PR #555, opened 2018-02-09
Fix cgroup hook race condition crash

minghui-liu:
#### Bug/feature Description
There is a race condition between different cgroup hooks accessing the cgroup file system. One hook can be reading the cgroup file system while another is deleting, causing hooks to crash.

#### Affected Platform(s)
All platform that cgroup hook supports

The constructor of `CgroupUtils` class calls `_get_assigned_cgroup_resources()`, which accesses thegroup file system, without obtaining a lock and causes the race condition.

#### Solution Description
1. In the `main()` method, hold the cgroup file lock from the moment we create a CgroupUtils instance to we finish executing hook event handlers. 
2. To support locking before a `CgroupUtils` instance is created, the parsing of configuration file is moved outside of `__init__` of `CgroupUtils`. The `parse_config_file` method is made a public and static method of the class. 

#### Checklist:

### DIFF
--- a/src/hooks/cgroups/pbs_cgroups.PY
+++ b/src/hooks/cgroups/pbs_cgroups.PY
@@ -505,11 +505,15 @@ class Lock(object):
     def __enter__(self):
         self.lockfd = open(self.path, 'w')
         fcntl.flock(self.lockfd, fcntl.LOCK_EX)
+        pbs.logmsg(pbs.EVENT_DEBUG4, "%s file lock acquired by %s" %
+                   (self.path, str(sys._getframe(1).f_code.co_name)))
 
     def __exit__(self, exc, val, trace):
         if self.lockfd:
             fcntl.flock(self.lockfd, fcntl.LOCK_UN)
             self.lockfd.close()
+        pbs.logmsg(pbs.EVENT_DEBUG4, "%s file lock released by %s" %
+                   (self.path, str(sys._getframe(1).f_code.co_name)))
 
 
 #
@@ -639,7 +643,7 @@ class HookUtils(object):
         """
         Return the event name for the supplied hook type.
         """
-        pbs.logmsg(pbs.EVENT_DEBUG4, "%s: Method called" % (caller_name()))
+        pbs.logmsg(pbs.EVENT_DEBUG4, "%s: Method called" % caller_name())
         if hooktype in self.hook_events:
             return self.hook_events[hooktype]['name']
         pbs.logmsg(pbs.EVENT_DEBUG4,
@@ -650,7 +654,7 @@ class HookUtils(object):
         """
         Return the handler for the supplied hook type.
         """
-        pbs.logmsg(pbs.EVENT_DEBUG4, "%s: Method called" % (caller_name()))
+        pbs.logmsg(pbs.EVENT_DEBUG4, "%s: Method called" % caller_name())
         if hooktype in self.hook_events:
             return self.hook_events[hooktype]['handler'] is not None
         return None
@@ -659,7 +663,7 @@ class HookUtils(object):
         """
         Call the appropriate handler for the supplied event.
         """
-        pbs.logmsg(pbs.EVENT_DEBUG4, "%s: Method called" % (caller_name()))
+        pbs.logmsg(pbs.EVENT_DEBUG4, "%s: Method called" % caller_name())
         pbs.logmsg(pbs.EVENT_DEBUG4, "%s: UID: real=%d, effective=%d" %
                    (caller_name(), os.getuid(), os.geteuid()))
         pbs.logmsg(pbs.EVENT_DEBUG4, "%s: GID: real=%d, effective=%d" %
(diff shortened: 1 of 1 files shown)


minghui-liu:
There is not a good way to write an automated test since this is a race condition bug. I ran a couple job arrays with 5000 short `date` jobs and attached the logs.

mike0042:
Two comments on your proposed changes...

The locking mechanism that currently exists is very simple, and supports only one lock at a time. In other words, if we establish a lock in one area of the code and later ask for another lock, we'll have a deadlock. The locking mechanism could incorporate a semaphore to keep track of the number of locks and releases. Only when the lock count is zero would the lock file actually be opened or closed. That would likely prevent the deadlock you encountered earlier and prevent future developers from running into the same problem. Then you could add a lock directly in _get_assigned_cgroup_resources() and not worry whether another portion of the hook already held the lock. It would simply increment and decrement the counter if the lock were already held.

My second comment involves our general approach to locking. If any cgroups hook is running, establishes a lock, releases it, and later establishes another lock, then there is no guarantee that another hook didn't grab the lock in between and modify the cgroup susbsystems. Would we be better off placing a lock at a higher level? Perhaps around the call to the handler itself? That would also eliminate the need for enhancing the locking mechanism since we'd only be creating/releasing the lock in one place.

mike0042:
It might also be useful to log some DEBUG4 messages within the lock code itself so we can identify who has the lock and when.

minghui-liu:
Placed the lock at a higher level (in `main()`) per suggestion of Mike. Attaching the new mom log for these changes.

[5000_job_test_mom_log.txt.zip](https://github.com/PBSPro/pbspro/files/1721439/5000_job_test_mom_log.txt.zip)

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -3844,33 +3848,36 @@ def main():
...
-        # Call the appropriate handler
-        if hooks.invoke_handler(event, cgroup, jobutil) is True:
-            pbs.logmsg(pbs.EVENT_DEBUG4, "%s: Hook handler returned success" %
-                       (caller_name()))
-            event.accept()
-        else:
-            pbs.logmsg(pbs.EVENT_DEBUG, "%s: Hook handler returned failure" %
-                       (caller_name()))
-            event.reject()
+            with Lock(cfg['cgroup_lock_file']):
```
mike0042: I think this block is indented too far. This portion of the code should be called regardless of whether the event has a vnode_list attribute.

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -2004,11 +1907,112 @@ class CgroupUtils(object):
             return os.path.join(subdir, jobid, cgfile)
         return os.path.join(subdir, jobid, (prefix + cgfile))
 
+    # Read the config file in json format
+    @staticmethod
```
mike0042: Rather than use the staticmethod decorator, should we just move the function outside of the class?

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -3844,33 +3848,36 @@ def main():
...
-        if hooks.invoke_handler(event, cgroup, jobutil) is True:
-            pbs.logmsg(pbs.EVENT_DEBUG4, "%s: Hook handler returned success" %
-                       (caller_name()))
-            event.accept()
-        else:
-            pbs.logmsg(pbs.EVENT_DEBUG, "%s: Hook handler returned failure" %
-                       (caller_name()))
-            event.reject()
+            with Lock(cfg['cgroup_lock_file']):
+                cgroup = CgroupUtils(hostname, vnode, cfg=cfg)
```
mike0042: Nicely done.

src/hooks/cgroups/pbs_cgroups.PY
minghui-liu: Because cgroup configurations are stored in `CgroupUtils` class, I think it makes more sense that `parse_config_file` stay in the same class. Convering `parse_config_file` a static function of the class made it possible to call the function without creating an instance.

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -1681,7 +1684,7 @@ class CgroupUtils(object):
     """
 
     def __init__(self, hostname, vnode, cfg=None, subsystems=None,
-                 paths=None, vntype=None, assigned_resources=None):
+                paths=None, vntype=None, assigned_resources=None):
```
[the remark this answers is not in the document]
minghui-liu: Fixed all PEP8 warnings.

src/hooks/cgroups/pbs_cgroups.PY
minghui-liu: You are right. Fixed.

src/hooks/cgroups/pbs_cgroups.PY
minghui-liu: Thank you.

src/hooks/cgroups/pbs_cgroups.PY
mike0042: Makes sense.

minghui-liu:
Thank you, @mike0042 @bhroam . I have rebased and squashed.

mike0042:
@PBSPro/pbspro-maintainers : Please merge this PR as @bhroam and I do not currently have permission.

PR #559, opened 2018-02-12
PP-1070: Revert ifl call pbs_statsched signature back to pre-mutlisched version.

Shrini-h:
#### Bug/feature Description
* *`pbs_statsched()` was changed to support multi-sched in [PP-746](https://pbspro.atlassian.net/browse/PP-746). However, this change broke backward compatibility. This Fix reverts the prototype without affecting multiched feature*
* `pbs_statsched()` prototype changed from
`struct batch_status *pbs_statsched(int, char *, struct attrl *, char *);`
to
`struct batch_status *pbs_statsched(int, struct attrl *, char *);`
*  [PP-1070](https://pbspro.atlassian.net/browse/PP-1070) 

#### Affected Platform(s)
* *All*

#### Solution Description
* *`pbs_statsched()` is made to return list of barch_status containing stat of all shceduler objects. The caller will loop through this list and use the one interested*

#### Checklist:

   - [ ] I have added  **manual test(s) to this pull request.
   - [x] I have explained why PTL is not appropriate** for this case.

#### Explanation for not adding new PTL test(s) to my commit:
"This commit is not eligible to have its own PTL test(s) as it does not introduce any new interface nor a bug that fails a test case. It reverts back an undocumented change in the prototype of `pbs_statsched()` IFL without affecting the multi-sched feature. However it now includes a PTL test case that was missed in a historic PR"


src/scheduler/fifo.c
```diff
@@ -2608,7 +2608,10 @@ update_svr_schedobj(int connector, int cmd, int alarm_time)
 		return 1;
 
 	/* Stat the scheduler to get details of sched */
-	ss = pbs_statsched(connector, sc_name, NULL, NULL);
+	ss = pbs_statsched(connector, NULL, NULL);
+	while(ss && strncmp(sc_name, ss->name, PBS_MAXSCHEDNAME)) {
```
bhroam: Check ss != NULL.  Also, I don't think you do an strncmp().  A strcmp() will be fine.  Lastly, you are iterating, use a for() loop.

src/scheduler/server_info.c
```diff
@@ -223,8 +223,12 @@ query_server(status *pol, int pbs_sd)
 		return NULL;
 	}
 
+	sched = pbs_statsched(pbs_sd, NULL, NULL);
+	while(sched && strncmp(PBS_DFLT_SCHED_NAME, sched->name, PBS_MAXSCHEDNAME)) {
+		sched = sched->next;
```
bhroam: Same as above, check sched != NULL.  Use strcmp() and use a for() loop.

src/scheduler/server_info.c
```diff
@@ -223,8 +223,12 @@ query_server(status *pol, int pbs_sd)
...
+	}
 
-	if ((sched = pbs_statsched(pbs_sd, PBS_DFLT_SCHED_NAME, NULL, NULL)) == NULL) {
+	if (sched == NULL) {
```
bhroam: Move this check up above the loop

src/server/req_stat.c
```diff
@@ -819,24 +819,15 @@ req_stat_sched(struct batch_request *preq)
...
-		}
-	} else {
-		psched = (pbs_sched *) GET_NEXT(svr_allscheds);
-		while (psched != (pbs_sched *) 0) {
-			rc = status_sched(psched, preq, &preply->brp_un.brp_status);
-			if (rc != 0) {
-				break;
-			}
-			psched = (pbs_sched *) GET_NEXT(psched->sc_link);
+	psched = (pbs_sched *) GET_NEXT(svr_allscheds);
```
bhroam: You're iterating, use a for() loop.

src/scheduler/server_info.c
```diff
@@ -223,8 +223,12 @@ query_server(status *pol, int pbs_sd)
 		return NULL;
 	}
 
+	sched = pbs_statsched(pbs_sd, NULL, NULL);
+	while(sched && strncmp(PBS_DFLT_SCHED_NAME, sched->name, PBS_MAXSCHEDNAME)) {
```
bhroam: Why are you only looking for the default scheduler?  Shouldn't you be looking for sc_name?  This is the policy object that will be used throughout the scheduling cycle.  A multisched should use its own policy object, not the default schedulers.

src/cmds/qmgr.c
```diff
@@ -2005,6 +2005,11 @@ display(int otype, int ptype, char *oname, struct batch_status *status,
 				printf("Server %s\n", status->name);
 		}
 		else if (otype == MGR_OBJ_SCHED) {
+			if(oname && *oname && strncmp(oname,status->name,PBS_MAXSCHEDNAME)) {
```
agrawalravi90: oname is passed into the function so please do the NULL checks on it at the top of the function itself, it'll make it more robust.

agrawalravi90: Also, please add a space after commas inside the strncmp() call

src/cmds/qmgr.c
```diff
@@ -2005,6 +2005,11 @@ display(int otype, int ptype, char *oname, struct batch_status *status,
...
+				status = status->next;
```
agrawalravi90: Can you please add a comment here explaining the logic? why is 'status' being moved to the sched object in the list?

src/lib/Libifl/pbs_ifl_wrap.c
```diff
@@ -5766,7 +5766,7 @@ SWIGINTERN PyObject *_wrap_pbs_statsched(PyObject *SWIGUNUSEDPARM(self), PyObjec
     SWIG_exception_fail(SWIG_ArgError(res3), "in method '" "pbs_statsched" "', argument " "3"" of type '" "char *""'");
   }
   arg3 = (char *)(buf3);
-  result = (struct batch_status *)pbs_statsched(arg1,PBS_DFLT_SCHED_NAME,arg2,arg3);
+  result = (struct batch_status *)pbs_statsched(arg1,arg2,arg3);
```
agrawalravi90: same nit here, please add spaces after commas

agrawalravi90:
Also, since this is a change in one of the IFL interfaces, can you please perform the following testing:
- Run all scheduler PTL tests in the repo and attach logs to the PR
- Do a basic manual test with valgrind to see that there are no leaks/memory errors. The test could be as simple as submitting a job and doing a qmgr -c 'list sched' to exercise the two most common paths that call pbs_statsched, while the scheduler is run via valgrind.

Thanks!

src/scheduler/server_info.c
suresh-thelkar: Thank you very much for catching this @bhroam. It looks like we added this as placeholder in the very beginning days of  multisched where the core interface was not developed at that time and forgot to replace it.  @Shrini-h  Since the core interface is now checked in, we should use sc_name only as suggested by Bhroam.  Also It would be better if we add couple of PTL test cases(There might already be test cases existing for these already just need to modify them to be suitable for Multisched) covering the following attributes for Multisched and add it our test suites TestMultipleSchedulers or TestSchedulerInterface whichever is fine.  Please let me know in case if you need any help on this.
* do_not_span_psets, only_explicit_psets
* preempt_targets_enable
* job_sort_formula_threshold
* opt_backfill_fuzzy

Also please run all the test cases from the following test suites in addition to what you are testing.
* TestMultipleSchedulers 
* TestSchedulerInterface

src/cmds/qmgr.c
Shrini-h: I agree NULL check for parameters should be done at the beginning of a function to make it robust.
However, it wont suit for this scenario. Due to following interrelated reasons:
   

1.  There a places of display() invocation with "oname" argument set to NULL, ex. in execute() of qmgr.c
2. I cannot fathom what to put under if (oname ==NULL) {} that is to be added at the begining.

src/cmds/qmgr.c
Shrini-h: ### Logic: 
Since now pbs_statsched(0 dumps batch_status for all the scheduler, we need to skip over the one we are not interested in. Hence advancing the status.

src/lib/Libifl/pbs_ifl_wrap.c
Shrini-h: I think we can ignore this file as it is auto-generated. I edited it by following the mutlit sched changes

src/scheduler/fifo.c
Shrini-h: 1.  `while (ss)` == `while (ss != NULL)` . I prefer former succinct format.
2.  why let `strcmp()` run amok if by chance we get a non-null termintated string. I say its safe to restrict it to "n"
3.  Ok I will change that, but I dont think there will be any difference in the compiled binary code

src/scheduler/server_info.c
Shrini-h: @suresh-thelkar I will change  the default scheduler to sc_name.  But I dont think PTL changes on Multisched will fall in the scope of this Bug or commit.

src/scheduler/server_info.c
Shrini-h: 1. while (ss) == while (ss != NULL) . I prefer former succinct format.
2. why let strcmp() run amok if by chance we get a non-null termintated string. I say its safe to restrict it to "n"
3. Ok I will change that, but I dont think there will be any difference in the compiled binary code

src/scheduler/server_info.c
Shrini-h: I think this check is at the right place. after looping the returned shcedstat batch_status linked list, we may end-up with `NULL` in `sched` if none of the sched stat match with sc_name

src/cmds/qmgr.c
agrawalravi90: Ok. I was a bit confused by the fact that none of the other objects do this so I looked a bit into how it's used. It seems like 'oname' is used by display just to ensure that only custom resources are displayed when somebody types "list resource" without specifying an object name. It seems like a very specific use case for just making sure that built-in resources don't get displayed. I don't think you need to bother with it at all for the sched object, the results are already filtered by the pbs_stat calls, display just needs to display them and not do any filtering itself, resources are an exception.

src/scheduler/server_info.c
suresh-thelkar: @Shrini-h 
As you are touching the  Multisched code it is always better to run the existing Multisched test cases as I mentioned above just to make sure that nothing is broken.  
Regarding the couple of new test cases that I have suggested, it is up to you whether to add as part of this bug or they can be added as part of QA Verification.

src/cmds/qmgr.c
agrawalravi90: Please see my comment to below about 'oname', I don't think you need this if statement at all.

Shrini-h: I disagree.  It is definitely required, as explained in the logic below, we will be getting all the sched objects, so if the user gives "list sched sc_abc",  the `status` will be linked list of "default_sched"-->"sc_abc"-->"sc__xyz"-->"etc".  So now in this case, except for "sc_abc" we need to advance the status to next.

However if the user gives "list sched",  we have execute the while loop for all the sched.

src/cmds/qmgr.c
Shrini-h: disagree, justified above

src/scheduler/server_info.c
Shrini-h: @suresh-thelkar , I agree to run the Multisched test cases here.  But lets take the PTL changes outside the scope from here

src/server/req_stat.c
Shrini-h: Ok I will change that, but I dont think there will be any difference in the compiled binary code

src/cmds/qmgr.c
agrawalravi90: Ok, so the server (pbs_statsched call) will not do any filtering, it is expected that it will happen at the client side, got it, thanks for clarifying.

Shrini-h:
[PTL logs for 1070.zip](https://github.com/PBSPro/pbspro/files/1720113/PTL.logs.for.1070.zip)

Shrini-h:
[val.zip](https://github.com/PBSPro/pbspro/files/1720613/val.zip)

src/scheduler/fifo.c
bhroam: 1. Follow scheduler style: use var != NULL.
2. The problem here is you're assuming the values will be PBS_MAXSCHEDNAME.  While this should be the case, the IFL interface doesn't limit the names it returns.  In the case where you have one string longer than the other with the same initial characters, you'll pass this test.  You should be in no danger of having non-null terminated strings.  IFL will terminate its strings.
3. A for() loop is a safer construct for iterating than a while() loop.  If at any point you use a continue in a while loop and forget to move your iteration variable along,  you'll get into an infinite loop.  In any case, follow scheduler style.  Use a for() loop.

Shrini-h:
image

@bhroam 
I have addressed your above comments in the latest commit

agrawalravi90:
@Shrini-h thanks for attaching the logs.
The valgrind logs are pretty big, there are errors which are definitely not related to the changes you made, but I can't go through all of them to see if there are any that can be related to the changes that you made. Could you please just run pbs_sched with "valgrind --show-possibly-lost=no --leak-check=full", perform the following 2 simple operations and post the output:
1) run qmgr -c 'list sched'
2) submit a job

Thanks.

Shrini-h:
[qmgr_out.txt](https://github.com/PBSPro/pbspro/files/1723421/qmgr_out.txt)

[val_out.txt](https://github.com/PBSPro/pbspro/files/1723422/val_out.txt)

Attaching the short Valgrind logs

bhroam:
@Shrini-h Please do not squash in the middle of a code review.  I makes it more difficult for your reviewers.

src/scheduler/fifo.c
```diff
@@ -2608,7 +2608,11 @@ update_svr_schedobj(int connector, int cmd, int alarm_time)
 		return 1;
 
 	/* Stat the scheduler to get details of sched */
-	ss = pbs_statsched(connector, sc_name, NULL, NULL);
+	for(ss = pbs_statsched(connector, NULL, NULL);
+			((ss != NULL) && strcmp(sc_name, ss->name));
+			ss = ss->next
+		);	//empty for loop
```
bhroam: Don't use C++ style comments.  Use /* */.  Also, since you are searching this multiple times, consider writing a function find_batch_status() and put it in libcmds.  It could be useful in the future.

src/scheduler/server_info.c
```diff
@@ -223,8 +223,12 @@ query_server(status *pol, int pbs_sd)
 		return NULL;
 	}
 
+	for(sched = pbs_statsched(pbs_sd, NULL, NULL);
+			((sched != NULL) && strcmp(sc_name, sched->name));
+			sched = sched->next
+		);	//empty for loop
```
bhroam: Same here, use /* */

src/server/req_stat.c
```diff
@@ -818,25 +818,16 @@ req_stat_sched(struct batch_request *preq)
...
-	} else {
-		psched = (pbs_sched *) GET_NEXT(svr_allscheds);
-		while (psched != (pbs_sched *) 0) {
-			rc = status_sched(psched, preq, &preply->brp_un.brp_status);
-			if (rc != 0) {
-				break;
-			}
-			psched = (pbs_sched *) GET_NEXT(psched->sc_link);
+	for(psched = (pbs_sched *) GET_NEXT(svr_allscheds);
+			(psched != (pbs_sched *) 0);
```
bhroam: Don't use (pbs_sched *) 0, use NULL

agrawalravi90:
Thanks for the logs. The changes look fine to me other than the comments that Bhroam already made.

**One question:** Why no PTL test?

Shrini-h:
> @Shrini-h Please do not squash in the middle of a code review. I makes it more difficult for your reviewers.

I'm sorry I didn't know force pushing an amended commit will purge the previous comments. Wont do that again.

Shrini-h:
> One question: Why no PTL test?

@agrawalravi90 , you mean why no PTL test case?
This is a BUG and not an RFE, the changes don't introduce any feature or interface. Hence I believe it doesn't  qualify for its own PTL test case.

Thank you for review and signoff

agrawalravi90:
"This is a BUG and not an RFE, the changes don't introduce any feature or interface. Hence I believe it doesn't qualify for its own PTL test case."

That's not true, each pull request, whether it addressed a BUG or an RFE, should be accompanied with PTL test(s), exceptions aside.

Shrini-h:
> That's not true, each pull request, whether it addresses a BUG or an RFE, should be accompanied with PTL test(s), exceptions aside.

@agrawalravi90 , Now I agree each PR whether BUG or RFE should be accompanied with PTL test(s) with exceptions

However the nature of the BUG this PR fixes makes it eligible for an exception. If you don't agree and have something in mind please help me know what PTL test this  PR should add?

agrawalravi90:
Here are 2 things that I can think of:
- you made a change to the qmgr 'list sched' codeline. So, one obvious thing to test for is whether qmgr "list sched" operation works correctly for 1) when there's only 1 scheduler and 2) when there are multiple schedulers. This can be easily tested using PTL.
- you made a change to the pbs_statsched() interface, which is actually an externally visible, stable and supported interface (one of the reasons why the original change to this interface was undesirable). So, another good thing to test for would be whether the pbs_statsched() interface itself works as expected, but I'm not sure if there's a way you can do that in PTL or not, so you'll have to investigate whether a PTL test for it makes sense or not.

Thinking more about how pbs_statsched() is an externally supported interface and you are changing its behavior (now it will return a list of all schedulers instead of just 1, which is how it was previously), I think this should be documented as an external change. I'd recommend modifying the multisched EDD (https://pbspro.atlassian.net/wiki/spaces/PD/pages/50947131/PP-337+Multiple+schedulers+servicing+the+PBS+cluster) to add this behavior change if its not already covered.

src/scheduler/fifo.c
Shrini-h: >Don't use C++ style comments. Use /* */.

The comment style is changed to /* */ in the second commit

>Also, since you are searching this multiple times, consider writing a function find_batch_status() and put it in libcmds. It could be useful in the future.

I understand it would be help full in future to add this new utility function,
However I like to pass on this for now, since I'm busy with other major RFE, and I'm repeating this code only twice for now.

But if you really insist, how about adding below macro in pbs_ifl.h or 
```
for_each_bach_status_sc( con, bs, sc_name )    for(\
                                             bs = pbs_statsched(con, NULL, NULL); \
                                             ((bs!= NULL) && strcmp(sc_name, bs->name));\
                                             bs = bs->next ) 
```

src/scheduler/server_info.c
Shrini-h: >Same here, use /* */

The comment style is changed to /* */ in the second commit

src/server/req_stat.c
Shrini-h: >Don't use (pbs_sched *) 0, use NULL

OK Changed it to NULL.  But if you see carefully this was existing code which was rearranged.

Shrini-h:
@agrawalravi90 
>you made a change to the qmgr 'list sched' codeline. So, one obvious thing to test for is whether qmgr "list sched" operation works correctly for 1) when there's only 1 scheduler and 2) when there are multiple schedulers. This can be easily tested using PTL.

this is already part of SmokeTest, results of which was already attached

>you made a change to the pbs_statsched() interface, which is actually an externally visible, stable and supported interface (one of the reasons why the original change to this interface was undesirable). So, another good thing to test for would be whether the pbs_statsched() interface itself works as expected, but I'm not sure if there's a way you can do that in PTL or not, so you'll have to investigate whether a PTL test for it makes sense or not.

That's what I meant when I said this PR is eligible for exception for no ptl tests

>Thinking more about how pbs_statsched() is an externally supported interface and you are changing its behavior (now it will return a list of all schedulers instead of just 1, which is how it was previously), I think this should be documented as an external change. I'd recommend modifying the multisched EDD (https://pbspro.atlassian.net/wiki/spaces/PD/pages/50947131/PP-337+Multiple+schedulers+servicing+the+PBS+cluster) to add this behavior change if its not already covered.

I checked the contents of man page in doc folder and also in function description regarding `pbs_statsched()` IFL call and they are already consistent with the behavior in this PR.

With reference to the multisched commit that introduced the change in the `pbs_statsched()` and its EDD
one thing shocks me!  that it made it to mainline without any mention in the docs or the EDD regarding the change in prototype which this PR reverses

suresh-thelkar:
@Shrini-h 
As we are aware that the signature of pbs_statsched() is going to be reverted as part of this ticket, there was no need to announce that in EDD. What @agrawalravi90  is suggesting is since now there is a possibility that pbs_statsched() can return statistics of more than one scheduler, clients i.e. whoever uses it should be aware of that and hence he is just recommending it to add to the EDD only about this change.

Shrini-h:
@suresh-thelkar 
The return type description in man page was already consistent with the new behavior.
However to make the behavior more visible I have now added "(s)" to "scheduler" and edited man page accordingly.  I will add a line to EDD

arungrover:
@Shrini-h to add to what Ravi mentioned about PTL tests, in this ticket you are also fixing a scheduler bug where now on each multi-sched will filter the output of pbs_statsched to use its own scheduler object. At the very least that change must have a PTL test associated to it. We couldn't catch this bug in our multi-sched checkin because there was no test for it.

agrawalravi90:
> this is already part of SmokeTest, results of which was already attached

Can you please point me to the particular test case which tests the qmgr list sched functionality? The SCHED object seems to appear in only one test in SmokeTest (test_job_sort_formula_threshold), and that is not in the context of qmgr list. I also went through the other tests that you posted logs for and I couldn't find any which tests the qmgr list sched functionality.

> That's what I meant when I said this PR is eligible for exception for no ptl tests

Did you actually investigate that the pbs_statsched() IFL interface cannot be tested via PTL? I know that PTL's IFL mode is not working a 100%, but it might work for a simple stat call. So, in case you haven't, please do investigate whether this is possible or not, you could take help from @hirenvadalia and @kjakkali .

For future reference, it's not enough to just say that your PR doesn't need PTL tests, contributors  are expected to explain why PTL tests are not needed (it's one of the checklist items). We are spending time here going back and forth on this when a simple explanation on the PR description would have clarified it. Although it does seem like this PR needs PTL tests.

bhroam:
"I'm sorry I didn't know force pushing an amended commit will purge the previous comments. Wont do that again."
@Shrini-h when you amend a commit, you are not creating a new one.  You are combining the previous commit with any new changes.  It's basically the same thing as making a new commit and squashing it into the previous commit.

"But if you really insist, how about adding below macro in pbs_ifl.h or

for_each_bach_status_sc( con, bs, sc_name )    for(\
                                             bs = pbs_statsched(con, NULL, NULL); \
                                             ((bs!= NULL) && strcmp(sc_name, bs->name));\
                                             bs = bs->next ) 
"

Make this a function and not a macro.  Also, accept just a batch_status and search it.  I know you are only repeating this code twice, but shared code is good.

doc/man3/pbs_statsched.3B
```diff
@@ -47,7 +47,7 @@
 .sp
 .B void pbs_statfree(\^struct batch_status *psj\^)
 .SH DESCRIPTION
-Issue a batch request to obtain the status of PBS scheduler.
+Issue a batch request to obtain the status of (all) PBS scheduler(s).
```
bhroam: Remove the () around all.

doc/man3/pbs_statsched.3B
```diff
@@ -126,5 +126,5 @@ When the batch request generated by
 .B pbs_statsched()
 has been completed successfully by the PBS server, 
 .B pbs_statsched() 
-will return a pointer to a batch_status structure.
+will return a pointer to a list of batch_status structure.
```
bhroam: Add (s) at the end.

Shrini-h:
> @Shrini-h to add to what Ravi mentioned about PTL tests, in this ticket you are also fixing a scheduler bug where now each multi-sched will filter the output of pbs_statsched to use its own scheduler object. At the very least that change must have a PTL test associated to it. We couldn't catch this bug in our multi-sched checkin because there was no test for it.

@arungrover 
I'm sorry, I don't agree. I'm not fixing any scheduler bug here or multi-sched RFE here. The fix reverts the pbs_statsched() changes without affecting the multi-sched feature. Whatever PTL test cases needed to test the multi-sched features should have been already checked in with mulit-sched feature itself.
All I have to do is to ensure not to break any of the mult-sched or sched related existing PTL test case, which I already did, they all passed successfully and have already attached the logs and reports above in this PR.  

Further for adding new PTL test cases related to multi-sched I suggest you create a new ticket.

Thank you

Shrini-h:
> Can you please point me to the particular test case which tests the qmgr list sched functionality? The SCHED object seems to appear in only one test in SmokeTest (test_job_sort_formula_threshold), and that is not in the context of qmgr list. I also went through the other tests that you posted logs for and I couldn't find any which tests the qmgr list sched functionality.

@agrawalravi90 
This PR does not introduce `qmgr list sched`. I wont be writin any PTL test case on it here. I have seen in the PTL log that `qmgr list sched` command is getting executed for every test case in SmokeTest, In fact `qmgr list sched` crashed SmokeTest during my development of this PR. When you grep `list sched` in the attached ptl log file "ptl_out.txt" it hits for 379 lines !!
If you are interested in getting PTL test cases written for `qmgr list sched` or `pbs_statsched()` please consult the author of PR's that introduced them.

>Did you actually investigate that the pbs_statsched() IFL interface cannot be tested via PTL? I know that PTL's IFL mode is not working a 100%, but it might work for a simple stat call. So, in case you haven't, please do investigate whether this is possible or not, you could take help from @hirenvadalia and @kjakkali .

Just checked with @hirenvadalia , and pbs_statsched() IFL interface testing through PTL's IFL mode is not yet available.

> For future reference, it's not enough to just say that your PR doesn't need PTL tests, contributors are expected to explain why PTL tests are not needed (it's one of the checklist items). We are spending time here going back and forth on this when a simple explanation on the PR description would have clarified it. Although it does seem like this PR needs PTL tests.

Hope the below addition at the bottom of PR description handles this

image

doc/man3/pbs_statsched.3B
Shrini-h: having () around all implies for multiple PBS schedulers. I say we keep the () around all

doc/man3/pbs_statsched.3B
Shrini-h: will add only 's' instead.

Shrini-h:
For reviewers determined to get Multisched PTL test cases done through this humble tiny PR, I suggest you create a separate ticket for it. 

Please cut me some slack, I got to get back to major ongoing works.

Thank You
Shrini

Shrini-h:
> Make this a function and not a macro. Also, accept just a batch_status and search it. I know you are only repeating this code twice, but shared code is good.

@bhroam 
I feel a function just to traverse a linked list is an overkill and cumbersome. As per my knowledge link list traversal in most OSS is best handled as a macro.

doc/man3/pbs_statsched.3B
bhroam: () means a parenthetical remark.  In this case, all is part of the sentence.  Please remove them.

arungrover:
@Shrini-h While the code works and you have attached the proof of testing, we maintainers (myself, @agrawalravi90  and @bhroam) think that it still needs more changes in terms of technical debt that it will accrue and we will not be able to take it in as it is. We understand that you are busy with other important work so you can come back to this pull request and complete it when you have more time available. Till then the PR can stay open as is.

Shrini-h:
@arungrover , Thank you for understanding my position, 
I was hurrying on this as it is a blocker for another project which uses this IFL
I will come back to this PR after completing some major work.

bhroam:
@Shrini-h 
Macros have their place, but I don't think this is one of them.  Macros should be short pieces of simple code.  They also have their own complexities.  Passing simple arguments like X into a macro might work, but passing X+1 might not.  If you pass X+1 into a function, the math happens first before you call the function.  Please write this function.  It leaves the code in a better place than where you left it.  This is always the goal of any pull request.

subhasisb:
@Shrini-h, I agree with @bhroam - we try not to add complicated macros to our codebase. Also, that sort of makes debugging more problematic. Besides, a function is much more readable. Also reusable in the future by others.

src/scheduler/server_info.c
Shrini-h: Now considering the impact of this new found bug I have filed a new bug https://pbspro.atlassian.net/browse/PP-1238

doc/man3/pbs_statsched.3B
Shrini-h: removed the ()

Shrini-h:
> Make this a function and not a macro. Also, accept just a batch_status and search it. I know you are only repeating this code twice, but shared code is good.

done

Shrini-h:
rebased to latest. I apologize for the unintentional squash

Shrini-h:
Now considering the impact of this new found bug I have filed a new bug https://pbspro.atlassian.net/browse/PP-1238

src/lib/Libcmds/batch_status_find.c
```diff
@@ -0,0 +1,72 @@
...
+ * trademark licensing policies.
+ *
+ */
+
+#include <stdio.h>
+
+#include "pbs_ifl.h"
+
+/**
+ * @file	batch_status_find
```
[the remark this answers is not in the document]
Shrini-h: I named it batch_status_find.c as there is neighboring file named batch_status_sort.c [which contains bs_isort()] , Now if I change it to find_batch_status.c, the file names in the folder I feel will look inconsistent.

With that, please let me know I still need to change the file name to find_batch_status.c

src/lib/Libcmds/batch_status_find.c
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * @param[in] bs - batch_status linked list
+ * @param[in] name - name of the batch_status structure to be searched
+ *
+ *
+ * @return 	batch_status structure handle
+ * @retval	batch_status structure pointer with given name or NULL
+ *
+ */
+struct batch_status *find_batch_status(struct batch_status *bs, const char * name) {
```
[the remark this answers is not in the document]
Shrini-h: done.

src/lib/Libcmds/batch_status_find.c
```diff
@@ -0,0 +1,72 @@
...
+
+	if (name == NULL)
+		return NULL;
+
+	for(;
+				((bs != NULL) && strcmp(name, bs->name));
+				bs = bs->next
+			);	/* empty for loop */
```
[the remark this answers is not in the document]
Shrini-h: done.

src/lib/Libcmds/batch_status_find.c
```diff
@@ -0,0 +1,72 @@
...
+#include <stdio.h>
+
+#include "pbs_ifl.h"
+
+/**
+ * @file	batch_status_find
+ *
+ * @brief
+ *	batch_status_find - find a batch_status in a batch_status structures list.
+ */
```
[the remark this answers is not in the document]
Shrini-h: done.

agrawalravi90:
""This commit is not eligible to have its own PTL test(s) as it does not introduce any new interface nor a bug that fails a test case. It reverts back an undocumented change in the prototype of pbs_statsched() IFL without affecting the multi-sched feature.""

In addition to reverting the pbs_statsched() prototype, your changes also enable stat'ing multi-sched info from qmgr & pbs_statsched(). It's important to have a PTL test which ensures that the change in behavior of qmgr -c "list sched" from before multi-sched to after is correct. Right now, there's no PTL test which tests that.  It's unfortunate that the original multisched PR which added the same functionality as this PR didn't include that PTL test. But let's not repeat the same mistake twice.

Shrini-h:
> your changes also enable stat'ing multi-sched info from qmgr & pbs_statsched()

Factually this PR doesn't do any thing new wrt stating multi-sched.

> It's important to have a PTL test which ensures that the change in behavior of qmgr -c "list sched" from before multi-sched to after is correct. Right now, there's no PTL test which tests that. It's unfortunate that the original multisched PR which added the same functionality as this PR didn't include that PTL test. But let's not repeat the same mistake twice.

To cover up the mistake I can write PTL test to test qmgr -c "list sched".

Shrini-h:
> But let's not repeat the same mistake twice.

@agrawalravi90 
I disagree. I do not understand how we are repeating a mistake with this PR.
I can write PTL test to test qmgr -c "list sched" which was missed in a historic PR.

bhroam:
@Shrini-h I disagree that you are not doing anything new wrt multi-sched.  Prior to the change to pbs_statsched(), it would return one scheduler (the default one).  With your change, it now returns all of the schedulers.

I'm happy with the changes you made.  I agree with @agrawalravi90 that this change in pbs_statsched() requires a test.  I'll sign off after that is written.

Shrini-h:
I've started writing PTL test named "test_list_sched" under the test suite "TestQmgr" in test/tests/functional/pbs_qmgr.py

I'm trying to cover below scenarios in it for qmgr -c "list sched"
1. No schedulers present.
2. Only Default sched exist
3. More than One sched exist

For each scenario I will call below cmds
1. qmgr -c "list sched"
2. qmgr -c "list sched default"
3. qmgr -c "list sched <sc_name>"      # for each sched
4. qmgr -c "list sched <invalid_sc_name>"

Please comment if you think of any other scenarios I need to cover or If the test case should be in a different test suite.

Thank you

Shrini-h:
Just discovered that scenario 1. "No schedulers present" is not valid as we cannot delete the "default" sched. Hence will cover below remaining scenarios

1. Only Default sched exist
2. More than One sched exist

Shrini-h:
I have added the PTL test case TestQmgr.test_list_sched, I have attached the PTL out put log here
[PTL_OUT_test_list_sched.txt](https://github.com/PBSPro/pbspro/files/1761750/PTL_OUT_test_list_sched.txt)

test/tests/functional/pbs_qmgr.py
```diff
@@ -159,3 +208,143 @@ def test_multi_attributes(self):
...
             self.assertTrue(msg in e.msg[0])
         self.server.expect(NODE, 'queue', op=UNSET, id=self.mom.shortname)
+
+    @tags('multisched')
+    def test_list_sched(self):
+        """
+        Test to verify that qmgr list sched works for multiple schedulers
+        """
+        # Scenario 1: only default sched present
+        # Scenario 2: multiple schedulers present
```
bhroam: These comments should be part of the doc string.  Also if you are doing two distinct scenarios in one test, it is probably better to make two tests.

test/tests/functional/pbs_qmgr.py
```diff
@@ -47,6 +47,55 @@ class TestQmgr(TestFunctional):
...
+    def setup_sc4(self):
+        a = {'partition': 'P5',
+             'sched_host': self.server.hostname,
+             'sched_port': '15053'}
+        self.server.manager(MGR_CMD_CREATE, SCHED,
+                            a, id="sc4")
+        self.scheds['sc4'].create_scheduler()
+        self.scheds['sc4'].start()
+        self.server.manager(MGR_CMD_SET, SCHED,
+                            {'scheduling': 'True'}, id="sc4", expect=True)
```
bhroam: If you are going to copy all this code over from the multisched test plan, just add your test to the multisched test plan.  It belongs there more than here anyway.

test/tests/functional/pbs_qmgr.py
```diff
@@ -159,3 +208,143 @@ def test_multi_attributes(self):
...
+        """
+        Test to verify that qmgr list sched works for multiple schedulers
+        """
+        # Scenario 1: only default sched present
+        # Scenario 2: multiple schedulers present
+
+        
+        # Scenario 1: only default sched present
+        self.server.manager(MGR_CMD_LIST, SCHED, id="default", expect=True,
+                                runas=ROOT_USER)
```
bhroam: Why are you running as ROOT_USER?  If you need to run as root, use sudo=True, but none of these commands require to run as root.  You can list sched as a normal user.

test/tests/functional/pbs_qmgr.py
```diff
@@ -159,3 +208,143 @@ def test_multi_attributes(self):
...
+        self.server.manager(MGR_CMD_LIST, SCHED, id="sc1", expect=True,
+                                runas=ROOT_USER)
+            
+        self.server.manager(MGR_CMD_LIST, SCHED, id="sc2", expect=True,
+                                runas=ROOT_USER)
+            
+        self.server.manager(MGR_CMD_LIST, SCHED, id="sc3", expect=True,
+                                runas=ROOT_USER)
+            
+        self.server.manager(MGR_CMD_LIST, SCHED, id="sc1,sc3,default", expect=True,
```
bhroam: I think the format of manager with multiple id's is to pass it a list instead of a comma delimited string

test/tests/functional/pbs_qmgr.py
```diff
@@ -159,3 +208,143 @@ def test_multi_attributes(self):
...
+            self.server.manager(MGR_CMD_LIST, SCHED, id="invalide_scname", expect=True,
+                                runas=ROOT_USER)
+        except PbsManagerError as e:
+            err_msg = "Unknown Scheduler"
+            self.assertTrue(err_msg in e.msg[0],
+                            "Error message is not expected")
+
+        try:
+            self.server.manager(MGR_CMD_LIST, SCHED, id="sc1,sc4", expect=True,
+                                runas=ROOT_USER)
```
bhroam: Same as above, use a list

test/tests/functional/pbs_qmgr.py
```diff
@@ -159,3 +208,143 @@ def test_multi_attributes(self):
...
+        self.server.manager(MGR_CMD_DELETE, SCHED, id="sc1", expect=True,
+                                runas=ROOT_USER)
+
+        try:
+            self.server.manager(MGR_CMD_LIST, SCHED, id="sc1", expect=True,
+                                runas=ROOT_USER)
+        except PbsManagerError as e:
+            err_msg = "Unknown Scheduler"
+            self.assertTrue(err_msg in e.msg[0],
+                            "Error message is not expected")
```
bhroam: You've already done this test.  You deleted sched sc3 and tested you couldn't list it.  No need to do it again for sc1.

test/tests/functional/pbs_qmgr.py
```diff
@@ -159,3 +208,143 @@ def test_multi_attributes(self):
...
+
+        self.server.manager(MGR_CMD_LIST, SCHED, id="default,sc4", expect=True,
+                                runas=ROOT_USER)
```
bhroam: Same as above, use a list.

bhroam:
Codacy is complaining about trailing whitespace.  Please run your test through autopep8.

Shrini-h:
@bhroam  I have tried addressing all your comments on new PTL test case.
I have moved the scenario 2 into a separate test case as TestMultipleSchedulers.test_list_multi_sched
[PTL_OUT.txt](https://github.com/PBSPro/pbspro/files/1765641/PTL_OUT.txt)

test/tests/functional/pbs_qmgr.py
```diff
@@ -159,3 +159,18 @@ def test_multi_attributes(self):
...
             self.assertTrue(msg in e.msg[0])
         self.server.expect(NODE, 'queue', op=UNSET, id=self.mom.shortname)
+
+    def test_list_default_sched(self):
+        """
+        Test to verify that qmgr list sched works when only default sched present
+        """
+        self.server.manager(MGR_CMD_LIST, SCHED, id="default", expect=True)
+
+        self.server.manager(MGR_CMD_LIST, SCHED, id="", expect=True)
```
visheshh: No need to give empty string as id

test/tests/functional/pbs_multi_sched.py
```diff
@@ -580,6 +580,60 @@ def test_run_limts_per_scheduler(self):
...
+    def test_list_multi_sched(self):
+        """
+        Test to verify that qmgr list sched works when multiple schedulers are present
+        """
+
+        self.setup_sc1()
+        self.setup_sc2()
+        self.setup_sc3()
+
+        self.server.manager(MGR_CMD_LIST, SCHED, id="", expect=True)
```
visheshh: No need to give empty string as id

test/tests/functional/pbs_qmgr.py
Shrini-h: done

test/tests/functional/pbs_multi_sched.py
Shrini-h: done

test/tests/functional/pbs_qmgr.py
```diff
@@ -209,141 +160,18 @@ def test_multi_attributes(self):
             self.assertTrue(msg in e.msg[0])
         self.server.expect(NODE, 'queue', op=UNSET, id=self.mom.shortname)
 
-    @tags('multisched')
-    def test_list_sched(self):
+    def test_list_default_sched(self):
```
bhroam: This test duplicates the test you put just moved into the multisched test plan.  There is no need for it here.

bhroam:
@Shrini-h you have a merge commit.  Merge commits are very bad.  When you squash and rebase, they will make it look like all the code you merged in from master will look like your own.  You need to reset your branch.  Unfortunately you checked in things after your merge commit.  This will add extra complication.  I suggest doing the following.  Since the only thing changed is the test plan, move the test plan aside.  Do a git log --oneline.  Look for the commit right before your merge commit.  Do a git reset --hard <commit hash>.  Copy the PTL test back into place and commit it.  Force push that to your fork.  To avoid doing this in the future, please follow our pages on rebasing.  You need to update your master branch and then rebase your dev branch from master.

src/lib/Libcmds/batch_status_find.c
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * @param[in] bs - batch_status linked list
+ * @param[in] name - name of the batch_status structure to be searched
+ *
+ *
+ * @return 	batch_status structure handle
+ * @retval	batch_status structure pointer with given name or NULL
+ *
+ */
+struct batch_status *find_batch_status(struct batch_status *bs, const char *name) {
```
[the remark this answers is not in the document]
agrawalravi90: also, the style in PBSPro for function headers is to put the return type on a separate line, and then the rest of the header below that. Also, the opening brace should go on a separate line as well (basically ANSI C standard). So, something like this:
struct batch_status *
find_batch_status(struct batch_status *bs, const char *name)
{
...
}

Please see the the "formatting" section of our coding standards for more info: https://pbspro.atlassian.net/wiki/spaces/DG/pages/5537826/Coding+Standards#CodingStandards-ProgrammingConstructs

src/lib/Libcmds/batch_status_find.c
```diff
@@ -0,0 +1,69 @@
...
+
+	if (name == NULL)
+		return NULL;
+
+	for( ; ((bs != NULL) && strcmp(name, bs->name)); bs = bs->next )
```
agrawalravi90: Another style nit: remove the spaces after '(' and before ')', see the first bullet under "Use of White Space" section of the coding standards.

test/tests/functional/pbs_multi_sched.py
```diff
@@ -580,6 +580,60 @@ def test_run_limts_per_scheduler(self):
...
+        """
+        Test to verify that qmgr list sched works when multiple schedulers are present
+        """
+
+        self.setup_sc1()
+        self.setup_sc2()
+        self.setup_sc3()
+
+        self.server.manager(MGR_CMD_LIST, SCHED, expect=True)
+
```
agrawalravi90: why the blank lines between manager calls? PEP8 says "Use blank lines in functions, sparingly, to indicate logical sections."

test/tests/functional/pbs_multi_sched.py
```diff
@@ -580,6 +580,60 @@ def test_run_limts_per_scheduler(self):
...
+        try:
+            self.server.manager(MGR_CMD_LIST, SCHED, id="sc3")
+        except PbsManagerError as e:
+            err_msg = "Unknown Scheduler"
+            self.assertTrue(err_msg in e.msg[0],
+                            "Error message is not expected")
+
+        self.server.manager(MGR_CMD_LIST, SCHED, expect=True)
+
```
agrawalravi90: Same nit here, please remove the extra blank lines

test/tests/functional/pbs_multi_sched.py
```diff
@@ -580,6 +580,60 @@ def test_run_limts_per_scheduler(self):
...
+        Test to verify that qmgr list sched works when multiple schedulers are present
+        """
+
+        self.setup_sc1()
+        self.setup_sc2()
+        self.setup_sc3()
+
+        self.server.manager(MGR_CMD_LIST, SCHED, expect=True)
+
+        self.server.manager(MGR_CMD_LIST, SCHED, id="default", expect=True)
```
agrawalravi90: What would 'expect=True' do for a qmgr list?

test/tests/functional/pbs_multi_sched.py
```diff
@@ -580,6 +580,60 @@ def test_run_limts_per_scheduler(self):
...
+        except PbsManagerError as e:
+            err_msg = "Unknown Scheduler"
+            self.assertTrue(err_msg in e.msg[0],
+                            "Error message is not expected")
+
+        self.server.manager(MGR_CMD_LIST, SCHED, expect=True)
+
+        self.server.manager(MGR_CMD_LIST, SCHED, id="default", expect=True)
+
+        self.server.manager(MGR_CMD_LIST, SCHED, id="sc1", expect=True)
```
agrawalravi90: What's the rationale behind these manager list calls? You already called them at the beginning of the function.

test/tests/functional/pbs_multi_sched.py
```diff
@@ -580,6 +580,60 @@ def test_run_limts_per_scheduler(self):
...
+        # delete sc1 sched
+        self.server.manager(MGR_CMD_DELETE, SCHED, id="sc1", sudo=True)
+
+        try:
+            self.server.manager(MGR_CMD_LIST, SCHED, id="sc1")
+        except PbsManagerError as e:
+            err_msg = "Unknown Scheduler"
+            self.assertTrue(err_msg in e.msg[0],
+                            "Error message is not expected")
+
```
agrawalravi90: One more thing that you might wanna test is calling manager() list with id = None, that will cause it to stat all scheduler objects.

Shrini-h:
> @Shrini-h you have a merge commit. Merge commits are very bad. When you squash and rebase, they will make it look like all the code you merged in from master will look like your own. You need to reset your branch. Unfortunately you checked in things after your merge commit. This will add extra complication. I suggest doing the following. Since the only thing changed is the test plan, move the test plan aside. Do a git log --oneline. Look for the commit right before your merge commit. Do a git reset --hard . Copy the PTL test back into place and commit it. Force push that to your fork. To avoid doing this in the future, please follow our pages on rebasing. You need to update your master branch and then rebase your dev branch from master.

@bhroam  Thanks for the caveat. If we rebase the master instead of merge, then we have to force push if there are conflicts. This force push will unintentionally alter the commits in the middle of the review (which is undersirable https://github.com/PBSPro/pbspro/pull/559#issuecomment-365741295 ). That's what happened when I commented this 
https://github.com/PBSPro/pbspro/pull/559#issuecomment-367413970

> rebased to latest. I apologize for the unintentional squash


Now I did an experiment. I cloned this branch and then rebased it to the master, surprisingly the rebase command handles the merge perfectly without any creeping of master changes into my commits, it also deleted the merge commit automatically during rebase, 
I have pushed it here https://github.com/Shrini-h/pbspro/tree/test_rebase_after_merge if you want to inspect.

May be a squash before a rebase does evil things when there are merge commits. Try rebase before squash next time.
git rebase is really intelligent. http://blog.woobling.org/2009/05/git-rebase-considered-awesome.html

test/tests/functional/pbs_qmgr.py
Shrini-h: This is for the scenario 1: Only default sched exist. Since there is no multiple scheds, I feel its appropriate here. Please let me know If you still want to remove this test case.

src/lib/Libcmds/batch_status_find.c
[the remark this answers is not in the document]
Shrini-h: > Nit: The file is called "batch_status_find" but the function is called "find_batch_status", I'd recommend making them consistent.

The same was discussed and justified here: https://github.com/PBSPro/pbspro/pull/559#pullrequestreview-98367934
and the function name "find_batch_status()" was suggested by @bhroam here: https://github.com/PBSPro/pbspro/pull/559#pullrequestreview-96652722

>also, the style in PBSPro for function headers is to put the return type on a separate line, and then the rest of the header below that. Also, the opening brace should go on a separate line as well (basically ANSI C standard). So, something like this:
struct batch_status *
find_batch_status(struct batch_status *bs, const char *name)
{ ...
Please see the the "formatting" section of our coding standards for more info: 

Actually for being on safer side, I used existing mainline code as template for writing this code (assuming we strictly adhered to the guidelines in existing mainline code). Used below function as template: https://github.com/PBSPro/pbspro/blob/3798b6bd739da79bfa5405b342a5d6b28ee86d81/src/lib/Libcmds/batch_status_sort.c#L61-L62

Anyways I will correct it in both the places.

src/lib/Libcmds/batch_status_find.c
Shrini-h: Ok will delete it.

test/tests/functional/pbs_multi_sched.py
Shrini-h: blank lines are used sparingly, since each manager call executes the targeted test objective, I consider each of them equivalent to a logical section.

test/tests/functional/pbs_multi_sched.py
Shrini-h: I don't know, it re-verifies the attributes?. You want me to remove it?

test/tests/functional/pbs_multi_sched.py
Shrini-h: blank lines are used sparingly, since each manager call executes the targeted test objective, I consider each of them equivalent to a logical section.

test/tests/functional/pbs_multi_sched.py
Shrini-h: Just verifying they all work after deleting sc3 here 
https://github.com/Shrini-h/pbspro/blob/e27d7ffa81f8fc6c44d5a54b52a04bd5f1bd475f/test/tests/functional/pbs_multi_sched.py#L610

test/tests/functional/pbs_multi_sched.py
Shrini-h: Its already done here 
https://github.com/Shrini-h/pbspro/blob/e27d7ffa81f8fc6c44d5a54b52a04bd5f1bd475f/test/tests/functional/pbs_multi_sched.py#L592
and here 
https://github.com/Shrini-h/pbspro/blob/e27d7ffa81f8fc6c44d5a54b52a04bd5f1bd475f/test/tests/functional/pbs_multi_sched.py#L613

src/lib/Libcmds/batch_status_find.c
```diff
@@ -57,12 +57,14 @@
...
-struct batch_status *find_batch_status(struct batch_status *bs, const char *name) {
+struct batch_status *
+find_batch_status(struct batch_status *bs, const char *name)
+{
 
 	if (name == NULL)
 		return NULL;
 
-	for( ; ((bs != NULL) && strcmp(name, bs->name)); bs = bs->next )
+	for (;((bs != NULL) && strcmp(name, bs->name)); bs = bs->next)
```
bhroam: Please add the space after the ';' back in please.

test/tests/functional/pbs_multi_sched.py
agrawalravi90: I see, that's cool. But I think that you also need to actually check that the manager call did the right thing. A manager list call updated the respective scheduler object's 'attribute' dictionary. I think the test should check that the dictionaries got populated correctly, at least for one manager call.

test/tests/functional/pbs_multi_sched.py
agrawalravi90: verifies them against what? usually 'expect' makes sense in context of a set operation as it can verify that the attributes got updated correctly. For a list, there's nothing to verify against. So ya, I think you can remove them.

src/lib/Libcmds/batch_status_find.c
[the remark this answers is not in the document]
agrawalravi90: "The same was discussed and justified here: #559 (review)
and the function name "find_batch_status()" was suggested by @bhroam here: #559 (review)"

Please use the same name for both the file and the function, no matter which name you like.

agrawalravi90:
"The same was discussed and justified here: #559 (review)
and the function name "find_batch_status()" was suggested by @bhroam here: #559 (review)"

Please use the same name for both the file and the function, no matter which name you like.

bhroam:
@Shrini-h Having merge commits is playing with fire.  No matter what your test has shown, I don't want to have to make sure your merge commit is handled correctly.  Please git reset like I have asked.

test/tests/functional/pbs_qmgr.py
bhroam: I still think this test is a repeat of what is in the multisched test plan.  It isn't needed here.

Shrini-h:
> Please use the same name for both the file and the function, no matter which name you like.

Why should file name be same as the function name?. They do convey what they do.

src/lib/Libcmds/batch_status_find.c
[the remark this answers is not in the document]
Shrini-h: >Please use the same name for both the file and the function, no matter which name you like.

Why should file name be same as the function name?. They do convey what they do.

Shrini-h:
>Having merge commits is playing with fire. No matter what your test has shown, I don't want to have to make sure your merge commit is handled correctly. Please git reset like I have asked.

@bhroam  For allaying your fears I have rebased to the master and force pushed the commits. However I did not have to do reset --hard followed by manuall merge as you suggested, a simple "git rebase master" worked flawlessly.

agrawalravi90:
> Why should file name be same as the function name?. They do convey what they do.

I think it would be less confusing, that file literally will only ever have that one function, so it'd be convenient if the function was called the same as the file so that people don't mix the names up, do you have a good reason to not make them the same?

test/tests/functional/pbs_qmgr.py
Shrini-h: OK deleted this test case.

test/tests/functional/pbs_multi_sched.py
Shrini-h: Added lines to test case to check updation of attributes dictionary of sched sc2 from list sched command

Shrini-h:
> I think it would be less confusing,
so it'd be convenient if the function was called the same as the file so that people don't mix the names up

I don't know how its confusing or inconvenient. Try in an IDE.

>that file literally will only ever have that one function,

We cant say that for sure, some one might add more functions in the future.

> do you have a good reason to not make them the same?

Yes I have many:
1. Till now there are no good reason to make them the same.
2. Coding guidelines doesn't say file name should be same as lone function name.
3. There's already exists in the mainline a file named batch_status_sort.c that has lone function bs_isort() which going by your view literally will only ever have that one function,
4. Bhroam has given the function name and I cant change that. Changing the file name will make the neighboring file look odd or inconsistent

test/tests/functional/pbs_multi_sched.py
```diff
@@ -715,7 +715,13 @@ def test_list_multi_sched(self):
 
         self.server.manager(MGR_CMD_LIST, SCHED, id="sc1")
 
-        self.server.manager(MGR_CMD_LIST, SCHED, id="sc2")
+        dir_path = '/var/spool/pbs/sched_dir'
```
bhroam: Use os.path.join(), don't use direct paths.

test/tests/functional/pbs_multi_sched.py
```diff
@@ -715,7 +715,13 @@ def test_list_multi_sched(self):
...
+        a = {'partition': 'P2',
+             'sched_priv': os.path.join(dir_path, 'sched_priv_sc2'),
+             'sched_log': os.path.join(dir_path, 'sched_logs_sc2'),
+             'sched_host': self.server.hostname,
+             'sched_port': '15051'}
+        self.server.manager(MGR_CMD_LIST, SCHED, a, id="sc2", expect=True)
```
bhroam: Don't use expect=True on a MGR_CMD_LIST

test/tests/functional/pbs_multi_sched.py
Shrini-h: This Line was directly copied from existing mainline code 
https://github.com/PBSPro/pbspro/blob/master/test/tests/functional/pbs_multi_sched.py#L59

Anyways I will change in both places

test/tests/functional/pbs_multi_sched.py
Shrini-h: Ok removed expect=True

Shrini-h:
I have implemented the review comments.

If you have any further comments please provide asap. Else I appreciate your signoffs.

Thank you

test/tests/functional/pbs_multi_sched.py
agrawalravi90: It's a little confusing, but expect=True with a list of attributes would actually verify that qmgr list returned the correct attribute values. If you don't wanna depend on expect=True, then you could also explicitly check that scheduler.attributes was populated correctly. Up to you, but your code was previously correct.

agrawalravi90:
@Shrini-h , since we were in disagreement over the naming of find_batch_status() and its new file batch_status_find.c, I had a chat with @PBSPro/pbspro-maintainers and we voted for the following solution:
- put the new function in batch_status_sort.c, rename the file to batch_status.c and rename the function to bs_find().

Bhroam wanted me to let you know that he's okay with the renaming the function from find_batch_status to bs_find.

This, along with the "expect=True" comment are now the only outstanding comments from my side, I'll sign-off after they are addressed, thanks for your patience.

test/tests/functional/pbs_multi_sched.py
Shrini-h: Ok I put back the expect=True

Shrini-h:
> put the new function in batch_status_sort.c, rename the file to batch_status.c and rename the function to bs_find().

That will have my vote too. However, the file name batch_status_utils.c would have been more apt.

Anyways I have done the two changes.
If you have any further comments please provide asap. Else I appreciate your signoffs.
I will rebase and squash after signoff.
Thank you

agrawalravi90:
@Shrini-h , Thanks for making the changes, before we move forward, please do the following:
- Squash all your commits into one SIGNED commit and push it to your fork, do NOT update your branch with master yet.
- Once you push to your fork, Travis and Appveyor will finally be able to test your changes, they haven't been running on your PR because of unsigned commits.
- If there are any issues reported by Travis/Appveyor, please fix them
- Finally, bring your branch up to date with master. github shows that there will be merge conflicts when you bring your branch to latest, so expect to see conflicts and resolve them.
- Push again. 

After this is done, I'll take a look at it again and sign-off if all looks good. Again, thanks for your patience.

agrawalravi90:
Actually Travis and Appveyor are not able to do anything because of the merge conflicts, you'll need to resolve them, please go ahead and update your branch with master, resolve the conflicts and upload to fork

Shrini-h:
PR is ready for merge.

agrawalravi90:
Please post new test logs Shrini, changes look fine to me otherwise.

Btw, you still need a second sign-off, Bhroam is on vacation, please find another reviewer for this PR.

Shrini-h:
I will execute the smoke, multisched and sched interface tests after reverting https://github.com/PBSPro/pbspro/commit/98235b6e3b44db35857d125d0f75f7ecb31e6223 in my local branch

https://github.com/PBSPro/pbspro/commit/98235b6e3b44db35857d125d0f75f7ecb31e6223 seems to have broken TestMultipleSchedulers

`2018-03-06 19:48:53,654 INFO     ================================================================================
error: test_backfill_per_scheduler (tests.functional.pbs_multi_sched.TestMultipleSchedulers)
error: test_equiv_multisched (tests.functional.pbs_multi_sched.TestMultipleSchedulers)
error: test_equiv_partition (tests.functional.pbs_multi_sched.TestMultipleSchedulers)
error: test_equiv_single_partition (tests.functional.pbs_multi_sched.TestMultipleSchedulers)
error: test_equiv_suspend_jobs (tests.functional.pbs_multi_sched.TestMultipleSchedulers)
error: test_job_array_partition (tests.functional.pbs_multi_sched.TestMultipleSchedulers)
error: test_job_queue_partition (tests.functional.pbs_multi_sched.TestMultipleSchedulers)
error: test_job_sorted_per_scheduler (tests.functional.pbs_multi_sched.TestMultipleSchedulers)
error: test_limits_partition (tests.functional.pbs_multi_sched.TestMultipleSchedulers)
error: test_multi_fairshare (tests.functional.pbs_multi_sched.TestMultipleSchedulers)
Test cases with failures: TestMultipleSchedulers.test_backfill_per_scheduler,TestMultipleSchedulers.test_equiv_multisched,TestMultipleSchedulers.test_equiv_partition,TestMultipleSchedulers.test_equiv_single_partition,TestMultipleSchedulers.test_equiv_suspend_jobs,TestMultipleSchedulers.test_job_array_partition,TestMultipleSchedulers.test_job_queue_partition,TestMultipleSchedulers.test_job_sorted_per_scheduler,TestMultipleSchedulers.test_limits_partition,TestMultipleSchedulers.test_multi_fairshare
Test suites with failures: TestMultipleSchedulers
run: 12, succeeded: 2, failed: 0, errors: 10, skipped: 0, timedout: 0
Tests run in 0:10:35.072401
2018-03-06 19:48:53,654 INFO     Cleaning up temporary files
2018-03-06 19:48:53,654 INFO     Cleaning up /var/tmp dir
2018-03-06 19:48:53,655 INFO     Cleaning up /tmp dir
`


The same is passing properly in its previous commit https://github.com/PBSPro/pbspro/commit/d6eb67f76f92528c29db7378424818b2cffe1a8f
`2018-03-06 19:23:59,449 INFO     ================================================================================
run: 24, succeeded: 24, failed: 0, errors: 0, skipped: 0, timedout: 0
Tests run in 0:04:58.943673
2018-03-06 19:23:59,449 INFO     Cleaning up temporary files
2018-03-06 19:23:59,450 INFO     Cleaning up /var/tmp dir
2018-03-06 19:23:59,450 INFO     Cleaning up /tmp dir`

I have attached PTL out for both commit ids here:
[Failing 98235b6 PP-747 Equivalence class tests with scheduler partitions.zip](https://github.com/PBSPro/pbspro/files/1785366/Failing.98235b6.PP-747.Equivalence.class.tests.with.scheduler.partitions.zip)
[Passing d6eb67f Adding additional tests for pbsfs.zip](https://github.com/PBSPro/pbspro/files/1785367/Passing.d6eb67f.Adding.additional.tests.for.pbsfs.zip)

subhasisb:
Codacy seems to be a bit unhappy, but can't tell me why- in the absence of that, I sign off!

mike0042:
@bhroam will be unable to review these changes. Dismissing his review.

mike0042:
Codacy is complaining because it detected a new duplicate block of code in test/tests/functional/pbs_multi_sched.py. I think we can safely ignore it.

mike0042:
This PR has the required approvals, pending review by @agrawalravi90.

Shrini-h:
Ran SmokeTest,TestSchedulerInterface,TestMultipleSchedulers tests after reverting https://github.com/PBSPro/pbspro/commit/d6eb67f76f92528c29db7378424818b2cffe1a8f

Its all green
[Final_PTL_run_logs.zip](https://github.com/PBSPro/pbspro/files/1785885/Final_PTL_run_logs.zip)

@agrawalravi90  Please let me if I need to rebase again.

@mike0042  Thank you for the sign-off

agrawalravi90:
>Ran SmokeTest,TestSchedulerInterface,TestMultipleSchedulers tests after reverting d6eb67f

I'm not sure why you had to revert anything, I took your branch as is and ran TestMultipleSchedulers, it ran 36 tests (not 24), all of them successfully. Anyways, please rebase the branch in preparation for merge. Thanks.

Shrini-h:
Rebased to latest, PR ready for merge

After using new setup I was able to run TestMultipleSchedulers successfully without reverting anything.
[Rebase_PTL_run_logs.zip](https://github.com/PBSPro/pbspro/files/1788379/Rebase_PTL_run_logs.zip)
For some reason SmokeTest.test_interactive_job is timing out

agrawalravi90:
@subhasisb , was Codacy check made mandatory? I can't merge this PR because Codacy didn't pass. The timezone difference is causing this PR to linger, could you please take over this and merge it after figuring out the Codacy check?

subhasisb:
Thanks Ravi. I will merge it now. Codacy is somehow erroring out, not really showing an error in the code, but erroring out itself - hope it's transient and will fix itself.

PR #568, opened 2018-02-16
Fixing invalid writes reported by valgrind for PP-746

suresh-thelkar:
#### Bug/feature Description
* *Valgrind is reporting some of invalid writes which were caused by PP-746*

#### Affected Platform(s)
* *All platforms*

* This issue occurs when scheduler/s state is either "down" or "idle" in the database which can happen in the following corner cases only.
* Scheduler is shutdown abruptly and server is restarted
* Server is shutdown abruptly when scheduler/s state is "idle".

In other words we can say these invalid writes occur by restarting server only when state of scheduler/s in database is "idle" or "down".

#### Solution Description
* Do not save the state of scheduler/s in to the database. This was the original intention of PP-746 but somehow we missed this during that time.  Server itself can decide the state of scheduler/s, so need of saving this state in to the database. This itself will resolve the current issue.
* In addition to the above we are now using corresponding setter functions so that it correctly frees and allocates memory.

#### Checklist:

### DIFF
--- a/src/include/server.h
+++ b/src/include/server.h
@@ -253,6 +253,9 @@ enum failover_state {
 #define SVR_CLEAN_JOBHIST_SECS	5	/* never spend more than 5 seconds in one sweep to clean hist */
 #define SVR_JOBHIST_DEFAULT	1209600	/* default time period to keep job history: 2 weeks */
 
+#define VALUE(str) #str
+#define TOSTR(str) VALUE(str)
+
 /* function prototypes */
 
 extern int			svr_recov_db(void);
@@ -263,6 +266,7 @@ extern enum failover_state	are_we_primary(void);
 extern int			have_socket_licensed_nodes(void);
 extern void			unlicense_socket_licensed_nodes(void);
 extern void			set_sched_default(pbs_sched *, int unset_flag);
+extern void			set_attr_svr(attribute *pattr, attribute_def *pdef, char *value);
 
 #ifdef	__cplusplus
 }
--- a/src/server/pbsd_init.c
+++ b/src/server/pbsd_init.c
@@ -256,16 +256,16 @@ init_server_attrs()
 {
 	resource_def    *prdef = NULL;
 	resource        *presc = NULL;
-	struct attribute attrib = {0};
-	int i = 0;
+	attribute	attrib;
+	int 		i = 0;
 
-	for (i=0; i<SRV_ATR_LAST; i++)
+
+	for (i = 0; i < SRV_ATR_LAST; i++)
 		clear_attr(&server.sv_attr[i], &svr_attr_def[i]);
 
-	server.sv_attr[(int)SRV_ATR_scheduler_iteration].at_val.at_long =
-		PBS_SCHEDULE_CYCLE;
-	server.sv_attr[(int)SRV_ATR_scheduler_iteration].at_flags =
-		ATR_VFLAG_SET|ATR_VFLAG_MODCACHE;
+	set_attr_svr(&(server.sv_attr[(int)SRV_ATR_scheduler_iteration]), &svr_attr_def[(int) SRV_ATR_scheduler_iteration],
+		TOSTR(PBS_SCHEDULE_CYCLE));
+
 	server.sv_attr[(int)SRV_ATR_State].at_val.at_long = SV_STATE_INIT;
 	server.sv_attr[(int)SRV_ATR_State].at_flags =
 		ATR_VFLAG_SET|ATR_VFLAG_MODCACHE;
@@ -313,6 +313,7 @@ init_server_attrs()
 	server.sv_attr[(int)SRV_ATR_max_concurrent_prov].at_flags =
 		ATR_VFLAG_DEFLT|ATR_VFLAG_SET|ATR_VFLAG_MODCACHE;
 
+	clear_attr(&attrib, &svr_attr_def[(int)	SVR_ATR_jobscript_max_size]);
 	svr_attr_def[(int)SVR_ATR_jobscript_max_size].at_decode(&attrib,ATTR_jobscript_max_size,NULL,DFLT_JOBSCRIPT_MAX_SIZE);
 	attr_jobscript_max_size.at_type  |= ATR_TYPE_SIZE;  /* get_bytes_from_attr() is checking for at_type */
 	set_size(&attr_jobscript_max_size,&attrib,SET);
--- a/src/server/run_sched.c
+++ b/src/server/run_sched.c
@@ -349,8 +349,10 @@ schedule_high(pbs_sched *psched)
 		return -1;
 
 	if (psched->scheduler_sock == -1) {
-		if ((s = contact_sched(psched->svr_do_sched_high, NULL, psched->pbs_scheduler_addr, psched->pbs_scheduler_port)) < 0)
+		if ((s = contact_sched(psched->svr_do_sched_high, NULL, psched->pbs_scheduler_addr, psched->pbs_scheduler_port)) < 0) {
+			set_attr_svr(&(psched->sch_attr[(int) SCHED_ATR_sched_state]), &sched_attr_def[(int) SCHED_ATR_sched_state], SC_DOWN);
 			return (-1);
+		}
 		set_sched_sock(s, psched);
 		if (psched->scheduler_sock2 == -1) {
 			if ((s = contact_sched(SCH_SCHEDULE_NULL, NULL, psched->pbs_scheduler_addr, psched->pbs_scheduler_port)) >= 0)
@@ -359,9 +361,8 @@ schedule_high(pbs_sched *psched)
 		psched->svr_do_sched_high = SCH_SCHEDULE_NULL;
 		return 0;
 	}
-	strncpy(psched->sch_attr[(int) SCHED_ATR_sched_state].at_val.at_str, SC_SCHEDULING, SC_STATUS_LEN);
-	psched->sch_attr[(int) SCHED_ATR_sched_state].at_val.at_str[SC_STATUS_LEN] = '\0';
-	psched->sch_attr[(int) SCHED_ATR_sched_state].at_flags = ATR_VFLAG_DEFLT | ATR_VFLAG_SET | ATR_VFLAG_MODCACHE;
+
+	set_attr_svr(&(psched->sch_attr[(int) SCHED_ATR_sched_state]), &sched_attr_def[(int) SCHED_ATR_sched_state], SC_SCHEDULING);
 
 	return 1;
 }
@@ -425,8 +426,10 @@ schedule_jobs(pbs_sched *psched)
 			pdefr = (struct deferred_request *)GET_NEXT(pdefr->dr_link);
 		}
 
-		if ((s = contact_sched(cmd, jid, psched->pbs_scheduler_addr, psched->pbs_scheduler_port)) < 0)
+		if ((s = contact_sched(cmd, jid, psched->pbs_scheduler_addr, psched->pbs_scheduler_port)) < 0) {
+			set_attr_svr(&(psched->sch_attr[(int) SCHED_ATR_sched_state]), &sched_attr_def[(int) SCHED_ATR_sched_state], SC_DOWN);
 			return (-1);
+		}
 		else if (pdefr != NULL)
 			pdefr->dr_sent = 1;   /* mark entry as sent to sched */
 		set_sched_sock(s, psched);
@@ -436,9 +439,7 @@ schedule_jobs(pbs_sched *psched)
 		}
 		psched->svr_do_schedule = SCH_SCHEDULE_NULL;
 
-		strncpy(psched->sch_attr[(int) SCHED_ATR_sched_state].at_val.at_str, SC_SCHEDULING, SC_STATUS_LEN);
-		psched->sch_attr[(int) SCHED_ATR_sched_state].at_val.at_str[SC_STATUS_LEN] = '\0';
-		psched->sch_attr[(int) SCHED_ATR_sched_state].at_flags = ATR_VFLAG_DEFLT | ATR_VFLAG_SET | ATR_VFLAG_MODCACHE;
+		set_attr_svr(&(psched->sch_attr[(int) SCHED_ATR_sched_state]), &sched_attr_def[(int) SCHED_ATR_sched_state], SC_SCHEDULING);
 
 		first_time = 0;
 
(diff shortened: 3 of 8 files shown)


suresh-thelkar:
No need to write separate PTL test cases for this. I have run all the multisched test cases under valgrind and got the clean report where invalid writes which caused due to PP-746 are absent now.

src/server/pbsd_init.c
```diff
@@ -256,14 +256,18 @@ init_server_attrs()
 {
 	resource_def    *prdef = NULL;
 	resource        *presc = NULL;
-	struct attribute attrib = {0};
-	int i = 0;
+	attribute 	temp_attr;
+	struct 		attribute attrib = {0};
```
mike0042: In attribute.h we have... typedef struct attribute attribute;
Normally, (for the sake of clarity) this would be... typedef struct attribute attribute_t;
You've used both "attribute" and "struct attribute" here, even though they are the same thing. Please use one or the other.

src/server/pbsd_init.c
```diff
@@ -256,14 +256,18 @@ init_server_attrs()
...
+	int 		i = 0;
+
 
 	for (i=0; i<SRV_ATR_LAST; i++)
 		clear_attr(&server.sv_attr[i], &svr_attr_def[i]);
 
-	server.sv_attr[(int)SRV_ATR_scheduler_iteration].at_val.at_long =
-		PBS_SCHEDULE_CYCLE;
+	temp_attr.at_flags = ATR_VFLAG_SET;
+	temp_attr.at_val.at_long = PBS_SCHEDULE_CYCLE;
```
arungrover: @suresh-thelkar Please use at_decode function to set the values here and in all other places because decode is a common way of populating an attribute and if that changes in future we will not have to worry about these special cases. It will also set the flag for you.
Also, if you end up using decode, don't forget to call at_free as well because that will then free up memory inside the attribute (if allocated)

src/server/svr_recov_db.c
```diff
@@ -504,8 +504,10 @@ sched_save_db(pbs_sched *ps, int mode)
 		attr_info.parent_obj_type = PARENT_TYPE_SCHED; /* svr attr */
 		attr_info.parent_id = ps->sc_name;
 
+		ps->sch_attr[SCHED_ATR_sched_state].at_flags &= ~(ATR_VFLAG_MODIFY | ATR_VFLAG_MODCACHE);
 		if (save_attr_db(conn, &attr_info, sched_attr_def, ps->sch_attr, (int)SCHED_ATR_LAST, flag) !=0)
 			goto db_err;
+		ps->sch_attr[SCHED_ATR_sched_state].at_flags &= (ATR_VFLAG_MODIFY | ATR_VFLAG_MODCACHE);
```
bhroam: Did you mean |= ?  This will get rid of all bits other than modify and modcache

src/server/run_sched.c
```diff
@@ -359,9 +365,9 @@ schedule_high(pbs_sched *psched)
 		psched->svr_do_sched_high = SCH_SCHEDULE_NULL;
 		return 0;
 	}
-	strncpy(psched->sch_attr[(int) SCHED_ATR_sched_state].at_val.at_str, SC_SCHEDULING, SC_STATUS_LEN);
-	psched->sch_attr[(int) SCHED_ATR_sched_state].at_val.at_str[SC_STATUS_LEN] = '\0';
-	psched->sch_attr[(int) SCHED_ATR_sched_state].at_flags = ATR_VFLAG_DEFLT | ATR_VFLAG_SET | ATR_VFLAG_MODCACHE;
```
arungrover: You may still want to retain this line to add "ATR_VFLAG_DEFLT"  (in this place and all other places) because no setter function will add this for you.

src/server/run_sched.c
```diff
@@ -344,13 +344,19 @@ int
...
 	if (psched == NULL)
 		return -1;
 
+	temp_attr.at_flags = ATR_VFLAG_SET;
+
 	if (psched->scheduler_sock == -1) {
-		if ((s = contact_sched(psched->svr_do_sched_high, NULL, psched->pbs_scheduler_addr, psched->pbs_scheduler_port)) < 0)
+		if ((s = contact_sched(psched->svr_do_sched_high, NULL, psched->pbs_scheduler_addr, psched->pbs_scheduler_port)) < 0) {
+			temp_attr.at_val.at_str = SC_DOWN;
+			sched_attr_def[(int) SCHED_ATR_sched_state].at_set(&psched->sch_attr[(int) SCHED_ATR_sched_state], &temp_attr, SET);
```
bhroam: set_str() will return an error if malloc() fails.  Do you want to catch that?  (for all other instances of setting the scheduler state as well).

src/server/pbsd_init.c
suresh-thelkar: @mike0042,
I agree with you that we could have actually defined it as struct attribute attribute_t. I am using attribute only in both the places.
Since I have now created a separate function to set value for any attribute through server this temp_attr is not required here.(But defined inside the function set_attr_svr)

src/server/pbsd_init.c
```diff
@@ -256,14 +256,18 @@ init_server_attrs()
...
 	resource_def    *prdef = NULL;
 	resource        *presc = NULL;
-	struct attribute attrib = {0};
-	int i = 0;
+	attribute 	temp_attr;
+	struct 		attribute attrib = {0};
+	int 		i = 0;
+
 
 	for (i=0; i<SRV_ATR_LAST; i++)
```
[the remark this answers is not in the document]
suresh-thelkar: Implemented.

src/server/pbsd_init.c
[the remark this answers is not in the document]
suresh-thelkar: Yes @bhroam, we need to do clear_attr() on temp_attr. Since I have now created a new function which is used to set a value for any attribute from server I am taking care of calling this in that function set_attr_svr().

suresh-thelkar: Thanks for letting me know this. My initial idea was we should use decode functions only when we are getting data from external source i.e. either from database or from the user as input. One more thing what made me to use only set functions is most of the places in server code we are directly setting  value for an attribute of an object. But it seems it is better to use decode() followed by set functions as suggested by you. I have actually come up with a function named set_attr_svr() which takes care of all these and I am calling this wherever required.

And to correct all the places where we are not following this convention as of now which are outside the scope of this bug planning to raise an infrastructure RFE so that all the code we follow one approach. When we implement this we can actually enhance the newly written function set_attr_svr().

src/server/run_sched.c
suresh-thelkar: I have actually written a function set_attr_svr() which takes care of decoding, setting and clearing the attribute. I am logging appropriate error messages if corresponding functions fail.

src/server/run_sched.c
suresh-thelkar: It looks we are not really making use of this attribute in Multisched code and hence not using the flag.

src/server/svr_recov_db.c
suresh-thelkar: Implemented.

suresh-thelkar:
@bhroam, @arungrover, @mike0042 
I have implemented all of your comments. Please look into the latest commit and provide your sign off or further review comments if any.

src/include/server.h
```diff
@@ -253,6 +253,9 @@ enum failover_state {
 #define SVR_CLEAN_JOBHIST_SECS	5	/* never spend more than 5 seconds in one sweep to clean hist */
 #define SVR_JOBHIST_DEFAULT	1209600	/* default time period to keep job history: 2 weeks */
 
+#define VALUE(str) #str
+#define TOSTR(str) VALUE(str)
```
mike0042: Well, isn't that clever.  :-)

src/server/svr_func.c
```diff
@@ -7193,3 +7193,38 @@ action_opt_bf_fuzzy(attribute *pattr, void *pobj, int actmode)
...
+ * @param[in]	value	-	value to be set
+ *
+ * @return	void
+ *
+ * @par MT-Safe: No
+ * @par Side Effects: None
+ *
+ */
+void
+set_attr_svr(attribute *pattr, attribute_def *pdef, char *value)
```
mike0042: We already have functions named set_attribute, mgr_set_attr, set_attr, and set_attr_resc. The name set_attr_svr is innocuous, but not very descriptive. Besides the changes proposed in this PR, aren't there dozens of places where this should be called as opposed to the "old way" of setting server attributes?

If you're going to introduce a new function, I suggest you look for all the places it could be applied and refactor accordingly.

src/include/server.h
bhroam: I think I'm missing something here.  Why do you need two macros?  Why can't you just have one macro that does #str?  According to my book, the # preprocessor will quote what follows.  I don't see what having the extra macro buys us.

src/server/pbsd_init.c
```diff
@@ -256,17 +256,15 @@ init_server_attrs()
 {
 	resource_def    *prdef = NULL;
 	resource        *presc = NULL;
-	attribute 	temp_attr;
-	struct 		attribute attrib = {0};
+	attribute	attrib = {0};
```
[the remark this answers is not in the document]
bhroam: I believe clear_attr() is the right answer.  It will memset() the structure and set some members to the right values.

src/server/svr_func.c
```diff
@@ -7193,3 +7193,38 @@ action_opt_bf_fuzzy(attribute *pattr, void *pobj, int actmode)
...
+{
+	attribute tempat;
+
+	if (pattr == NULL || pdef == NULL) {
+		(void)sprintf(log_buffer,  "Invalid pointer to attribute or its definition");
```
bhroam: use snprintf() (and other uses in this function)

src/server/svr_func.c
```diff
@@ -7193,3 +7193,38 @@ action_opt_bf_fuzzy(attribute *pattr, void *pobj, int actmode)
...
+		log_err(errno, __func__, log_buffer);
```
bhroam: errno will not be set.  You'll end up with a weird message which says "Success".  Maybe PBSE_INTERNAL?

src/include/server.h
suresh-thelkar: I would like share you how I came to write such a macro. 

Following is the simple program that I tried initially.

#include <stdio.h>
#define PBS_SCHEDULE_CYCLE    600 /* re-schedule even if no change, 10 min   */

#define VALUE(str) #str
#define TOSTR(str) VALUE(str)

unsigned char s1[60] = TOSTR (PBS_SCHEDULE_CYCLE);    /* generates "600" \*/
unsigned char s2[60] = VALUE(PBS_SCHEDULE_CYCLE);     /* generates "PBS_SCHEDULE_CYCLE" \*/
int main()
{
        printf("s1 = %s\n",s1);
        printf("s2 = %s\n",s2);
        return 0;
}
Since we need the value of PBS_SCHEDULE_CYCLE as a string, I had to use both of the above macros.

src/server/svr_func.c
suresh-thelkar: @mike0042,
I named the function as set_attr_svr because we can read it like "setting an attribute of any object from server" thinking that the older style of setting attributes is mostly used in server and hence named the function like that. Please suggest me a better name if you already have something in your mind. I am fine to change it.

As I mentioned earlier since I am going to log a separate infrastructure RFE to correct the code  where the older style is used(apparently it seems there are several places like that) to newer style we will revisit this function at that and modify it accordingly if required.

src/server/pbsd_init.c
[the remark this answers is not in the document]
suresh-thelkar: Thanks @bhroam for answering it.  As we are now calling clear_attr() as part of the new function I have changed the definition of this variable back to attribute attrib; Removed the initialization here.

src/server/svr_func.c
suresh-thelkar: Done

src/server/svr_func.c
suresh-thelkar: I am now collecting the return value in rc both from decode() and set() functions and using the same.

suresh-thelkar:
@bhroam / @mike0042,
I have implemented your review comments. Please look into the latest commit and provide your sign off or further comments if any.
@arungrover,
Thanks for your sign offf. I have made couple of changes to implement Bhroam and Mike's review comments. Please go through the latest commit and provide your sign off or further review comments if any.

src/server/svr_func.c
```diff
@@ -7212,19 +7212,20 @@ void
...
-		(void)sprintf(log_buffer,  "decode of %s failed", pdef->at_name);
-		log_err(errno, __func__, log_buffer);
-	} else if (pdef->at_set(pattr, &tempat, SET) != 0) {
-		(void)sprintf(log_buffer, "set of %s failed", pdef->at_name);
-		log_err(errno, __func__, log_buffer);
+	if ((rc = pdef->at_decode(&tempat, pdef->at_name, NULL, value)) != 0) {
+		snprintf(log_buffer,  sizeof(log_buffer), "decode of %s failed", pdef->at_name);
+		log_err(rc, __func__, log_buffer);
+	} else if ((rc = pdef->at_set(pattr, &tempat, SET)) != 0) {
+		(void)snprintf(log_buffer, sizeof(log_buffer), "set of %s failed", pdef->at_name);
```
mike0042: Please remove the cast to void for this call as well.

src/include/server.h
mike0042: FWIW, it looks correct to me.

src/server/svr_func.c
mike0042: I don't have a problem with the name you chose, it's just that we already have several functions with similar names. Every time we add another it makes it more difficult to determine which is the correct call to use in a given scenario. Please link the refactoring ticket to PP-746 so we can tell they are related.

src/server/svr_func.c
mike0042: That may be a better choice than using -1 as I saw in a couple places.

src/include/server.h
vstumpf: In my opinion, I would flip them. `TOSTR` should return the name of the `#define` as a string, and `VALUE` should return the value of `#define`.

bhroam: Thanks for the explanation.  I see it now.  I agree with @mike0042, this is quite clever.  As to @vstumpf's comment, the usable macro is TOSTR().  The other macro can be called anything.  I think TOSTR() correctly defines what is going on.

vstumpf: Ah, I see now that TOSTR is making the value of a `#define` to a string. Makes sense to me.

src/server/svr_func.c
suresh-thelkar: Done

src/server/svr_func.c
suresh-thelkar: Done

suresh-thelkar:
@bhroam / @mike0042 ,
I have implemented all of your comments. Can you please provide your sign off if you do not have any further comments.

suresh-thelkar:
@mike0042 / @arungrover,
Thank you for your approvals. 
@bhroam,
Since I do not see any further comments from you since yesterday I am assuming your sign off and as this code fix is urgent  I have squashed the commits and rebased it with master. Hope this is fine.

PR #569, opened 2018-02-17
Fix for invalid read of mem from req struct after its free'd

hirenvadalia:
#### Bug/feature Description
* *Invalid read reported by valgrind on request struct after its free'd*

#### Affected Platform(s)
* *All*

* *Use of object from request struct after its free'd*

#### Solution Description
* *Use object first then free request struct or store object value from struct in temporary variable then use temporary variable after request free'd*

#### Checklist:

### DIFF
--- a/src/server/hook_func.c
+++ b/src/server/hook_func.c
@@ -982,9 +982,9 @@ mgr_hook_delete(struct batch_request *preq)
 {
 
 
-	hook 		*phook;
-	char		*hookname;
-	char		hook_msg[HOOK_MSG_SIZE];
+	hook *phook = NULL;
+	char hookname[PBS_MAXSVRJOBID + 1] = {0};
+	char hook_msg[HOOK_MSG_SIZE] = {0};
 
 	memset(hook_msg, '\0', HOOK_MSG_SIZE);
 
@@ -993,7 +993,7 @@ mgr_hook_delete(struct batch_request *preq)
 		return;
 	}
 
-	hookname = preq->rq_ind.rq_manager.rq_objname;
+	strncpy(hookname, preq->rq_ind.rq_manager.rq_objname, PBS_MAXSVRJOBID);
 
 	phook = find_hook(hookname);
 
@@ -1127,21 +1127,21 @@ py_compile_and_run(char *input_file_path, char *hook_msg, size_t msg_len,
 void
 mgr_hook_import(struct batch_request *preq)
 {
-	svrattrl	*plist, *plx;
-	char		*hookname;
-	hook		*phook;
-	char		content_type[BUFSIZ];
-	char		content_encoding[BUFSIZ];
-	char		input_file[MAXPATHLEN+1];
-	char		input_file_path[MAXPATHLEN+1];
-	char		input_path[MAXPATHLEN+1];
-	char		temp_path[MAXPATHLEN+1];
-	char		output_path[MAXPATHLEN+1];
-	char		hook_msg[HOOK_MSG_SIZE];
-	int		overwrite;
-	struct python_script *py_test_script=NULL;
-	int		rc;
-	int		hook_obj;
+	svrattrl *plist, *plx;
+	char hookname[PBS_MAXSVRJOBID + 1] = {0};
+	hook *phook;
+	char content_type[BUFSIZ];
+	char content_encoding[BUFSIZ];
+	char input_file[MAXPATHLEN + 1];
+	char input_file_path[MAXPATHLEN + 1];
+	char input_path[MAXPATHLEN + 1] = {0};
+	char temp_path[MAXPATHLEN + 1];
+	char output_path[MAXPATHLEN + 1] = {0};
+	char hook_msg[HOOK_MSG_SIZE] = {0};
+	int overwrite;
+	struct python_script *py_test_script = NULL;
+	int rc;
+	int hook_obj;
 
 	hook_obj = preq->rq_ind.rq_manager.rq_objtype;
 
@@ -1150,19 +1150,14 @@ mgr_hook_import(struct batch_request *preq)
 		return;
 	}
 
-	memset(hook_msg, '\0', HOOK_MSG_SIZE);
-	memset(input_path, '\0', MAXPATHLEN+1);
-	memset(output_path, '\0', MAXPATHLEN+1);
-	hookname = preq->rq_ind.rq_manager.rq_objname;
+	strncpy(hookname, preq->rq_ind.rq_manager.rq_objname, PBS_MAXSVRJOBID);
 
 	phook = find_hook(hookname);
 
 	if ((phook == NULL) || phook->pending_delete) {
-		snprintf(hook_msg, HOOK_MSG_SIZE-1,
-			"%s does not exist!", hookname);
+		snprintf(hook_msg, HOOK_MSG_SIZE, "%s does not exist!", hookname);
+		log_event(PBSEVENT_ADMIN, PBS_EVENTCLASS_HOOK, LOG_INFO, hookname, hook_msg);
 		reply_text(preq, PBSE_HOOKERROR, hook_msg);
-		log_event(PBSEVENT_ADMIN, PBS_EVENTCLASS_HOOK, LOG_INFO,
-			hookname, hook_msg);
 		return;
 	}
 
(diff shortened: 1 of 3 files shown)


hirenvadalia:
[before_fix_valgrind_log.txt](https://github.com/PBSPro/pbspro/files/1733728/before_fix_valgrind_log.txt)
[after_fix_valgrind_log.txt](https://github.com/PBSPro/pbspro/files/1733727/after_fix_valgrind_log.txt)
[ptl_log.txt](https://github.com/PBSPro/pbspro/files/1733729/ptl_log.txt)

src/server/hook_func.c
```diff
@@ -982,9 +982,9 @@ mgr_hook_delete(struct batch_request *preq)
 {
 
 
-	hook 		*phook;
-	char		*hookname;
-	char		hook_msg[HOOK_MSG_SIZE];
+	hook *phook = NULL;
+	char hookname[PBS_MAXSVRJOBID + 1] = {0};
```
subhasisb: Why a size of PBS_MAXSVRJOBID? Is that not too small? can be hookname be larger than that.

src/server/hook_func.c
```diff
@@ -993,7 +993,7 @@ mgr_hook_delete(struct batch_request *preq)
 		return;
 	}
 
-	hookname = preq->rq_ind.rq_manager.rq_objname;
+	strncpy(hookname, preq->rq_ind.rq_manager.rq_objname, PBS_MAXSVRJOBID);
```
subhasisb: Perhaps doing a strdup and freeing after using is good enough?

src/server/hook_func.c
```diff
@@ -1127,21 +1127,21 @@ py_compile_and_run(char *input_file_path, char *hook_msg, size_t msg_len,
...
-	char		input_path[MAXPATHLEN+1];
-	char		temp_path[MAXPATHLEN+1];
-	char		output_path[MAXPATHLEN+1];
-	char		hook_msg[HOOK_MSG_SIZE];
-	int		overwrite;
-	struct python_script *py_test_script=NULL;
-	int		rc;
-	int		hook_obj;
+	svrattrl	*plist = NULL, *plx = NULL;
+	char		hookname[PBS_MAXSVRJOBID + 1] = {0};
```
subhasisb: Same question as above

test/tests/functional/pbs_hook_set_nonexist.py
```diff
@@ -0,0 +1,55 @@
+# coding: utf-8
```
subhasisb: Isn't there a hook test suite to which this can be added, instead of a new suite?

src/server/hook_func.c
hirenvadalia: @subhasisb I have used PBS_MAXSVRJOBID only for one reason, as source buffer is length of PBS_MAXSVRJOBID.
(see [here](https://github.com/PBSPro/pbspro/blob/master/src/include/batch_request.h#L146))

src/server/hook_func.c
hirenvadalia: I can change it to strdup.
But just for my knowledge why strdup instead strncpy? What we will gain by using strdup?

test/tests/functional/pbs_hook_set_nonexist.py
hirenvadalia: Will check and try to move to existing testsuite if any.

src/server/hook_func.c
```diff
@@ -1127,21 +1127,21 @@ py_compile_and_run(char *input_file_path, char *hook_msg, size_t msg_len,
...
+	char		input_file[MAXPATHLEN + 1] = {0};
+	char		input_file_path[MAXPATHLEN + 1] = {0};
+	char		input_path[MAXPATHLEN + 1] = {0};
+	char		temp_path[MAXPATHLEN + 1] = {0};
+	char		output_path[MAXPATHLEN + 1] = {0};
+	char		hook_msg[HOOK_MSG_SIZE] ={0};
+	int		overwrite = -1;
+	struct python_script *py_test_script = NULL;
+	int		rc = -1;
+	int		hook_obj = -1;
```
bhroam: I dislike initializing variables like this for three reasons.  First, if you don't need to do it, you shouldn't.  It takes extra time to zero buffers like this.  Second, while you won't actually read uninitialized memory, you will miss the fact that you're doing something wrong.  It will cover up a bug.  We now build with -Wall -Werror.  If you read from a variable before you initialize it, gcc will throw a warning and travis/jenkins will barf.  Lastly, I've seen codacy complain about this.  You set something just to set it again before using it. 
If you look further down, you memset 3 of your variables you are initializing here to the same thing.  Extra time zeroing buffers for no reason.

src/server/hook_func.c
```diff
@@ -1153,39 +1153,34 @@ mgr_hook_import(struct batch_request *preq)
...
-	hookname = preq->rq_ind.rq_manager.rq_objname;
+
+	strncpy(hookname, preq->rq_ind.rq_manager.rq_objname, PBS_MAXSVRJOBID);
 
 	phook = find_hook(hookname);
 
 	if ((phook == (hook *)0) || phook->pending_delete) {
-		snprintf(hook_msg, HOOK_MSG_SIZE-1,
-			"%s does not exist!", hookname);
+		snprintf(hook_msg, HOOK_MSG_SIZE-1, "%s does not exist!", hookname);
```
bhroam: No need for the -1.  snprintf() will do that for you(and with the other snprintf() calls).  Also, can you remove the '!'.  I dislike them in log messages.

src/server/hook_func.c
```diff
@@ -1515,40 +1510,34 @@ mgr_hook_export(struct batch_request *preq)
...
-	hookname = preq->rq_ind.rq_manager.rq_objname;
+	strncpy(hookname, preq->rq_ind.rq_manager.rq_objname, PBS_MAXSVRJOBID);
 
 	/* Else one and only one vhook */
 	phook = find_hook(hookname);
 
 	if ((phook == (hook *)0) || phook->pending_delete) {
-		snprintf(hook_msg, HOOK_MSG_SIZE-1,
-			"%s does not exist!", hookname);
+		snprintf(hook_msg, HOOK_MSG_SIZE-1, "%s does not exist!", hookname);
```
bhroam: Same here about the '!' (and for others)

src/server/hook_func.c
```diff
@@ -1764,44 +1753,32 @@ mgr_hook_set(struct batch_request *preq)
...
-
 		return;
 	}
 
 	if ((phook->type == HOOK_SITE) && (hook_obj == MGR_OBJ_PBS_HOOK)) {
-		sprintf(log_buffer,
-			"cannot set attributes of a '%s' hook", HOOKSTR_SITE);
+		sprintf(log_buffer, "cannot set attributes of a '%s' hook named %s", HOOKSTR_SITE, phook->hook_name);
+		log_event(PBSEVENT_ADMIN, PBS_EVENTCLASS_HOOK, LOG_INFO, hookname, log_buffer);
+		sprintf(log_buffer, "cannot set attributes of a '%s' hook", HOOKSTR_SITE);
```
bhroam: use snprintf() for the above 2 sprintf() calls

test/tests/functional/pbs_hook_set_nonexist.py
```diff
@@ -0,0 +1,55 @@
...
+from tests.functional import *
+
+
+class TestHookSetNonExist(TestFunctional):
+
+    def test_set_on_nonexist_hook(self):
+        try:
+            self.server.manager(MGR_CMD_DELETE, MGR_OBJ_HOOK, id="test")
+        except PbsManagerError:
+            pass
```
bhroam: You don't need to do this.  revert_to_defaults() will delete all hooks.

src/server/hook_func.c
bhroam: I think it will cause confusion.  It looks like a bug that you are sizing your buffer for a hook name by the max size for a job id.

hirenvadalia: Hey @bhroam, Sorry but I didn't get your point. Are you asking to change buffer size in source buffer? or Are you asking to make new macro with same size as PBS_MAXSVRJOBID and use it?

src/server/hook_func.c
hirenvadalia: I think you are right, will remove initialization in next commit.

src/server/hook_func.c
hirenvadalia: Regarding '!':
Removal of '!' may turn in to design change as I have seen tests written around them... as this message is used in reply_text which comes to client (aka qmgr command)...
Are you still want me to change?

src/server/hook_func.c
subhasisb: No additional gain. The are essentially the same. You declare a local variable and have to carefully determine the size and do a proper strncpy and ensure it is null terminated. Besides making sure that the size variable makes sense and does not confuse people (due to it's name  referring to jobid size). A strdup is easy, faster than strncpy in many cases and does the same work, and is less code. Just a suggestion, choose what you like best.

src/server/hook_func.c
bhroam: I did a little research and the internal hook code doesn't limit the length of a hook name.  The only limitation is from the batch protocol itself.  When the hook name is set in the hook structure, it uses strdup().  You shouldn't use a fixed length buffer here because if we ever increase the size of the name field in the rq_manage structure, you'll be limiting the size of a hook name.

src/server/hook_func.c
bhroam: Dang.  If it requires changes to tests then don't bother.  It isn't worth it.  Thanks for looking into it.

hirenvadalia:
Thanks @bhroam and @subhasisb for review... I have rebased and squashed branch with latest master... its ready for merge...

PR #192, opened 2016-11-11
pbs_ical correct icalerror_errors_are_fatal

honza801:
#### Issue-ID

* **[PP-XXXX](https://pbspro.atlassian.net/browse/PP-XXXX)**

#### Problem description
* cannot compile due to non-compatible libical api calls

#### Cause / Analysis
* one should not use icalerror_errors_are_fatal global variable, since it breaks compilation

#### Solution description
* use libical setters instead of using global (undefined) variables

#### Checklist:

- [x] I have joined the **[pbspro community forum](http://community.pbspro.org/)**.
- [x] My pull request contains a **single, signed** commit. See **[setting up gpg signature](https://pbspro.atlassian.net/wiki/display/DG/Signing+Your+Git+Commits)**.
- [x] My code follows the **[coding style](https://pbspro.atlassian.net/wiki/display/DG/Coding+Standards)** of this project.
- [x] My change requires project documentation. See **[required documentation checklist](https://pbspro.atlassian.net/wiki/display/DG/Checklist+for+Developing+Features+and+Bug+Fixes)** for details.
- [x] I have added documentation in the **[project documentation area](https://pbspro.atlassian.net/wiki/display/PD)**.
- [x] I have added new **PTL test(s) to my commit**. (See **[using PTL for testing](https://pbspro.atlassian.net/wiki/display/DG/Using+PTL+for+Testing)**) *(or)*
- [x] I have added  **manual test(s) to the Jira ticket and explained why PTL is not appropriate** for this case.
- [x] All new and existing automated tests have passed. (See **[running automated PTL tests](https://pbspro.atlassian.net/wiki/display/DG/PTL+Quick+Start+Guide)**).
- [x] I have attached **test logs to the Jira ticket** as evidence of testing/verification.

__***For further information please visit the [Developer Guide Home](https://pbspro.atlassian.net/wiki/display/DG/Developer+Guide+Home).***__

corresponding to libical api, setting ical errors are fatal should be handled by getters and setters. global variable is not available.

PR #570, opened 2018-02-21
Scheduler errors out if server_dyn_res script returns unexpected value

agrawalravi90:
#### Bug/feature Description
* Scheduler fails to run any jobs if there's a server_dyn_res script which returns bad output
* sched logs show the error "Problem with creating server data structure".
* This can be reproduced by creating a server_dyn_res script which returns a string value for a long type resource.

#### Affected Platform(s)
* All

* query_server_dyn_res() bails if set_resource() fails. Post the check-in of #192, set_resource() was enhanced to do error checking for the resources being set. So, for example, if a long type resource is being set to a string value, the function will catch it and error out. So, it errors out if the server_dyn_res script returns bad output for a resource, and query_server_dyn_res() just bails, causing any job in the system, regardless of whether they requested this resource or not, to fail.

#### Solution Description
* query_server_dyn_res() should handle the case of bad server_dyn_res script output in the same way that it handles not being able to open the script or reading its contents, which is that it sets the value of the resource to 0 and moves on. This will mean that any job that requests it will fail, which is fine as the server_dyn_res script is messed up, but the jobs which don't deal with it at all work just fine.

#### Testing:
- [Before fix](https://github.com/PBSPro/pbspro/files/1742632/failed.log)
- [After fix](https://github.com/PBSPro/pbspro/files/1742633/passed.log)

#### Checklist:

### DIFF
--- a/src/scheduler/server_info.c
+++ b/src/scheduler/server_info.c
@@ -709,7 +709,11 @@ query_server_dyn_res(server_info *sinfo)
 				}
 
 				if (set_resource(res, buf, RF_AVAIL) == 0) {
-					return -1;
+					snprintf(buf, sizeof(buf), "Script %s returned bad output",
+							conf.dynamic_res[i].program);
+					schdlog(PBSEVENT_DEBUG, PBS_EVENTCLASS_SERVER, LOG_DEBUG,
+										"server_dyn_res", buf);
+					(void) set_resource(res, res_zero, RF_AVAIL);
 				}
 			}
 			else {
--- a/test/tests/functional/pbs_svr_dyn_res.py
+++ b/test/tests/functional/pbs_svr_dyn_res.py
@@ -1,0 +1,90 @@
+# coding: utf-8
+
+# Copyright (C) 1994-2018 Altair Engineering, Inc.
+# For more information, contact Altair at www.altair.com.
+#
+# This file is part of the PBS Professional ("PBS Pro") software.
+#
+# Open Source License Information:
+#
+# PBS Pro is free software. You can redistribute it and/or modify it under the
+# terms of the GNU Affero General Public License as published by the Free
+# Software Foundation, either version 3 of the License, or (at your option) any
+# later version.
+#
+# PBS Pro is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.
+# See the GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+# Commercial License Information:
+#
+# For a copy of the commercial license terms and conditions,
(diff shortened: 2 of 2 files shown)


agrawalravi90:
Thanks for the sign-off Bhroam. Requesting one more review on this, thanks!

src/scheduler/server_info.c
```diff
@@ -709,7 +709,11 @@ query_server_dyn_res(server_info *sinfo)
 				}
 
 				if (set_resource(res, buf, RF_AVAIL) == 0) {
-					return -1;
+					snprintf(buf, sizeof(buf), "Script %s returned bad output",
```
subhasisb: Do we need the capital S for script? Guess it is not a norm we follow.

src/scheduler/server_info.c
```diff
@@ -709,7 +709,11 @@ query_server_dyn_res(server_info *sinfo)
...
+							conf.dynamic_res[i].program);
+					schdlog(PBSEVENT_DEBUG, PBS_EVENTCLASS_SERVER, LOG_DEBUG,
+										"server_dyn_res", buf);
+					(void) set_resource(res, res_zero, RF_AVAIL);
```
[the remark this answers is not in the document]
agrawalravi90: I just followed the style that was used in the other set_resource() call in the function, I'll remove it if you really hate it.

src/scheduler/server_info.c
```diff
@@ -709,7 +709,11 @@ query_server_dyn_res(server_info *sinfo)
 				}
 
 				if (set_resource(res, buf, RF_AVAIL) == 0) {
-					return -1;
+					snprintf(buf, sizeof(buf), "Script %s returned bad output",
+							conf.dynamic_res[i].program);
+					schdlog(PBSEVENT_DEBUG, PBS_EVENTCLASS_SERVER, LOG_DEBUG,
```
subhasisb: Do you want debug or error?

src/scheduler/server_info.c
agrawalravi90: It's a caps for start of the log message sentence, I just followed the style of other log messages in the function:
snprintf(buf, sizeof(buf), "**C**an't pipe to program %s: %s",
		conf.dynamic_res[i].program, strerror(pipe_err));
else
snprintf(buf, sizeof(buf), "**E**rror piping to program %s.",
		conf.dynamic_res[i].program);

src/scheduler/server_info.c
agrawalravi90: We don't error out when this happens, the scheduler behaves similarly to how it does when other similarly bad things happen in the function like bad pipe for the server_dyn_res script, it moves on without erroring out. So, I think DEBUG is consistent.

src/scheduler/server_info.c
subhasisb: Ok thanks for pointing out. I guess we do other things in other places..so it is fine.

src/scheduler/server_info.c
subhasisb: okay

agrawalravi90:
Thanks for the merge Subhasis!

PR #571, opened 2018-02-21
PP-1024: Secondary server stuck in "Starting" state even after Primary server is killed, with an error in tpp_em_wait()

sid2364:
#### Bug/feature Description
* **
* Bugs: failover does not function as expected, server logs have an error saying Invalid argument (22) in wait_request, tpp_em_wait() error, errno=22
* [PP-1024](https://pbspro.atlassian.net/browse/PP-1024)

#### Affected Platform(s)
* Windows only

* On Windows, select() on an empty (non-null) fd set returns a value of -1. Since wait_request() continously looks at this value and then sets the secondary state to SECONDARY_STATE_noconn, after it has been modified to be SECONDARY_STATE_takeov. This causes an infinite loop of first calling wait_request, checking "one last time" that the server is not connected, then seeing that primary is no longer connected, then trying to set state to takeover from as secondary and then again calling wait_request to start over again.
* Bugs: select() returns -1 for an empty (non-null) fd set on Windows. (On Linux, it returns a 0 and failover can done as expecting).

#### Solution Description
* Solution was to either add a dummy fd in the read set, or skip the call to select() entirely. Skipping the call to select() would require looping through the maxfd value and checking if something is set. This would slow everything down slightly since wait_request is called quite frequently. So, gone ahead with adding a dummy fd in the read set, so the call to select() does not fail.

#### Checklist:

### DIFF
--- a/src/lib/Libnet/net_server.c
+++ b/src/lib/Libnet/net_server.c
@@ -996,6 +996,7 @@ static int
 init_poll_context(void)
 {
 #ifdef WIN32
+	int sd_dummy;
 	max_connection = FD_SETSIZE;
 #else
 	int idx;
@@ -1018,6 +1019,25 @@ init_poll_context(void)
 		log_err(errno, __func__, "could not initialize poll_context");
 		return (-1);
 	}
+#ifdef WIN32
+	/* set a dummy fd in the read set so that	*/
+	/* select() does not return WSAEINVAL 		*/
+	sd_dummy = socket(AF_INET, SOCK_STREAM, 0);
+	if (sd_dummy < 0) {
+		errno = WSAGetLastError();
+		log_err(errno, __func__, "socket() failed");
+		return -1;
+	}
+	if ((tpp_em_add_fd(poll_context, sd_dummy, EM_IN) == -1)) {
+		int err = errno;
+		snprintf(logbuf, sizeof(logbuf), 
+			"Could not add socket %d to the read set", sd_dummy);
+		log_err(err, __func__, logbuf);
+		CLOSESOCKET(sd_dummy);
+		return -1;
+	}
+#endif /* WIN32 */
+
 	return 0;
 }
 


sid2364:
Server log capturing secondary takeover: [secondary_takes_over_successfully__failover_windows.txt](https://github.com/PBSPro/pbspro/files/1743602/secondary_takes_over_successfully__failover_windows.txt)

src/lib/Libnet/net_server.c
```diff
@@ -332,6 +332,23 @@ init_network_add(int sd, void (*readfunc)(int))
...
+			return (-1);
+		}
+		if((tpp_em_add_fd(poll_context, sd_dummy, EM_IN) == -1)){
+			int err = errno;
+			snprintf(logbuf, sizeof(logbuf), 
+				"Could not add socket %d to the read set", sd_dummy);
+			log_err(err, __func__, logbuf);
+			close_conn(sd);
+		}
+#endif /* WIN32 */
```
subhasisb: This is fine actually, but it would be cleaner to add this code block to init_poll_context (under WIN32 of course), so that in case init_poll_context gets call elsewhere too the right thing happens. Thanks!

src/lib/Libtpp/tpp_em.c
```diff
@@ -1086,22 +1077,7 @@ tpp_em_wait_win(void *em_ctx, em_event_t **ev_array, int timeout)
...
-			flag_fd_isset = 1;
-		if(FD_ISSET(i, &ctx->read_fds))
-			DBPRT(("## tpp_em_wait_win: fd=%d is set for read\n", i));
-		if(FD_ISSET(i, &ctx->write_fds))
-			DBPRT(("## tpp_em_wait_win: fd=%d is set for write\n", i));
-		if(FD_ISSET(i, &ctx->err_fds))
-			DBPRT(("## tpp_em_wait_win: fd=%d is set for err\n", i));
-	}
-	if(!flag_fd_isset)
-		return 0;
```
subhasisb: You should really squash and rebase for before initiating your review, since these changes are not existent on the codeline.

sid2364: Will keep that in mind.

src/lib/Libnet/net_server.c
sid2364: Done.

src/lib/Libnet/net_server.c
```diff
@@ -1035,6 +1019,25 @@ init_poll_context(void)
...
+#ifdef WIN32
+	/* set a dummy fd in the read set so that	*/
+	/* select() does not return WSAEINVAL 		*/
+	sd_dummy = socket(AF_INET, SOCK_STREAM, 0);
+	if (sd_dummy < 0) {
+		errno = WSAGetLastError();
+		log_err(errno, __func__, "socket() failed");
+		return -1;
+	}
+	if((tpp_em_add_fd(poll_context, sd_dummy, EM_IN) == -1)){
```
subhasisb: Minor nit - please add a space between ) and {

src/lib/Libnet/net_server.c
```diff
@@ -1035,6 +1019,25 @@ init_poll_context(void)
...
+		int err = errno;
+		snprintf(logbuf, sizeof(logbuf), 
+			"Could not add socket %d to the read set", sd_dummy);
+		log_err(err, __func__, logbuf);
+		close_conn(sd_dummy);
```
subhasisb: I think we cannot call close_conn() here since we had not yet done a add_conn(), so better to simply close the socket and return. So replace with CLOSESOCKET(sd_dummy);

src/lib/Libnet/net_server.c
sid2364: Fixed

src/lib/Libnet/net_server.c
sid2364: I've changed it.

subhasisb:
Please attach the manual test results (before and after) the fix.

sid2364:
@subhasisb, here are the before and after results:-
**Before fix** server logs excerpt: [before_fix_invalid_argument_error.txt](https://github.com/PBSPro/pbspro/files/1747405/before_fix_invalid_argument_error.txt)
**After fix** server log's excerpt: [after_fix__secondary_takes_over_successfully.txt](https://github.com/PBSPro/pbspro/files/1747408/after_fix__secondary_takes_over_successfully.txt)

PR #572, opened 2018-02-22
Fix for equiv classes not working with limited resources not in resou…

bhroam:
…rces line

#### Bug/feature Description
* Job's requesting limit resources could cause jobs not requesting those resources to not run if the resources were not on the sched_config resources line.
#### Affected Platform(s)
* All
* The scheduler only creates equivalence classes for jobs requesting resources on the sched_config resources line(and a few others).  If any limit was placed on a resource that was not on the resources line, an equivalence class would not be created for it.  This means a job requesting a limit resource could cause a job not requesting that limit resource to not run.
#### Solution Description
* The solution is to add the limit resources to the array of resources used to create equivalence classes.  The limits code already keeps track of all resources that are being limited.  This is stored in a static global to the limits code.  The limits code is treated more like a library to the scheduler.  I chose to leave this variable static to the limits code and write a getter function for it.  I use this function when creating the array of resources which are used to create equivalence classes.
#### Checklist:

### DIFF
--- a/src/include/limits_if.h
+++ b/src/include/limits_if.h
@@ -147,8 +147,7 @@ extern int	is_oldlimattr(const struct attrl *);
  * @retval NULL		: attribute value is not an old limit attribute
  *
  */
-char *
-convert_oldlim_to_new(const struct attrl *a);
+extern char *	convert_oldlim_to_new(const struct attrl *a);
 
 /**	@fn int lim_setlimits(const struct attrl *a, enum limtype lt, void *p)
  *	@brief set resource or run-time limits
@@ -215,7 +214,17 @@ extern int	check_soft_limits(server_info *, queue_info *, resource_resv *);
  *             resource definitions are updated.
  *      @return void
  */
-extern void clear_limres(void);
+extern void 	clear_limres(void);
+
+/**
+ * 	@fn schd_resource query_limres()
+ * 	@brief returns a linked list of resources being limited.
+ * 	@par This is to be treated as read-only.  Modifying this will adversely
+ * 	     affect the limits code
+ *
+ * 	@return schd_resource *
+ */
+extern schd_resource *query_limres(void);
 
 
 #ifdef	__cplusplus
--- a/src/scheduler/job_info.c
+++ b/src/scheduler/job_info.c
@@ -1483,7 +1483,7 @@ update_job_attr(int pbs_sd, resource_resv *resresv, char *attr_name,
 			}
 			pattr->next = pattr2;
 			/* extra may have been a list, let's find the end */
-			for(end = pattr2; end->next != NULL; end = end ->next)
+			for (end = pattr2; end->next != NULL; end = end ->next)
 				;
 		}
 	}
@@ -2005,7 +2005,7 @@ free_resresv_set_array(resresv_set **rsets) {
 	if (rsets == NULL)
 		return;
 
-	for(i = 0; rsets[i] != NULL; i++)
+	for (i = 0; rsets[i] != NULL; i++)
 		free_resresv_set(rsets[i]);
 
 	free(rsets);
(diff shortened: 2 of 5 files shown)


src/include/limits_if.h
```diff
@@ -217,6 +217,16 @@ extern int	check_soft_limits(server_info *, queue_info *, resource_resv *);
...
 
+/**
+ * 	@fn schd_resource query_limres()
+ * 	@brief returns a linked list of resources being limited.
+ * 	@par This is to be treated as read-only.  Modifying this will adversely
+ * 	     affect the limits code
+ *
+ * 	@return schd_resource *
+ */
+schd_resource *query_limres();
```
mike0042: Don't you want to prefix this with "extern" and add "void" between the parens?

src/scheduler/limits.c
```diff
@@ -2913,6 +2913,18 @@ clear_limres(void)
 	free_resource_list(limres);
 	limres = NULL;
 }
+/**
```
mike0042: There's normally an empty line before a doxygen header.

src/scheduler/limits.c
```diff
@@ -2913,6 +2913,18 @@ clear_limres(void)
...
+ *      @brief returns a linked list of resources being limited.
+ *
+ *      @par This is to be treated as read-only.  Modifying this will adversely
+ *           affect the limits code
+ *
+ *      @return schd_resource *
+ */
+schd_resource *query_limres()
```
mike0042: The type and function name should be on separate lines. If it takes no arguments, please add "void" in parens.

src/scheduler/job_info.c
```diff
@@ -2255,6 +2263,14 @@ create_resresv_sets_resdef(status *policy, server_info *sinfo) {
 		defs[i++] = getallres(RES_PREEMPT_TARGETS);
 	defs[i] = NULL;
 
+	for(cur_res = limres; cur_res != NULL; cur_res = cur_res->next) {
+		/* There may be overlap between resdef_to_check and limres */
+		if (!resdef_exists_in_array(defs, cur_res->def)) {
+			defs[i++] = cur_res->def;
+			defs[i] = NULL;
```
arungrover: you can probably remove this null assignment out of the for loop and even remove it from line 2264 to do it just once

src/scheduler/limits.c
arungrover: I'm not sure why this function is needed because it just returns a global pointer, but if we do, then should we use it in limits.c as well when we iterate through all the limit resources.

bhroam: This is a weirdity that I tried to cover in my PR description.  The limits code is treated more as a "library" to the scheduler even though it's in the scheduler directory.  It's header file lives in include instead of scheduler.  The rest of the scheduler only calls functions which are provided by the limits interface.  This is why I chose to go this way instead of just removing the static from limres.  The limits code itself is inside the "library" so it can directly use a private member.  I'd really rather not have to query this member at all.  It's needed by the limits code and if it is changed, the limits code will not function properly.

src/scheduler/job_info.c
bhroam: It's needed for resdef_exists_in_array().  It requires a NULL terminated array.  Every time I add something to the array, I need to terminate it again.

src/scheduler/limits.c
arungrover: Ok I get it now. Thanks!

mike0042:
Looks good @bhroam

src/scheduler/job_info.c
```diff
@@ -2255,6 +2263,14 @@ create_resresv_sets_resdef(status *policy, server_info *sinfo) {
 		defs[i++] = getallres(RES_PREEMPT_TARGETS);
 	defs[i] = NULL;
 
+	for(cur_res = limres; cur_res != NULL; cur_res = cur_res->next) {
```
minghui-liu: A minor style thing: your two for loops are missing spaces after the `for` keyword

bhroam:
Thanks for pointing that out @minghui-liu.  I always forget that space.  I also caught two other instances of missing spaces in the file that I fixed.

test/tests/functional/pbs_equiv_classes.py
```diff
@@ -1931,3 +1931,31 @@ def test_overlap_resv(self):
...
+        a = {ATTR_RESC_TYPE: 'long'}
+        self.server.manager(MGR_CMD_CREATE, RSC, a, id='foores')
+
+        a = {'max_run_res.foores': '[u:PBS_GENERIC=4]'}
+        self.server.manager(MGR_CMD_SET, SERVER, a)
+
+        a = {'Resource_List.foores': 1, 'Resource_List.select': '1:ncpus=1'}
+        self.submit_jobs(2, a)
+        a['Resource_List.foores'] = 2
+        (_, jid4) = self.submit_jobs(2, a)
```
minghui-liu: Parentheses are not necessary here.

bhroam: Oh, wow, you're right.  I thought because submit_jobs() returns a list that I needed the ().  In any case, the entire test plan uses the ().  I'll either leave them alone, or pull them from the entire plan, and I'd rather not pull them from the entire plan.

minghui-liu: That's fine. It just surprised me a little bit because parentheses makes it looks like `submit_jobs` always returns a tuple instead of a list. I had to go to the definition to check.

PR #573, opened 2018-02-22
Remove cached ip addresses when node is deleted

subhasisb:
#### Bug/feature Description
When a node is created or loaded from database during startup of PBS server, the IP address(es) of the node is cached in a avl tree. This ip caching is done to enhance performance so that jobs being sent to a mom does not need repeated name resolution. While this works great, there is a problem. If a node with name node1 is added to server, and later removed, and recreated again, and the node comes up with a different IP address then pbs server continues to use the old IP address cached inside itself.

#### Affected Platform(s)
* * All

* * PBS server caches IP addresses of moms even and does not invalidate/clear those entries even when the mom/host is removed from pbs server.

#### Solution Description
* * When the node is removed, delete the associated IP addresses from the internal cache. This way, if a node with the same name is added again, then it is resolved again by pbs server and the right ip address is added to the cache.

#### Checklist:

### DIFF
--- a/src/server/req_manager.c
+++ b/src/server/req_manager.c
@@ -2878,6 +2878,38 @@ make_host_addresses_list(char *phost, u_long **pul)
 
 /**
  * @brief
+ * 		remove the cached ip addresses of a mom from the host tree and the ipaddrs tree
+ *
+ * @param[in]	pmom - valid ptr to the mom info
+ *
+ * @return	error code
+ * @retval	0		- no error
+ * @retval	PBS error	- error
+ */
+int
+remove_mom_ipaddresses_list(mominfo_t *pmom)
+{
+	/* take ipaddrs from ipaddrs cache tree */
+	if (hostaddr_tree != NULL) {
+		struct pul_store *tpul;
+
+		if ((tpul = (struct pul_store *) find_tree(hostaddr_tree, pmom->mi_host)) != NULL) {
+			u_long *pul;
+			for (pul = tpul->pul; *pul; pul++)
+				tdelete2(*pul, pmom->mi_port, &ipaddrs);
+
+			if (tree_add_del(hostaddr_tree, pmom->mi_host, NULL, TREE_OP_DEL) != 0)
+				return (PBSE_SYSTEM);
+
+			free(tpul->pul);
+			free(tpul);
+		}
+	}
+	return 0;
+}
+
+/**
+ * @brief
  *		create pbs node structure, i.e. add a node
  *
  * @param[in]	objname	- Name of node
--- a/src/server/mom_info.c
+++ b/src/server/mom_info.c
@@ -95,6 +95,8 @@ int	    svr_num_moms = 0;
 extern char	*msg_daemonname;
 extern char	*path_hooks_rescdef;
 
+extern int remove_mom_ipaddresses_list(mominfo_t *pmom);
+
 /*
  * The following function are used by both the Server and Mom
  *	create_mom_entry()
@@ -469,6 +471,12 @@ delete_svrmom_entry(mominfo_t *pmom)
 		/* take stream out of tree */
 		(void)rpp_close(psvrmom->msr_stream);
 		tdelete2((unsigned long)psvrmom->msr_stream , 0, &streams);
+
+		if (remove_mom_ipaddresses_list(pmom) != 0) {
+			snprintf(log_buffer, sizeof(log_buffer), "Could not remove IP address for mom %s:%d from cache",
+					pmom->mi_host, pmom->mi_port);
+			log_err(errno, __func__, log_buffer);
+		}
 	}
 	memset((void *)psvrmom, 0, sizeof(mom_svrinfo_t));
 	psvrmom->msr_stream = -1; /* always set to -1 when deleted */


subhasisb:
The reproduction of this scenario needs to add a node, then remove it, change the IP address of the node, and then again it again and try to run a job on that node. All these are not supported by PTL framework yet, so I am performing these steps manually. Attaching the test results below.

subhasisb:
Test steps as root (except in steps specified otherwise):
========================================================
1. cat /etc/hosts, assuming hostname is centosvm1 here is contents:
	192.168.203.131 centosvm1 
2. Install PBS with a single mom in a single machine setup. 
3. Create a local node: qmgr -c "c n `hostname`"
4. Run pbs_mom
5. pbsnodes -av should show the node as up.
6. Run a job: qsub -- /bin/date and verify that it ran. (as non-root user)
7. Delete the node: qmgr -c "d n `hostname`"
8. Stop mom: pkill pbs_mom
9. Add another ip address to your network interface, as follows:
	ip addr add 192.168.203.151 dev ens33
10. Update /etc/hosts to point the hostname to this new ip address, and edit the existing ip address to a new name like centosvmsvr. After this update, /etc/hosts shows:
	192.168.203.131 centosvmsvr
	192.168.203.151 centosvm1  --> Note IP address of same node name is now different
11. Add centosvmsvr to the clienthost of mom. Edit /var/spool/pbs/mom_priv/config to add:
	$clienthost centosvmsvr
12. Create a local node: qmgr -c "c n `hostname`"
13. Run pbs_mom
14. pbsnodes -av should show the node as up.
15. Run a job: qsub -- /bin/date. (as non-root user)
16. qstat -f will show that the job is being tried continuously, but cannot run. 
17. Check server logs to find error message as follows:
	02/22/2018 18:31:05;0008;Server@centosvm1;Job;2725.centosvm1;Job Run at request of Scheduler@centosvmsvr on exec_vnode (centosvm1:ncpus=1)
02/22/2018 18:31:05;0001;Server@centosvm1;Req;;Could not connect to Mom
02/22/2018 18:31:05;0008;Server@centosvm1;Job;2725.centosvm1;send of job to centosvm1 failed error = 115
02/22/2018 18:31:05;0008;Server@centosvm1;Job;2725.centosvm1;Unable to Run Job, send to Mom failed

With fixed version, the job will run successfully.

subhasisb:
Test results attached
==============
[reproduction_log.txt](https://github.com/PBSPro/pbspro/files/1747984/reproduction_log.txt)
[fixed_log.txt](https://github.com/PBSPro/pbspro/files/1747985/fixed_log.txt)

subhasisb:
valgrind logs
=========
[val_log.txt](https://github.com/PBSPro/pbspro/files/1748056/val_log.txt)

src/server/req_manager.c
```diff
@@ -2876,6 +2878,41 @@ make_host_addresses_list(char *phost, u_long **pul)
...
+{
+	/* take ipaddrs from ipaddrs cache tree */
+	if (hostaddr_tree != NULL) {
+		struct pul_store *tpul;
+		u_long *pul;
+
+		if ((tpul = (struct pul_store *) find_tree(hostaddr_tree, pmom->mi_host)) != NULL) {
+			pul = tpul->pul;
+			while (*pul) {
+					tdelete2(*pul, pmom->mi_port, &ipaddrs);
```
mike0042: Need to check pmom before we dereference?

src/server/req_manager.c
```diff
@@ -167,6 +167,8 @@ extern char *msg_cannot_set_route_que;
 
 extern int check_req_aoe_available(struct pbsnode *, char *);
 int resize_prov_table(int);
+int remove_mom_ipaddresses_list(mominfo_t *pmom);
```
bhroam: No need for this.  You don't actually call the function in this file.

src/server/req_manager.c
```diff
@@ -2876,6 +2878,41 @@ make_host_addresses_list(char *phost, u_long **pul)
...
+remove_mom_ipaddresses_list(mominfo_t *pmom)
+{
+	/* take ipaddrs from ipaddrs cache tree */
+	if (hostaddr_tree != NULL) {
+		struct pul_store *tpul;
+		u_long *pul;
+
+		if ((tpul = (struct pul_store *) find_tree(hostaddr_tree, pmom->mi_host)) != NULL) {
+			pul = tpul->pul;
+			while (*pul) {
```
bhroam: You're iterating.  Use a for() loop.

bhroam:
Does this PR require a PTL test?

src/server/req_manager.c
```diff
@@ -2876,6 +2878,41 @@ make_host_addresses_list(char *phost, u_long **pul)
...
+					tdelete2(*pul, pmom->mi_port, &ipaddrs);
+					pul++;
+			}
+
+			if (tree_add_del(hostaddr_tree, pmom->mi_host, NULL, TREE_OP_DEL) != 0)
+				return (PBSE_SYSTEM);
```
minghui-liu: Parentheses are not necessary here.

subhasisb:
@bhroam We can do PTL, but the test will have to be complicated and deal with adding ip addresses to an existing network interface, update /etc/hosts or other DNS infra in place. Such things are currently not supported in PTL and would have to be custom, platform specific, root-required code. I was hoping not to go that route. What do you feel?

src/server/req_manager.c
subhasisb: That's true. I will remove it.

src/server/req_manager.c
subhasisb: okay - i know you love for() way more... :)

src/server/req_manager.c
subhasisb: I feel we don't need to, since this is way deep in a stack of calls which all expect pmom to be valid. I can add a note to the description saying pmom is expected to be a valid pointer

src/server/req_manager.c
subhasisb: removed

src/server/req_manager.c
minghui-liu: I disagree and think a NULL check should be put. Because the function might be called in other places where the call stack is not deep and `pmom` was not check. Even if there aren't such cases now in our code base there might be in the future.

subhasisb:
hoping to have addressed the codacy issue as well.

src/server/req_manager.c
mike0042: Agreed. Just double checking if this is one of the "sane" places where we can assume the value will never be NULL. Our increased use of tools like valgrind will help to identify any memory references that are out of bounds or dereferencing NULL pointers.

src/server/req_manager.c
mike0042: I've thought about the use of parens with return after I saw many of them removed in a recent PR. I tend to favor their use in case you ever want to define a macro for return(). If you don't have the parens, then a macro won't work. As of now, there is no rule governing whether to use them or not other than following the local style as @subhasisb has done here.

src/server/req_manager.c
mike0042: And I know you're teasing a bit, but I've promoted the same thing several times in reviews. All it takes is someone to throw a "continue" in a while loop, and poof... infinite loop. The for loop protects against that.

src/server/req_manager.c
minghui-liu: I agree that parens are useful for macros, but it makes more sense to me to put parens when defining them rather than when using.

minghui-liu: For this PR I am not going to insist on removing the parens so that it complies with the existing style in the file. I can open anther PR in the future and move the discussion there. Thanks @subhasisb and @mike0042.

bhroam:
@subhasisb I agree.  I do not like the idea of changing system configuration within a PTL test.  If a PTL test errors in the middle, you won't set everything back.  Thanks for explaining.

src/server/req_manager.c
mike0042: Looking at the current version of src/server/req_manager.c I don't see any consistency in the use of parens with return statements. Until there is a policy that mandates their use (or non-use), I don't see anything wrong with the change as it stands.

iestockdale:
Referring to Subhasis's answer regarding why he didn't write a PTL test, and Bhroam's follow-up comment on the perils of changing system configuration inside a PTL test, could a PTL RFE be filed to request a safe mechanism for supporting such tests?  That way we'lll get that into the PTL backlog.

src/server/req_manager.c
bhroam: I agree with @minghui-liu about the use of macros.  One of the perils of using macros is not using enough parenthese.  A macro of SQ(x) x * x is bad.  If you were to call SQ(x+1), you'll end up with x+1 * x+1 and end up with 2x+2.  SQ(x) (x) * (x) or even better ((x) * (x)).  I also agree with @mike0042 that until we have a standard for whether or not parenthese are needed for return statements, anything really goes.  Consistency with the file is the best we can hope for.

src/server/req_manager.c
subhasisb: of course and i like for too :)

subhasisb:
Do i have a second sign off? Thanks

mike0042:
Looks good. Please squash and rebase.

bhroam:
I merged in the PR.

PR #574, opened 2018-02-23
PP-905: Ctrl+C during Interactive job startup causes race condition on Cray

npadole20:
#### Bug Description

* Ctrl+C during Interactive job startup causes race condition on Cray. 
  This leaves behind orphaned ALPS reservation in the Cray system and jobs are held because of lack of resources.

* Issue ID: [PP-905](https://pbspro.atlassian.net/browse/PP-905)

#### Affected Platform(s)
* Cray

* A user immediately does a Ctrl-C after submitting an interactive qsub that requests Cray compute nodes. But by this time PBS mom has forked, and one process is trying to create the ALPS reservation while another part is being killed off due to the Ctrl-C. The ALPS reservation is made successfully, but the ALPS reservation ID is not available when PBS is deleting the job so the ALPS reservation is not canceled. This causes ALPS reservations to be left behind. And since PBS thinks those resources are free, but ALPS disagrees, we end up with held jobs due to the mismatch.
The hook that keeps ALPS and PBS in sync will not be able to help with undoing the problem, because it's main purpose is to notice nodes coming up or going down. It wouldn't know what to do with nodes that are busy with an ALPS reservation. And because it runs only periodically, it will not prevent jobs from being held in between the times it runs.

#### Solution Description
* Report the ALPS reservation ID to parent mom as soon as it is available.Thus when the job is being deleted, parent mom will also cancel the ALPS reservation associated with the job as it will have the value of ALPS reservation ID available. 

#### Checklist:

### DIFF
--- a/src/include/job.h
+++ b/src/include/job.h
@@ -1058,14 +1058,21 @@ task_find	(job		*pjob,
 #define JOB_EXEC_HOOK_DELETE   -19 /* a hook requested for job to be deleted */
 #define JOB_EXEC_RERUN_MS_FAIL -20 /* Mother superior connection failed */
 #define JOB_EXEC_FAIL_SECURITY -21 /* Security breach in PBS directory */
-#define JOB_EXEC_HOOKERROR	-22 /* job exec failed due to */
-				    /* unexpected exception or */
-				    /* hook execution timed out */
+#define JOB_EXEC_HOOKERROR	-22 /* job exec failed due to
+				     * unexpected exception or
+				     * hook execution timed out
+				     */
+#define JOB_EXEC_UPDATE_ALPS_RESV_ID 1 /* Update ALPS reservation ID to parent mom as soon
+					* as it is available.
+					* This is neither a success nor a failure exit code,
+					* so we are using a positive value
+					*/
 
-
-/* Fake "random" number added onto the end of the staging */
-/* and execution directory when sandbox=private	          */
-/* used in jobdirname()				          */
+/*
+ * Fake "random" number added onto the end of the staging
+ * and execution directory when sandbox=private
+ * used in jobdirname()
+ */
 #define FAKE_RANDOM	"x8z"
 
 /* The default project assigned to jobs when project attribute is unset */
--- a/src/resmom/start_exec.c
+++ b/src/resmom/start_exec.c
@@ -1628,6 +1628,16 @@ record_finish_exec(int sd)
 		exec_bail(pjob, JOB_EXEC_RETRY, log_buffer);
 		return;
 	}
+
+#if MOM_ALPS
+	if (sjr.sj_code == JOB_EXEC_UPDATE_ALPS_RESV_ID) {
+		pjob->ji_extended.ji_ext.ji_pagg = sjr.sj_pagg;
+		pjob->ji_extended.ji_ext.ji_reservation = sjr.sj_reservation;
+		(void)writepipe(pjob->ji_mjspipe, &sjr, sizeof(sjr));
+		return;
+	}
+#endif
+
 	/* send back as an acknowledgement that MOM got it */
 	(void)writepipe(pjob->ji_mjspipe, &sjr, sizeof(sjr));
 	(void)close_conn(pjob->ji_jsmpipe);
@@ -2125,6 +2135,9 @@ finish_exec(job *pjob)
 	int			downfds2 = -1;	/* init to invalid fd */
 	int			port_out, port_err;
 	struct startjob_rtn	sjr;
+#if MOM_ALPS
+	struct startjob_rtn     ack;
+#endif
 	char			*termtype;
 	pbs_task			*ptask;
 	struct	array_strings	*vstrs;
@@ -2863,6 +2876,13 @@ finish_exec(job *pjob)
 				LOG_NOTICE, pjob->ji_qs.ji_jobid, log_buffer);
 			starter_return(upfds, downfds, JOB_EXEC_FAIL1, &sjr);
 		}
+#if MOM_ALPS
+		sjr.sj_code = JOB_EXEC_UPDATE_ALPS_RESV_ID;
+		(void)writepipe(upfds, &sjr, sizeof(sjr));
+
+		/* wait for acknowledgement */
+		(void)readpipe(downfds, &ack, sizeof(ack));
+#endif
 
 		/* Open the slave pty as the controlling tty */
 


npadole20:
[fix_testing_23_feb.txt](https://github.com/PBSPro/pbspro/files/1750642/fix_testing_23_feb.txt)
Attaching testing log.

src/resmom/start_exec.c
```diff
@@ -2125,6 +2135,9 @@ finish_exec(job *pjob)
 	int			downfds2 = -1;	/* init to invalid fd */
 	int			port_out, port_err;
 	struct startjob_rtn	sjr;
+#if MOM_ALPS
+	struct startjob_rtn     ack;
```
vstumpf: What is ack used for, other than reading from the pipe? We assign it a value, but never read it.

src/include/job.h
```diff
@@ -1060,7 +1060,10 @@ task_find	(job		*pjob,
 #define JOB_EXEC_HOOKERROR	-22 /* job exec failed due to */
 				    /* unexpected exception or */
 				    /* hook execution timed out */
-
+#if MOM_ALPS
+#define JOB_EXEC_UPDATE_ALPS_RESV_ID 23  /*update ALPS reservation ID to parent mom
```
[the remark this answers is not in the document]
npadole20: @vstumpf the exit values are negative for job exec failure scenarios. This is neither a pass nor a failure job exec code. We just want to update reservation ID. In fact, I was not sure what the value should be. So I would be happy to go with the value you would suggest.

src/resmom/start_exec.c
npadole20: I wanted to accomplish what the starter_return() function does but without closing the fds. So I followed the same style as that of starter_return()

npadole20:
@arungrover thank you for the review. I will try that out and update you with the results in a bit.

npadole20:
@arungrover I tried the scenario you asked to test. In start_exec.c, by putting a breakpoint at 3527, if I try to delete the job, the qdel command gets hung. So I exited gdb to let the qdel command execute and then checked the tracejob of the submitted job. I am attaching the log. Please let me know if this is what you were expecting. 
[del_job_before_sid.txt](https://github.com/PBSPro/pbspro/files/1758593/del_job_before_sid.txt)

src/resmom/start_exec.c
vstumpf: Looking at it now, I see that it blocks until the ack comes in. There's nothing wrong with the code.

arungrover:
@npadole20 I'm not sure why qdel would hang in this scenario. Try qdel -Wforce instead to delete the job. I was hoping that you put breakpoint in the child mom but before it writes sjr back to parent mom (start_exec.c line 2880) and when debugger hits the break point Issue qdel to delete the job and then continue the debugger. I wanted to see what will be the behavior if job is deleted in parent mom and child tries to update sjr to it. If Alps reservation does not clear up in such situation then we might have to change the code to clear up the reservation in the child itself.
what do you think?

npadole20:
@arungrover I tested without "detach-on-fork off" and I am attaching the log. The reservation is being cancelled. The delete request comes to server at "03/02/2018 04:36:38  S    delete job request received" and at log from mom is "03/02/2018 04:36:46  M    signal job request received
03/02/2018 04:36:46  M    signal job with TermJob".  
[del_job_before_sid_2.txt](https://github.com/PBSPro/pbspro/files/1775121/del_job_before_sid_2.txt)

arungrover:
Thanks for running the tests @npadole20

npadole20:
@arungrover thank you for your help and for the review. 
@vstumpf thank you for the review and sign-off.

npadole20:
I have squashed the commits and rebased the branch. It's ready for merge. Thank you.

src/include/job.h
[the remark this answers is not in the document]
mike0042: We really need a definitive answer as to why the other values are negative. If we are going to start using positive numbers, why start at 23?

npadole20:
@mike0042 for all error scenario the values are negative. JOB_EXEC_OK is the only success case for which the value is 0. JOB_EXEC_UPDATE_ALPS_RESV_ID is neither a success nor a failure code. That's why I did not use a negative value for it.

mike0042:
@npadole20 : If that is the case, then you should add a comment for your explanation. Next time someone goes through this code, they'll have one less mystery to solve. However, I still believe we should start with one rather than 23. Since you're just defining a macro, you don't need to do so conditionally. You may remove the "#if MOM_ALPS"

src/include/job.h
```diff
@@ -1061,7 +1061,9 @@ task_find	(job		*pjob,
 #define JOB_EXEC_HOOKERROR	-22 /* job exec failed due to */
 				    /* unexpected exception or */
 				    /* hook execution timed out */
-
+#define JOB_EXEC_UPDATE_ALPS_RESV_ID 1  /*update ALPS reservation ID to parent mom as soon as it is available.
+					  This is neither a success nor a failure exit code,
+					  so we are using a positive value*/
```
mike0042: Please use the appropriate multi-line comment format (with spaces).
```
/*
 * This is a multi-
 * line comment.
 */
```
You are welcome (and encouraged) to fix additional comments that do not conform.

npadole20:
@mike0042 done. Please have a look.

src/include/job.h
npadole20: Done. Please have a look.

src/include/job.h
```diff
@@ -1058,14 +1058,20 @@ task_find	(job		*pjob,
 #define JOB_EXEC_HOOK_DELETE   -19 /* a hook requested for job to be deleted */
 #define JOB_EXEC_RERUN_MS_FAIL -20 /* Mother superior connection failed */
 #define JOB_EXEC_FAIL_SECURITY -21 /* Security breach in PBS directory */
-#define JOB_EXEC_HOOKERROR	-22 /* job exec failed due to */
-				    /* unexpected exception or */
-				    /* hook execution timed out */
+#define JOB_EXEC_HOOKERROR	-22 /* job exec failed due to
+				     * unexpected exception or
+				     * hook execution timed out
+				     */
```
mike0042: This multi-line comment is correct. Thank you for fixing it.

src/include/job.h
```diff
@@ -1058,14 +1058,20 @@ task_find	(job		*pjob,
...
+#define JOB_EXEC_UPDATE_ALPS_RESV_ID 1 /* Update ALPS reservation ID to parent mom as soon as it is available.
+					* This is neither a success nor a failure exit code,
+					* so we are using a positive value
+					*/
```
mike0042: The first line is too long in this comment, and the rest of the lines need to be indented one space more.

mike0042:
Please rebase on your next commit.

src/include/job.h
npadole20: The indentation looks off in the diff. But it appears correct if you view the file. I don't know why it looks this way in this diff snippet. 
Also, fixed the comment. Squashed commits and rebased.

PR #582, opened 2018-02-28
pbsnodes -r does not clear mom state, offline_vnodes offline all vnodes

arungrover:
#### Bug/feature Description
There are two problems we are dealing with here - 
   - when a mom hook fail action is set to offline_vnodes it offlines all the vnodes associating with that mom. It does not check whether or not the same vnodes are being reported by other moms as well and there would be a way to use these vnodes using other moms. 
   - When a mom has been offlined and admin issues 'pbsnodes -r' command for that mom host, pbs server resets the state of the node back to free but it does not clear the mom state internally which means that any job that ends up on that mom will not run.

To reproduce the problems mentioned above perform following steps
 - Have two moms having the same set of vnode definitions
 - Create a mom execjob_begin hook that sleeps for a long period of time
 - Set the fail action for this hook as 'offline_vnodes'
 - Set hook alarm smaller than the sleep time in the hook
 - Submit a job that runs on one of the moms
 - On a buggy build you will see the mom and all the vnodes being turned offline when the execjob_begin hook is alarmed out

Now to reproduce pbsnode -r problem do all the steps mentioned above and 
 - issue pbsnodes -r for the mom host that has been marked as offline
 - Submit another job which will run on the vnode that was previously marked as offline
 - On a buggy build you will notice that job stays in queued state even when pbsnodes is reporting it as free

#### Affected Platform(s)
* All platforms

#### Analysis / Design
*  [External design for offline_vnode problem](https://pbspro.atlassian.net/wiki/spaces/PD/pages/195002450/offline+vnodes+should+only+offline+vnodes+belonging+to+more+than+one+Mom+when+all+the+MoMs+are+offline)
* [pbspro community forum discussion](http://community.pbspro.org/t/offline-vnodes-should-only-offline-vnodes-belonging-to-more-than-one-mom-when-all-the-moms-are-offline/864)

#### Solution Description
* Fix includes check to make sure that when vnodes are offlined it checks that all the moms are also marked offline. Fix also includes a change where while setting node state it also sets its mom state internally.

#### Checklist:

   - [x] I have added documentation in the **[project documentation area](https://pbspro.atlassian.net/wiki/spaces/PD/pages/195002450/Draft+offline+vnodes+should+only+offline+vnodes+belonging+to+more+than+one+Mom+when+all+the+MoMs+are+offline)**)

[ptl_dev_test.txt](https://github.com/PBSPro/pbspro/files/1768592/ptl_dev_test.txt)

### DIFF
--- a/src/server/node_manager.c
+++ b/src/server/node_manager.c
@@ -466,8 +466,9 @@ get_addr_of_nodebyname(char *name, unsigned int *port)
 
 
 enum Set_All_State_When {
-	Set_ALL_State_All_Down,	 /* set set on vnodes when all Moms are down */
-	Set_All_State_Regardless /* set set on vnodes regardless */
+	Set_ALL_State_All_Down,	  /* set on vnodes when all Moms are down */
+	Set_All_State_Regardless, /* set on vnodes regardless */
+	Set_All_State_All_Offline /* set on vnodes when all Moms are offline */
 };
 
 /**
@@ -497,11 +498,19 @@ set_all_state(mominfo_t *pmom, int do_set, unsigned long bits, char *txt,
 	struct pbsnode *pvnd;
 	attribute	*pat;
 	int		nchild;
+	unsigned long	inuse_flag = 0;
 
 	if (do_set) { /* STALE is not meaning in the state of the Mom, don't set it */
 		psvrmom->msr_state |= (bits & ~INUSE_STALE);
 	} else {
 		psvrmom->msr_state &= ~bits;
+	}
+
+	/* Set the inuse_flag based off the value of setwhen */
+	if (setwhen == Set_ALL_State_All_Down) {
+		inuse_flag = INUSE_DOWN;
+	} else if (setwhen == Set_All_State_All_Offline) {
+		inuse_flag = INUSE_OFFLINE_BY_MOM;
 	}
 
 	for (nchild = 0; nchild < psvrmom->msr_numvnds; ++nchild) {
@@ -512,14 +521,16 @@ set_all_state(mominfo_t *pmom, int do_set, unsigned long bits, char *txt,
 
 		/*
 		 * If this vnode has more than one Mom and
-		 * setwhen is Set_ALL_State_All_Down, then we only change
-		 * state if all Moms are down
+		 * setwhen is Set_ALL_State_All_Down or
+		 * setwhen is Set_All_State_All_Offline, then we only change
+		 * state if all Moms are down/offline
 		 */
-		if ((pvnd->nd_nummoms > 1) &&
-			(setwhen == Set_ALL_State_All_Down)) {
+		if ((pvnd->nd_nummoms > 1) && 
+			((setwhen == Set_ALL_State_All_Down) ||
+			(setwhen == Set_All_State_All_Offline))) {
 			for (imom = 0; imom < pvnd->nd_nummoms; ++imom) {
 				mstate = ((mom_svrinfo_t *)(pvnd->nd_moms[imom]->mi_data))->msr_state;
-				if ((mstate & INUSE_DOWN) == 0) {
+				if ((mstate & inuse_flag) == 0) {
 					do_this_vnode = 0;
 					break;
 				}
@@ -8084,7 +8095,7 @@ momptr_offline_by_mom(mominfo_t *pmom, char *why)
 		log_event(PBSEVENT_SYSTEM, PBS_EVENTCLASS_NODE,
 			LOG_ALERT, pmom->mi_host, why);
 
-	set_all_state(pmom, 1, INUSE_OFFLINE_BY_MOM, why, Set_All_State_Regardless);
+	set_all_state(pmom, 1, INUSE_OFFLINE_BY_MOM, why, Set_All_State_All_Offline);
 	return;
 }
 
--- a/test/tests/functional/pbs_offline_vnodes.py
+++ b/test/tests/functional/pbs_offline_vnodes.py
@@ -1,0 +1,260 @@
+# coding: utf-8
+
+# Copyright (C) 1994-2018 Altair Engineering, Inc.
+# For more information, contact Altair at www.altair.com.
+#
+# This file is part of the PBS Professional ("PBS Pro") software.
+#
+# Open Source License Information:
+#
+# PBS Pro is free software. You can redistribute it and/or modify it under the
+# terms of the GNU Affero General Public License as published by the Free
+# Software Foundation, either version 3 of the License, or (at your option) any
+# later version.
+#
+# PBS Pro is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.
+# See the GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+# Commercial License Information:
+#
+# For a copy of the commercial license terms and conditions,
+# go to: (http://www.pbspro.com/UserArea/agreement.html)
+# or contact the Altair Legal Department.
+#
+# Altair’s dual-license business model allows companies, individuals, and
+# organizations to create proprietary derivative works of PBS Pro and
+# distribute them - whether embedded or bundled with other software -
+# under a commercial license agreement.
+#
+# Use of Altair’s trademarks, including but not limited to "PBS™",
+# "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
+# trademark licensing policies.
+
+from tests.functional import *
+
+
+class TestOfflineVnodeOnHookFailure(TestFunctional):
+    """
+    Tests if vnodes are marked offline if a hook fails
+    """
+    is_cray = True
+
+    def setUp(self):
+        if not self.du.get_platform().startswith('cray'):
+            self.is_cray = False
+
+        TestFunctional.setUp(self)
+        self.server.manager(MGR_CMD_SET, SERVER, {'log_events': 2047})
+
+    def create_mom_hook(self):
+        name = "h1"
+        body = ("import pbs\n"
+                "import time\n"
+                "time.sleep(60)\n"
+                "pbs.event.accept()")
+        attr = {'event': 'execjob_begin', 'fail_action': 'offline_vnodes',
+                'alarm': '3', 'enabled': 'true'}
+        self.server.create_import_hook(name, attr, body)
+
+    def create_multi_vnodes(self, num_moms, num_vnode=3):
+        if num_moms != len(self.moms.values()):
+            self.server.manager(MGR_CMD_DELETE, NODE, id="@default",
+                                expect=True)
+        if self.is_cray is True:
+            if num_moms == 1 and len(self.moms.values()) != 1:
+                self.server.manager(MGR_CMD_CREATE, NODE,
+                                    id=self.moms.values()[0].shortname)
+                # adding a sleep of two seconds because it takes some time
+                # before node resources start showing up
+                time.sleep(2)
+            return
+        # No need to create vnodes on a cpuset mom
+        if self.moms.values()[0].is_cpuset_mom() is True:
+            return
+        vn_attrs = {ATTR_rescavail + '.ncpus': 1,
+                    ATTR_rescavail + '.mem': '1024mb'}
+        for i in range(num_moms):
+            self.server.create_vnodes('vnode', vn_attrs, num_vnode,
+                                      self.moms.values()[i],
+                                      usenatvnode=True, delall=False,
+                                      expect=False)
+            # Calling an explicit expect on newly created nodes.
+            self.server.expect(NODE, {ATTR_NODE_state: 'free'},
+                               id=self.moms.values()[i].shortname)
+
+    def test_single_mom_hook_failure_affects_vnode(self):
+        """
+        Run an execjob_begin hook that sleep for sometime,
+        at the same time set an alarm value so less that
+        the hook alarms out and server executes the fail_action
+        After this check if vnodes are marked offline.
+        In case of a single mom reporting vnodes, it should mark
+        all the vnodes and mom as offline.
+        Once offlined, reset the mom by issueing pbsnodes -r
+        and check if the job runs on one of the vnodes.
+        """
+        single_mom = ""
+        single_mom = self.moms.values()[0]
+        cpuset_mom = single_mom.is_cpuset_mom()
+        self.create_multi_vnodes(1)
+        self.create_mom_hook()
+
+        self.server.expect(NODE, {ATTR_NODE_state: 'free'},
+                           id=single_mom.shortname, max_attempts=3,
+                           interval=2)
+        j1 = Job(TEST_USER)
+        j1.set_sleep_time(1000)
+        jid = self.server.submit(j1)
+        # mom hook will alarm out and job will get into Q state
+        self.server.expect(JOB, {ATTR_state: 'Q'}, id=jid)
+
+        # since mom hook alarm out it's fail_action will put the
+        # node in offline state
+        self.server.expect(
+            NODE, {ATTR_NODE_state: 'offline'},
+            id=single_mom.shortname, max_attempts=20, interval=2)
+
+        vname = None
+        if self.is_cray is True:
+            vlist = []
+            vnl = self.server.filter(
+                VNODE, {'resources_available.vntype': 'cray_compute'})
+            vlist = vnl["resources_available.vntype=cray_compute"]
+            # Loop through each compute vnode in the list and check if state
+            # is offline
+            for v1 in vlist:
+                if vname is None:
+                    vname = v1
+                # Check that the vnode is in offline state
+                self.server.expect(
+                    VNODE, {'state': 'offline'}, id=v1, max_attempts=3,
+                    interval=2)
+        elif cpuset_mom is True:
+            vnl = self.server.status(NODE)
+            for v1 in vnl:
+                if single_mom.shortname != v1['resources_available.vnode']:
+                    if vname is None:
+                        vname = v1['resources_available.vnode']
+                    # Check that the vnode is in offline state
+                    self.server.expect(
+                        VNODE, {'state': 'offline'},
+                        id=v1['resources_available.vnode'], max_attempts=3,
+                        interval=2)
+        else:
+            vname = "vnode[0]"
+            self.server.expect(
+                NODE, {ATTR_NODE_state: 'offline'}, id='vnode[0]',
+                max_attempts=3, interval=2)
+            self.server.expect(
+                NODE, {ATTR_NODE_state: 'offline'}, id='vnode[1]',
+                max_attempts=3, interval=2)
+
+        mom_host = single_mom.shortname
+        pbs_exec = self.server.pbs_conf['PBS_EXEC']
+        pbsnodes_cmd = os.path.join(pbs_exec, 'bin', 'pbsnodes')
+        pbsnodes_reset = pbsnodes_cmd + ' -r ' + mom_host
+
+        # Set mom sync hook timeout to be a low value because if mom fails to
+        # get the hook after disabling it then next sync will happen after
+        # 2 minutes by default and we don't want to wait that long.
+        self.server.manager(MGR_CMD_SET, SERVER,
+                            {'sync_mom_hookfiles_timeout': '5'})
+        self.server.manager(MGR_CMD_SET, HOOK, {'enabled': 'False'}, id="h1")
+        # Make sure that hook has been sent to mom
+        self.server.log_match("successfully sent hook file")
+        self.du.run_cmd(self.server.hostname, cmd=pbsnodes_reset, sudo=True)
+        self.server.delete(jid, wait=True)
+
+        j2 = Job(TEST_USER)
+        j2.set_attributes({ATTR_l + '.select': '1:vnode=' + vname})
+        jid2 = self.server.submit(j2)
+        self.server.expect(NODE, {ATTR_NODE_state: 'free'},
+                           id=single_mom.shortname, max_attempts=3,
+                           interval=2)
+        self.server.expect(JOB, {ATTR_state: 'R'}, id=jid2)
+
+    def test_multi_mom_hook_failure_affects_vnode(self):
+        """
+        Run an execjob_begin hook that sleeps for sometime,
+        at the same time set an alarm value so less that
+        the hook alarms out and server executes the fail_action
+        After this check if vnodes are marked offline.
+        In case of a multiple mom reporting same set of vnodes,
+        it should not mark the vnodes and mom as offline because
+        there are other moms active and reporting same vnodes.
+        NOTE: This test needs moms to report the same set of vnodes
+        """
+        if len(self.moms.values()) != 2:
+            self.skipTest("Provide 2 moms while invoking test")
+
+        if self.moms.values()[0].is_cpuset_mom() is True or self.moms.values()[
+                1].is_cpuset_mom() is True:
+            self.skipTest("Skipping test on cpuset moms")
+        # The moms provided to the test may have unwanted vnodedef files.
+        if self.moms.values()[0].has_vnode_defs():
+            self.moms.values()[0].delete_vnode_defs()
+        if self.moms.values()[1].has_vnode_defs():
+            self.moms.values()[1].delete_vnode_defs()
+
+        self.create_multi_vnodes(2)
+        self.create_mom_hook()
+
+        # set one natural node to have higher ncpus than the other one so
+        # that the job only goes to this natural node.
+        self.server.manager(MGR_CMD_SET, NODE, {
+                            ATTR_rescavail + '.ncpus': '256'},
+                            id=self.moms.values()[0].shortname)
+        self.server.manager(MGR_CMD_SET, NODE, {
+                            ATTR_rescavail + '.ncpus': '1'},
+                            id=self.moms.values()[1].shortname)
+
+        j1 = Job(TEST_USER)
+
+        if self.is_cray is True:
+            # on a cray, make sure job runs on login node with higher number of
+            # ncpus
+            j1.set_attributes(
+                {ATTR_l + '.select': '1:ncpus=256:vntype=cray_login'})
+        else:
+            j1.set_attributes({ATTR_l + '.select': '1:ncpus=256'})
+
+        jid = self.server.submit(j1)
+        self.server.expect(JOB, {ATTR_state: 'Q'}, id=jid)
+
+        self.server.expect(NODE, {ATTR_NODE_state: 'offline'},
+                           id=self.moms.values()[0].shortname, max_attempts=3,
+                           interval=2)
+        self.server.expect(NODE, {ATTR_NODE_state: 'free'},
+                           id=self.moms.values()[1].shortname, max_attempts=3,
+                           interval=2)
+
+        if self.is_cray is True:
+            vlist = []
+            vnl = self.server.filter(
+                VNODE, {'resources_available.vntype': 'cray_compute'})
+            vlist = vnl["resources_available.vntype=cray_compute"]
+            # Loop through each compute vnode in the list and check if state =
+            # free
+            for v1 in vlist:
+                # Check that the node is in free state
+                self.server.expect(
+                    VNODE, {'state': 'free'}, id=v1, max_attempts=3,
+                    interval=2)
+
+        else:
+            self.server.expect(NODE, {ATTR_NODE_state: 'free'}, id='vnode[0]',
+                               max_attempts=3, interval=2)
+            self.server.expect(NODE, {ATTR_NODE_state: 'free'}, id='vnode[1]',
+                               max_attempts=3, interval=2)
+
+        # Delete and create offline node because PTL's setup does not delete
+        # offline nodes
+        self.server.manager(MGR_CMD_DELETE, NODE,
+                            id=self.moms.values()[0].shortname, expect=True)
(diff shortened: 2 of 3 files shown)


test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -0,0 +1,211 @@
...
+            vlist = []
+            first_vnode = 1
+            vnl = self.server.filter(
+                VNODE, {'resources_available.vntype': 'cray_compute'})
+            vlist = vnl["resources_available.vntype=cray_compute"]
+            # Loop through each compute vnode in the list and check if state =
+            # free
+            for v1 in vlist:
+                if first_vnode == 1:
+                    first_vnode += 1
```
bhroam: Since we're only setting vname once, why don't you break?

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -0,0 +1,211 @@
...
+        j2 = Job(TEST_USER)
+        j2.set_attributes({ATTR_l + '.select': '1:vnode=' + vname})
+        jid2 = self.server.submit(j2)
+        self.du.run_cmd(self.server.hostname, cmd=pbsnodes_reset, sudo=True)
+        self.server.expect(NODE, {ATTR_NODE_state: 'free'},
+                           id=self.moms.values()[0].shortname)
+        self.server.manager(MGR_CMD_SET, HOOK, {'enabled': 'False'}, id="h1")
+        self.server.manager(MGR_CMD_SET, SCHED, {'scheduling': 'True'})
+        self.server.expect(JOB, {ATTR_state: 'R'}, id=jid2)
+        self.server.delete(jid2, wait=True)
```
bhroam: No need to delete the job, it'll happen in revert_to_defaults()

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -0,0 +1,211 @@
...
+
+        # The moms provided to the test may have unwanted vnodedef files.
+        if self.moms.values()[0].has_vnode_defs():
+            self.moms.values()[0].delete_vnode_defs()
+        if self.moms.values()[1].has_vnode_defs():
+            self.moms.values()[1].delete_vnode_defs()
+        TestFunctional.setUp(self)
+
+    def tearDown(self):
+        self.server.manager(MGR_CMD_DELETE, NODE, id='@default')
```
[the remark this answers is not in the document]
arungrover: I've removed it. Thanks for catching it

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -0,0 +1,211 @@
...
+        In case of a single mom reporting vnodes, it should mark
+        all the vnodes and mom as offline.
+        """
+        self.create_multi_vnodes(1, True)
+        self.create_mom_hook()
+        try:
+            self.server.manager(MGR_CMD_DELETE, NODE,
+                                id=self.moms.values()[1].shortname)
+        except PbsManagerError as e:
+            self.assertTrue("Unknown node" in e.msg[0])
```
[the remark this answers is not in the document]
arungrover: I've removed this as well

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -0,0 +1,211 @@
...
+
+        j1 = Job(TEST_USER)
+        self.server.expect(NODE, {ATTR_NODE_state: 'free'},
+                           id=self.moms.values()[0].shortname)
+        j1.set_sleep_time(1000)
+        jid = self.server.submit(j1)
+        self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': '0'})
```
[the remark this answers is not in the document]
arungrover: Done

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -0,0 +1,211 @@
...
+
+        else:
+            self.server.expect(
+                NODE, {ATTR_NODE_state: 'offline'}, id='vnode[0]')
+            self.server.expect(
+                NODE, {ATTR_NODE_state: 'offline'}, id='vnode[1]')
+
+    def test_multi_mom_hook_failure_affects_vnode(self):
+        """
+        Run an execjob_begin hook that sleep for sometime,
```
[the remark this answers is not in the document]
arungrover: Done

test/tests/functional/pbs_offline_vnodes.py
arungrover: Well after that I need to check the state of all vnodes reported so I can not break out of the loop

test/tests/functional/pbs_offline_vnodes.py
arungrover: done

arungrover:
[dev_ptl_test2.txt](https://github.com/PBSPro/pbspro/files/1769523/dev_ptl_test2.txt)

src/server/node_manager.c
```diff
@@ -467,8 +467,9 @@ get_addr_of_nodebyname(char *name, unsigned int *port)
 
 
 enum Set_All_State_When {
-	Set_ALL_State_All_Down,	 /* set set on vnodes when all Moms are down */
-	Set_All_State_Regardless /* set set on vnodes regardless */
+	Set_ALL_State_All_Down,	  /* set set on vnodes when all Moms are down */
+	Set_All_State_Regardless, /* set set on vnodes regardless */
```
vccardenas: set appears twice in the previous two lines

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -0,0 +1,204 @@
...
+    def test_single_mom_hook_failure_affects_vnode(self):
+        """
+        Run an execjob_begin hook that sleep for sometime,
+        at the same time set an alarm value so less that
+        the hook alarms out and server executes the fail_action
+        After this check if vnodes are marked offline.
+        In case of a single mom reporting vnodes, it should mark
+        all the vnodes and mom as offline.
+        """
+        self.create_multi_vnodes(1, True)
```
vccardenas: In the case of Cray, you start out with two login nodes. You might want to disable one of them so that your test will truly have one mom.

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -0,0 +1,204 @@
...
+        self.create_mom_hook()
+
+        j1 = Job(TEST_USER)
+        self.server.expect(NODE, {ATTR_NODE_state: 'free'},
+                           id=self.moms.values()[0].shortname)
+        j1.set_sleep_time(1000)
+        jid = self.server.submit(j1)
+        self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'False'})
```
vccardenas: If you only have one login/mom then you might not need to turn off scheduling here.

test/tests/functional/pbs_offline_vnodes.py
vccardenas: This single mom test might be good to run on the Altix/UV. But the setup requires two moms.

src/server/node_manager.c
arungrover: I'll fix that

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -0,0 +1,204 @@
...
+            j1.set_attributes({ATTR_l + '.select': '1:ncpus=256'})
+
+        jid = self.server.submit(j1)
+        self.server.expect(JOB, {ATTR_state: 'Q'}, id=jid)
+
+        self.server.expect(
+            NODE, {ATTR_NODE_state: 'offline'},
+            id=self.moms.values()[0].shortname)
+        self.server.expect(NODE, {ATTR_NODE_state: 'free'},
+                           id=self.moms.values()[1].shortname)
```
lisa-altair: Sometimes when you check for the node/vnode state you set max_attempts and an interval, other times you don't.  Is there a reason?  I think you should check a few times in all places, just in case.
Here it might be because you don't expect the state of mom1 to change...but just in case I think the test should wait a little to see if it does change.

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -0,0 +1,204 @@
...
+                    first_vnode += 1
+                    vname = v1
+                # Check that the node is in free state
+                self.server.expect(
+                    VNODE, {'state': 'free'}, id=v1, max_attempts=3,
+                    interval=2)
+
+        else:
+            self.server.expect(NODE, {ATTR_NODE_state: 'free'}, id='vnode[0]')
+            self.server.expect(NODE, {ATTR_NODE_state: 'free'}, id='vnode[1]')
```
lisa-altair: Add max_attempts and an interval here too.

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -0,0 +1,204 @@
...
+
+        self.server.delete(jid, wait=True)
+        self.server.manager(MGR_CMD_SET, HOOK, {'enabled': 'False'}, id="h1")
+
+        j2 = Job(TEST_USER)
+        j2.set_attributes({ATTR_l + '.select': '1:vnode=' + vname})
+        jid2 = self.server.submit(j2)
+        self.du.run_cmd(self.server.hostname, cmd=pbsnodes_reset, sudo=True)
+        self.server.expect(NODE, {ATTR_NODE_state: 'free'},
+                           id=self.moms.values()[0].shortname)
```
lisa-altair: Add max_attempts and an interval here too.

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -0,0 +1,204 @@
...
+                            id=self.moms.values()[0].shortname)
+        self.server.manager(MGR_CMD_SET, NODE, {
+                            ATTR_rescavail + '.ncpus': '1'},
+                            id=self.moms.values()[1].shortname)
+
+        j1 = Job(TEST_USER)
+        self.server.expect(NODE, {ATTR_NODE_state: 'free'},
+                           id=self.moms.values()[0].shortname)
+        self.server.expect(NODE, {ATTR_NODE_state: 'free'},
+                           id=self.moms.values()[1].shortname)
```
lisa-altair: Should max_attempts and interval be set on these 2 expect() too?

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -0,0 +1,204 @@
...
+        self.server.manager(MGR_CMD_SET, HOOK, {'enabled': 'False'}, id="h1")
+
+        j2 = Job(TEST_USER)
+        j2.set_attributes({ATTR_l + '.select': '1:vnode=' + vname})
+        jid2 = self.server.submit(j2)
+        self.du.run_cmd(self.server.hostname, cmd=pbsnodes_reset, sudo=True)
+        self.server.expect(NODE, {ATTR_NODE_state: 'free'},
+                           id=self.moms.values()[0].shortname)
+        self.server.manager(MGR_CMD_SET, SCHED, {'scheduling': 'True'})
+        self.server.expect(JOB, {ATTR_state: 'R'}, id=jid2)
```
lisa-altair: Since there is a second mom that has never been marked offline, I believe this test isn't really proving that the pbnodes -r problem is fixed because even with the bug the job can be run from the second mom.  Maybe you should either cause the second mom to be marked offline too.  Or add this part of the test to the single mom test above.

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -0,0 +1,204 @@
...
+        the hook alarms out and server executes the fail_action
+        After this check if vnodes are marked offline.
+        In case of a multiple mom reporting same set of vnodes,
+        it should not mark the vnodes and mom as offline because
+        there are other moms active and reporting same vnodes.
+        """
+        self.create_multi_vnodes(2, False)
+        self.create_mom_hook()
+
+        # set one  natural node to have higher ncpus than the other one
```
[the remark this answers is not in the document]
arungrover: Your understanding is correct. I'll add a comment.

test/tests/functional/pbs_offline_vnodes.py
arungrover: Ideally, it should not take more than a second or so to update the state that is why added max_attempts and interval in a few places. I'll update all other instances.

test/tests/functional/pbs_offline_vnodes.py
arungrover: Done

test/tests/functional/pbs_offline_vnodes.py
arungrover: Done

test/tests/functional/pbs_offline_vnodes.py
arungrover: That makes sense. Thanks for catching this. I'll move "pbsnodes -r" test to single node test.

arungrover:
Attaching t
[dev_ptl_test2.txt](https://github.com/PBSPro/pbspro/files/1790244/dev_ptl_test2.txt)
est logs

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -86,39 +87,80 @@ def test_single_mom_hook_failure_affects_vnode(self):
...
+        self.du.run_cmd(self.server.hostname, cmd=pbsnodes_reset, sudo=True)
+        self.server.delete(jid, wait=True)
+
+        j2 = Job(TEST_USER)
+        j2.set_attributes({ATTR_l + '.select': '1:vnode=' + vname})
+        jid2 = self.server.submit(j2)
+        self.server.expect(NODE, {ATTR_NODE_state: 'free'},
+                           id=self.moms.values()[0].shortname, max_attempts=3,
+                           interval=2)
+        self.server.manager(MGR_CMD_SET, SCHED, {'scheduling': 'True'})
```
lisa-altair: Since you don't make scheduling false anymore, you probably don't need to set it to true.  Submitting the job should cause a scheduling cycle to run.

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -130,10 +172,11 @@ def test_multi_mom_hook_failure_affects_vnode(self):
         it should not mark the vnodes and mom as offline because
         there are other moms active and reporting same vnodes.
         """
-        self.create_multi_vnodes(2, False)
+        self.create_multi_vnodes(2)
         self.create_mom_hook()
 
-        # set one  natural node to have higher ncpus than the other one
+        # set one  natural node to have higher ncpus than the other one so
```
lisa-altair: nit: extra space before natural

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -86,39 +87,80 @@ def test_single_mom_hook_failure_affects_vnode(self):
...
+                max_attempts=3, interval=2)
             self.server.expect(
-                NODE, {ATTR_NODE_state: 'offline'}, id='vnode[1]')
+                NODE, {ATTR_NODE_state: 'offline'}, id='vnode[1]',
+                max_attempts=3, interval=2)
+
+        mom_host = self.moms.values()[0].shortname
+        pbs_exec = self.server.pbs_conf['PBS_EXEC']
+        pbsnodes_cmd = os.path.join(pbs_exec, 'bin', 'pbsnodes')
+        pbsnodes_reset = pbsnodes_cmd + ' -r ' + mom_host
```
lisa-altair: Shouldn't we also check that all of the nodes are free (i.e. no longer offline)?

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -162,43 +205,25 @@ def test_multi_mom_hook_failure_affects_vnode(self):
...
-        self.du.run_cmd(self.server.hostname, cmd=pbsnodes_reset, sudo=True)
-        self.server.expect(NODE, {ATTR_NODE_state: 'free'},
-                           id=self.moms.values()[0].shortname)
-        self.server.manager(MGR_CMD_SET, SCHED, {'scheduling': 'True'})
-        self.server.expect(JOB, {ATTR_state: 'R'}, id=jid2)
+            self.server.expect(NODE, {ATTR_NODE_state: 'free'}, id='vnode[0]',
+                               max_attempts=3, interval=2)
+            self.server.expect(NODE, {ATTR_NODE_state: 'free'}, id='vnode[1]',
+                               max_attempts=3, interval=2)
+        self.server.manager(MGR_CMD_DELETE, NODE, id="@default")
```
lisa-altair: Out of curiosity, why do we have to delete the nodes?  Won't the next setUp do that anyway?

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+class TestOfflineVnodeOnHookFailure(TestFunctional):
+    """
+    Tests if vnodes are marked offline if a hook fails
+    """
+
+    def setUp(self):
+        if len(self.moms.values()) != 2:
+            self.skipTest("Provide 2 moms while invoking test")
```
lisa-altair: Hey @arungrover, since the first test is really a single mom test, can we move this out of here and into the multi-mom test case?  That way the single mom test can get run most of the time.

test/tests/functional/pbs_offline_vnodes.py
arungrover: I think that is not needed because there wasn't a problem with "pbsnodes -r" not freeing up the nodes.

test/tests/functional/pbs_offline_vnodes.py
arungrover: You are right, this is not needed.

test/tests/functional/pbs_offline_vnodes.py
arungrover: I'll change it. Thanks!

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -86,39 +87,80 @@ def test_single_mom_hook_failure_affects_vnode(self):
...
+                max_attempts=3, interval=2)
+
+        mom_host = self.moms.values()[0].shortname
+        pbs_exec = self.server.pbs_conf['PBS_EXEC']
+        pbsnodes_cmd = os.path.join(pbs_exec, 'bin', 'pbsnodes')
+        pbsnodes_reset = pbsnodes_cmd + ' -r ' + mom_host
+
+        # Set mom sync hook timeout to be a low value because if mom fails to
+        # get the hook after disabling it then next sync will happen after
+        # 2 minutes by default.
```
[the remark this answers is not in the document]
arungrover: I'll add it.

test/tests/functional/pbs_offline_vnodes.py
arungrover: Yes this step is not needed.

test/tests/functional/pbs_offline_vnodes.py
arungrover: I'll change it

arungrover:
[test_on_linux_single_mom.txt](https://github.com/PBSPro/pbspro/files/1804920/test_on_linux_single_mom.txt)
[test_on_linux.txt](https://github.com/PBSPro/pbspro/files/1804921/test_on_linux.txt)
[test_on_UV.txt](https://github.com/PBSPro/pbspro/files/1804922/test_on_UV.txt)

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -90,26 +83,28 @@ def test_single_mom_hook_failure_affects_vnode(self):
...
+                           id=single_mom.shortname, max_attempts=3,
                            interval=2)
         j1.set_sleep_time(1000)
         jid = self.server.submit(j1)
         self.server.expect(JOB, {ATTR_state: 'Q'}, id=jid)
 
         self.server.expect(
             NODE, {ATTR_NODE_state: 'offline'},
-            id=self.moms.values()[0].shortname, max_attempts=3, interval=2)
+            id=single_mom.shortname, max_attempts=20, interval=2)
```
lisa-altair: Earlier you had expected the check to be quick and that's why max_attempts=3.  But now it is set to 20.  Is that because you added UV testing?

test/tests/functional/pbs_offline_vnodes.py
lisa-altair: Okay.  Thanks.

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -90,26 +83,28 @@ def test_single_mom_hook_failure_affects_vnode(self):
         Once offlined, reset the mom by issueing pbsnodes -r
         and check if the job runs on one of the vnodes.
         """
-        self.create_multi_vnodes(1)
+        single_mom = ""
+        if len(self.moms.values()) != 1:
+            self.server.manager(MGR_CMD_DELETE, NODE, id="@default",
+                                expect=True)
+        single_mom = self.moms.values()[0]
+
```
lisa-altair: Do we need to add the following code in case the mom provided has an unwanted vnodedef file (sorry I can't get the formatting of the comment right)
```
if self.moms.values()[0].has_vnode_defs():
       self.moms.values()[0].delete_vnode_defs()
```

arungrover: I think so, this is just to safe guard that the two moms being added do not come with already existing vnode definitions.

test/tests/functional/pbs_offline_vnodes.py
arungrover: Yes Lisa this is because of UV testing. I found out that it took longer on a UV to set the nodes to offline state. Not sure why it was so but thought it is better to increase the number of attempts.

test/tests/functional/pbs_offline_vnodes.py
lisa-altair: I meant do we need to add that code to the single mom test case.  :)

test/tests/functional/pbs_offline_vnodes.py
lisa-altair: Okay.  Thanks for confirming.

test/tests/functional/pbs_offline_vnodes.py
arungrover: oh ok, I get it. I can add it, but if you are okay I'd like not to touch the single node test. The test is very delicate and is now working on all 3 platforms (cray, uv and normal linux). I'm worried that changing it may cause trouble on one or more of these platforms.

arungrover:
[test_on_linux_single.txt](https://github.com/PBSPro/pbspro/files/1808700/test_on_linux_single.txt)
[test_on_linux.txt](https://github.com/PBSPro/pbspro/files/1808701/test_on_linux.txt)
[test_on_UV.txt](https://github.com/PBSPro/pbspro/files/1808702/test_on_UV.txt)

test/tests/functional/pbs_offline_vnodes.py
lisa-altair: Since it's working then it seems the change is not necessary.  I certainly don't want to upset the delicate balance if it is working.  :)  Thanks for the response.

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -0,0 +1,262 @@
...
+        # node resources start showing up
+            time.sleep(2)
+        # No need to create vnodes on a cpuset mom
+        cpuset_mom = single_mom.is_cpuset_mom()
+        if cpuset_mom is False:
+            self.create_multi_vnodes(1)
+        self.create_mom_hook()
+
+        j1 = Job(TEST_USER)
+        self.server.expect(NODE, {ATTR_NODE_state: 'free'},
```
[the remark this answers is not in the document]
agrawalravi90: also, you might not need this expect if you set expect=True in create_vnodes() inside create_multi_vnodes()

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -0,0 +1,262 @@
...
+                           id=single_mom.shortname, max_attempts=3,
+                           interval=2)
+        j1.set_sleep_time(1000)
+        jid = self.server.submit(j1)
+        self.server.expect(JOB, {ATTR_state: 'Q'}, id=jid)
```
agrawalravi90: Can you please add a comment here explaining why you expect the job to be in 'Q' state and the node to be offline? I know you explained it in the docstring, but code comments help more.

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -0,0 +1,262 @@
...
+
+        self.server.expect(
+            NODE, {ATTR_NODE_state: 'offline'},
+            id=single_mom.shortname, max_attempts=20, interval=2)
+
+        vname = "vnode[0]"
+        if self.is_cray is True:
+            vlist = []
+            first_vnode = 1
```
agrawalravi90: how about setting vname to None when you enter the if block and just checking for vname inside the "for v1 in vlist" loop? if it's None, set it to v1. That way you won't need 'first_vnode' at all.

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -0,0 +1,262 @@
...
+            for v1 in vlist:
+                if first_vnode == 1:
+                    first_vnode += 1
+                    vname = v1
+                # Check that the vnode is in offline state
+                self.server.expect(
+                    VNODE, {'state': 'offline'}, id=v1, max_attempts=3,
+                    interval=2)
+        elif cpuset_mom is True:
+            first_vnode = 1
```
agrawalravi90: same comment here about removing first_vnode

test/tests/functional/pbs_offline_vnodes.py
agrawalravi90: in fact, you don't really need to set vname before the if-else blocks, you could just check whether vname is None after the if-else blocks, if it is then set it to 'vnode[0]'

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -0,0 +1,262 @@
...
+                    if first_vnode == 1:
+                        first_vnode += 1
+                        vname = v1['resources_available.vnode']
+                    # Check that the vnode is in offline state
+                    self.server.expect(
+                        VNODE, {'state': 'offline'},
+                        id=v1['resources_available.vnode'], max_attempts=3,
+                        interval=2)
+        else:
+            self.server.expect(
```
agrawalravi90: didn't you create just one vnode? how can there be 2?

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -0,0 +1,262 @@
...
+        pbs_exec = self.server.pbs_conf['PBS_EXEC']
+        pbsnodes_cmd = os.path.join(pbs_exec, 'bin', 'pbsnodes')
+        pbsnodes_reset = pbsnodes_cmd + ' -r ' + mom_host
+
+        # Set mom sync hook timeout to be a low value because if mom fails to
+        # get the hook after disabling it then next sync will happen after
+        # 2 minutes by default and we don't want to wait that long.
+        self.server.manager(MGR_CMD_SET, SERVER,
+                            {'sync_mom_hookfiles_timeout': '5'})
+        self.server.manager(MGR_CMD_SET, HOOK, {'enabled': 'False'}, id="h1")
```
agrawalravi90: Can you please change "create_mom_hook()" to return the id of the hook that it created, store that in a variable when calling the function and use that variable here instead of "h1"?

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -0,0 +1,262 @@
...
+        # that the job only goes to this natural node.
+        self.server.manager(MGR_CMD_SET, NODE, {
+                            ATTR_rescavail + '.ncpus': '256'},
+                            id=self.moms.values()[0].shortname)
+        self.server.manager(MGR_CMD_SET, NODE, {
+                            ATTR_rescavail + '.ncpus': '1'},
+                            id=self.moms.values()[1].shortname)
+
+        j1 = Job(TEST_USER)
+        self.server.expect(NODE, {ATTR_NODE_state: 'free'},
```
agrawalravi90: If you say expect=True in create_vnodes() inside create_multi_vnodes() then you probably won't need these expects here.

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -0,0 +1,262 @@
...
+        # The moms provided to the test may have unwanted vnodedef files.
+        if self.moms.values()[0].has_vnode_defs():
+            self.moms.values()[0].delete_vnode_defs()
+        if self.moms.values()[1].has_vnode_defs():
+            self.moms.values()[1].delete_vnode_defs()
+
+        self.create_multi_vnodes(2)
+        self.create_mom_hook()
+
+        # set one natural node to have higher ncpus than the other one so
```
agrawalravi90: just a suggestion: why not just have the job ask for 'vnode' to be one of the natural vnodes' resources_available.vnode value? that's a sure way of making the job run on a particular vnode.

arungrover: Well I can do that but even with ncpus it is almost doing the same thing. So if you do not feel strongly about it, I'd like to retain it as is.

arungrover:
@agrawalravi90 I've addressed your comments. Please have a look.
@lisa-altair there has been some more changes to test script (mostly refactoring) please have a look at it again.

arungrover:
[test_on_linux_single.txt](https://github.com/PBSPro/pbspro/files/1817501/test_on_linux_single.txt)
[test_on_linux.txt](https://github.com/PBSPro/pbspro/files/1817502/test_on_linux.txt)
[test_on_UV.txt](https://github.com/PBSPro/pbspro/files/1817503/test_on_UV.txt)

test/tests/functional/pbs_offline_vnodes.py
arungrover: Since both these tests need to create only one hook I just made it a function for simplicity. If you feel strongly about changing it then I can do that otherwise I'd like to keep it as is.

vccardenas:
@arungrover , the changes look good to me.

PR #589, opened 2018-03-07
PP-1242: Windows pbs_mom.exe crash in check_pwd()

sid2364:
#### Bug/feature Description
* *pbs_mom crashes*
* *Bugs: On submission of jobs from a very large number of users, check_pwd() fails.*
* *Bugs: When a lot of users (~100) are configured on the system (on Windows) this problem can be reproduced.*
* *[PP-1242](https://pbspro.atlassian.net/browse/PP-1242)*

#### Affected Platform(s)
* *Windows*

* *We cache user login information in an array. Initially the size of this array is limited to just 4 users. As more users are added to the system, information for them also needs to be added to the cache; we have a call to realloc() to resize the array, but don't handle this properly while passing back references of older users. This results in the MoM crashing, since these older entries in the cache array are now pointing to invalid memory locations after the call to realloc().*

#### Solution Description
* *Instead of caching this information in arrays, we can use linked lists instead. This would solve the problem of keeping track of a "resizing" cache, and there is no reallocation of memory involved.*

#### Checklist:

### DIFF
--- a/src/include/list_link.h
+++ b/src/include/list_link.h
@@ -78,12 +78,12 @@ typedef pbs_list_link pbs_list_head;
 
 /* function prototypes */
 
-extern void insert_link(pbs_list_link *old, pbs_list_link *new, void *pobj, int pos);
-extern void append_link(pbs_list_head *head, pbs_list_link *new, void *pnewobj);
-extern void delete_link(pbs_list_link *old);
+extern void insert_link(pbs_list_link *oldp, pbs_list_link *newp, void *pobj, int pos);
+extern void append_link(pbs_list_head *head, pbs_list_link *newp, void *pnewobj);
+extern void delete_link(pbs_list_link *oldp);
 extern void swap_link   (pbs_list_link *, pbs_list_link *);
-extern int  is_linked(pbs_list_link *head, pbs_list_link *old);
-extern void list_move(pbs_list_head *old, pbs_list_head *new);
+extern int  is_linked(pbs_list_link *head, pbs_list_link *oldp);
+extern void list_move(pbs_list_head *oldp, pbs_list_head *newp);
 
 #ifndef NDEBUG
 extern void *get_next(pbs_list_link, char *file, int line);
--- a/src/lib/Libifl/list_link.c
+++ b/src/lib/Libifl/list_link.c
@@ -69,8 +69,8 @@
  *	Entry is added either before (position=0) or after (position !=0)
  *	an old entry.
  *
- * @param[in] old	ptr to old entry in list
- * @param[in] new  	ptr to new link entry
+ * @param[in] oldp	ptr to old entry in list
+ * @param[in] newp	ptr to new link entry
  * @param[in] pobj 	ptr to object to link in
  * @param[in] position  0=before old, else after
  *
@@ -78,7 +78,7 @@
  */
 
 void
-insert_link(struct pbs_list_link *old, struct pbs_list_link *new,
+insert_link(struct pbs_list_link *oldp, struct pbs_list_link *newp,
  		void *pobj, int position)
 {
 
@@ -86,50 +86,50 @@ insert_link(struct pbs_list_link *old, struct pbs_list_link *new,
 	/* first make sure unlinked entries are pointing to themselves	    */
 
 	if ((pobj          == NULL)	      ||
-		(old           == NULL) ||
-		(old->ll_prior == NULL)        ||
-		(old->ll_next  == NULL)        ||
-		(new->ll_prior != (pbs_list_link *)new)      ||
-		(new->ll_next  != (pbs_list_link *)new)) {
+		(oldp           == NULL) ||
+		(oldp->ll_prior == NULL)        ||
+		(oldp->ll_next  == NULL)        ||
+		(newp->ll_prior != (pbs_list_link *)newp)      ||
+		(newp->ll_next  != (pbs_list_link *)newp)) {
 		(void)fprintf(stderr, "Assertion failed, bad pointer in insert_link\n");
 		abort();
 	}
 #endif
 
-	if (position == LINK_INSET_AFTER) {  /* insert new after old */
-		new->ll_prior = old;
-		new->ll_next  = old->ll_next;
-		(old->ll_next)->ll_prior = new;
-		old->ll_next = new;
-	} else {				/* insert new before old */
-		new->ll_next = old;
-		new->ll_prior = old->ll_prior;
-		(old->ll_prior)->ll_next = new;
-		old->ll_prior = new;
+	if (position == LINK_INSET_AFTER) {  /* insert newp after oldp */
+		newp->ll_prior = oldp;
+		newp->ll_next  = oldp->ll_next;
+		(oldp->ll_next)->ll_prior = newp;
+		oldp->ll_next = newp;
+	} else {				/* insert newp before oldp */
+		newp->ll_next = oldp;
+		newp->ll_prior = oldp->ll_prior;
+		(oldp->ll_prior)->ll_next = newp;
+		oldp->ll_prior = newp;
 	}
 	/*
 	 * its big trouble if ll_struct is null, it would make this
 	 * entry appear to be the head, so we never let that happen
 	 */
 	if (pobj)
-		new->ll_struct = pobj;
+		newp->ll_struct = pobj;
 	else
-		new->ll_struct = (void *)new;
+		newp->ll_struct = (void *)newp;
 }
 
 /**
  * @brief
  * 	-append_link - append a new entry to the end of the list
  *
- * @param[in] head      ptr to head of list
- * @param[in] new       ptr to new link entry
- * @param[in] pobj      ptr to object to link in
+ * @param[in] head		ptr to head of list
+ * @param[in] newp		ptr to new link entry
+ * @param[in] pobj		ptr to object to link in
  *
  * @return      Void
  */
 
 void
-append_link(pbs_list_head *head, pbs_list_head *new, void *pobj)
+append_link(pbs_list_head *head, pbs_list_head *newp, void *pobj)
 {
 
 #ifndef NDEBUG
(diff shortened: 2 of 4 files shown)


sid2364:
**Purify run output:-**
[pbs_mom_before.txt](https://github.com/PBSPro/pbspro/files/1788499/pbs_mom_before.txt)
[pbs_mom_after.txt](https://github.com/PBSPro/pbspro/files/1788500/pbs_mom_after.txt)

**Configuration:-**
100 users set up with pbs_password

**Tests run:-**
1. 100 jobs submitted from 100 users.
2. 100 array jobs submitted from 100 users.

src/include/list_link.h
```diff
@@ -78,12 +78,12 @@ typedef pbs_list_link pbs_list_head;
 
 /* function prototypes */
 
-extern void insert_link(pbs_list_link *old, pbs_list_link *new, void *pobj, int pos);
-extern void append_link(pbs_list_head *head, pbs_list_link *new, void *pnewobj);
+extern void insert_link(pbs_list_link *old, pbs_list_link *newp, void *pobj, int pos);
+extern void append_link(pbs_list_head *head, pbs_list_link *newp, void *pnewobj);
```
subhasisb: What is the significance of renaming the new to newp? Both cases the parameter was a pointer anyway!

src/lib/Libifl/list_link.c
```diff
@@ -89,47 +89,47 @@ insert_link(struct pbs_list_link *old, struct pbs_list_link *new,
 		(old           == NULL) ||
 		(old->ll_prior == NULL)        ||
 		(old->ll_next  == NULL)        ||
-		(new->ll_prior != (pbs_list_link *)new)      ||
-		(new->ll_next  != (pbs_list_link *)new)) {
+		(newp->ll_prior != (pbs_list_link *)newp)      ||
```
subhasisb: I feel you don't need these changes.

src/lib/Libifl/list_link.c
```diff
@@ -89,47 +89,47 @@ insert_link(struct pbs_list_link *old, struct pbs_list_link *new,
...
-		old->ll_prior = new;
+		newp->ll_prior = old;
+		newp->ll_next  = old->ll_next;
+		(old->ll_next)->ll_prior = newp;
+		old->ll_next = newp;
+	} else {				/* insert newp before old */
+		newp->ll_next = old;
+		newp->ll_prior = old->ll_prior;
+		(old->ll_prior)->ll_next = newp;
+		old->ll_prior = newp;
```
subhasisb: Again, why this renaming?

src/lib/Libwin/passwd.c
```diff
@@ -2688,23 +2688,26 @@ getHomedir(char *user)
...
 
 	if (homedir)
 		return (homedir);
 
-	/* look for a user login handle */
-	for (i=0; i<user_num; i++) {
-		struct passwd *pwdp;
-
-		pwdp = &user_array[i];
+	if(user_num == 0){
```
subhasisb: indentation - add space between ) and { and you don't need the flower braces anyway

src/lib/Libwin/passwd.c
```diff
@@ -2688,23 +2688,26 @@ getHomedir(char *user)
...
+		CLEAR_HEAD(passwd_cache_ll);
+	}
 
+	/* look for a user login handle */
+	pwdp = (struct passwd*) GET_NEXT(passwd_cache_ll);
+	while(pwdp){
```
subhasisb: indentation, add space

src/lib/Libwin/passwd.c
subhasisb: Since it is a linked list of user password structures anyway, why do you need the user_num variable at all?

src/lib/Libwin/passwd.c
```diff
@@ -3949,15 +3952,19 @@ setuser_with_password(char *user,
 	if ((user == NULL) || (decrypt_func == NULL))
 		return (0);
 
-	/* look in internal array for saved usertoken handle */
-	for (i=0; i<user_num; i++) {
-		pwdp = &user_array[i];
+	if(user_num == 0){
```
subhasisb: This variable should not be required. Initialize (CLEAR_HEAD) the linked list head once at startup, and that should we all required.

src/lib/Libwin/passwd.c
```diff
@@ -4028,13 +4035,17 @@ setuser_close_handle()
...
 
-		/* look in internal array for saved usertoken handle */
-		for (i=0; i<user_num; i++) {
-			pwdp = &user_array[i];
+		if(user_num == 0){
+			CLEAR_HEAD(passwd_cache_ll);
+		}
 
+		/* look in internal array for saved usertoken handle */
+		pwdp = (struct passwd*) GET_NEXT(passwd_cache_ll);
```
subhasisb: convert into for loop (all places) like this:
	for (pwdp = (struct passwd*) GET_NEXT(passwd_cache_ll); pwdp;	pwdp = (struct passwd*) GET_NEXT(pwdp->pw_allpasswds)) {

src/lib/Libwin/passwd.c
```diff
@@ -4647,137 +4665,112 @@ add_pwentry(char *name,
...
-			user_array[i].pw_uid = NULL;
-			user_array[i].pw_gid = NULL;
-			user_array[i].pw_gecos = NULL;
-			user_array[i].pw_dir = NULL;
-			user_array[i].pw_userlogin = INVALID_HANDLE_VALUE;
-		}
-	}
-
-	if (user_array == NULL)
+	pwdn = (struct passwd *)malloc(sizeof(struct passwd));
```
subhasisb: Why malloc and memset? Simply use calloc directly!

src/include/list_link.h
```diff
@@ -78,12 +78,12 @@ typedef pbs_list_link pbs_list_head;
 
 /* function prototypes */
 
-extern void insert_link(pbs_list_link *old, pbs_list_link *newp, void *pobj, int pos);
-extern void append_link(pbs_list_head *head, pbs_list_link *newp, void *pnewobj);
```
subhasisb: Sorry for making you do this change back, but from your build report i figured out the problem. Some other tools that use this header but uses C++ to compile are failing. So let's go ahead and change it back to what you had before, which is newp.

src/lib/Libwin/passwd.c
```diff
@@ -3952,19 +3951,18 @@ setuser_with_password(char *user,
 	if ((user == NULL) || (decrypt_func == NULL))
 		return (0);
 
-	if(user_num == 0){
+	if(user_num == 0)
```
subhasisb: Not sure why we still need user_num? It is a linked list, so it can be initialized by calling CLEAR_HEAD only once at server startup, and thus need not be called in every routine that you are calling here

src/lib/Libwin/passwd.c
```diff
@@ -4630,12 +4627,12 @@ print_pwentries(void)
 	/* look for a password */
 	struct passwd* pwdp;
 
-	if(user_num == 0){
+	if(user_num == 0)
```
subhasisb: Same comment as above.

src/lib/Libwin/passwd.c
```diff
@@ -4956,18 +4950,17 @@ getpwnam(const char *name)
 	struct  passwd *pwdp = NULL;
 	int found = 0;
 
-	if(user_num == 0){
+	if(user_num == 0)
```
subhasisb: Same comment as above. We should not need to check for user_num == 0 in each function, rather do a CLEAR_HEAD at server startup. That way you do not even need this variable. There is also a formatting error: Space between if and (

src/lib/Libwin/passwd.c
sid2364: Okay, I'll change this.

src/include/list_link.h
sid2364: Done.

src/lib/Libwin/passwd.c
sid2364: So, I've added a call to CLEAR_HEAD during server init (in pbsd_main.c), but also have one call to CLEAR_HEAD for pw_allpasswds in cache_usertoken_and_homedir since pbs_account has an entry point there and will crash if it's not initialised. This is called during the postinstall, which is why I've not removed that one.

src/server/pbsd_main.c
```diff
@@ -1291,6 +1291,10 @@ main(int argc, char **argv)
 	CLEAR_HEAD(svr_exechost_startup_hooks);
 	CLEAR_HEAD(svr_execjob_attach_hooks);
 	CLEAR_HEAD(svr_allscheds);
+#ifdef WIN32
+	if(!passwd_cache_init)
+		CLEAR_HEAD(passwd_cache_ll);
```
subhasisb: Why are you even checking passwd_cache_init? Why not simply call CLEAR_HEAD(passwd_cache_ll)? Just like how all the other lists are being initialized?

src/lib/Libwin/passwd.c
```diff
@@ -5036,12 +5016,15 @@ cache_usertoken_and_homedir(char *user,
 	if (user == NULL)
 		return;
 
-	if(user_num == 0)
+	if(passwd_cache_init == 0){
```
subhasisb: Why do the CLEAR_HEAD here at all? Why not just leave it to pbsd_main/init, where it gets initialized anyway?

sid2364: It gets initialised there after server startup, but these functions are called from pbs_account and pbs_ds_monitor as well. While testing these changed I saw both these crashing if I wasn't initialising the list here.

src/server/pbsd_main.c
sid2364: Okay

src/lib/Libwin/passwd.c
subhasisb: Ah ok, i see. Thanks for explaining that. In that case, it makes sense the way you had done it earlier. Since you are anyway going to have the "check and initialize" in the individual functions itself, there is no need to also add it to pbsd_main.c. Let's do only one kind of initialization. Otherwise it looks good.

sid2364: Okay, will revert those changes.

sid2364:
@subhasisb have addressed your comments.

src/lib/Libwin/passwd.c
```diff
@@ -2688,21 +2687,24 @@ getHomedir(char *user)
 	char	*homedir = NULL;
 	HANDLE	userlogin = INVALID_HANDLE_VALUE;
 	int	i;
+	struct passwd *pwdp;
```
bremanandjk: I feel, Its good to initialize NULL for avoiding dangling pointer related crashes due to a programming error.

sid2364: Done

sid2364:
@subhasisb, @bremanandjk, thank you for reviewing this. 

I've rebased and squashed my commits.

PR #578, opened 2018-02-26
Add automated test for cgroup race condition crash

vstumpf:
#### Bug/feature Description
Testing #555 can be automated.

#### Affected Platform(s)
N/A

#### Solution Description
This PR has a test for the hook crash.

#### Checklist:
- [x] I have joined the **[pbspro community forum](http://community.pbspro.org/)**.
- [x] My pull request contains a **single, signed** commit. See **[setting up gpg signature](https://pbspro.atlassian.net/wiki/display/DG/Signing+Your+Git+Commits)**.
- [x] My code follows the **[coding style](https://pbspro.atlassian.net/wiki/display/DG/Coding+Standards)** of this project.
- [ ] My change requires project documentation. See **[required documentation checklist](https://pbspro.atlassian.net/wiki/display/DG/Checklist+for+Developing+Features+and+Bug+Fixes)** for details.
   - [ ] I have added documentation in the **[project documentation area](https://pbspro.atlassian.net/wiki/display/PD)**.
- [x] I have added new **PTL test(s) to my commit**. (See **[using PTL for testing](https://pbspro.atlassian.net/wiki/display/DG/Using+PTL+for+Testing)**) *(or)*
   - [ ] I have added  **manual test(s) to this pull request and explained why PTL is not appropriate** for this case.
- [x] All new and existing automated tests have passed. (See **[running automated PTL tests](https://pbspro.atlassian.net/wiki/display/DG/PTL+Quick+Start+Guide)**).
- [x] I have attached **test logs to this pull request** as evidence of testing/verification.

__***For further information please visit the [Developer Guide Home](https://pbspro.atlassian.net/wiki/display/DG/Developer+Guide+Home).***__

[cg-afterfix.log](https://github.com/PBSPro/pbspro/files/1760229/cg-afterfix.log)
[cg-beforefix.log](https://github.com/PBSPro/pbspro/files/1760230/cg-beforefix.log)

PR #593, opened 2018-03-13
Fix race condition between cgroup periodic and begin handler

minghui-liu:
#### Bug/feature Description
There is a race condition between cgroup hook's `exechost_periodic` and `execjob_begin` handler that causes some jobs' cgroup to be mistaken as orphans and cleaned up by periodic hook. 

#### Affected Platform(s)
All platform that cgroup hook supports

The periodic hook relies on a list of jobs written to mom hook input file for identifying orphaned cgroups. However there is a gap between the time mom writes the list to the input file and the time that periodic hook is run. During this time gap, a begin hook of a new job could run and creates a cgroup, which could be cleaned up by periodic hook. Even with new locking method in place, the race still exists since mom itself does not participate in the locking. 

#### Solution Description
Let the begin hook write out a `.new` file to indicate a new job. The periodic handler can look for these `.new` files and `.JB` files and compile a list of job ids and avoid cleaning up cgroups corresponding to these job ids.  Hooks after the begin hook will need to clean these files up because after the begin hook finish a `.JB` file will be created. 

No new PTL test because `test_cgroups_race_condition` in #578 will also test this issue.

#### Checklist:

[PTL test log](https://github.com/PBSPro/pbspro/files/1831749/test_output2.txt)

### DIFF
--- a/src/hooks/cgroups/pbs_cgroups.PY
+++ b/src/hooks/cgroups/pbs_cgroups.PY
@@ -689,6 +689,9 @@ class HookUtils(object):
                    caller_name())
         pbs.logmsg(pbs.EVENT_DEBUG4, "%s: Host assigned job resources: %s" %
                    (caller_name(), jobutil.assigned_resources))
+        # add jobid to cgroup_jobs file to tell periodic handler that this
+        # job is new and its cgroup should not be cleaned up
+        cgroup.add_jobid_to_cgroup_jobs(event.job.id)
         # Make sure the parent cgroup directories exist
         cgroup.create_paths()
         # Make sure the cgroup does not already exist
@@ -737,6 +740,9 @@ class HookUtils(object):
         Handler for execjob_epilogue events.
         """
         pbs.logmsg(pbs.EVENT_DEBUG4, "%s: Method called" % caller_name())
+        # delete this jobid from cgroup_jobs in case hook events before me
+        # failed to do that
+        cgroup.remove_jobid_from_cgroup_jobs(event.job.id)
         # The resources_used information has a base type of pbs_resource.
         # Update the usage data
         cgroup.update_job_usage(event.job.id, event.job.resources_used)
@@ -751,6 +757,9 @@ class HookUtils(object):
         Handler for execjob_end events.
         """
         pbs.logmsg(pbs.EVENT_DEBUG4, "%s: Method called" % caller_name())
+        # delete this jobid from cgroup_jobs in case hook events before me
+        # failed to do that
+        cgroup.remove_jobid_from_cgroup_jobs(event.job.id)
         # The cgroup is usually deleted in the execjob_epilogue event
         # There are certain corner cases where epilogue can fail or skip
         # Delete files again here to make sure we catch those
@@ -777,6 +786,9 @@ class HookUtils(object):
         Handler for execjob_launch events.
         """
         pbs.logmsg(pbs.EVENT_DEBUG4, "%s: Method called" % caller_name())
+        # delete this jobid from cgroup_jobs in case hook events before me
+        # failed to do that
+        cgroup.remove_jobid_from_cgroup_jobs(event.job.id)
         # Add the parent process id to the appropriate cgroups.
         cgroup.add_pids(os.getppid(), jobutil.job.id)
         # FUTURE: Add environment variable to the job environment
@@ -911,6 +923,9 @@ class HookUtils(object):
         Handler for execjob_attach events.
         """
         pbs.logmsg(pbs.EVENT_DEBUG4, "%s: Method called" % caller_name())
+        # delete this jobid from cgroup_jobs in case hook events before me
+        # failed to do that
+        cgroup.remove_jobid_from_cgroup_jobs(event.job.id)
         pbs.logjobmsg(jobutil.job.id, "%s: Attaching PID %s" %
                       (caller_name(), event.pid))
         # Add the job process id to the appropriate cgroups.
@@ -1730,6 +1745,11 @@ class CgroupUtils(object):
         self.host_job_env_dir = os.path.join(PBS_MOM_HOME, 'aux')
         self.host_job_env_filename = os.path.join(self.host_job_env_dir,
                                                   '%s.env')
+        # temporarily stores list of new jobs that came after job_list was
+        # written to mom hook input file (work around for periodic
+        # and begin race condition)
+        self.cgroup_jobs_file = os.path.join(PBS_MOM_HOME, 'mom_priv', 'hooks',
+                                             'hook_data', 'cgroup_jobs')
         # information for offlining nodes
         self.offline_file = os.path.join(PBS_MOM_HOME, 'mom_priv', 'hooks',
                                          ('%s.offline' %
@@ -3445,12 +3465,29 @@ class CgroupUtils(object):
         """
         pbs.logmsg(pbs.EVENT_DEBUG4, "%s: Method called" % caller_name())
         pbs.logmsg(pbs.EVENT_DEBUG4, "Local jobs: %s" % (local_jobs))
+        new_jobs = set()
+        mom_priv_jobs_dir = os.path.join(PBS_MOM_HOME, 'mom_priv', 'jobs')
+        # Make a list of jobs from jobids in cgroup_jobs file and .JB files.
+        # These jobs are new since local_jobs was written to mom hook input
+        # file. We should not mistake their cgroups as orphans
+        cgroup_jobs = self.read_cgroup_jobs()
+        pbs.logmsg(pbs.EVENT_DEBUG4,
+                   'cgroup_jobs file content: %s' % cgroup_jobs)
+        if cgroup_jobs:
+            for jobid in cgroup_jobs:
+                new_jobs.add(jobid)
+            self.empty_cgroup_jobs_file()
+        for p in glob.glob(os.path.join(mom_priv_jobs_dir, '*.JB')):
+            pbs.logmsg(pbs.EVENT_DEBUG4, 'Found file %s' % p)
+            jobid = os.path.splitext(os.path.basename(p))[0]
+            new_jobs.add(jobid)
+        pbs.logmsg(pbs.EVENT_DEBUG4, 'New jobs: %s' % str(new_jobs))
         remaining = 0
         for key in self.paths:
             path = os.path.dirname(self._cgroup_path(key))
             for subdir in glob.glob(os.path.join(path, '[0-9]*')):
                 jobid = os.path.basename(subdir)
-                if jobid not in local_jobs:
+                if jobid not in local_jobs and jobid not in new_jobs:
                     if not jobid.endswith('-orphan'):
                         try:
                             os.rename(subdir, subdir + '-orphan')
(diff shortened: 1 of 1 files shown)


mike0042:
- Which call to cleanup_orphans is removing cgroups it shouldn't? Is it the call from the exechost_periodic handler where the job list is provided by the event? Or is it the call from configure_job where the job list is acquired by calling node.gather_jobs_on_node? If the job list is incomplete, then we should fix the problem at the source and not provide a workaround.

If we must provide a workaround...

- We should not be writing files into mom_priv/jobs if we can avoid it. It's just one more thing the hook is responsible for cleaning up. (And another thing it will get blamed for if it doesn't clean up.)

- If we must keep a record of jobs not to delete, then I suggest we use the file mom_priv/hooks/hook_data/cgroup_jobs (os.path.join(self.hook_storage_dir, 'croup_jobs')) and write the list into the file. Now that the cgroup hook maintains a lock, we don't need to worry about multiple writers.

minghui-liu:
The `cleanup_orphans` call in `exechost_periodic_handler()`. The job list here is provided by the event object. To make the list up to date when periodic handler runs we will need to change the hook infrastructure.

Yes you are right. Writing files to `mom_priv/jobs` was not a great decision. I will use `mom_priv/hooks/hook_data/cgroup_jobs` instead. Thank you for the suggestion!

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -3445,12 +3458,30 @@ class CgroupUtils(object):
         """
         pbs.logmsg(pbs.EVENT_DEBUG4, "%s: Method called" % caller_name())
         pbs.logmsg(pbs.EVENT_DEBUG4, "Local jobs: %s" % (local_jobs))
+        new_jobs = set()
+        mom_priv_jobs_dir = os.path.join(PBS_MOM_HOME, 'mom_priv', 'jobs')
+        # search for .new files and .JB files and extract the jobid
```
mike0042: You are no longer looking for "*.new" files. Please update the comment.

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -911,6 +921,9 @@ class HookUtils(object):
         Handler for execjob_attach events.
         """
         pbs.logmsg(pbs.EVENT_DEBUG4, "%s: Method called" % caller_name())
+        # delete new job marker files here in case hook events before me
```
mike0042: You are no longer looking for "*.new" files. Please update the comment.

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -3735,6 +3766,61 @@ class CgroupUtils(object):
                 raise
         return self.assigned_resources is not None
 
+    @staticmethod
+    def add_jobid_to_cgroup_jobs(jobid):
+        filename = os.path.join(PBS_MOM_HOME, 'mom_priv', 'hooks', 'hook_data',
+                                'cgroup_jobs')
```
bayucan: Can this filename assignment be made into a global/class variable? I see it repeated 5 times in various functions.

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -777,8 +786,9 @@ class HookUtils(object):
         Handler for execjob_launch events.
         """
         pbs.logmsg(pbs.EVENT_DEBUG4, "%s: Method called" % caller_name())
-        # Add the parent process id to the appropriate cgroups.
-        cgroup.add_pids(os.getppid(), jobutil.job.id)
```
bayucan: This line disappeared on the new code. Was this intended? Are we fixing another bug?

minghui-liu: No it was by accident. Fixed.

src/hooks/cgroups/pbs_cgroups.PY
minghui-liu: Done.

src/hooks/cgroups/pbs_cgroups.PY
minghui-liu: Fixed.

minghui-liu:
I don't think this needs a new PTL test because it would be the same as #578.

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -3735,6 +3772,48 @@ class CgroupUtils(object):
                 raise
         return self.assigned_resources is not None
 
+    def add_jobid_to_cgroup_jobs(self, jobid):
+        pbs.logmsg(pbs.EVENT_DEBUG4, 'Adding jobid %s to cgroup_jobs' % jobid)
+        with open(self.cgroup_jobs_file, 'a+') as f:
```
mike0042: What if the jobid is already in the list? Do you still want to append it?
What if the open fails? Should you add a try/except block? If you do, don't forget to add an except for KeyboardInterrupt.

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -3735,6 +3772,48 @@ class CgroupUtils(object):
...
+            f.write(jobid + '\n')
+        return True
+
+    def remove_jobid_from_cgroup_jobs(self, jobid):
+        if os.path.isfile(self.cgroup_jobs_file):
+            with open(self.cgroup_jobs_file, 'r+') as f:
+                jobid_list = f.readlines()
+                f.seek(0)
+                for j in jobid_list:
+                    if jobid != j.strip():
```
mike0042: Or you could read/write the list as a single line...
```
$ python
Python 2.7.5 (default, Aug  4 2017, 00:39:18)
[GCC 4.8.5 20150623 (Red Hat 4.8.5-16)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> desc = open('myfile', 'r+')
>>> mylist = desc.readline().split()
>>> type(mylist)
<type 'list'>
>>> print mylist
['123.foo', '456.bar', '789.baz']
>>> newlist = [x for x in mylist if x != '456.bar']
>>> print newlist
['123.foo', '789.baz']
>>> desc.seek(0)
>>> desc.write(' '.join(newlist))
>>> desc.truncate()
>>> desc.close()
>>> exit()
```

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -3735,6 +3772,48 @@ class CgroupUtils(object):
...
         return self.assigned_resources is not None
 
+    def add_jobid_to_cgroup_jobs(self, jobid):
+        pbs.logmsg(pbs.EVENT_DEBUG4, 'Adding jobid %s to cgroup_jobs' % jobid)
+        with open(self.cgroup_jobs_file, 'a+') as f:
+            f.write(jobid + '\n')
+        return True
+
+    def remove_jobid_from_cgroup_jobs(self, jobid):
+        if os.path.isfile(self.cgroup_jobs_file):
```
mike0042: Don't bother checking if the file exists. Open it in a try block.

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -3735,6 +3772,48 @@ class CgroupUtils(object):
...
+                   'Deleting file: %s' % self.cgroup_jobs_file)
+        if os.path.isfile(self.cgroup_jobs_file):
+            os.remove(self.cgroup_jobs_file)
+        return True
+
+    def empty_cgroup_jobs_file(self):
+        pbs.logmsg(pbs.EVENT_DEBUG4,
+                   'Emptying file: %s' % self.cgroup_jobs_file)
+        with open(self.cgroup_jobs_file, 'w+'):
+            pass
```
mike0042: You don't need 'w+', just use 'w'. You need to call truncate() on the descriptor.

src/hooks/cgroups/pbs_cgroups.PY
minghui-liu: Yes.  When the list is read, it is read into a set so duplicates will be removed. 
I will add a try/except.

src/hooks/cgroups/pbs_cgroups.PY
minghui-liu: I put a '+' because '+' will create the file if it does not exist. Or is there a better way to do this?

mike0042: I see. In that case, this is fine.

src/hooks/cgroups/pbs_cgroups.PY
mike0042: Since you're using sets, which guarantee unique entries, why not use set.add() and set.discard()?

minghui-liu: Great idea. Thanks. Done

lisa-altair:
@minghui-liu  said:
> I don't think this needs a new PTL test because it would be the same as #578.
test_cgroups_race_condition will also test this issue.

Can you please add a comment about this to the pull request text?

Also, can you please attach test logs from running the above PTL test that you mentioned.  Thanks!

minghui-liu:
[Here](https://github.com/PBSPro/pbspro/files/1826291/test_output2.txt) is the PTL test log.

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -3735,6 +3772,65 @@ class CgroupUtils(object):
                 raise
         return self.assigned_resources is not None
 
+    def add_jobid_to_cgroup_jobs(self, jobid):
+        pbs.logmsg(pbs.EVENT_DEBUG4, 'Adding jobid %s to cgroup_jobs' % jobid)
+        try:
+            with open(self.cgroup_jobs_file, 'a') as f:
+                f.write(jobid + ' ')
```
mike0042: I really like they way you read in the job list in remove_jobid_from_cgroup_jobs(), and I think it would make sense to do the same thing here. It's not that expensive since you already seek to the end of the file when you open it.

Alternatively, if f.tell indicates your offset is zero then you know it's the first entry. Then you only need to append jobid (and no space). For the other cases you would append (' ' + jobid).

What if the cgroup_jobs_file doesn't exist? Do you need to open as 'a+'?

minghui-liu: I don't understand. This function only needs to append one jobid so there is no reading involved. 

If the `cgroup_jobs` file does not exist, `open()` will create it. I was wrong about the `+` mode being necessary. I have fixed other places.

mike0042: In the current implementation, there will always be a space following the last jobid in the file. I don't think that will affect the behavior, but it just feels a little sloppy. You could do something like...
'''
with open(self.cgroup_jobs_file, 'a') as f:
    if f.tell() > 0:
        f.write(' ')
    f.write(jobid)
'''

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -3735,6 +3772,65 @@ class CgroupUtils(object):
...
+        except IOError:
+            pbs.logmsg(pbs.EVENT_DEBUG, 'Failed to open cgroup_jobs file.')
```
mike0042: The method will still return true when an IOError is encountered. Is this what you want?

src/hooks/cgroups/pbs_cgroups.PY
minghui-liu: Fixed. Thanks.

src/hooks/cgroups/pbs_cgroups.PY
minghui-liu: No that's not what I wanted. Thanks for catching that. Fixed.

minghui-liu:
I have rebased and squashed. Please take it further. Thanks.

mike0042:
Branch still shows it is out-of-date. Did you force push your rebase?

minghui-liu:
Should be good now.

PR #600, opened 2018-03-20
Automate t1-t6 of mom dynamic resource test

dinesh1306:
#### Bug/feature Description
* Automate test cases t1-t6 from mom dynamic resource test suite.

#### Affected Platform(s)
* All

#### Solution Description
* Added test for mom dynamic resource. 
* t1.host level consumable size resource with mom dynamic value string
* t2.host level float resource with mom dynamic value string
* t3.host level string resource with mom dynamic value integer
* t4.host level string_array resource with mom dynamic value float
* t5.host level string_array resource with mom dynamic value string
* t6.host level string_array resource with mom dynamic value “This is a test”

#### Checklist:

[result.txt](https://github.com/PBSPro/pbspro/files/1829680/result.txt)

### DIFF
--- a/test/tests/functional/pbs_mom_dynamic_resource_ext.py
+++ b/test/tests/functional/pbs_mom_dynamic_resource_ext.py
@@ -1,0 +1,147 @@
+# coding: utf-8
+
+# Copyright (C) 1994-2018 Altair Engineering, Inc.
+# For more information, contact Altair at www.altair.com.
+#
+# This file is part of the PBS Professional ("PBS Pro") software.
+#
+# Open Source License Information:
+#
+# PBS Pro is free software. You can redistribute it and/or modify it under the
+# terms of the GNU Affero General Public License as published by the Free
+# Software Foundation, either version 3 of the License, or (at your option) any
+# later version.
+#
+# PBS Pro is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.
+# See the GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+# Commercial License Information:
+#
+# For a copy of the commercial license terms and conditions,
+# go to: (http://www.pbspro.com/UserArea/agreement.html)
+# or contact the Altair Legal Department.
+#
+# Altair’s dual-license business model allows companies, individuals, and
+# organizations to create proprietary derivative works of PBS Pro and
+# distribute them - whether embedded or bundled with other software -
+# under a commercial license agreement.
+#
+# Use of Altair’s trademarks, including but not limited to "PBS™",
+# "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
+# trademark licensing policies.
+
+from tests.functional import *
+
+
+class TestMomDynRes(TestFunctional):
+
+    def create_mom_resources(self, resc_name,
+                             resc_type, resc_flag,
+                             script_body):
+        """
+        helper function for create mom_resources.
+        """
+        attr = {"type": resc_type, "flag": resc_flag}
+        self.server.manager(MGR_CMD_CREATE, RSC, attr,
+                            id=resc_name, expect=True)
+
+        fp = self.du.create_temp_file(prefix="mom_resc", suffix=".scr",
+                                      body=script_body)
+        self.du.chmod(path=fp, mode=0755)
+
+        mom_config_str = "!" + fp
+        self.mom.add_config({resc_name: mom_config_str})
+        self.scheduler.set_sched_config({'mom_resources': '"foo"'},
+                                        validate=True)
+        self.scheduler.add_resource(resc_name)
+
+    def test_t3(self):
+        """
+        Test for host level string resource
+        with string mom dynamic value
+        """
+        resc_name = "foo"
+        self.create_mom_resources(resc_name, "string", "h", "/bin/echo pqr")
+
+        a = {'scheduling': 'True'}
+        self.server.manager(MGR_CMD_SET, SERVER, a)
+
+        # Submit a job that requests different mom dynamic resource
+        # not return from script
+        attr = {"Resource_List." + resc_name: 'abc'}
+        j = Job(TEST_USER, attrs=attr)
+        jid = self.server.submit(j)
+
+        # The job shouldn't run
+        c = "Can Never Run: Insufficient amount of resource: foo (abc != pqr)"
+        self.server.expect(JOB, {'job_state': 'Q', 'comment': c},
+                           id=jid, attrop=PTL_AND)
+
+    def test_t5(self):
+        """
+        Tets for host level string_array resource
+        with string mom dynamic value
+        """
+
+        resc_name = "foo"
+        self.create_mom_resources(resc_name, "string_array",
+                                  "h", "/bin/echo red,green,blue")
+
+        a = {'scheduling': 'True'}
+        self.server.manager(MGR_CMD_SET, SERVER, a)
+
+        # Submit a job that requests mom dynamic resource
+        attr = {"Resource_List." + resc_name: 'red'}
+        j = Job(TEST_USER, attrs=attr)
+        jid = self.server.submit(j)
+
+        # The job should run
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid)
+
+        # Submit a job that requests different mom dynamic resource
+        # not return from script
+        attr = {"Resource_List." + resc_name: 'white'}
+        j = Job(TEST_USER, attrs=attr)
+        jid = self.server.submit(j)
+
+        # The job shouldn't run
+        c = "Can Never Run: Insufficient amount of resource:"
+        c += " foo (white != red,green,blue)"
+        self.server.expect(JOB, {'job_state': 'Q', 'comment': c},
+                           id=jid, attrop=PTL_AND)
+
+    def test_t6(self):
+        """
+        Test for host level string resource
+        with mom dynamic value “This is a test”
+        """
+        resc_name = "foo"
+        self.create_mom_resources(resc_name, "string",
+                                  "h", "/bin/echo This is a test")
+
+        a = {'scheduling': 'True'}
+        self.server.manager(MGR_CMD_SET, SERVER, a)
+
+        # Submit a job that requests different mom dynamic resource
+        attr = {"Resource_List." + resc_name: 'red'}
+        j = Job(TEST_USER, attrs=attr)
+        jid = self.server.submit(j)
+
+        # The job shouldn't run
+        c = "Can Never Run: Insufficient amount of resource:"
+        c += " foo (red != This is a test)"
+        self.server.expect(JOB, {'job_state': 'Q', 'comment': c},
+                           id=jid, attrop=PTL_AND)
+
+        # Submit a job that requests mom dynamic resource
+        attr = {"Resource_List." + resc_name: '"This is a test"'}
+        j = Job(TEST_USER, attrs=attr)
+        jid = self.server.submit(j)
+
+        # The job should run
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid)


test/tests/functional/pbs_mom_dynamic_resource_ext.py
```diff
@@ -0,0 +1,212 @@
...
+        """
+        helper function for create mom_dyn_res.
+        """
+        attr = {"type": resc_type, "flag": resc_flag}
+        self.server.manager(MGR_CMD_CREATE, RSC, attr,
+                            id=resc_name, expect=True)
+
+        filename = "foobar"
+        tmpfilepath = os.path.join(os.sep, "tmp", filename)
+        with open(tmpfilepath, "w") as fd:
```
anamikau: use mkstemp to create a file or dshutils open function. for system like ICE or CRAY mom is not on the same host.

test/tests/functional/pbs_mom_dynamic_resource_ext.py
```diff
@@ -0,0 +1,212 @@
...
+
+        # So, submit a job that requests the resource
+        attr = {"Resource_List." + resc_name: 'red'}
+        j = Job(TEST_USER, attrs=attr)
+        jid = self.server.submit(j)
+
+        # The job should run
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid)
+
+        # Submit a job that requests mom dynamic resource
```
anamikau: update the comment to reflect job is asking a different resource

dinesh1306:
@anamikau , i attached result with this PR (result.txt). now i am attaching result after addressed your comment.
[result.log](https://github.com/PBSPro/pbspro/files/1837329/result.log)

dinesh1306:
@anamikau , 
I have addressed your comments , please take a look.

dinesh1306:
@anamikau , i update the summary of PR.
Please take a look.

dinesh1306:
@PBSPro/pbspro-maintainers Please review the PR.

test/tests/functional/pbs_mom_dynamic_resource_ext.py
```diff
@@ -0,0 +1,212 @@
...
+
+from tests.functional import *
+
+
+class TestMomDynRes(TestFunctional):
+    def create_mom_dyn_resc(self, resc_name,
+                            resc_type, resc_flag,
+                            script_body):
+        """
+        helper function for create mom_dyn_res.
```
bhroam: I'd remove the underscores.  The real PBS keyword is mom_resources

test/tests/functional/pbs_mom_dynamic_resource_ext.py
```diff
@@ -0,0 +1,212 @@
...
+        # Submit a job that requests mom dynamic resource
+        attr = {"Resource_List." + resc_name: 1}
+        j = Job(TEST_USER, attrs=attr)
+        jid = self.server.submit(j)
+
+        # The job shouldn't run
+        self.server.expect(JOB, {'job_state': 'Q'}, id=jid)
+
+        # Check for the expected log message in scheduler logs
+        c = "Insufficient amount of resource: foo (1kb != (null))"
```
bhroam: Any time you see '(null)' in a string from PBS, then it is a bug.  It means a NULL pointer was passed into printf().  This is not something you should be testing for in a PTL test.

test/tests/functional/pbs_mom_dynamic_resource_ext.py
```diff
@@ -0,0 +1,212 @@
...
+        # Submit a job that requests mom dynamic resource
+        attr = {"Resource_List." + resc_name: 2}
+        j = Job(TEST_USER, attrs=attr)
+        jid = self.server.submit(j)
+
+        # The job shouldn't run
+        self.server.expect(JOB, {'job_state': 'Q'}, id=jid)
+
+        # Check for the expected log message in scheduler logs
+        c = "Insufficient amount of resource: foo (2 != (null))"
```
bhroam: Same as above, this is a bug in PBS

test/tests/functional/pbs_mom_dynamic_resource_ext.py
```diff
@@ -0,0 +1,212 @@
...
+        attr = {"Resource_List." + resc_name: 'abc'}
+        j = Job(TEST_USER, attrs=attr)
+        jid = self.server.submit(j)
+
+        # The job shouldn't run
+        self.server.expect(JOB, {'job_state': 'Q'}, id=jid)
+
+        # Check for the expected log message in scheduler logs
+        c = "Insufficient amount of resource: foo (abc != 3)"
+        self.scheduler.log_match(c)
```
bhroam: Checking the log message is less efficient than the comment message.  Do an expect() on the job comment instead.

test/tests/functional/pbs_mom_dynamic_resource_ext.py
```diff
@@ -0,0 +1,212 @@
...
+
+    def test_t4(self):
+        """
+        Test for host level string_array resource
+        with float mom dynamic value
```
bhroam: I don't think you are not really testing what you think you are testing.  What you are testing is that the scheduler is correctly not matching the string 'red' and the string '3.14'.

test/tests/functional/pbs_mom_dynamic_resource_ext.py
```diff
@@ -0,0 +1,212 @@
...
+        # The job shouldn't run
+        self.server.expect(JOB, {'job_state': 'Q'}, id=jid)
+
+        # Check for the expected log message in scheduler logs
+        c = "Insufficient amount of resource: foo (red != 3.14)"
+        self.scheduler.log_match(c)
+
+    def test_t5(self):
+        """
+        Tets for host level string_array resource
```
bhroam: any particular reason you are using a string_array resource?  You're really treating it as a string.

bhroam:
I don't know if you should spend time writing mom_resources tests.  The functionality has been replaced with a mom periodic hooks.  The option still exists, but could be removed at any time.

anamikau:
Hey @bhroam, you are right but I haven't heard of dropping that feature in 18.x at this point. These tests would still add value in testing 18.1 as there are customers still using dynamic resources. A critical bug was found earlier in server dynamic resource.

bhroam:
I'm incorrect.  I did a search for mom_resources in my PDF viewer and nothing came up.  It appears my PDF viewer is lame.  I tried a different viewer and mom_resources is still in the guides.  Maybe I'm wrong about it being deprecated yet.  It sure should be.  In any case, you are right, these tests should be added.  As a note, server_dyn_res and mom_resources are very different things.  server_dyn_res is heavily used.  mom_resources most likely isn't because of the scalability issues it entails.

dinesh1306:
@bhroam , i have incorporate your comments. Please take a look

test/tests/functional/pbs_mom_dynamic_resource_ext.py
```diff
@@ -60,98 +61,27 @@ def create_mom_dyn_resc(self, resc_name,
...
-        # Submit a job that requests mom dynamic resource
-        attr = {"Resource_List." + resc_name: 1}
-        j = Job(TEST_USER, attrs=attr)
-        jid = self.server.submit(j)
-
-        # The job shouldn't run
-        self.server.expect(JOB, {'job_state': 'Q'}, id=jid)
-
-        # Check for the expected log message in scheduler logs
-        c = "Insufficient amount of resource: foo (1kb != (null))"
```
bhroam: Did you file this bug?  It is a product bug.

test/tests/functional/pbs_mom_dynamic_resource_ext.py
```diff
@@ -181,32 +111,38 @@ def test_t5(self):
...
         # The job shouldn't run
-        self.server.expect(JOB, {'job_state': 'Q'}, id=jid)
-
-        # Check for the expected log message in scheduler logs
-        c = "Insufficient amount of resource: foo (white != red)"
-        self.scheduler.log_match(c)
+        c = "Can Never Run: Insufficient amount of resource:"
+        c += " foo (white != red,green,blue)"
+        self.server.expect(JOB, {'job_state': 'Q', 'comment': c},
+                           id=jid, attrop=PTL_AND)
```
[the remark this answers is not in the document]
dinesh1306: Yes , I have checked this also. I submitted job with value 'red' and checked that job is in running state.Code is present above these lines.

test/tests/functional/pbs_mom_dynamic_resource_ext.py
bhroam: Thanks for filing the bug!

test/tests/functional/pbs_mom_dynamic_resource_ext.py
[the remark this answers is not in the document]
bhroam: Ah OK, it didn't show up because I was looking at a diff.  Thanks for pointing it out.

bhroam:
Looks good.  I sign off.  Once @anamikau takes another look, squash and rebase for the merge.

test/tests/functional/pbs_mom_dynamic_resource_ext.py
anamikau: @bhroam and @dinesh1306 should not we keep the test but comment the log message due to PP-1263? And these tests as reference to PP-1263.

test/tests/functional/pbs_mom_dynamic_resource_ext.py
```diff
@@ -0,0 +1,212 @@
...
+        """
+        attr = {"type": resc_type, "flag": resc_flag}
+        self.server.manager(MGR_CMD_CREATE, RSC, attr,
+                            id=resc_name, expect=True)
+
+        filename = "foobar"
+        tmpfilepath = os.path.join(os.sep, "tmp", filename)
+        with open(tmpfilepath, "w") as fd:
+            fd.write(script_body)
+        self.du.chmod(path=tmpfilepath, mode=0755)
```
bhroam: Now that I'm looking at this again, dshutils now provides the ability to create temp files.  Use self.du.create_temp_file().  You can even provide it the body of your script, so you don't have to do the write.  It'll do everything for you.  It returns the path to the file.

test/tests/functional/pbs_mom_dynamic_resource_ext.py
bhroam: The tests have value as negative test cases.  The problem with including them now is that depending how PP-1263 is fixed, the tests will have to change.  Probably the best way the fix PP-1263 is to reject the resource at the time we query the mom.  The test as written will not work for that.  We might want to wait until PP-1263 is fixed before adding these tests.

anamikau: @bhroam you are right that functionality might change which will change the test but that should not be a reason not to include this test for 18.x. We have found an issue where negative values has caused scheduler to behave differently. At least having this test around for 18.x will help catching such issues (Today it is not crashing it might crash tomorrow with the same test). (PBS behavior might change but objective of the test would remain the same).

bhroam: The test as it stands tests for the buggy behavior.  I do not think we should include that.  If we try and fix the test, we'd need to know what the non-buggy behavior is.  We won't know that until the bug is fixed.

test/tests/functional/pbs_mom_dynamic_resource_ext.py
dinesh1306: @bhroam , i think you are looking into first commit. in second commit i am using "self.du.mkstemp" according to the Anamika's comment. please let me know if "self.du.mkstemp" is ok or i need to change to "self.du.create_temp_file()".

test/tests/functional/pbs_mom_dynamic_resource_ext.py
anamikau: okay thanks

test/tests/functional/pbs_mom_dynamic_resource_ext.py
bhroam: Oh, I must have been looking at the older commit.  I think self.du.create_temp_file() is a better solution because it will do everything you need.  You won't need to open the file and write to the file.  You just pass create_temp_file() the body of the file you want, and it returns the path to the newly created temp file.  It even closes the file descriptor that mkstemp() returns.

dinesh1306:
@bhroam  , i have addressed your comment. please take look.

dinesh1306:
rebased and squashed.Ready for merge

dinesh1306:
[res_pbs_mom_dynamic_resource_ext.txt](https://github.com/PBSPro/pbspro/files/1918991/res_pbs_mom_dynamic_resource_ext.txt)

final execution details

PR #542, opened 2018-01-31
PP-1067: Utilize new topology information to accurately count devices (sockets, MICs, GPUs) on nodes

nithinj:
#### Bug/feature Description
* *The server will need to be able to parse the new topology information gathered from hwloc to determine the number of devices on each node.*
* *PBS server will be able to detect devices such as GPU's and Co-processor(Xeon Phi)*
* *pbs_topologyinfo command can be used on Windows*
* *Existing issue: [PP-1067](https://pbspro.atlassian.net/browse/PP-1067)*

#### Solution Description
* *Hwloc is used to gather details on Linux. On Windows native API's are used to count devices. The mom will find the information and will send it to server with UPDATE2. Server will store it as a file under $PBS_HOME/server_priv/topology/$node_name. pbs_topologyinfo will parse this file to give the requested information.*

#### Checklist:

- [x] I have joined the **[pbspro community forum](http://community.pbspro.org/)**.
- [x] My pull request contains a **single, signed** commit. See **[setting up gpg signature](https://pbspro.atlassian.net/wiki/display/DG/Signing+Your+Git+Commits)**.
- [x] My code follows the **[coding style](https://pbspro.atlassian.net/wiki/display/DG/Coding+Standards)** of this project.
- [ ] My change requires project documentation. See **[required documentation checklist](https://pbspro.atlassian.net/wiki/display/DG/Checklist+for+Developing+Features+and+Bug+Fixes)** for details.
- [ ] I have added documentation in the **[project documentation area](https://pbspro.atlassian.net/wiki/display/PD)**.
- [ ] I have added new **PTL test(s) to my commit**. (See **[using PTL for testing](https://pbspro.atlassian.net/wiki/display/DG/Using+PTL+for+Testing)**) *(or)*
- [x] I have added  **manual test(s) to this pull request and explained why PTL is not appropriate** for this case.
- [x] All new and existing automated tests have passed. (See **[running automated PTL tests](https://pbspro.atlassian.net/wiki/display/DG/PTL+Quick+Start+Guide)**).
- [x] I have attached **test logs to this pull request** as evidence of testing/verification.

__***For further information please visit the [Developer Guide Home](https://pbspro.atlassian.net/wiki/display/DG/Developer+Guide+Home).***__

PR #608, opened 2018-03-23
PP-1067: pbs_topologyinfo enhancements

nithinj:
#### Bug/feature Description
* *enhance pbs_topologyinfo to account new devices discovered*
* *Node displays an extra attribute license='u'*
* *If there is an issue ID, link to it: [PP-1067](https://pbspro.atlassian.net/browse/PP-1067)*

* *Bugs: #542 failed to add necessary changes needed to count the discovered devices*

#### Solution Description
* *This PR includes the necessary pbs_topologyinfo changes to complete the goal of PP-1067*
* Generic check to identify hwloc version. All versions from hwloc 1.11 uses `<Package>` tag instead of `<Socket>`
* PTL test included as smoke test to ensure the correctness of pbsnodes output.
* Clearing additional attribute license='u' from the node attribute list

CC: @bremanandjk , @mike0042 

#### Checklist:

### DIFF
--- a/src/cmds/scripts/pbs_topologyinfo.py
+++ b/src/cmds/scripts/pbs_topologyinfo.py
@@ -1,5 +1,4 @@
 # coding: utf-8
-#
 
 # Copyright (C) 1994-2018 Altair Engineering, Inc.
 # For more information, contact Altair at www.altair.com.
@@ -41,13 +40,28 @@ from optparse import OptionParser
 import os
 import re
 import sys
+import math
 import platform
 
 
 def reportsockets_win(f):
-    temp = f.read().split(',')[0]
-    if temp.find('sockets:') != -1:
-        return int(temp.replace('sockets:', ''))
+    temp = f.read().split(',')
+    for item in temp:
+        if item.find('sockets:') != -1:
+            sockets = int(item[8:])  # len('sockets:') = 8
+        if item.find('gpus:') != -1:
+            gpus = int(item[5:])  # len('gpus:') = 5
+        if item.find('mics:') != -1:
+            coproc = int(item[5:])  # len('mics:') = 5
+    return (sockets, gpus, coproc)
+
+
+def latest_hwloc(hwlocVersion):
+    hwlocVersion = hwlocVersion.split('.')
+    major = int(hwlocVersion[0])
+    minor = int(hwlocVersion[1]) if len(hwlocVersion) > 1 else 0
+    if ((major == 1) and (minor >= 11)) or (major > 1):
+        return 1
     return 0
 
 
@@ -55,39 +69,52 @@ try:
     import xml.parsers.expat
     from xml.parsers.expat import ExpatError
 
-    def reportsockets(dir, files):
+    def reportsockets(dirs, files, options):
         """
-        Look for and report the number of "Socket" elements in a string
-        produced by hwloc_topology_export_xmlbuffer().
+        Look for and report the number of "Package" elements which stands
+        for Sockets in a string produced by hwloc_topology_export_xmlbuffer().
 
         This version of reportsockets uses expat to parse the XML.
         If the PBS version of Python does not allow import of expat,
         we go on to try a simpler approach (below).
         """
+
         def socketXMLstart(name, attrs):
             if name == "BasilResponse":
-                socketXMLstart.isCray = True
+                socketXMLstart.CrayVersion = attrs.get("protocol")
                 return
-            if (name == "info" and attrs.get("name") == "hwlocVersion" and
-                    re.search(r'1.11.*', attrs.get("value"))):
-                socketXMLstart.hwloclatest = 1
+            if (name == "info" and attrs.get("name") == "hwlocVersion"):
+                socketXMLstart.hwloclatest = latest_hwloc(attrs.get("value"))
                 return
-            if socketXMLstart.isCray:
-                if name == "Node":
+            if socketXMLstart.CrayVersion != "0.0":
+                if float(socketXMLstart.CrayVersion) <= 1.2 and name == "Node":
                     socketXMLstart.nsockets += 2
-            else:
-                if (name == "object" and (
-                    (socketXMLstart.hwloclatest == 1 and attrs.get("type") ==
-                        "Package") or (socketXMLstart.hwloclatest == 0 and
-                                       attrs.get("type") == "Socket"))):
+                elif name == "Socket":
                     socketXMLstart.nsockets += 1
+                if name == "Accelerator" and attrs.get("type") == "GPU":
+                    socketXMLstart.ngpus += 1
+            else:
+                if (name == "object" and ((socketXMLstart.hwloclatest == 1 and
+                    attrs.get("type") == "Package") or
+                    (socketXMLstart.hwloclatest == 0 and attrs.get("type") ==
+                        "Socket"))):
+                    socketXMLstart.nsockets += 1
+                if (name == "object" and attrs.get("type") == "OSDev" and
+                    attrs.get("osdev_type") == "1" and
+                        attrs.get("name").startswith("card")):
+                    socketXMLstart.ngpus += 1
+                if (name == "object" and attrs.get("type") == "OSDev" and
+                    attrs.get("osdev_type") == "5" and
+                        attrs.get("name").startswith("mic")):
+                    socketXMLstart.ncoproc += 1
 
         if files is None:
             compute_socket_nodelist = True
             try:
-                files = os.listdir(dir)
-            except (IOError, OSError), (e, strerror):
-                print "%s:  %s (%s)" % (dir, strerror, e)
+                files = os.listdir(dirs)
+            except (IOError, OSError) as err:
+                (e, strerror) = err.args
+                print "%s:  %s (%s)" % (dirs, strerror, e)
                 return
         else:
             compute_socket_nodelist = False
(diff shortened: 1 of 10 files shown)


nithinj:
The newly introduced test case `test_validate_pbsnodes_attrs`  fails to run in smoke mode, but passes when ran as a single test. I'm debugging it.

nithinj:
I've updated the test case. Please review.

test/tests/functional/pbs_nodes_json.py
```diff
@@ -91,3 +91,33 @@ def test_check_tab_escape_json(self):
             json.loads("\n".join(n_out))
         except ValueError:
             self.assertFalse(True, "Json failed to load")
+
+    @tags('smoke')
+    def test_validate_pbsnodes_attrs(self):
+        """
+        Test if the pbsnodes outputs all the default attributes.
+        And does not print any new attributes.
```
bhroam: You shouldn't check for an exact set of attributes.  In the future if we  add a new default attribute, your test will fail.  Since you are tagging this test as smoke, it will be run by travis.  This will hold up any subsequent PR until we fix it.  Fixing it requires Subhasis to jump through hoops.

src/server/req_stat.c
```diff
@@ -901,8 +901,13 @@ print_license_ct(enum licensing_backend lb, char *buf)
...
 				sockets_total(), sockets_available());
 			break;
+		case LIC_TRIAL:
+			sprintf(buf, "Avail_Local:%d Used:%d High_Use:%d",
+				licenses.lb_aval_floating, licenses.lb_used_floating,
+				licenses.lb_high_used_floating);
+			break;
 		default:
-			sprintf(buf, "NA");
+			sprintf(buf, "Avail_Nodes:%d Unused_Nodes:%d", 0, 0);
```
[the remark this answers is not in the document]
nithinj: "NA" was never used in PTL tests. FW has updated in function `is_server_licensed`, file: https://github.com/PBSPro/pbspro/blob/master/test/fw/ptl/utils/pbs_testsuite.py

test/tests/functional/pbs_nodes_json.py
nithinj: The intention is to check for the default set of attributes of pbsnodes output. Nobody should add/remove an attribute to/from pbsnodes output mistakenly. If they do, this test case will trigger a red flag on travis and the PR will be blocked. If the change is intentional, the smoke test also should get updated along with the PR.

test/tests/functional/pbs_nodes_json.py
```diff
@@ -91,3 +91,33 @@ def test_check_tab_escape_json(self):
...
+        """
+        cmd = [os.path.join(self.server.pbs_conf['PBS_EXEC'],
+                            'bin', 'pbsnodes')]
+        cmd += ['-v', self.server.shortname, '-Fjson']
+        n_out = self.du.run_cmd(self.server.hostname, cmd=cmd)['out']
+        try:
+            json_object = json.loads("\n".join(n_out))
+        except ValueError:
+            self.assertTrue(False)
```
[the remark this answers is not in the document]
nithinj: This test is not checking what type of resource the nodes displaying such as mem,vmem, ncpus...
It just checks whether resource_assigned and resource_available is present. For this, json parsing is convenient as it lists all the attributes in the second level and type of resources only in the next level.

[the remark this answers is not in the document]
bhroam: Whenever possible, use PTL's provided functionality.  Only use run_cmd when it is not.  In this case status() can be used to get the attributes of a node.

test/tests/functional/pbs_nodes_json.py
```diff
@@ -91,3 +91,35 @@ def test_check_tab_escape_json(self):
...
+            'Mom', 'Port', 'pbs_version', 'ntype', 'state', 'pcpus',
+            'resources_available', 'resources_assigned', 'resv_enable',
+            'sharing']
+        node_attr = []
+        self.server.expect(NODE, {'state': 'free'},
+                           id=self.server.shortname, interval=3)
+        for key, val in \
+                json_object['nodes'][self.server.shortname].iteritems():
+            if str(key) not in attrs_pbsnodes_dflts:
+                self.assertFalse(not (key == 'license' and val == 'l'),
```
bhroam: You're testing for three things here.  You want to see the existence of certain attributes and not others except sometimes this one attribute.  I'd make this third case more generic.  There might be more in this third case in the future.  Right now it'll be difficult to add those to the test.  If you make it more generic, then it'll be as easy as adding an attribute to a list.  Make a list of possible default attributes.  Those can appear or not.

src/cmds/scripts/pbs_topologyinfo.py
```diff
@@ -48,20 +48,21 @@ def reportsockets_win(f):
     temp = f.read().split(',')
     for item in temp:
         if item.find('sockets:') != -1:
-            sockets = int(item.replace('sockets:', ''))
+            sockets = int(item[len('sockets:'):])
```
bhroam: I find this hard to read.  I'd rather you add a comment saying 'len('sockets:') = 8 and make the splice of item[8:] (and for the subsequent two splices).

src/cmds/scripts/pbs_topologyinfo.py
```diff
@@ -48,20 +48,21 @@ def reportsockets_win(f):
...
 
 
 def latest_hwloc(hwlocVersion):
-    if hwlocVersion.count('.') > 1:
-        hwlocVersion = hwlocVersion[:hwlocVersion.rfind('.')]
-    hwlocVersion = float(hwlocVersion)
-    if hwlocVersion >= 1.11:
+    hwlocVersion = hwlocVersion.split('.')
+    major = int(hwlocVersion[0])
+    minor = int(hwlocVersion[1])
```
bhroam: Can you change this to: minor = int(hwlocVersion[1]) if len(hwlocVersion) > 1 else 0

test/tests/functional/pbs_nodes_json.py
bhroam: If that is what you want to do, then I'm OK with it.  I'm not sure the added maintenance of checking for things that shouldn't show up is warranted, but I see the point of it.

lisa-altair: @arungrover and I also don't understand the benefit of adding the smoke test.
We're not sure we understand how someone "mistakenly" adds/removes an attribute.  Can you please explain more?  It may help us to understand why such a smoke test is needed.
Because in my opinion if someone does mistakenly add/remove and attribute and the smoke test fails then the developer will see the red flag and change the smoke test.  No one else will necessarily be aware of any mistaken add/remove because the tests will now pass.
It only seems to add maintenance as @bhroam already mentioned.

test/tests/functional/pbs_nodes_json.py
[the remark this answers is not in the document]
nithinj: With that, the test does not fit in `TestPbsnodes_json` class anymore. I'll create a new one for it. Hope you are fine with it.

test/tests/functional/pbs_nodes_json.py
nithinj: It is helpful in catching bugs such as additional pbsnodes attribute `license='u'` which had addressed with this PR. In scenarios like these, this test will fail and notifies the developer that there is something strange. The test description explains the motives behind it. Even if the developer doesn't observe it and updates the test case to clear the red flag, the maintainer can raise the concern on why a smoke test got edited without appropriate explanation/design document.

This test also doesn't add any bottleneck to travis CI as it does not submit a new job, and will get completed quickly. I cannot add this test case without a smoke tag as it will not help in achieving the objective. I would say the maintenance cost of this test is not significant as the features which modifies the node attributes are not frequent and that can be justified by preventing possible bugs.

test/tests/functional/pbs_nodes_json.py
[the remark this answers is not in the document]
bhroam: If you want to make it a smoke test, why not just add it to pbs_smoketest.py?

nithinj:
Hi @bhroam .
I've moved the test case to pbs_smoketest.py. Please check.

test/tests/pbs_smoketest.py
```diff
@@ -1505,3 +1505,31 @@ def test_man_pages(self):
             msg = "Successfully retrieved man page for"
             msg += " %s command" % pbs_cmd
             self.logger.info(msg)
+
+    @tags('smoke')
```
anamikau: There is a class level smoke tag for this class. You do not need to tag this test as smoke explicitly.

test/tests/functional/pbs_nodes_json.py
lisa-altair: Thank you for the explanation @nithinj.  I suppose it's fine.   We can always remove it later if it turns out to hurt more than help. Thanks again.

bhroam:
Looks good to me.  I sign off.  Please squash and rebase for the merge.

nithinj:
Thanks @bhroam and @bremanandjk for your sign-off. I've squashed and re-based my branch. Please take this further.

bhroam:
I didn't see evidence of testing.  Could you add it?

nithinj:
[PTL result.txt](https://github.com/PBSPro/pbspro/files/1907339/PTL.result.txt)
`[root@centos cdrom]# pbs_topologyinfo -al`
`centos 1
`

nithinj:
Hi @bhroam ,
I've added my verification results. Please check.

bhroam:
A quick (non-code) comment.  The 3rd bullet in your solution description is cut off.  Could you complete the sentence?

nithinj:
Hi @bhroam ,
Thanks for the catch, I've updated the solution description. The branch is up-to date please have a look.

bhroam:
I merged in the PR.

PR #624, opened 2018-04-07
memory leak fixes in server

arungrover:
#### Bug/feature Description
Memory leaks found in server under following scenarios:
* When one tries to set/unset a node attribute that is not allowed to be unset or has no permission to be unset (like resources_assigned)
* When a queue has default_chunk set on it and then the queue is deleted
* When there are jobs running on an execution host and mom is started with -p option to preserve the running jobs.

#### Affected Platform(s)
* all

These leaks were trivial and straightforward.
here are the valgrind snippets of the leaks in the order they have been listed in previous section - 

==113219== 8 bytes in 1 blocks are definitely lost in loss record 28 of 1,733
==113219==    at 0x4C29BE3: malloc (vg_replace_malloc.c:299)
==113219==    by 0x45DEB5: mgr_node_unset (req_manager.c:2541)
==113219==    by 0x45FE77: req_manager (req_manager.c:4861)
==113219==    by 0x4A1782: wait_request (net_server.c:568)
==113219==    by 0x428BD0: main (pbsd_main.c:2140)

==113219== 40 (32 direct, 8 indirect) bytes in 1 blocks are definitely lost in loss record 122 of 1,733
==113219==    at 0x4C29BE3: malloc (vg_replace_malloc.c:299)
==113219==    by 0x476116: deflt_chunk_action (svr_func.c:1041)
==113219==    by 0x45A379: mgr_set_attr2 (req_manager.c:943)
==113219==    by 0x45D4A4: mgr_set_attr (req_manager.c:1037)
==113219==    by 0x45D4A4: mgr_queue_set (req_manager.c:1821)
==113219==    by 0x45FEF7: req_manager (req_manager.c:4816)
==113219==    by 0x4A1782: wait_request (net_server.c:568)
==113219==    by 0x428BD0: main (pbsd_main.c:2140)

==113219== 2 bytes in 2 blocks are definitely lost in loss record 4 of 1,733
==113219==    at 0x4C29BE3: malloc (vg_replace_malloc.c:299)
==113219==    by 0x4CE7D5: disrst (disrst.c:99)
==113219==    by 0x4482E2: mom_running_jobs (node_manager.c:4324)
==113219==    by 0x44D38C: is_request (node_manager.c:4662)
==113219==    by 0x452218: do_rpp (pbsd_main.c:466)
==113219==    by 0x45223C: rpp_request (pbsd_main.c:515)
==113219==    by 0x4A1782: wait_request (net_server.c:568)
==113219==    by 0x428BD0: main (pbsd_main.c:2140)

#### Solution Description
* Added code to free up memory when it is not needed anymore.
* I've also added code to suppress a leak that valgrind reports when hooks are used but it is not actually a leak because it gets freed when python interpreter restarts

#### Checklist:

* Valgrind Logs

[server_log_withoutleak.txt](https://github.com/PBSPro/pbspro/files/1885960/server_log_withoutleak.txt)
[server_log_withleak.txt](https://github.com/PBSPro/pbspro/files/1885961/server_log_withleak.txt)

### DIFF
--- a/src/server/queue_func.c
+++ b/src/server/queue_func.c
@@ -150,6 +150,7 @@ que_free(pbs_queue *pq)
 	int		 i;
 	attribute	*pattr;
 	attribute_def	*pdef;
+	key_value_pair  *pkvp = NULL;
 
 	/* remove any malloc working attribute space */
 
@@ -158,6 +159,15 @@ que_free(pbs_queue *pq)
 		pattr = &pq->qu_attr[i];
 
 		pdef->at_free(pattr);
+	}
+	/* free default chunks set on queue */
+	pkvp = pq->qu_seldft;
+	if (pkvp) {
+		for (i = 0; i < pq->qu_nseldft; ++i) {
+			free((pkvp+i)->kv_keyw);
+			free((pkvp+i)->kv_val);
+		}
+		free(pkvp);
 	}
 
 	/* now free the main structure */
--- a/src/server/req_manager.c
+++ b/src/server/req_manager.c
@@ -2297,8 +2297,10 @@ mgr_node_set(struct batch_request *preq)
 							reply_text(preq, PBSE_PARTITION_NOT_IN_QUE, log_buffer);
 							break;
 
-						default:  req_reject(rc, 0, preq);
+						default:
+							req_reject(rc, 0, preq);
 					}
+					free(warn_nodes);
 					return;
 				}
 
@@ -2587,8 +2589,10 @@ mgr_node_unset(struct batch_request *preq)
 							reply_badattr(rc, bad, plist, preq);
 							break;
 
-						default:  req_reject(rc, 0, preq);
+						default:
+							req_reject(rc, 0, preq);
 					}
+					free(warn_nodes);
 					return;
 				}
 
--- a/valgrind.supp
+++ b/valgrind.supp
@@ -634,4 +634,11 @@
    fun:log_record
    fun:*
 }
-
+{
+   From PBSPro (server) - Suppress allocations that clear up when python interpreter restarts
+   Memcheck:Leak
+   fun:malloc
+   fun:PyString_FromString
+   fun:pbs_python_populate_attributes_to_python_class
+   fun:*
+}
--- a/src/server/node_manager.c
+++ b/src/server/node_manager.c
@@ -4293,6 +4293,7 @@ mom_running_jobs(int stream)
 	char             exec_host_name[PBS_MAXHOSTNAME+2]="UNKNOWN2";
 	char             *slash_pos = NULL;
 	int              exec_host_hostlen = 0;
+	char		*pset = NULL;
 
 	njobs = disrui(stream, &rc);    /* number of jobs in update */
 	if (rc)
@@ -4321,7 +4322,7 @@ mom_running_jobs(int stream)
 		execvnod = disrst(stream, &rc);
 		if (rc)
 			goto err;
-		(void)disrst(stream, &rc);	/* pset is not currently used */
+		pset = disrst(stream, &rc);	/* pset is not currently used */
 		if (rc)
 			goto err;
 
@@ -4418,6 +4419,8 @@ mom_running_jobs(int stream)
 		jobid = NULL;
 		free(execvnod);
 		execvnod = NULL;
+		free(pset);
+		pset = NULL;
 	}
 	return;
 
@@ -4427,6 +4430,7 @@ err:
 	log_err(errno, "mom_running_jobs", log_buffer);
 	free(jobid);
 	free(execvnod);
+	free(pset);
 }
 
 


valgrind.supp
```diff
@@ -634,4 +634,11 @@
    fun:log_record
    fun:*
 }
-
+{
+   From PBSPro (server) - Suppress allocations that clear up when python interpreter restarts
+   Memcheck:Leak
+   fun:malloc
+   fun:PyString_FromString
```
hirenvadalia: @arungrover I don't think this is good to add in suppression file, because as per document of [PyString_FromString](https://docs.python.org/2/c-api/string.html#c.PyString_FromString) it returns new reference which we need to pass to Py_XDECREF (or Py_CLEAR) once we are done with it.

src/server/queue_func.c
```diff
@@ -159,6 +160,15 @@ que_free(pbs_queue *pq)
 
 		pdef->at_free(pattr);
 	}
+	/* free default chunks set on queue */
+	pkvp = pq->qu_seldft;
+	if (pkvp) {
+		for (i=0; i < pq->qu_nseldft; ++i) {
+			free((pkvp+i)->kv_keyw);
```
hirenvadalia: should we first check (pkvp+i) for not null?

valgrind.supp
arungrover: @hirenvadalia thanks for reviewing the pull, You are right that PyString_FromString will not be freed unless its reference count comes to 0 but it in this suppression file I've mentioned the function stack to include "pbs_python_populate_attributes_to_python_class" which I think will only be used while preparing input for the hook and will get freed when interpreter restarts.

Have you seen this sort of leak even after interpreter restarts?

src/server/queue_func.c
arungrover: qu_nseldft is actually a count of how many select defaults are there in qu_seldft. This gets set in "deflt_chunk_action()" function. Adding an additional check is not a bad thing to have but practically these two numbers have to be same so the memory must exist (unless there is a bug in deflt_chunk_action function). 
Similar code exists in deflt_chunk_action function as well. I would like to leave it as is unless you feel strongly about adding a null check.

bhroam: POSIX defines free(NULL) as a noop.  There's no real need for the check.

src/server/queue_func.c
```diff
@@ -159,6 +160,15 @@ que_free(pbs_queue *pq)
 
 		pdef->at_free(pattr);
 	}
+	/* free default chunks set on queue */
+	pkvp = pq->qu_seldft;
+	if (pkvp) {
+		for (i=0; i < pq->qu_nseldft; ++i) {
```
bhroam: Add spaces around the '='

src/server/req_manager.c
```diff
@@ -2589,6 +2589,8 @@ mgr_node_unset(struct batch_request *preq)
 
 						default:  req_reject(rc, 0, preq);
```
bhroam: I know this isn't your code, but could you put a newline in between the default and the reject?

src/server/queue_func.c
arungrover: I agree with you @bhroam, I think Hiren's concern about checking for null was that we don't dereference it when it is null.

src/server/req_manager.c
arungrover: Done

src/server/req_manager.c
```diff
@@ -2587,8 +2587,11 @@ mgr_node_unset(struct batch_request *preq)
 							reply_badattr(rc, bad, plist, preq);
 							break;
 
-						default:  req_reject(rc, 0, preq);
+						default:
+							req_reject(rc, 0, preq);
 					}
+					free(warn_nodes);
+					free(problem_nodes);
```
lisa-altair: I don't think we *need* this free(problem_nodes) here.  Because earlier on line 2531 we only allocate memory if numnodes > 1.  And here at line 2594 we are in the else case of numnodes > 1.
But here's a question...I also noticed that warn_nodes is allocated always...  In case someone decides to make problem_nodes look like warn_nodes and allocate always, should we leave the free(problem_nodes) there just in case?  Because it doesn't do any harm (except for the extra call to free with a NULL).  And it could help prevent leaks in the future.  What do you think?

lisa-altair:
Also, don't we have the same leak for warn_nodes in mgr_node_set around line 2302 when we return without first freeing warn_nodes?

src/server/req_manager.c
arungrover: You are right Lisa. While this free will have no effect when problem_nodes is not allocated but I think it is not suitable to have it here for now. I think I should remove it from the code.

arungrover:
@lisa-altair good find! I'll address that leak as well.

valgrind.supp
hirenvadalia: You are tight @arungrover that suppression you added will be skip leaks if call stack has "pbs_python_populate_attributes_to_python_class".

But "pbs_python_populate_attributes_to_python_class" is PBS' function and PBS prepares input for hook which means it's PBS' responsibility to free that input because PBS is owner of that reference (see first paragraph at [here](https://docs.python.org/2/c-api/intro.html#reference-count-details))

Regarding restart of interpreter:
As you have seen in above link that it is not good idea to restart interpreter (aka call to Py_Initilize and Py_Finalize) multiple time in process as it doesn't guaranty to free all objects which PBS' has allocated, but that is different story...

Answer to you question: Yes you can see that those reference (aka objects) are still live after interpreter restart, but to see that you have to do little hack in PBS' code as below:
Replace  https://github.com/PBSPro/pbspro/blob/master/src/lib/Libpython/pbs_python_external.c#L177
with below:
```
Py_IgnoreEnvironmentFlag = 0;
setenv("PYTHONDUMPREFS", "1", 1);
```
then run your test manually while PBS Server is running in console (with -N) as Python debug logs will not go into PBS' log file.

FYI: There are other variables also if you want to get more debugging info, see [here](https://docs.python.org/2/using/cmdline.html#debug-mode-variables)

src/server/queue_func.c
hirenvadalia: Thanks @arungrover for explanation. I think we don't need null check for (pkvp+i) as you said it much be valid memory as per "deflt_chunk_action()" function.

@bhroam Indeed free(NULL) is noop, but we were taking about check for "(pkvp+i)" not "(pkvp+i)->kv_keyw" as we are checking for "pkvp" as like 165 but we are not checking "(pkvp+i)" anywhere. Hope you understand 😄

arungrover:
Thanks for useful pointers Hiren. I'll look into them and try testing with the option you mentioned

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

valgrind.supp
hirenvadalia: @arungrover Based on our chat and testing this part, I think for now it is good to add this leak in suppression file till we fix (and remove) restarting of python interpreter so that we get clean valgrind report.

bhroam:
I'm happy with the changes.  Please squash and rebase for the merge.

bhroam:
I merged in the PR

PR #628, opened 2018-04-10
mail fork error is not logged

vchlum:
#### Bug/feature Description
* There is no error message logged on mail fork fail. If the failure happens it is difficult to determine why the mail wasn't sent.
* Steps to reproduce: Let the server occupy so much memory that the fork() cannot be done successfully and run a job with '-m b'.

#### Affected Platform(s)
* Linux

* The fork fail is already handled but no error message is logged.

#### Solution Description
* I added logging error message on fork fail. This error message is consistent with the others fork fail messages in the code.

* I also fixed the align of the sendmail code.

* test log: [mail_fork-ptl_output_smoketest.txt](https://github.com/PBSPro/pbspro/files/1893872/mail_fork-ptl_output_smoketest.txt)

#### Checklist:

### DIFF
--- a/src/server/svr_mail.c
+++ b/src/server/svr_mail.c
@@ -639,7 +639,12 @@ svr_mailowner_id(char *jid, job *pjob, int mailpoint, int force, char *text)
 	 */
 
 #ifndef WIN32
-	if (fork())
+	mcpid = fork();
+	if (mcpid == -1) { /* Error on fork */
+		log_err(errno, __func__, "fork failed\n");
+		return;
+	}
+	if (mcpid > 0)
 		return;		/* its all up to the child now */
 
 	/*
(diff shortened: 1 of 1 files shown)


src/server/svr_mail.c
```diff
@@ -752,16 +757,21 @@ svr_mailowner_id(char *jid, job *pjob, int mailpoint, int force, char *text)
...
+		}
+		(void)close(1);
+		(void)close(2);
+		if (execv(SENDMAIL_CMD, margs) == -1)
 			exit(1);
 	}
+	if (mcpid == -1) {/* Error on fork */
+		log_err(errno, __func__, "fork failed\n");
+		(void)close(mfds[0]);
+		exit(0);
```
bhroam: In the previous check for fork failure, you return out of the function.  Here you exit().  Why the difference?  Do we really want to exit the server on a fork() failure?  I usually try avoiding such measures for daemons.

src/server/svr_mail.c
```diff
@@ -995,16 +1010,21 @@ svr_mailownerResv(resc_resv *presv, int mailpoint, int force, char *text)
...
```
bhroam: Same question as above about exit() vs return

src/server/svr_mail.c
vchlum: This is the second fork in the row. If we fail the first fork we are still the main process - so the return is necessary, but after the first fork, the process only intends to send the e-mail, which means the exit is suitable after the second fork. 

Now, I see that it should be exit(1) instead of exit(0).

bhroam: Oh, I see.  This fork() is in the child, so exiting out of the child is the right thing to do.  Yeah, I agree that exit(1) is better than 0.

vchlum: Changed to exit(1).

vchlum:
Thank you for reviews. I have squashed and rebased. This PR is ready for merge.

bhroam:
I merged the PR.

PR #630, opened 2018-04-10
Scheduler reports preempted job as Can Never Run for one cycle

agrawalravi90:
#### Bug/feature Description
* For jobs preempted by suspension or checkpointing in the cycle that they are preempted, the scheduler will report them as "Job will never run". The preempted jobs do get run in the next cycle, but sched_logs end up with the never run log message, which can be confusing.

#### Affected Platform(s)
* All

#### Cause / Analysis
Bhroam helped me debug this. The analysis as as follows:
1) The resources that the scheduler uses for scheduling are the intersection between the sched_config resources line (stored in policy->resdef_to_check) and the resources requested by jobs/resvs in the cycle (stored in the job's select spec)
2) If a job is preempted via suspension or checkpointing, the job should be run on the same vnodes as before. To make this happen, the scheduler adds information about the vnodes from the job's exec_vnode to its select spec so that it seems like the job is requesting to be placed on the vnodes that it ran on previously.
3) Now, the 'vnode' resource needs to be added to the resources that the scheduler uses for scheduling (ref bullet 1). This happens conditionally inside update_universe_on_end(), the condition being that the job's select spec contains a resource that policy->resdef_to_check doesn't.
4) This condition doesn't gets fulfilled in the cycle that the job was preempted because, and this is the code that's buggy, 'vnode' already got added to policy->resdef_to_check for the original policy object when the scheduler was preempting the job. This happens because the original policy object is modified while simulating job end during preemption. So, the resource 'vnode' never gets added the the list of resources that the scheduler uses and while trying to schedule the preempted job, the scheduler sees that the job is requesting 'vnode', but 'vnode' doesn't seem to be a resource available in the system, and it ends up marks the job as Can Never Run.
5) The problem goes away in the next cycle as it recreates the resource list at the beginning of the cycle unconditionally while querying the universe, but the sched logs already got populated in the previous cycle with a log message saying "Job will never run".

#### Solution Description
* Despite the esoteric nature of the bug, the fix is quite simple: we just need to use the dup'd policy object while simulating job ends during preemption so that the original policy object is not touched. This will make update_universe_in_end() update the scheduler's list of resources to use with 'vnode', and the job will not be marked as Can Never run, it will instead correctly log the insufficient resources message.

#### Testing:
[after_fix.log](https://github.com/PBSPro/pbspro/files/1900368/after_fix.log)
[before_fix.log](https://github.com/PBSPro/pbspro/files/1896688/before_fix.log)

#### Checklist:

### DIFF
--- a/src/scheduler/job_info.c
+++ b/src/scheduler/job_info.c
@@ -3027,6 +3027,7 @@ find_jobs_to_preempt(status *policy, resource_resv *hjob, server_info *sinfo, in
 	int has_lower_jobs = 0;	/* there are jobs of a lower preempt priority */
 	int prev_prio;		/* jinfo's preempt field before simulation */
 	server_info *nsinfo;
+	status *npolicy;
 	resource_resv **rjobs = NULL;	/* the running jobs to choose from */
 	resource_resv **pjobs = NULL;	/* jobs to preempt */
 	int *pjobs_list = NULL;	/* list of job ids */
@@ -3166,12 +3167,14 @@ find_jobs_to_preempt(status *policy, resource_resv *hjob, server_info *sinfo, in
 		}
 	}
 
+	/* use locally dup'd copy of sinfo so we don't modify the original */
 	if ((nsinfo = dup_server_info(sinfo)) == NULL) {
 		free_schd_error_list(full_err);
 		free(pjobs);
 		free_string_array(preempt_targets_list);
 		return NULL;
 	}
+	npolicy = nsinfo->policy;
 
 	njob = find_resource_resv_by_rank(nsinfo->jobs, hjob->rank);
 	prev_prio = njob->job->preempt;
@@ -3237,7 +3240,7 @@ find_jobs_to_preempt(status *policy, resource_resv *hjob, server_info *sinfo, in
 	}
 
 	skipto=0;
-	while ((indexfound = select_index_to_preempt(policy, njob, rjobs, skipto, err, fail_list)) != NO_JOB_FOUND) {
+	while ((indexfound = select_index_to_preempt(npolicy, njob, rjobs, skipto, err, fail_list)) != NO_JOB_FOUND) {
 		if (indexfound == ERR_IN_SELECT) {
 			/* System error occurred, no need to proceed */
 			free_server(nsinfo, 1);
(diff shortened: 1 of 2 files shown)


src/scheduler/job_info.c
```diff
@@ -3057,26 +3058,37 @@ find_jobs_to_preempt(status *policy, resource_resv *hjob, server_info *sinfo, in
 	if (hjob == NULL || sinfo == NULL)
 		return NULL;
 
+	/* use locally dup'd copy of sinfo so we don't modify the original */
+	if ((nsinfo = dup_server_info(sinfo)) == NULL) {
+		free_schd_error_list(full_err);
+		free(pjobs);
+		free_string_array(preempt_targets_list);
+		return NULL;
+	}
```
bhroam: By moving this call where you did, you'll now dup the server for every job we don't run.  The server was dup'd where it was for a reason.  Duping the server is expensive.  We want to dup it as close to where it is used as possible.  That will give the rest of the function every possibility to return before we dup.

src/scheduler/job_info.c
```diff
@@ -3026,7 +3026,8 @@ find_jobs_to_preempt(status *policy, resource_resv *hjob, server_info *sinfo, in
 	int j = 0;
 	int has_lower_jobs = 0;	/* there are jobs of a lower preempt priority */
 	int prev_prio;		/* jinfo's preempt field before simulation */
-	server_info *nsinfo;
+	server_info *nsinfo = NULL;
+	status *npolicy = NULL;
```
bhroam: No need to initialize these.  They get set before they are used.

agrawalravi90:
Thanks for catching that Bhroam. I'd moved it up to make the code cleaner and less error prone, but I wasn't aware of the possible performance hit, which is unacceptable. I tried to divide this up into 2 functions, but it was going to lead to a large list of variables on the second function's stack, so I didn't do it. So, I just moved the dup back to where it was originally. Please let me know if it looks better now. Thanks again.

src/scheduler/job_info.c
agrawalravi90: All of the local pointers besides nsinfo in the function seem to be initialized, I checked a few of them and they seem to be getting set before use as well. Do you still want me to revert it?

bhroam: Please remove the initialization.  Feel free to fix other pointers that get set before they are used as well.  Now that we have gcc -Wall -Werror set, the compiler will bail if we use an uninitialized variable.  If we always initialize them all, gcc won't alert us and we'll have a bug.

agrawalravi90: I see, alright I'll remove them, thanks.

test/tests/functional/pbs_preemption.py
```diff
@@ -0,0 +1,87 @@
...
+        attr["queue"] = queue_id_h
+        j = Job(TEST_USER, attrs=attr)
+        jidh = self.server.submit(j)
+        self.server.expect(JOB, {ATTR_state: 'R'}, id=jidh)
+
+        # The low priority job should be preempted
+        self.server.expect(JOB, {ATTR_state: 'S'}, id=jidl)
+
+        # Check whether scheduler marked the preempted job as "will never run"
+        self.scheduler.log_match(jidl + ";Job will never run", existence=False)
```
bhroam: I'm put a max_attempts in this log_match().  If you don't, it will need to check 60 times before it decides the match didn't happen.

agrawalravi90:
Thanks Bhroam. One more review please!

arungrover:
@agrawalravi90 please rebase and squash

agrawalravi90:
@arungrover Thanks for the sign-off, I squashed, rebased and attached the latest test logs to the description, I'll let you know once the CI passes.

agrawalravi90:
@PBSPro/pbspro-maintainers this should be ready for merge, thanks!

bhroam:
I merged in the PR.

PR #634, opened 2018-04-12
PP-1261: Retain previous scheduling cycle's configuration of Multisched as much as necessary

suresh-thelkar:
#### Bug/feature Description
This ticket is one of the major user-stories of the following EPIC.
 
[PP-337](https://pbspro.atlassian.net/browse/PP-337) Multiple schedulers servicing the PBS cluster

Currently at the end of each scheduler_iteration or intermediate scheduling cycle Multisched validates its configuration assuming that admins can change its configuration at any point in time i.e. during the intermediate scheduling cycles or before submitting/deleting a job etc. The problem with this approach is we will end up in validating the necessary Multisched configuration for each iteration of the events as described above even though there is no change in its configuration data. The current user-story helps in avoiding this.

#### Affected Platform(s)
* All

#### Cause / Analysis / Design/Solution
Now Multisched maintains some sort of its previous cycle's configuration. This certainly helps in reducing the number of times we perform/validate configuration. We need to do validation only when the configuration is different from its previous cycle. This way Multisched performance is also improved to a better extent.

In addition to the above to achieve this goal we have to separate Multisched configuration change/applying it from the attribute scheduling i.e. configuration change should be applied in real time(immediately when admins change it irrespective of scheduling=true or not).
If scheduling=false then Multisched should apply its configuration but should not trigger scheduling cycle.

I have also fixed the following bug "multi-sched scheduler goes to scheduling=false intermittently" as part of this pull request. Root cause of this bug is as follows.

The system call “open” is failing while opening/writing one of the scheduler’s config files and hence Multisched is thinking that scheduler validation is failed and hence setting its comment to “PBS failed validation checks for sched_priv directory”. And as per the design once validation is failed we set its scheduling to false automatically. “open” system call is actually failing because the number of file descriptors per process are exhausted(which is 1024 in my system). After this no matter even we set scheduling to true any number of times “open” system call fails since it crossed the soft limit of number of fds open which is 1024.

solution to this is make sure that the closing of file happens correctly. Added this as a separate commit for clarity.
Also written all the necessary PTL test cases. Please find the logs given below.

[TestMultipleSchedulers_04122018.log](https://github.com/PBSPro/pbspro/files/1902543/TestMultipleSchedulers_04122018.log)
[TestSchedulerInterface_04122018.log](https://github.com/PBSPro/pbspro/files/1902545/TestSchedulerInterface_04122018.log)

**Latest PTL logs**
[TestMultipleSchedulers_04182018.log](https://github.com/PBSPro/pbspro/files/1927678/TestMultipleSchedulers_04182018.log)
[TestSchedulerInterface_04182018.log](https://github.com/PBSPro/pbspro/files/1927679/TestSchedulerInterface_04182018.log)

#### Checklist:

### DIFF
--- a/src/scheduler/fifo.c
+++ b/src/scheduler/fifo.c
@@ -603,12 +603,6 @@ intermediate_schedule(int sd, char *jobid)
 	int ret; /* to re schedule or not */
 	int cycle_cnt = 0; /* count of cycles run */
 
-	/*
-	 * This is required since there is a probability that scheduler's configuration has been changed at
-	 * server through qmgr.
-	 */
-	update_svr_schedobj(connector, 0, 0);
-
 	do {
 		ret = scheduling_cycle(sd, jobid);
 
@@ -2375,12 +2369,12 @@ sched_settings_frm_svr(struct batch_status *status)
 	struct attrl *attr;
 	char *tmp_priv_dir = NULL;
 	char *tmp_log_dir = NULL;
-	char *tmp_partitions = NULL;
+	static char *priv_dir = NULL;
+	static char *log_dir = NULL;
 	struct	attropl	*attribs;
 	char *tmp_comment = NULL;
 	int clear_comment = 0;
 	int ret = 0;
-	extern char *partitions;
 
 	attr = status->attribs;
 
@@ -2393,9 +2387,6 @@ sched_settings_frm_svr(struct batch_status *status)
 			} else if (!strcmp(attr->name, ATTR_sched_log)) {
 				if ((tmp_log_dir = string_dup(attr->value)) == NULL)
 					goto cleanup;
-			} else if (!strcmp(attr->name, ATTR_partition)) {
-				if ((tmp_partitions = string_dup(attr->value)) == NULL)
-					goto cleanup;
 			} else if (!strcmp(attr->name, ATTR_comment)) {
 				if ((tmp_comment = string_dup(attr->value)) == NULL)
 					goto cleanup;
@@ -2407,13 +2398,27 @@ sched_settings_frm_svr(struct batch_status *status)
 	if (!dflt_sched) {
 		int err;
 		int priv_dir_update_fail = 0;
+		int validate_log_dir = 0;
+		int validate_priv_dir = 0;
 		struct attropl *patt;
 		char comment[MAX_LOG_SIZE] = {0};
 
-		if (tmp_log_dir != NULL) {
-			(void)snprintf(path_log,  sizeof(path_log), "%s", tmp_log_dir);
+		if (log_dir == NULL)
+			validate_log_dir = 1;
+		else if (strcmp(log_dir, tmp_log_dir) != 0)
+			validate_log_dir = 1;
+
+		if (priv_dir == NULL)
+			validate_priv_dir = 1;
+		else if (strcmp(priv_dir, tmp_priv_dir) != 0)
+			validate_priv_dir = 1;
+
+		if (!validate_log_dir && !validate_priv_dir && tmp_comment != NULL)
+			clear_comment = 1;
+
+		if (validate_log_dir) {
 			log_close(1);
-			if (log_open(logfile, path_log) == -1) {
+			if (log_open(logfile, tmp_log_dir) == -1) {
 				/* update the sched comment attribute with the reason for failure */
 				attribs = calloc(2, sizeof(struct attropl));
 				if (attribs == NULL) {
@@ -2445,10 +2450,14 @@ sched_settings_frm_svr(struct batch_status *status)
 				snprintf(log_buffer, sizeof(log_buffer), "scheduler log directory is changed to %s", tmp_log_dir);
 				schdlog(PBSEVENT_DEBUG3, PBS_EVENTCLASS_SCHED, LOG_DEBUG,
 						"reconfigure", log_buffer);
+				free(log_dir);
+				if ((log_dir = string_dup(tmp_log_dir)) == NULL) {
+					return 0;
+				}
 			}
 		}
 
-		if (tmp_priv_dir != NULL) {
+		if (validate_priv_dir) {
 			int c = -1;
 #if !defined(DEBUG) && !defined(NO_SECURITY_CHECK)
 				c  = chk_file_sec(tmp_priv_dir, 1, 0, S_IWGRP|S_IWOTH, 1);
@@ -2468,7 +2477,6 @@ sched_settings_frm_svr(struct batch_status *status)
 					priv_dir_update_fail = 1;
 				} else {
 					int lockfds;
-					(void)unlink("sched.lock");
 					lockfds = open("sched.lock", O_CREAT|O_WRONLY, 0644);
 					if (lockfds < 0) {
 						snprintf(log_buffer, sizeof(log_buffer), "PBS failed validation checks for directory %s", tmp_priv_dir);
@@ -2484,11 +2492,16 @@ sched_settings_frm_svr(struct batch_status *status)
 						#endif
 							(void)sprintf(log_buffer, "%d\n", getpid());
 						(void)write(lockfds, log_buffer, strlen(log_buffer));
+						close(lockfds);
 						snprintf(log_buffer, sizeof(log_buffer), "scheduler priv directory has changed to %s", tmp_priv_dir);
 						schdlog(PBSEVENT_DEBUG3, PBS_EVENTCLASS_SCHED, LOG_DEBUG,
 								"reconfigure", log_buffer);
 						if (tmp_comment != NULL)
 							clear_comment = 1;
+						free(priv_dir);
+						if ((priv_dir = string_dup(tmp_priv_dir)) == NULL) {
+							return 0;
+						}
 					}
 				}
 			}
@@ -2521,12 +2534,6 @@ sched_settings_frm_svr(struct batch_status *status)
 				schdlog(PBSEVENT_ERROR, PBS_EVENTCLASS_SCHED, LOG_ERR, __func__, log_buffer);
 			}
 			goto cleanup;
-		}
-		if (cstrcmp(partitions, tmp_partitions) != 0) {
-			free(partitions);
-			partitions = NULL;
-			if ((tmp_partitions != NULL) && (partitions = string_dup(tmp_partitions)) == NULL)
-				goto cleanup;
 		}
 	}
 	if (clear_comment) {
@@ -2566,7 +2573,6 @@ cleanup:
 	free(tmp_log_dir);
 	free(tmp_priv_dir);
 	free(tmp_comment);
-	free(tmp_partitions);
 	return ret;
 
 }
@@ -2599,6 +2605,7 @@ update_svr_schedobj(int connector, int cmd, int alarm_time)
 	int	err;
 	struct	attropl	*attribs, *patt;
 	struct batch_status *ss = NULL;
+	char sched_host[PBS_MAXHOSTNAME + 1];
 
 	/* This command is only sent on restart of the server */
 	if (cmd == SCH_SCHEDULE_FIRST)
@@ -2619,12 +2626,6 @@ update_svr_schedobj(int connector, int cmd, int alarm_time)
 	if (!sched_settings_frm_svr(ss))
 		return 0;
 
-	if (!dflt_sched && (partitions == NULL)) {
-		sprintf(log_buffer, "Scheduler does not contain a partition");
-		log_err(-1, __func__, log_buffer);
-		return 0;
-	}
-
 	pbs_statfree(ss);
 
 	/* update the sched with new values */
@@ -2633,9 +2634,17 @@ update_svr_schedobj(int connector, int cmd, int alarm_time)
 		schdlog(PBSEVENT_DEBUG, PBS_EVENTCLASS_SCHED, LOG_INFO, __func__, MEM_ERR_MSG);
 		return 0;
 	}
+
+	if ((gethostname(sched_host, (sizeof(sched_host) - 1)) == -1) ||
+		(get_fullhostname(sched_host, sched_host, (sizeof(sched_host) - 1)) == -1)) {
+		log_err(-1, __func__, "Unable to get my host name");
+		free(attribs);
+		return 0;
+	}
+
 	patt = attribs;
 	patt->name = ATTR_SchedHost;
-	patt->value = scheduler_host_name;
+	patt->value = sched_host;
 	patt->next = patt + 1;
 	patt++;
 	patt->name = ATTR_sched_port;
--- a/src/scheduler/server_info.c
+++ b/src/scheduler/server_info.c
@@ -241,6 +241,14 @@ query_server(status *pol, int pbs_sd)
 	query_sched_obj(policy, sched, sinfo);
 	pbs_statfree(sched);
 
+	if (!dflt_sched && (sinfo->partitions == NULL)) {
+		snprintf(log_buffer, sizeof(log_buffer), "Scheduler does not contain a partition");
+		schdlog(PBSEVENT_SCHED, PBS_EVENTCLASS_SERVER, LOG_ERR, __func__, log_buffer);
+		pbs_statfree(server);
+		sinfo->fairshare = NULL;
+		free_server(sinfo, 0);
+		return NULL;
+	}
 
 	/* to avoid a possible race condition in which the time it takes to
 	 * query nodes is long enough that a reservation may have crossed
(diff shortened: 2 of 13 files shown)


src/scheduler/fifo.c
```diff
@@ -2445,10 +2455,15 @@ sched_settings_frm_svr(struct batch_status *status)
 				snprintf(log_buffer, sizeof(log_buffer), "scheduler log directory is changed to %s", tmp_log_dir);
 				schdlog(PBSEVENT_DEBUG3, PBS_EVENTCLASS_SCHED, LOG_DEBUG,
 						"reconfigure", log_buffer);
+				free(log_dir);
+				if ((log_dir = string_dup(tmp_log_dir)) == NULL) {
+					schdlog(PBSEVENT_ERROR, PBS_EVENTCLASS_SCHED, LOG_ERR, __func__, MEM_ERR_MSG);
```
bhroam: The scheduler only logs error messages after the malloc() itself.  Otherwise you'll see a cascade of logs for the same malloc() failure.  Really the whole point of string_dup() over strdup() is the fact that it logs a malloc() failure.  This log isn't needed.

src/scheduler/fifo.c
```diff
@@ -2484,11 +2498,17 @@ sched_settings_frm_svr(struct batch_status *status)
...
 						(void)write(lockfds, log_buffer, strlen(log_buffer));
+						close(lockfds);
 						snprintf(log_buffer, sizeof(log_buffer), "scheduler priv directory has changed to %s", tmp_priv_dir);
 						schdlog(PBSEVENT_DEBUG3, PBS_EVENTCLASS_SCHED, LOG_DEBUG,
 								"reconfigure", log_buffer);
 						if (tmp_comment != NULL)
 							clear_comment = 1;
+						free(priv_dir);
+						if ((priv_dir = string_dup(tmp_priv_dir)) == NULL) {
+							schdlog(PBSEVENT_ERROR, PBS_EVENTCLASS_SCHED, LOG_ERR, __func__, MEM_ERR_MSG);
```
bhroam: Same here, no need to log

test/tests/functional/pbs_multi_sched.py
```diff
@@ -1637,3 +1637,103 @@ def test_dedicated_time_multisched(self):
...
+        Test to make sure scheduling should be turned off automatically
+        when number of open files per process are exhausted
+        """
+
+        # get the number of open files per process
+        ret = self.du.run_cmd(
+            self.server.hostname, [
+                'ulimit', '-n'], as_script=True, logerr=False)
+        self.assertEqual(ret['rc'], 0)
+        open_files_limit = ret['out'][0]
```
bhroam: There's a python module which you should use instead of this.  It's called resource.  You want resource.RLIMIT_NOFILE

test/tests/functional/pbs_multi_sched.py
```diff
@@ -1637,3 +1637,103 @@ def test_dedicated_time_multisched(self):
...
+
+        # set the number of open files per process to 10
+        ret = self.du.run_cmd(
+            self.server.hostname, [
+                'ulimit', '-n', '10'], as_script=True, logerr=False)
+        self.assertEqual(ret['rc'], 0)
```
bhroam: You should use resource.setrlimit()

test/tests/functional/pbs_multi_sched.py
```diff
@@ -1637,3 +1637,103 @@ def test_dedicated_time_multisched(self):
...
+
+        self.setup_sc3()
+
+        for _ in range(0, 100):
+            self.server.manager(MGR_CMD_SET, SCHED,
+                                {'scheduling': 'True'}, id="sc3", expect=True)
```
bhroam: This probably won't start 100 scheduling cycles.  When multiple events that would start a cycle during a cycle, we only keep one and start a cycle after the current one finishes.  This means you could set scheduling=true more than once during a cycle and it just means you'll start one cycle afterwards.

test/tests/functional/pbs_multi_sched.py
```diff
@@ -1637,3 +1637,103 @@ def test_dedicated_time_multisched(self):
...
+
+        # scheduling should not go to false once all fds per process
+        # are exhausted.
+        self.server.expect(SCHED, {'scheduling': 'True'},
+                           id='sc3', max_attempts=10)
+
+        ret = self.du.run_cmd(
+            self.server.hostname, ['ulimit', '-n', open_files_limit],
+            as_script=True, logerr=False)
+        self.assertEqual(ret['rc'], 0)
```
bhroam: Same as above, use the resource module.

src/scheduler/fifo.c
```diff
@@ -603,12 +603,6 @@ intermediate_schedule(int sd, char *jobid)
 	int ret; /* to re schedule or not */
 	int cycle_cnt = 0; /* count of cycles run */
 
-	/*
-	 * This is required since there is a probability that scheduler's configuration has been changed at
-	 * server through qmgr.
-	 */
-	update_svr_schedobj(connector, 0, 0);
```
bhroam: Now that we're not calling this here, what happens if a multi-sched comes up when the server is down.  We no longer call SCH_CONFIGURE_ATTRS on recovery, so how will a scheduler initially get its sched_priv/sched_log directories.

bhroam:
After poking around a little, I noticed a couple of the global variables are no longer used.  They were not removed during the initial code review.  Now that you're poking around in the multisched code, do you want to remove them?  scheduler_host_name is set and never used.  partitions is set and only used once to do a little bit of validation.  That validation can happen at cycle query time when you really query partitions for use.

test/tests/functional/pbs_multi_sched.py
arungrover: Adding to what Bhroam mentioned, You will have to do this change a superuser.

src/scheduler/fifo.c
suresh-thelkar: Implemented.

src/scheduler/fifo.c
suresh-thelkar: Implemented.

test/tests/functional/pbs_multi_sched.py
suresh-thelkar: Thanks for your suggestion. I am now using resource module of python. I initially observed that we are directly using "ulimit -n" command in some other part of the PTL(pbs_hook_debug_nocrash.py) and hence used run_cmd earlier. May be we need to correct that as well. Also with run_cmd we can execute ulimit -n as root, but with resource module we can not do it like that and hence at the start of the test case I have used the following.

      if os.getuid() != 0 or sys.platform in ('cygwin', 'win32'):
            self.skipTest("Test need to run as root")

Hope it is fine.

test/tests/functional/pbs_multi_sched.py
suresh-thelkar: Done

test/tests/functional/pbs_multi_sched.py
suresh-thelkar: Thanks for catching this Bhroam. I did not notice it earlier. Now I am making scheduler_iteration=1 second and waiting for more than 10 seconds for making sure that at least 10 scheduling cycles will be completed and it also exhausts the set open files per process which is 10 in this case.

Earlier also for some reason when I run this test case it always used to elapse 10 open files process and hence the test was failing before the fix.

test/tests/functional/pbs_multi_sched.py
suresh-thelkar: Done

src/scheduler/fifo.c
suresh-thelkar: As soon as Server comes up the function update_svr_schedobj gets called(from pbs_sched.c) which in turn calls the function sched_settings_frm_svr because of the command SCH_SCHEDULE_FIRST. This is how Multisched gets its log and priv directories configured at Server. Till the time Server comes up it assumes that sched_logs_sc1 and sched_priv_sc1 are its desired directories if Multisched is named as sc1. Hope this answers your question.

suresh-thelkar:
@bhroam,
Please find the below reply for the last comment you have given.

scheduler_host_name is actually getting used to update its host name back to the Server in the function update_svr_schedobj. As we know this function gets called multiple times. This is actually useful in case if tomorrow we would like to change the direction of connection between Server and Scheduler. It is also needed for backward compatibility reasons I believe.

In order to eliminate this global variable I can actually get the value of sceduler_host_name in the function update_svr_schedobj itself by doing one of the following.
1. Get its host name by parsing pbs_conf.pbs_leaf_name if present
2. Using gethostname()

partitions variable is used in the following functions apart from sched_settings_frm_svr
1. node_in_partition
2. queue_in_partition
3. update_svr_schedobj

_Design to eliminate the global variable partition_
1. Create an entry named partitions inside server_info data structure.
2. Update its value in query_sched_obj.
3. Pass sinfo->partitions as additional parameter to query_sched_obj.
4. Pass sinfo->partitions as additional parameter to node_in_partition.
5. With the above steps now we no need to query partitions inside sched_settings_frm_svr so eliminate the corresponding code.

Please let me know If I am missing anything in the above approach. Since this is a kind of re-factoring the code I would like to do this as part of a new ticket so that it is easy for tracking and testing. This is just a thought, But please do let me know in case if you feel that I have to do these changes as part of the current ticket. I will be waiting for your confirmation on this so that I can proceed accordingly.

suresh-thelkar:
@bhroam / @arungrover ,
Thanks very much for your quick reviews. I have implemented your review comments. Please look into the latest commit for the code changes and provide your further comments/sign off.

test/tests/functional/pbs_multi_sched.py
```diff
@@ -1644,34 +1645,42 @@ def test_auto_sched_off_due_to_fds_limit(self):
...
-        for _ in range(0, 100):
-            self.server.manager(MGR_CMD_SET, SCHED,
-                                {'scheduling': 'True'}, id="sc3", expect=True)
+        self.server.manager(MGR_CMD_SET, SCHED, {'scheduler_iteration': 1},
+                            id="sc3", expect=True)
+        self.server.manager(MGR_CMD_SET, SCHED, {'scheduling': 'True'},
+                            id="sc3", expect=True)
 
+        # This will trigger required number of scheduling cycles that are
+        # needed to exhaust open files per process which is 10 in our case.
```
bhroam: Turn this comment into a self.logger().  It'll be nicer to a human running the test to know why the test is sleeping.

test/tests/functional/pbs_multi_sched.py
```diff
@@ -1644,34 +1645,42 @@ def test_auto_sched_off_due_to_fds_limit(self):
...
+            # get the number of open files per process
+            (open_files_soft_limit, open_files_hard_limit) =\
+                resource.getrlimit(resource.RLIMIT_NOFILE)
+
+            # set the soft limit of number of open files per process to 10
+            resource.setrlimit(resource.RLIMIT_NOFILE,
+                               (10, open_files_hard_limit))
+        except (ValueError, resource.error):
+            self.assertFalse("Error in accessing system RLIMIT_ variales, "
+                             "test fails.")
```
bhroam: The correct call to this method is self.assertFalse(condition, msg).  While a message is always true, you really should do something like self.assertFalse(True, msg).

test/tests/functional/pbs_multi_sched.py
```diff
@@ -1644,34 +1645,42 @@ def test_auto_sched_off_due_to_fds_limit(self):
...
-        ret = self.du.run_cmd(
-            self.server.hostname, ['ulimit', '-n', open_files_limit],
-            as_script=True, logerr=False)
-        self.assertEqual(ret['rc'], 0)
+        try:
+            resource.setrlimit(resource.RLIMIT_NOFILE, (open_files_soft_limit,
+                                                        open_files_hard_limit))
+        except (ValueError, resource.error):
+            self.assertFalse("Error in accessing system RLIMIT_ variales, "
+                             "test fails.")
```
bhroam: Same as the above comment about the assert.

test/tests/functional/pbs_multi_sched.py
```diff
@@ -1644,34 +1645,42 @@ def test_auto_sched_off_due_to_fds_limit(self):
...
         """
 
-        # get the number of open files per process
-        ret = self.du.run_cmd(
-            self.server.hostname, [
-                'ulimit', '-n'], as_script=True, logerr=False)
-        self.assertEqual(ret['rc'], 0)
-        open_files_limit = ret['out'][0]
+        if os.getuid() != 0 or sys.platform in ('cygwin', 'win32'):
+            self.skipTest("Test need to run as root")
```
bhroam: While I'm fine with this, I do know that PDO does not run their tests as root.  You might consider consulting with them.

bhroam:
@suresh-thelkar, I'd like to get rid of the two globals.  I think your solution for sched_host_name is actually better than having the global.  I also think that removing partitions is the way to go since there is no real reason to have it as a global.  You can query it at the start of the cycle.  I don't understand why you need to pass it into query_sched_obj() though.  The sinfo is passed into query_sched_obj().  That should be fine.

As for doing this in this ticket or another, I'd rather do it here.  If you submit it as a separate ticket, it will never be done.  Whenever I am working in an area of the code and there is cleanup that can be done, I do the cleanup.  Submitting these types of tickets just leaves cruft in the code and adds tickets we'll never get to on our backlog.

test/tests/functional/pbs_multi_sched.py
suresh-thelkar: Implemented.

test/tests/functional/pbs_multi_sched.py
suresh-thelkar: Implemented.

test/tests/functional/pbs_multi_sched.py
suresh-thelkar: Implemented.

suresh-thelkar:
@arungrover , Thanks for your sign off. 

@bhroam, If you do not have any further comments can you please sign off.

src/scheduler/fifo.c
```diff
@@ -2617,6 +2605,7 @@ update_svr_schedobj(int connector, int cmd, int alarm_time)
 	int	err;
 	struct	attropl	*attribs, *patt;
 	struct batch_status *ss = NULL;
+	char sched_host[PBS_MAXHOSTNAME+1];
```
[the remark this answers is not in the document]
suresh-thelkar: Done

src/scheduler/server_info.c
```diff
@@ -785,6 +790,9 @@ query_sched_obj(status *policy, struct batch_status *sched, server_info *sinfo)
 		if (!strcmp(attrp->name, ATTR_sched_cycle_len)) {
 			sinfo->sched_cycle_len = res_to_num(attrp->value, NULL);
 		}
+		else if (!strcmp(attrp->name, ATTR_partition)) {
+			sinfo->partitions = string_dup(attrp->value);
```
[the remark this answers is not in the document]
suresh-thelkar: Implemented.

test/tests/functional/pbs_multi_sched.py
```diff
@@ -1667,8 +1669,9 @@ def test_auto_sched_off_due_to_fds_limit(self):
         self.server.manager(MGR_CMD_SET, SCHED, {'scheduling': 'True'},
                             id="sc3", expect=True)
 
-        # This will trigger required number of scheduling cycles that are
-        # needed to exhaust open files per process which is 10 in our case.
+        self.logger.info('This sleep will trigger required number of '
```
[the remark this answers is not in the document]
suresh-thelkar: Done

src/scheduler/server_info.c
```diff
@@ -241,6 +241,11 @@ query_server(status *pol, int pbs_sd)
 	query_sched_obj(policy, sched, sinfo);
 	pbs_statfree(sched);
 
+	if (!dflt_sched && (sinfo->partitions == NULL)) {
+		snprintf(log_buffer, sizeof(log_buffer), "Scheduler does not contain a partition");
+		schdlog(PBSEVENT_SCHED, PBS_EVENTCLASS_SERVER, LOG_ERR, __func__, log_buffer);
+		return NULL;
```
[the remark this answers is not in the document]
suresh-thelkar: Thanks for catching this Bhroam. I have implemented this.

suresh-thelkar:
@bhroam,
I have implemented some more re-factoring changes as you suggested. Please look into the last commit for details. Please provide your further comments if any/sign off.

suresh-thelkar:
Thank you @bhroam for sign off. I have squashed and rebased with mainline. Can you please take this PR forward.

bhroam:
@suresh-thelkar you really should have waited for @arungrover to take another look before you squashed.  He should look at the additional code changes between when he signed off and now.  In any case, the PR is out of date.  Please rebase.

suresh-thelkar:
@bhroam, I have rebased it again. Please take it forward.

bhroam:
I merged in the PR.

PR #635, opened 2018-04-12
fix node oversubscription during preemption

arungrover:
#### Bug/feature Description
Jobs with restrict_res_to_release_on_suspend set may oversubscribe the node.

#### Affected Platform(s)
All

There is an issue that got introduced by mistake while addressing PP-863 (https://github.com/PBSPro/pbspro/commit/a6fc64db7290aff9ee28b17f977bf1f793da4e2c#diff-5f979fbfa8c6c0619fa4f2aee7fd90e6 ) This issue can cause a node to be oversubscribed if a filler job starts running on the resources released by the suspended job in the same cycle when it was suspended.

Even if we fix this, there are chances of hitting another issue where the suspended job may not end up in the calendar "sched_preempt_enforce_resumption" is also set to True. This is because in the cycle we suspend the job we do not recreate exec select of the job and it may result in job asking for non-released resources again which can prevent it from calendaring in that particular cycle. This has a potential to delay suspended jobs from resuming on time.

#### Solution Description
Made changes to release the correct amount of resources from the node when a job is suspended and also set execselect on the job as soon as it is suspended.

#### Checklist:

#### Test logs
[after_fix_logs.txt](https://github.com/PBSPro/pbspro/files/1904320/after_fix_logs.txt)
[before_fix.txt](https://github.com/PBSPro/pbspro/files/1904321/before_fix.txt)
[before_fix2.txt](https://github.com/PBSPro/pbspro/files/1904322/before_fix2.txt)


test/tests/functional/pbs_release_limited_res_suspend.py
```diff
@@ -848,3 +848,84 @@ def test_normal_user_unable_to_see_res_released(self):
...
+        self.server.manager(MGR_CMD_SET, SCHED, a)
+
+        j1 = Job(TEST_USER)
+        j1.set_attributes({ATTR_l + '.select': '1:ncpus=2:mem=512mb'})
+        jid1 = self.server.submit(j1)
+        self.server.expect(JOB, {ATTR_state: 'R'}, id=jid1)
+
+        # Submit a filler job
+        j2 = Job(TEST_USER)
+        j2.set_attributes({ATTR_l + '.select': '1:ncpus=3',
```
bhroam: You never set the number of ncpus on the machine.  This job will not run on a node with 1 cpu.

test/tests/functional/pbs_release_limited_res_suspend.py
```diff
@@ -848,3 +848,84 @@ def test_normal_user_unable_to_see_res_released(self):
...
+                           ATTR_q: 'expressq',
+                           ATTR_l + '.walltime': 100})
+        jid3 = self.server.submit(j3)
+        self.server.expect(JOB, {ATTR_state: 'R'}, id=jid3)
+        self.server.expect(JOB, {ATTR_state: 'S'}, id=jid1)
+        # Check that resources_assigned is not exceeding resources_available
+        ras_ncpus = ATTR_rescassn + '.ncpus'
+        rav_ncpus = ATTR_rescavail + '.ncpus'
+        rv = self.server.status(
+            NODE, [ras_ncpus, rav_ncpus], id=self.mom.shortname)
```
bhroam: Instead of creating two variables, just create one variable that contains a list and pass it into status()

test/tests/functional/pbs_release_limited_res_suspend.py
```diff
@@ -848,3 +848,84 @@ def test_normal_user_unable_to_see_res_released(self):
...
+        self.assertNotEqual(rv, None)
+
+        self.assertTrue(rv[0][ras_ncpus] <= rv[0][rav_ncpus],
```
bhroam: use assertLessEqual()

test/tests/functional/pbs_release_limited_res_suspend.py
```diff
@@ -848,3 +848,84 @@ def test_normal_user_unable_to_see_res_released(self):
...
+        """
+        # Set mem in restrict_res_to_release_on_suspend server attribute
+        a = {ATTR_restrict_res_to_release_on_suspend: 'mem'}
+        self.server.manager(MGR_CMD_SET, SERVER, a)
+
+        a = {ATTR_sched_preempt_enforce_resumption: True}
+        self.server.manager(MGR_CMD_SET, SCHED, a)
+
+        # Set 5 ncpus available on the node
+        a = {ATTR_rescavail + '.ncpus': 5}
```
bhroam: Maybe you should set mem as well to be at least 2gb?

test/tests/functional/pbs_release_limited_res_suspend.py
```diff
@@ -848,3 +848,84 @@ def test_normal_user_unable_to_see_res_released(self):
...
+        j2.set_attributes({ATTR_l + '.select': '1:ncpus=2:mem=2gb',
+                           ATTR_q: 'expressq',
+                           ATTR_l + '.walltime': 100})
+        jid2 = self.server.submit(j2)
+        self.server.expect(JOB, {ATTR_state: 'R'}, id=jid2)
+        self.server.expect(JOB, {ATTR_state: 'S'}, id=jid1)
+
+        # Check if the job is calendared
+        self.scheduler.log_match(
+            jid1 + ";Can't find start time estimate", existence=False,
```
bhroam: I'd search for the positive message of "Job is a top job and will run at"

test/tests/functional/pbs_release_limited_res_suspend.py
arungrover: setup in this test sets the node ncpus to 4

test/tests/functional/pbs_release_limited_res_suspend.py
arungrover: will do

test/tests/functional/pbs_release_limited_res_suspend.py
arungrover: It is set in setup

test/tests/functional/pbs_release_limited_res_suspend.py
arungrover: I did not choose that because next cycle happens very quickly and it gets calendared. I wanted to make sure calendaring never failed for suspended job

test/tests/functional/pbs_release_limited_res_suspend.py
arungrover: I'd like to keep it as is because the same variables are then used to read from the return value of status call.

arungrover:
@bhroam I've addressed your comments. Please have a look.

test/tests/functional/pbs_release_limited_res_suspend.py
```diff
@@ -848,3 +848,84 @@ def test_normal_user_unable_to_see_res_released(self):
...
+
+        # Submit a high priority job
+        j3 = Job(TEST_USER)
+        j3.set_attributes({ATTR_l + '.select': '1:ncpus=1:mem=2gb',
+                           ATTR_q: 'expressq',
+                           ATTR_l + '.walltime': 100})
+        jid3 = self.server.submit(j3)
+        self.server.expect(JOB, {ATTR_state: 'R'}, id=jid3)
+        self.server.expect(JOB, {ATTR_state: 'S'}, id=jid1)
+        # Check that resources_assigned is not exceeding resources_available
```
lisa-altair: @I don't think it really matters, but jid3 can't be running until jid1 is suspended, you could switch this check around to look for the suspended job before looking for the running job.  But again, I doubt it really matters.

test/tests/functional/pbs_release_limited_res_suspend.py
```diff
@@ -848,3 +848,84 @@ def test_normal_user_unable_to_see_res_released(self):
...
+        rv = self.server.status(
+            NODE, [ras_ncpus, rav_ncpus], id=self.mom.shortname)
+        self.assertNotEqual(rv, None)
+
+        self.assertLessEqual(rv[0][ras_ncpus], rv[0][rav_ncpus],
+                             msg="pbs released resource ncpus incorrectly")
+
+        # Expect filler job to be in queued state because
+        # suspended job did not release any ncpus
+        self.server.expect(JOB, {ATTR_state: 'Q'}, id=jid2)
```
lisa-altair: Can you explain why the suspended job doesn't release ncpus?  I think you only set memory not to be freed on suspend...what am I missing?

test/tests/functional/pbs_release_limited_res_suspend.py
arungrover: You are right Lisa, it would not matter. If you are ok with it, I'd like to retain it as is

test/tests/functional/pbs_release_limited_res_suspend.py
arungrover: That's because resources to release is set to only memory. So the job will retain ncpus.

test/tests/functional/pbs_release_limited_res_suspend.py
lisa-altair: I am fine to leave it as is.

test/tests/functional/pbs_release_limited_res_suspend.py
lisa-altair: Ah, I had it backwards.  You set only memory to be freed. Doh.  Thanks for explaining.

bhroam:
I merged in the PR

PR #636, opened 2018-04-12
Add deepcopy support to pbs.size type

minghui-liu:
#### Bug/feature Description
`pbs.size` type does not support deep copy. So if we do this:
```
import pbs
import copy
a = pbs.size(1000)
b = copy.deepcopy(a)
```
`b` will be ` 0` instead of `1000b`

#### Affected Platform(s)
All

#### Solution Description
Implement `__deepcopy__` method of `pbs.size` type. In this case it is pretty straightforward. The method simply creates and returns a new `pbs.size` object with the same value. 

#### Checklist:

#### Test Log
[PTL test log.txt](https://github.com/PBSPro/pbspro/files/1904937/PTL.test.log.txt)

### DIFF
--- a/src/modules/python/pbs/v1/_base_types.py
+++ b/src/modules/python/pbs/v1/_base_types.py
@@ -533,6 +533,9 @@ class size(_size):
 	# and not in _size's richcompare.
         return size(s.__sub__(o))
 
+    def __deepcopy__(self, mem):
+        return size(str(self))
+
 class duration(int):
     """
     Represents an interval or elapsed time object in number of seconds. This is
--- a/test/tests/functional/pbs_types.py
+++ b/test/tests/functional/pbs_types.py
@@ -1,0 +1,69 @@
+# coding: utf-8
+
+# Copyright (C) 1994-2018 Altair Engineering, Inc.
+# For more information, contact Altair at www.altair.com.
+#
+# This file is part of the PBS Professional ("PBS Pro") software.
+#
+# Open Source License Information:
+#
+# PBS Pro is free software. You can redistribute it and/or modify it under the
+# terms of the GNU Affero General Public License as published by the Free
+# Software Foundation, either version 3 of the License, or (at your option) any
+# later version.
+#
+# PBS Pro is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.
+# See the GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+# Commercial License Information:
+#
+# For a copy of the commercial license terms and conditions,
+# go to: (http://www.pbspro.com/UserArea/agreement.html)
+# or contact the Altair Legal Department.
+#
+# Altair’s dual-license business model allows companies, individuals, and
+# organizations to create proprietary derivative works of PBS Pro and
+# distribute them - whether embedded or bundled with other software -
+# under a commercial license agreement.
+#
+# Use of Altair’s trademarks, including but not limited to "PBS™",
+# "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
+# trademark licensing policies.
+
+
+from tests.functional import *
+
+
+class Test_pbs_types(TestFunctional):
+    """
+    This test suite tests pbs python types and their related functions
+    """
+    def test_pbs_size_deepcopy(self):
+        """
+        Test that deepcopy works for pbs.size type
+        """
+        hook_content = ("""
+import pbs
+import copy
+a = pbs.size(1000)
+b = copy.deepcopy(a)
+c = a
+pbs.logmsg(pbs.EVENT_DEBUG, 'a=%s, b=%s, c=%s' % (a, b, c))
+d = pbs.size('1m')
+e = copy.deepcopy(d)
+f = d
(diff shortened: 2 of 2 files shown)


src/modules/python/pbs/v1/_base_types.py
```diff
@@ -532,6 +532,9 @@ def __sub__(self,other):
 	# return would look in here for comparison operators
 	# and not in _size's richcompare.
         return size(s.__sub__(o))
+    
+    def __deepcopy__(self, mem):
```
hirenvadalia: @minghui-liu However code looks good but curious to know what will be benefits of using ```copy.deepcopy()``` over directly ```pbs.size()``` to create copy ?

test/tests/functional/pbs_types.py
```diff
@@ -0,0 +1,73 @@
...
+                        "a = pbs.size(1000)\n"
+                        "b = copy.deepcopy(a)\n"
+                        "c = a\n"
+                        "pbs.logmsg(pbs.EVENT_DEBUG, "
+                        "'a=%s, b=%s, c=%s' % (a, b, c))\n"
+                        "d = pbs.size('1m')\n"
+                        "e = copy.deepcopy(d)\n"
+                        "f = d\n"
+                        "pbs.logmsg(pbs.EVENT_DEBUG, "
+                        "'d=%s, e=%s, f=%s' % (d, e, f))\n")
```
bhroam: I think it a """ quoted string is clearer for strings like this.  It makes it extremely clear what the hook will look like because it's exactly what is in the string.  You have two lines which carry over to the next line.  It is easy to miss the fact that they don't end in a \n.

test/tests/functional/pbs_types.py
```diff
@@ -0,0 +1,73 @@
...
+        hook_name = 'deepcopy'
+        hook_attr = {'enabled': 'true', 'event': 'execjob_begin'}
```
bhroam: It might be simpler to use a queue job hook.  The pbs.size() type is available to any hook.  The execjob_begin hook does work, but you need to have a job run.  It won't matter with the queuejob hook.

test/tests/functional/pbs_types.py
```diff
@@ -0,0 +1,73 @@
...
+        self.server.create_import_hook(hook_name, hook_attr, hook_content)
+
+        j = Job(TEST_USER)
+        j.create_script('/bin/true')
```
bhroam: Why does the job have to exit?  You have an execjob_begin hook.  There's no reason to bother.  Just use the standard sleep.

test/tests/functional/pbs_types.py
```diff
@@ -0,0 +1,73 @@
...
+        jid = self.server.submit(j)
+        self.server.expect(JOB, 'queue', id=jid, op=UNSET, max_attempts=10,
+                           interval=1, offset=1)
```
bhroam: Once again, no reason to care that the job exits.

src/modules/python/pbs/v1/_base_types.py
```diff
@@ -533,6 +533,9 @@ def __sub__(self,other):
 	# and not in _size's richcompare.
         return size(s.__sub__(o))
 
+    def __deepcopy__(self, mem):
+        return size(str(self))
```
bhroam: I don't see the str() method defined for this type.  What will it return?

src/modules/python/pbs/v1/_base_types.py
minghui-liu: `deepcopy` recursively creates copies of members within a type. So if some other types have `pbs.size` as member and want to use deepcopy, then `pbs.size` needs to support it.

src/modules/python/pbs/v1/_base_types.py
minghui-liu: It returns a string representing the size. It's implemented in pbs_python_svr_size_type.c.

src/modules/python/pbs/v1/_base_types.py
hirenvadalia: Ah I see... Thanks for explanation... Now thinking more, should our all PBS types support __deepcopy__?
(Off-course not in this PR 😄)

test/tests/functional/pbs_types.py
```diff
@@ -47,27 +47,23 @@ def test_pbs_size_deepcopy(self):
...
+pbs.logmsg(pbs.EVENT_DEBUG, 'd=%s, e=%s, f=%s' % (d, e, f))
+""")
         hook_name = 'deepcopy'
-        hook_attr = {'enabled': 'true', 'event': 'execjob_begin'}
+        hook_attr = {'enabled': 'true', 'event': 'queuejob'}
         self.server.create_import_hook(hook_name, hook_attr, hook_content)
 
         j = Job(TEST_USER)
-        j.create_script('/bin/true')
         jid = self.server.submit(j)
```
hirenvadalia: Please remove jid as it is unused... and codacy is complaining about it.

minghui-liu: Removed.

src/modules/python/pbs/v1/_base_types.py
minghui-liu: Yes I think most types can benefit from deepcopy. Maybe except some types that are designed to have only one instance.

mike0042: @hirenvadalia : Yes, I think all types that could be referenced in a deepcopy() operation should be made to support it. Otherwise, the implementation would be incomplete, IMHO.

minghui-liu:
I have rebased again. Please take this PR further, @PBSPro/pbspro-maintainers . Thanks.

bhroam:
I merged in the PR.

minghui-liu:
Thank you.

PR #581, opened 2018-02-28
PP-1009: Enhancing sudo usage in PTL

kjakkali:
#### Bug/feature Description
 [PP-1009](https://pbspro.atlassian.net/browse/PP-1009)

#### Affected Platform(s)
* *All platforms*

#### Cause / Analysis / Design
In PTL
- the existing interface to create temporary file return file name and file descriptor (FD). If test case create a file as different users than the current user the file FD is invalid as current user can't make use the returned FD.
- For file copy operation PTL by default does copy file with keeping original permission and then change owner and permissions of file

#### Solution Description
- New interface to create a temporary file
- Adding a new parameter 'preserve_permission' to run_copy() to override the default file copy operation.

Doc : https://pbspro.atlassian.net/wiki/spaces/PD/pages/72974341/PP-1009+Enhancing+sudo+usage+in+PTL

#### Checklist:

- [x] I have joined the **[pbspro community forum](http://community.pbspro.org/)**.
- [x] My pull request contains a **single, signed** commit. See **[setting up gpg signature](https://pbspro.atlassian.net/wiki/display/DG/Signing+Your+Git+Commits)**.
- [x] My code follows the **[coding style](https://pbspro.atlassian.net/wiki/display/DG/Coding+Standards)** of this project.
- [ ] My change requires project documentation. See **[required documentation checklist](https://pbspro.atlassian.net/wiki/display/DG/Checklist+for+Developing+Features+and+Bug+Fixes)** for details.
   - [ ] I have added documentation in the **[project documentation area](https://pbspro.atlassian.net/wiki/display/PD)**.
- [ ] I have added new **PTL test(s) to my commit**. (See **[using PTL for testing](https://pbspro.atlassian.net/wiki/display/DG/Using+PTL+for+Testing)**) *(or)*
   - [ ] I have added  **manual test(s) to this pull request and explained why PTL is not appropriate** for this case.
- [ ] All new and existing automated tests have passed. (See **[running automated PTL tests](https://pbspro.atlassian.net/wiki/display/DG/PTL+Quick+Start+Guide)**).
- [ ] I have attached **test logs to this pull request** as evidence of testing/verification.

__***For further information please visit the [Developer Guide Home](https://pbspro.atlassian.net/wiki/display/DG/Developer+Guide+Home).***__

PR #649, opened 2018-04-20
Removing python-magic dependency from PBS Snapshot

agrawalravi90:
#### Bug/feature Description
* python-magic was a third-party dependency that was used by pbs_snapshot to determine if a file was a core file or not. Because that's the only reason why that module was used, we decided that it's better to just use the file command to do this and remove an additional 3rd party dependency.

#### Affected Platform(s)
* All Linux

#### Solution Description
* Replaced the use of python-magic module with the file command
* The trace capturing script had a bug, it didn't include the executable which generated it, fixed that.
* #581 had renamed DshUtils' mkstemp to create_temp_file but pbs_snaputils and pbs_anonutils were not updated with the new call, so updated those as well.

#### Testing logs/output
* Since an automated test for this would require the test to generate a core file (and possibly mess around with ulimit), I don't think an automated test is possible. I did manual testing by generating a core file from a dummy program called 'create_core', copied it over to server_priv/ and ran pbs_snapshot. PFB the snapshot that it captured (I had to convert it to a .zip as Github doesn't support tars):
[snapshot_20180420_22_12_40.zip](https://github.com/PBSPro/pbspro/files/1933833/snapshot_20180420_22_12_40.zip)

The stack trace captured is inside core_trace_bt/server_priv/core
* Here's also the output of the pbs_snapshot's existing ptl tests: 
[testpbssnapshot.log](https://github.com/PBSPro/pbspro/files/1930133/testpbssnapshot.log)

#### Checklist:

### DIFF
--- a/test/fw/ptl/utils/pbs_snaputils.py
+++ b/test/fw/ptl/utils/pbs_snaputils.py
@@ -38,7 +38,6 @@
 import os
 import time
 import tarfile
-import magic
 import logging
 
 from subprocess import STDOUT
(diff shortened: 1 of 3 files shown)


agrawalravi90:
Thanks for the quick review @bhroam. Requesting one more review, this is a tiny code change!

agrawalravi90:
Requesting another review, this is a tiny change! Thanks in advance.

test/fw/ptl/utils/pbs_snaputils.py
```diff
@@ -761,12 +759,16 @@ def __evaluate_core_file(self, file_path, core_dir):
...
+
         :returns: True if this was a valid core file, otherwise False
         """
         if not os.path.isfile(file_path):
             return False
 
-        file_header = magic.from_file(file_path)
+        # Get the header of this file
+        ret = self.du.run_cmd(cmd=["file", file_path], sudo=self.sudo)
+        assert(ret['err'] is None or len(ret['err']) == 0)
```
mike0042: Codacy is complaining that this assert call will be optimized out in the *.pyo file. If the behavior is critical to the test, then you should raise an error here or fail the test. The rest looks fine.

agrawalravi90: I read that, and the reason that it mentions doesn't apply here. So, I'm going to leave it as is. Thanks for the sign-off!

bhroam: I'm with @mike0042 here.  You put in the assert for a reason and it will be optimized out.  If you click the arrow next to it, it has example code of the correct way to handle the situation.  It also says using an assert() is not considered good style.

agrawalravi90: It says that it gets optimized out when used in a particular way, which I'm not doing here, but if it's not considered good style then Ill change it.

test/fw/ptl/utils/pbs_snaputils.py
```diff
@@ -767,7 +767,9 @@ def __evaluate_core_file(self, file_path, core_dir):
 
         # Get the header of this file
         ret = self.du.run_cmd(cmd=["file", file_path], sudo=self.sudo)
-        assert(ret['err'] is None or len(ret['err']) == 0)
+        if ret['err'] is not None and len(ret['err']) != 0:
+            raise AssertionError()
```
mike0042: Shouldn't you provide some descriptive text as an argument to AssertionError()?

agrawalravi90:
Sorry about that Mike, it's Friday, I guess I'm feeling lazy :)

I just added some text, please let me know if it looks fine now. Thanks!

agrawalravi90:
Thanks Mike, I squashed and rebased.

PR #660, opened 2018-04-27
Make alps reservation query requests tunable

arungrover:
#### Bug/feature Description
On Cray when a job finishes PBS sends a request to ALPS a release reservation request.  PBS then will intermittently poll until the ALPS response is "No entry for resId" is received. As of today, time delay between each polling cycle is exponential in nature and the according to Cray this can be sped up.
This pull request makes the default polling interval (and jitter) small and also exposes these tunables to admins wherein they can be set appropriately.

#### Affected Platform(s)
Cray platform

Lowering down the default value of polling wait time to 0.4 second and polling jitter that is added to the polling wait time to 0.12 second

#### Solution Description
Adding support for new tunable "alps_release_jitter" and "alps_release_interval" which will be used to wait for each subsequent alps reservation cancellation request.

#### Testing logs/output
This functionality can only be tested on an actual Cray machine. I've added a PTL script to the pull but cray simulator is skipped because on a simulator the reservation request gets canceled when the very first cancellation request is received (unlike the real world use case).

#### Checklist:

Documentation [Link](https://pbspro.atlassian.net/wiki/spaces/PD/pages/216498179/Make+the+wait+time+between+Cray+ALPS+release+reservation+requests+adjustable)

[Craysim testlogs](https://github.com/PBSPro/pbspro/files/1956707/test_logs_from_sim.txt)

### DIFF
--- a/src/resmom/catch_child.c
+++ b/src/resmom/catch_child.c
@@ -109,8 +109,9 @@ extern time_t		time_now;
 extern pbs_list_head	mom_polljobs;
 extern unsigned int	pbs_mom_port;
 #if MOM_ALPS
-extern int		alps_release_interval_max;
+extern useconds_t	alps_release_wait_time;
 extern int		alps_release_timeout;
+extern useconds_t	alps_release_jitter;
 #endif
 
 extern char		*path_hooks_workdir;
@@ -2401,25 +2402,19 @@ del_job_hw(job *pjob)
 			/* We are the child */
 			begin_time = time(NULL);
 			end_time = begin_time;
-			/* casting to unsigned per Open standards */
+			/* add jobid to the seed */
 			srandom((unsigned)(atoi(pjob->ji_qs.ji_jobid) + begin_time));
-			for (i=0;
-				(total_time = end_time - begin_time) < alps_release_timeout;
-				++i) {
+			for (i = 1; (total_time = end_time - begin_time) < alps_release_timeout; ++i, end_time = time(NULL)) {
 				/* calculate time to sleep */
-				sleeptime = i*i;
-				if (sleeptime > alps_release_interval_max) {
-					sleeptime = alps_release_interval_max;
-				}
-				/* Add randomness of 0 to 4 seconds to the
+				sleeptime = alps_release_wait_time;
+				/* Add randomness of 0 to 0.12 seconds to the
 				 * sleeptime so we don't overwhelm ALPS with
 				 * multiple ALPS release requests when jobs end
 				 * at the same time.
 				 */
-				jitter = random() % 5;
-				sleeptime = sleeptime + jitter;
-				sleep(sleeptime);
-				end_time = time(NULL);
+				jitter = random() % alps_release_jitter;
+				sleeptime += jitter;
+				usleep(sleeptime);
 				if ((j = alps_cancel_reservation(pjob)) <= 0)
 					break;
 			}
@@ -2441,9 +2436,8 @@ del_job_hw(job *pjob)
 			} else if (j == 0) {
 				sprintf(log_buffer,
 					"Cancelled ALPS reservation %ld after a "
-					"total of %ld seconds",
-					pjob->ji_extended.ji_ext.ji_reservation,
-					total_time);
+					"total of %d tries",
+					pjob->ji_extended.ji_ext.ji_reservation, i + 1);
 				log_event(PBSEVENT_DEBUG3, PBS_EVENTCLASS_JOB,
 					LOG_DEBUG, pjob->ji_qs.ji_jobid,
 					log_buffer);
--- a/src/resmom/mom_main.c
+++ b/src/resmom/mom_main.c
@@ -268,12 +268,14 @@ time_t		time_resc_updated = 0;
 extern pbs_list_head svr_requests;
 extern struct var_table vtable;	/* see start_exec.c */
 #if	MOM_ALPS
-#define	ALPS_REL_INTERVAL_MAX		30;	/* 30 sec */
+#define	ALPS_REL_WAIT_TIME_DFLT		400000;	/* 0.4 sec */
+#define	ALPS_REL_JITTER_DFLT		120000;	/* 0.12 sec */
 #define	ALPS_REL_TIMEOUT		600;	/* 10 min */
 #define	ALPS_CONF_EMPTY_TIMEOUT		10;	/* 10 sec */
 #define	ALPS_CONF_SWITCH_TIMEOUT	35;	/* 35 sec */
 char	       *alps_client = NULL;
-int		alps_release_interval_max;
+useconds_t	alps_release_wait_time = ALPS_REL_WAIT_TIME_DFLT;
+useconds_t	alps_release_jitter = ALPS_REL_JITTER_DFLT;
 int		vnode_per_numa_node;
 int		alps_release_timeout;
 int		alps_confirm_empty_timeout;
@@ -460,7 +462,8 @@ static handler_ret_t	set_alien_attach(char *);
 static handler_ret_t	set_alien_kill(char *);
 #if	MOM_ALPS
 static handler_ret_t	set_alps_client(char *);
-static handler_ret_t	set_alps_release_interval_max(char *);
+static handler_ret_t	set_alps_release_wait_time(char *);
+static handler_ret_t	set_alps_release_jitter(char *);
 static handler_ret_t	set_alps_release_timeout(char *);
 static handler_ret_t	set_alps_confirm_empty_timeout(char *);
 static handler_ret_t	set_vnode_per_numa_node(char *);
@@ -526,7 +529,8 @@ static struct	specials {
 #if	MOM_ALPS
 	{ "alps_client",		set_alps_client },
 	{ "alps_confirm_empty_timeout", set_alps_confirm_empty_timeout },
-	{ "alps_release_interval_max",	set_alps_release_interval_max },
+	{ "alps_release_wait_time",	set_alps_release_wait_time },
+	{ "alps_release_jitter",	set_alps_release_jitter },
 	{ "alps_release_timeout",	set_alps_release_timeout },
 	{ "vnode_per_numa_node", 	set_vnode_per_numa_node },
 	{ "alps_confirm_switch_timeout",set_alps_confirm_switch_timeout },
@@ -3732,6 +3736,45 @@ set_int(const char *id, char *value, int *var)
 
 /**
  * @brief
+ *	set float value
+ *
+ * @param[in] id - function name
+ * @param[in] value - value
+ * @param[out] var - output float value
+ *
+ * @return      handler_ret_t
+ * @retval      HANDLER_FAIL            Failure
+ * @retval      HANDLER_SUCCESS         Success
+ *
+ */
+handler_ret_t
+set_float(const char *id, char *value, float *var)
+{
+	char	*left;
+	float	val;
+
+	if (value == NULL || *value == '\0') {
+		log_event(PBSEVENT_SYSTEM, PBS_EVENTCLASS_SERVER, LOG_NOTICE,
+			id, "No value specified, no action taken.");
+		return HANDLER_FAIL;	/* error */
+	}
+
+	val = strtod(value, &left);
+	if (left == value || val <= 0) {
+		sprintf(log_buffer, "bad value \"%s\"", value);
+		log_event(PBSEVENT_SYSTEM, 0, LOG_ERR, id, log_buffer);
+		return HANDLER_FAIL;	/* error */
+	}
+	*var = val;
+
+	snprintf(log_buffer, sizeof(log_buffer), "setting %f", val);
+	log_event(PBSEVENT_SYSTEM, 0, LOG_DEBUG, id, log_buffer);
+
+	return HANDLER_SUCCESS;
+}
+
+/**
+ * @brief
  *	Set the configuration flag that defines whether the restart nunction
  * 	occurs in the background.
  *
@@ -3933,23 +3976,49 @@ set_vnode_per_numa_node(char *value)
 
 /**
  * @brief
- * 	Set the maximum time in seconds to wait between ALPS release
- * 	reservation requests
+ *	Set the alps_release_wait_time in micoseconds to wait between ALPS release
+ *	reservation requests
  *
- * @par
- * PBS will make release reservation requests at i*i intervals until
- * the result of i*i is larger than alps_release_interval_max, at
- * which point PBS will only make release reservation requests
- * every alps_release_interval_max seconds on a per job basis.
- *
- * @retval 0 failure
- * @retval 1 success
+ * @return returns value of set_float()
  *
  */
 static handler_ret_t
-set_alps_release_interval_max(char *value)
+set_alps_release_wait_time(char *value)
 {
-	return (set_int(__func__, value, &alps_release_interval_max));
+	float tmp;
+	handler_ret_t ret;
+	double fract, integ;
+	ret = set_float(__func__, value, &tmp);
+	if (ret == HANDLER_SUCCESS) {
+		fract = modf(tmp, &integ);
+		alps_release_wait_time = (useconds_t)integ*1000000 + (fract*1000000);
+	}
+	return (ret);
+}
+
+/**
+ * @brief
+ *	Set the alps_release_jitter value in microseconds.
+ *
+ * @par
+ *	PBS will randomly generate how much
+ *	microseconds to add to the alps_release_wait_time value
+ *
+ * @return returns value of set_float()
+ *
+ */
+static handler_ret_t
+set_alps_release_jitter(char *value)
+{
+	float tmp;
+	handler_ret_t ret;
+	double fract, integ;
+	ret = set_float(__func__, value, &tmp);
+	if (ret == HANDLER_SUCCESS) {
+		fract = modf(tmp, &integ);
+		alps_release_jitter = (useconds_t)integ*1000000 + (fract*1000000);
+	}
+	return (ret);
 }
 
 /**
@@ -4802,7 +4871,8 @@ read_config(char *file)
 #endif /* localmod 015 */
 
 #if MOM_ALPS
-	alps_release_interval_max = ALPS_REL_INTERVAL_MAX;
+	alps_release_wait_time = ALPS_REL_WAIT_TIME_DFLT;
+	alps_release_jitter = ALPS_REL_JITTER_DFLT;
 	vnode_per_numa_node = FALSE;
 	alps_release_timeout      = ALPS_REL_TIMEOUT;
 	alps_confirm_empty_timeout = ALPS_CONF_EMPTY_TIMEOUT;
--- a/test/tests/functional/pbs_alps_release_tunables.py
+++ b/test/tests/functional/pbs_alps_release_tunables.py
@@ -1,0 +1,162 @@
+# coding: utf-8
+
+# Copyright (C) 1994-2018 Altair Engineering, Inc.
+# For more information, contact Altair at www.altair.com.
+#
+# This file is part of the PBS Professional ("PBS Pro") software.
+#
+# Open Source License Information:
+#
+# PBS Pro is free software. You can redistribute it and/or modify it under the
+# terms of the GNU Affero General Public License as published by the Free
+# Software Foundation, either version 3 of the License, or (at your option) any
+# later version.
+#
+# PBS Pro is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.
+# See the GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+# Commercial License Information:
+#
+# For a copy of the commercial license terms and conditions,
+# go to: (http://www.pbspro.com/UserArea/agreement.html)
+# or contact the Altair Legal Department.
+#
+# Altair’s dual-license business model allows companies, individuals, and
+# organizations to create proprietary derivative works of PBS Pro and
+# distribute them - whether embedded or bundled with other software -
+# under a commercial license agreement.
+#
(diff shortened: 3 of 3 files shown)


test/tests/functional/pbs_alps_request_tunables.py
```diff
@@ -0,0 +1,150 @@
...
+        a = time.strptime(msg[:19], "%m/%d/%Y %H:%M:%S")
+        return int(time.mktime(a))
+
+    def test_alps_release_interval(self):
+        """
+        Set alps_release_interval to a higher value and then notice that
+        subsequest reservation cancellation requests are made atleast
+        after the set interval.
+        """
+        ari = 4.298
```
lisa-altair: How did you arrive at this value?  Perhaps add a comment explaining.

test/tests/functional/pbs_alps_request_tunables.py
```diff
@@ -0,0 +1,150 @@
...
+        else:
+            self.skipTest("Reservation cancelled without retry, Try again!")
+
+    def test_alps_release_jitter(self):
+        """
+        Set alps_release_jitter to a higher value and then notice that
+        subsequest reservation cancellation requests are made by adding
+        a random time interval (less than jitter) to alps_release_interval.
+        """
+        arj = 2.198
```
lisa-altair: How did you arrive at this value? Perhaps add a comment explaining.

test/tests/functional/pbs_alps_request_tunables.py
```diff
@@ -0,0 +1,150 @@
...
+        self.mom.add_config({'$alps_release_interval': ari})
+        # There is no good way to test jitter and it is a random number
+        # less than value set in alps_release_jitter. So in this case
+        # we can probably try deleting a reservation a few times.
+        for each in range(5):
+            # submit a job and then delete it after it starts running
+            start_time = time.time()
+            j1 = Job(TEST_USER)
+            jid1 = self.server.submit(j1)
+            self.server.expect(JOB, {ATTR_state: 'R'}, id=jid1)
```
lisa-altair: To ensure that there is enough time for the aprun to get started, maybe we should sleep for a couple seconds before we delete the job? The sleep will likely also cause the ALPS reservation not to be freed right away so that you know you are getting into the code that uses jitter.

test/tests/functional/pbs_alps_request_tunables.py
```diff
@@ -0,0 +1,150 @@
...
+            # cancellation at ALPS level itself.
+            if len(out) >= 2:
+                nline_prev, line_prev = out[0]
+                time_prev = self.get_epoch(line_prev)
+                for data in out[1:]:
+                    nline, line_current = data
+                    time_current = self.get_epoch(line_current)
+                    self.assertLessEqual(time_current - time_prev, max_delay,
+                                         msg="alps_release_jitter not working")
+                    time_prev = time_current
```
lisa-altair: In this test, you don't skip if you only find one cancellation request...But shouldn't you at least keep a count and if for all 5 jobs you only find one cancellation then you should skip/error?
Which does make me wonder, is skipping the test the right thing to do?  Since we're trying to test a base functionality, wouldn't it be better to error out so someone can look into why the wait_time/jitter code isn't being executed?

test/tests/functional/pbs_alps_request_tunables.py
arungrover: It was just a random value big enough to measure using log message timestamps.

test/tests/functional/pbs_alps_request_tunables.py
arungrover: Well, it gets us into a situation where we can not really say that functionality did not work. I thought for the first case it was probably ok to skip because then we can run it again or at the very least the test will not count as an error. For this test, I was not sure if I should be skipping it because there might be other retries where it might exercise the functionality.
Like you said, I can probably keep a count of retries, and skip if it did not exercise the code in any of the retries. I specifically don't want to error out because it is not PBS's fault if the first cancellation call succeeds.

arungrover:
@lisa-altair I've addressed your review comments and also addressed some warnings that Codacy was showing.

test/tests/functional/pbs_alps_request_tunables.py
lisa-altair: Okay, I see what you mean.

test/tests/functional/pbs_alps_release_tunables.py
```diff
@@ -0,0 +1,162 @@
...
+                                 n='ALL', max_attempts=5, regexp=True,
+                                 allmatch=True)
+
+        # We found something, Let's first check there are atleast 2 such
+        # log messages, If not then that means reservation was cancelled
+        # in the first attempt itself, at that point right thing to do is
+        # to either run it again or find out a way to delay the reservation
+        # cancellation at ALPS level itself.
+        if len(out) >= 2:
+            _, line_prev = out[0]
```
vccardenas: You might want to just index the list of tuples, e.g.:
  line_prev = out[0][1]

test/tests/functional/pbs_alps_release_tunables.py
```diff
@@ -0,0 +1,162 @@
...
+            time_prev = get_epoch(line_prev)
+            for data in out[1:]:
+                _, line_current = data
```
vccardenas: Similar to above comment:
    line_current = data[1]

test/tests/functional/pbs_alps_release_tunables.py
```diff
@@ -0,0 +1,162 @@
...
+                                     allmatch=True)
+
+            # We found something, Let's first check there are atleast 2 such
+            # log messages, If not then that means reservation was cancelled
+            # in the first attempt itself, at that point right thing to do is
+            # to either run it again or find out a way to delay the reservation
+            # cancellation at ALPS level itself.
+            if len(out) >= 2:
+                retry -= 1
+                _, line_prev = out[0]
```
vccardenas: You might want to just index the list of tuples, e.g.:
  line_prev = out[0][1]

test/tests/functional/pbs_alps_release_tunables.py
```diff
@@ -0,0 +1,162 @@
...
+                time_prev = get_epoch(line_prev)
+                for data in out[1:]:
+                    _, line_current = data
```
vccardenas: Similar to above comment,
    line_current = data[1]

src/resmom/mom_main.c
```diff
@@ -3730,6 +3734,46 @@ set_int(const char *id, char *value, int *var)
...
+ *
+ */
+handler_ret_t
+set_float(const char *id, char *value, float *var)
+{
+	char	*left;
+	float	val;
+
+	if (value == NULL || *value == '\0') {
+		sprintf(log_buffer, "No value specified, no action taken.");
```
bhroam: use snprintf()

src/resmom/mom_main.c
```diff
@@ -3730,6 +3734,46 @@ set_int(const char *id, char *value, int *var)
...
+		log_event(PBSEVENT_SYSTEM, PBS_EVENTCLASS_SERVER, LOG_NOTICE,
+			id, log_buffer);
+		return HANDLER_FAIL;	/* error */
+	}
+
+	val = strtod(value, &left);
+	if (left == value || val <= 0) {
+		sprintf(log_buffer, "bad value \"%s\"", value);
+		log_event(PBSEVENT_SYSTEM, 0, LOG_ERR, id, log_buffer);
```
bhroam: I'd use log_err() instead of log_event() here.

src/resmom/mom_main.c
```diff
@@ -3730,6 +3734,46 @@ set_int(const char *id, char *value, int *var)
...
+		return HANDLER_FAIL;	/* error */
+	}
+	*var = val;
+
+	sprintf(log_buffer, "setting %f", val);
```
bhroam: use snprintf().  Also you might consider putting more in this log message.  Maybe use id

src/resmom/mom_main.c
```diff
@@ -3933,23 +3977,51 @@ set_vnode_per_numa_node(char *value)
...
+static handler_ret_t
+set_alps_release_interval(char *value)
+{
+	float tmp;
+	handler_ret_t ret;
+	double fract, integ;
+	ret = set_float(__func__, value, &tmp);
+	if (ret == HANDLER_SUCCESS) {
+		fract = modf(tmp, &integ);
+		alps_release_interval = (int)integ*1000000 + (fract*1000000);
```
bhroam: Isn't this a usecond_t?

src/resmom/mom_main.c
```diff
@@ -3933,23 +3977,51 @@ set_vnode_per_numa_node(char *value)
...
- * the result of i*i is larger than alps_release_interval_max, at
- * which point PBS will only make release reservation requests
- * every alps_release_interval_max seconds on a per job basis.
+ *	PBS will randomly generate how much
+ *	microseconds to add to the alps_release_interval value
  *
- * @retval 0 failure
- * @retval 1 success
+ * @retval 0 HANDLER_FAIL
+ * @retval 1 HANDLER_SUCCESS
```
bhroam: You return HANDLER_FAIL or HABNDLER_SUCCESS, not 0 or 1.

src/resmom/mom_main.c
```diff
@@ -3933,23 +3977,51 @@ set_vnode_per_numa_node(char *value)
...
+set_alps_release_jitter(char *value)
 {
-	return (set_int(__func__, value, &alps_release_interval_max));
+	float tmp;
+	handler_ret_t ret;
+	double fract, integ;
+	ret = set_float(__func__, value, &tmp);
+	if (ret == HANDLER_SUCCESS) {
+		fract = modf(tmp, &integ);
+		alps_release_jitter = (int)integ*1000000 + (fract*1000000);
```
bhroam: Same as above, shouldn't this be a usecond_t?

test/tests/functional/pbs_alps_request_tunables.py
```diff
@@ -0,0 +1,150 @@
...
+    """
+
+    def setUp(self):
+        machine = self.du.get_platform()
+        if not machine.startswith('cray') or machine == "craysim":
+            self.skipTest("Test suite only meant to run on a Cray")
+        TestFunctional.setUp(self)
+
+    def get_epoch(self, msg):
+        a = time.strptime(msg[:19], "%m/%d/%Y %H:%M:%S")
```
bhroam: I'm not sure :19 will work.  Is the month, day, hour, and minute always 2 characters?  If not, you'll have a variable length time.  Since this is coming from a log message, you can always split on the ';'

test/tests/functional/pbs_alps_request_tunables.py
bhroam: Put a comment explaining the value.

test/tests/functional/pbs_alps_request_tunables.py
```diff
@@ -0,0 +1,150 @@
...
+        """
+        Set alps_release_interval to a higher value and then notice that
+        subsequest reservation cancellation requests are made atleast
+        after the set interval.
+        """
+        ari = 4.298
+        self.mom.add_config({'$alps_release_interval': ari})
+
+        # submit a job and then delete it after it starts running
+        start_time = time.time()
```
bhroam: You are using this in log_match().  time.time() returns a float.  You should int() it first.

test/tests/functional/pbs_alps_request_tunables.py
```diff
@@ -0,0 +1,150 @@
...
+                                 allmatch=True)
+
+        # We found something, Let's first check there are atleast 2 such
+        # log messages, If not then that means reservation was cancelled
+        # in the first attempt itself, at that point right thing to do is
+        # to either run it again or find out a way to delay the reservation
+        # cancellation at ALPS level itself.
+        if len(out) >= 2:
+            nline_prev, line_prev = out[0]
+            time_prev = self.get_epoch(line_prev)
```
bhroam: Looks like nline_prev is not used.  Also, if you only use line_prev once, just use out[0][1].

test/tests/functional/pbs_alps_request_tunables.py
```diff
@@ -0,0 +1,150 @@
...
+            for data in out[1:]:
+                nline, line_current = data
+                time_current = self.get_epoch(line_current)
```
bhroam: Same as above, use data[1]

test/tests/functional/pbs_alps_request_tunables.py
```diff
@@ -0,0 +1,150 @@
...
+                self.assertGreaterEqual(time_current - time_prev,
+                                        math.ceil(ari),
+                                        msg="alps_release_interval not working"
+                                        )
```
bhroam: Consider merging the ) onto the line above.

test/tests/functional/pbs_alps_request_tunables.py
```diff
@@ -0,0 +1,150 @@
...
+                time_prev = time_current
+
+        else:
+            self.skipTest("Reservation cancelled without retry, Try again!")
```
bhroam: Instead of skipping this test, maybe put the meat of it into a function and try running it a few times before skipping or failing the test.

test/tests/functional/pbs_alps_request_tunables.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def test_alps_release_jitter(self):
+        """
+        Set alps_release_jitter to a higher value and then notice that
+        subsequest reservation cancellation requests are made by adding
+        a random time interval (less than jitter) to alps_release_interval.
+        """
+        arj = 2.198
+        ari = 1
+        max_delay = (ari + math.ceil(2.198))
```
bhroam: Either put arj in math.ceil() or just use 3.  I dislike using functions on literals when the return value is known up front.

test/tests/functional/pbs_alps_request_tunables.py
```diff
@@ -0,0 +1,150 @@
...
+        self.mom.add_config({'$alps_release_jitter': arj})
+        self.mom.add_config({'$alps_release_interval': ari})
+        # There is no good way to test jitter and it is a random number
+        # less than value set in alps_release_jitter. So in this case
+        # we can probably try deleting a reservation a few times.
+        for each in range(5):
+            # submit a job and then delete it after it starts running
+            start_time = time.time()
```
bhroam: you probably want to int() this.

test/tests/functional/pbs_alps_request_tunables.py
```diff
@@ -0,0 +1,150 @@
...
+                                     allmatch=True)
+
+            # We found something, Let's first check there are atleast 2 such
+            # log messages, If not then that means reservation was cancelled
+            # in the first attempt itself, at that point right thing to do is
+            # to either run it again or find out a way to delay the reservation
+            # cancellation at ALPS level itself.
+            if len(out) >= 2:
+                nline_prev, line_prev = out[0]
+                time_prev = self.get_epoch(line_prev)
```
bhroam: as above, use out[0][1]

test/tests/functional/pbs_alps_request_tunables.py
bhroam: The usual reason to skip a test is if the environment you require for the test is not available.  For example, you don't have a Cray or you need 3 moms and only have one.  The problem here is that you're testing a race condition.  If the reservation is always deleted on the first try, then PBS didn't do anything wrong.  We still didn't test what we wanted to test.  This is a hard question to answer.  If we fail, it makes it sound like PBS did something wrong.  If we skip, it makes it sound like we don't have the right environment.  Neither are the case here.  I'm leaning towards skip over fail, but I don't know what the right answer is.

test/tests/functional/pbs_alps_request_tunables.py
arungrover: I do something similar to what you are saying while testing Jitter. That I do because Jitter is random in nature and I can never be certain of how long it will sleep. In this test there is minimal jitter (120 miliseconds) so it is more certain in nature. I believe ALPS cancels in first request itself then it will get cancelled in every other try as well. I'd like to retain it as is but If you still think it makes sense to retry then I'll do that.

arungrover:
@bhroam  and @vccardenas  Thanks for the review. I've addressed your comments.
@bhroam  I think you reviewed an old commit, some of the comments were resolved in the new commit. Please have a look again.
Thanks!

src/resmom/catch_child.c
```diff
@@ -2404,9 +2404,7 @@ del_job_hw(job *pjob)
 			end_time = begin_time;
 			/* add jobid to the seed */
 			srandom((unsigned)(atoi(pjob->ji_qs.ji_jobid) + begin_time));
-			for (i=1;
-				(total_time = end_time - begin_time) < alps_release_timeout;
-				++i) {
+			for (i = 1; (total_time = end_time - begin_time) < alps_release_timeout; ++i) {
```
bhroam: Don't set variables in the conditional portion of a loop.  In this case, I don't see total_time being used, so you can just remove it.

src/resmom/mom_main.c
arungrover: I actually retained the implementation from set_int and other setter functions.

src/resmom/mom_main.c
arungrover: It adds the ID as part of log_event call.

src/resmom/mom_main.c
```diff
@@ -3980,8 +3979,8 @@ set_vnode_per_numa_node(char *value)
  *	Set the alps_release_wait_time in micoseconds to wait between ALPS release
  *	reservation requests
  *
- * @retval 0 HANDLER_FAIL
- * @retval 1 HANDLER_SUCCESS
+ * @retval HANDLER_FAIL
```
bhroam: Usually the retval statements say what you are returning and when it will be returned.  In your case, it's the return value of set_float().  Maybe replace these two with "@return return value of set_float()" or something like that.

src/resmom/mom_main.c
```diff
@@ -4006,8 +4005,8 @@ set_alps_release_wait_time(char *value)
  *	PBS will randomly generate how much
  *	microseconds to add to the alps_release_wait_time value
  *
- * @retval 0 HANDLER_FAIL
- * @retval 1 HANDLER_SUCCESS
+ * @retval HANDLER_FAIL
+ * @retval HANDLER_SUCCESS
```
bhroam: Same as above, use a more descriptive return

test/tests/functional/pbs_alps_release_tunables.py
```diff
@@ -92,16 +93,14 @@ def test_alps_release_wait_time(self):
         # to either run it again or find out a way to delay the reservation
         # cancellation at ALPS level itself.
         if len(out) >= 2:
-            _, line_prev = out[0]
-            time_prev = get_epoch(line_prev)
+            time_prev = get_epoch(out[0][1])
             for data in out[1:]:
-                _, line_current = data
+                line_current = data[1]
```
bhroam: You only use line_current once, just use data[1] there instead.

src/resmom/catch_child.c
arungrover: Well, this part of the code is not something I have changed but I'll be happy to amend it if it creates a problem. I do see total time being used to log the amount of time we waited for canceling a reservation, it is outside the loop. Since the code has been there for years and has not caused any warnings or troubles I'd argue to have it as is, but, if you feel strongly about not having to set a variable in the conditional portion, I'll move it.

arungrover:
@bhroam I've addressed your comments and also made an additional change to the for loop in catch_child.c where end_time is set inside the for statement. Please have a look again.

@lisa-altair, @vccardenas Thanks for your sign-off. There has been another update to the pull, please have a look again.

vccardenas:
PTL tests still look good to me.

test/tests/functional/pbs_alps_release_tunables.py
bhroam: I don't see how it is more readable.  The only thing it provides is the name line_current.  If you want more explanation, use a comment.

src/resmom/catch_child.c
```diff
@@ -2438,7 +2437,7 @@ del_job_hw(job *pjob)
 				sprintf(log_buffer,
 					"Cancelled ALPS reservation %ld after a "
 					"total of %d tries",
-					pjob->ji_extended.ji_ext.ji_reservation, i);
+					pjob->ji_extended.ji_ext.ji_reservation, i+1);
```
bhroam: Add spaces around the '+'

src/resmom/mom_main.c
```diff
@@ -3979,8 +3979,7 @@ set_vnode_per_numa_node(char *value)
  *	Set the alps_release_wait_time in micoseconds to wait between ALPS release
  *	reservation requests
  *
- * @retval HANDLER_FAIL
- * @retval HANDLER_SUCCESS
+ * @return return value of set_float()
```
bhroam: 'return' is duplicated.  I'd remove it.

src/resmom/mom_main.c
```diff
@@ -4005,8 +4004,7 @@ set_alps_release_wait_time(char *value)
  *	PBS will randomly generate how much
  *	microseconds to add to the alps_release_wait_time value
  *
- * @retval HANDLER_FAIL
- * @retval HANDLER_SUCCESS
+ * @return return value of set_float()
```
bhroam: Same as above, remove the duplicate 'return'

src/resmom/catch_child.c
arungrover: will do

test/tests/functional/pbs_alps_release_tunables.py
arungrover: Well, you are right, a comment would have helped but I feel it is not complex enough that it would need a comment and it is also not clear enough that data[1] is the line where log_match() finds a hit. I thought the variable name does help here. 
I'd like to know if @lisa-altair and @vccardenas also think it should be removed then I'll remove it.
I personally like to be lenient with variable use in PTL script because I feel it does not save us much to have it optimized and everyone has their own personal style of dealing with things, but I may be wrong :)

vccardenas:
The PTL changes look good to me.

arungrover:
@bhroam I've squashed and rebased my branch. Thanks!

bhroam:
I merged in the PR.

PR #661, opened 2018-04-30
Mom crashes whenever server tries to delete a job in EXITING state

nithinj:
#### Bug/feature Description
* *Mom crashes whenever server tries to delete a job in EXITING state*
* Steps for reproduction:
1. Create a checkpoint_abort script

cat /tmp/test.sh

#!/bin/bash
kill $1
exit 0
2. Update the mom config file
cat config
$clienthost centos
$restrict_user_maxsysid 999
$action checkpoint_abort 30 !/tmp/test.sh %sid

3. You need to change the order preempt_order: "CSR" from the default "SCR" to checkpoint jobs.
4. HUP mom and scheduler to take effect.
5. Create an express_q
qmgr
c q express_q
s q queue_type = Execution
s q Priority = 200
s q enabled = True
s q started = True
6. submit 2 jobs
qsub -- /bin/sleep 100
qsub -q express_q -- /bin/sleep 10
7. Wait for a minute and check the state of job is 'H' after second job finishes.
8. qdel on the checkpointed job in 'H' state and you can see a mom crash.*
* Attaching pbs_checkpoint which can be used to test this. But I cannot checkin the PTL as this behavior is temporary and won't appear after the checkpoint fix.
[pbs_checkpoint.txt](https://github.com/PBSPro/pbspro/files/1959809/pbs_checkpoint.txt)

* In req_cpyfile function we call direct_write_requested() without checking if find_job was successful in finding the requested job or not.

#### Solution Description
* In direct_write_requested(), pjob is null checked before accessing elements.

#### Testing logs/output
[PTL test logs.txt](https://github.com/PBSPro/pbspro/files/1959802/PTL.test.logs.txt)

#### Checklist:

CC: @arungrover

### DIFF
--- a/src/server/job_func.c
+++ b/src/server/job_func.c
@@ -685,7 +685,7 @@ remove_stdouterr_files(job *pjob, char *suffix) {
  */
 int direct_write_requested(job *pjob) {
 	char *pj_attrk = NULL;
-	if ((pjob->ji_wattr[(int)JOB_ATR_keep].at_flags & ATR_VFLAG_SET)) {
+	if (pjob && (pjob->ji_wattr[(int)JOB_ATR_keep].at_flags & ATR_VFLAG_SET)) {
 		pj_attrk = pjob->ji_wattr[(int)JOB_ATR_keep].at_val.at_str;
 		if (strchr(pj_attrk, 'd') && (strchr(pj_attrk, 'o') || (strchr(pj_attrk, 'e'))))
 			return 1;


src/server/job_func.c
```diff
@@ -685,7 +685,7 @@ remove_stdouterr_files(job *pjob, char *suffix) {
  */
 int direct_write_requested(job *pjob) {
 	char *pj_attrk = NULL;
-	if ((pjob->ji_wattr[(int)JOB_ATR_keep].at_flags & ATR_VFLAG_SET)) {
+	if (pjob && (pjob->ji_wattr[(int)JOB_ATR_keep].at_flags & ATR_VFLAG_SET)) {
```
bhroam: While this is a fine check to add, shouldn't you really be returning after the find_job() in req_copyfile().  Is there any purpose to continue processing that function if we can't find the job?

arungrover: @bhroam This is really a corner case. I think instead of doing something special in this function we should add this check to the caller and take the decision of bailing out from there.
What do you think @nithinj?

nithinj: I think we can continue processing the stage out even if find_job didn't returned anything as rq_cpyfile structure contains stage out info.

bhroam: I'm not sure of the ramifications of the situation we'd be in.  What does it mean to the mom if she can't find the job?  While we might be able to continue with the rq_cpyfile structure, would anything in there be valid?  Would the files still exist any more?  Would Mom have cleaned up?

subhasisb: I looked at the callers of direct_write_requested(). One of them is req_cpyfile(). That functions does sections of code that is different when pjob is NULL vs not-NULL. Thus, it is clear, that some part of the copyfile processing is expected to continue even when pjob is NULL.

In the WIN32 counterpart req_cpyfile(), this function is called after a check for pjob being NON-NULL, thus:
      if (pjob != NULL && (dir == STAGE_DIR_OUT)) {
		direct_write = direct_write_requested(pjob);
	}

We could simply mimic this behavior for Linux. 

But then this function is also called from "open_std_out_err", and if this function fails, the code will go ahead to print some properties of the job as part of the error message (again access pjob).  However, this function "open_std_out_err" is called from functions where pjob CANNOT be NULL, so in effect, the caller is confident :)

The bigger issues is this: direct_write_requested() returns 0 if direct write is not requested by job, and 1 is direct write was requested. If pjob passed to it is NULL, and you handle it inside this function, what would you return? Either 0, or 1 would be a wrong return. If you add a different return value for pjob being NULL, then you anyway have to add code to the caller. 

Given all these above, I strongly suggest we put a check in the caller and do the "right" thing for that caller.

nithinj: We rely on pjob structure to determine whether direct_write is requested for the job, not rq_cpyfile. If the answer is no, or if it is something that cannot be determined the function will return false. In the worst case where the job has requested direct write and direct_write_requested function returning false will result in warning in mom log when it tries to stageout/delete stdout/err files from spool directory. This will still be the case if we handle the NULL case in the caller. But the issue being discussed is whether we should return from req_cpyfile() itself if the mom cannot find the job.

`What does it mean to the mom if she can't find the job? While we might be able to continue with the rq_cpyfile structure, would anything in there be valid? Would the files still exist any more? Would Mom have cleaned up?`
I still doesn't know the answer to these questions. I might need more time to investigate this (missing the bigger picture). I think @bayucan will take over this issue as I'm working on another one. Thanks for all the ideas!

bhroam:
Closing the PR since @bayucan is taking this over.

PR #674, opened 2018-05-08
PTL test added to check fairshare usage in TestMultipleSchedulers

Bhagat-Rajput:
#### Bug/feature Description
* Added ptl test case to write schedulers fairshare usage and also make sure that usage file is  updating correclty 

#### Affected Platform(s)
* ALL

I couldn't find the test case which verifies the read/write/update operations on schedulers fairshare usage file.

#### Solution Description
* Added a PTL test case to the TestMultipleSchedulers 

#### Testing logs/output

#### Checklist:

### DIFF
--- a/test/tests/functional/pbs_multi_sched.py
+++ b/test/tests/functional/pbs_multi_sched.py
@@ -647,6 +647,117 @@ class TestMultipleSchedulers(TestFunctional):
         self.assertEquals(n.nshares, sc3_shares)
         self.assertEquals(n.usage, sc3_usage)
 
+    @timeout(240)
+    def test_fairshare_usage(self):
+        """
+        Test the schedulers fairshare usage file and
+        check the usage file is updating correctly or not
+        """
+        self.setup_sc1()
+        a = {'queue_type': 'execution',
+             'started': 'True',
+             'enabled': 'True',
+             'partition': 'P1'}
+        self.server.manager(MGR_CMD_CREATE, QUEUE, a, id='wq1')
+        # Set resources to node
+        resc = {'resources_available.ncpus': 1,
+                'partition': 'P1'}
+        self.server.manager(MGR_CMD_SET, NODE, resc, self.mom.shortname)
+        # Add entry to the resource group of multisched 'sc1'
+        self.scheds['sc1'].add_to_resource_group('grp1', 100, 'root', 60)
+        self.scheds['sc1'].add_to_resource_group('grp2', 200, 'root', 40)
+        self.scheds['sc1'].add_to_resource_group(TEST_USER1,
+                                                 101, 'grp1', 40)
+        self.scheds['sc1'].add_to_resource_group(TEST_USER2,
+                                                 102, 'grp1', 20)
+        self.scheds['sc1'].add_to_resource_group(TEST_USER3,
+                                                 201, 'grp2', 30)
+        self.scheds['sc1'].add_to_resource_group(TEST_USER4,
+                                                 202, 'grp2', 10)
+        # Set scheduler iteration
+        sc_attr = {'scheduler_iteration': 7,
+                   'scheduling': 'False'}
+        self.server.manager(MGR_CMD_SET, SCHED, sc_attr, id='sc1')
+        # Update scheduler config file
+        sc_config = {'fair_share': 'True',
+                     'fairshare_usage_res': 'ncpus*100'}
+        self.scheds['sc1'].set_sched_config(sc_config)
+        # submit jobs to multisched 'sc1'
+        sc1_attr = {ATTR_queue: 'wq1',
+                    'Resource_List.select': '1:ncpus=1',
+                    'Resource_List.walltime': 10}
+        sc1_J1 = Job(TEST_USER1, attrs=sc1_attr)
+        sc1_jid1 = self.server.submit(sc1_J1)
+        sc1_J2 = Job(TEST_USER2, attrs=sc1_attr)
+        sc1_jid2 = self.server.submit(sc1_J2)
+        sc1_J3 = Job(TEST_USER3, attrs=sc1_attr)
+        sc1_jid3 = self.server.submit(sc1_J3)
+        self.server.manager(MGR_CMD_SET, SCHED, {'scheduling': 'True'},
+                            id='sc1')
+        # pbsuser1 job will run and other two will be queued
+        self.server.expect(JOB, {'job_state': 'R'}, id=sc1_jid1)
+        self.server.expect(JOB, {'job_state': 'Q'}, id=sc1_jid3)
+        self.server.expect(JOB, {'job_state': 'Q'}, id=sc1_jid2)
+        self.server.manager(MGR_CMD_SET, SCHED, {'scheduling': 'True'},
+                            id='sc1')
+        # pbsuser3 job will run after pbsuser1
+        self.server.expect(JOB, {'job_state': 'R'}, id=sc1_jid3)
+        self.server.expect(JOB, {'job_state': 'Q'}, id=sc1_jid2)
+        self.server.manager(MGR_CMD_SET, SCHED, {'scheduling': 'True'},
+                            id='sc1')
+        # pbsuser2 job will run in the end
+        self.server.expect(JOB, {'job_state': 'R'}, id=sc1_jid2)
+        self.server.manager(MGR_CMD_SET, SCHED, {'scheduling': 'True'},
+                            id='sc1')
+        self.server.log_match(sc1_jid2 + ";Exit_status")
+        # query fairshare and check usage
+        sc1_fs_user1 = self.scheds['sc1'].query_fairshare(name=str(TEST_USER1))
+        self.assertEquals(sc1_fs_user1.usage, 101)
+        sc1_fs_user2 = self.scheds['sc1'].query_fairshare(name=str(TEST_USER2))
+        self.assertEquals(sc1_fs_user2.usage, 101)
+        sc1_fs_user3 = self.scheds['sc1'].query_fairshare(name=str(TEST_USER3))
+        self.assertEquals(sc1_fs_user3.usage, 101)
+        sc1_fs_user4 = self.scheds['sc1'].query_fairshare(name=str(TEST_USER4))
+        self.assertEquals(sc1_fs_user4.usage, 1)
+        # Restart the scheduler
+        self.scheds['sc1'].restart()
+        # Check the multisched 'sc1' usage file whether it's updating or not
+        self.server.manager(MGR_CMD_SET, SCHED, {'scheduling': 'False'},
+                            id='sc1')
+        sc1_J1 = Job(TEST_USER1, attrs=sc1_attr)
+        sc1_jid1 = self.server.submit(sc1_J1)
+        sc1_J2 = Job(TEST_USER2, attrs=sc1_attr)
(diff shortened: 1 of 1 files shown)


Bhagat-Rajput:
Test logs has been attached.
[ptl_fairshare_usage.txt](https://github.com/PBSPro/pbspro/files/1983037/ptl_fairshare_usage.txt)

test/tests/functional/pbs_multi_sched.py
```diff
@@ -647,6 +647,268 @@ def test_multi_fairshare(self):
...
+        self.scheds['sc2'].add_to_resource_group('grp1', 100, 'root', 60)
+        self.scheds['sc2'].add_to_resource_group('grp2', 200, 'root', 40)
+        self.scheds['sc2'].add_to_resource_group('pbsuser1',
+                                                 101, 'grp1', 40)
+        self.scheds['sc2'].add_to_resource_group('pbsuser2',
+                                                 102, 'grp1', 20)
+        self.scheds['sc2'].add_to_resource_group('pbsuser3',
+                                                 201, 'grp2', 30)
+        self.scheds['sc2'].add_to_resource_group('pbsuser4',
+                                                 202, 'grp2', 10)
```
bhroam: Don't use username strings.  Pass in the PbsUser objects (e.g., TEST_USER)

test/tests/functional/pbs_multi_sched.py
```diff
@@ -647,6 +647,268 @@ def test_multi_fairshare(self):
...
+        # Set scheduler iteration
+        sc_iter = {'scheduler_iteration': 7}
+        self.server.manager(MGR_CMD_SET, SCHED, sc_iter, id='default')
+        self.server.manager(MGR_CMD_SET, SCHED, sc_iter, id='sc1')
+        self.server.manager(MGR_CMD_SET, SCHED, sc_iter, id='sc2')
+        # Update scheduler config file
+        formula = 'ceil(fabs(-ncpus*(foo/100.00)*sqrt(100)))'
+        sc_config = {'fair_share': 'True', 'fairshare_decay_time': '24:00:00',
+                     'fairshare_decay_factor': 0.5,
+                     'fairshare_usage_res': formula}
```
bhroam: fairshare_decay_time and fairshare_decay_factor default to those values.  You don't have to set them if you don't want to.

test/tests/functional/pbs_multi_sched.py
```diff
@@ -647,6 +647,268 @@ def test_multi_fairshare(self):
...
+        self.server.manager(MGR_CMD_SET, NODE, resc, self.mom.shortname)
+        self.server.manager(MGR_CMD_SET, NODE, resc, id='vnode[0]')
+        self.server.manager(MGR_CMD_SET, NODE, resc, id='vnode[1]')
+        self.server.manager(MGR_CMD_SET, NODE,
+                            {'resources_available.ncpus': 0},
+                            id='vnode[4]')
+
+        # submit jobs to default scheduler
+        def_attr = {'Resource_List.select': '1:ncpus=1:foo=20',
+                    'Resource_List.walltime': 4}
```
bhroam: A 4 second job could easily be missed by the scheduler.  Two cycles have to happen for this job for the usage to be accumulated.  The cycle that the job runs, and one subsequent cycle.  The scheduler_iteration is 7.  You might miss the second cycle.  I'd increase the walltime to 10s.

test/tests/functional/pbs_multi_sched.py
```diff
@@ -647,6 +647,268 @@ def test_multi_fairshare(self):
...
+        self.server.expect(JOB, {'job_state': 'R'}, id=def_jid3)
+        self.server.expect(JOB, {'job_state': 'Q'}, id=def_jid2)
+        # pbsuser2 job will run in the end
+        self.server.expect(JOB, {'job_state': 'R'}, id=def_jid2)
+        self.server.expect(JOB, {'job_state': 'F'}, id=def_jid2, extend='x')
+        # query fairshare and check usage
+        USER1 = str(TEST_USER1)
+        USER2 = str(TEST_USER2)
+        USER3 = str(TEST_USER3)
+        USER4 = str(TEST_USER4)
```
bhroam: Either use these throughout your test plan or just call query_fairshare() with str().  You do a combination of both.  My vote would be to use str() in query_fairshare() itself since it'd be easier to read.  You call Job() with one of the users.  This way you're using the same username.  Either that or update query_fairshare() in the framework to take a PbsUser object.  It isn't hard, just call str() on it right before its use.  That is a bit out of scope for this PR though.

test/tests/functional/pbs_multi_sched.py
```diff
@@ -647,6 +647,268 @@ def test_multi_fairshare(self):
         self.assertEquals(n.nshares, sc3_shares)
         self.assertEquals(n.usage, sc3_usage)
 
+    @timeout(300)
```
varunsonkar: Does this test really takes 300 sec? I think default is enough.

test/tests/functional/pbs_multi_sched.py
```diff
@@ -647,6 +647,268 @@ def test_multi_fairshare(self):
...
+    def test_fairshare_usage(self):
+        """
+        Test the schedulers fairshare usage file and
+        check the usage file is updating correctly or not
+        """
+        self.common_setup()
+        rv = self.server.add_resource('foo', 'float', 'nh')
+        self.assertTrue(rv)
+        self.server.manager(MGR_CMD_SET, SERVER,
+                            {'job_history_enable': 'True'}, expect=True)
```
varunsonkar: This is not required in the test case.
As we discussed remove this. Use log match to verify that the jobs has finished.

test/tests/functional/pbs_multi_sched.py
```diff
@@ -647,6 +647,268 @@ def test_multi_fairshare(self):
...
+        self.scheds['default'].add_to_resource_group('grp1', 100, 'root', 60)
+        self.scheds['default'].add_to_resource_group('grp2', 200, 'root', 40)
+        self.scheds['default'].add_to_resource_group('pbsuser1',
+                                                     101, 'grp1', 40)
+        self.scheds['default'].add_to_resource_group('pbsuser2',
+                                                     102, 'grp1', 20)
+        self.scheds['default'].add_to_resource_group('pbsuser3',
+                                                     201, 'grp2', 30)
+        self.scheds['default'].add_to_resource_group('pbsuser4',
+                                                     202, 'grp2', 10)
```
varunsonkar: You do not need to test this for default scheduler. I believe that is already being tested as part of smoke test.

test/tests/functional/pbs_multi_sched.py
```diff
@@ -647,6 +647,268 @@ def test_multi_fairshare(self):
...
+                                                 101, 'grp1', 40)
+        self.scheds['sc1'].add_to_resource_group('pbsuser2',
+                                                 102, 'grp1', 20)
+        self.scheds['sc1'].add_to_resource_group('pbsuser3',
+                                                 201, 'grp2', 30)
+        self.scheds['sc1'].add_to_resource_group('pbsuser4',
+                                                 202, 'grp2', 10)
+        # Add entry to the resource group of multisched 'sc2'
+        #(sched_priv_sc2 at non-default location)
+        self.scheds['sc2'].add_to_resource_group('grp1', 100, 'root', 60)
```
varunsonkar: I think testing for non-default location of multisched priv dir exclusively for fairshare scenario is not required if similar kind of test exists that checks the multiched functionality where priv dir is not at default location.
If there is any test case that verifies functionality when the changes made to multisched existing in non default location. Then this verification of ['sc2'] is not required.

test/tests/functional/pbs_multi_sched.py
```diff
@@ -647,6 +647,268 @@ def test_multi_fairshare(self):
...
+        # query fairshare and check usage
+        sc2_fs_user1 = self.scheds['sc2'].query_fairshare(name=str(TEST_USER1))
+        self.assertEquals(sc2_fs_user1.usage, 5)
+        sc2_fs_user2 = self.scheds['sc2'].query_fairshare(name=str(TEST_USER2))
+        self.assertEquals(sc2_fs_user2.usage, 5)
+        sc2_fs_user3 = self.scheds['sc2'].query_fairshare(name=str(TEST_USER3))
+        self.assertEquals(sc2_fs_user3.usage, 3)
+        sc2_fs_user4 = self.scheds['sc2'].query_fairshare(name=str(TEST_USER4))
+        self.assertEquals(sc2_fs_user4.usage, 3)
+
```
varunsonkar: I will recommend you to write one more test case that verifies the usage value remains same before and after the restart of multisched.

Bhagat-Rajput: Ok, do i need to write a new test case or just add restart code into this testcase itself.

varunsonkar: I think adding restart code and verifying usage after the restart can be added here. No need to write separate test case.

test/tests/functional/pbs_multi_sched.py
Bhagat-Rajput: Yes. we are checking functionality for non-default location of multisched in most of the test cases but there is no testcase in which checking the read/update operations on the fairshare usage file. Would you still recommend to remove the code ?

test/tests/functional/pbs_multi_sched.py
Bhagat-Rajput: Thanks @varunsonkar i will add here

test/tests/functional/pbs_multi_sched.py
```diff
@@ -647,6 +647,268 @@ def test_multi_fairshare(self):
         self.assertEquals(n.nshares, sc3_shares)
         self.assertEquals(n.usage, sc3_usage)
 
+    @timeout(300)
+    def test_fairshare_usage(self):
+        """
+        Test the schedulers fairshare usage file and
+        check the usage file is updating correctly or not
+        """
+        self.common_setup()
```
suresh-thelkar: Don't use common_setup(). This creates all the schedulers sc1,sc2 and sc3 and its set-up. Instead use only the schedulers that are needed and use its corresponding setup_scx() functions.

test/tests/functional/pbs_multi_sched.py
suresh-thelkar: I agree with @varunsonkar. No need to use the custom resource "foo" here.

test/tests/functional/pbs_multi_sched.py
suresh-thelkar: Yes. No need to test default scheduler's fairshare here.

test/tests/functional/pbs_multi_sched.py
varunsonkar: Yeah remove it. The code will always refer to PBS_HOME for referring the location of usage file. So similar tests are already covered as you said for checking the functionality for non-default location of multisched where they look for non-default PBS_HOME.
No need to specifically check for fairshare functinality for non-default PBS_HOME

Bhagat-Rajput:
Thanks to all, I have updated the test case based on your suggestions.Please review it.

Bhagat-Rajput:
Thanks to all for signoff, squashed and rebased so please take it further.

bhroam:
Looks like you are out of date again.  Please rebase again.

Bhagat-Rajput:
Rebased

bhroam:
I merged in the PR.

PR #693, opened 2018-05-16
PP-261: Microsecond Logging

bhroam:
#### Bug/feature Description
* This PR implements the RFE for microsecond logging.  
*  [PP-261](https://pbspro.atlassian.net/browse/PP-261)
* PTL bug where the dshutils functions pbs_set_config(), pbs_unset_config(), and pbs_set_environment() would only work if the test was run as root.  Plus there wasn't pbs_unset_environment(), so that was implemented.
* Fixed test pbs_init_script.py to be able to be run as non-root.

* Design: https://pbspro.atlassian.net/wiki/spaces/PD/pages/50836942/PP-261+micro-second+time+stamp+for+daemon+logging
* Discussion: http://community.pbspro.org/t/pp-261-micro-second-time-stamp-for-daemon-logging/467

#### Solution Description
* PTL bug: The root cause was that all the functions called an internal function called _set_file().  _set_file() would call run_copy() without sudo=True.  I fixed it by adding a sudo argument to _set_file() which is passed to run_copy().  In all the affected functions, I pass sudo=True.
* Test bug: The test didn't call run_cmd() with sudo=True.  This meant the init script wouldn't be able to stop/start PBS.  Also, the env argument to run_cmd() won't work with the sudo argument.  It will pass the environmental variables to the sudo process which strips them from the real command.  I had to pass the environmental variables on the command line.
#### Testing logs/output
*  [ptl.log](https://github.com/PBSPro/pbspro/files/2033238/ptl.log)
* Other tests that use affected functions:  [ptl2.log](https://github.com/PBSPro/pbspro/files/2011259/ptl2.log)

#### Checklist:

- [X] My change requires project documentation. See **[required documentation checklist](https://pbspro.atlassian.net/wiki/display/DG/Checklist+for+Developing+Features+and+Bug+Fixes)** for details.
   - [X] I have added documentation in the **[project documentation area](https://pbspro.atlassian.net/wiki/display/PD)**.

### DIFF
--- a/src/lib/Libifl/pbs_loadconf.c
+++ b/src/lib/Libifl/pbs_loadconf.c
@@ -121,9 +121,10 @@ struct pbs_config pbs_conf = {
 	NULL,					/* for router, default communication routers list */
 	0,					/* default comm logevent mask */
 	4,					/* default number of threads */
-	NULL					/* mom short name override */
+	NULL,					/* mom short name override */
+	0					/* high resolution timestamp logging */
 #ifdef WIN32
-	,NULL					/* remote viewer launcher executable alongwith launch options */
+	,NULL					/* remote viewer launcher executable along with launch options */
 #endif
 };
 
@@ -564,6 +565,10 @@ pbs_loadconf(int reload)
 				free(pbs_conf.pbs_mom_node_name);
 				pbs_conf.pbs_mom_node_name = strdup(conf_value);
 			}
+			else if (!strcmp(conf_name, PBS_CONF_LOG_HIGHRES_TIMESTAMP)) {
+				if (sscanf(conf_value, "%u", &uvalue) == 1)
+					pbs_conf.pbs_log_highres_timestamp = ((uvalue > 0) ? 1 : 0);
+			}
 #ifdef WIN32
 			else if (!strcmp(conf_name, PBS_CONF_REMOTE_VIEWER)) {
 				free(pbs_conf.pbs_conf_remote_viewer);
@@ -790,6 +795,10 @@ pbs_loadconf(int reload)
 	if ((gvalue = getenv(PBS_CONF_ENVIRONMENT)) != NULL) {
 		free(pbs_conf.pbs_environment);
 		pbs_conf.pbs_environment = shorten_and_cleanup_path(gvalue);
+	}
+	if ((gvalue = getenv(PBS_CONF_LOG_HIGHRES_TIMESTAMP)) != NULL) {
+		if (sscanf(gvalue, "%u", &uvalue) == 1)
+			pbs_conf.pbs_log_highres_timestamp = ((uvalue > 0) ? 1 : 0);
 	}
 
 #ifdef WIN32
--- a/src/lib/Liblog/pbs_log.c
+++ b/src/lib/Liblog/pbs_log.c
@@ -61,6 +61,7 @@
 #include <sys/stat.h>
 #include <sys/param.h>
 #include <sys/types.h>
+#include <sys/time.h>
 #include <limits.h>
 #include <time.h>
 #include <fcntl.h>
@@ -130,6 +131,41 @@ static char *class_names[] = {
 	"TPP"
 };
 
+#ifdef WIN32
+/**
+ * @brief
+ *		gettimeofday - This function returns the current calendar
+ *		time as the elapsed time since the epoch in the struct timeval
+ *		structure indicated by tp
+ *
+ * @param[in] - tp - pointer to timeval struct
+ * @param[in] - tzp - pointer to timezone struct (not used)
+ * @return int
+ * @retval -1 - failure
+ * @retval 0 - success
+ */
+int
+gettimeofday(struct timeval *tp, struct timezone *tzp)
+{
+	FILETIME file_time = {0};
+	ULARGE_INTEGER large_int = {0};
+	/*
+ 	 * Microsecond different from "January 1, 1601 (UTC)" to
+  	 * "00:00:00 January 1, 1970" as Windows's FILESYSTEM is represents from
+ 	 * "January 1, 1601 (UTC)"
+  	 */
+	static const unsigned __int64 epoch = 116444736000000000ULL;
+
+	GetSystemTimeAsFileTime(&file_time);
+	large_int.LowPart = file_time.dwLowDateTime;
+	large_int.HighPart = file_time.dwHighDateTime;
+	tp->tv_sec = (time_t)((large_int.QuadPart - epoch) / 10000000L);
+	tp->tv_usec = (time_t)((large_int.QuadPart - epoch) % 1000000L);
+	return 0;
+}
+#endif
+
+
 /* External functions called */
 
 /**
@@ -142,7 +178,8 @@ static char *class_names[] = {
  * @retval 0 - success
  */
 
-int set_msgdaemonname(char *ch)
+int 
+set_msgdaemonname(char *ch)
 {
 	if(!(msg_daemonname = strdup(ch))) {
 		return 1;
@@ -157,7 +194,8 @@ int set_msgdaemonname(char *ch)
  * @return void
  */
 
-void set_logfile(FILE *fp)
+void 
+set_logfile(FILE *fp)
 {
 	log_opened = 1;
 	logfile = fp;
@@ -181,7 +219,9 @@ void set_logfile(FILE *fp)
 static char *
 mk_log_name(char *pbuf, size_t pbufsz)
 {
+#ifndef WIN32
 	struct tm ltm;
+#endif
 	struct tm *ptm;
 	time_t time_now;
 
@@ -833,13 +873,16 @@ log_suspect_file(const char *func, const char *text, const char *file, struct st
 void
 log_record(int eventtype, int objclass, int sev, const char *objname, const char *text)
 {
-	time_t now;
+	time_t now = 0;
 	struct tm *ptm;
+#ifndef WIN32
 	struct tm ltm;
+#endif
 	int    rc = 0;
 	FILE  *savlog;
 	static char slogbuf[LOG_BUF_SIZE];
-
+	struct timeval tp;
+	char microsec_buf[8] = {0};
 
 #if SYSLOG
 	if (syslogopen != 0) {
@@ -858,7 +901,13 @@ log_record(int eventtype, int objclass, int sev, const char *objname, const char
 	if ((text == NULL) || (objname == NULL))
 		return;
 
-	now = time(NULL);	/* get time for message */
+	/* if gettimeofday() fails, log messages will be printed at the epoch */
+	if (gettimeofday(&tp, NULL) != -1) {
+		now = tp.tv_sec;
+
+		if (pbs_conf.pbs_log_highres_timestamp)
+			snprintf(microsec_buf, sizeof(microsec_buf), ".%06ld", (long)tp.tv_usec);
+	}
 
 #ifdef WIN32
 	ptm = localtime(&now);
@@ -889,14 +938,13 @@ log_record(int eventtype, int objclass, int sev, const char *objname, const char
 
 	if (pbs_conf.locallog != 0 || pbs_conf.syslogfac == 0) {
 		rc = fprintf(logfile,
-			"%02d/%02d/%04d %02d:%02d:%02d;%04x;%s;%s;%s;%s\n",
-			ptm->tm_mon+1, ptm->tm_mday, ptm->tm_year+1900,
-			ptm->tm_hour, ptm->tm_min, ptm->tm_sec,
-			eventtype & ~PBSEVENT_FORCE,
-			msg_daemonname,
-			class_names[objclass],
-			objname,
-			text);
+			     "%02d/%02d/%04d %02d:%02d:%02d%s;%04x;%s;%s;%s;%s\n",
+			     ptm->tm_mon + 1, ptm->tm_mday, ptm->tm_year + 1900,
+			     ptm->tm_hour, ptm->tm_min, ptm->tm_sec, microsec_buf,
+			     eventtype & ~PBSEVENT_FORCE, msg_daemonname,
+			     class_names[objclass], objname, text);
+
+
 
 		(void)fflush(logfile);
 		if (rc < 0) {
--- a/src/tools/tracejob.c
+++ b/src/tools/tracejob.c
@@ -82,6 +82,7 @@ const char *mid_path[] = {"server_priv/accounting", "server_logs", "mom_logs",
 struct log_entry *log_lines;
 int ll_cur_amm;
 int ll_max_amm;
+int has_high_res_timestamp = 0;
 
 static char none[1] = { '\0' };
 /**
@@ -337,8 +338,14 @@ main(int argc, char *argv[])
 				else
 					event_type = strtol(log_lines[i].event, &endp, 16);
 				if (!(log_filter & event_type) && !(log_lines[i].no_print)) {
-					printf("%-20s %-5c", log_lines[i].date, log_lines[i].log_file);
-					line_wrap(log_lines[i].msg, 26, wrap);
+					if (has_high_res_timestamp) {
+						printf("%-27s %-5c", log_lines[i].date, log_lines[i].log_file);
+						line_wrap(log_lines[i].msg, 33, wrap);
+					} else {
+						printf("%-20s %-5c", log_lines[i].date, log_lines[i].log_file);
+						line_wrap(log_lines[i].msg, 26, wrap);
+					}
+
 				}
 			}
 		}
@@ -349,10 +356,10 @@ main(int argc, char *argv[])
 			 */
 			unknw_job = 1;
 			if (strchr(argv[opt], '.') == NULL)
-			  fprintf(stderr, "\ntracejob: Couldn't find Job Id %s.%s in logs of past %d day%s\n\n", argv[opt],
+				fprintf(stderr, "\ntracejob: Couldn't find Job Id %s.%s in logs of past %d day%s\n\n", argv[opt],
 					pbs_conf.pbs_server_name, number_of_days, number_of_days == 1 ? "" : "s");
 			else
-			  fprintf(stderr, "\ntracejob: Couldn't find Job Id %s in logs of past %d day%s\n\n", argv[opt],
+				fprintf(stderr, "\ntracejob: Couldn't find Job Id %s in logs of past %d day%s\n\n", argv[opt],
 					number_of_days, number_of_days == 1 ? "" : "s");
 		}
 	}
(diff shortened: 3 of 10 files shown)


bhroam:
For some reason my pep8 and travis's pep8 report different things.  My pep8 reports cleanly, and travis fails.  This is causing a lot of commits.

bhroam:
Since no one had started my review yet, I just squashed all the pep8 fixing commits to make it easier to review.

src/tools/tracejob.c
```diff
@@ -481,13 +488,35 @@ parse_log(FILE *fp, char *job, int ind)
...
+ 						 * in the range 0 to 11 for mktime()
+ 						 */
+						tms.tm_mon--;
+						log_lines[ll_cur_amm].date_time = mktime(&tms);
+					}
+				} else {
+					if (sscanf(tmp.date, "%d/%d/%d %d:%d:%d", &tms.tm_mon, &tms.tm_mday, 
+					    &tms.tm_year, &tms.tm_hour, &tms.tm_min, &tms.tm_sec) != 6) {
+						log_lines[ll_cur_amm].date_time = -1;	/* error in date field */
+						log_lines[ll_cur_amm].highres_ts = -1;
```
agrawalravi90: pull this line outside of if-else as it is common to both

src/tools/tracejob.c
```diff
@@ -481,13 +488,35 @@ parse_log(FILE *fp, char *job, int ind)
...
+					} else {
+						has_high_res_timestamp = 1;
+						if (tms.tm_year > 1900)
+							tms.tm_year -= 1900;
+						/* The number of months since January, 
+ 						 * in the range 0 to 11 for mktime()
+ 						 */
+						tms.tm_mon--;
+						log_lines[ll_cur_amm].date_time = mktime(&tms);
```
agrawalravi90: Don;t you need to set 'log_lines[ll_cur_amm].highres_ts' here? I don't know what it's used for, but it's being set in other cases, so just checking

src/tools/tracejob.c
```diff
@@ -481,13 +488,35 @@ parse_log(FILE *fp, char *job, int ind)
...
-					tms.tm_mon--;         /* The number of months since January, in the range 0 to 11 for mktime */
-					log_lines[ll_cur_amm].date_time = mktime(&tms);
+				if ((ind != IND_ACCT) && (strchr(tmp.date, '.'))) {
+					if (sscanf(tmp.date, "%d/%d/%d %d:%d:%d.%ld", &tms.tm_mon, 
+					    &tms.tm_mday, &tms.tm_year, &tms.tm_hour, &tms.tm_min, 
+					    &tms.tm_sec, &(log_lines[ll_cur_amm].highres_ts)) != 7) {
+						log_lines[ll_cur_amm].date_time = -1;	/* error in date field */
+						log_lines[ll_cur_amm].highres_ts = -1;
+					} else {
+						has_high_res_timestamp = 1;
```
agrawalravi90: We have macros for True and False right? can you please use them instead of 1 and 0 literals?

src/tools/tracejob.c
agrawalravi90: Does it make sense to make -1  a macro?

src/tools/tracejob.c
```diff
@@ -563,6 +592,13 @@ sort_by_date(const void *v1, const void *v2)
 	else if (l1->date_time > l2->date_time)
 		return 1;
 	else {
+		if ((l1->highres_ts != -1) && (l2->highres_ts != -1)) {
```
agrawalravi90: same comment here, does it make sense to make -1 a macro?

test/fw/ptl/utils/pbs_dshutils.py
```diff
@@ -280,20 +285,33 @@ def _set_file(self, hostname, fin, fout, append, vars):
...
             self.chmod(path=fn, mode=0644)
             with open(fn, 'w') as fd:
                 for k, v in conf.items():
                     fd.write(str(k) + '=' + str(v) + '\n')
-            self.run_copy(hostname, fn, fout, level=logging.DEBUG2)
-            self.rm(path=fn)
+            rv = self.run_copy(hostname, fn, fout, uid=user, gid=group,
+                               level=logging.DEBUG2, sudo=sudo)
+            if rv['rc'] != 0:
+                raise PbsConfigError
```
agrawalravi90: Would it be better to pass an error message to PbsConfigError saying that the copy failed?

src/lib/Libifl/pbs_loadconf.c
```diff
@@ -564,6 +565,10 @@ pbs_loadconf(int reload)
 				free(pbs_conf.pbs_mom_node_name);
 				pbs_conf.pbs_mom_node_name = strdup(conf_value);
 			}
+			else if (!strcmp(conf_name, PBS_CONF_LOG_HIGHRES_TIMESTAMP)) {
+				if (sscanf(conf_value, "%u", &uvalue) == 1)
+					pbs_conf.pbs_log_highres_timestamp = ((uvalue > 0) ? 1 : 0);
```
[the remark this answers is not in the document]
bhroam: It's for consistency.  pbs.conf booleans are 1 or 0.  This is how the code works for all of them.

test/fw/ptl/utils/pbs_dshutils.py
bhroam: It doesn't matter.  This is just really to trigger the try/except.  The except block does have an error message.  If I put one here, it would just be ignored.

src/tools/tracejob.c
bhroam: This is the case where highres_ts was set by the sscanf().  The other cases are either an error in the sscanf() or no microseconds were detected.  Those need to have it set to -1.

src/tools/tracejob.c
bhroam: It appears we don't have global macros for true and false.  There are some localized macros in different header files, but none of them are appropriate for tracejob to include.

src/tools/tracejob.c
bhroam: Probably.  It's set to -1 up front and then the -1 is checked as a value.  If we set it to -1 and then checked >= 0 that'd be one thing.  Since we are specifically looking for -1, we should use a macro.

src/lib/Liblog/pbs_log.c
```diff
@@ -858,7 +910,14 @@ log_record(int eventtype, int objclass, int sev, const char *objname, const char
 	if ((text == NULL) || (objname == NULL))
 		return;
 
-	now = time(NULL);	/* get time for message */
+	if (pbs_conf.pbs_log_highres_timestamp && (gettimeofday(&tp, NULL) != -1)) {
+		now = tp.tv_sec;
+		microsec = tp.tv_usec;
+	} else {
+		now = time((time_t *)0);	/* get time for message */
+		microsec = 0;
```
arungrover: You don't have to set it to 0 in this case. But, even if you keep it, I'm ok with it.

src/tools/tracejob.h
```diff
@@ -94,6 +94,7 @@ struct log_entry
 {
 	char *date;           /* date of log entry */
 	time_t date_time;     /* number of seconds from the epoch to date */
+	long highres_ts;  /* high resolution time stamp */
```
arungrover: Please indent the comment. I think the variable name is misleading here. When we parse the log message microseconds are read into highres_ts which is part of a timestamp and not really a timestamp by itself.

test/fw/ptl/utils/pbs_dshutils.py
```diff
@@ -449,7 +472,46 @@ def set_pbs_environment(self, hostname=None,
...
+    def unset_pbs_environment(self, hostname=None,
+                              fin='/var/spool/pbs/pbs_environment', fout=None,
+                              environ=None):
+        """
+        Unset environment variables in a pbs_environment file
+
+        :param hostname: the name of the host on which to operate
+        :type hostname: str or None
+        :param fin: the input pbs.conf file
+        :type fin: str or None
```
arungrover: Default is not None, please change the comment

test/tests/functional/pbs_highreslog.py
```diff
@@ -0,0 +1,153 @@
...
+                    ok = False
+        _msg = 'Not Found high resolution time stamp in log'
+        self.assertTrue(ok, _msg)
+        _msg = self.server.shortname + \
+            ': High resolution time stamp found in log'
+        self.logger.info(_msg)
+
+    def tearDown(self):
+        confs = self.du.parse_pbs_config()
+        if 'PBS_LOG_HIGHRES_TIMESTAMP' in confs:
```
arungrover: Shouldn't this be added as part of default setup? why are you removing it here?

test/tests/functional/pbs_highreslog.py
```diff
@@ -0,0 +1,153 @@
...
+# under a commercial license agreement.
+#
+# Use of Altair’s trademarks, including but not limited to "PBS™",
+# "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
+# trademark licensing policies.
+
+from tests.functional import *
+
+
+class HighResLoggingTestSuite(TestFunctional):
```
arungrover: According to guideline test suites start with "Test" keyword

test/fw/ptl/utils/pbs_dshutils.py
```diff
@@ -356,18 +374,20 @@ def set_pbs_config(self, hostname=None, fin=None, fout=None,
...
         :param confs: The ``key/value`` pairs to create
-        :type confs: Dictionary
+        :type confs: Dictionary or None
         """
         if fin is None:
             fin = self.get_pbs_conf_file(hostname)
         if fout is None and fin is not None:
             fout = fin
-        if confs:
+        if confs is not None:
```
minghui-liu: `{}` is also not equal to `None`. Is this what you want?

test/tests/functional/pbs_highreslog.py
```diff
@@ -0,0 +1,153 @@
...
+
```
minghui-liu: Please remove the empty line here.

test/tests/functional/pbs_highreslog.py
bhroam: The problem is that pbs.conf is not reverted to default in the default setUp().  It needs to be removed here.

test/fw/ptl/utils/pbs_dshutils.py
bhroam: It is what I want.  I changed the default value of confs from {} to None because dictionaries are mutable types.  Originally, if confs was defaulted and it was changed, that change would persist through all calls to set_pbs_config().  By defaulting it to None and then setting it to an empty dictionary in the code, you avoid this.

bhroam:
I've addressed all comments except for the ones in the windows gettimeofday() function.  I didn't write it, so I'm a little scared to touch it.  I have emailed the people who did write it for assistance.  I'll address them in my next commit.

test/tests/functional/pbs_highreslog.py
```diff
@@ -98,17 +97,15 @@ def test_basic(self):
         self.du.set_pbs_config(confs=a, append=True)
         _msg = 'Failed to restart server: %s' % (self.server.shortname)
         self.assertTrue(self.server.restart(), _msg)
+        print self.du.parse_pbs_config()
```
minghui-liu: Please use log instead of printing directly.

src/lib/Liblog/pbs_log.c
```diff
@@ -910,12 +910,13 @@ log_record(int eventtype, int objclass, int sev, const char *objname, const char
...
 
-	if (pbs_conf.pbs_log_highres_timestamp && (gettimeofday(&tp, NULL) != -1)) {
+	if (gettimeofday(&tp, NULL) != -1) {
 		now = tp.tv_sec;
-		snprintf(microsec_buf, sizeof(microsec_buf), ".%ld", (long)tp.tv_usec);
-	} else
-		now = time((time_t *)0);	/* get time for message */
 
+	if (pbs_conf.pbs_log_highres_timestamp)
+		snprintf(microsec_buf, sizeof(microsec_buf), ".%0ld", (long)tp.tv_usec);
```
minghui-liu: The indentation here looks off.

test/tests/functional/pbs_highreslog.py
bhroam: Or I can just delete it.  It's a debug message I forgot to remove :)

src/lib/Liblog/pbs_log.c
bhroam: It's two tabs.  I think it is correct.

src/lib/Liblog/pbs_log.c
```diff
@@ -858,7 +901,13 @@ log_record(int eventtype, int objclass, int sev, const char *objname, const char
 	if ((text == NULL) || (objname == NULL))
 		return;
 
-	now = time(NULL);	/* get time for message */
+	if (gettimeofday(&tp, NULL) != -1) {
+		now = tp.tv_sec;
+
+	if (pbs_conf.pbs_log_highres_timestamp)
```
agrawalravi90: The indentation seems off here

src/lib/Liblog/pbs_log.c
```diff
@@ -858,7 +901,13 @@ log_record(int eventtype, int objclass, int sev, const char *objname, const char
...
+		snprintf(microsec_buf, sizeof(microsec_buf), ".%06ld", (long)tp.tv_usec);
+	} else
+		now = time(NULL);
```
agrawalravi90: you do error checking for gettimeofday() but not for time(), why? I feel like we should just bail if gettimeofday fails.

agrawalravi90:
Could you also fix Travis/Appveyor warnings? they've been failing for the last few commits.

bhroam:
Appvetor is failing due to an initialization issue.  It can't download one of its dependencies.  Hopefully it'll pass on my next push.

minghui-liu:
I restarted your travis and appveyor build.

bhroam:
OK I fixed the indentation and the pep8 issue.  As for gettimeofday() failing, instead of bailing and not printing the log message, I'll just print it at the epoch (Jan 1 1970 GMT).

test/tests/functional/pbs_highreslog.py
```diff
@@ -0,0 +1,148 @@
...
+        j = Job(TEST_USER)
+        jid = self.server.submit(j)
+        self.server.expect(JOB, {ATTR_state: 'R'}, id=jid)
+        lines = self.server.log_lines(logtype=self.server,
+                                      starttime=self.server.ctime)
+        ok = True
+        for line in lines:
+            m = self.tm_micro_re.match(line)
+            if m is not None:
+                ok = False
```
agrawalravi90: why not break from here?

test/tests/functional/pbs_highreslog.py
```diff
@@ -0,0 +1,148 @@
...
+        _msg = 'Found high resolution time stamp in log,' \
+               ' it shouldn\'t be there'
+        self.assertTrue(ok, _msg)
+        _msg = self.server.shortname + \
+            ': High resolution time stamp correctly not in log'
```
agrawalravi90: Did you mean "High resolution time stamp correctly not **set** in log"?

test/tests/functional/pbs_highreslog.py
```diff
@@ -0,0 +1,148 @@
...
+        self.server.expect(JOB, {ATTR_state: 'R'}, id=jid)
+        lines = self.server.log_lines(logtype='tracejob', id=jid, n='ALL')
+        if len(lines) > 1:
+            # Remove first line as it is not log line
+            lines = lines[1:]
+        ok = True
+        for line in lines:
+            m = self.tm_micro_re.match(line)
+            if m is not None:
+                ok = False
```
agrawalravi90: same here, shouldn't you break out?

test/tests/functional/pbs_highreslog.py
```diff
@@ -0,0 +1,148 @@
...
+        _msg = 'Found high resolution time stamp in tracejob, ' \
+            'it should not be there'
+        self.assertTrue(ok, _msg)
+        _msg = self.server.shortname + \
+            ': High resolution time stamp correctly not in tracejob output'
```
agrawalravi90: Again, did you mean to say "High resolution time stamp correctly not **set** in tracejob output" ?

test/tests/functional/pbs_init_script.py
```diff
@@ -43,23 +43,28 @@ class TestPbsInitScript(TestFunctional):
...
         Test PBS_START environment variables overrides values in pbs.conf file
         """
-        self.du.run_cmd(cmd=['/etc/init.d/pbs', 'stop'])
-
         conf = {'PBS_START_SERVER': '1', 'PBS_START_SCHED': '1',
-                'PBS_START_COMM': '1', 'PBS_START_MOM': '0'}
+                'PBS_START_COMM': '1', 'PBS_START_MOM': '1'}
+        self.du.set_pbs_config(confs=conf)
+
+        self.du.run_cmd(cmd=['/etc/init.d/pbs', 'stop'], sudo=True)
```
agrawalravi90: why didn't you use PBSInitServices().stop() instead?

test/tests/functional/pbs_init_script.py
```diff
@@ -43,23 +43,28 @@ class TestPbsInitScript(TestFunctional):
...
+
+        conf['START_PBS_MOM'] = '0'
         self.du.set_pbs_config(confs=conf)
 
-        env = {'PBS_START_SERVER': '0', 'PBS_START_SCHED': '0',
-               'PBS_START_COMM': '0', 'PBS_START_MOM': '1'}
+        cmd = ['PBS_START_SERVER=0', 'PBS_START_SCHED=0',
+               'PBS_START_COMM=0', 'PBS_START_MOM=1',
+               '/etc/init.d/pbs', 'start']
```
agrawalravi90: Please use os.sep and os.path.join() to create the path

test/tests/functional/pbs_init_script.py
```diff
@@ -43,23 +43,28 @@ class TestPbsInitScript(TestFunctional):
...
-        self.du.run_cmd(cmd=['/etc/init.d/pbs', 'stop'])
-
         conf = {'PBS_START_SERVER': '1', 'PBS_START_SCHED': '1',
-                'PBS_START_COMM': '1', 'PBS_START_MOM': '0'}
+                'PBS_START_COMM': '1', 'PBS_START_MOM': '1'}
+        self.du.set_pbs_config(confs=conf)
+
+        self.du.run_cmd(cmd=['/etc/init.d/pbs', 'stop'], sudo=True)
+
+        conf['START_PBS_MOM'] = '0'
```
agrawalravi90: Is START_PBS_MOM a typo? should it be PBS_START_MOM?

test/tests/selftest/pbs_dshutils_tests.py
```diff
@@ -0,0 +1,85 @@
...
+#
+# Use of Altair’s trademarks, including but not limited to "PBS™",
+# "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
+# trademark licensing policies.
+
+from tests.selftest import *
+from ptl.utils.pbs_dshutils import PbsConfigError
+
+
+class TestDshUtils(TestSelf):
```
agrawalravi90: Thanks for writing these unit tests! I hope this inspires others to do the same.

test/tests/functional/pbs_highreslog.py
bhroam: I actually just turned this into a self.assertIsNone().  No need to break any more since the test will fail if the assert fails

test/tests/functional/pbs_highreslog.py
bhroam: Thanks for pointing this out.  I fixed it.

test/tests/functional/pbs_highreslog.py
bhroam: Same as above, I replaced this with a self.assertIsNone()

test/tests/functional/pbs_highreslog.py
bhroam: Yep, same typo.  Thanks for pointing it out.

test/tests/functional/pbs_init_script.py
bhroam: I don't want PBSInitServices() to muck with pbs.conf.  The whole point of this test suite is to test if environmental variables override pbs.conf variables.  I don't think PBSInitServices() will muck with pbs.conf unless you only stop one daemon, but I don't want to take the chance.

test/tests/functional/pbs_init_script.py
bhroam: Good point.  I'll update this to use os.path.join()

test/tests/functional/pbs_init_script.py
bhroam: Good catch.  I tried to figure out why this test passed with this typo, and then I realized that PBS_START_MOM=1 from pbs.conf and worked fine by not being overridden.  I've corrected the typo.

agrawalravi90:
Thanks for posting the latest logs, you have my approval.

agrawalravi90:
@arungrover could you please sign-off on the latest changes as well so that this can be merged?

agrawalravi90:
@bhroam please rebase for the merge

PR #694, opened 2018-05-17
Fix mom crash when deleting job in exiting state.

bayucan:
#### Bug/feature Description
* Mom crashes whenever server tries to delete a job in EXITING state
* Steps for reproduction:

1. Create a checkpoint_abort script
cat /tmp/test.sh 
#!/bin/bash
kill $1
exit 0
2. Update the mom config file
cat config
$clienthost centos
$restrict_user_maxsysid 999
$action checkpoint_abort 30 !/tmp/test.sh %sid 
3. You need to change the order preempt_order: "CSR" from the default "SCR" to checkpoint jobs.
4. HUP mom and scheduler to take effect.
5. Create an express_q
 qmgr
c q express_q
s q queue_type = Execution
s q Priority = 200
s q enabled = True
 s q started = True
6. submit 2 jobs
qsub -- /bin/sleep 100
qsub -q express_q -- /bin/sleep 10
7. Wait for a minute and check the state of job is 'H' after second job finishes.
8. qdel on the checkpointed job in 'H' state and you can see a mom crash.*

#### Affected Platform(s)
* Linux

* In req_cpyfile function we call direct_write_requested() without checking if find_job was successful in finding the requested job or not.

#### Solution Description
* In resmom/requests.c:req_cpyfile() function, check pjob is not NULL first and is doing STAGE_DIR_OUT before calling direct_write_requested(). This essentially does what is in the existing Windows portion of the code. Note that pjob can be NULL under req_cpyfile() when job is doing STAGE_DIR_IN.

#### Testing logs/output
* PTL test, pbs_checkpoint.py, has been added and here are the results:
[ptl.checkpoint.PASS.txt](https://github.com/PBSPro/pbspro/files/2014928/ptl.checkpoint.PASS.txt)
[ptl.checkpoint.FAIL.txt](https://github.com/PBSPro/pbspro/files/2014929/ptl.checkpoint.FAIL.txt)
* The code fix is in the path of direct_write_requested() so regression test of the direct write feature has been run successfully:
[ptl.direct_write.txt](https://github.com/PBSPro/pbspro/files/2014931/ptl.direct_write.txt)

#### Checklist:

Manual test is done as it's not possible to create a PTL test that can reproduce a bug, that is dependent on a buggy checkpoint abort behavior (where job does not get properly checkpointed) that will soon also get fixed. Here is the PTL test that can be run manually but should not be checked-in as it is written in a way that is dependent on existing checkpoint buggy behavior:
[pbs_checkpoint.py.txt](https://github.com/PBSPro/pbspro/files/2028538/pbs_checkpoint.py.txt)
[ptl.pbs_checkpoint.FAIL.2.txt](https://github.com/PBSPro/pbspro/files/2028539/ptl.pbs_checkpoint.FAIL.2.txt)
[ptl.pbs_checkpoint.PASS.2.txt](https://github.com/PBSPro/pbspro/files/2028540/ptl.pbs_checkpoint.PASS.2.txt)

- [] All new and existing automated tests have passed. (See **[running automated PTL tests](https://pbspro.atlassian.net/wiki/display/DG/PTL+Quick+Start+Guide)**).

### DIFF
--- a/src/resmom/requests.c
+++ b/src/resmom/requests.c
@@ -3423,7 +3423,7 @@ req_cpyfile(struct batch_request *preq)
 		}
 	}
 
-	if (dir == STAGE_DIR_OUT && direct_write_requested(pjob))
+	if ((pjob != NULL) && (dir == STAGE_DIR_OUT) && direct_write_requested(pjob))
 		stage_inout.direct_write = 1;
 	else
 		stage_inout.direct_write = 0;


src/resmom/requests.c
```diff
@@ -3423,7 +3423,7 @@ req_cpyfile(struct batch_request *preq)
 		}
 	}
 
-	if (dir == STAGE_DIR_OUT && direct_write_requested(pjob))
+	if ((pjob != NULL) && (dir == STAGE_DIR_OUT) && direct_write_requested(pjob))
```
minghui-liu: The inner parentheses aren't really necessary.

test/tests/functional/pbs_checkpoint.py
```diff
@@ -0,0 +1,88 @@
...
+        attrs = {'Resource_List.select': '1:ncpus=' + ncpus}
+        j1 = Job(TEST_USER, attrs)
+        j1.set_sleep_time(5)
+        jid1 = self.server.submit(j1)
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid1)
+        attrs['queue'] = 'expressq'
+        j2 = Job(TEST_USER, attrs)
+        j2.set_sleep_time(1)
+        self.server.submit(j2)
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid1)
```
minghui-liu: Do you mean `jid2` here?

test/tests/functional/pbs_checkpoint.py
```diff
@@ -0,0 +1,88 @@
...
+        self.mom.add_config(c)
+        self.scheduler.set_sched_config({'preempt_order': 'C'})
+        a = {'queue_type': 'e',
+             'started': 'True',
+             'enabled': 'True',
+             'Priority': 150}
+        self.server.manager(MGR_CMD_CREATE, QUEUE, a, "expressq")
+        node_attrs = self.server.status(NODE, id=self.server.shortname)[0]
+        ncpus = node_attrs['resources_available.ncpus']
+        attrs = {'Resource_List.select': '1:ncpus=' + ncpus}
```
bhroam: It's probably easier to just set the number of ncpus at the start of the test than figure out how many ncpus the machine has.

test/tests/functional/pbs_checkpoint.py
```diff
@@ -0,0 +1,88 @@
...
+        j1 = Job(TEST_USER, attrs)
+        j1.set_sleep_time(5)
```
bhroam: Why are you setting the sleep time?  Such a small sleep time is likely to cause race conditions in the test.

test/tests/functional/pbs_checkpoint.py
bhroam: Check for jid1 to be in state S.

test/tests/functional/pbs_checkpoint.py
```diff
@@ -0,0 +1,88 @@
...
+        jid1 = self.server.submit(j1)
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid1)
+        attrs['queue'] = 'expressq'
+        j2 = Job(TEST_USER, attrs)
+        j2.set_sleep_time(1)
+        self.server.submit(j2)
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid1)
+        time.sleep(30)
+        self.server.expect(JOB, {'job_state': 'H'}, id=jid1)
```
bhroam: Turn this sleep into an offset in the expect().  You might also add a log message saying why you are sleeping.  It'll help the test runner to know why they are waiting.

test/tests/functional/pbs_checkpoint.py
```diff
@@ -0,0 +1,88 @@
...
+        self.server.delete(id=jid1)
+        time.sleep(1)
```
bhroam: Why are you sleeping?

src/resmom/requests.c
bayucan: In this case, I still prefer them to be included for  better readability.

test/tests/functional/pbs_checkpoint.py
arungrover: @bayucan I think the job should ideally not go into the held state because it should just run as soon as the high priority job ends. The reason it is happening that way is because of another bug in PBS. Once that bug is resolved, this test will start failing.
I had a chat about this test with Bhroam and we think we can make this a generic checkpointing test that checks the first job ran and finished after the high priority job ends and also check for the node to be in free state (which you are already doing).
This will be a generic test and can also go into our SmokeTest test suite.

test/tests/functional/pbs_checkpoint.py
bayucan: That setting of sleep time is unnecessary. Removed.

test/tests/functional/pbs_checkpoint.py
bayucan: It's actually jid1 that we need to check here.

bayucan: There's checkpoint abort bug right now that causes that job to go to 'E' state and then 'H' state. Once the job in 'H' state is qdel-ed, then we encounter a mom crash. The test is trying to catch that.
I'll just check for the job going to the eventual 'H' state, but I also added a comment to say that once the checkpoint abort bug gets fixed, then we'll need to check for the job going to 'S' state.

test/tests/functional/pbs_checkpoint.py
bayucan: I removed the time.sleep() and just added a max_attempts.

test/tests/functional/pbs_checkpoint.py
bayucan: Removed it.

test/tests/functional/pbs_checkpoint.py
bayucan: Yes, there's a checkpoint abort bug right now that is preventing the first job into going into 'S' state but instead, goes to 'E' state and eventually to 'H' state. Once the job is in 'H' state and I issue a qdel-ed, then I get a mom crash, which is the situation this test is trying to capture. For now, I'll add a comment to mention all of this, and that once the checkpoint abort bug is fixed, we can check for the job going into 'S' state. I've also added code to check that the high priority job ran and finished so it can also look like a generic checkpoint test.

bayucan:
After much discussion with the folks reviewing the code, the PTL test pbs_checkpoint.py to verify fix will NOT be checked-in. It is written in such a way that is dependent on an existing checkpoint abort buggy behavior in PBS. This checkpoint bug will eventually get fixed an nullify this PTL test, so the test should not be checked-in at all. Instead, I will just post the results under "manual test".

bayucan:
@PBSPro/pbspro-maintainers: I've rebased and squashed. All tests passed. Please move forward...

bhroam:
I merged in the PR.

PR #704, opened 2018-05-29
Added code in TestPBSSnapshot to verify that snapshot captures the non-default location schedulers

Bhagat-Rajput:
#### Bug/feature Description
* Added code to verfiy that snapshot also captures the multisched scheduler's home at non-default location inside PBS_HOME

#### Affected Platform(s)
* All 

#### Solution Description
* Updated the existing PTL because also need to check/verify the multisched scheduler's at non-default location inside PBS_HOME.

#### Testing logs/output
* Test logs : [ptl_pbs_snapshot.txt](https://github.com/PBSPro/pbspro/files/2061854/ptl_pbs_snapshot.txt)

#### Checklist:

### DIFF
--- a/test/tests/functional/pbs_snapshot_unittest.py
+++ b/test/tests/functional/pbs_snapshot_unittest.py
@@ -77,7 +77,8 @@ class TestPBSSnapshot(TestFunctional):
         if ret['rc'] != 0:
             self.skipTest("pbs_snapshot/PBSSnapUtils need root privileges")
 
-    def setup_sc(self, sched_id, partition, port):
+    def setup_sc(self, sched_id, partition, port,
+                 sched_priv=None, sched_log=None):
         """
         Setup a scheduler
 
@@ -97,9 +98,18 @@ class TestPBSSnapshot(TestFunctional):
         a = {'partition': partition,
              'sched_host': self.server.hostname,
              'sched_port': port}
+        if sched_priv is not None:
+            a['sched_priv'] = sched_priv
+        if sched_log is not None:
+            a['sched_log'] = sched_log
         self.server.manager(MGR_CMD_CREATE, SCHED, a, id=sched_id)
-        self.scheds[sched_id].create_scheduler()
-        self.scheds[sched_id].start()
+        if 'sched_priv' in a:
+            sched_dir = os.path.dirname(sched_priv)
+            self.scheds[sched_id].create_scheduler(sched_dir)
+            self.scheds[sched_id].start(sched_dir)
+        else:
+            self.scheds[sched_id].create_scheduler()
+            self.scheds[sched_id].start()
         self.server.manager(MGR_CMD_SET, SCHED,
                             {'scheduling': 'True'}, id=sched_id, expect=True)
 
@@ -447,13 +457,20 @@ class TestPBSSnapshot(TestFunctional):
         if self.pbs_snapshot_path is None:
             self.skip_test("pbs_snapshot not found")
 
-        # Setup 2 schedulers
-        sched_ids = ["sc1", "sc2", "default"]
+        # Setup 3 schedulers
+        sched_ids = ["sc1", "sc2", "sc3", "default"]
         self.setup_sc(sched_ids[0], "P1", "15050")
         self.setup_sc(sched_ids[1], "P2", "15051")
+        # Setup scheduler at non-default location
+        dir_path = os.path.join(os.sep, 'var', 'spool', 'pbs', 'sched_dir')
+        if not os.path.exists(dir_path):
+            self.du.mkdir(path=dir_path, sudo=True)
+        sched_priv = os.path.join(dir_path, 'sched_priv_sc3')
+        sched_log = os.path.join(dir_path, 'sched_logs_sc3')
+        self.setup_sc(sched_ids[2], "P3", "15052", sched_priv, sched_log)
 
-        # Add 2 partitions, each associated with a queue and a node
-        (q_ids, _) = self.setup_queues_nodes(2)
+        # Add 3 partitions, each associated with a queue and a node
+        (q_ids, _) = self.setup_queues_nodes(3)
 
         # Submit some jobs to fill the system up and get the multiple
         # schedulers busy
@@ -475,6 +492,7 @@ class TestPBSSnapshot(TestFunctional):
             else:
                 schedi_priv = os.path.join(snapdir, "sched_priv_" + sched_id)
                 schedi_logs = os.path.join(snapdir, "sched_logs_" + sched_id)
+
             self.assertTrue(os.path.isdir(schedi_priv))
             self.assertTrue(os.path.isdir(schedi_logs))
 
@@ -492,7 +510,7 @@ class TestPBSSnapshot(TestFunctional):
                     sched_id = sched_id.strip()
                     self.assertTrue(sched_id in sched_ids)
                     scheds_found += 1
-            self.assertEqual(scheds_found, 3)
+            self.assertEqual(scheds_found, 4)
 
     @classmethod
     def tearDownClass(self):


Bhagat-Rajput:
I have added the ptl output.
[ptl_snapshot.txt](https://github.com/PBSPro/pbspro/files/2048593/ptl_snapshot.txt)

test/tests/functional/pbs_snapshot_unittest.py
```diff
@@ -448,12 +459,17 @@ def test_multisched_support(self):
             self.skip_test("pbs_snapshot not found")
 
         # Setup 2 schedulers
-        sched_ids = ["sc1", "sc2", "default"]
+        sched_ids = ["sc1", "sc2", "sc3", "default"]
         self.setup_sc(sched_ids[0], "P1", "15050")
         self.setup_sc(sched_ids[1], "P2", "15051")
+        # Setup scheduler at non-default location
+        dir_path = os.path.join(os.sep, 'var', 'spool', 'pbs', 'sched_dir')
```
agrawalravi90: the default location is also PBS_HOME, so I think it'd be a better test if the non-default location was somewhere outside of PBS_HOME, I'd recommend putting it under /tmp and see if it pbs_snapshot can still capture it correctly.

agrawalravi90: Also, just set the complete path here (including 'sched_priv_sc3' and 'sched_logs_sc3') instead of hard-coding it inside setup_sc().

test/tests/functional/pbs_snapshot_unittest.py
```diff
@@ -94,12 +95,22 @@ def setup_sc(self, sched_id, partition, port):
...
         :type log_filter: int
         """
-        a = {'partition': partition,
-             'sched_host': self.server.hostname,
-             'sched_port': port}
+        if sched_priv is not None:
+            a = {'partition': partition,
+                 'sched_host': self.server.hostname,
+                 'sched_port': port,
+                 'sched_priv': os.path.join(sched_priv, 'sched_priv_sc3'),
```
agrawalravi90: Please don't hard-code the priv and log directory names inside the generic function setup_sc(), pass it the complete path from the caller itself.

test/tests/functional/pbs_snapshot_unittest.py
```diff
@@ -94,12 +95,22 @@ def setup_sc(self, sched_id, partition, port):
         :param log_filter: log filter value for the scheduler
         :type log_filter: int
         """
-        a = {'partition': partition,
-             'sched_host': self.server.hostname,
-             'sched_port': port}
+        if sched_priv is not None:
```
agrawalravi90: What if sched_log is None? you are not checking for that. Suggestion: create attribute list with just 'partition', 'sched_host' and 'sched_port'. Then, add 'sched_priv' or 'sched_log' if they are not None.

test/tests/functional/pbs_snapshot_unittest.py
```diff
@@ -94,12 +95,22 @@ def setup_sc(self, sched_id, partition, port):
...
+            a = {'partition': partition,
+                 'sched_host': self.server.hostname,
+                 'sched_port': port,
+                 'sched_priv': os.path.join(sched_priv, 'sched_priv_sc3'),
+                 'sched_log': os.path.join(sched_log, 'sched_logs_sc3')}
+        else:
```
agrawalravi90: See my comment above, I think it's cleaner if you build the attribute list with just these 3 attributes and then add 'sched_priv' and 'sched_log' conditionally.

test/tests/functional/pbs_snapshot_unittest.py
```diff
@@ -94,12 +95,22 @@ def setup_sc(self, sched_id, partition, port):
...
+            a = {'partition': partition,
+                 'sched_host': self.server.hostname,
+                 'sched_port': port}
         self.server.manager(MGR_CMD_CREATE, SCHED, a, id=sched_id)
         self.scheds[sched_id].create_scheduler()
-        self.scheds[sched_id].start()
+        if sched_priv is not None:
+            self.scheds[sched_id].start(sched_priv)
```
agrawalravi90: Doesn't start() take 'sched_home' as the first argument? why are you providing it sched_priv path?

test/tests/functional/pbs_snapshot_unittest.py
```diff
@@ -448,12 +459,17 @@ def test_multisched_support(self):
...
+        sched_ids = ["sc1", "sc2", "sc3", "default"]
         self.setup_sc(sched_ids[0], "P1", "15050")
         self.setup_sc(sched_ids[1], "P2", "15051")
+        # Setup scheduler at non-default location
+        dir_path = os.path.join(os.sep, 'var', 'spool', 'pbs', 'sched_dir')
+        if not os.path.exists(dir_path):
+            self.du.mkdir(path=dir_path, sudo=True)
+        self.setup_sc(sched_ids[2], "P3", "15052", dir_path, dir_path)
 
         # Add 2 partitions, each associated with a queue and a node
```
agrawalravi90: Modify this comment to say 3 partitions instead of 2

test/tests/functional/pbs_snapshot_unittest.py
Bhagat-Rajput: @agrawalravi90 Thanks for the review. Yes, it takes and i am also passing the same just forgot to change the name, will change the name from sched_priv to sched_dir

test/tests/functional/pbs_snapshot_unittest.py
agrawalravi90: Can you just create it inside the user's home directory (not PBS_HOME, just the user's home/?

Bhagat-Rajput: ok, i will try to create inside user's home and let you know.Thanks

test/tests/functional/pbs_snapshot_unittest.py
```diff
@@ -97,9 +98,17 @@ def setup_sc(self, sched_id, partition, port):
         a = {'partition': partition,
              'sched_host': self.server.hostname,
              'sched_port': port}
+        if sched_priv is not None:
+            a.update({'sched_priv': sched_priv})
```
agrawalravi90: this is inefficient, please replace this with simply a['sched_priv'] = sched_priv. Same for sched_log.

test/tests/functional/pbs_snapshot_unittest.py
```diff
@@ -97,9 +98,17 @@ def setup_sc(self, sched_id, partition, port):
...
+        if sched_log is not None:
+            a.update({'sched_log': sched_log})
         self.server.manager(MGR_CMD_CREATE, SCHED, a, id=sched_id)
         self.scheds[sched_id].create_scheduler()
-        self.scheds[sched_id].start()
+        if(len(a) > 3):
+            sched_dir = os.path.dirname(sched_priv)
+            self.scheds[sched_id].start(sched_dir)
```
agrawalravi90: I took a closer look at Scheduler.start(sched_home), it seems like 'sched_home' is passed as PBS_HOME argument to pbs_sched, which seems wrong to me. @suresh-thelkar, @arungrover and @bhroam , can you guys please help me understand if this is a bug? if 'sched_home', which is parent directory of a sched's sched_priv/log, is set as PBS_HOME for that sched, then it'd imply that the sched will search for this parent directory for the existence of the job_sort_formula file (instead of PBS_HOME) right? Won't this be an issue? job_sort_formula should only exist inside a file in PBS_HOME right?

suresh-thelkar: @agrawalravi90,

Following are my thoughts on this.

We need to pass sched_home directory to  Scheduler.start() only when we create scheduler's priv and log directories outside the regular PBS_HOME.  This parameter serves as value for -d argument of pbs_sched.  If scheduler's priv and log directories are inside PBS_HOME then we no need to pass value for sched_home.

Each multisched having its own job_sort_formula is not supported at the moment. We are ready with our design/analysis to for this enhancement which is not getting prioritized.

agrawalravi90: Thanks for your reply Suresh.

I was not thinking about each multisched having its own job_sort_formula, I was concerned about what would happen if we set job_sort_formula on the server with a sched that was run with -d /tmp (for example) ? Wouldn't that sched try to access the sched_formula file inside /tmp/sched_priv? Will it find a copy of the job_sort_formula's file (sched_formula?) there?

test/tests/functional/pbs_snapshot_unittest.py
Bhagat-Rajput: I had tried to create a directory inside user's home but couldn't start the scheduler. I had discussed with @suresh-thelkar so as per the current design, we can create it inside only such user's home whose have "uid" < "10" . @suresh-thelkar Please correct if i got it wrong.

test/tests/functional/pbs_snapshot_unittest.py
bhroam: @suresh-thelkar Was that fact that job_sort_formula didn't work with multi-sched clear in your EDD?  This needs to be communicated in the docs or release notes properly.

suresh-thelkar: @bhroam,
As we have not yet made job_sort_formula a scheduler attribute we initially thought of not adding it to the EDD. An enhancement to make job_sort_formula to work with Multisched has been logged already. We would like to add a new interface to Multisched EDD once we start working on this enhancement. 
Since scheduler's priv and log directories can be created in any path, planning to add the following caveat either to the documentation/release notes.  Thanks for reminding us. Hope the following information is enough. Please let me know your thoughts in case if we need to add more details.
**Caveat**: Each multisched uses/shares the same job_sort_formula set at server for the time being. Scheduler's sched_priv and sched_log directories should be at the same level as of default scheduler's priv and log directories as of now.

Hope @agrawalravi90 the above explanation answers your question as well. Please let us know in case if anything more needs to be added.

test/tests/functional/pbs_snapshot_unittest.py
agrawalravi90: ok, I'm fine with keeping the code as it is (inside PBS_HOME/<dir>)

test/tests/functional/pbs_snapshot_unittest.py
bhroam: @suresh-thelkar The scheduler doesn't read the job_sort_formula attribute from the server.  It is written into a file in sched_priv.  The scheduler reads it from the file.  This is for security reasons.  We don't want a non-root manager to set this attribute and have arbitrary python code run as root.  From what @agrawalravi90 just said, the file isn't being copied into the non-default scheduler's sched_priv directories.  This means the job_sort_formula isn't shared by all schedulers.  It will only work for the default scheduler.

suresh-thelkar: Thanks for letting us know the intent behind copying the sched_formula to sched_priv. Thereafter this file is used by scheduler to read the formula and proceed accordingly.  I am still thinking that job_sort_formula works for Multisched as well( with the same formula as that of default scheduler and with the above caveat I mentioned in the above comments).  In order to find more details I have executed the scenario attached to confirm whether the same formula is used for both default scheduler as well as Multisched. It looks to me that both are using the formula and jobs are sorted accordingly.  Please look into the attachment and correct me If I am missing anything.
[Existing_Job_sort_formula_Analysis.txt](https://github.com/PBSPro/pbspro/files/2066838/Existing_Job_sort_formula_Analysis.txt)

Also looked at the code to see how scheduler is reading the job_sort_formula. It calls the function read_formula() where the scheduler file is read from the macro FORMULA_ATTR_PATH_SCHED which is nothing but "sched_priv/sched_formula". So even in case of Multisched also since this code has not been changed as of now it still reads the formula from the same macro(only thing is that corresponding Multisched should have access to default scheduler's sched_priv directory). So with respect to this I am feeling that job_sort_formula works for Multisched as well except with the following caveat till the time we complete the job_sort_formula enhancement for Multisched.

**Caveat**: Each multisched tries to use the same job_sort_formula of default scheduler for the time being. As long as Multisched's sched_priv_<sc_name> directory is created at the same level as that of default scheduler's sched_priv directory.
Example: sched_priv_<sc_name> is strictly created in PBS_HOME.

bhroam: Please have this caveat added to the release notes for multi-sched.

test/tests/functional/pbs_snapshot_unittest.py
```diff
@@ -97,9 +98,18 @@ def setup_sc(self, sched_id, partition, port):
...
              'sched_host': self.server.hostname,
              'sched_port': port}
+        if sched_priv is not None:
+            a['sched_priv'] = sched_priv
+        if sched_log is not None:
+            a['sched_log'] = sched_log
         self.server.manager(MGR_CMD_CREATE, SCHED, a, id=sched_id)
-        self.scheds[sched_id].create_scheduler()
-        self.scheds[sched_id].start()
+        if(len(a) > 3):
```
agrawalravi90: If 'sched_priv' is None, but 'sched_log' is not None, len(a) > 3 will be satisfied and it will crash the test when os.path.dirname(sched_priv) gets executed. Maybe change this condition to "if 'sched_priv' in a" instead ?

test/tests/functional/pbs_snapshot_unittest.py
agrawalravi90: Thanks for clarifying Suresh. I'm sorry but I have one more question: Right now, pbs_sched -d has been overloaded to accept the parent directory of sched_priv and sched_log. So, what will happen if they are put in different locations? Or is that not supported? Or is pbs_sched -d not supported? I couldn't find any references to pbs_sched -d in the EDD.

agrawalravi90:
Thanks for clarifying @suresh-thelkar. I'm sorry but I have one more question: Right now, pbs_sched -d has been overloaded to accept the parent directory of sched_priv and sched_log. So, what will happen if they are put in different locations? Or is that not supported? Or is pbs_sched -d not supported to be the parent directory for sched_priv/sched_log? I couldn't find any references to pbs_sched -d in the EDD.

test/tests/functional/pbs_snapshot_unittest.py
Bhagat-Rajput: Thanks @agrawalravi90 for the valuable suggestion and have implemented the same.

suresh-thelkar:
@agrawalravi90,
As I was occupied with critical priority sorry I could not reply it yesterday. Just wanted to answer your query with an example. Lets take the following Multisched configuration as an example.
[root@stblr3 pbspro-1]# qmgr -c "l sc sc2"
Sched sc2
    sched_host = stblr3
    pbs_version = 18.1.0
    sched_cycle_length = 00:20:00
    sched_port = 15051
    partition = P2
    sched_priv = /var/spool/pbs/sched_dir/sched_priv_sc2
    sched_log = /tmp
    scheduling = True
    scheduler_iteration = 600
    state = idle

1. Here sched_priv and sched_log are totally in different locations and if sched_logs_sc2 is not present then start scheduler as follows. Now the scheduler won't be started.
pbs_sched -I sc2 -S 15051 -d /var/spool/pbs/sched_dir
[root@stblr3 pbspro-1]# pbs_sched -I sc2 -S 15051 -d /var/spool/pbs/sched_dir
pbs_sched: logfile could not be opened

2. If sched_logs_sc2 is present then scheduler is started. Here the initial set of logs go to sched_logs_sc2 directory and as soon as scheduler establishes contact with server then the logs from that time on words goes to /tmp directory.
[root@stblr3 sched_dir]# pbs_sched -I sc2 -S 15051 -d /var/spool/pbs/sched_dir
Creating usage database for fairshare.
**_Initial set of logs:_**
[root@stblr3 sched_logs_sc1]# tail -f 20180607 
06/07/2018 18:24:44;0c06;pbs_sched;TPP;pbs_sched(Main Thread);Max files too low - you may want to increase it.
06/07/2018 18:24:44;0d80;pbs_sched;TPP;pbs_sched(Main Thread);TPP initialization done
06/07/2018 18:24:44;0c06;pbs_sched;TPP;pbs_sched(Main Thread);Single pbs_comm configured, TPP Fault tolerant mode disabled
06/07/2018 18:24:44;0d80;pbs_sched;TPP;pbs_sched(Main Thread);Connecting to pbs_comm stblr3:17001
06/07/2018 18:24:44;0c06;pbs_sched;TPP;pbs_sched(Thread 0);Thread ready
06/07/2018 18:24:44;0c06;pbs_sched;TPP;pbs_sched(Thread 0);Registering address 192.168.75.129:15051 to pbs_comm
06/07/2018 18:24:44;0c06;pbs_sched;TPP;pbs_sched(Thread 0);Registering address 192.168.122.1:15051 to pbs_comm
06/07/2018 18:24:44;0c06;pbs_sched;TPP;pbs_sched(Thread 0);Registering address 172.17.0.1:15051 to pbs_comm
06/07/2018 18:24:44;0c06;pbs_sched;TPP;pbs_sched(Thread 0);Connected to pbs_comm stblr3:17001
06/07/2018 18:24:45;0002;pbs_sched;Svr;Log;Log closed

**_After scheduler gets an update from server: (Lets say a job is submitted)_**
tail -f 20180607
06/07/2018 18:25:05;0080;pbs_sched;Req;;Starting Scheduling Cycle
06/07/2018 18:25:05;0004;pbs_sched;Fil;holidays;The holiday file is out of date; please update it.
06/07/2018 18:25:05;0080;pbs_sched;Job;7.stblr3;Considering job to run
06/07/2018 18:25:05;0040;pbs_sched;Job;7.stblr3;Job run
06/07/2018 18:25:05;0080;pbs_sched;Req;;Leaving Scheduling Cycle

Looking at this I feel that ‘-d’ option to pbs_sched is mainly used to detect its sched_priv<sc_name> directory and it is used only if the directory is in non default path. If logs directory is present at the same location then scheduler will be started and later on or as soon as it gets an update from server then the logs from that time on words goes to the location of sched_logs that is configured at the qmgr for this Scheduler. In other words I can say ‘-d’ option is working fine but with a little bit of change in behavior as explained above. If you do feel that we need to document this either in EDD or our release notes please do let us know your opinion so that I will update them accordingly. In case if you would like us to modify the behavior of ‘-d’ option please file a ticket with the required description and we will work on it accordingly.

agrawalravi90:
Thanks for making the changes Bhagat, they look good to me. I'll take the multi-sched discussion offline as it's clearly out of scope for this PR. You need one more reviewer.

suresh-thelkar:
Code changes look good to me as well. I sign off.

Bhagat-Rajput:
Thanks @agrawalravi90 and @suresh-thelkar for sign off. I have rebased and squashed commits, ready to merge.

PR #705, opened 2018-05-30
prevent using GPUs without user awareness

vchlum:
#### Bug/feature Description
* Some programs (e.g. TensorFlow) autodetect GPU devices and use the devices without user awareness. This can lead to unwanted using of GPUs without its allocation in PBS Pro.

#### Affected Platform(s)
* CUDA

* This issue is related to the cgroup hook. If the CUDA_VISIBLE_DEVICES env is not set then some programs autodetect GPUs. The programs usually respect CUDA_VISIBLE_DEVICES and we can set the value of the CUDA_VISIBLE_DEVICES to the empty string in order to prevent using the GPUs.

#### Solution Description
* If the 'devices' subsystem is enabled we can simply set the CUDA_VISIBLE_DEVICES to the empty string. If the GPUs are actually requested the correct value will override the empty value.

#### Manual test
* The job requests the GPUs:
```
(JESSIE)vchlum@torque1:~$ qsub -I -l select=vnode=konos5:ngpus=4
qsub: waiting for job 32497.torque1.ics.muni.cz to start
qsub: job 32497.torque1.ics.muni.cz ready
konos5.fav.zcu.cz$ env | grep CUDA
CUDA_VISIBLE_DEVICES=0,1,2,3
konos5.fav.zcu.cz$ 
```

* The job does not request any GPU:
```
(JESSIE)vchlum@torque1:~$ qsub -I -l select=vnode=konos5
qsub: waiting for job 32496.torque1.ics.muni.cz to start
qsub: job 32496.torque1.ics.muni.cz ready
konos5.fav.zcu.cz$ env | grep CUDA
CUDA_VISIBLE_DEVICES=
konos5.fav.zcu.cz$ 
```

#### Testing logs/output
* [ptl_output_pbs_cgroups_hook.txt](https://github.com/PBSPro/pbspro/files/2053323/ptl_output_pbs_cgroups_hook.txt)

#### Checklist:

### DIFF
--- a/src/hooks/cgroups/pbs_cgroups.PY
+++ b/src/hooks/cgroups/pbs_cgroups.PY
@@ -2818,6 +2818,9 @@ class CgroupUtils(object):
         execjob_launch hook
         """
         pbs.logmsg(pbs.EVENT_DEBUG4, '%s: Method called' % caller_name())
+        if 'devices' in self.subsystems:
+            # prevent using GPUs without user awareness
+            pbs.event().env['CUDA_VISIBLE_DEVICES'] = ''
         if 'device_names' in self.assigned_resources:
             names = self.assigned_resources['device_names']
             pbs.logmsg(pbs.EVENT_DEBUG4,


vchlum:
Thank you, It's a good point.

I think we should not bother the user who does not request GPU with the initialization of CUDA_VISIBLE_DEVICES.

4) Another solution would be to add another hook and this hook will set the env for the job.

From my point of view, I'm fine with 1 or 4, but I think others could face this issue too, and the cgroups hook is the only place where PBS sets CUDA_VISIBLE_DEVICES out of the box. So, I think this is a good place where to prevent this issue. I do not see a better way how to provide the instant solution for others. An accidental use of an unallocated GPU is a serious issue.

Another solution would be to add a short warning to the cgroups documentation... Maybe it is enough:)

mike0042:
If the hook is the responsible for setting the environment variable, then it makes sense that it should initialize it regardless of whether the user requested GPUs. I was curious if you (or anyone reading this) could imagine a case where we would NOT want to initialize it. If not, then I'm fine with the change.

Do you think we should do the same for OFFLOAD_DEVICES?

vchlum:
Unfortunately, I do not know whether to do the same with OFFLOAD_DEVICES or not.  We do not have nodes with MIC cards, and I can't test it.

mike0042:
If OFFLOAD_DEVICES causes the same problem in the future, someone will have to file a ticket.

Please sign your commit.

vchlum:
Thank you @mike0042 , commit has been signed.

PR #709, opened 2018-06-01
Adding generic library function - break_delimited_str

nithinj:
#### Bug/feature Description
* Create a more generic version of the function break_comma_list()

* break_comma_list() can break a string based on the delimiter comma. The new function should break a list based on the delimiter passed to it.

#### Solution Description
* Created a new function break_delimited_str() which accepts delimiter as the second parameter.
* break_comma_list() is not removed as this is used in several places.

#### Checklist:

CC: @arungrover , @bhroam , @subhasisb

### DIFF
--- a/src/include/libutil.h
+++ b/src/include/libutil.h
@@ -241,10 +241,16 @@ lock_file(FILE *fp, int op, char *filename, int lock_retry,
 
 
 /*
- *      break_comma_list - break apart a comma delemetd string into an arary
+ *      break_comma_list - break apart a comma delimited string into an array
  *                         of strings
  */
 char **break_comma_list(char *list);
+
+/*
+ *      break_delimited_str - break apart a delimited string into an array
+ *                         of strings
+ */
+char **break_delimited_str(char *list, char delim);
 
 /*
  *      free_string_array - free an array of strings with NULL as sentinel
--- a/src/lib/Libutil/misc_utils.c
+++ b/src/lib/Libutil/misc_utils.c
@@ -1035,22 +1035,28 @@ validate_ext_auth_data(int auth_type, void *data, int data_len, char *ebuf, int 
 
 /**
  *
- *	@brief break apart a comma delimited string into an array of strings
+ *	@brief break apart a delimited string into an array of strings
  *
- *	  @param[in] strlist - the comma delimited string
+ *	@param[in] strlist - the delimited string
+ *	@param[in] sep - the separator character
  *
  *	@return char **
  *
+ *	@note
+ *		The returned array of strings has to be freed by the caller.
  */
 char **
-break_comma_list(char *strlist)
+break_delimited_str(char *strlist, char delim)
 {
+	char sep[2] = "\0";
 	int num_words = 1; /* number of words delimited by commas*/
 	char **arr = NULL; /* the array of words */
 	char *list;
 	char *tok; /* used with strtok() */
 	char *end;
 	int i;
+
+	sep[0] = delim;
 
 	if (strlist == NULL) {
 		pbs_errno = PBSE_BADATVAL;
@@ -1063,7 +1069,7 @@ break_comma_list(char *strlist)
 		char* saveptr = NULL;
 
 		for (i = 0; list[i] != '\0'; i++)
-			if (list[i] == ',')
+			if (list[i] == delim)
 				num_words++;
 
 		if ((arr = (char **) malloc(sizeof(char*) * (num_words + 1))) == NULL) {
@@ -1072,7 +1078,7 @@ break_comma_list(char *strlist)
 			return NULL;
 		}
 
-		tok = strtok_r(list, ",", &saveptr);
+		tok = strtok_r(list, sep, &saveptr);
 
 		for (i = 0; tok != NULL; i++) {
 			while (isspace((int) *tok))
@@ -1086,7 +1092,7 @@ break_comma_list(char *strlist)
 			}
 
 			arr[i] = strdup(tok);
-			tok = strtok_r(NULL, ",", &saveptr);
+			tok = strtok_r(NULL, sep, &saveptr);
 		}
 		arr[i] = NULL;
 	}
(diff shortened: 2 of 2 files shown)


src/lib/Libutil/misc_utils.c
```diff
@@ -1035,23 +1035,29 @@ validate_ext_auth_data(int auth_type, void *data, int data_len, char *ebuf, int
...
+ *	@param[in] sep - the separator character
  *
  *	@return char **
  *
+ *	@note
+ *		The returned array of strings has to be freed by the caller.
  */
 char **
-break_comma_list(char *strlist)
+break_delimited_str(char *strlist, char delim)
```
suresh-thelkar: Since we are anyhow making this function more generic why can't  we allow multiple delimiters instead of  only one. If we do this this is inline with strtok_r/strtok which supports multiple delimiters.

arungrover: I am not aware of any use case where we would need to split the string on the basis of multiple tokens. If there is no use case today, I'd suggest that we don't do it. It will complicate things when we don't need it.

suresh-thelkar: I am actually thinking from the semantics perspective of the generic functions break_delimited_str() vs strtok()/strtok_r(). Because we anyhow internally making use of these functions which actually allow multiple delimiters and planning to make our new function super generic  I felt it would be better to add this. But as Arun said if we do not have any use case in PBS as of now where there is a need of breaking the string by one or more delimiters I am fine leaving the code as it is.

bhroam:
Maybe this is a future feature, but consider adding the ability to escape the delimiter char.  Right now if you do a\,b,c, you'll get 'a\' 'b' and 'c'.  The other nice feature you might consider is strings.  Just ignore delimiters within a quoted string.  Right now break_comma_list() doesn't do this, so I see no reason break_delimited_list() has to do this now.

nithinj:
@bhroam ,
I'm not sure whether you are asking for this. `escape_delimiter()` will escape the delimiters in a string. available in `misc_utils.c`
```
escape_delimiter(char *str, char *delim, char esc)
Escape every occurrence of 'delim' in 'str' with 'esc'
```

nithinj:
Hi @bhroam / @arungrover ,
This branch is up to date. Please take this further.

src/include/libutil.h
```diff
@@ -241,11 +241,17 @@ lock_file(FILE *fp, int op, char *filename, int lock_retry,
...
 
 /*
- *      break_comma_list - break apart a comma delemetd string into an arary
+ *      break_comma_list - break apart a comma delemetd string into an array
  *                         of strings
  */
 char **break_comma_list(char *list);
 
+/*
+ *      break_delimited_str - break apart a delemeted string into an array
```
bremanandjk: Spelling error:  "delemetd" 
If you wish to change, It would be better.

nithinj:
@bremanandjk ,
I've made the corrections and included in the same commit. Hope that is fine.

arungrover:
@nithinj still looks good to me. Once you have another signoff from Brem please rebase the branch

nithinj:
@arungrover ,
I've re-based my branch. Please take it further.

src/lib/Libutil/misc_utils.c
```diff
@@ -1035,23 +1035,29 @@ validate_ext_auth_data(int auth_type, void *data, int data_len, char *ebuf, int
...
 {
+	char sep[2] = "\0";
```
mike0042: If the intent is to create a NULL terminated string with delim being the only character, you want to initialize this as...
char sep[2] = { '\0', '\0'};

Otherwise, you could sprintf(sep, "%c", delim) below.

What you have now is NOT null terminated.

vstumpf: It is null terminated, because the compiler will add the terminating null at the end, since it's a string of length 1.

It does look a little weird though.

mike0042: It's more than weird, it's wrong.

While the compiler may terminate the string literal for us, the string itself doesn’t contain what it should. If you pass comma as the delimiter then sep has the value “,0” which gets passed to strtok_r().

Also, we’re not flagging an error when strdup() fails. We end up returning NULL and pbs_errno is unchanged. Should we not be initializing pbs_errno to PBSE_NONE if the caller will be checking it?

IMHO, this code change is incorrect. Further, there are no test logs attached so it should not have been merged. I’d like to request this change be reverted.

arungrover: Mike I still don't see a problem. If a comma is passed to this function then wouldn't a null terminated comma be passed to strtok_r? With byte order being ,\0
Also while strdup failure is something we could check that is not something Nithin added. 
The code has been existing for as long as I remember and I didn't ask for an additional test because it gets exercised in all the cases where scheduler parses resources string in every sched startup

arungrover: Mike I still don't see a problem. If a comma is passed to this function then wouldn't a null terminated comma be passed to strtok_r? With byte order being ,\0
Also while strdup failure is something we could check that is not something Nithin added. 
The code has been existing for as long as I remember and I didn't ask for an additional test because it gets exercised in all the cases where scheduler parses resources string in every sched startup

mike0042: So I may have misunderstood one aspect of what you are saying. I was thinking that "\0" would result in sep[0] being '\' and sep[1] being '0'. That is not the case. Since the string literal implies a null terminated string, are we certain that the '\0' gets assigned to sep[1]?

I would feel much better if the initialization was done explicitly, without assuming the null terminator gets copied.

There is still the issue of initializing pbs_errno and checking the return code strdup(). Whether Nithin added it or not, it should be checked. Technically, this is a new function even though most of it was copied from break_comma_list().

vstumpf: `"\0"` is a character array of length 2; the first character is `'\0'`, and the second character is the null terminating byte, which is also `'\0'`.

What we end up with is:
```C
sep[0] = '\0'
sep[1] = '\0'
```

PR #712, opened 2018-06-06
qsub: parameter -v with commas in value is broken

vchlum:
#### Bug/feature Description
* man qsub says (-v parameter): 
If a variable=value pair contains any commas, the value must be enclosed in single  or double quotes, and the variable=value pair must be enclosed in the kind of quotes not used to enclose the value.  For example:
                      qsub -v "var1='A,B,C,D'" job.sh
* This feature is broken:
```
(JESSIE)vchlum@torque2:~$ qsub -v "var1='A,B,C,D'" job.sh
qsub: cannot send environment with the job
(JESSIE)vchlum@torque2:~$ 
```

#### Affected Platform(s)
* Linux

* copy_env_value() is called twice on the value of the var1. In the first call, the quotes are removed and the commas are escaped. In the second call, the first comma is considered to be the end of the value because the quotes are gone and the value ends with first ESC_CHAR.

#### Solution Description
* In the copy_env_value(), we should ensure that if the comma comes after ESC_CHAR, it is not the end of the value but the value continues. Moreover, we should not escape already escaped comma.

#### Testing logs/output
* Smoketest: [ptl_output_smoketest.txt](https://github.com/PBSPro/pbspro/files/2075019/ptl_output_smoketest.txt)
* New test: [ptl_passing_environment_variable_output.txt](https://github.com/PBSPro/pbspro/files/2075021/ptl_passing_environment_variable_output.txt)

#### Checklist:

### DIFF
--- a/src/cmds/qsub.c
+++ b/src/cmds/qsub.c
@@ -4055,6 +4055,7 @@ copy_env_value(char *dest, /* destination  */
 	int   go = 1;
 	int   q_ch = 0;
 	int   is_func = 0;
+	char  *dest_full = dest;
 
 	while (*dest)
 		++dest;
@@ -4087,12 +4088,15 @@ copy_env_value(char *dest, /* destination  */
 
 			case ESC_CHAR:			/* backslash in value, escape it */
 				*dest++ = *pv;
-				*dest++ = *pv;
+				if (*(pv+1) != ',') /* do not escape if ESC_CHAR already escapes */
+					*dest++ = *pv;
 				break;
 
 			case ',':
 				if (q_ch || quote_flg) {
 					*dest++ = ESC_CHAR;
+					*dest++ = *pv;
+				} else if (dest_full != dest && *(dest-1) == ESC_CHAR) { /* the comma is escaped, not finished yet */
 					*dest++ = *pv;
 				} else {
 					go = 0;		/* end of value string */
--- a/test/tests/functional/pbs_passing_environment_variable.py
+++ b/test/tests/functional/pbs_passing_environment_variable.py
@@ -1,0 +1,68 @@
+# coding: utf-8
+
+# Copyright (C) 1994-2018 Altair Engineering, Inc.
+# For more information, contact Altair at www.altair.com.
+#
+# This file is part of the PBS Professional ("PBS Pro") software.
+#
+# Open Source License Information:
+#
+# PBS Pro is free software. You can redistribute it and/or modify it under the
+# terms of the GNU Affero General Public License as published by the Free
+# Software Foundation, either version 3 of the License, or (at your option) any
+# later version.
+#
+# PBS Pro is distributed in the hope that it will be useful, but WITHOUT ANY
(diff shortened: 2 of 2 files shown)


src/cmds/qsub.c
```diff
@@ -4087,13 +4087,16 @@ copy_env_value(char *dest, /* destination  */
...
+				if (*(pv+1) != ',') /* do not escape if ESC_CHAR already escapes */
+					*dest++ = *pv;
 				break;
 
 			case ',':
 				if (q_ch || quote_flg) {
 					*dest++ = ESC_CHAR;
 					*dest++ = *pv;
+				} else if (*(dest-1) == ESC_CHAR) { /* the comma is escaped, not finished yet */
+					*dest++ = *pv;
```
mike0042: What if the string starts with a comma? We would not want to check the value of *(dest-1) in that case.

vchlum: This should not happen. ",var1='A,B,C,D'" is rejected before it enters copy_env_value(), but you are right, it is better to handle it. Please @mike0042, have a look.

src/cmds/qsub.c
```diff
@@ -4087,13 +4088,16 @@ copy_env_value(char *dest, /* destination  */
...
+				if (*(pv+1) != ',') /* do not escape if ESC_CHAR already escapes */
+					*dest++ = *pv;
 				break;
 
 			case ',':
 				if (q_ch || quote_flg) {
 					*dest++ = ESC_CHAR;
 					*dest++ = *pv;
+				} else if (dest_full != dest && *(dest-1) == ESC_CHAR) { /* the comma is escaped, not finished yet */
+					*dest++ = *pv;
```
mike0042: I guess there's really nothing we can do here if a comma is the first character, so we fall through to the else. Works for me.

bayucan:
@vchlum: the changes looked fine, although, can you make sure the following testsuite passes:  test/tests/functional/pbs_hook_set_jobenv.py. This contains many tests that exercise setting job environment with special characters. We want to make sure there's no regression with this fix.

vchlum:
@bayucan There is a problem with the test/tests/functional/pbs_hook_set_jobenv.py. I can not pass the test on the master branch nor this one. I think it is connected to the [PP-1008](https://pbspro.atlassian.net/browse/PP-1008). The TEST_RETURN is written into the log and log_match() does not read it correctly. As mentioned in [PP-1008](https://pbspro.atlassian.net/browse/PP-1008), everything after '\n' is lost.

I commented all the TEST_RETURN related lines in the pbs_hook_set_jobenv.py and it passed. Output of modified pbs_hook_set_jobenv.py [here](https://github.com/PBSPro/pbspro/files/2093711/ptl_output_hook_set_jobenv.txt).

bayucan:
@vchlum: Yes, I'm getting the same error running pbs_hook_set_jobenv.py. Ok, thanks for running it. Your testing logs are fine as is.

vchlum:
Thank you for the sign-offs. I have squashed and rebased.

PR #713, opened 2018-06-06
Include pbs_sleep for linux distribution

saimanikant:
#### Bug/feature Description
* *Include pbs_sleep for linux distribution*

#### Affected Platform(s)
* *Linux platforms*

* *Including pbs_sleep in linux distribution for supporting container images that might not have a sleep binary.*
* *A single common sleep command for PTL tests on all platforms (including windows) for test jobs*

#### Solution Description
* *make changes to existing pbs_sleep_win and make it generic.*

#### Testing logs/output

#### Checklist:

### DIFF
--- a/src/tools/Makefile.am
+++ b/src/tools/Makefile.am
@@ -43,7 +43,8 @@ bin_PROGRAMS = \
 	pbs_wish \
 	printjob.bin \
 	printjob_svr.bin \
-	tracejob
+	tracejob \
+	pbs_sleep
 
 sbin_PROGRAMS = \
 	pbs_ds_monitor \
@@ -59,6 +60,9 @@ common_libs = \
 	$(top_builddir)/src/lib/Libpbs/.libs/libpbs.a \
 	-lpthread \
 	@socket_lib@
+
+pbs_sleep_LDFLAGS = -all-static
+pbs_sleep_SOURCES = pbs_sleep.c
 
 chk_tree_CPPFLAGS = -I$(top_srcdir)/src/include
 chk_tree_LDADD = ${common_libs}
--- a/src/tools/pbs_sleep.c
+++ b/src/tools/pbs_sleep.c
@@ -37,33 +37,55 @@
  */
 /**
  * @file
- *		pbs_sleep_win.c
+ *		pbs_sleep.c
  *
  * @brief
- *		This file contains functions related to sleep of PBS on Windows.
+ *		This file contains functions related to sleep of PBS.
  *
  * Functions included are:
  * 	main()
  *
  */
 #include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <string.h>
+#include <signal.h>
+#ifdef WIN32
 #include <windows.h>
 #include "win.h"
+#endif
+
 /**
  * @Brief
- *      This is main function of pbs_sleep_win process.
- *      It calls sleep internally for the number of seconds passed to it.
+ *      This is main function of pbs_sleep process.
+ *      It calls sleep internally for the number of seconds passed to it, -1 for sleep indefinitely.
  *
  */
+
+
+int
 main(int argc, char *argv[])
 {
+	int i;
+	int forever = 0;
+	int secs = 0;
+
 	if (argc != 2) {
 		fprintf(stderr, "%s secs\n", argv[0]);
 		exit(1);
 	}
 
-	sleep(atoi(argv[1]));
 
-	exit(0);
+	/* if argv[1] is -1, loop with sleep 1 indefinitely */
+	if (strcmp(argv[1], "-1") == 0)
+		forever = 1;
+	else
+		secs = atoi(argv[1]);
+
+	for (i = 0; i < secs || forever; i++)
+		sleep(1);
+
+	return 0;
 
 }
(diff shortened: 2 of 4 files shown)


src/tools/Makefile.am
```diff
@@ -43,7 +43,8 @@ bin_PROGRAMS = \
 	pbs_wish \
 	printjob.bin \
 	printjob_svr.bin \
-	tracejob
+	tracejob \
+	pbs_sleep
```
hirenvadalia: In windows this binary is named as `pbs-sleep` instead `pbs_sleep` so if you really want to create one sleep command for all platforms then I suggest to change this to `pbs-sleep` for non-windows platform.

subhasisb: Makes sense that we should have the same name. However, PTL on windows is not yet available, so if we were to change the name in both platforms to pbs_sleep, that might be acceptable as well. Doesn't most of the PBS binaries on Windows have a underscore in the name as opposed to a dash?  If so, this is one chance to standardize on that as well.

src/tools/pbs_sleep.c
```diff
@@ -37,33 +37,47 @@
...
 		exit(1);
 	}
 
+	/* if argv[1] is -1, loop with sleep 1 indefinitely */
+	if (strcmp(argv[1], "-1")) {
+		while(1)
+			sleep(1);
+	}
+
 	sleep(atoi(argv[1]));
```
bhroam: Instead of doing one sleep(), do a loop around sleep(1) like above.  The sleep() function doesn't work well with a suspended process.  The sleep() function will sleep for N seconds, regardless if the process is suspended or not.  If the process is suspended and the sleep time is in the past, the sleep() will end when the process is resumed.  If you loop around sleep(1), one of the sleep(1) calls will end and we'll immediately start another.

src/tools/Makefile.am
bhroam: How many people do we think are using pbs-sleep on windows?  It would be bad if we suddenly changed the name of the process and broke anything a customer is doing.  I highly doubt anyone is using it except us for testing, but it's something to think about.  Do we need to deprecate pbs-sleep and add pbs_sleep for windows?  We can remove pbs-sleep one release later.

subhasisb: That sounds like a great idea Bhroam. Perhaps we should deprecate pbs-sleep. I think we also missed asking the community about this. The slight laziness about this is probably based on the fact that we have not supported Windows from 14.2 onwards and are anyways making a huge bunch of changes to windows. PTL is not on windows either, so very few tests should get affected.

Another option is add a "soft-link" in windows (junction in windows parlance) to have a pbs-sleep that points to pbs_sleep. (as a method of having both and deprecating one..eventually removing the link in the future)

src/tools/pbs_sleep.c
```diff
@@ -61,6 +62,14 @@
  *      It calls sleep internally for the number of seconds passed to it, -1 for sleep indefinitely.
  *
  */
+int getout=0;
+int i=0;
```
bhroam: Please put spaces around the '='

src/tools/pbs_sleep.c
```diff
@@ -70,13 +79,23 @@ main(int argc, char *argv[])
 		exit(1);
 	}
 
+	#ifndef WIN32
+	signal(SIGINT, handler);
+	signal(SIGHUP, handler);
+	signal(SIGTERM, handler);
```
bhroam: You shouldn't need this one.  It should just die on its own.  The default action on a SIGTERM is to terminate.

src/tools/pbs_sleep.c
```diff
@@ -61,6 +62,14 @@
  *      It calls sleep internally for the number of seconds passed to it, -1 for sleep indefinitely.
  *
  */
+int getout=0;
+int i=0;
+
+void
+handler(int sig)
+{
+	getout=1;
```
bhroam: Please put spaces around the '='

src/tools/pbs_sleep.c
```diff
@@ -70,13 +79,23 @@ main(int argc, char *argv[])
...
+		while (getout == 0)
 			sleep(1);
+		return 0;
 	}
 
-	sleep(atoi(argv[1]));
+	while (i < atoi(argv[1]) && getout ==0) {
+		sleep(1);
+		i++;
+	}
```
bhroam: You could merge these two loops.  Have a local variable called something like 'forever'.  Set forever to 1 if pbs_sleep -1 is passed in.  Do a loop that checks forever || i < N.

Plus, use a for() loop.  You're iterating.  
Also, don't put atoi() in the loop header.  You'll convert it to an int N times.  Do the conversion once.

saimanikant:
@hirenvadalia, Correction made in the same commit.
Thanks.

src/tools/pbs_sleep.c
```diff
@@ -62,40 +62,29 @@
...
-			sleep(1);
-		return 0;
-	}
+	if (strcmp(argv[1], "-1") == 0)
+		forever = 1;
+
+	secs = atoi(argv[1]);
 
-	while (i < atoi(argv[1]) && getout ==0) {
+	for (i = 0; i < atoi(argv[1]) || forever; i++)
```
hirenvadalia: atoi(argv[1]) == secs

saimanikant:
Hi @bhroam ,
I have pushed a commit addressing your comments. Please have a look.
Thanks

saimanikant:
rebased and squashed, ready for merge

subhasisb:
Please sign your commit

PR #720, opened 2018-06-12
Added test where hook is failing due to exception.

bayucan:
#### Bug/feature Description
Add a test case that causes an execjob_begin and exechost_startup hook to encounter an exception, causing the hook fail_action=offline_vnodes to get triggered. Also, the test involves a job that spans multiple host (-l place=scatter).

#### Affected Platform(s)
* All

The current testsuite only test hooks that gets an alarm signal, plus the jobs submitted are not actually spanning multiple hosts.

#### Solution Description
Testing with a bad execjob_begin hook and exechost_startup hook with hook fail_action=offline_vnodes.

#### Testing logs/output
* Testing log:
[testrun.txt](https://github.com/PBSPro/pbspro/files/2094567/testrun.txt)

#### Checklist:

### DIFF
--- a/test/tests/functional/pbs_offline_vnodes.py
+++ b/test/tests/functional/pbs_offline_vnodes.py
@@ -61,6 +61,23 @@ class TestOfflineVnodeOnHookFailure(TestFunctional):
                 'alarm': '3', 'enabled': 'true'}
         self.server.create_import_hook(name, attr, body)
 
+    def create_bad_begin_hook(self):
+        name = "h2"
+        body = ("import pbs\n"
+                "e=pbs.event()\n"
+                "if e.job.in_ms_mom():\n"
+                "    e.accept()\n"
+                "raise ValueError('invalid name')\n")
+        attr = {'event': 'execjob_begin', 'fail_action': 'offline_vnodes'}
+        self.server.create_import_hook(name, attr, body)
+
+    def create_bad_startup_hook(self):
+        name = "h3"
+        body = ("import pbs\n"
+                "raise ValueError('invalid name')\n")
+        attr = {'event': 'exechost_startup', 'fail_action': 'offline_vnodes'}
+        self.server.create_import_hook(name, attr, body)
+
     def create_multi_vnodes(self, num_moms, num_vnode=3):
         if num_moms != len(self.moms.values()):
             self.server.manager(MGR_CMD_DELETE, NODE, id="@default",
@@ -178,6 +195,14 @@ class TestOfflineVnodeOnHookFailure(TestFunctional):
                            interval=2)
         self.server.expect(JOB, {ATTR_state: 'R'}, id=jid2)
 
+        # Restore original node setup for future test cases.
+        self.server.cleanup_jobs(extend='force')
+        self.server.manager(MGR_CMD_DELETE, NODE, id="@default",
+                            expect=True)
+        for m in self.moms.values():
+            self.server.manager(MGR_CMD_CREATE, NODE,
+                                id=m.shortname)
+
     def test_multi_mom_hook_failure_affects_vnode(self):
         """
         Run an execjob_begin hook that sleeps for sometime,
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -61,10 +61,30 @@ def create_mom_hook(self):
...
+        attr = {'event': 'exechost_startup', 'fail_action': 'offline_vnodes'}
+        self.server.create_import_hook(name, attr, body)
+
     def create_multi_vnodes(self, num_moms, num_vnode=3):
-        if num_moms != len(self.moms.values()):
+        try:
             self.server.manager(MGR_CMD_DELETE, NODE, id="@default",
                                 expect=True)
+        except:
+            pass
```
bhroam: This is the wrong thing to do.  If you fail to delete all your nodes, your system will be in an unstable state for the rest of the test.  You should let manager() throw its exception.

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -61,10 +61,30 @@ def create_mom_hook(self):
...
+                "if e.job.in_ms_mom():\n"
+                "    e.accept()\n"
+                "x\n")
+        attr = {'event': 'execjob_begin', 'fail_action': 'offline_vnodes'}
+        self.server.create_import_hook(name, attr, body)
+
+    def create_bad_startup_hook(self):
+        name = "h3"
+        body = ("import pbs\n"
+                "x\n")
```
bhroam: Instead of using 'x' which will cause an exception, why not explicitly raise one?  Right now when someone reads these hooks, they might be confused why it isn't valid python.  If the hook explicitly raises an exception, it will be obvious that you are trying to cause an exception.

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -258,3 +278,64 @@ def test_multi_mom_hook_failure_affects_vnode(self):
...
+        Then run an exechost_startup hook that gets an
+        exception when local mom is restarted. Vnode representing
+        local mom would be marked offline.
+        """
+        if len(self.moms.values()) != 2:
+            self.skipTest("Provide 2 moms while invoking test")
+
+        if self.moms.values()[0].is_cpuset_mom() is True or self.moms.values()[
+                1].is_cpuset_mom() is True:
+            self.skipTest("Skipping test on cpuset moms")
```
bhroam: Consider putting this into a for loop.  It might look cleaner.

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -258,3 +278,64 @@ def test_multi_mom_hook_failure_affects_vnode(self):
...
+            self.moms.values()[1].delete_vnode_defs()
+
+        self.create_multi_vnodes(num_moms=2, num_vnode=1)
+
+        self.create_bad_begin_hook()
+
+        j1 = Job(TEST_USER)
+
+        j1.set_attributes({ATTR_l + '.select': '2:ncpus=1'})
+        j1.set_attributes({ATTR_l + '.place': 'scatter'})
```
bhroam: I've never seen two calls to set_attributes() before.  Make sure one doesn't override the other.  You might need to combine both dictionaries and make one call (or just provide it in attrs when you create the Job() object)

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -258,3 +278,64 @@ def test_multi_mom_hook_failure_affects_vnode(self):
...
+
+        jid = self.server.submit(j1)
+        self.server.expect(JOB, {ATTR_state: 'Q'}, id=jid)
```
bhroam: Move this under the expect for the sister mom.  Checking this first has a race condition.  You might check before the scheduler looks at the job yet.  The job would still be in the 'Q' state.

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -258,3 +278,64 @@ def test_multi_mom_hook_failure_affects_vnode(self):
...
+        self.server.expect(NODE, {ATTR_NODE_state: 'free'},
+                           id=self.moms.values()[0].shortname, max_attempts=3,
+                           interval=2)
+        # sister mom's vnode gets offlined due to hook exception
+        self.server.expect(NODE,
+                           {ATTR_NODE_state: 'offline',
+                            ATTR_comment:
+                            "offlined by hook 'h2' due to hook error"},
+                           id=self.moms.values()[1].shortname,
+                           max_attempts=6, interval=2)
```
bhroam: To check for two attributes in an expect(), you need to add atterop=PTL_AND.  If you don't, it'll check one or the other.

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -258,3 +278,64 @@ def test_multi_mom_hook_failure_affects_vnode(self):
...
+
+        self.create_bad_startup_hook()
```
bhroam: Consider moving this into its own test.  It is logically a different test.

test/tests/functional/pbs_offline_vnodes.py
```diff
@@ -198,6 +195,14 @@ def test_single_mom_hook_failure_affects_vnode(self):
                            interval=2)
         self.server.expect(JOB, {ATTR_state: 'R'}, id=jid2)
 
+        # Restore original node setup for future test cases.
+        self.server.cleanup_jobs(extend='force')
+        self.server.manager(MGR_CMD_DELETE, NODE, id="@default",
+                            expect=True)
+        for m in self.moms.values():
+            self.server.manager(MGR_CMD_CREATE, NODE,
+                                id=m.shortname)
```
bhroam: You shouldn't have to do this.  revert_to_defaults() should be doing this.  Does the test fail if you remove it?

bayucan: Yes

bhroam: @hirenvadalia any chance you could take a peek at the code I'm commenting on?  It's at the end of the last few tests where @bayucan has to delete all the nodes and recreate the natural vnode.  I thought this was all done in revert_to_defaults().

hirenvadalia: @bhroam No revert_to_defaults() don't delete nodes and re-create it (except on Cray), revert_to_defaults() only reverts attributes of natural vnode and it will delete all vnodes and vnode definition files.

bhroam:
I merged in the PR.

PR #729, opened 2018-06-22
If job in resv exceeds resv end time, future resv fails to confirm

arungrover:
#### Bug/feature Description
If a job is running with exclusive placement in a reservation and its walltime exceeds the reservation end time then scheduler will not confirm any other reservation on the same node until the job ends.

It turns out there were two more problems that needed to be fixed in for reservations to work - 
Problem - 1 
Scheduler crashes when it simulates a job run even where the job is part of a reservation. The reason is that job's node info array points to node in reservation and those nodes do not have node index and bucket index set on them. This caused "find_node_by_indrank()" function to dereference an out of bounds location and crash scheduler.
A check for node_ind against '-1' value fixed the problem. Similar fix was needed in update_node_on_run/end

Problem - 2
While confirming the reservation (confirm_reservation) it fails in the call to is_ok_to_run because scheduler used to think that it can not fit the reservation with the available resources in partition metadata. This check was not needed because reservations don't need this metadata and check_nodes will eventually find whether the reservation can run or not.

#### Affected Platform(s)
All

In an ideal scenario, if a job's walltime exceeds the end time of reservation, scheduler should just consider job's end time as end time of the reservation. This was not happening and because of this we were putting end-event of the job with wrong end-time in the calendar.

#### Solution Description
Changed the code to calculate the time left on the job to match the time left for the reservation.

#### Testing logs/output
[pbs_soft_walltime.txt](https://github.com/PBSPro/pbspro/files/2129493/pbs_soft_walltime.txt)
[pbs_reservations.txt](https://github.com/PBSPro/pbspro/files/2129494/pbs_reservations.txt)

#### Checklist:

### DIFF
--- a/src/scheduler/node_info.c
+++ b/src/scheduler/node_info.c
@@ -1575,7 +1575,7 @@ update_node_on_run(nspec *ns, resource_resv *resresv, char *job_state)
 	resource_resv **tmp_arr;
 	node_info *ninfo;
 	timed_event *te;
-	
+
 	if (ns == NULL || resresv == NULL)
 		return;
 
@@ -1693,17 +1693,17 @@ update_node_on_run(nspec *ns, resource_resv *resresv, char *job_state)
 			}
 		}
 	}
-	
+
 	for(te = resresv->server->calendar->next_event; te != NULL; te = te->next) {
 		if(te->event_type & TIMED_RUN_EVENT) {
 			if(te->event_ptr == resresv)
 				break;
 		}
 	}
-	
+
 	remove_te_list(&ninfo->node_events, te);
 
-	if (ninfo->bucket_ind != -1) {
+	if (ninfo->node_ind != -1 && ninfo->bucket_ind != -1) {
 		node_bucket *bkt = ninfo->server->buckets[ninfo->bucket_ind];
 		int ind = ninfo->node_ind;
 
@@ -1832,9 +1832,9 @@ update_node_on_end(node_info *ninfo, resource_resv *resresv, char *job_state)
 			}
 		}
 	}
-	
+
 	ind = ninfo->node_ind;
-	if (ind != -1) {
+	if (ind != -1 && ninfo->bucket_ind != -1) {
 		node_bucket *bkt = ninfo->server->buckets[ninfo->bucket_ind];
 
 		if (ninfo->node_events == NULL) {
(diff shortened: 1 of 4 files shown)


test/tests/functional/pbs_reservations.py
```diff
@@ -272,3 +274,221 @@ def test_resv_excl_future_resv(self):
...
+             'reserve_end': now + 3600}
+        r2 = Reservation(TEST_USER, attrs=a)
+        rid2 = self.server.submit(r2)
+
+        exp_attr = {'reserve_state': (MATCH_RE, 'RESV_CONFIRMED|2')}
+        self.server.expect(RESV, exp_attr, id=rid2)
+
+    def test_future_resv_conflicts_running_job(self):
+        """
+        Test if a running job without walltime will deny the future resv
```
bhroam: Maybe add the word 'exclusive' to the job?

arungrover:
Thanks for review @bhroam!
I've addressed your comments

src/scheduler/node_info.c
```diff
@@ -1693,17 +1693,17 @@ update_node_on_run(nspec *ns, resource_resv *resresv, char *job_state)
...
 			if(te->event_ptr == resresv)
 				break;
 		}
 	}
-	
+
 	remove_te_list(&ninfo->node_events, te);
 
-	if (ninfo->bucket_ind != -1) {
+	if (ninfo->node_ind != 1 && ninfo->bucket_ind != -1) {
```
bhroam: Shouldn't that be node_ind != -1?

arungrover: I had this weird feeling that something was not quite right there :)

arungrover:
@bhroam addressed your comments

src/scheduler/simulate.c
```diff
@@ -860,7 +860,7 @@ create_events(server_info *sinfo)
 	qsort(all, count_array((void **)all), sizeof(resource_resv *), cmp_events);
 
 	for (i = 0; all[i] != NULL && is_timed(all[i]); i++) {
-		if (all[i]->is_resv && all[i]->resv &&
+		if (all[i]->is_resv && all[i]->resv != NULL &&
```
lisa-altair: Why did you add the != NULL?  And then to only one of them instead of the first one too?

bhroam: It's the scheduler's style.  The scheduler will check if (ptr != NULL) instead of if (ptr).  This was introduced in a previous commit.  I asked Arun to fix it.

lisa-altair: Thanks for the response @bhroam.  I get it now.
Looks good to me.

arungrover:
Thanks @lisa-altair  and @bhroam  I'll rebase the branch

PR #412, opened 2017-10-06
PP-1017: Job ALPS reservation sometimes doesn't get cleared when preempt_order includes "C" (checkpoint)

arungrover:
#### Issue-ID

* **[PP-1017](https://pbspro.atlassian.net/browse/PP-1017)**

#### Problem description
On cray it has been observed that sometimes while attempting preemption, MOM ends up trying to create another ALPS reservation for the normal job.

#### Cause / Analysis
This is a timing issue where in server tries to preempt a job and by the time suspension request reaches mom the job ends. Mom then rejects the suspension request with badstate error number. This makes scheduler to try checkpointing the job and server sends checkpointing request to mom.
This reject is also rejected by mom. at the same time Obit is received by the server but the substate of the job instead of being running shows up as rerun.
This weird state/substate combination (running/rerun) makes server try to run the job again which ends up mom trying to create another ALPS reservation for a job whose previously ALPS reservation is still not deleted.

#### Solution description
Remove the code in req_holdjob.c to where we set the substate of the job to rerun. I have myself not been able to reproduce the issue, this is just by code inspection and after talking to peers that I've made this fix.

#### Checklist:

- [x] I have joined the **[pbspro community forum](http://community.pbspro.org/)**.
- [x] My pull request contains a **single, signed** commit. See **[setting up gpg signature](https://pbspro.atlassian.net/wiki/display/DG/Signing+Your+Git+Commits)**.
- [x] My code follows the **[coding style](https://pbspro.atlassian.net/wiki/display/DG/Coding+Standards)** of this project.
- [ ] My change requires project documentation. See **[required documentation checklist](https://pbspro.atlassian.net/wiki/display/DG/Checklist+for+Developing+Features+and+Bug+Fixes)** for details.
- [ ] I have added documentation in the **[project documentation area](https://pbspro.atlassian.net/wiki/display/PD)**.
- [ ] I have added new **PTL test(s) to my commit**. (See **[using PTL for testing](https://pbspro.atlassian.net/wiki/display/DG/Using+PTL+for+Testing)**) *(or)*
- [ ] I have added  **manual test(s) to the Jira ticket and explained why PTL is not appropriate** for this case.
- [x] All new and existing automated tests have passed. (See **[running automated PTL tests](https://pbspro.atlassian.net/wiki/display/DG/PTL+Quick+Start+Guide)**).
- [x] I have attached **test logs to the Jira ticket** as evidence of testing/verification.

[...]

PR #732, opened 2018-06-26
checkpoint_abort not working

prakashcv13:
#### Bug/feature Description
* checkpoint_abort is not working.
* Steps to reproduce:

           1. Set preempt_order to "C" in sched_config.
           2. Setup only one node in the complex having ncpus=4.
           3. Setup action for checkpoint_abort and restart in mom config.
           4. Have one normal priority and one express queue.
           5. Submit a job to the normal priority queue, requesting 3 ncpus confirm it starts running.
           6. Submit a job to the express queue, requesting 4 ncpus.

Expected Behavior: Job submitted to express queue should preempt the job submitted to the normal queue and starts running. The job submitted to the normal queue goes to the queued state 'Q'.
Observed Behavior: Job submitted to the normal queue remains in 'E' state for a while and then moved to hold state 'H'. Job submitted to the express queue never runs.
Logs - [steps to reproduce.txt](https://github.com/PBSPro/pbspro/files/2138508/steps.to.reproduce.txt)

#### Affected Platform(s)
* All

* This [fix](https://github.com/PBSPro/pbspro/pull/412) moved the setting of job substate to rerun from req_jobhold() to post_hold(). 
* Due to this, when MoM sends an obit for the preempted job, server was not sending the acknowledgement to the MoM as the logic falls under "rerun" block in job_obit().

#### Solution Description
* Moved the logic to send acknowledgement to MoM to the non-rerun block

#### Testing logs/output
* [testing_logs_after_fix.txt](https://github.com/PBSPro/pbspro/files/2138517/testing_logs_after_fix.txt)

#### Checklist:

### DIFF
--- a/src/server/req_jobobit.c
+++ b/src/server/req_jobobit.c
@@ -2251,6 +2251,26 @@ RetryJob:
 
 		check_block(pjob, "");		/* if block set, send word */
 
+		if ((pjob->ji_qs.ji_svrflags & JOB_SVFLG_CHKPT) &&
+			((pjob->ji_qs.ji_svrflags & JOB_SVFLG_SubJob) == 0)) {
+
+			/* non-migratable checkpoint, leave there
+			 * and just requeue the job.
+			 */
+
+			rel_resc(pjob);
+			ack_obit(stream, pjob->ji_qs.ji_jobid);
+			pjob->ji_qs.ji_svrflags |= JOB_SVFLG_HASRUN;
+			svr_evaljobstate(pjob, &newstate, &newsubst, 1);
+			(void)svr_setjobstate(pjob, newstate, newsubst);
+			if (pjob->ji_mom_prot == PROT_TCP)
+				svr_disconnect(pjob->ji_momhandle);
+
+			pjob->ji_momhandle = -1;
+			pjob->ji_mom_prot = PROT_INVALID;
+			return;
+		}
+
 		ptask = set_task(WORK_Immed, 0, on_job_exit, (void *)pjob);
 		append_link(&pjob->ji_svrtask, &ptask->wt_linkobj, ptask);
 
@@ -2269,24 +2289,6 @@ RetryJob:
 			(JOB_SVFLG_CHKPT | JOB_SVFLG_ChkptMig)) == 0) {
 			job_attr_def[(int)JOB_ATR_resc_used].at_free(
 				&pjob->ji_wattr[(int)JOB_ATR_resc_used]);
-
-		} else if ((pjob->ji_qs.ji_svrflags & JOB_SVFLG_CHKPT) &&
-			((pjob->ji_qs.ji_svrflags & JOB_SVFLG_SubJob)==0)) {
-
-			/* non-migratable checkpoint, leave there */
-			/* and just requeue the job		      */
-
-			rel_resc(pjob);
-			ack_obit(stream, pjob->ji_qs.ji_jobid);
-			pjob->ji_qs.ji_svrflags |= JOB_SVFLG_HASRUN;
-			svr_evaljobstate(pjob, &newstate, &newsubst, 1);
-			(void)svr_setjobstate(pjob, newstate, newsubst);
-			if (pjob->ji_mom_prot == PROT_TCP)
-				svr_disconnect(pjob->ji_momhandle);
-
-			pjob->ji_momhandle = -1;
-			pjob->ji_mom_prot = PROT_INVALID;
-			return;
 		}
 
 		(void)svr_setjobstate(pjob, JOB_STATE_EXITING,
--- a/test/fw/ptl/utils/pbs_crayutils.py
+++ b/test/fw/ptl/utils/pbs_crayutils.py
@@ -39,6 +39,7 @@ import socket
 import os
 
 from ptl.utils.pbs_dshutils import DshUtils
+from ptl.lib.pbs_ifl_mock import *
 
 
 class CrayUtils(object):
@@ -175,3 +176,11 @@ class CrayUtils(object):
                 key = line.split()
                 numthreads = int(key[1])
         return numthreads
+
+    def num_compute_vnodes(self, server):
+        """
+        Count the Cray compute nodes and return the value.
+        """
+        vnl = server.filter(MGR_OBJ_NODE,
+                            {'resources_available.vntype': 'cray_compute'})
+        return len(vnl["resources_available.vntype=cray_compute"])
--- a/test/tests/functional/pbs_checkpoint.py
+++ b/test/tests/functional/pbs_checkpoint.py
@@ -1,0 +1,107 @@
+# coding: utf-8
+
+# Copyright (C) 1994-2018 Altair Engineering, Inc.
+# For more information, contact Altair at www.altair.com.
+#
+# This file is part of the PBS Professional ("PBS Pro") software.
+#
+# Open Source License Information:
+#
+# PBS Pro is free software. You can redistribute it and/or modify it under the
+# terms of the GNU Affero General Public License as published by the Free
+# Software Foundation, either version 3 of the License, or (at your option) any
+# later version.
+#
+# PBS Pro is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.
+# See the GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+# Commercial License Information:
+#
+# For a copy of the commercial license terms and conditions,
+# go to: (http://www.pbspro.com/UserArea/agreement.html)
+# or contact the Altair Legal Department.
+#
+# Altair’s dual-license business model allows companies, individuals, and
+# organizations to create proprietary derivative works of PBS Pro and
+# distribute them - whether embedded or bundled with other software -
+# under a commercial license agreement.
+#
+# Use of Altair’s trademarks, including but not limited to "PBS™",
+# "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
+# trademark licensing policies.
+
+
+from tests.functional import *
+from ptl.utils.pbs_crayutils import CrayUtils
+
+
+class TestCheckpoint(TestFunctional):
+    """
+    This test suite targets Checkpoint functionality.
+    """
+    abort_file = ''
+    cu = CrayUtils()
+
+    def test_checkpoint_abort_with_preempt(self):
+        """
+        This test verifies that checkpoint_abort works as expected when
+        a job is preempted via checkpoint. It does so by submitting a job
+        in express queue which preempts a running job in the default queue.
+        """
+        abort_script = """#!/bin/bash
+kill $1
+exit 0
+"""
+        self.abort_file = self.du.create_temp_file(body=abort_script)
+        self.du.chmod(path=self.abort_file, mode=0755)
+        c = {'$action': 'checkpoint_abort 30 !' + self.abort_file + ' %sid'}
+        self.mom.add_config(c)
+
+        self.scheduler.set_sched_config({'preempt_order': 'C'})
+
+        a = {'queue_type': 'e',
+             'started': 'True',
+             'enabled': 'True',
+             'Priority': 200}
+        self.server.manager(MGR_CMD_CREATE, QUEUE, a, "expressq")
+
+        self.platform = self.du.get_platform()
+        if self.platform != 'cray' and self.platform != 'craysim':
+            attrs = {ATTR_l + '.select': '1:ncpus=1',
+                     ATTR_l + '.place': 'excl'}
+        else:
+            nv = self.cu.num_compute_vnodes(self.server)
+            attrs = {ATTR_l + '.select': '%d:ncpus=1' % nv,
+                     ATTR_l + '.place': 'scatter'}
+
+        j1 = Job(TEST_USER, attrs)
+        jid1 = self.server.submit(j1)
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid1)
+
+        attrs['queue'] = 'expressq'
+        j2 = Job(TEST_USER, attrs)
+        jid2 = self.server.submit(j2)
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid2)
+
+        self.server.expect(JOB, {'job_state': 'Q'}, id=jid1)
+        ck_dir = os.path.join(self.server.pbs_conf['PBS_HOME'], 'checkpoint',
+                              jid1 + '.CK')
+        self.assertTrue(os.path.isdir(ck_dir),
+                        msg="Checkpoint directory for job not found")
+        self.mom.log_match(msg=jid1 + ';req_holdjob: Checkpoint initiated.')
+        self.mom.log_match(msg=jid1 + ';checkpoint_abort script ' +
+                           self.abort_file + ': exit code 0')
+        self.mom.log_match(msg=jid1 + ';checkpointed to ' + ck_dir)
+        self.mom.log_match(msg=jid1 + ';task 00000001 terminated')
+
+    def tearDown(self):
+        TestFunctional.tearDown(self)
+        try:
+            os.remove(self.abort_file)
+        except OSError:
+            pass
--- a/test/tests/functional/pbs_cray_suspend_resume.py
+++ b/test/tests/functional/pbs_cray_suspend_resume.py
@@ -37,6 +37,7 @@
 
 import time
 from tests.functional import *
+from ptl.utils.pbs_crayutils import CrayUtils
 
 
 @tags('cray')
@@ -48,6 +49,7 @@ class TestSuspendResumeOnCray(TestFunctional):
     This test suite expects the platform to be 'cray' and assumes that
     suspend/resume feature is enabled on it.
     """
+    cu = CrayUtils()
 
     def setUp(self):
         if not self.du.get_platform().startswith('cray'):
@@ -210,18 +212,6 @@ at least resid .* is exclusive"
         # The job should be in the same state as it was prior to the signal
         self.server.expect(JOB, {ATTR_state: 'S', ATTR_substate: 43}, id=jid2)
 
-    def num_compute_vnodes(self):
-        """
-        Count the Cray compute nodes and return the value.
-        """
-        vlist = []
-        vnl = self.server.filter(
-            VNODE, {'resources_available.vntype': 'cray_compute'})
-        vlist = vnl["resources_available.vntype=cray_compute"]
-        nv = len(vlist)
-        self.assertNotEqual(nv, 0, "There are no cray_compute vnodes present.")
-        return nv
-
     def submit_resv(self, resv_start, chunks, resv_dur):
         """
         Function to request a PBS reservation with start time, chunks and
@@ -257,7 +247,8 @@ at least resid .* is exclusive"
         self.server.manager(MGR_CMD_SET, QUEUE, a, qname)
 
         # Reserve all the compute nodes
-        nv = self.num_compute_vnodes()
+        nv = self.cu.num_compute_vnodes(self.server)
+        self.assertNotEqual(nv, 0, "There are no cray_compute vnodes present.")
         now = time.time()
         resv_start = now + 100
         resv_dur = 7200
@@ -326,7 +317,8 @@ at least resid .* is exclusive"
         self.server.manager(MGR_CMD_CREATE, QUEUE, a, "expressq3")
 
         # Count the compute nodes
-        nv = self.num_compute_vnodes()
+        nv = self.cu.num_compute_vnodes(self.server)
+        self.assertNotEqual(nv, 0, "There are no cray_compute vnodes present.")
 
         j1 = Job(TEST_USER, {ATTR_l + '.select': '%d:ncpus=1' % nv,
                              ATTR_l + '.place': 'scatter',
@@ -372,7 +364,8 @@ at least resid .* is exclusive"
         self.server.manager(MGR_CMD_SET, SERVER, {'backfill_depth': '2'})
 
         # Count the compute nodes
-        nv = self.num_compute_vnodes()
+        nv = self.cu.num_compute_vnodes(self.server)
+        self.assertNotEqual(nv, 0, "There are no cray_compute vnodes present.")
 
         # Submit a job
         j = Job(TEST_USER, {ATTR_l + '.select': '%d:ncpus=1' % nv,


prakashcv13:
@arungrover - yes that is what is happening - and after that -
- scheduler releases the hold
- server sets the substate to queued (10).

lisa-altair:
@prakashcv13 please add a PTL test for this bug fix.

src/server/req_jobobit.c
```diff
@@ -2269,24 +2288,6 @@ job_obit(struct resc_used_update *pruu, int stream)
 			(JOB_SVFLG_CHKPT | JOB_SVFLG_ChkptMig)) == 0) {
 			job_attr_def[(int)JOB_ATR_resc_used].at_free(
 				&pjob->ji_wattr[(int)JOB_ATR_resc_used]);
-
-		} else if ((pjob->ji_qs.ji_svrflags & JOB_SVFLG_CHKPT) &&
```
arungrover: Are you sure we don't need this part of the code here? What if checkpoint_abort fails on mom and in post_hold we do not remove CHKPT flag from the job and then the job exits. I think in that case control will come here and will not reply ack to mom.
Since this is a timing issue, have you thought of order of operation where server processes obit before processing holdjob response (when job is ending at the same time it is getting held). These are some tricky scenarios and you may want to try some of those.

lisa-altair:
Also, since the culprit you point to (PP-1017) was for a Cray.  Can you please also test checkpointing with compute nodes on a Cray?  I'm worried that the ALPS reservation isn't going to be released.  So please run the second job (that causes the checkpoint) on the same compute node.

src/server/req_jobobit.c
prakashcv13: I had tested with the checkpoint_abort script returning 1 instead of 0. Checkpointing fails, but MoM decides to send an Obit that the server responds and the job is preempted.
line 2269 is when substate is rerun and checkpoint flag is NOT set.
I do not think this is a timing issue, PP-1017 was.
I believe, the testing I have done is enough, if there is a specific scenario you have in your mind, please let me know, will be happy to test it.

prakashcv13:
@lisa-altair and @arungrover - thank you for the review and the comments. I have added a PTL test case to this PR. The testing was done on Cray and local Linux. logs for the local Linux testing are attached.

Thanks again!
[ptl_local_linux.txt](https://github.com/PBSPro/pbspro/files/2143039/ptl_local_linux.txt)

test/tests/functional/pbs_checkpoint.py
```diff
@@ -0,0 +1,94 @@
...
+
+from tests.functional import *
+
+
+class TestCheckpoint(TestFunctional):
+    """
+    This test suite targets Checkpoint functionality.
+    """
+
+    def num_compute_vnodes(self):
```
lisa-altair: This same function also exists in pbs_cray_suspend_resume.py, please put this in the framework  fw/ptl/utils/pbs_crayutils.py.

test/tests/functional/pbs_checkpoint.py
```diff
@@ -0,0 +1,94 @@
...
+        self.du.chmod(path=abort_file, mode=0755)
+        c = {'$action': 'checkpoint_abort 30 !' + abort_file + ' %sid'}
+        self.mom.add_config(c)
+        self.scheduler.set_sched_config({'preempt_order': 'C'})
+        a = {'queue_type': 'e',
+             'started': 'True',
+             'enabled': 'True',
+             'Priority': 200}
+        self.server.manager(MGR_CMD_CREATE, QUEUE, a, "expressq")
+        self.platform = DshUtils().get_platform()
```
lisa-altair: I think this should be self.du.get_platform()

test/tests/functional/pbs_checkpoint.py
```diff
@@ -0,0 +1,94 @@
...
+            node_attrs = self.server.status(NODE, id=self.server.shortname)[0]
+            ncpus = node_attrs['resources_available.ncpus']
+            attrs = {'Resource_List.select': '1:ncpus=' + ncpus}
+        else:
+            nv = self.num_compute_vnodes()
+            attrs = {ATTR_l + '.select': '%d:ncpus=1' % nv,
+                     ATTR_l + '.place': 'scatter'}
+
+        j1 = Job(TEST_USER, attrs)
+        j1.set_sleep_time(5)
```
lisa-altair: Depending on the speed of the system, isn't it possible j1 could be finished before j2 can cause it to be checkpointed.
Also, if the test is running on a Cray system, will this call affect the Cray default job's aprun sleep of 100 sec?

test/tests/functional/pbs_checkpoint.py
```diff
@@ -0,0 +1,94 @@
...
+        jid1 = self.server.submit(j1)
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid1)
+        attrs['queue'] = 'expressq'
+        j2 = Job(TEST_USER, attrs)
+        j2.set_sleep_time(1)
+        jid2 = self.server.submit(j2)
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid2)
+        self.server.expect(JOB, {'job_state': 'Q'}, id=jid1)
+        self.server.delete(id=jid1)
```
lisa-altair: Since you're giving the jobs such short sleep times, can't you let both jobs finish normally?  That way we should see j2 end and j1 run again and then also end without errors.

src/server/req_jobobit.c
lisa-altair: > I had tested with the checkpoint_abort script returning 1 instead of 0. Checkpointing fails, but MoM decides to send an Obit that the server responds and the job is preempted.

I'm not sure I understand what you're saying.  If preempt_order is "C", and the Checkpoint fails, how does the server respond that the job is preempted...didn't preemption fail?

src/server/req_jobobit.c
```diff
@@ -2251,6 +2251,25 @@ job_obit(struct resc_used_update *pruu, int stream)
...
+			ack_obit(stream, pjob->ji_qs.ji_jobid);
+			pjob->ji_qs.ji_svrflags |= JOB_SVFLG_HASRUN;
+			svr_evaljobstate(pjob, &newstate, &newsubst, 1);
+			(void)svr_setjobstate(pjob, newstate, newsubst);
+			if (pjob->ji_mom_prot == PROT_TCP)
+				svr_disconnect(pjob->ji_momhandle);
+
+			pjob->ji_momhandle = -1;
+			pjob->ji_mom_prot = PROT_INVALID;
+			return;
```
lisa-altair: Why do we return here?  Wouldn't we want to continue in the code and set a work task?

test/tests/functional/pbs_checkpoint.py
prakashcv13: OK.

test/tests/functional/pbs_checkpoint.py
prakashcv13: right.

test/tests/functional/pbs_checkpoint.py
prakashcv13: makes sense.

test/tests/functional/pbs_checkpoint.py
prakashcv13: that would be testing if the restart works well and according to me that is out of scope of this bug.

src/server/req_jobobit.c
prakashcv13: the MoM sends an Obit after the checkpointing fails, the server responds to it, and the job gets preempted. The pre-emption goes well regardless of whether checkpointing was successful because the MoM is sending the Obit.

src/server/req_jobobit.c
prakashcv13: No we don't need it. The job needs to be requeued.

src/server/req_jobobit.c
lisa-altair: @prakashcv13 after digging through the code with Bhroam, I believe you are only seeing this behavior because you killed all processes but returned a non-zero exit.  Since there nothing else for the job to do it ends, and that's why you see the Obit.  What happens if your checkpoint script only returns non-zero exit (need a long enough job that it also doesn't end by then)?  According to what we saw in the code, Mom should be sending a req_reject and not an Obit.

bhroam: Just checking for the job in the Q state doesn't tell if the job was checkpointed.  You need to see if the exec_vnode attribute is also set.  In any case, this seems wrong.  If the checkpoint fails, then the qhold should be rejected.  Could you test this by hand?  Set up checkpoint where the checkpoint_abort script just returns 1.  Submit a job and then qhold it.  If it goes into the H state, something strange is going on.

prakashcv13: @lisa-altair and @bhroam - I tried this scenario and you guys are right that checkpointing will fail and so will preemption. But, isn't that an expected behavior?
What we are trying to fix here is what went wrong after the fix of PP-1017. It was the missing ack to the obit which is added now. Rest of the things are as they were.
I believe what we have tested so far is enough.

test/tests/functional/pbs_checkpoint.py
prakashcv13: I will increase the sleep time to 100 seconds.

prakashcv13:
@lisa-altair - I have addressed your comments.

test/tests/functional/pbs_checkpoint.py
bhroam: I believe the default sleep time is 100s.  If you have no reason to change it, I'd leave it as the default(don't call set_sleep_time()).  The main reason to change the sleep time is you want to wait until the jobs end.  I don't think that is the case here.

prakashcv13: To be on the safer side, I have increased it to 1000s.

prakashcv13:
@arungrover, @bhroam, @lisa-altair - please find attached the testing logs.
[ptl_local_linux_0628.txt](https://github.com/PBSPro/pbspro/files/2146391/ptl_local_linux_0628.txt)

test/tests/functional/pbs_checkpoint.py
prakashcv13: when I checked, the job was not deleted.

src/server/req_jobobit.c
```diff
@@ -2251,6 +2251,25 @@ job_obit(struct resc_used_update *pruu, int stream)
 
 		check_block(pjob, "");		/* if block set, send word */
 
+		if ((pjob->ji_qs.ji_svrflags & JOB_SVFLG_CHKPT) &&
+			((pjob->ji_qs.ji_svrflags & JOB_SVFLG_SubJob) == 0)) {
+
+			/* non-migratable checkpoint, leave there */
+			/* and just requeue the job		      */
```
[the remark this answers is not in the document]
prakashcv13: that is existing code which was moved, I will make it look standard.

src/server/req_jobobit.c
prakashcv13: If I do not kill the process, the pre-emption still happens. If the process is left lying around, it doesn't.

test/tests/functional/pbs_checkpoint.py
bhroam: That is because you only have 1 test.  Jobs get deleted in setUp() of the following test.

src/server/req_jobobit.c
bhroam: I think this is a side effect of our communication model between mom and the server.  I think what is happening is the server says, "checkpoint this job".  The checkpoint abort script fails but kills the process.  There are no processes left for the job, so the mom notices this and sends the normal end of job obit to the server.  The server was expecting an obit from the checkpoint so it assumes this is where the obit is from.  Now we have a case where the checkpoint_abort script has failed, but we checkpoint the job anyway.  In any case, this isn't part of your bug fix.  Can you file it?  Thanks again for testing it.

test/tests/functional/pbs_checkpoint.py
```diff
@@ -0,0 +1,83 @@
...
+        j1.set_sleep_time(1000)
+        jid1 = self.server.submit(j1)
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid1)
+        attrs['queue'] = 'expressq'
+        j2 = Job(TEST_USER, attrs)
+        j2.set_sleep_time(1)
+        jid2 = self.server.submit(j2)
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid2)
+        self.server.expect(JOB, {'job_state': 'Q'}, id=jid1)
+        self.server.delete(id=jid1)
```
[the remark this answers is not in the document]
lisa-altair: @vccardenas I don't agree with this statement.  @prakashcv13 has already written a test that reproduces the bug.  I see no reason for him to write a test for every way that the bug can be reproduced...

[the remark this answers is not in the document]
iestockdale: The fix should be addressing all observed failure modes.  In order to show that, there need to be PTL tests for each of those.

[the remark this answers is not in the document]
bhroam: I agree with @lisa-altair.  There are three methods of triggering checkpointing.  All three send a checkpoint command to the mom.  Everything after that is the same.  @prakashcv13 changed code after the mom responds to the checkpoint command she was given.  He wrote a test which used one of the three methods of triggering checkpointing, since his fix has nothing to do with how one triggers checkpointing.  It has to do with how the job is checkpointed.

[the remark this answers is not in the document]
iestockdale: The reasons to have the PTL tests for the bug fix include ensuring that the assumptions behind the fix are correct and that all affected scenarios beneft from the fix.  The fix may be independent of the triggering method, but the consequences of the fix could be different (or not).  The way to show which it is with a test.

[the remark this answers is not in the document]
subhasisb: @iestockdale while i agree with you that we must have tests for the 3 different methods of checkpointing, i agree with the others that for this particular fix we do not need the contributor to cover those tests. What he has used to test well covers his area of change. I feel we should raise a separate bug to add test coverage in the other triggering methods, but that is beyond the scope of this fix alone.

src/server/req_jobobit.c
lisa-altair: Okay, thanks for explaining.

lisa-altair:
@prakashcv13 thank you for moving num_compute_vnodes code to CrayUtils.  Earlier I mentioned that the same function also exists in pbs_cray_suspend_resume.py.  To finish creating a generic function, please update pbs_cray_suspend_resume.py to use the CrayUtils function instead of the local copy of the same function.  Thank you.

test/tests/functional/pbs_checkpoint.py
```diff
@@ -0,0 +1,83 @@
...
+from tests.functional import *
+from ptl.utils.pbs_crayutils import CrayUtils
+
+
+class TestCheckpoint(TestFunctional):
+    """
+    This test suite targets Checkpoint functionality.
+    """
+
+    def test_checkpoint_abort_with_preemept(self):
```
vccardenas: @prakashcv13 , please add a docstring for this test.

test/fw/ptl/utils/pbs_crayutils.py
```diff
@@ -175,3 +175,15 @@ def get_numthreads(self, nid):
...
+    def num_compute_vnodes(self):
+        """
+        Count the Cray compute nodes and return the value.
+        """
+        vlist = []
+        vnl = self.server.filter(
+            VNODE, {'resources_available.vntype': 'cray_compute'})
+        vlist = vnl["resources_available.vntype=cray_compute"]
+        nv = len(vlist)
+        self.assertNotEqual(nv, 0, "There are no cray_compute vnodes present.")
```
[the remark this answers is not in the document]
prakashcv13: While I can change the docstring and the function name, I think having the utility function just to count the number of vnodes is sufficient even though it would have only one single return statement because there could be tests to confirm that the number of vnodes is equal to 0. So, I am leaving the decision to assert or not up to the user of the utility function.

test/tests/functional/pbs_checkpoint.py
```diff
@@ -0,0 +1,83 @@
...
+        abort_script = """#!/bin/bash
+kill $1
+exit 0
+"""
+        abort_file = self.du.create_temp_file(body=abort_script)
+        self.du.chmod(path=abort_file, mode=0755)
```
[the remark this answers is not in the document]
prakashcv13: It doesn't seem like an issue to the MoM.

test/tests/functional/pbs_checkpoint.py
```diff
@@ -0,0 +1,83 @@
...
+        c = {'$action': 'checkpoint_abort 30 !' + abort_file + ' %sid'}
+        self.mom.add_config(c)
+        self.scheduler.set_sched_config({'preempt_order': 'C'})
+        a = {'queue_type': 'e',
+             'started': 'True',
+             'enabled': 'True',
+             'Priority': 200}
+        self.server.manager(MGR_CMD_CREATE, QUEUE, a, "expressq")
```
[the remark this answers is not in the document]
prakashcv13: done!

test/tests/functional/pbs_checkpoint.py
```diff
@@ -0,0 +1,83 @@
...
+        self.platform = self.du.get_platform()
+        if self.platform != 'cray' and self.platform != 'craysim':
+            node_attrs = self.server.status(NODE, id=self.server.shortname)[0]
+            ncpus = node_attrs['resources_available.ncpus']
+            attrs = {'Resource_List.select': '1:ncpus=' + ncpus}
```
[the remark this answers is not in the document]
prakashcv13: done!

test/tests/functional/pbs_checkpoint.py
```diff
@@ -0,0 +1,83 @@
...
+        j1 = Job(TEST_USER, attrs)
+        j1.set_sleep_time(1000)
+        jid1 = self.server.submit(j1)
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid1)
+        attrs['queue'] = 'expressq'
+        j2 = Job(TEST_USER, attrs)
+        j2.set_sleep_time(1)
+        jid2 = self.server.submit(j2)
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid2)
+        self.server.expect(JOB, {'job_state': 'Q'}, id=jid1)
```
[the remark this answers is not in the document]
prakashcv13: removed the sleep time.

test/tests/functional/pbs_checkpoint.py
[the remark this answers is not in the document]
prakashcv13: thank you everyone!

test/tests/functional/pbs_checkpoint.py
prakashcv13: added.

test/tests/functional/pbs_checkpoint.py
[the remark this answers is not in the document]
bhroam: Are you running pbs_benchpress as root?  I may be mistaken about the location of the file, but I'm pretty sure she makes sure the script is root owned.  PTL scripts are supposed to be able to be run as a non-root user.

prakashcv13:
Thank you all for the review comments, I have addressed them. Please have a look.

src/server/req_jobobit.c
```diff
@@ -2254,8 +2254,8 @@ job_obit(struct resc_used_update *pruu, int stream)
 		if ((pjob->ji_qs.ji_svrflags & JOB_SVFLG_CHKPT) &&
 			((pjob->ji_qs.ji_svrflags & JOB_SVFLG_SubJob) == 0)) {
 
-			/* non-migratable checkpoint, leave there */
-			/* and just requeue the job		      */
+			/* non-migratable checkpoint, leave there
+			 * and just requeue the job. */
```
bhroam: Put the closing */ on the line below

test/fw/ptl/utils/pbs_crayutils.py
```diff
@@ -176,14 +177,11 @@ def get_numthreads(self, nid):
...
-        vlist = []
-        vnl = self.server.filter(
-            VNODE, {'resources_available.vntype': 'cray_compute'})
+        vnl = server.filter(MGR_OBJ_NODE,
+                            {'resources_available.vntype': 'cray_compute'})
         vlist = vnl["resources_available.vntype=cray_compute"]
-        nv = len(vlist)
-        self.assertNotEqual(nv, 0, "There are no cray_compute vnodes present.")
-        return nv
+        return len(vlist)
```
bhroam: Can you just put vnl[] in the len directly and get rid of vlist and still have the line under 79 characters?  Actually if you wanted to shrink the function even further, you could just return len(server.filter())

test/tests/functional/pbs_checkpoint.py
```diff
@@ -44,40 +44,67 @@ class TestCheckpoint(TestFunctional):
...
+    cu = CrayUtils()
 
-    def test_checkpoint_abort_with_preemept(self):
+    def test_checkpoint_abort_with_preempt(self):
+        """
+        This test verifies that checkpoint_abort works as expected by
+        1. setting preempt_order in sched_config to 'C'.
+        2. submitting a job to the default queue.
+        3. submitting a job to the express queue.
+        4. confirming that the first job was checkpointed.
```
bhroam: Don't put the test steps in the doc string.  Just describe the test.

test/tests/functional/pbs_checkpoint.py
```diff
@@ -44,40 +44,67 @@ class TestCheckpoint(TestFunctional):
...
         attrs['queue'] = 'expressq'
         j2 = Job(TEST_USER, attrs)
-        j2.set_sleep_time(1)
         jid2 = self.server.submit(j2)
         self.server.expect(JOB, {'job_state': 'R'}, id=jid2)
+
         self.server.expect(JOB, {'job_state': 'Q'}, id=jid1)
-        self.server.delete(id=jid1)
+        ck_dir = os.path.join(self.server.pbs_conf['PBS_HOME'], 'checkpoint/' +
+                              jid1 + '.CK')
```
bhroam: Let os.path.join() combine your path, don't add a string with a '/' in it.

test/tests/functional/pbs_checkpoint.py
```diff
@@ -44,40 +44,67 @@ class TestCheckpoint(TestFunctional):
...
+        self.assertEqual(os.path.isdir(ck_dir), True,
+                         msg="Checkpoint directory for job not found")
```
bhroam: Use self.assertTrue()

test/fw/ptl/utils/pbs_crayutils.py
prakashcv13: len(server.filter()) would return 1. I will get rid of vlist by passing vnl[] to len().

test/tests/functional/pbs_checkpoint.py
prakashcv13: done!

test/tests/functional/pbs_checkpoint.py
prakashcv13: done!

test/tests/functional/pbs_checkpoint.py
prakashcv13: ok, done!

prakashcv13:
@bhroam - I have addressed your review comments - please have a look.

.gitignore
```diff
@@ -193,6 +193,7 @@ src/tools/printjob.bin
 src/tools/printjob_svr.bin
 src/tools/tracejob
 src/tools/wrap_tcl.sh
+src/tools/pbs_sleep
```
bhroam: Check if this is already there.  I added pbs_sleep to the .gitignore file a couple of weeks ago.

test/tests/functional/pbs_checkpoint.py
```diff
@@ -49,11 +49,9 @@ class TestCheckpoint(TestFunctional):
...
     def test_checkpoint_abort_with_preempt(self):
         """
-        This test verifies that checkpoint_abort works as expected by
-        1. setting preempt_order in sched_config to 'C'.
-        2. submitting a job to the default queue.
-        3. submitting a job to the express queue.
-        4. confirming that the first job was checkpointed.
+        This test verifies that checkpoint_abort works as expected when
+        preempt_order in sched_config is set to 'C' and a job in express
+        queue preempts a job in the default queue.
```
bhroam: I'd rephrase this.  How about 'when a job is preempted via checkpoint'.  That is what you are really testing.  There are other ways to cause it to happen than what you are doing which would test the same thing.  The real thing you are testing is preemption via checkpointing.

bhroam:
Travis failed due to pep8

test/tests/functional/pbs_checkpoint.py
```diff
@@ -44,40 +44,67 @@ class TestCheckpoint(TestFunctional):
...
+
         a = {'queue_type': 'e',
              'started': 'True',
              'enabled': 'True',
              'Priority': 200}
         self.server.manager(MGR_CMD_CREATE, QUEUE, a, "expressq")
+
         self.platform = self.du.get_platform()
         if self.platform != 'cray' and self.platform != 'craysim':
             node_attrs = self.server.status(NODE, id=self.server.shortname)[0]
```
lisa-altair: since the you changed how you form the job request, node_attrs is now no longer used.  Please remove it.

test/tests/functional/pbs_cray_suspend_resume.py
```diff
@@ -48,6 +49,7 @@ class TestSuspendResumeOnCray(TestFunctional):
     This test suite expects the platform to be 'cray' and assumes that
     suspend/resume feature is enabled on it.
     """
+    cu = CrayUtils()
```
lisa-altair: Thanks for making these changes!

lisa-altair:
@prakashcv13 please attach test logs from running your PTL test on a Linux system.  Thanks.

prakashcv13:
@lisa-altair - attaching the logs.
[test_local_linux_0711.txt](https://github.com/PBSPro/pbspro/files/2190673/test_local_linux_0711.txt)

lisa-altair:
Thanks @prakashcv13!  LGTM.

prakashcv13:
@bhroam, @vccardenas - could you please let me know if there are further comments.

prakashcv13:
@lisa-altair, @bhroam and @vcc, thank you for the review and approvals/sign-off. I have squashed my commits and rebased my branch. Request @PBSPro/pbspro-maintainers - to take it forward.

bhroam:
I merged in the PR.

PR #736, opened 2018-06-29
test_cgroup_cpuset failing with hyperthreads present

mike0042:
Signed-off-by: Michael Karo <mike0042@gmail.com>

#### Bug/feature Description
* The test_cgroup_cpuset test was failing on systems with hyperthreads

#### Affected Platform(s)
* Linux w/cgroups available

* The cgroup hook counts cores differently on systems where hyperthreads are present. When two jobs were submitted only one would run because the cgroup hook thought the node was fully allocated, thereby causing the test to fail.

#### Solution Description
* Update the test to submit both jobs immediately, reducing race conditions.
* Update the test so that use_hyperthreads is enabled in the cgroup configuration.
* Update the cgroup hook to count cores and threads correctly.
* In the past, if the test system had more than two cores the test would succeed which is likely why this is only observed on a two core system.
* No new PTL tests are necessary because the one in question is now passing.

#### Testing logs/output
* To be added shortly

#### Checklist:


test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -1232,52 +1232,56 @@ def test_cgroup_cpuset(self):
...
              self.hostA, ATTR_N: name + 'b'}
         j2 = Job(TEST_USER, attrs=b)
         j2.create_script(self.cpuset_mem_script)
         jid2 = self.server.submit(j2)
         a = {'job_state': 'R'}
+        # Make sure they are both running
+        self.server.expect(JOB, a, jid1)
         self.server.expect(JOB, a, jid2)
+        # Status the jobs for their output files
+        attrib = [ATTR_o, 'exec_host']
```
vstumpf: Do you still need exec_host?

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -3089,35 +3090,30 @@ class CgroupUtils(object):
                         empty_cores.add(corenum)
                 # Assign threads from the empty cores
                 for corenum in empty_cores:
+                    assigned['cpuset.cpus'].append(corenum)
```
vstumpf: I don't fully understand how the data is stored, but if we have a node with 4 empty cores, each with 2 threads, we should have 8 threads. If we request 4 cpus, then needed = 4.
We check the first core, append it to `cpuset.cpus`, remove it from `cores`, then reduce `needed` by 1 `\\needed = 3`. 
We check the first thread of first core, append it to `cpuset.cpus`, remove it from `cores`, then reduce `needed` by 1 `\\needed = 2`. 
We check the second thread of first core, append it to `cpuset.cpus`, remove it from `cores`, then reduce `needed` by 1 `\\needed = 1`. 
We check the second core, append it to `cpuset.cpus`, remove it from `cores`, then reduce `needed` by 1 `\\needed = 0`. Because `needed == 0`, we break out. 

At the end, we have core1, core1threadA, core1threadB, and core2. 
Shouldn't it be core1threadA, core1threadB, core2threadA, core2threadB?

test/tests/functional/pbs_cgroups_hook.py
mike0042: I think we still need it because it is used in the self.server.status() calls that follow.

vstumpf: I don't see exec_host used; since you put the host in the select line, you (correctly) assume that the host will be hostA.

src/hooks/cgroups/pbs_cgroups.PY
mike0042: The _discover_cpuinfo() method is responsible for populating self.cpuinfo in the NodeConfig class. The available_node_resources() method then removes assigned resources from the dictionary, resulting in the "available" dictionary that gets passed to _assign_resources(). At this point each physical CPU has a thread list associated with it (node.cpuinfo['cpu'][corenum]['threads']). When use_hyperthreads is enabled, the hook attempts to assign empty core/thread sets before it resorts to assigning those that are partially allocated.

Using your example of an empty 4 core (8 thread) node...
- All four cores are identified as empty
- The first core is assigned to the job (needed = 3)
- The thread associated with the previously assigned core is assigned to the job (needed = 2)
- The second core is assigned to the job (needed = 1)
- The thread associated with the second core is assigned to the job (needed = 0)
- We break from the loop because needed == 0

Ultimately, we end up with what you thought we should... core1threadA, core1threadB, core2threadA, core2threadB.

The test was failing on a hyperthread enabled system because it was attempting to place two jobs on different threads of the same core. The hook used to (erroneously) reject the request. It was a corner case when ncpus=1 and there was only one thread available to assign.

test/tests/functional/pbs_cgroups_hook.py
mike0042: Ahh... I think I see your point now. We're asking for the status of two attributes, when we really only need ATTR_o. I'll remove the 'exec_host' attribute from the list.

src/hooks/cgroups/pbs_cgroups.PY
vstumpf: I see; so each cpu has a thread, not two threads.

mike0042:
@PBSPro/pbspro-maintainers Please take this forward.

src/hooks/cgroups/pbs_cgroups.CF
```diff
@@ -7,6 +7,7 @@
     "vnode_per_numa_node"   : false,
     "online_offlined_nodes" : true,
     "use_hyperthreads"      : false,
+    "ncpus_are_cores"       : false,
```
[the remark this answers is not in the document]
mike0042: This was a remnant from another contributor. It's been there for a while now, but the value wasn't getting properly initialized. The design document has been updated here: https://pbspro.atlassian.net/wiki/spaces/PD/pages/11599882/PP-325+Support+Cgroups

What's an EDD?  ;-)

src/hooks/cgroups/pbs_cgroups.PY
mike0042: Correct.

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -1232,52 +1232,56 @@ def test_cgroup_cpuset(self):
...
         j2 = Job(TEST_USER, attrs=b)
         j2.create_script(self.cpuset_mem_script)
         jid2 = self.server.submit(j2)
         a = {'job_state': 'R'}
+        # Make sure they are both running
+        self.server.expect(JOB, a, jid1)
         self.server.expect(JOB, a, jid2)
+        # Status the jobs for their output files
+        attrib = [ATTR_o]
+        self.server.status(JOB, attrib, jid1)
```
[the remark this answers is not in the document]
mike0042: I honestly don't know. All I can say is that is how it was done previously and it produces the desired behavior. If you would like me to change it, please provide a specific suggestion on how it should be implemented.

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -1232,52 +1232,56 @@ def test_cgroup_cpuset(self):
...
-            % (jid2, tmp_host2))
+            % (jid2, self.hostA))
         self.logger.info('job dir check passed')
+        # Ensure the CPU ID for each job differs
+        cpuid1 = None
         for kv in tmp_out1:
             if 'CpuIDs=' in kv:
                 cpuid1 = kv
                 break
+        self.assertNotEqual(cpuid1, None, 'Could not read first CPU ID.')
```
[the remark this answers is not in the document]
mike0042: Good question. I used assertNotEqual() because cpuid1 is initialized to None, so that's what we're comparing against. It seems to work as-is, but I'm unaware if there is a preferred way to test the value.

test/tests/functional/pbs_cgroups_hook.py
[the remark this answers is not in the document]
bhroam: It just saves you the creation of the local variable is all.  I'll leave it up to you what you want to do.

test/tests/functional/pbs_cgroups_hook.py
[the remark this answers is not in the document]
bhroam: The way you have it is probably best then.

mike0042:
Squashed and rebased.

PR #754, opened 2018-07-24
checkpoint_abort not working when server is shutdown using qterm im…

prakashcv13:
…mediate and delay.

#### Bug/feature Description
* checkpoint_abort was not working when the server is shutdown using qterm -t immediate (and delay)
* This issue was found by running tests that are being reviewed under [https://github.com/PBSPro/pbspro/pull/749](PR#749)

#### Affected Platform(s)
* All

* checkpoint_abort was working in other scenarios.
* At the time of shutdown, the server was marking the substate of the job to be rerun before the MoM replied back on the checkpoint request.
* Due to this, in job_obit(), the flow was jumping to the rerun logic, which lead the server to send a delete job request.
#### Solution Description
* To not set the substate to rerun, the new state and substate will be evaluated and set in job_obit().

#### Testing logs/output
* Same 4 PTL test cases being reviewed in PR#749 were run. All test cases passed.
[ptl_0724.txt](https://github.com/PBSPro/pbspro/files/2224169/ptl_0724.txt)

#### Checklist:

### DIFF
--- a/src/include/job.h
+++ b/src/include/job.h
@@ -891,6 +891,7 @@ task_find	(job		*pjob,
 #define JOB_SVFLG_ChkptMig 0x100 /* job has migratable checkpoint */
 #define JOB_SVFLG_Suspend  0x200 /* job suspended (signal suspend) */
 #define JOB_SVFLG_StagedIn 0x400 /* job has files that have been staged in */
+#define JOB_SVFLG_HASHOLD  0x800 /* job has a hold request sent to MoM */
 #define JOB_SVFLG_HasNodes 0x1000 /* job has nodes allocated to it */
 #define JOB_SVFLG_RescAssn 0x2000 /* job resources accumulated in server/que */
 #define JOB_SVFLG_SPSwitch 0x2000 /* SP switch loaded for job, SP MOM only */
--- a/src/server/req_holdjob.c
+++ b/src/server/req_holdjob.c
@@ -209,7 +209,7 @@ req_holdjob(struct batch_request *preq)
 			req_reject(rc, 0, preq);
 		} else {
 			pjob->ji_qs.ji_svrflags |=
-				JOB_SVFLG_HASRUN | JOB_SVFLG_CHKPT;
+				(JOB_SVFLG_HASRUN | JOB_SVFLG_CHKPT | JOB_SVFLG_HASHOLD);
 			(void)job_save(pjob, SAVEJOB_QUICK);
 			log_event(PBSEVENT_JOB, PBS_EVENTCLASS_JOB, LOG_INFO,
 				pjob->ji_qs.ji_jobid, log_buffer);
--- a/src/server/req_shutdown.c
+++ b/src/server/req_shutdown.c
@@ -323,9 +323,7 @@ shutdown_chkpt(job *pjob)
 		if (pjob->ji_qs.ji_state == JOB_STATE_TRANSIT)
 			svr_setjobstate(pjob, JOB_STATE_RUNNING, JOB_SUBSTATE_RUNNING);
 
-		pjob->ji_qs.ji_substate = JOB_SUBSTATE_RERUN;
-		pjob->ji_qs.ji_svrflags |= JOB_SVFLG_HASRUN;
-		pjob->ji_qs.ji_svrflags |= JOB_SVFLG_CHKPT;
+		pjob->ji_qs.ji_svrflags |= (JOB_SVFLG_HASRUN | JOB_SVFLG_CHKPT | JOB_SVFLG_HASHOLD);
 		pjob->ji_modified = 1;
 		(void)job_save(pjob, SAVEJOB_QUICK);
 		return (0);
--- a/src/server/req_jobobit.c
+++ b/src/server/req_jobobit.c
@@ -2249,10 +2249,9 @@ RetryJob:
 		((pjob->ji_qs.ji_substate != JOB_SUBSTATE_RERUN) &&
 		(pjob->ji_qs.ji_substate != JOB_SUBSTATE_RERUN2))) {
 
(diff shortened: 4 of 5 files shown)


arungrover:
@prakashcv13 with this change now during server shutdown we will end up calling 'check_block' function in post job obit callback. Can you please run a test with and without your fix where there is a blocking job running on the system and see what happens when you run your test manually (one that reproduced the problem).
I suspect there might be some difference in behavior for blocking jobs.

arungrover:
Thanks for confirming that there will be a behavior change. 
Since it is going to be visible to users, it qualifies as an external behavior change. 
It will attract a design change and a forum discussion on whether or not it is the right thing to do.

prakashcv13:
@arungrover - I understand now what you were saying. The behavior change (calling check_block()) is due to the [PR-732](https://github.com/PBSPro/pbspro/pull/732). I will work on not calling check_block() for jobs that are being checkpointed.
@vccardenas - So far writing PTL with blocking jobs will block the test until the job gets over. I will work on using run_cmd to qsub a job with -Wblock.

prakashcv13:
@arungrover, @lisa-altair, and @vcc - I have made the below changes -
1) Moved the call to check_block() after the checkpoint logic.
2) Changed PTL test cases so that they pass even if run by a user instead of root.
3) Checkpointed jobs on completion were getting re-queued as the checkpoint logic in job_obit() did not differentiate between Obit sent after checkpointing the job and the one sent after actual job completion, so added a new flag.
4) Set the new flag in req_holdjob() and shutdown_chkpt().
5) Reset the new flag when Obit is received for a hold job request.

All the test cases pass when run as a user and root, logs are attached.
[ptl.txt](https://github.com/PBSPro/pbspro/files/2284964/ptl.txt)
[ptl_user_logs.txt](https://github.com/PBSPro/pbspro/files/2284965/ptl_user_logs.txt)

Request you to review the changes.

src/server/req_holdjob.c
```diff
@@ -210,6 +210,7 @@ req_holdjob(struct batch_request *preq)
 		} else {
 			pjob->ji_qs.ji_svrflags |=
 				JOB_SVFLG_HASRUN | JOB_SVFLG_CHKPT;
+			pjob->ji_qs.ji_svrflags |= JOB_SVFLG_HASHOLD;
```
arungrover: I think this flag can move inside the previous statement itself.

I do have a question, You are adding this flag because job_obit() is unable to identify whether the job it is dealing with was checkpointed or has actually ended. But, my point of confusion is that why does the job holds on to JOB_SVFLG_CHKPT? When it is requeued the first time after checkpointing shouldn't we just remove CHKPT flag from the job.
I'm probably missing something here.

src/include/job.h
```diff
@@ -891,6 +891,7 @@ task_find	(job		*pjob,
 #define JOB_SVFLG_ChkptMig 0x100 /* job has migratable checkpoint */
 #define JOB_SVFLG_Suspend  0x200 /* job suspended (signal suspend) */
 #define JOB_SVFLG_StagedIn 0x400 /* job has files that have been staged in */
+#define JOB_SVFLG_HASHOLD  0x800 /* job has a hold request sent to MoM */
```
arungrover: I'm surprised that you actually found a void in the job flags :)

src/server/req_shutdown.c
```diff
@@ -323,9 +323,9 @@ shutdown_chkpt(job *pjob)
 		if (pjob->ji_qs.ji_state == JOB_STATE_TRANSIT)
 			svr_setjobstate(pjob, JOB_STATE_RUNNING, JOB_SUBSTATE_RUNNING);
 
-		pjob->ji_qs.ji_substate = JOB_SUBSTATE_RERUN;
```
lisa-altair: Why did you remove setting the substate?

src/server/req_shutdown.c
```diff
@@ -323,9 +323,9 @@ shutdown_chkpt(job *pjob)
...
 		pjob->ji_qs.ji_svrflags |= JOB_SVFLG_HASRUN;
 		pjob->ji_qs.ji_svrflags |= JOB_SVFLG_CHKPT;
+		pjob->ji_qs.ji_svrflags |= JOB_SVFLG_HASHOLD;
```
lisa-altair: I think it would be best to OR all these flags together and then set ji_svrflags.

prakashcv13:
@arungrover, @lisa-altair - thank you for the review. I have OR-ed the flags in one statement as per the review comments.

src/include/job.h
lisa-altair: Yup!  Good idea.

src/server/req_shutdown.c
lisa-altair: Ah.  Okay, sorry I missed it.

prakashcv13:
Thank you @lisa-altair for the sign-off.
@arungrover - could you please let me know if you have any further comments.

prakashcv13:
Thank you @lisa-altair and @arungrover for the review and sign-off. I have squashed my commits and rebased my branch. 
@lisa-altair - as we decided, I have created [PP-1295](https://pbspro.atlassian.net/browse/PP-1295) for removing the duplicate flags.

vccardenas:
@prakashcv13, the PTL tests look good to me.

prakashcv13:
Thank you @vccardenas for the sign-off. I believe we can now take this PR forward.

prakashcv13:
@PBSPro/pbspro-maintainers - could you please merge this PR.

mike0042:
Please rebase your commit @prakashcv13.

prakashcv13:
Thank you @mike0042, I just rebased.

PR #755, opened 2018-07-24
fix pbs_probe realpath() buffer overflow on ubuntu 16

minghui-liu:
#### Bug/feature Description
Running pbs_probe on Ubuntu 16 fails becaue of a buffer overflow.

#### Affected Platform(s)
* Ubuntu 16

Call to `realpath()` in `get_realpath_value()` will put real path in array `real`, which was `MAXPATHLEN` long. On most systems this is defined to be 1024 which is not large enough. 

#### Solution Description
A length of `PATH_MAX` (defined to be 4096 in `limits.h` on most linux systems) should be used instead.

#### Testing logs/output
No new PTL test needed as test_pbs_probe in SmokeTest will be enough to cover this fix. 
* Ubuntu 16: [test_pbs_probe.txt](https://github.com/PBSPro/pbspro/files/2255121/test_pbs_probe.txt)
* RHEL 7: [test_pbs_probe_r7.txt](https://github.com/PBSPro/pbspro/files/2291963/test_pbs_probe_r7.txt)
* SLES 12: [test_pbs_probe_s12.txt](https://github.com/PBSPro/pbspro/files/2291964/test_pbs_probe_s12.txt)

#### Checklist:

### DIFF
--- a/src/lib/Liblog/chk_file_sec.c
+++ b/src/lib/Liblog/chk_file_sec.c
@@ -231,23 +231,17 @@ tempstat(struct stat *sp, int isdir, int sticky, int disallow)
  *              			ENOTDIR if not file/directory as specified
  *              			EACCESS if permissions are not ok
  */
-
 int
-chk_file_sec(path, isdir, sticky, disallow, fullpath)
-char	*path;		/* path to check */
-int	isdir;		/* 1 = path is directory, 0 = file */
-int	sticky;		/* allow write on directory if sticky set */
-int	disallow;	/* perm bits to disallow */
-int	fullpath;	/* check full path */
+chk_file_sec(char *path, int isdir, int sticky, int disallow, int fullpath)
 {
 	int		rc = 0;
 	struct	stat	sbuf;
-	char		real[MAXPATHLEN];
+	char 	*real = NULL;
 
 	assert(path != NULL);
 	assert(path[0] != '\0');
 
-	if (realpath(path, real) == NULL) {
+	if ((real = realpath(path, NULL)) == NULL) {
 		rc = errno;
 		goto chkerr;
 	}
@@ -342,6 +336,7 @@ chkerr:
 		}
 	}
 
+	free(real);
 	return (rc);
 }
 
@@ -374,12 +369,12 @@ tmp_file_sec(char *path, int isdir, int sticky, int disallow, int fullpath)
 {
 	int		rc = 0;
 	struct	stat	sbuf;
-	char		real[MAXPATHLEN];
+	char	*real = NULL;
 
 	assert(path != NULL);
 	assert(path[0] != '\0');
 
-	if (realpath(path, real) == NULL) {
+	if ((real = realpath(path, NULL)) == NULL) {
 		rc = errno;
 		goto chkerr;
 	}
@@ -455,7 +450,7 @@ chkerr:
 	if (rc != 0) {
 		char	*error_buf;
 
-		if ((error_buf = malloc(1024)) == 0) {
+		if ((error_buf = malloc(1024)) == NULL) {
 			log_err(rc, "chk_file_sec", "Malloc failed");
 		} else {
 			sprintf(error_buf,
@@ -472,5 +467,6 @@ chkerr:
 		}
 	}
 
+	free(real);
 	return (rc);
 }
--- a/src/lib/Libwin/util.c
+++ b/src/lib/Libwin/util.c
@@ -52,6 +52,9 @@
 #include <avltree.h>
 
 #define TIME_SIZE 26 /*String length of time string is 26*/
+#ifndef PATH_MAX
+#define PATH_MAX 4096
+#endif
 
 BOOL is_user_impersonated = FALSE;
 static AVL_IX_DESC *env_avltree = NULL;
@@ -242,7 +245,10 @@ fcntl(int fd, int cmd, long arg)
  * 	just a placeholder for the unix equivalent function.
  *
  * @param[in] path - file path
- * @param[in] resolved_path - resolved path
+ * @param[out] resolved_path - resolved path, if NULL is passed in, then
+ *     this function will allocate PATH_MAX bytes of memory and return the
+ *     address as return value. It is the caller's responsibility to free
+ *     this memory.
  *
  * @return	string
  * @retval	resolved path	succees
@@ -251,6 +257,12 @@ fcntl(int fd, int cmd, long arg)
 char *
 realpath(const char *path, char *resolved_path)
 {
+	if (resolved_path == NULL) {
+		if ((resolved_path = malloc(PATH_MAX)) == NULL) {
+			fprintf(stderr, "realpath() failed to allocate memory\n");
+			return NULL;
+		}
+	}
 	strcpy(resolved_path, path);
 	back2forward_slash(resolved_path);
 	return (resolved_path);
--- a/src/resmom/stage_func.c
+++ b/src/resmom/stage_func.c
@@ -147,23 +147,31 @@ add_bad_list(char **pbl, char *newtext, int nl)
  * @param[in]	path	-	path to check
  *
  * @return	int
- * @retval	FALSE - given not a child path
- * @retval	TRUE - given path is child of dir
- *
+ * @retval	1 - given path is child of dir
+ * @retval	0 - given not a child path
+ * @retval -1 - error encountered
  */
 int
 is_child_path(char *dir, char *path)
 {
 	char fullpath[2*MAXPATHLEN+2] = {'\0'};
-	char dir_real[MAXPATHLEN+1] = {'\0'};
-	char fullpath_real[2*MAXPATHLEN+2] = {'\0'};
+	char *dir_real = NULL;
+	char *fullpath_real = NULL;
 	char *pos = NULL;
+	int  return_value = 0;
 
 	/* if file path is relative, combine it with directory */
 	if (!is_full_path(path)) {
-		(void)snprintf(fullpath, sizeof(fullpath)-1, "%s/%s", dir, path);
+		(void)snprintf(fullpath, sizeof(fullpath), "%s/%s", dir, path);
 	} else {
-		strncpy(fullpath, path, sizeof(fullpath)-1);
+		strncpy(fullpath, path, sizeof(fullpath));
+	}
+
+	dir_real = malloc(sizeof(char) * (MAXPATHLEN + 1));
+	fullpath_real = malloc(sizeof(char) * (2 * MAXPATHLEN + 2));
+	if (dir_real == NULL || fullpath_real == NULL) {
+		return_value = -1;
+		goto error_exit;
 	}
 
 	/* even if the file path is relative to some directory, */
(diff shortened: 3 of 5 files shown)


src/tools/pbs_probe.c
```diff
@@ -2244,7 +2244,7 @@ pbsdotconf(char * path)
 static	int
 get_realpath_values(struct infrastruct *pinf)
 {
-	char real[MAXPATHLEN];
+	char real[PATH_MAX+1];
```
mike0042: According to the realpath(3) man page...
> The resulting pathname is stored as  a  null-terminated  string, up to a maximum of PATH_MAX bytes, in the buffer pointed to by resolved_path.
So you don't need the +1.
If the Windows version of realpath() from src/lib/libWin/util.c were beefed up, we wouldn't need to preallocate the variable "real" and could just use NULL instead. Later we would have to free the allocated space. I think it would be more platform agnostic, but I'll leave the decision up to you.

minghui-liu: Yes you are right +1 is not necessary. Thanks. I am going with the idea of enhancing the windows function. Although right now the windows version is not used anywhere, once the enhancement's done a few places can be refactored and we can get ride of some ugly `#ifdef`s.

src/tools/pbs_probe.c
```diff
@@ -2244,7 +2244,7 @@ pbsdotconf(char * path)
 static	int
 get_realpath_values(struct infrastruct *pinf)
 {
-	char real[MAXPATHLEN];
+	char real[PATH_MAX];
```
mike0042: Now that you've fixed the realpath function for Windows, this should be changed to:
`
char *real = NULL;
`
The realpath() calls should be passed NULL as the second argument, and the return value assigned to real. Don't forget to check the value after the call, and don't forget to free it when you're done.

minghui-liu: Yes will do in another commit. I am trying to logically separate the changes to multiple commits.

mike0042: Fair enough. Wasn't certain how you intended to proceed.

minghui-liu: Sorry for the confusion. I will remove the WIP tag once I am all done. Then you can review all changes at once.

src/lib/Libwin/util.c
```diff
@@ -242,7 +242,8 @@ fcntl(int fd, int cmd, long arg)
  * 	just a placeholder for the unix equivalent function.
  *
  * @param[in] path - file path
- * @param[in] resolved_path - resolved path
+ * @param[in] resolved_path - resolved path, if NULL then PATH_MAX bytes will
+ *     be allocated, which the caller of this function is responsible to free
```
bhroam: This is an out.  Also, did you mean 'if not NULL'?

src/lib/Libwin/util.c
```diff
@@ -251,6 +252,8 @@ fcntl(int fd, int cmd, long arg)
 char *
 realpath(const char *path, char *resolved_path)
 {
+	if (resolved_path == NULL)
+		resolved_path = (char *)malloc(PATH_MAX);
```
bhroam: No need to cast the return value of malloc().  It returns a void * which is a generic pointer.

src/lib/Libwin/util.c
minghui-liu: No, I mean if equal to NULL. And this parameter is an input. `realpath` will return the allocated memory address as return value.

src/lib/Libwin/util.c
minghui-liu: OK will remove.

src/lib/Libwin/util.c
minghui-liu: Sorry I was wrong. This is an output. The resolved real path will be stored in this address.

src/lib/Liblog/chk_file_sec.c
```diff
@@ -342,6 +342,9 @@ int	fullpath;	/* check full path */
 		}
 	}
 
+	if (real != NULL)
+		free(real);
```
vstumpf: You don't need to nullcheck this.

src/lib/Liblog/chk_file_sec.c
```diff
@@ -374,12 +375,12 @@ tmp_file_sec(char *path, int isdir, int sticky, int disallow, int fullpath)
...
 	int		rc = 0;
 	struct	stat	sbuf;
-	char		real[MAXPATHLEN];
+	char	*real = NULL;
 
 	assert(path != NULL);
 	assert(path[0] != '\0');
 
-	if (realpath(path, real) == NULL) {
+	if (((real = realpath(path, NULL)) == NULL) {
```
mike0042: Parenthesis are unbalanced here.

src/lib/Liblog/chk_file_sec.c
```diff
@@ -472,5 +473,6 @@ tmp_file_sec(char *path, int isdir, int sticky, int disallow, int fullpath)
 		}
 	}
 
+	free(real)
```
mike0042: This comment is for line 459, but I couldn't comment there. Please check the return value of malloc() against NULL and not zero.

mike0042: Need a semicolon at the end of this line.

mike0042:
You should perform a build on your local system before you update your PR. That would have caught this latest error.

minghui-liu:
Yeah. Sorry about all the careless mistakes. I usually push as I commit on my local branch and then test. But since this PR is already open all the maintainers get notified. I will stop pushing new commits until everything is done and tested. Once again I apologize.

src/resmom/stage_func.c
```diff
@@ -175,8 +175,8 @@ is_child_path(char *dir, char *path)
 	strncpy(dir_real, lpath2short(dir), sizeof(dir_real)-1);
 	strncpy(fullpath_real, lpath2short(fullpath), sizeof(fullpath_real)-1);
 #else
-	dir_real = realpath(dir, dir_real);
-	fullpath_real = realpath(fullpath, fullpath_real);
+	dir_real = realpath(dir, NULL);
```
bayucan: I see the realpath() call is done on the linux side but not on Windows. Without the call, we'll crash on Windows. Looking at the updated code, I see:

`158     char *dir_real = NULL;
`
...

```
183     #ifdef WIN32
184              forward2back_slash(fullpath);
185              strncpy(dir_real, lpath2short(dir), sizeof(dir_real)-1);    /*** dir_real is NULL! */
186              strncpy(fullpath_real, lpath2short(fullpath), sizeof(fullpath_real)-1);
187     #else
188                dir_real = realpath(dir, NULL);
189                fullpath_real = realpath(fullpath, NULL);

190    #endif

```

src/tools/pbs_probe.c
```diff
@@ -2257,29 +2257,25 @@ get_realpath_values(struct infrastruct *pinf)
...
 	 * data belonging to *pinf's "pri" member
 	 */
-
 	for (i=0; i<PBS_last; ++i) {
 
 		good_prime[i] = 0;
 		if (pinf->pri.pbs_mpug[i].path) {
-			if (realpath(pinf->pri.pbs_mpug[i].path, &real[0])) {
-				pinf->pri.pbs_mpug[i].realpath = strdup(&real[0]);
+			if ((real = realpath(pinf->pri.pbs_mpug[i].path, NULL)) != NULL) {
```
bhroam: Here you are allocating memory for real in a for() loop without freeing it.  Will it ever allocate twice and leak?  Also, I don't see a free() further down except in the middle of a for() loop.  Is it possible for this function to exit and leak it?

src/tools/pbs_probe.c
```diff
@@ -2413,8 +2410,8 @@ get_realpath_values(struct infrastruct *pinf)
 
 					strcpy(endhead, pmpug[j].path);
 
-					if (realpath(path, &real[0])) {
-						pmpug[j].realpath = strdup(&real[0]);
+					if ((real = realpath(path, NULL)) != NULL) {
```
bhroam: You called realpath() above that allocated memory without freeing it.  Can you allocate more here and leak?

minghui-liu: Fixed. Thanks. Will run tests again and attach test log.

src/resmom/stage_func.c
```diff
@@ -170,13 +170,11 @@ is_child_path(char *dir, char *path)
 	/* even if the file path is relative to some directory, */
 	/* it may use /../../ notation and escape that directory, */
 	/* so always perform full check of parent-child directory relation */
+	dir_real = realpath(dir, NULL);
+	fullpath_real = realpath(fullpath, NULL);
```
bayucan: And ensure also dir_real and fullpath_real are not NULL on Windows before executing strncpy() into it next...

minghui-liu: I changed windows version of `realpath` to allocate memory if second argument is NULL. So `dir_real` and `fullpath_real` should not be null after here.

minghui-liu: Oh OK I see what you mean. `malloc` in `realpath` could still fail.

src/resmom/stage_func.c
```diff
@@ -155,37 +155,40 @@ int
...
+		strncpy(fullpath, path, sizeof(fullpath));
 	}
 
 	/* even if the file path is relative to some directory, */
 	/* it may use /../../ notation and escape that directory, */
 	/* so always perform full check of parent-child directory relation */
+	dir_real = realpath(dir, NULL);
+	fullpath_real = realpath(fullpath, NULL);
 #ifdef WIN32
-	forward2back_slash(fullpath);
```
hirenvadalia: Why did you removed this line? Is this change intentional?

minghui-liu: Yes this change is intentional. Windows version of `realpath` function has a conversion from back slash to forward slash. So both `dir_real` and `fullpath_real` will be in forward slash for the substring search.

src/resmom/stage_func.c
```diff
@@ -167,18 +167,24 @@ is_child_path(char *dir, char *path)
...
+	forward2back_slash(fullpath);
 	strncpy(dir_real, lpath2short(dir), sizeof(dir_real)-1);
 	strncpy(fullpath_real, lpath2short(fullpath), sizeof(fullpath_real)-1);
+#else
+	realpath(dir, dir_real);
+	realpath(fullpath, fullpath_real);
 #endif
+
+	if (dir_real == NULL || fullpath_real == NULL)
+		goto error_exit;
```
mike0042: This check should be done immediately after the calls to malloc().

src/resmom/stage_func.c
```diff
@@ -167,18 +167,24 @@ is_child_path(char *dir, char *path)
...
-	dir_real = realpath(dir, NULL);
-	fullpath_real = realpath(fullpath, NULL);
-
-	if (dir_real == NULL || fullpath_real == NULL)
-		goto error_exit;
 #ifdef WIN32
+	forward2back_slash(fullpath);
 	strncpy(dir_real, lpath2short(dir), sizeof(dir_real)-1);
 	strncpy(fullpath_real, lpath2short(fullpath), sizeof(fullpath_real)-1);
+#else
```
mike0042: The above Windows code looks incorrect because sizeof(dir_real) and sizeof(fullpath_real) will both return the size of the pointer, not the size of the malloc()'ed block.

src/resmom/stage_func.c
```diff
@@ -169,22 +169,21 @@ is_child_path(char *dir, char *path)
 
 	dir_real = malloc(sizeof(char) * (MAXPATHLEN + 1));
 	fullpath_real = malloc(sizeof(char) * (2 * MAXPATHLEN + 2));
+	if (dir_real == NULL || fullpath_real == NULL)
+		goto error_exit;
```
mike0042: There is no way for the caller to determine if something went wrong since return_value will always be TRUE or FALSE. This looks like a case where we want to return...
1 = path is a child
0 = path is not a child
-1 = error encountered

The function is only called in a few places, so it shouldn't be too difficult to update the callers.

minghui-liu:
@PBSPro/pbspro-maintainers I have squashed and rebased.

PR #770, opened 2018-08-08
A qstat job_id.serverA@serverB does not work as specified

Jitendra-Rajak:
#### Bug/feature Description
* For qstat command which require or accept a job identifier operand, if the server is specified in the job identifier operand as @server, then the batch requests will be sent to the server named by server.
* This can be reproduce by having a setup of 2 servers and 1 client installation. Submit the job on one server, move the job to server 2 and from the client run qstat job_id.serverA@serverB. According to correct behavior the request should go to server 2 but it was going server 1. 

#### Affected Platform(s)
* All

* The qstat should send the request to server2 it manages to set current_server to server2 but then it does not return finally the request is going to server1.

#### Solution Description
* Making sure to set current_server to the server1 only if it is NULL.

#### Testing logs/output
* [ptl_result.txt](https://github.com/PBSPro/pbspro/files/2270639/ptl_result.txt)

#### Checklist:


test/tests/pbs_qstat.py
```diff
@@ -0,0 +1,65 @@
...
+# "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
+# trademark licensing policies.
+
+from ptl.utils.pbs_testsuite import *
+
+
+class QstatTest(PBSTestSuite):
+
+    """
+    This test suite is to check which server is handeling request of qstat
```
prakashcv13: this test suite checks whether qstat sends the status request to the correct server or not for requests like job_id.serverA@serverB

test/tests/pbs_qstat.py
```diff
@@ -0,0 +1,65 @@
...
+        self.server2 = self.servers.values()[1]
+        a = {'scheduling': 'false', 'flatuid': 'true'}
+        self.server1.manager(MGR_CMD_SET, SERVER, a)
+        self.server2.manager(MGR_CMD_SET, SERVER, a)
+
+    def test_qstat_req_server(self):
+        j = Job(TEST_USER)
+        jid = self.server1.submit(j)
+        self.server1.expect(JOB, {'job_state': 'Q'}, id=jid)
+        destination = 'workq@%s' % self.server2.hostname
```
prakashcv13: are you sure that workq exists on server2?

test/tests/pbs_qstat.py
```diff
@@ -0,0 +1,65 @@
...
+        self.server1.movejob(jobid=jid, destination=destination)
+        self.server1.status(JOB, id=jid+'@%s' % self.server2.hostname)
+        expmsg = "Type 19 request received" \
+                 " from pbsroot@%s" % self.server.client
```
prakashcv13: will this always be pbsroot?

test/tests/pbs_qstat.py
```diff
@@ -0,0 +1,65 @@
...
+
+class QstatTest(PBSTestSuite):
+
+    """
+    This test suite is to check which server is handeling request of qstat
+    """
+
+    def setUp(self):
+        if len(self.servers.keys()) != 2:
+            self.skipTest("Need two servers")
```
prakashcv13: shall we say "This test needs two servers."

test/tests/pbs_qstat.py
```diff
@@ -0,0 +1,65 @@
...
+    def test_qstat_req_server(self):
+        j = Job(TEST_USER)
+        jid = self.server1.submit(j)
+        self.server1.expect(JOB, {'job_state': 'Q'}, id=jid)
+        destination = 'workq@%s' % self.server2.hostname
+        self.server1.movejob(jobid=jid, destination=destination)
+        self.server1.status(JOB, id=jid+'@%s' % self.server2.hostname)
+        expmsg = "Type 19 request received" \
+                 " from pbsroot@%s" % self.server.client
+        self.server2.log_match(msg=expmsg)
```
prakashcv13: can this test case be part of an existing test suite?

test/tests/pbs_qstat.py
Jitendra-Rajak: I tried with sudo but it failed because it was asking for password from other machines apart from the machine on which the ptl is running and pbsroot has password less ssh.

test/tests/pbs_qstat.py
Jitendra-Rajak: It can also be done directly without mentioning the workq. There are three ways in which we can do qmove.

prakashcv13: let us change it to not use workq.

test/tests/pbs_qstat.py
prakashcv13: it is my understanding that we need not be looking for a particular user in the log message. The only thing to make sure is that the log appears after the stat request is sent.

src/lib/Libcmds/get_server.c
```diff
@@ -171,8 +171,8 @@ get_server(char *job_id_in, char *job_id_out, char *server_out)
 #else
 		strcat(job_id_out, parent_server);
 #endif /* localmod 086 */
-
-		strcpy(server_out, parent_server);
+		if (server_out[0] == '\0')
+			strcpy(server_out, parent_server);
```
prakashcv13: I am of the view that this whole if block can be skipped if we have copied current_server to server_out. The logic to append the server name to the job id is there below. What do you think?

test/tests/pbs_qstat.py
prakashcv13: In that case, let us do it without mentioning the queue name.

test/tests/pbs_qstat.py
prakashcv13: It is my understanding that we need not be looking for the user information in the log message. The only thing that matters is that the log appears after the status request is sent.

src/lib/Libcmds/get_server.c
Jitendra-Rajak: If we provide server2 in the pbs_conf of client on that time this logic will make sense as block of code above where it tries to match parent server and pbs_conf server, that block will not be executed and control will come here and if we skip this block the below block will do strcat between the job_id_out and the second server.

Jitendra-Rajak:
[ptl_result_new.txt](https://github.com/PBSPro/pbspro/files/2274529/ptl_result_new.txt)
Adding the result of ptl after making the changes in the test file.

test/tests/pbs_qstat.py
```diff
@@ -0,0 +1,74 @@
...
+        j = Job(TEST_USER)
+        actualclient = self.server.client
+        self.server.client = self.server1.hostname
+        jid = self.server1.submit(j)
+        self.server1.expect(JOB, {'job_state': 'Q'}, id=jid)
+        destination = '@%s' % self.server2.hostname
+        self.server1.movejob(jobid=jid, destination=destination)
+        self.server.client = actualclient
+        # Allowing 5 secs of time difference in server and client
+        starttime = time.time()-5
```
prakashcv13: could you please explain the reason behind this time difference?

Jitendra-Rajak: When I was writing and running the test  every time I was experiencing 1 or 2 seconds difference in the client and server while doing qstat, so I thought allowing the 5 seconds time difference would be a good Idea.

prakashcv13: If I understand correctly you are saying that there is a time difference of 1-2 seconds between two different machines, which may or may not always happen. It is also possible that 5 seconds may not be enough if the time difference is more than 5 seconds.

test/tests/pbs_qstat.py
```diff
@@ -0,0 +1,74 @@
...
+
+    """
+    This test suite checks whether qstat sends the
+    status request to the correct server or not for
+    requests like job_id.serverA@serverB
+    """
+
+    def setUp(self):
+        if len(self.servers.keys()) != 2:
+            self.skipTest("This needs two servers")
```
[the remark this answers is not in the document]
bhroam: Just a quick suggestion: You can pass dictionaries into len().  Calling self.server.keys() requires python to create a list.  Passing self.servers directly into len(), you can skip that part.  Since self.servers is going to be so short, it really doesn't matter.

test/tests/pbs_qstat.py
Jitendra-Rajak: Yes @prakashcv13, I was seeing a delay if I was not using 5 second of difference because while I was doing qstat it was checking for the logs from the start time and later but in the server it was printing just a second before so I considered 5 seconds time difference. If the difference is more than that the test will fail and one needs to debug in such case ss there any time limit that we can consider for the difference between two machines?

Or we can also check for the logs by using tail, by this we can check the server log for the expected message from the bottom to top by mentioning some range. What you suggest?

prakashcv13: ok. In that case, how will you make sure that the "Type 19" being matched is not for the status request on line 64?

test/tests/pbs_qstat.py
```diff
@@ -0,0 +1,72 @@
...
+class QstatTest(PBSTestSuite):
+
+    """
+    This test suite checks whether qstat sends the
+    status request to the correct server or not for
+    requests like job_id.serverA@serverB
+    """
+
+    def setUp(self):
+        if len(self.servers) != 2:
```
hirenvadalia: I think you need to put check ensuring that given client on pbs_benchpress is 3 machine other than two server.

test/tests/pbs_qstat.py
```diff
@@ -0,0 +1,72 @@
...
+        self.server1 = self.servers.values()[0]
+        self.server2 = self.servers.values()[1]
+        a = {'scheduling': 'false', 'flatuid': 'true'}
+        self.server1.manager(MGR_CMD_SET, SERVER, a)
+        self.server2.manager(MGR_CMD_SET, SERVER, a)
+        self.du.set_pbs_config(self.server.client, confs={'PBS_SERVER': ''})
+
+    def test_qstat_req_server(self):
+        j = Job(TEST_USER)
+        actualclient = self.server.client
```
hirenvadalia: I don't think you need this client toggling. PBSTestSuite.setup() will set proper client to whatever given on pbs_bechpress -p client

test/tests/pbs_qstat.py
hirenvadalia: @prakashcv13 Before movejob Type 19 request will go only in server1's log file because request if going to server1 not server2 so there if not Type 19 request log in server2's log file and once job gets moved then Type 19 request will go in server2's log file not in server1's log file as request if going to server2 not server1.
So at line 64 (which expect) will go to server1 and this status will go to server2.
Hope this what you are looking for.

prakashcv13: Yes, but I am also worried that type 19 log does not include the job id. How can we be sure that the type 19 if for the job we are looking for. Are we assuming that the no other client will be sending any stat request to server2 when the test is running?

Jitendra-Rajak: Yes, as per the test case only one client is there and that will make the request.

prakashcv13: well, it is an assumption that only this test case will be running in the complex. The setup I am referring to is where there are other clients apart from the client used by this test suite.

Jitendra-Rajak: In the log match with type 19 request we are also checking which client has requested.

test/tests/pbs_qstat.py
Jitendra-Rajak: I think this will be required because in the setup I am making the PBS_SERVER empty in pbs_conf of client. So I have to toggle the client so that the job can be submitted and moved.

Jitendra-Rajak:
Hey @prakashcv13  and @hirenvadalia, have made the changes in the ptl please have a look. also adding the latest logs.
[ptl_result_new1.txt.txt](https://github.com/PBSPro/pbspro/files/2314205/ptl_result_new1.txt.txt)

test/tests/pbs_qstat.py
hirenvadalia: I don't think you need to unset PBS_SERVER entry in pbs.conf on client, because if job id has server name in it then qstat (and as far as I know almost all PBS client command) will use that as server to connect rather than PBS_SERVER in pbs.conf.
So even it you keep pbs.conf as it is on client (or set it to server1 so all default request goes to server1) and remove this toggling then also your test should run fine.
(See https://github.com/PBSPro/pbspro/blob/master/src/lib/Libcmds/get_server.c#L142 and https://github.com/PBSPro/pbspro/blob/master/src/cmds/qstat.c#L2628)

Jitendra-Rajak:
Hey @hirenvadalia, thanks for the reference and the explanation I have made the changes accordingly and uploaded the logs as well. Please have a look.
[result.txt.txt](https://github.com/PBSPro/pbspro/files/2314633/result.txt.txt)

test/tests/pbs_qstat.py
prakashcv13: @hirenvadalia - The issue that is being fixed is that qstat is not sending the status request to the proper server. I had explained @Jitendra-Rajak in an internal discussion as to how we can reproduce this particular issue by not having PBS_SERVER=server1 in client's pbs.conf.
The test case is passing with the latest change as PBS_SERVER is set to server1 in pbs.conf on the client machine and the actual issue itself is not getting reproduced. With the last change, I think the test case will pass on builds without fix as well, @Jitendra-Rajak - can confirm.
This is because the comparison on line 153 (https://github.com/PBSPro/pbspro/blob/b8eb6ffb5eb68009aee1c87837f5325022326574/src/lib/Libcmds/get_server.c#L153) in get_server.c will fail and get_server() will not return on line 157.
It is my opinion that we should reverse the last change.

test/tests/pbs_qstat.py
prakashcv13: There is a possibility that a status request is sent from outside the test case on the same client for another job id. The test case will still pass.
I do agree that such a thing will not happen in the ideal world.

test/tests/pbs_qstat.py
Jitendra-Rajak: Yes @hirenvadalia, I was not able to reproduce the issue with having PBS_SERVER in client's pbs.conf. So after that @prakashcv13 and I were able to reproduce the issue by making PBS_SERVER empty in client's pbs.conf. 
That is why in the setup I was making PBS_SERVER empty so that the test follows the actual scenario in which the bug was logged. I think reversing the change should be the perfect solution.

Jitendra-Rajak:
Hey @prakashcv13, I have made the changes in the ptl test case please have a look. 
[result_after.txt](https://github.com/PBSPro/pbspro/files/2316982/result_after.txt)

bhroam:
@Jitendra-Rajak thanks for moving the file.  I have one last comment which I really want @hirenvadalia to weigh in on.  You are modifying self.server.client.  To me that is a private member of the Server class.  Should a test be modifying it?  There is no setter for it, so the only way to modify it is either to directly set it like you are doing or create a new instance of the Server class which seems wasteful.  If @hirenvadalia doesn't have a problem with it, then I'll sign off.

test/tests/functional/pbs_qstat.py
```diff
@@ -0,0 +1,73 @@
...
+        self.du.set_pbs_config(self.server.client, confs={'PBS_SERVER': ''})
+
+    def test_qstat_req_server(self):
+        j = Job(TEST_USER)
+        actualclient = self.server.client
+        self.server.client = self.server1.hostname
+        jid = self.server1.submit(j)
+        self.server1.expect(JOB, {'job_state': 'Q'}, id=jid)
+        destination = '@%s' % self.server2.hostname
+        self.server1.movejob(jobid=jid, destination=destination)
```
bhroam: Actually I do have one more comment.  If the movejob() fails, it will raise an exception.  This will leave self.server.client set to the alternate value.

test/tests/functional/pbs_qstat.py
```diff
@@ -55,19 +55,16 @@ def setUp(self):
...
     def test_qstat_req_server(self):
         j = Job(TEST_USER)
-        actualclient = self.server.client
-        self.server.client = self.server1.hostname
         jid = self.server1.submit(j)
-        self.server1.expect(JOB, {'job_state': 'Q'}, id=jid)
         destination = '@%s' % self.server2.hostname
         self.server1.movejob(jobid=jid, destination=destination)
-        self.server.client = actualclient
+        self.du.set_pbs_config(self.server.client, confs={'PBS_SERVER': ''})
```
prakashcv13: I believe you moved this after movejob() based on @bhroam's comment. Shouldn't we set PBS_SERVER to the previous value in teardown()?

Jitendra-Rajak: Hey @prakashcv13, I believe when ptl starts a new test case it will revert pbs configuration back to out of box. Do you still think we need to take care ?

test/tests/functional/pbs_qstat.py
```diff
@@ -68,3 +68,8 @@ def test_qstat_req_server(self):
                  " from %s@%s" % (str(TEST_USER), self.server.client)
         self.server1.log_match(msg=expmsg, existence=False, max_attempts=5)
         self.server2.log_match(msg=expmsg)
+
+    def tearDown(self):
+        a = {'PBS_SERVER': '@%s' % self.server1.hostname}
```
prakashcv13: this will set PBS_SERVER to server1.hostname regardless of the previous value.
You need to save the previous value and use it here. Also, attach the test logs.

Jitendra-Rajak:
Hey @prakashcv13, made the changes and attached logs as well. Please have a look.
[before.txt](https://github.com/PBSPro/pbspro/files/2339381/before.txt)
[after.txt](https://github.com/PBSPro/pbspro/files/2339382/after.txt)

Jitendra-Rajak:
Thanks @hirenvadalia and @prakashcv13 for reviews and approving the PR.
@PBSPro/pbspro-maintainers need one more review from one of you.

Jitendra-Rajak:
Squashed and re-based, Please take the PR forward.

subhasisb:
@Jitendra-Rajak your commits need to be signed.

PR #771, opened 2018-08-09
Provide a pkg-config file libpbs.pro

e4t:
This should help other project to build against this library.

Signed-off-by: Egbert Eich <eich@suse.com>

#### Bug/feature Description
Pkg-config file can be used by autotooled projects, projects using cmake and other build systems to obtain information about availability and required flags for building. It nicely provides linker flags of dependent libraries for static builds, adds a lot of flexibility and removes a need for guess work finding the right file locations.
This is provided here for libpbspro as well. 

#### Affected Platform(s)
* *All*

* [This page describes the use of pkg-config](https://people.freedesktop.org/~dbn/pkg-config-guide.html)

#### Solution Description
* Add pkg-config template, made adjustments to Makefile.am-s to generate and install the pkg-config file.

#### Testing logs/output
* File is available at default location after installation.

#### Checklist:

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -291,6 +291,9 @@ AC_CHECK_FUNCS([ \
 	utime \
 ])
 
+PKG_PROG_PKG_CONFIG
+PKG_INSTALLDIR
+
 # PBS macros (order matters for some of these)
 PBS_AC_PBS_VERSION
 PBS_AC_MACHINE_TYPE
@@ -365,6 +368,7 @@ AC_CONFIG_FILES([
 	src/lib/Liblog/Makefile
 	src/lib/Libnet/Makefile
 	src/lib/Libpbs/Makefile
+	src/lib/Libpbs/pbs.pc
 	src/lib/Libpython/Makefile
 	src/lib/Libsec/Makefile
 	src/lib/Libsite/Makefile
--- a/src/lib/Libpbs/Makefile.am
+++ b/src/lib/Libpbs/Makefile.am
@@ -46,6 +46,8 @@ libpbs_la_CPPFLAGS = -I$(top_srcdir)/src/include
 # to the following article prior to modifying the library version:
 # https://autotools.io/libtool/version.html
 #
+pkgconfig_DATA = pbs.pc
+
 libpbs_la_LDFLAGS = -version-info 0:0:0
 
 libpbs_la_LIBADD= \
--- a/src/lib/Libpbs/pbs.pc.in
+++ b/src/lib/Libpbs/pbs.pc.in
@@ -1,0 +1,11 @@
+prefix=@prefix@
+exec_prefix=@exec_prefix@
+libdir=@libdir@
+includedir=@includedir@
+
+Name: @PACKAGE_NAME@
+Description: Library for the PBSPro Workload Manager
+Version: @PACKAGE_VERSION@
+Libs: -L${libdir} -lpbs
+Cflags: -I${includedir}
+Requires.private: libcrypto


mike0042:
Closing and reopening to restart Appveyor build.

mike0042:
Reopening to rerun CI checks.

subhasisb:
Since this is a pkg-config for the libpbs.so library, should the name of the file be libpbs.pc? Instead of pbspro.pc?

e4t:
@subhasisb This is actually a good question: if there are plans to add additional libraries to PBS Pro - which are to provide external ABIs - it would indeed be better to use the library name - this way, it would be easy to have a .pc file for each library. It looks like there are different conventions used in the real world: either `libfoo.pc` or just `foo.pc` - corresponding to the linker argument `-lfoo` for  a library `libfoo`. I'd personally opt for `pbs.pc`. But `libpbs.pc` would be fine with me as well.

Let me know what you guys think and I will change it accordingly.

subhasisb:
yes @e4t it would be better just in case we have more shared libraries in the future. I am fine with pbs.pc as well - just felt that if it matched the actual library name it would cleaner. Matching with the actual linker argument sounds great! Thanks.

e4t:
@subhasisb I'm fine either way - so we may need more votes ;)

No, pkg-config can help with static libraries as well: while I was looking into packaging libraries into separate binary packages for SUSE, I stumbled over a number of shared libraries that got installed:
`libattr.a  liblog.a  libnet.a  libpbs.a  libpbs_sched.a  libsite.a`.
`libpbs.a` corresponds to `libpbs.so.*`, however, a quick comparison of some of the symbols provided by the other ones with the public header files seems to indicate that those do not provide a public API. Also, they don't seem to be needed to link against the libpbs.a static library (I may be wrong as I haven't checked systematically).
If the libraries `libattr.a  liblog.a  libnet.a libpbs_sched.a  libsite.a` are only internal, they should probably not be installed - however, if they are needed to link against libpbs.a, this should be set in the pkg-config file.
Please let me know, I'll make a patch to not install them or fix up the pkg-config file.

mike0042:
There are many places where "pbs" and "pbspro" are used interchangeably, and we've never had much reason for concern since we were packaging as an ISV application. The best short term solution (IMHO) would be to install the PBS Pro libraries in %{_libdir}/%{name} in cases where we are packaging for the distro itself. A better solution would be to refactor the library names (and possibly content) such that the names would become:
libpbs.so
libpbs_attr.a
libpbs_log.a
libpbs_net.a
libpbs_site.a
etc.

If we were to take that approach we could install directly into %{_libdir} since there should no longer be any name conflicts. I haven't tried, but I believe this would also avoid breaking builds for OpenMPI and other packages that rely on PBS Pro libraries.

e4t:
@mike0042 It is not about name spacing - I was wondering if these libraries need to be installed at all, ie. whether they provide symbols for consumption by non-PBSPro software.
I went ahead and checked the text symbols in these libraries against the header files and `libpbs.a`.
Here's the result: only the following symbols

- `activereq`

- `addreq`

- `allreq`

- `closerm`

- `configrm`

- `downrm`

- `flushreq`

- `fullresp`

- `getreq`

- `openrm`
are represented in the *installed* header files. All of which are in `rm.c`. These are all from in rm.c in Libnet, and are thus part of `libnet.a`. They are all represented in the header `rm.h`. AFAIKT, rm.c does not contain any undefined symbols which are resolved by any of the other libraries.
If `libnet.a` provides a public API, it brings up the question why there is no shared version of it.
In fact, why isn't `rm.c` not just part of `libpbs`?
But this is getting off-topic here. Maybe I should take this discussion to the developers forum.

mike0042:
@e4t What you are pointing out is very much inline with the refactoring suggested. The only library that had to be a shared object when we publish PBS Pro as open source was libpbs, so the others got ignored in the interest of time. We should not install any libraries that are needed by PBS Pro internally, and only install those that external packages might link against. And I tend to agree, all external libraries should be distributed as shared objects.

e4t:
@mike0042 If you can check if the symbols declared in `src/include/rm.h` are meant to be part of the public API, I would make the changes - depending on the results either drop `rm.h` from the installed headers or move `rm.c` to the files included in `libpbs.*` - either way would be really easy.

e4t:
@mike0042 Just for the record - the static library topic is now discussed in #782 and #783.
I went ahead and pushed an update to change the pkgconfig file to `pbs.pc`: I've done a quick check across all pkgconfig files shipped with openSUSE Tumbleweed: about 3 out of 4 don't use the prefix `lib` - although not all of them deal with libraries, there seems to be a majority which do not put the `lib`. I'm fine either way, I also have a patch adding the `lib`-prefix.

subhasisb:
@mike0042 do you have any further feedback?

subhasisb:
Hi @e4t please rebase your commit in preparation for merge

mike0042:
@e4t: Please rebase this pull request in preparation for merge.

e4t:
@subhasisb, @mike0042 - I've rebased the patch and re-pushed it. Sorry for the delay. 
Thank you!

PR #780, opened 2018-08-15
PP-1026: Stranded array subjobs

Shrini-h:
#### Bug/feature Description
* *Sometimes instantiated subjobs can get stranded in queued state due to communication failure with execution node*
* *Not easily reproducible, usually observed during communication hiccups in large clusters using large job arrays*
* *Refer issue ID: [PP-1026](https://pbspro.atlassian.net/browse/PP-1026)*

#### Affected Platform(s)
* *All*

#### Cause
* *Instantiated subjobs in Q state. the subjobs were instantiated when the server first tried to run them, the find_jobs() in req_runjob() finds an entry for them, causing the server to reject them.*

#### Solution Description
* *Purge an instantiated Queued subjob and re-instantiate it afresh during runjob*

#### Testing logs/output
* *PTL test cannot be automated cause there is no way to create a stranded subjob through commands*
* *Hence I devised below manual steps to verify the fix*

1. Manually build the server by commenting out below lines (for force creating an instantiated Q subjob)
https://github.com/PBSPro/pbspro/blob/882f33a732d82736aca098d8ee0a7e493287628c/src/server/req_runjob.c#L677-L681
2. Comment out below line
https://github.com/PBSPro/pbspro/blob/882f33a732d82736aca098d8ee0a7e493287628c/test/tests/functional/pbs_sched_subjob_badstate.py#L77
3. Run PTL testcase
`pbs_benchpress -t TestSchedSubjobBadstate`
4. After the above test completes we can see
 from `/etc/init.d/pbs status` that `pbs_mom` is not running but `pbsnodes -av` shows the node as free
 from `qstat -t`  that all the subjob are in Q state,
5. Manually Q sub some more regular jobs and array jobs. all will get queued. The first subjob of all the array are stranded subjobs.
6. now start the pbs_mom.
7. We should see that the stranded Q subjobs from the PTL test case and step 5 goes to Running
8. Verify the server logs. We should see as seen in the sample attached below
[strandedsubjob_logs.zip](https://github.com/PBSPro/pbspro/files/2315311/strandedsubjob_logs.zip)
PTL test output attached below
[ptl_test_results.zip](https://github.com/PBSPro/pbspro/files/2317518/ptl_test_results.zip)

#### Checklist:

### DIFF
--- a/src/server/req_runjob.c
+++ b/src/server/req_runjob.c
@@ -376,7 +376,7 @@ req_runjob(struct batch_request *preq)
 		if (i == -1) {
 			req_reject(PBSE_IVALREQ, 0, preq);
 			return;
-		} else if ((i != JOB_STATE_QUEUED) || (find_job(jid) != NULL)) {
+		} else if (i != JOB_STATE_QUEUED) {
 			/* job already running */
 			req_reject(PBSE_BADSTATE, 0, preq);
 			return;
@@ -505,20 +505,9 @@ req_runjob(struct batch_request *preq)
 
 		/* single subjob, if parent qeueud, it can be run */
 
-		offset = subjob_index_to_offset(parent, get_index_from_jid(jid));
-		if (offset == -1) {
-			req_reject(PBSE_UNKJOBID, 0, preq);
-			return;
-		}
-		i = get_subjob_state(parent, offset);
-		if (i == -1) {
-			req_reject(PBSE_IVALREQ, 0, preq);
-			return;
-		} else if ((i != JOB_STATE_QUEUED) || (find_job(jid) != NULL)) {
-			/* job already running */
-			req_reject(PBSE_BADSTATE, 0, preq);
-			return;
-		}
+		if ((pjobsub = find_job(jid)) != NULL)
+			job_purge(pjobsub);
+
 		if ((pjobsub = create_subjob(parent, jid, &j)) == NULL) {
 			req_reject(j, 0, preq);
 			return;
@@ -572,27 +561,27 @@ req_runjob(struct batch_request *preq)
 
 			if (get_subjob_state(parent, i) == JOB_STATE_QUEUED) {
 				jid  = mk_subjob_id(parent, i);
-				pjob = find_job(jid);
-				if (pjob == NULL) {
-					if ((pjobsub = create_subjob(parent, jid, &j)) == NULL) {
-						req_reject(j, 0, preq);
-						continue;
-					}
-					if (call_to_process_hooks(preq, hook_msg, sizeof(hook_msg),
-							pbs_python_set_interrupt) == 0) {
-						/* subjob reject from hook*/
-						job_purge(pjobsub);
-						reply_text(preq, PBSE_HOOKERROR, hook_msg);
-						return;
-					}
-					if ((pjob = where_to_runjob(preq, pjobsub)) == NULL) {
-						/* requeue subjob */
-						pjobsub->ji_qs.ji_substate = JOB_SUBSTATE_RERUN3;
-						job_purge(pjobsub);
-						continue;
-					}
-					dup_br_for_subjob(preq, pjob, req_runjob2);
+				if ((pjobsub = find_job(jid)) != NULL)
+					job_purge(pjobsub);
+
+				if ((pjobsub = create_subjob(parent, jid, &j)) == NULL) {
+					req_reject(j, 0, preq);
+					continue;
 				}
+				if (call_to_process_hooks(preq, hook_msg, sizeof(hook_msg),
+						pbs_python_set_interrupt) == 0) {
+					/* subjob reject from hook*/
+					job_purge(pjobsub);
+					reply_text(preq, PBSE_HOOKERROR, hook_msg);
+					return;
+				}
+				if ((pjob = where_to_runjob(preq, pjobsub)) == NULL) {
+					/* requeue subjob */
+					pjobsub->ji_qs.ji_substate = JOB_SUBSTATE_RERUN3;
+					job_purge(pjobsub);
+					continue;
+				}
+				dup_br_for_subjob(preq, pjob, req_runjob2);
 			}
 		}
 		range = pc;
@@ -1084,7 +1073,8 @@ svr_strtjob2(job *pjob, struct batch_request *preq)
 			pjob->ji_qs.ji_jobid,
 			"Unable to Run Job, send to Mom failed");
 
-		if (pjob->ji_qs.ji_substate == JOB_SUBSTATE_PROVISION)
+		if (pjob->ji_qs.ji_substate == JOB_SUBSTATE_PROVISION ||
+				pjob->ji_qs.ji_substate == JOB_SUBSTATE_PRERUN)
 			rel_resc(pjob);
 		else
 			free_nodes(pjob);
@@ -1500,9 +1490,6 @@ post_sendmom(struct work_task *pwt)
 			break;
 #endif
 		case SEND_JOB_NODEDW:	/* node (mother superior) is down? */
-			if (jobp->ji_qs.ji_substate == JOB_SUBSTATE_PROVISION ||
-					jobp->ji_qs.ji_substate == JOB_SUBSTATE_PRERUN)
-				rel_resc(jobp);
 			mark_node_down(jobp->ji_qs.ji_destin, "could not send job to mom");
 
 			/* fall through to requeue job */
--- a/src/server/svr_movejob.c
+++ b/src/server/svr_movejob.c
@@ -530,14 +530,17 @@ send_job_exec(job *jobp, pbs_net_t hostaddr, int port, struct batch_request *req
 	stream = svr_connect(hostaddr, port, NULL, ToServerDIS, rpp);
 	if (stream < 0) {
 		log_event(PBSEVENT_ERROR, PBS_EVENTCLASS_REQUEST, LOG_WARNING, "", "Could not connect to Mom");
+		pbs_errno = PBSE_NORELYMOM;
 		goto send_err;
 	}
 
 	pmom = tfind2((unsigned long) jobp->ji_qs.ji_un.ji_exect.ji_momaddr,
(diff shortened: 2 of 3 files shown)


src/scheduler/job_info.c
```diff
@@ -4248,7 +4248,7 @@ queue_subjob(resource_resv *array, server_info *sinfo,
 	if (subjob_index >= 0) {
 		subjob_name = create_subjob_name(array->name, subjob_index);
 		if (subjob_name != NULL) {
-			if ((rresv = find_resource_resv(sinfo->jobs, subjob_name)) != NULL) {
+			if ((rresv = find_resource_resv(qinfo->jobs, subjob_name)) != NULL) {
```
bhroam: Just a quick comment: While this might look right, it is not.  The point of this check is to see if the job is in the system.  You want to search the entire system, not just a subset.

Shrini-h: Was just testing the water. I thought of speeding this up since array jobs do not span across queues from server's point of view. But now after looking at movejob server code I realize there are possibilities of subjob instances internally spanning across queues (moving partially completed array job to a different queue)

However I feel that there is an opportunity to speed things up here, as we are scanning the whole system job pool (O(n)) before creation of each new subjob.
Anyways since this is not needed for the problem at hand, I will revert this change and leave it to the sched owner's discretion on the speed up opportunity.

bhroam: Thanks for thinking of scheduler speed.  This is something that is always in the back of our minds.

src/server/req_runjob.c
```diff
@@ -1084,7 +1054,10 @@ svr_strtjob2(job *pjob, struct batch_request *preq)
 			pjob->ji_qs.ji_jobid,
 			"Unable to Run Job, send to Mom failed");
 
-		if (pjob->ji_qs.ji_substate == JOB_SUBSTATE_PROVISION)
+		mark_node_down(pjob->ji_qs.ji_destin, "could not send job to mom");
```
arungrover: can send_job not fail due to multiple reasons? How will you know that under what kind of failure you want to mark the node down?

src/server/req_jobobit.c
```diff
@@ -1481,14 +1481,6 @@ on_job_rerun(struct work_task *ptask)
 				unset_extra_attributes(pjob);
 
 
-				/* Now  if not a Sub Job, then re-queue the job */
-
-				if (pjob->ji_qs.ji_svrflags & JOB_SVFLG_SubJob) {
-					/* for a sub job, just purge it */
-					/* note: substate already JOB_SUBSTATE_RERUN3 */
-					job_purge(pjob);
```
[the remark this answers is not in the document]
Shrini-h: Its updated in below flow
After twenty lines from here we have svr_setjobstate() and from there
```
svr_setjobstate()
       update_subjob_state()
              chk_array_doneness()
```

src/server/req_runjob.c
Shrini-h: Nice catch, fixed in commit with tag "rc5"

src/scheduler/job_info.c
Shrini-h: This change now stands reverted.

Shrini-h:
@arungrover  Thanks for your inputs

> Up until today only running sub jobs were reported to the scheduler. This change will make queued sub jobs to be reported to the scheduler as well. I understand your intent to speed things up and optimize code but maybe it needs a little more thinking.

I agree this needs more thinking. I feel as per current scheduler code, it should be able to handle queued subjobs.

> With this change in place now there is a chance that the queued jobs will now end up in HELD state (because of run_count) which was never the case before (in fact qhold rejects such requests on sub jobs).

As discussed by community for [PP-479](http://community.pbspro.org/t/pp-479-running-subjobs-to-be-able-to-survive-a-pbs-server-restart/835), we are moving towards making subjob equivalent to a regular job, I feel we should allow the subjob to be Held from both qhold and run_count limiting. We already have old open issue [PP-72](https://pbspro.atlassian.net/browse/PP-72). 

May be I should reset the run_count and run_version in req_runjob() for now.

I'm still dual minded if I should stretch this PR to include these changes or should I take the easy route of purging an instantiated queued subjob during runjob

> Because of this change, it might happen that queued sub jobs will start accruing eligible-time, this was also not the case before.

Addressed this in latest commit.

> I think all these things are leading to more of a design change on how queued sub jobs were earlier perceived.

Agree. But I feel we should consider this redesign asap since [PP-479](https://pbspro.atlassian.net/browse/PP-479) already laid the ground work for our goal for subjob robustness

Shrini-h:
@bhroam 
>making all subjobs real jobs is not just an internal design change. There will be external changes. 

Yes true, my changes did not do so.
>I think the latter is a mistake.

Yes, but the fact that we are seeing this bug means the system is encountering this mistake in real world.

Hence I was attempting to make the system handle when it encounter's an instantiated Queued subjob to be run.

Anyways I have decided to move the current approach to back up branch https://github.com/Shrini-h/pbspro/tree/zbackup_dev_PP-1026_approach_1 
we should create a RFE and forum to discuss with community on if and how we should improve the array job.

I will soon work and push a new approach in this PR.

Shrini-h:
New fresh easy fix pushed. Please review

src/server/req_runjob.c
```diff
@@ -1084,7 +1073,11 @@ svr_strtjob2(job *pjob, struct batch_request *preq)
 			pjob->ji_qs.ji_jobid,
 			"Unable to Run Job, send to Mom failed");
 
-		if (pjob->ji_qs.ji_substate == JOB_SUBSTATE_PROVISION)
+		if(pbs_errno == PBSE_NORELYMOM)
+			mark_node_down(pjob->ji_qs.ji_destin, "could not send job to mom");
```
arungrover: What happens when the server fails to connect to mom while staging in the files? It seems that svr_strtjob2 may not even get called in that case.

src/server/svr_movejob.c
```diff
@@ -530,14 +530,17 @@ send_job_exec(job *jobp, pbs_net_t hostaddr, int port, struct batch_request *req
...
+		pbs_errno = PBSE_NORELYMOM;
 		goto send_err;
 	}
 
 	pmom = tfind2((unsigned long) jobp->ji_qs.ji_un.ji_exect.ji_momaddr,
 		jobp->ji_qs.ji_un.ji_exect.ji_momport,
 		&ipaddrs);
-	if (!pmom || (((mom_svrinfo_t *)(pmom->mi_data))->msr_state & INUSE_DOWN))
+	if (!pmom || (((mom_svrinfo_t *)(pmom->mi_data))->msr_state & INUSE_DOWN)) {
+		pbs_errno = PBSE_NORELYMOM;
```
arungrover: I'm not sure why you set this error number here. If the mom is already not there then why would you want to set this error number and mark the node down in runjob. Also, if the node is already marked down then it does not really help to mark it down again. I could not find a better error number for this case, maybe create a new error number like PBSE_NODEDOWN and purge the subjob upon receiving this error number as well.

Shrini-h: As seen in the bug description and logs, the error number displayed is "0" which does not provide correct information on the type of failure. hence setting appropriate error number here
```
06/01/2016 17:38:20;0008;Server@pbs-1;Job;42399[1188].pbs-1;Job Run at 
request of Scheduler@pbs-1 on exec_vnode (r311i0n3:ncpus=16) 
06/01/2016 17:38:20;0001;Server@pbs-1;Req;;Could not connect to Mom 
06/01/2016 17:38:20;0008;Server@pbs-1;Job;42399[1188].pbs-1;send of job to 
r311i0n3 failed error = 0 
06/01/2016 17:38:20;0008;Server@pbs-1;Job;42399[1188].pbs-1;Unable to 
Run Job, send to Mom failed 
```

Now after looking at the code of svr_connect() I think You are right that I might be marking the node down again.

I will remove the added call to `mark_node_down()` in `svr_strtjob2()`, but keep the updates to `pbs_errno` for better reporting in the logs.

src/server/req_runjob.c
Shrini-h: I was actually following the failure cases in of `post_sendmom()` here. Now after looking at the code of `svr_connect()` I think You are right that I might be marking the node down again.  I will remove this `mark_node_down()` call in the next push.

src/server/req_runjob.c
```diff
@@ -1084,7 +1073,11 @@ svr_strtjob2(job *pjob, struct batch_request *preq)
 			pjob->ji_qs.ji_jobid,
 			"Unable to Run Job, send to Mom failed");
 
-		if (pjob->ji_qs.ji_substate == JOB_SUBSTATE_PROVISION)
+		if(pbs_errno == PBSE_NORELYMOM)
```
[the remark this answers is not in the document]
Shrini-h: This line has been removed in new push rc2 : https://github.com/PBSPro/pbspro/pull/780/commits/2f8f2d15aef0e97bea4c75f72249339d3986f6a0

src/server/svr_movejob.c
Shrini-h: >and purge the subjob upon receiving this error number as well.

Now that `pbs_error` is appropriately set to non-zero, the subjob will get purged down the line at below location in `req_runjob2()`
https://github.com/PBSPro/pbspro/blob/882f33a732d82736aca098d8ee0a7e493287628c/src/server/req_runjob.c#L674-L684

Shrini-h:
PR ready for Sign-Off or Review

Shrini-h:
new Test logs after branch update are attached below
[new_ptl_test_results.zip](https://github.com/PBSPro/pbspro/files/2326861/new_ptl_test_results.zip)
[new_manual_test_logs.zip](https://github.com/PBSPro/pbspro/files/2326862/new_manual_test_logs.zip)

Shrini-h:
Ready for Squash?
Codacy seems to be broken due to external factors.

Shrini-h:
PR ready for merge!!
Squashed the commits!!
pre-squash history saved at https://github.com/Shrini-h/pbspro/tree/zbackup_PP-1026

PR #781, opened 2018-08-16
default_qsub_arguments -V not working

sakshamgarg:
#### Bug/feature Description
* *qsub -V option is not working when it is set using default_qsub_arguments.*
* *Bugs: Once the default_qsub_arguments is set for -V option, "qstat -f" doesn't display environment variable except those which are related to PBS or which are set using -v option.*

#### Affected Platform(s)
* *ALL*

* *Bugs: Environment variable list is not stored after checking that -V option is set or not in default_qsub_arguments.*

#### Solution Description
* *After the options for default_qsub_arguments are processed, -V option is checked whether it is set or not.*
* *If the -V option is set then using the variable 'environ', all the environment variable are stored as a list in a variable.*

#### Testing logs/output
* [logs_default_qsub_arg.txt](https://github.com/PBSPro/pbspro/files/2293481/logs_default_qsub_arg.txt)

#### Checklist:

### DIFF
--- a/src/cmds/qsub.c
+++ b/src/cmds/qsub.c
@@ -160,6 +160,7 @@
 #define QSUB_DMN_TIMEOUT_LONG 60 /* timeout for qsub background process */
 #define QSUB_DMN_TIMEOUT_SHORT 5
 
+#define DMN_REFUSE_EXIT 7 /* return code when daemon can't serve a job and exits */
 
 #ifdef NAS /* localmod 005 */
 extern void set_attr_resc(struct attrl **attrib, char *attrib_name, char *attrib_resc, char *attrib_value);
@@ -231,6 +232,11 @@ static char *pbs_hostvar = NULL; /* buffer containing ",PBS_O_HOST=" and host na
 static int pbs_o_hostsize = sizeof(",PBS_O_HOST=") + 1; /* size of prefix for hostvar */
 static char *display; /* environment variable DISPLAY */
 
+/*
+ * Flag to check if current process is the background process.
+ * This variable is set only once and is read-only afterwards.
+ */
+static int is_background = 0;
 static int no_background = 0; /* flag to disable backgrounding */
 static char roptarg = 'y'; /* whether the job is rerunnable */
 static char *v_value = NULL; /* expanded variable list from v opt */
@@ -4904,7 +4910,8 @@ do_connect(char *server_out, char *retmsg)
  * @return int
  * @retval 0 - Success
  * @retval 1/-1/pbs_errno - Failure, retmsg paramter is set
- *
+ * @retval DMN_REFUSE_EXIT - If daemon can't submit the job
+ * 
  */
 static int
 do_submit(char *retmsg)
@@ -4915,7 +4922,7 @@ do_submit(char *retmsg)
 	char *errmsg;
 	int retries;
 
-	if (dfltqsubargs != NULL) {
+    if (dfltqsubargs != NULL) {
 		/*
 		 * Setting options from the server defaults will not overwrite
 		 * options set from the job script. CMDLINE-2 means
@@ -4930,6 +4937,16 @@ do_submit(char *retmsg)
 		}
 		if (rc != 0)
 			return (rc);
+	}
+
+	/* 
+	 * get environment variable if -V option is set. Return the code
+	 * DMN_REFUSE_EXIT if -V option is detected in background qsub.
+	 */
+	if (V_opt) {
+		if (is_background)
+			return DMN_REFUSE_EXIT;
+		qsub_envlist = env_array_to_varlist(environ);
 	}
 
 	/* set_job_env must be done here to pick up -v, -V options passed by default_qsub_arguments */
@@ -5898,6 +5915,10 @@ do_daemon_stuff(void)
 			free(cred_buf);
 			cred_buf = NULL;
 		}
+
+		/* Exit the daemon if it can't submit the job */
+		if (rc == DMN_REFUSE_EXIT)
+			goto out;
 #if defined(PBS_PASS_CREDENTIALS)
 		memset(passwd_buf, 0, PBS_MAXPWLEN);
 #endif
@@ -5966,6 +5987,8 @@ fork_and_stay(void)
 		/* set single threaded mode */
 		pbs_client_thread_set_single_threaded_mode();
 
+		/* set when background qsub is running */
+		is_background = 1; 
 		do_daemon_stuff();
 		/*
 		 * Control should never reach here.
@@ -6064,16 +6087,21 @@ again:
 
 			/* read back response from background daemon */
 			if ((recv_string(&sock, retmsg) != 0) ||
-				dorecv(&sock, (char *) &rc, sizeof(int)) != 0) {
+				(dorecv(&sock, (char *) &rc, sizeof(int)) != 0) ||
+				rc == DMN_REFUSE_EXIT) {
 
 				/*
 				 * Something bad happened, either background submitted
 				 * and failed to send us response, or it failed before
-				 * submitting.
+				 * submitting. If background qsub detects -V option, then
+				 * submit the job through foreground.
 				 */
-				rc = -1;
-				sprintf(retmsg, "Failed to recv data from background qsub\n");
-				/* Error message will be printed in caller */
+				if (rc != DMN_REFUSE_EXIT) {
+					rc = -1;
+					sprintf(retmsg, "Failed to recv data from background qsub\n");
+					/* Error message will be printed in caller */
+				} else
+					*do_regular_submit = 1;
 			}
 		}
 		/* going down, no need to free stuff */
@@ -6107,7 +6135,7 @@ regular_submit(int daemon_up)
 			rc = -1;
 	}
 #ifndef WIN32
-	if ((rc == 0) && !(Interact_opt != FALSE || block_opt) && (daemon_up == 0) && (no_background == 0))
+	if ((rc == 0) && !(Interact_opt != FALSE || block_opt) && (daemon_up == 0) && (no_background == 0) && !V_opt)
 		fork_and_stay();
 #endif
 	return rc;
--- a/test/tests/functional/pbs_passing_environment_variable.py
+++ b/test/tests/functional/pbs_passing_environment_variable.py
@@ -105,3 +105,53 @@ class Test_passing_environment_variable_via_qsub(TestFunctional):
         self.assertEqual(job_output, match,
                          msg="Environment variable foo content does "
                          "not match original")
+
+    def test_option_V_dfltqsubargs(self):
+        """
+        Test exporting environment variable when -V is enabled
+        in default_qsub_arguments.
+        """
+        os.environ["SET_IN_SUBMISSION"] = "true"
+        self.server.manager(MGR_CMD_SET, SERVER,
+                            {'default_qsub_arguments': '-V'})
+        j = Job(self.du.get_current_user())
+        jid = self.server.submit(j)
+        self.server.expect(JOB, {'Variable_List': (MATCH_RE,
+                           'SET_IN_SUBMISSION=true')},
+                           id=jid)
+
+    def test_option_V_cmdline(self):
+        """
+        Test exporting environment variable when -V is passed
+        through command line.
+        """
+        os.environ["SET_IN_SUBMISSION"] = "true"
+        self.ATTR_V = 'Full_Variable_List'
+        api_to_cli.setdefault(self.ATTR_V, 'V')
+        a = {self.ATTR_V: None}
+        j = Job(self.du.get_current_user(), attrs=a)
+        jid = self.server.submit(j)
+        self.server.expect(JOB, {'Variable_List': (MATCH_RE,
+                           'SET_IN_SUBMISSION=true')},
+                           id=jid)
+
+    def test_option_V_dfltqsubargs_qsub_daemon(self):
+        """
+        Test whether the changed value of the exported
+        environment variable is reflected if the submitted job
+        goes to qsub daemon.
+        """
+        self.server.manager(MGR_CMD_SET, SERVER,
+                            {'default_qsub_arguments': '-V'})
+        os.environ["SET_IN_SUBMISSION"] = "true"
+        j = Job(self.du.get_current_user())
+        jid = self.server.submit(j)
+        os.environ["SET_IN_SUBMISSION"] = "false"
+        j1 = Job(self.du.get_current_user())
+        jid1 = self.server.submit(j1)
+        self.server.expect(JOB, {'Variable_List': (MATCH_RE,
+                           'SET_IN_SUBMISSION=true')},
+                           id=jid)
+        self.server.expect(JOB, {'Variable_List': (MATCH_RE,
+                           'SET_IN_SUBMISSION=false')},
+                           id=jid1)


src/cmds/qsub.c
```diff
@@ -6218,8 +6226,10 @@ main(int argc, char **argv, char **envp) /* qsub */
 	basic_envlist = job_env_basic();
 	if (basic_envlist == NULL)
 		exit_qsub(3);
-	if (V_opt)
-		qsub_envlist = env_array_to_varlist(envp);
+	
+	/* qsub_envlist might be needed if -V option is set in default flags.
+	 * It will be unset if not needed then. */
+	qsub_envlist = env_array_to_varlist(envp);
```
suresh-thelkar: We need to call the function env_array_to_varlist() only when desired i.e. when V_opt is set either while calling qsub or through default_qsub_arguments.
 In other words before calling the function set_job_env() make sure that V_opt(This should be set only when qsub is called through -V option or default_qsub_arguments contains -V option) is set or not. If set then only call env_array_to_varlist to populate qsub_envlist and then pass it to set_job_env()

suresh-thelkar:
Also we need to write a few PTL test cases to verify the fix. Following are the test cases that I can think of right now.

1. -V option is not set in qsub but in default_qsub_arguments it is set. 
2. -V option is not set in qsub but in default_qsub_arguments it is set. Submit a job script with and without -V option. 
3. Submit a test case by only setting default_qsub_arguments through job script i.e. default_qsub_arguments should not contain -V option.

If required all the above test cases can be clubbed into a single test case.

test/tests/functional/pbs_default_qsub_argument.py
```diff
@@ -0,0 +1,78 @@
...
+    def test_option_V(self):
+        """
+        This Test case exports an environment variable. It sets
+        default_qsub_argument with -V option and then submits a job.
+        It expects the Variable list to contain the exported environment
+        variable.
+        """
+        os.environ["SET_IN_SUBMISSION"] = "true"
+        self.server.manager(MGR_CMD_SET, SERVER, {'default_qsub_arguments': '-V'})
+        j = Job(ROOT_USER)
```
hirenvadalia: Why do need to submit job as root user?

test/tests/functional/pbs_default_qsub_argument.py
```diff
@@ -0,0 +1,78 @@
...
+        j.create_script("#PBS -N default_qsub_arguments"
+                        "\n#PBS -V" + 
+                        "\nif [ -z '$SET_IN_SUBMISSION' ]" +
+                        "\nthen" +
+                        "\nexit 1" +
+                        "\nelse" +
+                        "\nexit 0" +
+                        "\nfi" )
+        jid = self.server.submit(j)
+        self.server.log_match(jid + ";Exit_status=0", id=jid)
```
hirenvadalia: Missing new line.

test/tests/functional/pbs_default_qsub_argument.py
```diff
@@ -0,0 +1,78 @@
...
+        This Test case exports an environment variable. It sets
+        default_qsub_argument with -V option through a job script and 
+        then submits a job. It expects the Variable list to contain
+        the exported environment variable.
+        """
+        os.environ["SET_IN_SUBMISSION"] = "true"
+        j = Job(ROOT_USER)
+        j.create_script("#PBS -N default_qsub_arguments"
+                        "\n#PBS -V" + 
+                        "\nif [ -z '$SET_IN_SUBMISSION' ]" +
```
hirenvadalia: In `if [ -z '$SET_IN_SUBMISSION' ]`, I doubt that this SET_IN_SUBMISSION will get expanded because it is in single quote

test/tests/functional/pbs_default_qsub_argument.py
sakshamgarg: I want to submit the job through a single user only, need not be a root user. 
This is needed because the sample environment variable which is set will not be accessed by the job running on other users. The sample environment variable is set only for the current user.

test/tests/functional/pbs_default_qsub_argument.py
```diff
@@ -0,0 +1,87 @@
...
+        self.server.expect(JOB, {'Variable_List': (MATCH_RE,
+                           'SET_IN_SUBMISSION=%s' % env_var)},
+                           id=jid)
+
+    def test_option_V_with_script(self):
+        """
+        This Test case exports an environment variable. It sets
+        -V option through a job script directive and then changes
+        the value of the environment variable. It expects another job
+        script to have the changed value of the environment varible.
```
suresh-thelkar: Instead of saying another job script we can say that "The job under execution should get the updated value of the environment variable" since its(Job Script Directives) precedence is higher than the others"

test/tests/functional/pbs_default_qsub_argument.py
```diff
@@ -0,0 +1,87 @@
...
+        """
+        os.environ["SET_IN_SUBMISSION"] = "true"
+        j = Job(ROOT_USER)
```
suresh-thelkar: Instead of ROOT_USER you can actually use one of TEST_USER variables.

test/tests/functional/pbs_default_qsub_argument.py
```diff
@@ -0,0 +1,87 @@
...
+        """
+        This Test case exports an environment variable. It sets
+        default_qsub_argument with -V option and then submits a job.
+        It expects the Variable list to contain the exported environment
+        variable.
+        """
+        os.environ["SET_IN_SUBMISSION"] = "true"
+        self.server.manager(MGR_CMD_SET, SERVER,
+                            {'default_qsub_arguments': '-V'})
+        j = Job(ROOT_USER)
```
suresh-thelkar: Instead of ROOT_USER you can actually use one of TEST_USER variables.

test/tests/functional/pbs_default_qsub_argument.py
```diff
@@ -0,0 +1,87 @@
...
+        jid = self.server.submit(j)
+        env_var = os.environ.get("SET_IN_SUBMISSION")
+        self.server.expect(JOB, {'Variable_List': (MATCH_RE,
+                           'SET_IN_SUBMISSION=%s' % env_var)},
+                           id=jid)
+
+    def test_option_V_with_script(self):
```
suresh-thelkar: Better to rename the test case as test_qsub_opt_V_with_jobscript().

sakshamgarg:
Thanks for all the suggestions and comments. I have updated the PTL test and tested it on the test machine. I have attached the log file for the same.
[LOG_default_qsub_argument_V.txt](https://github.com/PBSPro/pbspro/files/2313241/LOG_default_qsub_argument_V.txt)

src/cmds/qsub.c
```diff
@@ -4914,6 +4914,10 @@ do_submit(char *retmsg)
 		if (rc != 0)
 			return (rc);
 	}
+	
+	/* Get required environment variables */
+	if (V_opt)
```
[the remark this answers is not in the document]
suresh-thelkar: As I said yesterday during our offline discussion it is not good to compare qsub performance with large environment before and after the current code fix. Prior to this code change the function env_array_to_varlist() was not getting called when we set default_qsub_arguments and hence the current bug.
As part of the fix this function call is reintroduced at the right place. So if we compare with and without code fix the performance of qsub after the fix is expected to slow down with -V option. So the ideal test case for performance measure would be to compare qsub with a older version where this bug was not present vs with the current code fix. @sakshamgarg  is verifying this test case and will be posting the results and logs to the ticket.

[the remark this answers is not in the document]
sakshamgarg: I agree with what @suresh-thelkar has suggested in the second part. I did the qsub performance test between the older version where this bug was not there and the current code fix which has been done using the above python script. There is no significant change in the performance test.

sakshamgarg:
I have done manual testing for the test case which uses a job script to set -V option as I was facing user permission issue while trying to automate the test.

This test case exports an environment variable. It sets -V option through a job script directive and then changes the value of the environment variable. It expects the job under execution to get the updated value of the environment variable since its (Job Script Directive's) precedence is higher than the others.

Below is the log for manual testing and I have attached the script files.

[script.sh.txt](https://github.com/PBSPro/pbspro/files/2317666/script.sh.txt)
[sample.py.txt](https://github.com/PBSPro/pbspro/files/2317669/sample.py.txt)

[hadoop@centos2 ~]$ export EXAMPLE=true
[hadoop@centos2 ~]$ qsub script.sh 
12558.centos2
[hadoop@centos2 ~]$ cat sample.e12558
[hadoop@centos2 ~]$ cat sample.o12558
Value of EXAMPLE variable before changing: 
EXAMPLE=true
Value of EXAMPLE variable after changing
EXAMPLE=false
job script returned changed value

suresh-thelkar:
@sakshamgarg, Thanks for making all the suggested changes and PTL test cases. I sign off.

suresh-thelkar:
Good catch @hirenvadalia. It looks like qsub's environment is not passed to qsub daemon that we launch internally. Whenever a request is made to qsub daemon we should pass the new environment to it if V_opt is set. If passing this environment turns out to be a costly operation with respect to performance then we can also think of not redirecting the request to qsub daemon when V_opt is set. 
Also it would be better to check whether this behaviour was working earlier or not. This might help us in measuring the qsub performance with large environment correctly.

sakshamgarg:
Below is the log after fixing the issue as reported by @hirenvadalia.

[saksham@centos ~]$ qmgr -c "s s default_qsub_arguments='-V'"
[saksham@centos ~]$ export EXAMPLE=true
[saksham@centos ~]$ qsub -- /bin/sleep 1000
6.centos
[saksham@centos ~]$ export EXAMPLE=false
[saksham@centos ~]$ qsub -- /bin/sleep 1000
7.centos
[saksham@centos ~]$ qstat -f | grep EXAMPLE
	/jre,LANG=en_US.UTF-8,GDM_LANG=en_US.UTF-8,EXAMPLE=true,
	/jre,LANG=en_US.UTF-8,GDM_LANG=en_US.UTF-8,EXAMPLE=false,
[saksham@centos ~]$ 

This is the log for the PTL test case.
[res_pbs_default_qsub_arguments.txt](https://github.com/PBSPro/pbspro/files/2339553/res_pbs_default_qsub_arguments.txt)

I am also attaching the logs for the manual testing as described in previous comment.
[log_manualTesting.txt](https://github.com/PBSPro/pbspro/files/2339223/log_manualTesting.txt)

src/cmds/qsub.c
```diff
@@ -4914,6 +4914,13 @@ do_submit(char *retmsg)
 		if (rc != 0)
 			return (rc);
 	}
+	
+	/* Free the environment variable list if V_opt is not set */
+	if (!V_opt && qsub_envlist!=NULL)
```
hirenvadalia: Please add space before and after `!=`

test/tests/functional/pbs_default_qsub_argument.py
```diff
@@ -0,0 +1,61 @@
...
+
+from tests.functional import *
+
+
+class TestDefaultQsubArgument(TestFunctional):
+    """
+    validate default_qsub_argument option.
+    """
+
+    def test_option_V(self):
```
hirenvadalia: Please add test for recently discovered use cases (EXAMPLE=true and false one)

test/tests/functional/pbs_default_qsub_argument.py
```diff
@@ -0,0 +1,61 @@
...
+# Use of Altair’s trademarks, including but not limited to "PBS™",
+# "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
+# trademark licensing policies.
+
+from tests.functional import *
+
+
+class TestDefaultQsubArgument(TestFunctional):
+    """
+    validate default_qsub_argument option.
```
hirenvadalia: `default_qsub_argument` == `default_qsub_arguments`

Also I suggest this docstring to something like this: Validate whether qsub sends all environ when -V is enabled via default_qsub_arguments and or on command line.

test/tests/functional/pbs_default_qsub_argument.py
```diff
@@ -0,0 +1,61 @@
...
+        """
+        os.environ["SET_IN_SUBMISSION"] = "true"
+        self.server.manager(MGR_CMD_SET, SERVER,
+                            {'default_qsub_arguments': '-V'})
+        j = Job(PBSROOT_USER)
+        jid = self.server.submit(j)
+        env_var = os.environ.get("SET_IN_SUBMISSION")
+        self.server.expect(JOB, {'Variable_List': (MATCH_RE,
+                           'SET_IN_SUBMISSION=%s' % env_var)},
+                           id=jid)
```
hirenvadalia: Please add test case for testing passing full environ via -V on qsub command line

test/tests/functional/pbs_default_qsub_argument.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+class TestDefaultQsubArgument(TestFunctional):
+    """
+    validate default_qsub_argument option.
+    """
+
+    def test_option_V(self):
+        """
+        This Test case exports an environment variable. It sets
```
hirenvadalia: This docstring looks like steps of test. I suggest to changing it to something like: Test exporting variable when -V is enabled in default_qsub_arguments

test/tests/functional/pbs_default_qsub_argument.py
```diff
@@ -0,0 +1,61 @@
...
+        default_qsub_argument with -V option and then submits a job.
+        It expects the Variable list to contain the exported environment
+        variable.
+        """
+        os.environ["SET_IN_SUBMISSION"] = "true"
+        self.server.manager(MGR_CMD_SET, SERVER,
+                            {'default_qsub_arguments': '-V'})
+        j = Job(PBSROOT_USER)
```
hirenvadalia: You are setting `SET_IN_SUBMISSION` in current user's session and submitting job as pbsroot user, which internally PTL will do sudo qsub, and `SET_IN_SUBMISSION` might not get passed by sudo.
I suggest to submit job as current user's only, you can use `self.du.get_current_user()` to get current user name

test/tests/functional/pbs_default_qsub_argument.py
```diff
@@ -0,0 +1,61 @@
...
+        jid = self.server.submit(j)
+        env_var = os.environ.get("SET_IN_SUBMISSION")
```
hirenvadalia: Why to get is back from os.environ while we know value of `SET_IN_SUBMISSION`?

test/tests/functional/pbs_default_qsub_argument.py
```diff
@@ -0,0 +1,61 @@
...
+        self.server.expect(JOB, {'Variable_List': (MATCH_RE,
+                           'SET_IN_SUBMISSION=%s' % env_var)},
```
hirenvadalia: I suggest to change `'SET_IN_SUBMISSION=%s' % env_var` to `'SET_IN_SUBMISSION=true` since we already know value of `SET_IN_SUBMISSION`

test/tests/functional/pbs_default_qsub_argument.py
```diff
@@ -0,0 +1,61 @@
...
+# under a commercial license agreement.
+#
+# Use of Altair’s trademarks, including but not limited to "PBS™",
+# "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
+# trademark licensing policies.
+
+from tests.functional import *
+
+
+class TestDefaultQsubArgument(TestFunctional):
```
hirenvadalia: I suggest to put these test cases in already existing test suite under https://github.com/PBSPro/pbspro/blob/master/test/tests/functional/pbs_passing_environment_variable.py

src/cmds/qsub.c
```diff
@@ -6218,8 +6225,8 @@ main(int argc, char **argv, char **envp) /* qsub */
 	basic_envlist = job_env_basic();
 	if (basic_envlist == NULL)
 		exit_qsub(3);
-	if (V_opt)
-		qsub_envlist = env_array_to_varlist(envp);
+	
+	qsub_envlist = env_array_to_varlist(envp);
```
hirenvadalia: Calling env_array_to_varlist in every condition with definitely decrease performance of qsub even with out -V because env_array_to_varlist is little heavy function which traverse whole environ char by char.

sakshamgarg: I agree with your statement that it will affect performance of qsub.
The other alternative solution to this problem was to connect to the server at the very starting, before it was loading value to qsub_envlist in the "main" function, and fetch the default qsub arguments from the server. But this renders the use of a daemon pointless. 
Although this result in performance degradation but it is the only (valid) solution for this bug. 
If needed, we can enforce that -V option never be set in default arguments to ensure performance.

src/cmds/qsub.c
```diff
@@ -5809,6 +5822,25 @@ do_daemon_stuff(void)
...
 
+		/*
+		 * Check if -V option set in default_qsub_arguments.
+		 * If the option is set, then don't submit the job using
+		 * background qsub and send an error code (VOPT_ERROR) to the
+		 * foreground process.
+		 */
+		refresh_dfltqsubargs();
+		rc = parse_dfltqsubargs(retmsg);
+		if(rc || V_opt){
```
suresh-thelkar: You can make it more readable by saying rc == 0 && V_opt.

src/cmds/qsub.c
```diff
@@ -6041,12 +6074,17 @@ daemon_submit(int *daemon_up, int *do_regular_submit)
 				 * We were able to send data to the background daemon.
 				 * Now, even if we fail to read back response from
 				 * background, we do not want to submit again.
+				 * If -V option is detected in background qsub, then
+				 * do a regular submit.
 				 */
-				*do_regular_submit = 0;
+				if(rc != VOPT_ERROR)
```
suresh-thelkar: Instead of checking whether VOPT_ERROR is not set, do a check for rc == VOPT_ERROR, if yes bypass the following code. Basically you need to optimise the code here.

subhasisb:
@sakshamgarg the server forces the background daemons to update the default qsub args (When it changes via qmgr), this code is already there. We need to check it after do_dir() in do_submit() and return with a "Unique" error code. This should percolate all the way back to the foreground qsub as a submission return code (only that it is not), and we need to check that in the foreground and set do_forground_qsub=1

src/cmds/qsub.c
```diff
@@ -5982,6 +5961,8 @@ fork_and_stay(void)
 		/* set single threaded mode */
 		pbs_client_thread_set_single_threaded_mode();
 
+		/* set when background qsub is running */
+        is_background = 1;
```
subhasisb: indentation

src/cmds/qsub.c
```diff
@@ -236,6 +236,7 @@ static char *pbs_hostvar = NULL; /* buffer containing ",PBS_O_HOST=" and host na
 static int pbs_o_hostsize = sizeof(",PBS_O_HOST=") + 1; /* size of prefix for hostvar */
 static char *display; /* environment variable DISPLAY */
 
+static int is_background = 0; /* flag to check if background qsub is active */
```
subhasisb: Change the comment to something like "am i the background process? This variable is set only once and is read-only afterwards"

src/cmds/qsub.c
```diff
@@ -4919,12 +4898,32 @@ do_submit(char *retmsg)
...
+		}
+		if (rc != 0)
+			return (rc);
+	}
 
+	/* 
+	 * get environment variable if -V option is set. Return the code
+	 * VOPT_SET if -V option is detected in background qsub.
+	 */
 	if(V_opt) {
```
subhasisb: space before (

src/cmds/qsub.c
```diff
@@ -4919,12 +4898,32 @@ do_submit(char *retmsg)
...
+		if(is_background)
```
subhasisb: space before (

src/cmds/qsub.c
```diff
@@ -6074,27 +6055,28 @@ daemon_submit(int *daemon_up, int *do_regular_submit)
...
 				 * Something bad happened, either background submitted
 				 * and failed to send us response, or it failed before
-				 * submitting.
+				 * submitting. If background qsub detects -V option, then
+				 * submit the job through foreground.
 				 */
-				rc = -1;
-				sprintf(retmsg, "Failed to recv data from background qsub\n");
-				/* Error message will be printed in caller */
+				if(rc != VOPT_SET){
```
subhasisb: space before (

src/cmds/qsub.c
```diff
@@ -164,7 +164,7 @@
 #define QSUB_DMN_TIMEOUT_LONG 60 /* timeout for qsub background process */
 #define QSUB_DMN_TIMEOUT_SHORT 5
 
-#define VOPT_ERROR 4321 /* Error code if -V option is detected in background qsub*/
+#define VOPT_SET 4 /* return code if -V option is detected in background qsub*/
```
subhasisb: It is probably best to rename this return code to be more generic, so that it can be used later for other similar purposes, if necessary. Rename it to something like DMN_CANT_SERVE (or think something better)

src/cmds/qsub.c
```diff
@@ -4919,12 +4898,32 @@ do_submit(char *retmsg)
...
+			return (rc);
+	}
 
+	/* 
+	 * get environment variable if -V option is set. Return the code
+	 * VOPT_SET if -V option is detected in background qsub.
+	 */
 	if(V_opt) {
+		if(is_background)
+			return VOPT_SET;
```
subhasisb: Also, it would be better if we can check this in do_daemon_stuff and exit the daemon in this return case. This way future qsubs will not have to continuously consult with the background qsub just to know that it should not be going to the background

src/cmds/qsub.c
```diff
@@ -6062,20 +6071,19 @@ daemon_submit(int *daemon_up, int *do_regular_submit)
...
+				rc == DMN_REFUSE_EXIT) {
 
 				/*
 				 * Something bad happened, either background submitted
 				 * and failed to send us response, or it failed before
 				 * submitting. If background qsub detects -V option, then
 				 * submit the job through foreground.
 				 */
-				if(rc != VOPT_SET){
+				if (rc != DMN_REFUSE_EXIT){
```
subhasisb: still has a indentation issue between ){

src/cmds/qsub.c
```diff
@@ -4915,6 +4922,16 @@ do_submit(char *retmsg)
 			return (rc);
 	}
 
+	/* 
+	 * get environment variable if -V option is set. Return the code
```
suresh-thelkar: Can you repharse it as "get the variable _environ_ which points to an array of pointers to strings of the user environment if -V option is set....".

sakshamgarg:
Attaching the logs of PTL tests.
[ptl_output_-V.txt](https://github.com/PBSPro/pbspro/files/2374600/ptl_output_-V.txt)

sakshamgarg:
I have squashed all commits and rebased.

PR #791, opened 2018-08-24
Scheduler may oversubscribe node while confirming reservation

arungrover:
#### Bug/feature Description
It is found that in some cases scheduler may end up oversubscribing the node. Below are the two cases where this can happen:
* If a reservation that is scheduled to start in future is deleted, the server puts the reservation in 'RESV_BEING_DELETED' state and goes on to delete the jobs present in that reservation. Now during this time if the scheduling cycle is triggered, then the scheduler assumes that the reservation must be running and it is going to release resources very soon. This makes scheduler release resources requested by the reservation when it is simulating future. Due to this problem scheduler sometimes end up oversubscribing the node.
* Second problem was introduced in "[PP-479](https://pbspro.atlassian.net/browse/PP-479)". After this change, when scheduler queries server to select all the jobs, it also gets a list of expired sub jobs of job-array. This was not the case before because the server was only reporting array-parent and running sub jobs to the scheduler. Now, upon receiving expired sub jobs scheduler releases more resources during simulation (for top jobs or reservation) and can end up oversubscribing the node.

#### Affected Platform(s)
All

#### Solution Description
It has 3 way fix - 
  - Server now does not report expired sub jobs to the scheduler. It only reports queued array-parent, running sub jobs and other non-finished jobs to the scheduler.
 - Scheduler detects that if it received reservation in being_deleted state or sub jobs in non-running state then it ignores it.
- While simulating, if scheduler ever falls into a case where resources assigned on node/server/queue are turning negative then it sets them to 0.

#### Testing logs/output
[test_est.txt](https://github.com/PBSPro/pbspro/files/2316682/test_est.txt)

[testlogs_node_oversubscription.txt](https://github.com/PBSPro/pbspro/files/2316657/testlogs_node_oversubscription.txt)
[test_array_jobs.txt](https://github.com/PBSPro/pbspro/files/2319437/test_array_jobs.txt)

#### Checklist:

* NOTE: I've added a PTL test to make sure top jobs do not get affected when there are expired subjobs in the system. To test reservation case, I used gdb and that is why there is no automated test  for that scenario.

### DIFF
--- a/src/scheduler/job_info.c
+++ b/src/scheduler/job_info.c
@@ -613,12 +613,14 @@ query_jobs(status *policy, int pbs_sd, queue_info *qinfo, resource_resv **pjobs,
 	}
 	resresv_arr[num_prev_jobs] = NULL;
 
-	cur_job = jobs;
 	err = new_schd_error();
-	if(err == NULL)
+	if(err == NULL) {
+		free_resource_resv_array(resresv_arr);
 		return NULL;
+	}
 
-	for (i = num_prev_jobs; cur_job != NULL; i++) {
+	i = num_prev_jobs;
+	for (cur_job = jobs; cur_job != NULL; cur_job = cur_job->next) {
 		char *selectspec = NULL;
 		if ((resresv = query_job(cur_job, qinfo->server, err)) ==NULL) {
 			free_schd_error(err);
@@ -638,9 +640,16 @@ query_jobs(status *policy, int pbs_sd, queue_info *qinfo, resource_resv **pjobs,
 			 * header.  We to continue adding valid jobs to our array.  We're
 			 * freeing what we allocated and ignoring this job completely.
 			 */
-			i--;
 			free_resource_resv(resresv);
-			cur_job = cur_job->next;
+			continue;
+		}
+
+		/* Make sure scheduler does not process a subjob in undesirable state*/
+		if (resresv->job->is_subjob && !resresv->job->is_running &&
+			!resresv->job->is_suspended && !resresv->job->is_provisioning) {
+			schdlog(PBSEVENT_SCHED, PBS_EVENTCLASS_RESV, LOG_DEBUG,
+				resresv->name, "Subjob found in undesirable state, ignoring this job");
+			free_resource_resv(resresv);
 			continue;
 		}
 
@@ -890,10 +899,9 @@ query_jobs(status *policy, int pbs_sd, queue_info *qinfo, resource_resv **pjobs,
 			clear_schd_error(err);
 		}
 
-		resresv_arr[i] = resresv;
-		resresv_arr[i+1] = NULL;	/* Make array searchable */
+		resresv_arr[i++] = resresv;
+		resresv_arr[i] = NULL;	/* Make array searchable */
 
-		cur_job = cur_job->next;
 	}
 	resresv_arr[i] = NULL;
 
--- a/src/scheduler/node_info.c
+++ b/src/scheduler/node_info.c
@@ -447,6 +447,8 @@ query_node_info(struct batch_status *node, server_info *sinfo)
 			count = strtol(attrp->value, &endp, 10);
 			if (*endp == '\0')
 				ninfo->last_used_time = count;
+		} else if (!strcmp(attrp->name, ATTR_NODE_resvs)) {
+			ninfo->resvs = break_comma_list(attrp->value);
 		}
 		attrp = attrp->next;
 	}
@@ -512,6 +514,7 @@ new_node_info()
 	new->mom = NULL;
 	new->port = pbs_rm_port;
 	new->jobs = NULL;
+	new->resvs = NULL;
 	new->job_arr = NULL;
 	new->run_resvs_arr = NULL;
 	new->res = NULL;
@@ -600,6 +603,9 @@ free_node_info(node_info *ninfo)
 
 		if (ninfo->jobs != NULL)
 			free_string_array(ninfo->jobs);
+
+		if (ninfo->resvs != NULL)
+			free_string_array(ninfo->resvs);
 
 		if (ninfo->job_arr != NULL)
 			free(ninfo->job_arr);
@@ -1270,6 +1276,7 @@ dup_node_info(node_info *onode, server_info *nsinfo,
 	nnode->priority = onode->priority;
 
 	nnode->jobs = dup_string_array(onode->jobs);
+	nnode->resvs = dup_string_array(onode->resvs);
 	if (flags & DUP_INDIRECT)
 		nnode->res = dup_ind_resource_list(onode->res);
 	else
@@ -1802,9 +1809,10 @@ update_node_on_end(node_info *ninfo, resource_resv *resresv, char *job_state)
 						res->assigned -= resreq->amount;
 						if (res->assigned < 0) {
 							snprintf(logbuf, MAX_LOG_SIZE,
-								"Setting %s assigned to %.2lf", res->name, res->assigned);
+								"%s turned negative %.2lf, setting it to 0", res->name, res->assigned);
 							schdlog(PBSEVENT_DEBUG, PBS_EVENTCLASS_NODE,
 								LOG_DEBUG, ninfo->name, logbuf);
+							res->assigned = 0;
 						}
 						if (res->def == getallres(RES_NCPUS)) {
 							ninfo->loadave -= resreq->amount;
@@ -2397,7 +2405,7 @@ eval_placement(status *policy, selspec *spec, node_info **ninfo_arr, place *pl,
 	 * remark: reorder_nodes doesn't reorder in place, returns
 	 *         a ptr to a reordered static array
 	 */
-	if ((pl->pack && spec->total_chunks == 1 && nspec_arr != NULL) ||
+	if ((pl->pack && spec->total_chunks == 1) ||
 		(conf.provision_policy == AVOID_PROVISION && resresv->aoename != NULL) ||
 		(resresv->is_resv && resresv->resv != NULL && resresv->resv->check_alternate_nodes))
 		nptr = reorder_nodes(ninfo_arr, resresv);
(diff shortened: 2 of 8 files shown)


src/scheduler/job_info.c
```diff
@@ -644,6 +644,17 @@ query_jobs(status *policy, int pbs_sd, queue_info *qinfo, resource_resv **pjobs,
 			continue;
 		}
 
+		/* Make sure scheduler does not process a subjob in undesirable */
```
bhroam: Maybe "in an undesirable state"?

test/tests/functional/pbs_check_calendaring.py
```diff
@@ -0,0 +1,98 @@
...
+        In this test we test that the top job which gets added to the
+        calendar has estimated start time correctly set for future.
+        """
+
+        self.scheduler.set_sched_config({'strict_ordering': 'true all'})
+        a = {'resources_available.ncpus': 1}
+        self.server.manager(MGR_CMD_SET, NODE, a, self.mom.shortname)
+
+        a = {'backfill_depth': 2}
+        self.server.manager(MGR_CMD_SET, SERVER, {'backfill_depth': '2'})
```
bhroam: Why set 'a' if you don't use it.

test/tests/functional/pbs_check_calendaring.py
```diff
@@ -0,0 +1,98 @@
...
+
+        res_req = {'Resource_List.select': '1:ncpus=1',
+                   'Resource_List.walltime': 10,
+                   'array_indices_submitted': '1-10'}
+        j1 = Job(TEST_USER)
+        j1.set_attributes(res_req)
```
bhroam: Consider doing this in one step by passing attrs=res_req to Job()

test/tests/functional/pbs_check_calendaring.py
```diff
@@ -0,0 +1,98 @@
...
+
+        self.server.expect(JOB, {'job_state': 'X'}, j1_sub1)
+        self.server.expect(JOB, {'job_state': 'R'}, j1_sub2)
+        self.server.expect(JOB, {'job_state': 'Q'}, jid2)
+        job1 = self.server.status(JOB, id=jid1)
+        job2 = self.server.status(JOB, id=jid2)
+        time_now = int(time.time())
+
+        # get estimated start time of both the jobs
+        self.assertTrue('estimated.start_time' in job1[0])
```
bhroam: use assertIn()

src/scheduler/job_info.c
arungrover: I'll change it, sorry about that typo

src/server/req_select.c
```diff
@@ -494,6 +494,9 @@ select_job(job *pjob, struct select_list *psel, int dosubjobs, int dohistjobs)
 	else if ((dosubjobs != 2) &&
 		(pjob->ji_qs.ji_svrflags & JOB_SVFLG_SubJob))
 		return 0;	/* don't bother to look at sub job */
+	else if ((dosubjobs == 2) && (pjob->ji_qs.ji_svrflags & JOB_SVFLG_SubJob) &&
+		(pjob->ji_qs.ji_state != JOB_STATE_RUNNING)) /* select only running subjobs */
```
agrawalravi90: Just checking, what about subjobs in states 'H', 'M', 'S', 'T', 'U' or 'W'? I'm not sure if subjobs can go into these states or not, but just in case, do we need to select subjobs in any of these states?

src/scheduler/resv_info.c
```diff
@@ -231,6 +220,28 @@ query_reservations(server_info *sinfo, struct batch_status *resvs)
 			continue;
 		}
 
+		/* Make sure it is not a future reservation that is being deleted, if so ignore it */
+		if (resresv->resv->resv_state == RESV_BEING_DELETED && resresv->start > sinfo->server_time) {
```
agrawalravi90: one question: if a reservation is in RESV_BEING_DELETED, is it for certain that the server has not freed the resources on the vnodes yet? for instance, If a user deletes a previously running reservation so that it's in RESV_BEING_DELETED state when the scheduler queries it, is it possible that the reservation's vnodes have been free'd even though it is in state RESV_BEING_DELETED? If yes, then this we can fall into the same trap even with this check right?

src/scheduler/resv_info.c
```diff
@@ -231,6 +220,28 @@ query_reservations(server_info *sinfo, struct batch_status *resvs)
...
+			schdlog(PBSEVENT_SCHED, PBS_EVENTCLASS_RESV, LOG_DEBUG,
+				resresv->name, "Future reservation is being deleted, ignoring this reservation");
+
+			sinfo->num_resvs--;
+			free_resource_resv(resresv);
+			cur_resv = cur_resv->next;
```
agrawalravi90: Consider converting this while loop into a for loop so that cur_resv = cur_resv->next is part of the for loop header.

test/tests/functional/pbs_check_calendaring.py
```diff
@@ -0,0 +1,100 @@
...
+#
+# Use of Altair’s trademarks, including but not limited to "PBS™",
+# "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
+# trademark licensing policies.
+
+import time
+from tests.functional import *
+
+
+class TestCalendaring(TestFunctional):
```
agrawalravi90: Nit: How about renaming the name of this file to pbs_calendaring.py ? It will go well with the test suite name TestCalendaring, and it will be more noticeable that way among the many test files that we have.

test/tests/functional/pbs_check_calendaring.py
```diff
@@ -0,0 +1,100 @@
...
+
+    """
+    This test suite tests if PBS scheduler calendars events correctly
+    """
+    def test_topjob_start_time(self):
+        """
+        In this test we test that the top job which gets added to the
+        calendar has estimated start time correctly set for future when
+        job history is enabled and opt_backfill_fuzzy is turned off.
```
agrawalravi90: Is turning off opt_backfill_fuzzy necessary for doing any kind of calendaring testing? If yes, then please add a custom setUp to this testsuite and turn it off there. If not, then can you please explain why this test needs it to be turned off?

agrawalravi90: it might also be useful to set strict_ordering in setUp.

test/tests/functional/pbs_check_calendaring.py
```diff
@@ -0,0 +1,100 @@
...
+import time
+from tests.functional import *
+
+
+class TestCalendaring(TestFunctional):
+
+    """
+    This test suite tests if PBS scheduler calendars events correctly
+    """
+    def test_topjob_start_time(self):
```
agrawalravi90: Do we also need to add a test for reservations?

src/scheduler/job_info.c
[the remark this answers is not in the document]
agrawalravi90: Please ignore this, I think bhroam provided the same comment, and I think you already addressed it

src/scheduler/job_info.c
```diff
@@ -644,6 +644,17 @@ query_jobs(status *policy, int pbs_sd, queue_info *qinfo, resource_resv **pjobs,
 			continue;
 		}
 
+		/* Make sure scheduler does not process a subjob in undesirable state*/
+		if (resresv->job->is_subjob && !resresv->job->is_running &&
```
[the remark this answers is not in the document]
arungrover: This comment made me worry that I missed to check for checkpointed job. Then I found out that checkpointing is not supported for subjobs. So we are covered there.

src/server/req_select.c
arungrover: Subjobs cannot go into some of the states like 'H', it can go into the suspended state but at that time the state is actually 'R' it is the sub-state that actually tells us whether it is suspended. For all other states, the scheduler does not really care.

src/scheduler/job_info.c
```diff
@@ -644,6 +644,17 @@ query_jobs(status *policy, int pbs_sd, queue_info *qinfo, resource_resv **pjobs,
...
+			!resresv->job->is_suspended && !resresv->job->is_provisioning) {
+			schdlog(PBSEVENT_SCHED, PBS_EVENTCLASS_RESV, LOG_DEBUG,
+				resresv->name, "Subjob found in undesirable state, ignoring this job");
+			i--;
```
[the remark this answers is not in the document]
arungrover: I'll change it

src/scheduler/resv_info.c
arungrover: Server Ideally should not free resources unless the reservation has been completely deleted. In case we are also safe because the reservation which is being deleted has not yet started, so no resources would be assigned to the reservation and scheduler does not really care about a future reservation that is being deleted.

arungrover: @agrawalravi90 I now get the problem you are stating, thanks for explaining it. I'll fix it

src/scheduler/resv_info.c
arungrover: Done

test/tests/functional/pbs_check_calendaring.py
arungrover: Well, Since there is only one test in the suite, I have all the config settings in the test case itself. Also, since the file is pretty generic, I suppose moving it to a setup may not really be helpful for tests someone might add in future.
The reason I'm turning backfill_fuzzy off is because the subjobs that are in expired state have only 10 seconds of walltime and opt_backfill_fuzzy by default (when unset) tries to simulate events at 60 seconds boundary, which means the expired subjob's end event will be  performed and running subjob's end event will also be performed (since both of them lie in this 60 second boundary), after performing both the events the third subjob's run event gets placed at the right time (which is after the running subjob's end event).
If I turn off backfill-fuzzy then we will check if the job can run or not after performing each event which will mean now third subjob's run event will be placed right after the expired subjob (which will be in the past).

test/tests/functional/pbs_check_calendaring.py
arungrover: Reservations bug is more of a timing issue so I've tested it manually with the help of gdb.

test/tests/functional/pbs_check_calendaring.py
arungrover: I'll rename the file

test/tests/functional/pbs_check_calendaring.py
agrawalravi90: Thanks for clarifying Arun. Could you please add a comment either inline in the test, or in the header, explaining why opt_backfill_fuzzy should be turned off? I bet I won't remember this the next time I look at this test :)

I'm okay with no custom setUp()

arungrover: I'll add a comment

arungrover:
Thanks for reviewing @agrawalravi90, I've addressed your review comments. @bhroam Please have a look again.

arungrover:
[node_oversubscription_running_resv.txt](https://github.com/PBSPro/pbspro/files/2334914/node_oversubscription_running_resv.txt)

arungrover:
@agrawalravi90, @bhroam I've addressed some of the warnings codacy found, Please have a look at the last commit. Change is small :)

src/scheduler/node_info.c
```diff
@@ -4763,7 +4763,7 @@ reorder_nodes(node_info **nodes, resource_resv *resresv)
 	if (last_node_name[0] == '\0')
 		strcpy(last_node_name, nodes[0]->name);
 
-	if (resresv->aoename != NULL && conf.provision_policy == AVOID_PROVISION) {
+	if (resresv != NULL && resresv->aoename != NULL && conf.provision_policy == AVOID_PROVISION) {
```
bhroam: Wouldn't it be better to factor the if(resresv != NULL) out of this block and the block two higher and but both of them in an if (reseresv != NULL) if block?

src/scheduler/queue_info.c
```diff
@@ -326,7 +326,7 @@ query_queues(status *policy, int pbs_sd, server_info *sinfo)
 	pbs_statfree(queues);
 	free_schd_error(sch_err);
 	if (err) {
-		if (qinfo_arr != NULL)
+		if (qinfo_arr[0] != NULL)
```
bhroam: You should leave this one alone.  If qinfo_arr[0] == NULL, then qinfo_arr has memory allocated to it.  This was correct before.

agrawalravi90:
Changes still look good Arun, thanks for fixing all the issues, some of them seem pretty bad

src/scheduler/node_info.c
arungrover: I could do that, I just wasn't comfortable changing a whole lot of code that I'm not working on and then end up with doing a lot of testing around it.
If you feel strongly about it then I'll change it, else I'd like to leave it as is.

src/scheduler/queue_info.c
arungrover: I think the check is redundant then. At line 323 we are already dereferencing qinfo_arr so Ideally, we don't even need a check before freeing it up.

bhroam: All the checks are redundant.  I didn't know it at the time I wrote the code, but free(NULL) is a noop.  You don't need these.  Until we go through and remove them all, I'd rather leave them in for consistency.

src/scheduler/node_info.c
bhroam: It's redundant though.  I'd rather do it right instead of having redundant checks.

agrawalravi90:
sorry for the delay, looks good Arun.

arungrover:
@PBSPro/pbspro-maintainers the pull request is ready for merge. Please take it forward.

arungrover:
Thanks @agarwalravi90

PR #803, opened 2018-09-05
memory is not freed when action function fails in mgr_set_attr

nithinj:
#### Bug/feature Description
* memory is not freed when action function fails in mgr_set_attr
* memory leak in set_entlim

* When action function was failing PBS was just freeing the main attribute structure instead of freeing individual elements in the array. 
* memory allocated by entlim_get_next() is not freed.

#### Solution Description
* Freeing each item in the attribute
* Updating a duplicate test suite name
* Adding more hook related memory leaks to suppress file
* fix for memory leak in set_entlim 

#### Testing logs/output
[memleak1 logs.txt](https://github.com/PBSPro/pbspro/files/2351151/memleak1.logs.txt)
[set_entlim.log](https://github.com/PBSPro/pbspro/files/2356591/set_entlim.log)

#### Checklist:

CC: @subhasisb , @bayucan

### DIFF
--- a/src/lib/Libattr/attr_fn_entlim.c
+++ b/src/lib/Libattr/attr_fn_entlim.c
@@ -848,6 +848,7 @@ set_entlim(attribute *old, attribute *new, enum batch_op op)
 					if (entlim_replace(pkey->key, newptr, oldctx, svr_freeleaf) != 0) {
 						/* failed to add */
 						svr_freeleaf(newptr);
+						free(pkey);
 						return (PBSE_SYSTEM);
 					}
 				}
@@ -912,7 +913,7 @@ set_entlim(attribute *old, attribute *new, enum batch_op op)
 			/* having removed one or more elements from the value tree */
 			/* see if any entries are left or if the value is now null */
 			pkey = NULL;
-			if (entlim_get_next(pkey, oldctx) == NULL) {
+			if ((pkey = entlim_get_next(pkey, oldctx)) == NULL) {
 				/* no entries left set, clear the entire attribute */
 				free_entlim(old);
 				/* set _MODIFY flag so up level functions */
@@ -920,6 +921,7 @@ set_entlim(attribute *old, attribute *new, enum batch_op op)
 				old->at_flags |= ATR_VFLAG_MODIFY;
 				return (0);
 			}
+			free(pkey);
 			break;
 
 		default:	return (PBSE_INTERNAL);
--- a/valgrind.supp
+++ b/valgrind.supp
@@ -671,3 +671,53 @@
    fun:pbs_python_populate_attributes_to_python_class
    fun:*
 }
+{
+   From PBSPro (server) - Suppress intentional unfreed memory from loading hook script at mgr_hook_import 
+   Memcheck:Leak
+   fun:malloc
+   fun:strdup
+   fun:pbs_python_ext_alloc_python_script
+   fun:mgr_hook_import
+   fun:req_manager
+   fun:dispatch_request
+   fun:process_request
+   fun:wait_request
+   fun:main
+}
+{
+   From PBSPro (server) - Suppress intentional unfreed memory from loading hook script at mgr_hook_import 
+   Memcheck:Leak
+   fun:malloc
+   fun:pbs_python_ext_alloc_python_script
+   fun:mgr_hook_import
+   fun:req_manager
+   fun:dispatch_request
+   fun:process_request
+   fun:wait_request
+   fun:main
+}
+{
+   From PBSPro (server) - Suppress intentional unfreed memory from allocating hook data at mgr_hook_create 
+   Memcheck:Leak
+   fun:malloc
+   fun:hook_alloc
+   fun:mgr_hook_create
+   fun:req_manager
+   fun:dispatch_request
+   fun:process_request
+   fun:wait_request
+   fun:main
+}
+{
+   From PBSPro (server) - Suppress intentional unfreed memory from allocating hook name at mgr_hook_create 
+   Memcheck:Leak
+   fun:malloc
+   fun:strdup
+   fun:set_hook_name
+   fun:mgr_hook_create
+   fun:req_manager
+   fun:dispatch_request
+   fun:process_request
+   fun:wait_request
+   fun:main
+}
--- a/src/server/req_manager.c
+++ b/src/server/req_manager.c
@@ -878,6 +878,7 @@ mgr_set_attr2(attribute *pattr, attribute_def *pdef, int limit, svrattrl *plist,
 						attr_atomic_kill(new, pdef, limit);
 						attr_atomic_kill(pre_copy, pdef, limit);
 						attr_atomic_kill(attr_save, pdef, limit);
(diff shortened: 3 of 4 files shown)


bhroam:
Thanks for fixing the leaks.  Are you sure about the 3rd and 4th leaks you are suppressing? They might be in the hooks code, but they aren't in python.

nithinj:
Thanks @bhroam for the review. Hooks created as part of 3rd and 4th leak are part of global structure `svr_allhooks`. So they're intentional and you can also find similar stack trace in suppress file from `hook_recov()`

valgrind.supp
```diff
@@ -671,3 +671,52 @@
    fun:pbs_python_populate_attributes_to_python_class
    fun:*
 }
+{
+   From PBSPro (server) - Suppress intentional unfreed memory from loading hook script at mgr_hook_import 
+   Memcheck:Leak
```
subhasisb: @nithinj how can we confirm that these allocations are intended not to be freed.

bhroam:
@nithinj valgrind does not report memory held at the end of a process as a leak.  If it reports a leak, it is really being leaked.  Are we sure about this?  Once something is added to the suppression file, it is hard to remove it since we won't see the leaks any more.

nithinj:
Valgrind is reporting this as a `possible leak` and not a `definite leak`. And we're freeing the structure when the hook is getting deleted and during server shutdown. 
I'll need more time to find why this is being reported as a `possible leak`. I'll get back to you on this. Thanks!

nithinj:
So we are not freeing the hooks at the server shutdown. I just added code to verify this behavior at svr_shutdown(). The code will purge all hooks from svr_allhooks before shutting down. After this change valgrind is not complaining anymore. It is not a leak as it is evident that we are tracking the hook as part of global structure svr_allhooks.
@subhasisb figured out that valgrind complain is coming due to the way we are handling the linked list. It is worth spending some time to fix them so that valgrind won't generate anymore warning. I'll file an issue and would like to proceed on this one.

src/lib/Libattr/attr_fn_entlim.c
```diff
@@ -912,14 +912,15 @@ set_entlim(attribute *old, attribute *new, enum batch_op op)
...
-			if (entlim_get_next(pkey, oldctx) == NULL) {
+			if ((pkey = entlim_get_next(pkey, oldctx)) == NULL) {
 				/* no entries left set, clear the entire attribute */
 				free_entlim(old);
 				/* set _MODIFY flag so up level functions */
 				/* know the attribute has been changed    */
 				old->at_flags |= ATR_VFLAG_MODIFY;
 				return (0);
 			}
+			free(pkey);
```
Shrini-h: Should we also free(pkey) in the two while loops occurring in this function?

nithinj: pkey memory allocation happens at the beginning of the tree traversal. It is freed when the traversal is complete. But if we return the control from the loop in between, the memory has to be freed. I've addressed this with my recent commit. Please check.

subhasisb:
I was always confused about valgrind reporting "possibly leaked" cases - especially associated to the pbs_list_link data structure. I am attaching a test program that demonstrates that a sample code with pbs_list_link will show possible leaks even when there are no leaks (whereas it should actually show those blocks as "still reachable"). Yet to figure out why Valgrind gets confused...so any help is appreciated.
[tst2.c.txt](https://github.com/PBSPro/pbspro/files/2427334/tst2.c.txt) (please rename to .c extension)

subhasisb:
Sorry, i had attached the wrong program. The tst2.c is the one that uses a "regular" style linked list and does NOT show any possibly leaked. @Shrini-h when you advance the pointer it is supposed to show the possibly leaked, so that's also okay.

Here is the code using the pbs list link. This shows possibly leaked even when it seems it should not. 
[tst.c.txt](https://github.com/PBSPro/pbspro/files/2433259/tst.c.txt)

Shrini-h:
@subhasisb , I wrote another simple code that can demonstrate the difference bw `possibly lost` and `still reachable`. Compiling the file [test3.c.txt](https://github.com/PBSPro/pbspro/files/2434381/test3.c.txt) with `-DI_LIKE_POSSIBLE_LEAK` will make valgrind throw `possibly lost` and without the flag the same memory is shown as `still reachable`, that's cause we are storing pointers to the middle of allocated block.

And for the [tst.c.txt](https://github.com/PBSPro/pbspro/files/2433259/tst.c.txt) that you shared, the reason  we get `possibly lost` is because we are storing a pointer to the middle of malloc-ed block in the second arg of invocation `append_link(&list, &p->link, p);` which gets stored as the "links" in the link list. So I think if we erase the link pointers without freeing the memory should make valgrind show them as only `still reachable`. However I'm yet to try this out.

subhasisb:
Its actually fairly easy to create code that will generate possibly leaked - but that is not the question here. About the list_link code, the fact that we store a pointer to the middle should be okay, i cant see where we are losing the pointer to the start of the memory anyway...

Shrini-h:
The catch is the memory pointers are stored back in the block itself as part of `void *ll_struct`, which is not relied by valgrind as valid reference to the malloc block. what we do with `void *ll_struct` is like locking the main door and throwing the key inside the house.

I experimented by changing `append_link()` function to store in the `ll_struct` the pointer to the next malloced obj pointer to make it single linked list of objects, like locking the door and throwing the key in the next house, Now give valgrind the one key to the first house, it will open all other house doors 😄  one by one.

Just replace the last 4 lines of `append_link()` with below 4 to see valgrind report all the blocks as `still reachable`
```
	if (pobj)
		newp->ll_struct = pobj;
	else
		newp->ll_struct = (void *)newp;
```
also use `CLEAR_HEAD(p->link);` instead of `CLEAR_LINK(p->link);` in `allocate()` of [tst.c.txt](https://github.com/PBSPro/pbspro/files/2433259/tst.c.txt)

subhasisb:
Thanks - that sounds reasonable. Yes that's how list_link.c works - but i thought valgrind should figure it all out. Recall we have not thrown off all the keys inside the room. There is definitely the pbs_list_head that we (our code) itself follows to get to any node we want - so the keys all exist - its just that valgrind can't find it anymore. Anyways, your experiment (in changing the code) does prove that valgrind gets confused with our pointer manipulation - and thus we should ignore all "possible leaks" related to our link lists (if we want to fix it in a reasonable way, we need to be super careful)

Shrini-h:
One way to mitigate many of the cases is to keep `pbs_list_link` as the first member of the object structure. Then the links will point to the start of block

iestockdale:
Ya we can do that - but would a whole bunch of structures to change possibly. But we should look into that.

On 01-Oct-2018, at 11:49 PM, Shrinivas Harapanahalli <notifications@github.com<mailto:notifications@github.com>> wrote:


One way to mitigate many of the cases is to keep pbs_list_link as the first member of the object structure. Then the links will point to the start of block

nithinj:
@subhasisb ,
I've re-based my changes. Please take this further.

PR #813, opened 2018-09-20
 Server allows jobs/resvs to be run/confirmed on down/stale vnodes

shivaprasad-bs:
#### Bug/feature Description
When a job is run or a reservation is confirmed, the server is sent a list of nodes.  It does not detect if any of these nodes are down or stale and reject the request.

This can happen in the following cases:

When a node goes down/stale after the scheduler has queried it for the cycle.  The scheduler will think it is available for use.

What does this mean:
If a reservation is confirmed on a down or stale vnode, it will go into the RESV_CONFIRMED state.  It will not be degraded unless there is some other event that will cause it to go degraded.
If a job is run on a down node, the server will attempt to connect to the mom (even though it knows its down) and that will fail.
If a job is run on a stale node, it will successfully run on the mom that is no longer reporting that vnode.

One of the way to Reproduce the issue:
We need to use gdb to reproduce the issue.
1. Attach scheduler to one session of gdb and server to another session of gdb.
2. At scheduler we need to put break point after where the scheduler is querying the stats of server.
3. At Server put a break point inside set_nodes function.
4. run qsub <test script>
5. at scheduler it will enter the break point , kill the mom, so all node state goes down.
6. continue the break point at scheduler to allow it to schedule.
7. it comes back to server, check the steps in set_nodes where job is allocated to vnode which is down. newly added checks will put the job into the Queue for reschedule.

#### Affected Platform(s)
All Platforms supported by PBS Professional is affected.
 
The function that parses the list of nodes given the server when a job runs or a reservation is confirmed is called set_nodes().  This function is a very generic and highly critical function to the server.  It is the function that sets the node pointers on the jobs/reservations.  After the Scheduler schedules the job to a node, call comes back to set_nodes - here if the node is down after the job is scheduled, there is no check in server, before sending request to mom. which is causing the mom being contacted by server to push the job.

#### Solution Description
Added checks to the set_nodes function to check whether the assigned node is down/stale, if it's down/stale function returns PBSE_BAD_NODE_STATE.

#### Checklist:

### DIFF
--- a/src/include/pbs_error.h
+++ b/src/include/pbs_error.h
@@ -259,6 +259,7 @@ extern "C" {
 #define PBSE_STDG_RESV_OCCR_CONFLICT 15179 /* cannot change start time of a non-empty reservation */
 
 #define PBSE_SOFTWT_STF		     15180 /* soft_walltime is incompatible with STF jobs */
+#define PBSE_BAD_NODE_STATE          15181 /* node is in the wrong state for the operation */
 
 /*
  ** 	Resource monitor specific
--- a/src/server/node_manager.c
+++ b/src/server/node_manager.c
@@ -7140,9 +7140,7 @@ set_nodes(void *pobj, int objtype, char *execvnod_in, char **execvnod_out, char 
 					}
 					else {
 						resv_setResvState(presv, RESV_DEGRADED, RESV_DEGRADED);
-						free(phowl);
-						free(execvncopy);
-						return (PBSE_BAD_NODE_STATE);
+						set_resv_retry(presv, time_now+10);
 					}
 				}
                         }
--- a/src/server/node_manager.c
+++ b/src/server/node_manager.c
@@ -7125,6 +7125,11 @@ set_nodes(void *pobj, int objtype, char *execvnod_in, char **execvnod_out, char 
 				return (PBSE_UNKNODE);
 			}
 
+                        if (svr_init == FALSE && (pnode->nd_state & (INUSE_DOWN | INUSE_STALE))) {
+                                   free(phowl);
+                                   free(execvncopy);
+                                   return (PBSE_BAD_NODE_STATE);
+                        }
 
 			if (pjob != NULL) { /* only for jobs do we warn if a mom */
 				/* hook has not been sent */
@@ -8404,8 +8409,10 @@ set_old_subUniverse(resc_resv	*presv)
 		free(sp);
 		return;
 	}
+	int svr_init;
+	svr_init = TRUE;
 	/* set the nodes on the reservation */
-	rc = assign_resv_resc(presv, sp);
+	rc = assign_resv_resc(presv, sp, svr_init);
 	if (rc != PBSE_NONE) {
 		sprintf(log_buffer,
 			"problem assigning resource to reservation %d", rc);
--- a/src/server/node_manager.c
+++ b/src/server/node_manager.c
@@ -7125,10 +7125,23 @@ set_nodes(void *pobj, int objtype, char *execvnod_in, char **execvnod_out, char 
 				return (PBSE_UNKNODE);
 			}
 
-                        if (svr_init == FALSE && (pnode->nd_state & (INUSE_DOWN | INUSE_STALE))) {
+                        if ((pnode->nd_state & (INUSE_DOWN | INUSE_STALE)) && (svr_init == FALSE)) {
+
+				if(objtype == JOB_OBJECT) {
                                    free(phowl);
                                    free(execvncopy);
                                    return (PBSE_BAD_NODE_STATE);
+				} 
+				else {
+					if(presv->ri_qs.ri_state == RESV_UNCONFIRMED) {
+						set_resv_retry(presv, time_now+10);
+						free(phowl);
+                	                	free(execvncopy);
+                        	      		return (PBSE_BAD_NODE_STATE);
+					}
+					else
+						resv_setResvState(presv, RESV_DEGRADED, RESV_DEGRADED);
+				}
                         }
 
 			if (pjob != NULL) { /* only for jobs do we warn if a mom */
--- a/src/server/node_manager.c
+++ b/src/server/node_manager.c
@@ -7134,13 +7134,16 @@ set_nodes(void *pobj, int objtype, char *execvnod_in, char **execvnod_out, char 
 				} 
 				else {
 					if(presv->ri_qs.ri_state == RESV_UNCONFIRMED) {
-						set_resv_retry(presv, time_now+10);
 						free(phowl);
                 	                	free(execvncopy);
                         	      		return (PBSE_BAD_NODE_STATE);
 					}
-					else
+					else {
 						resv_setResvState(presv, RESV_DEGRADED, RESV_DEGRADED);
+						free(phowl);
+						free(execvncopy);
+						return (PBSE_BAD_NODE_STATE);
+					}
 				}
                         }
 
@@ -8422,10 +8425,8 @@ set_old_subUniverse(resc_resv	*presv)
 		free(sp);
 		return;
 	}
-	int svr_init;
-	svr_init = TRUE;
 	/* set the nodes on the reservation */
-	rc = assign_resv_resc(presv, sp, svr_init);
+	rc = assign_resv_resc(presv, sp, TRUE);
 	if (rc != PBSE_NONE) {
 		sprintf(log_buffer,
 			"problem assigning resource to reservation %d", rc);
--- a/src/server/req_rescq.c
+++ b/src/server/req_rescq.c
@@ -312,6 +312,7 @@ req_confirmresv(struct batch_request *preq)
 	int		is_being_altered = 0;
 	char		*tmp_buf = NULL;
 	size_t		tmp_buf_size = 0;
+	int 		svr_init=FALSE;
 
 	if ((preq->rq_perm & (ATR_DFLAG_MGWR | ATR_DFLAG_OPWR)) == 0) {
 		req_reject(PBSE_PERM, 0, preq);
@@ -583,7 +584,6 @@ req_confirmresv(struct batch_request *preq)
 	 */
 	if (is_being_altered)
 		free_resvNodes(presv);
-	int svr_init;
 	svr_init = FALSE;
 	rc = assign_resv_resc(presv, next_execvnode, svr_init);
 
--- a/src/server/req_rescq.c
+++ b/src/server/req_rescq.c
@@ -242,7 +242,7 @@ cnvrt_timer_init()
  * @retval	non-zero	: error code if problem occurs
  */
 int
-assign_resv_resc(resc_resv *presv, char *vnodes)
+assign_resv_resc(resc_resv *presv, char *vnodes, int svr_init)
 {
 	int		  ret;
 	char     *node_str = NULL;
@@ -252,7 +252,7 @@ assign_resv_resc(resc_resv *presv, char *vnodes)
 		return (PBSE_BADNODESPEC);
 
 	ret = set_nodes((void *)presv, presv->ri_qs.ri_type, vnodes,
-		&node_str, &host_str, &host_str2, 0, FALSE);
+		&node_str, &host_str, &host_str2, 0, svr_init);
 
 	if (ret == PBSE_NONE) {
 		/*update resc_resv object's RESV_ATR_resv_nodes attribute*/
@@ -583,7 +583,9 @@ req_confirmresv(struct batch_request *preq)
 	 */
 	if (is_being_altered)
 		free_resvNodes(presv);
-	rc = assign_resv_resc(presv, next_execvnode);
+	int svr_init;
+	svr_init = FALSE;
+	rc = assign_resv_resc(presv, next_execvnode, svr_init);
 
 	if (rc != PBSE_NONE) {
 		free(next_execvnode);
--- a/src/server/req_rescq.c
+++ b/src/server/req_rescq.c
@@ -249,7 +249,7 @@ assign_resv_resc(resc_resv *presv, char *vnodes, int svr_init)
 	char		 *host_str = NULL;	/* used only as arg to set_nodes */
 	char * host_str2 = NULL;
 	struct work_task *pwt;
-        extern void resv_retry_handler(struct work_task *ptask);
+	extern void resv_retry_handler(struct work_task *ptask);
 
 	if ((vnodes == NULL) || (*vnodes == '\0'))
 		return (PBSE_BADNODESPEC);
--- a/src/server/req_rescq.c
+++ b/src/server/req_rescq.c
@@ -248,6 +248,9 @@ assign_resv_resc(resc_resv *presv, char *vnodes, int svr_init)
 	char     *node_str = NULL;
 	char		 *host_str = NULL;	/* used only as arg to set_nodes */
 	char * host_str2 = NULL;
+	struct work_task *pwt;
+        extern void resv_retry_handler(struct work_task *ptask);
+
 	if ((vnodes == NULL) || (*vnodes == '\0'))
 		return (PBSE_BADNODESPEC);
 
@@ -268,6 +271,18 @@ assign_resv_resc(resc_resv *presv, char *vnodes, int svr_init)
 
 		presv->ri_modified = 1;
 	}
+	else if (ret == PBSE_BAD_NODE_STATE) {
+           /* Set a work task to initiate a scheduling cycle when the time to check
+            * for alternate nodes to assign the reservation comes
+            */
+           if ((pwt = set_task(WORK_Timed, time_now+10, resv_retry_handler, presv)) != NULL) {
+           /* set things so that the reservation going away will result in
+            * any "yet to be processed" work tasks also going away
+            */
+ 	          append_link(&presv->ri_svrtask, &pwt->wt_linkobj, pwt);
+           }
+	}
+
 
 	return (ret);
 }
@@ -312,7 +327,6 @@ req_confirmresv(struct batch_request *preq)
 	int		is_being_altered = 0;
 	char		*tmp_buf = NULL;
 	size_t		tmp_buf_size = 0;
-	int 		svr_init=FALSE;
 
 	if ((preq->rq_perm & (ATR_DFLAG_MGWR | ATR_DFLAG_OPWR)) == 0) {
 		req_reject(PBSE_PERM, 0, preq);
@@ -584,8 +598,7 @@ req_confirmresv(struct batch_request *preq)
 	 */
 	if (is_being_altered)
 		free_resvNodes(presv);
-	svr_init = FALSE;
-	rc = assign_resv_resc(presv, next_execvnode, svr_init);
+	rc = assign_resv_resc(presv, next_execvnode, FALSE);
 
 	if (rc != PBSE_NONE) {
 		free(next_execvnode);
--- a/src/server/svr_jobfunc.c
+++ b/src/server/svr_jobfunc.c
@@ -3600,7 +3600,9 @@ Time4occurrenceFinish(resc_resv *presv)
 	/* Assign the allocated resources to the reservation
 	 * and the reservation to the associated vnodes
 	 */
-	rc = assign_resv_resc(presv, newxc);
+	int svr_init;
+	svr_init = FALSE;
+	rc = assign_resv_resc(presv, newxc, svr_init);
 	free(newxc);
 
 	if (rc != PBSE_NONE) {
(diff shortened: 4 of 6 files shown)


bhroam:
I see a problem with the code.  Standing reservations move from occurrence to occurrence when one finishes.  If a vnode goes down (or become stale) between when the reservation is confirmed and when the occurrence is the next occurrence, we'll call set_nodes() with down/stale nodes.  The new change will reject that change and the reservation be left in an unstable state.  The reservation will lose its vnodes for the next occurrence.  
Right now we at least set the vnodes (since we ignore state), but we don't degrade the reservation.  
The set_nodes() function is an extremely critical function and some parts of the code don't handle when it fails well.

How about this as a solution:
Handle reservations differently than jobs.  The current code works fine for jobs.  Other than server restart, all calls to set_nodes() for jobs are from requests.  We can reject those.  For reservations, we detect but do not reject.  When we detect the down/stale situation, we correctly reallocate all the nodes to the reservation, but immediately degrade it.  The scheduler will reconfirm it on other nodes later.

Another idea is slightly different.  We can reject a reservation confirm, but on other calls to set_nodes() we detect and degrade.

What do you think?

src/server/req_rescq.c
```diff
@@ -242,7 +242,7 @@ cnvrt_timer_init()
  * @retval	non-zero	: error code if problem occurs
  */
 int
-assign_resv_resc(resc_resv *presv, char *vnodes)
+assign_resv_resc(resc_resv *presv, char *vnodes, int svr_init)
```
suresh-thelkar: Update the function header with the new parameter svr_init along with its purpose.

src/server/req_rescq.c
```diff
@@ -583,7 +583,9 @@ req_confirmresv(struct batch_request *preq)
 	 */
 	if (is_being_altered)
 		free_resvNodes(presv);
-	rc = assign_resv_resc(presv, next_execvnode);
+	int svr_init;
+	svr_init = FALSE;
+	rc = assign_resv_resc(presv, next_execvnode, svr_init);
```
suresh-thelkar: You do not need extra variable here. Directly pass the desired value to the function assign_resv_resc.

src/server/req_rescq.c
```diff
@@ -583,7 +583,9 @@ req_confirmresv(struct batch_request *preq)
 	 */
 	if (is_being_altered)
 		free_resvNodes(presv);
-	rc = assign_resv_resc(presv, next_execvnode);
+	int svr_init;
+	svr_init = FALSE;
```
suresh-thelkar: Also declaring variables in this portion of the code will fail Windows Build. It looks like this is one of the reason appveyor is failing the windows build. Please look into the Windows build and also test it on Windows.

src/include/pbs_error.h
```diff
@@ -259,6 +259,7 @@ extern "C" {
 #define PBSE_STDG_RESV_OCCR_CONFLICT 15179 /* cannot change start time of a non-empty reservation */
 
 #define PBSE_SOFTWT_STF		     15180 /* soft_walltime is incompatible with STF jobs */
+#define PBSE_BAD_NODE_STATE          15181 /* node is in the wrong state for the operation */
```
suresh-thelkar: If possible try to use the existing return codes like PBSE_NODESTALE etc.

src/server/req_rescq.c
shivaprasad-bs: > I see a problem with the code. Standing reservations move from occurrence to occurrence when one finishes. If a vnode goes down (or become stale) between when the reservation is confirmed and when the occurrence is the next occurrence, we'll call set_nodes() with down/stale nodes. The new change will reject that change and the reservation be left in an unstable state. The reservation will lose its vnodes for the next occurrence.
> Right now we at least set the vnodes (since we ignore state), but we don't degrade the reservation.
> The set_nodes() function is an extremely critical function and some parts of the code don't handle when it fails well.
> 
> How about this as a solution:
> Handle reservations differently than jobs. The current code works fine for jobs. Other than server restart, all calls to set_nodes() for jobs are from requests. We can reject those. For reservations, we detect but do not reject. When we detect the down/stale situation, we correctly reallocate all the nodes to the reservation, but immediately degrade it. The scheduler will reconfirm it on other nodes later.
> 
> Another idea is slightly different. We can reject a reservation confirm, but on other calls to set_nodes() we detect and degrade.
> 
> What do you think?

Thanks Bhroam - Made changes for code to take care of Normal reservation and Standing reservation as well.
In case of normal reservation added retry after 10 seconds, as the nodes may be available - if the call has come to set nodes.
Standing reservation - I've set it for degraded, which will be considered for next scheduling cycle.

please check the changes and let me know the comments.

bhroam:
It looks like you did a git merge from master instead of a rebase.  This will be bad since when you squash, all these commits will look like you made the changes.  You need to do a git rebase --hard to the point before the merge and then rebase from master.

src/server/node_manager.c
```diff
@@ -7114,11 +7114,27 @@ set_nodes(void *pobj, int objtype, char *execvnod_in, char **execvnod_out, char
...
-                        if (svr_init == FALSE && (pnode->nd_state & (INUSE_DOWN | INUSE_STALE))) {
-                                   free(phowl);
-                                   free(execvncopy);
-                                   return (PBSE_BAD_NODE_STATE);
-                        }
+                	if ((pnode->nd_state & (INUSE_DOWN | INUSE_STALE))&& (svr_init == FALSE)) {
+                            if (objtype == JOB_OBJECT) {
+				free(phowl);
+                                free(execvncopy);
+                                return (PBSE_BAD_NODE_STATE);
```
bhroam: I think these two lines are indented further than they should be.  Either that or the first free() is not indented enough.

src/server/node_manager.c
```diff
@@ -7114,11 +7114,27 @@ set_nodes(void *pobj, int objtype, char *execvnod_in, char **execvnod_out, char
...
+				resv_setResvState(presv, RESV_DEGRADED, RESV_DEGRADED);
+			    }
+			    else {
+				int est_retryTime;
+				est_retryTime = time_now+ 10; /* Time for Normal reservation to retry, value in Seconds */
+				
+				set_resv_retry(presv, est_retryTime);
+				free(phowl);
+                                free(execvncopy);
+				return (PBSE_BAD_NODE_STATE);
```
bhroam: This is not right.  If you reject the set_nodes() call, the confirm will fail.  This means it will attempt to be reconfirmed when it is not degraded.  The more I think about this, I think we should treat all reservations the same.  Just mark them degraded and let the degraded reservation machinery handle reconfirming it.

bhroam:
@shivaprasad-bs until you fix your branch, anything you push will be lost after the git reset.  Please fix your branch before continuing work.

src/server/node_manager.c
```diff
@@ -7119,6 +7119,24 @@ set_nodes(void *pobj, int objtype, char *execvnod_in, char **execvnod_out, char
...
+                        if ((pnode->nd_state & (INUSE_DOWN | INUSE_STALE)) && (svr_init == FALSE)) {
+
+				if(objtype == JOB_OBJECT) {
+                                   free(phowl);
+                                   free(execvncopy);
+                                   return (PBSE_BAD_NODE_STATE);
+				} 
+				else {
+					if(presv->ri_qs.ri_state == RESV_UNCONFIRMED) {
+						set_resv_retry(presv, time_now+10);
```
bhroam: This is unneeded and has bad side effects.  This function was meant for reservations in the degraded state.  It sets the retry time.  That isn't supposed to be set until the resv is degraded.  The scheduler will take a rejection as a retry.  It will immediately stop its cycle and start a new one.  If a node has gone down or stale on the scheduler since it queried the world, the new cycle will see that.

src/server/node_manager.c
```diff
@@ -7119,6 +7119,24 @@ set_nodes(void *pobj, int objtype, char *execvnod_in, char **execvnod_out, char
...
+						free(phowl);
+                	                	free(execvncopy);
+                        	      		return (PBSE_BAD_NODE_STATE);
+					}
+					else
+						resv_setResvState(presv, RESV_DEGRADED, RESV_DEGRADED);
```
bhroam: Speaking of reservations in the degraded state, look at the vnode_unavailable() function.  That is the function that usually marks reservations as degraded.  It sets the retry time.  I think just setting the state will mark the reservation in state degraded, but the scheduler will never reconfirm it.

src/server/svr_jobfunc.c
```diff
@@ -3595,7 +3596,8 @@ Time4occurrenceFinish(resc_resv *presv)
 	/* Assign the allocated resources to the reservation
 	 * and the reservation to the associated vnodes
 	 */
-	rc = assign_resv_resc(presv, newxc);
+	svr_init = FALSE;
```
bhroam: Same here, no need for the svr_init variable.  Just pass FALSE.

src/server/node_manager.c
shivaprasad-bs: If the Reservation is still unconfirmed - is it Ok to be degraded with out nodes assigned to it and if we make it degraded, it will stay as degraded till node is available for Standing reservation. if the reservaion is unconfirmed even we make standing reservation is degraded it will not be rescheduled.

In case of normal reservation, even if it is degraded - it will not be considered for reschedule or retry resv in both the cases (unconfirmed or state moved from confirmed->degraded), 
even when nodes are available.
would you think directly using set_resv_retry call will impact scheduling cycle?

Standing Reservation:
if degraded, it will be put back to resv retry.
if unconfirmed, it will be rejected, so I've manually used set_resv_retry to put it back for retry.
In case of Standing resv, setting the main state to degraded will mean to put it for rescheduling/reallocating the nodes.

Normal Reservation:
if degraded, the nodes will not be reallocated with new ones.
if unconfirmed, the current changes will put the resv to try again (set_resv_retry), if the scheduled node is down/stale.
In case of normal reservation, setting the reservation state to degraded, it will not be set to resv retry.

src/server/node_manager.c
shivaprasad-bs: For Standing reservation, if state is set to degraded it will be reconfirmed by scheduler by setting retry time.

bhroam: Yes, that is my point.  You are not setting the retry time.  You are just setting the state.

src/server/node_manager.c
bhroam: The issue is with calling set_resv_retry().  It does more than just sets a task on the list to retry the reservation.  It also sets the internal attribute/member for retrying.  If the reservation goes degraded later, these are supposed to be unset.  If you want to just kick the scheduler to retry, set a work task for it (look at set_resv_retry() for how to do it).  Now you do need to call set_resv_retry for when you set the reservation in the degraded state on the other side of the if.

src/server/node_manager.c
```diff
@@ -7128,13 +7128,16 @@ set_nodes(void *pobj, int objtype, char *execvnod_in, char **execvnod_out, char
...
                 	                	free(execvncopy);
                         	      		return (PBSE_BAD_NODE_STATE);
 					}
-					else
+					else {
 						resv_setResvState(presv, RESV_DEGRADED, RESV_DEGRADED);
+						free(phowl);
+						free(execvncopy);
+						return (PBSE_BAD_NODE_STATE);
+					}
```
bhroam: You can't reject reservations that are not in the unconfirmed state.  If you do then when one occurrence of a standing reservation ends and the next occurrence's nodes are down/stale, the reservation won't get nodes attached to it.  You need to mark the reservation degraded and call set_resv_retry() appropriately.

src/server/req_rescq.c
```diff
@@ -248,6 +248,9 @@ assign_resv_resc(resc_resv *presv, char *vnodes, int svr_init)
 	char     *node_str = NULL;
 	char		 *host_str = NULL;	/* used only as arg to set_nodes */
 	char * host_str2 = NULL;
+	struct work_task *pwt;
+        extern void resv_retry_handler(struct work_task *ptask);
```
bhroam: I think the indentation is off.

test/tests/pbs_smoketest.py
```diff
@@ -118,6 +118,47 @@ def test_standing_reservation(self):
...
+        else:
+            self.logger.info('Missing timezone, using America/Los_Angeles')
+            tzone = 'America/Los_Angeles'
+       	a = {'resources_available.ncpus': 1}
+       	self.server.create_vnodes('vn', a, num=4, mom=self.mom)
+       	a = {'Resource_List.select': '1:ncpus=2',
+             ATTR_resv_rrule: 'FREQ=HOURLY;COUNT=2',
+             ATTR_resv_timezone: tzone,
+             ATTR_resv_standing: '1',
+             'reserve_start': time.time() + 4,
```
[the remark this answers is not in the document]
bhroam: also say time_now = int(time.time()) and use time_now for reserve_start and reserve_end.  Using time.time(), you will end up with a floating point number.

test/tests/pbs_smoketest.py
```diff
@@ -118,6 +118,47 @@ def test_standing_reservation(self):
...
+             'reserve_end': time.time() + 9, }
+       	r = Reservation(TEST_USER, a)
+        rid = self.server.submit(r)
+        a = {'reserve_state': (MATCH_RE, "RESV_CONFIRMED|2")}
+        a2 = {'reserve_state': (MATCH_RE, "RESV_DEGRADED|2")}
```
bhroam: 2 is confirmed, use the correct number for degraded.

test/tests/pbs_smoketest.py
```diff
@@ -118,6 +118,47 @@ def test_standing_reservation(self):
...
+        self.server.expect(RESV, a, id=rid)
+        self.server.status(RESV, 'resv_nodes', id=rid)
+        resv_node = self.server.reservations[rid].get_vnodes()[0]
+        b = {'state': 'down'}
+        self.server.manager(MGR_CMD_SET, NODE, b, id=resv_node)
+        self.server.expect(RESV, a2, id=rid)
+	time.sleep(13)
```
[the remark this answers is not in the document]
bhroam: You might also add a self.logger.info() to inform the test runner what the sleep is for.

test/tests/pbs_smoketest.py
```diff
@@ -118,6 +118,47 @@ def test_standing_reservation(self):
...
+        self.server.expect(RESV, a, id=rid)
```
bhroam: Since you are using expect(), consider using an offset to expect() instead of time.sleep()

bhroam:
Travis failed due to pep8 issues.

test/tests/pbs_smoketest.py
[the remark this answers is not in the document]
shivaprasad-bs: Can you please elaborate on this point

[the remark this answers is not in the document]
anamikau: currently test is only checking if reservation has been confirmed in the next cycle. but it is not checking if it is being confirmed on a different node than the one which is marked as down.

test/tests/pbs_smoketest.py
```diff
@@ -135,27 +138,35 @@ def test_standing_reservation_stale_nodes(self):
...
-             'reserve_end': time.time() + 9, }
-       	r = Reservation(TEST_USER, a)
+             'reserve_start': time_now + 8,
+             'reserve_end': time_now + 14, 
+             'Resource_List.place': 'free'}
+        r = Reservation(TEST_USER, a)
+        #Submitting recurring reservation to check the resv status for one of the vnode being stale
         rid = self.server.submit(r)
+        #To check the standing resv once confirmed.
         a = {'reserve_state': (MATCH_RE, "RESV_CONFIRMED|2")}
```
anamikau: I still think it is a good idea to check for confirmed or running state both

test/tests/pbs_smoketest.py
```diff
@@ -135,27 +138,35 @@ def test_standing_reservation_stale_nodes(self):
...
-        a2 = {'reserve_state': (MATCH_RE, "RESV_DEGRADED|2")}
+        #To check the standing resv for Degraded status when one of the vnode becomes down
+        a2 = {'reserve_state': (MATCH_RE, "RESV_DEGRADED|10")}
         self.server.expect(RESV, a, id=rid)
         self.server.status(RESV, 'resv_nodes', id=rid)
         resv_node = self.server.reservations[rid].get_vnodes()[0]
+        #To check the status of standing resv in next cycle and also check the down node doesn't exist in resv_nodes.
```
anamikau: rephrase it as "Check that standing reservation gets confirmed in next cycle to a different node than the one which is down"

test/tests/pbs_smoketest.py
```diff
@@ -112,12 +112,15 @@ def test_standing_reservation(self):
              'reserve_start': time.time() + 20,
              'reserve_end': time.time() + 30, }
         r = Reservation(TEST_USER, a)
+        nodes = r.get_vnodes(r.exec_vnode)
+        print "NODES:"
+        print nodes
```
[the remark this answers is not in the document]
anamikau: Also are you sure this change is at the right place. I mean why are you getting the vnodes before submitting the reservation.

test/tests/pbs_smoketest.py
```diff
@@ -112,22 +112,20 @@ def test_standing_reservation(self):
...
         self.server.expect(RESV, a, id=rid)
         if _m == PTL_API:
             self.server.set_op_mode(PTL_API)
+
     @skipOnCray
     def test_standing_reservation_stale_nodes(self):
         """
-        Test to submit a standing reservation for nodes which are stale after scheduling cycle and before set_nodes in server.
+        Test to submit a standing reservation for nodes which are stale
+	after scheduling cycle and before set_nodes in server.
```
bhroam: Looks like the indent is off here.

test/tests/pbs_smoketest.py
```diff
@@ -150,21 +147,27 @@ def test_standing_reservation_stale_nodes(self):
              'reserve_end': time_now + 14, 
              'Resource_List.place': 'free'}
         r = Reservation(TEST_USER, a)
-        #Submitting recurring reservation to check the resv status for one of the vnode being stale
+        #Submitting recurring reservation to check the resv status 
+        #for one of the vnode being stale
```
bhroam: Suggestion: add spaces after the #.  It makes it easier to read.

test/tests/pbs_smoketest.py
```diff
@@ -150,21 +147,27 @@ def test_standing_reservation_stale_nodes(self):
...
+        #check the down node doesn't exist in resv_nodes.
+        a = {'reserve_state': (MATCH_RE, "RESV_CONFIRMED|2"),
+             'resv_nodes': 
+             (MATCH_RE, '^((?!' + re.escape(resv_node) + ').)*$')}
         b = {'state': 'down'}
         self.server.manager(MGR_CMD_SET, NODE, b, id=resv_node)
         self.server.expect(RESV, a2, id=rid)
-        self.logger.info('testinfo: waiting for recurring reservation, while expecting status')
+        self.logger.info('testinfo: waiting for recurring reservation,' + 
+                                                'while expecting status')
```
bhroam: No need for the '+'.  Two string literals next to each other will be concatenated together.

test/tests/pbs_smoketest.py
```diff
@@ -147,20 +147,20 @@ def test_standing_reservation_stale_nodes(self):
...
         a2 = {'reserve_state': (MATCH_RE, "RESV_DEGRADED|10")}
         self.server.expect(RESV, a, id=rid)
         self.server.status(RESV, 'resv_nodes', id=rid)
         resv_node = self.server.reservations[rid].get_vnodes()[0]
-        #To check the status of standing resv in next cycle and also
-        #check the down node doesn't exist in resv_nodes.
-        a = {'reserve_state': (MATCH_RE, "RESV_CONFIRMED|2"),
+        # To check the status of standing resv in next cycle and also
+        # check the down node doesn't exist in resv_nodes.
+        a = {'reserve_state': (MATCH_RE, "RESV_CONFIRMED|2|RESV_RUNNING|5"),
```
bhroam: You should know the state you are looking for.  Why are you checking for confirmed or running?  When you pass this to expect(), you'll need to pass attrop=PTL_AND to check for both of these.

shivaprasad-bs: I don't think we need check for both RUNNING and CONFIRMED, just CONFIRMED is enough.  Looks like Anamika has suggested to check if it's in race condition, when test case is waiting and it skipped the 8 seconds time.

anamikau: Hi Bhroam, yes I suggested to look for both since reservation start time is 8s from now and in the past we have observed related race conditions where reservation has started running by the time we are looking for reservation state. Also it should be PTL_OR as state will be one of them.

test/tests/pbs_smoketest.py
anamikau: please update this instance as well with resv_confirmed or resv_running

test/tests/pbs_smoketest.py
bhroam: Oh, OK.  It's OK then.  I don't think it should be PTL_OR.  That is an or between attributes.  He has correctly done the or as an regex.  It needs to be PTL_AND because there are two attributes he's expecting.  I don't think the regex he has will work for matching the resv_nodes though.

anamikau: I see. Sorry got confused about PTL_AND. you are right about it.

anamikau:
@shivaprasad-bs please attach the latest execution results for smoketest.

agrawalravi90:
No activity on this for 3+ months so I'm closing this PR

PR #836, opened 2018-10-08
Sched skips marking node for provisioning

arungrover:
#### Bug/feature Description
While trying to find a node solution for a job requesting provisioning scheduler could in some cases (with place scatter or vscatter) modify the current aoe on the node. If this job could not run because of some reason  (like resource crunch) then the scheduler might end up producing a wrong node solution for the next provisioning job.

#### Affected Platform(s)
All

The problem is that scheduler sets the current aoe on the node as soon as it finds provisioning is needed for a job. When the job in hand fails to run, the current aoe set on the node persists and then the next provisioning job's solution gets affected (because the node already looked as provisioned to the next job).

#### Solution Description
Do not set current aoe when placement is scatter/vscatter because these are the two placement specification where the nodes are not duplicated while finding a node solution for a job. In all other cases it does not matter because modifying duplicate set of nodes does not affect the next job scheduler looks at.

#### Testing logs/output
[test_after_fix.txt](https://github.com/PBSPro/pbspro/files/2458211/test_after_fix.txt)
[test_before_fix.txt](https://github.com/PBSPro/pbspro/files/2458214/test_before_fix.txt)

#### Checklist:


agrawalravi90:
"Do not set current aoe when placement is scatter/vscatter " .. The answer to this is probably no, but is it possible to fix this issue generically so that we don't have to write special case code?

test/tests/functional/pbs_provisioning_enhancement.py
```diff
@@ -63,15 +63,32 @@ class TestProvisioningJob_Enh(TestFunctional):
...
+        if len(self.moms.values()) < 2:
+            self.skipTest("Provide at least 2 moms while invoking test")
+
         TestFunctional.setUp(self)
         self.momA = self.moms.values()[0]
         self.momB = self.moms.values()[1]
         self.hostA = self.momA.shortname
         self.hostB = self.momB.shortname
 
+        if self.hostA == self.server.shortname:
```
agrawalravi90: I'm not sure i understand this, can you please add a comment to explain what's happening? Also, you could probably check for self.server.shortname before the first time you initialize these variables itself

test/tests/functional/pbs_provisioning_enhancement.py
```diff
@@ -338,3 +355,86 @@ def test_provisioning_with_placement(self):
...
+                            attrib={'current_aoe': 'osimage1'}, expect=True)
+
+        # submit one job that will run on local node
+        a = {'Resource_List.select': '1:ncpus=10'}
+        j1 = Job(TEST_USER1, attrs=a)
+        j1.set_sleep_time(200)
+        jid1 = self.server.submit(j1)
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid1)
+
+        # Turn off scheduling
```
agrawalravi90: Why are you turning scheduling off?

arungrover: That's because I want next two jobs to be considered in the same cycle

test/tests/functional/pbs_provisioning_enhancement.py
arungrover: I'll add a comment

arungrover:
@agrawalravi90 I've addressed your review comments

test/tests/functional/pbs_provisioning_enhancement.py
arungrover: I made this change because the test sets up aoe on momA that means it expects momA to be a remote mom. So I added this code to make sure that even if the first mom is local mom, the test swaps them and uses the correct mom for provisioning.

src/scheduler/node_info.c
```diff
@@ -3596,7 +3596,7 @@ resources_avail_on_vnode(resource_req *specreq_cons, node_info *node,
 					} else if (is_p == PROVISIONING_NEEDED ) {
 						if (ns != NULL)
 							ns->go_provision = 1;
-						if (resresv->select->total_chunks >1)
+						if (resresv->select->total_chunks > 1 && pl->scatter != 1 && pl->vscatter != 1)
```
bhroam: The eoe code was fashioned off of the aoe code.  It has the same bug in it.

test/tests/functional/pbs_provisioning_enhancement.py
```diff
@@ -63,15 +63,35 @@ class TestProvisioningJob_Enh(TestFunctional):
...
+import pbs
+e = pbs.event()
+j = e.job
+pbs.logmsg(pbs.LOG_DEBUG, "job " + str(j) + " solution " + str(j.exec_vnode))
+e.reject()
 """
 
     def setUp(self):
+
+        if len(self.moms.values()) < 2:
```
bhroam: You can pass a dictionary to len().  No need to say self.moms.values().

arungrover:
> "Do not set current aoe when placement is scatter/vscatter " .. The answer to this is probably no, but is it possible to fix this issue generically so that we don't have to write special case code?

Well, @bhroam can correct me if I'm wrong. For scatter/vscatter we do not create a duplicate copy of nodes while finding the node solution. This means setting anything on nodes would result into a change in the universe scheduler queried. I think not creating a duplicate copy was done as part of an optimization but I'm not certain.

agrawalravi90:
> Well, @bhroam can correct me if I'm wrong. For scatter/vscatter we do not create a duplicate copy of nodes while finding the node solution. This means setting anything on nodes would result into a change in the universe scheduler queried. I think not creating a duplicate copy was done as part of an optimization but I'm not certain.

Ok, I'll let Bhroam reply, I've approved your changes since this might end up being refactoring, but I'd really like us to have a good reason to write special case code for only scatter/vscatter, and possibly mention this reason as a comment in the code so we don't have to revisit this.

bhroam:
Look at eval_complex_selspec() and the free placement section of eval_placement().  It will answer your questions.  It is where the duping of nodes happens.  Maybe a code comment is in order?

arungrover:
Thanks for reviewing the code @agrawalravi90 and @bhroam, I've rebased and squashed the commits.

arungrover:
Codacy is stuck in its analysis. I'm going to close and open the PR again

arungrover:
@agrawalravi90, @bhroam I've added the comment

agrawalravi90:
still looks good to me, thanks

bhroam:
I merged in the PR.

arungrover:
Thanks @bhroam

PR #839, opened 2018-10-12
Scheduler requeue/checkpoint fails when RRTROS is set

agrawalravi90:
#### Bug/feature Description
* When RRTROS is set and preemption method is checkpoint or requeue, scheduler fails to preempt jobs

#### Affected Platform(s)
* All

* Inside find_jobs_to_preempt, we were calling update_universe_in_end() with job state set to 'S'. This was not an issue before RRTROS as all preemption methods free'd up all resources. But post RRTROS, the 'S' method could release limited resources. So, if the preemption method is set to 'R' or 'C' and RRTROS is set, we were releasing limited resources during preemption simulation, and we would end up rejecting jobs as preemption candidates, even though they might be perfect.

#### Solution Description
* Now using preempt order, the code passes the appropriate job_state to update_universe_on_end()
* Also added a new internal flag to job_info called is_preempt_candidate which is checked for inside update_resresv_on_end before it deallocates resources from the job object itself. The resources on the job (like ninfo_arr) are needed later in preemption_similarity.

#### Testing logs/output
Before:
[before_checkpoint_rel_all.log](https://github.com/PBSPro/pbspro/files/2475019/before_checkpoint_rel_all.log)
[before_requeue_rel_all.log](https://github.com/PBSPro/pbspro/files/2475020/before_requeue_rel_all.log)

After:
[testrequeuerelease.log](https://github.com/PBSPro/pbspro/files/2548389/testrequeuerelease.log)
[testcheckpointrelease.log](https://github.com/PBSPro/pbspro/files/2548390/testcheckpointrelease.log)

RRTROS testsuite logs (25 tests): 
[testrrtros.log](https://github.com/PBSPro/pbspro/files/2548399/testrrtros.log)

#### Checklist:

### DIFF
--- a/src/scheduler/job_info.c
+++ b/src/scheduler/job_info.c
@@ -3286,6 +3286,8 @@ find_jobs_to_preempt(status *policy, resource_resv *hjob, server_info *sinfo, in
 
 	skipto=0;
 	while ((indexfound = select_index_to_preempt(npolicy, nhjob, rjobs_subset, skipto, err, fail_list)) != NO_JOB_FOUND) {
+		struct preempt_ordering *po;
+
 		if (indexfound == ERR_IN_SELECT) {
 			/* System error occurred, no need to proceed */
 			free_server(nsinfo, 1);
@@ -3300,8 +3302,13 @@ find_jobs_to_preempt(status *policy, resource_resv *hjob, server_info *sinfo, in
 		schdlog(PBSEVENT_DEBUG2, PBS_EVENTCLASS_JOB, LOG_DEBUG, pjob->name,
 			"Simulation: preempting job");
 
-		pjob->job->resreleased = create_res_released_array(npolicy, pjob);
-		pjob->job->resreq_rel = create_resreq_rel_list(npolicy, pjob);
+		po = get_preemption_order(pjob, sinfo);
+		if (po != NULL) {
+			if (po->order[0] == PREEMPT_METHOD_SUSPEND && pjob->job->can_suspend) {
+				pjob->job->resreleased = create_res_released_array(npolicy,pjob);
+				pjob->job->resreq_rel = create_resreq_rel_list(npolicy, pjob);
+			}
+		}
 
 		update_universe_on_end(npolicy, pjob,  "S", NO_ALLPART);
 		rjobs_count--;
--- a/test/tests/functional/pbs_release_limited_res_suspend.py
+++ b/test/tests/functional/pbs_release_limited_res_suspend.py
@@ -929,3 +929,68 @@ class TestReleaseLimitedResOnSuspend(TestFunctional):
         self.scheduler.log_match(
             jid1 + ";Can't find start time estimate", existence=False,
             max_attempts=2)
+
+    def helper_test_preempt_release_all(self, preempt_method):
+        """
+        Helper function to test that when preempting jobs, all resources
+        are released during preemption simulation for R and C methods
+        """
+        if preempt_method == "R":
+            schedlog_msg = "Job preempted by requeuing"
+        elif preempt_method == "C":
+            schedlog_msg = "Job preempted by checkpointing"
+        else:
+            raise Exception("Unexpected value of argument preempt_method: %s"
+                            % (preempt_method))
+
+        a = {ATTR_restrict_res_to_release_on_suspend: 'mem'}
+        self.server.manager(MGR_CMD_SET, SERVER, a)
+
+        self.scheduler.set_sched_config({'preempt_order': preempt_method})
+
+        # Set 1gb mem available on the node
+        a = {ATTR_rescavail + '.ncpus': "2"}
+        self.server.manager(MGR_CMD_SET, NODE, a, self.mom.shortname)
+
+        # Submit a low priority jobs which takes up all of the ncpus
+        j1 = Job(TEST_USER)
+        j1.set_attributes({ATTR_l + '.select': '1:ncpus=2'})
+        jid1 = self.server.submit(j1)
+        self.server.expect(JOB, {ATTR_state: 'R'}, id=jid1)
+
+        # Submit a high priority job which requests 1 ncpus
+        j2 = Job(TEST_USER)
+        j2.set_attributes({ATTR_l + '.select': '1:ncpus=1',
+                           ATTR_q: 'expressq'})
+        jid2 = self.server.submit(j2)
+
+        # Even though server is configured to only release mem for suspend,
+        # for requeue and checkpointing, we should have released ncpus as well
+        # and correctly preempted the low priority job
+        self.server.expect(JOB, {ATTR_state: 'R'}, id=jid2)
+        self.scheduler.log_match(jid1 + ";" + schedlog_msg)
+
+    def test_preempt_requeue_release_all(self):
+        """
+        Test that when preempting jobs via Requeue, all resources
+        are release during the preemption simulation
+        """
+        self.helper_test_preempt_release_all("R")
+
+    def test_preempt_checkpoint_release_all(self):
+        """
+        Test that when preempting jobs via Checkpointing, all resources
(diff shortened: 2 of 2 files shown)


test/tests/functional/pbs_release_limited_res_suspend.py
```diff
@@ -929,3 +929,68 @@ def test_suspended_job_gets_calendered(self):
...
+            raise Exception("Unexpected value of argument preempt_method: %s"
+                            % (preempt_method))
+
+        a = {ATTR_restrict_res_to_release_on_suspend: 'mem'}
+        self.server.manager(MGR_CMD_SET, SERVER, a)
+
+        self.scheduler.set_sched_config({'preempt_order': preempt_method})
+
+        # Set 1gb mem available on the node
+        a = {ATTR_rescavail + '.ncpus': "2"}
```
arungrover: You are not setting memory on the node

src/scheduler/job_info.c
```diff
@@ -2791,6 +2791,7 @@ preempt_job(status *policy, int pbs_sd, resource_resv *pjob, server_info *sinfo)
 	if (!pjob->job->is_running || pjob->ninfo_arr == NULL)
 		return 0;
 
+	pjob->job->is_preempt_candidate = 0;
```
arungrover: Please add a comment why this flag has to be 0 here.

test/tests/functional/pbs_release_limited_res_suspend.py
```diff
@@ -929,3 +929,68 @@ def test_suspended_job_gets_calendered(self):
...
+        self.server.manager(MGR_CMD_SET, NODE, a, self.mom.shortname)
+
+        # Submit a low priority jobs which takes up all of the ncpus
+        j1 = Job(TEST_USER)
+        j1.set_attributes({ATTR_l + '.select': '1:ncpus=2'})
```
arungrover: ATTR_restrict_res_to_release_on_suspend is set to mem in the test but the job is not requesting it, is it intentional?

test/tests/functional/pbs_release_limited_res_suspend.py
agrawalravi90: That's fine, it doesn't matter since the jobs don't request memory

test/tests/functional/pbs_release_limited_res_suspend.py
agrawalravi90: Yes it is. The idea is to create a situation where PBS will NOT release anything other than mem. So, when the low priority job requests ncpus, PBS will not release ncpus from the nodes for suspend, but it should release them for checkpoint and requeue

agrawalravi90:
I've addressed your comment Arun, please let me know if things look okay now. Thanks!

agrawalravi90:
Thanks for your feedback Bhroam. I'm afraid of doing what you suggest because I think it's too clever. The function update_universe_on_end() assumes that it is updating the universe for a job suspend, without being aware of the special way in which find_jobs_to_preempt() is calling it. I'm worried about somebody modifying update_universe_on_end() in the future in a way that makes perfect sense for update_universe_on_end(), but breaks the special relationship that it has with find_jobs_to_preempt(). So, I'd prefer fixing it in a more reliable, easy to understand way, if possible.

I decided to add metadata on the job object because there seemed to be other similar metadata on the object which don't exactly match with the job states that pbspro supports: there's "can_not_preempt", "is_preempted", "is_susp_sched" etc., all of which seem like metadata that help us make decisions on the job. So, why do you dislike a flag that says that the job is a preemption candidate?

agrawalravi90:
>>  I'm worried about somebody modifying update_universe_on_end() in the future in a way that makes perfect sense for update_universe_on_end(), but breaks the special relationship that it has with find_jobs_to_preempt(). 

Actually, now that I think about it, the reason why the bug exist is exactly this, we modified update_universe_on_end() for RRTROS in a way which made perfect sense for the function but broke the way find_jobs_to_preempt() used it.

bhroam:
I disagree on your reasoning.  This bug exists because when rrtros was implemented, only suspension was considered during implementation.  If all preemption methods were considered, then this bug would have been caught back then.

None of the members you said are metadata.  They either come directly from the server as attributes, or are set at query time.  They are not used as a flag to tell certain functions how to act.  Your piece of metadata is not a global piece of metadata about the job.  is_susp_sched says the job has been preempted by the scheduler (comes from substate=45).  No matter what version of this job we are looking at, it is true.  The version we query in query_jobs(), the dup'd version in calendaring, or the dup'd version in preemption, it is always true.  Your member is only true for the dup'd copy in preemption when trying to preempt for a specific high priority job.

I like leaving the job "suspended" because we need to preemption candidates need to act more like suspended jobs than queued jobs.  preempt_similarity() needs the ninfo_arr.  Why write all this special case code if the jobs can remain suspended.

agrawalravi90:
Thanks for clarifying Bhroam, I didn't know that all of the members correspond to global states, in that case yes, I agree with you, my solution will introduce a piece of metadata just to pass data around. But I still don't like the idea of passing "S" for all preemption methods, let me try to explain:

>>I disagree on your reasoning. This bug exists because when rrtros was implemented, only suspension was considered during implementation. If all preemption methods were considered, then this bug would have been caught back then.

What I meant to say was that update_universe_on_end() as an interface should be agnostic to how it is being called. So, just like RRTROS, if we add a feature which makes suspend special and it requires changes to update_universe_on_end(), one should be able to reliably say if (job_state == "S") inside update_universe_on_end() and do the special thing for suspend. By passing in "S" for requeue and checkpoint as well, i fear we will create opportunities for bugs.

bhroam:
The code in update_universe_on_end() is really doing the right thing.  It should only release resources differently on suspend when rrtors is set.  rrtros usually isn't set, so releasing all resources is the right thing to do.  No matter what we do, we're going to have to trick update_universe_on_end() to do the wrong thing.  It is the wrong thing to leave the nspec_arr/ninfo_arr on the job if you requeue a job.  You either have to add more special case code into update_universe_on_end() or trick it to do what we need to trick it to do what we need it to do.  Since nothing we do is the right answer, I say we take the easiest approach.

agrawalravi90:
Alright, you've convinced me that there's no elegant way to fix this, barring a major refactoring of the code, so I'll go with the approach you suggested.

agrawalravi90:
@arungrover and @bhroam , I've changed the approach a bit according to Bhroam's suggestion, please review the changes again, thanks for your patience.

src/scheduler/resource_resv.c
```diff
@@ -1292,7 +1292,8 @@ update_resresv_on_end(resource_resv *resresv, char *job_state)
 		resresv->job->is_provisioning = 0;
 
 		/* free resources allocated to job since it's now been requeued */
-		if (resresv->job->is_queued && !resresv->job->is_checkpointed) {
+		if (resresv->job->is_queued
+				&& !resresv->job->is_checkpointed) {
```
bhroam: I'd merge these back to one line.

agrawalravi90:
Thanks Bhroam, I've addressed your nit.
@arungrover , requesting you to review this as well once more, thanks

bhroam:
Looks good, @agrawalravi90.  Please squash and rebase for the merge.

agrawalravi90:
New test results:
[rrtrostest.log](https://github.com/PBSPro/pbspro/files/2566666/rrtrostest.log)

arungrover:
appveyor build failed again, I've restarted the build

mike0042:
Filed ticket pertaining to Appveyor failures: https://pbspro.atlassian.net/browse/PP-1315

agrawalravi90:
Appveyor finally passed! Requesting @PBSPro/pbspro-maintainers to merge this, thanks

agrawalravi90:
Thanks for the merge @mike0042 !

PR #850, opened 2018-10-22
fix server memory leak in fix_indirectness involving indirect resources

nithinj:
#### Bug/feature Description
* *Memory leak occurs when a custom resource is getting deleted on which a direct value had overwritten a previously existing indirect value.*

* *When an indirect value assigned to a custom resource, the resource assigned is populated to maintain the same string value. But when it is over-written using a direct resource, only resource_available is freed and modified and resource_assigned is kept intact.*

#### Solution Description
* *Free the resource_assigned value when an indirect resource is over-written using a direct resource.*

#### Testing logs/output
PTL is added.
[ptl_after_fix.log](https://github.com/PBSPro/pbspro/files/2500559/ptl_after_fix.log)
[ptl_before_fix.log](https://github.com/PBSPro/pbspro/files/2500560/ptl_before_fix.log)
[server_after_fix.log](https://github.com/PBSPro/pbspro/files/2500561/server_after_fix.log)

#### Checklist:

<!
[server_before_fix.log](https://github.com/PBSPro/pbspro/files/2500562/server_before_fix.log)
--- If you're unsure about any of these, don't hesitate to ask. We're here to help! -->

CC: @bayucan

### DIFF
--- a/src/server/node_func.c
+++ b/src/server/node_func.c
@@ -77,7 +77,7 @@
  *	delete_a_subnode()		-	mark a (last) single subnode entry as deleted
  *	mod_node_ncpus()		-	when resources_available.ncpus changes, need to update the number of subnodes,
  *								 creating or deleting as required
- *	set_clear_target()		-	set or clear ATR_VFLAG_TARGET flag in a target resource "index"
+ *	fix_indirect_resc_targets()	-	set or clear ATR_VFLAG_TARGET flag in a target resource "index"
  *								is the index into the node's attribute array.
  *	indirect_target_check()	-	called via a work task to (re)set ATR_VFLAG_TARGET
  *									in any resource which is the target of another indirect resource.
@@ -532,7 +532,8 @@ initialize_pbsnode(struct pbsnode *pnode, char *pname, int ntype)
 	/* or  ATR_DFLAG_ANASSN set in the resource definition          */
 
 	for (prd = svr_resc_def; prd; prd = prd->rs_next) {
-		if (prd->rs_flags & (ATR_DFLAG_FNASSN |  ATR_DFLAG_ANASSN)) {
+		if ((prd->rs_flags & (ATR_DFLAG_FNASSN | ATR_DFLAG_ANASSN)) &&
+									(prd->rs_flags & ATR_DFLAG_MOM)) {
 			presc = add_resource_entry(pat2, prd);
 			presc->rs_value.at_flags = ATR_VFLAG_SET |
 				ATR_VFLAG_MODCACHE;
@@ -1747,9 +1748,12 @@ mod_node_ncpus(struct pbsnode *pnode, long ncpus, int actmode)
 
 /**
  * @brief
- * 		set_clear_target - set or clear ATR_VFLAG_TARGET flag in a target resource
- *		"index" is the index into the node's attribute array (which attr).
- *		either ND_ATR__ResourceAvail or ND_ATR_ResourceAssn
+ * 		fix_indirect_resc_targets - set or clear ATR_VFLAG_TARGET flag in
+ * 		a target resource "index" is the index into the node's attribute
+ * 		array (which attr). If invoked with ND_ATR__ResourceAvail or 
+ * 		ND_ATR_ResourceAssn, the target flag is applied on both. We need
+ * 		to do this as the check for target flag in fix_indirectness relies
+ * 		on resources_assigned as resources_available is already got over-written.
  *
  * @param[out]	psourcend	- Vnode structure
  * @param[in]	psourcerc	- target resource
@@ -1758,10 +1762,10 @@ mod_node_ncpus(struct pbsnode *pnode, long ncpus, int actmode)
  *
  * @return	int
  * @retval	-1	- error
- * @retval	-	- success
+ * @retval	0	- success
  */
 int
-set_clear_target(struct pbsnode *psourcend, resource *psourcerc, int index, int set)
+fix_indirect_resc_targets(struct pbsnode *psourcend, resource *psourcerc, int index, int set)
 {
 	char		*nname;
 	char		*pn;
@@ -1796,7 +1800,29 @@ set_clear_target(struct pbsnode *psourcend, resource *psourcerc, int index, int 
 			ptargetrc->rs_value.at_flags |= ATR_VFLAG_TARGET;
 		else
 			ptargetrc->rs_value.at_flags &= ~ATR_VFLAG_TARGET;
+
+		if (index == ND_ATR_ResourceAvail)
+			index = ND_ATR_ResourceAssn;
+		else
+			index = ND_ATR_ResourceAvail;
+
+		ptargetrc = find_resc_entry(&pnode->nd_attr[index], psourcerc->rs_defin);
+		if (!ptargetrc) {
+			/* For unset if the avail/assign counterpart is null, just return without creating the rescource.
+			* This happens only during node clean-up stage */
+			if (!set || index == ND_ATR_ResourceAvail)
+				return 0;
+			ptargetrc = add_resource_entry(&pnode->nd_attr[index], psourcerc->rs_defin);
+			if (!ptargetrc)
+				return PBSE_SYSTEM;
+		}
+
+		if (set)
+			ptargetrc->rs_value.at_flags |= ATR_VFLAG_TARGET;
+		else
+			ptargetrc->rs_value.at_flags &= ~ATR_VFLAG_TARGET;
 	}
+
 	return 0;
 }
 
@@ -1835,7 +1861,7 @@ indirect_target_check(struct work_task *ptask)
 				presc = (resource *)GET_NEXT(presc->rs_link)) {
 
 				if (presc->rs_value.at_flags & ATR_VFLAG_INDIRECT) {
-					set_clear_target(pnode, presc, (int)ND_ATR_ResourceAvail, 1);
+					fix_indirect_resc_targets(pnode, presc, (int)ND_ATR_ResourceAvail, 1);
 				}
 			}
 		}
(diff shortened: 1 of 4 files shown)


src/server/node_func.c
```diff
@@ -2001,8 +2001,16 @@ fix_indirectness(resource *presc, struct pbsnode *pnode, int doit)
...
 				ptargetan->rs_value.at_flags &= ~ATR_VFLAG_INDIRECT;
 			}
-
 			run_safety_check = 1;
+		} else {
+			/* Rescource assigned is not getting freed when a resource is set on an indirect rescource*/
+			ptargetan = find_resc_entry(&pnode->nd_attr[(int)ND_ATR_ResourceAssn], prdef);
+			if (consumable && (ptargetan != NULL) && (ptargetan->rs_value.at_flags & ATR_VFLAG_INDIRECT)) {
+				free_str(&ptargetan->rs_value);
+				prdef->rs_decode(&ptargetan->rs_value, NULL, NULL, NULL);
```
arungrover: What happens if there are running jobs on this vnode? If you just set it to 0 then the indirect resources assigned to the job on this vnode will suddenly become 0 and resources_assigned on the vnode from where the resource was being shared will never get decremented.

nithinj: You are right. The resources consumed by job are not getting released.
By looking more into the code, we are expecting old values in `porignalrc` and new values in `presc`. But `porignalrc` got modified before arriving at action function and it maintains the same values as `presc`. So any check in this function related to "porignalrc" simply doesnt work. One such check is the presence of job in the node we are setting the indirect resource. Another, check which is failing is when you try to set an indirect resource on a target resource as it would create a graph like structure. Then the third usage of `porignalrc` is to do what I fixed by adding another if condition.
An obvious way to fix all the three issues by checking `rescource_assigned` value for flags instead of `resource_available`. `resource_assigned` will still maintain the flags as it is not overwritten in `mgr_set_attr2`. I need to check the implications of fixing it this way.
These three cases also needs to be covered using automated test cases. I'll update you once the changes are ready.

arungrover: I don't think we should use resources_assigned to do the job. That just seems like a hack.
I'd suggest you create a setter function that you add to node's resources_available attribute and let that setter first call fix_indirectness and see whether we should even be setting the attribute. If it fails, don't set anything.

nithinj: The setter function is common for all resources - `set_resc`. It is a very generic function and it does not pass the parent object during set.

arungrover: Well I'm not asking you to change set_resc but to implement a function that will call set_resc inside it. I didn't notice the parent object, is there some other way to get that? I'll look into it tomorrow.

arungrover: Me and Bhroam had a long conversation about this issue, we both are in agreement that for short term using resources_assigned would solve the problem. Currently action functions are serving dual purpose (to validate the inputs and perform action). Unless we separate them out it is really hard to solve this problem. IMO validation should occur before we set anything but the reason it is done this way is to make sure that conflicting attributes do not get set at the same time (like walltime and soft_walltime).

nithinj:
Hi @arungrover ,
I've updated my solution. Please have a look.
Here is the PTL logs for the new solution: 
[PTL_failure_cases.txt](https://github.com/PBSPro/pbspro/files/2565319/PTL_failure_cases.txt)
[PTL_success_case.txt](https://github.com/PBSPro/pbspro/files/2565320/PTL_success_case.txt)

test/tests/functional/pbs_indirect_resources.py
```diff
@@ -0,0 +1,137 @@
...
+        self.server.expect(NODE, {'state': 'job-busy'}, id='vnode[1]')
+        try:
+            self.server.manager(MGR_CMD_SET, NODE,
+                                {'resources_available.fooi': 100},
+                                'vnode[1]')
+        except PbsManagerError:
+            pass
+        else:
+            _msg = "Setting resources on busy object should fail"
+            self.assertTrue(False, _msg)
```
arungrover: Instead, assert if fooi is set to 100

src/server/node_func.c
```diff
@@ -1948,8 +1968,8 @@ fix_indirectness(resource *presc, struct pbsnode *pnode, int doit)
 
 			/* new is not indirect, was the original? */
 
-			if (porignalrc) {
-				if (porignalrc->rs_value.at_flags & ATR_VFLAG_INDIRECT) {
+			if (presc_assn) {
```
arungrover: Please add a comment that this is a hack that we are using resource-assigned to identify that the resource was an indirect resource because attribute's set function has already changed resources-available

src/server/node_func.c
```diff
@@ -1899,9 +1919,9 @@ fix_indirectness(resource *presc, struct pbsnode *pnode, int doit)
 			/* setting this resource to be indirect, make serveral checks */
 
 			/* this vnode may not be a target of another indirect */
-			if (porignalrc) {
-				if (porignalrc->rs_value.at_flags & ATR_VFLAG_TARGET) {
-					if ((resc_in_err = strdup(porignalrc->rs_defin->rs_name)) == NULL)
+			if (presc_assn) {
```
arungrover: Does this need to be assigned? or can it be resources_available?

src/server/node_func.c
```diff
@@ -533,7 +533,8 @@ initialize_pbsnode(struct pbsnode *pnode, char *pname, int ntype)
 	/* or  ATR_DFLAG_ANASSN set in the resource definition          */
 
 	for (prd = svr_resc_def; prd; prd = prd->rs_next) {
-		if (prd->rs_flags & (ATR_DFLAG_FNASSN |  ATR_DFLAG_ANASSN)) {
+		if ((prd->rs_flags & (ATR_DFLAG_FNASSN | ATR_DFLAG_ANASSN)) &&
+									(prd->rs_flags & ATR_DFLAG_MOM)) {
```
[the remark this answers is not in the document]
nithinj: If you create a consumable resource and create a vnode/node, this resource will appear on the node resources_assigned part even though this resource is never been set on that particular node. The code above is going through all the resources available in `svr_resc_def` and adds them nodes resources_available section. This is not correct.
But there are some consumable resources in the master definition file which is required to be consumed by mom tagged with `ATR_DFLAG_MOM` and has to be added to newly created nodes/vnodes anyways.

src/server/node_func.c
```diff
@@ -1797,7 +1798,25 @@ set_clear_target(struct pbsnode *psourcend, resource *psourcerc, int index, int
 			ptargetrc->rs_value.at_flags |= ATR_VFLAG_TARGET;
 		else
 			ptargetrc->rs_value.at_flags &= ~ATR_VFLAG_TARGET;
+
```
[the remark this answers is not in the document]
nithinj: I'll update the header.

src/server/node_func.c
```diff
@@ -1797,7 +1798,25 @@ set_clear_target(struct pbsnode *psourcend, resource *psourcerc, int index, int
...
+		if (index == ND_ATR_ResourceAvail)
+			index = ND_ATR_ResourceAssn;
+		else
+			index = ND_ATR_ResourceAvail;
+
+		ptargetrc = find_resc_entry(&pnode->nd_attr[index], psourcerc->rs_defin);
+		if (!ptargetrc) {
+			ptargetrc = add_resource_entry(&pnode->nd_attr[index], psourcerc->rs_defin);
```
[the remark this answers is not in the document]
nithinj: We need to ensure that an indirect resource is not been set on a node which is already a target. This is been achieved using resources_assigned flags. For that flags in both resources_available and assigned should be in sync. 
Here a corresponding resources_assigned is added and a target flag is set on it. This will not be displayed during a stat command as the attribute is not `SET` for non-consumable resources.

src/server/node_func.c
nithinj: Earlier this was checking for flags present in resources_available structure which got updated already.

nithinj:
Hi @arungrover ,

src/server/node_func.c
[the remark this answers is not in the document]
arungrover: In that case, just doing this when index is resources_available will not suffice? I also feel that the purpose of this function was just to set the target flag and add resources_assigned to the object. It does not fit the purpose of the function.

src/server/node_func.c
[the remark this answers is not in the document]
nithinj: Along with `ATR_DFLAG_MOM` we are checking for either `ATR_DFLAG_FNASSN` or `ATR_DFLAG_ANASSN`. I think the purpose is to convey the usage in stat command. Between it's set for the following resources on a fresh node - `hbmem, mem, ncpus, vmem`

src/server/node_func.c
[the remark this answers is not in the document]
nithinj: There was no conventions. It was being invoked from `unset_indirect()` to clear flag for `resources_assigned`

src/server/node_func.c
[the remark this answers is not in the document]
nithinj: Thanks for the info about  "resource_unset_infinite". I have tested this part. If there is any change in the resources_assigned it will get set in node. So we are safe here.

src/server/node_func.c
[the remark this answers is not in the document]
arungrover: I strongly feel to move this logic out of this function. All these special cases make this function very special. It does not match the intent this function was written for in the first place.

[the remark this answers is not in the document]
nithinj: Say, I'm going to make this outside the function `set_clear_target`. I need to do the following:
1) getting the node name from the resource value.
2) Invoking find_nodebyname() to find the node structure (takes log(n) time)
3) Finding the resource entry and if not found adding the appropriate entry.
4) Setting the flag on the rescource_entry

As we need to do this in multiple places it's better to be a function. And this fucntion will look similar to `set_clear_target` but with additional complexity. I'd like to hear from you before I proceed with this method. Personally, I still feel the existing function is good enough.

nithinj:
@Bhagat-Rajput ,
Can you start reviewing this PR.

Bhagat-Rajput:
@nithinj appveyor build is failed, so please have a look at it.

nithinj:
Here is the lastest test logs: 
[latest logs.txt](https://github.com/PBSPro/pbspro/files/2632956/latest.logs.txt)

I've rebased the branch. Please take this further.

arungrover:
Well, something else went in and appveyor on debug build is failing. I have triggered a rebuild.

arungrover:
In any case, you will have to rebase again. If everything goes fine, I'll merge it in the evening today

nithinj:
@arungrover ,
Sorry for the delay. I've rebased it and all the tests have passed. Please take this PR further,

PR #851, opened 2018-10-23
test_prologue_hook_does_not_execute_twice is failing due to incorrect mom in log_match

VishrutArora:
#### Bug/feature Description
* test_prologue_hook_does_not_execute_twice is failing due to incorrect mom object in the log_match

#### Affected Platform(s)
* All

* test_prologue_hook_does_not_execute_twice is failing due to incorrect mom being specified for the log_match function

#### Solution Description
* Changed the mom object of the log_match function using the job output file.

#### Testing logs/output

#### Checklist:

### DIFF
--- a/test/tests/functional/pbs_hook_execjob_prologue.py
+++ b/test/tests/functional/pbs_hook_execjob_prologue.py
@@ -239,6 +239,9 @@ class TestPbsExecutePrologue(TestFunctional):
 
     def test_prologue_hook_does_not_execute_twice_on_pbsdsh(self):
         """
+        This test creates a hook and then submits a job.
+        It then uses the job output file to do a log_match
+        on both the moms
         """
         hook_name = 'prologue'
         hook_body = ("import pbs\n"
@@ -249,10 +252,17 @@ class TestPbsExecutePrologue(TestFunctional):
 
         j = Job(TEST_USER, {'Resource_List.select': '2:ncpus=1',
                             'Resource_List.place': 'scatter'})
-        j.create_script('pbsdsh -n 1 hostname')
+        j.create_script('#!/bin/sh\npbsdsh  hostname\nsleep 10\n')
         jid = self.server.submit(j)
-        self.server.expect(JOB, 'queue', op=UNSET, id=jid)
-        match = self.momB.log_match("Job;%s;executed prologue hook" % jid,
-                                    allmatch=True)
-        self.assertTrue(match is not None)
-        self.assertTrue(len(match) == 1)
+        attribs = self.server.status(JOB, id=jid)
+        self.server.expect(JOB, 'queue', op=UNSET, id=jid, offset=10)
+        host, opath = attribs[0]['Output_Path'].split(':', 1)
+        ret = self.du.cat(hostname=host, filename=opath, runas=TEST_USER)
+        _msg = "cat command failed with error: %s" % ret['err']
+        self.assertEqual(ret['rc'], 0, _msg)
+        mom1 = ret['out'][2].split(".")[0]
+        mom2 = ret['out'][3].split(".")[0]
+        self.exec_mom1 = self.moms[mom1]
+        self.exec_mom2 = self.moms[mom2]
+        self.exec_mom1.log_match("Job;%s;executed prologue hook" % jid)
+        self.exec_mom2.log_match("Job;%s;executed prologue hook" % jid)


test/tests/functional/pbs_hook_execjob_prologue.py
```diff
@@ -249,10 +249,13 @@ def test_prologue_hook_does_not_execute_twice_on_pbsdsh(self):
 
         j = Job(TEST_USER, {'Resource_List.select': '2:ncpus=1',
                             'Resource_List.place': 'scatter'})
-        j.create_script('pbsdsh -n 1 hostname')
+        j.set_sleep_time("5")
```
anamikau: however I have learnt that it is not mandatory to have pbsdsh for exec job_prologue hook now to run prologue hook on sister moms, but this test is about running pbsdsh.

Your changes are fine but not testing what this test was supposed to do. instead of reading from exec_vnode read from the job output file. update the job script to have "pbsdsh hostname" and submit a 2 chunk job. Read the output file and get the hostnames on which this job has run. for example,

anamika@testdev-01-s11p2:~> cat job1.sh.o9
x40-64-centos64.pbspro.com
testdev-01-s11p2

and then look for the "Job;9.testdev-01-s11p2;executed prologue hook" in mom_logs of both the hosts.

bhroam:
@VishrutArora Please fill out the top part of the template.

bhroam:
Travis is failing due to pep8

test/tests/functional/pbs_hook_execjob_prologue.py
VishrutArora: I've made the changes as you suggested. However, I've removed the "-n 1" from the job script as it was causing only the mother superior to be printed in the job o/p file.

test/tests/functional/pbs_hook_execjob_prologue.py
```diff
@@ -247,12 +250,20 @@ def test_prologue_hook_does_not_execute_twice_on_pbsdsh(self):
...
+                               'Resource_List.place': 'scatter'})
+        j.create_script('pbsdsh  hostname')
         jid = self.server.submit(j)
+        attribs = self.server.status(JOB, id=jid)
         self.server.expect(JOB, 'queue', op=UNSET, id=jid)
-        match = self.momB.log_match("Job;%s;executed prologue hook" % jid,
-                                    allmatch=True)
-        self.assertTrue(match is not None)
-        self.assertTrue(len(match) == 1)
+        host, opath = attribs[0]['Output_Path'].split(':', 1)
```
anamikau: if the job status is unset how will you get the value of output_path? you need to read it from the server_logs or enable job history or add a sleep for 10-12s in the job script.

test/tests/functional/pbs_hook_execjob_prologue.py
```diff
@@ -247,12 +250,20 @@ def test_prologue_hook_does_not_execute_twice_on_pbsdsh(self):
...
+        ret = self.du.cat(hostname=self.mom.shortname, filename=opath,
```
anamikau: again why do you need to be root to read the output file?

test/tests/functional/pbs_hook_execjob_prologue.py
```diff
@@ -247,12 +250,20 @@ def test_prologue_hook_does_not_execute_twice_on_pbsdsh(self):
...
+                          sudo=True)
+        mom1 = ret['out'][0].split(".")[0]
```
anamikau: I believe you also need to check if above cat command is successful or output is not empty before reading from the file.

test/tests/functional/pbs_hook_execjob_prologue.py
VishrutArora: @anamikau I believe the output file will be available only when the job is finished right? And by this, I'm actually getting the contents of the output file after the job has finished.
Let me know if I'm missing something.

anamikau: hey output_path will be set as soon as the job starts. you are reading this from status/qstat. once the job is done you won't get any output from qstat. you are getting this just by luck I would say as the systems you will be using are fast. But it might not be true all the time.

test/tests/functional/pbs_hook_execjob_prologue.py
VishrutArora: I've put a check for cat command as suggested.

test/tests/functional/pbs_hook_execjob_prologue.py
```diff
@@ -252,13 +252,14 @@ def test_prologue_hook_does_not_execute_twice_on_pbsdsh(self):
 
         j = Job(PBSROOT_USER, {'Resource_List.select': '2:ncpus=1',
                                'Resource_List.place': 'scatter'})
-        j.create_script('pbsdsh  hostname')
+        j.create_script('pbsdsh  hostname\nsleep 10\n')
```
anamikau: I think you need to add shebang as on some platform (ubuntu) you might face problems running this.

VishrutArora:
@anamikau I've addressed the comments given by you. Kindly review.

test/tests/functional/pbs_hook_execjob_prologue.py
```diff
@@ -252,13 +252,14 @@ def test_prologue_hook_does_not_execute_twice_on_pbsdsh(self):
...
         jid = self.server.submit(j)
         attribs = self.server.status(JOB, id=jid)
         self.server.expect(JOB, 'queue', op=UNSET, id=jid)
```
anamikau: add an offset of 10s since job is now sleeping longer.

test/tests/functional/pbs_hook_execjob_prologue.py
```diff
@@ -252,13 +252,14 @@ def test_prologue_hook_does_not_execute_twice_on_pbsdsh(self):
...
         host, opath = attribs[0]['Output_Path'].split(':', 1)
-        ret = self.du.cat(hostname=self.mom.shortname, filename=opath,
-                          sudo=True)
+        ret = self.du.cat(hostname=self.mom.shortname, filename=opath)
```
anamikau: should not the hostname be "host" that you are reading from qstat output above?

test/tests/functional/pbs_hook_execjob_prologue.py
anamikau: Thanks Vishrut. what if the hostnames are not in FQDN but short hostname?

test/tests/functional/pbs_hook_execjob_prologue.py
VishrutArora: It is one and the same thing. Here we just mean to get the location of the o/p file which is fixed. We do not mean to test the host obtained from the o/p path.

VishrutArora:
@anamikau Even if the obtained hostname is already a "shortname" instead of FQDN, the split will cause no harm and will return everything in the string. That is the exact case on the machine I've used over here.

test/tests/functional/pbs_hook_execjob_prologue.py
anamikau: I think on systems like CRAY or HPE ICE where mom is not on the same machine I think it makes more sense to use the host from the qstat -f output.

VishrutArora:
@anamikau I've addressed the comments. Kindly review.

anamikau:
Thanks Vishrut. Code change looks good. Please provide the execution results.

VishrutArora:
@anamikau 
[prologue_log_30th.txt](https://github.com/PBSPro/pbspro/files/2530785/prologue_log_30th.txt)
Please find execution results as requested

test/tests/functional/pbs_hook_execjob_prologue.py
```diff
@@ -247,12 +250,21 @@ def test_prologue_hook_does_not_execute_twice_on_pbsdsh(self):
         attr = {'event': 'execjob_prologue'}
         self.server.create_import_hook(hook_name, attr, hook_body)
 
-        j = Job(TEST_USER, {'Resource_List.select': '2:ncpus=1',
-                            'Resource_List.place': 'scatter'})
-        j.create_script('pbsdsh -n 1 hostname')
+        j = Job(PBSROOT_USER, {'Resource_List.select': '2:ncpus=1',
```
bhroam: Why did you change this from TEST_USER to PBSROOT_USER?

test/tests/functional/pbs_hook_execjob_prologue.py
```diff
@@ -247,12 +250,21 @@ def test_prologue_hook_does_not_execute_twice_on_pbsdsh(self):
...
-        self.server.expect(JOB, 'queue', op=UNSET, id=jid)
-        match = self.momB.log_match("Job;%s;executed prologue hook" % jid,
-                                    allmatch=True)
-        self.assertTrue(match is not None)
-        self.assertTrue(len(match) == 1)
+        attribs = self.server.status(JOB, id=jid)
+        self.server.expect(JOB, 'queue', op=UNSET, id=jid, offset=10)
+        host, opath = attribs[0]['Output_Path'].split(':', 1)
+        ret = self.du.cat(hostname=host, filename=opath)
+        _msg = "cat command failed with error:%s" % ret['err']
```
bhroam: nit: put a space between error: and the error

test/tests/functional/pbs_hook_execjob_prologue.py
```diff
@@ -247,12 +250,21 @@ def test_prologue_hook_does_not_execute_twice_on_pbsdsh(self):
...
+        self.assertEqual(ret['rc'], 0, _msg)
+        mom1 = ret['out'][0].split(".")[0]
+        mom2 = ret['out'][1].split(".")[0]
+        self.exec_mom1 = self.moms[mom1]
+        self.exec_mom2 = self.moms[mom2]
+        self.exec_mom1.log_match("Job;%s;executed prologue hook" % jid,
+                                 allmatch=True)
+        self.exec_mom2.log_match("Job;%s;executed prologue hook" % jid,
+                                 allmatch=True)
```
bhroam: Since you are just checking for the existence of this message, why pass allmatch?  If you were doing something with the return value of log_match(), I could see it.  Since you aren't, I don't think it is necessary.

test/tests/functional/pbs_hook_execjob_prologue.py
VishrutArora: @bhroam I changed this because the job o/p file will then be accessible to pbsroot easily in the test case.

bhroam: I think you are making an assumption this test will always be run by pbsroot.  This may be the case by internal Altair automated systems, but the test should be able to be run by any user.

VishrutArora: Yes, you are right. I've changed the user back to TEST_USER.

VishrutArora:
@bhroam I've addressed your comments. Kindly Review

PR #872, opened 2018-11-06
PBS server reports expired subjobs to scheduler

arungrover:
#### Bug/feature Description
While fixing the issue in PR #791 I added a new log message in scheduler which is logged if scheduler find a subjob in invalid state (!RUNNING). Along with this server was also changed to not report non-running subjobs to scheduler. But that part of the fix is not working and as a result, scheduler logs a bunch of messages about subjobs being invalid every cycle when job history is enabled.

#### Affected Platform(s)
All

In server the check to bail out if the subjob was a not running subjob was placed at a wrong place.

#### Solution Description
Moved the check to the right place.

#### Testing logs/output
[test_without_fix.txt](https://github.com/PBSPro/pbspro/files/2551199/test_without_fix.txt)
[test_out.txt](https://github.com/PBSPro/pbspro/files/2551200/test_out.txt)
[test_smoketest.txt](https://github.com/PBSPro/pbspro/files/2551201/test_smoketest.txt)

#### Checklist:

### DIFF
--- a/test/tests/functional/pbs_job_array.py
+++ b/test/tests/functional/pbs_job_array.py
@@ -610,3 +610,31 @@ class TestJobArray(TestFunctional):
             self.server.submit(j)
         except PbsSubmitError as e:
             raise self.failureException("Failed to submit job: " + str(e.msg))
+
+    def test_expired_subjobs_not_reported(self):
+        """
+        Test if a subjob is finished and moves to expired state,
+        it is not reported to scheduler in the next scheduling
+        cycle. Scheduler expects only running subjobs to be reported to it.
+        """
+
+        a = {'job_history_enable': 'True'}
+        self.server.manager(MGR_CMD_SET, SERVER, a)
+
+        req_node = ":host=" + self.mom.shortname
+        res_req = {'Resource_List.select': '1:ncpus=1' + req_node,
+                   'array_indices_submitted': '1-16',
+                   'Resource_List.place': 'excl'}
+        j1 = Job(TEST_USER, attrs=res_req)
+        j1.set_sleep_time(2)
+        jid1 = self.server.submit(j1)
+        j1_sub1 = j1.create_subjob_id(jid1, 1)
+
+        self.server.expect(JOB, {'job_state': 'X'}, j1_sub1)
+        # Trigger a sched cycle
+        a = {'scheduling': 'True'}
+        self.server.manager(MGR_CMD_SET, SERVER, a)
+
+        msg = j1_sub1 + ";" + "Subjob found in undesirable state"
+        msg += ", ignoring this job"
+        self.scheduler.log_match(msg, existence=False, max_attempts=10)
--- a/src/server/req_select.c
+++ b/src/server/req_select.c
@@ -489,14 +489,15 @@ select_job(job *pjob, struct select_list *psel, int dosubjobs, int dohistjobs)
 		return 0;
 	}
 
+	if ((dosubjobs == 2) && (pjob->ji_qs.ji_svrflags & JOB_SVFLG_SubJob) &&
+		(pjob->ji_qs.ji_state != JOB_STATE_RUNNING)) /* select only running subjobs */
+		return 0;
+
 	if ((pjob->ji_qs.ji_svrflags & JOB_SVFLG_ArrayJob) == 0)
 		dosubjobs = 0;  /* not an Array Job,  ok to check state */
 	else if ((dosubjobs != 2) &&
 		(pjob->ji_qs.ji_svrflags & JOB_SVFLG_SubJob))
 		return 0;	/* don't bother to look at sub job */
-	else if ((dosubjobs == 2) && (pjob->ji_qs.ji_svrflags & JOB_SVFLG_SubJob) &&
-		(pjob->ji_qs.ji_state != JOB_STATE_RUNNING)) /* select only running subjobs */
-		return 0;
 
 	while (psel) {
 


test/tests/functional/pbs_array_job.py
```diff
@@ -0,0 +1,72 @@
...
+# under a commercial license agreement.
+#
+# Use of Altair’s trademarks, including but not limited to "PBS™",
+# "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
+# trademark licensing policies.
+
+from tests.functional import *
+
+
+class TestArrayJobs(TestFunctional):
```
agrawalravi90: There already is a file called "pbs_job_array.py", please add this test to that test suite.

test/tests/functional/pbs_array_job.py
```diff
@@ -0,0 +1,72 @@
...
+
+    """
+    This test suite tests if PBS's array job functionality
+    """
+    def test_expired_subjobs_not_reported(self):
+        """
+        Test if a subjob is finished ans moves to expired state,
```
agrawalravi90: typo and* instead of ans

test/tests/functional/pbs_array_job.py
```diff
@@ -0,0 +1,72 @@
...
+
+        a = {'resources_available.ncpus': 1}
+        self.server.manager(MGR_CMD_SET, NODE, a, self.mom.shortname)
+        a = {'job_history_enable': 'True'}
+        self.server.manager(MGR_CMD_SET, SERVER, a)
+
+        res_req = {'Resource_List.select': '1:ncpus=1',
+                   'array_indices_submitted': '1-6'}
+        j1 = Job(TEST_USER, attrs=res_req)
+        j1.set_sleep_time(2)
```
agrawalravi90: 2 seconds seems to small to me for slower machines, how about turning scheduling off after submitting the job, doing expect to see if it expired, then triggering the next cycle to search for the log message?

test/tests/functional/pbs_array_job.py
```diff
@@ -0,0 +1,72 @@
...
+    """
+    def test_expired_subjobs_not_reported(self):
+        """
+        Test if a subjob is finished ans moves to expired state,
+        it is not reported to scheduler in the next scheduling
+        cycle. Scheduler expects only running subjobs to be reported to it.
+        """
+
+        a = {'resources_available.ncpus': 1}
+        self.server.manager(MGR_CMD_SET, NODE, a, self.mom.shortname)
```
anamikau: why do you need to set ncpus=1? looking at your test it doesn't matter if you have 1 or more.

test/tests/functional/pbs_array_job.py
anamikau: @agrawalravi90 you are right in general. but here since we are not looking for job state in R but only X it is safe.

test/tests/functional/pbs_array_job.py
arungrover: Thanks for pointing me to that file, I'll move this test

test/tests/functional/pbs_array_job.py
arungrover: oops

test/tests/functional/pbs_array_job.py
arungrover: My intent is to have some subjobs in 'X' state and some in 'Q' or 'R'. Also, to have the test not take a long time I reduced the time interval to 2 seconds. I'd like to retain as is.

test/tests/functional/pbs_array_job.py
```diff
@@ -0,0 +1,72 @@
...
+# "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
+# trademark licensing policies.
+
+from tests.functional import *
+
+
+class TestArrayJobs(TestFunctional):
+
+    """
+    This test suite tests if PBS's array job functionality
```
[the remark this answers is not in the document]
arungrover: Ah.. I missed adding 'working properly' at the end of the sentence. removing 'if' will also work in this case. I'll change it;

test/tests/functional/pbs_array_job.py
arungrover: The reason is that I want some of the subjobs to be either in running state or queued state. This is why I made the available ncpus as 1 because then only one job can run at a time.

anamikau: okay but then this configuration would be hard to run on a CRAY or cpuset systems. why don't you submit an exclusive job so that PBS cannot run more than one job one a node?

test/tests/functional/pbs_array_job.py
anamikau: @arungrover if that is your intent then I suggest submit a job array requesting 100 subjobs or so.

test/tests/functional/pbs_array_job.py
arungrover: Well, you are right, but even if I take the node exclusively it might still not work because those systems come with loads of vnodes prebuilt. What I can probably do is that I can submit a job requesting mom name as the node where the job needs to run. This way it will only run on login node.

test/tests/functional/pbs_array_job.py
arungrover: Ok, I can increase the number of subjobs here

arungrover:
Thanks @agrawalravi90 and @anamikau  for your prompt review. I have addressed your comments and here is the test log - 
[test_out.txt](https://github.com/PBSPro/pbspro/files/2559527/test_out.txt)

test/tests/functional/pbs_job_array.py
```diff
@@ -610,3 +610,32 @@ def test_job_id_duplicate_server_restart(self):
...
+
+    def test_expired_subjobs_not_reported(self):
+        """
+        Test if a subjob is finished and moves to expired state,
+        it is not reported to scheduler in the next scheduling
+        cycle. Scheduler expects only running subjobs to be reported to it.
+        """
+
+        a = {'resources_available.ncpus': 1}
+        self.server.manager(MGR_CMD_SET, NODE, a, self.mom.shortname)
```
anamikau: I thought we decided to remove this and rather submit the jobs with exclusive flag with host=mom node to make it more generic.

arungrover: Thanks for pointing that out Anamika, I've pushed that change in

agrawalravi90:
Looks good to me as well, thanks.

mike0042:
Looks like the Appveyor release build timed out. Please squash and rebase, and I suspect it will work.

mike0042:
This time the debug build failed @arungrover. Please re-sign your commit to restart the CI checks. I also notice that the Codacy check isn't listed, but I saw that it passed previously.

arungrover:
@mike0042 I restarted it, I don't know why debug build is failing.

arungrover:
Thanks @mike0042

PR #878, opened 2018-11-12
Jobid/Resvid wrap it to zero after server restart multiple times consecutively 

Bhagat-Rajput:
#### Bug/feature Description
* Jobid/Resvid wrap it to zero after server restart multiple times consecutively 
* Steps to reproduce: 
   1. Submit a job and wait for the completion.
   2. Restart the server.
   3. Restart the server once again.
   4. Submit a job and this time will get the same jobid again (i.e. 0.servername)

#### Affected Platform(s)
* All

* PBS uses the global variable (sv_jobidnumber) to update the next jobidnumber in the database during shutdown.
 After server restart the variable intializes with zero again, now if server is restarted again (without submit a single job/resv) then its initialized value gets updated into the database. So next time whenever will start the PBS will get 0(Zero) jobid/resvid. 

#### Solution Description
* PBS will assign the updated jobidnumber value to the variable (sv_jobidnumber) at the time of server start/restart.

#### Testing logs/output
* Test logs: [ptl_pbs_trillion_jobid.txt](https://github.com/PBSPro/pbspro/files/2572311/ptl_pbs_trillion_jobid.txt)

#### Checklist:

### DIFF
--- a/src/server/req_quejob.c
+++ b/src/server/req_quejob.c
@@ -3236,7 +3236,7 @@ long long get_next_svr_sequence_id(void)
 	++svr_sequence_window_count;
 	ret_svr_sequence_id = svr_jobidnumber;
 	/* sequence window count is more than 1000, reset back to zero*/
-	if (svr_sequence_window_count > SEQ_WIN_INCR) {
+	if (svr_sequence_window_count >= SEQ_WIN_INCR) {
 		svr_sequence_window_count = 0;
 	}
 	/* If server job limit is over, reset back to zero */
--- a/test/tests/functional/pbs_trillion_jobid.py
+++ b/test/tests/functional/pbs_trillion_jobid.py
@@ -144,7 +144,7 @@ exit 0
         self.assertTrue(self.server.isUp(), restart_msg)
 
     def submit_job(self, sleep=10, lower=0,
-                   upper=0, job_id=None, job_msg=None):
+                   upper=0, job_id=None, job_msg=None, verify=False):
         """
         Helper method to submit a normal/array job
         and also checks the R state and particular jobid if success,
@@ -165,6 +165,9 @@ exit 0
         :param job_msg : Expected message upon submission failure
         :type  job_msg : int
 
+        :param verify : Checks Job status R
+        :type  verify : boolean(True/False)
+
         """
         arr_flag = False
         j = Job(TEST_USER)
@@ -178,15 +181,15 @@ exit 0
             if job_id is not None:
                 self.assertEqual(jid.split('.')[0], job_id)
             if arr_flag:
-                self.server.expect(JOB, {'job_state': 'B'}, id=jid)
-                self.server.expect(JOB, {'job_state=R': '%d' % (total_jobs)},
-                                   count=True, id=jid, extend='t')
-                if sleep == 1:
-                    self.server.expect(JOB, 'queue', op=UNSET, id=jid)
+                if verify:
+                    self.server.expect(JOB, {'job_state': 'B'}, id=jid)
+                    self.server.expect(
+                        JOB,
+                        {'job_state=R': total_jobs},
+                        count=True, id=jid, extend='t')
             else:
-                self.server.expect(JOB, {'job_state': 'R'}, id=jid)
-                if sleep == 1:
-                    self.server.expect(JOB, 'queue', op=UNSET, id=jid)
+                if verify:
+                    self.server.expect(JOB, {'job_state': 'R'}, id=jid)
         except PbsSubmitError as e:
             if job_msg is not None:
                 # if JobId already exist
@@ -323,13 +326,14 @@ exit 0
         # Check default limit(9999999) and wrap it 0
         a = {'resources_available.ncpus': 20}
         self.server.manager(MGR_CMD_SET, NODE, a, self.mom.shortname)
-        self.submit_job()
-        self.submit_job(lower=1, upper=2)
+        self.submit_job(verify=True)
+        self.submit_job(lower=1, upper=2, verify=True)
         self.submit_resv()
         sv_jobidnumber = 9999999  # default
         self.set_svr_sv_jobidnumber(sv_jobidnumber)
-        self.submit_job(job_id='%s' % (sv_jobidnumber))
-        self.submit_job(lower=1, upper=2, job_id='0[]')  # wrap it
+        self.submit_job(job_id='%s' % (sv_jobidnumber), verify=True)
+        self.submit_job(lower=1, upper=2, job_id='0[]',
+                        verify=True)  # wrap it
         self.submit_resv(resv_id='R1')
 
         # Check max limit (999999999999) and wrap it 0
@@ -342,12 +346,13 @@ exit 0
             runas=ROOT_USER,
             expect=True)
         self.server.expect(SERVER, seq_id)
-        self.submit_job()
-        self.submit_job(lower=1, upper=2)
+        self.submit_job(verify=True)
+        self.submit_job(lower=1, upper=2, verify=True)
         self.submit_resv()
         self.set_svr_sv_jobidnumber(sv_jobidnumber)
-        self.submit_job(job_id='%s' % (sv_jobidnumber))
-        self.submit_job(lower=1, upper=2, job_id='0[]')  # wrap it
+        self.submit_job(job_id='%s' % (sv_jobidnumber), verify=True)
+        self.submit_job(lower=1, upper=2, job_id='0[]',
+                        verify=True)  # wrap it
         self.submit_resv(resv_id='R1')
 
         # Someone set the max_job_sequence_id less than current jobid then also
@@ -363,8 +368,8 @@ exit 0
         self.server.expect(SERVER, seq_id)
         sv_jobidnumber = 123456789
         self.set_svr_sv_jobidnumber(sv_jobidnumber)
-        self.submit_job(job_id='%s' % (sv_jobidnumber))
-        self.submit_job(lower=1, upper=2, job_id='123456790[]')
+        self.submit_job(job_id='%s' % (sv_jobidnumber), verify=True)
+        self.submit_job(lower=1, upper=2, job_id='123456790[]', verify=True)
         self.submit_resv(resv_id='R123456791')
         # Set smaller(12345678) than current jobid(123456790)
         sv_jobidnumber = 12345678
@@ -376,8 +381,8 @@ exit 0
             runas=ROOT_USER,
             expect=True)
         self.server.expect(SERVER, seq_id)
-        self.submit_job(job_id='0')  # wrap it to zero
-        self.submit_job(lower=1, upper=2, job_id='1[]')
+        self.submit_job(job_id='0', verify=True)  # wrap it to zero
+        self.submit_job(lower=1, upper=2, job_id='1[]', verify=True)
         self.submit_resv(resv_id='R2')
 
     def test_verify_sequence_window(self):
@@ -387,8 +392,6 @@ exit 0
         """
         # Abruptly kill the server so next jobid should be 1000 after server
         # start
-        a = {'resources_available.ncpus': 15}
-        self.server.manager(MGR_CMD_SET, NODE, a, self.mom.shortname)
         self.set_svr_sv_jobidnumber(0)
         self.submit_job(job_id='0')
         self.submit_job(lower=1, upper=2, job_id='1[]')
@@ -411,14 +414,18 @@ exit 0
         self.submit_job(lower=1, upper=2, job_id='2004[]')
         self.submit_resv(resv_id='R2005')
 
+        # Verify the sequence window, incase of submitting more than 1001 jobs
+        # and all jobs should submit successfully without any duplication error
+        for _ in xrange(1010):
+            j = Job(TEST_USER)
+            self.server.submit(j)
+
     def test_jobid_duplication(self):
         """
                 Tests the JobId/ResvId duplication after wrap
                 Job/Resv shouldn't submit because previous
                 jobs with the same id's are still running
         """
-        a = {'resources_available.ncpus': 8}
-        self.server.manager(MGR_CMD_SET, NODE, a, self.mom.shortname)
         seq_id = {ATTR_max_job_sequence_id: 99999999}
         self.server.manager(
             MGR_CMD_SET,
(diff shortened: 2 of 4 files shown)


suresh-thelkar:
It looks like the global variable svr_jobidnumber is not upto date in all the scenarios which is why the issue is happening. Since this is the root cause of the issue, I suggest to fix this part of the code instead of having the code to not save to database when svr_jobidnumber  is equal to zero. 

The other way to fix this is to eliminate the global variable svr_jobidnumber(we can make it a local variable in the appropriate place) and make sure that we always have upto date value in server.sv_qs.sv_jobidnumber. This is because when we save we always use this variable only.

test/tests/functional/pbs_trillion_jobid.py
```diff
@@ -444,3 +444,27 @@ def test_jobid_duplication(self):
...
         self.submit_job(lower=1, upper=2, job_id='3[]')
+
+    def test_jobid_resvid_after_multiple_restart(self):
+        """
+        Test to check the Jobid/Resvid should not wrap to 0 during
+        server restart multiple times consecutively either gracefully/abruptly
+        """
+        a = {'resources_available.ncpus': 15}
+        self.server.manager(MGR_CMD_SET, NODE, a, self.mom.shortname)
+        self.set_svr_sv_jobidnumber(0)
```
suresh-thelkar: I don't think we need to reset svr_jobidnumber to zero in the database. In the system where this bug is present even without this step it is reproduced.

test/tests/functional/pbs_trillion_jobid.py
```diff
@@ -444,3 +444,27 @@ def test_jobid_duplication(self):
...
+        self.submit_job(job_id='0')
+        self.submit_job(lower=1, upper=2, job_id='1[]')
+        self.submit_resv(resv_id='R2')
+        # Gracefully stop and start the server twice consecutively
+        self.stop_and_restart_svr('normal')
+        self.stop_and_restart_svr('normal')
+        self.submit_job(job_id='3')
+        self.submit_job(lower=1, upper=2, job_id='4[]')
+        self.submit_resv(resv_id='R5')
+        # Abruptly kill and start the server twice consecutively
```
suresh-thelkar: Do we really need to check the case of abrupt killing of server as well ? In this case whatever value gets saved at the timing of abrupt kill of server, the same value gets recovered while server is getting restarted.

Bhagat-Rajput:
> 
> 
> It looks like the global variable svr_jobidnumber is not upto date in all the scenarios which is why the issue is happening. Since this is the root cause of the issue, I suggest to fix this part of the code instead of having the code to not save to database when svr_jobidnumber is equal to zero.
> 
Thanks @suresh-thelkar for the review, svr_jobidnumber gets updated on a single job/resv run. If in a time frame PBS doesn't run a single job/resv then as per my knowledge no need to update the jobidnumber to the database because anyhow it would be the same.
But we can implement this behaviour as well.

> The other way to fix this is to eliminate the global variable svr_jobidnumber(we can make it a local variable in the appropriate place) and make sure that we always have upto date value in server.sv_qs.sv_jobidnumber. This is because when we save we always use this variable only.

If will go this way will loose the jobid/resvid because server saves its data from several other places as well.

suresh-thelkar:
> > It looks like the global variable svr_jobidnumber is not upto date in all the scenarios which is why the issue is happening. Since this is the root cause of the issue, I suggest to fix this part of the code instead of having the code to not save to database when svr_jobidnumber is equal to zero.
> 
> Thanks @suresh-thelkar for the review, svr_jobidnumber gets updated on a single job/resv run. If in a time frame PBS doesn't run a single job/resv then as per my knowledge no need to update the jobidnumber to the database because anyhow it would be the same.
> But we can implement this behaviour as well.
>> As discussed you can actually update the variable svr_jobidnumber as soon as svr_recov_db() gets called. This way before svr_shutdown() gets called in the said scenario it contains upto date value.
> 
> > The other way to fix this is to eliminate the global variable svr_jobidnumber(we can make it a local variable in the appropriate place) and make sure that we always have upto date value in server.sv_qs.sv_jobidnumber. This is because when we save we always use this variable only.
> 
> If will go this way will loose the jobid/resvid because server saves its data from several other places as well.
>> Since the save is happening in several other places this may not be trivial and hence option-1 is fine with me.

test/tests/functional/pbs_trillion_jobid.py
Bhagat-Rajput: Its necessary because i am checking the job sequence as well.

test/tests/functional/pbs_trillion_jobid.py
Bhagat-Rajput: Yes it's important @suresh-thelkar. Now this functionality has been changed, after abruptly kill the server user will not get the next jobid/resvid.

Bhagat-Rajput:
@suresh-thelkar I have addressed your comments please review it.

test/tests/functional/pbs_trillion_jobid.py
```diff
@@ -444,3 +444,27 @@ def test_jobid_duplication(self):
...
         # passed
         self.submit_job(lower=1, upper=2, job_id='3[]')
+
+    def test_jobid_resvid_after_multiple_restart(self):
+        """
+        Test to check the Jobid/Resvid should not wrap to 0 during
+        server restart multiple times consecutively either gracefully/abruptly
+        """
+        a = {'resources_available.ncpus': 15}
+        self.server.manager(MGR_CMD_SET, NODE, a, self.mom.shortname)
```
arungrover: Why do you need to set node's resources available? You do not really care if the job is running or not.

test/tests/functional/pbs_trillion_jobid.py
arungrover: I agree with @suresh-thelkar, Instead of setting it in DB, submit a job, get the job id and use that numeral to do further calculations

Bhagat-Rajput: Addressed!

test/tests/functional/pbs_trillion_jobid.py
Bhagat-Rajput: Using a generic function self.submit_job() in which we were checking the R state, however have updated that as well.

Bhagat-Rajput:
Thanks @arungrover for finding the sequence window bug, I have addressed that as well along with this commit and added the testing scenario in PTL also. Please review it.

test/tests/functional/pbs_trillion_jobid.py
```diff
@@ -144,7 +144,7 @@ def stop_and_restart_svr(self, restart_type):
         self.assertTrue(self.server.isUp(), restart_msg)
 
     def submit_job(self, sleep=10, lower=0,
-                   upper=0, job_id=None, job_msg=None):
+                   upper=0, job_id=None, job_msg=None, status=None):
```
arungrover: Since the function only checks for 'R' state, I'd suggest you rename to status to something like 'running' or 'check_running' and make it a boolean with a default value as 'False'.

test/tests/functional/pbs_trillion_jobid.py
```diff
@@ -178,15 +181,15 @@ def submit_job(self, sleep=10, lower=0,
...
-                self.server.expect(JOB, {'job_state': 'B'}, id=jid)
-                self.server.expect(JOB, {'job_state=R': '%d' % (total_jobs)},
-                                   count=True, id=jid, extend='t')
-                if sleep == 1:
-                    self.server.expect(JOB, 'queue', op=UNSET, id=jid)
+                if status is not None:
+                    self.server.expect(JOB, {'job_state': 'B'}, id=jid)
+                    self.server.expect(
+                        JOB,
+                        {'job_state=R': '%d' % (total_jobs)},
```
arungrover: Can you not write total_jobs to the dictionary instead of substituting it as a string?

src/server/req_quejob.c
```diff
@@ -3236,7 +3236,7 @@ long long get_next_svr_sequence_id(void)
 	++svr_sequence_window_count;
 	ret_svr_sequence_id = svr_jobidnumber;
 	/* sequence window count is more than 1000, reset back to zero*/
-	if (svr_sequence_window_count > SEQ_WIN_INCR) {
```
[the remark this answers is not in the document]
Bhagat-Rajput: As per my understanding, we can't put this line in the end because at line 3243 we are checking the next jobid that should not be greater than the svr_max_job_sequence_id attribute. If  greater then need to wrap it to 0 from here itself.
Will not loose the jobid because we are assigning first (at line 3223) and then increasing the jobid later.
I think previously also PBS saves the advance jobid in the database and written a testcase as well to check this scenario.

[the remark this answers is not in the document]
arungrover: Ok, in that case change the operator in te check to be '>=' just to be on a safer side

test/tests/functional/pbs_trillion_jobid.py
Bhagat-Rajput: Addressed!

test/tests/functional/pbs_trillion_jobid.py
Bhagat-Rajput: Addressed!

src/server/req_quejob.c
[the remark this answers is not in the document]
Bhagat-Rajput: Done!

Bhagat-Rajput:
Thanks @arungrover and @vstumpf , addressed all the comments. Please review it.

Bhagat-Rajput:
Rebased and squashed, ready to merge.

test/tests/functional/pbs_trillion_jobid.py
```diff
@@ -165,28 +165,31 @@ def submit_job(self, sleep=10, lower=0,
...
+        :type  running : boolean(True/False)
+
         """
         arr_flag = False
         j = Job(TEST_USER)
         if((lower >= 0) and (upper > lower)):
             j.set_attributes({ATTR_J: '%d-%d' % (lower, upper)})
             arr_flag = True
-            total_jobs = upper - lower + 1
+            arr_jobs = {'total_jobs': '%d' % (upper - lower + 1)}
```
agrawalravi90: why did you turn this into a dict? looks like you only ever store one key-val pair into it

test/tests/functional/pbs_trillion_jobid.py
```diff
@@ -165,28 +165,31 @@ def submit_job(self, sleep=10, lower=0,
...
-                    self.server.expect(JOB, 'queue', op=UNSET, id=jid)
+                if running:
+                    self.server.expect(JOB, {'job_state': 'B'}, id=jid)
+                    self.server.expect(
+                        JOB,
+                        {'job_state=R': arr_jobs['total_jobs']},
+                        count=True, id=jid, extend='t')
             else:
-                self.server.expect(JOB, {'job_state': 'R'}, id=jid)
-                if sleep == 1:
```
agrawalravi90: not sure what the sleep == 1 check was for, but can you please comment on why you removed it?

test/tests/functional/pbs_trillion_jobid.py
```diff
@@ -144,7 +144,7 @@ def stop_and_restart_svr(self, restart_type):
         self.assertTrue(self.server.isUp(), restart_msg)
 
     def submit_job(self, sleep=10, lower=0,
-                   upper=0, job_id=None, job_msg=None):
+                   upper=0, job_id=None, job_msg=None, running=False):
```
agrawalravi90: how about 'verify' instead of 'running' ?

agrawalravi90: Also, you call this with running=True everywhere, so why did you decide to add this argument? Earlier it was always going to check that the job was in the 'R' state, and that's exactly what it will do even with this new argument because you call it with 'True' everywhere, right?

test/tests/functional/pbs_trillion_jobid.py
```diff
@@ -444,3 +451,34 @@ def test_jobid_duplication(self):
...
+        j = Job(TEST_USER)
+        jid = self.server.submit(j)
+        curr_id = int(jid.split('.')[0])
+        self.submit_job(job_id='%s' % str(curr_id + 1))
+        self.submit_job(lower=1, upper=2, job_id='%s[]' % str(curr_id + 2))
+        self.submit_resv(resv_id='R%s' % str(curr_id + 3))
+        # Gracefully stop and start the server twice consecutively
+        self.stop_and_restart_svr('normal')
+        self.stop_and_restart_svr('normal')
+        self.submit_job(job_id='%s' % str(curr_id + 4))
```
agrawalravi90: you aren't really checking that the job id wrapped to 0 or not, wouldn't it be a simpler and more targeted test if you just submitted a single job after the server restarts and check that its id is not 0 ?

test/tests/functional/pbs_trillion_jobid.py
```diff
@@ -444,3 +451,34 @@ def test_jobid_duplication(self):
...
+        self.stop_and_restart_svr('kill')
+        # Adding 1000 in current jobid for the sequence window buffer and
+        # 4 for the jobs that ran already after server start
+        curr_id += 1000 + 4
+        self.submit_job(job_id='%s' % str(curr_id))
+        self.submit_job(lower=1, upper=2, job_id='%s[]' % str(curr_id + 1))
+        self.submit_resv(resv_id='R%s' % str(curr_id + 2))
+
+    def tearDown(self):
+        self.server.cleanup_jobs(extend='force')
```
agrawalravi90: forcefully deleting the jobs doesn't really delete the jobs on the mom, which means that if another test is run immediately following this test, the next test might run with all of those jobs lurking around on the mom. So, I think you should remove the 'force' flag and delete jobs properly in tearDown.

test/tests/functional/pbs_trillion_jobid.py
agrawalravi90: It's always wise to do whatever Arun tells you to :) but I think you misunderstood what he meant to say. I think he was just asking you to pass total_jobs directly to the attribute dictionary which self.server.expect() takes as its second argument, instead of passing it as a string. i.e - he asked you to do
self.server.expect(JOB, {'job_state=R': total_jobs}, ..)
instead of 
{'job_state=R': '%d' % (total_jobs)}, ..)
which you had previously.

test/tests/functional/pbs_trillion_jobid.py
Bhagat-Rajput: self.submit_job() is a generic function and using throughout the whole testsuite in which we are checking the exact jobid numbers in other test cases.
So that's why used the same in this test also and if it will wrap to zero after server restart then testcase will fail anyway.
Checking the array jobs and reservation as well because they also gets the id from the same function.

test/tests/functional/pbs_trillion_jobid.py
Bhagat-Rajput: No @agrawalravi90, only in one testcase i am checking the running state of jobs.

agrawalravi90: my bad, thanks for clarifying

test/tests/functional/pbs_trillion_jobid.py
Bhagat-Rajput: Thanks @agrawalravi90 for the valuable suggestion, implemented.

test/tests/functional/pbs_trillion_jobid.py
Bhagat-Rajput: Addressed!

test/tests/functional/pbs_trillion_jobid.py
```diff
@@ -480,5 +480,5 @@ def test_jobid_resvid_after_multiple_restart(self):
         self.submit_resv(resv_id='R%s' % str(curr_id + 2))
 
     def tearDown(self):
-        self.server.cleanup_jobs(extend='force')
+        #self.server.cleanup_jobs()
```
agrawalravi90: I think you forgot to un-comment this line

Bhagat-Rajput: Done!

Bhagat-Rajput:
I've attached the latest test logs below: 
[ptl_pbs_trillion_jobid.txt](https://github.com/PBSPro/pbspro/files/2585558/ptl_pbs_trillion_jobid.txt)

agrawalravi90:
Looks good Bhagat, please allow Arun to look at the latest changes once, then either of us can take this forward.

arungrover:
Change looks good to me

agrawalravi90:
Please rebase for merge bhagat

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

Bhagat-Rajput:
Squashed and Rebased, please take this further. Thanks

arungrover:
Appveyor failed initially, I have restarted the build

Bhagat-Rajput:
Thanks @arungrover, it's passed now -:)

agrawalravi90:
Sorry Bhagat, I noticed your comment just now, I had already merged a PR before that. Can you please rebase again? I'll merge your PR before anybody else's. Thanks.

Bhagat-Rajput:
It's fine @agrawalravi90. I've rebased it. Thanks!

PR #888, opened 2018-11-16
PTL setUp() does not revert pbs.conf changes

agrawalravi90:
#### Bug/feature Description
* If anyone makes changes to pbs.conf in their tests, those changes won't be reverted before the next test is run.  This may cause subsequent tests to fail.

#### Affected Platform(s)
* All Linux

* Well, we just didn't revert pbs.conf in setUp

#### Solution Description
* Added a function called revert_pbsconf_startbits() in PBSTestSuite to revert pbs.conf, which now gets called from setUp()
* If run on a single host type 1 installation, the function sets all start bits to 1
* If run on a multi-node setup with moms or comms, the function sets all mom/comm start bits on those hosts to 1, and on the server host it sets all bits to 1 except the mom bit.

#### Testing logs/output
* 5 new tests, results on a single node machine (expect to see the remote mom test skipped): 
[testpbstestsuite.log](https://github.com/PBSPro/pbspro/files/2652640/testpbstestsuite.log)

* results for remote mom test:
[testremotemom.log](https://github.com/PBSPro/pbspro/files/2652643/testremotemom.log)

#### Checklist:

### DIFF
--- a/test/fw/ptl/utils/pbs_testsuite.py
+++ b/test/fw/ptl/utils/pbs_testsuite.py
@@ -440,6 +440,7 @@ class PBSTestSuite(unittest.TestCase):
             return
         self.log_enter_setup()
         self.init_proc_mon()
+        self.revert_pbsconf()
         self.revert_servers()
         self.revert_comms()
         self.revert_schedulers()
@@ -830,6 +831,331 @@ class PBSTestSuite(unittest.TestCase):
                 self.start_proc_monitor(name='|'.join(_proc_mon), regexp=True,
                                         frequency=freq)
                 self._process_monitoring = True
+
+    def _get_dflt_pbsconfval(self, conf, primary_host, hosttype, hostobj):
+        """
+        Helper function to revert_pbsconf, tries to determine and return
+        default value for the pbs.conf variable given
+
+        :param conf: the pbs.conf variable
+        :type conf: str
+        :param primary_host: hostname of the primary server host
+        :type primary_host: str
+        :param hosttype: type of host being reverted
+        :type hosttype: str
+        :param hostobj: PTL object associated with the host
+        :type hostobj: PBSService
+
+        :return default value of the pbs.conf variable if it can be determined
+        as a string, otherwise None
+        """
+        if conf == "PBS_SERVER":
+            return primary_host
+        elif conf == "PBS_START_SCHED":
+            if hosttype == "server":
+                return "1"
+            else:
+                return "0"
+        elif conf == "PBS_START_COMM":
+            if hosttype == "comm":
+                return "1"
+            else:
+                return "0"
+        elif conf == "PBS_START_SERVER":
+            if hosttype == "server":
+                return "1"
+            else:
+                return "0"
+        elif conf == "PBS_START_MOM":
+            if hosttype == "mom":
+                return "1"
+            else:
+                return "0"
+        elif conf == "PBS_CORE_LIMIT":
+            return "unlimited"
+        elif conf == "PBS_SCP":
+            scppath = self.du.which(hostobj.hostname, "scp")
+            if scppath != "scp":
+                return scppath
+
+        return None
+
+    def _revert_pbsconf_comm(self, primary_server, vals_to_set):
+        """
+        Helper function to revert_pbsconf to revert all comm daemons' pbs.conf
+
+        :param primary_server: object of the primary PBS server
+        :type primary_server: PBSService
+        :param vals_to_set: dict of pbs.conf values to set
+        :type vals_to_set: dict
+        """
+        for comm in self.comms.values():
+            if comm.hostname == primary_server.hostname:
+                continue
+
+            new_pbsconf = dict(vals_to_set)
+            restart_comm = False
+            pbs_conf_val = self.du.parse_pbs_config(comm.hostname)
+            if not pbs_conf_val:
+                raise ValueError("Could not parse pbs.conf on host %s" %
+                                 (comm.hostname))
+
+            # to start with, set all keys in new_pbsconf with values from the
+            # existing pbs.conf
+            keys_to_delete = []
+            for conf in new_pbsconf:
+                if conf in pbs_conf_val:
+                    new_pbsconf[conf] = pbs_conf_val[conf]
+                else:
+                    # existing pbs.conf doesn't have a default variable set
+                    # Try to determine the default
+                    val = self._get_dflt_pbsconfval(conf,
+                                                    primary_server.hostname,
+                                                    "comm", comm)
+                    if val is None:
+                        self.logger.error("Couldn't revert %s in pbs.conf"
+                                          " to its default value" %
+                                          (conf))
+                        keys_to_delete.append(conf)
+                    else:
+                        new_pbsconf[conf] = val
+
+            for key in keys_to_delete:
+                del(new_pbsconf[key])
+
+            # Set the comm start bit to 1
+            if new_pbsconf["PBS_START_COMM"] != "1":
+                new_pbsconf["PBS_START_COMM"] = "1"
+                restart_comm = True
+
+            # Set PBS_CORE_LIMIT, PBS_SCP and PBS_SERVER
+            if new_pbsconf["PBS_CORE_LIMIT"] != "unlimited":
+                new_pbsconf["PBS_CORE_LIMIT"] = "unlimited"
+                restart_comm = True
+            if new_pbsconf["PBS_SERVER"] != primary_server.hostname:
+                new_pbsconf["PBS_SERVER"] = primary_server.hostname
+                restart_comm = True
+            if "PBS_SCP" not in new_pbsconf:
+                scppath = self.du.which(comm.hostname, "scp")
+                if scppath != "scp":
+                    new_pbsconf["PBS_SCP"] = scppath
+                    restart_comm = True
+
+            # Check if existing pbs.conf has more/less entries than the
+            # default list
+            if len(pbs_conf_val) != len(new_pbsconf):
+                restart_comm = True
+
+            if restart_comm:
+                self.du.set_pbs_config(comm.hostname, confs=new_pbsconf)
+                comm.pbs_conf = new_pbsconf
+                comm.pi.initd(comm.hostname, "restart", daemon="comm")
+
+    def _revert_pbsconf_mom(self, primary_server, vals_to_set):
+        """
+        Helper function to revert_pbsconf to revert all mom daemons' pbs.conf
+
+        :param primary_server: object of the primary PBS server
+        :type primary_server: PBSService
+        :param vals_to_set: dict of pbs.conf values to set
+        :type vals_to_set: dict
+        """
+        for mom in self.moms.values():
+            if mom.hostname == primary_server.hostname:
+                continue
+
+            new_pbsconf = dict(vals_to_set)
+            restart_mom = False
+            pbs_conf_val = self.du.parse_pbs_config(mom.hostname)
+            if not pbs_conf_val:
+                raise ValueError("Could not parse pbs.conf on host %s" %
+                                 (mom.hostname))
+
+            # to start with, set all keys in new_pbsconf with values from the
+            # existing pbs.conf
+            keys_to_delete = []
+            for conf in new_pbsconf:
+                if conf in pbs_conf_val:
+                    new_pbsconf[conf] = pbs_conf_val[conf]
+                else:
+                    # existing pbs.conf doesn't have a default variable set
+                    # Try to determine the default
+                    val = self._get_dflt_pbsconfval(conf,
+                                                    primary_server.hostname,
+                                                    "mom", mom)
+                    if val is None:
+                        self.logger.error("Couldn't revert %s in pbs.conf"
+                                          " to its default value" %
+                                          (conf))
+                        keys_to_delete.append(conf)
+                    else:
+                        new_pbsconf[conf] = val
+
+            for key in keys_to_delete:
+                del(new_pbsconf[key])
+
+            # Set the mom start bit to 1
+            if (new_pbsconf["PBS_START_MOM"] != "1"):
+                new_pbsconf["PBS_START_MOM"] = "1"
+                restart_mom = True
+
+            # Set PBS_CORE_LIMIT, PBS_SCP and PBS_SERVER
+            if new_pbsconf["PBS_CORE_LIMIT"] != "unlimited":
+                new_pbsconf["PBS_CORE_LIMIT"] = "unlimited"
+                restart_mom = True
+            if new_pbsconf["PBS_SERVER"] != primary_server.hostname:
+                new_pbsconf["PBS_SERVER"] = primary_server.hostname
+                restart_mom = True
+            if "PBS_SCP" not in new_pbsconf:
+                scppath = self.du.which(mom.hostname, "scp")
+                if scppath != "scp":
+                    new_pbsconf["PBS_SCP"] = scppath
+                    restart_mom = True
+
+            # Check if existing pbs.conf has more/less entries than the
+            # default list
+            if len(pbs_conf_val) != len(new_pbsconf):
+                restart_mom = True
+
+            if restart_mom:
+                self.du.set_pbs_config(mom.hostname, confs=new_pbsconf,
+                                       append=False)
+                mom.pbs_conf = new_pbsconf
+                mom.pi.initd(mom.hostname, "restart", daemon="mom")
+
+    def _revert_pbsconf_server(self, primary_server, vals_to_set):
+        """
+        Helper function to revert_pbsconf to revert all servers' pbs.conf
+
+        :param primary_server: object of the primary PBS server
+        :type primary_server: PBSService
+        :param vals_to_set: dict of pbs.conf values to set
+        :type vals_to_set: dict
+        """
+        for server in self.servers.values():
+            new_pbsconf = dict(vals_to_set)
+            cmds_to_exec = []
+            dmns_to_restart = 0
+            restart_pbs = False
+            pbs_conf_val = self.du.parse_pbs_config(server.hostname)
+            if not pbs_conf_val:
+                raise ValueError("Could not parse pbs.conf on host %s" %
+                                 (server.hostname))
+
+            # to start with, set all keys in new_pbsconf with values from the
+            # existing pbs.conf
+            keys_to_delete = []
+            for conf in new_pbsconf:
+                if conf in pbs_conf_val:
+                    new_pbsconf[conf] = pbs_conf_val[conf]
+                else:
+                    # existing pbs.conf doesn't have a default variable set
+                    # Try to determine the default
+                    val = self._get_dflt_pbsconfval(conf,
+                                                    primary_server.hostname,
+                                                    "server", server)
+                    if val is None:
+                        self.logger.error("Couldn't revert %s in pbs.conf"
+                                          " to its default value" %
+                                          (conf))
+                        keys_to_delete.append(conf)
+                    else:
+                        new_pbsconf[conf] = val
+
+            for key in keys_to_delete:
+                del(new_pbsconf[key])
+
+            # Set all start bits
+            if (new_pbsconf["PBS_START_SERVER"] != "1"):
+                new_pbsconf["PBS_START_SERVER"] = "1"
+                dmns_to_restart += 1
+                cmds_to_exec.append(["server", "start"])
+            if (new_pbsconf["PBS_START_SCHED"] != "1"):
+                new_pbsconf["PBS_START_SCHED"] = "1"
+                cmds_to_exec.append(["sched", "start"])
+                dmns_to_restart += 1
+            if self.moms and server.hostname not in self.moms:
+                if new_pbsconf["PBS_START_MOM"] != "0":
+                    new_pbsconf["PBS_START_MOM"] = "0"
+                    cmds_to_exec.append(["mom", "stop"])
+                    dmns_to_restart += 1
+            else:
+                if (new_pbsconf["PBS_START_MOM"] != "1"):
+                    new_pbsconf["PBS_START_MOM"] = "1"
+                    cmds_to_exec.append(["mom", "start"])
+                    dmns_to_restart += 1
+            if self.comms and server.hostname not in self.comms:
+                if new_pbsconf["PBS_START_COMM"] != "0":
+                    new_pbsconf["PBS_START_COMM"] = "0"
+                    cmds_to_exec.append(["comm", "stop"])
+            else:
+                if (new_pbsconf["PBS_START_COMM"] != "1"):
+                    new_pbsconf["PBS_START_COMM"] = "1"
+                    cmds_to_exec.append(["comm", "start"])
+                    dmns_to_restart += 1
+
+            if dmns_to_restart == 4:
+                # If all daemons need to be started again, just restart PBS
+                # instead of making PTL start each of them one at a time
+                restart_pbs = True
+
+            # Set PBS_CORE_LIMIT, PBS_SCP and PBS_SERVER
+            if new_pbsconf["PBS_CORE_LIMIT"] != "unlimited":
+                new_pbsconf["PBS_CORE_LIMIT"] = "unlimited"
+                restart_pbs = True
+            if new_pbsconf["PBS_SERVER"] != primary_server.hostname:
+                new_pbsconf["PBS_SERVER"] = primary_server.hostname
+                restart_pbs = True
+            if "PBS_SCP" not in new_pbsconf:
+                scppath = self.du.which(server.hostname, "scp")
+                if scppath != "scp":
+                    new_pbsconf["PBS_SCP"] = scppath
+                    restart_pbs = True
+
+            # Check if existing pbs.conf has more/less entries than the
+            # default list
+            if len(pbs_conf_val) != len(new_pbsconf):
+                restart_pbs = True
+
+            if restart_pbs or dmns_to_restart > 0:
+                # Write out the new pbs.conf file
+                self.du.set_pbs_config(server.hostname, confs=new_pbsconf,
+                                       append=False)
+                server.pbs_conf = new_pbsconf
+
+                if restart_pbs:
+                    # Restart all
+                    server.pi.restart(server.hostname)
+                else:
+                    for initcmd in cmds_to_exec:
+                        # start/stop the particular daemon
+                        server.pi.initd(server.hostname, initcmd[1],
+                                        daemon=initcmd[0])
+
+    def revert_pbsconf(self):
+        """
+        Revert contents of the pbs.conf file
+        Also start/stop the appropriate daemons
+        """
+        primary_server = self.server
+
+        vals_to_set = {
+            "PBS_HOME": None,
+            "PBS_EXEC": None,
+            "PBS_SERVER": None,
+            "PBS_START_SCHED": None,
+            "PBS_START_COMM": None,
+            "PBS_START_SERVER": None,
+            "PBS_START_MOM": None,
+            "PBS_CORE_LIMIT": None,
+            "PBS_SCP": None
+        }
+
+        self._revert_pbsconf_server(primary_server, vals_to_set)
+
+        self._revert_pbsconf_mom(primary_server, vals_to_set)
+
+        self._revert_pbsconf_comm(primary_server, vals_to_set)
 
     def revert_servers(self, force=False):
         """
--- a/test/tests/selftest/pbs_pbstestsuite.py
+++ b/test/tests/selftest/pbs_pbstestsuite.py
@@ -1,0 +1,260 @@
+# coding: utf-8
+
+# Copyright (C) 1994-2018 Altair Engineering, Inc.
+# For more information, contact Altair at www.altair.com.
+#
+# This file is part of the PBS Professional ("PBS Pro") software.
+#
+# Open Source License Information:
+#
+# PBS Pro is free software. You can redistribute it and/or modify it under the
+# terms of the GNU Affero General Public License as published by the Free
+# Software Foundation, either version 3 of the License, or (at your option) any
+# later version.
+#
+# PBS Pro is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.
+# See the GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+# Commercial License Information:
+#
+# For a copy of the commercial license terms and conditions,
+# go to: (http://www.pbspro.com/UserArea/agreement.html)
+# or contact the Altair Legal Department.
+#
+# Altair’s dual-license business model allows companies, individuals, and
+# organizations to create proprietary derivative works of PBS Pro and
+# distribute them - whether embedded or bundled with other software -
+# under a commercial license agreement.
+#
+# Use of Altair’s trademarks, including but not limited to "PBS™",
+# "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
+# trademark licensing policies.
+
+from tests.selftest import *
+from ptl.utils.pbs_snaputils import *
+
+
+class TestPBSTestSuite(TestSelf):
+    """
+    Contains tests for pbs_testsuite module's functionality
+    """
+
+    def test_revert_pbsconf_onehost(self):
+        """
+        Test the functionality of PBSTestSuite.revert_pbsconf()
+        for a single host type 1 installation
+        """
+        pbs_conf_val = self.du.parse_pbs_config(self.server.hostname)
+        self.assertTrue(pbs_conf_val and len(pbs_conf_val) >= 1,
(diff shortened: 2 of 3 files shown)


bhroam:
If you revert anything, you will need to restart the PBS services.  You don't want to have to do that if nothing changes.  You will want to verify something has changed before you restart them.  Also, the mom gets restarted as part of setUp().  You might want to figure out how to only restart the mom once.

agrawalravi90:
Thanks guys, I've addressed your comments.

@arungrover run, for PBS_SCP, I'm only reverting it if it is not set. My thinking is that the only test scenario for it is if somebody unsets it to test if PBS uses rcp or not. So, if it is unset, I go through the effort to find the path to it and set PBS_SCP, otherwise i leave it alone. Please let me know if this sounds reasonable.

agrawalravi90:
@arungrover and @bhroam , gentle reminder, thanks

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -828,6 +829,132 @@ def init_proc_mon(self):
...
+                    dmns_to_start += 1
+            else:
+                pbs_conf_val["PBS_START_MOM"] = "0"
+            if self.comms is None or len(self.comms) == 0:
+                if (pbs_conf_val["PBS_START_COMM"] != "1"):
+                    pbs_conf_val["PBS_START_COMM"] = "1"
+                    cmds_to_exec.append(["comm", "start"])
+                    dmns_to_start += 1
+            else:
+                pbs_conf_val["PBS_START_MOM"] = "0"
```
arungrover: why are we turning mom off when we are able to find comm?

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -828,6 +829,132 @@ def init_proc_mon(self):
...
+                pbs_conf_val["PBS_START_SCHED"] = "1"
+                cmds_to_exec.append(["sched", "start"])
+                dmns_to_start += 1
+            if self.moms is None or len(self.moms) == 0:
+                if (pbs_conf_val["PBS_START_MOM"] != "1"):
+                    pbs_conf_val["PBS_START_MOM"] = "1"
+                    cmds_to_exec.append(["mom", "start"])
+                    dmns_to_start += 1
+            else:
+                pbs_conf_val["PBS_START_MOM"] = "0"
```
arungrover: I am sure we have discussed it, but can you help me jog my memory about why are we disabling mom when self.moms is set. Also, don't you want to append mom stop to cmds_to_exec?

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -828,6 +829,132 @@ def init_proc_mon(self):
...
+            pbs_conf_val["PBS_SERVER"] = primary_server.hostname
+            if "PBS_SCP" not in pbs_conf_val:
+                scppath = self.du.which(mom.hostname, "scp")
+                if scppath != "scp":
+                    pbs_conf_val["PBS_SCP"] = scppath
+
+            self.du.set_pbs_config(mom.hostname, pbs_conf_filepath,
+                                   confs=pbs_conf_val)
+
+            if start_mom:
```
arungrover: If you change PBS_CORE_LIMITS or PBS_SERVER or PBS_SCP you would still want to restart mom

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -828,6 +829,132 @@ def init_proc_mon(self):
...
+
+            # Set PBS_CORE_LIMIT, PBS_SCP and PBS_SERVER
+            pbs_conf_val["PBS_CORE_LIMIT"] = "unlimited"
+            pbs_conf_val["PBS_SERVER"] = primary_server.hostname
+            if "PBS_SCP" not in pbs_conf_val:
+                scppath = self.du.which(server.hostname, "scp")
+                if scppath != "scp":
+                    pbs_conf_val["PBS_SCP"] = scppath
+
+            self.du.set_pbs_config(server.hostname, pbs_conf_filepath,
```
arungrover: If you change PBS_CORE_LIMITS or PBS_SERVER or PBS_SCP you would still want to restart all the daemons, isn't it?

test/tests/selftest/pbs_pbstestsuite.py
```diff
@@ -0,0 +1,132 @@
...
+        # Start PBS again
+        self.server.pi.start()
+
+        # Verify that the scheduler didn't come up
+        self.assertFalse(self.scheduler.isUp())
+
+        # Now call revert_pbsconf() & restart PBS again
+        self.server.pi.stop()
+        self.revert_pbsconf()
+        self.server.pi.start()
```
arungrover: Do you have to do this? isn't revert_pbsconf starting sched daemon automatically?

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -828,6 +829,132 @@ def init_proc_mon(self):
                                         frequency=freq)
                 self._process_monitoring = True
 
+    def revert_pbsconf(self):
+        """
+        Revert the following about the pbs.conf file:
```
arungrover: Since the function also restarts daemons when needed, It is better that we mention that in the docstring

test/fw/ptl/utils/pbs_testsuite.py
agrawalravi90: Thanks for catching that

test/fw/ptl/utils/pbs_testsuite.py
agrawalravi90: If self.moms is set, we are disabling the mom on the server host as the tests will pick up moms from the self.moms hosts. Does that make sense, or do you think that's a bad idea?

test/fw/ptl/utils/pbs_testsuite.py
agrawalravi90: I think just doing a HUP should suffice as it'll make them do pbs_loadconf() again and read the new values.

agrawalravi90:
@arungrover thanks I've addressed your comments, please provide further feedback

test/fw/ptl/utils/pbs_testsuite.py
arungrover: But doesn't PTL guarantee a mom on the local host to all the tests? If we do this then many tests will start failing

test/tests/selftest/pbs_pbstestsuite.py
```diff
@@ -105,28 +109,114 @@ def test_revert_pbsconf_remotemom(self):
...
+        # Now, set the core limit to 0 and restart PBS
+        self.server.pi.stop()
+        pbs_conf_val["PBS_CORE_LIMIT"] = "0"
+        self.du.set_pbs_config(fin=pbs_conf_filepath, confs=pbs_conf_val)
+        self.server.pi.start()
+
+        # First, check that there's no existing core file in mom_priv
+        parent_dir = os.getcwd()
+        snapdirs = []
+        with PBSSnapUtils(out_dir=parent_dir, acct_logs=0, daemon_logs=0,
```
arungrover: Why do we need to take snapshots? Can we not run an 'ls' command on the mom_priv directory?

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -828,6 +829,142 @@ def init_proc_mon(self):
...
+
+            if start_mom:
+                mom.pi.initd(mom.hostname, "start",
+                             pbs_conf_filepath, daemon="mom")
+            else:
+                # Send a HUP to mom
+                mom.signal("-HUP")
+
+        # Revert comm hosts' pbs.conf
+        for comm in self.comms.values():
```
bhroam: I just had a thought.  You might be setting the same pbs.conf multiple times if self.moms or self.comms share a common host.

test/tests/selftest/pbs_pbstestsuite.py
```diff
@@ -0,0 +1,222 @@
...
+    """
+
+    def test_revert_pbsconf_onehost(self):
+        """
+        Test the functionality of PBSTestSuite.revert_pbsconf()
+        for a single host type 1 installation
+        """
+        pbs_conf_filepath = self.du.get_pbs_conf_file(self.server.hostname)
+        pbs_conf_val = self.du.parse_pbs_config(self.server.hostname,
+                                                pbs_conf_filepath)
```
bhroam: I just checked, you don't need to call get_pbs_conf_file() before parse_pbs_conf.  If the second argument is not passed to parse_pbs_config(), it will default to calling get_pbs_conf_file(). (and other places in the code)

test/tests/selftest/pbs_pbstestsuite.py
```diff
@@ -0,0 +1,222 @@
...
+        # Verify that the scheduler didn't come up
+        self.assertFalse(self.scheduler.isUp())
+
+        # Now call revert_pbsconf() & restart PBS again
+        self.revert_pbsconf()
+
+        # Verify that the scheduler came up and start bit is 1
+        self.assertTrue(self.scheduler.isUp())
+        pbs_conf_val = self.du.parse_pbs_config(self.server.hostname)
+        self.assertEqual(pbs_conf_val["PBS_START_SCHED"], "1")
```
bhroam: Maybe you should check self.server.pbs_conf[]? (and other places in the test)

test/tests/selftest/pbs_pbstestsuite.py
```diff
@@ -0,0 +1,222 @@
...
+            self.skip_test("Test needs at least one remote Mom host,"
+                           " use -p moms=<hostname>")
+
+        pbs_conf_val = self.du.parse_pbs_config(self.server.hostname)
+        if pbs_conf_val is None or len(pbs_conf_val) < 1:
+            raise ValueError("Could not parse pbs.conf on host %s" %
+                             (server.hostname))
+
+        # Since the test requested for a mom, the server host's mom should not
+        # be ON, the rest should be set to 1
```
bhroam: Why does it matter if the server's mom is up or not?  We're checking a remote mom.

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -828,6 +829,142 @@ def init_proc_mon(self):
...
+                self.comm.signal("-HUP")
+                for sched in self.scheds.values():
+                    sched.signal("-HUP")
+
+        # Revert mom hosts' pbs.conf
+        for mom in self.moms.values():
+            start_mom = False
+            pbs_conf_filepath = self.du.get_pbs_conf_file(mom.hostname)
+            pbs_conf_val = self.du.parse_pbs_config(mom.hostname,
+                                                    pbs_conf_filepath)
```
bhroam: By doing this, you will keep any pbs.conf variables that are set during a test that are not the default values (e.g. PBS_LOG_HIGHRES_TIMESTAMP).  You need to create a dictionary of the variables you want, and set them.

test/tests/selftest/pbs_pbstestsuite.py
```diff
@@ -0,0 +1,222 @@
...
+
+        # Confirm that the mom came up and start bit is 1
+        self.assertTrue(remotemom.isUp())
+        pbs_conf_val = self.du.parse_pbs_config(remotemom.hostname)
+        self.assertEqual(pbs_conf_val["PBS_START_MOM"], "1")
+
+    def test_revert_pbsconf_corelimit(self):
+        """
+        Test the functionality of PBSTestSuite.revert_pbsconf() when
+        PBS_CORE_LIMIT is reverted
```
bhroam: Maybe: 'Test the functionality of PBSTestSuite.revert_pbsconf() when PBS_CORE_LIMIT is set to a value other than the default'

test/tests/selftest/pbs_pbstestsuite.py
```diff
@@ -0,0 +1,222 @@
...
+            snap_dir = snap_obj.capture_nodes(with_mom_logs=False)
+            snapdirs.append(snap_dir)
+            core_dir_path = os.path.join(snap_dir, CORE_MOM_PATH)
+            if os.path.isdir(core_dir_path):
+                # Delete all existing core files
+                core_filenames = os.listdir(core_dir_path)
+                for corefile in core_filenames:
+                    pbs_corepath = os.path.join(pbs_conf_val["PBS_HOME"],
+                                                "mom_priv", corefile)
+                    self.du.rm(path=pbs_corepath, force=True, sudo=True)
```
bhroam: core files are dropped in mom_priv.  This will delete everything in mom's priv directory (like her config file).  Also, I don't think we need to test this.  We aren't testing the mom functionality of dropping a core or not.  We are testing whether or not we revert pbs.conf.

test/tests/selftest/pbs_pbstestsuite.py
```diff
@@ -0,0 +1,222 @@
...
+        """
+        Test the functionality of PBSTestSuite.revert_pbsconf() when
+        PBS_CORE_LIMIT is reverted
+        """
+        pbs_conf_filepath = self.du.get_pbs_conf_file(self.server.hostname)
+        pbs_conf_val = self.du.parse_pbs_config(self.server.hostname,
+                                                pbs_conf_filepath)
+        if pbs_conf_val is None or len(pbs_conf_val) < 1:
+            raise ValueError("Could not parse pbs.conf on host %s" %
+                             (server.hostname))
```
bhroam: This is a test, don't raise exceptions.  Use asserts. (and other places in the test)

test/tests/selftest/pbs_pbstestsuite.py
agrawalravi90: we can, but how will we know which file is a core file?

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -828,6 +829,142 @@ def init_proc_mon(self):
...
+                scppath = self.du.which(server.hostname, "scp")
+                if scppath != "scp":
+                    pbs_conf_val["PBS_SCP"] = scppath
+
+            self.du.set_pbs_config(server.hostname, pbs_conf_filepath,
+                                   confs=pbs_conf_val)
+
+            if dmns_to_start == 4:
+                # Restart all
+                server.restart(server.hostname)
```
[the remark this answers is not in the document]
agrawalravi90: Thanks for catching that, I meant to use self.pi.restart(), i even provided it a hostname argument, server.restart() doesn't accept any arguments, I'm surprised that Codacy didn't catch this.

test/tests/selftest/pbs_pbstestsuite.py
agrawalravi90: I'm aware of that, the reason I'm doing it this way is so that it doesn't have to call get_pbs_conf_file() again when I set new pbs.conf attributes inside set_pbs_config()

test/tests/selftest/pbs_pbstestsuite.py
agrawalravi90: Since I'm reverting existing pbs.conf values, I didn't want to rely on the cache'd values from self.server.pbs.conf.

test/fw/ptl/utils/pbs_testsuite.py
agrawalravi90: hmm .. ok, this will need some refactoring, so I'll mark this WIP

test/tests/selftest/pbs_pbstestsuite.py
arungrover: I thought they all start with prefix core and if there are already existing core files in the 'priv' directory then daemons store the previous core files by appending an integer value to their name, like core.12434 will become core_001

test/tests/selftest/pbs_pbstestsuite.py
arungrover: Well, it does test that calling revert_pbsconf() makes the changes take effect in daemons. If we do not add this test then we don't know if daemons have been restarted and setup has correctly reverted it.

bhroam: Fair enough.  I can see the logic in that.  How do we know if the daemon has been restarted if we don't test the functionality that has been reverted.  The first part of my comment is still true though.  There are other files in mom_priv other than core files.  We should not delete them.

test/fw/ptl/utils/pbs_testsuite.py
agrawalravi90: actually, earlier my code was first turning the local mom off, and then bringing it back up when it parsed the self.moms list. I'm going to just check for it first now and not bring it down unnecessarily

test/tests/selftest/pbs_pbstestsuite.py
agrawalravi90: nope, I've seen core file names start with 'vgcore' or something completely different, it depends on how the system is configured: https://sigquit.wordpress.com/2009/03/13/the-core-pattern/

test/tests/selftest/pbs_pbstestsuite.py
agrawalravi90: that's exactly why I was doing it, I guess it won't be too much of an overhead so I'll do what you suggest

test/tests/selftest/pbs_pbstestsuite.py
agrawalravi90: oh yes, thanks for catching that, I'll update the cache

test/tests/selftest/pbs_pbstestsuite.py
agrawalravi90: We aren't deleting anything other than the core files. core_filenames literally is a list of just the core files, and we only delete them from mom_priv.

agrawalravi90:
@bhroam and @arungrover I've addressed your comments, made some changes so that the pbs.conf file is reverted to have only the default list of variables, please let me know what you guys think. Thanks.

test/tests/selftest/pbs_pbstestsuite.py
bhroam: Isn't core_filenames a listdir() of mom_priv?

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -828,6 +829,294 @@ def init_proc_mon(self):
...
+        :type primary_hot - str
+        :param hosttype - type of host being reverted
+        :type hosttype - str
+        :param hostobj - PTL object associated with the host
+        :type PBSService
+
+        :return default value of the pbs.conf variable if it can be determined
+        as a string, otherwise None
+        """
+        if conf == "PBS_SERVER":
```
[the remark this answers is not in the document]
agrawalravi90: the problem is what if the daemons share hosts. For example, if i create a pbs.conf for mom with start bit of only mom set to 1 and the rest set to 0, then it'll be bad if the mom shared that host with the comm or the server daemon.

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -828,6 +829,294 @@ def init_proc_mon(self):
...
+                    default_vals[conf] = pbs_conf_val[conf]
+                except KeyError:
+                    # existing pbs.conf doesn't have a default variable set
+                    # Try to determine the default value for it
+                    val = self._get_dflt_pbsconfval(conf,
+                                                    primary_server.hostname,
+                                                    "server", server)
+                    if val is None:
+                        print "Couldn't revert %s in pbs.conf" % (key)
+                        raise
```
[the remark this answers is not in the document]
agrawalravi90: when it is raised from the setUp, since we don't catch it, I'm guessing that the test execution will stop and we will bail out of pbs_benchpress. I think this is the right thing to do because if we continue on without a pristine pbs.conf then the results of test execution wouldn't be reliable.

test/tests/selftest/pbs_pbstestsuite.py
agrawalravi90: hmm .. I can see what you are trying to say, why not just go for the simple approach as it might be good enough .. since this is just a PTL test which will most probably run in a controlled environment anyways, I'll change it to the way you are suggesting

test/tests/selftest/pbs_pbstestsuite.py
agrawalravi90: no, it is a listdir on "core_file_bt/mom_priv", that directory only contains backtraces from core files. I'm going to take Arun's advice and just look for files which start with 'core' instead of taking a snapshot, that should make the test look much more straight-forward

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -828,6 +829,294 @@ def init_proc_mon(self):
...
+            else:
+                if (default_vals["PBS_START_COMM"] != "1"):
+                    default_vals["PBS_START_COMM"] = "1"
+                    cmds_to_exec.append(["comm", "start"])
+                    dmns_to_start += 1
+
+            # Set PBS_CORE_LIMIT, PBS_SCP and PBS_SERVER
+            if default_vals["PBS_CORE_LIMIT"] != "unlimited":
+                default_vals["PBS_CORE_LIMIT"] = "unlimited"
+                dmns_to_start = 4
```
anamikau: why are you setting this to 4 explicitly? what if a server do not have a mom on it?

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -840,128 +888,232 @@ def revert_pbsconf(self):
...
         # Revert server hosts' pbs.conf
         for server in self.servers.values():
+            default_vals = none_vals
             cmds_to_exec = []
             dmns_to_start = 0
             pbs_conf_filepath = self.du.get_pbs_conf_file(server.hostname)
             pbs_conf_val = self.du.parse_pbs_config(server.hostname,
                                                     pbs_conf_filepath)
-            if pbs_conf_val is None or len(pbs_conf_val) < 1:
+            if not pbs_conf_val or len(pbs_conf_val) < 1:
```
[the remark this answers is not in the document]
agrawalravi90: I didn't know that an empty list is interpreted as None/False by Python, thanks for the tip

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -840,128 +888,232 @@ def revert_pbsconf(self):
...
             pbs_conf_val = self.du.parse_pbs_config(mom.hostname,
                                                     pbs_conf_filepath)
-            if pbs_conf_val is None or len(pbs_conf_val) < 1:
+            if not pbs_conf_val or len(pbs_conf_val) < 1:
                 raise ValueError("Could not parse pbs.conf on host %s" %
                                  (mom.hostname))
 
+            # to start with, set all keys in default_vals with values from the
+            # existing pbs.conf
+            for conf in default_vals:
```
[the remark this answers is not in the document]
agrawalravi90: This is so that we only restart the daemons if necessary. If the existing values are correct then we won't restart the daemons.

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -840,128 +888,232 @@ def revert_pbsconf(self):
...
+                        # start/stop the particular daemon
+                        server.pi.initd(server.hostname, initcmd[1],
+                                        pbs_conf_filepath, daemon=initcmd[0])
 
         # Revert mom hosts' pbs.conf
         for mom in self.moms.values():
+            if mom.hostname == primary_server.hostname:
+                continue
+
+            default_vals = none_vals
```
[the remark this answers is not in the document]
agrawalravi90: Thanks for catching this!

test/fw/ptl/utils/pbs_testsuite.py
agrawalravi90: 4 is just a way of expressing "restart all daemons on the server". I could have used another flag for this but I didn't think that it was needed. Please let me know if you think that I shouldn't overload dmns_to_start and use another flag instead.

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -902,30 +898,30 @@ def revert_pbsconf(self):
...
         for server in self.servers.values():
-            default_vals = none_vals
+            default_vals = dict(none_vals)
             cmds_to_exec = []
             dmns_to_start = 0
-            pbs_conf_filepath = self.du.get_pbs_conf_file(server.hostname)
-            pbs_conf_val = self.du.parse_pbs_config(server.hostname,
-                                                    pbs_conf_filepath)
-            if not pbs_conf_val or len(pbs_conf_val) < 1:
+            self.du.get_pbs_conf_file(server.hostname)
```
bhroam: Why are you calling this?  It returns the conf file name.  By calling it like this, it does nothing.  parse_pbs_config() will call it itself.

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -989,36 +984,36 @@ def revert_pbsconf(self):
...
                 continue
 
-            default_vals = none_vals
+            default_vals = dict(none_vals)
             start_mom = False
-            pbs_conf_filepath = self.du.get_pbs_conf_file(mom.hostname)
-            pbs_conf_val = self.du.parse_pbs_config(mom.hostname,
-                                                    pbs_conf_filepath)
-            if not pbs_conf_val or len(pbs_conf_val) < 1:
+            self.du.get_pbs_conf_file(mom.hostname)
```
bhroam: Same as above, this doesn't do anything.

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -828,6 +829,282 @@ def init_proc_mon(self):
...
+            # Check if existing pbs.conf has more/less entries than the
+            # default list
+            if len(pbs_conf_val) != len(default_vals):
+                start_mom = True
+
+            if start_mom:
+                self.du.set_pbs_config(mom.hostname, confs=default_vals,
+                                       append=False)
+                mom.pbs_conf = default_vals
+                mom.pi.initd(mom.hostname, "start", daemon="mom")
```
bhroam: Don't you want to restart the mom?

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -828,6 +829,282 @@ def init_proc_mon(self):
...
+
+            # Check if existing pbs.conf has more/less entries than the
+            # default list
+            if len(pbs_conf_val) != len(default_vals):
+                start_comm = True
+
+            if start_comm:
+                self.du.set_pbs_config(comm.hostname, confs=default_vals)
+                comm.pbs_conf = default_vals
+                comm.pi.initd(comm.hostname, "start", daemon="comm")
```
bhroam: Don't you want to restart the comm?

bhroam:
Travis is failing in revert_pbsconf().  You should look into it.

test/fw/ptl/utils/pbs_testsuite.py
agrawalravi90: Actually i ended up refactoring it, please let me know if it's still confusing, thanks!

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -1012,98 +976,167 @@ def revert_pbsconf(self):
...
+                        # start/stop the particular daemon
+                        server.pi.initd(server.hostname, initcmd[1],
+                                        daemon=initcmd[0])
+
+    def revert_pbsconf(self):
+        """
+        Revert contents of the pbs.conf file
+        Also start/stop the appropriate daemons
+        """
+        primary_server = self.servers.values()[0]
```
bhroam: Is this the primary server, or is self.server the primary server?

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -1012,98 +976,167 @@ def revert_pbsconf(self):
...
+
+        self._revert_pbsconf_server(primary_server, vals_to_set)
+
+        self._revert_pbsconf_server(primary_server, vals_to_set)
 
-            if start_comm:
-                self.du.set_pbs_config(comm.hostname, confs=default_vals)
-                comm.pbs_conf = default_vals
-                comm.pi.initd(comm.hostname, "start", daemon="comm")
+        self._revert_pbsconf_server(primary_server, vals_to_set)
```
bhroam: Why do you call this 3 times?  I don't see any difference between the 3 calls.

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -911,10 +912,13 @@ def _revert_pbsconf_comm(self, primary_server, vals_to_set):
                     if val is None:
                         self.logger.error("Couldn't set %s in pbs.conf" %
                                           (conf))
-                        del(new_pbsconf[conf])
+                        keys_to_delete.append(new_pbsconf[conf])
                     else:
                         new_pbsconf[conf] = val
 
+            for key in keys_to_delete:
+                del(new_pbsconf[key])
```
bhroam: I don't understand this.  How is this different than deleting the keys as you see them?

agrawalravi90: Python gets mad at you if you delete elements from a dict while iterating on it:
RuntimeError: dictionary changed size during iteration

test/fw/ptl/utils/pbs_testsuite.py
agrawalravi90: If you look at init_servers(), self.server is actually set to servers.values()[0]:
        if cls.servers:
            cls.server = cls.servers.values()[0]

But I think it's probably safer to use self.server as the primary server, so I'll make that change, thanks.

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -1012,98 +976,167 @@ def revert_pbsconf(self):
...
+                dmns_to_start += 1
+                cmds_to_exec.append(["server", "start"])
+            if (new_pbsconf["PBS_START_SCHED"] != "1"):
+                new_pbsconf["PBS_START_SCHED"] = "1"
+                cmds_to_exec.append(["sched", "start"])
+                dmns_to_start += 1
+            if self.moms and server.hostname not in self.moms:
+                if new_pbsconf["PBS_START_MOM"] != "0":
+                    new_pbsconf["PBS_START_MOM"] = "0"
+                    cmds_to_exec.append(["mom", "stop"])
```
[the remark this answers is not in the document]
agrawalravi90: Actually, PTL's pi.init.d takes care of this. It creates a temporary file with only the start bit of the relevant daemon set to 1, then calls init script with the operation requested (start/stop) and passes PBS_CONF_FILE set to the temp file. This leaves the original pbs.conf file untouched and still does the right thing with starting/stopping the correct daemon. @saritakh did a great job on this code.

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -828,6 +829,327 @@ def init_proc_mon(self):
...
+                if conf in pbs_conf_val:
+                    new_pbsconf[conf] = pbs_conf_val[conf]
+                else:
+                    # existing pbs.conf doesn't have a default variable set
+                    # Try to determine the default
+                    val = self._get_dflt_pbsconfval(conf,
+                                                    primary_server.hostname,
+                                                    "comm", comm)
+                    if val is None:
+                        self.logger.error("Couldn't set %s in pbs.conf" %
```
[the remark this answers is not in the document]
agrawalravi90: > Using couldn't feels like that the program tried but it couldn't make it

Well, that actually is what's happening though, we try to find the default value for this variable in _get_dflt_pbsconfval(), but if that returns None, we say that we tried but couldn't figure out how to do it. Does it make sense now?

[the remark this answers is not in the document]
arungrover: Well, I'm ok if you want to retain it the way it is now. I thought, in this case, the log could be that it couldn't figure out a default value for the configuration. 

To me, the current log message just feels like a thing to say when setup tried putting the configuration in the conf file but couldn't do so because of factors like unable to open the file or file not present or write failed etc.

test/fw/ptl/utils/pbs_testsuite.py
bhroam: Now that you say this, it makes perfect sense.  Changing the list/dict you are iterating sounds like a really bad idea.

test/fw/ptl/utils/pbs_testsuite.py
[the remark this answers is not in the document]
bhroam: Oh cool.  I didn't know that.

test/fw/ptl/utils/pbs_testsuite.py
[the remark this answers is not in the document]
agrawalravi90: ok, i see what you are saying, alright I'll change it to say "Couldn't revert %s in pbs.conf to its default value"

agrawalravi90:
@arungrover and @bhroam Thanks for all the comments, I've addressed them. I also, finally, ran the remote mom PTL test and fixed some minor issues to make it work. Please let me know if things look okay. Latest test logs attached in the description.

test/fw/ptl/utils/pbs_testsuite.py
anamikau: Thanks @agrawalravi90. I looked at the updated  code and I am clear now :)

agrawalravi90:
Thanks @bhroam 
@arungrover please let me know if things look okay to you as well. Thanks!

arungrover:
looks good @agrawalravi90 please squash and rebase

agrawalravi90:
Thanks a lot @arungrover , I've rebased, hopefully the CI will pass, please merge this in if possible. Thanks again.

PR #842, opened 2018-10-13
Job Performance tests for PBS

visheshh:
#### Bug/feature Description
Job Performance tests for PBS to measure job submission time , job run_rate, job throughput, job deletion time , job stat time , server restart time .

#### Affected Platform(s)
all

#### Solution Description
Addition of performance tests

#### Testing logs/output
[output.txt](https://github.com/PBSPro/pbspro/files/2940779/output.txt)
Latest test logs:
[ptl_output.txt](https://github.com/PBSPro/pbspro/files/3047049/ptl_output.txt)
Latest test logs - Apr 11th:
[ptl_output_latest_run.txt](https://github.com/PBSPro/pbspro/files/3068303/ptl_output_latest_run.txt)

#### Checklist:

- [x] I have joined the **[pbspro community forum](http://community.pbspro.org/)**.
- [x] My pull request contains a **single, signed** commit. See **[setting up gpg signature](https://pbspro.atlassian.net/wiki/display/DG/Signing+Your+Git+Commits)**.
- [ ] My code follows the **[coding style](https://pbspro.atlassian.net/wiki/display/DG/Coding+Standards)** of this project.
- [ ] My change requires project documentation. See **[required documentation checklist](https://pbspro.atlassian.net/wiki/display/DG/Checklist+for+Developing+Features+and+Bug+Fixes)** for details.
   - [ ] I have added documentation in the **[project documentation area](https://pbspro.atlassian.net/wiki/display/PD)**.
- [x] I have added new **PTL test(s) to my commit**. (See **[using PTL for testing](https://pbspro.atlassian.net/wiki/display/DG/Using+PTL+for+Testing)**) *(or)*
   - [ ] I have added  **manual test(s) to this pull request and explained why PTL is not appropriate** for this case.
- [x] All new and existing automated tests have passed. (See **[running automated PTL tests](https://pbspro.atlassian.net/wiki/display/DG/PTL+Quick+Start+Guide)**).
- [x] I have attached **test logs to this pull request** as evidence of testing/verification.

__***For further information please visit the [Developer Guide Home](https://pbspro.atlassian.net/wiki/display/DG/Developer+Guide+Home).***__

PR #893, opened 2018-11-22
Disable Nagles Algorithm in scheduler

sakshamgarg:
#### Bug/feature Description
* *Disable Nagles algorithm for latency in scheduler*

#### Affected Platform(s)
* *ALL*

* *When submitting jobs it was noticed that due to nagle's algorithm, the submission rate could drop.*
* *Nagle's algorithm works by combining a number of small outgoing packets and sending them all at once, but enabling this algorithm is not beneficial in scheduler.*

#### Solution Description
* *Calling "set_nodelay()" function to set no-delay option (TCP_NODELAY) on connection.*

#### Testing logs/output
* [disable_nagles_Algo_test.txt](https://github.com/PBSPro/pbspro/files/2623004/disable_nagles_Algo_test.txt)
* Test case scenario:
Testing job completion performance from multiple users with scheduling on. Submitting 100 jobs from each user with a total of 10 users for 10 iterations.
Comparing throughput between the master branch and the PR branch:
 Master Branch throughput: 54
 PR Branch throughput: 57

#### Checklist:

### DIFF
--- a/src/include/port_forwarding.h
+++ b/src/include/port_forwarding.h
@@ -85,7 +85,6 @@ struct pfwdsock
 };
 /*Functions available in port_forwarding.h*/
 void port_forwarder(struct pfwdsock *, int(*connfunc)(char *phost, long pport), char *, int , int inter_read_sock, int (*readfunc)(int), void (*logfunc)(char *));
-int set_nodelay(int);
 int connect_local_xsocket(u_int);
 int x11_connect_display(char *, long);
 int set_nonblocking(int);
--- a/src/scheduler/pbs_sched.c
+++ b/src/scheduler/pbs_sched.c
@@ -567,6 +567,12 @@ server_command(char **jid)
 		return SCH_ERROR;
 	}
 
+	if (set_nodelay(new_socket) == -1) {
+		snprintf(log_buffer, sizeof(log_buffer), "cannot set nodelay on primary socket connection %d (errno=%d)\n", new_socket, errno);
+		log_err(-1, __func__, log_buffer);
+		return SCH_ERROR;
+	}
+
 	if (ntohs(saddr.sin_port) >= IPPORT_RESERVED) {
 		badconn("non-reserved port");
 		close(new_socket);
@@ -626,6 +632,12 @@ server_command(char **jid)
 			log_err(errno, __func__,
 				"warning: failed to get second_connection");
 			return cmd; /* bail out early */
+		}
+
+		if (set_nodelay(second_connection) == -1) {
+			snprintf(log_buffer, sizeof(log_buffer), "cannot set nodelay on secondary socket connection %d (errno=%d)\n", second_connection, errno);
+			log_err(-1, __func__, log_buffer);
+			return cmd;
 		}
 
 		if (ntohs(saddr.sin_port) >= IPPORT_RESERVED) {
--- a/src/server/run_sched.c
+++ b/src/server/run_sched.c
@@ -319,6 +319,15 @@ contact_sched(int cmd, char *jobid, pbs_net_t pbs_scheduler_addr, unsigned int p
 
 	net_add_close_func(sock, scheduler_close);
 
+	if (set_nodelay(sock) == -1) {
+#ifdef WIN32
+		errno = WSAGetLastError();
+#endif
+		snprintf(log_buffer, sizeof(log_buffer), "cannot set nodelay on connection %d (errno=%d)\n", sock, errno);
+		log_err(-1, __func__, log_buffer);
+		return (-1);
+	}
+
 	/* send command to Scheduler */
 
 	if (put_sched_cmd(sock, cmd, jobid) < 0) {
--- a/src/lib/Libcmds/cnt2server.c
+++ b/src/lib/Libcmds/cnt2server.c
@@ -108,16 +108,6 @@ cnt2server_extend(char *server, char *extend)
 		return (connect);
 	}
 
-	/*
-	 * Disable Nagle's algorithm on the TCP connection to server.
-	 * Nagle's algorithm is hurting cmd-server communication.
-	 */
-	if (pbs_connection_set_nodelay(connect) == -1) {
-		fprintf(stderr, "Cannot set nodelay on connection %d (errno=%d)\n",
-			connect, pbs_errno);
-		return (PBSE_SYSTEM);
-	}
-
 	return (connect);
 }
 
--- a/src/lib/Libifl/pbsD_connect.c
+++ b/src/lib/Libifl/pbsD_connect.c
@@ -803,6 +803,17 @@ __pbs_connect_extend(char *server, char *extend_data)
 	DIS_tcp_setup(connection[out].ch_socket);
 	pbs_tcp_timeout = PBS_DIS_TCP_TIMEOUT_VLONG;	/* set for 3 hours */
 
+	/*
+	 * Disable Nagle's algorithm on the TCP connection to server.
+	 * Nagle's algorithm is hurting cmd-server communication.
+	 */
+	if (pbs_connection_set_nodelay(out) == -1) {
+		CLOSESOCKET(connection[out].ch_socket);
+		connection[out].ch_inuse = 0;
+		pbs_errno = PBSE_SYSTEM;
+		return -1;
+	}
+
 	return out;
 }
 
--- a/src/scheduler/pbs_sched_win.c
+++ b/src/scheduler/pbs_sched_win.c
@@ -544,6 +544,15 @@ server_command(char **jid)
 		return SCH_ERROR;
 	}
 
+	if (set_nodelay(new_socket) == -1) {
+#ifdef WIN32
+		errno = WSAGetLastError();
+#endif
+		snprintf(log_buffer, sizeof(log_buffer), "cannot set nodelay on primary socket connection %d (errno=%d)\n", new_socket, errno);
+		log_err(-1, __func__, log_buffer);
+		return SCH_ERROR;
+	}
+
 	if (ntohs(saddr.sin_port) >= IPPORT_RESERVED) {
 		badconn("non-reserved port");
 #ifdef WIN32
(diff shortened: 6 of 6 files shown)


agrawalravi90:
"
    When submitting jobs it was noticed that depending on the size of the job packet sent from qsub the submission rate could drop.
    It was determined that the nagles algorithm was to blame for this issue.
"

Could you please mention some numbers? what was the submission rate before your fix? What kind of drop in the submission rate did you notice? What was the improvement after your fix?

Also, when you say  "depending on the size of the job packet", could you again mention what kind of packet size causes the said drop in submission rate?

src/cmds/qsub.c
```diff
@@ -4440,6 +4440,11 @@ do_connect(char *server_out, char *retmsg)
 		return (pbs_errno);
 	}
 
+	if (pbs_connection_set_nodelay(sd_svr) == -1) {
```
suresh-thelkar: Calling of the function pbs_connection_set_nodelay here is not required because we already call this function inside cnt2server_extend.

src/scheduler/pbs_sched.c
```diff
@@ -568,7 +568,7 @@ server_command(char **jid)
 	}
 
 	if (set_nodelay(new_socket) == -1) {
-		sprintf(log_buffer, "cannot set nodelay on primary socket connection %d (errno=%d)\n", new_socket, errno);
+		snprintf(log_buffer, sizeof(log_buffer), "cannot set nodelay on primary socket connection %d (errno=%d)\n", new_socket, errno);
```
suresh-thelkar: In order to get errno in case of Windows OS you need to use WSAGetLastError. So wherever you are expecting/logging errno better to use #ifdef WIN32 for Windows Code. Please apply this comment wherever it is needed. 
Also please make sure that set_nodelay() function indeed works for Windows also. If the current definition does not work we need to modify it accordingly.

mike0042:
Please run full regressions against your branch and post logs for both Linux and Windows.

subhasisb:
Hi @sakshamgarg - i think the pbs_connection_set_nodelay function is already defined in pbs_internal.h. I am not sure we have a need to expose the function to the external world via defining it in pbs_ifl.h. In fact, if we do that we will need to bump the SONAME (also since the function is called internally from IFL connect automatically, we do not need to expose it).

sakshamgarg:
> Hi @sakshamgarg - i think the pbs_connection_set_nodelay function is already defined in pbs_internal.h. I am not sure we have a need to expose the function to the external world via defining it in pbs_ifl.h. In fact, if we do that we will need to bump the SONAME (also since the function is called internally from IFL connect automatically, we do not need to expose it).

Hi @subhasisb , yeah you are right. "pbs_connection_set_nodelay" function is already defined in pbs_internal.h and there seems no need for defining the same function in pbs_ifl.h . 
I will remove it from pbs_ifl.h asap.

sakshamgarg:
Hi @agrawalravi90 , I am working on testing the performance. I will submit the results asap.

sakshamgarg:
Hi @agrawalravi90. I did ran the performance test on a development machine.
Test case scenario:
Testing job completion performance from multiple users with scheduling on. Submitting 100 jobs from each user with a total of 10 users for 10 iterations.

Comparing throughput between the master branch and the PR branch:
Master Branch throughput: 54
PR Branch throughput: 57

agrawalravi90:
@sakshamgarg how did you measure the throughput?

also, looking at the code more closely, we had already disabled Nagle's algorithm for command to server connections, so qsub should already be fast, so can you please help me understand why disabling Nagle's algorithm for server-scheduler connectivity will affect the job submission rate?

src/include/port_forwarding.h
```diff
@@ -85,7 +85,6 @@ struct pfwdsock
 };
 /*Functions available in port_forwarding.h*/
 void port_forwarder(struct pfwdsock *, int(*connfunc)(char *phost, long pport), char *, int , int inter_read_sock, int (*readfunc)(int), void (*logfunc)(char *));
-int set_nodelay(int);
```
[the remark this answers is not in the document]
mike0042: Because it was also being declared in net_connect.h

sakshamgarg:
Hi @agrawalravi90 , I have used the job performance test case which is there in the following PR [https://github.com/PBSPro/pbspro/pull/842](url). This test case calculates throughput by dividing total number of jobs by total time taken. The time difference is from first enqueue to last dequeue.

I agree with you that change in the server-scheduler connectivity will not affect the job submission rate, although it will affect the throughput.

agrawalravi90:
Thanks for explaining @sakshamgarg . I'm happy with your explanation and the code changes as such, but I feel that we should either wait for the other PR to be checked-in, or check-in the test which you have used to measure performance improvement with this PR itself. What do the other @PBSPro/pbspro-maintainers think?

subhasisb:
One challenge with the test is that the results can be very intermittent since this is to do with internal buffering inside TCP. One try we can do is do a bunch of qstats on a system that shows a difference - but not all systems will.

agrawalravi90:
If the results are intermittent, are we really sure that this will improve performance?

agrawalravi90:
Alright, thanks for explaining, I guess it generally make sense to disable Nagle's algorithm for applications like PBS where the daemons can be very chatty. I'm ok with checking this in without a test.

agrawalravi90:
@sakshamgarg please rebase in preparation for merge

sakshamgarg:
@agrawalravi90 I have rebased and squashed my commits. Although I am running full regression as suggested by Mike. It will get finished by tomorrow. After the analysis, it will get ready for merge.

agrawalravi90:
Alright, please let us know when you are ready.

src/scheduler/pbs_sched.c
```diff
@@ -567,6 +567,15 @@ server_command(char **jid)
 		return SCH_ERROR;
 	}
 
+	if (set_nodelay(new_socket) == -1) {
+#ifdef WIN32
```
arungrover: why are we adding windows specific code in pbs_sched.c when there is pbs_sched_win.c for windows already present. I think you don't need it here.

src/server/run_sched.c
```diff
@@ -319,6 +319,15 @@ contact_sched(int cmd, char *jobid, pbs_net_t pbs_scheduler_addr, unsigned int p
 
 	net_add_close_func(sock, scheduler_close);
 
+	if (set_nodelay(sock) == -1) {
+#ifdef WIN32
+		errno = WSAGetLastError();
+#endif
+		snprintf(log_buffer, sizeof(log_buffer), "run_sched: cannot set nodelay on connection %d (errno=%d)\n", sock, errno);
```
arungrover: I don't think you need the file name in the log message here. \_\_func\_\_ will log the function name and that's what we need.

src/scheduler/pbs_sched.c
sakshamgarg: Yeah, you are right. I will remove it.

anamikau:
Hello All, I am sorry for not catching up on this early. I understand that there cannot be a test to test exactly the change but we can test around it. Also for internal purposes we use dedicated hosts to compare the results and do benchmarking. I looked into our performance and resilience folder and I could not find a test for following scenario. I request to add a test for following. Looking at @subhasisb explanation results cannot be improved but they should not be showing negative performance (Is -1% - -3% okay). Right?

"Testing job completion performance from multiple users with scheduling on. Submitting 100 jobs from each user with a total of 10 users for 10 iterations."

Also please run pbs_equiv_classes_perf.py and pbs_client_nagle_performance.py with and without fix and post the comparison results.

sakshamgarg:
Hi all, I have successfully completed the full regression and the analysis is also done. I have rebased and squashed all my commits.

agrawalravi90:
Thanks @sakshamgarg , let @anamikau confirm that she's okay with your testing approach, we can merge this right after. Thanks for being patient.

anamikau:
> Hi @anamikau. For testing the performance of the changes in the current PR, I ran "test_submit_jobs_sched_on" test which is raised in the PR [https://github.com/PBSPro/pbspro/pull/842](url) as I mentioned in [#893 (comment)](https://github.com/PBSPro/pbspro/pull/893#issuecomment-444003315). Since this test case's PR is ongoing, there is no need for duplicating the work.
> 
okay thank you for pointing it out.

> In the following comment ( [#893 (comment)](https://github.com/PBSPro/pbspro/pull/893#issuecomment-443486030) ), I have posted the throughput results which were obtained by running the above test case. The results show positive increase in the performance. As far as I know, in the worst case scenario, the performance can be equivalent to that before this PR but not negative. I also want to know, how did you infer that the performance would be negative ?
> 
Because it is very system dependent it is quite possible you get 1 or 2 secs less some time instead of exact.
> Also I believe that running pbs_client_nagle_performance.py test is not necessary as it is related to client-server communication and my changes are related to server-scheduler communication.

okay I will go with you on that. can you provide the execution results for "pbs_equiv_classes_perf.py"? I do not see them attached under the testing summary.
Also a request, please attach all test related data under testing summary as it is kind of hard to go over each comment specially if there are many.

agrawalravi90:
@sakshamgarg I noticed that you are doing forced pushes to this PR, if you are making any code changes then please push separate commits so that we can see what code you changed.

sakshamgarg:
@agrawalravi90 I haven't changed any part of the code. I did a rebase that is why there is a forced commit.

agrawalravi90:
Ok, thanks for clarifying. Are you still in the process of addressing Anamika's comments?

sakshamgarg:
Hi @agrawalravi90 , I ran the test suggested by Anamika, but they are getting timed out.
I have talked to @anamikau  about this and I will be locking the bug. Other than that, I guess everything is solved.

agrawalravi90:
Thanks for the update saksham.
@anamikau are you happy with this PR?

anamikau:
Hi @agrawalravi90 and @sakshamgarg, please go forward with this PR. We need to run the tests as part of regressions later.

sakshamgarg:
Thanks @anamikau for your input. I will rebase this PR.

subhasisb:
On second thoughts, i think we should check if we can put the nodelay call inside pbs_connect_extend itself, so that it affects in every connection and not just the ones going via cnt2server.

agrawalravi90:
@sakshamgarg , did you get a chance to address Subhasis's comment?

sakshamgarg:
Hi @agrawalravi90. I am working on it. Will address it as soon as I am done.

sakshamgarg:
I have squashed my commits and rebased.

PR #895, opened 2018-11-22
qrun hangs indefinitely

SudeshnaMoulik:
#### Bug/feature Description
- If we submit "qrun jobid" immediately after disabling "scheduling = 0", that request may processed in existing scheduling cycle. In case the job was unable to run because of some reason (like Not Running: Not enough free nodes available) then qrun hangs.

#### Affected Platform(s)
-All Linux platform OS

- After job comment modification, another scheduler cycle is not started as we are checking there whether "scheduling" is enabled.

#### Solution Description
- If there are any qrun request in server deffered list, it should start another scheduling cycle, irrespective of "scheduling" is enabled.

#### Testing logs/output

#### Checklist:

### DIFF
--- a/test/tests/functional/pbs_qrun.py
+++ b/test/tests/functional/pbs_qrun.py
@@ -36,6 +36,8 @@
 # trademark licensing policies.
 
 from tests.functional import *
+import os
+import signal
 
 
 class TestQrun(TestFunctional):
@@ -92,3 +94,48 @@ class TestQrun(TestFunctional):
         msg = "Server is not up"
         self.assertTrue(self.server.isUp(), msg)
         self.logger.info("As expected server is up and running")
+
+    def test_qrun_hangs(self):
+        """
+        This test submit 500 jobs with differnt equivalence class,
+        turn of scheduling and qrun job to
+        verify whether qrun hangs.
+        """
+        node = self.mom.shortname
+        self.server.manager(MGR_CMD_SET, SCHED,
+                            {'scheduling': 'False'})
+        self.server.manager(MGR_CMD_SET, NODE,
+                            {'resources_available.ncpus': 1}, id=node)
+        for walltime in range(1, 501):
+            j = Job(TEST_USER)
+            a = {'Resource_List.walltime': walltime}
+            j.set_attributes(a)
+            if walltime == 500:
+                jid = self.server.submit(j)
+            else:
+                self.server.submit(j)
+        self.logger.info("Submitted 500 jobs with different walltime")
+        now = int(time.time())
+        self.server.manager(MGR_CMD_SET, SERVER,
+                            {'scheduling': 'True'})
+        self.server.manager(MGR_CMD_SET, SERVER,
+                            {'scheduling': 'False'}, expect=True)
+        pid = os.fork()
+        if pid == 0:
+            try:
+                self.server.runjob(jobid=jid)
+                self.logger.info("Successfully runjob. Child process exit.")
+            except PbsRunError as e:
+                self.logger.info("Runjob throws error: " + e.msg[0])
+        else:
+            out = self.scheduler.log_match(
+                "Starting Scheduling Cycle",
+                allmatch=True, interval=5, starttime=now, max_attempts=10)
+            if len(out) < 2:
+                os.kill(pid, signal.SIGKILL)
+                os.waitpid(pid, 0)
+                self.logger.info("Runjob hung. Child process exit.")
+                self.assertFalse(True,
+                                 "Qrun didn't start another sched cycle")
+            else:
+                self.logger.info("No hangs. Parent process exit")
--- a/src/server/pbsd_main.c
+++ b/src/server/pbsd_main.c
@@ -2097,8 +2097,7 @@ try_db_again:
 
 					psched->sch_next_schedule = time_now +
 							psched->sch_attr[(int)	SCHED_ATR_schediteration].at_val.at_long;
-					if (psched->sch_attr[SCHED_ATR_scheduling].at_val.at_long &&
-							(schedule_jobs(psched) == 0) && (svr_unsent_qrun_req))
+					if ((schedule_jobs(psched) == 0) && (svr_unsent_qrun_req))
 						svr_unsent_qrun_req = 0;
 				}
 			}


SudeshnaMoulik:
qrun hangs indefinitely. Verified the code changes via following script:

`
while true
do
   qmgr -c "s n $(hostname) resources_available.ncpus=1"
   j1=$(qsub – /bin/date)
   j2=$(qsub – /bin/date)
   qmgr -c "s s scheduling=1"
   qmgr -c "s s scheduling=0"
   qmgr -c "s n $(hostname) resources_available.ncpus=2"
   qrun $j2
done
`
Smoketest and Jenkins build are successful.

mike0042:
Please attach test logs to the PR.

bhroam:
You can write a PTL test for this.  You can set an alarm to see if the qrun hangs.

SudeshnaMoulik:
Thanks @bhroam @mike0042 
I will attach the PTL testcase in my next commit.

SudeshnaMoulik:
Please find the PTL logs below from openSuse platform, with/without fix.

[PTL_qrun_hangs_without_fix_openSuse.txt](https://github.com/PBSPro/pbspro/files/2628192/PTL_qrun_hangs_without_fix_openSuse.txt)
[PTL_qrun_noHangs_with_fix_openSuse.txt](https://github.com/PBSPro/pbspro/files/2628195/PTL_qrun_noHangs_with_fix_openSuse.txt)

Note:
I have rebased my branch as appveyor check is failing with following note:
"AppVeyor was unable to build non-mergeable pull request"

test/tests/functional/pbs_qrun.py
```diff
@@ -36,6 +36,7 @@
 # trademark licensing policies.
 
 from tests.functional import *
+import time
```
bhroam: You don't need to import time.  PTL does that for you.

test/tests/functional/pbs_qrun.py
```diff
@@ -92,3 +93,39 @@ def test_invalid_host_val(self):
...
+                                {'resources_available.ncpus': 1}, id=node)
+            j1 = Job(TEST_USER)
+            j1.create_script('/bin/date')
+            self.server.submit(j1)
+            jid2 = self.server.submit(j1)
+
+            self.server.manager(MGR_CMD_SET, SCHED,
+                                {'scheduling': 'True'})
+            self.server.manager(MGR_CMD_SET, SCHED,
+                                {'scheduling': 'False'})
```
bhroam: Add expect=True.  It will wait for the current cycle to end before returning.

test/tests/functional/pbs_qrun.py
```diff
@@ -92,3 +93,39 @@ def test_invalid_host_val(self):
...
+                                         starttime=self.server.ctime,
+                                         max_attempts=10)
+            except PtlLogMatchError:
+                self.assertFalse(True,
+                                 "New scheduling cycle not started.Qrun hangs")
+            else:
+                time2 = int(time.time())
+                if (time2 - time1) > 60:
+                    self.logger.info("Loop ran since 1 min,no hang noticed")
+                    break
```
bhroam: I don't think this will work.  First, the runjob will call qrun.  Won't that hang?  Also, you set scheduling to true at the top.  This should run a cycle regardless of qrun.  The log_match() should match that every time since you set starttime=self.server.ctime

test/tests/functional/pbs_qrun.py
SudeshnaMoulik: ok will address this in my next commit.

test/tests/functional/pbs_qrun.py
SudeshnaMoulik: Actually as per the shell script, the bug is reproducible only when that qrun job is processed in existing cycle and there are no free nodes available to run the job.
So, i guess we need the current cycle on-going, when qrun process the job.

test/tests/functional/pbs_qrun.py
SudeshnaMoulik: starttime=self.server.ctime => Used this to indicate the log_match should start matching from the current server time. Is there any other variable/parameter I should use?

Regarding the "scheduling", at first it set to true to get a new cycle and then set to false, as the script expect qrun to trigger the scheduling cycle.

test/tests/functional/pbs_qrun.py
bhroam: That gets tricky.  Since there is no real way to force the scheduler to stay in cycle, you'll need to submit a ton of jobs to cause the scheduler to do a lot of work.  They need to be different jobs or the scheduler will just ignore them after it realized it can't run one in the job equivalence class.

test/tests/functional/pbs_qrun.py
bhroam: self.server.ctime is the time that the server object was created.  If I understand this correctly, this should be the time the test suite started.  This means any scheduling cycle that has started since you ran bench_press will match.

test/tests/functional/pbs_qrun.py
SudeshnaMoulik: Yes @bhroam , if we look at the shell script, where the bug was reported and also able to re-produce, there we need to submit different jobs infinitely in a while loop, to hit that particular scenario.
While converting the shell script to PTL test case, we need to submit jobs infinitely (but in PTL, I submitted job till 1 min) and the issue was reproducable (as the qrun hangs), in some cycle. Attached the log above.

SudeshnaMoulik:
[PTL_qrun_hangs_without_fix_openSuse2.txt](https://github.com/PBSPro/pbspro/files/2638056/PTL_qrun_hangs_without_fix_openSuse2.txt)

test/tests/functional/pbs_qrun.py
bhroam: race conditions are tricky to reproduce in automated tests.  I may have been amiss asking you to write one.  I didn't understand fully what you had to do (sorry).  That said, I think we can craft a test that does what you need.  Turn scheduling off.  Submit 500-1000 different jobs.  The easiest way to do this is to increment the walltime by 1s for each job.  This will force the scheduler to put each job into its own equivalence class.  Turn scheduling on, and the cycle will take several seconds.  During this time you can do your qrun.  I worry doing what you are doing will fail on different speed machines.

SudeshnaMoulik:
Thanks @bhroam . I have addressed the walltime increment of each job, which caused the sched cycle to be longer.
Please review. Attached the PTL logs.

[PTL_qrun_hangs_without_fix3.txt](https://github.com/PBSPro/pbspro/files/2642525/PTL_qrun_hangs_without_fix3.txt)
[PTL_qrun_noHangs_with_fix3.txt](https://github.com/PBSPro/pbspro/files/2642526/PTL_qrun_noHangs_with_fix3.txt)

test/tests/functional/pbs_qrun.py
```diff
@@ -100,30 +100,46 @@ def test_qrun_hangs(self):
         """
         node = self.mom.shortname
         time1 = int(time.time())
+        walltime = 1
+        self.server.manager(MGR_CMD_SET, SCHED,
+                            {'scheduling': 'False'})
+        err_msg = 'qrun: Not enough free nodes available'
         while True:
             self.server.manager(MGR_CMD_SET, NODE,
                                 {'resources_available.ncpus': 1}, id=node)
```
bhroam: Can you move this outside of the while?  It doesn't change each iteration

test/tests/functional/pbs_qrun.py
```diff
@@ -100,30 +100,46 @@ def test_qrun_hangs(self):
...
             j1.create_script('/bin/date')
+            a = {'Resource_List.walltime': walltime}
+            j1.set_attributes(a)
             self.server.submit(j1)
+
+            walltime = walltime + 1
+            a = {'Resource_List.walltime': walltime}
+            j1.set_attributes(a)
             jid2 = self.server.submit(j1)
+            walltime = walltime + 1
```
bhroam: Two things:
1) just having two jobs in the system in different equivalence classes won't slow things down.  If you had 500 or 1000 equivalence classes, that'd slow things down.
2) You don't delete the previous jobs, so the number of jobs will increase over time.  This does increase the number of equivalence classes, but not in a major way.  After 60 times, you'll have 120 jobs in the system.

If what you are trying to achieve is to have the scheduler in cycle when you do your qrun, then submit 500-1000 different jobs and run a cycle.  This cycle should take several seconds.  Do your qrun and see if it hangs.

test/tests/functional/pbs_qrun.py
SudeshnaMoulik: @bhroam  that is already outside of the while.

SudeshnaMoulik:
I have modified the PTL, can you please review.
Attached the PTL logs.
While doing qrun, do we need to submit all those 500 jobs one by one to qrun? As we are hitting the bug when submitting the 1st job itself. As of now the loop is there for qrun.

[PTL_qrun.txt](https://github.com/PBSPro/pbspro/files/2647403/PTL_qrun.txt)

test/tests/functional/pbs_qrun.py
```diff
@@ -95,52 +95,51 @@ def test_invalid_host_val(self):
...
-            a = {'Resource_List.walltime': walltime}
-            j1.set_attributes(a)
-            self.server.submit(j1)
-
-            walltime = walltime + 1
+        self.server.manager(MGR_CMD_SET, NODE,
+                            {'resources_available.ncpus': 1}, id=node)
+        for walltime in range(1, 501):
+            j = Job(TEST_USER)
+            j.create_script('/bin/date')
```
bhroam: You aren't really going to run these jobs, so the default sleep job should be fine.

test/tests/functional/pbs_qrun.py
```diff
@@ -95,52 +95,51 @@ def test_invalid_host_val(self):
...
-            self.server.manager(MGR_CMD_SET, SCHED,
-                                {'scheduling': 'True'})
-            self.server.manager(MGR_CMD_SET, SCHED,
-                                {'scheduling': 'False'})
+        self.server.manager(MGR_CMD_SET, SCHED,
+                            {'scheduling': 'True'})
+        self.server.manager(MGR_CMD_SET, SCHED,
+                            {'scheduling': 'False'})
+        job = 10  # when we turn on scheduling, it scheduled few job.
+        while job < 501:
```
bhroam: I don't see job changing.  Won't this always be true?

test/tests/functional/pbs_qrun.py
```diff
@@ -95,52 +95,51 @@ def test_invalid_host_val(self):
...
                         self.scheduler.log_match("Starting Scheduling Cycle",
-                                                 starttime=time2,
+                                                 starttime=now,
                                                  max_attempts=10)
                     except PtlLogMatchError:
                         self.assertFalse(True,
-                                         "Sched cycle not started.Qrun hangs")
+                                         "Sched cycle not started by qrun")
                     else:
                         continue
```
bhroam: I don't think you need this else clause.  There is nothing else before the end of the loop

bhroam:
The 500 jobs aren't there to qrun.  They are there to give the scheduler something to do.  Since it has to look at 500 jobs, the cycle will take longer.  This gives you a chance to do your qrun while the scheduler is in cycle.  From what I understood from you, this is what causes the issue.

test/tests/functional/pbs_qrun.py
```diff
@@ -92,3 +92,54 @@ def test_invalid_host_val(self):
...
+                    try:
+                        self.scheduler.log_match("Starting Scheduling Cycle",
+                                                 starttime=now,
+                                                 max_attempts=10)
+                    except PtlLogMatchError:
+                        self.assertFalse(True,
+                                         "Sched cycle not started by qrun")
+                    else:
+                        continue
+            except Exception as e:
```
bhroam: I just looked at self.server.runjob() and it won't raise an exception other than PbsRunError.  If this triggers, won't the self.server.runjob() hang?  Will this try/except do what you want it to do?

test/tests/functional/pbs_qrun.py
SudeshnaMoulik: Sure will keep it default. Thanks

test/tests/functional/pbs_qrun.py
SudeshnaMoulik: Yes as now the scheduling cycle is longer, so we dont need to turn scheduling off. I will address this in my next commit.

SudeshnaMoulik: No, we need to turn off scheduling. Because if its enabled, then qrun can trigger another cycle for the jobid we submitted to this.
The bug was, if the scheduling was disabled and qrun is submitted job in current sched cycle then it was hung. As it wasn't able to trigger another cycle, because before triggering we were checking whether scheduling is enabled. 
But as per the feature, irrespective of scheduling status, qrun should trigger the cycle.

test/tests/functional/pbs_qrun.py
SudeshnaMoulik: Yes.
When this "except Exception as e" triggers, that mean qrun hang. If we don't handle this in except block, then it will raise Timeout exception as below (unhandled exception caused stack unwinding)

2018-12-06 09:36:07,546 INFO     TIMEDOUT
2018-12-06 09:36:07,546 INFO     ==========================
2018-12-06 09:36:07,546 INFO      Entered TestQrun tearDown
2018-12-06 09:36:07,546 INFO     ==========================
2018-12-06 09:36:07,547 INFO     ===========================
2018-12-06 09:36:07,547 INFO     Completed TestQrun tearDown
2018-12-06 09:36:07,547 INFO     ===========================
2018-12-06 09:36:07,549 INFO     
2018-12-06 09:36:07,549 INFO     ======================================================================
2018-12-06 09:36:07,549 INFO     TIMEDOUT: test_qrun_hangs (tests.functional.pbs_qrun.TestQrun)

2018-12-06 09:36:07,550 INFO     ___m_oo_m___
2018-12-06 09:36:07,550 INFO     Traceback (most recent call last):
  File "/home/sudemo/work/golden/pbspro/test/tests/functional/pbs_qrun.py", line 124, in test_qrun_hangs
    self.server.runjob(jobid=jid)
  File "/usr/lib/python2.7/site-packages/ptl/lib/pbs_testlib.py", line 7269, in runjob
    logerr=logerr)
  File "/usr/lib/python2.7/site-packages/ptl/utils/pbs_dshutils.py", line 951, in run_cmd
    (o, e) = p.communicate(input)
  File "/usr/lib64/python2.7/subprocess.py", line 800, in communicate
    return self._communicate(input)
  File "/usr/lib64/python2.7/subprocess.py", line 1401, in _communicate
    stdout, stderr = self._communicate_with_poll(input)
  File "/usr/lib64/python2.7/subprocess.py", line 1455, in _communicate_with_poll
    ready = poller.poll()
  File "/usr/lib/python2.7/site-packages/ptl/utils/plugins/ptl_test_runner.py", line 602, in timeout_handler
    raise TimeOut('Timed out after %s second' % timeout)
TimeOut: Timed out after 180 second

SudeshnaMoulik:
[PTL_qrun.txt](https://github.com/PBSPro/pbspro/files/2651814/PTL_qrun.txt)

test/tests/functional/pbs_qrun.py
bhroam: Oh, I see what is going on.  The test case times out from the PTL timeout.  You're catching that exception to cause the test to fail.  Nice.

test/tests/functional/pbs_qrun.py
```diff
@@ -119,27 +118,22 @@ def test_qrun_hangs(self):
...
-                # sched cycle started by qrun and failed as no free nodes
-                if err_msg in e.msg[0]:
-                    self.logger.info(
-                        "No free nodes available. Checking for sched cycle")
-                    try:
-                        self.scheduler.log_match("Starting Scheduling Cycle",
-                                                 starttime=now,
-                                                 max_attempts=10)
-                    except PtlLogMatchError:
+        jid = joblist[10]  # when we turn on scheduling, it scheduled few job.
```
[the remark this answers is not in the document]
SudeshnaMoulik: Yes. Modified the comment phrase.

agrawalravi90:
In your PR description, you say SLES, openSUSE and then you say Linux (very rarely), I'm not sure I understand what that means, can you please clarify?

test/tests/functional/pbs_qrun.py
```diff
@@ -92,3 +92,48 @@ def test_invalid_host_val(self):
...
+        self.logger.info("Submitted 500 jobs with different walltime")
+
+        self.server.manager(MGR_CMD_SET, SCHED,
+                            {'scheduling': 'True'})
+        self.server.manager(MGR_CMD_SET, SCHED,
+                            {'scheduling': 'False'})
+        jid = joblist[10]  # it scheduled a few jobs
+        now = int(time.time())
+        try:
+            self.server.runjob(jobid=jid)
```
agrawalravi90: runjob() is a blocking call by default, so if qrun actually gets hung, your test will just remain hung and timeout. So, I'd recommend that you set runjob()'s async argument to True while calling it, and then all that you need to do is check that a sched cycle was started by the call, which you might want to do inside a 'finally' block as you should check for that regardless of PbsRunError being thrown.

test/tests/functional/pbs_qrun.py
```diff
@@ -92,3 +92,48 @@ def test_invalid_host_val(self):
...
+            j.set_attributes(a)
+            jid = self.server.submit(j)
+            joblist.append(jid)
+        self.logger.info("Submitted 500 jobs with different walltime")
+
+        self.server.manager(MGR_CMD_SET, SCHED,
+                            {'scheduling': 'True'})
+        self.server.manager(MGR_CMD_SET, SCHED,
+                            {'scheduling': 'False'})
+        jid = joblist[10]  # it scheduled a few jobs
```
agrawalravi90: Why not qrun the last job id? That way you'll have the least chance of running into a race condition.

SudeshnaMoulik:
Hi Ravi,
When the issues reported it was only for SLES11 and opensuse/leap:42. 
But after we modified steps (given above by @bhroam) to reproduce the bugs, its now reproducable in Linux also.
Let me know in-case I need to modify the description.

agrawalravi90:
opensuse and sles are Linux based distros as well. Did you mean centos by any chance?

SudeshnaMoulik:
> opensuse and sles are Linux based distros as well. Did you mean centos by any chance?

Yes.

test/tests/functional/pbs_qrun.py
SudeshnaMoulik: Thanks @agrawalravi90 for the review. 
Using "async" option will not help in this case. Below are the explanation. Let me know if am missing anything.
As per the doc:
qrun -a jobid : The qrun command exits before the job actually starts execution.
I believe in PTL "self.server.runjob(jobid=jid,async=True)" does the same thing.

If the job is unable to run and scheduling is also disabled, then irrespective of "async" parameter, it will hung. And which will turn runjob() call to blocking (unless its time out).

test/tests/functional/pbs_qrun.py
SudeshnaMoulik: Yes sure. Then I don't need to save all jobid also in a list.
Will remove the list too.

SudeshnaMoulik:
[PTL_qrun.txt](https://github.com/PBSPro/pbspro/files/2663160/PTL_qrun.txt)

agrawalravi90:
> Yes.

Then please correct the PR description

test/tests/functional/pbs_qrun.py
agrawalravi90: Oh, ok, thanks for explaining. How about calling runjob in a separate Python thread/process? the main thread can then go ahead and check sched logs while the child thread is doing runjob, that way even if it gets hung the test will do its job correctly.

A test that checks for a command getting hung should not get hung itself, it should be able to fail/pass depending on whether the command got hung or not.

test/tests/functional/pbs_qrun.py
```diff
@@ -110,15 +109,16 @@ def test_qrun_hangs(self):
             j = Job(TEST_USER)
             a = {'Resource_List.walltime': walltime}
             j.set_attributes(a)
-            jid = self.server.submit(j)
-            joblist.append(jid)
+            if walltime == 500:
```
agrawalravi90: you don't need this condition, you can just always say jid = self.server.submit(j) and then check for jid later, it will be job id of the last job submitted

SudeshnaMoulik: It will be an extra task to access and save a variable value each time (500 times) that we don't need.
Doing a conditional check before save the value should be more effective.
Let me know your suggestion.

test/tests/functional/pbs_qrun.py
SudeshnaMoulik: Sure.  I will fork another process for runjob().

SudeshnaMoulik:
[PTL_qrun.txt](https://github.com/PBSPro/pbspro/files/2667545/PTL_qrun.txt)

test/tests/functional/pbs_qrun.py
```diff
@@ -119,21 +120,19 @@ def test_qrun_hangs(self):
...
-                except PtlLogMatchError:
+        pid = os.fork()
+        if pid == 0:
+            try:
+                self.server.runjob(jobid=jid)
+            except PbsRunError as e:
+                self.logger.info("Runjob throws error: " + e.msg[0])
+        else:
+            time.sleep(10)
+            now = int(time.time())
```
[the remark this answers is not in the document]
SudeshnaMoulik: Thanks @agrawalravi90 
I realized getting the current time before I fork, can also include the "Starting scheduling cycle" the normal one also. As all is happening at the same time (as per the log). So better will be to count the number of that message.

SudeshnaMoulik:
Attached logs.

test/tests/functional/pbs_qrun.py
```diff
@@ -115,25 +115,26 @@ def test_qrun_hangs(self):
...
+                self.logger.info("Runjob throws error: " + e.msg[0])
+        else:
+            time.sleep(5)
+            try:
+                out = self.scheduler.log_match("Starting Scheduling Cycle",
+                                               allmatch=True,
+                                               starttime=now,
+                                               max_attempts=10)
+                if len(out) < 2:
+                    self.assertFalse(True,
```
bhroam: Use self.assertLess() instead of an if and assertFalse(True)

SudeshnaMoulik: Thanks @bhroam. I have modified the PTL. Added extra steps to handle the hung process.
@agrawalravi90 @bhroam Can you kindly review the latest commit "Incorporated Changes".

test/tests/functional/pbs_qrun.py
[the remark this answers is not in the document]
agrawalravi90: I'm not sure I understand. If you want to check that there was a sched cycle triggered by qrun, then shouldn't you set 'now' to just before the fork call? then the only sched cycle that gets triggered will be the one that qrun triggers. I'm not sure I understand the complication.

test/tests/functional/pbs_qrun.py
```diff
@@ -92,3 +94,44 @@ def test_invalid_host_val(self):
...
+            if walltime == 500:
+                jid = self.server.submit(j)
+            else:
+                self.server.submit(j)
+        self.logger.info("Submitted 500 jobs with different walltime")
+
+        now = int(time.time())
+        self.server.manager(MGR_CMD_SET, SCHED,
+                            {'scheduling': 'True'})
+        self.server.manager(MGR_CMD_SET, SCHED,
```
agrawalravi90: Please set Expect to True here to make sure that scheduling is turned off before you proceed

test/tests/functional/pbs_qrun.py
```diff
@@ -92,3 +94,44 @@ def test_invalid_host_val(self):
...
+            except PbsRunError as e:
+                self.logger.info("Runjob throws error: " + e.msg[0])
+        else:
+            out = self.scheduler.log_match(
+                "Starting Scheduling Cycle",
+                allmatch=True, interval=5, starttime=now, max_attempts=10)
+            if len(out) < 2:
+                os.kill(os.getppid(), signal.SIGTERM)
+                self.assertFalse(True,
+                                 "Qrun didn't start another sched cycle")
```
agrawalravi90: you should probably do os.waitpid() and make sure the child process is finished before you return from this function

test/tests/functional/pbs_qrun.py
SudeshnaMoulik: Sure.

test/tests/functional/pbs_qrun.py
SudeshnaMoulik: As we are killing here the hung process (after checking the logs), so we don't need here waitpid.

Waitpid will wait for the process to time out (hung period), but to avoid that (as the PTL test case should not hung) we are forking another process.

test/tests/functional/pbs_qrun.py
SudeshnaMoulik: Here we are setting value to a SCHED object, if I use "expect=True" It will fail with below error.

2018-12-14 10:16:18,088 INFO     expect on server centosvm: scheduling set False && server_state set Idle sched attempt: 60
2018-12-14 10:16:18,594 INFO     FAILED
File "/usr/lib/python2.7/site-packages/ptl/lib/pbs_testlib.py", line 8112, in expect
    raise PtlExpectError(rc=1, rv=False, msg=_msg)
PtlExpectError: rc=1, rv=False, msg=expected on server centosvm: scheduling set False && server_state set Idle sched attempt: 60

test/tests/functional/pbs_qrun.py
agrawalravi90: I was suggesting doing a waitpid() after the if statement, which means it would be after you have already sent the kill signal to the process. I just the function to reap the child process cleanly before exiting.

test/tests/functional/pbs_qrun.py
```diff
@@ -92,3 +94,44 @@ def test_invalid_host_val(self):
...
+                self.server.submit(j)
+        self.logger.info("Submitted 500 jobs with different walltime")
+
+        now = int(time.time())
+        self.server.manager(MGR_CMD_SET, SCHED,
+                            {'scheduling': 'True'})
+        self.server.manager(MGR_CMD_SET, SCHED,
+                            {'scheduling': 'False'})
+        pid = os.fork()
+        if pid > 0:
```
agrawalravi90: wait, why are you making the parent process execute the runjob command? So the child process is calling os.kill() on the parent? Why are you doing that?

test/tests/functional/pbs_qrun.py
SudeshnaMoulik: If I use a waitpid after the kill then it will throw error, as that process is already killed.
    os.waitpid(hungpid, 0)
OSError: [Errno 10] No child processes

test/tests/functional/pbs_qrun.py
SudeshnaMoulik: Sure.

test/tests/functional/pbs_qrun.py
SudeshnaMoulik: Addressed this in my recent commit.

SudeshnaMoulik:
Attached logs:
[before_fix.txt](https://github.com/PBSPro/pbspro/files/2679662/before_fix.txt)
[after_fix.txt](https://github.com/PBSPro/pbspro/files/2679663/after_fix.txt)

SudeshnaMoulik:
Thanks @agrawalravi90

agrawalravi90:
@SudeshnaMoulik you haven't fixed the PR description yet, it still says that the bug appears on SLES and SUSE and very rarely on Linux. Since you were able to reproduce this consistently on different Linux distros, maybe just say "All Linux"?

agrawalravi90:
Actually https://github.com/PBSPro/pbspro/pull/893 is ahead of you, please rebase once that one is merged

agrawalravi90:
#893 got merged, please rebase now

PR #899, opened 2018-11-28
Unsupported usage of qsub should throw error

vstumpf:
#### Bug/feature Description
* Using qsub with both a script and executable should fail.
* Currently, it does not.

#### Affected Platform(s)
* Linux

* qsub's usage says `qsub ... [script | -- command [arg1 ...]]`
* getopt would rearrange argv to put all of the options before the non-options. This would include `--`.
* If someone put both, it would move the `--` before the script, leading qsub to believe that the script is actually an executable with arguments.
* getopt returns the index of the first non-option.

#### Solution Description
* Before calling getopt, make a copy of argv. After calling getopt, if the string at the index of the first non-option is different between the copy and the original, then getopt must have changed the order.
* The only reason why getopt would change the order is that there was an argument before the `--`. The usage says this is unsupported, and we return an error.
* PBS_GNU_GETOPTS never gets defined, so I have removed the dead code. PBS_GNU_GETOPTS used to get defined by the configure script by compiling a little .c program and checking the return code, but from looking at it, I don't think it tested the option correctly.

#### Testing logs/output
[my_version_windows.txt](https://github.com/PBSPro/pbspro/files/2622308/my_version_windows.txt)
[after_linux.txt](https://github.com/PBSPro/pbspro/files/2622309/after_linux.txt)

#### Checklist:

### DIFF
--- a/src/cmds/qsub.c
+++ b/src/cmds/qsub.c
@@ -2346,20 +2346,14 @@ process_opts(int argc, char **argv, int passet)
 	int ddash_index = -1;
 
 #ifdef WIN32
-#define GETOPT_ARGS_ORIG "a:A:c:C:e:fGhIj:J:k:l:m:M:N:o:p:q:r:R:S:u:v:VW:zP:"
+#define GETOPT_ARGS "a:A:c:C:e:fGhIj:J:k:l:m:M:N:o:p:q:r:R:S:u:v:VW:zP:"
 #else
 #if !defined(PBS_NO_POSIX_VIOLATION)
-#define GETOPT_ARGS_ORIG "a:A:c:C:e:fhIj:J:k:l:m:M:N:o:p:q:r:R:S:u:v:VW:XzP:"
+#define GETOPT_ARGS "a:A:c:C:e:fhIj:J:k:l:m:M:N:o:p:q:r:R:S:u:v:VW:XzP:"
 #else
-#define GETOPT_ARGS_ORIG "a:A:c:C:e:fhj:J:k:l:m:M:N:o:p:q:r:R:S:u:v:VW:zP:"
+#define GETOPT_ARGS "a:A:c:C:e:fhj:J:k:l:m:M:N:o:p:q:r:R:S:u:v:VW:zP:"
 #endif /* PBS_NO_POSIX_VIOLATION */
 #endif /* WIN32 */
-
-#ifdef PBS_GNU_GETOPTS
-#define GETOPT_ARGS "+"GETOPT_ARGS_ORIG
-#else
-#define GETOPT_ARGS GETOPT_ARGS_ORIG
-#endif /* PBS_GNU_GETOPTS */
 
 /*
  * The following macro, together the value of passet is used
@@ -2954,6 +2948,11 @@ process_special_args(int argc, char **argv, char *script)
 			if (!N_opt)	/* '-N' is not set */
 				set_attr(&attrib, ATTR_N, "STDIN");
 		} else {
+			if (optind + 1 != argc) {
+				/* argument is a job script, it should be last */
+				print_usage();
+				exit_qsub(2);
+			}
 			snprintf(script, MAXPATHLEN, "%s", argv[optind]);
 		}
 	}
@@ -4470,7 +4469,7 @@ do_connect(char *server_out, char *retmsg)
  * @retval 0 - Success
  * @retval 1/-1/pbs_errno - Failure, retmsg paramter is set
  * @retval DMN_REFUSE_EXIT - If daemon can't submit the job
- * 
+ *
  */
 static int
 do_submit(char *retmsg)
@@ -4498,7 +4497,7 @@ do_submit(char *retmsg)
 			return (rc);
 	}
 
-	/* 
+	/*
 	 * get environment variable if -V option is set. Return the code
 	 * DMN_REFUSE_EXIT if -V option is detected in background qsub.
 	 */
@@ -5548,7 +5547,7 @@ fork_and_stay(void)
 		pbs_client_thread_set_single_threaded_mode();
 
 		/* set when background qsub is running */
-		is_background = 1; 
+		is_background = 1;
 		do_daemon_stuff();
 		/*
 		 * Control should never reach here.
@@ -5718,6 +5717,8 @@ main(int argc, char **argv, char **envp) /* qsub */
 	char qsub_exe[MAXPATHLEN + 1];
 #endif
 	int daemon_up = 0;
+	char **argv_cpy; /* copy argv for getopt */
+	int i;
 
 	/* Set signal handlers */
 	set_sig_handlers();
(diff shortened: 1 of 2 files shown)


vstumpf:
Requesting a review from @latha-subramanian @lisa-altair.
Thanks!

src/cmds/qsub.c
```diff
@@ -2924,16 +2918,27 @@ process_opts(int argc, char **argv, int passet)
...
+		if (strcmp(argv[optind], argv_cpy[optind]) != 0) {
+			/* The arguments changed, the script and "--" must have been present.
+			 * getopt will move all non-options to the end of the array. In qsub's
+			 * case, it will only happen if both the "script" and "-- executable"
+			 * were present in the qsub command. This is unsupported usage and
+			 * should exit.
+			 */
+			print_usage();
+			exit_qsub(2);
+		}
```
bhroam: You passed in argv_cpy for just one check.  I'd put that check straight in main instead of passing in the copy.

vstumpf: I put it in here because it looked out of place in main; I'll put it back.

vstumpf:
I've updated the tests and the source changes. Please review again.

src/cmds/qsub.c
```diff
@@ -2346,21 +2346,15 @@ process_opts(int argc, char **argv, int passet)
...
-#define GETOPT_ARGS_ORIG "a:A:c:C:e:fhj:J:k:l:m:M:N:o:p:q:r:R:S:u:v:VW:zP:"
+#define GETOPT_ARGS "a:A:c:C:e:fhj:J:k:l:m:M:N:o:p:q:r:R:S:u:v:VW:zP:"
 #endif /* PBS_NO_POSIX_VIOLATION */
 #endif /* WIN32 */
 
-#ifdef PBS_GNU_GETOPTS
-#define GETOPT_ARGS "+"GETOPT_ARGS_ORIG
-#else
-#define GETOPT_ARGS GETOPT_ARGS_ORIG
-#endif /* PBS_GNU_GETOPTS */
```
lisa-altair: Did PBS_GNU_GETOPTS not get defined previously?  Is this basically dead code?

test/tests/functional/pbs_qsub_two_args.py
```diff
@@ -0,0 +1,111 @@
...
+# under a commercial license agreement.
+#
+# Use of Altair’s trademarks, including but not limited to "PBS™",
+# "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
+# trademark licensing policies.
+
+from tests.functional import *
+
+
+class TestQsubTwoArgs(TestFunctional):
```
lisa-altair: This test suite is mainly focused on cases where there is a script and an executable right?
Can you rename the suite, TwoArgs is seems a bit generic it could be any 2 arguments.

test/tests/functional/pbs_qsub_two_args.py
```diff
@@ -0,0 +1,111 @@
...
+        self.fn = self.du.create_temp_file(body=script)
+
+    def test_qsub_with_script_executable(self):
+        """
+        submit a job with a script and executable
+        """
+        cmd = 'qsub ' + self.fn + ' -- /bin/sleep 10'
+        rv = self.du.run_cmd(self.server.hostname, cmd=cmd)
+        failed = rv['rc'] == 2 and rv['err'][0].split(' ')[0] == 'usage:'
+        self.assertTrue(failed, 'qsub did not fail')
```
lisa-altair: I don't run a lot of PTL tests, and because of that, when I see that a test failed and then I look at the error message that says 'qsub did not fail', I find it a bit confusing.  Could you make this error message a bit more clear?

test/tests/functional/pbs_qsub_two_args.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def test_qsub_with_script_dashes(self):
+        """
+        submit a job with a script and dashes
+        """
+        cmd = 'qsub ' + self.fn + ' --'
+        rv = self.du.run_cmd(self.server.hostname, cmd=cmd)
+        failed = rv['rc'] == 2 and rv['err'][0].split(' ')[0] == 'usage:'
+        self.assertTrue(failed, 'qsub did not fail')
```
lisa-altair: Could you make this error message a bit more clear?

test/tests/functional/pbs_qsub_two_args.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def test_qsub_with_dashes(self):
+        """
+        submit a job with only dashes
+        """
+        cmd = 'qsub --'
+        rv = self.du.run_cmd(self.server.hostname, cmd=cmd)
+        failed = rv['rc'] == 2 and rv['err'][0].split(' ')[0] == 'usage:'
+        self.assertTrue(failed, 'qsub did not fail')
```
lisa-altair: Could you make this error message a bit more clear?

test/tests/functional/pbs_qsub_two_args.py
```diff
@@ -93,3 +93,19 @@ def test_qsub_with_executable(self):
         cmd = 'qsub -- /bin/sleep 10'
         rv = self.du.run_cmd(self.server.hostname, cmd=cmd)
         self.assertEquals(rv['rc'], 0, 'qsub failed')
+
+    def test_qsub_with_option_executable(self):
+        """
+        submit a job with an option and executable
+        """
+        cmd = 'qsub -V -- /bin/sleep 10'
```
bhroam: Hirin once told me that it is better to pass run_cmd() a list of command arguments instead of a string.

src/cmds/qsub.c
vstumpf: PBS_GNU_GETOPTS never gets defined, so I have removed the dead code. PBS_GNU_GETOPTS used to get defined by the configure script by compiling a little .c program and checking the return code, but from looking at it, I don't think it tested the option correctly.

vstumpf:
Here's an updated log
[test_qsub_opt_args.txt](https://github.com/PBSPro/pbspro/files/2641557/test_qsub_opt_args.txt)

test/tests/functional/pbs_qsub_two_args.py
```diff
@@ -55,57 +55,57 @@ def test_qsub_with_script_executable(self):
         """
         submit a job with a script and executable
         """
-        cmd = 'qsub ' + self.fn + ' -- /bin/sleep 10'
+        cmd = ['qsub', self.fn, '--',  '/bin/sleep 10']
```
bhroam: Consider doing os.path.join(self.server.pbs_conf['PBS_EXEC'], 'qsub')

vstumpf:
Thanks for the reviews.
Addressing bhroam's changes.
Here's the new log:
[test_qsub_opt_args.txt](https://github.com/PBSPro/pbspro/files/2641938/test_qsub_opt_args.txt)

bhroam:
I merged in the PR.

PR #901, opened 2018-11-28
asynchronous mom job saves

sakshamgarg:
#### Bug/feature Description
* *Asynchronously writing files in MOM to improve the performance.*

#### Affected Platform(s)
* *All*

* *Currently MOM uses O_SYNC flag to write files. This adds additional time to the job before it can start.*

#### Solution Description
* *Removing the O_SYNC flag while writing files in MOM.*

#### Testing logs/output
* *For Linux: [async_mom_job_saves_Test_linux.txt](https://github.com/PBSPro/pbspro/files/2625569/async_mom_job_saves_Test_linux.txt)*
* *For Windows: [async_mom_job_saves_Test_windows.txt](https://github.com/PBSPro/pbspro/files/2625570/async_mom_job_saves_Test_windows.txt)*

#### Checklist:

### DIFF
--- a/src/server/job_recov.c
+++ b/src/server/job_recov.c
@@ -115,13 +115,12 @@ static const size_t extndsize = sizeof(union jobextend);
  *			 - a full write for a new job
  *
  *		For a quick update, the data written is less than a disk block
- *		size and no size change occurs; so it is rewritten in place
- *		with O_SYNC.
+ *		size and no size change occurs.
  *
- *		For a full update (usually following modify job request), to
- *		insure no data is ever lost due to system crash:
- *		1. write (with O_SYNC) new image to a new file using a temp name
- *		2. rename the temp file to the old name
+ *		No need of O_SYNC flag as this will improve the performance.
+ *		This might lead to data loss from file system in case of system
+ *		crash. This is not an issue as data is mostly recovered from the
+ *		database.
  *
  *		For a new file write, first time, the data is written directly to
  *		the file.
@@ -183,7 +182,7 @@ job_save_fs(job *pjob, int updatetype)
 
 	if (updatetype == SAVEJOB_QUICK) {
 
-		openflags =  O_WRONLY | O_Sync;
+		openflags =  O_WRONLY;
 		fds = open(namebuf1, openflags, pmode);
 		if (fds < 0) {
 			log_err(errno, "job_save", "error on open");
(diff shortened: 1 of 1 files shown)


mike0042:
I have reservations about this change. In the comment you mention job data is "mostly" recovered from the database. Which parts are recovered from files? What if the file was buffered, but never written? What if the file was partially written, resulting in incomplete/corrupt contents? How much will performance improve? How will we test this?

subhasisb:
I agree @mike0042 - this was a change i had done for the 10k jobs/sec.

1. The way to test this would be to measure synchronous writes to no-sync writes for a large number of job saves (possibly using a completely separate piece of code). The other way test this would be via PBS by removing all other bottlenecks we know and then only enabling and disabling this change to see the effect on throughput of a  set of short jobs.

2. While recovering the job, if the size read is less than the expected size of the job (fixed size + the variable size) then the job_recov() aborts loading the job and the job is deleted from the system (this is existing functionality). 

The idea behind the change was that a complete system power loss is not very frequent. If a mom crash happens, a no-sync write is still fine, since the linux sync cycle will anyway sync it. If there is a kernel crash (or a host power failure) then we would have a partially written file. In such a case, on recovery, the job will be removed from the mom, resulting in the server rescheduling the job (we should ensure this is what is happening). Thus the impact is in the worst case scenario and the hope is that a system crash is quite infrequent, thus we optimize for the best case (yet be able to handle the worst case scenario with a bit of setback).

What do you think?

mike0042:
Thank you for the explanation @subhasisb. I agree that a server crash is a rare occurance, so that does mitigate the risk to some extent. And disk and/or controller failures are also rare. Do we currently have a throughput test that could be used to measure performance? We must have something from the 10K jobs/sec work that could be used. My primary reason for writing the comment was to make folks aware that this change is not zero-risk, and to acknowledge that was the case. Since we have done that, I'm happy to move forward with this.

subhasisb:
Thanks @mike0042 - your feedback was absolutely proper and welcome. As you mentioned, we can make use of the existing 10k jobs/min test suite to test this end to end in terms of performance.

sakshamgarg:
Thank you @mike0042 for your suggestion and feedback.
I will test the performance and will post the results as suggested by @subhasisb and @mike0042 .

sakshamgarg:
I ran the following performance test on a development machine via PBS.
Test case scenario:
Testing job completion performance from multiple users with scheduling on. Submitting 100 jobs from each user with a total of 10 users for 10 iterations.

Comparing throughput between the master branch and the PR branch:
Master Branch throughput: 54
PR Branch throughput: 60

PR #912, opened 2018-12-05
increase the default timeout for PTL tests

dinesh1306:
#### Bug/feature Description
* Increase the default timeout for PTL tests to 600sec.
* [PP-1047](https://pbspro.atlassian.net/browse/PP-1047)*

#### Affected Platform(s)
* All

* sometime if setup takes longer time in some tests where there are too many jobs or nodes to cleanup then it affects test execution time. default time for each test case execution is 180sec including setUp and tearDown. If setUp and test execution is not completed in 180sec then test got timeout.

#### Solution Description
* Increase the default timeout for PTL tests to 600sec. if timeout decorator have value less than 600sec(Default testcase timeout) then PTL consider testcase timeout value as Default testcase timeout otherwise consider value provided by decorator. 

#### Testing logs/output
* [res_TestDefaultTimeout.txt](https://github.com/PBSPro/pbspro/files/2648964/res_TestDefaultTimeout.txt)

#### Checklist:

### DIFF
--- a/test/fw/ptl/utils/pbs_testsuite.py
+++ b/test/fw/ptl/utils/pbs_testsuite.py
@@ -133,6 +133,7 @@ SMOKE = 'smoke'
 REGRESSION = 'regression'
 NUMNODES = 'numnodes'
 TIMEOUT_KEY = '__testcase_timeout__'
+MINIMUM_TESTCASE_TIMEOUT = 600
 
 
 def skip(reason="Skipped test execution"):
@@ -155,7 +156,18 @@ def timeout(val):
     """
     Decorator to set timeout value of test case
     """
+    logger = logging.getLogger(__name__)
+    old_val = None
+    if val < MINIMUM_TESTCASE_TIMEOUT:
+        old_val = val
+        val = MINIMUM_TESTCASE_TIMEOUT
+
     def wrapper(obj):
+        msg = 'for test ' + obj.func_name
+        msg += ' minimum-testcase-timeout updated to '
+        msg += str(val) + ' from ' + str(old_val)
+        if old_val:
+            logger.info(msg)
         setattr(obj, TIMEOUT_KEY, int(val))
         return obj
     return wrapper
@@ -564,7 +576,7 @@ class PBSTestSuite(unittest.TestCase):
         cls._validate_param('revert-queues')
         cls._validate_param('revert-resources')
         if 'default-testcase-timeout' not in cls.conf.keys():
-            cls.conf['default_testcase_timeout'] = 180
+            cls.conf['default_testcase_timeout'] = MINIMUM_TESTCASE_TIMEOUT
         else:
             cls.conf['default_testcase_timeout'] = int(
                 cls.conf['default-testcase-timeout'])
--- a/test/tests/selftest/pbs_default_timeout.py
+++ b/test/tests/selftest/pbs_default_timeout.py
@@ -1,0 +1,110 @@
+# coding: utf-8
+
+# Copyright (C) 1994-2018 Altair Engineering, Inc.
+# For more information, contact Altair at www.altair.com.
+#
+# This file is part of the PBS Professional ("PBS Pro") software.
+#
+# Open Source License Information:
+#
+# PBS Pro is free software. You can redistribute it and/or modify it under the
+# terms of the GNU Affero General Public License as published by the Free
+# Software Foundation, either version 3 of the License, or (at your option) any
+# later version.
+#
+# PBS Pro is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.
+# See the GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+# Commercial License Information:
+#
+# For a copy of the commercial license terms and conditions,
+# go to: (http://www.pbspro.com/UserArea/agreement.html)
+# or contact the Altair Legal Department.
+#
+# Altair’s dual-license business model allows companies, individuals, and
+# organizations to create proprietary derivative works of PBS Pro and
+# distribute them - whether embedded or bundled with other software -
+# under a commercial license agreement.
+#
+# Use of Altair’s trademarks, including but not limited to "PBS™",
+# "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
+# trademark licensing policies.
+
+from tests.selftest import *
+from ptl.utils.plugins.ptl_test_runner import TimeOut
+
+
+class TestDefaultTimeout(TestSelf):
+    """
+    Test suite to verify increase default testcase timeout working properly.
+    """
+
+    def test_default_timeout(self):
+        """
+        Verify that test not timedout after 180 sec.
+        i.e. after previous default timeout value
+        """
+        try:
+            mssg = 'sleeping for %ssec and minimum-testcase-timeout is %ssec' \
+                   % (MINIMUM_TESTCASE_TIMEOUT/2, MINIMUM_TESTCASE_TIMEOUT)
+            self.logger.info(mssg)
+            time.sleep(MINIMUM_TESTCASE_TIMEOUT/2)
+        except TimeOut as e:
+            mssg = 'Timed out after %s second' % MINIMUM_TESTCASE_TIMEOUT
+            err_mssg = 'The test timed out for an incorrect timeout duration'
+            self.assertEqual(mssg, e.message, err_mssg)
+
+    def test_timeout_greater_default_value(self):
+        """
+        Verify that test timedout after 600 sec.
+        """
+        try:
+            mssg = 'sleeping for %ssec and minimum-testcase-timeout is %ssec' \
+                   % (MINIMUM_TESTCASE_TIMEOUT+1, MINIMUM_TESTCASE_TIMEOUT)
+            self.logger.info(mssg)
+            time.sleep(MINIMUM_TESTCASE_TIMEOUT+1)
+        except TimeOut as e:
+            mssg = 'Timed out after %s second' % MINIMUM_TESTCASE_TIMEOUT
+            err_mssg = 'The test timed out for an incorrect timeout duration'
+            self.assertEqual(mssg, e.message, err_mssg)
+        else:
+            msg = 'Test did not timeout after the min. timeout period of %d' \
+                  % MINIMUM_TESTCASE_TIMEOUT
+            self.fail(msg)
+
+    @timeout(200)
+    def test_timeout_decorator_less_default_value(self):
+        """
+        If timeout decorator value is less than 600 then
+        default testcase timeout is considered
+        """
+        try:
+            mssg = 'sleeping for %ssec and minimum-testcase-timeout is %ssec' \
+                   % (MINIMUM_TESTCASE_TIMEOUT/2, MINIMUM_TESTCASE_TIMEOUT)
+            self.logger.info(mssg)
+            time.sleep(MINIMUM_TESTCASE_TIMEOUT/2)
+        except TimeOut as e:
+            mssg = 'Timed out after %s second' % MINIMUM_TESTCASE_TIMEOUT
+            err_mssg = 'The test timed out for an incorrect timeout duration'
+            self.assertEqual(mssg, e.message, err_mssg)
+
+    @timeout(800)
+    def test_timeout_decorator_greater_default_value(self):
+        """
+        If timeout decorator value is greater than 600 then
+        testcase timeout value consider as timeout decorator value
+        """
+        try:
+            mssg = 'sleeping for %ssec and minimum-testcase-timeout is %ssec' \
+                   % (MINIMUM_TESTCASE_TIMEOUT+1, 800)
+            self.logger.info(mssg)
+            time.sleep(MINIMUM_TESTCASE_TIMEOUT+1)
+        except TimeOut as e:
+            mssg = 'Timed out after 800 second'
+            err_mssg = 'The test timed out for an incorrect timeout duration'
+            self.assertEqual(mssg, e.message, err_mssg)


bhroam:
Is this the right answer, or is just not counting the PbsTestSuite.setUp() and PbsTestSuite.tearDown() time?  I would include the part of setUp() and tearDown() that the test writer wrote themselves.

test/tests/selftest/pbs_default_timeout.py
```diff
@@ -0,0 +1,67 @@
...
+        """
+        Verify that test not timedout after 180 sec.
+        i.e. after previous default timeout value
+        """
+        time.sleep(300)
+
+    @timeout(200)
+    def test_timeout_decorator_less_default_value(self):
+        """
+        If timeout decorator value is less than 600 then testcase timeout value
```
anamikau: reword it to say "If timeout decorator value is less than 600 then default test case timeout is considered"

bhroam:
@anamikau All of setUp() and tearDown() are included.  I'm suggesting we don't include the time for PbsTestSuite.setUp() and PbsTestSuite.tearDown().  The example given is if setUp() has to delete a whole bunch of jobs.  That happens in PbsTestSuite.setUp().  If time spent in that function wasn't included in the timeout, the 180s might be fine.

anamikau:
@agrawalravi90 and @bhroam you are absolutely right. This is a temporary fix to avoid failures that we are observing during the internal testing. For long term we would want to remove setUp and tearDown from the timeout.
I suggest we do not increase the scope of this one and we can resolve the actual problem in different PR.

agrawalravi90:
> For long term we would want to remove setUp and tearDown from the timeout.

Actually that's not what we had discussed. We had just decided to move the reverting to defaults action to tearDown, that way the timeout will correctly apply to the test case which submits the large number of jobs and we don't have to change the default timeout. It's important for the timeout to apply to setUp and tearDown as well because those functions can also get stuck and we'd wanna timeout and bail if that happens.

Is this very urgent to fix? Could we just not run the tests which cause timeout issues until we fix this correctly? It shouldn't be that big of a change to do this properly, I fear that a duct tape fix now will remove the motivation to fix this the right way.

bhroam:
@agrawalravi90 I'm not sure you want to move revert_to_defaults() to tearDown().  Right now it is in setUp().  If setUp() times out, the system is in some in between state.  The next call to setUp() will call revert_to_defaults() and continue the process.  If you move it to tearDown(), it will time out and leave the complex in the in between state which could cause the next test to fail.  I guess it is bad in either case.  Today we will timeout the following test.  In the proposed changes, we'll timeout the current test and fail the following test (or worse, not find a bug when there is one).

agrawalravi90:
@bhroam, actually i didn't lay out the whole design that was discussed, we had discussed putting a "conditional" revert_to_default in setUp() which will be executed only if the previous tearDown() failed. This is why I had proposed that we move this to a forum discussion.

agrawalravi90:
The nice thing about moving revert to tearDown is that the test that is creating a lot of cruft to be cleaned up later can create a custom timeout for itself and ensure that it doesn't time out.

anamikau:
@dinesh1306 please fix codacy error.

@bhroam and @agrawalravi90, can we just have the short term fix in this ticket and take the longer term approach in a different PR?

anamikau:
> > For long term we would want to remove setUp and tearDown from the timeout.
> 
> Actually that's not what we had discussed. We had just decided to move the reverting to defaults action to tearDown, that way the timeout will correctly apply to the test case which submits the large number of jobs and we don't have to change the default timeout. It's important for the timeout to apply to setUp and tearDown as well because those functions can also get stuck and we'd wanna timeout and bail if that happens.

Sorry you are right. I had to look up the notes again to recall.
> 
> Is this very urgent to fix? Could we just not run the tests which cause timeout issues until we fix this correctly? It shouldn't be that big of a change to do this properly, I fear that a duct tape fix now will remove the motivation to fix this the right way.

Skipping these tests would be tricky as these failures are very intermittent on different systems. Again this fix is not going to solve the problem 100% so we need to work on the long term approach sooner.

bhroam:
@anamikau if the short term fix is really important, then yeah I can see it.  I can see Ravi's point though.  After the short term fix, would the long term fix be not as high priority?

anamikau:
Thanks @bhroam. This fix should not change the priority for the other bug.

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -155,6 +156,9 @@ def timeout(val):
     """
     Decorator to set timeout value of test case
     """
+    if val < DEFAULT_TESTCASE_TIMEOUT:
+        val = DEFAULT_TESTCASE_TIMEOUT
```
agrawalravi90: why are you doing this? If a test case explicitly says it needs x seconds of timeout, why do you wanna change it to be a minimum of DEFAULT_TESTCASE_TIMEOUT ? I think we should leave custom timeouts alone.

test/tests/selftest/pbs_default_timeout.py
```diff
@@ -0,0 +1,67 @@
...
+    """
+    Test suite to verify increase default testcase timeout working properly.
+    """
+
+    def test_default_timeout(self):
+        """
+        Verify that test not timedout after 180 sec.
+        i.e. after previous default timeout value
+        """
+        time.sleep(300)
```
agrawalravi90: How about this: Catch the TimeOut exception and mark the test as 'failed' if the exception is caught in 300 seconds? Right now the test is not really very helpful, it will itself timeout while testing timeout, it should either pass or fail correctly.

test/tests/selftest/pbs_default_timeout.py
agrawalravi90: I strongly feel that PTL should not modify custom timeouts provided by tests internally.

test/tests/selftest/pbs_default_timeout.py
agrawalravi90: And instead of 300 seconds, you should probably use something like DEFAULT_TESTCASE_TIMEOUT / 2

agrawalravi90:
@anamikau , alright sounds good, I'm ok putting in this fix if it is urgent.

test/fw/ptl/utils/pbs_testsuite.py
dinesh1306: @agrawalravi90 , previously timeout value was 180 and there are some tests have value less than 600 and greater than 180 in timeout tag, to make affect on those tests also i added above condition.

agrawalravi90:
Ok, let's call it 'minimum_testcase_timeout' instead of 'default_testcase_timeout' then. Yes, a warning message would be good.

anamikau:
Thanks @agrawalravi90. your suggestion looks fine to me. 
@dinesh1306 please address Ravi's comments above and also fixed codacy errors.

dinesh1306:
@anamikau @agrawalravi90 , timeout decorator evaluate initially in test suite. If we put warning message in timeout decorator function then it will comes initially in testsuite. Please let me know can i add warning message or not.

dinesh1306:
@anamikau @agrawalravi90 , I have incorporate review comments other than warning message. Please review.

test/tests/selftest/pbs_default_timeout.py
```diff
@@ -36,32 +36,57 @@
...
+from ptl.utils.plugins.ptl_test_runner import TimeOut
 
 
 class TestDefaultTimeout(TestSelf):
     """
     Test suite to verify increase default testcase timeout working properly.
     """
 
+    @staticmethod
+    def common_sleep_function(sleep_val):
```
agrawalravi90: why did you need to create this function? couldn't you have just called time.sleep() ?

test/tests/selftest/pbs_default_timeout.py
```diff
@@ -36,32 +36,57 @@
...
         """
         Verify that test not timedout after 180 sec.
         i.e. after previous default timeout value
         """
-        time.sleep(300)
+        try:
+            self.common_sleep_function(MINIMUM_TESTCASE_TIMEOUT/2)
+        except TimeOut:
+            mssg = "Test Timeout before complete default-testcase-timeout"
+            raise TimeOut(mssg)
```
agrawalravi90: don't raise TimeOut exception, that would have been the same thing as letting it timeout without catching it. we want to just let the user know that the test failed. So, maybe do "self.fail(\<error message>)" instead of raising TimeOut?

test/tests/selftest/pbs_default_timeout.py
```diff
@@ -36,32 +36,57 @@
...
+
+    def test_timeout_greater_default_value(self):
+        """
+        Verify that test timedout after 600 sec.
+        """
+        mssg = 'Timed out after 600 second'
```
agrawalravi90: shouldn't it be 600 seconds (not second) ?

test/tests/selftest/pbs_default_timeout.py
```diff
@@ -36,32 +36,57 @@
...
-        If timeout decorator value is less than 600 then testcase timeout value
-        consider as default timeout value
+        If timeout decorator value is less than 600 then
+        default testcase timeout is considered
         """
-        time.sleep(300)
+        try:
+            self.common_sleep_function(MINIMUM_TESTCASE_TIMEOUT/2)
+        except TimeOut:
+            raise TimeOut('Test Timeout before complete expected time')
```
agrawalravi90: same comment here, don't raise TimeOut

test/tests/selftest/pbs_default_timeout.py
```diff
@@ -36,32 +36,57 @@
...
     def test_timeout_decorator_greater_default_value(self):
         """
         If timeout decorator value is greater than 600 then
         testcase timeout value consider as timeout decorator value
         """
-        time.sleep(700)
+        try:
+            self.common_sleep_function(MINIMUM_TESTCASE_TIMEOUT+1)
+        except TimeOut:
+            raise TimeOut('Test Timeout before complete expected time')
```
agrawalravi90: same comment here, don't raise TimeOut

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -156,8 +156,8 @@ def timeout(val):
     """
     Decorator to set timeout value of test case
     """
-    if val < DEFAULT_TESTCASE_TIMEOUT:
-        val = DEFAULT_TESTCASE_TIMEOUT
+    if val < MINIMUM_TESTCASE_TIMEOUT:
+        val = MINIMUM_TESTCASE_TIMEOUT
```
agrawalravi90: this is where you might wanna add a warning message

test/tests/selftest/pbs_default_timeout.py
dinesh1306: @agrawalravi90 , Initially i did the same but codacy giving error "Method could be a function". to incorporate it i make static common function and called in every test.

test/fw/ptl/utils/pbs_testsuite.py
dinesh1306: I will add warning message here, but timeout decorator evaluate initially in test suite it will comes initially in testsuite not at the test level.

test/tests/selftest/pbs_default_timeout.py
dinesh1306: This message is from timeout_handler function.

def timeout_handler(signum, frame):
    raise TimeOut('Timed out after %s second' % timeout)

test/fw/ptl/utils/pbs_testsuite.py
anamikau: @dinesh1306 thanks for the info. I suggest still adding a generic message mentioning that some of the tests are updated to use minimum_test_timeout

test/tests/selftest/pbs_default_timeout.py
agrawalravi90: i think it's safe to change the string to 'seconds', changes of the timeout period being 1 second is pretty small

test/tests/selftest/pbs_default_timeout.py
bhroam: Strange, because that is not what the instructions say: https://docs.python.org/2/library/unittest.html#skipping-tests-and-expected-failures
It says that it's more like a skip since it's an expected failure.

test/tests/selftest/pbs_default_timeout.py
agrawalravi90: no, i meant to say that you don't need this function altogether, you can just call time.sleep() instead of calling this function in your code

test/tests/selftest/pbs_default_timeout.py
agrawalravi90: If you mean this line: " In addition, it supports marking a test as an “expected failure,” a test that is broken and will fail, but shouldn’t be counted as a failure on a TestResult."

Then that refers to the decorator "@unittest.expectedFailure", not the unittest.fail() function. If you look at the logs that I posted, self.fail() basically raises an AssertionError

bhroam: I read it a bit too fast.  Thanks for pointing it out.  We can stop using self.assertTrue(False) now.  This will work for us.  Thanks for testing it out @agrawalravi90

anamikau:
Marking as WIP as @dinesh1306 is on leave and there is no one who can work on this.

test/tests/selftest/pbs_default_timeout.py
agrawalravi90: I'm okay with you leaving this as is as the message is generated by PTL, other tests might also be searching for the same string, so we can deal with that separately.

dinesh1306:
@agrawalravi90 , @anamikau I have incorporate review comments. Please Take a look.

dinesh1306:
[result.txt](https://github.com/PBSPro/pbspro/files/2726096/result.txt)
Latest Execution Result

test/fw/ptl/utils/pbs_testsuite.py
dinesh1306: @anamikau , I have added message.

test/tests/selftest/pbs_default_timeout.py
```diff
@@ -87,6 +84,7 @@ def test_timeout_decorator_greater_default_value(self):
         testcase timeout value consider as timeout decorator value
         """
         try:
-            self.common_sleep_function(MINIMUM_TESTCASE_TIMEOUT+1)
+            time.sleep(MINIMUM_TESTCASE_TIMEOUT+1)
         except TimeOut:
-            raise TimeOut('Test Timeout before complete expected time')
+            mssg = 'Test Timeout before complete expected time'
```
agrawalravi90: actually, there can be a race condition here: if the setUp takes a long time (e.g - if it has to delete a large number of jobs from the previous test), then the time taken by setUp + time.sleep(MINIMUM_TESTCASE_TIMEOUT+1) might actually cross timeout(800). So, instead of failing the test outright, have a assert to check that the timeout was for 800, if not then fail the test.

test/tests/selftest/pbs_default_timeout.py
```diff
@@ -44,28 +44,24 @@ class TestDefaultTimeout(TestSelf):
...
         Verify that test not timedout after 180 sec.
         i.e. after previous default timeout value
         """
         try:
-            self.common_sleep_function(MINIMUM_TESTCASE_TIMEOUT/2)
+            time.sleep(MINIMUM_TESTCASE_TIMEOUT/2)
         except TimeOut:
-            mssg = "Test Timeout before complete default-testcase-timeout"
-            raise TimeOut(mssg)
+            mssg = 'Test Timeout before complete default-testcase-timeout'
```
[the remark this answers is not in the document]
agrawalravi90: same comment here about the race condition, please check my comment and add a check here to see if the timeout was for the correct duration or not.

test/tests/selftest/pbs_default_timeout.py
```diff
@@ -76,9 +72,10 @@ def test_timeout_decorator_less_default_value(self):
         default testcase timeout is considered
         """
         try:
-            self.common_sleep_function(MINIMUM_TESTCASE_TIMEOUT/2)
+            time.sleep(MINIMUM_TESTCASE_TIMEOUT/2)
         except TimeOut:
-            raise TimeOut('Test Timeout before complete expected time')
+            mssg = 'Test Timeout before complete expected time'
```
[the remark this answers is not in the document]
agrawalravi90: same comment here about the race condition

test/tests/selftest/pbs_default_timeout.py
```diff
@@ -0,0 +1,104 @@
...
+        try:
+            mssg = 'sleeping for %ssec and minimum-testcase-timeout is %ssec' \
+                   % (MINIMUM_TESTCASE_TIMEOUT+1, MINIMUM_TESTCASE_TIMEOUT)
+            self.logger.info(mssg)
+            time.sleep(MINIMUM_TESTCASE_TIMEOUT+1)
+        except TimeOut, e:
+            mssg = 'Timed out after 600 second'
+            self.assertEqual(mssg, e.message)
+        else:
+            self.fail('Test is not timeout after minimum timeout period')
```
agrawalravi90: How about: "Test did not timeout after the minimum timeout period of %d"%MINIMUM_TESTCASE_TIMEOUT

test/tests/selftest/pbs_default_timeout.py
```diff
@@ -0,0 +1,104 @@
...
+        default testcase timeout is considered
+        """
+        try:
+            mssg = 'sleeping for %ssec and minimum-testcase-timeout is %ssec' \
+                   % (MINIMUM_TESTCASE_TIMEOUT/2, MINIMUM_TESTCASE_TIMEOUT)
+            self.logger.info(mssg)
+            time.sleep(MINIMUM_TESTCASE_TIMEOUT/2)
+        except TimeOut, e:
+            mssg = 'Timed out after 600 second'
+            self.assertEqual(mssg, e.message)
```
agrawalravi90: Please add an error message to assert and do this for the rest of the code as well.

dinesh1306:
@agrawalravi90 , I have incorporate review comments. Please take a look.

test/tests/selftest/pbs_default_timeout.py
```diff
@@ -55,8 +55,9 @@ def test_default_timeout(self):
             self.logger.info(mssg)
             time.sleep(MINIMUM_TESTCASE_TIMEOUT/2)
         except TimeOut, e:
-            mssg = 'Timed out after 600 second'
-            self.assertEqual(mssg, e.message)
+            mssg = 'Timed out after %s second' % MINIMUM_TESTCASE_TIMEOUT
+            err_mssg = 'failed due to mismatch in expected message'
```
agrawalravi90: this log message doesn't really explain the main reason why the test failed. It might be more useful to say something like "the test timed out for an incorrect timeout duration". If you agree, then please make change the log message in the rest of the code as well.

dinesh1306:
@agrawalravi90 ,  i made changes according to your comment. Please review.

test/tests/selftest/pbs_default_timeout.py
```diff
@@ -50,20 +50,28 @@ def test_default_timeout(self):
...
         """
         try:
+            mssg = 'sleeping for %ssec and minimum-testcase-timeout is %ssec' \
+                   % (MINIMUM_TESTCASE_TIMEOUT/2, MINIMUM_TESTCASE_TIMEOUT)
+            self.logger.info(mssg)
             time.sleep(MINIMUM_TESTCASE_TIMEOUT/2)
-        except TimeOut:
-            mssg = 'Test Timeout before complete default-testcase-timeout'
-            self.fail(mssg)
+        except TimeOut, e:
```
bhroam: Please use the modern style of except: except TimeOut as e:

test/tests/selftest/pbs_default_timeout.py
```diff
@@ -72,10 +80,13 @@ def test_timeout_decorator_less_default_value(self):
...
         """
         try:
+            mssg = 'sleeping for %ssec and minimum-testcase-timeout is %ssec' \
+                   % (MINIMUM_TESTCASE_TIMEOUT/2, MINIMUM_TESTCASE_TIMEOUT)
+            self.logger.info(mssg)
             time.sleep(MINIMUM_TESTCASE_TIMEOUT/2)
-        except TimeOut:
-            mssg = 'Test Timeout before complete expected time'
-            self.fail(mssg)
+        except TimeOut, e:
```
bhroam: same here, use 'as'

test/tests/selftest/pbs_default_timeout.py
```diff
@@ -84,7 +95,10 @@ def test_timeout_decorator_greater_default_value(self):
...
         """
         try:
+            mssg = 'sleeping for %ssec and minimum-testcase-timeout is %ssec' \
+                   % (MINIMUM_TESTCASE_TIMEOUT+1, 800)
+            self.logger.info(mssg)
             time.sleep(MINIMUM_TESTCASE_TIMEOUT+1)
-        except TimeOut:
-            mssg = 'Test Timeout before complete expected time'
-            self.fail(mssg)
+        except TimeOut, e:
```
bhroam: here as well, use 'as'

bhroam:
@dinesh1306 thanks for fixing my suggestions.  I didn't do a full review, so please continue without my signoff.

dinesh1306:
@agrawalravi90 , @anamikau Thanks for the review. Branch is rebased and squashed. Ready for merge.
[pbs_final_result.txt](https://github.com/PBSPro/pbspro/files/2749594/pbs_final_result.txt)

PR #922, opened 2018-12-13
Race conditions in PTL tests when using manager()

agrawalravi90:
#### Bug/feature Description
* Many PTL tests suffer from race conditions wrt using the manager() call, many tests set attributes via manager and move on assuming that the attributes got set even when they might not be set yet, which leads to race conditions. Many PTL calls automatically do verification, but manager() doesn't unless explicitly instructed to do so.

#### Affected Platform(s)
* All Linux

* manager()'s expect is set to False by default and it doesn't do any verification automatically

#### Solution Description
* For set and delete operations, set expect to True inside manager(). expect() doesn't correctly handle the unset operation for default attributes, so not including unset in the list for now.
* manager() now returns 0 for Success for non-zero for error
* Modified expect() to not error out for operators that it doesn't support (like INCR)

#### Testing logs/output
* Since this is for fixing race conditions, it's not possible to write a PTL test for this as such. If existing tests pass, it'll verify that the changes don't do any harm.
* Wrote a test for the secondary changes to expect, the part where it was fixed to handle unsupported operators: 
[testunsupportedoperator.log](https://github.com/PBSPro/pbspro/files/2732292/testunsupportedoperator.log)

#### Checklist:

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -6429,10 +6429,7 @@ class Server(PBSService):
         issue a management command to the server, e.g to set an
         attribute
 
-        Returns the return code of ``qmgr/pbs_manager()`` on
-        success, if expect is set to True, the return value is
-        that of the call to expect.Raises ``PbsManagerError`` on
-        error
+        Returns 0 for Success and non 0 number for Failure
 
         :param cmd: The command to issue,
                     ``MGR_CMD_[SET,UNSET, LIST,...]`` see pbs_ifl.h
@@ -6470,12 +6467,21 @@ class Server(PBSService):
                        i.e. silent mode
         :type logerr: bool
         :raises: PbsManagerError
-
-        When expect is ``False``, return the value, ``0/!0``
-        returned by pbs_manager
-        When expect is ``True``, return the value, ``True/False``,
-        returned by expect
+        :returns: On success:
+                - if expect is False, return code of qmgr/pbs_manager
+                - if expect is True, 0 for success
+        :raises: On Error/Failure:
+                - PbsManagerError if qmgr/pbs_manager() failed
+                - PtlExpectError if expect() failed
         """
+
+        # Currently, expect() doesn't validate the values being set for
+        # create operations.
+        # For unset operations, expect does not handle attributes that are
+        # reset to default after unset.
+        # So, only call expect automatically for set and delete operations.
+        if cmd in (MGR_CMD_SET, MGR_CMD_DELETE):
+            expect = True
 
         if isinstance(id, str):
             oid = id.split(',')
@@ -6689,15 +6695,17 @@ class Server(PBSService):
                     attrop = PTL_AND
 
             if oid is None:
-                return self.expect(obj_type, attrib, oid, op=op,
-                                   max_attempts=max_attempts,
-                                   attrop=attrop, offset=offset)
-            for i in oid:
-                rc = self.expect(obj_type, attrib, i, op=op,
-                                 max_attempts=max_attempts,
-                                 attrop=attrop, offset=offset)
-                if not rc:
-                    break
+                self.expect(obj_type, attrib, op=op,
+                            max_attempts=max_attempts,
+                            attrop=attrop, offset=offset)
+
+            else:
+                for i in oid:
+                    self.expect(obj_type, attrib, i, op=op,
+                                max_attempts=max_attempts,
+                                attrop=attrop, offset=offset)
+            rc = 0  # If we've reached here then expect passed, so return 0
+
         return rc
 
     def sigjob(self, jobid=None, signal=None, extend=None, runas=None,
@@ -8080,8 +8088,9 @@ class Server(PBSService):
                     PtlExpectError.
         :type msg: str or None
 
-        :returns: True if attributes are as expected and False
-                  otherwise
+        :returns: True if attributes are as expected
+
+        :raises: PtlExpectError if attributes are not as expected
         """
 
         if attempt == 0 and offset > 0:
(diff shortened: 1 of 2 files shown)


agrawalravi90:
This PR is blocked on https://github.com/PBSPro/pbspro/pull/931

agrawalravi90:
this PR is ready for review now

agrawalravi90:
@bhroam and @anamikau requesting reviews from you guys, thanks.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -6477,6 +6474,13 @@ def manager(self, cmd, obj_type, attrib=None, id=None, extend=None,
         returned by expect
         """
 
+        # expect doesn't handle 'unset' correctly for default attributes
+        # so, don't do expect automatially for the unset operation
+        # Also, expect doesn't seem to work for create, so don't
```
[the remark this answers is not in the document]
anamikau: I agree with Bhroam here. This message is very ambiguous.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -8145,6 +8157,11 @@ def expect(self, obj_type, attrib=None, id=None, op=EQ, attrop=PTL_OR,
             else:
                 operator = op
                 val = v
+            if operator not in PTL_OP_TO_STR:
+                self.logger.log(level, "Operator not supported by expect(), "
```
anamikau: please also print which operator is not supported in the error message.

anamikau: Also I feel this piece of code needs a test case.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -6689,15 +6693,22 @@ def manager(self, cmd, obj_type, attrib=None, id=None, extend=None,
...
-                                 max_attempts=max_attempts,
-                                 attrop=attrop, offset=offset)
-                if not rc:
-                    break
+                expectrc = self.expect(obj_type, attrib, oid, op=op,
+                                       max_attempts=max_attempts,
+                                       attrop=attrop, offset=offset)
+            else:
+                for i in oid:
+                    expectrc = self.expect(obj_type, attrib, i, op=op,
```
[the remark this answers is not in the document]
agrawalravi90: Thanks for catching that, I did it this way because expect's dostring says:
":returns: True if attributes are as expected and False
                  otherwise"
I'll change it to try-except

test/fw/ptl/lib/pbs_testlib.py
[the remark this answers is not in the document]
agrawalravi90: I've tried to phrase it better, please let me know if it makes sense now.

test/fw/ptl/lib/pbs_testlib.py
agrawalravi90: The problem with printing the operator is that 'operator' is just a integer into the 'range' tuple that's defined inside pbs_testlib.py. So, when i print out the operator, the log message comes out as: "Operator 2 not supported by expect(), cannot verify change in managers", which doesn't seem very useful.
I'll try writing a test case

test/fw/ptl/lib/pbs_testlib.py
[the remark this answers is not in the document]
bhroam: Or you can let the exception go through and just reword the doc string.  Or you can catch the exception and raise a manager exception.  I think either is better than catching the exception and returning 0.

test/fw/ptl/lib/pbs_testlib.py
anamikau: Thanks for explaining. Make sense to not to include it in the message then.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -6429,10 +6429,7 @@ def manager(self, cmd, obj_type, attrib=None, id=None, extend=None,
         issue a management command to the server, e.g to set an
         attribute
 
-        Returns the return code of ``qmgr/pbs_manager()`` on
-        success, if expect is set to True, the return value is
-        that of the call to expect.Raises ``PbsManagerError`` on
-        error
+        Returns 0 for Success and non 0 number for Failure
```
anamikau: instead of non-zero value for failure, manager() should also raise exception. Why are we changing that?

agrawalravi90: Thanks for catching that Anamika, I got confused because I didn't realize that expect raises an exception as well, which means that manager() used to raise an exception even when expect was set to True. I'll fix it back.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -6467,13 +6467,19 @@ def manager(self, cmd, obj_type, attrib=None, id=None, extend=None,
...
-
-        # expect doesn't work for 'create' operations. It also doesn't
-        # work for 'unset' operations for default PBS attributes.
-        # So, only do expect automatically for 'set' and 'delete'
+        :returns: On success:
+                - if expect is False, return code of qmgr/pbs_manager
+                - if expect is True, 0 for success
+                On Error/Failure:
+                - if qmgr/pbs_manager() failed, raises PbsManagerError
+                - if expect() failed, it raises PtlExpectError
```
bhroam: I'm not sure the convention with Sphinx comments.  Should these be in a :raises: tag or here?

bhroam:
The appveyor debug build failed.  I suggest rerunning it to see if it was a fluke.

bhroam:
Looks like you are out of date again.  Please rebase.

anamikau:
merged the changes in. everything looks good.

PR #933, opened 2018-12-20
Fix scheduler not logging reason denying reservation

vstumpf:
#### Bug/feature Description
* The scheduler would not log the reason why a reservation was denied; it would just say "Failed to confirm reservation".

#### Affected Platform(s)
* All

* The wrong buffer was being written into the buffer.

#### Solution Description
* Print `logmsg` instead of `errmsg` into `msgbuf`.

#### Testing logs/output
[before-fix.txt](https://github.com/PBSPro/pbspro/files/2700639/before-fix.txt)
[after-fix.txt](https://github.com/PBSPro/pbspro/files/2700640/after-fix.txt)

#### Checklist:

### DIFF
--- a/src/scheduler/resv_info.c
+++ b/src/scheduler/resv_info.c
@@ -1506,12 +1506,12 @@ confirm_reservation(status *policy, int pbs_sd, resource_resv *unconf_resv, serv
 				(void) translate_fail_code(err, NULL, logmsg);
 
 				/* If the reservation is degraded, we log a message and continue */
-				pbs_asprintf(&msgbuf, "Reservation Failed to Reconfirm: %s", logmsg);
 				if (nresv->resv->resv_substate == RESV_DEGRADED) {
+					pbs_asprintf(&msgbuf, "Reservation Failed to Reconfirm: %s", logmsg);
 					schdlog(PBSEVENT_RESV, PBS_EVENTCLASS_RESV, LOG_INFO,
 						nresv->name, msgbuf);
+					free(msgbuf);
 				}
-				free(msgbuf);
 				/* failed to confirm so move on. This will throw flow out of the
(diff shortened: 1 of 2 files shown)


src/scheduler/resv_info.c
```diff
@@ -1570,7 +1570,7 @@ confirm_reservation(status *policy, int pbs_sd, resource_resv *unconf_resv, serv
 			errmsg = "";
 
 		if (rconf == RESV_CONFIRM_FAIL)
-			pbs_asprintf(&msgbuf, "PBS Failed to confirm resv: %s", errmsg);
+			pbs_asprintf(&msgbuf, "PBS Failed to confirm resv: %s", logmsg);
```
mike0042: Why was this changed to logmsg when errmsg was assigned to the return value of pbs_geterrmsg()?

bhroam: There are two reasons rconf can equal RESV_CONFIRM_FAIL.  The first is if the scheduler tried to find space for the reservation and couldn't.  The second is if the scheduler found space, but the server rejected the pbs_confirmresv() call.  pbs_geterrmsg() only will return something in the second case.  In the first case, we will print the reason the resv couldn't fit in logmsg.

mike0042: Thanks @bhroam. A comment might be helpful in the code. Since we only use errmsg in the else block, perhaps we should set it there rather than before the conditional. Otherwise, it is a bit misleading.

vstumpf: I've moved setting the error message inside the else block, as well as adding a comment in the code.

vstumpf:
I've rebased, can someone take this forward when the checks are done?
Thanks

PR #939, opened 2019-01-04
Fix crash in PBS daemons when one with compression enabled talks to another without compression enabled

subhasisb:
#### Bug/feature Description
Recently the M4 macro that detects whether compression (zlib) is available was fixed, and thus newer builds of PBS has compression enabled for TPP. When a pbs daemon from the newer version (with compression enabled) talks to an older version (compression disabled due to the M4 macro bug), the older version daemon would crash when the sender sends it a compressed packet. The receiver detects that the packet is compressed and calls a routine to decompress it, however that routine returns NULL (since compression was compiled out) - this NULL return is not handled properly in the code and the daemon crashes.

#### Affected Platform(s)
* *All*

#### Solution Description
* This fix is not to ensure that the PBS daemon does not crash in such a circumstance.

#### Testing logs/output
* Manual test logs are attached

#### Checklist:

**Test steps: [test_steps.txt](https://github.com/PBSPro/pbspro/files/2730202/test_steps.txt)**
**Bug reproduction: [bef_fix.txt](https://github.com/PBSPro/pbspro/files/2730203/bef_fix.txt)**
**Fix validation: [after_fix.txt](https://github.com/PBSPro/pbspro/files/2730204/after_fix.txt)**

### DIFF
--- a/src/lib/Libtpp/tpp_client.c
+++ b/src/lib/Libtpp/tpp_client.c
@@ -1045,7 +1045,6 @@ tpp_send(int sd, void *data, int len)
 			tpp_log_func(LOG_CRIT, __func__, "tpp deflate failed");
 			return -1;
 		}
-		tpp_log_func(LOG_INFO, __func__, "tpp_deflate called inside tpp_send routine");
 		pkt = tpp_cr_pkt(outbuf, cmprsd_len, 0);
 		if (pkt == NULL) {
 			free(outbuf);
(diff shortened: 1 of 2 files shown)


subhasisb:
> Change looks good to me. I do have a question though - What happens when add_part_packet() returns NULL in case of decompression failure? Do we dump the packet?

Yes @arungrover we will dump the packet and thus the information from mom (in this particular case) will not be processed at all - which would mean that mom might appear down/unreachable. However, at least with this fix, the daemon will not crash. With the log message, one should then find and fix the situation so that compression is enabled on both sides

subhasisb:
Rebased and pushed now.

PR #944, opened 2019-01-09
Added new performance test for node buckets

bhroam:
#### Bug/feature Description
Added new automated performance test for node buckets

#### Solution Description
I wrote a test which compares place=free jobs using both the node bucket algorithm and the normal node search algorithm.

#### Testing logs/output
[node_bucket.log](https://github.com/PBSPro/pbspro/files/2743133/node_bucket.log)

#### Checklist:

### DIFF
--- a/test/fw/ptl/utils/pbs_logutils.py
+++ b/test/fw/ptl/utils/pbs_logutils.py
@@ -103,10 +103,9 @@ Analyze ``server``, ``scheduler``, ``MoM``, and ``accounting`` logs.
     utilization
 """
 
-
-tm_re = "(?P<datetime>\d\d/\d\d/\d{4,4}[\s]\d\d:\d\d:\d\d)"
-job_re = ";(?P<jobid>[\d\[\d*\]]+)\."
-fail_re = ";(?P<jobid>[\d\[\[]+)\."
+tm_re = r'(?P<datetime>\d\d/\d\d/\d{4}\s\d\d:\d\d:\d\d(\.\d{6})?)'
+job_re = r';(?P<jobid>[\d\[\d*\]]+)\.'
+fail_re = r';(?P<jobid>[\d\[\[]+)\.'
 
 # Server metrics
 NUR = 'node_up_rate'
@@ -210,6 +209,8 @@ PARSER_OK_STOP = 1
 PARSER_ERROR_CONTINUE = 2
 PARSER_ERROR_STOP = 3
 
+epoch_datetime = datetime.datetime.fromtimestamp(0)
+
 
 class PBSLogUtils(object):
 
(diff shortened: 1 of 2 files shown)


test/tests/performance/pbs_node_buckets_perf.py
```diff
@@ -65,13 +65,43 @@ def cust_attr_func(self, name, totalnodes, numnode, attribs):
...
+            jid = self.server.submit(J)
+            jids.append(jid)
+            
+        return jids
+    
+    def run_cycle(self):
+        """
+        Run a cycle and return the length of the cycle
+        """
+        t = int(time.time())
```
agrawalravi90: If the scheduler was already in a scheduling cycle when you executed this then there could be a race condition in this function. I think we should first check that server/scheduler state is idle before doing anything in this function.

test/tests/performance/pbs_node_buckets_perf.py
```diff
@@ -65,13 +65,43 @@ def cust_attr_func(self, name, totalnodes, numnode, attribs):
...
+    def test_node_bucket_perf_scatter(self):
         """
         Submit a large number of jobs which use node buckets.  Run a cycle and
         compare that to a cycle that doesn't use node buckets.  Jobs require
-        place=scatter:excl to use node buckets.
+        place=excl to use node buckets.
+        This test uses place=scatter.  Scatter placement is quicker than free
         """
         num_jobs = 1000
         a = {'Resource_List.select': '1429:ncpus=1:color=yellow',
```
agrawalravi90: I know this is old code, but can you please explain why you are submitting this job before proceeding with submitting the large number of jobs?

test/tests/performance/pbs_node_buckets_perf.py
bhroam: Good point.  When I added the test, we were in cycle.  I hadn't turned scheduling off before submitting all 1000 jobs.  I did that and checked for server_stat = Idle.

test/tests/performance/pbs_node_buckets_perf.py
bhroam: It's to make sure the scheduler can't run any of the 1000 jobs.  We want to treat each of the jobs equally, so that means we need to either run them all or not run any.  Submitting one first means we won't run any.  I added a code comment.

bhroam:
@agrawalravi90 I updated the tests and cleaned them up a bit.  Since the two tests practically the same, I factored out the entire test and put it into a helper function.  Now the two tests call the helper function with a place statement and the number of jobs.  I modified run_cycle() as you suggested.  I make sure we're not in a cycle.  I kick a cycle and make sure we're in a cycle, and then I wait for that cycle to end.  It's now all done via the server_state attribute instead of parsing logs.  Let me know what you think.

agrawalravi90:
Codacy didn't pass, it's complaining about trailing whitespaces, please fix them.

test/tests/performance/pbs_node_buckets_perf.py
```diff
@@ -66,69 +69,97 @@ def cust_attr_func(self, name, totalnodes, numnode, attribs):
...
+            self.server.alterjob(jid, a)
+
+        cycle2_time = self.run_cycle()
 
-        c = self.scheduler.cycles(lastN=1)[0]
-        cycle2_time = c.end - c.start
         self.logger.info('Cycle 1: %d Cycle 2: %d Cycle time difference: %d' %
                          (cycle1_time, cycle2_time, cycle1_time - cycle2_time))
-        self.assertTrue(cycle1_time > cycle2_time)
+        self.assertGreater(cycle1_time, cycle2_time)
```
visheshh: Hi @bhroam , Can you make use of set_test_measurements API so that the test results are stored in the JSON output file .
self.set_test_measurements({"test_measure": cycle_time_without_node_buckets,"value": cycle_time})
self.set_test_measurements({"test_measure": cycle_time_with_node_buckets,"value": cycle2_time}).
We are planning to store all results from all the performance tests to json file , So we can use them to compare performance across releases.

test/fw/ptl/utils/pbs_logutils.py
```diff
@@ -920,9 +919,9 @@ class PBSServerLog(PBSLogAnalyzer):
     tm_tag = re.compile(tm_re)
     server_run_tag = re.compile(tm_re + ".*" + job_re + ".*;Job Run at.*")
     server_nodeup_tag = re.compile(tm_re + ".*Node;.*;node up.*")
-    server_enquejob_tag = re.compile(tm_re + ".*" + job_re +
+    server_enquejob_tag = re.compile(tm_re + ".*" + job_re +
```
agrawalravi90: looks like your editor configured to add spaces after  '+'

bhroam:
@agrawalravi90 Pydev changed its default python formatter.  It was being run when I saved and added the spaces(weird).  I changed it back to running autopep8.  You should make sure it didn't change for you as well.

PR #958, opened 2019-01-22
PTL test case test_qrun_hangs fails because of race condition

SudeshnaMoulik:
#### Bug/feature Description
qrun_hangs test case fails in some platform because of race condition.

#### Affected Platform(s)

- ARM
- SLES15  OS

Previously we didn't wait for the scheduler log to end the cycle. Because of this below problems have observed in some platforms:
- Parent process were doing log_match from previous cycle and trying to kill the child process. Even though child process did runjob().

expect on server a01: scheduling set False && server_state set Idle server a01.pbspro.com ...  OK
run on a01: 501.a01
a01: /opt/pbs/bin/qrun 501.a01
scheduler a01 log match: searching for "Starting Scheduling Cycle" - on all matches ... OK
Runjob hung. Child process exit.
err: ["'unknown': unknown terminal type.", "'unknown': unknown terminal type."]
FAILED

- If the parent process executed before child process does runjob, it matches the previous cycle log. While it tried to kill child process below error thrown:

expect on server x55-s12: scheduling set False && server_state set Idle server x55-s12.pbspro.com ...  OK
log match: searching for "Starting Scheduling Cycle" - on all matches ... OK
Runjob hung. Child process exit.
err: ["'unknown': I need something more specific.", "'unknown': I need something more specific."]

#### Solution Description
Synchronized the event of log_match in parent process and runjob in child process. Before forking waited for scheduler to end the current cycle and noted the time. So that after fork the parent process will wait to log_match for the current cycle triggered from child process.

#### Testing logs/output
[sles12.txt](https://github.com/PBSPro/pbspro/files/2781424/sles12.txt)
[sles15.txt](https://github.com/PBSPro/pbspro/files/2781425/sles15.txt)
[arm.txt](https://github.com/PBSPro/pbspro/files/2781426/arm.txt)

#### Checklist:

### DIFF
--- a/test/tests/functional/pbs_qrun.py
+++ b/test/tests/functional/pbs_qrun.py
@@ -115,11 +115,12 @@ class TestQrun(TestFunctional):
             else:
                 self.server.submit(j)
         self.logger.info("Submitted 500 jobs with different walltime")
-        now = int(time.time())
         self.server.manager(MGR_CMD_SET, SERVER,
                             {'scheduling': 'True'})
         self.server.manager(MGR_CMD_SET, SERVER,
                             {'scheduling': 'False'}, expect=True)
+        time.sleep(1)
+        now = int(time.time())
         pid = os.fork()
         if pid == 0:
             try:
@@ -128,14 +129,13 @@ class TestQrun(TestFunctional):
             except PbsRunError as e:
                 self.logger.info("Runjob throws error: " + e.msg[0])
         else:
-            out = self.scheduler.log_match(
-                "Starting Scheduling Cycle",
-                allmatch=True, interval=5, starttime=now, max_attempts=10)
-            if len(out) < 2:
+            try:
+                self.scheduler.log_match("Starting Scheduling Cycle",
+                                         interval=5, starttime=now,
+                                         max_attempts=10)
+                self.logger.info("No hangs. Parent process exit")
+            except PtlLogMatchError:
                 os.kill(pid, signal.SIGKILL)
                 os.waitpid(pid, 0)
                 self.logger.info("Runjob hung. Child process exit.")
-                self.assertFalse(True,
-                                 "Qrun didn't start another sched cycle")
-            else:
-                self.logger.info("No hangs. Parent process exit")
+                self.fail("Qrun didn't start another sched cycle")


test/tests/functional/pbs_qrun.py
```diff
@@ -115,11 +115,13 @@ def test_qrun_hangs(self):
             else:
                 self.server.submit(j)
         self.logger.info("Submitted 500 jobs with different walltime")
-        now = int(time.time())
         self.server.manager(MGR_CMD_SET, SERVER,
                             {'scheduling': 'True'})
         self.server.manager(MGR_CMD_SET, SERVER,
                             {'scheduling': 'False'}, expect=True)
+        now = int(time.time())
+        self.scheduler.log_match("Leaving Scheduling Cycle", starttime=now)
```
agrawalravi90: I'm not sure I understand how this will work. Before setting 'now', you make sure that the scheduling cycle is finished. So, there shouldn't be any "Leaving Scheduling Cycle" message in sched logs since 'now' right?

test/tests/functional/pbs_qrun.py
```diff
@@ -115,11 +115,13 @@ def test_qrun_hangs(self):
...
+        now = int(time.time())
```
agrawalravi90: if you are setting 'now' after ensuring that the scheduler is not in a sched cycle, then do you really need to do the log_match() above?

test/tests/functional/pbs_qrun.py
```diff
@@ -128,14 +130,14 @@ def test_qrun_hangs(self):
...
+            try:
+                self.scheduler.log_match("Starting Scheduling Cycle",
+                                         interval=5, starttime=now,
+                                         max_attempts=10)
+                self.logger.info("No hangs. Parent process exit")
+            except PtlLogMatchError:
                 os.kill(pid, signal.SIGKILL)
                 os.waitpid(pid, 0)
                 self.logger.info("Runjob hung. Child process exit.")
                 self.assertFalse(True,
```
agrawalravi90: please use 'self.fail' instead of doing this

test/tests/functional/pbs_qrun.py
SudeshnaMoulik: Thanks @agrawalravi90  for review.
self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'False'}, expect=True), will make sure in server the "scheduling" attribute is set false. But in some platform, in scheduler log the current cycle is still going on, so the log is coming after we set the server attribute.
But yes, it will be better if time "now" set before we set scheduling false.

test/tests/functional/pbs_qrun.py
SudeshnaMoulik: Yes the log match for "Leaving Scheduling Cycle" is necessary. 
1. The first "now" am setting because we need the current cycle we just started that needs to complete, in sched log. Otherwise, even though in server the "scheduling" attribute is set false, but in sched log the end message of current cycle still not received.

2. For fast system, the first now timestamp can also be same as when we are starting the cycle. So when we start checking log_match() in parent process after fork, it matches with the 1st cycle starting log.
To overcome this, noticed the timestamp when we got the "Leaving Scheduling Cycle" message in sched log. Started next log match after this time, to check whether new sched cycle triggered by runjob from child.

test/tests/functional/pbs_qrun.py
SudeshnaMoulik: self.fail() this will fail the PTL test case. But what if the child process hung, then this test case will wait for timeout period runjob. May be instead of waiting that longer, after checking few times in the the sched log if new cycle not started we can kill the child, as we know it didnt start the cycle and test case hung.

test/tests/functional/pbs_qrun.py
agrawalravi90: Actually, PTL does more than just check that 'scheduling' is set to False. It also checks that the server state is 'idle' before giving control back to the test. This means that by the time the test executes  "now = int(time.time())", the server should already be in the 'idle' state, meaning that the sched cycle should be finished.

test/tests/functional/pbs_qrun.py
agrawalravi90: self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'False'}, expect=True) should ensure that the scheduling cycle is finished before returning control back to the test. If that's not happening then we should fix the PTL framework, not this test.

test/tests/functional/pbs_qrun.py
agrawalravi90: i agree with killing the child process, I just wanted you to replace the self.assertFalse(True, .. ) with self.fail(...)

test/tests/functional/pbs_qrun.py
agrawalravi90: There can be a case where on fast machines the sched cycle finishes in the same second as when you set 'now'. So, try adding a sleep(1) after self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'False'}, expect=True) and before setting 'now'. That should ensure that log_match() only reads the next sched cycle.

SudeshnaMoulik: Yes then that should be useful. I will check for microsecond logging.

agrawalravi90: ah, ok, my bad, i thought it was there. @SudeshnaMoulik i guess you should go the route of adding a sleep(1) before setting 'now', sorry for making you go through a fruitless path.

test/tests/functional/pbs_qrun.py
SudeshnaMoulik: Yes @bhroam 
If we see the below log, even though in server "scheduling= false/ expect=True", still the scheduler cycle is not ended according to the sched log.

2019-01-21 02:07:53,242 INFOCLI a01: /opt/pbs/bin/qmgr -c set server scheduling=True
2019-01-21 02:07:53,280 INFO manager on a01: set server {'scheduling': 'False'}
2019-01-21 02:07:53,280 INFOCLI a01: /opt/pbs/bin/qmgr -c set server scheduling=False
2019-01-21 02:07:53,371 INFOCLI a01: /opt/pbs/bin/qstat -Bf a01.pbspro.com
**2019-01-21 02:07:53,406 INFO expect on server a01: scheduling set False && server_state set Idle server a01.pbspro.com ... OK**
**2019-01-21 02:07:53,414 INFO scheduler a01 log match: searching for "Leaving Scheduling Cycle" - No match
2019-01-21 02:07:54,015 INFO scheduler a01 log match: searching for "Leaving Scheduling Cycle"... OK**
2019-01-21 02:07:54,018 INFO run on a01: 499.a01

SudeshnaMoulik:
@agrawalravi90 have changed the code, tested among several systems.
Can you please review the changes.

SudeshnaMoulik:
[PTL.txt](https://github.com/PBSPro/pbspro/files/2785825/PTL.txt)

test/tests/functional/pbs_qrun.py
```diff
@@ -128,14 +129,13 @@ def test_qrun_hangs(self):
...
+                self.logger.info("No hangs. Parent process exit")
+            except PtlLogMatchError:
                 os.kill(pid, signal.SIGKILL)
                 os.waitpid(pid, 0)
                 self.logger.info("Runjob hung. Child process exit.")
-                self.assertFalse(True,
-                                 "Qrun didn't start another sched cycle")
-            else:
-                self.logger.info("No hangs. Parent process exit")
+                self.fail("Qrun didn't start another sched cycle")
```
agrawalravi90: you are failing the test if log_match() is successful? This line should be part of the except block right?

agrawalravi90: d'oh! you removed the else block completely, my bad

agrawalravi90:
@SudeshnaMoulik please rebase your branch in preparation for merge

SudeshnaMoulik:
Thanks @npadole20 for review the changes.
@agrawalravi90 rebased the branch. Can you please merge.

PR #979, opened 2019-02-04
holidays file enhancements

agrawalravi90:
#### Feature Description
* Holidays file has information about prime/non-prime windows for days of the week for a site. Right now, there's no documented behavior for what happens if data for any of the days is missing or if the data is inconsistent. Also, the behavior for when the year entry is missing is not clearly defined. The general use case, as well as the theme of the prime-time feature in PBS, seems to be to assume prime time if data is insufficient, so it seems like a good idea to officially enforce this behavior. It is also desired to have an empty/commented out holidays file which doesn't need to be updated every year so that it is all prime-time by default.

#### Affected Platform(s)
* All

#### Design
* EDD: https://pbspro.atlassian.net/wiki/spaces/PD/pages/1046151225/Holidays+file+missing+inconsistent+days+will+be+treated+as+ALL+PRIME
* Forum discussion:
http://community.pbspro.org/t/holidays-file-missing-inconsistent-days-will-be-treated-as-all-prime/1440
http://community.pbspro.org/t/wrong-holidays-file-makes-scheduler-infinite-loop/1387/

#### Solution Description
* Now, after parsing holidays file, if there's data missing for a day, we enforce 'all' prime-time for it
* If an entry in holidays file sets both prime and non-prime start times to 'all'/'none', we enforce 'all' prime-time for it.
* The holidays file will now be completely commented out by default, so it'll be 24x7 prime-time

#### Testing logs/output
* [testholidays.log](https://github.com/PBSPro/pbspro/files/2850263/testholidays.log)

#### Checklist:

- [X] My change requires project documentation. See **[required documentation checklist](https://pbspro.atlassian.net/wiki/display/DG/Checklist+for+Developing+Features+and+Bug+Fixes)** for details.
   - [X] I have added documentation in the **[project documentation area](https://pbspro.atlassian.net/wiki/display/PD)**.

### DIFF
--- a/src/scheduler/pbs_holidays
+++ b/src/scheduler/pbs_holidays
@@ -1,77 +1,58 @@
-*
- * Copyright (C) 1994-2018 Altair Engineering, Inc.
- * For more information, contact Altair at www.altair.com.
- *
- * This file is part of the PBS Professional ("PBS Pro") software.
- *
- * Open Source License Information:
- *
- * PBS Pro is free software. You can redistribute it and/or modify it under the
- * terms of the GNU Affero General Public License as published by the Free
- * Software Foundation, either version 3 of the License, or (at your option) any
- * later version.
- *
- * PBS Pro is distributed in the hope that it will be useful, but WITHOUT ANY
- * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
- * FOR A PARTICULAR PURPOSE.
- * See the GNU Affero General Public License for more details.
- *
- * You should have received a copy of the GNU Affero General Public License
- * along with this program.  If not, see <http://www.gnu.org/licenses/>.
- *
- * Commercial License Information:
- *
- * For a copy of the commercial license terms and conditions,
- * go to: (http://www.pbspro.com/UserArea/agreement.html)
- * or contact the Altair Legal Department.
- *
- * Altair’s dual-license business model allows companies, individuals, and
- * organizations to create proprietary derivative works of PBS Pro and
- * distribute them - whether embedded or bundled with other software -
- * under a commercial license agreement.
- *
- * Use of Altair’s trademarks, including but not limited to "PBS™",
- * "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
- * trademark licensing policies.
-*
-
-*
-YEAR  2018
-*
-* Prime/Nonprime Table 
-*
-*   Prime Non-Prime
-* Day   Start Start
-*
-  weekday 0600  1730
-  saturday  none  all
-  sunday  none  all
-*
-* Day of  Calendar  Company
-* Year    Date    Holiday
-*
-
-* if a holiday falls on a saturday, it is observed on the friday before
-* if a holiday falls on a sunday, it is observed on the monday after
-
-* Jan 1 
-    1           Jan 1           New Year's Day
-* Third Monday of Jan
-   15           Jan 15          Martin Luther King Day
-* Third Monday in Feb
-   50           Feb 19          President's Day
-* Last Monday in May
-  148           May 28          Memorial Day
-* July 4th
-  185           Jul 4           Independence Day
-* First Monday in Sept 
-  246           Sep 3           Labor Day
-* Second Monday in Oct
-  281           Oct 8           Columbus Day
-* Nov 11 
-  316           Nov 12          Veterans Day
-* Fourth Thursday in Nov
-  326           Nov 22          Thanksgiving
-* Dec 25 
-  359           Dec 25          Christmas Day
+#
+# Copyright (C) 1994-2019 Altair Engineering, Inc.
+# For more information, contact Altair at www.altair.com.
+#
+# This file is part of the PBS Professional ("PBS Pro") software.
+#
+# Open Source License Information:
+#
+# PBS Pro is free software. You can redistribute it and/or modify it under the
+# terms of the GNU Affero General Public License as published by the Free
+# Software Foundation, either version 3 of the License, or (at your option) any
+# later version.
+#
+# PBS Pro is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.
+# See the GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+# Commercial License Information:
+#
+# For a copy of the commercial license terms and conditions,
+# go to: (http://www.pbspro.com/UserArea/agreement.html)
+# or contact the Altair Legal Department.
+#
+# Altair’s dual-license business model allows companies, individuals, and
+# organizations to create proprietary derivative works of PBS Pro and
+# distribute them - whether embedded or bundled with other software -
+# under a commercial license agreement.
+#
+# Use of Altair’s trademarks, including but not limited to "PBS™",
+# "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
+# trademark licensing policies.
+#
+#
+# UNCOMMENT AND CHANGE THIS TO THE CURRENT YEAR
+#YEAR  1970
+#
+# Prime/Nonprime Table 
+#
+#   Prime Non-Prime
+# Day   Start Start
+#
+# UNCOMMENT AND SET THE REQUIRED PRIME/NON-PRIME START TIMES
+#  weekday 0600  1730
+#  saturday  none  all
+#  sunday  none  all
+#
+# Day of  Calendar  Company
+# Year    Date    Holiday
+#
+#
+# UNCOMMENT AND ADD CALENDAR HOLIDAYS TO BE CONSIDERED AS NON-PRIME DAYS
+#    1           Jan 1           New Year's Day
+#  359           Dec 25          Christmas Day
 
--- a/src/scheduler/prime.c
+++ b/src/scheduler/prime.c
@@ -229,6 +229,34 @@ is_holiday(long date)
 }
 
 /**
+ * @brief	Set conf.prime values to reflect "ALL PRIME" before we start parsing
+ * 			the holidays file
+ *
+ * @param	void
+ *
+ * @return void
+ */
+static void
+handle_missing_prime_info(void)
+{
+	enum days d;
+
+	for (d = SUNDAY; d < HIGH_DAY; d++) {
+		if (conf.prime[d][PRIME].all + conf.prime[d][PRIME].none
+				+ conf.prime[d][PRIME].hour + conf.prime[d][PRIME].min == 0) {
+			conf.prime[d][PRIME].all = TRUE;
+			conf.prime[d][PRIME].none = FALSE;
+			conf.prime[d][PRIME].hour = UNSPECIFIED;
+			conf.prime[d][PRIME].min = UNSPECIFIED;
+			conf.prime[d][NON_PRIME].none = TRUE;
+			conf.prime[d][NON_PRIME].all = FALSE;
+			conf.prime[d][NON_PRIME].hour = UNSPECIFIED;
+			conf.prime[d][NON_PRIME].min = UNSPECIFIED;
+		}
+	}
+}
+
+/**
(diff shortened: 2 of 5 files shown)


src/scheduler/prime.c
```diff
@@ -228,6 +228,34 @@ is_holiday(long date)
...
+/**
+ * @brief	Set conf.prime values to reflect "ALL PRIME" before we start parsing
+ * 			the holidays file
+ *
+ * @param	void
+ *
+ * @return void
+ */
+static void
+handle_missing_prime_info(void)
```
bhroam: This should happen in init_config()

test/tests/functional/pbs_holidays.py
```diff
@@ -0,0 +1,148 @@
...
+        """
+        Helper function to test that it is prime-time right now
+        """
+        # Submit a job to the non-prime queue
+        a = {ATTR_queue: self.np_queue}
+        j1 = Job(attrs=a)
+        jid1 = self.server.submit(j1)
+
+        # Verify that the job does not run as it is 24hrs prime-time
+        self.server.expect(JOB, {'job_state': 'Q'}, jid1)
```
bhroam: Check to see that comment is set before checking for job_state: Q.  There is a race that you do your expect() before the scheduler even sees the job.

src/scheduler/prime.c
```diff
@@ -467,16 +499,30 @@ load_day(enum days d, enum prime_time pr, char *tok)
 
 	if (tok != NULL) {
 		if (!strcmp(tok, "all") || !strcmp(tok, "ALL")) {
+			if (pr == NON_PRIME && conf.prime[d][PRIME].all == TRUE) {
+				/*
+				 * The user specified both all prime and all non-prime for this day
+				 * all prime wins in such a situation, so just ignore the non-prime info
+				 */
```
bhroam: Turn this comment into a more concise log message.  This is an error case, so we shouldn't just assume something without saying it.

src/scheduler/prime.c
```diff
@@ -467,16 +499,30 @@ load_day(enum days d, enum prime_time pr, char *tok)
...
 			conf.prime[d][pr].min = UNSPECIFIED;
-		}
-		else if (!strcmp(tok, "none") || !strcmp(tok, "NONE")) {
-			conf.prime[d][pr].none = TRUE;
+			conf.prime[d][pr].none = FALSE;
+		} else if (!strcmp(tok, "none") || !strcmp(tok, "NONE")) {
+			if (pr == NON_PRIME && conf.prime[d][PRIME].none == TRUE) {
+				/*
+				 * The user specified both none prime and none non-prime for this day
+				 * Make it all prime-time for this day as we don't know better
```
bhroam: Same as above about the log message.

src/scheduler/data_types.h
```diff
@@ -1069,7 +1069,7 @@ struct config
 	fairshare_head *fairshare;		/* fairshare tree */
 	time_t decay_time;			/*  time in seconds for the decay period*/
 	time_t sync_time;			/* time between syncing usage to disk */
-	struct t prime[HIGH_DAY][HIGH_PRIME];	/* prime time start and prime time end*/
+	struct t prime[HIGH_DAY][2];	/* prime time start and prime time end*/
```
bhroam: Why the change from HIGH_PRIME to 2?

agrawalravi90: the other 2 rows (corresponding to ALL and NONE) were not used, so it was just unused memory.

src/scheduler/prime.c
agrawalravi90: Just to be sure, you want me to actually log a message to sched logs here, or just trim the comment?

src/scheduler/prime.c
agrawalravi90: I wanted to do the same thing, but, if somebody adds "all all" as prime and non prime start in holidays file, then we need to be able to catch that in parse_holidays and set all prime-time, not all non-prime time. We don't have a good way of catching that if we set the bits to "all prime" inside init_config() itself as we won't be able to tell inside load_day() whether the bits are all prime because the holidays file has "all" entry for prime, or if it is what was set by init_config(). So, we could refactor the function if you really hate this, but otherwise i can't put it inside init_config().

src/scheduler/prime.c
bhroam: The comment as a comment was fine.  I want a scheduler log message.  If we're adding a scheduler log message, I don't think we need both a comment and the log message.  For a log message, the comment is a bit wordy.

src/scheduler/data_types.h
bhroam: This will probably never happen, but the idea was if we ever add more time windows than just prime and non-prime, this will just automatically update itself.  We'll add a new value to the enum and HIGH_PRIME will increase.  The amount of lost memory is pretty minor.

src/scheduler/prime.c
bhroam: I don't understand.  You call handle_missing_prime_info() right at the top of parse_holidays().  There should be no difference between the same code being called in init_config() or in parse_holidays()

agrawalravi90: No, I call it at the end after parsing the holidays file

agrawalravi90:
@bhroam thanks for your comments, I've addressed them. I also made some additional changes according to the comments on the design from the forums. Please provide feedback.

test/tests/functional/pbs_holidays.py
```diff
@@ -71,15 +71,39 @@ def check_prime(self):
...
         j2 = Job(attrs=a)
         jid2 = self.server.submit(j2)
 
-        # Verify that the job runs as it is 24hrs prime-time
+        # Verify that the job runs as it is all prime-time
+        self.server.expect(JOB, {'job_state': 'R'}, jid2)
+
+    def check_nonprime(self):
+        """
+        Helper function to test that it is non-prime time right now
```
bhroam: If the log_filter isn't filtering DEBUG2, the scheduler will tell you what prime status it is in and when it ends in the log.  Consider using the log message.

test/tests/functional/pbs_holidays.py
```diff
@@ -146,3 +170,117 @@ def test_only_year(self):
...
+
+        # Set today as all non-prime time
+        self.scheduler.holidays_set_day(today, "none", "all")
+        self.check_nonprime()
+
+    def test_missing_year(self):
+        """
+        Test that scheduler assumes all prime time if the year entry is
+        missing from holidays file
+        """
```
bhroam: You really have tested this above.  You are testing what the scheduler will do with a completely commented out holidays file.  Above, the year was comment out, and here the commented out year is missing.  In either case the entire file is commented out.

src/scheduler/pbs_holidays
```diff
@@ -34,44 +34,25 @@
  * "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
  * trademark licensing policies.
 *
-
 *
-YEAR  2018
+* UNCOMMENT AND CHANGE THIS TO THE CURRENT YEAR
+*YEAR  1970
```
bhroam: As I said in the forum post, use '#' to comment out the lines.  The '#'  always worked, but '*' was the correct comment character for this file to be compliant with UNICOS.

test/tests/functional/pbs_holidays.py
```diff
@@ -146,3 +170,117 @@ def test_only_year(self):
...
+        h_content = self.du.cat(filename=self.scheduler.holidays_file,
+                                sudo=True)['out']
+        self.assertEqual(len(h_content), 3)
+
+        self.check_prime()
+
+        # Verify that scheduler didn't log a message for out of date file
+        msg = "holidays;The holiday file is out of date; please update it."
+        self.scheduler.log_match(msg, starttime=ctime, existence=False,
+                                 max_attempts=5)
```
bhroam: One of the use cases of commenting out the holiday file is so the scheduler doesn't add prime status events to the calendar.  Consider writing a test to check for that.  To do it, you need to set log_filter to 2048.  You then add a job that takes up all the resources without requesting walltime.  You then add a second job that will get calendared.  You look in the log that no policy change events happened.

test/tests/functional/pbs_holidays.py
agrawalravi90: I know, but I'd like to keep it just to be thorough.

src/scheduler/pbs_holidays
agrawalravi90: Ok, I'll make it '#' and keep the check for '*' for backward compatibility.

btw, I actually don't see your comment in the forum, could you please post it once more? Thanks.

agrawalravi90:
@bhroam I've addressed your comments, please provide further feedback. Thanks!

test/tests/functional/pbs_holidays.py
bhroam: Then make the test different.  Set non-prime such that it should be in non-prime, but since the year is missing it will be 24hr prime.

bhroam:
Codacy is complaining about an unused variable in your test.

bhroam:
My only comment is the one I mentioned above for the test without a YEAR line.

test/tests/functional/pbs_holidays.py
agrawalravi90: Actually I think the test already does that, it sets 24x7 non-prime time and then checks that it's 24x7 prime-time because of the missing year entry

agrawalravi90:
@bhroam thanks for the review, I've addressed your comments. I also decided to add 2 more PTL tests to cover some corner cases. Please review the latest commit and provide feedback. Thanks.

test/tests/functional/pbs_holidays.py
bhroam: It has 3 lines that are commented out.  I'm saying don't have them commented out.  Have them set non-prime.  This way we can easily check if 24hr prime is working.

test/tests/functional/pbs_holidays.py
```diff
@@ -334,10 +332,64 @@ def test_prime_events_calendar(self):
...
+        # Add a weekday entry with all prime-time
+        self.scheduler.holidays_set_day("weekday", "all", "none")
+
+        # Set today as all non-prime time
+        today_idx = datetime.datetime.today().weekday()
+        today = self.days[today_idx]
+        self.scheduler.holidays_set_day(today, "none", "all")
+
+        h_content = self.du.cat(filename=self.scheduler.holidays_file,
+                                sudo=True)['out']
```
bhroam: Why are you catting the file?  It isn't used anywhere.

bhroam:
My only comment is about the test with the YEAR line.  I'm fine after we come to consensus about that test.

test/tests/functional/pbs_holidays.py
agrawalravi90: D'oh! Those lines were never meant to commented out, sorry about that, yes, the test would be very much the same as the previous one with them commented out :) I've uncommented them now, please let me know if it looks good now.

agrawalravi90:
Thanks a lot for the thorough review Bhroam!
@arungrover could you please review this PR as well?

src/scheduler/prime.c
```diff
@@ -467,16 +501,26 @@ load_day(enum days d, enum prime_time pr, char *tok)
...
+				return 0;
+			}
 			conf.prime[d][pr].all = TRUE;
 			conf.prime[d][pr].hour = UNSPECIFIED;
 			conf.prime[d][pr].min = UNSPECIFIED;
-		}
-		else if (!strcmp(tok, "none") || !strcmp(tok, "NONE")) {
-			conf.prime[d][pr].none = TRUE;
+			conf.prime[d][pr].none = FALSE;
+		} else if (!strcmp(tok, "none") || !strcmp(tok, "NONE")) {
```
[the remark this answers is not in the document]
agrawalravi90: I'm not sure if values like NoNe or NOne are supposed to be allowed. Since you left it up to me, I'd like to keep the code as is for now. Thanks for being diligent :)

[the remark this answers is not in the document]
mike0042: +1 for strcasecmp()
+2 for strncasecmp()

agrawalravi90:
@arungrover and @bhroam thanks for the sign-offs guys, I've rebase the branch, please merge it if possible. Thanks!

bhroam:
Thanks @agrawalravi90.  I've merged in the PR.

PR #613, opened 2018-03-27
PP-1252: daemonized qsub causes foreground qsub failures when it decides to quit.

praveenhandigol:
#### Bug/feature Description
* *daemonized qsub causes foreground qsub failures when it decides to quit.*

* *If there is an issue ID, link to it: [PP-1252](https://pbspro.atlassian.net/browse/PP-1252)*

#### Affected Platform(s)
* *Linux*

#### Cause / Analysis / Design
* Foreground qsub process fails to submit the jobs after background qsub hits the CREDENTIAL_LIFETIME timeout of 60 mins.
The issue is not reproducible locally.
* *If there is an external design, link to it **[project documentation area](https://pbspro.atlassian.net/wiki/display/PD)***
* *If there is a discussion on the Forum, link to it **[pbspro community forum](http://community.pbspro.org/)***

#### Solution Description
Background qsub daemon timeout has been reduced to 30 mins.
After the 30 mins timeout, the UNIX domain socket file is being unlinked and the daemon waits for 5 seconds to service the existing connection if any.
Fix on Windows platform:
1.	Problem is consistently reproducible on Windows platform.
2.	I tried incorporating the similar fix that of Linux i.e. disconnecting the pipe and retaining it for 
seconds before closing it.
3.	But, the behaviour of DisconnectNamedPipe API on windows is little different than that of unlink() call on linux. Unlinking the UD socket on linux still retains the existing connection object, but DisconnectNamedPipe doesn’t do that. It just allows the client to re-connect before doing the closehandle.
4.	So, I just moved the code for checking the Credential timeout (30 mins) much before WaitForMultipleObjects call Just after while(1) loop and problem is resolved. I could see a different qsub process id coming up after 29 mins of cred time-out and no error message is being displayed.

#### Checklist:

- [x] I have joined the **[pbspro community forum](http://community.pbspro.org/)**.
- [x] My pull request contains a **single, signed** commit. See **[setting up gpg signature](https://pbspro.atlassian.net/wiki/display/DG/Signing+Your+Git+Commits)**.
- [x] My code follows the **[coding style](https://pbspro.atlassian.net/wiki/display/DG/Coding+Standards)** of this project.
- [ ] My change requires project documentation. See **[required documentation checklist](https://pbspro.atlassian.net/wiki/display/DG/Checklist+for+Developing+Features+and+Bug+Fixes)** for details.
- [ ] I have added documentation in the **[project documentation area](https://pbspro.atlassian.net/wiki/display/PD)**.
[...]

PR #984, opened 2019-02-07
qsub background daemon timesout every minute after CREDENTIAL_LIFETIME expiry

visheshh:
#### Bug/feature Description
qsub background daemon remains for first half an hour till its CREDENTIAL_LIFETIME i.e, 30 mins .
After CREDENTIAL_LIFETIME lifetime expiry every minute a background daemon exits and new process is spawned.

#### Affected Platform(s)
All

file descriptor was unlinked twice in the case of cred_timeout which caused to unlink the fd of a new daemon.

#### Solution Description
If the file is already unlinked do not unlink again

#### Testing logs/output
[test logs.txt](https://github.com/PBSPro/pbspro/files/2840403/test.logs.txt)

#### Checklist:

### DIFF
--- a/src/cmds/qsub.c
+++ b/src/cmds/qsub.c
@@ -5486,7 +5486,8 @@ do_daemon_stuff(void)
 
 out:
 	close(bindfd);
-	unlink(fl);
+	if (cred_timeout != 1)
+		unlink(fl);
 	exit(0);
 
 error:


anamikau:
Thanks @visheshh. Do we need to add a new test or this can be covered by our existing performance/resilience tests?

visheshh:
@anamikau , We need to add a new test case .
To reproduce we need to submit job infinitely . Bug does not appear till first half an hour and we need to monitor atleast for 1-1.30 hr to verify fix is working . On a good machine it will submit more than half a million jobs by that time. So i felt automating wouldn't be a good idea . So i have tested it manually.

anamikau:
@visheshh I see it is not a functional but would be a performance/stress test which can only be run on large systems. Please add a manual test in the meantime.

suresh-thelkar:
@visheshh,
Code changes look good to me.  I think we also need to test the following scenario in order to complete our testing.

Just before qsub daemon credentials expires i.e. one minute before or at 29th minute we unlink qsub daemon's unix domain socket file in order to allow this daemon to still serve the request/s of existing clients for next 5 more seconds. During this time a second qsub daemon comes up which also uses the same unix domain socket file and also start serving the new qsub clients. At this point of time we really have two qsub daemons. 

During the above scenario we need to have a test case which tells or confirms us that the old qsub daemon serves its existing clients successfully before it actually exits which happens in the next 5 seconds. Although from theory it is expected that both daemons during this time should work it is always good to test this scenario to make sure it indeed works. Again I know this is very difficult to automate but we can at least have a manual test case for verifying the same.

suresh-thelkar:
@visheshh, Thanks for testing the scenario. Test results look good to me.

visheshh:
@PBSPro/pbspro-maintainers , Please review PR

prakashcv13:
@visheshh - could you please clarify if the issue was that **every** new qsub daemon exited every 1 minute or only the one getting created after CREDENTIAL_LIFETIME expiry?

src/cmds/qsub.c
```diff
@@ -5485,7 +5485,8 @@ do_daemon_stuff(void)
 
 out:
 	close(bindfd);
-	unlink(fl);
+	if (cred_timeout != 1)
```
arungrover: why don't you do what do_daemon_stuff does for windows i.e. just goto out label? I also find it interesting that the place we check that the credentials have expired, we unlink the file and then continue receiving data from the foreground process.

subhasisb: @arungrover yes this is part of the out label itself, the issue was that we were unlinking the socket file created by the next daemon that would have come up (so removing the file twice). 
Yes, we unlink the socket file first, so that no new qsub can connect to this daemon any more as it has decided to quit soon, but then continue to read and process data that was "already sent" to this socket by a already connected qsub front-end..

arungrover: Thanks, @subhasisb for the explanation. I am still a little confused about the implementation. The way we are exiting the background process is different between windows and linux. Maybe it is because in windows things work differently but I just want to be sure.

In case of windows, as soon as we see that the credentials are about to expire then we just exit from the process but in case of linux we process any pending request and then exit. To maintain consistency one of them should change, either in Windows, we should not cleanup so quickly, serve pending request and also provide 5 seconds of grace period for foreground process to exit OR in Linux we should just exit as soon as we see credentials are expiring and send DMN_REFUSE_EXIT or something similar to foreground process before exiting.

visheshh:
@prakashcv13 issue is that **every** new qsub daemon exits every 1 minute created after CREDENTIAL_LIFETIME expiry. I have updated the Bug description and title.

src/cmds/qsub.c
visheshh: @arungrover Please check https://github.com/PBSPro/pbspro/pull/613 for fix details.
The Bug fix was targeted differently for both the platforms. This part of code affects only Linux. 

Explaination from Praveen on the PR: 

Fix on Windows platform:

1. Problem is consistently reproducible on Windows platform.
2. I tried incorporating the similar fix that of Linux i.e. disconnecting the pipe and retaining it for
   seconds before closing it.
3. But, the behaviour of DisconnectNamedPipe API on windows is little different than that of unlink() call on linux. Unlinking the UD socket on linux still retains the existing connection object, but DisconnectNamedPipe doesn’t do that. It just allows the client to re-connect before doing the closehandle.
4. So, I just moved the code for checking the Credential timeout (30 mins) much before WaitForMultipleObjects call Just after while(1) loop and problem is resolved. I could see a different qsub process id coming up after 29 mins of cred time-out and no error message is being displayed.

arungrover: Thanks @visheshh I now understand why it was done this way. I feel we could have also fixed the problem if the process becoming the daemon would unlink the file itself before opening the socket. By doing that we could have the code same in both places.

prakashcv13:
looks good to me...

src/cmds/qsub.c
visheshh: @arungrover Do you think we can go ahead with the current fix ?

arungrover: It is your choice, you already have signoff on the change. It just felt inconsistent to me but then this change was not introduced by you in the first place.
#613  just does not seems like the right change to me.

subhasisb:
@arungrover the methods of communication between windows and unix domain sockets are totally different, and hence the difference. 
Also, we cannot leave it to the next daemon coming up to unlink the file before opening the socket, because - when the previous daemon was to go down, it needs to unlink the file first, clear up the backlog of data already in the pipe and then quit - if it left the file there, more connects will keep coming.

visheshh:
@prakashcv13 Please take this PR forward its rebased.

arungrover:
@subhasisb  If the first background serves the current client connection and closes it and after that, it checks for credential timeout and closes the listening socket as well then wouldn't other connecting clients get a connrefused at their end? I think even the client unlinks the socket file in case of connrefused.

subhasisb:
@arungrover that actually was the code before and does not work. By the time the background decides to quit, and decides to close and get out, another foreground had already connected and pumped data to it (recall in unix domain sockets, the client side connect will go ahead even before accept happened in the server side) - so, when we decide to go out, we first unlink the file, so that nobody new can connect, then we drain the pipe (so cannot close the socket first) process the data and then go out.

visheshh:
@arungrover While testing the fix we have added few logs after cred_timeout at dosend and dorecv we have made sure that foreground qsub is still connecting to the daemon even after the unlink has happened.

subhasisb:
@visheshh i think you got confused - i dont think that had happened - data can be passed even after the unlink has happened, not a connect..

arungrover:
ok, I agree with the implementation we currently have. It did intrigue me to go looking for an alternative. There is a possibility to use abstract namespace but it very Linux centric implementation.
In any case, I take my concern about the code back, it looks good to go.

visheshh:
@subhasisb Thanks Subhasis for clarifying got it .

prakashcv13:
merged #984

subhasisb:
> ok, I agree with the implementation we currently have. It did intrigue me to go looking for an alternative. There is a possibility to use abstract namespace but it very Linux centric implementation.
> In any case, I take my concern about the code back, it looks good to go.

Thanks @arungrover - yes abstract namespaces are linux specific feature, but not just that, the semantics of it auto closing without actually also enforcing a synchronous connect has a possibility to bring in more issues than solutions...perhaps!

PR #986, opened 2019-02-08
requirements decorator in PTL to provide cluster information required…

saritakh:
… for a test

#### Bug/feature Description
* *[PP-1281](https://pbspro.atlassian.net/browse/PP-1281): New decorator in PTL using which user can provide cluster information required for a test*

#### Affected Platform(s)
* *Linux*

 * **[requirements decorator design](https://pbspro.atlassian.net/wiki/spaces/PD/pages/458883073/PP-1281+New+decorator+in+PTL+using+which+user+can+provide+cluster+information+required+for+a+test.)**
* **[requirements decorator forum discussion](http://community.pbspro.org/t/pp-1281-new-decorator-in-ptl-using-which-user-can-provide-cluster-information-required-for-a-test/1050)**
* **

#### Solution Description
* *New "requirements" decorator added  in PTL to provide cluster information required for a test.*

#### Testing logs/output
[reqtest_194_TestRequirementsDecorator.txt](https://github.com/PBSPro/pbspro/files/2844537/reqtest_194_TestRequirementsDecorator.txt)
[ptl_ts_tree.json.txt](https://github.com/PBSPro/pbspro/files/2849901/ptl_ts_tree.json.txt)

#### Checklist:

### DIFF
--- a/test/fw/ptl/utils/plugins/ptl_test_info.py
+++ b/test/fw/ptl/utils/plugins/ptl_test_info.py
@@ -40,9 +40,25 @@ import logging
 from nose.plugins.base import Plugin
 from ptl.utils.pbs_testsuite import PBSTestSuite
 from ptl.utils.plugins.ptl_test_tags import TAGKEY
+from ptl.utils.pbs_testsuite import REQUIREMENTS_KEY
+from ptl.utils.pbs_testsuite import default_requirements
 from copy import deepcopy
 
 log = logging.getLogger('nose.plugins.PTLTestInfo')
+
+
+def get_effective_reqs(ts_reqs=None, tc_reqs=None):
+    """
+    get effective requirements at test case
+    """
+    tc_effective_reqs = {}
+    if (tc_reqs is None and ts_reqs is None):
+        tc_effective_reqs = deepcopy(default_requirements)
+    else:
+        tc_effective_reqs = deepcopy(default_requirements)
+        tc_effective_reqs.update(ts_reqs)
+        tc_effective_reqs.update(tc_reqs)
+    return tc_effective_reqs
 
 
 class FakeRunner(object):
@@ -203,6 +219,7 @@ class PTLTestInfo(Plugin):
         tsd['module'] = suite.__module__
         dcl = suite.__dict__
         tcs = {}
+        ts_req = getattr(suite, REQUIREMENTS_KEY, {})
         for k in dcl.keys():
             if k.startswith('test_'):
                 tcd = {}
@@ -213,8 +230,10 @@ class PTLTestInfo(Plugin):
                     # not a test case, ignore
                     continue
                 tcd['doc'] = str(tc.__doc__)
-                tctags = sorted(set(tstags + getattr(tc, TAGKEY, [])))
+                tc_req = getattr(tc, REQUIREMENTS_KEY, {})
+                tcd['requirements'] = get_effective_reqs(ts_req, tc_req)
                 numnodes = 1
+                tctags = sorted(set(tstags + getattr(tc, TAGKEY, [])))
                 for tag in tctags:
                     if 'numnodes' in tag:
                         numnodes = tag.split('=')[1].strip()
--- a/test/fw/ptl/utils/plugins/ptl_test_runner.py
+++ b/test/fw/ptl/utils/plugins/ptl_test_runner.py
@@ -42,6 +42,7 @@ import datetime
 import unittest
 import tempfile
 import platform
+import socket
 import pwd
 import signal
 import ptl
@@ -56,7 +57,9 @@ from nose.plugins.skip import SkipTest
 from nose.suite import ContextSuite
 from ptl.utils.pbs_testsuite import PBSTestSuite
 from ptl.utils.pbs_testsuite import TIMEOUT_KEY
+from ptl.utils.pbs_testsuite import REQUIREMENTS_KEY
 from ptl.utils.pbs_dshutils import DshUtils
+from ptl.utils.plugins.ptl_test_info import get_effective_reqs
 from ptl.lib.pbs_testlib import PBSInitServices
 from ptl.utils.pbs_covutils import LcovUtils
 try:
@@ -476,6 +479,7 @@ class PTLTestRunner(Plugin):
         self.__failed_tc_count = 0
         self.__tf_count = 0
         self.__failed_tc_count_msg = False
+        self._test_marker = 'test_'
 
     def options(self, parser, env):
         """
@@ -518,6 +522,7 @@ class PTLTestRunner(Plugin):
         """
         self.config = config
         self.enabled = True
+        self.param_dict = self.__get_param_dictionary()
 
     def prepareTestRunner(self, runner):
         """
@@ -577,6 +582,98 @@ class PTLTestRunner(Plugin):
         test.duration = test.end_time - test.start_time
         test.captured_logs = self.result.handler.get_logs()
 
+    def __get_param_dictionary(self):
+        """
+        Method to convert data in param into dictionary of cluster
+        information
+        """
+        tparam_contents = {}
+        nomomlist = []
+        shortname = (socket.gethostname()).split('.', 1)[0]
+        for key in ['servers', 'moms', 'comms', 'clients']:
+            tparam_contents[key] = []
+        if self.param is not None:
+            for h in self.param.split(','):
+                if '=' in h:
+                    k, v = h.split('=', 1)
+                    if (k == 'server' or k == 'servers'):
+                        tparam_contents['servers'].extend(v.split(':'))
+                    elif (k == 'mom' or k == 'moms'):
+                        tparam_contents['moms'].extend(v.split(':'))
+                    elif k == 'comms':
+                        tparam_contents['comms'] = v.split(':')
+                    elif k == 'client':
+                        tparam_contents['clients'] = v.split(':')
+                    elif k == 'nomom':
+                        nomomlist = v.split(':')
+        for pkey in ['servers', 'moms', 'comms', 'clients']:
+            if not tparam_contents[pkey]:
+                tparam_contents[pkey] = set([shortname])
+            else:
+                tparam_contents[pkey] = set(tparam_contents[pkey])
+        if nomomlist:
+            tparam_contents['moms'] -= set(nomomlist)
+        return tparam_contents
+
+    @staticmethod
+    def __are_requirements_matching(param_dic=None, test=None):
+        """
+        Validates test requirements against test cluster information
+        returns True on match or False otherwise None
+
+        :param param_dic: dictionary of cluster information from data passed
+                          to param list
+        :param_dic type: dic
+        :param test: test object
+        :test type: object
+
+        :returns True or False or None
+        """
+        ts_requirements = {}
+        tc_requirements = {}
+        param_count = {}
+        shortname = (socket.gethostname()).split('.', 1)[0]
+        if test is None:
+            return None
+        test_name = getattr(test.test, '_testMethodName', None)
+        if test_name is not None:
+            method = getattr(test.test, test_name, None)
+        if method is not None:
+            tc_requirements = getattr(method, REQUIREMENTS_KEY, {})
+            cls = method.im_class
+            ts_requirements = getattr(cls, REQUIREMENTS_KEY, {})
+        if not tc_requirements:
+            if not ts_requirements:
+                return None
+        eff_tc_req = get_effective_reqs(ts_requirements, tc_requirements)
+        for key in ['servers', 'moms', 'comms', 'clients']:
+            param_count['num_' + key] = len(param_dic[key])
+        for pk in param_count:
+            if param_count[pk] < eff_tc_req[pk]:
+                return False
+        if set(param_dic['moms']) & set(param_dic['servers']):
+            if eff_tc_req['no_mom_on_server']:
+                return False
+        else:
+            if not eff_tc_req['no_mom_on_server']:
+                return False
+        if set(param_dic['comms']) & set(param_dic['servers']):
+            if eff_tc_req['no_comm_on_server']:
+                return False
+        else:
+            if not eff_tc_req['no_comm_on_server']:
+                return False
+        comm_mom_list = set(param_dic['moms']) & set(param_dic['comms'])
+        if comm_mom_list and shortname in comm_mom_list:
+            # Excluding the server hostname for flag 'no_comm_on_mom'
+            comm_mom_list.remove(shortname)
+        if comm_mom_list:
+            if eff_tc_req['no_comm_on_mom']:
+                return False
+        else:
+            if not eff_tc_req['no_comm_on_mom']:
+                return False
+
     def startTest(self, test):
         """
         Start the test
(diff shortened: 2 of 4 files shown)


test/fw/ptl/utils/plugins/ptl_test_info.py
```diff
@@ -213,8 +234,12 @@ def _gen_ts_tree(self, suite):
                     # not a test case, ignore
                     continue
                 tcd['doc'] = str(tc.__doc__)
-                tctags = sorted(set(tstags + getattr(tc, TAGKEY, [])))
+                tc_req = getattr(tc, REQUIREMENTS_KEY, {})
+                eff_tc_req = get_eff_requirements(ts_req, tc_req)
+                if eff_tc_req:
+                    tcd['requirements'] = eff_tc_req
```
hirenvadalia: I don't think we need to put this under if condition, we should simply `requirements` key irrespective of whether the test has a requirement or not and by doing that we will get uniform JSON output.
If we add requirements key in JSON based on condition then some JSON output will have requirement key and some not which makes non-uniform output. What do you say?

test/fw/ptl/utils/plugins/ptl_test_info.py
```diff
@@ -40,11 +40,31 @@
...
 from ptl.utils.pbs_testsuite import PBSTestSuite
 from ptl.utils.plugins.ptl_test_tags import TAGKEY
+from ptl.utils.pbs_testsuite import REQUIREMENTS_KEY
+from ptl.utils.pbs_testsuite import default_requirements
 from copy import deepcopy
 
 log = logging.getLogger('nose.plugins.PTLTestInfo')
 
 
+def get_eff_requirements(ts_requirements=None, tc_requirements=None):
```
hirenvadalia: Optional comment: rename fn name to `get_effective_reqs` and rename `ts_requirements` to `ts_reqs` and `tc_requirements` to `tc_reqs`

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -577,6 +581,79 @@ def __set_test_end_data(self, test, err=None):
...
+        tparam_dic = {}
+        tparam_contents = {}
+        for key in paramkeys:
+            tparam_contents[key] = []
+        tparam_dic.update(default_requirements)
+        if self.param is None:
+            return tparam_dic
+        for h in self.param.split(','):
+            if '=' in h:
+                k, v = h.split('=')
```
hirenvadalia: `.split('=')` == `.split('=', 1)`

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -577,6 +581,79 @@ def __set_test_end_data(self, test, err=None):
...
+                if k in paramkeys:
+                    tparam_contents[k] = [v.replace(':', ',')]
+                    if (k == 'server' or k == 'servers'):
+                        tparam_dic['num_servers'] = len(v.split(':'))
```
hirenvadalia: `.split(':')` == `.split(':', 1)` (applies below also)

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -577,6 +581,79 @@ def __set_test_end_data(self, test, err=None):
...
+                    if (k == 'mom' or k == 'moms'):
+                        tparam_dic['num_moms'] = len(v.split(':'))
+                    if k == 'comms':
+                        tparam_dic['num_comms'] = len(v.split(':'))
+                    if k == 'clients':
+                        tparam_dic['num_clients'] = len(v.split(':'))
+        for pkey in paramkeys:
+            if pkey in ['server', 'mom', 'comms', 'client']:
+                if not tparam_contents[pkey]:
```
hirenvadalia: since you are calling multiple time `set` on `tparam_contents` below, I will suggest to put else part here and modify `tparam_contents` value by calling `set` so that you don't have to call `set` below

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -577,6 +581,79 @@ def __set_test_end_data(self, test, err=None):
...
+    @staticmethod
+    def __are_requirements_matching(param_count=None, test=None):
+        """
+        Validates test requirements against test cluster information
+        returns True on match or False otherwise
+        """
+        keylist = ['num_servers', 'num_moms', 'num_comms', 'num_clients']
+        flaglist = ['no_mom_on_server', 'no_comm_on_server', 'no_comm_on_mom']
+        if test is not None:
+            method = getattr(test.test, getattr(test.test, '_testMethodName'))
```
hirenvadalia: To be safer side, add default value in getattr and put if condition after this line

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -599,6 +676,11 @@ def startTest(self, test):
             self.__failed_tc_count_msg = True
             raise TCThresholdReached
         timeout = self.__get_timeout(test)
+        param_dict = self.__get_param_dictionary()
```
hirenvadalia: Do we need to populate this here? I think we can populate this in `configure` only since it doesn't depend on test which is about to start

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -599,6 +676,11 @@ def startTest(self, test):
...
+        rv = self.__are_requirements_matching(param_dict, test)
+        if rv is False:
+            self.result.startTest(test)
```
hirenvadalia: I suggest adding a comment here about why we are calling `self.result.startTest(test)` for future (and for other devs) reference

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -611,16 +612,17 @@ def __get_param_dictionary(self):
...
         if not ((tparam_contents['servers']) & (tparam_contents['moms'])):
             tparam_dic['no_mom_on_server'] = True
         if not ((tparam_contents['servers']) & (tparam_contents['comms'])):
             tparam_dic['no_comm_on_server'] = True
-        if ((tparam_contents['moms']) & (tparam_contents['comms'])):
-            tparam_dic['no_comm_on_mom'] = False
+        commtrack = ((tparam_contents['moms']) & (tparam_contents['comms']))
+        if commtrack:
+            if not ((len(commtrack) == 1) and (shortname in commtrack)):
+                tparam_dic['no_comm_on_mom'] = False
```
hirenvadalia: As we discussed, we don't need to calculate these flags value, we just have to use it in if else condition.

agrawalravi90:
A comment on the overall design of this decorator: PTL right now doesn't really support a multi-server configuration properly. The last time I had tested it out, it seems to have worked only if all the servers were passed as "nomom" hosts. So, you should think about how you want to handle multiple servers with options like "no_mom_on_server", which don't really say which server to operate on.

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -577,6 +582,87 @@ def __set_test_end_data(self, test, err=None):
...
+                    if not ts_requirements:
+                        return None
+            eff_tc_req = get_effective_reqs(ts_requirements, tc_requirements)
+        for key in ['servers', 'moms', 'comms', 'clients']:
+            param_count['num_'+key] = len(param_dict[key])
+        for pk in param_count:
+            if pk in keylist:
+                if param_count[pk] < eff_tc_req[pk]:
+                    return False
+        if set(param_dict['moms']) & set(param_dict['servers']):
```
agrawalravi90: So, looking at this check, it seems like the function will return False for a multi-server setup if ANY of the servers have a mom running on them, and no_mom_on_server is set. If you want to retain this behavior, then please make sure to document it.

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -577,6 +582,87 @@ def __set_test_end_data(self, test, err=None):
...
+            if eff_tc_req['no_mom_on_server']:
+                return False
+        else:
+            if not eff_tc_req['no_mom_on_server']:
```
agrawalravi90: here as well, when no_mom_on_server is False, if there are no moms on any servers in a multi-server scenario, it will fail. I think we can just document that these flags will apply to all servers in a multi-server setup

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -577,6 +582,87 @@ def __set_test_end_data(self, test, err=None):
...
+                return False
+        if set(param_dict['comms']) & set(param_dict['servers']):
+            if eff_tc_req['no_comm_on_server']:
+                return False
+        else:
+            if not eff_tc_req['no_comm_on_server']:
+                return False
+        if set(param_dict['moms']) & set(param_dict['comms']):
+            if not eff_tc_req['no_comm_on_mom']:
```
agrawalravi90: shouldn't this be "if eff_tc_req['no_comm_on_mom']" instead of "if not eff_tc_req['no_comm_on_mom']" ? Right now, it says that if there are any comms on mom hosts and no_comm_on_mom is False, then the function will return False, which seems wrong.

test/tests/selftest/pbs_requirements_decorator.py
```diff
@@ -0,0 +1,84 @@
+# coding: utf-8
+
+# Copyright (C) 1994-2018 Altair Engineering, Inc.
```
agrawalravi90: 2019

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -599,6 +685,13 @@ def startTest(self, test):
             self.__failed_tc_count_msg = True
             raise TCThresholdReached
         timeout = self.__get_timeout(test)
+        rv = self.__are_requirements_matching(self.param_dict, test)
```
agrawalravi90: The decorator isn't really doing much besides setting REQUIREMENTS_KEY, the actual checking for requirements is happening here as part of startTest(). Is there no way to move this check to the decorator itself? The whole point of a decorator is to run it before starting a test.

test/tests/selftest/pbs_requirements_decorator.py
```diff
@@ -0,0 +1,84 @@
...
+class TestRequirementsDecorator(TestSelf):
+
+    """
+    This test suite is to verify functionality of requirements
+    decorator. This test suite when run on single node without any
+    hostnames passed in -p or param file, tests run and skip functionality
+    of this decorator
+
+    """
+    @requirements(num_servers=1, num_comms=1)
```
agrawalravi90: I feel like these tests aren't really helpful. One will just see "skipped" in the output instead of "failed" or "passed", and it'll be difficult to know whether "skipped" means that the test passed as expected, or if it failed.

Not sure if this approach will work for you as the decorator doesn't seem to be doing the actual check itself (you seem to be doing the actual check inside startTest()), but do check this out and see if you can write tests similarly:
https://stackoverflow.com/questions/34648337/how-to-test-a-decorator-in-a-python-package

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -577,6 +582,87 @@ def __set_test_end_data(self, test, err=None):
         test.duration = test.end_time - test.start_time
         test.captured_logs = self.result.handler.get_logs()
 
+    def __get_param_dictionary(self):
+        """
+        Method to convert data in param into dictionary of counts
+        """
+        paramkeys = ['server', 'servers', 'mom', 'moms', 'comms', 'client',
```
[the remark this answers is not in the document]
saritakh: "comm" is not an option since currently PTL has only below param input types for PBS daemon hostnames and there is no "comm" specification here, but only "comms" specification being present:

    :param server: The hostname on which the PBS ``server/scheduler``
                   are running
    :param mom: The hostname on which the PBS MoM is running
    :param servers: Colon-separated list of hostnames hosting a PBS server.
                    Servers are then accessible as a dictionary in the
                    instance variable servers.
    :param client: For CLI mode only, name of the host on which the PBS
                   client commands are to be run from. Format is
                   ``<host>@<path-to-config-file>``
    :param moms: Colon-separated list of hostnames hosting a PBS MoM.
                 MoMs are made accessible as a dictionary in the instance
                 variable moms.
    :param comms: Colon-separated list of hostnames hosting a PBS Comm.
                  Comms are made accessible as a dictionary in the
                  instance variable comms.
    :param nomom=<host1>\:<host2>...: expect no MoM on given set of hosts

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -620,46 +618,61 @@ def __get_param_dictionary(self):
...
-        if set(param_dict['comms']) & set(param_dict['servers']):
+        if set(param_dic['comms']) & set(param_dic['servers']):
             if eff_tc_req['no_comm_on_server']:
                 return False
         else:
             if not eff_tc_req['no_comm_on_server']:
                 return False
-        if set(param_dict['moms']) & set(param_dict['comms']):
+        comm_mom_list = set(param_dic['moms']) & set(param_dic['comms'])
+        if shortname in comm_mom_list:
```
agrawalravi90: if comm_mom_list is None then this will error out, so I think you need to change this to:  "if comm_mom_list and if shortname in comm_mom_list":

saritakh:
@agrawalravi90 , yes, I have some pending comments to be addressed; I am working on the possibilities. Sorry, I missed to update comment here with my last commit.

test/fw/ptl/utils/plugins/ptl_test_runner.py
saritakh: This decorator is supported at test suite level and test case level. I found it difficult to get both test suite level and test case level requirements data together in the same call inside the decorator function, hence could not compute the effective requirements for a given test case context.

Regarding the other data which is param data of benchpress, in order to get that inside decorator it could be got by a function defined (in pbs_testsuite.py) to set a PBSTestSuite class level macro to the merged param data (-p and paramfile); and call this function from pbs_benchpress after importing. 

So, I preferred to have this current code flow where the merged param data is already available in runner plugin and the requirements data that is set on test object could be captured at different levels i.e. from test method (current test name) and its class which is the test suite. (code lines 640 - 646 in ptl_test_runner.py)

test/tests/selftest/pbs_requirements_decorator.py
saritakh: With the current approach mentioned in my previous note about check not being done in requirements decorator; the most I could do is verify the requirements at test case level by extracting the REQUIREMENTS_KEY and verify that it is set. I have modified test_tc_run with this check. The execution of test_tc_run indicates test case level requirements are effective over the test suite level requirements and rest of the tests getting skipped indicate working skip functionality. 
I have kept the rest part of tests as is. Please let me know your suggestions.

test/fw/ptl/utils/plugins/ptl_test_runner.py
saritakh: I have updated the design documents for the multi-server setup.

saritakh:
@agrawalravi90 I have updated most of your review comments and reasoned my approach for the others. Please review the same and let me know of any comments or suggestions.

agrawalravi90:
Thanks for the explanations Sarita. I think ideally it would have been great to have the checks be done as part of the decorator itself, but it's not trivial (if at all possible) to combine a class decorator and method decorator which does all the necessary checks, so let's go ahead with this and we can come back to it if any issues come up.

saritakh:
Thanks @agrawalravi90 for the approval. I agree with your decision.

saritakh:
Thanks @hirenvadalia for the approval.

saritakh:
@agrawalravi90 I have squashed and rebased by branch. Request you to take it forward. Thanks.

agrawalravi90:
@saritakh I'm afraid your branch is stale again, can you please update it once more?

PR #990, opened 2019-02-14
fix for chk_file_sec() to verify the file security

bremanandjk:
#### Bug/feature Description
Function chk_file_sec() on windows does not work properly, as a result, there are two problems one can face one windows -
1. There is a binary called "chk_tree" which is available on windows. This binary is used to check if the files passed to it as an argument have security issues and have open write permissions for non-admin users. Due to the bug, this binary never reports any security issues

2. Mom/Server dynamic resource scripts also suffer from the same bug on windows. PBS checks that the dynamic resource scripts do not have any security issues but, since the check on windows is broken, it never finds out any security issues and executes the scripts.

#### Affected Platform(s)
* Windows

* After populating ACL entities, we are verifying each ACE elements in a loop, the logical expressions used for file permissions check and ownership verifications are not cautiously written. 

#### Solution Description
*  Correct the evaluation expressions in file permission checks.

#### Testing logs/output
* [Test_logs_windows.txt](https://github.com/PBSPro/pbspro/files/2865763/Test_logs_windows.txt)
* [chk_tree_file_test_results.docx](https://github.com/PBSPro/pbspro/files/2865847/chk_tree_file_test_results.docx)

#### Checklist:

- [] I have added documentation in the **[project documentation area](https://pbspro.atlassian.net/wiki/display/PD)**.

### DIFF
--- a/src/lib/Libwin/accesinfo.c
+++ b/src/lib/Libwin/accesinfo.c
@@ -811,7 +811,7 @@ perm_granted_admin_and_owner(char *path, int disallow, char *owner, char *errmsg
 		powner_s = getgrpname(powner);
 
 	if (powner_s != NULL) {
-		if ( !sidIsAdminPrivilege(powner) && \
+		if ( !sidIsAdminPrivilege(powner) || \
 			( (owner != NULL) && strcmp(owner, powner_s) != 0 )  ) {
 			rc = EPERM;
 			sprintf(errmsg, "File %s not owned by user %s or an admin-type user!",
@@ -852,7 +852,7 @@ perm_granted_admin_and_owner(char *path, int disallow, char *owner, char *errmsg
 			 ( EqualSid((SID *)&pace->SidStart, esid) || \
 			sid2rid((SID *)&pace->SidStart) != \
 					 SECURITY_CREATOR_OWNER_RID ) && \
-			(!sidIsAdminPrivilege( (SID *)&pace->SidStart)) && \
+			(!sidIsAdminPrivilege( (SID *)&pace->SidStart)) || \
    			((owner != NULL) && strcmp(name, owner) != 0) ) {
 			(void)accessinfo_add(allowed, sizeInfo.AceCount,
 				name, (mask & 0xFFFF));


src/lib/Libwin/accesinfo.c
```diff
@@ -852,7 +852,7 @@ perm_granted_admin_and_owner(char *path, int disallow, char *owner, char *errmsg
 			 ( EqualSid((SID *)&pace->SidStart, esid) || \
 			sid2rid((SID *)&pace->SidStart) != \
 					 SECURITY_CREATOR_OWNER_RID ) && \
-			(!sidIsAdminPrivilege( (SID *)&pace->SidStart)) && \
+			(!sidIsAdminPrivilege( (SID *)&pace->SidStart)) || \
```
bayucan: I believe in here, we add to the "allowed" list if file's owner has admin privilege, or if not admin-privilege, and yet there's a passed 'owner' to the function (i.e. owner != NULL), that owner must be the same as the file's owner (i.e. name). So I think the if condition here should be:
 && (sidIsAdminPrivilege( (SID *)&pace->SidStart) || ((owner != NULL) && (strcmp(name, owner) == 0))
*** kind of like the opposite of the one line 814 that we're saying is invalid.

bremanandjk: Thanks for your review @bayucan 

I am reading the code this way, 
- In the first check after entering this api at https://github.com/PBSPro/pbspro/blob/master/src/lib/Libwin/accesinfo.c#L814, we are validating whether the file is owned by "Admin" user or passed "owner", else throwing EPERM error(1).
-  If the control goes further means, file is owned by Admin/passed owner. then we are getting ACL and ACE's. 
- At second check, in loop, we are parsing each ACE and filtering only the allowed list of file permissions. here. we are interested only in ***Non***-Admin's, ***Non***-passed-owner's allowed file attributes. So that we check with passed mask value for NON-admins/NON-passed-owners permissions. 
Eg: if we dont want to allow NON-admin write permissions, we are filtering only non-admin users allowed permissions.
- Later we are filtering "Disallowed" permissions and comparing all data with disallow mask flag and return EACCES(13) error.

I have attached test results for below four cases in attached doxs with all possibilites, its working as expected without proposed changes.
1) Admin only owned file without sharing write permission to Non-admin users.
2) Admin owned, but sharing write permission with non-admin user
3) Non-admin owned file
4) Non-Admin owned file with sharing write permission with Non-admin users

FYI: I have tried building with suggested way, It seems not working in perfect case.(ie: Admin owned , Sharing no write permissions with NON-admin user.)

Kindly let me know if understaning need some change. Thanks

bayucan: Your explanation makes sense! Thanks @bremanandjk

PR #1003, opened 2019-02-25
pbsnodes_output: Add skip code to skip the test if not run by root

VishrutArora:
#### Bug/feature Description
* Add skip code to skip the test if it is being run by a non-root user. The test needs to be run a root user in order to succesfully edit the /etc/hosts file.

#### Affected Platform(s)
* *All*

* The skip code to handle the condition where the test is being run by a non-root user was absent

#### Solution Description
* Skip coded added

#### Testing logs/output
[output_trimmed_logs.txt](https://github.com/PBSPro/pbspro/files/2952986/output_trimmed_logs.txt)

#### Checklist:

### DIFF
--- a/test/tests/functional/pbs_pbsnodes_output_trimmed.py
+++ b/test/tests/functional/pbs_pbsnodes_output_trimmed.py
@@ -45,46 +45,21 @@ class TestPbsnodesOutputTrimmed(TestFunctional):
     This test suite tests pbsnodes executable with -l
     and makes sure that the node names are not trimmed.
     """
-    temp_list = []
-    is_host_file_changed = False
-
-    @classmethod
-    def hostname_resolves(self, hostname):
-        try:
-            socket.gethostbyname(hostname)
-            return 1
-        except socket.error:
-            return 0
 
     def test_pbsnodes_output(self):
         """
-        This method adds a new entry to /etc/hosts and creates
-        a mom with name more than 20 characters, then makes the
-        node offline and checks the pbsnodes -l output and
+        This method creates a new vnode with name more than
+        20 characters, and checks the pbsnodes output and
         makes sure it is not trimmed
         """
-        with open("/etc/hosts", "r")as fd:
-            self.temp_list = fd.readlines()
+        self.server.manager(MGR_CMD_DELETE, NODE, None, '')
         pbsnodes = os.path.join(self.server.pbs_conf['PBS_EXEC'],
                                 "bin", "pbsnodes")
         hname = "long123456789012345678901234567890.pbspro.com"
-        ret = self.hostname_resolves(hname)
-        if ret == 0:
-            with open("/etc/hosts", "a") as fd:
-                tmpentry = '127.0.0.1' + " " + hname + " " + hname
-                fd.write(tmpentry)
-                fd.write("\n")
-            self.is_host_file_changed = True
-        self.server.manager(MGR_CMD_DELETE, NODE, None, '')
-        self.server.manager(MGR_CMD_CREATE, NODE,
-                            {'resources_available.ncpus': '1'}, hname)
-        command = pbsnodes + " -l"
+        a = {'resources_available.ncpus': 4}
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_pbsnodes_output_trimmed.py
```diff
@@ -70,6 +70,8 @@ def test_pbsnodes_output(self):
         hname = "long123456789012345678901234567890.pbspro.com"
         ret = self.hostname_resolves(hname)
         if ret == 0:
+            if os.getuid() != 0:
```
agrawalravi90: shouldn't this be the first thing that you do in this test? why is this check so far down?

anamikau: @agrawalravi90 the test will pass if hostname is resolvable and will try to edit /etc/hosts file in case it is not hence we do not want to skip at first place.

agrawalravi90: Thanks for your reply Anamika. The possibility of there being an actual host with that ridiculously long and made up hostname ("long123456789012345678901234567890.pbspro.com") is negligible. So, I think we should make the test less confusing by just checking for root permission at the top and just editing /etc/hosts file without checking for the hostname to be resolvable. Even if there are multiple entries in /etc/hosts (just in case that hostname actually exists), the test should work fine, the resolvers will just pick the first entry in /etc/hosts for that hostname

VishrutArora: @agrawalravi90 What you are saying makes sense, except that for the internal testbed systems this absurd hostname has already been added for testing purposes. The reason for adding was to avoid editing an important file such as  /etc/hosts file. Also, if we write the skip code at the beginning of the test like you suggested, we wouldn't be able to run it as pbsroot user.

agrawalravi90: wow, ok. If this test exists to just check that pbsnodes output doesn't trim a really long vnode name, then how about instead of editing the /etc/hosts file, we edit the vnode def file to create a vnode with that ridiculous name with the existing mom? that won't require this test to run as root at all.

agrawalravi90: and we can rename that poor host in our testbed systems

anamikau: @agrawalravi90 that is a good suggestion. @VishrutArora please try it out.

bhroam: If all you want is a very long vnode name, then you can create one.  You say qmgr -c 'c n <very long name> mom=hostname'.  The vnode will immediately go stale, but it will exist for the purposes of using pbsnodes.

VishrutArora:
@agrawalravi90 @bhroam  I've addressed the comments given by you. Kindly Review

test/tests/functional/pbs_pbsnodes_output_trimmed.py
```diff
@@ -1,6 +1,6 @@
 # coding: utf-8
 
-# Copyright (C) 1994-2019 Altair Engineering, Inc.
+# Copyright (C) 1994-2018 Altair Engineering, Inc.
```
agrawalravi90: 2019

VishrutArora: @agrawalravi90 I've pushed the changes for the same. Kindly review.

agrawalravi90:
Looks good, please post test logs, thanks.

VishrutArora:
I've attached the testing logs as well. Thanks

prakashcv13:
AppVeyor passed for debug, but failed for release - have restarted the build.

VishrutArora:
> AppVeyor passed for debug, but failed for release - have restarted the build.

Thanks Prakash. I'll squash and rebase for merge.

PR #1017, opened 2019-03-06
Fix buffer overrun on long TMPDIR path

riyazhakki:
#### Bug/feature Description
* *If the user define the TMPDIR path longer than ~128 characters then job submission fails with error 'qsub: cannot access script file'*
* ***Steps to reproduce:***
 1.  *export TMPDIR=/home/mohammh/aaaaaaaaaa/bbbbbbbbbb/cccccccccc/eeeeeeeeee/ffffffffff/gggggggggg/hhhhhhhhhh/iiiiiiiiii/jj/afdj/hlppoo/jkloiytupoo/bhtiusabsdlg/*
  2.  *mkdir -p $TMPDIR*
  3.  *qsub job.scr*
 *qsub: cannot access script file*

#### Affected Platform(s)
* *All Linux*

* *Function get_comm_filename appends the tmpdir path to the path of unixsocket filename, which is restricted to 108 charecters and thus when the user update the TMPDIR path longer than 108 characters, buffer overflow occurs and that leads to invalid path.*

#### Solution Description
* *Using snprintf to avoid the bufferflow and if the tmpdir path is greater than buffer size we use /var/tmp path for the unixsocket file.*

#### Testing logs/output
[test_qsub_with_script_with_long_TMPDIR_BeforeFix.ptl.txt](https://github.com/PBSPro/pbspro/files/2932817/test_qsub_with_script_with_long_TMPDIR_BeforeFix.ptl.txt)
[test_qsub_with_script_with_long_TMPDIR_AfterFix.ptl.txt](https://github.com/PBSPro/pbspro/files/2932816/test_qsub_with_script_with_long_TMPDIR_AfterFix.ptl.txt)
[TestQsubOptionsArguments.ptl.txt](https://github.com/PBSPro/pbspro/files/2932848/TestQsubOptionsArguments.ptl.txt)

#### Checklist:

### DIFF
--- a/src/cmds/qsub.c
+++ b/src/cmds/qsub.c
@@ -227,7 +227,8 @@ static char script_tmp[MAXPATHLEN + 1] = {'\0'}; /* name of script file copy */
 #ifdef WIN32
 static char fl[(2 * MAXPATHLEN) + 1] = {'\0'}; /* the filename used as the pipe name */
 #else
-static char fl[sizeof(((struct sockaddr_un *)0)->sun_path)] = {'\0'}; /* the filename used as the pipe name */
+#define MAXPIPENAME sizeof(((struct sockaddr_un *)0)->sun_path)
+static char fl[MAXPIPENAME] = {'\0'}; /* the filename used as the pipe name */
 #endif
 #define BUFSIZE 1024 /* windows default pipe buffer size */
 #define PIPE_TIMEOUT 0 /* default windows pipe timeout */
@@ -5241,8 +5242,10 @@ get_comm_filename(char *fname)
 {
 	char *env_svr = getenv(PBS_CONF_SERVER_NAME);
 	char *env_port = getenv(PBS_CONF_BATCH_SERVICE_PORT);
+	int count = 0;
 
-	sprintf(fname, "%s/pbs_%.16s_%lu_%.8s_%.32s_%.16s_%.5s",
+
+	count = snprintf(fname, MAXPIPENAME, "%s/pbs_%.16s_%lu_%.8s_%.32s_%.16s_%.5s",
 		tmpdir,
 		((server_out == NULL || server_out[0] == 0) ?
 		"default" : server_out),
@@ -5252,6 +5255,19 @@ get_comm_filename(char *fname)
 		(env_svr == NULL)?"":env_svr,
 		(env_port == NULL)?"":env_port
 		);
+
+	if (count >= MAXPIPENAME) {
+		snprintf(fname, MAXPIPENAME, "%s/pbs_%.16s_%lu_%.8s_%.32s_%.16s_%.5s",
+		TMP_DIR,
+		((server_out == NULL || server_out[0] == 0) ?
+		"default" : server_out),
+		(unsigned long int)getuid(),
+		cred_name,
+		get_conf_path(),
+		(env_svr == NULL)?"":env_svr,
+		(env_port == NULL)?"":env_port
+		);
+	}
 }
 
 /**
--- a/test/tests/functional/pbs_qsub_opts_args.py
+++ b/test/tests/functional/pbs_qsub_opts_args.py
@@ -36,6 +36,7 @@
 # trademark licensing policies.
 
 from tests.functional import *
+import os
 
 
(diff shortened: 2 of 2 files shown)


prakashcv13:
@riyazhakki - tmpdir is being used in get_script() as well.
Also, the solution that you are proposing would silently override the TMPDIR setting in qsub.c, which the user might not know or expect, which I do not think is what we would want.
How about using dynamically allocated memory for fname (get_comm_filename) and tmp_name (get_script). If memory allocation fails, we display "out of memory" and exit.

mike0042:
Dynamically allocated memory won't make a difference in this case. The limitation is on the length of the named pipe which is restricted to 108 characters in /usr/include/linux/un.h. The full path must fit in the sockaddr_un.sun_path array. We would only use the alternate location in the unlikely event that the user has defined a very long TMPDIR. The only other solution I can think of would be to create a (much shorter) symbolic link to the named pipe and use that instead, but I prefer the solution being proposed.

riyazhakki:
Thank you Mike for providing clarification. Prakash, about tmpdir in get_script () we are not changing the tmpdir for the job script.

prakashcv13:
@mike0042 - I did not realize we are using this path for the Unix domain socket. Thank you for pointing out.
@riyazhakki - in get_script() we do use tmpdir which may have the value from the environment variable. As far as I know, env variables can be more than 1024 (MAXPATHLEN) characters. If the user sets the TMPDIR to a path more than 1024 characters which is less than PATH_MAX (4096), mkstemp() will fail.

here is an example run.
[qsub buffer overrun.txt](https://github.com/PBSPro/pbspro/files/2940639/qsub.buffer.overrun.txt)

This may not cause a buffer overrun, as we are already using snprintf().
We can either use the same logic as you have in get_comm_filename().

riyazhakki:
Prakash, this environment variable we are using for the file path and I think most of the filesysetems limit it to the 255 bytes, even in case user provide the long path we are doing right throwing the message "qsub: could not create/open tmp file".

prakashcv13:
filenames are limited to 255 characters. Most of the modern FSs do not have a limit on the path length. [This](https://en.wikipedia.org/wiki/Comparison_of_file_systems) is a comparison of the fFSs.
Coming to doing the right thing - the user is in no way aware that qsub will be copying the script to a temporary location, so this error will not be something expected. Copying the file is internal to qsub.

riyazhakki:
I think it is long odds the user will provide the path greater than MAXPATHLEN as compare to MAXPIPNAME (for unix pipe).
I would like to go with existing code until the reviewer strongly feels we need the similar change here.

subhasisb:
@prakashcv13 for the temporary script file, we do not want to override if user has set anything in TMPDIR. That is because the user wants us to use TMPDIR for all temporary stuff..script files might e large etc. However, The the domain socket filename is different, it's internal and 0 bytes in length, so we chose to try /tmo before we give up. I don't think we should do that for scripts - it current thows an error which can be enhanced, but that is probably a separate ticket

prakashcv13:
@subhasisb - copying the script too is internal to qsub, which makes the error message unexpected for the user.
However, I do see a point in not overriding TMPDIR for copying scripts because it beats the reasons for which the TMPDIR was set to something else in the first place. How we go about handling it, can be done as a separate ticket as you suggested.

subhasisb:
> @subhasisb - copying the script too is internal to qsub, which makes the error message unexpected for the user.
> However, I do see a point in not overriding TMPDIR for copying scripts because it beats the reasons for which the TMPDIR was set to something else in the first place. How we go about handling it, can be done as a separate ticket as you suggested.

@prakashcv13 , You are right - both of these things are actually internal. However, the socket filename is zero byte file whereas the temporary script might take actual space. The socket filename length is limited by our choice of the communication infrastructure - which is the unix domain socket. On the other hand, the temporary file path is just that a path with a maximum allowed length. In that sense, the socket filename path is more "internal" implementation dependant that the temporary script path...and thus this attempt to work around the issue.

prakashcv13:
@subhasisb - alright, let us go ahead with this workaround.

riyazhakki:
Thank you Subhasis and Prakash for the review and approval.

PR #1022, opened 2019-03-06
test_revert_pbsconf_corelimit fails intermittently

agrawalravi90:
#### Bug/feature Description
* The test seems to fail intermittently as follows:
2018-12-14 12:01:36,883 INFO     running init script to start pbs on pbi-01-r6.pbspro.com using /etc/pbs.conf init_cmd=['sudo', '/opt/pbs/libexec/pbs_init.d', 'start']
2018-12-14 12:01:36,884 INFOCLI2 pbi-01-r6: sudo /opt/pbs/libexec/pbs_init.d start
2018-12-14 12:01:40,758 INFOCLI2 pbi-01-r6: sudo -H /opt/tools/wrappers/cat /var/spool/pbs/mom_priv/mom.loc
2018-12-14 12:01:40,793 INFO     mom pbi-01-r6: sent signal -SEGV
2018-12-14 12:01:40,794 INFOCLI2 pbi-01-r6: sudo -H kill -SEGV 15602
2018-12-14 12:01:40,859 INFOCLI2 pbi-01-r6: sudo -H /opt/tools/wrappers/cat /var/spool/pbs/mom_priv/mom.lock
2018-12-14 12:01:40,894 INFO     FAILED
Traceback (most recent call last): File "/usr/pbsroot13590/tmp/ptl_src/tests/selftest/pbs_pbstestsuite.py", line 168, in test_revert_pbsconf_corelimit self.assertFalse(self.mom.isUp()) AssertionError: True is not false

#### Affected Platform(s)
* Linux

* There's a race condition in the test w.r.t checking whether the mom is up or not. The test just passes SEGV and immediately does an assert on mom.isUp() to see whether the mom is down or not. Sometimes the mom might be up when SEGV is sent, so the test fails.

#### Solution Description
* mom.isUp() was recently enhanced to wait 20 seconds, checking every second to see if the mom was up yet. This was done to solve similar race condition issues. So, I enhanced the function to take an 'expectation' argument. If set to false, the function checks for the mom to be down instead of up. This argument is set to True by default.

#### Testing logs/output
* 
[test_revert_pbsconf_corelimit.log](https://github.com/PBSPro/pbspro/files/2936148/test_revert_pbsconf_corelimit.log)

#### Checklist:

 - [X] I have attached **test logs to this pull request** as evidence of testing/verification

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -12718,12 +12718,12 @@ class MoM(PBSService):
         self.version = None
         self._is_cpuset_mom = None
 
-    def isUp(self):
+    def isUp(self, max_attempts=20):
         """
         Check for PBS mom up
         """
         # Poll for few seconds to see if mom is up and node is free
-        for _ in range(20):
+        for _ in range(max_attempts):
             rv = super(MoM, self)._isUp(self)
             if rv:
                 break
--- a/test/tests/selftest/pbs_pbstestsuite.py
+++ b/test/tests/selftest/pbs_pbstestsuite.py
@@ -165,7 +165,12 @@ class TestPBSTestSuite(TestSelf):
         # Send SIGSEGV to pbs_mom
         self.assertTrue(self.mom.isUp())
         self.mom.signal("-SEGV")
-        self.assertFalse(self.mom.isUp())
+        for _ in range(20):
+            ret = self.mom.isUp(max_attempts=1)
+            if not ret:
+                break
+            time.sleep(1)
+        self.assertFalse(ret, "Mom was expected to go down but it didn't")
 
         # Confirm that no core file was generated
         mom_priv_filenames = self.du.listdir(self.server.hostname,
@@ -188,7 +193,12 @@ class TestPBSTestSuite(TestSelf):
         # Send another SIGSEGV to pbs_mom
         self.assertTrue(self.mom.isUp())
         self.mom.signal("-SEGV")
-        self.assertFalse(self.mom.isUp())
+        for _ in range(20):
+            ret = self.mom.isUp(max_attempts=1)
+            if not ret:
+                break
+            time.sleep(1)
+        self.assertFalse(ret, "Mom was expected to go down but it didn't")
 
         # Confirm that a core file was generated this time
         mom_priv_filenames = self.du.listdir(self.server.hostname,


agrawalravi90:
@latha-subramanian can you please review this? Thanks.

latha-subramanian:
@agrawalravi90, the changes related to adding 'expectation' argument  looks good to me. But why is the code for 'node state' check removed now ?  I feel that the check is still required for cases when 'expectation' is True.

agrawalravi90:
Thanks for your review Latha. I removed it because I felt that the mom.isUp() interface shouldn't automatically be checking that the nodes associated to that mom are 'free'/'down'. It also gets complicated if there are multiple nodes associated to that mom: what if only one node is in the correct state and others aren't? do we return True or False? So, I felt that it's better that the test writers test that explicitly outside of this function, if needed. What do you think?

bhroam:
I'm not sure I agree with your change.  First, I find it counterintuitive to call a function called "isUp()" to check if the mom is down.  Why can't the user handle it properly on the outside?  If they expect isUp() to fail, they can put it in an assertFalse().  If you are trying to avoid the 20s wait, then add a max_attempts parameter or something like that.

I personally like the check for state=free.  While it doesn't check all vnodes on the host, it does check the natural vnode.  This should be sufficient for the check.  You're really just checking to see if the mom came back and has reported in to the server.  If you want to be a little more exact, then checking for (MATCH_RE, 'down'), op=NE is better.  There's nothing saying the node wasn't offline before the mom went down.

agrawalravi90:
> Why can't the user handle it properly on the outside? If they expect isUp() to fail, they can put it in an assertFalse()

That's exactly why the test was failing: it signals a SEGV to mom and immediately does assertFalse on isUp, it's a race condition, which i think is very similar to the race condition that isUp() currently solves for when it expects the mom to be up by waiting doing 20 attempts to check if it is up. That's why I thought of solving this race condition in the same way.

> First, I find it counterintuitive to call a function called "isUp()" to check if the mom is down

I see what you mean, but isn't it the same as log_match() being called with existence=False?

> I personally like the check for state=free

Ok, I see your point, but currently the function just returns a False if the mom is up but the natural vnode is down, I feel that that's not the right thing to do and can be misleading: does False mean the mom is down? or does it mean that the natural vnode is in state 'down' ? If this is an obvious error then shouldn't we be raising an exception?

agrawalravi90:
>I think just adding expectation may not help in all cases. Consider a case where MoM is trying to come up and a test writer calls isUp(False). It might just return false positive from the very first poll result and the test will think that MoM is not up.
I think it is better to add a wait option to wait for a certain period of time to be sure whether the mom is up or not.

I don't like the idea of a wait because users will have to make it large to accommodate slower machines, and it will increase the test run time on faster machines. Also, they can just use time.sleep() in the test.

I'm more and more thinking that this function should just not do anything special, let the user deal with race conditions in the tests, add a sleep(), iterate 20 times, check for node state to be down etc., whatever is relevant for their test. What do you think?

latha-subramanian:
> > Why can't the user handle it properly on the outside? If they expect isUp() to fail, they can put it in an assertFalse()
> 
> That's exactly why the test was failing: it signals a SEGV to mom and immediately does assertFalse on isUp, it's a race condition, which i think is very similar to the race condition that isUp() currently solves for when it expects the mom to be up by doing 20 attempts to check if it is up. That's why I thought of solving this race condition in the same way.
> 
> > First, I find it counterintuitive to call a function called "isUp()" to check if the mom is down
> 
> I see what you mean, but isn't it the same as log_match() being called with existence=False?
> 
> > I personally like the check for state=free
> 
> Ok, I see your point, but currently the function just returns a False if the mom is up but the natural vnode is down, I feel that that's not the right thing to do and can be misleading: does False mean the mom is down? or does it mean that the natural vnode is in state 'down' ? If this is an obvious error then shouldn't we be raising an exception?
@agrawalravi90 , In my opinion, it is ok to return a False when the mom daemon is up, but the natural node is down, because we cannot do any useful operation when the node is down. This check was indeed added recently to address test failures due to such a scenario.

bhroam:
@agrawalravi90 log_match() is a much more complex function.  I don't really mind the extra complexity of existence=False.  I find isUp() a simple function, I find it more confusing.  Maybe it is just personal taste.

As for checking if the node is up, I look at this function as doing two things.  It first checks if the mom process is running, and  it also checks if the mom has successfully contacted the server.  I see the real purpose of isUp() is to tell you if the mom is up and ready for use.  If you just check if the process is running, you're adding yet another race condition to the mix.  Having the caller fix this means that every caller of isUp() has to fix it.  While the existing code isn't quite right (it should be checking for not down), I still think it is the right answer.  As a note, the server's isUp() does the same thing.  It waits until the server is up and is responsive.

agrawalravi90:
@bhroam and @latha-subramanian thanks for clarifying things.

Since it seemed confusing to have isUp check for mom to be down as well, I added a function isDown() instead which does that. Since that completely changes the approach, I decided to just revert my changes and create a fresh commit. Please let me know what you think about this approach.

agrawalravi90:
@bhroam @latha-subramanian  and @arungrover,  a gentle reminder, thanks.

latha-subramanian:
@agrawalravi90,  Thinking more about it,  I feel the change could be done at the test level itself: After -SEGV signal was issued to mom,  loop for 20 seconds with isUp() to make sure mom really went down,  and then do assertion failure if it was up still.  Two reasons, first, I don’t see any other tests, (other than TestPBSTestSuite.test_revert_pbsconf_corelimit) doing such a check and failing, second, there are isUp() functions for server, mom and scheduler and comm. If we add isDown for mom should we also be adding isDown() for all  other daemons to keep uniformity ?

bhroam:
I agree with @latha-subramanian.  Put a check around isUp() in the test itself.  I don't even think you need to wait 20s.  You probably could do a 2-5s sleep and check isUp() after that.  it doesn't take long for a process to die.

agrawalravi90:
@latha-subramanian thanks for your inputs, I agree with you, it's just one test, and we should add isDown in other objects too if we add it in one, so I've just enhanced the test to handle it.

@bhroam To weed out possible race conditions on slower machines, I've kept it 20 iterations of 1 sec each.

Since I changed my approach completely again, I reverted all previous changes and created a fresh commit, please let me know if this looks good. Thanks!

arungrover:
I realized that @latha-subramanian asked for recent test logs. @agrawalravi90 Please post that to the ticket. Sorry about the inconvenience guys.

agrawalravi90:
@arungrover thanks for the merge!

@latha-subramanian here are the final logs:
[final_logs.log](https://github.com/PBSPro/pbspro/files/2973636/final_logs.log)

PR #812, opened 2018-09-19
misuse of qrun -H crashes server

saimanikant:
#### Bug/feature Description
* qrun -H call crashes the server when the node list is ill-formed.
* example:  jid=$(echo "sleep 10" | qsub -l select=ncpus=36:host=centos7+ncpus=36:host=centos7)
                   qrun -H "'(centos7)+(centos7)'" $jid
                   qrun: End of File 4.centos7

#### Affected Platform(s)
* All Platforms

#### Cause / Analysis / Design
* parse_plus_spec_r of src/lib/Libifl/grunt_parse.c explicitly skips everything within quotes including the plus sign , whereas build_execvnode of src/server/node_manager.c  disagrees with it about the significance of plus signs within the quoted string.
* Having identified the number of nodes (2 in the above example) in build_execvnode, ndarray is defined (with one node value as NULL) 
* Seg Fault occurs in strlen() call for the above ndarray.

#### Solution Description
* if qrun is used with a bad -H specification, then reject the call with Unknown node error.
* if the number of nodes defined above in ndarray is not equal to the number of nodes identified, then skip the loop of breaking up the select spec into separate chunks.

#### Testing logs/output
* *Please attach your test log output from running the test you added (or from existing tests that cover your changes)*

#### Checklist:

- [x] I have joined the **[pbspro community forum](http://community.pbspro.org/)**.
- [x] My pull request contains a **single, signed** commit. See **[setting up gpg signature](https://pbspro.atlassian.net/wiki/display/DG/Signing+Your+Git+Commits)**.
- [x] My code follows the **[coding style](https://pbspro.atlassian.net/wiki/display/DG/Coding+Standards)** of this project.
- [ ] My change requires project documentation. See **[required documentation checklist](https://pbspro.atlassian.net/wiki/display/DG/Checklist+for+Developing+Features+and+Bug+Fixes)** for details.
   - [ ] I have added documentation in the **[project documentation area](https://pbspro.atlassian.net/wiki/display/PD)**.
- [x] I have added new **PTL test(s) to my commit**. (See **[using PTL for testing](https://pbspro.atlassian.net/wiki/display/DG/Using+PTL+for+Testing)**) *(or)*
   - [ ] I have added  **manual test(s) to this pull request and explained why PTL is not appropriate** for this case.
- [ ] All new and existing automated tests have passed. (See **[running automated PTL tests](https://pbspro.atlassian.net/wiki/display/DG/PTL+Quick+Start+Guide)**).
[...]

PR #1024, opened 2019-03-07
Server crashes intermittently when the node list in qrun -H is ill formed

saimanikant:
#### Bug/feature Description
* Server crashes intermittently  when the node list in qrun -H is ill formed
* example: jid=$(echo "sleep 10" | qsub -l select=ncpus=36:host=centos7+ncpus=36:host=centos7)
qrun -H "'(centos7)+(centos7)'" $jid
qrun: End of File 4.centos7

#### Affected Platform(s)
* All

* The original fix for this issue in https://github.com/PBSPro/pbspro/pull/812 was returning garbage values in outbuf from build_execvnode().
* This value was part of the modified execvnode string in set_nodes().
* Eventually, this value is passed to find_nodebyname which intermittently returns a non NULL value which was causing the crash (Please check Valgrind logs).

#### Solution Description
* Make sure the modified execvnode string in set_nodes() has the right vnode value by treating the -H argument to qrun as a single node and passing the same value to the 2nd element of ndarray in build_execvnode().
* Before the fix, for the job in the example above, build_execvnode() will return: '(centos7)': ncpus=36:host=centos7+  '(centos7)': ncpus=36:host=centos7 for qrun  command:  
qrun -H "'(centos7)'"  $jid 

Replicating the above behavior , after fix, for qrun -H "'(centos7)+(centos7)'" $jid, 
build_execvnode() will return:
 '(centos7)+(centos7)': ncpus=36:host=centos7+  '(centos7)+(centos7)': ncpus=36:host=centos7

* After Fix Behavior:
   qrun -H "'(centos7)+(centos7)'" $jid
   qrun: Unknown node  '(centos7)+(centos7)'

#### Testing logs/output
* [valgrind_logs_before_fix.txt](https://github.com/PBSPro/pbspro/files/2940804/valgrind_logs_before_fix.txt)
*  [valgrind_logs_after_fix.txt](https://github.com/PBSPro/pbspro/files/2969617/valgrind_logs_after_fix.txt)
* [Test_logs_before_fix.txt](https://github.com/PBSPro/pbspro/files/2940824/Test_logs_before_fix.txt)
* [Test_logs_after_fix.txt](https://github.com/PBSPro/pbspro/files/2969577/Test_logs_after_fix.txt)

#### Checklist:

### DIFF
--- a/src/server/node_manager.c
+++ b/src/server/node_manager.c
@@ -6758,27 +6758,20 @@ build_execvnode(job *pjob, char *nds)
 		pc = parse_plus_spec(NULL, &rc);
 	}
 
+	*outbuf = '\0';
+
        /* 
         * if the number of nodes defined ndarray is not equal
         * to the number of nodes identified by parse_plus_spec, then there's a 
         * invalid vnode specification on the qrun
         */
 
-	if (i != nnodes) {
-		while (i != nnodes) {
-                	if ((*(ndarray+i) = strdup(nds)) == NULL)
-                        	rc = errno;
-			++i;
-		}
-        }
-
-	if (rc)
+	if (rc || i != nnodes)
 		goto done;
 
 	/* now loop breaking up the select spec into separate chunks */
 	/* and determining how many times each chunk is to be used   */
 
-	*outbuf = '\0';
 	i  = 0;
 	pc = parse_plus_spec(selspec, &rc);
 	while (pc) {
@@ -7037,6 +7030,9 @@ set_nodes(void *pobj, int objtype, char *execvnod_in, char **execvnod_out, char 
 		}
 		if (execvnod == NULL)
 			return PBSE_BADNODESPEC;
+
+		if (!strlen(execvnod))
+                        return PBSE_UNKNODE;
 
 		/* are we to allocate the nodes "excl" ? */
 		prsdef = find_resc_def(svr_resc_def, "place", svr_resc_size);
--- a/src/server/node_manager.c
+++ b/src/server/node_manager.c
@@ -6759,11 +6759,20 @@ build_execvnode(job *pjob, char *nds)
 	}
 
        /* 
-        * if the number of nodes defined above in ndarray is not equal
-        * to the number of nodes identified, then skip the below loop
+        * if the number of nodes defined ndarray is not equal
+        * to the number of nodes identified by parse_plus_spec, then there's a 
+        * invalid vnode specification on the qrun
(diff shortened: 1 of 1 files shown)


saimanikant:
@prakashcv13, returning NULL from build_execvnode will result in a PBSE_BADNODESPEC error, which would be misleading. The proposed changes will make sure  build_execvnode understands that qrun has a single node as its argument for "'(centos)+(centos)'" and throw a  Unknown node error by passing that value to find_nodebyname.

prakashcv13:
@saimanikant - To me unknown node is misleading. When the user provides the vnode specification, they do not consider (d_server)+(d_server) as a single node. The error message unknown node for (d_server)+(d_server) is not helping the user.
The issue is parse_plus_spec_r is looking for a parenthesis first and then for the quote, therefore it fails to parse the provided vnode specification. If we remove the quotes qrun will be successful. So, to me, it makes sense that we return PBSE_BADNODESPEC instead of unknown node OR we change parse_plus_spec_r to work with quotes before the parenthesis.

saimanikant:
@prakashcv13, when a user provides the vnode specification inside **single quotes**, like qrun -H "'(d_server)+(d_server)'", PBS considers it as a single node value inside parse_plus_spec_r, just like you pointed out. I don't think we should make changes in parse_plus_spec_r and accept such inputs because the node list here is ill-formed (shouldn't be quoted). 
The reason for saying we should error out with Unknown node message here is, if we take the following case and notice the behaviour:
qsub -- /bin/sleep 100
qrun -H "'(d_server)'" $jid
this request will be rejected with the following message:
qrun: Unknown node '(d_server)'
Here, we will not hit the  (i != nnodes) condition. The proposed changes will maintain the same behavior even when the  (i != nnodes) condition is true.

bhroam:
As a note, you can pass an exec_vnode to qrun -H.  This is the entire fully formed exec_vnode as the scheduler does.  So an example would be (centos:ncpus=1)+(centos:ncpus=1).  The server will consider something passed to qrun -H an exec_vnode if it has a ':' in it.  It think you are playing scheduler and will take it as is.

saimanikant:
[valgrind_logs.txt](https://github.com/PBSPro/pbspro/files/2991354/valgrind_logs.txt)
[ptl_test_results.txt](https://github.com/PBSPro/pbspro/files/2991355/ptl_test_results.txt)

prakashcv13:
merged #1024

PR #1034, opened 2019-03-13
PTL setUp takes too long when there's large no. of vnodes

agrawalravi90:
#### Bug/feature Description
* PTL's setUp() was taking to long (15+ minutes) when there are large number of nodes (~10k) in the system (reference testsuite: TestSchedPerf)

#### Affected Platform(s)
* Linux

#### Cause / Analysis
* There were 2 bottlenecks:
  1) pbs_init.d restart: This was taking 10+ minutes. Cause: the server function check_and_set_multivnode() takes too long when there are that many nodes in the system
  2) delete custom resource: Deleting just one custom resource was taking 4+ minutes. Cause: when PBS deletes a resource, it has to unset it from all the PBS objects where it is set

#### Solution Description
* Modified Server's revert_to_default() to delete all nodes before deleting all custom resources
* Reordered the revert_to_defaults calls in setUp() so that revert_server is called first, this makes sure that we don't restart pbs before all the nodes have been deleted from the server

#### Testing logs/output
* I saw a performance improvement of around 51 minutes (~39%) when running TestSchedPerf. The run time went down from 2 hours 11 mins to 1 hour 20 mins. Logs here:
[testschedperf_log.tar.gz](https://github.com/PBSPro/pbspro/files/2964010/testschedperf_log.tar.gz)

Please note that one of the test fails in both the original and after changes logs, so it has nothing to do with these changes. Hopefully we can run performance tests more often with this performance boost.

#### Checklist:

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -5031,7 +5031,8 @@ class Server(PBSService):
 
     def revert_to_defaults(self, reverthooks=True, revertqueues=True,
                            revertresources=True, delhooks=True,
-                           delqueues=True, delscheds=True, server_stat=None):
+                           delqueues=True, delscheds=True, delnodes=True,
+                           server_stat=None):
         """
         reset server attributes back to out of box defaults.
 
@@ -5060,6 +5061,8 @@ class Server(PBSService):
                           The sched_priv and sched_logs directories will be
                           deleted.
         :type delscheds: bool
+        :param delnodes: If True all vnodes are deleted
+        :type delnodes: bool
         :returns: True upon success and False if an error is
                   encountered.
         :raises: PbsStatusError or PbsManagerError
@@ -5141,6 +5144,12 @@ class Server(PBSService):
                                recursive=True, force=True)
                     self.manager(MGR_CMD_DELETE, SCHED, id=name)
 
+        if delnodes:
+            try:
+                self.manager(MGR_CMD_DELETE, VNODE, id="@default")
+            except PbsManagerError as e:
+                if "Unknown node" not in e.msg[0]:
+                    raise
         if reverthooks:
             if self.platform == 'cray' or self.platform == 'craysim':
                 if self.du.cmp(self.hostname, self.dflt_mpp_hook,
--- a/test/fw/ptl/utils/pbs_testsuite.py
+++ b/test/fw/ptl/utils/pbs_testsuite.py
@@ -477,11 +477,11 @@ class PBSTestSuite(unittest.TestCase):
             return
         self.log_enter_setup()
         self.init_proc_mon()
-        self.revert_pbsconf()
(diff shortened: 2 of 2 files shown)


agrawalravi90:
@arungrover and @bhroam can you please take a look at this when you get time? Thanks.

agrawalravi90:
> My only comment is that I see you deleting the nodes in a new place, but I don't see you removing the old place. Aren't you deleting them twice now?

Earlier we were doing it as part of reverting mom to default. We'd delete vnodes for each mom, 1 mom at a time. Now, revert of server deletes all vnodes from the server, so the revert of mom tries to query vnodes on the mom and just returns when it doesn't find any vnodes on it. I had previously tried to do it as part of reverting mom itself, but it didn't work for the following reasons:
1) Mom's delete_vnodes() only deletes vnodes which are on the same host as the mom node. So, it wasn't actually deleting all vnodes which reported to that mom.
2) Even if we fix 1), reverting moms deletes only the vnodes associated to that mom, and each mom revert needs to restart PBS as it also reverts the mom config files. Restarting PBS is very expensive when there are many nodes in the system, so this wouldn't give the desired performance boost.

bhroam:
I actually like how you have done this.  My comment was more to the point that you added code to delete nodes.  Why didn't you delete the code that was doing it before.  You said that the old code now queries the nodes and then does nothing.  If it is going to do nothing, why keep it?

agrawalravi90:
> You said that the old code now queries the nodes and then does nothing. If it is going to do nothing, why keep it?

So, it does nothing because it sees no nodes to delete (as they've already been deleted by the server's revert. I didn't remove the deletion of nodes from mom's revert because I didn't want to change the behavior of Mom's revert. As an interface, I kind of feel like revert_to_defaults() for Mom should probably delete the associated vnodes as this function can be called directly as well. What do you think?

arungrover:
I also feel it is ok to leave deletion of nodes in mom's revert to default because a test might want to call revert_to_default() explicitly on a specific mom and in that case we might want to delete the nodes.

bhroam:
What nodes does mom.revert_to_defaults() delete?  If it deletes @default, then it's a little confusing.  If we keep the code, then it should only delete the nodes that the mom provides.  To do this, you'd have to do a s = self.server.status(NODE) and then [s[x]['id'] for x in s if self.shortname in s[x]['moms']].  If it does that, then I'm fine with it.  If reverting any mom deletes all the nodes, then I find it confusing.

agrawalravi90:
@bhroam If reverting a mom deletes any node that is not associated with it, then we should fix revert_to_defaults() Mom. Looking at the code, and my experience with it, it should not delete any node not associated with the mom, so we should be fine.

bhroam:
Looks good to me.  Thanks for explaining, @agrawalravi90.

bhroam:
I merged in the PR.

PR #1036, opened 2019-03-14
LogUtils.convert_date_time() doesn't handle DST properly

kjakkali:
#### Bug/feature Description
LogUtils.convert_date_time() doesn't handle DST properly

#### Affected Platform(s)
All

* The convert_date_time() function was recently rewritten to handle microsecond logging.  The original method of converting times to epoch times was using the time object.  The time object doesn't support microseconds.  The newer datetime object does support microseconds.  The new convert_date_time() function uses the datetime object.  Unfortunately it does not handle DST properly.  For example, converting to PDT will convert to PST and be off by an hour.

#### Solution Description
* Updated LogUtils.convert_date_time() to deduct extra 3600 sec if DST

#### Checklist:

- [ ] My code follows the **[coding style](https://pbspro.atlassian.net/wiki/display/DG/Coding+Standards)** of this project.

### DIFF
--- a/test/fw/ptl/utils/pbs_logutils.py
+++ b/test/fw/ptl/utils/pbs_logutils.py
@@ -236,6 +236,9 @@ class PBSLogUtils(object):
         if dt is None:
             return None
 
+        offset = 0
+        if time.daylight:
+            offset = 3600
         micro = False
         if fmt is None:
             if '.' in dt:
@@ -250,7 +253,7 @@ class PBSLogUtils(object):
             # Get timedelta object of epoch time
             t -= epoch_datetime
             # get epoch time from timedelta object
-            tm = t.total_seconds()
+            tm = t.total_seconds() - offset
         except:
             cls.logger.debug("could not convert date time: " + str(datetime))
             return None
--- a/test/fw/ptl/utils/pbs_logutils.py
+++ b/test/fw/ptl/utils/pbs_logutils.py
@@ -38,7 +38,7 @@
 import re
 import time
 import sys
-import datetime
+from datetime import tzinfo, timedelta, datetime
 import logging
 import traceback
 import math
@@ -209,7 +209,7 @@ PARSER_OK_STOP = 1
 PARSER_ERROR_CONTINUE = 2
 PARSER_ERROR_STOP = 3
 
-epoch_datetime = datetime.datetime.fromtimestamp(0)
+epoch_datetime = datetime.fromtimestamp(0)
 
 
 class PBSLogUtils(object):
@@ -225,7 +225,8 @@ class PBSLogUtils(object):
     def convert_date_time(cls, dt=None, fmt=None):
         """
         convert a date time string of the form given by fmt into
-        number of seconds since epoch (with possible microseconds)
+        number of seconds since epoch (with possible microseconds).
+        it considers the current system's timezone to convert the datetime to epoch time
 
         :param dt: the datetime string to convert
         :type dt: str or None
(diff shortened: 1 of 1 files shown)


prakashcv13:
@kjakkali - could we not use 3600 directly and use the logic similar to what is used on this [page](https://docs.python.org/2/library/datetime.html)
STDOFFSET = timedelta(seconds = -_time.timezone)
if _time.daylight:
    DSTOFFSET = timedelta(seconds = -_time.altzone)
else:
    DSTOFFSET = STDOFFSET

DSTDIFF = DSTOFFSET - STDOFFSET

test/fw/ptl/utils/pbs_logutils.py
```diff
@@ -236,6 +238,9 @@ def convert_date_time(cls, dt=None, fmt=None):
         if dt is None:
             return None
 
+        offset = timedelta(seconds=-time.timezone)
```
agrawalravi90: I'm not sure if I understand this change. time.timezone will return the offset of current Tz frm UTC, so for me it gives offset as 5:30:00. Which means that if time.daylight is not True, then the code will directly subtract offset.total_seconds() from the epoch time, so that will subtract 19800.0 seconds from the epoch time?
Shouldn't you be setting the offset to just (-time.altzone - time.timezone) ?

test/fw/ptl/utils/pbs_logutils.py
```diff
@@ -236,6 +238,12 @@ def convert_date_time(cls, dt=None, fmt=None):
         if dt is None:
             return None
 
+        stdoffset = timedelta(seconds=-time.timezone)
+        if time.daylight:
+            dstoffset = timedelta(seconds=-time.altzone)
+        else:
+            dstoffset = stdoffset
+        offsetdiff = dstoffset - stdoffset
```
agrawalravi90: ok, you fixed it yourself by the time i posted my review. I think we can make this simpler by just setting offsetdiff to timedelta(seconds=-time.altzone) - timedelta(seconds=-time.timezone) in all cases. When daylight savings is off, timedelta(seconds=-time.altzone) seems to get set to timedelta(seconds=-time.timezone).

prakashcv13: @agrawalravi90 - timedelta(seconds=-time.altzone) - timedelta(seconds=-time.timezone) would not work as it will not give the number of seconds.
We should have -
offsetdiff = (int)(timedelta(seconds=-time.altzone) - timedelta(seconds=-time.timezone)).total_seconds()
the typecast (int) may not be needed. but we would need total_seconds()

kjakkali: @agrawalravi90 : I think making it generic may not work on DST timezone. 
Also Python doc say "time.altzone
The offset of the local DST timezone, in seconds west of UTC, if one is defined. This is negative if the local DST timezone is east of UTC (as in Western Europe, including the UK). **Only use this if daylight is nonzero**."  Shall we keep the changes as it is?

agrawalravi90: Ah, I didn't read that part, yes, please keep the changes as is then.

kjakkali:
@vstumpf , @agrawalravi90 , @prakashcv13 : Thank you.

test/fw/ptl/utils/pbs_logutils.py
prakashcv13: I think generic will work. I tried the below code on my VM and one in SV, the results were as expected.
from datetime import timedelta, datetime
import time
offsetdiff = (int)((timedelta(seconds=-time.altzone) - timedelta(seconds=-time.timezone)).total_seconds())
print offsetdiff

However, we should follow what the documentation says.

PR #1040, opened 2019-03-16
Removing expect=True from manager()

bhroam:
#### Bug/feature Description
* PTL's manager() call has an argument called expect.  It was thought that if you set it to true, it would make sure your values were set properly.  All it was really doing was checking if the values are set to anything.  This functionality was not all that useful to begin with.  An exception is raised if manager() fails.  We should trust PBS to correctly set the value if it succeeds.

It also tried to wait until a scheduling cycle ended if you set scheduling=False.  It wouldn't work properly because it tried to check if server_state != scheduling.  Since op=SET instead of op=EQ, it only really checked if server_state was set.  Which it was, it was set to 'scheduling'

manager() was passing op=SET into expect() instead of op=EQ.  We can't just change this to op=EQ since there are many ways the server morphs the input to different output (e.g. queue_type=e turns into Execution or scheduling=1 turns into scheduling=True).

#### Solution Description
Remove the call expect and max_attempts arguments and the call to expect() except for scheduling=False.  When scheduling is set to False, we wait for the cycle to end.  Since a cycle can take up to 20m, max_attempts is set to 1200 and interval is set to 1.

Forum Discussion: http://community.pbspro.org/t/ptl-manager-with-expect-true/1427/28

#### Testing logs/output
[manager_test.log](https://github.com/PBSPro/pbspro/files/2973262/manager_test.log)

#### Checklist:

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -5106,7 +5106,7 @@ class Server(PBSService):
             hooks = self.status(HOOK, level=logging.DEBUG)
             hooks = [h['id'] for h in hooks]
             if len(hooks) > 0:
-                self.manager(MGR_CMD_DELETE, HOOK, id=hooks, expect=True)
+                self.manager(MGR_CMD_DELETE, HOOK, id=hooks)
         if delqueues:
             revertqueues = False
             queues = self.status(QUEUE, level=logging.DEBUG)
@@ -5120,11 +5120,11 @@ class Server(PBSService):
                                          node['id'])
                 except:
                     pass
-                self.manager(MGR_CMD_DELETE, QUEUE, id=queues, expect=True)
+                self.manager(MGR_CMD_DELETE, QUEUE, id=queues)
             a = {ATTR_qtype: 'Execution',
                  ATTR_enable: 'True',
                  ATTR_start: 'True'}
-            self.manager(MGR_CMD_CREATE, QUEUE, a, id='workq', expect=True)
+            self.manager(MGR_CMD_CREATE, QUEUE, a, id='workq')
             setdict.update({ATTR_dfltque: 'workq'})
         if delscheds:
             self.manager(MGR_CMD_LIST, SCHED)
@@ -5152,8 +5152,7 @@ class Server(PBSService):
             hooks = [h['id'] for h in hooks]
             a = {ATTR_enable: 'false'}
             if len(hooks) > 0:
-                self.manager(MGR_CMD_SET, MGR_OBJ_HOOK, a, hooks,
-                             expect=True)
+                self.manager(MGR_CMD_SET, MGR_OBJ_HOOK, a, hooks)
         if revertqueues:
             self.status(QUEUE, level=logging.DEBUG)
             queues = []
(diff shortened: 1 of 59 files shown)


test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -5179,7 +5178,7 @@ def revert_to_defaults(self, reverthooks=True, revertqueues=True,
             except:
                 rescs = []
             if len(rescs) > 0:
-                self.manager(MGR_CMD_DELETE, RSC, id=rescs, expect=True)
```
hirenvadalia: Shouldn't we need to add expext() after manager() since we removed expect=True?

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -5106,7 +5106,7 @@ def revert_to_defaults(self, reverthooks=True, revertqueues=True,
             hooks = self.status(HOOK, level=logging.DEBUG)
             hooks = [h['id'] for h in hooks]
             if len(hooks) > 0:
-                self.manager(MGR_CMD_DELETE, HOOK, id=hooks, expect=True)
+                self.manager(MGR_CMD_DELETE, HOOK, id=hooks)
```
agrawalravi90: I think all of the places where the framework was previously doing an expect=True, we should now do an expect() explicitly to check the same. This will help catch errors at the setUp() stage itself.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -6469,22 +6460,9 @@ def manager(self, cmd, obj_type, attrib=None, id=None, extend=None,
                        i.e. silent mode
         :type logerr: bool
         :raises: PbsManagerError
-        :returns: On success:
-                - if expect is False, return code of qmgr/pbs_manager
-                - if expect is True, 0 for success
-        :raises: On Error/Failure:
                 - PbsManagerError if qmgr/pbs_manager() failed
```
agrawalravi90: line 6462 already mentions "raises: PbsManagerError", so I think we can remove this line as well

test/tests/selftest/pbs_expect.py
```diff
@@ -57,7 +57,7 @@ def test_attribute_case(self):
         # Set the Priority attribute on the queue but provide 'p' lowercase
         # Set other attributes normally
         a = {'enabled': 'True', 'started': 'True', 'priority': 150}
-        self.server.manager(MGR_CMD_SET, QUEUE, a, 'expressq', expect=True)
+        self.server.manager(MGR_CMD_SET, QUEUE, a, 'expressq')
```
agrawalravi90: This test really doesn't make any sense without a call to expect(). Can you please add an explicit call to expect() after the call to manager() ?

agrawalravi90:
Besides the SmokeTest logs, can you please also post logs for all the tests that you had to modify with this change? Thanks.

bhroam:
I did run all the tests, but I did it through the internal test harness.  If you want to see the logs, I can give you a range of TH job ids.  I did not run them manually on my dev system.

test/fw/ptl/lib/pbs_testlib.py
bhroam: I can, but the logic behind removing expect=True was that it was not necessary.  The manager() call itself will raise an exception if it fails.  We should trust PBS to do the operation properly if it didn't fail.

test/fw/ptl/lib/pbs_testlib.py
bhroam: As I said in @hirenvadalia's comment, I don't think we should.  The logic behind removing expect=True was it was not necessary.  Adding in expect() calls will just make revert_to_defaults() take longer for no real benefit.

agrawalravi90: Ok, but by not doing an expect, I fear that there could be race conditions: setUp fires off PBS commands which might still be running (on a slow machine for example) while setUp ends and the test starts.

agrawalravi90:
@bhroam that's not necessary, thanks for doing a full regression for this

test/fw/ptl/lib/pbs_testlib.py
bhroam: I'm not sure I understand.  The manager() call doesn't return until qmgr has been called.  This means the value has been set.  How can we have a race condition on slow machines?

agrawalravi90: Ah, yes, you are correct, manager() is a blocking call so it shouldn't cause any issues, my bad.

PR #1044, opened 2019-03-19
Strict ordering does not work if backfill_depth on queue is 0

arungrover:
#### Bug/feature Description
When strict_ordering is true and backfill_depth is set to 0, the scheduler should stop considering jobs to run when it can't run one.  This means it will run as many jobs as it can until it can't run one.  It will then end the cycle. 
This does not happen if backfill_depth on a queue is set to 0.

#### Affected Platform(s)
All

When backfill is enabled but backfill_depth on a queue is set to 0  and strict ordering is enabled, we do not mark the jobs in that queue as can-not-run. This results into considering all the jobs of the queue one after another.

#### Solution Description
When strict_ordering is true and backfill_depth on a queue is set to 0, mark all the jobs of that queue as can-not-run.

#### Testing logs/output
[test_strict_ordering.txt](https://github.com/PBSPro/pbspro/files/2985871/test_strict_ordering.txt)

#### Checklist:

### DIFF
--- a/src/scheduler/fifo.c
+++ b/src/scheduler/fifo.c
@@ -1071,6 +1071,10 @@ main_sched_loop(status *policy, int sd, server_info *sinfo, schd_error **rerr)
 				set_schd_error_arg(err, SPECMSG, "Job would conflict with starving job");
 				update_jobs_cant_run(sd, sinfo->jobs, NULL, err, START_WITH_JOB);
 			}
+			else if (policy->backfill && policy->strict_ordering && qinfo->backfill_depth == 0) {
+				set_schd_error_codes(err, NOT_RUN, STRICT_ORDERING);
+				update_jobs_cant_run(sd, qinfo->jobs, NULL, err, START_WITH_JOB);
+			}
 		}
 
 		time(&cur_time);
--- a/test/tests/functional/pbs_strict_ordering.py
+++ b/test/tests/functional/pbs_strict_ordering.py
@@ -54,8 +54,8 @@ class TestStrictOrderingAndBackfilling(TestFunctional):
         self.server.create_vnodes('vn', a, 1, self.mom, usenatvnode=True)
 
         rv = self.scheduler.set_sched_config(
-            {'round_robin': 'false all', 'by_queue': 'false prime',
-             'by_queue': 'false non_prime', 'strict_ordering': 'true all',
+            {'round_robin': 'false all', 'by_queue': 'false all',
+             'strict_ordering': 'true all',
              'help_starving_jobs': 'false all'})
         self.assertTrue(rv)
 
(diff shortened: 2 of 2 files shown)


src/scheduler/fifo.c
```diff
@@ -1071,6 +1071,10 @@ main_sched_loop(status *policy, int sd, server_info *sinfo, schd_error **rerr)
 				set_schd_error_arg(err, SPECMSG, "Job would conflict with starving job");
 				update_jobs_cant_run(sd, sinfo->jobs, NULL, err, START_WITH_JOB);
 			}
+			else if (policy->backfill && policy->strict_ordering && qinfo->backfill_depth == 0) {
```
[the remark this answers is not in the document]
arungrover: If you set server's backfill_depth to 0 then policy->backfill is turned off and that case is handled on line 1064 just above this code

test/tests/functional/pbs_strict_ordering.py
```diff
@@ -164,3 +163,104 @@ def test_t2(self):
                            id=j5id,
                            max_attempts=30,
                            interval=2)
+
+    def test_zero_backfill_depth_on_queue(self):
+        """
+        Test if scheduler tries to run a job when strict ordering is enabled
+        and backfill_depth is set to 0 on the queue
```
bhroam: These tests are fine, but I think we can make them slightly better.  Instead of having 1 cpu and submitting 3 identical jobs, how about we have 2 cpus and submit jobs requesting 1 cpu, 2 cpus, 1cpu.  If backfill_depth wasn't 0, we'd run jobs 1 and 3.  Since backfill_depth is 0, we won't run job 3.

arungrover:
@bhroam thanks for your review comment, I have addressed it in my latest commit

test/tests/functional/pbs_strict_ordering.py
```diff
@@ -204,6 +206,42 @@ def test_zero_backfill_depth_on_queue(self):
...
+        j6 = Job(TEST_USER)
+        j6.set_attributes(a)
+        jid6 = self.server.submit(j6)
+
+        a = {'scheduling': 'True'}
+        self.server.manager(MGR_CMD_SET, SERVER, a)
+
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid4)
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid6)
+        self.scheduler.log_match(jid5 + ";Job is a top job")
```
bhroam: If you submit your these 3 jobs as your first 3 jobs (they're the same but with walltime), you don't need these.  You can reuse the same 3 jobs for both sides of this test.

arungrover: I feel the way it is right now is cleaner because in the other implementation I'll have to make sure the jobs have enough sleep/walltime so that on slower machines they don't end in between and for log match I'll have to make sure that we specify the timestamp.
I'd like to retain it as is because functionality wise I feel it is not different than keeping the same jobs across cycles.

bhroam: OK, I'll buy that.  Thanks for explaining.

PR #1048, opened 2019-03-21
Fixed race condition in test_pbsfs

bhroam:
#### Bug/feature Description
* This test had a race condition because it did the following:
1) Run a cycle
2) wait for the cycle to finish by looking in the log for the 'Leaving scheduling cycle' message
3) Look at the order of jobs that ran in the last cycle.
--> One ran, one didn't.

The race condition is that a second cycle can run after 2 and before 3.  Since one job is running, only the second job is considered in the new cycle.

#### Solution Description
What needed to happen is to turn scheduling false immediately after turning scheduling to true.  Now that PTL will wait for the cycle to finish when you turn scheduling to false, this should remove the race condition.

#### Testing logs/output
[pbsfs.log](https://github.com/PBSPro/pbspro/files/2994381/pbsfs.log)

### DIFF
--- a/test/tests/functional/pbs_fairshare.py
+++ b/test/tests/functional/pbs_fairshare.py
@@ -313,11 +313,8 @@ class TestFairshare(TestFunctional):
         J2 = Job(TEST_USER1)
         jid2 = self.server.submit(J2)
 
-        t = int(time.time())
         self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'True'})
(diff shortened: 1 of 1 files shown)


bhroam:
@agrawalravi90 if you could take a look at this PR, that'd be great.  It's just a couple of lines of code change.

agrawalravi90:
Sorry for the delay, changes look good!

PR #1055, opened 2019-03-27
Node buckets test fails due to race condition

arungrover:
#### Bug/feature Description
Some of the node buckets tests fail once in a while.

#### Affected Platform(s)
All

Tests fail because at the beginning of the test we are trying to make a config change and then expect that config change to have taken effect but the problem occurs when scheduling cycle was already running when this config change happens and as a result, it has not taken effect in that cycle. This will make the test think that scheduler has taken wrong decisions and fail the test.

#### Solution Description
Turn off the sched cycle before we do config changes and then turn them on before we expect the job to run.

#### Testing logs/output
[test_node_buckets.txt](https://github.com/PBSPro/pbspro/files/3015274/test_node_buckets.txt)

#### Checklist:

- [xgoto] My code follows the **[coding style](https://pbspro.atlassian.net/wiki/display/DG/Coding+Standards)** of this project.

### DIFF
--- a/test/tests/functional/pbs_node_buckets.py
+++ b/test/tests/functional/pbs_node_buckets.py
@@ -614,7 +614,9 @@ class TestNodeBuckets(TestFunctional):
         Test that jobs in the calendar fit within a placement set
         """
         self.scheduler.set_sched_config({'strict_ordering': 'True'})
-        self.server.manager(MGR_CMD_SET, SERVER, {'backfill_depth': 5})
+        svr_attr = {'node_group_key': 'shape', 'node_group_enable': 'True',
+                    'backfill_depth': 5}
+        self.server.manager(MGR_CMD_SET, SERVER, svr_attr)
 
         chunk1 = '10010:ncpus=1'
         a = {'Resource_List.select': chunk1,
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_node_buckets.py
```diff
@@ -695,6 +695,10 @@ def test_psets_spanning(self):
         Request more nodes than available in one placement set and see
         the job span or not depending on the value of do_not_span_psets
         """
+        # Turn off scheduling to be sure there is no cycle running when
+        # configurations are changed
+        a = {'scheduling': 'False'}
+        self.server.manager(MGR_CMD_SET, SERVER, a)
         a = {'node_group_key': 'shape', 'node_group_enable': 'True'}
```
agrawalravi90: combine the two manager calls?

arungrover: You are right these two calls can be combined, but I'd like to retain if you think the impact of this code is not huge. The reason is that this bug is a timing issue and occurs once in a while which makes verification of this bug difficult. I have to run the same testsuite on a slow machine 20 times to make sure it is fixed which takes about 15hrs.

PR #1060, opened 2019-03-28
PP-848: Ability to include requested custom resources in mom hook input file.

ashwathraop:
#### Bug/feature Description
* *Currently hook writers only get a limited amount of information in the hook input file. For example, if I request ngpus, I need to parse the exec_vnode attribute and then identify the node to find out if a gpu was requested for the node. If I want the software attribute or custom resources, I have to query the server for it. This can have a negative impact on the server and introduces potential hangs. I would like for all the information requested by the job or set by the server using defaults to be available in the hook input file.*
* *Issue ID: [PP-848](https://pbspro.atlassian.net/browse/PP-848)*

#### Affected Platform(s)
* *All*

* *PBS does not send custom resources associated with jobs to MoMs.*

#### Solution Description
* *Introduce a new flag "m" for custom resources which will allow PBS to send those custom resources defined with the aforementioned flag to MoMs*
* EDD can be found here: [EDD](https://pbspro.atlassian.net/wiki/spaces/PD/pages/1034158087/Ability+to+include+requested+custom+resources+in+mom+hook+input+file.)

#### Testing logs/output
[ptl_log.txt](https://github.com/PBSPro/pbspro/files/3020083/ptl_log.txt)

#### Checklist:

### DIFF
--- a/test/fw/ptl/utils/pbs_snaputils.py
+++ b/test/fw/ptl/utils/pbs_snaputils.py
@@ -946,12 +946,15 @@ class _PBSSnapUtils(object):
 
         :returns: numeric value of the resource flag
         """
+        # Variable assignments below mirrors definitions
+        # from src/include/pbs_internal.h
         ATR_DFLAG_USRD = 0x01
         ATR_DFLAG_USWR = 0x02
         ATR_DFLAG_OPRD = 0x04
         ATR_DFLAG_OPWR = 0x08
         ATR_DFLAG_MGRD = 0x10
         ATR_DFLAG_MGWR = 0x20
+        ATR_DFLAG_MOM = 0x400
         ATR_DFLAG_RASSN = 0x4000
         ATR_DFLAG_ANASSN = 0x8000
         ATR_DFLAG_FNASSN = 0x10000
@@ -971,6 +974,8 @@ class _PBSSnapUtils(object):
             resc_flag |= ATR_DFLAG_ANASSN
         if "h" in flag:
             resc_flag |= ATR_DFLAG_CVTSLT
+        if "m" in flag:
+            resc_flag |= ATR_DFLAG_MOM
         if "r" in flag:
             resc_flag &= ~READ_WRITE
             resc_flag |= NO_USER_SET
--- a/test/tests/functional/pbs_resc_custom_perm.py
+++ b/test/tests/functional/pbs_resc_custom_perm.py
@@ -1,0 +1,140 @@
+# coding: utf-8
+
+# Copyright (C) 1994-2019 Altair Engineering, Inc.
+# For more information, contact Altair at www.altair.com.
+#
+# This file is part of the PBS Professional ("PBS Pro") software.
+#
+# Open Source License Information:
+#
+# PBS Pro is free software. You can redistribute it and/or modify it under the
+# terms of the GNU Affero General Public License as published by the Free
+# Software Foundation, either version 3 of the License, or (at your option) any
+# later version.
+#
+# PBS Pro is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.
+# See the GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+# Commercial License Information:
+#
+# For a copy of the commercial license terms and conditions,
+# go to: (http://www.pbspro.com/UserArea/agreement.html)
+# or contact the Altair Legal Department.
+#
+# Altair’s dual-license business model allows companies, individuals, and
+# organizations to create proprietary derivative works of PBS Pro and
+# distribute them - whether embedded or bundled with other software -
+# under a commercial license agreement.
+#
+# Use of Altair’s trademarks, including but not limited to "PBS™",
+# "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
+# trademark licensing policies.
+
+from tests.functional import *
+
+
+class Test_Custom_Resource_Perm(TestFunctional):
+    rsc_list = ['foo_str', 'foo_strm', 'foo_strh']
+    hook_list = ["start", "begin"]
+
+    def setUp(self):
+        TestFunctional.setUp(self)
+
+        self.momA = self.moms.values()[0]
+        self.momA.delete_vnode_defs()
+        self.hostA = self.momA.shortname
+
+        rc = self.server.manager(MGR_CMD_DELETE, NODE, None, "")
+        self.assertEqual(rc, 0)
+
+        rc = self.server.manager(MGR_CMD_CREATE, NODE, id=self.hostA)
+        self.assertEqual(rc, 0)
+        self.server.expect(NODE, {'state': 'free'}, id=self.hostA)
+
+        # Next set custom resources with m flag and without m flag
+        # via qmgr -c 'create resource'
+        attr = {'type': 'string', 'flag': 'h'}
+        r = 'foo_str'
+        self.server.manager(
+            MGR_CMD_CREATE, RSC, attr, id=r, logerr=False)
+
+        attr = {'type': 'string', 'flag': 'hm'}
+        r = 'foo_strm'
+        self.server.manager(MGR_CMD_CREATE, RSC, attr, id=r, logerr=False)
+
+        # Create a custom resources via exechost_startup hook.
+        startup_hook_body = """
+import pbs
+e=pbs.event()
+localnode=pbs.get_local_nodename()
+e.vnode_list[localnode].resources_available['foo_strh'] = "str_hook"
+"""
+        hook_name = "start"
+        a = {'event': "exechost_startup", 'enabled': 'True'}
+        self.server.create_import_hook(
+            hook_name,
+            a,
+            startup_hook_body,
+            overwrite=True)
+
+        # Restart the MoM so that exechost_startup hook can run.
+        self.momA.restart()
+
+        # Give the moms a chance to receive the updated resource.
+        # Ensure the new resource is seen by all moms.
+        self.momA.log_match("resourcedef;copy hook-related file",
+                            max_attempts=20, interval=1)
+
+        # Create a exechost_begin hook.
+        begin_hook_body = """
+import pbs
+e=pbs.event()
+res_list = getattr(e.job, 'Resource_List')
+pbs.logjobmsg(e.job.id, "Resource_List is %s" % str(res_list))
+"""
+        hook_name = "begin"
+        a = {'event': "execjob_begin", 'enabled': 'True'}
+        self.server.create_import_hook(
+            hook_name,
+            a,
+            begin_hook_body,
+            overwrite=True)
+
+    def test_custom_resc_single_node(self):
+        """
+        Test permission flag of resources of a single node job
+        using a execjob_begin hook.
+        """
+        self.logger.info("test_custom_resc__single_node")
+
+        a = {'Resource_List.foo_str': 'str_noperm',
+             'Resource_List.foo_strm': 'str_perm',
+             'Resource_List.foo_strh': 'str_hook'
+             }
+        j = Job(TEST_USER, a)
+        j.set_sleep_time("100")
+        jid = self.server.submit(j)
+
+        self.server.expect(JOB, {
+            'job_state': 'R',
+            'Resource_List.foo_str': 'str_noperm',
+            'Resource_List.foo_strm': 'str_perm',
+            'Resource_List.foo_strh': 'str_hook'},
+            offset=1, id=jid)
+
+        # Match mom logs entry, only resoruces with "m" flag would show up.
+        msg = 'foo_strm=str_perm'
+        self.momA.log_match("Job;%s;.*%s.*" % (jid, msg),
+                            regexp=True, n=10, max_attempts=10, interval=2)
+        msg = 'foo_strh=str_hook'
+        self.momA.log_match("Job;%s;.*%s.*" % (jid, msg),
+                            regexp=True, n=10, max_attempts=10, interval=2)
+        msg = 'foo_str=str_noperm'
+        self.momA.log_match("Job;%s;.*%s.*" % (jid, msg),
+                            regexp=True, n=5, max_attempts=10, interval=2,
+                            existence=False)
--- a/test/tests/pbs_smoketest.py
+++ b/test/tests/pbs_smoketest.py
@@ -48,7 +48,7 @@ class SmokeTest(PBSTestSuite):
     # Class variables
     resc_types = [None, 'long', 'float', 'boolean', 'size', 'string',
                   'string_array']
-    resc_flags = [None, 'n', 'h', 'nh', 'q', 'f', 'fh']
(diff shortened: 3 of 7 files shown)


test/tests/pbs_smoketest.py
```diff
@@ -48,7 +48,7 @@ class SmokeTest(PBSTestSuite):
     # Class variables
     resc_types = [None, 'long', 'float', 'boolean', 'size', 'string',
                   'string_array']
-    resc_flags = [None, 'n', 'h', 'nh', 'q', 'f', 'fh']
+    resc_flags = [None, 'n', 'h', 'nh', 'q', 'f', 'fh', 'm']
```
mike0042: Might "mh" also be a valid combination?

subhasisb:
@ashwathraop we should think of extending this to all attributes. And if we do that then we probably may not need this "send_job" vehicle? I think we need to consider this along with the overall flow here.

test/tests/pbs_smoketest.py
ashwathraop: Yes. Added mh to the list now.

ashwathraop:
@subhasisb I would recommend having separate PR for sending server/queue attributes to MoMs as we might need a different approach for that. For resources associated with jobs, it makes sense to go through send job route. What do you think?

ashwathraop:
[ptl_smoke.txt](https://github.com/PBSPro/pbspro/files/3139211/ptl_smoke.txt)
[ptl.txt](https://github.com/PBSPro/pbspro/files/3139213/ptl.txt)
Test run after pep8 and adding 'mh' flag to smoke test.

subhasisb:
> @subhasisb I would recommend having separate PR for sending server/queue attributes to MoMs as we might need a different approach for that. For resources associated with jobs, it makes sense to go through send job route. What do you think?

Agreed! For the job related attribute, sending them with send_job sounds good. For the other, non-job, attributes/resources, we will need a separate mechanism. However, two things to confirm before u go ahead:

1. To be consistent in our design, the "m" flag would make sense on ALL kinds of attributes/resources? (we would not want to introduce a flag called "m" for one kind of attributes and that is called something else, or has a separate syntax for other attributes. Hence the design for the entire thing must be proposed and reviewed, hopefully in one place.

2. We have to be certain that we are not doing to double send the jobs related attributes if we create "another" mechanism to send the rest of the attributes. Thus, a bit of thought about what the other mechanism is would be good, so that we are sure about this approach before we go ahead and commit.

Does this make sense?

billnitzberg:
> Agreed! For the job related attribute, sending them with send_job sounds good. For the other, non-job, attributes/resources, we will need a separate mechanism. However, two things to confirm before u go ahead:

The 'm' flag makes sense for resources, and it's clear it fits with the current overall design of PBS Pro, in part, because there is already an existing interface to specify resource flags.   Anything beyond this should be in a separate PR.

Regarding 'm' for attributes, we need to be more careful, as there is no supported interface for changing attribute flags (outside the source code) today.  The design for that new interface (for modifying attribute flags outside of source code) should consider all the flags and all the types of attributes, not just the new 'm' one.   It makes sense to do this more carefully and then decide how much of that broader design should be implemented first, and all this should be a separate PR.

test/fw/ptl/utils/pbs_snaputils.py
```diff
@@ -632,6 +632,7 @@ def __convert_flag_to_numeric(flag):
         ATR_DFLAG_OPWR = 0x08
         ATR_DFLAG_MGRD = 0x10
         ATR_DFLAG_MGWR = 0x20
+        ATR_DFLAG_MOM = 0x400
```
dilip-krishnan: Just wondering why 0x400 and not 0x40,  it makes more sense to utilize bit in incremental order.

ashwathraop: @dilip-krishnan existing definitions match with what we have in pbs_internal.h. For example we have ATR_DFLAG_RASSN = 0x4000. I tried to follow it and defined the same value for ATR_DFLAG_MOM as defined in pbs_internal.h.

mike0042: @ashwathraop please add a comment to this effect. Also note that the value of ATR_DFLAG_RASSN is defined as 0x4000 and not 0x400 in pbs_internal.h.

ashwathraop: pbs_internal.h has value of ATR_DFLAG_RASSN is 0x4000 and ATR_DFLAG_MOM is 0x400. That is how I defined. I am not sure if I follow your comment.

Here is the list of flags we used in the snaputils file:
        ATR_DFLAG_USRD = 0x01
        ATR_DFLAG_USWR = 0x02
        ATR_DFLAG_OPRD = 0x04
        ATR_DFLAG_OPWR = 0x08
        ATR_DFLAG_MGRD = 0x10
        ATR_DFLAG_MGWR = 0x20
        ATR_DFLAG_MOM = 0x400
        ATR_DFLAG_RASSN = 0x4000
        ATR_DFLAG_ANASSN = 0x8000
        ATR_DFLAG_FNASSN = 0x10000
        ATR_DFLAG_CVTSLT = 0x20000

mike0042: Sorry @ashwathraop, I was looking at the wrong line. Your definition of ATR_DFLAG_MOM is correct. Please do add a comment indicating where the value originated.

ashwathraop:
Added new ptl test case for combination of i and m flags.
[ptl_logs.txt](https://github.com/PBSPro/pbspro/files/3272980/ptl_logs.txt)

test/fw/ptl/utils/pbs_snaputils.py
ashwathraop: @mike0042 I have added the comment to refer pbs_internal.h

bhroam:
@ashwathraop why did you disallow m and i?  The i flag means invisible to the user.  Hooks run as managers.  They should see the i flag resources.

ashwathraop:
@bhroam I thought with 'i' flag no external interface can see the resource. I was wrong. Even today mom hooks can access the resources flagged with 'i'. I tried a simple example hook and I could see the information being updated on the job.
----------------------------------------------------
create resource test
set resource test type = string
set resource test flag = i
03:06:04[root: /home/ashr/work]# cat end.py 
import pbs
this_event = pbs.event()
this_job = this_event.job
this_job.resources_used["test"] = "hellojob"
03:06:11[root: /home/ashr/work]# qstat -xf 1002|grep test
    resources_used.test = hellojob
----------------------------------------------------
It is just that any update made to the custom resources at server level or requested at job submission won't reach mom but the changes done at the hooks can be seen by the managers. I guess I do not see any harm in allowing flag 'm' to go along with 'i'. @subhasisb what do you think?

subhasisb:
> @bhroam I thought with 'i' flag no external interface can see the resource. I was wrong. Even today mom hooks can access the resources flagged with 'i'. I tried a simple example hook and I could see the information being updated on the job.
> Example:
> create resource test
> set resource test type = string
> set resource test flag = i
> 
> 03:06:04[root: /home/ashr/work]# cat end.py
> import pbs
> this_event = pbs.event()
> this_job = this_event.job
> this_job.resources_used["test"] = "hellojob"
> 
> 03:06:11[root: /home/ashr/work]# qstat -xf 1002|grep test
> resources_used.test = hellojob
> 
> Without 'm' flag today it is just that any update made to the custom resources at server level or requested at job submission won't reach mom but the changes done at the hooks can be seen by the managers. I guess I do not see any harm in allowing flag 'm' to go along with 'i'. @subhasisb what do you think?

After our offline discussion, i agree with Bhroam and you. And i see you have made the changes, so i am good with the PR.

ashwathraop:
[ptl_smoke.txt](https://github.com/PBSPro/pbspro/files/3279609/ptl_smoke.txt)
Results after reverting the commit.

ashwathraop:
@bhroam I reverted the changes. Can you look at it now?

test/tests/functional/pbs_resc_custom_perm.py
```diff
@@ -0,0 +1,164 @@
...
+# trademark licensing policies.
+
+from tests.functional import *
+
+
+class Test_Custom_Resource_Perm(TestFunctional):
+    rsc_list = ['foo_str', 'foo_strm', 'foo_strh']
+    hook_list = ["start", "begin"]
+
+    def tearDown(self):
```
bhroam: Everything you are reverting in tearDown() will be reverted by PTL.  I don't see the purpose of doing it here.

test/tests/functional/pbs_resc_custom_perm.py
```diff
@@ -0,0 +1,164 @@
...
+            self.server.manager(MGR_CMD_DELETE, HOOK, id=hook, runas=ROOT_USER)
+
+    def setUp(self):
+        TestFunctional.setUp(self)
+        for rsc in self.rsc_list:
+            try:
+                self.server.manager(MGR_CMD_DELETE, RSC,
+                                    id=rsc, runas=ROOT_USER)
+            except PbsManagerError:
+                pass
```
bhroam: Why are you deleting resources?  The system will be in the default state after TestFunctional.setUp() runs.

test/tests/functional/pbs_resc_custom_perm.py
```diff
@@ -0,0 +1,164 @@
...
+
+        self.momA = self.moms.values()[0]
+        self.momA.restart()
```
bhroam: Why are you restarting the mom?

test/tests/functional/pbs_resc_custom_perm.py
```diff
@@ -0,0 +1,164 @@
...
+        self.momA.delete_vnode_defs()
+        self.hostA = self.momA.shortname
+
+        rc = self.server.manager(MGR_CMD_DELETE, NODE, None, "")
+        self.assertEqual(rc, 0)
```
bhroam: manager() wil raise an exception if it fails.  There is no need to check its return code. (and other places)

test/tests/functional/pbs_resc_custom_perm.py
```diff
@@ -0,0 +1,164 @@
...
+        rc = self.server.manager(MGR_CMD_CREATE, NODE, id=self.hostA)
+        self.assertEqual(rc, 0)
+        self.server.expect(NODE, {'state': 'free'}, id=self.hostA)
+
+        # Next set custom resources with m flag and without m flag
+        # via qmgr -c 'create resource'
+        attr = {'type': 'string', 'flag': 'h'}
+        r = 'foo_str'
+        rc = self.server.manager(
+            MGR_CMD_CREATE, RSC, attr, id=r, runas=ROOT_USER, logerr=False)
```
bhroam: The person running PTL is a manager, you don't need to run as root.

test/tests/functional/pbs_resc_custom_perm.py
```diff
@@ -0,0 +1,164 @@
...
+        hook_name = "start"
+        a = {'event': "exechost_startup", 'enabled': 'True'}
+        rv = self.server.create_import_hook(
+            hook_name,
+            a,
+            startup_hook_body,
+            overwrite=True)
+        self.assertTrue(rv)
+
+        self.momA.restart()
```
bhroam: Why are you restarting the mom?  The hook will be sent to the mom after it is created.

test/tests/functional/pbs_resc_custom_perm.py
```diff
@@ -0,0 +1,164 @@
...
+        using a execjob_begin hook.
+        """
+        self.logger.info("test_custom_resc__single_node")
+
+        a = {'Resource_List.foo_str': 'str_noperm',
+             'Resource_List.foo_strm': 'str_perm',
+             'Resource_List.foo_strh': 'str_hook'
+             }
+        j = Job(TEST_USER)
+        j.set_attributes(a)
```
bhroam: Consider making this more concise by setting the attributes in the Job() call.

test/tests/functional/pbs_resc_custom_perm.py
```diff
@@ -0,0 +1,164 @@
...
+        j.set_sleep_time("100")
+        jid = self.server.submit(j)
+
+        self.server.expect(JOB, {
+            'job_state': 'R',
+            'Resource_List.foo_str': 'str_noperm',
+            'Resource_List.foo_strm': 'str_perm',
+            'Resource_List.foo_strh': 'str_hook'},
+            offset=1, attrop=PTL_AND, id=jid)
```
bhroam: attrop=PTL_AND is the default now.

test/tests/functional/pbs_resc_custom_perm.py
ashwathraop: Well, this restart is needed to run a startup hook on the MoM which creates a custom resource.

test/tests/functional/pbs_resc_custom_perm.py
ashwathraop: I will remove it.

test/tests/functional/pbs_resc_custom_perm.py
ashwathraop: Okay, makes sense. I will remove it.

test/tests/functional/pbs_resc_custom_perm.py
ashwathraop: Not exactly needed. I will remove it.

test/tests/functional/pbs_resc_custom_perm.py
ashwathraop: I will remove it.

test/tests/functional/pbs_resc_custom_perm.py
ashwathraop: Okay. I will remove the teardown().

test/tests/functional/pbs_resc_custom_perm.py
ashwathraop: Okay. I will remove it.

ashwathraop:
@bhroam Thank you for the comments on the ptl script. I have fixed them and here is the result from the latest run.
[ptl_log.txt](https://github.com/PBSPro/pbspro/files/3283179/ptl_log.txt)

test/tests/functional/pbs_resc_custom_perm.py
ashwathraop: Done.

test/tests/functional/pbs_resc_custom_perm.py
bhroam: Ahh, got it.  I didn't think about that.  Thanks for the explanation.

test/tests/functional/pbs_resc_custom_perm.py
```diff
@@ -79,15 +60,13 @@ def setUp(self):
...
+        self.server.manager(
+            MGR_CMD_CREATE, RSC, attr, id=r, logerr=False)
 
         attr = {'type': 'string', 'flag': 'hm'}
         r = 'foo_strm'
-        rc = self.server.manager(
-            MGR_CMD_CREATE, RSC, attr, id=r, runas=ROOT_USER, logerr=False)
-        self.assertEqual(rc, 0)
+        self.server.manager(
+            MGR_CMD_CREATE, RSC, attr, id=r, logerr=False)
```
bhroam: Not sure, but you might be able to merge these back to 1 line since they are shorter now.

ashwathraop: It is done.

ashwathraop:
I rebased and merged the commits as well. Once the checks are done PR can be merged.

PR #1067, opened 2019-04-01
Scheduler memory leak while parsing its general or fairshare configur…

suresh-thelkar:
…ation

#### Bug/feature Description
* Scheduler leaks huge amount of memory when the frequency of SIGHUP is more. Most of these leaks are related to the scheduler's general and fairshare configuration.

#### Affected Platform(s)
* All

* Whenever SIGHUP is issued Scheduler rereads its configuration present in sched_config file and then proceeds further with this new configuration. But before it rereads the new configuration it is not freeing most of the memory allocated for the conf(of type struct config used to read scheduler's configuration) data structure/s and hence the memory leak.

#### Solution Description
* As most of the leaks are related to the parsing of scheduler configuration and fairshare, started running various scenarios relating to these in Valgrind and proceeded towards fixing the “definitely lost” leaks.
* Well freeing of some of these leaks is not straight forward. This is because at some places where we parse the configuration in Scheduler we sometimes are using the read-only memory and at other times we are using dynamically allocated memory for example for conf.fairshare_res and conf.fairshare_ent. The solution here is to allocate memory dynamically all the times wherever required.
* Some of the data structures are pointing to the same memory location. If  we free one data structure then we can get core dump while freeing this memory from another data structure. As there is no reference counting mechanism we should make sure that whenever we free the memory we should not free the same memory again even though it is part of another data structure.

#### Testing logs/output
* Have run all the required automated test cases under valgrind before and after the fix. This is to make sure that the leaks identified are indeed fixed. Attached are the detailed logs. Valgrind logs after before and after testing are huge and since we cannot attach a tar.gz file not adding the same here. 
[test_fairshare_and_enhanced.log](https://github.com/PBSPro/pbspro/files/3027960/test_fairshare_and_enhanced.log)
[TestFairshareJobSortKeys_ext.log](https://github.com/PBSPro/pbspro/files/3027961/TestFairshareJobSortKeys_ext.log)
[TestMultipleSchedulers.log](https://github.com/PBSPro/pbspro/files/3027962/TestMultipleSchedulers.log)
[TestSchedulerInterface.log](https://github.com/PBSPro/pbspro/files/3027963/TestSchedulerInterface.log)

#### Checklist:

### DIFF
--- a/src/scheduler/fifo.c
+++ b/src/scheduler/fifo.c
@@ -562,11 +562,7 @@ schedule(int cmd, int sd, char *runjobid)
 		case SCH_CONFIGURE:
 			schdlog(PBSEVENT_SCHED, PBS_EVENTCLASS_SCHED, LOG_INFO,
 				"reconfigure", "Scheduler is reconfiguring");
-			free_fairshare_head(conf.fairshare);
 			reset_global_resource_ptrs();
-			free(conf.prime_sort);
-			free(conf.non_prime_sort);
-
 			if(schedinit() != 0) {
 				return 0;
 			}
--- a/src/scheduler/parse.c
+++ b/src/scheduler/parse.c
@@ -344,8 +344,11 @@ parse_config(char *fname)
 					} else
 						error = 1;
 				}
-				else if (!strcmp(config_name, PARSE_FAIRSHARE_RES))
+				else if (!strcmp(config_name, PARSE_FAIRSHARE_RES)) {
+					if (conf.fairshare_res != NULL)
+						free(conf.fairshare_res);
 					conf.fairshare_res = string_dup(config_value);
+				}
 				else if (!strcmp(config_name, PARSE_FAIRSHARE_ENT)) {
 					if (strcmp(config_value, ATTR_euser) &&
 						strcmp(config_value, ATTR_egroup) &&
@@ -356,6 +359,8 @@ parse_config(char *fname)
 						sprintf(errbuf, "%s %s is erroneous (or deprecated).",
 							PARSE_FAIRSHARE_ENT, config_value);
 					}
+					if (conf.fairshare_ent != NULL)
+						free(conf.fairshare_ent);
 					conf.fairshare_ent = string_dup(config_value);
 				}
 				else if (!strcmp(config_name, PARSE_NODE_GROUP_KEY)) {
@@ -438,64 +443,6 @@ parse_config(char *fname)
 								conf.non_prime_smp_dist = (enum smp_cluster_dist) smp_cluster_info[i].value;
 						}
 				}
-				else if (!strcmp(config_name, PARSE_SORT_BY)) {
-					obsolete[0] = config_name;
-					obsolete[1] = PARSE_JOB_SORT_KEY;
-
-					error = 1;
-					for (i = 0; sort_convert[i].config_name != NULL; i++) {
-						if (!strcmp(config_value, sort_convert[i].config_name)) {
-							error = 0;
-							if (prime == PRIME || prime == ALL) {
-								conf.prime_sort[0].res_name = sort_convert[i].res_name;
-								conf.prime_sort[0].order = sort_convert[i].order;
-								pkey_num++;
-							}
-							if (prime == NON_PRIME || prime == ALL) {
-								conf.non_prime_sort[0].res_name = sort_convert[i].res_name;
-								conf.non_prime_sort[0].order = sort_convert[i].order;
-								npkey_num++;
-							}
-						}
-					}
-					/* no_sort converts to not having any job_sort_key lines at all
-					 * but is not an erro
-					 */
-					if (!strcmp(config_value, "no_sort"))
-						error = 0;
-				}
-				else if (!strcmp(config_name, PARSE_KEY)) {
-					obsolete[0] = config_name;
-					obsolete[1] = PARSE_JOB_SORT_KEY;
-					error = 1;
-					for (i = 0; i < MAX_SORTS && sort_convert[i].config_name != NULL; i++) {
-						if (!strcmp(config_value, sort_convert[i].config_name)) {
-							error = 0;
-							if (((prime == PRIME||prime == ALL) && pkey_num <= MAX_SORTS) ||
-								((prime == NON_PRIME||prime == ALL) && npkey_num <= MAX_SORTS)
-								) {
-								if (prime == PRIME || prime == ALL) {
-									conf.prime_sort[pkey_num].res_name = sort_convert[i].res_name;
-									conf.prime_sort[pkey_num].order = sort_convert[i].order;
-									pkey_num++;
-								}
-								if (prime == NON_PRIME || prime == ALL) {
-									conf.non_prime_sort[npkey_num].res_name =
-										sort_convert[i].res_name;
-									conf.non_prime_sort[npkey_num].order =
-										sort_convert[i].order;
-									npkey_num++;
-								}
-							}
-							else {
-								sprintf(logbuf, "Too many keys for %s , ignoring %s.",
-									prime == PRIME ? "prime" : "non-prime", config_value);
-								schdlog(PBSEVENT_SCHED, PBS_EVENTCLASS_FILE, LOG_NOTICE,
-									fname, logbuf);
-							}
-						}
-					}
-				}
 				else if (!strcmp(config_name, PARSE_PREEMPT_PRIO)) {
 					obsolete[0] = PARSE_PREEMPT_PRIO;
 					obsolete[1] = "nothing - set via qmgr";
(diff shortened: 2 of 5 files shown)


src/scheduler/fifo.c
```diff
@@ -564,9 +566,64 @@ schedule(int cmd, int sd, char *runjobid)
 				"reconfigure", "Scheduler is reconfiguring");
 			free_fairshare_head(conf.fairshare);
 			reset_global_resource_ptrs();
+
+			if (conf.prime_sort != NULL) {
```
bhroam: Put this code in init_config().  The point of that function is to initialize conf and make it ready for parsing the config file.  Checking for members that need to be freed is a fine thing to do there.  You could even move the freeing of free_fairshare_head() in there as well.

src/scheduler/parse.c
```diff
@@ -356,6 +359,8 @@ parse_config(char *fname)
 						sprintf(errbuf, "%s %s is erroneous (or deprecated).",
 							PARSE_FAIRSHARE_ENT, config_value);
 					}
+					if (conf.fairshare_ent)
```
bhroam: Check against != NULL.

src/scheduler/fifo.c
```diff
@@ -564,9 +566,64 @@ schedule(int cmd, int sd, char *runjobid)
...
+			if (conf.non_prime_sort != NULL) {
+				for (i = 0; i < MAX_SORTS; i++) {
+					if (conf.non_prime_sort[i].res_name != NULL)
+						free(conf.non_prime_sort[i].res_name);
+				}
+			}
 			free(conf.non_prime_sort);
 
+			if (conf.prime_node_sort != NULL) {
+				for (i = 0; i < MAX_SORTS; i++) {
```
bhroam: Make a helper function for this loop.  You do it 4 times.

suresh-thelkar:
@bhroam, Thanks very much for the review. I have implemented your comments in the latest commit. Please look into it and provide further review comments if any.

src/scheduler/fifo.c
suresh-thelkar: Moved the code at the start of init_config.

src/scheduler/fifo.c
suresh-thelkar: Created a new function named free_sort_info() and used it appropriately.

src/scheduler/parse.c
```diff
@@ -1161,3 +1181,57 @@ preempt_cmp(const void *p1, const void *p2)
...
+			break;
+		default:
+			return;
+	}
+
+	if (so_info != NULL) {
+		int i;
+		for (i = 0; i < MAX_SORTS; i++) {
+			if (so_info[i].res_name != NULL) {
+				free(so_info[i].res_name);
```
bhroam: There is a very long since (like decades old) sort parameter called sort_by/key which used to be how we sorted jobs.  It takes static strings and converts them into the job_sort_key structure.  The code is still there to read sort_by/key.  You should probably remove the parsing code for them so it isn't possible for this to be a static string.

src/scheduler/parse.c
```diff
@@ -1161,3 +1181,57 @@ preempt_cmp(const void *p1, const void *p2)
...
+				if (si_type == PRIME_SORT || si_type == PRIME_NODE_SORT) {
+					if (si_type == PRIME_SORT) {
+						if (so_info[i].res_name != conf.non_prime_sort[i].res_name)
+							so_info[i].res_name = NULL;
+						else
+							so_info[i].res_name = conf.non_prime_sort[i].res_name = NULL;
```
bhroam: I'm not sure I understand this.  Wouldn't setting conf.non_prime_sort[i].res_name to NULL leak it?  Could you add a code comment explaining why this is necessary?

src/scheduler/parse.c
```diff
@@ -1161,3 +1181,57 @@ preempt_cmp(const void *p1, const void *p2)
...
+	}
+
+	if (so_info != NULL) {
+		int i;
+		for (i = 0; i < MAX_SORTS; i++) {
+			if (so_info[i].res_name != NULL) {
+				free(so_info[i].res_name);
+				if (si_type == PRIME_SORT || si_type == PRIME_NODE_SORT) {
+					if (si_type == PRIME_SORT) {
+						if (so_info[i].res_name != conf.non_prime_sort[i].res_name)
```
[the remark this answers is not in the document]
suresh-thelkar: Yes. But here we are just assigning it to NULL so that if same memory location is referred from an another sort_info structure we should not accidentally free it or in other words to avoid double free.

src/scheduler/parse.c
```diff
@@ -1161,3 +1181,57 @@ preempt_cmp(const void *p1, const void *p2)
...
+		default:
+			return;
+	}
+
+	if (so_info != NULL) {
+		int i;
+		for (i = 0; i < MAX_SORTS; i++) {
+			if (so_info[i].res_name != NULL) {
+				free(so_info[i].res_name);
+				if (si_type == PRIME_SORT || si_type == PRIME_NODE_SORT) {
```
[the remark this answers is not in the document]
suresh-thelkar: I have refactored the function free_sort_info() further. Please take a look.

src/scheduler/parse.c
suresh-thelkar: Thanks for letting us know this @bhroam. It looks like there are lot of obsolete parameters that we are passing in parse_config() apart from sort_by/key. I feel it would be better if we remove all these obsolete parameters as part of another ticket if we are no longer using them instead of only deleting the parsing of these two parameters.

src/scheduler/parse.c
suresh-thelkar: I have refactored the function free_sort_info() further. Please take a look. 

As there is a possibility that a memory location pointed by res_name can also be referred from another sort_info structures like non_prime_sort we should initialise the corresponding variable in the other sort_info structure to NULL to avoid double free or program crash. And also we need to make sure that free_sort_info() calls should work irrespective of the sequence in which it is called for PRIME_SORT, NON_PRIME_SORT, PRIME_NODE_SORT and NON_PRIME_NODE_SORT.

Example: 
	free_sort_info(PRIME_SORT);
	free_sort_info(NON_PRIME_SORT);
	free_sort_info(PRIME_NODE_SORT);
	free_sort_info(NON_PRIME_NODE_SORT);
OR
      free_sort_info(NON_PRIME_SORT);
      free_sort_info(PRIME_SORT);
      ...
Also I have added comments at the appropriate places in the code. I have made sure that I have run several test suites in valgrind to make sure we do not have any invalid reads/writes or any more leaks after the fix.

suresh-thelkar:
@bhroam, I have implemented your review comments and answered the questions. Can you please take a look and provide your further comments or sign off.

src/scheduler/parse.c
bhroam: Normally I would agree with you.  The reason I think we should remove sort_by and key now is that they will copy a static string into the res_name member.  You later expect this to be allocated memory so you free it.  If someone uses sort_by/key now, you will free static memory and crash.

src/scheduler/parse.c
bhroam: Would it just be easier to always allocate memory for the name?  It's such a small amount of memory, we're not saving much by reusing it.  It makes this area of the code more confusing to clean up, and adds the possibility for future bugs/double free()s.

src/scheduler/parse.c
suresh-thelkar: I am not copying a static string into the res_name member. Please look into line numbers 455 and 460 in parse.c. Instead calling sort_convert[i].res_name and then making use of it in the corresponding sort_info structure variables. So freeing this memory is safe and it does not lead to any crash.

suresh-thelkar:
@bhroam, I am trying to address your recent comments. I will update you once done.

src/scheduler/parse.c
bhroam: I see.  I don't think making a modification to sort_by/key is worthwhile.  I'd rather see it go away.  These have literally been deprecated for 1-2 decades.

suresh-thelkar: I have removed the code relating to sort_by/key. Please take a look.

src/scheduler/parse.c
suresh-thelkar: I have refactored free_sort_info() and also not using the read-only memory at the required place which avoids the need of adding complex logic to the free_sort_info() function. Please look into the latest commit for more details.

suresh-thelkar:
@bhroam, I have implemented your review comments. Please look at the last commit and give further review comments if any or sign off.

suresh-thelkar:
Thank you very much @bhroam for sign off. I have squashed and rebased. Can you please take it forward ?

bhroam:
@suresh-thelkar I don't see a second reviewer.  Did I miss a second sign off?

suresh-thelkar:
@bhroam, Thanks for reminding me about the second review. @sakshamgarg  has signed off. 
Have squashed and rebased after that. I request you to take this PR forward.

PR #1070, opened 2019-04-01
Node maintenance window enhancement

vchlum:
#### Bug/feature Description
* The new feature provides a special reservation that can be easily used for maintenance. The reservations can be submitted only on entire host(s) and they are "prioritized" over common reservations. This feature is only for managers and operators.

#### Affected Platform(s)
* All

* **[design document](https://pbspro.atlassian.net/wiki/spaces/PD/pages/879493121/Node+maintenance+window+enhancement)**
* **[pbspro community forum](http://community.pbspro.org/t/allowing-to-schedule-node-maintenance-with-a-possibility-to-run-new-jobs-until-the-maintenance-begins/1306)**

#### Solution Description
* The special reservation is submitted with extend parameter, which is acl restricted. Once the reservation is submitted, the command pbs_rsub immediately confirms the reservation. The confirmation process recognizes reservation prefix and detects all overlapping reservation. The overlapping reservations are degraded (other reservations are altered in order not to overlap the special reservation) and the scheduler will try to reconfirm them later. Already running reservations are not reconfirmed - just degraded.

#### Testing logs/output
* [ptl_reservations_output.txt](https://github.com/PBSPro/pbspro/files/3029400/ptl_reservations_output.txt)
* [ptl_maintenance_output.txt](https://github.com/PBSPro/pbspro/files/3029401/ptl_maintenance_output.txt)

#### Checklist:


test/tests/functional/pbs_maintenance_reservations.py
```diff
@@ -0,0 +1,192 @@
...
+    """
+    Various tests to verify behavior of maintenance reservations
+    """
+
+    def test_maintenance_acl_denied(self):
+        """
+        Test if the maintenance reservation is denied for common user
+        """
+        a = {'--hosts': self.mom.shortname}
+        r = Reservation(TEST_USER, attrs=a)
```
bhroam: You are making an assumption that the key to attrs can be an rsub option.  The interface says they're attributes.  Since rsub itself is doing a lot of work for this feature, I'd use run_cmd(). (and other places in the test plan)

test/tests/functional/pbs_maintenance_reservations.py
```diff
@@ -0,0 +1,192 @@
...
+        self.server.expect(RESV, exp_attr, id=rid1)
+
+        a2 = {'reserve_start': now + 3602,
+              'reserve_end': now + 3604,
+              '--hosts': self.mom.shortname}
+        r2 = Reservation(TEST_USER, attrs=a2)
+        self.server.submit(r2)
+
+        exp_attr = {'reserve_state': (MATCH_RE, 'RESV_DEGRADED|10'),
+                    'reserve_substate': 12}
```
bhroam: I thought maintenance reservations marked overlapping reservations in conflict instead of degraded.  Also check for the shortened resv_nodes

test/tests/functional/pbs_maintenance_reservations.py
```diff
@@ -0,0 +1,192 @@
...
+              'reserve_start': now + 3600,
+              'reserve_end': now + 3605}
+        r1 = Reservation(TEST_USER, attrs=a1)
+        rid1 = self.server.submit(r1)
+
+        exp_attr = {'reserve_state': (MATCH_RE, 'RESV_CONFIRMED|2')}
+        self.server.expect(RESV, exp_attr, id=rid1)
+
+        a2 = {'reserve_start': now + 3602,
+              'reserve_end': now + 3604,
```
bhroam: Could we use longer reservations than 5 seconds?  Since they're an hour out it shouldn't matter.  We've run into weird race conditions with short timeframes.  They shouldn't happen here, but it can't hurt to make these longer.

test/tests/functional/pbs_maintenance_reservations.py
```diff
@@ -0,0 +1,192 @@
...
+              '--hosts': self.momA.shortname}
+        r2 = Reservation(TEST_USER, attrs=a2)
+        self.server.submit(r2)
+
+        exp_attr = {'reserve_state': (MATCH_RE, 'RESV_CONFIRMED|2'),
+                    'resv_nodes': '(%s:ncpus=1)' % self.momB.shortname}
+        self.server.expect(RESV, exp_attr, id=rid1)
```
bhroam: Consider more tests (either some or all of these ideas).  Your maintenance reservation starts after the overlapping reservation and end before it.  Write one where the maintenance reservation starts before the overlapping reservation and ends during it.  Another where the maintenance reservation starts during the overlapping reservation and ends after it.  The third is where the maintenance reservation starts before and ends after the overlapping reservation.   Write a test where the two reservations overlap in nodes but not time and make sure the normal reservation is not marked degraded.  Write an end to end test.  Submit a maintenance reservation, see it overlap with another reservation, and make sure they work properly when they are running.  Write one where you see an overlapping reservation get reconfirmed after it is degraded.

src/cmds/pbs_rsub.c
```diff
@@ -696,8 +765,154 @@ main(int argc, char *argv[], char *envp[])
...
+
+		/* add place=exclhost */
+		if (set_resources(&attrib, "place=exclhost", 0, &erp) != 0) {
+			fprintf(stderr, "pbs_rsub: illegal -l value\n");
+
+			CS_close_app();
+			exit(pbs_errno);
+		}
+
+		extend = strdup("m");
```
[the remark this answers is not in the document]
vchlum: It is freed on line 958.

src/server/req_rescq.c
```diff
@@ -222,6 +222,165 @@ cnvrt_timer_init()
...
+				/* remove possible leading '+' */
+				if (presv->ri_wattr[(int)RESV_ATR_resv_nodes].at_val.at_str[0] == '+')
+					memmove(presv->ri_wattr[(int)RESV_ATR_resv_nodes].at_val.at_str,
+						presv->ri_wattr[(int)RESV_ATR_resv_nodes].at_val.at_str + 1,
+						strlen(presv->ri_wattr[(int)RESV_ATR_resv_nodes].at_val.at_str));
+
+				/* remove possible trailing '+' */
+				tmp_len = strlen(presv->ri_wattr[(int)RESV_ATR_resv_nodes].at_val.at_str);
+				if (presv->ri_wattr[(int)RESV_ATR_resv_nodes].at_val.at_str[tmp_len - 1] == '+')
+					presv->ri_wattr[(int)RESV_ATR_resv_nodes].at_val.at_str[tmp_len - 1] = '\0';
```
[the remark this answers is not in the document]
vchlum: This is after the modification of resv_nodes and it covers the cases: "+(node2:whatever)+(node3:whatever)" and "(node1:whatever)+(node2:whatever)+", but I realized that I forgot "(node1:whatever)++(node3:whatever)".

I also realized that "select=1:ncpus=2" and "select=2" produce different resv_nodes on the same multi-noded host. "select=1:ncpus=2" produces: "resv_nodes = (vn[0]:ncpus=1+vn[1]:ncpus=1)" but "select=2" produces: "resv_nodes = (vn[0]:ncpus=1)+(vn[1]:ncpus=1)". There was a bug in this.

I fixed the issues in the commit "bug fix in changing resv_nodes". A commit actually addressing your comments will come soon.

src/server/svr_jobfunc.c
```diff
@@ -4131,6 +4131,10 @@ change_enableORstart(resc_resv *presv, int which, char *value)
 	if (presv->ri_qs.ri_type != RESC_RESV_OBJECT)
 		return (0);
 
+	if (which == Q_CHNG_START && strcmp(value, ATR_TRUE) == 0 &&
+		! presv->ri_wattr[RESV_ATR_resv_nodes].at_flags & ATR_VFLAG_SET)
+		return (0);
+
```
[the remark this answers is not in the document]
vchlum: It means: Never start a reservation queue if its reservation is without RESV_ATR_resv_nodes.

If such a reservation queue is started (the associated RESV_ATR_resv_nodes is not set) then the associated jobs could run on any node!

It covers the case when the maintenance wipes the whole RESV_ATR_resv_nodes of common reservation and the reservation is degraded but it will start anyway.

test/tests/functional/pbs_maintenance_reservations.py
vchlum: The state is DEGRADED. Only the substate is IN_CONFLICT, which is the '12'. It is possible to set also the state to IN_CONFLICT (just for being more obvious), but it is not necessary for the feature. Should I set also the state to IN_CONFLICT?

src/server/req_rescq.c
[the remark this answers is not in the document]
bhroam: I'm confused how resv_nodes can contain a ++.  The case I was talking about was a chunk between two others (vn1:ncpus=1)+(vn2:ncpus=1)+(vn3:ncpus=1).  The middle chunk has a plus on both sides.  If you remove them both, the chunks will slide together.

src/server/svr_jobfunc.c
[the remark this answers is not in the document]
bhroam: Ahh got it.  Thanks for the explanation.

test/tests/functional/pbs_maintenance_reservations.py
bhroam: Oh, I thought degraded was a substate.  The reservation was in state confirmed or running, but the substate was degraded.  This would degrade the reservation.  I guess I'm not as familiar with the degraded reservation code as I thought.

src/server/req_rescq.c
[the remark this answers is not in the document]
vchlum: The modification of resv_nodes works like this: First, the string "(vn[n]:whatever)" is removed from resv_nodes but no '+' is removed yet - only the chunk itself. The extra '+' is removed afterwards.

So in your example, it means:
- For removing vn1 we get: "+(vn2:ncpus=1)+(vn3:ncpus=1)"
- For removing vn2 we get: "(vn1:ncpus=1)++(vn3:ncpus=1)"
- For removing vn3 we get: "(vn1:ncpus=1)+(vn2:ncpus=1)+"

After this, the extra '+' is removed:
- First, the resv_nodes leading '+' is checked and removed (not the chunk leading, but the whole resv_nodes leading - this is engaged when vn1 is removed). 
- Second, the resv_nodes trailing '+' is checked and removed (not the chunk trailing, but the whole resv_nodes trailing - this is engaged when vn3 is removed). 
- Third, the resv_nodes is searched for double '+' and if found, one of the plus is removed (this is engaged when vn2 is removed).

I will add some simple comment to the code.

src/cmds/pbs_rsub.c
```diff
@@ -696,8 +765,154 @@ main(int argc, char *argv[], char *envp[])
...
+			/* lets craft select_str */
+			for (bstat = bstat_head; bstat; bstat = bstat->next) {
+				for (pattr = bstat->attribs; pattr; pattr = pattr->next) {
+				    if (pattr->resource && strcmp(pattr->name, "resources_available") == 0 && strcmp(pattr->resource, "ncpus") == 0)
+					ncpus = pattr->value;
+				}
+
+				if (strcmp(bstat->name, *hostp) == 0 && atoi(ncpus) > 0) {
+					if (!select_str) {
+						select_str = (char*) malloc(strlen("select=host=") + strlen(bstat->name) + strlen(":ncpus=") +strlen(ncpus) + 1 + 1);
```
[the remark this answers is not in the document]
vchlum: OK, fixed on all occurrences.

src/cmds/pbs_rsub.c
```diff
@@ -710,9 +925,43 @@ main(int argc, char *argv[], char *envp[])
 			fprintf(stderr, "pbs_rsub: Error (%d) submitting reservation\n", pbs_errno);
 		CS_close_app();
 		exit(pbs_errno);
+	}
+
+	if (is_maintenance_resv) {
+		char *resv_start_time_str;
+		time_t resv_start_time = 0;
+
+		reservid = strtok(new_resvname, " ");
```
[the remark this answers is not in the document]
vchlum: Done.

src/scheduler/resv_info.c
```diff
@@ -736,7 +736,8 @@ query_resv(struct batch_status *resv, server_info *sinfo)
 	 */
 	if (advresv->resv->resv_state ==RESV_DEGRADED) {
```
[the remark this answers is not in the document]
vchlum: Done

src/server/req_rescq.c
```diff
@@ -222,6 +222,165 @@ cnvrt_timer_init()
...
+ */
+void
+remove_node_from_resv(resc_resv *presv, struct pbsnode *pnode) {
+	char *begin = NULL;
+	char *end = NULL;
+	char *tmp_buf;
+	struct	resvinfo *rinfp, *prev;
+	attribute   tmpatr;
+
+	tmp_buf = (char *)malloc(sizeof(char) * strlen(pnode->nd_name) + 2);
```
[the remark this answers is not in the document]
vchlum: OK, done.

src/server/req_rescq.c
```diff
@@ -222,6 +222,165 @@ cnvrt_timer_init()
...
+
+			while (*begin != '(')
+				begin--;
+
+			end = strchr(begin, ')');
+			end++;
+
+			if (presv->ri_giveback) {
+				/* resources were actually assigned to this reservation
+				 * we must accounted the resources back into the loaner's pool.
```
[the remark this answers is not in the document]
vchlum: Done

src/server/req_rescq.c
```diff
@@ -222,6 +222,165 @@ cnvrt_timer_init()
...
+		if ((begin = strstr(presv->ri_wattr[(int)RESV_ATR_resv_nodes].at_val.at_str, tmp_buf)) != NULL) {
+
+			while (*begin != '(')
+				begin--;
+
+			end = strchr(begin, ')');
+			end++;
+
+			if (presv->ri_giveback) {
+				/* resources were actually assigned to this reservation
```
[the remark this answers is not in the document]
vchlum: OK, it is shorter now.

src/server/req_rescq.c
```diff
@@ -222,6 +222,165 @@ cnvrt_timer_init()
...
+					presv->ri_wattr[(int)RESV_ATR_resv_nodes].at_val.at_str[tmp_len - 1] = '\0';
+			}
+
+			presv->ri_wattr[(int)RESV_ATR_resv_nodes].at_flags |= ATR_VFLAG_MODIFY | ATR_VFLAG_MODCACHE;
+			presv->ri_modified = 1;
+		}
+	}
+
+	/* traverse the reservations of the node and remove the reservation if found */
+	for (prev=NULL, rinfp = pnode->nd_resvp; rinfp; prev=rinfp, rinfp = rinfp->next) {
```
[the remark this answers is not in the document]
vchlum: Done

test/tests/functional/pbs_maintenance_reservations.py
```diff
@@ -0,0 +1,192 @@
...
+# organizations to create proprietary derivative works of PBS Pro and
+# distribute them - whether embedded or bundled with other software -
+# under a commercial license agreement.
+#
+# Use of Altair’s trademarks, including but not limited to "PBS™",
+# "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
+# trademark licensing policies.
+
+from tests.functional import *
+import time
```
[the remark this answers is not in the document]
vchlum: OK, thanks, removed.

test/tests/functional/pbs_maintenance_reservations.py
```diff
@@ -0,0 +1,192 @@
...
+        r = Reservation(TEST_USER, attrs=a)
+
+        msg = ""
+
+        try:
+            self.server.submit(r)
+        except PbsSubmitError as err:
+            msg = err.msg[0].strip()
+
+        self.assertTrue("pbs_rsub: Unauthorized Request" == msg)
```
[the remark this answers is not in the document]
vchlum: Changed across all tests.

test/tests/functional/pbs_maintenance_reservations.py
vchlum: It makes sense, done.

test/tests/functional/pbs_maintenance_reservations.py
vchlum: OK, I use much longer intervals now.

test/tests/functional/pbs_maintenance_reservations.py
vchlum: Thank you for suggestions. Several new tests have been added. I also added tests on top of suggested.

vchlum:
Latest test logs:
[ptl_reservations_output2.txt](https://github.com/PBSPro/pbspro/files/3043022/ptl_reservations_output2.txt)
[ptl_maintenance_output2.txt](https://github.com/PBSPro/pbspro/files/3043023/ptl_maintenance_output2.txt)

src/server/req_rescq.c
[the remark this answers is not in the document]
bhroam: Oh OK, I understand now.  Yes a code comment would be appreciated.

test/tests/functional/pbs_maintenance_reservations.py
```diff
@@ -172,21 +395,120 @@ def test_maintenance_reconfirm_resv(self):
...
+        maintenance reservation at first, and then do full overlap.
+        Test if the job in fully overlapped reservation will not run.
+        Two moms (-p "servers=M1,moms=M1:M2") are needed for this test.
+        """
+        now = int(time.time())
+
+        self.server.manager(MGR_CMD_SET, SERVER,
+                            {'managers': '%s@*' % TEST_USER})
+        self.server.manager(MGR_CMD_SET, SERVER,
+                            {'scheduler_iteration': 3})
```
bhroam: Curious, why lower the scheduler_iteration?  The server should kick cycles when needed.

test/tests/functional/pbs_maintenance_reservations.py
```diff
@@ -172,21 +395,120 @@ def test_maintenance_reconfirm_resv(self):
...
+
+        a = {'Resource_List.select': '1:ncpus=1',
+             'Resource_List.walltime': '1000',
+             'queue': rid1.split('.')[0]}
+        j = Job(TEST_USER, attrs=a)
+        j.set_sleep_time(990)
+        jid1 = self.server.submit(j)
+
+        self.logger.info("Wait for the job to try to run (30 seconds)")
+        time.sleep(30)
```
bhroam: Why are we waiting 30s for the job to run?  If the reservation is running, a cycle should be run when the job is submitted.

vchlum: The job is expected to stay in Q and do not run because its reservation was fully overlapped. There could be a false positive result without any timeout. Maybe 10 seconds is enough, but I think 30 is bearable.

test/tests/functional/pbs_maintenance_reservations.py
vchlum: The same occurrence above was unnecessary (removed). But it is needed in the last two tests where the reconfirmation is done. If the reservation is degraded by maintenance reservation then no scheduler cycle is kicked. It will wait for the regular cycle. AFAIS, this is the same behavior as if the vnode of confirmed reservation turns down. Reconfirmation will also wait for the regular cycle.

test/tests/functional/pbs_maintenance_reservations.py
```diff
@@ -36,60 +36,81 @@
...
         Test if the select and place is forbidden with maintenance
         """
-        a = {'Resource_List.select': '1:ncpus=1',
-             '--hosts': self.mom.shortname}
-        r = Reservation(TEST_USER, attrs=a)
+        start = self.now_plus_to_rsub_format(hours=1)
+        end = self.now_plus_to_rsub_format(hours=2)
 
-        msg = ""
+        ret = self.submit_maintenance(TEST_USER, start, end,
```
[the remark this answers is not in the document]
vchlum: I am sorry, I am not sure what do you mean by "have one reason the reservation can't be submitted". Please, could you clarify? Do you mean (for example) to always test whether the return code is not zero?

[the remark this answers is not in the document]
bhroam: The test can also be rejected for permissions reasons.  Could you add TEST_USER to the managers line like you do in other tests?  This means the only reason the reservation can be rejected is due to -lselect and --hosts.

test/tests/functional/pbs_maintenance_reservations.py
vchlum: Actually, only one occurrence is necessary. The rest is removed.

test/tests/functional/pbs_maintenance_reservations.py
```diff
@@ -135,27 +220,165 @@ def test_maintenance_degrade_reservation(self):
...
+                    'reserve_substate': 2,
+                    'resv_nodes': (MATCH_RE, resv_nodes_re),
+                    'Resource_List.select': (MATCH_RE, select_re),
+                    'Resource_List.place': 'exclhost'}
+        self.server.expect(RESV, exp_attr, id=rid)
+
+    def test_maintenance_reconfirm_resvation_and_run(self):
+        """
+        Test if the overlapping reservation is reconfirmed correctly.
+        Wait for both reservation to start and submit jobs into this
```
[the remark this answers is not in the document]
vchlum: Yes, fixed.

test/tests/functional/pbs_maintenance_reservations.py
```diff
@@ -135,27 +220,165 @@ def test_maintenance_degrade_reservation(self):
...
+        Two moms (-p "servers=M1,moms=M1:M2") are needed for this test.
+        """
+        self.server.manager(MGR_CMD_SET, SERVER,
+                            {'managers': '%s@*' % TEST_USER})
+        self.server.manager(MGR_CMD_SET, SERVER,
+                            {'scheduler_iteration': 3})
+
+        if len(self.moms) != 2:
+            cmt = "need 2 mom hosts: -p servers=<m1>,moms=<m1>:<m2>"
+            self.skip_test(reason=cmt)
```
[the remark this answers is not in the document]
vchlum: Done

test/tests/functional/pbs_maintenance_reservations.py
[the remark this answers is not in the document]
vchlum: Thank you for clarification, done.

test/tests/functional/pbs_maintenance_reservations.py
```diff
@@ -36,60 +36,81 @@
...
+        now_plus = datetime.datetime.today() + interval
+        return now_plus.strftime("%Y%m%d%H%M")
+
+    def submit_maintenance(self, user, start, end, hosts, others=""):
+        """
+        Submit maintenance reservation
+        """
+        resv_cmd = "%s -R %s -E %s %s --hosts %s" % \
+                   (self.rsub_cmd, start, end, others, " ".join(hosts))
+        cmd = "sudo -u %s %s" % (user, resv_cmd)
```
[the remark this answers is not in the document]
vchlum: Fixed.

test/tests/functional/pbs_maintenance_reservations.py
```diff
@@ -36,60 +36,81 @@
...
+    def setUp(self):
+        """
+        Tests setup
+        """
+        TestFunctional.setUp(self)
+
+        self.rsub_cmd = os.path.join(self.server.pbs_conf[
+            'PBS_EXEC'], 'bin', 'pbs_rsub')
+
+    def now_plus_to_rsub_format(self, hours=0, minutes=0):
```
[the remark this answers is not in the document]
vchlum: Fixed

src/cmds/pbs_rsub.c
```diff
@@ -696,8 +765,158 @@ main(int argc, char *argv[], char *envp[])
...
+						free(tmp_str);
+					}
+				}
+			}
+		}
+
+		pbs_statfree(bstat_head);	/* free info returned by pbs_stathost() */
+
+		pbs_errno = 0;
+		bstat_head = pbs_statvnode(connect, "", NULL, NULL);
```
prakashcv13: instead of using pbs_statvnode(), we can use pbs_stathost(). We can make the code simpler instead of re-inventing the wheel. What do you think?

vchlum: pbs_stathost() provides only list of hosts but we need the vnodes and its attributes because of crafting execvnodes. So, pbs_statvnode() is necessary.

I also use pbs_stathost() because we need to craft the select in the form like: "host=<host1>:ncpus=<ncpus_host1>...".

Maybe, we could use pbs_statvnode() and count the ncpus on all vnodes of the same host, but I think it would be more confusing than the used solution, and so it is better to call both pbs_stathost() and pbs_statvnode(). 

@prakashcv13 Do you think I should use only pbs_statvnode() and count the ncpus of the requested host from pbs_statvnode() (based on the same "host" attribute)?

src/server/req_rescq.c
```diff
@@ -222,6 +222,199 @@ cnvrt_timer_init()
...
+ * @parm[in,out]	pnode	-	pointer to node
+ *
+ */
+void
+remove_node_from_resv(resc_resv *presv, struct pbsnode *pnode) {
+	char *begin = NULL;
+	char *end = NULL;
+	char *tmp_buf;
+	struct	resvinfo *rinfp, *prev;
+	attribute   tmpatr;
```
prakashcv13: extra space after struct and attribute

src/server/req_rescq.c
```diff
@@ -222,6 +222,199 @@ cnvrt_timer_init()
...
+ *		the node is removed from RESV_ATR_resv_nodes and assigned
+ *		resources are accounted back to loaner's pool and finally the
+ *		reservation is removed from the node
+ *
+ * @parm[in,out]	presv	-	reservation structure
+ * @parm[in,out]	pnode	-	pointer to node
+ *
+ */
+void
+remove_node_from_resv(resc_resv *presv, struct pbsnode *pnode) {
```
prakashcv13: '{' should be on the first column of the next line. Similar change is also needed in the newly added functions.

src/server/req_rescq.c
```diff
@@ -222,6 +222,199 @@ cnvrt_timer_init()
...
+	tmp_buf = (char *)malloc(strlen(pnode->nd_name) + 2);
+	if (tmp_buf == NULL) {
+		snprintf(log_buffer, LOG_BUF_SIZE-1, "malloc failure (errno %d)", errno);
+			log_err(PBSE_SYSTEM, __func__, log_buffer);
+		return;
+	}
+
+	strcpy(tmp_buf, pnode->nd_name);
+	tmp_buf[strlen(pnode->nd_name)] = ':';
+	tmp_buf[strlen(pnode->nd_name) + 1] = '\0';
```
prakashcv13: we can use a single sprintf() call instead of the above 3 statements.

src/cmds/pbs_rsub.c
```diff
@@ -696,8 +765,158 @@ main(int argc, char *argv[], char *envp[])
...
+						}
+						sprintf(select_str, "select=host=%s:ncpus=%s", bstat->name, ncpus);
+					} else {
+						tmp_str = strdup(select_str);
+						if (tmp_str == NULL) {
+							fprintf(stderr, "pbs_rsub: Out of memory\n");
+							exit(2);
+						}
+						/* +14 stands for "+host=" + ":ncpus=" + '\0' */
+						select_str = (char*) realloc(select_str, strlen(tmp_str) + strlen(bstat->name) + strlen(ncpus) + 14);
```
prakashcv13: I would suggest we initially malloc() a buffer of a fixed length say 500, and then use pbs_strcat() to concatenate subsequent values, instead of realloc()ing every time. This will save us from the calculations.

src/cmds/pbs_rsub.c
```diff
@@ -696,8 +765,158 @@ main(int argc, char *argv[], char *envp[])
...
+
+		/* here, the select is crafted from hosts */
+		hostp = maintenance_hosts;
+		for (;*hostp ;hostp++) {
+			/* lets craft select_str */
+			for (bstat = bstat_head; bstat; bstat = bstat->next) {
+				for (pattr = bstat->attribs; pattr; pattr = pattr->next) {
+				    if (pattr->resource && strcmp(pattr->name, "resources_available") == 0 && strcmp(pattr->resource, "ncpus") == 0)
+					ncpus = pattr->value;
+				}
```
prakashcv13: wouldn't moving this for loop under "if (strcmp(bstat->name, *hostp) == 0)" would be better?

src/cmds/pbs_rsub.c
```diff
@@ -696,8 +765,158 @@ main(int argc, char *argv[], char *envp[])
...
+			exit(pbs_errno);
+		}
+
+		/* here, the select is crafted from hosts */
+		hostp = maintenance_hosts;
+		for (;*hostp ;hostp++) {
+			/* lets craft select_str */
+			for (bstat = bstat_head; bstat; bstat = bstat->next) {
+				for (pattr = bstat->attribs; pattr; pattr = pattr->next) {
+				    if (pattr->resource && strcmp(pattr->name, "resources_available") == 0 && strcmp(pattr->resource, "ncpus") == 0)
```
prakashcv13: Let us use "ATTR_rescavail" macro instead of the string literal.

src/cmds/pbs_rsub.c
prakashcv13: @vchlum - I should have spent more time in understanding why we are calling pbs_statvnode(). If we can avoid call to the pbs_stathost() and make the code better, we should do it. Could you please let me know why do you think using only pbs_statvnode() would be more confusing.

src/cmds/pbs_rsub.c
```diff
@@ -696,8 +765,158 @@ main(int argc, char *argv[], char *envp[])
...
+
+		/* add place=exclhost */
+		if (set_resources(&attrib, "place=exclhost", 0, &erp) != 0) {
+			fprintf(stderr, "pbs_rsub: illegal -l value\n");
+
+			CS_close_app();
+			exit(pbs_errno);
+		}
+
+		extend = strdup("m");
```
prakashcv13: instead of calling strdup() and checking for it's success - let us use char *extend = "s".

src/cmds/pbs_rsub.c
vchlum: @prakashcv13 I thought this way there will be two similar blocks of code with easier reading, but I thought about that and you are right. I will remove the calling of pbs_stathost(). It is not a problem. We should avoid calling server if possible.

src/cmds/pbs_rsub.c
vchlum: You were right, but this optimization is not feasible with the new implementation (pbs_stathost() is removed).

src/cmds/pbs_rsub.c
vchlum: OK, strcat() is used now.

src/cmds/pbs_rsub.c
vchlum: Done

src/cmds/pbs_rsub.c
vchlum: Definitely, good suggestion, done.

src/server/req_rescq.c
vchlum: OK, done.

src/server/req_rescq.c
vchlum: Done.

src/server/req_rescq.c
vchlum: Done.

src/cmds/pbs_rsub.c
```diff
@@ -696,8 +765,158 @@ main(int argc, char *argv[], char *envp[])
...
+				fprintf(stderr, "pbs_rsub: No nodes found\n");
+			}
+
+			CS_close_app();
+			exit(pbs_errno);
+		}
+
+		/* here, the select is crafted from hosts */
+		hostp = maintenance_hosts;
+		for (;*hostp ;hostp++) {
```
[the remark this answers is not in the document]
vchlum: Fixed on all locations.

src/cmds/pbs_rsub.c
```diff
@@ -268,6 +277,32 @@ process_opts(int argc, char **argv, struct attrl **attrp, char *dest)
 		errflg++;
 	}
 
+	if (is_maintenance_resv) {
+		if (argc - optind > 0) {
+			maintenance_hosts = (char **)malloc(sizeof(char *) * (argc - optind + 1));
```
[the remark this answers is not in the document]
vchlum: OK, done.

vchlum:
@prakashcv13 I addressed your comments. New ptl logs follow:
[ptl_reservations_output3.txt](https://github.com/PBSPro/pbspro/files/3137311/ptl_reservations_output3.txt)
[ptl_maintenance_output3.txt](https://github.com/PBSPro/pbspro/files/3137312/ptl_maintenance_output3.txt)

test/tests/functional/pbs_maintenance_reservations.py
```diff
@@ -0,0 +1,513 @@
...
+        select_re = "%s|%s" % (possibility1, possibility2)
+
+        exp_attr = {'reserve_state': (MATCH_RE, 'RESV_CONFIRMED|2'),
+                    'reserve_substate': 2,
+                    'resv_nodes': (MATCH_RE, resv_nodes_re),
+                    'Resource_List.select': (MATCH_RE, select_re),
+                    'Resource_List.place': 'exclhost'}
+        self.server.expect(RESV, exp_attr, id=rid)
+
+    def test_maintenance_reconfirm_resvation_and_run(self):
```
prakashcv13: resvation ==> reservation

src/cmds/pbs_rsub.c
```diff
@@ -696,8 +767,131 @@ main(int argc, char *argv[], char *envp[])
...
+			if (host_ncpus > 0) {
+				if (!select_str) {
+					select_str = malloc(BUF_SIZE);
+					if (select_str == NULL) {
+						fprintf(stderr, "pbs_rsub: Out of memory\n");
+						exit(2);
+					}
+					select_str_size = BUF_SIZE;
+
+					sprintf(select_str, "select=host=%s:ncpus=%d", *hostp, host_ncpus);
```
prakashcv13: shall we use snprintf() instead of sprintf(), here and in other places?

prakashcv13:
Hi @vchlum - Thank you so much for making the changes. I have a few more comments.

src/cmds/pbs_rsub.c
```diff
@@ -696,8 +767,131 @@ main(int argc, char *argv[], char *envp[])
 		}
 	}
 
+	if (is_maintenance_resv) {
+		int i;
+		char tmp_str[BUF_SIZE];
```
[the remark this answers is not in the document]
vchlum: I think 512 is acceptable. The worst case is 255 (for PBS_MAXHOSTNAME) + 10 (for max integer) + 13 (for "+host=:ncpus="), which is 278. Let me know if you still consider it to be wasting.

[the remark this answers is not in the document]
prakashcv13: hmmm. thank you for the explanation. let us keep it 512 bytes.

src/cmds/pbs_rsub.c
```diff
@@ -696,8 +767,131 @@ main(int argc, char *argv[], char *envp[])
...
+						fprintf(stderr, "pbs_rsub: Out of memory\n");
+						exit(2);
+					}
+				}
+			}
+		}
+
+		/* here, the execvnodes is crafted from vnodes */
+		hostp = maintenance_hosts;
+		for (; *hostp; hostp++) {
```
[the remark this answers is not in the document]
vchlum: @prakashcv13 If I understand correctly, you have suggested combining the crafting of 'select' and 'execvnodes' into one loop. It is possible but IMHO it will harm the readability pretty badly, and it will not improve the complexity. I would prefer keeping the crafting of 'select' and 'execvnodes' separately.

src/cmds/pbs_rsub.c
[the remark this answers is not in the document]
mike0042: The pbs_asprintf() function was introduced when we added "-Wall -Werror" to the compilation parameters. That might be another option in this case. Several buffer overrun issues were addressed with this.

src/cmds/pbs_rsub.c
```diff
@@ -788,98 +791,74 @@ main(int argc, char *argv[], char *envp[])
...
-						select_str = (char*) realloc(select_str, strlen(tmp_str) + strlen(bstat->name) + strlen(ncpus) + 14);
-						if (select_str == NULL) {
-							fprintf(stderr, "pbs_rsub: Out of memory\n");
-							exit(2);
-						}
-						sprintf(select_str, "%s+host=%s:ncpus=%s", tmp_str, bstat->name, ncpus);
-						free(tmp_str);
-					}
+				if (strcmp(host, *hostp) == 0 && atoi(ncpus) > 0) {
+					host_ncpus += atoi(ncpus);
```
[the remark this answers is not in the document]
vchlum: strtol() is used now.

src/cmds/pbs_rsub.c
vchlum: snprintf() is used in all places now.

test/tests/functional/pbs_maintenance_reservations.py
vchlum: The is typo fixed.

test/tests/functional/pbs_maintenance_reservations.py
```diff
@@ -0,0 +1,513 @@
...
+        return now_plus.strftime("%Y%m%d%H%M")
+
+    def submit_maintenance(self, user, start, end, hosts, others=""):
+        """
+        Submit maintenance reservation
+        """
+        cmd = "%s -R %s -E %s %s --hosts %s" % \
+              (self.rsub_cmd, start, end, others, " ".join(hosts))
+
+        return self.du.run_cmd(self.server.hostname, cmd, runas=user)
```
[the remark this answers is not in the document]
vchlum: PTL framework updated. I have added 'hosts' to the class Reservation().

test/tests/functional/pbs_maintenance_reservations.py
```diff
@@ -0,0 +1,513 @@
...
+    """
+
+    def setUp(self):
+        """
+        Tests setup
+        """
+        TestFunctional.setUp(self)
+
+        self.rsub_cmd = os.path.join(self.server.pbs_conf[
+            'PBS_EXEC'], 'bin', 'pbs_rsub')
```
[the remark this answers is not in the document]
vchlum: OK, changed along with changing PTL framework.

src/server/req_rescq.c
```diff
@@ -222,6 +222,198 @@ cnvrt_timer_init()
...
+	attribute tmpatr;
+
+	tmp_buf = (char *)malloc(strlen(pnode->nd_name) + 2);
+	if (tmp_buf == NULL) {
+		snprintf(log_buffer, LOG_BUF_SIZE-1, "malloc failure (errno %d)", errno);
+			log_err(PBSE_SYSTEM, __func__, log_buffer);
+		return;
+	}
+
+	sprintf(tmp_buf, "%s:", pnode->nd_name);
```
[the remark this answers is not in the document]
vchlum: It is used now.

src/server/req_rescq.c
```diff
@@ -222,6 +222,198 @@ cnvrt_timer_init()
...
+void
+remove_node_from_resv(resc_resv *presv, struct pbsnode *pnode)
+{
+	char *begin = NULL;
+	char *end = NULL;
+	char *tmp_buf;
+	struct resvinfo *rinfp, *prev;
+	attribute tmpatr;
+
+	tmp_buf = (char *)malloc(strlen(pnode->nd_name) + 2);
```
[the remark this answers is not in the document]
vchlum: Done

src/server/req_rescq.c
```diff
@@ -222,6 +222,198 @@ cnvrt_timer_init()
...
+	if (tmp_buf == NULL) {
+		snprintf(log_buffer, LOG_BUF_SIZE-1, "malloc failure (errno %d)", errno);
```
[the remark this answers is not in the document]
vchlum: OK

src/server/req_quejob.c
```diff
@@ -2356,6 +2365,12 @@ req_resvSub(struct batch_request *preq)
 					ATR_VFLAG_SET | ATR_VFLAG_MODCACHE;
 		}
 	}
+
+	if (is_maintenance) {
+		rid[0] = PBS_MNTNC_RESV_ID_CHAR;
+		qbuf[0] = PBS_MNTNC_RESV_ID_CHAR;
```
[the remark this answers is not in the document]
vchlum: Done

src/cmds/pbs_rsub.c
```diff
@@ -696,8 +767,131 @@ main(int argc, char *argv[], char *envp[])
...
+
+				for (pattr = bstat->attribs; pattr; pattr = pattr->next) {
+					if (pattr->resource && strcmp(pattr->resource, "host") == 0)
+						host = pattr->value;
+
+					if (pattr->resource && strcmp(pattr->name, ATTR_rescavail) == 0 && strcmp(pattr->resource, "ncpus") == 0)
+						ncpus = pattr->value;
+				}
+
+				if (strcmp(host, *hostp) == 0 && atoi(ncpus) > 0){
```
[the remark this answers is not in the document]
vchlum: Done

src/cmds/pbs_rsub.c
```diff
@@ -696,8 +767,131 @@ main(int argc, char *argv[], char *envp[])
...
+			for (bstat = bstat_head; bstat; bstat = bstat->next) {
+				for (pattr = bstat->attribs; pattr; pattr = pattr->next) {
+					if (pattr->resource && strcmp(pattr->name, ATTR_rescavail) == 0 && strcmp(pattr->resource, "host") == 0)
+						host = pattr->value;
+					if (pattr->resource && strcmp(pattr->name, ATTR_rescavail) == 0 && strcmp(pattr->resource, "ncpus") == 0)
+						ncpus = pattr->value;
+				}
+
+				if (strcmp(host, *hostp) == 0 && atoi(ncpus) > 0) {
+					host_ncpus += atoi(ncpus);
```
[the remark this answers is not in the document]
vchlum: Removed.

src/cmds/pbs_rsub.c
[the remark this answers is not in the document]
prakashcv13: @vchlum - I would vouch for performance instead of readability. What do you think?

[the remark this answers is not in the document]
vchlum: This change is not included in the last commit yet. Let me know if you really prefer to combine the two for loops into one.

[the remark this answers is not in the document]
vchlum: @prakashcv13 Well, I think the performance gain will be quite small, but no problem. OK, I will change it into one for loop in next commit. I will add some comments if necessary.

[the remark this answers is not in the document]
vchlum: Done

vchlum:
Latest PTL log: 
[ptl_maintenance_output4.txt](https://github.com/PBSPro/pbspro/files/3142546/ptl_maintenance_output4.txt)

src/cmds/pbs_rsub.c
```diff
@@ -839,24 +848,33 @@ main(int argc, char *argv[], char *envp[])
...
+							CS_close_app();
 							exit(2);
 						}
 						execvnodes_str_size = BUF_SIZE;
 
-						sprintf(execvnodes_str, "(%s:ncpus=%s)", bstat->name, ncpus);
+						snprintf(execvnodes_str, BUF_SIZE - 1, "(%s:ncpus=%d)", bstat->name, ncpus);
 					} else {
-						sprintf(tmp_str, "+(%s:ncpus=%s)", bstat->name, ncpus);
+						snprintf(tmp_str, BUF_SIZE - 1, "+(%s:ncpus=%d)", bstat->name, ncpus);
```
bhroam: You don't need the -1.  snprintf() will only print N-1 bytes to leave space for the null byte.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -13872,10 +13874,11 @@ class Reservation(ResourceResv):
 
     dflt_attributes = {}
 
-    def __init__(self, username=None, attrs={}):
+    def __init__(self, username=None, attrs={}, hosts=[]):
```
bhroam: It's dangerous to use mutable types as default arguments.  It means every call to this function will use the same memory.  If the list is modified in one call, it will still be modified in subsequent call.  It's safer to set it to None and then say if hosts is None: hosts = [].  I see that attrs in the same way, but it's also wrong.  Fix that one as well?

src/cmds/pbs_rsub.c
vchlum: Right, fixed.

test/fw/ptl/lib/pbs_testlib.py
vchlum: Both the attrs and hosts fixed.

test/tests/functional/pbs_maintenance_reservations.py
```diff
@@ -44,74 +44,62 @@ class TestMaintenanceReservations(TestFunctional):
...
 
-        ret = self.submit_maintenance(TEST_USER, start, end,
-                                      [self.mom.shortname])
+        msg = ""
 
-        msg = ret['err'][0].strip()
+        try:
+            self.server.submit(r)
+        except PbsSubmitError as err:
+            msg = err.msg[0].strip()
```
bhroam: Since maintenance reservations are handled in the pbs_rsub command, could you force PTL into CLI mode in setUp()?  This way if API mode is ever fixed, then this test won't fail.

src/cmds/pbs_rsub.c
```diff
@@ -811,10 +810,35 @@ main(int argc, char *argv[], char *envp[])
 					exit(2);
 				}
 
-				if (strcmp(host, *hostp) == 0 && ncpus > 0)
+				/* here, the execvnodes is crafted */
+				if (strcmp(host, *hostp) == 0 && ncpus > 0) {
+					/* count ncpus of a host across vnodes
+					 * it will be used for crafting select */
```
bhroam: Please move the */ to the line below

vchlum: Done

test/tests/functional/pbs_maintenance_reservations.py
vchlum: Done

prakashcv13:
I have restarted debian and SUSE builds in travis as these builds had errored

prakashcv13:
@vchlum - there is one comment from @mike0042 - that would need a reply or code changes. Could you please look into it. Thank you for your patience.

src/cmds/pbs_rsub.c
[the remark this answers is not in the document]
vchlum: pbs_asprintf() always allocates the memory.  Here, we repeatedly snprintf to the tmp_str in the for loop. Using pbs_asprintf() instead of snprintf, we would need to free the tmp_str after each use and it would be allocated with pbs_asprintf() again. Now, calculations are spared opposed to using pbs_asprintf(). I think that the string buffer that is large enough is more suitable in this case.

vchlum:
@prakashcv13 thx for the reminder. I replied to @mike0042. The debian build has succeeded but the suse build timeouts on retreiving the repository:-( Should I rebase?

src/cmds/pbs_rsub.c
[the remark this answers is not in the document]
mike0042: Sounds reasonable. Thank you.

prakashcv13:
@vchlum - travis seems to be having an issue. Let us squash and rebase.

src/cmds/pbs_rsub.c
```diff
@@ -100,7 +102,7 @@ process_opts(int argc, char **argv, struct attrl **attrp, char *dest)
 	char* temp_apvalue = NULL;
 #endif
 
-	while ((c = getopt(argc, argv, "D:E:I:l:m:M:N:q:r:R:u:U:g:G:H:W:")) != EOF) {
+	while ((c = getopt(argc, argv, "D:E:I:l:m:M:N:q:r:R:u:U:g:G:H:W:-:")) != EOF) {
```
mike0042: Shouldn't you be using getopt_long() for this? If you use getopt() then the following are equivalent:
pbs_rsub --hosts
pbs_rsub -- hosts

src/cmds/pbs_rsub.c
```diff
@@ -256,6 +258,13 @@ process_opts(int argc, char **argv, struct attrl **attrp, char *dest)
 				}
 				break;
 
+			case '-':
+				if (strcmp(optarg, "hosts") == 0)
+					is_maintenance_resv = 1;
```
mike0042: Will this impact all vnodes assigned to the list of hosts?

src/cmds/pbs_rsub.c
vchlum: 'pbs_rsub -- hosts' returns directive error. OK, I will change it to getopt_long() anyway.

mike0042: This will be the first time we've added a "long" parameter in PBS Pro. I suspect we will start using them more often, so thank you for making the change.

bhroam: You can't use getopt_long().  That won't work on Windows.

src/cmds/pbs_rsub.c
vchlum: The vnodes are obtained later with pbs_statvnode(). For each host passed as a parameter (here), the associated vnodes are found and used in crafting execvnodes.

src/cmds/pbs_rsub.c
vchlum: Do we need to implement getopt_long() for Windows or will we use the current solution?

prakashcv13:
I restarted the SUSE build and travis is green now.

src/cmds/pbs_rsub.c
prakashcv13: As per comments from @mike0042, we need not implement getopt_long(). But, have we tested these changes on Windows?

vchlum: No, I do not have any Windows machine. I probably could prepare some Windows testbed if necessary but it will take some time. I would appreciate help with the test on Windows.

mike0042: I don't want to force all this extra work on you @vchlum. I suggest we file a ticket to implement getopt_long() for Windows and refactor the existing getopt() calls in the future.

bhroam:
Thanks for the nice new feature, @vchlum.  I merged in the PR.

PR #1077, opened 2019-04-04
Tunable test parameters

visheshh:
#### Describe Bug or Feature
Tunable test parameters

#### Describe Your Change
@testparams is a decorator to specify default values of test specific parameters
Test Parameters can be specified using testparams decorator in key value pair format. The parameters and its values will be stored in self.conf dictionary of the test case.
Key's must be in the format TestSuiteName.VariableName=Value
As the decorator stores the key value pair data in self.conf dictionary it can be modified through pbs_benchpress.
The test parameters will be added to test doc string and can be viewed by pbs_benchpress -i --verbose -t <pbstestsuite>

#### Link to Design Doc

https://pbspro.atlassian.net/wiki/spaces/PD/pages/1198981123/Tunable+test+parameters

Modified tests with params
1. scheduling false , 60 jobs
[j60sfalse.txt](https://github.com/PBSPro/pbspro/files/3042313/j60sfalse.txt)
2. scheduling True 50 jobs
[j50strue.txt](https://github.com/PBSPro/pbspro/files/3042314/j50strue.txt)

### DIFF
--- a/test/fw/ptl/utils/pbs_testsuite.py
+++ b/test/fw/ptl/utils/pbs_testsuite.py
@@ -47,6 +47,7 @@ import socket
 import time
 import calendar
 import ptl
+from distutils.util import strtobool
 from ptl.utils.pbs_logutils import PBSLogAnalyzer
 from ptl.utils.pbs_dshutils import DshUtils
 from ptl.utils.pbs_cliutils import CliUtils
@@ -249,6 +250,40 @@ def requirements(*args, **kwargs):
     return wrap_obj
 
 
+def testparams(**kwargs):
+    """
+    Decorator to set or modify test specific parameters
+    """
+    def decorated(function):
+        function.__doc__ += "Test Params:" + "\n\t"
+        for key, value in kwargs.iteritems():
+            function.__doc__ += str(key) + ' : ' + str(value) + '\n\t'
+
+        def wrapper(self, *args):
+            for key, value in kwargs.iteritems():
+                keyname = type(self).__name__ + "." + key
+                if keyname not in self.conf.keys():
+                    self.conf[keyname] = value
+                    self.testconf[keyname] = value
+                else:
+                    self.testconf[keyname] = self.conf[keyname]
+                    t = type(value)
+                    if t == bool:
+                        if strtobool(self.conf[keyname]):
+                            self.conf[keyname] = True
+                        else:
+                            self.conf[keyname] = False
+                    else:
+                        # If value is not a boolean then typecast
+                        self.conf[keyname] = t(self.conf[keyname])
+
+            function(self, *args)
+        wrapper.__doc__ = function.__doc__
+        wrapper.__name__ = function.__name__
+        return wrapper
+    return decorated
+
+
 class PBSServiceInstanceWrapper(dict):
 
     """
@@ -435,6 +470,7 @@ class PBSTestSuite(unittest.TestCase):
     measurements = []
     additional_data = {}
     conf = {}
+    testconf = {}
     param = None
     du = DshUtils()
     _procmon = None
@@ -1526,6 +1562,8 @@ class PBSTestSuite(unittest.TestCase):
         verify that ``server`` and ``scheduler`` are up
         clean up jobs and reservations
         """
+        if self.conf:
+            self.set_test_measurements({'testconfig': self.testconf})
         if 'skip-teardown' in self.conf:
             return
         self.log_enter_teardown()
--- a/test/tests/selftest/pbs_testparams_decorator.py
+++ b/test/tests/selftest/pbs_testparams_decorator.py
@@ -1,0 +1,68 @@
+# coding: utf-8
+
+# Copyright (C) 1994-2019 Altair Engineering, Inc.
+# For more information, contact Altair at www.altair.com.
+#
+# This file is part of the PBS Professional ("PBS Pro") software.
+#
+# Open Source License Information:
+#
+# PBS Pro is free software. You can redistribute it and/or modify it under the
+# terms of the GNU Affero General Public License as published by the Free
+# Software Foundation, either version 3 of the License, or (at your option) any
+# later version.
+#
+# PBS Pro is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.
+# See the GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+# Commercial License Information:
+#
+# For a copy of the commercial license terms and conditions,
+# go to: (http://www.pbspro.com/UserArea/agreement.html)
+# or contact the Altair Legal Department.
+#
+# Altair’s dual-license business model allows companies, individuals, and
+# organizations to create proprietary derivative works of PBS Pro and
+# distribute them - whether embedded or bundled with other software -
+# under a commercial license agreement.
+#
+# Use of Altair’s trademarks, including but not limited to "PBS™",
+# "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
+# trademark licensing policies.
+
+from tests.selftest import *
+
+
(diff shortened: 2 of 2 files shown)


agrawalravi90:
@visheshh this is still being discussed in the forums right? Can you please mark this as WIP ?

visheshh:
@PBSPro/pbspro-maintainers Please review

test/tests/selftest/pbs_testparams_decorator.py
```diff
@@ -0,0 +1,68 @@
...
+        scheduling = self.conf["Test_TestparamsDecorator.scheduling"]
+        num_jobs = self.conf["Test_TestparamsDecorator.num_jobs"]
+        a = {'resources_available.ncpus': int(num_jobs)}
+        self.server.manager(MGR_CMD_SET, NODE, a, id=self.mom.shortname)
+        a = {'scheduling': scheduling}
+        self.server.manager(MGR_CMD_SET, SERVER, a)
+        j = Job(TEST_USER)
+        for _ in xrange(int(num_jobs)):
+            self.server.submit(j)
+        if eval(scheduling):
```
vstumpf: I dislike this, anyone can put any python code into the scheduling parameter and it will be run.

test/tests/selftest/pbs_testparams_decorator.py
```diff
@@ -0,0 +1,68 @@
...
+        parameter.
+        """
+        scheduling = self.conf["Test_TestparamsDecorator.scheduling"]
+        num_jobs = self.conf["Test_TestparamsDecorator.num_jobs"]
+        a = {'resources_available.ncpus': int(num_jobs)}
+        self.server.manager(MGR_CMD_SET, NODE, a, id=self.mom.shortname)
+        a = {'scheduling': scheduling}
+        self.server.manager(MGR_CMD_SET, SERVER, a)
+        j = Job(TEST_USER)
+        for _ in xrange(int(num_jobs)):
```
vstumpf: I don't think the test writer should have to convert the number of jobs to a number.

test/tests/selftest/pbs_testparams_decorator.py
```diff
@@ -0,0 +1,68 @@
...
+
+class Test_TestparamsDecorator(TestSelf):
+
+    """
+    This test suite is to verify functionality of requirements
+    decorator. The test specific parameters can be modified by
+    -p or param file.
+
+    """
+    @testparams(num_jobs=100, scheduling="False")
```
vstumpf: Generally in python, you just use `False`, not `"False"`.

test/tests/selftest/pbs_testparams_decorator.py
visheshh: @vstumpf  Would you suggest using 1 or 0 instead of true or False .

test/tests/selftest/pbs_testparams_decorator.py
visheshh: @vstumpf  As the decorator might have any type of variables . It is not feasible to typecast inside library .
Test writer knows what he has passed in the decorator , So he can easily typecast it the way he wants.

test/tests/selftest/pbs_testparams_decorator.py
visheshh: @vstumpf eval() passes it as True if it is just False it can evaluate only if it is "False". That is the reason i had it.

visheshh:
> Right now, the test parameters are being loaded as strings, even though the defaults could be of any type. Maybe the wrapper should convert the string from the config to the type of the default parameter?
> I dislike having to `int()` or worse even `eval()` parameters.

@vstumpf If we have a default typecasting to int , What will happen to strings . There will be a wrong value stored in self.conf dictionary

visheshh:
@vstumpf Please review.

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -248,6 +248,30 @@ def wrap_obj(obj):
...
+            function.__doc__ += str(key) + ' : ' + str(value) + '\n\t'
+
+        def wrapper(self, *args):
+            for key, value in kwargs.iteritems():
+                keyname = str(self.__class__.__name__) + "." + key
+                if keyname not in self.conf.keys():
+                    self.conf[keyname] = value
+                    self.testconf[keyname] = value
+                else:
+                    self.testconf[keyname] = self.conf[keyname]
```
vstumpf: At this point, you know the default value of keyname.
In theory, you can use `type(value)`  to tell you the type, and then cast `self.conf[keyname]` to that value.
The code could look like this:
```Python
t = type(value)
if t is bool: # this would require a special function, as casting any string with len > 1 to bool will return True
    self.testconf[keyname] = strasbool(self.conf[keyname]) 
else:
    self.testconf[keyname] = t(self.conf[keyname])
```

visheshh: Thanks @vstumpf , I have addressed your comments, Please review.

test/tests/selftest/pbs_testparams_decorator.py
```diff
@@ -0,0 +1,68 @@
...
+    @testparams(num_jobs=100, scheduling=False)
+    def test_testparams(self):
+        """
+        Test to submit n number of jobs and expect jobs to be
+        in running or queued state as per tunable scheduling
+        parameter.
+        """
+        scheduling = self.conf["Test_TestparamsDecorator.scheduling"]
+        num_jobs = self.conf["Test_TestparamsDecorator.num_jobs"]
+        a = {'resources_available.ncpus': int(num_jobs)}
```
vstumpf: now that num_jobs is an int, you can remove the `int()` cast.

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -248,6 +249,40 @@ def wrap_obj(obj):
...
+    Decorator to set or modify test specific parameters
+    """
+    def decorated(function):
+        function.__doc__ += "Test Params:" + "\n\t"
+        for key, value in kwargs.iteritems():
+            function.__doc__ += str(key) + ' : ' + str(value) + '\n\t'
+
+        def wrapper(self, *args):
+            for key, value in kwargs.iteritems():
+                keyname = str(self.__class__.__name__) + "." + key
```
vstumpf: I think it's more pythonic to not use two double underscore variables. Instead, you can use `type(self).__name__`
See https://stackoverflow.com/questions/510972/getting-the-class-name-of-an-instance

visheshh:
Thanks @vstumpf for the comments. Please review these minor comment fixes.

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -248,6 +249,40 @@ def wrap_obj(obj):
...
+            for key, value in kwargs.iteritems():
+                keyname = type(self).__name__ + "." + key
+                if keyname not in self.conf.keys():
+                    self.conf[keyname] = value
+                    self.testconf[keyname] = value
+                else:
+                    self.testconf[keyname] = self.conf[keyname]
+                    t = type(value)
+                    try:
+                        if strtobool(self.conf[keyname]):
```
vstumpf: I missed this, it seems strtobool will take `1` or `0` as valid boolean strings, which the test writer could have intended to use them as integers. See https://docs.python.org/2/distutils/apiref.html?highlight=distutils.util#distutils.util.strtobool

So I think you have to do a `if t is bool / else` block.

visheshh: Thanks @vstumpf , Nice catch . Well that means if we use 0 or 1 as integers then also its type will be bool. May be we should make a note of it somewhere not to use 0 or 1 for boolean.

visheshh: Hi @vstumpf, sorry for the delay.  Please review the latest change.

visheshh:
Hi @vstumpf , I have added the if type ==bool condition. Please review.

visheshh:
@vstumpf, Thanks Vincent for the review.

prakashcv13:
I have restarted the AppVeyor build.

visheshh:
Thanks @prakashcv13

visheshh:
Hi @prakashcv13, Appveyor passed please take this PR ahead.

prakashcv13:
merged #1077

PR #1079, opened 2019-04-04
Revert managers and operators in tearDown method

shilpakodli:
#### Describe Bug or Feature
Currently PTL is not reverting server attributes "managers" and "operators" at the end of the test case. 

#### Describe Your Change
Unsetting server attributes "managers" and "operators" in cleanup_jobs method which is called in tearDown function.

[pbs_managers_operators_execution_logs.txt](https://github.com/PBSPro/pbspro/files/3043326/pbs_managers_operators_execution_logs.txt)

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -8328,6 +8328,11 @@ class Server(PBSService):
             select_xt = 'x'
         job_ids = self.select(extend=select_xt)
 
+        revert_server_attribs = ['managers', 'operators']
+
+        for attrib in revert_server_attribs:
+            self.manager(MGR_CMD_UNSET, SERVER, attrib, sudo=True)
+
         # Make sure the current user is a manager. Some tests might have
         # unset the mangers attribute. Ignore 'Duplicate entry in the list'
         # error if the current user was already a manager
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -5190,6 +5190,20 @@ class Server(PBSService):
                 self.manager(MGR_CMD_DELETE, RSC, id=rescs)
         return True
 
+    def revert_server_teardown(self):
+        """
+        Method to unset only server attributes during tearDown
+        """
+        unset_list = ['managers', 'operators']
+        self.manager(MGR_CMD_UNSET, SERVER, unset_list, sudo=True)
+
+        current_user = pwd.getpwuid(os.getuid())[0]
+        a = {ATTR_managers: (INCR, current_user + '@*')}
+        try:
+            self.manager(MGR_CMD_SET, SERVER, a, sudo=True)
+        except PbsManagerError:
+            pass
+
     def save_configuration(self, outfile, mode='a'):
         """
         Save a server configuration, this includes:
@@ -11158,20 +11172,6 @@ class Scheduler(PBSService):
         self.fairshare_tree = None
         self.resource_group = None
         return self.isUp()
-
-    def revert_server_attribs(self):
-        """
-        Method to unset only server attributes during tearDown
-        """
-        unset_list = ['managers', 'operators']
-        self.manager(MGR_CMD_UNSET, SERVER, unset_list, sudo=True)
-
-        current_user = pwd.getpwuid(os.getuid())[0]
-        a = {ATTR_managers: (INCR, current_user + '@*')}
-        try:
-            self.manager(MGR_CMD_SET, SERVER, a, sudo=True)
-        except PbsManagerError:
-            pass
 
     def create_scheduler(self, sched_home=None):
         """
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -8328,11 +8328,6 @@ class Server(PBSService):
             select_xt = 'x'
         job_ids = self.select(extend=select_xt)
 
-        revert_server_attribs = ['managers', 'operators']
-
-        for attrib in revert_server_attribs:
-            self.manager(MGR_CMD_UNSET, SERVER, attrib, sudo=True)
-
         # Make sure the current user is a manager. Some tests might have
         # unset the mangers attribute. Ignore 'Duplicate entry in the list'
         # error if the current user was already a manager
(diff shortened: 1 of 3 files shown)


test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -8328,6 +8328,11 @@ def cleanup_jobs(self, extend=None, runas=None):
             select_xt = 'x'
         job_ids = self.select(extend=select_xt)
 
+        revert_server_attribs = ['managers', 'operators']
+
+        for attrib in revert_server_attribs:
+            self.manager(MGR_CMD_UNSET, SERVER, attrib, sudo=True)
```
bhroam: There is no need for the for loop.  You can pass a attrib list to manager().

test/tests/selftest/pbs_managers_operators_ext.py
```diff
@@ -60,3 +60,28 @@ def test_managers_unset_setup(self):
...
+        mgr_user2 = str(TEST_USER1)
+        a = {ATTR_managers: (INCR, mgr_user1 + '@*,' + mgr_user2 + '@*')}
+        self.server.manager(MGR_CMD_SET, SERVER, a, runas=runas)
+        a = {ATTR_operators: str(OPER_USER) + '@*'}
+        self.server.manager(MGR_CMD_SET, SERVER, a, sudo=True)
+        self.logger.info("Calling test tearDown:")
+        TestSelf.tearDown(self)
+        self.server.expect(SERVER, attrib=attr)
+        try:
+            self.server.expect(SERVER, attrib=a, max_attempts=5)
```
bhroam: I don't think the try/except is needed.  I think you can do an expect() with op=NE with the same effect.

test/fw/ptl/lib/pbs_testlib.py
hirenvadalia: Why we are reverting server attribute in cleanup_jobs()? This method is to clean up jobs only, right.

shilpakodli: Since there is code relating setting server's attribute "managers" to current user, I have made changes at same place.

bhroam: @latha-subramanian had to revert the managers here so the jobs could be deleted.

shilpakodli:
@bhroam and @hirenvadalia I have addressed your review comments. Please find attached latest execution logs.
[pbs_managers_operators_execution_logs.txt](https://github.com/PBSPro/pbspro/files/3053283/pbs_managers_operators_execution_logs.txt)

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -11163,6 +11158,20 @@ def revert_to_defaults(self):
         self.resource_group = None
         return self.isUp()
 
+    def revert_server_attribs(self):
```
bhroam: I dislike this name since it could be confused with revert_to_defaults().  How about something more specific: revert_server_tearDown() or something like that.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -11163,6 +11158,20 @@ def revert_to_defaults(self):
...
+        """
+        Method to unset only server attributes during tearDown
+        """
+        unset_list = ['managers', 'operators']
+        self.manager(MGR_CMD_UNSET, SERVER, unset_list, sudo=True)
+
+        current_user = pwd.getpwuid(os.getuid())[0]
+        a = {ATTR_managers: (INCR, current_user + '@*')}
```
bhroam: I don't know if it is worthwhile, but if the user is already in managers, this spits out an error.  It's not a harmful error though.  We want the user in managers.  We could get rid of this error by checking first, but it will require a stat call to the server.  Is it worth it to avoid printing an error?

test/fw/ptl/lib/pbs_testlib.py
hirenvadalia: @latha-subramanian Can you tell us why you had to revert the managers here? As far as I know in Server.revert_to_defaults() cleanup jobs happen first then it reverts server attributes. So why we need to revert the managers here (I mean before reverting server attributes)?

test/fw/ptl/lib/pbs_testlib.py
shilpakodli: @bhroam I am unsetting managers attribute before I set it to current user. I am doing this because it might be set to some other users in addition to current user as part of few test cases. I am also handling error which will be likely thrown from "self.manager(MGR_CMD_SET, SERVER, a, sudo=True)"

shilpakodli:
@bhroam I have addressed your review comments. Please let me know if ant changes needs to be done.

test/fw/ptl/lib/pbs_testlib.py
bhroam: Great, thanks

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -5190,6 +5190,20 @@ def revert_to_defaults(self, reverthooks=True, revertqueues=True,
                 self.manager(MGR_CMD_DELETE, RSC, id=rescs)
         return True
 
+    def revert_server_attribs(self):
```
bhroam: I don't see the name being changed to revert_server_teardown

shilpakodli: Sorry Bhroam, I missed to push one file change. Please review it now.

shilpakodli:
Closing this PR as this is already covered in setUp.

test/fw/ptl/lib/pbs_testlib.py
latha-subramanian: @hirenvadalia ,Sorry for missing this earlier. cleanup_jobs() wants mangers attribute to be set in order to delete jobs.  Some tests unset managers attribute(to test some specific scenarios) . Hence cleanup_jobs() had to set it back.  Did it answer your question?

hirenvadalia: @latha-subramanian If I understand your answer correct then we need manager attribute set to current user to delete other user's job in PBS right? if that's the only case then I don't think we need set manager here because Server.manager() can access sudo and/or runas argument and in PBS root user can delete any user's job.

latha-subramanian: We would also want to kill all the job processes (in _cleanup_large_num_jobs()). If I remember right, I wasn't able to do this without manager attribute set.

hirenvadalia: @latha-subramanian Even in case of that you don't need to set manager attribute because _cleanup_large_num_jobs() has runas argument and we can pass root in that and all 3 API (Server.status, DshUtils.run_cmd and Server.delete) which are called by _cleanup_large_num_jobs can take runas argument. So I still don't think we need to set manager attribute to delete jobs before we revert server attribute.

latha-subramanian: @hirenvadalia , Yes. I agree for the status(),run_cmd() and delete functions 'runas' argument will suffice. But there is also a qmgr call to make scheduling 'false' and 'true' in cleanup_jobs().  Is there a way out for this without setting the 'manager' attribute ?

latha-subramanian: @hirenvadalia , I checked manager() call was successful with 'runas' as root. I will raise a bug and get this one fixed. Thanks again for pointing this out !

PR #1081, opened 2019-04-04
Scheduler leaks memory while suspending job

arungrover:
#### Describe Bug or Feature
Scheduler leaks memory in two cases - 
1 - When a running job is suspended, we create execselect for the suspended job based on resources held by the job and overwrite the previously created execselect without freeing it.
2 - When we try to check whether the queued job can fit node partition, it calls check_available_resources() on the node partition but the additional errors returned by check_available_resources is overwritten by the calling function.

#### Describe Your Change
free execselect before rewriting the pointer.
do not overwrite the error returned by check_available_resources() function and move to the end of error list and then append the new error.

[sched_leak_without_fix.txt](https://github.com/PBSPro/pbspro/files/3045697/sched_leak_without_fix.txt)
[sched_leak_with_fix.txt](https://github.com/PBSPro/pbspro/files/3045698/sched_leak_with_fix.txt)

### DIFF
--- a/src/scheduler/job_info.c
+++ b/src/scheduler/job_info.c
@@ -4800,6 +4800,7 @@ void create_res_released(status *policy, resource_resv *pjob)
 		pjob->job->resreq_rel = create_resreq_rel_list(policy, pjob);
 	}
 	selectspec = create_select_from_nspec(pjob->job->resreleased);
+	free_selspec(pjob->execselect);
 	pjob->execselect = parse_selspec(selectspec);
 	free(selectspec);
 	return;
--- a/src/scheduler/node_partition.c
+++ b/src/scheduler/node_partition.c
@@ -1037,7 +1037,8 @@ resresv_can_fit_nodepart(status *policy, node_partition *np, resource_resv *resr
 				INSUFFICIENT_RESOURCE, err) == 0) {
 		if ((flags & RETURN_ALL_ERR)) {
 			can_fit = 0;
-			for (; err->next != NULL; err = err->next);
+			for (; err->next != NULL; err = err->next)
+			;
 			err->next = new_schd_error();
 			prev_err = err;
 			err = err->next;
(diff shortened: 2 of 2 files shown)


arungrover:
I've not written any new test for this fix. Exiting tests related to restrict_res_to_release_on_suspend are able to catch the issue.

src/scheduler/node_partition.c
```diff
@@ -1037,6 +1037,7 @@ resresv_can_fit_nodepart(status *policy, node_partition *np, resource_resv *resr
 				INSUFFICIENT_RESOURCE, err) == 0) {
 		if ((flags & RETURN_ALL_ERR)) {
 			can_fit = 0;
+			for (; err->next != NULL; err = err->next);
```
bhroam: Put the ; on the following line to make it clear it is an empty for loop.

src/scheduler/job_info.c
```diff
@@ -4800,6 +4800,7 @@ void create_res_released(status *policy, resource_resv *pjob)
 		pjob->job->resreq_rel = create_resreq_rel_list(policy, pjob);
 	}
 	selectspec = create_select_from_nspec(pjob->job->resreleased);
+	free_selspec(pjob->execselect);
```
agrawalravi90: Just curious, execselect already gets set inside query_jobs, so why do we need to set it again here?

arungrover: Well this is needed to add the job back to calendar. We use execselect (if it is set) to find node solution in check_nodes.

bhroam: @agrawalravi90 We reset it here since it has changed.  When it is set in query_jobs(), it uses the exec_vnode (all the nodes and all the resources).  Here we are setting it to just the resources we released.

agrawalravi90: ok, so we kind of overload execselect so that get_resresv_spec() can return only the released resources and we only look for them when finding resources for the suspended job?

I feel like we could leverage this for degraded reservations as well, set execselect to only the node(s)/resources which went down so that we only look for them when re-confirming the reservation (the server will have to set resources_released on the reservation for the node(s) which went down).

Anyways, thanks for the clarification!

PR #1082, opened 2019-04-05
Fix for invalid cache in getgids and register rshd in execution mode

hirenvadalia:
#### Describe Bug or Feature
* While adding group names in cache getgids/getdefgrpname passes invalid group name buffer length so catch_data() caches invalid group name and in next call to getgids/getdefgrpname uses that invalid group name which leads to failure to create user token in LogonNoPass() and which leads to "Failed to Impersonate user" error

* While installing Execution only, even if pbs_rshd binary exists, win_postinstall.py doesn't register PBS_RSHD service.

#### Describe Your Change
* Pass correct group name buffer length in catch_data()
* Register pbs_rshd service in execution mode

#### Link to Design Doc
* None

* [before_fix.txt](https://github.com/PBSPro/pbspro/files/3046530/before_fix.txt)
* [after_fix.txt](https://github.com/PBSPro/pbspro/files/3046531/after_fix.txt)

### DIFF
--- a/src/cmds/scripts/win_postinstall.py
+++ b/src/cmds/scripts/win_postinstall.py
@@ -413,14 +413,14 @@ def register_and_start_services(username, password):
                      username, password)
         __svc_helper(os.path.join(pbs_sbin, 'pbs_server.exe'),
                      username, password)
-        rshd_path = os.path.join(pbs_sbin, 'pbs_rshd.exe')
-        if os.path.isfile(rshd_path):
-            __svc_helper(rshd_path, username, password)
         set_svr_defaults()
     if installtype in ('server', 'comm'):
         __svc_helper(os.path.join(pbs_sbin, 'pbs_comm.exe'),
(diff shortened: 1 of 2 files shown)


subhasisb:
we need another pair of eyes on this

prakashcv13:
@hirenvadalia - the changes look good.
Only we need to remove getgids() from the bug description as it is already taking care of adding +1 to GNLEN while calling cache_data().

hirenvadalia:
> @hirenvadalia - the changes look good.
> Only we need to remove getgids() from the bug description as it is already taking care of adding +1 to GNLEN while calling cache_data().

But @prakashcv13 this is happening in getgids (and getdefgrpname) function only not.

prakashcv13:
@hirenvadalia - I am waiting for Codacy check to get over - will merge it in once done.

prakashcv13:
Merged #1082

PR #1085, opened 2019-04-09
preempt retry not working

prakashcv13:
#### Describe Bug or Feature
* scheduler tried only once to preempt jobs. If the server reported failure in preempting a job, scheduler will not try again.

#### Describe Your Change
* server was not sending the job id - made the change.
* scheduler flag "done" in find_and_preempt_jobs() was not getting unset when preemption failed for a job.

[ptl_after_fix.txt](https://github.com/PBSPro/pbspro/files/3060370/ptl_after_fix.txt)
[ptl_before_fix.txt](https://github.com/PBSPro/pbspro/files/3060371/ptl_before_fix.txt)

### DIFF
--- a/src/server/req_preemptjob.c
+++ b/src/server/req_preemptjob.c
@@ -86,7 +86,6 @@ void
 req_preemptjobs(struct batch_request *preq)
 {
 	int			i = 0;
-	int			j = 0;
 	int			count = 0;
 	job			*pjob = NULL;
 	preempt_job_info	*ppj = NULL;
@@ -102,7 +101,7 @@ req_preemptjobs(struct batch_request *preq)
 
 		pjob->preempt_order = svr_get_preempt_order(pjob, psched);
 		pjob->preempt_order_index = 0;
-		switch((int)pjob->preempt_order[0].order[j]) {
+		switch((int)pjob->preempt_order[0].order[0]) {
 			case PREEMPT_METHOD_SUSPEND:
 				issue_signal(pjob, SIG_SUSPEND, post_signal_req, pjob, preq);
 				break;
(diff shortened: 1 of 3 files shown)


src/server/req_preemptjob.c
```diff
@@ -102,7 +101,7 @@ req_preemptjobs(struct batch_request *preq)
 
 		pjob->preempt_order = svr_get_preempt_order(pjob, psched);
 		pjob->preempt_order_index = 0;
-		switch((int)pjob->preempt_order[0].order[j]) {
+		switch((int)pjob->preempt_order[0].order[0]) {
```
bhroam: Won't this always use the first preemption method?  If I set preempt_order to SCR, won't it only try S and not C or R if S fails?

test/tests/functional/pbs_preemption.py
```diff
@@ -194,3 +194,60 @@ def test_preempt_sort_when_unset(self):
...
+        a = {'resources_available.ncpus': 2}
+        self.server.manager(MGR_CMD_SET, NODE, a, id=self.mom.shortname)
+
+        abort_script = """#!/bin/bash
+exit 3
+"""
+        self.abort_file = self.du.create_temp_file(body=abort_script)
+        self.du.chmod(path=self.abort_file, mode=0755)
+        self.du.chown(path=self.abort_file, uid=0, gid=0, runas=ROOT_USER)
+        c = {'$action': 'checkpoint_abort 30 !' + self.abort_file + ' %sid'}
```
bhroam: You don't actually need the %sid since you aren't using it in the script.  All %sid is doing is passing the session ID as the first argument of the script.

test/tests/functional/pbs_preemption.py
```diff
@@ -194,3 +194,60 @@ def test_preempt_sort_when_unset(self):
...
+        """
+        Test to make sure that preemption is retried if it fails.
+        """
+        a = {'resources_available.ncpus': 2}
+        self.server.manager(MGR_CMD_SET, NODE, a, id=self.mom.shortname)
+
+        abort_script = """#!/bin/bash
+exit 3
+"""
+        self.abort_file = self.du.create_temp_file(body=abort_script)
```
bhroam: Why save this in a class variable?

src/server/req_preemptjob.c
prakashcv13: trying the other methods is done in reply_preempt_jobs_request(). the variable j was not serving any purpose.

PR #1087, opened 2019-04-09
Failed calendared job took up a top job slot

bhroam:
#### Describe Bug or Feature
When a topjob fails to be added to the calendar (usually since it can never run), it still takes up a topjob slot in the backfill_depth.

#### Describe Your Change
The function that adds jobs to the calendar returned success when a job failed to be added to the calendar.  We just needed to return the failure, and it no longer takes up a top job slot.

The other half of the top job code which didn't change is in main_sched_loop().  This might be helpful to look at when reviewing.

[topjob_before.log](https://github.com/PBSPro/pbspro/files/3061146/topjob_before.log)
[topjob_after.log](https://github.com/PBSPro/pbspro/files/3061148/topjob_after.log)

### DIFF
--- a/src/scheduler/fifo.c
+++ b/src/scheduler/fifo.c
@@ -2054,14 +2054,9 @@ add_job_to_calendar(int pbs_sd, status *policy, server_info *sinfo,
 		schdlog(PBSEVENT_DEBUG, PBS_EVENTCLASS_JOB, LOG_DEBUG,
 			bjob->name, log_buf);
 	} else if (start_time == 0) {
-		/* In the case where start_time = 0, we don't want mark the job as
-		 * can_never_run because there are transient cases (like node state)
-		 * that we don't handle in our simulation that can fix themselves in
-		 * real life.  Reconsider this decision once the simulation is more
-		 * robust
-		 */
 		schdlog(PBSEVENT_SCHED, PBS_EVENTCLASS_JOB, LOG_WARNING, topjob->name,
 			"Error in calculation of start time of top job");
+		return 0;
 	}
 	free_server(nsinfo, 1);
 	
--- a/test/tests/functional/pbs_calendaring.py
+++ b/test/tests/functional/pbs_calendaring.py
@@ -140,7 +140,7 @@ class TestCalendaring(TestFunctional):
 
     def test_topjob_fail(self):
         """
-        Test that when we fail to add a job to the calendar it doesn't 
+        Test that when we fail to add a job to the calendar it doesn't
         take up a topjob slot.  The server's backfill_depth is 1 by default,
         so we just need to submit a job that can never run and a job that can.
         The can never run job will fail to be added to the calendar and the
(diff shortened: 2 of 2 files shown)


src/scheduler/fifo.c
```diff
@@ -2054,14 +2054,9 @@ add_job_to_calendar(int pbs_sd, status *policy, server_info *sinfo,
 		schdlog(PBSEVENT_DEBUG, PBS_EVENTCLASS_JOB, LOG_DEBUG,
 			bjob->name, log_buf);
 	} else if (start_time == 0) {
-		/* In the case where start_time = 0, we don't want mark the job as
```
agrawalravi90: why did you remove this comment? is this information incorrect? is yes then we should probably modify the comment to mention the correct reason why we aren't marking the job as can_never_run..

test/tests/functional/pbs_calendaring.py
```diff
@@ -137,3 +137,45 @@ def test_topjob_start_time_of_subjob(self):
...
+    def test_topjob_fail(self):
+        """
+        Test that when we fail to add a job to the calendar it doesn't
+        take up a topjob slot.  The server's backfill_depth is 1 by default,
+        so we just need to submit a job that can never run and a job that can.
+        The can never run job will fail to be added to the calendar and the
+        second job will be.
+        """
+        attrs = {'resources_available.ncpus': 1}
+        self.server.create_vnodes('vn', attrib=attrs, num=2,
```
agrawalravi90: just curious, any particular reason why you decided to create 2 vnodes instead of just using the one available and setting its ncpus?

src/scheduler/fifo.c
bhroam: The comment was obsolete.  The old message that was printed said the job would never run.  This confused people since they'd see a message saying the job would never run, and then later in the day it'd run.  The problem could have been that a bunch of nodes were down making it impossible to run the job in the system at that time.  Once the nodes came back online, the job ran.  I changed this message to say we had an error adding the job to the calendar.  The comment referred to the previous log message.

test/tests/functional/pbs_calendaring.py
bhroam: I needed to create a situation where a job could never run, but the scheduler doesn't realize that.  In this case there are two nodes with 1 cpu on it.  The can never run job is requesting 1 node with 2 cpus.  We have to be careful when creating can never run jobs, because the scheduler is really good at detecting that jobs can never run.  If it does, it won't add the job to the calendar.  If I turned it around and had one node with 2 cpus and requested 2 nodes with 1 cpu, the scheduler would have caught it as a can never run job and not added it to the calendar

agrawalravi90: Thanks for clarifying. Can you please add a comment explaining this in the test? It's not really obvious from the code.

bhroam: I added a comment.  Take a look and see if you like it.

PR #1089, opened 2019-04-10
Add quotes to string attributes when printing in qmgr

vstumpf:
#### Describe Bug or Feature
* [1] When string attribute values with spaces or quotes are printed with qmgr, it does not quote them. When the output is used as input into qmgr, qmgr can't parse the input and fails.
* [2] When PTL tries to set an string attribute with a value with spaces, it does not quote it, and PTL fails to set the attribute.

#### Describe Your Change
* [1] When qmgr prints a string attribute that needs to be quoted, it will quote it.
* [2] When `server.manager()` is called with an attrval with spaces, it will quote it.

#### Link to Design Doc
None

new:
[after-fix3.txt](https://github.com/PBSPro/pbspro/files/3082120/after-fix3.txt)

[before-fix.txt](https://github.com/PBSPro/pbspro/files/3065676/before-fix.txt)
[after-fix.txt](https://github.com/PBSPro/pbspro/files/3065677/after-fix.txt)

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -5071,7 +5071,7 @@ class Server(PBSService):
         ignore_attrs += [ATTR_status, ATTR_total, ATTR_count]
         ignore_attrs += [ATTR_rescassn, ATTR_FLicenses, ATTR_SvrHost]
         ignore_attrs += [ATTR_license_count, ATTR_version, ATTR_managers]
-        ignore_attrs += [ATTR_pbs_license_info,  ATTR_power_provisioning]
+        ignore_attrs += [ATTR_pbs_license_info, ATTR_power_provisioning]
         unsetlist = []
         setdict = {}
         skip_site_hooks = ['pbs_cgroups']
@@ -6523,8 +6523,13 @@ class Server(PBSService):
                                 op = '='
                             # handle string arrays as double quotes if
                             # not already set:
-                            if isinstance(v, str) and ',' in v and v[0] != '"':
-                                v = '"' + v + '"'
+                            if isinstance(v, str):
+                                if ',' in v and v[0] != '"':
+                                    v = '"' + v + '"'
+                                elif any((c in v) for c in set(", \n'")):
+                                    v = '"%s"' % v
+                                elif '"' in v:
+                                    v = "'%s'" % v
                             kvpairs += [str(k) + op + str(v)]
                         if kvpairs:
                             execcmd += [",".join(kvpairs)]
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -6526,10 +6526,11 @@ class Server(PBSService):
                             if isinstance(v, str):
                                 if ',' in v and v[0] != '"':
                                     v = '"' + v + '"'
-                                elif any((c in v) for c in set(", \n'")):
-                                    v = '"%s"' % v
-                                elif '"' in v:
-                                    v = "'%s'" % v
+                                elif any((c in v) for c in set(', \'\n"')):
+                                    if '"' in v:
+                                        v = "%s'" % v
+                                    else:
+                                        v = '"%s"' % v
                             kvpairs += [str(k) + op + str(v)]
                         if kvpairs:
                             execcmd += [",".join(kvpairs)]
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -6528,7 +6528,7 @@ class Server(PBSService):
                                     v = '"' + v + '"'
                                 elif any((c in v) for c in set(', \'\n"')):
                                     if '"' in v:
-                                        v = "%s'" % v
+                                        v = "'%s'" % v
                                     else:
                                         v = '"%s"' % v
                             kvpairs += [str(k) + op + str(v)]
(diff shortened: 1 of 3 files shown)


test/tests/functional/pbs_qmgr.py
```diff
@@ -159,3 +159,27 @@ def test_multi_attributes(self):
...
+        """
+        Test to verify that if a string attribute has spaces, the value
+        is quoted when printed
+        """
+        a = {'default_qsub_arguments': '-V -lplace=scatter'}
+        self.server.manager(MGR_CMD_SET, SERVER, a)
+        qmgr_cmd = \
+            os.path.join(self.server.pbs_conf['PBS_EXEC'], 'bin', 'qmgr') + \
+            " -c '%s'"
+        qmgr_cmd_print = qmgr_cmd % ('p s @default default_qsub_arguments')
```
bhroam: You don't need @default on a print server.  It does that by default.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -6523,8 +6523,13 @@ def manager(self, cmd, obj_type, attrib=None, id=None, extend=None,
...
                             # not already set:
-                            if isinstance(v, str) and ',' in v and v[0] != '"':
-                                v = '"' + v + '"'
+                            if isinstance(v, str):
+                                if ',' in v and v[0] != '"':
+                                    v = '"' + v + '"'
+                                elif any((c in v) for c in set(", \n'")):
+                                    v = '"%s"' % v
+                                elif '"' in v:
+                                    v = "'%s'" % v
```
lisa-altair: What if the string has a double quote " and a space?  Wouldn't there be a problem?

bhroam:
Looks good to me.  I sign off.

lisa-altair:
Please attach your latest test results.

test/tests/functional/pbs_qmgr.py
```diff
@@ -159,3 +159,25 @@ def test_multi_attributes(self):
             self.logger.info("looking for error, %s" % msg)
             self.assertTrue(msg in e.msg[0])
         self.server.expect(NODE, 'queue', op=UNSET, id=self.mom.shortname)
+
+    def test_string_quoting(self):
+        """
+        Test to verify that if a string attribute has spaces, the value
```
latha-subramanian: @vstumpf , should we also a test to address the case where string attribute has quotes ?

test/tests/functional/pbs_qmgr.py
```diff
@@ -160,24 +160,42 @@ def test_multi_attributes(self):
...
+        fn = self.du.create_temp_file()
         for line in ret['out']:
             if '#' in line:
                 continue
-            qmgr_cmd_check = qmgr_cmd % (line)
-            ret_s = self.du.run_cmd(
-                self.server.hostname, cmd=qmgr_cmd_check, as_script=True)
+            if 'create node' in line:
+                continue
+            with open(fn, 'w') as f:
```
latha-subramanian: Shouldn't the mode be 'w+' as it has to be read by qmgr?

test/tests/functional/pbs_qmgr.py
```diff
@@ -160,24 +160,42 @@ def test_multi_attributes(self):
...
-            " -c '%s'"
-        qmgr_cmd_print = qmgr_cmd % ('p s default_qsub_arguments')
+        a = {'comment': comment}
+        self.server.manager(MGR_CMD_SET, NODE, a, id=self.mom.hostname)
+        qmgr_path = \
+            os.path.join(self.server.pbs_conf['PBS_EXEC'], 'bin', 'qmgr')
+        qmgr_cmd_print = qmgr_path + \
+            (' -c "p n %s comment"' % self.mom.hostname)
         ret = self.du.run_cmd(self.server.hostname,
                               cmd=qmgr_cmd_print, as_script=True)
```
latha-subramanian: qmgr output could be redirected to a file directly via run_cmd() 's stdout parameter. Other tests in the suite seem to do that. Any reason you want to do it this way ?

test/tests/functional/pbs_qmgr.py
vstumpf: `w+` open the file for updating, if it's already created. Otherwise it will create a new file.
`w` will truncate the file (which is what we want to do, since we only create one file and write to it many times.

test/tests/functional/pbs_qmgr.py
vstumpf: For a reason unknown to me, when we run `print node <nodename> comment`, it prints the command to create the node and the command to set the comment. If we pipe it back into qmgr, it will print an error.

bhroam: The print option is meant to recreate.  The idea is you do a qmgr -c 'print server' > file and then on another server you redirect that back in to qmgr to recreate your settings.  So you are printing a node, you want to recreate that node.  This is why create node is printed.  If you wanted to list the attributes, you would use the list command.

latha-subramanian: @vstumpf . Thanks for clarifying. I am fine with this way then.

vstumpf: It would make more sense if I could print multiple attributes at once, for example `print node <nodename> resv_enable sharing vn_type`, and it would be able to recreate the node with the specific attributes. But because I can only print one at a time, I would have to print each separately and remove the duplicate `creates`.

But I understand probably no one uses it this way, so it's fine as it is.

PR #1090, opened 2019-04-12
Fixed race condition in test_fairshare_enhanced

bhroam:
#### Describe Bug or Feature
SmoteTest.test_fairshare_enhanced can fail on slow machines due to a race condition.  The test jobs are 4s long.  If a slow machine freezes at the wrong time, one of the 4s jobs can complete before PTL will notice it ever ran.

#### Describe Your Change
Extend sleep time for jobs to 10s

[fairshare.log](https://github.com/PBSPro/pbspro/files/3071136/fairshare.log)

### DIFF
--- a/test/tests/pbs_smoketest.py
+++ b/test/tests/pbs_smoketest.py
@@ -1364,13 +1364,13 @@ class SmokeTest(PBSTestSuite):
         self.server.manager(MGR_CMD_SET, NODE, node_attr, self.mom.shortname)
         self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'False'})
         job_attr = {'Resource_List.select': '1:ncpus=1:foo1=20',
-                    'Resource_List.walltime': 4}
+                    'Resource_List.walltime': 10}
         J1 = Job(TEST_USER2, attrs=job_attr)
-        J1.set_sleep_time(4)
+        J1.set_sleep_time(10)
         J2 = Job(TEST_USER3, attrs=job_attr)
-        J2.set_sleep_time(4)
+        J2.set_sleep_time(10)
         J3 = Job(TEST_USER1, attrs=job_attr)
-        J3.set_sleep_time(4)
+        J3.set_sleep_time(10)
         j1id = self.server.submit(J1)
         j2id = self.server.submit(J2)
         j3id = self.server.submit(J3)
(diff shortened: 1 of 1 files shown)


agrawalravi90:
Maybe this will not work well with fairshare, but is it possible to just leave the jobs at the default sleep time (100 seconds?) and then delete jobs manually instead of waiting for them to finish?

bhroam:
@agrawalravi90 I considered something similar to that.  From a scheduler's point of view, the jobs can be deleted.  We just need to know the order in which they are run.  I had to guess why the test was written the way it was.  I thought it was because QA wanted to test the way the customer would run it.  They want to see the jobs run and finish and then run the next one.  Your idea would work with fairshare because of the formula in use.  It accumulated all the fairshare usage for the job in one cycle.  We just need to make sure a cycle is run after each job has started for this to happen.  The test does run those cycles.  Do you want me to change the test?

agrawalravi90:
> Do you want me to change the test?

I just think that it will make the test more robust against such race conditions, and also make it run faster. The scheduler is kind of self contained, and for this test, so as long as we are triggering the sched cycle to accumulate usage, it shouldn't matter how the jobs end (deleted vs run their due course). So if it's not too much trouble, I'd prefer that you change it to delete the jobs instead of waiting out.

test/tests/pbs_smoketest.py
```diff
@@ -1414,36 +1415,33 @@ def test_fairshare_enhanced(self):
...
+        msg += j1id + " is deleted"
         self.logger.info(msg)
         self.server.expect(JOB, {'job_state': 'R'}, id=j3id)
         self.server.expect(JOB, {'job_state': 'Q'}, id=j2id)
         self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'True'})
 
+        self.server.delete(j3id)
         msg = "Checking the job state of " + j2id + ", runs after "
-        msg += j1id + " completes"
+        msg += j1id + " is deleted"
```
arungrover: shouldn't this comment use j3id? because that is what we deleted.

bhroam: Looks to be a copy paste error when the test was initially written.  It's funny since this is a smoke test and one of the tests that is run most often.  We always missed it :)

Issue #1080, opened 2019-04-04
Mom: Double free with multiple exechost_periodic hooks


vchlum:
Process two or more hooks in mom_process_hooks() results in double free of php structure in the post_run_hook(). The problem is that the same php structure is passed to post_run_hook() in run_hook() and the php structure is freed in each of the post_run_hook():

==4767== Invalid read of size 4
==4767==    at 0x44E6DB: post_run_hook (mom_hook_func.c:3150)
==4767==    by 0x4C7C4D: dispatch_task (work_task.c:142)
==4767==    by 0x4C7F26: default_next_task (work_task.c:301)
==4767==    by 0x45F039: main (mom_main.c:9739)
==4767==  Address 0xa2130d0 is 32 bytes inside a block of size 64 free'd
==4767==    at 0x4C29E90: free (vg_replace_malloc.c:473)
==4767==    by 0x44E6F0: post_run_hook (mom_hook_func.c:3151)
==4767==    by 0x4C7C4D: dispatch_task (work_task.c:142)
==4767==    by 0x4C7F26: default_next_task (work_task.c:301)
==4767==    by 0x45F039: main (mom_main.c:9739)
==4767== 
==4767== Invalid free() / delete / delete[] / realloc()
==4767==    at 0x4C29E90: free (vg_replace_malloc.c:473)
==4767==    by 0x44E6F0: post_run_hook (mom_hook_func.c:3151)
==4767==    by 0x4C7C4D: dispatch_task (work_task.c:142)
==4767==    by 0x4C7F26: default_next_task (work_task.c:301)
==4767==    by 0x45F039: main (mom_main.c:9739)
==4767==  Address 0xa2130b0 is 0 bytes inside a block of size 64 free'd
==4767==    at 0x4C29E90: free (vg_replace_malloc.c:473)
==4767==    by 0x44E6F0: post_run_hook (mom_hook_func.c:3151)
==4767==    by 0x4C7C4D: dispatch_task (work_task.c:142)
==4767==    by 0x4C7F26: default_next_task (work_task.c:301)
==4767==    by 0x45F039: main (mom_main.c:9739)
==4767==

vchlum:
This issue is fixed.

PR #1092, opened 2019-04-12
double free with multiple exechost_periodic mom hooks

vchlum:
#### Describe Bug or Feature
Mom crashes with two or more exechost_periodic hooks in one calling of mom_process_hooks(). The issue is #1080.

exechost_periodic hook can be processed in parallel. In the mom_process_hooks() each exechost_periodic hook is forked and the php structure is passed to each hook run and freed in each post_run_hook() call.

#### Describe Your Change
We can not use only one instance of php because we would not know where or when to freed. So, the php structure is duplicated after each exechost_periodic hook is started. This way each exechost_periodic calling has its own copy of php and it can be safely freed. Since the duplication is done after the calling we do not need the exception for not freeing php for exechost_periodic hook in the end of mom_process_hooks(). The php is always freed in this place.

#### Link to Design Doc

None

[ptl_php_double_free.txt](https://github.com/PBSPro/pbspro/files/3074233/ptl_php_double_free.txt)

### DIFF
--- a/src/resmom/mom_hook_func.c
+++ b/src/resmom/mom_hook_func.c
@@ -640,6 +640,39 @@ vnl_add_vnode_entries(vnl_t *vnl, vmpiprocs *vnode_entry, int num_vnodes,
 
 /**
  * @brief
+ *	Duplicates pointer to hooks parameters.
+ *
+ * @param[in]   php - structure for duplication
+ *
+ * @return new_php	for success
+ * @return NULL		for error
+ *
+ */
+mom_process_hooks_params_t
+*duplicate_php(mom_process_hooks_params_t *php) {
+	mom_process_hooks_params_t *new_php;
+
+	if ((new_php = (mom_process_hooks_params_t *)malloc(
+		sizeof(mom_process_hooks_params_t))) == NULL) {
+		log_err(errno, __func__, MALLOC_ERR_MSG);
+		return NULL;
+	}
+
+	new_php->hook_event = php->hook_event;
+	new_php->req_user = php->req_user;
+	new_php->req_host = php->req_host;
+	new_php->hook_msg = php->hook_msg;
+	new_php->msg_len = php->msg_len;
+	new_php->update_svr = php->update_svr;
+	new_php->hook_input = php->hook_input;
+	new_php->hook_output = php->hook_output;
+	new_php->parent_wait = php->parent_wait;
+
+	return new_php;
+}
+
+/**
+ * @brief
  *	Runs the hook 'phook' in a child process in response to 'event_type'
  *	with input parameter 'hook_input'.
  *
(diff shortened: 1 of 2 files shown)


arungrover:
@vchlum Thanks for addressing this issue.
While your fix works perfectly and make each periodic hook have its own copy of hook parameters, I feel we didn't have to allocate the parameters in the first place.

If you see first thing post_run_hook() does in case of PERIODIC hook is free up php and every subsequent call to run_hook from run_periodic_hook_bg passes hook parameters as NULL.

I also find it useless that we pass in parent_wait parameter to run_hook when the same parameter is part of php. It would only make sense to have this parameter passed separately when the caller wants to overload the value of parent_wait in php structure, but I couldn't find any place where we are doing that.

I'll leave it upto you if you want to update the fix by not allocating hook parameter structure. I'm fine if you want to leave it as is.

vchlum:
@arungrover I am open to update the fix with your suggestion.

I am maybe missing something, but we need php in post_run_hook() because of php->hook_event. We wouldn't know that the hook is EXECHOST_PERIODIC without php here, which means that we wouldn't know that php is not needed anymore without php:-). AFAIS, there is no other way how to read the hook event type in post_run_hook(). Please, correct me.

arungrover:
@vchlum you are right, post_run_hook() uses php to detect which hook event the call is pertaining to. Although in case of periodic hooks there is no reason to register post_run_hook as the callback, post_periodic_hook is what is needed to be called. So, in mom_process_hooks we could just pass post_periodic_hook as the callback to run_hook() for PERIODIC hooks.

In any case, I think even if we fix it this way, we will end up adding a bunch of special code for managing periodic hooks. Your current fix will allocate php for the first time when all periodic hooks are run but after that allocation does not happen again, and it is cleaner than the fix I'm proposing. 
I guess, it is better to have it the way it is now. I'll sign-off.

src/resmom/mom_hook_func.c
```diff
@@ -638,6 +638,39 @@ vnl_add_vnode_entries(vnl_t *vnl, vmpiprocs *vnode_entry, int num_vnodes,
...
+ *	Duplicates pointer to hooks parameters.
+ *
+ * @param[in]   php - structure for duplication
+ *
+ * @return new_php	for success
+ * @return NULL		for error
+ *
+ */
+mom_process_hooks_params_t
+*duplicate_php(mom_process_hooks_params_t *php) {
```
prakashcv13: please move the "{" to the first column of the next line.

vchlum: Done

prakashcv13:
@vchlum - thank you for making the change - LGTM.

prakashcv13:
merged #1092

PR #1094, opened 2019-04-12
libical1a is not in Debian stretch/current stable

enckse:
#### Describe Bug or Feature
in current debian stable (stretch) `libical1a` does not exist,  `libical2` does (yes it builds). I believe this will change when buster releases in the coming months as well...

#### Describe Your Change
This is install text/doc change only

#### Link to Design Doc
n/a

n/a

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -46,7 +46,7 @@ How to install PBS Pro using the configure script.
   For Debian systems you should run the following command as root:
 
     apt-get install expat libedit2 postgresql python postgresql-contrib sendmail-bin \
-      sudo tcl tk libical1a
+      sudo tcl tk libical2
 
 3. Open a terminal as a normal (non-root) user, unpack the PBS Pro
   tarball, and cd to the package directory.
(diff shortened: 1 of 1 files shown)


INSTALL
```diff
@@ -46,7 +46,7 @@ How to install PBS Pro using the configure script.
   For Debian systems you should run the following command as root:
 
     apt-get install expat libedit2 postgresql python postgresql-contrib sendmail-bin \
-      sudo tcl tk libical1a
+      sudo tcl tk libical2
```
mike0042: Thanks for the contribution. At one point, the old instructions were accurate. Do you happen to know if this change will fail with older versions of Debian? If so, which ones?

enckse: I think it probably worked with jessie (oldstable) but I can't confirm (when I was searching debian packages this appeared like it was true). Anything before that I have no idea/it would take a lot more digging.

enckse: (clarify): the old instructions would work with jessie but the new instruction/package would not

mike0042: Looks like Jessie will be officially supported until June 30, 2020. Source: https://wiki.debian.org/LTS/

I think it would be better if we provide two instructions for Debian, one for Jessie and one for Stretch.

mike0042:
@enckse - Could you please re-sign your latest commit? That will restart the CI checks. One of the Windows builds looks like it got wedged. Nothing to do with your code.

PR #1096, opened 2019-04-16
pbs_snapshot --obfuscate doesn't obfuscate everything

agrawalravi90:
#### Describe Bug or Feature

pbs_snapshot --obfuscate has many issues: it doesn’t obfuscate everything, there are bugs with obfuscating special attributes (like it only obfuscates the first entry in managers, acl attributes etc.), the obfuscated string has the same length as the original string, which could be decrypted back to the original, etc.

#### Describe Your Change

Suggesting a redesign of pbs_snapshot --obfuscate. Please refer to the linked design for more details.

#### Link to Design Doc

https://pbspro.atlassian.net/wiki/spaces/PD/pages/1233125400/Redesigning+pbs+snapshot+--obfuscate

Ran pbs_snapshot.unittest.py
Single node test results:
[snapshot_test.log](https://github.com/PBSPro/pbspro/files/3152131/snapshot_test.log)

Will post multi-node results soon.

### DIFF
--- a/test/fw/bin/pbs_snapshot
+++ b/test/fw/bin/pbs_snapshot
@@ -171,9 +171,12 @@ def remotesnap_thread(logger, host):
     du.rm(hostname=host, path=snap_home, recursive=True, force=True)
 
 
-def capture_local_snap():
+def capture_local_snap(sudo_val):
     """
     Helper method to capture snapshot of the local host
+
+    :param sudo_val - Value of the --with-sudo option
+    :type sudo_val - bool
 
     :returns Name of the snapshot directory/tar file captured
     """
@@ -185,12 +188,12 @@ def capture_local_snap():
 
     with PBSSnapUtils(out_dir, acct_logs=acct_logs,
                       daemon_logs=daemon_logs, create_tar=create_tar,
-                      log_path=log_path, with_sudo=with_sudo) as snap_utils:
+                      log_path=log_path, with_sudo=sudo_val) as snap_utils:
         snap_name = snap_utils.capture_all()
 
     if obfuscate:
         obj = ObfuscateSnapshot()
-        obj.obfuscate_snapshot(snap_name, map_file)
+        obj.obfuscate_snapshot(snap_name, map_file, sudo_val)
         if additional_hosts is None:
             # Now create the tar
             outtar = snap_name + ".tgz"
@@ -319,7 +322,7 @@ if __name__ == '__main__':
             thread_p.start()
         else:
             # Capture a local snapshot
-            main_snap = capture_local_snap()
+            main_snap = capture_local_snap(with_sudo)
 
         if thread_p is not None:
             # Let's get the main host's snapshot first
@@ -368,7 +371,7 @@ if __name__ == '__main__':
         du.rm(path=p_snappath, force=True)
     else:
         # Capture snapshot of the local host
-        outtar = capture_local_snap()
+        outtar = capture_local_snap(with_sudo)
 
     if outtar is not None:
         print "Snapshot available at: " + outtar
--- a/test/fw/bin/pbs_snapshot
+++ b/test/fw/bin/pbs_snapshot
@@ -49,7 +49,7 @@ from getopt import GetoptError
 from threading import Thread
 
 from ptl.lib.pbs_testlib import PtlConfig
-from ptl.utils.pbs_snaputils import PBSSnapUtils
+from ptl.utils.pbs_snaputils import PBSSnapUtils, ObfuscateSnapshot
 from ptl.utils.pbs_cliutils import CliUtils
 from ptl.utils.pbs_dshutils import DshUtils
 
@@ -136,7 +136,7 @@ def remotesnap_thread(logger, host):
     cmd = [host_pbssnappath, "-o", snap_home,
            "--daemon-logs=" + str(daemon_logs),
            "--accounting-logs=" + str(acct_logs)]
-    if anonymize:
+    if obfuscate:
         cmd.extend(["--obfuscate", "--map=" + map_file])
     if with_sudo:
         cmd.append("--with-sudo")
@@ -171,6 +171,38 @@ def remotesnap_thread(logger, host):
     du.rm(hostname=host, path=snap_home, recursive=True, force=True)
 
 
+def capture_local_snap():
+    """
+    Helper method to capture snapshot of the local host
+
+    :returns Name of the snapshot directory/tar file captured
+    """
+    if obfuscate or additional_hosts is not None:
+        # We will need to the captured snapshot directory in these 2 cases
+        create_tar = False
+    else:
+        create_tar = True
+
+    with PBSSnapUtils(out_dir, acct_logs=acct_logs,
+                      daemon_logs=daemon_logs, create_tar=create_tar,
+                      log_path=log_path, with_sudo=with_sudo) as snap_utils:
+        snap_name = snap_utils.capture_all()
+
+    if obfuscate:
+        obj = ObfuscateSnapshot()
+        obj.obfuscate_snapshot(snap_name, map_file)
+        if additional_hosts is None:
+            # Now create the tar
+            outtar = snap_name + ".tgz"
+            with tarfile.open(outtar, "w:gz") as tar:
+                tar.add(snap_name, arcname=os.path.basename(snap_name))
+            # Delete the snapshot directory itself
+            du.rm(path=snap_name, recursive=True, force=True)
+            snap_name = outtar
+
+    return snap_name
+
+
 if __name__ == '__main__':
 
     # Arguments to PBSSnapUtils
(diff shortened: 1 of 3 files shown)


agrawalravi90:
FYI, I won't fix the rest of Codacy's warning as they are not relevant.

agrawalravi90:
@arungrover can you please review this PR?

test/fw/ptl/utils/pbs_snaputils.py
```diff
@@ -212,11 +216,307 @@
...
+    vals_to_del = []
+    bu = BatchUtils()
+    du = DshUtils()
+    logger = logging.getLogger(__name__)
+
+    job_attrs_del = [ATTR_v, ATTR_e, ATTR_jobdir,
+                     ATTR_submit_arguments, ATTR_o, ATTR_S]
+    resv_attrs_del = [ATTR_v]
+    svr_attrs_del = [ATTR_mailfrom]
+    job_attrs_obf = [ATTR_euser, ATTR_egroup, ATTR_project, ATTR_A,
```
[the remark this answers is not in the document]
agrawalravi90: We do a sed at the end to replace all instances of usernames etc. from all files, so comment will get obfuscated that way.
The idea is: Find all sensitive users, groups, hostnames etc., create obfuscated values for them and then use that to obfuscate all files in the snapshot. So, job_attrs_obf only needs to contain the attributes which can add to the list of users, groups etc. The usernames that the Comment attribute will contain should already be covered by the other attributes in that list. Does that make sense?

agrawalravi90:
Thanks @arungrover !

test/fw/ptl/utils/pbs_snaputils.py
```diff
@@ -212,11 +216,307 @@
...
+        sed_path = self.du.which(exe="sed")
+        if sed_path is None or sed_path == "sed":
+            raise IOError("Couldn't find path to sed command, aborting .. ")
+        for root, _, fnames in os.walk(snap_dir):
+            for fname in fnames:
+                fpath = os.path.join(root, fname)
+
+                # Obfuscate values from val_obf_map
+                for key, val in self.val_obf_map.iteritems():
+                    sedcmd = r"%s -i 's/\b%s\b/%s/g' %s" % (sed_path, key,
```
bhroam: Why run sed when we can do this directly in python with re.sub()

test/tests/functional/pbs_snapshot_unittest.py
```diff
@@ -798,6 +798,109 @@ def test_remote_primary_multinode(self):
...
+        manager2 = str(TEST_USER) + "@*"
+        self.server.manager(MGR_CMD_SET, SERVER,
+                            {ATTR_managers: (INCR, manager2)},
+                            sudo=True)
+        real_values[ATTR_managers] = [manager1, manager2]
+
+        operator = str(OPER_USER) + '@*'
+        self.server.manager(MGR_CMD_SET, SERVER,
+                            {ATTR_operators: (INCR, operator)},
+                            sudo=True)
```
bhroam: You don't need to use sudo.  PTL makes the user running the tests a manager.

test/tests/functional/pbs_snapshot_unittest.py
```diff
@@ -798,6 +798,109 @@ def test_remote_primary_multinode(self):
...
+        real_values[ATTR_operators] = [operator]
+
+        real_values[ATTR_SvrHost] = [self.server.hostname]
+
+        # Create a queue with acls set
+        a = {ATTR_qtype: 'execution', ATTR_start: 't', ATTR_enable: 't',
+             ATTR_aclgren: 't', ATTR_aclgroup: TSTGRP0,
```
bhroam: Spell out 'True'.

test/tests/functional/pbs_snapshot_unittest.py
```diff
@@ -798,6 +798,109 @@ def test_remote_primary_multinode(self):
...
+             ATTR_acluser: TEST_USER}
```
bhroam: I think you need to str() TSTGRP0 and TEST_USER

test/fw/bin/pbs_snapshot
```diff
@@ -171,6 +171,38 @@ def remotesnap_thread(logger, host):
     du.rm(hostname=host, path=snap_home, recursive=True, force=True)
 
 
+def capture_local_snap():
+    """
+    Helper method to capture snapshot of the local host
+
+    :returns Name of the snapshot directory/tar file captured
```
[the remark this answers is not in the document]
agrawalravi90: I'm not sure what the Doxygen convention is. @hirenvadalia  or @borlesanket do you guys know?

test/fw/ptl/utils/pbs_snaputils.py
agrawalravi90: I think I had issues with word boundaries when trying out re.sub(), sed's \b handles things like "key", "key=value", "key: value", "key1, key2, key3", "key1 key2 key3" etc., I don't remember exactly but I was having trouble doing the same with re.sub(). I'm not very good with regex, so please do let me know if you can think of a regex that will work here.

test/fw/ptl/utils/pbs_snaputils.py
```diff
@@ -212,11 +216,307 @@
...
+        # We will later do a sed on the whole snapshot, that's when these
+        # will get obfuscated
+        custom_rscs = []
+        custrscs_path = os.path.join(snap_dir, "server_priv", "resourcedef")
+        with open(custrscs_path, "r") as fd:
+            for line in fd:
+                try:
+                    rscs_name = line.split(" ", 1)[0]
+                except IndexError:
+                    continue
```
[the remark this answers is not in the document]
agrawalravi90: Ya, I'm not sure why i did that, I'll try to remove it and see if everything still works fine.

test/tests/functional/pbs_snapshot_unittest.py
agrawalravi90: manager() does it already, that's why some tests pass in int values when setting ncpus on vnodes.

test/fw/ptl/utils/pbs_snaputils.py
agrawalravi90: One more thing: sed can also obfuscate binary files:

[root@pbspro jobs]# grep "pbspro" 2017.pbspro.JB
Binary file 2017.pbspro.JB matches
[root@pbspro jobs]# sed -i 's/pbspro/blah/g' 2017.pbspro.JB 
[root@pbspro jobs]# grep "pbspro" 2017.pbspro.JB
[root@pbspro jobs]# /opt/pbs/bin/printjob 2017.pbspro.JB
---------------------------------------------------
jobid:	2017.blah
---------------------------------------------------
state:		0x4
substate:	0x2a (42)
svrflgs:	0x1 (1)
ordering:	0
inter prior:	0
stime:		1556891829
file base:	17.blah
queue:		
--bad union type -1408172029
--attributes--
short read of attribute

agrawalravi90:
@bhroam I've addressed your comments, please provide further feedback.
@arungrover request you to look at the new changes as well. Thanks!

test/fw/ptl/utils/pbs_snaputils.py
bhroam: I'm not so sure you want to run sed on a binary file.  According to this stack overflow article I just read it is meant for text.  It might not work properly on binary files even if there is valid text within the files.  You don't know what other non-ascii characters are in there.  Even if we do go the sed route, I wouldn't run it on the .JB files.  I'd run printjob on each .JB file and obfuscate that output.  I still think we should use re.sub().  I just checked and python regex has \\b.

agrawalravi90: I had tried \b with Python regex but wasn't happy with it, but I tried it again and it seems to be working fine, so I'll go ahead with it. And I'll take your and @scc138 's advice on how to deal with .JB files, I will take printjob outputs for them, and delete the binaries.

agrawalravi90:
@bhroam I've addressed your comments, please let me know what you think.
@arungrover request you to review the latest commit as well, I changed the approach from using sed to re.sub()

test/fw/ptl/utils/pbs_snaputils.py
```diff
@@ -239,7 +241,7 @@ class ObfuscateSnapshot(object):
                       ATTR_rescavail + ".vnode"]
     sched_attrs_obf = [ATTR_SchedHost]
     queue_attrs_obf = [ATTR_acluser, ATTR_aclgroup, ATTR_aclhost]
-    skip_vals = ["_pbs_project_default"]
+    skip_vals = ["_pbs_project_default", "*"]
```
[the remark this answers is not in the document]
agrawalravi90: I had to add this to skip over obfuscating the "*" from values like "pbsuser@*" for 'managers' attribute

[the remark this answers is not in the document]
bhroam: Ahh, I understand now.  That's a literal '*'.

bhroam:
Travis is failing due to pep8 issues.  Fix that and we should be good.

agrawalravi90:
@bhroam thanks, I've fixed pep8 issues, please merge this if all looks ok.

agrawalravi90:
Thanks @arungrover , can you please merge it?

arungrover:
I can merge it, I thought @bhroam might want to have a look again after you resolved pep8 problems.

bhroam:
I'm good with the changes.

PR #1100, opened 2019-04-23
revert_pbsconf() need to be updated to wait for pbs daemons to be up and ready

shilpakodli:
#### Describe Bug or Feature
PTL test cases sometimes used to fail with error "'Connection refused', 'qmgr: cannot connect to server '" due to server was not up when pbs.conf file is reverted and restarted all the daemons.
revert_pbsconf() reverts pbs.conf file and restarts the daemons if any changes are made in pbs.conf file.

#### Describe Your Change
Wait for all the daemons to be up and running after PBS restart.

### DIFF
--- a/test/fw/ptl/utils/pbs_testsuite.py
+++ b/test/fw/ptl/utils/pbs_testsuite.py
@@ -1168,15 +1168,39 @@ class PBSTestSuite(unittest.TestCase):
                 if restart_pbs:
                     # Restart all
                     server.pi.restart(server.hostname)
-                    if not server.isUp():
-                        self.fail("Server is not up")
-                    if not self.mom.isUp():
-                        self.fail("Mom is not up")
-                    if not self.scheduler.isUp():
-                        self.fail("Scheduler is not up")
-                    if not self.comm.isUp():
-                        self.fail("comm is not up")
-
+                    rv = server.isUp()
+                    if not rv:
+                        self.logger.error('Server '
+                                          + comm.hostname
+                                          + ' is down')
+                        msg = 'Failed to restart server '
+                        msg += server.hostname
+                        self.assertTrue(rv, msg)
+                    if new_pbsconf["PBS_START_MOM"] == "1":
+                        rv = server.moms.values()[0].isUp()
+                        if not rv:
+                            self.logger.error('mom '
+                                              + server.moms.values()[0]
+                                              + ' is down')
+                            msg = 'Failed to restart mom '
+                            msg += server.moms.values()[0]
+                            self.assertTrue(rv, msg)
+                    rv = server.schedulers['default'].isUp()
+                    if not rv:
+                        self.logger.error('Scheduler is down')
+                        msg = 'Failed to restart scheduler '
+                        msg += server.scheduler.hostname
+                        self.assertTrue(rv, msg)
+                    if self.du.is_localhost(server.hostname):
+                        if new_pbsconf["PBS_START_COMM"] == "1":
+                            rv = self.comm.isUp()
+                            if not rv:
+                                self.logger.error('comm '
+                                                  + self.comm.hostname
+                                                  + ' is down')
+                                msg = 'Failed to restart comm '
+                                msg += self.comm.hostname
+                                self.assertTrue(rv, msg)
                 else:
                     for initcmd in cmds_to_exec:
                         # start/stop the particular daemon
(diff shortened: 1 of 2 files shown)


test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -1169,6 +1171,7 @@ def _revert_pbsconf_server(self, vals_to_set):
                         # start/stop the particular daemon
                         server.pi.initd(server.hostname, initcmd[1],
                                         daemon=initcmd[0])
+                server.isUP()
```
kjakkali: On server node along with server we need to wait for other daemons to up and ready depending action.

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -989,6 +989,7 @@ def _revert_pbsconf_comm(self, primary_server, vals_to_set):
                 self.du.set_pbs_config(comm.hostname, confs=new_pbsconf)
                 comm.pbs_conf = new_pbsconf
                 comm.pi.initd(comm.hostname, "restart", daemon="comm")
+                comm.isUp()
```
agrawalravi90: I think you should do an assertTrue to make sure that the daemon is up, same comment for the rest of the daemons.

agrawalravi90: I think you should do an assertTrue to make sure that the daemon is up, same comment for the rest of the daemons.

test/fw/ptl/utils/pbs_testsuite.py
agrawalravi90: This function can restart multiple daemons, you need to check for each one depending on what's being restarted. 

Also, typo, server.isUP() should be server.isUp()

shilpakodli:
@agrawalravi90 and @kjakkali I have addressed you review comments, please let me know if any changes need to be done.

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -1164,11 +1168,33 @@ def _revert_pbsconf_server(self, vals_to_set):
                 if restart_pbs:
                     # Restart all
                     server.pi.restart(server.hostname)
+                    if not server.isUp():
+                        self.fail("Server is not up")
+                    if not self.mom.isUp():
```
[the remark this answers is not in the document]
agrawalravi90: Also, you'll need to do server.mom.isUp(), server.scheduler.isUp() and server.comm.isUp()

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -1164,11 +1168,33 @@ def _revert_pbsconf_server(self, vals_to_set):
...
                     for initcmd in cmds_to_exec:
                         # start/stop the particular daemon
                         server.pi.initd(server.hostname, initcmd[1],
                                         daemon=initcmd[0])
+                        if initcmd[1] == "start":
+                            if initcmd[0] == "server":
+                                if not server.isUp():
+                                    self.fail("Server is not up")
+                            if initcmd[1] == "sched":
+                                if not self.sched.isUp():
```
agrawalravi90: same comment here, you'll need to do server.mom.isUp(), server.scheduler.isUp() and server.comm.isUp()

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -989,6 +989,8 @@ def _revert_pbsconf_comm(self, primary_server, vals_to_set):
                 self.du.set_pbs_config(comm.hostname, confs=new_pbsconf)
                 comm.pbs_conf = new_pbsconf
                 comm.pi.initd(comm.hostname, "restart", daemon="comm")
+                if not comm.isUp():
+                    self.fail("comm is not up")
```
kjakkali: instead of fail() Can you update it as follows

                rv = comm.isUp()
                if not rv:
                    self.logger.error('comm ' + comm.hostname + ' is down')
                    comm.start()
                    msg = 'Failed to restart comm ' + comm.hostname
                    self.assertTrue(comm.isUp(), msg)

Applies to other daemons.

agrawalravi90: but we already do a comm restart/start just before calling comm.isUp(), will it really help to try it again?

kjakkali: Nothing specific just wanted to retry before exiting.

agrawalravi90: I'm not sure if running the same command again will help, but I think it might help if we modify isUp to wait and poll for a few times before giving up. Mom's isUp already does this:

        for _ in range(max_attempts):
            rv = super(MoM, self)._isUp(self)
            if rv:
                break
            time.sleep(1)

If this sounds like a good idea for all isUp functions, then maybe we should add this to __isUp() itself

shilpakodli:
@agrawalravi90 and @kjakkali I have addressed your review comments. Please review the same.

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -1168,33 +1168,80 @@ def _revert_pbsconf_server(self, vals_to_set):
...
-                        self.fail("comm is not up")
-
+                    rv = server.isUp()
+                    if not rv:
+                        self.logger.error('Server '
+                                          + comm.hostname
+                                          + ' is down')
+                        msg = 'Failed to restart server '
+                        msg += server.hostname
+                        self.assertTrue(rv, msg)
```
agrawalravi90: Instead of this, please use self.fail() instead, same comment for other instances

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -1168,33 +1168,80 @@ def _revert_pbsconf_server(self, vals_to_set):
...
+                            msg = 'Failed to restart mom '
+                            msg += server.moms.values()[0]
+                            self.assertTrue(rv, msg)
+                    rv = server.schedulers['default'].isUp()
+                    if not rv:
+                        self.logger.error('Scheduler is down')
+                        msg = 'Failed to restart scheduler '
+                        msg += server.scheduler.hostname
+                        self.assertTrue(rv, msg)
+                    if self.du.is_localhost(server.hostname):
```
agrawalravi90: Instead of this, it might be better to just add 'comms' to the server object, then you'll be able to call isUp() on all server daemons, just like mom and scheduler. You'll just need to add cls.server.comms = cls.comms at the end in PBSTestSuite.init_comms()

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -1168,33 +1168,80 @@ def _revert_pbsconf_server(self, vals_to_set):
...
+                                        self.logger.error('mom '
+                                                          + server.mom.hostname
+                                                          + ' is down')
+                                        msg = 'Failed to restart mom '
+                                        msg += server.mom.hostname
+                                        self.assertTrue(rv, msg)
                             if initcmd[0] == "comm":
-                                if not self.comm.isUp():
-                                    self.fail("comm is not running")
+                                if self.du.is_localhost(server.hostname):
```
agrawalravi90: This is up to you, but i feel like we could do some refactoring here to reuse a lot of the code. Suggestion: create a function called _check_daemon_on_server(server_obj, daemon_name), which will check the daemon name and call isUp on sched, mom or comm and print the error message as relevant. I think this method will shrink down a lot if we do that.

shilpakodli:
@agrawalravi90 I have addressed your review comments. Please review it.

shilpakodli:
@agrawalravi90 and @kjakkali Thanks for the review. I will run full regression and update the same.

shilpakodli:
@kjakkali and @agrawalravi90 I have run full regression on two testbed machines and I don't see any failures related to my changes also I have run multinode where it takes two servers, a client and another test case with multimom. Both test cases ran successfully.

PR #1101, opened 2019-04-23
Sched mem leak when estimating for top job fails

prakashcv13:
**Problem**
Scheduler leaks memory when estimating for top job fails - 
==24836== 78,274 (2,928 direct, 75,346 indirect) bytes in 6 blocks are definitely lost in loss record 1,723 of 1,724
==24836== at 0x4C27BE3: malloc (vg_replace_malloc.c:299)
==24836== by 0x459043: new_server_info (server_info.c:1244)
==24836== by 0x45AFCF: dup_server_info (server_info.c:2235)
==24836== by 0x4207D0: add_job_to_calendar (fifo.c:1905)
==24836== by 0x41EB3F: main_sched_loop (fifo.c:963)
==24836== by 0x41E57A: scheduling_cycle (fifo.c:742)
==24836== by 0x41E20C: intermediate_schedule (fifo.c:614)
==24836== by 0x41E0CB: schedule (fifo.c:559)
==24836== by 0x41CFD8: main (pbs_sched.c:1455)

Steps to reproduce:
Execute the below PTL tests -
TestCalendaring.test_topjob_fail
from pbs_multi_sched.py

A return statement was added through #1087 which caused this leak.

**Solution**
Free the duped server_info before returning from add_job_to_calendar().
Once this is done - we see another mem leak being reported - 
==395== 3,630 (32 direct, 3,598 indirect) bytes in 1 blocks are definitely lost in loss record 1,568 of 1,595
==395==    at 0x4C27BE3: malloc (vg_replace_malloc.c:299)
==395==    by 0x435644: query_nodes (node_info.c:197)
==395==    by 0x456813: query_server (server_info.c:267)
==395==    by 0x41E2AF: scheduling_cycle (fifo.c:676)
==395==    by 0x41E20C: intermediate_schedule (fifo.c:614)
==395==    by 0x41E0CB: schedule (fifo.c:559)
==395==    by 0x41CFD8: main (pbs_sched.c:1455)

For this - we need to pass 1 to free_server when query_queues() fail.

**Logs**
[val_sched_final.txt](https://github.com/PBSPro/pbspro/files/3107072/val_sched_final.txt)
[val_sched_no_changes.txt](https://github.com/PBSPro/pbspro/files/3107073/val_sched_no_changes.txt)
[val_sched_one_change.txt](https://github.com/PBSPro/pbspro/files/3107074/val_sched_one_change.txt)

### DIFF
--- a/src/scheduler/server_info.c
+++ b/src/scheduler/server_info.c
@@ -280,7 +280,7 @@ query_server(status *pol, int pbs_sd)
 	if ((sinfo->queues = query_queues(policy, pbs_sd, sinfo)) == NULL) {
 		pbs_statfree(server);
 		sinfo->fairshare = NULL;
-		free_server(sinfo, 0);
+		free_server(sinfo, 1);
 		return NULL;
 	}
 
--- a/src/scheduler/server_info.c
+++ b/src/scheduler/server_info.c
@@ -222,7 +222,7 @@ query_server(status *pol, int pbs_sd)
 	if( query_server_dyn_res(sinfo) == -1 ) {
 		pbs_statfree(server);
 		sinfo -> fairshare = NULL;
-		free_server( sinfo, 0 );
+		free_server(sinfo);
 		return NULL;
 	}
 
@@ -239,7 +239,7 @@ query_server(status *pol, int pbs_sd)
 		pbs_statfree(server);
 		pbs_statfree(all_sched);
 		sinfo->fairshare = NULL;
-		free_server(sinfo, 0);
+		free_server(sinfo);
 		return NULL;
 	}
 	query_sched_obj(policy, sched, sinfo);
@@ -250,7 +250,7 @@ query_server(status *pol, int pbs_sd)
 		schdlog(PBSEVENT_SCHED, PBS_EVENTCLASS_SERVER, LOG_ERR, __func__, log_buffer);
 		pbs_statfree(server);
 		sinfo->fairshare = NULL;
-		free_server(sinfo, 0);
+		free_server(sinfo);
 		return NULL;
 	}
 
@@ -267,7 +267,7 @@ query_server(status *pol, int pbs_sd)
 	if ((sinfo->nodes = query_nodes(pbs_sd, sinfo)) == NULL) {
 		pbs_statfree(server);
 		sinfo->fairshare = NULL;
-		free_server(sinfo, 0);
+		free_server(sinfo);
 		return NULL;
 	}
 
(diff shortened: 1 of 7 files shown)


src/scheduler/server_info.c
```diff
@@ -280,7 +280,7 @@ query_server(status *pol, int pbs_sd)
 	if ((sinfo->queues = query_queues(policy, pbs_sd, sinfo)) == NULL) {
 		pbs_statfree(server);
 		sinfo->fairshare = NULL;
-		free_server(sinfo, 0);
+		free_server(sinfo, 1);
```
agrawalravi90: Not sure if there was a particular reason why this was called with a 0, but if we are changing this, then shouldn't we also change the other free_server() calls in this function?

prakashcv13: the subsequent calls are already using 1, instead of 0.
The calls before this need not change as when called with 1 - free_server() would free queues, nodes and resource_resv_array. Any of these have not been allocated until this point.

prakashcv13:
thank you @agrawalravi90 for the review and approval :)

src/scheduler/server_info.c
arungrover: I'll suggest to just remove the option to free only certain things in free_server()
It does not make sense to free only partial things. I understand that free will potentially be called on unallocated things but if they are initialized properly (with NULL) it should not be a problem. This will also reduce the possibility of future problems with free_server()

prakashcv13: I thought about the same - but chose to be lazy and do it if this is suggested as a review comment :)

arungrover: Well now you have it 😄

agrawalravi90:
There must be a reason why that flag exists, before we remove it, I think we should check with @bhroam once.

arungrover:
@agrawalravi90 Bhroam and I had a chat about removing flag from free_server() sometime back but we didn't get a chance to fix it. Now that we are making the change in the same part of the code, I think Bhroam will be ok removing the flag.

prakashcv13:
hey @arungrover - I have removed the flag from free_server() and free_queues() as well.
As discussed offline we decided to remove the flag from free_queues() as it was being called only with a 1 and never with a 0.

prakashcv13:
@arungrover - I have run the same test with scheduler under valgrind.
[val_sched_ac2.txt](https://github.com/PBSPro/pbspro/files/3108944/val_sched_ac2.txt).
Please let me know if there is anything specific you would like me to run and test.

agrawalravi90:
Ok, sounds good, thanks for clarifying @arungrover 

Changes still look fine to me.

prakashcv13:
@arungrover - I executed the whole TestCalendaring test suite. Attached are the PTL and valgrind logs.
[val_sched_ac3.txt](https://github.com/PBSPro/pbspro/files/3108993/val_sched_ac3.txt)
[TestCalendering.txt](https://github.com/PBSPro/pbspro/files/3109003/TestCalendering.txt)

arungrover:
Thanks @prakashcv13, I'll merge this in

prakashcv13:
Thank you @agrawalravi90 and @arungrover.

PR #1102, opened 2019-04-23
Remove unnecessary setting of managers attribute in cleanup_jobs()

latha-subramanian:
#### Describe Bug or Feature
cleanup_jobs() in pbs_testlib.py sets the 'manager' attribute so the current user is a manager in order to set 'scheduling' to True or False. This is unnecessary as the server.manager() call itself can achieve this when passing 'root' user as the 'runas' argument.

#### Describe Your Change
Remove code related to setting manager attribute in cleanup_jobs().  Set 'scheduling' to True/False via server.manager() with 'runas' argument as 'root'

#### Link to Design Doc
NA

[test_insufficient_resc_non_cons.txt](https://github.com/PBSPro/pbspro/files/3109648/test_insufficient_resc_non_cons.txt)

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -8334,21 +8334,15 @@ class Server(PBSService):
             select_xt = 'x'
         job_ids = self.select(extend=select_xt)
 
-        # Make sure the current user is a manager. Some tests might have
-        # unset the mangers attribute. Ignore 'Duplicate entry in the list'
-        # error if the current user was already a manager
-        current_user = pwd.getpwuid(os.getuid())[0]
-        a = {ATTR_managers: (INCR, current_user + '@*')}
-        try:
-            self.manager(MGR_CMD_SET, SERVER, a, sudo=True)
-        except PbsManagerError:
-            pass
-
         # Turn off scheduling so jobs don't start when trying to
         # delete. Restore the orignial scheduling state
         # once jobs are deleted.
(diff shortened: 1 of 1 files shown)


latha-subramanian:
@hirenvadalia , Requesting your review

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -8334,21 +8334,15 @@ def cleanup_jobs(self, extend=None, runas=None):
...
-
         # Turn off scheduling so jobs don't start when trying to
         # delete. Restore the orignial scheduling state
         # once jobs are deleted.
         sched_state = self.status(SERVER, {'scheduling'})[0]['scheduling']
-        self.manager(MGR_CMD_SET, SERVER, {'scheduling': 'False'})
+
+        # runas is required here because some tests remove current user
+        # from managers list
+        self.manager(MGR_CMD_SET, SERVER, {'scheduling': 'False'},
```
visheshh: Hi @latha-subramanian , In case multiple schedulers are present, we should be setting false for all the schedulers instead on server which will only set False on default scheduler.

latha-subramanian: @visheshh , I have made changes such that it works for multischeds too. Please take a look

visheshh: @latha-subramanian Thanks for making changes.

prakashcv13:
merged #1102

PR #1104, opened 2019-04-25
qsub -a argument causes duplicate job entry in datastore in 18.x or before and gets rejected in 19.x

Bhagat-Rajput:
#### Describe Bug or Feature

* job submitted with qsub -a argument in the past causes duplicate job and job_attr rows in datastore and ultimately server crash in 18.x and before, causes job to be rejected in 19.x
* In 18.x PBS saves the job in job_wait_over as a new job and later again save the same job through job_or_resv_save as a fresh job resulting duplication in datastore happens.
* In 19.x this thing is not happening because jobid has been guarded with primary key so PBS doesn't accept the job.

**Steps to Reproduce**:(18.x or older)
1. submit a job with past time using qsub 'a' option.
   e.g. 
   [root@centos7_2 PBSPro_18.2.3]# date
   Thu Apr 25 16:11:59 IST 2019
    [root@centos7_2 PBSPro_18.2.3]# qsub -a 201904251500 -- /bin/sleep 100
    0.centos7_2
2. Restart the PBS complex (/etc/init.d/pbs restart)
3. Run qstat and can see two jobs with the same name.
4. qdel \<jobid\>, server will be crashed here. 

#### Describe Your Change

* PBS will use job_or_resv_save() only to save a new job.
* If PBS tries to save a same job twice as a new job then PBS will dump backtrace in server logs.

[ptl_job_array.txt](https://github.com/PBSPro/pbspro/files/3161766/ptl_job_array.txt)
[ptl_qsub_opt_args.txt](https://github.com/PBSPro/pbspro/files/3161767/ptl_qsub_opt_args.txt)
valgrind report: [server_multi_job_save_val.txt](https://github.com/PBSPro/pbspro/files/3171794/server_multi_job_save_val.txt)

### DIFF
--- a/src/server/array_func.c
+++ b/src/server/array_func.c
@@ -856,6 +856,7 @@ create_subjob(job *parent, char *newjid, int *rc)
 		*rc = PBSE_SYSTEM;
 		return NULL;
 	}
+	subj->ji_newjob = 1; /* help to save in db as a new job */
 	subj->ji_qs = parent->ji_qs;	/* copy the fixed save area */
 	parent->ji_ajtrk->tkm_tbl[indx].trk_psubjob = subj;
 	subj->ji_qhdr     = parent->ji_qhdr;
@@ -904,8 +905,6 @@ create_subjob(job *parent, char *newjid, int *rc)
 
 	subj->ji_qs.ji_svrflags &= ~JOB_SVFLG_ArrayJob;
 	subj->ji_qs.ji_svrflags |=  JOB_SVFLG_SubJob;
-	subj->ji_modified = 0;  /* to avoid db save in svr_setjobstate()*/
-	subj->ji_newjob = 1;    /* Hack to use ji_newjob to mean SAVEJOB_NEW for subjobs */
 	subj->ji_qs.ji_substate = JOB_SUBSTATE_TRANSICM;
 	(void)svr_setjobstate(subj, JOB_STATE_QUEUED, JOB_SUBSTATE_QUEUED);
 	subj->ji_modified = 1;  /* to force a SAVEJOB_NEW db save in svr_setjobstate() for subjobs */
--- a/src/server/array_func.c
+++ b/src/server/array_func.c
@@ -856,7 +856,7 @@ create_subjob(job *parent, char *newjid, int *rc)
 		*rc = PBSE_SYSTEM;
 		return NULL;
 	}
-	subj->ji_newjob = 1; /* help to save in db as a new job */
+	subj->ji_newjob = 1; /* flag to indicate a new job is being added to the db */
 	subj->ji_qs = parent->ji_qs;	/* copy the fixed save area */
 	parent->ji_ajtrk->tkm_tbl[indx].trk_psubjob = subj;
 	subj->ji_qhdr     = parent->ji_qhdr;
--- a/src/server/job_recov_db.c
+++ b/src/server/job_recov_db.c
@@ -383,10 +383,9 @@ job_save_db(job *pjob, int updatetype)
 		return (0);
 
 	if (updatetype == SAVEJOB_NEW && pjob->ji_newjob == 0) {
-		/* If new_job flag is not set but still try to save as a new job, changed to SAVEJOB_FULL*/
 		updatetype = SAVEJOB_FULL;
-		sprintf(log_buffer,"job has already been saved earlier as a new job, thus changing to SAVEJOB_FULL");
-		log_joberr(-1, "job_save_db", log_buffer,  pjob->ji_qs.ji_jobid);
+		sprintf(log_buffer, "job already saved to db, so changing updatetype to SAVEJOB_FULL");
+		log_joberr(-1, "job_save_db", log_buffer, pjob->ji_qs.ji_jobid);
 #ifndef WIN32
 		print_backtrace(pjob->ji_qs.ji_jobid);
 #endif
@@ -819,15 +818,15 @@ print_backtrace(char *jobid) {
 	total_frames = backtrace(bt_buffer, BACKTRACE_BUF_SIZE);
 	if (total_frames != 0) {
 		log_err(-1, "job_save_db", "----- BACKTRACE START -----");
-		sprintf(log_buffer,"backtrace() has returned %d addresses",total_frames);
-		log_joberr(-1, "job_save_db", log_buffer,  jobid);
+		sprintf(log_buffer, "backtrace() has returned %d addresses", total_frames);
+		log_joberr(-1, "job_save_db", log_buffer, jobid);
 		/* backtrace_symbols() resolve addresses into strings containing "filename(function+address)",
 		 * this array must be free()-ed at the end.
 		 */
 		bt_strings = backtrace_symbols(bt_buffer, total_frames);
 		if (bt_strings == NULL) {
-			sprintf(log_buffer,"no backtrace_symbols() found");
-			log_joberr(-1, "job_save_db", log_buffer,  jobid);
+			sprintf(log_buffer, "no backtrace symbols found");
+			log_joberr(-1, "job_save_db", log_buffer, jobid);
 		} else {
 			for (frame_num = 0; frame_num < total_frames; frame_num++) {
 				snprintf(log_buffer, LOG_BUF_SIZE-1, "%s", bt_strings[frame_num]);
@@ -837,8 +836,8 @@ print_backtrace(char *jobid) {
 		free(bt_strings);
 		log_err(-1, "job_save_db", "----- BACKTRACE END -----");
 	} else {
-		sprintf(log_buffer,"backtrace() has not returned any address, might be corrupted");
-		log_joberr(-1, "job_save_db", log_buffer,  jobid);
+		sprintf(log_buffer, "No backtrace symbols present");
+		log_joberr(-1, "job_save_db", log_buffer, jobid);
 	}
 #endif /*-- GLIBC --*/
 #endif /*-- GNUC  --*/
--- a/src/server/job_recov_db.c
+++ b/src/server/job_recov_db.c
@@ -57,6 +57,7 @@
  *	db_to_svr_resv		  -	Load data from database resv object to a server resv object
  *	resv_save_db		  -	Save resv to database
  *	resv_recov_db		  - Recover resv from database
+ *	print_backtrace		  - Print backtrace of the program
  *
  */
 
@@ -106,6 +107,14 @@
 
 #ifndef PBS_MOM
 extern pbs_db_conn_t	*svr_db_conn;
+#ifndef WIN32
+#ifdef __GNUC__
+/* backtrace() functionality provided in glibc since from version 2.1 */
+#if (__GLIBC__ >= 2 && __GLIBC_MINOR__ >= 1)
+void print_backtrace(char *);
+#endif
+#endif
+#endif
 #endif
 
 #ifdef NAS /* localmod 005 */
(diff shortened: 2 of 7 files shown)


src/server/job_recov_db.c
```diff
@@ -376,6 +378,46 @@ job_save_db(job *pjob, int updatetype)
...
+		log_joberr(-1, "job_save_db", log_buffer,  pjob->ji_qs.ji_jobid);
+#ifndef WIN32
+		/* let's print the backtrace to identify the faulty scenario */
+		int total_frames; /* total number of frames returned by backtrace() */
+		int frame_num;
+		void *bt_buffer[BACKTRACE_BUF_SIZE];
+		char **bt_strings;
+
+		/* backtrace() returns current stack addresses */
+		total_frames = backtrace(bt_buffer, BACKTRACE_BUF_SIZE);
```
bhroam: Why are we dumping a backtrace?  There are a lot of error conditions in PBS, but we don't log backtraces.  What makes this one so special?  Also, this is a GNU extension, so we shouldn't use it unless it is correctly guarded.

test/tests/functional/pbs_qsub_opts_args.py
```diff
@@ -126,3 +126,37 @@ def test_qsub_with_option_script(self):
...
+        """
+        self.server.manager(MGR_CMD_SET, NODE,
+                            {'resources_available.ncpus': 2},
+                            self.mom.shortname)
+        # submit a job with future time and should start whenever time hits
+        present_tm = int(time.time())
+        future_tm = time.strftime(
+            "%Y%m%d%H%M", time.localtime(
+                present_tm + 60))
+        qsub_cmd = [self.qsub_cmd, '-a', future_tm, '--', '/bin/sleep', '100']
```
bhroam: Why run qsub directly?  Why not use PTL's submit() function to submit a job?

src/server/job_recov_db.c
Bhagat-Rajput: Thanks for the review @bhroam.
Trying to identify a scenario in which PBS tries to save the already saved job again as a new job which causes panic db shutdown. We had identified one and fixed but there might be others so with the help of backtrace will find and fix it without any panic shutdown.
PBS is already using backtrace functionality in qsub command without any special guard and it seems working fine.

prakashcv13: @Bhagat-Rajput - backtrace in qsub.c is guarded with BACKTRACE_SIZE.

Bhagat-Rajput: @prakashcv13, I didn't noticed that macro BACKTRACE_SIZE is permanently commented and that may be the reason of qsub to work properly.
Do we need to guard the same here ?

prakashcv13: @Bhagat-Rajput - backtrace() is a good GNU extension that helps in debugging the issue at hand, but this can be added/removed as and when needed. As @bhroam mentioned there are a lot of error conditions, but we do not log backtraces for all of them. It is a debugging tool which a developer can use for debugging, but need not be part of the "committed" code. So, I would suggest removing the backtrace() code.

subhasisb: @Bhagat-Rajput i agree with Bhroam that we should guard it. Please use a ifdef for __GNUC__ before using this macro. As to why this is required: The current fix is based on our finding of one specific code flow path that can cause this. There could be more, but it is difficult to find out the flow, and should we hit another flow, this trace will absolutely tell us. Thus this is valuable. We have precedence of adding such things to the code when we had the need to debug or troubleshoot. Recall the famous attributes getting deleted from database after server restart. That code absolutely benefited from dumping data like this.

test/tests/functional/pbs_qsub_opts_args.py
```diff
@@ -135,28 +135,41 @@ def test_qsub_with_option_a(self):
...
-        qsub_cmd = [self.qsub_cmd, '-a', future_tm, '--', '/bin/sleep', '100']
-        rv = self.du.run_cmd(self.server.hostname, cmd=qsub_cmd)
-        if rv['rc'] != 0:
-            self.fail("ERROR in submitting a job with future time: %s" % rv)
-        self.server.expect(JOB, {'job_state': 'W'}, id=str(rv['out'][0]))
+        j1 = Job(TEST_USER, {ATTR_a: future_tm})
+        try:
+            jid_1 = self.server.submit(j1)
+        except PbsSubmitError as e:
+            if e.rc == 57:
```
bhroam: I believe the fact that e.rc == 57 is a bug in PTL.  I wouldn't trust that it will always remain so.  @hirenvadalia can confirm this.

hirenvadalia: @bhroam I don't think it is a bug in PTL because error code 57 is returned by qsub (as per the chat with @Bhagat-Rajput). So PTL is simply returning error code returned by qsub. However, first we need to find why qsub is returning 57 and what is the meaning of 57? @Bhagat-Rajput can you please find out why qsub is returning 57 and what is the meaning of that?

Bhagat-Rajput: Had a chat with @suresh-thelkar we came to know that the no. 57 is the result of 15161%256. As @bhroam said we also believe that these things should be handle at framework level.

bhroam: The framework can't handle this.  A mod operator is one way.  The framework can't take 57 and figure out it was 15161.  How about adding a comment in the test explaining what 57 means?

suresh-thelkar: As per my understanding, if PTL is using PTL_API mode then we should get 15161 error code only(The value should not be masked). If PTL is using PTL_CLI then we should do the mod operator as @bhroam suggested.

Bhagat-Rajput: As of now @suresh-thelkar PTL_API mode is not working due to some bugs so we cannot check that mode. I'll fix the PTL using one of the solution as @bhroam suggested.

Bhagat-Rajput:
I have fixed the PTL testcase and uploaded the latest test results. Please review it.

test/tests/functional/pbs_qsub_opts_args.py
```diff
@@ -54,6 +54,23 @@ def setUp(self):
...
+        ret_msg = 'qsub: Failed to save job/resv, '\
+            'refer server logs for details'
+        # PBS returns 15161 error code when it fails to save the job in db
+        # but in PTL_CLI mode it returns modulo of the error code.
+
+        # for PTL_API mode
+        if e.rc == 15161:
+            self.assertFalse(e.msg[0], ret_msg)
+        # for PTL_CLI mode
+        elif 15161 % 256 == e.rc:
```
bhroam: Why not combine these into one if statement with an 'or' between them?  Also, swap the two sides of the '==' to be consistent with the above check.

src/server/req_runjob.c
```diff
@@ -613,6 +613,18 @@ req_runjob2(struct batch_request *preq, job *pjob)
 	/* Check if prov is required, if so, reply_ack and let prov finish */
 	/* else follow normal flow */
 	prov_rc = check_and_provision_job(preq, pjob, &need_prov);
+
+	/* in case of subjob, it was never saved to the database so far.
+	 * Save it now, before a possiblity to return from the routine
+	 */
+	if (pjob->ji_qs.ji_svrflags & JOB_SVFLG_SubJob) {
+		if (job_save(pjob, SAVEJOB_NEW)) {
```
Shrini-h: @Bhagat-Rajput  Will it be better if we avoid `job_save(subjob)` when `check_and_provision_job()` fails ? If you agree you can move the `if (pjob->ji_qs.ji_svrflags & JOB_SVFLG_SubJob) {...}` block to location after `if (prov_rc) {...}` block below

src/server/job_recov_db.c
```diff
@@ -376,6 +378,52 @@ job_save_db(job *pjob, int updatetype)
 	if (pjob->ji_newjob == 1 && updatetype != SAVEJOB_NEW)
 		return (0);
 
+	if (updatetype == SAVEJOB_NEW && pjob->ji_newjob == 0) {
+		/* If new_job flag is not set but still try to save as a new job, changed to SAVEJOB_FULL*/
+		updatetype = SAVEJOB_FULL;
+		sprintf(log_buffer,"job has already been saved earlier as a new job, thus changing to SAVEJOB_FULL");
+		log_joberr(-1, "job_save_db", log_buffer,  pjob->ji_qs.ji_jobid);
+
+#ifndef WIN32
```
Shrini-h: I would recommend moving this to a new utility function. So that it can be handy for future use.

Bhagat-Rajput:
Thanks for the review @Shrini-h, I have addressed your comments. Please review it.

src/server/job_recov_db.c
```diff
@@ -106,6 +107,14 @@
 
 #ifndef PBS_MOM
 extern pbs_db_conn_t	*svr_db_conn;
+#ifndef WIN32
```
Shrini-h: Would be tidier if you move the conditional compilation directives inside the body of `print_backtrace()`

src/server/job_recov_db.c
```diff
@@ -388,40 +397,10 @@ job_save_db(job *pjob, int updatetype)
...
-			free(bt_strings);
-			log_err(-1, "job_save_db", "----- BACKTRACE END -----");
-		} else {
-			sprintf(log_buffer,"backtrace() has not returned any address, might be corrupted");
-			log_joberr(-1, "job_save_db", log_buffer,  pjob->ji_qs.ji_jobid);
-		}
-#endif /*-- GLIBC --*/
-#endif /*-- GNUC --*/
-#endif /*-- WIN32 --*/
+		print_backtrace(pjob->ji_qs.ji_jobid);
```
Shrini-h: Would be tidier if you move the conditional compilation directives inside the body of `print_backtrace()`

src/server/job_recov_db.c
```diff
@@ -825,4 +804,51 @@ job_or_resv_recov_db(char *id, int objtype)
...
+ *		Print backtrace of the program
+ *
+ * @see
+ * 		job_save_db
+ *
+ * @param[in]	jobid - Print call trace of the job
+ *
+ */
+void
+print_backtrace(char *jobid) {
```
Shrini-h: Would be good to move the implementation to an utility library. I wont mandate it but leave it to you and other reviewers/ maintainers

Bhagat-Rajput: @Shrini-h , I think for the time being it's good to be here only and however it's restricted to server only.

Bhagat-Rajput:
@PBSPro/pbspro-maintainers, PR is ready to merge.Thanks!

src/server/req_runjob.c
prakashcv13: @Bhagat-Rajput and @Shrini-h - wouldn't we still need the subjob to be saved to db, even if we could not start provisioning for the subjob?

Shrini-h: @prakashcv13  you are right, I mistook the free_nodes() as job_purge(), hence I suggested it.
@Bhagat-Rajput  please revert this change and verify once with provision fail test when Job Array is submitted

src/server/array_func.c
```diff
@@ -904,8 +905,6 @@ create_subjob(job *parent, char *newjid, int *rc)
 
 	subj->ji_qs.ji_svrflags &= ~JOB_SVFLG_ArrayJob;
 	subj->ji_qs.ji_svrflags |=  JOB_SVFLG_SubJob;
-	subj->ji_modified = 0;  /* to avoid db save in svr_setjobstate()*/
```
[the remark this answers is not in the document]
Bhagat-Rajput: @prakashcv13, removed because we don't want to save a new job through svr_setjobstate

src/server/job_recov_db.c
```diff
@@ -376,6 +382,16 @@ job_save_db(job *pjob, int updatetype)
 	if (pjob->ji_newjob == 1 && updatetype != SAVEJOB_NEW)
 		return (0);
 
+	if (updatetype == SAVEJOB_NEW && pjob->ji_newjob == 0) {
+		/* If new_job flag is not set but still try to save as a new job, changed to SAVEJOB_FULL*/
```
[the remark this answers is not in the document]
Bhagat-Rajput: I think it's fine, explaining the current block of code.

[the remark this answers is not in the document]
prakashcv13: the log message already explains what we are doing, so the comment is redundant, please remove.

Bhagat-Rajput:
@prakashcv13, I have addressed all the comments and @Shrini-h code has also reverted back. Please review it.

Bhagat-Rajput:
@prakashcv13, if there are no further comments then i appreciate your sign-off.

PR #1109, opened 2019-05-01
Sched crash when array job and reservation is submitted

arungrover:
#### Describe Bug or Feature
Scheduler crashes when it encounters an array job and a reservation in the same scheduling cycle.

#### Describe Your Change
As part of speeding up scheduler, we added a way to find a job/reservation based on their indexes in the all_resresv array.
But after these indexes are created, all_resresv array gets sorted to bring timed events in from in order to create a calendar. This sort messes up the indexes and causes the crash. While the sort is needed there is no reason to have all_resresv sorted. 
As part of the fix, create_events code which creates the calendar will create a local copy of the array and sort that local copy instead of sorting all_resresv array.

#### Link to Design Doc
None

[sched_dump_valgrind.txt](https://github.com/PBSPro/pbspro/files/3136152/sched_dump_valgrind.txt)
[test_out.txt](https://github.com/PBSPro/pbspro/files/3136153/test_out.txt)

### DIFF
--- a/src/scheduler/simulate.c
+++ b/src/scheduler/simulate.c
@@ -858,11 +858,21 @@ create_events(server_info *sinfo)
 	int		errflag = 0;
 	int		i = 0;
 	time_t 		end = 0;
+	resource_resv	**all_resresv_copy;
+	int		all_resresv_len;
 
-	/* all_resresv is sorted such that the timed events are in the front of
-	 * the array.  Once the first non-timed event is reached, we're done
+	/* create a temporary copy of all_resresv array which is sorted such that
+	 * the timed events are in the front of the array.
+	 * Once the first non-timed event is reached, we're done
 	 */
-	all = sinfo->all_resresv;
+	all_resresv_len = count_array((void **)sinfo->all_resresv);
+	all_resresv_copy = (resource_resv **)malloc((all_resresv_len + 1) * sizeof(resource_resv *));
+	if (all_resresv_copy == NULL)
+		return 0;
+	for (i = 0; sinfo->all_resresv[i] != NULL; i++)
+		all_resresv_copy[i] = sinfo->all_resresv[i];
+	all_resresv_copy[i] = NULL;
+	all = all_resresv_copy;
 
 	/* sort the all resersv list so all the timed events are in the front */
 	qsort(all, count_array((void **)all), sizeof(resource_resv *), cmp_events);
@@ -901,8 +911,10 @@ create_events(server_info *sinfo)
 		if (node->is_sleeping) {
 			te = create_event(TIMED_NODE_UP_EVENT, sinfo->server_time + PROVISION_DURATION,
 					(event_ptr_t *) node, (event_func_t) node_up_event, NULL);
-			if (te == NULL)
+			if (te == NULL) {
+				free(all_resresv_copy);
 				return 0;
+			}
 			events = add_timed_event(events, te);
 		}
 	}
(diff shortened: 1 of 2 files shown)


src/scheduler/simulate.c
```diff
@@ -858,11 +858,21 @@ create_events(server_info *sinfo)
...
 
-	/* all_resresv is sorted such that the timed events are in the front of
-	 * the array.  Once the first non-timed event is reached, we're done
+	/* create a temporary copy of all_resresv array which is sorted such that
+	 * the timed events are in the front of the array.
+	 * Once the first non-timed event is reached, we're done
 	 */
-	all = sinfo->all_resresv;
+	all_resresv_len = count_array((void **)sinfo->all_resresv);
+	all_resresv_copy = (resource_resv **)malloc((all_resresv_len + 1) * sizeof(resource_resv *));
```
bhroam: No need to cast the return value of malloc()

src/scheduler/simulate.c
```diff
@@ -901,18 +911,22 @@ create_events(server_info *sinfo)
 		if (node->is_sleeping) {
 			te = create_event(TIMED_NODE_UP_EVENT, sinfo->server_time + PROVISION_DURATION,
 					(event_ptr_t *) node, (event_func_t) node_up_event, NULL);
-			if (te == NULL)
+			if (te == NULL) {
+				free(all_resresv_copy);
 				return 0;
+			}
```
bhroam: I think this is actually a bug.  If we return here, we'll leak events.  Instead of returning, errflag++ and break (that is what happens above).  That will cause the code below to free everything and return.

test/tests/pbs_smoketest.py
```diff
@@ -79,15 +79,38 @@ def test_submit_job_array(self):
...
         """
-        Test to submit an advanced reservation
+        Test to submit an advanced reservation and submit jobs to that
+        reservation. Check if the reservation gets confimed and the jobs
+        inside the reservation starts running when the reservation runs.
         """
-        r = Reservation()
-        a = {'Resource_List.select': '1:ncpus=1'}
+        a = {'resources_available.ncpus': 4}
+        self.server.manager(MGR_CMD_SET, NODE, a, id=self.mom.shortname)
```
bhroam: If you do this, I think you need to add the skip on cpuset decorator to this test.  You can't modify the resources on a cpuset machine to differ from what the mom reported.

test/tests/pbs_smoketest.py
```diff
@@ -79,15 +79,38 @@ def test_submit_job_array(self):
...
+        r = Reservation(TEST_USER)
+        now = int(time.time())
+        a = {'Resource_List.select': '1:ncpus=4',
+             'reserve_start': now + 5,
```
bhroam: 5s might be too short on slow machines.  I'd use 10s.

test/tests/pbs_smoketest.py
```diff
@@ -79,15 +79,38 @@ def test_submit_job_array(self):
...
         self.server.expect(RESV, a, id=rid)
 
+        # submit a normal job and an array job to the reservation
+        a = {'Resource_List.select': '1:ncpus=1',
+             ATTR_q: rid.split('.')[0]}
+        j1 = Job(TEST_USER, attrs=a)
+        jid1 = self.server.submit(j1)
+
+        a = {'Resource_List.select': '1:ncpus=1',
+             ATTR_q: rid.split('.')[0], ATTR_J: '1-2'}
```
bhroam: Since you're splitting the rid twice, why not do it once and store it into a local variable?

arungrover:
Thanks for quick review @bhroam. I've addressed your review comments.

PR #1110, opened 2019-05-03
Measure hook performance infrastructure

bayucan:
#### Describe Bug or Feature

Put infrastructure in place so that we can accurately measure the time it takes a hook to execute.

#### Describe Your Change

* The amount of walltime and cputime that sections of PBS code in support of hooks are reported, via  server_logs messages for server hooks, and mom_logs messages for mom hooks, under the high log level of DEBUG4.
* In order to see these messages, set server log_event value to:
`# qmgr -c "set server log_events = 4095"` 
and set mom_priv/config value of $logevent to 4095:
```
 # cat /var/spool/pbs/mom_priv/config
 $logevent 4095
```
* The log messages are in the form of space space-separated token headed by the substring "hook_perf_stat;":
   * **basic log message:**
`"hook_perf_stat;label=<description_of_section_of_code> action=<action_of_section_of_code> walltime=<num_of_seconds> cputime=<num_of_seconds>"`
    * **header message**: marks the start of profiling of a particular hook:
`"hook_perf_stat;label=<description_of_section_of_code> action=<action_of_section_of_code> profile_start"`
    * **trailer log message:** shows the amount taken from when the hook first got loaded (marked by the header message) to when it actually executed and completed:
`"hook_perf_stat;label=<description_of_section_of_code> action=<action_of_section_of_code> walltime=<num_of_seconds> cputime=<num_of_seconds> profile_stop"`
    * **label value format**: <hook_event>_<hook_name>_<unique_identifier_like_object_id_or_pid>
    * **special label value: "hook_func"** means a function inside the hook script, like pbs.server(), has been profiled and this was the result.
    * **special action value: "run_code"** shows elapsed time that the hook python script actually took.
    * **special action value: "start_interpreter"** shows up in mom hooks when the pbs_python command actually makes a call to start the interpreter.
   * s**pecial action value: "populate*"** for showing how long it takes for a particularly named Python object to be populated.
   * **special action value: "load_hook_input_file"** for showing the time it takes to load values from hook input file, which is done in mom hooks via pbs_python.
   * s**special action value: "hook_output"** for showing the time it takes for PBS to consume the values that were set by the target hook.

* [hook_perf_stat.ptl.050719.txt](https://github.com/PBSPro/pbspro/files/3153923/hook_perf_stat.ptl.050719.txt)
* [valgrind_server_output.050319.txt](https://github.com/PBSPro/pbspro/files/3143141/valgrind_server_output.050319.txt)
* [valgrind_mom_output.050319.txt]* (https://github.com/PBSPro/pbspro/files/3143143/valgrind_mom_output.050319.txt)

### DIFF
--- a/src/lib/Libpython/pbs_python_svr_internal.c
+++ b/src/lib/Libpython/pbs_python_svr_internal.c
@@ -4199,7 +4199,7 @@ create_py_vnodelist(pbs_list_head *vnlist, char *perf_label, char *perf_action)
 				goto create_py_vnodelist_exit;
 			}
 
-			rc = pbs_python_populate_python_class_from_svrattrl(py_vn, &rqs.rq_attr, perf_label, perf_action);
+			rc = pbs_python_populate_python_class_from_svrattrl(py_vn, &rqs.rq_attr, NULL, NULL);
 
 			if (rc == -1) {
 				snprintf(log_buffer, sizeof(log_buffer),
@@ -4432,7 +4432,7 @@ create_py_joblist(pbs_list_head *joblist, char *perf_label, char *perf_action)
 				goto create_py_joblist_exit;
 			}
 
-			rc = pbs_python_populate_python_class_from_svrattrl(py_jn, &rqs.rq_attr, perf_label, perf_action);
+			rc = pbs_python_populate_python_class_from_svrattrl(py_jn, &rqs.rq_attr, NULL, NULL);
 
 			if (rc == -1) {
 				snprintf(log_buffer, sizeof(log_buffer),
@@ -4665,7 +4665,7 @@ create_py_resvlist(pbs_list_head *resvlist, char *perf_label, char *perf_action)
 				goto create_py_resvlist_exit;
 			}
 
-			rc = pbs_python_populate_python_class_from_svrattrl(py_rn, &rqs.rq_attr, perf_label, perf_action);
+			rc = pbs_python_populate_python_class_from_svrattrl(py_rn, &rqs.rq_attr, NULL, NULL);
 
 			if (rc == -1) {
 				snprintf(log_buffer, sizeof(log_buffer),
--- a/src/lib/Libpython/pbs_python_svr_internal.c
+++ b/src/lib/Libpython/pbs_python_svr_internal.c
@@ -3346,7 +3346,7 @@ _pps_helper_get_queue(pbs_queue *pque, const char *que_name, char *perf_label)
 		que->qu_jobstbuf);
 	/* stuff all the attributes */
 	snprintf((char *)hook_debug.objname, HOOK_BUF_SIZE-1, "%s(%s)", SERVER_QUEUE_OBJECT, que->qu_qs.qu_name);
-	snprintf(perf_action, sizeof(perf_action), "populate:%s", hook_debug.objname);
+	snprintf(perf_action, sizeof(perf_action), "%s:%s", HOOK_PERF_POPULATE, hook_debug.objname);
 	tmp_rc = pbs_python_populate_attributes_to_python_class(py_que,
 		py_que_attr_types,
 		que->qu_attr,
@@ -3496,7 +3496,7 @@ _pps_helper_get_server(char *perf_label)
 
 	/* stuff all the attributes */
 	strncpy((char *)hook_debug.objname, SERVER_OBJECT, HOOK_BUF_SIZE-1);
-	snprintf(perf_action, sizeof(perf_action), "populate:%s", hook_debug.objname);
+	snprintf(perf_action, sizeof(perf_action), "%s:%s", HOOK_PERF_POPULATE, hook_debug.objname);
 	tmp_rc = pbs_python_populate_attributes_to_python_class(py_svr,
 		py_svr_attr_types,
 		server.sv_attr,
@@ -3622,7 +3622,7 @@ _pps_helper_get_job(job *pjob_o, const char *jobid, const char *qname, char *per
 	 * OK, At this point we need to start populating the job class.
 	 */
 	snprintf((char *)hook_debug.objname, HOOK_BUF_SIZE-1, "%s(%s)", SERVER_JOB_OBJECT, pjob->ji_qs.ji_jobid);
-	snprintf(perf_action, sizeof(perf_action), "populate:%s", hook_debug.objname);
+	snprintf(perf_action, sizeof(perf_action), "%s:%s", HOOK_PERF_POPULATE, hook_debug.objname);
 	tmp_rc = pbs_python_populate_attributes_to_python_class(py_job,
 		py_job_attr_types,
 		pjob->ji_wattr,
@@ -3758,7 +3758,7 @@ _pps_helper_get_resv(resc_resv *presv_o, const char *resvid, char *perf_label)
 	 * OK, At this point we need to start populating the resv class.
 	 */
 	snprintf((char *)hook_debug.objname, HOOK_BUF_SIZE-1, "%s(%s)", SERVER_RESV_OBJECT, presv->ri_qs.ri_resvID);
-	snprintf(perf_action, sizeof(perf_action), "populate:%s", hook_debug.objname);
+	snprintf(perf_action, sizeof(perf_action), "%s:%s", HOOK_PERF_POPULATE, hook_debug.objname);
 	tmp_rc = pbs_python_populate_attributes_to_python_class(py_resv,
 		py_resv_attr_types,
 		presv->ri_wattr,
@@ -3882,7 +3882,7 @@ _pps_helper_get_vnode(struct pbsnode *pvnode_o, const char *vname, char *perf_la
 	 * OK, At this point we need to start populating the vnode class.
 	 */
 	snprintf((char *)hook_debug.objname, HOOK_BUF_SIZE-1, "%s(%s)", SERVER_VNODE_OBJECT, pvnode->nd_name);
-	snprintf(perf_action, sizeof(perf_action), "populate:%s", hook_debug.objname);
+	snprintf(perf_action, sizeof(perf_action), "%s:%s", HOOK_PERF_POPULATE, hook_debug.objname);
 	tmp_rc = pbs_python_populate_attributes_to_python_class(py_vnode,
 		py_vnode_attr_types,
 		pvnode->nd_attr,
@@ -5342,7 +5342,7 @@ _pbs_python_event_set(unsigned int hook_event, char *req_user, char *req_host,
 			goto event_set_exit;
 		}
 
-		snprintf(perf_action, sizeof(perf_action), "populate:%s(%s)", EVENT_JOB_OBJECT, rqj->rq_jid);
+		snprintf(perf_action, sizeof(perf_action), "%s:%s(%s)", HOOK_PERF_POPULATE, EVENT_JOB_OBJECT, rqj->rq_jid);
 		rc = pbs_python_populate_python_class_from_svrattrl(py_job,
 			&rqj->rq_attr, perf_label, perf_action);
 
@@ -5386,7 +5386,7 @@ _pbs_python_event_set(unsigned int hook_event, char *req_user, char *req_host,
 				PY_TYPE_EVENT,  PY_EVENT_PARAM_RESV);
 			goto event_set_exit;
 		}
-		snprintf(perf_action, sizeof(perf_action), "populate:%s(%s)", EVENT_RESV_OBJECT, rqj->rq_jid);
+		snprintf(perf_action, sizeof(perf_action), "%s:%s(%s)", HOOK_PERF_POPULATE, EVENT_RESV_OBJECT, rqj->rq_jid);
 		rc = pbs_python_populate_python_class_from_svrattrl(py_resv,
 			&rqj->rq_attr, perf_label, perf_action);
 
@@ -5432,7 +5432,7 @@ _pbs_python_event_set(unsigned int hook_event, char *req_user, char *req_host,
 			goto event_set_exit;
 		}
 
-		snprintf(perf_action, sizeof(perf_action), "populate:%s(%s)", EVENT_JOB_OBJECT, rqj->rq_objname);
+		snprintf(perf_action, sizeof(perf_action), "%s:%s(%s)", HOOK_PERF_POPULATE, EVENT_JOB_OBJECT, rqj->rq_objname);
 		rc = pbs_python_populate_python_class_from_svrattrl(py_job,
 			&rqj->rq_attr, perf_label, perf_action);
 
@@ -5624,7 +5624,7 @@ _pbs_python_event_set(unsigned int hook_event, char *req_user, char *req_host,
 		/* SET VNODE_LIST param */
 		(void)PyDict_SetItemString(py_event_param, PY_EVENT_PARAM_VNODELIST,
 			Py_None);
-		py_vnodelist = create_py_vnodelist(vnlist, perf_label, "populate:pbs.event().vnode_list");
+		py_vnodelist = create_py_vnodelist(vnlist, perf_label, HOOK_PERF_POPULATE_VNODELIST);
 		if (py_vnodelist == NULL) {
 			LOG_ERROR_ARG2("%s: failed to create a Python vnodelist object for param['%s']",
 				PY_TYPE_EVENT, PY_EVENT_PARAM_VNODELIST);
@@ -5646,7 +5646,7 @@ _pbs_python_event_set(unsigned int hook_event, char *req_user, char *req_host,
 
 		(void)PyDict_SetItemString(py_event_param, PY_EVENT_PARAM_RESVLIST,
 			Py_None);
-		py_resvlist = create_py_resvlist(resvlist, perf_label, "populate:pbs.event().resv_list");
+		py_resvlist = create_py_resvlist(resvlist, perf_label, HOOK_PERF_POPULATE_RESVLIST);
 		if (py_resvlist == NULL) {
 			LOG_ERROR_ARG2("%s: failed to create a Python resvlist object for param['%s']",
 				PY_TYPE_EVENT, PY_EVENT_PARAM_RESVLIST);
@@ -5820,7 +5820,7 @@ _pbs_python_event_set(unsigned int hook_event, char *req_user, char *req_host,
 			goto event_set_exit;
 		}
 
-		snprintf((char *)perf_action, sizeof(perf_action), "populate:%s(%s)", EVENT_JOB_OBJECT, rqj->rq_jid);
+		snprintf((char *)perf_action, sizeof(perf_action), "%s:%s(%s)", HOOK_PERF_POPULATE, EVENT_JOB_OBJECT, rqj->rq_jid);
 		rc = pbs_python_populate_python_class_from_svrattrl(py_job, &rqj->rq_attr, perf_label, perf_action);
 
 		if (rc == -1) {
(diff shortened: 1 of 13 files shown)


src/lib/Libpython/pbs_python_svr_internal.c
```diff
@@ -6754,9 +6857,12 @@ _pbs_python_event_to_request(unsigned int hook_event, hook_output_param_t *req_p
...
 			log_err(PBSE_INTERNAL, __func__,
 				"unexpected hook event type");
-			return -1;
+			goto event_to_request_exit;
 	}
-	return 0;
+event_to_request_exit:
+	rc = 0;
+	hook_perf_stat_stop(perf_label, perf_action, 0);
+	return rc;
```
vstumpf: does this mean this function will always return 0 (success)

bayucan: Woops...the "rc = 0" should be before the "event_to_request_exit" label. I'll fix.

src/lib/Libpython/pbs_python_svr_internal.c
```diff
@@ -6743,7 +6846,7 @@ _pbs_python_event_to_request(unsigned int hook_event, hook_output_param_t *req_p
 					log_err(PBSE_INTERNAL, __func__, log_buffer);
 					Py_CLEAR(py_attr_keys);
 					free(key_str);
-					return -1;
+					goto event_to_request_exit;
```
mike0042: Shouldn't you be setting rc before the goto?

src/lib/Libpython/pbs_python_svr_internal.c
```diff
@@ -6982,7 +7088,7 @@ pbsv1mod_meth_get_queue(PyObject *self, PyObject *args, PyObject *kwds)
 	}
 
 	hook_set_mode = C_MODE;
-	py_que = _pps_helper_get_queue(NULL, name);
+	py_que = _pps_helper_get_queue(NULL, name, "hook_func");
```
mike0042: Should "hook_func" be a macro?

src/lib/Libpython/pbs_python_svr_internal.c
```diff
@@ -5988,7 +6077,7 @@ _pbs_python_event_set(unsigned int hook_event, char *req_user, char *req_host,
 
 			(void)PyDict_SetItemString(py_event_param, PY_EVENT_PARAM_JOBLIST,
 				Py_None);
-			py_joblist = create_py_joblist(joblist);
+			py_joblist = create_py_joblist(joblist, perf_label, "populate:pbs.event().job_list");
```
mike0042: Should multiple instances of the string "populate" use macros instead?

src/lib/Libpython/pbs_python_svr_internal.c
```diff
@@ -10720,8 +10825,8 @@ py_get_queue_static(char *qname, char *svr_name)
 		goto get_queue_error_exit;
 	}
 
-	rc = pbs_python_populate_python_class_from_svrattrl(\
-				       py_queue, &queuel);
+	snprintf(perf_label, sizeof(perf_label), "hook_func:%s(%s)", SERVER_QUEUE_OBJECT, qname);
```
mike0042: Should multiple instances of "hook_func" use a macro?

src/lib/Libutil/hook.c
```diff
@@ -3868,3 +3868,82 @@ cleanup_hooks_workdir(struct work_task *ptask)
...
+ * @return void
+ *
+ */
+void
+hook_perf_stat_start(char *label, char *action, int print_start_msg)
+{
+	char instance[MAXBUFLEN];
+
+	if (!will_log_event(PBSEVENT_DEBUG4))
+		return;
```
mike0042: Is it possible for an admin to change the log event mask while a hook is running? That would mean hook_perf_stat_start() would not get called, but hook_perf_stat_stop() would. Might that cause problems?

src/lib/Libutil/misc_utils.c
```diff
@@ -1520,3 +1532,221 @@ get_preemption_order(struct preempt_ordering *porder, int req, int used)
...
+{
+	LARGE_INTEGER time, freq;
+
+	if (QueryPerformanceFrequency(&freq) == 0)
+        	return (0);
+
+	if (QueryPerformanceCounter(&time) == 0)
+        	return (0);
+
+	return ((double)time.QuadPart / freq.QuadPart);
```
mike0042: Do we need to check freq.QuadPart to make sure it's not zero?

src/lib/Libutil/misc_utils.c
```diff
@@ -1520,3 +1532,221 @@ get_preemption_order(struct preempt_ordering *porder, int req, int used)
...
+ * @return double - number of seconds.
+ **/
+static double
+get_cputime()
+{
+	clock_t	clock_cycles;
+	clock_cycles = clock();
+	if (clock_cycles == -1)
+		return (0);
+	return (double)clock_cycles / CLOCKS_PER_SEC;
```
mike0042: CLOCKS_PER_SEC should never be zero, but it would cause a failure if it were.

src/server/hook_func.c
```diff
@@ -4697,20 +4730,24 @@ recreate_request(struct batch_request *preq)
 	hook_output_param_init(&req_params);
 	if (preq->rq_type == PBS_BATCH_QueueJob) {
 		req_params.rq_job = (struct rq_quejob *)&preq->rq_ind.rq_queuejob;
+		snprintf(perf_label, sizeof(perf_label), "hook_%s_%s_%d", HOOKSTR_QUEUEJOB, preq->rq_ind.rq_queuejob.rq_jid, getpid());
 		rc = pbs_python_event_to_request(HOOK_EVENT_QUEUEJOB,
-			&req_params);
+			&req_params, perf_label, "hook_output");
```
mike0042: Should the "hook_output" string be a macro?

src/server/svr_func.c
```diff
@@ -6180,9 +6182,11 @@ execute_python_prov_script(hook  *phook,
 	}
 
 	/* let rc pass through */
-	rc=pbs_python_run_code_in_namespace(&svr_interp_data,
+	hook_perf_stat_start(perf_label, "run_code", 0);
+	rc = pbs_python_run_code_in_namespace(&svr_interp_data,
 		phook->script,
 		&exit_code);
+	hook_perf_stat_stop(perf_label, "run_code", 0);
```
mike0042: Should "run_code" be a macro?

src/server/w32_svr_periodic_hook.c
```diff
@@ -291,8 +292,10 @@ execute_python_periodic_hook(hook  *phook)
 	req_ptr.vns_list = (pbs_list_head *)get_vnode_list();
 	req_ptr.resv_list = (pbs_list_head *)get_resv_list();
 
+	snprintf(perf_label, sizeof(perf_label), "hook_%s_%s_%d", hook_event_as_string(hook_event), phook->hook_name, mypid);
+
 	rc = pbs_python_event_set(hook_event, username,
-		"server", &req_ptr);
+		"server", &req_ptr, perf_label);
```
mike0042: Should "server" be a macro?

src/tools/pbs_python.c
```diff
@@ -2826,7 +2873,9 @@ main(int argc, char *argv[], char *envp[])
 		(void)pbs_python_ext_alloc_python_script(hook_script,
 			(struct python_script **) &py_script);
 
+		hook_perf_stat_start(perf_label, "start_interpreter", 0);
 		pbs_python_ext_start_interpreter(&svr_interp_data);
+		hook_perf_stat_stop(perf_label, "start_interpreter", 0);
```
mike0042: Should "start_interpreter" be a macro?

src/tools/pbs_python.c
```diff
@@ -2709,7 +2755,8 @@ main(int argc, char *argv[], char *envp[])
 				&server_jobs, &server_jobs_ids,
 				&server_queues, &server_queues_names,
 				&server_resvs, &server_resvs_resvids,
-				&server_vnodes, &server_vnodes_names);
+				&server_vnodes, &server_vnodes_names,
+				the_data, "load_hook_data");
```
mike0042: Should "load_hook_data" be a macro?

src/tools/pbs_python.c
```diff
@@ -3156,8 +3196,10 @@ main(int argc, char *argv[], char *envp[])
 
 			rc=Py_Main(1, tmp_argv);
 		} else {
+			hook_perf_stat_start(perf_label, "run_code", 0);
 			rc=pbs_python_run_code_in_namespace(&svr_interp_data,
 				py_script, 0);
+			hook_perf_stat_stop(perf_label, "run_code", 0);
```
mike0042: Should "run_code" be a macro?

src/tools/pbs_python.c
```diff
@@ -3572,6 +3622,8 @@ main(int argc, char *argv[], char *envp[])
 			free(progname);
 		if (env_str != NULL)
 			free(env_str);
+
+		hook_perf_stat_stop(perf_label, "pbs_python", 1);
```
mike0042: Should "pbs_python" be a macro?

test/tests/functional/pbs_hook_perf_stat.py
```diff
@@ -0,0 +1,482 @@
...
+    def test_queuejob_hook(self):
+        """
+        Test that pbs_server collects performance stats for queuejob hook
+        """
+        hook_content = ("""
+import pbs
+pbs.logmsg(pbs.LOG_DEBUG, "server hook called")
+s = pbs.server()
+pbs.logmsg(pbs.LOG_DEBUG, "server data collected for %s" % (s.name,))
+""")
```
mike0042: Why use parenthesis here?

src/lib/Libutil/hook.c
bayucan: If hook_perf_stop() is called and yet hook_perf_start() was called but returned early since DEBUG4 was not in the log_event_mask at the time, then nothing would really happen since the 'instance' has not been added to the performance stats list; that is, hook_perf_stop() would not find an entry for 'instance', so nothing is logged. It's the corollary that needs to be addressed. If hook_perf_start() was called, and yet hook_perf_stop() returns early since it does not find DEBUG4 in log_event_mask, then we'll have a leftover instance entry, that is not freed, and may hang around forever. I've updated the code in this case to make sure to free up always the instance entry that was set up by hook_perf_start().

src/lib/Libutil/misc_utils.c
bayucan: Checked msdn/win32 info on QueryPerformanceFrequency():
https://msdn.microsoft.com/en-us/library/windows/desktop/ms644905(v=vs.85).aspx

It asserts:

BOOL WINAPI QueryPerformanceFrequency(
  _Out_ LARGE_INTEGER *lpFrequency
);

Parameters
lpFrequency [out]
A pointer to a variable that receives the current performance-counter frequency, in counts per second. **If the installed hardware doesn't support a high-resolution performance counter, this parameter can be zero (this will not occur on systems that run Windows XP or later).**

Return value
If the installed hardware supports a high-resolution performance counter, the return value is nonzero.

If the function fails, the return value is zero. To get extended error information, call GetLastError. On systems that run Windows XP or later, the function will always succeed and will thus never return zero.
--
Since we return early if QueryPerformanceFrequency()  has a return code of 0, then we should get to a divide by zero crash.

src/lib/Libutil/misc_utils.c
bayucan: Ok, I'll put an assert(CLOCKS_PER_SEC != 0) statement out there, to reflect that this is not expected.

src/server/hook_func.c
bayucan: Yes...updated.

src/server/svr_func.c
bayucan: Yes...updated.

src/server/w32_svr_periodic_hook.c
bayucan: Yes...updated.

src/tools/pbs_python.c
bayucan: Yes...updated.

src/tools/pbs_python.c
bayucan: Yes...updated.

src/tools/pbs_python.c
bayucan: Yes...updated.

src/tools/pbs_python.c
bayucan: Yes...updated.

test/tests/functional/pbs_hook_perf_stat.py
bayucan: Yes, it's unnecessary.

mike0042: More specifically, why use parenthesis around the text block that hook_content is being assigned? This is present in a few places.

bayucan: Yes, that is also unnecessary. Ok, I've updated ptl test, removing all instances of those not needed outer parenthesis.

src/lib/Libpython/pbs_python_svr_internal.c
```diff
@@ -4144,8 +4199,7 @@ create_py_vnodelist(pbs_list_head *vnlist)
 				goto create_py_vnodelist_exit;
 			}
 
-			rc = pbs_python_populate_python_class_from_svrattrl(\
-						       py_vn, &rqs.rq_attr);
+			rc = pbs_python_populate_python_class_from_svrattrl(py_vn, &rqs.rq_attr, perf_label, perf_action);
```
vstumpf: Wouldn't this also call hook_perf_stat_start and hook_perf_stat_stop with the same perf_label and perf_action as you created in 4101?
`create_py_vnodelist`:4101 - `hook_perf_stat_start` - creates a new p_stat and sets initial values
`pbs_python_populate_python_class_from_svr_attrl`:2279 - `hook_perf_stat_start` - finds the existing p_stat and overwrites values
`pbs_python_populate_python_class_from_svr_attrl`:2336 - `hook_perf_stat_stop` - finds the p_stat, stops it, prints a summary, frees p_stat
`create_py_vnodelist`:4275 - `hook_perf_stat_stop` - can't find the p_stat because it's been freed, returns without printing.

src/lib/Libpython/pbs_python_svr_internal.c
```diff
@@ -4365,8 +4432,7 @@ create_py_joblist(pbs_list_head *joblist)
 				goto create_py_joblist_exit;
 			}
 
-			rc = pbs_python_populate_python_class_from_svrattrl(\
-						       py_jn, &rqs.rq_attr);
+			rc = pbs_python_populate_python_class_from_svrattrl(py_jn, &rqs.rq_attr, perf_label, perf_action);
```
vstumpf: Same as above

src/lib/Libpython/pbs_python_svr_internal.c
```diff
@@ -4585,8 +4665,7 @@ create_py_resvlist(pbs_list_head *resvlist)
 				goto create_py_resvlist_exit;
 			}
 
-			rc = pbs_python_populate_python_class_from_svrattrl(
-						       py_rn, &rqs.rq_attr);
+			rc = pbs_python_populate_python_class_from_svrattrl(py_rn, &rqs.rq_attr, perf_label, perf_action);
```
vstumpf: Same as above

src/lib/Libutil/misc_utils.c
```diff
@@ -1520,3 +1534,251 @@ get_preemption_order(struct preempt_ordering *porder, int req, int used)
...
+ */
+static perf_stat_t *
+perf_stat_alloc(char *instance)
+{
+	perf_stat_t	*p_stat;
+
+	if ((instance == NULL) || (instance[0] == '\0'))
+		return NULL;
+
+	p_stat = (perf_stat_t *)malloc(sizeof(perf_stat_t));
```
vstumpf: You don't need to cast malloc.

src/lib/Libutil/misc_utils.c
```diff
@@ -1520,3 +1534,251 @@ get_preemption_order(struct preempt_ordering *porder, int req, int used)
...
+perf_stat_find(char *instance)
+{
+	perf_stat_t *p_stat;
+
+	if ((instance == NULL) || (instance[0] == '\0') || (perf_stats_initialized == 0))
+		return (NULL);
+
+	p_stat = (perf_stat_t *)GET_NEXT(perf_stats);
+	while (p_stat) {
+		if ( strcmp(p_stat->instance, instance) == 0) {
```
vstumpf: Remove the extra space before `strcmp`

src/lib/Libutil/misc_utils.c
```diff
@@ -1520,3 +1534,251 @@ get_preemption_order(struct preempt_ordering *porder, int req, int used)
...
+perf_stat_remove(char *instance)
+{
+	perf_stat_t *p_stat;
+
+	if ((instance == NULL) || (instance[0] == '\0') || (perf_stats_initialized == 0))
+		return;
+
+	p_stat = (perf_stat_t *)GET_NEXT(perf_stats);
+	while (p_stat) {
+		if ( strcmp(p_stat->instance, instance) == 0) {
```
vstumpf: Remove the extra space before `strcmp`

test/tests/functional/pbs_hook_perf_stat.py
```diff
@@ -0,0 +1,482 @@
...
+        self.server.manager(MGR_CMD_SET, SERVER, a)
+
+        self.mom.add_config({'$logevent': 4095})
+        self.mom.signal('-HUP')
+
+    def test_queuejob_hook(self):
+        """
+        Test that pbs_server collects performance stats for queuejob hook
+        """
+        hook_content = """
```
vstumpf: The hook content is repeated multiple times in different tests, maybe add it once under the class?

src/lib/Libpython/pbs_python_svr_internal.c
bayucan: Good catch, Vincent! I was supposed to NULL the values of  'perf_label' and 'perf_action' parameters to the inner call of pbs_pbs_python_populate_python_class_from_svrattrl(), so perf stat collecting task does not happen. That function is being called for each object created by create_vnodelist(), create_joblist() and create_resvlist(), and that could be a lot (e.g. ex 10,000 vnodes!) which would fill the daemon logs too much. So I opted to just do 1 perf stat perf_hook_start at the beginning, and 1 perf_hook_stop at end of create_vnodelist(), create_joblist(), and create_resvlist(). So a call to create_vnodelist() would give a single elapsed time for how long it took say 10,000 vnodes to get created, rather than 10,000 separate daemon log entries. I just missed NULL-ing the perf_* parameters to *populate_class_from_svrattrl()...

src/lib/Libpython/pbs_python_svr_internal.c
bayucan: Will fix.

src/lib/Libpython/pbs_python_svr_internal.c
bayucan: Will fix.

test/tests/functional/pbs_hook_perf_stat.py
bayucan: Good idea. I'll update.

mike0042:
Still looks good. Merging.

PR #1111, opened 2019-05-06
Fixed PTL issue trigger_sched_cycle set to False by default

visheshh:
#### Describe Bug or Feature
trigger_sched_cycle value was set to False by default rather than setting the value passed

#### Describe Your Change
Assigned the appropriate value


prakashcv13:
will merge once travis passes.

prakashcv13:
merged #1111

PR #1114, opened 2019-05-07
Fixed when high priority requesting vnode fails to preempt

arungrover:
#### Describe Bug or Feature
If a high priority job requests a vnode/host it fails to find a low priority job for preemption.

#### Describe Your Change
The problem was that when we pass through all the low priority jobs in scheduler to find out which job can help in running the high priority job, we check for chunk level resources only. While it works in most cases but there is a special case for vnode/host resource. These resources are node level resources but they don't usually end up as a chunk level resource because of which scheduler fails to find the preemptible candidates and that results in high priority job not running.

FIX: To fix this problem, Scheduler now checks in execvnode of the running job to match the vnode name when the resource in contention is vnode. In case of host, it looks through all the nodes the job is running on matches the host.

#### Link to Design Doc

NA

[test_after_fix.txt](https://github.com/PBSPro/pbspro/files/3150794/test_after_fix.txt)
[test_before_fix.txt](https://github.com/PBSPro/pbspro/files/3150795/test_before_fix.txt)

### DIFF
--- a/src/scheduler/job_info.c
+++ b/src/scheduler/job_info.c
@@ -5054,14 +5054,35 @@ static int cull_preemptible_jobs(resource_resv *job, void *arg)
 			}
 			break;
 		case INSUFFICIENT_RESOURCE:
-			for (index = 0; job->select->chunks[index] != NULL; index++)
-			{
-				for (req_scan = job->select->chunks[index]->req; req_scan != NULL; req_scan = req_scan->next)
+			/* special check for vnode and host resource because those resources
+			 * do not get into chunk level resources. So in such a case we 
+			 * compare the resource name with the chunk name
+			 */
+			if (inp->err->rdef == getallres(RES_VNODE)) {
+				resource_req *hreq = find_resource_req(inp->job->resreq, inp->err->rdef);
+				if (hreq == NULL)
+					return 0;
+				for (index = 0; job->execselect->chunks[index] != NULL; index++)
 				{
-					if (req_scan->def == inp->err->rdef) {
-						if (req_scan->type.is_non_consumable ||
-						    req_scan->amount > 0) {
+					resource_req *lreq = find_resource_req(job->execselect->chunks[index]->req, inp->err->rdef);
+					if (lreq != NULL)
+						if (strcmp(hreq->res_str, lreq->res_str) == 0)
 							return 1;
+				}
+			} else if (inp->err->rdef == getallres(RES_HOST)) {
+				resource_req *hreq = find_resource_req(inp->job->resreq, inp->err->rdef);
+				if (find_node_by_host(job->ninfo_arr, hreq->res_str) != NULL)
+					return 1;
+			} else {
+				for (index = 0; job->select->chunks[index] != NULL; index++)
+				{
+					for (req_scan = job->select->chunks[index]->req; req_scan != NULL; req_scan = req_scan->next)
+					{
+						if (req_scan->def == inp->err->rdef) {
+							if (req_scan->type.is_non_consumable ||
+							    req_scan->amount > 0) {
+									return 1;
+							}
 						}
 					}
 				}
--- a/src/scheduler/node_info.h
+++ b/src/scheduler/node_info.h
@@ -671,6 +671,10 @@ int add_event_to_nodes(timed_event *te, nspec **nspecs);
 
 int add_node_events(timed_event *te, void *arg1, void *arg2);
 
+/*
+ * Find a node by its hostname
+ */
+node_info *find_node_by_host(node_info **ninfo_arr, char *host);
 #ifdef	__cplusplus
 }
 #endif
--- a/test/tests/functional/pbs_preemption.py
+++ b/test/tests/functional/pbs_preemption.py
@@ -54,6 +54,14 @@ class TestPreemption(TestFunctional):
              'enabled': 'True',
              'Priority': 200}
         self.server.manager(MGR_CMD_CREATE, QUEUE, a, "expressq")
+        if len(self.moms) == 2:
+            self.mom1 = self.moms.keys()[0]
+            self.mom2 = self.moms.keys()[1]
+            # Since some tests need multi-node setup and majority don't,
+            # delete the second node so that single node tests don't fail.
+            # Tests needing multi-node setup will create the second node
+            # explicity.
+            self.server.manager(MGR_CMD_DELETE, NODE, id=self.mom2)
 
     def submit_jobs(self):
         """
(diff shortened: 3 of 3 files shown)


test/tests/functional/pbs_preemption.py
```diff
@@ -251,3 +251,38 @@ def test_preempt_retry(self):
...
+        self.server.create_vnodes(name='vnode', attrib=a, num=11,
+                                  mom=self.mom, usenatvnode=False)
+
+        a = {'Resource_List.select': '1:ncpus=2+1:ncpus=2'}
+        j = Job(TEST_USER, attrs=a)
+        jid1 = self.server.submit(j)
+        jid2 = self.server.submit(j)
+        jid3 = self.server.submit(j)
+        jid4 = self.server.submit(j)
+        jid5 = self.server.submit(j)
```
bhroam: Don't reuse the job object, it caches data from for the specific job itself.

arungrover:
@bhroam Thanks for reviewing this pull request.
I've addressed your comment, fixed the issue with 'host' resource and added a new test in the latest commit.

src/scheduler/node_info.h
```diff
@@ -671,6 +671,10 @@ int add_event_to_nodes(timed_event *te, nspec **nspecs);
 
 int add_node_events(timed_event *te, void *arg1, void *arg2);
 
+/*
+ * Find a node by its hostname
+ */
+node_info * find_node_by_host(node_info **ninfo_arr, char *host);
```
bhroam: Please attach the * to the function name

test/tests/functional/pbs_preemption.py
```diff
@@ -263,12 +267,16 @@ def test_vnode_resource_contention(self):
...
+        j1 = Job(TEST_USER, attrs=a)
+        jid1 = self.server.submit(j1)
+        j2 = Job(TEST_USER, attrs=a)
+        jid2 = self.server.submit(j2)
+        j3 = Job(TEST_USER, attrs=a)
+        jid3 = self.server.submit(j3)
+        j4 = Job(TEST_USER, attrs=a)
+        jid4 = self.server.submit(j4)
+        j5 = Job(TEST_USER, attrs=a)
+        jid5 = self.server.submit(j5)
```
bhroam: Put this into a loop.  Create the job, submit the job, and then check if it is in the R state.  You don't use the job ids after checking if they are in the R state.

test/tests/functional/pbs_preemption.py
```diff
@@ -286,3 +294,51 @@ def test_vnode_resource_contention(self):
...
+        self.server.expect(JOB, {'job_state': 'S'}, id=pjid)
+
+        # Submit another express queue job requesting the host,
+        # this job will stay queued
+        a = {ATTR_q: 'expressq', 'Resource_List.host': self.mom1}
+        hj2 = Job(TEST_USER, attrs=a)
+        hjid2 = self.server.submit(hj2)
+        self.server.expect(JOB, {'job_state': 'R'}, id=hjid)
+        job_stat = self.server.status(JOB, id=hjid2)
+        print job_stat[0]['comment']
```
bhroam: This looks like a debug message.  Is it necessary?

arungrover: Ahh... a partially baked test was pushed, I've rectified it in the latest commit

agrawalravi90:
Can you please describe the solution and summarize your changes in the "Describe your changes" section" You've only described the bug there :)

src/scheduler/job_info.c
```diff
@@ -5054,14 +5054,35 @@ static int cull_preemptible_jobs(resource_resv *job, void *arg)
...
+					return 0;
+				for (index = 0; job->execselect->chunks[index] != NULL; index++)
 				{
-					if (req_scan->def == inp->err->rdef) {
-						if (req_scan->type.is_non_consumable ||
-						    req_scan->amount > 0) {
+					resource_req *lreq = find_resource_req(job->execselect->chunks[index]->req, inp->err->rdef);
+					if (lreq != NULL)
+						if (strcmp(hreq->res_str, lreq->res_str) == 0)
 							return 1;
```
agrawalravi90: this return seems to be indented too much

src/scheduler/job_info.c
```diff
@@ -5054,14 +5054,35 @@ static int cull_preemptible_jobs(resource_resv *job, void *arg)
...
+			if (inp->err->rdef == getallres(RES_VNODE)) {
+				resource_req *hreq = find_resource_req(inp->job->resreq, inp->err->rdef);
+				if (hreq == NULL)
+					return 0;
+				for (index = 0; job->execselect->chunks[index] != NULL; index++)
 				{
-					if (req_scan->def == inp->err->rdef) {
-						if (req_scan->type.is_non_consumable ||
-						    req_scan->amount > 0) {
+					resource_req *lreq = find_resource_req(job->execselect->chunks[index]->req, inp->err->rdef);
```
agrawalravi90: just trying to understand the logic: if a job requests a vnode explicitly, then here we are checking if any other job has also requested that vnode? What if a lowp job is just running on a vnode that's requested by the highp job? Shouldn't we also preempt such a job? (similar to how you handle the host case below)

test/tests/functional/pbs_preemption.py
```diff
@@ -54,6 +54,10 @@ def setUp(self):
              'enabled': 'True',
              'Priority': 200}
         self.server.manager(MGR_CMD_CREATE, QUEUE, a, "expressq")
+        if len(self.moms) == 2:
+            self.mom1 = self.moms.keys()[0]
+            self.mom2 = self.moms.keys()[1]
+            self.server.manager(MGR_CMD_DELETE, NODE, id=self.mom2)
```
agrawalravi90: can you please add a comment here to explain why you are deleting the nodes on mom2 ?

test/tests/functional/pbs_preemption.py
```diff
@@ -251,3 +255,80 @@ def test_preempt_retry(self):
...
         self.server.expect(JOB, {'job_state': 'R'}, id=jid3)
+
+    def test_vnode_resource_contention(self):
+        """
+        Test to make sure that preemption happens when the resource in
+        contention is vnode.
+        """
+
+        a = {'resources_available.ncpus': 2}
+        self.server.create_vnodes(name='vnode', attrib=a, num=11,
```
agrawalravi90: any particular reason why you are using 11 vnodes and 5 low priority jobs? Why not just 1 vnode and 1 low priority job?

test/tests/functional/pbs_preemption.py
```diff
@@ -251,3 +255,80 @@ def test_preempt_retry(self):
...
+                                  mom=self.mom, usenatvnode=False)
+
+        a = {'Resource_List.select': '1:ncpus=2+1:ncpus=2'}
+        for each in range(5):
+            j = Job(TEST_USER, attrs=a)
+            jid = self.server.submit(j)
+            self.server.expect(JOB, {'job_state': 'R'}, id=jid)
+
+        # Make sure vnode[4] has jobs running on it
```
agrawalravi90: is it guaranteed that vnode[4] will be job-busy? you created 11 vnodes with 2 ncpus each, so 1 vnode should be free after 5 jobs requesting 4 ncpus, couldn't that be vnode[4] ?

src/scheduler/job_info.c
arungrover: Looks fine to me in my editor. How does it look in yours?

agrawalravi90: In Github's editor, it looks like it's indented by one extra tab, but if it looks fine in yours then it must be ok

src/scheduler/job_info.c
arungrover: When resource requested by high priority job is a vnode then we check the execvnode list of running jobs and matches vnode name. This is because whether or not low priority job has requested a specific vnode, the vnode it is running on makes it into its execvnode list.
If I understand correctly, your second question is the bug we are trying to fix.

test/tests/functional/pbs_preemption.py
arungrover: I wanted to have a sufficient amount of nodes with free ncpus so that scheduler does not apply preemption because of ncpus. 11 is just a random number, I could have done it with lesser. If you don't feel strongly about it, I would like to retain it as is.

test/tests/functional/pbs_preemption.py
arungrover: Well, there is no node sort key, so unless we change the default configs I think vnode[4] should be used.

arungrover:
@agrawalravi90 Thanks for reviewing the code, I have addressed your comments please have a look again

src/scheduler/job_info.c
agrawalravi90: but here you are looking for the vnode resource in lowp's execselect, not execvnode, right? I'm not sure I understand how this will check the execvnode list.

test/tests/functional/pbs_preemption.py
```diff
@@ -54,6 +54,13 @@ def setUp(self):
              'enabled': 'True',
              'Priority': 200}
         self.server.manager(MGR_CMD_CREATE, QUEUE, a, "expressq")
+        if len(self.moms) == 2:
+            self.mom1 = self.moms.keys()[0]
+            self.mom2 = self.moms.keys()[1]
+            # Since some tests need multi-node setup and majority don't
```
agrawalravi90: The comment is a bit confusing, can you please reword it?

test/tests/functional/pbs_preemption.py
agrawalravi90: This kind of makes it seem like the test is testing some functionality by checking that vnode[4] is job-busy. How about just finding the first node that is job-busy and using that?

arungrover: How about I write a comment that randomly select one of the busy nodes and request that vnode in the high priority job?

src/scheduler/job_info.c
arungrover: I'm sorry I meant execselect, not execvnode

test/tests/functional/pbs_preemption.py
```diff
@@ -57,9 +57,10 @@ def setUp(self):
         if len(self.moms) == 2:
             self.mom1 = self.moms.keys()[0]
             self.mom2 = self.moms.keys()[1]
-            # Since some tests need multi-node setup and majority don't
-            # deleting the second node for majority, tests which need
-            # multi-node setup will create the second node itself.
+            # Since some tests need multi-node setup and majority don't,
+            # Delete the second node so that single node tests don't fail.
```
bhroam: The previous line ended with a comma.  This is in the middle of the sentence.  Lowercase 'Delete'

arungrover:
I've made the change @bhroam. Please have a look.

arungrover:
@bhroam can you please merge the pull or do you want me to rebase it?

bhroam:
It's been  9 days since the PR started.  I think it'd be best if you rebase first just incase something has changed.

arungrover:
I have squashed and rebased, please take it forward

PR #1116, opened 2019-05-07
Replace pbs_diag with pbs_snapshot for post-analysis data capturing

shilpakodli:
#### Describe Bug or Feature
Replacing pbs_diag with pbs_snapshot to capture post_analysis data while running PTL tests.
This will capture all the logs which were capturing by pbs_diag, apart from that it captures mom logs on local and remote host and also if there is any crash is detected then it created stack trace and saves the logs.

#### Describe Your Change
Replaced pbs_diag with pbs_snapshot in ptl_test_data.py file.

### DIFF
--- a/test/fw/ptl/utils/plugins/ptl_test_data.py
+++ b/test/fw/ptl/utils/plugins/ptl_test_data.py
@@ -1,5 +1,5 @@
 # coding: utf-8
-
+#
 # Copyright (C) 1994-2019 Altair Engineering, Inc.
 # For more information, contact Altair at www.altair.com.
 #
@@ -56,7 +56,7 @@ class PTLTestData(Plugin):
     Save post analysis data on test cases failure or error
     """
     name = 'PTLTestData'
-    score = sys.maxint - 6
+    score = sys.maxsize - 6
     logger = logging.getLogger(__name__)
 
     def __init__(self):
@@ -119,7 +119,7 @@ class PTLTestData(Plugin):
             signal.alarm(0)
         self.logger.log(logging.DEBUG2, 'Saving post analysis data...')
         current_host = socket.gethostname().split('.')[0]
-        self.du.mkdir(current_host, path=datadir, mode=0755,
+        self.du.mkdir(current_host, path=datadir, mode=0o755,
                       parents=True, logerr=False, level=logging.DEBUG2)
         if err is not None:
             if isclass(err[0]) and issubclass(err[0], SkipTest):
@@ -149,9 +149,10 @@ class PTLTestData(Plugin):
             self.logger.error(_msg)
             f.close()
             return
-        svr = getattr(_test, 'server', None)
-        if svr is not None:
-            svr_host = svr.hostname
+
+        servers = getattr(_test, 'servers', None)
+        if servers is not None:
+            server_host = servers.values()[0].shortname
         else:
             _msg = 'Could not find Server Object in given test object'
             _msg += ', skipping saving post analysis data'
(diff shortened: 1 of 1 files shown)


test/fw/ptl/utils/plugins/ptl_test_data.py
```diff
@@ -149,160 +149,140 @@ def __save_home(self, test, status, err=None):
...
-        ret = self.du.run_copy(current_host, diag_out_r, diag_out_dest,
-                               sudo=True, level=logging.DEBUG2)
+
+            snap_out = ret['out'][0].strip(" ")
+            output_tar = snap_out.split(":")[1]
+
+        snap_out_dest = os.path.join(datadir, os.path.basename(snap_out))
+
+        snap_out_r = output_tar.strip(" ")
+        ret = self.du.run_copy(current_host, snap_out_r, snap_out_dest,
```
hirenvadalia: what will happen is server_host is not current_host? can you please try out this scenario?

test/fw/ptl/utils/plugins/ptl_test_data.py
```diff
@@ -149,160 +149,140 @@ def __save_home(self, test, status, err=None):
...
-                f.close()
-                return
-            self.du.rm(hostname=current_host, path=diag_out_dest,
-                       force=True, sudo=True, level=logging.DEBUG2)
-            diag_out_dest = diag_out_dest.rstrip('.tar')
-            for c in cores:
-                cmd = [pbs_diag, '-g', c]
-                ret = self.du.run_cmd(svr_host, cmd, sudo=True,
-                                      level=logging.DEBUG2)
+        snap_out_dest = snap_out_dest.rstrip('.tgz')
```
hirenvadalia: Why we are extracting snapshot tarball? previously in diag code, it was not extracting final diag tarball so I suggest we should keep that way... and let the caller of pbs_benchpress decide what to do with snapshot tarball.

NOTE: don't get confused with extracting at line no 220 in old code, it was there to extract backtrace of core file and remove large core file from diag tarball

test/fw/ptl/utils/plugins/ptl_test_data.py
```diff
@@ -149,160 +149,140 @@ def __save_home(self, test, status, err=None):
...
+            pbs_snapshot_path,
+            '--daemon-logs',
+            '2',
+            '--accounting-logs',
+            '2',
+            '--with-sudo']
+        if len(add_hosts) > 0:
+            cmd += ['--additional-hosts=' + ','.join(map(str, add_hosts))]
+        cmd += ['-o', parent_dir]
+        cmd += ['> /dev/null', '2>&1']
```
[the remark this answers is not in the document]
shilpakodli: To redirect output and error to /dev/null but this doesn't seem to be working. I will remove above line of code and include logerr=False in du.run_cmd. This should be ok right ?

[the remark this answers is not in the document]
hirenvadalia: yes, logerr=False will do the job.

test/fw/ptl/utils/plugins/ptl_test_data.py
shilpakodli: The reason behind untarig the snapshot tarball is if we capture logs on multinode setup then pbs_snapshot stores remote host logs inside localhost logs.
snapshot_datestamp_logs.tar.gz has below directories inside when we run on multinode setup having one remote host mom
- server_logs
- comm_logs
- etc
- remote_host_snapshot_log.tar.gz

In above directory structure, you can see remote_host tar ball. When TH2 extracts main snapshot tar file then it will have this remote host tar file and we cannot just open it and see the logs. That's why I am untarring the main tar file and extracting the tar files inside and renaming them.

hirenvadalia: @shilpakodli I know this and that's why I told that let the caller of the pbs_benchpress handle that, why PTL has to take care of it? PTL will simply give one tarball which has all info and caller will do whatever it wants to do with that tarball. If we want PTL to give extracted tarball then why pbs_snapshot is storing data in a tarball at all? we should change that right? But we don't want that :smile:

agrawalravi90: I agree with Hiren, PTL should just generate snapshot tarball, which might be used by external apps (like TH2) in whatever way they want. There's a lot of code here which untars the snapshot and the sub-snapshots inside it, I think you should remove all of it and instead add it to TH2.

test/fw/ptl/utils/plugins/ptl_test_data.py
```diff
@@ -149,160 +149,140 @@ def __save_home(self, test, status, err=None):
...
+        ret = self.du.run_cmd(server_host, cmd, level=logging.DEBUG2)
```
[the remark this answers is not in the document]
hirenvadalia: Oh yes, I completely forget out -H option in pbs_snapshot, @shilpakodli I agree with Ravi we should use -H option and run pbs_snapshot in current_host

shilpakodli:
@agrawalravi90 If we use -H option, then for multi node do we need to run pbs_snapshot command seperately for every host ? if so then we will have multiple tar files right ?

shilpakodli:
@agrawalravi90 and @hirenvadalia Thanks for the suggestions. I have addressed your review comments. Please let me know if change sneed to be done.

test/fw/ptl/utils/plugins/ptl_test_data.py
```diff
@@ -150,23 +150,26 @@ def __save_home(self, test, status, err=None):
...
             return
 
-        server = getattr(_test, 'server', None)
+        servers = getattr(_test, 'servers', None)
         moms = getattr(_test, 'moms', None)
         comms = getattr(_test, 'comms', None)
-        client = getattr(_test.server, 'client', None)
+        client = getattr(_test.servers, 'client', None)
         add_hosts = []
+        if len(servers) > 1:
```
agrawalravi90: I think you should add only servers[1:] values to add_hosts as you are already capturing the servers[0] host via -H

test/fw/ptl/utils/plugins/ptl_test_data.py
```diff
@@ -149,160 +149,140 @@ def __save_home(self, test, status, err=None):
...
+        comms = getattr(_test, 'comms', None)
+        client = getattr(_test.server, 'client', None)
+        add_hosts = []
+        if moms is not None:
+            for param in moms.values():
+                add_hosts.append(param.shortname)
+        if comms is not None:
+            for param_dict in comms.values():
+                add_hosts.append(param.shortname)
+        if client is not None:
```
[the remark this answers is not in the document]
agrawalravi90: Can you please clarify this?

test/fw/ptl/utils/plugins/ptl_test_data.py
```diff
@@ -149,160 +149,140 @@ def __save_home(self, test, status, err=None):
...
+            add_hosts.append(client.split('.')[0])
+        add_hosts = list(dict.fromkeys(add_hosts))
```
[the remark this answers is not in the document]
agrawalravi90: I think you didn't address this comment

test/fw/ptl/utils/plugins/ptl_test_data.py
```diff
@@ -150,23 +150,26 @@ def __save_home(self, test, status, err=None):
...
+            for param in comms.values():
                 add_hosts.append(param.shortname)
         if client is not None:
             add_hosts.append(client.split('.')[0])
         add_hosts = list(dict.fromkeys(add_hosts))
         add_hosts = [host for host in add_hosts if host != current_host]
-        if server is not None:
-            server_host = server.hostname
+        if servers is not None:
+            server_host = (servers.values()[0]).shortname
```
agrawalravi90: any particular reason why you are using 'shortname' ? I think 'hostname' will be safer to use, no? Same comment for other instances

test/fw/ptl/utils/plugins/ptl_test_data.py
```diff
@@ -175,24 +178,26 @@ def __save_home(self, test, status, err=None):
             f.close()
             return
         pbs_snapshot_path = os.path.join(
-            server.pbs_conf["PBS_EXEC"], "sbin", "pbs_snapshot")
+            servers.values()[0].pbs_conf["PBS_EXEC"], "sbin", "pbs_snapshot")
```
agrawalravi90: please save servers.values()[0] once and reuse it

test/fw/ptl/utils/plugins/ptl_test_data.py
```diff
@@ -149,160 +149,140 @@ def __save_home(self, test, status, err=None):
...
             _msg += ', skipping saving post analysis data'
             f.write(_msg + '\n')
             self.logger.warning(_msg)
             f.close()
             return
-        pbs_diag = os.path.join(svr.pbs_conf['PBS_EXEC'],
-                                'unsupported', 'pbs_diag')
+        pbs_snapshot_path = os.path.join(
+            server.pbs_conf["PBS_EXEC"], "sbin", "pbs_snapshot")
+        parent_dir = os.getcwd()
```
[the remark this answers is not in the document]
agrawalravi90: you are using "pwd.getpwnam(cur_user).pw_dir" now instead of 'parent_dir', so this should be removed

test/fw/ptl/utils/plugins/ptl_test_data.py
```diff
@@ -175,24 +178,26 @@ def __save_home(self, test, status, err=None):
...
+            '--with-sudo'
+            ]
         if len(add_hosts) > 0:
-            cmd += ['--additional-hosts=' + ','.join(map(str, add_hosts))]
-        cmd += ['-o', parent_dir]
-        cmd += ['> /dev/null', '2>&1']
-        ret = self.du.run_cmd(server_host, cmd, level=logging.DEBUG2)
+            cmd += ['--additional-hosts=' + ','.join(add_hosts)]
+        cmd += ['-o', pwd.getpwnam(cur_user).pw_dir]
+        ret = self.du.run_cmd(current_host, cmd, level=logging.DEBUG2,
```
agrawalravi90: Any particular reason for setting level to logging.DEBUG2 ?

test/fw/ptl/utils/plugins/ptl_test_data.py
```diff
@@ -204,76 +209,13 @@ def __save_home(self, test, status, err=None):
...
-        ret = self.du.run_cmd(current_host, cmd, level=logging.DEBUG2)
-        if ret['rc'] != 0:
-            _msg = 'Failed to tar %s' % snap_out_dest
-            _msg += '\n\n' + '\n'.join(ret['err']) + '\n\n'
-            f.write(_msg + '\n')
-            self.logger.error(_msg)
-            f.close()
-            return
+        snap_out = ret['out'][0]
+        snap_out_dest = (snap_out.split(":")[1]).strip(" ")
```
agrawalravi90: you don't need to provide " " to strip, it does that by default, so you can call it simply as 'strip()'

test/fw/ptl/utils/plugins/ptl_test_data.py
```diff
@@ -149,160 +149,82 @@ def __save_home(self, test, status, err=None):
...
-            f.close()
-            return
-        self.du.rm(current_host, path=diag_out_dest + '.tar.gz',
-                   force=True, sudo=True, level=logging.DEBUG2)
+        self.du.rm(current_host, path=snap_out_dest,
+                   recursive=True, force=True, level=logging.DEBUG2)
+
+        list_tgz_fles = os.listdir(self.post_data_dir)
+
+        for f_tgz in list_tgz_fles:
```
agrawalravi90: do you still need this?

shilpakodli:
@agrawalravi90 I have addressed your comments. Please review.

test/fw/ptl/utils/plugins/ptl_test_data.py
shilpakodli: Since, pbs_diag used to run with DEBUG2 logging, I have maintained the same.

test/fw/ptl/utils/plugins/ptl_test_data.py
```diff
@@ -151,9 +151,19 @@ def __save_home(self, test, status, err=None):
...
+            self.logger.warning(_msg)
+            f.close()
+            return
         moms = getattr(_test, 'moms', None)
         comms = getattr(_test, 'comms', None)
         client = getattr(_test.servers, 'client', None)
+        server = servers.values()[0]
         add_hosts = []
         if len(servers) > 1:
             for param in servers.values():
```
agrawalravi90: add_hosts will still add servers.values()[0], please replace 'servers' here with 'servers[1:]'

test/fw/ptl/utils/plugins/ptl_test_data.py
shilpakodli: Yes, if I pass the hostname then snapshot logs will be saved with FQDN name on some machines which is not a good practice to save logs with full hostname.

test/fw/ptl/utils/plugins/ptl_test_data.py
shilpakodli: I am going to address this comment. and also I will be removing the line add_hosts = list(set(add_hosts) - set(current_host)) as now pbs_snapshot is run irrespective of current_host.

agrawalravi90: ya but if somebody passes 'servers=<remote_host1>,<remote_host2>', then servers[0] will not be "current_host" right? In fact I think that can be quite useful, being able to capture data from remote hosts while running pbs_snapshot on the local client host.

shilpakodli: I have changed current_host to server_host to avoid duplicates.

agrawalravi90:
Thanks Shilpa, looks good, can you please add some logs to show this working?

test/fw/ptl/utils/plugins/ptl_test_data.py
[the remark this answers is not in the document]
shilpakodli: Hi Ravi, we need to capture logs on client as well though it doesn't have pbs directory in it. It still has conf file and system logs.

test/fw/ptl/utils/plugins/ptl_test_data.py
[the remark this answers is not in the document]
shilpakodli: I have changed this to add_host.remove(server_host)

agrawalravi90:
@shilpakodli the title of this PR, and the commit message, are kind of confusing. This PR is specific to post analysis data. So, please rename the title of the PR and the commit message (when you squash it after the final changes) to say something like "replace pbs_diag with pbs_snapshot for post-analysis data capturing"

agrawalravi90:
@shilpakodli are you still working on this PR? Can you please provide us an update?

shilpakodli:
@agrawalravi90 There are still some permission issues on testbed machines which Harry yet to fix. He is going to fix in few days, probably by next week. I will keep post updates on permission issues fix.
Thanks.

shilpakodli:
Hi @agrawalravi90, I have tested on testdeb machines using TH2, and logs were saved properly without any interrupt. We can go ahead with PR merge as permission issue is not a blocker in TH2. 
Only when you run manually through commandline, then it will asks for a password for dmesg.

test/fw/ptl/utils/plugins/ptl_test_data.py
```diff
@@ -149,164 +149,82 @@ def __save_home(self, test, status, err=None):
...
             _msg += ', skipping saving post analysis data'
             f.write(_msg + '\n')
             self.logger.warning(_msg)
             f.close()
             return
-        pbs_diag = os.path.join(svr.pbs_conf['PBS_EXEC'],
-                                'unsupported', 'pbs_diag')
+        moms = getattr(_test, 'moms', None)
+        comms = getattr(_test, 'comms', None)
+        client = getattr(_test.servers, 'client', None)
```
hirenvadalia: you need to get client from _test.servers.values()[0] not _test.servers

hirenvadalia:
> on second thought not just _test.servers but you need to iterate through _test.servers.values() and get all clients for all servers.
> Also, you can use the servers variable instead of _test.servers.

As per the offline discussion, this is not necessary for now since PTL doesn't support multiple clients. so I am ok with change as it is.

PR #1117, opened 2019-05-09
Enhanced save & load configuration methods for server object in PTL

RKORavi:
#### Describe Bug or Feature
Currently cPickle is used to dump configuration using save_configuration() & load_configuration() is used to load it, which does not work as intended. Pickle is slow, unsafe and is not human readable. Storing configurations in json format would be a better option.

#### Describe Your Change
PBSTestSuite's Server.save_configuration() method should save all the configurations of PBS server listed below:
Commands output:
qmgr print server
qmgr print sched
qmgr print resource
qmgr list pbshook
save hooks .CF, .PY & .HK files.
pbsnodes -av
PBS Configuration file:
pbs.conf

* Load_configuration is used to load saved above configurations

#### Link to Design Doc
 
https://pbspro.atlassian.net/wiki/spaces/PD/pages/1156284419/Update+file+format+and+file+name+for+data+saved+using+save+configuration+method+in+PTL

[custom_setup_logs.txt](https://github.com/PBSPro/pbspro/files/3593469/custom_setup_logs.txt)
[use_cur_setup OOB logs.txt](https://github.com/PBSPro/pbspro/files/3593470/use_cur_setup.OOB.logs.txt)

### DIFF
--- a/test/fw/bin/pbs_benchpress
+++ b/test/fw/bin/pbs_benchpress
@@ -171,6 +171,7 @@ def usage():
     msg += ['--list-tags: List all currenlty used tags\n']
     msg += [
         '--verbose: show verbose output (used with -i, -L or --tag-info)\n']
+    msg += ['--use-current-setup: Runs test on current PBS setup\n']
     msg += ['--version: show version number and exit\n']
 
     print(''.join(msg))
@@ -187,6 +188,7 @@ if __name__ == '__main__':
     outfile = None
     dbtype = None
     dbname = None
+    use_cur_setup = False
     dbaccess = None
     testfiles = None
     testsuites = None
@@ -232,7 +234,7 @@ if __name__ == '__main__':
     largs += ['stop-on-failure', 'enable-nose-debug']
     largs += ['post-analysis-data=', 'gen-ts-tree']
     largs += ['tc-failure-threshold=', 'cumulative-tc-failure-threshold=']
-    largs += ['max-postdata-threshold=', 'user-plugins=']
+    largs += ['max-postdata-threshold=', 'user-plugins=', 'use-current-setup']
 
     try:
         opts, args = getopt.getopt(
@@ -336,6 +338,8 @@ if __name__ == '__main__':
         elif o == '-h':
             usage()
             sys.exit(0)
+        elif o == '--use-current-setup':
+            use_cur_setup = True
         elif o == '--version':
             print(ptl.__version__)
             sys.exit(0)
@@ -454,9 +458,11 @@ if __name__ == '__main__':
         data = PTLTestData()
         loader.set_data(testgroup, testsuites, excludes, follow, testfiles)
         testtags.set_data(tags, eval_tags)
-        runner.set_data(paramfile, testparam, lcov_bin, lcov_data, lcov_out,
+        runner.set_data(paramfile, testparam, lcov_bin,
+                        lcov_data, lcov_out,
                         genhtml_bin, lcov_nosrc, lcov_baseurl,
-                        tc_failure_threshold, cumulative_tc_failure_threshold)
+                        tc_failure_threshold, cumulative_tc_failure_threshold,
+                        use_cur_setup)
         db.set_data(dbtype, dbname, dbaccess)
         data.set_data(post_data_dir, max_postdata_threshold)
         plugins = (loader, testtags, runner, db, data)
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -52,6 +52,9 @@ import tempfile
 import threading
 import time
 import traceback
+import json
+import base64
+import ast
 from collections import OrderedDict
 from distutils.version import LooseVersion
 from operator import itemgetter
@@ -65,16 +68,16 @@ from ptl.utils.pbs_testusers import ROOT_USER, TEST_USER, PbsUser
 try:
     import psycopg2
     PSYCOPG = True
-except BaseException:
+except:
     PSYCOPG = False
 
 try:
     from ptl.lib.pbs_ifl import *
     API_OK = True
-except BaseException:
+except:
     try:
         from ptl.lib.pbs_ifl_mock import *
-    except BaseException:
+    except:
         sys.stderr.write("failed to import pbs_ifl, run pbs_swigify " +
                          "to make it\n")
         raise ImportError
@@ -271,7 +274,7 @@ class PtlConfig(object):
                 k = k.strip()
                 v = v.strip()
                 self.options[k] = v
-            except BaseException:
+            except:
                 self.logger.error('Error parsing line ' + line)
         # below two if block are for backword compatibility
         if 'PTL_EXPECT_MAX_ATTEMPTS' in self.options:
@@ -866,7 +869,7 @@ class PbsTypeSelect(list):
             if 'mem' in k:
                 try:
                     v = PbsTypeSize(v).value
-                except BaseException:
+                except:
                     # failed so we guessed wrong on the type
                     pass
             if isinstance(v, int) or v.isdigit():
@@ -2185,7 +2188,7 @@ class BatchUtils(object):
         """
         try:
             ret = time.mktime(time.strptime(st, '%a %b %d %H:%M:%S %Y'))
-        except BaseException:
+        except:
             ret = st
         return ret
 
@@ -2209,7 +2212,7 @@ class BatchUtils(object):
                                                             fmt=_f)
                 dtime_to = self.convert_datetime_to_epoch(m.group('dt_to'),
                                                           fmt=_f)
-            except BaseException:
+            except:
                 self.logger.error('error converting dedicated time')
         return (dtime_from, dtime_to)
 
@@ -2300,7 +2303,7 @@ class BatchUtils(object):
         :type op: str or None
         :param hostname: The name of the host on which to operate
         :type hostname: str or None
-        :param dflt_conf: Whether we are using the default PBS
+        :param dflt_conf: Whether we are using the default PBS
                           configuration
         :type dflt_conf: bool
         :param exclude_attrs: Optional list of attributes to not
@@ -3448,7 +3451,7 @@ class PBSService(PBSObject):
                     self.logger.info('FQDN name ' + self.fqdn + ' differs '
                                      'from name provided ' + self.hostname)
                     self.hostname = self.fqdn
-            except BaseException:
+            except:
                 pass
         else:
             self.fqdn = self.hostname
@@ -4004,7 +4007,7 @@ class PBSService(PBSObject):
         """
         try:
             from ptl.utils.pbs_logutils import PBSLogUtils
-        except BaseException:
+        except:
             _msg = 'error loading ptl.utils.pbs_logutils'
             raise ImportError(_msg)
 
@@ -4060,7 +4063,7 @@ class PBSService(PBSObject):
             # an open file descriptor, we close here but ignore errors in case
             # any were raised for all irrelevant cases
             lines.close()
-        except BaseException:
+        except:
             pass
         if (rv is None and existence) or (rv is not None and not existence):
             _msg = infomsg + attemptmsg
@@ -4199,7 +4202,7 @@ class PBSService(PBSObject):
 
     def _load_configuration(self, infile, objtype=None):
         """
-        Load configuration as was saved in infile
+        Load server configuration as was saved in infile
 
         :param infile: the file in which configuration
                        was saved
@@ -4211,45 +4214,99 @@ class PBSService(PBSObject):
         if os.path.isfile(infile):
             conf = {}
             with open(infile, 'r') as f:
-                # load all objects from the Pickled file
-                while True:
-                    try:
-                        conf = cPickle.load(f)
-                    except BaseException:
-                        break
-
-            if objtype and objtype in conf:
-                conf = conf[objtype]
-            else:
-                # load all object types that could be in infile
-                newconf = {}
-                for ky in [MGR_OBJ_SERVER, MGR_OBJ_SCHED, MGR_OBJ_NODE]:
-                    if ky not in conf:
-                        conf[ky] = {}
-                    newconf = {**newconf, **conf[ky]}
-                conf = newconf
-
+                try:
+                    conf = json.load(f)
+                except ValueError:
+                    self.logger.info("Error loading JSON file: %s"
+                                     % infile)
+                    return False
+            qmgr = os.path.join(self.client_conf['PBS_EXEC'],
+                                'bin', 'qmgr')
             for k, v in conf.items():
-                fn = self.du.create_temp_file()
-                with open(fn, 'w') as fd:
-                    # handle server data saved as output of qmgr commands
-                    # by piping data back into qmgr
-                    if k.startswith('qmgr_'):
-                        qmgr = os.path.join(self.client_conf['PBS_EXEC'],
-                                            'bin', 'qmgr')
-                        fd.write("\n".join(v))
-                        self.du.run_cmd(self.hostname, [qmgr], cstdin=fd,
-                                        sudo=True)
-                    else:
-                        fd.write("\n".join(v))
-                        # append the last line
-                        fd.write("\n")
-                        self.du.run_cmd(self.hostname, ['cp', fn, k],
-                                        sudo=True)
-                    os.remove(fn)
-
+                # Load server configuration
+                if k.startswith('qmgr_'):
+                    fpath = self.du.create_temp_file()
+                    print_svr = '\n'.join(v)
+                    with open(fpath, 'w') as f:
+                        f.write(print_svr)
+                    file_qmgr = open(fpath)
+                    d = self.du.run_cmd(
+                        self.hostname, [qmgr], stdin=file_qmgr, sudo=True,
+                        logerr=False, level=logging.DEBUG)
+                    err_msg = "Failed to load server configurations"
+                    file_qmgr.close()
+                    if d['rc'] != 0:
+                        self.logger.error("%s" % err_msg)
+                        return False
+                # Load pbs.conf file
+                elif k == "pbs_conf":
+                    enc_utf = v.encode('UTF-8')
+                    dec_b64 = base64.b64decode(enc_utf)
+                    cfg_vals = dec_b64.decode('UTF-8')
+                    config = ast.literal_eval(cfg_vals)
+                    self.du.set_pbs_config(self.hostname, confs=config)
+                # Load hooks
+                elif k == "hooks":
+                    fpath = self.du.create_temp_file()
+                    print_hooks = '\n'.join(v)
+                    with open(fpath, 'w') as f:
+                        f.write(print_hooks)
+                    file_qmgr = open(fpath)
+                    d = self.du.run_cmd(
+                        self.hostname, [qmgr], stdin=file_qmgr, sudo=True,
+                        level=logging.DEBUG)
+                    file_qmgr.close()
+                    if d['rc'] != 0:
+                        self.logger.error("Failed to load site hooks")
+            if 'pbsnodes' in conf:
+                nodes = conf['pbsnodes']
+                for node in nodes:
+                    node_name = str(node['id'])
+                    nodes_created = self.create_pbsnode(node_name, node)
+                    if not nodes_created:
+                        self.logger.error("Failed to create node: %s"
+                                          % node)
+                        return False
             return True
         return False
+
+    def create_pbsnode(self, node_name, attrs):
+        """
+        Create node in PBS with given attributes
+        """
+        qmgr = os.path.join(self.client_conf['PBS_EXEC'],
+                            'bin', 'qmgr')
+        execcmd = "create node " + node_name
+        execcmd += " Port=" + attrs['Port']
+        cmd = [qmgr, "-c", execcmd]
+        ret = self.du.run_cmd(self.hostname, cmd, sudo=True)
+        if ret['rc'] != 0:
+            self.logger.info("Failed to create node: %s" % node_name)
+            self.logger.error("Error: %s" % ret['err'])
+            return False
+        # skip all read-only attributes
+        skip_atb_list = ['id', 'pbs_version', 'pcpus',
+                         'last_state_change_time', 'ntype',
+                         'Mom', 'sharing', 'resources_available.vnode',
+                         'resources_available.host', 'last_used_time',
+                         'resource_assigned', 'resv', 'Port'
+                         ]
+        for node_atb, val in attrs.items():
+            # only offline state of node is read, write attribute
+            if(node_atb in skip_atb_list or
+               'resources_assigned' in node_atb or
+               (node_atb == 'state' and val != 'offline')):
+                continue
+            k = str(node_atb)
+            v = str(val)
+            execcmd = "set node %s %s=%s" % (node_name, k, v)
+            cmd = [qmgr, "-c", execcmd]
+            ret = self.du.run_cmd(self.hostname, cmd, sudo=True,
+                                  level=logging.DEBUG)
+            if ret['rc'] != 0:
+                self.logger.info("Failed to set node attribute %s=%s" % (k, v))
+                return False
+        return True
 
     def get_tempdir(self):
         """
@@ -4982,13 +5039,6 @@ class Server(PBSService):
                   encountered.
         :raises: PbsStatusError or PbsManagerError
         """
-        ignore_attrs = ['id', 'pbs_license', ATTR_NODE_ProvisionEnable]
-        ignore_attrs += [ATTR_status, ATTR_total, ATTR_count]
-        ignore_attrs += [ATTR_rescassn, ATTR_FLicenses, ATTR_SvrHost]
-        ignore_attrs += [ATTR_license_count, ATTR_version, ATTR_managers]
-        ignore_attrs += [ATTR_operators]
-        ignore_attrs += [ATTR_pbs_license_info, ATTR_power_provisioning]
-        unsetlist = []
         setdict = {}
         skip_site_hooks = ['pbs_cgroups']
         self.logger.info(self.logprefix +
@@ -5001,31 +5051,7 @@ class Server(PBSService):
                                           'lib', 'python', 'altair',
                                           'pbs_hooks',
                                           'PBS_translate_mpp.HK')
-        self.jacs_hk = os.path.join(self.pbs_conf['PBS_HOME'],
-                                    'server_priv', 'hooks',
-                                    'PBS_cray_jacs.HK')
-        self.dflt_jacs_hk = os.path.join(self.pbs_conf['PBS_EXEC'],
-                                         'lib', 'python', 'altair',
-                                         'pbs_hooks',
-                                         'PBS_cray_jacs.HK')
-        self.jacs_cf = os.path.join(self.pbs_conf['PBS_HOME'],
-                                    'server_priv', 'hooks',
-                                    'PBS_cray_jacs.CF')
-        self.dflt_jacs_cf = os.path.join(self.pbs_conf['PBS_EXEC'],
-                                         'lib', 'python', 'altair',
-                                         'pbs_hooks',
-                                         'PBS_cray_jacs.CF')
-        if server_stat is None:
-            server_stat = self.status(SERVER, level=logging.DEBUG)[0]
-        for k in server_stat.keys():
-            if (k in ignore_attrs) or (k in self.dflt_attributes.keys()):
-                continue
-            elif (('.' in k) and (k.split('.')[0] in ignore_attrs)):
-                continue
-            else:
-                unsetlist.append(k)
-        if len(unsetlist) != 0:
-            self.manager(MGR_CMD_UNSET, MGR_OBJ_SERVER, unsetlist)
+        self.unset_svr_attrib()
         for k in self.dflt_attributes.keys():
             if(k not in self.attributes or
                self.attributes[k] != self.dflt_attributes[k]):
@@ -5038,53 +5064,20 @@ class Server(PBSService):
                 reverthooks = True
             else:
                 reverthooks = False
-            hooks = self.status(HOOK, level=logging.DEBUG)
-            hooks = [h['id'] for h in hooks]
-            for h in skip_site_hooks:
-                if h in hooks:
-                    hooks.remove(h)
-            if len(hooks) > 0:
-                self.manager(MGR_CMD_DELETE, HOOK, id=hooks)
+            self.delete_site_hooks()
         if delqueues:
             revertqueues = False
-            queues = self.status(QUEUE, level=logging.DEBUG)
-            queues = [q['id'] for q in queues]
-            if len(queues) > 0:
-                try:
-                    nodes = self.status(VNODE, logerr=False)
-                    for node in nodes:
-                        if 'queue' in node.keys():
-                            self.manager(MGR_CMD_UNSET, NODE, 'queue',
-                                         node['id'])
-                except BaseException:
-                    pass
-                self.manager(MGR_CMD_DELETE, QUEUE, id=queues)
+            self.delete_queues()
             a = {ATTR_qtype: 'Execution',
                  ATTR_enable: 'True',
                  ATTR_start: 'True'}
             self.manager(MGR_CMD_CREATE, QUEUE, a, id='workq')
             setdict.update({ATTR_dfltque: 'workq'})
         if delscheds:
-            self.manager(MGR_CMD_LIST, SCHED)
-            for name in list(self.schedulers.keys()):
-                if name != 'default':
-                    self.schedulers[name].terminate()
-                    sched_log = self.schedulers[
-                        name].attributes['sched_log']
-                    sched_priv = self.schedulers[
-                        name].attributes['sched_priv']
-                    self.du.rm(path=sched_log, sudo=True,
-                               recursive=True, force=True)
-                    self.du.rm(path=sched_priv, sudo=True,
-                               recursive=True, force=True)
-                    self.manager(MGR_CMD_DELETE, SCHED, id=name)
+            self.delete_sched_config()
 
         if delnodes:
-            try:
-                self.manager(MGR_CMD_DELETE, VNODE, id="@default")
-            except PbsManagerError as e:
-                if "Unknown node" not in e.msg[0]:
-                    raise
+            self.delete_nodes()
         if reverthooks:
             if self.platform == 'cray' or self.platform == 'craysim':
                 if self.du.cmp(self.hostname, self.dflt_mpp_hook,
@@ -5139,6 +5132,87 @@ class Server(PBSService):
                 self.manager(MGR_CMD_DELETE, RSC, id=rescs)
         return True
 
+    def unset_svr_attrib(self, server_stat=None):
+        """
+        Unset server attributes
+        """
+        ignore_attrs = ['id', 'pbs_license', ATTR_NODE_ProvisionEnable]
+        ignore_attrs += [ATTR_status, ATTR_total, ATTR_count]
+        ignore_attrs += [ATTR_rescassn, ATTR_FLicenses, ATTR_SvrHost]
+        ignore_attrs += [ATTR_license_count, ATTR_version, ATTR_managers]
+        ignore_attrs += [ATTR_pbs_license_info, ATTR_power_provisioning]
+        unsetlist = []
+        self.cleanup_jobs_and_reservations()
+        if server_stat is None:
+            server_stat = self.status(SERVER, level=logging.DEBUG)[0]
+        for k in server_stat.keys():
+            if (k in ignore_attrs) or (k in self.dflt_attributes.keys()):
+                continue
+            elif (('.' in k) and (k.split('.')[0] in ignore_attrs)):
+                continue
+            else:
+                unsetlist.append(k)
+        if len(unsetlist) != 0:
+            self.manager(MGR_CMD_UNSET, MGR_OBJ_SERVER, unsetlist)
+
+    def delete_site_hooks(self):
+        """
+        Delete site hooks from PBS
+        """
+        skip_site_hooks = ['pbs_cgroups']
+        hooks = self.status(HOOK, level=logging.DEBUG)
+        hooks = [h['id'] for h in hooks]
+        for h in skip_site_hooks:
+            if h in hooks:
+                hooks.remove(h)
+        if len(hooks) > 0:
+            self.manager(MGR_CMD_DELETE, HOOK, id=hooks)
+
+    def delete_queues(self):
+        """
+        Delete queues
+        """
+        queues = self.status(QUEUE, level=logging.DEBUG)
+        queues = [q['id'] for q in queues]
+        if len(queues) > 0:
+            try:
+                nodes = self.status(VNODE, logerr=False)
+                for node in nodes:
+                    if 'queue' in node.keys():
+                        self.manager(MGR_CMD_UNSET, NODE, 'queue',
+                                     node['id'])
+            except BaseException:
+                pass
+            self.manager(MGR_CMD_DELETE, QUEUE, id=queues)
+
+    def delete_sched_config(self):
+        """
+        Delete sched_priv & sched_log files
+        """
+        self.manager(MGR_CMD_LIST, SCHED)
+        for name in list(self.schedulers.keys()):
+            if name != 'default':
+                self.schedulers[name].terminate()
+                sched_log = self.schedulers[
+                    name].attributes['sched_log']
+                sched_priv = self.schedulers[
+                    name].attributes['sched_priv']
+                self.du.rm(path=sched_log, sudo=True,
+                           recursive=True, force=True)
+                self.du.rm(path=sched_priv, sudo=True,
+                           recursive=True, force=True)
+                self.manager(MGR_CMD_DELETE, SCHED, id=name)
+
+    def delete_nodes(self):
+        """
+        Remove all the nodes from PBS
+        """
+        try:
+            self.manager(MGR_CMD_DELETE, VNODE, id="@default")
+        except PbsManagerError as e:
+            if "Unknown node" not in e.msg[0]:
+                raise
+
     def save_configuration(self, outfile, mode='a'):
         """
         Save a server configuration, this includes:
@@ -5164,47 +5238,85 @@ class Server(PBSService):
         """
         conf = {}
         sconf = {MGR_OBJ_SERVER: conf}
-
-        rd = os.path.join(self.pbs_conf['PBS_HOME'], 'server_priv',
-                          'resourcedef')
-        self._save_config_file(conf, rd)
-
+        # save pbs.conf file
+        cfg_path = self.du.get_pbs_conf_file()
+        with open(cfg_path, 'r') as p:
+            pbs_cfg = p.readlines()
+            config = self.utils.convert_to_dictlist(pbs_cfg)
+            cfg_str = str(config[0])
+            encode_utf = cfg_str.encode('UTF-8')
+            pbs_cfg_b64 = base64.b64encode(encode_utf)
+            decode_utf = pbs_cfg_b64.decode('UTF-8')
+        conf['pbs_conf'] = decode_utf
+        # save hook files
+        hooks_str = self._save_hook_files()
+        if hooks_str:
+            conf.update(hooks_str)
+            conf['hooks'] = hooks_str
+        else:
+            self.logger.error('Failed to save site hooks')
+            return False
         qmgr = os.path.join(self.client_conf['PBS_EXEC'], 'bin', 'qmgr')
-
-        ret = self.du.run_cmd(self.client, [qmgr, '-c', 'p s'], sudo=True)
+        pbsnodes = os.path.join(
+            self.client_conf['PBS_EXEC'], 'bin', 'pbsnodes')
+        ret = self.du.run_cmd(
+            self.hostname, [
+                qmgr, '-c', 'print server'], sudo=True,
+            logerr=False, level=logging.DEBUG)
         if ret['rc'] != 0:
+            self.logger.error('Failed to get Server attributes')
             return False
         else:
             conf['qmgr_print_server'] = ret['out']
-
-        ret = self.du.run_cmd(self.hostname, [qmgr, '-c', 'p sched'],
-                              sudo=True)
+        ret = self.du.run_cmd(self.hostname, [qmgr, '-c', 'print sched'],
+                              logerr=False, level=logging.DEBUG, sudo=True)
         if ret['rc'] != 0:
+            self.logger.error('Failed to get sched attributes')
             return False
         else:
             conf['qmgr_print_sched'] = ret['out']
-
-        ret = self.du.run_cmd(self.hostname, [qmgr, '-c', 'p h'], sudo=True)
-        if ret['rc'] != 0:
+        ret = self.du.run_cmd(self.hostname, [pbsnodes, '-av'],
+                              logerr=False, level=logging.DEBUG)
+        err_msg = "Server has no node list"
+        # pbsnodes -av returns a non zero exit code when there are
+        # no nodes in cluster
+        if ret['rc'] != 0 and err_msg in ret['err']:
+            self.logger.error('Failed to get nodes info')
             return False
         else:
-            conf['qmgr_print_hook'] = ret['out']
-
+            nodes_val = self.utils.convert_to_dictlist(ret['out'])
+            conf['pbsnodes'] = nodes_val
         try:
             with open(outfile, mode) as f:
-                cPickle.dump(sconf, f)
+                json.dump(conf, f)
         except BaseException:
             self.logger.error('Error processing file ' + outfile)
             return False
 
         return True
 
+    def _save_hook_files(self):
+        """
+        save all the hooks .CF, .PY, .HK files
+        """
+        qmgr = os.path.join(self.client_conf['PBS_EXEC'], 'bin', 'qmgr')
+        cfg = {"hooks": ""}
+        cmd = [qmgr, '-c', 'print hook @default']
+        ret = self.du.run_cmd(self.hostname, cmd,
+                              sudo=True)
+        if ret['rc'] != 0:
+            self.logger.error('Failed to save hook files ')
+            return False
+        else:
+            cfg['qmgr_print_hook'] = ret['out']
+        return cfg
+
     def load_configuration(self, infile):
         """
         load configuration from saved file ``infile``
         """
-        self.revert_to_defaults()
-        self._load_configuration(infile, MGR_OBJ_SERVER)
+        rv = self._load_configuration(infile, MGR_OBJ_SERVER)
+        return rv
 
     def get_hostname(self):
         """
@@ -5708,7 +5820,7 @@ class Server(PBSService):
                             elif obj_type == PBS_HOOK:
                                 return [h.attributes for h in
                                         self.pbshooks.values()]
-                    except BaseException:
+                    except:
                         pass
                 else:
                     bs = pbs_stathook(c, id, a, extend)
@@ -7824,7 +7936,7 @@ class Server(PBSService):
                 name = e0
                 r = Resource(name, rtype, flag)
                 resources[name] = r
-        except BaseException:
+        except:
             raise PbsResourceError(rc=1, rv=False,
                                    msg="error in parse_resources")
         return resources
@@ -8901,7 +9013,7 @@ class Server(PBSService):
                         if a.endswith('mem'):
                             try:
                                 amt = PbsTypeSize().encode(amt)
-                            except BaseException:
+                            except:
                                 # we guessed the type incorrectly
                                 pass
                     else:
@@ -9008,7 +9120,7 @@ class Server(PBSService):
                             avail_nodes_by_time[tm].append(nodes[n])
                             try:
                                 nodes_id.remove(n)
-                            except BaseException:
+                            except:
                                 pass
                         else:
                             ncopy = copy.copy(nodes[n])
@@ -9045,7 +9157,7 @@ class Server(PBSService):
                                 avail_nodes_by_time[tm].append(nodes[n])
                                 try:
                                     nodes_id.remove(n)
-                                except BaseException:
+                                except:
                                     pass
                             else:
                                 ncopy = copy.copy(nodes[n])
@@ -9311,7 +9423,7 @@ class Server(PBSService):
         if createnode:
             try:
                 statm = self.status(NODE, id=natvnode)
-            except BaseException:
+            except:
                 statm = []
             if len(statm) >= 1:
                 _m = 'Mom %s already exists, not creating' % (natvnode)
@@ -9488,7 +9600,7 @@ class Server(PBSService):
         srv_stat = self.status(SERVER, 'sync_mom_hookfiles_timeout')
         try:
             sync_val = srv_stat[0]['sync_mom_hookfiles_timeout']
-        except BaseException:
+        except:
             self.logger.info("Setting sync_mom_hookfiles_timeout to 15s")
             self.manager(MGR_CMD_SET, SERVER,
                          {"sync_mom_hookfiles_timeout": 15})
@@ -10223,7 +10335,7 @@ class Server(PBSService):
                               snapmap=self.snapmap)
             try:
                 svr.manager(MGR_CMD_DELETE, NODE, None, id="")
-            except BaseException:
+            except:
                 pass
             svr.revert_to_defaults(delqueues=True, delhooks=True)
             local = svr.pbs_conf['PBS_HOME']
@@ -10264,7 +10376,7 @@ class Server(PBSService):
             for a in ['pbs_license_info', 'mail_from', 'acl_hosts']:
                 try:
                     svr.manager(MGR_CMD_UNSET, SERVER, a, sudo=True)
-                except BaseException:
+                except:
                     pass
 
             for (d, l) in _fcopy:
@@ -10298,7 +10410,7 @@ class Server(PBSService):
                 if vdef:
                     try:
                         svr.manager(MGR_CMD_DELETE, NODE, None, "")
-                    except BaseException:
+                    except:
                         pass
                     MoM(h, pbsconf_file=conf_file).insert_vnode_def(vdef)
                     svr.restart()
@@ -10950,7 +11062,7 @@ class Scheduler(PBSService):
             conf_opts = self.du.cat(self.hostname, schd_cnfg,
                                     sudo=(not self.has_snap),
                                     level=logging.DEBUG2)['out']
-        except BaseException:
+        except:
             self.logger.error('error parsing scheduler configuration')
             return False
 
@@ -11100,7 +11212,7 @@ class Scheduler(PBSService):
             os.remove(fn)
 
             self.logger.debug(self.logprefix + "updated configuration")
-        except BaseException:
+        except:
             m = self.logprefix + 'error in apply_config '
             self.logger.error(m + str(traceback.print_exc()))
             raise PbsSchedConfigError(rc=1, rv=False, msg=m)
@@ -12033,7 +12145,7 @@ class Scheduler(PBSService):
                     (dtime_from, dtime_to) = self.utils.convert_dedtime(line)
                     self.dedicated_time.append({'from': dtime_from,
                                                 'to': dtime_to})
-        except BaseException:
+        except:
             self.logger.error('error in parse_dedicated_time')
             return None
 
@@ -12113,7 +12225,7 @@ class Scheduler(PBSService):
             self.du.run_copy(self.hostname, fn, ddfile, sudo=True,
                              preserve_permission=False)
             os.remove(fn)
-        except BaseException:
+        except:
             raise PbsSchedConfigError(rc=1, rv=False,
                                       msg='error adding dedicated time')
 
@@ -12183,7 +12295,7 @@ class Scheduler(PBSService):
         """
         try:
             from ptl.utils.pbs_logutils import PBSSchedulerLog
-        except BaseException:
+        except:
             self.logger.error('error loading ptl.utils.pbs_logutils')
             return None
 
@@ -13267,7 +13379,7 @@ class MoM(PBSService):
                         self.config[k] = [self.config[k], v]
                 else:
                     self.config[k] = v
-        except BaseException:
+        except:
             self.logger.error('error in parse_config')
             return None
 
@@ -13356,7 +13468,7 @@ class MoM(PBSService):
             self.du.run_copy(self.hostname, fn, dest,
                              preserve_permission=False, sudo=True)
             os.remove(fn)
-        except BaseException:
+        except:
             raise PbsMomConfigError(rc=1, rv=False,
                                     msg='error processing add_config')
         if restart:
@@ -13397,7 +13509,7 @@ class MoM(PBSService):
         """
         try:
             fn = self.du.create_temp_file(self.hostname, body=vdef)
-        except BaseException:
+        except:
             raise PbsMomConfigError(rc=1, rv=False,
                                     msg="Failed to insert vnode definition")
         if fname is None:
@@ -13601,7 +13713,7 @@ class MoM(PBSService):
         :param host: the hostname on which dyn res script is created
         :type host: str or None
         :param perm: perm to use while creating scripts
-                     (must be octal like 0777)
+                     (must be octal like 0o777)
         :param prefix: the file name will begin with this prefix
         :type prefix: str
         :param suffix: the file name will end with this suffix
@@ -13864,7 +13976,7 @@ class Job(ResourceResv):
         else:
             try:
                 homedir = pwd.getpwnam(user)[5]
-            except BaseException:
+            except:
                 homedir = ""
 
         self.username = user
@@ -14127,7 +14239,7 @@ class InteractiveJob(threading.Thread):
         """
         try:
             import pexpect
-        except BaseException:
+        except:
             self.logger.error('pexpect module is required for '
                               'interactive jobs')
             return None
@@ -14164,7 +14276,7 @@ class InteractiveJob(threading.Thread):
             self.job.interactive_handle = _p
             time.sleep(_st)
             expstr = "qsub: waiting for job "
-            expstr += r"(?P<jobid>\d+.[0-9A-Za-z-.]+) to start"
+            expstr += "(?P<jobid>\d+.[0-9A-Za-z-.]+) to start"
             _p.expect(expstr)
             if _p.match:
                 self.jobid = _p.match.group('jobid').decode()
--- a/test/fw/ptl/utils/pbs_testsuite.py
+++ b/test/fw/ptl/utils/pbs_testsuite.py
@@ -430,6 +430,7 @@ class PBSTestSuite(unittest.TestCase):
     del_queues = True
     del_scheds = True
     del_vnodes = True
+    config_saved = False
     server = None
     scheduler = None
     mom = None
@@ -448,6 +449,15 @@ class PBSTestSuite(unittest.TestCase):
         cls.init_param()
         cls.check_users_exist()
         cls.init_servers()
+        if cls.use_cur_setup:
+            _, path = tempfile.mkstemp(prefix="saved_custom_setup",
+                                       suffix=".json")
+            ret = cls.server.save_configuration(path, 'w')
+            if ret:
+                cls.saved_file = path
+            else:
+                cls.logger.error("Failed to save custom setup")
+                raise Exception("Failed to save custom setup")
         cls.init_comms()
         cls.init_schedulers()
         cls.init_moms()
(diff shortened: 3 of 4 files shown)


test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4247,45 +4250,71 @@ def _load_configuration(self, infile, objtype=None):
...
+                    #    fd.write("#!/bin/bash\n")
+                    #    fd.write("\n".join(v))
+                    fd, path = tempfile.mkstemp(prefix="runcmd_")
+                    os.chmod(path, 0o700)
+                    with open(path, 'a') as f:
+                        f.write('\n'.join(v))
+                    qmgr_cmd = qmgr +" < "+path
+                    d = self.du.run_cmd(self.hostname, qmgr_cmd, sudo=True, as_script=True)
+                    os.unlink(path)
+                    # Load site hooks
```
SudeshnaMoulik: As below else block is handling pbs conf loading, can you please modify the comment

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4247,45 +4250,97 @@ def _load_configuration(self, infile, objtype=None):
...
+                                with open(fn, 'w') as fd:
+                                    fd.write(hook_py_str)
+                                i_h = " -c 'i h %s application/x-python"
+                                i_h += " default %s'"
+                                import_py_cmd = (qmgr + "%s" % i_h
+                                                 % (hook_name, fn))
+                                self.du.run_cmd(
+                                    self.hostname, import_py_cmd,
+                                    sudo=True, as_script=True)
+                                # os.unlink(fn)
```
ashwathraop: Remove this line if not intended to use.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4247,45 +4250,97 @@ def _load_configuration(self, infile, objtype=None):
...
+                                with open(fn, 'w') as fd:
+                                    fd.write(hook_cf_str)
+                                i_cfg = " -c 'i h %s application/x-config"
+                                i_cfg += " default %s'"
+                                import_cfg_cmd = (qmgr + "%s" % i_cfg
+                                                  % (hook_name, fn))
+                                self.du.run_cmd(
+                                    self.hostname, import_cfg_cmd,
+                                    sudo=True, as_script=True)
+                                # os.unlink(fn)
```
ashwathraop: Remove this line if not intended to use.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4247,45 +4250,97 @@ def _load_configuration(self, infile, objtype=None):
...
+                                    if k == 'event':
+                                        e_v = val.split(',')
+                                        for v in e_v:
+                                            cmd = qmgr + \
+                                                " -c 's h %s event+=%s'" % (
+                                                    hook_name, v)
+                                            self.du.run_cmd(
+                                                self.hostname, cmd,
+                                                sudo=True, as_script=True)
+                                    cmd = (qmgr +
```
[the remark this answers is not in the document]
RKORavi: I am loading hook here, i was unable to assign multiple events in one command. So, i had to run a loop of all the hook events & add them one by one.
E.g;
cgroup_hooks event = execjob_begin,execjob_epilogue,execjob_end,execjob_launch,
	execjob_attach,
	execjob_resize,
	exechost_periodic,
	exechost_startup

Running a loop of all the events here and running a qmgr command for each event.
for h_event in events:
        qmgr -c "s h hook_name event += h_event

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4247,45 +4250,97 @@ def _load_configuration(self, infile, objtype=None):
...
+                    with open(path, 'a') as f:
+                        f.write('\n'.join(v))
+                    qmgr_cmd = qmgr + " < " + path
+                    d = self.du.run_cmd(
+                        self.hostname, qmgr_cmd, sudo=True, as_script=True)
+                    os.unlink(path)
+                    # Load pbs.conf file
+                elif k == "pbs_conf":
+                    f_w = base64.decodestring(v)
+                    conf_path = os.path.join(os.sep, 'etc', 'pbs.conf')
```
sujatapatnaik52: would recommend to try `PBS_CONF_PATH` first else go for `/etc/pbs.conf`.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4247,45 +4250,97 @@ def _load_configuration(self, infile, objtype=None):
...
+                    cmd = f_w + " > " + path
+                    self.du.run_cmd(
+                        self.hostname, cmd, sudo=True, as_script=True)
+                # Load hooks
+                elif k == "hooks":
+                    for hpath in conf['hooks']:
+                        h_name = hpath.split(os.sep)[-1]
+                        hook_name = h_name[:-3]
+                        hook_name = str(hook_name)
```
sujatapatnaik52:
```suggestion
                        hook_name = str(h_name[:-3])
```

test/fw/ptl/lib/pbs_testlib.py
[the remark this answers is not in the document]
sujatapatnaik52: It is possible to use though, you may need to use some additional string formatters - `qmgr -c "c h move_hook_1 event='movejob,queuejob,runjob'"`. If it is **possible**, try to use one command.

test/fw/ptl/lib/pbs_testlib.py
sujatapatnaik52: Similar set of comments at line 4261. Have a look at them and remove it.

test/fw/ptl/lib/pbs_testlib.py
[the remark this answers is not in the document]
sujatapatnaik52: Also there is an API `create_and_import_hook` recently added. Please have a look at the same if it will be of any use here.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -5222,42 +5277,73 @@ def save_configuration(self, outfile, mode='a'):
...
-
+        try:
+            with open(rd, 'r') as rdef_file:
+                rdef_str = rdef_file.read()
+            b64_rdef = base64.encodestring(rdef_str)
+            conf['resourcedef_file'] = b64_rdef
+        except BaseException:
+            print("Resourcedef file not found")
+        # save pbs.conf file
+        pcfg = os.path.join(os.sep, 'etc', 'pbs.conf')
```
sujatapatnaik52: Same comment as above.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4247,45 +4250,85 @@ def _load_configuration(self, infile, objtype=None):
...
+                            hk_contents = base64.decodestring(
+                                conf['hooks'][hk_hpath])
+                            hook_py_b64str = conf['hooks'][hpath]
+                            hook_py_str = base64.decodestring(
+                                hook_py_b64str)
+                            hk_vals = self.utils.convert_to_dictlist(
+                                hk_contents.splitlines())
+                            for k, val in hk_vals[0].items():
+                                if k == 'event':
+                                    events = "'%s'" % val
```
npadole20: Where is the 'events' variable being used?
Also, the function Sujata is referring to is "create_import_hook".

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -5222,42 +5263,74 @@ def save_configuration(self, outfile, mode='a'):
         """
         conf = {}
         sconf = {MGR_OBJ_SERVER: conf}
-
+        # save resourcedef file
         rd = os.path.join(self.pbs_conf['PBS_HOME'], 'server_priv',
                           'resourcedef')
-        self._save_config_file(conf, rd)
-
+        try:
```
agrawalravi90: I didn't see code to load resourcedef file?

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -5222,42 +5263,74 @@ def save_configuration(self, outfile, mode='a'):
...
+            with open(rd, 'r') as rdef_file:
+                rdef_str = rdef_file.read()
+            b64_rdef = base64.encodestring(rdef_str)
+            conf['resourcedef_file'] = b64_rdef
+        except BaseException:
+            self.logger.info("Resourcedef file not found")
+        # save pbs.conf file
+        cfg_path = os.path.join(os.sep, 'etc', 'pbs.conf')
+        with open(cfg_path, 'r') as p:
```
agrawalravi90: As i suggested earlier, can you try to encode the pbs.conf dict itself (server.pbs_conf) so you can decode it into a dict which can be passed to set_pbs_config when loading it back?

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -5222,42 +5263,74 @@ def save_configuration(self, outfile, mode='a'):
...
         else:
             conf['qmgr_print_sched'] = ret['out']
-
-        ret = self.du.run_cmd(self.hostname, [qmgr, '-c', 'p h'], sudo=True)
-        if ret['rc'] != 0:
-            return False
-        else:
-            conf['qmgr_print_hook'] = ret['out']
-
+        ret = self.du.run_cmd(self.hostname, [pbsnodes, '-av'])
```
agrawalravi90: I didn't see equivalent code to load pbsnodes output?

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -5222,42 +5263,74 @@ def save_configuration(self, outfile, mode='a'):
...
-            cPickle.dump(sconf, f)
-            f.close()
-        except:
+            with open(outfile, mode) as f:
+                json.dump(conf, f)
+        except BaseException:
             self.logger.error('Error processing file ' + outfile)
             return False
 
         return True
```
agrawalravi90: don't we need to save sched configuration files (sched_config, holidays, etc.? )

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4247,45 +4250,83 @@ def _load_configuration(self, infile, objtype=None):
...
+                    fd, path = tempfile.mkstemp(prefix="runcmd_")
+                    os.chmod(path, 0o700)
+                    with open(path, 'a') as f:
+                        f.write('\n'.join(v))
+                    qmgr_cmd = qmgr + " < " + path
+                    d = self.du.run_cmd(
+                        self.hostname, qmgr_cmd, sudo=True, as_script=True)
+                    os.unlink(path)
+                # Load pbs.conf file
+                elif k == "pbs_conf":
```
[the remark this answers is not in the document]
RKORavi: Changed

test/fw/ptl/lib/pbs_testlib.py
RKORavi: changed

test/fw/ptl/lib/pbs_testlib.py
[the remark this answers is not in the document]
RKORavi: @sujatapatnaik52  i have made the changes, using import_hook api now

test/fw/ptl/lib/pbs_testlib.py
RKORavi: @sujatapatnaik52 @agrawalravi90 I have made the changes

test/fw/ptl/lib/pbs_testlib.py
RKORavi: Changed

test/fw/ptl/lib/pbs_testlib.py
RKORavi: Removed variable events

RKORavi:
> I have an over-arching question: do we really need this capability in PTL? What is this used for?

Please refer to EDD : https://pbspro.atlassian.net/wiki/spaces/PD/pages/557088789/Design+for+a+supported+way+to+change+default+setup+in+PTL?showComments=true&showCommentArea=true

test/fw/ptl/lib/pbs_testlib.py
RKORavi: Right now scope of this PR is to save only server attributes & files.
e.g, print server, print sched, hook files, pbs.conf file, pbsnodes -av.

test/fw/ptl/lib/pbs_testlib.py
RKORavi: Thanks for pointing it out, I forgot to add code to load pbsnodes

test/fw/ptl/lib/pbs_testlib.py
RKORavi: I have removed resourcedef from both save & load config, because i am already using print server command to save resources & then loading them back using qmgr.

RKORavi:
[ptl_run_logs.txt](https://github.com/PBSPro/pbspro/files/3210198/ptl_run_logs.txt)
[save_config.docx](https://github.com/PBSPro/pbspro/files/3210201/save_config.docx)

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4247,45 +4250,97 @@ def _load_configuration(self, infile, objtype=None):
...
-                # load all object types that could be in infile
-                newconf = {}
-                for ky in [MGR_OBJ_SERVER, MGR_OBJ_SCHED, MGR_OBJ_NODE]:
-                    if ky not in conf:
-                        conf[ky] = {}
-                    newconf = dict(newconf.items() + conf[ky].items())
-                conf = newconf
-
+            with open(infile, 'r') as f:
+                conf = json.load(f)
```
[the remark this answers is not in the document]
sujatapatnaik52: I see you are yet to incorporate the above comment.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4247,45 +4250,69 @@ def _load_configuration(self, infile, objtype=None):
...
+                            continue
+                        cf_hpath = ""
+                        if ".PY" in hpath:
+                            cf_hpath = hpath.replace('.PY', '.CF', 1)
+                            hk_hpath = hpath.replace('.PY', '.HK', 1)
+                            rep_str = "hook_name=" + hook_name
+                            cmd = qmgr + " -c 'create hook %s'" % hook_name
+                            self.du.run_cmd(
+                                self.hostname, cmd, sudo=True,
+                                as_script=True)
```
sujatapatnaik52: This is just a question. why do you need as_script=True? Did you try making it work without that?

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -5222,42 +5249,67 @@ def save_configuration(self, outfile, mode='a'):
...
-        rd = os.path.join(self.pbs_conf['PBS_HOME'], 'server_priv',
-                          'resourcedef')
-        self._save_config_file(conf, rd)
-
+        # save pbs.conf file
+        cfg_path = self.du.get_pbs_conf_file()
+        with open(cfg_path, 'r') as p:
+            pbs_cfg = p.readlines()
+            config = self.utils.convert_to_dictlist(pbs_cfg)
+            print("PBS config\n ---------\n")
```
sujatapatnaik52: If the print statements are for debugging, you can remove them.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -5222,42 +5249,67 @@ def save_configuration(self, outfile, mode='a'):
...
+        ret = self.du.run_cmd(self.hostname, [pbsnodes, '-av'])
+        nodes_val = self.utils.convert_to_dictlist(ret['out'])
+        conf['pbsnodes'] = nodes_val
         try:
-            f = open(outfile, mode)
-            cPickle.dump(sconf, f)
-            f.close()
-        except:
+            with open(outfile, mode) as f:
+                json.dump(conf, f)
```
sujatapatnaik52: same try-except suggestion for the above block as well.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4247,45 +4250,83 @@ def _load_configuration(self, infile, objtype=None):
...
-                        fd.write("\n")
-                        self.du.run_cmd(self.hostname, ['cp', fn, k],
-                                        sudo=True)
-                    os.remove(fn)
-
+                # Load server configuration
+                if k.startswith('qmgr_'):
+                    fd, path = tempfile.mkstemp(prefix="runcmd_")
+                    os.chmod(path, 0o700)
+                    with open(path, 'a') as f:
```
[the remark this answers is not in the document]
agrawalravi90: please ignore this comment and only consider my comment below for using create_temp_file()

test/fw/bin/pbs_benchpress
```diff
@@ -1,7 +1,7 @@
-#!/usr/bin/env python
+#!/bin/python
 # coding: utf-8
 
-# Copyright (C) 1994-2019 Altair Engineering, Inc.
+# Copyright (C) 1994-2018 Altair Engineering, Inc.
```
sujatapatnaik52: I did not get this change. This is supposed to be 2019?

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -466,6 +468,10 @@ def setUpClass(cls):
         cls.parse_param()
         cls.init_param()
         cls.check_users_exist()
+        fd, path = tempfile.mkstemp(prefix="new_server_output_",
+                                    suffix=".json")
+        cls.saved_file = path
+        Server().save_configuration(path, 'w')
```
sujatapatnaik52: I have few questions here - 
1. The file name `setupclass_output_`, the last underscore is not needed. If I am missing something here please point me the same. 
2. Is it mandatory to do `save_configuration` before `init_servers`? I was thinking `init_server` is creating the server object and thus you can use the `save_configuration` method of the same instead of creating another `Server()` object.

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -1526,3 +1552,4 @@ def tearDown(self):
     @classmethod
     def tearDownClass(cls):
         cls._testMethodName = 'tearDownClass'
+        Server().load_configuration(cls.saved_file)
```
sujatapatnaik52: I think you can try using `cls.server` here?

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -483,6 +483,10 @@ def setUp(self):
             return
         self.log_enter_setup()
         self.init_proc_mon()
+        fd, path = tempfile.mkstemp(prefix="setup_output_",
+                                    suffix=".json")
+        self.saved_file = path
+        Server().save_configuration(path, 'w')
```
sujatapatnaik52: you can use self.servers here?

prakashcv13:
@RKORavi - if you are not working on this PR - could you please mark it [WIP] or close it.

RKORavi:
@agrawalravi90 @kjakkali @hirenvadalia @saritakh @sujatapatnaik52 
I have made changes to revert_to_defaults() method could you please review the PR

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -9600,6 +9713,51 @@ def import_hook(self, name, body, level=logging.INFO):
...
+        :param body: The body of the hook config as a string.
+        :type body: str
+        :returns: True on success.
+        :raises: PbsManagerError
+        """
+        # sync_mom_hookfiles_timeout is 15min by default
+        # Setting it to lower value to avoid the race condition at hook copy
+        srv_stat = self.status(SERVER, 'sync_mom_hookfiles_timeout')
+        try:
+            sync_val = srv_stat[0]['sync_mom_hookfiles_timeout']
```
agrawalravi90: where do you use sync_val?

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -9600,6 +9713,51 @@ def import_hook(self, name, body, level=logging.INFO):
...
+        :param name: The name of the hook to import body to
+        :type name: str
+        :param body: The body of the hook config as a string.
+        :type body: str
+        :returns: True on success.
+        :raises: PbsManagerError
+        """
+        # sync_mom_hookfiles_timeout is 15min by default
+        # Setting it to lower value to avoid the race condition at hook copy
+        srv_stat = self.status(SERVER, 'sync_mom_hookfiles_timeout')
```
agrawalravi90: where do you use srv_stat?

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -9600,6 +9713,51 @@ def import_hook(self, name, body, level=logging.INFO):
...
+            self.du.run_copy(self.hostname, fn, rfile)
+        else:
+            rfile = fn
+
+        a = {'content-type': 'application/x-config',
+             'content-encoding': 'default',
+             'input-file': rfile}
+        self.manager(MGR_CMD_IMPORT, HOOK, a, name)
+
+        os.remove(rfile)
```
agrawalravi90: you can just use self.du.rm(self.hostname, rfile) for local as well, then you won't need the if condition below

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -494,7 +495,11 @@ def setUpClass(cls):
         cls.parse_param()
         cls.init_param()
         cls.check_users_exist()
+        _, path = tempfile.mkstemp(prefix="setupclass_output",
+                                   suffix=".json")
+        cls.saved_file = path
         cls.init_servers()
+        cls.server.save_configuration(path, 'w')
```
agrawalravi90: trying to understand the logic, shouldn't you be saving the configuration only when '--use-current-setup' was passed by the user?

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -505,11 +510,16 @@ def setUp(self):
...
         self.log_enter_setup()
         self.init_proc_mon()
-        self.revert_servers()
-        self.revert_moms()
-        self.revert_comms()
-        self.revert_schedulers()
-        self.revert_pbsconf()
+        _, path = tempfile.mkstemp(prefix="setup_output",
+                                   suffix=".json")
+        self.saved_file = path
```
agrawalravi90: won't this overwrite the 'saved_file' path that you created in setupclass() ?

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -1564,9 +1574,16 @@ def tearDown(self):
 
         for sched in self.scheds:
             self.scheds[sched].cleanup_files()
+        self.revert_server(self.server)
```
agrawalravi90: why did you add the revert operation to tearDown() ?

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -1564,9 +1574,16 @@ def tearDown(self):
...
+        self.revert_pbsconf()
+        self.server.load_configuration(self.saved_file)
+        self.du.rm(path=self.saved_file)
 
         self.log_end_teardown()
 
     @classmethod
     def tearDownClass(cls):
         cls._testMethodName = 'tearDownClass'
+        cls.server.load_configuration(cls.saved_file)
```
agrawalravi90: why are you loading configuration in tearDownClass()?

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -509,6 +510,7 @@ def set_data(self, paramfile, testparam,
             else:
                 testparam = _f
         self.param = testparam
+        self.cur_setup = use_cur_setup
```
agrawalravi90: I like use_cur_setup more than cur_setup, I'd suggest renaming cur_setup to use_cur_setup, it's more intuitive

test/fw/ptl/utils/pbs_testsuite.py
RKORavi: Sorry my bad, added this to debug and forgot to remove.

test/fw/ptl/utils/pbs_testsuite.py
RKORavi: @agrawalravi90 as per the EDD: https://pbspro.atlassian.net/wiki/spaces/PD/pages/557088789/Design+for+a+supported+way+to+change+default+setup+in+PTL
It says to load PBS everytime in teardown and teardown class. 
Goal is to revert pbs to the same state after running the tests. Even when --use-current-setup is not used

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4247,45 +4250,83 @@ def _load_configuration(self, infile, objtype=None):
...
+                elif k == "pbs_conf":
+                    cfg_vals = base64.decodestring(v)
+                    conf_path = os.path.join(os.sep, 'etc', 'pbs.conf')
+                    # Clear current pbs.conf file and
+                    # rewite with saved config contents
+                    with open(conf_path, 'w') as f:
+                        f.truncate(0)
+                        f.write(cfg_vals)
+                # Load hooks
+                elif k == "hooks":
```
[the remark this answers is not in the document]
RKORavi: Missed it, I will make the changes

RKORavi:
@sujatapatnaik52 @agrawalravi90 should i go ahead and squash my commits, it is getting really difficult when i try to rebase and resolve the conflicts.

test/fw/ptl/utils/pbs_testsuite.py
RKORavi: prefix of both the files are different so, i will not overwrite file created in setupclass

agrawalravi90:
> @sujatapatnaik52 @agrawalravi90 should i go ahead and squash my commits, it is getting really difficult when i try to rebase and resolve the conflicts.

I'd have preferred if you hadn't rebased (and resolved conflicts) until the PR was signed off. But since you've already done this, I don't mind i you squash your commits since I'll have to review it completely anyways. But please don't rebase until you get sign-offs for future PRs.

RKORavi:
Attached the logs of for --use-current-setup option in pbs_benchpress.
1. Logs of running pbs_benchpress on a OOB configuration with --use-current-setup
2. Logs of running pbs_benchpress on a custom setup with --use-current-setup option

test/fw/ptl/utils/pbs_testsuite.py
RKORavi: As per EDD system should revert to original setup after running PTL even when --use-current-setup is not provided in benchpress command

RKORavi:
@agrawalravi90 @sujatapatnaik52 Sorry for delay in addressing all the comments, took me a while to get a travis as passed, though i am still seeing error in centos:7 Build:santize node. I will try to fix that too.

test/fw/ptl/utils/plugins/ptl_test_runner.py
RKORavi: Changed

test/fw/ptl/lib/pbs_testlib.py
RKORavi: Function is no longer required since i have moved to qmgr command for getting hooks output and importing it

test/fw/ptl/lib/pbs_testlib.py
RKORavi: Function is no longer required since i have moved to qmgr command for getting hooks output and importing it

test/fw/ptl/lib/pbs_testlib.py
RKORavi: Function is no longer required since i have moved to qmgr command for getting hooks output and importing it

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4323,49 +4327,117 @@ def _load_configuration(self, infile, objtype=None):
                         for, one of server, scheduler, mom or
                         if None, load all objects in infile
         """
+        #self.unset_svr_attrib()
+        self.delete_hooks()
```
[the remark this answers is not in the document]
RKORavi: changed

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4323,49 +4327,117 @@ def _load_configuration(self, infile, objtype=None):
                         for, one of server, scheduler, mom or
                         if None, load all objects in infile
         """
+        #self.unset_svr_attrib()
```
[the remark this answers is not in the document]
RKORavi: removed commented line

agrawalravi90:
@RKORavi please don't do force push(es) to your branch during the review phase, and do NOT do any rebase(es) until your PR is approved.

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -1474,9 +1487,15 @@ def tearDown(self):
 
         for sched in self.scheds:
             self.scheds[sched].cleanup_files()
+        self.server.del_hooks_nodes_qs()
```
agrawalravi90: Interesting, this will add some part of "revert" action to tearDown() as well. We will delete hooks, nodes, queues and reservations in this function and then load the "default" configuration, where the default could either be PTL's default, or the user's provided configuration.
Ideally, we should do the following:
- split the "revert" functions into two parts: delete and load, where delete would be common but the load function will either load the PTL's default, or the user provided configuration.

But, since this PR has been hanging around for a long time, if you don't want to do the refactoring, then I suggest that you only call "self.server.del_hooks_nodes_qs()" and "self.server.load_configuration(self.saved_file)" if "self.use_cur_setup" is True. Similarly, only call "self.server.save_configuration(path, 'w')" in setUp() if "self.use_cur_setup" is True. This will mean that when "self.use_cur_setup" is not True, PTL will behave exactly like before.

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -1474,9 +1487,15 @@ def tearDown(self):
...
+        self.server.load_configuration(self.saved_file)
+        self.du.rm(path=self.saved_file)
```
agrawalravi90: Do you really need to remove the saved configuration? Why save and delete for every test? You only need to save it once in setUp() and then just load it in tearDown() right?

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4187,6 +4190,27 @@ def tracejob_match(self, msg, id=None, n=50, tail=True,
...
+                           del_nodes=True,
+                           del_qs=True,
+                           del_vdef=True):
+        """
+        Delete nodes, queues and hooks
+        """
+        if del_hooks:
+            self.delete_hooks()
+        if del_vdef or del_nodes:
+            self.cleanup_reservations()
```
agrawalravi90: since you also delete reservations here, please add that to the method name and docstring

agrawalravi90: or you can call the method something generic, like "delete_current_state", up to you. But do mention what all will get deleted in the docstring

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4187,6 +4190,27 @@ def tracejob_match(self, msg, id=None, n=50, tail=True,
                                regexp, max_attempts, interval, starttime,
                                endtime, level, existence)
 
+    def del_hooks_nodes_qs(self, del_hooks=True,
+                           del_nodes=True,
+                           del_qs=True,
+                           del_vdef=True):
+        """
+        Delete nodes, queues and hooks
+        """
```
agrawalravi90: please document the function's arguments in the doctring

test/fw/ptl/utils/pbs_testsuite.py
RKORavi: Hi Ravi,

As per the EDD : https://pbspro.atlassian.net/wiki/spaces/PD/pages/557088789/Design+for+a+supported+way+to+change+default+setup+in+PTL
- We have to bring back PBS to the same custom setup on which it started.
- Saved configuration in setup() has to be loaded in teardown() everytime.
- When --use-current-setup option is used, tests should run on given custom setup.

Instead of saving the configurations for each test case. I could save configuration the first time when setup() is called for the first time in a test suite

agrawalravi90: right, that's exactly what i was suggesting. Save it only once, don't save again if already saved, then in tearDown() you don't need to delete self.saved_file

test/fw/ptl/lib/pbs_testlib.py
RKORavi: Added function arguments as docstring

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4247,45 +4250,83 @@ def _load_configuration(self, infile, objtype=None):
...
-                    if ky not in conf:
-                        conf[ky] = {}
-                    newconf = dict(newconf.items() + conf[ky].items())
-                conf = newconf
-
+            with open(infile, 'r') as f:
+                conf = json.load(f)
+            qmgr = os.path.join(self.client_conf['PBS_EXEC'],
+                                'bin', 'qmgr')
             for k, v in conf.items():
```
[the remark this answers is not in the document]
RKORavi: iteritems() depricated in python3

test/fw/ptl/lib/pbs_testlib.py
RKORavi: Added docstring & changed function name

test/fw/ptl/utils/pbs_testsuite.py
RKORavi: updated changes by adding class variable: PBSTestSuite.config_saved as mentioned in design doc.

RKORavi:
Travis is taking 9 mins more after this change in PTL framework

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4208,46 +4241,77 @@ def _load_configuration(self, infile, objtype=None):
...
-                                        sudo=True)
-                    os.remove(fn)
-
+                # Load server configuration
+                if k.startswith('qmgr_'):
+                    fpath = self.du.create_temp_file()
+                    print_svr = '\n'.join(v)
+                    with open(fpath, 'w') as f:
+                        f.write(print_svr)
+                    file_qmgr = open(fpath)
```
sujatapatnaik52: I am not sure why are you opening the file in the above line again? Can't this be achieved with the above `with open as f`?

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4208,46 +4241,77 @@ def _load_configuration(self, infile, objtype=None):
...
-                        # append the last line
-                        fd.write("\n")
-                        self.du.run_cmd(self.hostname, ['cp', fn, k],
-                                        sudo=True)
-                    os.remove(fn)
-
+                # Load server configuration
+                if k.startswith('qmgr_'):
+                    fpath = self.du.create_temp_file()
+                    print_svr = '\n'.join(v)
```
sujatapatnaik52: Are we just loading server here or all objects? Then I think we should update the docstring accordingly.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -5158,36 +5239,121 @@ def save_configuration(self, outfile, mode='a'):
...
-        ret = self.du.run_cmd(self.hostname, [qmgr, '-c', 'p h'], sudo=True)
+        return True
+
+    def _save_hook_files(self):
+        """
+        save all the hooks .CF, .PY, .HK files
+        """
+        qmgr = os.path.join(self.client_conf['PBS_EXEC'], 'bin', 'qmgr')
+        cfg = {"hooks": ""}
+        cmd = [qmgr, '-c', 'print hook @default']
```
sujatapatnaik52: Just have a doubt here. @default is for specifying server right? If we do `print hook` then that should also return the same output. Why are we using @default explicitly?

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -5158,36 +5239,121 @@ def save_configuration(self, outfile, mode='a'):
...
+        sconf = {MGR_OBJ_SERVER: conf}
+        # save pbs.conf file
+        cfg_path = self.du.get_pbs_conf_file()
+        with open(cfg_path, 'r') as p:
+            pbs_cfg = p.readlines()
+            config = self.utils.convert_to_dictlist(pbs_cfg)
+            pbs_cfg_b64 = base64.encodestring(str(config[0]))
+        conf['pbs_conf'] = pbs_cfg_b64
+        # save hook files
+        hooks_str = self._save_hook_files()
```
sujatapatnaik52: I think you should add a check here - if `_save_hook_files()` is returning `False` then you should maybe throw an exception or so?

test/fw/ptl/utils/pbs_testsuite.py
sujatapatnaik52: Since saving configuration is kind of mandatory now, I think in the `save_configuration()`, you are returning `False` so maybe you should add a check here as well to proceed only if it succeeds.

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -441,11 +446,22 @@ def setUp(self):
             return
         self.log_enter_setup()
         self.init_proc_mon()
-        self.revert_servers()
+        if not PBSTestSuite.config_saved:
+            _, path = tempfile.mkstemp(prefix="setup_output",
+                                       suffix=".json")
+            self.saved_file = path
+            self.server.save_configuration(path, 'w')
```
sujatapatnaik52: Same comment as above.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4208,46 +4241,77 @@ def _load_configuration(self, infile, objtype=None):
...
+            nodes = conf['pbsnodes']
+            for node in nodes:
+                node_name = str(node['id'])
+                self.create_pbsnode(node_name, node)
             return True
         return False
 
+    def create_pbsnode(self, node_name, attrs):
+        """
+        Create node in PBS with given attributes
```
sujatapatnaik52: I have a doubt here, Suppose the server has a pre-config of one node and some vnodes of that execution host. Will this API `create_pbsnode` take care of that?

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -5127,12 +5127,93 @@ def revert_to_defaults(self, reverthooks=True, revertqueues=True,
...
+                nodes = self.status(VNODE, logerr=False)
+                for node in nodes:
+                    if 'queue' in node.keys():
+                        self.manager(MGR_CMD_UNSET, NODE, 'queue',
+                                     node['id'])
+            except BaseException:
+                pass
+            self.manager(MGR_CMD_DELETE, QUEUE, id=queues)
+
+    def delete_sched(self):
```
sujatapatnaik52: Do you think will it be more appropriate if we say `delete_sched_config` or maybe something of that sort?

RKORavi:
[ptl_run_with_use_current_setup.txt](https://github.com/PBSPro/pbspro/files/3619401/ptl_run_with_use_current_setup.txt)
[printing_configurations.txt](https://github.com/PBSPro/pbspro/files/3619403/printing_configurations.txt)

Test logs with latest changes

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -456,6 +457,8 @@ def setUp(self):
         # setup block, rest of them to be added to this block
         # once save & load configurations are implemented for
         # comm, mom, scheduler
+        if self.use_cur_setup:
+            self.server.delete_nodes()
         if not self.use_cur_setup:
```
agrawalravi90: make this an else:

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -456,6 +457,8 @@ def setUp(self):
         # setup block, rest of them to be added to this block
         # once save & load configurations are implemented for
         # comm, mom, scheduler
+        if self.use_cur_setup:
```
agrawalravi90: I think you should move the

        if self.use_cur_setup:
            self.server.delete_current_state()
            self.server.load_configuration(self.saved_file)

block here from tearDown() instead of adding this.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4208,46 +4238,77 @@ def _load_configuration(self, infile, objtype=None):
...
-                    if ky not in conf:
-                        conf[ky] = {}
-                    newconf = dict(newconf.items() + conf[ky].items())
-                conf = newconf
-
+                try:
+                    conf = json.load(f)
+                except ValueError:
+                    self.logger.info("Error loading JSON file: %s"
+                                     % infile)
```
hirenvadalia: shouldn't we return from here if we fail to load data from file?

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4208,46 +4238,77 @@ def _load_configuration(self, infile, objtype=None):
...
-                    os.remove(fn)
-
+                # Load server configuration
+                if k.startswith('qmgr_'):
+                    fpath = self.du.create_temp_file()
+                    print_svr = '\n'.join(v)
+                    with open(fpath, 'w') as f:
+                        f.write(print_svr)
+                    file_qmgr = open(fpath)
+                    cmd = [qmgr]
```
hirenvadalia: you don't need cmd here you can pass [qmgr] directly to run_cmd

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4208,46 +4238,77 @@ def _load_configuration(self, infile, objtype=None):
...
+                    d = self.du.run_cmd(
```
hirenvadalia: please add error check here

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4208,46 +4238,77 @@ def _load_configuration(self, infile, objtype=None):
...
+        """
+        Create node in PBS with given attributes
+        """
+        qmgr = os.path.join(self.client_conf['PBS_EXEC'],
+                            'bin', 'qmgr')
+        execcmd = "create node " + node_name
+        cmd = [qmgr, "-c", execcmd]
+        self.du.run_cmd(
+            self.hostname, cmd, sudo=True)
+        skip_atb_list = ['id', 'Port', 'pbs_version', 'pcpus',
```
hirenvadalia: we can't ignore port and state, as in case of multi mom on a single host, the port will be different.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4208,46 +4238,77 @@ def _load_configuration(self, infile, objtype=None):
...
+                         ]
+        for node_atb, val in attrs.items():
+            if(node_atb in skip_atb_list or
+               'resources_assigned' in node_atb):
+                continue
+            k = str(node_atb)
+            v = str(val)
+            execcmd = "set node %s %s=%s" % (node_name, k, v)
+            cmd = [qmgr, "-c", execcmd]
+            self.du.run_cmd(
```
hirenvadalia: please add error check here

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -5032,59 +5062,26 @@ def revert_to_defaults(self, reverthooks=True, revertqueues=True,
...
 
         if delnodes:
-            try:
-                self.manager(MGR_CMD_DELETE, VNODE, id="@default")
-            except PbsManagerError as e:
-                if "Unknown node" not in e.msg[0]:
-                    raise
+            self.delete_nodes()
         if reverthooks:
             if self.platform == 'cray' or self.platform == 'craysim':
```
hirenvadalia: shouldn't this be part of delete_hooks?

test/fw/ptl/lib/pbs_testlib.py
RKORavi: Hi Hiren,

I am running the command with logerr=False, which will not print anything even if there is a failure. Do you want me log error when exit status is non-zero here?

test/fw/ptl/lib/pbs_testlib.py
RKORavi: For me it looks like setter part in revert to defaults here. I have separated delete configurations from this method

test/fw/ptl/lib/pbs_testlib.py
hirenvadalia: no I mean what will happen if the command fails? shouldn't we stop or do some error handling

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4995,31 +5049,7 @@ def revert_to_defaults(self, reverthooks=True, revertqueues=True,
                                           'lib', 'python', 'altair',
                                           'pbs_hooks',
                                           'PBS_translate_mpp.HK')
-        self.jacs_hk = os.path.join(self.pbs_conf['PBS_HOME'],
```
hirenvadalia: you need to restore this as it is used under shasta

test/fw/ptl/lib/pbs_testlib.py
hirenvadalia: but test calls delete_hooks then it will delete these hooks which it shouldn't right... if we don't put this in delete_hooks then test writer has to take care of restoring this hook...

RKORavi: Thanks I got it now.

test/fw/ptl/lib/pbs_testlib.py
RKORavi: As per admin guide: @default refers to default server.

test/fw/ptl/lib/pbs_testlib.py
RKORavi: In with block i am writing contents to a file. Then I am opening the file again and passing file descriptor it to run_cmd as stdin. If I add the same thing within with block then i does not work as expected. It is unable to load the configurations.

test/fw/ptl/lib/pbs_testlib.py
RKORavi: I forgot to check for this condition, but here vnodes will not get created back. The reason is that I am deleting vnodedef file in teardown. I can add the save of vnodedef file in save configurations and load it back in teardown. But i feel that vnodedef file should be under save & load of Mom, since the scope of this PR is limited only to server.
What do you think @sujatapatnaik52 @hirenvadalia @agrawalravi90

hirenvadalia: I am ok with handling vnodedefs with --use-current-setup in separate PR for MoM object

sujatapatnaik52: I do agree with you. However, we should mention this in the EDD probably, if there are vnodes associated to the nodes then this option is not going load them back.

agrawalravi90: I'm ok doing this in a different PR as well. I don't think the EDD needs to be modified, an EDD can be implemented via multiple PRs. Maybe instead make a note of it in the PR description?

test/fw/ptl/lib/pbs_testlib.py
RKORavi: Changed, returning false when saving of hook has failed.

test/fw/ptl/utils/pbs_testsuite.py
RKORavi: changed

test/fw/ptl/lib/pbs_testlib.py
RKORavi: changed

test/fw/ptl/utils/pbs_testsuite.py
RKORavi: changed

test/fw/ptl/utils/pbs_testsuite.py
RKORavi: I tried to move this block from teardown to setup() but i was still facing the same problem.
So keeping this change

test/fw/bin/pbs_benchpress
```diff
@@ -339,6 +341,8 @@ if __name__ == '__main__':
         elif o == '--version':
             print ptl.__version__
             sys.exit(0)
+        elif o == '--use-current-setup':
```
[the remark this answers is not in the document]
RKORavi: changed

test/fw/bin/pbs_benchpress
```diff
@@ -454,7 +458,8 @@ if __name__ == '__main__':
         data = PTLTestData()
         loader.set_data(testgroup, testsuites, excludes, follow, testfiles)
         testtags.set_data(tags, eval_tags)
-        runner.set_data(paramfile, testparam, lcov_bin, lcov_data, lcov_out,
+        runner.set_data(paramfile, testparam, use_cur_setup, lcov_bin,
```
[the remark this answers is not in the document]
RKORavi: changed

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4187,6 +4190,33 @@ def tracejob_match(self, msg, id=None, n=50, tail=True,
...
+        """
+        if del_hooks:
+            self.delete_hooks()
+        if del_vdef or del_nodes:
+            self.cleanup_reservations()
+        if del_vdef:
+            for m in self.moms:
+                self.moms[m].delete_vnode_defs()
+                self.moms[m].delete_vnodes()
+                self.moms[m].restart()
```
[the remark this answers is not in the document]
RKORavi: changed

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4187,6 +4190,33 @@ def tracejob_match(self, msg, id=None, n=50, tail=True,
...
+        :param del_vdef: If True, delete vnodedef file
+        :type del_vdef: boolean
+        """
+        if del_hooks:
+            self.delete_hooks()
+        if del_vdef or del_nodes:
+            self.cleanup_reservations()
+        if del_vdef:
+            for m in self.moms:
+                self.moms[m].delete_vnode_defs()
```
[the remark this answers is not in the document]
RKORavi: added the check

test/fw/ptl/lib/pbs_testlib.py
RKORavi: changed

test/fw/ptl/lib/pbs_testlib.py
RKORavi: Added

test/fw/ptl/lib/pbs_testlib.py
RKORavi: Changed

test/fw/ptl/lib/pbs_testlib.py
RKORavi: Added a check

RKORavi:
@agrawalravi90 @hirenvadalia @sujatapatnaik52 should i go head and rebase my branch. I would like to test my PR with recent changes of python3

hirenvadalia:
@RKORavi Yes, please go ahead with rebase, it will solve a conflict also in testlib.

agrawalravi90:
@RKORavi No, please give me a chance to review your new changes, you can rebase after you get sign-offs

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4246,10 +4227,12 @@ def _load_configuration(self, infile, objtype=None):
...
                         f.write(print_svr)
                     file_qmgr = open(fpath)
-                    cmd = [qmgr]
                     d = self.du.run_cmd(
-                        self.hostname, cmd, stdin=file_qmgr, sudo=True,
+                        self.hostname, [qmgr], stdin=file_qmgr, sudo=True,
                         logerr=False, level=logging.DEBUG)
+                    err_msg = "Failed to load server configurations"
+                    if d['rc'] != 0:
+                        self.logger.error("%s" % err_msg)
```
agrawalravi90: so, if this (or any of the other object load) fails, what will happen? will pbs_benchpress continue running? Should we stop and error out instead?

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4263,15 +4246,20 @@ def _load_configuration(self, infile, objtype=None):
                     with open(fpath, 'w') as f:
                         f.write(print_hooks)
                     file_qmgr = open(fpath)
-                    cmd = [qmgr]
                     d = self.du.run_cmd(
-                        self.hostname, cmd, stdin=file_qmgr, sudo=True,
-                        logerr=False, level=logging.DEBUG)
+                        self.hostname, [qmgr], stdin=file_qmgr, sudo=True,
+                        logerr=True, level=logging.DEBUG)
```
agrawalravi90: logerr is True by default, did you want to set it to False instead?

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4283,24 +4271,34 @@ def create_pbsnode(self, node_name, attrs):
...
         execcmd = "create node " + node_name
         cmd = [qmgr, "-c", execcmd]
-        self.du.run_cmd(
-            self.hostname, cmd, sudo=True)
-        skip_atb_list = ['id', 'Port', 'pbs_version', 'pcpus',
-                         'state', 'last_state_change_time', 'ntype',
+        ret = self.du.run_cmd(self.hostname, cmd, sudo=True)
+        if ret['rc'] != 0:
+            self.logger.info("Failed to create node: %s" % node_name)
+            self.logger.error("Error: %s" % ret['err'])
```
agrawalravi90: Since you didn't provide logerr=False to run_cmd, it will print out ret['err'] automatically, so please remove this line.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4283,24 +4271,34 @@ def create_pbsnode(self, node_name, attrs):
...
             execcmd = "set node %s %s=%s" % (node_name, k, v)
             cmd = [qmgr, "-c", execcmd]
-            self.du.run_cmd(
-                self.hostname, cmd, sudo=True,
-                logerr=False, level=logging.DEBUG)
+            ret = self.du.run_cmd(self.hostname, cmd, sudo=True,
+                                  logerr=False, level=logging.DEBUG)
+            if ret['rc'] != 0:
+                self.logger.info("Failed to set node attribute %s=%s" % (k, v))
+                self.logger.error("Error: %s" % ret['err'])
```
agrawalravi90: shouldn't you return False here?

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -5248,19 +5250,28 @@ def save_configuration(self, outfile, mode='a'):
...
             conf['qmgr_print_sched'] = ret['out']
         ret = self.du.run_cmd(self.hostname, [pbsnodes, '-av'],
                               logerr=False, level=logging.DEBUG)
-        nodes_val = self.utils.convert_to_dictlist(ret['out'])
-        conf['pbsnodes'] = nodes_val
+        err_msg = "Server has no node list"
+        # pbsnodes -av returns a non zero exit code when there are
+        # no nodes in cluster
+        if ret['rc'] != 0 and err_msg in ret['err']:
+            self.logger.error('Failed to nodes info')
```
agrawalravi90: type: "failed to get nodes info" instead of "Failed to nodes info"

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -445,17 +450,24 @@ def setUp(self):
...
-        self.server.save_configuration(path, 'w')
+        if not PBSTestSuite.config_saved and self.use_cur_setup:
+            _, path = tempfile.mkstemp(prefix="saved_test_setup",
+                                       suffix=".json")
+            self.saved_test_setup = self.server.save_configuration(path, 'w')
+            if self.saved_test_setup:
+                self.saved_file = path
+            else:
+                self.logger.error("Failed to save test setup")
+                PBSTestSuite.config_saved = True
```
agrawalravi90: I think we should stop executing and error out if the save fails, what do you think?

test/fw/ptl/lib/pbs_testlib.py
RKORavi: I am deleting only site hooks. So, i did not need to add these in delete_hooks() method.
I have changed function name to delete_site_hooks()

test/fw/ptl/lib/pbs_testlib.py
RKORavi: Right now, it will continue and will not save the configuration for the failed command.
So, at the time of loading we will just not load any failed commands.

agrawalravi90: shouldn't we error out if that happens?

test/fw/ptl/lib/pbs_testlib.py
RKORavi: Set node fails even if i try to set Port on mom, but i have to load Port on node to care of condition: multi mom on single host. That is the reason I am not returning False here

test/fw/ptl/lib/pbs_testlib.py
hirenvadalia: I agree with @agrawalravi90 we should error out if any command fails instead continue because it won't save correct conf nither it will restore correct.

test/fw/ptl/lib/pbs_testlib.py
RKORavi: @hiren 
1. I created multiple moms on a single host and on different port using pbs_config,
pbs_conf --multi-mom --create=4 --init-port=60000
2. Deleted one of the moms & tried to create it back using qmgr command
3. I was getting an error in create node
qmgr -c "c n spartan-6 Port=60006"
qmgr obj=spartan-6 svr=default: Unknown node.

Whereas, when i run the same command for a node on different host it gets created successfully.
qmgr -c "c n 10.75.37.51 Port=60006".
Looks like it will fail in create node and will not go ahead with set node for multi mom on single host.

hirenvadalia: @RKORavi You tagged wrong hiren :) my username is @hirenvadalia

Can you provide me pbs_config output with -lDEBUG when u created multiple moms.

test/fw/ptl/utils/pbs_testsuite.py
RKORavi: Stopping the execution when a command fails.

test/fw/ptl/lib/pbs_testlib.py
RKORavi: Changed

test/fw/ptl/lib/pbs_testlib.py
RKORavi: Changed

test/fw/ptl/lib/pbs_testlib.py
RKORavi: changed

test/fw/ptl/lib/pbs_testlib.py
RKORavi: changed

RKORavi:
@hirenvadalia @agrawalravi90 @sujatapatnaik52 please review the latest change

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4208,6 +4208,7 @@ def _load_configuration(self, infile, objtype=None):
                         for, one of server, scheduler, mom or
                         if None, load all objects in infile
         """
+        self.unset_svr_attrib()
```
agrawalravi90: Please move this outside of _load_configuration()

test/fw/ptl/lib/pbs_testlib.py
RKORavi: Sorry my bad, corrected it now :)

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -435,11 +435,12 @@ def setUpClass(cls):
...
                                        suffix=".json")
-            cls.saved_cus_setup = cls.server.save_configuration(path, 'w')
-            if cls.saved_cus_setup:
+            ret = cls.server.save_configuration(path, 'w')
+            if ret:
                 cls.saved_file = path
             else:
-                self.logger.error("Failed to save custom setup")
+                cls.logger.error("Failed to save custom setup")
+                sys.exit(1)
```
agrawalravi90: I think you should do self.fail("Failed to save custom setup") instead of doing sys.exit(1), that way pbs_benchpress can exit gracefully. Same comment elsewhere in this file

hirenvadalia: `self.fail` is only available in test case code, not in this class, you can raise an exception to stop execution flow here

RKORavi:
@hirenvadalia Please find the -lDEBUG logs of pbs_config attached
[debug logs.txt](https://github.com/PBSPro/pbspro/files/3653592/debug.logs.txt)

test/fw/ptl/lib/pbs_testlib.py
hirenvadalia: @RKORavi can you please try this commands:
`pbs_config --multi-mom --create=4 --init-port=60000`
`qmgr -c "d n spartan-6"`
`qmgr -c "c n spartan-6 port=60006,Mom=spartan"`

RKORavi:
Thanks @hirenvadalia @agrawalravi90 for sign off, rebasing my branch to test with python3 changes

RKORavi:
[qmgr server output.txt](https://github.com/PBSPro/pbspro/files/3656947/qmgr.server.output.txt)
[smoketest with option use_current_setup.txt](https://github.com/PBSPro/pbspro/files/3656948/smoketest.with.option.use_current_setup.txt)

RKORavi:
Build report:

Machine | OS | Build_Status | Artifacts link | PTL_REPORT | Total_Tests | Passed | Skipped | Failed | Error | Timedout | Build_Start_Time[PDT] | Test_Time | Total_Build_Time
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
pbi-01-r6 | linux26_x86_64 | Success | Artifacts | Results_Link | 1282 | 1272 | 10 | 0 | 0 | 0 | 05:28:17 | 1:16:31 | 01:43:34
pbi-04-r7 | linux310_x86_64 | Success | Artifacts | Results_Link | 1282 | 1272 | 10 | 0 | 0 | 0 | 08:01:06 | 1:23:40 | 01:53:51
pbi-02-r6-altix | linux26_x86_64 | Success | Artifacts | Results_Link | 1282 | 1272 | 10 | 0 | 0 | 0 | 07:50:49 | 1:22:54 | 01:55:15
pbi-11-s11 | linux26_x86_64 | Success | Artifacts | Results_Link | 1282 | 1272 | 10 | 0 | 0 | 0 | 04:34:19 | 0:56:21 | 01:21:36
pbi-18-s12 | linux312_x86_64 | Success | Artifacts | Results_Link | 1282 | 1272 | 10 | 0 | 0 | 0 | 06:15:56 | 0:53:41 | 01:42:47
pbi-23-r7p1-altix | linux310_x86_64 | Success | Artifacts | Results_Link | 1282 | 1272 | 10 | 0 | 0 | 0 | 05:47:07 | 1:18:47 | 01:51:07
pbi-24-s12-altix | linux312_x86_64 | Success | Artifacts | Results_Link | 1282 | 1272 | 10 | 0 | 0 | 0 | 06:35:47 | 0:57:34 | 01:52:12
pbi-25-s11p1-altix | linux26_x86_64 | Success | Artifacts | Results_Link | 1282 | 1272 | 10 | 0 | 0 | 0 | 04:46:37 | 1:08:59 | 01:38:17
pbi-26-s12-cray | linux312_x86_64_crayxt | Success | Artifacts | Results_Link | 1282 | 1271 | 11 | 0 | 0 | 0 | 06:31:41 | 0:56:22 | 01:49:21
pbi-30-s11-cray | linux26_x86_64_crayxt | Success | Artifacts | Results_Link | 1282 | 1271 | 11 | 0 | 0 | 0 | 04:39:09 | 1:01:27 | 01:28:15
pbi-31-u16 | linux413_x86_64 | Success | Artifacts | Results_Link | 1282 | 1272 | 10 | 0 | 0 | 0 | 07:26:31 | 1:10:04 | 02:05:55
pbi-32-s12 | linux312_x86_64_sanitize | Success | Artifacts | Results_Link | 1276 | 1256 | 10 | 0 | 9 | 1 | 11:46:23 | 1:18:59 | 02:12:29
pbi-33-r7 | linux310_x86_64_sanitize | Success | Artifacts | Results_Link | 1282 | 1272 | 10 | 0 | 0 | 0 | 07:09:15 | 1:26:02 | 01:57:31
pbi-34-s15 | linux412_x86_64 | Success | Artifacts | Results_Link | 1282 | 1272 | 10 | 0 | 0 | 0 | 05:48:14 | 1:27:35 | 02:07:44
pbi-35-u18 | linux415_x86_64 | Success | Artifacts | Results_Link | 1282 | 1272 | 10 | 0 | 0 | 0 | 06:57:51 | 1:10:02 | 02:04:23
pbi-36-s15-altix | linux412_x86_64 | Success | Artifacts | Results_Link | 1282 | 1272 | 10 | 0 | 0 | 0 | 05:17:03 | 1:43:59 | 02:13:46
pbi-37-s15-cray | linux412_x86_64_crayxt | Success | Artifacts | Results_Link | 1282 | 1271 | 11 | 0 | 0 | 0 | 04:05:19 | 1:32:20 | 01:58:08
pbi-a01-s12-cray | linux44_aarch64_crayxt | Success | Artifacts | Results_Link | 1282 | 1271 | 11 | 0 | 0 | 0 | 02:33:39 | 2:06:28 | 03:04:57
pbi-a02-s12 | linux44_aarch64.tar.gz | Fail | NA | NA | NA | NA | NA | NA | NA | NA | NA | NA | NA
pbi-a03-r7 | linux45_aarch64 | Success | Artifacts | Results_Link | 1282 | 1272 | 10 | 0 | 0 | 0 | 02:33:36 | 1:08:53 | 01:46:08
pbi-xarm-01-s15 | linux412_aarch64 | Success | Artifacts | Results_Link | 1282 | 1272 | 10 | 0 | 0 | 0 | 05:50:35 | 1:11:43 | 01:43:47
pbi-xarm-02-s15-cray | linux412_aarch64_crayxt | Success | Artifacts | Results_Link | 1282 | 1271 | 11 | 0 | 0 | 0 | 05:39:37 | 1:11:53 | 01:45:34
pbi-38-2k16 | windows | Success | Artifacts |   |   |   |   |   |   |   | 02:34:05 |   | 00:41:30

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -5139,6 +5132,87 @@ def revert_to_defaults(self, reverthooks=True, revertqueues=True,
...
+            except BaseException:
+                pass
+            self.manager(MGR_CMD_DELETE, QUEUE, id=queues)
+
+    def delete_sched_config(self):
+        """
+        Delete sched_priv & sched_log files
+        """
+        self.manager(MGR_CMD_LIST, SCHED)
+        for name in self.schedulers.keys():
```
npadole20: Please do list(self.schedulers.keys()) here as you are modifying the object at line 5195.

RKORavi: thanks

RKORavi:
squashing reviewed commits, so that it will be easier to rebase the branch

RKORavi:
[regression results.docx](https://github.com/PBSPro/pbspro/files/3660993/regression.results.docx)

RKORavi:
@hirenvadalia @agrawalravi90 please review the changes w.r.t python3 changes, I have attached all the logs.

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -457,7 +457,7 @@ def setUpClass(cls):
                 cls.saved_file = path
             else:
                 cls.logger.error("Failed to save custom setup")
-                sys.exit(1)
+                raise Exception("Failed to load custom setup")
```
hirenvadalia: error message doesn't match with the above line

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -474,9 +474,10 @@ def setUp(self):
             ret = self.server.save_configuration(path, 'w')
             if ret:
                 self.saved_file = path
+                PBSTestSuite.config_saved = True
             else:
                 self.logger.error("Failed to save test setup")
-                sys.exit(1)
+                raise Exception("Failed to load custom setup")
```
hirenvadalia: same as above

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4240,7 +4240,9 @@ def _load_configuration(self, infile, objtype=None):
                         return False
                 # Load pbs.conf file
                 elif k == "pbs_conf":
-                    cfg_vals = base64.decodestring(v)
+                    enc_utf = v.encode('UTF-8')
```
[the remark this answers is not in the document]
RKORavi: I had to make this change to be compatible with python3. I cannot directly encode a string to base64 in python3. I had to either convert it to byte or encode it in first UTF-8 format. Converting in byte and then trying to use the string proved a little problematic for me. That is the reason i went for encoding it with UTF-8 format & then encoding it in base64 format.

Issue #1118, opened 2019-05-09
Job array subjobs are not being held after failing/requeueing 20 times.


Shrini-h:
Currently run_count attribute of an instantiated subjob is not incremented for each retry by server to run it on mom and hence any exception or rejection by Mom to run it will result in to and fro transfer of subjob between server and mom for eternity causing unnecessary mom and server cpu time and log space.

dilip-krishnan:
The question that comes here is, should we just held the failing subjob or the whole array job. Personally
I feel we should maintain the fail count in the parent job and at any point if the cumulatively failure of subjobs reaches to 20 we should held the whole array job instead of putting a hold on the failing subjob. 

This actually reminds me of a forum discussion on how the array job is percieved in PBS. AFAIK we consider array jobs as a part of single normal job, instead of considering each subjob as a normal job.  Here is the link to the discussion http://community.pbspro.org/t/understanding-array-jobs/212/2 .

PR #1119, opened 2019-05-09
Array subjobs are not being held after requeueing 20 times

Shrini-h:
#### Describe Bug or Feature
https://github.com/PBSPro/pbspro/issues/1118
Currently run_count attribute of an instantiated subjob is not incremented for each retry by server to run it on mom and hence any exception or rejection by Mom to run it will result in to and fro transfer of subjob between server and mom for eternity causing unnecessary mom and server cpu time and log space.

Also fixes below invalid write
==29750== Invalid read of size 4
==29750== at 0x44FE93: node_down_requeue (node_manager.c:740)
==29750== by 0x4DA7CA: dispatch_task (work_task.c:142)
==29750== by 0x4DAB26: default_next_task (work_task.c:316)
==29750== by 0x466B61: next_task (pbsd_main.c:2373)
==29750== by 0x466389: main (pbsd_main.c:2033)
==29750== Address 0x148180b8 is 232 bytes inside a block of size 6,440 free'd
==29750== at 0x4C2ACBD: free (vg_replace_malloc.c:530)
==29750== by 0x444B31: job_free (job_func.c:580)
==29750== by 0x44518D: job_purge (job_func.c:975)
==29750== by 0x487C7A: force_reque (req_rerun.c:178)
==29750== by 0x450145: post_discard_job (node_manager.c:850)
==29750== by 0x4530B6: discard_job (node_manager.c:2802)
==29750== by 0x44FE8E: node_down_requeue (node_manager.c:737)
==29750== by 0x4DA7CA: dispatch_task (work_task.c:142)
==29750== by 0x4DAB26: default_next_task (work_task.c:316)
==29750== by 0x466B61: next_task (pbsd_main.c:2373)
==29750== by 0x466389: main (pbsd_main.c:2033)
==29750== Block was alloc'd at
==29750== at 0x4C29BC3: malloc (vg_replace_malloc.c:299)
==29750== by 0x4445EF: job_alloc (job_func.c:321)
==29750== by 0x42D702: create_subjob (array_func.c:855)
==29750== by 0x48A47B: req_runjob (req_runjob.c:511)
==29750== by 0x468654: dispatch_request (process_request.c:773)
==29750== by 0x468112: process_request (process_request.c:488)
==29750== by 0x4DF425: process_socket (net_server.c:504)
==29750== by 0x4DF5A0: wait_request (net_server.c:585)
==29750== by 0x46666B: main (pbsd_main.c:2137)

#### Describe Your Change

- removed `job_purge()` of subjob from all fail/reque paths
- extracted and backedup `run_count` and `run_version` attributes before purging existing instance of subjob job object in `req_runjob()`
- reverted the `run_count` and `run_version` attributes after fresh instantiation of subjob job object in `req_runjob()`
- Now subjobs breaching 20 `run_count` will move to held state `H`. Also there will be a system hold placed on its parent job Array.
- releasing system hold on job array will also release any system holds on any of its sujobs

#### Link to Design Doc

https://pbspro.atlassian.net/wiki/spaces/PD/pages/1049854035/Changes+to+run+count+attribute+hold+and+release+operation+of+subjob+s+in+a+Job+Array

[PTL_test_run_count_subjob.txt](https://github.com/PBSPro/pbspro/files/3173076/PTL_test_run_count_subjob.txt)
[Test_report_logs.zip](https://github.com/PBSPro/pbspro/files/3173077/Test_report_logs.zip)
[val_mem_err_fix_verification.txt](https://github.com/PBSPro/pbspro/files/3176688/val_mem_err_fix_verification.txt)
[valgrind_test_run_count_subjob.log](https://github.com/PBSPro/pbspro/files/3176809/valgrind_test_run_count_subjob.log)

New PTL logs 27 May
[ptl_out_27_may.txt](https://github.com/PBSPro/pbspro/files/3222254/ptl_out_27_may.txt)
[Test_report_logs_27_may.zip](https://github.com/PBSPro/pbspro/files/3222257/Test_report_logs_27_may.zip)

New PTL logs 29 May
[ptl_out_29_may.txt](https://github.com/PBSPro/pbspro/files/3230926/ptl_out_29_may.txt)


Shrini-h:
PR ready for review, Test logs attached !
Tagging @suresh-thelkar  @Bhagat-Rajput  @visheshh  @nithinj  @subhasisb  @sakshamgarg  @bhroam  @arungrover  and others

Shrini-h:
Valgrind report attached, verified fix of invalid read/writes
[val_mem_err_fix_verification.txt](https://github.com/PBSPro/pbspro/files/3176688/val_mem_err_fix_verification.txt)

arungrover:
I’m reviewing this pull. I’ll post my review by tomorrow

src/server/req_holdjob.c
```diff
@@ -300,9 +300,54 @@ req_releasejob(struct batch_request *preq)
...
+				if (psubjob->ji_wattr[(int)JOB_ATR_hold].at_val.at_long == 0)
+					job_attr_def[(int)JOB_ATR_Comment].at_free(&psubjob->ji_wattr[(int)JOB_ATR_Comment]);
+				(void)sprintf(log_buffer, msg_jobholdrel, pset, preq->rq_user,
+					preq->rq_host);
+				log_event(PBSEVENT_JOB, PBS_EVENTCLASS_JOB, LOG_INFO,
+					psubjob->ji_qs.ji_jobid, log_buffer);
+			}
+		}
+	}
+	if (pjob->ji_wattr[(int)JOB_ATR_hold].at_val.at_long == 0) {
```
arungrover: Instead of comparing it to 0 use HOLD_n

src/server/req_runjob.c
```diff
@@ -502,36 +502,45 @@ req_runjob(struct batch_request *preq)
...
+		}
 
 		if ((pjobsub = create_subjob(parent, jid, &j)) == NULL) {
 			req_reject(j, 0, preq);
 			return;
 		}
+
+		if (sub_run_version.at_flags & ATR_VFLAG_SET) {
+			pjobsub->ji_wattr[(int)JOB_ATR_run_version].at_val.at_long = sub_run_version.at_val.at_long;
+			pjobsub->ji_wattr[(int)JOB_ATR_run_version].at_flags |= (ATR_VFLAG_SET | ATR_VFLAG_MODCACHE);
```
arungrover: Why don't you set ATR_VFLAG_MODIFY on this attribute?

src/server/req_runjob.c
```diff
@@ -502,36 +502,45 @@ req_runjob(struct batch_request *preq)
...
 			reply_text(preq, PBSE_HOOKERROR, hook_msg);
 			return;
 		}
 		pjob = where_to_runjob(preq, pjobsub);
-		if (pjob == NULL) {
-			/* requeue subjob */
-			pjobsub->ji_qs.ji_substate = JOB_SUBSTATE_RERUN3;
-			job_purge(pjobsub);
-			return;
+		if (pjob) {
```
arungrover: what will happen when pjob returned is NULL?

src/server/req_runjob.c
```diff
@@ -1414,7 +1425,7 @@ post_sendmom(struct work_task *pwt)
 				 * array starts its comment is set to a begun message and
 				 * should not change after that
 				 */
-				if (jobp->ji_parentaj->ji_qs.ji_state != JOB_STATE_BEGUN) {
+				if (jobp->ji_parentaj->ji_qs.ji_state == JOB_STATE_QUEUED) {
```
arungrover: Please modify the comment above

src/server/req_runjob.c
Shrini-h: It doesn't matter as we will be doing full save later. Frankly I go this line form other places, where also it is missing.

src/server/req_runjob.c
Shrini-h: Addressed in https://github.com/PBSPro/pbspro/pull/1119/commits/2d3a5a6fe3c51e0f88bb63f19c7624fbca4a48d0

src/server/req_holdjob.c
Shrini-h: Adressed in https://github.com/PBSPro/pbspro/pull/1119/commits/2d3a5a6fe3c51e0f88bb63f19c7624fbca4a48d0

dilip-krishnan:
I guess I wrote this comment in a wrong place moving it to PR

"The question that comes here is, should we just held the failing subjob or the whole array job. Personally
I feel we should maintain the fail count in the parent job and at any point if the cumulative failure of subjobs reaches to 20 we should held the whole array job instead of putting a hold on the failing subjob.

This actually reminds me of a forum discussion on how the array job is percieved in PBS. AFAIK we consider array jobs as a part of single normal job, instead of considering each subjob as a normal job. Here is the link to the discussion http://community.pbspro.org/t/understanding-array-jobs/212/2 ."

>>Now subjobs breaching 20 run_count will move to held state H. Also there will be a system hold placed on its parent job Array.

From your proposed solution i do see that you are putting hold on parent job , but only on 20 failure on single subjob. But that can be problematic for large array job , where there are lot of failures on each subjob but none has reached 20.

Shrini-h:
> From your proposed solution i do see that you are putting hold on parent job , but only on 20 failure on single subjob. But that can be problematic for large array job , where there are lot of failures on each subjob but none has reached 20.

@dilip-krishnan  thank you for digging out your old discussion.

Your proposal sounds good, but does not seem to be practical. For example, in a small job array of 25 subjobs, if each of the subjob is requeued just once due to transient nw failures, we will end-up holding the array unnecessarily. Adding up the run_counts of all subjobs is also cumbersome and prone to inconsistency.

I suggest you checkout and use the recent discussion forum for this topic here http://community.pbspro.org/t/proposal-of-interface-regarding-hold-release-of-subjob-s-and-job-array/1442

dilip-krishnan:
>> Your proposal sounds good, but does not seem to be practical. For example, in a small job array of 25 subjobs, if each of the subjob is requeued just once due to transient nw failures, we will end-up holding the array unnecessarily. Adding up the run_counts of all subjobs is also cumbersome and prone to inconsistency.

So you are suggesting that 20 transient nw failures on a single subjob can put hold on parent job, but transient nw failures on 20 individual subjob should not. I don't see how that is practical. 

And usually array jobs are huge, so an array job of 1000 jobs can technically have 19000 failures before going to held state with the proposed solution. That doesn't seems to be addressing the issue here.  

I'll have a look on the discussion to get a deeper understanding at this.

Shrini-h:
> 
> So you are suggesting that 20 transient nw failures on a single subjob can put hold on parent job, but transient nw failures on 20 individual subjob should not. I don't see how that is practical.
> 
> And usually array jobs are huge, so an array job of 1000 jobs can technically have 19000 failures before going to held state with the proposed solution. That doesn't seems to be addressing the issue here.
> 

@dilip-krishnan, may be I have misunderstood your proposal.

> I'll have a look on the discussion to get a deeper understanding at this.

Thank's, please continue discussions there so that we can focus this PR page for code review only.

src/server/req_runjob.c
nithinj: I feel it's better to add the MODIFY flag so there is no issue even if we move the full save logic in the future.

Shrini-h: Added the flag in https://github.com/PBSPro/pbspro/pull/1119/commits/982610ec6b04f71ab4769eca4907fa8c612eb884

Shrini-h:
I have addressed all the comments, Please provide further comments or sign offs. Thank you

Shrini-h:
> I have addressed all the comments, Please provide further comments or sign offs. Thank you

@arungrover  and @nithinj

nithinj:
@Shrini-h ,
Thanks for making the change. I'll continue reviewing once we reach a consensus on the run_count discussion.

src/server/req_runjob.c
arungrover: A similar change is needed on line 588 as well

Shrini-h: Addressed in https://github.com/PBSPro/pbspro/pull/1119/commits/ec73366673c4c1438ad7dfb4ae574888aa17c57b

Shrini-h:
> @Shrini-h ,
> Thanks for making the change. I'll continue reviewing once we reach a consensus on the run_count discussion.

@nithinj Thanks for pointing out.
The consensus has been reached on the run_count discussion. Please continue the review or sign off

Shrini-h:
> Change looks good to me.
> @Shrini-h is there any plan to add more tests related to this fix?

@arungrover and @nithinj 
Thanks for approvals.

No plans to add more tests from my side. However if there is some test scenario that is recommended to add, I'm open to it.

arungrover:
I could be wrong but since this is a new thing that we are introducing, we should have a bunch of more tests around it. Like,  have a few subjobs in the expired state when the array is being held, submit job array with a different run_count and see it gets held, restart moms with held subjobs, check parent's job comment when it moves from H to B state, on a qrerun does subjob retain its run_count/run_version etc.

Shrini-h:
> I could be wrong but since this is a new thing that we are introducing, we should have a bunch of more tests around it. Like, have a few subjobs in the expired state when the array is being held, submit job array with a different run_count and see it gets held, restart moms with held subjobs, check parent's job comment when it moves from H to B state, on a qrerun does subjob retain its run_count/run_version etc.

OK, I will work and push the suggested PTL TCs.

Shrini-h:
> I could be wrong but since this is a new thing that we are introducing, we should have a bunch of more tests around it. Like, have a few subjobs in the expired state when the array is being held, submit job array with a different run_count and see it gets held, restart moms with held subjobs, check parent's job comment when it moves from H to B state, on a qrerun does subjob retain its run_count/run_version etc.

Added 4 new PTL tests to cover most of the recommended scenarios.
[ptl_out_27_may.txt](https://github.com/PBSPro/pbspro/files/3222254/ptl_out_27_may.txt)
[Test_report_logs_27_may.zip](https://github.com/PBSPro/pbspro/files/3222257/Test_report_logs_27_may.zip)

Did not add Tests case for `restart moms with held subjobs` as the subjobs are not held in the middle of run but before run. hence I feel scenario is not valid to test here.

@arungrover and @nithinj and @PSPro/maintainers
Please review and comment/ signoff / merge.

Thank you

test/tests/functional/pbs_test_run_count.py
```diff
@@ -65,6 +65,15 @@ def disable_reject_begin_hook(self):
         self.mom.log_match("h1.HK;copy hook-related file request received",
                            existence=True, starttime=start_time)
 
+    def enable_reject_begin_hook(self):
+        start_time = int(time.time())
```
nithinj: Do you need this start_time variable or can you use the time directly in log_match function if it not reused?

test/tests/functional/pbs_test_run_count.py
```diff
@@ -112,6 +121,19 @@ def test_less_than_20_run_count(self):
         """
         self.check_run_count(input_count="15", output_count="21")
 
+    def subjob_check(self, jid, sjid, maxruncount="21"):
```
nithinj: Can you add a doc string for functions added?

test/tests/functional/pbs_test_run_count.py
```diff
@@ -123,16 +145,87 @@ def test_run_count_subjob(self):
...
+        # Create an execjob_begin hook that rejects the job
+        self.create_reject_begin_hook()
+        time.sleep(2)
+        self.disable_reject_begin_hook()
 
-        self.server.expect(JOB, {ATTR_state: "H", ATTR_comment: (MATCH_RE,
-                           ja_comment)}, attrop=PTL_AND, id=jid)
+        a = {ATTR_J: '1-6'}
+        j = Job(TEST_USER, a)
+        j.set_sleep_time(10)
```
nithinj: Can you reduce the sleep time so that the overall test execution time would be less?

test/tests/functional/pbs_test_run_count.py
```diff
@@ -123,16 +145,87 @@ def test_run_count_subjob(self):
...
+        self.server.expect(JOB, {ATTR_state: "X"},
+                           id=j.create_subjob_id(jid, 2))
+
+        self.subjob_check(jid=jid, sjid=j.create_subjob_id(jid, 3))
+
+    def test_large_run_count_subjob(self):
+        """
+        Submit a job array with a large (>20) but valid run_count value and
+        check if the subjob and the parent are getting
+        held after 1 rejection from mom
```
nithinj: > held after 1 rejection from mom

This is confusing. Did you meant to say "after first job getting rejected run_count times?"

test/tests/functional/pbs_test_run_count.py
```diff
@@ -123,16 +145,87 @@ def test_run_count_subjob(self):
...
+
+        # Create an execjob_begin hook that rejects the job
+        self.create_reject_begin_hook()
+        time.sleep(2)
         self.disable_reject_begin_hook()
-        self.server.rlsjob(jid, 's')
+
+        a = {ATTR_W: "run_count=453", ATTR_J: '1-6'}
+        j = Job(TEST_USER, a)
+        j.set_sleep_time(10)
```
nithinj: Can you reduce the sleep time?

test/tests/functional/pbs_test_run_count.py
Shrini-h: Nope, can cause race condition.

test/tests/functional/pbs_test_run_count.py
Shrini-h: Its a helper function like other existing function which dont have docstrings either

test/tests/functional/pbs_test_run_count.py
Shrini-h: can cause problems in slower machines

test/tests/functional/pbs_test_run_count.py
Shrini-h: its clear, i mean "one" rejection not first. "1st" means first

test/tests/functional/pbs_test_run_count.py
Shrini-h: Can cause problem on slower machines

test/tests/functional/pbs_test_run_count.py
nithinj: got it. Thanks!

test/tests/functional/pbs_test_run_count.py
nithinj: Well, this helper function can be reused when we enhance the test suite so docstring is useful. But I'm leaving it up to you.

test/tests/functional/pbs_test_run_count.py
nithinj: I dont think we are accounting for such slower machines anywhere else in the framework. I guess it should be good enough with 2-3 seconds. In that case you wont be needing the following sleep of 8+7 either.

test/tests/functional/pbs_test_run_count.py
nithinj: But does that supposed to get held after "1" rejection or maxruncount rejections?

test/tests/functional/pbs_test_run_count.py
Shrini-h: Thanks, the subroutine is small enough to be understood what its doing without doc string. hence i'm passing on this

test/tests/functional/pbs_test_run_count.py
Shrini-h: This is not framework. There are many race conditions especially in job array test  cases that have cropped up on slow machines and have had corrections made using subsequent commits. Hence trying to be more careful here. trying to avoid race conditions. sleep of 8+7 is needed for faster machines

test/tests/functional/pbs_test_run_count.py
Shrini-h: This is the behavior for all jobs when run_count is more than 20. May be you are not familiar with this. please read through docs or experiment on product before commenting.

Shrini-h:
@nithinj I have justified all your comments.

@PBSPro/maintainers, @arungrover 
PR ready for merge

subhasisb:
@nithinj any further comments? @arungrover can you take this forward please? Thanks.

test/tests/functional/pbs_test_run_count.py
nithinj: There are two ways to go about this. Either you can start with lower values and re-adjust it whenever you detect a failure or you can go with an arbitrary upper limit which will never hit even in the slowest machines.
If we go with approach # 1 then need to update the sleep values whenever we observe a race condition or some other failure related to the slowness of the machine. But if you can identify the slow test machine which is the bottleneck, you can fixate the value in the beginning itself.
The approach you have chosen selects an arbitrary upper limit at the cost of test run time. Is there a reason why you have chosen 10s or is it just a random value? If every test writer chooses large sleep value, It could increase the total test execution time exponentially. Execution time is a usability feature and we should try to minimize it if we can.
In your test case the wait of 9s and 8s are done to see it the job in running state and to see it finished which is not required when you have jobs with a lower sleep time. If you still don't agree, please help me understand what problems can occur with respect to your test case in a slower machine?

Shrini-h: we cannot be expected to run this test on all slower machines to try converge on an accurate value (after all this code wont make it inside any rocket guidance systems). It has already been experienced in past. see PRs https://github.com/PBSPro/pbspro/pull/952 and https://github.com/PBSPro/pbspro/pull/918 where sleep time has been increased after experiencing problems on slower machines. 

Each person has definition of "large". and for me 10 seconds is not large. If, as per your definition of "large", 10 seconds is large, then please feel free to reduce it to what ever you feel is "small" before firing the tests. I wont be changing it here.

Shrini-h:
> @nithinj any further comments? @arungrover can you take this forward please? Thanks.

@arungrover and @subhasisb 
The only new comment from @nithinj  has been neutralized. 

So PR Ready for Merge !

arungrover:
@subhasisb, it was a holiday yesterday, I plan to do this review today.

test/tests/functional/pbs_test_run_count.py
```diff
@@ -102,3 +120,112 @@ def test_less_than_20_run_count(self):
...
+        Submit a job array and check if the subjob and the parent are getting
+        held after 20 rejection from mom when there is another subjob in X
+        """
+        self.server.manager(MGR_CMD_SET, NODE,
+                            {'resources_available.ncpus': 1},
+                            id=self.mom.shortname)
+
+        # Create an execjob_begin hook that rejects the job
+        self.create_reject_begin_hook()
+        time.sleep(2)
```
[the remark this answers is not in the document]
Shrini-h: Addressed in https://github.com/PBSPro/pbspro/pull/1119/commits/9aed3c2973459d4f123b0f3f8d4c61256902fe67

Reran `Test_run_count` suite again. Report: [ptl_out_29_may.txt](https://github.com/PBSPro/pbspro/files/3230926/ptl_out_29_may.txt)

Shrini-h:
PR Ready for Merge!

arungrover:
@Shrini-h you still need a sign-off from Nithin

Shrini-h:
@nithinj  do you have anymore review comments.? 
else pls comment your approval

nithinj:
I've no further comments/suggestions. Thanks!

Shrini-h:
PR ready for merge!. Take it or leave it !

PR #1121, opened 2019-05-09
Support in PTL log_match() for microsec logging

tpai26:
#### Describe Bug or Feature
Currently log_match() is not able to match messages with microsecond timestamps. It will only match messages till seconds. This PR will allow starttime and endtime in log_match() to be able to accept time in microseconds. This will help us test out log messages with microsecond timestamps. 

[microsec_log.txt](https://github.com/PBSPro/pbspro/files/3192675/microsec_log.txt)
[before_change_log.txt](https://github.com/PBSPro/pbspro/files/3192679/before_change_log.txt)


test/fw/ptl/utils/pbs_logutils.py
```diff
@@ -365,13 +365,13 @@ def match_msg(self, lines, msg, allmatch=False, regexp=False,
         if lines:
             for l in lines:
                 if starttime is not None:
-                    # l[:19] captures the log record time
-                    tm = self.convert_date_time(l[:19])
+                    # l.split(';', 1)[0] gets the time stamp string
+                    tm = self.convert_date_time(l.split(';', 1)[0])
```
agrawalravi90: Do the split only once so you can reuse the value in "if endtime is not None" block

test/tests/selftest/pbs_microsecondlogging.py
```diff
@@ -0,0 +1,86 @@
+# coding: utf-8
+
+# Copyright (C) 1994-2018 Altair Engineering, Inc.
```
agrawalravi90: 2019

test/tests/selftest/pbs_microsecondlogging.py
```diff
@@ -0,0 +1,86 @@
...
+# "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
+# trademark licensing policies.
+
+from tests.selftest import *
+import time
+
+
+class TestMicrosecondLogging(TestSelf):
+    """
+    Test pbs microsec logging
```
agrawalravi90: If this is a selftest, then you are not testing PBS's microsecond logging feature right?  You are actually testing PTL's support for microsecond logging.

test/tests/selftest/pbs_microsecondlogging.py
```diff
@@ -0,0 +1,86 @@
...
+        """
+        du = DshUtils()
+        if hostname is None:
+            hostname = self.server.hostname
+        a = {'PBS_LOG_HIGHRES_TIMESTAMP': highrestimestamp}
+        self.du.set_pbs_config(hostname=hostname, confs=a, append=True)
+        PBSInitServices().restart()
+        self.assertTrue(self.server.isUp(), 'Failed to restart PBS Daemons')
+
+    def test_microsec_logging(self):
```
agrawalravi90: how about "test_log_match_microsec" since you are actually testing log_match?

test/tests/selftest/pbs_microsecondlogging.py
```diff
@@ -0,0 +1,86 @@
...
+from tests.selftest import *
+import time
+
+
+class TestMicrosecondLogging(TestSelf):
+    """
+    Test pbs microsec logging
+    """
+
+    def setup_microsecondlogging(self, hostname=None, highrestimestamp=1):
```
agrawalravi90: "setup_microsecondlogging" makes me think that this method should always turn microsecond logging ON. How about calling it 'switch_microsecondlogging" instead?

test/tests/selftest/pbs_microsecondlogging.py
```diff
@@ -41,32 +41,33 @@
...
+    values are in microseconds
     """
 
-    def setup_microsecondlogging(self, hostname=None, highrestimestamp=1):
+    def switch_microsecondlogging(self, hostname=None, highrestimestamp=1):
         """
         Set micro second logging in pbs.conf
         """
-        du = DshUtils()
+        self.du = DshUtils()
```
agrawalravi90: You shouldn't have to initialize 'du', PTL already provides you with one, just try using 'self.du' without initializing it.

test/tests/selftest/pbs_microsecondlogging.py
```diff
@@ -41,32 +41,33 @@
...
-    def test_microsec_logging(self):
+    def test_log_match_microsec_logging(self):
         """
         Test that log_match will work when microsecond logging
         is turned on and then off
         """
 
         # Turn of microsecond logging and test log_match()
-        self.setup_microsecondlogging(highrestimestamp=1)
+        self.st = time.time()
```
agrawalravi90: Why are you making 'st' and 'et' Class variables? they are local to this test case, so i think you should just declare them as local variables (without the 'self')

test/tests/selftest/pbs_microsecondlogging.py
```diff
@@ -0,0 +1,86 @@
...
+        jid1 = self.server.submit(J1)
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid1)
+        msg_str = "Job;" + jid1 + ";Job Run at request of Scheduler"
+        self.et = time.time()
+        self.server.log_match(msg_str, starttime=self.st, endtime=self.et)
+        self.server.deljob(jid1, wait=True)
+
+        # Turn off microsecond logging and test log_match()
+        self.st = time.time()
```
bhroam: You need to int() time.time() for logging to work without microsecond logging.

test/tests/selftest/pbs_log_match.py
```diff
@@ -0,0 +1,84 @@
...
+        self.assertTrue(self.server.isUp(), 'Failed to restart PBS Daemons')
+
+    def test_log_match_microsec_logging(self):
+        """
+        Test that log_match will work when microsecond logging
+        is turned on and then off
+        """
+
+        # Turn of microsecond logging and test log_match()
+        st = time.time()
```
bhroam: Put this right above the submit.  Right now it's above the restart of PBS which takes over a second.  We want to test fractions of a second.

PR #1122, opened 2019-05-09
Fix manager() adding extra quotes

vstumpf:
#### Describe Bug or Feature
 A previous check-in #1089 introduced a bug where PTL would quote string values if it contained certain characters. This would fail with string arrays. It would also fail with tests that added quotes to the strings so qmgr could import it; it would double-quote some strings. This will cause tests that set Fairshare order to fail. 

#### Describe Your Change
manager() will now also take lists of strings.
Design: https://pbspro.atlassian.net/wiki/spaces/PD/pages/1281458177/manager+should+use+python+lists+for+string+arrays

**String Arrays**
If any of the strings in the list have quotes or special characters, it will separately set each one, so qmgr does not get confused as to what is a string boundary and what is actually in the string.
Otherwise, it will join the string with commas and fall through to the strings case.

**Strings**
If the first and last character are both single or double quotes, don't add quotes.

~~[after-fix-self.txt](https://github.com/PBSPro/pbspro/files/3180075/after-fix-self.txt)~~
[after-fix8.txt](https://github.com/PBSPro/pbspro/files/3184749/after-fix8.txt)
[after-fix9.txt](https://github.com/PBSPro/pbspro/files/3184750/after-fix9.txt)
[after-fix10.txt](https://github.com/PBSPro/pbspro/files/3188475/after-fix10.txt)

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -6527,11 +6527,40 @@ class Server(PBSService):
                                 v = v[1]
                             else:
                                 op = '='
-                            # handle string arrays as double quotes if
-                            # not already set:
+                            if isinstance(v, list) and not isinstance(v, str):
+                                # handle string arrays with strings
+                                # that contain special characters
+                                # with multiple manager calls
+                                if any((c in vv) for c in set(', \'\n"')
+                                       for vv in v):
+                                    if op == '+=' or op == '=':
+                                        oper = INCR
+                                    else:
+                                        oper = DECR
+                                    for vv in v:
+                                        a = {k: (oper, vv)}
+                                        rc = self.manager(cmd=cmd,
+                                                          obj_type=obj_type,
+                                                          attrib=a, id=id,
+                                                          extend=extend,
+                                                          level=level,
+                                                          sudo=sudo,
+                                                          runas=runas,
+                                                          logerr=logerr)
+                                        if rc:
+                                            return rc
+                                    return 0
+                                # if there are no special characters, then
+                                # join the list and parse it normally.
+                                v = ','.join(v)
                             if isinstance(v, str):
-                                if ',' in v and v[0] != '"':
+                                # don't quote if already quoted
+                                if v[0] == v[-1] and v[0] in set('"\''):
+                                    pass
+                                # handle string arrays
+                                elif ',' in v and v[0] != '"':
                                     v = '"' + v + '"'
+                                # handle strings that need to be quoted
                                 elif any((c in v) for c in set(', \'\n"')):
                                     if '"' in v:
                                         v = "'%s'" % v
--- a/test/tests/functional/pbs_python_restart_settings.py
+++ b/test/tests/functional/pbs_python_restart_settings.py
@@ -256,7 +256,8 @@ localnode = pbs.get_local_nodename()
 vn = pbs.server().vnode(localnode)
 pbs.event().accept()
 """
-        a = {'event': "queuejob,movejob,modifyjob,runjob", 'enabled': "True"}
+        a = {'event': ["queuejob", "movejob", "modifyjob", "runjob"],
+             'enabled': "True"}
         self.server.create_import_hook("test", a, hook_body, overwrite=True)
         # Create workq2
         a = {'queue_type': 'e', 'started': 't', 'enabled': 't'}
@@ -337,7 +338,7 @@ pbs.event().accept()
 import pbs
 pbs.event().accept()
 """
-        a = {'event': "queuejob,modifyjob", 'enabled': 'True'}
+        a = {'event': ["queuejob", "modifyjob"], 'enabled': 'True'}
         self.server.create_import_hook("test", a, hook_body, overwrite=True)
         # Set max_objects and min_interval so that further changes
         # will generate a log message.
--- a/test/tests/functional/pbs_qmgr.py
+++ b/test/tests/functional/pbs_qmgr.py
@@ -165,11 +165,11 @@ class TestQmgr(TestFunctional):
         Set the node's comment, then print it and re-import it
         """
         a = {'comment': comment}
-        self.server.manager(MGR_CMD_SET, NODE, a, id=self.mom.hostname)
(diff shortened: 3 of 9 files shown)


test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -6527,8 +6527,13 @@ def manager(self, cmd, obj_type, attrib=None, id=None, extend=None,
...
                             # not already set:
                             if isinstance(v, str):
-                                if ',' in v and v[0] != '"':
+                                # don't quote if already quoted
+                                if v[0] == v[-1] and v[0] in set('"\''):
+                                    pass
+                                # handle string arrays
+                                elif ',' in v and v[0] != '"':
                                     v = '"' + v + '"'
+                                # handle strings that need to be quoted
```
latha-subramanian: Changes look good to me

vstumpf:
I've made this WIP due to some tests failing. I think the best way to continue this fix is to change how string arrays are handled with PTL. 

Expect a design doc soon.

vstumpf:
I've opened this up for review again.
Please re-review. I've dismissed the previous reviews, as I've added new changes.
I also fixed a test (TestQmgr) that was added in the previous PR.

test/tests/functional/pbs_python_restart_settings.py
```diff
@@ -256,8 +256,10 @@ def test_max_hooks(self):
 vn = pbs.server().vnode(localnode)
 pbs.event().accept()
 """
-        a = {'event': "queuejob,movejob,modifyjob,runjob", 'enabled': "True"}
+        a = {'event': ["queuejob", "movejob", "modifyjob", "runjob"],
+             'enabled': "True"}
         self.server.create_import_hook("test", a, hook_body, overwrite=True)
+        return
```
lisa-altair: Why did you add this return?  I don't think it should be here.

test/tests/functional/pbs_qmgr.py
```diff
@@ -165,12 +165,12 @@ def set_and_test_comment(self, comment):
...
         a = {'comment': comment}
-        self.server.manager(MGR_CMD_SET, NODE, a, id=self.mom.hostname)
+        self.server.manager(MGR_CMD_SET, NODE, a, id=self.mom.shortname)
         qmgr_path = \
             os.path.join(self.server.pbs_conf['PBS_EXEC'], 'bin', 'qmgr')
         qmgr_cmd_print = qmgr_path + \
-            (' -c "p n %s comment"' % self.mom.hostname)
-        ret = self.du.run_cmd(self.server.hostname,
+            (' -c "p n %s comment"' % self.mom.shortname)
+        ret = self.du.run_cmd(self.server.shortname,
```
lisa-altair: Why are you changing this to self.server.shortname?

test/tests/selftest/pbs_manager.py
```diff
@@ -0,0 +1,63 @@
...
+        """
+        attr = {}
+        attr['type'] = 'string_array'
+        attr['flag'] = 'h'
+        r = 'foo2'
+        rc = self.server.manager(
+            MGR_CMD_CREATE, RSC, attr, id=r, runas=ROOT_USER, logerr=False)
+        self.assertEqual(rc, 0)
+        self.server.manager(MGR_CMD_SET, SERVER,
+                            {"resources_available.foo2": ["A'A", 'B"B', 'C C']}
```
lisa-altair: Should there be a test for commas and newlines too?

test/tests/selftest/pbs_manager.py
```diff
@@ -0,0 +1,63 @@
...
+    def test_pbs_conf(self):
+        """
+        Test setting an string array with quote characters
+        """
+        attr = {}
+        attr['type'] = 'string_array'
+        attr['flag'] = 'h'
+        r = 'foo2'
+        rc = self.server.manager(
+            MGR_CMD_CREATE, RSC, attr, id=r, runas=ROOT_USER, logerr=False)
```
bhroam: You don't need to run this as ROOT_USER.  PTL makes the user who runs pbs_benchpress a manager.  That is sufficient.

test/tests/selftest/pbs_manager.py
```diff
@@ -0,0 +1,63 @@
...
+    Test ptl.server.manager()
+    """
+
+    def test_pbs_conf(self):
+        """
+        Test setting an string array with quote characters
+        """
+        attr = {}
+        attr['type'] = 'string_array'
+        attr['flag'] = 'h'
```
bhroam: Just curious, why use 3 lines when this should all fit on one?

test/tests/selftest/pbs_manager.py
```diff
@@ -0,0 +1,63 @@
...
+        r = 'foo2'
+        rc = self.server.manager(
+            MGR_CMD_CREATE, RSC, attr, id=r, runas=ROOT_USER, logerr=False)
+        self.assertEqual(rc, 0)
+        self.server.manager(MGR_CMD_SET, SERVER,
+                            {"resources_available.foo2": ["A'A", 'B"B', 'C C']})
+        nodes = self.server.filter(SERVER,
+                                   {'resources_available.foo2':
+                                    """A'A,B"B,C C"""})
+        self.logger.info(nodes)
```
bhroam: You're querying the server, nodes is a poor variable name.  Also, don't log a blanket variable.  Either turn it into a full log message, or maybe assert instead.

test/tests/selftest/pbs_manager.py
vstumpf: It's weird, even though the code is there in pbs, it doesn't support new lines or commas in string array values.

latha-subramanian:
@vstumpf, Are the attached test results the latest ? Logs show failures. Please attach the latest logs.

vstumpf:
Sorry, that was an old version of the test. I ran the new test without resaving the output.
I've attached the new output.

latha-subramanian:
> Sorry, that was an old version of the test. I ran the new test without resaving the output.
> I've attached the new output.



> Sorry, that was an old version of the test. I ran the new test without resaving the output.
> I've attached the new output.

Thanks @vstumpf

vstumpf:
I've rebased, please take this forward @bhroam

PR #1125, opened 2019-05-13
remove extra double quotes in events in order to load hook

vccardenas:
#### Describe Bug or Feature
Extra '"' in events string causes the pbs_cgroups hook to not get loaded.

#### Describe Your Change
Remove the extra '"' in the events string in load_hook() and tearDown() in pbs_cgroups_hook.py.

[pbs-22861_after.txt](https://github.com/PBSPro/pbspro/files/3175034/pbs-22861_after.txt)
[pbs-22861_before.txt](https://github.com/PBSPro/pbspro/files/3175035/pbs-22861_before.txt)

### DIFF
--- a/test/tests/functional/pbs_cgroups_hook.py
+++ b/test/tests/functional/pbs_cgroups_hook.py
@@ -844,9 +844,9 @@ if %s e.job.in_ms_mom():
                 script = fd.read()
         except IOError:
             self.assertTrue(False, 'Failed to open hook file %s' % filename)
-        events = '"execjob_begin,execjob_launch,execjob_attach,'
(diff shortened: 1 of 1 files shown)


mike0042:
Just to be clear, this update is necessary due to a change in the test harness. It had been working prior to that, correct?

vccardenas:
@mike0042 , yes that is correct, the test had been working prior to the change in pbs_testlib.py in PR 1089.  I inspected the other tests and only this one is using the events = '"<event>"' format.  The other tests use the form events = '<event>' or events = "<event>", and are passing.

PR #1127, opened 2019-05-15
Use initgroups to get user's supplementary groups

arwild01:
#### Describe Bug or Feature

Direct Write of Output and Error files fails to work if the target pathname resides under a directory that is restricted to a group that's not the user's primary gid.  The solution is to access the user's supplementary groups in addition to their primary gid.

#### Describe Your Change

Inserted call(s) to initgroups().  Unfortuantely, initgroups needs the username (string) and not just the uid.  I considered modifying impersonate_user to require a char *eusrname, but it turns out to implement this you modify the api of several other functions.  Using getpwuid to look up the username means the api of imerpersonate_user() and revert_from_user() can be retained.

#### Link to Design Doc

Sorry... don't have one.  I'm customer trying to request this commercially and decided to take a stab at the code myself.

Test logs are attached in the comments.

### DIFF
--- a/src/resmom/start_exec.c
+++ b/src/resmom/start_exec.c
@@ -901,6 +901,14 @@ internal_mkjobdir(char *jobid, char *jobdir)
 int
 impersonate_user(uid_t uid, gid_t gid)
 {
+#if defined(HAVE_GETPWUID) && defined(HAVE_INITGROUPS)
+	struct passwd *pwd = getpwuid(uid);
+
+	if (initgroups(pwd->pw_name, gid) == -1) {
+		return -1;
+	}
+
+#endif
 #if defined(HAVE_SETEUID) && defined(HAVE_SETEGID)
 	/* most systems */
 	if ((setegid(gid) == -1) ||
@@ -923,15 +931,23 @@ impersonate_user(uid_t uid, gid_t gid)
 void
 revert_from_user(void)
 {
-#if defined(HAVE_SETEUID) && defined(HAVE_SETEGID)
+#if defined(HAVE_SETEUID)
 	/* most systems */
 	(void)seteuid(0);
-	(void)setegid(pbsgroup);
-#elif defined(HAVE_SETRESUID) && defined(HAVE_SETRESGID)
+#elif defined(HAVE_SETRESUID)
 	(void)setresuid(-1, 0, -1);
+#else
+#error No function to change effective UID
+#endif
+#if defined(HAVE_INITGROUPS)
+	(void)initgroups("root", pbsgroup);
+#endif
+#if defined(HAVE_SETEGID)
+	(void)setegid(pbsgroup);
+#elif defined(HAVE_SETRESGID)
 	(void)setresgid(-1, pbsgroup, -1);
 #else
-#error No function to change effective UID or GID
+#error No function to change effective GID
 #endif
 }
 
--- a/src/resmom/start_exec.c
+++ b/src/resmom/start_exec.c
@@ -903,6 +903,8 @@ impersonate_user(uid_t uid, gid_t gid)
 {
 #if defined(HAVE_GETPWUID) && defined(HAVE_INITGROUPS)
 	struct passwd *pwd = getpwuid(uid);
+	if (pwd == NULL)
+		return -1;
 
 	if (initgroups(pwd->pw_name, gid) == -1) {
 		return -1;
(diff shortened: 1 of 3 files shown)


src/resmom/start_exec.c
```diff
@@ -901,6 +901,16 @@ internal_mkjobdir(char *jobid, char *jobdir)
 int
 impersonate_user(uid_t uid, gid_t gid)
 {
+#if defined(HAVE_GETPWUID) && defined(HAVE_INITGROUPS)
+	struct passwd *pwd = getpwuid(uid);
+	if (pwd == NULL)
+		return -1;
+
+	if (initgroups(pwd->pw_name, gid) == -1) {
```
mike0042: I suggest you use getgroups()/setgroups() rather than initgroups() to accomplish this. The man page for initgroups() specifically mentions the /etc/group file is used. Environments using LDAP, NIS, etc. to manage user data would not utilize the settings in /etc/nsswitch.conf as one might expect. You may also need to store a list of supplementary groups to be restored in revert_from_user() since there is no "effective" supplementary group list, only the "real" one. The saved list could be stored as a static since it only needs to be looked up the first time impersonate_user() is called.

arwild01: Based on testing in my own environment where we use centrify (Active directory) and NIS (in addition to local /etc/group files).  It works just fine.  I'm actually using very similiar logic from python in one of my hooks (which is configured to run as "pbsadmin", but becomes the user (with all their groups via the initgroups and setresgid/setresuid calls).  Looking at the source code for initgroups (at least from GNU libc) is based on getgroups()/setgroups() calls.  

See:   https://github.com/lattera/glibc/blob/master/grp/initgroups.c  

(not saying this is the "definitive" version, but it's what a quick google search returned).

I'll admit I don't completely "get" the intention of revert_from_user().  The name implies it wants to restore to the previous state... but (as you noted) the previous state isn't saved (and that includes the uid/gid).  As it stands the code is just hardcoded to are return you to root.  As such I think it's appropriate (and certainly simpler) to just return the supplementary group list to whatever root's "real".... supplementary real group list is.

mike0042: You're right about initgroups(). I was concerned because the the CentOS7 man page reads:
```
DESCRIPTION
       The initgroups() function initializes the group access list by reading the group
       database /etc/group and using all groups of which user is a member.   The  addi‐
       tional group "group" is also added to the list.

       The user argument must be non-NULL.
```
The should probably change it to indicate it uses the underlying NSS functionality.

We only modify the effective user ID in impersonate_user(), so revert_from_user() just sets it back. But if we're going to modify the supplemental groups, we need to set those back as well. These functions are primarily used for file staging. We could have also changed the uid/gid after a fork/exec and never had to change the uid/gid of the parent process. I'm sure there were reasons someone chose to implement it this way, but I couldn't tell you exactly what they were.

arwild01: I agree it makes sense to not leave the supplementary groups the way "impersonate_user()" left them... that's why I added the call to initgroups("root",....) in revert_from_user().  I realize that it's "notionally possible" that something else could have modified the supplementary groups list so restoring to root's default list may not be the "right" thing.  Certainly adding something to store the original list isn't hard to do... I'm not sure in PBS's design where the right place to store something like that is.  

However, I think it's reasonable to assume that root's supplementary groups coming into PBS are the same thing that initgroup("root", ...) returns.  Changing supplementary groups isn't common... I don't believe there's a standard userspace command to even do it.

mike0042: Agreed. I'm fine with what you have. Thanks for talking me through.

mike0042: Could you please take a look at test/tests/functional/pbs_qsub_direct_write.py and let us know if you would be opposed to adding a test?

arwild01: I took a stab at writing one test:  test_direct_write_when_job_succeeds_controlled

I'm typing this message before Travis has run so I have no idea if it's even syntactically correct, but let's see what happens...

mike0042:
@arwild01, please attach the log of the qsub direct write tests in the PR description.

arwild01:
@mike0042 as request here's the ptl_output.txt from the one test (after fixing the issues with it)

[ptl_output.txt](https://github.com/PBSPro/pbspro/files/3204476/ptl_output.txt)

mike0042:
Merged. Thanks again @arwild01!

PR #1128, opened 2019-05-15
Fairshare usage can become << 1, leading to unexpected errors

agrawalravi90:
#### Describe Bug or Feature

Inside decay_fairshare_tree(), scheduler has a check to prevent the usage from going below 0, but it's not very effective as usage is a float value:

	root->usage *= conf.fairshare_decay_factor;
    
    	if (root->usage == 0)
    		root->usage = 1;

So, the usage can end up becoming really small (like 1e-40 small), which can end up causing various errors:
- If the job_sort_formula is set to something unrelated to fairshare, like 'queue_priority',  it starts failing once fairshare usage factor reaches such small values because the value of a resource_resv's ginfo->usage_factor gets set to 'nan' after a call to calc_usage_factor(), and 'fairshare_tree_usage' is  part of the data that's passed to Python 'formula_evaluate'. It fails as:
  File "<string>", line 1, in <module>NameError: name 'nan' is not defined

- Then, if job_sort_formula is set to "pow(2,(fairshare_tree_usage/fairshare_perc))", I noticed the scheduler logging errors w.r.t formula evaluation after some decay cycles:

[ravi@pbspro sched_logs]$ grep "Formula evaluation for job had an error" *
20190515:05/15/2019 09:25:54;0100;pbs_sched;Job;2.pbspro;Formula evaluation for job had an error. Zero value will be used: float division by zero
20190515:05/15/2019 09:25:54;0100;pbs_sched;Job;3.pbspro;Formula evaluation for job had an error. Zero value will be used: float division by zero
20190515:05/15/2019 09:25:54;0100;pbs_sched;Job;0.pbspro;Formula evaluation for job had an error. Zero value will be used: float division by zero
20190515:05/15/2019 09:25:54;0100;pbs_sched;Job;1.pbspro;Formula evaluation for job had an error. Zero value will be used: float division by zero
20190515:05/15/2019 09:28:38;0100;pbs_sched;Job;2.pbspro;Formula evaluation for job had an error. Zero value will be used: float division by zero
20190515:05/15/2019 09:28:38;0100;pbs_sched;Job;3.pbspro;Formula evaluation for job had an error. Zero value will be used: float division by zero
20190515:05/15/2019 09:28:38;0100;pbs_sched;Job;2.pbspro;Formula evaluation for job had an error. Zero value will be used: float division by zero
20190515:05/15/2019 09:28:38;0100;pbs_sched;Job;3.pbspro;Formula evaluation for job had an error. Zero value will be used: float division by zero

#### Describe Your Change

The fix is pretty simple, we just change the condition "if (root->usage == 0)" to "if (root->usage < 1)"

[before_fix.log](https://github.com/PBSPro/pbspro/files/3181973/before_fix.log)
[after_fix.log](https://github.com/PBSPro/pbspro/files/3181974/after_fix.log)

### DIFF
--- a/src/scheduler/constant.h
+++ b/src/scheduler/constant.h
@@ -314,6 +314,8 @@ enum fairshare_flags
 	FS_TRIM = 1
 };
 
+#define FAIRSHARE_MIN_USAGE 1
+
 /* flags used for copy constructors - bit field */
 enum dup_flags
 {
--- a/test/tests/functional/pbs_fairshare.py
+++ b/test/tests/functional/pbs_fairshare.py
@@ -366,13 +366,12 @@ class TestFairshare(TestFunctional):
 
         t = int(time.time())
         time.sleep(2)
-
         self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'True'})
-        self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'False'})
 
-        self.scheduler.log_match("Decaying Fairshare Tree", starttime=t,
-                                 max_attempts=5)
+        self.scheduler.log_match("Decaying Fairshare Tree", starttime=t)
 
         # Check that TEST_USER's usage is 1
         fs = self.scheduler.query_fairshare(name=str(TEST_USER))
-        self.assertEquals(fs.usage, 1)
+        fs_usage = int(fs.usage)
+        self.assertEquals(
+            fs_usage, 1, "Fairshare usage %d not equal to 1" % int(fs_usage))
--- a/test/tests/functional/pbs_fairshare.py
+++ b/test/tests/functional/pbs_fairshare.py
@@ -351,3 +351,29 @@ class TestFairshare(TestFunctional):
         job_order = [jid3, jid4]
         for i in range(len(job_order)):
             self.assertEqual(job_order[i].split('.')[0], c.political_order[i])
+
+    def test_fairshare_decay_min_usage(self):
+        """
+        Test that fairshare decay doesn't reduce the usage below 1
+        """
+        self.scheduler.set_sched_config({'fair_share': 'True'})
+        self.scheduler.set_sched_config({'log_filter': 0})
+        self.scheduler.add_to_resource_group(TEST_USER, 10, 'root', 50)
+        self.scheduler.set_fairshare_usage(TEST_USER, 1)
+        self.scheduler.set_sched_config({"fairshare_decay_time": "00:00:02"})
+        self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'True'})
+        self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'False'})
+
+        t = int(time.time())
+        # Sleep for 2 seconds
+        time.sleep(2)
+
+        self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'True'})
+        self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'False'})
+
+        self.scheduler.log_match("Decaying Fairshare Tree", starttime=t,
+                                 max_attempts=5)
+
+        # Check that TEST_USER's usage is 1
+        fs = self.scheduler.query_fairshare(name=str(TEST_USER))
+        self.assertEquals(fs.usage, 1)
--- a/test/tests/functional/pbs_fairshare.py
+++ b/test/tests/functional/pbs_fairshare.py
@@ -373,5 +373,5 @@ class TestFairshare(TestFunctional):
         # Check that TEST_USER's usage is 1
         fs = self.scheduler.query_fairshare(name=str(TEST_USER))
         fs_usage = int(fs.usage)
-        self.assertEquals(
-            fs_usage, 1, "Fairshare usage %d not equal to 1" % int(fs_usage))
+        self.assertEquals(fs_usage, 1,
+                          "Fairshare usage %d not equal to 1" % fs_usage)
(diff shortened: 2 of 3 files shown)


src/scheduler/constant.h
```diff
@@ -314,6 +314,8 @@ enum fairshare_flags
 	FS_TRIM = 1
 };
 
+#define FAIRSHARE_MIN_USAGE 1
```
bhroam: Modify new_group_info to use this macro instead of initializing usage and temp_usage to 1.

test/tests/functional/pbs_fairshare.py
```diff
@@ -351,3 +351,29 @@ def test_pbsfs(self):
...
+        self.scheduler.set_sched_config({'fair_share': 'True'})
+        self.scheduler.set_sched_config({'log_filter': 0})
+        self.scheduler.add_to_resource_group(TEST_USER, 10, 'root', 50)
+        self.scheduler.set_fairshare_usage(TEST_USER, 1)
+        self.scheduler.set_sched_config({"fairshare_decay_time": "00:00:02"})
+        self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'True'})
+        self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'False'})
+
+        t = int(time.time())
+        # Sleep for 2 seconds
```
bhroam: The code says this.  Either say why you are sleeping or remove this comment.

test/tests/functional/pbs_fairshare.py
```diff
@@ -351,3 +351,29 @@ def test_pbsfs(self):
...
+        time.sleep(2)
+
+        self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'True'})
+        self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'False'})
+
+        self.scheduler.log_match("Decaying Fairshare Tree", starttime=t,
+                                 max_attempts=5)
```
bhroam: Why set max_attempts=5 instead of leaving it to the default.  This might lead to a race condition on slow machines.

test/tests/functional/pbs_fairshare.py
```diff
@@ -351,3 +351,29 @@ def test_pbsfs(self):
...
+        """
+        Test that fairshare decay doesn't reduce the usage below 1
+        """
+        self.scheduler.set_sched_config({'fair_share': 'True'})
+        self.scheduler.set_sched_config({'log_filter': 0})
+        self.scheduler.add_to_resource_group(TEST_USER, 10, 'root', 50)
+        self.scheduler.set_fairshare_usage(TEST_USER, 1)
+        self.scheduler.set_sched_config({"fairshare_decay_time": "00:00:02"})
+        self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'True'})
+        self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'False'})
```
arungrover: Why do you set scheduling to False here and below? When there is no workload then just setting scheduling to True should do it, right?

arungrover:
Forgot to mention, use the new macro in new_group_info as well

test/tests/functional/pbs_fairshare.py
agrawalravi90: Well, after the manager set scheduling=False call, this log message should really be there, so no point waiting 60 attempts. I actually think we should just make it 1

test/tests/functional/pbs_fairshare.py
agrawalravi90: I set it to False because PTL makes sure that the sched cycle is over before returning control back. So, this is just to make sure that the sched cycle is over before we proceed. I actually think we should enhance PTL to just do this for scheduling=True, but that'd be out of scope for this PR.

agrawalravi90:
@arungrover and @bhroam Thanks for the review, I've addressed your comments, please let me know what you think.

test/tests/functional/pbs_fairshare.py
arungrover: I don't think that is necessary. Your test setup has nothing in it (no jobs & reservation). Your test sleeps for 2 seconds between the cycles, and then there is a log match which will be attempted 5 times. I think there is enough time to complete empty cycles.

src/scheduler/constant.h
arungrover: I didn't realize Bhroam had given the same comment :)

test/tests/functional/pbs_fairshare.py
agrawalravi90: hmm .. ok, I guess I was being a bit pedantic, I'll remove the turning scheduling off. But I think I'll remove the 5 attempts and let it be default, that way it's still guaranteed to run ok on slower machines.

agrawalravi90:
@arungrover I've addressed your comment, please let me know if it looks okay now.

test/tests/functional/pbs_fairshare.py
agrawalravi90: Ok, after addressing Arun's comment, I've removed max_attempts, leaving it to default, please let me know if it looks fine now.

test/tests/functional/pbs_fairshare.py
```diff
@@ -366,13 +366,12 @@ def test_fairshare_decay_min_usage(self):
...
-        self.scheduler.log_match("Decaying Fairshare Tree", starttime=t,
-                                 max_attempts=5)
+        self.scheduler.log_match("Decaying Fairshare Tree", starttime=t)
 
         # Check that TEST_USER's usage is 1
         fs = self.scheduler.query_fairshare(name=str(TEST_USER))
-        self.assertEquals(fs.usage, 1)
+        fs_usage = int(fs.usage)
+        self.assertEquals(
+            fs_usage, 1, "Fairshare usage %d not equal to 1" % int(fs_usage))
```
bhroam: You've already int()'d fs_usage.  You don't need to do it again.

agrawalravi90:
@bhroam thanks, I've removed the cast, please let me know if it looks fine now

agrawalravi90:
Thanks @bhroam ! @arungrover Can you please see if you are happy with the latest commit and merge it if possible? Thanks.

PR #1129, opened 2019-05-15
Address rpmlint warning: dangerous-command

minghui-liu:
#### Describe Bug or Feature
rpmlint complains about dangerous commands in `%post`, `%preun` and `%posttrans` sections in the RPM spec file

#### Describe Your Change
* Copy `pbs_init.d` script to `/etc/init.d/pbs` during the `%install` phase rather than `%post`
* Move commands in `%preun` and `%posttrans` to separate scripts
* Removed the copying of `python/pbs_bootcheck.py` to `var/tmp`
 
[build_and_test.txt](https://github.com/PBSPro/pbspro/files/3189017/dangerous_cmd_build_and_test.txt)

### DIFF
--- a/pbspro.spec
+++ b/pbspro.spec
@@ -308,6 +308,7 @@ cd build
 %make_install
 mandir=$(find %{buildroot} -type d -name man)
 [ -d "$mandir" ] && find $mandir -type f -exec gzip -9 -n {} \;
+install -D %{buildroot}/%{pbs_prefix}/libexec/pbs_init.d %{buildroot}/etc/init.d/pbs
 
 %post %{pbs_server}
 ldconfig %{_libdir}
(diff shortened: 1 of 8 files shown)


src/cmds/scripts/pbs_preuninstall
```diff
@@ -0,0 +1,63 @@
...
+have_systemd="${4:-0}"
+
+if [ `basename $pbs_exec` = $pbs_version ]; then
+    top_level=`dirname $pbs_exec`
+    if [ -h $top_level/default ]; then
+        link_target=`readlink $top_level/default`
+        [ `basename "$link_target"` = $pbs_version ] && rm -f $top_level/default
+    fi
+fi
+rm -f /opt/modulefiles/pbs/%{version}
```
mike0042: Shouldn't this be ${version}?

minghui-liu: Yes. Thanks for catching this. Fixed.

pbspro.spec
```diff
@@ -308,6 +308,7 @@ cd build
 %make_install
 mandir=$(find %{buildroot} -type d -name man)
 [ -d "$mandir" ] && find $mandir -type f -exec gzip -9 -n {} \;
+install -D %{buildroot}/%{pbs_prefix}/libexec/pbs_init.d %{buildroot}/etc/init.d/pbs
```
vstumpf: The client rpm does not have pbs_init.d, and will show a warning when it hits this line.

minghui-liu: I tested installing client package and there was not warning. The %install section happens during RPM build not during package install. And because`pbs_init.d` is excluded from the client package, RPM will not install it.

PR #1131, opened 2019-05-16
Default setting for preempt_sort needed in PTL scheduler class

saritakh:
#### Describe Bug or Feature

Default setting for preempt_sort needed in PTL scheduler class:
With update of scheduler configuration parameter 'preempt_sort' from sched_config file to 'qmgr' interface; it needed to be set to default value in PTL; which otherwise was getting unset i.e. removed during cleanup.

#### Describe Your Change

Added ATTR_preempt_sort and set its default value in dflt_attributes dic of class Scheduler in order to set its default value
Added 'preempt_sort' to ignore_attributes since it was getting unset - removed at the cleanup time

[Tests_after_fix.txt](https://github.com/PBSPro/pbspro/files/3185771/Tests_after_fix.txt)
[Tests_before_fix.txt](https://github.com/PBSPro/pbspro/files/3185772/Tests_before_fix.txt)


saritakh:
@agrawalravi90 Sorry I had goofed up while singing and squashing my commits. Thanks for looking into this; I have now updated the exact changes. Please review.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -10483,6 +10483,9 @@ class Scheduler(PBSService):
         "log_filter": "3328 ",
 
     }
+    dflt_attributes = {
+        ATTR_preempt_sort: "min_time_since_start"
```
agrawalravi90: PTL should not be doing this. If preempt_sort is supposed to have a default value, then PBS should be the one re-setting it (after an 'unset' operation). PTL should just be doing an unset. So, I think PTL was doing the correct thing, we need to fix PBS.

saritakh: In PBS, the sched_config behavior of preempt_sort is retained as is moved to qmgr. In sched_config, either the value of preempt_sort was set (to min_time_since_start); or the parameter had to be removed to get the reverse behavior. But, that makes it inconsistent when compared to the other moved sched parameters by not having a default behavior when unset.

arungrover: The reverse behavior, when the attribute is unset, is an undocumented behavior where we sort the jobs based on their priority. Unless we decide to make that as a documented behavior, PTL is the only place you can revert it back. This is all in line with the way sched_config worked.
From your comment, it seems like that you are proposing to document the undocumented behavior, is that correct?

agrawalravi90: Yes i am. qmgr unset for default attributes should 'reset' them to their defaults. This doesn't happen for some attributes (like node_fail_requeue) because of a bug in PBS, not by design, and even then the code still uses the default value of node_fail_requeue (310) internally. So, I feel that it's very confusing to say that 'min_time_since_start' is the default value for preempt_sort, but when unset, we not only won't reset it to its default, we will actually exhibit a completely different, undocumented behavior.

bhroam: The unset version will sort first by preemption priority and then submission order.  Likely all running jobs will be in the same preemption priority, so we'll be preempting them in submission order.  This will likely mean we'll be choosing jobs that have been running the longest.  Do we need preempt_sort?  Does anyone really want to preempt jobs in the order they were submitted in?  The idea behind preempt_sort is that we'd add more sorting methods.  If we haven't in the decade or so since we added preempt_sort.  Is it required?  Should we remove it and  just sort by min time since start?  If we ever want to add a new sort, we can bring back preempt_sort.  Just incase someone is wants the unset behavior, we probably should check with the field first.

arungrover: I'm fine if all of us think that preempt_sort is not needed as an attribute for time being. I'd suggest opening a forum discussion to let the community know that we will be providing one type of sort and there is no way to change it.

agrawalravi90: @bhroam can you please send an email out to the field before we start a forum discussion?

saritakh: @all, I have queried the field team for preempt_sort's usage.

agrawalravi90: Thanks @saritakh Please let us know what feedback you get.

agrawalravi90: @saritakh any updates on this ?

saritakh: Hi @agrawalravi90 and All,

Sorry for the late reply on this. As I understand from the feedback, there are lot of enhancement expectations regarding preempt_sort from the customers. From the feedback I got, only one customer used the non-default behavior with actually not much impact of its consequences.

agrawalravi90: Thanks Sarita. It seems like there is demand for a configurable preemption sort, especially one that sorts by priority, which actually is supported, it's just that it's the "internal default" behavior which makes it less visible. So, I still feel that we should just make the "internal default" behavior a new preempt_sort option, decide which of the two methods should be the default preempt_sort method, and use that when one unsets preempt_sort.

agrawalravi90: for this PR, if the PTL change is urgent, then I'm ok with doing what you had implemented. But we should really try to enhance PBS and fix this if it's not urgent. So, what do you wanna do?

saritakh: I too feel that we need to update PBS instead of PTL. I have raised an issue for the same: https://github.com/PBSPro/pbspro/issues/1185. Hence, closing this PR.
Hence, closing this PR for now.

PR #1139, opened 2019-05-21
Adding scenario for multinode job array with advance reservation

Mahalty:
#### Description

- Automate following scenario - "multinode job array with advance reservation"

[execution_logs.txt](https://github.com/PBSPro/pbspro/files/3202408/execution_logs.txt)

### DIFF
--- a/test/tests/functional/pbs_reservations.py
+++ b/test/tests/functional/pbs_reservations.py
@@ -1003,3 +1003,87 @@ class TestReservations(TestFunctional):
                            extend='x', attrop=PTL_AND, id=jid2)
         self.server.expect(JOB, {'job_state': 'F', 'Exit_status': '0'},
                            extend='x', attrop=PTL_AND, id=jid3)
+
+    def test_advance_resv_with_multinode_job_array(self):
+        """
+        Test multinode job array with advance reservation
+        """
+        if (len(self.moms) < 2):
+            self.skip_test("Test requires 2 moms: use -p mom1:mom2")
+        a = {'resources_available.ncpus': 4}
+        for mom in self.moms.values():
+            self.server.manager(MGR_CMD_SET, NODE, a, id=mom.shortname)
+        self.server.manager(MGR_CMD_SET, SERVER,
+                            {'job_history_enable': 'True'})
+        # Submit reservation with placement type scatter
+        now = int(time.time())
+        a = {'Resource_List.select': '2:ncpus=2',
+             'Resource_List.place': 'scatter',
+             'reserve_start': now + 30,
+             'reserve_end': now + 300}
+        r = Reservation(PBSROOT_USER, attrs=a)
+        rid = self.server.submit(r)
+        exp_attr = {'reserve_state': (MATCH_RE, "RESV_CONFIRMED|2")}
+        self.server.expect(RESV, exp_attr, id=rid)
+        resv_queue = rid.split(".")[0]
+
+        # Submit job array in reservation queue
+        attrs = {ATTR_q: resv_queue, ATTR_J: '1-5',
+                 'Resource_List.select': '2:ncpus=1'}
+        j = Job(PBSROOT_USER, attrs)
+        j.set_sleep_time(60)
+        jid = self.server.submit(j)
+        subjid = []
+        for i in range(1, 6):
+            subjid.append(j.create_subjob_id(jid, i))
+
+        self.logger.info("Wait 30s for resv to be in Running state")
+        a = {'reserve_state': (MATCH_RE, 'RESV_RUNNING|5')}
+        self.server.expect(RESV, a, id=rid, offset=30)
+        self.server.expect(JOB, {'job_state': 'B'}, id=jid)
+        self.server.expect(JOB, {'job_state=R': 2}, count=True,
+                           extend='t', id=jid)
+        self.server.expect(JOB, {'job_state=Q': 3}, count=True,
+                           extend='t', id=jid)
+        self.server.sigjob(jobid=subjid[0], signal="suspend")
+        self.server.expect(JOB, {'job_state': 'S'}, id=subjid[0])
+        self.server.expect(JOB, {'job_state': 'R'}, id=subjid[2])
+
+        # Submit job array with placement type scatter in resv queue
+        attrs = {ATTR_q: resv_queue, ATTR_J: '1-5',
+                 'Resource_List.place': 'scatter',
+                 'Resource_List.select': '2:ncpus=1'}
+        j1 = Job(PBSROOT_USER, attrs)
+        j1.set_sleep_time(60)
+        jid2 = self.server.submit(j1)
+        subjid2 = []
+        for i in range(1, 6):
+            subjid2.append(j.create_subjob_id(jid2, i))
+        self.server.expect(JOB, {'job_state': 'Q'}, id=jid2)
+
+        self.server.sigjob(subjid[0], 'resume')
+        self.logger.info("Wait 180s for all the subjobs to complete")
+        self.server.expect(JOB, {'job_state': 'F', 'exit_status': '0'},
+                           id=jid, extend='x', offset=180)
+
+        self.server.expect(JOB, {'job_state': 'B'}, id=jid2)
+        self.server.expect(JOB, {'job_state=R': 2}, count=True,
+                           extend='t', id=jid2)
+        self.server.expect(JOB, {'job_state=Q': 3}, count=True,
+                           extend='t', id=jid2)
+        self.server.sigjob(jobid=subjid2[0], signal="suspend")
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_reservations.py
```diff
@@ -883,3 +883,80 @@ def test_fail_confirm_resv_message(self):
         # The scheduler should log reason why it was denied
         self.scheduler.log_match(rid + ";PBS Failed to confirm resv: " +
                                  "Insufficient amount of resource: ncpus")
+
+    def test_advance_resv_with_multinode_job_array(self):
+        """
+        Test multinode job array with advance reservation
+        """
+        if (len(self.moms) < 2):
+            self.skip_test("Test requires 2 moms as input")
```
bhroam: Say "Test requires 2 moms: use -p mom1:mom2:

test/tests/functional/pbs_reservations.py
```diff
@@ -883,3 +883,80 @@ def test_fail_confirm_resv_message(self):
...
+
+        a = {'reserve_state': (MATCH_RE, 'RESV_RUNNING|5')}
+        self.server.expect(RESV, a, id=rid, offset=30)
+        self.server.expect(JOB, {'job_state': 'B'}, id=jid)
+        self.server.expect(JOB, {'job_state=R': 2}, count=True,
+                           extend='t', id=jid)
+        self.server.expect(JOB, {'job_state=Q': 3}, count=True,
+                           extend='t', id=jid)
+        self.server.expect(JOB, {'job_state': 'Q'}, id=subjid[3])
+        self.server.deljob(subjid[3])
```
bhroam: I think self.server.delete() is the right PTL interface to use. (and other places in the test plan)

test/tests/functional/pbs_reservations.py
```diff
@@ -883,3 +883,80 @@ def test_fail_confirm_resv_message(self):
...
+        self.server.expect(JOB, {'job_state': 'X'}, id=subjid[3])
+        self.server.sigjob(jobid=subjid[0], signal="suspend")
+        self.server.expect(JOB, {'job_state': 'S'}, id=subjid[0])
+        self.server.expect(JOB, {'job_state': 'R'}, id=subjid[2])
+        attrs = {ATTR_q: resv_queue, ATTR_J: '1-5',
+                 'Resource_List.place': 'scatter',
+                 'Resource_List.select': '2:ncpus=1'}
+        j1 = Job(PBSROOT_USER)
+        j1.set_sleep_time(60)
+        j1.set_attributes(attrs)
```
bhroam: Consider passing attrs to Job().  it'll save you the call to set_attributes().

Mahalty:
Latest execution-logs:
[Latest_execution_logs.txt](https://github.com/PBSPro/pbspro/files/3206049/Latest_execution_logs.txt)

test/tests/functional/pbs_reservations.py
```diff
@@ -883,3 +883,80 @@ def test_fail_confirm_resv_message(self):
...
+        self.server.expect(JOB, {'job_state': 'B'}, id=jid2)
+        self.server.expect(JOB, {'job_state=R': 2}, count=True,
+                           extend='t', id=jid2)
+        self.server.expect(JOB, {'job_state=Q': 3}, count=True,
+                           extend='t', id=jid2)
+        self.server.sigjob(jobid=subjid2[0], signal="suspend")
+        self.server.expect(JOB, {'job_state': 'S'}, id=subjid2[0])
+        self.server.sigjob(jobid=subjid2[1], signal="suspend")
+        self.server.expect(JOB, {'job_state': 'S'}, id=subjid2[1])
+        self.server.deljob(subjid2[1])
```
Klovely: Remove this line .
Instead of deleting  subjid2[1] , check subjid2[2] and subjid[3] started running.

test/tests/functional/pbs_reservations.py
Klovely: As we discussed, even if we delete this line, next job-array would be in Q and no effect in test case objective.
So I would suggest remove this line.

Klovely: Generic comment  could you please add doc string to increase readability of test steps.

Mahalty:
Latest execution-logs:
[Latest_execution_logs_1.txt](https://github.com/PBSPro/pbspro/files/3207378/Latest_execution_logs_1.txt)

test/tests/functional/pbs_reservations.py
```diff
@@ -914,46 +916,60 @@ def test_advance_resv_with_multinode_job_array(self):
         subjid = []
         for i in range(1, 6):
             subjid.append(j.create_subjob_id(jid, i))
-
+        # Wait for reservation to be in running state
```
bhroam: Turn this comment into a self.logger.info().  It is nicer for the test runner to know why they are waiting 30s.  It also serves the same purpose as the comment.

test/tests/functional/pbs_reservations.py
```diff
@@ -914,46 +916,60 @@ def test_advance_resv_with_multinode_job_array(self):
...
         self.server.expect(JOB, {'job_state': 'S'}, id=subjid2[0])
         self.server.sigjob(jobid=subjid2[1], signal="suspend")
         self.server.expect(JOB, {'job_state': 'S'}, id=subjid2[1])
-        self.server.delete(subjid2[1])
-        self.server.expect(JOB, {'job_state': 'X'}, id=subjid2[1])
+        # Verify third and fourth subjobs are running
         self.server.expect(JOB, {'job_state': 'R'}, id=subjid2[2])
+        self.server.expect(JOB, {'job_state': 'R'}, id=subjid2[3])
+        # Delete running jobs
+        self.server.delete(subjid2[2], subjid2[3])
```
bhroam: You need to pass the jobs in as a list.  This passes them in as two different arguments.  The second argument to delete() is the extend field.  So make this self.server.delete([subjob[2], subjob[3]])

test/tests/functional/pbs_reservations.py
```diff
@@ -914,46 +916,60 @@ def test_advance_resv_with_multinode_job_array(self):
...
+        # Verify fifth subjob is in R state
+        self.server.expect(JOB, {'job_state': 'R'}, id=subjid2[4])
+        # Resume the suspended jobs
         self.server.sigjob(subjid2[0], 'resume')
-        self.server.expect(JOB, {'job_state': 'F'}, id=jid2,
-                           extend='x', offset=120)
+        self.server.sigjob(subjid2[1], 'resume')
+        self.server.expect(JOB, {'job_state=R': 2}, count=True,
+                           extend='t', id=jid2)
```
bhroam: Resuming subjobs 1 and 2 will not resume them immediately.  It will change their substate from 43 to 45.  A job in substate 45 is looked at by the scheduler to resume it when there is space to do so.  If I understand properly,  there should be 3 running jobs and 2 suspended jobs right now.  What this does is wait for the 3 running jobs to end, and then the 2 suspended jobs resume.  Is this what you want?

bhroam:
Some of your new comments are explain what you are doing and not why.  The code nicely explains what you are doing.  Saying "See job array in the B state" and then immediately following it by an expect to see it in the B state doesn't really help.  Add a comment if you need to explain why you are doing it.  If a code reader sees a few comments that explain what the code is doing, they'll stop reading comments.  They'll just look at the code.  Consider rephrasing some of your comments or removing some.

Mahalty:
Latest execution-logs:
[Latest_execution_logs_2.txt](https://github.com/PBSPro/pbspro/files/3210783/Latest_execution_logs_2.txt)

Klovely:
Latest changes looks good to me .

test/tests/functional/pbs_reservations.py
```diff
@@ -1074,22 +1070,18 @@ def test_advance_resv_with_multinode_job_array(self):
...
         self.server.expect(JOB, {'job_state': 'R'}, id=subjid2[4])
-        # Resume the suspended jobs
+        self.server.expect(JOB, {'job_state': 'X'}, id=subjid2[4], offset=60)
         self.server.sigjob(subjid2[0], 'resume')
         self.server.sigjob(subjid2[1], 'resume')
         self.server.expect(JOB, {'job_state=R': 2}, count=True,
                            extend='t', id=jid2)
         self.server.expect(JOB, {'job_state': 'F'},
-                           id=jid, extend='x', offset=180)
+                           id=jid2, extend='x', offset=180)
```
bhroam: Consider adding a self.logger.info() explaining what the 120s sleep is for.  It'd be nicer to the test runner to know why they are waiting.  It isn't necessary though.

Mahalty:
Thanks @bhroam  for the review. I have addressed your review comment

@PBSPro/pbspro-maintainers I have squashed commits and rebased my branch. Please take this PR forward

PR #1141, opened 2019-05-22
MGR_USR and OPR_USR added in PTL setup

tpai26:
#### Describe Bug or Feature
The current PTL framework adds the user by which the PTL is run to the list of managers in the setup.
This change will add  MGR_USER and OPER_USER to the list of managers and operators in the PTL setup irrespective of the user by which PTL is run

[test_mgr_opr_log.txt](https://github.com/PBSPro/pbspro/files/3284332/test_mgr_opr_log.txt)
[snapshot_unit_test.txt](https://github.com/PBSPro/pbspro/files/3287145/snapshot_unit_test.txt)
[qstat_format_logs.txt](https://github.com/PBSPro/pbspro/files/3287194/qstat_format_logs.txt)

### DIFF
--- a/test/fw/ptl/utils/pbs_testsuite.py
+++ b/test/fw/ptl/utils/pbs_testsuite.py
@@ -107,6 +107,7 @@ OPER_USER = PbsUser('pbsoper', uid=4356, groups=[TSTGRP0, TSTGRP2, GRP_PBS,
 ADMIN_USER = PbsUser('pbsadmin', uid=4357, groups=[TSTGRP0, TSTGRP2, GRP_PBS,
                                                    GRP_AGT])
 PBSROOT_USER = PbsUser('pbsroot', uid=4371, groups=[TSTGRP0, TSTGRP2])
+
 ROOT_USER = PbsUser('root', uid=0, groups=[ROOT_GRP])
 
 PBS_USERS = (TEST_USER, TEST_USER1, TEST_USER2, TEST_USER3, TEST_USER4,
@@ -1286,10 +1287,10 @@ class PBSTestSuite(unittest.TestCase):
         except PbsManagerError as e:
             self.logger.error(e.msg)
         a = {ATTR_managers: (INCR, current_user + '@*,' +
-             MGR_USER.name + '@*')}
+             str(MGR_USER) + '@*')}
         server.manager(MGR_CMD_SET, SERVER, a, sudo=True)
 
-        a1 = {ATTR_operators: (INCR, OPER_USER.name + '@*')}
+        a1 = {ATTR_operators: (INCR, str(OPER_USER) + '@*')}
         server.manager(MGR_CMD_SET, SERVER, a1, sudo=True)
         if ((self.revert_to_defaults and self.server_revert_to_defaults) or
                 force):
--- a/test/fw/ptl/utils/pbs_testsuite.py
+++ b/test/fw/ptl/utils/pbs_testsuite.py
@@ -1281,10 +1281,16 @@ class PBSTestSuite(unittest.TestCase):
         try:
             # Unset managers list
             server.manager(MGR_CMD_UNSET, SERVER, 'managers', sudo=True)
+            # Unset operators list
+            server.manager(MGR_CMD_UNSET, SERVER, 'operators', sudo=True)
         except PbsManagerError as e:
             self.logger.error(e.msg)
-        a = {ATTR_managers: (INCR, current_user + '@*')}
+        a = {ATTR_managers: (INCR, current_user + '@*,' +
+             MGR_USER.name + '@*')}
         server.manager(MGR_CMD_SET, SERVER, a, sudo=True)
+
+        a1 = {ATTR_operators: (INCR, OPER_USER.name + '@*')}
+        server.manager(MGR_CMD_SET, SERVER, a1, sudo=True)
         if ((self.revert_to_defaults and self.server_revert_to_defaults) or
                 force):
             server.revert_to_defaults(reverthooks=self.revert_hooks,
--- a/test/tests/functional/pbs_qstat_formats.py
+++ b/test/tests/functional/pbs_qstat_formats.py
@@ -177,7 +177,7 @@ class TestQstatFormats(TestFunctional):
             #        "PBS_O_QUEUE":"workq",
             #    },
 
-            if ',' in val and key != 'managers':
+            if key == ATTR_v:
                 for v in val.split(','):
                     qstat_attrs.append(str(v).split('=')[0])
         return qstat_attrs
--- a/test/tests/functional/pbs_qstat_formats.py
+++ b/test/tests/functional/pbs_qstat_formats.py
@@ -177,7 +177,7 @@ class TestQstatFormats(TestFunctional):
             #        "PBS_O_QUEUE":"workq",
             #    },
 
-            if ',' in val:
+            if ',' in val and key != 'managers':
                 for v in val.split(','):
                     qstat_attrs.append(str(v).split('=')[0])
         return qstat_attrs
--- a/test/tests/selftest/pbs_managers_operators.py
+++ b/test/tests/selftest/pbs_managers_operators.py
@@ -73,7 +73,6 @@ class TestManagersOperators(TestSelf):
         """
         Check that default operator user is set on PTL setup
         """
-        svr_opr = self.server.status(SERVER, 'operators')[0]['operators'] \
-            .split(",")
-        opr_usr = {str(OPER_USER) + '@*'}
-        self.assertEqual(opr_usr, set(svr_opr))
+        svr_opr = str(self.server.status(SERVER, 'operators'))
+        opr_usr = str(OPER_USER) + '@*'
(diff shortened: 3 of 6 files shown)


test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -1281,10 +1281,16 @@ def revert_server(self, server, force=False):
         try:
             # Unset managers list
             server.manager(MGR_CMD_UNSET, SERVER, 'managers', sudo=True)
+            # Unset operators list
+            server.manager(MGR_CMD_UNSET, SERVER, 'operators', sudo=True)
         except PbsManagerError as e:
             self.logger.error(e.msg)
-        a = {ATTR_managers: (INCR, current_user + '@*')}
+        a = {ATTR_managers: (INCR, current_user + '@*,' +
+             MGR_USER.name + '@*')}
```
agrawalravi90: just use str(MGR_USER) instead of accessing its private variable 'name'

test/tests/functional/pbs_qstat_formats.py
```diff
@@ -177,7 +177,7 @@ def get_qstat_attribs(self, obj_type):
             #        "PBS_O_QUEUE":"workq",
             #    },
 
-            if ',' in val:
+            if ',' in val and key != 'managers':
```
agrawalravi90: what about 'operators'? Since this seems to be specific for the Variable_List attribute, I think you should just match that exactly (if key == ATTR_v) instead of matching a comma

test/tests/selftest/pbs_managers_operators_ext.py
```diff
@@ -38,7 +38,7 @@
 from tests.selftest import *
```
agrawalravi90: I know that you didn't create this file, but can you please rename it to remove the "ext" at the end ?

test/tests/selftest/pbs_managers_operators_ext.py
```diff
@@ -50,13 +50,27 @@ def test_managers_unset_setup(self):
         after test setUp run
```
agrawalravi90: please change docstring to say "except current user & MGR_USER"

test/tests/selftest/pbs_managers_operators_ext.py
```diff
@@ -50,13 +50,27 @@ def test_managers_unset_setup(self):
...
         """
         runas = ROOT_USER
+        manager_usr_str = str(MGR_USER) + '@*'
         current_usr = pwd.getpwuid(os.getuid())[0]
-        attr = {ATTR_managers: current_usr + '@*'}
-        self.server.expect(SERVER, attrib=attr)
+        current_usr_str = str(current_usr) + '@*'
+        svr_mgr = str(self.server.status(SERVER, 'managers'))
```
agrawalravi90: The code right now only checks that the MGR_USER and the current user are part of managers, it will pass even if managers contains other users in addition to MGR_USER & current user.

How about this instead:
mgr_users = [str(MGR_USER) + '@\*', str(current_usr) + '@\*']
svr_mgr = self.server.status(SERVER, 'managers')[0]['managers'].split(",")
self.assertEquals(set(svr_mgr), set(mgr_users))

test/tests/selftest/pbs_managers_operators_ext.py
```diff
@@ -50,13 +50,27 @@ def test_managers_unset_setup(self):
...
+
         mgr_user1 = str(TEST_USER)
         mgr_user2 = str(TEST_USER1)
         a = {ATTR_managers: (INCR, mgr_user1 + '@*,' + mgr_user2 + '@*')}
         self.server.manager(MGR_CMD_SET, SERVER, a, runas=runas)
         self.logger.info("Calling test setUp:")
         TestSelf.setUp(self)
-        self.server.expect(SERVER, attrib=attr)
+
+        svr_mgr = str(self.server.status(SERVER, 'managers'))
```
agrawalravi90: same suggestion here: get the managers value as a list, create a list of all expected managers users, and compare the two as sets

test/tests/selftest/pbs_managers_operators_ext.py
```diff
@@ -50,13 +50,27 @@ def test_managers_unset_setup(self):
...
+        self.assertIn(str(manager_usr_str), svr_mgr)
+        self.assertIn(str(current_usr_str), svr_mgr)
+
+    def test_default_oper(self):
+        """
+        Check that default operator user is set on PTL setup
+        """
+        svr_opr = str(self.server.status(SERVER, 'operators'))
+        self.assertIn(str(OPER_USER) + '@*', svr_opr)
```
agrawalravi90: Again, this doesn't check that there aren't other operator users in the 'operators' attribute. Get the value and match it exactly.

test/tests/selftest/pbs_managers_operators_ext.py
```diff
@@ -41,21 +41,22 @@
...
         manager_usr_str = str(MGR_USER) + '@*'
         current_usr = pwd.getpwuid(os.getuid())[0]
         current_usr_str = str(current_usr) + '@*'
-        svr_mgr = str(self.server.status(SERVER, 'managers'))
-        self.assertIn(str(manager_usr_str), svr_mgr)
-        self.assertIn(str(current_usr_str), svr_mgr)
+        mgr_users = [manager_usr_str, current_usr_str]
+        svr_mgr = self.server.status(SERVER, 'managers')[0]['managers']\
+            .split(",")
+        self.assertEqual(mgr_users, svr_mgr)
```
bhroam: set() these two just to make them order independent.

test/tests/selftest/pbs_managers_operators.py
```diff
@@ -38,25 +38,42 @@
...
         runas = ROOT_USER
+        manager_usr_str = str(MGR_USER) + '@*'
         current_usr = pwd.getpwuid(os.getuid())[0]
-        attr = {ATTR_managers: current_usr + '@*'}
-        self.server.expect(SERVER, attrib=attr)
+        current_usr_str = str(current_usr) + '@*'
+        mgr_users = {manager_usr_str, current_usr_str}
+        svr_mgr = self.server.status(SERVER, 'managers')[0]['managers']\
+            .split(",")
+        self.assertEqual(mgr_users, set(svr_mgr))
```
agrawalravi90: you are comparing a dict with a set ? I think you need to make them both sets.

test/tests/selftest/pbs_managers_operators.py
```diff
@@ -38,25 +38,42 @@
...
         mgr_user2 = str(TEST_USER1)
         a = {ATTR_managers: (INCR, mgr_user1 + '@*,' + mgr_user2 + '@*')}
         self.server.manager(MGR_CMD_SET, SERVER, a, runas=runas)
         self.logger.info("Calling test setUp:")
         TestSelf.setUp(self)
-        self.server.expect(SERVER, attrib=attr)
+
+        svr_mgr = self.server.status(SERVER, 'managers')[0]['managers'] \
+            .split(",")
+        self.assertEqual(mgr_users, set(svr_mgr))
```
agrawalravi90: Same comment here

test/tests/selftest/pbs_managers_operators.py
```diff
@@ -38,25 +38,42 @@
...
+
+    def test_default_oper(self):
+        """
+        Check that default operator user is set on PTL setup
+        """
+        svr_opr = self.server.status(SERVER, 'operators')[0]['operators'] \
+            .split(",")
+        opr_usr = {str(OPER_USER) + '@*'}
+        self.assertEqual(opr_usr, set(svr_opr))
```
agrawalravi90: It's a single user, so you don't need to wrap it in another data structure, just compare them as strings

test/tests/selftest/pbs_managers_operators.py
tpai26: both are sets. I initialized the set with  ->  set = {'str a', 'str b'}

test/tests/selftest/pbs_managers_operators.py
tpai26: the value of svr_opr is a list. The reason not to compare them as strings is, if svr_opr list contains more than one element the test should fail.

tpai26: Oh I got it, I will change both to strings

test/tests/selftest/pbs_managers_operators.py
agrawalravi90: ah, my bad, thanks for explaining

agrawalravi90:
@tpai26 one advice: when you are trying things out and creating intermediate commits, please don't push them to your fork, it makes reviewing incremental changes very difficult.

test/tests/functional/pbs_qstat_formats.py
```diff
@@ -177,7 +177,7 @@ def get_qstat_attribs(self, obj_type):
             #        "PBS_O_QUEUE":"workq",
             #    },
 
-            if ',' in val:
+            if key == ATTR_v:
```
agrawalravi90: can you please run this test as well and post logs for it?

PR #1143, opened 2019-05-23
refactoring sched for performance

arungrover:
#### Describe Bug or Feature
There were reports to preemption being slow when soft-limits are in place in 13.0 release.

#### Describe Your Change
This pull request does not really address soft-limits issue partly because the optimization we did a few months back to cull the list of preemptible jobs speed things up during preemption to a large extent.
In our testing, we did find a few places where scheduler was unnecessarily spending time in creating log messages which are not needed. In this pull we remove those messages and also changed a function to do lookup of a resource reservation by its index (whenever necessary)

We found that the test added in this pull request takes around 772 seconds to preempt and the changed code does the same in 613 seconds, giving us an improvement of almost 20%.
In the attached logs search for multiple hashes to find the relevant timing information ("##########") between the tests done on 'master' branch and this dev branch.
[test_out_master.txt](https://github.com/PBSPro/pbspro/files/3214378/test_out_master.txt)
[test_out.txt](https://github.com/PBSPro/pbspro/files/3214379/test_out.txt)

### DIFF
--- a/src/scheduler/resource_resv.c
+++ b/src/scheduler/resource_resv.c
@@ -543,8 +543,6 @@ find_resource_resv_by_indrank(resource_resv **resresv_arr, int rank, int index)
  * @param[in]	resresv_arr -	array of resource_resvs to search
  * @param[in]	name        -	name of resource_resv to find
  * @param[in]	start_time  -	the start time of the resource_resv
- * @param[in]	index	    -	index of resource_resv in all_resresv array
- *				It is ignored when negative
  *
  * @return	resource_resv *
  * @retval	resource_resv	: if found
@@ -552,14 +550,11 @@ find_resource_resv_by_indrank(resource_resv **resresv_arr, int rank, int index)
  *
  */
 resource_resv *
-find_resource_resv_by_time_index(resource_resv **resresv_arr, char *name, time_t start_time, int index)
+find_resource_resv_by_time(resource_resv **resresv_arr, char *name, time_t start_time)
 {
 	int i;
 	if (resresv_arr == NULL || name == NULL)
 		return NULL;
-
-	if ((index >= 0) && (resresv_arr[index]->start == start_time))
-		return resresv_arr[index];
 
 	for (i = 0; resresv_arr[i] != NULL;i++) {
 		if ((strcmp(resresv_arr[i]->name, name) == 0) && (resresv_arr[i]->start == start_time))
@@ -1351,8 +1346,8 @@ update_resresv_on_end(resource_resv *resresv, char *job_state)
 					next_occr_time = get_occurrence(resresv->resv->rrule,
 						resresv->resv->req_start, resresv->resv->timezone, 2);
 					if (next_occr_time >= 0) {
-						next_occr = find_resource_resv_by_time_index(resresv->server->resvs,
-							resresv->name, next_occr_time, -1);
+						next_occr = find_resource_resv_by_time(resresv->server->resvs,
+							resresv->name, next_occr_time);
 						if (next_occr != NULL) {
 							if (resv_queue->jobs != NULL) {
 								for (i = 0; resv_queue->jobs[i] != NULL; i++) {
--- a/src/scheduler/resource_resv.c
+++ b/src/scheduler/resource_resv.c
@@ -538,7 +538,7 @@ find_resource_resv_by_indrank(resource_resv **resresv_arr, int rank, int index)
 
 /**
  * @brief
- * 		find a resource_resv by name/index and start time
+ * 		find a resource_resv by name and start time
  *
  * @param[in]	resresv_arr -	array of resource_resvs to search
  * @param[in]	name        -	name of resource_resv to find
--- a/src/scheduler/resource_resv.c
+++ b/src/scheduler/resource_resv.c
@@ -538,11 +538,13 @@ find_resource_resv_by_indrank(resource_resv **resresv_arr, int rank, int index)
 
 /**
  * @brief
- * 		find a resource_resv by name and start time
+ * 		find a resource_resv by name/index and start time
  *
  * @param[in]	resresv_arr -	array of resource_resvs to search
  * @param[in]	name        -	name of resource_resv to find
  * @param[in]	start_time  -	the start time of the resource_resv
+ * @param[in]	index	    -	index of resource_resv in all_resresv array
+ *				It is ignored when negative
  *
  * @return	resource_resv *
  * @retval	resource_resv	: if found
@@ -550,11 +552,14 @@ find_resource_resv_by_indrank(resource_resv **resresv_arr, int rank, int index)
  *
  */
 resource_resv *
-find_resource_resv_by_time(resource_resv **resresv_arr, char *name, time_t start_time)
+find_resource_resv_by_time_index(resource_resv **resresv_arr, char *name, time_t start_time, int index)
 {
 	int i;
 	if (resresv_arr == NULL || name == NULL)
 		return NULL;
+
+	if ((index >= 0) && (resresv_arr[index]->start == start_time))
+		return resresv_arr[index];
 
 	for (i = 0; resresv_arr[i] != NULL;i++) {
 		if ((strcmp(resresv_arr[i]->name, name) == 0) && (resresv_arr[i]->start == start_time))
@@ -1346,8 +1351,8 @@ update_resresv_on_end(resource_resv *resresv, char *job_state)
 					next_occr_time = get_occurrence(resresv->resv->rrule,
 						resresv->resv->req_start, resresv->resv->timezone, 2);
 					if (next_occr_time >= 0) {
-						next_occr = find_resource_resv_by_time(resresv->server->resvs,
-							resresv->name, next_occr_time);
+						next_occr = find_resource_resv_by_time_index(resresv->server->resvs,
+							resresv->name, next_occr_time, -1);
 						if (next_occr != NULL) {
 							if (resv_queue->jobs != NULL) {
 								for (i = 0; resv_queue->jobs[i] != NULL; i++) {
--- a/src/scheduler/resource_resv.h
+++ b/src/scheduler/resource_resv.h
@@ -93,9 +93,9 @@ resource_resv *find_resource_resv(resource_resv **resresv_arr, char *name);
 resource_resv *find_resource_resv_by_indrank(resource_resv **resresv_arr, int rank, int index);
 
 /**
- *  find_resource_resv_by_time_index - find a resource_resv by name/index and start time
+ *  find_resource_resv_by_time - find a resource_resv by name/index and start time
  */
-resource_resv *find_resource_resv_by_time_index(resource_resv **resresv_arr, char *name, time_t start_time, int index);
+resource_resv *find_resource_resv_by_time(resource_resv **resresv_arr, char *name, time_t start_time);
 
 /*
  *      find_resource_req - find a resource_req from a resource_req list
--- a/src/scheduler/resource_resv.h
+++ b/src/scheduler/resource_resv.h
@@ -93,7 +93,7 @@ resource_resv *find_resource_resv(resource_resv **resresv_arr, char *name);
 resource_resv *find_resource_resv_by_indrank(resource_resv **resresv_arr, int rank, int index);
 
 /**
- *  find_resource_resv_by_time - find a resource_resv by name/index and start time
+ *  find_resource_resv_by_time - find a resource_resv by name and start time
  */
 resource_resv *find_resource_resv_by_time(resource_resv **resresv_arr, char *name, time_t start_time);
 
--- a/src/scheduler/resource_resv.h
+++ b/src/scheduler/resource_resv.h
@@ -93,9 +93,9 @@ resource_resv *find_resource_resv(resource_resv **resresv_arr, char *name);
 resource_resv *find_resource_resv_by_indrank(resource_resv **resresv_arr, int rank, int index);
 
 /**
- *  find_resource_resv_by_time - find a resource_resv by name and start time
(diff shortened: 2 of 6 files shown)


src/scheduler/resource_resv.c
```diff
@@ -538,24 +538,29 @@ find_resource_resv_by_indrank(resource_resv **resresv_arr, int rank, int index)
...
+ *				It is ignored when negative
  *
  * @return	resource_resv *
  * @retval	resource_resv	: if found
  * @retval	NULL	: if not found or on error
  *
  */
 resource_resv *
-find_resource_resv_by_time(resource_resv **resresv_arr, char *name, time_t start_time)
+find_resource_resv_by_time_index(resource_resv **resresv_arr, char *name, time_t start_time, int index)
```
agrawalravi90: I don't think you need to change the function's name, it focused on finding resresv by time, not how the list is searched. If you are going to mention index, then you should also mention name.

src/scheduler/resource_resv.c
```diff
@@ -538,24 +538,29 @@ find_resource_resv_by_indrank(resource_resv **resresv_arr, int rank, int index)
...
 {
 	int i;
 	if (resresv_arr == NULL || name == NULL)
```
agrawalravi90: i think you should move name's NULL check to after the index check below, because one should be able to call this function with name = NULL and a valid index.

src/scheduler/resource_resv.c
```diff
@@ -1346,8 +1351,8 @@ update_resresv_on_end(resource_resv *resresv, char *job_state)
 					next_occr_time = get_occurrence(resresv->resv->rrule,
 						resresv->resv->req_start, resresv->resv->timezone, 2);
 					if (next_occr_time >= 0) {
-						next_occr = find_resource_resv_by_time(resresv->server->resvs,
-							resresv->name, next_occr_time);
+						next_occr = find_resource_resv_by_time_index(resresv->server->resvs,
+							resresv->name, next_occr_time, -1);
```
agrawalravi90: here and other places, why not pass in index of resresv? actually, why not just remove 'name' argument and always search by index?

test/tests/performance/pbs_preemptperformance.py
```diff
@@ -480,6 +480,69 @@ def test_preemption_basic(self):
...
+             'resources_available.mem': '6400mb'}
+        self.server.create_vnodes('vn', a, 500, self.mom, usenatvnode=False)
+        p = "express_queue, normal_jobs, server_softlimits, queue_softlimits"
+        a = {'preempt_prio': p}
+        self.server.manager(MGR_CMD_SET, SCHED, a, runas=ROOT_USER)
+
+        a = {'max_run_res_soft.ncpus': "[u:" + str(TEST_USER)+"=1]"}
+        self.server.manager(MGR_CMD_SET, QUEUE, a, 'workq')
+        self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'False'})
+        # submit a bunch of jobs under projectrB
```
agrawalravi90: typo: project B

test/tests/performance/pbs_preemptperformance.py
```diff
@@ -480,6 +480,69 @@ def test_preemption_basic(self):
...
+        a = {ATTR_l + '.select=1:ncpus': 1, ATTR_project: "ProjectB"}
```
agrawalravi90: what's the significance of setting the job's project attribute?

arungrover: that will make all the jobs have the same project and land us in softlimits code

src/scheduler/resource_resv.c
arungrover: Your comment made me think about it again, I think I should just call a different function in simulation to get the resource_resv object (likely find_resource_resv_by_ind) and not touch this function at all. That is the only place I am passing the index and I think there also I'll retain this call just for standing reservations.

src/scheduler/resource_resv.c
arungrover: I'll revert this change

src/scheduler/resource_resv.c
arungrover: Since standing reservations can occur multiple times in the array, I didn't pass in any index.

arungrover:
@agrawalravi90 I've addressed your comments. I also did a rebase with master by mistake, please ignore that.

test/tests/performance/pbs_preemptperformance.py
agrawalravi90: Won't all jobs have the same project anyways (_pbs_project_default)? also, the limit is on the user, not the project, so how will setting the project affect limits?

src/scheduler/simulate.c
```diff
@@ -1284,9 +1284,16 @@ find_event_ptr(timed_event *ote, server_info *nsinfo)
...
-				oep->name, oep->start);
+			if (oep->is_resv)
+				event_ptr =
+					find_resource_resv_by_time(nsinfo->all_resresv,
+					oep->name, oep->start);
+			else
+				/* In case of jobs there can be only one occurance of job in
+				 * all_resresv list, so no need to search using start time of job
+				 */
+				event_ptr = find_resource_resv_by_indrank(nsinfo->all_resresv,
```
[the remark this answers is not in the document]
arungrover: search by rank is O(N) operation and search by index is O(1).
It was earlier search by rank but we changed it to introduce index.

arungrover:
@agrawalravi90, I have changed the PR title, I'll amend the commit message during rebase after I get sign-off.
About the project, you are right all the jobs will get the same project. I just didn't want to depend on that behavior and explicitly provided a project name.
If you look into update_server_on_end/run() you will see why a matching project can result into updating preempt priority when the limit is not even set for project. Same is true for other combinations as well  - 

```
if (sinfo->has_soft_limit || resresv->job->queue->has_soft_limit) {
    for (i = 0; sinfo->jobs[i] != NULL; i++) {
        if (sinfo->jobs[i]->job !=NULL) {
            if (!strcmp(resresv->user, sinfo->jobs[i]->user) ||
                !strcmp(resresv->group, sinfo->jobs[i]->group) ||
                !strcmp(resresv->project, sinfo->jobs[i]->project))
                set_preempt_prio(sinfo->jobs[i],
                sinfo->jobs[i]->job->queue, sinfo);
         }
     }
}
```

src/scheduler/simulate.c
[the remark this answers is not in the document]
agrawalravi90: I kind of feel that it'll be better (and more OO) if we have just 1 generic function that takes as arguments "name", "rank" and "index", it first searches by index if it's not negative, then by rank if it's not negative, and then, finally, by name. If there's a 'time' argument then it can also check for that. I'd personally prefer that over multiple functions, but I'll leave it to you to decide if you wanna refactor the code or not.

agrawalravi90:
Thanks for explaining Arun. In the code that you pasted, it seems like the condition will get fulfilled anyways because we are submitting the jobs as the same user, no?

arungrover:
Ahh... That's correct, it will hit the condition with a different user as well. I'll remove setting Project in the jobs.

src/scheduler/simulate.c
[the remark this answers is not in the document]
arungrover: I agree with you, maybe we can have a single generic function to find the resource resv.
I don't want to change that with this pull request though, it will expand the scope of this work.

arungrover:
@agrawalravi90 I've addressed your concern about project

src/scheduler/resource_resv.c
```diff
@@ -538,7 +538,7 @@ find_resource_resv_by_indrank(resource_resv **resresv_arr, int rank, int index)
 
 /**
  * @brief
- * 		find a resource_resv by name and start time
+ * 		find a resource_resv by name/index and start time
```
bhroam: This no longer does index.

src/scheduler/resource_resv.h
```diff
@@ -93,7 +93,7 @@ resource_resv *find_resource_resv(resource_resv **resresv_arr, char *name);
 resource_resv *find_resource_resv_by_indrank(resource_resv **resresv_arr, int rank, int index);
 
 /**
- *  find_resource_resv_by_time - find a resource_resv by name and start time
+ *  find_resource_resv_by_time - find a resource_resv by name/index and start time
```
bhroam: Same here, this no longer does index.

test/tests/performance/pbs_preemptperformance.py
```diff
@@ -480,6 +480,69 @@ def test_preemption_basic(self):
...
+        """
+        Measure the time scheduler takes to preempt when the high priority
+        job hits soft limits under a considerable amount of workload.
+        """
+        a = {'resources_available.ncpus': 4,
+             'resources_available.mem': '6400mb'}
+        self.server.create_vnodes('vn', a, 500, self.mom, usenatvnode=False)
+        p = "express_queue, normal_jobs, server_softlimits, queue_softlimits"
+        a = {'preempt_prio': p}
+        self.server.manager(MGR_CMD_SET, SCHED, a, runas=ROOT_USER)
```
bhroam: The user running the test is a manager.  You don't need to run this as ROOT_USER.

test/tests/performance/pbs_preemptperformance.py
```diff
@@ -480,6 +480,69 @@ def test_preemption_basic(self):
...
+        self.server.expect(JOB, {'job_state=R': 2000}, interval=10, offset=5,
+                           max_attempts=100)
+
+        qname = 'highp'
+        a = {'queue_type': 'execution', 'priority': '200',
+             'started': 'True', 'enabled': 'True'}
+        self.server.manager(MGR_CMD_CREATE, QUEUE, a, qname)
+
+        self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'False'})
+        a = {ATTR_l + '.select=1:ncpus': 1, ATTR_q: 'highp'}
```
bhroam: I'd say to either use the qname variable or just remove the qname variable.

test/tests/performance/pbs_preemptperformance.py
```diff
@@ -480,6 +480,69 @@ def test_preemption_basic(self):
...
+        jid = ""
```
bhroam: You don't need to set this.  It will be set in the for loop.

test/tests/performance/pbs_preemptperformance.py
```diff
@@ -480,6 +480,69 @@ def test_preemption_basic(self):
...
+        for _ in range(2000):
+            j = Job(TEST_USER2, attrs=a)
+            j.set_sleep_time(3000)
+            jid = self.server.submit(j)
```
bhroam: You don't use jid.  It'll just get reset the next time through the loop.  I also don't see it used anywhere either.

test/tests/performance/pbs_preemptperformance.py
```diff
@@ -480,6 +480,69 @@ def test_preemption_basic(self):
...
+        scycle = time.time()
+        self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'True'})
+        (_, str1) = self.scheduler.log_match("Starting Scheduling Cycle",
+                                             n='ALL', starttime=int(scycle),
+                                             max_attempts=1, interval=2)
```
bhroam: Don't do this yourself.  Use the cycles() function to get a cycle and look at the start and end.  Look at pbs_job_sort_formula.py for how to use it.  It doesn't look at the start and end times, but it'll show you how to use cycles().  There is a start and end member to the cycles object that is returned.

test/tests/performance/pbs_preemptperformance.py
arungrover: I use it outside the loop to get the start time of the last job that was submitted

arungrover:
@bhroam I've addressed your review comments. Please have a look again
[test_out.txt](https://github.com/PBSPro/pbspro/files/3242755/test_out.txt)

test/tests/performance/pbs_preemptperformance.py
```diff
@@ -513,24 +513,25 @@ def test_preemption_with_soft_limits(self):
...
             j.set_sleep_time(3000)
-            jid = self.server.submit(j)
+            ljid = self.server.submit(j)
         scycle = time.time()
         self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'True'})
-        (_, str1) = self.scheduler.log_match("Starting Scheduling Cycle",
-                                             n='ALL', starttime=int(scycle),
+
+        # Rollback 2000 job ids to get to the first high priority job id
+        fjid = str((int(ljid.split('.')[0]) - 2000)) + '.' + ljid.split('.')[1]
```
bhroam: Don't you submit the high priority job separately?  Why do you need to figure out which job is the high priority job?

arungrover: I submit 2000 high priority jobs in a loop, I have the last jobid in ljid variable, from that job id I compute which is the first high priority job id. Once I have both the job ids, I run log_match() for both jobs to compute the time it takes to run 2000 high priority job by preempting 2000 normal jobs.

agrawalravi90: might be more readable if you set fjid in the loop:
"if fjid is None:
    fjid = self.server.submit(j)
else:
    ljid = self.server.submit(j)"

arungrover: @agrawalravi90 I've addressed the code for readability like the way you suggested, Please have a look at the latest commit.

arungrover:
[test_out.txt](https://github.com/PBSPro/pbspro/files/3248891/test_out.txt)

arungrover:
@bhroam Please don't merge this yet, I'm working on more changes

bhroam:
Ooops, too late.  I just merged it.

PR #1144, opened 2019-05-25
Fixed issue where custom resources requesting -1/-2 didn't work

bhroam:
#### Describe Bug or Feature
The scheduler uses special values for unspecified and infinity values (-1/-2).  The values they used are actually valid values that can be requested by a job for a custom resource.   The default value for a resource is infinity (-2).  If a job requested -2, the scheduler would set the value to the resource. It would then think it failed to set it since it was still set to the default value.  The job would be considered invalid for the cycle and be ignored.

#### Describe Your Change
The values for UNSPECIFIED and SCHD_INFINITY are too widely used on too many data types to change them to some sort of hugely negative or positive value.  It would require changing everything from a 5 bit field to many ints to longs.  The resource data type is a double.  There is a special floating point value called HUGE_VAL.  This is usually used for errors from floating point math functions.  I used HUGE_VAL for SCHD_INFINITY_RES and -HUGE_VAL for UNSPECIFIED_RES.

The customer who reported this bug used the custom resources in the job_sort_formula, so I wrote a generic job_sort_formula having a job request -2.  The test would fail if the job was ignored since the order of the two jobs would not be correct.

[job_sort_formula.log](https://github.com/PBSPro/pbspro/files/3219014/job_sort_formula.log)

### DIFF
--- a/src/scheduler/constant.h
+++ b/src/scheduler/constant.h
@@ -41,6 +41,8 @@
 extern "C" {
 #endif
 
+#include <math.h>
+
 /* macro to turn a value from enum preempt into it's bit for the bitfield */
 #define PREEMPT_TO_BIT(X) (1 << (X) )
 
@@ -92,12 +94,21 @@ enum select_job_status
 
 #define INIT_ARR_SIZE 2048
 
+/* We need two sets of UNSPECIFIED/SCHD_INFINITY constants.  One for resources
+ * which can be negative, and one for positive integer values.  While we could
+ * use some numbers near -LONG_MAX, that would mean every integer used in the
+ * scheduler would need to be a long, when an int or smaller type is fine.
+ */ 
+
 /* Unspecified resource value */
-#define UNSPECIFIED -1
+#define UNSPECIFIED_RES -HUGE_VAL
 #define UNSPECIFIED_STR "UNSPECIFIED"
 /* infinity value for resources */
-#define SCHD_INFINITY -2
+#define SCHD_INFINITY_RES HUGE_VAL
 #define SCHD_INFINITY_STR "SCHD_INFINITY"
+
+#define UNSPECIFIED -1
+#define SCHD_INFINITY -2
 
 /* infinity walltime value for forever job. This is 5 years(=60 * 60 * 24 * 365 * 5 seconds) */
 #define JOB_INFINITY (60 * 60 * 24 * 365 * 5)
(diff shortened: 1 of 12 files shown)


test/tests/functional/pbs_job_sort_formula.py
```diff
@@ -0,0 +1,71 @@
...
+
+class TestJobSortFormula(TestFunctional):
+    """
+    Tests for the job_sort_formula
+    """
+
+    def test_job_sort_formula_negative_value(self):
+        """
+        Test to see that negative values in the job_sort_formula
+        correctly sort properly
```
agrawalravi90: "correctly sort properly" sounds a bit excessive, i think you should remove either correctly or properly.

test/tests/functional/pbs_job_sort_formula.py
```diff
@@ -0,0 +1,71 @@
...
+        a = {'resources_available.ncpus': 2}
+        self.server.manager(MGR_CMD_SET, NODE, a, self.server.shortname)
+        self.server.manager(MGR_CMD_CREATE, RSC, {'type': 'float'}, id='foo')
+
+        a = {'job_sort_formula': 'foo', 'scheduling': 'False'}
+        self.server.manager(MGR_CMD_SET, SERVER, a, runas=ROOT_USER)
+
+        j1 = Job(TEST_USER, attrs={'Resource_List.foo': -2})
+        jid1 = self.server.submit(j1)
+        j2 = Job(TEST_USER, attrs={'Resource_List.foo': 1})
```
agrawalravi90: -1 right?

src/scheduler/constant.h
```diff
@@ -92,13 +94,22 @@ enum select_job_status
...
 /* Unspecified resource value */
-#define UNSPECIFIED -1
+#define UNSPECIFIED_RES -HUGE_VAL
 #define UNSPECIFIED_STR "UNSPECIFIED"
 /* infinity value for resources */
-#define SCHD_INFINITY -2
+#define SCHD_INFINITY_RES HUGE_VAL
 #define SCHD_INFINITY_STR "SCHD_INFINITY"
 
+#define UNSPECIFIED -1
```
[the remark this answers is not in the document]
bhroam: I don't think so.  They'd be defined to the same thing.  I think until we need to have them be separate strings, we should just use the macros we have now.

test/tests/functional/pbs_job_sort_formula.py
bhroam: I meant 1 to be higher than -2, but I see your point.  I'll change it to -1.  The issue where the job was being ignored really had to do with -2, but testing them both is better.

bhroam:
@agrawalravi90 I addressed your comments.  Please re-review.
[job_sort_formula.log](https://github.com/PBSPro/pbspro/files/3230137/job_sort_formula.log)

PR #1148, opened 2019-05-30
Refactor cgroup orphan cleanup

mike0042:
#### Describe Bug or Feature

When a job ends the cgroup hook attempts to remove the cgroups it had previously created. In some cases the removal fails. The remaining cgroups are no longer associated with active jobs and are referred to as orphans. Further attempts are made to remove these orphan cgroups in the exechost_periodic handler, 

#### Describe Your Change

- Orphan cleanup was enhanced by adding timestamps to the cgroup_jobs file to distinguish between newly arriving jobs and actual orphans
- An offline file was added to reduce interaction with the server
- The Timeout class is used more consistently
- The value of memory.swappiness is now configurable
- Added online_nodes_min_delay configuration parameter to prevent a node being brought online too soon after it had been taken offline
- Added server_timeout as a configuration parameter
- New methods for taking nodes online and offline rather than embedding them in other methods
- Other minor fixes and enhancements 

#### Link to Design Doc

Bug fix, no design doc.

[cgroups-ptl-0531.log](https://github.com/PBSPro/pbspro/files/3241816/cgroups-ptl-0531.log)
No changes to C code, so no valgrind log attached

### DIFF
--- a/src/hooks/cgroups/pbs_cgroups.PY
+++ b/src/hooks/cgroups/pbs_cgroups.PY
@@ -928,7 +928,7 @@ class HookUtils(object):
                     result = False
                 try:
                     val = int(cgroup.cfg['cgroup']['memory']['swappiness'])
-                    cgroup.set_swappiness(value)
+                    cgroup.set_swappiness(val)
                 except Exception:
                     pbs.logmsg(pbs.EVENT_DEBUG2,
                                '%s: Failed to set swappiness' % caller_name())
@@ -1640,8 +1640,8 @@ class NodeUtils(object):
         # Calculate reserved memory
         reserved = 0
         if not ignore_reserved:
-            reserve_percent = self.cfg['cgroup']['memory']['reserve_percent']
-            reserved += int(total * (int(reserve_percent) / 100))
+            reserve_pct = int(self.cfg['cgroup']['memory']['reserve_percent'])
+            reserved += int(total * (reserve_pct / 100.0))
             reserve_amount = self.cfg['cgroup']['memory']['reserve_amount']
             reserved += size_as_int(reserve_amount)
         pbs.logmsg(pbs.EVENT_DEBUG4, 'reserved mem: %d' % reserved)
@@ -1702,8 +1702,8 @@ class NodeUtils(object):
         # Calculate reserved vmem
         reserved = 0
         if not ignore_reserved:
-            reserve_percent = self.cfg['cgroup']['memsw']['reserve_percent']
-            reserved += int(total * (int(reserve_percent) / 100))
+            reserve_pct = int(self.cfg['cgroup']['memsw']['reserve_percent'])
+            reserved += int(total * (reserve_pct / 100.0))
             reserve_amount = self.cfg['cgroup']['memsw']['reserve_amount']
             reserved += size_as_int(reserve_amount)
         pbs.logmsg(pbs.EVENT_DEBUG4, 'reserved vmem: %d' % reserved)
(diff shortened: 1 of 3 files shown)


src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -1955,6 +1923,114 @@ class NodeConfig(object):
...
+            pbs.logmsg(pbs.EVENT_DEBUG2,
+                       '%s: Offline file already exists, not overwriting' %
+                       caller_name())
+            return
+        try:
+            # Write a timestamp so that exechost_periodic can avoid
+            # cleaning up before this event has sent updates to server
+            with open(self.offline_file, 'w') as fd:
+                fd.write(str(time.time()))
+                fd.truncate()
```
vstumpf: Why truncate here? `w` will truncate the file if it already exists.

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -3266,24 +3345,46 @@ class CgroupUtils(object):
         else:
             sockets = available.keys()
             # If placement type is load_balanced, reorder the sockets
```
vstumpf: With the addition of job_balanced, maybe update this comment?

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -3866,35 +3983,43 @@ class CgroupUtils(object):
             else:
                 avail_resc = self.available_node_resources(node)
             assigned = self.assign_job(hostresc, avail_resc, node)
+            if attempt != 0:
```
vstumpf: Is there a reason why we're stopping here if it failed?

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -4467,13 +4602,12 @@ class CgroupUtils(object):
...
         try:
-            with open(self.cgroup_jobs_file, 'r+') as f:
-                joblist = f.readline().split()
-                jobset = set(joblist)
-                jobset.add(jobid)
-                f.seek(0)
-                f.write(' '.join(jobset))
-                f.truncate()
+            with open(self.cgroup_jobs_file, 'r+') as fd:
+                jobdict = eval(fd.read())
```
vstumpf: I'm not the biggest fan of `eval()`, have you thought of using `ast.literal_eval`? See https://docs.python.org/2.7/library/ast.html#ast.literal_eval

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -4519,13 +4655,14 @@ class CgroupUtils(object):
...
+        Remove all keys from the file where local jobs are maintained
         """
         pbs.logmsg(pbs.EVENT_DEBUG4, 'Emptying file: %s' %
                    self.cgroup_jobs_file)
         try:
-            with open(self.cgroup_jobs_file, 'w') as f:
-                f.truncate()
+            with open(self.cgroup_jobs_file, 'w') as fd:
+                fd.write(str(dict()))
+                fd.truncate()
```
vstumpf: Don't think you need truncate() here.

src/hooks/cgroups/pbs_cgroups.PY
mike0042: I had tested with 'r+' manually, which requires truncate(). You are correct, it's not necessary with 'w'.

src/hooks/cgroups/pbs_cgroups.PY
mike0042: I added a comment to the code.

src/hooks/cgroups/pbs_cgroups.PY
mike0042: The description of ast.literal_eval() mentions untrusted sources. In this case, the hook itself is not an untrusted source since it populated the file. I'd prefer to leave it as-is.

PR #1149, opened 2019-05-31
Fix for intermittent failure of test_cgroup_prefix_and_devices

vccardenas:
#### Describe Bug or Feature
TestCgroupsHook.test_cgroup_prefix_and_devices would intermittently fail on some machines.

#### Describe Your Change
The test checks that only the devices subsystem is enabled with the correct allowable devices from within a job script.  The change was to move this check outside the job script, after a job is submitted and goes into R state.

test showing intermittent failures:
[TestCgroupsHook.test_cgroup_prefix_and_devices_x48_19.4_before.txt](https://github.com/PBSPro/pbspro/files/3242448/TestCgroupsHook.test_cgroup_prefix_and_devices_x48_19.4_before.txt)

test passes consistently:
[TestCgroupsHook.test_cgroup_prefix_and_devices_x48_19.4_after.txt](https://github.com/PBSPro/pbspro/files/3242294/TestCgroupsHook.test_cgroup_prefix_and_devices_x48_19.4_after.txt)
[TestCgroupsHook.test_cgroup_prefix_and_devices_x69_19.4_after.txt](https://github.com/PBSPro/pbspro/files/3242299/TestCgroupsHook.test_cgroup_prefix_and_devices_x69_19.4_after.txt)

### DIFF
--- a/test/tests/functional/pbs_cgroups_hook.py
+++ b/test/tests/functional/pbs_cgroups_hook.py
@@ -267,8 +267,8 @@ else
 fi
 sleep 10
 """
-        self.check_dirs_script = """#!/bin/bash
-#PBS -joe
+        self.check_dirs_script = """
+PBS_JOBID=$(qstat | tail -1 | awk '{print $1}')
 
 check_file_diff() {
     for filename in $1/*.*; do
@@ -283,7 +283,8 @@ check_file_diff() {
 }
 
 jobnum=${PBS_JOBID%%.*}
-cpuset_base=`grep cgroup /proc/mounts | grep cpuset | cut -d' ' -f2`
+cpuset_base=`grep cgroup /proc/mounts | grep cpuset | cut -d' ' -f2 | \
+             tr " " "\n" | sed -n '1p'`
 if [ -d "$cpuset_base/propbs" ]; then
     cpuset_job="$cpuset_base/propbs/$PBS_JOBID"
 else
@@ -334,7 +335,6 @@ if [ -d $devices_job ]; then
 else
     echo "Devices directory should be populated"
 fi
-sleep 10
 """
         self.check_gpu_script = """#!/bin/bash
 #PBS -joe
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -283,7 +283,8 @@ def setUp(self):
 }
 
 jobnum=${PBS_JOBID%%.*}
-cpuset_base=`grep cgroup /proc/mounts | grep cpuset | cut -d' ' -f2`
+cpuset_base=`grep cgroup /proc/mounts | grep cpuset | cut -d' ' -f2 | \
+             tr " " "\n" | sed -n '1p'`
```
mike0042: What is tr accomplishing? You are guaranteed there will be no spaces by the fact you're using space as a delimiter in the cut command. If you just want the first mount point, why not do this...
```
cpuset_base=`grep cgroup /proc/mounts | grep cpuset | cut -d' ' -f2 | head -1`
```

vccardenas: I wanted to be consistent with what was previously done in cpuset_mem_script (line 213).  I have now changed both cpuset_base assignments in lines 213 and 286.

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -211,7 +211,7 @@ def setUp(self):
 #PBS -joe
 echo $PBS_JOBID
 cpuset_base=`grep cgroup /proc/mounts | grep cpuset | cut -d' ' -f2 | \
-             tr " " "\n" | sed -n '1p'`
+             grep cgroup`
```
mike0042: Isn't the last command supposed to be "head -1" in case there were multiple mount points? You already grep'ed for cgroup.

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -284,7 +284,7 @@ def setUp(self):
 
 jobnum=${PBS_JOBID%%.*}
 cpuset_base=`grep cgroup /proc/mounts | grep cpuset | cut -d' ' -f2 | \
-             tr " " "\n" | sed -n '1p'`
+             grep cgroup`
```
mike0042: Isn't the last command supposed to be "head -1" in case there were multiple mount points? You already grep'ed for cgroup.

vccardenas: I have changed that to use "sort | tail -1" instead.

PR #1150, opened 2019-05-31
Flag values that are truncated in qstat

vstumpf:
#### Describe Bug or Feature
Qstat truncates values if it is longer than the column width, leading to some confusion.

#### Describe Your Change
If qstat will truncate a value, substitute the last character with an asterisk `*`.

#### Link to Design Doc
https://pbspro.atlassian.net/wiki/spaces/PD/pages/1289420812/Flag+fields+where+data+is+truncated+in+qstat

[after-fix.txt](https://github.com/PBSPro/pbspro/files/3242478/after-fix.txt)

### DIFF
--- a/src/cmds/qstat.c
+++ b/src/cmds/qstat.c
@@ -102,6 +102,8 @@ char *cnvt_est_start_time(char *start_time, int shortform);
 #define HPCBP_EXEC_TAG  "jsdl-hpcpa:Executable"
 #define CHAR_LINE_LIMIT 78 /* maximum number of characters which can be printed on a line */
 
+#define DISPLAY_TRUNC_CHAR '*'
+
 static struct attrl basic_attribs[] = {
 	{	&basic_attribs[1],
 		ATTR_N,
@@ -302,6 +304,8 @@ states(char *string, char *q, char *r, char *h, char *w, char *t, char *e, int l
 			*c = '\0';
 			if (strlen(s) > (size_t)len) {
 				f = s + len;
+				if (f > s)
+					*(f - 1) = DISPLAY_TRUNC_CHAR;
 				*f = '\0';
 			}
 			strcpy(d, s);
@@ -537,13 +541,50 @@ prt_attr(char *name, char *resource, char *value, int one_line) {
 #define LOCL    16  /* printf of jobs */
 #define SIZEL	6   /* length of "SIZE" fields in printf */
 
+
+/* format widths defined for display in normal format output */
+#define SIZEJOBID 15    /* length of JobId field in printf */
+#define SIZEJOBID_INCR 20 /*length of jobId field in printf with incr_width */
+#define SIZEJOBNAME 10  /* length of JobName field in printf */
+#define SIZEQUEUENAME 8 /* length of Queue field in printf */
+#define SIZESESSID 6    /* length of session id in printf */
+#define SIZENDS 3	/* length of nds in printf */
+#define SIZETSK 3	/* length of tsk in printf */
+#define SIZEUSER 8	/* length of user name in printf */
+
 /* format widths defined for display in wide format output */
-#define SIZEJOBID 30    /* length of JobId field in printf */
-#define SIZEJOBNAME 15  /* length of JobName field in printf when wide is true */
-#define SIZESESSID 8    /* length of session id in printf */
-#define SIZENDS 4	/* length of nds in printf */
-#define SIZETSK 5	/* length of tsk in printf */
-#define SIZEUSER 15	/* length of user name in printf */
+#define SIZEJOBID_W 30    /* length of JobId field in printf */
+#define SIZEJOBNAME_W 15  /* length of JobName field in printf */
+#define SIZESESSID_W 8    /* length of session id in printf */
+#define SIZENDS_W 4	/* length of nds in printf */
+#define SIZETSK_W 5	/* length of tsk in printf */
+#define SIZEUSER_W 15	/* length of user name in printf */
+
+
+/**
+ * @brief
+ *	Check if value is too long, truncate and append DISPLAY_TRUNC_CHAR if needed
+ *
+ * @param[in] value - value that may be truncated
+ * @param[in] len - non-wide length that must be met
+ * @param[in] wide_len - wide length that must be met
+ * @param[in] wide - whether or not wide format is used
+ *
+ */
+
+static void
+trunc_value(char *value, int len, int wide_len, int wide)
+{
+	if (wide) {
+		if (strlen(value) > wide_len && wide_len > 0) {
+			*(value + wide_len - 1) = DISPLAY_TRUNC_CHAR;
+		}
+	} else {
+		if (strlen(value) > len && len > 0) {
+			*(value + len - 1) = DISPLAY_TRUNC_CHAR;
+		}
+	}
+}
 
 /**
  * @brief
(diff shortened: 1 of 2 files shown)


src/cmds/qstat.c
```diff
@@ -302,6 +304,7 @@ states(char *string, char *q, char *r, char *h, char *w, char *t, char *e, int l
 			*c = '\0';
 			if (strlen(s) > (size_t)len) {
 				f = s + len;
+				*(f - 1) = DISPLAY_TRUNC_FLAG;
```
mike0042: Perhaps a bit pedantic, but please check (f > s) before you assign *(f - 1).

src/cmds/qstat.c
```diff
@@ -545,6 +548,32 @@ prt_attr(char *name, char *resource, char *value, int one_line) {
...
+ *
+ * @param[in] value - value that may be truncated
+ * @param[in] len - non-wide length that must be met
+ * @param[in] wide_len - wide length that must be met
+ * @param[in] wide - whether or not wide format is used
+ *
+ */
+
+static void
+trunc_value(char *value, int len, int wide_len, int wide)
```
mike0042: C99 allows you to use bool for the wide parameter. Why do you need two length parameters if you'll only ever use one or the other? Can we remove wide_len?

src/cmds/qstat.c
```diff
@@ -545,6 +548,32 @@ prt_attr(char *name, char *resource, char *value, int one_line) {
...
+{
+	if (wide) {
+		if (strlen(value) > wide_len) {
+			*(value + wide_len - 1) = DISPLAY_TRUNC_FLAG;
+		}
+	} else {
+		if (strlen(value) > len) {
+			*(value + len - 1) = DISPLAY_TRUNC_FLAG;
```
mike0042: You should ensure (len > 0) before you make this assignment.

src/cmds/qstat.c
```diff
@@ -924,10 +961,14 @@ altdsp_statjob(struct batch_status *pstat, struct batch_status *prtheader, int a
 		else
 			timeval = usecput ? eltimecpu : eltimewal;
 
-
 		if (wide) {
+			if (strlen(pstat->name) > SIZEJOBID) {
```
mike0042: Should this be jname instead of pstat->name?

src/cmds/qstat.c
```diff
@@ -942,11 +983,21 @@ altdsp_statjob(struct batch_status *pstat, struct batch_status *prtheader, int a
 				timeval);
 		} else {
 			if(how_opt & ALT_DISPLAY_INCR_WIDTH) {
```
mike0042: Please add a space after the if.

src/cmds/qstat.c
```diff
@@ -942,11 +983,21 @@ altdsp_statjob(struct batch_status *pstat, struct batch_status *prtheader, int a
...
+				if (strlen(pstat->name) > 20) {
```
mike0042: Should this be jname instead of pstat->name?

src/cmds/qstat.c
```diff
@@ -1362,16 +1413,19 @@ display_statjob(struct batch_status *status, struct batch_status *prtheader, int
 				if (wide) {
 					if (l > SIZEJOBID) {
 						c = p->name + SIZEJOBID;
+						*(c - 1) = DISPLAY_TRUNC_FLAG;
```
mike0042: This should be safe since we know SIZEJOBID > 0.

src/cmds/qstat.c
```diff
@@ -1362,16 +1413,19 @@ display_statjob(struct batch_status *status, struct batch_status *prtheader, int
...
 						*c = '\0';
 					}
 				} else if (how_opt & ALT_DISPLAY_INCR_WIDTH) {
 					if (l > (PBS_MAXSEQNUM+10)) {
 						c = p->name + PBS_MAXSEQNUM + 10;
+						*(c - 1) = DISPLAY_TRUNC_FLAG;
 						*c = '\0';
 					}
 				} else {
 					if (l > (PBS_MAXSEQNUM+5)) {
```
mike0042: Please add space around the plus sign.

src/cmds/qstat.c
vstumpf: jname hadn't been filled out yet.

src/cmds/qstat.c
vstumpf: If I change `trunc_value` to only take one length, then everywhere I call it I have to put `if (wide) ... else` blocks.
I kept wide as an int to follow the style of the file, but I can change it if you want.

src/cmds/qstat.c
vstumpf: Same as above, jname hasn't been filled yet.

src/cmds/qstat.c
mike0042: Ahh... thanks for explaining, that makes sense.

src/cmds/qstat.c
```diff
@@ -836,25 +867,32 @@ altdsp_statjob(struct batch_status *pstat, struct batch_status *prtheader, int a
 		while (pat) {
 			if (strcmp(pat->name, ATTR_N) == 0) {
 				jobn = pat->value;
+				trunc_value(jobn, 10, SIZEJOBNAME, wide);
```
mike0042: We should define all field widths as macros, as we do for the wide values.

src/cmds/qstat.c
```diff
@@ -102,6 +102,8 @@ char *cnvt_est_start_time(char *start_time, int shortform);
 #define HPCBP_EXEC_TAG  "jsdl-hpcpa:Executable"
 #define CHAR_LINE_LIMIT 78 /* maximum number of characters which can be printed on a line */
 
+#define DISPLAY_TRUNC_FLAG '*'
```
bhroam: This really isn't a flag.  A flag changes how something works.  This is just a character that indicates something.  Maybe DISPLAY_TRUNC_CHAR?

src/cmds/qstat.c
```diff
@@ -545,6 +549,32 @@ prt_attr(char *name, char *resource, char *value, int one_line) {
 #define SIZETSK 5	/* length of tsk in printf */
 #define SIZEUSER 15	/* length of user name in printf */
 
+
+/**
+ * @brief
+ *	Check if value is too long, truncate and append with * if needed
```
bhroam: Don't say '*'.  It's dependent on what your macro is set to.  I'd say 'and indicate it is truncated' or something like that.

src/cmds/qstat.c
```diff
@@ -924,10 +962,14 @@ altdsp_statjob(struct batch_status *pstat, struct batch_status *prtheader, int a
 		else
 			timeval = usecput ? eltimecpu : eltimewal;
 
-
 		if (wide) {
+			if (strlen(pstat->name) > SIZEJOBID) {
```
bhroam: Why aren't you using your function to handle the truncating?

test/tests/functional/pbs_qstat_formats.py
```diff
@@ -678,3 +678,40 @@ def test_qstat_answ(self):
...
         self.run_namelength_test('-answ')
+
+    def test_qstat_ans(self):
+        """
+        Test if qstat -ans correctly prints with truncation.
+        """
+        self.server.stop()
+        self.assertFalse(self.server.isUp(), 'Failed to stop PBS')
+
+        conf = self.du.parse_pbs_config(self.server.hostname)
```
bhroam: You don't need to parse pbs config.  Use self.server.pbs_conf[]

vstumpf:
I'm not sure why I have two travis checks, any ideas how to fix it?

mike0042:
You could try re-signing your commit to restart the build.

bhroam:
LGTM

bhroam:
Please squash and rebase once.  I'll take it from there.

PR #1151, opened 2019-06-04
database shutdown / server abort relating to duplicate queue creation

nithinj:
Cherry-pick from master: https://github.com/PBSPro/pbspro/pull/1076
#### Describe Bug or Feature
Three issues are being addressed here related to queues
1) database shutdown / server abort when trying to create the same queue second time which is having max queue name length .
que_alloc() will copy queue name to a string which can hold only 14 characters. During the next attempt, 15 char long queue name is compared with this string in find_queuebyname() and proceeds to insert in database as it couldn't find a match. Database will see this as a duplicate entry and will throw a primary key violation resulting in a panic_stop_db

2) If you create a queue, by default it wont be having any attributes with it. So it inserts empty string to the attribute field of pbs.queue table. However our database API's are not able to handle this empty string value and hence will not be able to load these values back to the memory after a restart. Hence `qstat -Qf` will not display this queue and re-creating this queue will cause server abort as it is unable to handle primary key violation exception in the db.

3) After an overlay upgrade all the queues are found missing if one of the queue does not have any attribute associated with it:
Overlay upgrade sql commands are trying to insert a NULL (as no attribute is present) and hence resulting in an error due to the NOT NULL constraint in the table defenition. Due to this error attribute fields of other entries are also filled with empty string rather than the actual value. After upgrade a fresh incarnation of the server wont be able to read these values due to bug mentioned above. 

#### Describe Your Change
1) Add one more byte to hold null character.
2) Update the database API's to read empty string
3) Following changes are made to the sql command to fix overlay upgrade issue:
Create attribute field without NOT NULL constraint -> Insert values including NULL -> convert all the NULL to empty string -> Introduce NOT NULL constraint for attribute field.

[logs_19_1_2.txt](https://github.com/PBSPro/pbspro/files/3252158/logs_19_1_2.txt)

CC: @subhasisb , @suresh-thelkar , @agrawalravi90

### DIFF
--- a/.travis/opensuse_leap_15.sh
+++ b/.travis/opensuse_leap_15.sh
@@ -1,9 +1,9 @@
 #!/bin/bash -xe
-OS_NAME=$(${DOCKER_EXEC} cat /etc/os-release | awk -F[=\"] '/^NAME=/ {print $3}')
-OS_VERSION=$(${DOCKER_EXEC} cat /etc/os-release | awk -F[=\"] '/^VERSION=/ {print $3}')
-OS_NAME_VERSION=${OS_NAME// /_}_${OS_VERSION// /_}
-${DOCKER_EXEC} zypper -n ar -f -G http://download.opensuse.org/repositories/devel:/tools/${OS_NAME_VERSION}/devel:tools.repo
-${DOCKER_EXEC} zypper -n ar -f -G http://download.opensuse.org/repositories/devel:/libraries:/c_c++/${OS_NAME_VERSION}/devel:libraries:c_c++.repo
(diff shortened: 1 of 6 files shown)


nithinj:
Thanks @suresh-thelkar for your sign-off.

@PBSPro/pbspro-maintainers ,
Please have a look.

agrawalravi90:
@subhasisb it seems like I don't have merge permission for 19.1 branch, can you please merge this PR and provide me merge permission?

prakashcv13:
@subhasisb - same goes for me, unable to merge.

PR #1155, opened 2019-06-06
Fixed preemption with multiple scheduler (and misc server crashes)

bhroam:
#### Describe Bug or Feature
When two or more schedulers preempt jobs at the same time, the server would crash.  This is due to the use of global variables.  They would be used intermixed between the schedulers, and then when one scheduler was replied to, set to NULL.  The next use of that variable would dereference it and crash.

#### Describe Your Change
The three global variables used by preemption are already on the preq.  One is the total number of jobs to preempt.  Next is the jobs that were preempted.  Last is how many jobs were preempted.  The global variables were used to keep track of preemption progress, and then stores on the preq for replying.  Now I just store the progress on the preq directly.

[multi_preemption.log](https://github.com/PBSPro/pbspro/files/3259574/multi_preemption.log)
[preemption.log](https://github.com/PBSPro/pbspro/files/3259575/preemption.log)

### DIFF
--- a/src/scheduler/job_info.c
+++ b/src/scheduler/job_info.c
@@ -2874,7 +2874,8 @@ find_and_preempt_jobs(status *policy, int pbs_sd, resource_resv *hjob, server_in
 		for (i = 0; i < no_of_jobs; i++) {
 			job = find_resource_resv(sinfo->running_jobs, preempt_jobs_reply[i].job_id);
 			if (job == NULL) {
-				snprintf(log_buffer, sizeof(log_buffer), "Server replied to preemption request with job %s which does not exist.", preempt_jobs_reply[i].job_id);
+				snprintf(log_buffer, sizeof(log_buffer), "Server replied to preemption request with job which does not exist.");
+				schdlog(PBSEVENT_DEBUG, PBS_EVENTCLASS_JOB, LOG_DEBUG, preempt_jobs_reply[i].job_id, log_buffer);
 				continue;
 			}
 
--- a/src/scheduler/job_info.c
+++ b/src/scheduler/job_info.c
@@ -2873,6 +2873,11 @@ find_and_preempt_jobs(status *policy, int pbs_sd, resource_resv *hjob, server_in
 
 		for (i = 0; i < no_of_jobs; i++) {
 			job = find_resource_resv(sinfo->running_jobs, preempt_jobs_reply[i].job_id);
+			if (job == NULL) {
+				snprintf(log_buffer, sizeof(log_buffer), "Server replied to preemption request with job %s which does not exist.", preempt_jobs_reply[i].job_id);
+				continue;
+			}
+
 			if (preempt_jobs_reply[i].order[0] == '0') {
 				done = 0;
 				fail_list[fail_count++] = job->rank;
--- a/src/server/req_holdjob.c
+++ b/src/server/req_holdjob.c
@@ -84,7 +84,7 @@ extern char	*msg_jobholdrel;
 extern char	*msg_mombadhold;
 extern char	*msg_postmomnojob;
 extern time_t	 time_now;
-extern job  *chk_job_request(char *, struct batch_request *, int *);
+extern job  *chk_job_request(char *, struct batch_request *, int *, int *);
 
 
 int chk_hold_priv(long val, int perm);
@@ -137,13 +137,17 @@ req_holdjob(struct batch_request *preq)
 	job		*pjob;
 	char		*pset;
 	int		 rc;
-	char             date[32];
-	time_t           now;
+	char 		 date[32];
+	time_t 		 now;
+	int err = PBSE_NONE;
 
-
-	pjob = chk_job_request(preq->rq_ind.rq_hold.rq_orig.rq_objname, preq, &jt);
-	if (pjob == NULL)
+	pjob = chk_job_request(preq->rq_ind.rq_hold.rq_orig.rq_objname, preq, &jt, &err);
+	if (pjob == NULL) {
+		pjob = find_job(preq->rq_ind.rq_hold.rq_orig.rq_objname);
+		if (pjob != NULL && pjob->ji_pmt_preq != NULL)
+			reply_preempt_jobs_request(err, PREEMPT_METHOD_CHECKPOINT, pjob);
 		return;
+	}
 	if ((jt != IS_ARRAY_NO) && (jt != IS_ARRAY_ArrayJob)) {
 		/*
 		 * We need to find the job again because chk_job_request() will return
@@ -275,7 +279,7 @@ req_releasejob(struct batch_request *preq)
 	int		 rc;
 
 
-	pjob = chk_job_request(preq->rq_ind.rq_release.rq_objname, preq, &jt);
+	pjob = chk_job_request(preq->rq_ind.rq_release.rq_objname, preq, &jt, NULL);
 	if (pjob == NULL)
 		return;
 
--- a/src/server/req_preemptjob.c
+++ b/src/server/req_preemptjob.c
@@ -86,14 +86,15 @@ const static char *preempt_methods[] = {
  * @param[in] job_id - the job to mark as failed
  * @return void
  */
-static void job_preempt_fail(struct batch_request *preempt_preq, char *job_id) {
+static void job_preempt_fail(struct batch_request *preempt_preq, char *job_id) 
+{
 	int preempt_index = preempt_preq->rq_reply.brp_un.brp_preempt_jobs.count;
-	preempt_job_info *preempt_jobs_list = preempt_jobs_list = preempt_preq->rq_reply.brp_un.brp_preempt_jobs.ppj_list;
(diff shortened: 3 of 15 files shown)


bhroam:
@arungrover @prakashcv13 this PR fixes several server crashes, including the one which is currently failing our nightly tests.

bhroam:
This PR fixes an infinite loop that can OOM machines(including our nightly test machines).  Please review this PR promptly.

src/server/req_preemptjob.c
```diff
@@ -71,23 +71,29 @@ extern struct server server;
...
-	log_event(PBSEVENT_DEBUG, PBS_EVENTCLASS_JOB, LOG_DEBUG, job_id, log_buffer);
+static void job_preempt_fail(struct batch_request *preempt_preq, char *job_id) {
+	int preempt_index = preempt_preq->rq_reply.brp_un.brp_preempt_jobs.count;
+	preempt_job_info *preempt_jobs_list = preempt_jobs_list = preempt_preq->rq_reply.brp_un.brp_preempt_jobs.ppj_list;
+
 	preempt_preq->rq_reply.brp_code = 1;
 	strcpy(preempt_jobs_list[preempt_index].order, "000");
 	sprintf(preempt_jobs_list[preempt_index].job_id, "%s", job_id);
-	preempt_index++;
+	preempt_preq->rq_reply.brp_un.brp_preempt_jobs.count = preempt_index + 1;
```
prakashcv13: why don't we simply preempt_preq->rq_reply.brp_un.brp_preempt_jobs.count++ instead of using a new variable?

src/server/req_preemptjob.c
```diff
@@ -71,23 +71,29 @@ extern struct server server;
...
  * @param[in] preempt_preq - the preemption preq from the scheduler
  * @param[in] job_id - the job to mark as failed
  * @return void
  */
-static void job_preempt_fail(struct batch_request *preempt_preq, char *job_id, int code) {
-	snprintf(log_buffer, sizeof(log_buffer), "preemption failed for job (%d)", code);
-	log_event(PBSEVENT_DEBUG, PBS_EVENTCLASS_JOB, LOG_DEBUG, job_id, log_buffer);
+static void job_preempt_fail(struct batch_request *preempt_preq, char *job_id) {
+	int preempt_index = preempt_preq->rq_reply.brp_un.brp_preempt_jobs.count;
+	preempt_job_info *preempt_jobs_list = preempt_jobs_list = preempt_preq->rq_reply.brp_un.brp_preempt_jobs.ppj_list;
```
prakashcv13: extra preempt_jobs_list = ?

src/server/req_preemptjob.c
```diff
@@ -288,23 +276,24 @@ req_preemptjobs(struct batch_request *preq)
 	job *pjob = NULL;
 	preempt_job_info *ppj = NULL;
 	pbs_sched *psched;
+	int preempt_index = 0;
+	int preempt_total;
+	preempt_job_info *preempt_jobs_list;
 
 	preq->rq_reply.brp_code = 0;
 	count = preq->rq_ind.rq_preempt.count;
 	psched = find_sched_from_sock(preq->rq_conn);
```
prakashcv13: should we handle the case when find_sched_from_sock() returns NULL?

src/server/req_preemptjob.c
```diff
@@ -377,14 +369,30 @@ reply_preempt_jobs_request(int code, int aux, struct job *pjob)
...
+				pjob->ji_pmt_preq = NULL;
+			}
+		} else {
+			job_preempt_fail(preq, pjob->ji_qs.ji_jobid);
 			pjob->ji_pmt_preq = NULL;
 		}
-		pjob->preempt_order_index++;
-	} else {
+	}
+	else {
```
prakashcv13: } and else should be on the same line.

src/server/req_signal.c
```diff
@@ -113,12 +113,17 @@ req_signaljob(struct batch_request *preq)
...
 	jid = preq->rq_ind.rq_signal.rq_jid;
 
-	parent = chk_job_request(jid, preq, &jt);
-	if (parent == NULL)
-		return;		/* note, req_reject already called */
+	parent = chk_job_request(jid, preq, &jt, &err);
+	if (parent == NULL) {
+		pjob = find_job(jid);
+		if (pjob != NULL && pjob->ji_pmt_preq != NULL)
+			reply_preempt_jobs_request(err, PREEMPT_METHOD_SUSPEND, pjob);
```
prakashcv13: instead of doing this for every caller, wouldn't it be good to move this to chk_job_request() itself?

src/server/req_signal.c
```diff
@@ -465,9 +470,11 @@ post_signal_req(struct work_task *pwt)
 			rel_resc(pjob);
 		}
 
+		if (pjob == NULL)
+			pjob = find_job(preq->rq_ind.rq_signal.rq_jid);
```
prakashcv13: we should check for the return value of find_job() like we do on line 528

src/scheduler/job_info.c
```diff
@@ -2873,6 +2873,11 @@ find_and_preempt_jobs(status *policy, int pbs_sd, resource_resv *hjob, server_in
 
 		for (i = 0; i < no_of_jobs; i++) {
 			job = find_resource_resv(sinfo->running_jobs, preempt_jobs_reply[i].job_id);
+			if (job == NULL) {
+				snprintf(log_buffer, sizeof(log_buffer), "Server replied to preemption request with job %s which does not exist.", preempt_jobs_reply[i].job_id);
```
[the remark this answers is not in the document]
bhroam: It is mostly a precaution.  If the server ever has a bug where it passes a job back to us that doesn't exist (like it's in a different partition), we don't want to crash.  I did see the crash once when the jobs for 2 partitions were being intermixed.

src/server/req_holdjob.c
```diff
@@ -275,7 +279,7 @@ req_releasejob(struct batch_request *preq)
 	int		 rc;
 
 
-	pjob = chk_job_request(preq->rq_ind.rq_release.rq_objname, preq, &jt);
+	pjob = chk_job_request(preq->rq_ind.rq_release.rq_objname, preq, &jt, NULL);
```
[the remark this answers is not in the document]
bhroam: All calls to chk_job_request() take this new argument.  The argument are for req_deletejob(), req_holdjob(), and req_signaljob().  It's so chk_job_request() can return the reason why the request was rejected.  I'm not sure what exactly you are asking.  Could you explain further?

src/server/req_signal.c
bhroam: Originally I thought chk_job_request() was just used for deletion.  It turns out it is called all over the place.  So my option was to pass in the preempt method, or to do it in the caller.  Since there are 3 calls which use the preempt method and about 10 without, I went this way.  I also thought it might be useful for this function to return the reason why it failed for the future.

src/server/req_holdjob.c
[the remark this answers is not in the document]
prakashcv13: Please ignore. I missed to see that this change was in req_releasejob().

src/server/svr_chk_owner.c
```diff
@@ -412,7 +412,7 @@ authenticate_user(struct batch_request *preq, struct connection *pcred)
 
 
 job *
-chk_job_request(char *jobid, struct batch_request *preq, int *rc)
+chk_job_request(char *jobid, struct batch_request *preq, int *rc, int *err)
```
arungrover: Please update function header of this function

src/server/req_preemptjob.c
```diff
@@ -377,14 +375,29 @@ reply_preempt_jobs_request(int code, int aux, struct job *pjob)
...
+		snprintf(log_buffer, sizeof(log_buffer), "preemption method %s failed for job (%d)", preempt_methods[aux], code);
+		log_event(PBSEVENT_DEBUG, PBS_EVENTCLASS_JOB, LOG_DEBUG, pjob->ji_qs.ji_jobid, log_buffer);
+
 		if (pjob->preempt_order[0].order[pjob->preempt_order_index] == PREEMPT_METHOD_CHECKPOINT)
 			clear_preempt_hold(pjob);
-		if (issue_preempt_request((int)pjob->preempt_order[0].order[pjob->preempt_order_index + 1], pjob, preq)) {
-			job_preempt_fail(preq, pjob->ji_qs.ji_jobid, code);
+		
+		pjob->preempt_order_index++;
+		if (pjob->preempt_order[0].order[pjob->preempt_order_index] != PREEMPT_METHOD_LOW) {
```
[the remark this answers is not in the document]
bhroam: I looked at the data in the debugger.  The extra data in the order array is PREEMPT_METHOD_LOW.  I suspect the array is just zeroed, and PREEMPT_METHOD_LOW is 0.

bhroam:
Updated log: [preemption.log](https://github.com/PBSPro/pbspro/files/3263886/preemption.log)

PR #1156, opened 2019-06-06
Test test_soft_job_array of TestSoftWalltime is failing while verifying the job's attribute

dinesh1306:
#### Describe Bug or Feature
Test test_soft_job_array of TestSoftWalltime is failing while verifying the job's attribute

#### Describe Your Change
Problem:
Test test_soft_job_array failling while checking job_state of first subjob in Array. 
In the Test array_job is submitted of 5 subjobs. then 5 sec. soft_walltime is set on array_job and check that sub_job is not deleted after completetion of soft_walltime.
The race condition occured such that if machine having more than 5 resources_available.ncpus value then all subjobs running parrallely and completed. PTL try to get job_state 'X' of first sub_job, but cannot get expected job_state in 60 attempts and test got failed. machines having resources_available.ncpus value less than 5 after completetion of first job PTL easily get 'X'. 

Solution:
Set resorces_available.ncpus value to 1 in test. After completion of first job, second job is in R state and PTL easily get job_state 'X' for first job.

- Before fix
[before_fix.txt](https://github.com/PBSPro/pbspro/files/3260206/before_fix.txt)
- After fix
[after_fix.txt](https://github.com/PBSPro/pbspro/files/3260207/after_fix.txt)

### DIFF
--- a/test/tests/functional/pbs_soft_walltime.py
+++ b/test/tests/functional/pbs_soft_walltime.py
@@ -1036,6 +1036,8 @@ e.accept()
         regular jobs
         """
 
+        a = {'resources_available.ncpus': 1}
+        self.server.manager(MGR_CMD_SET, NODE, a, id=self.mom.shortname)
         self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'False'})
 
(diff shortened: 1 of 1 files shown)


Klovely:
Looks good to me .

PR #1158, opened 2019-06-07
Sched crash when array job and reservation is submitted

arungrover:
#### Describe Bug or Feature
Scheduler crashes when it encounters an array job and a reservation in the same scheduling cycle.

#### Describe Your Change
As part of speeding up scheduler, we added a way to find a job/reservation based on their indexes in the all_resresv array.
But after these indexes are created, all_resresv array gets sorted to bring timed events in front in order to create a calendar. This sort messes up the indexes and causes the crash. While the sort is needed there is no reason to have all_resresv sorted.
As part of the fix, create_events code which creates the calendar will create a local copy of the array and sort that local copy instead of sorting all_resresv array.

#### Link to Design Doc
None

[smoketest_out.txt](https://github.com/PBSPro/pbspro/files/3267010/smoketest_out.txt)

#### NOTE
It is a chery-pick of #1109

### DIFF
--- a/.travis/opensuse_leap_15.sh
+++ b/.travis/opensuse_leap_15.sh
@@ -1,9 +1,9 @@
 #!/bin/bash -xe
-OS_NAME=$(${DOCKER_EXEC} cat /etc/os-release | awk -F[=\"] '/^NAME=/ {print $3}')
-OS_VERSION=$(${DOCKER_EXEC} cat /etc/os-release | awk -F[=\"] '/^VERSION=/ {print $3}')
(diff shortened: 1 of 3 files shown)


bhroam:
@arungrover this ticket is having the same issue as the other 19.1 PR that is out.  We need to cherry-pick the travis change, or someone has to pull it into their branch for travis to pass.

arungrover:
@bhroam I've cherry-picked the opensuse travis change as well.

bhroam:
Thanks for cherry-picking the travis PR.

PR #1160, opened 2019-06-12
Fixed server crash with expired subjobs in reservation

arungrover:
#### Describe Bug or Feature
When there are expired subjobs in a reservation and reservation ends PBS server tries to clean up all the jobs inside the reservation and it crashes while doing so.
This works fine when job_history is disabled, but when job_history is enabled then reservation's job list also consists of expired subjobs. 
What happens is that in req_deleteReservation() function we iterate through the list of all jobs and issue a local request to delete the jobs. While doing so we maintain a pointer to next job in the list because the job in hand will get deleted and then we can not iterate to next job.
Now, when we encounter an array parent, we issue a delete job request and then in req_deletejob we delete all the expired subjobs of this parent. This results in req_deleteReservations holding onto a dangling next job pointer (because the next job to delete was the expired subjob), hence the crash.

#### Describe Your Change
While trying to find the next job to delete in the list, we just skip all the expired subjobs. This makes server work only on array parents and other jobs.

[server_leak.txt](https://github.com/PBSPro/pbspro/files/3279120/server_leak.txt)
[test_fix.txt](https://github.com/PBSPro/pbspro/files/3279121/test_fix.txt)

### DIFF
--- a/src/server/req_delete.c
+++ b/src/server/req_delete.c
@@ -1138,9 +1138,6 @@ req_deleteReservation(struct batch_request *preq)
 				continue;
 			}
 
-			if (pjob == NULL)
-				break;
-
 			newreq = alloc_br(PBS_BATCH_DeleteJob);
 			if (newreq != NULL) {
 
--- a/src/server/req_delete.c
+++ b/src/server/req_delete.c
@@ -1120,6 +1120,12 @@ req_deleteReservation(struct batch_request *preq)
 
 			pnxj = (job *) GET_NEXT(pjob->ji_jobque);
 
+			/* skip all expired subjobs, expired subjobs are deleted when array parent is
+			 * issued delete request
+			 */
+			for (; pnxj != NULL && (pnxj->ji_qs.ji_svrflags & JOB_SVFLG_SubJob) &&
+			     pnxj->ji_qs.ji_state == JOB_STATE_EXPIRED; pnxj = (job *) GET_NEXT(pnxj->ji_jobque))
+				;
 			/*
 			 * If a history job (job state is JOB_STATE_MOVED
 			 * or JOB_STATE_FINISHED, then no need to delete
@@ -1131,6 +1137,10 @@ req_deleteReservation(struct batch_request *preq)
 				pjob = pnxj;
 				continue;
 			}
+
+			if (pjob == NULL)
+				break;
+
 			newreq = alloc_br(PBS_BATCH_DeleteJob);
 			if (newreq != NULL) {
 
--- a/test/tests/functional/pbs_reservations.py
+++ b/test/tests/functional/pbs_reservations.py
@@ -1098,8 +1098,8 @@ class TestReservations(TestFunctional):
         # Submit a advance reservation and an array job to the reservation
         # once reservation confirmed
         now = int(time.time())
-        a = {'reserve_start': now + 20,
-             'reserve_end': now + 80}
+        a = {'reserve_start': now + 10,
+             'reserve_end': now + 40}
         r = Reservation(TEST_USER, attrs=a)
         rid = self.server.submit(r)
         rid_q = rid.split('.')[0]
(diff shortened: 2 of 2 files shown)


src/server/req_delete.c
```diff
@@ -1120,6 +1120,12 @@ req_deleteReservation(struct batch_request *preq)
 
 			pnxj = (job *) GET_NEXT(pjob->ji_jobque);
 
+			/* skip all expired subjobs, expired subjobs are deleted when array parent is
+			 * issued delete request
+			 */
+			for (; pnxj != NULL && (pnxj->ji_qs.ji_svrflags & JOB_SVFLG_SubJob) &&
```
agrawalravi90: Aren't ALL subjobs deleted when an array parent is deleted? So shouldn't you skip over all subjobs regardless of their state?

agrawalravi90: One more thing, can there ever be a case where subjobs of an arrayjob are not listed all together? Suppose the arrayjob's user exceeds some limit after 2 of the subjobs were run, so the scheduler runs 2 of the subjobs and some other job. Then later more subjobs are run. Since subjobs aren't real jobs until they get run, can the subjobs possibly be listed non-consecutively in ji_jobque in such a case?

test/tests/functional/pbs_reservations.py
```diff
@@ -1087,3 +1087,39 @@ def test_advance_resv_with_multinode_job_array(self):
...
+        Test that an array job submitted to a reservation ends when
+        there are expired subjobs in the array job and job history is
+        enabled
+        """
+        self.common_steps()
+        # Submit a advance reservation and an array job to the reservation
+        # once reservation confirmed
+        now = int(time.time())
+        a = {'reserve_start': now + 20,
+             'reserve_end': now + 80}
```
agrawalravi90: isn't 60 seconds too long? How about just 10 seconds, and you can make the sleep time of subjobs 1 second

bhroam: How about make the reservation 30 seconds.  You can start it in 10s.  Make the jobs sleep for 2s.  This means you should run through 15 of the 20 subjobs during the reservation.

src/server/req_delete.c
arungrover: Regarding your first query, all subjobs (expired and running are deleted) when a parent job is deleted. expired subjobs are deleted instantaneously but for running subjobs it goes through a different code path of sending a signal to mom to delete the running job and then delete the array parent when all the subjobs are deleted.
For your second comment, As long as array parent is part of the reservation queue's job list, it should be fine. deletion of parent job should trigger delete of all expired subjobs which will make them dequeue from reservation's job queue and the server will not see those pointers anymore in the req_deleteReservation function.

test/tests/functional/pbs_reservations.py
arungrover: I feel 10 seconds is too small and on slower machines it may cause problems. I'll make it a reservation of 30 seconds and 2 second subjobs.

src/server/req_delete.c
arungrover: Since we already skip sending any subjob for deletion  (line 1135) I do not need the block to check pjob is null. I'll remove that block.

test/tests/functional/pbs_reservations.py
bhroam: Since the point of the test is to have a job array with some expired and some queued subjobs at the end of the reservation, then 10s should be fine.  If on slower machines the reservation is submitted and starts before the job array is submitted, then fewer subjobs will run.  They're still 2s each, so some number of them will run.  I think 10s is fine in this case.  In others, you are probably right.

src/server/req_delete.c
agrawalravi90: "Regarding your first query, all subjobs (expired and running are deleted) when a parent job is deleted. expired subjobs are deleted instantaneously but for running subjobs it goes through a different code path of sending a signal to mom to delete the running job and then delete the array parent when all the subjobs are deleted."

So, if we only skip over the expired jobs, then we might set next pointer to a queued/running subjob. In such a case, isn't it possible that by the time we access the next pointer, the queued/running subjob has been deleted and the pointer is invalid? Or what about the case where the list looks like this: parent job->expired subjob->a qrun job->a running subjob, now, isn't it possible that we set the next pointer to the running subjob and by the time we dereference it, the running subjob gets deleted and the pointer is invalid?

test/tests/functional/pbs_reservations.py
arungrover: @agrawalravi90 @bhroam what's the consensus? do you want me to reduce the time further from 30 seconds to 10 and 1 second jobs?

src/server/req_delete.c
arungrover: We are actually safe in case of queued subjobs because subjobs are enqueued when they start running. Now if the next pointer is pointing to the running subjob then that should be ok too because running subjobs are not deleted instantaneously. 
Now for the case related to parent job->expired subjob->qrun job-> running subjob. In this case, what will happen is that server will try to delete parent job and as part of that it will delete all the expired subjobs and do an issue_signal to moms for running subjobs and register a Deferred_Reply task in case mom replies back. Since the server is single threaded I think we are safe here too.

PR #1162, opened 2019-06-12
updates to load_config() to match pbs_cgroups_hook.py

vccardenas:
#### Describe Bug or Feature
Intermittent failure of TestCgroupsStress when verifying mom log message after loading hook configuration. 

#### Describe Your Change
1. Removed max_attempts from log_match in load_config().
2. Updated load_config() to match what's being done in the functional test pbs_cgroups_hook.py.

[TestCgroupsStress_x69_19.4_after_1.txt](https://github.com/PBSPro/pbspro/files/3282756/TestCgroupsStress_x69_19.4_after_1.txt)
[TestCgroupsStress_x36_19.4_after_1.txt](https://github.com/PBSPro/pbspro/files/3282757/TestCgroupsStress_x36_19.4_after_1.txt)

### DIFF
--- a/test/tests/performance/pbs_cgroups_stress.py
+++ b/test/tests/performance/pbs_cgroups_stress.py
@@ -190,6 +190,7 @@ class TestCgroupsStress(TestPerformance):
         Create a hook configuration file with the provided contents.
         """
         fn = self.du.create_temp_file(body=cfg)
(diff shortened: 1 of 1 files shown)


mike0042:
My only concern is that the string "pbs_cgroups.CF differ" does not appear in either log file, so the tests never actually triggered the condition you are attempting to fix.

PR #1163, opened 2019-06-13
Sched crash while creating node partition with indirect resource

arungrover:
#### Describe Bug or Feature
scheduler crashes while creating node partition when the node resource it is looking at is an indirect resource.

#### Describe Your Change
Since in case of indirect resources, we do not set value of the resource on the resource structure. Scheduler should instead be using the resource structure of the node on which this indirect resource is dependent on.

#### Link to Design Doc
NA

[test_out.txt](https://github.com/PBSPro/pbspro/files/3283641/test_out.txt)

### DIFF
--- a/src/scheduler/node_partition.c
+++ b/src/scheduler/node_partition.c
@@ -423,6 +423,9 @@ create_node_partitions(status *policy, node_info **nodes, char **resnames, unsig
 				res = &unset_res;
 			}
 			if (res != NULL) {
+				/* Incase of indirect resource, point it to the right place */
+			        if (res->indirect_res != NULL)
+					res = res->indirect_res;
 				for (val_i = 0; res->str_avail[val_i] != NULL; val_i++) {
 					/* 2: 1 for '=' 1 for '\0' */
 					if (reslen + strlen(res->str_avail[val_i]) + 2 < 1024) {
--- a/src/scheduler/node_partition.c
+++ b/src/scheduler/node_partition.c
@@ -535,6 +535,9 @@ create_node_partitions(status *policy, node_info **nodes, char **resnames, unsig
 				res = &unset_res;
 			}
 			if (res != NULL) {
+				/* Incase of indirect resource, point it to the right place */
+			        if (res->indirect_res != NULL)
+					res = res->indirect_res;
 				if (compare_res_to_str(res, np_arr[np_i]->res_val, CMP_CASE)) {
 					if (np_arr[np_i]->ok_break) {
 						tmpres = find_resource(nodes[node_i]->res, getallres(RES_HOST));
--- a/test/tests/functional/pbs_schedule_indirect_resources.py
+++ b/test/tests/functional/pbs_schedule_indirect_resources.py
@@ -114,7 +114,7 @@ class TestSchedulingIndirectResources(TestFunctional):
             self.submit_job(attr)]
 
         # since there are 6 vnodes and the test grouped them on foostr
-        # resource, we now have groups in pair of vnode 0-3, 1-4, 2-5
+        # resource, we now have groups in pair of vnode 0+3, 1+4, 2+5
         # check that the jobs are running on correct vnode groups.
         for i in range(3):
             self.server.expect(JOB, {'job_state': 'R'}, id=j[i][0])
--- a/test/tests/functional/pbs_schedule_indirect_resources.py
+++ b/test/tests/functional/pbs_schedule_indirect_resources.py
@@ -113,7 +113,9 @@ class TestSchedulingIndirectResources(TestFunctional):
             self.submit_job(attr),
             self.submit_job(attr)]
 
-        # Verify that the jobs are running on correct vnodes
+        # since there are 6 vnodes and the test grouped them on foostr
+        # resource, we now have groups in pair of vnode 0-3, 1-4, 2-5
+        # check that the jobs are running on correct vnode groups.
         for i in range(3):
             self.server.expect(JOB, {'job_state': 'R'}, id=j[i][0])
             self.server.status(JOB, 'exec_vnode', j[i][0])
--- a/test/tests/functional/pbs_schedule_indirect_resources.py
+++ b/test/tests/functional/pbs_schedule_indirect_resources.py
@@ -1,0 +1,128 @@
+# coding: utf-8
+
+# Copyright (C) 1994-2019 Altair Engineering, Inc.
+# For more information, contact Altair at www.altair.com.
+#
+# This file is part of the PBS Professional ("PBS Pro") software.
+#
+# Open Source License Information:
+#
+# PBS Pro is free software. You can redistribute it and/or modify it under the
+# terms of the GNU Affero General Public License as published by the Free
+# Software Foundation, either version 3 of the License, or (at your option) any
+# later version.
+#
+# PBS Pro is distributed in the hope that it will be useful, but WITHOUT ANY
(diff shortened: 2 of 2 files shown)


test/tests/functional/pbs_schedule_indirect_resources.py
```diff
@@ -0,0 +1,128 @@
...
+    """
+    Test Scheduler resolve indirect resources correctly
+    """
+
+    def config_complex_for_grouping(self, res, res_type='string', flag='h'):
+        """
+        Configure the PBS complex for node grouping test
+        """
+        # Create a non-consumable custom resource foostr'
+        self.server.add_resource(res, res_type, flag)
```
bhroam: add_resource() is an old function that manually manipulates the resourcedef file.  Use self.server.manager() instead.

test/tests/functional/pbs_schedule_indirect_resources.py
```diff
@@ -0,0 +1,128 @@
...
+        set as the node_group_key
+        -> Set 'foostr' to 'A', 'B', and 'C' on the first three vnodes
+        -> Set 'foostr' for the next three vnodes to point to the first three
+        vnodes correspondingly
+        -> Verify that the last three vnodes are part of their respective
+        placement sets
+        """
+
+        # Delete any existing vnodes on the system
+        self.server.manager(MGR_CMD_DELETE, NODE, None, "")
```
bhroam: I think create_vnodes() does this for you.

test/tests/functional/pbs_schedule_indirect_resources.py
```diff
@@ -0,0 +1,128 @@
...
+        attr = {'resources_available.foostr': '@vnode[2]'}
+        self.server.manager(MGR_CMD_SET, NODE, attr, 'vnode[5]')
+
+        # Verify that the last three vnodes are part of their respective
+        # placement sets
+        # Submit 3 jobs requesting placement sets 'A', 'B' and 'C', and vnodes
+        # vnode[3], vnode[4] and vnode[5] respectively
+        attr1 = {'Resource_List.select': '1:foostr=A:vnode=vnode[3]'}
+        attr2 = {'Resource_List.select': '1:foostr=B:vnode=vnode[4]'}
+        attr3 = {'Resource_List.select': '1:foostr=C:vnode=vnode[5]'}
```
bhroam: This isn't really testing node grouping.  It is testing indirect resources.  If you want to test node grouping, you can't request the node grouping resource.  Also requesting a specific vnode won't help either.  You need to submit 3 jobs requesting 2 nodes.  Make sure they're on the correct set of vnodes.

arungrover: Thanks for this comment, It made me find another problem in the code where I was missing the change.

arungrover:
@bhroam as discussed offline, I've added an additional comment in the test explaining how nodes are grouped and what we are looking for. Please have a look.

arungrover:
Thanks @bhroam! I've changed the comment

src/scheduler/node_partition.c
```diff
@@ -423,6 +423,9 @@ create_node_partitions(status *policy, node_info **nodes, char **resnames, unsig
 				res = &unset_res;
 			}
 			if (res != NULL) {
+				/* Incase of indirect resource, point it to the right place */
```
agrawalravi90: im sure there's good reason to do it this way, i'm just curious, why don't we just make 'res' be 'indirect_res' when it is an indirect resource?

test/tests/functional/pbs_schedule_indirect_resources.py
```diff
@@ -0,0 +1,123 @@
...
+# under a commercial license agreement.
+#
+# Use of Altair’s trademarks, including but not limited to "PBS™",
+# "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
+# trademark licensing policies.
+
+from tests.functional import *
+
+
+class TestSchedulingIndirectResources(TestFunctional):
```
agrawalravi90: instead of creating a specialized class for indirect resources, how about creating a more generic "TestIndirectResource" class?

test/tests/functional/pbs_schedule_indirect_resources.py
```diff
@@ -0,0 +1,123 @@
...
+    """
+    Test Scheduler resolve indirect resources correctly
+    """
+
+    def config_complex_for_grouping(self, res, res_type='string', flag='h'):
```
agrawalravi90: you are calling this function just once, in 1 test, so i kind of feel that it should just be part of the test instead of being in a separate function.

test/tests/functional/pbs_schedule_indirect_resources.py
```diff
@@ -0,0 +1,123 @@
...
+        """
+        Configure the PBS complex for node grouping test
+        """
+        # Create a custom resource
+        attr = {"type": res_type, "flag": flag}
+        self.server.manager(MGR_CMD_CREATE, RSC, attr, id=res)
```
agrawalravi90: there's also a server.add_resource() which basically does a qmgr, with some additional checking, i think. Just wondering if we should be using that interface to create the resource. What do you think?

test/tests/functional/pbs_schedule_indirect_resources.py
```diff
@@ -0,0 +1,123 @@
...
+            self.submit_job(attr),
+            self.submit_job(attr)]
+
+        # since there are 6 vnodes and the test grouped them on foostr
+        # resource, we now have groups in pair of vnode 0+3, 1+4, 2+5
+        # check that the jobs are running on correct vnode groups.
+        for i in range(3):
+            self.server.expect(JOB, {'job_state': 'R'}, id=j[i][0])
+            self.server.status(JOB, 'exec_vnode', j[i][0])
+            vn = j[i][1].get_vnodes()
```
agrawalravi90: this might be easier to read if you just added "jid, j = self.submit_job(attr)" in the for loop, and then referenced jid and j in the loop.

src/scheduler/node_partition.c
arungrover: IMO it is because there will be multiple node objects pointing to the same object and when you free them you will have dangling pointers all over the place. Unless we decide to add a reference count to sched_resource structure.

test/tests/functional/pbs_schedule_indirect_resources.py
arungrover: It was server.add_resource() earlier but Bhroam said that that's an old interface and suggested to change it to this.

test/tests/functional/pbs_schedule_indirect_resources.py
arungrover: In the future, When want to expand this test suite to add more tests for node grouping, these helper functions will be useful

arungrover:
[test_fix.txt](https://github.com/PBSPro/pbspro/files/3297871/test_fix.txt)
[test_bug.txt](https://github.com/PBSPro/pbspro/files/3297872/test_bug.txt)
@agrawalravi90 I've addressed your comments, please have a look again. Also, attached are before and after test results.

test/tests/functional/pbs_schedule_indirect_resources.py
bhroam: Oh, interesting, @agrawalravi90 is right.  server.add_resource() does use manager().  It existed before qmgr modified resources and modified the resourcedef file itself.  I thought it still did.  I guess it got rewritten at some point.  The only difference is that it keeps the PTL cache up to date.

PR #1164, opened 2019-06-13
Use UUID in CUDA_VISIBLE_DEVICES

minghui-liu:
#### Describe Bug or Feature
The numbers used in `CUDA_VISIBLE_DEVICES` (`0, 1, 2 ...`) are not stable, 

#### Describe Your Change
1. Use GPU UUID instead
2. Set `CUDA_DEVICE_ORDER` to `PCI_BUS_ID`

[cuda_visible_devices_uuid.txt](https://github.com/PBSPro/pbspro/files/3288314/cuda_visible_devices_uuid.txt)

### DIFF
--- a/src/hooks/cgroups/pbs_cgroups.PY
+++ b/src/hooks/cgroups/pbs_cgroups.PY
@@ -779,7 +779,7 @@ class HookUtils(object):
                     if key.startswith('mic'):
                         mics.append(key[3:])
                     elif key.startswith('nvidia'):
-                        gpus.append(key[6:])
+                        gpus.append(node.devices['gpu'][key]['uuid'])
                 if mics:
                     env_list.append('OFFLOAD_DEVICES=%s' %
                                     string.join(mics, ','))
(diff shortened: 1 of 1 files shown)


src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -856,7 +858,7 @@ class HookUtils(object):
             pbs.logmsg(pbs.EVENT_DEBUG4,
                        'assigned_resources: %s' %
                        (cgroup.assigned_resources))
-            cgroup.setup_job_devices_env()
+            cgroup.setup_job_devices_env(node.devices['gpu'])
```
bayucan: Would it be possible that node.devices does not have a key and value for 'gpu'? If so, it would cause an exception here. Maybe we need to check if key exists first?

minghui-liu: `node.devices` is populated by `_discover_devies()` when a NodeConfig instance is created. And `_discover_devies()` always initialize `self.devices['gpu']` to be an empty dictionary.  So this should be safe since the key `gpu` always exist. 

See line 1430, in function `_discover_devices`.

mike0042:
Please rebase your branch @minghui-liu. The NodeConfig class was renamed to NodeUtils.

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -997,7 +999,7 @@ class HookUtils(object):
                     if key.startswith('mic'):
                         mics.append(key[3:])
                     elif key.startswith('nvidia'):
-                        gpus.append(key[6:])
+                        gpus.append(node.devices['gpu'][key]['uuid'])
```
mike0042: I had some concern over whether the prior use of key[6:] was done for a reason. Looking at the rest of the changes, I'm convinced that you are doing the right thing here.

src/hooks/cgroups/pbs_cgroups.PY
bayucan: Looks good.

PR #1165, opened 2019-06-14
Update PTL FW to execute test apart from tests dir

shilpakodli:
#### Describe Bug or Feature
Currently in PTL, to run a test case using "pbs_benchpress -t" option, user has to go default test directory and then run the test case. If user is in non test directory then PTL used to throw an error saying " Unknown testsuite/Unknown testcase"

#### Describe Your Change
PTL will now search tests in current directory first, if it doesn't find test in current directory then it will search in PTL installed tests directory. If it doesn't find in both of the paths then it will search in environment variable "PTLTESTS" where user can explictly mention the path to tests directory.

#### Logs of test execution
[smoktest_results.txt](https://github.com/PBSPro/pbspro/files/3290142/smoktest_results.txt)

### DIFF
--- a/test/fw/bin/pbs_benchpress
+++ b/test/fw/bin/pbs_benchpress
@@ -437,7 +437,7 @@ if __name__ == '__main__':
         loader = PTLTestLoader()
         testinfo = PTLTestInfo()
         loader.set_data(testgroup, testsuites, excludes, True,
-                        need_info=True)
+                        testfiles)
         testinfo.set_data(testsuites, list_test,
                           showinfo, verbose, gen_ts_tree)
         plugins = (loader, testinfo)
@@ -453,7 +453,7 @@ if __name__ == '__main__':
         runner = PTLTestRunner()
         db = PTLTestDb()
         data = PTLTestData()
-        loader.set_data(testgroup, testsuites, excludes, follow)
+        loader.set_data(testgroup, testsuites, excludes, follow, testfiles)
         testtags.set_data(tags, eval_tags)
         runner.set_data(paramfile, testparam, lcov_bin, lcov_data, lcov_out,
                         genhtml_bin, lcov_nosrc, lcov_baseurl,
--- a/test/fw/bin/pbs_benchpress
+++ b/test/fw/bin/pbs_benchpress
@@ -434,10 +434,13 @@ if __name__ == '__main__':
         follow = True
 
     if only_info:
+        loader = PTLTestLoader()
         testinfo = PTLTestInfo()
+        loader.set_data(testgroup, testsuites, excludes, True,
+                        need_info=True)
         testinfo.set_data(testsuites, list_test,
                           showinfo, verbose, gen_ts_tree)
-        plugins = (testinfo,)
+        plugins = (loader, testinfo)
     elif (tags_info or list_tags):
         loader = PTLTestLoader()
         testtags = PTLTestTags()
--- a/test/fw/bin/pbs_benchpress
+++ b/test/fw/bin/pbs_benchpress
@@ -433,8 +433,9 @@ if __name__ == '__main__':
         testsuites = 'PBSTestSuite'
         follow = True
 
+    loader = PTLTestLoader()
+
     if only_info:
-        loader = PTLTestLoader()
         testinfo = PTLTestInfo()
         loader.set_data(testgroup, testsuites, excludes, True,
                         testfiles)
@@ -442,13 +443,11 @@ if __name__ == '__main__':
                           showinfo, verbose, gen_ts_tree)
         plugins = (loader, testinfo)
     elif (tags_info or list_tags):
-        loader = PTLTestLoader()
         testtags = PTLTestTags()
         loader.set_data(testgroup, testsuites, excludes, True)
         testtags.set_data(tags, eval_tags, tags_info, list_tags, verbose)
         plugins = (loader, testtags)
     else:
-        loader = PTLTestLoader()
         testtags = PTLTestTags()
         runner = PTLTestRunner()
         db = PTLTestDb()
--- a/test/fw/ptl/utils/plugins/ptl_test_info.py
+++ b/test/fw/ptl/utils/plugins/ptl_test_info.py
@@ -77,7 +77,7 @@ class PTLTestInfo(Plugin):
     Load test cases from given parameter
     """
     name = 'PTLTestInfo'
-    score = sys.maxint - 1
+    score = sys.maxint - 2
     logger = logging.getLogger(__name__)
 
     def __init__(self):
@@ -128,26 +128,9 @@ class PTLTestInfo(Plugin):
         """
         Is the class wanted?
         """
-        if not issubclass(cls, PBSTestSuite):
-            return False
-        if cls.__name__ == 'PBSTestSuite':
-            return False
         self._tree.setdefault(cls.__name__, cls)
         if len(cls.__bases__) > 0:
             self.wantClass(cls.__bases__[0])
-        return False
-
-    def wantFunction(self, function):
-        """
-        Is the function wanted?
-        """
-        return False
-
-    def wantMethod(self, method):
-        """
-        Is the method wanted?
-        """
-        return False
 
     def _get_hierarchy(self, cls, level=0):
         delim = '    ' * level
--- a/test/fw/ptl/utils/plugins/ptl_test_loader.py
+++ b/test/fw/ptl/utils/plugins/ptl_test_loader.py
@@ -208,7 +208,7 @@ class PTLTestLoader(Plugin):
                     tests_dir += [ptl_test_dir]
                 rv = old_loadTestsFromNames(tests_dir, module)
             else:
-                rv = old_loadTestsFromNames(names, module)
+                rv = old_loadTestsFromNames(tests_dir, module)
             self.check_unknown()
             return rv
         loader.loadTestsFromNames = check_loadTestsFromNames
--- a/test/fw/ptl/utils/plugins/ptl_test_loader.py
+++ b/test/fw/ptl/utils/plugins/ptl_test_loader.py
@@ -51,7 +51,7 @@ class PTLTestLoader(Plugin):
     Load test cases from given parameter
     """
     name = 'PTLTestLoader'
-    score = sys.maxint - 2
+    score = sys.maxsize - 1
     logger = logging.getLogger(__name__)
 
     def __init__(self):
@@ -67,7 +67,7 @@ class PTLTestLoader(Plugin):
         self.__tests_list_copy = {self._only_ts: [], self._only_tc: []}
         self.__allowed_cls = []
         self.__allowed_method = []
-        self.__need_info = False
+        self.testfiles = None
 
     def options(self, parser, env):
         """
@@ -75,7 +75,7 @@ class PTLTestLoader(Plugin):
         """
         pass
 
-    def set_data(self, testgroup, suites, excludes, follow, need_info=False):
+    def set_data(self, testgroup, suites, excludes, follow, testfiles=None):
         """
         Set the data required for loading test data
 
@@ -92,7 +92,7 @@ class PTLTestLoader(Plugin):
         if excludes is not None:
             self.excludes.extend(excludes.split(','))
         self.follow = follow
-        self.__need_info = need_info
+        self.testfiles = testfiles
 
     def configure(self, options, config):
         """
@@ -193,33 +193,35 @@ class PTLTestLoader(Plugin):
         old_loadTestsFromNames = loader.loadTestsFromNames
 
         def check_loadTestsFromNames(names, module=None):
-            rv = old_loadTestsFromNames(names, module)
-            test_list = self.__tests_list_copy
-            test_suites = self.__tests_list_copy['__only__ts__']
-            if (len(test_list) > 2) or (len(test_suites) != 0):
-                if 'PBSTestSuite' not in test_suites:
-                    self._du = DshUtils()
-                    ptl_test_dir = self._du.which(exe='pbs_benchpress')
-                    ptl_test_dir = ptl_test_dir.replace('bin/pbs_benchpress',
-                                                        'tests')
-                    user_test_dir = os.environ.get("PTL_TESTS_DIR", None)
-                    if user_test_dir is not None:
-                        if os.path.isdir(user_test_dir):
-                            rv = old_loadTestsFromNames([user_test_dir],
-                                                        module)
+            if not self.testfiles:
+                self._du = DshUtils()
+                ptl_test_dir = self._du.which(exe='pbs_benchpress')
+                ptl_test_dir = ptl_test_dir.replace('bin/pbs_benchpress',
+                                                    'tests')
+                user_test_dir = os.environ.get("PTL_TESTS_DIR", None)
+                if user_test_dir is not None:
+                    if os.path.isdir(user_test_dir):
+                        if os.path.isdir(ptl_test_dir):
+                            rv = old_loadTestsFromNames(
+                                names + [ptl_test_dir, user_test_dir], module)
                         else:
-                            _msg = "Invalid directory specified"
-                            _msg += " in PTL_TESTS_DIR"
-                            logging.error(_msg)
-                            sys.exit(1)
-                    elif os.path.isdir(ptl_test_dir):
-                        rv = old_loadTestsFromNames([ptl_test_dir], module)
-                    if not self.__need_info:
-                        self.check_unknown()
+                            rv = old_loadTestsFromNames(
+                                names + [user_test_dir], module)
+                    else:
+                        if os.path.isdir(ptl_test_dir):
+                            rv = old_loadTestsFromNames(names +
+                                                        [ptl_test_dir], module)
+                else:
+                    if os.path.isdir(ptl_test_dir):
+                        rv = old_loadTestsFromNames(names +
+                                                    [ptl_test_dir], module)
+                    else:
+                        rv = old_loadTestsFromNames(names, module)
             else:
-                if not self.__need_info:
-                    self.check_unknown()
+                rv = old_loadTestsFromNames(names, module)
+            self.check_unknown()
             return rv
+
         loader.loadTestsFromNames = check_loadTestsFromNames
         return loader
 
--- a/test/fw/ptl/utils/plugins/ptl_test_loader.py
+++ b/test/fw/ptl/utils/plugins/ptl_test_loader.py
@@ -202,9 +202,8 @@ class PTLTestLoader(Plugin):
                 ptl_test_dir = os.path.join(ptl_test_dir.split('ptl')[0],
                                             "ptl", "tests")
                 user_test_dir = os.environ.get("PTL_TESTS_DIR", None)
-                if user_test_dir is not None:
-                    if os.path.isdir(user_test_dir):
-                        tests_dir += [user_test_dir]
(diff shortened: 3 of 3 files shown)


agrawalravi90:
Hi @shilpakodli , thanks for proposing this change, this has irked me for so long!
But I do think that this deserves a design document for 2 reasons:
1) So that a wider audience gets to know that this is finally happening and
2) You are introducing a new environment variable, and you've taken some design decisions for how the tests will be found etc., i think it'll be useful to take feedback from others on this.

Thanks again for doing this

test/fw/ptl/utils/plugins/ptl_test_loader.py
```diff
@@ -191,7 +194,30 @@ def prepareTestLoader(self, loader):
 
         def check_loadTestsFromNames(names, module=None):
             rv = old_loadTestsFromNames(names, module)
-            self.check_unknown()
+            test_list = self.__tests_list_copy
+            test_suites = self.__tests_list_copy['__only__ts__']
+            if (len(test_list) > 2) or (len(test_suites) != 0):
+                if 'PBSTestSuite' not in test_suites:
```
hirenvadalia: why do you need this if condition? what if the user is trying to run `pbs_benchpress -i -t PBSTestSuite` when PTL_TESTS_DIR is set to a valid non-install directory

test/fw/ptl/utils/plugins/ptl_test_loader.py
```diff
@@ -191,7 +194,30 @@ def prepareTestLoader(self, loader):
...
+                            rv = old_loadTestsFromNames([user_test_dir],
+                                                        module)
+                        else:
+                            logging.error("Invalid dircetory specified in
+                                          PTL_TESTS_DIR")
+                            sys.exit(1)
+                    elif os.path.isdir(ptl_test_dir):
+                        rv = old_loadTestsFromNames([ptl_test_dir], module)
+                    if not self.__need_info:
+                        self.check_unknown()
```
hirenvadalia: What will be an error message if the user runs `pbs_benchpress -i -t UnknownTestSuite` with your changes?

test/fw/ptl/utils/plugins/ptl_test_loader.py
shilpakodli: When pbs_benchpress is run with -f option where file is not present in current directory then, self.__test_list_copy will have PBSTestSuite in self.__tests_list_copy['__only__ts__'] and it will go in if condition and try to run all of the test suites which is not expected. I have put this condition to skip for -f option.

test/fw/ptl/utils/plugins/ptl_test_loader.py
shilpakodli: So, in pbs_benchpress, I am calling loader plugin and then testinfo plugin. Loader will load all of the test cases in current/User specified/Installed directory and runs check_unknown function which I don't want to call for -i option as it is taken care when testinfo plugin is called.

test/fw/ptl/utils/plugins/ptl_test_loader.py
```diff
@@ -191,7 +194,30 @@ def prepareTestLoader(self, loader):
 
         def check_loadTestsFromNames(names, module=None):
             rv = old_loadTestsFromNames(names, module)
-            self.check_unknown()
+            test_list = self.__tests_list_copy
+            test_suites = self.__tests_list_copy['__only__ts__']
+            if (len(test_list) > 2) or (len(test_suites) != 0):
```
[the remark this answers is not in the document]
shilpakodli: "old_loadTestsFromNames(names, module)" is called at the beginning of check_loadTestsFromNames method with names = "Current Dircetory".
If the given test case/test suite to be run is not present in current directory then "old_loadTestsFromNames" will update self.__tests_list_copy with non-existing test suite/test case.
self.__test_list_copy is assigned initially to "{self._only_ts: [], self._only_tc: []}" when loader plugin is called.
> Suppose if "MytestSuite" which I want to run is not present in current directory then self__test_list_copy will have "{self._only_ts: ["MytestSuite"], self._only_tc: []}"
> If "MytestSuite.testcase1" which is a single test case in MytestSuite test suite is not present in current directory, then self.__test_list_copy will have "{self._only_ts: [], self._only_tc: [], " MytestSuite" : ["testcase1"]}
So "if (len(test_list) > 2) or (len(test_suites) != 0):" is for above two scenarios.

test/fw/ptl/utils/plugins/ptl_test_loader.py
shilpakodli: When I run this below command 
"pbs_benchpress -i -t PBSTestSuite" , it is giving output of all methods present in ptl/utils/pbs_testsuite.py

test/fw/ptl/utils/plugins/ptl_test_loader.py
```diff
@@ -191,7 +194,31 @@ def prepareTestLoader(self, loader):
 
         def check_loadTestsFromNames(names, module=None):
             rv = old_loadTestsFromNames(names, module)
-            self.check_unknown()
+            test_list = self.__tests_list_copy
```
hirenvadalia: @shilpakodli As per our offline discussion, we don't need this change we just need to pass a new directory in `old_loadTestsFromNames` along with names. Also, you won't need a new argument to `set_data` after new changes

shilpakodli:
@hirenvadalia Thanks for the review. I have addressed your offline review comments and tested on my VM.

shilpakodli:
> Hi @shilpakodli , thanks for proposing this change, this has irked me for so long!
> But I do think that this deserves a design document for 2 reasons:
> 
> 1. So that a wider audience gets to know that this is finally happening and
> 2. You are introducing a new environment variable, and you've taken some design decisions for how the tests will be found etc., i think it'll be useful to take feedback from others on this.
> 
> Thanks again for doing this

EDD got sign off now and code is ready for review. Thanks for bringing up this @agrawalravi90

test/fw/ptl/utils/plugins/ptl_test_loader.py
```diff
@@ -190,9 +193,35 @@ def prepareTestLoader(self, loader):
...
 
         def check_loadTestsFromNames(names, module=None):
-            rv = old_loadTestsFromNames(names, module)
+            if not self.testfiles:
+                self._du = DshUtils()
+                ptl_test_dir = self._du.which(exe='pbs_benchpress')
+                ptl_test_dir = ptl_test_dir.replace('bin/pbs_benchpress',
+                                                    'tests')
+                user_test_dir = os.environ.get("PTL_TESTS_DIR", None)
+                if user_test_dir is not None:
```
hirenvadalia: as discussed offline please remove this if-else condition and create a new list with names and append ptl_test_dir and user_test_dir based on availability and in order.

shilpakodli:
@hirenvadalia  As per offline discussion, I have addressed your review comments.

shilpakodli:
@hirenvadalia I am attaching test execution results with following observation.
test case : pbs_test_pbs.py
Same test case is present in PTL_TESTS_DIR, PTL Installed Directory and current directory but with different print statement.

PTL_TESTS_DIR (/tmp/tests) - Executed from PTL TESTS Directory
PTL Installed Dir (/opt/ptl/tests) - Executed from PTL Installed Directory
Current Directory (/home/pbsroot/usertemp/tests) - Executed from Current Directory

I have attached logs with three scenarios with condition as same test suite present in all three directories:

scenario1 : Ran from Non-tests directory and PTL_TESTS_DIR set to /tmp/tests/
            Result : Test should execute from PTL_TESTS_DIR.
Scenario2 : Ran from current dircetory and PTL_TESTS_DIR set to /tmp/tests/
            Result : Test should execute from current directory
Scenario3 : Ran from non-tests dircetory and unset PTL_TESTS_DIR
            Result : Test should execute from PTL Installed directory.

[scenario1_results.txt](https://github.com/PBSPro/pbspro/files/3381686/scenario1_results.txt)
[scenario2_results.txt](https://github.com/PBSPro/pbspro/files/3381687/scenario2_results.txt)
[scenario3_results.txt](https://github.com/PBSPro/pbspro/files/3381688/scenario3_results.txt)

test/fw/ptl/utils/plugins/ptl_test_loader.py
```diff
@@ -190,7 +195,20 @@ def prepareTestLoader(self, loader):
         old_loadTestsFromNames = loader.loadTestsFromNames
 
         def check_loadTestsFromNames(names, module=None):
-            rv = old_loadTestsFromNames(names, module)
+            tests_dir = names
+            if not self.testfiles:
+                self._du = DshUtils()
+                ptl_test_dir = self._du.which(exe='pbs_benchpress')
```
hirenvadalia: what if pbs_benchpress is not in the path? you can use sys.argv[0].

shilpakodli: Can we use PBS_EXEC path and then replace pbs with ptl ?

hirenvadalia: I don't think use PBS_EXEC is good. it's not guaranteed that use has installed PTL in /opt/ptl.
So it is best to use sys.argv[0]

shilpakodli: I agree. Wish we could have something called PTL_HOME like PBS_HOME which represents PTL installed path. 
For now, I have changed to sys.argv[0] considering user will always use full path of pbs_benchpress to run the test case.

test/fw/bin/pbs_benchpress
```diff
@@ -434,10 +434,13 @@ if __name__ == '__main__':
         follow = True
 
     if only_info:
+        loader = PTLTestLoader()
```
prakashcv13: we can move this out of if-elif-else block

test/fw/ptl/utils/plugins/ptl_test_loader.py
```diff
@@ -66,14 +68,15 @@ def __init__(self):
...
+        self.testfiles = None
 
     def options(self, parser, env):
         """
         Register command line options
         """
         pass
 
-    def set_data(self, testgroup, suites, excludes, follow):
+    def set_data(self, testgroup, suites, excludes, follow, testfiles=None):
```
prakashcv13: add param testfiles in the docstring

shilpakodli:
@hirenvadalia Addressed offline review comments.

shilpakodli:
@prakashcv13 I have addressed your review comments. Please have a look.

test/fw/ptl/utils/plugins/ptl_test_loader.py
```diff
@@ -202,7 +203,8 @@ def check_loadTestsFromNames(names, module=None):
                                             "ptl", "tests")
                 user_test_dir = os.environ.get("PTL_TESTS_DIR", None)
                 if user_test_dir is not None:
-                    tests_dir += [user_test_dir]
+                    if os.path.isdir(user_test_dir):
```
hirenvadalia: you can merge this if with above if: `if user_test_dir and os.path.isdir(user_test_dir):`

test/fw/ptl/utils/plugins/ptl_test_info.py
```diff
@@ -128,26 +128,9 @@ def wantClass(self, cls):
         """
         Is the class wanted?
         """
-        if not issubclass(cls, PBSTestSuite):
```
agrawalravi90: why did you remove this?

test/fw/ptl/utils/plugins/ptl_test_info.py
```diff
@@ -128,26 +128,9 @@ def wantClass(self, cls):
...
-            return False
-        if cls.__name__ == 'PBSTestSuite':
-            return False
         self._tree.setdefault(cls.__name__, cls)
         if len(cls.__bases__) > 0:
             self.wantClass(cls.__bases__[0])
-        return False
-
-    def wantFunction(self, function):
```
agrawalravi90: was this dead code?

test/fw/ptl/utils/plugins/ptl_test_loader.py
```diff
@@ -165,11 +170,12 @@ def check_unknown(self):
         Check for unknown test suite and test case
         """
         log.debug('check_unknown called')
-        only_ts = self.__tests_list_copy.pop(self._only_ts)
-        only_tc = self.__tests_list_copy.pop(self._only_tc)
+        tests_list_copy = copy.deepcopy(self.__tests_list_copy)
```
agrawalravi90: are you modifying the contents of self.__tests_list_copy in this function? I don't see them getting modified, in which case you don't need the deepcopy

test/fw/ptl/utils/plugins/ptl_test_loader.py
```diff
@@ -190,7 +196,19 @@ def prepareTestLoader(self, loader):
         old_loadTestsFromNames = loader.loadTestsFromNames
 
         def check_loadTestsFromNames(names, module=None):
-            rv = old_loadTestsFromNames(names, module)
+            tests_dir = names
+            if not self.testfiles:
+                ptl_test_dir = __file__
+                ptl_test_dir = os.path.join(ptl_test_dir.split('ptl')[0],
```
agrawalravi90: don't you also need to modify the Makefiles so that tests get copied to this path when PTL is installed?

test/fw/ptl/utils/plugins/ptl_test_loader.py
```diff
@@ -190,7 +196,19 @@ def prepareTestLoader(self, loader):
...
+                                            "ptl", "tests")
+                user_test_dir = os.environ.get("PTL_TESTS_DIR", None)
+                if user_test_dir and os.path.isdir(user_test_dir):
+                    tests_dir += [user_test_dir]
+                if os.path.isdir(ptl_test_dir):
+                    tests_dir += [ptl_test_dir]
```
agrawalravi90: So, this will mean that if PTL_TESTS_DIR is set, then tests will be picked from \<ptl install\>/tests as well as PTL_TESTS_DIR. Is that really what we want? I kind of feel that if users specify PTL_TESTS_DIR, then that should be the ONLY directory where tests should be searched for. What do you think?

test/fw/ptl/utils/plugins/ptl_test_loader.py
```diff
@@ -190,7 +196,19 @@ def prepareTestLoader(self, loader):
...
+                rv = old_loadTestsFromNames(tests_dir, module)
+            else:
+                rv = old_loadTestsFromNames(names, module)
```
agrawalravi90: you set tests_dir to name before, so you can just call "rv = old_loadTestsFromNames(tests_dir, module)" right?

test/fw/ptl/utils/plugins/ptl_test_loader.py
shilpakodli: We are assuming here that when PTL is installed, tests will reside in parallel to PTL library.

test/fw/ptl/utils/plugins/ptl_test_loader.py
shilpakodli: tests_list_copy.pop removes items. That's why copying self.__tests_list_copy to another variable.

test/fw/ptl/utils/plugins/ptl_test_info.py
shilpakodli: Yes. This was not called anywhere.

hirenvadalia: @agrawalravi90 It was not the dead code and @shilpakodli it was getting called by nose internal call. And test info plugin was designed such a way that it works without any other ptl node plugin but with this PR changes now we are using loader plugin along with info plugin and loader plugin already has wantFunction so we don't need it here.

test/fw/ptl/utils/plugins/ptl_test_info.py
hirenvadalia: This was there just ensure that we are not adding any class in self._tree which are not subclass of PBSTestSuite (as we want only test class to be added in self._tree) but now that is covered by loader plugin so we don't need to do it here. That's why it was removed

test/fw/ptl/utils/plugins/ptl_test_loader.py
hirenvadalia: > don't you also need to modify the Makefiles so that tests get copied to this path when PTL is installed?

This is already done. see https://github.com/PBSPro/pbspro/blob/master/test/tests/Makefile.am

test/fw/ptl/utils/plugins/ptl_test_loader.py
hirenvadalia: We are giving 3 dir to the loader (in a specific order: 1. current dir 2. PTL_TESTS_DIR 3. PTL install dir) to find given tests and loader will load tests from only first dir it find test rest will be ignored for perticular test.

test/fw/ptl/utils/plugins/ptl_test_loader.py
hirenvadalia: Agree.

test/fw/ptl/utils/plugins/ptl_test_loader.py
agrawalravi90: Ok, but I feel that it's more intuitive to only look at PTL_TESTS_DIR if it is set by the user. If unset, we can look in current dir + the default install location.

hirenvadalia: Ok, but what if the given test is not in PTL_TESTS_DIR but in default install location?

agrawalravi90: then we'll say test not found, and that'll be expected behavior. If you disagree then we can ask for an opinion on forums.

hirenvadalia: @agrawalravi90 that's what we discussed on forums if I recall it correctly and it concluded that we will give all 3 directly to the loader and let loader find test from one of them and if not found in all 3 then only give test not found error. But saying that I am happy re-discuss that but current code are in line with the current design so if we want to change what you say then first we need to change the design.

agrawalravi90: if this has already been discussed then never-mind, I don't want to delay the PR. Let's check it in as is and if somebody else feels that it should change or gets confused by the behavior then we can bring this up again.

agrawalravi90:
@shilpakodli essentially there's only 1 outstanding comment from me, thanks to @hirenvadalia for clarifying other comments. So, you just need to address this one:
"you set tests_dir to name before, so you can just call "rv = old_loadTestsFromNames(tests_dir, module)" right?"

Thanks.

shilpakodli:
Thank you @hiren for clarifying Ravi's questions on behalf of me. Thanks @agrawalravi90 for the review. I have addressed @agrawalravi90 comments.

test/fw/ptl/utils/plugins/ptl_test_loader.py
```diff
@@ -208,7 +208,7 @@ def check_loadTestsFromNames(names, module=None):
                     tests_dir += [ptl_test_dir]
                 rv = old_loadTestsFromNames(tests_dir, module)
             else:
-                rv = old_loadTestsFromNames(names, module)
+                rv = old_loadTestsFromNames(tests_dir, module)
```
hirenvadalia: you can remove else part and put old_loadTestsFromNames part outside of if condition

PR #1169, opened 2019-06-18
Fix for schema upgrade with dbuser 'pbs'

bremanandjk:
#### Describe Bug or Feature
Overlay upgrade fails due to a mismatch in "hstore" extension's schema name. "hstore" extension was being created without any specified schema value. By default, it was chosen "public"/"pbs". 

#### Describe Your Change
 Modified the extension creation and its usage under "pbs" schema.

#### Link to Design Doc

NA

[After_fix.txt](https://github.com/PBSPro/pbspro/files/3302596/After_fix.txt)
[Before_fix.txt](https://github.com/PBSPro/pbspro/files/3302597/Before_fix.txt)

### DIFF
--- a/src/cmds/scripts/pbs_schema_upgrade
+++ b/src/cmds/scripts/pbs_schema_upgrade
@@ -161,9 +161,9 @@ upgrade_pbs_schema_from_v1_3_0() {
 
 	${PGSQL_DIR}/bin/psql -p ${PBS_DATA_SERVICE_PORT} -d pbs_datastore -U ${PBS_DATA_SERVICE_USER} <<-EOF > /dev/null
 		
-		CREATE EXTENSION hstore;
+		CREATE EXTENSION hstore SCHEMA pbs;
 		
-		ALTER TABLE pbs.job ADD attributes public.hstore DEFAULT ''::public.hstore;
+		ALTER TABLE pbs.job ADD attributes pbs.hstore DEFAULT ''::pbs.hstore;
 		UPDATE pbs.job SET attributes=(
 			SELECT hstore(array_agg(attr.key ), array_agg(attr.value))
 				FROM ( SELECT concat(attr_name, '.' , attr_resource) AS key,
@@ -172,7 +172,7 @@ upgrade_pbs_schema_from_v1_3_0() {
 		UPDATE pbs.job SET attributes='' WHERE attributes IS NULL;
 		ALTER TABLE pbs.job ALTER COLUMN attributes SET NOT NULL;
 
-		ALTER TABLE pbs.node ADD attributes public.hstore DEFAULT ''::public.hstore;
+		ALTER TABLE pbs.node ADD attributes pbs.hstore DEFAULT ''::pbs.hstore;
 		UPDATE pbs.node SET attributes=(
 			SELECT hstore(array_agg(attr.key ), array_agg(attr.value))
 				FROM ( SELECT concat(attr_name, '.' , attr_resource) AS key,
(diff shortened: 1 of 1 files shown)


arungrover:
@bremanandjk would this change affect upgrades again where schema has changed name from public to pbs?

subhasisb:
> @bremanandjk would this change affect upgrades again where schema has changed name from public to pbs?

Yes, we must test upgrading from versions where schema was public to this

bremanandjk:
> > @bremanandjk would this change affect upgrades again where schema has changed name from public to pbs?
> 
> Yes, we must test upgrading from versions where schema was public to this

Thanks, @arungrover & @subhasisb for your reviews.

I have tested the upgrade case from PBS Pro v19.1 source code with public.hstore to latest PBS Pro(master) with this fix.
The overlay upgrade succeeded smoothly, 

As there is no DBMS version change, there is no issue with upgrades.   

P:S: Even after the upgrade, the "hstore" will remain in "public" schema. I think it's fine to remain as it 
       is. While performing any ALTER statements in the future, we need to take care of "hstore" object 
       access. 

Please let me know, your thoughts on this!. 

[test_upgrade_from_public_hstore_to_pbs_hstore.txt](https://github.com/PBSPro/pbspro/files/3305109/test_upgrade_from_public_hstore_to_pbs_hstore.txt)

subhasisb:
> > > @bremanandjk would this change affect upgrades again where schema has changed name from public to pbs?
> > 
> > 
> > Yes, we must test upgrading from versions where schema was public to this
> 
> Thanks, @arungrover & @subhasisb for your reviews.
> 
> I have tested the upgrade case from PBS Pro v19.1 source code with public.hstore to latest PBS Pro(master) with this fix.
> The overlay upgrade succeeded smoothly,
> 
> As there is no DBMS version change, there is no issue with upgrades.
> 
> P:S: Even after the upgrade, the "hstore" will remain in "public" schema. I think it's fine to remain as it
> is. While performing any ALTER statements in the future, we need to take care of "hstore" object
> access.
> 
> Please let me know, your thoughts on this!.
> 
> [test_upgrade_from_public_hstore_to_pbs_hstore.txt](https://github.com/PBSPro/pbspro/files/3305109/test_upgrade_from_public_hstore_to_pbs_hstore.txt)

That sounds fine. Whatever worked with public will continue to work anyway. The part where we access hstore extension is only at the creation or during upgrade, so sounds good to me.

subhasisb:
> While performing any ALTER statements in the future, we need to take care of "hstore" object
> access.

Yes, true, we need to be careful in furture upgrade steps.

bremanandjk:
> I do have a question. Can we not Alter schema to rename it from public to PBS? That way we don’t have to worry about any Alter we might do in future?

Correct me if I am wrong!

I am not sure about renaming the schema from "public" -> "pbs" would be the perfect approach, since some pgsql default modules might depend on "public" schema, those functionalities may affect due to this conversion.

My perspective is, we can handle this situation while writing **upgrade_pbs_schema_from_v1_4_0()** like 

```
upgrade_pbs_schema_from_v1_4_0() {

CREATE EXTENSION [ IF NOT EXISTS ] hstore SCHEMA pbs;
ALTER ....
....
}
```

**Note**: I just learned from [https://www.postgresql.org/docs/9.1/sql-alterschema.html](https://www.postgresql.org/docs/9.1/sql-alterschema.html ) that "There is no ALTER SCHEMA statement in the SQL standard."

arungrover:
Can you please check if it works that way. I will be surprised if Postgres modifies its default modules because our Postgres user does not own them. I think the change is harmless but I could be wrong.

bremanandjk:
I just tried to rename the "public" schema, I get below error!

```
pbs_datastore=# ALTER SCHEMA public RENAME To pbs;
ERROR:  schema "pbs" already exists
```
Since we have already existed "pbs" schema for all our relations and roles, renaming won't be our preferred choice. What do you say?

arungrover:
Fair enough! Thanks for trying!

bremanandjk:
Thanks! @arungrover for merging in.

PR #1171, opened 2019-06-19
single-node non-interactive job cannot start a task using pbsdsh or…

vchlum:
… pbs_tmrsh

#### Describe Bug or Feature
This bug was introduced in #1127 and it is caused by calling initgroups() as the user (not superuser). It fails. The function becomeuser() is called within set_credentils() and set_credentils() is called before open_std_out_err().  open_std_out_err() leads to calling impersonate_user() and the initgroups() subsequently.

It can lead to troubles with MPI too.

Steps to reproduce:
`$ echo 'pbs_tmrsh $(hostname -f) hostname' | qsub -l select=vnode=torque4
28739.torque4.ics.muni.cz`
It causes following error in mom log:
`06/19/2019 13:28:28;0001;pbs_mom;Job;28739.torque4.ics.muni.cz;Unable to open standard output/error`
`06/19/2019 13:28:28;0001;pbs_mom;Job;28739.torque4.ics.muni.cz;task not started, Retry hostname -3`
and the output is empty:
`-rw-------  1 vchlum meta            0 čen 19 13:28 STDIN.o28739`

#### Describe Your Change
Call initgroups() only if it is not called as the user.

#### Link to Design Doc
None

[ptl_initgroups_as_user.txt](https://github.com/PBSPro/pbspro/files/3305780/ptl_initgroups_as_user.txt)

### DIFF
--- a/src/resmom/start_exec.c
+++ b/src/resmom/start_exec.c
@@ -906,7 +906,8 @@ impersonate_user(uid_t uid, gid_t gid)
 	if (pwd == NULL)
 		return -1;
 
-	if (initgroups(pwd->pw_name, gid) == -1) {
+	if ((geteuid() != uid) &&
+		(initgroups(pwd->pw_name, gid) == -1)) {
 		return -1;
 	}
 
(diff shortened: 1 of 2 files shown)


mike0042:
Thanks again @vchlum!

PR #1172, opened 2019-06-19
Sched's run_update_resresv is perf bottleneck for large no. of jobs

agrawalravi90:
#### Describe Bug or Feature

When there are a large number of jobs in the system (test used 1 million), scheduler's run_update_resresv() becomes a performance bottleneck and takes around 75+% of the total sched cycle time. This happens because the code calls check_run_job() to recreate the array of running jobs after every job is run.

#### Describe Your Change

Instead of calling check_run_job(), just call add_resresv_to_array() to add a job run to the array.

Before:
Screen Shot 2019-06-11 at 2 54 11 PM

After:
Screen Shot 2019-06-12 at 10 43 09 PM

### DIFF
--- a/src/scheduler/queue_info.c
+++ b/src/scheduler/queue_info.c
@@ -674,9 +674,8 @@ update_queue_on_run(queue_info *qinfo, resource_resv *resresv, char *job_state)
 
 		req = req->next;
 	}
-	free(qinfo->running_jobs);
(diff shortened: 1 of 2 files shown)


agrawalravi90:
Thanks @bhroam 
@PBSPro/pbspro-maintainers requesting a maintainer review on this, thanks.

PR #1173, opened 2019-06-19
Fixed pbspro.spec for opensuse 15.1

bhroam:
#### Describe Bug or Feature
rpmbuild fails to build RPMs on opensuse 15.1

#### Describe Your Change
opensuse changed the name of lib directories from 'lib' to 'lib64'.  We hard coded the 'lib' directory in our pbspro.spec file.  This fix wildcards the lib directory so it will work for both CentOS and opensuse.

Thanks @mike0042 for figuring this out!

### DIFF
--- a/pbspro.spec
+++ b/pbspro.spec
@@ -423,7 +423,7 @@ ${RPM_INSTALL_PREFIX:=%{pbs_prefix}}/libexec/pbs_posttrans \
 %{pbs_prefix}/*
 %attr(4755, root, root) %{pbs_prefix}/sbin/pbs_rcp
 %attr(4755, root, root) %{pbs_prefix}/sbin/pbs_iff
-%attr(644, root, root) %{pbs_prefix}/lib/libpbs.la
+%attr(644, root, root) %{pbs_prefix}/lib*/libpbs.la
 %{_sysconfdir}/profile.d/pbs.csh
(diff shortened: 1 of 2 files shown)


mike0042:
The opensuse.org repo is causing problems. We may want to switch to a more reliable mirror and update .travis/opensuse_leap_15.sh accordingly. I don't know where the Travis CI systems are geographically located, so I don't know where the closest mirror would be. I think we should just pick another in the U.S. I've been using the one at Georgia Tech.

bhroam:
I didn't change the mirror.  I just restarted the travis build.  If you want to change the mirror, I'd say we ask @hirenvadalia to do it.  He's the expert on Travis.

mike0042:
I remember multiple times where the OpenSUSE build failed due to the package server. The opensuse.org server is likely the busiest of them all because it is the default. There are several reliable mirrors that might lead to fewer failures. What do you think @hirenvadalia?

hirenvadalia:
@bhroam and @mike0042 By default OpenSuSE main download server redirect to nearby mirror server only (read the 2nd paragraph on https://mirrors.opensuse.org/). But currently, OpenSuSE main download server is facing some problem so download is a little slow (see https://status.opensuse.org/).

mike0042:
Thanks @hirenvadalia. Wasn't aware that's how it worked.

PR #1177, opened 2019-06-21
Fixed invalid read in qrerun, qhold, qdel, and qsig

bhroam:
#### Describe Bug or Feature
The functions req_deletejob(), req_signaljob(), req_holdjob(), and req_rerunjob() call chk_job_request() to do some checking and return the pjob.  If the checks fail, chk_job_request() returns NULL.  The calling functions need the pjob in order to check and reply to the preemption request for the job.  If chk_job_request() returned NULL, the functions would call find_job() on the job id.  The job id is passed into these functions as part of the preq.  Since chk_job_request() replied to the preq, it was freed.  When find_job() was called, we were accessing freed memory.

#### Describe Your Change
Copy the job id into a local variable before calling chk_job_request()

[valgrind-before.log](https://github.com/PBSPro/pbspro/files/3316197/valgrind-before.log)
[valgrind-after.log](https://github.com/PBSPro/pbspro/files/3316196/valgrind-after.log)

Look for the four calling functions in the before log and for them not being there in the after log.  There is an invalid read from python, but that's because I use a python compiled for valgrind.

### DIFF
--- a/src/server/req_delete.c
+++ b/src/server/req_delete.c
@@ -355,7 +355,7 @@ req_deletejob(struct batch_request *preq)
 	int forcedel = 0;
 	int i;
 	int j;
-	char *jid;
+	char jid[PBS_MAXSVRJOBID + 1];
 	int jt; /* job type */
 	int offset;
 	char *pc;
@@ -370,7 +370,7 @@ req_deletejob(struct batch_request *preq)
 	int count = 0;
 	int err = PBSE_NONE;
 
-	jid = preq->rq_ind.rq_delete.rq_objname;
+	snprintf(jid, sizeof(jid), "%s", preq->rq_ind.rq_delete.rq_objname);
 
 	if (preq->rq_extend && strstr(preq->rq_extend, DELETEHISTORY))
 		delhist = 1;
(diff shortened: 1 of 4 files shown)


bhroam:
chk_job_request() does not call reply_preempt_jobs_request().  There was one left over call that I removed (either in this PR or another one).  It can't call reply_preempt_jobs_request() because 
A) it is called in a lot of places, most of them not dealing with preemption
B) We won't know what preemption method is being rejected inside this function since it is called by all 4 preemption methods.

src/server/req_delete.c
```diff
@@ -370,7 +370,7 @@ req_deletejob(struct batch_request *preq)
 	int count = 0;
 	int err = PBSE_NONE;
 
-	jid = preq->rq_ind.rq_delete.rq_objname;
+	snprintf(jid, sizeof(jid), "%s", preq->rq_ind.rq_delete.rq_objname);
```
agrawalravi90: Can preq->rq_ind.rq_delete.rq_objname be NULL?

bhroam: It can't be NULL.  We read the name off the wire and if there is a problem, we will catch it then.

PR #1178, opened 2019-06-23
Use non-blocking write in message_job()

prakashcv13:
#### Describe Bug or Feature
* [PP-1000](https://pbspro.atlassian.net/browse/PP-1000)

When an interactive job sends a lot of spam to stdout/stderr, MoM hangs trying to kill the job (on exceeding walltime).

#### Describe Your Change
When MoM decides to kill the job - it calls message_job() that does a BLOCKING write() to display a message telling why the job is being killed. The file that the MoM writes to is a pseudo-tty (for an interactive job) that is being used by the child pbs_mom (job process) to send data to qsub.
As there is a lot of data being sent - there is a chance that a lot of data is buffered and until all the data is being read by the other side, the MoM process that called message_job is blocked.

As a solution - we use O_NONBLOCK in message_job().

To reproduce - I am using the cgroup freezer subsystem to freeze qsub processes so that the buffering happens. It is not straightforward to do that through PTL - so I am attaching manual test logs.

Test Logs
[after_fix.txt](https://github.com/PBSPro/pbspro/files/3336592/after_fix.txt)
[before_fix.txt](https://github.com/PBSPro/pbspro/files/3336593/before_fix.txt)
[mom_valgrind.txt](https://github.com/PBSPro/pbspro/files/3349310/mom_valgrind.txt)


prakashcv13:
Hi All, This PR is open for review.

src/resmom/requests.c
```diff
@@ -765,15 +783,37 @@ message_job(job *pjob, enum job_file jft, char *text)
 		pstr[len++] = '\n';	/* append new-line */
 		text = pstr;
 	}
-	(void)write(fds, text, len);
 #ifdef	WIN32
+	(void)write(fds, text, len);
 	(void)_commit(fds);
+#else
+	flags = fcntl(fds, F_GETFL, 0);
+	fcntl(fds, F_SETFL, flags | O_NONBLOCK);
```
riyazhakki: fds is already opened with O_NONBLOCK flag, why to set it again?

src/resmom/requests.c
```diff
@@ -765,15 +783,37 @@ message_job(job *pjob, enum job_file jft, char *text)
...
+			}
+		}
+	}
 #endif
 	(void)close(fds);
 	if (pstr)
 		free(pstr);
 
-	return PBSE_NONE;
+	if (total_bytes_written == len)
```
riyazhakki: What if above write() call returns -1 (in case errno=EAGAIN) at least one time in the loop ? Does this condition satisfy on next successful write?

src/resmom/requests.c
```diff
@@ -746,16 +750,30 @@ message_job(job *pjob, enum job_file jft, char *text)
...
 	/* set to append mode */
 	SetFilePointer((HANDLE)_get_osfhandle(fds), (LONG)NULL,
 		(PLONG)NULL, FILE_END);
-#endif
+#else
+	while ((slept < 5) && (fds < 0)) {
+		fds = open_std_file(pjob, jft, O_WRONLY | O_APPEND | O_NONBLOCK,
+	                         pjob->ji_qs.ji_un.ji_momt.ji_exgid);
+		if (fds < 0) {
+			if ( errno == EAGAIN || errno == EWOULDBLOCK) {
```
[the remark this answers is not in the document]
prakashcv13: seteuid() and seteguid() do not set errno to EAGAIN. However, setresuid() and setresgid() do. If any of these calls failed and set errno to EAGAIN, the attempt to open the file was not successful as open() is not called at all. We would need to retry as EAGAIN is a temporary error.
It is my view, it really does not matter which system call failed and set the errno to EAGAIN in this case.

src/resmom/requests.c
```diff
@@ -746,16 +750,30 @@ message_job(job *pjob, enum job_file jft, char *text)
...
+				sleep(1);
```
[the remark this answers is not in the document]
prakashcv13: yes, the mom will be blocked if the errno is set to EAGAIN/EWOULDBLOCK, which is better than being blocked forever. We can use usleep() to sleep for 0.5 seconds, to reduce the amount of time mom will be blocked.

src/resmom/requests.c
prakashcv13: yes.

src/resmom/requests.c
```diff
@@ -765,15 +783,37 @@ message_job(job *pjob, enum job_file jft, char *text)
...
+	(void)write(fds, text, len);
 	(void)_commit(fds);
+#else
+	flags = fcntl(fds, F_GETFL, 0);
+	fcntl(fds, F_SETFL, flags | O_NONBLOCK);
+
+	while ((slept < 5) && (total_bytes_written < len)) {
+		bytes_written = write(fds, text, len-total_bytes_written);
+		total_bytes_written += bytes_written;
+		text += bytes_written;
```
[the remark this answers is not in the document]
prakashcv13: how?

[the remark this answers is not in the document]
riyazhakki: On write failure bytes_written will have -1 and the line 796 'text += bytes_written' will move text pointer to one location backward.

prakashcv13:
@riyazhakki - I have addressed your review comment, please take a look.

src/resmom/requests.c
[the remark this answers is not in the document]
mike0042: I tried using WORK_Immed to instantiate the PMIx server process and noticed a delay of around a second before the work task actually started. This was unacceptable for a rapid-fire launch scenario. The call to open_std_file() should return success the first time in most cases. I don't think we need to try to open the file more than three times, pausing 0.25 seconds between attempts for a total of 0.5 seconds of sleep. Also, this should be a for loop to instantiate and increment slept.

src/resmom/requests.c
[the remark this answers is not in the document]
arungrover: I disagree, we are trying to find a problem in one call and checking errno which can potentially be set by another system call.
In function open_file_as_user() open is called before calling setresuid/setresgid. This means if open fails for some reason you do not want to retry, this code might end up retrying even in that case because it will be working off of errno set by setresuid/setresgid calls.

prakashcv13:
@mike0042 and @arungrover - I have tried to address your comments, requesting another look.

src/resmom/requests.c
```diff
@@ -757,16 +756,18 @@ message_job(job *pjob, enum job_file jft, char *text)
...
+	unsigned int usecs = 250 * 1000; /* 250 milliseconds */
+	for (i = 0; i < 3; i++) {
 		fds = open_std_file(pjob, jft, O_WRONLY | O_APPEND | O_NONBLOCK,
 	                         pjob->ji_qs.ji_un.ji_momt.ji_exgid);
-		if (fds < 0) {
-			if (errno == EAGAIN || errno == EWOULDBLOCK) {
-				sleep(1);
-				slept++;
-			} else
+		if (fds > 0)
```
arungrover: Was this change intentional? If fd is > 0 then that would mean open() succeeded but we will still retry if errno is set to EAGAIN or EWOULDBLOCK.

prakashcv13: not intentional. I have pushed the correct changes now.

src/resmom/start_exec.c
```diff
@@ -6379,18 +6379,23 @@ starter_return(int upfds, int downfds, int code, struct startjob_rtn *sjrtn)
...
+	int open_errno = 0; 
 	extern gid_t pbsgroup;
 
 	/* must open or create file as the user */
 
 	if (impersonate_user(exuid, exgid) == -1)
 		return -1;
 
-	fds = open(path, oflag, mode);
+	if ((fds = open(path, oflag, mode)))
```
arungrover: Please set errno only when open returns -1

src/resmom/requests.c
```diff
@@ -793,6 +793,8 @@ message_job(job *pjob, enum job_file jft, char *text)
 				usleep(usecs);
 			else {
 				(void)close(fds);
+				if (pstr)
+					free(pstr);
```
mike0042: You have properly initialized pstr to NULL when you declared it, so no need to check for NULL. Just call free().

src/resmom/start_exec.c
prakashcv13: done

src/resmom/requests.c
prakashcv13: done.

prakashcv13:
Hi @arungrover and @mike0042 - Thank you for the comments, I have addressed the latest review comments, requesting another look.

src/resmom/requests.c
```diff
@@ -746,34 +747,71 @@ message_job(job *pjob, enum job_file jft, char *text)
...
 		(void)strcpy(pstr, text);
 		pstr[len++] = '\n';	/* append new-line */
 		text = pstr;
 	}
-	(void)write(fds, text, len);
 #ifdef	WIN32
+	(void)write(fds, text, len);
 	(void)_commit(fds);
+#else
+	for (i = 0; i < 5; i++) {
```
mike0042: Final comment... I think this should be three tries as well.

prakashcv13: ok - let us have only three tries...

prakashcv13:
thank you @arungrover and @mike0042 for the review and approvals, shall we take this PR forward.

PR #1179, opened 2019-06-25
Fixed memory leak when a job is preempted multiple times

bhroam:
#### Describe Bug or Feature
There is a memory leak in the scheduler during preemption.  The following has to happen to trigger this memory leak:
1) Job A is preempted for high priority job H1
2) H1 fails to run
3) SInce H1 fails to run, A  is run again
4) In the same cycle, we preempt A again for H2.

In the process of preempting A the first time, we set a couple of job members.  These are not cleared when we run the job again.  When we preempt the job the second time, we write the same members again leaking the memory from the first.

#### Describe Your Change
The members mentioned above were for RRTROS (not releasing all resources on suspend).
I did the following fixes:
1) When a job is run and the RRTROS members are set, free them and NULL them.
2) If RRTROS is not set, don't set the RRTROS members on the preempted job.

Random fix: I removed a prolific log message that provided little benefit.

[valgrind-before.log](https://github.com/PBSPro/pbspro/files/3327563/valgrind-before.log)
[valgrind-after.log](https://github.com/PBSPro/pbspro/files/3327562/valgrind-after.log)
Look for the memory leaked from dup_job_info() in the before log and for it not being there in the after.
[preemption.log](https://github.com/PBSPro/pbspro/files/3327567/preemption.log)

### DIFF
--- a/src/scheduler/job_info.c
+++ b/src/scheduler/job_info.c
@@ -2774,10 +2774,6 @@ struct preempt_ordering *schd_get_preempt_order(resource_resv *resresv)
 	if (get_job_req_used_time(resresv, &req, &used) != 0)
 		return NULL;
 
-	if (req == -1 || used == -1)
-		schdlog(PBSEVENT_JOB, PBS_EVENTCLASS_JOB, LOG_INFO, resresv->name,
-				"No walltime/cput to determine percent of time left - will use first preempt order");
-
 	po = get_preemption_order(conf.preempt_order, req, used);
 
 	return po;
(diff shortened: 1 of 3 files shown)


src/scheduler/job_info.c
```diff
@@ -3247,15 +3243,15 @@ find_jobs_to_preempt(status *policy, resource_resv *hjob, server_info *sinfo, in
 
 		po = schd_get_preempt_order(pjob);
 		if (po != NULL) {
-			if (po->order[0] == PREEMPT_METHOD_SUSPEND && pjob->job->can_suspend) {
+			if (policy->rel_on_susp != NULL && po->order[0] == PREEMPT_METHOD_SUSPEND && pjob->job->can_suspend) {
```
arungrover: You need a similar change in find_and_preempt_jobs() too. This is only happening in the simulated universe.

PR #1180, opened 2019-06-27
Sched refactor to only query necessary attribs

agrawalravi90:
#### Describe Bug or Feature

Right now, scheduler queries all attributes of jobs when querying the universe every cycle, but only uses a limited set from them. So, there's time and memory wasted in packaging & sending unnecessary data over the wire, so we should refactor the code to only query the attributes that it needs.

#### Describe Your Change

Just added a list of attributes to query in the selstat request.

Since this is just refactoring, smoke testing should be enough. We start seeing performance benefits of this when there are more than 300k jobs in the queue, so it's not suitable to write a performance test for this.

### DIFF
--- a/src/scheduler/job_info.c
+++ b/src/scheduler/job_info.c
@@ -521,7 +521,7 @@ query_jobs(status *policy, int pbs_sd, queue_info *qinfo, resource_resv **pjobs,
 	struct attropl opl = { NULL, ATTR_q, NULL, NULL, EQ };
 	static struct attropl opl2[2] = { { &opl2[1], ATTR_state, NULL, "Q", EQ},
 		{ NULL, ATTR_array, NULL, "True", NE} };
-	struct attrl *attrib = NULL;
+	static struct attrl *attrib = NULL;
 	int i;
 
 	/* linked list of jobs returned from pbs_selstat() */
@@ -609,14 +609,16 @@ query_jobs(status *policy, int pbs_sd, queue_info *qinfo, resource_resv **pjobs,
 
 	server_time = qinfo->server->server_time;
 
-	for (i = 0; jobattrs[i] != NULL; i++) {
-		struct attrl *temp_attrl = NULL;
+	if (attrib == NULL) {
+		for (i = 0; jobattrs[i] != NULL; i++) {
+			struct attrl *temp_attrl = NULL;
 
-		temp_attrl = new_attrl();
-		temp_attrl->name = strdup(jobattrs[i]);
-		temp_attrl->next = attrib;
-		temp_attrl->value = "";
-		attrib = temp_attrl;
+			temp_attrl = new_attrl();
+			temp_attrl->name = strdup(jobattrs[i]);
+			temp_attrl->next = attrib;
+			temp_attrl->value = "";
+			attrib = temp_attrl;
+		}
 	}
 
 	/* get jobs from PBS server */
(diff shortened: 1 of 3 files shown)


src/scheduler/job_info.c
```diff
@@ -521,6 +521,8 @@ query_jobs(status *policy, int pbs_sd, queue_info *qinfo, resource_resv **pjobs,
 	struct attropl opl = { NULL, ATTR_q, NULL, NULL, EQ };
 	static struct attropl opl2[2] = { { &opl2[1], ATTR_state, NULL, "Q", EQ},
 		{ NULL, ATTR_array, NULL, "True", NE} };
+	struct attrl *attrib = NULL;
```
arungrover: Why don't we make this a static list and reuse it every time scheduler queries jobs?

src/scheduler/job_info.c
```diff
@@ -573,8 +609,18 @@ query_jobs(status *policy, int pbs_sd, queue_info *qinfo, resource_resv **pjobs,
...
 	server_time = qinfo->server->server_time;
 
+	for (i = 0; jobattrs[i] != NULL; i++) {
+		struct attrl *temp_attrl = NULL;
+
+		temp_attrl = new_attrl();
+		temp_attrl->name = strdup(jobattrs[i]);
+		temp_attrl->next = attrib;
+		temp_attrl->value = "";
+		attrib = temp_attrl;
```
arungrover: You are not freeing the members of the list. You may not have to do that if you make attrib a static link list.

test/tests/performance/pbs_sched_perf.py
```diff
@@ -216,3 +216,33 @@ def test_pset_fuzzy_perf(self):
...
+    def test_sched_query(self):
+        """
+        Test scheduler's turnaround time with focus on querying the universe
+        """
+        self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'False'})
+
+        # We have 10010 vnodes, submit 10050 jobs so that it can run almost
+        # all of them
+        a = {'Resource_List.select': '1:ncpus=1'}
+        jids = self.submit_jobs(attribs=a, num=10050, step=0, wt_start=900)
```
arungrover: I think we should submit jobs with the environment so that the job size becomes big enough that it makes a difference querying limited attributes.

test/tests/performance/pbs_sched_perf.py
```diff
@@ -216,3 +216,33 @@ def test_pset_fuzzy_perf(self):
...
+
+        # Kick a sched cycle which will run 1010 jobs
+        self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'True'})
+        self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'False'})
+        jid1010 = jids[1009]
+        self.server.expect(JOB, {"job_state": "R"}, jid1010)
```
bhroam: Use a job_state=R count instead of checking for a specific job.  You're making an assumption that the scheduler chose all the jobs before that.  It should, but you never know what might change in the future.

src/scheduler/job_info.c
```diff
@@ -573,8 +609,18 @@ query_jobs(status *policy, int pbs_sd, queue_info *qinfo, resource_resv **pjobs,
 
 	server_time = qinfo->server->server_time;
 
+	for (i = 0; jobattrs[i] != NULL; i++) {
```
bhroam: Why don't you put this code in a helper function and then call it for nodes as well.

Also, I don't see you freeing the list.

agrawalravi90: I think I'll make it a static list as Arun suggested, that way we don't have to recreate it every cycle. But I do like the idea of also doing this for nodes, so I'll add similar code in query_nodes.

agrawalravi90:
@bhroam and @arungrover since the performance boost from this isn't significant unless we have 300k+ jobs in the queue, a performance test for this change isn't practical. So, I've removed the one I'd written, which wasn't showing much benefit because there were only 10k jobs in the queue.

Please resume review for this and let me know if changes look ok, sorry for the delay on this.

agrawalravi90:
Thanks Bhroam! @arungrover please review this whenever you get a chance

agrawalravi90:
@arungrover gentle reminder, thanks.

PR #1182, opened 2019-06-28
Fix clang static analyzer warnings

bhroam:
#### Describe Bug or Feature
Clang (alternate compiler to gcc) has a static analyzer.  I ran it on PBS and it came up with many warnings.  I fixed two types: Possible memory leaks and possible NULL pointer dereferences.

Mostly the possible memory leaks were in error cases where you needed to clean up, and some variables weren't being freed.
Possible NULL pointer dereferences usually were some strange path through the code that could lead to a pointer being NULL and being dereferenced.

There is a 3rd I fixed here and there (but not all of them): Variables being set and never used.

For testing I am currently running tests through an internal test harness tool.  It takes time, so I thought I'd put the PR out for review before it is done.

### DIFF
--- a/src/cmds/qmgr.c
+++ b/src/cmds/qmgr.c
@@ -4225,6 +4225,9 @@ parse_request(char *request, char ***req)
 		fprintf(stderr, "malloc failure (errno %d)\n", errno);
 		exit(1);
 	}
+	for (i = IND_FIRST; i <= IND_LAST; i++)
+		(*req)[i] = NULL;
+		
 	for (i = 0; !EOL(*foreptr) && i < MAX_REQ_WORDS && error == 0;) {
 		while (White(*foreptr))
 			foreptr++;
--- a/src/cmds/qmgr.c
+++ b/src/cmds/qmgr.c
@@ -656,7 +656,7 @@ attrlist_add(struct attropl  **attrlist, char *attname,
 	*attrlist = paol;
 
 	ltxt = attname_len;
-	Mstring(paol->name, ltxt+1);
+	Mstring(paol->name, ltxt + 1);
 	strncpy(paol->name, attname, ltxt);
 	paol->name[ltxt] = '\0';
 
@@ -2641,7 +2641,7 @@ execute(int aopt, int oper, int type, char *names, struct attropl *attribs)
 			}
 
 			sp = sname->svr;
-			if ((oper == MGR_CMD_LIST)) {
+			if (oper == MGR_CMD_LIST) {
 				sa = attropl2attrl(attribs);
 				switch (type) {
 					case MGR_OBJ_SERVER:
@@ -2685,7 +2685,7 @@ execute(int aopt, int oper, int type, char *names, struct attropl *attribs)
 
 				pbs_statfree(ss);
 			}
-			else if ((oper == MGR_CMD_PRINT)) {
+			else if (oper == MGR_CMD_PRINT) {
 
 				sa = attropl2attrl(attribs);
 				switch (type) {
@@ -4225,10 +4225,6 @@ parse_request(char *request, char ***req)
 		fprintf(stderr, "malloc failure (errno %d)\n", errno);
 		exit(1);
 	}
-	for (i=IND_FIRST;i<=IND_LAST;i++) {
-		(*req)[i] = '\0';
-	}
-
 	for (i = 0; !EOL(*foreptr) && i < MAX_REQ_WORDS && error == 0;) {
 		while (White(*foreptr))
 			foreptr++;
@@ -4245,15 +4241,14 @@ parse_request(char *request, char ***req)
 			pstderr("qmgr: max word length exceeded\n");
 			CaretErr(request, chars_parsed);
 		}
-		if (len != 0) {
-			(*req)[i] = (char *) malloc(len + 1);
-			if ((*req)[i] == NULL) {
-				fprintf(stderr, "malloc failure (errno %d)\n", errno);
-				exit(1);
-			}
-			((*req)[i])[len] = '\0';
-			strncpy((*req)[i], backptr, len);
+		(*req)[i] = (char *)malloc(len + 1);
+		if ((*req)[i] == NULL) {
+			fprintf(stderr, "malloc failure (errno %d)\n", errno);
+			exit(1);
 		}
+		((*req)[i])[len] = '\0';
+		if (len > 0)
+			strncpy((*req)[i], backptr, len);
 		i++;
 
 	}
--- a/src/lib/Libcmds/pbs_json.c
+++ b/src/lib/Libcmds/pbs_json.c
@@ -203,7 +203,7 @@ free_json_node(JsonNode *node)
 
 /**
  * @brief
- *	frees a json node
+ *	frees the json node list
  *
  * @return	Void
  *
--- a/src/lib/Libcmds/pbs_json.c
+++ b/src/lib/Libcmds/pbs_json.c
@@ -35,6 +35,7 @@
 * trademark licensing policies.
 */
 
+#include <stdlib.h>
 #include <ctype.h>
 #include "pbs_json.h"
 #include "libutil.h"
@@ -244,8 +245,10 @@ add_json_node(JsonNodeType ntype, JsonValueType vtype, JsonEscapeType esc_type, 
 	if (node->value_type == JSON_STRING) {
 		if (value != NULL) {
 			ptr = strdup_escape(esc_type, value);
-			if (ptr == NULL)
+			if (ptr == NULL) {
+				free(node);
 				return NULL;
+			}
 		}
 		node->value.string = ptr;
 	}
--- a/src/lib/Libcmds/pbs_json.c
+++ b/src/lib/Libcmds/pbs_json.c
@@ -184,6 +184,45 @@ strdup_escape(JsonEscapeType esc_type, const char *str)
 }
 
 /**
+ * @brief free an individual json node
+ * @param[in] node - node to free
+ * 
+ * @return void
+ */
+void
+free_json_node(JsonNode *node)
+{
+	if ((node->value_type == JSON_STRING) || (node->value_type == JSON_NUMERIC)) {
+		if (node->value.string != NULL)
+			free(node->value.string);
+	}
+	if (node->key != NULL)
+		free(node->key);
+	free(node);
+}
+
+/**
+ * @brief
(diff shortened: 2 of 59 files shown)


bhroam:
I'm a little confused why Travis failed.  I ran the smoke test on my machine and it passed without incident.  I'll figure it out.

src/lib/Libifl/pbsD_selectj.c
```diff
@@ -97,8 +97,10 @@ __pbs_selectjob(int c, struct attropl *attrib, char *extend)
 		ret = PBSD_select_get(c);
 
 	/* unlock the thread lock and update the thread context data */
-	if (pbs_client_thread_unlock_connection(c) != 0)
+	if (pbs_client_thread_unlock_connection(c) != 0) {
+		free(ret);
```
agrawalravi90: ret is a char **, should we be calling free_string_array() on it?

src/lib/Libpython/pbs_python_svr_size_type.c
```diff
@@ -587,20 +587,13 @@ pps_size_number_methods_subtract(PyObject *left, PyObject *right)
 
 		if (rc != 0)
 			goto QUIT;
-#ifdef NAS /* localmod 005 */
```
agrawalravi90: why did you remove this NAS mod?

src/lib/Libtpp/tpp_client.c
```diff
@@ -89,6 +89,7 @@
 #include <sys/time.h>
 #ifndef WIN32
 #include <stdint.h>
+#include <stdlib.h>
```
agrawalravi90: you included this header file, but you didn't really make any other change to the file, so is this change needed?

src/lib/Libtpp/tpp_router.c
```diff
@@ -288,6 +286,7 @@ send_leaves_to_router(tpp_router_t *parent, tpp_router_t *target)
 		free(lf_data->addrs);
 		free(lf_data);
 	}
+	free(pkey);
```
agrawalravi90: why did you move this? pkey is not really used in the while block right?

src/scheduler/check.c
```diff
@@ -505,15 +505,15 @@ shrink_to_run_event(status *policy, server_info *sinfo,
 			/* If there are no events left to check or if we have reached the front of event list or
 			 * if the event is falling before min end time, break.
 			 */
-			if ((te == NULL && last_skipped_event == NULL) || te == initial_event || te->event_time < min_end_time)
-				break;
-			/* If no events in this segment, then try last skipped event of the previous segment */
-			else if (te == NULL) {
+			if (te == NULL)
```
agrawalravi90: so we don't wanna break if te == NULL && last_skipped_event == NULL ?

src/lib/Libecl/ecl_verify.c
```diff
@@ -355,15 +355,13 @@ verify_an_attribute(int batch_request, int parent_object, int cmd,
...
-	if ((err_code !=0) && (*err_msg == NULL)) {
+	if ((err_code != 0) && (*err_msg == NULL)) {
 		/* find err_msg and update it */
 		p = pbse_to_txt(err_code);
 		if (p) {
 			*err_msg = strdup(p);
-			if (*err_msg == NULL) {
-				err_code = PBSE_SYSTEM;
+			if (*err_msg == NULL)
 				return -1;
```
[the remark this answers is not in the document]
bhroam: The doxygen comment says the function returns -1 when it can't allocate memory.  I'd rather leave it alone since if we return PBSE_SYSTEM, the caller might get confused that something is in err_msg and dereference it.

src/lib/Libecl/pbs_client_thread.c
```diff
@@ -639,6 +644,10 @@ __pbs_client_thread_init_thread_context(void)
 	 * to read the error code out of it.
 	 */
 	pbs_client_thread_set_single_threaded_mode();
+	if (free_ptr) {
```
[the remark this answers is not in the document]
bhroam: I don't know much about the pthread stuff, but what I assuming what is happening is at line 617 we call pthread_set_specific() with ptr.  I'm assuming this associates ptr with a thread since we don't free ptr in the success case.  There are some times we goto err after the call to pthread_set_specific().  We can't just check if ptr is NULL, because we don't know if we got there before or after the call to pthread_set_specific().  You're more of an expert on pthreads than I am, so please let me know if I am wrong.

src/server/req_jobobit.c
```diff
@@ -2206,6 +2208,7 @@ job_obit(struct resc_used_update *pruu, int stream)
 					svr_setjobstate(pjob, JOB_STATE_HELD, JOB_SUBSTATE_HELD);
 					FREE_RUU(pruu)
 					free(mailbuf);
+					free(acctbuf);
```
arungrover: Same change is needed for "case JOB_EXEC_QUERST:" as well.

src/lib/Libifl/pbsD_selectj.c
bhroam: This isn't a NULL ptr terminated string array.  PBSD_select_get() will allocate enough memory for all the strings in one go.  We only need to free it.  It does look weird though.  I'll add a code comment.

src/lib/Libpython/pbs_python_svr_size_type.c
bhroam: NAS mod 5 has to do with removing warnings from the Intel compiler.  I didn't remove their code, I removed ours.  Since the intel compiler is a bit more pedantic than gcc, following what it wants should be fine for gcc.  I removed the mod because it cleaned up the code a bit.

src/lib/Libtpp/tpp_client.c
bhroam: clang complained about the return value of malloc() being converted into an int and back to a pointer.  Since malloc() was being used, this header should have been included.

src/lib/Libtpp/tpp_router.c
bhroam: I moved it since it could cause a double free before.  It was freed on line 274.  On line 284 we goto err.  In err we free pkey.  I just moved pkey down past the goto err so we didn't have a chance to double free.  Aren't static analyzers grand?  They find the weirdest esoteric paths through the code that might cause a problem :)

src/scheduler/check.c
bhroam: Clang complained about the end of that if statement.  If te == NULL last_skipped_event != NULL and te->event_time.  That last part would dereference and crash.  You are right that swapping the order of the two if/else cases is wrong by itself.  That first part would never trigger.  I broke the condition into two parts.  It should work better now.

src/cmds/qmgr.c
```diff
@@ -4245,15 +4244,14 @@ parse_request(char *request, char ***req)
...
 		}
-		if (len != 0) {
-			(*req)[i] = (char *) malloc(len + 1);
-			if ((*req)[i] == NULL) {
-				fprintf(stderr, "malloc failure (errno %d)\n", errno);
-				exit(1);
-			}
-			((*req)[i])[len] = '\0';
-			strncpy((*req)[i], backptr, len);
+		(*req)[i] = (char *)malloc(len + 1);
```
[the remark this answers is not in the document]
bhroam: Clang complained about line 4228.  The code was assigning a null byte into a pointer.  I'm assuming the original author wanted their strings to be initialized as null strings and messed up what they were doing.  Code wise, I didn't really do anything at line 4228 except use the correct symbol when initializing the variable (NULL == 0 and '\0' == 0).  I took it one step farther and made the initial good on my initial assumption.  Now if len == 0, we'll allocate 1 byte and shove a '\0' into it.

src/lib/Libecl/pbs_client_thread.c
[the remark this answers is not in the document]
agrawalravi90: Oh, my bad, I hadn't seen the call to pthread_setspecific. You are right, we should not be freeing it after pthread_setspecific call until we destroy the thread itself.

src/lib/Libtpp/tpp_router.c
agrawalravi90: ah, ok, that makes sense. Just to be safe, can we set pkey to NULL after the free() ? Just in case somebody changes this code in the future.
Static analyzers are grand indeed.

src/scheduler/check.c
agrawalravi90: I'm not sure I'm following you. Couldn't you just change the last condition from (te->event_time < min_end_time) to (te != NULL && te->event_time < min_end_time) ? Or do we want to change the behavior so that we won't break out of the loop if (te == NULL && last_skipped_event == NULL) ? Because with your changes that won't happen.

src/lib/Libutil/misc_utils.c
```diff
@@ -1464,7 +1464,7 @@ show_nonprint_chars(char *str)
 		return str;
 
 	nsize = (strlen(str) * 2) + 1;
-	if (nsize > locbuf_size) {
+	if (nsize > locbuf_size || locbuf == NULL) {
```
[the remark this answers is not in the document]
bhroam: It's to shut up the clang analyzer.  It doesn't know we're going to fall in there the first time and allocate memory.  I could remove it, but then the next time we do this exercise, we'll see it again.

src/scheduler/check.c
bhroam: I think I fixed this without changing behavior.  I am not too familiar with this function and it is very complex.  I'd rather not make changes that might break things.  I left the if/else flipped, and added a check if last_skipped_event == NULL inside the check for if (te == NULL).   This basically broke the single break in to two.

src/lib/Libtpp/tpp_router.c
bhroam: I don't really think it is necessary.  We return immediately after freeing pkey.  If someone modifies the code, I think it is pretty obvious that they shouldn't use pkey.  It is freed right above a return.

src/lib/Libcmds/pbs_json.c
```diff
@@ -183,6 +183,45 @@ strdup_escape(JsonEscapeType esc_type, const char *str)
...
+			free(node->value.string);
+	}
+	if (node->key != NULL)
+		free(node->key);
+	free(node);
+}
+
+/**
+ * @brief
+ *	frees a json node
```
arungrover: This comment would be "frees a json node list"

PR #1188, opened 2019-07-04
queues have their own has_user/grp/proj_limits flags

arungrover:
#### Describe Bug or Feature
This is a performance optimization.
Scheduler up until now used to check if there is any limit (user/group/proj) is set on queue and if so, it used to set flags on server object that there is a limit (user/group/proj) set in the complex.
This means that scheduler used to check limits on server and all the queues even when the limit was only set on one (or some) queues.
Also, while ending (or preempting) a job scheduler used to check if there is any soft limit set on server or queue object of the job that is ending and used to call set_preempt_prio() on all the jobs with matching user/group/project name. Since all the jobs in pbs by default have the same project name, the check was turning out to be true almost all the time. Checking and updating preempt priority on all the jobs was an expensive operation and takes a lot of time.

#### Describe Your Change
In this change, every queue will keep track of what kind of limit (user/group/proj) is set on it and will not update server object to show that this is a complex-wide limit.
This gives us more control in our limit checking code to return early in each function if a relevant limit is not set.
Another change is that when a job ends scheduler will now check what kind of limit is set on server/queue and then accordingly match the entity (user/group/proj) of the ending job to other jobs and update their preempt priority.

Setting the type of limit on queues also resulted in a change in job equivalence classes. Earlier any limit set on any queue used to result into setting limit flags on the server object, which would then make equivalence classes code used to consider user/group/project of the job into consideration while finding/creating an equivalence class (even when the limit was neither set on the server or the queue these jobs were in). With this change, these entities (user/group/project) are only considered when job's queue (or server) has the limit set on it.

#### Link to Design Doc
NA

[test_equiv_classes.txt](https://github.com/PBSPro/pbspro/files/3357195/test_equiv_classes.txt)
[test_perf.txt](https://github.com/PBSPro/pbspro/files/3357196/test_perf.txt)
[test_perf_old.txt](https://github.com/PBSPro/pbspro/files/3374226/test_perf_old.txt)

### DIFF
--- a/src/scheduler/data_types.h
+++ b/src/scheduler/data_types.h
@@ -72,6 +72,7 @@ struct node_info;
 struct job_info;
 struct schd_resource;
 struct resource_req;
+struct resource_count;
 struct holiday;
 struct prev_job_info;
 struct group_info;
@@ -106,6 +107,7 @@ typedef struct job_info job_info;
 typedef struct node_info node_info;
 typedef struct schd_resource schd_resource;
 typedef struct resource_req resource_req;
+typedef struct resource_count resource_count;
 typedef struct usage_info usage_info;
 typedef struct group_info group_info;
 typedef struct prev_job_info prev_job_info;
@@ -372,6 +374,7 @@ struct server_info
 	resresv_set **equiv_classes;
 	node_bucket **buckets;		/* node bucket array */
 	node_info **unordered_nodes;
+	int preempt_bit;		/* Overall preempt bit to mark server is over max run softlimits */
 #ifdef NAS
 	/* localmod 049 */
 	node_info **nodes_by_NASrank;	/* nodes indexed by NASrank */
@@ -439,6 +442,7 @@ struct queue_info
 	int num_topjobs;	/* current number of top jobs in this queue */
 	int backfill_depth;	/* total allowable topjobs in this queue*/
 	char *partition;	/* partition to which queue belongs to */
+	int preempt_bit;	/* Overall preempt bit to mark queue is over max run softlimits */
 };
 
 struct job_info
@@ -739,7 +743,6 @@ struct resource_type
 	unsigned is_time:1;
 };
 
-
 struct schd_resource
 {
 	char *name;			/* name of the resource - reference to the definition name */
@@ -796,8 +799,18 @@ struct counts
 {
 	char *name;		/* name of entitiy */
 	int running;		/* count of running jobs in object */
-	resource_req *rescts;	/* resources used */
+	int preempt_bit;	/* Place to store preempt bit if entity is over limits */
+	resource_count *rescts;	/* resources used */
 	counts *next;
+};
+
+struct resource_count
+{
+	char *name;		    /* resource name */
+	resdef *def;		    /* definition of resource */
+	sch_resource_t amount;	    /* amount of resource used */
+	int preempt_bit;	    /* Place to store preempt bit if resource of an entity is over limits */
+	struct resource_count *next;
 };
 
 /* global data types */
--- a/src/scheduler/data_types.h
+++ b/src/scheduler/data_types.h
@@ -374,7 +374,7 @@ struct server_info
 	resresv_set **equiv_classes;
 	node_bucket **buckets;		/* node bucket array */
 	node_info **unordered_nodes;
-	int preempt_bit;		/* Overall preempt bit to mark server is over max run softlimits */
+	int soft_limit_preempt_bit;	/* Overall preempt bit to mark server is over max run softlimits */
 #ifdef NAS
 	/* localmod 049 */
 	node_info **nodes_by_NASrank;	/* nodes indexed by NASrank */
@@ -438,11 +438,11 @@ struct queue_info
 	char **node_group_key;		/* node grouping resources */
 	struct node_partition **nodepart; /* array pointers to node partitions */
 	struct node_partition *allpart;   /* partition w/ all nodes assoc with queue*/
-	int num_parts;		/* number of node partitions(node_group_key) */
-	int num_topjobs;	/* current number of top jobs in this queue */
-	int backfill_depth;	/* total allowable topjobs in this queue*/
-	char *partition;	/* partition to which queue belongs to */
-	int preempt_bit;	/* Overall preempt bit to mark queue is over max run softlimits */
+	int num_parts;			/* number of node partitions(node_group_key) */
+	int num_topjobs;		/* current number of top jobs in this queue */
+	int backfill_depth;		/* total allowable topjobs in this queue*/
+	char *partition;		/* partition to which queue belongs to */
+	int soft_limit_preempt_bit;	/* Overall preempt bit to mark queue is over max run softlimits */
 };
 
 struct job_info
@@ -797,10 +797,10 @@ struct mom_res
 
 struct counts
 {
-	char *name;		/* name of entitiy */
-	int running;		/* count of running jobs in object */
-	int preempt_bit;	/* Place to store preempt bit if entity is over limits */
-	resource_count *rescts;	/* resources used */
+	char *name;			/* name of entitiy */
+	int running;			/* count of running jobs in object */
+	int soft_limit_preempt_bit;	/* Place to store preempt bit if entity is over limits */
+	resource_count *rescts;		/* resources used */
 	counts *next;
 };
 
@@ -809,7 +809,7 @@ struct resource_count
 	char *name;		    /* resource name */
 	resdef *def;		    /* definition of resource */
 	sch_resource_t amount;	    /* amount of resource used */
-	int preempt_bit;	    /* Place to store preempt bit if resource of an entity is over limits */
+	int soft_limit_preempt_bit; /* Place to store preempt bit if resource of an entity is over limits */
 	struct resource_count *next;
 };
 
--- a/src/scheduler/data_types.h
+++ b/src/scheduler/data_types.h
@@ -393,7 +393,10 @@ struct queue_info
 	unsigned has_soft_limit:1;	/* queue has a soft user/grp limit set */
 	unsigned has_hard_limit:1;	/* queue has a hard user/grp limit set */
 	unsigned is_peer_queue:1;	/* queue is a peer queue */
-	unsigned has_resav_limit:1;		/* queue has resources_available limits */
+	unsigned has_resav_limit:1;	/* queue has resources_available limits */
+	unsigned has_user_limit:1;	/* queue has user hard or soft limit */
+	unsigned has_grp_limit:1;	/* queue has group hard or soft limit */
+	unsigned has_proj_limit:1;	/* queue has project hard or soft limit */
 	struct server_info *server;	/* server where queue resides */
 	char *name;			/* queue name */
 	state_count sc;			/* number of jobs in different states */
--- a/src/scheduler/fifo.c
+++ b/src/scheduler/fifo.c
@@ -458,12 +458,18 @@ init_scheduling_cycle(status *policy, int pbs_sd, server_info *sinfo)
 	 * 2. we need all the jobs to be created and up to date for soft run limits
 	 */
 
+	/* Before setting preempt priorities on all jobs, make sure that entity's preempt bit
+	 * is updated for all running jobs
+	 */
+	if ((sinfo->running_jobs != NULL) && (policy->preempting)) {
+		for (i = 0; sinfo->running_jobs[i] != NULL; i++)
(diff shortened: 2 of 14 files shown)


agrawalravi90:
Can you please post "before" and "after" performance numbers to show how the performance has improved?

agrawalravi90:
> Also, while ending (or preempting) a job scheduler used to check if there is any soft limit set on server or queue object of the job that is ending and used to call set_preempt_prio() on all the jobs with matching user/group/project name.

I didn't understand how this has changed. You mentioned that the code will now check for the type of limit set on the job that's ending/being preempted. Can you please explain how that will avoid the update of all jobs with the same project ?

arungrover:
> > Also, while ending (or preempting) a job scheduler used to check if there is any soft limit set on server or queue object of the job that is ending and used to call set_preempt_prio() on all the jobs with matching user/group/project name.
> 
> I didn't understand how this has changed. You mentioned that the code will now check for the type of limit set on the job that's ending/being preempted. Can you please explain how that will avoid the update of all jobs with the same project ?

So scheduler will still call set_preempti_prio for all the job that matches the entity but now the 16 limits function that set_preempt_prio calls inside will return early by checking if the relevant limit if set on server/queue. Does this make sense or have I confused you more :)

src/scheduler/limits.c
```diff
@@ -1203,6 +1203,9 @@ check_server_max_user_run(server_info *si, queue_info *qi, resource_resv *rr,
 	if ((si == NULL) || (user == NULL) || (sc == NULL))
 		return (SCHD_ERROR);
 
+	if (si->has_user_limit != 1)
```
bhroam: This is a bit, use a !, don't check if it is not equal to 1. (and other places in the PR)

test/tests/functional/pbs_equiv_classes.py
```diff
@@ -390,6 +390,39 @@ def test_user_queue(self):
...
+        self.server.manager(MGR_CMD_SET, QUEUE,
+                            {'max_run': '[u:PBS_GENERIC=4]'}, id='workq')
+
+        # Eat up all the resources, this job will make first equiv class
+        a = {'Resource_List.select': '1:ncpus=8'}
+        J = Job(TEST_USER, attrs=a)
+        self.server.submit(J)
+
+        # Create a new queue and submit jobs to this queue
+        a = {'queue_type': 'e', 'started': 't', 'enabled': 't'}
```
bhroam: Spell out execution and True (and other places)

test/tests/functional/pbs_equiv_classes.py
```diff
@@ -390,6 +390,39 @@ def test_user_queue(self):
...
+        jids1 = self.submit_jobs(3, user=TEST_USER, attrs=a)
+        jids2 = self.submit_jobs(3, user=TEST_USER2, attrs=a)
+
+        self.server.manager(MGR_CMD_SET, SERVER,
+                            {'scheduling': 'True'})
+
+        # Two equivalence classes.  One for the resource eating job and
+        # one for all jobs in workq2.
+        self.scheduler.log_match("Number of job equivalence classes: 2",
+                                 starttime=self.t)
```
bhroam: This test works as described, but I think you should take it one step further.  Submit another job to workq as a different user and check for 3 equiv classes.

test/tests/performance/pbs_preemptperformance.py
```diff
@@ -106,7 +96,7 @@ def create_workload_and_preempt(self):
         sched_on = {'scheduling': 'True'}
         self.server.manager(MGR_CMD_SET, SERVER, sched_on)
 
-        self.server.expect(JOB, {'job_state=R': 1590},
+        self.server.expect(JOB, {'substate=42': 1590},
```
bhroam: This bug has been fixed and merged.  You can go back to using job_state=R

test/tests/performance/pbs_preemptperformance.py
```diff
@@ -513,29 +504,86 @@ def test_preemption_with_soft_limits(self):
...
+        self.server.manager(MGR_CMD_SET, SCHED, a)
+
+        a = {'max_run_res_soft.ncpus': "[u:" + str(TEST_USER)+"=1]"}
+        self.server.manager(MGR_CMD_SET, QUEUE, a, 'workq')
+        self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'False'})
+
+        # submit a bunch of jobs as different users
+        a = {ATTR_l + '.select=1:ncpus': 1}
+        usr_list = [TEST_USER, TEST_USER2, TEST_USER3, TEST_USER4]
+        num_usr = 4
```
bhroam: Instead of 4, use len(usr_list)

arungrover:
@agrawalravi90 , @bhroam  I have changed the implementation to store the preempt bits on the entities themselves (which are part of counts structure). These entities are updated when check_* functions are called and then in set_preempt_prio instead of calling check_* soft limit functions scheduler now only picks relevant preempt_bit from entities job is part of and use that for setting the preempt value on the job.
This approach as sped up two specific tests related to soft limits in the preempt performance test suite by almost by 50+ seconds. The tests still take atleast 120+ seconds to preempt 2000 jobs and run the high priority job but then most of this time is spent in server to suspend 2000 jobs (more than 70 to 80 seconds).
Please have a look at the change and provide comments.

arungrover:
[test_equiv.txt](https://github.com/PBSPro/pbspro/files/3433545/test_equiv.txt)
[test_new_branch_all.txt](https://github.com/PBSPro/pbspro/files/3433546/test_new_branch_all.txt)
[test_master_branch_all.txt](https://github.com/PBSPro/pbspro/files/3433547/test_master_branch_all.txt)

arungrover:
travis seems to be stuck, I'm gonna close the PR and reopen it again

arungrover:
oh well, that didn't work. I guess I'll rebase it, resolve conflict and push the change in

arungrover:
@bhroam Thanks for your review comments, I have addressed them in my latest commit.
Additionally, I have also addressed 3 bugs in the code.
1 - Update_preemption_on_run used to sort all the jobs. This caused next_job() function to overlook some of the jobs because now it correctly keeps track of the index where it left and ifsf sorting resulted into running jobs to move at the end then it used to skip jobs.
2 - During preemption, we used to call filter_preemptible_jobs function to get rid of jobs that are not going to help in running the high priority job. That function was missing an err code check for insufficient server/queue resource. I added that.
3 - During preemption, update_server/queue_on_end were using released resource request to release resources on server/queue. But now since we do not create released resources list unless RRTROS is set, the functions were failing to release the resources. I have added similar check to update_server/queue_on_end() functions as well.

@agrawalravi90 Please have a look when you have time.

test/tests/performance/pbs_preemptperformance.py
```diff
@@ -495,6 +495,7 @@ def test_preemption_with_unrelated_soft_limits(self):
             j = Job(TEST_USER2, attrs=a)
             j.set_sleep_time(3000)
             self.server.submit(j)
+        time.sleep(15)
```
bhroam: Put a self.logger.info() ahead of this sleep saying the reason why we're sleeping.

arungrover: Good catch! Actually, there was no reason to sleep here. I added that because I was trying to run gprof compiled scheduler and I used to swap the normal sched binary with gprof compiled sched binary in these 15 seconds :)
I did not clean up well

src/scheduler/data_types.h
```diff
@@ -791,12 +797,22 @@ struct mom_res
...
+	char *name;			/* name of entitiy */
+	int running;			/* count of running jobs in object */
+	int soft_limit_preempt_bit;	/* Place to store preempt bit if entity is over limits */
+	resource_count *rescts;		/* resources used */
 	counts *next;
 };
 
+struct resource_count
+{
+	char *name;		    /* resource name */
```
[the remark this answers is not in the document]
arungrover: Actually the structure is made similar to our resource_req structure. It is for easy lookup of name. when we assign resource_count->name we point it to resdef->name only.

src/scheduler/fifo.c
```diff
@@ -458,6 +458,13 @@ init_scheduling_cycle(status *policy, int pbs_sd, server_info *sinfo)
 	 * 2. we need all the jobs to be created and up to date for soft run limits
 	 */
 
+	/* Before setting preempt priorities on all jobs, make sure that entity's preempt bit
+	 * is updated for all running jobs
+	 */
+	if ((sinfo->running_jobs != NULL) && (policy->preempting)) {
+		for (i = 0; sinfo->running_jobs[i] != NULL; i++)
```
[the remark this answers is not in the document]
arungrover: Good catch!! I'll add that change

src/scheduler/resource_resv.c
```diff
@@ -759,6 +759,40 @@ dup_resource_req_list(resource_req *oreq)
...
+	resource_count *rcount;
+	resource_count *nrcount;
+	resource_count *head;
+	resource_count *prev;
+
+	head = NULL;
+	prev = NULL;
+	rcount = orcount;
+
+	while (rcount != NULL) {
```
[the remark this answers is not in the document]
arungrover: You are right, I can do this. Currently, code is in sync with how dup_resource_req_list() operates. If you are ok I'd like to retain it as is.

src/scheduler/limits.c
```diff
@@ -1159,15 +1210,34 @@ int
...
 {
 	int	rc = 0;
-	int	i;
+
+	if (si == NULL || qi == NULL || rr == NULL)
+		return 0;
 
 #ifdef NAS /* localmod 097 */
-	if (! si->has_soft_limit) {
+	if (!si->has_soft_limit) {
```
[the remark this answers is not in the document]
arungrover: This is NAS compilation block that is why I didn't touch it at all.

src/scheduler/server_info.c
```diff
@@ -1977,13 +1977,13 @@ update_server_on_end(status *policy, server_info *sinfo, queue_info *qinfo,
...
-					if (((resresv->job->queue->has_user_limit || sinfo->has_user_limit) &&
-					    (!strcmp(resresv->user, sinfo->jobs[i]->user))) ||
-					    ((resresv->job->queue->has_grp_limit || sinfo->has_grp_limit) &&
-					    (!strcmp(resresv->group, sinfo->jobs[i]->group))) ||
-					    ((resresv->job->queue->has_proj_limit || sinfo->has_proj_limit) &&
-					    (!strcmp(resresv->project, sinfo->jobs[i]->project))))
+				if (sinfo->jobs[i]->job != NULL) {
+					int usrlim = resresv->job->queue->has_user_limit | sinfo->has_user_limit;
+					int grplim = resresv->job->queue->has_grp_limit | sinfo->has_grp_limit;
+					int projlim = resresv->job->queue->has_proj_limit | sinfo->has_proj_limit;
```
bhroam: I think we want to use || instead of the bitwise |.  This is because we're using the structure bit fields instead of doing it ourselves.  This only works because has_use_limit is :1.  If it was :2 or :3, it'd hold values that were not bits.

arungrover: You are right Bhroam, but do you see the need for has_user_limit flags to support more than one bit in future? I can not imagine what it would store.
Even if it does, all we are looking for is that if there is any user limit set on queue or server so as long as OR value is a non zero it would work here.

bhroam: structure bitfields are not bits.  They are variables smaller than a byte.  If we were change has_user_limit and turn it into a :4, it just means the value that can be stored can have 4 bits (0-15).  This isn't the same as having a 2 or 4 or 8 stored in an unsigned integer which to use the bitwise operators on.  It just so happens that has_user_limits is a :1, meaning it can contain one bit (0-1).  They aren't meant to be bits, they are meant to be normal variables with smaller ranges.

arungrover: I know that it just limits the range of values a variable can store. But does it matter to us if we are doing 3 | 5 or 3||5 because as long as it is non-zero we know there is a limit set. right?

bhroam: All I am saying is we should use the correct operator.  The '|' operator is meant for bits.  The || operator is meant for normal variables.  A structure bit field is a normal variable with a limited range.  You are correct that both will work, but I don't think the '|' operator is right.  We wouldn't store a set of bits in a bitfield.  We'd use an unsigned long and use the '&' and '|' operators.

arungrover: I've made the change @bhroam and pushed the change.

arungrover:
@agrawalravi90 I have run all limits related tests present in test harness and all of them passed

Issue #1183, opened 2019-07-01
sched_preempt_enforce_resumption broken on master branch


saritakh:
Steps to reproduce issue:
1. Single node setup with 4 ncpus; having expressq with >150 priority and sched attribute sched_preempt_enforce_preemption enabled
2. Submit normal job asking 4 ncpus & 300secs walltime and 2 single ncpus job one asking 50secs walltime and other asking 200secs walltime
=> j1 runs, j2, j3 are queued
3. Submit high priority job to expressq asking 2 ncpus & 100secs walltime
=> j4 runs after preempting j1. Here due to sched_preempt_enforce_resumption j3 should not have run; but it runs.

[saritah@nx15 jobs]$ qsub -lselect=ncpus=4 -l walltime=300 -- /bin/sleep 500
0.nx15
[saritah@nx15 jobs]$ sleep 1
[saritah@nx15 jobs]$ qsub -lselect=ncpus=1 -l walltime=50 -- /bin/sleep 500
1.nx15
[saritah@nx15 jobs]$ sleep 1
[saritah@nx15 jobs]$ qsub -lselect=ncpus=1 -l walltime=200 -- /bin/sleep 500
2.nx15
[saritah@nx15 jobs]$ sleep 4
[saritah@nx15 jobs]$ qsub -lselect=ncpus=2 -l walltime=100 -q expressq -- /bin/sleep 500
3.nx15
[saritah@nx15 jobs]$ qstat -s

nx15:
                                                            Req'd  Req'd   Elap
Job ID          Username Queue    Jobname    SessID NDS TSK Memory Time  S Time
--------------- -------- -------- ---------- ------ --- --- ------ ----- - -----
0.nx15* saritah  workq    STDIN       13671   1   4    --  00:05 S 00:00
   Not Running: Not enough free nodes available
1.nx15* saritah  workq    STDIN       13679   1   1    --  00:00 R 00:00
   Job run at Sun Jun 30 at 23:24 on (nx15:ncpus=1)
2.nx15* saritah  workq    STDIN       13680   1   1    --  00:03 R 00:00
   Job run at Sun Jun 30 at 23:24 on (nx15:ncpus=1)
3.nx15* saritah  expressq STDIN       13678   1   2    --  00:01 R 00:00
   Job run at Sun Jun 30 at 23:24 on (nx15:ncpus=2)
[saritah@nx15 jobs]$ qmgr -c "p sched"
#
# Create and define scheduler default
#
create sched default
...
[...]

Issue #1185, opened 2019-07-02
scheduler attribute preempt_sort needs to be revisited


saritakh:
An unset of attribute should get it back to default value; which is not the case with scheduler attribute preempt_sort.
A recent scheduler enhancement updated configuration of sched parameter preempt_sort from 'sched_config' file to qmgr interface; whose change summary is as below:

**Old sched_config behavior:**
Default:
	preempt_sort = min_time_since_start
Non-default:
	#preempt_sort = min_time_since_start

**New qmgr behavior:**
Default:
	preempt_sort = min_time_since_start
Non-default: qmgr unset action
	<Attribute removed from qmgr "p sched">

Now in order to get this attribute back to default value; we actually need to set it back to "min_time_since_start". By convention, the "unset" of "preempt_sort" should result in default value i.e. min_time_since_start; which is not the case right now.  
Right now once we unset "preempt_sort", in order to get it to back to default value, we need to remember the string "min_time_since_start" and set the same.
This is not the case with other scheduler parameters as shown below:

[root@n27 ~]# qmgr -c "print sched"
...
set sched scheduling = True
set sched scheduler_iteration = 600
set sched state = idle
set sched preempt_queue_prio = 150
set sched preempt_prio = "express_queue, normal_jobs"
set sched preempt_order = SCR
set sched preempt_sort = min_time_since_start
[root@n27 ~]# qmgr -c "unset sched preempt_sort"
[root@n27 ~]# qmgr -c "print sched"
...
set sched scheduling = True
set sched scheduler_iteration = 600
set sched state = idle
set sched preempt_queue_prio = 150
set sched preempt_prio = express_queue, normal_jobs
set sched preempt_order = SCR
[root@n27 ~]# qmgr -c "unset sched preempt_prio"
[root@n27 ~]# qmgr -c "print sched"
...
set sched scheduling = True
set sched scheduler_iteration = 600
set sched state = idle
set sched preempt_queue_prio = 150
set sched preempt_prio = express_queue, normal_jobs
set sched preempt_order = SCR
[root@n27 ~]# qmgr -c "u sched preempt_queue_prio"
[root@n27 ~]# qmgr -c "p sched"
...
set sched scheduling = True
set sched scheduler_iteration = 600
set sched state = idle
set sched preempt_queue_prio = 150
set sched preempt_prio = express_queue, normal_jobs
set sched preempt_order = SCR
[root@n27 ~]# qmgr -c "u sched preempt_order"
[...]

PR #1189, opened 2019-07-05
Scheduler skips adding preempted job to calendar

arungrover:
#### Describe Bug or Feature

#1183 
#1166 made a change where scheduler started marking a preempted job as can_not_run in the cycle it gets preempted.
Due to this change, while finding the next job in the same cycle, scheduler started skipping this job because find_non_normal_job_ind and find_runnable_resresv_ind functions ignore jobs that have "can_not_run" flag set on them. 
Since these jobs were getting skipped, they were never added to the calendar and undeserving filler jobs started running on its resource resulting in delaying the start time of the preempted job.

#### Describe Your Change
Instead of checking for the job to be preempted (by suspended or checkpointed) and then marking the jobs as "can not run", we will rather change the preemption code to break the association of preempted job and the equivalence class.
This way suspended jobs will still be considered to run and get into the calendar when they cannot run and their equivalence class can still be marked as "can not run" if scheduler could not resume/rerun them.

No new test needed, pbs_sched_preempt_enforce_resumption.py  tests and pbs_equiv_classses.py tests are enough for testing this
[test_susp_bug.txt](https://github.com/PBSPro/pbspro/files/3364072/test_susp_bug.txt)
[test_equiv.txt](https://github.com/PBSPro/pbspro/files/3364073/test_equiv.txt)
[test_susp.txt](https://github.com/PBSPro/pbspro/files/3364074/test_susp.txt)

### DIFF
--- a/src/scheduler/data_types.h
+++ b/src/scheduler/data_types.h
@@ -854,9 +854,6 @@ struct resresv_set
 	place *place_spec;		/* place spec of set */
 	resource_req *req;		/* ATTR_L (qsub -l) resources of set.  Only contains resources on the resources line */
 	queue_info *qinfo;		/* The queue the resresv is in if the queue has nodes associated */
-
-	resource_resv **resresv_arr;	/* The resresvs in the set */
-	int num_resresvs;		/* The number of resresvs in the set */
 };
 
 struct node_partition
--- a/src/scheduler/job_info.c
+++ b/src/scheduler/job_info.c
@@ -2002,8 +2002,6 @@ new_resresv_set(void)
 	rset->req = NULL;
 	rset->select_spec = NULL;
 	rset->qinfo = NULL;
-	rset->resresv_arr = NULL;
-	rset->num_resresvs = 0;
 
 	return rset;
 }
@@ -2023,7 +2021,6 @@ free_resresv_set(resresv_set *rset) {
 	free_selspec(rset->select_spec);
 	free_place(rset->place_spec);
 	free_resource_req_list(rset->req);
-	free(rset->resresv_arr);
 	free(rset);
 }
 /**
@@ -2100,15 +2097,8 @@ dup_resresv_set(resresv_set *oset, server_info *nsinfo)
 		free_resresv_set(rset);
 		return NULL;
 	}
-	rset->resresv_arr = copy_resresv_array(oset->resresv_arr, nsinfo->all_resresv);
-	if (rset->resresv_arr == NULL) {
-		free_resresv_set(rset);
-		return NULL;
-	}
 	if(oset->qinfo != NULL)
 		rset->qinfo = find_queue_info(nsinfo->queues, oset->qinfo->name);
-
-	rset->num_resresvs = oset->num_resresvs;
 
 	return rset;
 }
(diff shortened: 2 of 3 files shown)


arungrover:
I had a discussion with @bhroam  and we think instead of checking for the job to be preempted (by suspended or checkpointed) and then NOT updating equivalence class as "can not run" we will rather change the preemption code to break the association of preempted job and the equivalence class.
This way suspended jobs will still be considered to run and get into the calendar when the cannot run and their equivalence class can still be marked as "can not run" if scheduler could not resume/rerun them.
[test_susp.txt](https://github.com/PBSPro/pbspro/files/3379829/test_susp.txt)
[test_equiv.txt](https://github.com/PBSPro/pbspro/files/3379830/test_equiv.txt)

Please ignore two test failures in equiv class test results. Those failures are because of #1185

arungrover:
Rebasing the branch to resolve merge conflicts because appveyor is unable to build with merge conflicts

src/scheduler/job_info.c
```diff
@@ -2508,36 +2498,19 @@ create_resresv_sets(status *policy, server_info *sinfo)
...
-		tmp_arr = realloc(rsets[i]->resresv_arr, (rsets[i]->num_resresvs + 1) * sizeof(resource_resv *));
-		if (tmp_arr != NULL)
-			rsets[i]->resresv_arr = tmp_arr;
-	}
-
 	tmp_rset_arr = realloc(rsets,(j + 1) * sizeof(resresv_set *));
 	if (tmp_rset_arr != NULL)
 		rsets = tmp_rset_arr;
-
+	i = count_array((void **)rsets);
```
bhroam: Since this is no longer an index variable, consider changing the name of it.

arungrover: Well, this variable is used to iterate through all jobs in the loop above this code. I am just reusing this variable to get the number of sets created.

bhroam: I'd still use a different variable.  Something like num_rsets.

agrawalravi90:
@arungrover can you please update the "Describe your change" section to only mention the latest approach to fixing this issue?

src/scheduler/data_types.h
```diff
@@ -854,9 +854,6 @@ struct resresv_set
 	place *place_spec;		/* place spec of set */
 	resource_req *req;		/* ATTR_L (qsub -l) resources of set.  Only contains resources on the resources line */
 	queue_info *qinfo;		/* The queue the resresv is in if the queue has nodes associated */
-
-	resource_resv **resresv_arr;	/* The resresvs in the set */
```
agrawalravi90: did you remove this because it was not used?

arungrover: Yes, resresv_arr stored the list of all jobs in the equivalence class but we were not using it anywhere. Thus removed it.

arungrover:
@bhroam @agrawalravi90  apologies for pushing the changes so late. I have addressed Bhroam's concern in the latest commit and also changed the description of the change in the PR template. Please have a look again

arungrover:
Thanks @agrawalravi90!
@bhroam Please have a look at the most recent commit.
I have also retested the code and here are the test logs
[test_susp.txt](https://github.com/PBSPro/pbspro/files/3429254/test_susp.txt)
[test_equiv.txt](https://github.com/PBSPro/pbspro/files/3429255/test_equiv.txt)

src/scheduler/data_types.h
bhroam: Yeah, this was left over when I did the original POC of equivalence classes.  In the original POC, when a job in a class could not run, we'd mark all jobs in the class at that point.  I'd walk down the resresv_arr then.  This disturbed the political order.  When the final equivalence class RFE was finished, we now mark the class as can not run and when we reach new jobs from next_job(), we immediately know they can't run.  This way the political order is not changed.

PR #1192, opened 2019-07-09
[WIP] Fix assert error in periodic_update_check_values

minghui-liu:
#### Describe Bug or Feature

Test_cgroup_periodic_update_check_values fails due to an assertion error.
I think this is caused by a typo.

#### Describe Your Change

Changed the value in the assertion

[pbs_cgroup_hook_test_output.txt](https://github.com/PBSPro/pbspro/files/3374880/pbs_cgroup_hook_test_output.txt)

### DIFF
--- a/test/tests/functional/pbs_cgroups_hook.py
+++ b/test/tests/functional/pbs_cgroups_hook.py
@@ -1202,13 +1202,13 @@ if %s e.job.in_ms_mom():
         match = re.match(r'(\d+)kb', mem)
         self.assertFalse(match is None)
         usage = int(match.groups()[0])
-        self.assertGreater(30000, usage)
+        self.assertGreater(300000, usage)
         if self.swapctl == 'true':
             vmem = qstat[0]['resources_used.vmem']
             match = re.match(r'(\d+)kb', vmem)
             self.assertFalse(match is None)
             usage = int(match.groups()[0])
-            self.assertGreater(30000, usage)
+            self.assertGreater(300000, usage)
         err_msg = "Unexpected error in pbs_cgroups " + \
             "handling exechost_periodic event: TypeError"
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -1202,13 +1202,13 @@ def test_cgroup_periodic_update_check_values(self):
...
         usage = int(match.groups()[0])
-        self.assertGreater(30000, usage)
+        self.assertGreater(300000, usage)
         if self.swapctl == 'true':
             vmem = qstat[0]['resources_used.vmem']
             match = re.match(r'(\d+)kb', vmem)
             self.assertFalse(match is None)
             usage = int(match.groups()[0])
-            self.assertGreater(30000, usage)
+            self.assertGreater(300000, usage)
```
bayucan: If the one in line 1205 and this one are supposed to be the same number, then perhaps put it in a variable.

minghui-liu: Since the value is used only twice I don't think it's really necessary. And having a number instead of a variable in an assert statement helps with debugging when the assertion fails.

bayucan: ok

mike0042: I'm not entirely clear how increasing the value is going to fix the failed assertion. If the usage never made it to 30000 previously it's certainly not going to make it to 300000.

minghui-liu: The assertion is testing that the memory usage of the job is under 30000KB or 30MB, not the other way around. The actually usage is around 120 MB at the time of the assertion.

PR #1194, opened 2019-07-10
Fix memory leak due to unfreed batch request in end of job processing

subhasisb:
#### Describe Bug or Feature

Memory growth is observed at several customer sites related to post job processing errors.
When post_cpyfile() on mom detects that a stageout failed, it sends a second obit to server, which makes it traverse through on_job_exit() once more, this causes a second batch request to be sent out to mom at whatever exiting stage the job is - could be a deletefiles or deletejob batch request (depends on timing). 

When the mom responds to the first deletejob request, the on_job_exit() routine goes to save the job in history, and in that process clears off ALL PENDING tasks on the job. (however, it DOES NOT free any associated batch requests, which free_job does)

However, by the time the mom responds to the second (duplicate) batch request, the job's tasks are all deleted and thus on_job_exit is never triggered, and the allocated batch structure is never freed. 

Another related issue is when the tpp connection between the server and mom is broken while a deletejob or copyfiles operation is pending, we fall into error path, but do not free the previously allocated batch request.

The test to reproduce these leaks is by breaking ssh (so scp stageout fails, say by breaking passwordless ssh) or by breaking the TPP connection between the server and mom (e.g. by killing pbs_comm). These are quite corner cases and we probably do not need PTL tests to cover these for detecting leaks.

On Valgrind this shows up as reachable memory, because server will continue to have a pointer to it, but will never free it, and this will accumulate, and thus is a leak: **(seen in svr_before_fix.log)**
_==124338== 2,264 bytes in 1 blocks are still reachable in loss record 1,437 of 1,555
==124338==    at 0x4C29BC3: malloc (vg_replace_malloc.c:299)
==124338==    by 0x46933B: alloc_br (process_request.c:1010)
==124338==    by 0x470B4F: on_job_exit (req_jobobit.c:977)
==124338==    by 0x4DD377: dispatch_task (work_task.c:142)
==124338==    by 0x444EB5: process_DreplyRPP (issue_request.c:899)
==124338==    by 0x457448: is_request (node_manager.c:4650)
==124338==    by 0x463C37: do_rpp (pbsd_main.c:462)
==124338==    by 0x463CC5: rpp_request (pbsd_main.c:510)
==124338==    by 0x4E1F48: process_socket (net_server.c:504)
==124338==    by 0x4E2257: wait_request (net_server.c:620)
==124338==    by 0x466F0B: main (pbsd_main.c:2137)_

#### Describe Your Change

The fix is to add the code block from free_job(or rather make it a function to be called from both places) so that pending batch requests are deleted when job is freed or when job goes to history.

[svr_before_fix.log](https://github.com/PBSPro/pbspro/files/3376793/svr_before_fix.log)
[svr_fixed.log](https://github.com/PBSPro/pbspro/files/3376794/svr_fixed.log)

### DIFF
--- a/src/server/job_func.c
+++ b/src/server/job_func.c
@@ -416,6 +416,57 @@ job_alloc(void)
 	return (pj);
 }
 
+#ifndef PBS_MOM
+
+/**
+ * @brief
+ * 	free work tasks and pending batch requests related to this job
+ *
+ * @param[in]	pj - pointer to job structure
+ *
+ * @return	void
+ */
+void
+free_job_work_tasks(job *pj)
+{
+	struct work_task *pwt;
+	struct batch_request *tbr = NULL;
+	/*
+	* Delete any work task entries associated with the job.
+	* mom deferred tasks via TPP are also hooked into the
+	* ji_svrtask now, so they also get automatically cleared
+	* in this following loop
+	*/
+	while ((pwt = (struct work_task *)GET_NEXT(pj->ji_svrtask)) != NULL) {
+		if (pwt->wt_type == WORK_Deferred_Reply) {
+			tbr = (struct batch_request *)pwt->wt_parm1;
+			if (tbr != NULL) {
+				/* Check if the reply is for scheduler
+					* If so, then reject the request.
+					*/
+				if ((tbr->rq_orgconn != -1) &&
+					(find_sched_from_sock(tbr->rq_orgconn) != NULL)) {
+					tbr->rq_conn = tbr->rq_orgconn;
+					req_reject(PBSE_HISTJOBID, 0, tbr);
+				}
+				/*
+				* free batch request from task struct
+				* if task is deferred reply
+				*/
+				else
+					free_br(tbr);
+			}
+		}
+
+		/* wt_event2 either has additional data (like msgid) or NULL */
+		free(pwt->wt_event2);
+
+		delete_task(pwt);
+	}
+
+}
+#endif
+
 /**
  * @brief
  * 		job_free - free job structure and its various sub-structures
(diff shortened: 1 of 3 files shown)


src/server/req_jobobit.c
```diff
@@ -1250,6 +1260,8 @@ on_job_rerun(struct work_task *ptask)
 					(preq->rq_reply.brp_code == DIS_EOD)) {
 					/* tcp connection to Mom broken */
 					conn_to_mom_failed(pjob, on_job_rerun);
+					free_br(preq);
+					NULL;
```
mike0042: preq = NULL;

subhasisb: Oops - no idea how that happened! Thanks will fix asap

mike0042:
I'd like to note that the leak summaries before and after the fix are very similar, but the error summary has been reduced by two. This may interest future reviewers.

subhasisb:
Sorry for the confusion @mike0042 , I should have pointed out the section in the valgrind logs that are the culprit for this case. Have edited the description to point that out now.

src/server/job_func.c
```diff
@@ -416,6 +416,57 @@ job_alloc(void)
...
+	* ji_svrtask now, so they also get automatically cleared
+	* in this following loop
+	*/
+	while ((pwt = (struct work_task *)GET_NEXT(pj->ji_svrtask)) != NULL) {
+		if (pwt->wt_type == WORK_Deferred_Reply) {
+			tbr = (struct batch_request *)pwt->wt_parm1;
+			if (tbr != NULL) {
+				/* Check if the reply is for scheduler
+					* If so, then reject the request.
+					*/
```
bhroam: Is the indentation off?  Github shows the second two lines of the comment indented further than the first.

src/server/job_func.c
```diff
@@ -416,6 +416,57 @@ job_alloc(void)
...
+				if ((tbr->rq_orgconn != -1) &&
+					(find_sched_from_sock(tbr->rq_orgconn) != NULL)) {
+					tbr->rq_conn = tbr->rq_orgconn;
+					req_reject(PBSE_HISTJOBID, 0, tbr);
```
bhroam: Is HISTJOBID the right reject code here?  Job history might not be on.

subhasisb: This was the same error code used in job_free() which is called when history is disabled. 
I will modify the code to return HISTJOBID in case of history enabled and UNKNOWN job id in case of no history. Thanks!

subhasisb: Actually, on second thoughts, I think i would like to leave it as is. This is the exact same code that was being used in the case of history-disabled, right, so it must have been working - or the callers do not care. In other words, i think tampering the return code is probably more dangerous than leaving it as is. 

The change here is returning an error in the case of history jobs and for that the return code is appropriate anyway.

PR #1195, opened 2019-07-11
Incorrect gpu count

Jitendra-Rajak:
#### Describe Bug or Feature
* all 'osdev_type="1"' devices are being counted as devices.
* It was the issue while counting the number of gpu devices, we were also counting the non gpu devices in number of gpu's.

#### Describe Your Change
* Counting the gpu devices seperately before adding it to ndevices so that if there is a non gpu device it should not get added in the gpu devices.

*  [topology_output_brfore.txt](https://github.com/PBSPro/pbspro/files/3381123/topology_output_brfore.txt)
* [topology_output_after.txt](https://github.com/PBSPro/pbspro/files/3381124/topology_output_after.txt)

 Topology_file
* [jitendra.txt](https://github.com/PBSPro/pbspro/files/3380922/jitendra.txt)

### DIFF
--- a/src/cmds/scripts/pbs_topologyinfo.py
+++ b/src/cmds/scripts/pbs_topologyinfo.py
@@ -257,7 +257,7 @@ def socketXMLstart(name, attrs):
             inventory.gpudevices += 1
         if (name == "object" and attrs.get("type") == "OSDev" and
             attrs.get("osdev_type") == "1" and
-                attrs.get("name").startswith("control")):
+                attrs.get("name").startswith("controlD")):
             if (inventory.gpudevices > 0):
                 inventory.gpudevices -= 1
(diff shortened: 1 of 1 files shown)


src/cmds/scripts/pbs_topologyinfo.py
```diff
@@ -247,7 +254,12 @@ def socketXMLstart(name, attrs):
         if (name == "object" and attrs.get("type") == "OSDev" and
             attrs.get("osdev_type") == "1" and
                 attrs.get("name").startswith("card")):
-            inventory.ndevices += 1
+            inventory.gpudevices += 1
+        if (name == "object" and attrs.get("type") == "OSDev" and
+            attrs.get("osdev_type") == "1" and
+                attrs.get("name").startswith("control")):
```
nithinj: Use "controlD"

PR #1200, opened 2019-07-12
Create a configuration/environment variable for the scheduler host name

lisa-altair:
#### Describe Bug or Feature
When running the PBS server and scheduler in a Kubernetes pod the scheduler can’t connect to the server because it looks like the connections are coming from a different IP address than the server’s.

#### Describe Your Change
Created a new PBS configuration variable called PBS_SCHEDULER_HOST_NAME.
This configuration/environment variable can be set by admins when starting PBS.

#### Link to Design Doc
There is a design, located here: [project documentation area](https://pbspro.atlassian.net/wiki/spaces/PD/pages/1318387715/New+scheduler+service+name+conf+variable)

[ptl_output.txt](https://github.com/PBSPro/pbspro/files/3388059/ptl_output.txt)

### DIFF
--- a/src/server/pbsd_main.c
+++ b/src/server/pbsd_main.c
@@ -1074,6 +1074,11 @@ main(int argc, char **argv)
 		pbs_scheduler_addr = get_hostaddr(pbs_conf.pbs_primary);
 	}
 
+	/* set the scheduler name if it is set in an env var or conf file */
+	if (pbs_conf.pbs_scheduler_host_name != NULL) {
+		pbs_scheduler_addr = get_hostaddr(pbs_conf.pbs_scheduler_host_name);
+	}
+
 	/* parse the parameters from the command line */
 
 	while ((c = getopt(argc, argv, "A:a:Cd:e:F:p:t:lL:M:NR:S:g:G:s:P:-:")) != -1) {
--- a/test/tests/functional/pbs_conf_var.py
+++ b/test/tests/functional/pbs_conf_var.py
@@ -65,12 +65,10 @@ class TestConf(TestFunctional):
         self.server.log_match(logmsg, starttime=now)
 
         # Now set it to cvs and show that the server can't talk to the sched
-        self.server.stop()
         self.du.set_pbs_config(
             self.server.hostname,
             confs={'PBS_SCHEDULER_HOST_NAME': 'cvs.pbspro.com'})
         now = int(time.time())
         self.server.restart()
-        self.assertTrue(self.server.isUp(), 'Failed to start PBS')
         logmsg = 'Could not contact Scheduler'
         self.server.log_match(logmsg, starttime=now)
--- a/test/tests/functional/pbs_conf_var.py
+++ b/test/tests/functional/pbs_conf_var.py
@@ -54,9 +54,6 @@ class TestConf(TestFunctional):
         the server can't talk to the sched, to prove that the
         variable is being used.
         """
-        self.server.stop()
-        self.assertFalse(self.server.isUp(), "Failed to stop PBS")
-
         conf = self.du.parse_pbs_config(self.server.hostname)
         self.du.set_pbs_config(
             self.server.hostname,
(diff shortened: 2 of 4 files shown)


test/tests/functional/pbs_conf_var.py
```diff
@@ -0,0 +1,81 @@
...
+            confs={'PBS_SCHEDULER_HOST_NAME': conf['PBS_SERVER']})
+        now = int(time.time())
+        self.server.start()
+        self.assertTrue(self.server.isUp(), 'Failed to start PBS')
+        logmsg = "request received from Scheduler@%s" % (self.server.hostname)
+        self.server.log_match(logmsg, starttime=now)
+
+        """
+        Now set it to cvs and show that the server can't talk to the sched
+        """
```
bhroam: Don't use triple quoted strings as comments.  Use #.

test/tests/functional/pbs_conf_var.py
```diff
@@ -0,0 +1,81 @@
...
+        Start the server with the PBS_SCHEDULER_HOST_NAME
+        set to the name of the local host to show the variable works
+        Also set it to some other existing machine that isn't set up
+        to work with this machine (e.g. cvs.pbspro.com) and show
+        the server can't talk to the sched, to prove that the
+        variable is being used.
+        """
+        self.server.stop()
+        self.assertFalse(self.server.isUp(), "Failed to stop PBS")
```
bhroam: isUp() will check for 20 times.  This unnecessarily waits.  Just set your variable and do a restart().  This means you only have to call isUp() once when you are really checking for the machine to be up.

test/tests/functional/pbs_conf_var.py
```diff
@@ -0,0 +1,81 @@
...
+            self.server.hostname,
+            confs={'PBS_SCHEDULER_HOST_NAME': conf['PBS_SERVER']})
+        now = int(time.time())
+        self.server.start()
+        self.assertTrue(self.server.isUp(), 'Failed to start PBS')
+        logmsg = "request received from Scheduler@%s" % (self.server.hostname)
+        self.server.log_match(logmsg, starttime=now)
+
+        """
+        Now set it to cvs and show that the server can't talk to the sched
```
vstumpf: I would change this to a line comment with `#`, but that's just preference. Up to you to change it.

src/server/pbsd_main.c
```diff
@@ -1074,6 +1074,11 @@ main(int argc, char **argv)
 		pbs_scheduler_addr = get_hostaddr(pbs_conf.pbs_primary);
 	}
 
+	/* set the scheduler name if it is set in an env var or conf file */
+	if (pbs_conf.pbs_scheduler_host_name != NULL) {
+		pbs_scheduler_addr = get_hostaddr(pbs_conf.pbs_scheduler_host_name);
+	}
+
```
bhroam: I'm not sure how this should interact with failover.  The non-failover case is obvious.  The comment above the failover case says the default scheduler is on the primary's host.  If we change it to this, what happens when the secondary takes over?  Is it still here?  Also, how does this interact with multi-sched?

test/tests/functional/pbs_conf_var.py
vstumpf: isUp only loops 20 times for MoM's isUp()
I didn't think of using restart, I agree that's better.

bhroam: Both isUp() functions loop 20 times.  Since it takes the server a while to read the database and do other start up things, it will be running but not responsive for some time.  The loop waits for it to become responsive.

test/tests/functional/pbs_conf_var.py
bhroam: pep8 tells us to use #'s for multi-line comments.  See: https://www.python.org/dev/peps/pep-0008/#block-comments

src/server/pbsd_main.c
lisa-altair: If this is set, then I think it should be what is used (over-written only by -S), this is how the other /etc/pbs.conf file variables work.  That is why the code is located after the failover code. Also, for the secondary server that might be located on a different host, it will have it's own configuration file, and should have the same PBS_SCHEDULER_HOST_NAME set.  What do you think?
For multisched it should not matter.  It's a setting used for the server. But just in case I'll try running a few tests to see.

test/tests/functional/pbs_conf_var.py
vstumpf: Ah, I see. MoM is the only one that can be changed with an argument.

src/server/pbsd_main.c
bhroam: That's a good point that the secondary will have its own conf file.  As for multi-sched, right now they all have to be on the server's host.  It was always in the plans to make it so they could be elsewhere.  I don't know if that changes things.

test/tests/functional/pbs_conf_var.py
```diff
@@ -62,20 +62,18 @@ def test_sched_host_name_envvar(self):
             self.server.hostname,
             confs={'PBS_SCHEDULER_HOST_NAME': conf['PBS_SERVER']})
         now = int(time.time())
-        self.server.start()
+        self.server.restart()
```
bhroam: Please remove the call to self.server.stop()/isUp() above, this call to restart() will handle it.  The part of your code I disliked was the call to isUp() after the stop.  It's going to add 20s to the test that we don't really need.

vstumpf: If the pid is gone isUp() will return false on the first iteration of the loop, so it won't add 20s.
But I do agree not stop/starting and instead just restarting is better.

lisa-altair:
It was determined that it would be better to not implement the feature this way.
But rather to use qmgr to set the default scheduler's sched_host.
Closing this pull request.

PR #1203, opened 2019-07-17
Optimize scheduler for 1M jobs/50k nodes

bhroam:
#### Describe Your Change
Several scheduler optimizations when running many jobs on many nodes.  The system I was using had 950k jobs and 50k nodes.

The following optimizations were done:
1) Before the placement set meta data (including the allpart) was recomputed every time a job runs.  The snapshot I was using ran 50k jobs in one cycle.  Even though we were only updating the allpart, this slowed things down.  Instead of walking down each placement set and each node in the placement set to recalculate the metadata, we walk down the exec_vnode of the job we just run and update the correct placement sets.  Each node now knows which placement sets it is part of.  The previous code was O(N^3) (jobs\*psets\*nodes).  The worst case scenario of the new case is the same.  The best case is with no placement sets (just the allpart) and only single node jobs.  This is O(N) (jobs).   Even with placement sets enabled, we only update the placement sets we need to update instead of all of them.

2) The function exists_run_event() checked to see if a run event existed during a jobs lifetime.  This would walk down the calendar looking for a run event.  The calendar is 99.9% made up of end events, so we'd walk past mostly end events looking for the run event.  Now the calendar keeps track of the the first run event.  The function changed from O(N) to O(1).

3) Deep in the node searching code, the scheduler used to call esists_run_event() to determine if there was a run event in the calendar before doing a simulation.  This wouldn't tell it if there was a run event on the node, so we still might not have needed to do the simulation.  The node bucket code added a list of run events on each node.  This allowed me to look at the first run event on that node to check if we need to do the simulation.  Since we were calling exists_run_event(), before it was O(N).  It is now O(1).  This is significant since this is deep in node search which is one of the most expensive areas of the scheduler.

4) A job now knows its run and end event in the calendar.  These were used in several places where we were searching for the event.  It turned several O(N) searches into O(1).

5) Fixed a bug when the scheduler was selecting the next job to run.  The code should have been O(N) (jobs) since next_job() kept track of where it was in the list so it can jump back there.  The bug had us reset our last used index.  It turned what should have been O(N) into O(N^2).

6) The normal node code path would allocate an array of pointers for the exec_vnode.  Since it might need to split chunks across vnodes of the same host, it allocates the worst case scenario.  The number of chunks * number of vnodes (every chunk split across every vnode).  In the case of 50k nodes, this is 390kb per job.  With 50k jobs running, that is 18.6gb.  I now trim down the exec_vnode to the size needed at the end of eval_selspec().

Running 10k jobs on 10k nodes:
Old code: 56.5 seconds
New code: 21.5 seconds

Simulation of new code in 950k jobs with 50k nodes:
5.5m to run 50k jobs.  This is longer than 21.5 * 5 since 50k nodes requires more searching.
30s to dismiss the other 900k jobs using equivalence classes.  All the jobs were in the same class, so we didn't have to do any work.  It dismissed jobs at a rate of 29000/sec.

I will eventually create the numbers from master (most likely released 19).  When I started, the simulation only ran a partial number of the jobs in a 20m cycle.  It now makes it all the way through the cycle in 6.5m.

### DIFF
--- a/src/scheduler/constant.h
+++ b/src/scheduler/constant.h
@@ -74,15 +74,15 @@ enum sort_status
 };
 
 
-/* enum used to find out what to skip while searching for the next job to schedule */
+/* enum used to find out what to skip while searching for the next job to schedule.  Values are bits in a bitfield */
 
 enum skip
 {
 	SKIP_NOTHING,
 	/* Value used to know whether reservations are already scheduled or not */
-	SKIP_RESERVATIONS,
+	SKIP_RESERVATIONS = 1,
 	/* Value used to know whether express, preempted, starving jobs are already scheduled or not */
-	SKIP_NON_NORMAL_JOBS
+	SKIP_NON_NORMAL_JOBS = 2
 };
 
 /* return value of select_index_to_preempt function */
@@ -360,6 +360,11 @@ enum add_resource_list_flags
 	ADD_AVAIL_ASSIGNED = 8,
 	ADD_ALL_BOOL = 16
 	/* next flag 32 */
+};
+
+enum incr_decr {
+	SCHD_INCR,
+	SCHD_DECR
 };
 
 /* run update resresv flags is a bitfield = 0, 1, 2, 4, 8, ...*/
--- a/src/scheduler/fifo.c
+++ b/src/scheduler/fifo.c
@@ -1049,7 +1049,7 @@ main_sched_loop(status *policy, int sd, server_info *sinfo, schd_error **rerr)
 			schdlog(PBSEVENT_SCHED, PBS_EVENTCLASS_JOB, LOG_WARNING,
 				njob->name, "Job will never run with the resources currently configured in the complex");
 		}
-		if ((rc != SUCCESS) && njob->job->resv ==NULL) {
+		if ((rc != SUCCESS) && njob->job->resv == NULL) {
 			/* jobs in reservations are outside of the law... they don't cause
 			 * the rest of the system to idle waiting for them
 			 */
--- a/src/scheduler/fifo.c
+++ b/src/scheduler/fifo.c
@@ -2255,7 +2255,7 @@ next_job(status *policy, server_info *sinfo, int flag)
 		skip = SKIP_NOTHING;
 		sort_jobs(policy, sinfo);
 		sort_status = SORTED;
-		last_job_index = 0;
+		last_job_index = -1; /* We search from last_job_index + 1 */
 		return NULL;
 	}
 
@@ -2279,7 +2279,7 @@ next_job(status *policy, server_info *sinfo, int flag)
 		|| (flag == MUST_RESORT_JOBS)) {
 		sort_jobs(policy, sinfo);
 		sort_status = SORTED;
-		last_job_index = 0;
+		last_job_index = -1;
 	}
 	if (policy->round_robin) {
 		/* Below is a pictorial representation of how queue_list
@@ -2353,11 +2353,11 @@ next_job(status *policy, server_info *sinfo, int flag)
 		}
 	} else if (policy->by_queue) {
 		if (!(skip & SKIP_NON_NORMAL_JOBS)) {
-			ind = find_non_normal_job_ind(sinfo->jobs, last_job_index);
+			ind = find_non_normal_job_ind(sinfo->jobs, last_job_index + 1);
 			if (ind == -1) {
 				/* No more preempted jobs */
 				skip |= SKIP_NON_NORMAL_JOBS;
-				last_job_index = 0;
+				last_job_index = -1;
 			} else {
 				rjob = sinfo->jobs[ind];
 				last_job_index = ind;
@@ -2365,9 +2365,9 @@ next_job(status *policy, server_info *sinfo, int flag)
 		}
 		if (skip & SKIP_NON_NORMAL_JOBS) {
 			while(last_queue < sinfo->num_queues &&
-			     ((ind = find_runnable_resresv_ind(sinfo->queues[last_queue]->jobs, last_job_index)) == -1)) {
+			     ((ind = find_runnable_resresv_ind(sinfo->queues[last_queue]->jobs, last_job_index + 1)) == -1)) {
 				last_queue++;
-				last_job_index = 0;
+				last_job_index = -1;
 			}
 			if (last_queue < sinfo->num_queues && ind != -1) {
 				rjob = sinfo->queues[last_queue]->jobs[ind];
--- a/src/scheduler/fifo.c
+++ b/src/scheduler/fifo.c
@@ -1655,12 +1655,15 @@ run_update_resresv(status *policy, int pbs_sd, server_info *sinfo,
 		}
 
 		if (ns != NULL) {
+			int update_nodepart = 0;
 			for (i = 0; ns[i] != NULL; i++) {
 				int j;
 				update_node_on_run(ns[i], rr, &old_state);
 				if (ns[i]->ninfo->np_arr != NULL) {
-					for (j = 0; ns[i]->ninfo->np_arr[j] != NULL; j++)
+					for (j = 0; ns[i]->ninfo->np_arr[j] != NULL; j++) {
 						modify_resource_list(ns[i]->ninfo->np_arr[j]->res, ns[i]->resreq, SCHD_INCR);
+						update_nodepart = 1;
+					}
 				}
 				/* if the node is being provisioned, it's brought down in
 				 * update_node_on_run().  We need to add an event in the calendar to
@@ -1673,6 +1676,8 @@ run_update_resresv(status *policy, int pbs_sd, server_info *sinfo,
 					}
 				}
 			}
+			if (update_nodepart)
+				sort_all_nodepart(policy, sinfo);
 		}
 
 		update_queue_on_run(qinfo, rr, &old_state);
--- a/src/scheduler/fifo.c
+++ b/src/scheduler/fifo.c
@@ -1656,13 +1656,18 @@ run_update_resresv(status *policy, int pbs_sd, server_info *sinfo,
 
 		if (ns != NULL) {
 			for (i = 0; ns[i] != NULL; i++) {
+				int j;
 				update_node_on_run(ns[i], rr, &old_state);
+				if (ns[i]->ninfo->np_arr != NULL) {
+					for (j = 0; ns[i]->ninfo->np_arr[j] != NULL; j++)
+						modify_resource_list(ns[i]->ninfo->np_arr[j]->res, ns[i]->resreq, SCHD_INCR);
+				}
 				/* if the node is being provisioned, it's brought down in
 				 * update_node_on_run().  We need to add an event in the calendar to
 				 * bring it back up.
 				 */
 				if (ns[i]->go_provision) {
-					if (add_prov_event(sinfo->calendar, sinfo->server_time + PROVISION_DURATION, ns[i]->ninfo)== 0) {
+					if (add_prov_event(sinfo->calendar, sinfo->server_time + PROVISION_DURATION, ns[i]->ninfo) == 0) {
 						set_schd_error_codes(err, NOT_RUN, SCHD_ERROR);
 						return -1;
 					}
@@ -1671,8 +1676,6 @@ run_update_resresv(status *policy, int pbs_sd, server_info *sinfo,
 		}
 
 		update_queue_on_run(qinfo, rr, &old_state);
-
-		sinfo->pset_metadata_stale = 1;
 
 		update_server_on_run(policy, sinfo, qinfo, rr, &old_state);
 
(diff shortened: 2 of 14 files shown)


bhroam:
@agrawalravi90 @arungrover I've removed the WIP tag from this PR.  It is ready for review.  Let me know if you have any questions.

src/scheduler/constant.h
```diff
@@ -74,15 +74,15 @@ enum sort_status
...
 
-/* enum used to find out what to skip while searching for the next job to schedule */
+/* enum used to find out what to skip while searching for the next job to schedule.  Values are bits in a bitfield */
 
 enum skip
 {
 	SKIP_NOTHING,
 	/* Value used to know whether reservations are already scheduled or not */
-	SKIP_RESERVATIONS,
+	SKIP_RESERVATIONS = 1,
```
agrawalravi90: did you do this as part of debugging? they should get set to 1 & 2 anyways

src/scheduler/job_info.c
```diff
@@ -3252,12 +3251,9 @@ find_jobs_to_preempt(status *policy, resource_resv *hjob, server_info *sinfo, in
 
 		update_universe_on_end(npolicy, pjob,  "S", NO_ALLPART);
 		rjobs_count--;
-		if (nsinfo->calendar != NULL) {
-			te = find_timed_event(nsinfo->calendar->events, 0, pjob->name, TIMED_END_EVENT, 0);
-			if (te != NULL) {
-				if (delete_event(nsinfo, te, DE_NO_FLAGS) == 0)
+		if (pjob->end_event != NULL) {
+				if (delete_event(nsinfo, pjob->end_event, DE_NO_FLAGS) == 0)
```
agrawalravi90: might be safer to set pjob->end_event to NULL after the delete_event

src/scheduler/resource_resv.c
```diff
@@ -1330,13 +1329,10 @@ update_resresv_on_end(resource_resv *resresv, char *job_state)
 			resresv->execselect = NULL;
 		}
 		/* We need to correct our calendar */
-		if (resresv->server->calendar != NULL) {
-			te = find_timed_event(resresv->server->calendar->events, 0, resresv->name, TIMED_END_EVENT, 0);
-			if (te != NULL)
-				set_timed_event_disabled(te, 1);
-		}
+			if (resresv->end_event != NULL)
+				set_timed_event_disabled(resresv->end_event, 1);
```
agrawalravi90: when should one call delete_event() vs set_timed_event_disabled() ?

src/scheduler/resource_resv.c
```diff
@@ -1330,13 +1329,10 @@ update_resresv_on_end(resource_resv *resresv, char *job_state)
 			resresv->execselect = NULL;
 		}
 		/* We need to correct our calendar */
-		if (resresv->server->calendar != NULL) {
-			te = find_timed_event(resresv->server->calendar->events, 0, resresv->name, TIMED_END_EVENT, 0);
-			if (te != NULL)
-				set_timed_event_disabled(te, 1);
-		}
+			if (resresv->end_event != NULL)
```
agrawalravi90: the indentation looks off here

src/scheduler/server_info.c
```diff
@@ -1420,6 +1420,38 @@ create_resource(char *name, char *value, enum resource_fields field)
 	return nres;
 }
 
+int
```
agrawalravi90: please add a function header

src/scheduler/server_info.c
```diff
@@ -1420,6 +1420,38 @@ create_resource(char *name, char *value, enum resource_fields field)
...
+		return 0;
+	
+	for(cur_req = req_list; cur_req != NULL; cur_req = cur_req->next) {
+		if (cur_req->type.is_consumable) {
+			cur_res = find_resource(res_list, cur_req->def);
+			if (cur_res == NULL) {
+				if (end_res == NULL)
+					for (end_res = res_list; end_res->next != NULL; end_res = end_res->next)
+						;
+				end_res->next = create_resource(cur_req->name, cur_req->res_str, RF_AVAIL);
```
agrawalravi90: shouldn't you check here whether 'type' is SCHD_INCR and only add to the list if true?

src/scheduler/server_info.c
```diff
@@ -1420,6 +1420,38 @@ create_resource(char *name, char *value, enum resource_fields field)
...
+				if (end_res->next == NULL)
+					return 0;
+				end_res = end_res->next;
+			} else {
+				if (type == SCHD_INCR)
+					cur_res->avail += cur_req->amount;
```
agrawalravi90: just checking, 'avail' changes instead of 'assigned' ?

src/scheduler/simulate.c
```diff
@@ -621,27 +621,30 @@ perform_event(status *policy, timed_event *event)
...
+ * @param[in]	ninfo - the node to check for
+ * @param[in]	search between now and end
+ */
+int
+exists_run_event_on_node(node_info *ninfo, time_t end)
+{
+	if (ninfo == NULL || ninfo->node_events == NULL)
 		return 0;
+	
+	if (ninfo->node_events->event->event_time < end)
```
agrawalravi90: shouldn't you check here if 'event' is a run event or not?

agrawalravi90: and will there only ever be a single event in 'node_events' ?

src/scheduler/constant.h
bhroam: I added a comment to the enum to explain this.  They are now bits.  While they do get set to 1 and 2, if we need any further bits, it'd be 4.  This is why I set them directly.

src/scheduler/node_info.c
```diff
@@ -2269,7 +2273,12 @@ eval_selspec(status *policy, selspec *spec, place *placespec,
 		if (rc == 0) {
 			free_nspecs(*nspec_arr);
 			*nspec_arr = NULL;
+		} else if (resresv->server->has_multi_vnode) {
```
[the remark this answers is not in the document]
bhroam: If we don't want to take up huge amounts of memory, there is no way around this.  We need to count the nspecs to realloc() it.  I'll test it, but I think the optimization is worthwhile regardless.

src/scheduler/node_partition.c
```diff
@@ -277,6 +277,35 @@ dup_node_partition(node_partition *onp, server_info *nsinfo)
...
+{
+	int cnt;
+	int i;
+	node_partition **nnp_arr;
+
+	if (onp_arr == NULL || new_nps == NULL)
+		return NULL;
+
+	cnt = count_array((void **)onp_arr);
+	if ((nnp_arr = calloc(cnt + 1, sizeof(node_partition *))) == NULL) {
```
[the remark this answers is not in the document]
bhroam: I thought about it, but that would have meant I'd have to walk through the entire node search code and make sure I always added the NULL in the right place.  I was worried I'd miss some place.  In any case, calloc() didn't show up on the callgrind graph.  I don't think it is slowing things.  If you want, I can try and use malloc() and see if how hard it would be to modify the node search code.

src/scheduler/job_info.c
bhroam: delete_event() does that itself.

src/scheduler/resource_resv.c
bhroam: I'm not exactly sure.  The disabled event code is used in degraded reservations.  You disable events to check if you can reconfirm the reservation without losing what you already have.  I don't know tha reconfirming code all that well.  I just kept the code here the same.  Instead of searching for an end event to disable it, I just directly do it.

src/scheduler/server_info.c
bhroam: Good catch!

src/scheduler/simulate.c
bhroam: node_events only contains run events.  Since it contains all the run events, it is possible there is only one.  I'm not sure I understand the question.

agrawalravi90: ah, if it only contains end events then it makes sense that you just check for one. Can we rename the variable to "run_events" so that it's clear that these are expected to be only run events?

src/scheduler/node_partition.c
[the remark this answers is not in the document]
agrawalravi90: but you loop until 'cnt' in the loop below, so pretty much all NULL values of that array, except the last one, are being overwritten in that loop right?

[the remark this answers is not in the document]
bhroam: Oh, sorry, I thought you meant the calloc() in eval_selspec().  You're right, this one can be a malloc().

src/scheduler/simulate.c
bhroam: I'd rather not rename it.  I named it node_events because I wasn't sure if we were going to expand it in the future to contain all events on that node.

agrawalravi90: Ok, can you please add a comment here then that says that node_events only contains run events so just check the first one?

test/tests/performance/pbs_sched_perf.py
```diff
@@ -165,17 +165,27 @@ def test_node_bucket_perf_free(self):
...
+    def test_run_10000_jobs(self):
         """
         Submit 10000 jobs and time the cycle that runs all of them.
         """
 
         a = {'Resource_List.select': '1:ncpus=1'}
-        self.submit_jobs(a, 10000)
+        jids = self.submit_jobs(a, 10000)
+        t = self.run_cycle()
+        self.logger.info('#' * 80)
```
agrawalravi90: please add a check here to see that all jobs actually ran

test/tests/performance/pbs_sched_perf.py
```diff
@@ -165,17 +165,27 @@ def test_node_bucket_perf_free(self):
...
+        m = 'Time taken in cycle to run 10000 normal jobs: %.2f'
+        self.logger.info(m % (t))
+        self.logger.info('#' * 80)
+
+        for j in jids:
+            self.server.alterjob(j, {'Resource_List.place': 'excl'})
```
agrawalravi90: shouldn't these jobs already be running?

src/scheduler/simulate.c
bhroam: I just changed the comment for the member to say that.

bhroam:
Thanks for the review @agrawalravi90.  I addressed your comments and fixed the problem that I wasn't sorting the placement sets.  Now I wrote a generic function that sorts the placement sets that is used in both update_all_nodepart() (used when a job ends) and run_update_resresv() (used when a job runs).

src/scheduler/fifo.c
```diff
@@ -2349,18 +2352,18 @@ next_job(status *policy, server_info *sinfo, int flag)
...
 				last_job_index = 0;
 			} else {
 				rjob = sinfo->jobs[ind];
 				last_job_index = ind;
 			}
 		}
-		if (skip == SKIP_NON_NORMAL_JOBS) {
+		if (skip & SKIP_NON_NORMAL_JOBS) {
 			while(last_queue < sinfo->num_queues &&
 			     ((ind = find_runnable_resresv_ind(sinfo->queues[last_queue]->jobs, last_job_index)) == -1)) {
```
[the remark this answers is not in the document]
bhroam: I had already done this here in a subsequent submit.  I did forget to do this in the by_queue: false case, so it's good you pointed it out.

src/scheduler/job_info.c
```diff
@@ -3252,12 +3252,9 @@ find_jobs_to_preempt(status *policy, resource_resv *hjob, server_info *sinfo, in
...
```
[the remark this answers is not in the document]
bhroam: There are two cases where we call update_server_on_end().  The first is here, and the other is in calendaring.  In calendaring, we process and move past each event as we call update_universe_on_end().  We then don't care about that event again.  If we deleted the events in update_server_on_end(), we'd be doing an O(N) search that provides no benefit to calendaring.  It would only slow it down.  Knowing that calendaring is one of the slowest areas of the code, I think we should leave this alone.

src/scheduler/node_info.c
```diff
@@ -2237,14 +2236,19 @@ eval_selspec(status *policy, selspec *spec, place *placespec,
...
-		sizeof(nspec*))) == NULL) {
+	if (resresv->server->has_multi_vnode) {
+		/* Worst case is that split all chunks onto all nodes */
+		tot_nodes = count_array((void **)ninfo_arr);
+		num_nspecs = tot_nodes * spec->total_chunks;
+	}
+	else
+		num_nspecs = spec->total_chunks;
+
+	if ((*nspec_arr = (nspec **) calloc(num_nspecs + 1, sizeof(nspec*))) == NULL) {
```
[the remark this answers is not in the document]
bhroam: I'm going to leave the if statement in place for clarity sake.  In all other places, I have this check.  This one would be an outlier, so I'd probably have to add a comment to explain why.  I figured just to leave the if statement in.  It's harmless.

PR #1204, opened 2019-07-17
PTL: Fixing race condition around scheduling=True

agrawalravi90:
#### Describe Bug or Feature

- manager() checks that scheduler is not scheduling when scheduling is set to False. However, expect(), by default, triggers sched cycles in between each attempt. This means that manager() can potentially trigger multiple sched cycles before it catches the scheduler's state = Idle and returns.
- Often tests set scheduling=True and assume that the operation is instantaneous. This leads to race conditions in the tests which want to trigger one sched cycle completely leading to intermittent test failures.

#### Describe Your Change

- Set expect()'s trigger_sched_cycle arg to False inside manager() when checking for scheduling to be off.
- Added a convenience method called 'run_sched_cycle' in Server class to kick and finish a new scheduling cycle reliably. It takes care of any race conditions and returns a tuple of the start and end times of the sched cycle that's run, which can aid in further log matches for that cycle.

#### Link to Design Doc

http://community.pbspro.org/t/new-ptl-method-to-run-a-sched-cycle-to-completion/1704

[test_pbsfs.log](https://github.com/PBSPro/pbspro/files/3400986/test_pbsfs.log)

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -6701,10 +6701,55 @@ class Server(PBSService):
                     sname = 'default'
                 else:
                     sname = id
+
                 # Default max cycle length is 1200 seconds (20m)
                 self.expect(SCHED, {'state': 'scheduling'}, op=NE, id=sname,
-                            interval=1, max_attempts=1200)
+                            interval=1, max_attempts=1200,
+                            trigger_sched_cycle=False)
         return rc
+
+    def run_sched_cycle(self, sched="default"):
+        """
+        Convenience method to start and finish a sched cycle
+
+        :param sched - the scheduler to kick a cycle for
+        :type sched - str
+
+        :return tuple of (start time, end time) of the new sched cycle
+        """
+        old_val = self.status(SCHED, 'scheduling', id=sched)
+
+        # Make sure that we aren't in a sched cycle already
+        self.manager(MGR_CMD_SET, SCHED, {'scheduling': 'False'}, id=sched)
+
+        # Kick a new cycle
+        tbefore = time.time()
+        self.manager(MGR_CMD_SET, SCHED, {'scheduling': 'True'}, id=sched)
+        m1 = self.schedulers[sched].log_match("Starting Scheduling",
+                                              starttime=tbefore)
+        m2 = self.schedulers[sched].log_match("Leaving Scheduling",
+                                              starttime=tbefore, interval=1,
+                                              max_attempts=1200)
+
+        # Get start and end timestamps for the sched cycle
+        if self.logutils is None:
+            try:
+                from ptl.utils.pbs_logutils import PBSLogUtils
+            except:
+                _msg = 'error loading ptl.utils.pbs_logutils'
+                raise PtlLogMatchError(rc=1, rv=False, msg=_msg)
+            self.logutils = PBSLogUtils()
+        t1 = m1[1].split(";", 1)[0]
+        t1 = self.logutils.convert_date_time(t1)
+        t2 = m2[1].split(";", 1)[0]
+        t2 = self.logutils.convert_date_time(t2)
+
+        # Restore original value of scheduling
+        if old_val[0]['scheduling'] == 'False':
+            self.manager(MGR_CMD_SET, SCHED, {'scheduling': 'False'},
+                         id=sched)
+
+        return (t1, t2)
 
     def sigjob(self, jobid=None, signal=None, extend=None, runas=None,
                logerr=True):
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -10904,14 +10904,16 @@ class Scheduler(PBSService):
                             'scheduling': 'True'}, id=sched)
         self.log_match("Starting Scheduling",
                        starttime=tbefore)
-        self.log_match("Leaving Scheduling",
-                       starttime=tbefore, interval=1,
-                       max_attempts=1200)
 
-        # Restore original value of scheduling
         if old_val == 'False':
+            # This will also ensure that the sched cycle is over before
+            # returning
             self.server.manager(MGR_CMD_SET, SCHED, {'scheduling': 'False'},
                                 id=sched)
+        else:
+            self.server.expect(SCHED, {'state': 'scheduling'}, op=NE,
+                               id=sched, interval=1, max_attempts=1200,
+                               trigger_sched_cycle=False)
 
     def pbs_version(self):
         """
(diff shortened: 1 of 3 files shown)


test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -6701,11 +6701,56 @@ def manager(self, cmd, obj_type, attrib=None, id=None, extend=None,
...
+    def run_sched_cycle(self, sched="default"):
+        """
+        Convenience method to start and finish a sched cycle
+
+        :param sched - the scheduler to kick a cycle for
+        :type sched - str
+
+        :return tuple of (start time, end time) of the new sched cycle
+        """
+        old_val = self.status(SCHED, 'scheduling', id=sched)
```
bhroam: Consider adding '[0]['scheduling']' to the end.  This way old_val will be the 'True' or 'False' you need to use to set it back.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -6701,11 +6701,56 @@ def manager(self, cmd, obj_type, attrib=None, id=None, extend=None,
...
+            try:
+                from ptl.utils.pbs_logutils import PBSLogUtils
+            except:
+                _msg = 'error loading ptl.utils.pbs_logutils'
+                raise PtlLogMatchError(rc=1, rv=False, msg=_msg)
+            self.logutils = PBSLogUtils()
+        t1 = m1[1].split(";", 1)[0]
+        t1 = self.logutils.convert_date_time(t1)
+        t2 = m2[1].split(";", 1)[0]
+        t2 = self.logutils.convert_date_time(t2)
```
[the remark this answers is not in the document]
agrawalravi90: My thought was that with this callers will reliably get timestamps of the correct sched cycle. It's possible that multiple sched cycles get triggered by the time we turn scheduling off, in which case cycles(lastN=1) won't get us the correct cycle.

But callers can just use the 'start' argument of cycles() to do it. So, it should be good enough to just return the 'start' timestamp from this function, so I'll change it to do that, let me know what you think.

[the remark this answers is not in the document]
agrawalravi90: Actually, the users can just note then the time before calling run_sched_cycle, so it's not worth it. I'll remove the timestamp code.

agrawalravi90:
@bhroam I've addressed your comments, please let me know if it looks fine now. Thanks.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -10929,6 +10886,48 @@ def log_match(self, msg=None, id=None, n=50, tail=True, allmatch=False,
...
+
+        # Kick a new cycle
+        tbefore = time.time()
+        self.server.manager(MGR_CMD_SET, SCHED, {
+                            'scheduling': 'True'}, id=sched)
+        msg = self.log_match("Starting Scheduling",
+                             starttime=tbefore)
+        self.log_match("Leaving Scheduling",
+                       starttime=tbefore, interval=1,
+                       max_attempts=1200)
```
bhroam: This is not the most efficient way to check if the scheduling cycle is over.  Do an expect() on scheduling NE 'scheduling'.  That will limit the amount of string comparisons.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -10929,6 +10886,48 @@ def log_match(self, msg=None, id=None, n=50, tail=True, allmatch=False,
...
+
+        # Get start timestamp of the sched cycle
+        if self.logutils is None:
+            try:
+                from ptl.utils.pbs_logutils import PBSLogUtils
+            except:
+                _msg = 'error loading ptl.utils.pbs_logutils'
+                raise PtlLogMatchError(rc=1, rv=False, msg=_msg)
```
bhroam: This isn't really a log match error, is there a better exception to raise?  A more generic log utils one?

test/fw/ptl/lib/pbs_testlib.py
agrawalravi90: I was going for reliability over performance. I didn't do what you are suggesting because PBS might execute multiple sched cycles, so i thought that checking for the leaving sched cycle log message with starttime set correctly would be more reliable than checking for the scheduling attribute. I wish we could enhance PBS to run 'n' sched cycles, that way we could just run 1 cycle reliably, but that doesn't really make much sense for end users.

test/fw/ptl/lib/pbs_testlib.py
agrawalravi90: you are right, i think i copied this from somewhere. This really is just an import error, so I'll raise the same

agrawalravi90:
@bhroam sorry for the delay, I've addressed your comments, please provide further feedback

test/fw/ptl/lib/pbs_testlib.py
bhroam: While you are correct, your method will definitely catch the one cycle, I'm not sure it matters.  In situation 1, we check if scheduler's state != scheduling.  We miss our current cycle, and catch a subsequent one.  We eventually return.  In situation 2, run_cycle() makes sure that it matches the end of the first cycle.  It then returns back to the caller.  By the time we get control back, the scheduler has run another one.  In both cases we have run two cycles.  The reason I like checking for state != scheduling is that manager() does that.  If we went with log_match() we'd have two functions checking for the same thing in two different ways.

bhroam:
@agrawalravi90 I'm fine with the latest commit.  Once we finish our conversation above, I can sign off.

agrawalravi90:
@bhroam Ok, i changed it to check for the state since it didn't matter much. Please let me know if it looks fine now. Thanks.

agrawalravi90:
Thanks @bhroam. @arungrover can you please take this up whenever you get some time?

agrawalravi90:
@arungrover gentle reminder, thanks

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -10884,6 +10886,35 @@ def log_match(self, msg=None, id=None, n=50, tail=True, allmatch=False,
...
+        """
+        Convenience method to start and finish a sched cycle
+        """
+        sched = self.attributes['id']
+        old_val = self.server.status(SCHED, 'scheduling', id=sched)[
+            0]['scheduling']
+
+        # Make sure that we aren't in a sched cycle already
+        self.server.manager(MGR_CMD_SET, SCHED, {
+                            'scheduling': 'False'}, id=sched)
```
prakashcv13: would this stop the already running cycle?

agrawalravi90: no, it will just make sure that IF there's an existing cycle running, we'll wait until it's done

agrawalravi90:
@arungrover and @bhroam thanks for the sign-offs, can one of you please merge this?

agrawalravi90:
thanks for the merge @prakashcv13 !

PR #1210, opened 2019-07-22
Test "test_qalter_remove_files" of "TestQsub_remove_files" is failing due to race condition

Mahalty:
#### Describe Bug or Feature

- Test "test_qalter_remove_files" of "TestQsub_remove_files" is failing while verifying the job's state. 

- Intent of the test is to submit a job and alter the job's "Remove_Files" attribute to oe before the job starts running and check whether the changes are reflecting in qstat -f output.

- In the failed scenario, job started running before qalter command was issued

- In the test we are capturing the current time,rounding off the value(using int) and adding up 1 min to the above captured time and passing this value to qsub option in -a parameter(which is job's start time)

- By the time test is altering the job, the job has already started its execution hence the mismatch in the job state

#### Describe Your Change

- Change the start time of the job(current time+2 mins)

[Execution_logs_TestQsub_remove_files.txt](https://github.com/PBSPro/pbspro/files/3417858/Execution_logs_TestQsub_remove_files.txt)

### DIFF
--- a/test/tests/functional/pbs_qsub_remove_files.py
+++ b/test/tests/functional/pbs_qsub_remove_files.py
@@ -198,7 +198,7 @@ class TestQsub_remove_files(TestFunctional):
         alter the job with -Roe and check whether it is
         reflecting in qstat -f output.
         """
-        mydate = int(time.time()) + 60
+        mydate = int(time.time()) + 120
         j = Job(TEST_USER)
         attribs = {
             ATTR_a: time.strftime(
--- a/test/tests/functional/pbs_qsub_remove_files.py
+++ b/test/tests/functional/pbs_qsub_remove_files.py
@@ -198,22 +198,14 @@ class TestQsub_remove_files(TestFunctional):
         alter the job with -Roe and check whether it is
         reflecting in qstat -f output.
         """
-        mydate = int(time.time()) + 120
+        self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'False'})
         j = Job(TEST_USER)
-        attribs = {
-            ATTR_a: time.strftime(
-                '%m%d%H%M',
-                time.localtime(
-                    float(mydate)))}
-        j.set_attributes(attribs)
         jid = self.server.submit(j)
+        self.server.expect(JOB, {'job_state': 'Q'}, id=jid)
         attribs = {ATTR_R: 'oe'}
         try:
             self.server.alterjob(jid, attribs)
-            if self.server.expect(JOB, {'job_state': 'W'},
-                                  id=jid):
-                self.server.expect(JOB, attribs,
-                                   id=jid)
+            self.server.expect(JOB, attribs, id=jid)
         except PbsAlterError as e:
             print str(e)
 


dinesh1306:
Just a thought, instead of submitting job with -a can we do following steps:
- set scheduling=False
- submit simple job
- alter the job and verify qstat -f output
- set scheduling=True

because we are sure that job is in Q state

test/tests/functional/pbs_qsub_remove_files.py
```diff
@@ -198,7 +198,7 @@ def test_qalter_remove_files(self):
         alter the job with -Roe and check whether it is
         reflecting in qstat -f output.
         """
-        mydate = int(time.time()) + 60
+        mydate = int(time.time()) + 120
```
prakashcv13: we are not doing much between submitting the job and altering the attribute. 60 seconds seems enough. Are we sure that the test is failing because the job started running.

Mahalty: @prakashcv13: This issue was observed during one of the regressions. The test failed while verifying the job's state
_**Error:**_
2018-09-10 **15:39:55**,611 INFO job: executable set to /bin/sleep with arguments: 100
2018-09-10 15:39:55,612 INFOCLI x40-64-centos64: sudo -H -u pbsuser /opt/pbs/bin/qsub -a 09101540 – /bin/sleep 100
2018-09-10 15:39:59,592 INFO submit to x40-64-centos64 as pbsuser: job 1.x40-64-centos64 OrderedDict([('Execution_Time', '**09101540**')])
2018-09-10 15:40:01,033 INFO alter on x40-64-centos64: 1.x40-64-centos64 {'Remove_Files': 'oe'}
2018-09-10 15:40:01,033 INFOCLI x40-64-centos64: /opt/pbs/bin/qalter -R oe 1.x40-64-centos64
2018-09-10 15:40:01,075 INFOCLI x40-64-centos64: /opt/pbs/bin/qstat -f 1.x40-64-centos64
2018-09-10 15:40:01,146 INFO expect on server x40-64-centos64: job_state = W job 1.x40-64-centos64 got: job_state = Q
2018-09-10 15:40:01,649 INFOCLI x40-64-centos64: /opt/pbs/bin/qmgr -c set server scheduling=True
2018-09-10 15:40:01,703 INFOCLI x40-64-centos64: /opt/pbs/bin/qstat -f 1.x40-64-centos64
2018-09-10 **15:40:01**,831 INFO expect on server x40-64-centos64: job_state = W job 1.x40-64-centos64 attempt: 2 got: job_state = R
2018-09-10 15:40:02,332 INFOCLI x40-64-centos64: /opt/pbs/bin/qmgr -c set server scheduling=True

- In the test we are capturing the current time,rounding off the value(using int) and in the test we are adding up 1 min to the above captured time and passing this value to qsub option in -a parameter(which is job's start time)

- If we observe the logs, the current time is "15:39:55", that means the round off value would be "15:39", so accordingly the job's start time should be 15:40 which is 5 secs after the current time

Mahalty:
Latest execution results:
[Execution_logs_TestQsub_remove_files.txt](https://github.com/PBSPro/pbspro/files/3426764/Execution_logs_TestQsub_remove_files.txt)

test/tests/functional/pbs_qsub_remove_files.py
prakashcv13: time.time() returns no.of seconds since epoch as a float. (int) will round it off to the nearest int value - it does not return a string formatted to "hh:mm:ss" - see example below -
>>> time.time()
**1563977785.850314**
>>> int(time.time())
**1563977786**                       //Here it rounded it to a "higher" value
>>> time.time()
**1563977789.368164**
>>> int(time.time())
**1563977789**                       //Here it truncated to a "lower" value
>>> 
In the logs that you have shown, it is clear to me that time.time() returned something that was equivalent to 15:39:<some seconds>.
adding 60 to it would make it 15:40:<some seconds>

But the format that is being passed to strftime() on line 205 is '%m%d%H%M'. It ignores the seconds part completely, so we see only 15:40.
If we use '%m%d%H%M.%S', we will have 15:40:<some seconds>, the job will remain in 'W' state and that will fix the issue at hand.

However, the test case just needs the job to not be running, which can be simply achieved by disabling the scheduling of jobs, it is not about testing the '-a' option of qsub, so we can go ahead with @dinesh1306's suggestion.

Mahalty: Thanks @prakashcv13 for clarification. I have changed the test accordingly. Kindly review

prakashcv13:
merged #1210

PR #1211, opened 2019-07-23
Adding multi sched support for Scheduler::cycles()

tpai26:
#### Describe Bug or Feature
Currently scheduler.cycles() reads into the default sched_log file to read cycle info. This is to the update the cycle() to read into the sched_log file of the scheduler object which is calling the function.

[cycles_selftest.txt](https://github.com/PBSPro/pbspro/files/3455722/cycles_selftest.txt)

[multi_sched_logs.txt](https://github.com/PBSPro/pbspro/files/3448845/multi_sched_logs.txt)
[smoketests_log.txt](https://github.com/PBSPro/pbspro/files/3448846/smoketests_log.txt)

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -12176,6 +12176,10 @@ class Scheduler(PBSService):
             self.logger.error('error loading ptl.utils.pbs_logutils')
             return None
 
+        if self.id != "default":
+            day = time.strftime("%Y%m%d", time.localtime())
+            self.logfile = os.path.join(self.attributes['sched_log'], day)
+
         if start is not None or end is not None:
             analyze_path = os.path.dirname(self.logfile)
         else:
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -12178,13 +12178,16 @@ class Scheduler(PBSService):
 
         if 'sched_log' in self.attributes:
             logdir = self.attributes['sched_log']
-            tm = time.strftime("%Y%m%d", time.localtime())
-            self.logfile = os.path.join(logdir, tm)
+        else:
+            logdir = os.path.join(self.pbs_conf['PBS_HOME'], 'sched_logs')
+
+        tm = time.strftime("%Y%m%d", time.localtime())
+        log_file = os.path.join(logdir, tm)
 
         if start is not None or end is not None:
-            analyze_path = os.path.dirname(self.logfile)
+            analyze_path = os.path.dirname(log_file)
         else:
-            analyze_path = self.logfile
+            analyze_path = log_file
 
         sl = PBSSchedulerLog()
         sl.analyze(analyze_path, start, end, self.hostname)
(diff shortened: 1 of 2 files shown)


test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -12176,6 +12176,10 @@ def cycles(self, start=None, end=None, firstN=None, lastN=None):
             self.logger.error('error loading ptl.utils.pbs_logutils')
             return None
 
+        if self.id != "default":
+            day = time.strftime("%Y%m%d", time.localtime())
+            self.logfile = os.path.join(self.attributes['sched_log'], day)
+
```
bhroam: This might not be the right place to put this.  the PBSService class has init_logfile_path() which sets self.logfile.  Maybe you should overload it and set self.logfile there.  I mean self.logfile is not correctly set for the multischeds.

test/tests/selftest/pbs_cycles_test.py
```diff
@@ -0,0 +1,102 @@
...
+        jid1 = self.server.submit(j1)
+        j2 = Job(TEST_USER1, attrs=a)
+        jid2 = self.server.submit(j2)
+
+        self.server.manager(MGR_CMD_SET, SCHED,
+                            {'scheduling': 'True'}, id='sc')
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid1)
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid2)
+
+        cycle = self.scheds['sc'].cycles(start=self.server.ctime, lastN=1)[0]
```
agrawalravi90: don't you wanna set start to 'st' ?

agrawalravi90: also, don't access the index right-away. Instead, store the value in 'cycles', assert that it's not None and len(cycles) > 0, then do cycle = cycles[0]. That way, the test can fail instead of erroring out if the multisched support doesn't work.

test/tests/selftest/pbs_cycles_test.py
```diff
@@ -0,0 +1,102 @@
...
+
+        firstconsidered = cycle.political_order[0]
```
agrawalravi90: same comment here, don't access the index of political_order directly.

test/tests/selftest/pbs_cycles_test.py
tpai26: Added a assert to check if the value of cycle is greater than 0 before accessing the index

test/tests/selftest/pbs_cycles_test.py
```diff
@@ -0,0 +1,105 @@
...
+        jid1 = self.server.submit(j1)
+        j2 = Job(TEST_USER1, attrs=a)
+        jid2 = self.server.submit(j2)
+
+        self.server.manager(MGR_CMD_SET, SCHED,
+                            {'scheduling': 'True'}, id='sc')
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid1)
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid2)
+
+        cycles = self.scheds['sc'].cycles(start=st, lastN=1)
```
agrawalravi90: don't you want firstN=1 from st time instead of lastN ?

tpai26: yep, that is right.  I had also forgotten to assert the first considered is jid 1 in the test, added that to the test and attached the new log for it.

tpai26:
I have added the new log for the test I just changed. The other logs are also from the current change to the framework. I did not change the tests, so did not update them again.

PR #1217, opened 2019-07-29
Fix preemption with requeue on Cray

riyazhakki:
#### Describe Bug or Feature

*Preemption with preempt_order 'R' is not working on Cray.
(it is also reproducable on non-Cray by setting node's share attribute to force_excelhost)*

*Server replies to preemption request to scheduler immediately after sending requeue signal to mom, while mom and server still in obituary process of the preempting job.  Since scheduler receives confirmation of preemption it asks server to run the high priority job and end up with server's response resource unavailable (15044) and fails to preempt the job.*

#### Describe Your Change

*Replying to preemption batch request only after resources are released by the preempting job.*

[test_preempt_requeue_exclhost.txt](https://github.com/PBSPro/pbspro/files/3440204/test_preempt_requeue_exclhost.txt)
[Testpreemption.txt](https://github.com/PBSPro/pbspro/files/3440205/Testpreemption.txt)

### DIFF
--- a/src/server/req_jobobit.c
+++ b/src/server/req_jobobit.c
@@ -1490,6 +1490,8 @@ on_job_rerun(struct work_task *ptask)
 
 				}
 				rel_resc(pjob);		/* free resc assigned to job */
+
+				/* Respond to pending preemption request from the scheduler, if any */
 				if (pjob->ji_pmt_preq != NULL)
 					reply_preempt_jobs_request(PBSE_NONE, PREEMPT_METHOD_REQUEUE, pjob);
 
--- a/src/server/req_jobobit.c
+++ b/src/server/req_jobobit.c
@@ -627,6 +627,9 @@ rel_resc(job *pjob)
 		if (pjob->ji_rerun_preq->rq_conn != PBS_LOCAL_CONNECTION)
 			conn = get_conn(pjob->ji_rerun_preq->rq_conn);
 
+		if (pjob->ji_pmt_preq != NULL)
+			reply_preempt_jobs_request(PBSE_NONE, PREEMPT_METHOD_REQUEUE, pjob);
+
 		reply_ack(pjob->ji_rerun_preq);
 
 		/* clear no-timeout flag on connection to prevent stale connections */
--- a/src/server/req_jobobit.c
+++ b/src/server/req_jobobit.c
@@ -627,9 +627,6 @@ rel_resc(job *pjob)
 		if (pjob->ji_rerun_preq->rq_conn != PBS_LOCAL_CONNECTION)
 			conn = get_conn(pjob->ji_rerun_preq->rq_conn);
 
-		if (pjob->ji_pmt_preq != NULL)
-			reply_preempt_jobs_request(PBSE_NONE, PREEMPT_METHOD_REQUEUE, pjob);
-
 		reply_ack(pjob->ji_rerun_preq);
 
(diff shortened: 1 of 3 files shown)


src/server/req_jobobit.c
```diff
@@ -627,6 +627,9 @@ rel_resc(job *pjob)
 		if (pjob->ji_rerun_preq->rq_conn != PBS_LOCAL_CONNECTION)
 			conn = get_conn(pjob->ji_rerun_preq->rq_conn);
 
+		if (pjob->ji_pmt_preq != NULL)
+			reply_preempt_jobs_request(PBSE_NONE, PREEMPT_METHOD_REQUEUE, pjob);
+
```
bhroam: This can't go in rel_resc().  rel_resc() is a very generic function that is called whenever resources need to be released.  It is not just for requeuing jobs.  If you put it here, then any time rel_resc() is called, the preemption would be replied to thinking it was for requeue.  I think the place you want is just after the rel_resc() call is made in on_job_rerun() at line 1422 in req_jobobit.c.

test/tests/functional/pbs_preemption.py
```diff
@@ -268,9 +285,11 @@ def test_preempt_requeue_never_run(self):
         Test that a job is preempted by requeue and the scheduler does not
         report the job as can never run
         """
+        start_time = int(time.time())
```
bhroam: You no longer have to int() time().  PTL uses microsecond logging by default.

test/tests/functional/pbs_preemption.py
```diff
@@ -202,6 +202,23 @@ def test_preempt_requeue(self):
...
+        """
+        # set node share attribute to force_exclhost
+        self.server.manager(MGR_CMD_DELETE, NODE, None, "")
+        self.mom.delete_vnode_defs()
+        vnode_prefix = self.mom.shortname
+        a = {'resources_available.ncpus': '1',
+             'sharing': 'force_exclhost'}
+        vnodedef = self.mom.create_vnode_def(vnode_prefix, a, 0)
+        self.assertNotEqual(vnodedef, None)
+        self.mom.insert_vnode_def(vnodedef, 'vnode.def')
```
bhroam: Is there a reason you jump through the hoops of creating a vnodedef file and inserting it instead of using self.server.create_vnodes() ?

riyazhakki:
[Testpreemption_before_fix.txt](https://github.com/PBSPro/pbspro/files/3446740/Testpreemption_before_fix.txt)
[Testpreemption.txt](https://github.com/PBSPro/pbspro/files/3446741/Testpreemption.txt)

src/server/req_jobobit.c
riyazhakki: Code change was inside a rerun request check, which run only for requeuing job, but I do feel the better place for replying preemption is from on_job_rerun(), thanks for pointing to the right function. I have made the suggested changes.

src/server/req_jobobit.c
```diff
@@ -1490,6 +1490,9 @@ on_job_rerun(struct work_task *ptask)
 
 				}
 				rel_resc(pjob);		/* free resc assigned to job */
```
prakashcv13: only one minor suggestion - 
add a comment here saying "respond to pending preemption request from the scheduler, if any"

prakashcv13:
merged #1217

PR #1221, opened 2019-08-01
ptl.sh/csh does net set environment correctly when installed via make

sandisamp:
#### Describe Bug or Feature

Paths were not being correctly set when installed via make

#### Describe Your Change

Previously ptl.sh file was only checking for rpm installation, made it so that it checks for ptl installation as well and sets corresponding path.

[o/p log of ptl.sh](https://github.com/PBSPro/pbspro/files/3456160/ptlsh.op.txt)

### DIFF
--- a/test/fw/ptl.csh
+++ b/test/fw/ptl.csh
@@ -37,15 +37,43 @@
 
 # This file will set path variables in case of ptl installation
 
-if [ -f "/etc/debian_version" ]; then
-	ptl_prefix_lib=$( dpkg -L pbspro-ptl 2>/dev/null | grep -m 1 lib$ 2>/dev/null )
+if ( -f "/etc/debian_version" ) then
+	set ptl_prefix_lib=`dpkg -L pbspro-ptl 2>/dev/null | grep -m 1 lib$ 2>/dev/null`
 else
-	ptl_prefix_lib=$( rpm -ql pbspro-ptl 2>/dev/null | grep -m 1 lib$ 2>/dev/null )
-fi
-if [ "x${ptl_prefix_lib}" != "x" ]; then
-	python_dir=$( /bin/ls -1 ${ptl_prefix_lib} )
-	prefix=$( dirname ${ptl_prefix_lib} )
+	set ptl_prefix_lib=`rpm -ql pbspro-ptl 2>/dev/null | grep -m 1 lib$ 2>/dev/null`
+endif
+if ( ! $?ptl_prefix_lib ) then
+	set python_dir=`/bin/ls -1 ${ptl_prefix_lib}`
+	set prefix=`dirname ${ptl_prefix_lib}`
 
 	setenv PATH=${prefix}/bin/:${PATH} 
-	setenv PYTHONPATH=${prefix}/lib/${python_dir}/site-packages/:$PYTHONPATH 
-fi
+	setenv PYTHONPATH=${prefix}/lib/${python_dir}/site-packages/:$PYTHONPATH
+	unset python_dir
+	unset prefix
+	unset ptl_prefix_lib
+else
+	if ( $?PBS_CONF_FILE ) then
+		set conf = "$PBS_CONF_FILE"
+	else
+		set conf = /etc/pbs.conf
+	endif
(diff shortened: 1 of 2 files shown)


test/fw/ptl.sh
```diff
@@ -52,18 +52,18 @@ else
...
+		if [ "X${__PBS_EXEC}" != "X" ]; then
+			# Define PATH and PYTHONPATH for the users
 			PTL_PREFIX=$( dirname ${__PBS_EXEC} )/ptl
 			python_dir=$( /bin/ls -1 ${PTL_PREFIX}/lib )/site-packages
-      		[ -d "${PTL_PREFIX}/bin" ] && export PATH="${PATH}:${PTL_PREFIX}/bin"
+			[ -d "${PTL_PREFIX}/bin" ] && export PATH="${PATH}:${PTL_PREFIX}/bin"
 			[ -d "${PTL_PREFIX}/lib/${python_dir}" ] && export PYTHONPATH="${PYTHONPATH}:${PTL_PREFIX}/lib/${python_dir}"
-      		if [ `whoami` = "root" ]; then
-         		[ -d "${PTL_PREFIX}/bin" ] && export PATH="${PATH}:${PTL_PREFIX}/bin"
+			if [ `whoami` = "root" ]; then
```
hirenvadalia: Why we need this if block?

test/fw/ptl.sh
```diff
@@ -48,4 +48,22 @@ if [ "x${ptl_prefix_lib}" != "x" ]; then
 
 	export PATH=${prefix}/bin/:${PATH} 
 	export PYTHONPATH=${prefix}/lib/${python_dir}/site-packages/:$PYTHONPATH
```
hirenvadalia: do unset in this if block also

test/fw/ptl.sh
```diff
@@ -48,4 +48,22 @@ if [ "x${ptl_prefix_lib}" != "x" ]; then
...
+			PTL_PREFIX=$( dirname ${__PBS_EXEC} )/ptl
+			python_dir=$( /bin/ls -1 ${PTL_PREFIX}/lib )/site-packages
+			[ -d "${PTL_PREFIX}/bin" ] && export PATH="${PATH}:${PTL_PREFIX}/bin"
+			[ -d "${PTL_PREFIX}/lib/${python_dir}" ] && export PYTHONPATH="${PYTHONPATH}:${PTL_PREFIX}/lib/${python_dir}"
+			if [ `whoami` = "root" ]; then
+				[ -d "${PTL_PREFIX}/bin" ] && export PATH="${PATH}:${PTL_PREFIX}/bin"
+				[ -d "${PTL_PREFIX}/lib/${python_dir}" ] && export PYTHONPATH="${PYTHONPATH}:${PTL_PREFIX}/lib/${python_dir}"
+			fi
+		fi
+		unset __PBS_EXEC
```
hirenvadalia: also unset PTL_PREFIX, python_dir, and conf

prakashcv13:
@sandisamp - please remove the PBS-* ticket id from the summary.

test/fw/ptl.csh
```diff
@@ -37,15 +37,43 @@
...
+	unset prefix
+	unset ptl_prefix_lib
+else
+	if ( $?PBS_CONF_FILE ) then
+		set conf = "$PBS_CONF_FILE"
+	else
+		set conf = /etc/pbs.conf
+	endif
+	if ( -r "${conf}" ) then
+		# we are not sourcing pbs.conf file because we don't want to export any variable, just parse PBS_EXEC
```
prakashcv13: why don't we do just that, source the file and use only PBS_EXEC?

sandisamp: That would create a bunch of sourced variables in the environment. Every time the user logs in the file will be sourced which may or may not cause problems. We want not to tamper too much with the shell environment and only set the variables that we require.

prakashcv13: I don't think there will be any harm if the variables are set in the environment, but I am alright with the current logic.
The comment is too verbose, let us shorten it.

prakashcv13:
will merge once the checks pass.

PR #1223, opened 2019-08-02
'test_non_manager' of 'TestPythonRestartSettings' fails while verifying the error msg

Mahalty:
#### Describe Bug or Feature

- Test 'test_non_manager' of 'TestPythonRestartSettings' fails while verifying the error msg

#### Describe Your Change

- test_non_manager() tests that operator users and other non manager users cannot set attribute in server. But in this test the operators were never set in the server. So the error message we got was -"Unauthorized Request". 

- The error message we get when an operator user added to the server list of operators tries to set the attribute is - "Cannot set attribute, read only or insufficient permission python_restart_max_hooks"

- Now by default the operator user in the server will be pbsoper. We need to update the test and the error message accordingly.

[Execution_logs_TestPythonRestartSettings.txt](https://github.com/PBSPro/pbspro/files/3460290/Execution_logs_TestPythonRestartSettings.txt)

### DIFF
--- a/test/tests/functional/pbs_python_restart_settings.py
+++ b/test/tests/functional/pbs_python_restart_settings.py
@@ -117,51 +117,52 @@ class TestPythonRestartSettings(TestFunctional):
 
     def test_non_manager(self):
         """
-        Test that values are not set as operator or users
+        Test that hook values can not be set as operator or users.
         """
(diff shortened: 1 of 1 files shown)


VishrutArora:
@Mahalty The changes look good to me. Although, you can consider changing the docstring to "Test that hook values can not be set as operator or users."

PR #1224, opened 2019-08-02
Support PBS on Shasta

vstumpf:
#### Describe Bug or Feature
Support Cray's new Shasta Architecture

#### Describe Your Change
Used a hook to interface with Cray services

#### Link to Design Doc
https://pbspro.atlassian.net/wiki/spaces/PD/pages/1086029883/PBS+Design+Changes+for+Shasta

### DIFF
--- a/src/modules/python/pbs_hooks/PBS_cray_jacs.PY
+++ b/src/modules/python/pbs_hooks/PBS_cray_jacs.PY
@@ -273,6 +273,49 @@ def merge_dict(base, new):
     return merged
 
 
+def retry_post(data):
+    """
+    In the case where a POST fails due to a 400 error,
+    it could be because there is already a job on the cray side.
+    In that case, we should try to delete the existing job and
+    resubmit a new one.
+    """
+    event = pbs.event()
+    jid = event.job.id
+
+    joburl = HookHelper.build_path(resource='job', jobid=jid)
+    del_timeout = HookHelper.get_config()['delete_timeout']
+    try:
+        r2 = delete(joburl, timeout=del_timeout)
+        r2.raise_for_status()
+    except requests.Timeout:
+        log_with_caller(pbs.EVENT_ERROR, 'DELETE timed out')
+        raise OfflineError('Job delete timed out')
+    except requests.HTTPError:
+        # If 404, then maybe the job that was there is now gone,
+        # try posting again. Otherwise, raise an OfflineError
+        if r2.status_code != 404:
+            log_with_caller(pbs.EVENT_ERROR, 'DELETE job failed')
+            raise OfflineError('Job delete failed')
+
+    url = HookHelper.build_path(resource='job')
+    post_timeout = HookHelper.get_config()['post_timeout']
+    try:
+        r3 = post(url, json=data, timeout=post_timeout)
+        r3.raise_for_status()
+    except requests.Timeout:
+        log_with_caller(pbs.EVENT_ERROR, 'POST timed out')
+        raise OfflineError('Job POST timed out')
+    except requests.HTTPError:
+        log_with_caller(pbs.EVENT_ERROR,
+                        'Invalid status code %d' % r3.status_code)
+        raise OfflineError('Job POST encountered invalid status code')
+
+    # if we got here, we've successfully deleted and re-posted the job
+    log_with_caller(pbs.EVENT_DEBUG, 'Job %s registered' % jid)
+    return
+
+
 def handle_execjob_begin():
     """
     Handler for execjob_begin events.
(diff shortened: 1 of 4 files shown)


src/modules/python/pbs_hooks/PBS_cray_jacs.PY
```diff
@@ -0,0 +1,339 @@
...
+        raise timeout_exc
+
+
+if __name__ == '__builtin__':
+    try:
+        main()
+    except OfflineError:
+        vnode = HookHelper.get_event().vnode_list[pbs.get_local_nodename()]
+        vnode.state = pbs.ND_OFFLINE
+        vnode.comment = 'Node offlined due to Health Check Failure'
```
lisa-altair: Instead of Health Check Failure, maybe we should say "Node offlined due to timeout of PBS_cray_jacs hook"?
Or if we can, maybe even tell them which timeout (post or delete)?

src/modules/python/pbs_hooks/PBS_cray_jacs.PY
```diff
@@ -0,0 +1,339 @@
...
+    """
+    Wrapper to requests.post
+
+    Logs before and after
+    """
+    pbs.logmsg(pbs.EVENT_DEBUG2, 'Sending POST to %s' % url)
+    pbs.logmsg(pbs.EVENT_DEBUG2, 'Sending JSON: %s' % JSON.dumps(json))
+    r = requests.post(url, json=json, **kwargs)
+    pbs.logmsg(pbs.EVENT_DEBUG2, 'Received status code = %s' % r.status_code)
+    pbs.logmsg(pbs.EVENT_DEBUG2, 'Received text %s' % r.text)
```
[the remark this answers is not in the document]
vstumpf: I agree, I've made the change.

src/modules/python/pbs_hooks/PBS_cray_jacs.PY
vstumpf: According to the design, if the POST encounters a timeout, the hook rejects, it does not offline. Should I change the design?
I've added the exception message to the vnode comment.

src/modules/python/pbs_hooks/PBS_cray_jacs.PY
```diff
@@ -0,0 +1,339 @@
...
+    Return the name of the nth calling function or method.
+    """
+    return str(sys._getframe(frames).f_code.co_name)
+
+
+def log_function_name():
+    """
+    Log the caller's name
+    """
+    pbs.logmsg(pbs.EVENT_DEBUG4, '%s: Method called' % caller_name(2))
```
bhroam: Since you aren't the only pbs hook that uses this terminology in the logs, you might want to put the hook name in the log message.

src/modules/python/pbs_hooks/PBS_cray_jacs.PY
```diff
@@ -0,0 +1,339 @@
...
+
+    @classmethod
+    def get_event(cls):
+        """
+        Fetch the event if it hasn't already been fetched.
+        Return the event
+        """
+        if not cls.event:
+            cls.event = pbs.event()
+        return cls.event
```
bhroam: @bayucan can correct me, but I don't think you are saving anything by caching the event type.  It is a variable that is part of the pbs object.  You're just copying one variable into another.

src/modules/python/pbs_hooks/PBS_cray_jacs.PY
```diff
@@ -0,0 +1,339 @@
...
+    timeout = HookHelper.get_config()['delete_timeout']
+    try:
+        r = delete(url, timeout=timeout)
+        r.raise_for_status()
+    except requests.Timeout:
+        log_with_caller(pbs.EVENT_ERROR, 'POST timed out for job %s' % jid)
+        raise RejectError()
+    except requests.HTTPError:
+        log_with_caller(pbs.EVENT_ERROR, 'DELETE job failed')
+        raise RejectError('Job delete failed')
```
bhroam: You don't handle the timeout like you did above?

src/modules/python/pbs_hooks/PBS_cray_jacs.PY
lisa-altair: oh, that's right, I forgot that POST timeouts/failures doesn't offline nodes.

src/modules/python/pbs_hooks/PBS_cray_jacs.PY
vstumpf: You're right, pbs.event() just returns a PyObject. I've removed it.

src/modules/python/pbs_hooks/PBS_cray_jacs.PY
vstumpf: We talked offline, this is resolved.

src/modules/python/pbs_hooks/PBS_cray_jacs.PY
```diff
@@ -324,16 +327,17 @@ def main():
...
         main()
-    except OfflineError:
-        vnode = HookHelper.get_event().vnode_list[pbs.get_local_nodename()]
-        vnode.state = pbs.ND_OFFLINE
-        vnode.comment = 'Node offlined due to Health Check Failure'
-        HookHelper.get_event().reject()
+    except OfflineError as e:
+        for vnode in pbs.event().vnode_list.keys():
+            vnode.state = pbs.ND_OFFLINE
+            vnode.comment = '%s: Node offlined due to JACS timeout: %s' % (
```
lisa-altair: Do we have to say "due to JACS timeout"?  Also, couldn't it be due to the hook alarm?

vstumpf: I've changed it to print the message in the OfflineError exception.

src/modules/python/pbs_hooks/PBS_cray_jacs.PY
```diff
@@ -273,6 +273,49 @@ def merge_dict(base, new):
...
+        # try posting again. Otherwise, raise an OfflineError
+        if r2.status_code != 404:
+            log_with_caller(pbs.EVENT_ERROR, 'DELETE job failed')
+            raise OfflineError('Job delete failed')
+
+    url = HookHelper.build_path(resource='job')
+    post_timeout = HookHelper.get_config()['post_timeout']
+    try:
+        r3 = post(url, json=data, timeout=post_timeout)
+        r3.raise_for_status()
```
bhroam: You might want to rename these from r2 and r3 to r1 and r2 or rdelete and rpost or something.  Starting something at 2 is weird.

PR #1226, opened 2019-08-05
Fixed over-limit jobs accruing eligible time with dependencies

bhroam:
## Describe Bug or Feature
Job's over their limit do not accrue eligible time.  The bug is when such a job was the parent of a dependency (i.e. another job has an after dependency on it), the job starts accruing eligible time.

#### Describe Your Change
The server has a function called determine_accruetype().  This will set the accrue type depending on certain criteria.  There are two based on dependencies.  One is if there is a system hold on the job and it is in substate 22 to not accrue eligible time.  This is correct.  The second is if the depend attribute is set to accrue eligible time.  This is not correct.  This is because when you apply an after dependency, the parent job gets a before dependency.  This will always force the job to accrue eligible time regardless of what the scheduler thinks about the job.

I fixed this by removing check in determine_accruetype() that checked if the depend attribute is set.  I can't think of any reason why this was a good idea.

It is a little more complicated than this.  The determine_accruetype() function is only called when the job state changes or when qalter is done.  This means you have to force the function to be called after the dependency is applied to see the bug.

[eligible_before.log](https://github.com/PBSPro/pbspro/files/3469014/eligible_before.log)
[eligible_after.log](https://github.com/PBSPro/pbspro/files/3469013/eligible_after.log)

### DIFF
--- a/test/tests/functional/pbs_eligible_time.py
+++ b/test/tests/functional/pbs_eligible_time.py
@@ -48,6 +48,7 @@ class TestEligibleTime(TestFunctional):
         TestFunctional.setUp(self)
         a = {'eligible_time_enable': 'True'}
         self.server.manager(MGR_CMD_SET, SERVER, a)
+        self.accrue = {'ineligible': 1, 'eligible': 2, 'run': 3, 'exit': 4}
 
     def test_qsub_a(self):
         """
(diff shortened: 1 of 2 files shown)


test/tests/functional/pbs_eligible_time.py
```diff
@@ -148,3 +148,48 @@ def test_job_array(self):
...
+        Make sure jobs accrue eligible time (or not) approprately with an
+        after dependency
+        """
+
+        self.server.manager(MGR_CMD_SET, NODE,
+                            {'resources_available.ncpus': 2},
+                            id=self.mom.shortname)
+        J1 = Job(TEST_USER)
+        jid1 = self.server.submit(J1)
+        attribs = {'job_state': 'R', 'accrue_type': 3}
```
arungrover: Can you please use a variable instead of using numbers. This will make it clear what kind of accrue type we are looking for.

PR #1229, opened 2019-08-06
PTL should know if it's running on Shasta

ashwathraop:
#### Describe Bug or Feature

Find out a way for PTL to figure out if it is running on a Shasta machine.

#### Describe Your Change

PTL checks for a shasta specific file to identify if the system is Shasta or not. 

#### Link to Design Doc

https://pbspro.atlassian.net/wiki/spaces/PD/pages/1327235075/Support+PTL+on+Shasta

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -13123,6 +13123,15 @@ class MoM(PBSService):
         else:
             return True
 
+    def is_shasta(self):
+        """
+        Returns True if the version of PBS used is installed on Shasta platform
+        """
+        if self.platform == 'shasta':
+            return True
+        else:
+            return False
+
     def is_cpuset_mom(self):
         """
         Check for cpuset mom
--- a/test/fw/ptl/utils/pbs_dshutils.py
+++ b/test/fw/ptl/utils/pbs_dshutils.py
@@ -159,8 +159,8 @@ class DshUtils(object):
         """
         Get a local or remote platform info, essentially the value of
         Python's sys.platform, in case of Cray it will return a string
-        as "cray" for actual Cray cluster and "craysim" for Cray ALPS
-        simulator
+        as "cray" or "shasta" for actual Cray cluster and "craysim"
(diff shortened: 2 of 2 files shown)


test/fw/ptl/utils/pbs_dshutils.py
```diff
@@ -184,6 +184,10 @@ def get_platform(self, hostname=None, pyexec=None):
             else:
                 splatform = 'craysim'
             found_already = True
+        if self.isfile(hostname=hostname, path='/etc/cray/xname',
+                       level=logging.DEBUG2):
```
vstumpf: This if also needs a `if not found_already`

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -13123,6 +13123,15 @@ def is_cray(self):
...
             return True
 
+    def is_shasta(self):
+        """
+        Returns True if the version of PBS used is installed on Shasta platform
+        """
+        if self.platform == 'shasta':
+            return True
+        else:
+            return False
```
vstumpf: This could all be replaced with
`return self.platform == 'shasta'`
But it's up to you if you want to change it.

lisa-altair: I like the readability of the code as it is.  :)

ashwathraop:
[test_ptl.txt](https://github.com/PBSPro/pbspro/files/3474860/test_ptl.txt)
[test_with_shasta_file.txt](https://github.com/PBSPro/pbspro/files/3474861/test_with_shasta_file.txt)
[test_without_shasta_file.txt](https://github.com/PBSPro/pbspro/files/3474862/test_without_shasta_file.txt)

I wrote a simple ptl test which print if the machine is shasta or not.

test/fw/ptl/utils/pbs_dshutils.py
ashwathraop: found_already should not be true since we check for cray_xt before this check. So I did not see a need for that extra check.

PR #1236, opened 2019-08-09
Fix for timeouts when tests are run on system under high load

hirenvadalia:
#### Describe Bug or Feature
* When the system is under high load, PBS stop/restart operation takes time to gracefully shutdown but PBS init script only waits for 30 sec for graceful shutdown of PBS daemons and exit with error after 30 sec if PBS daemons are not shutdown
* PTL has the capability to override default max attempts and the interval between attempts but as of now, that overridden value is only used Server.expect method, but in PTL there are iterator methods like PBSService._log_match, *.isUp, etc... and use overridden value in these methods will help when tests are run on the system under high load
* User can override default test case timeout via pbs_benchpress but PTL doesn't consider that if a test case has timeout applied via `@timeout` decorator and overridden value is greater than test case timeout
* if log_match is called with existence=False then even if the log message is not found still log_match iterates till max attempts reached

#### Describe Your Change
* Increased wait time to 180 sec from 30 sec in pbs_init.d, i.e. wait for 180 sec for PBS daemons to graceful shutdown
* Made max attempts and attempt interval generic for all PBS objects and use it all iterator method
* Modified get timeout method in runner plugins such a way that it will return max(default_testcase_timeout, testcase_timeout)
* Fixed log_match not iterate again if the message is not found when existence is False

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -35,25 +35,31 @@
 # "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
 # trademark licensing policies.
 
-import sys
-import os
-import socket
-import pwd
+import copy
+import cPickle
+import datetime
 import grp
 import logging
-import time
-import re
+import os
+import pwd
 import random
+import re
+import socket
 import string
+import sys
 import tempfile
-import cPickle
-import copy
-import datetime
-import traceback
 import threading
-from operator import itemgetter
+import time
+import traceback
 from collections import OrderedDict
 from distutils.version import LooseVersion
+from operator import itemgetter
+
+from ptl.lib.pbs_api_to_cli import api_to_cli
+from ptl.utils.pbs_cliutils import CliUtils
+from ptl.utils.pbs_dshutils import DshUtils, PtlUtilError
+from ptl.utils.pbs_fileutils import FILE_TAIL, FileUtils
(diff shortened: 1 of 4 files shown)


hirenvadalia:
@PBSPro/pbspro-maintainers please do needful for this PR... thanks

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -3354,9 +3390,17 @@ class PBSObject(object):
...
     """
 
+    logger = logging.getLogger(__name__)
     utils = BatchUtils()
     platform = sys.platform
 
+    ptl_conf = {
+        'mode': PTL_API,
+        'max_attempts': 60,
+        'attempt_interval': 0.5,
```
mike0042: Should these values be set based on object.options['PTL_MAX_ATTEMPTS'] and object.options['PTL_ATTEMPT_INTERVAL'] as opposed to hardcoding them? I noticed that the same values are defined in two places.

hirenvadalia: These are default values if a user or any PTL command doesn't call `PtlConfig()` and options dict is only accessible in PtlConfig class not in PBSObject or its child

hirenvadalia: @mike0042 Sorry for pushing but this PR is little urgent for other dependent work so if you don't have any further comment then can you please do needfull...

PR #1239, opened 2019-08-12
update 'test_finished_jobs' from 'SmokeTest' to verify job used resources

dinesh1306:
#### Describe Bug or Feature

update test "test_finished_jobs" from "SmokeTest" to verify job's Resources_used values.
Currently test check that job is finished or not. Need to check job is finished with Exit_status=0 and Resources_used values should be updated.

#### Describe Your Change

Update test to check Exit_status is 0 and Resources_used values are updated.
submit job with ncpus, memory and cput so we can verify Resources_used value for those attributes also.

[res_test_finished_jobs.txt](https://github.com/PBSPro/pbspro/files/3492292/res_test_finished_jobs.txt)

### DIFF
--- a/test/tests/pbs_smoketest.py
+++ b/test/tests/pbs_smoketest.py
@@ -335,19 +335,23 @@ class SmokeTest(PBSTestSuite):
         self.server.manager(MGR_CMD_SET, NODE, a, self.mom.shortname)
         a = {'job_history_enable': 'True'}
         self.server.manager(MGR_CMD_SET, SERVER, a)
-        a = {'Resource_List.select': '1:ncpus=2:mem=20mb'}
+        a = {'Resource_List.select': '1:ncpus=2:mem=20mb',
+             'Resource_List.walltime': '15'}
         j = Job(TEST_USER, a)
-        j.set_sleep_time(15)
         j.create_eatcpu_job(15)
         jid = self.server.submit(j)
         self.server.expect(JOB, {'job_state': 'F'}, extend='x', offset=15,
                            interval=1, id=jid)
-        a = {'resources_used.ncpus': 2, 'resources_used.cpupercent': 100,
-             'resources_used.cput': '00:00:15', 'Exit_status': '0'}
-        self.server.expect(JOB, a, id=jid, extend='x', attrop=PTL_AND)
-        a = {'resources_used.walltime': '00:00:00',
-             'resources_used.mem': '0kb'}
-        self.server.expect(JOB, a, id=jid, extend='x', op=NE, attrop=PTL_AND)
+        jobs = self.server.status(JOB, id=jid, extend='x')
+        expect_val = {}
+        expect_val[ATTR_used+'.ncpus'] = '2'
+        expect_val[ATTR_used+'.cpupercent'] = '100'
+        expect_val[ATTR_used+'.cput'] = '00:00:15'
+        expect_val[ATTR_exit_status] = '0'
+        for key in expect_val:
+            self.assertEquals(expect_val[key], jobs[0][key])
+        self.assertNotEquals(jobs[0][ATTR_used+'.walltime'], '00:00:00')
+        self.assertNotEquals(jobs[0][ATTR_used+'.mem'], '0kb')
 
     def test_project_based_limits(self):
         """
--- a/test/tests/pbs_smoketest.py
+++ b/test/tests/pbs_smoketest.py
@@ -331,27 +331,26 @@ class SmokeTest(PBSTestSuite):
         """
         Test for finished jobs and resource used for jobs.
         """
-        a = {'resources_available.ncpus': '4'}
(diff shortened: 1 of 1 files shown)


test/tests/pbs_smoketest.py
```diff
@@ -329,18 +329,25 @@ def test_limits(self):
...
         jid = self.server.submit(j)
-        self.server.expect(JOB, {'job_state': 'F'}, extend='x', offset=5,
+        self.server.expect(JOB, {'job_state': 'F'}, extend='x', offset=15,
                            interval=1, id=jid)
+        a = {'resources_used.ncpus': 2, 'resources_used.cpupercent': 100,
+             'resources_used.cput': '00:00:15', 'Exit_status': '0'}
+        self.server.expect(JOB, a, id=jid, extend='x', attrop=PTL_AND)
+        a = {'resources_used.walltime': '00:00:00',
+             'resources_used.mem': '0kb'}
+        self.server.expect(JOB, a, id=jid, extend='x', op=NE, attrop=PTL_AND)
```
bhroam: I'd convert these expect() calls to one status() call and asserts.  The job is finished.  The values are not going to change.  Either they will be set the first time expect checks or not.  There is no reason to wait 30s to fail the test.  Just do one call to status(), and then do self.assertEquals() for the values you care about.

dinesh1306:
@bhroam , I have incorporate review comment. Please take a look.
[res_new_test_job_finished.txt](https://github.com/PBSPro/pbspro/files/3496510/res_new_test_job_finished.txt)

Klovely:
LGTM

test/tests/pbs_smoketest.py
```diff
@@ -329,18 +329,29 @@ def test_limits(self):
...
         """
         a = {'resources_available.ncpus': '4'}
         self.server.manager(MGR_CMD_SET, NODE, a, self.mom.shortname)
         a = {'job_history_enable': 'True'}
         self.server.manager(MGR_CMD_SET, SERVER, a)
-        a = {'Resource_List.walltime': '10', ATTR_k: 'oe'}
-        j = Job(TEST_USER, attrs=a)
-        j.set_sleep_time(5)
+        a = {'Resource_List.select': '1:ncpus=2:mem=20mb',
+             'Resource_List.walltime': '15'}
```
prakashcv13: we can remove walltime from select as we are only checking for it to be non-zero.

test/tests/pbs_smoketest.py
```diff
@@ -329,18 +329,29 @@ def test_limits(self):
...
+        j = Job(TEST_USER, a)
+        j.create_eatcpu_job(15)
```
prakashcv13: do we really need to increase the time from 5 to 15? the smoketest runs every time we execute a commit to github.

test/tests/pbs_smoketest.py
prakashcv13: In fact, we can have select to only be "ncpus=2"

test/tests/pbs_smoketest.py
prakashcv13: based on the offline discussion - 
resources_used values are getting updated only after 10 seconds, so we will keep the time to be 15 seconds.

test/tests/pbs_smoketest.py
```diff
@@ -329,18 +329,29 @@ def test_limits(self):
     @skipOnCpuSet
     def test_finished_jobs(self):
         """
-        Test for finished jobs
+        Test for finished jobs and resource used for jobs.
         """
         a = {'resources_available.ncpus': '4'}
```
prakashcv13: based on offline discussion - we reduce ncpus to 2.

test/tests/pbs_smoketest.py
```diff
@@ -329,18 +329,29 @@ def test_limits(self):
...
+        j = Job(TEST_USER, a)
+        j.create_eatcpu_job(15)
         jid = self.server.submit(j)
-        self.server.expect(JOB, {'job_state': 'F'}, extend='x', offset=5,
+        self.server.expect(JOB, {'job_state': 'F'}, extend='x', offset=15,
                            interval=1, id=jid)
+        jobs = self.server.status(JOB, id=jid, extend='x')
+        expect_val = {}
+        expect_val[ATTR_used+'.ncpus'] = '2'
+        expect_val[ATTR_used+'.cpupercent'] = '100'
```
prakashcv13: cpu utilization may not always be 100, so let us move this to assertNotEquals

dinesh1306:
@prakashcv13 , I have incorporate review comments. Please take a look.
[new_res_test_job_finished.txt](https://github.com/PBSPro/pbspro/files/3500469/new_res_test_job_finished.txt)

prakashcv13:
i have restarted the failed travis build.

prakashcv13:
merged #1239

PR #1243, opened 2019-08-13
Changed class name PtlTestRunner to PtlTextTestRunner

Klovely:
#### Describe Bug or Feature
The ptl_test_runner.py file contains two classes with the same name and different capitalization.  
ptl_test_runner.py contains class PtlTestRunner and  class PTLTestRunner
They do different things which is  confusing.  
We  should probably name these two classes differently.

#### Describe Your Change
Changed class name PtlTestRunner to PtlTextTestRunner in ptl_test_runner.py file

### DIFF
--- a/test/fw/ptl/utils/plugins/ptl_test_runner.py
+++ b/test/fw/ptl/utils/plugins/ptl_test_runner.py
@@ -410,7 +410,7 @@ class _PtlTestResult(unittest.TestResult):
         self.logger.info('\n'.join(msg))
 
 
-class PtlTestRunner(TextTestRunner):
+class PtlTextTestRunner(TextTestRunner):
 
     """
     Test runner that uses ``PtlTestResult`` to enable errorClasses,
@@ -529,7 +529,7 @@ class PTLTestRunner(Plugin):
         """
         Prepare test runner
         """
-        return PtlTestRunner(verbosity=3, config=self.config)
+        return PtlTextTestRunner(verbosity=3, config=self.config)
 
     def prepareTestResult(self, result):
         """


test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -410,7 +410,7 @@ def printSummary(self):
         self.logger.info('\n'.join(msg))
 
 
-class PtlTestRunner(TextTestRunner):
+class PtlTextTestRunner(TextTestRunner):
```
prakashcv13: shall we capitalise Ptl to PTL like it is for PTLTestRunner.

agrawalravi90: I don't really like this name. It doesn't intuitively make sense and it makes it closely related to the nose framework, which we might change in the future. So, can we name this to mean something closer to its purpose? What is this class used for? @hirenvadalia can you provide some insights?

prakashcv13: I didn't like the name as well, but like PTLTestRunner which is derived from TestRunner, I thought we can have PTLTextTestRunner which is derived from TextTestRunner.

agrawalravi90: TestRunner makes sense intuitively, it runs tests, TextTestRunner doesn't really make sense, and it's a nose functionality, so if we decide to go away from nose, TextTestRunner will be replaced with something else. But, now that I think about it, we might just remove 'PtlTextTestRunner' entirely if we move away from nose, so maybe it's okay to name it closely related to nose. Again, I'd like @hirenvadalia to comment once as he knows the history behind this code

hirenvadalia: I don't mind keeping `PtlTextTestRunner` (not PTLTextTestRunner as PTL* naming used for nose plugin name).
Also, @agrawalravi90 and @prakashcv13 `TextTestRunner` name is not nose functionality specific but its unittest functionality. see https://docs.python.org/2/library/unittest.html#unittest.TextTestRunner

agrawalravi90: Alright, thanks for clarifying

prakashcv13:
merged #1243

PR #1246, opened 2019-08-15
Support PTL to run non-root commands from other hosts using IP and port

shilpakodli:
#### Describe Bug or Feature
- PTL doesn't have support to run PBS non-root commands from remote hosts provided IP address and port of remote host.
- This requires pbs_benchpress to accept test-users in -p parameter as "username@IP+port"
- PTL has to check username provided in "-p test-users" param on given IP with port number.

#### Describe Your Change
- Changed _set_user function in pbs_testsuite.py to update users list as {username : IP+port} only when username is provided with IP and port.
- Added a check in check_user_exist in pbs_testsuite.py to see if that specific user exist on the specified IP with port number.
- Made changes in run_cmd in pbs_dshutils.py to ssh to remote hosts with IP and port associated with thai username.
- Moved PbsUser and PbsGroup out of pbs_testlib to utils.ptl.pbs_testusers and all of pre-defined users are moved from pbs_testsuite.py to pbs_testusers.py.
  Reason behind this change is in run_cmd we need values of hostname and port which are set by _set_user in pbs_testsuite.py in order to run a command on remote host. If we import pbs_testuite's modules inside pbs_dshutils.py then it will create a circular reference as pbs_testsuite itself importing DshUtils inside it. 
In order to avoid that, we are creating a seperate file which has class along with users created.


vstumpf:
The check_user_exists call will fail because root can't ssh to the remote hosts.
DshUtils.check_user_exists:607, the run_cmd needs to be run as the user being checked.

hirenvadalia:
@shilpakodli As discussed offline, it will be a good idea to put IP/hostname and port in PbsUser class only instead of replacing the name with dict

shilpakodli:
> The check_user_exists call will fail because root can't ssh to the remote hosts.
> DshUtils.check_user_exists:607, the run_cmd needs to be run as the user being checked.

Hi @vstumpf , since we are making changes as generic, I am passing runas as current user if runas is not passed in run_cmd. Let me know your thoughts.

shilpakodli:
@hirenvadalia Addressed review comments.

test/fw/ptl/utils/pbs_dshutils.py
```diff
@@ -932,24 +923,10 @@ def run_cmd(self, hosts=None, cmd=None, sudo=False, stdin=None,
...
-
         for hostname in hosts:
-            if not check_runas:
-                hostname_port = hostname.split('+')
-                hostname = hostname_port[0]
-                port = None
-                if len(hostname_port) > 1:
-                    port = hostname_port[1]
+            if (runas is not None) and (runas in PBS_ALL_USERS):
+                    hostname = runas.host
```
hirenvadalia: I still don't understand this... Why we are overriding user given hostname? here `hosts` is list of hosts on which user wants to run command.
I understand the port part that ssh might be running on a given host on a different port than default (22) so while running the command we need to use given port but why hostname, by overriding hostname, we are forcing run_cmd to run all command for a given user to only one host.

test/fw/ptl/utils/pbs_dshutils.py
```diff
@@ -932,24 +923,10 @@ def run_cmd(self, hosts=None, cmd=None, sudo=False, stdin=None,
...
+                    port = runas.port
```
hirenvadalia: runas can be uid or str (as username) you need to handle that also.

test/fw/ptl/utils/pbs_testusers.py
```diff
@@ -209,3 +217,8 @@ def __int__(self):
 PBS_ROOT_USERS = (PBSROOT_USER, ROOT_USER)
 
 PBS_BUILD_USERS = (BUILD_USER,)
+
+PBS_TEST_USERS = (TEST_USER, TEST_USER1, TEST_USER2, TEST_USER3)
```
hirenvadalia: Please rename this variable to something else like `REQUIRED_USERS` as these are not *only* test users, tests users are what we have defined in PBS_USERS, this list is a list which specifies a list of users which must exist in systems.

Also why we are checking the only a subset of PBS_USERS for existence? Shouldn't all user/group define in this file should exist in the system? By checking a subset of PBS_USERS for existence, we are saying that we don't need other users specified in PBS_USERS so if that's the case then we should remove those users from PTL but that is not the case...

test/fw/ptl/utils/pbs_dshutils.py
hirenvadalia: So I understand the requirement correctly then what we need it hostname vs ip+port mapping, correct? Like for hostname xyz IP is 1.2.3.4 and port is 8022 so whenever run_cmd try to run a command on host xyz it should IP 1.2.3.4 and port 8022 instead xyz... correct?

If my understanding is correct here then we are completely changing wrong place

vstumpf: > we are forcing run_cmd to run all command for a given user to only one host.

That's what we want, all run_cmds for non-root users must be ran on their specific host (what I'll call a userhost) .

The userhosts have no hostname, they can only be accesed by an ip and a port. So if a user wants to run a command, it HAS TO run on their userhost.

vstumpf:
> Hi @vstumpf , since we are making changes as generic, I am passing runas as current user if runas is not passed in run_cmd. Let me know your thoughts

The current user (root) can't access the userhosts, only the user that owns it. So the ssh will fail. If you pass the user you're checking as runas, the other code you're checking in will make it work.

test/fw/ptl/utils/pbs_dshutils.py
hirenvadalia: In that case I don't think we can make this change for all platform becase Linux doesn't work that way, right? so we have to make this change specific to platform who uses configuration which you said (user's commands must be run on userhost)

test/fw/ptl/utils/pbs_testusers.py
shilpakodli: Was there any reason why we were checking initially only 4 TEST_USERS ? I have still kept this code as it is until we decide why it was there before and why it doesn't include all users.

shilpakodli:
@hirenvadalia and @vstumpf I have addressed your review comments and made changes specific to shasta as it has its own username+host requirement.

test/fw/ptl/utils/pbs_dshutils.py
vstumpf: If it's only for shasta that's fine, we can use the DshUtils.get_platform() function to determine if the machine is a shasta.

shilpakodli: @vstumpf I have already implemented this change. Please review.

vstumpf:
There's a bug with `Server.submit()`, PTL calls `pwd.getpwnam(username)`, but the user doesn't exist where PTL is running. On shasta, we should not do the `os.chdir(submit_dir)` in `submit`.

vstumpf:
Another issue, when I do get around that bug, the jobs aren't being submitted correctly.
My setup is as follows:
shecil - server sched comm
katrinn - mom, clients

create 5 users on katrinn that aren't on shecil (pbstest1-5)

run benchpress as root `pbs_benchpress -t SmokeTest.test_fairshare -p test-users=pbstest1@katrinn+22:pbstest2@katrinn+22:pbstest3@katrinn+22:pbstest4@katrinn+22:pbstest5@katrinn+22,mom=katrin`

When the test submits jobs, it does not do it as the user, instead as the user running the test.
```
2019-08-20 13:02:34,665 INFO     job: executable set to /bin/sleep with arguments: 100
2019-08-20 13:02:34,665 INFOCLI  shecil: /opt/pbs/bin/qsub -l select=1:ncpus=4 -- /bin/sleep 100
2019-08-20 13:02:34,699 INFO     submit to shecil as pbstest2: job 163.shecil OrderedDict([('Resource_List.select', '1:ncpus=4')])
```

```
2019-08-20 13:02:44,909 INFO     job: executable set to /bin/sleep with arguments: 100
2019-08-20 13:02:44,909 INFOCLI  shecil: /opt/pbs/bin/qsub -- /bin/sleep 100
2019-08-20 13:02:44,913 INFO     submit to shecil as pbstest3: job 173.shecil
```

This is because submit passes the username instead of PbsUser.

I have an idea for fixing this, in run_cmd. We can loop through PBS_ALL_USERS and look for the PbsUser object that matches the username/uid passed in runcmd.

I have made some changes on an older version, here's what I have: https://github.com/PBSPro/pbspro/commit/16a9a0a462eb1691139d1a255afa48a78f700eff

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -561,8 +497,8 @@ def check_users_exist(cls):
         Check whether the user is exist or not
         """
         testusersexist = True
-        for u in [TEST_USER, TEST_USER1, TEST_USER2, TEST_USER3]:
-            rv = cls.du.check_user_exists(str(u))
+        for u in PBS_ALL_USERS:
```
vstumpf: Do we have to check for all users? I don't recommend changing this to PBS_ALL_USERS. REQUIRED_USERS is enough.

hirenvadalia: @vstumpf I don't agree with you here because as per PTL all users should exist. If you think that all users/groups defined in PTL is not required then we should remove it completely from PTL (not just this check)... But changing users/groups in PTL will be separate stuff than what's this PR is doing...

We define users/groups but we don't check them then what's point of defining them?

test/fw/ptl/utils/pbs_dshutils.py
```diff
@@ -920,6 +933,11 @@ def run_cmd(self, hosts=None, cmd=None, sudo=False, stdin=None,
         ret = {'out': '', 'err': '', 'rc': 0}
 
         for hostname in hosts:
+            if platform == "shasta":
+                if (runas is not None) and isinstance(_runas_user, PbsUser):
+                    if _runas_user in PBS_ALL_USERS:
+                        hostname = _runas_user.host
```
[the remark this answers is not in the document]
hirenvadalia: Also you can merge all 3 if in one if...

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -5962,10 +5854,14 @@ def submit(self, obj, script=None, extend=None, submit_dir=None):
             return None
 
         if submit_dir is None:
```
hirenvadalia: you can replace this if block as follow
```py
if not submit_dir and self.du.get_platform() != 'shasta':
   submit_dir = pwd.getpwnam(obj.username)[5]
```

shilpakodli: I realized that, this will fail if platform is non-shasta and submit_dir is not none, then it will go in else condition and assigns submit_dir to None which is incorrect. 
Proper if-else condition would be 
```suggestion
       if submit_dir and self.du.get_platform == 'shasta':
            submit_dir = None
       else:
            submit_dir = pwd.getpwnam(obj.username)[5]
```
Agree ?

hirenvadalia: @shilpakodli I was talking about replacing if including internal if-else block. So we want to modify submit dir only when submit_dir is None and platform != 'shasta' right? which `if not submit_dir and self.du.get_platform() != 'shasta':` condition will do

test/fw/ptl/utils/pbs_testsuite.py
vstumpf: I agree the changes of users/groups should be in a different PR. For now, we should put this back to the original 4 users.

hirenvadalia: > I agree the changes of users/groups should be in a different PR.

Thank you.

> For now, we should put this back to the original 4 users.

I don't agree with this as previously PTL was doing the wrong thing by checking only 4 users and so we changed since we are changing the respective area in code.
We should change back to the previous 4 users check only if we are facing any problem which checking all users (which you will if you are creating proper users and groups before running PTL).
@vstumpf Saying above leads me to ask you a question, why are you insisting to put back check for only 4 users? Are you facing any problem having a check for all users?

kjakkali: @vstumpf : We might need only few users for test cases from but in overall framework we need all the required users to exists . So i agree with @hirenvadalia and @shilpakodli  here. 

Lets fix it instead of postponing.

shilpakodli:
Attaching smoke test results. 
[smoketest_results.txt](https://github.com/PBSPro/pbspro/files/3570879/smoketest_results.txt)

PR #1249, opened 2019-08-16
installed ptl (via make install) is not usable because missing __init__.py in ptl folder && PTL rpm does not include utils/jobs dir

sandisamp:
#### Describe Bug or Feature

* 1->The \_\_init\_\_.py file is not being installed via make when ran from a different folder
* 2->The newly added jobs/utils folder was missing from makefile, thus the files were missing from rpm.

#### Describe Your Change

* 1->It was not able to find the file in the specified directory, made it so that it is able to locate the correct file.
* 2-> Made suitable changes to the makefile to include these files.

* Found after make install:
~/git_chng/pbspro$ ls /opt/ptl/lib/python2.7/site-packages/ptl/
\_\_init_\_\.py  \_\_init_\_\.pyc  \_\_init\_\_.pyo  lib  utils
* [sanidhya@centos7 commercial_build]$ rpm -qlp buildwork/17.2.0.20190816143014/cdrom/pbspro-ptl-17.2.0.20190816143014-0.el7.x86_64.rpm | grep utils/j
/opt/ptl/lib/python2.7/site-packages/ptl/utils/jobs
/opt/ptl/lib/python2.7/site-packages/ptl/utils/jobs/eatcpu.py
/opt/ptl/lib/python2.7/site-packages/ptl/utils/jobs/eatcpu.pyc
/opt/ptl/lib/python2.7/site-packages/ptl/utils/jobs/eatcpu.pyo

* Find both logs in comments.

### DIFF
--- a/test/fw/Makefile.am
+++ b/test/fw/Makefile.am
@@ -59,6 +59,10 @@ ptlpkg_pylib_pluginsdir = $(ptlpkg_pylib_utilsdir)/plugins
 
 dist_ptlpkg_pylib_plugins_PYTHON = $(wildcard $(srcdir)/ptl/utils/plugins/*.py)
 
+ptlpkg_pylib_jobsdir = $(ptlpkg_pylib_utilsdir)/jobs
+
+dist_ptlpkg_pylib_jobs_PYTHON = $(wildcard $(srcdir)/ptl/utils/jobs/*.py)
+
 sysprofiledir = /etc/profile.d
 
 dist_sysprofile_DATA = \
--- a/test/fw/Makefile.am
+++ b/test/fw/Makefile.am
@@ -45,7 +45,7 @@ dist_ptlpkg_bin_SCRIPTS = $(wildcard $(srcdir)/bin/*)
 
 ptlpkg_pylib_topdir = ${ptl_prefix}/lib/python$(PYTHON_VERSION)/site-packages/ptl
 
-dist_ptlpkg_pylib_top_PYTHON = ptl/__init__.py
+dist_ptlpkg_pylib_top_PYTHON = $(wildcard $(srcdir)/ptl/*.py $(builddir)/ptl/*.py)
 
 ptlpkg_pylib_libdir = $(ptlpkg_pylib_topdir)/lib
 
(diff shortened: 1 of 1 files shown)


test/fw/Makefile.am
```diff
@@ -45,7 +45,7 @@ dist_ptlpkg_bin_SCRIPTS = $(wildcard $(srcdir)/bin/*)
 
 ptlpkg_pylib_topdir = ${ptl_prefix}/lib/python$(PYTHON_VERSION)/site-packages/ptl
 
-dist_ptlpkg_pylib_top_PYTHON = $(wildcard $(srcdir)/ptl/*.py)
+dist_ptlpkg_pylib_top_PYTHON = ptl/__init__.py
```
hirenvadalia: I think you should do like this: `$(wildcard $(srcdir)/ptl/*.py $(builddir)/ptl/*.py)`

sandisamp:
Adding install logs:
[install_log.txt](https://github.com/PBSPro/pbspro/files/3508769/install_log.txt)

sandisamp:
Adding build log for new rpm generated:
[Build_log.txt](https://github.com/PBSPro/pbspro/files/3508877/Build_log.txt)

test/fw/Makefile.am
```diff
@@ -45,7 +45,7 @@ dist_ptlpkg_bin_SCRIPTS = $(wildcard $(srcdir)/bin/*)
 
 ptlpkg_pylib_topdir = ${ptl_prefix}/lib/python$(PYTHON_VERSION)/site-packages/ptl
 
-dist_ptlpkg_pylib_top_PYTHON = $(wildcard $(srcdir)/ptl/*.py)
+dist_ptlpkg_pylib_top_PYTHON = $(wildcard $(srcdir)/ptl/*.py $(builddir)/ptl/*.py)
```
hirenvadalia: on second thought why don't we do this change for all dir? it will be useful in future if user add anything other dirs which we need to pick from builddir instead srcdir

prakashcv13:
@sandisamp - I agree with @hirenvadalia that we can do this change for all directories.

sandisamp:
Since we don't have other dir in builddir it fails, also i tried putting ifelse which doesn't work, so when any user adds in future they have to modify makefile also.

hirenvadalia:
> Since we don't have other dir in builddir it fails, also i tried putting ifelse which doesn't work, so when any user adds in future they have to modify makefile also.

I am ok with this...

prakashcv13:
merged #1249

PR #1250, opened 2019-08-16
fix typo in sudo_cmd

kjakkali:
#### Describe Bug or Feature
PBSLogUtils doesn't have sudo_cmd attribute so the cat command was not prefixing with sudo command

#### Describe Your Change

sudo_cmd is attribute of DshUtils class so add missing object

### DIFF
--- a/test/fw/ptl/utils/pbs_logutils.py
+++ b/test/fw/ptl/utils/pbs_logutils.py
@@ -296,7 +296,7 @@ class PBSLogUtils(object):
         try:
             if hostname is None or self.du.is_localhost(hostname):
                 if sudo:
-                    cmd = copy.copy(self.sudo_cmd) + ['cat', log]
+                    cmd = copy.copy(self.du.sudo_cmd) + ['cat', log]
                     self.logger.info('running ' + " ".join(cmd))
                     p = Popen(cmd, stdout=PIPE)
                     f = p.stdout


test/fw/ptl/utils/pbs_logutils.py
```diff
@@ -296,7 +296,7 @@ def open_log(self, log, hostname=None, sudo=False):
         try:
             if hostname is None or self.du.is_localhost(hostname):
                 if sudo:
-                    cmd = copy.copy(self.sudo_cmd) + ['cat', log]
+                    cmd = copy.copy(self.du.sudo_cmd) + ['cat', log]
```
bhroam: This might be better for another ticket, but if you have access to du inside of log utils, why not use self.du.cat instead of directly calling Popen()?

prakashcv13:
restarted the AppVeyor build as only Debug build was failing.

test/fw/ptl/utils/pbs_logutils.py
hirenvadalia: @bhroam open_log returns File like instance which caller can iterate to get lines one by one. Where DshUtils.cat returns contents of the file as a list of strings instead File like object.

But saying that we can do one thing here as the list is also iterable we can either return a list of lines or if we still want File like object after using self.du.cat then we can make use of StringIO to get File like object from string.

hirenvadalia: > From what I know, File objects are better than lists because they take less memory.

True


> I don't see the File object we return helping us any more than returning a list. Please correct me if I am wrong.

True

agrawalravi90:
It seems like this was approved a month back, any reason why we haven't merged this yet?

kjakkali:
@agrawalravi90   :The fix went in as part of [https://github.com/PBSPro/pbspro/pull/1268](#1268 ).
I forgot close this PR.

kjakkali:
closing this PR

PR #1252, opened 2019-08-17
build fails as make attempts to copy multiple __init__.py

sandisamp:
#### Describe Bug or Feature

Build was failing as make was  trying to copy __init__.py file multiple times.

#### Describe Your Change

Removed $(srcdir)/ptl/__init__.py and now taking it only from $(builddir)

[build.txt](https://github.com/PBSPro/pbspro/files/3511608/build.txt)

### DIFF
--- a/test/fw/Makefile.am
+++ b/test/fw/Makefile.am
@@ -45,7 +45,7 @@ dist_ptlpkg_bin_SCRIPTS = $(wildcard $(srcdir)/bin/*)
 
 ptlpkg_pylib_topdir = ${ptl_prefix}/lib/python$(PYTHON_VERSION)/site-packages/ptl
 
-dist_ptlpkg_pylib_top_PYTHON = $(wildcard $(srcdir)/ptl/*.py $(builddir)/ptl/*.py)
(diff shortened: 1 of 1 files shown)


test/fw/Makefile.am
```diff
@@ -45,7 +45,7 @@ dist_ptlpkg_bin_SCRIPTS = $(wildcard $(srcdir)/bin/*)
 
 ptlpkg_pylib_topdir = ${ptl_prefix}/lib/python$(PYTHON_VERSION)/site-packages/ptl
 
-dist_ptlpkg_pylib_top_PYTHON = $(wildcard $(srcdir)/ptl/*.py $(builddir)/ptl/*.py)
+dist_ptlpkg_pylib_top_PYTHON = $(wildcard $(builddir)/ptl/*.py)
```
hirenvadalia: what will happen if `srcdir` has other `.py` files? those will be missed in installation or rpm right?

sandisamp:
The build is failing in pbspro-extensions Makefile, Closing this PR as OSS is working correctly and we should modify the extensions Makefile accordingly.

PR #1260, opened 2019-08-21
Allow jobs with $reject_root_scripts to get held

bayucan:
#### Describe Bug or Feature

* Jobs submitted as root with job script, and mom's config file has $reject_root_scripts turned true, indefinitely gets reject and rerun.
* It's expected for the job to get rejected, but it should also eventually get held after too many retries (21).

#### Describe Your Change

* When server received a PBSE_REJECT_ROOT_SCRIPTS error from mom trying to start a job, the code for checking failed attempts must be called and allow job to be held after too many retries.

* [ptl.root_owned_scripts.td-08.txt](https://github.com/PBSPro/pbspro/files/3526809/ptl.root_owned_scripts.td-08.txt)
* [ptl.pbs_root_owned_script.PASS.txt](https://github.com/PBSPro/pbspro/files/3526826/ptl.pbs_root_owned_script.PASS.txt)
* [ptl.pbs_root_owned_script.FAIL.txt](https://github.com/PBSPro/pbspro/files/3526827/ptl.pbs_root_owned_script.FAIL.txt)
* [valgrind.server.31375.txt](https://github.com/PBSPro/pbspro/files/3526829/valgrind.server.31375.txt)

### DIFF
--- a/src/server/req_runjob.c
+++ b/src/server/req_runjob.c
@@ -1278,6 +1278,34 @@ parse_hook_rejectmsg(char *reject_msg, char *hook_name, int hook_name_size)
 
 /**
  * @brief
+ *		Check and put a hold on a job if it has already been run
+ *		too many times.
+ *
+ * @param[in,out]	pjob	-	job pointer
+ *
+ * @return	void
+ */
+static void
+check_failed_attempts(job *jobp)
+{
+	if (jobp->ji_wattr[(int)JOB_ATR_runcount].at_val.at_long > PBS_MAX_HOPCOUNT + PBS_MAX_HOPCOUNT) {
+		jobp->ji_wattr[(int)JOB_ATR_hold].at_val.at_long |= HOLD_s;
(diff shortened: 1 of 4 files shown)


src/server/req_runjob.c
```diff
@@ -1276,6 +1276,34 @@ parse_hook_rejectmsg(char *reject_msg, char *hook_name, int hook_name_size)
...
+ *		Check and put a hold on a job if it has already been run
+ *		too many times.
+ *
+ * @param[in,out]	pjob	-	job pointer
+ *
+ * @return	void
+ */
+static void
+check_failed_attempts(job *jobp)
+{
```
mike0042: Thank you for making this a function and reducing duplicate code. There is another instance of this in req_jobobit.c that also includes a NAS mod. Would be nice to consolidate that one as well.

bayucan: Good idea. I'll do.

PR #1261, opened 2019-08-21
Refactor job's run/end_event pointers to not dangle

bhroam:
#### Describe Bug or Feature
A recent change went in to have a job/reservation know its run/end event.  It was possible for these pointers to be left dangling if the event is deleted.

Also, the timed_event doubly linked list was broken.  The prev pointers wouldn't properly be updated.

#### Describe Your Change
I've changed the code so the run/end_event pointers are set in create_event() and NULL'd in free_timed_event().  This will ensure the pointers will never dangle, and are set properly.
I changed all timed_event functions to use create_event() and free_timed_event(), so we only have to set/NULL the run/end_event pointers in one place
I fixed add_event(), dup_timed_event_list(), and delete_event() to properly add and delete events using the doubly linked list.  The prev pointer code has been fixed to properly set them.

SInce this is a refactor project, no new PTL test was written.  The travis SmokeTest run should be fine to show testing.

### DIFF
--- a/src/scheduler/job_info.c
+++ b/src/scheduler/job_info.c
@@ -3284,10 +3284,8 @@ find_jobs_to_preempt(status *policy, resource_resv *hjob, server_info *sinfo, in
 
 		update_universe_on_end(npolicy, pjob,  "S", NO_ALLPART);
 		rjobs_count--;
-		if (pjob->end_event != NULL) {
-				if (delete_event(nsinfo, pjob->end_event, DE_NO_FLAGS) == 0)
-					schdlog(PBSEVENT_SCHED, PBS_EVENTCLASS_JOB, LOG_INFO, pjob->name, "Failed to delete end event for job.");
-		}
+		if (pjob->end_event != NULL)
+			delete_event(nsinfo, pjob->end_event);
 
 		pjobs[j] = pjob;
 		j++;
--- a/src/scheduler/simulate.c
+++ b/src/scheduler/simulate.c
@@ -1554,16 +1554,17 @@ delete_event(server_info *sinfo, timed_event *e)
 
 	if (calendar->next_event == e)
 		calendar->next_event = e->next;
+	
+	if (calendar->first_run_event == e)
+		calendar->first_run_event = find_timed_event(calendar->events, 0, NULL, TIMED_RUN_EVENT, 0);
 
-	if (e->prev == NULL) {
+	if (e->prev == NULL)
 		calendar->events = e->next;
-		if (e->next != NULL)
-			e->next->prev = NULL;
-	} else {
+	else
 		e->prev->next = e->next;
-		if (e->next != NULL)
-			e->next->prev = e->prev;
-	}
+
+	if (e->next != NULL)
+		e->next->prev = e->prev;
 
 	free_timed_event(e);
 }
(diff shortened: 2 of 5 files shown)


src/scheduler/job_info.c
```diff
@@ -3284,10 +3284,8 @@ find_jobs_to_preempt(status *policy, resource_resv *hjob, server_info *sinfo, in
 
 		update_universe_on_end(npolicy, pjob,  "S", NO_ALLPART);
 		rjobs_count--;
-		if (pjob->end_event != NULL) {
-				if (delete_event(nsinfo, pjob->end_event, DE_NO_FLAGS) == 0)
-					schdlog(PBSEVENT_SCHED, PBS_EVENTCLASS_JOB, LOG_INFO, pjob->name, "Failed to delete end event for job.");
-		}
+		if (pjob->end_event != NULL)
+			delete_event(nsinfo, pjob->end_event);
```
arungrover: We need to set pjob->end_event to NULL here, right?

src/scheduler/simulate.c
```diff
@@ -1547,54 +1538,34 @@ add_timed_event(timed_event *events, timed_event *te)
...
-		}
-
-		return 1;
+	if (e->prev == NULL) {
+		calendar->events = e->next;
+		if (e->next != NULL)
+			e->next->prev = NULL;
+	} else {
+		e->prev->next = e->next;
+		if (e->next != NULL)
```
arungrover: you can move this condition and assignment out of the if/else block. because effectively you are doing the same operation in both places, its just that e->prev is NULL in the 'if' block.

src/scheduler/simulate.c
```diff
@@ -1547,54 +1538,34 @@ add_timed_event(timed_event *events, timed_event *te)
...
 	event_list *calendar;
 
 	if (sinfo == NULL || e == NULL)
-		return 0;
+		return;
 
 	calendar = sinfo->calendar;
 
+	if (calendar->next_event == e)
+		calendar->next_event = e->next;
```
arungrover: What will happen to first_run_event if the event being deleted is the 'first_run_event'. I think we need to update that too.

src/scheduler/job_info.c
bhroam: delete_event() calls free_timed_event() which NULL's it.  One of the things I refactored is to move where the run/end_event pointers are set and NULL'd to one place.

arungrover: oh right, forgot about that. Thanks!

src/scheduler/simulate.c
```diff
@@ -1453,11 +1446,6 @@ add_event(event_list *calendar, timed_event *te)
 
 	calendar->events = add_timed_event(calendar->events, te);
 
-	if (te->event_type & TIMED_RUN_EVENT)
-		((resource_resv *)te->event_ptr)->run_event = te;
-	else if (te->event_type & TIMED_END_EVENT)
-		((resource_resv *)te->event_ptr)->end_event = te;
```
agrawalravi90: don't you need to modify add_timed_event() to set run_event and end_event pointers?

bhroam: We don't need to set the pointers when the event is added since they are set in create_event().  The idea behind this change was to set the pointers in one place and NULL them in one place that everybody calls.  All events are created via create_event() and freed via free_timed_event().

PR #1262, opened 2019-08-21
Fixed issue where nodepart's free_nodes wasn't being updated

bhroam:
#### Describe Bug or Feature
A recent optimization changed the way placement set metadata was kept up to date.  Instead of recreating the placement set metadata everytime a job is run, it only updates the nodes that were used.  The resources were modified, but the total count of free nodes was not.

#### Describe Your Change
When updating the metadata for nodes when a job is run, decrement the total number of free nodes if the node is no longer in the free state.

Before: [all.log](https://github.com/PBSPro/pbspro/files/3527718/all.log)
After: [all.after.log](https://github.com/PBSPro/pbspro/files/3527717/all.after.log)

### DIFF
--- a/test/tests/functional/pbs_allpart.py
+++ b/test/tests/functional/pbs_allpart.py
@@ -1,0 +1,145 @@
+# coding: utf-8
+
+# Copyright (C) 1994-2019 Altair Engineering, Inc.
+# For more information, contact Altair at www.altair.com.
+#
+# This file is part of the PBS Professional ("PBS Pro") software.
+#
+# Open Source License Information:
+#
+# PBS Pro is free software. You can redistribute it and/or modify it under the
+# terms of the GNU Affero General Public License as published by the Free
+# Software Foundation, either version 3 of the License, or (at your option) any
+# later version.
+#
+# PBS Pro is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.
+# See the GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+# Commercial License Information:
+#
+# For a copy of the commercial license terms and conditions,
+# go to: (http://www.pbspro.com/UserArea/agreement.html)
+# or contact the Altair Legal Department.
+#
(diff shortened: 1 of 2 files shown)


test/tests/functional/pbs_allpart.py
```diff
@@ -0,0 +1,145 @@
...
+# under a commercial license agreement.
+#
+# Use of Altair’s trademarks, including but not limited to "PBS™",
+# "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
+# trademark licensing policies.
+
+from tests.functional import *
+
+
+class TestAllPart(TestFunctional):
```
agrawalravi90: how about adding 'sched' in the name of the testsuite?

test/tests/functional/pbs_allpart.py
```diff
@@ -0,0 +1,145 @@
...
+        j2 = Job(TEST_USER, a)
+        jid2 = self.server.submit(j2)
+        self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'True'})
+
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid1)
+        a = {'job_state': 'Q', 'comment':
+             'Not Running: Not enough free nodes available'}
+        self.server.expect(JOB, a, id=jid2)
+
+    def test_vscatter2(self):
```
agrawalravi90: please add a docstring

test/tests/functional/pbs_allpart.py
```diff
@@ -0,0 +1,145 @@
...
+        j = Job(TEST_USER, a)
+        jid = self.server.submit(j)
+
+        a = {'job_state': 'Q', 'comment':
+             'Can Never Run: Not enough total nodes available'}
+        self.server.expect(JOB, a, id=jid)
+
+    def test_rassn(self):
+        """
+        Test rassn resource (ncpus) is empty and the comment is shown
```
agrawalravi90: what you mean by "rassn resource (ncpus) is empty" ?

test/tests/functional/pbs_allpart.py
agrawalravi90: Also, i feel like all of these tests are good candidates for being smoke tests, they are small and test important basic scheduler functionality. How do you feel about adding these to the SmokeTest suite instead?

bhroam: I thought about that, but I think these tests are a little specific.  They are testing a specific optimization in the scheduler.  They aren't testing general scheduler functionality

PR #1263, opened 2019-08-22
test_cgroup_periodic_update_check_values: add loop to check cput, mem…

vccardenas:
…, vmem usage expected values

#### Describe Bug or Feature
The test TestCgroupsHook.test_cgroup_periodic_update_check_values got an assertion error after checking for max memory usage.

#### Describe Your Change
The test checks for the mem and vmem usage too soon. On slower systems the wrong values are taken while mem and vmem usage continues to increment up to the expected amount.
Added a loop to keep checking for the expected cput, mem, and vmem usage up to a limited time.

#### Attach Test Logs/Output
[test_cgroup_periodic_update_check_values_x52_19.4_before.txt](https://github.com/PBSPro/pbspro/files/3527788/test_cgroup_periodic_update_check_values_x52_19.4_before.txt)
[test_cgroup_periodic_update_check_values_x58_19.4_before.txt](https://github.com/PBSPro/pbspro/files/3527789/test_cgroup_periodic_update_check_values_x58_19.4_before.txt)
[test_cgroup_periodic_update_check_values_x70_19.4_before.txt](https://github.com/PBSPro/pbspro/files/3527790/test_cgroup_periodic_update_check_values_x70_19.4_before.txt)
[test_cgroup_periodic_update_check_values_x52_19.4_after.txt](https://github.com/PBSPro/pbspro/files/3527791/test_cgroup_periodic_update_check_values_x52_19.4_after.txt)
[test_cgroup_periodic_update_check_values_x70_19.4_after.txt](https://github.com/PBSPro/pbspro/files/3527792/test_cgroup_periodic_update_check_values_x70_19.4_after.txt)
[test_cgroup_periodic_update_check_values_x58_19.4_after.txt](https://github.com/PBSPro/pbspro/files/3527793/test_cgroup_periodic_update_check_values_x58_19.4_after.txt)

after adding a check for values at end of each loop:
[test_cgroup_periodic_update_check_values_x70_19.4_after1.txt](https://github.com/PBSPro/pbspro/files/3532248/test_cgroup_periodic_update_check_values_x70_19.4_after1.txt)
[test_cgroup_periodic_update_check_values_x58_19.4_after1.txt](https://github.com/PBSPro/pbspro/files/3532249/test_cgroup_periodic_update_check_values_x58_19.4_after1.txt)
also ran the tests 15 times and now passing consistently.

### DIFF
--- a/test/tests/functional/pbs_cgroups_hook.py
+++ b/test/tests/functional/pbs_cgroups_hook.py
@@ -1326,9 +1326,11 @@ if %s e.job.in_ms_mom():
                         if vmem_usage > 400000:
                             break
             if cput_usage > 1.0 and mem_usage > 400000:
-                if self.swapctl == 'true' and vmem_usage > 400000:
+                if self.swapctl == 'true':
+                    if vmem_usage > 400000:
+                        break
+                else:
                     break
-                break
         self.assertGreater(cput_usage, 1.0)
         self.assertGreater(mem_usage, 400000)
         if self.swapctl == 'true':
--- a/test/tests/functional/pbs_cgroups_hook.py
+++ b/test/tests/functional/pbs_cgroups_hook.py
@@ -1325,6 +1325,10 @@ if %s e.job.in_ms_mom():
                         vmem_usage = int(match.groups()[0])
                         if vmem_usage > 400000:
                             break
+            if cput_usage > 1.0 and mem_usage > 400000:
+                if self.swapctl == 'true' and vmem_usage > 400000:
+                    break
+                break
         self.assertGreater(cput_usage, 1.0)
         self.assertGreater(mem_usage, 400000)
         if self.swapctl == 'true':
--- a/test/tests/functional/pbs_cgroups_hook.py
+++ b/test/tests/functional/pbs_cgroups_hook.py
@@ -1285,46 +1285,50 @@ if %s e.job.in_ms_mom():
         # Allow some time to pass for values to be updated
         begin = int(time.time())
         self.logger.info('Waiting for periodic hook to update usage data.')
-        time.sleep(15)
-        if self.paths['cpuacct']:
-            lines = self.moms_list[0].log_match(
-                '%s;update_job_usage: CPU usage:' %
-                jid, allmatch=True, starttime=begin)
-            usage = 0.0
-            for line in lines:
-                match = re.search(r'CPU usage: ([0-9.]+) secs', line[1])
-                if not match:
-                    continue
-                usage = float(match.groups()[0])
-                if usage > 1.0:
-                    break
-            self.assertGreater(usage, 1.0)
-        if self.paths['memory']:
-            lines = self.moms_list[0].log_match(
-                '%s;update_job_usage: Memory usage: mem=' % jid,
-                allmatch=True, max_attempts=5, starttime=begin)
-            usage = 0
-            for line in lines:
-                match = re.search(r'mem=(\d+)kb', line[1])
-                if not match:
-                    continue
-                usage = int(match.groups()[0])
-                if usage > 400000:
-                    break
-            self.assertGreater(usage, 400000, 'Max memory usage: %dkb' % usage)
-            if self.swapctl == 'true':
+        # loop to check if cput, mem, vmem are expected values
+        cput_usage = 0.0
+        mem_usage = 0
+        vmem_usage = 0
+        for count in range(3):
+            # Faster systems might have expected usage after 8 seconds
+            time.sleep(8)
+            if self.paths['cpuacct'] and cput_usage <= 1.0:
                 lines = self.moms_list[0].log_match(
-                    '%s;update_job_usage: Memory usage: vmem=' % jid,
-                    allmatch=True, max_attempts=5, starttime=begin)
-                usage = 0
+                    '%s;update_job_usage: CPU usage:' %
+                    jid, allmatch=True, starttime=begin)
                 for line in lines:
-                    match = re.search(r'vmem=(\d+)kb', line[1])
+                    match = re.search(r'CPU usage: ([0-9.]+) secs', line[1])
                     if not match:
                         continue
-                    usage = int(match.groups()[0])
-                    if usage > 400000:
+                    cput_usage = float(match.groups()[0])
+                    if cput_usage > 1.0:
                         break
-                self.assertGreater(usage, 400000)
+            if self.paths['memory'] and mem_usage <= 400000:
+                lines = self.moms_list[0].log_match(
+                    '%s;update_job_usage: Memory usage: mem=' % jid,
+                    allmatch=True, starttime=begin)
+                for line in lines:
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -1285,46 +1285,50 @@ def test_cgroup_periodic_update_check_values(self):
...
                 for line in lines:
-                    match = re.search(r'vmem=(\d+)kb', line[1])
+                    match = re.search(r'CPU usage: ([0-9.]+) secs', line[1])
                     if not match:
                         continue
-                    usage = int(match.groups()[0])
-                    if usage > 400000:
+                    cput_usage = float(match.groups()[0])
+                    if cput_usage > 1.0:
                         break
```
mike0042: Don't break here, we haven't checked mem usage yet. You could "continue" here if you want.

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -1285,46 +1285,50 @@ def test_cgroup_periodic_update_check_values(self):
...
+                        '%s;update_job_usage: Memory usage: vmem=' % jid,
+                        allmatch=True, starttime=begin)
+                    for line in lines:
+                        match = re.search(r'vmem=(\d+)kb', line[1])
+                        if not match:
+                            continue
+                        vmem_usage = int(match.groups()[0])
+                        if vmem_usage > 400000:
+                            break
+        self.assertGreater(cput_usage, 1.0)
```
mike0042: You should check cput, mem_usage, and vmem_usage at the end of the loop. If they are all above the expected values, then break from the loop. Otherwise, one may be correct but the others are not.

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -1285,46 +1285,50 @@ def test_cgroup_periodic_update_check_values(self):
...
+                lines = self.moms_list[0].log_match(
+                    '%s;update_job_usage: Memory usage: mem=' % jid,
+                    allmatch=True, starttime=begin)
+                for line in lines:
+                    match = re.search(r'mem=(\d+)kb', line[1])
+                    if not match:
+                        continue
+                    mem_usage = int(match.groups()[0])
+                    if mem_usage > 400000:
+                        break
```
mike0042: Don't break here, we haven't checked vmem_usage yet. You could "continue" here if you want.

test/tests/functional/pbs_cgroups_hook.py
vccardenas: Since we have our expected cput value we can break from the for loop and not have to read in the remaining lines. After the break it will then move on to checking mem_usage.

test/tests/functional/pbs_cgroups_hook.py
vccardenas: We might not get the expected values within the loop, a reasonable time frame within which we can expect the correct values.  So I check the cput, mem_usage, and vmem_usage at the end of the loop.

test/tests/functional/pbs_cgroups_hook.py
vccardenas: Since we have our expected mem_usage value we can break from the for loop and not have to read in the remaining lines. After the break it will then move on to checking vmem_usage.

test/tests/functional/pbs_cgroups_hook.py
mike0042: OK, I was thinking that was breaking from the enclosing loop. Thanks for making me look closer. In the event you find a match, why do you only break if the CPU time is greater than 1.0? Even if it isn't you should still break since you've found the line you are looking for. I'm suggesting that the conditional be removed.

test/tests/functional/pbs_cgroups_hook.py
mike0042: Same WRT removing the conditional.

test/tests/functional/pbs_cgroups_hook.py
mike0042: Same WRT removing the conditional for the break. Also, you should check all three at the end of the loop. If they are all above the expected values then you can break from the 8 second sleep loop. No need to iterate again if all the values are where they need to be.

test/tests/functional/pbs_cgroups_hook.py
vccardenas: If I remove the conditional then the loop will just see the earliest entry for cput usage in the mom logs (relative to 'begin') and never get to the latest entry due to the assert at the end of each loop per your suggestion.

test/tests/functional/pbs_cgroups_hook.py
vccardenas: I check that the cput, mem, vmem values match what is expected within each loop. If within a loop the expected value is reached then that parameter is not checked in the next loop. If all the parameters have reached expected values then the mom logs are not reread in the following loop. If at least one parameter has not reached the expected value after all the loops are exhausted then we get an error.

test/tests/functional/pbs_cgroups_hook.py
vccardenas: If I remove the conditional then the loop will just see the earliest entry for mem usage in the mom logs (relative to 'begin') and never get to the latest entry due to the assert at the end of each loop per your suggestion.

test/tests/functional/pbs_cgroups_hook.py
mike0042: Thanks for the explanation. Isn't there a way to search the log in reverse order?

test/tests/functional/pbs_cgroups_hook.py
mike0042: Thanks for the explanation. If all three values are above the threshold we're looking for upon the first iteration, what stops the loop from executing again and sleeping for another 8 seconds?

test/tests/functional/pbs_cgroups_hook.py
vccardenas: You might be thinking of tali=True which is the default.

test/tests/functional/pbs_cgroups_hook.py
vccardenas: Per your suggestion I have added a check for all three at the end of each loop and break if appropriate.  This will stop looping again and sleeping for 8 seconds.

mike0042: What happens when cput_usage > 1.0 and mem_usage > 400000 and self.swapctl == 'true'? The loop will break whether or not vmem_usage > 400000. Also, I find it very strange that self.swapctl is a string and not a boolean. I think is_memsw_enabled() needs updating.

vccardenas: I have corrected the value checks at the end of each loop.  Changing is_memsw_enabled() to return a boolean (e.g. True) will affect the settings in the config files which needs 'true', for example:
        "memsw":
        {
            "enabled"         :  true

PR #1265, opened 2019-08-22
Use make && make install method in Travis

hirenvadalia:
#### Describe Bug or Feature
Currently, we are using rpm build in Travis so make && make install is not getting tested properly and because of that, we found few issues in it when user/customer tried to make && make install and found bugs. Also newly added --with ptl in rpmbuild is not getting tested in Travis.

#### Describe Your Change
* Use make && make install method (along with pip install for PTL) in all platforms except sanitize mode
* Use rpmbuild method with `--with ptl` in sanitize mode so we can keep rpmbuild testing also but enhance it to use PTL rpm instead pip install which is getting tested in other platforms
* Also refactored Travis script such a way that user or other tools can use it like
   * User can do only installation of PBS/PTL deps by running `ONLY_INSTALL_DEPS=1 .travis/do.sh`
   * User can do compilation and installation of PBS/PTL but don't start PBS by running `DONT_START_PBS=1 .travis/do.sh`
   * User can do just rebuild (aka only make, make install and PBS restart) by running `ONLY_REBUILD=1 .travis/do.sh` (NOTE: This requires user must run `.travis/do.sh` first without any flags)
* Remove platform-dependent files from buildutils directly, Autoconf/autogen.sh with add required file in that directory based on the platform when the user runs autogen.sh

#### Link to Design Doc
None

Not required as this is Travis change


kjakkali:
@hirenvadalia : If other users or tools are using then i think you need to delete the docker previous docker container instance  before starting fresh build.

Also for sanitize build can you install debug binaries it will help in debugging.

hirenvadalia:
> @hirenvadalia : If other users or tools are using then i think you need to delete the docker previous docker container instance before starting fresh build.

I think you are talking about docker container in .travis.yml file, if yes then we don't need to delete any docker container as in Travis we will get fresh machine where no containers are running.

If you are taking about user or tools running .travis/do*.sh script inside the container then its user or tools' responsibility of starting fresh container every time (except for ONLY_REBUILD mode as ONLY_REBUILD mode is meant to run multiple time in the same container in which previously user/tool has run .travis/do*.sh without any flag)


> Also for sanitize build can you install debug binaries it will help in debugging.

Done, please have a look into latest commit

hirenvadalia:
[dont_start_pbs.txt](https://github.com/PBSPro/pbspro/files/3529075/dont_start_pbs.txt)
[only_install_deps.txt](https://github.com/PBSPro/pbspro/files/3529076/only_install_deps.txt)
[only_rebuild.txt](https://github.com/PBSPro/pbspro/files/3529077/only_rebuild.txt)

kjakkali:
> 
> 
> [dont_start_pbs.txt](https://github.com/PBSPro/pbspro/files/3529075/dont_start_pbs.txt)
> [only_install_deps.txt](https://github.com/PBSPro/pbspro/files/3529076/only_install_deps.txt)
> [only_rebuild.txt](https://github.com/PBSPro/pbspro/files/3529077/only_rebuild.txt)

Thank you @hirenvadalia

mike0042:
Please note that the removal of certain files in this PR now requires an Internet connection to build PBS Pro.

subhasisb:
> Please note that the removal of certain files in this PR now requires an Internet connection to build PBS Pro.

Oops, did not realize that. Which files are those?

mike0042:
The files in buildutils. IIRC, autotools will look for newer versions of the scripts. If it can't find any it falls back to the files in the repo, otherwise it downloads the newer versions. We assume an Internet connection is present, but that is not always the case in secure environments.

hirenvadalia:
@mike0042 I removed those files in buildutils because those files were platform-dependent and when you compile pbspro source on a different platform with same repo then git shows that file as changed, which is confusing to the user who is doing changes in the source.

However, I just tried compiling PBS without Internet connectivity (after installation of PBS deps) and it worked fine also I was able to run a job.
Here is a log of compiling and running a job without Internet connectivity: 
[compile_without_internet.txt](https://github.com/PBSPro/pbspro/files/3530549/compile_without_internet.txt)

Also, I tried to build on Altair's internal build tool and that tool does compile and test on the machines which don't have Internet connectivity and it worked fine there.

But still if you find any file which we need to restore in buildutils then please let me know, I will restore that file.

mike0042:
I grep'ed the log files for "buildutils" from before and after this change. Before the change I see:
```
AUTOGEN.LOG:libtoolize: putting auxiliary files in AC_CONFIG_AUX_DIR, `buildutils'.
AUTOGEN.LOG:libtoolize: copying file `buildutils/ltmain.sh'
AUTOGEN.LOG:configure.ac:51: installing 'buildutils/config.guess'
AUTOGEN.LOG:configure.ac:51: installing 'buildutils/config.sub'
```
After the change I see:
```
AUTOGEN.LOG:libtoolize: putting auxiliary files in AC_CONFIG_AUX_DIR, `buildutils'.
AUTOGEN.LOG:libtoolize: copying file `buildutils/ltmain.sh'
AUTOGEN.LOG:configure.ac:71: installing 'buildutils/ar-lib'
AUTOGEN.LOG:configure.ac:72: installing 'buildutils/compile'
AUTOGEN.LOG:configure.ac:51: installing 'buildutils/config.guess'
AUTOGEN.LOG:configure.ac:51: installing 'buildutils/config.sub'
AUTOGEN.LOG:configure.ac:56: installing 'buildutils/install-sh'
AUTOGEN.LOG:configure.ac:56: installing 'buildutils/missing'
AUTOGEN.LOG:src/cmds/Makefile.am: installing 'buildutils/depcomp'
AUTOGEN.LOG:src/cmds/scripts/Makefile.am:66: installing 'buildutils/py-compile'
```
As it turns out, I was mistaken. These files don't get downloaded, but are copied from /usr/share/libtool/config. My apologies for the confusion. I'm satisfied this change will work properly.

hirenvadalia:
Thanks @mike0042 for confirmation.

PR #1268, opened 2019-08-28
Fix log_match failure at day change and revert_mom not to restart mom unnecessary

hirenvadalia:
#### Describe Bug or Feature
* Currently PBSService.log_lines use a different method to get log file contents one of them is FileUtils which fails when day change and PBS creates a new log file
* Also FileUtils returns log lines in reverse order while other method returns correct order.
* Also FileUtils returns '\n' at end of every log lines in the returned list while other method doesn't.
* PBSTestSuite.revert_mom always restarts mom when it's not required then also, this is because after #1034 Server.reverts_to_defaults delete nodes and in revert_mom it always goes inside `if not nodes:` condition at line 1397 and inside the if-block it again reverts to default mom and restarts it.

#### Describe Your Change
* Remove buggy FileUtils, AFAIK FileUtils was implemented to make common utility for all local or remote file/logs traversing file content like File object in Python. But that functionality is not used as of now anywhere... (except one func `finished_jobs_nodes` but that func is not used anywhere)
  * IF we require utility like FileUtils in future then we can make use of StringIO and cat/head/tail to get File like object from the file content.
* Use cat, tail or head commands to get log lines in PBSService.log_lines, this way it will be in the correct order and all methods will have the same content without '\n' at end
* Fix PBSTestSuite.revert_mom to restart mom only when mom has vnodedefs else send HUP signal to reload changed config

#### Link to Design Doc
None

Not required as Travis will be sufficient


hirenvadalia:
@bhroam All suggestions are accepted and implemented, please have a look at the latest commit

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4036,7 +4036,7 @@ def log_lines(self, logtype, id=None, n=50, tail=True, starttime=None,
                     n = n - len(day_lines)
                     if n <= 0:
                         break
-        except:
+        except (BaseException, IOError, PtlLogMatchError):
```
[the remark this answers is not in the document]
hirenvadalia: @bhroam No, we don't want to catch SIGINT here. Changed to Exception...

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4036,7 +4036,7 @@ def log_lines(self, logtype, id=None, n=50, tail=True, starttime=None,
                     n = n - len(day_lines)
                     if n <= 0:
                         break
-        except (BaseException, IOError, PtlLogMatchError):
+        except (Exception, IOError, PtlLogMatchError):
```
bhroam: Exception seems kind of broad?  The only exceptions that should be thrown are from du.run_cmd(), du.cat(), and PtlLogMatchError.  What gets thrown from the two du functions?

hirenvadalia: As of now du.run_cmd and du.cat doesn't throw any particular exception so it could be anything, so we have to Exception only.

prakashcv13:
I restarted the failed travis build and it passed this time.

test/fw/ptl/lib/pbs_testlib.py
bhroam: I still think Exception is a little broad, but at least it is better than a blanket except like it was.  I'm fine with it.

PR #1272, opened 2019-08-30
minimum requirement for ptl tests

birbalsain:
#### Describe Bug or Feature
Minimum hardware requirements for ptl test

#### Describe Your Change
Added min_mom_ram<in mb>, min_mom_disk<in mb>, min_server_ram<mb>, min_server_disk<in mb> options in existing requirement decorator

#### Link to Design Doc
https://pbspro.atlassian.net/wiki/spaces/PD/pages/1221427203/Addition+of+hardware+requirements+option+to+existing+requirements+decorator

[test_logs_with_changes.txt](https://github.com/PBSPro/pbspro/files/3599182/test_logs_with_changes.txt)
[test_logs_without_changes.txt](https://github.com/PBSPro/pbspro/files/3599185/test_logs_without_changes.txt)
[HARDWARE_REQUIREMENT_LOGS.txt](https://github.com/PBSPro/pbspro/files/3610289/HARDWARE_REQUIREMENT_LOGS.txt)

### DIFF
--- a/test/fw/ptl/utils/pbs_testsuite.py
+++ b/test/fw/ptl/utils/pbs_testsuite.py
@@ -81,6 +81,7 @@ default_requirements = {
     'num_moms': 1,
     'num_comms': 1,
     'num_clients': 1,
+# unit of ram and disk is GB
     'min_mom_ram': 1,
     'min_mom_disk': 5,
     'min_server_ram': 1,
--- a/test/fw/ptl/utils/pbs_testsuite.py
+++ b/test/fw/ptl/utils/pbs_testsuite.py
@@ -81,7 +81,7 @@ default_requirements = {
     'num_moms': 1,
     'num_comms': 1,
     'num_clients': 1,
-    'min_mom_ram': 1
+    'min_mom_ram': 1,
     'min_mom_disk': 5,
     'min_server_ram': 1,
     'min_server_disk': 5,
--- a/test/fw/ptl/utils/pbs_testsuite.py
+++ b/test/fw/ptl/utils/pbs_testsuite.py
@@ -81,10 +81,10 @@ default_requirements = {
     'num_moms': 1,
     'num_comms': 1,
     'num_clients': 1,
-    'min_mom_ram' : 500,
-    'min_mom_disk' : 500,
-    'min_server_ram' : 500,
-    'min_server_disk' : 500,
+    'min_mom_ram': 1
+    'min_mom_disk': 5,
+    'min_server_ram': 1,
+    'min_server_disk': 5,
     'no_mom_on_server': False,
     'no_comm_on_server': False,
     'no_comm_on_mom': True
--- a/test/fw/ptl/utils/pbs_testsuite.py
+++ b/test/fw/ptl/utils/pbs_testsuite.py
@@ -81,6 +81,10 @@ default_requirements = {
     'num_moms': 1,
     'num_comms': 1,
     'num_clients': 1,
+    'min_mom_ram' : 500,
+    'min_mom_disk' : 500,
+    'min_server_ram' : 500,
+    'min_server_disk' : 500,
     'no_mom_on_server': False,
     'no_comm_on_server': False,
     'no_comm_on_mom': True
--- a/test/fw/ptl/utils/pbs_testsuite.py
+++ b/test/fw/ptl/utils/pbs_testsuite.py
@@ -76,12 +76,12 @@ TIMEOUT_KEY = '__testcase_timeout__'
 MINIMUM_TESTCASE_TIMEOUT = 600
 REQUIREMENTS_KEY = '__PTL_REQS_LIST__'
 
+# unit of min_ram and min_disk is GB
 default_requirements = {
     'num_servers': 1,
     'num_moms': 1,
     'num_comms': 1,
     'num_clients': 1,
-# unit of ram and disk is GB
     'min_mom_ram': 1,
     'min_mom_disk': 5,
     'min_server_ram': 1,
--- a/test/fw/ptl/utils/plugins/ptl_test_runner.py
+++ b/test/fw/ptl/utils/plugins/ptl_test_runner.py
@@ -452,7 +452,6 @@ class SystemInfo:
             self.system_disk = float(disk_size[3]) / (2**20)
 
 
-
 class PtlTextTestRunner(TextTestRunner):
 
     """
--- a/test/fw/ptl/utils/plugins/ptl_test_runner.py
+++ b/test/fw/ptl/utils/plugins/ptl_test_runner.py
@@ -426,18 +426,20 @@ class SystemInfo:
         mem_info = du.cat(hostname, "/proc/meminfo")
         if mem_info['rc'] != 0:
             self.logger.error('failed to get content of /proc/meminfo')
-        ram_info = dict((i.split()[0].rstrip(':'), int(i.split()[1]))
-                        for i in mem_info['out'])
-        self.system_ram = float(ram_info['MemAvailable']) / (2**20)
+        else:
+            ram_info = dict((i.split()[0].rstrip(':'), int(i.split()[1]))
+                            for i in mem_info['out'])
+            self.system_ram = float(ram_info['MemAvailable']) / (2**20)
         # getting disk size in gb
         pbs_conf = du.parse_pbs_config(hostname)
         pbs_home_info = du.run_cmd(hostname, cmd=['df', '-k',
                                    pbs_conf['PBS_HOME']])
-        if mem_info['rc'] != 0:
+        if pbs_home_info['rc'] != 0:
             self.logger.error('failed to get output of df -k command')
-        disk_info = pbs_home_info['out']
-        disk_size = disk_info[1].split()
-        self.system_disk = float(disk_size[3]) / (2**20)
+        else:
+            disk_info = pbs_home_info['out']
+            disk_size = disk_info[1].split()
+            self.system_disk = float(disk_size[3]) / (2**20)
 
 
 class PtlTextTestRunner(TextTestRunner):
@@ -695,15 +697,27 @@ class PTLTestRunner(Plugin):
                 return False
         for hostname in param_dic['moms']:
             si = SystemInfo(hostname)
-            if eff_tc_req['min_mom_ram'] >= si.system_ram:
+            Available_sys_ram = getattr(si, 'system_ram', None)
+            if Available_sys_ram is None:
                 return False
-            if eff_tc_req['min_mom_disk'] >= si.system_disk:
+            elif eff_tc_req['min_mom_ram'] >= Available_sys_ram:
+                return False
+            Available_sys_disk = getattr(si, 'system_disk', None)
+            if Available_sys_disk is None:
+                return False
+            elif eff_tc_req['min_mom_disk'] >= Available_sys_disk:
                 return False
         for hostname in param_dic['servers']:
             si = SystemInfo(hostname)
-            if eff_tc_req['min_server_ram'] >= si.system_ram:
+            Available_sys_ram = getattr(si, 'system_ram', None)
+            if Available_sys_ram is None:
                 return False
-            if eff_tc_req['min_server_disk'] >= si.system_disk:
+            elif eff_tc_req['min_server_ram'] >= Available_sys_ram:
+                return False
+            Available_sys_disk = getattr(si, 'system_disk', None)
+            if Available_sys_disk is None:
+                return False
+            elif eff_tc_req['min_server_disk'] >= Available_sys_disk:
                 return False
         if set(param_dic['moms']) & set(param_dic['servers']):
             if eff_tc_req['no_mom_on_server']:
--- a/test/fw/ptl/utils/plugins/ptl_test_runner.py
+++ b/test/fw/ptl/utils/plugins/ptl_test_runner.py
@@ -35,11 +35,9 @@
 # "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
 # trademark licensing policies.
 
-import os
-import subprocess
-import sys
 import datetime
 import logging
+import os
 import platform
 import pwd
 import re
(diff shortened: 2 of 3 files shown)


test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -411,6 +414,42 @@ def printSummary(self):
         self.logger.info('\n'.join(msg))
 
 
+class system_availability:
```
hirenvadalia: Rename it to SystemInfo

hirenvadalia: Also add docstring

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -411,6 +414,42 @@ def printSummary(self):
...
+    def __init__(self, hostname=None):
+        # getting RAM size
+        ssh = subprocess.Popen(['ssh', hostname, 'cat', '/proc/meminfo'],
```
hirenvadalia: Use DshUtils.cat here

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -411,6 +414,42 @@ def printSummary(self):
...
+                               stdout=subprocess.PIPE)
+        meminfo = dict((i.split()[0].rstrip(':'), int(i.split()[1]))
+                       for i in ssh.stdout)
+        self.system_ram = meminfo['MemAvailable'] / (2**10)
+
+        # for home and exec size
+        home_disk = []
+        pbsconf_home = Server().pbs_conf['PBS_HOME']
```
hirenvadalia: Use DshUtils.parse_pbs_config here so you can eliminate user of Server here...

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -411,6 +414,42 @@ def printSummary(self):
...
+        pbsconf_exec = Server().pbs_conf['PBS_EXEC']
+        # getting mounte point for pbs home
+        home_info = subprocess.Popen(['ssh', hostname, 'df -k', pbsconf_home],
```
hirenvadalia: Same as above, Use DshUtils here

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -411,6 +414,42 @@ def printSummary(self):
...
+                                     stdout=subprocess.PIPE)
+        home_disk.extend((i.split())
+                         for i in home_info.stdout)
+        home_size = home_disk[1][3]
+        home_mount = home_disk[1][5]
+        # getting mounte point for pbs exec
+        exec_disk = []
+        exec_info = subprocess.Popen(['ssh', hostname, 'df -k', pbsconf_exec],
```
hirenvadalia: Why do we need to check PBS_EXEC size? Do we write anything in PBS_EXEC? I think checking PBS_HOME size will be sufficient?

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -145,6 +145,10 @@ class SkipTest(Exception):
     'num_moms': 1,
     'num_comms': 1,
     'num_clients': 1,
+    'min_mom_ram' : 500,
+    'min_mom_disk' : 500,
+    'min_server_ram' : 500,
+    'min_server_disk' : 500,
```
hirenvadalia: Don't you think 500MB of the disk is too less? At least we should use 2Gb or 5Gb...

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -145,6 +145,10 @@ class SkipTest(Exception):
     'num_moms': 1,
     'num_comms': 1,
     'num_clients': 1,
+    'min_mom_ram' : 500,
+    'min_mom_disk' : 500,
+    'min_server_ram' : 500,
```
hirenvadalia: Don't you think 500Mb is less? pbs_benchpress will eat up ~300MB of RAM when you run performance tests... I think the default should be at least 1Gb

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -664,6 +704,22 @@ def __are_requirements_matching(param_dic=None, test=None):
         for pk in param_count:
             if param_count[pk] < eff_tc_req[pk]:
                 return False
+        for hostname in param_dic['moms']:
+            pi = system_availability(hostname)
+            sys_avail['min_mom_ram'] = pi.system_ram
+            if eff_tc_req['min_mom_ram'] > sys_avail['min_mom_ram']:
```
hirenvadalia: you can directly use pi.system_ram here no need of sys_avail... And I think condition should be `>=` instead `>`

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -411,6 +411,29 @@ def printSummary(self):
         self.logger.info('\n'.join(msg))
 
 
+class SystemInfo:
+    """
+        used to get system's ram size and disk size information.
+
+        :system_ram: ram of the test running machine
+        :system_disk: disk size of the test running machine
```
visheshh: Mention the units are in gb

visheshh:
@birbalsain How will the user know that the units are in GB?

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -411,6 +411,29 @@ def printSummary(self):
...
+    """
+    def __init__(self, hostname=None):
+        du = DshUtils()
+        # getting RAM size in gb
```
birbalsain: Here i mentioned the unit

test/fw/ptl/utils/plugins/ptl_test_runner.py
birbalsain: I mentioned unit in next part

test/fw/ptl/utils/plugins/ptl_test_runner.py
visheshh: It should be in the docstring, else it won't be documented.

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -411,6 +411,29 @@ def printSummary(self):
...
+        ssh = du.cat(hostname, "/proc/meminfo")
+        meminfo = dict((i.split()[0].rstrip(':'), int(i.split()[1]))
+                       for i in ssh['out'])
+        self.system_ram = meminfo['MemAvailable'] / (2**20)
+        # getting disk size in gb
```
birbalsain: Here i mentioned unit of disk size.

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -411,6 +411,29 @@ def printSummary(self):
...
+    """
+        used to get system's ram size and disk size information.
+
+        :system_ram: ram of the test running machine
+        :system_disk: disk size of the test running machine
+    """
+    def __init__(self, hostname=None):
+        du = DshUtils()
+        # getting RAM size in gb
+        ssh = du.cat(hostname, "/proc/meminfo")
```
hirenvadalia: check exit status here, also rename ssh to some meaning full variable name

test/tests/selftest/pbs_requirements_decorator.py
```diff
@@ -38,7 +38,8 @@
 from tests.selftest import *
 
 
-@requirements(num_servers=2, num_comms=2)
+@requirements(num_servers=1, num_comms=1, min_mom_ram=1,
+              min_mom_disk=5, min_server_ram=1, min_server_disk=5)
```
visheshh: Can you only have these at test suite level?

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -411,6 +411,29 @@ def printSummary(self):
...
+        meminfo = dict((i.split()[0].rstrip(':'), int(i.split()[1]))
+                       for i in ssh['out'])
+        self.system_ram = meminfo['MemAvailable'] / (2**20)
+        # getting disk size in gb
+        pbs_conf_file = du.parse_pbs_config(hostname)
```
hirenvadalia: I suggest renaming to pbs_conf as its dict of pbs.conf not a path to pbs.conf

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -411,6 +411,29 @@ def printSummary(self):
...
+        home_info = du.run_cmd(hostname, cmd=['df', '-k',
```
hirenvadalia: check exit status here

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -664,6 +687,18 @@ def __are_requirements_matching(param_dic=None, test=None):
         for pk in param_count:
             if param_count[pk] < eff_tc_req[pk]:
                 return False
+        for hostname in param_dic['moms']:
+            pi = SystemInfo(hostname)
```
hirenvadalia: suggest renaming to `si`

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -415,23 +415,29 @@ class SystemInfo:
...
         du = DshUtils()
         # getting RAM size in gb
-        ssh = du.cat(hostname, "/proc/meminfo")
-        meminfo = dict((i.split()[0].rstrip(':'), int(i.split()[1]))
-                       for i in ssh['out'])
-        self.system_ram = meminfo['MemAvailable'] / (2**20)
+        mem_info = du.cat(hostname, "/proc/meminfo")
+        if mem_info['rc'] != 0:
+            self.logger.error('failed to get content of /proc/meminfo')
+        ram_info = dict((i.split()[0].rstrip(':'), int(i.split()[1]))
```
hirenvadalia: this will fail if mem_info['rc'] is non-0

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -415,23 +415,29 @@ class SystemInfo:
...
-                               pbs_conf_file['PBS_HOME']])
-        disk_info = home_info['out']
-        home_disk = disk_info[1].split()
-        self.system_disk = float(home_disk[3]) / (2**20)
+        pbs_conf = du.parse_pbs_config(hostname)
+        pbs_home_info = du.run_cmd(hostname, cmd=['df', '-k',
+                                   pbs_conf['PBS_HOME']])
+        if mem_info['rc'] != 0:
+            self.logger.error('failed to get output of df -k command')
+        disk_info = pbs_home_info['out']
```
hirenvadalia: I think you have to use pbs_home_info instead mem_info here

hirenvadalia:
@birbalsain Please don't do rebase or force push while a review is going on unless someone from review asks you to do.

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -695,15 +697,27 @@ def __are_requirements_matching(param_dic=None, test=None):
                 return False
         for hostname in param_dic['moms']:
             si = SystemInfo(hostname)
-            if eff_tc_req['min_mom_ram'] >= si.system_ram:
+            Available_sys_ram = getattr(si, 'system_ram', None)
```
hirenvadalia: Available === available (aka use small case for variable name)

hirenvadalia:
@birbalsain latest changes look good but can you please attach test log with latest changes...

visheshh:
@PBSPro/pbspro-maintainers : Please review the PR

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -420,22 +420,32 @@ class SystemInfo:
     """
     logger = logging.getLogger(__name__)
 
-    def __init__(self, hostname=None):
+    def get_system_info(self, hostname=None):
+        """
+        used to get system's ram size and disk size information.
+
+        :system_ram: Available ram(in GB) of the test running machine
```
agrawalravi90: system_ram and system_disk are not the arguments to this function, document 'hostname' instead.
Also, the docstring for params is written as:
```
:param <param name>: <description>
:type <param name>: <type>
```

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -420,22 +420,32 @@ class SystemInfo:
...
-            self.logger.error('failed to get content of /proc/meminfo')
+            _msg = 'failed to get content of /proc/meminfo of host: '
+            self.logger.error(_msg + hostname)
         else:
-            ram_info = dict((i.split()[0].rstrip(':'), int(i.split()[1]))
-                            for i in mem_info['out'])
-            self.system_ram = float(ram_info['MemAvailable']) / (2**20)
+            ram_dict = {}
+            for i in mem_info['out']:
+                ram_info = i.split()
```
agrawalravi90: why not just split once at the colon? right now, it will split as many times as the spaces in the string, which might give you unexpected results. The following might be safer:
key, val = i.split(":", 1)
ram_dict[key.strip()] = val.strip()

agrawalravi90: Actually, why are you using ram_dict at all? All that you need is MemAvailable right? So you can just do this:

```
for i in mem_info['out']:
    if "MemAvailable" in i:
        self.system_ram = float(i.split(":", 1)[1].strip()) / (2**20)
        break
```

birbalsain: Here i is "MemAvailable: 1234 kb".i think split by space is nice.

birbalsain:
I have added latest logs.

PR #1274, opened 2019-09-03
Fixed sched crash due to freeing in wrong order

bhroam:
## Describe Bug or Feature
A recent change in the scheduler has the job/resv objects keep track of their run/end events.  When the event is freed, the run/end event on the job/resv is NULL'd.  These objects were freed out of order.  The jobs were freed dangling pointers in the events.  When the events were freed, they'd try and NULL pointers on the already freed job/resv objects and crash

#### Describe Your Change
Free the calendar first before the jobs/resvs.

[valgrind-before.log](https://github.com/PBSPro/pbspro/files/3571589/valgrind-before.log)
[valgrind-after.log](https://github.com/PBSPro/pbspro/files/3571588/valgrind-after.log)
grep for free_timed_event

### DIFF
--- a/src/scheduler/server_info.c
+++ b/src/scheduler/server_info.c
@@ -1172,8 +1172,6 @@ free_server_info(server_info *sinfo)
 	if (sinfo->nodes_by_NASrank != NULL)
 		free(sinfo->nodes_by_NASrank);
 #endif
-
-	free(sinfo);
 }
 
 /**
@@ -1700,6 +1698,8 @@ free_server(server_info *sinfo)
 	if (sinfo == NULL)
 		return;
 
+	free_server_info(sinfo);
+
 	free_queues(sinfo->queues);
 	free_nodes(sinfo->nodes);
 	free_resource_resv_array(sinfo->resvs);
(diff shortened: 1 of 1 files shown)


src/scheduler/server_info.c
```diff
@@ -1700,14 +1698,16 @@ free_server(server_info *sinfo)
 	if (sinfo == NULL)
 		return;
 
+	free_server_info(sinfo);
```
arungrover: Can you please write a comment explaining this move. It will help other developers to pay attention to order of operation.

src/scheduler/server_info.c
```diff
@@ -1172,8 +1172,6 @@ free_server_info(server_info *sinfo)
 	if (sinfo->nodes_by_NASrank != NULL)
 		free(sinfo->nodes_by_NASrank);
 #endif
-
-	free(sinfo);
```
agrawalravi90: I'm not sure if I like this, this will make it inconsistent with the other destructors work right? as an interface, one would expect free_server_info() to free the object itself as well. How about this instead: you add an additional argument to this function which says "also free the other reference objects like queues, nodes, resvs etc." so that you could do that inside this function itself?

bhroam: Well It depends what you call the "destructor" function.  Is it free_server() or free_server_info()?  The call to free sinfo was originally inside free_server_info(), but with it being moved to the top of free_server(), we can't.  We reference sinfo when freeing the queues, nodes, jobs, and reservations.  It has to be the last thing freed.  This is why I moved it.

Issue #1267, opened 2019-08-26
SmokeTest.test_finished_jobs fails due to permission issue


vstumpf:
Taken off master, 52dab14b644f684c6d13cce60289a34532265602.
Installed pbspro_server.rpm
Installed ptl from source, using `sudo pip install -U -r requirements.txt .`

As a non-root user, called `pbs_benchpress -t SmokeTest.test_finished_jobs`

Test fails.
I've attached the test log, but here's a snippet:
```
2019-08-26 12:04:11,211 INFOCLI2 shecil: which chmod
2019-08-26 12:04:11,215 INFOCLI2 shecil: /usr/bin/chmod 0755 /usr/lib/python2.7/site-packages/ptl/utils/jobs/eatcpu.py
2019-08-26 12:04:11,219 ERROR    err: ['/usr/bin/chmod: changing permissions of \xe2\x80\x98/usr/lib/python2.7/site-packages/ptl/utils/jobs/eatcpu.py\xe2\x80\x99: Operation not permitted']
2019-08-26 12:04:11,219 INFO     job: executable set to /usr/lib/python2.7/site-packages/ptl/utils/jobs/eatcpu.py with arguments: 15
2019-08-26 12:04:11,220 INFOCLI  shecil: sudo -H -u pbsuser /opt/pbs/bin/qsub -l ncpus=2 -- /usr/lib/python2.7/site-packages/ptl/utils/jobs/eatcpu.py 15
```

I'm not sure we should be changing the permissions of things in /usr/lib to executable. Maybe the job should be `python <script.py>`?

[smoketest.txt](https://github.com/PBSPro/pbspro/files/3542553/smoketest.txt)

PR #1277, opened 2019-09-04
PTL Changes for Shasta

vstumpf:
Still some things need to be changed in PTL for shasta.

#### Describe Bug or Feature
is_localhost fails on shasta
PBS_PUBLIC_HOST_NAME needs to be set
run_copy needs to handle user-hosts
create_temp_file needs to handle user-hosts
Job.create_script needs to handle user-hosts
proc_utils does not support the shasta platform
\@skipOnShasta needs to be implemented
Some usernames are hardcoded in PTL

#### Describe Your Change
is_localhost should check the hostnames reported, and if they're the server hostname, return `True`
Set PBS_PUBLIC_HOST_NAME on the server conf.
Add user-host specific code to run_copy, create_temp_file, and Job.create_script
Allow the shasta platform to use proc_utils
Implement \@skipOnShasta
Change hard-coded usernames to str(TEST_USER)

#### Link to Design Doc
https://pbspro.atlassian.net/wiki/spaces/PD/pages/1327235075/Support+PTL+on+Shasta

soon

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -13901,7 +13901,6 @@ class Job(ResourceResv):
                          be created
         :type hostname: str or None
         """
-        shasta_set_host = False
 
         if body is None:
             return None
@@ -13938,7 +13937,6 @@ class Job(ResourceResv):
             user = PbsUser.get_user(self.username)
             if user.host:
                 hostname = user.host
-                shasta_set_host = True
                 asuser = user.name
 
         self.script_body = body
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -8010,7 +8010,7 @@ class Server(PBSService):
         operation.
 
         :param obj_type: The type of object to query, JOB, SERVER,
-                         SCHEDULER, QUEUE NODE
+                         SCHEDULER, QUEUE, NODE
         :type obj_type: str
         :param attrib: Attributes to query, can be a string, a list,
                        or a dict
@@ -13932,6 +13932,8 @@ class Job(ResourceResv):
                     body[i] = " ".join(line_arr)
             body = '\n'.join(body)
 
+        # If the user has a userhost, the job will run from there
+        # so the script should be made there
         if self.platform == 'shasta' and self.username:
             user = PbsUser.get_user(self.username)
             if user.host:
@@ -13947,10 +13949,6 @@ class Job(ResourceResv):
         fn = self.du.create_temp_file(hostname, prefix='PtlPbsJobScript',
                                       asuser=asuser, body=body)
         self.du.chmod(hostname, fn, mode=0755)
-        # on shasta, if we have already set the hostname,
-        # it does not need to be copied again
-        if not self.du.is_localhost(hostname) and not shasta_set_host:
-            self.du.run_copy(hostname, fn, fn)
         self.script = fn
         return fn
 
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -13933,7 +13933,7 @@ class Job(ResourceResv):
 
         # If the user has a userhost, the job will run from there
         # so the script should be made there
-        if self.platform == 'shasta' and self.username:
+        if self.username:
             user = PbsUser.get_user(self.username)
             if user.host:
                 hostname = user.host
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -13937,6 +13937,7 @@ class Job(ResourceResv):
             if user.host:
                 hostname = user.host
                 shasta_set_host = True
+                asuser = user.name
 
         self.script_body = body
         if self.du is None:
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -13901,6 +13901,8 @@ class Job(ResourceResv):
                          be created
         :type hostname: str or None
         """
+        shasta_set_host = False
+
         if body is None:
             return None
 
@@ -13930,6 +13932,12 @@ class Job(ResourceResv):
                     body[i] = " ".join(line_arr)
             body = '\n'.join(body)
 
+        if self.platform == 'shasta' and self.username:
+            user = PbsUser.get_user(self.username)
+            if user.host:
+                hostname = user.host
+                shasta_set_host = True
+
         self.script_body = body
         if self.du is None:
             self.du = DshUtils()
@@ -13937,9 +13945,10 @@ class Job(ResourceResv):
         # its mode once the current user has written to it
         fn = self.du.create_temp_file(hostname, prefix='PtlPbsJobScript',
                                       asuser=asuser, body=body)
-
         self.du.chmod(hostname, fn, mode=0755)
-        if not self.du.is_localhost(hostname):
+        # on shasta, if we have already set the hostname,
+        # it does not need to be copied again
+        if not self.du.is_localhost(hostname) and not shasta_set_host:
             self.du.run_copy(hostname, fn, fn)
         self.script = fn
         return fn
--- a/test/fw/ptl/utils/pbs_dshutils.py
+++ b/test/fw/ptl/utils/pbs_dshutils.py
@@ -979,7 +979,8 @@ class DshUtils(object):
                     # TODO: get a valid remote temporary file rather than
                     # assume that the remote host has a similar file
                     # system layout
-                    self.run_copy(hostname, _script, _script, level=level)
+                    self.run_copy(hostname, _script, _script,
+                                  runas=runas, level=level)
                     os.remove(_script)
                 runcmd = rshcmd + sudocmd + [_script]
             else:
@@ -1019,7 +1020,7 @@ class DshUtils(object):
             if as_script:
                 # must pass as_script=False otherwise it will loop infinite
                 self.rm(hostname, path=_script, as_script=False,
-                        level=level)
+                        level=level, runas=runas)
 
             # handle the case where stdout is not a PIPE
             if o is not None:
@@ -1111,6 +1112,9 @@ class DshUtils(object):
         if sudo is True and not self.sudo_cmd:
             sudo = False
 
+        if runas:
+            _runas_user = PbsUser.get_user(runas)
+
         for targethost in hosts:
             islocal = self.is_localhost(targethost)
             if sudo and not islocal:
@@ -1148,6 +1152,9 @@ class DshUtils(object):
                 cmd += copy_cmd
                 if recursive:
                     cmd += ['-r']
+                if (self.get_platform() == 'shasta' and _runas_user and
+                   _runas_user.port):
+                    cmd += ['-P', _runas_user.port]
                 cmd += [src]
                 if islocal:
                     cmd += [dest]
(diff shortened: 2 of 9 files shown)


test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -1126,6 +1131,13 @@ def _revert_pbsconf_server(self, vals_to_set):
                 new_pbsconf["PBS_LOG_HIGHRES_TIMESTAMP"] = "1"
                 restart_pbs = True
 
+            # if shasta, set PBS_PUBLIC_HOST_NAME
+            if server.platform == 'shasta':
+                localhost = socket.gethostname()
+                if new_pbsconf["PBS_PUBLIC_HOST_NAME"] != localhost:
+                    new_pbsconf["PBS_PUBLIC_HOST_NAME"] = localhost
+                    restart_pbs = True
```
bhroam: Isn't PBS_PUBLIC_HOST_NAME there for the commands?  If it is, we won't need to restart PBS.

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -1202,7 +1214,12 @@ def revert_pbsconf(self):
             "PBS_LOG_HIGHRES_TIMESTAMP": None
         }
 
-        self._revert_pbsconf_server(vals_to_set)
+        server_vals_to_set = vals_to_set
+
+        if primary_server.platform == 'shasta':
+            server_vals_to_set["PBS_PUBLIC_HOST_NAME"] = None
```
bhroam: Why are you setting this to None?  Don't you need it set?

test/fw/ptl/utils/pbs_dshutils.py
```diff
@@ -1933,13 +1933,20 @@ def create_temp_file(self, hostname=None, suffix='', prefix='PtlPbs',
             else:
                 os.write(fd, body)
         os.close(fd)
+
+        if self.get_platform() == 'shasta' and not hostname and asuser:
+            asuser = PbsUser.get_user(asuser)
+            if asuser.host:
+                self.logger.info('set hostname to userhost')
```
bhroam: I don't find this log message useful by itself.  I'd either delete it, or add the host.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -13930,16 +13932,23 @@ def create_script(self, body=None, asuser=None, hostname=None):
...
             self.du = DshUtils()
         # First create the temporary file as current user and only change
         # its mode once the current user has written to it
         fn = self.du.create_temp_file(hostname, prefix='PtlPbsJobScript',
                                       asuser=asuser, body=body)
-
         self.du.chmod(hostname, fn, mode=0755)
-        if not self.du.is_localhost(hostname):
+        # on shasta, if we have already set the hostname,
+        # it does not need to be copied again
```
bhroam: Could you add a why in here?  Have you already copied it?  Maybe say that and where in the code that copied it.

test/tests/pbs_smoketest.py
```diff
@@ -327,6 +327,7 @@ def test_limits(self):
         self.server.expect(JOB, a, attrop=PTL_AND)
 
     @skipOnCpuSet
+    @skip('This test is broken for oss')
```
bhroam: Don't skip this test.  If you do, it will likely not be unskipped when it is fixed.

test/tests/functional/pbs_equiv_classes.py
```diff
@@ -1495,29 +1495,30 @@ def test_server_queue_limit(self):
         # Set queue limit
         a = {
             'max_run': '[o:PBS_ALL=100],[g:PBS_GENERIC=20],\
-                       [u:PBS_GENERIC=20],[g:tstgrp01 = 8],[u:pbsuser1=10]'}
+                       [u:PBS_GENERIC=20],[g:tstgrp01 = 8],[u:%s=10]' %
```
bhroam: Since you are fixing usernames, you should fix the group names while you are at it.

test/fw/ptl/utils/pbs_dshutils.py
```diff
@@ -1019,8 +1019,12 @@ def run_cmd(self, hosts=None, cmd=None, sudo=False, stdin=None,
...
             if as_script:
                 # must pass as_script=False otherwise it will loop infinite
-                self.rm(hostname, path=_script, as_script=False,
-                        level=level, runas=runas)
+                if platform == 'shasta' and not runas:
+                    self.rm(hostname, path=_script, as_script=False,
+                            level=level, runas=runas)
+                else:
+                    self.rm(hostname, path=_script, as_script=False,
+                            level=level)
```
bhroam: OK, this looks wrong.  First you took runas=runas out of the normal case.  Second, you said 'not runas' and then added runas=runas.  Won't it be empty?

test/fw/ptl/utils/pbs_dshutils.py
```diff
@@ -1919,13 +1944,20 @@ def create_temp_file(self, hostname=None, suffix='', prefix='PtlPbs',
             else:
                 os.write(fd, body)
         os.close(fd)
+
+        if self.get_platform() == 'shasta' and not hostname and asuser:
+            asuser = PbsUser.get_user(asuser)
+            if asuser.host:
```
agrawalravi90: same comment, i don't think this needs to be shasta specific.

test/fw/ptl/utils/pbs_dshutils.py
```diff
@@ -1919,13 +1944,20 @@ def create_temp_file(self, hostname=None, suffix='', prefix='PtlPbs',
...
+                hostname = asuser.host
+
         # if temp file to be created on remote host
         if not self.is_localhost(hostname):
             if asuser is not None:
                 # by default mkstemp creates file with 0600 permission
                 # to create file as different user first change the file
                 # permission to 0644 so that other user has read permission
-                self.chmod(hostname, tmpfile, mode=0644)
+                self.chmod(tmpfile, mode=0644)
```
agrawalravi90: why did you remove the hostname arg?

test/tests/pbs_smoketest.py
agrawalravi90: this is a smoketest, how is it broken?

test/fw/ptl/utils/pbs_dshutils.py
vstumpf: Interesting, this is the commit i force-pushed but the changes still didn't get added. I must not have committed them.
You're right, it should be `if platform == 'shasta' and runas:`

test/fw/ptl/utils/pbs_dshutils.py
```diff
@@ -1146,11 +1156,17 @@ def run_copy(self, hosts=None, src=None, dest=None, sudo=False, uid=None,
                 cmd += copy_cmd
                 if recursive:
                     cmd += ['-r']
+                if (self.get_platform() == 'shasta' and _runas_user and
+                   _runas_user.port):
+                    cmd += ['-P', _runas_user.port]
```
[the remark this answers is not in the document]
vstumpf: run_cmd()'s port argument is the port option for ssh. This port is for scp.

test/fw/ptl/utils/pbs_dshutils.py
vstumpf: It was a bug, PTL creates a file on the local host, then chmods the file on the remote host? The file hasn't been copied yet.

test/fw/ptl/utils/pbs_testsuite.py
vstumpf: vals_to_set is weird, we use the keys of the dictionary to determine if the conf var needs to be set. It could be replaced with a list, but that's out of scope.

test/tests/pbs_smoketest.py
vstumpf: Sorry, this wasn't supposed to be in the PR, just for testing.
It's broken due to #1267

test/tests/functional/pbs_equiv_classes.py
lisa-altair: I think it would be reasonable to file a bug to fix group names.

bhroam: Since it won't take long to fix the group names, I'd rather do it now.

test/fw/ptl/lib/pbs_testlib.py
vstumpf: I'm not even sure why this is in the code, create_temp_file should create the job script on the host specified by hostname. We then chmod the file on that host. Then we do another copy? It seems wrong, I'll remove it.

test/fw/ptl/utils/pbs_dshutils.py
vstumpf: Sorry, it was a debug message.

test/fw/ptl/utils/pbs_testsuite.py
vstumpf: In order to set the pbsconfig, restart_pbs needs to be true.

vstumpf:
Please don't merge until I test tomorrow. It is still open for review, however.

test/fw/ptl/utils/pbs_dshutils.py
agrawalravi90: ok, but now you are passing the file path 'tmpfile' into the 'hostname' argument of the function, you should change this to path=tmpfile

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -13932,11 +13931,12 @@ def create_script(self, body=None, asuser=None, hostname=None):
                     body[i] = " ".join(line_arr)
             body = '\n'.join(body)
 
+        # If the user has a userhost, the job will run from there
+        # so the script should be made there
```
agrawalravi90: This sounds like a pretty generic logic, so can we remove "if self.platform == 'shasta'" from the check below?

test/fw/ptl/utils/pbs_dshutils.py
```diff
@@ -1019,7 +1019,7 @@ def run_cmd(self, hosts=None, cmd=None, sudo=False, stdin=None,
 
             if as_script:
                 # must pass as_script=False otherwise it will loop infinite
-                if platform == 'shasta' and not runas:
+                if platform == 'shasta' and runas:
```
agrawalravi90: I think we can just pass 'runas' to rm() in either case, did you see any issue doing that in the generic case?

test/fw/ptl/utils/pbs_dshutils.py
```diff
@@ -1156,9 +1154,8 @@ def run_copy(self, hosts=None, src=None, dest=None, sudo=False, uid=None,
                 cmd += copy_cmd
                 if recursive:
                     cmd += ['-r']
-                if (self.get_platform() == 'shasta' and _runas_user and
-                   _runas_user.port):
-                    cmd += ['-P', _runas_user.port]
+                if self.get_platform() == 'shasta' and runas and runas.port:
```
agrawalravi90: Again, I feel that the shasta check is not needed here, if runas and runas.port is set, I think we should just use it.

test/fw/ptl/utils/pbs_dshutils.py
```diff
@@ -1300,7 +1297,8 @@ def is_localhost(self, host=None):
             self._h2l[host] = True
             return True
         # on a shasta machine, the name returned by `hostname` (pbs-host) is
-        # different than the one we tell PTL to use (pbs-service-nmn).
+        # different than the one we tell PTL to use (pbs-service-nmn). This
```
agrawalravi90: Just for my understanding, can you please explain why we ask PTL to use a different "local" hostname than the one that `hostname` returns ? Or if this is documented, point me to it? Thanks.

test/fw/ptl/utils/pbs_dshutils.py
agrawalravi90: > ok, but now you are passing the file path 'tmpfile' into the 'hostname' argument of the function, you should change this to path=tmpfile

I think you missed this comment

vstumpf: Ah, I missed that. Thanks!

PR #1279, opened 2019-09-05
Add Sleep for 0.2 seconds in client commands

Jitendra-Rajak:
#### Describe Bug or Feature
To help relieve pressure on the server from abusive query client command users, we should have certain client commands sleep for .2 seconds.
The idea is to have a fast server and "slow" client commands, which will help the server not be overrun by senseless client command requests. When people run "while $(true) ; qstat -f ; done" for hours on end they would get at most 5 replies in a second (rather than ~90, with 500 jobs in queue), which is more than they really need anyway, and the server can do other useful things (like dispatch and reap jobs) without being as easily overburdened.

#### Describe Your Change
Added sleep for 0.2 seconds in the certain client commands. for qstat, pbs_rstat, and pbsnodes (it would be OK to NOT do the sleep in pbsnodes for -r, -o, and -C, where the admin is making changes via the command).

#### Testing And Output
Tested With PTL:
[ptl_before_fix.txt](https://github.com/PBSPro/pbspro/files/3590687/ptl_before_fix.txt)
[Ptl_after_fix.txt](https://github.com/PBSPro/pbspro/files/3590690/Ptl_after_fix.txt)

Tested Manually:
[jitenr@jitendra ~]$ cat test.sh 
#!/bin/bash
for i in {1..90}
 do
    pbsnodes -av
 done
date

[Beforechange.txt.txt](https://github.com/PBSPro/pbspro/files/3589361/Beforechange.txt.txt)

[jitenr@jitendra ~]$ cat test.sh 
#!/bin/bash
for i in {1..5}
 do
    pbsnodes -av
 done
date

[afterchange.txt.txt](https://github.com/PBSPro/pbspro/files/3589325/afterchange.txt.txt)

Before the changes in 1 second approx 90  responses were getting received after the changes 5 responses are getting received in 1 second.

### DIFF
--- a/src/cmds/qstat.c
+++ b/src/cmds/qstat.c
@@ -2361,7 +2361,7 @@ main(int argc, char **argv, char **envp) /* qstat */
 	/*test for real deal or just version and exit*/
 
 	PRINT_VERSION_AND_EXIT(argc, argv);
-	check_last_query();
+	delay_query();
 #ifdef WIN32
 	if (winsock_init()) {
 		return 1;
@@ -3174,7 +3174,6 @@ svr_no_args:
 	 */
 	if (any_failed == PBSE_JOBHISTNOTSET)
 		any_failed = 0;
-	create_query_file();
 	exit(any_failed);
 }
 
--- a/src/cmds/qstat.c
+++ b/src/cmds/qstat.c
@@ -61,6 +61,11 @@
 #include "pbs_json.h"
 #include "pbs_internal.h"
 #include "libutil.h"
+#ifdef WIN32
+#include <windows.h>
+#else
+#include <unistd.h>
+#endif
 
 #if	TCL_QSTAT
 #include	<sys/stat.h>
@@ -3174,7 +3179,11 @@ svr_no_args:
 	 */
 	if (any_failed == PBSE_JOBHISTNOTSET)
 		any_failed = 0;
-
+#ifdef WIN32
+	Sleep(200);
+#else
+	usleep(200000);
+#endif
 	exit(any_failed);
 }
 
--- a/src/cmds/qstat.c
+++ b/src/cmds/qstat.c
@@ -61,11 +61,6 @@
 #include "pbs_json.h"
 #include "pbs_internal.h"
 #include "libutil.h"
-#ifdef WIN32
-#include <windows.h>
-#else
-#include <unistd.h>
-#endif
 
 #if	TCL_QSTAT
 #include	<sys/stat.h>
@@ -2366,7 +2361,7 @@ main(int argc, char **argv, char **envp) /* qstat */
 	/*test for real deal or just version and exit*/
 
 	PRINT_VERSION_AND_EXIT(argc, argv);
-
+	check_last_query();
 #ifdef WIN32
 	if (winsock_init()) {
 		return 1;
@@ -3179,11 +3174,7 @@ svr_no_args:
 	 */
 	if (any_failed == PBSE_JOBHISTNOTSET)
 		any_failed = 0;
-#ifdef WIN32
-	Sleep(200);
-#else
-	usleep(200000);
-#endif
+	create_query_file();
 	exit(any_failed);
 }
 
--- a/src/lib/Libutil/misc_utils.c
+++ b/src/lib/Libutil/misc_utils.c
@@ -1782,3 +1782,59 @@ perf_stat_stop(char *instance)
 
 	return (stat_summary);
 }
+
+/**
+ * @brief
+ *	creates an empty file in /tmp/ and saves timestamp of that file
+ *
+ * @param[in] - void
+ *
+ * @return - void
+ */
+void
+create_query_file(void)
+{
+	char filename[30];
+	char command[30];
+	snprintf(filename, 30, "/tmp/last_query_%d", getuid());
+#ifdef WIN32
+	snprintf(command, 30, "type nul > %d", filename);
+#else
+	snprintf(command, 30, "touch %s", filename);
+#endif
+	system(command);
+}
+
+/**
+ * @brief
+ *	stats te information of the empty file created in /tmp/ to decide 
+ *  whether to add sleep for .2 secons or not
+ * 
+ * @param[in] - void
+ *
+ * @return - void
+ */
+void
+check_last_query(void)
+{
+	char filename[30];	
+#ifdef WIN32
+	struct _stat buf;
+#else
+	struct stat buf;
+#endif
+	snprintf(filename, 30, "/tmp/last_query_%d", getuid());
+#ifdef WIN32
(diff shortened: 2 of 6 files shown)


bhroam:
I'm not sure I agree with this PR.  We shouldn't be in the business of slowing down our code.  If you are worried about the server getting overloaded, we should work on multi-server rather than slowing down our code.  If sites are getting people doing while loops around qstat, they have the option to slow it down themselves with a qstat wrapper.  I don't think we should slow down the commands for everyone just for those sites.

subhasisb:
> I'm not sure I agree with this PR. We shouldn't be in the business of slowing down our code. If you are worried about the server getting overloaded, we should work on multi-server rather than slowing down our code. If sites are getting people doing while loops around qstat, they have the option to slow it down themselves with a qstat wrapper. I don't think we should slow down the commands for everyone just for those sites.

Hi @bhroam, slowing the qstat by a very small (unnoticeable) margin is in an effort to minimize the impact of a "qstat in a loop" that ultimately hurts pbs_server performance (and thus the entire clusters). Even with multi-server, qstat in a loop would hurt a single server instance at the very least and will make requests from others to that instance slower. We can, of course, do more sophisticated Denial of Service checks, like check if a particular user is abusing the system and make that specific users responses slower, but this seemed to be a simple way to thwart such a looping qstat in a very simple, but effective manner.

subhasisb:
We probably don't need to put it for pbs_rstat (and even pbsnodes -av). We can target this only towards qstat, what say?

src/lib/Libutil/misc_utils.c
```diff
@@ -1782,3 +1782,59 @@ perf_stat_stop(char *instance)
...
+ * @param[in] - void
+ *
+ * @return - void
+ */
+void
+create_query_file(void)
+{
+	char filename[30];
+	char command[30];
+	snprintf(filename, 30, "/tmp/last_query_%d", getuid());
```
mike0042: Suggest you use $TMPDIR if it is defined in the environment. Also suggest you use pbs in the filename and make it a hidden file. Something like $TMPDIR/.pbs_query_%d

src/lib/Libutil/misc_utils.c
```diff
@@ -1782,3 +1782,59 @@ perf_stat_stop(char *instance)
...
+#ifdef WIN32
+	snprintf(command, 30, "type nul > %d", filename);
+#else
+	snprintf(command, 30, "touch %s", filename);
+#endif
+	system(command);
```
mike0042: It's cheaper to call open(2)/fopen(3) than it is to call system(3). If you want to guarantee you have fine granularity in the value you should write the timestamp into the file as a floating point value rather than relying on the underlying filesystem.

src/lib/Libutil/misc_utils.c
```diff
@@ -1782,3 +1782,59 @@ perf_stat_stop(char *instance)
...
+ * @brief
+ *	stats te information of the empty file created in /tmp/ to decide 
+ *  whether to add sleep for .2 secons or not
+ * 
+ * @param[in] - void
+ *
+ * @return - void
+ */
+void
+check_last_query(void)
```
mike0042: delay_query() may be a more descriptive name and would allow us to delay different amounts of time based on different criteria (e.g. machine load) in the future.

src/cmds/qstat.c
```diff
@@ -3179,11 +3174,7 @@ qstat -B [-f] [-F format] [-D delim] [ server_name... ]\n";
 	 */
 	if (any_failed == PBSE_JOBHISTNOTSET)
 		any_failed = 0;
-#ifdef WIN32
-	Sleep(200);
-#else
-	usleep(200000);
-#endif
+	create_query_file();
```
mike0042: I would call this from check_last_query() if the file isn't present and make create_query_file() a static or eliminate it altogether. It doesn't need to be exposed to the clients themselves.

src/lib/Libutil/misc_utils.c
```diff
@@ -1782,3 +1782,59 @@ perf_stat_stop(char *instance)
...
+#endif
+	snprintf(filename, 30, "/tmp/last_query_%d", getuid());
+#ifdef WIN32
+	if(_stat(filename, &buf) == 0) {
+		if(((time(NULL)*1000) - (buf.st_mtime * 1000)) < 10) {
+			Sleep(200);
+		}
+	}
+#else
+	if(stat(filename, &buf) == 0) {
```
subhasisb: Please ensure we are getting milliseconds level granularity on both Linux and Windows, from stat()

src/cmds/qstat.c
subhasisb: @mike0042 i think we need to do the "touch" (or update timstamp) activity at the very exit of the program, else we will need to account for the varying time that the program might spend from the start (where check_last_query()) was called. In case of some qstat -f's it could be in the range of tens of seconds, making out check for the minimum delay bad or ineffective.

I suggest we register an atexit() handler that automatically calls the "create_query_file()" function internally. This way we do not have to cover all the exit paths of the client program. Also, the create_query_file() function does not have to be exposed to the client.

src/lib/Libutil/misc_utils.c
```diff
@@ -92,6 +92,7 @@
 #include <malloc.h>
 #endif
 
+#define MAXLENGTH 40
```
mike0042: We generally use MAXPATHLEN+1 for path names. Forty characters is very small.

subhasisb:
Thanks for the note @mike0042 - please do not restart the build, we will need to debug this intermittent failure with the tests.

test/tests/functional/pbs_client_response.py
```diff
@@ -0,0 +1,56 @@
...
+    client command in 1 second
+    """
+    def test_qstat_reponse(self):
+        """
+        Test to check how many qstat can be done in 1 second.
+        """
+        count = 0
+        t = time.time() + 1
+        while time.time() < t:
+            self.server.status()
```
arungrover: This will fail if PTL ever runs under API mode because there is no sleep there

test/tests/functional/pbs_client_response.py
```diff
@@ -0,0 +1,56 @@
...
+            count += 1
+            print(count)
```
arungrover: You may want to move the print out of loop

bhroam: Don't use print() in a PTL test.  Use self.logger.info()

src/lib/Libutil/misc_utils.c
Jitendra-Rajak: Yes @subhasisb, we are getting seconds in both Linux as well as Windows and then I am converting them to miliseconds for our calculations

PR #1282, opened 2019-09-05
Fixed race condition in test_wait_to_queued of TestqstatStateCount

Klovely:
#### Describe Bug or Feature
Test  test_wait_to_queued  failed sometime due to a race condition.

**Problem : -** 
test_wait_to_queued of TestqstatStateCount failed some time while verifying job_state=Q 
at below step
self.server.expect(JOB, {'job_state': 'Q'}, id=jid, offset=3)
Rarely hit this issue where test expect job to be in Q but got as got: job_state = R , even after 60th attempt
**Reason : -** 
In test we are submitting 4 jobs (say J1, J2 ,J3 and J4 )  test will fail in case where 
all three jobs J2 , J3 and J4  enqued at workq at same time and if  by that time start time of job J2 also meet and  as we have only two ncpus on node, job J2  can run instead of job J4. While as per test J4 should be running (R) state and J2 should move  to Q from W  state because ncpus is not available .

**Solution : -** 
Increased 'timedelta'  from 3 to 7 in 
jid = self.submit_waiting_job(3)
so that even if all three jobs enqued at same time at workq only jobs J3 and J4  will get resources and run , and job J2  will be in Q from W .
and also look for job J3 and J4 is in R state after submission to confirm that 2  ncpus actually consumed and no resources available to run the job J2   .

#### Describe Your Change
Increased 'timedelta'  from 3 to 7 in 
jid = self.submit_waiting_job(3)
so that even if all three jobs enqued at same time at workq only jobs J3 and J4  will get resources and run , and job J2  will be in Q from W .
and also look for job J3 and J4 is in R state after submission to confirm that 2  ncpus actually consumed and no resources available to run the job J2   .

Note : - More detail is given in internal ticket .

[after_fix_test_logs_x37.txt](https://github.com/PBSPro/pbspro/files/3579438/after_fix_test_logs_x37.txt)
[after_fix_test_wait_to_queued_logs_x37.txt](https://github.com/PBSPro/pbspro/files/3579439/after_fix_test_wait_to_queued_logs_x37.txt)

### DIFF
--- a/test/tests/functional/pbs_qstat_count.py
+++ b/test/tests/functional/pbs_qstat_count.py
@@ -80,10 +80,10 @@ class TestqstatStateCount(TestFunctional):
         counts['all_state_count'] = all_state_count
         counts['total_jobs'] = int(qstat[0]['total_jobs'])
         # Find queued count from output of qstat
-        counts['expected_queued_count'] = (counts['total_jobs']
-                                           - counts['Held']
-                                           - counts['Waiting']
-                                           - counts['Running'])
+        counts['expected_queued_count'] = (counts['total_jobs'] -
+                                           counts['Held'] -
+                                           counts['Waiting'] -
+                                           counts['Running'])
         return counts
 
     def verify_count(self):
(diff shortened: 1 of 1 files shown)


Klovely:
Latest execution logs:
[after_fix_test_logs2.txt](https://github.com/PBSPro/pbspro/files/3580225/after_fix_test_logs2.txt)
[after_fix_test_logs2_x37.txt](https://github.com/PBSPro/pbspro/files/3580226/after_fix_test_logs2_x37.txt)

Klovely:
@prakashcv13 as per discussion , I changed 'timedelta' from 7 to 10 so that we have enough time on slower machines as well, Please take a look .

Latest logs: 
[after_fix_test_logs3_a01.txt](https://github.com/PBSPro/pbspro/files/3582337/after_fix_test_logs3_a01.txt)

prakashcv13:
merged #1282

PR #1284, opened 2019-09-09
Apply requirements decorator to test which require mom to be installed on host other than server host

Mahalty:
#### Describe Bug or Feature
Apply requirements decorator to tests which require mom to be installed on  host other than server host

### DIFF
--- a/test/tests/functional/pbs_power_provisioning_sgi.py
+++ b/test/tests/functional/pbs_power_provisioning_sgi.py
@@ -38,6 +38,7 @@
 from tests.functional import *
 
 
+@requirements(no_mom_on_server=True)
 class Test_power_provisioning_sgi(TestFunctional):
 
     """


test/tests/functional/pbs_power_provisioning_sgi.py
```diff
@@ -38,6 +38,7 @@
 from tests.functional import *
 
 
+@requirements(no_mom_on_server=True)
```
agrawalravi90: Is this really necessary? The test suite's custom setUp() deletes any mom on the server if present.

hirenvadalia: @agrawalravi90 Yes it is necessary because there are external tools which uses this information to create PBS cluster while running these tests... + Purpose of `@requirements` is to provide information to such external tools and skip tests if requirements don't match.

If you think this is not right then I suggest you to start discussion whatever you think but for now, I think we should allow this PR.

For more detail on requirements decorator please have a look at https://pbspro.atlassian.net/wiki/spaces/PD/pages/458883073/PP-1281+New+decorator+in+PTL+using+which+user+can+provide+cluster+information+required+for+a+test.

agrawalravi90: So, allowing this PR will mean that I as a developer will not be able to run this test on my dev system unless I first delete the mom on my system before running this test, otherwise the requirements decorator will skip it, right? As of now, the test itself does this for me, just like all other PTL tests which create/delete various PBS daemons. That seems undesirable to me.

If we need a decorator to convey information to an external tool then why don't we create one which will not skip the test ? or add a skip argument to the same decorator which can be set to "False".

shilpakodli: @agrawalravi90 TestSuite's setup is run after decorators are applied. If we add skip argument to existing decorator then TH2 will not be able to find if it is single node test or multi-node test.

agrawalravi90: Thanks for your inputs @shilpakodli . I'm actually not aware of how TH2 works, we can brainstorm offline what's the best way to solve this since TH2 is Altair's internal software. But as long as this change affects PTL testing negatively/doesn't make sense as far as the open source version is concerned, I cannot allow it.

prakashcv13: As per my understanding of the document - we compare the requirements flags with what is passed to pbs_benchpress command. I do not think we would need to delete the mom on the server host for the test/testsuite to not be skipped. If we do need to delete the mom, the design needs to be clear about that.
If my understanding is correct - we can allow this PR to go ahead.

kjakkali: @agrawalravi90 : The test is specific to SGI platforms. To simulate the SGI cluster it expects no mom on server node. More ever this test doesn't do much of non SGI platforms. This PR is about adding test requirements using new decorator, so that external tools can create cluster and run the test cases.   
May be @ashwathraop  can add more details.

shilpakodli: > As per my understanding of the document - we compare the requirements flags with what is passed to pbs_benchpress command. I do not think we would need to delete the mom on the server host for the test/testsuite to not be skipped. If we do need to delete the mom, the design needs to be clear about that.
> If my understanding is correct - we can allow this PR to go ahead.

I agree.

hirenvadalia: > So, allowing this PR will mean that I as a developer will not be able to run this test on my dev system unless I first delete the mom on my system before running this test, otherwise the requirements decorator will skip it, right?

Well, this depends on what you pass in -p in pbs_benchpress if you don't pass nomom=<sever node> while running pbs_benchpress then yes requirements decorator will skip it and that's exectly what we want. We don't want to run this kind a test which is meant to run with only remote mom.

> If we need a decorator to convey information to an external tool then why don't we create one which will not skip the test ? or add a skip argument to the same decorator which can be set to "False".

As I said in an earlier comment, if you think we need to change requirements decorator then we can discuss it but blocking this PR is not the right thing to do.

> But as long as this change affects PTL testing negatively/doesn't make sense as far as the open source version is concerned, I cannot allow it.

Can you please elaborate more on how this change is affecting PTL testing negatively?

hirenvadalia: > I do not think we would need to delete the mom on the server host for the test/testsuite to not be skipped. If we do need to delete the mom, the design needs to be clear about that.

@prakashcv13 You are absolutely right, we don't need to delete or do anything to skip test. It should be handled by requirements decorator which it is doing... aka skip test if mom is specified on the server node.

prakashcv13:
@agrawalravi90 - as per @hirenvadalia's explanation I think there is no issue with going ahead with this PR. Could you please elaborate further on your dislike of this change. If I have understood, we are not affecting PTL testing negatively, but maybe I am missing something, it would be great if you explain your dislike.

agrawalravi90:
> @agrawalravi90 - as per @hirenvadalia's explanation I think there is no issue with going ahead with this PR. Could you please elaborate further on your dislike of this change. If I have understood, we are not affecting PTL testing negatively, but maybe I am missing something, it would be great if you explain your dislike.

Sure. Without this change, one can run this PTL test without worrying about what kind of setup they have on their machine as the setUp() in the test suite takes care of deleting any mom if present on the server host. With the addition of this change, from what I understand, one either has to delete the local mom, or provide pbs_benchpress -p nomom=<server host>, otherwise the test will get skipped by this decorator. So, I don't like the idea of adding an additional pain point for the test runner without adding any benefit to the community. If tomorrow an outside contributor wants to remove this decorator from the test as they see that it serves no purpose to the test (and adds a pain point while running it), will we ask them to not do that since it helps our internal tooling? It doesn't seem justifiable and that's why I don't like it.

hirenvadalia:
> Without this change, one can run this PTL test without worrying about what kind of setup they have on their machine as the setUp() in the test suite takes care of deleting any mom if present on the server host.

@agrawalravi90 Looks like you get test and -p options in pbs_benchpress wrong, this test will not run without passing remote mom, so anyways to run this test you have to pass -p moms which include remote mom
+ As of now in PTL if you delete mom on server node (outside of PTL) or not it doesn't affect PTL because by default PTL initialize mom on the server node. if you want PTL to not consider and don't initialize mom on server node you have to pass nomom option. So in my view what you are saying is not possible at all as of now in PTL and to run this test (with or without this change) you have to pass nomom=servernode in pbs_benchpress.

prakashcv13:
@agrawalravi90 - I was about to say the same, but @hirenvadalia beat me to it. I do understand your dislike, but that would need discussion on an already accepted design which was floated more than 6 months back. As of now, I would go ahead and merge this, we can surely revisit the design, if the need be.

agrawalravi90:
> As of now in PTL if you delete mom on server node (outside of PTL) or not it doesn't affect PTL because by default PTL initialize mom on the server node
Thanks for reminding me, yes, so the test runner will have to provide the nomom option after this change if they want their test to not be skipped, it's still an additional option that they didn't have to provide before but will have to provide now.

> So in my view what you are saying is not possible at all as of now in PTL and to run this test (with or without this change) you have to pass nomom=servernode in pbs_benchpress

Well not really, the setUp() of the test suite takes care of deleting any mom on the server host, so one could run this test suite with other test suites which might require a mom on the server node without providing the nomom option, right?

agrawalravi90:
> @agrawalravi90 - I was about to say the same, but @hirenvadalia beat me to it. I do understand your dislike, but that would need discussion on an already accepted design which was floated more than 6 months back. As of now, I would go ahead and merge this, we can surely revisit the design, if the need be.

Sure, your call. But I'd just like to point out that the design was for a decorator which can be applied to tests which we want to get skipped if the correct requirements are not fulfilled. This test only required -p moms, now we've made it require -p nomom along with -p moms, even though the test doesn't need it.

prakashcv13:
@agrawalravi90 - I am of the view that we should remove the redundant code that deletes the mom on the server node as that will already be taken care of by no_mom_on_server. But, that I believe can be done as part of another PR, so as to not stall work dependent on this PR.

hirenvadalia:
> Thanks for reminding me, yes, so the test runner will have to provide the nomom option after this change if they want their test to not be skipped, it's still an additional option that they didn't have to provide before but will have to provide now.

NO... after this change also there is no effect on how you run this test... it will be same as before...

> Well not really, the setUp() of the test suite takes care of deleting any mom on the server host, so one could run this test suite with other test suites which might require a mom on the server node without providing the nomom option, right?

You can run this test without nomom but it won't run as it needs a remote mom.

agrawalravi90:
> You can run this test without nomom but it won't run as it needs a remote mom.

Ok, so this test was broken before? Well, it would have been useful to know that!

> @agrawalravi90 - I am of the view that we should remove the redundant code that deletes the mom on the server node as that will already be taken care of by no_mom_on_server. But, that I believe can be done as part of another PR, so as to not stall work dependent on this PR.

If what Hiren says is correct and this test was actually broken before then why don't we fix it properly and just remove the code in setUp()? It's just a select and delete, shouldn't take a lot of time.

agrawalravi90:
@ashwathraop I think you authored this test, can you please clarify if this test needed to be run with -p nomom=serverhost, or could you run it before with only -p moms=momhosts? It seems like setUp() can handle not being provided -p nomom, but Hiren is suggesting that it was necessary, so it would be great if you could clarify, thanks.

hirenvadalia:
> You can run this test without nomom but it won't run as it needs a remote mom.

Just clarifying this statement: "it won't run" === "it will get skipped", not that it will fail

prakashcv13:
Looking at it only from the perspective of the purpose of this PR which is to add the decorator with flag no_mom_on_server=True (which by default is False). 
On reading the testsuite, it is clear that it needs mom to not be there on the server host. The setUp() is written that way.
Coming to @agrawalravi90's dislike, which to my understanding is that adding this decorator creates a pain-point for the user. This I believe can be addressed by revisiting the design of this decorator which was discussed and accepted a long time ago. But if the need be, must be done.
Whether to remove the code in setUp(), I believe is out of scope for this PR.

PR #1294, opened 2019-09-13
adjustments on newer OS versions that have shell functions in environ…

vccardenas:
…ment

#### Describe Bug or Feature
Newer OS versions like SLES15, RHEL8, RHEL7, and Ubuntu 18 might have shell functions in the user environment, for example,

1) on SLES 15:
BASH_FUNC_module%%=() { eval $($LMOD_CMD bash "$@") && eval $(${LMOD_SETTARG_CMD:-:} -s sh)
}
BASH_FUNC_ml%%=()
{ eval $($LMOD_DIR/ml_cmd "$@") }

2) on RHEL7:
BASH_FUNC_module()=() {  eval $($LMOD_CMD bash "$@") && eval $(${LMOD_SETTARG_CMD:-:} -s sh)
}
BASH_FUNC_ml()=() {  eval $($LMOD_DIR/ml_cmd "$@")
}

that will cause the TestNonprintingCharacters tests test_nonprint_shell_function, test_terminal_control_shell_function to fail.

#### Describe Your Change
Changed the grep in the job script.

#### Attach Test Logs/Output
[test_shell_func_s15_before.txt](https://github.com/PBSPro/pbspro/files/3611976/test_shell_func_s15_before.txt)
[shell_function_s15_after.txt](https://github.com/PBSPro/pbspro/files/3618685/shell_function_s15_after.txt)
[shell_function_r6_after.txt](https://github.com/PBSPro/pbspro/files/3618686/shell_function_r6_after.txt)

[shell_function_s15_after1.txt](https://github.com/PBSPro/pbspro/files/3622291/shell_function_s15_after1.txt)

### DIFF
--- a/test/tests/functional/pbs_nonprint_characters.py
+++ b/test/tests/functional/pbs_nonprint_characters.py
@@ -141,8 +141,7 @@ sleep 5
         Check if unescaped variable is in job output
         """
         self.server.expect(JOB, 'queue', op=UNSET, id=jid, offset=1)
-        cmd = ['cat', job_outfile]
-        ret = self.du.run_cmd(self.server.hostname, cmd=cmd)
+        ret = self.du.cat(filename=job_outfile)
         job_out = '\n'.join(ret['out'])
         self.logger.info('job output from %s:\n%s' % (job_outfile, job_out))
         job_output = ""
--- a/test/tests/functional/pbs_nonprint_characters.py
+++ b/test/tests/functional/pbs_nonprint_characters.py
@@ -89,11 +89,6 @@ unset -f foo
 exit 0
 """
         self.script = """#PBS -V
-for e in $(env | grep -v BASH_FUNC_foo | grep BASH_FUNC_ | \
-sed 's/BASH_FUNC_//' | sed 's/[%(].*//')
-do
-    unset -f $e
-done
 env | grep -A2 foo
 foo
 sleep 5
@@ -146,6 +141,10 @@ sleep 5
         Check if unescaped variable is in job output
         """
         self.server.expect(JOB, 'queue', op=UNSET, id=jid, offset=1)
+        cmd = ['cat', job_outfile]
+        ret = self.du.run_cmd(self.server.hostname, cmd=cmd)
+        job_out = '\n'.join(ret['out'])
+        self.logger.info('job output from %s:\n%s' % (job_outfile, job_out))
         job_output = ""
         with open(job_outfile, 'r') as f:
             job_output = f.read().strip()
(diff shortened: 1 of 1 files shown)


mike0042:
Is the qstat -f output actually showing inaccurate data when the BASH_FUNC_* environment variables are set? If so, unsetting the variables in the test does not fix the underlying issue.

hirenvadalia:
@mike0042 and @vccardenas I am not sure about qstat but PTL's BatchUtils.convert_to_dictlist has a bug where if you have multiline line (more than 2 lines) env variable then it fails to parse qstat output.
Env like:
```
export MYMULTILINE_VAR='a
a
}
{
|
asd,

?
@
#
$
%
^
\&
\*
,
\n
    a = aa                  # This line start with 4 spaces
	PBS_O_LOGNAME=pbsroot   # This line start with one tab
    run_count = 1           # This line start with 4 spaces
as
,
\t
'
```

mike0042:
Thanks for mentioning the PTL bug @hirenvadalia. I think this PR needs further analysis prior to review. Please mark as [WIP] and determine what is working and what is broken.

vccardenas:
@mkaro and @hirenvadalia, I have used the run_cmd to work around the PTL bug.

@mkaro, if the other non-foo shell functions exist in the environment we get this foo shell function:
BASH_FUNC_foo%%=() {  a=\x03;\n echo XX${a}YY}\n}\nBASH_FUNC_module%%=() {  eval $($LMOD_CMD bash "$@") && eval $(${LMOD_SETTARG_CMD:-:} -s sh)\nXX\x03YY}

mike0042:
@vccardenas - Having looked a little closer at this, it appears that we're dealing with a name collision. Both BASH_FUNC_foo() and foo() both show up in the job output due to the fact the job greps for "foo". I would suggest you either grep for "^foo" or use a different name like "ptlfoo".

mike0042:
This looks much better @vccardenas. Please update the test logs.

vccardenas:
The test logs from the latest changes have been attached.

test/tests/functional/pbs_nonprint_characters.py
```diff
@@ -136,6 +141,10 @@ def check_jobout(self, chk_var, jid, job_outfile):
         Check if unescaped variable is in job output
         """
         self.server.expect(JOB, 'queue', op=UNSET, id=jid, offset=1)
+        cmd = ['cat', job_outfile]
```
hirenvadalia: I suggest using self.du.cat()

PR #1296, opened 2019-09-17
Support PTL to import and export hook config file

shilpakodli:
#### Describe Bug or Feature
Support PTL to import and export hook config files.

#### Describe Your Change
Added import_hook_config and export_hook_config in pbs_testlib.py to support import and export hook config files respectively.

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -9732,7 +9732,8 @@ class Server(PBSService):
         else:
             hook_t = PBS_HOOK
 
-        fn = self.du.create_temp_file(body=json.dumps(hook_conf))
+        hook_body = json.dumps(hook_conf, indent=4)
+        fn = self.du.create_temp_file(body=hook_body)
 
         if not self._is_local:
             tmpdir = self.du.get_tempdir(self.hostname)
@@ -9752,7 +9753,7 @@ class Server(PBSService):
             self.du.rm(self.hostname, rfile)
         self.logger.log(level, 'server ' + self.shortname +
                         ': imported hook config\n---\n' +
-                        str(hook_conf) + '---')
+                        str(hook_body) + '\n---\n')
         return True
 
     def export_hook_config(self, hook_name, hook_type):
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -9711,6 +9711,79 @@ class Server(PBSService):
 
         return ret
 
+    def import_hook_config(self, hook_name, hook_conf, hook_type,
+                           level=logging.INFO):
+        """
+        Helper function to import hook config body into hook by name.
+        The hook must have been created prior to calling this
+        function.
+
+        :param hook_name: The name of the hook to import body to
+        :type name: str
+        :param hook_conf: The body of the hook config as a string.
+        :type hook_conf: str
+        :param hook_type: The hook type "site" or "pbshook"
+        :type hook_type: str
+        :returns: True on success.
+        :raises: PbsManagerError
+        """
+        if hook_type == "site":
+            hook_t = HOOK
+        else:
+            hook_t = PBS_HOOK
+
+        fn = self.du.create_temp_file(body=hook_conf)
+
+        if not self._is_local:
+            tmpdir = self.du.get_tempdir(self.hostname)
+            rfile = os.path.join(tmpdir, os.path.basename(fn))
+            self.du.run_copy(self.hostname, fn, rfile)
+        else:
+            rfile = fn
+
+        a = {'content-type': 'application/x-config',
+             'content-encoding': 'default',
+             'input-file': rfile}
+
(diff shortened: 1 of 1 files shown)


test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -9578,6 +9579,79 @@ def create_import_hook(self, name, attrs=None, body=None, overwrite=True,
...
+                           level=logging.INFO):
+        """
+        Helper function to import hook config body into hook by name.
+        The hook must have been created prior to calling this
+        function.
+
+        :param hook_name: The name of the hook to import body to
+        :type name: str
+        :param hook_conf: The body of the hook config as a string.
+        :type hook_conf: str
```
hirenvadalia: shouldn't this be dict instead str?

shilpakodli: User should pass dict quoted as string so that we will write it into a file and export

hirenvadalia: I think it will be better to allow use pass dict and this method will use `json.dumps` then write in the file.
Think of this use case: user calls export_hook_conf then updates returned dict then calls this method. it will be simple for the user else the user has to convert dict into a string then pass it to this.

prakashcv13: i agree with @hirenvadalia here, the user need not do the conversion.

shilpakodli:
@hirenvadalia and @prakashcv13 Sorry for addressing this late. Please review code changes.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -9600,7 +9600,7 @@ def import_hook_config(self, hook_name, hook_conf, hook_type,
         else:
             hook_t = PBS_HOOK
 
-        fn = self.du.create_temp_file(body=hook_conf)
+        fn = self.du.create_temp_file(body=json.dumps(hook_conf))
```
hirenvadalia: Optional: I suggest to pass indent=2 for it will dump JSON in a pretty format, will be easy to read later (mostly when it get saved in post-analysis data)

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -9620,7 +9620,7 @@ def import_hook_config(self, hook_name, hook_conf, hook_type,
             self.du.rm(self.hostname, rfile)
         self.logger.log(level, 'server ' + self.shortname +
                         ': imported hook config\n---\n' +
-                        hook_conf + '---')
+                        str(hook_conf) + '---')
```
hirenvadalia: `'---'` === '`\n---'`

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -9578,6 +9579,81 @@ def create_import_hook(self, name, attrs=None, body=None, overwrite=True,
...
+                        str(hook_body) + '\n---\n')
+        return True
+
+    def export_hook_config(self, hook_name, hook_type):
+        """
+        Helper function to export hook config body.
+        The hook must have been created prior to calling this
+        function.
+
+        :param hook_name: The name of the hook to import body to
```
prakashcv13: import or export?

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -9578,6 +9579,81 @@ def create_import_hook(self, name, attrs=None, body=None, overwrite=True,
...
+        if hook_type == "site":
+            hook_t = "hook"
+        else:
+            hook_t = "pbshook"
+        cmd = ["export", hook_t, hook_name]
+        cmd += ["application/x-config", "default"]
+        cmd = " ".join(cmd)
+        pcmd = [os.path.join(self.pbs_conf['PBS_EXEC'], 'bin', 'qmgr'),
+                '-c', cmd]
+        ret = self.du.run_cmd(self.hostname, pcmd, sudo=True)
```
prakashcv13: would it make sense to put this in try/catch?

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -9578,6 +9579,81 @@ def create_import_hook(self, name, attrs=None, body=None, overwrite=True,
...
+        else:
+            hook_t = PBS_HOOK
+
+        hook_body = json.dumps(hook_conf, indent=4)
+        fn = self.du.create_temp_file(body=hook_body)
+
+        if not self._is_local:
+            tmpdir = self.du.get_tempdir(self.hostname)
+            rfile = os.path.join(tmpdir, os.path.basename(fn))
+            self.du.run_copy(self.hostname, fn, rfile)
```
prakashcv13: would it help to check the return value here?

test/fw/ptl/lib/pbs_testlib.py
shilpakodli: du.run_cmd doesn't raise any exceptions. It returns 0 on success or non-zero value on failure.

PR #1297, opened 2019-09-18
miscellaneous maintenance fixes

vchlum:
#### Describe Bug or Feature
* The First bug is regarding creating maintenance while node(s) without ncpus is in PBS. This situation is not usual but it is not a bug to have a node without ncpus. For example, If we add a node that is not up/online, no ncpus is set automatically and the creation of maintenance is denied with error "pbs_rsub: Attribute value error" for any host. This is not correct behavior. IMHO, it is better to simply ignore such a node while creating maintenance.

* The second bug: The maintenance can be created only by a manager or operator but it is possible to delete or alter the maintenance by a regular user.

#### Describe Your Change
* With the first bug, we can simply ignore nodes without ncpus by not throwing an error.

* The second bug needs to test the first char in maintenance id and for such maintenance check the correct permissions. This is done in chk_rescResv_request().

Manual test for first bug:
```
(JESSIE)vchlum@torque3:~$ pbsnodes -v torque4
torque4
     Mom = torque4.ics.muni.cz
     Port = 15002
     pbs_version = unavailable
     ntype = PBS
     state = state-unknown,down
     resources_available.host = torque4
     resources_available.vnode = torque4
     resources_assigned.accelerator_memory = 0kb
     resources_assigned.hbmem = 0kb
     resources_assigned.mem = 0kb
     resources_assigned.naccelerators = 0
     resources_assigned.ncpus = 0
     resources_assigned.vmem = 0kb
     comment = node down: communication closed
     resv_enable = True
     sharing = default_shared
     last_state_change_time = Wed Sep 18 09:12:44 2019

(JESSIE)vchlum@torque3:~$ pbs_rsub -R 1200 -D 3600 --hosts torque4
pbs_rsub: missing host(s)
usage: pbs_rsub [-I seconds] [-m mail_points] [-M mail_list]
                [-N reservation_name] [-u user_list] [-g group_list]
                [-U auth_user_list] [-G auth_group_list] [-H auth_host_list]
                [-R start_time] [-E end_time] [-D duration] [-q destination]
                [-r rrule_expression] [-W otherattributes=value...]
                -l resource_list | --hosts host1 [... hostn]
       pbs_rsub --version
(JESSIE)vchlum@torque3:~$ 
```

PTL maintenance test:
[ptl_maintenance_output.txt](https://github.com/PBSPro/pbspro/files/3625847/ptl_maintenance_output.txt)

### DIFF
--- a/src/cmds/pbs_rsub.c
+++ b/src/cmds/pbs_rsub.c
@@ -806,7 +806,7 @@ main(int argc, char *argv[], char *envp[])
 				if (ncpus_str != NULL)
 					ncpus = strtol(ncpus_str, &endp, 0);
 				
-				if (*endp != '\0' || ncpus_str == NULL) {
+				if (*endp != '\0') {
 					fprintf(stderr, "pbs_rsub: Attribute value error\n");
 					CS_close_app();
 					exit(2);
--- a/src/cmds/pbs_rsub.c
+++ b/src/cmds/pbs_rsub.c
@@ -95,6 +95,7 @@ process_opts(int argc, char **argv, struct attrl **attrp, char *dest)
 	char dur_buf[800];
 	char badw[] = "pbs_rsub: illegal -W value\n";
 	int opt_re_flg = FALSE;
+	int opt_res_req_flg = FALSE;
 #ifdef WIN32
 	struct attrl *ap = NULL;
 	short nSizeofHostName = 0;
@@ -143,6 +144,7 @@ process_opts(int argc, char **argv, struct attrl **attrp, char *dest)
 				break;
 
 			case 'l':
+				opt_res_req_flg = TRUE;
 				if ((i = set_resources(&attrib, optarg, 0, &erp)) != 0) {
 					if (i > 1) {
 						pbs_prt_parse_err("pbs_rsub: illegal -l value\n", optarg,
@@ -274,6 +276,11 @@ process_opts(int argc, char **argv, struct attrl **attrp, char *dest)
 
 	if (opt_re_flg == TRUE && qmoveflg ==TRUE) {
 		fprintf(stderr, "pbs_rsub: -Wqmove is not compatible with -R or -E option\n");
+		errflg++;
+	}
+
+	if (opt_res_req_flg && is_maintenance_resv) {
+		fprintf(stderr, "pbs_rsub: can't use -l with --hosts\n");
 		errflg++;
 	}
 
--- a/src/cmds/pbs_rsub.c
+++ b/src/cmds/pbs_rsub.c
@@ -293,17 +293,20 @@ process_opts(int argc, char **argv, struct attrl **attrp, char *dest)
 
 	if (is_maintenance_resv) {
 		char **hostp = NULL;
+		int num_hosts = argc - optind;
 
-		if (argc - optind > 0) {
-			maintenance_hosts = malloc(sizeof(char *) * (argc - optind + 1));
+		if (num_hosts > 0) {
+			int i;
+
+			maintenance_hosts = malloc(sizeof(char *) * (num_hosts + 1));
 			if (maintenance_hosts == NULL) {
 				fprintf(stderr, "pbs_rsub: Out of memory\n");
 				return (++errflg);
 			}
 
-			maintenance_hosts[argc - optind] = NULL;
+			maintenance_hosts[num_hosts] = NULL;
 
-			for (; optind < argc; optind++) {
+			for (i = 0; optind < argc; optind++, i++) {
 				hostp = maintenance_hosts;
 				for (; *hostp; hostp++) {
 					if (strcmp(*hostp, argv[optind]) == 0) {
@@ -312,8 +315,15 @@ process_opts(int argc, char **argv, struct attrl **attrp, char *dest)
 					}
 				}
 
-				maintenance_hosts[argc - (optind + 1)] = strdup(argv[optind]);
-				if (maintenance_hosts[argc - (optind + 1)] == NULL) {
+				if (strlen(argv[optind]) == 0) {
+					num_hosts--;
+					i--;
+					maintenance_hosts[num_hosts] = NULL;
+					continue;
+				}
+
+				maintenance_hosts[i] = strdup(argv[optind]);
+				if (maintenance_hosts[i] == NULL) {
 					fprintf(stderr, "pbs_rsub: Out of memory\n");
 					return (++errflg);
 				}
@@ -816,9 +826,6 @@ main(int argc, char *argv[], char *envp[])
 		for (; *hostp; hostp++) {
 			int host_ncpus = 0;
 
-			if (strlen(*hostp) == 0)
-				continue;
-
 			for (bstat = bstat_head; bstat; bstat = bstat->next) {
 				char *ncpus_str = NULL;
 				int ncpus = 0;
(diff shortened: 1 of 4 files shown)


src/server/svr_chk_owner.c
```diff
@@ -550,6 +550,11 @@ chk_rescResv_request(char *resvID, struct batch_request *preq)
 		return NULL;
 	}
 
+	if (resvID[0] == PBS_MNTNC_RESV_ID_CHAR && ! (preq->rq_perm & (ATR_DFLAG_OPWR | ATR_DFLAG_MGWR))) {
```
arungrover: I do not understand this change. Essentially svr_authorize_resvReq does the same thing right? It checks if the request is coming from an Operator or Manager, if not then it compares the reservation owner with the user requesting a rdel or ralter.

vchlum: @arungrover It is not the same. svr_authorize_resvReq() authorizes all Operators and Managers for all reservation, but this change rejects access for all regular users accessing maintenance reservation. It fixes the following problem: A user with the Manager role creates maintenance. Later, the Manager role is removed for this user but the user is still able to remove the maintenance.

vchlum:
> @vchlum Thanks for making this contribution.
> We have seen a few minor bugs in maintenance reservation implementation. I do not intend to derail this contribution, it is definitely useful for all of us. But, if you are interested, following are few issues:
> 
> * When a Maintenance reservation is submitted, it can be submitted with -I  option but this option is only useful because the server needs to report the response of scheduler to the user. Since in case of maintenance reservation requests never reach scheduler this option does not really go with "--hosts" option.
> * If "--hosts" option is given with wrong syntax like a comma-separated list of hosts then it considers the last-mentioned hostname as the host where the reservation will run.
> * When a maintenance reservation is confirmed on the nodes where there is already a running reservation then the nodes are taken away from that running reservation and it gets a new "IN-CONFLICT" substate. But when the maintenance reservation ends, the "IN-CONFLICT" reservation never gets the node back again. I am not sure if this is by design or really an anomaly.

Thank you @arungrover for sharing issues.
* You are right. The -l parameter should be forbidden with --hosts. I will fix it in this contribution.
* If you use --hosts option, a space-separated list of hosts is expected. If some host is not found in pbs, it is simply ignored. It means that if you have: '--hosts host1,host2 host3' then 'host1,host2' is considered to be one host. This host is not found and is ignored. The second host is 'host3'. This host is recognized and the maintenance is confirmed on this host only. How do you suggest to change the behavior? I see a possible change like this: in case of an error in hosts syntax, we could always return an error (and do not confirm any maintenance)?
* I consider this to be by design. Do we want to change this behavior? or simply add the fact into the design document?

vchlum:
I added the fix for the first issue that @arungrover mentioned. The -l and --hosts can not be used together now.
[ptl_maintenance_output2.txt](https://github.com/PBSPro/pbspro/files/3634677/ptl_maintenance_output2.txt)

src/cmds/pbs_rsub.c
```diff
@@ -143,6 +144,7 @@ process_opts(int argc, char **argv, struct attrl **attrp, char *dest)
 				break;
 
 			case 'l':
+				opt_res_req_flg = TRUE;
```
arungrover: @vchlum my comment was about option -I (ATTR_inter) but this works too because --hosts will consume all ncpus on the hosts anyways. Can you please add similar change to "ATTR_inter" option as well?

arungrover:
> * If you use --hosts option, a space-separated list of hosts is expected. If some host is not found in pbs, it is simply ignored. It means that if you have: '--hosts host1,host2 host3' then 'host1,host2' is considered to be one host. This host is not found and is ignored. The second host is 'host3'. This host is recognized and the maintenance is confirmed on this host only. How do you suggest to change the behavior? I see a possible change like this: in case of an error in hosts syntax, we could always return an error (and do not confirm any maintenance)?
> * I consider this to be by design. Do we want to change this behavior? or simply add the fact into the design document?

For the first problem, I feel we should not confirm the reservation if the host does not exist. It gives no indication to the user that reservation has been accepted for only those nodes which are visible to PBS server. 
One more problem with not checking the hosts is that if the same host is given twice in pbs_rsub command, it confirms the reservation on the host. This has no impact as such but it does make server leak a small amount of memory. Here is trace you get from this leak - 
`
==82992== 32 (16 direct, 16 indirect) bytes in 1 blocks are definitely lost in loss record 61 of 1,524

==82992==    at 0x4C29BE3: malloc (vg_replace_malloc.c:299)
==82992==    by 0x44CD00: set_nodes (in /opt/pbs/sbin/pbs_server.bin)
==82992==    by 0x46E940: assign_resv_resc (in /opt/pbs/sbin/pbs_server.bin)
==82992==    by 0x46EBDF: req_confirmresv (in /opt/pbs/sbin/pbs_server.bin)
==82992==    by 0x4B070B: process_socket (in /opt/pbs/sbin/pbs_server.bin)
==82992==    by 0x4B08C1: wait_request (in /opt/pbs/sbin/pbs_server.bin)
==82992==    by 0x42AB89: main (in /opt/pbs/sbin/pbs_server.bin)`

Regarding the second problem, my opinion is that at least the reservation which is marked in conflict should be treated as degraded and another node should be found to keep it running. Currently, a node vanishes underneath the reservation and it never gets it back. The design document does state that IN_CONFLICT state is treated as DEGRADED and gets reconfirmed on other nodes.
Maybe we can have a forum discussion about it and see what others think.
Thanks!

vchlum:
I addressed your comments @arungrover. The branch is in conflict now and AppVeyor will not run. Should I rebase?

I will revive the maintenance topic in the forum regarding the last issue.

vchlum:
New PTL log: [ptl_maintenance_output3.txt](https://github.com/PBSPro/pbspro/files/3645922/ptl_maintenance_output3.txt)

vchlum:
@arungrover It is rebased. There is a problem with Travis on deb 9. I tried several times with the same error. Seems like other PR have the same issue...

arungrover:
I have restarted Travis build again. Now that PTL problem is fixed, it might pass.
@prakashcv13 I don't know if you reviewed recent commits, can you please sign-off again if you are ok with the change?

src/cmds/pbs_rsub.c
```diff
@@ -792,6 +816,9 @@ main(int argc, char *argv[], char *envp[])
 		for (; *hostp; hostp++) {
 			int host_ncpus = 0;
 
+			if (strlen(*hostp) == 0)
+				continue;
```
prakashcv13: @vchlum, I am not sure, if we really need this check.

src/cmds/pbs_rsub.c
```diff
@@ -863,6 +890,13 @@ main(int argc, char *argv[], char *envp[])
 					}
 				}
 			} /* end of part that crafts select */
+
+			/* host not found or host has zero ncpus */
+			if (host_ncpus == 0) {
```
prakashcv13: this if belongs just above the block where we craft the select statement.

test/tests/functional/pbs_maintenance_reservations.py
```diff
@@ -68,9 +68,9 @@ def test_maintenance_acl_denied(self):
 
         self.assertEqual("pbs_rsub: Unauthorized Request", msg)
 
-    def test_maintenance_forbidden_resources(self):
+    def test_maintenance_forbidden_parameters(self):
```
prakashcv13: shall we say "conflicting" parameters?

src/cmds/pbs_rsub.c
vchlum: It handles the special case 'pbs_rsub ... --hosts ""'. It can happen e.g. within the PTL test. I suggest to handle it instead of changing the test.

prakashcv13: Ok. Then, I believe we should handle this in process_opts()?

vchlum:
> A few minor comments.

Thank you @prakashcv13. Please, see the latest changes.

prakashcv13:
@arungrover - now it is your turn to review the latest changes :). if you find them good, please go ahead and merge this PR. Thank you.

vchlum:
@arungrover It is rebased.

PR #1305, opened 2019-09-20
degrading reservations if the scheduled node went down when a scheduling

prakashcv13:
cycle was in progress

#### Describe Bug or Feature

If a reservation is confirmed on a down or stale vnode, it will go into the RESV_CONFIRMED state.  It will not be degraded unless there is some other event that will cause it to go degraded.
The same can happen for jobs as well, but we need not handle that case - for the below reasons -
1) If we reject the runjob request - the scheduler will not be looking for other nodes to run the job on.
2) It does not add value as the node can go down just after the check, so the best place to check is when we send the job for running and that check is already present.

#### Describe Your Change

In set_nodes(), we check for the node to be down, offline or stale and degrade the reservation. Set the retry time and degraded time.

No new PTL tests are written as it is not possible to time the offline-ing of a node during a scheduling cycle. Smoke test for degraded reservation has passed.
[ptl_test_degrade_resv_smoke.txt](https://github.com/PBSPro/pbspro/files/3637492/ptl_test_degrade_resv_smoke.txt)
Manual testing logs are attached
[Standing reservations - manual testing.txt](https://github.com/PBSPro/pbspro/files/3637607/Standing.reservations.-.manual.testing.txt)
[Advance reservations - manual testing.txt](https://github.com/PBSPro/pbspro/files/3637608/Advance.reservations.-.manual.testing.txt)

### DIFF
--- a/src/server/node_manager.c
+++ b/src/server/node_manager.c
@@ -2493,7 +2493,7 @@ stat_update(int stream)
 				/* or resources_used was modified         */
 
 				pjob->ji_wattr[(int)JOB_ATR_session_id].at_flags &= ~ATR_VFLAG_MODIFY;
-				for (i=0; i<JOB_ATR_LAST; ++i) {
+				for (i = 0; i < JOB_ATR_LAST; ++i) {
 					if (pjob->ji_wattr[i].at_flags & ATR_VFLAG_MODIFY) {
 						job_save(pjob, SAVEJOB_FULL);
 						break;
@@ -2767,7 +2767,7 @@ discard_job(job *pjob, char *txt, int noack)
 		/* had better be the "natural" vnode with only the one parent */
 		if (pnode != NULL) {
 
-			for (i=0; i < nmom; ++i) {
+			for (i = 0; i < nmom; ++i) {
 				if ((pdsc+i)->jdcd_mom == pnode->nd_moms[0])
 					break;		/* already have this Mom */
 			}
@@ -3343,7 +3343,7 @@ ping_nodes(struct work_task *ptask)
 				return;
 			}
 
-			for (i=0; i<mominfo_array_size; i++) {
+			for (i = 0; i < mominfo_array_size; i++) {
 				if (mominfo_array[i]) {
 					ping_a_mom_mcast(mominfo_array[i], 0,
 						mtfd_ishello, mtfd_isnull,
@@ -3425,7 +3425,7 @@ setup_pnames(char *namestr)
 
 		/* is the resource name already in the pnames attribute? */
 		if (pparst) {
-			for (i=0; i<pparst->as_usedptr; ++i) {
+			for (i = 0; i < pparst->as_usedptr; ++i) {
 				if (strcasecmp(ps, pparst->as_string[i]) == 0)
 					break;
 			}
@@ -3521,7 +3521,7 @@ cross_link_mom_vnode(struct pbsnode *pnode, mominfo_t *pmom)
 
 	/* see if the node already has this Mom listed,if not add her */
 
-	for (i=0; i<pnode->nd_nummoms; ++i) {
+	for (i = 0; i < pnode->nd_nummoms; ++i) {
 		if (pnode->nd_moms[i] == pmom)
 			break;
 	}
@@ -3563,7 +3563,7 @@ cross_link_mom_vnode(struct pbsnode *pnode, mominfo_t *pmom)
 	/* Now set reverse linkage Mom -> node */
 
 	prmomsvr = pmom->mi_data;
-	for (i=0; i<prmomsvr->msr_numvnds; ++i) {
+	for (i = 0; i < prmomsvr->msr_numvnds; ++i) {
 		if (prmomsvr->msr_children[i] == pnode)
 			break;
 	}
@@ -3664,7 +3664,7 @@ update2_to_vnode(vnal_t *pvnal, int new, mominfo_t *pmom, int *madenew, int from
 
 		int have_topology   = 0;
 		int is_compute_node = 0;
-		for (i=0; i < pvnal->vnal_used; i++) {
+		for (i = 0; i < pvnal->vnal_used; i++) {
 			psrp = VNAL_NODENUM(pvnal, i);
 			if (strcasecmp(psrp->vna_name, ATTR_NODE_TopologyInfo) == 0)
 				have_topology = 1;
@@ -3785,7 +3785,7 @@ update2_to_vnode(vnal_t *pvnal, int new, mominfo_t *pmom, int *madenew, int from
 	 */
 
 	if (!from_hook) {
-		for (i=0; i < ND_ATR_LAST; ++i) {
+		for (i = 0; i < ND_ATR_LAST; ++i) {
 			/* if this vnode has been updated earlier in this update2 */
 			/* then don't free anything but topology */
 			if ((i != ND_ATR_TopologyInfo) && (pnode->nd_modified & NODE_UPDATE_VNL))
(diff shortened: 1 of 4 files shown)


prakashcv13:
This PR is open for review.

src/server/node_manager.c
```diff
@@ -8776,3 +8779,57 @@ update_node_rassn(attribute *pexech, enum batch_op op)
...
+	(void) resv_setResvState(presv, presv->ri_qs.ri_state, RESV_DEGRADED);
+
+	/* the number of vnodes down could exceed the number of vnodes in
+	 * the reservation only in the case of a standing reservation for
+	 * which the vnodes unavailable are associated to later occurrences
+	 */
+	if (presv->ri_vnodes_down > presv->ri_vnodect) {
+		attribute *rsv_attr = presv->ri_wattr;
+		/* If a standing reservation we print the execvnodes sequence
+		 * string for debugging purposes */
```
bhroam: Move the trailing */ to the next line

src/server/node_manager.c
```diff
@@ -8776,3 +8779,57 @@ update_node_rassn(attribute *pexech, enum batch_op op)
...
+		if (rsv_attr[RESV_ATR_resv_standing].at_val.at_long) {
+			snprintf(log_buffer, sizeof(log_buffer), " execvnodes sequence %s",
```
bhroam: Make this a fully fledged log message.  It looks like a message used to debug your code.  It won't really be useful except for the person who put in the log message

src/server/node_manager.c
```diff
@@ -8776,3 +8779,57 @@ update_node_rassn(attribute *pexech, enum batch_op op)
...
+				rsv_attr[RESV_ATR_resv_execvnodes].at_val.at_str);
+			log_event(PBSEVENT_DEBUG, PBS_EVENTCLASS_RESV, LOG_DEBUG,
+				presv->ri_qs.ri_resvID, log_buffer);
```
bhroam: DEBUG is printed at the default logging level.  This seems to me like a lengthy debug message that should be at a higher log level.

src/server/node_manager.c
```diff
@@ -7130,10 +7130,13 @@ set_nodes(void *pobj, int objtype, char *execvnod_in, char **execvnod_out, char
 				return (PBSE_UNKNODE);
 			}
 
+			if ((pnode->nd_state & (INUSE_DOWN | INUSE_STALE | INUSE_OFFLINE)) && (svr_init == FALSE))
+				if ((objtype == RESC_RESV_OBJECT) && (presv->ri_qs.ri_resvID[0] != PBS_MNTNC_RESV_ID_CHAR) /*&& (presv->ri_qs.ri_state == RESV_UNCONFIRMED)*/)
+					set_resv_for_degrade(pnode, presv);
```
bhroam: There is a chunk of code in this function where we check the same thing for jobs and reservations.  Please combine these two checks.

src/server/node_manager.c
```diff
@@ -1771,7 +1772,7 @@ vnode_unavailable(struct pbsnode *np, int account_vnode)
 				/* If reservation is currently running and a node is down then
 				 * set its substate to degraded
 				 */
-				resv_setResvState(presv, presv->ri_qs.ri_state, RESV_DEGRADED);
+				(void) resv_setResvState(presv, presv->ri_qs.ri_state, RESV_DEGRADED);
```
bhroam: Why void this?  No warning will be emitted.

bhroam:
You can test this manually via a debugger.  You put a breakpoint after query_nodes() and then offline your nodes.

src/server/node_manager.c
```diff
@@ -8776,3 +8779,57 @@ update_node_rassn(attribute *pexech, enum batch_op op)
...
 }
+
+static void
+set_resv_for_degrade(struct pbsnode *pnode, resc_resv *presv)
+{
+	char *str_time;
+	long degraded_time;
+
+	if (presv->ri_wattr[RESV_ATR_resv_standing].at_val.at_long == 0)
+		presv->ri_degraded_time = presv->ri_wattr[RESV_ATR_start].at_val.at_long;
```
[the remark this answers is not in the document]
prakashcv13: i have kept the logic same as in vnode_unavailable()

src/server/node_manager.c
prakashcv13: there are 2 other calls to the same function and we coid void the return return value there.

prakashcv13:
hey @bhroam, thank you for the review. For testing I have put the breakpoint in is_ok_to_run()

src/server/node_manager.c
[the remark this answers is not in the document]
bhroam: I guess I don't understand what presv->ri_degraded_time is used for.  I was assuming it was the time the reservation entered the degraded state.  Speaking of vnode_unavailable(), is there any chance you can call your new function from there, or do they do different things?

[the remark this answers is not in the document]
prakashcv13: ri_degraded_time is used as the soonest time in the reservation window when the reservation becomes degraded. For a standing reservation, it could be that an instance after the first is degraded. Using ri_degraded_time, reserve_retry_cutoff and reserve_retry_init, we find the retry time.
when a node goes down, vnode_unavailable() is called. In that function we look for all the reservations associated with the vnode that went down, set the degraded time and calculate the retry time somewhere between now and the degraded time.
Over here, we know the reservation as well as the node that went down, we need not "find" the reservation and also not go through the logic of finding things which we already know of.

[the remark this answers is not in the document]
bhroam: I wasn't thinking a wholesale replacement of vnode_unavailable().  I was thinking that this function handles what should happen if a reservation goes degraded.  So, do what you need to do in vnode_unavailable() to find the reservations, and then call your new function.  I guess I'm not seeing something.  If it doesn't make sense, then don't do it.  I was just hoping for some code reuse.

src/server/node_manager.c
prakashcv13: this is copied as-is from vnode_unavailable(), we can remove it. could you please let me know if you would like me to remove it from there as well.

src/server/node_manager.c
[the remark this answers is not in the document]
prakashcv13: vnode_unavailable() relies on what is there in ri_pbsnode_list to find the reservations. We set ri_pbsnode_list in set_nodes() once everything goes fine.
vnode_unavailable() is called when a node goes down. So, it was called when the scheduling cycle was in progress and the node went down. But, at that time "this" reservation was not updated with "this" node. Even if we call vnode_unavailable() from here, it will not help.

src/server/node_manager.c
prakashcv13: I am unable to find the chunk where we check if the node is (down, stale or offline) for a job in this function.

bhroam: Oh you are right.  I added it, then removed it due to all the problems it was causing.  The code worked for both jobs and reservations.  Reservations were the problematic code (which you've correctly handled).  We should probably submit a ticket to handle jobs in this fashion too.  If a mom is down, we still try to send the job to the mom even though we know it is down.

src/server/node_manager.c
bhroam: At least make them both DEBUG3 messages.

src/server/node_manager.c
prakashcv13: I am of the view that we need not handle the job case. The node can go down right after we check if it is down. So, it is better that we fail at the time of sending the job to the mom, which I think is very soon after we assign nodes to the job.
Also, if we reject the request on finding the node is down, the scheduler is not going to look for alternate nodes. What do you think?

prakashcv13:
@bhroam - I have updated the log level and moved the trailing "*/" to the next line.

src/server/node_manager.c
bhroam: If we know the node is down, why should we go through the extra work to try and send it to the mom just to fail?  We know it will fail.  If the node goes down after set_nodes(), we'll still catch it like we do now.  As for the scheduler finding a new node, this will only happen in the case where throughput mode is false.

prakashcv13:
thank you @bhroam

src/server/node_manager.c
```diff
@@ -7130,10 +7130,13 @@ set_nodes(void *pobj, int objtype, char *execvnod_in, char **execvnod_out, char
 				return (PBSE_UNKNODE);
 			}
 
+			if ((pnode->nd_state & (INUSE_DOWN | INUSE_STALE | INUSE_OFFLINE)) && (svr_init == FALSE))
```
riyazhakki: How about checking pnode->nd_state flag with VNODE_UNAVAILABLE ?

PR #1307, opened 2019-09-20
[WIP] Start time window

prakashcv13:


src/scheduler/fifo.c
```diff
@@ -917,28 +917,38 @@ main_sched_loop(status *policy, int sd, server_info *sinfo, schd_error **rerr)
...
-						sort_again = MUST_RESORT_JOBS;
-					else
-						sort_again = MAY_RESORT_JOBS;
-				} else {
-					/* if run_update_resresv() returns 0 and pbs_errno == PBSE_HOOKERROR,
-					 * then this job is required to be ignored in this scheduling cycle
-					 */
-					rc = err->error_code;
+				if (!(tj->job->window_enabled)) {
+					rc = JOB_WINDOW_NOT_STARTED;
```
bhroam: This is the wrong place for checking the time window.  Put the check for the time window in is_ok_to_run().

prakashcv13: hey @bhroam - thank you for looking into this.
If we check in is_ok_to_run() and return NULL, we fall into the else part and try preemption. We need another check there. Another reason is that we need to set rc = JOB_WINDOW_NOT_STARTED, only if the job did not run because it's time window has not started, not for some other reason, so I added this check just before we send the job to the server.

bhroam: The check still belongs in is_ok_to_run().  The error code still comes back in the err structure.  Another reason this has to be in is_ok_to_run() is because of calendaring.  We need to be able to know when the job can run via is_ok_to_run().  Otherwise we'll add it on the calendar in places it can't run since it is out of the time window.  If you are worried about preemption, there is a section of code there that checks a set of static reasons that we bail out of the preemption code.  For instance if do_not_span_psets is on and we need to span.  Preemption can't help there.  Just add a check for being out of your time window there.

prakashcv13: done. there is some code refactoring needed still...

drtoss:
I have been testing this for a bit and have a suggestion that will remove some of the special case code in qsub.c and parse_equal.c. The problem is with the window_rrule option, which has the form "FREQ=BYDAY=MO,TU". The argument parser that qsub uses (parse_equal) is not happy with the option -W window_rrule=FREQ=BYDAY=MO,TU. There are too many "=" for a single option. The current code has special cases to deal with this. I suggest instead that you just require the value to be quoted: `-W window_rrule="FREQ=BYDAY=MO,TU"`. This hides the extra equals from parse_equal.

In addition, the current code goes into an infinite loop if passed 

 `qsub -W 'window_rrule="FREQ=WEEKLY;BYDAY=TU,"' ... # Note trailing ,`

The attached diff implements these changes, in addition to fixing a bug in parse_equal that ended up adding a space to a quoted string value.

It also includes the iCal error number in messages if the server has trouble parsing the rrule. This helps debugging a tiny bit.

[thediffs.txt](https://github.com/PBSPro/pbspro/files/4459213/thediffs.txt)

drtoss:
Currently, the window_duration option is validated by verify_value_window_start_duration(). This means the duration must be 4 digits in the form hhmm. Would it be better to allow the same forms as for pbs_rsub?

prakashcv13:
hi @drtoss - Thank you for testing the time window feature, providing feedback and also fixing the issue that you found. The qsub changes had a hack, and I needed to work on a better solution before we merge this PR, but maybe I would not need to that now. 
I will try out and test the fix.
Coming to the format for providing time windows, I agree for consistency, we can use the same format as pbs_rsub, however, that would need refactoring of the current implementation.

riyazhakki:
Hi @drtoss , I am addressing your comments on this PR.

I have rebased the branch to the latest master and applied patch for the diff you shared, and it has fixed an infinite loop issue had with a trailing comma.

For the window rrule quoting suggestion you menioned -W window_rrule="FREQ=BYDAY=MO,TU" I believe it is a typo having "FREQ=" without its value(WEEKLY|DAILY|HOURLY), I have now handled this invalid syntax in the code and crrently we are reading only BYDAY element of the rrule, regarding quoting the value of window_rrule we need to add additional single quote around the value to be quoted, like -W window_rrule='"BYDAY=MO,TU"' because the command line arguments strips the double quotes from its argument.

riyazhakki:
> Currently, the window_duration option is validated by verify_value_window_start_duration(). This means the duration must be 4 digits in the form hhmm. Would it be better to allow the same forms as for pbs_rsub?

Hi @drtoss,
I have addressed this in the latest commit, attaching test logs here.
[window_duration_test.txt](https://github.com/openpbs/openpbs/files/4681065/window_duration_test.txt)

subhasisb:
hi @prakashcv13 - this PR has been open for too long and has not seen any activity recently. Should we not close this PR for now, and reopen when work resumes?

PR #1309, opened 2019-09-20
tests for hook events phase 1.2

vccardenas:
#### Describe Bug or Feature
Tests that cover linux hook event phase 1.2.

#### Describe Your Change
Added new tests involving execjob_prologue and execjob_attach.

#### Attach Test Logs/Output
[TestSupportLinuxHookEventPhase1_2_py3.txt](https://github.com/PBSPro/pbspro/files/3647493/TestSupportLinuxHookEventPhase1_2_py3.txt)


test/tests/functional/pbs_support_linux_cgroup_phase1_2.py
```diff
@@ -0,0 +1,229 @@
...
+        self.server.create_import_hook("hook1", a, hook_body)
+
+        self.momA = self.moms.values()[0]
+        self.momB = self.moms.values()[1]
+        self.hostA = self.momA.shortname
+        self.hostB = self.momB.shortname
+        if self.momA.is_cpuset_mom():
+            self.hostA = self.hostA + '[0]'
+        if self.momB.is_cpuset_mom():
+            self.hostB = self.hostB + '[0]'
```
mike0042: The filename contains "cgroup" and you are testing whether the cpuset MoM is being used. The cgroup hook should be disabled when the cpuset MoM is running. It doesn't look like you're testing the cgroup hook at all here, so the filename is misleading.  Rather, you are testing hook event presence and order. Perhaps we should consider a different filename.

test/tests/functional/pbs_support_linux_cgroup_phase1_2.py
```diff
@@ -0,0 +1,229 @@
...
+        pbs_attach = self.server.pbs_conf['PBS_EXEC'] + "/bin/pbs_attach"
+        pbs_tmrsh = self.server.pbs_conf['PBS_EXEC'] + "/bin/pbs_tmrsh"
```
mike0042: Please use os.path.join()

test/tests/functional/pbs_support_linux_hook_event_phase1_2.py
```diff
@@ -0,0 +1,229 @@
...
+# "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
+# trademark licensing policies.
+
+from tests.functional import *
+
+
+@requirements(num_moms=2)
+class TestSupportLinuxHookEventPhase1_2(TestFunctional):
+    """
+    Tests that cover support for Linux cgroups in phase 1.2.
```
mike0042: This comment still references cgroups.

PR #1311, opened 2019-09-24
Use find_node_by_indrank() when possible

bhroam:
#### Describe Bug or Feature
When node buckets were introduced, some find node operations changed from O(N) to O(1).  This new method was used with the node bucket algorithm, but the normal node algorithm was not modified to use it.

#### Describe Your Change
Replace find_node_by_rank() to find_node_by_indrank() whenever possible.  

[ptl.before.log](https://github.com/PBSPro/pbspro/files/3649421/ptl.before.log)
[ptl.after.log](https://github.com/PBSPro/pbspro/files/3649420/ptl.after.log)

### DIFF
--- a/src/scheduler/node_info.c
+++ b/src/scheduler/node_info.c
@@ -1371,7 +1371,7 @@ dup_node_info(node_info *onode, server_info *nsinfo,
 	nnode->last_used_time = onode->last_used_time;
 
 	if (onode->svr_node != NULL)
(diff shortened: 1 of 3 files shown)


prakashcv13:
i have restarted the failed travis builds

bhroam:
@agrawalravi90 I've run the performance test and added the output to the description.

PR #1313, opened 2019-09-25
Add log_eventf() and refactor scheduler to use it

bhroam:
#### Describe Bug or Feature
Implementation of log_eventf() (a combination of log_event() and sprintf()), and refactoring the scheduler to use it.  See the design document for more details of the function.

#### Link to Design Doc
https://pbspro.atlassian.net/wiki/spaces/PD/pages/1351319574/Scheduler+s+log+events

I ran a full regression and no tests failed.

### DIFF
--- a/src/lib/Liblog/log_event.c
+++ b/src/lib/Liblog/log_event.c
@@ -56,6 +56,7 @@
 #include <time.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <stdarg.h>
 #include "log.h"
 #include "server_limits.h"
 #include "list_link.h"
@@ -74,6 +75,8 @@ PBSEVENT_RESV;
 
 extern char *path_home;
 long	    *log_event_mask = &log_event_lvl_priv;
+
+extern char *pbs_asprintf_format(int len, const char *fmt, va_list args);
 
 /**
  * @brief
@@ -121,4 +124,49 @@ log_event(int eventtype, int objclass, int sev, const char *objname, const char 
 {
 	if (will_log_event(eventtype))
 		log_record(eventtype, objclass, sev, objname, text);
+}
+
+/**
+ * @brief
+ * 	log_eventf - a combination of log_event() and printf()
+ * 
+ * @param[in] eventtype - event type
+ * @param[in] objclass - event object class 
+ * @param[in] sev - indication for whether to syslogging enabled or not
+ * @param[in] objname - object name stating log msg related to which object
+ * @param[in] fmt - format string
+ * @param[in] ... - arguments to format string
+ * 
+ * @return void
+ */
+void
+log_eventf(int eventtype, int objclass, int sev, const char *objname, const char *fmt, ...)
+{
+	va_list args;
+	int len;
+	char logbuf[LOG_BUF_SIZE];
+	char *buf;
+	
+	if (will_log_event(eventtype) == 0)
+		return;
+
+	va_start(args, fmt);
+
+	len = vsnprintf(logbuf, sizeof(logbuf), fmt, args);
+
+	if (len >= sizeof(logbuf)) {
+		buf = pbs_asprintf_format(len, fmt, args);
+		if (buf == NULL) {
+			va_end(args);
+			return;
+		}
+	}
+	else
+		buf = logbuf;
+	
+	log_record(eventtype, objclass, sev, objname, buf);
+	
+	if (len >= sizeof(logbuf))
+		free(buf);
+	va_end(args);
 }
--- a/src/lib/Libutil/misc_utils.c
+++ b/src/lib/Libutil/misc_utils.c
@@ -427,6 +427,63 @@ pbs_fgets_extend(char **pbuf, int *pbuf_size, FILE *fp)
 
 /**
  * @brief
+ * 	Internal helper function for pbs_asprintf() to determine the length of post-formatted string
+ * 
+ * @param[in] fmt - printf format string
+ * @param[in] args - va_list arguments from pbs_asprintf()
+ * 
+ * @return int
+ * @retval length of post-formatted string
+ */
+int
+pbs_asprintf_len(const char *fmt, va_list args) 
+{
+	int len;
+#ifdef WIN32
+	len = _vscprintf(fmt, args);
+#else
+	{
+		va_list dupargs;
+		char c;
+
+		va_copy(dupargs, args);
+		len = vsnprintf(&c, 0, fmt, dupargs);
+		va_end(dupargs);
+	}
+#endif
+	return len;
+}
+
+/**
+ * @brief
+ * 	Internal helper function for pbs_asprintf() to allocate memory and format the string
+ * 
+ * @param[in] len - length of post-formatted string
+ * @param[in] fmt - format for printed string
+ * @param[in] args - va_list arguments from pbs_asprintf()
+ * 
+ * @return char *
+ * @retval formatted string in allocated buffer
+ */
+
+char *
+pbs_asprintf_format(int len, const char *fmt, va_list args)
+{
+	char *buf;
+	int rc;
+	buf = malloc(len + 1);
+	if (!buf)
+		return NULL;
+	rc = vsnprintf(buf, len + 1, fmt, args);
+	if (rc != len) {
+		free(buf);
+		return NULL;
+	}
+	return buf;
+}
+
+/**
+ * @brief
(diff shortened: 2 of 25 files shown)


src/scheduler/node_info.c
```diff
@@ -4576,8 +4513,8 @@ parse_execvnode(char *execvnode, server_info *sinfo)
 
 	if (invalid) {
 		snprintf(logbuf, sizeof(logbuf), "Failed to parse execvnode: %s", execvnode);
```
arungrover: This snprintf is not needed

src/scheduler/queue_info.c
```diff
@@ -153,7 +153,8 @@ query_queues(status *policy, int pbs_sd, server_info *sinfo)
 		if (errmsg == NULL)
 			errmsg = "";
 		sprintf(log_buffer, "Statque failed: %s (%d)", errmsg, pbs_errno);
```
arungrover: This sprintf is not needed

src/scheduler/resource.c
```diff
@@ -132,16 +132,16 @@ query_resources(int pbs_sd)
 			errmsg = "";
 
 		sprintf(logbuf, "pbs_statrsc failed: %s (%d)", errmsg, pbs_errno);
```
arungrover: same comment here

src/scheduler/resv_info.c
```diff
@@ -1535,8 +1524,8 @@ confirm_reservation(status *policy, int pbs_sd, resource_resv *unconf_resv, serv
 			short_xc = string_dup(execvnodes);
 
 		sprintf(logmsg, "Confirming %d Occurrences", occr_count);
```
arungrover: same comment here

src/scheduler/site_code.c
```diff
@@ -3258,7 +3258,7 @@ reconcile_share_tree(share_info *root, share_info *def, int cnt)
 						"%s share for %s too small for children: %d < %d",
 						root->name, shr_class_name_by_idx(i),
 						gross, c_sum);
-					schdlog(PBSEVENT_SCHED, PBS_EVENTCLASS_FILE,
+					log_event(PBSEVENT_SCHED, PBS_EVENTCLASS_FILE,
```
arungrover: Can this be log_eventf and we can get rid of sprintf as well?

src/scheduler/site_code.c
```diff
@@ -3723,7 +3723,7 @@ check_for_cycle_interrupt(int do_logging)
 		sprintf(log_buffer, "Too early to short circuit (%ds elapsed, need %ds)",
 			(int)(time(NULL) - interrupted_cycle_start_time),
 			conf.min_intrptd_cycle_length);
-		schdlog(PBSEVENT_DEBUG2, PBS_EVENTCLASS_SERVER, LOG_DEBUG, __func__, log_buffer);
+		log_event(PBSEVENT_DEBUG2, PBS_EVENTCLASS_SERVER, LOG_DEBUG, __func__, log_buffer);
```
arungrover: Can this be log_eventf and we can get rid of sprintf as well?

src/scheduler/site_code.c
bhroam: This is NAS site code.  I chose to not convert it.  I am not going to test it, so it is probably best to not convert it.  I had to change the function name since schdlog() no longer exists.

src/scheduler/site_code.c
bhroam: Same as above, this is NAS code.  They can make the modification if they wish.

src/lib/Liblog/log_event.c
```diff
@@ -122,3 +125,48 @@ log_event(int eventtype, int objclass, int sev, const char *objname, const char
...
+
+	va_start(args, fmt);
+
+	len = vsnprintf(logbuf, sizeof(logbuf), fmt, args);
+
+	if (len >= sizeof(logbuf)) {
+		buf = pbs_asprintf_format(len, fmt, args);
+		if (buf == NULL) {
+			va_end(args);
+			return;
```
agrawalravi90: shouldn't we log something when this fails? Since this function returns a void, the user won't really know if anything is wrong.

src/lib/Libutil/misc_utils.c
```diff
@@ -425,6 +425,63 @@ pbs_fgets_extend(char **pbuf, int *pbuf_size, FILE *fp)
...
+ * @param[in] len - length of post-formatted string
+ * @param[in] fmt - format for printed string
+ * @param[in] args - va_list arguments from pbs_asprintf()
+ * 
+ * @return char *
+ * @retval formatted string in allocated buffer
+ */
+
+char *
+pbs_asprintf_format(int len, const char *fmt, va_list args)
```
agrawalravi90: i'm not sure I like the name of this function. pbs_asprintf() also takes a format argument, so 'format' doesn't seem like the right distinction for this function. I actually think that pbs_asprintf_len() would be a good name for it since it needs the length as input. How about renaming the current "pbs_asprintf_len" to "get_str_va_len" and renaming "pbs_asprintf_format" to "pbs_asprintf_len" instead?

src/scheduler/fifo.c
```diff
@@ -511,15 +504,13 @@ init_scheduling_cycle(status *policy, int pbs_sd, server_info *sinfo)
 				if (sinfo->job_formula != NULL) {
 					double threshold = policy->job_form_threshold;
 					resresv->job->formula_value = formula_evaluate(sinfo->job_formula, resresv, resresv->resreq);
-					sprintf(log_buffer, "Formula Evaluation = %.*f",
-						float_digits(resresv->job->formula_value, FLOAT_NUM_DIGITS), resresv->job->formula_value);
-					schdlog(PBSEVENT_DEBUG3, PBS_EVENTCLASS_JOB, LOG_DEBUG, resresv->name, log_buffer);
+					log_eventf(PBSEVENT_DEBUG2, PBS_EVENTCLASS_JOB, LOG_DEBUG, resresv->name, "Formula Evaluation = %.*f",
```
agrawalravi90: This was a 'PBSEVENT_DEBUG3' right? Is this change intentional?

src/lib/Libutil/misc_utils.c
bhroam: I disagree.  The reason I named the functions the way I did was because of their purpose, not what they take.  pbs_asprintf_len() determines the length of the string.  It shouldn't be based on the fact that it takes a length parameter.  pbs_asprintf_format() actually does the formatting of the string.

src/scheduler/fifo.c
bhroam: Yes, it shouldn't have been DEBUG3 to begin with.  Per-job messages are at DEBUG2.  Per-job per-node messages are at DEBUG3.

src/lib/Liblog/log_event.c
bhroam: To what end?  The message would be very generic and not really helpful.  If you want a log message, I can put it in pbs_asprintf_format() when the malloc() fails, if you want.  Here all we'd log is 'log_eventf() failure' or something like that.  It wouldn't really be helpful.  We failed to format the string, so we would have no idea what log message had an error.

src/lib/Libutil/misc_utils.c
agrawalravi90: I personally got a bit confused because they are all "public" functions, so when one is choosing one of them, it didn't seem obvious to me which one I should be calling. But maybe it's just me, so I'll leave it up to you, naming is hard!

src/lib/Liblog/log_event.c
agrawalravi90: The effect of this failing is that some log messages would potentially not get printed right? I think we should add some hint in the logs that this happened as the absence of log messages might get construed as "things look fine". So even a generic log message might help.

agrawalravi90: Actually I'm ok with not doing this if you feel this is a 1 in a million scenario, so up to you

bhroam: The only failure cases are malloc() failure and if vsnprintf() returns an error.  Today neither are checked for.  This error is going to be very rare, so I don't think we should bother.

agrawalravi90:
@bhroam unfortunately, there are conflicts, please resolve them so this can be merged, thanks.

bhroam:
Thanks for your sign-off @agrawalravi90.  I have resolved the conflicts.

PR #1315, opened 2019-09-26
In PTL, the check for arch is not happening in is_cpuset_mom() as the variable is being overwritten

npadole20:
#### Describe Bug or Feature
In is_cpuset_mom(), we are overwriting the variable which will be used to check arch of the node after the node is free. Hence we end up checking if the node is free twice and do not check the arch of the node. Found through code walkthrough.

#### Describe Your Change
Rearranged the code to ensure check for arch happens

[test_log_is_cpuset_mom.txt](https://github.com/PBSPro/pbspro/files/3656201/test_log_is_cpuset_mom.txt)

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -13108,13 +13108,13 @@ class MoM(PBSService):
         """
         if self._is_cpuset_mom is not None:
             return self._is_cpuset_mom
-        raa = ATTR_rescavail + '.arch'
-        a = {raa: None}
+        a = {'state': 'free'}
         try:
-            a = {'state': 'free'}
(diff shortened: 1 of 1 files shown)


agrawalravi90:
@npadole20 Please add 'PTL' to the title of the PR and the commit message, I wasn't sure if this is a PBSPro bug or a PTL bug until I saw the code, we can merge this after that, thanks.

prakashcv13:
@agrawalravi90 - is it a process to mention 'PTL'/'PBS' in the title?

agrawalravi90:
> @agrawalravi90 - is it a process to mention 'PTL'/'PBS' in the title?

No, but it does help to know whether a particular issue is a product issue or a test issue. It's usually pretty obvious which is probably why we don't have a process around it, but in this case it wasn't.

PR #1318, opened 2019-09-26
Reporting hardware status

birbalsain:
#### Describe Bug or Feature
Ptl framework doesn't provide the information about generated core files and Ptl test doesn't stop if used disk size is greater than 95% .

#### Describe Your Change
It provides the information, where core files are generated. Ptl test stop if used disk size reaches 95 %. It checks hardware and core file information in every 5 minutes. 

#### Link to Design Doc
https://pbspro.atlassian.net/wiki/spaces/PD/pages/1254686727/Test+framework+should+sense+indicate+or+stop+tests+before+system+breakdown+during+a+test+run

[disk_limit_exceed_error.txt](https://github.com/PBSPro/pbspro/files/3711639/disk_limit_exceed_error.txt)
[hardware_report_test_success.txt](https://github.com/PBSPro/pbspro/files/3711643/hardware_report_test_success.txt)
[hardware_report_warning_message.txt](https://github.com/PBSPro/pbspro/files/3711644/hardware_report_warning_message.txt)
[disk_cleanup_message.txt](https://github.com/PBSPro/pbspro/files/3712145/disk_cleanup_message.txt)

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -503,7 +503,7 @@ class PbsResvAlterError(PtlException):
     pass
 
 
-class PbsSystemError(PtlException):
+class PbsHardwareError(PtlException):
     pass
 
 
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -503,7 +503,7 @@ class PbsResvAlterError(PtlException):
     pass
 
 
-class PbsHardwareError(PtlException):
+class PbsHardwareMonitorError(PtlException):
     pass
 
 
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -503,6 +503,10 @@ class PbsResvAlterError(PtlException):
     pass
 
 
+class PbsSystemError(PtlException):
+    pass
+
+
 class PbsTypeSize(str):
 
     """
--- a/test/fw/ptl/utils/plugins/ptl_test_runner.py
+++ b/test/fw/ptl/utils/plugins/ptl_test_runner.py
@@ -763,7 +763,7 @@ class PTLTestRunner(Plugin):
         """
         du = DshUtils()
         self.hardware_report_timer = Timer(
-            300, self.checks_hardware_status_and_core_files)
+            300, self.check_hardware_status_and_core_files)
         self.hardware_report_timer.start()
         systems = list(self.param_dict['servers'])
         systems.extend(self.param_dict['moms'])
@@ -781,12 +781,12 @@ class PTLTestRunner(Plugin):
                 self.logger.error(_msg)
                 self.hardware_report_timer.cancel()
                 raise PbsHardwareMonitorError(msg=_msg)
-            elif 70 < used_disk_percent < 95:
-                _msg = hostname + ": used disk is "
+            elif 70 <= used_disk_percent < 95:
+                _msg = hostname + ": disk usage is at "
                 _msg += str(used_disk_percent) + "%"
                 _msg += ", disk cleanup is recommended."
                 self.logger.warning(_msg)
-            elif used_disk_percent > 95:
+            elif used_disk_percent >= 95:
                 _msg = hostname + ":disk usage > 95%, stopping the test(s)"
                 self.logger.error(_msg)
                 self.hardware_report_timer.cancel()
@@ -794,38 +794,46 @@ class PTLTestRunner(Plugin):
             # checks for core files
             pbs_conf = du.parse_pbs_config(hostname)
             mom_priv_path = os.path.join(pbs_conf["PBS_HOME"], "mom_priv")
-            if du.isdir(hostname, mom_priv_path):
-                mom_priv_files = du.listdir(hostname, mom_priv_path,
-                                            sudo=True, fullpath=False)
+            if du.isdir(hostname=hostname, path=mom_priv_path):
+                mom_priv_files = du.listdir(
+                    hostname=hostname,
+                    path=mom_priv_path,
+                    sudo=True,
+                    fullpath=False)
                 if fnmatch.filter(mom_priv_files, "core*"):
                     _msg = hostname + ": core files found in "
                     _msg += mom_priv_path
                     self.logger.warning(_msg)
             server_priv_path = os.path.join(
                 pbs_conf["PBS_HOME"], "server_priv")
-            if du.isdir(hostname, server_priv_path):
-                server_priv_files = du.listdir(hostname, server_priv_path,
-                                               sudo=True, fullpath=False)
+            if du.isdir(hostname=hostname, path=server_priv_path):
+                server_priv_files = du.listdir(
+                    hostname=hostname,
+                    path=server_priv_path,
+                    sudo=True,
+                    fullpath=False)
                 if fnmatch.filter(server_priv_files, "core*"):
                     _msg = hostname + ": core files found in "
                     _msg += server_priv_path
                     self.logger.warning(_msg)
             sched_priv_path = os.path.join(pbs_conf["PBS_HOME"], "sched_priv")
-            if du.isdir(hostname, sched_priv_path):
-                sched_priv_files = du.listdir(hostname, sched_priv_path,
-                                              sudo=True, fullpath=False)
+            if du.isdir(hostname=hostname, path=sched_priv_path):
+                sched_priv_files = du.listdir(
+                    hostname=hostname,
+                    path=sched_priv_path,
+                    sudo=True,
+                    fullpath=False)
                 if fnmatch.filter(sched_priv_files, "core*"):
                     _msg = hostname + ": core files found in "
                     _msg += sched_priv_path
                     self.logger.warning(_msg)
             for u in PBS_ALL_USERS:
-                user_home_files = du.listdir(hostname, u.home,
+                user_home_files = du.listdir(hostname=hostname, path=u.home,
                                              sudo=True, fullpath=False)
-                for filename in user_home_files:
-                    if filename.startswith("core"):
-                        _msg = hostname + ": user-" + u
-                        _msg += ": core files found in "
-                        self.logger.warning(_msg + user_home['out'])
+                if fnmatch.filter(user_home_files, "core*"):
+                    _msg = hostname + ": user-" + str(u)
+                    _msg += ": core files found in "
+                    self.logger.warning(_msg + u.home)
 
     def startTest(self, test):
         """
@@ -856,7 +864,7 @@ class PTLTestRunner(Plugin):
             self.result.startTest(test)
             raise SkipTest('Test requirements are not matching')
         # function report hardware status and core files
-        self.checks_hardware_status_and_core_files()
+        self.check_hardware_status_and_core_files()
 
         def timeout_handler(signum, frame):
             raise TimeOut('Timed out after %s second' % timeout)
--- a/test/fw/ptl/utils/plugins/ptl_test_runner.py
+++ b/test/fw/ptl/utils/plugins/ptl_test_runner.py
@@ -46,6 +46,7 @@ import socket
 import sys
 import tempfile
 import unittest
+from threading import Timer
 from logging import StreamHandler
 from traceback import format_exception
 from types import ModuleType
@@ -58,6 +59,7 @@ from nose.util import isclass
 
 import ptl
 from ptl.lib.pbs_testlib import PBSInitServices
+from ptl.lib.pbs_testlib import PbsSystemError
 from ptl.utils.pbs_covutils import LcovUtils
 from ptl.utils.pbs_dshutils import DshUtils
 from ptl.utils.pbs_testsuite import (MINIMUM_TESTCASE_TIMEOUT,
@@ -428,13 +430,15 @@ class SystemInfo:
             self.logger.error(_msg + hostname)
         else:
             for i in mem_info['out']:
-                if "MemAvailable" in i:
+                if "MemTotal" in i:
+                    self.system_total_ram = float(i.split()[1]) / (2**20)
+                elif "MemAvailable" in i:
                     self.system_ram = float(i.split()[1]) / (2**20)
                     break
         # getting disk size in gb
         pbs_conf = du.parse_pbs_config(hostname)
         pbs_home_info = du.run_cmd(hostname, cmd=['df', '-k',
-                                   pbs_conf['PBS_HOME']])
+                                                  pbs_conf['PBS_HOME']])
         if pbs_home_info['rc'] != 0:
             _msg = 'failed to get output of df -k command of host: '
             self.logger.error(_msg + hostname)
@@ -442,6 +446,7 @@ class SystemInfo:
             disk_info = pbs_home_info['out']
             disk_size = disk_info[1].split()
             self.system_disk = float(disk_size[3]) / (2**20)
+            self.system_disk_used_percent = float(disk_size[4].rstrip('%'))
 
 
 class PtlTextTestRunner(TextTestRunner):
@@ -600,10 +605,10 @@ class PTLTestRunner(Plugin):
                                    {}).get('default-testcase-timeout',
                                            MINIMUM_TESTCASE_TIMEOUT))
         tc_timeout = int(getattr(getattr(_test,
-                                 getattr(_test, '_testMethodName', ''),
-                                 None),
-                         TIMEOUT_KEY,
-                         0))
+                                         getattr(_test, '_testMethodName', ''),
+                                         None),
+                                 TIMEOUT_KEY,
+                                 0))
         return max([dflt_timeout, tc_timeout])
 
     def __set_test_end_data(self, test, err=None):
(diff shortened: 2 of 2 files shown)


birbalsain:
@hirenvadalia please review my PR.

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -744,10 +749,221 @@ def __are_requirements_matching(param_dic=None, test=None):
...
+        self.timer_start = True
+        # mom hardware report
+        for hostname in self.param_dict['moms']:
+            hr = SystemInfo()
+            hr.get_system_info(hostname)
+            # monitors ram
+            avail_ram = getattr(hr, 'system_ram', None)
+            total_ram = getattr(hr, 'system_total_ram', None)
+            if (avail_ram or total_ram) is None:
+                _msg = "hostname " + hostname
```
hirenvadalia: please change this to `hostname + ": "` (same for other places in change)

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -500,6 +500,10 @@ class PbsResvAlterError(PtlException):
     pass
 
 
+class PbsSystemError(PtlException):
```
hirenvadalia: I suggest renaming this to `PbsHardwareError` or `PbsRequirementsError` - generally in python SystemError is used when some internal error occurs but here the case is different

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -744,10 +749,221 @@ def __are_requirements_matching(param_dic=None, test=None):
...
+                self.logger.error(_msg + "unable to get ram info")
+                self.hardware_report_timer.cancel()
```
hirenvadalia: after canceling timer make `self.timer_start = False`

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -744,10 +749,221 @@ def __are_requirements_matching(param_dic=None, test=None):
...
+                                        mom_priv_path, sudo=True,
+                                        fullpath=False)
+            for filename in mom_priv_files:
+                if filename.startswith("core"):
+                    _msg = "hostname- " + hostname + ": core file exist at "
+                    self.logger.warning(_msg + mom_priv_path)
+                    self.hardware_report_timer.cancel()
+                    raise PbsSystemError(msg=_msg + mom_priv_path)
+            # checks core files in server_priv
+            server_priv_path = os.path.join(pbs_conf["PBS_HOME"],
```
hirenvadalia: This will fail if you have only an execution installation... as server_priv won't exist in execution-only install

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -744,10 +749,221 @@ def __are_requirements_matching(param_dic=None, test=None):
...
+                                           server_priv_path, sudo=True,
+                                           fullpath=False)
+            for filename in server_priv_files:
+                if filename.startswith("core"):
+                    _msg = "hostname- " + hostname + ": core file exist at "
+                    self.logger.warning(_msg + server_priv_path)
+                    self.hardware_report_timer.cancel()
+                    raise PbsSystemError(msg=_msg + server_priv_path)
+            # checks core files in sched_priv
+            sched_priv_path = os.path.join(pbs_conf["PBS_HOME"],
```
hirenvadalia: same as above

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -744,10 +749,221 @@ def __are_requirements_matching(param_dic=None, test=None):
...
+                                          sched_priv_path, sudo=True,
+                                          fullpath=False)
+            for filename in sched_priv_files:
+                if filename.startswith("core"):
+                    _msg = "hostname- " + hostname + ": core file exist at "
+                    self.logger.warning(_msg + sched_priv_path)
+                    self.hardware_report_timer.cancel()
+                    raise PbsSystemError(msg=_msg + sched_priv_path)
+            # checks core files in user home directory
+            ptl_users = ['pbsuser1', 'pbsuser2', 'pbsuser3', 'pbsuser4',
```
hirenvadalia: Make user of PBS_ALL_USERS from pbs_testusers.py

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -744,10 +749,221 @@ def __are_requirements_matching(param_dic=None, test=None):
...
+                         'pbsuser5', 'pbsuser6', 'pbsuser7', 'pbstest',
+                         'pbsroot', 'pbsadmin']
+            for user in ptl_users:
+                cmd = "getent passwd " + user
```
hirenvadalia: once you use PBS_ALL_USERS you don't need to do this as PbsUser() has `.home` an attribute which will give you home dir path for that user

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -744,10 +749,221 @@ def __are_requirements_matching(param_dic=None, test=None):
...
+                    for filename in user_home_files:
+                        if filename.startswith("core"):
+                            _msg = "hostname- " + hostname
+                            _msg = ", user-" + user
+                            _msg += ": core file exist at"
+                            self.logger.warning(_msg + user_home['out'])
+                            self.hardware_report_timer.cancel()
+                            raise PbsSystemError(msg=_msg+user_home['out'])
+        # server hardware report
+        for hostname in self.param_dict['moms']:
```
hirenvadalia: looks like for block is duplicate of above...

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -744,10 +749,221 @@ def __are_requirements_matching(param_dic=None, test=None):
...
+        """
+        function used to get hardware report
+        in every 5 minutes
+        """
+        du = DshUtils()
+        self.hardware_report_timer = Timer(300, self.hardware_report)
+        self.hardware_report_timer.start()
+        self.timer_start = True
+        # mom hardware report
+        for hostname in self.param_dict['moms']:
```
hirenvadalia: also why only moms? shouldn't we do for servers, scheds, and comms also?

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -788,6 +1006,8 @@ def stopTest(self, test):
         if old_sigalrm_handler is not None:
             signal.signal(signal.SIGALRM, old_sigalrm_handler)
             signal.alarm(0)
+        if self.timer_start:
```
hirenvadalia: I suggest moving this if condition in `self.hardware_report()`, then you can get rid of `self.timer_start`

birbalsain: I have to stop timer at the time of test end. In above I am checking, timer is started then true and timer cancel at test end otherwise no need to take any action.
According to you i have to monitor test end and stop timer. That is reverse of above.

hirenvadalia: To stop the timer at test end you can call self.hardware_report_timer.cancle().

birbalsain: In stopTest i am stopping timer if self.timer_start(timer is started). If i don't use timer monitoring in stopTest function then timer continue to run even the test end successfully.
if i use if condition in hardware_report function then i have to monitor when test is going to end. so it is reverse of the above.
anyhow i have to use one variable which can monitor timer_start/test_end also one if block.

test/fw/ptl/utils/plugins/ptl_test_runner.py
birbalsain: I also did for sever. is scheduler also a different machine.

hirenvadalia: I don't see use of self.param_dict['servers'] in this method? can you please tell me line no. where you used it?

test/fw/ptl/utils/plugins/ptl_test_runner.py
hirenvadalia: I understand the use case so you call self.hardware_report_timer.cancel() then the timer will be stopped forcefully so I don't think we need self.timer_start.

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -744,10 +749,300 @@ def __are_requirements_matching(param_dic=None, test=None):
...
+            # checks core files in mom_priv
+            pbs_conf = du.parse_pbs_config(hostname)
+            mom_priv_path = os.path.join(pbs_conf["PBS_HOME"],
+                                         "mom_priv")
+            mom_priv_files = du.listdir(hostname,
+                                        mom_priv_path, sudo=True,
+                                        fullpath=False)
+            for filename in mom_priv_files:
+                if filename.startswith("core"):
+                    _msg = hostname + ": core file exist at "
```
prakashcv13: shall we use "core file found in" instead of "core file exist at"

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -744,10 +749,300 @@ def __are_requirements_matching(param_dic=None, test=None):
...
+                    self.logger.warning(_msg + mom_priv_path)
+                    self.hardware_report_timer.cancel()
+                    raise PbsHardwareError(msg=_msg + mom_priv_path)
+            # checks core files in user home directory
```
prakashcv13: checks for core files ...

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -744,10 +749,300 @@ def __are_requirements_matching(param_dic=None, test=None):
...
+                         'pbsroot', 'pbsadmin']
+            for user in ptl_users:
+                cmd = "getent passwd " + user
+                user_home = du.run_cmd(hostname, cmd=cmd)
+                if user_home['rc'] != 0:
+                    _msg = hostname + ", user-" + user
+                    self.logger.error(_msg + ": unable to get user home")
+                    self.hardware_report_timer.cancel()
+                    raise PbsHardwareError(
+                        msg=_msg+": unable to get user home")
```
prakashcv13: this cannot be hardware error.

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -744,10 +749,300 @@ def __are_requirements_matching(param_dic=None, test=None):
...
+                    for filename in user_home_files:
+                        if filename.startswith("core"):
+                            _msg = hostname
+                            _msg = ", user-" + user
+                            _msg += ": core file exist at "
+                            self.logger.warning(_msg + user_home['out'])
+                            self.hardware_report_timer.cancel()
+                            raise PbsHardwareError(msg=_msg+user_home['out'])
+        # server hardware report
+        for hostname in self.param_dict['servers']:
```
prakashcv13: there is a lot of repetition of code here, shall we combine the two for loops?

test/fw/ptl/utils/plugins/ptl_test_runner.py
birbalsain: I have added servers and comms block.

test/fw/ptl/utils/plugins/ptl_test_runner.py
birbalsain: i have used for loop in  new changes

test/fw/ptl/utils/plugins/ptl_test_runner.py
birbalsain: changed to HardwareReportError

test/fw/ptl/utils/plugins/ptl_test_runner.py
birbalsain: changed

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -747,6 +755,83 @@ def __are_requirements_matching(param_dic=None, test=None):
...
+    def hardware_report(self):
+        """
+        function used to get hardware report
+        every 5 minutes
+        """
+        du = DshUtils()
+        self.hardware_report_timer = Timer(300, self.hardware_report)
+        self.hardware_report_timer.start()
+        self.hardware_monitor_start = True
+        for system in ['servers', 'moms', 'comms']:
```
hirenvadalia: There will be chances of a duplicate host in servers, moms and comms so you can optimize these two for loop with the below code.
```
systems = self.param_dict['servers'].host_keys()
systems.extend(self.param_dict['moms'].host_keys())
systems.extend(self.param_dict['comms'].host_keys())
systems = list(set(systems))
for system in systems:
   ...
```

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -747,6 +755,83 @@ def __are_requirements_matching(param_dic=None, test=None):
...
 
+    def hardware_report(self):
+        """
+        function used to get hardware report
+        every 5 minutes
+        """
+        du = DshUtils()
+        self.hardware_report_timer = Timer(300, self.hardware_report)
+        self.hardware_report_timer.start()
+        self.hardware_monitor_start = True
```
hirenvadalia: remove all reference of `self.hardware_monitor_start` from this file as you are not using it anymore.

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -747,6 +755,83 @@ def __are_requirements_matching(param_dic=None, test=None):
             if not eff_tc_req['no_comm_on_mom']:
                 return False
 
+    def hardware_report(self):
```
hirenvadalia: how about renaming this to `check_and_report_hardware_status` to make more readable?

test/fw/ptl/utils/plugins/ptl_test_runner.py
birbalsain: moms doesn't have server_priv, comm_priv directory if all are a different system. So i have to check according to moms, servers and comms. But all or two on one system i have to use one other condition.

hirenvadalia: as per offline discussion, you can make use of du.isdir() to check whether require directory exists or not then check for the files in that directory

birbalsain:
I am not getting status of license/cla check because of that i am closing this PR. PR will be reopened.

test/fw/ptl/utils/plugins/ptl_test_runner.py
birbalsain: @prakashcv13 i have reviewed your all comments. if any suggestions, please suggest me.

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -747,6 +755,80 @@ def __are_requirements_matching(param_dic=None, test=None):
...
+            if used_disk_percent is None:
+                _msg = hostname
+                _msg += ": unable to get disk info"
+                self.logger.error(_msg)
+                self.hardware_report_timer.cancel()
+                raise PbsHardwareMonitorError(msg=_msg)
+            elif 70 < used_disk_percent < 95:
+                _msg = hostname + ": used disk is "
+                _msg += str(used_disk_percent) + "%"
+                _msg += "Please clean the disk"
```
prakashcv13: can we make the message better - something like - "disk usage is at str(used_disk_percent)%, disk cleanup is recommended."

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -747,6 +755,80 @@ def __are_requirements_matching(param_dic=None, test=None):
...
+                self.logger.warning(_msg)
+            elif used_disk_percent > 95:
+                _msg = hostname + ": used disk is "
+                _msg += str(used_disk_percent) + "%"
+                _msg += "test will be  stopped"
```
prakashcv13: here too - shall we say "disk usage > 95%, stopping the test(s)"

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -747,6 +755,80 @@ def __are_requirements_matching(param_dic=None, test=None):
             if not eff_tc_req['no_comm_on_mom']:
                 return False
 
+    def check_and_report_hardware_status(self):
```
prakashcv13: we are checking for core files as well. shall we name it check_core_files_and_disk_usage()?
if we are planning to add more hardware diagnostics to this function, we can keep "hardware" in the name, otherwise, all we are checking for is the disk usage.

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -747,6 +755,80 @@ def __are_requirements_matching(param_dic=None, test=None):
...
+                self.logger.error(_msg)
+                self.hardware_report_timer.cancel()
+                raise PbsHardwareMonitorError(msg=_msg)
+            # checks for core files
+            pbs_conf = du.parse_pbs_config(hostname)
+            mom_priv_path = os.path.join(pbs_conf["PBS_HOME"], "mom_priv")
+            if du.isdir(hostname, mom_priv_path):
+                mom_priv_files = du.listdir(hostname, mom_priv_path,
+                                            sudo=True, fullpath=False)
+                for filename in mom_priv_files:
```
prakashcv13: i am wondering if we could use fnmatch module's filter() and avoid the for loop below.

test/fw/ptl/utils/plugins/ptl_test_runner.py
birbalsain: checks_core_files_and_hardware_status. 
        """
        function checks hardware status and core files
        every 5 minutes
        """

prakashcv13: instead of checks_ - it should be check_
are we going to add more diagnostics later on?

birbalsain:
@prakashcv13 i have reviewed your comments, is any suggestions?

test/fw/ptl/utils/plugins/ptl_test_runner.py
birbalsain: yes

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -755,14 +756,14 @@ def __are_requirements_matching(param_dic=None, test=None):
             if not eff_tc_req['no_comm_on_mom']:
                 return False
 
-    def check_and_report_hardware_status(self):
+    def checks_hardware_status_and_core_files(self):
```
prakashcv13: remove the 's' in "checks"

test/fw/ptl/utils/plugins/ptl_test_runner.py
prakashcv13: this comment is not addressed completely.

prakashcv13:
@birbalsain - some comments were not replied to or addressed.

birbalsain:
@prakashcv13  I have done all the requested changes and atteched the  latest log with different cases.

prakashcv13:
@birbalsain - hardware_report_warning_message.txt does not have the warning message. Also, could you please test the logging of core file warning as well. Thank you!

birbalsain:
it has warning message 
2019-10-09 18:27:53,594 INFOCLI2 centos7: sudo -H ls /var/spool/pbs/sched_priv
2019-10-09 18:27:53,630 WARNING  centos7: core files found in /var/spool/pbs/sched_priv
2019-10-09 18:27:53,630 INFOCLI2 centos7: sudo -H ls /home/pbsuser
2019-10-09 18:27:53,670 INFOCLI2 centos7: sudo -H ls /home/pbsuser1
2019-10-09 18:27:53,705 INFOCLI2 centos7: sudo -H ls /home/pbsuser2
2019-10-09 18:27:53,740 WARNING  centos7: user-pbsuser2: core files found in /home/pbsuser2
2019-10-09 18:27:53,741 INFOCLI2 centos7: sudo -H ls /home/pbsuser3
2019-10-09 18:27:53,774 INFOCLI2 centos7: sudo -H ls /home/pbsuser4

birbalsain:
@prakashcv13  warning message at line number 394,398

prakashcv13:
great - we also need test logs for the disk usage warning message (one related to 70%) - if it is already there in the logs - please point me to that. Thanks!

birbalsain:
@prakashcv13 i have atteched  test logs for the disk usage warning message (one related to 70%)

prakashcv13:
which file?

prakashcv13:
@birbalsain - kindly attach logs before claiming that they are attached.

PR #1320, opened 2019-09-26
Correct modulefile MANPATH from $PBS_EXEC/man to $PBS_EXEC/share/man

scc138:
The modulefile sets an incorrect MANPATH.  The end result should be $PBS_EXEC/share/man, not $PBS_EXEC/man.

### DIFF
--- a/src/cmds/scripts/modulefile.in
+++ b/src/cmds/scripts/modulefile.in
@@ -14,5 +14,5 @@ puts stderr ""
 set _module_name [module-info name]
 set is_module_rm [module-info mode remove]
 set package_root @prefix@
-prepend-path MANPATH [file join ${package_root} man]
+prepend-path MANPATH [file join ${package_root} share/man]
 prepend-path PATH [file join ${package_root} bin]


prakashcv13:
thank you for your contribution @scc138 
i have restarted the failed debian field. the failures are totally unrelated and the build passes on other platforms.

subhasisb:
Oops! I was hoping to use the failed test results for this PR as a guinea pig to debug the intermittent test failures, possible after upgrading to python 3...

prakashcv13:
ohh, i was not aware...

mike0042:
The Debian build failed again. Please do not restart it. Also seeing the problem on #1279.

arungrover:
> Oops! I was hoping to use the failed test results for this PR as a guinea pig to debug the intermittent test failures, possible after upgrading to python 3...

I think Bhroam saw a similar problem in his testing recently. He has a possible root cause ->  now that log_events for the scheduler is set from qmgr, it sets the scheduler flag to update Config_attrs which means it sets a flag for sending high priority message to the scheduler. On the other hand when we submit jobs (which this is not high priority command for sched) it marks a separate flag to schedule a cycle. Now when scheduler processes the first command (CONFIG log_events) server clears up all the flags and schedule-a-job command never makes it to scheduler thereby running no jobs.

hirenvadalia:
> > Oops! I was hoping to use the failed test results for this PR as a guinea pig to debug the intermittent test failures, possible after upgrading to python 3...
> 
> I think Bhroam saw a similar problem in his testing recently. He has a possible root cause -> now that log_events for the scheduler is set from qmgr, it sets the scheduler flag to update Config_attrs which means it sets a flag for sending high priority message to the scheduler. On the other hand when we submit jobs (which this is not high priority command for sched) it marks a separate flag to schedule a cycle. Now when scheduler processes the first command (CONFIG log_events) server clears up all the flags and schedule-a-job command never makes it to scheduler thereby running no jobs.

Also, I found another possible root cause for Travis Debian failures (as I can reproduce the issue on my local machine):
In very last of PBSTestSuite.rever_mom we HUP/restart mom, create a node and immediately calls Server.expect() with `a = {'state': 'free', 'resources_available.ncpus': (GE, 1)}` but since the node is just created and HUPed/restarted, it is still coming online so it doesn't have `resources_available.ncpus` set on it and state is `state-unknown,down`. Now in Server.expect() at [here](https://github.com/PBSPro/pbspro/blob/master/test/fw/ptl/lib/pbs_testlib.py#L8233) we return False if we don't find given attribute name in status returned list.
I think returning False is wrong here, we should return from here unless we hit max attempts.
So I added calling expect again instead of returning False and all failed tests passed, I ran multiple time and all time it passes, if I revert to return False then it starts failing.

I will put my changes in a branch in my fork and will run multiple time Travis ci on it to see if it passes or not.

arungrover:
> > > Oops! I was hoping to use the failed test results for this PR as a guinea pig to debug the intermittent test failures, possible after upgrading to python 3...
> > 
> > 
> > I think Bhroam saw a similar problem in his testing recently. He has a possible root cause -> now that log_events for the scheduler is set from qmgr, it sets the scheduler flag to update Config_attrs which means it sets a flag for sending high priority message to the scheduler. On the other hand when we submit jobs (which this is not high priority command for sched) it marks a separate flag to schedule a cycle. Now when scheduler processes the first command (CONFIG log_events) server clears up all the flags and schedule-a-job command never makes it to scheduler thereby running no jobs.
> 
> Also, I found another possible root cause for Travis Debian failures (as I can reproduce the issue on my local machine):
> In very last of PBSTestSuite.rever_mom we HUP/restart mom, create a node and immediately calls Server.expect() with `a = {'state': 'free', 'resources_available.ncpus': (GE, 1)}` but since the node is just created and HUPed/restarted, it is still coming online so it doesn't have `resources_available.ncpus` set on it and state is `state-unknown,down`. Now in Server.expect() at [here](https://github.com/PBSPro/pbspro/blob/master/test/fw/ptl/lib/pbs_testlib.py#L8233) we return False if we don't find given attribute name in status returned list.
> I think returning False is wrong here, we should return from here unless we hit max attempts.
> So I added calling expect again instead of returning False and all failed tests passed, I ran multiple time and all time it passes, if I revert to return False then it starts failing.
> 
> I will put my changes in a branch in my fork and will run multiple time Travis ci on it to see if it passes or not.

I had a small chat with Bhroam and it seems the problem he was talking about was different than what I mentioned. In fact, I tried it myself and I do not see server clearing both flags. So just ignore what I said. I hope Hiren's fix will solve this problem. I apologize if I misled anyone.

subhasisb:
> > > > Oops! I was hoping to use the failed test results for this PR as a guinea pig to debug the intermittent test failures, possible after upgrading to python 3...
> > > 
> > > 
> > > I think Bhroam saw a similar problem in his testing recently. He has a possible root cause -> now that log_events for the scheduler is set from qmgr, it sets the scheduler flag to update Config_attrs which means it sets a flag for sending high priority message to the scheduler. On the other hand when we submit jobs (which this is not high priority command for sched) it marks a separate flag to schedule a cycle. Now when scheduler processes the first command (CONFIG log_events) server clears up all the flags and schedule-a-job command never makes it to scheduler thereby running no jobs.
> > 
> > 
> > Also, I found another possible root cause for Travis Debian failures (as I can reproduce the issue on my local machine):
> > In very last of PBSTestSuite.rever_mom we HUP/restart mom, create a node and immediately calls Server.expect() with `a = {'state': 'free', 'resources_available.ncpus': (GE, 1)}` but since the node is just created and HUPed/restarted, it is still coming online so it doesn't have `resources_available.ncpus` set on it and state is `state-unknown,down`. Now in Server.expect() at [here](https://github.com/PBSPro/pbspro/blob/master/test/fw/ptl/lib/pbs_testlib.py#L8233) we return False if we don't find given attribute name in status returned list.
> > I think returning False is wrong here, we should return from here unless we hit max attempts.
> > So I added calling expect again instead of returning False and all failed tests passed, I ran multiple time and all time it passes, if I revert to return False then it starts failing.
> > I will put my changes in a branch in my fork and will run multiple time Travis ci on it to see if it passes or not.
> 
> I had a small chat with Bhroam and it seems the problem he was talking about was different than what I mentioned. In fact, I tried it myself and I do not see server clearing both flags. So just ignore what I said. I hope Hiren's fix will solve this problem. I apologize if I misled anyone.

Thanks for jumping in and trying to find the root cause. Very much appreciated.

scc138:
I restarted the Travis build and it works now, thanks!  I think this is good to go?

prakashcv13:
merged #1320

Issue #1317, opened 2019-09-26
scheduler may oversubscribe node with subjob


vchlum:
I observed some oversubscribing. The scheduler may probably oversubscribe the node with a subjob. The problem is probably caused by the filter added in #872 .  The scheduler can miscount free resources now. The scheduler considers all not running subjobs to be gone (has_ghost_job==1). In order to fix this and do not break #791, I would suggest changing the filter added in #872 into filtering jobs in JOB_STATE_EXPIRED only.

Please, could someone check it with me (@arungrover)? I am not raising PR because I am not really sure yet.

arungrover:
@vchlum can you elaborate more on this issue?
When scheduler detects that there are ghost jobs in the system, it reassigns the resources on the node according to the jobs it gets by querying server.
Are you worries that sub-jobs with exiting state will not be reported to scheduler because of the check added in #872 and this would mean that scheduler will start running jobs on the node while the mom is still processing exiting jobs? Maybe in that case filter should be that if the job is not in the running state or not in the exiting state then do not report those jobs to scheduler.

vchlum:
@arungrover Yes, I think that is the issue - the exiting state of sub-jobs is not reported to the scheduler. In such a case, the scheduler set the has_ghost_job to true and reassigns the resources to the node incorrectly - it does not count the exiting sub-jobs. But the exiting sub-jobs can still be there. The exiting state can take a while... and meanwhile the oversubscribing can probably happen.

I think you are right that the filter that reports running or exiting sub-jobs in #872 is probably enough to fix this issue. The two states mean that a job occupies a node. My other question is: Is that policy in the right place? Shouldn't we send all the jobs to the scheduler and let it decide what states occupy a node?

vchlum:
I can randomly reproduce the issue by submitting 'echo "sleep 1" | qsub -J 1-1000' into a system with two nodes and each node has 2 ncpus:
```
(JESSIE)vchlum@torque1:~$ pbsnodes -v torque2 | grep -E "cpus|jobs"
     pcpus = 2
     jobs = 1145[719].torque1.ics.muni.cz/0, 1145[723].torque1.ics.muni.cz/1, 1145[722].torque1.ics.muni.cz/1, 1145[720].torque1.ics.muni.cz/1
     resources_available.ncpus = 2
     resources_assigned.ncpus = 4
```

It happens even if the filter added in #872 is removed... So the problem may be somewhere else...

vchlum:
[...]

PR #1328, opened 2019-10-01
scheduler may oversubscribe node with exiting subjob

vchlum:
#### Describe Bug or Feature
The scheduler can oversubscribe the node with exiting sub-job on it. As described in #1317, the sub-jobs in the exiting state are not processed by the scheduler and if such a sub-job occupies the node, the resources are miscounted.

Submitting 'sleep 1" | qsub -J 1-1000', the bug randomly appears and it looks like this:
```
(JESSIE)vchlum@torque1:~$ pbsnodes -v torque2 | grep -E "cpus|jobs"
     pcpus = 2
     jobs = 1145[719].torque1.ics.muni.cz/0, 1145[723].torque1.ics.muni.cz/1, 1145[722].torque1.ics.muni.cz/1, 1145[720].torque1.ics.muni.cz/1
     resources_available.ncpus = 2
     resources_assigned.ncpus = 4
```

#### Describe Your Change
The scheduler should process all states that occupy a node, which includes the exiting state.  Let the scheduler process JOB_STATE_EXITING too.

#### Link to Design Doc
None

PTL test is not suitable for this fix because the problem can not be reliably reproduced, but a manual test with the fix confirmed correctness - With the fix, I wasn't able to reproduce the bug at all.

### DIFF
--- a/src/server/req_select.c
+++ b/src/server/req_select.c
@@ -490,6 +490,7 @@ select_job(job *pjob, struct select_list *psel, int dosubjobs, int dohistjobs)
 	}
 
 	if ((dosubjobs == 2) && (pjob->ji_qs.ji_svrflags & JOB_SVFLG_SubJob) &&
+		(pjob->ji_qs.ji_state != JOB_STATE_EXITING) &&
 		(pjob->ji_qs.ji_state != JOB_STATE_RUNNING)) /* select only running subjobs */
 		return 0;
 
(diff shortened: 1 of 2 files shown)


src/server/req_select.c
```diff
@@ -490,6 +490,7 @@ select_job(job *pjob, struct select_list *psel, int dosubjobs, int dohistjobs)
 	}
 
 	if ((dosubjobs == 2) && (pjob->ji_qs.ji_svrflags & JOB_SVFLG_SubJob) &&
+		(pjob->ji_qs.ji_state != JOB_STATE_EXITING) &&
 		(pjob->ji_qs.ji_state != JOB_STATE_RUNNING)) /* select only running subjobs */
```
bhroam: Can you update this comment?

vchlum: @bhroam done

bhroam:
Thanks @vchlum.  We're currently having issues with a change in OpenSuse Leap with Travis.  It's being worked on.  Once that is fixed, you'll need to rebase to get those changes and then Travis will pass.  We can merge after that.

vchlum:
@bhroam It is rebased.

PR #1329, opened 2019-10-01
pbs.get_local_nodename() data inconsistencies for nodes added by IP address

sujatapatnaik52:
#### Describe Bug or Feature
* When a node is created using IP address, the hook API pbs.get_local_nodename() returns the host name of the machine. In cgroups hook, the pbs.get_local_nodename() returning hostname is matched with the exec_vnode of the job which is set to the IP address. There is a mismatch of these two, the job does not run successfully.
* When a node is created using IP address and the PBS_MOM_NODE_NAME variable is set in /etc/pbs.conf, pbs.get_local_nodename() returns a part of the IP address since the mom name is split by a dot i.e. Ex- If the IP adrdess is 172.0.0.1 then the above api returns just "172". In cgroups hooks, this is matched with exec_vnode of the job which is the full IP address and there is a mismatch.

#### Describe Your Change
For a case, where the nodes are created using IP address, we will make it mandatory for the users to set the variable PBS_MOM_NODE_NAME in /etc/pbs.conf and check if it's an IP address then don't split it by the dot.

I have ran all automated tests and none failed. 

[verification_logs.txt](https://github.com/PBSPro/pbspro/files/3675519/verification_logs.txt)
[cgroups_hook_logs.txt](https://github.com/PBSPro/pbspro/files/3675520/cgroups_hook_logs.txt)

### DIFF
--- a/src/resmom/mom_main.c
+++ b/src/resmom/mom_main.c
@@ -125,6 +125,7 @@
 #include	"mom_mach.h"
 #endif	/* MOM_CSA or MOM_ALPS */
 #include	"pbs_reliable.h"
+#include	<arpa/inet.h>
 
 #define STATE_UPDATE_TIME 10
 #ifndef	PRIO_MAX
@@ -8212,6 +8213,8 @@ main(int argc, char *argv[])
 	char				path_hooks_rescdef[MAXPATHLEN+1];
 	int					sock_bind_rm;
 	int					sock_bind_mom;
+	struct				sockaddr_in check_ip;
+	int				is_mom_host_ip;
 #ifdef	WIN32
 	/* Win32 only */
 	struct arg_param	*p = (struct arg_param *)pv;
@@ -8873,9 +8876,10 @@ main(int argc, char *argv[])
 	}
 	(void)strncpy(mom_short_name, mom_host, (sizeof(mom_short_name) - 1));
 	mom_short_name[(sizeof(mom_short_name) - 1)] = '\0';
-	if ((ptr = strchr(mom_short_name, (int)'.')) != NULL)
-		*ptr = '\0';  /* terminate shortname at first dot */
-
+	is_mom_host_ip = inet_pton(AF_INET, mom_host, &(check_ip.sin_addr));
+	if (!(is_mom_host_ip > 0))
+		if ((ptr = strchr(mom_short_name, (int)'.')) != NULL)
+			*ptr = '\0';  /* terminate shortname at first dot */
 	/*
 	 * Now get mom_host, which determines resources_available.host
 	 * and also the interface used to register to pbs_comm if
--- a/src/server/req_manager.c
+++ b/src/server/req_manager.c
@@ -3800,10 +3800,9 @@ struct batch_request *preq;
 		}
 	}
 	is_node_ip = inet_pton(AF_INET, preq->rq_ind.rq_manager.rq_objname, &(check_ip.sin_addr));
-	if(is_node_ip > 0) {
+	if (is_node_ip > 0) {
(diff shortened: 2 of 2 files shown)


sujatapatnaik52:
Force pushing once to verify travis.

src/resmom/mom_main.c
```diff
@@ -8873,9 +8876,10 @@ main(int argc, char *argv[])
 	}
 	(void)strncpy(mom_short_name, mom_host, (sizeof(mom_short_name) - 1));
 	mom_short_name[(sizeof(mom_short_name) - 1)] = '\0';
-	if ((ptr = strchr(mom_short_name, (int)'.')) != NULL)
-		*ptr = '\0';  /* terminate shortname at first dot */
-
+	is_mom_host_ip = inet_pton(AF_INET, mom_host, &(check_ip.sin_addr));
+	if (!(is_mom_host_ip > 0))
+		if ((ptr = strchr(mom_short_name, (int)'.')) != NULL)
+			*ptr = '\0';  /* terminate shortname at first dot */
```
prakashcv13: if the user is not setting PBS_MOM_NODE_NAME parameter in pbs.conf, we would still call gethostname(). Should we log the case where we need the PBS_MOM_NODE_NAME and it is not set.

sujatapatnaik52: @prakashcv13 As discussed offline, I cannot find out if the mom is created using IP address at this place, as per your suggestion, I have added a log message for debugging while creating the node. Please look at the latest commit.

sujatapatnaik52:
[updated_logs.txt](https://github.com/PBSPro/pbspro/files/3690137/updated_logs.txt)

src/server/req_manager.c
```diff
@@ -3796,6 +3799,12 @@ struct batch_request *preq;
 			return;
 		}
 	}
+	is_node_ip = inet_pton(AF_INET, preq->rq_ind.rq_manager.rq_objname, &(check_ip.sin_addr));
+	if(is_node_ip > 0) {
```
prakashcv13: space after if is needed

prakashcv13: space after if is needed.

src/server/req_manager.c
```diff
@@ -3796,6 +3799,12 @@ struct batch_request *preq;
...
+		log_eventf(PBSEVENT_DEBUG, PBS_EVENTCLASS_NODE, LOG_INFO, preq->rq_ind.rq_manager.rq_objname,
+			"Add PBS_MOM_NODE_NAME='IP address' in /etc/pbs.conf in Mom host and restart MOM to avoid node inconsistencies.");
+
```
prakashcv13: extra line.

src/server/req_manager.c
```diff
@@ -3796,6 +3799,12 @@ struct batch_request *preq;
 			return;
 		}
 	}
+	is_node_ip = inet_pton(AF_INET, preq->rq_ind.rq_manager.rq_objname, &(check_ip.sin_addr));
+	if(is_node_ip > 0) {
+		log_eventf(PBSEVENT_DEBUG, PBS_EVENTCLASS_NODE, LOG_INFO, preq->rq_ind.rq_manager.rq_objname,
+			"Add PBS_MOM_NODE_NAME='IP address' in /etc/pbs.conf in Mom host and restart MOM to avoid node inconsistencies.");
```
prakashcv13: shall we say - "Node added using IP address\nVerify that PBS_MOM_NODE_NAME is configured on the respective host"

sujatapatnaik52:
[updated_logs.txt](https://github.com/PBSPro/pbspro/files/3690407/updated_logs.txt)

sujatapatnaik52:
Thank you @hirenvadalia @sakshamgarg @prakashcv13 for the sign off. @prakashcv13 PR ready for merge.

PR #1331, opened 2019-10-01
Build fails in opensuse leap-15 due to third party issues

nithinj:
#### Describe Bug or Feature
Travis builds are failing in OpenSuse Leap 15 as it fetches the latest version of hwloc available in its repo (hwloc 2.0) which does not work with PBSPro resulting in compilation failure

#### Describe Your Change
* hwloc version should be lesser than 2.0 (updating rpm spec file)
* Travis build scripts should make use of version specified in the rpm spec file

#### Link to Design Doc
NA

Travis Build Fix - Infra changes.

CC: @arungrover

### DIFF
--- a/.travis/do.sh
+++ b/.travis/do.sh
@@ -37,8 +37,8 @@ if [ "x${ONLY_REBUILD}" != "x1" ]; then
     zypper -n update
     zypper -n install rpmdevtools python3-pip sudo which net-tools man time.x86_64
     rpmdev-setuptree
-    zypper -n install --force-resolution $(rpmspec --buildrequires -q ${SPEC_FILE} | awk '{print $1}'| sort -u | grep -vE '^(/bin/)?(ba)?sh$')
-    zypper -n install --force-resolution $(rpmspec --requires -q ${SPEC_FILE} | awk '{print $1}'| sort -u | grep -vE '^(/bin/)?(ba)?sh$')
+    zypper -n install --force-resolution $(rpmspec --buildrequires -q ${SPEC_FILE} | sort -u | grep -vE '^(/bin/)?(ba)?sh$')
+    zypper -n install --force-resolution $(rpmspec --requires -q ${SPEC_FILE} | sort -u | grep -vE '^(/bin/)?(ba)?sh$')
     pip3 install -r ${REQ_FILE}
   elif [ "x${ID}" == "xdebian" ]; then
     if [ "x${DEBIAN_FRONTEND}" == "x" ]; then
--- a/pbspro.spec
+++ b/pbspro.spec
@@ -100,7 +100,7 @@ BuildRequires: rpm-build
 BuildRequires: autoconf
 BuildRequires: automake
 BuildRequires: libtool
-BuildRequires: hwloc-devel
+BuildRequires: hwloc-devel < 2.0
 BuildRequires: libX11-devel
 BuildRequires: libXt-devel
 BuildRequires: libedit-devel
--- a/pbspro.spec.in
+++ b/pbspro.spec.in
@@ -100,7 +100,7 @@ BuildRequires: rpm-build
 BuildRequires: autoconf
 BuildRequires: automake
 BuildRequires: libtool
-BuildRequires: hwloc-devel
+BuildRequires: hwloc-devel < 2.0
 BuildRequires: libX11-devel
 BuildRequires: libXt-devel
 BuildRequires: libedit-devel


.travis/do.sh
```diff
@@ -37,8 +37,8 @@ if [ "x${ONLY_REBUILD}" != "x1" ]; then
     zypper -n update
     zypper -n install rpmdevtools python3-pip sudo which net-tools man time.x86_64
     rpmdev-setuptree
-    zypper -n install --force-resolution $(rpmspec --buildrequires -q ${SPEC_FILE} | awk '{print $1}'| sort -u | grep -vE '^(/bin/)?(ba)?sh$')
-    zypper -n install --force-resolution $(rpmspec --requires -q ${SPEC_FILE} | awk '{print $1}'| sort -u | grep -vE '^(/bin/)?(ba)?sh$')
+    zypper -n install --force-resolution $(rpmspec --buildrequires -q ${SPEC_FILE} | sort -u | grep -vE '^(/bin/)?(ba)?sh$')
```
arungrover: Why did you remove awk from the command?

bhroam: I ran that command on my opensuse 15 leap machine and got several multi-fielded lines.  I think the awk is required.

nithinj: awk just keeps the first column and strips off the rest. It cuts out the version field which is required to install the right dependencies.
```
[nithinj@centos7 pbspro]$ rpmspec --buildrequires -q pbspro.spec | awk '{print $1}'| sort -u | grep -vE '^(/bin/)?(ba)?sh$'
autoconf
automake
expat-devel
gcc
hwloc-devel
libedit-devel
libical-devel
libtool
libX11-devel
libXext
libXft
libXt-devel
make
ncurses-devel
openssl-devel
perl
postgresql-contrib
postgresql-devel
python3-devel
rpm-build
swig
tcl-devel
tk-devel
zlib-devel
[nithinj@centos7 pbspro]$ rpmspec --buildrequires -q pbspro.spec | sort -u | grep -vE '^(/bin/)?(ba)?sh$'
autoconf
automake
expat-devel
gcc
hwloc-devel < 2.0
libedit-devel
libical-devel
libtool
libX11-devel
libXext
libXft
libXt-devel
make
ncurses-devel
openssl-devel
perl
postgresql-contrib >= 9.1
postgresql-devel >= 9.1
python3-devel >= 3.5
rpm-build
swig
tcl-devel
tk-devel
zlib-devel
```

hirenvadalia: we need awk because > & < is redirection in shell script

prakashcv13: from the travis build logs it looks like zypper has the intelligence to parse <, >

prakashcv13: on my opensuse container - 
d6588032cbb1:/ # zypper install hwloc-devel '<' 2.0

Building repository 'Non-OSS Repository' cache .......................................................................................................................................................[done]
Building repository 'Main Repository' cache ..........................................................................................................................................................[done]
Building repository 'Main Update Repository' cache ...................................................................................................................................................[done]
Building repository 'Update Repository (Non-Oss)' cache ..............................................................................................................................................[done]
Loading repository data...
Reading installed packages...
Resolving package dependencies...

The following 16 NEW packages are going to be installed:
  glibc-devel hwloc-data hwloc-devel libhwloc5 libnuma1 libpciaccess-devel libpciaccess0 libxml2-devel libxml2-tools linux-glibc-devel ncurses-devel pkg-config readline-devel tack xz-devel zlib-devel

16 new packages to install.
Overall download size: 8.4 MiB. Already cached: 0 B. After the operation, additional 27.8 MiB will be used.
Continue? [y/n/v/...? shows all options] (y): y
Retrieving package hwloc-data-1.11.8-lp151.2.3.noarch                                                                                                                 (1/16),  35.8 KiB (  4.4 KiB unpacked)
Retrieving: hwloc-data-1.11.8-lp151.2.3.noarch.rpm ...................................................................................................................................................[done]

hirenvadalia: Thanks @prakashcv13, IIRC awk was added to handle shell redirection due to two python-devel declarations in spec file (Before Py3) since now we have only one python-devel we can remove awk command.
@nithinj You can also remove `| sort -u` as it is not needed anymore.

nithinj: `sort -u` is there to pick unique values as `rpmspec --requires` does not pick values corresponding to a daemon but parses every Requires irrespective of the daemon. It is not really an issue as Travis build is meant for all daemons and we don't build a specific one. But filtering it through shell does make sense as it is more efficient compared to passing onto zypper and waiting for it's decision on whether the package is already installed.

PR #1332, opened 2019-10-01
Fixed call to find_node_by_indrank()

bhroam:
## Describe Bug or Feature
A recent change to the normal node codepath changed several calls of find_node_by_rank() to find_node_by_indrank().  I was a little overzealous on this and changed too many.  The two calls in eval_complex_selspec() and eval_placement() are correct.  The call in eval_simple_selspec() is not.  The reason the call in eval_simple_selspec() is wrong is the fact that eval_placement() dups the nodes for free placement.  We need the exec_vnode structures to hold pointers to the dup'd nodes, so we can modify the nodes.  When you call find_node_by_indrank(), you get pointers back to the real nodes.  This means that when eval_placement() is trying to modify the dup'd nodes, it would modify the real node and perturb future attempts to run jobs.

#### Describe Your Change
Revert the call to find_node_by_indrank() in eval_selspec() to find_node_by_rank()

This was a refactoring project that went a little far.  No extra testing should be required.

### DIFF
--- a/src/scheduler/node_info.c
+++ b/src/scheduler/node_info.c
@@ -3211,14 +3211,14 @@ eval_simple_selspec(status *policy, chunk *chk, node_info **pninfo_arr,
 							ns->end_of_chunk = 1;
 						}
 
-						if (ns != NULL) {
-							/* replace the dup'd node with the real one */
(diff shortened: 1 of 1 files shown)


bhroam:
@agrawalravi90 I ran a regression and everything passed

PR #1338, opened 2019-10-07
memory leak when maintenance overlaps a reservation

prakashcv13:
#### Describe Bug or Feature
valgrind detected memory leak while running TestMaintenanceReservations with the below stack -
> 64 bytes in 4 blocks are definitely lost in loss record 96 of 734
   by 0x483EDC: set_nodes (node_manager.c:7568)
   by 0x4ABD99: assign_resv_resc (req_rescq.c:448)
   by 0x4AC69B: req_confirmresv (req_rescq.c:776)
   by 0x50C5F6: process_socket (net_server.c:504)
   by 0x50C730: wait_request (net_server.c:585)
   by 0x4911A9: main (pbsd_main.c:2141)

#### Describe Your Change

When a maintenance overlaps a reservation, we remove the nodes belonging to the maintenance from the resv_nodes of the reservation and the reservation is also removed from all the vnodes belonging to the host(s) on which the maintenance is submitted. Now, when we delete the reservation, we call free_resvNodes(). This function parses all the nodes on the server and free()s the ri_pbsnode_list. However, we had removed the reference of this reservation from the node when we submitted the maintenance, hence ri_pbsnode_list is not free()-ed.

**Solution:** While removing the host from the reservation, delete the vnode(s) of that host from ri_pbsnode_list as well.

#### Valgrind Logs/Output

[val_out_before.txt](https://github.com/PBSPro/pbspro/files/3698913/val_out_before.txt)
[val_out_after.txt](https://github.com/PBSPro/pbspro/files/3698914/val_out_after.txt)

### DIFF
--- a/src/server/req_rescq.c
+++ b/src/server/req_rescq.c
@@ -250,7 +250,7 @@ remove_node_from_resv(resc_resv *presv, struct pbsnode *pnode)
 		return;
 	}
 
-	snprintf(tmp_buf, strlen(pnode->nd_name) + 1,"%s:", pnode->nd_name);
+	snprintf(tmp_buf, strlen(pnode->nd_name) + 1, "%s:", pnode->nd_name);
 
 	/* remove the node '(vn[n]:foo)' from RESV_ATR_resv_nodes attribute */
 	if (presv->ri_wattr[RESV_ATR_resv_nodes].at_flags & ATR_VFLAG_SET) {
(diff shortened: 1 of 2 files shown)


src/server/req_rescq.c
```diff
@@ -351,9 +351,21 @@ remove_node_from_resv(resc_resv *presv, struct pbsnode *pnode)
...
 remove_host_from_resv(resc_resv *presv, char *hostname) {
 	pbsnode_list_t *pl = NULL;
-	for (pl = presv->ri_pbsnode_list; pl != NULL; pl = pl->next) {
-		if (strcmp(pl->vnode->nd_hostname, hostname) == 0)
+	pbsnode_list_t *prev = NULL;
+
+	for (prev = NULL, pl = presv->ri_pbsnode_list; pl != NULL;) {
+		if (strcmp(pl->vnode->nd_hostname, hostname) == 0) {
 			remove_node_from_resv(presv, pl->vnode);
+			if (prev == NULL) {
```
arungrover: I don't understand this change. "prev" is never assigned a value so it will always be NULL. That means you are always removing from the front of the list.

prakashcv13: i had removed pl = pl->next from the for statement, to add it, but forgot. thanks for catching it.

prakashcv13:
[val_out_after2_1.txt](https://github.com/PBSPro/pbspro/files/3703647/val_out_after2_1.txt)
[val_out_after2.txt](https://github.com/PBSPro/pbspro/files/3703648/val_out_after2.txt)

prakashcv13:
thank you @mike0042 and @arungrover

PR #1340, opened 2019-10-07
release_19_1_branch: Make mom accept only privileged requests

bayucan:
#### Describe Bug or Feature

* An attacker can sniff the network (e.g. using tcpdump) and as a regular user modify the execution user attribute of a job.

* The cause is that pbs_mom just blindly accepts any request, trusting that it is coming from a pbs_server process (daemon process) running on a privilege port.

#### Describe Your Change

* The privileged port origin check is missing in pbs_mom (it's there on the pbs_server side) the job can execute as an unauthorized user, including root.

* [fix_19_1_test.txt](https://github.com/PBSPro/pbspro/files/3699058/fix_19_1_test.txt)

### DIFF
--- a/src/server/process_request.c
+++ b/src/server/process_request.c
@@ -472,6 +472,12 @@ process_request(int sfds)
 		return;
 	}
 
+	if ((conn->cn_authen & PBS_NET_CONN_FROM_PRIVIL) == 0) {
+		req_reject(PBSE_BADCRED, 0, request);
+		close_client(sfds);
+		return;
+	}
+
 	request->rq_fromsvr = 1;
 	request->rq_perm = ATR_DFLAG_USRD | ATR_DFLAG_USWR |
 			   ATR_DFLAG_OPRD | ATR_DFLAG_OPWR |


mike0042:
Is your branch up to date @bayucan? This looks like the same failure we've seen previously on OpenSUSE Leap. If not, please rebase.

bayucan:
@mike0042 : The source is from release_19_1_branch, which is the earlier OSS release, which may not be compatible with opensuse-leap15. Nothing has been checked-in to release_19_1_branch for a very long time, and I just checked and my branch is up to date...

prakashcv13:
@mike0042, @bayucan - we would need to cherry-pick PR #1331 to this branch. The issue is because travis now installs hwloc-2 instead of hwloc-1.

mike0042:
I suggest we merge #1331 as part of this PR. Leap is a supported platform on this branch, and it should build properly. It's a relatively small change. @PBSPro/pbspro-maintainers - would you agree?

prakashcv13:
@mike0042 - I have just created a PR (#1342) for picking relevant changes from PR #1331. I was of the view that keeping the two things separate would be good.

mike0042:
Agreed. Thank you @prakashcv13.

bayucan:
@prakashcv13 @mike0042 Thanks, guys! It finally passed.

PR #1341, opened 2019-10-07
Node wrongly assigned to suspended job after server restart

arungrover:
#### Describe Bug or Feature
If a running job is suspended and then server is restarted, the node gets assigned back again to the job. This only happens when restrict_res_to_release_on_suspend attribute is set in PBS server.
The cause of the problem is that when "restrict_res_to_release_on_suspend" is set on server, PBS server on a restart calls "reassign_resc()" on suspended jobs with released resources. This function internally calls set_nodes which sets the node's "jobs" attribute and assigns virtual cpu slots to this job.

#### Describe Your Change
Added a check in set_nodes to check if the job that is suspended and set_nodes is called on a restart for the first time, do not allow set_nodes to modify node's "jobs" attribute or state.

#### Link to Design Doc
NA

[test_before.txt](https://github.com/PBSPro/pbspro/files/3700018/test_before.txt)
[test_after.txt](https://github.com/PBSPro/pbspro/files/3700019/test_after.txt)

### DIFF
--- a/test/tests/functional/pbs_release_limited_res_suspend.py
+++ b/test/tests/functional/pbs_release_limited_res_suspend.py
@@ -949,7 +949,7 @@ class TestReleaseLimitedResOnSuspend(TestFunctional):
         self.server.manager(MGR_CMD_SET, SCHED,
                             {'preempt_order': preempt_method}, runas=ROOT_USER)
 
-        # Set 1gb mem available on the node
+        # Set 2 ncpus available on the node
         a = {ATTR_rescavail + '.ncpus': "2"}
         self.server.manager(MGR_CMD_SET, NODE, a, self.mom.shortname)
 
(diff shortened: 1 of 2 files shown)


test/tests/functional/pbs_release_limited_res_suspend.py
```diff
@@ -997,3 +997,42 @@ def test_preempt_checkpoint_release_all(self):
...
+        PBS server is restarted, the job is able to resume back on the same
+        node.
+        """
+        # Set ncpus in restrict_res_to_release_on_suspend server attribute
+        a = {ATTR_restrict_res_to_release_on_suspend: 'ncpus'}
+        self.server.manager(MGR_CMD_SET, SERVER, a)
+
+        # Set 2 ncpus available on the node
+        self.server.manager(MGR_CMD_DELETE, NODE, id="@default")
+        self.server.manager(MGR_CMD_CREATE, NODE, id=self.mom.shortname)
```
bhroam: Why delete the nodes?  setUp() should do this for you.  The default should just be the natural vnode.

test/tests/functional/pbs_release_limited_res_suspend.py
```diff
@@ -997,3 +997,42 @@ def test_preempt_checkpoint_release_all(self):
...
+        a = {ATTR_rescavail + '.ncpus': "2"}
+        self.server.manager(MGR_CMD_SET, NODE, a, self.mom.shortname)
+
+        # Submit a job which takes up all of the ncpus
+        j1 = Job(TEST_USER)
```
bhroam: Up to you: You don't have to say TEST_USER any more.  It is the default.  Of course you might want it just to be consistent with the rest of the test plan.

test/tests/functional/pbs_release_limited_res_suspend.py
```diff
@@ -997,3 +997,42 @@ def test_preempt_checkpoint_release_all(self):
...
+        j1.set_attributes({ATTR_l + '.select': '1:ncpus=2'})
+        jid1 = self.server.submit(j1)
+        self.server.expect(JOB, {ATTR_state: 'R'}, id=jid1)
+
+        # make sure that job id is part of node's jobs attribute
+        node = self.server.status(NODE, id=self.mom.shortname)
+        self.assertTrue(jid1 in node[0]['jobs'])
```
bhroam: use assertIn()

test/tests/functional/pbs_release_limited_res_suspend.py
```diff
@@ -997,3 +997,42 @@ def test_preempt_checkpoint_release_all(self):
...
+
+        self.server.restart()
+
+        self.assertTrue(self.server.isUp())
+        self.server.expect(NODE, {'state': 'free'}, id=self.mom.shortname)
+        self.server.expect(NODE, 'jobs', op=UNSET, id=self.mom.shortname)
+
+        # resume job
+        self.server.sigjob(jobid=jid1, signal="resume")
+        self.server.expect(JOB, {ATTR_state: 'R'}, id=jid1)
```
bhroam: Since you check that the job is not in the jobs line above, you should check that it is in the jobs line here.

test/tests/functional/pbs_release_limited_res_suspend.py
arungrover: Right, Since setup of this test suite sets node with 4 cpus I was recreating it. But you are correct, I guess just setting right resources_available on the node should do the job.

test/tests/functional/pbs_release_limited_res_suspend.py
arungrover: That is true, for consistency sake, I'd like to retain it.

arungrover:
Thanks for reviewing this pull request @bhroam. I have addressed your review comments

arungrover:
Thanks @bhroam and @vstumpf for the review. Can you please merge it.

PR #1347, opened 2019-10-10
while deleting sub-job, the server can get stuck in an infinite loop

vchlum:
#### Describe Bug or Feature
I observed that the server can get stuck in an infinite loop. If the sub-job state is JOB_STATE_EXITING and the server tries to delete it, then the sub-job is meant to be skipped but the array index is not increased, which leads to infinite loop.

#### Describe Your Change
Increase the sub-job index, please see the context of the change. It is obvious. It is a similar case to the case for 'no such index' a few lines above.

#### Link to Design Doc
None

None. The PTL test is not suitable because it is not easy to reproduce this bug.

### DIFF
--- a/src/server/req_delete.c
+++ b/src/server/req_delete.c
@@ -552,8 +552,11 @@ req_deletejob(struct batch_request *preq)
 			}
 
 			sjst = get_subjob_state(parent, i);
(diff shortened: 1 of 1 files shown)


prakashcv13:
thank you for your contribution @vchlum. 
merged #1347 to master.

PR #1348, opened 2019-10-10
TestNonprintingCharacters: python3 adjustments to open()

vccardenas:
#### Describe Bug or Feature
Non printing character \r is getting converted to \n when using open() on the output file. 

#### Describe Your Change
Use newline="" in open().

#### Attach Test Logs/Output
[TestNonprintingCharacters_a01.txt](https://github.com/PBSPro/pbspro/files/3714006/TestNonprintingCharacters_a01.txt)

image

### DIFF
--- a/test/tests/functional/pbs_nonprint_characters.py
+++ b/test/tests/functional/pbs_nonprint_characters.py
@@ -145,7 +145,7 @@ sleep 5
         job_out = '\n'.join(ret['out'])
         self.logger.info('job output from %s:\n%s' % (job_outfile, job_out))
         job_output = ""
-        with open(job_outfile, 'r') as f:
+        with open(job_outfile, 'r', newline="") as f:
             job_output = f.read().strip()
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_nonprint_characters.py
```diff
@@ -741,7 +741,7 @@ def test_nonprint_character_interactive_job(self):
             # Once all commands sent and matched, job exits
             self.server.expect(JOB, 'queue', op=UNSET, id=jid, offset=1)
             # Check for the non-printable character in the output file
-            with open(self.job_out1_tempfile) as fd:
+            with open(self.job_out1_tempfile, newline="") as fd:
```
mike0042: Do we need the same change at lines 786 and 859?

vccardenas: No we don't because those parts test terminal control characters and \r is not used.

PR #1353, opened 2019-10-14
Fixed maintenance resv test to set managers correctly

arungrover:
#### Describe Bug or Feature
test_maintenance_resv test was creating TEST_USER as manager and then unsetting the managers attribute. This was overwriting the existing "managers" attribute on the server object (which by default is the user who is running the test). Because of overwriting managers attribute, unset command used to fail with permission denied error.

#### Describe Your Change
Instead of overwriting the attribute, use INCR/DECR operator to change managers

#### Link to Design Doc
NA

[test.txt](https://github.com/PBSPro/pbspro/files/3726247/test.txt)
[test_fixed.txt](https://github.com/PBSPro/pbspro/files/3726248/test_fixed.txt)

### DIFF
--- a/test/tests/functional/pbs_maintenance_reservations.py
+++ b/test/tests/functional/pbs_maintenance_reservations.py
@@ -244,7 +244,7 @@ class TestMaintenanceReservations(TestFunctional):
         now = int(time.time())
 
         self.server.manager(MGR_CMD_SET, SERVER,
-                            {'managers': '%s@*' % TEST_USER})
+                            {'managers': (INCR, '%s@*' % TEST_USER)})
 
         a = {'reserve_start': now + 3600,
              'reserve_end': now + 7200}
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_maintenance_reservations.py
```diff
@@ -244,7 +244,7 @@ def test_maintenance_delete(self):
         now = int(time.time())
 
         self.server.manager(MGR_CMD_SET, SERVER,
-                            {'managers': '%s@*' % TEST_USER})
+                            {'managers': (INCR, '%s@*' % TEST_USER)})
```
bhroam: This is fine, but why don't we just submit the reservation as MGR_USER if we want to submit it as a manager.  That way we don't have to add and delete TEST_USER from managers.

arungrover: Well, there was a bug in PBS wherein a user when acting as manager is allowed to submit a maintenance reservation but when the same user is taken off the manager's list, it was still able to delete the submitted reservation.
As you said, it can be done with MGR_USER too, in that case the test will be assuming that the user is already part of managers, it will then remove MGR_USER off the list of managers and then add it back again. I guess, Vaclav wanted the test to be more explicit and self contained.

PR #1355, opened 2019-10-16
PTL tests get skipped if no_mom_on_server is true when requirement decorator is set

shilpakodli:
#### Describe Bug or Feature
Scenario-1:
PTL multi-node tests get skipped if moms are on remote host.
For example: pbs_benchpress -t TestMultiNode -p servers=host1,moms=host2:host3 

Reason is, in requirement decorator, if there is no common host specified in servers and moms then it will check for no_mom_on_server flag is set or not. If it is not set then tests will be skipped.

Scenario-2:
Requirement decorator doesn't handle if we pass PBS confile in -p param "host@pbs_conf_file".
For example, if we pass -pserver=host1@/etc/pbs.conf2 then in requirement decorator, server will be considered as "host@/etc/pbs.conf2"

#### Describe Your Change
Do not skip tests if no_mom_on_server flag is not set in testcase/testsuite requirement decorator and in -p param in "pbs_benchpress" command.
Handle host names if pbs_conf file is passed in -p param list.

Modified code for no_comm_on_server as well as its behavior was also same as no_mom_on_server.

[test_execution_logs.txt](https://github.com/PBSPro/pbspro/files/3732718/test_execution_logs.txt)

### DIFF
--- a/test/fw/ptl/utils/plugins/ptl_test_runner.py
+++ b/test/fw/ptl/utils/plugins/ptl_test_runner.py
@@ -657,29 +657,37 @@ class PTLTestRunner(Plugin):
         tparam_contents = {}
         nomomlist = []
         shortname = (socket.gethostname()).split('.', 1)[0]
-        for key in ['servers', 'moms', 'comms', 'clients']:
+        for key in ['servers', 'moms', 'comms', 'clients', 'nomom']:
             tparam_contents[key] = []
+        tparam_contents['no_mom_on_server'] = False
+        tparam_contents['no_comm_on_server'] = False
+        tparam_contents['no_comm_on_mom'] = False
         if self.param is not None:
             for h in self.param.split(','):
                 if '=' in h:
                     k, v = h.split('=', 1)
+                    hosts = [x.split('@')[0] for x in v.split(':')]
                     if (k == 'server' or k == 'servers'):
-                        tparam_contents['servers'].extend(v.split(':'))
+                        tparam_contents['servers'].extend(hosts)
                     elif (k == 'mom' or k == 'moms'):
-                        tparam_contents['moms'].extend(v.split(':'))
+                        tparam_contents['moms'].extend(hosts)
                     elif k == 'comms':
-                        tparam_contents['comms'] = v.split(':')
+                        tparam_contents['comms'] = hosts
                     elif k == 'client':
-                        tparam_contents['clients'] = v.split(':')
+                        tparam_contents['clients'] = hosts
                     elif k == 'nomom':
-                        nomomlist = v.split(':')
+                        nomomlist = hosts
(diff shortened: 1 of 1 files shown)


test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -734,16 +750,20 @@ def __are_requirements_matching(param_dic=None, test=None):
...
-            if eff_tc_req['no_mom_on_server']:
+            if eff_tc_req['no_mom_on_server'] or \
+               (set(_nomom) - set(_servers)) or \
+               _no_mom_on_server:
                 return False
         else:
-            if not eff_tc_req['no_mom_on_server']:
+            if not eff_tc_req['no_mom_on_server'] or \
+               not _no_mom_on_server \
+               and (set(_nomom) - set(_servers)):
```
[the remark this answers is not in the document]
shilpakodli: Thanks Vincent for pointing. I made changes and tested. Please have review.

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -734,16 +750,19 @@ def __are_requirements_matching(param_dic=None, test=None):
             elif eff_tc_req['min_server_disk'] >= available_sys_disk:
                 return False
         if set(param_dic['moms']) & set(param_dic['servers']):
-            if eff_tc_req['no_mom_on_server']:
+            if eff_tc_req['no_mom_on_server'] or \
+               (set(_nomom) - set(_servers)) or \
+               _no_mom_on_server:
                 return False
         else:
```
vstumpf: Now that I think about it, do we even need this else? Is it really necessary?

shilpakodli: That's right Vincent. I thought the same. Removed else condition and it's working fine without else.

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -687,6 +695,11 @@ def __are_requirements_matching(param_dic=None, test=None):
         ts_requirements = {}
         tc_requirements = {}
         param_count = {}
+        _servers = param_dic['servers']
```
hirenvadalia: small suggestion: create a set here only instead of every place we use these variables.

shilpakodli:
Please find latest execution logs:
[TestPbsExecjobSuspendResume.txt](https://github.com/PBSPro/pbspro/files/3771275/TestPbsExecjobSuspendResume.txt)

vstumpf:
Looks good, you have to resolve some merge conflicts though.

shilpakodli:
@vstumpf I have rebased and squashed commits. Please take this PR forward.

PR #1356, opened 2019-10-16
Fix potential memory corruption in hook alarm signal handler

minghui-liu:
#### Describe Bug or Feature
Hook alarm signal handler could potentially cause memory corruption. It calls `log_event()`, which has a `calloc()` and `calloc()` is not safe to call from within a signal handler. 

#### Describe Your Change
Use stack memory for `log_lock` to instead allocating heap memory.

#### Link to Design Doc
N/A

[SmokeTest.txt](https://github.com/PBSPro/pbspro/files/3735742/SmokeTest.txt)
[TestHookSmokeTest.txt](https://github.com/PBSPro/pbspro/files/3735743/TestHookSmokeTest.txt)

### DIFF
--- a/src/lib/Liblog/pbs_log.c
+++ b/src/lib/Liblog/pbs_log.c
@@ -275,24 +275,16 @@ log_get_tls_data(void)
 int
 log_mutex_lock()
 {
-	struct pbs_log_lock *log_lock = NULL;
-	if ((log_lock = pthread_getspecific(pbs_log_tls_key)) == NULL) {
-		log_lock = calloc(1, sizeof(struct pbs_log_lock));
-		if (!log_lock)
-			return -1;
+	long log_lock;
+	if ((log_lock = (long)pthread_getspecific(pbs_log_tls_key)) != 0)
+		return -1;
 
-		if (pthread_setspecific(pbs_log_tls_key, log_lock) != 0) {
-			free(log_lock);
-			return -1;
-		}
-		log_lock->locked = 0;
-	}
+	if (pthread_mutex_lock(&log_mutex) != 0)
+		return -1;
+	
+	log_lock = 1;
+	pthread_setspecific(pbs_log_tls_key, (void*)log_lock);
 
-	if (log_lock->locked == 0) {
-		if (pthread_mutex_lock(&log_mutex) != 0)
-			return -1;
-	}
-	log_lock->locked++;
 	return 0;
 }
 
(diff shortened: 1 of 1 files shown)


src/lib/Liblog/pbs_log.c
```diff
@@ -275,24 +275,16 @@ log_get_tls_data(void)
...
-		if (!log_lock)
-			return -1;
-
-		if (pthread_setspecific(pbs_log_tls_key, log_lock) != 0) {
-			free(log_lock);
-			return -1;
-		}
-		log_lock->locked = 0;
-	}
+	long log_lock;
```
mike0042: Change this to void *, no need to initialize since we assign it immediately after it is defined. Please don't store pointers in type long, or even unsigned long.

Please remove the definition of struct pbs_log_lock since this is the only place it is used.

src/lib/Liblog/pbs_log.c
```diff
@@ -275,24 +275,16 @@ log_get_tls_data(void)
...
+	if ((log_lock = (long)pthread_getspecific(pbs_log_tls_key)) != 0)
```
mike0042: No need for cast to long once type is changed.

src/lib/Liblog/pbs_log.c
```diff
@@ -275,24 +275,16 @@ log_get_tls_data(void)
...
+		return -1;
+
+	if (pthread_mutex_lock(&log_mutex) != 0)
+		return -1;
+	
+	log_lock = 1;
+	pthread_setspecific(pbs_log_tls_key, (void*)log_lock);
```
mike0042: Add space between void and *

src/lib/Liblog/pbs_log.c
```diff
@@ -313,17 +305,16 @@ log_mutex_lock()
 int
 log_mutex_unlock()
 {
-	struct pbs_log_lock *log_lock = NULL;
-	if ((log_lock = pthread_getspecific(pbs_log_tls_key)) == NULL)
+	long log_lock;
```
mike0042: void *

src/lib/Liblog/pbs_log.c
```diff
@@ -313,17 +305,16 @@ log_mutex_lock()
...
+	if ((log_lock = (long)pthread_getspecific(pbs_log_tls_key)) == 0)
```
mike0042: Remove cast.

src/lib/Liblog/pbs_log.c
```diff
@@ -275,15 +271,15 @@ log_get_tls_data(void)
...
-	long log_lock;
-	if ((log_lock = (long)pthread_getspecific(pbs_log_tls_key)) != 0)
+	void *log_lock;
+	if ((log_lock = pthread_getspecific(pbs_log_tls_key)) != 0)
 		return -1;
 
 	if (pthread_mutex_lock(&log_mutex) != 0)
 		return -1;
 	
 	log_lock = 1;
```
mike0042: I think you can use log_lock = &log_lock here. That should provide a unique non-NULL value.

src/lib/Liblog/pbs_log.c
```diff
@@ -305,15 +301,15 @@ log_mutex_lock()
...
-	long log_lock;
-	if ((log_lock = (long)pthread_getspecific(pbs_log_tls_key)) == 0)
+	void *log_lock;
+	if ((log_lock = pthread_getspecific(pbs_log_tls_key)) == 0)
 		return -1;
 	
 	if (pthread_mutex_unlock(&log_mutex) != 0)
 		return -1;
 
 	log_lock = 0;
```
mike0042: You should use NULL rather than zero here.

PR #1358, opened 2019-10-17
Job executing prologue hook could get stuck in substate=41.

bayucan:
#### Describe Bug or Feature

* A job that is in pre-execution stage (i.e. executing a prologue hook, in substate=41/PRERUN), can be interrupted with qrerun -W force. This causes the server to immediately rerun the job without waiting for an acknowledgement from Mom (it's "forced") that she has killed the job. Server would just send a discard message to the mom to delete the previous version of the job.
* A race condition can occur where Server reruns the job, but then immediately receives an Obit request from the mom as she had killed the original job.Now server  rejects the Obit request from 
 Mom since it had already started a new instance of the job. When Mom receives the server's reject obit reply, it proceeds to delete the job that happens to be the newly instantiated job!
* The end result is a hung job: server knows of the newly instantiated job; mom forgets about the job.
* Server_logs snippet shows server rerunning the job, but then immediately receives an Obit request from mom:
10/15/2019 15:00:24.156178;0008;Server@x100-s15;Job;107.x100-s15;Discard running job, Force rerun
...
10/15/2019 15:00:29.553551;0008;Server@x100-s15;Job;107.x100-s15;Job Run at request of Scheduler@x100-s15.pbspro.com on exec_vnode (x100-s15:ncpus=1)
10/15/2019 15:00:29.564214;0080;Server@x100-s15;Job;107.x100-s15;Obit received momhop:1 serverhop:2 state:4 substate:41
10/15/2019 15:00:29.564246;0001;Server@x100-s15;Svr;Server@x100-s15;Success (0) in job_obit, reject_obit(107.x100-s15): run is for an older run version of the job

* Mom_logs show it has received the reject obit reply and proceeds to forget the job:
10/16/2019 09:14:12.046179;0008;pbs_mom;Job;107.x100-s15;Job removed, Server rejected Obit

#### Describe Your Change

* When mom receives the reject obit reply from server, it calls the function rid_job(). Code has been adjusted to only issue a mom_deljob() (i.e. forget the job) if job is actually running (not in PRERUN state like executing the prologue hook) so as to prevent the race condition. When job is in PRERUN state and is forgotten by Mom, job would never progress at all. 

#### Link to Design Doc

N/A

* Ran: pbs_benchpress -t TestQdel_Wforce_ext.test_qrerun_job_in_prologue_hook 100 times, and on the 55th try, it failed stuck looking for job to advance from substate=41 (PRERUN) to substate=42 (RUNNING):
[ptl.prologue_qrerun.FAIL.txt](https://github.com/PBSPro/pbspro/files/3746287/ptl.prologue_qrerun.FAIL.txt)

* Ran: pbs_benchpress -t TestQdel_Wforce_ext.test_qrerun_job_in_prologue_hook 100 times on the fixed code and it all passed:
[ptl.prologue_qrerun.PASS.txt](https://github.com/PBSPro/pbspro/files/3746288/ptl.prologue_qrerun.PASS.txt)

### DIFF
--- a/src/resmom/catch_child.c
+++ b/src/resmom/catch_child.c
@@ -2735,18 +2735,30 @@ send_sisters_deljob_wait(job *pjob)
  * 	rid_job - rid mom of a job that the server says no longer exists
  *
  * @param[in] jobid - char pointer holding jobid
+ * @param[in] msg - message to log if job has been deleted.
  *
  * @return Void
  *
  */
 void
-rid_job(char *jobid)
+rid_job(char *jobid, char *msg)
 {
 	job	*pjob;
 
 	pjob = find_job(jobid);
-	if (pjob && !pjob->ji_hook_running_bg_on)
+	if (pjob && !pjob->ji_hook_running_bg_on && (pjob->ji_qs.ji_substate != JOB_SUBSTATE_PRERUN)) {
+		/* Allowing only to rid a job that has actually started
+		   (i.e. not in JOB_SUBSTATE_PRERUN), avoids the race condition where
+		   a job that is just executing a prologue hook (in PRERUN state)
+		   gets signaled to qrerun -Wforce, causing server to immediately
+		   rerun the job, ahead of the Obit sent by mom, which is expectedly
+		   rejected by the server, but has a side effect of the Mom deleting
+		   its copy of the job due to the rejected obit. And this results in
+		   a hung job: server knows about the job but mom forgets it!
+		*/
+		log_event(PBSEVENT_JOB, PBS_EVENTCLASS_JOB, LOG_NOTICE, jobid, msg);
 		mom_deljob(pjob);
+	}
 }
 
 /**
--- a/src/resmom/catch_child.c
+++ b/src/resmom/catch_child.c
@@ -2748,15 +2748,18 @@ rid_job(char *jobid, char *msg)
 	pjob = find_job(jobid);
 	if (pjob && !pjob->ji_hook_running_bg_on && (pjob->ji_qs.ji_substate != JOB_SUBSTATE_PRERUN)) {
 		/* Allowing only to rid a job that has actually started
-		   (i.e. not in JOB_SUBSTATE_PRERUN), avoids the race condition where
-		   a job that is just executing a prologue hook (in PRERUN state)
-		   gets signaled to qrerun -Wforce, causing server to immediately
-		   rerun the job, ahead of the Obit sent by mom, which is expectedly
-		   rejected by the server, but has a side effect of the Mom deleting
-		   its copy of the job due to the rejected obit. And this results in
-		   a hung job: server knows about the job but mom forgets it!
-		*/
-		log_event(PBSEVENT_JOB, PBS_EVENTCLASS_JOB, LOG_NOTICE, jobid, msg);
+		 * (i.e. not in JOB_SUBSTATE_PRERUN), would avoid the race condition
+		 * resulting in a hung job: server force reruns a job which
+		 * is lingering in PRERUN state, and an Obit request for
+		 * the previous instance of the job is received by the server and
+		 * rejected, causing mom to delete the new instance of the
+		 * job.
+		 * If the job has passed the PRERUN stage, then it would have already
+		 * synced up with the server on status, and not end up in this
+		 * race condition.
+		 */
+		if (msg != NULL)
+			log_event(PBSEVENT_JOB, PBS_EVENTCLASS_JOB, LOG_NOTICE, jobid, msg);
 		mom_deljob(pjob);
 	}
 }
--- a/src/resmom/catch_child.c
+++ b/src/resmom/catch_child.c
@@ -2730,39 +2730,6 @@ send_sisters_deljob_wait(job *pjob)
 		return 0;
 
 }
-/**
- * @brief
- * 	rid_job - rid mom of a job that the server says no longer exists
- *
- * @param[in] jobid - char pointer holding jobid
- * @param[in] msg - message to log if job has been deleted.
- *
- * @return Void
- *
- */
-void
-rid_job(char *jobid, char *msg)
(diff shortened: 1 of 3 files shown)


src/resmom/catch_child.c
```diff
@@ -2735,18 +2735,30 @@ send_sisters_deljob_wait(job *pjob)
...
+		/* Allowing only to rid a job that has actually started
+		   (i.e. not in JOB_SUBSTATE_PRERUN), avoids the race condition where
+		   a job that is just executing a prologue hook (in PRERUN state)
+		   gets signaled to qrerun -Wforce, causing server to immediately
+		   rerun the job, ahead of the Obit sent by mom, which is expectedly
+		   rejected by the server, but has a side effect of the Mom deleting
+		   its copy of the job due to the rejected obit. And this results in
+		   a hung job: server knows about the job but mom forgets it!
+		*/
+		log_event(PBSEVENT_JOB, PBS_EVENTCLASS_JOB, LOG_NOTICE, jobid, msg);
```
mike0042: Please check for msg != NULL before this line.

src/resmom/catch_child.c
```diff
@@ -2735,18 +2735,30 @@ send_sisters_deljob_wait(job *pjob)
...
+	if (pjob && !pjob->ji_hook_running_bg_on && (pjob->ji_qs.ji_substate != JOB_SUBSTATE_PRERUN)) {
+		/* Allowing only to rid a job that has actually started
+		   (i.e. not in JOB_SUBSTATE_PRERUN), avoids the race condition where
+		   a job that is just executing a prologue hook (in PRERUN state)
+		   gets signaled to qrerun -Wforce, causing server to immediately
+		   rerun the job, ahead of the Obit sent by mom, which is expectedly
+		   rejected by the server, but has a side effect of the Mom deleting
+		   its copy of the job due to the rejected obit. And this results in
+		   a hung job: server knows about the job but mom forgets it!
+		*/
```
mike0042: Standard multi-line comment format:
```
/*
 * line 1
 * line 2
 */
```
If you could try to wordsmith this comment a bit. Most of it is a run-on sentence.

src/resmom/catch_child.c
```diff
@@ -2735,18 +2735,30 @@ send_sisters_deljob_wait(job *pjob)
...
  */
 void
-rid_job(char *jobid)
+rid_job(char *jobid, char *msg)
 {
 	job	*pjob;
 
 	pjob = find_job(jobid);
-	if (pjob && !pjob->ji_hook_running_bg_on)
+	if (pjob && !pjob->ji_hook_running_bg_on && (pjob->ji_qs.ji_substate != JOB_SUBSTATE_PRERUN)) {
```
mike0042: Just so I am clear... if the job has started, it already sent an update to the server so there is no longer a race condition in that case, correct? If so, you might want to mention that in the comment.

src/resmom/catch_child.c
bayucan: Yes, that should be done.

src/resmom/catch_child.c
bayucan: Correct. I'll add to the comment.

src/resmom/catch_child.c
```diff
@@ -2735,18 +2735,33 @@ send_sisters_deljob_wait(job *pjob)
...
  *
  * @param[in] jobid - char pointer holding jobid
+ * @param[in] msg - message to log if job has been deleted.
  *
  * @return Void
  *
  */
 void
-rid_job(char *jobid)
+rid_job(char *jobid, char *msg)
```
bhroam: I'd rather not pass in a message that is only used for printing.  I'd rather the function return a value that says if it deleted the job or not.  This way the caller can use that to print or not.

bayucan: @bhroam Thanks for your suggestion. But I feel the code is fine as is. The function is called only once, and I just try to match how it was originally invoked.  The original invocation didn't really care whether the job got deleted or not, but strangely just assumed it did, and the message was always printed. If someone else wants to use the function, and would like something done besides logging a message when job is deleted, then the function can be improved...

mike0042: Please let us know if you feel strongly about this @bhroam.

arungrover: If the function is very scenario-specific and is called nowhere else then is there really a need to pass in a message? function can log it by itself, right?

bayucan: @arungrover : The original code in src/resmom/mom_server.c:is_request() under case BAD_OBIT:

log_event(..., ""Job removed, Server rejected Obit")
...
rid(jobid)

Now if I put inside rid(jobid) the "Job removed, Server reject Obit" hardwired message, it would look odd.

arungrover: Thanks Al! I got it.
IMHO (not suggesting you change), there is not much of a need for this function itself because it is small and there is nowhere else we call it from. 
I understand that you want to keep the function generic by passing in the message but I feel the function is already very specific to the case when no background hook is running for the job, it is not in prerun state.

mike0042: I'm not generally a fan of increasing the size of a function (is_request) that is already hundreds of lines long, but we're only talking a couple lines. @arungrover is correct in that this is the only place rid_job() gets called, so it doesn't make a lot of sense to keep it a function. Especially now that we are specializing the conditions for which the job gets deleted. We also get to remove the external definition from the header file. While there is nothing wrong with the changes you propose @bayucan, I tend to agree with @arungrover here. That should also address the concerns from @bhroam.

bayucan: @mike0042 @arungrover @bhroam : Ok then, I'll work on updating this fix, removing the rid_job() function and putting its content in line under case BAD_OBIT in is_request().

bayucan:
@mike0042 @hirenvadalia : Thanks for the review and merge!

PR #1361, opened 2019-10-17
Fix two test case bugs

sujatapatnaik52:
#### Describe Bug or Feature
* Most of the test cases in Testsuite pbs_test_entity_limits.py uses TST_USR. 
* TestMultipleSchedulers.test_auto_sched_off_due_to_fds_limit sets the limit of open files to 10, then reverts it to its original at the end of the test. However, if the test fails between them, it does not revert the limit of open files and all the future tests fail.

#### Describe Your Change
1. Modifed the test user TST_USR to TEST_USER in pbs_test_entity_limits.py to make it generic.
2. Added teardown to the TestMultipleSchedulers to reset the rlimit even if the test fails. 

[sched_test_case_logs.txt](https://github.com/PBSPro/pbspro/files/3739831/sched_test_case_logs.txt)
[TestEntityLimits_test_logs.txt](https://github.com/PBSPro/pbspro/files/3739836/TestEntityLimits_test_logs.txt)

### DIFF
--- a/test/tests/functional/pbs_multi_sched.py
+++ b/test/tests/functional/pbs_multi_sched.py
@@ -2089,4 +2089,3 @@ class TestMultipleSchedulers(TestFunctional):
         self.server.manager(MGR_CMD_SET, SCHED,
                             {'scheduling': 'True'}, id='sc2')
         self.server.log_match("processing priority socket", starttime=t)
-
--- a/test/tests/functional/pbs_multi_sched.py
+++ b/test/tests/functional/pbs_multi_sched.py
@@ -45,6 +45,10 @@ class TestMultipleSchedulers(TestFunctional):
     """
     Test suite to test different scheduler interfaces
     """
+    rlimit_flag = 0
+
+    open_files_soft_limit = ''
+    open_files_hard_limit = ''
 
     def setup_sc1(self):
         a = {'partition': 'P1,P4',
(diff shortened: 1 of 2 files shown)


test/tests/functional/pbs_multi_sched.py
```diff
@@ -2084,3 +2083,16 @@ def test_multi_sched_priority_sockets(self):
         self.server.manager(MGR_CMD_SET, SCHED,
                             {'scheduling': 'True'}, id='sc2')
         self.server.log_match("processing priority socket", starttime=t)
+
+    def tearDown(self):
+        if TestMultipleSchedulers.rlimit_flag == 1:
```
prakashcv13: instead of writing a tearDown() which is effective only for one test (at this time) and we need to use class variables to do that, how about handling this in the test case itself, where regardless of the test result, we revert the limits, maybe by using PtlExpectError.

sujatapatnaik52: Hi @prakashcv13, If I am not wrong, PtlExpectError is for catching the expect statements, after setting the rlimit, the test can fail at any point. I agree with you regarding having teardown is kind of trivial for now. I am just thinking to add try...except...finally for rest of the test case after the limit is set. How does that sound?

prakashcv13: you are right, I just assumed that the test case would fail at the except(). Your idea sounds good to me.

sujatapatnaik52: @prakashcv13 I have addressed your suggestions. Please look at the latest commit.

PR #1362, opened 2019-10-17
Hook with config file configured is unable to access os.environ object

npadole20:
#### Describe Bug or Feature

If a hook has a config file configured and we try accessing os.environ in the hook, we get "'str' object has no attribute 'decode'" error.

#### Describe Your Change
We were setting environment variables to os.environ.data in Python 2 which changed to _data in Python 3. As part of this fix, we will set environment variables in os.environ instead of _data.

#### Link to Design Doc

NA

[test_hook_config_os_env.txt](https://github.com/PBSPro/pbspro/files/3740734/test_hook_config_os_env.txt)
[test_hook_config_os_env_bf.txt](https://github.com/PBSPro/pbspro/files/3740735/test_hook_config_os_env_bf.txt)
[os_env_testlog.txt](https://github.com/PBSPro/pbspro/files/3740743/os_env_testlog.txt)

### DIFF
--- a/src/lib/Libpython/pbs_python_svr_internal.c
+++ b/src/lib/Libpython/pbs_python_svr_internal.c
@@ -11778,9 +11778,9 @@ int
 pbs_python_set_os_environ(char *env_var, char *env_val)
 {
 	PyObject *pystr_env_val = NULL;
+	PyObject *pystr_env_var = NULL;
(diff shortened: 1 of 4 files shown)


mike0042:
There is a test repeatedly failing under TravisCI.

PR #1363, opened 2019-10-21
[WIP] TestMomDynRes suite fails if mom is on remote host

shilpakodli:
#### Describe Bug or Feature
TestMomDynRes test suite fails to run if mom is on remote host.
TestMomDynRes test suite creates a mom_dyn_resource file on local host and adds path of this file inside mom_priv/config file. If moms are on remote hosts, this file will not be present and resource value will be null.

#### Describe Your Change
Copy resource file to remote host and delete when done with it.

#### Attach Test Logs
[TestMomDynRes_logs.txt](https://github.com/PBSPro/pbspro/files/3749676/TestMomDynRes_logs.txt)

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -13832,10 +13832,8 @@ class MoM(PBSService):
         else:
             if dirname is None:
                 dirname = self.pbs_conf['PBS_HOME']
-            local_host = socket.gethostname()
             tmp_file = self.du.create_temp_file(prefix=prefix, suffix=suffix,
-                                                body=script_body,
-                                                hostname=local_host)
+                                                body=script_body)
             res_file = os.path.join(dirname, tmp_file.split(os.path.sep)[-1])
             self.du.run_copy(host, tmp_file, res_file, sudo=True,
                              preserve_permission=False)
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -2397,10 +2397,6 @@ class BatchUtils(object):
             if isinstance(v, list):
                 v = ','.join(v)
 
-            if isinstance(v, str):
-                if " " in v:
-                    v = "'" + v + "'"
-
             # when issuing remote commands, escape spaces in attribute values
             if (((hostname is not None) and
                  (not self.du.is_localhost(hostname))) or
@@ -6037,6 +6033,13 @@ class Server(PBSService):
                     del obj.custom_attrs[ATTR_queue]
 
             _conf = self.default_client_pbs_conf
+            user_host = PbsUser.get_user(obj.username).host
+            if (not self._is_local) or \
+                    (not self.du.is_localhost(user_host)):
+                for k, v in obj.custom_attrs.items():
+                    if isinstance(v, str) and (" " in v):
+                        v = "'" + v + "'"
+                        obj.custom_attrs[k] = v
             cmd = self.utils.convert_to_cli(obj.custom_attrs, IFL_SUBMIT,
                                             self.hostname, dflt_conf=_conf,
                                             exclude_attrs=exclude_attrs)
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -13832,10 +13832,10 @@ class MoM(PBSService):
         else:
             if dirname is None:
                 dirname = self.pbs_conf['PBS_HOME']
+            local_host = socket.gethostname()
             tmp_file = self.du.create_temp_file(prefix=prefix, suffix=suffix,
                                                 body=script_body,
-                                                hostname=host)
-
+                                                hostname=local_host)
             res_file = os.path.join(dirname, tmp_file.split(os.path.sep)[-1])
             self.du.run_copy(host, tmp_file, res_file, sudo=True,
                              preserve_permission=False)
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -6037,8 +6037,14 @@ class Server(PBSService):
             if (not self._is_local) or \
                     (not self.du.is_localhost(user_host)):
                 for k, v in obj.custom_attrs.items():
-                    if isinstance(v, str) and (" " in v):
-                        v = "'" + v + "'"
+                    if isinstance(v, str):
+                        if " " in v:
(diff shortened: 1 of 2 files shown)


test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -13832,10 +13832,10 @@ def add_mom_dyn_res(self, custom_resource, script_body=None,
         else:
             if dirname is None:
                 dirname = self.pbs_conf['PBS_HOME']
+            local_host = socket.gethostname()
```
hirenvadalia: do we need to pass localhost in create_temp_file? I think localhost is default right?

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -13832,10 +13832,10 @@ def add_mom_dyn_res(self, custom_resource, script_body=None,
...
             tmp_file = self.du.create_temp_file(prefix=prefix, suffix=suffix,
                                                 body=script_body,
-                                                hostname=host)
-
+                                                hostname=local_host)
             res_file = os.path.join(dirname, tmp_file.split(os.path.sep)[-1])
             self.du.run_copy(host, tmp_file, res_file, sudo=True,
```
hirenvadalia: cp command will fail if given `host` is localhost:
```
$ cp /tmp/1 /tmp/1
cp: '/tmp/1' and '/tmp/1' are the same file
$ echo $?
1
```

test/tests/functional/pbs_mom_dynamic_resource.py
```diff
@@ -162,7 +164,7 @@ def test_res_string_perticular_value(self):
                            id=jid, attrop=PTL_AND)
 
         # Submit a job that requests mom dynamic resource
-        attr = {"Resource_List." + resc_name[0]: '"This is a test"'}
+        attr = {"Resource_List." + resc_name[0]: '\'\"This is a test\"\''}
```
hirenvadalia: why do we need to add single quotes in the string?

shilpakodli: This is generic change. If we have to pass a string with spaces in custom resources, format should be "qsub -l foo='\"This is a test\"' -- /bin/sleep 100" according to documentation.

hirenvadalia: oh ok, then I think we should fix PTL lib to add quotes instead user do it in tests.

test/fw/ptl/lib/pbs_testlib.py
shilpakodli: It doesn't fail because create_teno_file will create file in /tmp/ and in cp command we are copying to either PBS_HOME directory user given directory.

test/fw/ptl/lib/pbs_testlib.py
shilpakodli: Removed host in create_temp_file call as we always want to create this file on local host by default and copy to remote host or to the location on localhost where user specifies or to PBS_HOME/

test/tests/functional/pbs_mom_dynamic_resource.py
shilpakodli: @hirenvadalia I have made changes in PTL library to handle strings having spaces.

shilpakodli:
@hirenvadalia and @vstumpf Please re-review the code as I have made changes in "submit" function itself.

test/tests/functional/pbs_mom_dynamic_resource.py
```diff
@@ -438,4 +441,8 @@ def tearDown(self):
             self.du.rm(path=self.dirnames, sudo=True, force=True,
                        recursive=True)
             self.dirnames[:] = []
+        if len(self.resc_files) != 0:
```
vstumpf: This can be simplified to
```suggestion
        if self.resc_files:
```

shilpakodli:
Attaching latest logs with latest code change:
[TestMomDynRes_latest.txt](https://github.com/PBSPro/pbspro/files/3813669/TestMomDynRes_latest.txt)

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -6033,6 +6033,13 @@ def submit(self, obj, script=None, extend=None, submit_dir=None):
                     del obj.custom_attrs[ATTR_queue]
 
             _conf = self.default_client_pbs_conf
+            user_host = PbsUser.get_user(obj.username).host
+            if (not self._is_local) or \
+                    (not self.du.is_localhost(user_host)):
+                for k, v in obj.custom_attrs.items():
+                    if isinstance(v, str) and (" " in v):
```
hirenvadalia: what if v has `"` or `'`? don't we want to put under quotes that value also?

shilpakodli: @hirenvadalia I have tried with strings having 'or " in it("This'is'test" and "This"is"test") but it doesn't have any effect as in string is submitted as "Thisistest" which doesn;t have any single quotes.

hirenvadalia: @shilpakodli Assuming you are talking about running qsub command from the shell, not PTL, I think you missed to escape those quotes while running qsub command in the shell.

shilpakodli:
@hirenvadalia I have addressed your comments. Please review.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -6037,8 +6037,14 @@ def submit(self, obj, script=None, extend=None, submit_dir=None):
...
                     (not self.du.is_localhost(user_host)):
                 for k, v in obj.custom_attrs.items():
-                    if isinstance(v, str) and (" " in v):
-                        v = "'" + v + "'"
+                    if isinstance(v, str):
+                        if " " in v:
+                            v = "'%s'" % v
+                        if any((c in v) for c in set(',\'"')):
+                            if '"' in v:
+                                v = "\\'%s\\'" % v
```
hirenvadalia: why do we need to add `\` here?
Also, I think we should move this fix to `BatchUtils.convert_to_cli` (see https://github.com/PBSPro/pbspro/blob/master/test/fw/ptl/lib/pbs_testlib.py#L2405)

shilpakodli: I have moved this part of code to convert_to_cli function.
We need \ to escape single quote in set of characters to match.

agrawalravi90:
Closing because this PR has been inactive for a long time, please re-open when you are ready with your changes.

PR #1365, opened 2019-10-21
Added new accounting records

bhroam:
#### Describe Bug or Feature
Added new alter accounting log record and extended the Q record.

#### Describe Your Change
Added a new accounting log record when an attribute is modified.  It happens when someone uses qalter, or a server hook modifies an attribute.

Also added extra job based information to the 'Q' record including dependencies.  
#### Link to Design Doc
https://pbspro.atlassian.net/wiki/spaces/PD/pages/1363443719/Enhancing+accounting+logs+for+job+lifecycle

[acct.log](https://github.com/PBSPro/pbspro/files/3753281/acct.log)

### DIFF
--- a/src/include/acct.h
+++ b/src/include/acct.h
@@ -94,7 +94,7 @@ extern void account_jobstr(job *pjob);
 extern void account_jobstr2(job *pjob, int type);
 extern void account_job_update(job *pjob, int type);
 extern void account_jobend(job *pjob, char * used, int type);
-
+extern void log_alter_records_for_attrs(job *pjob, svrattrl *plist);
 extern void set_job_ProvAcctRcd(job *pjob, long time_se, int type);
 
 extern int concat_rescused_to_buffer(char **buffer, int *buffer_size, svrattrl *patlist, char *delim, job *pjob);
--- a/src/include/acct.h
+++ b/src/include/acct.h
@@ -65,6 +65,7 @@ extern "C" {
 #define PBS_ACCT_UPDATE	(int)'u'	/* phased job update record */
 #define PBS_ACCT_NEXT	(int)'c'	/* phased job next record */
 #define PBS_ACCT_LAST	(int)'e'	/* phased job last usage record */
+#define PBS_ACCT_ALTER  (int)'a'	/* Job attribute is being altered */
 
 /* for RESERVATION accounting */
 
--- a/src/server/accounting.c
+++ b/src/server/accounting.c
@@ -2240,8 +2240,18 @@ void log_alter_records_for_attrs(job *pjob, svrattrl *plist) {
 	svrattrl *cur_svr;
 	pbs_list_head phead;
 	svrattrl *cur_plist;
-	char *logstr = NULL;
+	char *per_alter_buf = NULL;
+	char *entire_record = NULL;
+	int entire_record_len = 0;
+
 	int i;
+	entire_record = malloc(1024 * sizeof(char));
+	if (entire_record == NULL)
+		return;
+	entire_record_len = 1024;
+	entire_record[0] = '\0';
+	int error = 0;
+
 	CLEAR_HEAD(phead);
 	for (i = 0; i < JOB_ATR_LAST; i++) {
 		if (pjob->ji_wattr[i].at_flags & ATR_VFLAG_MODIFY) {
@@ -2259,7 +2269,7 @@ void log_alter_records_for_attrs(job *pjob, svrattrl *plist) {
 										fmt = "%s.%s=%s";
 									else
 										fmt = "%s.%s=\"%s\"";
-									pbs_asprintf(&logstr, fmt, cur_svr->al_name, cur_svr->al_resc, cur_svr->al_value);
+									pbs_asprintf(&per_alter_buf, fmt, cur_svr->al_name, cur_svr->al_resc, cur_svr->al_value);
 									break;
 								}
 							}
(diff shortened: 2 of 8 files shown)


src/server/hook_func.c
```diff
@@ -3148,6 +3152,9 @@ set_job_reslist(job *pjob, char *hook_name, char *msg, int msg_len,
 		log_event(PBSEVENT_JOB, PBS_EVENTCLASS_JOB, LOG_INFO,
 			pjob->ji_qs.ji_jobid, msg);
 
+		snprintf(log_buffer, sizeof(log_buffer), "%s.%s=%s", ATTR_l, resc, new_rescval_str);
+		account_record(PBS_ACCT_ALTER, pjob, log_buffer);
```
agrawalravi90: can there be multiple ATTR_l being set ? If so, wouldn't it be better to just call account_record() once for all of them?

src/server/req_modify.c
```diff
@@ -387,6 +388,48 @@ req_modifyjob(struct batch_request *preq)
 		return;
 	}
 
+	if(find_sched_from_sock(preq->rq_conn) == NULL)
```
agrawalravi90: nit: add a space after if

src/server/req_modify.c
```diff
@@ -387,6 +388,48 @@ req_modifyjob(struct batch_request *preq)
...
+							else {
+								pbs_asprintf(&logstr, "%s=%s", cur_svr->al_name, cur_svr->al_value);
+								break;
+							}
+						}
+					}
+					if (logstr == NULL && cur_plist->al_value[0] == '\0') { /* unset */
+						pbs_asprintf(&logstr, "%s%s%s=UNSET", cur_plist->al_name, cur_plist->al_resc ? "." : "", cur_plist->al_resc ? cur_plist->al_resc : "");
+					}
+					account_record(PBS_ACCT_ALTER, pjob, logstr);
```
agrawalravi90: Won't this print a record for each attribute being modified? Is that what we want?

bhroam: Yes and Yes.  There should be no difference between modifying two attributes in 2 qalters or in one.  I described this in the design document and just clarified it further today.

src/server/hook_func.c
bhroam: As I said below.  There should be one record per attribute or resource.

src/server/accounting.c
```diff
@@ -554,29 +555,45 @@ acct_job(job *pjob, int type, char *buf, int len)
...
-	pb  += i;
+	pb += i;
 	len -= i;
 
-	/* start time */
-	(void)sprintf(pb, "start=%ld ", (long)pjob->ji_qs.ji_stime);
-	i = strlen(pb);
-	pb  += i;
-	len -= i;
+	if (type != PBS_ACCT_QUEUE) {
```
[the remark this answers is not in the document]
bhroam: Interesting, it's also used for the 'S' and and 's' and other similar records

src/server/accounting.c
```diff
@@ -307,6 +307,7 @@ cpy_quote_value(char *pb, char *value)
 #define	ARRAY_INDICES_FMT	"array_indices=%s "
 #define	EXEC_HOST_FMT		"exec_host=%s "
 #define	EXEC_VNODE_FMT		"exec_vnode=%s "
+#define DEPEND_FMT		"depend=%s "
```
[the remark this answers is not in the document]
bhroam: argh, I caught this and fixed it... but didn't commit it.

src/include/acct.h
```diff
@@ -65,6 +65,7 @@ extern "C" {
 #define PBS_ACCT_UPDATE	(int)'u'	/* phased job update record */
 #define PBS_ACCT_NEXT	(int)'c'	/* phased job next record */
 #define PBS_ACCT_LAST	(int)'e'	/* phased job last usage record */
+#define PBS_ACCT_ALTER  (int)'a'	/* Job attribute is being altered */
```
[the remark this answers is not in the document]
bhroam: It's attribute singular since there will be one record per change

src/server/req_modify.c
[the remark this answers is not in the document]
bhroam: I can make the design document more explicit, but it does say 'Whenever a job attribute is altered, an 'a' record will be emitted to the accounting log.'  To me this says there will be one 'a' record printed per attribute.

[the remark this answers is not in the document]
arungrover: I think a more explicit way would be that whenever a job is altered there will be an ‘a’ record logged for each modified attribute.
I do feel this will make logs too verbose but I guess that’s a forum discussion at this point.

src/server/accounting.c
```diff
@@ -2240,8 +2240,18 @@ void log_alter_records_for_attrs(job *pjob, svrattrl *plist) {
 	svrattrl *cur_svr;
 	pbs_list_head phead;
 	svrattrl *cur_plist;
-	char *logstr = NULL;
+	char *per_alter_buf = NULL;
```
agrawalravi90: how about 'per_attr_buf' since this buffer is really per attribute?

src/server/accounting.c
```diff
@@ -2270,19 +2280,33 @@ void log_alter_records_for_attrs(job *pjob, svrattrl *plist) {
...
+					return;
 				}
-				account_record(PBS_ACCT_ALTER, pjob, logstr);
-				free(logstr);
-				logstr = NULL;
 			}
 			free_svrattrl(svrattrl_list);
 		}
 	}
+	account_record(PBS_ACCT_ALTER, pjob, entire_record);
```
agrawalravi90: do you wanna check if entire_record[0] == '\0'  before calling account_record() ?

src/server/accounting.c
```diff
@@ -2201,3 +2222,91 @@ account_job_update(job *pjob, int type)
...
+				if (strcmp(cur_plist->al_name, job_attr_def[i].at_name) != 0)
+					continue;
+				else {
+					for (cur_svr = svrattrl_list; cur_svr != NULL; cur_svr = (svrattrl *)GET_NEXT(cur_svr->al_link)) {
+						if (pjob->ji_wattr[i].at_type == ATR_TYPE_RESC) {
+							if (cur_plist->al_resc != NULL) {
+								if (strcmp(cur_plist->al_resc, cur_svr->al_resc) == 0) {
+									char *fmt;
+									if (strchr(cur_svr->al_value, ' ') == NULL)
+										fmt = "%s.%s=%s";
```
arungrover: can you add a space in the beginning of the format so that you don't have to do pbs_strcat for adding a space later on? Of course, you might have to keep track of not having space when first key-value pair is added to log. I feel it will perform better that way.

bhroam: There isn't a really good way to do this.  I don't want the space in the for the first record.  This means I'd have to do an if/else for each pbs_asprintf() (or an ugly ? : ) case to either add the space or not.  It is highly unlikely the pbs_strcat() will cause an allocation.  I allocate 1024 bytes.  That'll be enough space for most log records.  I'd rather leave this alone.

src/server/accounting.c
```diff
@@ -2280,25 +2282,25 @@ void log_alter_records_for_attrs(job *pjob, svrattrl *plist) {
...
-					if (pbs_strcat(&entire_record, &entire_record_len, per_alter_buf) == NULL)
+					if (pbs_strcat(&entire_record, &entire_record_len, per_attr_buf) == NULL)
 						error = 1;
 
-				free(per_alter_buf);
-				per_alter_buf = NULL;
+				free(per_attr_buf);
+				per_attr_buf = NULL;
 				if (error) {
 					free(entire_record);
```
arungrover: This free is not needed now that the pointer is static.

src/server/accounting.c
```diff
@@ -2307,6 +2309,7 @@ void log_alter_records_for_attrs(job *pjob, svrattrl *plist) {
 			free_svrattrl(svrattrl_list);
 		}
 	}
-	account_record(PBS_ACCT_ALTER, pjob, entire_record);
+	if (entire_record[0] != '\0')
+		account_record(PBS_ACCT_ALTER, pjob, entire_record);
 	free(entire_record);
```
arungrover: same holds true for this call to free

bhroam:
@arungrover I've been running with Valgrind.  It's somewhat a messy report, so I missed the free() issue previously.  I just double checked it.  It's clean.

bhroam:
Here's the valgrind log
[valgrind.log](https://github.com/PBSPro/pbspro/files/3785015/valgrind.log)

PR #1367, opened 2019-10-22
changed default ram 1 gb to run ptl test

birbalsain:
#### Describe Bug or Feature

The default ram requirement to run ptl test is 0.128GB. Even pbs_benchpress needs ~200MB to run. With 0.128  GB , We can't run simple job on mom also. Requirement of ram to run ptl test should be more than 1 GB.

#### Describe Your Change

Changed default ram requirement to 1 GB

[CHANGED_RAM_SIZE.txt](https://github.com/PBSPro/pbspro/files/3754333/CHANGED_RAM_SIZE.txt)


birbalsain:
It is still under discussion.

PR #1368, opened 2019-10-22
Improve process and system monitoring for tests in PTL

birbalsain:
#### Describe Bug or Feature

Monitor processes provided as input in procmon with ps command and sar command and store the results to JSON output file.
More information needed of monitoring pbs process and system for load/stress tests

#### Describe Your Change

Added few more sensors to monitor for ps command added info from sar command and system load average.
The monitoring output is stored to proc_monitor.json file on a perticular provided interval and stored in ptl_test_results.json file at the end of test.

Json file during test runnig
[proc_monitor_json.txt](https://github.com/PBSPro/pbspro/files/3966367/proc_monitor_json.txt)

ptl output
[ptl_output_file.txt](https://github.com/PBSPro/pbspro/files/3938089/ptl_output_file.txt)

json file at the test end
[ptl_test_result_json_file.txt](https://github.com/PBSPro/pbspro/files/3938091/ptl_test_result_json_file.txt)

### DIFF
--- a/test/fw/ptl/utils/pbs_procutils.py
+++ b/test/fw/ptl/utils/pbs_procutils.py
@@ -43,7 +43,6 @@ import logging
 import socket
 import os
 import json
-from unittest.mock import sentinel
 from ptl.utils.pbs_dshutils import DshUtils
 
 
@@ -249,9 +248,11 @@ class ProcInfo(object):
         self.command = None
 
     def __str__(self):
-        return (f"{self.name} pid: {self.pid} rss: {self.rss} vsz: "
-                f"{self.vsz} pcpu: {self.pcpu} pmem: {self.pmem} size: "
-                f"{self.size} cputime: {self.cputime} command: {self.command}")
+        return "%s pid: %s rss: %s vsz: %s pcpu: %s pmem: %s \
+               size: %s cputime: %s command: %s" % \
+               (self.name, str(self.pid), str(self.rss), str(self.vsz),
+                str(self.pcpu), str(self.pmem), str(self.size),
+                str(self.cputime), self.command)
 
 
 class ProcMonitor(threading.Thread):
@@ -269,7 +270,6 @@ class ProcMonitor(threading.Thread):
         self._pu = ProcUtils()
         self.stop_thread = threading.Event()
         self.db_proc_info = []
-        self.sysstat = {}
 
     def set_frequency(self, value=60):
         """
@@ -286,6 +286,7 @@ class ProcMonitor(threading.Thread):
         Run system monitoring
         """
         timenow = int(time.time())
+        sysstat = {}
         # if no protocols set, use default
         if not nw_protocols:
             nw_protocols = ['TCP']
@@ -296,11 +297,11 @@ class ProcMonitor(threading.Thread):
         op = rv['out'][2:]
         op = [i.split()[2:] for i in op if
               (i and not i.startswith('Average'))]
-        self.sysstat['name'] = "System"
-        self.sysstat['time'] = time.ctime(timenow)
+        sysstat['name'] = "System"
+        sysstat['time'] = time.ctime(timenow)
         for i in range(0, len(op), 2):
-            self.sysstat.update(dict(zip(op[i], op[i + 1])))
-        return self.sysstat
+            sysstat.update(dict(zip(op[i], op[i + 1])))
+        return sysstat
 
     def run(self):
         """
--- a/test/fw/ptl/utils/pbs_procutils.py
+++ b/test/fw/ptl/utils/pbs_procutils.py
@@ -43,6 +43,7 @@ import logging
 import socket
 import os
 import json
+from unittest.mock import sentinel
 from ptl.utils.pbs_dshutils import DshUtils
 
 
@@ -248,11 +249,9 @@ class ProcInfo(object):
         self.command = None
 
     def __str__(self):
-        return "%s pid: %s rss: %s vsz: %s pcpu: %s pmem: %s \
-               size: %s cputime: %s command: %s" % \
-               (self.name, str(self.pid), str(self.rss), str(self.vsz),
-                str(self.pcpu), str(self.pmem), str(self.size),
-                str(self.cputime), self.command)
+        return (f"{self.name} pid: {self.pid} rss: {self.rss} vsz: "
+                f"{self.vsz} pcpu: {self.pcpu} pmem: {self.pmem} size: "
+                f"{self.size} cputime: {self.cputime} command: {self.command}")
 
 
 class ProcMonitor(threading.Thread):
@@ -282,12 +281,19 @@ class ProcMonitor(threading.Thread):
         self.logger.debug('procmonitor: set frequency to ' + str(value))
         self.frequency = value
 
-    def get_system_stats(self, nw_protocols=['TCP']):
+    def get_system_stats(self, nw_protocols=sentinel):
         """
         Run system monitoring
         """
+        if nw_protocols is sentinel:
+            self.skipTest("network protocol values is not set.")
         cmd = 'sar -rSub -n %s 1 1' % ','.join(nw_protocols)
-        rv = self.du.run_cmd(cmd=cmd, as_script=True)
+        try:
+            rv = self.du.run_cmd(cmd=cmd, as_script=True)
+        except BaseException:
+            msg = f"Is sar package installed"
+            self.logger.error(msg)
+            self.skipTest(msg)
         op = rv['out'][2:]
         op = [i.split()[2:] for i in op if
               (i and not i.startswith('Average'))]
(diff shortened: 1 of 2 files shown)


agrawalravi90:
Since this is about enhancing process & system monitoring via PTL, have we considered using the psutil module (https://pypi.org/project/psutil/) instead of writing our own code to monitor using system commands like ps ? It supports multiple platforms (including Windows) and seems to provide a very easy to use API to do process and system monitoring.

birbalsain:
@agrawalravi90 psutil doesn't work for remote host.https://github.com/giampaolo/psutil/issues/336

birbalsain:
@agrawalravi90 psutils doesn't work for remote host.

agrawalravi90:
> @agrawalravi90 psutil doesn't work for remote host.

Even ps or sar don't work for remote host right? We use run_cmd() to ssh and run them on the remote host. If psutils is better, then it's easy enough to just run a Python script on a remote host which contains psutil based monitoring code, right?

birbalsain:
@agrawalravi90 I added three more parameters to the existing code. We will implement your suggestion in future with different PR.

vstumpf:
Is there a way to disable all this monitoring? Sometimes you just want to run a ptl test and see if it passes/fails...

birbalsain:
@vstumpf If you use  pbs_benchpress -t testsuite.testcase to run a test then it won't monitor the system and process. If you want to monitor then you have to use  pbs_benchpress -t testsuite.testcase -p procmon=pbs_server:pbs_sched,procmon-freq=120  with default test run format.

visheshh:
Hi Birbal, Code changes look good to me, Will this work only on server machines or can we monitor mom running on a different machine? If Yes, Can you please attach logs for that.

birbalsain:
@visheshh I just added few more parameters in existing checks. currently we are not monitoring the mom/moms. If it is required then we will generate a different ticket.

birbalsain:

test/fw/ptl/utils/pbs_procutils.py
```diff
@@ -269,21 +282,55 @@ def set_frequency(self, value=60):
         self.logger.debug('procmonitor: set frequency to ' + str(value))
         self.frequency = value
 
+    def get_system_stats(self, nw_protocols=['TCP']):
+        """
+        Run system monitoring
+        """
+        cmd = 'sar -rSub -n %s 1 1' % ','.join(nw_protocols)
```
vstumpf: If sar isn't installed, this will fail. Maybe check if it is installed: if not, then don't run.

test/fw/ptl/utils/pbs_procutils.py
```diff
@@ -269,21 +282,55 @@ def set_frequency(self, value=60):
         self.logger.debug('procmonitor: set frequency to ' + str(value))
         self.frequency = value
 
+    def get_system_stats(self, nw_protocols=['TCP']):
```
vstumpf: Using mutable objects as defaults is not the greatest, as if you mutate the object it will not reset.
Use a sentinel value for the default

test/fw/ptl/utils/pbs_procutils.py
```diff
@@ -235,20 +241,26 @@ def __init__(self, name=None, pid=None):
...
         self.pcpu = None
+        self.pmem = None
+        self.size = None
+        self.cputime = None
         self.time = time.time()
         self.command = None
 
     def __str__(self):
-        return "%s pid: %s rss: %s vsz: %s pcpu: %s command: %s" % \
+        return "%s pid: %s rss: %s vsz: %s pcpu: %s pmem: %s \
```
vstumpf: You can replace this with an fstring:
```Python
return f"{self.name} pid: {self.pid} rss: {self.rss} ..."
```
I find it more readable than % formatting, and it's faster. You also don't need to call str() on the values.

test/fw/ptl/utils/pbs_procutils.py
```diff
@@ -235,20 +241,26 @@ def __init__(self, name=None, pid=None):
...
+               size: %s cputime: %s command: %s" % \
```
vstumpf: This will insert all of those spaces into the line. Instead, break the string like so:
```Python
         return (f"really long string {value}"
                 f"really long string {value2}")
```

mike0042:
Please answer the question that @vstumpf posed.

test/fw/ptl/utils/pbs_procutils.py
birbalsain: @vstumpf test skip on sar command is not found.

test/fw/ptl/utils/pbs_procutils.py
birbalsain: @modified

test/fw/ptl/utils/pbs_procutils.py
birbalsain: @vstumpf modified. Is it ok?

test/fw/ptl/utils/pbs_procutils.py
```diff
@@ -282,12 +281,19 @@ def set_frequency(self, value=60):
...
+        if nw_protocols is sentinel:
+            self.skipTest("network protocol values is not set.")
         cmd = 'sar -rSub -n %s 1 1' % ','.join(nw_protocols)
-        rv = self.du.run_cmd(cmd=cmd, as_script=True)
+        try:
+            rv = self.du.run_cmd(cmd=cmd, as_script=True)
+        except BaseException:
+            msg = f"Is sar package installed"
+            self.logger.error(msg)
+            self.skipTest(msg)
```
birbalsain: @vstumpf test skip on exception occurred. Is it ok?

test/fw/ptl/utils/pbs_procutils.py
```diff
@@ -282,12 +281,19 @@ def set_frequency(self, value=60):
         self.logger.debug('procmonitor: set frequency to ' + str(value))
         self.frequency = value
 
-    def get_system_stats(self, nw_protocols=['TCP']):
+    def get_system_stats(self, nw_protocols=sentinel):
         """
         Run system monitoring
         """
+        if nw_protocols is sentinel:
+            self.skipTest("network protocol values is not set.")
```
birbalsain: @vstumpf used the sentinel instead of default.

test/fw/ptl/utils/pbs_procutils.py
```diff
@@ -248,11 +249,9 @@ def __init__(self, name=None, pid=None):
...
 
     def __str__(self):
-        return "%s pid: %s rss: %s vsz: %s pcpu: %s pmem: %s \
-               size: %s cputime: %s command: %s" % \
-               (self.name, str(self.pid), str(self.rss), str(self.vsz),
-                str(self.pcpu), str(self.pmem), str(self.size),
-                str(self.cputime), self.command)
+        return (f"{self.name} pid: {self.pid} rss: {self.rss} vsz: "
+                f"{self.vsz} pcpu: {self.pcpu} pmem: {self.pmem} size: "
+                f"{self.size} cputime: {self.cputime} command: {self.command}")
```
birbalsain: used f" instead of % concept.

test/fw/ptl/utils/pbs_procutils.py
vstumpf: Sorry, i meant use something like
```Python3
def get_stats(self, mw_proto=None):
    # if no protocols set, use default 
    if not nw_proto:
        nw_proto = ['TCP']
```
If you think you should skip the test, that also works.
If you're curious, take a look at [this link](https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html) to see why I made the original comment.

test/fw/ptl/utils/pbs_procutils.py
```diff
@@ -281,24 +281,23 @@ def set_frequency(self, value=60):
...
         cmd = 'sar -rSub -n %s 1 1' % ','.join(nw_protocols)
-        try:
-            rv = self.du.run_cmd(cmd=cmd, as_script=True)
-        except BaseException:
-            msg = f"Is sar package installed"
-            self.logger.error(msg)
-            self.skipTest(msg)
+        rv = self.du.run_cmd(cmd=cmd, as_script=True)
+        if rv['err']:
+            return False
```
birbalsain: @vstumpf self.du.run method itself check command available or not. it through error "command not available"
Now test doesn't check sar command related data.I have atteched the log when command not available.

vstumpf: I think get_system_stats() should return the dictionary already made. That way, you won't have to create self.sysstat AND the _sys_info dictionaries. If the command errors, just return None.
PS: you can also make dictionaries in one statement, like so:
```Python3
_sys_info = {
    'name': "System",
    'time': time.ctime(timenow),
     ...
}
return _sys_info
```

birbalsain: @vstumpf done the changes.

test/fw/ptl/utils/pbs_procutils.py
```diff
@@ -41,7 +41,9 @@
 import threading
 import logging
 import socket
-
+import os
+import json
+from unittest.mock import sentinel
```
vstumpf: You should remove this import

test/fw/ptl/utils/pbs_procutils.py
```diff
@@ -258,6 +269,7 @@ def __init__(self, name=None, regexp=False, frequency=60):
         self._pu = ProcUtils()
         self.stop_thread = threading.Event()
         self.db_proc_info = []
+        self.sysstat = {}
```
vstumpf: You don't need this here, make it a local variable to get_system_stats.

PR #1375, opened 2019-10-25
Fixed test_finished_jobs test when mom is on different node

RKORavi:
#### Describe Bug or Feature
When mom node is different from server node, test_finished_jobs fails as it expects eatcpu.py file on mom.

#### Describe Your Change

1. When job pbs_benchpress is run by providing moms as -p option, then eatcpu script is copied to all test user home directory.
2. Job is submitted as : -S PBS_EXEC/bin/pbs_python shell path.
3. Changed shebang of eatcpu script to "#!/usr/bin/env python3".
#### Link to Design Doc

NA

[ptl_fix_logs.txt](https://github.com/PBSPro/pbspro/files/3771056/ptl_fix_logs.txt)

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -14225,8 +14225,8 @@ class Job(ResourceResv):
         """
         script_dir = os.path.dirname(os.path.dirname(__file__))
         script_path = os.path.join(script_dir, 'utils', 'jobs', 'eatcpu.py')
-        d = os.path.join(os.path.sep, 'home', self.username)
         if mom:
+            d = pwd.getpwnam(self.username).pw_dir
             DshUtils().run_copy(hosts=mom, src=script_path, dest=d)
             script_path = os.path.join(d, "eatcpu.py")
         DshUtils().chmod(path=script_path, mode=0o755)
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -14223,18 +14223,20 @@ class Job(ResourceResv):
         Create a job that eats cpu indefinitely or for the given
         duration of time
         """
+        if self.du is None:
+            self.du = DshUtils()
         script_dir = os.path.dirname(os.path.dirname(__file__))
         script_path = os.path.join(script_dir, 'utils', 'jobs', 'eatcpu.py')
-        if not DshUtils().is_localhost(mom):
-            pbs_conf = DshUtils().parse_pbs_config(mom)
-            shell_path = os.path.join(pbs_conf['PBS_EXEC'],
-                                      'bin', 'pbs_python')
-            a = {ATTR_S: shell_path}
-            self.set_attributes(a)
+        pbs_conf = self.du.parse_pbs_config(mom)
+        shell_path = os.path.join(pbs_conf['PBS_EXEC'],
+                                  'bin', 'pbs_python')
+        a = {ATTR_S: shell_path}
+        self.set_attributes(a)
+        if not self.du.is_localhost(mom):
             d = pwd.getpwnam(self.username).pw_dir
-            DshUtils().run_copy(hosts=mom, src=script_path, dest=d)
+            self.du.run_copy(hosts=mom, src=script_path, dest=d)
             script_path = os.path.join(d, "eatcpu.py")
-        DshUtils().chmod(path=script_path, mode=0o755)
+        self.du.chmod(path=script_path, mode=0o755)
         self.set_execargs(script_path, duration)
 
 
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -14227,15 +14227,18 @@ class Job(ResourceResv):
             self.du = DshUtils()
         script_dir = os.path.dirname(os.path.dirname(__file__))
         script_path = os.path.join(script_dir, 'utils', 'jobs', 'eatcpu.py')
+        if not self.du.is_localhost(mom):
+            d = pwd.getpwnam(self.username).pw_dir
+            ret = self.du.run_copy(hosts=mom, src=script_path, dest=d)
+            if ret is None or ret['rc'] != 0:
+                raise AssertionError("Failed to copy file %s"
+                                     % (script_path))
+            script_path = os.path.join(d, "eatcpu.py")
         pbs_conf = self.du.parse_pbs_config(mom)
         shell_path = os.path.join(pbs_conf['PBS_EXEC'],
                                   'bin', 'pbs_python')
         a = {ATTR_S: shell_path}
         self.set_attributes(a)
-        if not self.du.is_localhost(mom):
-            d = pwd.getpwnam(self.username).pw_dir
-            self.du.run_copy(hosts=mom, src=script_path, dest=d)
-            script_path = os.path.join(d, "eatcpu.py")
         self.du.chmod(path=script_path, mode=0o755)
         self.set_execargs(script_path, duration)
 
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -14231,8 +14231,8 @@ class Job(ResourceResv):
             d = pwd.getpwnam(self.username).pw_dir
             ret = self.du.run_copy(hosts=mom, src=script_path, dest=d)
             if ret is None or ret['rc'] != 0:
-                raise AssertionError("Failed to copy file %s"
-                                     % (script_path))
+                raise AssertionError("Failed to copy file %s to %s"
+                                     % (script_path, mom))
             script_path = os.path.join(d, "eatcpu.py")
         pbs_conf = self.du.parse_pbs_config(mom)
         shell_path = os.path.join(pbs_conf['PBS_EXEC'],
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -14218,13 +14218,17 @@ class Job(ResourceResv):
         return job_array_id[:idx + 1] + str(subjob_index) + \
             job_array_id[idx + 1:]
 
-    def create_eatcpu_job(self, duration=None):
+    def create_eatcpu_job(self, duration=None, mom=None):
         """
         Create a job that eats cpu indefinitely or for the given
         duration of time
         """
         script_dir = os.path.dirname(os.path.dirname(__file__))
         script_path = os.path.join(script_dir, 'utils', 'jobs', 'eatcpu.py')
+        d = os.path.join(os.path.sep, 'home', self.username)
+        if mom:
+            DshUtils().run_copy(hosts=mom, src=script_path, dest=d)
+            script_path = os.path.join(d, "eatcpu.py")
         DshUtils().chmod(path=script_path, mode=0o755)
         self.set_execargs(script_path, duration)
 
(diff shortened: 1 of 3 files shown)


test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -14218,13 +14218,17 @@ def create_subjob_id(self, job_array_id, subjob_index):
...
 
-    def create_eatcpu_job(self, duration=None):
+    def create_eatcpu_job(self, duration=None, mom=None):
         """
         Create a job that eats cpu indefinitely or for the given
         duration of time
         """
         script_dir = os.path.dirname(os.path.dirname(__file__))
         script_path = os.path.join(script_dir, 'utils', 'jobs', 'eatcpu.py')
+        d = os.path.join(os.path.sep, 'home', self.username)
```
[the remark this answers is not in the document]
agrawalravi90: Also, use 'pwd.getpwnam(user).pw_dir' to get the home directory of the user

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -14218,13 +14218,17 @@ def create_subjob_id(self, job_array_id, subjob_index):
...
+        if mom:
+            DshUtils().run_copy(hosts=mom, src=script_path, dest=d)
```
agrawalravi90: you should probably check here if mom is localhost or not, if it's localhost then you don't need to do the remote copy.

test/tests/pbs_smoketest.py
```diff
@@ -335,10 +335,16 @@ def test_finished_jobs(self):
...
-        a = {'Resource_List.ncpus': 2}
+        shell_path = os.path.join(self.server.pbs_conf['PBS_EXEC'],
+                                  "bin", "pbs_python")
+        a = {'Resource_List.ncpus': 2, ATTR_S: shell_path}
         j = Job(TEST_USER, a)
         j.set_sleep_time(15)
-        j.create_eatcpu_job(15)
+        mom = self.moms.values()[0].shortname
+        server = self.server.shortname
+        if server == mom:
```
agrawalravi90: just move the localhost check inside create_eatcpu_job() itself, that way PTL can take care of this for the test writer

test/tests/pbs_smoketest.py
```diff
@@ -335,10 +335,16 @@ def test_finished_jobs(self):
         self.server.manager(MGR_CMD_SET, NODE, a, self.mom.shortname)
         a = {'job_history_enable': 'True'}
         self.server.manager(MGR_CMD_SET, SERVER, a)
-        a = {'Resource_List.ncpus': 2}
+        shell_path = os.path.join(self.server.pbs_conf['PBS_EXEC'],
+                                  "bin", "pbs_python")
+        a = {'Resource_List.ncpus': 2, ATTR_S: shell_path}
```
agrawalravi90: If Shell_Path_List is needed for running the eat cpu script, then it might be better to set this attribute inside create_eatcpu_job(). What do you think?

test/fw/ptl/utils/jobs/eatcpu.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
```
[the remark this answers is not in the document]
RKORavi: If we use this shebang: #!/usr/bin/python3, we have to create a softlink for this. It might break something on system as some applications will be dependent on python2. 
While #!/usr/bin/env python will run whichever python is installed on the system.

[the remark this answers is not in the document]
agrawalravi90: Well, this script is supposed to be used for testing purposes, at least chiefly. We use "#!/usr/bin/env python3" in all of our scripts now. So, I think we should be consistent unless there's an issue with using "#!/usr/bin/env python3"

test/fw/ptl/lib/pbs_testlib.py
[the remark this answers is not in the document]
RKORavi: I tried self.du but i was getting AttributeError as:
self.du.run_copy(hosts=mom, src=script_path, dest=d)
AttributeError: 'NoneType' object has no attribute 'run_copy'

test/tests/pbs_smoketest.py
RKORavi: I too agree that these checks can be moved to create_eatcpu_job() but I was unable to access server or mom attributes from Job class, that is the reason i have added the code here

test/tests/pbs_smoketest.py
RKORavi: If i move this code to create_eatcpu_job() then i need to pass shell path to the method as well. 
Since i did not find a way to access PBS_CONF from Job class.

test/fw/ptl/lib/pbs_testlib.py
RKORavi: In some of the text in ptl i see this: pbs_benchpress -p server=remote@/etc/pbs.conf, in this case checking for localhost will not help. I feel checking if server name and mom name are equal is better option here. What do you think?

test/fw/ptl/utils/jobs/eatcpu.py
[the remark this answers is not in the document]
RKORavi: changed the shebang to #!/usr/bin/env python3 as with earlier shebang "/usr/bin/env python" it was failing in debian and ubuntu18 (Travis)

test/fw/ptl/lib/pbs_testlib.py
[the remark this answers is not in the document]
agrawalravi90: Oh, I didn't realize that this function was part of the Job class. I don't think that's the right place for it, this function should create a Job object itself, so I think it should be in the Server class. Anyways, now that it's in the Job class, most of my comments are not valid.

test/tests/pbs_smoketest.py
hirenvadalia: You can use DshUtils.parse_pbs_conf in Job class to get pbs.conf

test/tests/pbs_smoketest.py
hirenvadalia: you can pass self.server as a argument like you are passing self.mom

test/fw/ptl/lib/pbs_testlib.py
hirenvadalia: I agree with @agrawalravi90, make use DshUtils.is_localhost on mom and do run_copy only if it is a remote mom.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -14225,7 +14225,12 @@ def create_eatcpu_job(self, duration=None, mom=None):
         """
         script_dir = os.path.dirname(os.path.dirname(__file__))
         script_path = os.path.join(script_dir, 'utils', 'jobs', 'eatcpu.py')
-        if mom:
+        if not DshUtils().is_localhost(mom):
+            pbs_conf = DshUtils().parse_pbs_config(mom)
+            shell_path = os.path.join(pbs_conf['PBS_EXEC'],
+                                   'bin', 'pbs_python')
+            a = {ATTR_S: shell_path}
+            self.set_attributes(a)
```
hirenvadalia: this should be outside of `if not localhost` condition right? because we want to use that shell as python whether mom is remote or not, right?

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -14225,7 +14225,12 @@ def create_eatcpu_job(self, duration=None, mom=None):
         """
         script_dir = os.path.dirname(os.path.dirname(__file__))
         script_path = os.path.join(script_dir, 'utils', 'jobs', 'eatcpu.py')
-        if mom:
+        if not DshUtils().is_localhost(mom):
```
hirenvadalia: create `self.du = DshUtils()` in Job class and use it everywhere

hirenvadalia:
@RKORavi Please don't do force push once the review is started.

test/fw/ptl/lib/pbs_testlib.py
RKORavi: yes, thanks for catching that i will make the change

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -14218,14 +14218,25 @@ def create_subjob_id(self, job_array_id, subjob_index):
         return job_array_id[:idx + 1] + str(subjob_index) + \
             job_array_id[idx + 1:]
 
-    def create_eatcpu_job(self, duration=None):
+    def create_eatcpu_job(self, duration=None, mom=None):
         """
         Create a job that eats cpu indefinitely or for the given
         duration of time
         """
+        if self.du is None:
```
hirenvadalia: you don't need these 2 lines as I just checked Job class has `self.du` already defined as a class variable and it will never be None

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -14218,14 +14218,25 @@ def create_subjob_id(self, job_array_id, subjob_index):
...
         script_path = os.path.join(script_dir, 'utils', 'jobs', 'eatcpu.py')
-        DshUtils().chmod(path=script_path, mode=0o755)
+        pbs_conf = self.du.parse_pbs_config(mom)
+        shell_path = os.path.join(pbs_conf['PBS_EXEC'],
+                                  'bin', 'pbs_python')
+        a = {ATTR_S: shell_path}
+        self.set_attributes(a)
+        if not self.du.is_localhost(mom):
+            d = pwd.getpwnam(self.username).pw_dir
+            self.du.run_copy(hosts=mom, src=script_path, dest=d)
```
hirenvadalia: what if this run_copy() fails?

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -14218,14 +14218,25 @@ def create_subjob_id(self, job_array_id, subjob_index):
...
+        if self.du is None:
+            self.du = DshUtils()
         script_dir = os.path.dirname(os.path.dirname(__file__))
         script_path = os.path.join(script_dir, 'utils', 'jobs', 'eatcpu.py')
-        DshUtils().chmod(path=script_path, mode=0o755)
+        pbs_conf = self.du.parse_pbs_config(mom)
+        shell_path = os.path.join(pbs_conf['PBS_EXEC'],
+                                  'bin', 'pbs_python')
+        a = {ATTR_S: shell_path}
+        self.set_attributes(a)
```
hirenvadalia: keeping in mind that run_copy() can fail, I suggest moving this part after if condition

test/fw/ptl/lib/pbs_testlib.py
hirenvadalia: sounds good to me for raising AssertionError.

test/fw/ptl/lib/pbs_testlib.py
RKORavi: Hi Hiren, 
At first I tried self.du but i was getting AttributeError as:
self.du.run_copy(hosts=mom, src=script_path, dest=d)
AttributeError: 'NoneType' object has no attribute 'run_copy'
That is the reason I had to add these two lines here

RKORavi:
[failed_copy_logs.txt](https://github.com/PBSPro/pbspro/files/3812566/failed_copy_logs.txt)

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -14227,15 +14227,18 @@ def create_eatcpu_job(self, duration=None, mom=None):
             self.du = DshUtils()
         script_dir = os.path.dirname(os.path.dirname(__file__))
         script_path = os.path.join(script_dir, 'utils', 'jobs', 'eatcpu.py')
+        if not self.du.is_localhost(mom):
+            d = pwd.getpwnam(self.username).pw_dir
+            ret = self.du.run_copy(hosts=mom, src=script_path, dest=d)
+            if ret is None or ret['rc'] != 0:
+                raise AssertionError("Failed to copy file %s"
```
hirenvadalia: Please include mom name in error message like `"Failed to copy file %s to %s" % (script_path, mom)`

PR #1377, opened 2019-10-30
Fix for sched crash with entity limits

arungrover:
#### Describe Bug or Feature
Scheduler crashes when it is trying to check limits for an entity that has no other registered running job requesting the limit resource.

#### Describe Your Change
Change is simple, an 'if' condition was wrongly scoped. It ended up dereferencing uninitialized stack variables.

#### Link to Design Doc
NA

I ended up extending an existing limits smoketest.
[test_before.txt](https://github.com/PBSPro/pbspro/files/3790263/test_before.txt)
[test_after.txt](https://github.com/PBSPro/pbspro/files/3790264/test_after.txt)

### DIFF
--- a/src/scheduler/check.c
+++ b/src/scheduler/check.c
@@ -1341,10 +1341,9 @@ find_counts_elm(counts *cts_list, char *name, resdef *rdef, counts **cnt, resour
 			*cnt = cts;
 		if (rdef == NULL)
 			return cts->running;
-		else {
-			if ((res_lim = find_resource_count(cts->rescts, rdef)) != NULL)
-				if (rcount != NULL)
-					*rcount = res_lim;
+		else if ((res_lim = find_resource_count(cts->rescts, rdef)) != NULL) {
+			if (rcount != NULL)
+				*rcount = res_lim;
 			return res_lim->amount;
 		}
 	}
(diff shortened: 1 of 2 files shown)


src/scheduler/check.c
```diff
@@ -1341,10 +1341,9 @@ find_counts_elm(counts *cts_list, char *name, resdef *rdef, counts **cnt, resour
...
 		if (rdef == NULL)
 			return cts->running;
-		else {
-			if ((res_lim = find_resource_count(cts->rescts, rdef)) != NULL)
-				if (rcount != NULL)
-					*rcount = res_lim;
+		else if ((res_lim = find_resource_count(cts->rescts, rdef)) != NULL) {
+			if (rcount != NULL)
+				*rcount = res_lim;
 			return res_lim->amount;
```
agrawalravi90: So, I think this might be changing existing logic. Previously, the code would return res_lim->amount regardless of the if ((res_lim = find_resource_count(cts->rescts, rdef)) condition, now it will return it only if that condition is true, is this change intentional?

arungrover: Right that's the bug, because, find_resource_count returns "res_lim". so unless it is not-null, we end up dereferencing an uninitialized pointer.

PR #1378, opened 2019-10-31
Fixed server crash with alter record

bhroam:
#### Describe Bug or Feature
A recent PR introduced a new accounting log record ('a').  A structure which is used to encode the attribute to print was not cleared between uses.  This would cause a double free.

#### Describe Your Change
I moved the variable to the block it was being used, so it leaves scope and is reintroduced between uses in the loop.

[alter.log](https://github.com/PBSPro/pbspro/files/3795997/alter.log)
[valgrind.log](https://github.com/PBSPro/pbspro/files/3796004/valgrind.log)

### DIFF
--- a/src/server/accounting.c
+++ b/src/server/accounting.c
@@ -2236,7 +2236,6 @@ writeit:
  * @returns void 
  */
 void log_alter_records_for_attrs(job *pjob, svrattrl *plist) {
-	svrattrl *svrattrl_list;
 	svrattrl *cur_svr;
 	pbs_list_head phead;
 	svrattrl *cur_plist;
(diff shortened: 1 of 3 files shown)


src/server/accounting.c
```diff
@@ -2257,6 +2256,7 @@ void log_alter_records_for_attrs(job *pjob, svrattrl *plist) {
 	CLEAR_HEAD(phead);
 	for (i = 0; i < JOB_ATR_LAST; i++) {
 		if (pjob->ji_wattr[i].at_flags & ATR_VFLAG_MODIFY) {
+			svrattrl *svrattrl_list = {0};
```
agrawalravi90: Can't you use NULL instead?

PR #1380, opened 2019-11-04
Do not emit 'a' record for estimated resources

bhroam:
#### Describe Bug or Feature
The estimated resources provide little use for accounting purposes.  They are there for the user to know when their job will run and where it will run.  Do not emit an 'a' record for these resources.

#### Describe Your Change
If the log_alter_records_for_attrs() function sees an estimated resource, don't emit an 'a' record for it.

The only way to test this is through the API.  You can't alter estimated attributes through qalter.  Here's some manual testing through an internal tool
[estimated_test.log](https://github.com/PBSPro/pbspro/files/3806146/estimated_test.log)

### DIFF
--- a/src/server/accounting.c
+++ b/src/server/accounting.c
@@ -79,20 +79,21 @@
 
 /* Local Data */
 
-static FILE	    *acctfile;		/* open stream for log file */
-static volatile int  acct_opened = 0;
-static int	     acct_opened_day;
-static int	     acct_auto_switch = 0;
-static char	    *acct_buf = 0;
-static int	     acct_bufsize = PBS_ACCT_MAX_RCD;
+static FILE *acctfile;		/* open stream for log file */
+static volatile int acct_opened = 0;
+static int acct_opened_day;
+static int acct_auto_switch = 0;
+static char *acct_buf = 0;
+static int acct_bufsize = PBS_ACCT_MAX_RCD;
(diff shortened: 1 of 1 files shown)


src/server/accounting.c
```diff
@@ -2259,7 +2260,14 @@ void log_alter_records_for_attrs(job *pjob, svrattrl *plist) {
 			svrattrl *svrattrl_list = NULL;
 			job_attr_def[i].at_encode(&pjob->ji_wattr[i], &phead, job_attr_def[i].at_name, NULL, ATR_ENCODE_CLIENT, &svrattrl_list);
 			for (cur_plist = plist; cur_plist != NULL; cur_plist = (svrattrl *)GET_NEXT(cur_plist->al_link)) {
-				if (strcmp(cur_plist->al_name, job_attr_def[i].at_name) != 0)
+				int j;
+				int found = 0;
```
arungrover: This is really a nit pick- is it possible to change the variable name to "ignore" instead of found?

src/server/accounting.c
```diff
@@ -2286,7 +2294,7 @@ void log_alter_records_for_attrs(job *pjob, svrattrl *plist) {
 							break;
 						}
 					}
-				}
+						}
```
arungrover: This tab makes it look like it is wrongly indented

PR #1381, opened 2019-11-04
qsub fails with long select PBS directive

arungrover:
#### Describe Bug or Feature
qsub command fails when select specification given in a job script file exceeds 4096 characters in length.

#### Describe Your Change
We were expecting all directives to be no more than 4K characters long which is why qsub was failing.
I've changed the code to make it dynamic.

#### Link to Design Doc
NA

[test_before.txt](https://github.com/PBSPro/pbspro/files/3806289/test_before.txt)
[test_after.txt](https://github.com/PBSPro/pbspro/files/3806290/test_after.txt)

### DIFF
--- a/src/cmds/qsub.c
+++ b/src/cmds/qsub.c
@@ -3160,7 +3160,7 @@ get_script(FILE *file, char *script, char *prefix)
 	char tmp_name[MAXPATHLEN + 1];
 	FILE *TMP_FILE;
 	char *in;
-	char *s_in;
+	char *s_in = NULL;
 	int s_len = 0;
 #ifndef WIN32
 	int fds;
--- a/src/cmds/qsub.c
+++ b/src/cmds/qsub.c
@@ -2979,7 +2979,6 @@ make_argv(int *argc, char *argv[], char *line)
 	char *l, *b, *c;
 	char static_buffer[MAX_LINE_LEN + 1];
 	char *buffer;
-	int buf_allocated = 0;
 	int line_len = 0;
 	int len;
 	char quote;
@@ -2995,7 +2994,6 @@ make_argv(int *argc, char *argv[], char *line)
 			fprintf(stderr, "qsub: out of memory\n");
 			exit_qsub(2);
 		}
-		buf_allocated = 1;
 	}
 	else
 		buffer = static_buffer;
@@ -3052,7 +3050,7 @@ make_argv(int *argc, char *argv[], char *line)
 		free(argv[i]);
 		argv[i++] = NULL;
 	}
-	if (buf_allocated)
+	if (buffer != static_buffer)
 		free(buffer);
 }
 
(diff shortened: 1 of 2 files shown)


arungrover:
Thanks @bhroam for your review comments, I've addressed them in the latest commit.

test/tests/functional/pbs_job_script.py
```diff
@@ -63,4 +60,30 @@ def test_long_select_spec(self):
         j = Job()
         j.create_script(scr)
         jid = self.server.submit(j)
-        self.server.expect(JOB, {'job_state': 'R'}, id=jid)
+        return jid
+
+    def test_long_select_spec_without_newline(self):
```
bhroam: How about we call this just 'test_long_select_spec'

test/tests/functional/pbs_job_script.py
```diff
@@ -63,4 +60,30 @@ def test_long_select_spec(self):
...
+        """
+        jid = self.submit_job()
+        execvnode = ""
+        for node in range(500):
+            execvnode += "(Verylongvnodename[" + str(node) + "]:ncpus=1)+"
+        execvnode = execvnode[0:-1]
+        self.server.expect(JOB, {'job_state': 'R', 'exec_vnode': execvnode},
+                           id=jid)
+
+    def test_long_select_spec_with_newline(self):
```
bhroam: And this one could be test_long_select_spec_extend

arungrover:
@bhroam I have changed the name of the test. Please have a look.
[test_after.txt](https://github.com/PBSPro/pbspro/files/3816102/test_after.txt)

src/cmds/qsub.c
```diff
@@ -3038,6 +3052,8 @@ make_argv(int *argc, char *argv[], char *line)
 		free(argv[i]);
 		argv[i++] = NULL;
 	}
+	if (buf_allocated)
```
mike0042: You don't need buf_allocated if you change the conditional to:
```suggestion
	if (buffer != static_buffer)
```

arungrover: Thanks @mike0042! I've addressed your comments. Please have a look at the recent commit.

src/cmds/qsub.c
```diff
@@ -3140,14 +3154,14 @@ set_opt_defaults(void)
...
 {
-	char s[MAX_LINE_LEN + 1];
 	char *sopt;
 	int err = 0;
 	int exec = FALSE;
-	char *cont;
 	char tmp_name[MAXPATHLEN + 1];
 	FILE *TMP_FILE;
 	char *in;
+	char *s_in;
```
agrawalravi90: To be on he safer side, I think you should set s_in to NULL here, so that just in case pbs_fgets_extend fails, you don't end up freeing garbage memory with free(s_in) later.

arungrover: This is a remote possibility, but it is a possibility nonetheless. I have addressed your comment Ravi, please have a look.

PR #1382, opened 2019-11-05
[WIP] - Update valgrind.supp to suppress memory leak caused by Python 3 source

npadole20:
#### Describe Bug or Feature

We use Py_InitializeEx() in pbs_python_ext_quick_start_interpreter() and pbs_python_ext_start_interpreter(). Py_InitializeEx() has leaks which show up in PBS when run under valgrind

#### Describe Your Change

Added those leaks to the suppression file as they originate from the Python 3 source code

#### Link to Design Doc

NA

[generated-suppfile-3.5 .txt](https://github.com/PBSPro/pbspro/files/3808014/generated-suppfile-3.5.txt)
[sample-program.txt](https://github.com/PBSPro/pbspro/files/3808023/sample-program.txt)

### DIFF
--- a/valgrind.supp
+++ b/valgrind.supp
@@ -427,6 +427,7 @@
    fun:_PyUnicode_Init
    fun:_Py_InitializeEx_Private
    fun:Py_InitializeEx
+   ...
    fun:main
 }
 {
@@ -3177,6 +3178,7 @@
    fun:_PyLong_Init
    fun:_Py_InitializeEx_Private
    fun:Py_InitializeEx
+   ...
    fun:main
 }
 {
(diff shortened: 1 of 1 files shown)


subhasisb:
Hi @npadole20 - two thoughts:
1. We should remove the python 2.7 specific leak suppressions from the file
2. We should only add suppressions published by the python community; we cannot add suppressions that just appear in our tests without having confirmation from python dev as to whether they should be suppressed. Since it is not our code, we cannot know whether they are actual leaks or not.

agrawalravi90:
Closing because of inactivity, please re-open/create new when you get time to work on this.

PR #1386, opened 2019-11-05
Fix log_match dst failure

vstumpf:
#### Describe Bug or Feature
On some machines, all log_matches are failing, even though the message appears.

#### Describe Your Change
For some reason, `time.daylight` is returning 1, even though daylight savings time has ended.
Instead of doing all the math the code does now, we can use things introduced in python3.
Datetime objects now have a `.timestamp()` method that returns the epoch timestamp, while assuming the datetime is in the local timezone. It's exactly what we need!

[afterfix.txt](https://github.com/PBSPro/pbspro/files/3811006/afterfix.txt)
[beforefix.txt](https://github.com/PBSPro/pbspro/files/3811007/beforefix.txt)

### DIFF
--- a/test/fw/ptl/utils/pbs_logutils.py
+++ b/test/fw/ptl/utils/pbs_logutils.py
@@ -210,9 +210,6 @@ PARSER_OK_STOP = 1
 PARSER_ERROR_CONTINUE = 2
 PARSER_ERROR_STOP = 3
 
-epoch_datetime = datetime.fromtimestamp(0)
-
-
 class PBSLogUtils(object):
 
     """
@@ -240,12 +237,6 @@ class PBSLogUtils(object):
         if dt is None:
             return None
 
-        stdoffset = timedelta(seconds=-time.timezone)
-        if time.daylight:
-            dstoffset = timedelta(seconds=-time.altzone)
-        else:
-            dstoffset = stdoffset
-        offsetdiff = dstoffset - stdoffset
         micro = False
         if fmt is None:
             if '.' in dt:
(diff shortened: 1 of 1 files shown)


test/fw/ptl/utils/pbs_logutils.py
```diff
@@ -369,6 +359,8 @@ def match_msg(self, lines, msg, allmatch=False, regexp=False,
                 dt_str = l.split(';', 1)[0]
                 if starttime is not None:
                     tm = self.convert_date_time(dt_str)
+                    self.logger.info(f'tm is {tm} | starttime is {starttime}'
```
hirenvadalia: is this debugging message? if yes then please remove. If not then I suggest logging at debug level else this message will appear for each and every line passed to this method (i.e. `lines`) which will be too much info.

shilpakodli:
@vstumpf Could you please test with PDT set manually in your local machine and verify the same.

test/fw/ptl/utils/pbs_logutils.py
vstumpf: Sorry, it failed in jenkins and I was trying to debug it here, will WIP PR

vstumpf:
[afterfix-dst.txt](https://github.com/PBSPro/pbspro/files/3817343/afterfix-dst.txt)
I've attached the logs when dst is enabled
Please continue with this PR

test/fw/ptl/utils/pbs_logutils.py
vstumpf: It now passes in jenkins

bhroam:
I'm going to merge this since it's blocking the build.  @hirenvadalia asked for a debug statement to be removed and @vstumpf removed it.  If @hirenvadalia has more comments, we can make a new PR.

hirenvadalia:
No, I don't have any more comments on this change.

PR #1388, opened 2019-11-06
Skipping few test cases for a specific platform

sujatapatnaik52:
#### Describe Bug or Feature
Few of the test cases from pbs_test_entity_limits does not work when linux users or groups are not present at the server.

#### Describe Your Change
Skipping them for a specific platform where the linux users or groups will not be present at the server. 

#### Attach Test Logs/Output
[skipped_test_logs.txt](https://github.com/PBSPro/pbspro/files/3812995/skipped_test_logs.txt)

### DIFF
--- a/test/tests/functional/pbs_test_entity_limits.py
+++ b/test/tests/functional/pbs_test_entity_limits.py
@@ -134,7 +134,6 @@ class TestEntityLimits(TestFunctional):
             self.assertFalse(True, "Job violating limits got submitted after "
                              "server restart.")
 
-    @skipOnShasta
     def test_server_generic_user_limits_queued(self):
         """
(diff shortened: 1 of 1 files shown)


vstumpf:
User tests work even if the user doesn't exist on the machine.
It's only the group limits tests, since the pbs-server has to look up the gid of the user, which doesn't exist.
For user limits, the pbs-server just takes the username from the qsub request.
Please unskip the user tests.

sujatapatnaik52:
[skipped_test_logs.txt](https://github.com/PBSPro/pbspro/files/3818585/skipped_test_logs.txt)

sujatapatnaik52:
@vstumpf Please review the latest commit.

Issue #1383, opened 2019-11-05
Server failing to recover queue on restart and having panic shutdown


saritakh:
Server is unable to recover a queue having preempt_targets set as a queue and has a panic shutdown when we try to create this un-recovered queue.

Steps to reproduce:

1. Create workq2 and workq3
2. On workq2 set preempt_targets as a queu
set queue workq2 preempt_targets=queue=workq3		
3. Restart PBS
/etc/init.d/pbs restart
PBS fails to recover workq2!
4. Try to create workq2
create queue workq2 queue_type=e,enabled=t,started=t
The server has a panic shutdown!

vstumpf:
Fixed in 36458ef

PR #1308, opened 2019-09-20
big array server restart crash and qalter usage correction

Shrini-h:
#### Describe Bug or Feature

1. Server crashes during restart, while recovering a job array whose size is bigger than the currently set `max_array_size` server attribute.

2. `qalter` usage string wrongly mentions Job array option `[-J X-Y[:Z]]`, which is infact not supported by it.
```
$ qalter -J 1-20 7[].pbsserver
qalter: invalid option -- 'J'
usage: qalter [-a date_time] [-A account_string] [-c interval] [-e path]
        [-h hold_list] [-j y|n] [-k keep] [-J X-Y[:Z]] [-l resource_list]
        [-m mail_options] [-M user_list] [-N jobname] [-o path] [-p priority]
        [-R o|e|oe] [-r y|n] [-S path] [-u user_list] [-W dependency_list]
        [-P project_name] job_identifier...
       qalter --version
```

3. Invalid read of 4 bytes with below stack while executing TestJobArray PTL testsuite
```
==64194== Invalid read of size 4
==64194==    at 0x47B31F: req_deletejob (req_delete.c:445)
==64194==    by 0x4779CA: dispatch_request (process_request.c:683)
==64194==    by 0x477666: process_request (process_request.c:488)
==64194==    by 0x4F117D: process_socket (net_server.c:504)
==64194==    by 0x4F148C: wait_request (net_server.c:620)
==64194==    by 0x475BC3: main (pbsd_main.c:2141)
```

4. Dead code found functions: `node_recov_db()` and `db_to_svr_node()` in `node_recov_db.c`

#### Describe Your Change

1. made `decode_attr_db()` to respond to the return code of `at_action()` action functions of attributes
2. made `mk_subjob_index_tbl()` to compare Job array size with `max_array_size` server attribute only for `ATR_ACTION_NEW` and `ATR_ACTION_ALTER`
3. added validation of job array flags at the beginning of action function of `array_indices_remaining`, so that we can avoid server crash
4. corrected `qalter`'s usage by removing Job array option  `[-J X-Y[:Z]]`
5. Fixed invalid read by moving the offending lines to appropriate locations
6. purged the dead code functions: `node_recov_db()` and `db_to_svr_node()` in `node_recov_db.c`

#### Attach Test and Valgrind Logs/Output

[before-fail.zip](https://github.com/PBSPro/pbspro/files/3637418/before-fail.zip)
[after-pass.zip](https://github.com/PBSPro/pbspro/files/3637420/after-pass.zip)
[...]

PR #1394, opened 2019-11-08
Server failing to recover queue on restart and having panic shutdown

Shrini-h:
#### Describe Bug or Feature

1. #1383 
2. two memory leaks found after #1308 when server encounters db recovery failures

1:
Server is unable to recover a queue having preempt_targets set as a queue and has a panic shutdown when we try to create this un-recovered queue.

Steps to reproduce:

Create workq2 and workq3
On workq2 set preempt_targets as a queu
set queue workq2 resources_default.preempt_targets=queue=workq3
Restart PBS
/etc/init.d/pbs restart
PBS fails to recover workq2!
Try to create workq2
create queue workq2 queue_type=e,enabled=t,started=t
The server has a panic shutdown!

2:
The two memory leak signatures when a queue db recovery fails during server init
363 bytes in 3 blocks are definitely lost in loss record 848 of 1,130
   by 0x460CC7: make_attr (attr_recov_db.c:109)
   by 0x460FAB: decode_attr_db (attr_recov_db.c:300)
   by 0x4943EA: db_to_svr_que (queue_recov_db.c:141)
   by 0x4943EA: que_recov_db (queue_recov_db.c:234)
   by 0x48C96F: pbsd_init (pbsd_init.c:889)
   by 0x490B6D: main (pbsd_main.c:1691)

&

1,169 (1,064 direct, 105 indirect) bytes in 1 blocks are definitely lost in loss record 964 of 1,130
   by 0x53EC56: convert_array_to_db_attr_list (db_postgres_attr.c:115)
   by 0x53D71F: pg_db_load_que (db_postgres_que.c:237)
   by 0x494390: que_recov_db (queue_recov_db.c:231)
   by 0x48C96F: pbsd_init (pbsd_init.c:889)
   by 0x490B6D: main (pbsd_main.c:1691)

#### Describe Your Change

1. Made the action function `preempt_targets_action()`  return `PBSE_NONE` when the `(actmode == ATR_ACTION_RECOV)`
2. Freed few malloc's when there is db recovery failure

[Before_test_preempt_queue_restart.zip](https://github.com/PBSPro/pbspro/files/3825237/Before_test_preempt_queue_restart.zip)
[After_test_preempt_queue_restart.zip](https://github.com/PBSPro/pbspro/files/3825241/After_test_preempt_queue_restart.zip)
[SmokeTest.zip](https://github.com/PBSPro/pbspro/files/3825238/SmokeTest.zip)
[valgrind_before.log](https://github.com/PBSPro/pbspro/files/3825240/valgrind_before.log)
[valgrind_after.log](https://github.com/PBSPro/pbspro/files/3825239/valgrind_after.log)

### DIFF
--- a/src/server/job_recov_db.c
+++ b/src/server/job_recov_db.c
@@ -545,14 +545,18 @@ job_recov_db(char *jid)
 
 	pj = job_recov_db_spl(&dbjob);
 	if (!pj)
-		goto db_err;
+		goto db_err_reset;
 
 	if (pbs_db_end_trx(conn, PBS_DB_COMMIT) != 0)
-		goto db_err;
+		goto db_err_reset;
 
 	pbs_db_reset_obj(&obj);
 
 	return (pj);
+
+db_err_reset:
+	pbs_db_reset_obj(&obj);
+
 db_err:
 	if (pj)
 		job_free(pj);
@@ -690,7 +694,7 @@ resv_recov_db(char *resvid)
 		goto db_err;
 
 	if (db_to_svr_resv(presv, &dbresv) != 0)
-		goto db_err;
+		goto db_err_reset;
 
 	pbs_db_reset_obj(&obj);
 
@@ -698,6 +702,9 @@ resv_recov_db(char *resvid)
 		goto db_err;
 
 	return (presv);
+
+db_err_reset:
+	pbs_db_reset_obj(&obj);
 db_err:
 	if (presv)
 		resv_free(presv);
--- a/src/server/attr_recov_db.c
+++ b/src/server/attr_recov_db.c
@@ -423,6 +423,15 @@ decode_attr_db(
 							snprintf(log_buffer,LOG_BUF_SIZE, "Action function failed for %s attr, errn %d",
 									(padef+index)->at_name, act_rc);
 							log_err(act_rc, __func__, log_buffer);
+							for ( index++; index <= limit; index++) {
+								while (pal) {
+									tmp_pal = pal->al_sister;
+									free(pal);
+									pal = tmp_pal;
+								}
+								if (index < limit)
+									pal = palarray[index];
+							}
 							free(palarray);
 							/* bailing out from this function */
 							/* any previously allocated attrs will be */
--- a/src/server/queue_recov_db.c
+++ b/src/server/queue_recov_db.c
@@ -224,20 +224,23 @@ que_recov_db(char *qname)
 		return NULL;
 	}
 
-	/* load server_qs */
+	/* load queue name */
 	snprintf(dbque.qu_name, sizeof(dbque.qu_name), "%s", qname);
 
-	/* read in job fixed sub-structure */
+	/* read in queue fixed sub-structure */
 	if (pbs_db_load_obj(conn, &obj) != 0)
 		goto db_err;
 
 	if (db_to_svr_que(pq, &dbque) != 0)
-		goto db_err;
+		goto db_err_reset;
 
 	pbs_db_reset_obj(&obj);
 
 	/* all done recovering the queue */
 	return (pq);
+
+db_err_reset:
+	pbs_db_reset_obj(&obj);
 
 db_err:
 	snprintf(log_buffer, LOG_BUF_SIZE, "Failed to recover queue %s", qname);
--- a/src/server/resc_attr.c
+++ b/src/server/resc_attr.c
@@ -414,7 +414,7 @@ preempt_targets_action(resource *presc, attribute *pattr, void *pobject, int typ
 	resource_def *resdef = NULL;
 	char ch;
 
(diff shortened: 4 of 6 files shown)


src/server/job_recov_db.c
```diff
@@ -545,14 +545,18 @@ job_recov_db(char *jid)
...
 
 	if (pbs_db_end_trx(conn, PBS_DB_COMMIT) != 0)
-		goto db_err;
+		goto db_err_reset;
 
 	pbs_db_reset_obj(&obj);
 
 	return (pj);
+
+db_err_reset:
```
subhasisb: Far i recall, we can call pbs_db_reset_obj() directly - the function checks whether to free or not. So perhaps we can it anyways db_err label itself, instead of making another label db_err_reset. Please check that.

src/server/job_recov_db.c
```diff
@@ -690,14 +694,17 @@ resv_recov_db(char *resvid)
...
+		goto db_err_reset;
 
 	pbs_db_reset_obj(&obj);
 
 	if (pbs_db_end_trx(conn, PBS_DB_COMMIT) != 0)
 		goto db_err;
 
 	return (presv);
+
+db_err_reset:
```
subhasisb: Same comment as above

src/server/job_recov_db.c
Shrini-h: Hey @subhasisb I had the same plan initially but after analyzing the db interface code I saw conditional jump on uninitialized stack memory in `free_db_attr_list()` if the failure happens before a call to `convert_array_to_db_attr_list()`, hence chose the new label. I spent some effort and tried to simulate this flow and as suspected got below result using valgrind
```
==31371== Conditional jump or move depends on uninitialised value(s)
==31371==    at 0x53E77A: free_db_attr_list (db_postgres_attr.c:222)
==31371==    by 0x53BFAF: pg_db_reset_que (db_postgres_que.c:372)
==31371==    by 0x535F84: pbs_db_reset_obj (db_postgres_impl.c:903)
==31371==    by 0x4844C5: que_recov_db (queue_recov_db.c:243)
==31371==    by 0x47A431: pbsd_init (pbsd_init.c:889)
==31371==    by 0x47F5E1: main (pbsd_main.c:1691)
==31371==  Uninitialised value was created by a stack allocation
==31371==    at 0x4843F9: que_recov_db (queue_recov_db.c:212)
```

I agree, we can remove the label `db_err_reset`, if we agree on initializing the stack memory `pbs_db_que_info_t	dbque;` to zero in `que_recov_db()` and similar `*_recov_db()` functions?

Shrini-h: In addition to the Conditional jump on uninitialised values, I'm also seeing many invalid reads and invalid free's
I'm attaching this valgrind log below
[valgrind_simulated_error.log](https://github.com/PBSPro/pbspro/files/3828701/valgrind_simulated_error.log)

subhasisb: Yes, so the attribute array is uninitialized before read, we can actually initialize it pretty easily. 
If we could initialize the dbque attribute list thus:
	dbque.attr_list.attributes = NULL;
	dbque.attr_list.attr_count = 0;

So, your set of lines at the top of que_recov_db() (similarly for jobs and other objects as well) would then be:
	obj.pbs_db_obj_type = PBS_DB_QUEUE;
	obj.pbs_db_un.pbs_db_que = &dbque;
	dbque.attr_list.attributes = NULL;
	dbque.attr_list.attr_count = 0;


The alternative is to do it inside pg_db_load_que() (similarly for jobs and other objects as well):
	pq->attr_list.attr_count = 0;
	pq->attr_list.attributes = NULL;

This would be better than adding another goto label.

Shrini-h: > If we could initialize the dbque attribute list thus:
> dbque.attr_list.attributes = NULL;
> dbque.attr_list.attr_count = 0;
> 

Or how about this one line init of full struct ? I dont feel the object is too big for it to have a penalty.
`pbs_db_que_info_t dbque = {0};`

subhasisb: Yah, that sounds fine too. Yes, it is not too big to have a penalty.

Shrini-h: Fixed in https://github.com/PBSPro/pbspro/pull/1394/commits/019f046bf1570332e453e9d6451c79abfe7d5bb7

tests repeated and attached in the header!

src/server/job_recov_db.c
Shrini-h: Fixed in https://github.com/PBSPro/pbspro/pull/1394/commits/019f046bf1570332e453e9d6451c79abfe7d5bb7

subhasisb:
Can we get one more review please?

Shrini-h:
Thank you @riyazhakki for taking time for review.

@maintainers PR ready for merge!

PR #1395, opened 2019-11-09
cgroups hook under big cpuset systems not assigning cpuset.cpus,mems correctly

bayucan:
#### Describe Bug or Feature

1. WIth cgroups hook enabled on a big cpuset system with sockets (or vnodes)
   \>= 10, jobs run but the resulting cpuset.cpus and cpuset.mem cgroup files  are incomplete; ids and memory sockets ids that are in the assigned vnodes number 10 and up are missing.
2.  Also, when cgroups exechost_startup hook is invoked to map sockets into vnodes,
   some vnodes fail to be created due to "KeyError". This is traced to a type error exception
    with pbs.size(val) which expects val to be of type int or str, but system has returned a mem, vmem, or hpmem value of type float. 
 

#### Describe Your Change

* For problem 1 above,  in _get_assigned_job_resources() , the pattern used for searching for matching vnodes is just:

   vnhost pattern: **<vnode-name>\[[\d+]\]**

   But this would only match vnode[0...9] so resources from anything above vnode 9, will not be
   written into cpuset.cpus and cpuset.mem files. Fix is to have the '+' in pattern outside of the [] as in: **\[[\d]+\]** so as to match one or more digits, otherwise, it's matching digits or '+' sign.

* For problem 1 also, with pbs_cgroups hook enabled + standard mom, use_hyperthreads = True + vnode_numa_nodes = True, some cpus are not written to the cpuset.cpus file. This is because in function _assign_resources(), 2 lists are consulted: the cores list and the threads list. From these lists, cpus are assigned to populate the cpuset.cpus file. But there are some repeats, for example cores = [90], threads list = [384, 90], and then pbs would select 90, 384, then 90 again in but only 90 and 384 are actually written to cpuset.cpus file. But pbs_cgroups thinks it had already written 3 items. So the fix is to check to make sure to select only cpu ids that have not been actually written in the cpuset.cpus file. 
* For problem 2, if a mem, hpmem, or vmem value is detected to be of type float, just convert it to int.
* Additionally, the pbs_cgroups_hook ptl testsuite has been updated to be able to run on big cpuset systems like the uv300. The updates to the testsuite involve:
   - Allowing pbs_cgroups_hook testsuite to run without memory subsystem configured. On big cpuset systems, memory enforcement can be handled by cpuset instead.
   - Taking into account that on a system with bigger pool of available resources, jobs may not always be assigned cpuIDs=0 or memory socket = 0.

* [ptl.cgroups.r7.txt](https://github.com/PBSPro/pbspro/files/3827477/ptl.cgroups.r7.txt)
* [ptl.cgroups_hook.uv300.txt](https://github.com/PBSPro/pbspro/files/3838548/ptl.cgroups_hook.uv300.txt)
* [ptl.cgroups.s12.txt](https://github.com/PBSPro/pbspro/files/3827479/ptl.cgroups.s12.txt)
* [ptl.test_big_cgroup_cpuset.PASS.txt](https://github.com/PBSPro/pbspro/files/3826779/ptl.test_big_cgroup_cpuset.PASS.txt)
* [ptl.test_big_cgroup_cpuset.FAIL.txt](https://github.com/PBSPro/pbspro/files/3826780/ptl.test_big_cgroup_cpuset.FAIL.txt)

### DIFF
--- a/src/hooks/cgroups/pbs_cgroups.PY
+++ b/src/hooks/cgroups/pbs_cgroups.PY
@@ -1077,7 +1077,7 @@ class JobUtils(object):
             raise CgroupProcessingError('No job information available')
         # Create a list of local vnodes
         vnodes = []
-        vnhost_pattern = r'%s\[[\d+]\]' % hostname
+        vnhost_pattern = r'%s\[[\d]+\]' % hostname
         pbs.logmsg(pbs.EVENT_DEBUG4, '%s: vnhost pattern: %s' %
                    (caller_name(), vnhost_pattern))
         pbs.logmsg(pbs.EVENT_DEBUG4, '%s: Job exec_vnode list: %s' %
@@ -1926,7 +1926,10 @@ class NodeUtils(object):
                 elif key in ['mem', 'vmem', 'hpmem']:
                     # Used for vnodes per NUMA socket
                     if vnodes:
-                        vnode_resc_avail[key] = pbs.size(val)
+                        mem_val = val
+                        if isinstance(val, float):
+                            mem_val = int(val)
+                        vnode_resc_avail[key] = pbs.size(mem_val)
                 elif isinstance(val, list):
                     pass
                 elif isinstance(val, dict):
@@ -3260,9 +3263,13 @@ class CgroupUtils(object):
                     if needed <= 0:
                         break
                     for thread in node.cpuinfo['cpu'][corenum]['threads']:
+                        if thread in assigned['cpuset.cpus']:
+                            # thread core already assigned
+                            continue
                         assigned['cpuset.cpus'].append(thread)
                         needed -= 1
-                        cores.remove(thread)
+                        if thread in cores:
+                            cores.remove(thread)
                         if needed <= 0:
                             break
                     if needed <= 0:
--- a/test/tests/functional/pbs_cgroups_hook.py
+++ b/test/tests/functional/pbs_cgroups_hook.py
@@ -57,6 +57,8 @@ def is_memsw_enabled(mem_path):
     Check if system has swapcontrol enabled, then return true
     else return false
     """
+    if not mem_path:
+        return 'false'
     # List all files and check if memsw files exists
     for files in os.listdir(mem_path):
         if 'memory.memsw' in files:
@@ -87,6 +89,28 @@ def systemd_escape(buf):
     return ret
 
 
+def count_items(items):
+    """
+    Given a comma-separated string of numerical items of either
+    singular value or a range of values (<start>-<stop>),
+    return the actual number of items.
+    For example,
+         items="4-6,9,12-15"
+         count(items) = 8
+    since items expands to "4,5,6,9,12,13,14,15"
+    """
+    ct = 0
+    if items is None:
+        return ct
+    for i in items.split(','):
+        j = i.split('-')
+        if len(j) == 2:
+            ct += len(range(int(j[0]), int(j[1]))) + 1
+        else:
+            ct += 1
+    return ct
+
+
 @tags('mom', 'multi_node')
 class TestCgroupsHook(TestFunctional):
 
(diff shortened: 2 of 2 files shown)


test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -1549,13 +1657,15 @@ def test_cgroup_cpuset_ncpus_are_cores(self):
...
-        if sibs/cores == 1:
+        if sibs / cores == 1:
             self.skipTest('This test requires hyperthreading to be enabled.')
         name = 'CGROUP18'
         self.load_config(self.cfg8 % ('', '', '', self.swapctl, ''))
         # Submit M*N jobs, where M is the amount of physical processors and
         # N is number of 'cpu cores' per M. Expect them to run.
         njobs = phys * cores
+        if njobs > 100:
+            self.skipTest("too many jobs (%d) to submit" % njobs)
```
mike0042: Why skip the test in this case? Would it be better to simply put an upper limit on njobs of 100 and let the test run?

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -1926,7 +1926,10 @@ class NodeUtils(object):
                 elif key in ['mem', 'vmem', 'hpmem']:
                     # Used for vnodes per NUMA socket
                     if vnodes:
-                        vnode_resc_avail[key] = pbs.size(val)
+                        mem_val = val
+                        if isinstance(val, float):
+                            mem_val = int(val)
+                        vnode_resc_avail[key] = pbs.size(mem_val)
```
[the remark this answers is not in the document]
bayucan: I could do it this way, but I'm also worried that 'val' could be a string (not just an int) of the form "<value><size_suffix" (e.g. "124gb"), and pbs.size() also accepts strings of this form and process it accordingly. The function that calls this, create_vnodes(), could be taking the numa_nodes input from a vnodedefs file.

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -1363,10 +1443,24 @@ def test_cgroup_cpuset_and_memory(self):
...
         if self.paths['cpuacct']:
-            self.assertTrue('CpuIDs=0' in tmp_out)
+            cpuids = None
+            for kv in tmp_out:
+                if 'CpuIDs=' in kv:
+                    cpuids = kv.split("=")[1]
+                    break
+            self.assertNotEqual(cpuids, None, 'Could not read CPU IDs.')
+            self.assertEqual(count_items(cpuids), 1,
+                             'Number of CpIDs assigned not 1')
```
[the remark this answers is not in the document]
bayucan: Ok. I'll enhance count_items() to return 0 if parameter is None, and to get rid of the extra self.assertNotEqual() call.

test/tests/functional/pbs_cgroups_hook.py
bayucan: The test, test_cgroup_cpuset_ncpus_are_cores, tries to submit 1-cpu jobs up to the total number of cpu cores (njobs = pys * cores) available, and then check that an additional job would stay queued due to no more resources. So I can't really put a limit, for it's an absolute limit that is needed to run the test; otherwise, the test will fail for the additional job would still run since resources would still be available.

src/hooks/cgroups/pbs_cgroups.PY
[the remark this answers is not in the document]
mike0042: Understood. I'm fine with the change you propose.

test/tests/functional/pbs_cgroups_hook.py
mike0042: In that case, I'm OK with skipping it. I would suggest you pick a power of two rather than a power of ten as the limit. Maybe 64 or 128?

bayucan: Sure, power of 2.

mike0042:
Something is preventing the CLA check from passing @bayucan. I suggest you rebase your branch and force push the update.

bayucan:
@mike0042  Ok, I've rebased, squashed (since I already received 2 approvals), and push the updates. Hope it all passses this time.

PR #1396, opened 2019-11-11
CI tool for developers 

sandisamp:
#### Describe Bug or Feature

This is an introduction of a new tool, which will allow developers to run ci locally.

#### Describe Your Change

* The travis script is now moved to a "ci" directory where it will run inside a container specified by the user.
* A python script for ci will help run and manage pbs in these containers.


ci/do.sh
```diff
@@ -0,0 +1,159 @@
...
+  ./autogen.sh
+  _cflags="-g -O2 -Wall -Werror"
+  if [ "x${ID}" == "xubuntu" ]; then
+    _cflags="${_cflags} -Wno-unused-result"
+  fi
+  cd ${_targetdirname}
+  if [ -f /src/ci ];then
+    if [ -f ${workdir}/.configure_opt ]; then
+      configure_opt="$(cat ${workdir}/.configure_opt)"
+      _cflags="$(echo ${configure_opt} | awk -F'"' '{print $2}')"
```
hirenvadalia: I guess you mean configure_opt instead _cflags here right? else it will override _cflags from line no. 99 and 101

ci/do.sh
```diff
@@ -0,0 +1,159 @@
...
+set +e
+. /etc/profile.d/ptl.sh
+set -e
+pbs_config --make-ug
+
+if [ "x${RUN_TESTS}" == "x1" ];then
+  ptl_tests_dir=$(dirname ${prefix})/ptl/tests
+  cd ${ptl_tests_dir}/
+  pbs_benchpress "$( cat ${workdir}/.benchpress_opt )" --db-type=html --db-name=result -o ${logdir}/logfile
+  mv result.html ${logdir}/
```
hirenvadalia: instead of this, you can pass `--db-name=${logdir}/result.html`

ci/do.sh
sandisamp: It won't override the _cflags unless the user has specified new cflags. The 2 conditionals check wether it's a ci build and wether .configure_opt file is created by the user. This will ensure it only overrides default when user manually specifies.

ci/ci
```diff
@@ -0,0 +1,344 @@
...
+    image_name = read_from_file(env_dir)
+    image_name = image_name.split('=')[1]
+    if image_name.find('_ci_pbs') == -1:
+        image_name += '_ci_pbs'
+        image_name = image_name.replace(':', '_')
+        try:
+            build_id = 'docker-compose ps -q pbs-build'
+            build_id = run_cmd(build_id,True)
+            build_id = build_id.split("'")[1]
+            build_id = build_id[:12]
```
prakashcv13: explain the magic number 12 here.

ci/ci
```diff
@@ -0,0 +1,344 @@
...
+            cmd = 'docker commit '+build_id+' '+image_name+':latest'
+            log_info(cmd)
+            run_cmd(cmd)
+            # remove all bad_images after commit alternatively remove image then commit
```
prakashcv13: we have already run the commit command.

ci/ci
```diff
@@ -0,0 +1,344 @@
...
+                bad_images = (bad_images.split("'")[1]).replace("\\n", " ")
+                print("The following untagged images will be removed -> " + bad_images)
+                cmd = 'docker rmi ' + bad_images
+                run_cmd(cmd)
+        except Exception as e:
+            log_error(e)
+
+
+def delete_older_image(bad_images):
+    if val.find('_ci_pbs') == -1:
```
prakashcv13: what is "val" here?

ci/ci
```diff
@@ -0,0 +1,344 @@
...
+        write_to_file(env_dir, "CONTAINER_PLATFORM="+search_str)
+        return True
+    else:
+        write_to_file(env_dir, "CONTAINER_PLATFORM="+val)
+        return False
+
+
+def check_prerequisites():
+    # This function will check wether docker docker-compose commands are available
+    # or a  container is running that has pbs
```
prakashcv13: two spaces after a.
and shall we say "container that has pbs is running"?

ci/ci
```diff
@@ -0,0 +1,344 @@
...
+
+    try:
+        subprocess.run([cmd, "docker"], stdout=subprocess.DEVNULL)
+    except:
+        log_error("docker not found in PATH")
+
+    try:
+        subprocess.run([cmd, "docker-compose"], stdout=subprocess.DEVNULL)
+    except:
+        log_error("docker-compose not found in PATH")
```
prakashcv13: should we not return from here itself?

ci/ci
```diff
@@ -0,0 +1,344 @@
...
+
+    try:
+        docker_container_process = subprocess.run(
+            ["docker-compose", "ps", "pbs-build"], stdout=subprocess.PIPE)
+        docker_container = str(docker_container_process.stdout)
+        if docker_container.find('Up') == -1:
+            log_error("No running service found")
```
prakashcv13: why is this an error?

ci/ci
```diff
@@ -0,0 +1,344 @@
...
+            try:
+                log_info('Attempting to start container')
+                os.chdir(pbs_dirname)
+                subprocess.run(["docker-compose", "down"],
+                               stdout=subprocess.DEVNULL)
+                image_check = check_for_existing_image()
+                subprocess.run(["docker-compose", "up", "-d"])
```
prakashcv13: we should run only if image_check is True

ci/ci
```diff
@@ -0,0 +1,344 @@
...
+                if image_check == True:
+                    return 0
+                log_info('Waiting for container build to complete ')
+                build_log_path = os.path.join(pbs_dirname, 'logs', 'build')
+                log_info("You can check the logs in " + build_log_path)
```
prakashcv13: I think we should say - "build logs can be found in: "

ci/ci
```diff
@@ -0,0 +1,344 @@
...
+                # wait for build to complete and commit newly built container
+                timer = 1
+                while True:
+                    time.sleep(30)
+                    last_line = tail_file(build_log_path)
+                    log_info("Get current status ...... " + last_line)
```
prakashcv13: i am not sure, if this log is reporting progress to the user.

ci/killit.sh
```diff
@@ -0,0 +1,71 @@
...
+	sleep 10	
+	pid=$(ps -ef 2>/dev/null | grep $1 | grep -v grep | awk '{print $2}')	
+	if [ ! -z "${pid}" ]	
+	then	
+		echo "kill -KILL ${pid}"	
+		kill -KILL ${pid} 2>/dev/null	
+	fi	
+}
+
+check_process(){
```
prakashcv13: kill pbs_processes() would make more sense here.

ci/killit.sh
```diff
@@ -0,0 +1,71 @@
...
+    ps_count=$(ps -eaf 2>/dev/null | grep pbs_ | grep -v grep | wc -l )
```
prakashcv13: we are not using this ps_count.

prakashcv13:
add docstrings for all the new functions added.

ci/ci
```diff
@@ -0,0 +1,344 @@
...
+    except Exception as e:
+        log_error("Failed\n:")
+        log_error(e)
+
+
+def write_to_file(file_path, value):
+    with open(file_path, "w+") as f:
+        f.write(value)
+
+
```
prakashcv13: why 2 lines instead of 1?

ci/ci
sandisamp: It shortens the container id returned by **docker-compose** to the one displayed by **docker** command, as docker-compose command will return full container id.
It will be then be displayed as something below 
`14:21:46 ---> docker commit 50315b45f569 centos_7_ci_pbs:latest`
It's helpful to user as the container id is similar to one shown by command "docker ps"
```
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS               NAMES
50315b45f569        centos:7            "src/docker-entrypoi…"   2 minutes ago       Up 2 minutes                            ci_pbs-build_1
```

ci/ci
sandisamp: Running pep8 on the file produces 2 lines between functions.

ci/ci
sandisamp: No we need not do that, if image exists it will use that image and we wouldn't need to build dependencies, 
Image check returning False only means to build dependencies, not to stop the services all together.

ci/ci
sandisamp: This is not reporting progress exactly to the user, for that the user must go to the log file, however it gives some idea as to what is the current status of the build as opposed to a blinking cursor on the screen for some 10 mins.
```
14:50:12 ---> Waiting for container build to complete 
14:50:12 ---> Build logs can be found in /home/sanidhya/instantCI/pbspro/ci/logs/build
14:50:42 ---> Get current status ...... Public key for device-mapper-1.02.158-2.el7_7.2.x86_64.rpm is not installed

14:51:12 ---> Get current status ......   Cleanup    : bash-4.2.46-31.el7.x86_64                                124/130
14:51:42 ---> Get current status ......  * updates: centos.myfahim.com

14:52:12 ---> Get current status ......   Installing : openssh-clients-7.4p1-21.el7.x86_64                        38/57
14:52:42 ---> Get current status ...... Downloading packages:

14:53:12 ---> Get current status ......   Verifying  : libXau-devel-1.0.8-2.1.el7.x86_64                          73/73
14:53:42 ---> Get current status ...... ### First time build is complete ###

14:53:42 ---> build is now complete
14:53:43 ---> docker commit b6136476d886 centos_7_ci_pbs:latest
14:53:50 --->  done start container 
```
Any alternatives would be welcome as well, do you have better suggestions?

ci/ci
prakashcv13: okay. thank you for the explanation. we need to add a comment in the code to explain why are we using 12 instead of any other number.

ci/ci
prakashcv13: you are calling check_for_existing_image(), starting the containers and then checking the value of image_check.
is there a reason to do that?

ci/ci
prakashcv13: hmmm, what you say is like reporting progress...
let us remove "Get" from the log...

ci/ci
```diff
@@ -0,0 +1,367 @@
...
+                while True:
+                    time.sleep(30)
+                    last_line = tail_file(build_log_path)
+                    log_info("Get current status ...... " + last_line)
+                    if last_line.find('### First time build is complete ###') != -1:
+                        log_info("build is now complete")
+                        commit_docker_image()
+                        return 0
+                    timer += 1
+                    if timer > 120:
```
prakashcv13: add a comment "time-out after 60 minutes" to the above line.

ci/ci
```diff
@@ -0,0 +1,367 @@
...
+        args = ap.parse_args(sys.argv[1:])
+    except Exception as e:
+        log_error(e)
+    if args.pbs_dir != None:
+        pbs_dirname = os.path.join(args.pbs_dir, 'ci')
+        try:
+            os.chdir(pbs_dirname)
+        except Exception as e:
+            log_error(e)
+            log_error("Was correct pbs path provided")
```
prakashcv13: looks like a question, should we add a '?'

ci/ci
```diff
@@ -0,0 +1,367 @@
...
+    except:
+        log_error("failed to start ci services")
+
+
+if __name__ == "__main__":
+
+    pbs_dirname = find_ci()
+    pbs_dirname = os.path.join(pbs_dirname, 'ci')
+    os.chdir(pbs_dirname)
+    build_log_path = os.path.join(pbs_dirname, 'logs', 'build')
```
prakashcv13: pbs_dirname changes if we have args.pbs_dir, will that effect build_log_path?

ci/ci
```diff
@@ -0,0 +1,367 @@
...
+    try:
+        if args.configure != None:
+            val = args.configure
+            configure_opt = os.path.join(pbs_dirname, '.configure_opt')
+            if val == 'called':
+                val = read_from_file(configure_opt)
+                if val == '':
+                    val = '--prefix=/opt/pbs --enable-ptl'
+            log_info("using configure parameters : " + val)
+            write_to_file(configure_opt, val)
```
prakashcv13: we need not call write_to_file() if read_from_file() returned a non-empty string.

ci/ci
sandisamp: This comment was part of an earlier code which is no longer there so i have removed it. It does not indicate any functionality of the present code.

ci/ci
```diff
@@ -0,0 +1,367 @@
...
+            log_info(" done install ")
+        if args.test != None:
+            val = args.test
+            benchpress_opt = os.path.join(pbs_dirname, '.benchpress_opt')
+            if val == 'called':
+                val = read_from_file(benchpress_opt)
+                if val == '':
+                    val = '--tags=smoke'
+            log_info("using test parameters : " + val)
+            write_to_file(benchpress_opt, val)
```
prakashcv13: we need not call write_to_file() if read_from_file() returned a non-empty string.

ci/ci
```diff
@@ -0,0 +1,367 @@
...
+            benchpress_opt = os.path.join(pbs_dirname, '.benchpress_opt')
+            val = ''
+            if args.run == 'called':
+                val = read_from_file(benchpress_opt)
+                if val == '':
+                    val = '--tags=smoke'
+            else:
+                val = args.run
+            log_info("using test parameters : " + val)
+            write_to_file(benchpress_opt, val)
```
prakashcv13: we need not call write_to_file() if read_from_file() returned a non-empty string.

ci/ci
sandisamp: No i guess if you are asking if check image can be called after 197, then yes. it would work too.

ci/ci
```diff
@@ -279,13 +280,17 @@ if __name__ == "__main__":
...
             if val == 'called':
                 val = read_from_file(configure_opt)
                 if val == '':
                     val = '--prefix=/opt/pbs --enable-ptl'
+                else:
+                    write = False
             log_info("using configure parameters : " + val)
-            write_to_file(configure_opt, val)
+            if write:
+                write_to_file(configure_opt, val)
```
prakashcv13: why don't you just move the call to write_to_file() to lije 289 and do away with the else part and write variable?

sandisamp: This is because val == 'called' comes into picture when no value is assigned to the option '--configure' but when there is some value assigned then like --configure='--prefix=/tmp/pbs --enable-ptl' then value in val will be the value assigned and we still need to update this value to the file.

sandisamp:
Adding test logs:
[CI_LOGS.txt](https://github.com/PBSPro/pbspro/files/3857882/CI_LOGS.txt)

PR #1281, opened 2019-09-05
PP-468: Kerberos support with GSSAPI

vchlum:
#### Describe Bug or Feature
Kerberos support with GSSAPI. Work includes encryption on TCP, TPP, and distributing of Kerberos credentials for jobs.

#### Describe Your Change
Please, see the [design document](https://pbspro.atlassian.net/wiki/spaces/PD/pages/29655043/PP-468+Kerberos+support). It includes implementation notes.

Information about GSSAPI can be found [here](https://docs.oracle.com/cd/E19683-01/816-1331/index.html):

#### Link to Design Doc
[project documentation](https://pbspro.atlassian.net/wiki/spaces/PD/pages/29655043/PP-468+Kerberos+support)

[forum link](http://community.pbspro.org/t/kerberos-support)

#### Attach Test and Valgrind Logs/Output
[PTL smoketest log with disabled Kerberos](https://github.com/PBSPro/pbspro/files/3578902/ptl_smoke_kerberos_support_disabled.txt)

As agreed in the [forum](http://community.pbspro.org/t/kerberos-support/48/40), automated test will be added later.

Manual test1: Debian 8, Heimdal, AFS:

```
vchlum@torque1:~$ qsub  -I -l select=2 -l place=scatter
qsub: waiting for job 113.torque1.ics.muni.cz to start
qsub: job 113.torque1.ics.muni.cz ready

(JESSIE)vchlum@torque1:~$ klist
Credentials cache: FILE:/tmp/krb5cc_pbsjob_113.torque1.ics.muni.cz
        Principal: vchlum@META

  Issued                Expires               Principal
Sep  5 12:58:21 2019  Sep  5 22:58:21 2019  krbtgt/META@META
Sep  5 12:58:21 2019  Sep  5 22:58:21 2019  afs/ics.muni.cz@META
Sep  5 12:58:21 2019  Sep  5 22:58:21 2019  krbtgt/ZCU.CZ@META
Sep  5 12:58:21 2019  Sep  5 22:58:21 2019  afs/zcu.cz@ZCU.CZ
Sep  5 12:58:21 2019  Sep  5 22:58:21 2019  krbtgt/RUK.CUNI.CZ@META
Sep  5 12:58:21 2019  Sep  5 22:58:21 2019  afs/ruk.cuni.cz@RUK.CUNI.CZ
(JESSIE)vchlum@torque1:~$ pbsdsh klist
[...]

PR #1397, opened 2019-11-13
cached credentials not used with kerberos

vchlum:
#### Describe Bug or Feature
This bug was introduced during the review process #1281. I changed strcpy to strncpy but I forgot that strncpy will not implicitly copy null termination of the string. This means the cred->credid was not terminated. cred->credid is used for finding credentials in the cache. Since cred->credid was not terminated, the cached credentials were never found and new credentials were always created.

#### Describe Your Change
Include  '\0' in strncpy while caching credentials.

Manual test:
* First run of a job for a particular user logs acquiring credentials: 
 11/13/2019 07:43:58;0080;Server@torque1;Svr;Server@torque1;using cred_renew_tool '/usr/bin/krb525_renew' to acquire credentials for user: vchlum@META
* another job of the same user uses the credential without acquiring a new one

### DIFF
--- a/src/server/req_cred.c
+++ b/src/server/req_cred.c
@@ -184,7 +184,7 @@ get_cached_cred(job  *pjob)
 	}
 
 	strncpy(cred->credid, pjob->ji_wattr[(int)JOB_ATR_cred_id].at_val.at_str,
-		strlen(pjob->ji_wattr[(int)JOB_ATR_cred_id].at_val.at_str));
+		strlen(pjob->ji_wattr[(int)JOB_ATR_cred_id].at_val.at_str) + 1); /* +1 for '\0' */
 	cred->type = cred_type;
 	cred->validity = validity;
 	cred->size = strlen(buf);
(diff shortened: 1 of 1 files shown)


src/server/req_cred.c
```diff
@@ -184,7 +184,7 @@ get_cached_cred(job  *pjob)
 	}
 
 	strncpy(cred->credid, pjob->ji_wattr[(int)JOB_ATR_cred_id].at_val.at_str,
-		strlen(pjob->ji_wattr[(int)JOB_ATR_cred_id].at_val.at_str));
+		strlen(pjob->ji_wattr[(int)JOB_ATR_cred_id].at_val.at_str) + 1); /* +1 for '\0' */
```
subhasisb: This use of strncpy does not seem to be any more safe that a direct strcpy. This intends to copy ALL the bytes from the source string to the destination irrespective of the size of the destination buffer. strncpy is useful only if we use the destination size in the 3rd argument in an effort to not cross of the destination length, and in that case we always explicitly  put a null character at the last position of the destination string.

vchlum: Thanks, it makes sense. Changed.

PR #1398, opened 2019-11-13
Test Framework should allow testers to run a test for multiple times

birbalsain:
#### Describe Bug or Feature

Test Framework should allow testers to run a test for multiple times. Tester can use delay between two repetition.

#### Describe Your Change

Introduce new pbs_benchpress option --repeat_tests=<int<int>> for how many times test will run and
 --repeat_tests_delay for delay=<seconds<seconds>> delay between repeated test. 

Each repetition includes all submitted tests. next repetition start after finishing of all tests of first repetition.
If submits three tests then in first repetition all three tests run and after finishing all three tests, next repetition starts and again all three tests run.

At the end we get message " All tests are repeated <number of repetition> times"

#### Link to Design Doc

https://pbspro.atlassian.net/wiki/spaces/PD/pages/1386938446/Test+Framework+should+allow+testers+to+run+a+test+for+multiple+times

[repetition_single_test.txt](https://github.com/openpbs/openpbs/files/4936258/repetition_single_test.txt)
[repetition_single_test_json.txt](https://github.com/openpbs/openpbs/files/4936261/repetition_single_test_json.txt)
[repetition_mixed_tests.txt](https://github.com/openpbs/openpbs/files/4936263/repetition_mixed_tests.txt)
[repetition_mixed_tests_json.txt](https://github.com/openpbs/openpbs/files/4936264/repetition_mixed_tests_json.txt)

### DIFF
--- a/test/fw/doc/howtotest.rst
+++ b/test/fw/doc/howtotest.rst
@@ -62,6 +62,7 @@ PBSTestSuite offers the following:
     - conn_timeout: set a timeout in seconds after which a pbs_connect IFL call is refreshed (i.e., disconnected)
     - skip-setup: Bypasses setUp of PBSTestSuite (not custom ones)
     - skip-teardown: Bypasses tearDown of PBSTestSuite (not custom ones)
+    - test-repitation: How many times test repeat, It depends on the value of test-repitation
     - procinfo: Enables process monitoring thread, logged into ptl_proc_info test metrics.
     - procmon: Colon-separated process name to monitor. For example to monitor server, sched, and mom use procmon=pbs_server:pbs_sched:pbs_mom
     - procmon-freq: Sets a polling frequency for the process monitoring tool. Defaults to 10 seconds.
--- a/test/fw/doc/howtotest.rst
+++ b/test/fw/doc/howtotest.rst
@@ -62,7 +62,8 @@ PBSTestSuite offers the following:
     - conn_timeout: set a timeout in seconds after which a pbs_connect IFL call is refreshed (i.e., disconnected)
     - skip-setup: Bypasses setUp of PBSTestSuite (not custom ones)
     - skip-teardown: Bypasses tearDown of PBSTestSuite (not custom ones)
-    - test-repitation: How many times test repeat, It depends on the value of test-repitation
+    - repeat-tests: How many times test repeat
+    - repeat-tests-delay: How many seconds delay between two repeated test
     - procinfo: Enables process monitoring thread, logged into ptl_proc_info test metrics.
     - procmon: Colon-separated process name to monitor. For example to monitor server, sched, and mom use procmon=pbs_server:pbs_sched:pbs_mom
     - procmon-freq: Sets a polling frequency for the process monitoring tool. Defaults to 10 seconds.
--- a/test/fw/ptl/utils/plugins/ptl_test_runner.py
+++ b/test/fw/ptl/utils/plugins/ptl_test_runner.py
@@ -467,9 +467,11 @@ class PtlTextTestRunner(TextTestRunner):
     """
 
     def __init__(self, stream=sys.stdout, descriptions=True, verbosity=3,
-                 config=None):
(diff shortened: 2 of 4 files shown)


agrawalravi90:
Hi @birbalsain can you please explain the use cases for this feature? Why would one want to run a test repeatedly ?

test/fw/doc/howtotest.rst
```diff
@@ -62,6 +62,7 @@ PBSTestSuite offers the following:
     - conn_timeout: set a timeout in seconds after which a pbs_connect IFL call is refreshed (i.e., disconnected)
     - skip-setup: Bypasses setUp of PBSTestSuite (not custom ones)
     - skip-teardown: Bypasses tearDown of PBSTestSuite (not custom ones)
+    - test-repitation: How many times test repeat, It depends on the value of test-repitation
```
shilpakodli: Please correct repitation->repitition

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -518,6 +519,9 @@ class PTLTestRunner(Plugin):
     def __init__(self):
         Plugin.__init__(self)
         self.param = None
+        self.repeat_tests = 1
+        self.test_repeat = 1
```
shilpakodli: You can rename self.test_repeat name to something like self.default_test_run_count.

birbalsain:
@agrawalravi90 Test repetition is used to find race condition, to reproduce bug and run stress tests.

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -518,6 +519,9 @@ class PTLTestRunner(Plugin):
...
+        self.repeat_test_delay = 30
```
shilpakodli: I think 30 seconds delay is little bit high between tests run. Is there any reason behind this value ? It can be reduced to 10 seconds or even 5 seconds.

test/fw/ptl/utils/plugins/ptl_test_runner.py
birbalsain: @shilpakodli I think self.test_run_count is also ok.

agrawalravi90:
> @agrawalravi90 Test repetition is used to find race condition, to reproduce bug and run stress tests.

Please add this as a "Motivation" section on your design document

vstumpf:
Is there a way to disable all this monitoring? Sometimes you just want to run a ptl test and see if it passes/fails...

RKORavi:
@birbalsain If repetition is 3 and test fails at 2nd run, will it mark the test as fail, and will go to next run or will it skip that test all along.
Please answer the same in forum as it has already been asked by Vishwa.

vishwaks:
@birbalsain: I was going through the test logs attached, for the test "test_maintenance_duplicate_host" it ran 3 times and in the status, after repeat cycle, it shows failed 2 times which means it got passed 1 time?
Having different summary for each time is not good (for me), so after the repeat cycle, it's better to have a total repeat cycle summary like below:

repeat run: 3, succeeded: 1, failed: 2, errors: 0, skipped: 0, timedout: 0

birbalsain:
> @birbalsain: I was going through the test logs attached, for the test "test_maintenance_duplicate_host" it ran 3 times and in the status, after repeat cycle, it shows failed 2 times which means it got passed 1 time?
It doesn't mean third is passed. we get last repetition summary at the end. but we get error/skip message for all repetition just after teardown if they have been skipped/failed otherwise ok message for passed.

agrawalravi90:
@birbalsain is this PR still being worked on? There's been no activity on this since Nov 2019.

birbalsain:
Because of some other tasks are going on but it will be started in next few weeks.

birbalsain:
A new PR "https://github.com/openpbs/openpbs/pull/1908" is raised. changes scenario is different.

PR #1401, opened 2019-11-15
Fix for race of TestMultipleSchedulers.test_auto_sched_off_due_to_fds_limit

SudeshnaMoulik:
#### Describe Bug or Feature
TestMultipleSchedulers.test_auto_sched_off_due_to_fds_limit fails in server system  
- where system needs to connect to multiple remote client and mom in a cluster to run PTL 
- pbs_benchpress runs with "-o, --db-name, --db-type" options.
Example: 
PTL_SUDO_CMD= PBS_CONF_FILE=/etc/pbs.conf2 pbs_benchpress -t TestHostResources.test_set_direct_on_target_node,TestMultipleSchedulers.test_auto_sched_off_due_to_fds_limit  --param-file paramfile -o out.txt --db-name=out.html --db-type=html

This was failing because of race condition that occurred as we were trying to get new fd while creating and setting scheduler, after the system's  fd limit exhausted.

#### Describe Your Change
- create and setup scheduler and its attributes before setting system fd limit
- check for scheduler attribute after re-setting back the fd limit

[before_fix.txt](https://github.com/PBSPro/pbspro/files/3850116/before_fix.txt)
[after_fix.txt](https://github.com/PBSPro/pbspro/files/3850117/after_fix.txt)

Build across all platform is successful (20.0.0.20191114090204)
Regression along with this test successful.

### DIFF
--- a/test/tests/functional/pbs_multi_sched.py
+++ b/test/tests/functional/pbs_multi_sched.py
@@ -1833,6 +1833,11 @@ class TestMultipleSchedulers(TestFunctional):
         if os.getuid() != 0 or sys.platform in ('cygwin', 'win32'):
             self.skipTest("Test need to run as root")
 
+        self.setup_sc3()
+        self.server.manager(MGR_CMD_SET, SCHED,
+                            {'scheduler_iteration': 1}, id="sc3")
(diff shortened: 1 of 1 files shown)


SudeshnaMoulik:
Thanks @sujatapatnaik52 for reviewing this PR

PR #1402, opened 2019-11-18
Fix tests that pass environment variables to remote host cluster test systems	

RKORavi:
#### Describe Bug or Feature
When job was submitted from remote host(server on one node & mom on another node), environment variables were not set in same ssh session. It resulted in environment variable not being reflected in qstat -f output.

#### Describe Your Change
1. Changed submitting job from current user to TEST_USER, because in some platform it results in submitting job as root causes problem.
2. Changed pbs_testlib.py so that qsub can be done from script file.
3. Exporting environment variable & submitting job from the same script file.
4. To match if environment variable was set properly, job's output file is copied back to host on which PTL is running.
Reason:
When contents of job output file is printed on console, it results in string mismatch. e.g, If carriage return "\r" is present as text in the file then it will give a different text when it is printed on console.
For this reason it is assumed that there is a passwordless ssh setup between current user(root) where PTL is running and TEST_USER where mom is running. 

#### Link to Design Doc
NA

[remote_job_submission.txt](https://github.com/PBSPro/pbspro/files/3858059/remote_job_submission.txt)
[server_mom_on_same_host.txt](https://github.com/PBSPro/pbspro/files/3858060/server_mom_on_same_host.txt)

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -5984,7 +5984,8 @@ class Server(PBSService):
             continue
         return ij.jobid
 
-    def submit(self, obj, script=None, extend=None, submit_dir=None):
+    def submit(self, obj, script=None, extend=None, submit_dir=None,
+               env=None):
         """
         Submit a job or reservation. Returns a job identifier
         or raises PbsSubmitError on error
@@ -6005,7 +6006,6 @@ class Server(PBSService):
         _interactive_job = False
         as_script = False
         rc = None
-
         if isinstance(obj, Job):
             if self.platform == 'cray' or self.platform == 'craysim':
                 m = False
@@ -6062,14 +6062,26 @@ class Server(PBSService):
                 os.chdir(submit_dir)
         c = None
         # 1- Submission using the command line tools
+        runcmd = []
+        if env:
+            runcmd += ['#!/bin/bash\n']
+            for k, v in env.items():
+                if '()' in k:
+                    f_name = k.replace('()', '')
+                    runcmd += [k, v, "\n", "export", "-f", f_name]
+                else:
+                    runcmd += ['export %s=\"%s\"' % (k, v)]
+                runcmd += ["\n"]
+
+        script_file = None
         if self.get_op_mode() == PTL_CLI:
             exclude_attrs = []  # list of attributes to not convert to CLI
             if isinstance(obj, Job):
-                runcmd = [os.path.join(self.client_conf['PBS_EXEC'], 'bin',
-                                       'qsub')]
+                runcmd += [os.path.join(self.client_conf['PBS_EXEC'], 'bin',
+                                        'qsub')]
             elif isinstance(obj, Reservation):
-                runcmd = [os.path.join(self.client_conf['PBS_EXEC'], 'bin',
-                                       'pbs_rsub')]
+                runcmd += [os.path.join(self.client_conf['PBS_EXEC'], 'bin',
+                                        'pbs_rsub')]
                 if ATTR_resv_start in obj.custom_attrs:
                     start = obj.custom_attrs[ATTR_resv_start]
                     obj.custom_attrs[ATTR_resv_start] = \
@@ -6115,7 +6127,6 @@ class Server(PBSService):
                 except OSError:
                     pass
                 return None
-
             runcmd += cmd
 
             if script:
@@ -6154,10 +6165,17 @@ class Server(PBSService):
                 runcmd = [
                     'PBS_CONF_FILE=' + self.client_pbs_conf_file] + runcmd
                 as_script = True
-
+            if env:
+                user = PbsUser.get_user(obj.username)
+                host = user.host
+                run_str = " ".join(runcmd)
+                script_file = self.du.create_temp_file(hostname=host,
+                                                       body=run_str)
+                self.du.chmod(hostname=host, path=script_file, mode=0o755)
+                runcmd = [script_file]
             ret = self.du.run_cmd(self.client, runcmd, runas=runas,
                                   level=logging.INFOCLI, as_script=as_script,
-                                  logerr=False)
+                                  env=env, logerr=False)
             if ret['rc'] != 0:
                 objid = None
             else:
--- a/test/tests/functional/pbs_nonprint_characters.py
+++ b/test/tests/functional/pbs_nonprint_characters.py
@@ -40,7 +40,6 @@ import json
 import os
 
 
-@skipOnShasta
 class TestNonprintingCharacters(TestFunctional):
     """
     Test to check passing non-printable environment variables
@@ -90,7 +89,7 @@ unset -f foo
 exit 0
 """
         self.script = """#PBS -V
-env | grep -A2 foo
+env | grep -A2 BASH_FUNC_foo
 foo
 sleep 5
 """
@@ -115,14 +114,13 @@ sleep 5
                                          'bin', 'pbsnodes')
 
     def create_and_submit_job(self, user=None, attribs=None, content=None,
-                              content_interactive=None, preserve_env=False):
+                              content_interactive=None, preserve_env=False,
+                              set_env={}):
         """
         Create the job object and submit it to the server as 'user',
         attributes list 'attribs' script 'content' or 'content_interactive',
         and to 'preserve_env' if interactive job.
         """
-        # A user=None value means job will be executed by current user
-        # where the environment is set up
         if attribs is None:
             use_attribs = {}
         else:
@@ -134,21 +132,22 @@ sleep 5
         elif content_interactive is not None:
             retjob.interactive_script = content_interactive
             retjob.preserve_env = preserve_env
+        return self.server.submit(retjob, env=set_env)
 
-        return self.server.submit(retjob)
-
-    def check_jobout(self, chk_var, jid, job_outfile):
+    def check_jobout(self, chk_var, jid, job_outfile, host=None):
         """
         Check if unescaped variable is in job output
         """
         self.server.expect(JOB, 'queue', op=UNSET, id=jid, offset=1)
-        ret = self.du.cat(filename=job_outfile)
-        job_out = '\n'.join(ret['out'])
-        self.logger.info('job output from %s:\n%s' % (job_outfile, job_out))
-        job_output = ""
-        with open(job_outfile, 'r', newline="") as f:
-            job_output = f.read().strip()
-        self.assertEqual(job_output, chk_var)
+        ret = self.du.cat(hostname=host, sudo=True, filename=job_outfile,
+                          option="-v")
+        j_output = ""
+        if len(ret['out']) > 0:
+            if len(ret['out']) > 1:
+                j_output = '\n'.join(ret['out'])
+            else:
+                j_output = ret['out'][0].strip()
+        self.assertEqual(j_output, chk_var)
         self.logger.info('job output has: %s' % chk_var)
 
     def check_qstatout(self, chk_var, jid):
@@ -157,8 +156,13 @@ sleep 5
         """
         cmd = [self.qstat_cmd, '-f', jid]
         ret = self.du.run_cmd(self.server.hostname, cmd=cmd)
-        qstat_out = '\n'.join(ret['out'])
-        self.assertIn(chk_var, qstat_out)
+        k = chk_var.split('=')[0]
+        for elem in ret['out']:
+            if k in elem:
+                i = ret['out'].index(elem)
+                job_str = elem.strip('\t') + ret['out'][i + 1].strip('\t')
+                break
+        self.assertIn(chk_var, job_str)
         self.logger.info('qstat -f output has: %s' % chk_var)
 
     def test_nonprint_character_qsubv(self):
@@ -169,6 +173,7 @@ sleep 5
         qsub -v "var1='A,B,<non-printable character>,C,D'"
         and check that the value with the character is passed correctly
         """
+        uhost = PbsUser.get_user(TEST_USER).host
         for ch in self.npcat:
             self.logger.info('##### non-printable char: %s #####' % repr(ch))
             if ch in self.npch_exclude:
@@ -177,8 +182,11 @@ sleep 5
             # variable to check if with escaped nonprinting character or not
             chk_var = r'var1=A\,B\,%s\,C\,D' % self.npcat[ch]
             if ch in self.npch_asis:
-                chk_var = 'var1=A\,B\,%s\,C\,D' % ch
-            a = {ATTR_v: "var1=\'A,B,%s,C,D\'" % ch}
+                chk_var = r'var1=A\,B\,%s\,C\,D' % ch
+            if uhost is None or self.du.is_localhost(uhost):
+                a = {ATTR_v: "var1=\'A,B,%s,C,D\'" % ch}
+            else:
+                a = {ATTR_v: r"var1=\'A\,B\,%s\,C\,D\'" % ch}
             script = ['sleep 5']
             script += ['env | grep var1']
             jid = self.create_and_submit_job(attribs=a, content=script)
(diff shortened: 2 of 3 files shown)


test/tests/functional/pbs_nonprint_characters.py
```diff
@@ -271,16 +280,22 @@ def test_nonprint_character_default_qsubV(self):
...
             script += ['env | grep NONPRINT_VAR']
-            j = Job(self.du.get_current_user())
+            j = Job(TEST_USER)
             j.create_script(body=script)
-            jid = self.server.submit(j)
+            exp = "export NONPRINT_VAR=\"X%sY\"" % ch
+            file_n = j.create_script(body=script)
+            script = [exp]
+            script += ['/opt/pbs/bin/qsub %s' % file_n]
+            j.create_script(body=script)
```
bhroam: I'm not sure I understand this.  Why are you calling j.create_script() twice?  Won't the second call overwrite the first?

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -5981,8 +5982,9 @@ def submit(self, obj, script=None, extend=None, submit_dir=None):
         c = None
         # 1- Submission using the command line tools
         if self.get_op_mode() == PTL_CLI:
+            runcmd = []
             exclude_attrs = []  # list of attributes to not convert to CLI
-            if isinstance(obj, Job):
+            if isinstance(obj, Job) and not qsub_from_script:
```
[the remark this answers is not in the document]
RKORavi: Hi Bhroam,
I did try out putting env on command line. But since we are setting special characters like \t, \r in this test, it starts to error out. That is the reason i went with using script file to set environment variable.

test/tests/functional/pbs_nonprint_characters.py
RKORavi: Yes, first script is copied to remote host & is being called by second script.
First script contains - script1
sleep 5
env | grep NONPRINT_VAR

second script - 
export NONPRINT_VAR=value
/opt/pbs/bin/qsub script1
So, that environment variable is set in same session

test/fw/ptl/lib/pbs_testlib.py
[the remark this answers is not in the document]
bhroam: If the special characters are quoted properly, it should work.  If you can't get environmental variables on the command line to work, I still think the env dictionary is the right way to go.  You just have submit use du.create_temp_file() on the correct host and create the shell script that is used to submit the job.

test/tests/functional/pbs_nonprint_characters.py
bhroam: As with my above comment, use du.create_temp_file()

test/fw/ptl/lib/pbs_testlib.py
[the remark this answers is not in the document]
RKORavi: Thanks Broam, looks like i was not quoting the special characters properly. I will make the changes in submit.

RKORavi:
There are other test cases in TestNonprintingCharacters with similar problem that i have missed, I will mark the PR as WIP for now.

test/fw/ptl/lib/pbs_testlib.py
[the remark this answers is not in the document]
RKORavi: Had to use script file to set environment variable. Reason being we cannot set a function as environment variable due to shell shock malware fix

RKORavi:
[logs_mom_qub_on_remote_host.txt](https://github.com/PBSPro/pbspro/files/3903925/logs_mom_qub_on_remote_host.txt)

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -5981,15 +5981,31 @@ def submit(self, obj, script=None, extend=None, submit_dir=None,
...
+        runcmd = []
+        exp_str = ""
+        if len(env_set) > 0:
+            if '()' in list(env_set.keys())[0]:
+                func = list(env_set.keys())[0]
+                f_body = list(env_set.values())[0]
+                f_name = func.replace('()', '')
+                runcmd += [func, f_body, "\n", "export", "-f", f_name]
+            else:
+                exp_str = "export "
```
bhroam: You don't need export.  You can say 'VAR=val ./a.out' and VAR will be set as an environmental variable.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -6071,10 +6086,18 @@ def submit(self, obj, script=None, extend=None, submit_dir=None,
                 runcmd = [
                     'PBS_CONF_FILE=' + self.client_pbs_conf_file] + runcmd
                 as_script = True
-
+            if len(env_set) > 0:
+                user = PbsUser.get_user(TEST_USER)
```
bhroam: TEST_USER is the default user.  Shouldn't you use the user from the job/resv object?

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -5981,15 +5981,31 @@ def submit(self, obj, script=None, extend=None, submit_dir=None,
...
         c = None
         # 1- Submission using the command line tools
+        runcmd = []
+        exp_str = ""
+        if len(env_set) > 0:
+            if '()' in list(env_set.keys())[0]:
+                func = list(env_set.keys())[0]
+                f_body = list(env_set.values())[0]
+                f_name = func.replace('()', '')
+                runcmd += [func, f_body, "\n", "export", "-f", f_name]
```
bhroam: I'm not sure I fully understand this.  This only checks the first element of the env_set if it is a function.  What if a further element is?

test/fw/ptl/lib/pbs_testlib.py
RKORavi: Hi Bhroam,

I tried the following as: 
#!/bin/bash
NONPRINT_VAR="X^AY"
/opt/pbs/bin/qsub /tmp/PtlPbsJobScriptbpfw37yc

Environment variable NONPRINT_VAR was not seen in qstat -f output.
I could see the environment variable only when i add export NONPRINT_VAR="X^AY"

RKORavi:
@bhroam If possible, could you please review the PR with latest changes.

test/fw/ptl/lib/pbs_testlib.py
bhroam: Yes, it won't show up if you create a script.  We had previously talked about not using a script and adding the variables directly to the command line.  VAR=val ./a.out is exported as an environmental variable.

bhroam:
My only comment left is about using a script vs puting the env vars on the command line.  Once we finish that discussion, I'll sign off.  I'm sorry about the delay, I must have missed the email when you pushed.

RKORavi:
> My only comment left is about using a script vs puting the env vars on the command line. Once we finish that discussion, I'll sign off. I'm sorry about the delay, I must have missed the email when you pushed.

Hi @bhroam,
Sorry for delay in reply but could you please clear couple of concerns that i have:

1. when we do not use export & try to set a variable as VAR=val. we will be setting shell variable instead of environment variable. If a shell variable is set it will not be reflected in child process.
2. when we have to add functions as environment variable we have to use export:
ex: $ name () { echo "value print"; }
$ export -f name

bhroam:
In a script, if you do not use export, you create a shell variable.  You are correct.  This is not what I am suggesting.  If you run a program, you can define environmental variables on the command line.
Example:
% VAR=val ./a.out
For that run of a.out, the VAR environmental variable will be set to 'val'.  No script or export needed.  This is what I am suggesting we do.  We craft the run_cmd using environmental variables on the command line itself.  This saves us from writing a script.

If we need to add functions, I am unsure if this method will work.

RKORavi:
> In a script, if you do not use export, you create a shell variable. You are correct. This is not what I am suggesting. If you run a program, you can define environmental variables on the command line.
> Example:
> % VAR=val ./a.out
> For that run of a.out, the VAR environmental variable will be set to 'val'. No script or export needed. This is what I am suggesting we do. We craft the run_cmd using environmental variables on the command line itself. This saves us from writing a script.
> 
> If we need to add functions, I am unsure if this method will work.

Thank you for explaining this for me. I will try implementing the changes in code. I tried too add function and it seems to work.
[root@mom4 pbsuser]# name () { echo "value print"; }; qsub -V job_sc 
7038.ravi

I tried to grep "name" in job script(printenv | grep "name") and got the following in output file 
[root@mom4 pbsuser]# cat job_sc.o7038
BASH_FUNC_name()=() {  echo "value print"

RKORavi:
> % VAR=val ./a.out

Hi @bhroam, as per out discussion i was trying to make it working for shell function. But i have hit one roadblock. in previous example where i tried to show "name () { echo "value print"; }; qsub -V job_sc". It was working because i had "name()" shell function exported earlier and i did not notice that while running the command. 
To make the command working i had to make one small change as: 
name () {echo "value print";}; **export -f name**; qsub -V job_sc.

But due to this i am unable to make it working for PTL, since it will prefix the command with : sudo -H -u pbsuser [rest of the command here]. To make this work with sudo some workaround needs to be added and it might cause security issue.

I am unable to bypass this problem. If possible could you please suggest how should i proceed next.
Thanks

RKORavi:
Latest [commit](https://github.com/PBSPro/pbspro/pull/1402/commits/6b1e4024586569857bfc8e89116ab9d561fc4c62)

Include the following changes:
1. Instead of trying to open Job's output file and and check for NON_PRINTING character, I am changing it to use cat command with -v option (prints NON_PRINTING character except tab spaces).
2. Interactive job is not working as expected looks like a Bug in PTL, but i am investigating further to be sure.

RKORavi:
@hirenvadalia could you also review the PR

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -5980,14 +5981,30 @@ def submit(self, obj, script=None, extend=None, submit_dir=None):
...
+            for k, v in env.items():
+                if '()' in k:
+                    f_name = k.replace('()', '')
+                    runcmd += [k, v, "\n", "export", "-f", f_name]
+                else:
+                    if 'export' not in runcmd:
+                        runcmd += ['export']
+                    exp_str = "%s=\"%s\" " % (k, v)
+                    runcmd += [exp_str]
+            runcmd += "\n"
```
hirenvadalia: put `"\n"` in `[]`

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -5980,14 +5981,30 @@ def submit(self, obj, script=None, extend=None, submit_dir=None):
...
+        runcmd = []
+        exp_str = ""
+        if env:
+            runcmd += ['#!/bin/bash\n']
+            for k, v in env.items():
+                if '()' in k:
+                    f_name = k.replace('()', '')
+                    runcmd += [k, v, "\n", "export", "-f", f_name]
+                else:
+                    if 'export' not in runcmd:
```
hirenvadalia: can you explain here why you are checking 'export' in runcmd instead k?

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -6069,10 +6085,18 @@ def submit(self, obj, script=None, extend=None, submit_dir=None):
                 runcmd = [
                     'PBS_CONF_FILE=' + self.client_pbs_conf_file] + runcmd
                 as_script = True
-
+            if env:
+                user = PbsUser.get_user(obj.username)
+                host = user.host
+                run_str = " "
```
hirenvadalia: remove this line and, change below line to `run_str = " ".join(runcmd)`

test/tests/functional/pbs_nonprint_characters.py
```diff
@@ -127,28 +127,30 @@ def create_and_submit_job(self, user=None, attribs=None, content=None,
             use_attribs = {}
         else:
             use_attribs = attribs
-        retjob = Job(username=user, attrs=use_attribs)
+        retjob = Job(TEST_USER, attrs=use_attribs)
```
hirenvadalia: why did you change `username=user` to `TEST_USER`?

test/fw/ptl/lib/pbs_testlib.py
RKORavi: I am passing {"Key":"value", "k2":"v2",...}, export command is added to runcmd. e.g: ['export', 'k1=v1']. So, i am checking here if export command is already present in runcmd.
Thanks for pointing this out i will add a comment here.

test/tests/functional/pbs_nonprint_characters.py
RKORavi: In some platforms job is submitted as root user, it causes problem while copying error & output files.
So I have changed it to TEST_USER

hirenvadalia: then remove user argument from func

RKORavi:
@bhroam thank you for review & sign off

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -5980,14 +5981,30 @@ def submit(self, obj, script=None, extend=None, submit_dir=None):
             os.chdir(submit_dir)
         c = None
         # 1- Submission using the command line tools
+        runcmd = []
+        exp_str = ""
+        if env:
+            runcmd += ['#!/bin/bash\n']
+            for k, v in env.items():
```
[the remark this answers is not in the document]
RKORavi: @hirenvadalia thankyou for catching this, i have missed this condition completely. I will make the the code changes to handle this condition.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -5990,11 +5990,10 @@ def submit(self, obj, script=None, extend=None, submit_dir=None,
                     f_name = k.replace('()', '')
                     runcmd += [k, v, "\n", "export", "-f", f_name]
                 else:
-                    if 'export' not in runcmd:
-                        runcmd += ['export']
-                    exp_str = "%s=\"%s\" " % (k, v)
+                    runcmd += ['export']
```
hirenvadalia: replace 3 lines in else part with: `runcmd += ['export %s="%s"' % (k, v)]`

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -5980,14 +5981,27 @@ def submit(self, obj, script=None, extend=None, submit_dir=None):
             os.chdir(submit_dir)
         c = None
         # 1- Submission using the command line tools
+        runcmd = []
+        exp_str = ""
```
hirenvadalia: remove this unused variable

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -5924,7 +5925,7 @@ def submit(self, obj, script=None, extend=None, submit_dir=None):
         _interactive_job = False
         as_script = False
         rc = None
-
+        run_str = ""
```
hirenvadalia: remove this line as it is not required

RKORavi:
Latest commit contains the following changes:
1. While running regression sometimes i had noticed that checking NON_PRINTVAR variable in qstat -f output is not right.
2. Reason is sometimes NON_PRINTVAR value are printed in two lines 
E.g: for NONPRINT_VAR=X\\,^[[1m\\,^[[31m\\,Y check qstat -f contents were split in two line. as: 
 NONPRINT_VAR=X\\,^[[1m\\,**\t\n**^[[31m\\,Y due to limit on length of character that can be displayed.
3. I have changed the code in PTL test suite where the check if performed in qstat -f output.
4. Remaining are addressed review comments.

test/tests/functional/pbs_nonprint_characters.py
```diff
@@ -157,8 +157,13 @@ def check_qstatout(self, chk_var, jid):
         """
         cmd = [self.qstat_cmd, '-f', jid]
         ret = self.du.run_cmd(self.server.hostname, cmd=cmd)
-        qstat_out = '\n'.join(ret['out'])
-        self.assertIn(chk_var, qstat_out)
+        k = chk_var.split('=')[0]
+        for elem in ret['out']:
+            if k in elem:
+                i = ret['out'].index(elem)
+                job_str = elem.strip('\t') + ret['out'][i+1].strip('\t')
```
bhroam: Add spaces around the +

mike0042:
Is this PR still being pursued @RKORavi? The previous comment is over two months old. If not, please close it.

hirenvadalia:
@RKORavi Can you please address @bhroam's comments and rebase your branch with the latest mainline (as there are many new changes in the latest code) and run a regression with your changes?

RKORavi:
regression results

RKORavi:
thanks for review and signoff @bhroam @hirenvadalia. Rebasing and squashing the commits now.

RKORavi:
final_reg_results

PR #1403, opened 2019-11-18
Update Makefile to remove libundolr_lib from common_libs

riyazhakki:
#### Describe Bug or Feature

UndoLR's LD flag appended to common_libs instead of pbs_sched_LDADD.

#### Describe Your Change

Updated  Makefile, to append UndoLR's LD flag to pbs_sched_LDADD.

### DIFF
--- a/src/scheduler/Makefile.am
+++ b/src/scheduler/Makefile.am
@@ -136,5 +136,5 @@ dist_sysconf_DATA = \
 	pbs_sched_config
 
 if UNDOLR_ENABLED
-common_libs += @libundolr_lib@
+pbs_sched_LDADD += @libundolr_lib@
 endif
--- a/src/scheduler/Makefile.am
+++ b/src/scheduler/Makefile.am
@@ -121,7 +121,7 @@ common_libs = \
 	@KRB5_LIBS@
 
 pbs_sched_CPPFLAGS = ${common_cppflags}
-pbs_sched_LDADD = ${common_libs}
+pbs_sched_LDADD = ${common_libs} @libundolr_lib@
 pbs_sched_SOURCES = pbs_sched.c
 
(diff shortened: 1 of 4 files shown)


src/scheduler/Makefile.am
```diff
@@ -136,5 +136,5 @@ dist_sysconf_DATA = \
 	pbs_sched_config
 
 if UNDOLR_ENABLED
-common_libs += @libundolr_lib@
+pbs_sched_LDADD += @libundolr_lib@
```
mike0042: You shouldn't need to check UNDOLR_ENABLED since libundolr_lib should be an empty string if it was not requested. In that case you can just add @libundolr_lib@ after ${common_libs} where pbs_sched_LDADD is defined. What you currently have will work, but you have to examine the whole file to see that it's only added for pbs_sched.

src/scheduler/Makefile.am
```diff
@@ -121,7 +121,7 @@ common_libs = \
 	@KRB5_LIBS@
 
 pbs_sched_CPPFLAGS = ${common_cppflags}
-pbs_sched_LDADD = ${common_libs}
+pbs_sched_LDADD = ${common_libs} @libundolr_lib@
```
hirenvadalia: I like this change, I suggest doing the same change for server, mom, and comm also.

riyazhakki: Yeah, I did changes testing it now, will push the commit.

riyazhakki: I have now pushed the commit

riyazhakki:
Thank you @mike0042 and @hirenvadalia for the review and approval.
This branch is now ready to merge, @PBSPro/pbspro-maintainers please take it forward.

PR #1405, opened 2019-11-19
PBS_ALL softlimit affect jobs that are not over limit

arungrover:
#### Describe Bug or Feature
PBS_ALL softlimit applied to a resource affects the jobs that are not even requesting that resource.

#### Describe Your Change
Earlier any job exceeding PBS_ALL softlimit used to set a preempt_bit on server/queue (depending on where the limit was set) and this resulted into all the jobs in that server/queue to be marked over PBS_ALL softlimit even when these jobs are not requesting the resource to which the limit was applied.
This change removes preempt_bit from server/queue and manages PBS_ALL limit just like it handles user/group/project limit. It now checks if the job in hand has requested the resource to which PBS_ALL limit is applied and then only copy over the preempt_bit from the relevant resource.

#### Link to Design Doc
NA

[test.txt](https://github.com/PBSPro/pbspro/files/3865268/test.txt)

### DIFF
--- a/src/scheduler/limits.c
+++ b/src/scheduler/limits.c
@@ -2039,20 +2039,16 @@ check_queue_max_run_soft(server_info *si, queue_info *qi, resource_resv *rr)
 	max_running = (int) lim_get(key, LI2RUNCTXSOFT(qi->liminfo));
 	free(key);
 
-	if ((max_running == SCHD_INFINITY) ||
-		(max_running > qi->sc.running)) {
-		qi->alljobcounts->soft_limit_preempt_bit = 0;
-		return (0);
+	/* at this point, we know a limit is set for PBS_ALL*/
+	used = find_counts_elm(qi->alljobcounts, PBS_ALL_ENTITY, NULL, &cnt, NULL);
+	if (max_running != SCHD_INFINITY && used > max_running) {
+		if (cnt != NULL)
+			cnt->soft_limit_preempt_bit = PREEMPT_TO_BIT(PREEMPT_OVER_QUEUE_LIMIT);
+		return (PREEMPT_TO_BIT(PREEMPT_OVER_QUEUE_LIMIT));
 	} else {
-		/* at this point, we know a limit is set  for PBS_ALL*/
-		used = find_counts_elm(qi->alljobcounts, PBS_ALL_ENTITY, NULL, &cnt, NULL);
-		if (used > max_running) {
-			if (cnt != NULL)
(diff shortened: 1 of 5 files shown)


src/scheduler/limits.c
```diff
@@ -2022,22 +2024,35 @@ check_queue_max_run_soft(server_info *si, queue_info *qi, resource_resv *rr)
...
 
 	if ((max_running == SCHD_INFINITY) ||
 		(max_running > qi->sc.running)) {
-		qi->soft_limit_preempt_bit = 0;
+		qi->alljobcounts->soft_limit_preempt_bit = 0;
 		return (0);
 	} else {
-		qi->soft_limit_preempt_bit = PREEMPT_TO_BIT(PREEMPT_OVER_QUEUE_LIMIT);
-		return (PREEMPT_TO_BIT(PREEMPT_OVER_QUEUE_LIMIT));
+		/* at this point, we know a limit is set  for PBS_ALL*/
```
bhroam: Is the above code necessary?  We will return if the queue doesn't have a soft limit on line 2034. (and in other places).  Aso, there is two spaces between set and for, and no space between PBS_ALL and the end comment. (here and other places)

arungrover:
@agrawalravi90 @bhroam I've addressed the review comment. Please have a look.
[test_out.txt](https://github.com/PBSPro/pbspro/files/3898247/test_out.txt)

src/scheduler/limits.c
```diff
@@ -2039,20 +2039,16 @@ check_queue_max_run_soft(server_info *si, queue_info *qi, resource_resv *rr)
...
-		used = find_counts_elm(qi->alljobcounts, PBS_ALL_ENTITY, NULL, &cnt, NULL);
-		if (used > max_running) {
-			if (cnt != NULL)
-				cnt->soft_limit_preempt_bit = PREEMPT_TO_BIT(PREEMPT_OVER_QUEUE_LIMIT);
-			return (PREEMPT_TO_BIT(PREEMPT_OVER_QUEUE_LIMIT));
-		} else {
-			return (0);
-		}
+		if (cnt != NULL)
+			cnt->soft_limit_preempt_bit = 0;
```
agrawalravi90: is this equivalent to setting 'qi->alljobcounts->soft_limit_preempt_bit = 0' ? if not then do we not need to set qi->alljobcounts->soft_limit_preempt_bit = 0 anymore?

arungrover: Yes it is, find_counts_elm will return the first count itself which will have PBS_ALL_ENTITY

PR #1407, opened 2019-11-20
fix travis showing passed despite build fail

sandisamp:
#### Describe Bug or Feature

The ci wrapper python script for do.sh was not returning the exit code. Since the test part is also in the same script '|| true' part of earlier code was suppressing errors.

#### Describe Your Change

* Handled error code in ci script
* In travis config passing scripts without '|| true' and taking care of sanitize conditions in the script itself as BUILD_MODE env variable.

[Failed test scenario](https://travis-ci.org/sandisamp/pbspro/builds/614379367?utm_medium=notification&utm_source=github_status)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ before_install:
   - '[ "${OS_TYPE}" == "opensuse/leap:15" ] && export DOCKER_EXTRA_ARG="-e LC_ALL=C.utf8" || true'
   - docker run -it -d -h testdev.pbspro.org --name testdev -v $(pwd):$(pwd) --privileged -w $(pwd) ${DOCKER_EXTRA_ARG} ${OS_TYPE} /bin/bash
   - docker ps -a
-  - export DOCKER_EXEC="docker exec -it ${DOCKER_EXTRA_ARG} --privileged testdev"
+  - export DOCKER_EXEC="docker exec -it ${DOCKER_EXTRA_ARG} -e BUILD_MODE="${BUILD_MODE}" --privileged testdev"
 install:
   - '${DOCKER_EXEC} .travis/do_sanitize_mode.sh'
   - '${DOCKER_EXEC} .travis/do.sh'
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -22,8 +22,8 @@ before_install:
   - '[ "${OS_TYPE}" == "opensuse/leap:15" ] && export DOCKER_EXTRA_ARG="-e LC_ALL=C.utf8" || true'
   - docker run -it -d -h testdev.pbspro.org --name testdev -v $(pwd):$(pwd) --privileged -w $(pwd) ${DOCKER_EXTRA_ARG} ${OS_TYPE} /bin/bash
   - docker ps -a
-  - export DOCKER_EXEC="docker exec -it ${DOCKER_EXTRA_ARG} --privileged testdev"
+  - export DOCKER_EXEC="docker exec -it ${DOCKER_EXTRA_ARG} -e BUILD_MODE="${BUILD_MODE}" --privileged testdev"
 install:
-  - '[ "x${BUILD_MODE}" == "xsanitize" ] && ${DOCKER_EXEC} .travis/do_sanitize_mode.sh || true'
-  - '[ "x${BUILD_MODE}" != "xsanitize" ] && ${DOCKER_EXEC} .travis/do.sh || true'
+  - '${DOCKER_EXEC} .travis/do_sanitize_mode.sh'
+  - '${DOCKER_EXEC} .travis/do.sh'
```
prakashcv13: from what I understand, we can still check for BUILD_MODE here and execute the appropriate script based on it's value.

sandisamp:
```
[ "x${BUILD_MODE}" == "xsanitize" ] && ${DOCKER_EXEC} .travis/do_sanitize_mode.sh

The command "[ "x${BUILD_MODE}" == "xsanitize" ] && ${DOCKER_EXEC} .travis/do_sanitize_mode.sh" failed and exited with 1 during .
```
This always gives an error as mentioned above unless we run it like the following:

`'[ "x${BUILD_MODE}" == "xsanitize" ] && ${DOCKER_EXEC} .travis/do_sanitize_mode.sh || true'`
The above solution was however exiting with 0 regardless of build failure. Hence I have gone with the current approach.

PR #1409, opened 2019-11-20
PTL tests are failing in setUp() when run on cpuset mom

Mahalty:
#### Describe Bug or Feature
PTL Tests are failing in setUp(), when run on cpuset mom

#### Describe Your Change

- When setUp() is called at the beginning of the test case, setUp() calls revert_mom(). revert_mom() deletes and recreates the node and expects the node state to be free and resources_available.ncpus to be >= 1. But on cpuset mom, the value of resources_available.ncpus will be zero on base node and hence test fails.

- We should update revert_mom() function to check that total available ncpus should be greater than 1 in the cluster.

- [Execution_logs_with_fix.txt](https://github.com/PBSPro/pbspro/files/3977910/Execution_logs_with_fix.txt)

- [Execution_logs_without_fix.txt](https://github.com/PBSPro/pbspro/files/3977911/Execution_logs_without_fix.txt)

### DIFF
--- a/test/fw/ptl/utils/pbs_testsuite.py
+++ b/test/fw/ptl/utils/pbs_testsuite.py
@@ -1410,7 +1410,7 @@ class PBSTestSuite(unittest.TestCase):
         if not mom.isUp():
             self.logger.error('mom ' + mom.shortname + ' is down after revert')
         self.server.manager(MGR_CMD_CREATE, NODE, None, mom.shortname)
-        a = {'state': 'free', 'resources_available.ncpus': (GE, 1)}
+        a = {'state': 'free'}
         self.server.expect(NODE, a, id=mom.shortname, interval=1)
         return mom
 


Mahalty:
@Bhroam: I wasn't clear before. I have updated the description of the changes. Please let me know if it makes sense

bhroam:
What happens on a multi-mom test where one of the nodes is free and the other is not?  In any case, I think if you just check for state=free, you are OK.  It is telling you the mom has reported in and everything is back to default.

Lastly, should this fail the test or skip it?  Neither is great, but just because PBS couldn't revert to defaults doesn't mean the test failed.

Mahalty:
@bhroam: Currently revert_mom() will check the expected attributes(state and available ncpus >=1) only for the mom specified, It will not check for other moms.

Also earlier the test was checking for the expected attributes via expect function which returns error if the attributes doesn't match and now it checks using the assert function which fails the test if the condition fails.

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -1410,8 +1410,12 @@ def revert_mom(self, mom, force=False):
         if not mom.isUp():
             self.logger.error('mom ' + mom.shortname + ' is down after revert')
         self.server.manager(MGR_CMD_CREATE, NODE, None, mom.shortname)
-        a = {'state': 'free', 'resources_available.ncpus': (GE, 1)}
+        a = {'state': 'free'}
         self.server.expect(NODE, a, id=mom.shortname, interval=1)
+        exp_attrib = {'resources_available.ncpus': (GE, 1)}
+        counter = self.server.counter(NODE, exp_attrib)
+        if not counter:
```
prakashcv13: counter() returns None or a dictionary, would it make more sense to check "if counter is not None".

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -1410,8 +1410,12 @@ def revert_mom(self, mom, force=False):
...
+            self.fail("Value of 'resources_available.ncpus' is less than 1")
```
prakashcv13: Personally, I feel, reporting error is better than failing the test case.

bhroam:
@Mahalty thanks for the explanation about the moms.  Why are we checking if we have GE 1 cpu vs just checking state=free?  I guess it doesn't really matter, but if state=free, it means the mom has contacted the server.  With state=free will come all the resources available.  In any case, it can't hurt.

test/fw/ptl/utils/pbs_testsuite.py
Mahalty: @prakashcv13: In case of failed condition, counter doesn't return None but a dictionary. Below is the snippet
Code-snippet:
        exp_attrib = {'resources_available.ncpus': (GE, 100)}
        counter = self.server.counter(NODE, exp_attrib)
        print("====>counter=", counter)

Log-snippet:
====>counter= {'resources_available.ncpus>=100': 0}

Mahalty:
@bhroam: State of the node will be irrespective of total ncpus and here we are ensuring that given mom has atleast 1 ncpus so that job can run on that node.

test/fw/ptl/utils/pbs_testsuite.py
prakashcv13: hmmm, I am looking at the definition of counter in pbs_testlib.py - and see that it does return None. What you have tested is not an error condition.
In the function _filter() (called from counter()), if self.status raises PbsStatusError, counter will return None. There are other similar conditions as well in that function.

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -1414,8 +1414,9 @@ def revert_mom(self, mom, force=False):
         self.server.expect(NODE, a, id=mom.shortname, interval=1)
         exp_attrib = {'resources_available.ncpus': (GE, 1)}
         counter = self.server.counter(NODE, exp_attrib)
-        if not counter:
-            self.fail("Value of 'resources_available.ncpus' is less than 1")
+        msg = "Value of 'resources_available.ncpus' is less than 1"
+        rc = counter is not None and counter['resources_available.ncpus>=1'] < 1
+        raise PtlExpectError(rc=rc, msg=msg)
```
bhroam: Two things.  First, if you are going to raise an expect error, why don't you just use expect() and let it handle it?  Second, there is no if here.  If counter['resources_available.ncpus>=1'] < 1, rc will be 0, but you will still raise the expect error.

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -1414,8 +1414,9 @@ def revert_mom(self, mom, force=False):
         self.server.expect(NODE, a, id=mom.shortname, interval=1)
         exp_attrib = {'resources_available.ncpus': (GE, 1)}
         counter = self.server.counter(NODE, exp_attrib)
```
bhroam: I know I asked this before, but I am still concerned about it.  In a multi-mom setup, won't this pass if a different mom is up?  How can self.server.counter() know which mom we are talking about?  It's part of the server object, not the mom.

bhroam:
Travis failed because one line is longer than 79 characters.

test/fw/ptl/utils/pbs_testsuite.py
Mahalty: I agree with you @bhroam so I have reverted my changes to my initial change of just checking the state of the node to be free.

Mahalty:
**Latest Execution-logs:**
[Latest_execution_logs_aft_fix.txt](https://github.com/PBSPro/pbspro/files/4035609/Latest_execution_logs_aft_fix.txt)

bhroam:
For some reason Travis is listed twice.  One of the instances isn't finishing.  I'm going to close and reopen this PR to see if it fixes it.

mike0042:
One of the TravisCI instances is more than five hours older than the other, but has still not completed. Looking at the individual builds of the older instance, they all appear to have passed though the summary screen still shows them in progress.

PR #1410, opened 2019-11-20
Making job_sort_formula a sched atttribute

agrawalravi90:
#### Describe Bug or Feature

Right now, job_sort_formula is a server attribute, which means that in a multi-sched scenario, all the schedulers use the same formula. This is an unnecessary limitation, the scheds are intended to be individually configured, possibly very differently from each other. So, making job_sort_formula a sched attribute will allow one to configure different sorting formulas for each sched.

#### Describe Your Change

Implemented the design as described in the design doc, please check the link below for details.

#### Link to Design Doc

Design: https://pbspro.atlassian.net/wiki/spaces/PD/pages/1393721405/Making+job+sort+formula+a+sched+attribute

[test_job_sort_formula_multisched.log](https://github.com/PBSPro/pbspro/files/3878020/test_job_sort_formula_multisched.log)


src/lib/Libattr/master_sched_attr_def.xml
```diff
@@ -211,7 +211,7 @@
 	<member_at_comp>comp_f</member_at_comp>
 	<member_at_free>free_null</member_at_free>
 	<member_at_action>NULL_FUNC</member_at_action>
-	<member_at_flags><both>NO_USER_SET</both></member_at_flags>
+	<member_at_flags><both>PRIV_READ | ATR_DFLAG_MGWR</both></member_at_flags>
```
bhroam: Why is this different than job_sort_formula (MGR_ONLY_SET)

src/scheduler/server_info.c
```diff
@@ -929,6 +931,15 @@ query_sched_obj(status *policy, struct batch_status *sched, server_info *sinfo)
 				conf.preempt_min_wt_used = 1;
 			else
 				conf.preempt_min_wt_used = 0;
+		} else if (!strcmp(attrp->name, ATTR_job_sort_formula)) {
+			if (conf.job_sort_formula != NULL)
+				free(conf.job_sort_formula);
+			conf.job_sort_formula = read_formula();
```
bhroam: If you are going to read this every cycle, leave it where it was (sinfo).  If you want to read it only when it changes, follow the code for command SCH_ATTRS_CONFIGURE and shove it in global space somewhere.

src/scheduler/server_info.c
```diff
@@ -3403,7 +3410,10 @@ read_formula(void)
 	FILE *fp;
 
 
-	sprintf(pathbuf, "%s/%s", pbs_conf.pbs_home_path, FORMULA_ATTR_PATH_SCHED);
+	if (dflt_sched)
+		sprintf(pathbuf, "%s/%s", pbs_conf.pbs_home_path, FORMULA_ATTR_PATH_SCHED);
+	else
+		sprintf(pathbuf, "%s/sched_priv_%s/%s", pbs_conf.pbs_home_path, sc_name, FORMULA_FILENAME);
```
bhroam: This is incorrect.  There is a global that holds the sched_priv name.  A scheduler's sched_priv can be called anything.  This is just the default.

src/server/req_manager.c
```diff
@@ -1747,6 +1747,16 @@ mgr_server_unset(struct batch_request *preq)
...
 				set_attr_svr(&(server.sv_attr[(int)SRV_ATR_scheduling]),
 					    &svr_attr_def[(int) SRV_ATR_scheduling], "TRUE");
+			else if (strcasecmp(plist->al_name, ATTR_job_sort_formula) == 0) {
+				pbs_sched *psched;
+
+				/* Ask all scheds to reconfigure */
+				for (psched = (pbs_sched *) GET_NEXT(svr_allscheds);
+								psched != NULL;
+								psched = (pbs_sched *) GET_NEXT(psched->sc_link)) {
+					set_scheduler_flag(SCH_CONFIGURE, psched);
```
bhroam: Why a SCH_CONFIGURE?  That just forces the scheduler to read sched_config.  There is nothing in there about the formula.  If you want to follow some of the other sched attrs, send a SCH_ATTRS_CONFIGURE.

test/tests/functional/pbs_multi_sched.py
```diff
@@ -122,6 +122,78 @@ def check_vnodes(self, j, vnodes, jid):
...
+    def test_job_sort_formula_multisched(self):
+        """
+        Test that job_sort_formula can be set for each sched
+        """
+        self.common_setup()
+
+        self.server.expect(SCHED, 'job_sort_formula', id='sc1', op=UNSET)
+        self.server.expect(SCHED, 'job_sort_formula', id='sc2', op=UNSET)
+        self.server.expect(SCHED, 'job_sort_formula', id='sc3', op=UNSET)
+        self.server.expect(SERVER, 'job_sort_formula', op=UNSET)
```
bhroam: No reason to do this.  The system is reverted back to the default state in setUp().

src/scheduler/data_types.h
```diff
@@ -1098,6 +1097,7 @@ struct config
 	char **res_to_check;			/* the resources schedule on */
 	resdef **resdef_to_check;		/* the res to schedule on in def form */
 	char **ignore_res;			/* resources - unset implies infinite */
+	char *job_sort_formula;			/* job sort formula for the sched */
```
bhroam: This is yet another way to hold data from the scheduler object.  Some are in globals by themselves.  Some are on sinfo.  This would be yet another place for them.  You should follow one of the other ways, or move the others to conf.

test/tests/functional/pbs_multi_sched.py
```diff
@@ -122,6 +122,78 @@ def check_vnodes(self, j, vnodes, jid):
...
+            pass
+
+        # Now, unset server's JSF and set sc1's JSF again
+        self.server.manager(MGR_CMD_UNSET, SERVER, 'job_sort_formula')
+        self.server.manager(MGR_CMD_SET, SCHED,
+                            {'job_sort_formula': '2*walltime'}, id='sc1')
+
+        self.server.cleanup_jobs()
+
+        # Now, submit 2 jobs with different walltimes to each sched again
```
bhroam: Consider removing the 'Now' from the start of each comment

src/lib/Libattr/master_sched_attr_def.xml
agrawalravi90: The reference guide says that job_sort_formula_threshold is not user readable while job_sort_formula is, that's why I had to set it differently

src/scheduler/server_info.c
agrawalravi90: In order to support backwards compatibility for server's job_sort_formula attribute, we need to read it every cycle. So I'll just leave it with sinfo, we can change this to follow the SCH_ATTRS_CONFIGURE path when we remove the server's JSF.

src/scheduler/server_info.c
agrawalravi90: That's what my understanding was as well, but I couldn't find it. Since pbs_sched.c also does this (https://github.com/PBSPro/pbspro/blob/a280b6b9573d626db3cf055bf4f81cf6ebbcd506/src/scheduler/pbs_sched.c#L1075), I thought this would be correct. Do you now where we keep the sched_priv global? if this is incorrect then we should fix pbs_sched.c as well.

src/server/req_manager.c
agrawalravi90: Actually, we don't need this at all since it's being read every sched cycle.

SCH_ATTRS_CONFIGURE would work if we were explicitly setting the sched object's job_sort_formula when the server's job_sort_formula was set. But we can't do that if we want to make server & sched's job_sort_formula attributes incompatible, right? e.g - if we set server's job_sort_formula to "1*walltime" and also set it on all sched objects inside validate_job_formula(), then later if we change the server's job_sort_formula to "2*walltime", it becomes difficult to tell if the previously set job_sort_formula value on sched objects was set via server's job_sort_formula, or if they were set explicitly on the sched objects. If they were set explicitly, then they would be incompatible, if they were set via the server's job_sort_formula, then it's okay to change them. There's no way to tell.

agrawalravi90:
@bhroam I've addressed your comments, please provide further feedback, thanks!

src/scheduler/server_info.c
bhroam: OK, after looking through the code, I have figured out what is going on (and it is wrong).  A scheduler comes up and assumes the default sched_priv and reads the sched_config.  On the first scheduling cycle, it will read the real sched_priv from the server and does a chdir() into it.  It will not reread its sched_config though.  This means it has either read the wrong sched_config, or none at all.  It isn't until the first HUP that the real sched_config will be read.

None of this fixes your problem.  Your problem might be simpler to fix though.  Can't you just read a relative path to the sched formula file since you will be chdir()'d into the sched_priv?

src/scheduler/data_types.h
```diff
@@ -368,6 +368,7 @@ struct server_info
 	/* policy structure for the server.  This is an easy storage location for
 	 * the policy struct.  The policy struct will be passed around separately
 	 */
+	char *job_formula;	/* formula used for sorting */
```
bhroam: Is the comment missing a tab?

src/scheduler/server_info.c
```diff
@@ -558,9 +558,9 @@ query_server_info(status *pol, struct batch_status *server)
 		} else if (!strcmp(attrp->name, ATTR_NodeGroupKey))
 			sinfo->node_group_key = break_comma_list(attrp->value);
 		else if (!strcmp(attrp->name, ATTR_job_sort_formula)) {	/* Deprecated */
-			if (conf.job_sort_formula != NULL)
-				free(conf.job_sort_formula);
-			conf.job_sort_formula = read_formula();
+			if (sinfo->job_formula != NULL)
+				free(sinfo->job_formula);
```
bhroam: You are querying the job sort formula for the first time.  It can't be set.

src/scheduler/server_info.c
agrawalravi90: That's true, we'd already be in the sched_priv by that point, I'll make that change thanks.

agrawalravi90:
@bhroam I've addressed your comments, please let me know if it looks okay now.

src/scheduler/data_types.h
```diff
@@ -365,10 +365,10 @@ struct server_info
 	np_cache **npc_arr;
 
 	resource_resv *qrun_job;	/* used if running a job via qrun request */
-	char *job_formula;		/* formula used for sorting */
 	/* policy structure for the server.  This is an easy storage location for
 	 * the policy struct.  The policy struct will be passed around separately
 	 */
+	char *job_formula;		/* formula used for sorting */
```
arungrover: please remove this change because now policy comment is out of place.

src/server/sched_func.c
```diff
@@ -43,23 +43,356 @@
...
+ *
+ * @param[in]	formula - the formula to write
+ * @param[in]	sched_priv_path - path to scheduler's sched_priv
+ *
+ * @return	int
+ * @return 	PBSE_NONE for Success
+ * @return	PBSE_ error code for Failure
+ */
+static int
+write_job_formula(char *formula, char *sched_priv_path)
```
arungrover: you rename this to write_job_sort_formula() instead?

src/server/sched_func.c
```diff
@@ -43,23 +43,356 @@
...
+
+		/* check if any sched's JSF is set to a different value */
+		for (psched = (pbs_sched *) GET_NEXT(svr_allscheds);
+				psched != NULL;
+				psched = (pbs_sched *) GET_NEXT(psched->sc_link)) {
+			attribute *jsf_attr = NULL;
+
+			jsf_attr = &(psched->sch_attr[SCHED_ATR_job_sort_formula]);
+			if (jsf_attr->at_flags & ATR_VFLAG_SET) {
+				if (strcmp(psched->sch_attr[SCHED_ATR_job_sort_formula].at_val.at_str, formula) != 0) {
```
arungrover: just use jsf_attr instead of sch_attr[...]

src/server/sched_func.c
```diff
@@ -43,23 +43,356 @@
...
+					incompatible = 1;
+					break;
```
arungrover: cant we just return from here?

src/include/pbs_share.h
```diff
@@ -59,8 +59,9 @@
 #define FORMULA_ACCRUE_TYPE 	"accrue_type"
 
 /* Well known file to store job sorting formula */
-#define FORMULA_ATTR_PATH "server_priv/sched_formula"
-#define FORMULA_ATTR_PATH_SCHED "sched_priv/sched_formula"
+#define FORMULA_FILENAME "sched_formula"
+#define FORMULA_ATTR_PATH "server_priv/" FORMULA_FILENAME
+#define FORMULA_ATTR_PATH_SCHED "sched_priv/" FORMULA_FILENAME
```
arungrover: I don't think this is needed anymore, now that we are using sched_priv attribute

src/scheduler/server_info.c
```diff
@@ -3410,14 +3410,17 @@ read_formula(void)
...
+		log_event(PBSEVENT_SYSTEM, PBS_EVENTCLASS_REQUEST, LOG_INFO, __func__,
+				"Error getting current working directory path");
+		return NULL;
+	}
 
-	if (dflt_sched)
-		sprintf(pathbuf, "%s/%s", pbs_conf.pbs_home_path, FORMULA_ATTR_PATH_SCHED);
-	else
-		sprintf(pathbuf, "%s/sched_priv_%s/%s", pbs_conf.pbs_home_path, sc_name, FORMULA_FILENAME);
+	snprintf(pathbuf, sizeof(pathbuf), "%s/%s", cwd, FORMULA_FILENAME);
```
bhroam: Why bother getting the cwd?  Why not just open FORMULA_FILENAME.  It will open it in the current directory.

agrawalravi90:
@bhroam  and @arungrover Thanks guys, I've addressed your comments, please let me know if it looks okay now.

PR #1411, opened 2019-11-21
Updating last running job id in cleanup_jobs

tpai26:
#### Describe Bug or Feature
In cleanup_jobs in PTL we wait for the last running job to get finished. Currently PTL expects all jobs to be running. This change will filter out only the running jobs and then wait for the last running job in the list to get finished. 

Logs:
[deletion_jobs.txt](https://github.com/PBSPro/pbspro/files/3877251/deletion_jobs.txt)
[deletion_queued_jobs.txt](https://github.com/PBSPro/pbspro/files/3877252/deletion_queued_jobs.txt)

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -8471,7 +8471,7 @@ class Server(PBSService):
             select_xt = 'x'
         jobs = self.status(JOB, extend=select_xt)
         job_ids = sorted(list(set([x['id'] for x in jobs])))
-        running_jobs = [j['id'] for j in jobs if j['job_state'] == 'R']
+        running_jobs = sorted([j['id'] for j in jobs if j['job_state'] == 'R'])
         host_pid_map = {}
         for job in jobs:
             exec_host = job.get('exec_host', None)
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -8471,9 +8471,9 @@ class Server(PBSService):
             select_xt = 'x'
         jobs = self.status(JOB, extend=select_xt)
         job_ids = sorted(list(set([x['id'] for x in jobs])))
-        running_jobs = self.filter(JOB, {'job_state': 'R'})
-        if running_jobs.values():
-            last_running_job = list(running_jobs.values())[0][-1]
+        running_jobs = [j['id'] for j in jobs if j['job_state'] == 'R']
+        if running_jobs:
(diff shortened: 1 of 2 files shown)


test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -8471,6 +8471,9 @@ def cleanup_jobs(self):
             select_xt = 'x'
         jobs = self.status(JOB, extend=select_xt)
         job_ids = sorted(list(set([x['id'] for x in jobs])))
+        running_jobs = self.filter(JOB, {'job_state': 'R'})
```
hirenvadalia: You don't need to call Server.filter here (which internally will run qstat again), you already have the full status of jobs in `jobs`, you can make use of that then filter out running jobs like:
```python
last_running_job = job_ids[-1]
running_jobs = sorted([x['id'] for x in jobs if x['job_status'] == 'R'])
if running_jobs:
    last_running_job = running_jobs[-1]
```

bhroam: Won't, sorted() do an alpha sort?  Don't we need to chop the IDs at the first '.' so sorted() does a numeric sort?  So:
```
running_jobs = sorted(x['id'].split('.')[0] for x in jobs if x['job_state'] == 'R')
```

agrawalravi90: we need to make it an int() as well right? I think we should do sorted(int(x['id'].split('.')[0]) ...

tpai26: I have not sorted the list of jobs, I do not think we will need it since the status of jobs will return us jobs in the sequence they are present on the server.
Do correct me if I am wrong

tpai26:
@hirenvadalia Thanks for the catching the issue.
I have added the new log file in the PR description section

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -8510,7 +8513,7 @@ def cleanup_jobs(self):
                     self.du.run_cmd(host, ['kill', '-9'] + chunk,
                                     runas=ROOT_USER, logerr=False)
             if running_job is True:
```
bhroam: I don't think you need running_job any more now that you have last_running_job.  You can remove the variable.

test/fw/ptl/lib/pbs_testlib.py
bhroam: Actually don't care about sorting persay.  We care about the running job we had on the qdel line.  This means we should sort both lists or sort neither.

PR #1415, opened 2019-11-23
a fork in a hook continues to run unexpected code

vchlum:
#### Describe Bug or Feature
Let's have a hook:
```
import pbs
import os
import time
import sys

e = pbs.event()

pid = os.fork()
if pid == 0:
    time.sleep(10)
    sys.exit()

e.reject("test")
```
Starting this hook:
```
pbs_python  --hook -i /root/test.in -o /tmp/testout -L /tmp -e 65535 -r /var/spool/pbs/mom_priv/hooks/resourcedef /root/test.py 
```

At first, the hook will write expected output:
```
~# cat /tmp/testout 
pbs.event().reject=True
pbs.event().accept=False
pbs.event().reject_msg=test
```
but after 10 seconds, the output will be rewritten to 
```
~# cat /tmp/testout 
pbs.event().accept=True
pbs.event().reject=False
```
If we remove the line 'time.sleep(10)' from the hook file then random data are read by mom or by the server. This is just an example of how the pbs_python behaves. We can also lose changes in attributes and it can be a source of unexpected behavior.

As a workaround, you can exit the fork with os._exit(0).

#### Describe Your Change
After running the hook code, check for the pid and if the pid changed then it is a fork and the fork should be terminated immediately.

[ptl_smoke_hook.txt](https://github.com/PBSPro/pbspro/files/3882272/ptl_smoke_hook.txt)

### DIFF
--- a/src/lib/Libpython/pbs_python_external.c
+++ b/src/lib/Libpython/pbs_python_external.c
@@ -693,6 +693,7 @@ pbs_python_run_code_in_namespace(struct python_interpreter_data *interp_data,
 	PyObject *retval;
 	char      *pStr;
 	int rc=0;
+	pid_t orig_pid;
 
 	if (!interp_data || !py_script) {
 		log_err(-1, __func__, "Either interp_data or py_script is NULL");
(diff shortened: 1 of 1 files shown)


vchlum:
It was my intention to provide such a hook. My use case is to start an asynchronous task (fork) in the hook, which can prepare data for further hook execution.

Thank you for review and merging.

PR #1416, opened 2019-11-25
test_prologue_hook_does_not_execute_twice_on_pbsdsh fails due to out of range error

sandisamp:
#### Describe Bug or Feature
The test failed with out of range error.

#### Describe Your Change

Sometimes the machines returned warning messages along with the mom names. An example is given below
`{'out': ['Warning: no access to tty (Bad file descriptor).', 'Thus no job control in this shell.', 'x52-rhel7.pbspro.com', 'x58-r7.pbspro.com'], 'err': [], 'rc': 0}`

> This is an innocuous warning that simply means you are running a script (rather than a binary) under a job that has no access to the TTY. In otherwords, you cannot interrupt it (^C), suspend it (^Z) or use other interactive commands because there is no screen or keyboard to interact with it. It can be safely ignored.

Fix now always takes the last  2 elements from the dictionary key ['out']

[test_prologue_hook_does_not_execute_twice_on_pbsdsh](https://github.com/PBSPro/pbspro/files/3885068/prologue_log.txt)

### DIFF
--- a/test/tests/functional/pbs_hook_execjob_prologue.py
+++ b/test/tests/functional/pbs_hook_execjob_prologue.py
@@ -261,12 +261,9 @@ class TestPbsExecutePrologue(TestFunctional):
         ret = self.du.cat(hostname=host, filename=opath, runas=TEST_USER)
         _msg = "cat command failed with error: %s" % ret['err']
         self.assertEqual(ret['rc'], 0, _msg)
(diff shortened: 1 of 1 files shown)


sandisamp:
[Latest_logs](https://github.com/PBSPro/pbspro/files/3885599/prologue_log.txt)

PR #1417, opened 2019-11-25
add Kerberos build into Travis CI

vchlum:
#### Describe Bug or Feature
New PBS Pro build for Travis CI with enabled Kerberos. The build runs on epel7 and uses MIT Kerberos. This is the first phase of adding automated tests for testing Kerberos.

Discussion [here](http://community.pbspro.org/t/kerberos-support/48/80).

#### Describe Your Change
Adding a new build with enabled Kerberos.

See travis log.


ci/do.sh
```diff
@@ -109,12 +114,19 @@ if [ "x${ONLY_REBUILD}" != "x1" -a "x${ONLY_INSTALL}" != "x1" -a "x${ONLY_TEST}"
     else 
       configure_opt='--prefix=/opt/pbs --enable-ptl'
     fi
+    if [ "x${BUILD_MODE}" == "xkerberos" ]; then
+      configure_opt="${configure_opt} --with-krbauth PATH_KRB5_CONFIG=/usr/bin/krb5-config"
+    fi
```
sandisamp: This condition is not useful here as BUILD_MODE variable is not passed in CI.

vchlum: Thank you, The condition is removed. It is not needed.

PR #1420, opened 2019-11-27
Sched segfaults accessing free'd node partition cache

agrawalravi90:
#### Describe Bug or Feature

Scheduler's node structure was recently enhanced to have pointers to node partitions. This was done for performance reasons. This cache is created when the universe is first queried, and is not expected to be updated through the course of a sched cycle. But, we were updating it in a corner case during calendaring if the simulation code ends up creating host sets, which can happen in certain situations, one of which is if calendaring a suspended job on a multi-vnoded host, this leads to the creation of host sets, which ends up creating placements sets and attaching a new copy of the cache on the duplicated nodes, which gets free'd, and scheduler seg-faults when the free'd cache is access for the next event in the calendar.

#### Describe Your Change

Added a new flag to tell create_node_partitions to not update the node partition cache.

I actually wrote a specialized PTL test for this, but it didn't seem functionally valuable to me. It's only interesting to trigger this segfault, but not otherwise. So, I'm not adding it to the PR. Please let me know if you think that I should add it.
The test i did involved creating 2 vnodes (8 ncpus each) on the same host, an expressq (Priority 200), turning preemption on, submitting 16 low priority jobs requesting 1ncpu, then submitting a high priority job which requests 2:ncpus=8. Then triggered the seg fault.

Valgrind logs below (scroll to the bottom of the files, skip all the Python related errors):
[valgrind_after.log](https://github.com/PBSPro/pbspro/files/3898653/valgrind_after.log)
[valgrind_before.log](https://github.com/PBSPro/pbspro/files/3898654/valgrind_before.log)

### DIFF
--- a/src/scheduler/constant.h
+++ b/src/scheduler/constant.h
@@ -523,7 +523,7 @@ enum nodepart
 	NP_IGNORE_EXCL = 1,
 	NP_CREATE_REST = 2,
 	NP_NO_ADD_NP_ARR = 4
-	/* next 4, 8, etc */
+	/* next 8, 16, etc */
 };
 
 /* It is used to identify the provisioning policy set on scheduler */
--- a/src/scheduler/constant.h
+++ b/src/scheduler/constant.h
@@ -522,7 +522,7 @@ enum nodepart
 	NP_LOW = 0,
 	NP_IGNORE_EXCL = 1,
 	NP_CREATE_REST = 2,
-	NP_NO_CACHE_UPDATE = 4
+	NP_NO_ADD_NP_ARR = 4
 	/* next 4, 8, etc */
 };
 
--- a/src/scheduler/constant.h
+++ b/src/scheduler/constant.h
@@ -521,7 +521,8 @@ enum nodepart
 {
 	NP_LOW = 0,
 	NP_IGNORE_EXCL = 1,
-	NP_CREATE_REST = 2
+	NP_CREATE_REST = 2,
+	NP_NO_CACHE_UPDATE = 4
 	/* next 4, 8, etc */
 };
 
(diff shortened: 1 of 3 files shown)


src/scheduler/constant.h
```diff
@@ -521,7 +521,8 @@ enum nodepart
 {
 	NP_LOW = 0,
 	NP_IGNORE_EXCL = 1,
-	NP_CREATE_REST = 2
+	NP_CREATE_REST = 2,
+	NP_NO_CACHE_UPDATE = 4
```
bhroam: This isn't a cache.  It's an array of node partitions it is in.  The cache is when we create node partitions on the fly . It is stored elsewhere.  This should be something like NP_NO_ADD_NP_NODE_ARR or something like that.

src/scheduler/node_partition.c
```diff
@@ -897,10 +899,12 @@ find_alloc_np_cache(status *policy, np_cache ***pnpc_arr,
 	npc = find_np_cache(*pnpc_arr, resnames, ninfo_arr);
 
 	if (npc == NULL) {
+		int flags;
+
+		flags = policy->only_explicit_psets ? NP_NO_CACHE_UPDATE : NP_CREATE_REST | NP_NO_CACHE_UPDATE;
```
bhroam: Say int flags = NP_NO_CACHE_UPDATE at declaration and then do a flags |= here.

agrawalravi90:
@bhroam I've addressed your comments, please let me know if it looks okay now. Thanks!

agrawalravi90:
@arungrover Thanks, I've addressed your comment, please let me know if it looks okay now.

PR #1421, opened 2019-11-27
Removed _SX related code

SudeshnaMoulik:
#### Describe Bug or Feature
There are few codes in our code base related to _SX, which we dont define anywhere. This PR is for removal of those codes.

#### Describe Your Change
Removal of _SX related code

### DIFF
--- a/src/tools/pbs_probe.c
+++ b/src/tools/pbs_probe.c
@@ -2802,14 +2802,14 @@ check_paths(struct infrastruct *pinf)
 	MPUG	*pmpug;
 	char	*realpath;
 
-#ifndef	_SX
-        for (i=0; i<PBS_last; ++i) {
+
+    for (i=0; i<PBS_last; ++i) {
 		msg_table_set_defaults(pinf, SRC_pri, MSG_po);
-                if ((realpath = pinf->pri.pbs_mpug[i].realpath))
+        if ((realpath = pinf->pri.pbs_mpug[i].realpath))
 			check_owner_modes(realpath, &pinf->pri.pbs_mpug[i], 0);
 	}
-#endif
-        for (i=0; i<PH_last; ++i) {    
+
+    for (i=0; i<PH_last; ++i) {    
 		msg_table_set_defaults(pinf, SRC_home, MSG_po);
 
 		if ((pmpug = pinf->home[i]) == NULL)
--- a/src/tools/pbs_probe.c
+++ b/src/tools/pbs_probe.c
@@ -2803,13 +2803,13 @@ check_paths(struct infrastruct *pinf)
 	char	*realpath;
 
 
-    for (i=0; i<PBS_last; ++i) {
+	for (i=0; i<PBS_last; ++i) {
 		msg_table_set_defaults(pinf, SRC_pri, MSG_po);
-        if ((realpath = pinf->pri.pbs_mpug[i].realpath))
+		if ((realpath = pinf->pri.pbs_mpug[i].realpath))
 			check_owner_modes(realpath, &pinf->pri.pbs_mpug[i], 0);
 	}
 
-    for (i=0; i<PH_last; ++i) {    
+	for (i=0; i<PH_last; ++i) {    
 		msg_table_set_defaults(pinf, SRC_home, MSG_po);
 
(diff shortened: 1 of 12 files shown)


agrawalravi90:
We also don't define NAS defined anywhere but we have a bunch of #ifdef NAS throughout the code, do we plan to remove that as well?

mike0042:
> We also don't define NAS defined anywhere but we have a bunch of #ifdef NAS throughout the code, do we plan to remove that as well?

@bhroam would know best, but I highly doubt the NAS defines are going away anytime soon. They only appeared in the past few years.

src/tools/pbs_upgrade_job.c
```diff
@@ -154,14 +154,8 @@ typedef struct taskfix_PRE19
...
 	union {
 		int ti_hold[16]; 	/* reserved space */
-		struct
-		{
-#ifdef _SX
-			pid_t ti_parent;
-			id_t ti_jid;
-#else
+		struct {
 			int ti_jj;	/* has to have a member */
```
bhroam: Now that there is only one element in the struct, consider removing the struct and promoting its member to the parent union.

bhroam:
Yes, the NAS defines need to stay.  NAS has their own mods to our code.  Having some of them in our code makes it easier for them to bring their codeline up to current with ours.

src/tools/pbs_upgrade_job.c
bhroam: I actually think you can remove the structure altogether.  We had a structure with 2 elements on one side of the ifdef.  It would need at least one on the other side of the else for the code to compile.  According to my IDE, ti_jj is never used.  As a note, there are two copies of this structure (the other in job.h).

SudeshnaMoulik:
Thank you @bhroam for your review. Yes you are correct that struct element is never used anywhere. So removed the same from both job.h and pbs_upgrade_job.c.

src/tools/pbs_probe.c
```diff
@@ -2802,14 +2802,14 @@ check_paths(struct infrastruct *pinf)
...
 	char	*realpath;
 
-#ifndef	_SX
-        for (i=0; i<PBS_last; ++i) {
+
+    for (i=0; i<PBS_last; ++i) {
 		msg_table_set_defaults(pinf, SRC_pri, MSG_po);
-                if ((realpath = pinf->pri.pbs_mpug[i].realpath))
+        if ((realpath = pinf->pri.pbs_mpug[i].realpath))
 			check_owner_modes(realpath, &pinf->pri.pbs_mpug[i], 0);
```
agrawalravi90: Same here, is this over-indented?

src/tools/pbs_probe.c
```diff
@@ -2802,14 +2802,14 @@ check_paths(struct infrastruct *pinf)
 	MPUG	*pmpug;
 	char	*realpath;
 
-#ifndef	_SX
-        for (i=0; i<PBS_last; ++i) {
+
+    for (i=0; i<PBS_last; ++i) {
```
agrawalravi90: is the line below this over-indented now?

src/tools/pbs_probe.c
```diff
@@ -2802,14 +2802,14 @@ check_paths(struct infrastruct *pinf)
...
 		msg_table_set_defaults(pinf, SRC_pri, MSG_po);
-                if ((realpath = pinf->pri.pbs_mpug[i].realpath))
+        if ((realpath = pinf->pri.pbs_mpug[i].realpath))
 			check_owner_modes(realpath, &pinf->pri.pbs_mpug[i], 0);
 	}
-#endif
-        for (i=0; i<PH_last; ++i) {    
+
+    for (i=0; i<PH_last; ++i) {
```
agrawalravi90: Same comment for the line below this

SudeshnaMoulik:
Thanks @agrawalravi90 for review this PR. Corrected the indentation. 

This happened because in my IDE the oss source code (https://github.com/PBSPro/pbspro/blob/688e905d75ce87431a2bf646bcb6c8afc26ce34e/src/tools/pbs_probe.c#L2808)  was lacking of indentation.

PR #1423, opened 2019-11-28
test_direct_write_when_job_succeeds_controlled fails because it uses mkdtemp

birbalsain:
#### Describe Bug or Feature

test_direct_write_when_job_succeeds_controlled fails because it uses mkdtemp.

#### Describe Your Change

Use du.create_temp_dir to  creat temp directory.

[qsub_temp_dir.txt](https://github.com/PBSPro/pbspro/files/3899875/qsub_temp_dir.txt)
[qsub_direct_write.txt](https://github.com/PBSPro/pbspro/files/3899870/qsub_direct_write.txt)

### DIFF
--- a/test/tests/functional/pbs_qsub_direct_write.py
+++ b/test/tests/functional/pbs_qsub_direct_write.py
@@ -87,9 +87,9 @@ class TestQsub_direct_write(TestFunctional):
         """
         j = Job(TEST_USER4, attrs={ATTR_k: 'doe'})
         j.set_sleep_time(10)
-        sub_dir = self.du.mkdtemp(uid=TEST_USER4.uid)
-        mapping_dir = self.du.mkdtemp(
-            uid=TEST_USER5.uid, gid=TSTGRP4.gid, mode=0o770)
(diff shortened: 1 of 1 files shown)


birbalsain:
qsub_temp_dir.txt has linux logs
qsub_direct_write.txt has shasta logs

birbalsain:
All test failed on shasta because of one other bug of remote directory.

birbalsain:
If i create mapping directory by job submitting user and that user's other than primary group name.
and create sub_dir with any user(different and same as job user) then test will pass.
But if i create sub_dir with job submitting user. mapping directory by a different user,
owned by a group that is not the job user's primary gid, (but is a gid that the user is a member of),Then i got error, expect job_stat run but got held.

birbalsain:
In tracejob i got error is unable to open std file.

birbalsain:
Is it product issue?

mike0042:
@birbalsain - Your changes to the test look correct and have been merged. If you believe there may be a product issue, please file a ticket and it will be investigated.

PR #1425, opened 2019-11-29
qstat doesn't supply full machine name with -n when nodes added by IP address

npadole20:
#### Describe Bug or Feature
* When nodes are created using IP address, qstat -n does not show the full node name. Node name is truncated after the first dot.
* When we create nodes using FQDN, qstat -n does not display a valid exec_host string in the expected format of <host1>/<T1>*<P1>.

#### Describe Your Change

* Changed how "." is handled in a node name. If the node name is a valid IP address, the string is not truncated at the first ".".
* Concatenating the <T1>*<P1> information to the short name of the node, if nodes are created using FQDN.

#### Link to Design Doc

NA

[qstat_n.txt](https://github.com/PBSPro/pbspro/files/3904359/qstat_n.txt)
[qstat_n_win.txt](https://github.com/PBSPro/pbspro/files/3911762/qstat_n_win.txt)
[valgrind.log](https://github.com/PBSPro/pbspro/files/3926570/valgrind.log)

### DIFF
--- a/src/cmds/qstat.c
+++ b/src/cmds/qstat.c
@@ -602,7 +602,6 @@ prt_nodes(char *nodes, int no_newl)
 	int  i, l;
 	char linebuf[78];
 	char *rest;
-	char *saveptr;
 	char *token;
 	char *token_cp;
 	char *subtoken;
--- a/src/cmds/qstat.c
+++ b/src/cmds/qstat.c
@@ -601,13 +601,13 @@ prt_nodes(char *nodes, int no_newl)
 {
 	int  i, len;
 	char linebuf[78];
-	char *rest;
-	char *saveptr;
-	char *token;
-	char *token_cp;
-	char *subtoken;
+	char *rest = NULL;
+	char *saveptr = NULL;
+ 	char *token = NULL;
+	char *token_cp = NULL;
+	char *subtoken = NULL;
 	char *node_name = NULL;
-	char *chunk;
+	char *chunk = NULL;
 	struct sockaddr_in check_ip;
 	int ret = 0;
 
@@ -668,10 +668,14 @@ prt_nodes(char *nodes, int no_newl)
 		printf((no_newl ? "%s\n" : "   %s\n"), show_nonprint_chars(linebuf));
 	} else if (no_newl)
 		printf("\n");
-	if (token_cp)
+	if (token_cp) {
 		free(token_cp);
-	if (rest)
+		token_cp = NULL;
+	}
+	if (rest) {
 		free(rest);
+		rest = NULL;
+	}
 }
 
 /**
--- a/src/cmds/qstat.c
+++ b/src/cmds/qstat.c
@@ -61,7 +61,7 @@
 #include "pbs_json.h"
 #include "pbs_internal.h"
 #include "libutil.h"
-#include	<arpa/inet.h>
+#include <arpa/inet.h>
 
 #if	TCL_QSTAT
 #include	<sys/stat.h>
@@ -599,75 +599,79 @@ trunc_value(char *value, int len, int wide_len, int wide)
 static void
 prt_nodes(char *nodes, int no_newl)
 {
-	int  i, l;
+	int  i, len;
 	char linebuf[78];
 	char *rest;
+	char *saveptr;
 	char *token;
 	char *token_cp;
 	char *subtoken;
-	char *node_name;
+	char *node_name = NULL;
 	char *chunk;
-	struct	sockaddr_in check_ip;
+	struct sockaddr_in check_ip;
 	int ret = 0;
 
 	if ((nodes == NULL) || (*nodes == '\0'))
 		return;
 
 	i = 0;
-	rest = malloc(strlen(nodes) + 1);
-	if(rest)
-		strcpy(rest, nodes);
-	else
+	rest = strdup(nodes);
+	if (rest == NULL)
 		exit_qstat("out of memory");
-	while ((token = strtok_r(rest, "+", &rest))){
-		token_cp = (char*)malloc(strlen(token) + 1);
-		if(token_cp)
-			strcpy(token_cp, token);
-		else
+	/* The exec_host string has the format <host1>/<T1>*<P1>[+<host2>/<T2>*<P2>+... ].
+	 * We are using '+' delimiter to find each <host1>/<T1>*<P1> string.
+	 */
+	token = strtok_r(rest, "+", &saveptr);
+	while (token != NULL) {
+		token_cp = strdup(token);
+		if (token_cp == NULL)
 			exit_qstat("out of memory");
+		/* We are using '/' delimiter to extract the <host1> value
+		 * from <host1>/<T1>*<P1> string. We use the <host1> to identify if
+		 * the node is created using IP address.
+		 */
 		subtoken = strtok(token, "/");
 		chunk = token_cp + strlen(subtoken);
 		ret = inet_pton(AF_INET, subtoken, &(check_ip.sin_addr));
-		if (ret == 1){
+		if (ret == 1) {
 			/* node name is an IP address */
-			node_name = subtoken;
-			strcat(node_name, chunk);
-
+			pbs_asprintf(&node_name, "%s%s", subtoken, chunk);
 		} else {
 			/* Node name is not an IP address */
-			node_name = strtok(subtoken, ".");
-			strcat(node_name, chunk);
+			pbs_asprintf(&node_name, "%s%s", strtok(subtoken, "."), chunk);
 		}
-		l = strlen(node_name);
-		if (i + l < 77){
-			while (l) {
+		len = strlen(node_name);
+		if (i + len < 77) {
+			while (len) {
 				linebuf[i++] = *node_name++;
-				l--;
+				len--;
 			}
+			/* Appedning a  '+' here because we want to maintain the
+			 * exec_host format i.e. <host1>/<T1>*<P1>[+<host2>/<T2>*<P2>+.
+			 */
 			linebuf[i++] = '+';
 		} else {
 			/* flush line and start next */
 			linebuf[i] = '\0';
-			if (no_newl)
-				printf("%s", show_nonprint_chars(linebuf));
-			else
-				printf("   %s\n", show_nonprint_chars(linebuf));
+			printf((no_newl ? "%s" : "   %s\n"), show_nonprint_chars(linebuf));
 			i = 0;
-			while (l) {
+			while (len) {
 				linebuf[i++] = *node_name++;
-				l--;
+				len--;
 			}
 			linebuf[i++] = '+';
 		}
+		token = strtok_r(NULL, "+", &saveptr);
 	}
-	if (i != 0) {
+	if (i > 0) {
 		linebuf[--i] = '\0';
-		if (no_newl)
-			printf("%s\n", show_nonprint_chars(linebuf));
-		else
-			printf("   %s\n", show_nonprint_chars(linebuf));
+		printf((no_newl ? "%s\n" : "   %s\n"), show_nonprint_chars(linebuf));
 	} else if (no_newl)
 		printf("\n");
+	if (token_cp)
+		free(token_cp);
+	if (rest)
+		free(rest);
 }
 
 /**
--- a/src/cmds/qstat.c
+++ b/src/cmds/qstat.c
@@ -61,6 +61,7 @@
 #include "pbs_json.h"
 #include "pbs_internal.h"
 #include "libutil.h"
+#include	<arpa/inet.h>
 
 #if	TCL_QSTAT
 #include	<sys/stat.h>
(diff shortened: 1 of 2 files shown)


npadole20:
[qstat_seg_fix.txt](https://github.com/PBSPro/pbspro/files/3911779/qstat_seg_fix.txt)

npadole20:
PTL logs -
[qstat_fqdn_af.txt](https://github.com/PBSPro/pbspro/files/3916269/qstat_fqdn_af.txt)
[qstat_fqdn_bf.txt](https://github.com/PBSPro/pbspro/files/3916270/qstat_fqdn_bf.txt)
[qstat_ip_af.txt](https://github.com/PBSPro/pbspro/files/3916271/qstat_ip_af.txt)
[qstat_ip_bf.txt](https://github.com/PBSPro/pbspro/files/3916272/qstat_ip_bf.txt)

src/cmds/qstat.c
```diff
@@ -598,49 +599,69 @@ trunc_value(char *value, int len, int wide_len, int wide)
...
+		strcpy(rest, nodes);
+	else
+		exit_qstat("out of memory");
+	while ((token = strtok_r(rest, "+", &rest))){
+		token_cp = (char*)malloc(strlen(token) + 1);
+		if(token_cp)
+			strcpy(token_cp, token);
+		else
+			exit_qstat("out of memory");
+		subtoken = strtok(token, "/");
```
mike0042: Please add a comment. Why are we searching for "/"? Is it meant to be a netmask?

src/cmds/qstat.c
```diff
@@ -598,49 +599,69 @@ trunc_value(char *value, int len, int wide_len, int wide)
...
-					linebuf[i++] = *nodes++;
-			}
+	rest = malloc(strlen(nodes) + 1);
+	if(rest)
+		strcpy(rest, nodes);
+	else
+		exit_qstat("out of memory");
+	while ((token = strtok_r(rest, "+", &rest))){
+		token_cp = (char*)malloc(strlen(token) + 1);
+		if(token_cp)
```
mike0042: Add a space after if.

src/cmds/qstat.c
```diff
@@ -598,49 +599,69 @@ trunc_value(char *value, int len, int wide_len, int wide)
...
+			strcpy(token_cp, token);
+		else
+			exit_qstat("out of memory");
+		subtoken = strtok(token, "/");
+		chunk = token_cp + strlen(subtoken);
+		ret = inet_pton(AF_INET, subtoken, &(check_ip.sin_addr));
+		if (ret == 1){
+			/* node name is an IP address */
+			node_name = subtoken;
+			strcat(node_name, chunk);
```
mike0042: Are you certain this string will be null terminated and that we'll never overrun the buffer?

src/cmds/qstat.c
```diff
@@ -598,49 +599,69 @@ trunc_value(char *value, int len, int wide_len, int wide)
...
-				stp++;
-			nodes = stp++;
 		} else {
-			stp++;
+			/* Node name is not an IP address */
+			node_name = strtok(subtoken, ".");
+			strcat(node_name, chunk);
+		}
+		l = strlen(node_name);
+		if (i + l < 77){
```
mike0042: Add a space before the brace.

What is happening here? Why 77? Please add a comment.

src/cmds/qstat.c
```diff
@@ -598,49 +599,69 @@ trunc_value(char *value, int len, int wide_len, int wide)
...
+			while (l) {
+				linebuf[i++] = *node_name++;
+				l--;
+			}
+			linebuf[i++] = '+';
```
mike0042: What is the significance of "+", and why are we not dealing with "/" as we did previously?

src/cmds/qstat.c
```diff
@@ -598,49 +599,69 @@ trunc_value(char *value, int len, int wide_len, int wide)
...
+		} else {
+			/* flush line and start next */
+			linebuf[i] = '\0';
+			if (no_newl)
+				printf("%s", show_nonprint_chars(linebuf));
+			else
+				printf("   %s\n", show_nonprint_chars(linebuf));
```
mike0042: I think you can utilize a tertiary operator here.
```suggestion
				printf((no_newl ? "%s" : "   %s\n"), show_nonprint_chars(linebuf));
```

src/cmds/qstat.c
```diff
@@ -598,49 +599,69 @@ trunc_value(char *value, int len, int wide_len, int wide)
...
+			while (l) {
+				linebuf[i++] = *node_name++;
+				l--;
+			}
+			linebuf[i++] = '+';
 		}
 	}
 	if (i != 0) {
-		linebuf[i] = '\0';
+		linebuf[--i] = '\0';
```
mike0042: This looks dangerous. The conditional should be (i > 0). Otherwise, you run the risk of data corruption.

test/tests/functional/pbs_qstat.py
```diff
@@ -102,3 +102,42 @@ def test_qstat_qselect(self):
...
+        self.server.manager(MGR_CMD_DELETE, NODE, None, '')
+        ipaddr = socket.gethostbyname(self.mom.hostname)
+        attr_A = {'Mom': self.mom.hostname}
+        self.server.manager(MGR_CMD_CREATE, NODE, id=ipaddr, attrib=attr_A)
+        self.server.expect(NODE, {'state=free': 1})
+        j = Job(TEST_USER)
+        jid = self.server.submit(j)
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid)
+        qstat_cmd = \
+            os.path.join(self.server.pbs_conf['PBS_EXEC'], 'bin', 'qstat')
```
mike0042: Please avoid the use of backslash for line continuation where possible in Python. Align the last two arguments one space to the right of the open parenthesis on the next line and pycodestyle will be happy.

test/tests/functional/pbs_qstat.py
```diff
@@ -102,3 +102,42 @@ def test_qstat_qselect(self):
...
+
+    def test_qstat_n_fqdn(self):
+        self.server.manager(MGR_CMD_DELETE, NODE, None, '')
+        self.server.manager(MGR_CMD_CREATE, NODE, id=self.mom.hostname)
+        self.server.expect(NODE, {'state=free': 1})
+        j = Job(TEST_USER)
+        jid = self.server.submit(j)
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid)
+        qstat_cmd = \
+            os.path.join(self.server.pbs_conf['PBS_EXEC'], 'bin', 'qstat')
```
mike0042: Please avoid the use of backslash for line continuation where possible in Python. Align the last two arguments one space to the right of the open parenthesis on the next line and pycodestyle will be happy.

test/tests/functional/pbs_qstat.py
```diff
@@ -102,3 +102,42 @@ def test_qstat_qselect(self):
                               as_script=True)
         if ret['rc'] != 0:
             self.assertFalse(ret['err'][0], ret_msg)
+
+    def test_qstat_n_ip(self):
```
sujatapatnaik52: Can you add docstring for both the test cases?

src/cmds/qstat.c
```diff
@@ -598,55 +599,83 @@ trunc_value(char *value, int len, int wide_len, int wide)
...
+		/* We are using '/' delimiter to extract the <host1> value
+		 * from <host1>/<T1>*<P1> string. We use the <host1> to identify if
+		 * the node is created using IP address.
+		 */
+		subtoken = strtok(token, "/");
+		chunk = token_cp + strlen(subtoken);
+		ret = inet_pton(AF_INET, subtoken, &(check_ip.sin_addr));
+		if (ret == 1) {
+			/* node name is an IP address */
+			pbs_asprintf(&node_name, "%s%s", subtoken, chunk);
```
sujatapatnaik52: you may need to free `node_name` after it's usage.

src/cmds/qstat.c
```diff
@@ -598,55 +599,83 @@ trunc_value(char *value, int len, int wide_len, int wide)
...
+		} else {
+			/* Node name is not an IP address */
+			pbs_asprintf(&node_name, "%s%s", strtok(subtoken, "."), chunk);
+		}
+		len = strlen(node_name);
+		if (i + len < 77) {
```
sujatapatnaik52: As Mike suggested, please add the reason for 77.

src/cmds/qstat.c
```diff
@@ -598,55 +599,83 @@ trunc_value(char *value, int len, int wide_len, int wide)
...
+			while (len) {
+				linebuf[i++] = *node_name++;
+				len--;
 			}
-
-			/* strip off domain name to keep string short */
-			while ((*stp != '+') && (*stp != ':') && (*stp != '\0'))
-				stp++;
-			nodes = stp++;
+			/* Appedning a  '+' here because we want to maintain the
```
sujatapatnaik52: Please correct the Typo for Appending.

src/cmds/qstat.c
```diff
@@ -599,75 +599,83 @@ trunc_value(char *value, int len, int wide_len, int wide)
...
 		exit_qstat("out of memory");
-	while ((token = strtok_r(rest, "+", &rest))){
-		token_cp = (char*)malloc(strlen(token) + 1);
-		if(token_cp)
-			strcpy(token_cp, token);
-		else
+	/* The exec_host string has the format <host1>/<T1>*<P1>[+<host2>/<T2>*<P2>+... ].
+	 * We are using '+' delimiter to find each <host1>/<T1>*<P1> string.
+	 */
+	token = strtok_r(rest, "+", &saveptr);
```
mike0042: Thank you for adding the comment. Why are you using the reentrant version of strtok?

src/cmds/qstat.c
```diff
@@ -599,75 +599,83 @@ trunc_value(char *value, int len, int wide_len, int wide)
...
 			/* flush line and start next */
 			linebuf[i] = '\0';
-			if (no_newl)
-				printf("%s", show_nonprint_chars(linebuf));
-			else
-				printf("   %s\n", show_nonprint_chars(linebuf));
+			printf((no_newl ? "%s" : "   %s\n"), show_nonprint_chars(linebuf));
 			i = 0;
-			while (l) {
+			while (len) {
```
mike0042: When a loop involves incrementing/decrementing a variable, it's almost always better suited as a for loop rather than a while loop.

src/cmds/qstat.c
```diff
@@ -599,75 +599,83 @@ trunc_value(char *value, int len, int wide_len, int wide)
...
-		else
-			printf("   %s\n", show_nonprint_chars(linebuf));
+		printf((no_newl ? "%s\n" : "   %s\n"), show_nonprint_chars(linebuf));
 	} else if (no_newl)
 		printf("\n");
+	if (token_cp) {
+		free(token_cp);
+		token_cp = NULL;
+	}
+	if (rest) {
```
mike0042: No need to test if a variable is NULL before you free it.

src/cmds/qstat.c
bhroam: I find using the reentrant version of strtok() best at all times.  It saves you from doing silly things like calling strtok() to break on fields, and then calling strtok() again to break into subfields.  There is no harm in using it.

src/cmds/qstat.c
npadole20: @sujatapatnaik52 @mike0042 we were using 77 in this function originally. So I went with it.

npadole20:
[qstat_n_review_2.txt](https://github.com/PBSPro/pbspro/files/3925946/qstat_n_review_2.txt)

src/cmds/qstat.c
mike0042: I find it strange that the variable "i" was initialized to zero and hasn't changed to this point. Perhaps a remnant of some old code that never got properly cleaned up? I suggest we remove it here.

Please use the following:
```suggestion
		if (len < PBS_MAXNODENAME) {
```
Also, change the declaration of linebuf to be of size PBS_MAXNODENAME + 1.

src/cmds/qstat.c
```diff
@@ -598,55 +599,81 @@ trunc_value(char *value, int len, int wide_len, int wide)
...
+			pbs_asprintf(&node_name, "%s%s", subtoken, chunk);
+		} else {
+			/* Node name is not an IP address */
+			pbs_asprintf(&node_name, "%s%s", strtok(subtoken, "."), chunk);
+		}
+		/* Backing up node_name as we are modifying the pointer further in the code */
+		node_name_bkp = node_name;
+		len = strlen(node_name);
+		if (i + len < 77) {
+			for (;len > 0; i++, len--) {
```
mike0042: Initialize i to zero in the for loop and remove the initialization earlier in this function. Better yet, use strncpy() to copy the string.

src/cmds/qstat.c
mike0042: Ahh... thanks @bhroam. I overlooked the fact there is another call to strtok here.

src/cmds/qstat.c
```diff
@@ -598,55 +599,81 @@ trunc_value(char *value, int len, int wide_len, int wide)
...
-				stp++;
-			nodes = stp++;
+			/* Appending a  '+' here because we want to maintain the
+			 * exec_host format i.e. <host1>/<T1>*<P1>[+<host2>/<T2>*<P2>+.
+			 */
+			linebuf[i++] = '+';
 		} else {
-			stp++;
+			/* flush line and start next */
+			linebuf[i] = '\0';
```
mike0042:
```suggestion
			linebuf[0] = '\0';
```

src/cmds/qstat.c
```diff
@@ -598,55 +599,81 @@ trunc_value(char *value, int len, int wide_len, int wide)
...
+			printf((no_newl ? "%s" : "   %s\n"), show_nonprint_chars(linebuf));
+			for (i = 0;len > 0; i++, len--) {
```
mike0042: Again, you could use strncpy here.

src/cmds/qstat.c
```diff
@@ -598,55 +599,81 @@ trunc_value(char *value, int len, int wide_len, int wide)
...
-			nodes = stp++;
+			/* Appending a  '+' here because we want to maintain the
+			 * exec_host format i.e. <host1>/<T1>*<P1>[+<host2>/<T2>*<P2>+.
+			 */
+			linebuf[i++] = '+';
 		} else {
-			stp++;
+			/* flush line and start next */
+			linebuf[i] = '\0';
+			printf((no_newl ? "%s" : "   %s\n"), show_nonprint_chars(linebuf));
```
mike0042: If linebuf is a zero length string, I don't see the point of this printf().

src/cmds/qstat.c
```diff
@@ -598,55 +599,81 @@ trunc_value(char *value, int len, int wide_len, int wide)
...
+	struct sockaddr_in check_ip;
+	int ret = 0;
 
 	if ((nodes == NULL) || (*nodes == '\0'))
 		return;
 
 	i = 0;
-	stp = nodes;
-	while (*nodes != '\0') {
-		if (*stp == '[')
```
mike0042: It looks like you removed the check for brackets, which is what "i" was being used for. That's the source of my confusion. I think the result is a slight change in behavior, and I don't think that's what we want.

src/cmds/qstat.c
mike0042: Ignore this. I think "i" needs to stay.

src/cmds/qstat.c
mike0042: Ignore this. I think "i" needs to stay.

src/cmds/qstat.c
mike0042: Ignore this.

src/cmds/qstat.c
mike0042: We may still need "i" here, but do please make the PBS_MAXNODENAME change.

npadole20: @hirenvadalia explained the value 77 comes from CHAR_LINE_LIMIT which has the value 78. It is not related to PBS_MAXNODENAME. I will change 77 and linebuf[78] to use CHAR_LINE_LIMIT - 1 and CHAR_LINE_LIMIT  respectively.

src/cmds/qstat.c
npadole20: The check here for '[' and ']' made sure that we did not remove the brackets from the the <host> of the exec_host string (<host>/<t1>*<p1> + <host1>/<t2>*<p2>). After the fix too, we are retaining the brackets. We are not specifically checking for brackets because if <host> is not a valid IP address, we are using the value of <host> as is.
I have attached the log which shows the behavior of qstat -n before and after the fix. 
[qstat_with_sq_bracket.txt](https://github.com/PBSPro/pbspro/files/3931568/qstat_with_sq_bracket.txt)

src/cmds/qstat.c
npadole20: I will have to change all place to use str* function. I would prefer to keep it as it is if that's okay.

src/cmds/qstat.c
```diff
@@ -598,55 +599,81 @@ trunc_value(char *value, int len, int wide_len, int wide)
...
+			/* node name is an IP address */
+			pbs_asprintf(&node_name, "%s%s", subtoken, chunk);
+		} else {
+			/* Node name is not an IP address */
+			pbs_asprintf(&node_name, "%s%s", strtok(subtoken, "."), chunk);
+		}
+		/* Backing up node_name as we are modifying the pointer further in the code */
+		node_name_bkp = node_name;
+		len = strlen(node_name);
+		if (i + len < CHAR_LINE_LIMIT - 1) {
```
mike0042: The variable "i" was initialized to zero at the start of the function and has remains unchanged. It's not needed here. Also, please add parenthesis around (CHAR_LINE_LIMIT - 1).

src/cmds/qstat.c
```diff
@@ -598,55 +599,81 @@ trunc_value(char *value, int len, int wide_len, int wide)
...
+			for (;len > 0; i++, len--) {
```
mike0042: This is the first use of i in this function. Please initialize it to zero and remove the prior initialization.

src/cmds/qstat.c
mike0042: That's fine, but please add a space after the first semicolon.

src/cmds/qstat.c
npadole20: Every "node_name" we read we append it to "linebuf" till we don't reach the limit of (CHAR_LINE_LIMIT - 1). So we need "i" here to keep track of what was the index when we updated "linebuf" with the "node_name" value from the previous iteration.

src/cmds/qstat.c
npadole20: If I initialize "i" to zero here, we will overwrite the value of "linebuf" from the previous iteration.

src/cmds/qstat.c
mike0042: OK, I get it now. Thanks for the explanation.

npadole20:
@mike0042 thank you.

PR #1427, opened 2019-12-02
pbs start/stop script fails to stop a busy scheduler

agrawalravi90:
#### Describe Bug or Feature

Post multi-threading, pbs start/stop script seems to not be able to stop a busy scheduler. The reason why this happens is that the MT code blocks some signals explicitly, one of them is SIGPIPE, which is what the scheduler gets when the server is down while the scheduler is still up and tries to talk to it. The scheduler already has signal handling for most signals, so the threads don't really need to block any explicitly. 

#### Describe Your Change

Removed the code in sched multi-threading which explicitly blocked signals.

**Before:**
[ravi@pbspro ~]$ sudo /etc/init.d/pbs stop
Stopping PBS
Shutting server down with qterm.
PBS server - was pid: 28901
PBS mom - was pid: 28513
PBS sched - was pid: 28525
PBS comm - was pid: 28482
Waiting for shutdown to complete
Unable to stop PBS, pbs_sched still active
[ravi@pbspro ~]$ 

[ravi@pbspro ~]$ ps -ef | grep pbs_sched
root     28525     1 16 18:16 ?        00:00:56 [pbs_sched] <defunct>
ravi     49527    17  0 18:22 pts/0    00:00:00 grep --color=auto pbs_sched
[ravi@pbspro ~]$ 

**After**
[root@pbspro /]# /etc/init.d/pbs stop
Stopping PBS
Shutting server down with qterm.
PBS server - was pid: 91480
PBS mom - was pid: 91090
PBS sched - was pid: 91102
PBS comm - was pid: 91059
Waiting for shutdown to complete
[root@pbspro /]# /etc/init.d/pbs status
pbs_server is not running
pbs_mom is not running
pbs_sched is not running
pbs_comm is not running
[root@pbspro /]#

### DIFF
--- a/src/scheduler/pbs_sched.c
+++ b/src/scheduler/pbs_sched.c
@@ -151,6 +151,9 @@ extern int do_hard_cycle_interrupt;
 static int	engage_authentication(struct connect_handle *);
 
 extern char *msg_startup1;
+
+static pthread_mutex_t cleanup_lock;
+
 /**
  * @brief
  * 		cleanup after a segv and re-exec.  Trust as little global mem
@@ -161,33 +164,35 @@ extern char *msg_startup1;
 void
 on_segv(int sig)
 {
-	int *thid;
+	int ret_lock = -1;
 
-	thid = (int *) pthread_getspecific(th_id_key);
-
-	if (thid == NULL || *thid == 0) {
-		/* we crashed less then 5 minutes ago, lets not restart ourself */
-		if ((segv_last_time - segv_start_time) < 300) {
-			log_record(PBSEVENT_SYSTEM, PBS_EVENTCLASS_SERVER, LOG_INFO,
-					"on_segv",
-					"received a sigsegv within 5 minutes of start: aborting.");
-			schedexit();
-			abort();
-		}
-
-		log_record(PBSEVENT_SYSTEM, PBS_EVENTCLASS_SERVER, LOG_INFO, "on_segv",
-				"received segv and restarting");
-
-		schedexit();
-
-		if (fork() > 0) { /* the parent rexec's itself */
-			sleep(10); /* allow the child to die */
-			execv(glob_argv[0], glob_argv);
-			exit(3);
-		} else
-			abort(); /* allow to core and exit */
-	} else
+	ret_lock = pthread_mutex_trylock(&cleanup_lock);
+	if (ret_lock != 0)
 		pthread_exit(NULL);
+
+	/* we crashed less then 5 minutes ago, lets not restart ourself */
+	if ((segv_last_time - segv_start_time) < 300) {
+		log_record(PBSEVENT_SYSTEM, PBS_EVENTCLASS_SERVER, LOG_INFO, __func__,
+				"received a sigsegv within 5 minutes of start: aborting.");
+		schedexit();
+		pthread_mutex_unlock(&cleanup_lock);
+		abort();
+	}
+
+	log_record(PBSEVENT_SYSTEM, PBS_EVENTCLASS_SERVER, LOG_INFO, __func__,
+			"received segv and restarting");
+
+	schedexit();
+
+	if (fork() > 0) { /* the parent rexec's itself */
+		sleep(10); /* allow the child to die */
+		execv(glob_argv[0], glob_argv);
+		pthread_mutex_unlock(&cleanup_lock);
+		exit(3);
+	} else {
+		pthread_mutex_unlock(&cleanup_lock);
+		abort(); /* allow to core and exit */
+	}
 }
 
 /**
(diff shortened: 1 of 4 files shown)


src/scheduler/pbs_sched.c
```diff
@@ -164,27 +164,30 @@ on_segv(int sig)
 	int *thid;
 
 	thid = (int *) pthread_getspecific(th_id_key);
-	if (thid != NULL && *thid != 0)
-		pthread_exit(NULL);
 
-	if (num_threads > 1)
-		kill_threads();
+	if (thid == NULL || *thid == 0) {
```
bhroam: I don't understand.  You only want to terminate if thread 0 gets a segv?  What if a global caused the segv?  What if there is massive memory corruption that crosses threads?  Don't we always want to exit on a segv?

src/scheduler/pbs_sched.c
```diff
@@ -164,27 +164,30 @@ on_segv(int sig)
...
+		/* we crashed less then 5 minutes ago, lets not restart ourself */
+		if ((segv_last_time - segv_start_time) < 300) {
+			log_record(PBSEVENT_SYSTEM, PBS_EVENTCLASS_SERVER, LOG_INFO,
+					"on_segv",
```
bhroam: Turn this into \_\_func\_\_ and move it up to the line above

src/scheduler/pbs_sched.c
```diff
@@ -164,27 +164,30 @@ on_segv(int sig)
...
+			abort();
+		}
 
-	/* we crashed less then 5 minutes ago, lets not restart ourself */
-	if ((segv_last_time - segv_start_time) < 300) {
-		log_record(PBSEVENT_SYSTEM, PBS_EVENTCLASS_SERVER, LOG_INFO, "on_segv", "received a sigsegv within 5 minutes of start: aborting.");
-		abort();
-	}
+		log_record(PBSEVENT_SYSTEM, PBS_EVENTCLASS_SERVER, LOG_INFO, "on_segv",
+				"received segv and restarting");
```
bhroam: use \_\_func\_\_

src/scheduler/pbs_sched.c
```diff
@@ -215,33 +218,33 @@ die(int sig)
...
-	if (sig > 0) {
-		sprintf(log_buffer, "caught signal %d", sig);
-		log_record(PBSEVENT_SYSTEM, PBS_EVENTCLASS_SERVER, LOG_INFO,
-			__func__, log_buffer);
-	}
-	else {
-		log_record(PBSEVENT_SYSTEM, PBS_EVENTCLASS_SERVER, LOG_INFO,
-				__func__, "abnormal termination");
-	}
+	if (thid == NULL || *thid == 0) {
```
bhroam: Once again here, I don't understand why we don't want to die if we get a signal.  Anytime we got this signal before, we'd die.  Now if it happens in a thread other than 0, we just exit the thread.

src/scheduler/pbs_sched.c
agrawalravi90: So, this code just says that if I'm thread 0, do some cleanup before I die. If you look at the 'else' part of this if, we do a pthread_exit(), that's the code that the other threads will execute if they get the segv. Does that make sense?

agrawalravi90: SEGV should go to all threads in a process, not just the one which caused it right?

bhroam: I saw the pthread_exit().  That doesn't kill the whole process though.  Why aren't we dying on all segvs, not just the ones that happen in thread 1?  Or is it that this handler is called N times for each of the N threads?

agrawalravi90:
@bhroam Thanks for your comments, I incorrectly thought that a signal sent to a process goes to all of its threads. I've now changed the code to be more robust for a multi-threading environment, please let me know what you think

src/scheduler/pbs_sched.c
```diff
@@ -161,30 +164,37 @@ extern char *msg_startup1;
...
 	}
 
-	log_record(PBSEVENT_SYSTEM, PBS_EVENTCLASS_SERVER, LOG_INFO, "on_segv", "received segv and restarting");
+	log_record(PBSEVENT_SYSTEM, PBS_EVENTCLASS_SERVER, LOG_INFO, __func__,
+			"received segv and restarting");
 
 	if (fork() > 0) { /* the parent rexec's itself */
-		sleep(10);		/* allow the child to die */
+		sleep(10); /* allow the child to die */
+		pthread_mutex_unlock(&cleanup_lock);
```
arungrover: I don't think you need unlock in either case. This will ensure that no thread is context switched before execv is called.

src/scheduler/pbs_sched.c
```diff
@@ -240,6 +249,8 @@ die(int sig)
 		}
 	}
 
+	pthread_mutex_unlock(&cleanup_lock);
```
arungrover: here too, if a thread got the lock and the process is going to die then we don't need to release the lock, do we?

src/scheduler/pbs_sched.c
```diff
@@ -1445,6 +1450,23 @@ main(int argc, char *argv[])
 		}
 	}
 
+	/* Initialize cleanup lock */
+	if (pthread_mutexattr_init(&attr) != 0) {
```
arungrover: It seems like this block of code is also used in multi_threading.c. Can this be made into a common function to generate recursive mutex?

PR #1428, opened 2019-12-02
Fixed issue preemption issue with RASSN resources

arungrover:
#### Describe Bug or Feature
There are two problems being fixed in this pull - 
1 - Scheduler fails to preempt a job when the resource in contention is a RASSN resource that can also be converted to select specification (like ncpus/mem). 
2 - Scheduler tries to filter preemptable candidates even in cases when the reason of not running a high priority job in the simulated universe does not change (especially when the reason is not related to a resource).

#### Describe Your Change
The first problem is fixed by checking that in case of INSUFFICIENT_SERVER_RESOURCE error lower priority job has requested the said resource. 
The second problem is fixed by checking if the old error code related to why job didn't run matches the new error code. If it matches and the resource definition is not set then the scheduler knows that the reason for not running the job is the same and it does not try to filter the running jobs again.

#### Link to Design Doc
NA

[test_out_after.txt](https://github.com/PBSPro/pbspro/files/3913794/test_out_after.txt)
[test_out_before.txt](https://github.com/PBSPro/pbspro/files/3913795/test_out_before.txt)
I have not added any test for the second fix, I thought that externally there is no visible change. It will have some improvement in performance but in my manual test I could see only milliseconds difference between filter_preemptable_jobs being called 3900 time to only 2 times.
If reviewers strongly feel that I should add a test for the second fix then I will add one.

!--- Pull Request Guidelines: [Pull Request Guidelines](https://pbspro.atlassian.net/wiki/spaces/DG/pages/1187348483/Pull+Request+Guidelines) -->

### DIFF
--- a/src/scheduler/job_info.c
+++ b/src/scheduler/job_info.c
@@ -3479,7 +3479,8 @@ find_jobs_to_preempt(status *policy, resource_resv *hjob, server_info *sinfo, in
 				"Simulation: Preempted enough work to run job");
 			rc = sim_run_update_resresv(npolicy, nhjob, ns_arr, NO_ALLPART);
 			break;
-		} else if (old_errorcode == err->error_code && err->rdef != NULL) {
+		} else if (old_errorcode == err->error_code) {
+			if (err->rdef != NULL) {
 			/* If the error code matches, make sure the resource definition is also matching.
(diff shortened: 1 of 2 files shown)


bhroam:
Travis failed due to pep8.  It says a blank line at the end of the file.

PR #1430, opened 2019-12-04
reset the size of TCP buffers

vchlum:
#### Describe Bug or Feature
The TCP buffers are never freed but they are reallocated and the size is increased sometimes (if necessary). This can cause the overall consumed memory of the server growing - especially with plenty of requests to the server and with plenty of jobs in the system.

I have prepared a simple test. Let's have 5030 jobs in the systems and no nodes.
```
qstat -x | wc -l
5030

pbsnodes -l
pbsnodes: Server has no node list
```
The steps are:

0) build PBS Pro with kerberos support (since kerberos adds some tcp buffers - this issue is more significant)
1) Started only pbs_server.bin (disabled scheduler, no pbs_comm started)
2) Waited for server initialization - cpu utilization 0
3) Noted the FIRST line (in the table below) from top command
4) Run: "for i in $(seq 1 10) ; do qstat -fwx > /dev/null ; done"
5) Waited for cpu usage returns to 0
4) Noted the SECOND line (in the table below) from top command
```
  PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
27198 root      20   0  214828  78740  15312 S   0.0   3.9   0:00.80 pbs_server.bin (FIRST)
27198 root      20   0  288232 153204  16440 S   0.0   7.5   0:02.87 pbs_server.bin (SECOND)
```

The consumed memory remains high.

#### Describe Your Change

I added a new function and this function resets the size of the buffers to the original size. This function is called on TCP setup and on connection close.

With the provided patch the same experiment looks like this:
```
  PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
 3290 root      20   0  214828  78528  15376 S   0.0   3.8   0:00.74 pbs_server.bin (FIRST)
 3290 root      20   0  234260  99328  16364 S   0.0   4.9   0:02.81 pbs_server.bin (SECOND)
```

I have considered using realloc instead of free + malloc but it would lead to high fragmentation in the memory. Free + malloc provides better results in this case. We do not need to preserve any data while the new function is called.

I have considered different places for the new function and IMHO the most important place where to call the resize function is on connection close. We could call it only on initialization but then unnecessary memory will remain consumed if the particular fd will not be used for a while. It is IMHO useful to call it on both the setup and close.

### DIFF
--- a/src/lib/Libifl/tcp_dis.c
+++ b/src/lib/Libifl/tcp_dis.c
@@ -764,6 +764,82 @@ DIS_tcp_funcs()
 
 /**
  * @brief
+ *	-resets tcp buffers size to the default size
+ *
+ * @param[in] fd - socket descriptor
+ * @param[in] on_release - indication connection is closing
+ *
+ * @return	Void
+ *
+ */
+void
+DIS_tcp_reset_buffers_size(int fd, int on_release)
+{
+	int rc;
+	struct tcp_chan *tcp;
+
+	if ((fd < 0) || (fd >= tcparraymax))
+		return;
+
+	rc = pbs_client_thread_lock_tcp();
+	assert(rc == 0);
+
+	tcp = tcparray[fd];
+	if (tcp) {
+		if (tcp->readbuf.tdis_bufsize > THE_BUF_SIZE) {
+			free(tcp->readbuf.tdis_thebuf);
+			tcp->readbuf.tdis_thebuf = malloc(THE_BUF_SIZE);
+			assert(tcp->readbuf.tdis_thebuf != NULL);
+			tcp->readbuf.tdis_bufsize = THE_BUF_SIZE;
+		}
+
+		if (tcp->writebuf.tdis_bufsize > THE_BUF_SIZE) {
+			free(tcp->writebuf.tdis_thebuf);
+			tcp->writebuf.tdis_thebuf = malloc(THE_BUF_SIZE);
+			assert(tcp->writebuf.tdis_thebuf != NULL);
+			tcp->writebuf.tdis_bufsize = THE_BUF_SIZE;
+		}
+
+#if defined(PBS_SECURITY) && (PBS_SECURITY == KRB5)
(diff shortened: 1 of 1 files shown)


subhasisb:
Calling it from DIS_tcp_setup() could be a bit of a performance impact. DIS_tcp_setup() is called every time from dis_request_read() which is basically for every packet received by the PBS server from any client, which could be hundreds per second. And now for each such packet, we will be freeing the buffer and allocating again. Seems like too often. 

How about we call the resize once in a while (like check a duration from last - perhaps once in 5 seconds or something) - maybe || it with a condition where try resize if the size is > than some threshold. That way, we will not try to resize if the size is already somewhat manageable, and not too frequently. What say?

subhasisb:
hi @vchlum one more thought. One of our developers is working on removing the global tcpbuffers array and attempting to put the buffers as part of the connection structure itself. Thus a buffer is only created when a connection is established - thus there may not be any need to resize at all.

Do you mind holding onto this change for a while? If we are successful in removing the global tcpbuffers array, then we do not need any of these changes. What do you think?

vchlum:
OK, it sounds like a great solution. I will withdraw this PR now. I was writing my response while you commented, so it will follow anyway. It is probably not important for now.
****
OK, it makes sense not to call it too often within DIS_tcp_setup(). I agree.

I still think that the most important place for this is on connection close. Maybe, we could call it only here...? When you say to call it only once in 5 seconds do you agree with calling it on connection close? IMHO, the cooldown would be barely used in such a case.

How about a configurable threshold? Let's say > 5\*THE_BUF_SIZE within DIS_tcp_setup() and > 1\*THE_BUF_SIZE on connection close? 

If we call the function only in DIS_tcp_setup() then I think the 5 seconds cooldown and threshold are both suitable but as a result, memory occupation would be higher because the fd may not be used for a while.

Idea: How about to use a simple counter instead of cooldown?

subhasisb:
yeah calling at close sounds much better. I will circle back once we know whether our refactoring is going to work or not. Thanks!

PR #1431, opened 2019-12-04
Python Interpreter Initialization Error Checking

ericpershey:
Added error checking to verify the python interpreter started inside pbs_main.  

#### Describe Bug or Feature
If the interpreter was not started, pbs_main would continue until python was needed and then segfault.

#### Describe Your Change
Added error checking after the call to pbs_python_ext_start_interpreter.


src/server/pbsd_main.c
```diff
@@ -1677,6 +1677,15 @@ main(int argc, char **argv)
 
 	pbs_python_ext_start_interpreter(&svr_interp_data);
 
+	if (!svr_interp_data.pbs_python_types_loaded ||
+			!svr_interp_data.interp_started ||
+			!svr_interp_data.data_initialized) {
+		log_err(-1, msg_daemonname, "Failed to start python interpreter!");
+		stop_db();
```
mike0042: Thank you for your contribution!

It might make more sense to initialize the Python interpreter before we attempt to connect to the database. We can detect the problem and exit sooner, avoiding unnecessary work. What do you think?

When referring to Python (as a proper name, not a command) please capitalize it.

bayucan: @mike0042 : I believe we have to start the database first before starting the Python  interpreter, for the interpreter would need PBS attributes and resources info from database, in order to instantiate server, queue, node, vnode, job objects...

ericpershey: Also, it looks like pbs_python_ext_start_interpreter call and supporting code would go around line 1251 but it looks like python is started and stopped somewhere below that with pbs_python_ext_quick_start_interpreter(1399) and pbs_python_ext_quick_shutdown_interpreter(1420).  

Updated python -> Python.

hirenvadalia: @ericpershey How about we move this check inside `pbs_python_ext_start_interpreter` itself and change the signature of `pbs_python_ext_start_interpreter` to `int` from `void` and return failure (-1 or 1). Because `pbs_python_ext_start_interpreter` is called from few other places also in PBS code.

```
pbspro/src/tools/pbs_python.c:2909:		pbs_python_ext_start_interpreter(&svr_interp_data);
pbspro/src/server/node_func.c:1440:			pbs_python_ext_start_interpreter(&svr_interp_data);
pbspro/src/server/pbsd_main.c:1678:	pbs_python_ext_start_interpreter(&svr_interp_data);
pbspro/src/server/setup_resc.c:85:	pbs_python_ext_start_interpreter(&svr_interp_data);
pbspro/src/server/node_manager.c:3460:		pbs_python_ext_start_interpreter(&svr_interp_data);
pbspro/src/server/node_manager.c:4171:		pbs_python_ext_start_interpreter(&svr_interp_data);
pbspro/src/lib/Libpython/pbs_python_svr_internal.c:5246:		pbs_python_ext_start_interpreter(&svr_interp_data);
``` 

@mike0042 and @bayucan What do you guys say about moving failure check inside `pbs_python_ext_start_interpreter`?

hirenvadalia: @ericpershey I don't think you need to return `(!svr_interp_data.pbs_python_types_loaded || !svr_interp_data.interp_started || !svr_interp_data.data_initialized)`, you can simply return 1 or 0.

mike0042: Either way is fine, IMHO. Zero for success, non-zero for failure.

bayucan: I'm fine with putting in the failure check inside pbs_python_ext_start_interpreter() and then going through each of the cases where the function is called and deal with having it or exit or just ignored.

ericpershey: I have pushed changes.  There are a few places where we have open questions.  Please see the FIXME comments.

src/server/setup_resc.c
```diff
@@ -78,11 +78,23 @@ struct resc_sum *svr_resc_sum;
...
 	log_event(PBSEVENT_DEBUG2, PBS_EVENTCLASS_HOOK,
 		LOG_INFO, (char *) caller,
 		"Restarting Python interpreter as resourcedef file has changed.");
-	pbs_python_ext_shutdown_interpreter(&svr_interp_data);
-	pbs_python_ext_start_interpreter(&svr_interp_data);
+	pyrc = pbs_python_ext_start_interpreter(&svr_interp_data);
+	if (pyrc != 0) {
+		log_err(PBSE_INTERNAL, (char *) caller, "Failed to restart Python interpreter");
+		/*
+		 * FIXME:
```
hirenvadalia: I think we have to ignore an error here, as callers are IS requests (qmgr resource set/unset/delete/create) and as of they don't return any error.

src/server/setup_resc.c
```diff
@@ -78,11 +78,23 @@ struct resc_sum *svr_resc_sum;
...
+		 * Should we return an error?
+		 */
+	}
+	/*
+	 * FIXME:
```
hirenvadalia: same as above.

src/server/pbsd_main.c
mike0042: All of the functions that call restart_python_interpreter() return void themselves, so aside from logging a message nothing else happens when a restart fails. For now, I think we should return whatever value pbs_python_ext_start_interpreter() returned from restart_python_interpreter() rather than void. At least that way the callers can choose to do something in the event of a failure in the future.

src/server/node_manager.c
```diff
@@ -4171,19 +4167,13 @@ update2_to_vnode(vnal_t *pvnal, int new, mominfo_t *pmom, int *madenew, int from
...
-		pyrc = pbs_python_ext_start_interpreter(&svr_interp_data);
-		if (pyrc != 0) {
+		if (pbs_python_ext_start_interpreter(&svr_interp_data) != 0) {
 			log_err(PBSE_INTERNAL, __func__, "Failed to restart Python interpreter");
-			/*
-			 * FIXME:
-			 * Should we return an error?
-			 * If so, does anything need to be undone?
-			 */
+			return PBSE_PYTHON_UNINITIALIZED;
```
mike0042: Rather than adding a new error type, I think we can just use PBSE_INTERNAL since that's what we log it as.

src/server/pbsd_main.c
ericpershey: So basically change the signature of restart_python_interpreter to return an int that is captured from pbs_python_ext_start_interpreter?

src/server/node_manager.c
ericpershey: Understood.

src/server/pbsd_main.c
mike0042: Exactly.

src/include/pbs_error.h
```diff
@@ -297,6 +297,7 @@ extern "C" {
 #define PBSE_SCHED_PARTITION_ALREADY_EXISTS 15224 /* Partition already exists */
 #define PBSE_INVALID_MAX_JOB_SEQUENCE_ID 15225 /* Invalid max_job_sequence_id < 9999999, or > 999999999999 */
 #define PBSE_SVR_SCHED_JSF_INCOMPAT 15226	/* Server's job_sort_formula is incompatible with sched's */
+#define PBSE_PYTHON_UNINITIALIZED 15227 /* Python is not initialized */
```
mike0042: This is no longer necessary. The rest of the changes look good. Thank you!

mike0042:
@bayucan : please review the latest commits from @ericpershey

src/server/setup_resc.c
```diff
@@ -75,14 +75,18 @@ struct resc_sum *svr_resc_sum;
...
+int
 restart_python_interpreter(const char *caller)
 {
+	int rc;
 	log_event(PBSEVENT_DEBUG2, PBS_EVENTCLASS_HOOK,
 		LOG_INFO, (char *) caller,
 		"Restarting Python interpreter as resourcedef file has changed.");
-	pbs_python_ext_shutdown_interpreter(&svr_interp_data);
-	pbs_python_ext_start_interpreter(&svr_interp_data);
+	rc = pbs_python_ext_start_interpreter(&svr_interp_data);
```
bayucan: The pbs_python_ext_shutdown_interpreter() call disappeared here....

mike0042:
@bayucan : Please re-review.

src/lib/Libpython/pbs_python_external.c
```diff
@@ -249,17 +249,19 @@ pbs_python_ext_start_interpreter(
...
 
 	interp_data->pbs_python_types_loaded = 1; /* just in case */
-	return;
-
+SUCCESS_EXIT:
+	return 0;
 ERROR_EXIT:
-	pbs_python_ext_shutdown_interpreter(interp_data);
-	return;
+	if (interp_data->interp_started) {
```
agrawalravi90: Styling: no need to surround single line if blocks with curly braces (Ref: https://pbspro.atlassian.net/wiki/spaces/DG/pages/5537826/Coding+Standards)

src/server/setup_resc.c
```diff
@@ -75,14 +75,19 @@ struct resc_sum *svr_resc_sum;
...
 {
+	int rc;
 	log_event(PBSEVENT_DEBUG2, PBS_EVENTCLASS_HOOK,
 		LOG_INFO, (char *) caller,
 		"Restarting Python interpreter as resourcedef file has changed.");
 	pbs_python_ext_shutdown_interpreter(&svr_interp_data);
-	pbs_python_ext_start_interpreter(&svr_interp_data);
+	rc = pbs_python_ext_start_interpreter(&svr_interp_data);
+	if (rc != 0) {
+		log_err(PBSE_INTERNAL, (char *) caller, "Failed to restart Python interpreter");
```
agrawalravi90: Styling: no need to surround single line if blocks with curly braces

src/tools/pbs_python.c
```diff
@@ -2906,7 +2906,10 @@ main(int argc, char *argv[], char *envp[])
 			(struct python_script **) &py_script);
 
 		hook_perf_stat_start(perf_label, HOOK_PERF_START_PYTHON, 0);
-		pbs_python_ext_start_interpreter(&svr_interp_data);
+		if (pbs_python_ext_start_interpreter(&svr_interp_data) != 0) {
+			fprintf(stderr, "Failed to start Python interpreter");
+			exit(1);
```
agrawalravi90: free py_script and svr_interp_data.daemon_name before exiting ?

agrawalravi90:
the PR got merged while I was reviewing :) there were no major comments, so it's ok.

mike0042:
> the PR got merged while I was reviewing :) there were no major comments, so it's ok.

Sorry about that @agrawalravi90, but the PR met the criteria for merging. Like you mentioned, no major issues.

PR #1434, opened 2019-12-09
Server resolves node name when they are added using IP address

SudeshnaMoulik:
#### Describe Bug or Feature
Load balancing doesn't work when nodes are added using ip address and Server resolves that to node name (hostname).

Snippet of "pbsnodes -av"
10.0.9.120
     Mom = ip-10-0-9-120.us-west-1.compute.internal
     

#### Describe Your Change
When nodes are added using their ip address, server shouldn't try to resolve their name.

[Before_fix.txt](https://github.com/PBSPro/pbspro/files/3938493/Before_fix.txt)
[After_fix.txt](https://github.com/PBSPro/pbspro/files/3938494/After_fix.txt)
[PTL_logs.txt](https://github.com/PBSPro/pbspro/files/3938509/PTL_logs.txt)

Valgring Log : [val.txt](https://github.com/PBSPro/pbspro/files/3938864/val.txt)

Passed all existing regression tests & Build and tests across platforms.

### DIFF
--- a/src/server/node_func.c
+++ b/src/server/node_func.c
@@ -2380,15 +2380,13 @@ decode_Mom_list(struct attribute *patr, char *name, char *rescn, char *val)
 	for (i = 0; (p = str_arr[i]) != NULL; i++) {
 		clear_attr(&new, &node_attr_def[(int)ND_ATR_Mom]);
 		is_node_name_ip = inet_pton(AF_INET, p, &(check_ip.sin_addr)) ;
-		if(!is_node_name_ip) {
-			if (get_fullhostname(p, buf, (sizeof(buf) - 1)) != 0) {
-				strncpy(buf, p, (sizeof(buf) - 1));
-				buf[sizeof(buf) - 1] = '\0';
-			}
-		} else {
+		if(!is_node_name_ip)
+			is_node_name_ip = inet_pton(AF_INET6, p, &(check_ip.sin_addr)) ;
+		if(is_node_name_ip || get_fullhostname(p, buf, (sizeof(buf) - 1)) != 0) {
 			strncpy(buf, p, (sizeof(buf) - 1));
 			buf[sizeof(buf) - 1] = '\0';
 		}
+		
 		rc = decode_arst(&new, ATTR_NODE_Mom, NULL, buf);
 		if (rc != 0)
 			continue;
--- a/src/server/node_func.c
+++ b/src/server/node_func.c
@@ -2380,8 +2380,6 @@ decode_Mom_list(struct attribute *patr, char *name, char *rescn, char *val)
 	for (i = 0; (p = str_arr[i]) != NULL; i++) {
 		clear_attr(&new, &node_attr_def[(int)ND_ATR_Mom]);
 		is_node_name_ip = inet_pton(AF_INET, p, &(check_ip.sin_addr)) ;
-		if(!is_node_name_ip)
-			is_node_name_ip = inet_pton(AF_INET6, p, &(check_ip.sin_addr)) ;
 		if(is_node_name_ip || get_fullhostname(p, buf, (sizeof(buf) - 1)) != 0) {
 			strncpy(buf, p, (sizeof(buf) - 1));
 			buf[sizeof(buf) - 1] = '\0';
--- a/src/server/node_func.c
+++ b/src/server/node_func.c
@@ -2340,6 +2340,8 @@ decode_Mom_list(struct attribute *patr, char *name, char *rescn, char *val)
 	static char		**str_arr = NULL;
 	static long int		  str_arr_len = 0;
 	attribute		  new;
+	struct sockaddr_in check_ip;
+	int is_node_name_ip;
 
 	if ((val == NULL) || (strlen(val) == 0) || count_substrings(val, &ns)) {
 		node_attr_def[(int)ND_ATR_Mom].at_free(patr);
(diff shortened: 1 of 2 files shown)


src/server/node_func.c
```diff
@@ -2377,7 +2379,13 @@ decode_Mom_list(struct attribute *patr, char *name, char *rescn, char *val)
 
 	for (i = 0; (p = str_arr[i]) != NULL; i++) {
 		clear_attr(&new, &node_attr_def[(int)ND_ATR_Mom]);
-		if (get_fullhostname(p, buf, (sizeof(buf) - 1)) != 0) {
+		is_node_name_ip = inet_pton(AF_INET, p, &(check_ip.sin_addr)) ;
+		if(!is_node_name_ip) {
+			if (get_fullhostname(p, buf, (sizeof(buf) - 1)) != 0) {
+				strncpy(buf, p, (sizeof(buf) - 1));
+				buf[sizeof(buf) - 1] = '\0';
```
npadole20: Since we are doing the above 2 steps in both if and else part, is it possible to take them out of the if...else condition?

SudeshnaMoulik: Hey @npadole20 I think that wouldn't be correct, as we can see the purpose of that part of code is different here. That mean setting of buffer "when we don't get the full hostname of a node which we added using node name" vs "when we added node using ip address".

agrawalravi90: I think I agree with Neha, you can convert this to:
```
if (if_node_name_ip || get_fullhostname(p, buf, (sizeof(buf) - 1)) == 0) {
    strncpy(buf, p, (sizeof(buf) - 1));
    buf[sizeof(buf) - 1] = '\0';
}
```

right?

src/server/node_func.c
```diff
@@ -2377,7 +2379,13 @@ decode_Mom_list(struct attribute *patr, char *name, char *rescn, char *val)
 
 	for (i = 0; (p = str_arr[i]) != NULL; i++) {
 		clear_attr(&new, &node_attr_def[(int)ND_ATR_Mom]);
-		if (get_fullhostname(p, buf, (sizeof(buf) - 1)) != 0) {
+		is_node_name_ip = inet_pton(AF_INET, p, &(check_ip.sin_addr)) ;
```
agrawalravi90: Do we need to check for a IPv6 address as well?

test/tests/functional/pbs_sched_load_balancing.py
```diff
@@ -0,0 +1,75 @@
...
+class TestSchedLoadBalancing(TestFunctional):
+    """
+    Test suite for PBSPro's Scheduler load balancing
+    """
+    def setUp(self):
+        """
+        Set scheduler load_balancing
+        """
+        TestFunctional.setUp(self)
+        self.scheduler.set_sched_config({'load_balancing': 'true	ALL'})
```
agrawalravi90: since you have just one test right now, it's probably best not to turn on load balancing in setUp(), if we add more tests to this suite and all of them need it to be on by default then we can add this to setUp().

test/tests/functional/pbs_sched_load_balancing.py
```diff
@@ -0,0 +1,75 @@
...
+    def test_load_bal_node_ip(self):
+        """
+        This test case tests scheduler load balancing
+        feature, when node is added using ip address.
+        """
+        # Check current load in system
+        cmd = os.path.join(self.server.pbs_conf[
+                           'PBS_EXEC'], 'unsupported', 'pbs_rmget')
+        cmd += ' -m ' + self.mom.hostname + ' loadave '
+        ret = self.server.du.run_cmd(self.mom.hostname, cmd)
```
agrawalravi90: you can just do "self.du"

test/tests/functional/pbs_sched_load_balancing.py
```diff
@@ -0,0 +1,75 @@
...
+        current_load = float(ret['out'][0].split('=')[1])
+        # Considering our job will need 1 cpu
+        ideal_load = current_load + 1
+        max_load = current_load + 2
+        self.mom.add_config({'$ideal_load': ideal_load})
+        self.mom.add_config({'$max_load': max_load})
+        self.server.manager(MGR_CMD_DELETE, NODE, None, '')
+        ipaddr = socket.gethostbyname(self.mom.hostname)
+        self.server.manager(MGR_CMD_CREATE, NODE,  id=ipaddr)
```
agrawalravi90: don't you need to add the node to the vnode def file?

src/server/node_func.c
SudeshnaMoulik: Yes agree with both of you. Thanks, I will make the change

test/tests/functional/pbs_sched_load_balancing.py
SudeshnaMoulik: But this is a Scheduler's load_balancing test suite and that set up is to turn on the feature. So later also if we add any test cases under this suite they probably need this.

test/tests/functional/pbs_sched_load_balancing.py
SudeshnaMoulik: I don't think so. Is there any reason to add node in vnode def file in this scenario ?

SudeshnaMoulik:
Latest logs:

[PTL_logs.txt](https://github.com/PBSPro/pbspro/files/3943081/PTL_logs.txt)

test/tests/functional/pbs_sched_load_balancing.py
agrawalravi90: AFAIK, just creating a node via qmgr is not very useful, the mom has to report the nodes as well, that's when the nodes become "free", otherwise they remain in the "stale" state. For the mom to report any nodes, they need to be part of its vnodedef file. Right?

SudeshnaMoulik: Even though we don't add that server will run it on natural vnode. Its similar like when we add any node using 
"qmgr -c 'c n node_name' " , node adds to cluster and state becomes as free. And in the next line we are already checking that also.

agrawalravi90: When I try to create a vnode via qmgr only, this is what happens:

[ravi@pbspro ~]$ /opt/pbs/bin/qmgr -c 'c n node_name'
qmgr obj=node_name svr=default: Unknown node 
qmgr: Error (15062) returned from server
[ravi@pbspro ~]$ 

Can you try running this test on a test system?

agrawalravi90: ok wait ,you are adding the node with node name as the hostname/ipaddress of the server host itself, ok ya in that case it will work, nevermind

SudeshnaMoulik:
@npadole20  @agrawalravi90  Thanks for the review.
Please let me know whether am I good to squash multiple commits ?

src/server/node_func.c
agrawalravi90: So, were you going to handle IPv6 as well?

SudeshnaMoulik: In PBS we don't support adding node with their IPv6 address. So i reverted the corresponding changes i made.

SudeshnaMoulik:
In PBS we don't support adding node with their IPv6 address. So i reverted the corresponding changes i made.

PR #1435, opened 2019-12-09
TestPbsAccumulateRescUsed expects string in wrong order

sandisamp:
#### Describe Bug or Feature

In tests TestPbsAccumulateRescUsed.test_mom_down2 and TestPbsAccumulateRescUsed.test_server_restart2, the mom hooks set foo_str as a json object. The primary mom then merges the json objects and returns it.

The test expects the object's keys to appear in a certain order, but the mom returns it in a different order:

ptl.lib.pbs_testlib.PtlExpectError: rc=1, rv=False, msg=expected on server pbs-server@/etc/pbs.conf2: resources_used.foo_i = 29 && resources_used.foo_f = 0.29 && resources_used.foo_str = '{"eight": 8, "seven": 7, "nine": 9}' job 85.pbs-server attempt: 60 got: resources_used.foo_str = '{"eight": 8, "nine": 9, "seven": 7}'

#### Describe Your Change

While decoding the values it was being returned in a string form. The function decode_value now identifies a dictionary in string form and returns a dictionary value instead.

[PTL logs for TestPbsAccumulateRescUsed.test_mom_down2 and TestPbsAccumulateRescUsed.test_server_restart2 ](https://github.com/PBSPro/pbspro/files/3938750/ptl_logs.txt)

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -2091,7 +2091,7 @@ class BatchUtils(object):
         :returns: int or float or string
         """
         if re.search('(\{.*:.*[,]?.*\})', str(value)):
-            value = json.loads(eval(str(value)))
+            value = ast.literal_eval(eval(str(value)))
             return value
 
         if value is None or isinstance(value, collections.Callable):
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -2083,16 +2083,12 @@ class BatchUtils(object):
         of it, if it is made of alphanumeric values only, return
         it as a string, if it is of type size, i.e., with a memory
         unit such as b,kb,mb,gb then return the converted size to
-        kb without the unit, if value is a dictionary string return
-        dictionary value
+        kb without the unit
 
         :param value: attribute/resource value
         :type value: str or int
         :returns: int or float or string
         """
-        if re.search('(\{.*:.*[,]?.*\})', str(value)):
-            value = ast.literal_eval(ast.literal_eval(str(value)))
-            return value
 
         if value is None or isinstance(value, collections.Callable):
             return value
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -2083,12 +2083,17 @@ class BatchUtils(object):
         of it, if it is made of alphanumeric values only, return
         it as a string, if it is of type size, i.e., with a memory
         unit such as b,kb,mb,gb then return the converted size to
-        kb without the unit
+        kb without the unit, if value is a dictionary string return 
+        dictionary value
 
         :param value: attribute/resource value
         :type value: str or int
         :returns: int or float or string
         """
+        if re.search('\'{\".*\":.*',str(value)):
+            value = eval(eval(str(value)))
+            return value
+        
         if value is None or isinstance(value, collections.Callable):
             return value
 
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -2083,7 +2083,7 @@ class BatchUtils(object):
         of it, if it is made of alphanumeric values only, return
         it as a string, if it is of type size, i.e., with a memory
         unit such as b,kb,mb,gb then return the converted size to
-        kb without the unit, if value is a dictionary string return 
+        kb without the unit, if value is a dictionary string return
         dictionary value
 
         :param value: attribute/resource value
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -2091,7 +2091,7 @@ class BatchUtils(object):
         :returns: int or float or string
         """
         if re.search('(\{.*:.*[,]?.*\})', str(value)):
-            value = eval(eval(str(value)))
+            value = json.loads(eval(str(value)))
             return value
 
         if value is None or isinstance(value, collections.Callable):
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -2090,10 +2090,10 @@ class BatchUtils(object):
         :type value: str or int
         :returns: int or float or string
         """
-        if re.search('\'{\".*\":.*',str(value)):
+        if re.search('\'{\".*\":.*', str(value)):
             value = eval(eval(str(value)))
             return value
-        
+
         if value is None or isinstance(value, collections.Callable):
             return value
 
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -2090,7 +2090,7 @@ class BatchUtils(object):
         :type value: str or int
         :returns: int or float or string
         """
-        if re.search('\'{\".*\":.*', str(value)):
+        if re.search('(\{.*:.*[,]?.*\})', str(value)):
             value = eval(eval(str(value)))
             return value
 
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -2091,7 +2091,7 @@ class BatchUtils(object):
         :returns: int or float or string
         """
         if re.search('(\{.*:.*[,]?.*\})', str(value)):
-            value = ast.literal_eval(eval(str(value)))
+            value = ast.literal_eval(ast.literal_eval(str(value)))
             return value
 
         if value is None or isinstance(value, collections.Callable):
--- a/test/tests/functional/pbs_accumulate_resc_used.py
+++ b/test/tests/functional/pbs_accumulate_resc_used.py
@@ -35,6 +35,7 @@
 # "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
 # trademark licensing policies.
 from tests.functional import *
+import ast
 
 
 @requirements(num_moms=3)
@@ -1252,11 +1253,16 @@ time.sleep(15)
 
         # Verify the resource values
         a = {'resources_used.foo_i': 29,
-             'resources_used.foo_f': 0.29,
-             'resources_used.foo_str':
-             "\'{\"eight\": 8, \"seven\": 7, \"nine\": 9}\'"}
+             'resources_used.foo_f': 0.29}
+        a_dict = {'eight': 8, 'seven': 7, 'nine': 9}
+
         self.server.expect(JOB, a, extend='x', attrop=PTL_AND,
                            offset=5, id=jid, interval=1)
+        # check for dictionary resource
+        job_status = self.server.status(JOB, id=jid, extend='x')
+        job_str_resource = dict(job_status[0])['resources_used.foo_str']
+        job_str_resource = ast.literal_eval(ast.literal_eval(job_str_resource))
+        self.assertEqual(job_str_resource, a_dict)
 
         # Restart server while hook is still executing
         self.server.restart()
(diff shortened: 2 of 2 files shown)


test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -2083,12 +2083,17 @@ def decode_value(cls, value):
...
         unit such as b,kb,mb,gb then return the converted size to
-        kb without the unit
+        kb without the unit, if value is a dictionary string return
+        dictionary value
 
         :param value: attribute/resource value
         :type value: str or int
         :returns: int or float or string
         """
+        if re.search('\'{\".*\":.*', str(value)):
```
agrawalravi90: You are assuming that the dict will have string key & value, so this will not work for a dict that has numeric key or values.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -2090,7 +2090,7 @@ def decode_value(cls, value):
         :type value: str or int
         :returns: int or float or string
         """
-        if re.search('\'{\".*\":.*', str(value)):
+        if re.search('(\{.*:.*[,]?.*\})', str(value)):
             value = eval(eval(str(value)))
```
agrawalravi90: Why 2 evals ? It seems to fail for me:
>>> mystr = "{1:\"blah\", 2:[1, 'a', \"blah\"], 3: {1:2, 3:4}}"
>>> eval(str(mystr))
{1: 'blah', 2: [1, 'a', 'blah'], 3: {1: 2, 3: 4}}
>>> eval(eval(str(mystr)))
TypeError: eval() arg 1 must be a string, bytes or code object
>>>

sandisamp: With single eval statement we get this:
`2019-12-11 03:45:50,520 INFO     expect on server pbs-server@/etc/pbs.conf2: resources_used.foo_i = 29 && resources_used.foo_f = 0.29 && resources_used.foo_str = '{"eight": 8, "seven": 7, "nine": 9}' job 60.pbs-server attempt: 60  got: resources_used.foo_str = {"eight": 8, "nine": 9, "seven": 7}`

This is because with single eval type(value) is still string. That is why I have used double eval.

hirenvadalia: I would suggest using `json.loads` instead eval, as "eval" is "evil".

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -2083,12 +2083,17 @@ def decode_value(cls, value):
...
-        kb without the unit
+        kb without the unit, if value is a dictionary string return
+        dictionary value
 
         :param value: attribute/resource value
         :type value: str or int
         :returns: int or float or string
         """
+        if re.search('(\{.*:.*[,]?.*\})', str(value)):
+            value = json.loads(eval(str(value)))
```
agrawalravi90: unfortunately, it doesn't seem like this works either. It only works for dicts which have string keys, for dicts which have numeric keys, it doesn't work:
```

>>> mystr = '{1:"blah", 2:[1, "a", "blah"], 3: {1:2, 3:4}}'
>>> json.loads(mystr)
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/json/__init__.py", line 348, in loads
    return _default_decoder.decode(s)
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/json/decoder.py", line 353, in raw_decode
    obj, end = self.scan_once(s, idx)
json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes: line 1 column 2 (char 1)
>>> json.loads(eval(str(mystr)))
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/json/__init__.py", line 341, in loads
    raise TypeError(f'the JSON object must be str, bytes or bytearray, '
TypeError: the JSON object must be str, bytes or bytearray, not dict
```

How about using ast.literal_eval() instead?

hirenvadalia: I am ok with `ast.literal_eval()`.

sandisamp:
Adding new test logs:
[TestPbsAccumulateRescUsed.test_mom_down2](https://github.com/PBSPro/pbspro/files/3954233/logfile_PASS.txt)
[TestPbsAccumulateRescUsed.test_server_restart2](https://github.com/PBSPro/pbspro/files/3954243/logfile_PASS.txt)

sandisamp:
Adding new test logs:
[TestPbsAccumulateRescUsed.test_server_restart2](https://github.com/PBSPro/pbspro/files/3954606/logfile_PASS.txt)
[TestPbsAccumulateRescUsed.test_mom_down2](https://github.com/PBSPro/pbspro/files/3954618/logfile_PASS.txt)

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -2083,12 +2083,17 @@ def decode_value(cls, value):
...
-        kb without the unit
+        kb without the unit, if value is a dictionary string return
+        dictionary value
 
         :param value: attribute/resource value
         :type value: str or int
         :returns: int or float or string
         """
+        if re.search('(\{.*:.*[,]?.*\})', str(value)):
+            value = ast.literal_eval(ast.literal_eval(str(value)))
```
agrawalravi90: Why do you need to call it twice? Calling it once works for me:

```
>>> value='{1:"blah", 2:[1, "a", "blah"], 3: {1:2, 3:4}}'
>>> ast.literal_eval(value)
{1: 'blah', 2: [1, 'a', 'blah'], 3: {1: 2, 3: 4}}
```

^ The value above is already a dict type now, not string. Calling it twice infact errors out:

```
>>> ast.literal_eval(ast.literal_eval(value))
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.7/lib/python3.7/ast.py", line 91, in literal_eval
    return _convert(node_or_string)
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.7/lib/python3.7/ast.py", line 90, in _convert
    return _convert_signed_num(node)
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.7/lib/python3.7/ast.py", line 63, in _convert_signed_num
    return _convert_num(node)
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.7/lib/python3.7/ast.py", line 55, in _convert_num
    raise ValueError('malformed node or string: ' + repr(node))
ValueError: malformed node or string: {1: 'blah', 2: [1, 'a', 'blah'], 3: {1: 2, 3: 4}}
```

sandisamp: I have added test results, those infact work fine.
The value passed in the function is actually this:
`value='\'{1:"blah", 2:[1, "a", "blah"], 3: {1:2, 3:4}}\''`

So first time the method is called it evaluates to string, and second time to dict

agrawalravi90: That's not really a dict in string form though right? That's just a string. The function doesn't decode \'\2\'' to numeric 2, so I don't think it should decode '\'{1:"blah", 2:[1, "a", "blah"], 3: {1:2, 3:4}}\'' to a dict. Where is this value coming from? If the test creates it then it should remove the escaped quotes before passing it to this function.

hirenvadalia: I agree with @agrawalravi90 here.

sandisamp:
> When all values are valid JSON format, the resulting string resource value is a union of all dictionary items, shown in qstat -f output and accounting logs as:
> 
> If MoM A returns '{ "a":1, "b":2 }', MoM B returns '{ "c":1 }', and MoMC returns '{"d":4}' for
> resources_used.foo_str='{"a": 1, "b": 2, "c":1,"d": 4}'

Given the code of the test case:
```
def test_server_restart2(self):
        """
        Test that server restart during hook execution
        has no impact
        """

        hook_body = """
import pbs
import time

e = pbs.event()

pbs.logmsg(pbs.LOG_DEBUG, "executed epilogue hook")
if e.job.in_ms_mom():
        e.job.resources_used["vmem"] = pbs.size("9gb")
        e.job.resources_used["foo_i"] = 9
        e.job.resources_used["foo_f"] = 0.09
        e.job.resources_used["foo_str"] = '{"seven":7}'
        e.job.resources_used["cput"] = 10
else:
        e.job.resources_used["vmem"] = pbs.size("10gb")
        e.job.resources_used["foo_i"] = 10
        e.job.resources_used["foo_f"] = 0.10
        e.job.resources_used["foo_str"] = '{"eight":8,"nine":9}'
        e.job.resources_used["cput"] = 20

time.sleep(15)
"""
.
.
.

        # Verify the resource values
        a = {'resources_used.foo_i': 29,
             'resources_used.foo_f': 0.29,
             'resources_used.foo_str':
             "\'{\"eight\": 8, \"seven\": 7, \"nine\": 9}\'"}
        self.server.expect(JOB, a, extend='x', attrop=PTL_AND,
                           offset=5, id=jid, interval=1)

        # Restart server while hook is still executing
        self.server.restart()

        # Verify that values again
        self.server.expect(JOB, a, extend='x', attrop=PTL_AND,
                           id=jid)
```

It is correctly defining foo_str and expecting the return format as string ,PBS qstat behavior is dictating this. therefore this should not be modified in the test case itself.

This makes this a special scenario in case of dictionary values and should be handled accordingly by the decode_value method itself as it's expected to decode all qstat -f attributes/resources. 

According to me this is the place where this string should be handled and returned as dict.

hirenvadalia:
@sandisamp But your above example you are asking Server.expect to look for str for resources_used.foo_str and as you said in above qstat -f returns str for resources_used.foo_str also hook sets resources_used.foo_str to str value then why we have to convert that value in python dict?
In PBS everywhere resources_used.foo_str is str then why in PTL we need dict?

sandisamp:
@hirenvadalia The value returned is union of dictionary in str format. As i have quoted in my previous comment. The order of this union is different from what the expected order which the test case is passing in expect. Comparing them as string will fail. 

The PTL framework should be able to handle this scenario. That's why we are converting both expected value and returned value back to dict for comparison. We cannot expect the user to predict the order in which the union of dict will be returned.

hirenvadalia:
@sandisamp In that case how PBS is deciding the order of union of dict in str format.
Can you try changing the key names in the hook and see what order it comes in qstat -f?
If I am not wrong then PBS will return in alphabetic order of keys.

sandisamp:
I added {"one" :1} to a test case and this is the value that was returned:
got: resources_used.foo_str = {'eight': 8, 'nine': 9, 'seven': 7, 'one': 1}

agrawalravi90:
Just to get this out of the way, PBS doesn't support a json type resource, so I don't think PTL has to either. So, I think it does come down to this test doing something peculiar.

I'm actually very confused why this test is doing what it's doing. string type resources are non-consumables in PBS. Which means that PBS doesn't set 'resources_used' for non-consumables,  ergo, string type resources. So, we are forcing this to be set via the hook, and PBS is probably just doing a string concatenate in whatever order it receives the values from Mom. But the test doesn't really need to do this right? I see foo_str adding no value to this test's objective. So, can we just remove it?

If we do want to keep it, then I suggest doing this:
- stat the value of foo_str from the job
- convert it to a dict inside the test itself and verify all keys and values

Again, I think PTL doesn't need to support json type resource because PBS doesn't either. decode_value() is doing the right thing today by decoding it to a string type resource, because that's exactly what it is.

sandisamp:
[test_server_restart2.txt](https://github.com/PBSPro/pbspro/files/3967720/test_server_restart2.txt)
[test_mom_down2.txt.txt](https://github.com/PBSPro/pbspro/files/3967722/test_mom_down2.txt.txt)
Added new logs

PR #1436, opened 2019-12-09
Fixed permission issue in pbs_testlib.create_eatcpu_job.

ericpershey:
eatcpu.py requires execute permissions in order to run.  chmod required
sudo to change file permissions.  This fix is for issue #1267

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -14218,6 +14218,11 @@ class Job(ResourceResv):
         """
         Create a job that eats cpu indefinitely or for the given
         duration of time
+
+        :param duration: The duration, in seconds, to sleep
+        :type duration: int
+        :param mom: the MoM object on which to execute the job
+        :type mom: str
         """
         if self.du is None:
             self.du = DshUtils()
@@ -14235,7 +14240,10 @@ class Job(ResourceResv):
                                   'bin', 'pbs_python')
         a = {ATTR_S: shell_path}
         self.set_attributes(a)
-        self.du.chmod(path=script_path, mode=0o755)
+        mode = 0o755
+        if not self.du.chmod(path=script_path, mode=mode, sudo=True):
+            raise AssertionError("Failed to set permissions for file %s"
+                                 " to %s" % (script_path, oct(mode)))
         self.set_execargs(script_path, duration)
 
 
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -14214,34 +14214,34 @@ class Job(ResourceResv):
         return job_array_id[:idx + 1] + str(subjob_index) + \
             job_array_id[idx + 1:]
 
-    def create_eatcpu_job(self, duration=None, mom=None):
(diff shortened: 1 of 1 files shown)


test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -14235,7 +14240,10 @@ def create_eatcpu_job(self, duration=None, mom=None):
                                   'bin', 'pbs_python')
         a = {ATTR_S: shell_path}
         self.set_attributes(a)
-        self.du.chmod(path=script_path, mode=0o755)
+        mode = 0o755
+        if not self.du.chmod(path=script_path, mode=mode, sudo=True):
```
bhroam: Does this need a host?

toonen: Can you please elaborate?  @ericpershey can correct me if I am wrong, but the environment in which we have been testing is a single node running both the server and a mom.  My understanding from the du.chmod docstring is that specifying a node is only necessary in a multi-node environment.  Again, we are new to PBS Pro and my reading of the code and resulting output could be wrong, but our observation is that the eatcpu.py code installed in /opt/ptl is missing the executable bit which results in the test failing when the test is on a single node.

ericpershey: @toonen I think he is right, it needs a host.  The signature of create_eatcpu_job does contain mom, a hostname, and that hostname is passed to run_copy.  @toonen and I talked about adding it to run_copy, but some of the other functions seem to do a run_copy then chmod right after.  Thoughts?

kjakkali:
@toonen : This issue is already fix as a part of #1375 . Are you still seeing this issue?

ericpershey:
@kjakkali This problem seems to be from the line ```self.du.chmod(path=script_path, mode=0o755)``` from #1375.  The chmod seems to need to run as sudo, which fails and only logs an error, then it cannot run the eatcpu.py on the localhost as a mom.  It was also pointed out that the du.chmod should probably have a host.  If it doesn't I am guessing it's not changing the permission on the mom, but only locally.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -14218,6 +14218,11 @@ def create_eatcpu_job(self, duration=None, mom=None):
         """
         Create a job that eats cpu indefinitely or for the given
         duration of time
+
+        :param duration: The duration, in seconds, to sleep
+        :type duration: int
+        :param mom: the MoM object on which to execute the job
+        :type mom: str
```
bhroam: My only other comment is that the mom parameter in the rest of this file means a mom object.  Here it is a host name for a mom.  This is existing code, so feel free to do what you want, but I'd either rename this 'host' or allow a mom or str.  Check for the instance of a mom object, and then call mom.shortname() on it.  If you do choose to keep it a string, I wouldn't call this a MoM object.  There is a mom object in this library.

ericpershey: I renamed it to hostname as the rest of the functions use hostname.  Just pushed.

vstumpf:
BTW, you don't have to keep the branch up-to-date, unless you (or a maintainer) thinks there might be an integration issue with newer code.

PR #1438, opened 2019-12-09
pbs_snapshot fails because file command not found

agrawalravi90:
#### Describe Bug or Feature

pbs_snapshot relies on the file command to be present on the system to find core files and capture traces from them. Some special systems might not have the file command, right now the command just fails, we should instead check to see if file command is present and only then use it.

Also, pbs_snapshot was not able to find the python and PTL libraries correctly and was not working properly, so I had to make necessary changes to the script, these are derived from the ptl.csh script

[snapptl.log](https://github.com/PBSPro/pbspro/files/3942233/snapptl.log)

pbs_snapshot output on a system with no file command:
[snapout.log](https://github.com/PBSPro/pbspro/files/3943188/snapout.log)

### DIFF
--- a/src/cmds/scripts/pbs_snapshot
+++ b/src/cmds/scripts/pbs_snapshot
@@ -35,20 +35,52 @@
 # "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
 # trademark licensing policies.
 
-pbs_conf=${PBS_CONF_FILE:-/etc/pbs.conf}
-if [ -r "${pbs_conf}" ]; then
-	. ${pbs_conf}
+ptllibpath=
+ptlbinpath=
+
+if [ -f "/etc/debian_version" ]; then
+        ptl_prefix_lib=$( dpkg -L pbspro-ptl 2>/dev/null | grep -m 1 lib$ 2>/dev/null )
 else
-	echo "***" >&2
-	echo "*** No PBS_CONF_FILE found, exiting" >&2
-	echo "***" >&2
-	exit 1
+        ptl_prefix_lib=$( rpm -ql pbspro-ptl 2>/dev/null | grep -m 1 lib$ 2>/dev/null )
+fi
+if [ "x${ptl_prefix_lib}" != "x" ]; then
+        python_dir=$( /bin/ls -1 ${ptl_prefix_lib} )
+        prefix=$( dirname ${ptl_prefix_lib} )
+
+        ptllibpath=${prefix}/lib/${python_dir}/site-packages
+        ptlbinpath=${prefix}/bin
+else
+        conf="${PBS_CONF_FILE:-/etc/pbs.conf}"
+        if [ -r "${conf}" ]; then
+                # we only need PBS_EXEC from pbs.conf
+                __PBS_EXEC=$( grep '^[[:space:]]*PBS_EXEC=' "$conf" | tail -1 | sed 's/^[[:space:]]*PBS_EXEC=\([^[:space:]]*\)[[:space:]]*/\1/' )
+                if [ "X${__PBS_EXEC}" != "X" ]; then
+                        # Define PATH and PYTHONPATH for the users
+                        PTL_PREFIX=$( dirname ${__PBS_EXEC} )/ptl
+                        python_dir=$( /bin/ls -1 ${PTL_PREFIX}/lib )/site-packages
+                        if [ -d "${PTL_PREFIX}/bin" ];then
+                            ptllibpath=${PTL_PREFIX}/lib/${python_dir}
+                        fi
+                        if [ -d "${PTL_PREFIX}/lib/${python_dir}" ];then
+                            ptlbinpath=${PTL_PREFIX}/bin
+                        fi
+                fi
+                unset __PBS_EXEC
+                unset PTL_PREFIX
+                unset conf
+                unset python_dir
+        fi
 fi
 
-export PYTHONPATH=${PBS_EXEC}/unsupported/fw:${PYTHONPATH}
+export PYTHONPATH=${ptllibpath}:${PYTHONPATH}
+
+if [ -z $ptlbinpath ];then
+    echo "Error: PTL bin path not found"
+    exit 1
+fi
 
 if [ -x "${PBS_EXEC}/python/bin/python" ]; then
-	${PBS_EXEC}/python/bin/python ${PBS_EXEC}/unsupported/fw/bin/pbs_snapshot.py "${@}"
+	${PBS_EXEC}/python/bin/python ${ptlbinpath}/pbs_snapshot "${@}"
 else
-	python3 ${PBS_EXEC}/unsupported/fw/bin/pbs_snapshot.py "${@}"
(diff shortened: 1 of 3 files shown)


test/fw/ptl/utils/pbs_snaputils.py
```diff
@@ -1137,8 +1138,15 @@ def __evaluate_core_file(self, file_path, core_dir):
             self.logger.debug("Could not find file path " + str(file_path))
             return False
 
+        filecmd = "file"
+        cmdpath = self.du.which(exe=filecmd)
+        # du.which returns the input cmd name if it can't find the cmd
+        if cmdpath is filecmd:
+            self.filecmdnotfound = True
+            return False
```
bhroam: Do this at the start of the process.  Print the error message then, and then rename this variable 'capture_core_files' and use it here.

src/cmds/scripts/pbs_snapshot
```diff
@@ -35,20 +35,52 @@
...
+ptllibpath=
+ptlbinpath=
+
+if [ -f "/etc/debian_version" ]; then
+        ptl_prefix_lib=$( dpkg -L pbspro-ptl 2>/dev/null | grep -m 1 lib$ 2>/dev/null )
+else
+        ptl_prefix_lib=$( rpm -ql pbspro-ptl 2>/dev/null | grep -m 1 lib$ 2>/dev/null )
+fi
+if [ "x${ptl_prefix_lib}" != "x" ]; then
+        python_dir=$( /bin/ls -1 ${ptl_prefix_lib} )
```
bhroam: Is it common practice to put spaces after and before your parentheses?

src/cmds/scripts/pbs_snapshot
```diff
@@ -35,20 +35,52 @@
...
-pbs_conf=${PBS_CONF_FILE:-/etc/pbs.conf}
-if [ -r "${pbs_conf}" ]; then
-	. ${pbs_conf}
+ptllibpath=
+ptlbinpath=
+
+if [ -f "/etc/debian_version" ]; then
+        ptl_prefix_lib=$( dpkg -L pbspro-ptl 2>/dev/null | grep -m 1 lib$ 2>/dev/null )
+else
+        ptl_prefix_lib=$( rpm -ql pbspro-ptl 2>/dev/null | grep -m 1 lib$ 2>/dev/null )
```
bhroam: This assumes the PTL rpm is installed.  What if we installed via pip like is in our OSS docs?  Shouldn't you also check /usr/lib/python3.X/?

src/cmds/scripts/pbs_snapshot
agrawalravi90: I actually just copied the code from ptl.csh, I'll remove the spaces

src/cmds/scripts/pbs_snapshot
```diff
@@ -35,20 +35,62 @@
...
+        ptllibpath=${prefix}/lib/${python_dir}/site-packages
+        ptlbinpath=${prefix}/bin
+else
+        conf="${PBS_CONF_FILE:-/etc/pbs.conf}"
+        if [ -r "${conf}" ]; then
+                # we only need PBS_EXEC from pbs.conf
+                __PBS_EXEC=$( grep '^[[:space:]]*PBS_EXEC=' "$conf" | tail -1 | sed 's/^[[:space:]]*PBS_EXEC=\([^[:space:]]*\)[[:space:]]*/\1/' )
+                if [ "X${__PBS_EXEC}" != "X" ]; then
+                        # Define PATH and PYTHONPATH for the users
+                        PTL_PREFIX=$(dirname ${__PBS_EXEC})/ptl
```
npadole20: I don't understand this part. Are we looking for 'ptl' in PBS_EXEC path? What I understood is we will come to this else block only if PTL rpm is not installed, right? From pbspro.spec, `%define ptl_prefix %{pbs_prefix}/../ptl`

npadole20: Ignore this comment. I understood it.

test/tests/functional/pbs_snapshot_unittest.py
```diff
@@ -933,12 +933,11 @@ def test_basic_option(self):
                     if not fpath.startswith(sched_priv_dir):
                         self.fail("Unexpected file " + fpath + " captured")
 
-    @classmethod
-    def tearDownClass(self):
+    def tearDown(self):
```
npadole20: Why did you change this?

agrawalravi90: Yes! Can you tell me how to get rid of that? I just assumed that we can no longer override tearDownClass() in PTL and was going to file a ticket for it.

npadole20: There is a ticket for it already. But I think it's okay if you keep these changes. As you explained, it will delete snapshots after every test case as opposed to at the end of the test suite, which is fine.

src/cmds/scripts/pbs_snapshot
```diff
@@ -38,14 +38,24 @@
...
 else
-        ptl_prefix_lib=$( rpm -ql pbspro-ptl 2>/dev/null | grep -m 1 lib$ 2>/dev/null )
+        ptl_prefix_lib=$(rpm -ql pbspro-ptl 2>/dev/null | grep -m 1 lib$ 2>/dev/null)
 fi
+
+# Try system paths
+if [ ! -d  ${ptl_prefix_lib} ]; then
+    ptl_prefix_lib=/usr/local/lib/python3.[[:digit:]]/site-packages
+    if [ ! -d  ${ptl_prefix_lib} ]; then
+        ptl_prefix_lib=/usr/lib/python3.[[:digit:]]/site-packages
```
bhroam: Can ls take [[:digit:]], I thought it can only take globs.

agrawalravi90: Seems to work:

[ravi@pbspro ~]$ ptl_prefix_lib=/usr/lib/python3.[[:digit:]]/site-packages
[ravi@pbspro ~]$ echo $ptl_prefix_lib
/usr/lib/python3.6/site-packages
[ravi@pbspro ~]$ ls $ptl_prefix_lib
easy_install.py  pip  pip-9.0.3.dist-info  pkg_resources  __pycache__  setuptools  setuptools-39.2.0.dist-info
[ravi@pbspro ~]$

bhroam: This is interesting.  I wouldn't have thought it would work, but from the looks of it, the regex is being evaluated at the assignment, not at ls.  When it hits ls, it is a path.

mike0042: Looks like it works for bash, but would probably fail in Bourne/Korn/C shell.

agrawalravi90: Thanks for your input Mike. It seems to work on Korn and Bourne as well, but it failed on C shell.

mike0042: Good to know. Thanks @agrawalravi90!

PR #1439, opened 2019-12-10
afslog should be done in forked process

vchlum:
#### Describe Bug or Feature
This bug was introduced in Kerberos support (#1281). The initial idea on how to deal with afslog turned out to be wrong. I thought I can manipulate the pag (process authentication group) of a process in order to provide afs tokens for the processes of a job. This is not true the pag number is not as decisive as I thought.

Right now, the afslog is done in the mom main process and the pag number is changed for a particular job task. The job always uses the same pag for each task and the afslog is done in the main process. This simplification is not possible. It can lead to losing afs tokens for a particular process or even for interchanging afs tokens between users! It is not reliable.

#### Describe Your Change
The proper afslog solution is provided: The storing of pag is removed and after the Kerberos credential cache is created and assigned to the job, we need to set a pag for each job task process. This pag is naturally inherited by the child precesses and they share the same pag. PBS starts a new child process with a new task as a child of the main mom process. Thus, we need to provide new pag for each job task. (also job process does not exist on sister mom - without a running task) After setting the pag, the afslog is done and the new child process is forked. This forked process naturally inherits the pag and thus it can do the afslog, which is its only purpose. If we do the renewal of the Kerberos credentials then afslog needs to be done again. At this moment a HUP signal is sent to the forked process and on this signal, the new process does the afslog. Once the job task ends, the TERM signal is sent to this process. 

The whole afs support is covered by macro HAVE_LIBKAFS or HAVE_LIBKOPENAFS - same as before.

Starting an interactive job and checking for the afs tokens: 
```
qsub -I
qsub: waiting for job 28511.torque1.grid.cesnet.cz to start
qsub: job 28511.torque1.grid.cesnet.cz ready

torque1.grid.cesnet.cz$ klist -T 
Credentials cache: FILE:/tmp/krb5cc_pbsjob_28511.torque1.grid.cesnet.cz
        Principal: vchlum@META

  Issued                Expires               Principal
Dec 10 12:29:27 2019  Dec 11 12:29:27 2019  krbtgt/META@META
Dec 10 12:29:27 2019  Dec 11 12:29:27 2019  afs/ics.muni.cz@META
Dec 10 12:29:27 2019  Dec 11 12:29:27 2019  krbtgt/ZCU.CZ@META
Dec 10 12:29:27 2019  Dec 11 04:29:27 2019  afs/zcu.cz@ZCU.CZ
Dec 10 12:29:27 2019  Dec 11 12:29:27 2019  krbtgt/RUK.CUNI.CZ@META

Dec 10 12:29:27 201  Dec 11 12:29:27 201  Tokens for ics.muni.cz
Dec 10 12:29:27 201  Dec 11 04:29:27 201  Tokens for zcu.cz
```
Starting a few of jobs of other users, waiting for renewing the credentials and checking the afs tokens again:
```
klist -T 
Credentials cache: FILE:/tmp/krb5cc_pbsjob_28511.torque1.grid.cesnet.cz
        Principal: vchlum@META

  Issued                Expires               Principal
Dec 10 12:37:09 2019  Dec 11 12:37:09 2019  krbtgt/META@META
Dec 10 12:37:09 2019  Dec 11 12:37:09 2019  afs/ics.muni.cz@META
Dec 10 12:37:09 2019  Dec 11 12:37:09 2019  krbtgt/ZCU.CZ@META
Dec 10 12:37:09 2019  Dec 11 04:37:09 2019  afs/zcu.cz@ZCU.CZ
Dec 10 12:37:09 2019  Dec 11 12:37:09 2019  krbtgt/RUK.CUNI.CZ@META

Dec 10 12:37:09 201  Dec 11 12:37:09 201  Tokens for ics.muni.cz
Dec 10 12:37:09 201  Dec 11 04:37:09 201  Tokens for zcu.cz
```

The afs tokens are still OK.

### DIFF
--- a/src/resmom/mom_main.c
+++ b/src/resmom/mom_main.c
@@ -6502,15 +6502,11 @@ kill_job(job *pjob, int sig)
 		ct += tsk_ct;
 
 #if defined(PBS_SECURITY) && (PBS_SECURITY == KRB5)
-#if defined(HAVE_LIBKAFS) || defined(HAVE_LIBKOPENAFS)
 		if (sig == SIGKILL) { /* only stop afslog when the task is finally dying */
-			if (signal_afslog(ptask, SIGTERM)) {
-				log_record(PBSEVENT_ERROR, PBS_EVENTCLASS_JOB, LOG_ERR,
-					ptask->ti_job->ji_qs.ji_jobid, "sending SIGTERM to afslog process failed");
-			}
+			AFSLOG_TERM(ptask);
 		}
 #endif
-#endif
+
 		/*
 		 ** If this is an orphan task, force it to be EXITED
 		 ** since it will not be seen by scan_for_terminated.
@@ -6533,12 +6529,7 @@ kill_job(job *pjob, int sig)
 			if (ptask->ti_qs.ti_parenttask == TM_NULL_TASK)
 				exiting_tasks = 1;
 #if defined(PBS_SECURITY) && (PBS_SECURITY == KRB5)
-#if defined(HAVE_LIBKAFS) || defined(HAVE_LIBKOPENAFS)
-			if (signal_afslog(ptask, SIGTERM)) {
-				log_record(PBSEVENT_ERROR, PBS_EVENTCLASS_JOB, LOG_ERR,
-					ptask->ti_job->ji_qs.ji_jobid, "sending SIGTERM to afslog process failed");
-			}
-#endif
+			AFSLOG_TERM(ptask);
 #endif
 		}
 	}
--- a/src/resmom/mom_main.c
+++ b/src/resmom/mom_main.c
@@ -6501,6 +6501,16 @@ kill_job(job *pjob, int sig)
 		tsk_ct = kill_task(ptask, sig, 0);
 		ct += tsk_ct;
 
+#if defined(PBS_SECURITY) && (PBS_SECURITY == KRB5)
+#if defined(HAVE_LIBKAFS) || defined(HAVE_LIBKOPENAFS)
+		if (sig == SIGKILL) { /* only stop afslog when the task is finally dying */
+			if (signal_afslog(ptask, SIGTERM)) {
+				log_record(PBSEVENT_ERROR, PBS_EVENTCLASS_JOB, LOG_ERR,
+					ptask->ti_job->ji_qs.ji_jobid, "sending SIGTERM to afslog process failed");
+			}
+		}
+#endif
+#endif
 		/*
 		 ** If this is an orphan task, force it to be EXITED
 		 ** since it will not be seen by scan_for_terminated.
@@ -6522,6 +6532,14 @@ kill_job(job *pjob, int sig)
 			 */
 			if (ptask->ti_qs.ti_parenttask == TM_NULL_TASK)
 				exiting_tasks = 1;
+#if defined(PBS_SECURITY) && (PBS_SECURITY == KRB5)
+#if defined(HAVE_LIBKAFS) || defined(HAVE_LIBKOPENAFS)
+			if (signal_afslog(ptask, SIGTERM)) {
+				log_record(PBSEVENT_ERROR, PBS_EVENTCLASS_JOB, LOG_ERR,
+					ptask->ti_job->ji_qs.ji_jobid, "sending SIGTERM to afslog process failed");
+			}
+#endif
+#endif
 		}
 	}
 
--- a/src/resmom/renew_creds.c
+++ b/src/resmom/renew_creds.c
@@ -128,7 +128,8 @@ extern int decode_block_base64(unsigned char *ascii_data, ssize_t ascii_len, uns
 static int get_job_info_from_job(const job *pjob, const task *ptask, eexec_job_info job_info);
 static int get_job_info_from_principal(const char *principal, const char* jobid, eexec_job_info job_info);
 static krb5_error_code get_ticket_from_storage(struct krb_holder *ticket, char *errbuf, size_t errbufsz);
-static krb5_error_code get_renewed_creds(struct krb_holder *ticket, char *errbuf, size_t errbufsz, int cred_action);
+static krb5_error_code get_ticket_from_ccache(struct krb_holder *ticket, char *errbuf, size_t errbufsz);
+static krb5_error_code get_renewed_creds(struct krb_holder *ticket, char *errbuf, size_t errbufsz);
 static int init_ticket(struct krb_holder *ticket, int cred_action);
 
 static svrcred_data *find_cred_data_by_jobid(char *jobid);
@@ -157,12 +158,6 @@ init_ticket_from_req(char *principal, char *jobid, struct krb_holder *ticket, in
 		log_err(errno, __func__, buf);
 		return ret;
 	}
-
-#if defined(HAVE_LIBKAFS) || defined(HAVE_LIBKOPENAFS)
-	if (cred_action != CRED_DESTROY) {
-		setpag(0);
-	}
-#endif
 
 	ret = init_ticket(ticket, cred_action);
 	if (ret == 0) {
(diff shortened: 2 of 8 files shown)


mike0042:
Please don't use C++ style comments, use /* */ instead.

I realize AFS relies on Kerberos for authentication (I supported DCE/DFS with NQS a very long time ago), but I don't understand why there are function names that reference AFS. PBS should be agnostic to any service that relies on Kerberos, filesystem or otherwise. Could you please explain why AFS is specifically mentioned?

vchlum:
@mike0042 The reason why to have this optional feature in PBS is that the renewing of Kerberos credentials requests the new afslog to be called no matter the afslog was called before the renew.

The afs functions in PBS provide the user-friendly possibility of using afs filesystem within jobs. It would be also possible to simply instruct the user that he/she should call afslog command in order to have the afs available within the job. But since the credentials are renewed by PBS,  the user can not know that the credentials ware renewed and the afslog needs to be run again. It means that the user would need to call afslog command every time he/she wants to use afs within the job, which is not good.

You can still build PBS without afs support. The afs functions (optionally and naturally) extend the possibility of using Kerberos within PBS.

We use the afs filesystem and I would like to have the support of it in PBS. I think this is the right approach on how to use afs with PBS.

mike0042:
Thank you for the explanation @vchlum. I don't recall having to do anything like that for DFS, but that was long ago. In addition, I see you were calling krb5_afslog() even before these changes so you're really not extending the design, just making life easier for the user. No arguments from me.

vchlum:
There was a conflict - so I resolved it. Just a gentle reminder of this PR... it is active PR.

src/resmom/mom_main.c
```diff
@@ -6522,6 +6532,14 @@ kill_job(job *pjob, int sig)
...
 			if (ptask->ti_qs.ti_parenttask == TM_NULL_TASK)
 				exiting_tasks = 1;
+#if defined(PBS_SECURITY) && (PBS_SECURITY == KRB5)
+#if defined(HAVE_LIBKAFS) || defined(HAVE_LIBKOPENAFS)
+			if (signal_afslog(ptask, SIGTERM)) {
+				log_record(PBSEVENT_ERROR, PBS_EVENTCLASS_JOB, LOG_ERR,
+					ptask->ti_job->ji_qs.ji_jobid, "sending SIGTERM to afslog process failed");
+			}
+#endif
+#endif
```
mike0042: This section of code is repeated at least three times. Please consider making it a macro or an inline function.

src/resmom/renew_creds.c
```diff
@@ -197,15 +192,51 @@ init_ticket_from_job(job *pjob, const task *ptask, struct krb_holder *ticket, in
...
+ * @retval	!= 0 on error
+ */
+int
+init_ticket_from_ccache(job *pjob, const task *ptask, struct krb_holder *ticket)
+{
+	int ret;
+	char buf[LOG_BUF_SIZE];
+
+	if ((ret = get_job_info_from_job(pjob, ptask, ticket->job_info)) != 0)  {
+		snprintf(buf, sizeof(buf), "Could not fetch GSSAPI information from job (get_job_info_from_job returned %d).",ret);
```
mike0042: Please add a space after the last comma on this line.

src/resmom/renew_creds.c
```diff
@@ -1093,153 +1185,357 @@ send_cred_sisters(job *pjob)
...
-	gid_t *grplist = NULL;
-	int i;
-	int numsup;
-	static int maxgroups = 0;
-	int32_t pag = 0;
+	krb5_error_code ret = 0;
 
-	if (k_hasafs() == 0)
-		return 0;
+	if(k_hasafs() && (ret = krb5_afslog(context, job_info->ccache, NULL, NULL)) != 0) {
```
mike0042: Please add a space after "if"

src/resmom/renew_creds.c
```diff
@@ -1093,153 +1185,357 @@ send_cred_sisters(job *pjob)
...
+	return(ret);
+}
 
-	numsup = getgroups(maxgroups, grplist);
+/**
+ * @brief
+ * 	singleshot_afslog - do imminent afslog, no process started
+ *
+ * @param[in] ticket - kerberos ticket
+ *
```
mike0042: Please add "@return void" for all functions returning void.

src/resmom/renew_creds.c
```diff
@@ -1093,153 +1185,357 @@ send_cred_sisters(job *pjob)
...
+ */
+void
+singleshot_afslog(struct krb_holder *ticket) {
+	if(k_hasafs()) {
```
mike0042: Please add space after "if"

src/resmom/renew_creds.c
```diff
@@ -1093,153 +1185,357 @@ send_cred_sisters(job *pjob)
...
+ *
+ * @param[in] signal - received signal
+ *
+ */
+static void
+do_afslog_on_signal(int signal) {
+	krb5_error_code ret;
+
+	if (signal == SIGHUP) {
+		if ((ret = do_afslog(afslog_ticket->context, afslog_ticket->job_info))) {
```
mike0042: You don't use ret for anything, so why is it being set?

src/resmom/renew_creds.c
```diff
@@ -1093,153 +1185,357 @@ send_cred_sisters(job *pjob)
...
-	static int   maxgroups = 0;
+int
+start_afslog(const task *ptask, struct krb_holder *ticket, int fd1, int fd2) {
+	char buf[LOG_BUF_SIZE * 2] ;
+	char errbuf[LOG_BUF_SIZE];
+	int ret = PBS_KRB5_OK;
+	char pid_file[MAXPATHLEN];
+	int local_ticket = 0;
+
+	if(!k_hasafs())
```
mike0042: Please add space after "if"

src/resmom/renew_creds.c
```diff
@@ -1093,153 +1185,357 @@ send_cred_sisters(job *pjob)
...
 
-	if (k_hasafs() == 0)
-		return;
+	if (ptask == NULL)
+		return PBS_KRB5_ERR_INTERNAL;
 
-	if (pag == 0) {
-		k_setpag();
-		return;
+	job *pjob = ptask->ti_job;
```
mike0042: pjob should be declared at the top of the function and assigned here.

src/resmom/renew_creds.c
```diff
@@ -1093,153 +1185,357 @@ send_cred_sisters(job *pjob)
...
+
+	if (*pjob->ji_qs.ji_fileprefix != '\0')
+		snprintf(pid_file, sizeof(pid_file), "%s%s_afslog_%8.8X.pid", path_jobs, pjob->ji_qs.ji_fileprefix, (unsigned int)ptask->ti_qs.ti_task);
+	else
+		snprintf(pid_file, sizeof(pid_file), "%s%s_afslog_%8.8X.pid", path_jobs, pjob->ji_qs.ji_jobid, (unsigned int)ptask->ti_qs.ti_task);
+
+	int fd = open(pid_file, O_CREAT|O_EXCL|O_WRONLY, 0600);
```
mike0042: Please declare new variables at the beginning of the code block, be it a function or otherwise.

src/resmom/renew_creds.c
```diff
@@ -1093,153 +1185,357 @@ send_cred_sisters(job *pjob)
...
+		/* another afslog process is running ? */
+		snprintf(buf, sizeof(buf), "opening PID file for afslog process (%s) uid = %d", pid_file, getuid());
+		log_err(errno, __func__, buf);
+		return PBS_KRB5_ERR_CANT_OPEN_FILE;
 	}
 
-	/* first remove any other pag - for sure */
-	removepag();
+	/* Go user */
+	if(seteuid(pjob->ji_qs.ji_un.ji_momt.ji_exuid) < 0) {
```
mike0042: Please add a space after "if"

src/resmom/renew_creds.c
```diff
@@ -1093,153 +1185,357 @@ send_cred_sisters(job *pjob)
...
+/**
+ * @brief
+ * 	getpag - recognize afs pag in groups and return the pag.
+ *
+ * @return 	int32_t
+ * @retval	pag > 0 on success
+ * @retval	0 otherwise
+ */
+int32_t
+getpag()
```
mike0042: I'm not certain if you have removed all references to getpag outside of this file. If so, it should become static and removed from the corresponding header file.

vchlum: It is also used in becomeuser_args() because initgroups() removes pag here and we need to save it.

vchlum:
Thank you @mike0042 for the review. I addressed your comments.

src/resmom/renew_creds.h
```diff
@@ -97,9 +97,15 @@ int im_cred_send(job *pjob, hnodent *xp, int stream);
...
+int start_afslog(const task *ptask, struct krb_holder *ticket, int, int);
+int signal_afslog(const task *ptask, int signal);
+
 int32_t getpag();
-void setpag(int32_t pag);
-void removepag();
+
+#define AFSLOG_TERM(x) {if (signal_afslog(x, SIGTERM)) log_record(PBSEVENT_ERROR, PBS_EVENTCLASS_JOB, LOG_ERR, x->ti_job->ji_qs.ji_jobid, "sending SIGTERM to afslog process failed");}
+#else
+#define AFSLOG_TERM(x) {x = x;}
```
mike0042: Is the intention to implement a NOP here? If so, {} should suffice. Don't think you need to perform a reflexive assignment.

vchlum: I was afraid of using {} because of some compiler versions. You are right. From my googling, {} looks good. Changed.

vchlum:
I fixed a bug. Please, see @mike0042. The forked process kept the logfile open and thus running jobs prevent the log rotation. From now on, the afslog process logs only into syslog.

vchlum:
Found and fixed a bug @mike0042. I tried to do the afslog with root privileges for singleshot_afslog, which is part of req_cpyfile(). Other afslogs are OK.

src/resmom/renew_creds.c
```diff
@@ -1234,9 +1234,28 @@ do_afslog(krb5_context context, eexec_job_info job_info)
  */
 void
 singleshot_afslog(struct krb_holder *ticket) {
+	char buf[LOG_BUF_SIZE * 2];
+	char errbuf[LOG_BUF_SIZE];
+
 	if (k_hasafs()) {
+		/* Go user */
+		if(seteuid(ticket->job_info->job_uid) < 0) {
```
mike0042: Please add a space after "if"

src/resmom/renew_creds.c
```diff
@@ -1234,9 +1234,28 @@ do_afslog(krb5_context context, eexec_job_info job_info)
...
+			log_err(errno, __func__, buf);
+
+			return;
+		}
+
 		k_setpag();
 		do_afslog(ticket->context, ticket->job_info);
+
+		/* Go root */
+		if(seteuid(0) < 0) {
```
mike0042: Please add a space after "if"

vchlum:
@mike0042 I found multiple related occurrences without space after 'if'. Fixed.

vchlum:
Please, could I get one more review? As mentioned in #1505, these routines will be moved into lib, but this PR is related to only part of the planned refactoring and I think it is useful to merge the changes before the work on the refactoring of renewing code will begin.

PR #1440, opened 2019-12-10
test_restart_server fails because the sched_host name doesn't match

birbalsain:
#### Describe Bug or Feature

On some Platform PTL consider different sched hostname then hostname get by the pbs.
i. e PBS takes hostname as server_host and ptl takes hostname as pbs_host
#### Describe Your Change
In mullti sched doesn't require to match sched hostname.

### DIFF
--- a/test/tests/functional/pbs_multi_sched.py
+++ b/test/tests/functional/pbs_multi_sched.py
@@ -601,7 +601,6 @@ class TestMultipleSchedulers(TestFunctional):
         sched_logs = os.path.join(
             self.server.pbs_conf['PBS_HOME'], 'sched_logs_sc1')
         a = {'sched_port': 15050,
-             'sched_host': self.server.hostname,
              'sched_priv': sched_priv,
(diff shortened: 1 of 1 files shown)


birbalsain:
@agrawalravi90 i updated the bug summary

agrawalravi90:
I'm sorry but I'm not sure that I understand your new description. But if the problem happens on a particular platform, then maybe this is not a good test to execute for that platform and maybe we should skip it instead of removing the sched host attr form the list, which might be a useful thing to test in general.

birbalsain:
> the

@agrawalravi90 it sounds good.

agrawalravi90:
@birbalsain any update on this? It seemed like you agreed with my opinion, but I don't see any new commits, so just wanted to make sure that we are on the same page.

agrawalravi90:
@birbalsain I'm closing this PR because it's been inactive for more than 2 months, please re-open when you are ready with your changes.

PR #1444, opened 2019-12-11
Scheduler preempts even when preemptee changes priority

arungrover:
#### Describe Bug or Feature
This PR is a combination of the following problems - 

PBS Scheduler could preempt a job that could result in escalating the preemption priority of the preempted job or other jobs that the scheduler could possibly preempt. This is wrong because there is a potential that this escalated priority job could possibly preempt its preemptor.
This problem only happens when the jobs are just on the edge of being over the soft limits and preempting them would release resources and bring them below the soft limits.

There is another problem in soft limits code where scheduler was updating preempt bit and preempt_status of all the relevant jobs in the call to update_server_on_end but it was updating the entities in update_universe_on_end after calling update_server_on_end. This seems wrong because jobs were getting updated before the entities were updated that means preempt bit was not changing on any of the jobs. I have removed the functionality to update preempt bits on the job from update_server_on_end and now update_universe_on_end calls a function to update the preemption priorities on relevant jobs (after the entities are updated).

Check_max_user/group/project_res_soft functions were not resetting the soft limit preempt bit when the entities move under the said limit. 

Check_max_user/group/project_res_soft functions were returning right as soon as they were hitting a resource that was over its soft limits. This is wrong because there could be more than one resource on which limit might be set and it needs to update (possibly reset) the preempt bit for other resources too.

This PR also changes the order in which arguments are passed to function find_resource_resv_by_indrank to match the function name (i.e. index is passed before rank to match indrank)

#### Describe Your Change
There has been a slight change in how the scheduler finds preempt-able candidates to run a high priority job.
- When it ends the lower priority job in the simulated universe, it checks if the preemption priority of the preempted job or any of the previously (in the simulated universe) preempted jobs is equal to or more than the higher priority job (remember ending a job in the simulated universe changes preemption priority of all relevant jobs).
- If it finds that there is a job with equal or higher priority as the high priority job then it resumes back the most recent preempted job (because that is the job that caused the priority to change) and moves on.

[test_after_fix.txt](https://github.com/PBSPro/pbspro/files/3953127/test_after_fix.txt)
[test_before_fix.txt](https://github.com/PBSPro/pbspro/files/3947807/test_before_fix.txt)

### DIFF
--- a/src/scheduler/job_info.c
+++ b/src/scheduler/job_info.c
@@ -2783,7 +2783,7 @@ dup_job_info(job_info *ojinfo, queue_info *nqinfo, server_info *nsinfo)
 
 	if (ojinfo->resv != NULL) {
 		njinfo->resv = find_resource_resv_by_indrank(nqinfo->server->resvs,
-			ojinfo->resv->rank, ojinfo->resv->resresv_ind);
+			ojinfo->resv->resresv_ind, ojinfo->resv->rank);
 	}
 
 	njinfo->resused = dup_resource_req_list(ojinfo->resused);
@@ -3020,7 +3020,7 @@ find_and_preempt_jobs(status *policy, int pbs_sd, resource_resv *hjob, server_in
 		}
 
 		for (i = 0; i < no_of_jobs; i++) {
-			job = find_resource_resv_by_indrank(sinfo->running_jobs, jobs[i], -1);
+			job = find_resource_resv_by_indrank(sinfo->running_jobs, -1, jobs[i]);
 			if (job != NULL) {
 				if ((preempt_jobs_list[i] = strdup(job->name)) == NULL) {
 					log_err(errno, __func__, MEM_ERR_MSG);
@@ -3121,7 +3121,7 @@ find_and_preempt_jobs(status *policy, int pbs_sd, resource_resv *hjob, server_in
 			log_event(PBSEVENT_DEBUG, PBS_EVENTCLASS_JOB, LOG_DEBUG, hjob->name,
 				"Preempted work didn't run job - rerun it");
 			for (i = 0; i < preempted_count; i++) {
-				job = find_resource_resv_by_indrank(sinfo->jobs, preempted_list[i], -1);
+				job = find_resource_resv_by_indrank(sinfo->jobs, -1, preempted_list[i]);
 				if (job != NULL && !job->job->is_running) {
 					clear_schd_error(serr);
 					if (run_update_resresv(policy, pbs_sd, sinfo, job->job->queue, job, NULL, RURR_NO_FLAGS, serr) == 0) {
@@ -3300,6 +3300,7 @@ find_jobs_to_preempt(status *policy, resource_resv *hjob, server_info *sinfo, in
 		free_schd_error_list(full_err);
 		return NULL;
 	}
+	pjobs[0] = NULL;
 
 	if (sinfo->preempt_targets_enable) {
 		preempt_targets_req = find_resource_req(hjob->resreq, getallres(RES_PREEMPT_TARGETS));
@@ -3326,7 +3327,7 @@ find_jobs_to_preempt(status *policy, resource_resv *hjob, server_info *sinfo, in
 		return NULL;
 	}
 	npolicy = nsinfo->policy;
-	nhjob = find_resource_resv_by_indrank(nsinfo->jobs, hjob->rank, hjob->resresv_ind);
+	nhjob = find_resource_resv_by_indrank(nsinfo->jobs, hjob->resresv_ind, hjob->rank);
 	prev_prio = nhjob->job->preempt;
 
 	if (nsinfo->preempt_targets_enable) {
@@ -3401,6 +3402,8 @@ find_jobs_to_preempt(status *policy, resource_resv *hjob, server_info *sinfo, in
 	skipto = 0;
 	while ((indexfound = select_index_to_preempt(npolicy, nhjob, rjobs_subset, skipto, err, fail_list)) != NO_JOB_FOUND) {
 		struct preempt_ordering *po;
+		int dont_preempt_job = 0;
+		int ind = 0;
 
 		if (indexfound == ERR_IN_SELECT) {
 			/* System error occurred, no need to proceed */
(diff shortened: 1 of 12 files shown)


test/tests/functional/pbs_preemption.py
```diff
@@ -556,3 +556,131 @@ def test_insufficient_server_rassn_select_resc(self):
...
+                # normal and scheduler will not attempt to preempt it.
+                a = {ATTR_l + '.select=1:ncpus': 1}
+            j = Job(attrs=a)
+            jid = self.server.submit(j)
+            jid_list.append(jid)
+            time.sleep(1)
+
+        # Submit 5th job that will make all the job in workq to go over its
+        # softlimits because if resource ncpus
+        time.sleep(1)
```
bhroam: You will have just done a sleep 1.  Do you need to do another?

test/tests/functional/pbs_preemption.py
```diff
@@ -556,3 +556,131 @@ def test_insufficient_server_rassn_select_resc(self):
...
+
+        a = {'max_run_res_soft.ncpus': "[u:PBS_GENERIC=5]"}
+        self.server.manager(MGR_CMD_SET, QUEUE, a, 'workq')
+        # Set a soft limit on resource foo to 0 so that all jobs requesting
+        # this resource are over soft limits.
+        a = {'max_run_res_soft.foo': "[u:PBS_GENERIC=0]"}
+        self.server.manager(MGR_CMD_SET, QUEUE, a, 'workq')
+        a = {'max_run_res_soft.ncpus': "[u:PBS_GENERIC=10]"}
+        self.server.manager(MGR_CMD_SET, QUEUE, a, 'workq2')
+
```
bhroam: Why set a limit on workq2?  You submit 1 job requesting 8 cpus.  It'll never cross its limit.

src/scheduler/limits.c
```diff
@@ -2965,11 +2971,15 @@ check_max_user_res_soft(resource_resv **rr_arr, resource_resv *rr,
 		} else if (max_genuser_res_soft < used) {
 			if (rescts != NULL)
 				rescts->soft_limit_preempt_bit = preempt_bit;
-			return (preempt_bit);
+			rc = preempt_bit;
+		} else {
+			/* usage is under generic user soft limit, reset the preempt bit */
+			if (rescts != NULL)
+				rescts->soft_limit_preempt_bit = 0;
```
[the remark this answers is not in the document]
arungrover: In case of a specific limit, wouldn't control fall into the block - [L2961 ](https://github.com/PBSPro/pbspro/pull/1444/files/0e83e409f0e2730568309770e3718616086eb5f4#diff-4a1b68c3622f68d77b25310418c0e602R2961-R2970). Since we already check for specific limit, I thought this was the right place to put it.

test/tests/functional/pbs_preemption.py
```diff
@@ -556,3 +556,131 @@ def test_insufficient_server_rassn_select_resc(self):
...
+        # Submit 5th job that will make all the job in workq to go over its
+        # softlimits
+        a = {ATTR_l + '.select=1:ncpus=1:foo': 4}
+        j = Job(attrs=a)
+        jid = self.server.submit(j)
+        jid_list.append(jid)
+        self.server.expect(JOB, {'job_state=R': 5})
+
+        # Submit a job in workq2 which requests for 4 ncpus and 3 foo resource
+        a = {ATTR_l + '.select=1:ncpus=4:foo': 3, ATTR_q: 'workq2'}
```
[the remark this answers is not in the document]
arungrover: I started writing the test with 2 resources in mind and it over the time morphed to use only one. I'l remove foo.

test/tests/functional/pbs_preemption.py
arungrover: You are right, that's not needed.

arungrover:
@bhroam I've addressed your review comments. Please have a look.

src/scheduler/job_info.c
```diff
@@ -3426,11 +3429,42 @@ find_jobs_to_preempt(status *policy, resource_resv *hjob, server_info *sinfo, in
 
 		update_universe_on_end(npolicy, pjob,  "S", NO_ALLPART);
 		rjobs_count--;
+		/* Check if any of the previously preempted job increased its preemption priority */
+		for (ind = 0; pjobs[ind] != NULL; ind++) {
+			if (pjobs[ind]->job->preempt >= nhjob->job->preempt) {
+				dont_preempt_job = 1;
```
agrawalravi90: not sure that I understand this, why don't we want to preempt 'pjob' if any of the previously preempted jobs increased their p-priority? What has other jobs' p-priority got to do with pjob?

src/scheduler/job_info.c
```diff
@@ -3426,11 +3429,42 @@ find_jobs_to_preempt(status *policy, resource_resv *hjob, server_info *sinfo, in
...
+				break;
+			}
+		}
+		/* Check if the job we just ended increases its priority. If so, don't preempt this job */
```
agrawalravi90: So, this if statement isn't really checking that pjob's p-priority increased, it's checking whether its p-priority increased more than high priority job's, right? If so, can you please update the comment to say that instead?

src/scheduler/job_info.c
```diff
@@ -3426,11 +3429,42 @@ find_jobs_to_preempt(status *policy, resource_resv *hjob, server_info *sinfo, in
 
 		update_universe_on_end(npolicy, pjob,  "S", NO_ALLPART);
 		rjobs_count--;
+		/* Check if any of the previously preempted job increased its preemption priority */
+		for (ind = 0; pjobs[ind] != NULL; ind++) {
```
agrawalravi90: When we find a preemption candidate inside select_index_to_preempt(), is there a way to check how much its preempt priority will increase if it got preempted? If we could determine that, then we can just not select such a job inside select_index_to_preempt() and we wouldn't need to check stuff here.

src/scheduler/job_info.c
```diff
@@ -3426,11 +3429,42 @@ find_jobs_to_preempt(status *policy, resource_resv *hjob, server_info *sinfo, in
...
+				break;
+			}
+		}
+		/* Check if the job we just ended increases its priority. If so, don't preempt this job */
+		if (dont_preempt_job || pjob->job->preempt >= nhjob->job->preempt) {
+			schd_error *tmp_err = new_schd_error();
+			clear_schd_error(tmp_err);
+			remove_resresv_from_array(rjobs_subset, pjob);
+			log_event(PBSEVENT_DEBUG2, PBS_EVENTCLASS_JOB, LOG_INFO, pjob->name,
+				  "Preempting job will escalate its priority or priority of other jobs, not preempting it");
```
agrawalravi90: Sorry for the naive question :) but why would preempting one job increase the preemption priority of some other job?

src/scheduler/job_info.c
```diff
@@ -3426,11 +3429,42 @@ find_jobs_to_preempt(status *policy, resource_resv *hjob, server_info *sinfo, in
...
+			if (sim_run_update_resresv(npolicy, pjob, ns_arr, NO_ALLPART) != 1) {
+				log_event(PBSEVENT_DEBUG2, PBS_EVENTCLASS_JOB, LOG_INFO, nhjob->name,
+					  "Trouble finding preemptable candidates");
+				free_schd_error_list(full_err);
```
agrawalravi90: you do these series of free's earlier as well, how about putting them under a goto label at the end of the function?

src/scheduler/limits.c
```diff
@@ -2803,11 +2804,15 @@ check_max_group_res_soft(resource_resv *rr, counts *cts_list, void *limitctx, in
 		} else if (max_gengroup_res_soft < used) {
 			if (rescts != NULL)
 				rescts->soft_limit_preempt_bit = preempt_bit;
-			return (preempt_bit);
+			rc = preempt_bit;
+		} else {
```
agrawalravi90: just say "else if (rescts != NULL) "?

src/scheduler/job_info.c
arungrover: We don't want to preempt pjob because if we preempt this job then it increases the preemption priority of one of the previously preempted job. For example, jobs j1 runs with 8 ncpus and it is over the limit by 12 ncpus. So preempting J1 would still mean that the job is over its soft limit. Now comes job j2 which is consuming 6 ncpus and preempting this job will mean that previously preempted job J1 would now be under the softlimit and change its preemption priority.
Now we don't want to rollback the calculations we have done to preempt J1 and jobs after that so the easiest thing to do is to rollback J2 and not preempt it.

bhroam: Now that I look at this again, I think you need to change '>=' to just '>'.  It isn't a problem that we preempt a job and the preemptee is equal to the high priority job after it is preempted.  Jobs at equal preemption priority can't preempt each other.  It's also wrong to allow an entity to become immune to preemption just because they are over their limit by one.  Until they are at their limit, they are preemptable.

src/scheduler/job_info.c
arungrover: We can add that check but that check will be kind of similar to what we have right now. We will have to end the job in a simulated universe and then check its preemption priority. 
Even if we have this check, We can not predict if its preemption priority will move if any of the other running jobs get preempted.

src/scheduler/job_info.c
arungrover: One of the comment above covers this with an example.

src/scheduler/limits.c
arungrover: I purposefully did this because I wanted to have this else for the case where generic limit is under the usage. Within that case when resource counts are not null I rest the preempt_bit.
What you are suggesting is correct but for better readability I kept it the way it is, If you feel strongly about it, I'll change it.

arungrover:
[test_after_fix.txt](https://github.com/PBSPro/pbspro/files/3981087/test_after_fix.txt)

src/scheduler/job_info.c
```diff
@@ -3429,29 +3415,25 @@ find_jobs_to_preempt(status *policy, resource_resv *hjob, server_info *sinfo, in
 
 		update_universe_on_end(npolicy, pjob,  "S", NO_ALLPART);
 		rjobs_count--;
-		/* Check if any of the previously preempted job increased its preemption priority */
+		/* Check if any of the previously preempted job increased its preemption priority to be more than the
+		 * high priority job */
```
bhroam: Move the \*/ to the line below to follow our multi-line comment style.

PR #1445, opened 2019-12-11
Fix NoneType error when server hooks receive alarm

minghui-liu:
#### Describe Bug or Feature

Runjob hook fails with the following error when it receives hook alarm

> PBS server internal error (15011) in Error evaluating Python script, 'NoneType' object is not callable

This is caused by a behavior change in Python 3. Unlike Python 2, Python 3 interpreter converts a signal received to a keyboard interrupt automatically. So the hook scripts receive two interrupts. 

#### Describe Your Change

Since python 3 sends a keyboard interrupt to the hook script by default, we no longer need to call `PyErr_SetInterrup()` in `pbs_python_set_interrupt`. 

Manual test result on centos 7: 
[after_fix_centos_7.txt](https://github.com/PBSPro/pbspro/files/3952126/after_fix_centos_7.txt)

### DIFF
--- a/src/lib/Libpython/pbs_python_svr_external.c
+++ b/src/lib/Libpython/pbs_python_svr_external.c
@@ -584,10 +584,7 @@ pbs_python_do_vnode_set(void)
 void
 pbs_python_set_interrupt(void)
 {
-
-#ifdef PYTHON
-	PyErr_SetInterrupt();
-#endif
+	return;
 }
 
 /**


src/lib/Libpython/pbs_python_svr_external.c
```diff
@@ -584,10 +584,7 @@ pbs_python_do_vnode_set(void)
 void
 pbs_python_set_interrupt(void)
 {
-
-#ifdef PYTHON
-	PyErr_SetInterrupt();
-#endif
+	return;
```
vstumpf: Without PyErr_SetInterrupt(), the signal never gets sent to the python interpreter.

minghui-liu: Yes you are right. Removing this does not fix the problem. I'll close this pull request.

PR #1448, opened 2019-12-12
Fix mom crash due to release_nodes_stageout=true

bayucan:
#### Describe Bug or Feature

* MS mom crashed while updating a job's resources_used info, obtained from a sister mom that has been released early from the job (i.e. release_nodes_stageout=true).
* For some reason, a sister mom reported resources_used info tagged as FQDN nodehost earlier in the job process, but later tagged in the form of short hostname nodehost. This caused MS mom to allocate an extra pjob-ji.resources[] entry for the different nodehost name, but the new pjob->ji_resources[x].nr_used entry was not properly initialized. This caused the crash in update_ajob_status_using_cmd().
* When a mom starts up, it uses as nodehost to report back to MS mom the mom_host value. The mom_host value is fairly static through the life of a mom, either getting the value from gethostbyname() or from the PBS_MOM_NODE_NAME. Perhaps while job is running, the sister mom could have been restarted and a different form of mom_host is used.

#### Describe Your Change

* In mom_comm.c:im_request() under IM_SEND_RESC, ensure if new entry to pjob->ji_resources[] is added, the pjob->ji_resources[].nr_used should be initialized as well.
* Whether or not the resources_used data is tagged under FQDN mom hostname or short mom hostname, both data are for the same Mom. Thus, compare_short_hostname() is done instead of strcmp() when deciding if a new entry to pjob->ji_resources[] is needed.

* Verification of this bug is manual, as writing a PTL test is tricky given that it requires from the time a job is ending, and before stageout kicks in, the mom_host value of a sister mom would need to change in form.  To reproduce the bug, a special mom was built and run on the sister node, as detailed below:
[bug_verify.txt](https://github.com/PBSPro/pbspro/files/3957251/bug_verify.txt)
* Valgrind log when the fix is executed:
[valgrind.mom.txt](https://github.com/PBSPro/pbspro/files/3957260/valgrind.mom.txt)

### DIFF
--- a/src/resmom/mom_comm.c
+++ b/src/resmom/mom_comm.c
@@ -5409,7 +5409,7 @@ join_err:
 			resc_idx = -1;
 			for (i=0; i < pjob->ji_numrescs; i++) {
 				if ((pjob->ji_resources[i].nodehost != NULL) &&
-				    (strcmp(pjob->ji_resources[i].nodehost, nodehost) == 0)) {
+				    (compare_short_hostname(pjob->ji_resources[i].nodehost, nodehost) == 0)) {
 					resc_idx = i;
 					break;
 				}
@@ -5439,6 +5439,8 @@ join_err:
 				 	 	"strdup failure setting nodehost");
 					goto err;
 				}
+				clear_attr(&pjob->ji_resources[resc_idx].nr_used,
+						&job_attr_def[JOB_ATR_resc_used]);
 				pjob->ji_numrescs++;
 
 			}
--- a/src/include/net_connect.h
+++ b/src/include/net_connect.h
@@ -188,6 +188,7 @@ void close_conn(int socket);
 pbs_net_t get_connectaddr(int sock);
 int  get_connecthost(int sock, char *namebuf, int size);
 pbs_net_t get_hostaddr(char *hostname);
+int  compare_short_hostname(char *shost, char *lhost);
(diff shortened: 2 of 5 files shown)


src/resmom/mom_comm.c
```diff
@@ -5439,6 +5439,8 @@ im_request(int stream, int version)
 				 	 	"strdup failure setting nodehost");
 					goto err;
 				}
+				clear_attr(&pjob->ji_resources[resc_idx].nr_used,
+						&job_attr_def[JOB_ATR_resc_used]);
```
mike0042: Do you need to check nr_used.flags for ATR_VFLAG_SET? Is there the potential to leak memory here?

bayucan: No because this is the part in the code where we’re initializing/filling in data for a new slot: pjob->ji_resources[resc_idx].
It has not been set yet.

vstumpf:
I have a question, would this cause issues if the mom was created and named with an ip? If we're checking everything until the first period, wouldnt 10.1.2.3 match with 10.2.3.4?

bayucan:
@vstumpf : yes, it would be a problem if mom hosts are in IP address form only and added to PBS in that form. The node rampdown feature in PBS would be affected when it comes to reporting of resources_used, when nodes are removed early from the job. Info on the first sister node would always get the info since the match before the dot. I will send out an updated fix.

PR #1449, opened 2019-12-13
Adjustment to fix 4f7bed8 (mom crash due release_nodes_stageout=true)

bayucan:
#### Describe Bug or Feature

* There's a slight problem to the fix to "mom crash due release_nodes_stageout=true)" in that if mom hosts are added to the PBS as IP addresses, the compare_short_hostname() would always be true.

PR#1448: https://github.com/PBSPro/pbspro/pull/1448

#### Describe Your Change

* Adjust compare_short_hostname() function to check if at least one of the given hostnames is in IPV4 dotted-decimal form. If so, then just do a strcasecmp().

* WIth the updated fix, reran the manual test and it still passed:
[bug_verify.txt](https://github.com/PBSPro/pbspro/files/3958290/bug_verify.txt)
* valgrind of mom with reproduction case rerun:
[valgrind.mom.txt](https://github.com/PBSPro/pbspro/files/4024996/valgrind.mom.txt)

### DIFF
--- a/src/lib/Libnet/get_hostaddr.c
+++ b/src/lib/Libnet/get_hostaddr.c
@@ -158,7 +158,8 @@ compare_short_hostname(char *shost, char *lhost)
 	is_shost_ip = inet_pton(AF_INET, shost, &(check_ip.sin_addr));
 	is_lhost_ip = inet_pton(AF_INET, lhost, &(check_ip.sin_addr));
 	if ((is_shost_ip > 0) || (is_lhost_ip > 0)) {
-		if (strcasecmp(shost, lhost) == 0)
+		/* ((3 * 4) + 3) = 15 characters, max length dotted decimal addr */
+		if (strncmp(shost, lhost, 15) == 0)
 			return 0;
 		return 1;
 	}
--- a/src/lib/Libnet/get_hostaddr.c
+++ b/src/lib/Libnet/get_hostaddr.c
@@ -42,6 +42,7 @@
 #include <netdb.h>
 #include <netinet/in.h>
 #include <memory.h>
+#include <arpa/inet.h>
 #include "portability.h"
 #include "server_limits.h"
 #include "pbs_ifl.h"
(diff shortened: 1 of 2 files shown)


src/resmom/mom_comm.c
```diff
@@ -5409,7 +5409,7 @@ im_request(int stream, int version)
 			resc_idx = -1;
 			for (i=0; i < pjob->ji_numrescs; i++) {
 				if ((pjob->ji_resources[i].nodehost != NULL) &&
-				    (compare_short_hostname(pjob->ji_resources[i].nodehost, nodehost) == 0)) {
+				    (strcmp(pjob->ji_resources[i].nodehost, nodehost) == 0)) {
```
mike0042: Would it be better to enhance compare_short_hostname() to accommodate IP addresses as well? Then it would work in other cases where it's called. If not, we should use strncmp() here.

bayucan: @mike0042 : yes, I agree, that compare_short_hostname() should recognize if IP addresses are involved and if so, just do the strcasecmp(). I've updated the fix and pushed a new commit. Please review.

src/lib/Libnet/get_hostaddr.c
```diff
@@ -146,10 +147,23 @@ compare_short_hostname(char *shost, char *lhost)
...
+	struct	sockaddr_in check_ip;
 
 	if ((shost == NULL) || (lhost == NULL))
 		return 1;
 
+	/* check if hostnames given are in IPV4 dotted-decimal form: ddd.ddd.ddd.ddd */
+	is_shost_ip = inet_pton(AF_INET, shost, &(check_ip.sin_addr));
+	is_lhost_ip = inet_pton(AF_INET, lhost, &(check_ip.sin_addr));
+	if ((is_shost_ip > 0) || (is_lhost_ip > 0)) {
+		if (strcasecmp(shost, lhost) == 0)
```
mike0042: We know at least one of the strings is the dotted decimal representation, and we know the maximum length of the string is ((3 * 4) + 3) = 15 characters. We should be able to do a strncmp() of length 15. Not a huge performance improvement, but the right tool for the job, IMHO. Otherwise, it looks good.

bayucan: I like your suggestion. Ok, I'll incorporate.

PR #1450, opened 2019-12-13
PP-823: qsub -Wblock=true hangs server for 190 seconds per job at exit if submission host is unresponsive

sujatapatnaik52:
#### Describe Bug or Feature
[PP-823](https://pbspro.atlassian.net/browse/PP-823) 

#### Describe Your Change
Introduced a new structure for handling block jobs. 

Once the job has finished, Server checks if the job is a block job and tries to connect to a non-blocking socket with the client. If the socket is connected or it is in progress  then the server sets a work task which will be triggered after 10 seconds and stores the socket alive in the block job structure. 

In the work task - 
1. If the socket is connected, server immediately sends the response to the client and deletes the job. 
2. If the socket connection is in progress, then work task tries to poll the socket for POLLOUT event with a 0 timeout.  If the polling does not work in the first try, server tries to connect for maximum of 60 seconds and after that it deletes the block job structure irrespective of client is available to receive. 

[PP-823-valgrind-logs.txt](https://github.com/PBSPro/pbspro/files/3960381/PP-823-valgrind-logs.txt)
[automated_test_logs.txt](https://github.com/PBSPro/pbspro/files/3960382/automated_test_logs.txt)
[block_job_retries_logs.txt](https://github.com/PBSPro/pbspro/files/3960383/block_job_retries_logs.txt)

### DIFF
--- a/src/include/job.h
+++ b/src/include/job.h
@@ -477,6 +477,18 @@ struct jbdscrd {
 #define TKMFLG_REVAL_IND_REMAINING 0x02 /* Flag to re-evaluate "array_indices_remaining" */
 #define TKMFLG_CHK_ARRAY           0x04 /* chk_array_doneness() already in call stack*/
 
+/* Structure for block job reply processing */
+struct block_job_reply {
+	char jobid[PBS_MAXSVRJOBID + 1];
+	char client[PBS_MAXHOSTNAME + 1];
+	int port;
+	int exitstat;
+	time_t reply_time;	/* The timestamp at which the block job tried it's first attempt to reply */
+	char *msg;	/* Abort message to be send to client */
+	int fd;
+};
+#define BLOCK_JOB_REPLY_TIMEOUT 60
+
 /*
  * THE JOB
  *
--- a/src/server/svr_jobfunc.c
+++ b/src/server/svr_jobfunc.c
@@ -98,6 +98,7 @@
 #include <signal.h>
 #include <sys/types.h>
 #include <sys/socket.h>
+#include <sys/poll.h>
 #include <netinet/in.h>
 #include <arpa/inet.h>
 #include "pbs_ifl.h"
@@ -1472,6 +1473,135 @@ svr_chkque(job *pjob, pbs_queue *pque, char *hostname, int mtype)
 
 /**
  * @brief
+ *		check_block_wt	-	A worktask to reply to the blocked job client
+ *
+ * @param[in]	ptask	-	work_task structure
+ */
+void
+check_block_wt(struct work_task *ptask)
+{
+	struct block_job_reply *blockj = ptask->wt_parm1;
+	struct pollfd fds[1];
+	int rc;
+	pbs_socklen_t	len = sizeof(rc);
+	int conn = 0;
+	int ret = 0;
+	int check_error;
+
+	if (blockj->fd == -1) {
+		int sock_flags;
+		struct hostent		*hp;
+		struct sockaddr_in	remote;
+
+		if ((hp = gethostbyname(blockj->client)) == NULL) {
+			sprintf(log_buffer, "client host %s not found for block job %s",
+			blockj->client, blockj->jobid);
+			goto err;
+		}
+
+		memset(&remote, 0, sizeof(remote));
+		memcpy(&remote.sin_addr, hp->h_addr, hp->h_length);
+		remote.sin_port = htons((unsigned short)blockj->port);
+		remote.sin_family = hp->h_addrtype;
+
+		if ((blockj->fd = socket(AF_INET, SOCK_STREAM, 0)) == -1) {
+			sprintf(log_buffer, "Failed to create socket for job %s", blockj->jobid);
+			goto err;
+		}
+
+		/* Set socket to Non-blocking */
+		sock_flags = fcntl(blockj->fd, F_GETFL, 0);
+		if (fcntl(blockj->fd, F_SETFL, sock_flags | O_NONBLOCK) == -1) {
+			sprintf(log_buffer, "Failed to set non-blocking flag on socket for job %s", 
+			blockj->jobid);
+			goto err;
+		}
+
+		conn = connect(blockj->fd, (struct sockaddr *)&remote, sizeof(remote));
+		if ((conn == -1) && !(errno == EINPROGRESS || errno == EWOULDBLOCK)) {
+			goto retry;
+		}
+	}
+
+	while (1) {
+		fds[0].fd = blockj->fd;
+		fds[0].events  = POLLOUT;
+		fds[0].revents = 0;
+
+		rc = poll(fds, (nfds_t)1, 0);
+		if (rc == -1) {
+			if ((errno != EAGAIN) && (errno != EINTR))
+				break;
+		} else
+			break;	/* no error */
+	}
+
+	if (rc <= 0)
+		goto retry;
+
+	rc = 0;
+	check_error = getsockopt(fds[0].fd, SOL_SOCKET, SO_ERROR, &rc, &len);
+	if ((rc != 0) || (check_error != 0) || (fds[0].revents != POLLOUT))
+		goto retry;
+
+	rc = CS_server_auth(blockj->fd);
+	if ((rc != CS_SUCCESS) && (rc != CS_AUTH_CHECK_PORT)) {
+		sprintf(log_buffer, "Unable to authenticate with %s:%d", blockj->client, blockj->port);
+		goto err;
+	}
+
+	/*
+	**	All ready to talk... now send the info.
+	*/
+
+	DIS_tcp_setup(blockj->fd);
+	ret = diswsi(blockj->fd, 1);		/* version */
+	if (ret != DIS_SUCCESS)
+		goto err;
+	ret = diswst(blockj->fd, blockj->jobid);
+	if (ret != DIS_SUCCESS)
+		goto err;
+	if (blockj->msg == NULL) {
+		ret = diswst(blockj->fd, "");
+	} else {
+		ret = diswst(blockj->fd, blockj->msg);
+	}
+	if (ret != DIS_SUCCESS)
+		goto err;
+	ret = diswsi(blockj->fd, blockj->exitstat);
+	if (ret != DIS_SUCCESS)
+		goto err;
+	(void)DIS_tcp_wflush(blockj->fd);
+
+	sprintf(log_buffer, "%s: Write successful to client %s for job %s ", __func__,
+	blockj->client, blockj->jobid);
+	log_event(PBSEVENT_DEBUG, PBS_EVENTCLASS_JOB, LOG_NOTICE, blockj->jobid, log_buffer);
+	CS_close_socket(blockj->fd);
+	goto end;
+
+retry:
+	if ((time(0) - blockj->reply_time) < BLOCK_JOB_REPLY_TIMEOUT) {
+		set_task(WORK_Timed, time_now + 10, check_block_wt, blockj);
+		return;
+	} else {
+		sprintf(log_buffer, "Unable to reply to client %s for job %s", 
+		blockj->client, blockj->jobid);
+	}
+err:
+	if (ret != DIS_SUCCESS) {
+		sprintf(log_buffer, "DIS error while replying to client %s for job %s",
+		blockj->client, blockj->jobid);
+	}
+	log_err(-1, __func__, log_buffer);
+end:
+	if (blockj->fd != -1) 
+		close(blockj->fd);
+	free(blockj->msg);
+	free(blockj);
+}
+
+/**
+ * @brief
  *		check_block	-	See if "block" is set and send reply.
  *
  * @param[in,out]	pjob	-	job structure
@@ -1481,16 +1611,9 @@ void
 check_block(job *pjob, char *message)
 {
 	int			port;
-	int			ret;
 	char			*phost;
 	char			*jobid = pjob->ji_qs.ji_jobid;
-	struct hostent		*hp;
-#ifdef WIN32
-	struct in_addr		addr;
-#endif
-	int			sock;
-	struct sockaddr_in	remote;
-	short			remote_sin_family;
+	struct 			block_job_reply *blockj; 
 
 	if ((pjob->ji_wattr[(int)JOB_ATR_block].at_flags & ATR_VFLAG_SET) == 0)
 		return;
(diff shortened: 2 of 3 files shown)


src/server/svr_jobfunc.c
```diff
@@ -1526,83 +1639,63 @@ check_block(job *pjob, char *message)
...
+			*(int *)ptask->wt_parm2 = 0;
+		}
+		else {
+			sprintf(log_buffer, "%s: Unable to connect to client %s for job %s", __func__, phost, jobid);
+			log_event(PBSEVENT_ERROR, PBS_EVENTCLASS_JOB, LOG_NOTICE, jobid, log_buffer);
+			close(blockj->ji_block_fd);
+			return;
+		}
+	}
+	if (is_connect == 0) {
```
subhasisb: use else here

src/server/svr_jobfunc.c
```diff
@@ -1526,83 +1639,63 @@ check_block(job *pjob, char *message)
...
+		ptask = set_task(WORK_Timed, time_now + 10, check_block_wt, blockj);
```
subhasisb: since this is connected right away, call the work_task function with WORK_immed to call it immediately

src/server/svr_jobfunc.c
```diff
@@ -1526,83 +1639,63 @@ check_block(job *pjob, char *message)
...
+		ptask->wt_parm2 = malloc(sizeof(int));
+		*(int *)ptask->wt_parm2 = 1;
```
subhasisb: just assign the value with a malloc

src/server/svr_jobfunc.c
```diff
@@ -1470,6 +1471,115 @@ svr_chkque(job *pjob, pbs_queue *pque, char *hostname, int mtype)
...
+	is_replied = 1;
+	}
+
+	if (is_replied == 0 && blockj->ji_reply_retries != BLOCK_JOB_REPLY_RETRIES) {
+		ptask = set_task(WORK_Timed, time_now + 10, check_block_wt, blockj);
+		ptask->wt_parm2 = malloc(sizeof(int));
+		*(int *)ptask->wt_parm2 = 0;
+	}
+
+	if (blockj->ji_reply_retries == BLOCK_JOB_REPLY_RETRIES || is_error == 1) {
```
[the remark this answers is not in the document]
sujatapatnaik52: There is a chance where the connection may never succeed and the control may not reach err:label. So keeping it outside.

src/include/job.h
```diff
@@ -477,6 +477,18 @@ struct jbdscrd {
 #define TKMFLG_REVAL_IND_REMAINING 0x02 /* Flag to re-evaluate "array_indices_remaining" */
 #define TKMFLG_CHK_ARRAY           0x04 /* chk_array_doneness() already in call stack*/
 
+/* Structure for block job reply processing */
+struct block_job_reply {
+	char ji_jobid[PBS_MAXSVRJOBID+1];
```
subhasisb: put a space around +

src/include/job.h
```diff
@@ -477,6 +477,18 @@ struct jbdscrd {
...
+	char ji_submit_host[PBS_MAXHOSTNAME+1];
+	int ji_port;
+	int ji_exitstat;
+	time_t ji_reply_timestamp;	/* The timestamp at which the block job tried it's first attempt to reply */
+	char *ji_msg;	/* Abort message to be send to client */
+	int ji_block_fd;	/* blocked job fds */
```
subhasisb: remove this comment, understandable without it

src/server/svr_jobfunc.c
```diff
@@ -1470,6 +1471,111 @@ svr_chkque(job *pjob, pbs_queue *pque, char *hostname, int mtype)
...
 }
 
+/**
+ * @brief
+ *		check_block_wt	-	A worktask to reply to the blocked job client
+ *
+ * @param[in]	ptask	-	work_task structure
+ */
+void
+check_block_wt(struct work_task *ptask) {
```
subhasisb: braces in the next line for functions

src/server/svr_jobfunc.c
```diff
@@ -1470,6 +1471,111 @@ svr_chkque(job *pjob, pbs_queue *pque, char *hostname, int mtype)
...
+		blockj->ji_submit_host, blockj->ji_port, dis_emsg[ret]);
+	log_event(PBSEVENT_ERROR, PBS_EVENTCLASS_JOB, LOG_NOTICE,
+		blockj->ji_jobid, log_buffer);
+done:
+	if ((ret = CS_close_socket(blockj->ji_block_fd)) != CS_SUCCESS) {
+		sprintf(log_buffer, "Problem closing security context, (%s:%d)",
+			blockj->ji_submit_host, blockj->ji_port);
+		log_joberr(-1, __func__, log_buffer, blockj->ji_jobid);
+		}
+	is_replied = 1;
```
subhasisb: indent

src/server/svr_jobfunc.c
```diff
@@ -1470,6 +1471,111 @@ svr_chkque(job *pjob, pbs_queue *pque, char *hostname, int mtype)
...
+	}
+
+	if ((time_difference >= BLOCK_JOB_REPLY_TIMEOUT) || is_error == 1) {
+		sprintf(log_buffer, "%s: Unable to send word to client %s for job %s", __func__,
+		blockj->ji_submit_host, blockj->ji_jobid);
+		log_event(PBSEVENT_ERROR, PBS_EVENTCLASS_JOB, LOG_NOTICE,
+		blockj->ji_jobid, log_buffer);
+	}
+
+	if (((time_difference >= BLOCK_JOB_REPLY_TIMEOUT)) || (is_replied == 1)){
```
subhasisb: indent

src/server/svr_jobfunc.c
```diff
@@ -1526,83 +1635,64 @@ check_block(job *pjob, char *message)
...
-	if ((ret != CS_SUCCESS) && (ret != CS_AUTH_CHECK_PORT)) {
-
-		sprintf(log_buffer,
-			"Unable to authenticate with , %s (%s:%d)", phost,
-			inet_ntoa(remote.sin_addr), remote.sin_port);
-
-		log_joberr(-1, __func__, log_buffer, jobid);
-		goto done;
+	blockj->ji_msg = strdup(message);
+	memset(blockj->ji_submit_host, '\0', PBS_MAXHOSTNAME + 1);
```
subhasisb: put a null at the last char after strncpy instead

src/server/svr_jobfunc.c
```diff
@@ -1470,6 +1471,111 @@ svr_chkque(job *pjob, pbs_queue *pque, char *hostname, int mtype)
...
+	 	*/
+
+		DIS_tcp_setup(blockj->ji_block_fd);
+		ret = diswsi(blockj->ji_block_fd, 1);		/* version */
+		if (ret != DIS_SUCCESS)
+			goto err;
+		ret = diswst(blockj->ji_block_fd, blockj->ji_jobid);
+		if (ret != DIS_SUCCESS)
+			goto err;
+		ret = diswst(blockj->ji_block_fd, blockj->ji_msg);
```
subhasisb: check if not null, if null send empty("") message

src/server/svr_jobfunc.c
```diff
@@ -1526,83 +1635,64 @@ check_block(job *pjob, char *message)
...
-		log_joberr(-1, __func__, log_buffer, jobid);
-		goto done;
+	blockj->ji_msg = strdup(message);
+	memset(blockj->ji_submit_host, '\0', PBS_MAXHOSTNAME + 1);
+	strncpy(blockj->ji_submit_host, phost, PBS_MAXHOSTNAME);
+	blockj->ji_port = port;
+	blockj->ji_reply_timestamp = time(NULL);
+	blockj->ji_exitstat = pjob->ji_qs.ji_un.ji_exect.ji_exitstat;
+	memset(blockj->ji_jobid, '\0', PBS_MAXSVRJOBID + 1);
+	strncpy(blockj->ji_jobid, pjob->ji_qs.ji_jobid, PBS_MAXSVRJOBID);
```
subhasisb: just use strcpy

src/server/svr_jobfunc.c
```diff
@@ -1526,83 +1635,64 @@ check_block(job *pjob, char *message)
...
 	}
 
-#ifdef WIN32
-	closesocket(sock);
-#else
-	close(sock);
-#endif
-
+	is_connect = connect(blockj->ji_block_fd, (struct sockaddr *)&remote, sizeof(remote));
+	connect_err = errno;
```
subhasisb: maybe the extra variable is not required

src/server/svr_jobfunc.c
```diff
@@ -1526,83 +1635,64 @@ check_block(job *pjob, char *message)
...
+	if (is_connect == -1) {
+		if (connect_err == EINPROGRESS || connect_err == ECONNREFUSED || connect_err == EADDRINUSE || (connect_err == ETIMEDOUT) ||
+			connect_err == ECONNREFUSED || connect_err == EWOULDBLOCK) {
+			ptask = set_task(WORK_Timed, time_now + 10, check_block_wt, blockj);
+			ptask->wt_parm2 = malloc(sizeof(int));
+			*(int *)ptask->wt_parm2 = 0;
+		}
+		else {
```
subhasisb: indent

src/server/svr_jobfunc.c
```diff
@@ -1470,6 +1471,111 @@ svr_chkque(job *pjob, pbs_queue *pque, char *hostname, int mtype)
...
+check_block_wt(struct work_task *ptask) {
+	
+	struct block_job_reply	*blockj = ptask->wt_parm1;
+	int time_difference;
+	time_t current_timestamp = time(NULL);
+	int is_connected = *(int *)ptask->wt_parm2;
+	int is_replied = 0;
+	int is_error = 0;
+
+	if (is_connected != 1) {
```
subhasisb: you can put the connect() also in this function itself, so that we can retry even if connect itself fails

sujatapatnaik52:
Hi @subhasisb, I have addressed your suggestions. Please review the latest commit.

src/server/svr_jobfunc.c
```diff
@@ -1470,6 +1471,176 @@ svr_chkque(job *pjob, pbs_queue *pque, char *hostname, int mtype)
...
+		/* Set socket to Non-blocking */
+		if (blockj->ji_block_fd != -1) {
+			sock_flags = fcntl(blockj->ji_block_fd, F_GETFL, 0);
+			if (fcntl(blockj->ji_block_fd, F_SETFL, sock_flags | O_NONBLOCK) == -1) {
+				close(blockj->ji_block_fd);
+				blockj->ji_block_fd = -1;
+			}
+		}
+
+		/* The flag "is_connected" indicates the readiness of the socket and could be one of the following:
```
bremanandjk: Just a suggestion: For ease of code readability!
Could we rename this flag as "**is_ready_to_write**"? and the flag values in the sequence order 0) Socket not created 1) Connect() in Progress 2) Successful connect() and poll(); Socket is ready to write msg.

sujatapatnaik52: Hi @bremanandjk, thank you for the suggestion. I agree with renaming the flag. I will make the changes. Regarding the sequence order, I kept it to be 1 for `is_connected` which indicates boolean for success(i.e. socket is ready for write). Also, in a case where I am polling and upon success, I make the variable from 2 to 1 again. I would keep the flag values as it is.

src/server/svr_jobfunc.c
```diff
@@ -1470,6 +1471,176 @@ svr_chkque(job *pjob, pbs_queue *pque, char *hostname, int mtype)
...
+		1. If it is 0, the socket is not created at all.
+		2. If it is 1, the socket is connected successfully and message can be send now.
+		3. If it is 2, the socket connection is in progress and needs to be polled now.
+		*/
+		if (blockj->ji_block_fd != -1) {
```
SudeshnaMoulik: Can we merge this 2 "if (blockj->ji_block_fd != -1)" block into one ? mean if we get a valid socket fd, then making it unblocking and verifying the status inside same block ?

src/server/svr_jobfunc.c
```diff
@@ -1470,6 +1471,176 @@ svr_chkque(job *pjob, pbs_queue *pque, char *hostname, int mtype)
...
+				is_connected = 1;
+			}
+		}
+		if (write_fds == -1) {
+			close(blockj->ji_block_fd);
+			blockj->ji_block_fd = -1;
+		}
+	}
+	
+	if (is_connected == 1) {
```
SudeshnaMoulik: Can we use "else if" here?

src/server/svr_jobfunc.c
```diff
@@ -1470,6 +1471,176 @@ svr_chkque(job *pjob, pbs_queue *pque, char *hostname, int mtype)
...
+		if (write_fds == 1) {
+			int rc = 0;
+			int check_error;
+			pbs_socklen_t len = sizeof(rc);
+			check_error = getsockopt(fds[0].fd, SOL_SOCKET, SO_ERROR, &rc, &len);
+			if ((rc == 0) && check_error == 0 && (fds[0].revents == POLLOUT)) {
+				is_connected = 1;
+			}
+		}
+		if (write_fds == -1) {
```
SudeshnaMoulik: Better to use "else if"

sujatapatnaik52:
Hi @bremanandjk, please review the latest commit.

src/server/svr_jobfunc.c
sujatapatnaik52: Hi @SudeshnaMoulik, Thank you for reviewing. Yes I could do that but I do not want it to try connecting if the socket non-blocking call fails.

src/server/svr_jobfunc.c
sujatapatnaik52: As per our offline discussion, I keep it as it is since I am setting this flag in the upper if block as well.

src/server/svr_jobfunc.c
SudeshnaMoulik: If we don't want to proceed further if "blockj->ji_block_fd == -1" and making the socket non-blocking fails, then can we ignore the blocks in-between and directly goto the part where we setting next timed_work_task ? This way we can ignore the in-between processing.

sujatapatnaik52:
Hi @SudeshnaMoulik, Please review the latest commit.

src/server/svr_jobfunc.c
```diff
@@ -1517,6 +1517,7 @@ check_block_wt(struct work_task *ptask)
 			if (fcntl(blockj->ji_block_fd, F_SETFL, sock_flags | O_NONBLOCK) == -1) {
 				close(blockj->ji_block_fd);
 				blockj->ji_block_fd = -1;
+				goto end;
 			}
 		}
```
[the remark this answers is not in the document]
SudeshnaMoulik: In that case we don't need the next if check, as we know if the flow goes till there, that mean we got valid non-blocking socket fd.

sujatapatnaik52:
@SudeshnaMoulik Addressed your suggestion. Please review the latest commit.

SudeshnaMoulik:
Can you please upload the latest PTL test log once?

sujatapatnaik52:
[updated_test_logs.txt](https://github.com/PBSPro/pbspro/files/3981949/updated_test_logs.txt)
[block_job_retries_logs.txt](https://github.com/PBSPro/pbspro/files/3981950/block_job_retries_logs.txt)

SudeshnaMoulik:
Thank you @sujatapatnaik52 for the changes. It looks good to me.
Do you think we should add one test case for the failure scenario, where after submitting the block job, we should try killing it and match for the log "check_block_wt: Unable to reply to client" in Server log ?

sujatapatnaik52:
> Thank you @sujatapatnaik52 for the changes. It looks good to me.
> Do you think we should add one test case for the failure scenario, where after submitting the block job, we should try killing it and match for the log "check_block_wt: Unable to reply to client" in Server log ?

Hi @SudeshnaMoulik, I tried to automate the test case and I think manual looks cleaner for now.

sujatapatnaik52:
Force-pushed for squash and rebase.

subhasisb:
Another set of eyes on this would be great!

sujatapatnaik52:
Hi @agrawalravi90, If my understanding is correct, the work task does not run in a different thread. We are considering 0 timeout for the poll and the non-blocking socket may not be ready for writing immediately. So, we are trying to give it some time via a work task and then retry.

subhasisb:
> I'm not sure I understand why the work task is needed, why not do this right after the job has finished instead of in a work task? Are work tasks run in different threads?

A work task is a "deferred" action - in the same server thread. Since this is triggered periodically, we come back to the wt function and check if a "connecting" socket has now completed the connection setup, and if it failed, try again. Thus this allows us to NOT block pbs_server, but still try the connection for a duration

agrawalravi90:
Ok, thanks for clarifying. I just don't like the fact that we'll make a job submitter wait for 10 seconds, especially if the job was tiny. This should ideally be done in a separate thread which would just wait for the host to respond and send it data. But I know that the server is not MT friendly, so this might be the best that we can do right now.

sujatapatnaik52:
Thank you @subhasisb and @agrawalravi90 for the sign off.

subhasisb:
> Ok, thanks for clarifying. I just don't like the fact that we'll make a job submitter wait for 10 seconds, especially if the job was tiny. This should ideally be done in a separate thread which would just wait for the host to respond and send it data. But I know that the server is not MT friendly, so this might be the best that we can do right now.

The first call to the work task is "immediate" - so if the connect is instant, then they will not have to wait. It is only waits 10 seconds for connections that could not complete in the first call to the work task function

agrawalravi90:
> > Ok, thanks for clarifying. I just don't like the fact that we'll make a job submitter wait for 10 seconds, especially if the job was tiny. This should ideally be done in a separate thread which would just wait for the host to respond and send it data. But I know that the server is not MT friendly, so this might be the best that we can do right now.
> 
> The first call to the work task is "immediate" - so if the connect is instant, then they will not have to wait. It is only waits 10 seconds for connections that could not complete in the first call to the work task function

Ok, that makes more sense, thanks for clarifying :)

PR #1352, opened 2019-10-14
PP-725: new "keep <select>" option for "pbs_release_nodes"

Shrini-h:
#### Describe Feature

RFE: [PP-725](https://pbspro.atlassian.net/browse/PP-725)

Implemented usage:
`pbs_release_nodes [-j job_identifier] -k <select string>`

As a user, when releasing nodes from my job I'd like to be able to provide a select statement that is a subset of the job in its current form and release the nodes which are not needed to satisfy that new select statement
Suppose I get a job by asking for:
```
qsub -l select=15:model=abc:mpiprocs=5+4:model=abc:bigmem=true:mpiprocs=1+30:model=def:mpiprocs=32
```

Then I want to release most of the nodes, but keep 2 of the non-bigmem abc nodes, and 2 of the bigmem abc nodes. It'd be much easier for the user to provide a new select statement:

```
pbs_release_nodes -k select=2:model=abc:mpiprocs=5+2:model=abc:bigmem=true:mpiprocs=1
```

than to have to determine via PBS_NODEFILE/qstat/pbsnodes which nodes have to be individually specified to pare back the job to that smaller select statement description.

#### Describe Your Change

1. Introduced -k option in the cmd by passing the new option to server using the `extend` parameter of api `pbs_relnodesjob()`
2. On the server side, the code flow reaches to `recreate_exec_vnode()` which invokes `pbs_release_nodes_given_select()` when select string is specified
3. The core logic is handled in `pbs_release_nodes_given_select()` which has been now modified to handle custom resources including non consumable ones like `string` and `boolean` type resources.

#### Link to Design Doc

https://pbspro.atlassian.net/wiki/spaces/PD/pages/1341325315/PP-725+new+keep+select+option+for+pbs+release+nodes

#### Attach Test and Valgrind Logs/Output

[SmokeTest.zip](https://github.com/PBSPro/pbspro/files/3898257/SmokeTest.zip)
RFE-New Test Suite: [TestPbsNodeRampDownKeepSelect.zip](https://github.com/PBSPro/pbspro/files/3898258/TestPbsNodeRampDownKeepSelect.zip)
Existing Test Suite[TestPbsNodeRampDown.zip](https://github.com/PBSPro/pbspro/files/3900997/TestPbsNodeRampDown.zip)

Valgrind 1:
[valgrind_leaks_found.txt](https://github.com/PBSPro/pbspro/files/3902554/valgrind_leaks_found.txt)

[...]

PR #1452, opened 2019-12-16
PP-724 and PP-725: new "keep <node count>" and "keep <select>" options for "pbs_release_nodes"

Shrini-h:
#### Describe Bug or Feature

RFE-1: [PP-724](https://pbspro.atlassian.net/browse/PP-724)
RFE-2: [PP-725](https://pbspro.atlassian.net/browse/PP-725)  (Originally https://github.com/PBSPro/pbspro/pull/1352)

Implemented usages:
1. `pbs_release_nodes [-j job_identifier] -k <node count>`
2. `pbs_release_nodes [-j job_identifier] -k <select string>`

RFE-1: [PP-724] :
As a user, when releasing nodes from my job, I'd like to be able to specify the number of sister nodes that should be kept. (i.e excluding the Mother Superior)

RFE-2: [PP-725] :
As a user, when releasing nodes from my job I'd like to be able to provide a select statement that is a subset of the job in its current form and release the nodes which are not needed to satisfy that new select statement
Suppose I get a job by asking for:
```
qsub -l select=15:model=abc:mpiprocs=5+4:model=abc:bigmem=true:mpiprocs=1+30:model=def:mpiprocs=32
```

Then I want to release most of the nodes, but keep 2 of the non-bigmem abc nodes, and 2 of the bigmem abc nodes. It'd be much easier for the user to provide a new select statement:

```
pbs_release_nodes -k select=2:model=abc:mpiprocs=5+2:model=abc:bigmem=true:mpiprocs=1
```

than to have to determine via PBS_NODEFILE/qstat/pbsnodes which nodes have to be individually specified to pare back the job to that smaller select statement description.

#### Describe Your Change

RFE-1: [PP-724] :
Developing on top of https://github.com/PBSPro/pbspro/pull/1352 RFE-2: [PP-725], I enabled `-k` option to accept natural number as argument which is then passed to the server after prefixing the string `select=`

RFE-2: [PP-725] : Originally https://github.com/PBSPro/pbspro/pull/1352
1. Introduced -k option in the cmd by passing the new option to server using the `extend` parameter of api `pbs_relnodesjob()`
2. On the server side, the code flow reaches to `recreate_exec_vnode()` which invokes `pbs_release_nodes_given_select()` when select string is specified
3. The core logic is handled in `pbs_release_nodes_given_select()` which has been now modified to handle custom resources including non consumable ones like `string` and `boolean` type resources.

#### Link to Design Doc

RFE-1: [PP-724] :
https://pbspro.atlassian.net/wiki/spaces/PD/pages/1403289607/PP-724+new+keep+node+count+option+for+pbs+release+nodes?atlOrigin=eyJpIjoiMjJiNTVjNTBkNmZkNDgzZDljNmVlYTU5NDczMGI3NzMiLCJwIjoiYyJ9

RFE-2: [PP-725] :
https://pbspro.atlassian.net/wiki/spaces/PD/pages/1341325315/PP-725+new+keep+select+option+for+pbs+release+nodes

RFE-1: [PP-724] :
[SmokeTest.zip](https://github.com/PBSPro/pbspro/files/3969536/SmokeTest.zip)
[TestPbsNodeRampDownKeepSelect.zip](https://github.com/PBSPro/pbspro/files/3969537/TestPbsNodeRampDownKeepSelect.zip)
[valgrind.zip](https://github.com/PBSPro/pbspro/files/3971841/valgrind.zip)

RFE-2: [PP-725] :
[SmokeTest.zip](https://github.com/PBSPro/pbspro/files/3898257/SmokeTest.zip)
RFE-New Test Suite: [TestPbsNodeRampDownKeepSelect.zip](https://github.com/PBSPro/pbspro/files/3898258/TestPbsNodeRampDownKeepSelect.zip)
Existing Test Suite[TestPbsNodeRampDown.zip](https://github.com/PBSPro/pbspro/files/3900997/TestPbsNodeRampDown.zip)

Valgrind 1:
[valgrind_leaks_found.txt](https://github.com/PBSPro/pbspro/files/3902554/valgrind_leaks_found.txt)

Valgrind out after mem leak fix: https://github.com/PBSPro/pbspro/pull/1352/commits/b960864fdd887973e31931c8651942a6c86a891b
[valgrind_after_leak_fix.zip](https://github.com/PBSPro/pbspro/files/3904587/valgrind_after_leak_fix.zip)

Test reports after commit https://github.com/PBSPro/pbspro/pull/1352/commits/616c2c3d9b323cc79bf5d212fdf261ae9fe9a0da : 
[SmokeTest.zip](https://github.com/PBSPro/pbspro/files/3959804/SmokeTest.zip)
[TestPbsNodeRampDown.zip](https://github.com/PBSPro/pbspro/files/3959783/TestPbsNodeRampDown.zip)
[TestPbsNodeRampDownKeepSelect.zip](https://github.com/PBSPro/pbspro/files/3959784/TestPbsNodeRampDownKeepSelect.zip)
[TestPbsReliableJobStartup.zip](https://github.com/PBSPro/pbspro/files/3959785/TestPbsReliableJobStartup.zip)
[New_valgrind.zip](https://github.com/PBSPro/pbspro/files/3960189/New_valgrind.zip)


Shrini-h:
Though this PR work was complete few weeks ago, I'm waiting for merge of the dependent PR https://github.com/PBSPro/pbspro/pull/1352, to remove the Draft label and make it ready for review

@bayucan

Shrini-h:
@bayucan and all,
as the dependent PR https://github.com/PBSPro/pbspro/pull/1352 is getting delayed for merge,
I would like to remove the `Draft` tag for this PR by requesting all to review only the commits whose comments start with text `PP-724` or alternately use the link https://github.com/PBSPro/pbspro/pull/1452/files/616c2c3d9b323cc79bf5d212fdf261ae9fe9a0da..29341930b6802d2bbd21e654e8a37629a042c60a

The other commits will be part of the baseline once the dependent PR https://github.com/PBSPro/pbspro/pull/1352 gets merged.

Thank You

src/server/vnparse.c
```diff
@@ -2486,6 +2481,139 @@ typedef struct resc_limit_entry {
...
+ * @param[in]	left - the left resource limit.
+ * @param[in]	right - the right resource limit.
+ *
+ * @return int
+ * @retval -1	- left < right
+ * @retval 0	- left = right
+ * @retval 1	- left > right
+ */
+static int
+resc_limit_list_cmp_name(resc_limit_t *left, resc_limit_t *right)
```
bayucan: Check to make sure params 'left' and 'right' are non-NULL before referencing them.

src/server/vnparse.c
```diff
@@ -2486,6 +2481,139 @@ typedef struct resc_limit_entry {
...
+ * @param[in]	left - the left resource limit.
+ * @param[in]	right - the right resource limit.
+ *
+ * @return int
+ * @retval -1	- left < right
+ * @retval 0	- left = right
+ * @retval 1	- left > right
+ */
+static int
+resc_limit_list_cmp_val(resc_limit_t *left, resc_limit_t *right)
```
bayucan: Check for NULL-ness of 'left' and 'right'.

src/server/vnparse.c
```diff
@@ -2595,6 +2745,84 @@ add_to_resc_limit_list_as_head(pbs_list_head *phead, resc_limit_t *resc)
...
+ * @param[in]	kv_keyw - resource name.
+ * @param[in]	kv_val - resource value.
+ * @param[in]	execv_f - flag to indicate if the resource was found in execvnode
+ *
+ * @return int
+ * @retval 0	- successful operation.
+ * @retval PBSE_Error	- Error Code.
+ */
+int
+resc_limit_insert_other_res(resc_limit_t *have, char *kv_keyw, char *kv_val, int execv_f)
```
bayucan: Check for NULL-ness of 'have', 'kv_keyw', and 'kv_val'...

src/server/vnparse.c
```diff
@@ -2931,11 +3239,92 @@ add_to_vnl(vnl_t **vnlp, char *noden, char *keyw, char *keyval)
...
+ *
+ * @param[in]	need - the need value of resc_limit_t type.
+ * @param[in]	have - the have value of resc_limit_t type.
+ *
+ * @return int
+ *	1 - if not satisfied
+ *	0 - otherwise
+ */
+static int
+check_other_res(resc_limit_t *need, resc_limit_t *have)
```
bayucan: Check for NULL-ness of 'need' and 'have'.

Shrini-h: Looking at the current invocation, Its impossible to get a NULL value in both parameters. Let me know if you still feel need a redundant check?

src/server/vnparse.c
Shrini-h: From all the location of invocations, `have` can never be NULL. For `kv_keyw` and `kv_val` has been handled in the new commit https://github.com/PBSPro/pbspro/pull/1452/commits/8cddc27f70d0dfe1f6738946eee0cfdcda7a311f

src/server/vnparse.c
```diff
@@ -2931,11 +3239,92 @@ add_to_vnl(vnl_t **vnlp, char *noden, char *keyw, char *keyval)
...
+ *
+ * @param[in/out]	new_schedselect - pointer to new sched select
+ * @param[in]		chunkstr - current chunk spec to append
+ * @param[in/out]	tmp_chunk_spec - buffer to accumulate identical chunk specs
+ * @param[in/out]	tmp_chunk_ct - count of gathered identical chunk specs
+ *
+ * @return none
+ */
+static void
+coagulate_append_sched_sel(char *new_schedselect, char *chunkstr, char *tmp_chunk_spec, int *tmp_chunk_ct)
```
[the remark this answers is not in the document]
Shrini-h: From the location of invocation except for `chunkstr` all other parameters can never be NULL. Hence only for `chunkstr`, the NULL check has been added in new commit https://github.com/PBSPro/pbspro/pull/1452/commits/8cddc27f70d0dfe1f6738946eee0cfdcda7a311f

Shrini-h:
@bayucan  I have addressed your comments by adding NULL checks at few places. However at other places I found from the invocations that it could never be NULL and so have accordingly replied. If you still feel an extra NULL check is needed, please let me know.

Thank You

src/server/vnparse.c
bayucan: Still, please do the check.

src/server/vnparse.c
bayucan: Can not say "never". I still want that NULL-ness check for 'have'. What if in the future, someone tries to reuse the function and inadvertently pass a NULL for 'have'. Then server will crash.

src/server/vnparse.c
bayucan: Yes, please still do the NULL-ness check.

src/server/vnparse.c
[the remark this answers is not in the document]
bayucan: The other parameters that are being deferenced (e.g. if (*tmp_chunk_spec)) still need to be checked for NULL-ness.

src/server/vnparse.c
bayucan: Yes, please still do the check.

src/server/vnparse.c
Shrini-h: Done, see commit https://github.com/PBSPro/pbspro/pull/1452/commits/6151018baa0b74a7f013bf2e272e4ce090374365

src/server/vnparse.c
Shrini-h: Done, see commit https://github.com/PBSPro/pbspro/pull/1452/commits/6151018baa0b74a7f013bf2e272e4ce090374365

src/server/vnparse.c
Shrini-h: OK, Done, see commit https://github.com/PBSPro/pbspro/pull/1452/commits/6151018baa0b74a7f013bf2e272e4ce090374365

src/server/vnparse.c
Shrini-h: Done, see commit https://github.com/PBSPro/pbspro/pull/1452/commits/6151018baa0b74a7f013bf2e272e4ce090374365

src/server/vnparse.c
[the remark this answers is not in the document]
Shrini-h: OK Done, see commit https://github.com/PBSPro/pbspro/pull/1452/commits/6151018baa0b74a7f013bf2e272e4ce090374365

Shrini-h:
@bayucan  I have now added all the NULL checks that was suggested. Please see commit https://github.com/PBSPro/pbspro/pull/1452/commits/6151018baa0b74a7f013bf2e272e4ce090374365
Please let me know if you have further comments
Thank You

src/cmds/pbs_release_nodes.c
```diff
@@ -51,8 +51,10 @@
 #include <pbs_version.h>
 
 #define USAGE	"usage: pbs_release_nodes [-j job_identifier] host_or_vnode1 host_or_vnode2 ...\n"
-#define USAGE2	"usage: pbs_release_nodes [-j job_identifier] -a\n"
-#define USAGE3	"       pbs_release_nodes --version\n"
+#define USAGE2	"       pbs_release_nodes [-j job_identifier] -a\n"
+#define USAGE3	"       pbs_release_nodes [-j job_identifier] -k <select string>\n"
+#define USAGE4	"       pbs_release_nodes [-j job_identifier] -k <node count>\n"
+#define USAGE5	"       pbs_release_nodes --version\n"
```
mike0042: Suggest you define the usage string in the following way:
```

#include <stdio.h>

#define USAGE \
        "This is line 1.\n" \
        "This is line 2.\n"

int
main(void)
{
        printf("%s", USAGE);
        return 0;
}
```

src/cmds/pbs_release_nodes.c
```diff
@@ -104,12 +110,23 @@ main(int argc, char **argv, char **envp) /* pbs_release_nodes */
...
+	}
+
 	if (errflg ||
-		((optind == argc) && !all_opt) ||
+		((optind == argc) && !(all_opt || keep_opt)) ||
 		((optind != argc) && all_opt) ) {
 		fprintf(stderr, USAGE);
 		fprintf(stderr, USAGE2);
 		fprintf(stderr, USAGE3);
+		fprintf(stderr, USAGE4);
```
mike0042: You missed USAGE5 here. Not required if you take my previous suggestion.

src/include/job.h
```diff
@@ -308,8 +308,11 @@ typedef struct resc_limit {		/* per node limits for Mom	*/
 	long long rl_vmem;		/* total mem space (virtual)	*/
 	int	  rl_naccels;		/* number of accelerators	*/
 	long long rl_accel_mem;		/* accelerator mem (real mem)	*/
+	pbs_list_head rl_oth_res;	/* list of all other resources found in execvnode and sched select*/
```
mike0042: It's good to keep variable names short, but I think we can add two characters for readability in this case. How about rl_other_res?

src/include/job.h
```diff
@@ -308,8 +308,11 @@ typedef struct resc_limit {		/* per node limits for Mom	*/
...
+	unsigned int  rl_res_count;	/* total count of resources */
```
mike0042: It's good to keep variable names short, but I think we can add two characters for readability in this case. How about rl_other_res?
I assume rl_res_count is the number of resources in rl_oth_res. How about naming it rl_other_count to show this relation?

src/lib/Libifl/pbsD_msgjob.c
```diff
@@ -202,6 +202,50 @@ char *extend;
...
+	/* first verify the resource list in keep_select option */
+	if (extend) {
+		struct attrl *attrib = NULL;
+		char ebuff[200], *erp, *emsg = NULL;
+		int i;
+		struct pbs_client_thread_connect_context *con;
+		char nd_ct_selstr[20];
+		char *endptr = NULL;
+
+		strtol(extend, &endptr, 10);
```
mike0042: You should check the return value and print strerror(errno) if the conversion fails.

src/lib/Libifl/pbsD_msgjob.c
```diff
@@ -202,6 +202,50 @@ char *extend;
 	if (pbs_client_thread_init_thread_context() != 0)
 		return pbs_errno;
 
+	/* first verify the resource list in keep_select option */
+	if (extend) {
+		struct attrl *attrib = NULL;
+		char ebuff[200], *erp, *emsg = NULL;
```
mike0042: Why is ebuf 200 characters? Should we use a macro here?

src/lib/Libifl/pbsD_msgjob.c
```diff
@@ -202,6 +202,50 @@ char *extend;
...
+		int i;
+		struct pbs_client_thread_connect_context *con;
+		char nd_ct_selstr[20];
```
mike0042: Why is this array 20 characters?

src/lib/Libifl/pbsD_msgjob.c
```diff
@@ -202,6 +202,50 @@ char *extend;
...
+			if (!attrib || strcmp(attrib->resource, "select")) {
+				emsg = strdup("only a \"select=\" string is valid in -k option\n");
+				pbs_errno = PBSE_IVALREQ;
+			} else
+				pbs_errno = PBSE_NONE;
+		}
+		if (pbs_errno) {
+			if ((con = pbs_client_thread_find_connect_context(c))) {
+				if (con->th_ch_errtxt)
+					free(con->th_ch_errtxt);
```
mike0042: Shouldn't have to check the value before calling free() assuming it was initialized properly.

src/lib/Libifl/pbsD_msgjob.c
```diff
@@ -202,6 +202,50 @@ char *extend;
...
+				con->th_ch_errtxt = emsg;
+				con->th_ch_errno = pbs_errno;
+			} else
+				connection[c].ch_errtxt = emsg;
```
mike0042: If c is the connection index passed in, why did we call pbs_client_thread_find_connect_context() to find it?
I assume the storage for emsg is free()'d when the connection is cleaned up?
Why don't we set connection[c].ch_errno here?

src/server/req_manager.c
```diff
@@ -3923,7 +3923,7 @@ check_resource_set_on_jobs_or_resvs(struct batch_request *preq, resource_def *pr
 static void
 timed_send_rescdef()
 {
-	send_rescdef(0);
+	send_rescdef(1);
```
mike0042: Might want to add a comment here indicating why the update is being forced.

src/server/req_message.c
```diff
@@ -341,7 +342,18 @@ req_relnodesjob(struct batch_request *preq)
 	if ((nodeslist != NULL) && (nodeslist[0] == '\0')) {
 		nodeslist = NULL;
 	}
-	rc = free_sister_vnodes(pjob, nodeslist, msg, LOG_BUF_SIZE, preq);
+
+	if (preq->rq_extend && *preq->rq_extend) {
+		keep_select = strchr(preq->rq_extend, '=');
+		if (!keep_select || !*(keep_select+1))
+			rc = PBSE_INVALSELECTRESC;
+		if (rc || (rc = validate_perm_res_in_select(++keep_select, 1))) {
```
mike0042: What you are doing here is not wrong, but it might be more readable as...
```
		if (keep_select && *(keep_select+1))
			keep_select++;
		else
			rc = PBSE_INVALSELECTRESC;
		if (rc || (rc = validate_perm_res_in_select(keep_select, 1))) {
```

src/server/svr_jobfunc.c
```diff
@@ -5964,6 +5964,86 @@ send_job_exec_update_to_mom(job *pjob, char *err_msg, int err_msg_sz,
...
+ *
+ * @return 	void
+ */
+static void
+populate_mom_list(pbs_list_head *to_head, char *exechostx)
+{
+	char *hostn = NULL, *last = NULL, *peh;
+	int  hasprn = 0;
+
+	if ((to_head == NULL) || (exechostx == NULL) || !*exechostx)
```
mike0042: We should be consistent with our checks. Either...
```
if ((to_head == NULL) || (exechostx == NULL) || (*exechostx != '\0'))
```
or...
```
if (!to_head || !exechostx || !*exechostx)
```
I'm partial to the latter.

src/server/svr_jobfunc.c
```diff
@@ -5964,6 +5964,86 @@ send_job_exec_update_to_mom(job *pjob, char *err_msg, int err_msg_sz,
...
+*/
+static char *
+get_ms_select_chunk(char *select_str)
+{
+	char *slast, *selbuf, *psubspec, *retval = NULL;
+	int   hpn;
+
+    if (select_str == NULL) {
+    	log_err(-1, __func__, "bad param passed");
+    	return (NULL);
```
mike0042: Indentation looks off here and in other parts of this function.

src/server/svr_jobfunc.c
```diff
@@ -6114,9 +6214,12 @@ recreate_exec_vnode(job *pjob, char *vnodelist, char *err_msg,
 		(void)update_resources_list(pjob, ATTR_l,
 			JOB_ATR_resource, new_exec_vnode, INCR, 0,
 				JOB_ATR_resource_orig);
+	} else {
+		log_err(-1, __func__, "new_exec_vnode is null or empty string");
+		goto recreate_exec_vnode_exit;
 	}
 
-	if (new_deallocated_execvnode[0] != '\0') {
+	if ((keep_select == NULL) && new_deallocated_execvnode && (new_deallocated_execvnode[0] != '\0')) {
```
mike0042: Please be consistent with how you perform these checks.

src/server/svr_jobfunc.c
```diff
@@ -6148,19 +6251,25 @@ recreate_exec_vnode(job *pjob, char *vnodelist, char *err_msg,
...
 			NULL,
 			new_exec_host2);
 		pjob->ji_modified = 1;
+	} else {
+		log_err(-1, __func__, "new_exec_host2 is null or empty string");
+		goto recreate_exec_vnode_exit;
 	}
 
-	if (new_select[0] != '\0') {
+	if (new_select && (new_select[0] != '\0')) {
```
mike0042: Not wrong, but other places you check (*newselect != '\0')

src/server/vnparse.c
```diff
@@ -1489,10 +1489,6 @@ manage_resc_sum_values(enum resc_sum_action action, resource_def *resc_def, char
 			return (NULL);
 		}
 		resc_sum_values_size = 20;
-		for (k = 0; k < resc_sum_values_size; k++) {
-			rs = resc_sum_values;
-			rs[k].rs_def = NULL;
-		}
```
mike0042: We no longer want to initialize these values?

src/server/vnparse.c
```diff
@@ -2486,6 +2481,147 @@ typedef struct resc_limit_entry {
 	resc_limit_t	*resc;
 } rl_entry;
 
+/* Reusing ATR_VFLAG_HOOK to mean that the resource key value pair was found
+ * in in execvnode
+ */
+#define IN_EXECVNODE_FLAG	ATR_VFLAG_HOOK
```
mike0042: I'm not sure that's the best idea. Should we define a new flag instead? If I was working in this file and hadn't read this comment, I'd probably be a little confused as to what is going on. Also, the comment contains "in in execvnode".

src/server/vnparse.c
```diff
@@ -2505,28 +2641,50 @@ add_to_resc_limit_list_sorted(pbs_list_head *phead, resc_limit_t *resc)
...
 		return (1);
 
-	plink_cur = phead;
-	p_entry_cur = (rl_entry *)GET_NEXT(*phead);
-
-	while (p_entry_cur) {
+	for (plink_cur = phead,
+		p_entry_cur = (rl_entry *)GET_NEXT(*phead);
+		p_entry_cur;
+		p_entry_cur = (rl_entry *)GET_NEXT(*plink_cur)) {
```
mike0042: Thank you for making this a for loop.

src/server/vnparse.c
```diff
@@ -2505,28 +2641,50 @@ add_to_resc_limit_list_sorted(pbs_list_head *phead, resc_limit_t *resc)
...
+#if defined(PBS_MOM) || defined(PBS_PYTHON)
 			/* order according to increasing # of cpus
 			 * if same # of cpus, use increasing amt of mem
 			 */
-			if ((p_entry_cur->resc->rl_ncpus > resc->rl_ncpus) ||
-			      ((p_entry_cur->resc->rl_ncpus == resc->rl_ncpus) &&
-				  (p_entry_cur->resc->rl_mem > resc->rl_mem))) {
+			if ((p_res_cur->rl_ncpus > resc->rl_ncpus) ||
+			      ((p_res_cur->rl_ncpus == resc->rl_ncpus) &&
+				  (p_res_cur->rl_mem > resc->rl_mem))) {
```
mike0042: Something not right with indentation on this line.

src/server/vnparse.c
```diff
@@ -2505,28 +2641,50 @@ add_to_resc_limit_list_sorted(pbs_list_head *phead, resc_limit_t *resc)
...
 				break;
 			}
-		}
+#else
+			if (p_res_cur->rl_res_count < resc->rl_res_count)
+				continue;
+			else if (p_res_cur->rl_res_count > resc->rl_res_count)
+				break;
+			else {
```
mike0042: This else block doesn't need to be here since this is the only code path we can take if we didn't break or continue. Don't get rid of the block, just don't indent it.

src/server/vnparse.c
```diff
@@ -2838,7 +3133,7 @@ map_need_to_have_resources(char *buf, size_t buf_sz, char *have_resc,
 {
 
 	if ((buf == NULL) || (buf_sz == 0) || (have_resc == NULL) ||
-	     (have_val == NULL) | (need == NULL)) {
+	     (have_val == NULL) || (need == NULL)) {
```
mike0042: Good catch.

src/server/vnparse.c
```diff
@@ -2931,11 +3262,102 @@ add_to_vnl(vnl_t **vnlp, char *noden, char *keyw, char *keyval)
...
+ *	Check if other resources in 'have' satisfy the remaining
+ *	 resources from 'need'.
+ *
+ * @param[in]	need - the need value of resc_limit_t type.
+ * @param[in]	have - the have value of resc_limit_t type.
+ *
+ * @return int
+ *	1 - if not satisfied
+ *	0 - otherwise
+ * -1 - error
```
mike0042: Indentation off.

src/server/vnparse.c
```diff
@@ -3426,15 +3961,6 @@ pbs_release_nodes_given_select(relnodes_input_t *r_input, relnodes_input_select_
...
 
-				if (parent_mom == NULL) { /* should not happen */
-
-					if ((err_msg != NULL) && (err_msg_sz > 0)) {
-						snprintf(err_msg, err_msg_sz, "no parent_mom for for %s", noden);
-        					log_event(PBSEVENT_DEBUG2, PBS_EVENTCLASS_JOB, LOG_DEBUG, r_input->jobid, err_msg);
-					}
-					goto release_nodes_exit;
-				}
-
```
mike0042: Even though this should not happen, why are we removing the check?

src/server/vnparse.c
```diff
@@ -3566,6 +4117,14 @@ pbs_release_nodes_given_select(relnodes_input_t *r_input, relnodes_input_select_
 					} else if (
 						strcmp(pkvp[j].kv_keyw, "accelerator_memory") == 0) {
 						have->rl_accel_mem += to_kbsize(pkvp[j].kv_val);
+#if !(defined(PBS_MOM) || defined(PBS_PYTHON))
+					} else {
+						if ((rc = resc_limit_insert_other_res(have, pkvp[j].kv_keyw, pkvp[j].kv_val, TRUE))) {
```
mike0042: Could you please set rc outside of the if statement? Then check if (rc != 0)

src/server/vnparse.c
```diff
@@ -2931,11 +3262,102 @@ add_to_vnl(vnl_t **vnlp, char *noden, char *keyw, char *keyval)
...
+ *
+ * @param[in/out]	new_schedselect - pointer to new sched select
+ * @param[in]		chunkstr - current chunk spec to append
+ * @param[in/out]	tmp_chunk_spec - buffer to accumulate identical chunk specs
+ * @param[in/out]	tmp_chunk_ct - count of gathered identical chunk specs
+ *
+ * @return none
+ */
+static void
+coagulate_append_sched_sel(char *new_schedselect, char *chunkstr, char *tmp_chunk_spec, int *tmp_chunk_ct)
```
mike0042: That's a strange name for a function. The word coagulate is almost always associated with blood or thickening of a liquid. I'd suggest append_and_group_sched_sel(). It's a little longer but more descriptive.

test/tests/functional/pbs_node_rampdown_keep_select.py
```diff
@@ -0,0 +1,1398 @@
...
+        self.job_stat_after = job_stat_after
+        self.expected_res_after = expected_res_after
+        self.skip_vnode_status_check = skip_vnode_status_check
+        self.use_script = use_script
+
+
+@requirements(num_moms=5)
+class TestPbsNodeRampDownKeepSelect(TestFunctional):
+    """
+    This tests the Node Rampdown Feature's extension called keep_select ,
```
mike0042: Space before comma.

test/tests/functional/pbs_node_rampdown_keep_select.py
```diff
@@ -0,0 +1,1398 @@
...
+        self.rel_user = rel_user
+        self.qsub_sel_after = qsub_sel_after
+        self.sched_sel_after = sched_sel_after
+        self.job_stat_after = job_stat_after
+        self.expected_res_after = expected_res_after
+        self.skip_vnode_status_check = skip_vnode_status_check
+        self.use_script = use_script
+
+
+@requirements(num_moms=5)
```
mike0042: Any way this requirement can be reduced?

Shrini-h:
@bayucan  and @mike0042 thank you for reviewing, The review comments in this PR are actually on the changes done in the base PR #1352 . I think its better if I close the base PR and modify the title of the current PR to have both the RFE's. Hence I will edit the title, description of this PR to reflect both the RFE's (Keep select spec and Keep node count).

Shrini-h:
This PR now also takes over the Base RFE from PR https://github.com/PBSPro/pbspro/pull/1352

src/lib/Libifl/pbsD_msgjob.c
Shrini-h: 20 is enough to fit string `"select=%d", INT_MAX`

src/lib/Libifl/pbsD_msgjob.c
Shrini-h: I thought of this, but since I'm using this `strtol()` to detect the case where the argument to `-k` option is just numbers and not a select spec, I feel the checks and the error logs will be intrusive with the else case and might confuse the user.

src/lib/Libifl/pbsD_msgjob.c
Shrini-h: OK, fixed in commit https://github.com/PBSPro/pbspro/pull/1452/commits/8f309272ad00f7587c830cad07238cb40598e24a

src/lib/Libifl/pbsD_msgjob.c
Shrini-h: the call to `pbs_client_thread_find_connect_context()` is to find `con` which is of type `struct pbs_client_thread_connect_context *`.
Yes, emsg is free()'d when the connection is cleaned up.
Frankly after erroring out from here, there's nobody reading `connection[c].ch_errno`. But anyways I added it in the commit https://github.com/PBSPro/pbspro/pull/1452/commits/8f309272ad00f7587c830cad07238cb40598e24a for completeness.

src/server/req_manager.c
Shrini-h: added it in commit https://github.com/PBSPro/pbspro/pull/1452/commits/8f309272ad00f7587c830cad07238cb40598e24a

src/server/req_message.c
Shrini-h: Agree, added it in commit https://github.com/PBSPro/pbspro/pull/1452/commits/8f309272ad00f7587c830cad07238cb40598e24a

src/server/svr_jobfunc.c
Shrini-h: OK, handled them in commit https://github.com/PBSPro/pbspro/pull/1452/commits/8f309272ad00f7587c830cad07238cb40598e24a

src/server/svr_jobfunc.c
Shrini-h: Thanks for the sharp eye. Fixed in commit https://github.com/PBSPro/pbspro/pull/1452/commits/8f309272ad00f7587c830cad07238cb40598e24a

src/server/svr_jobfunc.c
Shrini-h: Fixed in commit https://github.com/PBSPro/pbspro/pull/1452/commits/8f309272ad00f7587c830cad07238cb40598e24a

src/server/svr_jobfunc.c
Shrini-h: Fixed in commit https://github.com/PBSPro/pbspro/pull/1452/commits/8f309272ad00f7587c830cad07238cb40598e24a

src/server/vnparse.c
Shrini-h: Cause its already NULL due to the `calloc()` used in Line 1485

src/server/vnparse.c
Shrini-h: Well I understand your concern, however the number of bits is limited, so its best not to waste them in such minor isolated operations and save them for major distributed operations in future. The new name define with apt comments used in the code serves as an informant to coder working in future. The extra `in` has been removed in commit https://github.com/PBSPro/pbspro/pull/1452/commits/8f309272ad00f7587c830cad07238cb40598e24a

src/server/vnparse.c
Shrini-h: Fixed in commit https://github.com/PBSPro/pbspro/pull/1452/commits/8f309272ad00f7587c830cad07238cb40598e24a

src/server/vnparse.c
Shrini-h: handled in commit https://github.com/PBSPro/pbspro/pull/1452/commits/8f309272ad00f7587c830cad07238cb40598e24a

src/server/vnparse.c
Shrini-h: corrected in commit https://github.com/PBSPro/pbspro/pull/1452/commits/8f309272ad00f7587c830cad07238cb40598e24a

src/server/vnparse.c
Shrini-h: Agree, Addressed in commit https://github.com/PBSPro/pbspro/pull/1452/commits/8f309272ad00f7587c830cad07238cb40598e24a

src/server/vnparse.c
Shrini-h: Not removed. It just moved down to line #3996

src/server/vnparse.c
Shrini-h: Addressed in commit https://github.com/PBSPro/pbspro/pull/1452/commits/8f309272ad00f7587c830cad07238cb40598e24a

test/tests/functional/pbs_node_rampdown_keep_select.py
Shrini-h: Reducing it will require complete rework of all the test cases here. Moreover since the RFEs deal with node management, these minimum moms are kind of essential for thorough TC designs. By the way, Reliable Job Startup RFE also has 5 moms.

test/tests/functional/pbs_node_rampdown_keep_select.py
Shrini-h: Removed in commit https://github.com/PBSPro/pbspro/pull/1452/commits/8f309272ad00f7587c830cad07238cb40598e24a

src/server/vnparse.c
Shrini-h: the typo was a no harm though

Shrini-h:
@mike0042 , I have addressed most of your review comments and replied to others. Please have a look.
thanks

src/lib/Libifl/pbsD_msgjob.c
mike0042: I think I could break this code using: -k '123    '
*endptr would not point to '\0' in this case.

Shrini-h: Please try after the new commit https://github.com/PBSPro/pbspro/pull/1452/commits/b33b0c1672d0a8fc21ece2ce23438681e42db09f

src/server/req_message.c
```diff
@@ -345,9 +345,11 @@ req_relnodesjob(struct batch_request *preq)
 
 	if (preq->rq_extend && *preq->rq_extend) {
 		keep_select = strchr(preq->rq_extend, '=');
-		if (!keep_select || !*(keep_select+1))
+		if (keep_select && *(keep_select+1))
+			keep_select++;
+		else
 			rc = PBSE_INVALSELECTRESC;
-		if (rc || (rc = validate_perm_res_in_select(++keep_select, 1))) {
+		if (rc || (rc = validate_perm_res_in_select(keep_select, 1))) {
```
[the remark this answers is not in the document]
Shrini-h: Fixed in commit https://github.com/PBSPro/pbspro/pull/1452/commits/b33b0c1672d0a8fc21ece2ce23438681e42db09f

src/server/vnparse.c
Shrini-h: May be I was too conservative. So fixed it in commit https://github.com/PBSPro/pbspro/pull/1452/commits/b33b0c1672d0a8fc21ece2ce23438681e42db09f

src/server/vnparse.c
```diff
@@ -2658,17 +2658,18 @@ add_to_resc_limit_list_sorted(pbs_list_head *phead, resc_limit_t *resc)
...
 				break;
 			}
 #else
 			if (p_res_cur->rl_res_count < resc->rl_res_count)
 				continue;
 			else if (p_res_cur->rl_res_count > resc->rl_res_count)
 				break;
-			else {
+
+			{
```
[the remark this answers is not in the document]
Shrini-h: Done in commit https://github.com/PBSPro/pbspro/pull/1452/commits/b33b0c1672d0a8fc21ece2ce23438681e42db09f

Shrini-h:
@mike0042  I have addressed your new comments in commit https://github.com/PBSPro/pbspro/pull/1452/commits/b33b0c1672d0a8fc21ece2ce23438681e42db09f. Please have a look.

Also there seems to be dummy Travis CI stuck, I dont know how to resolve it.
Thank you

prakashcv13:
@Shrini-h - shall we try force push once and see if that helps?

Shrini-h:
> @Shrini-h - shall we try force push once and see if that helps?

thanks @prakashcv13  it got solved

src/lib/Libifl/pbsD_msgjob.c
```diff
@@ -227,12 +227,13 @@ char *extend;
 		int i;
 		struct pbs_client_thread_connect_context *con;
 		char nd_ct_selstr[20];
-		char *endptr = NULL;
+		char *endptr = NULL, *extend_dup = strdup(extend), *num_only;
```
bayucan: Need to check the return value of "strdup()" here, otherwise, you'll crash on the strtok(extend_dup,...) next.

Shrini-h: https://github.com/PBSPro/pbspro/pull/1452/commits/122ff5230827085550ab21708ca48da405d844a6 fixes it. thanks

src/lib/Libifl/pbsD_msgjob.c
```diff
@@ -229,27 +229,32 @@ char *extend;
...
-		} else if ((i = set_resources(&attrib, extend, 1, &erp))) {
-			if (i > 1) {
-				snprintf(ebuff, sizeof(ebuff), "%s: %s\n", emsg_illegal_k_value, pbs_parse_err_msg(i));
-				emsg = strdup(ebuff);
-			} else
-				emsg = strdup("illegal -k value\n");
-			pbs_errno = PBSE_INVALSELECTRESC;
+		if (extend_dup) {
+			num_only = strtok(extend_dup, " \'\"\n\t\v");
+			strtol(num_only, &endptr, 10);
```
mike0042: This has gotten way more complicated than I intended. My initial request was that you check the return code of strtol(). Please allow me to clarify...

Please set errno to zero before the call and check it after. Also check whether the return value is LONG_MIN or LONG_MAX, per the man page. Finally, check whether *endptr points to '\0'. If everything looks good, then the parameter is a valid number. If not, then try the set_resources() call.

Checking for whitespace or quotes addresses a case where someone supplies -k '\'123  ' but does not address the example I provided. Confirming that strtol() ran properly is all you really need to do. No need to strdup() the argument.

Otherwise, things are looking good to me.

Shrini-h: Hope the changes in https://github.com/PBSPro/pbspro/pull/1452/commits/5c04242d3f819f0c4004c997443a79b4e0a5286d suffice. Now ditched the endptr dependency. Not checking the return value for LONG_MIN or LONG_MAX as the errno will be set anyways for overlfows and underflows.

src/lib/Libifl/pbsD_msgjob.c
```diff
@@ -227,33 +228,28 @@ char *extend;
...
-					pbs_errno = PBSE_NONE;
-			}
-			free(extend_dup);
+		char *endptr = NULL;
+		long int rc_long;
+
+		errno = 0;
+		rc_long = strtol(extend, &endptr, 10);
+
+		if ((errno == 0) && (rc_long > 0)) {
```
mike0042: One minor suggestion, and then I'll be happy...
```suggestion
		if ((errno == 0) && (rc_long > 0) && (*endptr == '\0')) {
```

Shrini-h: > I think I could break this code using: -k '123 '
(Note there are spaces after 3.)
*endptr would not point to '\0' in this case.

This suggestion will fail to accept your initial breaking scenario `-k '123 '`. This is because `strtol()` ignores only leading white spaces and not trailing white space, so with this breaking input `*endptr` will not be `'\0'` but will be `' '`. Do you still want to commit this suggestion?

mike0042: Yes, it will fail to accept '123 ', and I think we want it to. You're testing whether the parameter is a number, and nothing more. If I supply 123abc as the argument to -k, then errno will be zero, rc_long will be 123, but *endptr will point to 'a'. Unless *endptr points to '\0', then the argument is not exclusively numeric. This will prevent people from trying to do anything out of the ordinary when supplying a numeric argument.

mike0042: One other thing I noticed was your check for rc_long > 0. That means negative numbers and zero will be identified as non-numeric. I think that is fine, since it will eventually throw an error.

Shrini-h: 🤕 ,Now I caught the confusion. when you meant 

> I could break this code

 I thought you were providing a valid input which the code won't accept as valid. Now I can say that the initial code was never broken!. Anyways with the confusion cleared, I accept the suggestion.

PR #1456, opened 2019-12-17
checkpoint_abort script does not exist on the mom

birbalsain:
#### Describe Bug or Feature
checkpoint_abort script does not exist on the remote mom.

#### Describe Your Change
create the temp file with the mom hostname.

[checkpoint_abort_file.txt](https://github.com/PBSPro/pbspro/files/4025248/checkpoint_abort_file.txt)

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -68,16 +68,16 @@ from ptl.utils.pbs_testusers import ROOT_USER, TEST_USER, PbsUser
 try:
     import psycopg2
     PSYCOPG = True
-except:
+except BaseException:
     PSYCOPG = False
 
 try:
     from ptl.lib.pbs_ifl import *
     API_OK = True
-except:
+except BaseException:
     try:
         from ptl.lib.pbs_ifl_mock import *
-    except:
+    except BaseException:
         sys.stderr.write("failed to import pbs_ifl, run pbs_swigify " +
                          "to make it\n")
         raise ImportError
@@ -274,7 +274,7 @@ class PtlConfig(object):
                 k = k.strip()
                 v = v.strip()
                 self.options[k] = v
-            except:
+            except BaseException:
                 self.logger.error('Error parsing line ' + line)
         # below two if block are for backword compatibility
         if 'PTL_EXPECT_MAX_ATTEMPTS' in self.options:
@@ -873,7 +873,7 @@ class PbsTypeSelect(list):
             if 'mem' in k:
                 try:
                     v = PbsTypeSize(v).value
-                except:
+                except BaseException:
                     # failed so we guessed wrong on the type
                     pass
             if isinstance(v, int) or v.isdigit():
@@ -1564,7 +1564,7 @@ class BatchUtils(object):
         :type name: str
         :param fmt: Optional formatting string, uses %n for
                     object name, %a for attributes, for example
-                    a format of '%nE{\}nE{\}t%aE{\}n' will display
+                    a format of r'%nE{\}nE{\}t%aE{\}n' will display
                     objects with their name starting on the first
                     column, a new line, and attributes indented by
                     a tab followed by a new line at the end.
@@ -1747,10 +1747,11 @@ class BatchUtils(object):
                     lines[_e] = lines[_e].strip('\r\n\t') + \
                         l[i].strip('\r\n\t')
                 elif (not l[i].startswith(' ') and i > count and
-                      l[i-count].startswith('\t')):
+                      l[i - count].startswith('\t')):
                     _e = len(lines) - count
                     lines[_e] = lines[_e] + l[i]
-                    if ((i+1) < len(l) and not l[i+1].startswith(('\t', ' '))):
+                    if ((i + 1) < len(l)
+                            and not l[i + 1].startswith(('\t', ' '))):
                         count += 1
                     else:
                         count = 1
@@ -2195,7 +2196,7 @@ class BatchUtils(object):
         """
         try:
             ret = time.mktime(time.strptime(st, '%a %b %d %H:%M:%S %Y'))
-        except:
+        except BaseException:
             ret = st
         return ret
 
@@ -2219,7 +2220,7 @@ class BatchUtils(object):
                                                             fmt=_f)
                 dtime_to = self.convert_datetime_to_epoch(m.group('dt_to'),
                                                           fmt=_f)
-            except:
+            except BaseException:
                 self.logger.error('error converting dedicated time')
         return (dtime_from, dtime_to)
 
@@ -3458,7 +3459,7 @@ class PBSService(PBSObject):
                     self.logger.info('FQDN name ' + self.fqdn + ' differs '
                                      'from name provided ' + self.hostname)
                     self.hostname = self.fqdn
-            except:
+            except BaseException:
                 pass
         else:
             self.fqdn = self.hostname
@@ -4014,7 +4015,7 @@ class PBSService(PBSObject):
         """
         try:
             from ptl.utils.pbs_logutils import PBSLogUtils
-        except:
+        except BaseException:
             _msg = 'error loading ptl.utils.pbs_logutils'
             raise ImportError(_msg)
 
@@ -4070,7 +4071,7 @@ class PBSService(PBSObject):
             # an open file descriptor, we close here but ignore errors in case
             # any were raised for all irrelevant cases
             lines.close()
-        except:
+        except BaseException:
             pass
         if (rv is None and existence) or (rv is not None and not existence):
             _msg = infomsg + attemptmsg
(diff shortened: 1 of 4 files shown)


test/tests/functional/pbs_checkpoint.py
```diff
@@ -144,7 +148,7 @@ def test_checkpoint_abort_with_preempt(self):
         self.server.manager(MGR_CMD_CREATE, QUEUE, a, "expressq")
 
         j1 = Job(TEST_USER, self.attrs)
-        j1.set_sleep_time(20)
+        j1.set_sleep_time(80)
```
arungrover: why did you increase the sleep time of the job?

test/tests/functional/pbs_release_limited_res_suspend.py
```diff
@@ -989,10 +989,13 @@ def test_preempt_checkpoint_release_all(self):
                 exit 0
                 """
         pbs_home = self.server.pbs_conf['PBS_HOME']
```
arungrover: shouldn't this be self.mompbs_conf['PBS_HOME'] since you are making a script on mom?

test/tests/functional/pbs_checkpoint.py
birbalsain: Because we need, both job must be available. but first job get finished before second job is started.

test/tests/functional/pbs_release_limited_res_suspend.py
birbalsain: we are just checking pbs home directory.

test/tests/functional/pbs_checkpoint.py
arungrover: Why is it that the first job finishes before second starts? I don't see any place in the test which would cause a delay unless there is some problem with the system itself.
In any case, If you really want to increase the sleep time, I'd say just don't set it. By default, every job sleeps for 100 seconds.

test/tests/functional/pbs_release_limited_res_suspend.py
arungrover: No, you are passing this PBS_HOME read from server's pbs.conf to create a file on mom.  Ideally you should be using PBS_HOME read from mom's pbs.conf file.

test/tests/functional/pbs_checkpoint.py
bhroam: Or if you don't need the to wait for the jobs to end, increase it longer than 100 seconds.

test/tests/functional/pbs_checkpoint.py
```diff
@@ -55,9 +55,12 @@ def setUp(self):
...
 """
-        self.abort_file = self.du.create_temp_file(body=abort_script)
-        self.du.chmod(path=self.abort_file, mode=0o755)
-        self.du.chown(path=self.abort_file, uid=0, gid=0, runas=ROOT_USER)
+        self.abort_file = self.du.create_temp_file(hostname=self.mom.hostname,
+                                                   body=abort_script)
+        self.du.chmod(hostname=self.mom.hostname,
+                      path=self.abort_file, mode=0o755)
+        self.du.chown(hostname=self.mom.hostname, path=self.abort_file,
+                      uid=0, gid=0, runas=ROOT_USER)
```
bhroam: Since this is a common mistake, maybe we should create a function as part of the mom similar to the function that creates a mom dynamic resource.  This one will create the checkpoint script, chmod/chown it, and set preempt_prio on the server.  This way people can always use this instead of this mistake popping up from time to time.

test/tests/functional/pbs_checkpoint.py
birbalsain: > Why is it that the first job finishes before second starts? I don't see any place in the test which would cause a delay unless there is some problem with the system itself.
> In any case, If you really want to increase the sleep time, I'd say just don't set it. By default, every job sleeps for 100 seconds.

Some platforms take long time to do job process.

arungrover: Hmmm, that seems odd that PBS could not preempt the job within 20 seconds of running one job when the test does not even have any sleep. In any case, just don't set and it will be set to 100 seconds by default.

test/tests/functional/pbs_checkpoint.py
birbalsain: @bhroam check the new commit.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -13452,6 +13453,21 @@ def create_vnode_def(self, name, attrs={}, numnodes=1, sharednode=True,
...
+        """
+        Add checkpoint script in the mom config.
+        returns: a temp file for checkpoint script
+        """
+        chk_file = self.du.create_temp_file(hostname=self.hostname, body=body,
+                                            dirname=dirname)
+        self.du.chmod(hostname=self.hostname, path=chk_file, mode=mode)
+        self.du.chown(hostname=self.hostname, path=chk_file, uid=uid,
+                      gid=gid, sudo=sudo, runas=runas)
+        c = {'$action': 'checkpoint_abort 30 !' + chk_file + ' %sid'}
```
bhroam: Add 30 as a default argument.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -13452,6 +13453,21 @@ def create_vnode_def(self, name, attrs={}, numnodes=1, sharednode=True,
         del attribs
         return "\n".join(vdef)
 
+    def add_checkpoint_script(self, dirname=None, body=None, mode=None,
```
bhroam: There are more than one checkpoint $action scripts in the mom config.  Call this function add_checkpoint_abort_script()

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -13452,6 +13453,21 @@ def create_vnode_def(self, name, attrs={}, numnodes=1, sharednode=True,
...
+                              runas=None, sudo=False, uid=None, gid=None):
```
bhroam: I don't think you need runas, sudo, mode, uid, or gid.  The mom will complain if the script isn't secure.  I think we should force these to be root with secure permissions.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -68,16 +68,16 @@
 try:
     import psycopg2
     PSYCOPG = True
-except:
+except BaseException:
```
bhroam: Why the change from the blanket excepts to BaseException?  The reason you don't do blanket excepts is because it catches things like KeyboardInterrupt.  BaseException includes that.  This isn't the right PR to do this change.  When we get rid of the blanket excepts in pbs_testlib.py, we should go through and actually look what the try/except block is expects and catch that.

test/tests/functional/pbs_release_limited_res_suspend.py
```diff
@@ -988,17 +988,14 @@ def test_preempt_checkpoint_release_all(self):
...
-                                                 body=chk_script,
-                                                 dirname=pbs_home)
-        self.du.chmod(hostname=self.mom.hostname,
-                      path=self.chk_file, mode=0o755)
-        self.du.chown(hostname=self.mom.hostname,
-                      path=self.chk_file, uid=0, gid=0, sudo=True)
-        c = {'$action': 'checkpoint_abort 30 !' + self.chk_file + ' %sid'}
-        self.mom.add_config(c)
-
+        pbs_home = self.mom.pbs_conf['PBS_HOME']
```
birbalsain: @arungrover updated pbs_home.

test/tests/functional/pbs_checkpoint.py
```diff
@@ -148,7 +142,6 @@ def test_checkpoint_abort_with_preempt(self):
         self.server.manager(MGR_CMD_CREATE, QUEUE, a, "expressq")
 
         j1 = Job(TEST_USER, self.attrs)
-        j1.set_sleep_time(80)
```
birbalsain: @arungrover , remove the set sleep time

test/fw/ptl/lib/pbs_testlib.py
birbalsain: @bhroam , removed the changes.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -13453,18 +13453,17 @@ def create_vnode_def(self, name, attrs={}, numnodes=1, sharednode=True,
         del attribs
         return "\n".join(vdef)
 
-    def add_checkpoint_script(self, dirname=None, body=None, mode=None,
-                              runas=None, sudo=False, uid=None, gid=None):
+    def add_checkpoint_abort_script(self, dirname=None, body=None,
+                                    mode=None, abort_time=30):
```
bhroam: Passing mode=None causes chmod() to fail.  We know what will work for a checkpoint_abort script.  For security reasons, I'd suggest we use 0o700 instead of 0o755.

birbalsain: I think, it is good to remove  mode as a argument. we can direct use mode=0o700 in chmod().

test/tests/functional/pbs_release_limited_res_suspend.py
birbalsain: @arungrover  I addressed your requested changes

PR #1458, opened 2019-12-17
Fix race condition on qrerun and qrun with end job hook

riyazhakki:
#### Describe Bug or Feature

job stays in substate 41 after qrerun in epilogue script and a single rerun with an execjob_end hook leads to run_count 3 and intermittently Exit_Status=-3

Cause:
Race condition occur on force qrerun and qrun of same job, while asynchronous hook (exec_job_end) is running in the background.

#### Describe Your Change

Allowing job to rerun only when it's discard request is acknowledged

[TestPbsExecjobEnd.ptl.log](https://github.com/PBSPro/pbspro/files/3983291/TestPbsExecjobEnd.ptl.log)
[TestPbsExecjobEnd.ptl.log](https://github.com/PBSPro/pbspro/files/3974527/TestPbsExecjobEnd.ptl.log)
[valgrind.log](https://github.com/PBSPro/pbspro/files/3974528/valgrind.log)

### DIFF
--- a/src/include/job.h
+++ b/src/include/job.h
@@ -441,6 +441,7 @@ struct ajtrk {
 	int trk_exitstat;  /* if executed and exitstat set */
 	int trk_substate; /* sub state    */
 	int trk_stgout; /* stageout status  */
+	int trk_discarding; /* indicate job is discarding */
 	struct job *trk_psubjob; /* pointer to instantiated subjob */
 };
 
@@ -585,6 +586,7 @@ struct job {
 	int		ji_ports[2];	/* ports for stdout/err */
 	int		ji_hook_running_bg_on; /* set when hook starts in the background*/
 #else					/* END Mom ONLY -  start Server ONLY */
+	int		ji_discarding;	/* discarding job */
 	struct batch_request *ji_prunreq; /* outstanding runjob request */
 	pbs_list_head	ji_svrtask;	/* links to svr work_task list */
 	struct pbs_queue  *ji_qhdr;	/* current queue header */
@@ -853,7 +855,7 @@ typedef struct	infoent {
 #define IM_RESTART		16
 #define IM_DELETE_JOB		17
 #define IM_REQUEUE		18
-#define	IM_DELETE_JOB_REPLY	19
+/* skip 19 to avoid confilict with IS_DISCARD_JOB */
 #define IM_SETUP_JOB		20
 #define IM_DELETE_JOB2		21	/* sent by sister mom to delete job early */
 #define IM_SEND_RESC		22
@@ -861,6 +863,7 @@ typedef struct	infoent {
 #define IM_EXEC_PROLOGUE	24
 #define IM_CRED 		25
 #define IM_PMIX			26
+#define	IM_DELETE_JOB_REPLY	27
 
 #define IM_ERROR		99
 #define IM_ERROR2		100
--- a/src/include/job.h
+++ b/src/include/job.h
@@ -504,6 +504,15 @@ struct jbdscrd {
 #define	ji_taskid	ji_extended.ji_ext.ji_taskidx
 #define	ji_nodeid	ji_extended.ji_ext.ji_nodeidx
 
+enum bg_hook_request {
+	BG_NONE,
+	BG_IS_DISCARD_JOB,
+	BG_PBS_BATCH_DeleteJob,
+	BG_PBSE_SISCOMM,
+	BG_IM_DELETE_JOB_REPLY,
+	BG_IM_DELETE_JOB
+};
+
 struct job {
 
 	/* 
@@ -584,7 +593,7 @@ struct job {
 	int		ji_stdout;	/* socket for stdout */
 	int		ji_stderr;	/* socket for stderr */
 	int		ji_ports[2];	/* ports for stdout/err */
-	int		ji_hook_running_bg_on; /* set when hook starts in the background*/
+	enum	bg_hook_request	ji_hook_running_bg_on; /* set when hook starts in the background*/
 #else					/* END Mom ONLY -  start Server ONLY */
 	int		ji_discarding;	/* discarding job */
 	struct batch_request *ji_prunreq; /* outstanding runjob request */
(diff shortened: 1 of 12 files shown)


riyazhakki:
Hi All,
This PR is up for the review.

src/include/job.h
```diff
@@ -853,14 +855,15 @@ typedef struct	infoent {
 #define IM_RESTART		16
 #define IM_DELETE_JOB		17
 #define IM_REQUEUE		18
-#define	IM_DELETE_JOB_REPLY	19
+/* skip 19 to avoid confilict with IS_DISCARD_JOB */
```
arungrover: Why is this change needed? One is inter-mom message and the other is inter-server message. How can those two conflict?

src/resmom/mom_hook_func.c
```diff
@@ -3424,7 +3424,32 @@ void reply_hook_bg(job *pjob)
 	struct	batch_request *preq = pjob->ji_preq;
 #endif
 
-	if (pjob->ji_qs.ji_svrflags & JOB_SVFLG_HERE) { /*MS*/
+	if (pjob->ji_hook_running_bg_on == IS_DISCARD_JOB) {
```
arungrover: I do not understand the purpose of ji_hook_running_bg_on. It stores inter-server/inter-mom/batch-request/batch-request+errno in it. These are all independent things, do you know why this one variable is used to store all of these things?

src/server/array_func.c
```diff
@@ -535,6 +535,24 @@ update_subjob_state(job *pjob, int newstate)
...
+ * @param[in]	parent - pointer to the parent job
+ * @param[in]	iindx - first number of range
+ *
+ * @return	status
+ * @retval	-1	-  error
+ */
+int
+get_subjob_discarding(job *parent, int iindx)
+{
+	if (iindx == -1)
```
arungrover: Is there a scenario that iindx can be -1? If not then this function isn't really needed because caller can just access discarding variable of the subjob.

src/server/node_manager.c
```diff
@@ -840,8 +846,14 @@ post_discard_job(job *pjob, mominfo_t *pmom, int newstate)
 		return;
 	}
 
-	if (pjob->ji_qs.ji_substate == JOB_SUBSTATE_RERUN3) {
-		static char ndreque[] = "Job requeued, execution node %s down";
+	if (pjob->ji_qs.ji_substate == JOB_SUBSTATE_RERUN3 || pjob->ji_discarding) {
+		
+		static char *ndreque;
```
arungrover: why is this static?

src/server/req_runjob.c
```diff
@@ -377,6 +381,9 @@ req_runjob(struct batch_request *preq)
 			/* job already running */
 			req_reject(PBSE_BADSTATE, 0, preq);
 			return;
+		} else if (get_subjob_discarding(parent, i) == 1) {
```
arungrover: Why are you passing in 'i' (state of the subjob). Shouldn't you be passing offset to this function? I'd also argue if the function is even needed.

src/include/job.h
riyazhakki: @arungrover Thank you for reviewing and providing the comments.

> @riyazhakki Can you please explain a little more about your change and about what happens in this race condition.

During execution of the epilogue hook if the user hits the 'rerun -W force job_id' then the server sends IS_DISCARD_JOB to mom (mom kills the job and starts the exec_jobend hook in the background) and force queue at #L469, in this case it get scheduled to the same exec_vnode where the previous version was running. While the mom starts running hook in the background at the same time it receives the request to rerun the next version of same job, as mom is not blocked for running exec_jobend (background) hook.
When this job version lands on mom, req_commit() will change its state to 41 and start_exec function will abort it, as it sees its previous version is not yet cleaned up. Later, after execution of execjob_end hook (of old job version) mom will purge the job structure and on the server it remains in 41 state and similarly it happens with qrun also.
 

> Why is this change needed? One is inter-mom message and the other is inter-server message. How can those two conflict?

Sister node can receive both messages IM_DELETE_JOB_REPLY and IS_DISCARD_JOB (in case of node fail requeue), and replying to these request is done in a function reply_hook_bg(), therefore the above change is required.

src/resmom/mom_hook_func.c
riyazhakki: ji_hook_running_bg_on is used indicate the hook is running in the background and it stores inter-server/inter-mom/batch-request/batch-request+errno in it to send appropriate reply from the deferred function (reply_bg_hook).

src/server/array_func.c
riyazhakki: Yes, function subjob_index_to_offset can make it -1.

src/server/node_manager.c
riyazhakki: I have retained it as before.

src/server/req_runjob.c
riyazhakki: Yes, this should be offset, will change it.

src/resmom/mom_hook_func.c
arungrover: This seems incorrect. A variable should not store 2 or more distinct things in it. Like in the case statement you check for "PBS_BATCH_DeleteJob + PBSE_SISCOMM" this will be same as "PBS_BATCH_Commit + PBSE_SVRDOWN". While the example above may not be practically possible but it is a possibility that at some point someone might hit this issue and it then it will be very difficult to diagnose.

riyazhakki: I see even it is two distinct values, will be unique for the event so there is little chance of getting into the issue.
The reason for holding these values is, when MS receives PBS_BATCH_DeleteJob request, it starts execution of the hook in the background and at the same time it sends the IM_DELETE_JOB_REPLY to all its sisters, if it is unable to send this request to one or more sisters, we need to reply the batch request with PBSE_SISCOMM error and this reply is done in the deferred function of hook running in the background, to carry this error value from function req_deletejob() to reply_hook_bg() we are using the ji_hook_running_bg_on variable.

riyazhakki:
[TestPbsExecjobEnd.ptl.log](https://github.com/PBSPro/pbspro/files/4030927/TestPbsExecjobEnd.ptl.log)

src/resmom/mom_hook_func.c
arungrover: I hear what you are saying, but I don't like the fact that there is already a clash of values stored in this variable which is causing you to change the inter-mom request command IDs. This means that older moms and newer moms will not understand the messaged being passed between them (although, this may not be supported but it is possible, especially during upgrades)
I'd suggest that instead of storing existing batch request IDs, IM/IS command IDs, error number (or combinations of all three) you create a separate enum and add all possible values this variable is currently used for in that enum and then you don't have to change the Inter-mom command.

riyazhakki:
[TestPbsExecjobEnd.ptl.log](https://github.com/PBSPro/pbspro/files/4045732/TestPbsExecjobEnd.ptl.log)

PR #1467, opened 2020-01-08
Fixing unknown state message in logs for maintainence state

tpai26:
#### Describe Bug or Feature
The scheduler doesn't know about the new maintenance state. This will cause the scheduler to log an error message that a node is in an unknown state for each node in that state. On large systems, this could clog up the log.
This fix adds a maintenance node state in node_info so that the scheduler knows it's not an unknown node state.

[bad_state_log.txt](https://github.com/PBSPro/pbspro/files/4033331/bad_state_log.txt)


test/tests/functional/pbs_sched_badstate.py
```diff
@@ -0,0 +1,153 @@
+# coding: utf-8
+
+# Copyright (C) 1994-2019 Altair Engineering, Inc.
```
agrawalravi90: 2020

test/tests/functional/pbs_sched_badstate.py
```diff
@@ -0,0 +1,153 @@
...
+        - check the sched log for "Leaving sched cycle" from the time
+          array job was submitted.
+        If we are unable to find a log match then scheduler is in
+          endless loop and test case has failed.
+        """
+
+        self.mom.signal('-KILL')
+
+        attr = {'state': 'free', 'resources_available.ncpus': '2'}
+        self.server.manager(MGR_CMD_SET, NODE, attr, self.mom.shortname)
```
agrawalravi90: I know that you just moved an existing test, but according to our documentation, 'state' cannot be set to 'free', so we shouldn't be setting it (as such, PBS should not even allow this).

tpai26: The bug for which test was written required a scenario where the server fails to send a subjob to a mom.... the way send job from server to mom could have failed is when mom is down but the node is seen as free by scheduler.
Therefore, even though the docs say we should not, in this case we have to set the node state to  'free'

agrawalravi90: I understand, but it's not that it "shouldn't" be set, it's that it "cannot" be set. If it can be set today to "Free", then that's a bug, which might get fixed at some point and then this test will start failing. If this test was written to catch a race condition, then we should just remove it, race conditions are race conditions because they cannot be reproduced reliably, so no point writing a test to reproduce them.

bhroam: @agrawalravi90 there is nothing stopping you from setting the state of a node to free.  It is a valid node state.  It just isn't a good idea.  If a node is down and you set it back to free, it will just be marked down again when PBS realizes it is down.  I don't think this has ever been considered a bug, although we might want to disallow it as you suggest.  In any case, I think this test requires you to set the node to free while the mom is down.  That's what the docstring says anyway.  I think @Shrini-h wrote this test.  Maybe he can comment.

agrawalravi90: Ok, I was going by the Ref guide which says that the permissions for the stat attr are 'r, r, r' when setting it to 'free'. I just don't want us to come back to this test and think about what to do with it if we decide to implement the documented behavior in the future.

agrawalravi90: actually, I don't wanna hold this PR because of this test as it is an already existing test. If it causes issues in the future then we can come back to it.

Shrini-h: > I think @Shrini-h wrote this test. Maybe he can comment.

@bhroam   Sorry, I did not author this test. From git history, this has been introduced in PR https://github.com/PBSPro/pbspro/pull/264

PR #1471, opened 2020-01-14
Passes user to create job script on remote and read job_output file from client

birbalsain:
#### Describe Bug or Feature
Job output file is not on the server. Read it from client.
We should use a user to create job script on remote.

#### Describe Your Change
Passes test_user as a user to create a job script on remote and read job output file on client by using client hostname and job submitting user.

[enviroment_variable (1).txt](https://github.com/PBSPro/pbspro/files/4084519/enviroment_variable.1.txt)

### DIFF
--- a/test/tests/functional/pbs_passing_environment_variable.py
+++ b/test/tests/functional/pbs_passing_environment_variable.py
@@ -74,14 +74,17 @@ class Test_passing_environment_variable_via_qsub(TestFunctional):
              'Resource_List.walltime': 10}
         script = ['#PBS -v "var1=\'A,B,C,D\'"']
         script += ['env | grep var1']
-        jid = self.create_and_submit_job(content=script)
+        jid = self.create_and_submit_job(user=TEST_USER, content=script)
         qstat = self.server.status(JOB, ATTR_o, id=jid)
         job_outfile = qstat[0][ATTR_o].split(':')[1]
 
         self.server.expect(JOB, 'queue', op=UNSET, id=jid, offset=10)
         job_output = ""
-        with open(job_outfile, 'r') as f:
-            job_output = f.read().strip()
+        runcmd = ['cat', job_outfile]
+        ret = self.du.run_cmd(self.server.client, runcmd,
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_passing_environment_variable.py
```diff
@@ -74,14 +74,17 @@ def test_commas_in_custom_variable(self):
...
-        jid = self.create_and_submit_job(content=script)
+        jid = self.create_and_submit_job(user=TEST_USER, content=script)
         qstat = self.server.status(JOB, ATTR_o, id=jid)
         job_outfile = qstat[0][ATTR_o].split(':')[1]
 
         self.server.expect(JOB, 'queue', op=UNSET, id=jid, offset=10)
         job_output = ""
-        with open(job_outfile, 'r') as f:
-            job_output = f.read().strip()
+        runcmd = ['cat', job_outfile]
```
bhroam: use self.du.cat()

test/tests/functional/pbs_passing_environment_variable.py
```diff
@@ -74,14 +74,17 @@ def test_commas_in_custom_variable(self):
...
+        ret = self.du.run_cmd(self.server.client, runcmd,
+                              level=logging.INFOCLI, runas=TEST_USER,
+                              logerr=False)
+        job_output =(' '.join(map(str, ret['out']))).strip()
```
bhroam: I believe the output of ret['out'] is a list of strings.  I don't think you need to map it through str.  In any case, the output of self.du.cat() is a list of strings, so it will work.

bhroam:
Travis failed due to pep8.  You are missing spaces around '='

test/tests/functional/pbs_passing_environment_variable.py
bhroam: I'm still not sure you need to do the map().  I thought the output of self.du.cat( ) was a list of strings.  You should be able to do the join() directly.

birbalsain: @bhroam i have remove the map().

PR #1473, opened 2020-01-15
Add a timeout for server_dyn_res scripts

vstumpf:
#### Describe Bug or Feature
If server_dyn_res scripts never return, they can hang the scheduler forever.

#### Describe Your Change
Add a timeout to the server_dyn_res scripts by changing popen() to fork() + select()

#### Link to Design Doc
https://pbspro.atlassian.net/wiki/spaces/PD/pages/51975257/PP-305+If+server+dyn+res+script+does+not+return+on+UNIX+LINUX+scheduler+will+hang?flashId=c207b48808fc0e81844292031df70e66003096d2

[serverdynres.txt](https://github.com/PBSPro/pbspro/files/4062764/serverdynres.txt)


src/include/pbs_sched.h
```diff
@@ -64,6 +64,10 @@ extern "C" {
 /* Default value of preempt_queue_prio */
 #define PBS_PREEMPT_QUEUE_PRIO_DEFAULT		150
 
+/* Default value of server_dyn_res_alarm */
+#define PBS_SERVER_DYN_RES_ALARM_DEFAULT	30
```
bhroam: Is this needed?  The scheduler will get the alarm timeout from the server

src/scheduler/fifo.c
```diff
@@ -2474,6 +2474,13 @@ sched_settings_frm_svr(struct batch_status *status)
 				if (*endp != '\0')
 					goto cleanup;
 				*log_event_mask = mask;
+			} else if (!strcmp(attr->name, ATTR_sched_server_dyn_res_alarm)) {
+				char *endp;
+				long alarm;
```
bhroam: alarm is a function name, use a different variable name.

src/scheduler/data_types.h
```diff
@@ -1190,6 +1190,7 @@ struct config
 	struct sort_info *non_prime_node_sort;	/* node sorting non primetime */
 	struct dyn_res dynamic_res[MAX_SERVER_DYN_RES]; /* for server_dyn_res */
 	struct peer_queue peer_queues[NUM_PEERS];/* peer local -> remote queue map */
+	long server_dyn_res_alarm;		/* alarm for server_dyn_res scripts */
```
bhroam: This is a global, it isn't needed here.

src/scheduler/server_info.c
```diff
@@ -704,21 +712,91 @@ query_server_dyn_res(server_info *sinfo)
...
+				break;
+			case 0:		/* child */
+				close(pdes[0]);
+				if (pdes[1] != STDOUT_FILENO) {
+					dup2(pdes[1], STDOUT_FILENO);
+					close(pdes[1]);
+				}
+				setpgid(0, 0);
+				if (sigemptyset(&allsigs) == -1) {
+					perror("sigemptyset");
```
bhroam: This is a daemon.  There is no stderr.  Write to the sched log

src/scheduler/server_info.c
```diff
@@ -704,21 +712,91 @@ query_server_dyn_res(server_info *sinfo)
...
+				}
+
+				char *argv[4];
+				argv[0] = "/bin/sh";
+				argv[1] = "-c";
+				argv[2] = conf.dynamic_res[i].command_line;
+				argv[3] = NULL;
+
+				execve("/bin/sh", argv, environ);
+				_exit(127);
```
bhroam: Why _exit() instead of exit()?

src/scheduler/server_info.c
```diff
@@ -704,21 +712,91 @@ query_server_dyn_res(server_info *sinfo)
...
+				}
+				if (sigprocmask(SIG_SETMASK, &allsigs, NULL) == -1) {	/* unblock all signals */
+					perror("sigprocmask");
+					exit(1);
+				}
+
+				char *argv[4];
+				argv[0] = "/bin/sh";
+				argv[1] = "-c";
+				argv[2] = conf.dynamic_res[i].command_line;
```
bhroam: Will this work with arguments to the server_dyn_res script?

src/scheduler/server_info.c
```diff
@@ -704,21 +712,91 @@ query_server_dyn_res(server_info *sinfo)
...
+				argv[3] = NULL;
+
+				execve("/bin/sh", argv, environ);
+				_exit(127);
+			}
+
+			FD_ZERO(&set);
+			FD_SET(pdes[0], &set);
+			log_eventf(PBSEVENT_ERROR, PBS_EVENTCLASS_SERVER, LOG_DEBUG, "server_dyn_res",
+				"timeout is %d", server_dyn_res_alarm);
```
bhroam: This is a debug message.  Remove it.

src/scheduler/server_info.c
```diff
@@ -704,21 +712,91 @@ query_server_dyn_res(server_info *sinfo)
...
+				ret = select(FD_SETSIZE, &set, NULL, NULL, &timeout);
+			} else {
+				ret = select(FD_SETSIZE, &set, NULL, NULL, NULL);
+			}
+			if (ret == -1) {
+				log_eventf(PBSEVENT_ERROR, PBS_EVENTCLASS_SERVER, LOG_DEBUG, "server_dyn_res",
+				"Select() failed for script %s", conf.dynamic_res[i].command_line);
+			} else if (ret == 0) {
+				log_eventf(PBSEVENT_DEBUG, PBS_EVENTCLASS_SERVER, LOG_DEBUG, "server_dyn_res",
+				"Program %s timed out", conf.dynamic_res[i].command_line);
```
bhroam: Say that you are setting the resource to zero.

src/scheduler/server_info.c
```diff
@@ -704,21 +712,91 @@ query_server_dyn_res(server_info *sinfo)
...
+				}
 			}
 
-			if (((fp = popen(conf.dynamic_res[i].command_line, "r")) == NULL) ||
-				(fgets(buf, 256, fp) == NULL)) {
+			/* Parent; assume fdopen can't fail. */
+			fp = fdopen(pdes[0], "r");
+			close(pdes[1]);
+
+			if (fgets(buf, 256, fp) == NULL) {
```
bhroam: Use sizeof(buf)

src/server/req_manager.c
```diff
@@ -79,6 +79,8 @@
 
 #include <arpa/inet.h>
 
+#include "pbs_error.h"
+
```
bhroam: Do you need this?  You didn't modify anything else.

test/tests/functional/pbs_svr_dyn_res.py
```diff
@@ -521,6 +521,69 @@ def test_res_boolean_invalid_input(self):
...
+        resname = ["foo"]
+        restype = ["boolean"]
+
+        # Prep for server_dyn_resource script
+        resval = ["sleep 60\necho true"]
+
+        filenames = self.setup_dyn_res(resname, restype, resval)
+
+        # Submit job
+        a = {'Resource_List.foo': '"true"'}
```
bhroam: Why do you need the extra quotes?

test/tests/functional/pbs_svr_dyn_res.py
```diff
@@ -521,6 +521,69 @@ def test_res_boolean_invalid_input(self):
...
+        j = Job(TEST_USER, attrs=a)
+        jid = self.server.submit(j)
+
+        self.logger.info('sleeping 30 seconds to wait for script to timeout')
```
bhroam: Capitalize sleeping

test/tests/functional/pbs_svr_dyn_res.py
```diff
@@ -521,6 +521,69 @@ def test_res_boolean_invalid_input(self):
...
+        time.sleep(30)
+        self.scheduler.log_match("%s timed out" % filenames[0])
+        self.scheduler.log_match("Setting resource foo to 0")
+
+        # The job shouldn't run
+        job_comment = "Can Never Run: Insufficient amount of server resource:"
+        job_comment += " foo (True != False)"
+        a = {'job_state': 'Q', 'comment': job_comment}
+        self.server.expect(JOB, a, id=jid, attrop=PTL_AND)
```
bhroam: PTL_AND is default.

test/tests/functional/pbs_svr_dyn_res.py
```diff
@@ -521,6 +521,69 @@ def test_res_boolean_invalid_input(self):
...
+        # Prep for server_dyn_resource script
+        resval = ["sleep 20\necho true"]
+
+        filenames = self.setup_dyn_res(resname, restype, resval)
+
+        self.server.manager(MGR_CMD_SET, SCHED,
+                            {ATTR_sched_server_dyn_res_alarm: 10})
+
+        # Submit job
+        a = {'Resource_List.foo': '"true"'}
```
bhroam: Same as above, why do you need the extra quotes?

src/scheduler/server_info.c
```diff
@@ -704,21 +712,91 @@ query_server_dyn_res(server_info *sinfo)
...
-				log_eventf(PBSEVENT_SECURITY, PBS_EVENTCLASS_SERVER, LOG_ERR, "server_dyn_res", 
+				log_eventf(PBSEVENT_SECURITY, PBS_EVENTCLASS_SERVER, LOG_ERR, "server_dyn_res",
 					"error: %s file has a non-secure file access, setting resource %s to 0, errno: %d",
 					filename, res->name, err);
-				(void) set_resource(res, res_zero, RF_AVAIL);
+				set_resource(res, res_zero, RF_AVAIL);
+				continue;
+			}
+
+			if (pipe(pdes) < 0) {
```
bhroam: Any chance you can put the fork() and pipe() code in a separate function?  As an added benefit, you can get rid of the goto which we don't use often in the scheduler.

test/tests/functional/pbs_svr_dyn_res.py
vstumpf: Copy-pasted it from another test, I've removed it

src/scheduler/server_info.c
vstumpf: It says it gets set to zero on line 820.

src/scheduler/server_info.c
vstumpf: exit() will call exit handlers, which could disrupt the parent process.
See: https://stackoverflow.com/questions/5422831/what-is-the-difference-between-using-exit-exit-in-a-conventional-linux-fo

src/scheduler/server_info.c
vstumpf: It does work

src/scheduler/server_info.c
vstumpf: I'd rather not, I'd have to keep track of too many variables (pipes, pid, errno).
I guess I can put everything in if blocks, I'll try that.

src/scheduler/server_info.c
vstumpf: Actually, it even gets set to zero down there, so I don't need to set it up here.

src/scheduler/server_info.c
```diff
@@ -704,21 +712,87 @@ query_server_dyn_res(server_info *sinfo)
 			/* Make sure file does not have open permissions */
 			err = tmp_file_sec(filename, 0, 1, S_IWGRP|S_IWOTH, 1);
 			if (err != 0) {
-				log_eventf(PBSEVENT_SECURITY, PBS_EVENTCLASS_SERVER, LOG_ERR, "server_dyn_res", 
+				log_eventf(PBSEVENT_SECURITY, PBS_EVENTCLASS_SERVER, LOG_ERR, "server_dyn_res",
 					"error: %s file has a non-secure file access, setting resource %s to 0, errno: %d",
 					filename, res->name, err);
-				(void) set_resource(res, res_zero, RF_AVAIL);
+				set_resource(res, res_zero, RF_AVAIL);
+				continue;
```
arungrover: Thanks for fixing this bug :)

PR #1475, opened 2020-01-16
Added readme file for ci

sandisamp:
#### Describe Bug or Feature

Added readme file for ci folder which describes usage of it's CLI, dependencies and setup

### DIFF
--- a/ci/Readme.md
+++ b/ci/Readme.md
@@ -1,0 +1,98 @@
+Instant-CI is a developer tool which aims at providing continous integration to the developers locally on their development systems.
+It runs a container which will be running in the background on the users system, the container will pick up on changes done by the user ( via commits or manual triggers) and has the capability to build install and 
+run PTL tests on the pbs of the users repository, for all this the user may not worry about underlying dependencies or any caveats.
+It also supports build and test history in the form of logs.
+
+Dependincies for this tool are:
+* python3.5 or above
+* docker
+* docker-compose
+
+***How to setup:***
+
+Simply invoke the following command:
+
+` ./ci --start`
+
+***CLI interface for ci:***
+
+* **./ci --start:** This will behave as the setup for ci containers to run in the background. User can provide desired platform ,which must be present in docker hub. Currently supports ubuntu,centos and opensuse.
+```bash
+./ci --start
+ 
+# The above command will run ci with centos:7 as it's base container.
+ 
+./ci --start="ubuntu"
+ 
+# This will run ci with container as "ubuntu", all docker hub images are compatible with ci.
+ 
+./ci --start="opensuse/leap:15" --build
+ 
+# The above command will force the container to rebuild all dependencies.
+```
+
+* **./ci --configure:** user can provide PBS configure options.
+
+```bash
+./ci --configure
+  
+# The above command will run the pbs configure with default options "--prefix=/opt/pbs --enable-ptl" .
+# Note that if you have already specified the configure options and are running it again then 
+# simply calling --configure will run the last called options
+  
+./ci --configure='CFLAGS="-g -O2 -Wall -Werror" --prefix=/tmp/pbs --enable-ptl'
+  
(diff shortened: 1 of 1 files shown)


ci/Readme.md
```diff
@@ -0,0 +1,98 @@
+Instant-CI is a developer tool which aims at providing continous integration to the developers locally on their development systems.
+It runs a container which will be running in the background on the users system, the container will pick up on changes done by the user ( via commits or manual triggers) and has the capability to build install and
```
bhroam: Remove the space after then (

ci/Readme.md
```diff
@@ -0,0 +1,98 @@
...
+
+***How to setup:***
+
+Simply invoke the following command:
+
+` ./ci --start`
+
+***CLI interface for ci:***
+
+* **./ci --start:** This will behave as the setup for ci containers to run in the background. User can provide desired platform ,which must be present in docker hub. Currently supports ubuntu,centos and opensuse.
```
bhroam: Move the space from before the comma to after the comma

ci/Readme.md
```diff
@@ -0,0 +1,98 @@
...
+* **./ci --configure:** user can provide PBS configure options.
+
+```bash
+./ci --configure
+  
+# The above command will run the pbs configure with default options "--prefix=/opt/pbs --enable-ptl" .
+# Note that if you have already specified the configure options and are running it again then 
+# simply calling --configure will run the last called options
+  
+./ci --configure='CFLAGS="-g -O2 -Wall -Werror" --prefix=/tmp/pbs --enable-ptl'
```
bhroam: There's no real reason to use -g.  It's for debugging and we will never be debugging PBS inside the container.

ci/Readme.md
```diff
@@ -0,0 +1,98 @@
...
+```
+
+* **./ci --install:** It installs pbs and ptl into the container. For this to run the source must be compiled, it will throw an error if this is not the case.
+```bash
+./ci --install
+ 
+# this will invoke 'make install' for pbs
+```
+
+* **./ci --test:** This command will be used to run PTL test cases inside the container. By default it will run "–smoke" tag for PTL, else the user can specify what pbs_benchpress options they need.
```
bhroam: Do you mean '-smoke' or just 'smoke' ?

ci/Readme.md
```diff
@@ -0,0 +1,98 @@
...
+ 
+# The above method are examples how to call a custom test case, 
+# keep in mind that benchpress is called from the installed 'tests' dir of PTL
+```
+* **./ci --run:** This command will manually trigger the build for automatic run of 'build install and test', with last specified options for configure and test (default if no options were specified previously).
+
+```bash
+./ci --run
+ 
+# This command will invoke configure(with default options) ,make, install and test.
```
bhroam: Move the space from before the comma to after it.

ci/Readme.md
```diff
@@ -0,0 +1,98 @@
...
+./ci --run
+ 
+# This command will invoke configure(with default options) ,make, install and test. 
+# With no options specified this will run "--tags=smoke" for ptl, 
+# however if you have already ran this command before it will run the last specified options
+ 
+./ci --run="-f pbs_smoketest.py"
+./ci --run="--tags=smoke"
+ 
+#the above method are examples how to call a custom test case, keep in mind that benchpress is called from the installed 'tests' dir of PTL
```
bhroam: Add a space after the # to be consistent with the rest of the comments in the file.

ci/Readme.md
```diff
@@ -0,0 +1,98 @@
...
+```
+
+* **./ci --delete:** which will stop any ongoing build in the container for the moment and remove it build.
+
+```bash
+./ci --delete
+ 
+#If you want to get rid of the container simply invoke this command.
```
bhroam: Same as above, add a space after the \#

PR #1476, opened 2020-01-16
Fix expect in perf test

tpai26:
#### Describe Bug or Feature
The bug can be caused in expect() where it checks the number of jobs in running state. expect() by default has max_attempts=60 and triggers scheduling cycle after every attempt

These tests look at cycle time after setting scheduling to false. Also default attempts in expect is 60 which might be less than number required for all jobs to go into running state. Therefore adding trigger_sched_cycle=False and max_attempts=240 to expect

Also divided _test_run_many_jobs_ into two tests _test_run_many_normal_jobs_ , _test_run_many_bucket_jobs_ -  so that it takes less time to run the tests and the test does not time out.

[test_run_many_bucket_jobs.txt](https://github.com/PBSPro/pbspro/files/4073929/test_run_many_bucket_jobs.txt)
[test_run_many_normal_jobs.txt](https://github.com/PBSPro/pbspro/files/4073931/test_run_many_normal_jobs.txt)
[test_pset_fuzzy_perf.txt](https://github.com/PBSPro/pbspro/files/4073932/test_pset_fuzzy_perf.txt)

### DIFF
--- a/test/tests/performance/pbs_sched_perf.py
+++ b/test/tests/performance/pbs_sched_perf.py
@@ -163,31 +163,36 @@ class TestSchedPerf(TestPerformance):
         self.compare_normal_path_to_buckets('free', num_jobs)
 
     @timeout(3600)
-    def test_run_many_jobs(self):
+    def test_run_many_normal_jobs(self):
         """
-        Submit many jobs and time the cycle that runs all of them.
+        Submit many normal path jobs and time the cycle that runs all of them.
         """
         num_jobs = 10000
         a = {'Resource_List.select': '1:ncpus=1'}
-        jids = self.submit_jobs(a, num_jobs, wt_start=1000)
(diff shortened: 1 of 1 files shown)


test/tests/performance/pbs_sched_perf.py
```diff
@@ -163,31 +163,36 @@ def test_node_bucket_perf_free(self):
...
-    def test_run_many_jobs(self):
+    def test_run_many_normal_jobs(self):
         """
-        Submit many jobs and time the cycle that runs all of them.
+        Submit many normal path jobs and time the cycle that runs all of them.
         """
         num_jobs = 10000
         a = {'Resource_List.select': '1:ncpus=1'}
-        jids = self.submit_jobs(a, num_jobs, wt_start=1000)
+        jids = self.submit_jobs(a, num_jobs, wt_start=10000)
```
bhroam: use num_jobs instead of 10000 (and below)

PR #1478, opened 2020-01-17
Reconfirm running degraded reservations

bhroam:
#### Describe Bug or Feature
When a reservation is degraded, PBS will try and reconfirm it up until it starts running.  The RFE changes that.  We now reconfirm reservations while they are running.

#### Describe Your Change
The time between reconfirmation attempts has changed.  Before it would wait until until the init time passed, then do a reconfirmation attempt every half the time between the previous attempt and the start.  If it hit the cutoff, it would stop.  Now there is a new variable reserve_retry_time that determines the length of time between attempts.  Most of the changes in the server are dealing with this.

The scheduler now has to deal with the case when a reservation is running.  It needs to make sure to keep all nodes that are up and replace only the down ones.  To this end, we create a hybrid execselect with some chunks with nodes and other chunks without.  This forces the scheduler to keep the nodes the reservation has and search for other nodes for the down ones.  Finally if a running job is on a downed node, the reservation won't be reconfirmed until that job finishes or is requeued.

TestReservations.test_standing_resv_with_job_array was broken.  I fixed it so it would pass.

#### Link to Design Doc
https://pbspro.atlassian.net/wiki/spaces/PD/pages/1401815043/Reconfirming+degraded+reservations+that+are+running

I ran the reservation test suite with the server and scheduler in valgrind.  I did the best I could parsing through the output.  Due to the massive amount of python warnings, it makes it hard to determine if there is anything wrong.  I don't think there is.

Here is the output from running TestReservations: 
[resv.log](https://github.com/PBSPro/pbspro/files/4100729/resv.log)

I did a TH3 regression and there were some failures, but nothing to do with my code.  Most of them were failing in master.


bhroam:
Travis failed due to a few smoke tests failing during setup.  I'm looking into it.

src/server/svr_func.c
```diff
@@ -664,8 +662,8 @@ cred_name_okay(attribute *pattr, void *pobj, int actmode)
 
 /**
  * @brief
- * 		set_reserve_retry_init - action routine for the server's
- * 		"reserve_retry_init" attribute.
+ * 		action_resv_retry_time - action routine for the server's
+ * 		"resv_retry_time" attribute.
```
arungrover: Please rename it to reserve_retry_time since that's the name of the attribute.

test/fw/ptl/lib/pbs_api_to_cli.py
```diff
@@ -241,7 +241,7 @@
     ATTR_job_sort_formula: 'job_sort_formula',
     ATTR_EligibleTimeEnable: 'eligible_time_enable',
     ATTR_resv_retry_init: 'reserve_retry_init',
-    ATTR_resv_retry_cutoff: 'reserve_retry_cutoff',
+    ATTR_resv_retry_time: 'resv_retry_time',
```
arungrover: It should be reserve_retry_time here and in pbs_ifl_mock.py

test/tests/functional/pbs_reservations.py
```diff
@@ -85,56 +86,324 @@ def submit_asap_reservation(self, user, jid):
...
+        a = {'reserve_retry_time': 5}
         self.server.manager(MGR_CMD_SET, SERVER, a)
 
-        a = {'resources_available.ncpus': 4}
+        a = {'resources_available.ncpus': 1}
         self.server.create_vnodes('vn', a, num=2, mom=self.mom)
 
         now = int(time.time())
 
         # submitting 25 seconds from now to allow some of the older testbed
```
arungrover: you might want to change this comment as the start and end time is now passed in by the caller.

test/tests/functional/pbs_reservations.py
```diff
@@ -85,56 +86,324 @@ def submit_asap_reservation(self, user, jid):
...
-                           attrop=PTL_AND)
+        if run:
+            a = {'reserve_substate': 5}
+        else:
+            a = {'reserve_substate': 2}
+        a.update({'resv_nodes': (MATCH_RE, re.escape(other_node))})
+
+        if not run:
+            resv_state = {'reserve_state': (MATCH_RE, 'RESV_CONFIRMED|2')}
+        self.server.expect(RESV, a, id=rid, interval=1)
```
arungrover: I think we should set trigger_sched_cycle option as false, just to make sure that the cycle is triggered based on the reserve_retry_time and not because of expect triggering it,

test/tests/functional/pbs_reservations.py
```diff
@@ -85,56 +86,324 @@ def submit_asap_reservation(self, user, jid):
...
+            else:
+                resv_state = {'reserve_state': (MATCH_RE, 'RESV_RUNNING|5'),
+                              'reserve_substate': 5}
+        else:
+            resv_state = {'reserve_state': (MATCH_RE, 'RESV_CONFIRMED|2'),
+                          'reserve_substate': 2}
+
+        self.server.expect(RESV, resv_state, id=rid)
+
+    def degraded_resv_node_down_back(self, start, end, rrule=None, run=False):
```
arungrover: This function is not needed

test/tests/functional/pbs_reservations.py
```diff
@@ -85,56 +86,324 @@ def submit_asap_reservation(self, user, jid):
...
+        Verify that degraded standing reservations are reconfirmed
+        on other nodes
+        """
+        t = int(time.time())
+        self.degraded_resv_reconfirm(start=t + 25, end=t + 625,
+                                     rrule='freq=HOURLY;count=5', run=True)
+
+    def test_degraded_advanced_running_reservations(self):
+        """
+        Verify that degraded advanced reservations are reconfirmed
```
arungrover: I think the description here is wrong, such a reservation will not be reconfirmed.

doc/man1/pbs_server_attributes.7B
```diff
@@ -1541,30 +1541,6 @@ is any built-in or custom resource
...
-.I int
-.br
-Valid values: Must be greater than 
-.I zero
-.br
-Default: 
-.I 7200 
-(2 hours)
-
 .IP reserve_retry_init 8
```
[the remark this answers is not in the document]
bhroam: I'm not exactly sure why I modified the man pages.  They'll be overridden when Anne documents them for 2020.  I'll leave them alone.

src/scheduler/queue_info.c
```diff
@@ -968,9 +968,18 @@ dup_queue_info(queue_info *oqinfo, server_info *nsinfo)
 
 	if (oqinfo->resv != NULL) {
 		nqinfo->resv = find_resource_resv_by_indrank(nsinfo->resvs, oqinfo->resv->resresv_ind, oqinfo->resv->rank);
-		/* just incase we we didn't set the reservation cross pointer */
-		if (nqinfo->resv != NULL && nqinfo->resv->resv != NULL)
-			nqinfo->resv->resv->resv_queue = nqinfo;
+		if (!nqinfo->resv->resv->is_standing) {
+			/* just incase we we didn't set the reservation cross pointer */
+			if (nqinfo->resv != NULL && nqinfo->resv->resv != NULL)
```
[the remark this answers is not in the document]
vstumpf: Yeah, if nqinfo->resv or nqinfo->resv->resv is null, the scheduler will have already crashed before accessing this.

test/tests/functional/pbs_reservations.py
bhroam: It won't matter if it is based on reserve_retry_time or not.  The scheduler will ignore a reservation whose retry time is not up.  There really is no reason to run cycles through.

src/scheduler/resv_info.c
```diff
@@ -1612,44 +1615,52 @@ confirm_reservation(status *policy, int pbs_sd, resource_resv *unconf_resv, serv
...
  * @param[out]	tot_vnodes	-	the total number of vnodes associated to the reservation
- * @param[in,out]	names_of_down_vnodes	-	the string of vnodes that are down that are then
+ * @param[out]	names_of_down_vnodes	-	the string of vnodes that are down that are then
  * 												printed into the log. This has to be of maximum length MAXVNODELIST
  *
  * @return	the number of vnodes down. The total number of vnodes is passed back
  * 			by reference.
- * @retval	-1	: on error.
+ * @retval	-1	: there is a running job on one of the unavailable nodes.
+ * @retbal	-2	: error
```
vstumpf: should be retval

src/server/node_manager.c
```diff
@@ -8873,3 +8817,25 @@ set_resv_for_degrade(struct pbsnode *pnode, resc_resv *presv)
...
+	long retry;
+	long resv_start = presv->ri_wattr[RESV_ATR_start].at_val.at_long;
+
+	if (time_now < resv_start && time_now + resv_retry_time > resv_start)
+		retry = resv_start;
+	else
+		retry = time_now + resv_retry_time;
+
+	return retry;
+}
```
vstumpf: Add a newline at this line please.

src/server/req_rescq.c
```diff
@@ -548,34 +547,17 @@ req_confirmresv(struct batch_request *preq)
 	 */
 	if (strcmp(preq->rq_extend, PBS_RESV_CONFIRM_FAIL) == 0) {
 		if (is_degraded && !is_being_altered) {
-			long degraded_time = presv->ri_degraded_time;
+			long retry_time;
 			DBPRT(("degraded_time of %s is %s", presv->ri_qs.ri_resvID, ctime(&degraded_time)));
```
vstumpf: Might as well remove this, degraded_time is removed.

test/tests/functional/pbs_reservations.py
```diff
@@ -85,56 +86,324 @@ def submit_asap_reservation(self, user, jid):
...
+        """
+        a = {'reserve_retry_time': 5}
+        self.server.manager(MGR_CMD_SET, SERVER, a)
+
+        a = {'resources_available.ncpus': 1}
+        self.server.create_vnodes('vn', a, num=2, mom=self.mom)
+
+        now = int(time.time())
+
+        # submitting 25 seconds from now to allow some of the older testbed
```
vstumpf: Also remove this comment.

bhroam:
@arungrover @vstumpf I've resolved the conflicts.  I ran another TH3 run and picked up a new failure.  It came from when I rebased with master.  The same test fails on master.  I think this PR is ready for merge.

PR #1490, opened 2020-01-29
Fix cgroups not cleaned up on checkpoint_abort

bayucan:
#### Describe Bug or Feature

* With cgroups hook enabled, when scheduler preempts a lower priority multi-node job causing the mom to do a checkpoint_abort, the high priority job does not run.
* The reason being the cgroups hook did not clean up the cgroups resources on the secondary node of the preempted job.
* The high priority job fails on execjob_begin pbs_cgroups hook on the sister node, complaining about "CgroupProcessingError...Failed to assign resources"

#### Describe Your Change

* Have the execjob_abort hook execute when mom does a successful IM_CHECKPOINT_ABORT (checkpoint job then kill job).
* In this way, the execjob_abort side of pbs_cgroups hook would be called which does the cgroups clean up.
* It was a bug in the initial implementation of execjob_abort hook to not be called on a mom checkpoint_job abort request.

* [ptl.test_checkpoint_abort_preemption.FAIL.txt](https://github.com/PBSPro/pbspro/files/4130800/ptl.test_checkpoint_abort_preemption.FAIL.txt)
* [ptl.test_checkpoint_abort_preemption.PASS.2.txt](https://github.com/PBSPro/pbspro/files/4201186/ptl.test_checkpoint_abort_preemption.PASS.2.txt)
* [ptl.cgroups.txt](https://github.com/PBSPro/pbspro/files/4130802/ptl.cgroups.txt)
* Valgrind Sister mom: [valgrind.td-08.sister.txt](https://github.com/PBSPro/pbspro/files/4201190/valgrind.td-08.sister.txt)

* Valgrind MS mom: [valgrind.td-09.ms.txt](https://github.com/PBSPro/pbspro/files/4201192/valgrind.td-09.ms.txt)

### DIFF
--- a/src/resmom/requests.c
+++ b/src/resmom/requests.c
@@ -3895,6 +3895,26 @@ post_chkpt(job *pjob, int  ev)
 		 **	obit is sent.
 		 */
 		if (abort) {
+			mom_hook_input_t	hook_input;
+			mom_hook_output_t	hook_output;
+			int			hook_errcode = 0;
+			hook			*last_phook = NULL;
+			unsigned int		hook_fail_action = 0;
+			char			hook_msg[HOOK_MSG_SIZE+1];
+
+			mom_hook_input_init(&hook_input);
+			hook_input.pjob = pjob;
+
+			mom_hook_output_init(&hook_output);
+			hook_output.reject_errcode = &hook_errcode;
+			hook_output.last_phook = &last_phook;
+			hook_output.fail_action = &hook_fail_action;
+
+			(void)mom_process_hooks(HOOK_EVENT_EXECJOB_ABORT,
+						PBS_MOM_SERVICE_NAME, mom_host,
+						&hook_input, &hook_output, hook_msg,
+						sizeof(hook_msg), 1);
+
 			exiting_tasks = 1;
 			term_job(pjob);
 		} else if (pjob->ji_preq) {
--- a/src/resmom/requests.c
+++ b/src/resmom/requests.c
@@ -3895,28 +3895,20 @@ post_chkpt(job *pjob, int  ev)
 		 **	obit is sent.
 		 */
 		if (abort) {
-			mom_hook_input_t	hook_input;
-			mom_hook_output_t	hook_output;
-			int			hook_errcode = 0;
-			hook			*last_phook = NULL;
-			unsigned int		hook_fail_action = 0;
-			char			hook_msg[HOOK_MSG_SIZE+1];
+			mom_hook_input_t  *hook_input = NULL;
 
-			mom_hook_input_init(&hook_input);
-			hook_input.pjob = pjob;
-
-			mom_hook_output_init(&hook_output);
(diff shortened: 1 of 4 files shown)


bhroam:
I'm not sure I agree the abort hook should fire.  It might be called checkpoint_abort, but it is not an error condition.  This is how the scheduler preempts via checkpointing.  I think the normal end hook should fire.

mike0042:
@bhroam - As I understand it, the scheduler effectively aborts the running job when it orders a preemption to take place. While it may not be an error, it is an exception to the normal job life cycle. For that reason, I think the execjob_abort is the correct event. I'm open to discussion.

bayucan:
@bhroam @mike0042 : In addition to what Mike said, I also feel the execjob_abort is the right one to execute, as I'm afraid some sites not using cgroups hook already have execjob_end hooks in place, and would be surprised to see that their hook be called under the checkpoint_abort mom action. The end hook may just be printing end of job statistics, and the job is officially not complete yet, and has been requeued. What I can do is update the execjob_abort hook design (https://pbspro.atlassian.net/wiki/spaces/PD/pages/1063878670/Execjob+abort+hook) to explicitly say that the abort hook will fire under a successful checkpoint_abort hook action.

bhroam:
What hook fires when a job is requeued?  This really should be handled in the same way.  The only difference between a checkpointed job and a requeued job is the fact that some attributes are kept on the server side.  If the execjob_end hook fires when a job is requeued, I think it should happen here.  To the mom there shouldn't be a difference between the two (except it runs a script to kill the processes instead of doing killing them itself).

bayucan:
Please don't review my latest commit yet. Switching to execjob_end hook is causing a sister mom crash. Perhaps it has something to do with the end hook now happening in the background.

src/resmom/requests.c
```diff
@@ -3895,8 +3895,20 @@ post_chkpt(job *pjob, int  ev)
 		 **	obit is sent.
 		 */
 		if (abort) {
-			exiting_tasks = 1;
-			term_job(pjob);
+			mom_hook_input_t  *hook_input = NULL;
+
+			hook_input = (mom_hook_input_t *)malloc(sizeof(mom_hook_input_t));
```
mike0042: Shouldn't we log an error if the malloc fails? Since post_chkpt() returns void, the caller won't know if anything has gone wrong.

bayucan: @mike0042: I'm not sure how important it is to log the error malloc failure. Sure it will tell the caller that malloc failed, but job would still be terminated. Also, even if the execjob_end hook execute successfully or not, the job would still reach its end point. pbs_mom would not prevent the job from being aborted, given that the checkpoint_abort script already told the job to exit. Let me know if you still feel that I should logged that malloc failure...

mike0042: @bayucan - I see your point. We do log a message from this function if readdir() fails. Besides, if we couldn't malloc a small amount of memory there are big problems happening on the system.

PR #1491, opened 2020-01-30
TestPbsNodeRampDown fails checking licenses

sandisamp:
#### Describe Bug or Feature

TestPbsNodeRampDown  uses the 'Used' part of the license_count attribute to determine if nodes have been released by the feature. In OSS, license_count Used doesn't change, it is always 0.
This means all tests that check this value will fail.

#### Describe Your Change

Before raising error, added code that checks if license used is equal to 0, and if this is true successfully return from the function.

[test_multi_release_nodes.txt](https://github.com/PBSPro/pbspro/files/4133584/test_multi_release_nodes.txt)

### DIFF
--- a/test/tests/functional/pbs_node_rampdown.py
+++ b/test/tests/functional/pbs_node_rampdown.py
@@ -166,6 +166,8 @@ class TestPbsNodeRampDown(TestFunctional):
                     break
             retry -= 1
             if retry == 0:
+                if str(lic_count).find("Used:0"):
+                    return
                 raise AssertionError("not found %d licenses" % (num_licenses,))
             self.logger.info("sleeping 3 secs before next retry")
             time.sleep(3)
--- a/test/tests/functional/pbs_node_rampdown.py
+++ b/test/tests/functional/pbs_node_rampdown.py
@@ -157,6 +157,8 @@ class TestPbsNodeRampDown(TestFunctional):
         for _ in range(n):
             server_stat = self.server.status(SERVER, 'license_count')
             lic_count = server_stat[0]['license_count']
+            if lic_count.find('Avail_Global:10000000 Avail_Local:10000000 Used:0') != -1:
+                return 
             for lic in lic_count.split():
                 lic_split = lic.split(':')
                 if lic_split[0] == 'Used':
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_node_rampdown.py
```diff
@@ -166,6 +166,8 @@ def license_count_match(self, num_licenses):
                     break
             retry -= 1
             if retry == 0:
+                if str(lic_count).find("Used:0"):
+                    return
```
bhroam: You probably could move this up to right after the split so we don't try for 5 times before ultimately returning.

sandisamp: Yes I considered doing that in the first place, but there could be a scenario where the "Used" count increases after the first attempt (if it takes time to update). In that case, the function will return prematurely and incorrectly.

bhroam: Now that I think about this, why not use a RE_MATCH and use expect()?  It'll handle this looping for you.  You can set a max_attempts=5 and interval=3.

sandisamp: * If we use expect() the condition will fail as soon as it's not satisfied and exit declaring the test as a failure. So we won't be able to check for the second condition in case the first isn't satisfied.
* A workaround is to use the 'try expect' block however Regex match is a feature which cant be found in server object's expect. So we will have to extract that attribute and check it after filtering out the 'used' attribute (o/p returned will be something like _got: license_count = Avail_Global:10000000 Avail_Local:10000000 Used:0 High_Use:0_ ) as is done in the current code.

bhroam: With the RE_MATCH, you can search for either license_count with 'Used: num_license | Used:0'.  This way if used is 0, then it'll stop right away.  The regexp is just a little more complicated.

vstumpf: Can we just do a check for license_count == `Avail_Global:10000000 Avail_Local:10000000 Used:0` at the beginning? If that's the case, then we can assume it's an OSS build that doesn't require licensing. And we wouldn't have to wait 15 seconds for each time we call license_count_match.

sandisamp: @bhroam If we use MATCH_RE operation there could be a situation where we get a false positive for OSS ( if at the first attempt the lic_count used is '0' but increases in further attempts)

I think the suggestion made by @vstumpf is the way to go... as we can match the exact string that would only be present in OSS thus eliminating any chances of false positives.

PR #1492, opened 2020-01-30
pbs_snapshot should use PTL library from pbs/unsupported dir

shilpakodli:
#### Describe Bug or Feature
 pbs_snapshot command is failing if PTL is installed in non PBS_EXEC path.

#### Describe Your Change
In pbs_snapshot script, we are by default assuming that PTL will be installed under "/opt/ptl/lib" location and this fails if we are installing PTL on non PBS_EXEC path. 
Also, run pbs_snapshot.py from PBS_EXEC/unsupported/fw/bin path if it is not found in /opt/ptl/bin path.

#### Test Logs:
[pbs_snapshot_capture_logs.txt](https://github.com/PBSPro/pbspro/files/4133738/pbs_snapshot_capture_logs.txt)


shilpakodli:
@prakashcv13 and @sujatapatnaik52 Thanks for the offline review comments. I have addressed review comments. Please review.

src/cmds/scripts/pbs_snapshot
```diff
@@ -84,13 +86,26 @@ fi
 
 export PYTHONPATH=${ptllibpath}:${PYTHONPATH}
 
-if [ -z $ptlbinpath ];then
-    echo "Error: PTL bin path not found"
-    exit 1
+if [ -z $PBS_EXEC ]; then
```
kjakkali: pbs.conf is already read in the above lines. we don't need read it again here.

src/cmds/scripts/pbs_snapshot
```diff
@@ -53,44 +53,48 @@ if [ ! -d  ${ptl_prefix_lib} ]; then
...
-                        fi
-                fi
-                unset __PBS_EXEC
-                unset PTL_PREFIX
-                unset conf
-                unset python_dir
+    if [ "X${__PBS_EXEC}" != "X" ]; then
+        # Define PATH and PYTHONPATH for the users
+        PTL_PREFIX=$(dirname ${__PBS_EXEC})/ptl
+        if [ -x "${PTL_PREFIX}/lib)/site-packages" ]; then
```
agrawalravi90: Instead of explicitly running pbs_snapshot.py, I suggest the following here to set ptlbinpath & ptllibpath correctly:
```suggestion
        if [ ! -x "${PTL_PREFIX}/lib/site-packages" ]; then
            ptlbinpath=${__PBS_EXEC}/sbin
            ptllibpath=${__PBS_EXEC}/unsupported/fw
        else
            python_dir=$(/bin/ls -1 ${PTL_PREFIX}/lib)/site-packages
            ptllibpath=${PTL_PREFIX}/lib/${python_dir}
            ptlbinpath=${PTL_PREFIX}/bin
        fi
```

src/cmds/scripts/pbs_snapshot
```diff
@@ -53,44 +53,48 @@ if [ ! -d  ${ptl_prefix_lib} ]; then
...
 export PYTHONPATH=${ptllibpath}:${PYTHONPATH}
 
 if [ -z $ptlbinpath ];then
-    echo "Error: PTL bin path not found"
-    exit 1
-fi
-
-if [ -x "${PBS_EXEC}/python/bin/python" ]; then
-	${PBS_EXEC}/python/bin/python ${ptlbinpath}/pbs_snapshot "${@}"
+    if [ -x "${__PBS_EXEC}/python/bin/python" ]; then
```
agrawalravi90: Don't run pbs_snapshot.py directly, please use the change that I suggested instead of making any of the following changes

src/cmds/scripts/pbs_snapshot
```diff
@@ -53,44 +53,48 @@ if [ ! -d  ${ptl_prefix_lib} ]; then
...
+            ptlbinpath=${PTL_PREFIX}/bin
+        fi
+       unset PTL_PREFIX
+       unset python_dir
+    fi
 fi
 
 export PYTHONPATH=${ptllibpath}:${PYTHONPATH}
 
 if [ -z $ptlbinpath ];then
```
agrawalravi90: it's better to change this to:
```suggestion
if [ ! -x $ptlbinpath ] || [ ! -x $ptllibpath ];then
```

And then echo that either ptl bin or lib path not found

src/cmds/scripts/pbs_snapshot
```diff
@@ -53,44 +53,48 @@ if [ ! -d  ${ptl_prefix_lib} ]; then
     fi
 fi
 
+conf="${PBS_CONF_FILE:-/etc/pbs.conf}"
```
agrawalravi90: If you do the change that I suggested, then you won't have to move this block of code

src/cmds/scripts/pbs_snapshot
shilpakodli: Hey @agrawalravi90, ptllibpath is set proper only if ptl is installed under /opt/ptl. Otherwise checking for ptllibpath doesn't make any sense. Because, we won't be able to find out exact user installed PTL location.
What I will do is I will just checkfor ptlbinpath and set only ptlbinpath if "/opt/ptl/lib/site-packages" not found.
Your thoughts on this?

shilpakodli:
@agrawalravi90 I have removed code which is not necessary to check PTL library path for different installations. 
Because, PBS comes with unsupported/fw/ which has ptl and bin directories. We should be able to run pbs_snapshot with this even when PTL is not installed.
Attaching logs where I have run pbs_snapshot when PTL is not installed.
[pbs_snapshot_capture_logs_review2.txt](https://github.com/PBSPro/pbspro/files/4138389/pbs_snapshot_capture_logs_review2.txt)

src/cmds/scripts/pbs_snapshot
```diff
@@ -38,59 +38,30 @@
...
-# Try rpm install paths
-if [ -f "/etc/debian_version" ]; then
-        ptl_prefix_lib=$(dpkg -L pbspro-ptl 2>/dev/null | grep -m 1 lib$ 2>/dev/null)
-else
-        ptl_prefix_lib=$(rpm -ql pbspro-ptl 2>/dev/null | grep -m 1 lib$ 2>/dev/null)
+conf="${PBS_CONF_FILE:-/etc/pbs.conf}"
+if [ -r "${conf}" ]; then
+    # we only need PBS_EXEC from pbs.conf
+    __PBS_EXEC=$( grep '^[[:space:]]*PBS_EXEC=' "$conf" | tail -1 | sed 's/^[[:space:]]*PBS_EXEC=\([^[:space:]]*\)[[:space:]]*/\1/' )
 fi
```
kjakkali: we need to add else part with exit here. As pbs_conf is not readable , there is not point in proceeding further

agrawalravi90:
@shilpakodli So, this is how the script was originally, it used to pick things up from PBS_EXEC/unsupported/fw, I had to change it to do what it does now in https://github.com/PBSPro/pbspro/pull/1438 because I encountered issues with it. I think 'fw' gets removed from unsupported/ when pbs is built with --enable-ptl and CFLAGS=--with ptl. Please check with @kjakkali  and @hirenvadalia  once.

shilpakodli:
> @shilpakodli So, this is how the script was originally, it used to pick things up from PBS_EXEC/unsupported/fw, I had to change it to do what it does now in #1438 because I encountered issues with it. I think 'fw' gets removed from unsupported/ when pbs is built with --enable-ptl and CFLAGS=--with ptl. Please check with @kjakkali and @hirenvadalia once.

I have tested with enabling ptl while build rpm in OSS. I also don't see unsupported/fw folder. So I putting back changes done by you. Thanks Ravi for this information.

src/cmds/scripts/pbs_snapshot
```diff
@@ -53,44 +53,52 @@ if [ ! -d  ${ptl_prefix_lib} ]; then
...
-                unset __PBS_EXEC
-                unset PTL_PREFIX
-                unset conf
-                unset python_dir
+    if [ "X${__PBS_EXEC}" != "X" ]; then
+        # Define PATH and PYTHONPATH for the users
+        PTL_PREFIX=$(dirname ${__PBS_EXEC})/ptl
+        if [ ! -x "${PTL_PREFIX}/lib/site-packages" ]; then
+            ptlbinpath=${__PBS_EXEC}/unsupported/fw/bin
+			ptllibpath=${__PBS_EXEC}/unsupported/fw
```
agrawalravi90: the indentation seems off here

src/cmds/scripts/pbs_snapshot
```diff
@@ -74,7 +74,7 @@ else
         # Define PATH and PYTHONPATH for the users
         PTL_PREFIX=$(dirname ${__PBS_EXEC})/ptl
         if [ ! -x "${PTL_PREFIX}/lib/site-packages" ]; then
-            ptlbinpath=${__PBS_EXEC}/unsupported/fw/bin
+			ptlbinpath=${__PBS_EXEC}/unsupported/fw/bin
```
agrawalravi90: i think you should reduce the indentation of this block, match it with the corresponding else block

shilpakodli: I have corrected the indentation now. Please review.

PR #1496, opened 2020-02-04
Queue resources becomes negative and remains even after unset the attribute

Bhagat-Rajput:
#### Describe Bug or Feature
Queue resources_assigned goes negative because we unset a resource which is already been occupied by the job.
**Steps to reproduce:**
1. set q \<queue_name\> resources_available.ncpus=\<value\>
2. run a job
3. unset q <qeueu_name> resources_available.ncpus
4. qdel the job
5. qstat -Qf \<queue_name\> and see "resources_assigned" value
#### Describe Your Change

If we want to unset the resc, we will check it's **resources_assigned** value. If the value is **zero** then we will remove both the attribute **resources_available** and **resources_assigned** else we will remove only **resources_available** attribute. Once **resources_assigned** value becomes zero, PBS will search for the corresponding **resources_available** attribute and if not found, it will remove **resources_assigned** attribute as well else leave both the attributes as it is.

PTL logs: [ptl_pbs_que_resc_usage.txt](https://github.com/PBSPro/pbspro/files/4213115/ptl_pbs_que_resc_usage.txt)

### DIFF
--- a/src/server/req_manager.c
+++ b/src/server/req_manager.c
@@ -1127,7 +1127,10 @@ mgr_unset_attr(attribute *pattr, attribute_def *pdef, int limit, svrattrl *plist
 							if (i >= 0) {
 								if ((pattr+i)->at_flags & ATR_VFLAG_SET) {
 									resource *nresc;
-									if ((nresc = find_resc_entry((pattr+i), prsdef)) != NULL) {
+									/* if resources_assigned value is non-zero then don't unset it, otherwise it
+									 * would appear as negative in qstat once job is finished.
+									 */
+									if (((nresc = find_resc_entry((pattr+i), prsdef)) != NULL) && (nresc->rs_value.at_val.at_long == 0)) {
 										nresc->rs_defin->rs_free(&nresc->rs_value);
 										delete_link(&nresc->rs_link);
 										free(nresc);
--- a/src/server/req_manager.c
+++ b/src/server/req_manager.c
@@ -1135,10 +1135,10 @@ mgr_unset_attr(attribute *pattr, attribute_def *pdef, int limit, svrattrl *plist
 											nresc->rs_defin->rs_free(&nresc->rs_value);
 											delete_link(&nresc->rs_link);
 											free(nresc);
-											nresc = (resource *)GET_NEXT((pattr+i)->at_val.at_list);
+											nresc = (resource *)GET_NEXT((pattr + i)->at_val.at_list);
 											if (nresc == NULL)
-												(pattr+i)->at_flags &= ~ATR_VFLAG_SET;
-											(pattr+i)->at_flags |= ATR_VFLAG_MODCACHE|ATR_VFLAG_MODIFY;
+												(pattr + i)->at_flags &= ~ATR_VFLAG_SET;
+											(pattr + i)->at_flags |= ATR_VFLAG_MODCACHE|ATR_VFLAG_MODIFY;
 										}
 									}
 								}
--- a/src/server/req_manager.c
+++ b/src/server/req_manager.c
@@ -1127,17 +1127,19 @@ mgr_unset_attr(attribute *pattr, attribute_def *pdef, int limit, svrattrl *plist
 							if (i >= 0) {
 								if ((pattr+i)->at_flags & ATR_VFLAG_SET) {
 									resource *nresc;
-									/* if resources_assigned value is non-zero then don't unset it, otherwise it
-									 * would appear as negative in qstat once job is finished.
-									 */
-									if (((nresc = find_resc_entry((pattr+i), prsdef)) != NULL) && (nresc->rs_value.at_val.at_long == 0)) {
-										nresc->rs_defin->rs_free(&nresc->rs_value);
-										delete_link(&nresc->rs_link);
-										free(nresc);
-										nresc = (resource *)GET_NEXT((pattr+i)->at_val.at_list);
-										if (nresc == NULL)
-											(pattr+i)->at_flags &= ~ATR_VFLAG_SET;
-										(pattr+i)->at_flags |= ATR_VFLAG_MODCACHE|ATR_VFLAG_MODIFY;
+									if ((nresc = find_resc_entry((pattr + i), prsdef)) != NULL) {
+										/* if resources_assigned value is non-zero then don't unset it, otherwise it
+										 * would appear as negative in qstat once job is finished.
+										 */
+										if (check_resc_assign_val(nresc)) {
+											nresc->rs_defin->rs_free(&nresc->rs_value);
+											delete_link(&nresc->rs_link);
+											free(nresc);
+											nresc = (resource *)GET_NEXT((pattr+i)->at_val.at_list);
+											if (nresc == NULL)
+												(pattr+i)->at_flags &= ~ATR_VFLAG_SET;
+											(pattr+i)->at_flags |= ATR_VFLAG_MODCACHE|ATR_VFLAG_MODIFY;
+										}
 									}
 								}
 							}
@@ -1194,7 +1196,6 @@ mgr_unset_attr(attribute *pattr, attribute_def *pdef, int limit, svrattrl *plist
 		indirect_target_check(0);
 	return (0);
 }
-
 
 /**
  * @brief
--- a/test/tests/functional/pbs_que_resc_usage.py
+++ b/test/tests/functional/pbs_que_resc_usage.py
@@ -1,0 +1,180 @@
+# coding: utf-8
+
+# Copyright (C) 1994-2020 Altair Engineering, Inc.
+# For more information, contact Altair at www.altair.com.
+#
+# This file is part of the PBS Professional ("PBS Pro") software.
+#
+# Open Source License Information:
+#
+# PBS Pro is free software. You can redistribute it and/or modify it under the
+# terms of the GNU Affero General Public License as published by the Free
+# Software Foundation, either version 3 of the License, or (at your option) any
+# later version.
+#
+# PBS Pro is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.
+# See the GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+# Commercial License Information:
+#
+# For a copy of the commercial license terms and conditions,
+# go to: (http://www.pbspro.com/UserArea/agreement.html)
+# or contact the Altair Legal Department.
+#
+# Altair’s dual-license business model allows companies, individuals, and
+# organizations to create proprietary derivative works of PBS Pro and
+# distribute them - whether embedded or bundled with other software -
+# under a commercial license agreement.
+#
+# Use of Altair’s trademarks, including but not limited to "PBS™",
+# "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
+# trademark licensing policies.
+
+from tests.functional import *
+import time
+
+
+class TestQueRescUsage(TestFunctional):
+    """
+    Test built-in and custom resource behavior after
+    set/unset/assign/release resource on the queue
+    """
+    err_msg = 'Failed to get the expected resource value/result'
+
+    def setUp(self):
+        TestFunctional.setUp(self)
+        self.server.manager(MGR_CMD_SET, NODE, {
+                            'resources_available.ncpus': 5},
+                            self.mom.shortname)
+        a = {'queue_type': 'execution', 'enabled': True, 'started': True}
+        self.server.manager(MGR_CMD_CREATE, QUEUE, a, id='new_q')
+
+    def test_built_in_resc(self):
+        """
+        Test built-in resource and check the assigned resource value
+        after unset and shouldn't set to any negative value
+        """
+        # Unset the resource after job completion
+        a = {'resources_available.ncpus': 4}
+        self.server.manager(MGR_CMD_SET, QUEUE, a, id='new_q')
+        j_attr = {ATTR_queue: 'new_q', 'Resource_List.select': '1:ncpus=3'}
+        j = Job(TEST_USER, j_attr)
+        j.set_sleep_time(20)
+        jid = self.server.submit(j)
+        self.server.expect(JOB, {'job_state': 'R'}, jid)
+        q_status = self.server.status(
+            QUEUE, 'resources_assigned.ncpus', id='new_q', extend='f')
+        self.assertEqual(
+            int(q_status[0]['resources_assigned.ncpus']), 3, self.err_msg)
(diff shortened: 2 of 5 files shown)


test/tests/functional/pbs_que_resc_usage.py
```diff
@@ -0,0 +1,180 @@
...
+        j.set_sleep_time(20)
+        jid = self.server.submit(j)
+        self.server.expect(JOB, {'job_state': 'R'}, jid)
+        q_status = self.server.status(
+            QUEUE, 'resources_assigned.ncpus', id='new_q', extend='f')
+        self.assertEqual(
+            int(q_status[0]['resources_assigned.ncpus']), 3, self.err_msg)
+        self.server.expect(JOB, 'queue', op=UNSET, id=jid)
+        q_status = self.server.status(
+            QUEUE, 'resources_assigned.ncpus', id='new_q', extend='f')
```
bhroam: what does extend=f mean?  I don't see any reference to it in req_statque().

test/tests/functional/pbs_que_resc_usage.py
```diff
@@ -0,0 +1,180 @@
...
+        self.assertEqual(
+            int(q_status[0]['resources_assigned.ncpus']), 0, self.err_msg)
+        self.server.manager(MGR_CMD_UNSET, QUEUE,
+                            'resources_available.ncpus', id='new_q')
+        q_status = self.server.status(QUEUE, id='new_q', extend='f')
+        self.assertNotIn('resources_available.ncpus',
+                         q_status[0].keys(), self.err_msg)
+        self.assertNotIn('resources_assigned.ncpus',
+                         q_status[0].keys(), self.err_msg)
+
```
bhroam: I thought the code specifically only unset resources_assigned on non-builtins.  ncpus is a builtin.

test/tests/functional/pbs_que_resc_usage.py
```diff
@@ -0,0 +1,180 @@
...
+        q_status = self.server.status(QUEUE, id='new_q', extend='f')
+        self.assertNotIn('resources_available.ncpus',
+                         q_status[0].keys(), self.err_msg)
+        self.assertEqual(
+            int(q_status[0]['resources_assigned.ncpus']), 3, self.err_msg)
+        self.server.delete(jid, wait=True)
+        q_status = self.server.status(
+            QUEUE, 'resources_assigned.ncpus', id='new_q', extend='f')
+        self.assertEqual(
+            int(q_status[0]['resources_assigned.ncpus']), 0, self.err_msg)
```
bhroam: Why are you testing this in a second queue?  There is nothing special to the server about workq.  It's just any other queue.  It is created by habitat as the default queue.  Testing a second queue is not necessary.

src/server/req_manager.c
```diff
@@ -1127,7 +1127,10 @@ mgr_unset_attr(attribute *pattr, attribute_def *pdef, int limit, svrattrl *plist
 							if (i >= 0) {
 								if ((pattr+i)->at_flags & ATR_VFLAG_SET) {
 									resource *nresc;
-									if ((nresc = find_resc_entry((pattr+i), prsdef)) != NULL) {
+									/* if resources_assigned value is non-zero then don't unset it, otherwise it
+									 * would appear as negative in qstat once job is finished.
+									 */
+									if (((nresc = find_resc_entry((pattr+i), prsdef)) != NULL) && (nresc->rs_value.at_val.at_long == 0)) {
```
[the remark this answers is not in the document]
Bhagat-Rajput: We can't create a resource with flag=q and type=string or string_array or boolean.
Still if we have a resource of type "string" then we never add "resources_assigned" attribute during consumption.
Added checks for "float" and "size" type resources.

test/tests/functional/pbs_que_resc_usage.py
Bhagat-Rajput: It's for both built-in and non-built-in resources, the behavior is different only during unset.
Here we have added a built-in resource explicitly to the queue and unset the resource after job completion but if we unset the resource before job completion then code unset/removes only resources_assigned on non-built-ins.

src/server/req_manager.c
```diff
@@ -1127,17 +1127,19 @@ mgr_unset_attr(attribute *pattr, attribute_def *pdef, int limit, svrattrl *plist
...
-										(pattr+i)->at_flags |= ATR_VFLAG_MODCACHE|ATR_VFLAG_MODIFY;
+									if ((nresc = find_resc_entry((pattr + i), prsdef)) != NULL) {
+										/* if resources_assigned value is non-zero then don't unset it, otherwise it
+										 * would appear as negative in qstat once job is finished.
+										 */
+										if (check_resc_assign_val(nresc)) {
+											nresc->rs_defin->rs_free(&nresc->rs_value);
+											delete_link(&nresc->rs_link);
+											free(nresc);
+											nresc = (resource *)GET_NEXT((pattr+i)->at_val.at_list);
```
bhroam: Put spaces around the '+' (here and below)

arungrover:
@Bhagat-Rajput why don't you unset the resources_assigned.<resc_name> attribute when admin unsets resource_available.<resc_name>? This way you will not have to care about removing the resources_assigned attribute after updating it.
I see that you are checking resources_assigned value to be zero or negative before unsetting it, but since PBS does not enforce admins to create resources that can only have zero-or-positive value only, there are chances that resource available on the queue had negative value, to begin with.
I think if you just unset resource_assigned attribute as and when we unset resources_available attribute, it will just work fine. Of course, you will have to make a change in set_resource_assigned function to make sure that you update resource_assigned only when its corresponding resources_available is present.
Also, wouldn't this change also apply to server's resources_assigned attribute?

Bhagat-Rajput:
> 
Thanks @arungrover for the review, please find my replies inline.
> 
> @Bhagat-Rajput why don't you unset the resources_assigned.<resc_name> attribute when admin unsets resource_available.<resc_name>? This way you will not have to care about removing the resources_assigned attribute after updating it.

Resources_assigned.<resc_name> gets set to some negative value just because of this behavior. Once we unset both the attributes, the job releases its acquired resources and server re-creates a resources_assigned attribute for those resources( without checking whether it's corresponding resources_available is present or not) and set the negative values to it.
> I see that you are checking resources_assigned value to be zero or negative before unsetting it, but since PBS does not enforce admins to create resources that can only have zero-or-positive value only, there are chances that resource available on the queue had negative value, to begin with.

True but in that case PBS will never run that job and its corresponding resources_assigned attribute will never get populated.

> I think if you just unset resource_assigned attribute as and when we unset resources_available attribute, it will just work fine. Of course, you will have to make a change in set_resource_assigned function to make sure that you update resource_assigned only when its corresponding resources_available is present.

Initially i was also thinking the same way but in this case if admin sets the same resc again then it's assigned values will be totally messed up.

> Also, wouldn't this change also apply to server's resources_assigned attribute?

then we have to search the same resources_available.<resc_name> everywhere(queues, server and nodes) because it could be anywhere.

arungrover:
@Bhagat-Rajput 

> Resources_assigned.<resc_name> gets set to some negative value just because of this behavior. Once we unset both the attributes, the job releases its acquired resources and server re-creates a resources_assigned attribute for those resources( without checking whether it's corresponding resources_available is present or not) and set the negative values to it.

I am suggesting that you unset resources_assigned.<resource> as soon as admin unsets resources_available.<resource>. There is really no benefit of having resources_assigned set if the corresponding resources_available is unset. The reason it is set to a negative value today is because we add a resource entry in queue's (and server's) resources_assigned list when we are not able to find one. What we can do is that before adding this entry into resources_assigned attribute we should check if the corresponding resource has an entry in resources_available attribute of queue (or server). [This is the line](https://github.com/PBSPro/pbspro/blob/master/src/server/svr_func.c#L502) where we should add this check.

> Initially i was also thinking the same way but in this case if admin sets the same resc again then it's assigned values will be totally messed up.

This is an unlikely event, but as long as you make sure that when a resource is added to resources_available list, you are correctly populating the resources_assigned for that resource by looking at all the jobs in the queue this should not be a problem.

>then we have to search the same resources_available.<resc_name> everywhere(queues, server and nodes) because it could be anywhere.

That is true, this change is needed for all three objects (not just queues).

Bhagat-Rajput:
> 
> 
> @Bhagat-Rajput
> 
> > Resources_assigned.<resc_name> gets set to some negative value just because of this behavior. Once we unset both the attributes, the job releases its acquired resources and server re-creates a resources_assigned attribute for those resources( without checking whether it's corresponding resources_available is present or not) and set the negative values to it.
> 
> I am suggesting that you unset resources_assigned. as soon as admin unsets resources_available.. There is really no benefit of having resources_assigned set if the corresponding resources_available is unset. The reason it is set to a negative value today is because we add a resource entry in queue's (and server's) resources_assigned list when we are not able to find one. What we can do is that before adding this entry into resources_assigned attribute we should check if the corresponding resource has an entry in resources_available attribute of queue (or server). [This is the line](https://github.com/PBSPro/pbspro/blob/master/src/server/svr_func.c#L502) where we should add this check.

I got your point here and it's a good suggestion though but we didn't implemented earlier because of that unlikely event where admin re-creates the same resource. 
> 
> > Initially i was also thinking the same way but in this case if admin sets the same resc again then it's assigned values will be totally messed up.
> 
> This is an unlikely event, but as long as you make sure that when a resource is added to resources_available list, you are correctly populating the resources_assigned for that resource by looking at all the jobs in the queue this should not be a problem.

As per my knowledge and understanding, i don't think it's an optimal/efficient solution performance wise because admin can create a resource on all the queues/nodes at a time then in that case we need to process all the jobs just to search a single attribute on all the queues/nodes.
Could you please elaborate more on how is your suggestion better than the current implementation ?
> 
> > then we have to search the same resources_available.<resc_name> everywhere(queues, server and nodes) because it could be anywhere.
> 
> That is true, this change is needed for all three objects (not just queues).

arungrover:
I agree that it might be inefficient to traverse through all the jobs and find out the resource was in use BUT it should not be that big of a deal as this does not happen that often.
The main reason I dislike your change is because you check resources assigned to be 0or negative to determine that it can be unset and in PBS long resources are not mandated anywhere to have a positive integer value. 
Now you already have a sign off and I can understand it might be too late for you to accept any more changes. You can discuss this with your reviewer and team and take a decision.

bhroam:
I think I agree with @arungrover here.  While it is expensive to traverse all the jobs, we're talking about a very esoteric edge case when we'll have to do it.  Arun's suggestion sounds like the right way to go.

subhasisb:
@bhroam @arungrover i am not sure i agree with your proposed approach of recounting everything when a resource_available is added. Since we won't know whether the resource was used by any jobs before, with this approach, we will have to check for the same for any resource limit we add. Granted it is probably not very frequent, but we should not add inefficient code when it appears there is a robust, fast and practical solution available. (case in point is that even server restarts are infrequent, and yet anybody doing a restart wishes restarts be faster).

@arungrover do we have a practical use case of resources_available being a negative number? AFAIK, our resources_assigned count always goes from 0 to positive, and thus a negative resoruces_available seems to be a moot point? (Or maybe i am missing a whole story here?). If not, we could treat that behavior as a bug and fix it.

The other practical difficulty with counting everything (like jobs etc) is when we go to multi-server. A single server simply may not have enough information about all the jobs - this will make the implementation complex, error-prone without gaining much in return.

The other alternative is to simply disallow unsetting the value, just like we disallow deleting the resource itself when jobs are using it - however, not sure that can be acceptable.

Thoughts?

arungrover:
> @arungrover do we have a practical use case of resources_available being a negative number? AFAIK, our resources_assigned count always goes from 0 to positive, and thus a negative value seems to be a moot point? (Or maybe I am missing a whole story here?). If not, we could treat that behavior as a bug and fix it.
> 
Customers have been setting resources as negative to use them as coefficients in formulas in many cases. I spoke to Alexis, he mentioned in one of the case he saw customers requesting negative resources in requests where the job is actually run to release the resource (like disk space).
In the recent past, we have fixed bugs in scheduler where scheduler was considering negative resources and some internal constants (https://github.com/PBSPro/pbspro/pull/1144)
The thing is that we do not have any concept of unsigned integer resources, our guides, say that the float format can be negative. Checking to see that a resource value turned negative means it is safe to unset it is semantically incorrect.

> The other practical difficulty with counting everything (like jobs etc) is when we go to multi-server. A single server simply may not have enough information about all the jobs - this will make the implementation complex, error-prone without gaining much in return.
> 
> The other alternative is to simply disallow unsetting the value, just like we disallow deleting the resource itself when jobs are using it - however, not sure that can be acceptable.
> 
IMO, when resource_available is unset, resources_assigned serves no purpose. So both of them should be unset together.
The thing that Bhagat is worried about is that if a resources_available is being unset when there is running in the entity and then set back again before this running job ends would make the resources_assigned negative is such a remote possibility. 
There are two things I can think of - 
- One, we must be setting resource_assigned to 0 somewhere when resource_available is unset. If we stop doing that, resources_assigned will not turn negative. 
- Another thing we can do is that we can make a change in set_resc_assigned and check that if resources_available.<res> value is a positive integer then never let resources_assigned.<res> to go negative when a job ends. This will take care of this corner case.

I think it is also a good solution to disallow the admin to unset resource_available on an entity as long as there are jobs using that resource in that entity.

subhasisb:
> Customers have been setting resources as negative to use them as coefficients in formulas in many cases. I spoke to Alexis, he mentioned in one of the case he saw customers requesting negative resources in requests where the job is actually run to release the resource (like disk space).
> In the recent past, we have fixed bugs in scheduler where scheduler was considering negative resources and some internal constants (#1144)
> The thing is that we do not have any concept of unsigned integer resources, our guides, say that the float format can be negative. Checking to see that a resource value turned negative means it is safe to unset it is semantically incorrect.

I think the issue is how do we know that the resources_assigned value is finally done, since we do not have an initial value. Can we assume zero is ALWAYS the initial value. If so, we can unset only if it is zero, not negative or positive. That should solve the issue, right?

> IMO, when resource_available is unset, resources_assigned serves no purpose. So both of them should be unset together

I differ a bit there. I think it does serve a purpose, exactly in this case - where you remember how much was assigned to the jobs, such that if the resource limit was to be set again, we do not miscalculate it (and do not need a very expensive recount)

> The thing that Bhagat is worried about is that if a resources_available is being unset when there is running in the entity and then set back again before this running job ends would make the resources_assigned negative is such a remote possibility.

yes, of course - that is the core of the bug

> One, we must be setting resource_assigned to 0 somewhere when resource_available is unset. If we stop doing that, resources_assigned will not turn negative.

It appears to me that  that is exactly this bug. That we nuke resources_assigned when somebody unsets resources_available, even when jobs are running. The proposed solution is to not unset it right away, rather unset it when "all jobs that are using it are done" - in other words, i think, when the value is zero.

> Another thing we can do is that we can make a change in set_resc_assigned and check that if resources_available. value is a positive integer then never let resources_assigned. to go negative when a job ends. This will take care of this corner case.

Not sure how this is fix this bug. We unset resources_assigned when we unset resources_available today and that causes the bug.

Bhagat-Rajput:
I've raised a new fresh PR https://github.com/PBSPro/pbspro/pull/1599 for this issue thus closing the current one. Please review it! Thanks

PR #1500, opened 2020-02-04
Management hook event

ericpershey:
#### Describe Bug or Feature
In order to track server management operations, such as adding a new node, importing a hook script, etc., a new management hook event has been added to the PBS server. Registered management hook scripts are executed once the management operation has completed. Each script is provided information about the requested operation, any associated attributes, and the status of the completed operation.

#### Link to Design Doc
**[Server Management Hook Event](https://pbspro.atlassian.net/wiki/spaces/PD/pages/1404239875/Server+Management+Hook+Event)**


test/tests/pbs_smoketest.py
```diff
@@ -750,13 +751,19 @@ def test_round_robin(self):
         a = {'scheduling': 'False'}
         self.server.manager(MGR_CMD_SET, SERVER, a)
         self.server.expect(JOB, {'job_state=R': 9})
-        cycle = self.scheduler.cycles(start=start_time, end=int(time.time()))
+        end_time = int(time.time()) + 1
```
ericpershey: Due to the start_time and end_time being truncated with int(), the cycles() call may not read all the required logs to find all the data it needs and fail.  This might be related to this: https://pbspro.atlassian.net/browse/PP-359

src/modules/python/pbs/v1/_svr_types.py
```diff
@@ -1365,3 +1365,136 @@ def __next__(self):
...
+            raise BadAttributeValueError(
+                "'%s' attribute in the management object is readonly" % (name,))
+        super().__setattr__(name, value)
+    #: m(__setattr__)
+
+_management.cmd = PbsAttributeDescriptor(_management, 'cmd', None, (int,))
+_management.objtype = PbsAttributeDescriptor(_management, 'objtype', None, (int,))
+_management.objname = PbsAttributeDescriptor(_management, 'objname', "", (str,))
+_management._connect_server = PbsAttributeDescriptor(
+    _management, '_connect_server', "", (str,))
```
ericpershey: Do we need these lines?  What purpose do they server?

src/modules/python/pbs/v1/_svr_types.py
```diff
@@ -1365,3 +1365,136 @@ def __next__(self):
...
+                lst.append(mask)
+        return lst
+    #: m(extract_flags_int)
+
+    def tup(self):
+        return self.name, self.resource, self.value, self.op, self.flags, self.sisters
+    #: m(tup)
+
+_server_attribute._connect_server = PbsAttributeDescriptor(
+    _server_attribute, '_connect_server', "", (str,))
```
ericpershey: Do we need these lines?  What purpose do they server?

ericpershey:
Valgrind output to come.  @toonen is working on it.

src/lib/Libpython/pbs_python_svr_internal.c
```diff
@@ -5749,6 +5759,59 @@ _pbs_python_event_set(unsigned int hook_event, char *req_user, char *req_host,
...
+			goto event_set_exit;
+		}
+
+		py_attr = svrattrl_list_to_pyobject(&rqj->rq_manager.rq_attr);
+		if (!py_attr){
+			log_err(PBSE_INTERNAL, __func__, "could not build the list of server attributes");
+			goto event_set_exit;
+		}
+
+		py_margs = Py_BuildValue("(iisliiisO)",
```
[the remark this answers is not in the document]
ericpershey: from https://docs.python.org/3/c-api/arg.html

O (object) [PyObject *]

  Store a Python object (without any conversion) in a C object pointer. The C program thus
  receives the actual object that was passed. The object’s reference count is not increased. 
  The pointer stored is not NULL.

src/lib/Libpython/pbs_python_svr_internal.c
```diff
@@ -12487,3 +12551,139 @@ pbsv1mod_meth_release_nodes(PyObject *self, PyObject *args, PyObject *kwds)
...
+ * @brief
+ *	Returns a String containing a space separated list of attribute name,
+ *  resource name and attribute/resource value.  For debugging only.
+ * @param[in]	head_str- some string to print out the beginning.
+ * @param[in]	phead	- pointer to the head of the list containing data.
+ *
+ * @return char*
+ */
+
+const char * asprint_svrattrl_list_all(char *head_str, pbs_list_head *phead)
```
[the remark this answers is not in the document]
ericpershey: We were using this to debug the code.  Should we just remove it out of the source code?

src/lib/Libpython/pbs_python_svr_internal.c
```diff
@@ -12487,3 +12551,139 @@ pbsv1mod_meth_release_nodes(PyObject *self, PyObject *args, PyObject *kwds)
...
+			PyObject* py_slist = PyObject_GetAttrString(py_server_attribute, "sisters");
+			if (py_slist) {
+				for(slist = plist->al_sister; slist != NULL; slist = slist->al_sister) {
+					PyObject *py_server_attribute_sister = svrattrl_to_server_attribute(slist);
+					if (py_server_attribute_sister) {
+						PyList_Append(py_slist, py_server_attribute_sister);
+						Py_CLEAR(py_server_attribute_sister);
+					}
+				}
+				Py_CLEAR(py_slist);
```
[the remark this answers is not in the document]
ericpershey: No good reason, removing.

src/lib/Libpython/pbs_python_svr_internal.c
```diff
@@ -12487,3 +12551,139 @@ pbsv1mod_meth_release_nodes(PyObject *self, PyObject *args, PyObject *kwds)
...
+			} else {
+				log_err(PBSE_INTERNAL, __func__,
+					"failed to acquire sisters in server_attribute object");
```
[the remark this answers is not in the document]
ericpershey: Should we remove the else entirely?

src/lib/Libpython/pbs_python_svr_internal.c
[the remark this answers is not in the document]
arungrover: you are right, it should be 'O'. I don't know why but it looked to me like a 0 (zero) in the format. Looks like that is not the case :)

src/lib/Libpython/pbs_python_svr_internal.c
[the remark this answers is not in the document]
arungrover: If you don't use the function or have no use for it in foreseeable future then it would be better to remove it.

src/lib/Libpython/pbs_python_svr_internal.c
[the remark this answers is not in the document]
arungrover: Yes, I think it is better to remove it. Thanks!

src/server/hook_func.c
```diff
@@ -3881,6 +3885,14 @@ process_hooks(struct batch_request *preq, char *hook_msg, size_t msg_len,
 				"Did not find a job tied to runjob request!");
 			return (-1);
 		}
+	} else if (preq->rq_type == PBS_BATCH_Manager) {
+		hook_event = HOOK_EVENT_MANAGEMENT;
+		preq->rq_ind.rq_management.rq_reply = &preq->rq_reply;
+		preq->rq_ind.rq_management.rq_time = preq->rq_time;
+		// Copying the pointer to rq_management below is safe since
```
arungrover: Please use c style comments.

test/tests/functional/pbs_hook_management.py
```diff
@@ -0,0 +1,241 @@
...
+
+@tags('hooks', 'smoke')
+class TestHookManagement(TestFunctional):
+
+    def test_hook_00(self):
+        """
+        By creating an import hook, it executes a management hook.
+        """
+        self.logger.info("**************** HOOK START ****************")
+        # hook_name = "test_management"
```
arungrover: Please remove commented code

test/tests/functional/pbs_hook_management.py
```diff
@@ -0,0 +1,241 @@
...
+        hook_name = "management"
+        hook_msg = 'running management hook_00'
```
arungrover: It would be better if you create a function to create the hook body and pass in the hook_msg to that function.

test/tests/functional/pbs_hook_management.py
```diff
@@ -0,0 +1,241 @@
...
+        e = pbs.event()
+        m = e.management
+        pbs.logmsg(pbs.LOG_DEBUG, '%s')
+        """ % hook_msg
+        hook_body = textwrap.dedent(hook_body)
+        # attrs = {'event': 'management', 'enabled': 'True'}
+        # rv = self.server.create_import_hook(
+        #     hook_name, attrs=attrs, body=hook_body, overwrite=True)
+
+        qmgr_path = os.path.join(self.server.pbs_conf["PBS_EXEC"], "bin",
```
arungrover: If you use self.server.create_import_hook() method, you don't have to do any of these things yourself. Please have a look at other hook related PTL tests (like pbs_hooksmoketest.py)

The same comment applies to all other tests as well.

test/tests/functional/pbs_hook_management.py
```diff
@@ -0,0 +1,241 @@
...
+            tempfd.write(hook_body)
+            self.logger.info("tempfd_name:%s" % tempfd_name)
+
+        self.logger.info("hook_name:%s" % hook_name)
+        qmgr_setup = [qmgr_path, "-c",
+                      "create hook %s event=management" % hook_name]
+        qmgr_cmd = [qmgr_path, "-c",
+                    "import hook %s application/x-python default %s" %
+                    (hook_name, tempfd_name)]
+        qmgr_cleanup = [qmgr_path, "-c",
```
arungrover: You don't have to worry about cleanup, setup() will take care of it, next time another test starts.

test/tests/pbs_smoketest.py
```diff
@@ -1371,17 +1378,27 @@ def test_resource_delete(self):
...
-                continue
-            v = self.obj_map[k]
-            for t in self.resc_types:
-                for f in self.resc_flags:
-                    for c in self.resc_flags_ctl:
-                        self.delete_resource_helper(
-                            self.resc_name, t, f, c, k, v)
-                        self.logger.info("")
+        count = 0
+        starttime = int(time.time())
```
arungrover: Can you please explain why this change is needed? was this test failing for you?

ericpershey: Yes this test was failing.  When we would run this in debian9 with python 3.5 and if there was type hinting code inside the python, it would take much longer to run.  There is some kind of bug in python 3.5 that causes this.  I think the version was specifically 3.5.3.  This test seems to be the most sensitive to time and I added a timer so that when it fails in the ptl, we can compare it to the other tests and also still get the error.  Brian might be able to explain it better as he dug into it.

test/tests/functional/pbs_hook_management.py
ericpershey: Since the qmgr commands are firing hooks now, I wanted to test the delete also.

test/tests/functional/pbs_hook_management.py
ericpershey: I will take a look at this.

test/tests/functional/pbs_hook_management.py
ericpershey: Yes, this makes much more sense and less duplicated code.

src/lib/Libpython/pbs_python_svr_internal.c
```diff
@@ -12487,3 +12551,139 @@ pbsv1mod_meth_release_nodes(PyObject *self, PyObject *args, PyObject *kwds)
...
+ *  resource name and attribute/resource value.  For debugging only.
+ * @param[in]	head_str- some string to print out the beginning.
+ * @param[in]	phead	- pointer to the head of the list containing data.
+ *
+ * @return char*
+ */
+
+const char * asprint_svrattrl_list_all(char *head_str, pbs_list_head *phead)
+{
+	char *buf = strdup(head_str);
```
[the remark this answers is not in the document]
ericpershey: Take a look at the new code.

src/lib/Libpython/pbs_python_svr_internal.c
```diff
@@ -12487,3 +12551,139 @@ pbsv1mod_meth_release_nodes(PyObject *self, PyObject *args, PyObject *kwds)
...
+	size_t size = strlen(buf);
+	svrattrl *plist = NULL;
+
+	if ((head_str == NULL) || (phead == NULL)) {
+		log_err(errno, __func__, "NULL input parameters!");
+		return buf;
```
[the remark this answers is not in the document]
ericpershey: Take a look at the new code.

src/lib/Libpython/pbs_python_svr_internal.c
```diff
@@ -12487,3 +12551,139 @@ pbsv1mod_meth_release_nodes(PyObject *self, PyObject *args, PyObject *kwds)
...
+	}
+
+	for (plist = (svrattrl *)GET_NEXT(*phead); plist != NULL;
+		plist = (svrattrl *)GET_NEXT(plist->al_link)) {
+		char * tmp_buf;
+		size_t tmp_size;
+		pbs_asprintf(&tmp_buf, " %s.%s[%s]=%s", head_str,
+			plist->al_name, plist->al_resc, plist->al_value);
+		tmp_size = strlen(tmp_buf);
```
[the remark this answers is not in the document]
ericpershey: Updated the code.

src/lib/Libpython/pbs_python_svr_internal.c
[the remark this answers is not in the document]
ericpershey: Removed

src/lib/Libpython/pbs_python_svr_internal.c
```diff
@@ -12487,3 +12551,139 @@ pbsv1mod_meth_release_nodes(PyObject *self, PyObject *args, PyObject *kwds)
...
+* @retval	NULL		- if an error occurred.
+ */
+PyObject *svrattrl_list_to_pyobject(pbs_list_head *phead)
+{
+	svrattrl *plist = NULL;
+	PyObject* py_list = PyList_New(0);
+
+	if (phead == NULL) {
+		log_err(errno, __func__, "NULL input parameters!");
+		return NULL;
```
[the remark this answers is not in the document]
ericpershey: Code updated

src/lib/Libpython/pbs_python_svr_internal.c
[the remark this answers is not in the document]
ericpershey: Removed

src/lib/Libpython/pbs_python_svr_internal.c
```diff
@@ -12487,3 +12551,139 @@ pbsv1mod_meth_release_nodes(PyObject *self, PyObject *args, PyObject *kwds)
...
+					}
+				}
+				Py_CLEAR(py_slist);
+			} else {
+				log_err(PBSE_INTERNAL, __func__,
+					"failed to acquire sisters in server_attribute object");
+			}
+			PyList_Append(py_list, py_server_attribute);
+			Py_CLEAR(py_server_attribute);
+		}
```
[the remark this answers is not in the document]
ericpershey: Take a look at the error code I added.

src/lib/Libpython/pbs_python_svr_internal.c
```diff
@@ -12487,3 +12551,139 @@ pbsv1mod_meth_release_nodes(PyObject *self, PyObject *args, PyObject *kwds)
...
+	}
+	return py_list;
+}
+
+/**
+ *
+ * @brief
+ *	Returns a Python List of _server_attribute objects.
```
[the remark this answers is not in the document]
ericpershey: Updated

src/server/hook_func.c
ericpershey: Updated

test/tests/functional/pbs_hook_management.py
```diff
@@ -48,6 +48,17 @@
...
 
+def get_hook_body(hook_msg):
+    hook_body = """
+    import pbs
+    e = pbs.event()
+    m = e.management
+    pbs.logmsg(pbs.LOG_DEBUG, '%s')
+    """ % hook_msg
+    hook_body = textwrap.dedent(hook_body)
+    return hook_msg
```
arungrover: This should be hook_body instead

ericpershey: Interesting that that got through.  I am trying to find a way to make sure my changes get correctly propagated to my docker container.  The test is now failing.  For now I will just rebuild the entire thing, which doesn't take long at all.

ericpershey: This should be fixed now.  There were other problems in the test runner that I just fixed while correcting this.

test/tests/functional/pbs_hook_management.py
ericpershey: I added delete_hook to pbs_testlib.py and then switched over to using create_hook, import_hook and delete_hook.  The code looks much cleaner now.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -9613,6 +9613,22 @@ def create_hook(self, name, attrs):
...
+        :param name: The name of the hook to create
+        :type name: str
+        :returns: False if hook does not exist
+        :raises: PbsManagerError, otherwise return True.
+        """
+        hooks = self.status(HOOK, level=logging.DEBUG)
+        for hook in hooks:
+            if hook['id'] == name:
+                self.logger.info("Removing hook:%s" % name)
+                self.manager(MGR_CMD_DELETE, HOOK, id=[name])
```
arungrover: Thanks for adding this helper function!
Just a minor comment - the name does not need to be a list, you can just pass in a string.

bayucan:
My code review is still in progress...

test/tests/pbs_smoketest.py
toonen: From commit 9993b91f2f99598ef4b2b81f9f4fe07bbd12eea8 (with a bit of clean up):

>The Python typing module seems to be unstable in Python 3.5.3, which is what Debian 9 uses.  With the type specifications in place, Py_Finalize() would take increasing longer on each Python interpreter restart, eventually taking more than two seconds to complete and causing the SmokeTest.test_resource_delete to timeout.  Removing the type specifications results in the test successfully completion in the expected amount of time.

BTW, the issue disappeared when we tested with Python 3.5.9.  The change log for Python 3.5 noted updates to the typing module between 3.5.3 and 3.5.9, so I'm guessing they fixed a bug in 3.5.3 that Debian didn't backport.

arungrover: In that case, can we remove this change? I am not sure we should add this change for something that one Linux distribution failed to backport.

ericpershey: I think we should leave it that way if there is something else that slows this down, we can get timings on what happened.  All it's doing is showing how long it took to iterate through all the tests.  This also may be associated with random failures in here: https://pbspro.atlassian.net/browse/PP-359

src/lib/Libpython/pbs_python_svr_internal.c
```diff
@@ -12487,3 +12551,107 @@ pbsv1mod_meth_release_nodes(PyObject *self, PyObject *args, PyObject *kwds)
...
+/**
+ *
+ * @brief
+ *	Returns a Python List of _server_attribute objects or  or NULL on error.
+ * @param[in]	phead	- pointer to the head of the list containing data.
+ *
+ * @return 	PyObject *
+ * @retval	<object>	- the Python list object holding
+ * @retval				the _server_attribute objects.
+ * @retval	NULL		- if an error occurred.
```
bayucan: Please put an @note here that the returned <object> contains a new reference, which must by Py_CLEAR-ed later. This is to guard its future use.

src/lib/Libpython/pbs_python_svr_internal.c
```diff
@@ -12487,3 +12551,107 @@ pbsv1mod_meth_release_nodes(PyObject *self, PyObject *args, PyObject *kwds)
...
+/**
+ *
+ * @brief
+ *	Returns a Python _server_attribute object or NULL on error.
+ * @param[in]	attribute	- pointer to the head of the list containing data.
+ *
+ * @return 	PyObject *
+ * @retval	<object>	- the Python _server_attribute object.
+ * @retval	NULL		- if an error occurred.
+ */
```
bayucan: Same here too, please add a @note that says the returned \<object> contains a new reference that must be Py_CLEAR-ed later.

test/tests/pbs_smoketest.py
arungrover: ok, I agree. Let us leave it as is.
Thanks!

src/modules/python/pbs/v1/_svr_types.py
bayucan: It does look like it may not have any use right now, but perhaps in the future. It sort of associates the current class object with the particular PBS server. There could be multiple servers in the future being handled in hooks. So I'd say just leave it for now.

src/lib/Libpython/pbs_python_svr_internal.c
ericpershey: Note was added.

src/lib/Libpython/pbs_python_svr_internal.c
ericpershey: Note was added.

src/modules/python/pbs/v1/_svr_types.py
ericpershey: I will leave it.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -9613,6 +9613,22 @@ def create_hook(self, name, attrs):
         self.manager(MGR_CMD_SET, HOOK, attrs, id=name)
         return True
 
+    def delete_hook(self, name):
+        """
+        Helper function to delete a hook by name.
+
+        :param name: The name of the hook to create
```
bayucan: I think you mean "name of the hook to delete".

src/modules/python/pbs/v1/_svr_types.py
```diff
@@ -1365,3 +1365,136 @@ def __next__(self):
...
+        self.name = name
+        self.resource = resource
+        self.value = value
+        self.op = op
+        self.flags = flags
+        self.sisters = []
+    #: m(__init__)
+
+    def __str__(self):
+        return "%s:%s:%s:%s:%s" % self.tup()
```
bayucan: self.tup() returns 6 tuples but you only have 5 %s, it should be 6 %s. I realized this hen I tried the following management hook: 
for a in pbs.event().management..attribs:
    pbs.logmsg(pbs.LOG_DEBUG, "%s" % (a))

I got an exception:
3/13/2020 02:17:56.296640;0001;Server@corretja;Svr;Server@corretja;PBS server internal error (15011) in Error evaluating Python script, <class 'TypeError'>
03/13/2020 02:17:56.296644;0001;Server@corretja;Svr;Server@corretja;PBS server internal error (15011) in Error evaluating Python script, not enough arguments for format string

Better yet, for readability, just do:
def __str__(self):
        return "name=%s:resource=%s:value=%s:op=%s:flags=%s:sisters=%s" % self.tup()

So the output is more readable:
3/13/2020 02:24:31.334731;0006;Server@corretja;Hook;Server@corretja;a=name=enabled:resource=None:value=true:op=0:flags=0:sisters=[]

test/tests/functional/pbs_hook_management.py
```diff
@@ -0,0 +1,151 @@
...
+
+def get_hook_body(hook_msg):
+    hook_body = """
+    import pbs
+    e = pbs.event()
+    m = e.management
+    pbs.logmsg(pbs.LOG_DEBUG, '%s')
+    """ % hook_msg
+    hook_body = textwrap.dedent(hook_body)
+    return hook_body
```
bayucan: I would like your ptl test to be expanded some more. There are things in your design doc need to be exercised. 

- The pbs.event().management.* attribute values need to be shown and verified as stated in the design:
    

> pbs.event().management: _management
>         cmd: integer value of the server management operation, constants defined in pbs.MGR_CMD_*, translation to text is found in pbs.REVERSE_MGR_CMDS
>         objname: name of the object, such as the hook name.
>         objtype: integer value of the server management operation type, constants defined in pbs.MGR_OBJ_*, translation to text is found in pbs.REVERSE_MGR_OBJS
>         request_time: time of the request in seconds since epoch
>         reply_text: text returned to the user via the server management operation
>         reply_code: code returned to the user via the server management operation
>         reply_choice: request choice, constants defined in pbs.BRP_CHOICE_*, translation to text is found in pbs.REVERSE_BRP_CHOICES
>         reply_auxcode: Auxiliary error code
>         attribs: List of server_attribute class objects
>     pbs.event().management.rq_attrs: List[_server_attribute]
>         name: name of the attribute
>         resource: resource of the attribute
>         value: value of the attribute
>         op: integer value of the attribute operation, constants defined in pbs.BATCH_OP_*, translation to text is found in pbs.REVERSE_BATCH_OPS
>         flags: integer value of the attribute flags, constants defined in pbs.ATR_VFLAG_*, translation to text is found in pbs.REVERSE_ATR_VFLAGS
>         sisters: list of co-resources for the attribute

As an example, I created this management hook:
`import pbs`
`import pbs`
`e = pbs.event()`
`m = e.management`
`pbs.logmsg(pbs.LOG_DEBUG, "executed management hook order 2")`
`pbs.logmsg(pbs.LOG_DEBUG, "cmd=%s" % (pbs.REVERSE_MGR_CMDS[m.cmd]))`
`pbs.logmsg(pbs.LOG_DEBUG, "objtype=%s" % (pbs.REVERSE_MGR_OBJS[m.objtype]))`
`pbs.logmsg(pbs.LOG_DEBUG, "reply_choice=%s" % (pbs.REVERSE_BRP_CHOICES[m.reply_choice]))`
`for a in m.attribs:`
 `\   pbs.logmsg(pbs.LOG_DEBUG, "%s" % (a))`

When I ran qmgr -c 'set hook qmgr enabled=true', the management hook executed and produced this output:
03/13/2020 02:24:31.334703;0006;Server@corretja;Hook;Server@corretja;executed management hook order 2
03/13/2020 02:24:31.334709;0006;Server@corretja;Hook;Server@corretja;cmd=MGR_CMD_SET
03/13/2020 02:24:31.334713;0006;Server@corretja;Hook;Server@corretja;objtype=MGR_OBJ_HOOK
03/13/2020 02:24:31.334717;0006;Server@corretja;Hook;Server@corretja;reply_choice=BRP_CHOICE_NULL
03/13/2020 02:24:31.334731;0006;Server@corretja;Hook;Server@corretja;a=name=enabled:resource=None:value=true:op=0:flags=0:sisters=[]`

- You should also have test cases for the following scenarios as mentioned in the design:

> Keeping with the hook design, if one management hook is rejected, the other management hooks with a higher order value will not run.
> 
> If the management hook script encounters an unexpected error causing an unhandled exception, or times out due to the hook's alarm setting, the hook will act similar to a pbs.event().reject().

- Finally, when I added in my test management hook the call "pbs.event().reject("bad hook"), I did not see the message "bad hook" in the server_logs. Whatever message is passed to the reject() call should show up in the server_logs.

bayucan:
@ericpershey : I've completed going through your code. Nice work! There are few requested changes, plus, I need additional test cases in your ptl testsuite, which would touch on some things in your design doc.

src/modules/python/pbs/v1/_svr_types.py
toonen: The response provided for the `_server_attribute` class answers the reason to keep the `PbsAttributeDescriptor` for `_connect_server`, and that seems reasonable to me.  I do still wonder about the other attributes in the `_management` class.  We only added those calls because the existing classes we used as examples has them and we did not understand their purpose.  However, we did not add them for all `_management` class attributes, so we should either expand the calls to include all attributes or remove all but the `_connect_server` attribute.  Again, not understanding the purpose of `PbsAttributeDescriptor`, I'm not sure which is appropriate.

test/tests/functional/pbs_hook_management.py
ericpershey: I am adding the tests now.  I have encountered two errors in server_process_hooks that need to be looked at before I can continue.

src/modules/python/pbs/v1/_svr_types.py
bayucan: The PbsAttributeDescriptor is a way to individually control an attribute per instance, especially if some attributes are to behave a certain way, like making them read-only.

bayucan:
@ericpershey : Thanks for taking my suggestions and making the change. Any more updates you need to push?

ericpershey:
Yes, please don't merge yet.  I am about to add some test to test all the variables.  We have a good set covering quite a bit, but wanted to add a few more based on things you pointed out.

bayucan:
@ericpershey : sounds good.

mike0042:
@arungrover please re-review with the new changes.

arungrover:
Thanks for the reminder @mike0042 .
Change looks good to me. I'll merge it.

Thanks for working on this PR @ericpershey, @toonen  and @bayucan!

PR #1501, opened 2020-02-05
Reducing sched attributes update to Server ...

suresh-thelkar:
#### Describe Bug or Feature

As of today Scheduler updates attributes like sched_host, sched_port etc. to Server during every scheduling cycle even though the frequency of them getting changed is very rare. Instead if we can send these attributes only once to Server we can improve the Scheduler's performance.

#### Describe Your Change

- Make sure that we do update sched_host, sched_port etc. to Server as soon as Scheduler comes up as part of first scheduling cycle. We do not repeatedly update these attributes from the next scheduling cycle on words.
- Server can also change sched object attributes like sched_log, sched_priv etc. especially in case of Multiple Schedulers and sent to the Scheduler upon which it has to override the corresponding attributes at the Scheduler. As this is the existing behavior we should make sure that the current refactoring should not break this.
- Server sends SCH_ATTRS_CONFIGURE whenever any of the Scheduler's attributes change during which time update_svr_schedobj() is called. The problem with this function is 
      1) It updates the attributes like sched_host, sched_port etc. to Server and  
      2) Validates the current scheduler attributes like sched_log, sched_priv etc. with these ones and overrides them. 
      Ideally we should separate these two functionalities and in this case we should only do step 2).
      Code changes are made accordingly to re-factor the code.
- In the absence of the above changes PTL is currently hiding the following bug.
  **Issue:** When scheduler_iteration is unset it actually gets reset to "0". The effect of this is Scheduler kicks of infinite scheduling cycles. 
  Even though this bug is present in our code PTL is unable to find it out because of the following  
  reasons.
PTL resets to default by unsetting all the sched attributes. This will briefly unset scheduler_iteration.   When unsetting all the attributes, it unsets sched_priv/sched_log.  If we unset sched_priv/log, this sends the scheduler a SCH_ATTRS_CONFIGURE command.  On that command the scheduler will set various sched object attributes like scheduler version and host.  Since a sched attribute was set, we check if any attributes need to be set back to the default, and reset scheduler_iteration back to 600.
- Now that we have done the above mentioned re-factoring PTL shows this as bug. Solution for this bug is to remove the unset_flag and make sure if scheduler_iteration is unset, we do have to set it back to 600. Code changes are made accordingly
- Through code inspection found a very corner case memory leak and fixed it.

Have written possible PTL test cases to verify the code changes. Tested by using gdb to make sure that scheduler attributes updates to server are happening only once. As we cannot directly write a PTL test case for it, I used to gdb for the verification.

Also done the following testing and attached the logs.
1. All test cases of the test suite TestMultipleSchedulers 
[TestMultipleSchedulers.log](https://github.com/PBSPro/pbspro/files/4157722/TestMultipleSchedulers.log)

2. All test cases of the test suite TestSchedulerInterface 
[TestSchedulerInterface.log](https://github.com/PBSPro/pbspro/files/4157724/TestSchedulerInterface.log)

Following are the logs after implementing review comments.
[TestSchedulerInterface.log](https://github.com/PBSPro/pbspro/files/4180250/TestSchedulerInterface.log)

### DIFF
--- a/src/include/server.h
+++ b/src/include/server.h
@@ -282,7 +282,7 @@ extern int			sched_save_db(pbs_sched *, int mode);
 extern enum failover_state	are_we_primary(void);
 extern int			have_socket_licensed_nodes(void);
 extern void			unlicense_socket_licensed_nodes(void);
-extern void			set_sched_default(pbs_sched *,  int from_scheduler);
+extern void			set_sched_default(pbs_sched *, int from_scheduler);
 extern void			set_attr_svr(attribute *pattr, attribute_def *pdef, char *value);
 extern int			license_sanity_check(void);
 extern void			memory_debug_log(struct work_task *ptask);
--- a/src/include/server.h
+++ b/src/include/server.h
@@ -282,7 +282,7 @@ extern int			sched_save_db(pbs_sched *, int mode);
 extern enum failover_state	are_we_primary(void);
 extern int			have_socket_licensed_nodes(void);
 extern void			unlicense_socket_licensed_nodes(void);
-extern void			set_sched_default(pbs_sched *, int unset_flag, int from_scheduler);
+extern void			set_sched_default(pbs_sched *,  int from_scheduler);
 extern void			set_attr_svr(attribute *pattr, attribute_def *pdef, char *value);
 extern int			license_sanity_check(void);
 extern void			memory_debug_log(struct work_task *ptask);
--- a/src/scheduler/fifo.c
+++ b/src/scheduler/fifo.c
@@ -569,8 +569,8 @@ schedule(int cmd, int sd, char *runjobid)
 			 * This is required since there is a probability that scheduler's configuration has been changed at
 			 * server through qmgr.
 			 */
-			if (!update_svr_schedobj(connector, 0, 0)) {
-				log_err(-1, __func__, "update_svr_schedobj failed");
+			if (!validate_sched_attrs(connector)) {
+				log_err(-1, __func__, "validate_sched_attrs failed");
 				return 0;
 			}
 			break;
@@ -2689,8 +2689,6 @@ update_svr_schedobj(int connector, int cmd, int alarm_time)
 	static int svr_knows_me = 0;
 	int err;
 	struct attropl*attribs, *patt;
-	struct batch_status *ss = NULL;
-	struct batch_status *all_ss = NULL; /* all scheduler objects */
 	char sched_host[PBS_MAXHOSTNAME + 1];
 
 	/* This command is only sent on restart of the server */
@@ -2700,20 +2698,8 @@ update_svr_schedobj(int connector, int cmd, int alarm_time)
 	if ((cmd != SCH_SCHEDULE_NULL && cmd != SCH_ATTRS_CONFIGURE && svr_knows_me) || cmd == SCH_ERROR || connector < 0)
 		return 1;
 
-	/* Stat the scheduler to get details of sched */
-	all_ss = pbs_statsched(connector, NULL, NULL);
-	ss = bs_find(all_ss, sc_name);
-
-	if (ss == NULL) {
-		sprintf(log_buffer, "Unable to retrieve the scheduler attributes from server");
-		log_err(-1, __func__, log_buffer);
-		pbs_statfree(all_ss);
+	if (!validate_sched_attrs(connector))
 		return 0;
-	}
-	if (!sched_settings_frm_svr(ss))
-		return 0;
-
-	pbs_statfree(all_ss);
 
 	/* update the sched with new values */
 	attribs = calloc(4, sizeof(struct attropl));
@@ -2759,3 +2745,45 @@ update_svr_schedobj(int connector, int cmd, int alarm_time)
 	free(attribs);
 	return 1;
 }
+
+/**
+ * @brief
+ *	Validates the sched object attributes changed from Server.
+ *
+ * @param[in] connector - socket descriptor to server
+ *
+ * @retval Error code
+ * @return 0 - Failure
+ * @return 1 - Success
+ *
+ * @par Side Effects:
+ *	None
+ *
+ *
+ */
+int
+validate_sched_attrs(int connector)
+{
+	struct batch_status *ss = NULL;
+	struct batch_status *all_ss = NULL;
+
+	/* Stat the scheduler to get details of sched */
+
+	all_ss = pbs_statsched(connector, NULL, NULL);
+	ss = bs_find(all_ss, sc_name);
+
+	if (ss == NULL) {
+		log_err(-1, __func__,  "Unable to retrieve the scheduler attributes from server");
+		pbs_statfree(all_ss);
+		return 0;
+	}
+	if (!sched_settings_frm_svr(ss)) {
+		pbs_statfree(all_ss);
+		return 0;
+	}
+
+	pbs_statfree(all_ss);
+
+	return 1;
+}
+
--- a/src/scheduler/fifo.c
+++ b/src/scheduler/fifo.c
@@ -570,7 +570,6 @@ schedule(int cmd, int sd, char *runjobid)
 			 * server through qmgr.
 			 */
 			if (!validate_sched_attrs(connector)) {
-				log_err(-1, __func__, "validate_sched_attrs failed");
 				return 0;
 			}
 			break;
@@ -2694,7 +2693,6 @@ update_svr_schedobj(int connector, int cmd, int alarm_time)
 		return 1;
 
 	if (!validate_sched_attrs(connector)) {
-		log_err(-1, __func__, "validate_sched_attrs failed");
 		return 0;
 	}
 
@@ -2766,6 +2764,8 @@ validate_sched_attrs(int connector)
 	ss = bs_find(all_ss, sc_name);
 
 	if (ss == NULL) {
+		snprintf(log_buffer, sizeof(log_buffer), "Unable to retrieve the scheduler attributes from server");
+		log_err(-1, __func__, log_buffer);
 		pbs_statfree(all_ss);
 		return 0;
 	}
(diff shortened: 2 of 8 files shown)


bhroam:
@subhasisb why are we blocking this until we fork?  This has to do with multi-sched, not multi-server.  This fixes certain bugs that would be nice to get in sooner rather than later.

subhasisb:
> @subhasisb why are we blocking this until we fork? This has to do with multi-sched, not multi-server. This fixes certain bugs that would be nice to get in sooner rather than later.

Hi Bhroam, we can certainly merge it if we think this is already super high quality and benefits, but at this point the idea was to not put the upcoming release at any additional risk than necessary. So, if the maintainers strongly feel this is high risk free, we can go ahead. Thanks

bhroam:
@subhasisb It is as risky as any other high quality fix we put in.  It's probably less so than the mid-sized RFEs that are going in.  I don't see any problem.

subhasisb:
> @subhasisb It is as risky as any other high quality fix we put in. It's probably less so than the mid-sized RFEs that are going in. I don't see any problem.

Alright, please go ahead then. I will remove the block

suresh-thelkar:
Thanks @subhasisb and @bhroam. I have now removed the WIP tag.  

@bhroam, Requesting you to review this PR and provide your comments if any.

src/scheduler/fifo.c
```diff
@@ -569,8 +569,8 @@ schedule(int cmd, int sd, char *runjobid)
 			 * This is required since there is a probability that scheduler's configuration has been changed at
 			 * server through qmgr.
 			 */
-			if (!update_svr_schedobj(connector, 0, 0)) {
-				log_err(-1, __func__, "update_svr_schedobj failed");
+			if (!validate_sched_attrs(connector)) {
+				log_err(-1, __func__, "validate_sched_attrs failed");
```
bhroam: The failure message should be in this function itself.  You shouldn't log here.  If you are logging in the function, then you'll get 2 messages.

src/scheduler/pbs_sched.c
```diff
@@ -1483,8 +1484,11 @@ main(int argc, char *argv[])
 		cmd = server_command(&runjobid);
 
 		if (connector >= 0) {
-			/* update sched object attributes on server */
-			update_svr_schedobj(connector, cmd, alarm_time);
+			if (update_svr) {
```
bhroam: You don't actually have to do this.  It is already done in update_svr_schedobj().  There is a static flag called 'svr_knows_me'.  Once that is set, you don't update anything.  If you want to do it here, remove the flag in update_svr_schedobj().

src/scheduler/pbs_sched.c
```diff
@@ -118,6 +118,7 @@ struct		connect_handle connection[NCONNECTS];
 int		connector;
 int		server_sock;
 int		second_connection = -1;
+static int	update_svr = 1;
```
bhroam: There is no need to make this global.  Make it a local variable in main().

test/tests/interfaces/pbs_sched_interface_test.py
```diff
@@ -289,3 +289,14 @@ def test_set_default_sched_attrs(self):
...
                            id='default', max_attempts=10)
+
+    def test_scheduling_iteration(self):
+        """
+        Test scheduler_itration attribute after it is unset. It should go
+        to its default value which is 600. If this happens Server will not
+        kickoff infinite scheduling cycles.
+        """
+        self.server.manager(MGR_CMD_UNSET, SCHED,
+                            ATTR_schedit, id='TestCommonSched')
```
bhroam: You should probably set it to something else first and then unset it.  It's 600 by default, so you don't know if it remained 600 or it was unset and set back to 600.  While you are at it, you might want to check other defaults.  This test seems a bit too specific.

src/scheduler/fifo.c
suresh-thelkar: I have now removed the logging inside the function validate_sched_attrs.

src/scheduler/pbs_sched.c
suresh-thelkar: Thanks for letting me know this @bhroam. I have removed the flag svr_knows_me and modified the code accordingly.

src/include/server.h
```diff
@@ -282,7 +282,7 @@ extern int			sched_save_db(pbs_sched *, int mode);
 extern enum failover_state	are_we_primary(void);
 extern int			have_socket_licensed_nodes(void);
 extern void			unlicense_socket_licensed_nodes(void);
-extern void			set_sched_default(pbs_sched *, int unset_flag, int from_scheduler);
+extern void			set_sched_default(pbs_sched *,  int from_scheduler);
```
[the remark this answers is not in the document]
suresh-thelkar: Yes. I have removed a single space now.

src/scheduler/pbs_sched.c
suresh-thelkar: Moved this variable under main() function.

test/tests/interfaces/pbs_sched_interface_test.py
suresh-thelkar: Thanks for the suggestion. I have now written elaborated test that covers the above scenario as well.

suresh-thelkar:
@bhroam, I have implemented your review comments. Can you please take a look.

src/scheduler/pbs_sched.c
```diff
@@ -938,6 +937,7 @@ main(int argc, char *argv[])
 	int num_cores;
 	char *endp = NULL;
 	pthread_mutexattr_t attr;
+	static int update_svr = 1;
```
bhroam: This doesn't need to be static.  When main ends, so does the scheduler.

test/tests/interfaces/pbs_sched_interface_test.py
```diff
@@ -294,9 +294,29 @@ def test_scheduling_iteration(self):
...
+            self.server.pbs_conf['PBS_HOME'], 'sched_logs_TestCommonSched')
+        a = {'sched_port': 15051,
+             'sched_host': self.server.hostname,
+             'sched_priv': sched_priv,
+             'sched_log': sched_logs,
+             'scheduling': 'False',
+             'scheduler_iteration': 600,
+             'sched_cycle_length': '00:20:00'}
+        self.server.expect(SCHED, a, id='TestCommonSched',
+                           attrop=PTL_AND, max_attempts=10)
```
bhroam: No need for PTL_AND.  It is the default.

src/scheduler/pbs_sched.c
suresh-thelkar: Done

test/tests/interfaces/pbs_sched_interface_test.py
suresh-thelkar: Done

suresh-thelkar:
@bhroam, I have implemented your review comments. Please check.

src/scheduler/fifo.c
```diff
@@ -2750,12 +2731,49 @@ update_svr_schedobj(int connector, int cmd, int alarm_time)
...
+	struct batch_status *ss = NULL;
+	struct batch_status *all_ss = NULL;
+
+	/* Stat the scheduler to get details of sched */
+
+	all_ss = pbs_statsched(connector, NULL, NULL);
+	ss = bs_find(all_ss, sc_name);
+
+	if (ss == NULL) {
+		pbs_statfree(all_ss);
```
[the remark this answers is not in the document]
suresh-thelkar: The idea is not to log anything inside the function validate_sched_attrs. Instead we can log wherever it is called. So I have added the log message inside update_svr_schedobj() function. 
I thought it is not good to add the same log message i.e. "Unable to retrieve the scheduler attributes from server" because validate_sched_attrs() is where we refactored the code and it is now doing more than fetching scheduler stats.

suresh-thelkar:
> LGTM

@bhroam, Thanks for your sign off. I have just added a log message to the function update_svr_schedobj when validate_sched_attrs fails. Please take a look.

suresh-thelkar:
@agrawalravi90, Thanks for the review. I have added the corresponding log message. Please take a look.

src/scheduler/fifo.c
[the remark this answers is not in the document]
agrawalravi90: Ok, but ss == NULL here actually does imply "Unable to retrieve the scheduler attributes from server" right? A generic message saying "validate_sched_attrs failed" might not be as useful. One option is to do "validate_sched_attrs: Unable to retrieve the scheduler attributes from server"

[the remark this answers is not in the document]
bhroam: @suresh-thelkar the scheduler generally logs where the error happens itself and not in the caller.  This way we can get a more specific reason of what went wrong.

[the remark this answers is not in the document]
suresh-thelkar: @bhroam / @agrawalravi90, I have now moved the log message inside the function validate_sched_attrs itself. Please provide further review comments if any or sign off.

src/scheduler/fifo.c
```diff
@@ -2766,6 +2764,8 @@ validate_sched_attrs(int connector)
 	ss = bs_find(all_ss, sc_name);
 
 	if (ss == NULL) {
+		snprintf(log_buffer, sizeof(log_buffer), "Unable to retrieve the scheduler attributes from server");
+		log_err(-1, __func__, log_buffer);
```
bhroam: This is a static string, just put it into log_err itself.

agrawalravi90: oops! I merged this PR already, I didn't realize that there was a commit that you hadn't reviewed yet, sorry about that!

suresh-thelkar: @bhroam, I am ready to implement this comment but I don't know whether I can since the PR is already merged. Since having a variable length string does not harm I am thinking to leave this as is for now and implement this comment in upcoming PR. Hope this is fine.

bhroam: It doesn't harm anything, but this was a common thing you did when you implemented multi-sched.  I just went through and fixed them all to call log_event() directly.  I'd hate to add another.  I guess we'll fix it later.

PR #1504, opened 2020-02-06
Scheduler does not requeue jobs for preemption if rrtros is set

vstumpf:
#### Describe Bug or Feature
If a high priority job that requests a static resource is submitted, and the scheduler had preempt order set to R, AND restrict_res_to_release_on_suspend was set, the scheduler would not preempt jobs to run the high priority job.

#### Describe Your Change
Instead of checking if the preemption type is S to use the released resource array, check if it is filled. If it is, use it, otherwise, use all the resources.

[after-fix.txt](https://github.com/PBSPro/pbspro/files/4162611/after-fix.txt)
[before-fix.txt](https://github.com/PBSPro/pbspro/files/4162612/before-fix.txt)

### DIFF
--- a/src/scheduler/queue_info.c
+++ b/src/scheduler/queue_info.c
@@ -151,8 +151,8 @@ query_queues(status *policy, int pbs_sd, server_info *sinfo)
 	if ((queues = pbs_statque(pbs_sd, NULL, NULL, NULL)) == NULL) {
 		errmsg = pbs_geterrmsg(pbs_sd);
 		if (errmsg == NULL)
-			errmsg = "";	
-	log_eventf(PBSEVENT_SCHED, PBS_EVENTCLASS_QUEUE, LOG_NOTICE, "queue_info", 
+			errmsg = "";
+	log_eventf(PBSEVENT_SCHED, PBS_EVENTCLASS_QUEUE, LOG_NOTICE, "queue_info",
 			"Statque failed: %s (%d)", errmsg, pbs_errno);
 		free_schd_error(sch_err);
 		return NULL;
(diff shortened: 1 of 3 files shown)


bhroam:
pycodestyle is complaining about too many blank lines.

vstumpf:
boo i thought i checked it

src/scheduler/queue_info.c
```diff
@@ -666,7 +666,7 @@ update_queue_on_run(queue_info *qinfo, resource_resv *resresv, char *job_state)
 			multi_node_sort);
 
 
-	if ((job_state != NULL) && (*job_state == 'S'))
+	if ((job_state != NULL) && (*job_state == 'S') && (resresv->job->resreq_rel != NULL))
```
agrawalravi90: This is a side effect of assuming preemption method "S" internally in sched when calling update_universe_on_end() for any preemption method. Ideally, job_state wouldn't be "S" for preempt method 'R' right? I'd wanted to change this in the past, but I think it wasn't worth the effort at the time, so I just wanted to check, is it worth fixing it now? We might be able to avoid such bugs in the future if we fix this.

test/tests/functional/pbs_preemption.py
```diff
@@ -676,3 +676,33 @@ def test_preemption_priority_escalation_2(self):
         self.server.expect(JOB, {'job_state': 'S'}, id=jid_list[1])
         self.server.expect(JOB, {'job_state': 'S'}, id=jid_list[3])
         self.server.expect(JOB, {'job_state': 'S'}, id=jid_list[4])
+
+    @skipOnCpuSet
+    def test_preempt_requeue_static_resc(self):
+        """
+        Test that scheduler will preempt jobs for static resources
```
agrawalravi90: Can you also mention RRTROS and preempt method R here ?

agrawalravi90: Also, we don't really qualify static resources specially, we just call them resources and qualify dynamic resources instead right? I had to double check to make sure that "static resource" was just regular resources. So, I'd advice just calling it 'resource', but it's up to you :)

src/scheduler/queue_info.c
agrawalravi90: Actually now that i look at the code, i see that update_universe_on_end() is in fact not called with job state "S" all the time. So, why is the job_state 'S' for preempt method R  ?

bhroam: We only call update_universe_on_end() with job state S inside the simulation.  Outside the simulation after we have preempted jobs, we correctly call it with job state Q.  Inside the simulation, it doesn't matter what we call it with because we won't look at the job again.  This isn't really an additional bug to what you fixed, but an extension of it.  What you fixed was this bug for nodes.  We just didn't realize it was also broken for queues and the server back then.  I think Vincent's current fix is fine.

agrawalravi90:
Thanks for addressing my comments Vincent. Can you also please change the title of the PR to summarize the bug instead of the fix?

PR #1505, opened 2020-02-06
Introduce LibAuth and refactoring in DIS/TPP support routines

hirenvadalia:
#### Describe Bug or Feature
* Refactor DIS support routines for various transports
* Support auth, encrypt and decrypt in TPP
* Introduce LibAuth

#### Link to Design Doc
Links to pages which will describe the changes:
* LibAuth API Interface:
  * Design Page: [Here](https://pbspro.atlassian.net/wiki/spaces/PD/pages/1506279513/LibAuth+API+Interface+design)
  * Forum Discussion: [Here](http://community.pbspro.org/t/libauth-api-interface/1984)
* This page explains other changes and how LibAuth is used by PBS: [Here](https://pbspro.atlassian.net/wiki/spaces/PD/pages/1510604832/PR-1505+Introduce+LibAuth+and+refactoring+in+DIS+TPP+support+routines)
* Design for new pbs.conf switches for multiple auth: [Here](https://pbspro.atlassian.net/wiki/spaces/PD/pages/1525776385/Allow+multiple+auth+methods+in+PBSPro)
* Forum discussion for multiple auth: [Here](http://community.pbspro.org/t/allow-multiple-auth-methods-in-pbspro/2010)

* Dirty local hack to PTL for switching between different auth method (not added in changes): [ptl_auth_hack.txt](https://github.com/PBSPro/pbspro/files/4164165/ptl_auth_hack.txt)
* [smoke_gss.txt](https://github.com/PBSPro/pbspro/files/4164169/smoke_gss.txt)
* [smoke_munge.txt](https://github.com/PBSPro/pbspro/files/4164170/smoke_munge.txt)
* For resvport auth, Travis CI will be sufficient.


hirenvadalia:
I am still creating documents to explain changes, till that time, marking PR as WIP

hirenvadalia:
The document is ready so removing WIP tag.

hirenvadalia:
Rebased with the master + Bug fixes found during more testing + replaced auth_set_logger() with auth_set_config() as per latest LibAuth API specification...

src/include/batch_request.h
```diff
@@ -410,7 +403,7 @@ extern void  req_confirmresv(struct batch_request *req);
 extern void  req_connect(struct batch_request *req);
 extern void  req_defschedreply(struct batch_request *req);
 extern void  req_locatejob(struct batch_request *req);
-extern void  req_manager(struct batch_request *req);
+extern void  req_manager(struct batch_request *req, conn_t *conn);
```
subhasisb: Why is req_manager the only IFL api handler that needs a conn_t *?

hirenvadalia: @subhasisb Thanks for catching this, reverted signature...

hirenvadalia:
Test logs as per latest changes:

[smoke_gss.txt](https://github.com/PBSPro/pbspro/files/4190135/smoke_gss.txt)
[smoke_munge.txt](https://github.com/PBSPro/pbspro/files/4190137/smoke_munge.txt)

src/lib/Libtpp/tpp_client.c
```diff
@@ -4224,6 +4330,102 @@ leaf_pkt_handler(int tfd, void *data, int len, void *ctx, void *extra)
...
+		}
+
+		if (len_out == 0) {
+			pbs_asprintf(&msgbuf, "invalid decrypted data len: %d, pktlen: %d", len_out, len - 1);
+			tpp_log_func(LOG_CRIT, __func__, msgbuf);
+			free(msgbuf);
+			return -1;
+		}
+
+		free(data);
```
vchlum: Are you sure about this free()? I am surprised it does not cause an error... AFAICS: data = pkt_start + sizeof(int) and pkt_start = conn->scratch.data, which is freed in free_phy_conn(). Am I missing something?

hirenvadalia: @vchlum Thank you! You are right, we don't need this free(). And till now it didn't give any error because we don't have any pre/post pkt_handler to encrypt/decrypt data in the router so router never sends encrypted data to leaf, till now the only leaf->router was sending encrypted data.

hirenvadalia:
Test logs as per latest changes:
[smoke_gss.txt](https://github.com/PBSPro/pbspro/files/4202916/smoke_gss.txt)
[smoke_munge.txt](https://github.com/PBSPro/pbspro/files/4202918/smoke_munge.txt)

SudeshnaMoulik:
Will it be better if we move IFF authentication also part of Libauth (something like Libauth_resrv_port) ?

hirenvadalia:
> Will it be better if we move IFF authentication also part of Libauth (something like Libauth_resrv_port) ?

Yes, definitely it will be good if we move resvport in libauth_resvport.so + pbs_iff binary.
But I don't know how much work it is and I don't want to break anything related to resvport right now.
So, will take it up after this PR.

src/server/process_request.c
```diff
@@ -962,12 +925,7 @@ dispatch_request(int sfds, struct batch_request *request)
 			req_register(request);
 			break;
 
-		case PBS_BATCH_AuthenResvPort:
-			if (pbs_conf.auth_method == AUTH_MUNGE) {
-                                req_reject(PBSE_BADCRED, 0, request);
-                                close_client(sfds);
-                                return;
-                        }
+		case PBS_BATCH_Authenticate:
```
SudeshnaMoulik: Now instead of "PBS_BATCH_AuthenResvPort and PBS_BATCH_AuthExternal" we will be using "PBS_BATCH_Authenticate".  Then why are we handling only req_authenResvPort() inside this (PBS_BATCH_Authenticate) ?

src/include/pbs_ifl.h
```diff
@@ -486,13 +486,6 @@ enum batch_op {	SET, UNSET, INCR, DECR,
 	EQ, NE, GE, GT, LE, LT, DFLT
 };
 
-/* PBS supported authentication methods */
-enum pbs_auth_method {
-	AUTH_RESV_PORT,      /* Reserved port authentication */
-	AUTH_MUNGE,           /* MUNGE Authentication */
-	AUTH_GSS           /* GSS Authentication */
-};
-
```
SudeshnaMoulik: If we remove this and we also don't have any PBS_SUPPORTED_AUTH_METHODS currently, then how do we invoke the appropriate libauth_xxx.so api ?

src/lib/Libifl/pbs_loadconf.c
```diff
@@ -582,18 +583,13 @@ __pbs_loadconf(int reload)
...
-				   pbs_conf.auth_method = AUTH_MUNGE;
-				} else if (!strcasecmp(conf_value, "GSS")) {
-					pbs_conf.auth_method = AUTH_GSS;
-				} else {
-					fprintf(stderr, "pbsconf error: illegal value for %s\n",PBS_CONF_AUTH);
-					goto err;
+				int i = 0;
+				memset(pbs_conf.auth_method, '\0', sizeof(pbs_conf.auth_method));
+				for (i = 0; i < strlen(conf_value); i++) {
+					pbs_conf.auth_method[i] = tolower(conf_value[i]);
```
SudeshnaMoulik: How will we validate "pbs_conf.auth_method" is valid supported auth method (like GSS, Munge or any future authentication method) ?

src/server/process_request.c
hirenvadalia: for other than resvport, it's handled in process_request().

src/include/pbs_ifl.h
hirenvadalia: load_auth_lib make use of pbs_conf.auth_method to get .so name.

src/lib/Libifl/pbs_loadconf.c
hirenvadalia: if the value is not resvport then it will go into load_auth_lib() and there it will try to load .so file, if the file doesn't exist or not an libauth library then it will fail, which eventually considered as not supported auth method. Means if load_auth_lib() succeeds then its supported method.

src/cmds/Makefile.am
```diff
@@ -86,6 +86,10 @@ dist_bin_SCRIPTS = \
 	pbsrun_unwrap \
 	pbsrun_wrap
 
+common_cflags = \
+	-I$(top_srcdir)/src/include \
+	@KRB5_CFLAGS@
```
subhasisb: This is kerberos specific. Since our goal is to make the core code free of specific authentication/encryption calls, this should also be made generic. @vchlum perhaps we need your help in this area. Also we might need help is carving out the other kerberos specific functionality like ticket renewal etc into a kerberos specific library.

vchlum: Sure @subhasisb, I am willing to help with that. What is your intention - do you want to refactor it as a part of this PR?

subhasisb: hi @vchlum - thanks. I did not mean we need to do as part of this PR - a separate PR is also fine. The point is, how do  we take out the individual authentication related code from the core of PBS and make it part of a library - it may even need an extension to the proposed libAuth library interface.

hirenvadalia:
Rebased with the master as I needed fix from https://github.com/PBSPro/pbspro/pull/1534 for windows testing...

hirenvadalia:
Few more changes coming into this PR so marking as WIP:
Changes coming:
* Multiple auth method support in PBS
* Separate method for authentication and encrypt/decrypt
* Control switch on what to encrypt/decrypt
* Pass auth error message given by libauth back to the client from server

hirenvadalia:
All pending changes pushed, so removing WIP tag...

hirenvadalia:
Test logs as per the latest changes:
Filename format: `smoke_<auth method name>_<encrypt method name>_<encrypt mode>.txt`
Since as of now we have only one method which supports encrypt/decrypt so \<encrypt method name> will be always `gss`

[smoke_gss_gss_0.txt](https://github.com/PBSPro/pbspro/files/4271458/smoke_gss_gss_0.txt)
[smoke_gss_gss_1.txt](https://github.com/PBSPro/pbspro/files/4271459/smoke_gss_gss_1.txt)
[smoke_gss_gss_2.txt](https://github.com/PBSPro/pbspro/files/4271460/smoke_gss_gss_2.txt)
[smoke_munge_gss_0.txt](https://github.com/PBSPro/pbspro/files/4271461/smoke_munge_gss_0.txt)
[smoke_munge_gss_1.txt](https://github.com/PBSPro/pbspro/files/4271462/smoke_munge_gss_1.txt)
[smoke_munge_gss_2.txt](https://github.com/PBSPro/pbspro/files/4271463/smoke_munge_gss_2.txt)
[smoke_resvport_gss_0.txt](https://github.com/PBSPro/pbspro/files/4271464/smoke_resvport_gss_0.txt)
[smoke_resvport_gss_1.txt](https://github.com/PBSPro/pbspro/files/4271465/smoke_resvport_gss_1.txt)
[smoke_resvport_gss_2.txt](https://github.com/PBSPro/pbspro/files/4271466/smoke_resvport_gss_2.txt)

src/include/libpbs.h
```diff
@@ -290,19 +301,19 @@ struct batch_reply {
...
-#define PBS_BATCH_AuthExternal	88
+/* Unused -- #define PBS_BATCH_AuthExternal	88 */
 #define PBS_BATCH_HookPeriodic  89
 #define PBS_BATCH_RelnodesJob	90
 #define PBS_BATCH_ModifyResv	91
 #define PBS_BATCH_ResvOccurEnd	92
 #define PBS_BATCH_PreemptJobs	93
-#define PBS_BATCH_Cred          94
+#define PBS_BATCH_Cred		94
+#define PBS_BATCH_Authenticate	95
```
vchlum: You mentioned the merge of PBS_BATCH_AuthenResvPort and PBS_BATCH_AuthExternal in the design doc. Since the PBS_BATCH_AuthenResvPort will be unused, I believe it should be explicitly documented that the backward compatibility with all older pbs clients is broken. Otherwise, admins could be surprised.

subhasisb:
@hirenvadalia can you update your branch and resolve the conflicts please

src/include/net_connect.h
```diff
@@ -65,7 +66,7 @@ typedef unsigned long pbs_net_t;        /* for holding host addresses */
 #define PBS_NET_CONN_NOTIMEOUT	   0x04
 #define PBS_NET_CONN_FROM_QSUB_DAEMON	0x08
 #define PBS_NET_CONN_FORCE_QSUB_UPDATE	0x10
-#define PBS_NET_CONN_GSSAPIAUTH 0x20
+#define PBS_NET_CONN_TO_SCHED	0x20
```
vchlum: Could it be 0x40 and PBS_NET_CONN_GSSAPIAUTH commented as unused?

src/lib/Libauth/gss/pbs_gss.c
```diff
@@ -0,0 +1,1184 @@
...
+ * @retval	!0 on error
+ */
+int
+pbs_auth_process_handshake_data(void *ctx, void *data_in, size_t len_in, void **data_out, size_t *len_out, int *is_handshake_done)
+{
+	pbs_gss_extra_t *gss_extra = (pbs_gss_extra_t *) ctx;
+	int rc = 0;
+
+	if (gss_extra == NULL) {
+		GSS_LOG_ERR("No GSS context given");
```
vchlum: I would call it the 'auth context'. 'GSS context' is gss_extra->gssctx.

src/lib/Libifl/pbs_loadconf.c
```diff
@@ -933,18 +958,59 @@ __pbs_loadconf(int reload)
...
+		free_string_array(pbs_conf.supported_auth_methods);
+		pbs_conf.supported_auth_methods = break_comma_list(value);
+		if (pbs_conf.supported_auth_methods == NULL) {
+			free(value);
+			goto err;
+		}
+		free(value);
+	}
+
+	if (pbs_conf.supported_auth_methods == NULL && strcmp(pbs_conf.auth_method, AUTH_RESVPORT_NAME) != 0) {
```
subhasisb: @hirenvadalia this would imply that the supported_auth_methods is populated at the client side as well. We can reduce another point of data duplication and make it mandatory only for the sever, no?

src/lib/Libauth/gss/pbs_gss.c
```diff
@@ -0,0 +1,1184 @@
...
+	}
+
+	if (len_in == 0) {
+		GSS_LOG_ERR("No data available to decrypt");
+		return PBS_GSS_ERR_INTERNAL;
+	}
+
+	if (gss_extra->is_secure == 0) {
+		GSS_LOG_ERR("wrapped data ready but auth context is not secure");
+		return PBS_GSS_ERR_INTERNAL;
```
vchlum: With GSS, we can wrap data but do not encrypt the data. Technically, it is not en error. The original idea behind this treated as the error was that we don't want to send credentials on an insecure line (only wrapped). Since we have PBS_ENCRYPT_MODE and we can disable encryption of everything anyway. It is a question whether this should be treated as a warning only... Just saying... I am not sure, and we can discuss it. I would slightly prefer still treat it as the error in PBS. Theoretically and in general, you could have clients using wrapped data but not encrypted.

src/lib/Libifl/pbs_loadconf.c
hirenvadalia: @subhasisb So if I understand your point right then you are saying don't mandate supported_auth_methods in PBS config (or as env) on client-side, and just load given auth method and/or encrypt method if supported_auth_methods is not given, right?
If Yes, then yes we can do that but since PBS config parser code is shared by client and server both, it won't be only for the client, it will also apply to the server also.

src/include/libpbs.h
hirenvadalia: sure will update the design page with details.

src/lib/Libauth/gss/pbs_gss.c
hirenvadalia: I also feel that we should treat as an error in PBS as sending cred in just wrapped (not encrypted) is not a good idea. If the user/admin doesn't care about the encryption part then it can disable it via PBS_ENCRYPT_MODE anyway. So I think if PBS_ENCRYPT_MODE is > 0 and GSS/Kerberos is not configured for encryption then we should treat as an error in PBS.
(Just FYI: If PBS_ENCRYPT_MODE is 0 (aka disabled) then PBS will never call pbs_auth_encrypt_data() and pbs_auth_decrypt_data() for any auth lib)

@subhasisb and others What's ur thought on this?

src/lib/Libifl/pbs_loadconf.c
hirenvadalia: @subhasisb As per our offline discussion, I have changed code please have a look into latest commit.

src/server/process_request.c
```diff
@@ -406,19 +357,81 @@ process_request(int sfds)
...
+
+		if (strcmp(request->rq_user, conn->cn_username) != 0) {
+			pbs_errno = PBSE_BADCRED;
+			log_eventf(PBSEVENT_ERROR | PBSEVENT_FORCE, PBS_EVENTCLASS_REQUEST, LOG_CRIT, "",
+					"Unauthorized access!, request user = %s@%s, authenticated user = %s, sock=%d",
+					request->rq_user, request->rq_host, conn->cn_credid, request->rq_conn);
+			req_reject(PBSE_BADCRED, 0, request);
+			close_client(sfds);
+			return;
+		}
```
vchlum: This check is wrong, we don't want it with Kerberos. You actually can and should be allowed to use PBS clients with a principal that is not that same as the username. The correct principal can be host/pbspro.test.test@PBSPRO but the user name can be root and it is OK. ...I think that it should be even possible to submit a job under a user that has a different principal then the username.

Lets have a hook like this:
```
import pbs
node = pbs.server().vnode("torque1")
```
and 
```
Qmgr: l h test-hook
Hook test-hook
    type = site
    enabled = true
    event = exechost_periodic
    user = pbsadmin
    alarm = 30
    freq = 10
    order = 1
    debug = false
    fail_action = none
```

the hook will fail with something like:
```
03/11/2020 13:20:18;0100;Server@torque1;Req;;Type 21 request received from root@torque2.grid.cesnet.cz, sock=17
03/11/2020 13:20:18;0001;Server@torque1;Req;;Unauthorized access!, request user = root@torque2.grid.cesnet.cz, authenticated user = host/torque2.grid.cesnet.cz@EINFRA-SERVICES, sock=17
03/11/2020 13:20:18;0080;Server@torque1;Req;req_reject;Reject reply code=15019, aux=0, type=21, from root@torque2.grid.cesnet.cz
```

hirenvadalia: Yes, you are right @vchlum we don't need this check, as this should be handled in auth library implementation (if library wants to).

hirenvadalia:
@vchlum I have also applied a patch from you about removing the use of isatty() in init_pbs_client_ccache_from_keytab() (Thanks for the patch :smile:).
Please have look at the latest commit...

vchlum:
Thank you @hirenvadalia  for the changes. I found two more bugs (with gss):

* peer scheduling does not work: the peer scheduler logs:
```
03/11/2020 18:31:44;0080;pbs_sched;Job;28731.torque1.grid.cesnet.cz;Job is invalid - ignoring for this cycle No User
```
* the acl check behaves oddly: it does not work with gss:
```
set server managers += host/torque1.grid.cesnet.cz@META
```
does not work, but this:
```
set server managers += host/torque1.grid.cesnet.cz@torque1.grid.cesnet.cz
```
works... host/torque1.grid.cesnet.cz@torque1.grid.cesnet.cz does not make sense, it is not a principal...

Could you please check?

bhroam:
The 'No User' message happens when the job does not have an 'euser' attribute.  This can happen either because the attribute is missing, or the scheduler is not a manager at the remote machine.

vchlum:
Thank you for the explanation @bhroam! It makes sense. So, I have only one bug:-) The second one. Due to the changes, the scheduler now uses the gss and that is why `set server managers += root@fqdn` is not used anymore for peer scheduling. It should be `set server managers += <server_host_principal>`... but it does not work...
I tested the peer scheduling with `set server managers += host/torque1.grid.cesnet.cz@torque1.grid.cesnet.cz` and it works.

It means you can ignore the peer scheduling bug. It will be resolved together with resolving the ACL bug. I also think we should document that for peer scheduling server host principal should be set as manager.

hirenvadalia:
Thanks @vchlum, I will look into ACL bug.

hirenvadalia:
@vchlum So after debugging ACL bug, I found that my fix for marking root@server-host in case of GSS was causing that issue + my that fix was incorrect as I was assuming that conn->cn_username will never be something like "host/pbs-server.pbspro.org" (aka client is authenticated by using /tmp/krb5cc_pbs_client - host cred). So, for now, I have reverted that change.

Also while debugging one more thing I found with your latest patch about removal isatty(): If the user has run `kinit` for root user before starting PBS (aka existence of valid /tmp/krb5cc_0 cred for root user) then PBS daemons will use that and it will never create /tmp/krb5cc_pbs_client.
Is this expected? shouldn't PBS daemons always use /tmp/krb5cc_pbs_client (aka host cred) instead /tmp/krb5cc_0 (aka user cred)?

vchlum:
Right... @hirenvadalia  Hmm, I think that the desired behavior is to use  init_pbs_client_ccache_from_keytab() always with service-oriented connection, like:
* TPP
* scheduler TCP 
* pbs_python TCP

...and with PBS clients (user-oriented connection) we should check credentials with pbs_gss_can_get_creds() and if the credentials are not available then call init_pbs_client_ccache_from_keytab(). 
What do you think?

I would suggest the following solution how to get the information into the right place: 
* Use of pbs_auth_set_config() to also set the type of connection - enum {USER_ORIENTED, SERVICE_ORIENTED},
* or maybe we could add a new Libauth API interface - something more general - something like int pbs_auth_customize_data_into_ctx(void **ctx, ...) and this way we could set the gss_extra->init_client_ccache

What do you think?

hirenvadalia:
> ...and with PBS clients (user-oriented connection) we should check credentials with pbs_gss_can_get_creds() and if the credentials are not available then call init_pbs_client_ccache_from_keytab().
> What do you think?

I don't think we should use init_pbs_client_ccache_from_keytab() for user-oriented connection, as ccache_from_keytab() generates host cred not user cred right? and in user-oriented connection passing host cred is not a good idea as we have to validate user, not host.


> I would suggest the following solution how to get the information into the right place:
> 
> * Use of pbs_auth_set_config() to also set the type of connection - enum {USER_ORIENTED, SERVICE_ORIENTED},
> * or maybe we could add a new Libauth API interface - something more general - something like int pbs_auth_customize_data_into_ctx(void **ctx, ...) and this way we could set the gss_extra->init_client_ccache
> 
> What do you think?

I like set_config() one more than new API because even if we add new API we have to decide on what info (or arguments) we pass into it, and whatever we decide, it has to be agnostic of the actual implementation of auth libraries.

vchlum:
> > ...and with PBS clients (user-oriented connection) we should check credentials with pbs_gss_can_get_creds() and if the credentials are not available then call init_pbs_client_ccache_from_keytab().
> > What do you think?
> 
> I don't think we should use init_pbs_client_ccache_from_keytab() for user-oriented connection, as ccache_from_keytab() generates host cred not user cred right? and in user-oriented connection passing host cred is not a good idea as we have to validate user, not host.

Yes, it generates credentials for the host principal but any principal can be user of PBS, which is also the host principal. So it's still user-oriented. I think it is just like `root@fqdn` with resvport... The init_pbs_client_ccache_from_keytab() actually succeeds only for root due to permissions of /etc/krb5.keytab.

The benefit of this would be that you simply add `set server manager += host/pbs_server_fqdn@PBSPRO` and the pbs clients will work under root using gss without need to take care of krb ticket. We would not need to use `PBS_AUTH_METHOD=resvport PBS_ENCRYPT_MODE=0 qmgr` (only for initial settings) IMHO, that is a great feature. We could also add explicit check for geteuid() == 0 on the beginning of init_pbs_client_ccache_from_keytab()... for sure.

I do not insist but I think it makes sense and it would be very useful.

hirenvadalia:
> The init_pbs_client_ccache_from_keytab() actually succeeds only for root due to permissions of /etc/krb5.keytab.

I didn't know this part, so now it makes sense.

Also, I am trying to implement set_config() approach just for seeing how it will work while we wait for @subhasisb's thought on this.
I will send you a separate branch once it's working to test things on your side.

subhasisb:
> > The init_pbs_client_ccache_from_keytab() actually succeeds only for root due to permissions of /etc/krb5.keytab.
> 
> I didn't know this part, so now it makes sense.
> 
> Also, I am trying to implement set_config() approach just for seeing how it will work while we wait for @subhasisb's thought on this.
> I will send you a separate branch once it's working to test things on your side.

set_config approach sounds good to me @hirenvadalia thanks

PR #1507, opened 2020-02-06
Normalize node table code

nithinj:
This is a code refactoring and clean-up effort for better maintainability of code and preparing the codebase better suitable for future scalability enhancements. Node object was handled completely different in comparison to other objects for various reasons. This is an effort to generalize it.

Targets on:
* Remove special save functions for node.
Special save functions for individual attributes are not required as we are encoding only the modified flags while writing into the database. Encoded attributes will get updated in the table which is handled by Postgres internally. 

* Remove associated flags.
* Remove special cases within svr_to_db_node and invoke db_to_attr_db()
* Replace node_recov_db_raw() with node_recov_db().
raw load won't trigger action routines. which was required as some action routines such as indirect resource will refer to other nodes that may not have loaded yet. So action functions are called only after loading all nodes. This behaviour is changed to load the node when it is being referred.

* Additional node saves?
Node saves are not introduced in the job run path to avoid any performance penalty. It will only save only the updated attribute reducing the I/O. Right now static area will get saved anyway. Which will be solved using checksums in the future.

notes for maintainers:
It can be targeted for the next major release to buy more time for regressions.


agrawalravi90:
@nithinj I don't really see any value keeping this PR open if it won't get reviewed until we fork off 2020.1. What do you think?

nithinj:
@agrawalravi90 
I can close the PR if it cannot be reviewed. But I was hoping to complete the review by the time we fork a new release branch. Thanks!

agrawalravi90:
Oh, ok, I thought that we were going to go the integration branch route. Aren't you planning to put this in the integration branch as well?

nithinj:
I wasnt planning to. But I guess this changes will be helpful for the integration branch if we push the release further. Closing this now.

PR #1510, opened 2020-02-10
Add skipOnCpuset decorator to the unsupported tests 

Mahalty:
#### Describe Bug or Feature

- Few of the tests failed during regression on Cpuset machine.

#### Describe Your Change

- The tests were failing as they were changing either changing the value of 'resources_available.ncpus' or creating vnodes. On cpuset mom if we try to change the available ncpus value or create vnodes then we get the following msg "make_cpuset_inner, vnode uv-01: hv_ncpus (1) > mvi_acpus (0) (you are not expected to understand this)" which results in unexpected behavior of the test

- Need to skip those tests on Cpuset machine

- Also I have removed timeout from few tests as the default timeout is now 600 and in the test they were specifying timeout less than the default value

- In one of the test(pbs_hook_timeout.py) I have changed the expect function. Instead of checking exact 3 nodes to be free, I am now checking all the nodes to be free

- [Execution_logs_Test_run_count.txt](https://github.com/PBSPro/pbspro/files/4179376/Execution_logs_Test_run_count.txt)

- [Execution_logs_TestAdminSuspend.txt](https://github.com/PBSPro/pbspro/files/4179377/Execution_logs_TestAdminSuspend.txt)

- [Execution_logs_TestRunJobHook.txt](https://github.com/PBSPro/pbspro/files/4179378/Execution_logs_TestRunJobHook.txt)

- [Execution_logs_TestSoftWalltime.txt](https://github.com/PBSPro/pbspro/files/4179379/Execution_logs_TestSoftWalltime.txt)

- [Execution_logs_TestStrictOrderingAndBackfilling.txt](https://github.com/PBSPro/pbspro/files/4179380/Execution_logs_TestStrictOrderingAndBackfilling.txt)

- [Execution_logs_TestTrillionJobid.txt](https://github.com/PBSPro/pbspro/files/4179381/Execution_logs_TestTrillionJobid.txt)

### DIFF
--- a/test/tests/functional/pbs_hook_execjob_prologue.py
+++ b/test/tests/functional/pbs_hook_execjob_prologue.py
@@ -61,6 +61,9 @@ class TestPbsExecutePrologue(TestFunctional):
         self.hostB = self.momB.shortname
         self.hostC = self.momC.shortname
 
+        for mom in self.moms.values():
+            self.server.expect(NODE, {'state': 'free'}, id=mom.shortname)
+
     @skipOnCpuSet
     def test_prologue_execute_on_all_moms(self):
         """
--- a/test/tests/functional/pbs_hook_execjob_prologue.py
+++ b/test/tests/functional/pbs_hook_execjob_prologue.py
@@ -64,7 +64,6 @@ class TestPbsExecutePrologue(TestFunctional):
         for mom in self.moms.values():
             self.server.expect(NODE, {'state': 'free'}, id=mom.shortname)
 
-    @skipOnCpuSet
     def test_prologue_execute_on_all_moms(self):
         """
         Test to make sure execjob_prologue always get
@@ -77,12 +76,6 @@ class TestPbsExecutePrologue(TestFunctional):
                      "pbs.logjobmsg(e.job.id, 'executed prologue hook')\n")
         attr = {'event': 'execjob_prologue', 'enabled': 'True'}
         self.server.create_import_hook(hook_name, attr, hook_body)
-
-        attr = {'resources_available.ncpus': 1,
-                'resources_available.mem': '2gb'}
-        self.server.manager(MGR_CMD_SET, NODE, attr, id=self.hostA)
-        self.server.manager(MGR_CMD_SET, NODE, attr, id=self.hostB)
-        self.server.manager(MGR_CMD_SET, NODE, attr, id=self.hostC)
 
         attr = {'Resource_List.select': '3:ncpus=1',
                 'Resource_List.place': 'scatter',
--- a/test/tests/functional/pbs_hook_execjob_prologue.py
+++ b/test/tests/functional/pbs_hook_execjob_prologue.py
@@ -61,9 +61,7 @@ class TestPbsExecutePrologue(TestFunctional):
         self.hostB = self.momB.shortname
         self.hostC = self.momC.shortname
 
-        self.server.expect(VNODE, {'state=free': 3}, op=GE, max_attempts=10,
-                           interval=2)
-
+    @skipOnCpuSet
     def test_prologue_execute_on_all_moms(self):
         """
         Test to make sure execjob_prologue always get
--- a/test/tests/functional/pbs_hook_timeout.py
+++ b/test/tests/functional/pbs_hook_timeout.py
@@ -72,10 +72,6 @@ class TestHookTimeout(TestFunctional):
         self.server.manager(MGR_CMD_CREATE, NODE, id=self.hostC)
         for mom in self.moms.values():
             self.server.expect(NODE, {'state': 'free'}, id=mom.shortname)
-        """
-        self.server.expect(VNODE, {'state=free': 3}, op=EQ, count=True,
-                           max_attempts=10, interval=2)
-        """
 
     @timeout(600)
     def test_hook_send(self):
--- a/test/tests/functional/pbs_hook_timeout.py
+++ b/test/tests/functional/pbs_hook_timeout.py
@@ -73,7 +73,6 @@ class TestHookTimeout(TestFunctional):
         for mom in self.moms.values():
             self.server.expect(NODE, {'state': 'free'}, id=mom.shortname)
 
-    @timeout(600)
     def test_hook_send(self):
         """
         Test when the server doesn't receive an ACK from a mom for
--- a/test/tests/functional/pbs_hook_timeout.py
+++ b/test/tests/functional/pbs_hook_timeout.py
@@ -72,7 +72,7 @@ class TestHookTimeout(TestFunctional):
 
         self.server.manager(MGR_CMD_CREATE, NODE, id=self.hostC)
 
-
+    @skipOnCpuSet
     def test_hook_send(self):
         """
         Test when the server doesn't receive an ACK from a mom for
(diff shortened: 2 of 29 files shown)


Mahalty:
Execution-logs:

[Execution_logs.zip](https://github.com/PBSPro/pbspro/files/4229853/Execution_logs.zip)

test/tests/functional/pbs_hook_timeout.py
```diff
@@ -71,9 +70,12 @@ def setUp(self):
         self.server.manager(MGR_CMD_CREATE, NODE, id=self.hostB)
 
         self.server.manager(MGR_CMD_CREATE, NODE, id=self.hostC)
-
+        for mom in self.moms.values():
+            self.server.expect(NODE, {'state': 'free'}, id=mom.shortname)
```
prakashcv13: why is this change needed? the setUp() checks if number of moms is exactly 3 or not.

test/tests/functional/pbs_hook_timeout.py
```diff
@@ -71,9 +70,12 @@ def setUp(self):
...
+        """
         self.server.expect(VNODE, {'state=free': 3}, op=EQ, count=True,
                            max_attempts=10, interval=2)
+        """
```
prakashcv13: this is commented out. if we decide to check for "all" nodes to be in free state, we should remove the commented out code.

test/tests/functional/pbs_hook_execjob_prologue.py
```diff
@@ -61,9 +61,7 @@ def setUp(self):
         self.hostB = self.momB.shortname
         self.hostC = self.momC.shortname
 
-        self.server.expect(VNODE, {'state=free': 3}, op=GE, max_attempts=10,
-                           interval=2)
-
```
prakashcv13: do you mean to replace this code or simply remove it.

test/tests/functional/pbs_node_buckets.py
```diff
@@ -187,7 +187,7 @@ def test_multi_bucket2(self):
             self.assertTrue(self.colors[i] in
                             n[0]['resources_available.color'])
 
-    @timeout(450)
```
prakashcv13: did we check if 450 seconds made more sense than the default of 600?

prakashcv13:
@Mahalty - if setting of ncpus is the only issue - can we do something to skip only the setting of ncpus on cpuset machines. I see we are skipping the whole testsuite just because we set ncpus in setUp().
There are chances that the test cases are designed to work only if the no.of cpus is n. Can we check if that is not the case and allow the test to be run on cpuset machines. We would run more tests on cpuset machines by doing that, just a thought. What do you think?

test/tests/functional/pbs_hook_timeout.py
Mahalty: Removed.

test/tests/functional/pbs_node_buckets.py
Mahalty: This was added when the default timeout used to be 180s. I have gone through the tests and I didn't find anything that would get affected due to the change in the timeout value(450s to 600s). If the tests passes it will exit even if it hasn't hit timeout value.

test/tests/functional/pbs_hook_timeout.py
prakashcv13: but the test will get skipped if the no.of moms is not 3.

test/tests/functional/pbs_hook_timeout.py
```diff
@@ -71,8 +70,14 @@ def setUp(self):
...
         self.server.manager(MGR_CMD_CREATE, NODE, id=self.hostC)
+        for mom in self.moms.values():
+            self.server.expect(NODE, {'state': 'free'}, id=mom.shortname)
+        """
+        self.server.expect(VNODE, {'state=free': 3}, op=EQ, count=True,
+                           max_attempts=10, interval=2)
+        """
 
-    @skipOnCpuSet
+    @timeout(600)
```
prakashcv13: isn't 600 the default?

Mahalty:
[Execution_logs_TestPbsExecutePrologue.txt](https://github.com/PBSPro/pbspro/files/4232753/Execution_logs_TestPbsExecutePrologue.txt)
[Execution_logs_TestPbsExecutePrologue_latest.txt](https://github.com/PBSPro/pbspro/files/4232754/Execution_logs_TestPbsExecutePrologue_latest.txt)
[Execution_logs_TestServerDynRes.txt](https://github.com/PBSPro/pbspro/files/4232755/Execution_logs_TestServerDynRes.txt)
[Execution_logs_Test_array_job_email.txt](https://github.com/PBSPro/pbspro/files/4232756/Execution_logs_Test_array_job_email.txt)

test/tests/functional/pbs_hook_timeout.py
Mahalty: It wasn't intentional, added it by mistake. I have now reverted back the file to orignal test

Mahalty:
> @Mahalty - if setting of ncpus is the only issue - can we do something to skip only the setting of ncpus on cpuset machines. I see we are skipping the whole testsuite just because we set ncpus in setUp().
> There are chances that the test cases are designed to work only if the no.of cpus is n. Can we check if that is not the case and allow the test to be run on cpuset machines. We would run more tests on cpuset machines by doing that, just a thought. What do you think?

@prakashcv13 Thanks for the suggestion. I checked the tests and almost all the tests that are being skipped are either changing ncpus/set mem/set resources on node/create vnodes which is affecting the intent of the tests. The only tests that are not affected are "pbs_array_job_email.py,pbs_hook_execjob_prologue.py,pbs_svr_dyn_res(All the tests except 3)". I have  removed the skip decorator and tested. I have attached the logs (**Execution_logs_Test_array_job_email.txt, Execution_logs_TestPbsExecutePrologue_latest.txt, Execution_logs_TestServerDynRes**)

test/tests/functional/pbs_hook_timeout.py
Mahalty: @prakashcv13: The count in expect function considers the vnode as well, so the count differs with no of moms, whereas self.moms.values just consider the natural(base) nodes.

prakashcv13: Ok. On a cpuset system, if there is one base node and 3 vnodes, the test will get skipped, is that correct?

prakashcv13: never mind. I understand what this change does.

test/tests/functional/pbs_svr_dyn_res.py
```diff
@@ -46,8 +46,10 @@ class TestServerDynRes(TestFunctional):
     def setUp(self):
         TestFunctional.setUp(self)
         # Setup node
+        """
         a = {'resources_available.ncpus': 4}
         self.server.manager(MGR_CMD_SET, NODE, a, id=self.mom.shortname)
+        """
```
prakashcv13: shall we remove the commented code.

Mahalty:
**Execution-logs:
[Execution_logs_cpuset_branch_Oss_repo.zip](https://github.com/PBSPro/pbspro/files/4234402/Execution_logs_cpuset_branch_Oss_repo.zip)

**

prakashcv13:
@shilpakodli - @Mahalty has done some changes after you approved, could you please go through the changes and let us know if those changes look good to you.

PR #1511, opened 2020-02-10
Removing obsolete sort_queues from sched_config

agrawalravi90:
#### Describe Bug or Feature

https://pbspro.atlassian.net/browse/PP-419

sort_queues was obsoleted a long time back but it was still present in sched_config, which meant that scheduler would log the following message every time it reconfigured:
02/10/2020 17:19:05.824393;0040;pbs_sched;Sched;reconfigure;Scheduler is reconfiguring
02/10/2020 17:19:05.825235;0040;pbs_sched;Fil;sched_config;Obsolete config name sort_queues

Since the option has been obsolete for some time, I'm removing it so that we stop seeing this warning.

#### Describe Your Change

Just remove sort_queues from sched_config

Smoke tests should suffice, sched logs can be inspected to verify that the log message goes away now. It's not worth writing a PTL test for this.

### DIFF
--- a/src/scheduler/pbs_sched_config
+++ b/src/scheduler/pbs_sched_config
@@ -244,15 +244,6 @@ node_sort_key: "sort_priority HIGH"	ALL
 
 provision_policy: "aggressive_provision"
 
-#
-# sort_queues 
-#	sort queues by the priority attribute
-#
-#	PRIME OPTION
-#
-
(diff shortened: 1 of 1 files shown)


bhroam:
Thanks for doing this Ravi.  It should have happened a long time ago.

PR #1515, opened 2020-02-12
Add support for custom local file transfer commands via PBS_CP variable

jdburton:
#### Describe Bug or Feature

Currently, administrators can specify a custom command for the PBS MoM to use for remote file transfers by specifying the command in the PBS_SCP and PBS_RCP variables in /etc/pbs.conf. If these variables are unset, PBS will default to /bin/scp or <PBS_HOME>/sbin/pbs_rcp, respectively. 

There is no way to specify a custom command for local copies, however. PBS will always use /bin/cp on *nix based systems and xcopy on Windows systems. Administrators should have the ability to specify custom commands for local copies like they do for remote copies.
 
#### Describe Your Change

This change adds support for the PBS_CP variable in /etc/pbs.conf. If PBS_CP is unset, the MoM will default to /bin/cp on *nix systems and xcopy on Windows systems.

#### Link to Design Doc

https://pbspro.atlassian.net/wiki/spaces/PD/pages/1521680421/Add+support+for+custom+local+copy+via+the+PBS+CP+variable


src/lib/Libifl/pbs_loadconf.c
```diff
@@ -919,6 +924,22 @@ __pbs_loadconf(int reload)
 			goto err;
 		}
 	}
+	if (pbs_conf.cp_path == NULL) {
+
+#ifdef WIN32
+		if ((pbs_conf.cp_path =
+			malloc(6 * sizeof(char))) != NULL) {
```
bhroam: Some editors get wonky when braces don't match due to ifdefs.  I suggest you move the malloc() on top of the ifdef and malloc 8 bytes.  You can do your error checking then.  The only thing in the ifdef will be the sprintf() for each platform.

bhroam:
@jdburton thank you for your contribution to PBS.  A feature (even a small one like this) has a requirement to come with one or more PTL automated tests.  If you could write them, that would be great.

src/lib/Libifl/pbs_loadconf.c
jdburton: Suggested changes have been committed and pushed.

jdburton:
PBS_CP has been added to the pbs_testlib.py tests. The functionality of PBS_CP is similar to PBS_SCP, therefore, the test is similar.

Also PBS_CP documentation has been added to the appropriate man pages.

src/lib/Libifl/pbs_loadconf.c
hirenvadalia: how about using `strdup()` instead of `malloc()+sprintf()` like below:
```c
if (pbs_conf.cp_path == NULL) {
#ifndef WIN32
   char *cmd = "/bin/cp";
#else
   char *cmd = "xcopy";
#endif
   pbs_conf.cp_path = strdup(cmd);
   if (pbs_conf.cp_path == NULL)
       goto err;
}

Comment (jdburton) [reply src/lib/Libifl/pbs_loadconf.c]: It doesn't make much difference, but `strdup()` is a bit easier to read. Checking #ifdef WIN32 is more consistent than checking #ifndef. 

`	if (pbs_conf.cp_path == NULL) {
#ifdef WIN32
		char *cmd = 'xcopy';
#else
		char *cmd = "/bin/cp";
#endif
		pbs_conf.cp_path = strdup(cmd);
		if (pbs_conf.cp_path == NULL) {
			goto err;
		}
	}`

File: test/fw/ptl/utils/pbs_testsuite.py [hunk 2]
```diff
@@ -1264,6 +1283,7 @@ def revert_pbsconf(self):
             "PBS_START_MOM": None,
             "PBS_CORE_LIMIT": None,
             "PBS_SCP": None,
+            "PBS_CP": None,
```
hirenvadalia: I am not sure we should add PBS_CP here as it is not a default key-value of pbs.conf, and this dict should have only default key-value which gets added at the time of installation. @bhroam What you think?

src/lib/Libifl/pbs_loadconf.c
```diff
@@ -919,6 +924,17 @@ __pbs_loadconf(int reload)
 			goto err;
 		}
 	}
+	if (pbs_conf.cp_path == NULL) {
```
hirenvadalia: Before this shouldn't we check for "PBS_CP" environment variable to get its value as we do for other pbs.conf key-value?

jdburton: Yes. Good catch.

test/fw/ptl/utils/pbs_testsuite.py
jdburton: Whatever you and @bhroam think is best. I have no strong opinions either way.

jdburton:
Perhaps I am missing something, but I don't see any tests for PBS_SCP or PBS_RCP in the test/tests directory. The PBS_CP code is based on the PBS_SCP code, so the tests should be similar.

agrawalravi90:
> PBS_CP has been added to the pbs_testlib.py tests. The functionality of PBS_CP is similar to PBS_SCP, therefore, the test is similar.

pbs_testlib.py or pbs_testsuite.py are not tests:
- pbs_testlib.py contains Python routines which talk to PBS, it's used by multiple programs, one of which is pbs_benchpress, which is the program that we use to run tests.
- pbs_testsuite.py provides a Python testing framework that we use to write and run tests for PBS written in Python, it is a wrapper around Python's unittest: https://docs.python.org/3.8/library/unittest.html

test/tests/ is the directory that contains the actual tests, most of them go inside tests/functional directory.
> Perhaps I am missing something, but I don't see any tests for PBS_SCP or PBS_RCP in the test/tests directory. The PBS_CP code is based on the PBS_SCP code, so the tests should be similar.
You are correct, there are currently no tests for PBS_SCP/PBS_RCP. The reason being, PBS has existed for way longer than our test framework has. We started writing tests based off of our Python framework much later. So, we write tests for any new features added to PBS, and any bugfixes made to existing features. Each new test adds better coverage.

Anyways, since you are adding a new feature, it would be great if you could add a PTL test for it. If you need pointers about PTL, we can help. Here's some documentation if you'd like to learn more about how to write tests: https://pbspro.atlassian.net/wiki/spaces/DG/pages/1324417323/PTL+How-Tos

test/fw/ptl/utils/pbs_testsuite.py
bhroam: It sounds like @hirenvadalia is right.  This should be for just the things in pbs.conf that are set by default when PBS starts for the first time.  PBS_CP is something that is added by the admin.

hirenvadalia:
@jdburton Are you still working on this PR?

jdburton:
I still think this is a worthwhile change, I just haven't had the time to
commit to learning the testing system in order to write the tests that you
want for it.

agrawalravi90:
@jdburton the maintainers discussed and it's ok if you'd not like to write a PTL test for this. Please address any other review comments on the PR so that this can be signed-off and merged.

test/fw/ptl/utils/pbs_testsuite.py
agrawalravi90: @jdburton Seems like this is the only outstanding comment left, please address this.

jdburton: Changes have been made.

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -3,40 +3,37 @@
 # Copyright (C) 1994-2020 Altair Engineering, Inc.
 # For more information, contact Altair at www.altair.com.
 #
-# This file is part of both the OpenPBS software ("OpenPBS")
-# and the PBS Professional ("PBS Pro") software.
+# This file is part of the PBS Professional ("PBS Pro") software.
```
hirenvadalia, line 6: @jdburton Looks like merge/rebase issue in this file, please revert changes in this file.

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -456,31 +449,18 @@ def setUpClass(cls):
         cls.init_param()
         cls.check_users_exist()
         cls.init_servers()
-        cls.init_schedulers()
```
agrawalravi90, line 459: I'm not sure I understand why you made changes to this file, can you please clarify?

test/fw/ptl/utils/pbs_testsuite.py
jdburton: Yes, it was a merge issue. I restored the file to what is in master because no changes are necessary.

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -173,16 +170,12 @@ def wrapper(self, *args, **kwargs):
...
     """
 
     def wrapper(self, *args, **kwargs):
-        for mom in self.moms.values():
-            if mom.is_cpuset_mom():
-                msg = 'capability not supported on cgroup cpuset system: ' +\
-                      mom.shortname
-                self.skipTest(reason=msg)
-                break
+        if self.mom.is_cpuset_mom():
```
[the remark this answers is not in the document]
jdburton: it was a merge issue. I restored the file to what is in master because no changes are necessary.

test/fw/ptl/utils/pbs_testsuite.py
jdburton: it was a merge issue. I restored the file to what is in master because no changes are necessary.

agrawalravi90:
@jdburton it seems like something went wrong again. Please make sure that your local 'master' branch is exactly in sync with the master branch from the main repository and doesn't have any of your changes. Then, follow these steps:
git checkout pbs_cp
git checkout master -- test/fw/ptl/utils/pbs_testsuite.py
git commit -m "Reverting pbs_testsuite.py"

Then, push to your fork

I think this will fix it, please give it a shot, thanks.

bhroam:
Not me.  My only comment was about the PTL tests and that has been resolved.

PR #1520, opened 2020-02-15
Fix sigpipe handler in pbs_sched

vstumpf:
#### Describe Bug or Feature
scheduler was not handling sigpipe correctly

#### Describe Your Change
Py_Initialize sets the handler of SIGPIPE to SIG_IGN. See [Signal#General-rules](https://docs.python.org/3/library/signal.html#general-rules) and [Py_InitializeEx](https://docs.python.org/3/c-api/init.html#c.Py_InitializeEx)

Changing Py_Initialize to Py_InitializeEx(0) tells python to not ignore SIGPIPE.

[after-fix.txt](https://github.com/PBSPro/pbspro/files/4207624/after-fix.txt)
[before-fix.txt](https://github.com/PBSPro/pbspro/files/4207625/before-fix.txt)

### DIFF
--- a/test/tests/functional/pbs_sched_signal.py
+++ b/test/tests/functional/pbs_sched_signal.py
@@ -42,7 +42,7 @@ class TestSchedSignal(TestFunctional):
 
     def test_sigpipe(self):
         """
-        Test that pbs_sched handles sigpipe correctly
+        Test that pbs_sched receives a SIGPIPE correctly and it is not ignored
         """
         self.scheduler.signal('-PIPE')
         self.scheduler.log_match("We've received a sigpipe:")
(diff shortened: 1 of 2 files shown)


test/tests/functional/pbs_sched_signal.py
```diff
@@ -0,0 +1,48 @@
...
+# trademark licensing policies.
+
+from tests.functional import *
+
+
+class TestSchedSignal(TestFunctional):
+
+    def test_sigpipe(self):
+        """
+        Test that pbs_sched handles sigpipe correctly
```
lisa-altair: I think the test description should be changed to something like "Test that pbs_sched is able to received a SIGPIPE" because the test isn't verifying any handling of the sigpipe...

arungrover:
Are there any side effects of using the new initialize method?

vstumpf:
Python won't handle SIGPIPE internally, it will go to the main process. So if we're evaluating python code in the scheduler and it encounters a SIGPIPE, it will go to the scheduler instead of python handling it and throwing a IOException. 
Because we only use python in the scheduler for formulas, I don't see this affecting current behavior.

Issue #1522, opened 2020-02-17
pbs_benchpress -> ptl_test_db.py JSONDecodeError


ericpershey:
When running the full test suite, it fails to complete due to corrupted json at different points on debian9.  We have seen this in the other distros, but don't have the output.  Specifically we are using the debian9 docker image with python 3.5.9 pulled from python.org and compiled with valgrind.

```
export PYTHON3_MAJOR_VERSION=3.5
export PYTHON3_VERSION=3.5.9
export PBS_BUILD_DIR=/tmp/build-pbspro
export PBS_INSTALL_BASE=/opt
export PBS_INSTALL_DIR=$PBS_INSTALL_BASE/pbs
export PTL_INSTALL_DIR=$PBS_INSTALL_BASE/ptl
export PYTHON3_INSTALL_DIR=$PBS_INSTALL_BASE/python-$PYTHON3_VERSION
export PATH=$PYTHON3_INSTALL_DIR/bin:$PBS_INSTALL_DIR/bin:$PTL_INSTALL_DIR/bin:$PATH
export PYTHONPATH=$PTL_INSTALL_DIR/lib/python$PYTHON3_MAJOR_VERSION/site-packages:$PYTHONPATH
# --- START PYTHON INSTALL ---
cd /tmp
wget https://www.python.org/ftp/python/$PYTHON3_VERSION/Python-$PYTHON3_VERSION.tgz
tar xzf Python-$PYTHON3_VERSION.tgz
cd Python-$PYTHON3_VERSION
./configure --prefix=$PYTHON3_INSTALL_DIR --without-pymalloc --with-pydebug --with-valgrind
make -j 8
sudo make install
sudo $PYTHON3_INSTALL_DIR/bin/pip3 install --upgrade pip
sudo $PYTHON3_INSTALL_DIR/bin/pip3 install nose
sudo $PYTHON3_INSTALL_DIR/bin/pip3 install pexpect
# --- END PYTHON INSTALL ---
```

```time $PTL_INSTALL_DIR/bin/pbs_benchpress --db-name $PBS_BUILD_DIR/test/tests/ptl_test_results.json -f $PTL_INSTALL_DIR/tests -o $PBS_BUILD_DIR/test/tests/ptl_output_all.txt |& tee out.txt```

Below is the Traceback:
```
2020-02-17 15:25:08,701 INFO     expect on server pbsdev-mgmt-ptl-debian9: job_state set 0 job ...  OK
[...]

PR #1523, opened 2020-02-17
pbs_benchpress -> ptl_test_db.py JSONDecodeError

ericpershey:
#### Describe Bug or Feature
The JSON database was being overwritten, but the size of the file was
not being reduced, this was leaving garbage at the end of the file and
causing the JSON parser to fail.

#### Describe Your Change
Truncated the file to prevent the file from having leftover garbage.
The default type is set to "json", therefore corrected the "default"
in pbs_benchpress.  Also set __dbtype to None as the default is 
"json" and 'JSONDb' was not valid.

This is a fix for issue #1522

### DIFF
--- a/test/fw/ptl/utils/plugins/ptl_test_db.py
+++ b/test/fw/ptl/utils/plugins/ptl_test_db.py
@@ -1622,6 +1622,7 @@ class JSONDb(DBType):
             self.__dbobj[_TESTRESULT_TN].seek(0)
             jdata = json.load(self.__dbobj[_TESTRESULT_TN])
         jsondata = self.res_data.get_json(data=data, prev_data=jdata)
+        self.__dbobj[_TESTRESULT_TN].truncate(0)
         self.__dbobj[_TESTRESULT_TN].seek(0)
         json.dump(jsondata, self.__dbobj[_TESTRESULT_TN], indent=2)
 
@@ -1637,6 +1638,7 @@ class JSONDb(DBType):
             df = json.load(self.__dbobj[_TESTRESULT_TN])
             dur = str(result.stop - result.start)
             df['test_summary']['test_duration'] = dur
+            self.__dbobj[_TESTRESULT_TN].truncate(0)
             self.__dbobj[_TESTRESULT_TN].seek(0)
             json.dump(df, self.__dbobj[_TESTRESULT_TN], indent=2)
         for v in self.__dbobj.values():
@@ -1657,7 +1659,7 @@ class PTLTestDb(Plugin):
     def __init__(self):
         Plugin.__init__(self)
         self.__dbconn = None
-        self.__dbtype = 'JSONDb'
+        self.__dbtype = None
         self.__dbpath = None
         self.__dbaccess = None
         self.__dbmapping = {'file': FileDb,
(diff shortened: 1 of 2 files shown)


toonen:
@agrawalravi90 I'm not sure what additional information you are looking for.  We did file an issue (#1522 ), which described the problem in detail, and then amended it with a test case to reproduce the problem.  The proposed fix in the PR is two lines to fix the problem (the calls to truncate) and three lines of other minor related issues.  What would you like us to explain further?

agrawalravi90:
@toonen Sorry about the confusion, I was trying to give general information and not just specific to this PR. For this PR, can you just change the title of the PR and the commit to summarize the bug (maybe make it the same as the title of the Github Issue)?

ericpershey:
I will change it shortly.  We have found another issue and a push is coming soon.  There is some kind of file race that still exists after those changes and takes longer to find.

test/fw/ptl/utils/plugins/ptl_test_db.py
```diff
@@ -1615,16 +1615,19 @@ def __init__(self, dbtype, dbpath, dbaccess):
         self.res_data = PTLJsonData(command=self.__cmd)
 
     def __write_test_data(self, data):
-        jdata = None
         if _TESTRESULT_TN not in self.__dbobj.keys():
-            self.__dbobj[_TESTRESULT_TN] = open(self.dbpath, 'w+')
+            self.__dbobj[_TESTRESULT_TN] = self.dbpath
```
agrawalravi90: isn't self.__dbobj[_TESTRESULT_TN] redundant now? you already have the instance variable self.dbpath, and it seems like self.__dbobj[_TESTRESULT_TN] only gets set to one value: self.dbpath. Why not just do open("self.dbpath") instead in this class?

test/fw/ptl/utils/plugins/ptl_test_db.py
```diff
@@ -1615,15 +1615,19 @@ def __init__(self, dbtype, dbpath, dbaccess):
...
-            jdata = json.load(self.__dbobj[_TESTRESULT_TN])
+            dbobj = self.__dbobj[_TESTRESULT_TN]
+            with open(dbobj, 'r') as fd:
+                jdata = json.load(fd)
         jsondata = self.res_data.get_json(data=data, prev_data=jdata)
-        self.__dbobj[_TESTRESULT_TN].seek(0)
-        json.dump(jsondata, self.__dbobj[_TESTRESULT_TN], indent=2)
+        with open(dbobj, 'w') as fd:
+            json.dump(jsondata, fd, indent=2)
+            fd.flush()
```
agrawalravi90: do you need the flush()? dump() also does a write(), so the next write() should just get buffered behind that, and all flushed when with block ends, right?

ericpershey: I think you are correct, I will remove the flush().

test/fw/ptl/utils/plugins/ptl_test_db.py
ericpershey: Yes, it is redundant now.  I will remove self.__dbobj.

ericpershey:
Updated, @agrawalravi90 can you take a look again.  Cleaned it up even more.  Also, how does the title, summary and issue look?

PR #1524, opened 2020-02-18
Set a function in env variables of a remote machine

birbalsain:
#### Describe Bug or Feature
On a remote machine, pbs -V doesn't find  "foo" in env variable  because function "foo" is not set in env variable of a remote machine

#### Describe Your Change
set a function "foo" in env variables of remote machine

[env_variable_plat1.txt](https://github.com/PBSPro/pbspro/files/4218791/env_variable_plat1.txt)
[env_variable_plat2.txt](https://github.com/PBSPro/pbspro/files/4218792/env_variable_plat2.txt)


birbalsain:

PR #1526, opened 2020-02-19
[WIP] performance: don't do db saves for accrue_type & unnecessary job state updates

agrawalravi90:
#### Describe Bug or Feature

I noticed that alterjob requests for accrue_type were taking much longer than those for job comment because we call svr_setjobstate(), which does a job_save_db() for accrue_type, but not for job comment. We call svr_setjobstate() even if the job state & substate are the same as they were before, which seems unnecessary. But it seems like svr_setjobstate() is used as a way to do job_save_db() indirectly.

I'm proposing that we make accrue_type and eligible_time ATR_DFLAG_NOSAVM attributes. eligible_time is anyways not saved to db, so I think we can just make accrue_type a no-save attr as well and save on the job saves.

#### Describe Your Change

- Added a check so that we only set job state/substate if they are different.
- Made accrue_type and eligible_time ATR_DFLAG_NOSAVM attributes.

Submitted 50k jobs on a 1 vnode system with 1 ncpus, eligible_time_enable set to False. Ran a cycle to let the job alters for comment go through. Then set eligible_time_enable to True and ran another cycle, this time no job comment alters happen as they are the same as before, so this measures only the accrue_type alter requests.

**Before**: Sched cycle took 7 minutes with server spending 60%+ time in req_modifyjob():
Screen Shot 2020-02-18 at 4 13 39 PM

**After**: Sched cycle took around 2 minutes with the server now spending ~35% inside req_modifyjob() instead of 60%+:
Screen Shot 2020-02-19 at 8 57 22 PM

### DIFF
--- a/src/lib/Libattr/master_job_attr_def.xml
+++ b/src/lib/Libattr/master_job_attr_def.xml
@@ -1937,7 +1937,7 @@
 				NULL_FUNC
 			#endif</member_at_action>
       <member_at_flags>
-         <both>NO_USER_SET | ATR_DFLAG_SSET | ATR_DFLAG_ALTRUN</both>
+         <both>NO_USER_SET | ATR_DFLAG_SSET | ATR_DFLAG_ALTRUN | ATR_DFLAG_NOSAVM</both>
       </member_at_flags>
       <member_at_type>
          <both>ATR_TYPE_LONG</both>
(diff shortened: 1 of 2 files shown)


subhasisb:
> This check should be redundant (not required) since the set_svrjobstate function checks this as the first thing anyway...
> int
> svr_setjobstate(job *pjob, int newstate, int newsubstate)
> {
> pbs_queue *pque = pjob->ji_qhdr;
> pbs_sched *psched;
> 
> ```
> /*
>  * If the job has already finished, then do not make any new changes
>  * to job state or substate.
>  */
> if (pjob->ji_qs.ji_state == JOB_STATE_FINISHED || 
> 	((pjob->ji_qs.ji_state == newstate) && (pjob->ji_qs.ji_substate == newsubstate)))
> 	return (0);
> ```

Turns out i was looking at a different branch (where i made a similar change). So yes, I like the change, but could u put the check inside the function, so that it affects all callers? (as in the snippet above) - thanks!

agrawalravi90:
Sounds like this change will get antiquated soon with the multi-server stuff coming in, so I'll close this PR.

PR #1527, opened 2020-02-19
Add absolute path for qselect command in TestQselect test

shilpakodli:
#### Describe Bug or Feature
TestQselect.test_qselect_buffer_overflow test case was failing due to "qselect" command not found on Systems.

#### Describe Your Change
Add absolute path for qselect command in test case.

[TestQselect_after_fix.txt](https://github.com/PBSPro/pbspro/files/4224116/TestQselect_after_fix.txt)
[TestQselect_before_fix.txt](https://github.com/PBSPro/pbspro/files/4224117/TestQselect_before_fix.txt)

### DIFF
--- a/test/tests/functional/pbs_qselect.py
+++ b/test/tests/functional/pbs_qselect.py
@@ -47,7 +47,9 @@ class TestQselect(TestFunctional):
         Check that various qselect option arguments does not buffer overflow
         """
         # test -q option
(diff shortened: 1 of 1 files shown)


shilpakodli:
> Instead of using run_cmd to run qselect use server.select()

Hi @Mahalty , As per our offline discussion, I am keeping code as it is. Please review

PR #1529, opened 2020-02-19
Change default rpp_max_pkt_check value from 1024 to 64

scc138:
On busy installations with lots of jobs beginning and ending we have frequently had a need to lower the value of rpp_max_pkt_check as the default 1024 value can cause the client commands to appear sluggish.  Values of 32 or 128 are common, so 64 seems like a good default.

### DIFF
--- a/src/include/rpp.h
+++ b/src/include/rpp.h
@@ -57,7 +57,7 @@ extern "C" {
 /*
  * Default number of RPP packets to check every server iteration
  */
-#define RPP_MAX_PKT_CHECK_DEFAULT	1024
+#define RPP_MAX_PKT_CHECK_DEFAULT	64
 
 /* TPP specific definitions and structures */
 #define TPP_DEF_ROUTER_PORT 17001


agrawalravi90:
Hey @scc138 I remember discussions where we've talked about adding a sleep to qstat so that it doesn't bug the server too much. Is there a particular client command that we don't want to appear sluggish?

scc138:
Hi @agrawalravi90, that was taken care of in https://github.com/PBSPro/pbspro/pull/1279 (already done).  That is a separate issue since that is about too many client command requests and this is about the server being too busy processing job starting/ending/etc. requests that it does not check the client command queue often enough.  

This is a proposed adjustment to the work done and discussed at http://community.pbspro.org/t/pp-1250-process-3-rpp-tpp-messages-from-moms-per-server-iteration/929.  The loop iteration value had been hardcoded as 3 for years.  3 was found to be too small at busy sites where there was a large backlog of job events not getting processed in time and bloating server memory.  The value was made adjustable via qmgr and a default set to 1024 (which was a guessed at reasonable value at the time).  In the last ~2 years or so there have been numerous sites complaining of sluggish client command performance and the support team has helped them lower this value to (usually) between 32 and 128 until things "feel right", so this change of default is to simply make the default value a better fit for most sites.

mike0042:
The reference guide and installation guide both mention 1024 as the default value. Please file a draft doc request if you haven't already.

PR #1532, opened 2020-02-21
rpp connection 'server -> mom' can be confused

vchlum:
#### Describe Bug or Feature
We experienced vast network issues in our PBS infrastructure. There were a significant packet loss and connection breaks. As a result of the issues, some nodes were confused in the server like this:

It happened that node like:
```
create node node1 Mom=node1.domain
```
was randomly changed to:
```
create node node1 Mom="node1.domain,random-node.domain"
```

The issue seems to be caused by not deleting mom stream from the main tree of streams after rpp_close() in the server. 

tdelete2() should be called after  rpp_close() or  rpp_destroy() (supposing the stream was inserted in the tree of streams). If the tdelete2() is not called and the stream remains in the tree of streams, it can lead to confusing the mom identity.

#### Describe Your Change
 I found two occurrences of rpp_close() with missing tdelete2(). tdelete2() was added here.

### DIFF
--- a/src/server/node_manager.c
+++ b/src/server/node_manager.c
@@ -1441,6 +1441,7 @@ ping_a_mom_mcast(mominfo_t *pmom, int force_hello, int mtfd_ishello, int mtfd_is
 				 "Failed to add mom at %s:%d to ping mcast", pmom->mi_host, pmom->mi_port);
 		log_err(-1, __func__, log_buffer);
 		rpp_close(psvrmom->msr_stream);
+		tdelete2((u_long)psvrmom->msr_stream, 0, &streams);
 		psvrmom->msr_stream = -1;
 	}
 }
--- a/src/server/req_jobobit.c
+++ b/src/server/req_jobobit.c
@@ -692,6 +692,7 @@ conn_to_mom_failed(job *pjob, void(*func)(struct work_task *))
 		svr_disconnect(pjob->ji_momhandle);
 	} else {
 		rpp_close(pjob->ji_momhandle);
+		tdelete2((u_long)pjob->ji_momhandle, 0, &streams);
 	}
 	pjob->ji_momhandle = -1;
 	ptask = set_task(WORK_Immed, 0, func, pjob);


mike0042:
So in this case you want to delete the stream, but leave the node in the hostaddr_tree. Is that correct? If you want to remove them both, you should call remove_mom_ipaddresses_list().

If it is the case that tdelete2() should always be called after rpp_close() and rpp_delete(), then should we simply call it at the end of those functions?

vchlum:
@mike0042 The stream should not be present in the tree of streams after rpp_close() (in pbs server). I do not see a problem with remove_mom_ipaddresses_list().

We can not simply add tdelete2() at the end of rpp_close(). The rpp_close() can be called on a stream that is not in the tree. It actually happens in is_request() and on multiple places in mom.

Please, check it with me:
* The tdelete2() is missing in ping_a_mom_mcast() (if rc == -1) because tinsert2() is called in mom_ping_need().
* The tdelete2() is also missing in conn_to_mom_failed() because pjob->ji_momhandle is set to msr_stream (msr_stream is in the tree) via svr_connect().

subhasisb:
> @mike0042 The stream should not be present in the tree of streams after rpp_close() (in pbs server). I do not see a problem with remove_mom_ipaddresses_list().
> 
> We can not simply add tdelete2() at the end of rpp_close(). The rpp_close() can be called on a stream that is not in the tree. It actually happens in is_request() and on multiple places in mom.
> 
> Please, check it with me:
> 
> * The tdelete2() is missing in ping_a_mom_mcast() (if rc == -1) because tinsert2() is called in mom_ping_need().
> * The tdelete2() is also missing in conn_to_mom_failed() because pjob->ji_momhandle is set to msr_stream (msr_stream is in the tree) via svr_connect().

Yes i think your finding is correct @vchlum - we must remove from the  tree of streams. The ip_addresses_list is only about a name resolution cache and does not concern itself with streams.

subhasisb:
> > @mike0042 The stream should not be present in the tree of streams after rpp_close() (in pbs server). I do not see a problem with remove_mom_ipaddresses_list().
> > We can not simply add tdelete2() at the end of rpp_close(). The rpp_close() can be called on a stream that is not in the tree. It actually happens in is_request() and on multiple places in mom.
> > Please, check it with me:
> > 
> > * The tdelete2() is missing in ping_a_mom_mcast() (if rc == -1) because tinsert2() is called in mom_ping_need().
> > * The tdelete2() is also missing in conn_to_mom_failed() because pjob->ji_momhandle is set to msr_stream (msr_stream is in the tree) via svr_connect().
> 
> Yes i think your finding is correct @vchlum - we must remove from the tree of streams. The ip_addresses_list is only about a name resolution cache and does not concern itself with streams.

Just curious though, how did we know that your problems were being caused by the stale stream in the tree? Is it because, a tfind2(stream tree) is finding the stale connection later on?

vchlum:
> Just curious though, how did we know that your problems were being caused by the stale stream in the tree? Is it because, a tfind2(stream tree) is finding the stale connection later on?

It is because of tinsert2(stream, 0, ...). If we find both key1 and key2 in the tree with tinsert2() then tinsert2() just return (since it is void) and no data are stored in the tree. We suppose the mom is already inserted in the tree but the assumption can be incorrect - if the stream was not deleted before it is now in the tree with incorrect momp.

subhasisb:
> > Just curious though, how did we know that your problems were being caused by the stale stream in the tree? Is it because, a tfind2(stream tree) is finding the stale connection later on?
> 
> It is because of tinsert2(stream, 0, ...). If we find both key1 and key2 in the tree with tinsert2() then tinsert2() just return (since it is void) and no data are stored in the tree. We suppose the mom is already inserted in the tree but the assumption can be incorrect - if the stream was not deleted before it is now in the tree with incorrect momp.
> 
> The confused connection started to log:
> Node;nympha34.<domain>;Not allowed to update vnode 'nympha32', as it is owned by a different mom

Ah ok - that explains - thanks @vchlum

PR #1535, opened 2020-02-24
pbs_benchpress --use-current-setup needs manager and operator privileges

sujatapatnaik52:
#### Describe Bug or Feature
`pbs_benchpress` has an option `--use-current-setup` which runs a testsuite with the current state of PBS. After doing a fresh installation of PBS, when `pbs_benchpress` is run with --use-current-setup, an error **Unauthorized request** is observed while running a `qmgr` command in the test case. 

#### Describe Your Change
Since the test execution user needs manager privileges for creating resources in PBS Server, Added the test execution user as manager role at the testsuite level. Giving `MGR_USER` and `MGR_OPER` as manager and operator roles respectively similar to the default roles assigned without `--use-current-setup`

[after_fix.txt](https://github.com/PBSPro/pbspro/files/4243013/after_fix.txt)
[before_fix.txt](https://github.com/PBSPro/pbspro/files/4243014/before_fix.txt)

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -5295,6 +5295,11 @@ class Server(PBSService):
             return False
         else:
             conf['qmgr_print_sched'] = ret['out']
+
+        # sudo=True is added while running "pbsnodes -av", to make
+        # sure that all the node attributes are preserved in
+        # save_configuration. If this command is run without sudo,
+        # some of the node attributes like port, version is not listed.
         ret = self.du.run_cmd(self.hostname, [pbsnodes, '-av'],
                               logerr=False, level=logging.DEBUG, sudo=True)
         err_msg = "Server has no node list"
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -5205,7 +5205,7 @@ class Server(PBSService):
                 pass
             self.manager(MGR_CMD_DELETE, QUEUE, id=queues)
 
-    def delete_sched_config(self, sudo=False):
+    def delete_sched_config(self):
         """
         Delete sched_priv & sched_log files
         """
@@ -5221,7 +5221,7 @@ class Server(PBSService):
                            recursive=True, force=True)
                 self.du.rm(path=sched_priv, sudo=True,
                            recursive=True, force=True)
-                self.manager(MGR_CMD_DELETE, SCHED, id=name, sudo=sudo)
+                self.manager(MGR_CMD_DELETE, SCHED, id=name)
 
     def delete_nodes(self):
         """
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -5151,7 +5151,7 @@ class Server(PBSService):
                 self.manager(MGR_CMD_DELETE, RSC, id=rescs)
         return True
 
-    def unset_svr_attrib(self, server_stat=None, sudo=False):
+    def unset_svr_attrib(self, server_stat=None):
         """
         Unset server attributes
         """
@@ -5173,9 +5173,9 @@ class Server(PBSService):
             else:
                 unsetlist.append(k)
         if len(unsetlist) != 0:
-            self.manager(MGR_CMD_UNSET, MGR_OBJ_SERVER, unsetlist, sudo=sudo)
+            self.manager(MGR_CMD_UNSET, MGR_OBJ_SERVER, unsetlist)
 
-    def delete_site_hooks(self, sudo=False):
+    def delete_site_hooks(self):
         """
         Delete site hooks from PBS
         """
@@ -5186,9 +5186,9 @@ class Server(PBSService):
             if h in hooks:
                 hooks.remove(h)
         if len(hooks) > 0:
-            self.manager(MGR_CMD_DELETE, HOOK, id=hooks, sudo=sudo)
+            self.manager(MGR_CMD_DELETE, HOOK, id=hooks)
 
-    def delete_queues(self, sudo=False):
+    def delete_queues(self):
         """
         Delete queues
         """
@@ -5203,7 +5203,7 @@ class Server(PBSService):
                                      node['id'])
             except:
                 pass
-            self.manager(MGR_CMD_DELETE, QUEUE, id=queues, sudo=sudo)
+            self.manager(MGR_CMD_DELETE, QUEUE, id=queues)
 
     def delete_sched_config(self, sudo=False):
         """
@@ -5223,12 +5223,12 @@ class Server(PBSService):
                            recursive=True, force=True)
                 self.manager(MGR_CMD_DELETE, SCHED, id=name, sudo=sudo)
 
-    def delete_nodes(self, sudo=False):
+    def delete_nodes(self):
         """
         Remove all the nodes from PBS
         """
         try:
-            self.manager(MGR_CMD_DELETE, VNODE, id="@default", sudo=sudo)
+            self.manager(MGR_CMD_DELETE, VNODE, id="@default")
         except PbsManagerError as e:
             if "Unknown node" not in e.msg[0]:
                 raise
@@ -13214,7 +13214,7 @@ class MoM(PBSService):
 
         return self.version
 
-    def delete_vnodes(self, sudo=False):
+    def delete_vnodes(self):
         rah = ATTR_rescavail + '.host'
         rav = ATTR_rescavail + '.vnode'
         a = {rah: self.hostname, rav: None}
@@ -13235,7 +13235,7 @@ class MoM(PBSService):
             if v[rav].split('.')[0] != v[rah].split('.')[0]:
                 vs.append(v['id'])
         if len(vs) > 0:
-            self.server.manager(MGR_CMD_DELETE, VNODE, id=vs, sudo=sudo)
+            self.server.manager(MGR_CMD_DELETE, VNODE, id=vs)
 
     def revert_to_defaults(self, delvnodedefs=True):
         """
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -5151,7 +5151,7 @@ class Server(PBSService):
                 self.manager(MGR_CMD_DELETE, RSC, id=rescs)
         return True
 
-    def unset_svr_attrib(self, server_stat=None):
+    def unset_svr_attrib(self, server_stat=None, sudo=False):
         """
         Unset server attributes
         """
@@ -5173,9 +5173,9 @@ class Server(PBSService):
             else:
                 unsetlist.append(k)
         if len(unsetlist) != 0:
-            self.manager(MGR_CMD_UNSET, MGR_OBJ_SERVER, unsetlist)
+            self.manager(MGR_CMD_UNSET, MGR_OBJ_SERVER, unsetlist, sudo=sudo)
 
-    def delete_site_hooks(self):
+    def delete_site_hooks(self, sudo=False):
         """
         Delete site hooks from PBS
         """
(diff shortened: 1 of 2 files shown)


test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -440,6 +440,8 @@ class PBSTestSuite(unittest.TestCase):
     scheds = {}
     moms = None
     comms = None
+    users = ""
+    added_users = []
```
hirenvadalia: I don't think you need 2 separate variables to store a username and its role. It can be done by one variable also.

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -450,6 +452,19 @@ def setUpClass(cls):
...
+                         str(MGR_USER) + '@*': "managers",
+                         str(OPER_USER) + '@*': "operators"}
+            server_stat = cls.server.status(SERVER, ["managers", "operators"])
+            if len(server_stat) > 0:
+                server_stat = server_stat[0]
+            for user, role in cls.users.items():
+                if role not in server_stat or user not in server_stat[role]:
+                    attr = {role: (INCR, user)}
+                    cls.server.manager(MGR_CMD_SET, SERVER, attr, sudo=True)
+                    cls.added_users.append(user)
```
hirenvadalia: rather than adding just user in `added_users`, you can add user + role, so you don't need 2nd class variable.

test/fw/ptl/utils/pbs_testsuite.py
hirenvadalia: Also, I suggest renaming `added_users` to something like `added_mgrs_oprs` for more readability

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -450,6 +452,19 @@ def setUpClass(cls):
...
+            cls.users = {test_user: "managers",
+                         str(MGR_USER) + '@*': "managers",
+                         str(OPER_USER) + '@*': "operators"}
+            server_stat = cls.server.status(SERVER, ["managers", "operators"])
+            if len(server_stat) > 0:
+                server_stat = server_stat[0]
+            for user, role in cls.users.items():
+                if role not in server_stat or user not in server_stat[role]:
+                    attr = {role: (INCR, user)}
+                    cls.server.manager(MGR_CMD_SET, SERVER, attr, sudo=True)
```
hirenvadalia: Also, these added mgrs and oprs will be saved in setup (which is not part of the original setup which user asked to save and use) so shouldn't we need to take care of removing added mgrs and oprs while restoring saved setup?

sujatapatnaik52: I agree with you. If I remove the added mgrs and oprs in teardown(while restoring the saved setup), I may have to add it again in the next testcase's setup. so even though I have added it in setupclass I will have to add it in setup as well. Does that sound fine?

hirenvadalia: On second thought (as per our offline skype chat): I think we should move this setting mgrs and oprs into `save_configuration` and reverting it to `load_configuration`, that way if someone else other than `setUp/tearDown(class)` is using save/load configuration then they will be also beneficial for this change.

sujatapatnaik52: Thank you @hirenvadalia for the suggestion. I am making the change.

sujatapatnaik52: @hirenvadalia As per your suggestion, I am using only one variable for the fix. Please look at it.

Also, I tried  adding the managers and operators in `save_configuration` and `load configuration`. It works upto the point where `load_configuration` is called in teardown. I am removing the users which were added in setup. But the catch is `load_configuration` is also called in `teardownclass`, by the time, it reaches there, the users are already removed in `teardown` and because of which there is **Unauthorized request** in `teardownclass`. 

As discussed offline, I feel `save_configuration` and `load_configuration` are API's as per their names i.e. related to configuration changes. Adding users is kind of a pre-requisite which we are doing based on our requirement.

hirenvadalia: " Adding users is kind of a pre-requisite."
Well, I don't agree on this as this is not pre-requisite, those mgrs and oprs are needed by save/load configuration only... so we should set and reset in those methods only rather than asking the user to do before using those methods. I understand there will be overhead to set and reset mgrs and oprs but we will get save/load config method fully independent of anything else, and a user can use without any headache.

sujatapatnaik52: I may have to add managers and operators in teardownclass before load_configuration and remove them again after the configuration is loaded. This seems to me as a bit of overhead. 

Please correct me if I am wrong. With my current changes, the test execution user is added as a manager role in the `setupclass` of `PBSTestSuite` if used with `--use-current-setup`. If the user is using `save/load configuration` in the test case, then the user shall still be permitted to use the API's.

hirenvadalia: "Please correct me if I am wrong. With my current changes, the test execution user is added as a manager role in the setupclass of PBSTestSuite if used with --use-current-setup. If the user is using save/load configuration in the test case, then the user shall still be permitted to use the API's."

Yes anybody (within PTL fw + tests) can use save and load conf method. and that's why I am suggesting that we should make those methods independent of anything.

I understand there will be overhead, but everything comes with a price :smile: 

Or try to modify save and load conf methods such a way that it doesn't need mgrs and opers set? like use sudo or root user when doing privilege operation inside save/load conf methods

sujatapatnaik52: okay how about this - Let's make load_configuration and save_configuration work with sudo privileges. The managers and operators will still be needed for the test execution to work properly so that can be added independent of these API's maybe after the configuration is saved. Sounds good?

sujatapatnaik52:
@hirenvadalia The `save_configuration` and `load_configuration` has most of the commands which run with sudo privileges, It is just the other API's like `delete_current_state` which needed the same. I added a optional sudo parameter to those API's. I have added the managers and operators in setup and removing them in teardown. Please look at the latest commit.

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -483,7 +484,28 @@ def setUp(self):
         # once save & load configurations are implemented for
         # comm, mom, scheduler
         if self.use_cur_setup:
-            self.server.delete_nodes()
+            self.server.delete_nodes(True)
+            attr = {}
+            test_user = pwd.getpwuid(os.getuid())[0] + '@*'
```
hirenvadalia: Now when you are using sudo everywhere why do we need to set mgrs and opers?

sujatapatnaik52: The `revert_moms` is failing. Also, In the test case, the user may create vnodes, queues or do some other manager oriented operations. So I feel adding them will be the right thing to do.

hirenvadalia: "Also, In the test case, the user may create vnodes, queues or do some other manager oriented operations"
This test case has to take care of it, not setUp.

"The revert_moms is failing."
I suggest using sudo there also... We should keep one way of doing stuff, either sudo or with mgrs/opers... not both of it.

sujatapatnaik52:
Thank you @hirenvadalia for the thorough review and suggestions. Requesting one more review from @pbspro-maintainers.

sujatapatnaik52:
[smoketest_run_logs_with_use_current_setup](https://github.com/PBSPro/pbspro/files/4244355/smoketest_logs_use_current_setup.txt)

hirenvadalia:
> should PTL really be doing this? --use-current-setup kind of relies on the test runner to set things up such that the test runs properly right? Also, with --use-current-setup, PTL shouldn't be creating any new resources/nodes/queues etc., so why does it need operator and manager privileges ?

I also agree @agrawalravi90 with you here but the problem is, in PTL (from starting) we set the current user as mgr/oper, and now we have many tests written keeping that in mind so either we have to change all those tests not to assume anything or adopt the same env in `--use-current-setup`.
And easiest is to adopt env in `--user-current-setup` rather than changing hundreds of test cases.

Also, we can ask the user to make sure it adds the current user as mgr/opers but that will be a headache for the user to keep this mind and do it every time before running pbs_benchpress.
So from a usability point of view, we thought changing fw will be good.

HTH!

agrawalravi90:
Thanks for explaining Hiren. Ok, I agree that this sounds like the most sensible thing to do right now. But we should document this. Can we modify the doc:
https://pbspro.atlassian.net/wiki/spaces/PD/pages/557088789/Design+for+a+supported+way+to+change+default+setup+in+PTL

to mention this as a caveat ?

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -5296,7 +5296,7 @@ def save_configuration(self, outfile, mode='a'):
         else:
             conf['qmgr_print_sched'] = ret['out']
         ret = self.du.run_cmd(self.hostname, [pbsnodes, '-av'],
-                              logerr=False, level=logging.DEBUG)
+                              logerr=False, level=logging.DEBUG, sudo=True)
```
agrawalravi90: why does this need a sudo ? it's just pbsnodes -av command right?

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -458,6 +458,7 @@ def setUpClass(cls):
             else:
                 cls.logger.error("Failed to save custom setup")
                 raise Exception("Failed to save custom setup")
+            cls().add_mgrs_opers(cls.server)
```
agrawalravi90: this looks kind of hack-y .. you are creating an instance of the class via cls() to call a non-classmethod from a classmethod. why not just make add_mgrs_opers() a classmethod? Then you can just call it as "cls.add_mgrs_opers()" here. Inside setUp() for the normal code path, you can just call it as self.add_mgrs_opers(). You'll have to change "server" to "cls.server" inside the method

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -1308,6 +1309,38 @@ def revert_moms(self, force=False):
...
+                attr[role] = (INCR, ','.join(users))
+            else:
+                add_users = []
+                for user in users:
+                    if user not in server_stat[role]:
+                        add_users.append(user)
+                if len(add_users) > 0:
+                    attr[role] = (INCR, ",".join(add_users))
+        if len(attr) > 0:
+            self.server.manager(MGR_CMD_SET, SERVER, attr, sudo=True)
```
agrawalravi90: you are passing in 'server' object so you don't need to call it as "self.server". If you take my suggestion for converting this into a class method, then you'll have to use 'cls.server' everywhere in the method

test/fw/ptl/utils/pbs_testsuite.py
sujatapatnaik52: I tried this yesterday and I was getting some error so I did not debug further. Maybe I was surely doing something wrong. I have made the changes as per your suggestion. Please review the latest commit.

test/fw/ptl/lib/pbs_testlib.py
sujatapatnaik52: `pbsnodes -av` without `sudo` i.e. if this command is run as a test execution user, it does not list all the nodes attributes like Port, version etc. We just proceed with saving the configuration without `Port`. While loading configuration, we create the node using `Port` and that gives a `KeyError`. So, I had to add this.

agrawalravi90: thanks for explaining, can you please add this as a comment in the code so that someone like me doesn't remove it thinking it's unnecessary ? :)

sujatapatnaik52: Added the comment. Also, I have updated the design document. Please review both.

sujatapatnaik52:
Thank you @agrawalravi90 and @hirenvadalia for the sign off.

test/fw/ptl/lib/pbs_testlib.py
bhroam: As a note for the future, you didn't need to use sudo.  There is no difference between running as root and another manager.  You could have done a runas=MGR_USR and it would have been a bit more secure.

PR #1537, opened 2020-02-25
A new fix for server hooks nonetype object not callable error

minghui-liu:
#### Describe Bug or Feature

This is a new fix for the problem describe in #1445 .

#### Describe Your Change

* Reverted #1445 
* Explicitly import Python's built-int `signal` module and register `signal.default_int_handler`, instead of temporarily set SIGINT handler to SIGDFL and then switch it back. 
* The change above also requires putting a block of code that only executes when the server starts a Python interpreter. So I made pbs_server link to `libpbspython_svr.a` only and compile `libpbspython_svr.a` with a special flag. All other daemons and commands (like pbs_python) still link to `libpbspython.a`. 

#### Link to Design Doc

N/A

Screen Shot 2020-03-02 at 5 56 39 PM

[717-ALL.zip](https://github.com/PBSPro/pbspro/files/4274710/717-ALL.zip)

Note: TestPbsExecutePrologue.test_prologue_hook_set_fail_action fails on oss master branch as well so is not caused by this PR.

### DIFF
--- a/src/lib/Libpython/pbs_python_external.c
+++ b/src/lib/Libpython/pbs_python_external.c
@@ -115,12 +115,6 @@ pbs_python_ext_start_interpreter(struct python_interpreter_data *interp_data)
 	char pbs_python_destlib2[MAXPATHLEN + 1] = {'\0'};
 	int  evtype;
 	int  rc;
-#ifndef WIN32
-	struct sigaction act;
-	struct sigaction oact;
-#else
-	void *oact;
-#endif
 
 	/*
 	 * initialize the convenience global pbs_python_daemon_name, as it is
@@ -188,46 +182,12 @@ pbs_python_ext_start_interpreter(struct python_interpreter_data *interp_data)
 		goto ERROR_EXIT;
 	}
 
-#ifndef WIN32
-	/*
-	 * Temporary set SIGINT to SIG_DFL, so Py_InitializeEx can setup proper SIGINT handler
-	 * see https://github.com/python/cpython/blob/3.6/Modules/signalmodule.c#L1280
-	 * as per above, If SIGINT is not set to SIG_DFL then Python won't register it's SIGINT handler
-	 * which means Python won't raise KeyBoardInterrupt on PyErr_SetInterrupt()
-	 * instead it will throw NoneType object is not callable exception
-	 */
-	sigemptyset(&act.sa_mask);
-	act.sa_flags   = 0;
-	act.sa_handler = SIG_DFL;
-	if (sigaction(SIGINT, &act, &oact) != 0) {
-		log_err(errno, __func__, "Failed to set SIG_DFL on SIGINT");
-		return 1;
-	}
-#else
-	oact = signal(SIGINT, SIG_DFL);
-	if (oact == SIG_ERR) {
-		log_err(errno, __func__, "Failed to set SIG_DFL on SIGINT");
-		return 1;
-	}
-#endif
 	/*
 	 * arg '1' means to not skip init of signals
 	 * we want signals to propagate to the executing
 	 * Python script to be able to interrupt it
 	 */
 	Py_InitializeEx(1);
-
-	/* revert SIGINT to original sig handler */
-#ifndef WIN32
-	if (sigaction(SIGINT, &oact, NULL) != 0)
-#else
-	if (signal(SIGINT, oact) == SIG_ERR)
-#endif
-	{
-		log_err(errno, __func__, "Failed to revert signal handler for SIGINT");
-		return 1;
-	}
-
 
 	if (Py_IsInitialized()) {
 		char *msgbuf;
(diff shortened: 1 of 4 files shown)


src/tools/Makefile.am
```diff
@@ -126,6 +124,15 @@ pbs_python_LDADD = \
...
 pbs_python_SOURCES = \
+	$(top_builddir)/src/lib/Libpython/shared_python_utils.c \
+	$(top_builddir)/src/lib/Libpython/common_python_utils.c \
+	$(top_builddir)/src/lib/Libpython/pbs_ifl_wrap.c \
+	$(top_builddir)/src/lib/Libpython/pbs_python_external.c \
+	$(top_builddir)/src/lib/Libpython/pbs_python_svr_external.c \
+	$(top_builddir)/src/lib/Libpython/module_pbs_v1.c \
+	$(top_builddir)/src/lib/Libpython/pbs_python_svr_internal.c \
+	$(top_builddir)/src/lib/Libpython/pbs_python_svr_size_type.c \
+	$(top_builddir)/src/lib/Libpython/pbs_python_import_types.c \
```
mike0042: Why are you explicitly listing the source files rather than linking against the archive libraries? We use the same technique in src/server/Makefile.am

src/lib/Libpython/pbs_python_external.c
```diff
@@ -278,6 +238,28 @@ pbs_python_ext_start_interpreter(struct python_interpreter_data *interp_data)
...
+        goto ERROR_EXIT;
+    }
+    Py_XDECREF(m);
+    Py_XDECREF(d);
+    Py_XDECREF(f);
+    Py_XDECREF(sigint);
+    Py_XDECREF(handler);
+    log_event(PBSEVENT_DEBUG, PBS_EVENTCLASS_SERVER, LOG_INFO, interp_data->daemon_name, "successfully set up signal.default_int_handler");
+#endif
+
```
mike0042: Could this also be the reason that SIGALRM isn't getting through? The Timeout class in the cgroups hook seems to be broken since we switched to Python 3. @alexis-cousein will also be interested.

src/tools/Makefile.am
minghui-liu: I made `pbs_python` to compile with the libpython source files instead of linking to the archive library, so that the I can use the `PBS_PYTHON` flag to determine if my code in `pbs_python_external.c` is compiled with the server or with the `pbs_python` cmd. The code only need to execute when the interpreter is started by the server.

mike0042: Wasn't the point of having libpbspython.a and libpbspython_svr.a to compile them with different flags? Perhaps we shouldn't be referencing them both in pbs_python_LDADD.

src/lib/Libpython/pbs_python_external.c
minghui-liu: Yes this could be the reason. Python 3 no longer sets up `default_int_handler` when it starts like python 2 did.

src/tools/Makefile.am
minghui-liu: `libpbspython.a` and `libpbspython_svr.a` were compiled with different files. But what you pointed out is a good idea and I have modify the Makefiles so that `libpbspython_svr.a` is used exclusively for the server code.

src/lib/Libpython/pbs_python_external.c
mike0042: Should we be doing this for additional signals, besides SIGINT? Or is there a way to cope with this in the hook itself?

minghui-liu: We are doing this for SIGINT because we rely on `default_int_handler`'s behavior (raising a KeyboardInterrupt exception) for hook alarms. 
For all the other signals that each hook want to handler, the hook itself should import the `signal` module and register and custom handler function using `signal.signal()`.

src/lib/Libpython/pbs_python_external.c
```diff
@@ -278,6 +238,24 @@ pbs_python_ext_start_interpreter(struct python_interpreter_data *interp_data)
...
+    m = PyImport_ImportModule("signal");
+    d = PyModule_GetDict(m);
+    f = PyDict_GetItemString(d, "signal");
+    handler = PyDict_GetItemString(d, "default_int_handler");
+    sigint = PyDict_GetItemString(d, "SIGINT");
+    if (f && PyCallable_Check(f)) {
+        PyObject_CallFunctionObjArgs(f, sigint, handler, NULL);
+    } else {
+        log_err(-1, __func__, "could not set up python signal.default_int_handler");
+        goto ERROR_EXIT;
```
hirenvadalia: You need free `m` here before jumping to `ERROR_EXIT`.

src/lib/Libpython/pbs_python_external.c
```diff
@@ -278,6 +238,24 @@ pbs_python_ext_start_interpreter(struct python_interpreter_data *interp_data)
 	}
 
 	interp_data->pbs_python_types_loaded = 1; /* just in case */
+
+#ifdef LIBPYTHONSVR
+	PyObject *m, *d, *f, *handler, *sigint;
```
hirenvadalia: please fix the indentation.

src/lib/Libpython/pbs_python_external.c
```diff
@@ -278,6 +238,24 @@ pbs_python_ext_start_interpreter(struct python_interpreter_data *interp_data)
...
+    m = PyImport_ImportModule("signal");
```
hirenvadalia: Shouldn't we need to do an error check?

src/lib/Libpython/pbs_python_external.c
```diff
@@ -278,6 +238,24 @@ pbs_python_ext_start_interpreter(struct python_interpreter_data *interp_data)
...
+    f = PyDict_GetItemString(d, "signal");
+    handler = PyDict_GetItemString(d, "default_int_handler");
+    sigint = PyDict_GetItemString(d, "SIGINT");
+    if (f && PyCallable_Check(f)) {
+        PyObject_CallFunctionObjArgs(f, sigint, handler, NULL);
+    } else {
+        log_err(-1, __func__, "could not set up python signal.default_int_handler");
+        goto ERROR_EXIT;
+    }
+    Py_XDECREF(m);
```
hirenvadalia: I suggest using `Py_CLEAR()` as per py doc.

src/lib/Libpython/pbs_python_external.c
```diff
@@ -278,6 +238,24 @@ pbs_python_ext_start_interpreter(struct python_interpreter_data *interp_data)
...
+
+#ifdef LIBPYTHONSVR
+	PyObject *m, *d, *f, *handler, *sigint;
+    m = PyImport_ImportModule("signal");
+    d = PyModule_GetDict(m);
+    f = PyDict_GetItemString(d, "signal");
+    handler = PyDict_GetItemString(d, "default_int_handler");
+    sigint = PyDict_GetItemString(d, "SIGINT");
+    if (f && PyCallable_Check(f)) {
+        PyObject_CallFunctionObjArgs(f, sigint, handler, NULL);
```
hirenvadalia: error check?

src/lib/Libpython/pbs_python_external.c
minghui-liu: Done.

PR #1539, opened 2020-02-26
Can't alter start time of running resv if jobs finished with history

vstumpf:
#### Describe Bug or Feature
If a reservation is empty, you should be able to alter its start time of a running reservation.  If there are history jobs in the reservation, the server thinks it is not empty and rejects the alter request.
 

```
[vstumpf@shecil pbspro]$ qstat -x
Job id Name User Time Use S Queue
---------------- ---------------- ---------------- -------- - -----
1347.shecil STDIN vstumpf 00:00:00 F R1346

[vstumpf@shecil pbspro]$ pbs_ralter -R 1530 R1346
pbs_ralter: Reservation not empty
```

#### Describe Your Change
If job history is enabled, subtract moved, finished, and expired jobs from the number of jobs in the queue.

[afte-fix-manual.txt](https://github.com/PBSPro/pbspro/files/4252796/afte-fix-manual.txt)
[after-fix.txt](https://github.com/PBSPro/pbspro/files/4252797/after-fix.txt)

### DIFF
--- a/src/server/req_modify.c
+++ b/src/server/req_modify.c
@@ -818,8 +818,7 @@ req_modifyReservation(struct batch_request *preq)
 
 		switch (index) {
 			case RESV_ATR_start:
-				if ((presv->ri_wattr[RESV_ATR_state].at_val.at_long != RESV_RUNNING) ||
-					!(num_jobs)) {
+				if ((presv->ri_wattr[RESV_ATR_state].at_val.at_long != RESV_RUNNING) || !num_jobs) {
 					temp = strtol(psatl->al_value, &end, 10);
 					if ((temp > time(NULL)) &&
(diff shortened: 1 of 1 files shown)


src/server/req_modify.c
```diff
@@ -812,7 +819,7 @@ req_modifyReservation(struct batch_request *preq)
 		switch (index) {
 			case RESV_ATR_start:
 				if ((presv->ri_wattr[RESV_ATR_state].at_val.at_long != RESV_RUNNING) ||
-					!(presv->ri_qp->qu_numjobs)) {
+					!(num_jobs)) {
```
bhroam: Could you remove the () around num_jobs.  I don't think it makes anything clearer.

PR #1544, opened 2020-02-26
Fixed case where comment could not be set and refactored schd_error

bhroam:
#### Describe Bug or Feature
There is a case where the scheduler can set a job comment to 'Not Running:' without a reason.  It happens in free placement when a host-set is too small.

#### Describe Your Change
I fixed the issue and also changed the message (since it never was printed before).  The node search code should return the reason why the job didn't run on one node, but shouldn't specify the node.  I changed the to be the same regardless if we're checking a vnode or a host.

In the process of this, I needed to free/NULL one of the args of the schd_error structure.  There is a setter function for the args, but no way to unset.  I modified the setter to take NULL.

### DIFF
--- a/src/scheduler/buckets.c
+++ b/src/scheduler/buckets.c
@@ -1012,7 +1012,7 @@ find_correct_buckets(status *policy, node_bucket **buckets, resource_resv *resre
 					total += buckets[j]->total * c;
 				} else {
 					if (failerr->status_code == SCHD_UNKWN)
-						copy_schd_error(failerr, err);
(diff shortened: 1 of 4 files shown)


bhroam:
@vstumpf I'd rather leave the code alone.  I believe if you pass a NULL to strdup() you'll crash.  To avoid confusion, I'd want to put a comment saying it wouldn't crash on a NULL.  I'd rather just leave the code alone which is self explanatory that you can't pass a NULL.

PR #1547, opened 2020-03-03
snapshot ptl test: adding custom resource check

agrawalravi90:
#### Describe Bug or Feature

The test was missing a check for custom resource with pbs_snapshot is run with --obfuscate, so just added that.

[test_snapshot_obf_stress.log](https://github.com/PBSPro/pbspro/files/4278902/test_snapshot_obf_stress.log)


agrawalravi90:
@hirenvadalia  thanks, @PBSPro/pbspro-maintainers can you please take this forward, thanks.

PR #1549, opened 2020-03-04
PTL test: Fixing test_obfuscate_resv_user_groups

agrawalravi90:
#### Describe Bug or Feature

The test was failing on single core machines because it was expecting  a reservation requesting 2 ncpus to be confirmed without checking that there were 2 ncpus available.

#### Describe Your Change

Changed the ncpus requested from 2 to 1, there's no reason to request 2 ncpus in this test.

[after.log](https://github.com/PBSPro/pbspro/files/4288687/after.log)
[before.log](https://github.com/PBSPro/pbspro/files/4288688/before.log)


agrawalravi90:
Thanks @bhroam ! requesting one more review, it's a tiny fix, thanks!

PR #1550, opened 2020-03-04
Added a mock run option to pbs_mom for testing

agrawalravi90:
#### Describe Bug or Feature

Often while testing, especially for performance profiling, devs have the need to run thousands of jobs to tax scheduler and server enough to find bottlenecks. In such scenarios, pbs_mom becomes a bottleneck instead when the testing is being done on a single machine, not an actual cluster. Major reasons why this happens:

- pbs_mom forks off a separate process for each job that it runs. If we create a system with 10k vnodes, which are able to run 100k jobs at once, imagine pbs_mom(s) forking off 100k processes for these jobs at the same time. This overloads the system.

- pbs_mom also walks through the entire proc table on the system on a regular basis. This again compounds the problem in bullet 1 as there are hundreds of thousands of processes.

- pbs_mom also sets a high nice value for its process, which means that the kernel gives it preference over other processes like the scheduler or server which are running on the same machine.

So, for testing/performance profiling scheduler or server, it would be immensely useful to add an option to pbs_mom which makes it lightweight and not burden the system, while still maintaining similar server communication so that server and sched don’t see any difference and work the same way that they would in a real cluster.

#### Describe Your Change

Added a new option to pbs_mom "-m" which will make it execute in mock run mode. For more details, please refer to the design document. This PR doesn't implement the change that's needed to update cput and cpupercent to something more accurate than 0, this can be taken up in a later enhancement.

#### Link to Design Doc

https://pbspro.atlassian.net/wiki/spaces/PD/pages/1555726337/Add+a+mock+run+option+to+pbs+mom+for+testing

[test_mock_rsc_used.log](https://github.com/PBSPro/pbspro/files/4338664/test_mock_rsc_used.log)

### DIFF
--- a/src/include/mom_func.h
+++ b/src/include/mom_func.h
@@ -143,6 +143,8 @@ enum hup_action {
  */
 extern enum hup_action	call_hup;
 
+extern int mock_run;
+
 /* public funtions within MOM */
 
 #ifdef	_PBS_JOB_H
@@ -397,6 +399,8 @@ extern int setcurrentworkdir(char *);
 extern int becomeuser(job *);
 extern int becomeuser_args(char *, uid_t, gid_t, gid_t);
 extern void close_update_pipes(job *);
+extern void mom_set_use_all(void);
+void job_purge_mom(job *pjob);
 
 /* From popen.c */
 extern FILE *pbs_popen(const char *, const char *);
--- a/src/resmom/linux/mom_start.c
+++ b/src/resmom/linux/mom_start.c
@@ -2436,13 +2436,7 @@ scan_for_terminated(void)
 
 	termin_child = 0;
 
-	if (mom_get_sample() == PBSE_NONE) {
-		pjob = (job *)GET_NEXT(svr_alljobs);
-		while (pjob) {
-			mom_set_use(pjob);
-			pjob = (job *)GET_NEXT(pjob->ji_alljobs);
-		}
-	}
+	mom_set_use_all();
 
 	/* Now figure out which task(s) have terminated (are zombies) */
 
--- a/src/resmom/mock_run.c
+++ b/src/resmom/mock_run.c
@@ -1,0 +1,283 @@
+/*
+ * Copyright (C) 1994-2020 Altair Engineering, Inc.
+ * For more information, contact Altair at www.altair.com.
+ *
+ * This file is part of the PBS Professional ("PBS Pro") software.
+ *
+ * Open Source License Information:
+ *
+ * PBS Pro is free software. You can redistribute it and/or modify it under the
+ * terms of the GNU Affero General Public License as published by the Free
+ * Software Foundation, either version 3 of the License, or (at your option) any
+ * later version.
+ *
+ * PBS Pro is distributed in the hope that it will be useful, but WITHOUT ANY
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.
+ * See the GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Commercial License Information:
+ *
+ * For a copy of the commercial license terms and conditions,
+ * go to: (http://www.pbspro.com/UserArea/agreement.html)
+ * or contact the Altair Legal Department.
+ *
+ * Altair’s dual-license business model allows companies, individuals, and
+ * organizations to create proprietary derivative works of PBS Pro and
+ * distribute them - whether embedded or bundled with other software -
+ * under a commercial license agreement.
+ *
+ * Use of Altair’s trademarks, including but not limited to "PBS™",
+ * "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
+ * trademark licensing policies.
+ *
+ */
+
+#include <pbs_config.h>
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <assert.h>
+#include <time.h>
+#include <errno.h>
+
+#include "attribute.h"
+#include "batch_request.h"
+#include "job.h"
+#include "log.h"
+#include "mock_run.h"
+#include "mom_func.h"
+#include "pbs_error.h"
+#include "resource.h"
+
+#if defined(PBS_SECURITY) && (PBS_SECURITY == KRB5)
+#include "renew_creds.h"
+#endif
+
+extern time_t time_now;
+extern time_t time_resc_updated;
+extern int min_check_poll;
+extern int next_sample_time;
+
+void
+mock_run_finish_exec(job *pjob)
+{
+	resource_def *rd;
+	attribute *wallt;
+	resource *wall_req;
+	int walltime = 0;
+
+	rd = find_resc_def(svr_resc_def, "walltime", svr_resc_size);
+	wallt = &pjob->ji_wattr[(int)JOB_ATR_resource];
+	wall_req = find_resc_entry(wallt, rd);
+	if (wall_req != NULL) {
+		walltime = wall_req->rs_value.at_val.at_long;
+		start_walltime(pjob);
+	}
+
+	time_now = time(NULL);
+
+	/* Add a work task that runs when the job is supposed to end */
+	set_task(WORK_Timed, time_now + walltime, mock_run_end_job_task, pjob);
+
+	sprintf(log_buffer, "Started mock run of job");
+	log_event(PBSEVENT_JOB, PBS_EVENTCLASS_JOB,
+		LOG_INFO, pjob->ji_qs.ji_jobid, log_buffer);
+
+	mock_run_record_finish_exec(pjob);
+
+	return;
+}
+
+void
+mock_run_record_finish_exec(job *pjob)
+{
+	pjob->ji_qs.ji_state = JOB_STATE_RUNNING;
+	pjob->ji_qs.ji_substate = JOB_SUBSTATE_RUNNING;
+	job_save(pjob, SAVEJOB_QUICK);
+
+	pjob->ji_wattr[(int)JOB_ATR_state].at_val.at_long = JOB_STATE_RUNNING;
+	pjob->ji_wattr[(int)JOB_ATR_state].at_val.at_char = 'R';
+	pjob->ji_wattr[(int)JOB_ATR_substate].at_val.at_long = JOB_SUBSTATE_RUNNING;
+	pjob->ji_wattr[(int)JOB_ATR_state].at_flags |= ATR_VFLAG_MODIFY;
+	pjob->ji_wattr[(int)JOB_ATR_substate].at_flags |= ATR_VFLAG_MODIFY;
+
+	time_resc_updated = time_now;
+	mock_run_mom_set_use(pjob);
+
+	update_ajob_status(pjob);
+	next_sample_time = min_check_poll;
+
+	return;
+
+}
+
+/**
+ * @brief	work task handler for end of a job in mock run mode
+ *
+ * @param[in]	ptask - pointer to the work task
+ *
+ * @return void
+ */
+void
+mock_run_end_job_task(struct work_task *ptask)
+{
+	job *pjob;
+
+	if (ptask == NULL)
+		return;
+
+	pjob = ptask->wt_parm1;
+
+	pjob->ji_qs.ji_substate = JOB_SUBSTATE_EXITING;
+	pjob->ji_qs.ji_state = JOB_STATE_EXITING;
+	pjob->ji_qs.ji_un.ji_momt.ji_exitstat = JOB_EXEC_OK;
+
+	scan_for_exiting();
+}
+
+
+/**
+ * @brief
+ * 	Update the resources used.<attributes> of a job when in mock run mode
+ *
+ * @param[in]	pjob - job in question.
+ *
+ *
+ * @return int
+ * @retval PBSE_NONE	for success.
+ */
+int
+mock_run_mom_set_use(job *pjob)
+{
+	int i;
+	resource *pres;
+	resource *pres_req;
+	attribute *at;
+	attribute *at_req;
+	resource_def *rdefp;
+	long val_req = 0;
+	static resource_def	**rd = NULL;
+	static resource_def *vmemd = NULL;
+	int memval = 0;
+	unsigned int mem_atsv_shift = 10;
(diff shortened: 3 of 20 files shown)


src/resmom/start_exec.c
```diff
@@ -2931,6 +2952,34 @@ finish_exec(job *pjob)
...
+		time_now = time(NULL);
+
+		/* Add a work task that runs when the job is supposed to end */
+		set_task(WORK_Timed, time_now + walltime, mock_run_end_job_task, pjob);
+
+		sprintf(log_buffer, "Started mock run of job");
+		log_event(PBSEVENT_JOB, PBS_EVENTCLASS_JOB,
+			LOG_INFO, pjob->ji_qs.ji_jobid, log_buffer);
+
+		mom_set_use(pjob);
```
arungrover: I don't see a change in mom_set_use, how will you ensure that resources_used are correctly set?

src/resmom/start_exec.c
```diff
@@ -2851,6 +2851,27 @@ receive_job_update_request(int sd)
...
+ * @param[in]	ptask - pointer to the work task
+ *
+ * @return void
+ */
+static void
+mock_run_end_job_task(struct work_task *ptask)
+{
+	job *pjob;
+
+	pjob = ptask->wt_parm1;
```
arungrover: Will it not crash in the cases where a job is deleted and this task triggers after the job is deleted? Or is the mom already deleting all the pending tasks on the basis of param1?

agrawalravi90:
@arungrover thanks a ton for the review! I've addressed your comments, please let me know if the changes look fine now.

src/resmom/mock_run.c
```diff
@@ -0,0 +1,179 @@
...
+			pres = add_resource_entry(at, rdefp);
+			pres->rs_value.at_flags |= ATR_VFLAG_SET;
+			pres->rs_value.at_type = rd[i]->rs_type;
+
+			/*
+			 * get pointer to list of resources *requested* for the job
+			 * so the res used can be set to res requested
+			 */
+			at_req = &pjob->ji_wattr[(int)JOB_ATR_resource];
+			assert(at->at_type == ATR_TYPE_RESC);
```
arungrover: haven't you already checked this at line 106? Also, I am not sure these asserts are necessary

src/resmom/start_exec.c
arungrover: @agrawalravi90 I didn't see a change for this comment. I am ok if you want to keep it as is because this mode is just for testing purposes. Let me know.

src/resmom/mock_run.c
agrawalravi90: I had copied this from set_mom_use(), it should be 'at_req', not 'at' here. I kind of agree with you that this is unnecessary, so ill remove it from both functions.

agrawalravi90:
I've addressed your comments @arungrover 

> @agrawalravi90 I didn't see a change for this comment. I am ok if you want to keep it as is because this mode is just for testing purposes. Let me know.

I did right? I added a NULL check inside mock_run_end_job_task. The function was moved to a new file mock_run.c, maybe that's why it was not clear. Please let me know if it looks okay.

src/resmom/mock_run.c
```diff
@@ -0,0 +1,177 @@
...
+ */
+void
+mock_run_end_job_task(struct work_task *ptask)
+{
+	job *pjob;
+
+	if (ptask == NULL)
+		return;
+
+	pjob = ptask->wt_parm1;
```
arungrover: @agrawalravi90 I guess I was not clear, I meant that if a job is deleted then ptask->wt_parm1 becomes a dangling job pointer, isn't it? In such a case, dereferencing it would result into undesirable outcome (like crash etc.).
Now I think to address it, either we can just store the job id in param1 and do a find_job on the job id in this function. But this will be a little more time consuming because of the find.
OR
We can just delete the task (see delete_task_by_param1_func()) using param1 (which is job pointer) whenever we delete a job in "mock_run" mode.
OR
We just don't care about any of these things because it in testing mode .

I personally would prefer second option. Let me know what you think.

Thanks!

agrawalravi90: thanks for clarifying. i tried a qdel and it got stuck, so I need to investigate the delete further, marking the PR as WIP for now.

agrawalravi90:
@arungrover sorry for the delay, I've fixed the issues with job deletes, also added a PTL test (logs attached). I kind of messed up the commits and had to force push, sorry about that. You can safely ignore the second last commit, you've already reviewed it, just the commit hash is different. thanks again for your help with this PR!

agrawalravi90:
I'm confused why Windows is still failing. The Appveyor logs don't make a lot of sense to me. Requesting help from those more knowledgeable in Windows. @hirenvadalia can you please take a look?

hirenvadalia:
@agrawalravi90 Sure will look into failures and see why they are coming...

src/include/mom_func.h
```diff
@@ -143,6 +143,8 @@ enum hup_action {
  */
 extern enum hup_action	call_hup;
 
+extern int mock_run;
```
hirenvadalia: I suggest (if possible) then moving this in "mock_run.h"

src/resmom/linux/mom_start.c
```diff
@@ -2436,11 +2436,13 @@ scan_for_terminated(void)
 
 	termin_child = 0;
 
-	if (mom_get_sample() == PBSE_NONE) {
-		pjob = (job *)GET_NEXT(svr_alljobs);
-		while (pjob) {
-			mom_set_use(pjob);
-			pjob = (job *)GET_NEXT(pjob->ji_alljobs);
+	if (!mock_run) {
+		if (mom_get_sample() == PBSE_NONE) {
```
hirenvadalia: there are other places also mom_get_sample and traversing of svr_alljobs is used, so shouldn't we update those?

src/resmom/mock_run.c
```diff
@@ -0,0 +1,236 @@
...
+ * organizations to create proprietary derivative works of PBS Pro and
+ * distribute them - whether embedded or bundled with other software -
+ * under a commercial license agreement.
+ *
+ * Use of Altair’s trademarks, including but not limited to "PBS™",
+ * "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
+ * trademark licensing policies.
+ *
+ */
+
```
hirenvadalia: add `#include <pbs_config.h>` to fix all windows build errors.... (Reason: Windows has different pbs_config.h (in win_configure\include directory) which defines a few windows specific macro, which leads to the different behavior of including windows header files (like winsock2.h) as well as few of PBS header files... One of the very important macro is `_WIN32_WINNT` and its value
based on `_WIN32_WINNT` winsock2.h and Windows.h heavily changes its includes... )

src/resmom/mock_run.h
```diff
@@ -0,0 +1,56 @@
...
+ *
+ */
+
+
+#include <stdio.h>
+
+#include "work_task.h"
+#include "job.h"
+
+#ifndef	_MOCK_RUN_H
```
hirenvadalia: move this before including other header files... also add cpp guard also aka
```c
#ifdef	__cplusplus
extern "C" {
#endif

...

#ifdef	__cplusplus
}
#endif
```

src/resmom/start_exec.c
```diff
@@ -6133,6 +6135,16 @@ start_exec(job *pjob)
 	if (do_tolerate_node_failures(pjob))
 		reliable_job_node_add(&pjob->ji_node_list, mom_host);
 
+	if (mock_run) {
```
hirenvadalia: Not sure whether this mock run is only for non-Windows or not... but windows have different "start_exec.c" (inside src\resmom_win), maybe you want to look into that file and change if needed...

Also, (no matter whether it supported or not on windows) just make sure that this code change is not affecting on windows... as we don't run tests on Appveyor... so it better to check manually :smile:

test/tests/functional/pbs_mom_mock_run.py
```diff
@@ -0,0 +1,71 @@
...
+
+        # Submit a job requesting ncpus, mem and walltime
+        attr = {ATTR_l + ".select": "1:ncpus=1:mem=5mb",
+                ATTR_l + ".walltime": "00:00:05"}
+        j = Job(attrs=attr)
+        jid = self.server.submit(j)
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid)
+
+        # This job should end in 5 seconds, let's sleep
+        time.sleep(7)
```
hirenvadalia: Can't we use server.expect to wait for job to finish?

src/include/mom_func.h
agrawalravi90: it's consistent with the way other globals, especially the ones set in getopt, are declared in mom right?

src/resmom/linux/mom_start.c
agrawalravi90: i hadn't seen mom_get_sample() crop up on perf profiling after I added it here, so I didn't look further into it, thanks for pointing it out, I'll add a check at other places as well.

src/resmom/mock_run.c
agrawalravi90: Great, thanks for catching this!

src/resmom/start_exec.c
agrawalravi90: I didn't have plans to add support for Windows for mock run, maybe it can be taken up as a future enhancement if needed.

Appveyor does basic tests right? As long as they pass we should be good.

test/tests/functional/pbs_mom_mock_run.py
agrawalravi90: the accounting_match() does expect like checks, it just happens after the sleep, that way we'd avoid the unnecessary messages on the console saying expect failed attempt x ..

src/resmom/start_exec.c
hirenvadalia: Appveyor just installs msi and runs win_postinstall.py, but doesn't run any tests, no even submitting simple job as we need Linux server to do that.

test/tests/functional/pbs_mom_mock_run.py
hirenvadalia: ok got it

src/resmom/start_exec.c
agrawalravi90: ah, ok, then ill install this on a Windows machine and do some basic checks, thanks for clarifying

src/include/mom_func.h
hirenvadalia: ok, I didn't know that, but in my view, we should keep all the required things in one place for a particular interface... but up you, if you want to keep here, I am fine with it.

src/resmom/start_exec.c
hirenvadalia: Just FYI: If you want you can get MSI from Appveyor, from the artifacts tab.

agrawalravi90:
@hirenvadalia  thanks for approving! @arungrover could you please take a look at the last few commits once ? Thanks in advance!

arungrover:
LGTM
@agrawalravi90 is there anything else left in this PR? I see it in WIP.

agrawalravi90:
@arungrover  thanks for reviewing, yes, I'm just adding a check inside the getopt() block itself for Windows so that it wouldn't be used for Windows.

agrawalravi90:
@arungrover  and @hirenvadalia  Sorry guys, but I realized that there were other areas in mom where it was forking a lot. So, I refactored some code out and made it so that it doesn't fork in mock run mode. Please review the latest commit, I promise not to snowball this PR any further.

src/server/job_func.c
```diff
@@ -958,18 +1012,8 @@ job_purge(job *pjob)
...
-		(void)strcat(namebuf, pjob->ji_qs.ji_jobid);
-	(void)strcat(namebuf, JOB_FILE_SUFFIX);
-	if (unlink(namebuf) < 0) {
-		if (errno != ENOENT) {
-			log_joberr(errno, __func__, msg_err_purgejob,
-				pjob->ji_qs.ji_jobid);
-		}
-	}
+	/* delete job file */
+	del_job_related_file(pjob, JOB_FILE_SUFFIX);
```
hirenvadalia: I think we have to do delete_cred and delete of `.JB` files (see line no. 1018) in mock run also...
Maybe you can create one function like `del_job_creds()` which will call `delete_cred()` as well as delete `.JB` files and call it here as well as in mock_run_job_purge()

Also, I know this won't affect normal mom run, but just to ensure that we are not breaking anything... can you please run a regression with these changes?

agrawalravi90: I'm already deleting .JB, del_job_related_file() will do it when i pass JOB_FILE_SUFFIX right?  I've added a call to delete_cred(), let me know what you think.

I'll do a full regression, thanks.

hirenvadalia: Sorry I mean, JOB_CRED_SUFFIX not JOB_FILE_SUFFIX.
https://github.com/PBSPro/pbspro/blob/b9ed6cf4bd108c3ef3d4da87e061e82514ee4a44/src/server/job_func.c#L1062-L1073

agrawalravi90: but that code is not executed for mom, it's not under #ifdef PBS_MOM, right?

src/resmom/mock_run.c
```diff
@@ -235,3 +237,41 @@ mock_run_mom_set_use(job *pjob)
...
+void
+mock_run_job_purge(job *pjob)
+{
+	char namebuf[MAXPATHLEN + 1] = {'\0'};
+
+	delete_link(&pjob->ji_jobque);
+	delete_link(&pjob->ji_alljobs);
+	delete_link(&pjob->ji_unlicjobs);
+
+#ifdef PBS_MOM
```
arungrover: why is there an ifdef for MOM, is this function called in server too?

agrawalravi90: 'ji_preq' field of pjob is wrapped around #ifdef PBS_MOM, so if I don't wrap it here then compiler barfs

agrawalravi90:
hey, I ran a full regression on this after rebasing with master, so I'm confident that these changes are safe. @arungrover Can i rebase the branch to resolve conflicts?

agrawalravi90:
@arungrover thanks! @hirenvadalia  requesting your feedback, thanks

src/server/job_func.c
hirenvadalia: I think that code is for both server as well as mom because AFAIK both will have cred file for job

agrawalravi90:
@hirenvadalia I've addressed your comment, let me know if it looks ok so that I can rebase, thanks.

agrawalravi90:
@arungrover Please let me know if this looks good now. Thanks!

PR #1551, opened 2020-03-05
Call execjob_end event hooks when releasing unneeded vnodes from a job

lisa-altair:
#### Describe Bug or Feature

Today when pbs_release_nodes (command), pbs_relnodesjob() (IFL), or release_nodes() (Python function) is used to release unneeded nodes from a job only the execjob_epilogue hook scripts will be executed on the early released nodes.  The execjob_end hooks will not be executed on the early released nodes.

#### Describe Your Change

This enhancement will add a call to the hooks registered for an execjob_end event in case of early release of unneeded nodes from a job.  The execjob_epilogue hook event will still execute as before.  The execjob_end hook event will happen after the execjob_epilogue hook event.

#### Link to Design Doc

https://pbspro.atlassian.net/wiki/spaces/PD/pages/1532821513/Call+execjob+end+event+hooks+when+releasing+unneeded+vnodes+from+a+job

TH3_master_output
TH3_addendjob_output
[end_called_output.txt](https://github.com/PBSPro/pbspro/files/4290455/end_called_output.txt)

[valgrind_mainline.txt](https://github.com/PBSPro/pbspro/files/4295645/valgrind_mainline.txt)
[valgrind_withfix.txt](https://github.com/PBSPro/pbspro/files/4295646/valgrind_withfix.txt)


src/include/job.h
```diff
@@ -527,6 +527,7 @@ enum bg_hook_request {
 	BG_PBSE_SISCOMM,
 	BG_IM_DELETE_JOB_REPLY,
 	BG_IM_DELETE_JOB,
+	BG_IM_DELETE_JOB2,
```
mike0042: Is this entry as descriptive as it could be? What distinguishes BG_IM_DELETE_JOB from BG_IM_DELETE_JOB2?

src/resmom/mom_comm.c
```diff
@@ -3784,6 +3784,12 @@ im_request(int stream, int version)
 				} else {
 					runver = pjob->ji_wattr[(int)JOB_ATR_runcount].at_val.at_long;
 				}
+				/* Call the execjob_end hook now */
+				if (mom_process_hooks( HOOK_EVENT_EXECJOB_END, PBS_MOM_SERVICE_NAME, mom_host, hook_input_ptr,
```
mike0042: There is an unnecessary space after the second open paren.

src/include/job.h
lisa-altair: This is the newly created background request that a corresponds to IM_DELETE_JOB2 which is a well defined interface that was introduced with the release nodes feature.

mike0042: I'm not a fan of adding a numeric suffix to distinguish things that are similar, but different. I much prefer a more descriptive name. Since you are simply mirroring the IM_DELETE_JOB2 name, there is really no other option here.

PR #1145, opened 2019-05-28
user reliability workflow - creating reservation out of a job

prakashcv13:
#### Describe Bug or Feature

When a job encounters a problem with application/data/script/etc, the job exits and the resources allocated to the job are released back to the server. If the user wants to run the job again (after correcting the issues that caused the problem), they need to re-submit the job which could take a while to run (depending on various factors like number of jobs in the queue, priority, scheduling policy etc.). If the user needs resources to be allocated for a time period so that the job can be re-submitted and can run without delays, they can use reservations. As of now reservations have a start and end time, so it is not yet possible to have a reservation that is schedule-able.

#### Describe Your Change

This PR focuses on creating a reservation out of a job.

#### Link to Design Doc

[Design](https://pbspro.atlassian.net/wiki/spaces/PD/pages/1450409985/Creating+reservation+out+of+a+job)

#### Attach Test and Valgrind Logs/Output

[hook demo.txt](https://github.com/PBSPro/pbspro/files/4045114/hook.demo.txt)
[qsub demo.txt](https://github.com/PBSPro/pbspro/files/4045115/qsub.demo.txt)
[pbs_rsub demo.txt](https://github.com/PBSPro/pbspro/files/4045125/pbs_rsub.demo.txt)

[ptl.txt](https://github.com/PBSPro/pbspro/files/4223564/ptl.txt)

PR #1552, opened 2020-03-05
Fix build error on RHEL 8

prakashcv13:
#### Describe Bug or Feature
With PR #1145 merged, there was a build issue on RHEL 8.
pbspro/src/server/req_runjob.c:2003:2: error: \xe2\x80\x98strncpy\xe2\x80\x99 output may be truncated copying 256 bytes from a string of length 256 [-Werror=stringop-truncation]

  **strncpy(newreq->rq_user, owner, PBS_MAXUSER)**;

  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

cc1: all warnings being treated as errors

#### Describe Your Change

Removed the strncpy, copying the username directly to preq->rq_user through get_jobowner().

[ptl.txt](https://github.com/PBSPro/pbspro/files/4291755/ptl.txt)

### DIFF
--- a/src/server/req_runjob.c
+++ b/src/server/req_runjob.c
@@ -1989,7 +1989,6 @@ convert_job_to_resv(job *pjob)
 	pbs_list_head *plhed;
 	struct work_task *pwt;
 	struct batch_request *newreq;
-	char owner[PBS_MAXUSER + 1];
 
 	newreq = alloc_br(PBS_BATCH_SubmitResv);
 	if (newreq == NULL) {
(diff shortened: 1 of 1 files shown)


src/server/req_runjob.c
```diff
@@ -1999,8 +1998,7 @@ convert_job_to_resv(job *pjob)
 	}
 	newreq->rq_type = PBS_BATCH_SubmitResv;
 
-	get_jobowner(pjob->ji_wattr[(int)JOB_ATR_job_owner].at_val.at_str, owner);
-	strncpy(newreq->rq_user, owner, PBS_MAXUSER);
+	get_jobowner(pjob->ji_wattr[(int)JOB_ATR_job_owner].at_val.at_str, newreq->rq_user);
```
riyazhakki: Don't you need the user for request authentication?

prakashcv13: instead of having the value in a local variable "owner", I am directly getting it into newreq->rq_user.

PR #1553, opened 2020-03-05
Modified get_log_files to consider the borderline timestamps

visheshh:
#### Describe Bug or Feature
Tests checking for political order were intermittently failing as scheduler analyze does not consider the end time passed in the test case.

#### Describe Your Change
Modified get_log_files to consider the borderline timestamps to parse the logs for analysis.

[Buggy_log.txt](https://github.com/PBSPro/pbspro/files/4292736/Buggy_log.txt)
[fixed_logs.txt](https://github.com/PBSPro/pbspro/files/4292737/fixed_logs.txt)

### DIFF
--- a/test/fw/ptl/utils/pbs_logutils.py
+++ b/test/fw/ptl/utils/pbs_logutils.py
@@ -542,11 +542,11 @@ class PBSLogUtils(object):
                     d1 = time.strftime("%Y%m%d", time.localtime(tm[0]))
                     if start is not None:
                         d2 = time.strftime("%Y%m%d", time.localtime(start))
-                        if d1 < d2:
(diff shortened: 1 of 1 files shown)


bhroam:
@visheshh I'm not sure I understand how your proposed change fixes the issue.  By adding the = in there, you are continuing if d1 and d2 are equal.  This means they won't be returned in the list.  Your description says to have get_log_files() consider the borderline timestamps.  The code looks like it currently does that.  It only skips files that are greater than the end time or less than the start time.  Your proposed change will have it skip files that are equal.

visheshh:
@bhroam Sorry, Looks like i have overlooked the code in hurry. Will open a new PR after the fix.

PR #1555, opened 2020-03-05
ralter resv runs across placement sets

bhroam:
#### Describe Bug or Feature
When a reservation is submitted it will  only run within a placement set.  This is not true for when it is altered.  It can run in multiple placement sets.

This was due to how the reservation ralter code worked.  It attempted to keep as many of the original nodes on the reservation as possible.  To this end, it ignored placement sets.

#### Describe Your Change
I made reconfirming an alter request much closer to how reconfirming a degraded reservation works.  Instead of going out of our way to keep as many of the original nodes as possible, we will reconfirm on any nodes matching our select spec (when the reservation is confirmed).

The fix was to remove the code which tried to keep all the same nodes it could and then tell the node searching code to use the placement sets (which previously it didn't).   Trying to keep the code where we attempt to keep as many nodes as possible is messy.  The first problem is we'll confirm in the first placement that the reservation fits.  This might be on completely different nodes.  The next problem is if we have overlapping placement sets.  We'll keep some nodes but replace others even if the ones we replaced are free.  Since the reservation isn't running yet, any node matching the select spec is as good as any other.  There is no reason to for the added complexity to attempt to keep the same ones.

Since a running reservation is required to stay on the same nodes, this issue didn't exist for them.

I encountered a different bug when fixing this where a job could overlap with a reservation when an ralter fails.   I filed the ticket.  I wrote a test for it and skipped it.  The test can be unskipped when the issue is fixed.

[ralter_before.log](https://github.com/PBSPro/pbspro/files/4295604/ralter_before.log)
[ralter_after.log](https://github.com/PBSPro/pbspro/files/4295616/ralter_after.log)

I ran a valgrind run on TH3 and found nothing of note.

### DIFF
--- a/src/scheduler/node_info.c
+++ b/src/scheduler/node_info.c
@@ -2793,8 +2793,7 @@ eval_placement(status *policy, selspec *spec, node_info **ninfo_arr, place *pl,
 	 *         a ptr to a reordered static array
 	 */
 	if ((pl->pack && spec->total_chunks == 1) ||
-		(conf.provision_policy == AVOID_PROVISION && resresv->aoename != NULL) ||
-		(resresv->is_resv && resresv->resv != NULL && resresv->resv->check_alternate_nodes))
+		(conf.provision_policy == AVOID_PROVISION && resresv->aoename != NULL))
 		nptr = reorder_nodes(ninfo_arr, resresv);
 
 	if (nptr == NULL)
@@ -4244,8 +4243,7 @@ check_resources_for_node(resource_req *resreq, node_info *ninfo,
 
 					if (is_excl(resc_resv->place_spec, ninfo->sharing) || resresv_excl) {
 						min_chunks = 0;
-					}
-					else {
+					} else {
 						cur_res = nres;
 						while (cur_res != NULL) {
 							if (cur_res->type.is_consumable) {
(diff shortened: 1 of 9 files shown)


src/scheduler/node_info.c
```diff
@@ -5093,21 +5091,6 @@ reorder_nodes(node_info **nodes, resource_resv *resresv)
...
-
-			memcpy(nptr, nodes, (nsize + 1) * sizeof(node_info *));
-			for (i = 0; nptr[i] != NULL; i++) {
-				temp = find_node_by_rank(resresv->ninfo_arr, nptr[i]->rank);
-				if (temp != NULL)
-					nptr[i]->nscr.to_be_sorted = 0;
-				else
-					nptr[i]->nscr.to_be_sorted = 1;
-			}
-			qsort(nptr, i, sizeof(node_info*), cmp_nodes_sort);
```
arungrover: You might want to remove cmp_nodes_sort as well. looks like it is of no use now.

test/tests/functional/pbs_ralter.py
```diff
@@ -1427,3 +1427,91 @@ def test_auth_group_restart(self):
...
+        color1 = self.server.status(NODE, c, id=nodes[0])[0][c]
+        color2 = self.server.status(NODE, c, id=nodes[1])[0][c]
+        self.assertEqual(color1, color2)
+
+        self.alter_a_reservation(rid1, start1, end1, shift=300,
+                                 alter_e=True, whichMessage=3)
+
+        t = start1 - int(time.time())
+        self.logger.info('Sleeping until reservation starts')
+        self.server.expect(RESV, {'reserve_state': (MATCH_RE, 'RESV_RUNNING')},
```
arungrover: you are missing "|5" in  "RESV_RUNNING"?

PR #1557, opened 2020-03-06
Round of the microsecond value of the time parsed from logs

visheshh:
#### Describe Bug or Feature
Time passed to in_range function to return log lines was in microseconds because of that condition tm<=endtime (ex: 55.234 <= 55)  would fail and was not returning the log lines.

#### Describe Your Change
Rounded of the time parsed to 0th decimal. So that would match the end time to return the log lines,

[buggy_output_sched_logs.txt](https://github.com/PBSPro/pbspro/files/4296331/buggy_output_sched_logs.txt)
[buggy_output.txt](https://github.com/PBSPro/pbspro/files/4296332/buggy_output.txt)
[fixed_output_sched_logs.txt](https://github.com/PBSPro/pbspro/files/4296333/fixed_output_sched_logs.txt)
[fixed_output.txt](https://github.com/PBSPro/pbspro/files/4296334/fixed_output.txt)

### DIFF
--- a/test/fw/ptl/utils/pbs_logutils.py
+++ b/test/fw/ptl/utils/pbs_logutils.py
@@ -1293,6 +1293,7 @@ class PBSSchedulerLog(PBSLogAnalyzer):
         m = self.tm_tag.match(rec)
         if m:
             tm = self.logutils.convert_date_time(m.group('datetime'))
+            tm = round(tm, 0)
             self.record_tm.append(tm)
             if self.logutils.in_range(tm, start, end):
                 rv = self._parse_line(rec)


test/fw/ptl/utils/pbs_logutils.py
```diff
@@ -1293,6 +1293,7 @@ def scheduler_parsing(self, rec, start, end):
         m = self.tm_tag.match(rec)
         if m:
             tm = self.logutils.convert_date_time(m.group('datetime'))
+            tm = round(tm, 0)
```
hirenvadalia: I suggest this doing a test case that needs it. Because in my view the library should not touch the value given to it.
What if someone doesn't want to round it and use exactly what test/user-provided.
From python side, python is happy comparing float with int so if the user passes float as argument then it won't cause any problem:
```shell
$ python3
Python 3.6.9 (default, Nov  7 2019, 10:44:02) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> 55.23 < 55
False
>>> 55.23 < 55.30
True
>>> 55 < 55.30
True
>>> exit()
```

visheshh: @Hiren this value is not passed from test case. This is the value that is parsed from sched logs. Are you suggesting all test cases should pass the start time and end time in microseconds from now on. We need to make changes to all those test cases.

hirenvadalia: Assuming you are talking about `Scheduler.cycles()`:
We are already have start and end time in `cycles()` which will be passed way down to sched log analyzer and method u are changing as below:

```Scheduler.cycles() -> PBSSchedulerLog.analyse() -> (PBSLogAnalyzer.get_log_files() + ) PBSLogAnalyzer._log_parser() -> PBSLogAnalyzer.comp_analyze() -> PBSLogAnalyzer.parse_block() (+ see below) -> PBSLogUtils.in_range()```

(PBSLogAnalyzer.parse_custom_tag() nobody uses it in PTL as of now, its kinda dead code...
PBSLogAnalyzer.parse_conditional() the only pbs_loganalyzer uses it and parse_conditional() don't have to care about the start and end time as it is based on conditional/pattern matching, however, start and end time is already applied in get_log_files() so given times will be honored by log analyzer with conditional matching)

visheshh:
@agrawalravi90 Value 55 would be the end time passed from the test case.

test/fw/ptl/utils/pbs_logutils.py
visheshh: @hirenvadalia The path for sched would be PBSSchedulerLog.comp_analyze() -> PBSSchedulerLog.scheduler_parsing if i am not wrong. from scheduler_parsing call to in_range is done.

get_log_files() checks start time and end time only with date to get the number of log files it does not parse the file for messages.

visheshh:
@agrawalravi90 @hirenvadalia I guess the older test cases were typecasting time to int before sending and that was the main issue. As we have microsecond logging now. We need to remove int()'s from time from all the test cases.

test/fw/ptl/utils/pbs_logutils.py
hirenvadalia: @visheshh Yes, but even in that case also it passes start/end time in in_range which is passed by the test/user, so we should change test case, not the library.

agrawalravi90:
> @agrawalravi90 @hirenvadalia I guess the older test cases were typecasting time to int before sending and that was the main issue. As we have microsecond logging now. We need to remove int()'s from time from all the test cases.

that makes more sense, yes

visheshh:
@agrawalravi90 @hirenvadalia Thanks for the suggestions will close the ticket.

PR #1559, opened 2020-03-06
set ncpus for user reliability tests

prakashcv13:
#### Describe Bug or Feature

My assumption in PR #1145 that the machine will have at least 4 ncpus is incorrect, so the tests will fail on machines with less than 3 ncpus as the job demands for that many ncpus, and would not start running.

#### Describe Your Change

Added setUp() to set ncpus.

[ptl_before.txt](https://github.com/PBSPro/pbspro/files/4298729/ptl_before.txt)

### DIFF
--- a/test/tests/functional/pbs_user_reliability.py
+++ b/test/tests/functional/pbs_user_reliability.py
@@ -41,6 +41,11 @@ import re
 
 class Test_user_reliability(TestFunctional):
 
+    def setUp(self):
+        TestFunctional.setUp(self)
+        a = {'resources_available.ncpus': 4}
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_user_reliability.py
```diff
@@ -41,6 +41,11 @@
 
 class Test_user_reliability(TestFunctional):
 
+    def setUp(self):
+        TestFunctional.setUp(self)
+        a = {'resources_available.ncpus': 4}
+        self.server.manager(MGR_CMD_SET, NODE, a, self.mom.shortname)
```
bhroam: Now that you are setting the number of cpus, you need to skip these tests on cpuset machines.

PR #1560, opened 2020-03-07
dup_timed_event() wasn't duping the disabled bit

bhroam:
#### Describe Bug or Feature
The function dup_timed_event() wasn't copying over the disabled bit.  This means when the universe was dup'd, the disabled events would be performed.  This could cause top jobs to be scheduled too soon.  This really only affected preemption with sched_preempt_enforce_resumption on.  This was the only place timed events would be marked disabled and then the universe was dup'd.  It would only affect the cycle jobs are preempted.  Future cycles would be fine.

#### Describe Your Change
In dup_timed_event(), copy over the disabled bit.

There really isn't any bang for our buck for writing a test.  It only affects a small corner case, and the outcome wasn't that bad.

### DIFF
--- a/src/scheduler/simulate.c
+++ b/src/scheduler/simulate.c
@@ -1099,6 +1099,7 @@ dup_timed_event(timed_event *ote, server_info *nsinfo)
 		return NULL;
 
 	nte = create_event(ote->event_type, ote->event_time, event_ptr, ote->event_func, ote->event_func_arg);
+	set_timed_event_disabled(nte, ote->disabled);
 
 	return nte;
 }


src/scheduler/simulate.c
```diff
@@ -1099,6 +1099,7 @@ dup_timed_event(timed_event *ote, server_info *nsinfo)
 		return NULL;
 
 	nte = create_event(ote->event_type, ote->event_time, event_ptr, ote->event_func, ote->event_func_arg);
+	set_timed_event_disabled(nte, ote->disabled);
```
agrawalravi90: why not just do "nte->disabled = ote->disabled" ?

bhroam: If there is a setter, why not use it?  It means if I ever need to do something every time the disabled bit is set, I can do it there and trust it will always happen when an event is disabled.

vstumpf: But the event should just be duped, right? So if we act on it, it might no longer be a dupe.

bhroam: A good example is the call to create_event().  It handles setting the run_event and end_event pointers to a resource_resv structure.  This is something that needs to be done every time we create an event.  The setter should do everything to disable an event.  If an event is disabled in the old universe, it should be disabled in the new universe.  I don't understand why we shouldn't call it.

agrawalravi90: > if there is a setter, why not use it?

Since the setter doesn't do anything other than assignment, it seemed unnecessary. If you fore-see us adding more code in the future, then cool. The -O2 flag should inline this function anyways, so there shouldn't be any overhead as such.

PR #1561, opened 2020-03-09
"PTL test_preempt_checkpoint_requeue uses the wrong checkpoint script

birbalsain:
#### Describe Bug or Feature
"PTL test_preempt_checkpoint_requeue expects checkpoint should fail and job preempt due to requeue. But job is preempted by checkpoint . So to match the test definition, test should use checkpoint fail script. 

#### Describe Your Change
uses checkpoint fail script so that checkpoint fails and job preempts by requeue.

[checkpoint_requeue.txt](https://github.com/PBSPro/pbspro/files/4305022/checkpoint_requeue.txt)

### DIFF
--- a/test/tests/functional/pbs_preemption.py
+++ b/test/tests/functional/pbs_preemption.py
@@ -190,7 +190,7 @@ exit 1
         """
         Test that when checkpoint fails, a job is correctly requeued
         """
-        self.mom.add_checkpoint_abort_script(body=self.chk_script)
+        self.mom.add_checkpoint_abort_script(body=self.chk_script_fail)
         self.submit_and_preempt_jobs(preempt_order='CR')
 
     @skipOnCpuSet


bhroam:
I like the current test name better.  With your name, it makes it sound like the entire preemption fails.  The test is about checkpoint failing and falling over to requeue.

bhroam:
Well actually if you were confused by the title, how about test_preempt_chkpfail_requeue or maybe add the _ betweek chkp and fail

agrawalravi90:
Oh, I wasn't actually suggesting a rename of the test's name. The test is currently called "test_preempt_checkpoint_requeue", I was just suggesting that we change the title of the PR to say that the test failed and explain the reason for failure in the PR description.
That said, I don't oppose your rename suggestion, but we don't really need the word 'preempt' there, the test suite's name contains preemption already.

bhroam:
@agrawalravi90 I agree that we can strip out preempt, but if we do it, we should strip it out of all the test names in the entire test plan.

birbalsain:
@agrawalravi90 I have changed the PR description and heading.

PR #1562, opened 2020-03-09
test_attribute_case.TestExpect failed while verifying queue attribute

sandisamp:
#### Describe Bug or Feature

Test failed in verifying queue attribute. In test 'expressq' created with priority 150. after creating test try to verify queue attribute, but test cannot get expected output and test got failed. 

Server.except can work with lower cased attribute but Server.status and/or BatchUtils.convert_to_dictlist doesn't understand lower cased attribute name so when Server.expect is called with lower cased attribute name, which internally passed to Server.status->BatchUtils.convert_to_dictlist, but BatchUtils.convert_to_dictlist doesn't incude that attribute from PBS command's response to returned dict.

#### Describe Your Change

Function  convert_to_dictlist checks for both lower case and regular attribute while comparing and includes in the same in the dictlist returned.

[test_attribute_case.TestExpect.txt](https://github.com/PBSPro/pbspro/files/4305758/test_attribute_case.TestExpect.txt)

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -1780,7 +1780,7 @@ class BatchUtils(object):
                 m = self.pbsobjattrval_re.match(l)
                 if m:
                     attr = m.group('attribute')
-                    if (attribs is None or str(attr).lower() in attribs
+                    if (attribs is None or attr.lower() in attribs
                             or attr in attribs):
                         if attr in d:
                             d[attr] = d[attr] + "," + m.group('value')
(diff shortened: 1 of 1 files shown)


test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -1780,7 +1780,8 @@ def convert_to_dictlist(self, l, attribs=None, mergelines=True, id=None):
                 m = self.pbsobjattrval_re.match(l)
                 if m:
                     attr = m.group('attribute')
-                    if attribs is None or attr in attribs:
+                    if (attribs is None or str(attr).lower() in attribs
+                            or attr in attribs):
```
bhroam: I'd either str() both instances of attr, or not str() them.  It should be a string coming from an re match, right?

PR #1564, opened 2020-03-09
Hup mom to run exechost_startup hook

birbalsain:
#### Describe Bug or Feature
custom resource is not set by the execjob_startup hook

#### Describe Your Change
HUP the mom to run execjob_startup hook

[custom_res.txt](https://github.com/PBSPro/pbspro/files/4306951/custom_res.txt)

### DIFF
--- a/test/tests/functional/pbs_resc_custom_perm.py
+++ b/test/tests/functional/pbs_resc_custom_perm.py
@@ -82,8 +82,8 @@ e.vnode_list[localnode].resources_available['foo_strh'] = "str_hook"
             startup_hook_body,
             overwrite=True)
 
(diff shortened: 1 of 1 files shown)


vstumpf:
Why doesn't a restart work? Restart should bring the mom down and back up, then the hook should run on startup. Can you reproduce the original bug?

bhroam:
@birbalsain please add context to the PR description.  It doesn't make much sense to me.

agrawalravi90:
@birbalsain as a general advice, we use the PR as a bug report as well, so try to write describe the issue exactly as you would in a bug report, and then you can explain the fix.

birbalsain:
Unable to reproduce issue.

PR #1568, opened 2020-03-10
Buffer overflow corrupts heap in MoM child responsible for file staging

Jitendra-Rajak:
#### Describe Bug or Feature
Bad_list is a fairly large string that can, amongst others, contain scp verbose logging. Frequently that is a multiline string that is larger than the log buffer.
When the bad_list is too large these will, without fortified source, corrupt the heap (but often of a process about to exit), and with fortified source will immediately make the process abort. That caused some grief  with a fortified source build made to detect an unrelated heap corruption issue.

#### Describe Your Change
Instead of using sprintf() now we are using pbs_asprintf() and also splitting the bad_list on the basis of  \n in the list to avoid the heap corruption. 

[log_del_buffer_overflow.txt](https://github.com/PBSPro/pbspro/files/4312502/log_del_buffer_overflow.txt)

[stage_buffer_overflow.txt](https://github.com/PBSPro/pbspro/files/4312499/stage_buffer_overflow.txt)

[mom_log_after_changes.txt](https://github.com/PBSPro/pbspro/files/4312503/mom_log_after_changes.txt)

[valgrind_log.txt](https://github.com/PBSPro/pbspro/files/4312511/valgrind_log.txt)

### DIFF
--- a/src/resmom/requests.c
+++ b/src/resmom/requests.c
@@ -3403,10 +3403,11 @@ req_cpyfile(struct batch_request *preq)
 				if ((pbs_asprintf(&temp_buff, "%s\n", token)) != -1) {
 					log_event(PBSEVENT_DEBUG2, PBS_EVENTCLASS_JOB, LOG_DEBUG,
 						dup_rqcpf_jobid, temp_buff);
-					token = strtok_r(NULL, "\n", &save_ptr);
 					free(temp_buff);
 					temp_buff = NULL;
-				}
+				} else
+					break;
+				token = strtok_r(NULL, "\n", &save_ptr);
 			}
 			log_event(PBSEVENT_DEBUG2, PBS_EVENTCLASS_JOB, LOG_DEBUG,
 					dup_rqcpf_jobid, "---->>>>");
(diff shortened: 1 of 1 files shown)


src/resmom/requests.c
```diff
@@ -3392,9 +3392,24 @@ req_cpyfile(struct batch_request *preq)
 		}
 	} else {
 		if (stage_inout.bad_files) {
-			sprintf(log_buffer, "Job files not copied: %s", stage_inout.bad_list);
+			char* token;
```
prakashcv13: "*" should be attached to token. same for other variables.

src/resmom/requests.c
```diff
@@ -3392,9 +3392,24 @@ req_cpyfile(struct batch_request *preq)
...
+			token = strtok_r(rest, "\n", &save_ptr);
+			while (token != NULL) {
+				char *temp_buff;
+				if ((pbs_asprintf(&temp_buff, "%s\n", token)) != -1) {
+					log_event(PBSEVENT_DEBUG2, PBS_EVENTCLASS_JOB, LOG_DEBUG,
+						dup_rqcpf_jobid, temp_buff);
+					token = strtok_r(NULL, "\n", &save_ptr);
+				}
+				free(temp_buff);
+				temp_buff = NULL;
```
prakashcv13: temp_buff should be free()-ed only when pbs_asprintf() succeeds.

prakashcv13:
@Jitendra-Rajak - could you please attach valgrind test logs with the latest change.

Jitendra-Rajak:
[new_valgrind_log.txt](https://github.com/PBSPro/pbspro/files/4314275/new_valgrind_log.txt)

@prakashcv13, new valgrind logs attached.

src/resmom/requests.c
```diff
@@ -3392,9 +3392,24 @@ req_cpyfile(struct batch_request *preq)
...
+			char *token = NULL; 
+			char *rest = stage_inout.bad_list; 
+			char *save_ptr = NULL;
 			log_event(PBSEVENT_DEBUG2, PBS_EVENTCLASS_JOB, LOG_DEBUG,
-					dup_rqcpf_jobid, log_buffer);
+				dup_rqcpf_jobid, "Job files not copied:---->>>>");
+			token = strtok_r(rest, "\n", &save_ptr);
+			while (token != NULL) {
+				char *temp_buff = NULL;
+				if ((pbs_asprintf(&temp_buff, "%s\n", token)) != -1) {
```
bhroam: If pbs_asprintf() returns -1, you don't advance token.  pbs_asprintf() will likely continue to return -1, and this will become an infinite loop.  You should move the strtok_r() outside of the if statement, break out of the loop, or bail out of the function if pbs_asprintf() returns -1.  It does that on a malloc() failure.

PR #1509, opened 2020-02-06
Support for reservation in multi-sched environment

arungrover:
#### Describe Bug or Feature
Adding support to confirm/run reservations in multi-sched environment.
Interface 8 of [Multi-sched](https://pbspro.atlassian.net/wiki/spaces/PD/pages/50947131/PP-337+Multiple+schedulers+servicing+the+PBS+cluster) design

#### Describe Your Change
With this change in place, from now on, PBS server will send reservation confirmation requests to all schedulers in the complex. Reservation is marked confirmed as soon as any of the schedulers is able to confirm the reservation.

#### Link to Design Doc
Interface 8 of [Multi-sched](https://pbspro.atlassian.net/wiki/spaces/PD/pages/50947131/PP-337+Multiple+schedulers+servicing+the+PBS+cluster) design

#### Attach Test and Valgrind Logs/Output
[test_multi_sched.txt](https://github.com/PBSPro/pbspro/files/4167363/test_multi_sched.txt)

PR #1569, opened 2020-03-10
Reservation queue from default partition fails to recover

arungrover:
#### Describe Bug or Feature
After the change made in #1509 PBS Server now disallows queue/scheduler/node to have the partition name set to pbs-default. This default partition is internally assigned to these three objects by the server.
PBS server failed to recover objects (queue & reservation) which had partition set to pbs-default because the action function for this attribute failed.

#### Describe Your Change
Changed the action functions to ignore the validation of attributes when the server is recovering attributes from the database.

#### Link to Design Doc
NA

[test_before_fix.txt](https://github.com/PBSPro/pbspro/files/4314768/test_before_fix.txt)
[test_after_fix.txt](https://github.com/PBSPro/pbspro/files/4314769/test_after_fix.txt)

### DIFF
--- a/test/tests/functional/pbs_reservations.py
+++ b/test/tests/functional/pbs_reservations.py
@@ -2156,3 +2156,28 @@ class TestReservations(TestFunctional):
 
         self.server.expect(RESV, 'queue', op=UNSET, id=rid, offset=10)
         self.server.expect(SERVER, {'resources_assigned.ncpus': 0})
+
+    def test_server_recover_resv_queue(self):
+        """
+        Test that PBS server can recover a reservation queue after a
+        restart
+        """
+
+        a = {'resources_available.ncpus': 1}
+        self.server.create_vnodes('vn', a, num=2, mom=self.mom)
+        now = int(time.time())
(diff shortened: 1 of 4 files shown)


arungrover:
[test_after_fix.txt](https://github.com/PBSPro/pbspro/files/4314903/test_after_fix.txt)

test/tests/functional/pbs_reservations.py
```diff
@@ -2156,3 +2156,28 @@ def test_standing_resv_resc_used(self):
...
+    def test_server_recover_resv_queue(self):
+        """
+        Test that PBS server can recover a reservation queue after a
+        restart
+        """
+
+        a = {'resources_available.ncpus': 1}
+        self.server.create_vnodes('vn', a, num=2, mom=self.mom)
+        now = int(time.time())
+        rid = self.submit_reservation(user=TEST_USER, select='1:ncpus=1',
```
prakashcv13: do we need to provide 1:ncpus=1?

arungrover: We don't have to. But since I'm using a helper function that has been used in other tests too, this function expects select, start, end and user to be provided

PR #1572, opened 2020-03-11
Tests of TestServerDynRes are failing while verifying the job's comment

Mahalty:
#### Describe Bug or Feature

- Tests "test_invalid_script_out, test_multiple_res,test_res_string,test_res_string_array,test_res_size" of TestServerDynRes are failing on certain machines during regression 

#### Describe Your Change

- In the test we are submitting more than 1 job and expecting certain job comment but the default "resources_available.ncpus" on few machines is 1 due to which job's comment differs and hence the test failure.

- Change the tests to not use default job script(sleep of 100s) but use short script(sleep 5) and remove skipOnCpuset decorator from the tests as they can be run on Cpuset machines with the said changes.

- Modiifed test 'test_res_size' to wait  for job1 to complete instead of deleting it, as we are now submitting shortjob and deleting would sometimes race condition. Also updated the third job submission to request foobar=100gb.

- [Execution_logs_TestServerDynRes_x55_aftr_fix.txt](https://github.com/PBSPro/pbspro/files/4316767/Execution_logs_TestServerDynRes_x55_aftr_fix.txt)

- [Execution_logs_TestServerDynRes_x55_bfr_fix.txt](https://github.com/PBSPro/pbspro/files/4316769/Execution_logs_TestServerDynRes_x55_bfr_fix.txt)

- [Execution_logs_TestServerDynRes_x57_aftr_fix.txt](https://github.com/PBSPro/pbspro/files/4316770/Execution_logs_TestServerDynRes_x57_aftr_fix.txt)

- [Execution_logs_TestServerDynRes_x57_bfr_fix.txt](https://github.com/PBSPro/pbspro/files/4316771/Execution_logs_TestServerDynRes_x57_bfr_fix.txt)

- [Execution_logs_TestServerDynRes_x69_aftr_fix.txt](https://github.com/PBSPro/pbspro/files/4316772/Execution_logs_TestServerDynRes_x69_aftr_fix.txt)

- [Execution_logs_TestServerDynRes_x69_bfr_fix.txt](https://github.com/PBSPro/pbspro/files/4316773/Execution_logs_TestServerDynRes_x69_bfr_fix.txt)

### DIFF
--- a/test/tests/functional/pbs_svr_dyn_res.py
+++ b/test/tests/functional/pbs_svr_dyn_res.py
@@ -91,11 +91,14 @@ class TestServerDynRes(TestFunctional):
         self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'True'})
         return scripts
 
+    @skipOnCpuSet
     def test_invalid_script_out(self):
         """
         Test that the scheduler handles incorrect output from server_dyn_res
         script correctly
         """
+        a = {'resources_available.ncpus': 2}
+        self.server.manager(MGR_CMD_SET, NODE, a, id=self.mom.shortname)
         # Create a server_dyn_res of type long
         resname = ["mybadres"]
         restype = ["long"]
@@ -227,6 +230,8 @@ class TestServerDynRes(TestFunctional):
         Test multiple dynamic resources specified in resourcedef
         and sched_config
         """
+        a = {'resources_available.ncpus': 3}
+        self.server.manager(MGR_CMD_SET, NODE, a, id=self.mom.shortname)
         # Create resources of type long
         resname = ["foobar_small", "foobar_medium", "foobar_large"]
         restype = ["long", "long", "long"]
@@ -263,11 +268,14 @@ class TestServerDynRes(TestFunctional):
         a = {'job_state': 'R', 'Resource_List.foobar_large': 18}
         self.server.expect(JOB, a, id=jid)
 
+    @skipOnCpuSet
     def test_res_string(self):
         """
         Test that server_dyn_res accepts a string value returned
         by a script
         """
+        a = {'resources_available.ncpus': 2}
+        self.server.manager(MGR_CMD_SET, NODE, a, id=self.mom.shortname)
         # Create a resource of type string
         resname = ["foobar"]
         restype = ["string"]
@@ -299,11 +307,14 @@ class TestServerDynRes(TestFunctional):
         a = {'job_state': 'Q', 'comment': job_comment}
         self.server.expect(JOB, a, id=jid, attrop=PTL_AND)
 
+    @skipOnCpuSet
     def test_res_string_array(self):
         """
         Test that server_dyn_res accepts string array returned
         by a script
         """
+        a = {'resources_available.ncpus': 2}
+        self.server.manager(MGR_CMD_SET, NODE, a, id=self.mom.shortname)
         # Create a resource of type string_array
         resname = ["foobar"]
         restype = ["string_array"]
@@ -341,6 +352,8 @@ class TestServerDynRes(TestFunctional):
         Test that server_dyn_res accepts type "size" and a "value"
         returned by a script
         """
+        a = {'resources_available.ncpus': 2}
+        self.server.manager(MGR_CMD_SET, NODE, a, id=self.mom.shortname)
         # Create a resource of type size
         resname = ["foobar"]
         restype = ["size"]
--- a/test/tests/functional/pbs_svr_dyn_res.py
+++ b/test/tests/functional/pbs_svr_dyn_res.py
@@ -344,8 +344,6 @@ class TestServerDynRes(TestFunctional):
         Test that server_dyn_res accepts type "size" and a "value"
         returned by a script
         """
-        a = {'resources_available.ncpus': 2}
-        self.server.manager(MGR_CMD_SET, NODE, a, id=self.mom.shortname)
         # Create a resource of type size
         resname = ["foobar"]
         restype = ["size"]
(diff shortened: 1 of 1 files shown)


Mahalty:
> In all the tests - instead of setting ncpus - let us submit short jobs like "echo hello" and wait for the jobs to get over.
> That way we need not set ncpus - and these tests can still be executable on cpuset machines.
> What do you think?

@prakashcv13: Few of the tests expects job to be in Q state with particular job comment. If we submit short jobs then queued jobs would end up running.

Mahalty:
- [Execution_logs_TestServerDynRes_uv.txt](https://github.com/PBSPro/pbspro/files/4318105/Execution_logs_TestServerDynRes_uv.txt)

- [Execution_logs_TestServerDynRes_x69_aftr_fix_1.txt](https://github.com/PBSPro/pbspro/files/4318106/Execution_logs_TestServerDynRes_x69_aftr_fix_1.txt)

test/tests/functional/pbs_svr_dyn_res.py
```diff
@@ -341,6 +345,8 @@ def test_res_size(self):
         Test that server_dyn_res accepts type "size" and a "value"
         returned by a script
         """
+        a = {'resources_available.ncpus': 2}
+        self.server.manager(MGR_CMD_SET, NODE, a, id=self.mom.shortname)
```
prakashcv13: if we want one job to run and for other we want "Can never run", we do not need the first job to be in 'R' state when we submit the second job. We can do away with this as well.

Mahalty: @prakashcv13 In the test server dynamic resource script has 100gb as the value for resource foobar and in the end we are submitting 2 jobs requesting foobar=50gb and verifying both the job should run. That's why I have added code for setting ncpus

prakashcv13: ok - I did not see the last 2 jobs.
I am of the view that we do not need that to be tested as from the first two jobs it is clear that "server_dyn_res accepts type "size" and a "value" returned by a script" which is the purpose of this test case.
What do you think?

Mahalty: @prakashcv13 Yeah the initial job submissions would verify the intent of the test but I am thinking that the last 2 jobs submitted with a intention to verify that jobs can run when resource(foobar) requested is 100gb in total(which is the limit set)

prakashcv13: in that case submit one job that uses 100gb. unless the test case itself is that more than one job must run, we can do that.

Mahalty:
**Latest execution-logs:**
[Execution_logs_TestServerDynRes.txt](https://github.com/PBSPro/pbspro/files/4322369/Execution_logs_TestServerDynRes.txt)

test/tests/functional/pbs_svr_dyn_res.py
```diff
@@ -371,24 +376,15 @@ def test_res_size(self):
...
         a = {'job_state': 'Q', 'comment': job_comment}
         self.server.expect(JOB, a, id=jid2, attrop=PTL_AND)
-
-        # Delete jobs
         self.server.deljob(jid1, wait=True, runas=TEST_USER)
         self.server.deljob(jid2, wait=True, runas=TEST_USER)
 
         # Submit jobs again
-        a = {'Resource_List.foobar': '50gb'}
+        a = {'Resource_List.foobar': '100gb'}
```
prakashcv13: now that we have only one job, do we still need to set ncpis on line 347, 348

test/tests/functional/pbs_svr_dyn_res.py
```diff
@@ -371,24 +374,15 @@ def test_res_size(self):
...
-        j2 = Job(TEST_USER, attrs=a)
-        jid2 = self.server.submit(j2)
-
-        # Both jobs must run successfully
-        a = {'job_state': 'R', 'Resource_List.foobar': '50gb'}
+        a = {'job_state': 'R', 'Resource_List.foobar': '100gb'}
         self.server.expect(JOB, a, id=jid1)
-        self.server.expect(JOB, a, id=jid2)
 
     @skipOnCpuSet
```
prakashcv13: shall we remove this skip as well? the test case does not look like setting ncpus.

Mahalty: @prakashcv13: As scope of this PR, I have just modified the tests which are mentioned in the PR. 
I see only one of the tests now use the decorator so I will change that as well and update the description of the PR accordingly.

Mahalty:
**Latest execution-logs:**
[Execution_logs_TestServerDynRes_td-09.txt](https://github.com/PBSPro/pbspro/files/4322990/Execution_logs_TestServerDynRes_td-09.txt)
[Execution_logs_TestServerDynRes_uv-01.txt](https://github.com/PBSPro/pbspro/files/4322991/Execution_logs_TestServerDynRes_uv-01.txt)
[Execution_logs_TestServerDynRes_x57.txt](https://github.com/PBSPro/pbspro/files/4323002/Execution_logs_TestServerDynRes_x57.txt)

I have updated the deescription of PR accordingly

prakashcv13:
@Mahalty - thank you for making the code changes and PR description. LGTM.
We should also change the PR summary.

Mahalty:
> @Mahalty - thank you for making the code changes and PR description. LGTM.
> We should also change the PR summary.

Done

prakashcv13:
@dinesh1306 - could you please let us know, if there are any comments. There have been changes after you approved.

PR #1573, opened 2020-03-11
all tests of "TestPbsHookAlarmLargeMultinodeJob" timeout on using  create_vnodes() to check each vnode state is free

birbalsain:
#### Describe Bug or Feature
all tests of "TestPbsHookAlarmLargeMultinodeJob" timeout on using create_vnodes() to check each vnodes state is free because it does a pbsnodes -v on each vnode. This takes a long time.

#### Describe Your Change
uses  expect=False in create_vnodes(). outside the create_vnodes(), checks vnodes state is free.

[multinode_timeout.txt](https://github.com/PBSPro/pbspro/files/4329470/multinode_timeout.txt)

### DIFF
--- a/test/tests/resilience/pbs_hook_alarm_large_multinode_job.py
+++ b/test/tests/resilience/pbs_hook_alarm_large_multinode_job.py
@@ -55,11 +55,16 @@ class TestPbsHookAlarmLargeMultinodeJob(TestResilience):
 
         a = {'resources_available.mem': '1gb',
              'resources_available.ncpus': '1'}
-        self.server.create_vnodes(self.mom.shortname, a, 5000, self.mom)
+        self.server.create_vnodes(self.mom.shortname, a, 5000, self.mom,
+                                  expect=False)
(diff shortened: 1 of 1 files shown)


test/tests/resilience/pbs_hook_alarm_large_multinode_job.py
```diff
@@ -59,7 +59,7 @@ def setUp(self):
         # Restart mom explicitly due to PP-993
         self.mom.restart()
 
-    @timeout(400)
+    @timeout(1800)
```
bhroam: This is a pretty serious increase from 6.5m to 30m.  Do you really think these tests will take that long?

vstumpf:
Don't do this, just create the vnodes and then do 1 expect for 5000 vnodes.
See an example here:
https://github.com/PBSPro/pbspro/blob/1e50b52f21dcd8c2779973abe9e45013fe971058/test/tests/functional/pbs_node_buckets.py#L71-L77

birbalsain:
@bhroam i have done the requested changes.

PR #1574, opened 2020-03-11
Fixing user reliability test cases

prakashcv13:
#### Describe Bug or Feature

My assumption in PR #1145 that the machine will have at least 4 ncpus is incorrect, so the tests will fail on machines with less than 3 ncpus as the job demands for that many ncpus, and would not start running.

#### Describe Your Change

Removed the select for 3 ncpus.
If we set ncpus, the test cases cannot be run on cpuset machines.

[ptl_th.txt](https://github.com/PBSPro/pbspro/files/4319392/ptl_th.txt)


prakashcv13:
thank you @bhroam and @vstumpf

PR #1576, opened 2020-03-12
some tests from 'TestReservations' failing on cpuset machines

dinesh1306:
#### Describe Bug or Feature
Following tests from testsuite 'TestReservations' failing on cpuset machines.
1. test_future_resv_conflicts_running_job
2. test_resv_server_restart
3. test_asap_delete_idle_resv_set

* test_future_resv_conflicts_running_job:
in test job is submitted exclusive job without walltime and expect job should be in R state. After that submitting a reservation on same node. on single node host Reservation declined and test got expected result, but on cpuset mom reservation was confirmed on another vnode and Test cannot get expected 'Reservation denied' in server logs and test got failed.

* test_resv_server_restart:
in test reservation is submitted on local host, but on cpuset machine submitted reservation is placed on cpuset base node due to PBS cannot confirmed reservation and test got failed.    

* test_asap_delete_idle_resv_set:
in test we are changing resources_available.ncpus=1. On cpuset mom if we try to change the available ncpus value then we get mom log message "make_cpuset

#### Describe Your Change
* test_future_resv_conflicts_running_job:
if machine is cpuset machine then submit job and reservation on same node using vnode.

*  test_resv_server_restart:
if machine is cpuset machine then find out vnode in cpuset mom and submit reservation with vnode, while checking node status use vnode value. 

* test_asap_delete_idle_resv_set:
add skipOnCpuSet decorator to skip test on cpuset machine.

* before_fix:
[res_before_fix_oncpuset.txt](https://github.com/PBSPro/pbspro/files/4323104/res_before_fix_oncpuset.txt)

* after_fix:
[res_after_fix_on_normal_machine.txt](https://github.com/PBSPro/pbspro/files/4323107/res_after_fix_on_normal_machine.txt)
[res_after_fix_oncpuset.txt](https://github.com/PBSPro/pbspro/files/4323108/res_after_fix_oncpuset.txt)

### DIFF
--- a/test/tests/functional/pbs_reservations.py
+++ b/test/tests/functional/pbs_reservations.py
@@ -627,17 +627,15 @@ class TestReservations(TestFunctional):
              'Resource_List.place': 'excl'}
         if self.mom.is_cpuset_mom():
             a['Resource_List.select'] = vnode_val
+
         j = Job(TEST_USER, attrs=a)
         jid = self.server.submit(j)
         self.server.expect(JOB, {'job_state': 'R'}, id=jid)
 
         # Submit a reservation that will start after the job starts running
-        a = {'Resource_List.select': '1:ncpus=1',
-             'Resource_List.place': 'excl',
-             'reserve_start': now + 360,
+        b = {'reserve_start': now + 360,
              'reserve_end': now + 3600}
-        if self.mom.is_cpuset_mom():
-            a['Resource_List.select'] = vnode_val
+        a.update(b)
 
         r1 = Reservation(TEST_USER, attrs=a)
         rid1 = self.server.submit(r1)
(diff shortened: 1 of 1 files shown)


shilpakodli:
Just for confirmation, could you please run these tests on non-cpuset machine as well and see.

test/tests/functional/pbs_reservations.py
```diff
@@ -630,6 +636,9 @@ def test_future_resv_conflicts_running_job(self):
              'Resource_List.place': 'excl',
```
prakashcv13: instead of a, use b. We will not need changes on line 639-641

test/tests/functional/pbs_reservations.py
```diff
@@ -1938,6 +1953,7 @@ def test_delete_idle_resv_job_standing(self):
                     'reserve_start': start_str}
         self.server.expect(RESV, exp_attr, id=rid, offset=idle_timer + 5)
 
+    @skipOnCpuSet
```
prakashcv13: As I understand the purpose of job j1 is to keep the complex busy so that j2 and j3 do not run.
We can set scheduling=False in the beginning of the test case and then submit two jobs.
Convert them to asap reservations. Would that work?

dinesh1306: @prakashcv13 , I tried your suggestion, but it wont work.

dinesh1306:
@prakashcv13 , I have incorporate review comments. Please take a look.
[res_after_review_comments_on_cpusetmom.txt](https://github.com/PBSPro/pbspro/files/4324189/res_after_review_comments_on_cpusetmom.txt)
[res_after_review_comments_on_normal_machine.txt](https://github.com/PBSPro/pbspro/files/4324190/res_after_review_comments_on_normal_machine.txt)

test/tests/functional/pbs_reservations.py
```diff
@@ -616,20 +616,27 @@ def test_future_resv_conflicts_running_job(self):
...
         jid = self.server.submit(j)
         self.server.expect(JOB, {'job_state': 'R'}, id=jid)
 
         # Submit a reservation that will start after the job starts running
-        a = {'Resource_List.select': '1:ncpus=1',
-             'Resource_List.place': 'excl',
-             'reserve_start': now + 360,
+        b = {'reserve_start': now + 360,
              'reserve_end': now + 3600}
+        a.update(b)
```
prakashcv13: we can directly add reserve_start/end to a like we add Resource_List.select on line 629.

dinesh1306:
@prakashcv13 , i have change the code. Please review.
[res_after_review_on_cpusetmom.txt](https://github.com/PBSPro/pbspro/files/4324774/res_after_review_on_cpusetmom.txt)
[res_after_review_on_normal_machine.txt](https://github.com/PBSPro/pbspro/files/4324775/res_after_review_on_normal_machine.txt)

test/tests/functional/pbs_reservations.py
bhroam: I wouldn't expect this to work because the scheduler needs to run to confirm the ASAP reservations.  If scheduling is off, then it doesn't run.

prakashcv13: and that is why it didn't work :)

PR #1579, opened 2020-03-13
qsub crashes if environment variables contain backslashes

nagavg:
#### Issue-ID

- PBS-23201

#### Describe Bug or Feature

- qsub crashes if environment variables contain backslashes 

#### Describe Your Change

- qsub crashes if environment variables contain backslashes because it does not have enough memory allocated for the exported environmental variable. 

- Earlier, when we copy an environment variable value,  "copy_env_value" was adding extra special characters. we were not accommodating for those characters. With this code change we are allocating memory for extra characters.

[Before_fix.txt](https://github.com/PBSPro/pbspro/files/4328485/Before_fix.txt)
[After_fix.txt](https://github.com/PBSPro/pbspro/files/4328486/After_fix.txt)

[ptl_test_log.txt](https://github.com/PBSPro/pbspro/files/4328494/ptl_test_log.txt)


nagavg:
Force-pushing to resolve conflicts.

PR #1582, opened 2020-03-17
Making log_record async-safe

agrawalravi90:
#### Describe Bug or Feature

log_record is called, either directly or indirectly via log_event, from various signal handlers in various PBS daemons. log_record is not currently async-safe as it calls system calls like localtime_r() which are not async-safe. So, this PR is to make log_record async-safe by blocking all signals.

#### Describe Your Change

Added code to block all signals inside log_record()

There's no good test for this, this solves a slim race condition, so code inspection might be the only way to analyze the changes + smoke tests to ensure that nothing breaks.

### DIFF
--- a/src/lib/Liblog/pbs_log.c
+++ b/src/lib/Liblog/pbs_log.c
@@ -960,6 +960,8 @@ log_record(int eventtype, int objclass, int sev, const char *objname, const char
 sigunblock:
 #ifndef WIN32
 	sigprocmask(SIG_SETMASK, &old_mask, NULL);
+#else
+	return;
 #endif
 }
 
--- a/src/lib/Liblog/pbs_log.c
+++ b/src/lib/Liblog/pbs_log.c
@@ -893,15 +893,15 @@ log_record(int eventtype, int objclass, int sev, const char *objname, const char
 			snprintf(microsec_buf, sizeof(microsec_buf), ".%06ld", (long)tp.tv_usec);
 	}
 
-	/* lock the log mutex */
-	if (log_mutex_lock() != 0)
-		goto sigunblock;
-
 #ifdef WIN32
 	ptm = localtime(&now);
 #else
 	ptm = localtime_r(&now, &ltm);
 #endif
+
+	/* lock the log mutex */
+	if (log_mutex_lock() != 0)
+		goto sigunblock;
 
 	/* Do we need to switch the log? */
 	if (log_auto_switch && (ptm->tm_yday != log_open_day)) {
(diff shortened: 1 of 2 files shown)


subhasisb:
Thanks, this was long pending!

src/lib/Liblog/pbs_log.c
```diff
@@ -947,6 +956,13 @@ log_record(int eventtype, int objclass, int sev, const char *objname, const char
 			fclose(logfile);
 		}
 	}
+
+sigunblock:
+#ifndef WIN32
+	sigprocmask(SIG_SETMASK, &old_mask, NULL);
+#else
+	return;
```
mike0042: Shouldn't have to call return as the last line of a void function. Otherwise, looks good.

agrawalravi90:
@mike0042 looks like Windows doesn't like an empty label, I'll need to revert the last commit

src/lib/Liblog/pbs_log.c
```diff
@@ -884,16 +893,16 @@ log_record(int eventtype, int objclass, int sev, const char *objname, const char
 			snprintf(microsec_buf, sizeof(microsec_buf), ".%06ld", (long)tp.tv_usec);
 	}
 
+	/* lock the log mutex */
```
subhasisb: moving the mutex log before the call to localtime actually is not required. It will not make anything any better. If there is a fork that happens and takes the localtime_r's internal mutex to deadlock, this will not prevent it. The original place for the mutex is actually fine, i.e. it only guards against the switching of the log files

src/lib/Liblog/pbs_log.c
```diff
@@ -908,7 +917,7 @@ log_record(int eventtype, int objclass, int sev, const char *objname, const char
 			log_err(rc, "log_record", "PBS cannot open its log");
 			fclose(logfile);
 		}
-		return;
+		goto sigunblock;
```
subhasisb: This jump will be bad, if it happens, will leave the mutex locked given where you have it now

agrawalravi90: We unlock the mutex at the beginning of this block (see line 913), so I think it should be fine.

agrawalravi90:
@subhasisb I've addressed your comment, please let me know if it looks okay now. Also, any idea why where are 2 Travis builds on this PR? @hirenvadalia do you know?

hirenvadalia:
@agrawalravi90 Looks like some intermediate network issue, it's passing now. If it fails again let me know and don't restart it for debugging...

PR #1585, opened 2020-03-18
TestPbsNodeRampDown Tests failing because of jobs not in R state soon after server restart

visheshh:
#### Describe Bug or Feature
Soon after server restart nodes are not yet available so job will be in Q state for sometime. Tests are checking assigned resources and R state of the job soon after server restart as assigned resources are not there test fails it does not retrigger sched cycle.

#### Describe Your Change
Check the job for R state first which retriggers scheduling cycle and waits for jobs to run and then check for assigned resources.

[shared_restart.txt](https://github.com/PBSPro/pbspro/files/4347058/shared_restart.txt)
[excl_restart.txt](https://github.com/PBSPro/pbspro/files/4347063/excl_restart.txt)
[pbs_node_ramp_down_19.2.6_logs.txt](https://github.com/PBSPro/pbspro/files/4347111/pbs_node_ramp_down_19.2.6_logs.txt)

### DIFF
--- a/test/tests/functional/pbs_node_rampdown.py
+++ b/test/tests/functional/pbs_node_rampdown.py
@@ -6040,10 +6040,11 @@ pbs.logjobmsg(pbs.event().job.id, "epilogue hook executed")
         self.server.start()
         self.assertTrue(self.server.isUp())
 
-        # make sure job is now running with assigned resources
+        # make sure job is now running after server restart
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid)
+        # make sure job is running with assigned resources
         # from the original request
-        self.server.expect(JOB, {'job_state': 'R',
-                                 'Resource_List.mem': '5gb',
+        self.server.expect(JOB, {'Resource_List.mem': '5gb',
                                  'Resource_List.ncpus': 7,
                                  'Resource_List.nodect': 3,
                                  'Resource_List.select': self.job11x_select,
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_node_rampdown.py
```diff
@@ -6040,10 +6040,11 @@ def test_release_nodes_excl_server_restart_immed(self):
...
         self.assertTrue(self.server.isUp())
 
-        # make sure job is now running with assigned resources
+        # make sure job is now running after server restart
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid)
+        # make sure job is running with assigned resources
         # from the original request
-        self.server.expect(JOB, {'job_state': 'R',
-                                 'Resource_List.mem': '5gb',
+        self.server.expect(JOB, {'Resource_List.mem': '5gb',
```
bhroam: I'm a little confused why this change is required.  I thought that expect() will trigger scheduling cycles unless you tell it not to with saying trigger_sched_cycle=False.

visheshh: Hi @bhroam , The cycle will be triggered if the data does not match, But here the data for exec_host is not available as the job is in Q state. So the test fails.

bhroam: That sounds like a bug in expect() to me.  If I understand what you are saying is that if exec_host was set but didn't match, it'd trigger a cycle.  Since it isn't set, it just fails.  Could you file a bug against expect?  I'll sign off on your workaround here.

PR #1587, opened 2020-03-18
customize TPP mcast packet for target comms

subhasisb:
#### Describe Bug or Feature

In typical large setups, it is often the case that leafs (recipients) are attached to different comm routers.
A sender leaf sends an MCAST packet to the comm router to which it is connected to, and that router forwards the packet, unchanged, to several target routers. The fact that it sends the packet without customizing (or changing) it to the downstream comm routers, poses the following issues:

1. The downstream routers often have only a small subset of the total list of leafs attached to them (that they manage). However, since the MCAST packet contains target leaf entries that do not belong to this downstream comm router, the router wastes time in processing those entries by making avltree lookups  (which is quite fast, but nevertheless wasteful) only to find that the entry is not a leaf attached to itself and thus discards that entry.

2. Some sites do not connect all the comm routers to each other. They rather like to maintain islands of leafs (moms/nodes) that are isolated in groups. Thus the comm router in one island has no knowledge of leafs attached to a different comm router in another node island. Thus when the above MCAST packet arrives, the comm router fails to find some leafs - and without knowing whether it is a deliberate configuration or a network error (that is a leaf actually went down), the only thing it can do, safely, is to reject the whole MCAST packet - which results in the job aborting at startup.

#### Describe Your Change

The proposed fix customizes the MCAST packet before it forwards it to downstream comm routers. Specifically, it only fills in the entries for leafs that are connected to a particular downstream comm router. This way, both the above listed issues are mitigated.

Test steps and results (Thanks to @saritakh ) : 
[tpp_fix_tests.txt](https://github.com/PBSPro/pbspro/files/4353620/tpp_fix_tests.txt)
[tpp_fix_valgrind.txt](https://github.com/PBSPro/pbspro/files/4353824/tpp_fix_valgrind.txt)

### DIFF
--- a/src/lib/Libtpp/tpp_client.c
+++ b/src/lib/Libtpp/tpp_client.c
@@ -3656,7 +3656,7 @@ add_part_packet(stream_t *strm, void *data, int sz)
 	pkt = strm->part_recv_pkt;
 	TPP_DBPRT(("*** pkt=%p, sd=%u, sz=%d, totlen=%d, cmprsd_len=%u", (void *) pkt, strm->sd, sz, totlen, cmprsd_len));
 	if (!pkt) {
-		/* some rare cases, compresses size can be larger than orig size, hence use the larger value */
+		/* some rare cases, compressed size can be larger than orig size, hence use the larger value */
 		pkt = tpp_cr_pkt(NULL, totlen > sz ? totlen : sz, 1);
 		if (!pkt)
 			return NULL;
--- a/src/lib/Libtpp/tpp_client.c
+++ b/src/lib/Libtpp/tpp_client.c
@@ -1223,7 +1223,7 @@ tpp_send(int sd, void *data, int len)
 
 	TPP_DBPRT(("Sending: sd=%d, len=%d", sd, len));
 
-	if ((tpp_conf->compress == 1) && (len > TPP_SEND_SIZE)) {
+	if ((tpp_conf->compress == 1) && (len > TPP_COMPR_SIZE)) {
 		void *outbuf;
 
 		outbuf = tpp_deflate(data, len, &cmprsd_len);
@@ -2326,7 +2326,7 @@ tpp_mcast_send(int mtfd, void *data, unsigned int len, unsigned int full_len, un
 	chunks[0].len = sizeof(tpp_mcast_pkt_hdr_t);
 	totlen = chunks[0].len;
 
-	if (tpp_conf->compress == 1 && minfo_len > TPP_SEND_SIZE) {
+	if (tpp_conf->compress == 1 && minfo_len > TPP_COMPR_SIZE) {
 		def_ctx = tpp_multi_deflate_init(minfo_len);
 		if (def_ctx == NULL)
 			goto err;
@@ -3656,7 +3656,8 @@ add_part_packet(stream_t *strm, void *data, int sz)
 	pkt = strm->part_recv_pkt;
 	TPP_DBPRT(("*** pkt=%p, sd=%u, sz=%d, totlen=%d, cmprsd_len=%u", (void *) pkt, strm->sd, sz, totlen, cmprsd_len));
 	if (!pkt) {
-		pkt = tpp_cr_pkt(NULL, totlen, 1);
+		/* some rare cases, compresses size can be larger than orig size, hence use the larger value */
+		pkt = tpp_cr_pkt(NULL, totlen > sz ? totlen : sz, 1);
 		if (!pkt)
 			return NULL;
 		TPP_DBPRT(("Total length = %d, sz=%d", totlen, sz));
--- a/src/lib/Libtpp/tpp_common.h
+++ b/src/lib/Libtpp/tpp_common.h
@@ -273,7 +273,7 @@ enum TPP_MSG_TYPES {
 #define TPP_STRM_TIMEOUT        600
 #define TPP_MIN_WAIT            2
 #define TPP_SEND_SIZE           8192
-#define TPP_COMPR_SIZE			8192
+#define TPP_COMPR_SIZE          8192
 
 /* tpp cmds used internally by the layer to notify messages between threads */
 #define TPP_CMD_SEND            1
--- a/src/lib/Libtpp/tpp_common.h
+++ b/src/lib/Libtpp/tpp_common.h
@@ -273,6 +273,7 @@ enum TPP_MSG_TYPES {
 #define TPP_STRM_TIMEOUT        600
 #define TPP_MIN_WAIT            2
 #define TPP_SEND_SIZE           8192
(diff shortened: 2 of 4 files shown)


subhasisb:
image

subhasisb:
Test logs and valgrind output attached. Valgrind output from the same tests of isolated islands - valgrind attached to all comms. TH3 regression output attached.

src/lib/Libtpp/tpp_common.h
```diff
@@ -273,6 +273,7 @@ enum TPP_MSG_TYPES {
 #define TPP_STRM_TIMEOUT        600
 #define TPP_MIN_WAIT            2
 #define TPP_SEND_SIZE           8192
+#define TPP_COMPR_SIZE			8192
```
prakashcv13: is the value of TPP_COMPR_SIZE properly indented?
TPP_SEND_SIZE and TPP_COMPR_SIZE - have the same value.
How is using TPP_COMPR_SIZE instead of TPP_SEND_SIZE helping?

src/lib/Libtpp/tpp_router.c
```diff
@@ -2064,7 +2072,7 @@ router_pkt_handler(int tfd, void *data, int len, void *c, void *extra)
 					memcpy(&shdr.dest_addr, &minfo->dest_addr, sizeof(tpp_addr_t));
 
 					TPP_DBPRT(("Send mcast indiv packet to %s", tpp_netaddr(&shdr.dest_addr)));
-
+
```
prakashcv13: let us remove the tabs on this blank line.

src/lib/Libtpp/tpp_router.c
```diff
@@ -1949,10 +1949,19 @@ router_pkt_handler(int tfd, void *data, int len, void *c, void *extra)
 		case TPP_MCAST_DATA: {
 			int i, k;
 			tpp_addr_t *src_host;
-			int *rlist = NULL;
+			typedef struct {
+				char *router_name;
+				int target_fd; /* target comm fd */
+				void *cmpr_ctx;
+				void *minfo_buf; /* allocate size for total members */
+				int num_streams; /* actual number of destination streams */
```
prakashcv13: shall we have int(s) after all the pointers or vice versa?

src/lib/Libtpp/tpp_router.c
```diff
@@ -2077,74 +2085,131 @@ router_pkt_handler(int tfd, void *data, int len, void *c, void *extra)
...
-								(unsigned long)(sizeof(int) * rsize));
-							tpp_log_func(LOG_CRIT, __func__, tpp_get_logbuf());
-							if (data_out)
-								free(data_out);
-							return -1;
-						}
-						csize = 0;
-					}
-
+					/* add this to list of routers to whom we need to send */
```
prakashcv13: shall we combine these two comments?

src/lib/Libtpp/tpp_router.c
```diff
@@ -2077,74 +2085,131 @@ router_pkt_handler(int tfd, void *data, int len, void *c, void *extra)
...
 					/**
-					 * now check list backwards if router already sent to
+					 * now check list backwards if router already added to
 					 * rational for checking backwards is that the last router
```
prakashcv13: rationale?

src/lib/Libtpp/tpp_router.c
```diff
@@ -2077,74 +2085,131 @@ router_pkt_handler(int tfd, void *data, int len, void *c, void *extra)
...
-								free(data_out);
-							return -1;
-						}
-						csize = 0;
-					}
-
+					/* add this to list of routers to whom we need to send */
 					/**
-					 * now check list backwards if router already sent to
+					 * now check list backwards if router already added to
```
prakashcv13: would saying
"now check list backwards if router is already added." make more sense?
this line and the next line do not seem to be part of one sentence.

src/lib/Libtpp/tpp_client.c
```diff
@@ -3656,7 +3656,8 @@ add_part_packet(stream_t *strm, void *data, int sz)
 	pkt = strm->part_recv_pkt;
 	TPP_DBPRT(("*** pkt=%p, sd=%u, sz=%d, totlen=%d, cmprsd_len=%u", (void *) pkt, strm->sd, sz, totlen, cmprsd_len));
 	if (!pkt) {
-		pkt = tpp_cr_pkt(NULL, totlen, 1);
+		/* some rare cases, compresses size can be larger than orig size, hence use the larger value */
```
prakashcv13: compressed size?

src/lib/Libtpp/tpp_router.c
```diff
@@ -1949,10 +1949,19 @@ router_pkt_handler(int tfd, void *data, int len, void *c, void *extra)
...
+				void *cmpr_ctx;
+				void *minfo_buf; /* allocate size for total members */
+				int num_streams; /* actual number of destination streams */ 
+			} target_comm_struct_t;
+
+			target_comm_struct_t *rlist = NULL;
 			int rsize = 0;
 			int csize = 0;
 			void *tmp;
+			tpp_chunk_t mchunks[3];
```
prakashcv13: shall we add a comment to explain why are we using exactly 3 elements?

src/lib/Libtpp/tpp_router.c
subhasisb: if it makes it look slightly nicer, sure

src/lib/Libtpp/tpp_router.c
subhasisb: ACtually they are better in two different blocks, since they are telling different things

src/lib/Libtpp/tpp_router.c
subhasisb: yes thanks

src/lib/Libtpp/tpp_router.c
subhasisb: changed

src/lib/Libtpp/tpp_client.c
subhasisb: yes

src/lib/Libtpp/tpp_router.c
subhasisb: done

src/lib/Libtpp/tpp_common.h
subhasisb: fixed the indentation.

TPP_SEND_SIZE was being overloaded. Used for deciding the packet size, beyond which we will break down packets into multiple. Using the same value for the compression threshold meant we could not set it to a smaller value than the packet size. For example, we might want to send 16k long packets each, but compress if the packet size > 8k. Two separate variables allow this - and makes for easy testing

PR #1590, opened 2020-03-18
changes affecting revert_pbsconf()

vccardenas:
#### Describe Bug or Feature
revert_pbsconf() does not revert to correct ownership of PBS config file

#### Describe Your Changes
- Check if config file is owned by root, otherwise change ownership to root.
- chown() now uses the form "chown uid:gid <file>" instead of "chown uid <file>"

Screen Shot 2020-03-18 at 10 11 55 AM

### DIFF
--- a/test/fw/ptl/utils/pbs_dshutils.py
+++ b/test/fw/ptl/utils/pbs_dshutils.py
@@ -1206,6 +1206,9 @@ class DshUtils(object):
                            recursive=recursive, runas=runas)
             if ((uid is not None and uid != self.get_current_user()) or
                     gid is not None):
+                if dest == self.get_pbs_conf_file(targethost):
+                    uid = pwd.getpwnam('root')[2]
+                    gid = pwd.getpwnam('root')[3]
                 self.chown(targethost, path=dest, uid=uid, gid=gid, sudo=True,
                            recursive=False)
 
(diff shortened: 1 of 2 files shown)


test/fw/ptl/utils/pbs_dshutils.py
```diff
@@ -1206,6 +1206,9 @@ def run_copy(self, hosts=None, src=None, dest=None, sudo=False, uid=None,
                            recursive=recursive, runas=runas)
             if ((uid is not None and uid != self.get_current_user()) or
                     gid is not None):
+                if dest == self.get_pbs_conf_file(targethost):
+                    uid = pwd.getpwnam('root')[2]
+                    gid = pwd.getpwnam('root')[3]
```
bhroam: Should we use 'root' or str(ROOT_USER)?

vccardenas: We should use 'root' since it is used in the pw database of most *nix systems.

bhroam: Yeah, but we've abstracted it out in our PTL interface.  Why not use our interface?

vccardenas: root is root and I didn't see the need to abstract it.

vccardenas: I also tested the change and got this:
    uid = pwd.getpwnam(str(ROOT_USER))[2]
NameError: name 'ROOT_USER' is not defined

bhroam: Thanks for testing it.  pbs_testusers must not be imported.  Yeah, just use root.

vccardenas: Yes I'll need to do add ROOT_USER in the line below:
from ptl.utils.pbs_testusers import PBS_ALL_USERS, PbsUser, ROOT_USER
Ok I'll keep 'root'.

PR #1596, opened 2020-03-23
Adding new option -D to ralter

tpai26:
#### Describe Bug or Feature
Currently we can change the start time and the end time of the reservation. With this new option '-D' the user has the option to change the duration of the reservation.

#### Link to Design Doc
https://pbspro.atlassian.net/wiki/spaces/PD/pages/1551400989/New+option+-D+in+pbs+ralter+to+change+duration+of+a+reservation

[ralter_fix9.txt](https://github.com/PBSPro/pbspro/files/4430458/ralter_fix9.txt)

### DIFF
--- a/src/cmds/pbs_ralter.c
+++ b/src/cmds/pbs_ralter.c
@@ -65,14 +65,17 @@ int
 process_opts(int argc, char **argv, struct attrl **attrp, char *dest)
 {
 	int	c = 0;
+	
 	int	errflg = 0;
 	time_t	t;
 
 	char	time_buf[80] = {0};
 	char	*endptr = NULL;
 	long	temp = 0;
+	char dur_buf[800];
+	int i = 0;
 
-	while ((c = getopt(argc, argv, "E:I:m:M:N:R:q:U:G:")) != EOF) {
+	while ((c = getopt(argc, argv, "E:I:m:M:N:R:q:U:G:D:")) != EOF) {
 		switch (c) {
 			case 'E':
 				t = cvtdate(optarg);
@@ -137,6 +140,14 @@ process_opts(int argc, char **argv, struct attrl **attrp, char *dest)
 				set_attr_error_exit(&attrib, ATTR_auth_g, optarg);
 				break;
 
+			case 'D':
+				sprintf(dur_buf, "%s", optarg);
+				i = set_attr(&attrib, ATTR_resv_duration, dur_buf);
+				if (i != 0) {
+					fprintf(stderr, "pbs_ralter: illegal -D time value\n");
+					errflg++;
+				}
+				break;				
 			default:
 				/* pbs_ralter option not recognized */
 				errflg++;
@@ -161,7 +172,7 @@ print_usage()
 	"                [-N reservation_name] [-R start_time] [-E end_time]\n"
 	"                [-U (+/-)username[,(+/-)username]...]\n"
 	"                [-G [(+/-)group[,(+/-)group]...]]\n"
-	"                resv_id\n";
+	"                [-D durationn] resv_id\n";
 	fprintf(stderr, "%s", usage);
 	fprintf(stderr, "%s", usag2);
 }
@@ -201,6 +212,8 @@ handle_attribute_errors(struct ecl_attribute_errors *err_list)
 			opt = "R";
 		else if (strcmp(attribute->name, ATTR_auth_u) == 0)
 			opt = "U";
+		else if (strcmp(attribute->name, ATTR_resv_duration) == 0)
+			opt = "D";
 		else
 			return ;
 
--- a/src/cmds/pbs_ralter.c
+++ b/src/cmds/pbs_ralter.c
@@ -73,7 +73,6 @@ process_opts(int argc, char **argv, struct attrl **attrp, char *dest)
 	char	*endptr = NULL;
 	long	temp = 0;
 	char dur_buf[800];
-	int i = 0;
 
 	while ((c = getopt(argc, argv, "E:I:m:M:N:R:q:U:G:D:")) != EOF) {
 		switch (c) {
@@ -141,12 +140,8 @@ process_opts(int argc, char **argv, struct attrl **attrp, char *dest)
 				break;
 
 			case 'D':
-				sprintf(dur_buf, "%s", optarg);
-				i = set_attr(&attrib, ATTR_resv_duration, dur_buf);
-				if (i != 0) {
-					fprintf(stderr, "pbs_ralter: illegal -D time value\n");
-					errflg++;
-				}
+				snprintf(dur_buf, sizeof(dur_buf), "%s", optarg);
+				set_attr_error_exit(&attrib, ATTR_resv_duration, dur_buf);
 				break;				
 			default:
 				/* pbs_ralter option not recognized */
@@ -172,7 +167,8 @@ print_usage()
 	"                [-N reservation_name] [-R start_time] [-E end_time]\n"
 	"                [-U (+/-)username[,(+/-)username]...]\n"
 	"                [-G [(+/-)group[,(+/-)group]...]]\n"
-	"                [-D durationn] resv_id\n";
+	"                [-D duration]\n"
+	"                resv_id\n";
 	fprintf(stderr, "%s", usage);
 	fprintf(stderr, "%s", usag2);
 }
@@ -196,7 +192,9 @@ handle_attribute_errors(struct ecl_attribute_errors *err_list)
 
 	for (i = 0; i < err_list->ecl_numerrors; i++) {
 		attribute = err_list->ecl_attrerr[i].ecl_attribute;
-		if (strcmp(attribute->name, ATTR_resv_end) == 0)
+		if (strcmp(attribute->name, ATTR_resv_duration) == 0)
+			opt = "D";
+		else if (strcmp(attribute->name, ATTR_resv_end) == 0)
 			opt = "E";
 		else if (strcmp(attribute->name, ATTR_auth_g) == 0)
 			opt = "G";
@@ -212,8 +210,6 @@ handle_attribute_errors(struct ecl_attribute_errors *err_list)
 			opt = "R";
 		else if (strcmp(attribute->name, ATTR_auth_u) == 0)
 			opt = "U";
-		else if (strcmp(attribute->name, ATTR_resv_duration) == 0)
-			opt = "D";
 		else
 			return ;
 
--- a/src/include/reservation.h
+++ b/src/include/reservation.h
@@ -63,7 +63,7 @@ extern "C" {
 
 #define RESV_START_TIME_MODIFIED	0x1
 #define RESV_END_TIME_MODIFIED		0x2
-#define RESV_DURATION_MODIFIED		0x3
+#define RESV_DURATION_MODIFIED		0x4
 
 
 /*
--- a/src/include/reservation.h
+++ b/src/include/reservation.h
@@ -63,6 +63,8 @@ extern "C" {
 
 #define RESV_START_TIME_MODIFIED	0x1
 #define RESV_END_TIME_MODIFIED		0x2
+#define RESV_DURATION_MODIFIED		0x3
+
 
 /*
  * The following resv_atr enum provide an index into the array of
--- a/src/server/req_modify.c
+++ b/src/server/req_modify.c
@@ -789,6 +789,7 @@ req_modifyReservation(struct batch_request *preq)
 	int r = 0;
 	int e = 0;
 
+
 	while (psatl) {
 		long temp = 0;
 		char *end = NULL;
@@ -850,7 +851,6 @@ req_modifyReservation(struct batch_request *preq)
 				}
 				
 				if (d == 1) {
-					presv->ri_alter_etime = presv->ri_wattr[RESV_ATR_end].at_val.at_long;
 					presv->ri_wattr[RESV_ATR_end].at_flags &= ~ATR_VFLAG_SET;
 				}
 
@@ -870,21 +870,25 @@ req_modifyReservation(struct batch_request *preq)
 				}
 				e = 1;
 				if (d == 1) {
-					presv->ri_alter_stime = presv->ri_wattr[RESV_ATR_start].at_val.at_long;
 					presv->ri_wattr[RESV_ATR_start].at_flags &= ~ATR_VFLAG_SET;
 				}
 
 				break;
 			case RESV_ATR_duration:
 				if (r == 1) {
-					presv->ri_alter_etime = presv->ri_wattr[RESV_ATR_end].at_val.at_long;
+					if (presv->ri_alter_etime == 0)
+						presv->ri_alter_etime = presv->ri_wattr[RESV_ATR_end].at_val.at_long;
+						
 					presv->ri_wattr[RESV_ATR_end].at_flags &= ~ATR_VFLAG_SET;
 				} else if (e == 1) {
-					presv->ri_alter_stime = presv->ri_wattr[RESV_ATR_start].at_val.at_long;
+					if(presv->ri_alter_stime == 0)
+						presv->ri_alter_stime = presv->ri_wattr[RESV_ATR_start].at_val.at_long;
+
 					presv->ri_wattr[RESV_ATR_start].at_flags &= ~ATR_VFLAG_SET;
 				}
 				else {
-					presv->ri_alter_etime = presv->ri_wattr[RESV_ATR_end].at_val.at_long;
+					if (presv->ri_alter_etime == 0)
+						presv->ri_alter_etime = presv->ri_wattr[RESV_ATR_end].at_val.at_long;
 					presv->ri_wattr[RESV_ATR_end].at_flags &= ~ATR_VFLAG_SET;
 				}
 				d = 1;
@@ -895,10 +899,12 @@ req_modifyReservation(struct batch_request *preq)
 				break;
 		}
 
-		if (d == 0) 
+		if (d == 0)
 			presv->ri_wattr[RESV_ATR_duration].at_flags &= ~ATR_VFLAG_SET;
 
-		if (d == 1 && r == 1 && e == 1) {
+		if (d == 1 && e == 1 && r == 1) {
+			presv->ri_alter_state = presv->ri_wattr[RESV_ATR_state].at_val.at_long;
+			resv_setResvState(presv, RESV_BEING_ALTERED, presv->ri_qs.ri_substate);
 			req_reject(PBSE_BADTSPEC, 0, preq);
 			resv_revert_alter_times(presv);
 			return;
--- a/src/server/req_modify.c
+++ b/src/server/req_modify.c
@@ -741,6 +741,7 @@ req_modifyReservation(struct batch_request *preq)
 	extern char	*msg_stdg_resv_occr_conflict;
 	resc_resv	*presv;
 	int num_jobs;
+	long temp_dur = 0;
 
 	if (preq == NULL)
 		return;
@@ -861,24 +862,7 @@ req_modifyReservation(struct batch_request *preq)
 				}
 
 				break;
-			case RESV_ATR_duration:
-				if (presv->ri_alter_etime == 0) {
-					temp = strtol(psatl->al_value, &end, 10) + presv->ri_wattr[RESV_ATR_end].at_val.at_long;
-					if (!is_standing || temp < next_occr_start) {
-						presv->ri_alter_flags |= RESV_START_TIME_MODIFIED;
-						presv->ri_alter_etime = presv->ri_wattr[RESV_ATR_end].at_val.at_long;
-					}
-					else {
-						snprintf(log_buffer, sizeof(log_buffer), "%s", msg_stdg_resv_occr_conflict);
-						log_event(PBSEVENT_RESV, PBS_EVENTCLASS_RESV, LOG_INFO,
-							preq->rq_ind.rq_modify.rq_objname, log_buffer);
-						req_reject(PBSE_STDG_RESV_OCCR_CONFLICT, 0, preq);
-						return;
-					}
-				}
-				if ((!is_standing || temp < next_occr_start) && presv->ri_alter_stime == 0)
-					presv->ri_alter_stime = presv->ri_wattr[RESV_ATR_start].at_val.at_long;			
-				
+			case RESV_ATR_duration:							
 				send_to_scheduler = RESV_DURATION_MODIFIED;
 				presv->ri_alter_flags |= RESV_DURATION_MODIFIED;
 				break;
@@ -898,6 +882,34 @@ req_modifyReservation(struct batch_request *preq)
 		psatl = (svrattrl *)GET_NEXT(psatl->al_link);
 	}
 	resc_access_perm = resc_access_perm_save; /* restore perm */
+
+	temp_dur = presv->ri_wattr[RESV_ATR_duration].at_val.at_long + presv->ri_wattr[RESV_ATR_end].at_val.at_long;
+	
+	if ((presv->ri_alter_flags & RESV_DURATION_MODIFIED) && presv->ri_alter_etime == 0) {
+		if (!is_standing || temp_dur < next_occr_start) {
+			presv->ri_alter_flags |= RESV_START_TIME_MODIFIED;
+			presv->ri_alter_etime = presv->ri_wattr[RESV_ATR_end].at_val.at_long;
+		} else {
+			log_event(PBSEVENT_RESV, PBS_EVENTCLASS_RESV, LOG_INFO,
+				preq->rq_ind.rq_modify.rq_objname, msg_stdg_resv_occr_conflict);
+			req_reject(PBSE_STDG_RESV_OCCR_CONFLICT, 0, preq);
+			resv_revert_alter_times(presv);
+			return;
+		}
+	}
+	
+	if ((presv->ri_alter_flags & RESV_DURATION_MODIFIED) && presv->ri_alter_stime == 0) {
+		if (!is_standing || temp_dur < next_occr_start) {
+			presv->ri_alter_stime = presv->ri_wattr[RESV_ATR_start].at_val.at_long;
+		} else {
+				log_event(PBSEVENT_RESV, PBS_EVENTCLASS_RESV, LOG_INFO,
+					preq->rq_ind.rq_modify.rq_objname, msg_stdg_resv_occr_conflict);
+				req_reject(PBSE_STDG_RESV_OCCR_CONFLICT, 0, preq);
+				resv_revert_alter_times(presv);
+				return;
+		}
+	}
+	
 
 	if (send_to_scheduler) {
 		presv->ri_alter_state = presv->ri_wattr[RESV_ATR_state].at_val.at_long;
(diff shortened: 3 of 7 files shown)


tpai26:
updated test logs

src/cmds/pbs_ralter.c
```diff
@@ -137,6 +140,14 @@ process_opts(int argc, char **argv, struct attrl **attrp, char *dest)
 				set_attr_error_exit(&attrib, ATTR_auth_g, optarg);
 				break;
 
+			case 'D':
+				sprintf(dur_buf, "%s", optarg);
+				i = set_attr(&attrib, ATTR_resv_duration, dur_buf);
```
vstumpf: Can you not just do set_attr_error_exit?

src/cmds/pbs_ralter.c
```diff
@@ -161,7 +172,7 @@ print_usage()
 	"                [-N reservation_name] [-R start_time] [-E end_time]\n"
 	"                [-U (+/-)username[,(+/-)username]...]\n"
 	"                [-G [(+/-)group[,(+/-)group]...]]\n"
-	"                resv_id\n";
+	"                [-D durationn] resv_id\n";
```
vstumpf: I would keep resv_id on its own line.

src/cmds/pbs_ralter.c
```diff
@@ -201,6 +212,8 @@ handle_attribute_errors(struct ecl_attribute_errors *err_list)
 			opt = "R";
 		else if (strcmp(attribute->name, ATTR_auth_u) == 0)
 			opt = "U";
+		else if (strcmp(attribute->name, ATTR_resv_duration) == 0)
```
vstumpf: The rest of the options are in alphabetical order, can you make it so?

src/include/reservation.h
```diff
@@ -63,6 +63,8 @@ extern "C" {
 
 #define RESV_START_TIME_MODIFIED	0x1
 #define RESV_END_TIME_MODIFIED		0x2
+#define RESV_DURATION_MODIFIED		0x3
```
vstumpf: Where does this get used? And are you sure you can use 3 or is it a bit field? Might need to change it to 0x4.

test/tests/functional/pbs_ralter.py
```diff
@@ -1516,3 +1550,254 @@ def test_failed_ralter(self):
...
+        self.assertEqual(t_duration, 40)
+
+        # Submit a job to the reservation and change its start time.
+        self.submit_job_to_resv(rid)
+        temp_end = t_end
+        temp_start = t_start
+
+        self.alter_a_reservation(rid, temp_start, temp_end, shift=shift,
+                                 alter_e=True, sequence=2,
+                                 a_duration=50, check_log=False)
```
bhroam: Could you modify alter_a_reservation()?  It takes options that are ignored and not needed.  Here you are passing the start and end time, but end time is being altered.  It isn't necessary.  Maybe take default values for start, end, shift, and duration and check which ones are required based on alter_s and alter_e.

test/tests/functional/pbs_ralter.py
```diff
@@ -1516,3 +1550,254 @@ def test_failed_ralter(self):
...
+        t_duration, t_start, t_end = self.get_resv_time_info(rid)
+
+        self.assertEqual(t_end, end+20)
+        self.assertEqual(t_start, start+10)
+        self.assertEqual(t_duration, 30)
+
+        # Submit a job to the reservation and change its start time.
+        self.submit_job_to_resv(rid)
+        temp_end = t_end
+        temp_start = t_start
```
bhroam: Why do we care if there is a job in the reservation?  The only time it matters if the reservation is empty is if it is running and you are changing the start time.

test/tests/functional/pbs_ralter.py
```diff
@@ -1516,3 +1550,254 @@ def test_failed_ralter(self):
...
+        self.server.alterresv(rid, attr)
+        self.server.log_match(rid + ";Reservation alter denied",
+                              id=rid, interval=2)
+        rid = rid.split('.')[0]
+        self.server.log_match(rid + ";deleted at request of pbs_server",
+                              id=rid, interval=2)
+
+    def test_adv_resv_endtime_starttime_dur_together(self):
+        """
+        Test that all three end, start and duration can be changed together
```
bhroam: I think you mean cannot be changed together.

test/tests/functional/pbs_ralter.py
```diff
@@ -1516,3 +1550,254 @@ def test_failed_ralter(self):
...
+        new_end = self.bu.convert_seconds_to_datetime(end + shift)
+        new_duration = self.bu.convert_seconds_to_datetime(duration + 10)
+
+        try:
+            attr = {'reserve_start': new_start, 'reserve_end': new_end,
+                    'reserve_duration': new_duration}
+            self.server.alterresv(rid, attr)
+        except PbsResvAlterError as e:
+            self.assertTrue("pbs_ralter: Bad time specification(s)" in
+                            e.msg[0])
```
bhroam: @vstumpf has a better way of doing this.  It's with assertRaises().  Could you provide suggested code @vstumpf ?

test/tests/functional/pbs_ralter.py
```diff
@@ -1516,3 +1550,254 @@ def test_failed_ralter(self):
...
+
+        1. Change duration and endtime of standing reservation
+        2. After the first occurrence of the reservation finishes, verify that
+           the start and end time of the second occurrence have not changed.
+
+        All the above operations are expected to be successful.
+        """
+        offset = 20
+        duration = 20
+        shift = 15
```
bhroam: Consider adding the new duration up here.  This way you can use it in the alter call plus the assert().  Also use these as much as possible in the below asserts (here and other tests).

src/include/reservation.h
vstumpf: Yeah, this is a bitfield, it should be 0x4

src/server/req_modify.c
```diff
@@ -815,17 +819,19 @@ req_modifyReservation(struct batch_request *preq)
...
 					temp = strtol(psatl->al_value, &end, 10);
-					if ((temp > time(NULL)) &&
-						(temp != presv->ri_wattr[RESV_ATR_start].at_val.at_long)) {
+					if (temp > time(NULL)) {
 						if (!is_standing || (temp < next_occr_start)) {
 							send_to_scheduler = RESV_START_TIME_MODIFIED;
 							presv->ri_alter_stime = presv->ri_wattr[RESV_ATR_start].at_val.at_long;
 							presv->ri_alter_flags |= RESV_START_TIME_MODIFIED;
+							pdef->at_decode(&presv->ri_wattr[RESV_ATR_start], 
+							psatl->al_name, psatl->al_resc, psatl->al_value);
```
vstumpf: I dont think you should be doing this, if the scheduler rejects the alter, the start time will still change, right?

src/server/req_modify.c
```diff
@@ -844,13 +850,15 @@ req_modifyReservation(struct batch_request *preq)
 						req_reject(PBSE_BADTSPEC, 0, preq);
 					return;
 				}
+				
+				if (d==1) {
+					presv->ri_alter_etime = presv->ri_wattr[RESV_ATR_end].at_val.at_long;
+					presv->ri_wattr[RESV_ATR_end].at_flags &= ~ATR_VFLAG_SET;
```
vstumpf: Why are you unsetting the end of the reservation?

src/server/req_modify.c
```diff
@@ -862,11 +870,43 @@ req_modifyReservation(struct batch_request *preq)
 					req_reject(PBSE_STDG_RESV_OCCR_CONFLICT, 0, preq);
 					return;
 				}
+				e = 1;
+				if (d==1) {
+					presv->ri_alter_stime = presv->ri_wattr[RESV_ATR_start].at_val.at_long;
+					presv->ri_wattr[RESV_ATR_start].at_flags &= ~ATR_VFLAG_SET;
```
vstumpf: Again, why are you unsetting start?

test/tests/functional/pbs_ralter.py
vstumpf: Sure, it could be:
```Python
with self.assertRaises(PbsResvAlterError) as e:
    attr = {'reserve_start': new_start, 'reserve_end': new_end,
        'reserve_duration': new_duration}
    self.server.alterresv(rid, attr)
self.assertIn('pbs_ralter: Bad time specification(s)', e.msg[0])

Comment (vstumpf) [reply test/tests/functional/pbs_ralter.py]: You should also make sure that the reservation has the original start/end/duration

Comment (bhroam) [reply test/tests/functional/pbs_ralter.py]: I'd move the setting of attr outside of the with, but it really doesn't matter.

Comment (vstumpf) [reply test/tests/functional/pbs_ralter.py]: I'd keep it in the with, it has more context being with the alterresv, and he doesn't need the attrs later.

File: test/tests/functional/pbs_ralter.py (hunk 5 above)
Comment (tpai26) [reply test/tests/functional/pbs_ralter.py]: Left this as is. This function returns the new start and new end, it also to checks start time and end time in log files in all non duration tests. So changing this we will have to change some tests too or calculate new start / endtime in the function. So I think fixing this comment might be an overkill for making test code simpler in a couple of tests.

File: test/tests/functional/pbs_ralter.py (hunk 6 above)
Comment (tpai26) [reply test/tests/functional/pbs_ralter.py]: Just an extra testing to check reservation start time can be changed when there is job in reservation that is not running.

File: src/server/req_modify.c (hunk 11 above)
Comment (tpai26) [reply src/server/req_modify.c]: start_end_dur_wall() function calculates the value of field with unset flag for us, so the need to unset the flag

**Comment by tpai26** [comment]
updated logs after addressing comments

File: src/server/req_modify.c (hunk 10 above)
Comment (vstumpf) [reply src/server/req_modify.c]: Have you addressed this comment?

File: src/server/req_modify.c (hunk 11 above)
Comment (vstumpf) [reply src/server/req_modify.c]: But if this function fails, the end of the reservation is now unset forever.

File: test/tests/functional/pbs_ralter.py [hunk 13]
```diff
@@ -1516,3 +1548,265 @@ def test_failed_ralter(self):
...
+        rid, start, end = self.submit_and_confirm_reservation(offset, duration)
+        new_start = self.bu.convert_seconds_to_datetime(start + shift)
+        new_end = self.bu.convert_seconds_to_datetime(end + shift)
+        new_duration = self.bu.convert_seconds_to_datetime(new_duration)
+
+        with self.assertRaises(PbsResvAlterError) as e:
+            attr = {'reserve_start': new_start, 'reserve_end': new_end,
+                    'reserve_duration': new_duration}
+            self.server.alterresv(rid, attr)
+            self.assertIn('pbs_ralter: Bad time specification(s)', e.msg[0])
```
vstumpf: This block of code will end when `self.server.alterresv` throws the reservation, so this line needs to be outside the with block.

Also, make sure that the reservation still has the original start/end/dur.

src/server/req_modify.c
tpai26: For this I have updated the resv_revert_alter_times() function. In case the function fails resv_revert_alter_times() will reset the flags and also the starttime and endtime to the previous value

src/server/req_modify.c
```diff
@@ -895,10 +899,12 @@ req_modifyReservation(struct batch_request *preq)
 				break;
 		}
 
-		if (d == 0) 
+		if (d == 0)
 			presv->ri_wattr[RESV_ATR_duration].at_flags &= ~ATR_VFLAG_SET;
 
-		if (d == 1 && r == 1 && e == 1) {
+		if (d == 1 && e == 1 && r == 1) {
+			presv->ri_alter_state = presv->ri_wattr[RESV_ATR_state].at_val.at_long;
```
bhroam, line 906: Why are you setting the reservation state to being altered if you are rejecting the alter request?

src/server/req_rescq.c
```diff
@@ -1001,6 +1001,24 @@ resv_revert_alter_times(resc_resv *presv)
...
+	if (presv->ri_alter_flags & RESV_DURATION_MODIFIED) {
+		if (presv->ri_alter_stime != 0) {
+			presv->ri_qs.ri_stime = presv->ri_alter_stime;
+			presv->ri_wattr[RESV_ATR_start].at_val.at_long = presv->ri_alter_stime;
+			presv->ri_wattr[RESV_ATR_start].at_flags
+			|= ATR_VFLAG_SET | ATR_VFLAG_MODIFY | ATR_VFLAG_MODCACHE;
+			presv->ri_alter_stime = 0;
+		} 
+		else {
+			if (presv->ri_alter_etime != 0){
```
bhroam, line 1013: Can't both the stime and etime be modified if you do an ralter -R -D?

test/tests/functional/pbs_ralter.py
```diff
@@ -1702,8 +1703,15 @@ def test_adv_resv_endtime_starttime_dur_together(self):
...
             attr = {'reserve_start': new_start, 'reserve_end': new_end,
                     'reserve_duration': new_duration}
-            self.server.alterresv(rid, attr)
-            self.assertIn('pbs_ralter: Bad time specification(s)', e.msg[0])
+            m = self.server.alterresv(rid, attr)
+        self.assertIn('pbs_ralter: Bad time specification(s)',
+                      e.exception.msg[0])
+
+        t_duration, t_start, t_end = self.get_resv_time_info(rid)
+        print(str(self.server.status(RESV, id=rid)))
```
bhroam, line 1711: Remove this debug message.

test/tests/functional/pbs_ralter.py
```diff
@@ -1702,8 +1703,15 @@ def test_adv_resv_endtime_starttime_dur_together(self):
         with self.assertRaises(PbsResvAlterError) as e:
             attr = {'reserve_start': new_start, 'reserve_end': new_end,
                     'reserve_duration': new_duration}
-            self.server.alterresv(rid, attr)
-            self.assertIn('pbs_ralter: Bad time specification(s)', e.msg[0])
+            m = self.server.alterresv(rid, attr)
```
bhroam, line 1706: I don't see m being used.  If you don't use it, remove it.

vstumpf:
Sorry tejas, I think this PR is getting a little too confusing.
We can't guarantee the order the attributes come in as, so it's possible that we enter `case RESV_ATR_start:` and check `if d == 1` before we would even set `d`, causing the code not to run.

I think you need to go about this a different way.
Continue what prakash did with changing the start/end times, but saving the originals as presv->ri_alter_stime/dtime/etime.
Then, in start_end_dur_wall, there's an if block:
```
	if (pstate != RESV_BEING_ALTERED)
		...
	}
	else {
		swcode = 3;
	}
```
Prakash has hardcoded 3, but instead you should check the values of presv->ri_alter_flags to determine what you need to use to calculate the others.
Then alter resv_revert_alter_times to also include duration.

test/tests/functional/pbs_ralter.py
```diff
@@ -1728,6 +1736,8 @@ def test_standing_resv_duration(self):
 
         t_duration, t_start, t_end = self.get_resv_time_info(rid)
         self.assertEqual(t_duration, new_duration)
+        print("new Start is " + str(t_start))
+        print("new End is " + str(t_end))
```
bhroam, line 1740: Either remove these or turn them into self.logger.info() messages

src/server/req_modify.c
tpai26: In resv_revert_alter_times() we evaluate what state the reservation must be set in after revert. Therefore we need to pass in RESV_BEING_ALTERED as state before revert so that the correct state will be evaluated and set after revert. If we do not pass in this the state is set to NO

src/server/req_modify.c
```diff
@@ -862,6 +859,17 @@ req_modifyReservation(struct batch_request *preq)
...
 					return;
 				}
+
+				break;
+			case RESV_ATR_duration:
+				if (presv->ri_alter_etime == 0)
+					presv->ri_alter_flags |= RESV_START_TIME_MODIFIED;
+					presv->ri_alter_etime = presv->ri_wattr[RESV_ATR_end].at_val.at_long;
+					if (presv->ri_alter_stime == 0)
+						presv->ri_alter_stime = presv->ri_wattr[RESV_ATR_start].at_val.at_long;
```
vstumpf, line 869: This should be like the other attributes, where you do something like:
```C
temp = strtol(psatl->al_value, &end, 10);
presv->ri_alter_stime = presv->ri_wattr[RESV_ATR_start].at_val.at_long;
presv->ri_alter_flags |= RESV_DURATION_MODIFIED;
send_to_scheduler = RESV_DURATION_MODIFIED;
```
Do you need to check if it's a standing reservation and will collide with the next occurrance?

src/server/svr_jobfunc.c
```diff
@@ -4603,35 +4603,43 @@ start_end_dur_wall(void *pobj, int objtype)
...
+			if (presv->ri_alter_flags & RESV_END_TIME_MODIFIED)
+				swcode += 6; /*calcualte start time*/
+			if (presv->ri_alter_flags & RESV_START_TIME_MODIFIED)
+				swcode += 5; /*calculate end time*/;
+			if (swcode == 11)
+				return (-1);
+		}
+		else {
+			swcode = 3; /*start, end*/
+		}
```
vstumpf, line 4616: I would change this to something like:
```C
if (presv->ri_alter_flags & RESV_START_TIME_MODIFIED)
    swcode += 1;
if (flags & end_modified)
    += 2;
if (flags & duration_modified)
    += 4;
```
Here you would also have to check if duration is the only thing passed, to use duration.
```C
if (flags == duration_modified)
    swcode += 1;
```

Then in the switch-case statement below, add cases for 1, 2, and 4, rejecting them.

tpai26:
Updated Design doc. All three start time, end time and duration can now be changed together.
Added tests and updated the test logs

src/server/svr_jobfunc.c
tpai26: I changed it according to first part of comment. 
For the second part I just combined case 1 and 2 similar to case 3 and case 4 similar to case 5.

src/server/req_modify.c
tpai26: added a check for this. Also added a test

src/server/req_modify.c
```diff
@@ -862,6 +859,28 @@ req_modifyReservation(struct batch_request *preq)
...
+					else {
+						snprintf(log_buffer, sizeof(log_buffer), "%s", msg_stdg_resv_occr_conflict);
+						log_event(PBSEVENT_RESV, PBS_EVENTCLASS_RESV, LOG_INFO,
+							preq->rq_ind.rq_modify.rq_objname, log_buffer);
+						req_reject(PBSE_STDG_RESV_OCCR_CONFLICT, 0, preq);
+						return;
+					}
+				}
+				if ((!is_standing || temp < next_occr_start) && presv->ri_alter_stime == 0)
+					presv->ri_alter_stime = presv->ri_wattr[RESV_ATR_start].at_val.at_long;
```
vstumpf, line 880: I'm not sure you can do this in the while loop. I think you need to do this check before the `if (send_to_scheduler)`. At that point, we know all the values that are going to change.

vstumpf: Also, you shouldn't have to set ri_alter_stime or etime, it will be set in their respective ATTR cases.

src/server/req_modify.c
```diff
@@ -862,12 +862,23 @@ req_modifyReservation(struct batch_request *preq)
...
-				if (presv->ri_alter_etime == 0)
-					presv->ri_alter_flags |= RESV_START_TIME_MODIFIED;
-					presv->ri_alter_etime = presv->ri_wattr[RESV_ATR_end].at_val.at_long;
-					if (presv->ri_alter_stime == 0)
-						presv->ri_alter_stime = presv->ri_wattr[RESV_ATR_start].at_val.at_long;
-
+				if (presv->ri_alter_etime == 0) {
+					temp = strtol(psatl->al_value, &end, 10) + presv->ri_wattr[RESV_ATR_end].at_val.at_long;
+					if (!is_standing || temp < next_occr_start) {
+						presv->ri_alter_flags |= RESV_START_TIME_MODIFIED;
```
bhroam, line 868: Why are we modifying the start time in the end time block?

src/server/req_modify.c
tpai26: I am setting this for standing reservations, if a backup is not present the standing reservation sets the new time for next occurances

src/server/svr_jobfunc.c
```diff
@@ -4603,35 +4603,41 @@ start_end_dur_wall(void *pobj, int objtype)
...
+				swcode += 2; /*calcualte start time*/
+		if (presv->ri_alter_flags & RESV_START_TIME_MODIFIED)
+				swcode += 1; /*calculate end time*/;
 	}
 
 	atemp.at_flags = ATR_VFLAG_SET;
 	atemp.at_type = ATR_TYPE_LONG;
 	switch (swcode) {
+		case  1:
+		case  2:
```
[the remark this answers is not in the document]
tpai26: Should I do something like

_if (swcode > 0 && swcode < 3)
     swcode = 3_

Because if state is RESV_BEING_ALTERED case 1, 2, 3 are the same.

src/server/req_modify.c
tpai26: I am doing this for standing reservations, in case alter_etime or alter_stime is not set the time for next occurrence also changes

tpai26: Also you are right, I am moving this block of code outside while loop

src/server/svr_jobfunc.c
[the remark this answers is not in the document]
vstumpf: I would do
```C
if (flags == RESV_START_TIME_MODIFIED || flags == RESV_END_TIME_MODIFIED)
    swcode = 3;
```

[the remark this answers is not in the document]
tpai26: That will not work because in addition start_time and end_time,  duration could be modified too. How about something like this
```
  swcode = 3;
  if (presv->ri_alter_flags & RESV_DURATION_MODIFIED) {
	swcode +=1;
	if (presv->ri_alter_flags & RESV_END_TIME_MODIFIED)
		swcode += 2; /*calcualte start time*/
	if (presv->ri_alter_flags & RESV_START_TIME_MODIFIED)
		swcode += 1; /*calculate end time*/;
 }
```

tpai26:
updated the logs

src/server/req_modify.c
```diff
@@ -899,6 +883,34 @@ req_modifyReservation(struct batch_request *preq)
 	}
 	resc_access_perm = resc_access_perm_save; /* restore perm */
 
+	temp_dur = presv->ri_wattr[RESV_ATR_duration].at_val.at_long + presv->ri_wattr[RESV_ATR_end].at_val.at_long;
```
bhroam, line 886: Rename this.  It isn't a temp duration, it is the new end time

src/server/req_modify.c
```diff
@@ -899,6 +883,34 @@ req_modifyReservation(struct batch_request *preq)
...
+		} else {
+				log_event(PBSEVENT_RESV, PBS_EVENTCLASS_RESV, LOG_INFO,
+					preq->rq_ind.rq_modify.rq_objname, msg_stdg_resv_occr_conflict);
+				req_reject(PBSE_STDG_RESV_OCCR_CONFLICT, 0, preq);
+				resv_revert_alter_times(presv);
+				return;
+		}
+	}
+	
+
```
vstumpf, line 913: At this point, you shouldn't be changing anything with ri_alter_etime/stime/duration. This is just a check to see if the reservation will run over to the next occurrance.
The checks for changing start/end already do theirs, we just need to do the duration check.
At this point in time, the start time (new one if changed, old if not) is stored in ri_wattr.
If duration has changed, then do if (start + duration) > next_occr_start: fail, else pass.

src/server/svr_jobfunc.c
```diff
@@ -4601,21 +4601,20 @@ start_end_dur_wall(void *pobj, int objtype)
...
-				swcode += 1; /*calculate end time*/;
+	} else {
+		swcode = 3;
+		if (presv->ri_alter_flags & RESV_DURATION_MODIFIED) {
+			swcode +=1;
+			if (presv->ri_alter_flags & RESV_END_TIME_MODIFIED)
+					swcode += 2; /*calcualte start time*/
+			if (presv->ri_alter_flags & RESV_START_TIME_MODIFIED)
+					swcode += 1; /*calculate end time*/;
+		}
```
vstumpf: If duration is the only thing modified, this would make swcode 4, which is wrong.
```
		if (presv->ri_alter_flags & RESV_START_TIME_MODIFIED)
				swcode += 1; /* use start time*/
		if (presv->ri_alter_flags & RESV_END_TIME_MODIFIED)
				swcode += 2; /* use end time*/
		if (presv->ri_alter_flags & RESV_DURATION_MODIFIED)
				swcode += 4; /* use duration */
		if (presv->ri_alter_flags == RESV_START_TIME_MODIFIED || presv->ri_alter_flags == RESV_END_TIME_MODIFIED)
				swcode = 3; /* only start or only end changed, use start and end to calculate duration */
```

vstumpf: Alternatively, you can do what you did before, which was add the switch cases for 1 and 2, and you just have to put a 
```C
        case 1:
        case 2:
        /* states only valid for altered reservations */ 
        if (pstate != RESV_BEING_ALTERED)
           return -1;
        /* fall through */
        case 3:
```

src/server/req_modify.c
tpai26: > At this point, you shouldn't be changing anything with ri_alter_etime/stime/duration.
But we have to store the original start time and endtime before calling on start_end_dur_wall(). If we do not then for standing reservations time for next occurrences will change too.
For example if we change end time and duration, new wall time will be calculated and set. If it's a standing reservation then the start_time will change for all occurrences. This is why I am storing stime and etime before going forward

src/server/svr_jobfunc.c
tpai26: It can be 4. The case for 4 and 5 is same. Also according to the current code it will never be 4. If we change only duration we make sure to set RESV_START_TIME_MODIFIED flag

tpai26:
updated logs

src/server/svr_jobfunc.c
```diff
@@ -4602,13 +4602,14 @@ start_end_dur_wall(void *pobj, int objtype)
...
 	} else {
-		swcode = 3;
-		if (presv->ri_alter_flags & RESV_DURATION_MODIFIED) {
-			swcode +=1;
-			if (presv->ri_alter_flags & RESV_END_TIME_MODIFIED)
-					swcode += 2; /*calcualte start time*/
-			if (presv->ri_alter_flags & RESV_START_TIME_MODIFIED)
-					swcode += 1; /*calculate end time*/;
+		if (presv->ri_alter_flags & RESV_DURATION_MODIFIED) 
+			swcode +=4;
```
bhroam, line 4606: Add s spacer after the +=

src/server/svr_jobfunc.c
```diff
@@ -4602,13 +4602,14 @@ start_end_dur_wall(void *pobj, int objtype)
...
+		if (presv->ri_alter_flags & RESV_END_TIME_MODIFIED)
+			swcode += 2; /*calcualte start time*/
+		if (presv->ri_alter_flags & RESV_START_TIME_MODIFIED)
+			swcode += 1; /*calculate end time*/;
```
bhroam, line 4610: Remove the semicolon after the comment.

PR #1597, opened 2020-03-23
[WIP] Making pbs_asynrunjob truely asynchronous

agrawalravi90:
#### Describe Bug or Feature

Scheduler today seems to spend the majority of its time waiting for an ACK from server for its runjob request. For my test setup with system with 50k jobs which can all run, the sched cycle took 172 seconds, out of which 161 seconds, or 94% of total time was spent by PBSD_rdrpy(), which is what waits for an ACK from server. pbs_asyrunjob() doesn't really do anything meaningful with the reply that it gets from the server, it frees it in the very next line. So, the purpose of PBSD_rdrpy() call seems to be to just make sure that the server got it, but since we use TCP to communicate, this shouldn't be necessary. We anyways don't send the runjob request again if we don't get a reply from server, so it really doesn't provide much value today, and removing it would make the scheduler much faster.

#### Describe Your Change

Removed the call to PBSD_rdrpy() from pbs_asyrunjob(), and added checks in server so that it won't send a reply for pbs_asyrunjob() requests.

**Performance numbers:**
50k jobs, 50k ncpus, time taken by 1 sched cycle:
before: 3 minutes
after: 12 seconds 

This is refactoring, so running existing tests should suffice.

### DIFF
--- a/src/server/reply_send.c
+++ b/src/server/reply_send.c
@@ -279,6 +279,10 @@ reply_send(struct batch_request *request)
 	int		    rc = 0;
 	int		    sfds = request->rq_conn;		/* socket */
 
+	/* No need to send reply for asynrunjob */
+	if (request->rq_type == PBS_BATCH_AsyrunJob)
+		return 0;
+
 	/* if this is a child request, just move the error to the parent */
 
 	if (request->rq_parentbr) {
@@ -362,6 +366,10 @@ reply_ack(struct batch_request *preq)
 	if (preq == NULL)
 		return;
 
+	/* No need to send ack for asynrunjob */
+	if (preq->rq_type == PBS_BATCH_AsyrunJob)
+		return;
+
 	if (preq->prot == PROT_TPP && preq->tpp_ack == 0) {
 		free_br(preq);
 		return;
@@ -442,6 +450,10 @@ req_reject(int code, int aux, struct batch_request *preq)
 	char  msgbuf[ERR_MSG_SIZE];
 
 	if (preq == NULL)
+		return;
+
+	/* No need to send reject for asynrunjob */
+	if (preq->rq_type == PBS_BATCH_AsyrunJob)
 		return;
 
 	if (code != PBSE_NONE) {
--- a/src/lib/Libifl/pbsD_asyrun.c
+++ b/src/lib/Libifl/pbsD_asyrun.c
@@ -66,7 +66,6 @@ int
 __pbs_asyrunjob(int c, char *jobid, char *location, char *extend)
 {
 	int	rc;
-	struct batch_reply   *reply;
 	unsigned long resch = 0;
 
 	if ((jobid == NULL) || (*jobid == '\0'))
@@ -107,13 +106,6 @@ __pbs_asyrunjob(int c, char *jobid, char *location, char *extend)
 		(void)pbs_client_thread_unlock_connection(c);
 		return pbs_errno;
 	}
-
-	/* get reply */
-
-	reply = PBSD_rdrpy(c);
-	rc = get_conn_errno(c);
-
-	PBSD_FreeReply(reply);
 
 	/* unlock the thread lock and update the thread context data */
 	if (pbs_client_thread_unlock_connection(c) != 0)


arungrover:
Hey Ravi,
I think if we do this then the server will stop sending a reply back in the case when a runjob hook rejects a job. While the scheduler does nothing with the reply, it does need the error number in order to identify that a runjob hook has rejected (or there is a protocol failure) the request and scheduler won't account for the job's requested resources.

agrawalravi90:
> Hey Ravi,
I think if we do this then the server will stop sending a reply back in the case when a runjob hook rejects a job. While the scheduler does nothing with the reply, it does need the error number in order to identify that a runjob hook has rejected (or there is a protocol failure) the request and scheduler won't account for the job's requested resources.

Thanks for your reply Arun. If I understand your concern correctly, then this is similar to the mom rejecting a job, which will cause the same issue today, so is that a big issue? The next sched cycle, which will hopefully be sooner after this, scheduler will see that the jobs are queued and schedule them again right? Am I missing something important?

src/server/reply_send.c
```diff
@@ -279,6 +279,10 @@ reply_send(struct batch_request *request)
 	int		    rc = 0;
 	int		    sfds = request->rq_conn;		/* socket */
 
+	/* No need to send reply for asynrunjob */
+	if (request->rq_type == PBS_BATCH_AsyrunJob)
+		return 0;
```
hirenvadalia: we need free request before returning

src/server/reply_send.c
```diff
@@ -362,6 +366,10 @@ reply_ack(struct batch_request *preq)
 	if (preq == NULL)
 		return;
 
+	/* No need to send ack for asynrunjob */
+	if (preq->rq_type == PBS_BATCH_AsyrunJob)
+		return;
```
hirenvadalia: same as above

src/server/reply_send.c
```diff
@@ -444,6 +452,10 @@ req_reject(int code, int aux, struct batch_request *preq)
 	if (preq == NULL)
 		return;
 
+	/* No need to send reject for asynrunjob */
+	if (preq->rq_type == PBS_BATCH_AsyrunJob)
+		return;
```
hirenvadalia: same as above

bhroam:
@agrawalravi90 Arun is correct.  The runjob hook needs to return back to the scheduler.  The runjob hook was designed to have the scheduler to not use those resources.  It also returns back a message that gets added to the comment/log.  This reply has to get back to the scheduler.

arungrover:
> 
> Thanks for your reply Arun. If I understand your concern correctly, then this is similar to the mom rejecting a job, which will cause the same issue today, so is that a big issue? The next sched cycle, which will hopefully be sooner after this, scheduler will see that the jobs are queued and schedule them again right? Am I missing something important?

I think it is a little different than a job getting rejected by a mom hook. The reason being, if a mom hook rejects a job, the job is penalized by increasing the run count and eventually such jobs are held by the system. But if a runjob hook rejects the job and this information is not passed on to the scheduler then in every cycle scheduler will take the wrong decision in every cycle by assuming that the job ran and will continue to do so because such a job will not be held automatically.

agrawalravi90:
Thanks for the inputs guys, I understand now that I'll need to come up with a workaround for the runjob hook issue. As you guys suggested, maybe we can make this configurable, tell scheduler to care/not care about runjob hooks and admins can decide if they want to speed things up by telling sched to not care. I would also like to see if we can hold the rejected job for 1/few cycle(s) so that other jobs can use those resources, see if it affects utilization in, and if it is ok to do this. Maybe there are some other ways to get around this as well.

So I'll mark this [WIP] for now.

agrawalravi90:
Started a forum discussion here: http://community.pbspro.org/t/scheduler-spends-94-of-its-time-waiting-for-job-run-ack/2053

agrawalravi90:
The discussion is going in a slightly different direction, a new sched attribute also needs to be created, so I'll close this PR and raise a new one when we've finalized the design

PR #1598, opened 2020-03-24
Fix memory leaks for some of obj_save_db functions in case of db save…

suresh-thelkar:
… failures

#### Describe Bug or Feature

When database failures like object recovery or save fails some of the obj_save_db functions memory leak.
#### Describe Your Change

Following is the backtrace when the issue happens.
76 bytes in 7 blocks are definitely lost in loss record 249 of 1,365
   at 0x4C2A433: malloc (vg_replace_malloc.c:309)
   by 0x6A9BC79: strdup (in /lib64/libc-2.22.so)
   by 0x46B73D: encode_attr_db (attr_recov_db.c:221)
   by 0x4AF652: svr_to_db_que (queue_recov_db.c:114)
   by 0x4AF748: que_save_db (queue_recov_db.c:170)
   by 0x4BACD4: mgr_queue_create (req_manager.c:1238)
   by 0x4C2E5A: req_manager (req_manager.c:4745)
   by 0x4AD98F: dispatch_request (process_request.c:797)
   by 0x483397: issue_Drequest (issue_request.c:446)
   by 0x4CD081: get_queue_for_reservation (req_quejob.c:3138)
   by 0x4CC03F: req_resvSub (req_quejob.c:2713)
   by 0x4AD8F4: dispatch_request (process_request.c:766)

We could not reproduce the issue either in our dev or test environment at present. Fixed the code with the help of backtrace and code inspection.

It looks like these leaks happen when a database recovery or save fails during which case the control reaches to the label db_err in most of the object save/recovery functions.  Here in some of the places we are not calling pbs_db_reset_obj() which actually frees db attributes. However at the end of this label we call panic_stop_db() which stops db as well as PBS Server from which time the memory gets freed anyways. But if tomorrow if we handle db errors without calling panic_stop_db() then these will occupy lot of memory of the Server host and hence the fix.

image
Attached are entire regression logs.

### DIFF
--- a/src/server/attr_recov_db.c
+++ b/src/server/attr_recov_db.c
@@ -172,6 +172,7 @@ encode_attr_db(struct attribute_def *padef, struct attribute *pattr, int numattr
 	pbs_db_attr_info_t *attrs = NULL;
 
 	attr_list->attr_count = 0;
+	attr_list->attributes = NULL;
 
 	/* encode each attribute which has a value (not non-set) */
 	CLEAR_HEAD(lhead);
--- a/src/server/job_recov_db.c
+++ b/src/server/job_recov_db.c
@@ -410,11 +410,12 @@ job_save_db(job *pjob, int updatetype)
 	else
 		savetype = PBS_UPDATE_DB_FULL;
 
+	obj.pbs_db_obj_type = PBS_DB_JOB;
+	obj.pbs_db_un.pbs_db_job = &dbjob;
+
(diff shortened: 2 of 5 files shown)


suresh-thelkar:
@arungrover / @subhasisb, Requesting you to review the changes as you are aware of the code of the code changes.

suresh-thelkar:
Thanks @subhasisb for the sign off. @arungrover  can you also please review the changes.

suresh-thelkar:
Thanks @arungrover.

PR #1599, opened 2020-03-24
Server won't unset the resources_assigned attribute during resource unset anymore, if present

Bhagat-Rajput:
#### Describe Bug or Feature

Queue resources_assigned goes negative because we unset a resource which is already been occupied by the job.
**Steps to reproduce:**
   1.  set q <queue_name> resources_available.ncpus=<value>
   2.  run a job
   3. unset q <qeueu_name> resources_available.ncpus
   4. qdel the job
   5. qstat -Qf <queue_name> and see "resources_assigned" value

#### Describe Your Change

As per the new changes, If we unset the resource then will remove only resources_available.<resc_name> attribute  and leave resources_assigned.<resc_name> attribute as it is with its original value((could be zero as well) if present. Resources_assigned.<resc_name> will be there in the system forever, until the resource is deleted or server can also unset this attribute if didn't find any jobs related to that particular resource in the system at the time of restart.

PTL logs:- [ptl_que_resc_usage.txt](https://github.com/PBSPro/pbspro/files/4386420/ptl_que_resc_usage.txt)

### DIFF
--- a/test/tests/functional/pbs_que_resc_usage.py
+++ b/test/tests/functional/pbs_que_resc_usage.py
@@ -50,6 +50,19 @@ class TestQueRescUsage(TestFunctional):
                             'resources_available.ncpus': 10},
                             self.mom.shortname)
 
+    def create_custom_resc(self):
+        """
+        function to create the resource named as "foo"
+        """
+        self.server.manager(MGR_CMD_CREATE, RSC, {
+                            'type': 'long', 'flag': 'q'}, id='foo')
+        self.scheduler.add_resource('foo')
+        self.server.manager(MGR_CMD_SET, QUEUE, {
+                            'resources_available.foo': 6}, id='workq')
+        q_status = self.server.status(QUEUE, id='workq')
+        self.assertEqual(
+            int(q_status[0]['resources_available.foo']), 6, self.err_msg)
+
     @skipOnCpuSet
     def test_resc_assigned_set_unset(self):
         """
@@ -127,15 +140,8 @@ class TestQueRescUsage(TestFunctional):
         in the system or else unset it.
         """
 
-        # create a custom resource
-        self.server.manager(MGR_CMD_CREATE, RSC, {
-                            'type': 'long', 'flag': 'q'}, id='foo')
-        self.scheduler.add_resource('foo')
-        self.server.manager(MGR_CMD_SET, QUEUE, {
-                            'resources_available.foo': 6}, id='workq')
-        q_status = self.server.status(QUEUE, id='workq')
-        self.assertEqual(
-            int(q_status[0]['resources_available.foo']), 6, self.err_msg)
+        # create a resource
+        self.create_custom_resc()
 
         # resources_assigned is zero but still jobs are in the system
         j1_attr = {ATTR_queue: 'workq',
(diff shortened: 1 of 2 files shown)


test/tests/functional/pbs_que_resc_usage.py
```diff
@@ -167,3 +173,36 @@ def test_resources_assigned_with_zero_val(self):
...
+
+    @skipOnCpuSet
+    def test_resources_assigned_deletion(self):
+        """
+        Test resources_assigned.<resc_name> deletion from the system
+        """
+        # create a resource
+        self.create_custom_resc()
+        # submit jobs
+        j_attr = {ATTR_queue: 'workq',
```
bhroam: No real reason to request workq.  It is the default.

test/tests/functional/pbs_que_resc_usage.py
```diff
@@ -167,3 +173,36 @@ def test_resources_assigned_with_zero_val(self):
...
+                  'Resource_List.select': '1', 'Resource_List.foo': '3'}
+        j1 = Job(TEST_USER, j_attr)
+        j1.set_sleep_time(30)
+        jid_1 = self.server.submit(j1)
+        j2 = Job(TEST_USER, j_attr)
```
bhroam: No need to request TEST_USER, it is the default.  Just says attrs=j_attr

test/tests/functional/pbs_que_resc_usage.py
```diff
@@ -167,3 +173,36 @@ def test_resources_assigned_with_zero_val(self):
...
+        j2.set_sleep_time(30)
```
bhroam: You aren't waiting for the job to finish, no need to set the sleep time.

test/tests/functional/pbs_que_resc_usage.py
```diff
@@ -167,3 +173,36 @@ def test_resources_assigned_with_zero_val(self):
...
+        jid_2 = self.server.submit(j2)
+        self.server.expect(JOB, {'job_state': 'R'}, jid_1)
+        self.server.expect(JOB, {'job_state': 'R'}, jid_2)
+        # try to delete the resource when it's busy on job
+        try:
+            self.server.manager(MGR_CMD_DELETE, RSC, id='foo')
+        except PbsManagerError as e:
+            self.assertTrue("Resource busy on job" in e.msg[0])
```
bhroam: Use assertIn()

test/tests/functional/pbs_que_resc_usage.py
```diff
@@ -167,3 +173,36 @@ def test_resources_assigned_with_zero_val(self):
...
+        self.server.expect(JOB, 'queue', op=UNSET, id=jid_1)
+        self.server.expect(JOB, 'queue', op=UNSET, id=jid_2)
+        # now jobs has been finished, try to delete the resource again
+        self.server.manager(MGR_CMD_DELETE, RSC, id='foo')
+        q_status = self.server.status(QUEUE, id='workq')
+        self.assertNotIn('resources_assigned.foo',
+                         q_status[0].keys(), self.err_msg)
+        self.assertNotIn('resources_available.foo',
+                         q_status[0].keys(), self.err_msg)
```
bhroam: No reason to say .keys(), just use the dictionary itself.

test/tests/functional/pbs_que_resc_usage.py
```diff
@@ -50,6 +50,19 @@ def setUp(self):
...
+    def create_custom_resc(self):
+        """
+        function to create the resource named as "foo"
+        """
+        self.server.manager(MGR_CMD_CREATE, RSC, {
+                            'type': 'long', 'flag': 'q'}, id='foo')
+        self.scheduler.add_resource('foo')
+        self.server.manager(MGR_CMD_SET, QUEUE, {
+                            'resources_available.foo': 6}, id='workq')
+        q_status = self.server.status(QUEUE, id='workq')
```
bhroam: I'm not sure you really need to do this.  If the manager() failed, it would raise an exception.

test/tests/functional/pbs_que_resc_usage.py
Bhagat-Rajput: Actually, here we are waiting for jobs to end after "except".

Bhagat-Rajput:
Thanks @bhroam for the review.
Since this PR was merged by that time so i've addressed all your comments in a new PR https://github.com/PBSPro/pbspro/pull/1619. Please review it. Thanks.

PR #1601, opened 2020-03-24
Suppress intentional unfreed memory of auth struct inside global tpp_conf struct

hirenvadalia:
#### Describe Bug or Feature
* Valgrind shows leak for international unfreed memory of auth struct inside global tpp_conf struct per PBS daemon
    ```
   82 (48 direct, 34 indirect) bytes in 1 blocks are definitely lost in loss record 69 of 600
       at 0x4C2BFB9: calloc (vg_replace_malloc.c:762)
       by 0x4B8A3F: make_auth_config (auth.c:715)
       by 0x4E0909: set_tpp_config (tpp_util.c:341)
       by 0x481177: main (mom_main.c:9281)
  9 bytes in 1 blocks are possibly lost in loss record 9 of 731
       at 0x4C2E2DF: malloc (in /usr/lib64/valgrind/vgpreload_memcheck-amd64-linux.so)
       by 0x6B15299: strdup (in /lib64/libc-2.26.so)
       by 0x5A732D: make_auth_config (auth.c:730)
       by 0x531897: set_tpp_config (tpp_util.c:341)
       by 0x4A8475: main (pbsd_main.c:1544)
   ```

#### Describe Your Change

* Updated suppression file to suppress intentional unfreed memory

#### Link to Design Doc

* None

* No code change, so no new test required

### DIFF
--- a/valgrind.supp
+++ b/valgrind.supp
@@ -601,3 +601,20 @@
    fun:wait_request
    fun:main
 }
+{
+   From PBSPro - Suppress intentional unfreed memory of auth struct inside global tpp_conf struct
+   Memcheck:Leak
+   fun:calloc
+   fun:make_auth_config
+   fun:set_tpp_config
+   fun:main
+}
+{
+   From PBSPro - Suppress intentional unfreed memory of auth struct inside global tpp_conf struct
+   Memcheck:Leak
+   fun:malloc
+   fun:strdup
+   fun:make_auth_config
+   fun:set_tpp_config
+   fun:main
+}


bhroam:
Valgrind won't report a leak when memory is still held onto at the end of the process.  This is a true leak.  In any case, you have your two approvals, I won't stand in the way.

hirenvadalia:
> Valgrind won't report a leak when memory is still held onto at the end of the process. This is a true leak. In any case, you have your two approvals, I won't stand in the way.

Hey @bhroam, I understand your concern but this is not true leak, if you see the code then make_auth_config() is called (in case of TPP) only in set_tpp_config() (https://github.com/PBSPro/pbspro/blob/master/src/lib/Libtpp/tpp_util.c#L341) and set_tpp_config() is only called once per daemon (from main()) and rest code will never modify `tpp_conf->auth_config`, it only reads from it.
(You can simply search code with `tpp_conf->auth_config = ` you will see there is only one place in the whole codebase)

mike0042:
You're not wrong @bhroam, this is truly a leak. We could make all the daemons call something like free_tpp_config() prior to exiting that would free the allocated memory and silence the valgrind warnings, or we could ignore them since we know we are exiting. The danger of ignoring the warnings is very low in this case, but I acknowledge your concern is valid.

PR #1602, opened 2020-03-25
Making job attr updates from sched asynchronous

agrawalravi90:
#### Describe Bug or Feature

This is a performance enhancement. Currently, job attr updates from scheduler are synchronous: pbs_alterjob() calls PBSD_manager() which calls PBSD_rdrpy() to wait for a reply from the server. This can be a performance bottleneck when there are a large number of alterjobs every scheduling cycle.

#### Describe Your Change

- Using the pbs_alterjob() & PBSD_manager()'s 'extend' argument to pass "0" from scheduler. PBSD_manager() will take this to mean "don't wait for reply from server"

Smoke tests should be good enough to test functionality.
Added a performance test for 10k jobs:
**before:** Time taken by the sched cycle: 6.609070062637329
**after:** Time taken by the sched cycle: 0.6671128273010254


bhroam:
@agrawalravi90 '0' seems like a poor choice to pass to the extend field.  All our current extend are a little more self explanatory.  I'd suggest either A) create new IFL calls or B) use an extend field like 'nowait' or something like that.

agrawalravi90:
@bhroam how about 'async' ?

bhroam:
Yeah that sounds great.

agrawalravi90:
@bhroam  and @ashwathraop thanks for the sign-offs, but I had to change my approach. I found some issues during regression and realized that I needed the server to not reply at all when sched does an async alterjob. So, I created a new batch request for async modifyjob for which server will not send any replies back. This passes regression.

bhroam:
How about we be consistent with our other asynchronous call pbs_asyrunjob() and name this pbs_asyalterjob()?

agrawalravi90:
@bhroam I've renamed the call, please provide further feedback, thanks.

test/tests/performance/pbs_ifl_perf.py
```diff
@@ -0,0 +1,70 @@
...
+# Use of Altair’s trademarks, including but not limited to "PBS™",
+# "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
+# trademark licensing policies.
+
+from tests.performance import *
+
+
+class TestIFLPerf(TestPerformance):
+    @timeout(3600)
+    def test_attr_update_perf_from_sched(self):
```
bhroam: A) this should have been part of the scheduler performance test suite,  B) the test already exists (TestSchedPerf.test_run_many_normal_jobs)

agrawalravi90:
@bhroam thanks for pointing that out, I've removed the PTL test, lemme know if it looks okay now.

bhroam:
Hey Ravi, go ahead and resolve your conflicts.  Hopefully they won't be bad.

src/scheduler/job_info.c
```diff
@@ -129,7 +129,6 @@
 #include "site_queue.h"
 #endif
 
-
 extern int pbs_asyalterjob(int c, char *jobid, struct attrl *attrib, char *extend);
```
bhroam: Shouldn't this be in pbs_ifl.h or pbs_internal.h?

agrawalravi90:
hey @bhroam , do you think this is safe enough to be checked in right now? Or should we wait until 2020.1 is forked off?

bhroam:
@agrawalravi90 good question.  Maybe ask @subhasisb or @scc138?

agrawalravi90:
@scc138  and @subhasisb  do you guys think it's ok to merge this in? Or should we wait until 2020.1 is forked off?

subhasisb:
I feel that this is safe enough, and also we have quite some time to harden 2020.1 between now and the release. If we find issues in tests down the line, we can always pull it back out, fairly easily.

agrawalravi90:
Great, thanks for confirming Subhasis

agrawalravi90:
@bhroam please let me know if there's anything else left for this, thanks!

bhroam:
@agrawalravi90 I'm fine with the change, is it ready for merge?

agrawalravi90:
@bhroam yes, it's ready for merge, thanks!

PR #1603, opened 2020-03-25
Implement support for cpu ctrlr and 0-CPU jobs

alexis-cousein:
#### Describe Bug or Feature

See https://pbspro.atlassian.net/wiki/spaces/PD/pages/1627750401/Support+for+the+cpu+cgroup+controller+and+zero-CPU+jobs

#### Describe Your Change

Discussed in the document

#### Link to Design Doc

https://pbspro.atlassian.net/wiki/spaces/PD/pages/1627750401/Support+for+the+cpu+cgroup+controller+and+zero-CPU+jobs

N/A -- MoM hook, not one of the main daemons

image

### DIFF
--- a/src/hooks/cgroups/pbs_cgroups.PY
+++ b/src/hooks/cgroups/pbs_cgroups.PY
@@ -4510,6 +4510,8 @@ class CgroupUtils(object):
                 # This only sets the cpu controller limits/shares,
                 # since the cpuset stuff is done through
                 # cpuset.cpus and cpuset.mems
+                # Note the set_limit in the final 'else' should be
+                # skipped -- its replacement is at the end here
                 htpc = 1
                 if (self.cfg['cgroup']['cpu']['enabled']
                         and self.cfg['use_hyperthreads']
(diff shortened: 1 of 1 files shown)


src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -4451,7 +4501,40 @@ class CgroupUtils(object):
...
+                    pbs.logmsg(pbs.EVENT_DEBUG,
+                               '%s: Failed to delete cpuset %s'
+                               ' for zero cpus job %s'
+                               % (caller_name(), path, jobid))
+                    pbs.event().reject('Failed to delete cpuset'
+                                       ' for zero cpus job')
+                    # Hopefully the orphan cleaner will eventually
+                    # get rid of it
+            else:
+                # For everything else just pass it down
```
mike0042: Either remove the else or add a call to pass here.

alexis-cousein: In my original copy the self.set_limit behind this was indented under the else...let me check the logic.

alexis-cousein: Yup, somehow an indentation was lost in translation when some things were merged. Go figure.

The logic is that seld.setlimit needs _NOT_ to be done for the two special cases. For "ncpus" there is a set_limit that takes care of multiplying what we need to set by the number of threads if necessarey; for cpuset.cpus we need not call set_limit for ncpus==0 (in that case hostresc['cpuset.cpus'] is empty, and we test for that), but delete the cpuset (after which the processes will be added to the main pbspro.slice cpuset).

In all *other* cases, _then_ we "just" have to call setlimit.

mike0042:
Tests ran as expected. Please re-review @bayucan.

PR #1604, opened 2020-03-26
Atom hook should offline all vnodes reported by a mom

vstumpf:
#### Describe Bug or Feature
The cray atom hook was not offlining vnodes, and raising an exception for an improper use of 'e.message'

#### Describe Your Change
Use fail_action = offline_vnodes hook attribute.

Manual test:
[after-fix.txt](https://github.com/PBSPro/pbspro/files/4384377/after-fix.txt)
[after-fix-multivnode.txt](https://github.com/PBSPro/pbspro/files/4390787/after-fix-multivnode.txt)

#### Design
https://pbspro.atlassian.net/wiki/spaces/PD/pages/1086029883/PBS+Design+Changes+for+Shasta

### DIFF
--- a/src/modules/python/pbs_hooks/PBS_cray_atom.HK
+++ b/src/modules/python/pbs_hooks/PBS_cray_atom.HK
@@ -4,3 +4,4 @@ user=pbsadmin
 event=execjob_begin,execjob_end
 order=100
 alarm=300
+fail_action=offline_vnodes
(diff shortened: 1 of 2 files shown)


lisa-altair:
Thanks for making these changes.  I believe this is the right direction to go in.
I realize that you are using an established feature to accomplish offlining all the vnodes belonging to one mom...but since that is the bulk of the change that you are making here, can you show the manual test output for multiple vnodes per mom?

vstumpf:
I've added logs where the mom has 4 vnodes. I also added the qstat -f output of the job, where you can see the run_count is 1, meaning the job tried to run once and offlined all the nodes.

PR #1606, opened 2020-03-26
changes to user reliability test to not expect a certain number of ncpus

prakashcv13:
#### Describe Bug or Feature

The test test_create_resv_from_array_job from Test_user_reliability would fail on machines having ncpus < 3.

#### Describe Your Change

Changed the test to not check for all subjobs to be running.

[ptl_24538.txt](https://github.com/PBSPro/pbspro/files/4386372/ptl_24538.txt)

### DIFF
--- a/test/tests/functional/pbs_user_reliability.py
+++ b/test/tests/functional/pbs_user_reliability.py
@@ -179,11 +179,17 @@ j.create_resv_from_job=1
         This test confirms that a reservation cannot be created out of an
         array job.
         """
+
+        # Let us not assume the number of ncpus
+        nodes = self.server.status(NODE)
+        ncpus = nodes[0]['resources_available.ncpus']
+        j_range = '1-' + ncpus
+
         j = Job(TEST_USER)
-        j.set_attributes({ATTR_J: '1-3'})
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_user_reliability.py
```diff
@@ -179,11 +179,15 @@ def test_create_resv_from_array_job(self):
         This test confirms that a reservation cannot be created out of an
         array job.
         """
+        nodes = self.server.status(NODE)
+        ncpus = nodes[0]['resources_available.ncpus']
+        j_range = '1-' + ncpus
```
sujatapatnaik52: If the available ncpus is 1, In that case the job submission will be "qsub -J 1-1"? I tried this in my local vm and that doesn't seem to work. Please verify once. Rest looks to me.

prakashcv13: oops. I missed that. Thanks for catching.
I have changed the test case to not look for all subjobs to be running.

prakashcv13:
test logs - 
[ptl_24538-2.txt](https://github.com/PBSPro/pbspro/files/4387232/ptl_24538-2.txt)

PR #1608, opened 2020-03-26
Fix for premature end of message when auth fails

hirenvadalia:
#### Describe Bug or Feature
* When authentication fails on the server and if the server sends an auth error message back to client still client fails with `Premature end of message (null)`, instead of printing auth error message from the server
* This mostly happens when LibAuth library is not actually doing the handshake, for example, Munge auth lib, on the client, it gets auth data and marks handshake is done (because it doesn't need any data back from server). So, if the auth library says a handshake is done, PBS proceeds further.
* Basically we need a way to say to the client that we have received your data and its valid or not

#### Describe Your Change
* (On server-side) Changed code such a way if auth fails then it sends AUTH_ERR_DATA (existing code) and if auth succeeds then send AUTH_CTX_OK to notify the client auth data is valid
* (On client-side) Always wait for on pkt from the server, which will be either AUTH_CTX_DATA (if handshake needs continuation), AUTH_ERR_DATA (if auth failed on the server) or AUTH_CTX_OK (if auth succeed on the server)
* Modified Munge LibAuth implementation to pass auth error message back to client

#### Link to Design Doc

* None

* [before-fix.txt](https://github.com/PBSPro/pbspro/files/4386580/before-fix.txt)
* [after-fix.txt](https://github.com/PBSPro/pbspro/files/4386582/after-fix.txt)

### DIFF
--- a/src/lib/Libauth/munge/munge_supp.c
+++ b/src/lib/Libauth/munge/munge_supp.c
@@ -140,6 +140,9 @@ err:
  * @brief
  *	munge_get_auth_data - Call Munge encode API's to get the authentication data for the current user
  *
+ * @param[in] ebuf - buffer to hold error msg if any
+ * @param[in] ebufsz - size of ebuf
+ *
  * @return char *
  * @retval !NULL - success
  * @retval  NULL - failure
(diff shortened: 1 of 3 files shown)


src/lib/Libauth/munge/munge_supp.c
```diff
@@ -146,35 +146,37 @@ init_munge(void)
  *
  */
 static char *
-munge_get_auth_data(void)
+munge_get_auth_data(char *ebuf, size_t ebufsz)
```
subhasisb: Add a comment that ebuf was initialized with nulls all through and ebufsz value passed is sizeof(ebuf) - 1, and that is why we do not need to null terminate the last byte in the snprintf

hirenvadalia:
Force pushed to get Travis fix...

src/lib/Libauth/munge/munge_supp.c
bhroam: This was actually unneeded.  snprintf() will write as much as N-1 bytes and always null terminate.  It's strncpy() which has the problem.

PR #1609, opened 2020-03-26
pbs_server daemon core dumps while setting "restrict_res_to_release_on_suspend" attribute

nagavg:
#### Describe Bug or Feature
pbs_server daemon core dumps while setting "restrict_res_to_release_on_suspend" attribute.

#### Describe Your Change
Changed code to run pbs_server daemon while setting "restrict_res_to_release_on_suspend" attribute.

#### Attach Test logs/Output
[Before_fix_logs.txt](https://github.com/PBSPro/pbspro/files/4388452/Before_fix_logs.txt)
[After_fix_logs.txt](https://github.com/PBSPro/pbspro/files/4388451/After_fix_logs.txt)

### DIFF
--- a/src/server/node_manager.c
+++ b/src/server/node_manager.c
@@ -8664,14 +8664,16 @@ int update_resources_rel(job *pjob, attribute *attrib, enum batch_op op)
 		 */
 		if ((prdef->rs_flags & ATR_DFLAG_RASSN) &&
 			(find_resc_entry(&pjob->ji_wattr[(int) JOB_ATR_resc_released_list], prdef) == NULL)) {
-			for (j = 0; j < server.sv_attr[(int)SVR_ATR_restrict_res_to_release_on_suspend].at_val.at_arst->as_usedptr; j++) {
-				if (strcmp(server.sv_attr[(int)SVR_ATR_restrict_res_to_release_on_suspend].at_val.at_arst->as_string[j],
-				    prdef->rs_name) == 0) {
-					presc = add_resource_entry(&pjob->ji_wattr[(int) JOB_ATR_resc_released_list], prdef);
-					if (presc == NULL)
-						return 1;
-					prdef->rs_set(&presc->rs_value, &presc_sq->rs_value, op);
-					break;
+			if (server.sv_attr[(int) SVR_ATR_restrict_res_to_release_on_suspend].at_val.at_arst) {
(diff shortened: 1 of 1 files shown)


arungrover:
@agrawalravi90 I think you are correct. 
Even through, this NULL check is probably the right thing to do anyway, we should try to find why resources_release_list is not set.
This problem can only happen when there is a job suspended while RRTROS was set and then admin unsets RRTROS and the server restarts. But even in that case, when a job is saved into the database it should save resource_released_list attributes into the DB. 
We should find out why resource_release_list attribute is not saved in DB and fix that.

nagavg:
Hi Arun / Ravi,

Thank you for reviewing and providing feedback.

The problem happens when we set, unset, submit a job - suspend it  and restart the server.
Because we submitted and suspended the job when RRTROS was not set and resource_released_list is also not set.

arungrover:
So @nagavg I think the correct order is that you set restrict_res_to_release_on_suspend (let us call it RRTROS) suspend a job, unset RRTROS, restart the server. 
What's interesting is that just after you unset RRTROS, the job still has resource_released_list but after the server restarts it becomes unset, which to me is wrong.
The change you have done in the code is valid. But, along with that change, we should make sure that the job attribute gets saved into database so that the control never falls into update_resources_rel() function for such jobs.

nagavg:
Hi @arungrover,

As per analysis after restarting the server, mom sends obit for the job _state=4 and substate=43 (suspend).
job_obit() in this function state and substate is changed to exiting, rerun in on_job_rerun, so we free this attribute.
that is why we do not see the value and end up calling update_resource_rel().

arungrover:
@nagavg I don't understand what has mom got to do with this. You are only restarting server right? 
If you restart the server, a running (or suspended) job should remain in the state it was. 
resource_released_list is an attribute that is created when a running job is suspended and RRTROS is set on the server. After this, the job should be saved to DB and upon a server restart this attribute should be read by the server and it never has to call update_resource_rel() on such a job (because it will already be set on the job).
You might have to find why resource_released_list attribute is not getting saved in the database when the job is suspended.

agrawalravi90:
@nagavg any update on this?

arungrover:
@nagavg I'm closing this PR since it has been inactive for more than 60 days. Please reopen it when you get back to it.

PR #1611, opened 2020-03-27
Updates to MoM class' save_configuration() & load_configuration() for saving mom configurations when option '--use-current-setup' is used in PTL's pbs_benchpress

saritakh:
#### Describe Bug or Feature
MoM class' save_configuration() & load_configuration() methods were not updated using json, for saving mom configurations when option '--use-current-setup' is used in PTL's pbs_benchpress

#### Describe Your Change
Updated MoM class' save_configuration() & load_configuration() for saving mom configurations when option '--use-current-setup' is used in PTL's pbs_benchpress
Updated below methods to save mom configurations
save_configuration()
load_configuration()
setUpClass()
tearDownClass()
setUp()
tearDown()

#### Link to Design Doc
Design Documentation https://pbspro.atlassian.net/wiki/spaces/PD/pages/557088789/Design+for+a+supported+way+to+change+default+setup+in+PTL

[test_steps.txt](https://github.com/PBSPro/pbspro/files/4401833/test_steps.txt)
[check_30mar.out4.txt](https://github.com/PBSPro/pbspro/files/4401834/check_30mar.out4.txt)

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -4313,7 +4313,6 @@ class PBSService(PBSObject):
                             self.logger.error("Failed to restore "
                                               + "configuration: %s" % k)
                             return False
-                        self.du.rm(path=fn, force=True, sudo=True)
                     except:
                         self.logger.error("Failed to restore "
                                           + "configuration: %s" % k)
@@ -13397,7 +13396,7 @@ class MoM(PBSService):
                 self._save_config_file(conf,
                                        os.path.join(mpriv, 'config.d', f))
         mconf = {self.hostname: conf}
-        if MGR_OBJ_NODE not in self.server.saved_config.keys():
+        if MGR_OBJ_NODE not in self.server.saved_config:
             self.server.saved_config[MGR_OBJ_NODE] = {}
         self.server.saved_config[MGR_OBJ_NODE].update(mconf)
         if outfile is not None:
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -4294,7 +4294,7 @@ class PBSService(PBSObject):
             elif objtype == MGR_OBJ_SCHED:
                 for k, v in conf.items():
                     fn = self.du.create_temp_file()
-                    if os.path.isfile(fn):
+                    try:
                         rv = self.du.chmod(path=fn, mode=0o644)
                         if not rv:
                             self.logger.error("Failed to restore "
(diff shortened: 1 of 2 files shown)


test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4303,11 +4305,41 @@ def _load_configuration(self, infile, objtype=None):
...
+                                              + "configuration: %s" % k)
+                            return False
                     finally:
                         if os.path.isfile(fn):
-                            self.du.rm(path=fn)
+                            self.du.rm(path=fn, force=True, sudo=True)
+                return True
+            elif objtype == MGR_OBJ_NODE:
+                conf = sconf[str(objtype)]
+                nconf = {}
```
sujatapatnaik52: I think the above initialization is not needed since you are assigning the value in the next line.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4243,6 +4243,7 @@ def _load_configuration(self, infile, objtype=None):
                     return False
             conf = sconf[str(objtype)]
             if objtype == MGR_OBJ_SERVER:
+                conf = sconf[str(objtype)]
```
sujatapatnaik52: you are setting `conf` just above one line, I did not understand why are you doing it here?

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4303,11 +4305,41 @@ def _load_configuration(self, infile, objtype=None):
...
-                            self.du.rm(path=fn)
+                            self.du.rm(path=fn, force=True, sudo=True)
+                return True
+            elif objtype == MGR_OBJ_NODE:
+                conf = sconf[str(objtype)]
+                nconf = {}
+                nconf = conf[str(self.hostname)]
+                for k, v in nconf.items():
+                    try:
+                        fn = self.du.create_temp_file()
```
sujatapatnaik52: If `create_temp_file()` fails due to some reason, it will go to `finally` block and may be after the loop executes, it will `return True` I guess, whereas for one of the moms, the `load_configuration` might have failed. Please verify this behaviour.

test/fw/ptl/lib/pbs_testlib.py
saritakh: It was residual part of code added while debugging.

test/fw/ptl/lib/pbs_testlib.py
saritakh: Thanks for catching it; I updated the code by removing the try block. The temp file creation for each mom happens on the local node and the file is then copied to the remote mom node.

saritakh:
Thanks @sujatapatnaik52 for your comments. I have incorporated your comments and tested the same, please have a look and let me know of any comments.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4311,18 +4313,22 @@ def _load_configuration(self, infile, objtype=None):
...
+                        return False
                 return True
             elif objtype == MGR_OBJ_NODE:
-                conf = sconf[str(objtype)]
-                nconf = {}
                 nconf = conf[str(self.hostname)]
                 for k, v in nconf.items():
-                    try:
-                        fn = self.du.create_temp_file()
-                        self.du.chmod(path=fn, mode=0o644)
```
sujatapatnaik52: Thanks @saritakh for making the changes. one suggestion, I see that if we remove the `try` block, there might be exceptions at places like `fd.write`. How about we keep the `try...finally` block as it was and add an `except` block which shall `return False`? Rest looks good to me.

saritakh:
Thanks @sujatapatnaik52

saritakh:
Thank you @bhroam , I have incorporated your comments. Please have a look.

PR #1614, opened 2020-03-27
TestMomMockRun.test_rsc_used fails on cpuset machines

agrawalravi90:
#### Describe Bug or Feature and Changes

The test fails on cpuset machines, it's not really relevant for cpuset machines so I'm just going to add the skipOnCpuSet decorator.

### DIFF
--- a/test/tests/functional/pbs_mom_mock_run.py
+++ b/test/tests/functional/pbs_mom_mock_run.py
@@ -39,6 +39,7 @@ from tests.functional import *
 
 
 class TestMomMockRun(TestFunctional):
+    @skipOnCpuSet
     def test_rsc_used(self):
         """
         Test that resources_used are set correctly by mom under mock run


vstumpf:
I don't see why it should fail on a cpuset machine, it's not like you're changing the ncpus on a node which can't be done.
The accountinglog match fails on resources_used.walltime, but the other accounting log matches are found. So is walltime not being updated?
Also, I found this in the mom logs, is it expected?
```
03/27/2020 03:43:22.313319;0008;pbs_mom;Job;0.uv-02-s12;Started mock run of job
03/27/2020 03:43:22.313356;0001;pbs_mom;Svr;pbs_mom;No such file or directory (2) in job_save, error on open
03/27/2020 03:43:27.318672;0008;pbs_mom;Job;0.uv-02-s12;kill_job
03/27/2020 03:43:27.318708;0100;pbs_mom;Job;0.uv-02-s12;uv-02-s12 cput=00:00:00 mem=5120kb
03/27/2020 03:43:27.318783;0100;pbs_mom;Job;0.uv-02-s12;Obit sent
```

agrawalravi90:
Hey @vstumpf That log message is harmless, it tries to open the file system db on mom side, which doesn't exist because I short-circuited some code for mock run, so basically job_save doesn't happen, which is harmless, but ya should probably get rid of that log message.

Anyways, the reason why it fails is that walltime used was set to: "00:00:06" instead of 00:00:05, so maybe it was just a race condition? I'm not sure. Do you know if mom can be slow on cpuset machines?

vstumpf:
> That log message is harmless

That makes sense

> I'm not sure. Do you know if mom can be slow on cpuset machines?

I can see that being the case, I think we do special things in for cpuset moms.
I'll approve the PR.

PR #1615, opened 2020-03-27
cgroup pbs.server() timeouts using multiprocessing

alexis-cousein:
#### Describe Bug or Feature

The Timeout class is meant to allow long operations to be interrupted so that the cgroup hook abandons what it was doing.

This is currently _only_ used when deciding whether to online vnodes: the hook verifies that the site admin did not change the comment to something else, which would signal that he does not want to online the vnode just yet or that he has other reasons to offline the vnode. The hook ONLY onlines the vnode again if the comment on the vnode is that which was set by the hook.

The server can be unresponsive because it is busy servicing other requests, which can then make the exechost_periodic quasihang. Sadly, it is holding the cgroup lock file, so as a cascading effect this may also make other cgroup operations fail for e.g. excejob_begin, since they may wait for too long and fail to finish before the hook alarm timeout is triggered.

There is a configuration file option "server_timeout" that is meant to interrupt this operation if querying the comments from the server takes too long. But the Timeout class used is relying upon handling of a SIGALRM signal, and sadly the implementation of Python signal handling in Python 2.7 and Python 3.x is such that signals are only checked before the interpreter starts to process a new statement. pbs.server().vnode(XXX).comment is using a PBS IFL C call wrapped in Python using SWIG, so the mechanism to trigger an alarm will not work.

In Python 2.6 and later, multiprocessing can be used to sidestep that limitation. You can fork into two processes (each with their own memory space and Python interpreter) so that a main thread can spawn a worker to fetch the comment and pass it using a multiprocessing queue, and the main thread can abort and kill the child if the child takes too long to exit. 

#### Describe Your Change

The new hook attempts to import multiprocessing; multiprocessing is set to None if this fails.

A wrapper routine is made to fetch all comments for a list of vnodes; it returns a tuple with a dictionary of comments and a boolean indicating failure (or not). 

If multiprocessing support is detected, it will fork a worker thread to contact the server and pass a dictionary with the relevant comments using a multiprocessing.Queue. If the thread fails to exit within server_timeout seconds, the function will indicate failure and return an empty dictionary. If the worker thread has finished before the timeout expires, it will read the comments dictionary from the queue in which the worker has pushed it, and indicate success.

If no multiprocessing support is detected, then it will call a routine that uses the existing Timeout class to attempt to implement a timeout around the pobs.server().vnode(xxx) calls.

### DIFF
--- a/src/hooks/cgroups/pbs_cgroups.PY
+++ b/src/hooks/cgroups/pbs_cgroups.PY
@@ -116,10 +116,19 @@ else:
     import operator
     import fnmatch
     import math
+    import types
     try:
         import json
     except Exception:
         import simplejson as json
+    multiprocessing = None
+    try:
+        # will fail in Python 2.5
+        import multiprocessing
+    except Exception:
+        # but we can use isinstance(multiprocessing, types.ModuleType) later
+        # to find out if it worked
+        pass
     import fcntl
     import pwd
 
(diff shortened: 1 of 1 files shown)


src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -2185,27 +2302,22 @@ class NodeUtils(object):
...
             return
         # Bring vnodes online that this hook has taken offline
         for vnode_name in vnode_comments:
             if vnode_comments[vnode_name] != self.offline_msg:
-                pbs.logmsg(pbs.EVENT_DEBUG, '%s: Comment for vnode %s '
-                           'was not set by this hook' %
-                           (caller_name(), vnode_name))
+                pbs.logmsg(pbs.EVENT_DEBUG, ('%s: Comment for vnode %s '
+                           + 'was not set by this hook')
+                           % (caller_name(), vnode_name))
```
mike0042: This change was unnecessary, but if pycodestyle isn't complaining, then neither will I.

alexis-cousein: Actually, pycodestyle was complaining about the indentation of the 'was not' on my system (hence the change). Don't know what confused it, since splitting "a b" into "a " "b" is common practice.

PR #1617, opened 2020-03-30
test_hostname_add uses local hostname instead of mom name

lisa-altair:
#### Describe Bug or Feature

The TestVerifyLogOutput.test_hostname_add test is testing for the hostname presence in the logs.  However, the test assumes the mom is on the same node as the server.  If the mom is on a separate node, then this test fails.

#### Describe Your Change

Changed the test to use the mom name when searching the logs for a match.

[hostbefore.txt](https://github.com/PBSPro/pbspro/files/4405674/hostbefore.txt)
[hostafter.txt](https://github.com/PBSPro/pbspro/files/4405675/hostafter.txt)

### DIFF
--- a/test/tests/functional/pbs_verify_log_output.py
+++ b/test/tests/functional/pbs_verify_log_output.py
@@ -84,6 +84,7 @@ class TestVerifyLogOutput(TestFunctional):
         Test for hostname presence in log files
         """
         log_val = socket.gethostname()
+        momname = self.moms.values()[0].shortname
         self.scheduler.log_match(
             log_val,
             regexp=False,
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_verify_log_output.py
```diff
@@ -84,6 +84,7 @@ def test_hostname_add(self):
         Test for hostname presence in log files
         """
         log_val = socket.gethostname()
+        momname = self.moms.values()[0].shortname
```
vstumpf: Just a nitpick, can you use self.mom.shortname?

lisa-altair:
@vstumpf I have made your requested change.  Please have another look.

PR #1619, opened 2020-03-31
Refactored the PTL testcase pbs_que_resc_usage.py

Bhagat-Rajput:
#### Describe Bug or Feature
code is not optimized and efficient

#### Describe Your Change

I've done some code refactoring

PTL logs :- [ptl_que_resc_usage.txt](https://github.com/PBSPro/pbspro/files/4406898/ptl_que_resc_usage.txt)
Valgrind logs :- [resc_val_logs.txt](https://github.com/PBSPro/pbspro/files/4406912/resc_val_logs.txt)

### DIFF
--- a/test/tests/functional/pbs_que_resc_usage.py
+++ b/test/tests/functional/pbs_que_resc_usage.py
@@ -82,16 +82,17 @@ class TestQueRescUsage(TestFunctional):
                          q_status[0], self.err_msg)
 
         # set the resource and unset it after using
-        self.server.manager(MGR_CMD_SET, QUEUE, {
-                            'resources_available.ncpus': 8}, id='workq')
+        a = {'resources_available.ncpus': 8}
+        self.server.manager(MGR_CMD_SET, QUEUE, a, id='workq')
         q_status = self.server.status(QUEUE, id='workq')
         self.assertEqual(
             int(q_status[0]['resources_available.ncpus']), 8, self.err_msg)
         # job submission
-        j1 = Job(attrs={'Resource_List.ncpus': '3'})
+        j_attr = {'Resource_List.ncpus': '3'}
+        j1 = Job(attrs=j_attr)
         j1.set_sleep_time(30)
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_que_resc_usage.py
```diff
@@ -71,31 +68,30 @@ def test_resc_assigned_set_unset(self):
...
         q_status = self.server.status(QUEUE, id='workq')
         self.assertNotIn('resources_available.ncpus',
-                         q_status[0].keys(), self.err_msg)
+                         q_status[0], self.err_msg)
 
         # set the resource and unset it after using
-        a = {'resources_available.ncpus': 8}
-        self.server.manager(MGR_CMD_SET, QUEUE, a, id='workq')
+        self.server.manager(MGR_CMD_SET, QUEUE, {
+                            'resources_available.ncpus': 8}, id='workq')
```
bhroam: While both the old and new code are pep8 compliant, I find the old code more readable (here and above)

test/tests/functional/pbs_que_resc_usage.py
```diff
@@ -71,31 +68,30 @@ def test_resc_assigned_set_unset(self):
...
         self.assertEqual(
             int(q_status[0]['resources_available.ncpus']), 8, self.err_msg)
         # job submission
-        j_attr = {ATTR_queue: 'workq', 'Resource_List.select': '1:ncpus=3'}
-        j1 = Job(TEST_USER, j_attr)
+        j1 = Job(attrs={'Resource_List.ncpus': '3'})
         j1.set_sleep_time(30)
         jid_1 = self.server.submit(j1)
-        j2 = Job(TEST_USER, j_attr)
+        j2 = Job(attrs={'Resource_List.ncpus': '3'})
```
bhroam: Since we're submitting the same job twice, setting the values once in a variable and using it twice is more generic.  If the test ever needs to be changed, we would change the value once.

PR #1627, opened 2020-04-01
Fix tests of TestPbsHookAlarmLargeMultinodeJob

shilpakodli:
#### Describe Bug or Feature
Tests of TestPbsHookAlarmLargeMultinodeJob failed on slow machines due to less alarm value of execjob_* hooks where before Job goes into R state, alarm would trigger and Job gets requeued and test fails with incorrect Job state.

#### Describe Your Change
Since, we are creating large number of vnodes(5000 vnodes) and submitting job on 5000 vnode setup, Job is taking time from changing it's state from Q->R. Increased max attempts to check Job state in R and also increased alarm value from 15 secs to 20 secs.

[TestPbsHookAlarmLargeMultinodeJob_after_fix.txt](https://github.com/PBSPro/pbspro/files/4412783/TestPbsHookAlarmLargeMultinodeJob_after_fix.txt)
[TestPbsHookAlarmLargeMultinodeJob_before_fix.txt](https://github.com/PBSPro/pbspro/files/4412784/TestPbsHookAlarmLargeMultinodeJob_before_fix.txt)

### DIFF
--- a/test/tests/resilience/pbs_hook_alarm_large_multinode_job.py
+++ b/test/tests/resilience/pbs_hook_alarm_large_multinode_job.py
@@ -153,7 +153,7 @@ pbs.logmsg(pbs.LOG_DEBUG, "executing epilogue hook %s" % (e.hook_name,))
 """
         search_after = int(time.time())
         a = {'event': hook_event, 'enabled': 'True',
-             'alarm': '20'}
+             'alarm': '30'}
         self.server.create_import_hook(hook_name, a, hook_body)
 
         j = Job(TEST_USER)
--- a/test/tests/resilience/pbs_hook_alarm_large_multinode_job.py
+++ b/test/tests/resilience/pbs_hook_alarm_large_multinode_job.py
@@ -64,7 +64,7 @@ class TestPbsHookAlarmLargeMultinodeJob(TestResilience):
         # Restart mom explicitly due to PP-993
         self.mom.restart()
 
-    @timeout(400)
+    @skipOnCpuSet
     def test_begin_hook(self):
         """
         Create an execjob_begin hook, import a hook content with a small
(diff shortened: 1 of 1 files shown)


test/tests/resilience/pbs_hook_alarm_large_multinode_job.py
```diff
@@ -164,7 +164,7 @@ def test_epi_hook(self):
 
         jid = self.server.submit(j)
         self.server.expect(JOB, {'job_state': 'R'},
-                           jid, max_attempts=15, interval=2)
+                           jid, max_attempts=10, interval=2)
```
prakashcv13, line 167: shouldn't this be 20?

test/tests/resilience/pbs_hook_alarm_large_multinode_job.py
```diff
@@ -153,7 +153,7 @@ def test_epi_hook(self):
 """
         search_after = int(time.time())
         a = {'event': hook_event, 'enabled': 'True',
-             'alarm': '15'}
+             'alarm': '20'}
```
prakashcv13, line 156: do you think if we made it 30, we will be on the safer side?

test/tests/resilience/pbs_hook_alarm_large_multinode_job.py
shilpakodli: Hi Prakash, this is max attempts when Job comes to R state. I have tested on slow machines and is working fine with 10 attempts. I have given interval as 2 so that's equivalent to 20 seconds.

shilpakodli:
@prakashcv13 I have addressed your review comments. Please review.

shilpakodli:
@prakashcv13 As discussed offline, I have made changes to make it run on cpuset mom. Please review.
[TestPbsHookAlarmLargeMultinodeJob_nonuv.txt](https://github.com/PBSPro/pbspro/files/4420479/TestPbsHookAlarmLargeMultinodeJob_nonuv.txt)
[TestPbsHookAlarmLargeMultinodeJob_uv.txt](https://github.com/PBSPro/pbspro/files/4420480/TestPbsHookAlarmLargeMultinodeJob_uv.txt)

PR #1630, opened 2020-04-01
Accounting records for Job Suspend and resume events

sujatapatnaik52:
#### Describe Bug or Feature
Accounting of suspend and resume events of a job for monitoring resources throughout the job's lifecycle. 

#### Describe Your Change
1. Added a new 'z' record for the suspend event which has few job resources that are not static across other records - requestor, resources_released, resources_used. 
2. Added a new 'r' record for the resume event and accounting the requestor for the same. 

#### Link to Design Doc
[Design document](http://community.pbspro.org/t/new-accounting-record-for-suspend-and-resume-event/1917/2
)

[valgrind_logs.txt](https://github.com/PBSPro/pbspro/files/4415019/valgrind_logs.txt)
[automated_test_logs.txt](https://github.com/PBSPro/pbspro/files/4415021/automated_test_logs.txt)

### DIFF
--- a/src/server/accounting.c
+++ b/src/server/accounting.c
@@ -317,7 +317,8 @@ cpy_quote_value(char *pb, char *value)
  * Get the resources_used job attribute
  *
  * @param[in]	pjob	- pointer to job structure
- * @param[in]	res	- resource entity index (e.g. JOB_ATR_resource)
+ * @param[in]	resc_used - pointer to resources used string
+ * @param[in]	resc_used_size - size of resources used string
  *
  * @return	int
  * @retval	0 upon success
@@ -2297,21 +2298,32 @@ void
 log_suspend_resume_record(job *pjob, char *user, char *host, int acct_type)
 {
 	char *buf = NULL;
-	char *tmp;
 	char *resc_used;
+	char *resc_released;
+	int buf_size = 1024;
 	int resc_used_size = 0;
-	int rc;
 
-	rc = pbs_asprintf(&buf, "requestor=%s@%s", user, host);
+	buf = malloc(buf_size * sizeof(char));
+	if (buf == NULL)
+		return;
+
+	buf[0] = '\0';
+
+	snprintf(buf, buf_size, "requestor=%s@%s", user, host);
 
 	if (acct_type == PBS_ACCT_SUSPEND) {
 		
 		if (pjob->ji_wattr[JOB_ATR_resc_released].at_flags & ATR_VFLAG_SET) {
-			tmp = buf;
-			rc = pbs_asprintf(&buf, "%s resources_released=%s", tmp,
-			pjob->ji_wattr[JOB_ATR_resc_released].at_val.at_str);
-			free(tmp);
-		}	
+			if (pbs_asprintf(&resc_released, " resources_released=%s",
+			pjob->ji_wattr[JOB_ATR_resc_released].at_val.at_str) > 0) {
+				if (pbs_strcat(&buf, &buf_size, resc_released) == NULL) {
+					free(buf);
+					free(resc_released);
+					return;
+				}
+				free(resc_released);
+			}
+		}
 
 		/* Allocate initial space for resc_used.  Future space will be allocated by pbs_strcat(). */
 		resc_used = malloc(RESC_USED_BUF_SIZE);
@@ -2327,16 +2339,16 @@ log_suspend_resume_record(job *pjob, char *user, char *host, int acct_type)
 		}
 
 		if (resc_used != NULL) {
-			tmp = buf;
-			rc = pbs_asprintf(&buf, "%s%s", tmp, resc_used);
+			if (pbs_strcat(&buf, &buf_size, resc_used) == NULL) {
+				free(buf);
+				free(resc_used);
+				return;
+			}
 			free(resc_used);
-			free(tmp);
 		}
 	}
 
 writeit:
-	if (rc > 0) {
-		write_account_record(acct_type, pjob->ji_qs.ji_jobid, buf);
-		free(buf);
-	}
+	write_account_record(acct_type, pjob->ji_qs.ji_jobid, buf);
+	free(buf);
 }
--- a/src/server/accounting.c
+++ b/src/server/accounting.c
@@ -2298,7 +2298,6 @@ log_suspend_resume_record(job *pjob, char *user, char *host, int acct_type)
 {
 	if (acct_type == PBS_ACCT_SUSPEND) {
 		char *resc_buf;
-		char *resc_released;
 		int resc_buf_size = RESC_USED_BUF_SIZE;
 
 		/* Allocating initial space as required by resc_used. Future space will be allocated by pbs_strcat(). */
--- a/src/server/accounting.c
+++ b/src/server/accounting.c
@@ -2309,19 +2309,18 @@ log_suspend_resume_record(job *pjob, char *user, char *host, int acct_type)
 
 		if (get_resc_used(pjob, resc_buf, resc_buf_size) < 0) {
 			write_account_record(acct_type, pjob->ji_qs.ji_jobid, NULL);
-			goto writeit;
+			goto end;
 		}
 
 		if (pjob->ji_wattr[JOB_ATR_resc_released].at_flags & ATR_VFLAG_SET) {
-			if (pbs_strcat(&resc_buf, &resc_buf_size, " resources_released=") == NULL) {
-				goto writeit;
-			}
-			if (pbs_strcat(&resc_buf, &resc_buf_size, pjob->ji_wattr[JOB_ATR_resc_released].at_val.at_str) == NULL) {
-				goto writeit;
-			}
+			if (pbs_strcat(&resc_buf, &resc_buf_size, " resources_released=") == NULL)
+				goto end;
+
+			if (pbs_strcat(&resc_buf, &resc_buf_size, pjob->ji_wattr[JOB_ATR_resc_released].at_val.at_str) == NULL)
+				goto end;
 		}
 		write_account_record(acct_type, pjob->ji_qs.ji_jobid, resc_buf + 1);
-writeit:
+end:
 	free(resc_buf);
 	return;
 	}
--- a/src/server/accounting.c
+++ b/src/server/accounting.c
@@ -312,6 +312,46 @@ cpy_quote_value(char *pb, char *value)
 /* start attributes */
 #define ACCTBUF_TIMES_NEED	72
 
+/**
+ * @brief
+ * Get the resources_used job attribute
+ *
+ * @param[in]	pjob	- pointer to job structure
+ * @param[in]	res	- resource entity index (e.g. JOB_ATR_resource)
+ *
+ * @return	int
+ * @retval	0 upon success
+ * @retval	-1	if error encountered.
+ *
+ */
+static int
+get_resc_used(job *pjob, char *resc_used, int resc_used_size) {
+
+	struct svrattrl *patlist = NULL;
+	pbs_list_head temp_head;
+	CLEAR_HEAD(temp_head);
+
+	if (pjob->ji_wattr[(int) JOB_ATR_resc_used].at_user_encoded != NULL)
+		patlist = pjob->ji_wattr[(int) JOB_ATR_resc_used].at_user_encoded;
+	else if (pjob->ji_wattr[(int) JOB_ATR_resc_used].at_priv_encoded != NULL)
+		patlist = pjob->ji_wattr[(int) JOB_ATR_resc_used].at_priv_encoded;
+	else
+		encode_resc(&pjob->ji_wattr[(int) JOB_ATR_resc_used],
+		&temp_head, job_attr_def[(int) JOB_ATR_resc_used].at_name,
+		NULL, ATR_ENCODE_CLIENT, &patlist);
+
+	while(patlist) {
+		/* log to accounting_logs only if there's a value */
+		if (strlen(patlist->al_value) > 0) {
+			if (concat_rescused_to_buffer(&resc_used, &resc_used_size, patlist, " ", NULL) != 0) {
+				return -1;
+			}
+		}
+		patlist = patlist->al_sister;
+	}
+	free_attrlist(&temp_head);
+	return 0;
+}
 
 /**
  * @brief
@@ -1194,12 +1234,9 @@ account_jobend(job *pjob, char *used, int type)
 	int i = 0;
 	int len = 0;
 	char *pb = NULL;
-	struct svrattrl *patlist = NULL;
 	char *resc_used;
 	int resc_used_size = 0;
-	pbs_list_head temp_head;
 
-	CLEAR_HEAD(temp_head);
 	/* pack in general information about the job */
 
 	pb = acct_job(pjob, type, acct_buf, acct_bufsize);
@@ -1264,14 +1301,6 @@ account_jobend(job *pjob, char *used, int type)
 			 * So we try to derive the resource usage information from resources_used attribute of
 			 * the job and then reconstruct the resources usage information into resc_used buffer.
 			 */
-			if (pjob->ji_wattr[(int) JOB_ATR_resc_used].at_user_encoded != NULL)
-				patlist = pjob->ji_wattr[(int) JOB_ATR_resc_used].at_user_encoded;
-			else if (pjob->ji_wattr[(int) JOB_ATR_resc_used].at_priv_encoded != NULL)
-				patlist = pjob->ji_wattr[(int) JOB_ATR_resc_used].at_priv_encoded;
-			else
-				encode_resc(&pjob->ji_wattr[(int) JOB_ATR_resc_used],
-					    &temp_head, job_attr_def[(int) JOB_ATR_resc_used].at_name,
-					    NULL, ATR_ENCODE_CLIENT, &patlist);
 
 			/* Allocate initial space for resc_used.  Future space will be allocated by pbs_strcat(). */
 			resc_used = malloc(RESC_USED_BUF_SIZE);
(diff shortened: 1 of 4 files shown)


src/server/accounting.c
```diff
@@ -2259,3 +2259,86 @@ void log_alter_records_for_attrs(job *pjob, svrattrl *plist) {
...
+ * 
+ * @param[in] pjob - job to log records for.
+ * @param[in] user - User who requested the signal suspend/resume 
+ * @param[in] host - Hostname where the signal was triggered
+ * @param[in] signal_type - flag indicating the type of event suspend/resume
+ * 
+ * @returns void 
+ */
+void
+log_suspend_resume_record(job *pjob, char *user, char *host, char *signal_type)
```
agrawalravi90, line 2277: instead of a char* signal_type, how about a "int acct_type", which can be passed PBS_ACCT_RESUME or  PBS_ACCT_SUSPEND? int comparisons are faster than string comparisons

src/server/accounting.c
```diff
@@ -2259,3 +2259,86 @@ void log_alter_records_for_attrs(job *pjob, svrattrl *plist) {
...
+		acct_type = PBS_ACCT_SUSPEND;
+		
+		if (pjob->ji_wattr[JOB_ATR_resc_released].at_flags & ATR_VFLAG_SET) {
+			tmp = log_record;
+			rc = pbs_asprintf(&log_record, "%s resources_released=%s", tmp, 
+			pjob->ji_wattr[JOB_ATR_resc_released].at_val.at_str);
+			free(tmp);
+		}	
+
+		CLEAR_HEAD(temp_head);
```
agrawalravi90, line 2300: Since most of the following code for creating resources_used string for accounting is common with the E record construction, put it in a separate function to avoid code duplication?

test/tests/functional/pbs_suspend_resume_accounting.py
```diff
@@ -0,0 +1,356 @@
...
+#
+# Use of Altair’s trademarks, including but not limited to "PBS™",
+# "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
+# trademark licensing policies.
+
+
+from tests.functional import *
+
+
+class TestSuspendResumeAccounting(TestFunctional):
```
agrawalravi90, line 42: There's a TestAcctLog test suite, maybe add these tests there instead of creating a new test suite?

test/tests/functional/pbs_suspend_resume_accounting.py
```diff
@@ -0,0 +1,356 @@
...
+        self.server.expect(JOB, {ATTR_state: 'R'}, id=jid)
+
+        self.server.sigjob(jobid=jid, signal="suspend")
+
+        record = 'z;%s;requestor=%s@%s .*resources_used.' % \
+                 (jid, self.server.current_user, self.server.hostname)
+        self.server.accounting_match(msg=record, id=jid, regexp=True)
+
+        self.server.sigjob(jobid=jid, signal="resume")
+        record = 'r;%s;requestor=Scheduler@%s' % \
```
agrawalravi90, line 71: Why is the requestor "Scheduler" if the user is resuming it via qsig ?

test/tests/functional/pbs_suspend_resume_accounting.py
```diff
@@ -0,0 +1,356 @@
...
+        j2.create_script(self.script)
+        j2.set_attributes(
+            {ATTR_l + '.select': '1:ncpus=2:mem=512mb',
+             ATTR_q: 'expressq'})
+        jid2 = self.server.submit(j2)
+
+        self.server.expect(JOB, {ATTR_state: 'R'}, id=jid2)
+        self.server.expect(JOB, {ATTR_state: 'S', ATTR_substate: 45}, id=jid1)
+
+        record = 'z;%s;requestor=Scheduler@%s .*resources_used.' % \
```
agrawalravi90, line 179: Since you have RRTROS set, check for resources_released as well here

test/tests/functional/pbs_suspend_resume_accounting.py
agrawalravi90: Actually, looking at how many tests you wrote, keep the testsuite and ignore my comment

test/tests/functional/pbs_suspend_resume_accounting.py
```diff
@@ -0,0 +1,356 @@
...
+    script += ['sleep 1']
+    script += ['done']
+
+    def test_suspend_resume_job_signal(self):
+        """
+        Test case to verify accounting suspend
+        and resume records when the events are
+        triggered by client for one job.
+        """
+        j = Job(TEST_USER)
```
bhroam, line 59: No need to say TEST_USER, it is the default. (here and other places in the test plan)

test/tests/functional/pbs_suspend_resume_accounting.py
```diff
@@ -0,0 +1,356 @@
...
+        self.server.manager(MGR_CMD_SET, NODE, a, self.mom.shortname)
+
+        j = Job(TEST_USER)
+        j.create_script(self.script)
+        j.set_attributes({ATTR_J: '1-2'})
+        jid = self.server.submit(j)
+
+        subjobs = self.server.status(JOB, id=jid, extend='t')
+        sub_jid1 = subjobs[1]['id']
+        sub_jid2 = subjobs[2]['id']
```
bhroam, line 92: You can use create_subjob_id() instead if you want.  It'll save the status() call.

test/tests/functional/pbs_suspend_resume_accounting.py
```diff
@@ -0,0 +1,356 @@
...
+        self.server.sigjob(jobid=jid, signal="resume")
+        record = 'r;%s;requestor=Scheduler@%s' % \
+                 (sub_jid1, self.server.hostname)
+        self.server.accounting_match(msg=record, id=sub_jid1)
+
+        record = 'r;%s;requestor=Scheduler@%s' % \
+                 (sub_jid2, self.server.hostname)
+        self.server.accounting_match(msg=record, id=sub_jid2)
+
+    def test_interactive_job_suspend_resume(self):
```
bhroam, line 118: Why does it matter if a job is interactive or batch for suspend resume?

test/tests/functional/pbs_suspend_resume_accounting.py
bhroam: When a user does a qsig -s resume, it moves the job from suspended substate 43 (user suspend) to suspended substate 45 (scheduler suspended), but does not resume the job.  It is the scheduler that does the resume when resources are available.

sujatapatnaik52: Thanks @bhroam. I guess the requestor is scheduler here because I am logging after the resume was successful by the scheduler i.e. post_signal processing. Should we also record user requested resume?

src/server/accounting.c
```diff
@@ -2259,3 +2259,86 @@ void log_alter_records_for_attrs(job *pjob, svrattrl *plist) {
...
+	int acct_type = PBS_ACCT_RESUME;
+	int rc;
+
+	rc = pbs_asprintf(&log_record, "requestor=%s@%s", user, host);
+
+	if (strcmp(signal_type, SIG_SUSPEND)==0 || strcmp(signal_type, SIG_ADMIN_SUSPEND) == 0) {
+		acct_type = PBS_ACCT_SUSPEND;
+		
+		if (pjob->ji_wattr[JOB_ATR_resc_released].at_flags & ATR_VFLAG_SET) {
+			tmp = log_record;
```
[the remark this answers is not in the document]
sujatapatnaik52: Hi @bhroam, I am using `tmp` basically to copy the initial string to the newly allocated memory by `pbs_asprintf` into `log_record`. If I free `log_record` here, I will lose the previously stored string. I hope my understanding is correct.

test/tests/functional/pbs_suspend_resume_accounting.py
agrawalravi90: if the requester is always going to be Scheduler for resume record then there isn't really a lot of value logging it.

sujatapatnaik52: In case of `admin-resume`, the job is resumed directly, without waiting for the scheduler and it could be a different user. So, maybe I think we should keep it.

The user requested resume is kind of "attempted-resume" which may or may not succeed always. What if we have an additional record to say that there was a user requested resume?

test/tests/functional/pbs_suspend_resume_accounting.py
sujatapatnaik52: I was actually verifying for all the scenarios of job suspend. so added for interactive job as well. If it's not really necessary, I can remove it. Please let me know.

test/tests/functional/pbs_suspend_resume_accounting.py
agrawalravi90: nah, that can be confusing. Since the request to resume was initiated by the user, I think the right thing to do is to print the actual user's name as the requester, even if it happens via the Scheduler. So, if possible, I think we should print the user's name as the requester.

What do you think @bhroam  ?

sujatapatnaik52:
[updated_logs.txt](https://github.com/PBSPro/pbspro/files/4420357/updated_logs.txt)

sujatapatnaik52:
@agrawalravi90, @bhroam I have addressed most of the review suggestions. Please look at the latest commit.

src/server/accounting.c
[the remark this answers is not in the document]
bhroam: Yes, you will lose the previously stored value, but you do anyway when you free(tmp) 2 lines down.  There isn't really isn't any difference between freeing it here and freeing it 2 lines down because you don't do anything with the value in between.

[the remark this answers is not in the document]
bhroam: Oh, I see, you use tmp IN the pbs_asprintf() call.  That is dangerous because you are now setting and using the same value in the same call.  You don't know what pbs_asprintf() is going to do.  You are probably better off doing 2 calls to pbs_strcat().

src/server/accounting.c
```diff
@@ -2274,40 +2284,25 @@ void log_alter_records_for_attrs(job *pjob, svrattrl *plist) {
...
 
-	if (strcmp(signal_type, SIG_SUSPEND)==0 || strcmp(signal_type, SIG_ADMIN_SUSPEND) == 0) {
-		acct_type = PBS_ACCT_SUSPEND;
+	if (acct_type == PBS_ACCT_SUSPEND) {
 		
 		if (pjob->ji_wattr[JOB_ATR_resc_released].at_flags & ATR_VFLAG_SET) {
-			tmp = log_record;
-			rc = pbs_asprintf(&log_record, "%s resources_released=%s", tmp, 
+			tmp = buf;
+			rc = pbs_asprintf(&buf, "%s resources_released=%s", tmp,
```
bhroam, line 2311: Don't use tmp and buf in the same call to pbs_asprintf().  Use snprintf() to write into a local buffer and then pbs_strcat() to concatenate it onto the final string.

test/tests/functional/pbs_suspend_resume_accounting.py
agrawalravi90: We could argue the same for job arrays as well right? I think for the sake of completeness, it's good to keep this test as well.

src/server/accounting.c
agrawalravi90: I like the idea of using pbs_strcat as well. It might also help reduce the number of mallocs, each call of pbs_asprintf() does an unconditional malloc, whereas pbs_strcat will do it only when the current buffer runs out.

sujatapatnaik52:
By mistake clicked on `close and comment`. Reopened the PR.

sujatapatnaik52:
[valgrind_logs_after_review.txt](https://github.com/PBSPro/pbspro/files/4426854/valgrind_logs_after_review.txt)
[updated_logs.txt](https://github.com/PBSPro/pbspro/files/4426855/updated_logs.txt)

sujatapatnaik52:
@bhroam, @agrawalravi90, I am making some changes to this PR. Please review after I push my changes.

src/server/accounting.c
```diff
@@ -2297,21 +2298,32 @@ void
 log_suspend_resume_record(job *pjob, char *user, char *host, int acct_type)
```
bhroam, line 2298: Update param signal_type above

src/server/accounting.c
```diff
@@ -2297,21 +2298,32 @@ void
...
 
 	if (acct_type == PBS_ACCT_SUSPEND) {
 		
 		if (pjob->ji_wattr[JOB_ATR_resc_released].at_flags & ATR_VFLAG_SET) {
-			tmp = buf;
-			rc = pbs_asprintf(&buf, "%s resources_released=%s", tmp,
-			pjob->ji_wattr[JOB_ATR_resc_released].at_val.at_str);
-			free(tmp);
-		}	
+			if (pbs_asprintf(&resc_released, " resources_released=%s",
```
bhroam, line 2317: Consider doing two calls to pbs_strcat() instead of calling pbs_asprintf().  It will allocate memory which is not necessary.  You are only concatenating two strings together.  If it was more than that, pbs_asprintf() would be better.  In this case, I think the two calls to pbs_strcat() is better.

sujatapatnaik52: Hi @bhroam, there is a change for removal of requestor attribute from the log records. Also, I have addressed your above comment in the latest commit. Please have a look at it.

agrawalravi90:
> @bhroam, @agrawalravi90, I am making some changes to this PR. Please review after I push my changes.

Can you please mark your PR "WIP" while you are working on it?

sujatapatnaik52:
> > @bhroam, @agrawalravi90, I am making some changes to this PR. Please review after I push my changes.
> 
> Can you please mark your PR "WIP" while you are working on it?

Hi @agrawalravi90, Please review the changes of the latest commit. I have pushed them.

src/server/accounting.c
```diff
@@ -312,6 +312,56 @@ cpy_quote_value(char *pb, char *value)
...
+ * @param[in]	resc_used - pointer to resources used string
+ * @param[in]	resc_used_size - size of resources used string
+ *
+ * @return	int
+ * @retval	0 upon success
+ * @retval	-1	if error encountered.
+ *
+ */
+static int
+get_resc_used(job *pjob, char *resc_used, int resc_used_size) {
```
agrawalravi90, line 329: Nit: put the function's starting brace '{' on its own line

agrawalravi90, line 329: Also, what's the purpose of 'resc_used_size' ? It seems like it's an output variable. Do you intend to store the size of the resources_used string here? If yes, then this should be an int pointer right?

agrawalravi90: same comment for *resc_used, it should be **resc_used since concat_rescused_to_buffer can realloc it

src/server/accounting.c
```diff
@@ -1284,33 +1323,14 @@ account_jobend(job *pjob, char *used, int type)
...
-				if (strlen(patlist->al_value) > 0) {
-					if (concat_rescused_to_buffer(&resc_used, &resc_used_size, patlist, " ", NULL) != 0) {
-						free(resc_used);
-						goto writeit;
-					}
-
-				}
-
-				patlist = patlist->al_sister;
+			if (get_resc_used(pjob, resc_used, resc_used_size) < 0) {
```
agrawalravi90, line 1326: check for -1 instead of < 0

src/server/accounting.c
```diff
@@ -2259,3 +2279,51 @@ void log_alter_records_for_attrs(job *pjob, svrattrl *plist) {
...
+		int resc_buf_size = RESC_USED_BUF_SIZE;
+
+		/* Allocating initial space as required by resc_used. Future space will be allocated by pbs_strcat(). */
+		resc_buf = malloc(RESC_USED_BUF_SIZE);
+		if (resc_buf == NULL)
+			return;
+
+		resc_buf[0] = '\0';
+
+		if (get_resc_used(pjob, resc_buf, resc_buf_size) < 0) {
```
agrawalravi90, line 2310: same here, check for -1 instead of < 0

src/server/accounting.c
```diff
@@ -2259,3 +2279,51 @@ void log_alter_records_for_attrs(job *pjob, svrattrl *plist) {
...
+			write_account_record(acct_type, pjob->ji_qs.ji_jobid, NULL);
+			goto end;
```
agrawalravi90, line 2312: Labels and goto are considered error prone/bad practice (https://stackoverflow.com/questions/3517726/what-is-wrong-with-using-goto) unless used in a very obvious way, like at the end of the function as an error label. In this case, your label is inside an inner block, so I'd recommend not using it. You actually are not saving any extra lines of code anyways, your label is 3 lines long including the label, and it saves exactly 3 free(resc_buf); lines, so it's not valuable either. So, please remove it.

src/server/accounting.c
sujatapatnaik52: Thank you @agrawalravi90 for catching this.

sujatapatnaik52:
[updated_logs.txt](https://github.com/PBSPro/pbspro/files/4439466/updated_logs.txt)
[valgrind_logs_review.txt](https://github.com/PBSPro/pbspro/files/4439468/valgrind_logs_review.txt)

sujatapatnaik52:
> LGTM, just one thing: can you please attach the latest PTL test logs?

Uploaded Ptl test logs and valgrind logs.

sujatapatnaik52:
Thank you @agrawalravi90 for the detailed review and sign off. PR is ready for merge.

mike0042:
There have been changes since @bhroam signed off. I'll ask that he reconfirm the signoff.

sujatapatnaik52:
Thanks @mike0042.

sujatapatnaik52:
Thank you @bhroam

PR #1633, opened 2020-04-02
Support for ARM Ubuntu18.04

kjakkali:
#### Describe Bug or Feature
Adding support for ARM-Ubuntu 

#### Describe Your Change
Updated zlib macro to look for library in /usr/lib/aarch64-linux-gnu location


kjakkali:
Before fix : 
  configure: error: libz shared object library not found.
After fix : 
  configure:19219: checking for libz
  configure:19252: result: /lib64

kjakkali:
Thanks @hirenvadalia , @prakashcv13

PR #1639, opened 2020-04-06
TestSoftWalltime.test_soft_before_dedicated failing due to incorrect Job comment

shilpakodli:
#### Describe Bug or Feature
TestSoftWalltime.test_soft_before_dedicated test is failing on few machines with incorrect Job comment. 
Test is submitting a Job with softwalltime which won't complete before dedicated time but on slow machines, Job submission time and dedicated start time are conflicting and Job's comment is resulting to "Not Running: Dedicated time conflict".

#### Describe Your Change
Increased dedicated time by 20 seconds more to avoid race conditions.

[test_soft_before_dedicated_after_fix.txt](https://github.com/PBSPro/pbspro/files/4439817/test_soft_before_dedicated_after_fix.txt)
[test_soft_before_dedicated_before_fix.txt](https://github.com/PBSPro/pbspro/files/4439818/test_soft_before_dedicated_before_fix.txt)

### DIFF
--- a/test/tests/functional/pbs_soft_walltime.py
+++ b/test/tests/functional/pbs_soft_walltime.py
@@ -380,7 +380,7 @@ e.accept()
         """
 
         now = int(time.time())
-        self.scheduler.add_dedicated_time(start=now + 60, end=now + 2500)
+        self.scheduler.add_dedicated_time(start=now + 80, end=now + 2500)
 
         J = Job(TEST_USER)
         jid = self.server.submit(J)
--- a/test/tests/functional/pbs_soft_walltime.py
+++ b/test/tests/functional/pbs_soft_walltime.py
@@ -383,6 +383,7 @@ e.accept()
         self.scheduler.add_dedicated_time(start=now + 120, end=now + 2500)
 
         J = Job(TEST_USER)
+        J.set_sleep_time(200)
         jid = self.server.submit(J)
         self.server.alterjob(jid, {'Resource_List.soft_walltime': 180})
         comment = 'Not Running: Job would cross dedicated time boundary'
--- a/test/tests/functional/pbs_soft_walltime.py
+++ b/test/tests/functional/pbs_soft_walltime.py
@@ -380,11 +380,11 @@ e.accept()
         """
 
         now = int(time.time())
-        self.scheduler.add_dedicated_time(start=now + 80, end=now + 2500)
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_soft_walltime.py
```diff
@@ -380,7 +380,7 @@ def test_soft_before_dedicated(self):
         """
 
         now = int(time.time())
-        self.scheduler.add_dedicated_time(start=now + 60, end=now + 2500)
+        self.scheduler.add_dedicated_time(start=now + 80, end=now + 2500)
```
dinesh1306, line 383: @shilpakodli , this is not due to slow machine. If you check before fix result when PTL set 'Resource_List.soft_walltime', dedicated time started. you increased start time from 60 sec. to 80 sec. and if you check dedicated start time not affecting anymore. your test pass because when PTL set 'Resource_List.soft_walltime' dedicated time not started.

shilpakodli: Thanks Dinesh for catching this. I have addressed your comments. please review.

shilpakodli:
@dinesh1306 I have change Job sleep time as discussed offline. 
[test_soft_before_dedicated_after_comments.txt](https://github.com/PBSPro/pbspro/files/4442522/test_soft_before_dedicated_after_comments.txt)

shilpakodli:
> I don't understand the new changes. This test works because if we ran the job, the soft walltime of the job would run into the reservation. The message reported in the description is what happens when you are in dedicated time and try to run a job outside of a dedicated time queue.
> 
> The expect() should time out after 30s, so we shouldn't even get to the original dedicated time of 60s.
> 
> How does extending things even further fix this issue? Once again, it shouldn't hurt anything. It still qualifies for what the test is trying to do. Submit a job with a soft_walltime which can't run before dedicated time.
> 
> In any case, the expect doesn't have id=jid, so I'm not exactly sure what what job it is checking for. I guess there is only one, so that has to be the one. Could you add an id=jid to it?

Hi Bhroam, I have realized that previous changes are not really affecting dedicated time start. Increase in seconds will remain same. We need to increase in minutes, that's why I have changed from 60 to 120 in dedicated start time and thus softwalltime of Jobs aswelll.
I have addressed your comments of including id=jid in expect.

PR #1640, opened 2020-04-06
runone test fails on slow machines and cpuset machines

arungrover:
#### Describe Bug or Feature
Some runone dependency tests fail on cpuset machines and machines which are really slow.
There is also a small memory leak of the dependency string when parse_runone_job_list fails.

#### Describe Your Change
Some of the tests need to be skipped on cpuset machine because the jobs in some tests need some special nodes (vnodes created during setup) but the mom rejects these jobs because it cannot recognize these vnodes.
The second problem is that some tests submit small jobs and on some slow machines, by the time the test submits these small jobs and then do a qstat on them, the jobs finish. I've increased the sleep time of such jobs.
Addressed a memory leak.

#### Link to Design Doc
NA

[test.txt](https://github.com/PBSPro/pbspro/files/4447185/test.txt)

### DIFF
--- a/src/scheduler/job_info.c
+++ b/src/scheduler/job_info.c
@@ -5396,8 +5396,10 @@ static char **parse_runone_job_list(char *depend_val) {
 	    depend_str = string_dup(depend_val);
 
 	start = strstr(depend_str, depend_type);
-	if (start == NULL)
+	if (start == NULL) {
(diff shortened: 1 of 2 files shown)


arungrover:
Tests on cpuset machine have passed. This PR is open for reviews.

PR #1641, opened 2020-04-07
Database code refactoring.

ashwathraop:
#### Describe Bug or Feature

Code refactor for database related code within PBS.

#### Describe Your Change

For easy maintainability, current database-related code within the PBS server needs to be refactored into a separate pluggable dynamic library. Another motivation behind this refactoring is the added advantage of moving to any database without having to do major code changes within PBS itself.

#### Link to Design Doc

https://pbspro.atlassian.net/wiki/spaces/PD/pages/1524563969/DB+Refactor+Design.

Tests info is attached in the comments below.

P.S: This is the first set of refactoring related to the database. This will be followed by another PR for the cleanup of unused attributes and MoM dead code.


src/lib/Libdb/pgsql/pbs_db_utility
```diff
@@ -0,0 +1,1302 @@
...
+change_locale=0
+opt_err=1
+opt="$1"
+
+#---------------------------------------------------------------------------------------------------
+# Schema handling functions
+
+generate_schema_file () {
+	cat <<-EOF > ${schema}
+/*
```
subhasisb, line 57: I feel we should keep the schema file separate, so that it can be edited and loaded by other tools for schema analysis. There is particularly no need to embed this into this script. So, for postgres, you could still have a db_schema.sql file to easy. Besides the schema file included in this makes pbs_db_utility a really large file

src/lib/Libdb/pgsql/pbs_db_utility
```diff
@@ -0,0 +1,1302 @@
...
+    ji_jobid    TEXT       NOT NULL,
+    script      TEXT
+);
+CREATE INDEX job_scr_idx ON pbs.job_scr (ji_jobid);
+
+---------------------- END OF SCHEMA -----------------------
+EOF
+}
+
+upgrade_pbs_schema_from_v1_0_0() {
```
subhasisb, line 231: We should take a look at which upgrade routines we need. 
1. We need to only support upgrades from last 2 versions of PBS. So, we can eliminate very old schema upgrade functions.
2. We can maintain just 2 files for upgrades from last 2 versions of PBS. upragde_minus_one.sql and upgrade_minus_two.sql, or similarly named. So we keep only two file that pbs_db_utility will refer to and fail in other cases. What say?

src/lib/Libdb/pgsql/pbs_db_utility
```diff
@@ -0,0 +1,1302 @@
...
+	err=`cp -p ${datastore_dir}/pg_hba.conf.orig ${datastore_dir}/pg_hba.conf 2>&1`
+	if [ $? -eq 0 ]; then
+		rm -f ${datastore_dir}/pg_hba.conf.orig
+	else
+		echo "${err}"
+		return 1
+	fi
+}
+
+utf8_to_sql_ascii() {
```
subhasisb, line 656: Which was our last utf8 database distro version - was it within the last 2 versions of PBS? If not, we should remove this code.

src/lib/Libdb/pgsql/pbs_db_utility
```diff
@@ -0,0 +1,1302 @@
...
+	set_db_trust_login "${old_data_dir}"
+	ret=$?
+	if [ $ret -ne 0 ]; then
+		return $ret
+	fi
+
+	CWD=`pwd`
+	cd "${data_dir}"
+	#Perform pg_upgrade -c to check if we can upgrade the cluster or not
+	err=`su ${user} -c "/bin/sh -c '${PGSQL_LIBSTR} ${inst_dir}/bin/pg_upgrade -b ${old_inst_dir}/bin -B ${inst_dir}/bin -d ${old_data_dir} -D ${data_dir} -c'" 2>&1`
```
subhasisb, line 875: When we move to PG 10 in the next few months, pg_upgrade will not work, and we will have to do a dump and restore upgrade to support overlay upgrades. Is this current code structure good for such a upgrade?

src/lib/Libdb/pgsql/db_attr.c
```diff
@@ -40,7 +40,7 @@
 
 
 /**
- * @file    db_postgres_attr.c
+ * @file    db_attr.c
```
subhasisb, line 43: We should consider removing the @file tag since we need to change everytime we rename a file. Just like we do not name our functions and doxygen knows automatically, removing the @file tag makes sense to me. Perhaps we can do this for all files u are touching?

src/lib/Libdb/pgsql/db_common.c
```diff
@@ -0,0 +1,1358 @@
...
+ *		inside it. The row data is loaded into this parameter.
+ *
+ * @return	Error code
+ * @retval	-1  - Failure
+ * @retval	0  - success
+ * @retval	1  - Success but no more rows
+ *
+ */
+int
+db_cursor_next(void *conn, void *st, pbs_db_obj_info_t *obj)
```
subhasisb, line 264: Why should this not be a static function?

src/lib/Libdb/pgsql/db_common.c
```diff
@@ -0,0 +1,1358 @@
...
+ * @param[in]   conn - Connected database handle
+ *
+ * @return      Error code
+ * @retval       0  - success
+ * @retval	-1  - Failure
+ *
+ *
+ */
+int
+db_prepare_sqls(void *conn)
```
subhasisb, line 331: Static function?

src/lib/Libdb/pgsql/db_common.c
```diff
@@ -0,0 +1,1358 @@
...
+{
+	if (pbs_db_prepare_job_sqls(conn) != 0)
```
subhasisb, line 333: why do these need to have a pbs_ prefix? They seem to be internal functions for postgres only and are not part of the DB API

src/include/pbs_db.h
```diff
@@ -838,18 +508,13 @@ void panic_stop_db(char *txt);
 
 /**
  * @brief
- *	Delete all the server attributes from the database
- *
- * @param[in]	conn - Connection handle
- * @param[in]	obj  - server information
+ *	Translates the error code to an error message
```
subhasisb, line 511: does panic_stop_db() belong to the pbs_db.h file? In other words is it a DB API? I think it is more of an appliaction/PBS api which then calls DB API's to stop the database, no? If so, it should not be in this file

src/include/pbs_db.h
```diff
@@ -866,6 +531,21 @@ int pg_db_delete_svrattr(pbs_db_conn_t *conn, pbs_db_obj_info_t *obj);
  */
 int compare_obj_hash(void *qs, int len, void *oldhash);
```
subhasisb, line 532: My bad, but compare_obj_hash() also a generic util API and is not part of libdb actually. It is used by the application, so we should move the declaration to libutil.h and definition to misc_util.c (or similar)

src/include/pbs_db.h
```diff
@@ -866,6 +531,21 @@ int pg_db_delete_svrattr(pbs_db_conn_t *conn, pbs_db_obj_info_t *obj);
...
+ *
+ * @retval       -1 - Failure
+ * @retval        0  - Success
+ *
+ */
+int pbs_db_password(void *conn, char *userid, char *password, char *olduser);
+
 /**
  * @brief
  *	Frees attribute list memory
```
subhasisb, line 551: free_db_attr_list() is also NOT a DB API that libdb will implement. In fact it is implemented in pbs_db_func.c, so put it in a different header file

src/lib/Libdb/pgsql/pbs_db_utility
ashwathraop: We do a pg_upgrade -c, which does the simulation of upgrade and if it fails maybe we can proceed with a dump and restore upgrade.

src/lib/Libdb/pgsql/pbs_db_utility
ashwathraop: I think I added it when we wanted to move from PG8.3 to 9.3 and it was in PBS v13. I guess this can go.

src/lib/Libdb/pgsql/pbs_db_utility
bhroam: While we only need to support the last 2 versions of PBS, I don't think we want to make it hard for people from older versions to upgrade to newer versions.  Is it a problem to keep the old upgrade from X to Y around, and then run them in sequence.  If they are on version 1.2, first upgrade them to 1.3, then 1.4, and then finally 1.5?  I agree writing an upgrade function from every old version to the new version isn't worth it, but if we can do it in stages, I don't see the harm in keeping them.

ashwathraop: I guess we can remove everything below v1_3_0. I moved back the schema upgrade code to its own file as it was before. But removed old schema versions and their code from it. We will have v1_3_0 and v1_4_0.

I am not sure having 2 files just for this would be a good idea. Wouldnt it be a maintenance headache?

ashwathraop: >  If they are on version 1.2, first upgrade them to 1.3, then 1.4, and then finally 1.5?

Yes, that's what we today. If you are on 1.2, we upgrade in stages till you have 1.5.

bhroam: If that is the case, I don't. see any reason why we don't want to keep the upgrade path.  While we don't support the version of PBS, we do want to entice people to upgrade to one we do support.  If we make it so they have to either upgrade multiple times, or have to install fresh, people will just continue using their old version.

subhasisb: > While we only need to support the last 2 versions of PBS, I don't think we want to make it hard for people from older versions to upgrade to newer versions. Is it a problem to keep the old upgrade from X to Y around, and then run them in sequence. If they are on version 1.2, first upgrade them to 1.3, then 1.4, and then finally 1.5? I agree writing an upgrade function from every old version to the new version isn't worth it, but if we can do it in stages, I don't see the harm in keeping them.

I think keeping more code means maintaining more code. We do not test upgrades from older releases, which means aside from this database upgrade part, a direct upgrade is probably anyway broken.

bhroam: I'm not sure I agree there is a maintenance issue here.  Since we upgrade in stages, there is no reason we'd ever change any of these older upgrade schema functions.  As for not testing them, we did.  When we were upgrading from X to Y, we tested it.  I don't think we need to test from old old old X to new Y to assume it will still work.  As long as it has been tested in sections, it should work.  By removing this code, the only way someone can upgrade from a non-supported version of PBS to a supported version is to either upgrade in stages (upgrade from 13 to 19 to 2020) or to do a fresh install.  Customers have active systems and won't want to do a fresh install and lose all their jobs.  What we are saying is that we don't want them to upgrade.  Normally I'd be all for removing code which isn't supported any more.  When it has to do with the desire to move from something unsupported to something that is supported, I disagree.

src/cmds/scripts/pbs_dataservice
```diff
@@ -179,232 +161,112 @@ fi
...
 		ret=$?
 		if [ ${ret} -eq 1 ]; then
-			if [ "$2" != "PBS" ]; then
-				echo "Starting PBS Data Service.."
-			fi
-			start_dataservice $1
+			echo "Starting PBS Data Service.."
+			${data_srv_bin} -s start
 			ret=$?
+			grep -q "database files are incompatible with server" $log_file
```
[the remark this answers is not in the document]
ashwathraop: Yes, it is DB specific. I will move it to libdb.

ashwathraop:
Had to rebase and force push to resolve conflicts and solve build issues.

src/lib/Libdb/pgsql/pbs_db_utility
subhasisb: No i am not saying we dont want them to upgrade, i am just saying that they must hop-upgrade. Upgrade to 2 version up in each hop.

subhasisb:
Code changes are looking good to me. However, looks like u have test cases still failing - please address those.

ashwathraop:
Description: Tests from given sources on platforms CENTOS7, UBUNTU1804, SUSE15, UBUNTU1604
ID | TOTAL | FAIL | ERROR | TIMEDOUT | SKIP | PASS
-- | -- | -- | -- | -- | -- | --
2454 | 3510 | 0 | 0 | 0 | 9 | 3501

src/server/job_recov_db.c
```diff
@@ -282,7 +282,7 @@ job_save_db(job *pjob)
 		pbs_db_get_errmsg(PBS_DB_ERR, &conn_db_err);
 		log_errf(PBSE_INTERNAL, __func__, "Failed to save job %s %s", pjob->ji_qs.ji_jobid, conn_db_err? conn_db_err : "");
 		if (conn_db_err) {
-			if (savetype == OBJ_SAVE_NEW && strstr(conn_db_err, "duplicate key value"))
+			if (savetype & OBJ_SAVE_NEW && strstr(conn_db_err, "duplicate key value"))
```
subhasisb, line 285: you really want the first part to be in its own brackets

src/server/job_recov_db.c
```diff
@@ -511,7 +511,7 @@ resv_save_db(resc_resv *presv)
 		pbs_db_get_errmsg(PBS_DB_ERR, &conn_db_err);
 		log_errf(PBSE_INTERNAL, __func__, "Failed to save resv %s %s", presv->ri_qs.ri_resvID, conn_db_err? conn_db_err : "");
 		if(conn_db_err) {
-			if (savetype == OBJ_SAVE_NEW && strstr(conn_db_err, "duplicate key value"))
+			if (savetype & OBJ_SAVE_NEW && strstr(conn_db_err, "duplicate key value"))
```
subhasisb, line 514: same here

src/server/queue_recov_db.c
```diff
@@ -221,7 +231,42 @@ que_recov_db(char *qname, pbs_queue	*pq)
...
+	pbs_queue *pque = NULL;
+	pbs_db_que_info_t *dbque = dbobj->pbs_db_un.pbs_db_que;
+
+	*refreshed = 0;
+	if ((pque = que_recov_db(dbque->qu_name, NULL)) != NULL) {
+		/* que_recov increments sv_numque */
+		sprintf(log_buffer, msg_init_recovque, pque->qu_qs.qu_name);
+		log_event(PBSEVENT_SYSTEM | PBSEVENT_ADMIN | PBSEVENT_DEBUG, PBS_EVENTCLASS_SERVER, LOG_INFO, msg_daemonname, log_buffer);
+		if (pque->qu_attr[(int) QE_ATR_ResourceAssn].at_flags & ATR_VFLAG_SET)
+			que_attr_def[(int) QE_ATR_ResourceAssn].at_free(&pque->qu_attr[(int) QE_ATR_ResourceAssn]);
```
nithinj, line 262: Why resource assigned is freed here? It is not getting decoded in decode_attr_db. Are we freeing it for server and node as well?

src/server/svr_func.c
```diff
@@ -6699,20 +6699,18 @@ svr_load_jobscript(job *pj)
 		log_err(-1, __func__, log_buffer);
 		return NULL;
 	}
-	script = strdup(jobscr.script);
-	pbs_db_cleanup_resultset(conn);
 
-	if (script == NULL) {
+	if (jobscr.script == NULL) {
```
nithinj, line 6703: Earlier this check served to handle strdup failure. Now as you have removed strdup, does this error logging still make sense? We also have another error logging prior to this which is more generic.

src/server/svr_recov_db.c
```diff
@@ -311,8 +313,9 @@ sched_recov_db(char *sname, pbs_sched *ps)
 	pbs_sched *psched  = NULL;
```
nithinj, line 313: looks like you don't need psched in this function!

src/server/svr_recov_db.c
```diff
@@ -385,9 +391,44 @@ sched_save_db(pbs_sched *ps)
...
+		if(!strncmp(dbsched->sched_name, PBS_DFLT_SCHED_NAME, strlen(PBS_DFLT_SCHED_NAME)))
+			dflt_scheduler = psched;
+		psched->pbs_scheduler_port = psched->sch_attr[SCHED_ATR_sched_port].at_val.at_long;
+		psched->pbs_scheduler_addr = get_hostaddr(psched->sch_attr[SCHED_ATR_SchedHost].at_val.at_str);
+		set_scheduler_flag(SCH_CONFIGURE, psched);
+	}
+
+	free_db_attr_list(&dbsched->db_attr_list);
+
+	*refreshed = 1;
```
nithinj, line 432: You are setting refreshed even if sched_recov_db fails

src/tools/printjob.c
```diff
@@ -316,53 +312,27 @@ print_db_job(char *id, int no_attributes)
 	pbs_db_job_info_t dbjob;
 	pbs_db_jobscr_info_t jobscr;
 	job xjob;
-	int db_conn_error;
 	char *db_errmsg = NULL;
-	char errmsg[PBS_MAX_DB_CONN_INIT_ERR + 1];
-
+	int failcode;
```
nithinj, line 316: I will call it return code  / rc

src/server/job_recov_db.c
```diff
@@ -545,22 +561,101 @@ resv_recov_db(char *resvid, resc_resv *presv)
...
+			if (pbs_db_delete_obj(svr_db_conn, dbobj) != 0)
+				log_errf(PBSE_SYSTEM, __func__, "job %s not purged", dbjob->ji_jobid);
+		}
+		goto err;
+	}
+
+	pbsd_init_job(pj, server_init_type);
+	*refreshed = 1;
+
+	if ((++totcount % 20) == 0) {
```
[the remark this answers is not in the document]
nithinj: You are keeping track of tot_count in pbs_db_search as well. Why not update svr_live there? You may also be loading a large number of nodes. So keeping at pbs_db_search is more sensible to me.

src/server/node_func.c
```diff
@@ -1083,60 +1068,20 @@ setup_nodes()
...
-			/* a vnode pointer will be returned */
-			if (np)
-				np->nd_moms[0]->mi_modtime = mom_modtime;
+	
+	rc = pbs_db_search(conn, &obj, NULL, (query_cb_t)&recov_node_cb);
+	if (rc == -1) {
+		pbs_db_get_errmsg(PBS_DB_ERR, &conn_db_err);
+		if (conn_db_err != NULL) {
+			sprintf(log_buffer, "%s", conn_db_err);
+			free(conn_db_err);
```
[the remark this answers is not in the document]
nithinj: yes, so do a log_errf and return. Same no of lines.

src/server/pbs_db_func.c
```diff
@@ -0,0 +1,627 @@
...
+
+	strcpy(log_buffer, "Starting PBS dataservice");
+	log_event(PBSEVENT_SYSTEM | PBSEVENT_FORCE,
+		PBS_EVENTCLASS_SERVER, LOG_CRIT,
+		msg_daemonname, log_buffer);
+
+	rc = pbs_start_db(conn_db_host, pbs_conf.pbs_data_service_port);
+	if (rc != 0) {
+		log_set_dberr("Failed to start PBS dataservice", failstr);
+		log_event(PBSEVENT_SYSTEM | PBSEVENT_ADMIN,
```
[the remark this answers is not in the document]
nithinj: failstr is not set!

src/server/pbs_db_func.c
```diff
@@ -0,0 +1,627 @@
...
+			return; /* dataservice not running, got killed? */
+
+		strcpy(log_buffer, "Stopping PBS dataservice");
+		log_event(PBSEVENT_SYSTEM | PBSEVENT_FORCE,
+			PBS_EVENTCLASS_SERVER, LOG_CRIT,
+			msg_daemonname, log_buffer);
+
+		if (pbs_stop_db(conn_db_host, pbs_conf.pbs_data_service_port) != 0) {
+			pbs_db_get_errmsg(PBS_DB_ERR, &db_err);
+			log_set_dberr("Failed to stop PBS dataservice", db_err);
```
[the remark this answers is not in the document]
nithinj: Isn't it confusing? Either you should pass the log_buffer as part of log_set_dberr, so the reader knows you are causing a side-effect. Else handle the whole thing inside liblog.

src/server/pbs_db_func.c
```diff
@@ -0,0 +1,627 @@
...
+		}
+		if (db_err) {
+			free(db_err);
+			db_err = NULL;
+		}
+
+		db_oper_failed_times++;
+
+		/* try stopping after some time again */
+		db_delay = (int)(1000 + db_oper_failed_times * 2);
```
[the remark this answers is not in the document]
nithinj: usleep was sleeping in microseconds. So earlier sleeping time was in the order of milli-seconds. Why should it be in seconds now?

src/server/pbs_db_func.c
```diff
@@ -0,0 +1,627 @@
...
+		}
+
+		if (conn_state == PBS_DB_CONNECT_STATE_NOT_CONNECTED) {
+			/* get fresh connection */
+			if ((conn = get_db_connect_information()) == NULL)
+				conn_state = PBS_DB_CONNECT_STATE_FAILED;
+			else
+				conn_state = PBS_DB_CONNECT_STATE_CONNECTED;
+		}
+		db_delay = (int)(1000 + db_oper_failed_times * 2);
```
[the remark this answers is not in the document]
nithinj: Same question. Is it intentional to change from milliseconds to order of seconds?

src/server/job_recov_db.c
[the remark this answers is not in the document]
ashwathraop: pbs_db_search is used for other pbs components too. But svr_live is checked only here.

src/server/pbs_db_func.c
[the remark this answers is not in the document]
ashwathraop: We had sleep but I had changed it to usleep which does not look necessary. So reverted it back now.

src/server/pbs_db_func.c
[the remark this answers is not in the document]
ashwathraop: Yes.

src/server/pbsd_init.c
```diff
@@ -828,67 +746,24 @@ pbsd_init(int type)
 	/* get jobs from DB */
 	obj.pbs_db_obj_type = PBS_DB_JOB;
 	obj.pbs_db_un.pbs_db_job = &dbjob;
-	state = pbs_db_cursor_init(conn, &obj, NULL);
-	if (state == NULL) {
-		sprintf(log_buffer, "%s", (char *) conn->conn_db_err);
+	rc = pbs_db_search(conn, &obj, NULL, (query_cb_t)&recov_job_cb);
+	if (rc == -1) {
+		pbs_db_get_errmsg(PBS_DB_ERR, &conn_db_err);
+		sprintf(log_buffer, "%s", conn_db_err);
```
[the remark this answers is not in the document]
nithinj: PBS_DB_ERR is handled within pbs_db_get_errmsg. If strdup itself fails, then PBS will be running out of memory.

Thinking on pbs_db_get_errmsg(). Why does it have to strdup all the time and free in the caller. Have you thought of a buffer internal to libdb and reusing that?

src/server/queue_recov_db.c
nithinj: If so, we should be seeing double counts with server and node resc assigned as well. Please confirm.

src/server/svr_func.c
nithinj: Better to handle strdup failure in libdb itself and pass a specific return code back.

src/server/svr_recov_db.c
nithinj: That can be done with ps as well right? right now as ps is assigned with NULL even if it is not allocated there. This is a memory leak.

src/tools/printjob.c
nithinj: Return value can be PBS_DB_SUCCESS

src/server/job_recov_db.c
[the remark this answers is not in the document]
nithinj: By other components do you mean other objects like node, resv etc? does it make more sense to keep it there then?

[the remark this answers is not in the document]
ashwathraop: Number of node and resv objects might be significantly less comapred to less. So I dont see a need to have it for others except jobs.

src/server/pbsd_init.c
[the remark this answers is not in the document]
ashwathraop: For me, this looks simple. You call pbs_db_get_errmsg only when a libdb API fails. And caller knows it failed. Maintaining cache inside libdb adds the overhead of managing that cache which I dont think is a good idea.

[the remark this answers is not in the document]
nithinj: On the contrary, libdb is delegating its responsibility to check the buffer allocation failure to the caller. Caller has to check the buffer is valid and free after use. Which can be eliminated with a static buffer inside libdb.

src/server/job_recov_db.c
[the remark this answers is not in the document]
nithinj: Yes, I agree that they are significantly lesser compared to jobs. But an order of magnitude more than 20 which is the count you used to update svr_live. For that reason pbs_db_search is the right place. Also, you don't have to maintain totcount at the job recovery function.

ashwathraop:
Latest test results
Tests from given sources on platforms CENTOS7, UBUNTU1804, SUSE15, UBUNTU1604 
ID | TOTAL | FAIL | ERROR | TIMEDOUT | SKIP | PASS
-- | -- | -- | -- | -- | -- | --
2605 | 3510 | 0 | 0 | 0 | 0 | 3510

src/server/node_recov_db.c
```diff
@@ -173,8 +176,11 @@ node_recov_db(char *nd_name, struct pbsnode *pnode)
 
 	if (rc == 0)
 		rc = db_to_node(pnode, &dbnode);
-	else
-		log_errf(PBSE_INTERNAL, __func__, "Failed to load node %s %s", nd_name, (conn->conn_db_err)? conn->conn_db_err : "");
+	else {
+		pbs_db_get_errmsg(PBS_DB_ERR, &conn_db_err);
+		log_errf(PBSE_INTERNAL, __func__, "Failed to load node %s %s", nd_name, conn_db_err? conn_db_err : "");
```
nithinj, line 181: space before conditional operator. Please do the same for other places where you use conditional operator.

src/server/pbs_db_func.c
```diff
@@ -0,0 +1,588 @@
...
+	while (1) {
+		if (pbs_status_db(conn_db_host, pbs_conf.pbs_data_service_port) != 0)
+			return; /* dataservice not running, got killed? */
+
+		log_eventf(PBSEVENT_SYSTEM | PBSEVENT_FORCE, PBS_EVENTCLASS_SERVER, LOG_CRIT, msg_daemonname, "Stopping PBS dataservice");
+
+		if (pbs_stop_db(conn_db_host, pbs_conf.pbs_data_service_port) != 0) {
+			pbs_db_get_errmsg(PBS_DB_ERR, &db_err);
+			snprintf(log_buffer, LOG_BUF_SIZE, "%s %s", "Failed to stop PBS dataservice.", db_err ? db_err : "");
+			log_eventf( PBSEVENT_SYSTEM | PBSEVENT_ADMIN, PBS_EVENTCLASS_SERVER, LOG_ERR, msg_daemonname, log_buffer);
```
nithinj, line 183: You don't need log_eventf here!

src/server/pbsd_init.c
```diff
@@ -1656,6 +1539,89 @@ pbsd_init_job(job *pjob, int type)
...
+				/* natural vnode being recovered, add to pool */
+				add_mom_to_pool(np->nd_moms[0]);
+			}
+		}
+	} else {
+		if (rc == PBSE_NODEEXIST)
+			sprintf(log_buffer, "duplicate node \"%s\"", dbnode->nd_name);
+		else
+			sprintf(log_buffer, "could not create node \"%s\", error = %d", dbnode->nd_name, rc);
+		log_err(-1, __func__, log_buffer);
```
nithinj, line 1620: Use log_errf

src/server/svr_func.c
nithinj: This one is not addressed!

src/tools/printjob.c
nithinj: Please do not resolve comments which are not addressed.

src/server/pbs_db_func.c
ashwathraop: I think its good to log to server log since writing just to stderr will not record the cause.

src/server/job_recov_db.c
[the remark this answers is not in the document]
ashwathraop: If you insist I can increase the number from 20. But I dont want to do any server related activity in a libdb API like pbs_db_search.

src/server/pbsd_init.c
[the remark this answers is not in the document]
ashwathraop: I still feel the caller should take care of this since it knows when it is interested in getting the error message and this way we can keep libdb simple.

ashwathraop:
@nithinj and @subhasisb I have addresses all the comments. Can you guys take a look and let me if the PR looks fine?

src/server/pbsd_init.c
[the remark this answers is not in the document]
nithinj: I do not understand why it has to be complex? Maybe something we can take up later!

src/server/job_recov_db.c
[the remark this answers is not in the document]
nithinj: Make sense if pbs_db_search is part of libdb.

PR #1644, opened 2020-04-07
Hugepagesize updates to test_cgroup_offline_node_vnpernuma

vccardenas:
#### Describe Bug or Feature
The test TestCgroupsHook.test_cgroup_offline_node_vnpernuma fails on systems that don't have Hugepagesize in meminfo.

#### Describe Your Change
The test will skip if Hugepagesize is not in meminfo.

[before.txt](https://github.com/PBSPro/pbspro/files/4446339/before.txt)
[after.txt](https://github.com/PBSPro/pbspro/files/4447311/after.txt)

### DIFF
--- a/test/tests/functional/pbs_cgroups_hook.py
+++ b/test/tests/functional/pbs_cgroups_hook.py
@@ -1754,6 +1754,9 @@ if %s e.job.in_ms_mom():
         the cgroup and brought back online once the cgroup is cleaned up.
         vnode_per_numa_node = true
         """
+        with open('/proc/meminfo', 'r') as desc:
+            if 'Hugepagesize' not in desc:
+                self.skipTest('Hugepagesize not in meminfo')
         name = 'CGROUP7.2'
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -1754,8 +1754,8 @@ def test_cgroup_offline_node_vnpernuma(self):
         the cgroup and brought back online once the cgroup is cleaned up.
         vnode_per_numa_node = true
         """
-        with open('/proc/meminfo', 'r') as desc:
-            if 'Hugepagesize' not in desc:
+        meminfo = open(os.path.join(os.sep, 'proc', 'meminfo'), 'r').read()
+        if 'Hugepagesize' not in meminfo:
```
bhroam, line 1758: I think you still want the with statement.  I don't believe the file will ever be closed.  I think you want the with open() as desc.  Then if 'Hugepagesize' not in desc.read()

PR #1645, opened 2020-04-08
Fixing mem leak in pbs_asyalterjob

agrawalravi90:
#### Describe Bug or Feature

pbs_asyalterjob was added recently for async alterjobs. It can leak memory if any of the function calls fail as it allocates attropl list at the beginning but only frees it at the end:
40 bytes in 1 blocks are definitely lost in loss record 52 of 823
   at 0x4C28C23: **malloc** (vg_replace_malloc.c:299)
   by 0x4B4AB1: **attrl_to_attropl** (pbsD_alterjo.c:69)
   by 0x4B4CA1: **pbs_asyalterjob** (pbsD_alterjo.c:177)
   by 0x452213: send_attr_updates (job_info.c:1788)
   by 0x45216F: send_job_updates (job_info.c:1757)
   by 0x44BAC8: main_sched_loop (fifo.c:1111)
   by 0x44AE87: scheduling_cycle (fifo.c:733)
   by 0x44AB01: intermediate_schedule (fifo.c:605)
   by 0x44AA4C: schedule (fifo.c:556)
   by 0x449802: main (pbs_sched.c:1386)

#### Describe Your Change

Now allocating attropl list right before the function call where it's needed, and freeing it immediately afterwards.

### DIFF
--- a/src/lib/Libifl/pbsD_alterjo.c
+++ b/src/lib/Libifl/pbsD_alterjo.c
@@ -174,8 +174,6 @@ pbs_asyalterjob(int c, char *jobid, struct attrl *attrib, char *extend)
 	if ((jobid == NULL) || (*jobid == '\0'))
 		return (pbs_errno = PBSE_IVALREQ);
 
-	attrib_opl = attrl_to_attropl(attrib);
-
 	/* initialize the thread context data, if not initialized */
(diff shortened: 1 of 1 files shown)


agrawalravi90:
haha, it does look like that doesn't it? I bet it's been that way ever since the first iteration of pbs_alterjob :)

PR #1649, opened 2020-04-09
Add support for multiple mom action scripts in PTL framework

minghui-liu:
#### Describe Bug or Feature
In "pbs_testlib.py", class MOM has the "parse_config" method for parsing and adding to mom config file. "parse_config" reads in the config file line by line and splits each line into key-value pairs by the first space. Because of this simple parsing method, all $action entries have the same key, and therefore only one will be added to the dictionary that this method returns.

#### Describe Your Change

* Modified the `parse_config` function. When `parse_config` reads a line that starts with "$action", it now includes the next word into the key, and use the rest of the string to be the value. 
* Modified `add_checkpoint_abort_script` to use "$action checkpoint_abort" as key instead of just "$action". 
* I also added another function that adds a restart action script to mom config file. This will be used in a coming PTL test case. 

#### Link to Design Doc

N/A

Screen Shot 2020-04-14 at 4 34 47 PM
Screen Shot 2020-04-14 at 4 35 21 PM

These two tests are also failing on nightly build of master branch.

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -13572,10 +13572,25 @@ class MoM(PBSService):
         self.du.chmod(hostname=self.hostname, path=chk_file, mode=0o700)
         self.du.chown(hostname=self.hostname, path=chk_file, runas=ROOT_USER,
                       uid=0, gid=0)
-        c = {'$action': 'checkpoint_abort ' +
+        c = {'$action checkpoint_abort':
              str(abort_time) + ' !' + chk_file + ' %sid'}
         self.add_config(c)
         return chk_file
+    
+    def add_restart_script(self, dirname=None, body=None,
+                           abort_time=30):
+        """
+        Add restart script in the mom config.
+        returns: a temp file for restart script
+        """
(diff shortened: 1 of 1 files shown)


test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -13572,11 +13572,26 @@ def add_checkpoint_abort_script(self, dirname=None, body=None,
...
              str(abort_time) + ' !' + chk_file + ' %sid'}
         self.add_config(c)
         return chk_file
 
+    def add_restart_script(self, dirname=None, body=None,
+                           abort_time=30):
+        """
+        Add restart script in the mom config.
+        returns: a temp file for restart script
+        """
```
mike0042, line 13585: Why are we adding a method that never gets called?

minghui-liu: Hi @mike0042 . I plan to use this method in an upcoming pull request. This bug is blocking another bug fix of mine, which has a PTL test case that needs to add both a checkpoint abort script and a restart script.

minghui-liu: In other words the motivation for fixing this bug is to add this method, so I am including it in this here.

PR #1651, opened 2020-04-13
Fixing two test cases of TestSuspendResumeAccounting

sujatapatnaik52:
#### Describe Bug or Feature
Testcases `test_admin_suspend_resume_signal` and `test_resc_released_susp_resume` was failing on specific platforms.

#### Describe Your Change
1. For `test_admin_suspend_resume_signal`, the intent of the test case is to check for the suspend and resume event accounting logs, so removing the check for state of the node. 
2. For `test_resc_released_susp_resume`, get the exec_vnode attribute from the job and search for the same in the accounting string. 

#### Attach Test Logs/Output
[test_logs.txt](https://github.com/PBSPro/pbspro/files/4469414/test_logs.txt)

### DIFF
--- a/test/tests/functional/pbs_suspend_resume_accounting.py
+++ b/test/tests/functional/pbs_suspend_resume_accounting.py
@@ -220,13 +220,9 @@ class TestSuspendResumeAccounting(TestFunctional):
         # check for both ncpus and mem are released
         resc_released = 'resources_released=(%s:ncpus=1:mem=524288kb)'
 
-        if self.mom.is_cpuset_mom():
-            node = self.server.status(JOB, 'exec_vnode',
-                                      id=jid)[0]['exec_vnode']
-            vn = node.split('+')[0].split(':')[0].split('(')[1]
-            resc_released = resc_released % vn
-        else:
-            resc_released = resc_released % self.server.shortname
+        node = self.server.status(JOB, 'exec_vnode', id=jid)[0]['exec_vnode']
+        vn = node.split('+')[0].split(':')[0].split('(')[1]
+        resc_released = resc_released % vn
         record = 'z;%s;resources_used.' % jid
         line = self.server.accounting_match(msg=record, id=jid)[1]
         self.assertIn(resc_released, line)
--- a/test/tests/functional/pbs_suspend_resume_accounting.py
+++ b/test/tests/functional/pbs_suspend_resume_accounting.py
@@ -219,8 +219,12 @@ class TestSuspendResumeAccounting(TestFunctional):
 
         # check for both ncpus and mem are released
         resc_released = 'resources_released=(%s:ncpus=1:mem=524288kb)'
+
         if self.mom.is_cpuset_mom():
-            resc_released = resc_released % self.server.status(NODE)[1]['id']
+            node = self.server.status(JOB, 'exec_vnode',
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_suspend_resume_accounting.py
```diff
@@ -178,27 +178,29 @@ def test_suspend_resume_job_scheduler(self):
...
         self.server.accounting_match(msg=record, id=jid1)
 
+    @skipOnCpuSet
     def test_admin_suspend_resume_signal(self):
         """
         Test case to verify accounting of admin-suspend
         and admin-resume records.
         """
+        a = {ATTR_rescavail + '.ncpus': 2}
+        self.server.create_vnodes('vn', a, 1, self.mom)
```
prakashcv13, line 188: can we avoid creating vnodes on cpuset machined, so that we can remove the skipOnCpuSet?

sujatapatnaik52:
[test_logs.txt](https://github.com/PBSPro/pbspro/files/4469602/test_logs.txt)

prakashcv13:
@sujatapatnaik52, 
1) can we also have the "before" logs.
2) let us change the description to reflect the changes

test/tests/functional/pbs_suspend_resume_accounting.py
```diff
@@ -221,8 +218,11 @@ def test_resc_released_susp_resume(self):
         self.server.sigjob(jobid=jid, signal="suspend")
 
         # check for both ncpus and mem are released
-        resc_released = 'resources_released=(%s:ncpus=1:mem=524288kb)' % \
-                        self.server.shortname
+        resc_released = 'resources_released=(%s:ncpus=1:mem=524288kb)'
+        if self.mom.is_cpuset_mom():
+            resc_released = resc_released % self.server.status(NODE)[1]['id']
```
Mahalty, line 223: Instead of using the first vnode here, can we do either of this

- Submit a job specifying the vnode and use it while verifying both ncpus and mem are released.
**(or)**
- Take the node name from exec_vnode attribute of the job while verifying

sujatapatnaik52: Thanks @Mahalty. I opted for the second one. Please review the latest commit.

sujatapatnaik52:
[test_logs.txt](https://github.com/PBSPro/pbspro/files/4470112/test_logs.txt)

test/tests/functional/pbs_suspend_resume_accounting.py
```diff
@@ -219,8 +219,12 @@ def test_resc_released_susp_resume(self):
 
         # check for both ncpus and mem are released
         resc_released = 'resources_released=(%s:ncpus=1:mem=524288kb)'
+
         if self.mom.is_cpuset_mom():
-            resc_released = resc_released % self.server.status(NODE)[1]['id']
+            node = self.server.status(JOB, 'exec_vnode',
```
Mahalty, line 224: Move this outside the if condition, that way we will not need if-else block

sujatapatnaik52:
[test_logs.txt](https://github.com/PBSPro/pbspro/files/4470223/test_logs.txt)

sujatapatnaik52:
Thanks @prakashcv13 and @Mahalty for the review and sign off.

PR #1653, opened 2020-04-13
New sched attribute to control asynchronicity of runjob requests

agrawalravi90:
#### Describe Bug or Feature

Adding a new sched attribute called "job_run_wait", which will control the asynchronicity of runjob requests from scheduler to server. Motivation is performance (http://community.pbspro.org/t/scheduler-can-spend-94-of-its-time-waiting-for-job-run-ack/2053), adding the mechanism to do a truly asynchronous runjob request can can make sched cycles an order of magnitude faster.

#### Describe Your Change

- Added new sched attribute 'job_run_wait', it takes 3 values:
1) "execjob_hook": this will behave similarly to "throughput_mode=False"
2) "runjob_hook": This will behave similarly to "throughput_mode=True"
3) "none": This is new functionality, it will make scheduler's runjob requests to server completely asynchronous.

- Deprecating sched attribute 'throughput_mode'

See the design document for more details.

#### Link to Design Doc

https://pbspro.atlassian.net/wiki/spaces/PD/pages/1633517573/New+sched+attribute+to+control+runjob+wait+making+pbs+asynrunjob+truly+async+deprecating+throughput+mode

[runjobperf.log](https://github.com/PBSPro/pbspro/files/4472123/runjobperf.log)
[schedjobrunwait.log](https://github.com/openpbs/openpbs/files/4720243/schedjobrunwait.log)

**Performance test results:**
Time taken for runjob_wait=execjob_hook(throughput_mode=False): **60.959792137145996**
Time taken for runjob_wait=runjob_hook(throughput_mode=True): **58.68516206741333**
Time taken for runjob_wait=none: **1.5238490104675293**


agrawalravi90:
Putting [WIP] for now because I'm testing some things

src/scheduler/server_info.c
```diff
@@ -832,6 +833,43 @@ query_server_dyn_res(server_info *sinfo)
...
+ * @brief	Check if there are any runjob hooks configured in the system
+ *
+ * @param	void
+ *
+ * @return	int
+ * @retval	1 if runjob hooks are configured
+ * @retval	0 if not
+ */
+static int
+check_runjob_hooks(int pbs_sd)
```
arungrover, line 846: I think it should be the server that should set the run_job_wait attribute accordingly.
But, I can understand that you are doing it in scheduler because you don't want to overwrite user-provided values in server. So to sort that out, how about the server object itself has an attribute set that says it has configured and enabled runjob hooks?
I just don't like that the way it is now, the scheduler will query hooks in every cycle and with multi-sched it will only get worse.

src/server/req_runjob.c
```diff
@@ -340,11 +340,26 @@ req_runjob(struct batch_request *preq)
 
 #ifndef NAS /* localmod 133 */
 	if ((psched->scheduler_sock != -1) && was_job_alteredmoved(parent)) {
-		int index = find_attr(sched_attr_def, ATTR_throughput_mode, SCHED_ATR_LAST);
-		/* do not blacklist altered/moved jobs when throughput_mode is enabled */
-		if ((index == -1) ||
-			((psched->sch_attr[index].at_flags & ATR_VFLAG_SET) &&
-			 (psched->sch_attr[index].at_val.at_long == 0))) {
+		int index_tp = find_attr(sched_attr_def, ATTR_throughput_mode, SCHED_ATR_LAST);
```
arungrover, line 343: I don't see any purpose for this find_attr. Attribute array is static, can't you directly index it using SCHED_ATR_job_run_wait or SCHED_ATR_throughput_mode?

src/server/sched_func.c
```diff
@@ -964,39 +963,34 @@ set_sched_default(pbs_sched *psched, int from_scheduler)
...
 		psched->sch_attr[SCHED_ATR_preempt_sort].at_flags |= ATR_VFLAG_SET | ATR_VFLAG_MODCACHE | ATR_VFLAG_DEFLT;
-		flag = 1;
 	}
 	if (!(psched->sch_attr[SCHED_ATR_server_dyn_res_alarm].at_flags & ATR_VFLAG_SET)) {
 		psched->sch_attr[SCHED_ATR_server_dyn_res_alarm].at_val.at_long = PBS_SERVER_DYN_RES_ALARM_DEFAULT;
 		psched->sch_attr[SCHED_ATR_server_dyn_res_alarm].at_flags |= ATR_VFLAG_SET | ATR_VFLAG_MODCACHE | ATR_VFLAG_DEFLT;
-		flag = 1;
 	}
 
+	/* Note: Intentionally not setting job_run_wait because throughput_mode still exists.
```
arungrover, line 990: is this comment misplaced?

src/server/sched_func.c
```diff
@@ -1068,3 +1062,90 @@ action_sched_server_dyn_res_alarm(attribute *pattr, void *pobj, int actmode)
 
 	return PBSE_NONE;
 }
+
+/* action function for opt_backfill_fuzzy -- only allow the correct values */
```
arungrover, line 1066: Please make a doxygen function header

src/scheduler/server_info.c
agrawalravi90: I understand where you are coming from, but if you look at the performance test logs that I've posted, there's a test which specifically looks at the performance lag due to this, and it's actually not noticeable, it's just one batch request out of the thousands that sched makes every cycle, and it queries just one attribute, so the data transferred isn't much either. So I don't think it makes much of a difference.

I'd love to prevent the extra request, but I don't know if adding another server attribute is better than this.

src/lib/Libattr/master_sched_attr_def.xml
```diff
@@ -235,6 +235,22 @@
...
    </attributes>
+   <attributes> /* job_run_wait */
+        <member_name><both>ATTR_job_run_wait</both></member_name>    <!-- job_run_wait -->
+        <member_at_decode>decode_str</member_at_decode>
+        <member_at_encode>encode_str</member_at_encode>
+        <member_at_set>set_str</member_at_set>
+        <member_at_comp>comp_str</member_at_comp>
+        <member_at_free>free_str</member_at_free>
+        <member_at_action>action_job_run_wait</member_at_action>
+        <member_at_flags><both>MGR_ONLY_SET</both></member_at_flags>
```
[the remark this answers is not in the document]
agrawalravi90: TBH, I wasn't sure what to set it to. Some of the attributes are set to MGR_ONLY_SET. It made sense to me that this attribute be set only by an admin so I marked it MGR_ONLY_SET. Now, are operators admins as well? I actually have no idea :) please let me know what you think.

src/lib/Libifl/pbsD_asyrun.c
```diff
@@ -108,11 +107,78 @@ __pbs_asyrunjob(int c, char *jobid, char *location, char *extend)
...
+	if (pbs_client_thread_unlock_connection(c) != 0)
+		return pbs_errno;
+
+	return 0;
+}
+
+/**
+ * @brief
+ *	-send a run job batch request which waits for an ack from server
+ *	this differs from pbs_runjob in that server sends an ack before contacting the mom
```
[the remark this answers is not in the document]
agrawalravi90: No, I meant pbs_runjob only, this call differs from pbs_runjob in that server sends an ACL before contacting mom.

src/lib/Libattr/master_sched_attr_def.xml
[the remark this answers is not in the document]
bhroam: I can tell you what an operator was supposed to be.  So many developers ignore the role and only set things to user or manager, so it has lost some of its power.  The operator role is supposed to be a computer operator.  They are the first line of support before sysadmins (I was one at one point in time).  The manager is the sysadmin.  What is really comes down to is if it is policy or not.  Operators can do things like delete jobs or stop a queue, but they can't create queues or turn on policies like preemption. This sounds like more of a policy decision to me.  I agree with MGR_ONLY_SET.  If you make this MGR_ONLY_SET, you should make throughput_mode the same.

src/scheduler/server_info.c
arungrover: I think you need to query 2 attributes - enabled and type. In the test, you have only one hook. The performance will vary depending on how many runjob hooks are present in the system.
If you have a way from server to tell the scheduler that yes there is a runjob hook enabled then that's all scheduler cares about. Adding a server attribute to me is more efficient than scheduler querying it because adding/removing a hook is a rare event as compared to the number of cycles the scheduler is going to run. So once the attribute is set, it is unlikely it will change.
Using encode function is also another approach but in that case, the server will do repetitive work every scheduling cycle.

src/lib/Libattr/master_sched_attr_def.xml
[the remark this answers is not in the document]
agrawalravi90: thanks for explaining Bhroam, ill change throughput_mode as well.

src/scheduler/server_info.c
agrawalravi90: Ok, I like the encode function idea, I didn't think about that, I'll try it out, thanks

src/server/sched_func.c
agrawalravi90: it's not, I meant to clarify why I'm not setting a default for job_run_wait.

src/scheduler/server_info.c
agrawalravi90: @arungrover about performance lag every cycle, I'm thinking of refactoring the code so that sched attributes are not queried every cycle, they will be globals which will be cached by sched, anytime it changes via qmgr, we'll update the cache. I'll do this in another PR though. So, performance won't be a problem. Please let me know if you still have any concerns.

arungrover: sounds good

bhroam: The code isn't repetitive.  The server keeps a list of every type of hook in a global.  It just needs to look to see if that global is NULL or not.  If it is kept in the scheduler, the scheduler has to ask the server for the hooks.  The server has to wrap up all the attributes as strings to send to the scheduler, send it, and then the scheduler has to unwrap it to look for a runjob hook.  I think it is much better done in the server in the encode function.

arungrover: The server also has to loop through each hook to check if it is enabled, right?

agrawalravi90: I think i do need to go over the list of runjob hooks and check to see if any of them are enabled or not, but that shouldn't be much computation.

arungrover: I'm fine with it, especially when you said that you are going to make the scheduler cache the sched object and make it query on need basis.

agrawalravi90: turns out, we can't know who the requester is inside the encode functions, so the encode solution won't work. I'll do the secondary server attribute solution, marking the PR as WIP for now.

src/server/req_runjob.c
agrawalravi90: you are right, I'd just extrapolated the existing code, we should be able to access the array directly.

test/tests/functional/pbs_sched_runjobwait.py
```diff
@@ -0,0 +1,315 @@
...
+# under a commercial license agreement.
+#
+# Use of Altair’s trademarks, including but not limited to "PBS™",
+# "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
+# trademark licensing policies.
+
+from tests.functional import *
+
+
+class TestSchedRunjobWait(TestFunctional):
```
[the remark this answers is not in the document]
agrawalravi90: "Can this be added in another test plan?"
I'm not entirely sure what that means, can you please clarify?

test/tests/functional/pbs_sched_runjobwait.py
```diff
@@ -0,0 +1,315 @@
...
+            self.fail("invalid str value for job_run_wait was accepted")
+        except PbsManagerError:
+            pass
+
+        try:
+            self.server.manager(MGR_CMD_SET, SCHED,
+                                {'job_run_wait': 0}, id='default')
+            self.fail("invalid int value for job_run_wait was accepted")
+        except PbsManagerError:
+            pass
```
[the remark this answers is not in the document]
agrawalravi90: I didn't know about self.assertRaises(), it's so cool! thanks for the suggestion :)

test/tests/functional/pbs_sched_runjobwait.py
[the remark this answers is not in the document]
bhroam: The general rule of thumb is to create general test plans that can be enhanced in the future and not create small one-off test plans.  This seems rather specific.  Is there another test plan out there that these tests could be included with?

agrawalravi90:
@arungrover  and @bhroam  I've finally addressed your comments, took time since I had to change approaches and PTL wasn't behaving well. Please provide further comments. Thanks!

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -6791,9 +6790,23 @@ def manager(self, cmd, obj_type, attrib=None, id=None, extend=None,
...
-                    self.update_attributes(obj_type, bsl)
+                    
+                    # Since we stat everything, overwrite the cache
+                    self.update_attributes(obj_type, bsl, overwrite=True)
+                    
+                    # Filter out the attributes requested
+                    if attrib:
+                        bsl_attr = []
+                        for obj in bsl:
+                            newd = {}
```
bhroam, line 6801: How about we change the name of this to dnew?

test/tests/functional/pbs_sched_runjobwait.py
```diff
@@ -195,6 +191,7 @@ def test_multisched_multival(self):
         # sched 1 with job_run_wait=none runs first job without waiting
         # for runjob reject, so it doesn't run second job.
         # Ultimately, neither jobs should run
+        print(self.scheds[sc_queue[0][0]].attributes)
```
bhroam, line 194: Remove this debug statement

agrawalravi90:
@bhroam I've addressed your comments, please let me know if it looks okay now, thanks!

src/server/req_stat.c
```diff
@@ -738,6 +738,10 @@ req_stat_svr(struct batch_request *preq)
 	CLEAR_HEAD(pstat->brp_attr);
 	append_link(&preply->brp_un.brp_status, &pstat->brp_stlink, pstat);
 
+	/* has_runjob_hook should get set every time via encode_isrunhook */
+	server.sv_attr[(int)SRV_ATR_has_runjob_hook].at_priv_encoded = NULL;
+	server.sv_attr[(int)SRV_ATR_has_runjob_hook].at_user_encoded = NULL;
```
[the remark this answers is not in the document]
agrawalravi90: I didn't see this leak during my testing, and it dumps core when i try to free here. I'll run a full regression with valgrind just to be sure.

src/scheduler/server_info.c
```diff
@@ -847,7 +853,7 @@ query_server_dyn_res(server_info *sinfo)
  *
  */
 int
-query_sched_obj(status *policy, struct batch_status *sched, server_info *sinfo)
+query_sched_obj(int pbs_sd, status *policy, struct batch_status *sched, server_info *sinfo)
```
bhroam, line 856: pbs_sd is no longer needed now that you aren't querying the hooks

src/server/req_stat.c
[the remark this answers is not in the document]
bhroam: I do see the encode functions allocating memory, but they are also linking it to a list.  The two pointers you NULL just point to that encoded attribute.  I think you are good.

agrawalravi90:
@bhroam thanks for the thorough review!
@arungrover please provide more comments whenever you can, thanks!

agrawalravi90:
@arungrover gentle reminder, thanks
@bhroam  turns out that i was indeed leaking memory, and the free worked this time, not sure why it was crashing server earlier, i must have been doing something wrong, anyways, please review the latest commit as well, thanks.

bhroam:
@agrawalravi90 The encode function links this memory into a list hung off the attriburte.  The reason the server is crashing is now you have a dangling pointer there.  svrcached() calls free_svrcache() to do the freeing.  You call that on an attribute.  Of course you can just add the MODCACHE flag and svrcached() will do that for you on the next call.

agrawalravi90:
I find this code weird:
`	if (resc_access_perm & PRIV_READ)
		encoded = pat->at_priv_encoded;
	else
		encoded = pat->at_user_encoded;

	if (pat->at_flags & ATR_VFLAG_MODCACHE)
		/* free old cache value if the value has changed */
		free_svrcache(pat);`

We point 'encoded' to memory that can get free'd by free_svrcache(pat)? Shouldn't we be calling free_svrcache(pat) before setting 'encoded' ?

src/lib/Libifl/pbsD_asyrun.c
```diff
@@ -49,24 +49,22 @@
 
 
 /**
- * @brief
- *	-send async run job batch request.
+ * @brief	Helper function for pbs_asynrunjob and pbs_asynrunjob_ack
```
arungrover, line 52: Can't we also update __pbs_runjob() to use the helper function. There is no difference between __pbs_runjob and pbs_asyrunjob_ack other than the batch request ID.

src/scheduler/fifo.c
```diff
@@ -1282,6 +1282,28 @@ update_job_can_not_run(int pbs_sd, resource_resv *job, schd_error *err)
...
+ * @param[in]	execvnode	-	the execvnode to run the job on
+ *
+ * @return	int
+ * @retval	return value of the runjob call
+ */
+static int
+send_runjob(int pbs_sd, enum runjob_mode rj_mode, char *jobid, char *execvnode)
+{
+	if (rj_mode == RJ_NOWAIT)
+		return pbs_asyrunjob(pbs_sd,jobid, execvnode, NULL);
```
arungrover, line 1300: Please add a space after the comma.

src/server/sched_func.c
arungrover: I don't see the header changed. Did you forget to change it?

test/tests/functional/pbs_sched_runjobwait.py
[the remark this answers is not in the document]
agrawalravi90: thanks for clarifying and sorry that i missed this comment. I did try to search for an existing place for this, but didn't really find any other generic test suite where this would fit. Please let me know if you can think of any.

agrawalravi90:
@arungrover  thanks I've addressed your comments, please provide further feedback.

src/lib/Libifl/pbsD_runjob.c
```diff
@@ -90,34 +135,103 @@ __pbs_runjob(int c, char *jobid, char *location, char *extend)
...
-		if (set_conn_errtxt(c, dis_emsg[rc]) != 0) {
+		(rc = encode_DIS_ReqExtend(c, extend)))
+	{
+		if (set_conn_errtxt(c, dis_emsg[rc]) != 0)
+		{
 			pbs_errno = PBSE_SYSTEM;
-		} else {
+		}
+		else
+		{
```
bhroam, line 147: Undo the extra newlines added

arungrover:
LGTM

agrawalravi90:
Thanks guys, I'll do a rebase so that I can run regressions properly.

agrawalravi90:
@bhroam  and @arungrover  I added function pointer for the async ack call, please review the second of the 2 new commits from the rebased changes, thanks

agrawalravi90:
@bhroam  and @arungrover  On further testing, I found an issue: the job's comment wasn't being updated to reflect a runjob hook reject if sched submitted async runjob request. So, I've added code to make the server update the job comment when this happens. Also added a couple of new tests. Please review the last commit, hopefully there won't be any further changes. Thanks!

src/server/hook_func.c
```diff
@@ -4530,6 +4529,18 @@ int server_process_hooks(int rq_type, char *rq_user, char *rq_host, hook *phook,
...
+				char *jcomment = NULL;
+
+				pbs_asprintf(&jcomment, "Not Running: PBS Error: %s", hook_msg);
+				/* For async run, sched won't update job's comment, so let's do that */
+				job_attr_def[(int)JOB_ATR_Comment].at_decode(
+					&pjob->ji_wattr[(int)JOB_ATR_Comment],
+					NULL,
+					NULL,
+					jcomment);
+			}
```
bhroam, line 4543: you need to free jcomment.  pbs_asprintf() will have allocated it.

test/tests/functional/pbs_sched_runjobwait.py
```diff
@@ -327,3 +327,73 @@ def test_throughput_ok(self):
...
+        self.server.create_import_hook('rj', hk_attrs, hook_txt)
+
+        t1 = time.time()
+        jid4 = self.server.submit(Job())
+        self.server.expect(JOB, {"job_state": "Q"}, id=jid4)
+        self.server.expect(JOB, "comment", op=SET, id=jid4)
+        ret = self.server.status(JOB, id=jid4)
+        jdata = ret[0]
+        comment = jdata["comment"]
+        self.assertIn("no walltime specified", comment)
```
bhroam, line 368: You can do this with an expect() with MATCH_RE

test/tests/functional/pbs_sched_runjobwait.py
```diff
@@ -327,3 +327,73 @@ def test_throughput_ok(self):
...
+        self.server.log_match("Type 96 request", starttime=t1, max_attempts=5)
```
bhroam, line 369: Should this be in the runjob hook test plan?

test/tests/functional/pbs_sched_runjobwait.py
```diff
@@ -327,3 +327,73 @@ def test_throughput_ok(self):
...
+        self.server.create_import_hook('rj', hk_attrs, hook_txt)
+
+        t1 = time.time()
+        jid = self.server.submit(Job())
+        self.server.expect(JOB, {"job_state": "Q"}, id=jid)
+        self.server.expect(JOB, "comment", op=SET, id=jid)
+        ret = self.server.status(JOB, id=jid)
+        jdata = ret[0]
+        comment = jdata["comment"]
+        self.assertIn("no walltime specified", comment)
```
bhroam, line 397: Same as above, use expect() with MATCH_RE

src/server/req_stat.c
```diff
@@ -691,8 +692,43 @@ status_node(struct pbsnode *pnode, struct batch_request *preq, pbs_list_head *ps
...
+	for (phook = (hook *) GET_NEXT(svr_runjob_hooks);
+	     phook != NULL;
+	     phook = (hook *) GET_NEXT(phook->hi_runjob_hooks)) {
+		if (phook->enabled) {
+			pattr->at_val.at_long = 1;
+			break;
+		}
+	}
+
+	pattr->at_flags |= ATR_VFLAG_SET | ATR_VFLAG_MODCACHE;
```
bhroam, line 730: I'd rather you check if you need to change it before you change it.  This way you won't have to invalidate the cache all the time.

src/server/req_stat.c
```diff
@@ -722,6 +758,11 @@ req_stat_svr(struct batch_request *preq)
 	update_license_ct(&server.sv_attr[(int)SRV_ATR_license_count],
 		server.sv_license_ct_buf);
 
+	if (preq->rq_fromsvr) {
```
bhroam, line 761: fromsvr or do you want to check if it is coming from the scheduler?

src/tools/pbs_python.c
```diff
@@ -749,7 +749,7 @@ encode_svrstate(const attribute *pattr, pbs_list_head *phead, char *atname, char
 }
 
 int
-encode_isrunhook(attribute *attr, pbs_list_head *phead, char *atname, char *rsname, int mode, svrattrl **rtnl)
+encode_isrunhook(const attribute *attr, pbs_list_head *phead, char *atname, char *rsname, int mode, svrattrl **rtnl)
```
bhroam, line 752: You just got rid of this function.  You can remove this.

src/server/req_stat.c
agrawalravi90: Is there a better way to check that it's coming from the scheduler? I noticed that this flag is set when it is requested by sched so I used it.

agrawalravi90:
@bhroam  I've addressed your comments, please provide further feedback, thanks!

src/server/req_stat.c
bhroam: I think you check the connection against the scheduler connection.  Look for any place we log a connection coming from the scheduler and see how it does it.  If it uses this flag, then great.

agrawalravi90: Ok ya, I think the other parts look at the job, get its psched object and compare against the sched connection. Since this is server stat we'll have to loop over the list of all sched objects and match the connection. Should I do that instead?

hirenvadalia: you can do this:
```c
conn_t *conn = get_conn(preq->rq_conn);
if ((conn->cn_authen & PBS_NET_CONN_TO_SCHED) != 0)) {
   < connection to sched >
} else {
   < connection not to sched >
}
```
or

```c
if (strcmp(preq->rq_user, PBS_SCHED_DAEMON_NAME) == 0) {
  < connection to sched >
} else {
  < connection not to sched >
}
```

hirenvadalia: For example you can see process_request() when its change rq_user to PBS_SCHED_DAEMON_NAME if connection is to sched.

src/include/pbs_share.h
```diff
@@ -106,6 +106,11 @@
 
 #define MAX_INT_LEN 10
 
+/* Values of sched attribute 'job_run_wait' */
+#define RW_NONE "none"
```
hirenvadalia, line 110: Just a suggestion: we should rename this to REPLY_WAIT_*  or maybe just WAIT_* and it took some time to get that RW means reply wait (generally RW stand for read-write) 😄

src/server/req_stat.c
agrawalravi90: Thanks a lot Hiren, I've used your first suggestion since it seems to be faster than comparing strings, and it seems to work.

bhroam: I would test this with a multisched just incase there is a difference between the default and a non-default scheduler.

agrawalravi90: I have a PTL test which checks that multisched works fine, and it passed after these changes. see "test_multisched_multival"

src/server/sched_func.c
```diff
@@ -966,39 +965,43 @@ set_sched_default(pbs_sched *psched, int from_scheduler)
...
+	if (!(psched->sch_attr[SCHED_ATR_throughput_mode].at_flags & ATR_VFLAG_SET) &&
+			strcmp(psched->sch_attr[SCHED_ATR_job_run_wait].at_val.at_str, RUN_WAIT_NONE)) {
+		set_attr_svr(&(psched->sch_attr[SCHED_ATR_throughput_mode]), &sched_attr_def[SCHED_ATR_throughput_mode],
+				ATR_TRUE);
+		psched->sch_attr[SCHED_ATR_throughput_mode].at_flags |= ATR_VFLAG_DEFLT;
+	}
 
-	if (flag)
-		set_scheduler_flag(SCH_ATTRS_CONFIGURE, psched);
+	set_scheduler_flag(SCH_ATTRS_CONFIGURE, psched);
```
bhroam, line 1004: Before we would only send this scheduler command if we set anything.  Now we always send it.  Why the change?

src/server/stat_job.c
```diff
@@ -112,16 +112,15 @@ svrcached(attribute *pat, pbs_list_head *phead, attribute_def *pdef)
 		(server.sv_attr[(int)SRV_ATR_show_hidden_attribs].at_val.at_long == 0)) {
 		return;
 	}
+	if (pat->at_flags & ATR_VFLAG_MODCACHE)
+		/* free old cache value if the value has changed */
+		free_svrcache(pat);
```
bhroam, line 117: If you free pat here, won't you just dereference it on lines 120/122?

src/include/pbs_ifl.h
```diff
@@ -360,6 +361,7 @@ extern "C" {
 #define ATTR_sched_preempt_order  "preempt_order"
 #define ATTR_sched_preempt_sort  "preempt_sort"
 #define ATTR_sched_server_dyn_res_alarm "server_dyn_res_alarm"
+#define ATTR_job_run_wait "job_run_wait"
```
bhroam, line 364: Since the functions are called pbs_runjob() or or similar, should this be runjob_wait?

test/tests/functional/pbs_sched_runjobwait.py
bhroam: I don't see a response to this comment.  I see no reason why you shouldn't use expect() here (and below)

src/server/stat_job.c
agrawalravi90: thanks for catching that!

src/include/pbs_ifl.h
agrawalravi90: this was a design discussion point, Scott had mentioned that he preferred job_run_wait instead of runjob_wait because we already have an existing association of the word "runjob" with hooks.

test/tests/functional/pbs_sched_runjobwait.py
agrawalravi90: Oops, looks like I completely missed your comments on the test file. I'll make this change, thanks for the suggestion.

src/include/pbs_ifl.h
bhroam: Thanks for reminding me.  I thought there must be a reason.  You may want to rename your tests since they are test_runjob_wait_X

agrawalravi90:
@hirenvadalia Gentle reminder, please review the last few commits whenever you get time. Thanks!

src/server/sched_func.c
```diff
@@ -1070,3 +1081,110 @@ action_sched_server_dyn_res_alarm(attribute *pattr, void *pobj, int actmode)
...
+ * @return error code
+ */
+int
+action_throughput_mode(attribute *pattr, void *pobj, int actmode)
+{
+	long val = pattr->at_val.at_long;
+	pbs_sched *psched = NULL;
+
+	psched = (pbs_sched *) pobj;
+	if (actmode == ATR_ACTION_ALTER || actmode == ATR_ACTION_RECOV) {
```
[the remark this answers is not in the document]
agrawalravi90: Good catch, thanks!

agrawalravi90:
@arungrover thanks for the review! I've addressed your comments, please let me know if it looks ok now.

PR #1654, opened 2020-04-14
PTL: Add mom states in MoM.isUp method

varunsonkar:
#### Describe Bug or Feature
In PTL the isUp method of MoM class supports only few states like free, provisioning. So when tests are written to restart mom when the mom is in the states like "offline" or "Job-Busy" this method fails. 

#### Describe Your Change
The fix is to check for other states like offline and job-busy in the MoM.isUp() method as well. 

#### Test_logs:
[test_isUp.txt](https://github.com/PBSPro/pbspro/files/4473014/test_isUp.txt)

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -13135,7 +13135,7 @@ class MoM(PBSService):
             try:
                 nodes = self.server.status(NODE, id=self.shortname)
                 if nodes:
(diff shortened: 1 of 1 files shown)


bhroam:
I know this is merged, but I'll comment anyway.  You missed 'maintenance'.

With that in mind, why not look for the negative instead of the positive?  Look for not equal to down or down,offline.  Or if you can do a RE search with NE, then do that with just 'down'

PR #1656, opened 2020-04-14
Remove PBS_ENCRYPT_MODE config key and few other auth related fixes

hirenvadalia:
#### Describe Bug or Feature

* Recently we added new pbs.conf key called PBS_ENCRYPT_MODE, which specified which communication to encrypt and which one, but we don't need that key, because its client-side configuration, and when the client sets PBS_ENCRYPT_METHOD which implies that client wants to encrypt then there is no question of what to encrypt or not. Now if the client doesn't set PBS_ENCRYPT_METHOD that means the client doesn't want encryption, which means no question of what to encrypt or what not to. Means, PBS_ENCRYPT_MODE is useless since we PBS_ENCRYPT_METHOD key
* currently if PBS_SUPPORTED_AUTH_METHODS is configured, and if resvport is not listed in it still PBS allows resvport authentication, which it should as per design
* Currently we MUST have to list whatever PBS_ENCRYPT_METHOD client is using into either PBS_SUPPORTED_AUTH_METHODS or PBS_ENCRYPT_METHOD on server-side, even server doesn't want to use it.

#### Describe Your Change

* Remove PBS_ENCRYPT_MODE
* Enforce PBS_SUPPORTED_AUTH_METHODS if configured
* Remove requirement listing encrypt_method in PBS_SUPPORTED_AUTH_METHODS or PBS_ENCRYPT_METHOD on server even server doesn't want to use encryption

#### Link to Design Doc

* [Here](https://pbspro.atlassian.net/wiki/spaces/PD/pages/1525776385/Allow+multiple+auth+methods+in+PBSPro)

Screen Shot 2020-04-14 at 5 54 32 PM

### DIFF
--- a/src/include/auth.h
+++ b/src/include/auth.h
@@ -51,12 +51,6 @@ extern "C" {
 #define FOR_AUTH 0
 #define FOR_ENCRYPT 1
 
-enum ENCRYPT_MODE {
-	ENCRYPT_DISABLE = 0,
-	ENCRYPT_ONLY_CLIENT_TO_SERVER,
-	ENCRYPT_ALL
-};
-
 enum AUTH_ROLE {
 	AUTH_ROLE_UNKNOWN = 0,
 	AUTH_CLIENT,
@@ -88,9 +82,6 @@ typedef struct pbs_auth_config {
 	/* Name of encryption method (aka PBS_ENCRYPT_METHOD in pbs.conf). This should be a null-terminated string. */
 	char *encrypt_method;
 
-	/* Encryption mode (aka PBS_ENCRYPT_MODE in pbs.conf) */
-	int encrypt_mode;
-
 	/*
 	 * Function pointer to the logging method with the same signature as log_event from Liblog.
 	 * With this, the user of the authentication library can redirect logs from the authentication
@@ -100,7 +91,8 @@ typedef struct pbs_auth_config {
 	void (*logfunc)(int type, int objclass, int severity, const char *objname, const char *text);
 } pbs_auth_config_t;
 
-typedef struct auth_def {
+typedef struct auth_def auth_def_t;
+struct auth_def {
 	/* name of authentication method name */
 	char name[MAXAUTHNAME + 1];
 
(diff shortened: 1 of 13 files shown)


src/include/auth.h
```diff
@@ -141,7 +133,12 @@ typedef struct auth_def {
 	 * the function pointer to decrypt data
 	 */
 	int (*decrypt_data)(void *ctx, void *data_in, size_t len_in, void **data_out, size_t *len_out);
-} auth_def_t;
+
+	/*
+	 * pointer to next authdef structure
+	 */
+	auth_def_t *next;
```
agrawalravi90, line 140: any particular reason why you are changing this from an array to a linked list?

src/lib/Libifl/auth.c
```diff
@@ -242,18 +242,25 @@ _unload_auth(auth_def_t *auth)
...
+	auth_def_t *auth = loaded_auths;
 
-	if (auths == NULL)
-		return NULL;
-
-	while (auths[i] != NULL) {
-		if (strcmp(auths[i]->name, method) == 0)
-			return auths[i];
-		i++;
+	while (auth != NULL) {
```
agrawalravi90, line 247: this really should be a for loop:
for (auth = loaded_auths; auth != NULL; auth = auth->next)

src/lib/Libifl/auth.c
```diff
@@ -339,17 +332,14 @@ load_auths(int mode)
 void
 unload_auths(void)
 {
-	int i = 0;
-
-	if (auths == NULL)
+	if (loaded_auths == NULL)
```
agrawalravi90, line 335: you don't really need this if condition, the while loop won't be executed anyways if it is NULL

src/include/auth.h
hirenvadalia: Yes, @agrawalravi90, earlier we were loading authdefs only which was listed in PBS_SUPPORTED_AUTH_METHODS, PBS_AUTH_METHOD, and PBS_ENCRYPT_METHOD (aka fixed set of authdefs), so I used array and now as I mentioned in the PR description about removing requirement listing PBS_ENCRYPT_METHOD of the client in PBS_SUPPORTED_AUTH_METHODS or PBS_ENCRYPT_METHOD on the server, even if server is not using them.

So now until the client connects with its selected PBS_ENCRYPT_METHOD we don't how what to load for authdef, and so unknown size of loaded authdefs...
+
the pointer to authdefs is bound to per connection struct (aka `pbs_tcp_chan_t->auths[]->def`, see dis.h), now if use array, and if new encrypt method comes which we need to load and add it to the array so we have to realloc() array which will change the pointers of authdef, which will lead bad pointer in connection struct

HTH!

agrawalravi90: Got it, thanks for explaining

PR #1659, opened 2020-04-15
Fix for invalid read and write in munge handshake

hirenvadalia:
#### Describe Bug or Feature

* Valgrind reports invalid read and write in munge handshake due to invalid length of data and missing null char in data while generating and validating handshake data

#### Describe Your Change

* Use correct length which includes null char also in handshake data

#### Link to Design Doc

* None

[before.txt](https://github.com/PBSPro/pbspro/files/4479434/before.txt)
[after.txt](https://github.com/PBSPro/pbspro/files/4479432/after.txt)


hirenvadalia:
Thanks @mike0042  and @npadole20

PR #1660, opened 2020-04-15
Degreaded running reservations could be reconfirmed on wrong vnodes

bhroam:
#### Describe Bug or Feature
When the scheduler reconfirms a running reservation, it disregards non-consumables.  This means a running reservation could be reconfirmed on the wrong set of nodes

#### Describe Your Change
To deal with running requests, the scheduler creates an internal version of the schedselect based on the exec_vnode.  The exec_vnode only has consumable resources in it.  This means the non-consumables were lost when looking to replace nodes.  The fix is to map the schedselect to the exec_vnode and use the resource request from the chunk in in the schedselect instead of the exec_vnode.

[resv.log](https://github.com/PBSPro/pbspro/files/4483446/resv.log)

### DIFF
--- a/test/tests/functional/pbs_reservations.py
+++ b/test/tests/functional/pbs_reservations.py
@@ -239,6 +239,7 @@ class TestReservations(TestFunctional):
 
         self.server.expect(RESV, resv_state, id=rid)
 
+    @skipOnCpuSet
     def test_degraded_standing_reservations(self):
         """
         Verify that degraded standing reservations are reconfirmed
@@ -248,6 +249,7 @@ class TestReservations(TestFunctional):
         self.degraded_resv_reconfirm(start=t + 25, end=t + 625,
                                      rrule='freq=HOURLY;count=5')
 
+    @skipOnCpuSet
     def test_degraded_advance_reservations(self):
         """
         Verify that degraded advance reservations are reconfirmed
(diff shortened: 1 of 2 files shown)


arungrover:
@bhroam Please amend the typo in commit message

test/tests/functional/pbs_reservations.py
```diff
@@ -90,20 +90,34 @@ def submit_asap_reservation(self, user, jid, extra_attrs=None):
 
         return self.server.submit(r)
 
-    @skipOnCpuSet
```
vstumpf, line 93: Why are you removing this? This would cause the tests that call this function to fail on cpuset.

test/tests/functional/pbs_reservations.py
```diff
@@ -142,7 +156,6 @@ def degraded_resv_reconfirm(self, start, end, rrule=None, run=False):
 
         self.server.expect(RESV, a, id=rid, interval=1)
 
-    @skipOnCpuSet
```
vstumpf, line 145: Why are you removing this? This would cause the tests that call this function to fail on cpuset.

test/tests/functional/pbs_reservations.py
bhroam: It's a helper function, not a test.  The tests do have the decorator,

vstumpf: I'm on master and I don't see the tests having the decorator. That being said, I'm not sure if the decorator works in a helper function. Should you put the decorator on the tests?

bhroam: Yeah, I must have been looking at different tests.  I've added the skip decorator now.

PR #1661, opened 2020-04-15
Scheduler runs jobs out of fairshare order after jobs are added to the calendar

bhroam:
#### Describe Bug or Feature
When fairshare is enabled and a job is added to the calendar, the entire usage of that job should be added to the user's usage for that cycle only.  This is to stop us from running more jobs from that user that will eventually kick the top job out of its slot.  We do add the usage, but we never resort the jobs.  This meant the political ordering was not changed due to the increase in usage.

#### Describe Your Change
Resort the jobs if a job is added to the calendar and fairshare is enabled.

[fairshare.log](https://github.com/PBSPro/pbspro/files/4494703/fairshare.log)

### DIFF
--- a/src/scheduler/fifo.c
+++ b/src/scheduler/fifo.c
@@ -869,6 +869,7 @@ main_sched_loop(status *policy, int sd, server_info *sinfo, schd_error **rerr)
 		comment[0] = '\0';
 		log_msg[0] = '\0';
 		qinfo = njob->job->queue;
+		sort_again = SORTED;
 
 		clear_schd_error(err);
 
--- a/src/scheduler/fifo.c
+++ b/src/scheduler/fifo.c
@@ -945,10 +945,8 @@ main_sched_loop(status *policy, int sd, server_info *sinfo, schd_error **rerr)
 			int cal_rc;
 #ifdef NAS /* localmod 034 */
 			int bf_rc;
-			sort_again = SORTED;
 			if ((bf_rc = site_should_backfill_with_job(policy, sinfo, njob, num_topjobs, num_topjobs_per_queues, err)))
 #else
-			sort_again = SORTED;
 			if (should_backfill_with_job(policy, sinfo, njob, num_topjobs) != 0) {
 #endif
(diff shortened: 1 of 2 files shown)


src/scheduler/fifo.c
```diff
@@ -945,10 +945,8 @@ main_sched_loop(status *policy, int sd, server_info *sinfo, schd_error **rerr)
 			int cal_rc;
 #ifdef NAS /* localmod 034 */
 			int bf_rc;
-			sort_again = SORTED;
 			if ((bf_rc = site_should_backfill_with_job(policy, sinfo, njob, num_topjobs, num_topjobs_per_queues, err)))
 #else
-			sort_again = SORTED;
```
arungrover, line 951: Isn't this flag needed? If we remove it then what might happen is that we may send MAY_RESORT_JOBS to next_job more often than needed and it might result into sorting the jobs again.
Like if fairshare is enabled, one job couldn't run and was a calendared we will set MAY_RESORT_JOBS and then the next job also couldn't run and that job isn't calendared, we will send MAY_RESORT_JOBS again to next_job.

bhroam: This is tricky.  We may have already set it to MAY_RESORT_JOBS based on preemption not working.  We don't want to arbitrarily set it to SORTED here.  We may not even add the job to the calendar.  How about I move this up to the top of the main loop where we initialize other variables.  This way if it is set to anything other than SORTED, we know it came from this iteration of the main loop.

arungrover: But we don't set it to MAY_RESORT_JOBS when preemption isn't working, right? We try preemption and if we cannot preempt flag is set it as SORTED.

bhroam: Sorry, I misspoke.  Yes, if preemption doesn't work, we set it to SORTED.  If it does work, we set it to MAY_RESORT_JOBS.  We'll lose that if we set it to SORTED here.  I'll just initialize the variable at the start of the main loop

PR #1664, opened 2020-04-16
Skip chdir for Shasta

visheshh:
#### Describe Bug or Feature
os.chdir fails as the submit_dir is on a client machine on Shasta platform

#### Describe Your Change
If the platform is Shasta skip the chdir


visheshh:
@PBSPro/pbspro-maintainers , Please review

visheshh:
> Is this not enough? Should you remove the platform check from this if statement?

Hi @vstumpf , Yes this is not enough, If submit_dir has value it still tries to do chdir. I removed the check from eariler if condition.

PR #1669, opened 2020-04-17
Cred related errors with stage in failure on Windows cluster

Bhagat-Rajput:
#### Describe Bug or Feature
credential data gets truncated due to EOF char in it.
#### Describe Your Change
Encoding the binary credential data to text using base64 encoding and decoding back after receival

[win_stage_test_logs.txt](https://github.com/PBSPro/pbspro/files/4492530/win_stage_test_logs.txt)
[val_win_stage_err.txt](https://github.com/PBSPro/pbspro/files/4492571/val_win_stage_err.txt)

### DIFF
--- a/src/lib/Libutil/pbs_aes_encrypt.c
+++ b/src/lib/Libutil/pbs_aes_encrypt.c
@@ -7,6 +7,8 @@
 
 #include <openssl/evp.h>
 #include <openssl/aes.h>
+#include <openssl/bio.h>
+#include <openssl/buffer.h>
 #include "ticket.h"
 /**
  * @file	pbs_aes_encrypt.c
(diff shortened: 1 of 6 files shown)


src/lib/Libutil/pbs_aes_encrypt.c
```diff
@@ -150,3 +152,101 @@ pbs_decrypt_pwd(char *crypted, int credtype, size_t len, char **uncrypted)
 
         return 0;
 }
+
+/**
+ * @brief
+ *	encode_to_base64 - Encode data into bas64 format
```
npadole20, line 158: typo: base64

src/lib/Libutil/pbs_aes_encrypt.c
```diff
@@ -150,3 +152,101 @@ pbs_decrypt_pwd(char *crypted, int credtype, size_t len, char **uncrypted)
...
+	mem_obj1 = BIO_new(BIO_s_mem());
+	mem_obj2 = BIO_new(BIO_f_base64());
+
+	mem_obj1 = BIO_push(mem_obj2, mem_obj1);
+	BIO_set_flags(mem_obj1, BIO_FLAGS_BASE64_NO_NL);
+	BIO_write(mem_obj1, buffer, buffer_len);
+	(void)BIO_flush(mem_obj1);
+	BIO_get_mem_ptr(mem_obj1, &b_mem);
+
+	*ret_encoded_data = (*b_mem).data;
```
npadole20, line 184: Why are you doing this? In the next step you are using this pointer to malloc.

Bhagat-Rajput: yes, there is no sense of this step since we are using memcpy now. Thanks

src/lib/Libutil/pbs_aes_encrypt.c
```diff
@@ -150,3 +152,101 @@ pbs_decrypt_pwd(char *crypted, int credtype, size_t len, char **uncrypted)
...
+
+	*ret_encoded_data = malloc(buf_len + 1);
+	if (*ret_encoded_data == NULL) {
+		BIO_free_all(mem_obj1);
+		return (1);
+	}
+	memcpy(*ret_encoded_data, (*b_mem).data, buf_len);
+	(*ret_encoded_data)[buf_len] = '\0';
+
+	BIO_free_all(mem_obj1);
```
npadole20, line 195: Why is mem_obj2 not freed?

Bhagat-Rajput: BIO_free_all() call frees up an entire BIO chain

src/resmom/stage_func.c
```diff
@@ -2284,30 +2329,39 @@ recv_rq_cpyfile_cred(struct rq_cpyfile *pcf)
...
+					log_err(-1, __func__, log_buffer);
+					return 0;
+				}
 			}
 
-			a_cnt = fread(cred_buf, sizeof(char), cred_len+1, stdin);
-			cred_buf[cred_len+1] = '\0';
-			if (cred_len+1 != a_cnt) {
+			if (cred_len != a_cnt) {
+				sprintf(log_buffer, "Mismatch: cred_len = %d, a_cnt=%d", cred_len, a_cnt);
```
npadole20, line 2354: Can we not use 'a_cnt' and instead mention that we are referring to decoded cred length? I won't know what a_cnt is just from the logs unless I go through the code.

src/lib/Libutil/pbs_aes_encrypt.c
```diff
@@ -150,3 +152,100 @@ pbs_decrypt_pwd(char *crypted, int credtype, size_t len, char **uncrypted)
...
+ * @retval 1 - for failure
+ */
+int
+encode_to_base64(const unsigned char* buffer, size_t buffer_len, char** ret_encoded_data)
+{
+	BUF_MEM *b_mem;
+	BIO *mem_obj1, *mem_obj2;
+	int buf_len = 0;
+
+	mem_obj1 = BIO_new(BIO_s_mem());
```
hirenvadalia, line 175: Don't you have to check for return value for BIO_new?

src/lib/Libutil/pbs_aes_encrypt.c
```diff
@@ -150,3 +152,100 @@ pbs_decrypt_pwd(char *crypted, int credtype, size_t len, char **uncrypted)
...
+	mem_obj2 = BIO_new(BIO_f_base64());
+
+	mem_obj1 = BIO_push(mem_obj2, mem_obj1);
+	BIO_set_flags(mem_obj1, BIO_FLAGS_BASE64_NO_NL);
+	BIO_write(mem_obj1, buffer, buffer_len);
+	(void)BIO_flush(mem_obj1);
+	BIO_get_mem_ptr(mem_obj1, &b_mem);
+
+	buf_len = (*b_mem).length;
```
hirenvadalia, line 184: I think right way of access data and length in BIO_MEM is to use `BIO_get_mem_data()`

Bhagat-Rajput: one step above at line 182 we are retrieving a BIO_MEM pointer using BIO_get_mem_ptr() call. is it not a valid call ?

hirenvadalia: Its valid call but I don't think you need that, you can directly do `buf_len = BIO_get_mem_data(mem_obj1, &buf)` define `char *buf`

subhasisb: > Its valid call but I don't think you need that, you can directly do `buf_len = BIO_get_mem_data(mem_obj1, &buf)` define `char *buf`

If it is valid, we do not need to change. The only thing we missed is the check for BIO_new() failing, and adding that is sufficient

subhasisb: > > Its valid call but I don't think you need that, you can directly do `buf_len = BIO_get_mem_data(mem_obj1, &buf)` define `char *buf`
> 
> If it is valid, we do not need to change. The only thing we missed is the check for BIO_new() failing, and adding that is sufficient

Actually, i take it back. Let's fix this also. I am just worried we have to do another test, but i think this is safe enough

hirenvadalia: @subhasisb Generally `BIO_get_mem_ptr` is used with `BIO_set_mem_buf()`, and accessing .length and .data is not documentented in OpenSSL docs, it says to use `BIO_get_mem_data()` to data from BUF_MEM structure (https://www.openssl.org/docs/man1.0.2/man3/BIO_set_mem_buf.html). But up to you guys...

subhasisb: > @subhasisb Generally `BIO_get_mem_ptr` is used with `BIO_set_mem_buf()`, and accessing .length and .data is not documentented in OpenSSL docs, it says to use `BIO_get_mem_data()` to data from BUF_MEM structure (https://www.openssl.org/docs/man1.0.2/man3/BIO_set_mem_buf.html). But up to you guys...

Yes the documentation is insufficient. However, consider this. BIO_get_mem_ptr() is a first class API documented in Openssl docs. It is supposed to return BUF_MEM structure. Therefore the structure is documented and you CAN access it's elements. Consider this snippet of code in the openssl docs https://linux.die.net/man/3/bio_s_mem  
BUF_MEM *bptr;
BIO_get_mem_ptr(mem, &bptr);
BIO_set_close(mem, BIO_NOCLOSE); /* So BIO_free() leaves BUF_MEM alone */
BIO_free(mem);

What do you do with bptr now, if you cannot access its contents?

hirenvadalia: I am not denying that we can't access .length and .data in BUF_MEM, I am just saying we have a better way (using documented API) of doing what we are trying to do here.
But as I said, it's up to you guys 😄

subhasisb: > I am not denying that we can't access .length and .data in BUF_MEM, I am just saying we have a better way (using documented API) of doing what we are trying to do here.
> But as I said, it's up to you guys 😄

Ok so that is a valid access, you agree? I thought you are saying the access is invalid :)

hirenvadalia: Yes, accessing .length and .data in BUF_MEM is absolutely valid.

PR #1672, opened 2020-04-17
Reservation starts unexpectedly on slow machine in ralter test

bhroam:
#### Describe Bug or Feature
Test TestPbsResvAlter.test_alter_advance_resv_end_time_before_run expects to be able to submit a reservation and alter it three times before it starts.  The reservation starts in 10s, so on slow machines it can start before all the operations are complete.

### Describe Your Change
Since this whole test is about testing operations prior to a reservation starting, it doesn't matter when the reservation starts.  I changed it to start in 60s instead of 10.

[ralter.log](https://github.com/PBSPro/pbspro/files/4495671/ralter.log)

### DIFF
--- a/test/tests/functional/pbs_ralter.py
+++ b/test/tests/functional/pbs_ralter.py
@@ -542,7 +542,7 @@ class TestPbsResvAlter(TestFunctional):
         """
         duration = 20
         shift = 10
-        offset = 10
+        offset = 60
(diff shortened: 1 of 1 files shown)


bhroam:
I agree with you @vstumpf.  Fortunately for this test, it doesn't matter how long in the future you make the reservation.  It won't slow things down on faster machines.

PR #1679, opened 2020-04-21
Fixed type for create_resv_from_job

bhroam:
#### Describe Bug or Feature
The create_resv_from_job attribute was listed in the design as a boolean.  It was implemented as a long.  This meant you couldn't say -Wcreate_resv_from_job=True and the python type was int so you couldn't do the same thing in hooks.

#### Describe Your Change
Changed the type from long to boolean.  Also removed some unneeded job related flags from the reservation reserve_job attribute.

#### Link to Design Doc
https://pbspro.atlassian.net/wiki/spaces/PD/pages/1450409985/Creating+reservation+out+of+a+job

This change isn't really worth an automated test.  Here's a manual test;

[bmann@mars server]$ qsub -Wcreate_resv_from_job=True -- /bin/sleep 1000
486.mars
[bmann@mars server]$ qsub -Wcreate_resv_from_job=false -- /bin/sleep 1000
488.mars
[bmann@mars server]$ qsub -Wcreate_resv_from_job=foo -- /bin/sleep 1000
qsub: Illegal attribute or resource value create_resv_from_job
[bmann@mars server]$ qsub -Wcreate_resv_from_job=1 -- /bin/sleep 1000
491.mars

[bmann@mars server]$ qstat -f | egrep '(^Job)|create_resv' | grep -v Submit
Job Id: 486.mars
    create_resv_from_job = True
Job Id: 488.mars
    create_resv_from_job = False
Job Id: 491.mars
    create_resv_from_job = True

[bmann@mars ~]$ cat qj.py 
import pbs
pbs.event().job.create_resv_from_job=True
[bmann@mars ~]$ qsub -- /bin/sleep 1000
493.mars
[bmann@mars ~]$ qstat -f | grep create
    create_resv_from_job = True

### DIFF
--- a/src/lib/Libattr/master_job_attr_def.xml
+++ b/src/lib/Libattr/master_job_attr_def.xml
@@ -2756,21 +2756,21 @@
       <member_name>
          <both>ATTR_create_resv_from_job</both>
       </member_name>
-      <member_at_decode>decode_l</member_at_decode>
-      <member_at_encode>encode_l</member_at_encode>
-      <member_at_set>set_l</member_at_set>
-      <member_at_comp>comp_l</member_at_comp>
+      <member_at_decode>decode_b</member_at_decode>
+      <member_at_encode>encode_b</member_at_encode>
+      <member_at_set>set_b</member_at_set>
+      <member_at_comp>comp_b</member_at_comp>
       <member_at_free>free_null</member_at_free>
       <member_at_action>NULL_FUNC</member_at_action>
       <member_at_flags>
          <both>READ_WRITE</both>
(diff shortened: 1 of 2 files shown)


bayucan:
The change looks fine and straightforward. With type being ATR_TYPE_BOOL, the hook side of things would map it to a Python bool. Can you also just show a trace of hooks testing, like setting the attribute in the hook that is allowed...
pbs.event().job.create_resv_from_job = True

bhroam:
Thanks for looking at this @bayucan.  I've updated the testing section with a hook example.

PR #1681, opened 2020-04-22
Cleaning licensing code

Jitendra-Rajak:
#### Describe Bug or Feature
Cleaning up license stuff that is not required in open source.

#### Describe Your Change
Open source does not require license stuff, so here we are removing the code related to license and added the changes which are required.


src/include/pbs_license.h
```diff
@@ -90,33 +89,31 @@ typedef enum node_topology_type ntt_t;
...
-	LIC_TRIAL,	/* nonzero number of trial license (to license CPUs) */
 	LIC_UNKNOWN  /* used to hold the value of previous lb */
 };
-
+struct	pbs_lic_counts {
+	long socket_lic_needed;
+	long node_lic_needed;
+	long nsockets;
+	long node_count;
+};
```
nithinj, line 101: As discussed, we only need two members in this structure. nsockets and nnodes. This structure can be also used within node structure instead of creating two different variables.

src/server/svr_func.c
```diff
@@ -1155,32 +1148,14 @@ unset_license_location(void)
...
-			LOG_ALERT, msg_daemonname,
-			"One or more PBS license keys are invalid, jobs may not run");
-	} else {
-		sprintf(log_buffer,
-			"Licenses valid for %d floating hosts",
-			licenses.lb_aval_floating);
-		log_event(PBSEVENT_ADMIN, PBS_EVENTCLASS_SERVER,
-			LOG_NOTICE, msg_daemonname, log_buffer);
-		relicense_svr_unlicensedjobs();
+		licstate_unconfigured(LIC_NODES);
```
nithinj, line 1157: Try combining them.

src/include/pbs_license.h
```diff
@@ -90,43 +89,40 @@ typedef enum node_topology_type ntt_t;
...
-extern int    sockets_total(void);
 extern int    sockets_consume(int);
 extern void   sockets_release(int);
-extern void   sockets_reset(void);
 extern void   inspect_license_path(void);
 extern int    licstate_is_up(enum licensing_backend);
 extern void   licstate_down(void);
 extern void   licstate_unconfigured(enum licensing_backend);
-extern int	nsockets_from_topology(char *, ntt_t);
+extern int	nsockets_from_topology(char *, ntt_t, long *, long *);
```
nithinj, line 119: Looks like this function is not being used. Can we remove it?

src/include/pbs_license.h
nithinj: This should be declared in pbs_nodes.h and used here. Also can you make the names more generic. like, nsockets and nnodes inside struct devices/inventory?

src/server/req_stat.c
```diff
@@ -923,16 +887,12 @@ void
...
-		print_license_ct(LIC_SOCKETS, buf);
-	else if (licenses.lb_trial == 1)
-		print_license_ct(LIC_TRIAL, buf);
-	else
-		print_license_ct(last_valid_attempt, buf);
+	if (licstate_is_up(LIC_NODES) || licstate_is_up(LIC_SOCKETS)) {
+		sprintf(buf, "Avail_Global:%d Avail_Local:%d Used:%d High_Use:%d",
+				licenses.lb_glob_floating, licenses.lb_aval_floating,
+				licenses.lb_used_floating, licenses.lb_high_used_floating);
+	} else
```
nithinj, line 894: Do we need this else case as numbers will be zero anyway if things are not configured?

src/server/req_stat.c
```diff
@@ -923,16 +887,12 @@ void
...
-		print_license_ct(LIC_SERVER, buf);
-	else if (licstate_is_up(LIC_NODES))
-		print_license_ct(LIC_NODES, buf);
-	else if (licstate_is_up(LIC_SOCKETS))
-		print_license_ct(LIC_SOCKETS, buf);
-	else if (licenses.lb_trial == 1)
-		print_license_ct(LIC_TRIAL, buf);
-	else
-		print_license_ct(last_valid_attempt, buf);
+	if (licstate_is_up(LIC_NODES) || licstate_is_up(LIC_SOCKETS)) {
```
nithinj, line 890: This function should be in license_client.c

Jitendra-Rajak: this is a part of stat functions its better to keep it here only

src/include/pbs_nodes.h
```diff
@@ -282,6 +288,7 @@ struct	pbsnode {
 	unsigned short		 nd_accted;	/* resc recorded in job acct */
 	struct pbs_queue	*nd_pque;	/* queue to which it belongs */
 	int			 nd_modified;	/* flag indicating whether state update is required */
+	struct devices lic_counts;
```
nithinj, line 291: Pls use a generic name instead of lic_counts, something like device/inventory.

src/server/svr_func.c
```diff
@@ -1155,34 +1136,16 @@ unset_license_location(void)
 		if (pbs_licensing_license_location[0] != '\0') {
 
 			close_licensing();
-			sockets_reset();
 			unlicense_socket_licensed_nodes();
-
-		} else { /* from no license server */
+			clear_license_info();
```
nithinj, line 1140: I think set_license_location() and unset_license_location() should be moved to license_client.c
I'll leave it up to you to handle it now or later.

Jitendra-Rajak: We can handle that later

Jitendra-Rajak:
Thanks @nithinj for reviews and comments

Jitendra-Rajak:
@PBSPro/pbspro-maintainers , please have look on the PR and take this forward.

PR #1682, opened 2020-04-22
SmokeTest.test_resource_delete timedout on slow machines

dinesh1306:
#### Describe Bug or Feature
SmokeTest.test_resource_delete timedout on slow machines. Need to increase timeout value.

#### Describe Your Change
Increase timeout value on test-case from 720 sec. to 1200 sec. 

Before fix:
[oss_before_fix.txt](https://github.com/PBSPro/pbspro/files/4517886/oss_before_fix.txt)
After fix:
[oss_result.txt](https://github.com/PBSPro/pbspro/files/4517890/oss_result.txt)

### DIFF
--- a/test/tests/functional/pbs_qmgr.py
+++ b/test/tests/functional/pbs_qmgr.py
@@ -41,11 +41,25 @@ from tests.functional import *
 from ptl.lib.pbs_ifl_mock import *
 
 
+@tags('smoke')
 class TestQmgr(TestFunctional):
 
     """
     Test suite for PBSPro's qmgr command
     """
+
+    resc_flags = [None, 'n', 'h', 'nh', 'q', 'f', 'fh', 'm', 'mh']
+    resc_flags_ctl = [None, 'r', 'i']
+    objs = [QUEUE, SERVER, NODE, JOB, RESV]
+    resc_name = "ptl_custom_res"
+    avail_resc_name = 'resources_available.' + resc_name
+
+    def setUp(self):
+        TestFunctional.setUp(self)
+        self.obj_map = {QUEUE: self.server.default_queue,
+                        SERVER: self.server.name,
+                        NODE: self.mom.shortname,
+                        JOB: None, RESV: None}
 
     def __check_whitespace_prefix(self, line):
         """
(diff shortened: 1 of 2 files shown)


test/tests/pbs_smoketest.py
```diff
@@ -1361,7 +1361,7 @@ def delete_resource_helper(self, r, t, f, c, obj_type, obj_id):
                 if d and len(d) > 0:
                     self.assertFalse(ar in d[0])
 
-    @timeout(720)
+    @timeout(1200)
```
bhroam, line 1364: This might be out of scope for this PR, but if this test takes so long (1200s is >11m), is it a good candidate for a smoke test?  Should we move it to a functional test?

hirenvadalia: @bhroam actually test_resource_* in SomkeTest are qmgr resource interface regression, it should be part of interfaces test (We already have a ticket for it in the old issue tracker: https://pbspro.atlassian.net/browse/PP-1226)

dinesh1306: Hi @bhroam ,@hirenvadalia i will remove test_resource_create and test_resource_delete tests from SmokeTest test suite and add slimmed test for resource. 
Please let me know we have TestQmgr test suite in functional. Is that fine to move these tests to TestQmgr or i have to create separate test suite in interfaces as mentioned by @hirenvadalia .

hirenvadalia: I am ok moving into TestQmgr. Also if possible I would suggest dividing test_resource_create and test_resource_delete into multiple that way will have multiple short-running tests instead of long-running tests.

bhroam: I'm fine with this.  Thanks for doing this @dinesh1306

dinesh1306:
@hirenvadalia, @bhroam I have made changes in code. Please have a look.
[res_TestQmgr.txt](https://github.com/PBSPro/pbspro/files/4579681/res_TestQmgr.txt)
[res_SmokeTest.txt](https://github.com/PBSPro/pbspro/files/4579682/res_SmokeTest.txt)

saritakh:
Just for information 'SmokeTest.test_jobscript_max_size' is skipped due to issue: https://github.com/PBSPro/pbspro/issues/1711

test/tests/functional/pbs_qmgr.py
```diff
@@ -41,12 +41,26 @@
 from ptl.lib.pbs_ifl_mock import *
 
 
+@tags('smoke')
```
hirenvadalia, line 44: I am not sure about this, shouldn't this be like "regression" instead? If we keep "smoke" then while it will still run full regression of qmgr resource interface tests, along with newly added tests in SmokeTests, it Means new tests in SmokeTests is kinda redundant test if we are running tests in this suite?

So, our goal was to move qmgr resource interface tests from SmokeTest to this suite, add lightweight new tests in SmokeTest to compensate for remove tests, and mark moved tests as regression, right? Am I missing anything here?

PR #1683, opened 2020-04-22
In test "test_rsc_used" used_walltime should not be zero

birbalsain:
#### Describe Bug or Feature
Test "test_rsc_used" fails intermittently on some platforms because not exact match of used_walltime.

#### Describe Your Change
In accounting logs, Test checks used_walltime should not be zero.

[test_mom_mock.txt](https://github.com/PBSPro/pbspro/files/4519147/test_mom_mock.txt)

### DIFF
--- a/test/tests/functional/pbs_mom_mock_run.py
+++ b/test/tests/functional/pbs_mom_mock_run.py
@@ -68,5 +68,6 @@ class TestMomMockRun(TestFunctional):
         self.server.accounting_match(msg=used_ncpus, id=jid)
         used_mem = "resources_used.mem=5mb"
         self.server.accounting_match(msg=used_mem, id=jid)
-        used_walltime = "resources_used.walltime=00:00:05"
-        self.server.accounting_match(msg=used_walltime, id=jid)
+        used_walltime = "resources_used.walltime=00:00:00"
+        self.server.accounting_match(
+            msg=used_walltime, existence=False, id=jid)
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_mom_mock_run.py
```diff
@@ -68,5 +68,6 @@ def test_rsc_used(self):
         self.server.accounting_match(msg=used_ncpus, id=jid)
         used_mem = "resources_used.mem=5mb"
         self.server.accounting_match(msg=used_mem, id=jid)
-        used_walltime = "resources_used.walltime=00:00:05"
-        self.server.accounting_match(msg=used_walltime, id=jid)
+        used_walltime = "resources_used.walltime=00:00:00"
+        self.server.accounting_match(
+            msg=used_walltime, existence=False, id=jid)
```
bhroam, line 73: doing existence=False without a max_attempts will take 30s.  I see two ways to go about this.  Just add a max_attempts=10 or something is one.  The other way is a little smarter,  You can add an existence match of just 'resources_used.walltime', and then do a existence=False, max_attempts=1 match for what you are looking for.

birbalsain:
@bhroam Thanks for the suggestion, I have done the changes.

vstumpf:
`self.server.accounting_match()` will return: `(x,y)` where `x` is the matching line number and `y` the line itself. You can then use `self.assertNotIn()` to make sure it's not 0.

PR #1687, opened 2020-04-23
Fix support for allow_zero_cpus flag

alexis-cousein:
#### Describe Bug or Feature

There is a missing default for allow_zero_cpus in the defaults, and this will make the hook crash when faced with -lselect=1:ncpus=0 jobs unless the cgroup hook config file contains a value for it.

Also, the code brought in was not actually aware of the new execjob_resize, so the code now takes the opportunaity to make sure all corner cases are "clean" in a way that minimises the differences you could see between a system where the cgroup hook is enabled and one where it is not.

#### Describe Your Change

1. For execjob_begin/resize, if zero cpus are not allowed and the cpuset controller is enabled, when no ncpus are requested on mother superior the execjob_begin will reject the job (and execjob_resize will reject the resize). When no cpus are requested on sister MoMs, then the per-job cpuset will be destroyed (which entails an attempt to get rid of processes to correctly support execjob_resize).

2. execjob_attach and execjob_launch will, if the per-job cpuset cgroup is missing (see 1.), either place processes in the main pbspro.slice cpuset if zero cpus are allowed, or simply reject the event if not (with a clear message and a log message). This is done in add_pids, the routine that actually adds the required PIDs to the "correct" cpuset tasks file.

As a result, if zero cpus is NOT allowed and cpusets are enabled, you can still have exec_vnode containing chunks without cpus on some nodes (e.g. to reserve some custom vnode-level resources on them). But you will not be able to launch processes on those nodes.

#### Link to Design Doc

https://pbspro.atlassian.net/wiki/spaces/PD/pages/1627750401/Support+for+the+cpu+cgroup+controller+and+zero-CPU+jobs

Has been revised.

This was written to make the cgroup hook pass one of the existing PTL tests which fails:

 test_cgroup_assign_resources_mem_only_vnode

which generates this exception:
04/03/2020 12:22:30.411561;0080;pbs_python;Hook;pbs_python;['Traceback (most recent call last):', ' File "<embedded code object>", line 5479, in main', ' File "<embedded code object>", line 935, in invoke_handler', ' File "<embedded code object>", line 965, in _execjob_begin_handler', ' File "<embedded code object>", line 4641, in configure_job', "KeyError: 'allow_zero_cpus'"]

### DIFF
--- a/src/hooks/cgroups/pbs_cgroups.PY
+++ b/src/hooks/cgroups/pbs_cgroups.PY
@@ -973,11 +973,12 @@ class HookUtils(object):
                                  'but job does not request ncpus on '
                                  'mother superior: rejecting job')
                 else:
+                    # will be done in configure_job
                     pbs.logmsg(pbs.EVENT_JOB_USAGE,
                                'cpuset enabled with mandatory ncpus >= 1, '
                                'but job does not request ncpus on host: '
                                'deleting cpuset cgroup')
-                # will be done in configure_job
+
         # Configure the new cgroup
         cgroup.configure_job(event.job.id, jobutil.assigned_resources,
                              node, cgroup, event.type)
@@ -1212,11 +1213,12 @@ class HookUtils(object):
                                  'but job does not request ncpus on '
                                  'mother superior: rejecting resize')
                 else:
+                    # will be done in configure_job
                     pbs.logmsg(pbs.EVENT_ERROR,
                                'cpuset enabled with mandatory ncpus >= 1, '
                                'but job no longer requests ncpus on host: '
                                'deleting cpuset cgroup')
-                # will be done in configure_job
+
         # Configure the cgroup
         cgroup.configure_job(event.job.id, jobutil.assigned_resources,
                              node, cgroup, event.type)
@@ -3491,7 +3493,7 @@ class CgroupUtils(object):
                     pbs.event().reject('No cpuset cgroup present '
                                        'and npcus==0 disallowed')
                 else:
-                    # zero CPU  job: should be attached to root pbspro cpuset
+                    # zero CPU job: should be attached to root pbspro cpuset
                     tasks_file = self._cgroup_path(subsys, 'tasks')
             pbs.logmsg(pbs.EVENT_DEBUG4, '%s: tasks file = %s' %
                        (caller_name(), tasks_file))
--- a/src/hooks/cgroups/pbs_cgroups.PY
+++ b/src/hooks/cgroups/pbs_cgroups.PY
@@ -960,9 +960,28 @@ class HookUtils(object):
         cgroup.assigned_resources = cgroup._get_assigned_cgroup_resources()
         # Create the cgroup(s) for the job
         cgroup.create_job(event.job.id, node)
+        if (cgroup.cfg['cgroup']['cpuset']['enabled']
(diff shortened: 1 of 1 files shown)


src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -960,9 +960,28 @@ class HookUtils(object):
...
+                               'mother superior: rejecting job')
+                    event.reject('cpuset enabled with mandatory ncpus >= 1, '
+                                 'but job does not request ncpus on '
+                                 'mother superior: rejecting job')
+                else:
+                    pbs.logmsg(pbs.EVENT_JOB_USAGE,
+                               'cpuset enabled with mandatory ncpus >= 1, '
+                               'but job does not request ncpus on host: '
+                               'deleting cpuset cgroup')
+                # will be done in configure_job
```
mike0042, line 980: Comment should appear above pbs.logmsg() call.

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -1180,6 +1199,24 @@ class HookUtils(object):
...
+                               'mother superior: rejecting resize')
+                    event.reject('cpuset enabled with mandatory ncpus >= 1, '
+                                 'but job does not request ncpus on '
+                                 'mother superior: rejecting resize')
+                else:
+                    pbs.logmsg(pbs.EVENT_ERROR,
+                               'cpuset enabled with mandatory ncpus >= 1, '
+                               'but job no longer requests ncpus on host: '
+                               'deleting cpuset cgroup')
+                # will be done in configure_job
```
mike0042, line 1219: Comment should appear above pbs.logmsg() call.

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -3444,8 +3482,17 @@ class CgroupUtils(object):
...
+                if (self.cfg['cgroup']['cpuset']['enabled']
+                        and not self.cfg['cgroup']['cpuset']
+                                        ['allow_zero_cpus']):
+                    pbs.logmsg(pbs.EVENT_JOB_USAGE,
+                               'No cpuset cgroup present '
+                               'and npcus==0 disallowed')
+                    pbs.event().reject('No cpuset cgroup present '
+                                       'and npcus==0 disallowed')
+                else:
+                    # zero CPU  job: should be attached to root pbspro cpuset
```
mike0042, line 3494: Two spaces after CPU

lisa-altair:
This is the before and after test results:
[before.txt](https://github.com/PBSPro/pbspro/files/4525970/before.txt)
[after_fix.txt](https://github.com/PBSPro/pbspro/files/4525969/after_fix.txt)

src/hooks/cgroups/pbs_cgroups.PY
alexis-cousein: Corrected.

src/hooks/cgroups/pbs_cgroups.PY
alexis-cousein: Corrected.

src/hooks/cgroups/pbs_cgroups.PY
alexis-cousein: Corrected (one space only)

alexis-cousein:
Changes requested by @mike0042 implemented.

PR #1691, opened 2020-04-24
Chain of dependent jobs is not deleted properly

arungrover:
#### Describe Bug or Feature
When a running job ends or it is deleted, the chain of jobs dependent on the running job, and their dependent jobs are not deleted/released properly.
This change will delete/release the dependent jobs and their dependent jobs.

#### Describe Your Change
With this change in place - 

- If a running job is deleted (with -Wforce or not) the chain of all dependent jobs is deleted.
- If a Held job is deleted, chain of all dependent jobs is also deleted.
- While deleting the chain, only jobs that are scheduled to run after the said job are deleted.

#### Link to Design Doc
This was a bug in the product, to begin with. No design changes.

[test_perf.txt](https://github.com/PBSPro/pbspro/files/4531902/test_perf.txt)
[test.txt](https://github.com/PBSPro/pbspro/files/4531903/test.txt)

### DIFF
--- a/src/scheduler/job_info.c
+++ b/src/scheduler/job_info.c
@@ -3959,6 +3959,10 @@ create_subjob_from_array(resource_resv *array, int index, char *subjob_name)
 
 	subjob = dup_resource_resv(array, array->server, array->job->queue, err);
 
+	/* make a copy of dependent jobs */
+	subjob->job->depend_job_str = string_dup(array->job->depend_job_str);
+	subjob->job->dependent_jobs = (resource_resv **) dup_array((void **)array->job->dependent_jobs);
+
 	array->job->queued_subjobs = tmp;
 
 	if (subjob == NULL) {
--- a/src/scheduler/job_info.c
+++ b/src/scheduler/job_info.c
@@ -3961,7 +3961,7 @@ create_subjob_from_array(resource_resv *array, int index, char *subjob_name)
 
 	/* make a copy of dependent jobs */
 	subjob->job->depend_job_str = string_dup(array->job->depend_job_str);
-	subjob->job->dependent_jobs = (resource_resv **) dup_array((void **)array->job->dependent_jobs);
+	subjob->job->dependent_jobs = (resource_resv **) dup_array(array->job->dependent_jobs);
 
 	array->job->queued_subjobs = tmp;
 
--- a/test/tests/functional/pbs_job_dependency.py
+++ b/test/tests/functional/pbs_job_dependency.py
@@ -353,3 +353,182 @@ e.accept()
 
         a = {ATTR_depend: 'beforenotok:' + j1 + ":" + j2}
         self.check_job(a, reject_msg, 'F')
+
+    def check_depend_delete_msg(self, pjid, cjid):
+        """
+        helper function to check ia message that the dependent job (cjid)
+        is deleted because of the parent job (pjid)
+        """
+        msg = cjid + ";Job deleted as result of dependency on job " + pjid
+        self.server.log_match(msg)
+
+    def test_job_end_deleting_chain_of_dependency(self):
+        """
+        Submit a chain of dependent jobs and see if one of the running jobs
+        ends, all the dependent jobs (and their dependent jobs)
+        are also deleted.
+        """
+
+        a = {'job_history_enable': 'True'}
+        self.server.manager(MGR_CMD_SET, SERVER, a)
+        job = Job()
+        job.set_sleep_time(10)
+        j1 = self.server.submit(job)
+
+        a = {ATTR_depend: "afternotok:" + j1}
+        job = Job(attrs=a)
+        j2 = self.server.submit(job)
+
+        a = {ATTR_depend: "afterok:" + j2}
+        job = Job(attrs=a)
+        j3 = self.server.submit(job)
+
+        a = {ATTR_depend: "afterok:" + j1}
+        job = Job(attrs=a)
+        j4 = self.server.submit(job)
+
+        self.server.expect(JOB, {ATTR_state: 'F'}, id=j1, extend='x')
+        self.server.expect(JOB, {ATTR_state: 'F'}, id=j2, extend='x',
+                           max_attempts=3)
+        self.check_depend_delete_msg(j1, j2)
+        self.server.expect(JOB, {ATTR_state: 'F'}, id=j3, extend='x',
+                           max_attempts=3)
+        self.check_depend_delete_msg(j2, j3)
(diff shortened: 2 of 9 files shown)


test/tests/functional/pbs_job_dependency.py
```diff
@@ -353,3 +353,182 @@ def test_dependency_on_multiple_finished_job(self):
 
         a = {ATTR_depend: 'beforenotok:' + j1 + ":" + j2}
         self.check_job(a, reject_msg, 'F')
+
```
bhroam, line 356: I just noticed on line 83 of this test there is an over-indent.  If you don't mind, could you correct it?
```
            for job_list in check_dl:
                     self.assertIn(job, job_list)
```

test/tests/functional/pbs_job_dependency.py
```diff
@@ -353,3 +353,182 @@ def test_dependency_on_multiple_finished_job(self):
...
+        self.server.expect(JOB, {ATTR_state: 'R'}, id=j1)
+        self.server.delete(j1)
+        self.server.expect(JOB, {ATTR_state: 'F'}, id=j1, extend='x')
+        self.server.expect(JOB, {ATTR_state: 'F'}, id=j2, extend='x',
+                           max_attempts=3)
+        self.check_depend_delete_msg(j1, j2)
+        self.server.expect(JOB, {ATTR_state: 'F'}, id=j3, extend='x',
+                           max_attempts=3)
+        self.check_depend_delete_msg(j2, j3)
+        self.server.expect(JOB, {ATTR_state: 'R'}, id=j4)
```
bhroam, line 433: Consider adding an identical line to this between 424 and 425.  This dependency should be released when j1 runs, so it might be nice to check that it is running then.

test/tests/functional/pbs_job_dependency.py
```diff
@@ -353,3 +353,182 @@ def test_dependency_on_multiple_finished_job(self):
...
+        job = Job(attrs=a)
+        j2 = self.server.submit(job)
+
+        a = {ATTR_depend: "afternotok:" + j2}
+        job = Job(attrs=a)
+        j3 = self.server.submit(job)
+
+        a = {ATTR_depend: "after:" + j1}
+        job = Job(attrs=a)
+        j4 = self.server.submit(job)
```
bhroam, line 422: Since J1 will be running. when J4 is submitted, the hold is immediately released.  Is this OK?

test/tests/functional/pbs_job_dependency.py
```diff
@@ -353,3 +353,182 @@ def test_dependency_on_multiple_finished_job(self):
...
+        self.server.expect(JOB, {ATTR_state: 'F'}, id=j3, extend='x',
+                           max_attempts=3)
+        self.check_depend_delete_msg(j2, j3)
+        self.server.expect(JOB, {ATTR_state: 'R'}, id=j4, max_attempts=3)
+
+    def test_qdel_deleting_chain_of_dependency(self):
+        """
+        Submit a chain of dependent jobs and see if one of the running jobs
+        is deleted, all the dependent jobs (and their dependent jobs)
+        are also deleted.
```
bhroam, line 403: Say something like jobs without their dependencies released.  Since literally j4 has a dependency on j1, but it is released by the time the delete happens.

test/tests/performance/test_dependency_perf.py
```diff
@@ -0,0 +1,83 @@
...
+        PBS takes to get rid of all dependent jobs.
+        """
+
+        a = {'job_history_enable': 'True'}
+        self.server.manager(MGR_CMD_SET, SERVER, a)
+        job = Job()
+        job.set_sleep_time(3600)
+        jid = self.server.submit(job)
+        j_arr = [jid]
+        for each in range(5000):
```
bhroam, line 67: you don't use each.  Use '_' instead

test/tests/performance/test_dependency_perf.py
```diff
@@ -0,0 +1,83 @@
...
+            a = {ATTR_depend: 'afternotok:' + jid}
+            jid = self.server.submit(Job(attrs=a))
+            j_arr.append(jid)
+
+        self.server.expect(JOB, {ATTR_state: 'R'}, id=j_arr[0])
+        self.server.expect(JOB, {ATTR_state: 'H'}, id=j_arr[5000])
```
bhroam, line 73: Use expexct(JOB, {ATTR_state + '=H': 5000})

test/tests/functional/pbs_job_dependency.py
arungrover: I just wanted to cover all sort of different dependency types.

test/tests/performance/test_dependency_perf.py
arungrover: Isn't it essentially the same thing?

test/tests/functional/pbs_job_dependency.py
bhroam: I wasn't saying that dependency type was wrong, but now that you have changed it to stop and start scheduling, j4 won't immediately get its dependency released as soon as it is submitted.  It should be good now.

test/tests/performance/test_dependency_perf.py
bhroam: If all goes well, yes.  If a bug sneaks in and only some of them are held, then this won't be the same.

src/scheduler/job_info.c
```diff
@@ -3959,6 +3959,10 @@ create_subjob_from_array(resource_resv *array, int index, char *subjob_name)
 
 	subjob = dup_resource_resv(array, array->server, array->job->queue, err);
 
+	/* make a copy of dependent jobs */
+	subjob->job->depend_job_str = string_dup(array->job->depend_job_str);
+	subjob->job->dependent_jobs = (resource_resv **) dup_array(array->job->dependent_jobs);
```
vstumpf, line 3964: I didn't think subjobs could have dependencies, is this required for the arrayjob?

arungrover: Well, actually this is a bug fix for an issue in runone dependency. 
What happens is that when the scheduler decides to run a subjob, it should ideally mark all the runone dependent jobs (dependent on its parent) as can_not_run. But while creating the subjobs inside the scheduler we were not copying the list of dependent jobs from the parent because of which we were never marking the dependent jobs as not run and end up running more than one job having runone dependency on each other.

vstumpf: Ah ok, makes sense.

PR #1692, opened 2020-04-25
Fixing start time change when endtime and duration changes

tpai26:
#### Describe Bug or Feature
1) Start time of a reservation was changing even if there were running jobs in it. This was happening when both end time and duration of a reservation changes which would cause the start time to change. 
2) If the alter fails, PBS reverts back original reservation values. In this case the reservation state should also be set to the correct value. In some cases the value was set to RESV_NONE

#### Describe Your Change
1) Added a if statement to check that when end time and duration of a reservation is changed,  there should not be running jobs in it.
2) Save the current reservation state. If alter fails before reverting the reservation values, the correct value will be set.

[ralter_log.txt](https://github.com/PBSPro/pbspro/files/4560080/ralter_log.txt)

### DIFF
--- a/src/server/req_modify.c
+++ b/src/server/req_modify.c
@@ -786,6 +786,7 @@ req_modifyReservation(struct batch_request *preq)
 	resc_access_perm_save = resc_access_perm;
 	psatl = (svrattrl *)GET_NEXT(preq->rq_ind.rq_modify.rq_attr);
 	presv->ri_alter_flags = 0;
+	presv->ri_alter_state = presv->ri_wattr[RESV_ATR_state].at_val.at_long;
 
 	while (psatl) {
 		long temp = 0;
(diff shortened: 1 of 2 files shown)


src/server/req_modify.c
```diff
@@ -885,6 +888,15 @@ req_modifyReservation(struct batch_request *preq)
 
 		psatl = (svrattrl *)GET_NEXT(psatl->al_link);
 	}
+
+	presv->ri_alter_state = presv->ri_wattr[RESV_ATR_state].at_val.at_long;
```
bhroam, line 892: Instead of setting this just to reset it in the above error cases, move this to the top of the while loop.  There is only one error case you don't call revert_alter_resv_times().  You can just clear this in that reject case.

Also, I'd like to see if we can combine this check with its counterpart in the running case above.  Instead of checking it up there, just do it once down here for both cases.

tpai26: We will have to keep the check to see if the reservation is in running state that there are no jobs in it. Otherwise it will execute the start time case in the while loop and presv->ri_alter_stime changes. This causes a problem with standing reservations with already altered start time. For eg this test will fail
1) Submit a standing reservation
2) Alter start time of the reservation after it starts
3) Submit a job to it and while the job is running alter start time again, this fails
4) After the reservation occurance ends, the start time of next occurance remains what it was altered too in step 2 and not the original one

tpai26:
Added latest test logs

PR #1696, opened 2020-04-27
Fix test failure for test_alter_advance_resv_start_time_before_run

npadole20:
#### Describe Bug or Feature

The failure happened because of a timing issue. The test case altered the reservation and checked for it to be in a confirmed state. But the reservation was already in running state because there was a delay in the pbs_rstat command. This is most likely due to the machine being slow when the test case was run.

#### Describe Your Change

Increased the offset to set the start time of the reservation further away from the current time.

[TestPbsResvAlter_fix.txt](https://github.com/PBSPro/pbspro/files/4541526/TestPbsResvAlter_fix.txt)

### DIFF
--- a/test/tests/functional/pbs_ralter.py
+++ b/test/tests/functional/pbs_ralter.py
@@ -476,7 +476,7 @@ class TestPbsResvAlter(TestFunctional):
 
         All the above operations are expected to be successful.
         """
-        offset = 20
+        offset = 40
         duration = 20
         shift = 10
         rid, start, end = self.submit_and_confirm_reservation(offset, duration)
--- a/test/tests/functional/pbs_ralter.py
+++ b/test/tests/functional/pbs_ralter.py
@@ -671,7 +671,7 @@ class TestPbsResvAlter(TestFunctional):
 
         All the above operations are expected to be successful.
         """
-        offset = 60
+        offset = 30
         duration = 20
         shift = 15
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_ralter.py
```diff
@@ -476,7 +476,7 @@ def test_alter_advance_resv_start_time_before_run(self):
 
         All the above operations are expected to be successful.
         """
-        offset = 20
+        offset = 40
```
bhroam, line 479: I'd bump this to 60.  I modified a similar test last week and used 60.  Since this is the second test that needs modification, you might consider looking at other similar tests.  Look for tests that don't care about when the reservation starts but wants to alter it before it starts.

npadole20:
[TestPbsResvAlter_fix3.txt](https://github.com/PBSPro/pbspro/files/4549781/TestPbsResvAlter_fix3.txt)

test/tests/functional/pbs_ralter.py
```diff
@@ -671,7 +671,7 @@ def test_alter_standing_resv_start_time_before_run(self):
 
         All the above operations are expected to be successful.
         """
-        offset = 30
+        offset = 60
```
bhroam, line 674: 30 is probably fine here since we do actually run the reservation.

npadole20:
@bhroam I have reverted the offset for "test_alter_advance_resv_start_time_before_run". Please review the latest changes.

PR #1699, opened 2020-04-28
Test test_resv_job_soft_hard2 of TestSoftWalltime is failing while verfiying job attributes

shilpakodli:
#### Describe Bug or Feature
Test test_resv_job_soft_hard2 of TestSoftWalltime failed with Job state in Queue state while test is expecting it in Running state.
This is due to, Reservation start time and when Job is altered with softwalltime=60 was conflicting and thus Job remained in Q state.

#### Describe Your Change
Move this test case from TestSoftWalltime to TestReservations test suite as this test is more related to reservation functionality by removing softwalltime part from test case.

[test_resv_job_soft_hard2_after_fix.txt](https://github.com/PBSPro/pbspro/files/4543948/test_resv_job_soft_hard2_after_fix.txt)
[test_resv_job_soft_hard2_before_fix.txt](https://github.com/PBSPro/pbspro/files/4543949/test_resv_job_soft_hard2_before_fix.txt)

### DIFF
--- a/test/tests/functional/pbs_reservations.py
+++ b/test/tests/functional/pbs_reservations.py
@@ -2254,11 +2254,8 @@ class TestReservations(TestFunctional):
         self.server.expect(RESV,
                            {'reserve_state': (MATCH_RE, 'RESV_CONFIRMED|2')},
                            id=rid)
-
-        self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'False'})
         a = {'Resource_List.ncpus': 4,
-             'Resource_List.walltime': 60}
+             'Resource_List.walltime': 50}
         J = Job(TEST_USER, attrs=a)
         jid = self.server.submit(J)
-        self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'True'})
         self.server.expect(JOB, {'job_state': 'R'}, id=jid)
--- a/test/tests/functional/pbs_reservations.py
+++ b/test/tests/functional/pbs_reservations.py
@@ -2235,3 +2235,30 @@ class TestReservations(TestFunctional):
         jid = self.server.submit(J)
 
         self.server.expect(JOB, {ATTR_state: 'R'}, id=jid)
+
+    @skipOnCpuSet
+    def test_resv_job_hard_walltime(self):
+        """
+        Test that a job with hard walltime will not conflict with
+        reservtion if hard walltime is less that reservation start time.
+        """
+        a = {'resources_available.ncpus': 4}
+        self.server.manager(MGR_CMD_SET, NODE, a, id=self.mom.shortname)
+
+        now = int(time.time())
+
+        a = {'Resource_List.ncpus': 4, 'reserve_start': now + 65,
+             'reserve_end': now + 240}
+        R = Reservation(TEST_USER, attrs=a)
+        rid = self.server.submit(R)
+        self.server.expect(RESV,
+                           {'reserve_state': (MATCH_RE, 'RESV_CONFIRMED|2')},
+                           id=rid)
+
+        self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'False'})
+        a = {'Resource_List.ncpus': 4,
+             'Resource_List.walltime': 60}
+        J = Job(TEST_USER, attrs=a)
+        jid = self.server.submit(J)
+        self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'True'})
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid)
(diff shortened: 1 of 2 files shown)


shilpakodli:
@agrawalravi90 The intent of test is to run Job with hard and softwalltime which do not overlap with Reservation start time. But when we are qaltering a job with softwalltime=60, on slow machines, this time is conflicting with Reservation start time.

agrawalravi90:
The test header:
"Test that a job with soft and hard walltime will not conflict with
 reservtion if hard walltime is less that reservation start time"

made me think that only the hard walltime is less than reservation start time and the soft walltime overlaps. Otherwise what is the point of the test really? It will submit a job with soft and hard walltimes less than a reservation's start time .. what functionality is that testing?

bhroam:
@agrawalravi90 you can't have a soft walltime > hard walltime.  That would not make sense knowing the job will be killed at the end of the hard walltime.  You are right that this test doesn't test too much, but it does have some value.  It is testing that a job can be backfilled before a reservation if it does not conflict with it. In any case, when checking for backfilling with reservations, the job's hard walltime is always used.

agrawalravi90:
Thanks for clarifying @bhroam , that make sense. I'm confused about the role of soft walltime in this test then, it doesn't seem to be playing any role. I think we should just get rid of this test.

shilpakodli:
> Thanks for clarifying @bhroam , that makes sense. I'm confused about the role of soft walltime in this test then, it doesn't seem to be playing any role. I think we should just get rid of this test.

Hi @agrawalravi90 Are you saying that we should remove this test ?

bhroam:
@agrawalravi90 This test is testing something.  It might be slightly misplaced though.  It is testing that if a reservation is in the future, you can backfill a job before it.  The soft_walltime aspect of this test doesn't do much.  Maybe remove the alter of soft walltime and move this test into the reservation test suite?  Or is there a test in the reservation test suite that is already testing this?

agrawalravi90:
@bhroam  I assumed that we'd have a test for this already, but I actually can't find one in the reservation test suite. So, yes, in that case we should remove the soft walltime part of it and move it to the reservation test suite with a different name

shilpakodli:
Addressed review comments. Please find attached new execution results:
[TestReservations_logs.txt](https://github.com/PBSPro/pbspro/files/4644493/TestReservations_logs.txt)

test/tests/functional/pbs_reservations.py
```diff
@@ -2235,3 +2235,30 @@ def test_server_recover_resv_queue(self):
...
+             'reserve_end': now + 240}
+        R = Reservation(TEST_USER, attrs=a)
+        rid = self.server.submit(R)
+        self.server.expect(RESV,
+                           {'reserve_state': (MATCH_RE, 'RESV_CONFIRMED|2')},
+                           id=rid)
+
+        self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'False'})
+        a = {'Resource_List.ncpus': 4,
+             'Resource_List.walltime': 60}
```
bhroam, line 2260: I'd decrease this.  Right now there is a race condition that if it takes more than 5 seconds to submit the reservation, do the expect(), and then submit the job, the job won't run.  We have some super-slow machines in the testbed where it can take more than 5s to do a pbs_rstat.  I'd decrease the walltime to 50s.

test/tests/functional/pbs_reservations.py
```diff
@@ -2235,3 +2235,30 @@ def test_server_recover_resv_queue(self):
...
+
+        a = {'Resource_List.ncpus': 4, 'reserve_start': now + 65,
+             'reserve_end': now + 240}
+        R = Reservation(TEST_USER, attrs=a)
+        rid = self.server.submit(R)
+        self.server.expect(RESV,
+                           {'reserve_state': (MATCH_RE, 'RESV_CONFIRMED|2')},
+                           id=rid)
+
+        self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'False'})
```
bhroam, line 2258: No real need to stop/start scheduling.  A cycle will be run when you submit the job.  Since you aren't trying to coordinate multiple jobs into one cycle, this is not really necessary.

shilpakodli:
Addressed review comments from Bhroam.
[TestReservations_logs_2.txt](https://github.com/PBSPro/pbspro/files/4650314/TestReservations_logs_2.txt)

agrawalravi90:
@shilpakodli can you please change the description of this PR to reflect the new changes?

shilpakodli:
@agrawalravi90 I have changed PR description. Please take this PR forward.

prakashcv13:
merged based on reviews from @bhroam and @agrawalravi90.
The outstanding comment related to the description is addressed. If the description needs further improvement, we can still do it.

PR #1700, opened 2020-04-28
Change cgroups top dirs for systemd Delegate=yes

alexis-cousein:
#### Describe Bug or Feature

Many versions of systemd are continuing to interfere with the cgroups set up by the hook; the latest incident was out devices list clashing with the default DevicePolicy, which made systemd rewrite the devices.allow to a "clean" ```a:*.* rwm```, making the cgroup hook lose view of GPU assignments to existing jobs.

The reason is that Delegate=yes is NOT supported on slices (for very nebulous reasons). In systemd v238, this change willl actually be ENFORCED and the systemctl commands to create slices are going to fail.

That is what actually caused us to constantly adhere to systemd's view of the world in our _entire_ cgroup tree -- without realising it, we were constantly forced to trick systemd into thinking the cgroups had been set up by it, by mimicking it as well as we could, since we did not TRULY get the delegate control we thought we had.

Also, in cgroups v2 this kind of control (with two writers, one of whichis systemd via systemctl to create slices, and the other the cgroup hook to manage them) is likely also not going to work anymore.

To fix that necessitates a fairly simply change in the directory hierarchies under which the per-job directories are made.

#### Describe Your Change

Delegate=yes is only supported on scopes and services. Given the fact that scopes do not support simple unit files, we create a pbs_job.services container for our jobs using a dummy service (one that simply starts a sleep infinity to keep systemd from cleaning up the cgroups if there are no jobs).

The cgroup files directly under this directory are NOT under our control, but that of systemd, but any sub-cgroups/subdirectories *are* under our exclusive control. So we create a "jobid" directory there --no need to conform to any systemd naming scheme. That will both house zero-CPU job processes and any per-job cgroups created, and the Delegate=yes on the parent means it's our private playground, with no need to adhere to any systemd naming convention.

#### Link to Design Doc

Does not affect documented interfaces, but changes paths for cgroups, so it cannot be installed "on the fly" with jobs on the node.

N/A. We cannot _trigger_ systemd to interfere with us, but it does. What should be tested is conformity with the current test suite, though.

image


test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -2985,10 +3040,11 @@ def test_cgroup_sismom_resize_fail(self):
             jid, execvnode1), starttime=stime)
         self.moms_list[1].log_match("Job;%s;pruned to exec_vnode=%s" % (
             jid, execvnode2), starttime=stime)
-        # Check that the sister mom failed to update the job
+        # Check that momsup saw that the sister mom failed to update the job
```
mike0042, line 3043: momsup?

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -2985,10 +3040,11 @@ def test_cgroup_sismom_resize_fail(self):
...
+        # This message is on momsup mom[1] but mentions sismom mom[0]
```
mike0042, line 3044: Took me a second to realize momsup meant mother superior. We usually abbreviate it as MS.

alexis-cousein:
Rebased on upstream/master

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -3423,6 +3825,47 @@ def test_cpu_controller_enforce_default_zero_job(self):
               zero_cpus_shares_fraction
               zero_cpus_quota_fraction
         """
+        root_quota_host1 = None
+        try:
+            root_quota_host1_str = \
+                self.du.run_cmd(hosts=self.hosts_list[0],
+                                cmd=['cat',
+                                     '/sys/fs/cgroup/cpu/cpu.cfs_quota_us'])
```
bayucan, line 3833: These look like repeating blocks of statements that are best to be put in a class variable.

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -1777,9 +1823,8 @@ def test_cgroup_cpuset_ncpus_are_cores(self):
                     cores = int(cores_match.groups()[0])
                 if phys_match:
                     pval = int(phys_match.groups()[0])
-                    if pval != prev:
-                        prev = pval
-                        phys += 1
```
bayucan, line 1782: This has already been fixed in https://github.com/PBSPro/pbspro/pull/1709, so best to just rebase master and pick that up.

alexis-cousein: That test has other problems -- so I'm sticking to my guns for the moment:

-on TH3 the timing introduces a spurious failure -- 10 attempts are not enough to make all jobs that run actually run; the fact the daemons and hooks are CPU starved and that the exechost_periodic interlopes makes it unsafe. Better sleep a short time between jobs...and increase the number of attempts.
-The job runs are so slow that a default sleep of 100 is too dangerous on large machines -- by the time you get to the last job the first ones will have finished.
-We're submitting far too many jobs for what the test tries to test. As a result, on large machines we'll just skip (with a "heuristic" limit of 64 jobs).
-The test at the end is also not safe. It "expects" a queued job, but since it looks just after the job is submitted the scheduler might not yet have gotten to it, which would produce false positives. If the job starts Q and then becomes R, "expect" will not do 10 attempts and then fail when the job becomes "R", it will just declare victory at the first Q.

So I'm making a version that will first submit one job per phys, of "cores" width. That should make it practical even on large _and_ slow machines. I'm setting the sleep to 3600 to avoid spurious "success" by jobs finishing early, and I'm _first_ sleeping for 30 seconds before doing an expect of state Q for the job that should not run (which I keep as a 1 CPU job).

alexis-cousein: In other words: I'll rebase, but I still plan to keep my test in the merge ;-).

test/tests/functional/pbs_cgroups_hook.py
alexis-cousein: Yes, but frankly, I am now trying to focus on making things work, not refactoring it.

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -3178,13 +3262,18 @@ def test_big_cgroup_cpuset(self):
...
         Test to make sure that when scheduler preempts a multi-node job with
         checkpoint_abort, execjob_abort cgroups hook on secondary node
         gets called.  The abort hook cleans up assigned cgroups, allowing
         the higher priority job to run on the same node.
         """
+
+        self.skipTest('Test replaced with alternative '
+                      'to avoid scheduling race')
+        return
+
```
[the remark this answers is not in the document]
alexis-cousein: That is because the test was useful to find another bug so I wanted to keep it around -- but Arun has found that one, so I'll remove it.

test/tests/functional/pbs_cgroups_hook.py
alexis-cousein: Done

test/tests/functional/pbs_cgroups_hook.py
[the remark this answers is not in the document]
alexis-cousein: Done.

test/tests/functional/pbs_cgroups_hook.py
alexis-cousein: Rebased. Did keep a set instead of a dictionary with separate keys and constant all-one values, since they are functionally equivalent but the set is more readable and natural.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -14040,6 +14040,9 @@ def enable_cgroup_cset(self):
                  'input-file': path}
             self.server.manager(MGR_CMD_IMPORT, HOOK, a,
                                 'pbs_cgroups')
+            self.log_match('pbs_cgroups.CF;copy hook-related '
+                           'file request received',
+                           starttime=self.server.ctime)
```
bayucan, line 14045: I've tried this on our local UV-like system and this is failing:

020-05-12 09:33:37,164 INFOCLI  uv-02-s12: /opt/pbs/bin/pbsnodes -s uv-02-s12.pbspro.com -v uv-02-s12
2020-05-12 09:33:37,245 ERROR    err: ['/opt/pbs/bin/pbsnodes: Server has no node list']
2020-05-12 09:33:37,246 INFOCLI2 uv-02-s12: sudo -H /opt/pbs/sbin/pbs_mom --version
2020-05-12 09:33:37,288 INFOCLI2 uv-02-s12: sudo -H /opt/tools/wrappers/rm -f /var/spool/pbs/mom_priv/epilogue
2020-05-12 09:33:37,306 INFOCLI2 uv-02-s12: sudo -H /opt/tools/wrappers/rm -f /var/spool/pbs/mom_priv/prologue
2020-05-12 09:33:37,322 INFOCLI  uv-02-s12: sudo -H /opt/pbs/sbin/pbs_mom -s list
2020-05-12 09:33:37,385 INFOCLI2 uv-02-s12: sudo -H /opt/tools/wrappers/cp /tmp/PtlPbsg6srmjew /var/spool/pbs/mom_priv/config
2020-05-12 09:33:37,404 INFOCLI2 uv-02-s12: /opt/tools/wrappers/cat /proc/mounts
2020-05-12 09:33:37,415 INFO     manager on uv-02-s12: import hook pbs_cgroups {'content-type': 'application/x-config', 'content-encoding': 'default', 'input-file': '/tmp/cfgztxy6wet.json'}
2020-05-12 09:33:37,416 INFOCLI  uv-02-s12: sudo -H /opt/pbs/bin/qmgr -c import hook pbs_cgroups application/x-config default /tmp/cfgztxy6wet.json
2020-05-12 09:33:37,506 ERROR    err: ["hook 'pbs_cgroups' contents overwritten"]
2020-05-12 09:33:37,517 INFO     mom uv-02-s12 log match: searching for "pbs_cgroups.CF;copy hook-related file request received" - with existence - No match
2020-05-12 09:33:38,043 INFO     mom uv-02-s12 log match: searching for "pbs_cgroups.CF;copy hook-related file request received" - with existence - attempt 2
2020-05-12 09:33:38,566 INFO     mom uv-02-s12 log match: searching for "pbs_cgroups.CF;copy hook-related file request received" - with existence - attempt 3
2020-05-12 09:33:39,093 INFO     mom uv-02-s12 log match: searching for "pbs_cgroups.CF;copy hook-related file request received" - with existence - attempt 4
2020-05-12 09:33:39,619 INFO     mom uv-02-s12 log match: searching for "pbs_cgroups.CF;copy hook-related file request received" - with existence - attempt 5
..
2020-05-12 09:34:07,952 INFO     mom uv-02-s12 log match: searching for "pbs_cgroups.CF;copy hook-related file request received" - with existence - attempt 59
2020-05-12 09:34:08,478 INFO     mom uv-02-s12 log match: searching for "pbs_cgroups.CF;copy hook-related file request received" - with existence - attempt 60
2020-05-12 09:34:08,992 INFO     FAILED

----------
Problem is this:
qmgr -c "import hook application/x-config default pbs_cgroups.json"   <-- problem here since there's no local node, server would not communicate the pbs_cgroups.CF file yet to the mom.
restart pbs_mom
check for "pbs_cgroups.CF; copy request"   but since there's no local node yet, then we would never get this message.
---
I'm working on an adjustment to your fix, where after the delete node @default, need to do create node <local_node>, and then do the enable_cgroups_cpuset(), and finally restarting mom.

alexis-cousein: Yes. Fixed since.

Since the race seemed to be that when you create the node (when the hook is disabled) and then immediately enable the node, you have two conflicting HK files being pushed to the node after the server hello, I changed to reconfigure the hook while the nodes are deleted. But then you need to skip those tests for the hook file copies -- they will occur when the node is created later.

You DO NOT want to create the nodes. There was one, and it was removed ON PURPOSE, because that was what created a race condition between the "create node" and the "qmgr set hook X enabled = true".

But you do want to skip the hook copies that will not occur if the nodes are not there, of course.

alexis-cousein: Ah -- I see I forgot about the one in that function, yes.

alexis-cousein: Let me fix it.

alexis-cousein: I think we should simply be guarding this after having done a status() on the node. If it's there, then look for the hook copy, if not, then skipt it -- the hook copy is then done when the node is created.

alexis-cousein: Something like:
            try:
                statm = self.status(NODE, id=self.shortname)
            except:
                statm = []
            if len(statm) >= 1:

alexis-cousein: Please look at the latest push.

alexis-cousein:
Had to rebase because I fixed a bug in log.h in master (one that causes build failures on some OS environments).

Please see if my latest push addresses your comment.

alexis-cousein:
Reverted to standard pbs_testlib.py, fixed sequence in pbs_testuite.py where we are less likely to break things for others.

test/fw/ptl/lib/pbs_testlib.py
alexis-cousein: Latest push resolved our differences :-)

PR #1701, opened 2020-04-29
Add skipOnCpuSet decorator SmokeTest tests

shilpakodli:
#### Describe Bug or Feature
Tests test_submit_job_with_script,test_exclhost of SmokTest are failing due to ncpus being set on Mom node and also creating vnodes in test respectively.

#### Describe Your Change
Added skipOnCpuSet decorator on test_exclhost test and modified test_submit_job_with_script test to run on cpuset machine.

### DIFF
--- a/test/tests/pbs_smoketest.py
+++ b/test/tests/pbs_smoketest.py
@@ -563,7 +563,7 @@ class SmokeTest(PBSTestSuite):
         Test to submit job with job script
         """
         j = Job(TEST_USER, attrs={ATTR_N: 'test'})
-        j.create_script('sleep 5\n', hostname=self.server.client)
+        j.create_script('sleep 120\n', hostname=self.server.client)
         jid = self.server.submit(j)
         self.server.expect(JOB, {'job_state': 'R'}, id=jid)
         self.server.delete(id=jid, extend='force', wait=True)
--- a/test/tests/pbs_smoketest.py
+++ b/test/tests/pbs_smoketest.py
@@ -558,17 +558,15 @@ class SmokeTest(PBSTestSuite):
                 (GE, a['Resource_List.min_walltime'])}
         self.server.expect(JOB, attr, id=jid)
 
-    @skipOnCpuSet
     def test_submit_job_with_script(self):
         """
         Test to submit job with job script
         """
-        a = {ATTR_rescavail + '.ncpus': '2'}
-        self.server.manager(MGR_CMD_SET, NODE, a, id=self.mom.shortname)
         j = Job(TEST_USER, attrs={ATTR_N: 'test'})
-        j.create_script('sleep 120\n', hostname=self.server.client)
+        j.create_script('sleep 5\n', hostname=self.server.client)
         jid = self.server.submit(j)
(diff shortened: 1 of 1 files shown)


test/tests/pbs_smoketest.py
```diff
@@ -558,6 +558,7 @@ def test_shrink_to_fit(self):
                 (GE, a['Resource_List.min_walltime'])}
         self.server.expect(JOB, attr, id=jid)
 
+    @skipOnCpuSet
```
hirenvadalia, line 561: I would suggest changing test case not to set ncpus instead skipping it because the test is about a testing job with job script and it has nothing to do about how many ncpus we have (as long as we can run simple 1-ncpus job)

shilpakodli:
@hirenvadalia Thanks for the review comments. I have changed test accordingly. Attaching execution logs.
[test_results_after_fix.txt](https://github.com/PBSPro/pbspro/files/4550467/test_results_after_fix.txt)

test/tests/pbs_smoketest.py
```diff
@@ -562,12 +562,11 @@ def test_submit_job_with_script(self):
         """
         Test to submit job with job script
         """
-        a = {ATTR_rescavail + '.ncpus': '2'}
-        self.server.manager(MGR_CMD_SET, NODE, a, id=self.mom.shortname)
         j = Job(TEST_USER, attrs={ATTR_N: 'test'})
-        j.create_script('sleep 120\n', hostname=self.server.client)
+        j.create_script('sleep 5\n', hostname=self.server.client)
```
bhroam, line 566: I'd leave the job alone and then just delete it after you do what you want instead of checking for it to end.  On slow machines you might miss the job running in 5 seconds.

test/tests/pbs_smoketest.py
```diff
@@ -562,12 +562,11 @@ def test_submit_job_with_script(self):
...
         jid = self.server.submit(j)
         self.server.expect(JOB, {'job_state': 'R'}, id=jid)
+        self.server.expect(JOB, 'queue', op=UNSET, id=jid, offset=5)
         self.logger.info("Testing script with extension")
         j = Job(TEST_USER)
         fn = self.du.create_temp_file(suffix=".scr", body="/bin/sleep 10",
```
bhroam, line 572: This isn't part of your change, but I don't understand the next part of this test.  We don't submit the job with -N, so why is it OK for the job submit to fail as long as we don't fail with a -N failure message?  Shouldn't it be a test failure if the job fails to submit?  If you want to leave it alone, that is OK.

shilpakodli: I Agree with you @bhroam . It doesn't make any sense to catch qsub error and checking error message here. Test should fail if qsub fails. I have removed try-except block. Please have a look.

test/tests/pbs_smoketest.py
bhroam: What I meant by leaving the job alone is to revert it to sleep 120 seconds.  5s is still too short of a sleep.

PR #1703, opened 2020-04-29
Fix: NULL environment variable names cause a crash

raviv1995:
#### Describe Bug or Feature
qsub -S causes a mom crash when an incorrect path is provided.

#### Describe Your Change
This crash is caused when trying to build an environment variable with NULL as it's name. While adding variables to the environment there is no check for whether or not it is NULL.

#### Link to Design Doc
No design changes

### DIFF
--- a/src/resmom_win/start_exec.c
+++ b/src/resmom_win/start_exec.c
@@ -850,7 +850,7 @@ bld_wenv_variables(char *name, char *value)
 	int	i;
 	char    str_buf[MAXPATHLEN+1] = {0};
 
-	if ((*name == '\0') || (*name == '\n'))
+	if ((name == NULL) || (name == '\0') || (name == '\n'))
 		return;			/* invalid name */
 
 	if (env_array == NULL) {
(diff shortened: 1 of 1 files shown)


src/resmom_win/start_exec.c
```diff
@@ -850,7 +850,7 @@ bld_wenv_variables(char *name, char *value)
 	int	i;
 	char    str_buf[MAXPATHLEN+1] = {0};
 
-	if ((*name == '\0') || (*name == '\n'))
+	if ((name == NULL) || (name == '\0') || (name == '\n'))
```
subhasisb, line 853: ooooh - big mistake you missed the (star) *name in the second and third comparison rendering those checks useless

subhasisb: The previous checks are fine, all you need to do is add a if (!name) || ... to it - rest is already good

raviv1995:
Thanks for the review @subhasisb @Bhagat-Rajput

PR #1704, opened 2020-04-29
test "pbs_hook_send" checks “Timing out” messages are received when one of the mom is unresponsive

birbalsain:
#### Describe Bug or Feature
test is checking "(send replies  expect and receive" along with timeout message when one of the mom is unresponsive. Number of expect and received are not constant, numbers vary on each test-run
It is important to check timeout messages are received at least 3 times.

#### Describe Your Change
Test checks “Timing out” messages are received at least 3 times when one of the mom is unresponsive

[timeout_message.txt](https://github.com/PBSPro/pbspro/files/4581933/timeout_message.txt)

### DIFF
--- a/test/tests/functional/pbs_hook_timeout.py
+++ b/test/tests/functional/pbs_hook_timeout.py
@@ -74,6 +74,19 @@ class TestHookTimeout(TestFunctional):
         for mom in self.moms.values():
             self.server.expect(NODE, {'state': 'free'}, id=mom.shortname)
 
+    def timeout_messages(self, num_msg=1, starttime=None):
+        msg_found = None
+        for count in range(10):
+            sleep(30)
+            allmatch_msg = self.server.log_match(
+                "Timing out previous send of mom hook updates ",
+                max_attempts=1, starttime=starttime, allmatch=True)
+            if len(allmatch_msg) >= num_msg:
+                msg_found = allmatch_msg
+                break
+        self.assertIsNotNone(msg_found,
+                             msg="Didn't get expected timeout messages")
+
     def test_hook_send(self):
         """
         Test when the server doesn't receive an ACK from a mom for
@@ -81,7 +94,6 @@ class TestHookTimeout(TestFunctional):
         """
         self.server.manager(MGR_CMD_SET, SERVER, {'log_events': 2047})
         timeout_max_attempt = 7
-        timeout_msg = False
 
         # Make momB unresponsive
         self.logger.info("Stopping MomB")
@@ -117,18 +129,7 @@ class TestHookTimeout(TestFunctional):
 
         # Second batch of hook update is for the *.PY files + resend of
         # *.HK file to momB
-        for count in range(10):
-            sleep(30)
-            allmatch_msg = self.server.log_match(
-                "Timing out previous send of mom hook updates ",
-                n=600, starttime=start_time, allmatch=True)
-            if len(allmatch_msg) >= 2:
-                timeout_msg = True
-                break
-        if not timeout_msg:
-            self.logger.error("Doesn't gets expected timeout messages")
-        timeout_msg = False
-
+        self.timeout_messages(2, start_time)
         # sent hook content file
         for h in [self.hostA, self.hostB, self.hostC]:
             hfile = os.path.join(self.server.pbs_conf['PBS_HOME'],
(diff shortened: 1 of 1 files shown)


birbalsain:
> Since we're now looking at the "Timing out" message in its generic form, we want to make sure at least 3 of these are seen in the logs, one when trying to send *.HK, *.PY, and and a resend of *.HK and *.PY to the unresponsive mom. One or two ways this can be done: refresh the start_time value after receiving each "Timing out" message, or just do a match for "Timing out" message at the end, and count it to make sure we see the log message at least 3 times.

I am getting 2 timeout message. one when trying to send and second for the resend. I am not getting third one.

test/tests/functional/pbs_hook_timeout.py
```diff
@@ -155,11 +155,11 @@ def test_hook_send(self):
 
         # Ensure that hook send updates are retried for
         # the *.HK and *.PY file to momB
```
birbalsain, line 157: @bayucan here I am not getting timeout message.

birbalsain, line 157: @bayucan Here i am not getting any timeout message because sometime i get exact message with exact expected and received number but I didn't this message. with or without exact match, I am getting only 2 timeout message.

bayucan: @birbalsain : There's a time gap between timeout sends messages, and perhaps the log_match was not retried enough. I think currently, it tries only 7 times.
Also, from the original email you sent me reporting the problem, the output had 4:

"next_sync_mom_hookfiles;Timing out previous send of mom hook updates (send replies expected=28 received=23)
next_sync_mom_hookfiles;Timing out previous send of mom hook updates (send replies expected=11 received=4)
next_sync_mom_hookfiles;Timing out previous send of mom hook updates (send replies expected=7 received=0)
next_sync_mom_hookfiles;Timing out previous send of mom hook updates (send replies expected=7 received=0)

--
Also, no need to look for 3 instances of "Timeout messages" at each step. I suggest either:
1) Check one instance of the log message at each step, with the start_time value adjusted at each step.
- or -
2) Just have one log_match() instance towards the end of the test, and counting to make sure 3 timeout messages are seen.

birbalsain: @bayucan at each step, I am checking number of times timeout messages are received.

birbalsain:

test/tests/functional/pbs_hook_timeout.py
```diff
@@ -155,12 +162,16 @@ def test_hook_send(self):
...
-        self.server.log_match(
-            "Timing out previous send of mom hook updates "
-            "(send replies expected=2 received=0)", n=600,
-            max_attempts=timeout_max_attempt, interval=30,
-            starttime=start_time)
-
+        for count in range(10):
```
agrawalravi90, line 165: code duplication: put this in a common method and invoke it from the 2 places instead. You can pass a "num_instances" arg which will be set to '2' for the first call and '3' for the second.

test/tests/functional/pbs_hook_timeout.py
```diff
@@ -116,11 +117,17 @@ def test_hook_send(self):
 
         # Second batch of hook update is for the *.PY files + resend of
         # *.HK file to momB
-        self.server.log_match(
-            "Timing out previous send of mom hook updates "
-            "(send replies expected=4 received=2)", n=600,
-            max_attempts=timeout_max_attempt, interval=30,
-            starttime=start_time)
+        for count in range(10):
```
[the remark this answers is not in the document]
birbalsain: here definitely we will get message because here we are checking for second time. we have been already checked this message in above part.

test/tests/functional/pbs_hook_timeout.py
```diff
@@ -116,11 +117,17 @@ def test_hook_send(self):
...
+            sleep(30)
+            allmatch_msg = self.server.log_match(
+                "Timing out previous send of mom hook updates ",
+                n=600, starttime=start_time, allmatch=True)
+            if len(allmatch_msg) >= 2:
+                timeout_msg = True
+                break
+        if not timeout_msg:
+            self.logger.error("Doesn't gets expected timeout messages")
```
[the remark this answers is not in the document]
birbalsain: I agree with you it should fail.

test/tests/functional/pbs_hook_timeout.py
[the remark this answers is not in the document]
agrawalravi90: Ah, I see, I didn't realize that start_time was before the previous log_match(), ok, this will work then.

birbalsain:
@agrawalravi90 I have done the requested changes.

PR #1705, opened 2020-04-29
multinode cgroup test fixes

mike0042:
#### Describe Bug or Feature

Some multinode cgroups tests were regularly failing per QA

#### Describe Your Change

Adjust tests in question to use three moms rather than a multi-vnode mom. Also cleaned up checks for mom requirements as decorators are in place now.

#### Link to Design Doc

N/A

[cgroup-multinode.log](https://github.com/PBSPro/pbspro/files/4554566/cgroup-multinode.log)

### DIFF
--- a/test/tests/functional/pbs_cgroups_hook.py
+++ b/test/tests/functional/pbs_cgroups_hook.py
@@ -2923,6 +2923,7 @@ event.accept()
         exec_job_resize failure in mom superior,
         tolerate_node_failures=job_start
         """
+        self.load_config(self.cfg7)
         # instantiate queuejob hook
         hook_event = 'queuejob'
         hook_name = 'qjob'
@@ -2983,6 +2984,7 @@ event.accept()
         job is using only vnodes from mother superior host,
         tolerate_node_failures=job_start
         """
+        self.load_config(self.cfg7)
         # disable queuejob hook
         hook_event = 'queuejob'
         hook_name = 'qjob'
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -2988,29 +2911,20 @@ def test_cgroup_sismom_resize_fail(self):
...
         Verify that exec_vnode values are trimmed
         when execjob_launch hook prunes job via release_nodes(),
         exec_job_resize failure in mom superior,
         tolerate_node_failures=job_start
         """
-        # Test requires 2 nodes
-        if len(self.moms) < 2:
-            self.skipTest('Test requires at least two Moms as input, '
-                          'use -p moms=<mom1:mom2>')
-        self.load_config(self.cfg7)
```
bayucan, line 3009: I don't think you mean to remove also the "self.load_config(..)" call.

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -3063,23 +2977,14 @@ def test_cgroup_msmom_resize_fail(self):
...
         Verify that exec_vnode values are trimmed
         when execjob_launch hook prunes job via release_nodes(),
         job is using only vnodes from mother superior host,
         tolerate_node_failures=job_start
         """
-        # Test requires 2 nodes
-        if len(self.moms) < 2:
-            self.skipTest('Test requires at least two Moms as input, '
-                          'use -p moms=<mom1:mom2>')
-        self.load_config(self.cfg7)
```
bayucan, line 3078: Same here too, I don't you mean to remove the load_config() call.

test/tests/functional/pbs_cgroups_hook.py
mike0042: Good catch. Fixed.

test/tests/functional/pbs_cgroups_hook.py
mike0042: Fixed as well.

PR #1707, opened 2020-05-02
Server does not reply to ralter cmd when scheduler fails to alter resv

arungrover:
#### Describe Bug or Feature
when ralter is issue with "-I" option, server fails to reply to the command when the scheduler fails to confirm the reservation.

#### Describe Your Change
This issue arose when we changed PBS to support reservations in multi-sched environment. The problem is that server keeps a count of request/replies it sent/received from scheduler. Now, the server used these req/reply counts to ensure that it has received all replies from all schedulers before reply to the command that it failed to confirm a reservation.

Now when the user tries to ralter a reservation, the server used to increment the reply count only when the reservation was in an unconfirmed state. This leads to a problem because, during ralter, state of the reservation is set to "being_altered".

The fix is to use req/reply count only when the reservation's partition is not set. It is only in this case that the notification to confirm a reservation is sent to all schedulers.  Once a partition is set on a reservation, it is never unset and the server only has 1:1 communication with the relevant scheduler about this reservation.

in this fix notify_scheds_about_resv() function resets req/reply count to 0 when it sees the partition is not set on the reservation. It also increments request count only when partition is not set.
In req_confirmresv we increment reply count in all the cases. This is because it is hard to tell if replies are coming because of the initial broadcast made to all schedulers of because of a notification sent to a particular scheduler.

Along with this bug fix, I've fixed a place where we were unsetting partition when a standard reservation moved to a degraded state. This was wrong because once a reservation is confirmed it sticks to the partition it was confirmed in.

#### Link to Design Doc
NA

I've not written any extra test, there wasn't an easy way to efficiently capture the ralter hang in a test.
Below are logs from multi-sched test suite.
[test.txt](https://github.com/PBSPro/pbspro/files/4567038/test.txt)

### DIFF
--- a/test/tests/functional/pbs_multi_sched.py
+++ b/test/tests/functional/pbs_multi_sched.py
@@ -214,7 +214,7 @@ class TestMultipleSchedulers(TestFunctional):
         pbs_home = self.server.pbs_conf['PBS_HOME']
         self.du.run_copy(self.server.hostname,
                          src=os.path.join(pbs_home, 'sched_priv'),
-                         dest=s.path.join(pbs_home, 'sc1_new_priv'),
+                         dest=os.path.join(pbs_home, 'sc1_new_priv'),
                          recursive=True)
         self.server.manager(MGR_CMD_SET, SCHED,
                             {'sched_priv': '/var/spool/pbs/sc1_new_priv'},
--- a/src/server/req_modify.c
+++ b/src/server/req_modify.c
@@ -965,11 +965,8 @@ req_modifyReservation(struct batch_request *preq)
 		}
 	}
 
-	if (send_to_scheduler) {
-		presv->rep_sched_count = 0;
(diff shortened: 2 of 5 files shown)


test/tests/functional/pbs_multi_sched.py
```diff
@@ -214,7 +214,7 @@ def test_set_sched_priv(self):
         pbs_home = self.server.pbs_conf['PBS_HOME']
         self.du.run_copy(self.server.hostname,
                          src=os.path.join(pbs_home, 'sched_priv'),
-                         dest=s.path.join(pbs_home, 'sc1_new_priv'),
+                         dest=os.path.join(pbs_home, 'sc1_new_priv'),
```
vstumpf, line 217: Thanks for this, I ran into this before and didn't fix it.

PR #1708, opened 2020-05-02
Jobs don't accrue eligible time by default

agrawalravi90:
#### Describe Bug or Feature

According to documented behavior, the default accrue_type for jobs should be eligible time, but the code uses initial time instead. This also has a performance implication: usually, the chances of of jobs accruing eligible time are more than ineligible time. Not setting the default to eligible means that the scheduler has to send attribute updates to server for such jobs to update their accrue_type to eligible, which we would save if we corrected the default to eligible time.

#### Describe Your Change

Made the default accrue_type eligible time and added a couple of tests.

[testeligible.log](https://github.com/PBSPro/pbspro/files/4567163/testeligible.log)

### DIFF
--- a/src/server/hook_func.c
+++ b/src/server/hook_func.c
@@ -3565,8 +3565,8 @@ do_runjob_reject_actions(job *pjob, char *hook_name)
 	}
 
 	/* update the eligible time to JOB_INITIAL */
-	if (server.sv_attr[(int)SRV_ATR_EligibleTimeEnable].at_val.at_long == 1 &&
-		server.sv_attr[(int)SRV_ATR_EligibleTimeEnable].at_flags & ATR_VFLAG_SET)
+	if ((server.sv_attr[SRV_ATR_EligibleTimeEnable].at_flags & ATR_VFLAG_SET)
+		&& server.sv_attr[SRV_ATR_EligibleTimeEnable].at_val.at_long == 1)
 		update_eligible_time(JOB_INITIAL, pjob);
 
 	/*
--- a/src/server/hook_func.c
+++ b/src/server/hook_func.c
@@ -3564,6 +3564,10 @@ do_runjob_reject_actions(job *pjob, char *hook_name)
 		}
 	}
 
+	/* update the eligible time to JOB_INITIAL */
+	if (server.sv_attr[(int)SRV_ATR_EligibleTimeEnable].at_val.at_long == 1)
+		update_eligible_time(JOB_INITIAL, pjob);
+
 	/*
 	 * Don't let the values in runjob_reject_attrlist (a_map) linger.
 	 * When resources get deleted and recreated between job runs, the
--- a/src/server/hook_func.c
+++ b/src/server/hook_func.c
@@ -3565,7 +3565,8 @@ do_runjob_reject_actions(job *pjob, char *hook_name)
 	}
 
 	/* update the eligible time to JOB_INITIAL */
-	if (server.sv_attr[(int)SRV_ATR_EligibleTimeEnable].at_val.at_long == 1)
+	if (server.sv_attr[(int)SRV_ATR_EligibleTimeEnable].at_val.at_long == 1 &&
+		server.sv_attr[(int)SRV_ATR_EligibleTimeEnable].at_flags & ATR_VFLAG_SET)
 		update_eligible_time(JOB_INITIAL, pjob);
 
 	/*
--- a/src/server/job_func.c
+++ b/src/server/job_func.c
@@ -425,6 +425,7 @@ job_alloc(void)
 	/* start accruing time from the time job was created */
 	time(&ctm);
 	pj->ji_wattr[(int)JOB_ATR_sample_starttime].at_val.at_long = (long) ctm;
+	pj->ji_wattr[(int)JOB_ATR_eligible_time].at_val.at_long = 0;
 
 	/* if eligible_time_enable is not true, then job does not accrue eligible time */
 	if (server.sv_attr[(int)SRV_ATR_EligibleTimeEnable].at_val.at_long == 1) {
--- a/src/server/job_func.c
+++ b/src/server/job_func.c
@@ -139,6 +139,8 @@
 #include "renew_creds.h"
 #endif
 
+extern int time_now;
+
 /* External functions */
 
 #ifdef WIN32
@@ -340,10 +342,6 @@ job_alloc(void)
 {
 	job	*pj;
 
-#ifndef PBS_MOM
-	time_t 	ctm;
-#endif
-
 	pj = (job *)malloc(sizeof(job));
 	if (pj == NULL) {
 		log_err(errno, "job_alloc", "no memory");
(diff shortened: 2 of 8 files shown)


src/server/hook_func.c
```diff
@@ -3564,6 +3564,10 @@ do_runjob_reject_actions(job *pjob, char *hook_name)
 		}
 	}
 
+	/* update the eligible time to JOB_INITIAL */
+	if (server.sv_attr[(int)SRV_ATR_EligibleTimeEnable].at_val.at_long == 1)
```
bhroam, line 3568: Check if ATR_VFLAG_SET is set as well

src/server/job_func.c
```diff
@@ -422,21 +422,16 @@ job_alloc(void)
 
 
 #ifndef PBS_MOM
-	/* mark as JOB_INITIAL, set accrue times to zero */
-	pj->ji_wattr[(int)JOB_ATR_accrue_type].at_val.at_long = JOB_INITIAL;
-	pj->ji_wattr[(int)JOB_ATR_eligible_time].at_val.at_long = 0;
```
bhroam, line 427: I don't think you can get rid of this.  update_eligible_time() adds to the existing eligible time.  pj was malloc()'d so this is garbage.

src/server/svr_jobfunc.c
```diff
@@ -4929,36 +4916,32 @@ determine_accruetype(job* pjob)
...
 	pque = find_queuebyname(pjob->ji_qs.ji_queue);
 	if (pque != NULL)
-		if (pque->qu_attr[(int)QA_ATR_Started].at_val.at_long == 0) {
-			newaccruetype = (long)JOB_ELIGIBLE;
-			return newaccruetype;
-		}
+		if (pque->qu_attr[(int)QA_ATR_Started].at_val.at_long == 0)
+			return JOB_ELIGIBLE;
+
+	/* New job without any apparent crimes, make it accrue eligible time */
```
bhroam, line 4939: Did you mean times instead of crimes?

src/server/job_func.c
```diff
@@ -422,21 +422,16 @@ job_alloc(void)
...
 	if (server.sv_attr[(int)SRV_ATR_EligibleTimeEnable].at_val.at_long == 1) {
-		pj->ji_wattr[(int)JOB_ATR_accrue_type].at_flags |= ATR_VFLAG_SET;
+		int elig_val;
 
-		pj->ji_wattr[(int)JOB_ATR_eligible_time].at_flags |= ATR_VFLAG_SET;
-
-
-		pj->ji_wattr[(int)JOB_ATR_sample_starttime].at_flags |= ATR_VFLAG_SET;
+		elig_val = determine_accruetype(pj);
+		update_eligible_time(elig_val, pj);
```
bhroam, line 434: These 3 attributes are no longer being set in update_eligible_time(), don't they need to be set here then?

src/server/hook_func.c
agrawalravi90: There were existing instances of this which don't check for ATR_VFLAG_SET, do you think we should always do that check? As such, this value should be set to 0 when we unset it, or if it's not set right?

src/server/job_func.c
agrawalravi90: they are right? the switch case for eligible doesn't have a break so it'll trickle down to the code which sets all of them

src/server/svr_jobfunc.c
agrawalravi90: no i did mean crimes :) I meant to say that this job doesn't seem like it has a hold, or has anything else that'd prevent it from accruing eligible time .. would you like me to word it differently?

bhroam: Yeah, I was a little confused on the wording.  Maybe something like "The job doesn't have any reason to not accrue eligible time (e.g. on hold), so it should accrue it". Or something like that

src/server/svr_jobfunc.c
```diff
@@ -4985,44 +4968,32 @@ update_eligible_time(long newaccruetype, job *pjob)
...
+			pjob->ji_wattr[JOB_ATR_sample_starttime].at_val.at_long = timestamp;
 
-			pjob->ji_wattr[(int)JOB_ATR_sample_starttime].at_flags |=
-				(ATR_VFLAG_MODIFY | ATR_VFLAG_MODCACHE);
-			pjob->ji_wattr[(int)JOB_ATR_accrue_type].at_flags |=
-				(ATR_VFLAG_MODIFY | ATR_VFLAG_MODCACHE);
-			pjob->ji_wattr[(int)JOB_ATR_eligible_time].at_flags |=
-				(ATR_VFLAG_MODIFY | ATR_VFLAG_MODCACHE);
+			pjob->ji_wattr[JOB_ATR_accrue_type].at_flags |= flags;
+			pjob->ji_wattr[JOB_ATR_eligible_time].at_flags |= flags;
```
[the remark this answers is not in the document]
agrawalravi90: I'm simply reducing repetition in this function, ATR_VFLAG_MODIFY was set in all cases even before.

I really wanted to use set_attr_svr, but it accepts the value only in string format, I don't want to convert long to string and then have the decode function convert it back. In an ideal world, we'd have a setter that could accept values in any format, or different setters for each type.

src/server/job_func.c
```diff
@@ -422,21 +422,17 @@ job_alloc(void)
 
 
 #ifndef PBS_MOM
-	/* mark as JOB_INITIAL, set accrue times to zero */
-	pj->ji_wattr[(int)JOB_ATR_accrue_type].at_val.at_long = JOB_INITIAL;
-	pj->ji_wattr[(int)JOB_ATR_eligible_time].at_val.at_long = 0;
 	/* start accruing time from the time job was created */
 	time(&ctm);
-	pj->ji_wattr[(int)JOB_ATR_sample_starttime].at_val.at_long = (long)ctm;
+	pj->ji_wattr[(int)JOB_ATR_sample_starttime].at_val.at_long = (long) ctm;
```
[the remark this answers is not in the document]
agrawalravi90: Thanks for catching that, I think time_now was the right thing to set there otherwise there could be cases where accrued time might be negative.

agrawalravi90:
@nithinj thanks a lot for the review! I've addressed your comments, please provide further feedback.

src/server/svr_jobfunc.c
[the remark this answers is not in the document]
nithinj: I see that it was set even before. But there is no need to really set the flag unless you modify those values which in this case we are not doing. In this case, we are not repeating the code even if you set on the right place.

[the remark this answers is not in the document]
agrawalravi90: Ah, I got your point now, yes you are right, we are setting eligible time only for the JOB_ELIGIBLE case but setting the modify flag for it for all, which is wrong, I'll remove it, thanks.

agrawalravi90:
@nithinj gentle reminder, thanks!

src/server/svr_jobfunc.c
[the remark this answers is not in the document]
nithinj: This is not addressed.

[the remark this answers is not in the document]
agrawalravi90: Sorry about that, I'd forgotten to push the commit, see the new one now.

PR #1710, opened 2020-05-05
revert to default with cgroups

vccardenas:
#### Describe Bug or Feature
Reverting to defaults should still work if instead of running cpuset mom we are running standard mom with pbs_cgroups hook enabled with vnode_per_numa_node = true and use_hyperthreads = true

This has to be merged together with PR #1662

#### Describe Your Change
- changed the logic in is_cpuset_mom() to detect the platform where cgroups cpuset would run
- added a method enable_cgroup_cset() to revert to pbs_cgroups hook enabled with vnode_per_numa_node = true and use_hyperthreads = true
- added to revert_mom() to call enable_cgroup_cset() on the appropriate platform.

Results from modified test that shows that after setUp():
    - nodes are free
    - expected pbs_cgroups hook configuration
    - pbs_cgroups hook is enabled
[TestVerifyLogOutput_uv300.txt](https://github.com/PBSPro/pbspro/files/4600848/TestVerifyLogOutput_uv300.txt)
[TestVerifyLogOutput_uv-01.txt](https://github.com/PBSPro/pbspro/files/4600849/TestVerifyLogOutput_uv-01.txt)

Test result showing skip on cgroup cpuset system:
[TestAdminSuspend.txt](https://github.com/PBSPro/pbspro/files/4578330/TestAdminSuspend.txt)

Test results with multiple moms (uv-01,cpuset system + testdev-07-r7, non-cpuset system):
[TestVerifyLogOutput_uv-01_td-07.txt](https://github.com/PBSPro/pbspro/files/4597469/TestVerifyLogOutput_uv-01_td-07.txt)  -- uv-01 is head node
[TestVerifyLogOutput_td-07_uv-01.txt](https://github.com/PBSPro/pbspro/files/4597470/TestVerifyLogOutput_td-07_uv-01.txt) -- testdev-07-r7 is head node

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -13994,9 +13994,10 @@ class MoM(PBSService):
         Configure and enable cgroups hook
         """
         # check if cgroups subsystems including cpusets are mounted
-        with open(os.path.join(os.sep, 'proc', 'mounts'), 'r') as fd:
-            mounts = fd.read()
-        if 'cgroup' in mounts and mounts.count('cpuset') >= 2:
+        file = os.path.join(os.sep, 'proc', 'mounts')
+        mounts = self.du.cat(self.hostname, file)['out']
+        pat = 'cgroup /sys/fs/cgroup'
+        if str(mounts).count(pat) >= 6 and str(mounts).count('cpuset') >= 2:
             pbs_conf_val = self.du.parse_pbs_config(self.hostname)
             f1 = os.path.join(pbs_conf_val['PBS_EXEC'], 'lib',
                               'python', 'altair', 'pbs_hooks',
@@ -14020,7 +14021,9 @@ class MoM(PBSService):
             self.server.manager(MGR_CMD_SET, HOOK,
                                 {'enabled': 'True'}, 'pbs_cgroups')
         else:
-            self.logger.error('cgroup subsystems not mounted')
+            self.logger.error('%s: cgroup subsystems not mounted' %
+                              self.hostname)
+            raise AssertionError('cgroup subsystems not mounted')
 
 
 class Hook(PBSObject):
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -14001,37 +14001,20 @@ class MoM(PBSService):
             f1 = os.path.join(pbs_conf_val['PBS_EXEC'], 'lib',
                               'python', 'altair', 'pbs_hooks',
                               'pbs_cgroups.CF')
-            vpn1 = r'"vnode_per_numa_node"\s*:\s*false,'
-            vpn2 = '"vnode_per_numa_node"   : true,'
-            uht1 = r'"use_hyperthreads"\s*:\s*false,'
-            uht2 = '"use_hyperthreads"      : true,'
-            mem1 = r'"memory"\s*:\s*{'
-            mem2 = r'"memsw"\s*:\s*{'
-            ena1 = r'"enabled"\s*:\s*true,'
-            ena2 = '"enabled"            : false,'
             # set vnode_per_numa_node = true, use_hyperthreads = true
             with open(f1, "r") as cfg:
-                lines = cfg.readlines()
-            for i, line in enumerate(lines):
-                if re.search(vpn1, line):
-                    lines[i] = re.sub(vpn1, vpn2, line)
-                if re.search(uht1, line):
-                    lines[i] = (re.sub(uht1, uht2, line))
-                # memory subsystem not mounted, do not enable in config
-                if (re.search(mem1, line) or re.search(mem2, line)) and \
-                        mounts.count('memory') == 0:
-                    if re.search(ena1, lines[i + 1]):
-                        lines[i + 1] = (re.sub(ena1, ena2, lines[i + 1]))
+                cfg_dict = json.load(cfg)
+            cfg_dict['vnode_per_numa_node'] = 'true'
+            cfg_dict['use_hyperthreads'] = 'true'
             _, path = tempfile.mkstemp(prefix="cfg", suffix=".json")
             with open(path, "w") as cfg1:
-                for line in lines:
-                    cfg1.write((line))
+                json.dump(cfg_dict, cfg1, indent=4)
             # read in the cgroup hook configuration
             a = {'content-type': 'application/x-config',
                  'content-encoding': 'default',
                  'input-file': path}
-            rv = self.server.manager(MGR_CMD_IMPORT, HOOK, a,
-                                     'pbs_cgroups')
+            self.server.manager(MGR_CMD_IMPORT, HOOK, a,
+                                'pbs_cgroups')
             os.remove(path)
             # enable cgroups hook
             self.server.manager(MGR_CMD_SET, HOOK,
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -6053,12 +6053,13 @@ class Server(PBSService):
             self.logger.error(m)
             return None
 
-        if not submit_dir and self.du.get_platform() != 'shasta':
+        if not submit_dir:
             submit_dir = pwd.getpwnam(obj.username)[5]
 
         cwd = os.getcwd()
-        if submit_dir:
-            os.chdir(submit_dir)
+        if self.platform != 'shasta':
+            if submit_dir:
+                os.chdir(submit_dir)
         c = None
         # 1- Submission using the command line tools
         if self.get_op_mode() == PTL_CLI:
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -6053,13 +6053,12 @@ class Server(PBSService):
             self.logger.error(m)
             return None
 
-        if not submit_dir:
+        if not submit_dir and self.du.get_platform() != 'shasta':
             submit_dir = pwd.getpwnam(obj.username)[5]
 
         cwd = os.getcwd()
-        if self.platform != 'shasta':
-            if submit_dir:
-                os.chdir(submit_dir)
+        if submit_dir:
+            os.chdir(submit_dir)
         c = None
         # 1- Submission using the command line tools
         if self.get_op_mode() == PTL_CLI:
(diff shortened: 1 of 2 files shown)


test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -14002,6 +13989,56 @@ def add_mom_dyn_res(self, custom_resource, script_body=None,
...
+            uht2 = '"use_hyperthreads"      : true,'
+            mem1 = r'"memory"\s*:\s*{'
+            mem2 = r'"memsw"\s*:\s*{'
+            ena1 = r'"enabled"\s*:\s*true,'
+            ena2 = '"enabled"            : false,'
+            # set vnode_per_numa_node = true, use_hyperthreads = true
+            with open(f1, "r") as cfg:
+                lines = cfg.readlines()
+            for i, line in enumerate(lines):
+                if re.search(vpn1, line):
```
arungrover, line 14016: Voltaire, what happens when the search fails to find vnode_per_numa_node and use_hyperthreads in the config file? Shouldn't we be adding the feilds even in cases search fails because the lines are not present?

vccardenas: Since I am getting the data from the default config file pbs_cgroups.CF, if vnode_per_numa_node and use_hyperthreads lines are missing then there is a bug in that file.  Even if the lines were intentionally left out of the pbs_cgroups.CF if we added those lines later on, then it would not be based on a default configuration anymore.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -14002,6 +13989,56 @@ def add_mom_dyn_res(self, custom_resource, script_body=None,
...
+                    lines[i] = re.sub(vpn1, vpn2, line)
+                if re.search(uht1, line):
+                    lines[i] = (re.sub(uht1, uht2, line))
+                # memory subsystem not mounted, do not enable in config
+                if (re.search(mem1, line) or re.search(mem2, line)) and \
```
lisa-altair, line 14021: Is there something in install that sets memsw depending on whether or not the subsystem is mounted?  I didn't think there was...I believe memsw is enabled to "true" by default.  If there is no install code that changes the setting of memsw, then it shouldn't be set here either.

vccardenas: Yes mem and memsw are enabled to true by default.  Setting them here as false will allow us to run PTL tests on systems that don't have memory subsystem mounted.

vccardenas: I will remove setting mem and memsw to false here in the next iteration since that would not be the intended cgroups hook default setting.

test/fw/ptl/lib/pbs_testlib.py
arungrover: Oh ok, I get it. But, what if in future we remove (maybe by mistake) it from the default config for some reason. The reason that you are adding it to revert_mom is that you want to provide a consistent configuration that works on cpuset machines. I feel it is better to just replace the lines (if found) or add them in case it is not there.

vccardenas: I will make changes accordingly in the next iteration.

arungrover: Thanks

lisa-altair: @arungrover  and @vccardenas But by adding it to revert_mom, testing will now miss the fact that those settings are not there in the PBS_EXEC area.  Maybe rather than just setting it, an error should be printed if those 2 particular configurations are not found?

arungrover: I think to check that default config is proper or not isn't really the purpose of a function that's supposed to enable cgroups for cpuset machines.
I actually feel that such a check is not needed. But If it is, then it can be done from the setUpClass itself to do some sanity check on the provided configuration and then bail out if a problem exists.

vstumpf: The file is a json file, so you should be able to read it, convert it to a dictionary, change the values, and print out json.

vstumpf:
The cgroups config file is a json file, so you should be able to read it, convert it to a dictionary, change the values, and print out json. You shouldn't have to do search/replace.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -14002,6 +13989,56 @@ def add_mom_dyn_res(self, custom_resource, script_body=None,
         self.add_config(a)
         return res_file
 
+    def enable_cgroup_cset(self):
+        """
+        Configure and enable cgroups hook
+        """
+        # check if cgroups subsystems including cpusets are mounted
+        with open(os.path.join(os.sep, 'proc', 'mounts'), 'r') as fd:
```
vstumpf, line 13997: If we're talking about a remote mom, you currently read the file on the local host. You need to use get the mounts from the remote machine.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -14002,6 +13989,56 @@ def add_mom_dyn_res(self, custom_resource, script_body=None,
...
+            _, path = tempfile.mkstemp(prefix="cfg", suffix=".json")
+            with open(path, "w") as cfg1:
+                for line in lines:
+                    cfg1.write((line))
+            # read in the cgroup hook configuration
+            a = {'content-type': 'application/x-config',
+                 'content-encoding': 'default',
+                 'input-file': path}
+            rv = self.server.manager(MGR_CMD_IMPORT, HOOK, a,
+                                     'pbs_cgroups')
```
vstumpf, line 14034: If you have two moms with different features, you're going to overwrite the first one with the second one.
Do tests ever run with 2 moms, one with cpusets and one with no cpusets? It will probably fail with this PR.

test/fw/ptl/lib/pbs_testlib.py
vccardenas: I switched to using json. We are on a local host.

test/fw/ptl/lib/pbs_testlib.py
vccardenas: On these cgroup cpuset platforms we use one mom/mom image.

vccardenas: The moms will have the same cgroups hook configuration.  Reverting all the moms are done in revert_moms().

mike0042:
@vstumpf, @arungrover - Where do we stand with the latest changes?

test/fw/ptl/lib/pbs_testlib.py
vstumpf: Will a mom on a cpuset node and a mom on a non-cpuset node work with the same cgroups configuration?

test/fw/ptl/lib/pbs_testlib.py
vstumpf: If you have two moms, one of them is probably on a remote host. However, you'll still be checking the local host for the cgroups mounts.

mike0042: Not only that, if I specify a mom list that doesn't include the local host, it will only work if the systems are identical.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -14002,6 +13989,42 @@ def add_mom_dyn_res(self, custom_resource, script_body=None,
...
 
+    def enable_cgroup_cset(self):
+        """
+        Configure and enable cgroups hook
+        """
+        # check if cgroups subsystems including cpusets are mounted
+        file = os.path.join(os.sep, 'proc', 'mounts')
+        mounts = self.du.cat(self.hostname, file)['out']
+        pat = 'cgroup /sys/fs/cgroup'
+        if str(mounts).count(pat) >= 6 and str(mounts).count('cpuset') >= 2:
```
vstumpf, line 14000: what is this? can you give an example of what it should look like

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -14002,6 +13989,42 @@ def add_mom_dyn_res(self, custom_resource, script_body=None,
...
+            pbs_conf_val = self.du.parse_pbs_config(self.hostname)
+            f1 = os.path.join(pbs_conf_val['PBS_EXEC'], 'lib',
+                              'python', 'altair', 'pbs_hooks',
+                              'pbs_cgroups.CF')
```
vstumpf, line 14004: does this file exist on an execution-only host?

test/fw/ptl/lib/pbs_testlib.py
vccardenas: I'm looking for the pattern consistent in cpuset systems in /proc/mounts, for example:

cgroup /sys/fs/cgroup/systemd cgroup rw,nosuid,nodev,noexec,relatime,xattr,release_agent=/usr/lib/systemd/systemd-cgroups-agent,name=systemd 0 0
cgroup /sys/fs/cgroup/cpuset cgroup rw,nosuid,nodev,noexec,relatime,cpuset 0 0
cgroup /sys/fs/cgroup/cpu,cpuacct cgroup rw,nosuid,nodev,noexec,relatime,cpuacct,cpu 0 0
cgroup /sys/fs/cgroup/memory cgroup rw,nosuid,nodev,noexec,relatime,memory 0 0
cgroup /sys/fs/cgroup/devices cgroup rw,nosuid,nodev,noexec,relatime,devices 0 0
cgroup /sys/fs/cgroup/freezer cgroup rw,nosuid,nodev,noexec,relatime,freezer 0 0
cgroup /sys/fs/cgroup/net_cls cgroup rw,nosuid,nodev,noexec,relatime,net_cls 0 0
cgroup /sys/fs/cgroup/blkio cgroup rw,nosuid,nodev,noexec,relatime,blkio 0 0
cgroup /sys/fs/cgroup/perf_event cgroup rw,nosuid,nodev,noexec,relatime,perf_event 0 0
cgroup /sys/fs/cgroup/hugetlb cgroup rw,nosuid,nodev,noexec,relatime,hugetlb 0 0
cpuset /dev/cpuset cgroup rw,relatime,cpuset 0 0

test/fw/ptl/lib/pbs_testlib.py
vccardenas: This file does not exist on the execution host.  Since we are running PTL on the server node, PTL is able to pick up that file.  The server then pushes out the cgroups config to the moms.  I attached tests that I ran on uv-01 and td-07 (e.g. TestVerifyLogOutput_uv-01_td-07.txt) that show that only the cpuset machine got vnoded and the non-cpuset machine that has sockets did not get vnoded.

test/fw/ptl/lib/pbs_testlib.py
vstumpf: Ah, that makes sense

test/fw/ptl/lib/pbs_testlib.py
vstumpf: Ok, looks good

arungrover:
Thanks for addressing the comments @vccardenas

vccardenas:
@vstumpf  and @arungrover thanks for the reviews and sign offs.

vccardenas:
@bayucan , @lisa-altair , @mike0042 thanks for the review comments.

PR #1717, opened 2020-05-07
Log supported_auth_methods on start/restart or HUP the Server/Comm daemon and added code to write pid in windows mom.lock file

Bhagat-Rajput:
#### Describe Bug or Feature

* Earlier PBS doesn't log pbs_supported_auth_methods in the log files of server and comm
* Missing process_id in mom.lock file on windows

#### Describe Your Change
* Added a function which will write pbs_supported_auth_methods in the respective log files
* Added code to write pid in the mom.lock file on windows

server/comm logs: [log_auth_method.txt](https://github.com/PBSPro/pbspro/files/4597172/log_auth_method.txt)

ptl logs : [ptl_supported_auth_method.txt](https://github.com/PBSPro/pbspro/files/4597176/ptl_supported_auth_method.txt)

### DIFF
--- a/src/include/libutil.h
+++ b/src/include/libutil.h
@@ -245,12 +245,6 @@ lock_file(FILE *fp, int op, char *filename, int lock_retry,
  */
 char **break_comma_list(char *list);
 
-/**
- *		join_str_list - Join an array of strings using separator and returns
- *		                as a single string
- */
-char *join_str_list(char **list, char *sep);
-
 /*
  *      break_delimited_str - break apart a delimited string into an array
  *                         of strings
--- a/src/include/libutil.h
+++ b/src/include/libutil.h
@@ -245,6 +245,12 @@ lock_file(FILE *fp, int op, char *filename, int lock_retry,
  */
 char **break_comma_list(char *list);
 
+/**
+ *		join_str_list - Join an array of strings using separator and returns
+ *		                as a single string
+ */
+char *join_str_list(char **list, char *sep);
+
 /*
  *      break_delimited_str - break apart a delimited string into an array
  *                         of strings
--- a/src/lib/Liblog/setup_env.c
+++ b/src/lib/Liblog/setup_env.c
@@ -43,6 +43,7 @@
 #include <string.h>
 #include "portability.h"
 #include "log.h"
+#include "libutil.h"
 #ifdef WIN32
 #include "win.h"
 #endif
@@ -168,19 +169,17 @@ setup_env(char *filen)
  */
 void
 log_supported_auth_methods(char **supported_auth_methods) {
-
-	if (supported_auth_methods != NULL) {
-		int i;
-		char temp_buf[100 + 1] = {'\0'}; //MAXAUTHNAME 100
-		char *sep = ",";
-		for (i = 0; supported_auth_methods[i] != NULL; i++) {
-			strcat(temp_buf, supported_auth_methods[i]);
-			if (supported_auth_methods[i + 1] != NULL)
-				strcat(temp_buf, sep);
+	char *ret_str;
+	char *sep = ",";
+	if (supported_auth_methods) {
+		ret_str = join_str_list(supported_auth_methods, sep);
+		if (ret_str != NULL) {
+			sprintf(log_buffer, "PBS_SUPPORTED_AUTH_METHODS=%s", ret_str);
+			log_eventf(PBSEVENT_FORCE, PBS_EVENTCLASS_SERVER, LOG_INFO, msg_daemonname, log_buffer);
+			free(ret_str);
+		} else {
+			log_eventf(PBSEVENT_FORCE, PBS_EVENTCLASS_SERVER, LOG_INFO, msg_daemonname,
+					   "Failed to get PBS_SUPPORTED_AUTH_METHODS list");
 		}
-		sprintf(log_buffer, "PBS_SUPPORTED_AUTH_METHODS=%s", temp_buf);
-		log_event(PBSEVENT_FORCE, PBS_EVENTCLASS_SERVER, LOG_INFO, msg_daemonname, log_buffer);
-	} else {
-		log_event(PBSEVENT_FORCE, PBS_EVENTCLASS_SERVER, LOG_INFO, msg_daemonname, "PBS_SUPPORTED_AUTH_METHODS=None");
 	}
 }
--- a/src/lib/Liblog/setup_env.c
+++ b/src/lib/Liblog/setup_env.c
@@ -157,3 +157,31 @@ setup_env(char *filen)
 	fclose(efile);
 	return (-1);
 }
+
+/**
+ * @brief
+ * 	log_supported_auth_methods() - print PBS_SUPPORTED_AUTH_METHODS to the log file.
+ *
+ * @param[in]	supported_auth_methods = An array of supported methods
+ *
+ * @return	void
+ */
+void
+log_supported_auth_methods(char **supported_auth_methods) {
+
+	if (supported_auth_methods != NULL) {
+		int i;
+		char temp_buf[100 + 1]; //MAXAUTHNAME 100
+		char *sep = ",";
+		memset(temp_buf, '\0', sizeof(temp_buf));
+		for (i = 0; supported_auth_methods[i] != NULL; i++) {
+			strcat(temp_buf, supported_auth_methods[i]);
+			if (supported_auth_methods[i + 1] != NULL)
+				strcat(temp_buf, sep);
+		}
+		sprintf(log_buffer, "PBS_SUPPORTED_AUTH_METHODS=%s", temp_buf);
+		log_event(PBSEVENT_FORCE, PBS_EVENTCLASS_SERVER, LOG_INFO, msg_daemonname, log_buffer);
+	} else {
+		log_event(PBSEVENT_FORCE, PBS_EVENTCLASS_SERVER, LOG_INFO, msg_daemonname, "PBS_SUPPORTED_AUTH_METHODS=None");
+	}
+}
--- a/src/lib/Liblog/setup_env.c
+++ b/src/lib/Liblog/setup_env.c
@@ -43,7 +43,6 @@
 #include <string.h>
 #include "portability.h"
 #include "log.h"
-#include "libutil.h"
 #ifdef WIN32
 #include "win.h"
 #endif
@@ -161,7 +160,7 @@ setup_env(char *filen)
 
 /**
  * @brief
- * 	log_supported_auth_methods() - print PBS_SUPPORTED_AUTH_METHODS to the log file.
+ * 	log_supported_auth_methods() - log PBS_SUPPORTED_AUTH_METHODS
  *
  * @param[in]	supported_auth_methods = An array of supported methods
  *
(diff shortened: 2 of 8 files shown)


src/lib/Liblog/setup_env.c
```diff
@@ -157,3 +157,31 @@ setup_env(char *filen)
...
+		for (i = 0; supported_auth_methods[i] != NULL; i++) {
+			strcat(temp_buf, supported_auth_methods[i]);
+			if (supported_auth_methods[i + 1] != NULL)
+				strcat(temp_buf, sep);
+		}
+		sprintf(log_buffer, "PBS_SUPPORTED_AUTH_METHODS=%s", temp_buf);
+		log_event(PBSEVENT_FORCE, PBS_EVENTCLASS_SERVER, LOG_INFO, msg_daemonname, log_buffer);
+	} else {
+		log_event(PBSEVENT_FORCE, PBS_EVENTCLASS_SERVER, LOG_INFO, msg_daemonname, "PBS_SUPPORTED_AUTH_METHODS=None");
+	}
```
npadole20, line 186: According to the test log, when PBS_SUPPORTED_AUTH_METHOD is not set, we are logging"PBS_SUPPORTED_AUTH_METHODS=resvport". I do not understand in what condition we will log "PBS_SUPPORTED_AUTH_METHODS=None".

test/tests/functional/pbs_verify_log_output.py
```diff
@@ -152,3 +152,47 @@ def test_auto_sched_cycle_trigger(self):
                                          starttime=started_time)
             except PtlLogMatchError:
                 self.assertFalse(True)
+
+    def test_supported_auth_method_msgs(self):
+        """
+        Test case to verify pbs_supported_auth_method messages in the log
+        during start/restart or HUP the SERVER/COMM daemon
```
npadole20, line 159: Could be reworded to - Test to verify PBS_SUPPORTED_AUTH_METHODS is logged in server and comm daemon logs after start, restart and HUP.

test/tests/functional/pbs_verify_log_output.py
```diff
@@ -152,3 +152,47 @@ def test_auto_sched_cycle_trigger(self):
...
+        """
+        attr_name = 'PBS_SUPPORTED_AUTH_METHODS'
+        started_time = time.time()
+        # check the logs after restart the server and comm daemon
```
npadole20, line 163: Nit: after restarting

src/lib/Liblog/setup_env.c
Bhagat-Rajput: Yes, that's right but i have added that for worst case. If somehow it's fail to load default as well so this way we will get to know from the logs. Hope it's fine or else we can remove it also.

src/lib/Liblog/setup_env.c
```diff
@@ -157,3 +157,31 @@ setup_env(char *filen)
...
+ * @param[in]	supported_auth_methods = An array of supported methods
+ *
+ * @return	void
+ */
+void
+log_supported_auth_methods(char **supported_auth_methods) {
+
+	if (supported_auth_methods != NULL) {
+		int i;
+		char temp_buf[100 + 1]; //MAXAUTHNAME 100
```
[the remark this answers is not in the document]
Bhagat-Rajput: These macros are in use at more than once and checking some buffer overflow stuff and in our case i don't think so it's a concern for us because our buffer is directly dependent on MAXAUTHNAME macro but for that don't want to load a whole file just for one macro.

src/lib/Liblog/setup_env.c
hirenvadalia: @Bhagat-Rajput In pbs_loadconf(), if loadconf finish successfully then `pbs_conf.supported_auth_methods` will never be null, and if you reach in this function means loadconf() is finished successfully. so @npadole20 is right, you need else part.

src/lib/Liblog/setup_env.c
```diff
@@ -157,3 +157,30 @@ setup_env(char *filen)
...
+ * @param[in]	supported_auth_methods = An array of supported methods
+ *
+ * @return	void
+ */
+void
+log_supported_auth_methods(char **supported_auth_methods) {
+
+	if (supported_auth_methods != NULL) {
+		int i;
+		char temp_buf[100 + 1] = {'\0'}; //MAXAUTHNAME 100
```
hirenvadalia, line 174: This is wrong, MAXAUTHNAME is max length of individual auth method name length, not max length of the value of `PBS_SUPPORTED_AUTH_METHODS`, that means we don't have a length limit on `PBS_SUPPORTED_AUTH_METHODS` so you can't use a buffer for it.

So to solve the above issue and making code better I have an alternate approach to below for loop:
How about we define one method in libutil called, `join_str_list(list, sep)` (aka reverse method of `break_comm_list()` and use it here:
```c
char *join_str_list(char **list, char *sep)
{
	size_t sep_len = 0;
	int i = 0;
	char *ret = NULL;
	size_t ret_len = 0;

	if (list == NULL || sep == NULL) // Add more check if needed
		return NULL;
	
	sep_len = strlen(sep);
	while(list[i] != NULL) {
		ret_len += strlen(list[i]);
		ret_len += sep_len;
		i++;
	}
	ret = (char *)calloc(1, ret_len + 1);
	if (ret == NULL)
		return NULL;
	i = 0;
	while(list[i] != NULL) {
		strcat(ret, list[i]);
		strcat(ret, sep);
	}
	return ret; // caller has to free this
}
```

src/lib/Liblog/setup_env.c
```diff
@@ -157,3 +157,30 @@ setup_env(char *filen)
...
+		char *sep = ",";
+		for (i = 0; supported_auth_methods[i] != NULL; i++) {
+			strcat(temp_buf, supported_auth_methods[i]);
+			if (supported_auth_methods[i + 1] != NULL)
+				strcat(temp_buf, sep);
+		}
+		sprintf(log_buffer, "PBS_SUPPORTED_AUTH_METHODS=%s", temp_buf);
+		log_event(PBSEVENT_FORCE, PBS_EVENTCLASS_SERVER, LOG_INFO, msg_daemonname, log_buffer);
```
hirenvadalia, line 182: use log_eventf

src/lib/Liblog/setup_env.c
```diff
@@ -168,19 +169,17 @@ setup_env(char *filen)
...
-
-	if (supported_auth_methods != NULL) {
-		int i;
-		char temp_buf[100 + 1] = {'\0'}; //MAXAUTHNAME 100
-		char *sep = ",";
-		for (i = 0; supported_auth_methods[i] != NULL; i++) {
-			strcat(temp_buf, supported_auth_methods[i]);
-			if (supported_auth_methods[i + 1] != NULL)
-				strcat(temp_buf, sep);
+	char *ret_str;
```
hirenvadalia, line 172: put this under if condition

src/lib/Liblog/setup_env.c
```diff
@@ -168,19 +169,17 @@ setup_env(char *filen)
...
+	char *sep = ",";
```
hirenvadalia, line 173: you don't need this variable, you can directly pass `","` in `job_str_list()`

src/lib/Liblog/setup_env.c
```diff
@@ -168,19 +169,17 @@ setup_env(char *filen)
...
+	if (supported_auth_methods) {
+		ret_str = join_str_list(supported_auth_methods, sep);
+		if (ret_str != NULL) {
+			sprintf(log_buffer, "PBS_SUPPORTED_AUTH_METHODS=%s", ret_str);
+			log_eventf(PBSEVENT_FORCE, PBS_EVENTCLASS_SERVER, LOG_INFO, msg_daemonname, log_buffer);
```
hirenvadalia, line 178: change this to `log_eventf(PBSEVENT_FORCE, PBS_EVENTCLASS_SERVER, LOG_INFO, msg_daemonname, "PBS_SUPPORTED_AUTH_METHODS = %s", ret_str);` and remove above sprintf

src/lib/Libutil/misc_utils.c
```diff
@@ -1177,6 +1177,52 @@ break_comma_list(char *strlist)
 	return (break_delimited_str(strlist, ','));
 }
 
+/**
+ *
+ *	@brief: Join an array of strings using separator and returns as a single string
```
hirenvadalia, line 1182: remove `:` after `@brief`

src/include/libutil.h
```diff
@@ -245,6 +245,12 @@ lock_file(FILE *fp, int op, char *filename, int lock_retry,
  */
 char **break_comma_list(char *list);
 
+/**
+ *		join_str_list - Join an array of strings using separator and returns
+ *		                as a single string
```
prakashcv13, line 250: remove as

src/lib/Liblog/setup_env.c
```diff
@@ -157,3 +158,24 @@ setup_env(char *filen)
 	fclose(efile);
 	return (-1);
 }
+
+/**
+ * @brief
+ * 	log_supported_auth_methods() - print PBS_SUPPORTED_AUTH_METHODS to the log file.
```
prakashcv13, line 164: just say log PBS_SUPPORTED_AUTH_METHODS

src/lib/Libutil/misc_utils.c
```diff
@@ -1177,6 +1177,52 @@ break_comma_list(char *strlist)
 	return (break_delimited_str(strlist, ','));
 }
 
+/**
+ *
+ *	@brief Join an array of strings using separator and returns as a single string
```
prakashcv13, line 1182: remove as

src/lib/Libutil/misc_utils.c
```diff
@@ -1177,6 +1177,52 @@ break_comma_list(char *strlist)
...
+ *
+ *	NOTE: Caller has to free the return value
+ *
+ *	@param[in] list	-	an array of string values
+ *	@param[in] sep	-	separator
+ *
+ *	@return char *
```
prakashcv13, line 1189: specify return value on success and failure.

src/lib/Liblog/setup_env.c
```diff
@@ -157,3 +158,24 @@ setup_env(char *filen)
...
+ *
+ * @param[in]	supported_auth_methods = An array of supported methods
+ *
+ * @return	void
+ */
+void
+log_supported_auth_methods(char **supported_auth_methods) {
+	if (supported_auth_methods) {
+		char *ret_str = join_str_list(supported_auth_methods, ",");
```
prakashcv13, line 173: instead of having a function that makes 2 passes over the values - can we do something like - 
int i = 0;
log_buffer[0]; // you can use a local buffer of around 1024 bytes as well. I feel 1024 would be enough, after all the value is coming from pbs.conf.
while (supported_auth_methods[i])
        snprintf(log_buffer, sizeof(log_buffer, sizeof(log_buffer) - strlen(log_buffer), "%s, ", supported_auth_methods[i]);
log_eventf(...)

hirenvadalia: @prakashcv13 I don't agree with you here, practically log_buffer will be enough in most cases, but theoretically, we don't have any limit on PBS_SUPPORTED_AUTH_METHODS so log_buffer will not sufficient.

prakashcv13: okay - not sure about PBS_SUPPORTED_AUTH_METHODS, but do we expect to have a value with length > 4096 from pbs.conf?

hirenvadalia: yes, we can have for many keys like PBS_LEAF_ROUTERS, PBS_COMM_ROUTERS...

subhasisb: This is s very simple thing and does not have to be so complicated code. I would just do a snprintf(sizeof(log_buffer) and take care of the theoretical correctness. Please change this to the 2 simple lines it needs to be

src/lib/Libutil/misc_utils.c
```diff
@@ -1177,6 +1177,52 @@ break_comma_list(char *strlist)
...
+ */
+char *
+join_str_list(char **list, char *sep)
+{
```
subhasisb, line 1193: delete this function

src/lib/Liblog/setup_env.c
hirenvadalia: @subhasisb honestly I like current code with join_str_list, it's more readable and it can handle all cases (including the negative case when you have very long value of PBS_SUPPORTED_AUTH_METHODS, while using log_buffer it can give buffer overflow error)
+ join_str_list can use used other places (maybe not now but in future, we need to join str list)

But finally up to u guys... 😄

subhasisb: Hiren, there is no practical chance of that variable being that large (you know that we wont have such a long list of thing to support). 

I would just code this as follows:
snprintf(log_buffer, sizeof(log_buffer), "support_aith_methods=", pbs_conf.supported_auth_methods);

PR #1688, opened 2020-04-24
modified run_copy to do cp on remotehost if src and target hosts are same

visheshh:
#### Describe Bug or Feature
run_copy does scp when asked to do a cp src dest on a remote machine. 

#### Describe Your Change
Added an attribute src host to tell run_copy where the src file is present.
Added a check if srchost and target host are same then run cp instead of scp.

#### Attach Test and Valgrind Logs/Output

run_copy tests
mainline_tests

PR #1720, opened 2020-05-08
Fix pbs_snaphot to copy logs from remote host to local host

shilpakodli:
#### Describe Bug or Feature
pbs_snapshot was failing to copy snapshot logs from remote host to local host and thus, finsla tar file was not containing remote host snap tar file. 

#### Describe Your Change
Made changes in run_copy in pbs_snapshot file to pass remote host in srchost param.

[snapshot_capure_after_fix.txt](https://github.com/PBSPro/pbspro/files/4597673/snapshot_capure_after_fix.txt)

### DIFF
--- a/test/fw/bin/pbs_snapshot
+++ b/test/fw/bin/pbs_snapshot
@@ -154,8 +154,7 @@ def remotesnap_thread(logger, host):
 
     # Copy over the snapshot tar file as <hostname>_snapshot.tgz
     dest_path = os.path.join(out_dir, host + "_snapshot.tgz")
-    src_path = host + ":" + snaptarname
-    ret = du.run_copy(src=src_path, dest=dest_path)
+    ret = du.run_copy(srchost=host, src=snaptarname, dest=dest_path)
     if ret['rc'] != 0:
         logger.error("Error copying child snapshot from host %s" % (host))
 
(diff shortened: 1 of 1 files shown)


kjakkali:
> 
> 
> @shilpakodli This used to work before, so we have existing PTL tests which test this functionality, can you please run them (pbs_snapshot_unittest.py) and post logs here?
> 
> @hirenvadalia and @kjakkali do we run pbs_snapshot tests in TH3? If not, can we please add them? This used to work before, so it must have started failing after a framework change. So, if we run these tests as part of the regression suite then issues like this can be caught easily with TH3 in the future.

@agrawalravi90  : I have not added snapshot test case to TH3 regression profile. as it keeps on failing. There were issue logged for it.

shilpakodli:
> @shilpakodli This used to work before, so we have existing PTL tests which test this functionality, can you please run them (pbs_snapshot_unittest.py) and post logs here?
> 
> @hirenvadalia and @kjakkali do we run pbs_snapshot tests in TH3? If not, can we please add them? This used to work before, so it must have started failing after a framework change. So, if we run these tests as part of the regression suite then issues like this can be caught easily with TH3 in the future.

Hi @agrawalravi90 Yes. This used to work before until PR merge: https://github.com/PBSPro/pbspro/pull/1688
In above PR, parameters of run_copy have changed for remote host copy. That's why pbs_snapshot was failing to copy logs from remote host to local.

agrawalravi90:
> @agrawalravi90 : I have not added snapshot test case to TH3 regression profile. as it keeps on failing. There were issue logged for it.

I just ran it and all the single host tests passed. That's why I think we should run the multi-node tests from that suite with the changes in this PR to check that the issues with capturing remote host data have been resolved now.

@shilpakodli thanks for clarifying, can you please run the whole testsuite with your changes on a multi-node setup and see that it passes now?

shilpakodli:
Hi @agrawalravi90 Yes. I ran multinode tests with my changes and now it's capturing remote host logs:
I have attached ls output of tar file in description which was captured in TH3 for multinode failed tests.

kjakkali:
@agrawalravi90 : Sure.

PR #1723, opened 2020-05-09
If condition didn't scope DBPRT message

arungrover:
#### Describe Bug or Feature
PR #1707 cause an issue where standing reservations after one occurrence were getting deleted.

#### Describe Your Change
The issue was that there was an if condition (without parenthesis) which had a DBPRT call inside the block.
Unfortunately, when the source is not complied in DEBUG mode, DBPRT is not defined. This caused the very next line (which was about setting the reservation state) to be considered inside the if condition which was wrong.

#### Link to Design Doc
NA

[Test_passed.txt](https://github.com/PBSPro/pbspro/files/4602620/Test_passed.txt)
[Test_failed.txt](https://github.com/PBSPro/pbspro/files/4602621/Test_failed.txt)

### DIFF
--- a/src/server/svr_jobfunc.c
+++ b/src/server/svr_jobfunc.c
@@ -3516,8 +3516,9 @@ Time4occurrenceFinish(resc_resv *presv)
 			set_resv_retry(presv, time_now + 120);
 	}
 
-	if (sub == RESV_DEGRADED)
+	if (sub == RESV_DEGRADED) {
 		DBPRT(("degraded_time of %s is %s", presv->ri_qs.ri_resvID, ctime(&presv->ri_degraded_time)))
+	}
 
 	/* Set the reservation state and substate */
 	resv_setResvState(presv, state, sub);


hirenvadalia:
Hey @arungrover I am not sure but shouldn't this caught by smoke tests/ci? or we don't have any standing reservation tests in smoke tests?
if we don't have then shouldn't we need to add at least one standing reservation test in smoke tests?

kjakkali:
> 
> 
> Hey @arungrover I am not sure but shouldn't this caught by smoke tests/ci? or we don't have any standing reservation tests in smoke tests?
> if we don't have then shouldn't we need to add at least one standing reservation test in smoke tests?

@hirenvadalia : We have " test_standing_reservation" test.

arungrover:
Guys the test needs to wait for one occurrence to end and another one to start. There are a few standing reservation tests in multi sched test suite as well but those didn’t fail as well.

vstumpf:
I can merge this in, right?

hirenvadalia:
@vstumpf yes please, test part we can handle in seperate PR

PR #1731, opened 2020-05-11
Fixed crash with pbs_ralter and superchunks

bhroam:
#### Describe Bug or Feature
The recent RFE to reconfirm running reservations would crash if a reservation had a superchunk (a chunk broken up across multiple vnodes on the same host).

#### Describe Your Change
This required a decent rewrite of the feature.  The scheduler now needs to map chunks of the select to the chunks of the exec_vnode.  When it sees a superchunk that any of the vnodes are unavailable, it removes the superchunk and replaces it with the original chunk from the select.  When we create the internal select, the original chunk will be requested again.

[resv.log](https://github.com/PBSPro/pbspro/files/4612185/resv.log)

### DIFF
--- a/src/scheduler/node_info.c
+++ b/src/scheduler/node_info.c
@@ -696,6 +696,7 @@ new_node_info()
 	new->is_stale = 0;
 	new->is_maintenance = 0;
 	new->is_provisioning = 0;
+	new->is_sleeping = 0;
 	new->is_multivnoded = 0;
 	new->has_ghost_job = 0;
 
@@ -2337,6 +2338,7 @@ new_nspec()
 	ns->go_provision = 0;
 	ns->ninfo = NULL;
 	ns->resreq = NULL;
+	ns->chk = NULL;
 
 	return ns;
 }
@@ -2368,12 +2370,13 @@ free_nspec(nspec *ns)
  *
  * @param[in]	ons	-	the nspec to duplicate
  * @param[in]	nsinfo	-	the new server info
+ * @param[in]	sel	-	select spec to map nspec to
  *
  * @return	newly duplicated nspec
  *
  */
 nspec *
-dup_nspec(nspec *ons, node_info **ninfo_arr)
+dup_nspec(nspec *ons, node_info **ninfo_arr, selspec *sel)
 {
 	nspec *nns;
 
@@ -2391,6 +2394,8 @@ dup_nspec(nspec *ons, node_info **ninfo_arr)
 	nns->go_provision = ons->go_provision;
 	nns->ninfo = find_node_by_indrank(ninfo_arr, ons->ninfo->node_ind, ons->ninfo->rank);
 	nns->resreq = dup_resource_req_list(ons->resreq);
+	if (sel != NULL)
+		nns->chk = find_chunk_by_seq_num(sel->chunks, ons->seq_num);
 
 	return nns;
 }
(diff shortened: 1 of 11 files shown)


src/scheduler/resv_info.c
```diff
@@ -1636,92 +1656,148 @@ confirm_reservation(status *policy, int pbs_sd, resource_resv *unconf_resv, serv
...
+ * @retval 1 - running jobs, no unavailable nodes
+ * @retval 0 no unavailable nodes, no running jobs
+ * @retval -1 unavailable nodes, but no running jobs on the chunk
+ * @retval -2 unavilable nodes, running jobs within the chunk
+ * 
+ */
+int check_down_running(resource_resv *resv, int chunk_ind)
+{
+	int i, j, k;
+	int ret = 0;
```
arungrover, line 1675: Can you please add a check to make sure resv is a reservation and not a job. Maybe also change the name of the function to say that it is only meant for reservation.

src/scheduler/resv_info.c
```diff
@@ -1636,92 +1656,148 @@ confirm_reservation(status *policy, int pbs_sd, resource_resv *unconf_resv, serv
...
 
-		if (j == tot_chk_cnt) {
-			chk_ind++;
-			chk = resv->select->chunks[chk_ind];
-			tot_chk_cnt += chk->num_chunks;
+	for (i = 0; resv->orig_nspec_arr[i] != NULL; i++) {
+
+		/* If the original select chunks haven't been mapped to the resv_nodes and the reservation is running, we can't continue */
+		if (resv->resv->resv_state == RESV_RUNNING && resv->orig_nspec_arr[i]->chk == NULL) {
+			free(chunks_to_remove);
```
arungrover, line 1752: Can we log a message here?

src/scheduler/resv_info.c
```diff
@@ -1636,92 +1656,148 @@ confirm_reservation(status *policy, int pbs_sd, resource_resv *unconf_resv, serv
...
+		if (ret == -2) {
+			free(chunks_to_remove);
+			return -1;
+		}
+		/* unavailable nodes in chunk */
+		else if (ret == -1) {
+			int j;
+			has_down_node = 1;
+			for (j = i; resv->orig_nspec_arr[j] != NULL && !resv->orig_nspec_arr[j]->end_of_chunk; j++) {
+				resv->orig_nspec_arr[j]->ninfo = NULL;
```
arungrover, line 1771: Don't we need to free resreq and reassign it as we do at line 1786, 1787?

src/scheduler/resv_info.c
bhroam: We could, but it could get messy.  Any reservation in pre-upgrade to 2020 will not be able to map.  Also, we can't request the parts of a superchunk, so any running reservation can only be reconfirmed once.  After that, the exec_vnode will have more chunks than the select.  This is why I didn't add a message.  I can add one if you think we should.

arungrover: No, it's ok, you can leave it as it is.

src/scheduler/resv_info.c
bhroam: No, these are the chunks we will delete.  When we detect a superchunk with a down node, we delete all chunks except the last one.  We then have it request all the resources for the superchunk so it will be allocated elsewhere.  It doesn't matter if any node of the superchunk is up, as long as there is one down, the whole thing gets reassigned.

src/scheduler/node_info.c
```diff
@@ -4889,7 +4934,8 @@ node_state_to_str(node_info *ninfo)
 /**
  * @brief
  *		combine_nspec_array - find and combine any nspec's for the same node
- *		in an nspec array
+ *		in an nspec array.  Because nspecs no longer map to the original chunks
+ *		they came from, seq_num and chk longer have meaning.  They are cleared.
```
arungrover, line 4938: I think there is a 'no' missing here. It should say seq_num and chk no longer have meaning.

src/scheduler/node_info.c
```diff
@@ -4849,6 +4852,12 @@ parse_execvnode(char *execvnode, server_info *sinfo, selspec *sel)
 			if (simplespec != NULL) {
 				int ret;
 
+				if (hp > 0) /* simplespec starts with '(' but doesn't end with ')' */
+					in_superchunk = 1;
+				else if (hp < 0) /* simplespec ends with ')' but does not start with '(' */
+					in_superchunk = 0;
```
vstumpf, line 4858: Do we need to change anything if hp == 0?

bhroam: No, I added a comment for what hp == 0 means just for completeness.  hp == 0 means either (chunk) or chunk without ().  (chunk) is not a superchunk, so we don't need to change anything.  chunk without () is the middle of a superchunk, so we don't need to change anything either.

PR #1732, opened 2020-05-12
Attribute "create_resv_from_job" value type  is boolean

birbalsain:
#### Describe Bug or Feature
In test "test_create_resv_from_job_using_runjob_hook" value of create_resv_from_job is 1 but value type is boolean.So it should be True.

#### Describe Your Change
set the attribute "create_resv_from_job=True" in test "test_create_resv_from_job_using_runjob_hook"

[test_user_reliability.txt](https://github.com/PBSPro/pbspro/files/4616262/test_user_reliability.txt)

### DIFF
--- a/test/tests/functional/pbs_user_reliability.py
+++ b/test/tests/functional/pbs_user_reliability.py
@@ -55,7 +55,7 @@ class Test_user_reliability(TestFunctional):
 import pbs
 e = pbs.event()
 j = e.job
-j.create_resv_from_job=1
+j.create_resv_from_job=True
 """
         hook_event = "runjob"
         hook_name = "rsub"


arungrover:
@birbalsain, While I understand that the test is failing, the problem is not in the test. A boolean attribute should accept 1 as an acceptable value.
I and @bayucan did talk about this issue and it seems that one python API "PyObject_IsInstance" is failing when it shouldn't. I'd suggest that you don't change the test but rather wait for a fix in the product.

birbalsain:
@arungrover PR https://github.com/PBSPro/pbspro/pull/1679 related to this issue.

birbalsain:
I got one other issue it accept both int and boolean when it set by qsub or qalter. but it accept only boolean on set by runjob hook.

arungrover:
Right, It should ideally accept integer '1' as well. This is why I'm saying just making the change in the test will just mask the issue but that can't be the right fix.

birbalsain:
It should be done after changes in the product.

PR #1733, opened 2020-05-12
fork only if we have epilog to run while handling exiting job

hirenvadalia:
#### Describe Bug or Feature

* Currently, MoM (only on Linux), while handling exiting job it does fork to run epilog and forked child check whether it has any epilog to run or not. If no epilog to run then child simply exits
But that heavy fork can be avoided if we check first whether we have any epilog to run or not, if we have an epilog then only fork else just send obit and continue because checking for epilog is way cheap then fork.
* Also (not sure why, maybe to limit # forked processes) MoM only checks for first 5 exiting job in one call to scan_for_exiting() then returns. 5 job seams too less, we should check more (unless there is a solid reason not to check more)

#### Describe Your Change

* Changed scan_for_exiting() to check first for epilog, and fork only if we have epilog to run and if no epilog else just send obit and continue.
* Increased limit to 20 for checking exiting job. Also changed code such a way that this limit comes into picture only when we have epilog, if no epilog then no fork so there is no reason to apply a limit.

#### Link to Design Doc

None

Screen Shot 2020-05-13 at 12 15 45 AM

### DIFF
--- a/src/resmom/catch_child.c
+++ b/src/resmom/catch_child.c
@@ -1098,7 +1098,7 @@ void
 update_ajob_status_using_cmd(job *pjob, int cmd, int use_rtn_list_ext)
 {
 	struct resc_used_update rused;
-		
+
 	/* pass user-client privilege to encode_resc() */
 
 	resc_access_perm = ATR_DFLAG_MGRD;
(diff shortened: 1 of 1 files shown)


src/resmom/catch_child.c
```diff
@@ -1846,79 +1813,71 @@ scan_for_exiting(void)
...
 
 		extval = 0;
-
 		if (num_eligible_hooks(HOOK_EVENT_EXECJOB_EPILOGUE) > 0) {
 			mom_hook_input_init(&hook_input);
 			hook_input.pjob = pjob;
-			(void)mom_process_hooks(HOOK_EVENT_EXECJOB_EPILOGUE,
-				PBS_MOM_SERVICE_NAME, mom_host, &hook_input,
-				NULL, NULL, 0, 0);
+			(void)mom_process_hooks(HOOK_EVENT_EXECJOB_EPILOGUE, PBS_MOM_SERVICE_NAME, mom_host, &hook_input, NULL, NULL, 0, 0);
```
bayucan, line 1862: I see that the #ifndef WIN32 has been removed so that the same mom_process_hooks() call is done. But there's a subtle difference. The last parameter is actually a '1' (update_server) on the Windows side since it is not executing under forked process. It should be:
if (mom_process_hooks(HOOK_EVENT_EXECJOB_EPILOGUE,
                        PBS_MOM_SERVICE_NAME, mom_host, &hook_input, NULL,
                        NULL, 0, **1**)

hirenvadalia: oops, nice catch, missed that. will change it.

PR #1734, opened 2020-05-12
create_resv_from_job=False would create a reservation

bhroam:
#### Describe Bug or Feature
qsub -Wcreate_resv_from_job=False would create a reservation for the job.  When checking to see if we needed to create a reservation for a job, the server would only check if create_resv_from_job was set, but not set to True.

#### Describe Your Change
Check to see if the attribute is set and is true before creating a reservation

[resv_job.log](https://github.com/PBSPro/pbspro/files/4618201/resv_job.log)

### DIFF
--- a/test/tests/functional/pbs_user_reliability.py
+++ b/test/tests/functional/pbs_user_reliability.py
@@ -44,6 +44,7 @@ class Test_user_reliability(TestFunctional):
     """
     This test suite is for testing the user reliability workflow feature.
     """
+
     def test_create_resv_from_job_using_runjob_hook(self):
         """
         This test is for creating a reservation out of a job using runjob hook.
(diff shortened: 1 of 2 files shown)


test/tests/functional/pbs_user_reliability.py
```diff
@@ -111,7 +112,7 @@ def test_create_resv_from_job_using_qsub(self):
 
         now = time.time()
 
-        a = {ATTR_W: 'create_resv_from_job=1'}
+        a = {ATTR_W: 'create_resv_from_job=True'}
```
arungrover, line 115: Is this change needed? If not, I'll say just leave it. 
There is already a PR that Birbal raised to change the attribute value in hooks to use 'True' instead of 1 because it fails when a numeric value is given in the hook. There seems to be an issue in our hook implementation.
Ideally, the attribute should accept both '1' or 'True' (irrespective that it is being set from qsub, qalter or hooks), and having it set to 1 here proves that point.

bhroam: I changed it because I wanted the second part of the test I was adding to request False.  If I leave this as 1, then I'll have to request 0.  This is a PTL test, not a hook, so birbal's PR doesn't affect this.  PTL just wraps this up and sends it to qsub, and qsub correctly deals with decoding booleans.  If you want me to change this back to 1 (and request 0 further on), I will.

PR #1736, opened 2020-05-13
Delete saved_test_setup file and use sudo listdir

visheshh:
#### Describe Bug or Feature
saved_test_setup file was not deleted in teardown
non root user was not able to listdir mom_priv/config.d

#### Describe Your Change
Deleted the saved_test_setup file and used self.du.listdir(path=file,sudo=True) do allow sudo user to listdir mom_priv directory

### DIFF
--- a/test/fw/ptl/utils/pbs_testsuite.py
+++ b/test/fw/ptl/utils/pbs_testsuite.py
@@ -1651,6 +1651,7 @@ class PBSTestSuite(unittest.TestCase):
                 ret = mom.load_configuration(self.saved_file)
                 if not ret:
                     raise Exception("Failed to load mom's test setup")
+            self.du.rm(path=self.saved_file)
         self.log_end_teardown()
 
     @classmethod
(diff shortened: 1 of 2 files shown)


test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -1669,5 +1670,4 @@ def tearDownClass(cls):
                 ret = mom.load_configuration(cls.saved_file)
                 if not ret:
                     raise Exception("Failed to load mom's custom setup")
-        if cls.use_cur_setup:
```
RKORavi, line 1672: why is this condition removed? 
This file is saved in setupclass only when --use-current-setup option is used

visheshh: Hi @RKORavi , Code is already under the same if condition, this was redundant.

RKORavi: Sorry my bad did not notice that, thanks for pointed that out.

PR #1727, opened 2020-05-11
add error logging wraps for CloseHandle and LocalFree

Shrini-h:
#### Describe Bug or Feature

No error logs printed if CloseHandle or LocalFree fails

#### Describe Your Change

adding new wrapper functions that log the error in case of failure.
wrapper is invoked using a macro overriding all the API invocations

#### Attach Test and Valgrind Logs/Output

[20200511.txt](https://github.com/PBSPro/pbspro/files/4609332/20200511.txt)

error was forced using a debugger
```
05/11/2020 04:52:31;0100;pbs_mom;Req;;Type 1 request received from root@192.168.153.132:15001, sock=2
05/11/2020 04:52:31;0100;pbs_mom;Req;;Type 5 request received from root@192.168.153.132:15001, sock=2
05/11/2020 04:55:04;0001;pbs_mom;n/a;has_privilege;CloseHandle(procToken) at line 2810 failed with error 0, Handle[638]
05/11/2020 04:55:50;0001;pbs_mom;n/a;getdefgrpname;LocalFree(sid) at line 1923 failed with error 122, Handle[94e4728]
05/11/2020 04:56:55;0001;pbs_mom;n/a;LogonUserNoPass;CloseHandle(hToken) at line 3901 failed with error 0, Handle[640]
05/11/2020 04:57:51;0100;pbs_mom;Job;0.mixlin;allowed pbsuser to access window station and desktop, 

```

PR #1738, opened 2020-05-13
Refactor some part of Libwin code

SudeshnaMoulik:
#### Describe Bug or Feature

It is difficult to debug Windows issue due to lack of error messages.

#### Describe Your Change

- Removed unused code
- Added more messages wherever necessary


src/lib/Libwin/accesinfo.c
```diff
@@ -84,6 +84,10 @@ create_secure_dacl(char *user, ACCESS_MASK mask, SID *owner_sid)
 
 	rids[0] = DOMAIN_ALIAS_RID_ADMINS;
 	k = getgids(getlogin(), grp, rids);
+	if (k <= 0) {
+		sprintf(logb, "failed to get group-id for %s user", getlogin());
+		log_err(-1, __func__, logb);
```
subhasisb, line 89: errno?

src/lib/Libwin/accesinfo.c
```diff
@@ -106,9 +110,11 @@ create_secure_dacl(char *user, ACCESS_MASK mask, SID *owner_sid)
 				grp[k] = sid;
 				k++;
 			}
+		} else {
+			sprintf(logb, "failed to get group and user sid for %s user", user);
+			log_err(-1, __func__, logb);
```
subhasisb, line 115: errno?

src/lib/Libwin/accesinfo.c
```diff
@@ -126,7 +132,11 @@ create_secure_dacl(char *user, ACCESS_MASK mask, SID *owner_sid)
 		log_err(-1, __func__, logb);
 		return NULL;
 	}
-	InitializeAcl(ndacl, cbAcl, ACL_REVISION);
+	if (InitializeAcl(ndacl, cbAcl, ACL_REVISION) == 0) {
+		sprintf(logb, "failed to InitializeAcl with errorno %d", GetLastError());
```
subhasisb, line 136: either say error code, or errno (like linux). errorno is bit unconventional

src/lib/Libwin/accesinfo.c
```diff
@@ -139,15 +149,15 @@ create_secure_dacl(char *user, ACCESS_MASK mask, SID *owner_sid)
 
 		if (user != NULL && mask != 0 && i == (k-1)) {
 			if (AddAccessAllowedAce(ndacl, ACL_REVISION, mask | 0x00100000, grp[i]) == 0) {
-				sprintf(logb, "failed to add %d to %s", mask,
-					name);
+				sprintf(logb, "failed to add %d to %s with errorno %d", mask,
```
subhasisb, line 152: instead of printing the errnot in the log buffer, can we pass it to first parameter of log_err()?

src/lib/Libwin/accesinfo.c
```diff
@@ -184,22 +194,43 @@ secure_file(char *path, char *user, ACCESS_MASK mask)
 	char	logb[LOG_BUF_SIZE] = {'\0' } ;
 	char	*gname = NULL;
 
-	if (path == NULL || *path == '\0')
+	if (path == NULL || *path == '\0') {
+		sprintf(logb, "Invalid path");
+		log_err(-1, __func__, logb);
```
subhasisb, line 199: errno?

src/lib/Libwin/accesinfo.c
```diff
@@ -184,22 +194,43 @@ secure_file(char *path, char *user, ACCESS_MASK mask)
...
 		return (0);
+	}
 
-	if (lstat(path, &sbuf) == -1)
+	if (lstat(path, &sbuf) == -1) {
+		sprintf(logb, "Path don't exists");
+		log_err(-1, __func__, logb);
```
subhasisb, line 205: errno?

src/lib/Libwin/accesinfo.c
```diff
@@ -184,22 +194,43 @@ secure_file(char *path, char *user, ACCESS_MASK mask)
...
+			log_err(-1, __func__, logb);
+		}
+	}
 
-	if (!has_privilege(SE_TAKE_OWNERSHIP_NAME))
-		ena_privilege(SE_TAKE_OWNERSHIP_NAME);
+	if (!has_privilege(SE_TAKE_OWNERSHIP_NAME)) {
+		if (ena_privilege(SE_TAKE_OWNERSHIP_NAME) == 0) {
+			sprintf(logb, "Unable to enable privilege: SE_TAKE_OWNERSHIP_NAME");
+			log_err(-1, __func__, logb);
```
subhasisb, line 219: errno?

src/lib/Libwin/accesinfo.c
```diff
@@ -184,22 +194,43 @@ secure_file(char *path, char *user, ACCESS_MASK mask)
...
+		}
+	}
 
-	if (!has_privilege(SE_SECURITY_NAME))
-		ena_privilege(SE_SECURITY_NAME);
+	if (!has_privilege(SE_SECURITY_NAME)) {
+		if (ena_privilege(SE_SECURITY_NAME) == 0) {
+			sprintf(logb, "Unable to enable privilege: SE_SECURITY_NAME");
+			log_err(-1, __func__, logb);
```
subhasisb, line 226: errno?

src/lib/Libwin/accesinfo.c
```diff
@@ -184,22 +194,43 @@ secure_file(char *path, char *user, ACCESS_MASK mask)
...
+		}
+	}
 
-	InitializeSecurityDescriptor(&sd, SECURITY_DESCRIPTOR_REVISION);
+	if (InitializeSecurityDescriptor(&sd, SECURITY_DESCRIPTOR_REVISION) == 0) {
+		sprintf(logb, "InitializeSecurityDescriptor failed with errorno %d", GetLastError());
+		log_err(-1, __func__, logb);
```
subhasisb, line 232: errno?

src/lib/Libwin/accesinfo.c
```diff
@@ -210,21 +241,24 @@ secure_file(char *path, char *user, ACCESS_MASK mask)
 
 	if (usid) {
 		if (SetSecurityDescriptorOwner(&sd, usid, FALSE) == 0) {
-			sprintf(logb, "error setting owner for file %s", path);
-			log_err(-1, "secure_file", logb);
+			sprintf(logb, "error setting owner for file %s with errorno %d", path, GetLastError());
+			log_err(-1, __func__, logb);
```
subhasisb, line 245: Can we print errno in log_err in please of -1

src/lib/Libwin/passwd.c
```diff
@@ -395,18 +393,25 @@ sid_dup(SID *src_sid)
 {
 	DWORD	sid_len_need;
 	SID	*dest_sid = NULL;
+	char logb[LOG_BUF_SIZE] = {'\0'};
```
subhasisb, line 396: why use logb, why not use log_buffer itself?

src/lib/Libwin/accesinfo.c
SudeshnaMoulik: As this (getgids) is pbs function and we already handled error no login in corresponding steps inside that function depending on which api failed, so didn't log the code again here.

src/lib/Libwin/accesinfo.c
SudeshnaMoulik: Same as above. getusersid already handles corresponding API failures

src/lib/Libwin/accesinfo.c
SudeshnaMoulik: Right, Will change all of them to "errno"

src/lib/Libwin/accesinfo.c
SudeshnaMoulik: I didn't understand. If the path is NULL then what error code we are expecting ? I mean we are not calling in API here

subhasisb: yeah ignore that, no api call there

src/lib/Libwin/accesinfo.c
Shrini-h: why repeat getlogin() again just to get the username?

SudeshnaMoulik: yes

src/lib/Liblog/pbs_log.c
```diff
@@ -724,15 +725,15 @@ log_err(int errnum, const char *routine, const char *text)
...
+		snprintf(buf, LOG_BUF_SIZE, "GetLastError() returned: %d. ", err);
 		FormatMessage(
 			FORMAT_MESSAGE_ALLOCATE_BUFFER |
 			FORMAT_MESSAGE_FROM_SYSTEM |
 			FORMAT_MESSAGE_IGNORE_INSERTS,
 			NULL, err,
 			MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT),
 			(LPTSTR)&lpMsgBuf, 0, NULL);
-		strncpy(buf, lpMsgBuf, sizeof(buf));
+		strncat(buf, lpMsgBuf, LOG_BUF_SIZE - (int)strlen(buf));
```
RKORavi, line 736: Should this line be strncat(buf, lpMsgBuf, LOG_BUF_SIZE - strlen(log_buffer) - 1)) ?

src/lib/Libwin/dir.c
```diff
@@ -71,20 +72,23 @@ opendir(const char *name)
...
+	if (hdir == INVALID_HANDLE_VALUE) {
+		log_err(-1, __func__, "failed in FindFirstFile");
 		return NULL;
-
+	}
 	dir = (DIR *)malloc(sizeof(DIR));
 
-	if (dir == NULL)
+	if (dir == NULL) {
+		log_err(errno, __func__, "failed to allocate memory for dir");
```
RKORavi, line 82: Did you miss adding errno = GetLastError(); before this line?. errno is not defined in this function won't this throw an error here. or am i missing something?

src/lib/Libwin/dir.c
```diff
@@ -71,20 +72,23 @@ opendir(const char *name)
...
 		return NULL;
-
+	}
 
 	dir->handle = hdir;
 	dir->pos = DIR_BEGIN;
 	dir->entry = (struct dirent *)malloc(sizeof(struct dirent));
 
 	if (dir->entry == NULL) {
+		log_err(errno, __func__, "failed to allocate memory for dir->entry");
```
RKORavi, line 91: same as above

src/lib/Libwin/dir.c
SudeshnaMoulik: no its intentional. malloc sets the errno

src/lib/Libwin/popen.c
```diff
@@ -177,7 +184,9 @@ win_popen(char	*cmd, const char *type, struct pio_handles *pio, struct proc_ctrl
 		else
 			fsuccess = CreateProcess(NULL, cmd_line, &sa, &sa, proc_info->bInheritHandle,
 				CREATE_NO_WINDOW, NULL, NULL, &si, &pio->pi);
-
+		if (fsuccess == 0) {
+			log_err(-1, __func__, "failed in CreateProcess");
```
RKORavi, line 188: this condition doesn't seem right.
did you mean if(!fsuccess)(...

src/lib/Libwin/gethostid.c
```diff
@@ -147,8 +147,10 @@ gethostid(void)
...
 	sz = sizeof(cname);
 
-	if (GetComputerName(cname, &sz) == 0)
+	if (GetComputerName(cname, &sz) == 0) {
+		log_err(-1, __func__, "failed in GetComputerName");
 		return (ret);
+	}
 
 	sz1 = 0;
 	sz2 = sizeof(domain);
```
RKORavi, line 156: No log messages for other call failures - LookupAccountName, malloc,

SudeshnaMoulik: For malloc one, yes i will add log. But for others its already handled in the code

src/lib/Libwin/popen.c
SudeshnaMoulik: No this is correct, as per https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessa

src/lib/Libwin/accesinfo.c
```diff
@@ -269,8 +281,7 @@ secure_file(char *path, char *user, ACCESS_MASK mask)
 	si |= DACL_SECURITY_INFORMATION;
 
 	if (SetFileSecurity(path, si, &sd) == 0) {
-		sprintf(logb, "error setting security for file %s", path);
-		log_err(-1, "secure_file", logb);
+		log_err(-1, __func__, "failed in SetFileSecurity");
 		if (usid) LocalFree(usid);
 		if (gsid) LocalFree(gsid);
 		(void)free(pdacl);
```
sakshamgarg, line 287: We should add log at https://github.com/PBSPro/pbspro/blob/6514592d1bca52adb5cd2bdb3233be1f9838a25a/src/lib/Libwin/accesinfo.c#L303

src/lib/Libwin/accesinfo.c
```diff
@@ -526,8 +549,7 @@ secure_file2(char *path, char *user, ACCESS_MASK mask, char *user2, ACCESS_MASK
 	si |= DACL_SECURITY_INFORMATION;
 
 	if (SetFileSecurity(path, si, &sd) == 0) {
-		sprintf(logb, "error setting security for file %s", path);
-		log_err(-1, "secure_file2", logb);
+		log_err(-1, __func__, "failed in SetFileSecurity");
 		if (usid) LocalFree(usid);
 		if (gsid) LocalFree(gsid);
 		(void)free(pdacl);
```
sakshamgarg, line 555: Should we add log here https://github.com/PBSPro/pbspro/blob/6514592d1bca52adb5cd2bdb3233be1f9838a25a/src/lib/Libwin/accesinfo.c#L571

src/lib/Liblog/pbs_log.c
```diff
@@ -724,15 +725,15 @@ log_err(int errnum, const char *routine, const char *text)
 		LPVOID	lpMsgBuf;
 		int	err = GetLastError();
```
npadole20, line 726: Can we change this to DWORD err? Return type of GetLastError() is DWORD.

src/lib/Liblog/pbs_log.c
```diff
@@ -724,15 +725,15 @@ log_err(int errnum, const char *routine, const char *text)
...
 		int		len;
-
+		snprintf(buf, LOG_BUF_SIZE, "GetLastError() returned: %d. ", err);
```
npadole20, line 728: if we change err to DWORD as per above comment, use %lu here.

src/lib/Liblog/pbs_log.c
```diff
@@ -724,15 +725,15 @@ log_err(int errnum, const char *routine, const char *text)
...
 		FormatMessage(
 			FORMAT_MESSAGE_ALLOCATE_BUFFER |
 			FORMAT_MESSAGE_FROM_SYSTEM |
 			FORMAT_MESSAGE_IGNORE_INSERTS,
 			NULL, err,
 			MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT),
 			(LPTSTR)&lpMsgBuf, 0, NULL);
-		strncpy(buf, lpMsgBuf, sizeof(buf));
+		strncat(buf, lpMsgBuf, LOG_BUF_SIZE - (int)strlen(buf) - 1);
```
npadole20, line 736: Why this change?

src/lib/Libwin/dir.c
```diff
@@ -187,6 +193,7 @@ link(const char *oldpath, const char *newpath)
 
 	if (ret != 0) {
 		errno = GetLastError();
```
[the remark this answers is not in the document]
SudeshnaMoulik: No. Because there could be a possibility where the caller check this.

src/lib/Libwin/gethostid.c
```diff
@@ -158,6 +160,7 @@ gethostid(void)
 
 	sid = malloc(sz1);
 	if (sid == NULL) {
+		log_err(errno, __func__, "memory allocation failed for sid");
 		return (ret);
 	}
```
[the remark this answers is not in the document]
SudeshnaMoulik: As per the below link, "If the function fails, the return value is undefined." https://docs.microsoft.com/en-us/windows/win32/api/securitybaseapi/nf-securitybaseapi-getsidsubauthority

src/lib/Libwin/popen.c
```diff
@@ -97,12 +97,14 @@ win_popen(char	*cmd, const char *type, struct pio_handles *pio, struct proc_ctrl
 	if (CreatePipe(&hReadPipe_out, &hWritePipe_out, &sa, 0) == 0) {
 		fprintf(stderr, "\npopen: pipe creation failed!\n");
 		errno = GetLastError();
```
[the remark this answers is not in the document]
SudeshnaMoulik: You are right except 271. Please take a look at pbs_attach.c. Thanks

src/lib/Liblog/pbs_log.c
SudeshnaMoulik: Because we are changing the buffer now before this.

src/lib/Libwin/accesinfo.c
```diff
@@ -80,16 +80,18 @@ create_secure_dacl(char *user, ACCESS_MASK mask, SID *owner_sid)
 	int k = 0;
 	int cbAcl = 0;
 	ACL *ndacl = NULL;
-	char logb[LOG_BUF_SIZE] = {'\0' };
+	char logb[LOG_BUF_SIZE] = {'\0' } ;
```
subhasisb, line 83: why not just use log_buffer instead of defining another variable. Anyways, we are using log.h and calling logging functions from liblog

subhasisb: Also, i thnk this kind of initilization of large buffers is slow. It will basically memset the entire 4096 byte with 0, whereas we probably dont need that

src/lib/Libwin/accesinfo.c
```diff
@@ -80,16 +80,18 @@ create_secure_dacl(char *user, ACCESS_MASK mask, SID *owner_sid)
...
+	if (k <= 0) {
+		log_eventf(PBSEVENT_ERROR, 0, LOG_DEBUG, __func__, "failed to get group-id");
+	}
 
 	if ((k < _MAX_GROUPS) && (owner_sid != NULL)) {
 		grp[k] = sid_dup(owner_sid);
 		if (grp[k] == NULL) {
-			sprintf(logb, "failed to copy owner_sid");
-			log_err(-1, __func__, logb);
+			log_eventf(PBSEVENT_ERROR, 0, LOG_DEBUG, __func__, "failed to copy owner_sid");
```
subhasisb, line 94: Are these logs going to appear only in debug mode? I think we should log ALL errors in all logging levels

src/lib/Libwin/accesinfo.c
```diff
@@ -106,9 +108,10 @@ create_secure_dacl(char *user, ACCESS_MASK mask, SID *owner_sid)
 				grp[k] = sid;
 				k++;
 			}
+		} else {
+			log_eventf(PBSEVENT_ERROR, 0, LOG_DEBUG, __func__, "failed to get group and user sid");
```
subhasisb, line 112: Are these logs going to appear only in debug mode? I think we should log ALL errors in all logging levels

src/lib/Libwin/accesinfo.c
```diff
@@ -80,16 +80,18 @@ create_secure_dacl(char *user, ACCESS_MASK mask, SID *owner_sid)
 	int k = 0;
 	int cbAcl = 0;
 	ACL *ndacl = NULL;
-	char logb[LOG_BUF_SIZE] = {'\0' };
+	char logb[LOG_BUF_SIZE] = {'\0' } ;
 
 	rids[0] = DOMAIN_ALIAS_RID_ADMINS;
 	k = getgids(getlogin(), grp, rids);
+	if (k <= 0) {
+		log_eventf(PBSEVENT_ERROR, 0, LOG_DEBUG, __func__, "failed to get group-id");
```
subhasisb, line 88: also no error code in these function failures?

src/lib/Libwin/accesinfo.c
subhasisb, line 94: also no error code in these function failures?

src/lib/Libwin/accesinfo.c
```diff
@@ -181,25 +185,39 @@ secure_file(char *path, char *user, ACCESS_MASK mask)
...
 
-	if (path == NULL || *path == '\0')
+	if (path == NULL || *path == '\0') {
+		log_eventf(PBSEVENT_ERROR, 0, LOG_DEBUG, __func__, "invalid path");
 		return (0);
+	}
 
-	if (lstat(path, &sbuf) == -1)
+	if (lstat(path, &sbuf) == -1) {
+		log_eventf(PBSEVENT_ERROR, 0, LOG_DEBUG, __func__, "Path don't exists");
```
subhasisb, line 196: We should pass on the errno here

src/lib/Libwin/accesinfo.c
```diff
@@ -181,25 +185,39 @@ secure_file(char *path, char *user, ACCESS_MASK mask)
...
 		return (0);	/* ignore non-existent files! */
+	}
 
-	if (!has_privilege(SE_RESTORE_NAME))
-		ena_privilege(SE_RESTORE_NAME);
+	if (!has_privilege(SE_RESTORE_NAME)) {
+		if (ena_privilege(SE_RESTORE_NAME) == 0) {
+			log_eventf(PBSEVENT_ERROR, 0, LOG_DEBUG, __func__, "Unable to enable privilege: SE_RESTORE_NAME");
```
subhasisb, line 202: No error codesin the previous failure?

src/lib/Libwin/accesinfo.c
```diff
@@ -181,25 +185,39 @@ secure_file(char *path, char *user, ACCESS_MASK mask)
...
+		}
+	}
 
-	if (!has_privilege(SE_TAKE_OWNERSHIP_NAME))
-		ena_privilege(SE_TAKE_OWNERSHIP_NAME);
+	if (!has_privilege(SE_TAKE_OWNERSHIP_NAME)) {
+		if (ena_privilege(SE_TAKE_OWNERSHIP_NAME) == 0) {
+			log_eventf(PBSEVENT_ERROR, 0, LOG_DEBUG, __func__, "Unable to enable privilege: SE_TAKE_OWNERSHIP_NAME");
```
subhasisb, line 208: same as above

src/lib/Libwin/accesinfo.c
```diff
@@ -181,25 +185,39 @@ secure_file(char *path, char *user, ACCESS_MASK mask)
...
+		}
+	}
 
-	if (!has_privilege(SE_SECURITY_NAME))
-		ena_privilege(SE_SECURITY_NAME);
+	if (!has_privilege(SE_SECURITY_NAME)) {
+		if (ena_privilege(SE_SECURITY_NAME) == 0) {
+			log_eventf(PBSEVENT_ERROR, 0, LOG_DEBUG, __func__, "Unable to enable privilege: SE_SECURITY_NAME");
```
subhasisb, line 214: Are these DEBUG only messages - we need error logs printed always

src/lib/Libwin/accesinfo.c
```diff
@@ -251,16 +268,14 @@ secure_file(char *path, char *user, ACCESS_MASK mask)
 
 	pdacl = create_secure_dacl(user, mask, usid);
 	if (pdacl == NULL) {
-		sprintf(logb, "failed to create secure dacl for file %s", path);
-		log_err(-1, "secure_file", logb);
+		log_eventf(PBSEVENT_ERROR, 0, LOG_DEBUG, __func__, "failed to create secure dacl for file %s", path);
```
subhasisb, line 271: no error code from previous call?

src/lib/Liblog/pbs_log.c
```diff
@@ -715,24 +715,25 @@ log_open_main(char *filename, char *directory, int silent)
...
 
 	if (errnum == -1) {
 
 #ifdef WIN32
 		LPVOID	lpMsgBuf;
-		int	err = GetLastError();
+		DWORD	err = GetLastError();
 		int		len;
-
+		snprintf(buf, LOG_BUF_SIZE, "Error code: %lu. ", err);
```
Shrini-h, line 728: Now if the message is long, should we save space by reducing "Error code: XXXXXXX." to something like "Err(XXXXX)" ?

src/lib/Libwin/accesinfo.c
Shrini-h: since this is used for adding formatted string to logerr down somewhere, I wish there was an api "log_errf" similar to "log_eventf", so that we could have completely avoid local log bufferers

hirenvadalia: @Shrini-h I suggest adding log_errf (see this https://github.com/PBSPro/pbspro/pull/1592#discussion_r395842996)...

src/lib/Libwin/passwd.c
```diff
@@ -1989,8 +2096,10 @@ getlogin(void)
 
 		if (OpenThreadToken(GetCurrentThread(), TOKEN_ALL_ACCESS,
 			TRUE, &token)  == 0) {
-			OpenProcessToken(GetCurrentProcess(), TOKEN_ALL_ACCESS,
-				&token);
+			if (OpenProcessToken(GetCurrentProcess(), TOKEN_ALL_ACCESS,
```
Shrini-h, line 2099: since we use getlogin at so many places, could be calling from multiple threads? if so we need to make this function reentrant by removing the static char usern[]

src/lib/Libwin/accesinfo.c
Shrini-h: is it really error to have k = 0, what if there are no group-id

src/lib/Libwin/accesinfo.c
Shrini-h: I think the LOG_DEBUG is for syslog, why dont we have it same as pbs event

src/lib/Libwin/accesinfo.c
Shrini-h: you can use log_err(-1,

src/lib/Libwin/accesinfo.c
Shrini-h: you can use log_err(-1,

src/lib/Libwin/accesinfo.c
Shrini-h: you can use log_err(-1,

src/lib/Libwin/accesinfo.c
```diff
@@ -210,21 +228,21 @@ secure_file(char *path, char *user, ACCESS_MASK mask)
 
 	if (usid) {
 		if (SetSecurityDescriptorOwner(&sd, usid, FALSE) == 0) {
-			sprintf(logb, "error setting owner for file %s", path);
-			log_err(-1, "secure_file", logb);
+			log_err(-1, __func__, "failed in SetSecurityDescriptorOwner");
```
Shrini-h, line 231: wont printing value of "path" help? or will it be printed in other log?

src/lib/Libwin/accesinfo.c
```diff
@@ -210,21 +228,21 @@ secure_file(char *path, char *user, ACCESS_MASK mask)
...
 			LocalFree(usid);
 			return (0);
 		}
 		si |= OWNER_SECURITY_INFORMATION;
 
 		/* to reset perms, trick is to set ownership of file first */
 		if (SetFileSecurity(path, si, &sd) == 0) {
-			sprintf(logb, "error setting actual owner for file %s", path);
-			log_err(-1, "secure_file", logb);
+			log_err(-1, __func__, "failed in SetFileSecurity");
```
Shrini-h, line 239: wont printing value of "path" help? or will it be printed in other log?

src/lib/Libwin/accesinfo.c
```diff
@@ -239,8 +257,7 @@ secure_file(char *path, char *user, ACCESS_MASK mask)
 
 	if (gsid) {
 		if (SetSecurityDescriptorGroup(&sd, gsid, FALSE) == 0) {
-			sprintf(logb, "error setting group for file %s", path);
-			log_err(-1, "secure_file", logb);
+			log_err(-1, __func__, "failed in SetSecurityDescriptorGroup");
```
Shrini-h, line 260: wont printing value of "path" help? or will it be printed in other log?

src/lib/Libwin/passwd.c
```diff
@@ -396,17 +394,21 @@ sid_dup(SID *src_sid)
 	DWORD	sid_len_need;
 	SID	*dest_sid = NULL;
 
-	if ((src_sid == NULL) || (!IsValidSid(src_sid)))
+	if ((src_sid == NULL) || (!IsValidSid(src_sid))) {
+		log_err(-1, __func__, "Invalid sid");
```
Shrini-h, line 398: if src_sid is NULL then log_err will have wrong GetLastError()

src/lib/Libwin/accesinfo.c
SudeshnaMoulik: Thanks @Shrini-h for pointing this out. Will remove this check

src/lib/Libwin/accesinfo.c
SudeshnaMoulik: Will change the log level to LOG_ERR

SudeshnaMoulik: No, because that is already handled inside "sid_dup"

src/lib/Libwin/accesinfo.c
SudeshnaMoulik: If lstat() fails then the corresponding implementation in util.c logs the error code and message.
Do you recommend to remove this log message from here?

src/lib/Libwin/accesinfo.c
SudeshnaMoulik: For both the has_privilege and ena_privilege, corresponding implementation logs the error depending on where it fails. That's why didn't log them again here

src/lib/Libwin/accesinfo.c
SudeshnaMoulik: Logging of error code already handles inside that API

src/lib/Libwin/accesinfo.c
```diff
@@ -496,8 +518,7 @@ secure_file2(char *path, char *user, ACCESS_MASK mask, char *user2, ACCESS_MASK
 	}
 	if (gsid) {
 		if (SetSecurityDescriptorGroup(&sd, gsid, FALSE) == 0) {
-			sprintf(logb, "error setting group for file %s", path);
-			log_err(-1, "secure_file2", logb);
+			log_err(-1, __func__, "failed in SetSecurityDescriptorGroup for %s", path);
```
Shrini-h, line 521: why not print "gname" ?

src/lib/Libwin/accesinfo.c
```diff
@@ -508,16 +529,14 @@ secure_file2(char *path, char *user, ACCESS_MASK mask, char *user2, ACCESS_MASK
 
 	pdacl = create_secure_dacl2(user, mask, user2, mask2, usid);
 	if (pdacl == NULL) {
-		sprintf(logb, "failed to create secure dacl for file %s", path);
-		log_err(-1, "secure_file2", logb);
+		log_eventf(PBSEVENT_ERROR, 0, LOG_ERR, __func__, "failed to create secure dacl for file %s", path);
```
Shrini-h, line 532: why not print user, mask, user2, mask2

src/lib/Libwin/accesinfo.c
```diff
@@ -816,22 +837,27 @@ perm_granted_admin_and_owner(char *path, int disallow, char *owner, char *errmsg
...
+		goto chkerr;
+	}
 
 	allowed = (struct accessinfo *)malloc(sizeof(struct accessinfo)*sizeInfo.AceCount);
 	denied = (struct accessinfo *)malloc(sizeof(struct accessinfo)*sizeInfo.AceCount);
 
 	if (allowed == NULL || denied == NULL) {
 		rc = errno;
 		strcpy(errmsg, "malloc of allowed or denied struct failed!");
+		log_err(errno, __func__, errmsg);
```
Shrini-h, line 860: errno could be wrong if first malloc failed and second one succeeded?

src/lib/Libwin/accesinfo.c
```diff
@@ -1197,8 +1044,7 @@ make_dir_files_everyone_read(char *path)
 		make_dir_files_everyone_read(dirfile);
 	}
 	if (errno != 0 && errno != ENOENT) {
-		sprintf(logb,"readdir error; %s", path);
-		log_err(-1, "make_dir_files_everyone_read", logb);
+		log_eventf(PBSEVENT_ERROR, 0, LOG_ERR, __func__, "readdir error; %s", path);
```
Shrini-h, line 1047: Not a comment, But just want to know where errno is getting set, and what may set it to ENOENT

src/lib/Libwin/dir.c
```diff
@@ -152,7 +156,9 @@ closedir(DIR *hdir)
 		return (-1);
```
Shrini-h, line 156: did you miss checking and loggin for error at line 124/128 for api FindNextFile() ?

src/lib/Libwin/gethostid.c
Shrini-h: Im not able to find the place where error logging of two invocations of LookupAccountName() is happening?

Shrini-h: also should we log error for any failures of 3 invocations of Netbios() ?

bhroam:
I noticed several calls to log_event() passing 0 as the second argument.  It is a required argument for our log_event()/log_eventf().  Please provide an event class.

src/lib/Libwin/accesinfo.c
```diff
@@ -106,9 +104,10 @@ create_secure_dacl(char *user, ACCESS_MASK mask, SID *owner_sid)
 				grp[k] = sid;
 				k++;
 			}
+		} else {
+			log_eventf(PBSEVENT_ERROR, 0, LOG_ERR, __func__, "failed to get group and user sid");
```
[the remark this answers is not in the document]
SudeshnaMoulik: As any couple of other places we log user name, so in my opinion we should add that. As it will add more info in error log

src/lib/Libwin/accesinfo.c
SudeshnaMoulik: Because at this time it will be inaccessible.

src/lib/Libwin/accesinfo.c
SudeshnaMoulik: Except malloc didn't find the place where we are setting the errno

SudeshnaMoulik: Getting set in lstat

src/lib/Libwin/passwd.c
SudeshnaMoulik: I don't recall any place (for Windows) which is multi-threaded. Apart from this if any API fails then as per the code we are setting usern as "".

src/lib/Libwin/gethostid.c
SudeshnaMoulik: actually LookupAccountName error handling path is covered in existing code. And for Netbios if that fails then we assume the id as pbs_server_addr. So don't think we need any extra login there. Let me know your thoughts

src/lib/Libwin/passwd.c
```diff
@@ -5813,6 +5829,9 @@ access_uncpath(char *path, int mode)
 
 	unmap = get_localpath(path_temp_buf, map_drive);
 	ret = _access(path_temp_buf, mode);
+	if (ret != 0) {
+		log_eventf(PBSEVENT_ERROR, 0, LOG_DEBUG, __func__, "file %s don't have access %d", path_temp_buf, mode);
```
Shrini-h, line 5833: you could have used log_err(errno ?

src/lib/Libwin/passwd.c
```diff
@@ -757,7 +777,11 @@ get_dcinfo(char *net_name,
 		if (dctrl) {
 			strncpy(domain_name, dctrl->DomainName, PBS_MAXHOSTNAME);
 			strncpy(domain_ctrl, dctrl->DomainControllerName, PBS_MAXHOSTNAME);
-			NetApiBufferFree(dctrl);
+			nBufferStat = NetApiBufferFree(dctrl);
```
Shrini-h, line 780: for this NetApiBufferFree() I suggest creating wrapper as in https://github.com/PBSPro/pbspro/pull/1727
Since there are many NetApiBufferFree() invocations in this file, with wrapper it will look clean

src/lib/Libwin/passwd.c
```diff
@@ -1550,8 +1599,11 @@ isLocalAdminMember(char *user)
 	mbstowcs(userw, user, PBS_MAXHOSTNAME+UNLEN+1);
 	mbstowcs(gnamew, gname, GNLEN);
 
-	if (NetLocalGroupGetMembers(NULL, gnamew, 2, (LPBYTE *)&members,
-		MAX_PREFERRED_LENGTH, &nread, &totentries, NULL) != NERR_Success) {
+	status = NetLocalGroupGetMembers(NULL, gnamew, 2, (LPBYTE *)&members,
+				MAX_PREFERRED_LENGTH, &nread, &totentries, NULL);
+	if (status != NERR_Success) {
+		log_eventf(PBSEVENT_ERROR, 0, LOG_DEBUG, __func__, "failed in NetLocalGroupGetMembers, with errno %d",
```
Shrini-h, line 1605: You might want to replace LOG_DEBUG with LOG_ERR and use server eventclass for the second param, for all such changes through out this PR.

src/lib/Libwin/popen.c
```diff
@@ -218,7 +223,7 @@ win_popen(char	*cmd, const char *type, struct pio_handles *pio, struct proc_ctrl
 		return (1);
 	} else {
 		fprintf(stderr, "\npopen: CreateProcess() failed!\n");
-		errno = GetLastError();
+		log_err(-1, __func__, "failed in CreateProcess");
```
Shrini-h, line 226: the log_err(-1,) should be called immediately after failure. so please move this log to immediately after CreateProcess with if (!fsuccess) block

src/lib/Libwin/popen.c
```diff
@@ -218,7 +223,7 @@ win_popen(char	*cmd, const char *type, struct pio_handles *pio, struct proc_ctrl
 		return (1);
```
Shrini-h, line 223: did you miss loggin error for WaitForSingleObject at line 212/217 ?

src/lib/Libwin/popen.c
```diff
@@ -342,15 +353,19 @@ void
 win_pclose(pio_handles *pio)
 {
 	DWORD rc = 0;
+	int ret = 0;
 	if (pio) {
 		win_pclose2(pio);
 		if ((pio->pi.hProcess != NULL) && (pio->pi.hProcess != INVALID_HANDLE_VALUE)) {
 			/* Terminate the jobs object or the process only if it is not already exited */
 			if ((GetExitCodeProcess(pio->pi.hProcess, &rc)) && (rc == STILL_ACTIVE)) {
```
Shrini-h, line 361: Should we log failure of GetExitCodeProcess() ?

src/lib/Libwin/passwd.c
```diff
@@ -4132,17 +4097,24 @@ wsystem(char *cmdline, HANDLE user_handle)
...
+	}
 
 	/* restore current working directory */
 	if (changed_dir)
 		chdir(current_dir);
 
 	if (rc) { /* CreateProcess* successful */
+		stat = WaitForSingleObject(pi.hProcess, INFINITE);
+		if (stat != WAIT_OBJECT_0) {
+			log_eventf(PBSEVENT_ERROR, 0, LOG_DEBUG, __func__, "failed in WaitForSingleObject, with errno %d", stat);
```
Shrini-h, line 4111: should we use log_err(-1,) when stat == WAIT_FAILED. from https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-waitforsingleobject

> 

WAIT_FAILED(DWORD)0xFFFFFFFF | The function has failed. To get extended error information, call GetLastError.
-- | --

src/lib/Libwin/winnet.c
```diff
@@ -54,15 +54,14 @@
...
 	WSADATA	data;
 
 	save_env(); 	/* need certain environment variables set in order */
 	/* make network calls like socket() or gethostbyname()*/
 
 	if (WSAStartup(MAKEWORD(2, 2), &data)) {
-		sprintf(logb,"winsock_init failed! error=%d", WSAGetLastError());
-		log_err(-1, "winsock_init", logb);
+		log_eventf(PBSEVENT_ERROR, 0, LOG_DEBUG, __func__, "winsock_init failed with errno %d",
+				WSAGetLastError());
```
Shrini-h, line 64: please print only the return code of WSAStartup. from https://docs.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-wsastartup

> The WSAStartup function directly returns the extended error code in the return value for this function. A call to the WSAGetLastError function is not needed and should not be used.

src/lib/Libwin/passwd.c
```diff
@@ -649,15 +663,17 @@ GetComputerDomainName(char domain_name[PBS_MAXHOSTNAME+1])
...
+	}
 	strncpy(domain_name, local_name, PBS_MAXHOSTNAME);
 
 	ZeroMemory(&obj_attrs, sizeof(obj_attrs));
-	if (LsaOpenPolicy(NULL, &obj_attrs, POLICY_VIEW_LOCAL_INFORMATION,
-	&h_policy) \
-                                                        != ERROR_SUCCESS )
+	lsa_stat = LsaOpenPolicy(NULL, &obj_attrs, POLICY_VIEW_LOCAL_INFORMATION, &h_policy);
+	if ( lsa_stat != ERROR_SUCCESS ) {
+		log_eventf(PBSEVENT_ERROR, 0, LOG_DEBUG, __func__, "failed in LsaOpenPolicy, with errno %d", LsaNtStatusToWinError(lsa_stat));
```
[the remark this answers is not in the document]
SudeshnaMoulik: My opinion will be to log the actual LsaNtStatus error, because 

> The return value is the Windows error code that corresponds to the Status parameter. If there is no corresponding Windows error code, the return value is ERROR_MR_MID_NOT_FOUND.

I thought of converting it to Windows error code and log the actual LsaNtStatus error code too, but thought that may confuse the user as there will be 2 error code.

src/lib/Libwin/passwd.c
```diff
@@ -2053,8 +2164,10 @@ getlogin_full(void)
```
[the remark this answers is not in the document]
SudeshnaMoulik: If we are using getusername() then error handling is already handled inside that. If we are using GetUserName() and that return 0, then the following code path handles the error.

src/lib/Libwin/passwd.c
```diff
@@ -2053,8 +2164,10 @@ getlogin_full(void)
...
 			TRUE, &token)  == 0) {
-			OpenProcessToken(GetCurrentProcess(), TOKEN_ALL_ACCESS,
-				&token);
+			if (OpenProcessToken(GetCurrentProcess(), TOKEN_ALL_ACCESS,
+					&token) == 0) {
+				log_err(-1, __func__, "failed in OpenProcessToken");
+			}
 		}
 		if (token == INVALID_HANDLE_VALUE) {
 			strcpy(usern, "");
```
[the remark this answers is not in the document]
SudeshnaMoulik: Actually there is already error path to follow there

src/lib/Libwin/passwd.c
```diff
@@ -2414,12 +2515,16 @@ default_local_homedir(char *username, HANDLE usertoken, int ret_profile_path)
 	profilepath[0] = '\0';
 	if (GetUserProfileDirectory(userlogin, profilepath,
 		&profsz) == 0) {
-		if (LoadUserProfile(userlogin, &pi) == 0)
+		if (LoadUserProfile(userlogin, &pi) == 0) {
```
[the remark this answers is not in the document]
SudeshnaMoulik: Is this comment for LoadUserProfile ?

src/lib/Libwin/passwd.c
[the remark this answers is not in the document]
Shrini-h: thanks, I think printing LsaNtStatus will make it more clear, rather than converting.

src/lib/Libwin/passwd.c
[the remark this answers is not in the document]
Shrini-h: yes error path is there in the else, but I'm talking about the 3 lines in the if block, is it not dead code. also can the api fail with reasons other than insuffecient buffer. now i also notice that the malloc is not error logged.

src/lib/Libwin/passwd.c
```diff
@@ -2508,10 +2617,9 @@ map_unc_path(char *path, struct passwd *pw)
 		(void)free(nr.lpRemoteName);
 
 	if (ret != 0) { /* unc path failed to map */
-
-		strncpy(local_drive,
-			default_local_homedir(pw->pw_name, pw->pw_userlogin, 0),
-			MAXPATHLEN+1);
+		tmp = default_local_homedir(pw->pw_name, pw->pw_userlogin, 0);
```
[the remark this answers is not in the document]
SudeshnaMoulik: I think no, because if WNetUseConnection fails then it fallback to default_local_homedir (which handles its own error handling)

src/lib/Libwin/passwd.c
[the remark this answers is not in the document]
Shrini-h: no. its for GetUserProfileDirectory()

src/lib/Libwin/passwd.c
```diff
@@ -2631,8 +2747,16 @@ getAssignedHomeDirectory(char *user)
 				homedir = (char *)malloc(len+1);
 
 				if (homedir == NULL) {
-					NetApiBufferFree(uinfo);
-					if (dcw) NetApiBufferFree(dcw);
+					stat = NetApiBufferFree(uinfo);
```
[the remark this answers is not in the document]
SudeshnaMoulik: nope, its there!!
https://github.com/PBSPro/pbspro/blob/9d07e02a5de32e4e4484d62bf3980578b2a1b116/src/lib/Libwin/passwd.c#L2751

src/lib/Libwin/passwd.c
```diff
@@ -2801,6 +2903,7 @@ has_privilege(char *privname)
 	}
 
 	if (!AdjustTokenPrivileges(procToken, FALSE, toke, 0, NULL, 0)) {
+		log_err(-1, __func__, "failed in AdjustTokenPrivileges");
```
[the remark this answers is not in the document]
SudeshnaMoulik: Didn't understand the question. Are you asking about the error logging ? or code?

src/lib/Libwin/passwd.c
```diff
@@ -3046,13 +3157,17 @@ create_token_privs_byuser(SID *usid, DWORD attrib, HANDLE hLsa)
...
 
 	if (hLsa == INVALID_HANDLE_VALUE || usid == NULL)
 		return NULL;
 
 	/* get user rights and add to list of user token privileges */
-	LsaEnumerateAccountRights(hLsa, usid, &lsaRights, &numRights);
-
+	lsa_stat = LsaEnumerateAccountRights(hLsa, usid, &lsaRights, &numRights);
+	if (lsa_stat != ERROR_SUCCESS) {
+		log_eventf(PBSEVENT_ERROR, 0, LOG_DEBUG, __func__, "failed in LsaEnumerateAccountRights, with errno %d",
```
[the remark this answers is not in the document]
SudeshnaMoulik: Same reason.

> The return value is the Windows error code that corresponds to the Status parameter. If there is no corresponding Windows error code, the return value is ERROR_MR_MID_NOT_FOUND.
https://docs.microsoft.com/en-us/windows/win32/api/ntsecapi/nf-ntsecapi-lsantstatustowinerror

Wanted to log the LsaNtStatus error code, even for the case when there is no corresponding Windows error code

src/lib/Libwin/passwd.c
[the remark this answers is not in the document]
Shrini-h: I think the comment was for the malloc 3 lines above. please check

src/lib/Libwin/passwd.c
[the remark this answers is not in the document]
Shrini-h: This is about the code actually. just trying to understand.

src/lib/Libwin/passwd.c
```diff
@@ -3086,6 +3203,7 @@ create_token_privs_bygroups(TOKEN_GROUPS *token_groups, DWORD attrib, HANDLE hLs
 	char privname[GNLEN+1] = {'\0'};
 	LUID	luid;
 	BOOL	found_match;
+	NTSTATUS lsa_stat;
 
 	if (hLsa == INVALID_HANDLE_VALUE)
 		return NULL;
```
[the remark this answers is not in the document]
SudeshnaMoulik: for point 2: didn't add because if for one token groups it is not successful then it will  continue to the next (so its not a error)

src/lib/Libwin/passwd.c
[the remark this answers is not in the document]
Shrini-h: So will it be better to directly print the error code without conversion ?

src/lib/Libwin/passwd.c
```diff
@@ -3454,15 +3526,15 @@ get_token_info(HANDLE token,
...
+			}
 
 	}
 
-	if (GetTokenInformation(token, TokenUser, buf, cb, &cb)) {
-		log_err(-1, "get_token_info", "User: 1st GetTokenInformation failed!");
-		return;
-	} else {
+	if ( !GetTokenInformation(token, TokenUser, buf, cb, &cb)) {
+		/* User: 1st GetTokenInformation failed! */
```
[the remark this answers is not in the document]
SudeshnaMoulik: Because if that fails, we are allocating the buffer and trying GetTokenInformation again, if still that fails then we are logging the corresponding error details

src/lib/Libwin/passwd.c
```diff
@@ -3476,15 +3548,12 @@ get_token_info(HANDLE token,
...
+			log_err(-1, __func__, "failed in GetTokenInformation");
 		}
 	}
 	cb = 0;
-	if (GetTokenInformation(token, TokenOwner, buf, cb, &cb)) {
-		log_err(-1, "get_token_info", "Owner: 1st GetTokenInformation failed!");
-	} else {
-
+	if ( !GetTokenInformation(token, TokenOwner, buf, cb, &cb)) {
+		/* Owner: 1st GetTokenInformation failed! */
```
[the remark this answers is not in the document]
SudeshnaMoulik: Same as above

src/lib/Libwin/passwd.c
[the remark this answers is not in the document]
Shrini-h: I understand that the first call is to query the required buffer lenght. but cant the first GetTokenInformation() fail for reason other than insufficient buffer size?

src/lib/Libwin/passwd.c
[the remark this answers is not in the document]
Shrini-h: I understand that the first call is to query the required buffer lenght. but cant the first GetTokenInformation() fail for reason other than insufficient buffer size?

src/lib/Libwin/passwd.c
```diff
@@ -3909,127 +3963,33 @@ LogonUserNoPass(char *user)
...
-			setuser_hdle = pwdp->pw_userlogin;
-			return (impersonate_user(setuser_hdle));
-		}
-	}
-
-	usid = getusersid2(user, realname);
-	if (usid == NULL) {
-		return (0);
+	if (setuser_hdle == 0) {
+		log_err(-1, __func__, "failed in LogonUserNoPass");
```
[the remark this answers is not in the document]
SudeshnaMoulik: As per https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-logonusera
if the api fails then GetLastError gives the error code, which is what we are doing in log_err !!

Why do you think this is a bug ? if "setuser_hdle == 0" then it will ultimately return setuser_hdle

src/lib/Libwin/passwd.c
```diff
@@ -3909,127 +3963,33 @@ LogonUserNoPass(char *user)
...
 		}
-		memset(thepass, 0, cred_len);
-
-		return (impersonate_user(setuser_hdle));
 	}
 
-	/* no password */
-
-	return (setuser(user));
+	return setuser_hdle;
```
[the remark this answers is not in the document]
SudeshnaMoulik: Good catch

src/lib/Libwin/passwd.c
```diff
@@ -4179,27 +4151,26 @@ add_window_station_ace(HWINSTA hwin, SID *usid)
...
 	if (GetUserObjectSecurity(hwin, &si, sd, sd_sz, &sd_sz_need) == 0) {
-
 		if (GetLastError() == ERROR_INSUFFICIENT_BUFFER) {
 			if ((sd=(SECURITY_DESCRIPTOR *)malloc(sd_sz_need)) == NULL) {
-				errno = GetLastError();
+				log_err(errno, __func__, "memory allocation failed for GetUserObjectSecurity");
 				goto fail;
 			}
 			memset((SECURITY_DESCRIPTOR *)sd, 0, sd_sz_need);
 		}
```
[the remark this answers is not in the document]
SudeshnaMoulik: This probably the older version of code. Have corrected it in recent version.

src/lib/Libwin/passwd.c
```diff
@@ -4930,11 +4890,16 @@ logon_pw(char *username,
 		/* check if userlogin handle is stale */
 		if (pwdp->pw_userlogin != INVALID_HANDLE_VALUE) {
 			if (!impersonate_user(pwdp->pw_userlogin)) {
-
 				CloseHandle(pwdp->pw_userlogin);
```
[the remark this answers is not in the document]
SudeshnaMoulik: I don't think so, because it will try to get the new user handle.

src/lib/Libwin/passwd.c
```diff
@@ -4930,11 +4890,16 @@ logon_pw(char *username,
...
 				pwdp->pw_userlogin = LogonUserNoPass(username);
+				if (pwdp->pw_userlogin == 0) {
+					log_err(-1, __func__, "failed in LogonUserNoPass");
```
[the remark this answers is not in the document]
SudeshnaMoulik: yes, https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-logonusera

src/lib/Libwin/passwd.c
```diff
@@ -5697,6 +5711,7 @@ get_processowner(DWORD processid, uid_t *puid, char *puname, size_t uname_len, c
 		*puid = hUser->User.Sid;
 
 	if (!LookupAccountSid(0, hUser->User.Sid, username, &username_len, domainname, &domainname_len, (PSID_NAME_USE)&sid_buf)) {
+		log_err(-1, __func__, "failed in LookupAccountSid");
```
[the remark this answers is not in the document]
SudeshnaMoulik: if the call fails how will we get the values!

src/lib/Libwin/passwd.c
SudeshnaMoulik: Really like the idea, it would help much.
But as per latest team discussion, it probably not a good idea to add code no and this api is mainly used only in passwd.c (other Libwin files don't use this).
Probably we can think of this is future refactoring

src/lib/Libwin/passwd.c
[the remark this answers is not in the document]
Shrini-h: But here the function LogonUserNoPass() used to set the "setuser_hdle" is not a windows api? more over looking at the implementation of LogonUserNoPass(), it is returning 0 at some place without setting last error. so I believe we should not call log_err(-1,), which might lead to log_err using a stale GetLastError() value which might be erroneous.

Sorry on the false BUG alarm! I failed to look down the lines 😄

src/lib/Libwin/passwd.c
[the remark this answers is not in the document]
Shrini-h: I mentioned "LogonUserNoPass" in the comment but the link you replied is of "LogonUserA". So please check.

Shrini-h:
@SudeshnaMoulik  a request. It would be easy for me to verify the next round of review if you dont mark the comments as resolved ex parte. 😃 
Thank You

SudeshnaMoulik:
Will add the winlog_buffer as that is needed by resmom and pbs_account

src/lib/Libwin/passwd.c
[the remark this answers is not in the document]
Shrini-h: Sorry, copy paste error, i meant the process id

src/lib/Libwin/accesinfo.c
```diff
@@ -128,7 +128,6 @@ create_secure_dacl(char *user, ACCESS_MASK mask, SID *owner_sid)
 	}
 	if (InitializeAcl(ndacl, cbAcl, ACL_REVISION) == 0) {
 		log_err(-1, __func__, "failed in InitializeAcl");
-		return NULL;
```
subhasisb, line 129: removed return?

src/lib/Libwin/accesinfo.c
```diff
@@ -396,7 +395,6 @@ create_secure_dacl2(char *user, ACCESS_MASK mask, char *user2, ACCESS_MASK mask2
...
```
subhasisb, line 384: removed return?

src/lib/Libwin/accesinfo.c
SudeshnaMoulik: Ya @subhasisb, added that by mistake

src/lib/Libwin/accesinfo.c
SudeshnaMoulik: Yes, added that by mistake. Actual code https://github.com/PBSPro/pbspro/blob/master/src/lib/Libwin/accesinfo.c#L386

subhasisb: ohk got it

src/lib/Libwin/accesinfo.c
subhasisb: got it thanks

src/lib/Libwin/passwd.c
```diff
@@ -4551,25 +4604,30 @@ use_window_station_desktop(SID *usid)
...
 
 	if (add_desktop_ace(hdesk, usid)) {
+		log_eventf(PBSEVENT_ERROR, PBS_EVENTCLASS_SERVER, LOG_ERR, __func__, "failed in add_desktop_ace");
 		goto end;
 	}
 
 	ret = 0;
 
-	end:  	if (hwin)
+	end:
```
Shrini-h, line 4629: make the label start from 0 column

src/lib/Libwin/passwd.c
```diff
@@ -5256,8 +5292,8 @@ wrap_NetUserGetInfo(LPCWSTR servername,
 			}
 		}
 		if (!found) {
-			sprintf(winlog_buffer, "No user token found for %s",
-				user_name);
+			sprintf(winlog_buffer, "No user token found for %s", user_name);
+			log_event(PBSEVENT_ERROR, PBS_EVENTCLASS_SERVER, LOG_DEBUG, __func__, winlog_buffer);
```
Shrini-h, line 5296: do you want to change the syslog param also to LOG_ERR

src/lib/Libwin/popen.c
```diff
@@ -79,7 +79,8 @@ win_popen(char	*cmd, const char *type, struct pio_handles *pio, struct proc_ctrl
 	char                    cmd_shell[MAX_PATH + 1] = {'\0'};       /* path to cmd shell */
 	char	                current_dir[MAX_PATH + 1] = {'\0'};
 	char	                *temp_dir = NULL;
-	int	                changed_dir = 0;
+	int                     changed_dir = 0;
+	int						return_code = 0;
```
Shrini-h, line 83: indentation is off

src/lib/Libwin/passwd.c
Shrini-h: the end: label is still not begining from colun 0

Shrini-h:
LGTM,  except one last pending cosmetic comment.  https://github.com/PBSPro/pbspro/pull/1738#discussion_r429059723

SudeshnaMoulik:
Thanks @Shrini-h for the review. Addressed it in latest commit [4cc2d90](https://github.com/PBSPro/pbspro/pull/1738/commits/4cc2d90998a3d94b35e519493bd105e15b7d1c40)

PR #1739, opened 2020-05-13
Change job submission user to pbsroot

RKORavi:
#### Describe Bug or Feature
To run cat command on job's output file user should have sudo privileges

#### Describe Your Change
Changing job submission user to pbsroot

### DIFF
--- a/test/tests/functional/pbs_nonprint_characters.py
+++ b/test/tests/functional/pbs_nonprint_characters.py
@@ -113,7 +113,7 @@ sleep 5
         self.pbsnodes_cmd = os.path.join(self.server.pbs_conf['PBS_EXEC'],
                                          'bin', 'pbsnodes')
 
-    def create_and_submit_job(self, user=None, attribs=None, content=None,
+    def create_and_submit_job(self, attribs=None, content=None,
                               content_interactive=None, preserve_env=False,
                               set_env={}):
         """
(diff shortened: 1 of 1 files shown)


RKORavi:
Test logs:
[1592-ALL.zip](https://github.com/PBSPro/pbspro/files/4621705/1592-ALL.zip)

bhroam:
@RKORavi I don't understand why the change to PBSROOT_USER over TEST_USER.  Your PR description says you need sudo privileges, but pbsroot is just another user.  It isn't root.  The benefit of using TEST_USER is it is the default user for jobs.

RKORavi:
> @RKORavi I don't understand why the change to PBSROOT_USER over TEST_USER. Your PR description says you need sudo privileges, but pbsroot is just another user. It isn't root. The benefit of using TEST_USER is it is the default user for jobs.

Hi @bhroam,
To run cat command on job's output file sudo privileges is required. On some setup TEST_USER do not have sudo privileges this results in failure of test cases.

bhroam:
@RKORavi I still don't understand.  When you do a self.du.cat(), it is run by the person who is running the test.  In our testbed case, that will be pbsroot.  If you need cat() to run as a different user, user the runas argument.

RKORavi:
> @RKORavi I still don't understand. When you do a self.du.cat(), it is run by the person who is running the test. In our testbed case, that will be pbsroot. If you need cat() to run as a different user, user the runas argument.

@bhroam did not think of that. Thankyou I will make the changes.

RKORavi:
> > @RKORavi I still don't understand. When you do a self.du.cat(), it is run by the person who is running the test. In our testbed case, that will be pbsroot. If you need cat() to run as a different user, user the runas argument.
> 
> @bhroam did not think of that. Thankyou I will make the changes.

@bhroam our testbed machines seems come permission issue. So, this PR might not be required. I will update on this PR after confirming.

hirenvadalia:
Closing long time inactive PR.

PR #1743, opened 2020-05-14
PTL is not updating its internal structures after qmgr set server operation

nithinj:
#### Describe Bug or Feature
PTL internally maintains the old state even after updating certain attributes such as scheduling using qmgr operation. As a consequence of this expect routine will trigger the scheduling cycle even if the last operation sets the scheduling to false.

#### Describe Your Change

Set the hostname as the id of the server object.

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -6866,6 +6866,9 @@ class Server(PBSService):
                             break
                     if rc == 0:
                         rc = tmprc
+
+        if id is None and obj_type == SERVER:
+            id = self.hostname
         bs_list = []
         if cmd == MGR_CMD_DELETE and oid is not None and rc == 0:
             for i in oid:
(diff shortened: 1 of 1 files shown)


nithinj:
That's the consequence of this bug. Expect often performs on job status, not on server status.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -6866,6 +6866,9 @@ def manager(self, cmd, obj_type, attrib=None, id=None, extend=None,
                             break
                     if rc == 0:
                         rc = tmprc
+
+        if id is None and obj_type == SERVER:
+            id = self.hostname
```
agrawalravi90, line 6871: should it be pbs_conf["PBS_SERVER"] instead of self.hostname?

nithinj:
@agrawalravi90 ,
I have addressed your comment. please have a look.

PR #1747, opened 2020-05-18
Jobs overlap with resvs with buckets and psets

bhroam:
#### Describe Bug or Feature
When using the node bucket code path with placement sets enabled, it is possible for a job to run and interfere with a reservation.

#### Describe Your Change
The change is easy, but requires a little bit of explanation of the how the node buckets work.
1. Nodes know all of the run events that affect that node
2. When creating a bucket, we create 3 pools of nodes: busy, busy later (free now, but busy later), and totally free.  We create the busy later pool by looking at the events on the nodes.  If there is one, we add the node to the busy later pool.
3. Each placement set has its own set of buckets

The issue is we create the placement sets before we add the reservation events on the nodes.  This means all nodes with reservations on them are added to the totally free pool.  Since they are in this pool, the bucket path doesn't bother checking for the reservations when running jobs.

The fix is to add the events to the nodes before creating the placement sets.

[resv_overlap.log](https://github.com/PBSPro/pbspro/files/4647312/resv_overlap.log)

### DIFF
--- a/src/scheduler/fifo.c
+++ b/src/scheduler/fifo.c
@@ -1680,13 +1680,6 @@ run_update_resresv(status *policy, int pbs_sd, server_info *sinfo,
 			int sort_nodepart = 0;
 			for (i = 0; ns[i] != NULL; i++) {
 				int j;
-				te_list *te;
-				for (te = ns[i]->ninfo->node_events; te != NULL; te = te->next) {
-					/*
-					log_eventf(PBSEVENT_DEBUG, PBS_EVENTCLASS_NODE, LOG_DEBUG, resresv->name,
-						   "Node: %s Event %s, time %ld", ns[i]->ninfo->name, te->event->name, te->event->event_time);
-					*/
-				}
 				update_node_on_run(ns[i], rr, &old_state);
 				if (ns[i]->ninfo->np_arr != NULL) {
 					for (j = 0; ns[i]->ninfo->np_arr[j] != NULL; j++) {
--- a/src/scheduler/fifo.c
+++ b/src/scheduler/fifo.c
@@ -1680,6 +1680,13 @@ run_update_resresv(status *policy, int pbs_sd, server_info *sinfo,
 			int sort_nodepart = 0;
 			for (i = 0; ns[i] != NULL; i++) {
 				int j;
+				te_list *te;
+				for (te = ns[i]->ninfo->node_events; te != NULL; te = te->next) {
+					/*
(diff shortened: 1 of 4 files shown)


src/scheduler/fifo.c
```diff
@@ -1680,6 +1680,13 @@ run_update_resresv(status *policy, int pbs_sd, server_info *sinfo,
 			int sort_nodepart = 0;
 			for (i = 0; ns[i] != NULL; i++) {
 				int j;
+				te_list *te;
+				for (te = ns[i]->ninfo->node_events; te != NULL; te = te->next) {
+					/*
+					log_eventf(PBSEVENT_DEBUG, PBS_EVENTCLASS_NODE, LOG_DEBUG, resresv->name,
+						   "Node: %s Event %s, time %ld", ns[i]->ninfo->name, te->event->name, te->event->event_time);
+					*/
+				}
```
vstumpf, line 1689: Did you mean to leave this here?

test/tests/functional/pbs_node_buckets.py
```diff
@@ -663,6 +663,43 @@ def test_psets_calendaring(self):
...
+                                                  'node_group_enable': True})
+
+        now = int(time.time())
+        a = {'Resource_List.select': '10010:ncpus=1',
+             'Resource_List.place': 'scatter:excl',
+             'reserve_start': now + 600, 'reserve_end': now + 3600}
+        r = Reservation(attrs=a)
+        rid = self.server.submit(r)
+        self.server.expect(RESV, {'reserve_state':
+                                  (MATCH_RE, 'RESV_CONFIRME|2')}, id=rid)
```
vstumpf, line 684: Change it to `RESV_CONFIRMED`

test/tests/functional/pbs_node_buckets.py
```diff
@@ -663,6 +663,43 @@ def test_psets_calendaring(self):
...
+        self.server.expect(JOB, 'estimated.exec_vnode', id=jid, op=SET)
+
+        n = self.server.status(NODE, 'resources_available.shape')
+        st = self.server.status(JOB, 'estimated.exec_vnode', id=jid)[0]
+        nodes = j.get_vnodes(st['estimated.exec_vnode'])
+
+        s = [x['resources_available.shape']
+             for x in n if x['id'] in nodes]
+        self.assertEqual(len(set(s)), 1,
+                         "Job1 will run in more than one placement set")
```
vstumpf, line 701: Only one job, just say `Job will run...`

PR #1749, opened 2020-05-19
Test failures on slow machines

arungrover:
#### Describe Bug or Feature
There are a bunch of test failures on slow machines and this PR addresses them
1 - test_limits_queues(Multi-sched test suite) is failing on slow machines. It is unable to find a log message in scheduler logs. The log message is logged in the same second log_match function decides to bail out because it reached max_attempts.

2 - test_advance_resv_in_multi_sched fails on slow machines. There has been a case that by the time the test does pbs_rstat and by the time it receives a reply more than 10 seconds pass and the reservation ends because it was only 10 seconds long.

3 - On some slow machines est_advance_confirmed_resv_reconfirm test case fails. It submits a reservation 15 seconds in the future and then sets server reserve_retry_time to 5 seconds. It then offlines the node where this reservation was confirmed. After this, the test issues pbs_rstat to see if the reservation is in a DEGRADED state. This check fails because by the time pbs_rstat command returns  5 seconds passes and the reservation gets confirmed on a different node.

4 - test_alarm_more_than_freq test fails on slow machines. The test expects a hook run and hook end message to be log between 10 to 11 seconds. On slow machines hook itself sometimes takes more than 10 seconds to run which means the hook end message is logged at a later time which makes the test fail.

5 - test_admin_resume_loop tests that a job is admin-suspended and resumed 15 times and checks that the node correctly moves in and out of maintenance state. On a slow machine, the job ends before the test could loop 15 times.

#### Describe Your Change
Respective solution - 
1 - Removed max_attemps from log_match function in test_limits_queues test.
2 - Increased the duration of the reservation.
3 - Fixed by stopping scheduling before offlining the node and start it back again after confirming the reservation's state.
4 - The issue is fixed by relaxing the test to accommodate for slow machines and expect the log to be logged anywhere between the hook sleep time to hook alarm time.
5 - Increased sleep time of the job.

#### Link to Design Doc
NA

[test_svr_periodic_hook.txt](https://github.com/PBSPro/pbspro/files/4647580/test_svr_periodic_hook.txt)
[test_multi_sched.txt](https://github.com/PBSPro/pbspro/files/4647581/test_multi_sched.txt)
[test_admin_suspend.txt](https://github.com/PBSPro/pbspro/files/4647582/test_admin_suspend.txt)

### DIFF
--- a/test/tests/functional/pbs_server_periodic_hook.py
+++ b/test/tests/functional/pbs_server_periodic_hook.py
@@ -110,7 +110,7 @@ pbs.logmsg(pbs.LOG_DEBUG, "periodic hook ended at %%d" %% time.time())
                 time_logged = self.get_timestamp(msg[1])
                 if alarm != 0:
                         self.assertLessEqual(time_logged - time_expected,
-                                        alarm - hook_run_time)
+                                             alarm - hook_run_time)
                 else:
                         self.assertLessEqual(time_logged - time_expected, 1)
 
--- a/test/tests/functional/pbs_server_periodic_hook.py
+++ b/test/tests/functional/pbs_server_periodic_hook.py
@@ -109,10 +109,10 @@ pbs.logmsg(pbs.LOG_DEBUG, "periodic hook ended at %%d" %% time.time())
                                             starttime=search_after)
                 time_logged = self.get_timestamp(msg[1])
                 if alarm != 0:
-                        self.assertLess((time_logged - time_expected),
-                                        (alarm - hook_run_time))
+                        self.assertLess(time_logged - time_expected,
+                                        alarm - hook_run_time)
                 else:
-                        self.assertFalse((time_logged - time_expected) > 1)
+                        self.assertLess(time_logged - time_expected, 1)
 
                 if hook_run_time <= freq:
                     intr = freq - hook_run_time
--- a/test/tests/functional/pbs_server_periodic_hook.py
+++ b/test/tests/functional/pbs_server_periodic_hook.py
@@ -109,10 +109,10 @@ pbs.logmsg(pbs.LOG_DEBUG, "periodic hook ended at %%d" %% time.time())
                                             starttime=search_after)
                 time_logged = self.get_timestamp(msg[1])
                 if alarm != 0:
-                        self.assertLess(time_logged - time_expected,
+                        self.assertLessEqual(time_logged - time_expected,
                                         alarm - hook_run_time)
                 else:
-                        self.assertLess(time_logged - time_expected, 1)
+                        self.assertLessEqual(time_logged - time_expected, 1)
 
                 if hook_run_time <= freq:
                     intr = freq - hook_run_time
--- a/test/tests/functional/pbs_server_periodic_hook.py
+++ b/test/tests/functional/pbs_server_periodic_hook.py
@@ -53,8 +53,8 @@ pbs.logmsg(pbs.LOG_DEBUG, "periodic hook ended at %%d" %% time.time())
         """
         function to create a hook script.
         It accepts 2 arguments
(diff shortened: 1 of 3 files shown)


test/tests/functional/pbs_server_periodic_hook.py
```diff
@@ -106,7 +108,11 @@ def check_next_occurances(self, count, freq,
                                             interval=(hook_run_time + 1),
                                             starttime=search_after)
                 time_logged = self.get_timestamp(msg[1])
-                self.assertFalse((time_logged - time_expected) > 1)
+                if alarm != 0:
+                        self.assertLess((time_logged - time_expected),
+                                        (alarm - hook_run_time))
+                else:
+                        self.assertFalse((time_logged - time_expected) > 1)
```
vstumpf, line 115: Can you change this to an assertLess?

test/tests/functional/pbs_server_periodic_hook.py
```diff
@@ -106,7 +108,11 @@ def check_next_occurances(self, count, freq,
                                             interval=(hook_run_time + 1),
                                             starttime=search_after)
                 time_logged = self.get_timestamp(msg[1])
-                self.assertFalse((time_logged - time_expected) > 1)
+                if alarm != 0:
+                        self.assertLess((time_logged - time_expected),
+                                        (alarm - hook_run_time))
```
bhroam, line 113: I don't think the two extra set of () are needed.  They are two different arguments into the assert function.  There can't be any confusion on what is going on.

test/tests/functional/pbs_server_periodic_hook.py
```diff
@@ -70,17 +70,19 @@ def get_timestamp(self, msg):
...
-        - check_for_hook_end	If it is true then the functions checks for
-                                hook end messages.
+        - count                     to know how many times to repeat
+                                    checking these logs
+        - freq                      is the frequency set in pbs server to run
+                                    this hook
+        - hook_run_time             is the amount of time hook takes to run.
+        - check_for_hook_end        If it is true then the functions checks for
+                                    hook end messages.
+        - alarm                     hook alarm
```
bhroam, line 85: If this is the hook alarm, don't you need to set it via a manager() call?

arungrover: The alarm is set by the caller while creating and importing the hook.

test/tests/functional/pbs_server_periodic_hook.py
```diff
@@ -106,7 +108,11 @@ def check_next_occurances(self, count, freq,
                                             interval=(hook_run_time + 1),
                                             starttime=search_after)
                 time_logged = self.get_timestamp(msg[1])
-                self.assertFalse((time_logged - time_expected) > 1)
+                if alarm != 0:
+                        self.assertLess(time_logged - time_expected,
+                                        alarm - hook_run_time)
+                else:
+                        self.assertLess(time_logged - time_expected, 1)
```
vstumpf, line 115: Oops, sorry, should this be assertLessEqual? Just to keep the original assert the same? Or should it actually be assertLess.

arungrover: You are right, since get_timestamp returns an integer, in some cases it could be 1 second apart and that should be acceptable. I guess I should do the same for the other assert as well.

bhroam: Is this a case where we should use the fuzzy assert?  I can't remember the name of it off the top of my head, but you provide X, Y, and a delta where X and Y should be equal within the delta?

arungrover:
Tests passed on slow machines. These are the test job ids - 939299 to 93930

arungrover:
PR is ready for merge

PR #1753, opened 2020-05-19
pbs_snapshot obfuscate fails for remote hosts

agrawalravi90:
#### Describe Bug or Feature

Fixing 2 issues:
- obfuscate.map file wasn't correctly copied over from remote host:
2020-05-19 18:22:35,841 INFOCLI2 pbs: /usr/bin/scp -p momhost:/home/ragrawal/obfuscate.map ./momhost_/home/ragrawal/obfuscate.map
2020-05-19 18:22:40,144 ERROR    err: ['./momhost_/home/ragrawal/obfuscate.map: No such file or directory']
2020-05-19 18:22:40,144 ERROR    ['./momhost_/home/ragrawal/obfuscate.map: No such file or directory']
2020-05-19 18:22:40,144 ERROR    Error copying map file from host momhost

- --obfuscate would fail on remote hosts when the remote host didn't have any .JB files as follows:
File "/opt/pbs/unsupported/fw/bin/pbs_snapshot.py", line 337, in <module>
main_snap = capture_local_snap(with_sudo)
File "/opt/pbs/unsupported/fw/bin/pbs_snapshot.py", line 197, in capture_local_snap
obj.obfuscate_snapshot(snap_name, map_file, sudo_val)
File "/opt/pbs/unsupported/fw/ptl/utils/pbs_snaputils.py", line 523, in obfuscate_snapshot
for name in os.listdir(jobspath):
FileNotFoundError: [Errno 2] No such file or directory: '/home/pbsroot/TEST/tmp/tests/snapshot_20191212_01_06_16/mom_priv/jobs''

#### Describe Your Change

- Added a check to see if listdir() on the jobs files returns None
- We were using the full path to copy the map file over, that's what was causing the copy to fail, now extracting the filename from path and using that for copy

### DIFF
--- a/test/fw/ptl/utils/pbs_snaputils.py
+++ b/test/fw/ptl/utils/pbs_snaputils.py
@@ -531,7 +531,7 @@ class ObfuscateSnapshot(object):
                     if printjob is not None:
                         cmd = [printjob, fpath]
                         ret = self.du.run_cmd(cmd=cmd, sudo=sudo_val,
-                                                as_script=True)
+                                              as_script=True)
                     self.du.rm(path=fpath)
                     if ret is not None and ret["out"] is not None:
                         jbcontent[name] = "\n".join(ret["out"])
--- a/test/fw/ptl/utils/pbs_snaputils.py
+++ b/test/fw/ptl/utils/pbs_snaputils.py
@@ -401,7 +401,7 @@ class ObfuscateSnapshot(object):
 
         shutil.move(fout, file_path)
 
-    def obfuscate_acct_logs(self, snap_dir):
+    def obfuscate_acct_logs(self, snap_dir, sudo_val):
         """
         Helper function to obfuscate accounting logs
 
@@ -417,7 +417,7 @@ class ObfuscateSnapshot(object):
         attrs_to_obf += acct_extras
 
         acct_path = os.path.join(snap_dir, "server_priv", "accounting")
-        acct_fnames = os.listdir(acct_path)
+        acct_fnames = self.du.listdir(path=acct_path, sudo=sudo_val)
         for acct_fname in acct_fnames:
             acct_fpath = os.path.join(acct_path, acct_fname)
             self._obfuscate_acct_file(attrs_to_obf, acct_fpath)
@@ -498,7 +498,7 @@ class ObfuscateSnapshot(object):
         # Note: We can't rely on sed to do this because there might be logs
         # From long back which have usernames & hostnames that didn't get
         # captured in the qstat/pbs_rstat/pbsnodes outputs
-        self.obfuscate_acct_logs(snap_dir)
+        self.obfuscate_acct_logs(snap_dir, sudo_val)
 
         # Until we can support obfuscating daemon logs, delete them
         svr_logs = os.path.join(snap_dir, SVR_LOGS_PATH)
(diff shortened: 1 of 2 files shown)


bhroam:
Travis failed due to a PEP8 failure.

test/fw/ptl/utils/pbs_snaputils.py
```diff
@@ -522,21 +522,23 @@ def obfuscate_snapshot(self, snap_dir, map_file, sudo_val):
...
-                                          as_script=True)
-                self.du.rm(path=fpath)
-                if ret is not None and ret["out"] is not None:
-                    jbcontent[name] = "\n".join(ret["out"])
-            # Also delete any other files/directories inside mom_priv/jobs
-            else:
-                path = os.path.join(jobspath, name)
-                self.du.rm(path=path, recursive=True, force=True)
+        jbfilelist = self.du.listdir(path=jobspath, sudo=sudo_val)
+        if jbfilelist is not None:
```
hirenvadalia, line 526: Just a suggestion (because looks like you have used `is not None` in the whole file): `is not None` is not required here and other places you can simply do `if jbfilelist:`

test/fw/ptl/utils/pbs_snaputils.py
```diff
@@ -522,21 +522,23 @@ def obfuscate_snapshot(self, snap_dir, map_file, sudo_val):
...
+            for name in jbfilelist:
+                if name.endswith(".JB"):
+                    ret = None
+                    fpath = os.path.join(jobspath, name)
+                    if printjob is not None:
+                        cmd = [printjob, fpath]
+                        ret = self.du.run_cmd(cmd=cmd, sudo=sudo_val,
+                                                as_script=True)
+                    self.du.rm(path=fpath)
+                    if ret is not None and ret["out"] is not None:
```
hirenvadalia, line 536: shouldn't you need to check for `ret['rc'] == 0` ? if not 0 then log some error message or log whatever in `ret['err']`

test/fw/ptl/utils/pbs_snaputils.py
agrawalravi90: Why? I specifically want to test that it is not None .. "if X" will return True/False for a number of reasons .. probably doesn't make a difference here, but I think it's good practice to always check for exactly what you expect the function to return.

test/fw/ptl/utils/pbs_snaputils.py
agrawalravi90: Since I'm not setting logerr=False for run_cmd(), it should automatically log any errors encountered right?

agrawalravi90:
@bhroam thanks for the sign-off!
@hirenvadalia Thanks a ton for the review! I've replied to your comments, please let me know what you think.

test/fw/ptl/utils/pbs_snaputils.py
hirenvadalia: In my view, self.du.listdir itself is wrong, in python, you should return only one return type not multiple from the method, and here self.du.listdir is returning either None or list, which it shouldn't. It should return list always and if the given path doesn't exist then it should either return an empty list or raise an exception. (Not sure but if AFAIK, this will be flagged as
But anyway, that's for future enhancement/refactoring of PTL, I am ok if you don't want to change right now, I was just giving suggestion, take it or leave it up to you 😄

test/fw/ptl/utils/pbs_snaputils.py
hirenvadalia: Ah, missed that! please ignore comment

test/fw/ptl/utils/pbs_snaputils.py
agrawalravi90: I agree, the functions should ideally return only 1 type. Thanks for the suggestions, they are very much welcome, but I'd like to keep this one as it is :)

PR #1729, opened 2020-05-11
Add logs in error path and remove unused code in pid.c

Shrini-h:
#### Describe Bug or Feature

some error paths were not logged, making it difficult to debug rare issues. 

#### Describe Your Change

added more logs in all the error path and deleted unused code

PR #1765, opened 2020-05-22
add to #7895adb Add logs in error path of pid.c

Shrini-h:
#### Describe Bug or Feature

additions and corrections to commit https://github.com/PBSPro/pbspro/commit/7895adb1ac177cf94ea9a169fd34c7128836ba5f  merged through https://github.com/PBSPro/pbspro/pull/1729 `Add logs in error path and remove unused code in pid.c`

#### Describe Your Change

used the new log_errf instead of log_eventf
reverted printf format specifier for HANDLE back to %d
added missing CloseHandles()
removed local log buffer variables

### DIFF
--- a/src/lib/Libwin/pid.c
+++ b/src/lib/Libwin/pid.c
@@ -125,15 +125,14 @@ void
 printpids(void)
 {
 	int	i;
-	char	logb[LOG_BUF_SIZE] = {'\0' } ;
 
 	for (i=0; i < pids_cnt; i++) {
-		sprintf(logb, "printpids: pid_handles[%d] = %d", i, pid_handles[i]);
-		log_event(PBSEVENT_SYSTEM | PBSEVENT_ADMIN | PBSEVENT_FORCE| PBSEVENT_DEBUG, PBS_EVENTCLASS_FILE, LOG_NOTICE, "", logb);
+		sprintf(log_buffer, "printpids: pid_handles[%d] = %d", i, pid_handles[i]);
+		log_event(PBSEVENT_SYSTEM | PBSEVENT_ADMIN | PBSEVENT_FORCE| PBSEVENT_DEBUG, PBS_EVENTCLASS_FILE, LOG_NOTICE, "", log_buffer);
 	}
 
-	sprintf(logb, "printpids: pids_cnt=%d pids_nextidx=%d", pids_cnt, pids_nextidx);
-	log_event(PBSEVENT_SYSTEM | PBSEVENT_ADMIN | PBSEVENT_FORCE| PBSEVENT_DEBUG, PBS_EVENTCLASS_FILE, LOG_NOTICE, "", logb);
+	sprintf(log_buffer, "printpids: pids_cnt=%d pids_nextidx=%d", pids_cnt, pids_nextidx);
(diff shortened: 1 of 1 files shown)


src/lib/Libwin/pid.c
```diff
@@ -198,14 +197,14 @@ waitpid(HANDLE pid, int *statp, int opt)
...
 				rval = 0;
 			else if (ret == WAIT_FAILED) { /* notion of an abnormal exit */
-				log_eventf(PBSEVENT_ERROR, 0, LOG_ERR, __func__, "WaitForSingleObject(%p, %d) failed with errno %d, i[%d]",
-					pid_handles[i], timeout, GetLastError(), i);
+				log_errf(-1, __func__, "WaitForSingleObject(%d, %d) failed i[%d/%d]", pid_handles[i], timeout, i, pids_cnt);
 				rval = (HANDLE)-1;
 				*statp = -1;
 			} else {
-				sprintf(logb,"found pid_handles[%d]=%p to have exited", i, pid_handles[i]);
+				sprintf(logb,"found pid_handles[%d]=%d to have exited", i, pid_handles[i]);
```
SudeshnaMoulik, line 204: Before this line should we check whether "ret == WAIT_OBJECT_0" ?

Shrini-h: as per the [doc](https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-waitforsingleobject#return-value) apart from `WAIT_OBJECT_0` the only other remaining return code remaining is `WAIT_ABANDONED`.  and `WAIT_ABANDONED` is only for mutex objects, and since we are not handling mutex handles here, we can safely ignore the check.   However if you still want to add the check let me know , i don't mind either ways

> 

WAIT_ABANDONED0x00000080L | The specified object is a mutex object that was not released by the thread that owned the mutex object before the owning thread terminated. Ownership of the mutex object is granted to the calling thread and the mutex state is set to nonsignaled.If the mutex was protecting persistent state information, you should check it for consistency.
-- | --

SudeshnaMoulik: Yes we are not handling any mutex here

Shrini-h:
Thank you @SudeshnaMoulik and @sakshamgarg

src/lib/Libwin/pid.c
```diff
@@ -129,7 +128,7 @@ printpids(void)
 	char	logb[LOG_BUF_SIZE] = {'\0' } ;
 
 	for (i=0; i < pids_cnt; i++) {
-		sprintf(logb, "printpids: pid_handles[%d] = %p", i, pid_handles[i]);
+		sprintf(logb, "printpids: pid_handles[%d] = %d", i, pid_handles[i]);
```
npadole20, line 131: Can we use log_buffer here? We can remove logb variable. 
Also I would suggest renaming [logb](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/logb-logbf-logbl-logb-logbf?view=vs-2019) to something else.

Shrini-h: Thank You @npadole20 I have removed all the local log buffers from entire file. see https://github.com/PBSPro/pbspro/pull/1765/commits/346f8d488e930ae0403dae9597bc32d7deec9823

PR #1766, opened 2020-05-25
if getting new mom addr fails on init job, preserve original mom addr

vchlum:
#### Describe Bug or Feature

The jobs can lose the mom addr on job initialization. The reason for the loss is quite rare. In our case, it was (temporary) bad configuration on DNS and we restarted the pbs server during that time. The jobs were initialized and get_addr_of_nodebyname() sometimes failed during the initialization -> mom addr was set to 0 for those jobs. If this happens the jobs can not be recovered. The server is not able to contact the jobs (even after another server restart).

#### Describe Your Change

I suggest preserving the original mom addr and mom port of jobs if the get_addr_of_nodebyname() fails on job initialization. This can be caused by only a temporary error from the DNS. I think it is too strict and unnecessary to lose such a job. We can try to use the original mom addr and port for those jobs.

This bug is not easily reproduced and the fix hard to test.

### DIFF
--- a/src/server/pbsd_init.c
+++ b/src/server/pbsd_init.c
@@ -1720,8 +1720,8 @@ pbsd_init_job(job *pjob, int type)
 					pjob->ji_qs.ji_un.ji_exect.ji_momaddr = new_momaddr;
 					pjob->ji_qs.ji_un.ji_exect.ji_momport = new_momport;
 				} else {
-					log_event(PBSEVENT_DEBUG3, PBS_EVENTCLASS_JOB,
-						LOG_DEBUG, pjob->ji_qs.ji_jobid,
+					log_event(PBSEVENT_DEBUG, PBS_EVENTCLASS_JOB,
+						LOG_INFO, pjob->ji_qs.ji_jobid,
 						"Failed to update mom address. Mom address not changed.");
 				}
 			} else {
--- a/src/server/pbsd_init.c
+++ b/src/server/pbsd_init.c
@@ -1488,8 +1488,6 @@ pbsd_init_job(job *pjob, int type)
 {
 	int	 newstate;
 	int	 newsubstate;
-	pbs_net_t new_momaddr;
-	unsigned int new_momport;
 
 	pjob->ji_momhandle = -1;
 	pjob->ji_mom_prot = PROT_INVALID;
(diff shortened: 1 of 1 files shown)


src/server/pbsd_init.c
```diff
@@ -1709,11 +1710,15 @@ pbsd_init_job(job *pjob, int type)
 			(pjob->ji_qs.ji_un.ji_exect.ji_momaddr != 0)) {
 
 			if (pjob->ji_wattr[(int)JOB_ATR_exec_host].at_flags & ATR_VFLAG_SET) {
-				pjob->ji_qs.ji_un.ji_exect.ji_momaddr =
+				new_momaddr =
```
hirenvadalia, line 1713: @vchlum thank you for change, change look good, just minor comment: it seems like we are not using new_momaddr and new_momport outside this if condition, so can we reduce its scope to this if condition only? also, can we add some debug log saying like failed to find new mom addr at DEBUG3 or up-level? so it can help in debugging in the future.

vchlum: Thank you for comments @hirenvadalia. Done.

src/server/pbsd_init.c
```diff
@@ -1709,11 +1708,22 @@ pbsd_init_job(job *pjob, int type)
...
 					pjob->ji_wattr[(int)JOB_ATR_exec_host].
-					at_val.at_str, &d);
-				pjob->ji_qs.ji_un.ji_exect.ji_momport = d;
+					at_val.at_str, &new_momport);
+
+				if (new_momaddr != 0) {
+					pjob->ji_qs.ji_un.ji_exect.ji_momaddr = new_momaddr;
+					pjob->ji_qs.ji_un.ji_exect.ji_momport = new_momport;
+				} else {
+					log_event(PBSEVENT_DEBUG3, PBS_EVENTCLASS_JOB,
```
prakashcv13, line 1723: I am just wondering if info would be a better log level here.

vchlum: @prakashcv13 Thanks.

LOG_INFO sounds good. I agree. Should we change the event type too?

I think yes. This log message will not spam the log. The occurrence of the message is rare, and what is more important the message can occur only once -  changing debug level and restarting service will not invoke the message reliably. I would say PBSEVENT_DEBUG is suitable as a type. Or maybe higher?

prakashcv13: I agree @vchlum

vchlum: Done, please see @prakashcv13 .

PR #1770, opened 2020-05-27
Windows: mom hook fix

SudeshnaMoulik:
#### Describe Bug or Feature

- For mom hook runs as user (pbsuser), the environment of the user always shows service user acc environment under which mom is running
- For mom hook, the getpass.getuser() always shows previous job's user name. Observed this in both for admin and user hook.

These behaviors are noticed only for Windows mom hook.

#### Describe Your Change

- Export the user environment when the hook is running as user
- Inside Windows mom process, we were setting 'USER' env variable at some places in code and didn't reset that.

[before_fix.txt](https://github.com/openpbs/openpbs/files/4686975/before_fix.txt)
After fix:
[admin_hook_test.txt](https://github.com/openpbs/openpbs/files/4686977/admin_hook_test.txt)
[user_hook_test.txt](https://github.com/openpbs/openpbs/files/4686983/user_hook_test.txt)
[mom_log.txt](https://github.com/openpbs/openpbs/files/4686985/mom_log.txt)

### DIFF
--- a/src/lib/Libwin/passwd.c
+++ b/src/lib/Libwin/passwd.c
@@ -4148,7 +4148,9 @@ wsystem(char *cmdline, HANDLE user_handle)
 	} else {
 		flags = flags|CREATE_UNICODE_ENVIRONMENT;
 		if (!CreateEnvironmentBlock(&user_env, user_handle, FALSE)) {
+			run_exit = GetLastError();
 			log_err(-1, __func__, "failed in CreateEnvironmentBlock");
+			goto end;
 		}
 		rc=CreateProcessAsUser(user_handle, NULL, cmd,
 			NULL, NULL, TRUE, flags,
@@ -4185,6 +4187,7 @@ wsystem(char *cmdline, HANDLE user_handle)
 	if (user_env)
 		DestroyEnvironmentBlock(user_env);
 
+end:
 	return (run_exit);
 }
 
--- a/src/lib/Libwin/passwd.c
+++ b/src/lib/Libwin/passwd.c
@@ -4118,6 +4118,7 @@ wsystem(char *cmdline, HANDLE user_handle)
 	char			*temp_dir = NULL;
 	int			changed_dir = 0;
 	DWORD		stat;
+	LPVOID  user_env = NULL;
 
 	si.cb = sizeof(si);
 	si.lpDesktop = NULL;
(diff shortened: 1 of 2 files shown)


src/resmom/requests.c
```diff
@@ -2834,6 +2834,7 @@ req_cpyfile(struct batch_request *preq)
 	 * the same permissions as TMPDIR
 	 */
 	(void)revert_impersonated_user();
+	SetEnvironmentVariable("USER", NULL);
```
hirenvadalia, line 2837: not sure, but AFAIK, in code as of now I can see only two setenv for USER, and which is in fork_to_user() (both version windows as well as *nix), so my question is instead doing this can we remove that setenv from fork_to_user()?

raviv1995: +1, going a bit further I think they were used in dce_login - refer to the deletions in the dce_login removal commit - https://github.com/openpbs/openpbs/commit/ad9f208faca09679082d889db56c677d8bfc7b85#diff-68d448a9a18cff40450417ba0ffc3e0dL280

SudeshnaMoulik: Hi @hirenvadalia, Thanks for review. I had this idea, but was not sure as pbs_idled search for that. And am not sure whether it break anything

hirenvadalia: That pbs_idled.c is for *nix one, for windows there is pbs_idled_win.c which doesn't find user name, and in *nix pbs_idled is stared by user only (or login bashrc when user log-in into system) when cycle harvesting is configured, so even if we remove setenv from fork_to_user it won't affect pbs_idled.

SudeshnaMoulik:
Attached latest logs for Windows:
[user_hook_test.txt](https://github.com/openpbs/openpbs/files/4688630/user_hook_test.txt)
[admin_hook_test.txt](https://github.com/openpbs/openpbs/files/4688631/admin_hook_test.txt)
[mom_log2.txt](https://github.com/openpbs/openpbs/files/4688635/mom_log2.txt)

Ran regression (job id: 1898)

src/lib/Libwin/passwd.c
```diff
@@ -4145,9 +4146,13 @@ wsystem(char *cmdline, HANDLE user_handle)
 			NULL, NULL, TRUE, flags,
 			NULL, NULL, &si, &pi);
 	} else {
+		flags = flags|CREATE_UNICODE_ENVIRONMENT;
+		if (!CreateEnvironmentBlock(&user_env, user_handle, FALSE)) {
+			log_err(-1, __func__, "failed in CreateEnvironmentBlock");
```
vstumpf, line 4151: if we fail to create the environment block, shouldn't we return out of the function? Instead of continuing to run the process.

subhasisb: sudehsna addressed it, hence resolving this

PR #1771, opened 2020-05-27
Job submission from ROOT_USER changed to TST_USR in TestJobPerf testsuite

visheshh:
#### Describe Bug or Feature
Job submitted as a root user fails in TestJobPerf testsuite, if test is run from sudo user

#### Describe Your Change
Replace ROOT_USER to TST_USR and ADMIN_USER to TST_USR1

[buggy_root.txt](https://github.com/openpbs/openpbs/files/4693420/buggy_root.txt)
[fixed.txt](https://github.com/openpbs/openpbs/files/4693422/fixed.txt)

### DIFF
--- a/test/tests/performance/pbs_jobperf.py
+++ b/test/tests/performance/pbs_jobperf.py
@@ -153,7 +153,7 @@ class TestJobPerf(TestPerformance):
             i = 0
             users = [TEST_USER1, TEST_USER2, TEST_USER3, TEST_USER4,
                      TEST_USER5, TEST_USER6, TEST_USER7, TEST_USER,
-                     TST_USR, ADMIN_USER]
+                     TST_USR, TST_USR1]
             a = {'log_events': config['svr_log_level']}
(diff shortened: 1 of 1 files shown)


test/tests/performance/pbs_jobperf.py
```diff
@@ -153,7 +153,7 @@ def test_job_performance_sched_off(self):
             i = 0
             users = [TEST_USER1, TEST_USER2, TEST_USER3, TEST_USER4,
                      TEST_USER5, TEST_USER6, TEST_USER7, TEST_USER,
-                     ROOT_USER, ADMIN_USER]
+                     TST_USR, ADMIN_USER]
```
bhroam, line 156: If ROOT_USER was a problem, is ADMIN_USER ok?

visheshh: It works fine, But changed it as it may cause issue in a non linux platform

visheshh:
@prakashcv13 @bhroam Updated.

Issue #1711, opened 2020-05-05
server attribute 'jobscript_max_size' not working properly. 


dinesh1306:
Steps to reproduce:

1. set 'jobscript_max_size' with some value
2. submit job with job-script have size more than settled jobscript_max_size expected is job should be rejected.
3. unset 'jobscript_max_size'
4. submit job with job-script have size more than settled jobscript_max_size expected is job should be submitted and running, but job gets terminated immediately with Job Exit status 127.

I submitted job with -koe option and keeping server and mom logs at high value and got error message in job error file that script file not available in /var/spool/pbs/mom_priv/jobs/ for resprctive job.

I restarted PBS server and submit new job with script, then job was running fine and finished normally.

arungrover:
The issue is that when a job is rejected due to job script size exceeding jobscript_max_size server does not unlink the job from its list.
In fact, if you submit more than one job exceeding jobscript_max_size and then submit jobs that do not exceed job_script_size then all those jobs will not run because Server locates the wrong job from its new jobs list and attaches the new job script to old jobs.

PR #1772, opened 2020-05-27
server does not cleanup job when jobscript exceeds size

arungrover:
#### Describe Bug or Feature
When a user submits a job with a job script that exceeds jobscript_max_size limit, server does not cleanup the job properly. This results into subsequent jobs getting their job scripts attached to the previously rejected job and makes these jobs fail as well.

#### Describe Your Change
This problem only happens when jobs are submitted through qsub daemon process. 
What happens is that inside the server we register a connection handler for qsub clients. In this connection handler, we cleanup jobs if the connection breaks.
This all works fine until qsub is spawned in the background. When a job is submitted with a job script larger than the jobscript_max_size limit, the server rejects the job. But, it does not cleanup the job thinking that it will be cleaned when the connection ends.
Now if the same user modifies the script and submits it after bringing its size below the limit, After receiving the job script, PBS server tries to locate the job. The locate function it calls, checks the socket-id the job is coming from and the username to find out the job structure. The problem is that the job is coming from the same background qsub process and the user is the same user whose previous job was rejected. This makes PBS server locate the wrong job and attach the new jobscript to a previously rejected job.

The solution is simply to cleanup jobs properly when the job is rejected and do not rely on the connection cleanup handler.

resolves: #1711 

#### Link to Design Doc
NA

I did not see a recurring benefit of writing a PTL test for this issue. Please let me know if you think otherwise.

[test_jobscript_after.txt](https://github.com/openpbs/openpbs/files/4691197/test_jobscript_after.txt)
[test_jobscript_before.txt](https://github.com/openpbs/openpbs/files/4691198/test_jobscript_before.txt)

### DIFF
--- a/src/server/req_quejob.c
+++ b/src/server/req_quejob.c
@@ -1480,12 +1480,14 @@ req_jobscript(struct batch_request *preq)
 	/* add the script to the job */
 	size = get_bytes_from_attr(&attr_jobscript_max_size);
 	if (preq->rq_ind.rq_jobfile.rq_size > size){
+		job_purge(pj);
 		req_reject(PBSE_JOBSCRIPTMAXSIZE, 0, preq);
 		return;
 	}
 	temp = realloc(pj->ji_script, pj->ji_qs.ji_un.ji_newt.ji_scriptsz +
 		preq->rq_ind.rq_jobfile.rq_size + 1);
 	if (!temp) {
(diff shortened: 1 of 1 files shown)


arungrover:
I don't know why CLA is not reporting status. I'll close this PR and then reopen again.

PR #1774, opened 2020-05-28
Allow pbs_cgroups testsuite to run even without mem subsystem.

bayucan:
#### Describe Bug or Feature

* On some UV/HPE related systems, the 'memory' subsystem of cgroups may not be mounted.
* This causes the pbs_cgroups hook PTL testsuite to fail to run, as PTL infrastructure would detect a UV/HPE-like system, and attempt to enable pbs_cgroups hook with a default hook config file that has 'mem' and 'memsw' enabled by default.
* Since there's no memory subsystem, the pbs_cgroups hook fails to run with 'no memory subsystem' detected message, even though hook config file has this enabled.

#### Describe Your Change

* The PTL infrastructure has been updated to provide an option of not enabling of pbs_cgroups hook with default hook config file containing 'mem', 'memsw' enabled setting, under UV/HPE-like systems.
* Since the pbs_cgroups PTL testsuite takes care of enabling pbs_cgroups hook and installing its own hook config file per each test case, the testsuite would make use of the option mentioned above.
* And in each test case, the 'mem' and 'memsw' are enabled and tested only if the memory subystem is mounted.

The pbs_cgroups hook PTL testsuite now runs on an HPE/UV-related system that does not have 'memory' subsystem mounted:
* [ptl.cgroups_hook3.txt](https://github.com/openpbs/openpbs/files/4702797/ptl.cgroups_hook3.txt)
* [ptl.cgroups_hook3.rerun.txt](https://github.com/openpbs/openpbs/files/4702799/ptl.cgroups_hook3.rerun.txt)
* [ptl.assign_resources_mem_only_vnode.not-cpuset.txt](https://github.com/openpbs/openpbs/files/4704036/ptl.assign_resources_mem_only_vnode.not-cpuset.txt)
* [ptl.assign_resources_mem_only_vnode.cpuset_withmem.txt](https://github.com/openpbs/openpbs/files/4704040/ptl.assign_resources_mem_only_vnode.cpuset_withmem.txt)
* [ptl.assign_resources_mem_only_vnode.cpuset_nomem.txt](https://github.com/openpbs/openpbs/files/4704039/ptl.assign_resources_mem_only_vnode.cpuset_nomem.txt)

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -13098,6 +13098,10 @@ class MoM(PBSService):
                        access or dictionary containing
                        {'dbname':...,'user':...,'port':...}
     :type db_access: str or dictionary
+    :param revert_to_default: set to True if you like this instance of
+                              mom to be reverted to its default when
+                              revert_mom() is called.
+    :type revert_to_default: bool
     """
     dflt_attributes = {}
     conf_to_cmd_map = {'PBS_MOM_SERVICE_PORT': '-M',
@@ -13150,6 +13154,7 @@ class MoM(PBSService):
             self.dflt_config = {'$clienthost': self.server.hostname}
         self.version = None
         self._is_cpuset_mom = None
+        self.revert_to_default = True
 
     def __del__(self):
         del self.__dict__
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -13151,8 +13151,9 @@ class MoM(PBSService):
         self.version = None
         self._is_cpuset_mom = None
 
-        # set to True if you like this instance of mom to be reverted to its
-        # default when revert_mom() is called.
+        # If this is true, the mom will revert to default.
+        # This is true by default, but can be set to False if
+        # required by a test
         self.revert_to_default = True
 
     def __del__(self):
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -13098,10 +13098,6 @@ class MoM(PBSService):
                        access or dictionary containing
                        {'dbname':...,'user':...,'port':...}
     :type db_access: str or dictionary
-    :param revert_to_default: set to True if you like this instance of
-                              mom to be reverted to its default when
-                              revert_mom() is called.
-    :type revert_to_default: bool
     """
     dflt_attributes = {}
     conf_to_cmd_map = {'PBS_MOM_SERVICE_PORT': '-M',
@@ -13154,6 +13150,9 @@ class MoM(PBSService):
             self.dflt_config = {'$clienthost': self.server.hostname}
         self.version = None
         self._is_cpuset_mom = None
+
+        # set to True if you like this instance of mom to be reverted to its
+        # default when revert_mom() is called.
         self.revert_to_default = True
 
     def __del__(self):
--- a/test/fw/ptl/utils/pbs_testsuite.py
+++ b/test/fw/ptl/utils/pbs_testsuite.py
@@ -1500,7 +1500,7 @@ class PBSTestSuite(unittest.TestCase):
             mom.signal('-HUP')
             self.server.expect(NODE, a, id=mom.shortname + '[0]', interval=1)
         else:
-            self.server.manager(MGR_CMD_CREATE, NODE, None, mom.shortname)
+            self.server.manager(MGR_CMD_CREATE, NODE, id=mom.shortname)
             self.server.expect(NODE, a, id=mom.shortname, interval=1)
         return mom
 
--- a/test/fw/ptl/utils/pbs_testsuite.py
+++ b/test/fw/ptl/utils/pbs_testsuite.py
@@ -1454,8 +1454,9 @@ class PBSTestSuite(unittest.TestCase):
             self.assertTrue(mom.isUp(), msg)
         mom.pbs_version()
         restart = False
-        if ((self.revert_to_defaults and self.mom_revert_to_defaults) or
-                force):
+        enabled_cpuset = False
+        if ((self.revert_to_defaults and self.mom_revert_to_defaults and
+             mom.revert_to_default) or force):
             # no need to delete vnodes as it is already deleted in
             # server revert_to_defaults
             mom.delete_pelog()
@@ -1482,16 +1483,15 @@ class PBSTestSuite(unittest.TestCase):
                 # when the natural node is created below
                 # HUP may not be enough if exechost_startup is delayed
                 restart = True
+                enabled_cpuset = True
         if restart:
             mom.restart()
         else:
             mom.signal('-HUP')
         if not mom.isUp():
             self.logger.error('mom ' + mom.shortname + ' is down after revert')
-        if not mom.is_cpuset_mom():
-            self.server.manager(MGR_CMD_CREATE, NODE, None, mom.shortname)
         a = {'state': 'free'}
-        if mom.is_cpuset_mom():
+        if enabled_cpuset:
             # Checking whether the CF file was copied really belongs in code
             # that changes the config file -- i.e. after enable_cgroup_cset
             # called above. We're not sure it is always called here,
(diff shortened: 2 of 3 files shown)


test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -1500,6 +1500,7 @@ def revert_mom(self, mom, force=False):
             mom.signal('-HUP')
             self.server.expect(NODE, a, id=mom.shortname + '[0]', interval=1)
         else:
+            self.server.manager(MGR_CMD_CREATE, NODE, None, mom.shortname)
```
bhroam, line 1503: Remove the None and say id=mom.shortname

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -1482,16 +1483,15 @@ def revert_mom(self, mom, force=False):
...
             mom.restart()
         else:
             mom.signal('-HUP')
         if not mom.isUp():
             self.logger.error('mom ' + mom.shortname + ' is down after revert')
-        if not mom.is_cpuset_mom():
-            self.server.manager(MGR_CMD_CREATE, NODE, None, mom.shortname)
         a = {'state': 'free'}
-        if mom.is_cpuset_mom():
+        if enabled_cpuset:
```
bhroam, line 1494: Why have the enabled_cpuset variable if it is set when mom.is_cpuset_mom() is set.  Just use mom.is_cpuset_mom() like before.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -13098,6 +13098,10 @@ class MoM(PBSService):
                        access or dictionary containing
                        {'dbname':...,'user':...,'port':...}
     :type db_access: str or dictionary
+    :param revert_to_default: set to True if you like this instance of
+                              mom to be reverted to its default when
+                              revert_mom() is called.
+    :type revert_to_default: bool
```
bhroam, line 13104: revert_to_defaults was not added as a param.  It is just set to true.  Do you really need it?

bayucan: You're right. It's not a param since it's not even passed to class MoM's init(). I just needed an instance variable that can be update per testsuite/test case. It defaults to True, but on the pbs_cgroups_hook ptl testsuite, I need to set it to False under setUp, so that the revert_to_defaults() mom action would not do the enable pbs_cgroups hook thing. I'll update the fix.

test/fw/ptl/utils/pbs_testsuite.py
bayucan: It's because it could be a mom.is_cpuset_mom() but mom.revert_to_default = False. I want to do the action specifically when enable_cpuset was done, where pbs_cgroups hook got enabled along with a default hook config file.

test/fw/ptl/utils/pbs_testsuite.py
bayucan: Ok, I'll update.

test/fw/ptl/utils/pbs_testsuite.py
bhroam: I see it now, I thought it was set to True only if is_cpuset_mom() is true.  That if block is also in the if revert_to_defaults block.  Thanks for the explanation.

test/fw/ptl/lib/pbs_testlib.py
bhroam: Maybe move this comment to where you set the instance variable?

bayucan: Good idea. Thanks.

vccardenas:
It looks like there were other tests that should not have ran:
    test_cgroup_cpuset_mem_fences(self):
    test_cgroup_cpuset_mem_hardwall(self):
    test_cgroup_cpuset_memory_spread_page(self):

bayucan:
@vccardenas : Ok, I'll update those tests too.

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -1457,7 +1465,7 @@ def load_default_config(self, mom_checks=True):
              'content-encoding': 'default',
              'input-file': self.config_file}
         self.server.manager(MGR_CMD_IMPORT, HOOK, a, self.hook_name)
-        if not mom_checks:
+        if not mom_checks or self.moms_list[0].is_cpuset_mom():
```
lisa-altair, line 1468: Don't we want to verify the configuration change has been made on a cpuset mom?  Did you mean to check mom.revert_to_default instead?

bayucan: True, I should be checking mom.revert_to_default instead. I'll update.

bayucan: It's not necessary under cpuset mom to verify the default hook config file was loaded in the local mom, since this pbs_cgroups hook testsuite uses its own hook config file at each test case.

lisa-altair:
`It appears that this test is creating vnodes. On a real UV300 we should not create vnodes.`
There are some folks who believe that even if the test does something like create_vnodes on a cpuset system if the test passes, then we should let the test run and not mark it with @skipOnCpuSet decorator that way we get the most testing possible.  Does this test fail on cpuset systems?  If not, we might want to let the test run.

bayucan:
@lisa-altair : Yes, the test  test_cgroup_assign_resources_mem_only_vnode is passing on non-cpuset systems. @vccardenas : are you okay if I remove the @skipOnCpuset tag on this test? On our real UV300, since there's no memory subsystem mounted, it will be skipped anyway.

vccardenas:
@bayucan yes I agree that we can remove the @skipOncpuset decorator on this test.  Since vnode_per_numa = False no vnodes are created by the hook, it seems ok to run tests that create vnodes.  It makes sense to skip tests that request memory if the cgroup memory subsystem is not mounted.

bayucan:
@vccardenas : Thanks! I've pushed an update and has been tested.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -13154,6 +13150,9 @@ def __init__(self, name=None, attrs={}, pbsconf_file=None, snapmap={},
             self.dflt_config = {'$clienthost': self.server.hostname}
         self.version = None
         self._is_cpuset_mom = None
+
+        # set to True if you like this instance of mom to be reverted to its
+        # default when revert_mom() is called.
```
bhroam, line 13155: How about 'If this is true, the mom will revert to default.  This is true by default, but can be set to False if required by a test'

bayucan: Sure, that's fine. I've updated the comment and pushed a new commit.

lisa-altair:
still works for me.

bhroam:
@bayucan the CLA is acting up.  You need to rename your fork.  Do the following:
1) Push all your work from local branches to pbspro.  You will need to delete your clone
2) Goto github.com an click on bayucan/pbspro.
3) Under bayucan/pbspro there is a row of tabs.  On the far right is a settings tab
4) Under settings there is a box that says Repository Name.  Rename 'pbspro' to 'openpbs'
5) Delete your local fork and reclone the openpbs repo.

If you know how to rename your remotes, you can rename your remotes instead of deleting and recloning.  Deleting your clone will remove the private remotes as well.  You will need to set those back up.

I don't know if the rename itself will trigger this to work, or you need to kick the PR again.  I suspect you'll need to kick it by closing and reopening it.

bayucan:
@bhroam : Ok, I've renamed my local repository mapping to openpbs from "pbspro" to "openpbs". Then I've updated my public_fork to the new path, and re-signed my PR, and pushed to the new public_fork. Let's see if this works...

bhroam:
Thanks for updating your fork, @bayucan.  It appeared to not have done the trick.  It has helped for some PRs and not others.  I'm unsure of what to do.  Maybe @subhasisb can help?

bayucan:
@bhroam : The one thing I didn't do is to close and reopen this pull request. Maybe that would help. If I reopen, would I lose all signoffs and commits?

agrawalravi90:
I did a "recheck PRs" on CLA assistant's website, it seems to have worked ..

arungrover:
Thanks @agrawalravi90 worked for me too  :)

bayucan:
Thanks @bhroam @agrawalravi90 !

PR #1778, opened 2020-05-28
Guard find job against freed jobs

Shrini-h:
#### Describe Bug

Server segfault when deleting job "found" even though the job ID shows the structure has been freed
```
(gdb) bt

#0  0x0000000000487147 in req_deletejob (preq=0x4043acf0) at /home/pbsbuild/ramdisk/workspace/build/pbspro/src/server/req_delete.c:437

#1  0x00000000004f66dc in process_socket (sock=sock@entry=20) at /home/pbsbuild/ramdisk/workspace/build/pbspro/src/lib/Libnet/net_server.c:504

#2  0x00000000004f6892 in wait_request (waittime=waittime@entry=2, priority_context=<optimized out>)

    at /home/pbsbuild/ramdisk/workspace/build/pbspro/src/lib/Libnet/net_server.c:620

#3  0x0000000000450f38 in main (argc=<optimized out>, argv=<optimized out>)

    at /home/pbsbuild/ramdisk/workspace/build/pbspro/src/server/pbsd_main.c:2133
```
#### Describe Your Change

guarded inside `find_job()` by checking the 'X' mark in the first char of freed jobs jobid

Pending !

### DIFF
--- a/src/server/job_func.c
+++ b/src/server/job_func.c
@@ -1205,6 +1205,13 @@ find_job(char *jobid)
 	if ((AVL_jctx != NULL) && ((pkey = svr_avlkey_create(buf)) != NULL)) {
 		if (avl_find_key(pkey, AVL_jctx) == AVL_IX_OK)
 			pj = (job *) pkey->recptr;
+		if (pj && (pj->ji_qs.ji_jobid[0] == 'X')) {
+			/* found a freed job obj hanging around in the job avl tree */
+			log_errf(PBSE_INTERNAL, __func__, "%s stale job found in job avl tree, deletting it",jobid);
+			avl_delete_key(pkey, AVL_jctx);
+			pj = NULL;
+		}
+
 		free(pkey);
 		return (pj);
 	}


src/server/job_func.c
```diff
@@ -1205,6 +1205,13 @@ find_job(char *jobid)
 	if ((AVL_jctx != NULL) && ((pkey = svr_avlkey_create(buf)) != NULL)) {
 		if (avl_find_key(pkey, AVL_jctx) == AVL_IX_OK)
 			pj = (job *) pkey->recptr;
+		if (pj && (pj->ji_qs.ji_jobid[0] == 'X')) {
```
agrawalravi90, line 1208: find_job() seems like the wrong place to do this, shouldn't this be taken care of by job_free() or job_purge()?

Shrini-h: There has been an instance where a job already was found by `find_job()` during a user qdel request which lead to server coredump. Hence @alexis-cousein suggested to add this check to at least thwart a crash, while still looking for the root cause of stale job object in avl index. his analysis is pasted below verbatim


> GDB analysis:
User requests to delete a subjob of an array:
(gdb) print preq->rq_user
$5 = "lg4615", '\000' <repeats 250 times>
(gdb) print preq->rq_ind.rq_delete  
$1 = {rq_cmd = 1, rq_objtype = 2, rq_objname = "1492386[3].pbs", '\000' <repeats 259 times>, rq_attr = {ll_prior = 0x4043b1a8, ll_next = 0x4043b1a8,  
   ll_struct = 0x0}}
The job is "found" using AVL tree indices, but:
(gdb) print  &parent->ji_qs.ji_jobid                      
$16 = (char (*)[274]) 0x3364f968
(gdb) print  parent->ji_qs.ji_jobid   
$17 = "X492386[].pbs", '\000' <repeats 260 times>
(gdb) print  pjob->ji_qs.ji_jobid        
$18 = "X492386[3].pbs", '\000' <repeats 259 times>
The fact that the job ID starts with "X" means that we freed the corresponding structures -- this is used as a marker:

    /* if a subjob (of a Array Job), do not free certain items */
    /* which are malloced and shared with the parent Array Job */
    /* They will be freed when the parent is removed           */

    pj->ji_qs.ji_jobid[0] = 'X';    /* as a "freed" marker */
    free(pj);    /* now free the main structure */

> So I would bet that the job (and the parent) is nowhere to be found in svr_alljobs, but that the AVL tree indices were not updated correctly so we have a vestigial way to 'find' these when we should not. That lets users use "qdel" on subjobs purged from the history to crash the server.
I have not investigated the root cause of the fact the AVL tree still contains stale references, but in find_job adding:

     if ((AVL_jctx != NULL) && ((pkey = svr_avlkey_create(buf)) != NULL)) {
        if (avl_find_key(pkey, AVL_jctx) == AVL_IX_OK)
            pj = (job *) pkey->recptr;
        free(pkey);
        if ( pj != NULL && pj->ji_qs.ji_jobid[0] != 'X')
            return (pj);
        else
            /* AVL tree contains pointer to stale job */
            return NULL;
    }

> at least ensures that if the job structures haven't been overwritten we stand _some_ chance of "not finding" jobs that were freed in job_free.
It's not a complete fix, because if pj points to something already overwritten by heap reuse ji_qs.ji_jobid might point out of valid segments.

I have additionally added a call to `avl_delete_key()` to remove the stale record and also logged a error line, to leave an evidence in the log file

src/server/job_func.c
```diff
@@ -1205,6 +1205,13 @@ find_job(char *jobid)
...
+			/* found a freed job obj hanging around in the job avl tree */
+			log_errf(PBSE_INTERNAL, __func__, "%s stale job found in job avl tree, deletting it",jobid);
```
Shrini-h, line 1210: will correct the typo in "deletting" in the next commit

src/server/job_func.c
agrawalravi90: Thanks for the explanation @Shrini-h . I was kind of suggesting the same thing: fix the bug that causes the free'd subjob to still be part of the job AVL tree, but do it inside job_free()/ job_purge(). The avl_delete_key() should be in job_free()/job_purge() right?

hirenvadalia: > The avl_delete_key() should be in job_free()/job_purge() right?

Its in svr_dequejob() which is gets called by job_purge() not job_free() and based on what I know usage of job_free(), I don't think we can put AVL tree operation in job_free() as somewhere is hook code (IIRC, its in get_hook_results()) we make lite copy of job to add hook/job action after hook runs, and then we call job_free() on it, so if we put AVL operation in job_free() it will remove original job from AVL tree, and which will lead to many errors...

Shrini-h:
> Yeah i feel this is not the right place to fix (or even avoid it). Even if you want to avoid a crash by checking for X, do it in req_deletejob() or similar, but not in find_job().
> 
> Besides, it seems like we know when we put an X on the job, so why not delete the indices right then?

I hope @hirenvadalia 's reply answers this question. https://github.com/openpbs/openpbs/pull/1778#discussion_r432809406

src/server/job_func.c
Shrini-h: Thank you @hirenvadalia for the insight.

@agrawalravi90 as @hirenvadalia mentioned there is already a code path from `job_purge()` that handles AVL record deletion for the job being purged, However since we have seen this instance the real unknown root cause is either making us to skip that code path while purging or the problem could be in the AVL tree operations itself, we might never know. So as stated this is just a guard and not a real bug fix.

alexis-cousein:
Op 31/05/20 om 16:53 schreef Shrinivas Harapanahalli:
>
> @agrawalravi90 <https://github.com/agrawalravi90> as @hirenvadalia <https://github.com/hirenvadalia> mentioned there is already a 
> code path from |job_purge()| that handles AVL record deletion for the job being purged,
>
My guess is that it's job array related. When we have subjobs that finish, IIRC they are _not_ purged, or are they? And when finally 
the array is purged, are we sure that we also mop up all the entries related to subjobs?

src/server/job_func.c
agrawalravi90: Since this is going into master (as opposed to a release branch), I think that we should spend the time to investigate the actual bug and fix that instead of adding a patch/guard which might mask the issue and possibly make it more difficult to debug later.

Also, If there was a bug in the AVL tree implementation, it would have shown up in other places as well, it's been there for years and is heavily used for almost every major operation done by the server, so I think we can probably rule that out and focus on the server's code path.

hirenvadalia: +1 on @agrawalravi90's point, I don't think what this PR is doing is right fix, we should find real culprit and fix it, instead putting bandage on it.

Shrini-h: I too agree. I just hope that the original proposer of the patch (@alexis-cousein) also agrees on this. I will keep this PR open for some more time for further discussions on root causing

Shrini-h: Moreover we might not see this crash location again as it has been already removed previously in https://github.com/openpbs/openpbs/commit/7ce44ce2ada7e6333f32925c82845a620b66f0cd#diff-c3101b9054e0987316d8a49addb67e38L446

alexis-cousein:
> 
> L
> I too agree. I just hope that the original proposer of the patch (@alexis-cousein) also agrees on this. I will keep this PR open for some more time for further discussions on root causing
> 
> 
> 
No, I don't agree: a band-aid is better than nothing, especially when it prevents the server from crashing. One site has applied it and it has stopped the bleeding. We might find it annually reminder that there is an unfixed bug, for the site it has been worthwhile.

Of course we should probably not fix this _silently_ and log a discovered error, but I'm not in favour of letting a server encounter SEGV just because it might get us debugging information (in this case we haven’t found the root case, right, _despite_ having a core?)

Once we find the root cause and fix it _then_ we can remove the band aid (which is indeed just that, since if the data was freed even earlier we might SEGV without seeing a job ID that was marked as stale).

Incidentally: I think this is job-array related (and only happens when deleting subjobs for arrays recently gone stale) and might even only occur with job history disabled.

Shrini-h:
> Incidentally: I think this is job-array related (and only happens when deleting subjobs for arrays recently gone stale) and might even only occur with job history disabled.

Yes, I believe I found the root cause and an end to this PR discussion, and its _NOT_ the case of AVL tree acting up. The real fix is already in the master https://github.com/openpbs/openpbs/commit/7ce44ce2ada7e6333f32925c82845a620b66f0cd#diff-c3101b9054e0987316d8a49addb67e38

Explanation:
So this should happen when we delete the last running subjob.
by design when last subjob is deleted, the call to `chk_array_doneness()`, while updating the last subjob's state,  makes a job_purge of the parent job before the purge of the current subjob. 

example log of completing job array:
```
05/29/2020 10:55:23.676833;0010;Server@x1933-2-0;Job;4001[200].x1933-2-0;Exit_status=0 resources_used.cpupercent=0 resources_used.cput=00:00:00 resources_used.mem=352kb resources_used.ncpus=1 resources_used.vmem=4364kb resources_used.walltime=00:01:40
05/29/2020 10:55:23.676931;0100;Server@x1933-2-0;Job;4001[].x1933-2-0;dequeuing from workq, state 7
05/29/2020 10:55:23.677728;0100;Server@x1933-2-0;Job;4001[200].x1933-2-0;dequeuing from workq, state 5
```
and due to the wrong code (which existed before https://github.com/openpbs/openpbs/pull/1308) of accessing the job tracking table after it was freed, we ended up in this crash.

so when the AVL tree search happened the parent was still alive, and AVL tree was wrongly blamed as it appeared like it was the cause.

Now since we know the root cause has been found to be already fixed in https://github.com/openpbs/openpbs/pull/1308 I will wait for any more comments before closing this PR as already fixed.

Shrini-h:
So I was able to verify that we are avoiding this crash in the current master. I reverted https://github.com/openpbs/openpbs/pull/1308/commits/1bf4a089cbdbe0b0bd0fc2c85fbc33deab77651e from the current master https://github.com/openpbs/openpbs/commit/82803b2a84e604ecdc10dafc4b21b4e16179dcc5 and then valgrinded the pbs_server while running `TestJobArray` PTL test suite. As observed during https://github.com/openpbs/openpbs/pull/1308#issuecomment-534186536 the test cases `test_queue_deletion_after_terminated_subjob` and `test_subjob_comments` generated below invalid reads, which disappeared when run on the master without the revert.

```
==8749== Invalid read of size 4
==8749==    at 0x483516: req_deletejob (req_delete.c:442)
==8749==    by 0x47FD33: dispatch_request (process_request.c:757)
==8749==    by 0x47FA39: process_request (process_request.c:590)
==8749==    by 0x50A8A1: process_socket (net_server.c:521)
==8749==    by 0x50ABB0: wait_request (net_server.c:637)
==8749==    by 0x47DAD1: main (pbsd_main.c:1803)
==8749==  Address 0x15bbf9d8 is 168 bytes inside a block of size 6,392 free'd
==8749==    at 0x4C2ACBD: free (vg_replace_malloc.c:530)
==8749==    by 0x45D9C4: job_free (job_func.c:628)
==8749==    by 0x45DFB6: job_purge (job_func.c:1047)
==8749==    by 0x4C9C4C: svr_saveorpurge_finjobhist (svr_jobfunc.c:4819)
==8749==    by 0x48444C: req_deletejob2 (req_delete.c:834)
==8749==    by 0x483511: req_deletejob (req_delete.c:441)
==8749==    by 0x47FD33: dispatch_request (process_request.c:757)
==8749==    by 0x47FA39: process_request (process_request.c:590)
==8749==    by 0x50A8A1: process_socket (net_server.c:521)
==8749==    by 0x50ABB0: wait_request (net_server.c:637)
==8749==    by 0x47DAD1: main (pbsd_main.c:1803)
==8749==  Block was alloc'd at
==8749==    at 0x4C29BC3: malloc (vg_replace_malloc.c:299)
==8749==    by 0x45D43D: job_alloc (job_func.c:353)
==8749==    by 0x447A15: create_subjob (array_func.c:879)
==8749==    by 0x4A51E7: req_runjob (req_runjob.c:529)
==8749==    by 0x47FF11: dispatch_request (process_request.c:848)
==8749==    by 0x47FA39: process_request (process_request.c:590)
==8749==    by 0x50A8A1: process_socket (net_server.c:521)
==8749==    by 0x50AA1C: wait_request (net_server.c:602)
==8749==    by 0x47DAD1: main (pbsd_main.c:1803)
```

Hence we can close this PR as already fixed.
Thank you all for identifying it as the wrong change

subhasisb:
Thanks Shrini - since we have conclusive proof that we found the root cause, it makes all the sense to close this PR.

agrawalravi90:
Thanks for investigating, I've closed it.

PR #1779, opened 2020-05-28
Fix pbs.get_local_nodename() returns truncated PBS_MOM_NODE_NAME when it is in URL format

minghui-liu:
#### Describe Bug or Feature

pbs.get_local_nodename() returns truncated PBS_MOM_NODE_NAME when it is in "dotted" format but not an IP address. This is preventing some sites from using URL format mom node names.

#### Describe Your Change

Set `mom_short_name` to the value `PBS_MOM_NODE_NAME` in pbs.conf if it is set, regardless of whether it is an IP address. If `PBS_MOM_NODE_NAME` is not set, then use the return value of `gethostname()` but truncated after the first dot.

#### Link to Design Doc

https://pbspro.atlassian.net/wiki/spaces/PD/pages/1824423937/Revision+to+PBS+MOM+NODE+NAME+configuration+variable

SmokeTest, HookSmokeTests, TestMomLocalNodeName
Screen Shot 2020-06-15 at 9 31 18 PM
Screen Shot 2020-06-15 at 9 31 33 PM

### DIFF
--- a/src/resmom/mom_main.c
+++ b/src/resmom/mom_main.c
@@ -8785,15 +8785,14 @@ main(int argc, char *argv[])
 			LOG_WARNING, msg_daemonname, winlog_buffer);
 	}
 #endif
+
 	/*
-	 * Set mom_host to PBS_MOM_NODE_NAME if it is defined.
-	 * Otherwise, set mom_host with a call to gethostname().
+	 * Query the hostname.
+	 * Set mom_host temporarily with a call to gethostname().
 	 */
-	c = 0;
-	if (pbs_conf.pbs_mom_node_name) {
-		/* Hostname was defined by PBS_MOM_NODE_NAME */
-		(void)strncpy(mom_host, pbs_conf.pbs_mom_node_name, (sizeof(mom_host) - 1));
-		mom_host[(sizeof(mom_host) - 1)] = '\0';
+	c = gethostname(mom_host, (sizeof(mom_host) - 1));
+	if (c != 0) {
+		c = 0;
 		ptr = mom_host;
 		/* First character must be alpha-numeric */
 		if (isalnum((int)*ptr)) {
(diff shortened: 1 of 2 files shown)


src/resmom/mom_main.c
```diff
@@ -8742,15 +8742,14 @@ main(int argc, char *argv[])
...
+	 * Query the hostname.
+	 * Set mom_host temporarily with a call to gethostname().
 	 */
-	c = 0;
-	if (pbs_conf.pbs_mom_node_name) {
-		/* Hostname was defined by PBS_MOM_NODE_NAME */
-		(void)strncpy(mom_host, pbs_conf.pbs_mom_node_name, (sizeof(mom_host) - 1));
-		mom_host[(sizeof(mom_host) - 1)] = '\0';
+	c = gethostname(mom_host, (sizeof(mom_host) - 1));
+	if (c != 0) {
```
bayucan, line 8751: Did you mean if (c == 0) which is the sucess return of gethostname()?

minghui-liu: Yes, good catch. Fixed. I'll redo the tests too.

bayucan:
@minghui-liu : code looks fine. @alexis-cousein : can you review this as well, and see if it's sufficient? thanks.

alexis-cousein:
Someone left a comment that uses the word "trucated" instead of "truncated" just to check if I was reading it thoroughly ;-).

Other that, I'm wondering about this:
	c = get_fullhostname(mom_host, mom_host, (sizeof(mom_host) - 1));
	if (c == -1) {
		log_err(-1, msg_daemonname, "Unable to resolve my host name");
#ifdef	WIN32
		g_dwCurrentState = SERVICE_STOPPED;
		ss.dwCurrentState = g_dwCurrentState;
		ss.dwWin32ExitCode = ERROR_INVALID_COMPUTERNAME;
		if (g_ssHandle != 0)
			SetServiceStatus(g_ssHandle, &ss);
#endif
		return (-1);
	}

PBS_MOM_NODE_NAME was taken to configure mom_host _even_ when gethostname() fails --also before this PR. 

*If* mom_host was actually taken from PBS_MOM_NODE_NAME, since it actually conforms to the RFCs for a valid hostname if we get here, I'm wondering if we should actually return(-1) here, and not instead continue with mom_host set to the contents of PBS_MOM_NODE_NAME.

It seems odd, in the case that PBS_MOM_NODE_NAME is set, to tolerate gethostname() failures and substitute PBS_MOM_NODE_NAME for what that call should have yielded, but then not to tolerate canonicalization failures of that name (though we should _certainly_ emit the error written there!)

 That prevents MoM from working if the name resolution framework is _entirely_ broken locally when MoM starts up, but presumably if someone sets PBS_MOM_NODE_NAME to a FQDN that doesn't mean it's not working for the _other_ hosts, and it doesn't mean it will not work later on the host starting up MoM (e.g. on Cloud nodes the name resolution can often start to work quite late).

Not a biggie, though, just wondering about the consistency about what failures we tolerate when the variable is set. 

One _plus_ side of failing here is that people are _eventually_ used to force to do things the really correct way -- e.g. make sure that MoM startup depends on the name resolution framework already being up. 

The downside may be erratic failures to node start up races that we could possibly survive.

alexis-cousein:
FWIW: I'm fine with any decision we may make about the desired semantics. As long as it is, indeed, a decision and not an accident ;-).

bayucan:
@alexis-cousein : I believe it was a real decision to not allow mom to continue if mom_host fails in hostname resolution. Even when adding nodes via qmgr, the given <node_name> is also checked for validity. PBS won't allow the node to be added if it fails on hostname resolution.

alexis-cousein:
OK for me (although I would prefer "truncate" to be properly spelled in the comment).

test/tests/functional/pbs_mom_local_nodename.py
```diff
@@ -0,0 +1,89 @@
...
+# "OpenPBS®", "PBS Professional®", and "PBS Pro™" and Altair's logos is
+# subject to Altair's trademark licensing policies.
+
+from tests.functional import *
+from ptl.lib.pbs_testlib import BatchUtils
+
+
+class TestMomLocalNodeName(TestFunctional):
+
+    def test_url_nodename_not_truncated(self):
```
vstumpf, line 46: Can you add docstrings for the test?

test/tests/functional/pbs_mom_local_nodename.py
```diff
@@ -0,0 +1,89 @@
...
+           'my local nodename is %s'
+           % pbs.get_local_nodename())
+"""
+        self.server.create_import_hook(hook_name, a, hook_body)
+        j = Job(TEST_USER)
+        jid = self.server.submit(j)
+        self.mom.log_match("my local nodename is a.b.c.d")
+
+    def test_ip_nodename_not_truncated(self):
+        ipaddr = socket.gethostbyname(self.mom.hostname)
```
vstumpf, line 66: If you're using socket, can you import it? I know it's imported by something in testlib, but I think it's safer to also import it here.

PR #1783, opened 2020-05-29
Introduce PBS_DAEMON_SERVICE_USER

vstumpf:
#### Describe Bug or Feature
The scheduler doesn't really need to run as root. This PR introduces PBS_DAEMON_SERVICE_USER, which is the user that will run the scheduler, and in the future, other daemons.

#### Describe Your Change
Introduce PBS_DAEMON_SERVICE_USER as a pbs conf variable.
The scheduler will be run as PBS_DAEMON_SERVICE_USER

#### Link to Design Doc
https://openpbs.atlassian.net/wiki/spaces/PD/pages/1692598273/Specify+scheduler+user

*Aug 13*
[scheduser.txt](https://github.com/openpbs/openpbs/files/5071348/scheduser.txt)
[scheduser2.txt](https://github.com/openpbs/openpbs/files/5071349/scheduser2.txt)

*Old*
[smoke.txt](https://github.com/openpbs/openpbs/files/4892875/smoke.txt)
[smoke-daemon.txt](https://github.com/openpbs/openpbs/files/4892876/smoke-daemon.txt)
[tdsu.txt](https://github.com/openpbs/openpbs/files/4892877/tdsu.txt)
[tdsu-daemon.txt](https://github.com/openpbs/openpbs/files/4892878/tdsu-daemon.txt)


src/lib/Libifl/pbs_loadconf.c
```diff
@@ -125,7 +125,8 @@ struct pbs_config pbs_conf = {
 	NULL,					/* mom short name override */
 	NULL,					/* pbs_lr_save_path */
 	0,					/* high resolution timestamp logging */
-	0					/* number of scheduler threads */
+	0,					/* number of scheduler threads */
+	NULL				/* default scheduler user */
```
bhroam, line 129: Is the comment lined up properly?

src/lib/Liblog/chk_file_sec.c
```diff
@@ -108,6 +108,7 @@ teststat(struct stat *sp, int isdir, int sticky, int disallow,
  * @param[in]   isdir - value indicating directory or not
  * @param[in]   sticky - value indicating whether to allow write on directory
  * @param[in]   disallow - value indicating whether admin and owner given access permission
+ * @param[in]   uid - if >0, can be value of owner. else, check if owner <=10
```
bhroam, line 111: I'm not sure you need 'can be' (below too)

src/lib/Liblog/chk_file_sec.c
```diff
@@ -228,7 +230,37 @@ tempstat(struct stat *sp, int isdir, int sticky, int disallow)
...
+
+/**
+ * @brief
+ * 	chk_file_sec_user() - Check file/directory security
+ *      Part of the PBS System Security "Feature"
+ *
+ * @par	To be secure, all directories (and final file) in path must be:
+ *		owned by userid < 10 or owned by uid if set
+ *		owned by groupid < 10 if group writable
+ *		not have world writable unless stick bit set & this is allowed.
```
bhroam, line 246: I'm not sure I understand this sentence.  The first part makes sense, but what about "& this is allowed"

It's also 'sticky bit', not 'stick bit'
(same below and above)

src/scheduler/node_info.c
```diff
@@ -231,6 +208,8 @@ query_node_info_chunk(th_data_query_ninfo *data)
 	data->oarr = ninfo_arr;
 }
 
+
+
```
bhroam, line 212: Remove these 2 added lines

src/scheduler/node_info.c
```diff
@@ -4741,7 +4585,7 @@ create_execvnode(nspec **ns)
  *
  * @param[in]	execvnode	-	the execvnode to parse
  * @param[in]	sinfo		-	server to get the nodes from
- * @param[in]	sel		- select to map 
+ * @param[in]	sel		- select to map
```
bhroam, line 4588: Is this lined up?

src/scheduler/pbs_sched.c
```diff
@@ -851,12 +846,25 @@ main(int argc, char *argv[])
...
 	}
 
-#ifndef DEBUG
-	if ((geteuid() != 0) || (getuid() != 0)) {
-		fprintf(stderr, "%s: Must be run by root\n", argv[0]);
+	if (pbs_loadconf(0) == 0)
+		return (1);
+
+	if (pbs_conf.pbs_daemon_service_user) {
+		struct passwd * pwd = getpwnam(pbs_conf.pbs_daemon_service_user);
```
bhroam, line 853: Attach the * to the variable name

src/scheduler/pbsfs.c
```diff
@@ -142,6 +144,24 @@ main(int argc, char *argv[])
 		fprintf(stderr, "Usage: pbsfs --version\n");
 		exit(1);
 	}
+
+	if (pbs_conf.pbs_daemon_service_user) {
+		struct passwd * user;
```
bhroam, line 149: Attach the * to the variable

src/tools/pbs_probe.c
```diff
@@ -3219,6 +3222,20 @@ conf4primary(FILE *fp, struct infrastruct *pinf)
...
 			}
+			else if (!strcmp(conf_name, "PBS_DAEMON_SERVICE_USER")) {
+				struct passwd *pw;
+				pw = getpwnam(conf_value);
+				if (pw != NULL) {
+					pbs_servicename[0] = strdup(conf_value);
+					pbsservice[0] = pw->pw_uid;
+				}
+				else {
+					char * msgbuf;
```
bhroam, line 3233: Attach the * to the variable (here and all the other places)

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -11069,6 +11077,16 @@ def setup_sched_priv(self, sched_priv=None):
         self.holidays_obj = Holidays()
         self.holidays_parse_file(level=logging.DEBUG)
 
+    def setup_sched_logs(self):
+        if 'sched_log' in self.attributes:
+            sched_logs = self.attributes['sched_log']
+        else:
+            sched_logs = os.path.join(self.pbs_conf['PBS_HOME'],
+                                      'sched_logs')
+
```
bhroam, line 11086: Check if the path exists first like setup_sched_priv()

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -11132,8 +11150,12 @@ def start(self, sched_home=None, args=None, launcher=None):
...
                 cmd += ['-d', sched_home]
             try:
-                ret = self.du.run_cmd(self.hostname, cmd, sudo=True,
-                                      logerr=False, level=logging.INFOCLI)
+                if self.user.name == 'root':
+                    ret = self.du.run_cmd(self.hostname, cmd, sudo=True,
+                                          logerr=False, level=logging.INFOCLI)
+                else:
+                    ret = self.du.run_cmd(self.hostname, cmd, runas=self.user,
+                                          logerr=False, level=logging.INFOCLI)
```
bhroam, line 11158: Can't you combine these both and say runas=self.user.name?

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -11629,25 +11656,30 @@ def revert_to_defaults(self):
...
         self.signal('-HUP')
         # Revert fairshare usage
         cmd = [os.path.join(self.pbs_conf['PBS_EXEC'], 'sbin', 'pbsfs'), '-e']
         if self.sc_name is not 'default':
             cmd += ['-I', self.sc_name]
-        self.du.run_cmd(cmd=cmd, sudo=True)
+        if self.user.name == 'root':
+            self.du.run_cmd(cmd=cmd, sudo=True)
+        else:
+            self.du.run_cmd(cmd=cmd, runas=self.user)
```
bhroam, line 11682: Same as above, can't you just do the runas?

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -12616,7 +12655,28 @@ def query_fairshare(self, name=None, id=None):
         cmd = [os.path.join(self.pbs_conf['PBS_EXEC'], 'sbin', 'pbsfs')]
         if self.sc_name != 'default':
             cmd += ['-I', self.sc_name]
-        ret = self.du.run_cmd(self.hostname, cmd, sudo=True, logerr=False)
+
+        # env = ['PBS_DAEMON_SERVICE_USER=%s' % str(self.user)]
+        # self.logger.info(env)
+        # env.extend(cmd)
+        # cmd = env
+        # env = {'PBS_DAEMON_SERVICE_USER': str(self.user)}
```
bhroam, line 12663: Remove the commented out code

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -12616,7 +12655,28 @@ def query_fairshare(self, name=None, id=None):
...
+            env = '%s=%s' % (dsu, self.user.name)
+            cmd.insert(0, env)
+
+        self.logger.info(cmd)
+        if self.user.name == 'root':
+            ret = self.du.run_cmd(self.hostname, cmd=cmd,
+                                  sudo=True, logerr=False)
+        else:
+            ret = self.du.run_cmd(self.hostname, cmd=cmd,
+                                  runas=self.user, logerr=False)
```
bhroam, line 12678: Same as above, use runas (here and elsewhere)

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -785,6 +786,7 @@ def init_schedulers(cls, init_sched_func=None, skip=None):
                                         func=init_sched_func)
 
         for sched in cls.scheds.values():
+            # sched.pi.restart_sched(sched.hostname)
```
bhroam, line 789: Remove commented out code

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -805,6 +807,12 @@ def init_moms(cls, init_mom_func=None, skip='nomom'):
                                       func=init_mom_func)
         if cls.moms:
             cls.mom = cls.moms.values()[0]
+
+        for mom in cls.moms.values():
+            if not mom.isUp():
+                cls.logger.error('mom ' + mom.hostname + ' is down')
+                mom.pi.start_mom(mom.hostname)
```
bhroam, line 814: Why are we starting mom?  Also, if mom is down, mom.isUp() will wait like 30s for the mom to come up before returning.   You might want to give it a max_attempts

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -1228,7 +1242,15 @@ def _revert_pbsconf_server(self, vals_to_set):
...
-
+            if DAEMON_SERVICE_USER.name == 'root':
+                if "PBS_DAEMON_SERVICE_USER" in new_pbsconf:
+                    del(new_pbsconf['PBS_DAEMON_SERVICE_USER'])
+                    restart_pbs = True
+            elif (new_pbsconf["PBS_DAEMON_SERVICE_USER"] !=
+                  DAEMON_SERVICE_USER.name):
+                new_pbsconf["PBS_DAEMON_SERVICE_USER"] = \
+                    DAEMON_SERVICE_USER.name
+                restart_pbs = True
```
bhroam, line 1253: What about the case where daemon service user is not provided to benchpress?  Then it is not in pbs.conf by default.  Shouldn't we remove it here?

test/tests/functional/pbs_daemon_service_user.py
```diff
@@ -0,0 +1,130 @@
+# coding: utf-8
+
+# Copyright (C) 1994-2020 Altair Engineering, Inc.
+# For more information, contact Altair at www.altair.com.
+#
+# This file is part of the PBS Professional ("PBS Pro") software.
```
bhroam, line 6: This isn't the new license

test/tests/functional/pbs_daemon_service_user.py
```diff
@@ -0,0 +1,130 @@
...
+    """
+
+    def setUp(self):
+        TestFunctional.setUp(self)
+        self.orig_pbs_conf = self.du.parse_pbs_config(self.server.shortname)
+
+    def tearDown(self):
+        self.du.set_pbs_config(self.server.shortname,
+                               confs=self.orig_pbs_conf,
+                               append=False)
```
bhroam, line 56: Why do you need to do this?  Won't revert to defaults revert pbs.conf?

test/tests/functional/pbs_svr_dyn_res.py
```diff
@@ -596,7 +596,7 @@ def test_svr_dyn_res_permissions(self):
         self.scheduler.add_resource('foo')
 
         scr_body = ['echo "10"', 'exit 0']
-        home_dir = os.path.expanduser("~")
+        home_dir = os.path.join('/', 'tmp')
```
bhroam, line 599: why /tmp?  Also use os.sep, not '/'

src/scheduler/node_info.c
```diff
@@ -1174,135 +1147,6 @@ add_node_state(node_info *ninfo, char *state)
...
- *
- * @param[in,out]	ninfo	-	the node to to talk to its mom
- *
- * @return	int
- * @return	1	: on error
- * @return	0	: on success
- *
- */
-int
-talk_with_mom(node_info *ninfo)
```
bhroam, line 1189: I see all the rm stuff being removed (which includes load), so the load won't be set, but I don't see all the load code going away.  You need to remove load_balancing from conf.  You need to remove lowest_load from smp_cluster_dist.  You need to remove all the code that uses them.  You need to remove the load member from node_info and all the code which uses that.

test/tests/functional/pbs_svr_dyn_res.py
```diff
@@ -615,17 +615,17 @@ def test_svr_dyn_res_permissions(self):
 
         # give write permission to user only
         self.du.chmod(path=fp, mode=0o744, sudo=True)
-        if os.getuid() != 0:
-            self.check_access_log(fp, exist=True)
-        else:
-            self.check_access_log(fp, exist=False)
+        self.check_access_log(fp, exist=False)
```
bhroam, line 618: What about the case where daemon user  is root and os.getuid() is not?

test/tests/functional/pbs_daemon_service_user.py
```diff
@@ -0,0 +1,130 @@
...
+        self.assertIn(msg, '\n'.join(ret['err']))
+
+    def test_sched_runas_root(self):
+        """
+        Test if running sched as root with
+        PBS_DAEMON_SERVICE_USER set as another user
+        """
+        self.common_test('pbs_sched', ROOT_USER, TEST_USER,
+                         'Must be run by PBS_DAEMON_SERVICE_USER')
+
```
bhroam, line 89: You are testing a whole bunch of ways the scheduler doesn't start if it isn't the daemon user.  I don't see a test to see if it starts properly as the daemon user.

test/tests/functional/pbs_daemon_service_user.py
```diff
@@ -0,0 +1,130 @@
...
+    def test_sched_runas_nonroot_notset(self):
+        """
+        Test if running sched as nonroot with
+        PBS_DAEMON_SERVICE_USER not set
+        """
+        self.common_test('pbs_sched', TEST_USER, None,
+                         'Must be run by PBS_DAEMON_SERVICE_USER if '
+                         'set or root if not set')
```
bhroam, line 121: I don't get this message fully.  Don't we know if PBS_DAEMON_SERVICE_USER is set?  Why is this message different than the other message?

src/lib/Liblog/chk_file_sec.c
vstumpf: I just copied this from chk_file_sec
For something to be allowed, it must not be in disallow.

test/fw/ptl/lib/pbs_testlib.py
vstumpf: setup_sched_priv only checks the path before parsing the config, we don't do that.

test/fw/ptl/lib/pbs_testlib.py
vstumpf: if we `runas=root`, it runs the command as `sudo -u root qstat` instead of `sudo qstat`. I think our test harness handles these differently

test/fw/ptl/utils/pbs_testsuite.py
vstumpf: if it's not provided, we do remove it, check line 1246

test/tests/functional/pbs_svr_dyn_res.py
vstumpf: if the daemon user doesn't have read permissions to the current user, it won't be able to read the file.

test/tests/functional/pbs_daemon_service_user.py
vstumpf: the fact that other tests work should be proof positive that it works, right? i can add a simple test for it though

test/tests/functional/pbs_daemon_service_user.py
vstumpf: that's the log message that we're matching with. to give more context, i used the whole message to include it not being set.

test/tests/functional/pbs_svr_dyn_res.py
vstumpf: we don't call os.getuid(), i don't understand

src/lib/Liblog/chk_file_sec.c
bhroam: I still don't think I understand.  What do you mean by it must not be in disallow?

test/fw/ptl/lib/pbs_testlib.py
bhroam: Could you give it a try?  It'd clean up the code.

test/tests/functional/pbs_daemon_service_user.py
bhroam: I disagree.  The current tests will check if the scheduler will fail to start if the scheduler is not the right user.  It doesn't make sure the scheduler starts when it is the right user.  It might never start.

test/tests/functional/pbs_daemon_service_user.py
bhroam: I understand you are matching to this message, that still doesn't explain the message.  Maybe I should have given this comment up when you spit out the message.

test/tests/functional/pbs_svr_dyn_res.py
bhroam: Moving this to /tmp will cause chk_file_sec() to fail in a different way.  It will still not allow the script, but it will fail because /tmp has rwx for all perms.  The doc string wants it to fail based on the fact that the script itself has the wrong perms.

test/tests/functional/pbs_svr_dyn_res.py
bhroam: Look at the code you replaced.  It was an if statement with os.getuid() != NULL to call check_access_log() with exist=True.  Otherwise call it with exist=False.  You just dumped the first part of that if statement

test/tests/functional/pbs_daemon_service_user.py
vstumpf: revert to defaults doesn't revert conf soon enough

bhroam: How can it not?  Your test doesn't start until after the revert happens.

test/tests/functional/pbs_svr_dyn_res.py
vstumpf: i dumped it because we only cared about os.getuid() when we used the current user's home directory. 
Since i changed it to /tmp,

test/tests/functional/pbs_daemon_service_user.py
vstumpf: ptl fails to query fairshare data because it's trying to use the old pbs_daemon_Service_user value

bhroam: So revert pbs.conf sooner.  It should probably be the first thing reverted anyway

src/lib/Liblog/chk_file_sec.c
```diff
@@ -174,32 +174,41 @@ tempstat(struct stat *sp, int isdir, int sticky, int disallow, int uid)
 	if ((~disallow & S_IWUSR) && (sp->st_uid > 10 && !(uid > 0 && uid == sp->st_uid))) {
 		/* Owner write is allowed, and UID is greater than 10 or owner is not self. */
 		rc = EPERM;
+		log_errf(rc, "tempstat", "BAD USER");
```
bhroam, line 177: Don't  say "tempstat", say \_\_func\_\_

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -808,10 +807,10 @@ def init_moms(cls, init_mom_func=None, skip='nomom'):
...
             cls.mom = cls.moms.values()[0]
 
-        for mom in cls.moms.values():
-            if not mom.isUp():
-                cls.logger.error('mom ' + mom.hostname + ' is down')
-                mom.pi.start_mom(mom.hostname)
+        # for mom in cls.moms.values():
+        #     if not mom.isUp():
+        #         cls.logger.error('mom ' + mom.hostname + ' is down')
+        #         mom.pi.start_mom(mom.hostname)
```
bhroam, line 813: If you don't need this code, remove it

doc/man8/pbs_sched.8B
```diff
@@ -473,28 +473,6 @@ Default: Not in force
...
-
-.IP load_balancing_rr 13
-.B Deprecated.
-To duplicate this setting, enable
-.I load_balancing
-and
-.I set smp_cluster_dist
-to
-.I round_robin.
-
```
bhroam, line 497: Make sure to file a doc ticket  for this

src/lib/Libifl/auth.c
```diff
@@ -280,6 +280,10 @@ load_auths(int mode)
 	if (loaded_auths != NULL)
 		return 0;
 
+	if (pbs_conf.auth_method[0] == '\0' && pbs_conf.encrypt_method[0] == '\0') {
```
hirenvadalia, line 283: Not sure but this looks wrong: because as per current pbs_conf's default value and pbs_loadconf(), pbs_conf.auth_method will never be the empty string, it will be either "resvport" (default value) or user/admin provided string. So if you are getting pbs_conf.auth_method[0] == '\0' then it will some bug somewhere in code.

vstumpf: you're right, this was leftover for when i was doing testing. I'll remove it

src/lib/Liblog/chk_file_sec.c
vstumpf: sorry, this is just for debugging things

src/lib/Liblog/chk_file_sec.c
```diff
@@ -116,12 +117,12 @@ teststat(struct stat *sp, int isdir, int sticky, int disallow,
...
 
 static int
-teststat(struct stat *sp, int isdir, int sticky, int disallow)
+teststat(struct stat *sp, int isdir, int sticky, int disallow, int uid)
 {
 	int	rc = 0;
 
-	if ((~disallow & S_IWUSR) && (sp->st_uid > 10)) {
-		/* Owner write is allowed, and UID is greater than 10. */
+	if ((~disallow & S_IWUSR) && (sp->st_uid > 10 && !(uid > 0 && uid == sp->st_uid))) {
```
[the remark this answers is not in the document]
vstumpf: Yes, because when we check /var/spool/pbs/sched_priv, /var/spool/pbs's owner is root. If i didn't leave this, sched_priv/logs would have to be in directories owned by PBS_DAEMON_SERVICE_USER

src/lib/Liblog/chk_file_sec.c
vstumpf: If OtherWrite (S_IWOTH) is not passed into disallow
		 * Other write is allowed, and at least one of the following
		 * is true:
		 * - target is not a directory
		 * - target does not have sticky bit set
		 * - the value of the sticky argument we were passed was zero

test/tests/functional/pbs_daemon_service_user.py
vstumpf: ptl fails to delete all the jobs in teardown
In Server.cleanup_jobs, we status the server for all the schedulers. this causes us to create a new scheduler object in ptl, which calls pbsfs. this will fail because the conf file has PDSU set.

src/lib/Liblog/chk_file_sec.c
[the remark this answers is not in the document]
arungrover: But isn't it wrong that if st_uid is <=10 then it will not give EPERM error even when daemon service user uid isn't the matching with st_uid?

test/tests/functional/pbs_daemon_service_user.py
bhroam: You should fix this in setUp().  If you fix it this way, you are forcing any test which uses PDSU to revert it in tearDown().  This is PBS functionality, and it should be reverted properly.

src/lib/Liblog/chk_file_sec.c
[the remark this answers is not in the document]
vstumpf: I'm not sure why we trust uids <=10, but we do. I only add the case were we dont fail if `uid == sp->st_uid`.
I really don't see another way around this.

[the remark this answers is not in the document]
bhroam: Take a gander at your /etc/passwd file.  The UIDs under 10 are system accounts.  It probably was more so 20 years back.

test/tests/functional/pbs_daemon_service_user.py
bhroam: I don't see how we made it all the way to tearDown() without creating  all the schedulers.  They should have either been made in setUp() or in the test if it added any.  Why are we creating any in tearDown()?

vstumpf: In teardown, the first thing we do is call `self.server.cleanup_jobs()`.
This calls `self.status(SCHED)`, which creates scheduler objects from the info returned by the status call.
In Scheduler's constructor, we query fairshare. The problem is in the test, we have set a custom PBS_DAEMON_SERVICE_USER in /etc/pbs.conf, and when we try to run pbsfs, pbsfs complains that the user we're running as (which is root by default) doesn't match PDSU.

openpbs.spec
```diff
@@ -510,6 +510,7 @@ ${RPM_INSTALL_PREFIX:=%{pbs_prefix}}/libexec/pbs_posttrans \
 %exclude %{pbs_prefix}/libexec/pbs_db_utility
 %exclude %{pbs_prefix}/sbin/pbs_comm
 %exclude %{pbs_prefix}/sbin/pbs_dataservice
+%exclude %{pbs_prefix}/sbin/pbs_systemd
```
arungrover, line 513: You can remove this change. @ashwathraop confirmed that pbs_systemd shouldn't have been in openpbs.spec.in file to begin with. I'm addressing it in one of my PR.

src/scheduler/fifo.c
```diff
@@ -2984,3 +2986,42 @@ set_validate_sched_attrs(int connector)
...
+ *
+ * @retval Error code
+ * @return 0 - Failure
+ * @return 1 - Success
+ *
+ * @par Side Effects:
+ *	None
+ */
+int
+validate_running_user(char * exename) {
```
arungrover, line 3004: Can this function move to more generic part of PBS? It seems to confirm daemon service user might be useful for other components too.

vstumpf: I think i'd prefer to leave it here until another daemon actually uses it. Hopefully by then we'll have OOP and a daemon class to put it in.

test/tests/functional/pbs_daemon_service_user.py
```diff
@@ -131,3 +121,14 @@ def test_pbsfs_runas_nonroot_notset(self):
...
+    def test_sched_runas_nonroot_pass(self):
+        """
+        Test if sched runs as non-root user
+        """
+        self.scheduler.stop()
+        self.common_test('pbs_sched', TEST_USER, TEST_USER, None,
+                         setup_sched=True)
+        j = Job(TEST_USER1)
+        jid = self.server.submit(j)
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid)
```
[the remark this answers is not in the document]
vstumpf: I don't htink it adds anything, if someone wanted ot they can run the multi-sched test suite with the daemon-user param.

src/scheduler/fifo.c
```diff
@@ -3018,6 +3019,7 @@ validate_running_user(char * exename) {
 
 		if (geteuid() == 0) {
 			setuid(user->pw_uid);
+			strncpy(pbs_current_user, pbs_conf.pbs_daemon_service_user, PBS_MAXUSER);
```
arungrover, line 3022: please change this to pbs_strncpy :)

bhroam:
The only part of this PR that I don't understand is the moving of pbs_current_user.  I'd like someone who knows the ECL threading portion of the code to review it.  Maybe @hirenvadalia or @subhasisb could take a look? Just look at the second to last commit named "Move pbs_current_user"

subhasisb:
Thanks for the nudge @bhroam I took a peek at the said commit and it looks fine to me. I think the current_user being part of the the TLS of the client thread is more of a vestigial thing than anything else. @vstumpf  is removing that and moving the current_user information into the pbs_conf structure, which is fine.

When we refactor things to be better, one of the things we need to be careful about is storing configuration separate from "state", we have currently often mixed configuration (read-only stuff) with state (volatile stuff). Storing the current_user in pbs_conf is doing that "not so good design" thing as well, but that is probably okay for now.

PR #1787, opened 2020-06-01
pbs_topologyinfo does not work in Linux Windows Interoperability

npadole20:
#### Describe Bug or Feature

pbs_topologyinfo fails for windows node's topology files because it fails to identify the node's platform correctly as we use platform.system() to identify the platform. Since the server is Linux, we identify the platform as Linux always. 

#### Describe Your Change

Instead of using platform.system(), check if topology data is in xml format by looking for '<'.  XML format is used only for Linux nodes and will not be present in the windows node's topology file.

[topofix.txt](https://github.com/openpbs/openpbs/files/4725315/topofix.txt)

### DIFF
--- a/src/cmds/scripts/pbs_topologyinfo.py
+++ b/src/cmds/scripts/pbs_topologyinfo.py
@@ -70,7 +70,7 @@ class Inventory(object):
         """
         counting devices by parsing topo_file
         """
-        temp = topo_file.split(',')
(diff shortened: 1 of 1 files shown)


nithinj:
Why not check for XML type instead of hwloc?

npadole20:
> 
> 
> Why not check for XML type instead of hwloc?

Changed the code to check for XML. Thank you. Please review the latest changes @nithinj

PR #1794, opened 2020-06-02
Add logs to error path in Libwin util.c

Shrini-h:
#### Describe Bug or Feature

some error paths were not logged in src/lib/Libwin/util.c, making it difficult to debug rare issues on windows nodes.

#### Describe Your Change

1. added logs to error paths
2. wrapped `revert_impersonated_user()` to print error log
3. removed unused functions `print_arg_param()` and `fix_temp_path`

#### Attach Test Logs/Output

[Testing_logs.txt](https://github.com/openpbs/openpbs/files/4736052/Testing_logs.txt)


src/resmom_win/start_exec.c
```diff
@@ -1768,7 +1767,7 @@ finish_exec(job *pjob)
 	if (pjob->ji_user->pw_userlogin != INVALID_HANDLE_VALUE) {
 		if (!impersonate_user(pjob->ji_user->pw_userlogin)) {
 			sprintf(log_buffer,
-				"failed to ImpersonateLoggedOnUser on %s", mom_host);
+				"failed to ImpersonateLoggedOnUser on %s, error=%ld", mom_host, GetLastError());
```
Shrini-h, line 1770: NOTE to reviewers: This change will be reflected in the comment field of held windows jobs. I wanted to capture logon failure codes in the comment itself, but looks like the the logon failure codes appear in logon_pw and the pw_userlogin will be null. so in reality I'm not getting the correct code in the comment, but just the code 87 for ERROR_INVALID_PARAMETER which I'm setting explicitly in impersonate_user as part of this PR.

```
Job ID                         Username        Queue           Jobname         SessID   NDS  TSK   Memory Time  S Time
------------------------------ --------------- --------------- --------------- -------- ---- ----- ------ ----- - -----
7.server                         pbsuser         workq           STDIN                --     1     1    --    --  H   --
   failed to ImpersonateLoggedOnUser on winmom1, error=87:job has bad password
```

So Thinking of reverting this line. please give your opinions

src/lib/Libwin/util.c
```diff
@@ -744,7 +704,6 @@ win_open(const char *filename, int oflag, ...)
 	va_end(vl);
 
 	if ((err) || (fd == -1)) {
-		errno = err;
```
SudeshnaMoulik, line 747: Why are we removing this? Caller of open() and create() in Windows may check errno for validation

src/lib/Libwin/util.c
```diff
@@ -951,32 +910,41 @@ impersonate_user(HANDLE hlogintoken)
...
+ * 
+ * @param[in]   funcname : function name where API was invoked
+ * @param[in]   lineno : line number of API invocation
  *
  *  @return     BOOL
  *  @retval     whether revert succeded, return value of RevertToSelf() API
  */
 BOOL
-revert_impersonated_user()
+_log_wrap_revert_impersonated_user(LPCSTR funcname, INT lineno)
```
SudeshnaMoulik, line 937: Looks like we are removing "revert_impersonated_user" then why this function name is "_log_wrap_revert_impersonated_user" ?

src/include/win.h
```diff
@@ -315,7 +315,11 @@ extern int	get_dcinfo(char *net_name, char domain_name[], char domain_ctrl[]);
 extern SID*	create_everyone_sid(void);
 extern HANDLE LogonUserNoPass(char *username);
 extern BOOL impersonate_user(HANDLE hlogintoken);
-extern BOOL revert_impersonated_user();
+
+/* wrap revert_impersonated_user to log in case of error */
+#define revert_impersonated_user() _log_wrap_revert_impersonated_user(__func__, __LINE__)
+extern BOOL _log_wrap_revert_impersonated_user(LPCSTR, INT);
```
SudeshnaMoulik, line 321: Instead of creating another wrapper of a non-existent function, why don't we modify the existing ?
Noticed later revert_impersonated_user() removed

src/resmom_win/start_exec.c
SudeshnaMoulik: In my opinion you can modify the if check and add "pjob->ji_user->pw_userlogin != NULL" and before the else block can verify for the same and log accordingly. And also keep the current else block to cover the situation when we get a valid handle but Windows API failed.
Also can keep this block's error login message as it was previously.

src/lib/Libwin/util.c
Shrini-h: Only the definition header is renamed, so as to be helpful with macro replacement, the invocation will still be called as `revert_impersonated_user`. This is to get the function name and line location of invocation of `revert_impersonated_user` for error logging. The `_log_wrap_` is as per convention followed for wrapping `CloseHandle` and `LocalFree` in win.h ,  hit +1 if you agree.

src/include/win.h
Shrini-h: Only the definition header is renamed, so as to be helpful with macro replacement, the invocation will still be called as `revert_impersonated_user`. This is to get the function name and line location of invocation of `revert_impersonated_user` for error logging. The `_log_wrap_` is as per convention followed for wrapping `CloseHandle` and `LocalFree` in win.h ,  hit +1 if you agree.

src/resmom_win/start_exec.c
Shrini-h: fixed in https://github.com/openpbs/openpbs/pull/1794/commits/5ca363a538c69345193a7ce576461e4932416e0b
Please +1 if you agree to the fix

SudeshnaMoulik: Do you think it will be better, if we add an "else if block" after this to verify whether the handle is NULL and log accordingly ?

src/include/win.h
SudeshnaMoulik: Actually both way we can handle this either rename the definition header or modify only the error logging inside the existing function. Up to you

src/resmom_win/start_exec.c
Shrini-h: I don't think its better, because both `NULL` and `INVALID_HANDLE_VALUE` are non valid values and we never differentiate them from each other while logging in many other places, I just wish the original code was more confident to check for just either `NULL` or `INVALID_HANDLE_VALUE`, so I doubt it will add any more helpful info. over that the code will look ugly. +1 if you agree

src/include/win.h
Shrini-h: the problem with keeping the same name in the definition header is, due to the introduced macro, we will get compiler error, to solve it we will have to end up doing the ugly `#undef` before the function and then again re define the macro after the function, so I thought its better to keep it with different name. +1 if you agree

src/resmom_win/start_exec.c
subhasisb: If this additional error code is not the right one, or even adding any more value, then why add this now - we should revert back to what it was

Shrini-h: reverted it and added an error log instead to log the error code. https://github.com/openpbs/openpbs/pull/1794/commits/aa46739ce40d19359d847bc5530e632a383fa4c4

src/resmom_win/start_exec.c
```diff
@@ -3465,7 +3467,8 @@ open_std_file(job *pjob, enum job_file which, int mode, gid_t exgid)
 
 		if (became_admin) {   /* go back to being user */
 			if (pjob->ji_user->pw_userlogin != INVALID_HANDLE_VALUE) {
-				(void)impersonate_user(pjob->ji_user->pw_userlogin);
+				if (impersonate_user(pjob->ji_user->pw_userlogin) == FALSE)
+					log_joberr(-1, __func__, "impersonate_user() failed", pjob->ji_qs.ji_jobid);
```
subhasisb, line 3471: Don't need the user (pw_name) to be printed here?

src/server/job_func.c
```diff
@@ -519,8 +519,10 @@ job_free(job *pj)
 			if ((pj->ji_wattr[JOB_ATR_euser].at_val.at_str) &&
 				(pwdp = getpwnam(pj->ji_wattr[JOB_ATR_euser].at_val.at_str))) {
 				if (pwdp->pw_userlogin != INVALID_HANDLE_VALUE) {
-					if (impersonate_user(pwdp->pw_userlogin) == 0)
+					if (impersonate_user(pwdp->pw_userlogin) == 0) {
+						log_joberr(-1, __func__, "impersonate_user() failed", pj->ji_qs.ji_jobid);
```
subhasisb, line 523: same - dont u want to log the username for whom the impersonation failed?

src/resmom_win/start_exec.c
Shrini-h: fixed in https://github.com/openpbs/openpbs/pull/1794/commits/90ed7603b8b003d419447d749bc57f0e30543573

src/server/job_func.c
Shrini-h: fixed in https://github.com/openpbs/openpbs/pull/1794/commits/90ed7603b8b003d419447d749bc57f0e30543573

PR #1795, opened 2020-06-03
Server gives false positive for pbs_ralter cmd

arungrover:
#### Describe Bug or Feature
There are multiple problems addressed in this PR
- When a confirmed reservation is altered and ralter fails, it gives a false positive to the user stating the changes were confirmed.
- When a degraded reservation is altered, the server does not log a message in PBS accounting logs. 
- The third problem was in a test that expected ralter to fail in a case it shouldn’t.

#### Describe Your Change
- This issue happens when alter is issues with the ‘Interactive’ attribute. The problem happens because the server had a misplaced check of checking for the reservation to be NOT CONFIRMED before replying that it is denied. This check was wrongly placed.
- Second problem was fixed by updating a flag server checks to ensure that it does not log ralter accounting messages for degraded reservations.
- A test was modified to ensure that it accounts for free nodes in the system and expect the reservation ralter to succeed in such cases

#### Link to Design Doc
NA

[test_resv.txt](https://github.com/openpbs/openpbs/files/4720534/test_resv.txt)
[test_ralter.txt](https://github.com/openpbs/openpbs/files/4720536/test_ralter.txt)
[test_multi_sched.txt](https://github.com/openpbs/openpbs/files/4720537/test_multi_sched.txt)
I added no new tests. Following are the tests from ralter test suite that were failing - 
test_alter_degraded_resv_mom_down, test_conflict_two_advance_resvs, test_conflict_two_standing_resvs

### DIFF
--- a/src/server/req_rescq.c
+++ b/src/server/req_rescq.c
@@ -586,7 +586,7 @@ req_confirmresv(struct batch_request *preq)
 				log_event(PBSEVENT_DEBUG2, PBS_EVENTCLASS_RESV, LOG_NOTICE, presv->ri_qs.ri_resvID, log_buffer);
 			}
 		} else {
-			if ((presv->rep_sched_count >= presv->req_sched_count) && !is_confirmed) {
+			if (presv->rep_sched_count >= presv->req_sched_count) {
 				/* Clients waiting on an interactive request must be
 				* notified of the failure to confirm
 				*/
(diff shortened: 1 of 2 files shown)


arungrover:
Thanks @agrawalravi90 
I have also changed design doc - https://pbspro.atlassian.net/wiki/spaces/PD/pages/1401815043/Reconfirming+degraded+reservations+that+are+running
to mention that going forward we will be logging a 'Y' record whenever a degraded reservation is reconfirmed. Please have a look.

PR #1797, opened 2020-06-03
[WIP]Added back the closeHandle in get_processowner

SudeshnaMoulik:
#### Describe Bug or Feature
As part of [#1738 ](https://github.com/openpbs/openpbs/pull/1738) while adding debug log in error path, removed CloseHandle() by mistake from get_processowner() of passwd.c

#### Describe Your Change
Added back the same in error path before returning from the function.

### DIFF
--- a/src/lib/Libwin/passwd.c
+++ b/src/lib/Libwin/passwd.c
@@ -1040,7 +1040,7 @@ getusersid2(char *uname,
 
 			/* get the size of the memory buffer needed for the SID */
 			ret = GetTokenInformation(hToken, TokenUser, NULL, 0, &dwBufferSize);
-			if ((ret ==0) && (GetLastError() == ERROR_INSUFFICIENT_BUFFER)) {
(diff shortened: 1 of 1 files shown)


src/lib/Libwin/passwd.c
```diff
@@ -5775,6 +5775,7 @@ get_processowner(DWORD processid, uid_t *puid, char *puname, size_t uname_len, c
 	hProcess = OpenProcess(PROCESS_ALL_ACCESS, FALSE, processid);
 	if ((hProcess == INVALID_HANDLE_VALUE) || (hProcess == NULL)) {
 		log_errf(-1, __func__, "failed in OpenProcess for process: %lu", processid);
+		CloseHandle(hProcess);
```
subhasisb, line 5778: if hProcess is NULL or INVALID_HANDLE_VALUE do you really need to call CloseHandle()? Is that call even valid?

SudeshnaMoulik: Thanks @subhasisb for clarifying the doubt. Will close this PR.

PR #1798, opened 2020-06-03
Fix test_runjob_hook log match failure caused by regular expression

minghui-liu:
#### Describe Bug or Feature

test_runjob_hook fails to log match `msg_4` if the exec_vnode name contains a pair of square brackets. Square brackets are used in regular expression to specify a set of allowed characters. They need to be escaped. 

#### Describe Your Change

Escape square brackets using `replace`. 

#### Link to Design Doc

N/A

[pbi-23-r7p1 build and test.txt](https://github.com/openpbs/openpbs/files/4734719/pbi-23-r7p1.build.and.test.txt)
[pbi-24-s12 build and test.txt](https://github.com/openpbs/openpbs/files/4734717/pbi-24-s12.build.and.test.txt)

[uv-01.txt](https://github.com/openpbs/openpbs/files/4730704/uv-01.txt)
[uv-02-s12.txt](https://github.com/openpbs/openpbs/files/4730705/uv-02-s12.txt)

### DIFF
--- a/test/tests/functional/pbs_hooksmoketest.py
+++ b/test/tests/functional/pbs_hooksmoketest.py
@@ -454,19 +454,15 @@ print_attribs(j)"""
         # Verify server logs
         self.logger.info("Verifying logs in server")
         msg_1 = "Server@.*;Hook;%s;started" % \
-                (self.hook_name)
+                (re.escape(self.hook_name))
         msg_2 = "Hook;Server@.*;requestor = Scheduler"
         msg_3 = "Hook;Server@.*;hook_name = %s" % \
-                (self.hook_name)
+                (re.escape(self.hook_name))
         msg_4 = "Hook;Server@.*;exec_vnode = %s" % \
-                (ev)
+                (re.escape(ev))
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_hooksmoketest.py
```diff
@@ -463,6 +463,8 @@ def print_attribs(pbs_obj):
         # Character escaping '()' as the log_match is regexp
         msg_4 = msg_4.replace("(", "\(")
         msg_4 = msg_4.replace(")", "\)")
+        msg_4 = msg_4.replace("[", "\[")
+        msg_4 = msg_4.replace("]", "\]")
```
bhroam, line 467: What calling re.escape(msg_4)?  That does all the regex escaping you need.

minghui-liu: Good suggestion. Fixed.

minghui-liu: Actually `re.escape(msg_4)` does not work because `msg` need to remain a regular express. But `ev` needs to be escaped.

PR #1799, opened 2020-06-03
TestSchedPreemptEnforceResumption.test_suspended_job_ded_time_calendared failed due to race condition.

dinesh1306:
#### Describe Bug or Feature
TestSchedPreemptEnforceResumption.test_suspended_job_ded_time_calendared failed due to race condition.
in test we are calculating temp value which will be used for start time of dedicated time and walltime of job, race condition such that sometime job will submitted after 1 sec. and walltime value is than 1 sec of dedicated start time and job remain in Q state due to test got failed

#### Describe Your Change
update code with submitting job with particular (i.e. 180 sec.) sleep time value.

before_fix:
[res_test_suspended_job_ded_time_calendared_before_fix.txt](https://github.com/openpbs/openpbs/files/4723216/res_test_suspended_job_ded_time_calendared_before_fix.txt)

After_fix:
i ran test multiple times and it passed.
[res_after_fix_1.txt](https://github.com/openpbs/openpbs/files/4723222/res_after_fix_1.txt)
[res_after_fix_2.txt](https://github.com/openpbs/openpbs/files/4723223/res_after_fix_2.txt)
[res_after_fix_3.txt](https://github.com/openpbs/openpbs/files/4723225/res_after_fix_3.txt)
[res_after_fix_4.txt](https://github.com/openpbs/openpbs/files/4723226/res_after_fix_4.txt)
[res_after_fix_5.txt](https://github.com/openpbs/openpbs/files/4723227/res_after_fix_5.txt)
[res_after_fix_6.txt](https://github.com/openpbs/openpbs/files/4723228/res_after_fix_6.txt)
[res_after_fix_7.txt](https://github.com/openpbs/openpbs/files/4723229/res_after_fix_7.txt)
[res_after_fix_8.txt](https://github.com/openpbs/openpbs/files/4723230/res_after_fix_8.txt)
[res_after_fix_9.txt](https://github.com/openpbs/openpbs/files/4723231/res_after_fix_9.txt)
[res_after_fix_10.txt](https://github.com/openpbs/openpbs/files/4723232/res_after_fix_10.txt)
[res_after_fix_11.txt](https://github.com/openpbs/openpbs/files/4723233/res_after_fix_11.txt)
[res_after_fix_12.txt](https://github.com/openpbs/openpbs/files/4723234/res_after_fix_12.txt)
[res_after_fix_13.txt](https://github.com/openpbs/openpbs/files/4723235/res_after_fix_13.txt)
[res_after_fix_14.txt](https://github.com/openpbs/openpbs/files/4723237/res_after_fix_14.txt)
[res_after_fix_15.txt](https://github.com/openpbs/openpbs/files/4723238/res_after_fix_15.txt)

### DIFF
--- a/test/tests/functional/pbs_sched_preempt_enforce_resumption.py
+++ b/test/tests/functional/pbs_sched_preempt_enforce_resumption.py
@@ -105,10 +105,9 @@ class TestSchedPreemptEnforceResumption(TestFunctional):
         end = start + 120
         self.scheduler.add_dedicated_time(start=start, end=end)
 
-        temp += 180
         j1 = Job(TEST_USER)
         j1.set_attributes({ATTR_l + '.select': '1:ncpus=2',
-                           ATTR_l + '.walltime': temp-10})
+                           ATTR_l + '.walltime': start-int(time.time())-10})
         jid1 = self.server.submit(j1)
         self.server.expect(JOB, {ATTR_state: 'R'}, id=jid1)
 
--- a/test/tests/functional/pbs_sched_preempt_enforce_resumption.py
+++ b/test/tests/functional/pbs_sched_preempt_enforce_resumption.py
@@ -105,10 +105,9 @@ class TestSchedPreemptEnforceResumption(TestFunctional):
         end = start + 120
         self.scheduler.add_dedicated_time(start=start, end=end)
 
-        temp += 180
         j1 = Job(TEST_USER)
         j1.set_attributes({ATTR_l + '.select': '1:ncpus=2',
-                           ATTR_l + '.walltime': temp})
+                           ATTR_l + '.walltime': 180})
         jid1 = self.server.submit(j1)
         self.server.expect(JOB, {ATTR_state: 'R'}, id=jid1)
 
--- a/test/tests/functional/pbs_sched_preempt_enforce_resumption.py
+++ b/test/tests/functional/pbs_sched_preempt_enforce_resumption.py
@@ -105,9 +105,10 @@ class TestSchedPreemptEnforceResumption(TestFunctional):
         end = start + 120
         self.scheduler.add_dedicated_time(start=start, end=end)
 
+        temp += 180
         j1 = Job(TEST_USER)
         j1.set_attributes({ATTR_l + '.select': '1:ncpus=2',
-                           ATTR_l + '.walltime': 180})
+                           ATTR_l + '.walltime': temp-10})
         jid1 = self.server.submit(j1)
         self.server.expect(JOB, {ATTR_state: 'R'}, id=jid1)
 


test/tests/functional/pbs_sched_preempt_enforce_resumption.py
```diff
@@ -105,10 +105,9 @@ def test_suspended_job_ded_time_calendared(self):
         end = start + 120
         self.scheduler.add_dedicated_time(start=start, end=end)
 
-        temp += 180
         j1 = Job(TEST_USER)
         j1.set_attributes({ATTR_l + '.select': '1:ncpus=2',
-                           ATTR_l + '.walltime': temp})
+                           ATTR_l + '.walltime': 180})
```
arungrover, line 110: I think this test will fail when on line 103 value of temp is evaluated to be 60. That would mean dedicated time will start 4 minutes in the future. low priority job will run for 3 mins. High priority job will preempt the low priority job for 60 seconds and then low priority job will resume back for the remainder of the time. 

Line 103 is causing the problem in this test. I don't think the test needs the complication of doing "now % 60" to start a dedicated time in the future. 
I'd say just, keep your current changes and remove line 103, update start time to something like "now +240" and that's it.

bhroam: Dedicated time starts on the minute boundary.  That's what the now % 60 is for.  It's to floor it to the minute boundary.

arungrover: Oh ok, I remember now, dedicated time does not take in seconds. In that case, how about increasing the walltime of J1 by a second and make it 181.

bhroam: If we want a job that will conflict with dedicated time, I'd say 200 or more.

arungrover: hmmm... that may not work because the job may not even start if there is a dedicated time starting in less than 200 seconds. How about instead of giving it a hardcoded value, just make walltime of the job as "temp -5" which will mean the job will end 5 seconds earlier than the dedicated time. It will also result in conflict when this job is suspended and resumed after 60 seconds.

dinesh1306: Hi @arungrover , if we use temp-5 according to me it wont work because temp value in between 0-60 and if we use it then job J1 will suspend and J2 will resume but J2 is only 60 sec. and dedicated start time will start at least after 180 sec. so J1 will resume before start_time and test will fail at line 133.

if we put walltime value less than 180 sec. like 170 sec. or 175 sec. then as mentioned by you if we get temp value 0 then also dedicated start_time will start after 180 sec. from current time and J1 will comes in R state. test will not fail at line 133

arungrover: ya, for it to work you will have to retain line 108 that you removed in this PR.

test/tests/functional/pbs_sched_preempt_enforce_resumption.py
```diff
@@ -105,9 +105,10 @@ def test_suspended_job_ded_time_calendared(self):
         end = start + 120
         self.scheduler.add_dedicated_time(start=start, end=end)
 
+        temp += 180
         j1 = Job(TEST_USER)
         j1.set_attributes({ATTR_l + '.select': '1:ncpus=2',
-                           ATTR_l + '.walltime': 180})
+                           ATTR_l + '.walltime': temp-10})
```
bhroam, line 111: This is still a race.  If a slow machine takes more than 10s to submit this job (unlikely, but out testbed machines are really slow), you should make this start - int(time.time()) - 10.  This takes into account any time passed between when we create temp and when we set the walltime.

dinesh1306:
@bhroam , I have incorporate review comment. Please have a look.
[res_after_change_1.txt](https://github.com/openpbs/openpbs/files/4737641/res_after_change_1.txt)
[res_after_change_2.txt](https://github.com/openpbs/openpbs/files/4737642/res_after_change_2.txt)
[res_after_change_3.txt](https://github.com/openpbs/openpbs/files/4737643/res_after_change_3.txt)
[res_after_change_4.txt](https://github.com/openpbs/openpbs/files/4737644/res_after_change_4.txt)
[res_after_change_5.txt](https://github.com/openpbs/openpbs/files/4737645/res_after_change_5.txt)

PR #1801, opened 2020-06-04
test_execjob_abort_ms_prologue fails because it takes longer for a jo…

lisa-altair:
…b to held when cgroups is enabled

#### Describe Bug or Feature

The TestPbsExecjobAbort.test_execjob_abort_ms_prologue  test is failing because it expects the job to be in the "H" state.   To be in the Held state, the job has to try to run at least 21 times.
But after checking the state for the default 60 times, the job had not been rerun 21 times on machines where the croups hook was enabled.  

I believe this is because the cgroups hook is registered for a lot of hook events, and that causes the job to go through the job path a little slower than when there are fewer hook events that it needs to go through.

#### Describe Your Change

 Increased the test's max_attempts to be 100.  This gives slower machines time to run the job enough times to have it get held.  I didn't want to increase the interval because that would make the test run slower on faster machine that were not seeing this problem.

[before.txt](https://github.com/openpbs/openpbs/files/4726749/before.txt)
[after.txt](https://github.com/openpbs/openpbs/files/4726748/after.txt)

### DIFF
--- a/test/tests/functional/pbs_hook_execjob_abort.py
+++ b/test/tests/functional/pbs_hook_execjob_abort.py
@@ -141,7 +141,7 @@ time.sleep(2)
         for mom in self.moms.values():
             mom.log_match(msg, starttime=start_time)
         self.server.expect(JOB, {ATTR_state: 'H'}, id=jid, max_attempts=100,
-            offset=7)
+                offset=7)
 
     def test_execjob_abort_exit_job_launch_reject(self):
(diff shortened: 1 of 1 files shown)


bhroam:
@lisa-altair Travis failed due to pep8.  Once you fix that, I'l merge it.

lisa-altair:
@bhroam I fixed the pep8

PR #1802, opened 2020-06-04
skipOnCpuSet decorator should not use mom object and work on class and test levels.

vccardenas:
#### Describe Bug or Feature
skipOnCpuSet decorator does not use mom object.

#### Describe Your Change
The decorator @skipOnCpuSet can be used on the class and test levels without need for the mom object being created first.

Tests are not skipped on non cpuset system:
[TestEquivClass_per_noncset_4.txt](https://github.com/openpbs/openpbs/files/4760648/TestEquivClass_per_noncset_4.txt)

Skip a test:
[TestEquivClass_per_test_4.txt](https://github.com/openpbs/openpbs/files/4760650/TestEquivClass_per_test_4.txt)

Skip all tests if decorator is used on the class:
[TestEquivClass_per_class_4.txt](https://github.com/openpbs/openpbs/files/4760651/TestEquivClass_per_class_4.txt)

### DIFF
--- a/test/tests/functional/pbs_equiv_classes.py
+++ b/test/tests/functional/pbs_equiv_classes.py
@@ -48,6 +48,12 @@ class TestEquivClass(TestFunctional):
     """
 
     def setUp(self):
+        # Skip the test on cgroup cpuset system
+        hpe_file1 = "/etc/sgi-compute-node-release"
+        hpe_file2 = "/etc/sgi-known-distributions"
+        if os.path.isfile(hpe_file1) or os.path.isfile(hpe_file2):
+            msg = 'test is not supported on cgroup cpuset system'
+            raise SkipTest(msg)
         TestFunctional.setUp(self)
         a = {'resources_available.ncpus': 8}
         self.server.create_vnodes('vnode', a, 1, self.mom, usenatvnode=True)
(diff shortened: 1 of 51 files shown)


test/tests/functional/pbs_equiv_classes.py
```diff
@@ -48,6 +48,12 @@ class TestEquivClass(TestFunctional):
     """
 
     def setUp(self):
+        # Skip the test on cgroup cpuset system
```
bayucan, line 51: Rather than put these test for cgroup cpuset system here, I suggest moving them to a function in class TestFunctional (PBSTestsuite) (test/fw/lib/pbs_testsuite.py). So then it can be called later from other testsuites, not this one. If it's not too much trouble, perhaps merge the common codes in class MoM.is_cpuset_mom() with this this TestFunctional.is_cpuset(hostname>...

vccardenas:
The changes to the decorator now takes an argument.

vccardenas:
The requirements() and skip() decorators both take arguments.  My changes to the skipOnCpuSet decorator also takes arguments.  To use these decorators on the class level we need to provide an argument.

lisa-altair:
Unfortunately, this won't work on multi-node tests where one of the sister moms (not where PTL is running) is a cpuset mom.  That means tests that should be skipped on a multi-node situation (and are skipped without this change) won't be skipped.

vccardenas:
Closing this PR in favor of a different approach in a future PR.

PR #1805, opened 2020-06-04
only_explicit_psets not working with -lplace=group

arungrover:
#### Describe Bug or Feature
PBS scheduler does not seem to honor only_explicit_psets setting when jobs are submitted with -lplace=group=<res>.

#### Describe Your Change
There was a mistake in how only_explicit_psets condition was checked.

#### Link to Design Doc
NA

[test_place.txt](https://github.com/openpbs/openpbs/files/4732874/test_place.txt)

No new test needed. The existing Test_explicit_psets test suite was failing.

### DIFF
--- a/src/scheduler/node_partition.c
+++ b/src/scheduler/node_partition.c
@@ -899,7 +899,7 @@ find_alloc_np_cache(status *policy, np_cache ***pnpc_arr,
 	if (npc == NULL) {
 		int flags = NP_NO_ADD_NP_ARR;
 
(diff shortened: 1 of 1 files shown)


bhroam:
Hey Arun, the code looks fine, but could you update the PR title/comit title?  It isn't that the scheduler always ignores only_explicit_psets, it is only when we call find_alloc_np_cache().  This is mainly for -lplace=group, but also when we need to create hostsets in eval_placement().

How about changing it so something about it is not working with -lplace=group?

PR #1808, opened 2020-06-05
Remove dead code and add debug logs

SudeshnaMoulik:
#### Describe Bug or Feature
Remove Windows dead code and add more logs in accesinfo.c and passwd.c.

#### Describe Your Change
Dead code removed.
Added more debug logs in error path.
Added validation before closing a user handle in setuser_close_handle.

[mom_log.txt](https://github.com/openpbs/openpbs/files/4735004/mom_log.txt)
[high_debug.txt](https://github.com/openpbs/openpbs/files/4735005/high_debug.txt)

### DIFF
--- a/src/lib/Libwin/passwd.c
+++ b/src/lib/Libwin/passwd.c
@@ -5153,10 +5153,7 @@ has_read_access_domain_users(wchar_t dctrlw[PBS_MAXHOSTNAME+1])
 			log_eventf(PBSEVENT_ERROR, PBS_EVENTCLASS_SERVER, LOG_ERR, __func__, "failed in NetUserGetInfo, with errno %d", netst);
 			goto has_read_access_domain_users_end;
 		} else {
-			if ((netst == ERROR_BAD_NETPATH) || (netst == ERROR_INVALID_LEVEL) ||
-				(netst == NERR_InvalidComputer) || (netst == NERR_UserNotFound)) {
-				log_eventf(PBSEVENT_DEBUG3, PBS_EVENTCLASS_SERVER, LOG_DEBUG, __func__, "failed in NetUserGetInfo, with errno %d", netst);
-			}
+				log_eventf(PBSEVENT_DEBUG4, PBS_EVENTCLASS_SERVER, LOG_DEBUG, __func__, "failed in NetUserGetInfo, with errno %d", netst);
 		}
 		ncheck++;
 	}
--- a/src/lib/Libwin/passwd.c
+++ b/src/lib/Libwin/passwd.c
@@ -3869,18 +3869,18 @@ int
 wsystem(char *cmdline, HANDLE user_handle)
 {
 
-	STARTUPINFO             si = { 0 };
-	PROCESS_INFORMATION     pi = { 0 };
-	int			flags = CREATE_DEFAULT_ERROR_MODE| CREATE_NEW_PROCESS_GROUP;
-	int			rc = 0;
-	int			run_exit = 0;
-	char			cmd[PBS_CMDLINE_LENGTH] = {'\0'};
-	char			cmd_shell[MAX_PATH + 1] = {'\0'};
-	char			current_dir[MAX_PATH+1] = {'\0'};
-	char			*temp_dir = NULL;
-	int			changed_dir = 0;
-	DWORD			stat;
-	LPVOID			user_env = NULL;
+	STARTUPINFO  si = { 0 };
+	PROCESS_INFORMATION  pi = { 0 };
+	int  flags = CREATE_DEFAULT_ERROR_MODE| CREATE_NEW_PROCESS_GROUP;
+	int  rc = 0;
+	int  run_exit = 0;
+	char  cmd[PBS_CMDLINE_LENGTH] = {'\0'};
+	char  cmd_shell[MAX_PATH + 1] = {'\0'};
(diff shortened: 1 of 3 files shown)


src/lib/Libwin/passwd.c
```diff
@@ -2533,7 +2534,7 @@ default_local_homedir(char *username, HANDLE usertoken, int ret_profile_path)
 		} else {
 			userlogin = LogonUserNoPass(username);
 			if (userlogin == 0) {
```
subhasisb, line 2536: Do you not want to print the last error code? Since you are moving from log_errf() which could print the last error to log_evebtf() which cannot, do you not want to mention the lasteerror explicitly?

src/lib/Libwin/passwd.c
```diff
@@ -4050,7 +3790,7 @@ setuser(char *user)
 
 	setuser_hdle = LogonUserNoPass(user);
 	if (setuser_hdle == 0) {
-		log_errf(-1, __func__, "failed in LogonUserNoPass for %s", user);
+		log_eventf(PBSEVENT_ERROR, PBS_EVENTCLASS_SERVER, LOG_ERR, __func__, "failed in LogonUserNoPass for %s", user);
```
subhasisb, line 3793: Same question here: Why is log_errf() not good here?

src/lib/Libwin/passwd.c
```diff
@@ -4139,8 +3879,8 @@ wsystem(char *cmdline, HANDLE user_handle)
 	char			current_dir[MAX_PATH+1] = {'\0'};
 	char			*temp_dir = NULL;
 	int			changed_dir = 0;
-	DWORD		stat;
-	LPVOID  user_env = NULL;
+	DWORD			stat;
```
subhasisb, line 3882: can we please format all the local variables to be aligned? Using spaces instead of tabs is the best idea!!

src/lib/Libwin/passwd.c
```diff
@@ -4170,7 +3910,9 @@ wsystem(char *cmdline, HANDLE user_handle)
 	} else {
 		flags = flags|CREATE_UNICODE_ENVIRONMENT;
 		if (!CreateEnvironmentBlock(&user_env, user_handle, FALSE)) {
+			run_exit = GetLastError();
```
subhasisb, line 3913: why do we need this line? This is being done later anyways

subhasisb: besides log_errf() could automatically log it?

src/lib/Libwin/passwd.c
```diff
@@ -4986,7 +4727,7 @@ logon_pw(char *username,
 				CloseHandle(pwdp->pw_userlogin);
 				pwdp->pw_userlogin = LogonUserNoPass(username);
 				if (pwdp->pw_userlogin == 0) {
-					log_errf(-1, __func__, "failed in LogonUserNoPass for %s", username);
+					log_eventf(PBSEVENT_ERROR, PBS_EVENTCLASS_SERVER, LOG_ERR, __func__, "failed in LogonUserNoPass for %s", username);
```
subhasisb, line 4730: Why is log_errf() not good?

src/lib/Libwin/passwd.c
SudeshnaMoulik: I don't think here GetLastError() will show the real reason of any error, as LogonUserNoPass() is PBS API and the corresponding function have proper error handling.

src/lib/Libwin/passwd.c
SudeshnaMoulik: Same as above, LogonUserNoPass is PBS API, so it don't set last error.

src/lib/Libwin/passwd.c
SudeshnaMoulik: This part of code was added via [reference](https://github.com/openpbs/openpbs/pull/1770/commits/196a3dff8fe7dcbd4cdb597a8bd7af6ba8ccc0fa) which by mistake got deleted via [here](https://github.com/openpbs/openpbs/pull/1781/files#diff-60063bfd099fee0809396e4aebb54f6aL4151)

src/lib/Libwin/passwd.c
```diff
@@ -5411,6 +5152,11 @@ has_read_access_domain_users(wchar_t dctrlw[PBS_MAXHOSTNAME+1])
 			(netst == ERROR_LOGON_FAILURE)) {
 			log_eventf(PBSEVENT_ERROR, PBS_EVENTCLASS_SERVER, LOG_ERR, __func__, "failed in NetUserGetInfo, with errno %d", netst);
 			goto has_read_access_domain_users_end;
+		} else {
+			if ((netst == ERROR_BAD_NETPATH) || (netst == ERROR_INVALID_LEVEL) ||
```
subhasisb, line 5156: since u are logging at DEBUG3 you can actually log all error cases - not just the specific 4 u are checking for. BTW, only one code is left it seems ERROR_ACCESS_DENIED :)

SudeshnaMoulik: As we are printing this in a loop, so if we want to log for all the error codes, I think it will be better to increase the debug level to DEBUG4

subhasisb: DEBUG3 would have been good enough, but its ok at DEBUG4 also

PR #1809, opened 2020-06-07
Updated instructions for Centos-8

kjakkali:
#### Describe Bug or Feature
Package instructions for Centos-8.

#### Describe Your Change
Updated instructions for Centos-8

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -5,11 +5,13 @@ How to install PBS using the configure script.
 
 1. Install the prerequisite packages for building PBS.
 
-  For CentOS-8 systems you should run the following command as root:
+  For CentOS-8 systems you should configure and enable PowerTools 
+  repo for hwloc-devel and libedit-devel packages.
 
-    Configure and enable PowerTools repo for hwloc-devel and libedit-devel packages
-      dnf install -y dnf-plugins-core
-      dnf config-manager --set-enabled PowerTools
+  You should run the following command as root:
+
+    dnf install -y dnf-plugins-core
+    dnf config-manager --set-enabled PowerTools
 
     dnf install -y gcc make rpm-build libtool hwloc-devel \
       libX11-devel libXt-devel libedit-devel libical-devel \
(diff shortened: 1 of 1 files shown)


INSTALL
```diff
@@ -5,7 +5,17 @@ How to install PBS using the configure script.
 
 1. Install the prerequisite packages for building PBS.
 
-  For CentOS systems you should run the following command as root:
+  For CentOS-8 systems you should run the following command as root:
```
prakashcv13, line 8: the following line is not a command.

INSTALL
```diff
@@ -5,7 +5,17 @@ How to install PBS using the configure script.
...
+
+    Configure and enable PowerTools repo for hwloc-devel and libedit-devel packages
```
prakashcv13, line 10: why not include the actual commands.

INSTALL
```diff
@@ -5,7 +5,19 @@ How to install PBS using the configure script.
 
 1. Install the prerequisite packages for building PBS.
 
-  For CentOS systems you should run the following command as root:
+  For CentOS-8 systems you should run the following command as root:
```
Jitendra-Rajak, line 8: I think here we can remove command as below we have both instruction and commands.

INSTALL
```diff
@@ -5,7 +5,19 @@ How to install PBS using the configure script.
...
+
+    Configure and enable PowerTools repo for hwloc-devel and libedit-devel packages
```
Jitendra-Rajak, line 10: This instruction should start a column before so that the dnf below can be in-lined with the other dnf.

INSTALL
```diff
@@ -5,7 +5,19 @@ How to install PBS using the configure script.
...
+      dnf install -y dnf-plugins-core
+      dnf config-manager --set-enabled PowerTools
+
+    dnf install -y gcc make rpm-build libtool hwloc-devel \
```
Jitendra-Rajak, line 14: An instruction can be added above saying install required packages or tools.

kjakkali: Mentioned at line 11 .

PR #1810, opened 2020-06-08
test_prologue_exception_sisters fails because it takes longer for a j…

Klovely:
…ob to held when cgroups is enabled

Signed-off-by: Lovely Kumari <lovely.kumari@altair.com>

#### Describe Bug or Feature
The TestPbsExecutePrologue.test_prologue_exception_sisters  is failing on  machine is part of the  because it expects the job to be in the "H" state. To be in the Held state, the job has to try to run at least 21 times.
But after checking the state for the default 60 times, the job had not been rerun 21 times on machines where the croups hook was enabled.
Believe this is because the cgroups hook is registered for a lot of hook events, and that causes the job to go through the job path a little slower than when there are fewer hook events that it needs to go through.

#### Describe Your Change
Increased the test's max_attempts to be 100 and interval to 2 . This gives slower machines time to run the job enough times to have it get held. 

[logtest_prologue_exception_sisters_after_fix.txt](https://github.com/openpbs/openpbs/files/4743916/logtest_prologue_exception_sisters_after_fix.txt)
[logtest_prologue_exception_sisters_before_fix.txt](https://github.com/openpbs/openpbs/files/4743917/logtest_prologue_exception_sisters_before_fix.txt)

### DIFF
--- a/test/tests/functional/pbs_hook_execjob_prologue.py
+++ b/test/tests/functional/pbs_hook_execjob_prologue.py
@@ -301,5 +301,5 @@ class TestPbsExecutePrologue(TestFunctional):
         held_cmt = "job held, too many failed attempts to run"
         criteria = {'job_state': 'H', 'comment': held_cmt}
         for jid in job_list:
-            self.server.expect(JOB, criteria, attrop=PTL_AND, id=jid,
-                               max_attempts=100, interval=2)
+            self.server.expect(JOB, criteria, id=jid, max_attempts=100,
+                               interval=2)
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_hook_execjob_prologue.py
```diff
@@ -301,4 +301,5 @@ def test_prologue_exception_sisters(self):
         held_cmt = "job held, too many failed attempts to run"
         criteria = {'job_state': 'H', 'comment': held_cmt}
         for jid in job_list:
-            self.server.expect(JOB, criteria, attrop=PTL_AND, id=jid)
+            self.server.expect(JOB, criteria, attrop=PTL_AND, id=jid,
+                               max_attempts=100, interval=2)
```
bhroam, line 305: The attrop=PTL_AND is no longer needed

Klovely:
Latest log:-
=====
[latest_log_prologue_exception_sisters_after_fix.txt](https://github.com/openpbs/openpbs/files/4750108/latest_log_prologue_exception_sisters_after_fix.txt)

PR #1811, opened 2020-06-08
Fix for slow qdel due to heavy pbs_loadconf

hirenvadalia:
#### Describe Bug or Feature

* As part of #1505, I added pbs_loadconf(0) in pbs_iff to get PBS_ENCRYPT_METHOD value from pbs.conf so it can pass that to the server along with PBS_BATCH_Authenticate request. But pbs_loadconf() is heavy function as it does IO + memory allocations + string parsing (aka key=value parsing) + much more stuff, so pbs_iff became slow, which slowdowns qdel sending requests to the server, which results in overall qdel slow.
And this issue is not only limited to qdel but all client commands who do connect/disconnect to the server per jobid...

#### Describe Your Change

* Remove heavy pbs_loadconf() and pass encrypt method via environment variable, like we pass PBS_IFF_CLIENT_ADDR (aka client address)

#### Link to Design Doc

* None

* valgrind logs - no new leak detected: [valgrind.txt](https://github.com/openpbs/openpbs/files/4763453/valgrind.txt)

### DIFF
--- a/src/iff/iff2.c
+++ b/src/iff/iff2.c
@@ -98,8 +98,6 @@ main(int argc, char *argv[], char *envp[])
 	/*the real deal or output pbs_version and exit?*/
 	PRINT_VERSION_AND_EXIT(argc, argv);
 
-	pbs_loadconf(0);
-
 	cln_hostaddr = getenv(PBS_IFF_CLIENT_ADDR);
 
 	/* Need to unset LOCALDOMAIN if set, want local host name */
@@ -239,7 +237,7 @@ main(int argc, char *argv[], char *envp[])
 		parentport = ntohs(parentsock_port);
 
 	pbs_errno = 0;
-	err = tcp_send_auth_req(sock, parentport, pwent->pw_name);
+	err = tcp_send_auth_req(sock, parentport, pwent->pw_name, AUTH_RESVPORT_NAME, getenv(PBS_CONF_ENCRYPT_METHOD));
 	if (err != 0 && pbs_errno != PBSE_BADCRED)
 		return 2;
 
--- a/src/lib/Libifl/auth.c
+++ b/src/lib/Libifl/auth.c
@@ -369,31 +369,41 @@ is_valid_encrypt_method(char *method)
  *	tcp_send_auth_req - encodes and sends PBS_BATCH_Authenticate request
  *
  * @param[in] sock - socket descriptor
+ * @param[in] port - parent port in pbs_iff (only used in resvport auth) else 0
+ * @param[in] user - authenticating user name
+ * @param[in] auth_method - auth method name
+ * @param[in] encrypt_method - encrypt method name
  *
  * @return	int
  * @retval	0 on success
  * @retval	-1 on error
  */
 int
-tcp_send_auth_req(int sock, unsigned int port, char *user)
+tcp_send_auth_req(int sock, unsigned int port, char *user, char *auth_method, char *encrypt_method)
 {
 	struct batch_reply *reply;
 	int rc;
-	int am_len = strlen(pbs_conf.auth_method);
-	int em_len = strlen(pbs_conf.encrypt_method);
+	int am_len;
+	int em_len = encrypt_method ? strlen(encrypt_method) : 0;
 
+	if (auth_method == NULL || *auth_method == '\0') {
+		/* auth method can't be null or empty string */
+		pbs_errno = PBSE_INTERNAL;
+		return -1;
+	}
+	am_len = strlen(auth_method);
 	set_conn_errno(sock, 0);
 	set_conn_errtxt(sock, NULL);
 
 	if (encode_DIS_ReqHdr(sock, PBS_BATCH_Authenticate, user) ||
 		diswui(sock, am_len) || /* auth method length */
-		diswcs(sock, pbs_conf.auth_method, am_len) || /* auth method */
+		diswcs(sock, auth_method, am_len) || /* auth method */
 		diswui(sock, em_len)) { /* encrypt method length */
 		pbs_errno = PBSE_SYSTEM;
 		return -1;
 	}
 
-	if (em_len > 0 && diswcs(sock, pbs_conf.encrypt_method, em_len)) { /* encrypt method */
+	if (em_len > 0 && diswcs(sock, encrypt_method, em_len)) { /* encrypt method */
 		pbs_errno = PBSE_SYSTEM;
 		return -1;
 	}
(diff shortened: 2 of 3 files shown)


src/iff/iff2.c
```diff
@@ -239,7 +237,7 @@ main(int argc, char *argv[], char *envp[])
 		parentport = ntohs(parentsock_port);
 
 	pbs_errno = 0;
-	err = tcp_send_auth_req(sock, parentport, pwent->pw_name);
+	err = tcp_send_auth_req(sock, parentport, pwent->pw_name, AUTH_RESVPORT_NAME, getenv(PBS_CONF_ENCRYPT_METHOD));
```
nithinj, line 240: something like RESVPORT_AUTH will be more readable compared to AUTH_RESVPORT_NAME

src/lib/Libifl/auth.c
```diff
@@ -369,31 +369,41 @@ is_valid_encrypt_method(char *method)
...
 int
-tcp_send_auth_req(int sock, unsigned int port, char *user)
+tcp_send_auth_req(int sock, unsigned int port, char *user, char *auth_method, char *encrypt_method)
 {
 	struct batch_reply *reply;
 	int rc;
-	int am_len = strlen(pbs_conf.auth_method);
-	int em_len = strlen(pbs_conf.encrypt_method);
+	int am_len;
+	int em_len = encrypt_method ? strlen(encrypt_method) : 0;
```
suresh-thelkar, line 387: Can we do this check after making sure that auth_method != NULL i.e. after the following block of code.

src/lib/Libifl/auth.c
```diff
@@ -478,6 +488,8 @@ _invoke_pbs_iff(int psock, char *server_name, int server_port, char *ebuf, size_
 	/* for compatibility with 12.0 pbs_iff */
 	(void)snprintf(cmd[1], sizeof(cmd[1]) - 1, "%s -i %s %s %u %d %u", pbs_conf.iff_path, pbs_client_addr, server_name, server_port, psock, psock_port);
 #ifdef WIN32
+	if (pbs_conf.encrypt_method[0] != '\0')
+		SetEnvironmentVariable(PBS_CONF_ENCRYPT_METHOD, pbs_conf.encrypt_method);
```
suresh-thelkar, line 492: Can we check the return value of SetEnvironmentVariable and use GetLastError or the corresponding API in case of an error.

src/iff/iff2.c
subhasisb: I feel verything in the auth API starts with AUTH, so it actually makes sense to have it start with AUTH_. This is in sync with what the others are being called for example. AUTH_GSS_NAME

src/lib/Libifl/auth.c
subhasisb: does it matter?

suresh-thelkar: I was thinking that if auth_method is NULL then we don't need to calculate em_len where we are using strlen to calculate the length.

subhasisb: yeah i got that, but that error path will seldom happen and we do not need to really bother much about a seldom happening error path's performance, right?

suresh-thelkar: Yes. If it doesn't happen frequently there is no much performance impact. I agree with you.

src/lib/Libifl/auth.c
```diff
@@ -369,31 +369,41 @@ is_valid_encrypt_method(char *method)
...
 	struct batch_reply *reply;
 	int rc;
-	int am_len = strlen(pbs_conf.auth_method);
-	int em_len = strlen(pbs_conf.encrypt_method);
+	int am_len;
+	int em_len = encrypt_method ? strlen(encrypt_method) : 0;
 
+	if (auth_method == NULL || *auth_method == '\0') {
+		/* auth method can't be null or empty string */
+		pbs_errno = PBSE_INTERNAL;
```
suresh-thelkar, line 391: Just a question. Earlier this function used to return only PBSE_SYSTEM in case of any error. Now we are returning PBSE_INTERNAL if PBS_CONF_ENCRYPT_METHOD is not provided as an environment variable. It looks like the caller of this function does not check pbs_errno in case of a failure and say the exact reason why it has failed and always give the error "Failed to send auth request". Do we want to take care of this at the caller ?

hirenvadalia: No, it will be taken care in pbs_connect() (aka caller of caller to this func), pbs_connect() is already printing message: `"auth: error returned: %d\n", pbs_errno`

suresh-thelkar: Thanks @hirenvadalia. Good that we are checking pbs_errno in pbs_connect(). Along with the error you mentioned above we also get(if errbuff not null) "auth:Failed to send auth request" which I think it is very much general, may be if we could see the second error as "auth:<env_ name> is not provided." it would be easier for the caller of caller to know that authentication is failing because the corresponding environement variable is not provided.
What do you think ?

hirenvadalia: I don't think we need that, as pbs_connect() is just printing error code and message return by engage_client_auth(), pbs_connect() doesn't care what auth methods it is using underneath.
Also this setting env is only comes into picture when auth is resvport, so its not generic.

suresh-thelkar: I am leaving it up to you @hirenvadalia. My only concern was that if we do the above,  one can easily know that authentication is failing due to the lack of the corresponding environment variable (even if resvport auth is less freqently used) instead of providing generic error.

subhasisb: @suresh-thelkar he is also printing the setenvironmentvar() api failure into stderr, so that would be seen as well.

suresh-thelkar: I only saw windows counter part for the above i.e. SetEnvironmentVariable() where we are doing this and didn't find a LINUX counterpart. @subhasisb/ @hirenvadalia, Am I missing anything here ?

hirenvadalia: @suresh-thelkar for *nix we don't need to set any env explicitly in current process, as `popen()` can understand `env=var <pbs_iff cmd>` syntax and that's what we use in *nix, please see `#else` part of `#ifdef WIN32`

suresh-thelkar: Got it @hirenvadalia. Thanks.

src/iff/iff2.c
nithinj: It's the postfix `_NAME` that causes confusion. That makes it sound more like a variable than a macro. It would have been better if you keep it as AUTH_RESVPORT and AUTH_GSS.

hirenvadalia: @nithinj I understand your confusion, but for me AUTH_RESVPORT looks like int value in some enum/macro (like earlier we had), where here I wanted to explicitly say its string and its name of auth method, that's why I added `_NAME` in it. And if you see `AUTH_RESVPORT_NAME` as variable then its kinda `const char *` variable right, just that I used `#define` instead making it const char *. HTH!

PR #1813, opened 2020-06-08
Fixing license text related inconsistencies

agrawalravi90:
#### Describe Bug or Feature

The license text for some files used the old name "pbs pro", this just fixes that.

No code change as such, smoke tests should be good enough to test this.


src/lib/Libifl/pbsD_runjob.c
```diff
@@ -49,55 +49,6 @@
 #include "dis.h"
 #include "pbs_ecl.h"
 
-/*
```
arungrover, line 52: How did this happen :)

PR #1814, opened 2020-06-09
PBS fails local peering on server with multiple IPs

arungrover:
#### Describe Bug or Feature
When a local peering complex is configured on a machine that has more than one network interface, PBS server fails to identify that the move-job request issued by the scheduler is for the local server. This makes it try network move which fails because the job already exists in the destination server (which is actually the local server).
This issue only happens when the destination queue in sched_config is configured with a server name that's attached to one of the network interface which does not match PBS_SERVER in pbs.conf.

#### Describe Your Change
Change is to compare all the IP addresses of the destination server to the local pbs server, if any one of them match then the request needs to be honored locally.

#### Link to Design Doc
NA

This issue happened on the Shasta setup. I was able to reproduce it manually in a container but I couldn't get ptl to work with this setup in a way that the test can run on Shasta as well as on the normal machine.
Since the issue is remote, I have attached manual test logs - 
[test_peer.txt](https://github.com/openpbs/openpbs/files/4749305/test_peer.txt)

### DIFF
--- a/src/lib/Libnet/get_hostaddr.c
+++ b/src/lib/Libnet/get_hostaddr.c
@@ -177,3 +177,55 @@ compare_short_hostname(char *shost, char *lhost)
 	else
 		return 1;	/* no match */
 }
+
+/**
+ * @brief
+ *
+ * comp_svraddr - get internal internet address of the given host
+ *		  check to see if any of the addresses match the given server
+ *		  net address.
+ *
+ *
+ * @param[in] svr_addr - net address of the server
+ * @param[in] hostname - hostname whose internet addr needs to be compared
+ *
+ * @return	int
+ * @retval	0 address found
+ * @retval	1 address not found
+ * @retval	2 failed to find address
+ *
+ */
+int
+comp_svraddr(pbs_net_t svr_addr, char *hostname)
+{
+	struct addrinfo *aip, *pai;
+	struct addrinfo hints;
+	struct sockaddr_in *inp;
+	int		err;
+	pbs_net_t	res;
+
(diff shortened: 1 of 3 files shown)


src/lib/Libnet/get_hostaddr.c
```diff
@@ -177,3 +177,55 @@ compare_short_hostname(char *shost, char *lhost)
...
+
+	if ((hostname == 0) || (*hostname == '\0')) {
+		return (2);
+	}
+
+	memset(&hints, 0, sizeof(struct addrinfo));
+	hints.ai_family = AF_UNSPEC;
+	hints.ai_socktype = SOCK_STREAM;
+	hints.ai_protocol = IPPROTO_TCP;
+	if ((err = getaddrinfo(hostname, NULL, &hints, &pai)) != 0) {
```
vstumpf, line 215: We don't use err, so I would remove it.

src/lib/Libnet/get_hostaddr.c
```diff
@@ -177,3 +177,55 @@ compare_short_hostname(char *shost, char *lhost)
...
+int
+comp_svraddr(pbs_net_t svr_addr, char *hostname)
+{
+	struct addrinfo *aip, *pai;
+	struct addrinfo hints;
+	struct sockaddr_in *inp;
+	int		err;
+	pbs_net_t	res;
+
+	if ((hostname == 0) || (*hostname == '\0')) {
```
vstumpf, line 207: check hostname == NULL, not 0

arungrover:
[test_peer2.txt](https://github.com/openpbs/openpbs/files/4755378/test_peer2.txt)

Issue #1817, opened 2020-06-09
cgroups hook problem with hyperthreading


drtoss:
If the pbs_cgroups hook is enabled on a python3 PBS and vnode_per_numa_node, use_hyperthreads, and ncpus_are_cores are all set to true, and the cpuset cgroup is enabled, then you get an error when a float is used to index a list.

The problem is the python3 decision that division of integers should produce a float.

A fix, partially tested, is to use floor division:

```
--- /home7/dtalcott/work/drtaltc/src/hooks/cgroups/pbs_cgroups.PY       2020-06-08 12:02:08.930018109 -0700
+++ nas_cgroups.py      2020-06-09 08:26:44.430456554 -0700
@@ -1853,7 +1850,7 @@
                 if 'ht' in cpuinfo['cpu'][0]['flags']:
                     cpuinfo['hyperthreads_per_core'] = \
                         (cpuinfo['cpu'][0]['siblings']
-                         / cpuinfo['cpu'][0]['cpu cores'])
+                         // cpuinfo['cpu'][0]['cpu cores'])
                     # Map hyperthreads to physical cores
                     if cpuinfo['hyperthreads_per_core'] > 1:
                         pbs.logmsg(pbs.EVENT_DEBUG4,
@@ -1888,7 +1885,7 @@
         except Exception:
             pbs.logmsg(pbs.EVENT_DEBUG, '%s: Hyperthreading check failed' %
                        caller_name())
-        cpuinfo['physical_cpus'] = cpuinfo['logical_cpus'] / \
+        cpuinfo['physical_cpus'] = cpuinfo['logical_cpus'] // \
             cpuinfo['hyperthreads_per_core']
         pbs.logmsg(pbs.EVENT_DEBUG4, '%s returning: %s' %
                    (caller_name(), cpuinfo))
@@ -2247,11 +2244,11 @@
                     if not self.cfg['use_hyperthreads']:
                         # Do not treat a hyperthread as a core when
                         # use_hyperthreads is false.
-                        threads /= self.cpuinfo['hyperthreads_per_core']
+                        threads //= self.cpuinfo['hyperthreads_per_core']
                     elif self.cfg['ncpus_are_cores']:
                         # use_hyperthreads and ncpus_are_cores are both true,
                         # advertise only one thread per core
-                        threads /= self.cpuinfo['hyperthreads_per_core']
+                        threads //= self.cpuinfo['hyperthreads_per_core']
                     if vnodes:
[...]

Issue #1818, opened 2020-06-09
cgroups hook issue getting systemd version


drtoss:
The cgroups hook routine _get_systemd_version invokes systemctl via subprocess.Popen() to learn the version of systemd by parsing the value out of stdout. Unfortunately, python3 changed Popen() to return the outputs as bytes arrays, rather than strings. You need to convert that to a string before using it in a regular expression test.

Suggested fix. Should work with python2, but untested.

```
--- /home7/dtalcott/work/drtaltc/src/hooks/cgroups/pbs_cgroups.PY       2020-06-08 12:02:08.930018109 -0700
+++ nas_cgroups.py      2020-06-09 08:52:47.835824332 -0700
@@ -3283,6 +3312,8 @@
                                        stdout=subprocess.PIPE,
                                        stderr=subprocess.PIPE)
             out, err = process.communicate()
+            if not isinstance(out, str):
+                out = out.decode('utf-8')
             ver = int(re.sub(r'Version=([0-9]+)', r'\1', out.splitlines()[0]))
         except Exception:
             # Note we also get here if we can't decode an int version

```

alexis-cousein:
Could you post an example of an error generated by not converting to a str in Python3 in the MoM log (or evidence that the systemd version was not accurately detected)?

It would also be more readable to actually call "decode" only if the type is seen as bytes or bytearray (which are the types for which the method exists), rather than "not a string (so it must be a bytes or bytearray)".

alexis-cousein:
An alternative would be to set universal_newlines to True in the Popen constructor. That seems even more elegant, given that we definitely don't mind universal newlines, and that should open everything in text mode and give us strings (not that having the guard behind it asa firewall is not still a good idea, mind you).

drtoss:
> An alternative would be to set universal_newlines to True in the Popen constructor. That seems even more elegant, given that we definitely don't mind universal newlines, and that should open everything in text mode and give us strings in both Python 2.5-2.7 and Python 3 (having the guard to check the type behind it as a firewall is still a good idea, mind you).

That's what I have done in my own hooks. I wasn't sure which versions of python 2 supported universal newlines.

alexis-cousein:
Indeed, that was broken:


    06/09/2020 13:48:37.102182;0800;pbs_python;Hook;pbs_python;_get_systemd_version: Method called 06/09/2020 13:48:37.161733;0800;pbs_python;Hook;pbs_python;__init__: systemd version seems to be 0

while on the host:

[...]

PR #1822, opened 2020-06-10
Fix issues #1817 and #1818 seen by drtoss

alexis-cousein:
#### Describe Bug or Feature

This is mainly cleanup for two issues seen by drtoss affecting only Python3-based versions of PBSPro/openPBS

1. 
Python 3 "x / y" produces floats even when x and y are integers and y is a divisor of x. This propagates into hyperthreads_per_core and ultimately the variable "needed" that describes how many CPU threads a job needs to assign. When use_hyperthreads and ncpus_are_cores are enabled, the code correctly assigns all cpus and needed becomes 0.0. Unfortunately, instead of adding a guard to see we're done, we're also going to pick the 0.0 CPUs needed from a list of remaining CPUs, using slice syntax, so we get a crash.

2. 
Popen.communicate() _by default_ returns a pair (stdout, stderr) of non-mutable array of bytes ("str" in Python 2, "bytes"  in Python3). 

We cannot use string operations on those, but we have 6 Popen constructors and often we do things like use splitlines(). Since subprocesses are always dangerous, we do have code to catch exceptions, but unfortunately that leads to silent failures. The most prominent one is that on Python 3 PBSPro we will always think a host does not have systemd enabled, but there are others. 

3.
A small corner case for memsw -- if specifying -lselect=1:ncpus=1:mem=X, if memsw is enabled and X is less than the memsw default, vmem will be set to default. If this is larger than X, the job will fail on hosts that have no swap (the hook will detect that you're setting nonsensical limits that the host will not honour).

#### Describe Your Change

For problem #1, use int(x // y) when you want an integer result. A guard was also inserted to avoid needless work when the amount of needed CPUs goes below 1 (i.e. zero). Also, there is a PTL test that _should_ have caught this, but didn't because the config file neutered the effects of ncpus_are_cores (if use_hyperthreads is False then ncpus_are_cores doesn't do anything, so the current test also doesn't test anything). That test's config file was changed to fully test ncpus_are_cores CPU assignment.

For problem #2, set universal_newlines in the Popen constructor so that the stdout and stderr objects returned are always strings (in both Python 2.x, x>=3, and Python 3). Add a stringified_output() function to nevertheless _convert_ anything we may find to strings (should Python change again, should we get "unicode" objects in a version of Python2, or if we ever have NoneType objects), so that we can safely call string operations. The function should basically do nothing but return the first argument (unchanged, without creating a new object) if we did our job in the calling routine, so the overhead is really small.

Error reporting was also reworked to remove some of the silent failures (especially at DEBUG4). All too often just running a command is seen as "success", so in the relevant cases checks on process.returncode were added to at least _log_ something.

For problem #3, on hosts without swap, leave vmem to mem regardless of the default if vmem_default>mem. Don't apply the default just to make the job fail.

#### Link to Design Doc

N/A

Problem #1 _should_ have been seen by the test_cgroup_cpuset_ncpus_are_cores PTL test that is already existing, but unfortunately the config file loaded disables hyperthread use, which makes the test fail to fully test ncpus_are_cores functionality.

Problem #2 has been seen on system MoM logs in almost _all_ existing tests: hosts that have a systemd running with a valid systemd version are reported as having systemd version 0 in the MoM logs. Sadly, it is well nigh impossible to design a PTL test for it; a generic PTL test has to guard on logging "failures" on hosts that do not have systemd running, but given these are also in Python would be using almost the exact same code (with the use of Popen) to test whether the system _can_ run the test. It would be possible to add a test if a guard can be found that is _not_ dependent on Python (but e.g. on the name of a host) if you can guarantee that host runs systemd.

An example on a particular host:
[cousein@pbi-33-r7 ~]$ systemctl --version 
systemd 208
+PAM +LIBWRAP +AUDIT +SELINUX +IMA +SYSVINIT +LIBCRYPTSETUP +GCRYPT +ACL +XZ

but the MoM logs for existing PTL tests reveal:
    06/09/2020 13:48:37.102182;0800;pbs_python;Hook;pbs_python;_get_systemd_version: Method called 06/09/2020 13:48:37.161733;0800;pbs_python;Hook;pbs_python;__init__: systemd version seems to be 0

That should have been 208, not 0.

As detailed above, this is hard to test _within_ PTL, since you would need the exact same code that is in the hook to determine whether the host has systemd enabled (here, it was checked _manually_ by logging into the host). Write a bug in the hook and you'll likely write the same one in the PTL test to determine whether the host should be skipped or not.

Problem #3 can be seen in the following MoM log snippet. memsw was enabled, the default was 256MB, and a job was submitted with -lncpus=1:mem=123gb on a host with no configured swap space:

06/10/2020 14:29:08;0800;pbs_python;Hook;pbs_python;get_vmem_on_node: No swap space detected
06/10/2020 14:29:08;0080;pbs_python;Hook;pbs_python;['Traceback (most recent call last):', '  File "<embedded code object>", line 5538, in main', '  File "<embedded code object>", line 968, in invoke_handler', '  File "<embedded code object>", line 1017, in _execjob_begin_handler', '  File "<embedded code object>", line 4504, in configure_job', 'CgroupLimitError: Job might utilize swap and no swap space available']
06/10/2020 14:29:08;0001;pbs_python;Hook;pbs_python;Unexpected error in pbs_cgroups handling execjob_begin event for job 323.tc72 (system hold set): CgroupLimitError ('Job might utilize swap and no swap space available',)
06/10/2020 14:29:08;0008;pbs_mom;Job;323.tc72;Unexpected error in pbs_cgroups handling execjob_begin event for job 323.tc72 (system hold set): CgroupLimitError ('Job might utilize swap and no swap space available',)

### DIFF
--- a/src/hooks/cgroups/pbs_cgroups.PY
+++ b/src/hooks/cgroups/pbs_cgroups.PY
@@ -3339,7 +3339,7 @@ class CgroupUtils(object):
             # before calling splitlines
             # should not happen since we pass universal_newlines True
             out_split = stringified_output(out).splitlines()
-            ver = int(re.sub(r'Version=([0-9]+)', r'\1', out_split[0]))
+            ver = int(re.sub(r'Version=\D*(\d+)', r'\1', out_split[0]))
         except Exception:
             # Note we also get here if we can't decode an int version
             return 0
@@ -3835,12 +3835,12 @@ class CgroupUtils(object):
                 # and ncpus_are_cores are both True)
                 pass
             elif needed <= len(cores):
-                    corelist = sorted(cores)
-                    assigned['cpuset.cpus'] += corelist[:needed]
+                corelist = sorted(cores)
+                assigned['cpuset.cpus'] += corelist[:needed]
             else:
-                    pbs.logmsg(pbs.EVENT_DEBUG4, '%s: %d ncpus still needed' %
-                               (caller_name(), len(cores) - needed))
-                    return {}
+                pbs.logmsg(pbs.EVENT_DEBUG4, '%s: %d ncpus still needed' %
+                           (caller_name(), len(cores) - needed))
+                return {}
 
             # Set cpuset.mems to the socketlist for now even though
             # there may not be sufficient memory. Memory gets
--- a/src/hooks/cgroups/pbs_cgroups.PY
+++ b/src/hooks/cgroups/pbs_cgroups.PY
@@ -138,10 +138,7 @@ else:
     import fcntl
     import pwd
 
-    if sys.version_info[0] < 3:
-        PYTHON2 = True
-    else:
-        PYTHON2 = False
+    PYTHON2 = sys.version_info[0] < 3
 
     try:
         bytearray
@@ -621,29 +618,28 @@ def printjob_info(jobid, include_attributes=False):
         cmd.append('-a')
     cmd.append(jobfile)
     try:
-        err_nonsplit = ''
         pbs.logmsg(pbs.EVENT_DEBUG4, 'Running: %s' % cmd)
         process = subprocess.Popen(cmd, shell=False,
                                    stdout=subprocess.PIPE,
                                    stderr=subprocess.PIPE,
                                    universal_newlines=True)
-        out_nonsplit, err_nonsplit = process.communicate()
+        out, err = process.communicate()
         if process.returncode != 0:
             pbs.logmsg(pbs.EVENT_DEBUG4,
                        'command return code non-zero: %s'
                        % str(process.returncode))
             pbs.logmsg(pbs.EVENT_DEBUG4,
                        'command stderr: %s'
-                       % stringified_output(err_nonsplit))
+                       % stringified_output(err))
     except Exception as exc:
         pbs.logmsg(pbs.EVENT_DEBUG2, 'Error running command: %s' % cmd)
         pbs.logmsg(pbs.EVENT_DEBUG2, 'Exception: %s' % exc)
         return {}
     # if we get a non-str type then convert before calling splitlines
     # should not happen since we pass universal_newlines True
-    out = stringified_output(out_nonsplit).splitlines()
+    out_split = stringified_output(out).splitlines()
     pattern = re.compile(r'^(\w.*):\s*(\S+)')
-    for line in out:
+    for line in out_split:
         result = re.match(pattern, line)
         if not result:
             continue
@@ -1781,7 +1777,7 @@ class NodeUtils(object):
                                        stdout=subprocess.PIPE,
                                        stderr=subprocess.PIPE,
                                        universal_newlines=True)
-            out_nonsplit = process.communicate()[0]
+            out = process.communicate()[0]
         except Exception:
             pbs.logmsg(pbs.EVENT_DEBUG4, 'Failed to execute: %s' %
                        " ".join(cmd))
@@ -1797,7 +1793,7 @@ class NodeUtils(object):
         try:
             # Try parsing the output
             import xml.etree.ElementTree as xmlet
-            root = xmlet.fromstring(stringified_output(out_nonsplit))
+            root = xmlet.fromstring(stringified_output(out))
             pbs.logmsg(pbs.EVENT_DEBUG4, 'root.tag: %s' % root.tag)
             for child in root:
                 if child.tag == 'gpu':
@@ -3067,7 +3063,7 @@ class CgroupUtils(object):
                                            stdout=subprocess.PIPE,
                                            stderr=subprocess.PIPE,
                                            universal_newlines=True)
-                out_nonsplit = process.communicate()[0]
+                out = process.communicate()[0]
                 if (process.returncode == 0):
                     # service is already active -- no need to create it
                     return
@@ -3106,13 +3102,12 @@ class CgroupUtils(object):
                        % (caller_name(), servicefile))
             raise
         try:
-            err_nonsplit = ''
             cmd = ['systemctl', 'start', os.path.basename(servicefile)]
             process = subprocess.Popen(cmd, shell=False,
                                        stdout=subprocess.PIPE,
                                        stderr=subprocess.PIPE,
                                        universal_newlines=True)
-            out_nonsplit, err_nonsplit = process.communicate()
+            out, err = process.communicate()
             if (process.returncode == 0):
                 pbs.logmsg(pbs.EVENT_DEBUG,
                            '%s: Started systemd service with file %s'
@@ -3123,7 +3118,7 @@ class CgroupUtils(object):
                            % (caller_name(), process.returncode))
                 pbs.logmsg(pbs.EVENT_ERROR, '%s: stderr of systemctl was: %s'
                            % (caller_name(),
-                              stringified_output(err_nonsplit)))
+                              stringified_output(err)))
         except Exception:
             pbs.logmsg(pbs.EVENT_DEBUG,
                        '%s: Failed to start systemd service: %s'
@@ -3339,12 +3334,12 @@ class CgroupUtils(object):
                                        stdout=subprocess.PIPE,
                                        stderr=subprocess.PIPE,
                                        universal_newlines=True)
-            out_nonsplit = process.communicate()[0]
+            out = process.communicate()[0]
             # if we get a non-str type then convert
             # before calling splitlines
             # should not happen since we pass universal_newlines True
-            out = stringified_output(out_nonsplit).splitlines()
-            ver = int(re.sub(r'Version=([0-9]+)', r'\1', out[0]))
+            out_split = stringified_output(out).splitlines()
+            ver = int(re.sub(r'Version=([0-9]+)', r'\1', out_split[0]))
         except Exception:
             # Note we also get here if we can't decode an int version
             return 0
@@ -3835,14 +3830,18 @@ class CgroupUtils(object):
             # assigned all of the fully avaiable cores. There still may
             # be cores to assign. When use_hyperthreads is disabled, we
             # assign all the cores here.
-            if needed >= 1:
-                if needed > len(cores):
+            if needed <= 0:
+                # we're done (will always hold if use_hyperthreads
+                # and ncpus_are_cores are both True)
+                pass
+            elif needed <= len(cores):
+                    corelist = sorted(cores)
+                    assigned['cpuset.cpus'] += corelist[:needed]
+            else:
                     pbs.logmsg(pbs.EVENT_DEBUG4, '%s: %d ncpus still needed' %
                                (caller_name(), len(cores) - needed))
                     return {}
-                else:
-                    corelist = sorted(cores)
-                    assigned['cpuset.cpus'] += corelist[:needed]
+
             # Set cpuset.mems to the socketlist for now even though
             # there may not be sufficient memory. Memory gets
             # checked later in this method.
@@ -4657,7 +4656,7 @@ class CgroupUtils(object):
                 except Exception:
                     jobdict = dict()
                     pbs.logmsg(pbs.EVENT_DEBUG2,
-                               '%s: Failed gathering local job data' %
+                               '%s: Failed to gather local job data' %
                                caller_name())
                 # There may not be a .JB file present for this job yet
                 if jobid not in jobdict:
@@ -5251,15 +5250,15 @@ class CgroupUtils(object):
             proc = subprocess.Popen(['dmesg'], shell=False,
                                     stdout=subprocess.PIPE,
                                     universal_newlines=True)
-            out_nonsplit = proc.communicate()[0]
+            out = proc.communicate()[0]
             # if we get a non-str type then convert before calling splitlines
             # should not happen since we pass universal_newlines True
-            out = stringified_output(out_nonsplit).splitlines()
+            out_split = stringified_output(out).splitlines()
         except Exception:
             return ''
-        out.reverse()
+        out_split.reverse()
         # Check to see if the job id is found in dmesg output
-        for line in out:
+        for line in out_split:
             start = line.find('Killed process ')
             if start < 0:
                 start = line.find('Task in /%s' % self.cfg['cgroup_prefix'])
--- a/src/hooks/cgroups/pbs_cgroups.PY
+++ b/src/hooks/cgroups/pbs_cgroups.PY
@@ -143,7 +143,7 @@ else:
     try:
         bytearray
         BYTEARRAY_EXISTS = True
-    except Exception:
+    except NameError:
         BYTEARRAY_EXISTS = False
 
 # Define some globals that get set in main
--- a/src/hooks/cgroups/pbs_cgroups.PY
+++ b/src/hooks/cgroups/pbs_cgroups.PY
@@ -3800,8 +3800,9 @@ class CgroupUtils(object):
             if self.cfg['use_hyperthreads'] and self.cfg['ncpus_are_cores']:
                 needed *= node.cpuinfo['hyperthreads_per_core']
             if needed > avail:
-                pbs.logmsg(pbs.EVENT_DEBUG4, '%s: Insufficient ncpus: %s/%s' %
-                           (caller_name(), needed, avail))
+                pbs.logmsg(pbs.EVENT_DEBUG4,
+                           '%s: insufficient ncpus: %s needed, %s available'
+                           % (caller_name(), needed, avail))
                 return {}
             if self.cfg['use_hyperthreads']:
                 # Find cores that are fully available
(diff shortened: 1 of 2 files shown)


minghui-liu:
Hi @alexis-cousein checkpep8 failed on Travis because of two indentation issues. 

`Running check: 'checkpep8' ... FAILED
495test/tests/functional/pbs_cgroups_hook.py:1384:20: E121 continuation line indentation is not a multiple of four
496                   # os.path.join(basedir, 'pbs.slice',
497                   ^
498test/tests/functional/pbs_cgroups_hook.py:1392:19: E121 continuation line indentation is not a multiple of four
499                  ]`

alexis-cousein:
> Hi @alexis-cousein checkpep8 failed on Travis because of two indentation issues.
> 
> `Running check: 'checkpep8' ... FAILED 495test/tests/functional/pbs_cgroups_hook.py:1384:20: E121 continuation line indentation is not a multiple of four 496 # os.path.join(basedir, 'pbs.slice', 497 ^ 498test/tests/functional/pbs_cgroups_hook.py:1392:19: E121 continuation line indentation is not a multiple of four 499 ]`

pycodestyle disagrees; it's a comment that is indented at the natural indentation corresponding to the square bracket that is above it. It's a checkpep8 bug.

Nevertheless, I shall resubmit with an indentation that _is_ a multiple of 4.

alexis-cousein:
There is no pleasing PEP8checker...if you actually _do_ indent with a mutliple of four, it complains about it being visually overindented.

Darmned if you do, damned if you don't...

Ah well...I'll make yet another pass at it --less readable and not more PEP8 compliant, but perhaps the tool will take this as the proverbial sacrificial goat.

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -1350,21 +1368,46 @@ def get_cgroup_job_dir(self, subsys, jobid, host):
...
+        cpath = None
+        if os.path.isdir(cdir):
+            cpath = os.path.join(cdir, 'pbs_jobs.service/jobid')
+            if not os.path.isdir(cpath):
+                cpath = os.path.join(cdir, 'pbs.service/jobid')
+            if not os.path.isdir(cpath):
+                cpath = os.path.join(cdir, 'pbs.slice')
+            if not os.path.isdir(cpath):
+                cpath = os.path.join(cdir, 'pbs')
+        return cpath
```
minghui-liu, line 1409: I suggest using iteration to make this function easier to understand
```
def find_main_cpath(self, cdir):
    if os.path.isdir(cdir):
        paths = ['pbs_jobs.service/jobid', 'pbs.service/jobid', 'pbs.slice', 'pbs']
        for p in paths:
            cpath = os.path.join(cdir, p)
            if os.path.isdir(cpath):
                return cpath
    return None
```

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -2627,27 +2671,34 @@ def test_cgroup_execjob_end_should_delete_cgroup(self):
...
+            #    filename = os.path.join(path, 'pbs.slice', 'pbs-%s.slice'
+            #                            % systemd_escape(jid))
+            #    self.logger.info('Checking that file %s should not exist'
+            #                     % filename)
+            #    self.assertFalse(os.path.isfile(filename))
+            #    filename = os.path.join(path, 'pbs.slice', 'pbs-%s.slice'
+            #                            % systemd_escape(jid))
+            #    self.logger.info('Checking that file %s should not exist'
+            #                     % filename)
+            #    self.assertFalse(os.path.isfile(filename))
```
minghui-liu, line 2701: Please remove commented code.

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -589,25 +621,29 @@ def printjob_info(jobid, include_attributes=False):
...
     try:
+        err_nonsplit = ''
         pbs.logmsg(pbs.EVENT_DEBUG4, 'Running: %s' % cmd)
         process = subprocess.Popen(cmd, shell=False,
                                    stdout=subprocess.PIPE,
-                                   stderr=subprocess.PIPE)
-        out, err = process.communicate()
+                                   stderr=subprocess.PIPE,
+                                   universal_newlines=True)
+        out_nonsplit, err_nonsplit = process.communicate()
```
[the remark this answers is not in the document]
alexis-cousein: As always, I was aiming for minimal changes in the remaining code. But I will implement your suggestion.

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -136,6 +138,17 @@ else:
...
+    if sys.version_info[0] < 3:
+        PYTHON2 = True
+    else:
+        PYTHON2 = False
+
+    try:
+        bytearray
+        BYTEARRAY_EXISTS = True
+    except Exception:
+        BYTEARRAY_EXISTS = False
```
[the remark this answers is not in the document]
alexis-cousein: Yes. The hook is also useful on versions of PBSPro that still used Python 2.5.

[Yes, there are constructs like "except Exception as exc:" that no longer work with Python 2.5, but I find it important to minimise work to port back to Python 2.5, especially those that cannot be fixed by simple substitution.]

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -589,25 +621,29 @@ def printjob_info(jobid, include_attributes=False):
         cmd.append('-a')
     cmd.append(jobfile)
     try:
+        err_nonsplit = ''
```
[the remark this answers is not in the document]
alexis-cousein: That is a remnant of some formerly badly arranged code where err was being used when encountering an exception. Now that err_nonsplit is only being used if the communicate() works, it is indeed no longer needed.

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -3774,12 +3835,14 @@ class CgroupUtils(object):
...
-                return {}
-            assigned['cpuset.cpus'] += corelist[:needed]
+            if needed >= 1:
+                if needed > len(cores):
+                    pbs.logmsg(pbs.EVENT_DEBUG4, '%s: %d ncpus still needed' %
+                               (caller_name(), len(cores) - needed))
+                    return {}
+                else:
+                    corelist = sorted(cores)
+                    assigned['cpuset.cpus'] += corelist[:needed]
```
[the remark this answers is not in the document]
alexis-cousein: Indeed, that is shorter. And it's true that if needed is not zero, then it cannot be larger than len(cores). That was an attempt to skip _the whole_ block (including checking the length of cores) for the most common case, in which needed is zero when we arrive here.

Rearranged, with most common conditions at the start (which does introduce a pass statement).

```
            if needed <= 0:
                # we're done (will always hold if use_hyperthreads
                # and ncpus_are_cores are both True)
                pass
            elif needed <= len(cores):
                    corelist = sorted(cores)
                    assigned['cpuset.cpus'] += corelist[:needed]
            else:
                    pbs.logmsg(pbs.EVENT_DEBUG4, '%s: %d ncpus still needed' %
                               (caller_name(), len(cores) - needed))
                    return {}
```

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -4589,7 +4657,7 @@ class CgroupUtils(object):
                 except Exception:
                     jobdict = dict()
                     pbs.logmsg(pbs.EVENT_DEBUG2,
-                               '%s: Failed to resyncing local job data' %
+                               '%s: Failed gathering local job data' %
```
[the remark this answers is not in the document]
alexis-cousein: Yes. I'd rather not substitute one grammatical error for another ;-).

test/tests/functional/pbs_cgroups_hook.py
alexis-cousein: Quite -- thanks, much better.

test/tests/functional/pbs_cgroups_hook.py
alexis-cousein: That was left there for a reason, to make it easier to reuse the test for an older hook.

Since get_cgroup_job_dir already lists the other possible paths, replaced with a comment containing a reference to that function.

alexis-cousein:
Attaching logs:
for #1817 , here's an output of the old hook for the ncpus_are_cores PTL test when the configuration file for it (more correctly) enables use_hyperthreads (the current test disables it, which has the side effect of not testing ncpus_are_cores, which is basically a no-op when HT is entirely disabled).

[ptl.ncpus_are_cores.FAIL.txt](https://github.com/openpbs/openpbs/files/4769651/ptl.ncpus_are_cores.FAIL.txt)

alexis-cousein:
Here are test results on TH2 for the new hook:
[ptl.cgroups_hook.r8.txt](https://github.com/openpbs/openpbs/files/4769669/ptl.cgroups_hook.r8.txt)
[ptl.cgroups_hook.u18.txt](https://github.com/openpbs/openpbs/files/4769672/ptl.cgroups_hook.u18.txt)
[ptl.cgroups_hook.r7.txt](https://github.com/openpbs/openpbs/files/4769673/ptl.cgroups_hook.r7.txt)
[ptl.cgroups_hook.s15.txt](https://github.com/openpbs/openpbs/files/4769675/ptl.cgroups_hook.s15.txt)

alexis-cousein:
For #1818 , here is a new test MoM log for a RedHat 8 machine, pbi-39-r8

[mom.log](https://github.com/openpbs/openpbs/files/4769789/mom.log)

It confirms correct detection of the systemd version, whereas on older tests on Python 3, the version detected would always be 0.

06/11/2020 12:34:26.433732;0400;pbs_python;Hook;pbs_python;main: Event type is exechost_periodic
06/11/2020 12:34:26.433742;0800;pbs_python;Hook;pbs_python;_get_systemd_version: Method called
06/11/2020 12:34:26.458846;0800;pbs_python;Hook;pbs_python;__init__: systemd version seems to be 239

alexis-cousein:
Unfortunately, I see from other tests that systemctl's interface to show the version of the running systemd is not stable (it is hazardous to usde systemctl --version, which does have a more stable interface, since I have seen VMs and containers where systemctl exists but systemd is not running!).

[cousein@pbi-04-r7 ~]$ systemctl --system show --property=Version 
Version=systemd 208
alexis@tc72host:~> systemctl --system show --property=Version 
Version=228

So the regular expression string will have to be adapted to cater for both versions for that issue to go away on _all_ systems.

alexis-cousein:
Changed the regular expression in _get_systemd_version to add a \D* to ignore a 'systemd ' that may be inserted in front of the version number in older releases of systemd.

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -3835,14 +3830,18 @@ class CgroupUtils(object):
...
+            if needed <= 0:
+                # we're done (will always hold if use_hyperthreads
+                # and ncpus_are_cores are both True)
+                pass
+            elif needed <= len(cores):
+                corelist = sorted(cores)
+                assigned['cpuset.cpus'] += corelist[:needed]
+            else:
+                pbs.logmsg(pbs.EVENT_DEBUG4, '%s: %d ncpus still needed' %
+                           (caller_name(), len(cores) - needed))
```
minghui-liu, line 3842: `len(cores) - needed` is negative in this case (needed > len(cores)). Do you mean `needed - len(cores)`?

alexis-cousein: I left the message as-is, but you are right that it is confusing; if you're missing 3 cpus it will print "-3 ncpus still needed".

Original code:
```
            if needed > len(corelist):
                pbs.logmsg(pbs.EVENT_DEBUG4, '%s: %d ncpus still needed' %
                           (caller_name(), len(corelist) - needed))
                return {}
```

So you're right that whoever the original author was probably meant needed - len(cores), or in his case needed - len(corelist).

I'll also change the error messages. "Still needed" doesn't say after what they are 'still' needed.

```
            else:
                pbs.logmsg(pbs.EVENT_DEBUG4, '%s: %d more ncpus needed than available' %
                           (caller_name(), needed - len(cores))
                return {}
```

alexis-cousein: But it's all a bit irrelevant, since at the top of the function we have:
```
            avail = len(cores)
            needed = int(requested['ncpus'])
            if self.cfg['use_hyperthreads'] and self.cfg['ncpus_are_cores']:
                needed *= node.cpuinfo['hyperthreads_per_core']
            if needed > avail:
                pbs.logmsg(pbs.EVENT_DEBUG4, 
                           '%s: Insufficient ncpus: %s needed, %s available'
                           % (caller_name(), needed, avail))
                return {}
```
The error message "should never" be printed, unless there is a logic error in the code in between these two.

Checking the code, "avail" is len(cores). In the block between the first and second check on ncpus available, we decrement needed whenever we remove an element in cores. So if the first check passes, it is _impossible_  for the second check to fail, unless someone modifies the code. [I would leave second the guard there -- we definitely do not want to _silently_ assign too few ncpus if we make an error in modified code between the two checks].

alexis-cousein: Note that the first message _is_ actually printed in valid circumstances: if vnode_per_numa_node is False, then jobs are tentatively placed on each socket first, and only if they cannot be fit on a single socket placed on the entire host.

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -2699,10 +2736,10 @@ class CgroupUtils(object):
         except NameError:
             basestring = str
         # We want a view/iterator object, not something that creates copied
-        # lists. Unfortunately, iteritems/viewitems() no longer exists in Py3
+        # lists. iteritems/viewitems() no longer exists in Py3
         # since items() itself is now implemented as a view object.
-        dict_iter_object = (config_dict.items() if sys.version_info[0] >= 3
-                            else config_dict.iteritems())
+        dict_iter_object = (config_dict.iteritems() if PYTHON2
+                            else config_dict.items())
```
bhroam, line 2742: Can't you use items() for both?  In python2 it is just more efficient to use iteritems()?

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -1350,21 +1368,46 @@ def get_cgroup_job_dir(self, subsys, jobid, host):
...
-                   os.path.join(basedir, 'pbs.service/jobid', jobid),
-                   os.path.join(basedir, 'pbs_jobs.service/jobid', jobid),
-                   os.path.join(basedir, 'pbs.slice',
-                                'pbs-%s.slice' % systemd_escape(jobid)),
-                   os.path.join(basedir, 'pbs',
-                                'pbs-%s.slice' % systemd_escape(jobid))]
+        # One of these should exist depending on platform.
+        #
+        # Entries for older hooks deleted. I am leaving them as comments should
+        # this be used to test old hooks, but DO NOT put them at the top
```
bhroam, line 1374: Don't leave code as comments.  If we need to test older hooks, we can dig it out of git.   That's what code revision control is for.

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -136,6 +138,14 @@ else:
     import fcntl
     import pwd
 
+    PYTHON2 = sys.version_info[0] < 3
+
+    try:
+        bytearray
+        BYTEARRAY_EXISTS = True
+    except Exception:
+        BYTEARRAY_EXISTS = False
```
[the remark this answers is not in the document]
alexis-cousein: Correct. I'll change it.

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -376,17 +406,17 @@ def decode_list(data):
...
+        if isinstance(item, list):
             item = decode_list(item)
         elif isinstance(item, dict):
             item = decode_dict(item)
+        elif PYTHON2:
+            if isinstance(item, unicode):
+                item = item.encode('utf-8')
+            elif BYTEARRAY_EXISTS and isinstance(item, bytearray):
+                item = str(item)
+        elif isinstance(item, (bytes, bytearray)):
```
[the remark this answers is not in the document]
alexis-cousein: No. 

If we have Python 2, then "bytes" either does not exist or is an alias for "str" (which means all "str" methods work for it.) bytearray either does not exist or it does, and if it does, it needs to be converted to bytes/str.

If we have Python 3 or beyond (if PYTHON2 is false), then both "bytes" and "bytearray" necessarily exist, and none of them are "str" ("str" in Python 3 is the old "unicode" of Python 2). In that case, they need to be converted to "str", specifying which encoding was used in the bytes or bytearray.

test/tests/functional/pbs_cgroups_hook.py
alexis-cousein: OK -- I'll simply specify the paths here, but not the code.

src/hooks/cgroups/pbs_cgroups.PY
alexis-cousein: No, you can't use items() for both.

alexis-cousein:
I can confirm from the latest TH2 pbi-04-r7 mom logs that indeed issue #1818 is now also fixed for older systemd versions that report their version attribute differently:

```06/12/2020 13:45:21.691645;0800;pbs_python;Hook;pbs_python;_get_systemd_version: Method called
06/12/2020 13:45:21.742625;0800;pbs_python;Hook;pbs_python;__init__: systemd version seems to be 208
```

Whereas on previous versions we saw:
```

06/11/2020 13:34:47.009429;0800;pbs_python;Hook;pbs_python;_get_systemd_version: Method called
06/11/2020 13:34:47.049513;0800;pbs_python;Hook;pbs_python;__init__: systemd version seems to be 0
```

PR #1823, opened 2020-06-11
[WIP] Bifurcated DB and cache saves and used pbs_list_link in DB layer

subhasisb:
#### Describe Bug or Feature

This set of changes are not adding any new feature, rather is primarily about code refactoring. Making the code more structured, while addressing a few inconsistencies in the code of how we handle attributes (ATR_MODIFY vs ATR_MODCACHE) and how we save unnecessarily even those attributes that were designed to not need persisting. This change should bring better structure (and discipline) to the code even for the current usage, but is also making the code base ready for the multiple server functionality to fit in, eventually. One of the things we want to ensure is that this change does not degrade any performance of the current PBS functionalities (so we will be adding performance test comparisons to the PR)

- __Don't replace full attribute list in database__
  For some objects like the server, scheduler, and node objects we do not depend any more on removing all the attributes and writing them back all again. This change updates/adds/deletes the exact attribute(s) that were changed/added/removed. This change required that in some cases where mgr_server_set() function calls depended on just saving the updated attribute list, we now need to find out which ones were actually deleted, and call database functions to delete those specific ones

-  __Bifurcated DB and cache saves and used pbs_list_link in DB layer__
  Separated the NO_SAVM attributes into a separate list called the cache_attr_list. This is skipped for single server, since we do not even need to encode such parameters (since heap is the distributed cache in that case). Also removed the need to pass the type of save and to track whether an object was modified or not. Introduced functions obj_qs_modified() that will store a hash per object (inside the object) and can check whether the quick save area was changed or not (the attributes already have flags to check for modification, but the quick save area had no such equivalent). For the current database library based on postgres, the cache_attributes will be written to the database itself.

- __passing pbs_list to db layer instead of converting to attribute array__
     This is an optimization that saves converting the attribute list contained in each object to a array of attributes first before handing over to db layer. Now, the db layer deals with the pbs_list_head datatype

- __Refactored Node save semantics__
     The node save code was quite different. Updated to use the same DB save routines

- __Added ATR_VFLAG_MODIFY to all places with ATR_VFLAG_MODCACHE__
    The DB routines were incorrectly checking MODCACHE for db saves, rather it should only rely on MODIFY (MODCACHE is for stat cache). However it is not consistently flagged everywhere and hence all lines with MODCACHE is not added with MODIFY

- __saving job on req_modifyjob__
- __Fixed unset attribute__
    Unset of an attribute. was relying on the fact that the entire attribute array will be saved. This is not required and all attributes that re unset can be removed directly

- __Saving all attributes for history job after job termination__
  While we do not want to save NO_SAVM attributes to disk, history jobs needs things like comments saved. Thus in the case of job termination and history jobs we save back all NO_SAVM attributes as well

Profile: regression
|--|--|--|--|--|--|--|
|2231|1815|0|0|0|0|1815|

[perf tests.txt](https://github.com/openpbs/openpbs/files/4775852/perf.tests.txt)


src/lib/Libdb/db_postgres_impl.c
```diff
@@ -867,41 +833,71 @@ pbs_db_save_obj(pbs_db_conn_t *conn, pbs_db_obj_info_t *obj, int savetype)
...
+ *
+ * @return      Error code
+ * @retval	 0 - success
+ * @retval	-1 - failure
+ */
+int
+dist_cache_del_attrs(char *id, pbs_db_attr_list_t *attr_list)
+{
+	attr_list->attr_count = 0;
+	CLEAR_HEAD(attr_list->attrs);
```
suresh-thelkar, line 901: Is it enough to just clear the head pointer. Just wanted to know don't we need to free all the attributes as well.

src/lib/Libdb/db_postgres_impl.c
```diff
@@ -867,41 +833,71 @@ pbs_db_save_obj(pbs_db_conn_t *conn, pbs_db_obj_info_t *obj, int savetype)
  *
  */
 int
-pbs_db_delete_attr_obj(pbs_db_conn_t *conn, pbs_db_obj_info_t *obj, void *obj_id, pbs_db_attr_list_t *attr_list) {
-	return (db_fn_arr[obj->pbs_db_obj_type].pg_db_del_attr_obj(conn, obj, obj_id, attr_list));
+pbs_db_delete_attr_obj(pbs_db_conn_t *conn, pbs_db_obj_info_t *obj, void *obj_id, char *sv_time, pbs_db_attr_list_t *cache_attr_list, pbs_db_attr_list_t *db_attr_list)
```
suresh-thelkar, line 836: It looks like cache_attr_list parameter is missing from function header. Please add it.

src/server/attr_recov_db.c
```diff
@@ -78,73 +70,74 @@ extern int resc_access_perm;
...
- * @param[in]	attr_value - value of the attribute
- * @param[in]	attr_flags - Flags associated with the attribute
+ *	Compute and check whether the quick save area has been modified
  *
- * @retval - Pointer to the newly created attribute
- * @retval - NULL - Failure
- * @retval - Not NULL - Success
+ * @param[in]	qs  - pointer to the quick save area
+ * @param[in]	len - length of the quick save area
+ * @param[in]   oh  - pointer to a opaque value of current quick save area signature/hash
```
suresh-thelkar, line 80: Please rename oh->oldhash. Also it looks like this also acts as OUT parameter. Can you please update that.

src/server/attr_recov_db.c
```diff
@@ -78,73 +70,74 @@ extern int resc_access_perm;
...
-	strcpy(psvrat->al_name, attr_name);
-	psvrat->al_nameln = strlen(attr_name);
-	p = psvrat->al_name + psvrat->al_nameln + 1;
-
-	if (attr_resc && strlen(attr_resc) > 0) {
-		psvrat->al_resc = p;
-		strcpy(psvrat->al_resc, attr_resc);
-		psvrat->al_rescln = strlen(attr_resc);
-		p = p + psvrat->al_rescln + 1;
+	if (SHA1((const unsigned char *) qs, len, (unsigned char *) &hash) == NULL) {
```
suresh-thelkar, line 91: Remove brackets as it is a single line statement.

src/include/pbs_sched.h
```diff
@@ -119,6 +119,7 @@ typedef struct pbs_sched {
 	struct preempt_ordering preempt_order[PREEMPT_ORDER_MAX + 1];
 	/* sched object's attributes  */
 	attribute sch_attr[SCHED_ATR_LAST];
+	char	  sc_savetm[DB_TIMESTAMP_LEN + 1];
```
suresh-thelkar, line 122: Please put a single space between char and sc_savetm.

src/include/server.h
```diff
@@ -188,10 +188,11 @@ struct server {
 	struct server_qs {
 		int		sv_numjobs;	/* number of job owned by server   */
 		int		sv_numque;	/* nuber of queues managed          */
-		long long	sv_jobidnumber;	/* next number to use in new jobid  */
-		time_t		sv_savetm;	/* time of server db update         */
+		long long sv_jobidnumber;	/* next number to use in new jobid  */
```
suresh-thelkar, line 191: Please put a tab between long long and sv_jobidnumber and others just to align it with the other data members.

src/server/attr_recov_db.c
```diff
@@ -78,73 +70,74 @@ extern int resc_access_perm;
...
+encode_single_attr_db(struct attribute_def *padef, struct attribute *pattr, pbs_db_attr_list_t *cache_attr_list, pbs_db_attr_list_t *db_attr_list, int all)
+{
+	pbs_list_head *lhead;
+	int rc = 0;
+	int chkflag = 0;
+	
+	if (!all) // && (pbs_conf.pbs_max_servers == 1) /* to be implemented later */
+		chkflag |= ATR_DFLAG_NOSAVM; /* skip all NOSAVM attributes for single server case, so it stays in process memory */
+
+	if (!(padef->at_flags & chkflag)) {
```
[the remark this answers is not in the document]
subhasisb: Updated the code

src/include/pbs_db.h
```diff
@@ -319,15 +309,16 @@ struct pbs_db_query_options {
...
+
+#define PBS_DB_SVR          0
+#define PBS_DB_SCHED        1
+#define PBS_DB_QUEUE        2
+#define PBS_DB_NODE         3
+#define PBS_DB_MOMINFO_TIME 4
+#define PBS_DB_JOB          5
+#define PBS_DB_JOBSCR       6
+#define PBS_DB_RESV         7
+#define PBS_DB_NUM_TYPES    8
```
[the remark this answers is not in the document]
subhasisb: I changed them to arrange in the order that made sense - eg the order they are loaded in init. But it is just aesthetics

src/lib/Libdb/db_postgres_impl.c
subhasisb: It is not enough. It will be implemented when the dist cache is actually implemented. This is just a placeholder for now.

src/include/server.h
subhasisb: formatted to remove alignment

src/include/svrfunc.h
```diff
@@ -81,33 +81,31 @@ extern int schedule_high(pbs_sched *);
...
+extern int   setup_nodes(void);
+extern int   setup_resc(int);
+extern void  update_job_node_rassn(job *, attribute *, enum batch_op);
+extern void  mark_node_down(char *, char *);
+extern void  mark_node_offline_by_mom(char *, char *);
+extern void  clear_node_offline_by_mom(char *, char *);
+extern void  mark_which_queues_have_nodes(void);
+extern void  set_sched_sock(int, pbs_sched *);
+extern void  pbs_close_stdfiles(void);
+extern int   is_job_array(char *id);
```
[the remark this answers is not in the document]
subhasisb: done

src/lib/Libdb/db_postgres_impl.c
subhasisb: added

src/lib/Libattr/attr_node_func.c
```diff
@@ -351,7 +351,7 @@ encode_state(const attribute *pattr, pbs_list_head *ph, char *aname, char *rname
 	if (rtnl)
 		*rtnl = pal;
 
-	return (0);             /*success*/
+	return (1);             /*success*/
```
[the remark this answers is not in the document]
subhasisb: Changed

subhasisb:
Addressed your comments @suresh-thelkar  please check

src/lib/Libdb/db_postgres_job.c
```diff
@@ -405,60 +415,78 @@ load_job(const  PGresult *res, pbs_db_job_info_t *pj, int row)
...
-	else if (savetype == PBS_UPDATE_DB_QUICK)
-		stmt = STMT_UPDATE_JOB_QUICK;
-	else
+	if (savetype & OBJ_SAVE_NEW)
 		stmt = STMT_INSERT_JOB;
 
-	rc = pg_db_cmd(conn, stmt, params);
-
-	free(raw_array);
+	if (stmt != NULL) {
```
arungrover, line 475: Can the statement be NULL at this point? It does not seem like it can every be NULL. If it can be NULL, then maybe it would be better to move free(raw_array) out of this block.

src/lib/Libdb/db_postgres_svr.c
```diff
@@ -324,37 +306,30 @@ pbs_db_get_schema_version(pbs_db_conn_t *conn, int *db_maj_ver, int *db_min_ver)
...
-	if ((len = convert_db_attr_list_to_array(&raw_array, attr_list)) <= 0)
+	if ((len = attrlist_2_dbarray_ex(&raw_array, attr_list, 1)) <= 0)
 		return -1;
 
-
 	SET_PARAM_BIN(conn, raw_array, len, 0);
 
-	if (pg_db_cmd(conn, STMT_REMOVE_SVRATTRS, 1) != 0)
+	if (pg_db_cmd(conn, STMT_REMOVE_SVRATTRS, 1) != 0) {
+		PQclear(conn->conn_resultset);
```
arungrover, line 322: you are missing a free(raw_array) here

src/lib/Libdb/db_postgres_impl.c
```diff
@@ -815,6 +815,53 @@ pbs_db_destroy_connection(pbs_db_conn_t *conn)
...
+
+		case PBS_DB_RESV:
+				cache_attrs = &((pbs_db_resv_info_t *) obj->pbs_db_un.pbs_db_resv)->cache_attr_list;
+				db_attrs    = &((pbs_db_resv_info_t *) obj->pbs_db_un.pbs_db_resv)->db_attr_list;
+				break;
+	}
+
+	/*  for postgres, implement distributed cache by saving to database */
+	if (cache_attrs && db_attrs && (cache_attrs->attr_count > 0)) {
+		list_move(&cache_attrs->attrs, &db_attrs->attrs);
```
suresh-thelkar, line 861: Just wanted to know whether we are adding cache_attrs on top of db_attrs here so that both are saved to database. Also as we are going to change this logic when distributed cache is implemented can we have this piece of code in a separate function which will make things easier in future.

src/server/job_func.c
```diff
@@ -1592,9 +1588,9 @@ resc_resv_alloc(void)
 {
 	resc_resv	*resvp;
 
-	resvp = (resc_resv *)malloc(sizeof(resc_resv));
+	resvp = (resc_resv *) calloc(1, sizeof(resc_resv));
```
suresh-thelkar, line 1591: As we are doing calloc do we still need memset in the below lines.

src/server/job_recov_db.c
```diff
@@ -519,60 +339,138 @@ job_recov_db_spl(pbs_db_job_info_t *dbjob)
...
 
-	if (pbs_db_begin_trx(conn, 0, 0) !=0)
-		goto db_err;
+	rc = pbs_db_load_obj(conn, &obj);
+	if (rc == -2)
+		return pjob; /* no change in job, return the same job */
 
-	strcpy(dbjob.ji_jobid, jid);
+	if (rc == 0) {
+		pjob = job_recov_db_spl(&dbjob, pjob);
```
suresh-thelkar, line 369: Please remove the branches as it is single line statement.

src/server/node_recov_db.c
```diff
@@ -88,152 +83,187 @@
...
+{
+	if (pdbnd->nd_name && pdbnd->nd_name[0] != 0) {
+		pnode->nd_name = strdup(pdbnd->nd_name);
+		if (pnode->nd_name == NULL)
+			return -1;
+	}
+	else
+		pnode->nd_name = NULL;
+
+	if (pdbnd->nd_hostname && pdbnd->nd_hostname[0]!=0) {
```
suresh-thelkar, line 110: Put a space around !=0 between pdbnd->nd_hostname[0] and 0.

src/server/queue_func.c
```diff
@@ -112,10 +112,9 @@ que_alloc(char *name)
...
 
-
-	pq = (pbs_queue *)malloc(sizeof(pbs_queue));
+	pq = (pbs_queue *) calloc(1, sizeof(pbs_queue));
 	if (pq == NULL) {
-		log_err(errno, "que_alloc", "no memory");
+		log_err(errno, __func__, "no memory");
 		return NULL;
 	}
 	(void)memset((char *)pq, (int)0, (size_t)sizeof(pbs_queue));
```
suresh-thelkar, line 120: Can we now remove memset as we are using calloc.

src/server/queue_recov_db.c
```diff
@@ -133,18 +119,18 @@ svr_to_db_que(pbs_queue *pque, pbs_db_que_info_t *pdbque, int updatetype)
...
 	pque->qu_qs.qu_type = pdbque->qu_type;
-	pque->qu_qs.qu_ctime = pdbque->qu_ctime;
-	pque->qu_qs.qu_mtime = pdbque->qu_mtime;
 
-	if ((decode_attr_db(pque, &pdbque->attr_list, que_attr_def,
-		pque->qu_attr, (int) QA_ATR_LAST, 0)) != 0)
+	if ((decode_attr_db(pque, &pdbque->cache_attr_list, &pdbque->db_attr_list, que_attr_def, pque->qu_attr, (int) QA_ATR_LAST, 0)) != 0)
 		return -1;
 
+	obj_qs_modified(&pque->qu_qs, sizeof(pque->qu_qs), pque->qs_hash);
```
suresh-thelkar, line 130: Can we collect the return value of the function obj_qs_modified and handle accordingly. It is because this function is also returning -1 in case of failure.

src/server/req_manager.c
```diff
@@ -1207,22 +1213,17 @@ struct batch_request *preq;
...
-		(void)sprintf(log_buffer, msg_manager, msg_man_cre,
-			preq->rq_user, preq->rq_host);
-		log_event(PBSEVENT_ADMIN, PBS_EVENTCLASS_QUEUE, LOG_INFO,
-			pque->qu_qs.qu_name, log_buffer);
-		mgr_log_attr(msg_man_set, plist,
-			PBS_EVENTCLASS_QUEUE,
-			preq->rq_ind.rq_manager.rq_objname, NULL);
+		que_save_db(pque);
+
+		(void)sprintf(log_buffer, msg_manager, msg_man_cre, preq->rq_user, preq->rq_host);
```
suresh-thelkar, line 1224: Please use snprintf.

src/server/req_manager.c
```diff
@@ -1417,37 +1412,53 @@ void
...
+			append_link(&unsetlist, &plist->al_link, plist);
+			plist = tmp;
+			continue;
 		}
 		plist = (struct svrattrl *)GET_NEXT(plist->al_link);
 	}
+	
+	/* if the unsetist has attributes, call server_unset to remove them separately */
+	ulist = (svrattrl *)GET_NEXT(unsetlist);
+	if (ulist) {
```
suresh-thelkar, line 1443: Please remove the brackets as it is single line statement.

src/server/req_manager.c
```diff
@@ -1417,37 +1412,53 @@ void
...
-					reply_badattr(PBSE_ATTRRO, bad_attr, plist, preq);
-					return;
-				}
-				break;
+		if (strcasecmp(plist->al_name, ATTR_aclroot) == 0) {
+			if (!is_local_root(preq->rq_user, preq->rq_host)) {
+				reply_badattr(PBSE_ATTRRO, bad_attr, plist, preq);
+				return;
 			}
+		} else if (plist->al_atopl.value == NULL || plist->al_atopl.value[0] == '\0') {
```
suresh-thelkar, line 1431: Just a question. Are we assuming that if a value for an attribute is not provided it is meant to be unset ?

src/server/req_manager.c
```diff
@@ -1417,37 +1412,53 @@ void
...
 		}
 		plist = (struct svrattrl *)GET_NEXT(plist->al_link);
 	}
+	
+	/* if the unsetist has attributes, call server_unset to remove them separately */
+	ulist = (svrattrl *)GET_NEXT(unsetlist);
+	if (ulist) {
+		mgr_unset_attr(server.sv_attr, svr_attr_def, SRV_ATR_LAST, ulist,
+			preq->rq_perm, &bad_attr, (void *)&server, PARENT_TYPE_SERVER, INDIRECT_RES_CHECK);
+	}
```
suresh-thelkar, line 1446: Don't we need to free list ?

src/server/req_modify.c
```diff
@@ -398,16 +398,13 @@ req_modifyjob(struct batch_request *preq)
...
-		(void)job_save(pjob, SAVEJOB_FULL);
-	}
-	(void)sprintf(log_buffer, msg_manager, msg_jobmod,
-		preq->rq_user, preq->rq_host);
-	log_event(PBSEVENT_JOB, PBS_EVENTCLASS_JOB, LOG_INFO,
-		pjob->ji_qs.ji_jobid, log_buffer);
+	
+	job_save_db(pjob); /* we must save the updates anyway, if any */
+	
+	(void)sprintf(log_buffer, msg_manager, msg_jobmod, preq->rq_user, preq->rq_host);
```
suresh-thelkar, line 404: Please use snprintf.

src/include/pbs_nodes.h
```diff
@@ -289,9 +289,9 @@ struct	pbsnode {
 	unsigned short	 	 nd_ntype;	/* node type */
 	unsigned short		 nd_accted;	/* resc recorded in job acct */
 	struct pbs_queue	*nd_pque;	/* queue to which it belongs */
-	int			 nd_modified;	/* flag indicating whether state update is required */
-	struct devices device;
-	attribute		 nd_attr[ND_ATR_LAST];
+	struct 				devices device;
```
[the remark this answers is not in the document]
subhasisb: Formatted the entire structure

src/include/server.h
```diff
@@ -178,42 +178,38 @@ enum srv_atr {
...
 
-
 struct server {
 	struct server_qs {
-		int		sv_numjobs;	/* number of job owned by server   */
-		int		sv_numque;	/* nuber of queues managed          */
-		long long	sv_jobidnumber;	/* next number to use in new jobid  */
-		time_t		sv_savetm;	/* time of server db update         */
+		int sv_numjobs;			  /* number of job owned by server   */
+		int sv_numque;			  /* nuber of queues managed          */
```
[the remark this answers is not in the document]
subhasisb: fixed

src/lib/Libdb/db_postgres_attr.c
```diff
@@ -91,123 +167,147 @@ struct pg_array {
...
+			*attr_resc = '\0';
+			attr_resc++;
 		}
+
+		if ((attr_value = strchr(attr_flags, '.'))) {
+			*attr_value = '\0';
+			attr_value++;
+		}
+
+		if (!(pal = make_attr(attr_name, attr_resc, attr_value, atol(attr_flags))))
```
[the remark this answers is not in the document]
subhasisb: Why? Per the man pages: The atol subroutine is equivalent to the strtol subroutine where the value of the EndPointer parameter is a null pointer and the Base parameter is a value of 10.

src/lib/Libdb/db_postgres_common.c
```diff
@@ -100,12 +102,11 @@ pg_set_error(pbs_db_conn_t *conn, char *fnc, char *msg)
 	while ((p >= str) && (*p == '\r' || *p == '\n'))
 		*p-- = 0; /* supress the last newline */
 
-	conn->conn_db_err = malloc(strlen(fnc) + strlen(msg) +
-		strlen(str) + sizeof(fmt) + 1);
+	conn->conn_db_err = malloc(strlen(fnc) + strlen(msg) + strlen(msg2) + strlen(str) + sizeof(fmt) + 1);
```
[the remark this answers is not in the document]
subhasisb: You are right, i will add a check

src/lib/Libdb/db_postgres_job.c
subhasisb: yes it could theretically be NULL, if it did not enter any of the above blocks - you are right, i move the free(raw_array) outisde

src/lib/Libdb/db_postgres_svr.c
subhasisb: ah thanks, added

src/lib/Libdb/db_postgres_impl.c
subhasisb: yes, so we are merging the two separate lists (which happens only in case of multiserver) to one list for the postgres implementation. So cache_attrs being moved to db_attrs (with a simple few link changes). This will change for aerospike implemnetation where aerospike provides a inbuilt caching mechanism. It could also change, if we decide to add a dist cache support like memcached. 

We don't know what the interface will look like, so making a function will not help now.

src/server/job_func.c
subhasisb: dont know how it became calloc. Reverted back.

src/server/job_recov_db.c
subhasisb: done

src/server/node_recov_db.c
subhasisb: done

src/server/queue_func.c
subhasisb: reverted to malloc

src/server/queue_recov_db.c
subhasisb: Actually, obj_qs_modified does not return error. SHA1 should not fail since we provide it with exact buffer we need, and it merely hashes it. If it fails, there is probably severe problems in the system setup itself, so we simply quit PBS

src/server/req_manager.c
subhasisb: done

src/server/svr_recov_db.c
```diff
@@ -447,49 +374,37 @@ sched_recov_db(char *sname)
...
+done:
+	free_db_attr_list(&dbsched.db_attr_list);
+	free_db_attr_list(&dbsched.cache_attr_list);
 
-	if (rc != 0)
-		goto db_err;
-
-	return (0);
+	if (rc != 0) {
+		sprintf(log_buffer, "Failed to save sched %s ", ps->sc_name);
```
suresh-thelkar, line 401: Please use snprintf.

src/server/req_manager.c
subhasisb: Yes, exactly. Earlier, we used to write the entire server structure directly, so any removed attributes from heap woudl also get "removed" from database. However, now we only update selected attributes. Thus, if we use this api to unset something (which is actually tested by some test cases), then we need to detect those and actually call unset_xxx routines to remove from database. I will add a comment to explain this.

src/server/req_manager.c
subhasisb: done

src/server/req_manager.c
subhasisb: just like mgr_set_attr(), mgr_unset_attr() frees each attribute inside.

src/server/req_modify.c
subhasisb: Changed to log_eventf

src/server/svr_recov_db.c
subhasisb: Done

subhasisb:
Addressed review comments, please check further

src/lib/Libdb/db_postgres_attr.c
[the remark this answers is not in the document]
arungrover: We usually tend to avoid using ato* functions. These functions return numerical value even in case of failures (like when the input itself is non numeric)

subhasisb:
Reran tests:
Profile: regression
|--|--|--|--|--|--|--|
|2235|1815|0|0|0|0|1815|

src/lib/Libdb/db_postgres_attr.c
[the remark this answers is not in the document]
subhasisb: ah ok - the data is not from user input, from the database actually, so is entirely numeric - it can get corrupted in very bad cases of course, but there is not much to salvage at that point anyway. Nevertheless, i changed to strtol() - please take a peek

src/server/node_recov_db.c
```diff
@@ -88,152 +83,187 @@
...
-						&& (pnode->nd_attr[ND_ATR_Sharing].at_val.at_long != VNS_DFLT_SHARED)));
-	numattr = vnode_sharing;
-	psvrl = (svrattrl *)GET_NEXT(wrtattr);
-	while (psvrl) {
-		if ((strcmp(psvrl->al_name, ATTR_rescavail) == 0) &&
-				(strcmp(psvrl->al_resc, "ncpus") == 0)) {
+	/* MSTODO: how can we optimize this loop - eliminate this? */
+	psvrl = (svrattrl *) GET_NEXT(pdbnd->db_attr_list.attrs);
+	while (psvrl != NULL) {
+		if ((strcmp(psvrl->al_name, ATTR_rescavail) == 0) && (strcmp(psvrl->al_resc, "ncpus") == 0)) {
```
arungrover, line 239: maybe a check for !wrote_np along with other two strcmp might be helpful and avoid unnecessary strcmp

src/server/node_recov_db.c
```diff
@@ -327,32 +320,37 @@ node_recov_db_raw(void *nd, pbs_list_head *phead)
...
+		strcpy(pnode->nd_savetm, dbnode.nd_savetm);
+done:
+	free_db_attr_list(&dbnode.db_attr_list);
+	free_db_attr_list(&dbnode.cache_attr_list);
+	
+	if (rc != 0) {
+		strcpy(log_buffer, "node_save failed ");
+		if (conn->conn_db_err != NULL)
+			strncat(log_buffer, conn->conn_db_err, LOG_BUF_SIZE - strlen(log_buffer) - 1);
+		log_err(-1, __func__, log_buffer);
```
arungrover, line 350: Please use log_errf

src/server/queue_recov_db.c
```diff
@@ -162,92 +148,95 @@ db_to_svr_que(pbs_queue *pque, pbs_db_que_info_t *pdbque)
  * @retval	1 - Failure
  *
  */
+
 int
-que_save_db(pbs_queue *pque, int mode)
+que_save_db(pbs_queue *pque)
```
arungrover, line 153: Please change the function header

src/server/queue_recov_db.c
```diff
@@ -162,92 +148,95 @@ db_to_svr_que(pbs_queue *pque, pbs_db_que_info_t *pdbque)
...
-	return (-1);
+done:
+	free_db_attr_list(&dbque.db_attr_list);
+	free_db_attr_list(&dbque.cache_attr_list);
+	
+	if (rc != 0) {
+		strcpy(log_buffer, "que_save failed ");
+		if (conn->conn_db_err != NULL)
+			strncat(log_buffer, conn->conn_db_err, LOG_BUF_SIZE - strlen(log_buffer) - 1);
+		log_err(-1, __func__, log_buffer);
```
arungrover, line 180: Please use log_errf and a proper errno.

src/server/req_quejob.c
```diff
@@ -3325,9 +3285,7 @@ long long get_next_svr_sequence_id(void)
  */
 void reset_svr_sequence_window(void)
 {
-	svr_sequence_window_count = 0;
 	server.sv_qs.sv_jobidnumber = 0;
```
ashwathraop, line 3288: Since this function got reduced to one line, do we still need it?

subhasisb:
> Some more comments.
> I wanted to ask, now that you are updating flags to have MODIFY wherever MODCACHE was set. What is the distinction left between these two? Can we get rid of one?

They are used in two different things. Yet, they must both be set for the attribute whenever the attribute changes. Both are to flag that the attribute has changed (dirty) but they have different needs in terms of who uses them. The MODIFY flag is used by the DB, MODCACHE is used the stat cache. If we used only one flag, say MODCACHE, the flag will get turned off as soon as a stat happens, and we will never write the change to DB. And vice versa.

src/server/req_manager.c
suresh-thelkar: Thanks for the detailed explanation and adding the comment in the code.

src/lib/Libdb/db_postgres_common.c
```diff
@@ -100,12 +102,14 @@ pg_set_error(pbs_db_conn_t *conn, char *fnc, char *msg)
...
 	if (!conn->conn_db_err)
 		return;
 
-	sprintf((char *) conn->conn_db_err, fmt, fnc, msg, str);
+	if (!msg2)
```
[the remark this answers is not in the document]
subhasisb: Eeeks, sure!

src/lib/Libdb/db_postgres_node.c
```diff
@@ -250,42 +266,59 @@ int
...
+			free(raw_array);
+			return -1;
+		}
+		if (fnums_inited == 0) {
+			nd_svtime_fnum = PQfnumber(conn->conn_resultset, "nd_savetm");
+			fnums_inited = 1;
+		}
+		GET_PARAM_STR(conn->conn_resultset, 0, pnd->nd_savetm, nd_svtime_fnum);
+		PQclear(conn->conn_resultset);
 		free(raw_array);
```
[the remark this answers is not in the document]
subhasisb: sure

src/lib/Libdb/db_postgres_resv.c
```diff
@@ -252,49 +283,66 @@ int
...
+			free(raw_array);
+			return -1;
+		}
+		if (fnums_inited == 0) {
+			ri_savetm_fnum = PQfnumber(conn->conn_resultset, "ri_savetm");
+			fnums_inited = 1;
+		}
+		GET_PARAM_STR(conn->conn_resultset, 0, presv->ri_savetm, ri_savetm_fnum);
+		PQclear(conn->conn_resultset);
 		free(raw_array);
```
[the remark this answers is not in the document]
subhasisb: Yes made the change to all the places as well

src/server/job_recov_db.c
```diff
@@ -114,88 +99,82 @@ void print_backtrace(char *);
...
+	if (pjob->ji_qs.ji_state == JOB_STATE_FINISHED)
+		save_all_attrs = 1;
+
+	if ((encode_attr_db(job_attr_def, pjob->ji_wattr, JOB_ATR_LAST, &dbjob->cache_attr_list, &dbjob->db_attr_list, save_all_attrs)) != 0)
+		return -1;
+
+	if (pjob->ji_savetm[0] == '\0') /* object was never saved/loaded before */
+		savetype |= (OBJ_SAVE_NEW | OBJ_SAVE_QS);
+
+	if (obj_qs_modified(&pjob->ji_qs, sizeof(pjob->ji_qs), pjob->qs_hash) == 1) {
```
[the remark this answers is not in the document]
subhasisb: pbs_db.h line 887

src/server/job_recov_db.c
```diff
@@ -351,133 +247,56 @@ db_to_svr_resv(resc_resv *presv, pbs_db_resv_info_t *pdresv)
  *
  * @param[in]	pjob - The job to save
  * @param[in]   updatetype:
- *		SAVEJOB_QUICK - Quick update, save only quick save area
- *		SAVEJOB_FULL  - Update along with attributes
- *		SAVEJOB_NEW   - Create new job in database (insert)
- *		SAVEJOB_FULLFORCE - Same as SAVEJOB_FULL
- *
+ *				- bitwise operator to speficy what part to save. OBJ_SAVE_QS, OBJ_SAVE_ATTRS
```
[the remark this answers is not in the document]
subhasisb: done

src/server/job_recov_db.c
```diff
@@ -351,133 +247,56 @@ db_to_svr_resv(resc_resv *presv, pbs_db_resv_info_t *pdresv)
...
-		 * The update has five parts:
-		 * (1) the job structure,
-		 * (2) the extended area,
-		 * (3) if a Array Job, the index tracking table
-		 * (4) the attributes in the "encoded "external form, and last
-		 * (5) the dependency list.
-		 */
+		/* don't save mtime, set it from ji_savetm - TODO */
+		pjob->ji_wattr[JOB_ATR_mtime].at_val.at_long = time_now;
+		pjob->ji_wattr[JOB_ATR_mtime].at_flags |= ATR_VFLAG_MODCACHE;
```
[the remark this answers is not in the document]
subhasisb: This is existing code. I think JOB_ATR_mtime is always set, so already has the ATR_VFLAG_SET set on it. Anyways, it won't harm to set it, so adding ATR_VFLAG_SET to it.

src/server/job_recov_db.c
```diff
@@ -351,133 +247,56 @@ db_to_svr_resv(resc_resv *presv, pbs_db_resv_info_t *pdresv)
...
-			if (updatetype == SAVEJOB_NEW && strstr(conn->conn_db_err, "duplicate key value")) {
-				/* new job has a jobid clash, allow retry with a new jobid */
-				pbs_db_reset_obj(&obj);
-				if (pbs_db_end_trx(conn, PBS_DB_COMMIT) != 0)
-					goto db_err;
+	if (rc != 0) {
+		sprintf(log_buffer, "Failed to save job %s ", pjob->ji_qs.ji_jobid);
+		if (conn->conn_db_err != NULL)
+			strncat(log_buffer, conn->conn_db_err, LOG_BUF_SIZE - strlen(log_buffer) - 1);
+		log_err(-1, __func__, log_buffer);
```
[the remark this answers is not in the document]
subhasisb: Yeah can do that, updated

src/server/node_recov_db.c
subhasisb: Yeah, added - could help if ATTR_rescavail/ncpus is at the early part of the list, else it does not. Ideally, we should eliminate this loop - but...

src/server/node_recov_db.c
subhasisb: done

src/server/queue_recov_db.c
subhasisb: done

src/server/queue_recov_db.c
subhasisb: done

src/server/req_quejob.c
subhasisb: I will leave it in a function, lest tomorrow we decide to add some more logic to it, like walk over and try to find a free slot first

subhasisb:
@arungrover  further changes pushed, please check

subhasisb:
Second round perf tests done:
[perf.tests2.txt](https://github.com/openpbs/openpbs/files/4792361/perf.tests2.txt)

Please do not merge, still need to do a valgrind run

subhasisb:
marked WIP to do the valgrind tests before unblocking for merging

arungrover:
> > Some more comments.
> > I wanted to ask, now that you are updating flags to have MODIFY wherever MODCACHE was set. What is the distinction left between these two? Can we get rid of one?
> 
> They are used in two different things. Yet, they must both be set for the attribute whenever the attribute changes. Both are to flag that the attribute has changed (dirty) but they have different needs in terms of who uses them. The MODIFY flag is used by the DB, MODCACHE is used the stat cache. If we used only one flag, say MODCACHE, the flag will get turned off as soon as a stat happens, and we will never write the change to DB. And vice versa.

It sounds like if we have a DB MODIFY flag on, it will always result into MODCACHE as well. Would it help if we made MODIFY flag as an or'ed value of 'ATR_VFLAG_MODIFY | ATR_VFLAG_MODCACHE'? This way whoever modifies the code in the future does not have to care about setting both? I can understand you have already made changes in multiple places and if it is a lot of work then don't bother.

bhroam:
I'd +1 a OR'd flag.  I am currently in the process of adding one for SET|MODIFY|MODCACHE since they are commonly set together.  The combination flag saves on line length.

src/server/node_recov_db.c
```diff
@@ -88,152 +83,189 @@
...
-				&& ((pnode->nd_attr[ND_ATR_Sharing].at_val.at_long != VNS_UNSET)
-						&& (pnode->nd_attr[ND_ATR_Sharing].at_val.at_long != VNS_DFLT_SHARED)));
-	numattr = vnode_sharing;
-	psvrl = (svrattrl *)GET_NEXT(wrtattr);
-	while (psvrl) {
-		if ((strcmp(psvrl->al_name, ATTR_rescavail) == 0) &&
-				(strcmp(psvrl->al_resc, "ncpus") == 0)) {
+	/* MSTODO: how can we optimize this loop - eliminate this? */
+	psvrl = (svrattrl *) GET_NEXT(pdbnd->db_attr_list.attrs);
+	while (psvrl != NULL) {
```
arungrover, line 240: How about we create a new flag and mark the attributes that are not supposed to be encoded to DB? Then encode_attr_db can ignore such attributes and this loop can be avoided

subhasisb:
Since there were a few changes, uploading the latest regression report:
Profile: regression
|--|--|--|--|--|--|--|
|2262|1815|0|0|0|0|1815|

subhasisb:
> I'd +1 a OR'd flag. I am currently in the process of adding one for SET|MODIFY|MODCACHE since they are commonly set together. The combination flag saves on line length.

Ok, yes makes sense, i will add a combined flag of the 3 and modify everywhere

subhasisb:
We have changed our design quite a bit. Will close this and raise a fresh PR soon. Thanks for all the reviews

PR #1824, opened 2020-06-11
Fix server indefinite hang

hirenvadalia:
#### Describe Bug or Feature

* Due to changes in openpbs/openpbs#1505, if server receives request from incompatible client then it will hang indefinitely to receive data from incompatible client, while client is not sending any more data

#### Describe Your Change

* Add check for fd ready before calling blocking read() in tcp_recv(), if incompatible client is not sending any more data then server will max wait for tcp timeout, then it will reject connection/request

#### Link to Design Doc

* None

* qmove 50k + sched off

|Before libauth|After fix|
|-|-|
|real	36m18.909s<br>user	1m21.727s<br>sys	2m49.302s|real	25m1.785s<br>user	1m20.452s<br>sys	2m50.000s|

* qsub 100k + sched off

|19|After fix|
|-|-|
|real	7m36.121s<br>user	1m1.326s<br>sys	4m10.856s|*Try 1:*<br>real	7m34.954s<br>user	1m0.912s<br>sys	4m11.010s<br><br>*Try 2:*<br>real	7m35.154s<br>user	1m1.473s<br>sys	4m11.201s|

* qdel 100k + sched off

|19|After fix|
|-|-|
|real	16m49.136s<br>user	1m48.411s<br>sys	11m6.905s<br>|*Try 1:*<br>real	16m4.863s<br>user	1m56.191s<br>sys	11m5.135s<br><br>*Try 2:*<br>real	16m36.950s<br>user	1m50.640s<br>sys	11m40.516s|

Screen Shot 2020-06-11 at 1 42 10 PM
(NOTE: 5 failures are known issue in one of the test suite)

### DIFF
--- a/src/lib/Libdis/dis_helpers.c
+++ b/src/lib/Libdis/dis_helpers.c
@@ -375,9 +375,8 @@ parse_pkt(void *pkt, size_t pkt_len, int *type, void **data_out, size_t *len_out
 		*data_out = NULL;
 		*len_out = 0;
 		return -1;
-	} else {
+	} else
 		pos += PKT_MAGIC_SZ;
-	}
 	*type = *((unsigned char *)pos);
 	pos++;
 	*len_out = ntohl(*((int *)pos));
@@ -458,9 +457,8 @@ transport_send_pkt(int fd, int type, void *data_in, size_t len_in)
 		}
 		free(data_out);
 	} else {
-		if (create_pkt(type, data_in, len_in, &pkt, &pktlen) != 0) {
+		if (create_pkt(type, data_in, len_in, &pkt, &pktlen) != 0)
 			return -1;
-		}
 	}
 
 	i = transport_send(fd, pkt, pktlen);
@@ -520,9 +518,8 @@ transport_recv_pkt(int fd, int *type, void **data_out, size_t *len_out)
 			*/
 			transport_chan_set_old_client(fd);
 			data_in = malloc(i);
-			if (data_in == NULL) {
+			if (data_in == NULL)
 				return -1;
-			}
 			memcpy(data_in, (void *)pkt_magic, i);
 			*data_out = data_in;
 			*len_out = i;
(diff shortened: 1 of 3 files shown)


src/lib/Libifl/tcp_dis.c
```diff
@@ -109,46 +109,35 @@ tcp_recv(int fd, void *data, int len)
...
+		 * deliver promptly
+		 */
+		do {
+			i = select(fd + 1, &readset, NULL, NULL, &timeout);
+			if (pbs_tcp_interrupt)
+				break;
+		}
 #ifdef WIN32
-	while (i == -1 && errno == WSAEINTR);
+		while (i == -1 && errno == WSAEINTR);
```
subhasisb, line 133: I think this is not ever set errno - since on windows we need to call WSAGetLastError() to get the error code. Besides that, WASEINTR seems to happen only if when a cancel API is called (which we never call).

From MSDN:
WSAEINTR 10004 | Interrupted function call.  A blocking operation was interrupted by a call to WSACancelBlockingCall.
-- | --

subhasisb:
@hirenvadalia can we also do qsub/qdel/migration performance test for 100k jobs before we merge this?

src/lib/Libdis/dis_helpers.c
```diff
@@ -435,11 +502,36 @@ transport_recv_pkt(int fd, int *type, void **data_out, size_t *len_out)
...
+	if (strncmp(pkt_magic, PKT_MAGIC, PKT_MAGIC_SZ) != 0) {
+		if (pkt_magic[0] == DIS_HDR_FSTCHR) {
+			/*
+			* Mark this fd as old client as it doesn't have
+			* pkt magic and first received char in data is
+			* DIS header's first char (aka DIS_HRD_FSTCHR)
+			*/
+			transport_chan_set_old_client(fd);
+			data_in = malloc(i);
+			if (data_in == NULL) {
```
ashwathraop, line 523: Upto you if you want to change and remove the braces.

PR #1827, opened 2020-06-11
Fix: Python is unable to initialize in windows jobs

raviv1995:
#### Describe Bug or Feature
Python jobs fail to start the interpreter when submitted with a shell (qsub -S) with an error: Fatal Python error: failed to get random numbers to initialize Python

#### Describe Your Change
This error occurs when the environment does not contain the SystemRoot variable. The environ variable was NULLing out because of a buffer overflow.

[test_logs.txt](https://github.com/openpbs/openpbs/files/4764666/test_logs.txt)

### DIFF
--- a/src/resmom_win/win/mom_start.c
+++ b/src/resmom_win/win/mom_start.c
@@ -96,7 +96,7 @@ set_shell(job *pjob, struct passwd *pwdp)
 {
 	char	*cp;
 	int	i;
-	static char	shell[MAX_PATH + 1] = {'\0'};
+	static char	shell[MAX_PATH + 1];
 	char    *temp_dir = NULL;
 	struct array_strings *vstrs;
 
@@ -121,6 +121,7 @@ set_shell(job *pjob, struct passwd *pwdp)
 			}
 		}
 	}
+	shell[sizeof(shell) - 1] = '\0';
 	for (i=0; shell[i] != '\0'; i++) {
 		if (shell[i] == '/')
 			shell[i] = '\\';
--- a/src/resmom_win/win/mom_start.c
+++ b/src/resmom_win/win/mom_start.c
@@ -104,8 +104,10 @@ set_shell(job *pjob, struct passwd *pwdp)
 	 * Find which shell to use, one specified or the login shell
 	 * If we fail to get cmd shell(unlikely), use "cmd.exe" as shell
 	 */
-	if (0 != get_cmd_shell(shell, sizeof(shell)))
+	if (0 != get_cmd_shell(shell, sizeof(shell))) {
 		strncpy(shell, "cmd.exe", sizeof(shell) - 1);
+		shell[sizeof(shell) - 1] = '\0';
+	}
 	if ((pjob->ji_wattr[(int)JOB_ATR_shell].at_flags & ATR_VFLAG_SET) &&
 		(vstrs = pjob->ji_wattr[(int)JOB_ATR_shell].at_val.at_arst)) {
 		for (i = 0; i < vstrs->as_usedptr; ++i) {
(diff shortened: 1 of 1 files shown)


src/resmom_win/win/mom_start.c
```diff
@@ -105,19 +105,19 @@ set_shell(job *pjob, struct passwd *pwdp)
 	 * If we fail to get cmd shell(unlikely), use "cmd.exe" as shell
 	 */
 	if (0 != get_cmd_shell(shell, sizeof(shell)))
-		(void)snprintf(shell, sizeof(shell) - 1, "cmd.exe");
+		strncpy(shell, "cmd.exe", sizeof(shell) - 1);
```
Shrini-h, line 108: I think we might be leaving a chance of having non null turnminated shell string. Please handle this case for all the 3 strncpy changes in this PR

raviv1995: We initialize the entire array to null. Since we are copying till sizeof(shell)-1 it will ensure the last element of the array will be null

Shrini-h: are you sure about that. may be this one is safe as "cmd.exe" is smaller than MAXPATH len but the I'm afraid about the below two strncpy

src/resmom_win/win/mom_start.c
```diff
@@ -105,19 +105,19 @@ set_shell(job *pjob, struct passwd *pwdp)
...
 	if ((pjob->ji_wattr[(int)JOB_ATR_shell].at_flags & ATR_VFLAG_SET) &&
 		(vstrs = pjob->ji_wattr[(int)JOB_ATR_shell].at_val.at_arst)) {
 		for (i = 0; i < vstrs->as_usedptr; ++i) {
 			cp = strchr(vstrs->as_string[i], '@');
 			if (cp) {
 				if (!strncmp(mom_host, cp+1, strlen(cp+1))) {
 					*cp = '\0';	/* host name matches */
-					strncpy(shell, vstrs->as_string[i], PBS_CMDLINE_LENGTH - 1);
+					strncpy(shell, vstrs->as_string[i], sizeof(shell) - 1);
```
Shrini-h, line 116: Null terminate shell or use snprintf !

src/resmom_win/win/mom_start.c
```diff
@@ -105,19 +105,19 @@ set_shell(job *pjob, struct passwd *pwdp)
...
 					break;
 				}
 			} else {
-				strncpy(shell, vstrs->as_string[i], PBS_CMDLINE_LENGTH - 1);	/* wildcard */
+				strncpy(shell, vstrs->as_string[i], sizeof(shell) - 1);	/* wildcard */
```
Shrini-h, line 120: Null terminate shell or use snprintf !

src/resmom_win/win/mom_start.c
Shrini-h: Can you point me where we are initializing the whole buffer ?

raviv1995: Will not matter, we are only copying MAX_PATH - 1 in strncpy. Also MAX_PATH is a windows standard.

raviv1995: We initialize the whole buffer on line 97.

raviv1995: We initialize the whole buffer on line 99.

Shrini-h: Since the buffer is static my concern was that it only gets initialized the first time only. But 
Yes I think its a false alarm since the "-1" saves the last NULL byte from being overwritten. Now I agree that shell will always be NULL terminated. Thanks

src/resmom_win/win/mom_start.c
Shrini-h: Agree shell is terminated

src/resmom_win/win/mom_start.c
Shrini-h: Agree shell is terminated

src/resmom_win/win/mom_start.c
subhasisb: Better to not depend on the static variable to never be used again by the next programmer down the line (and overwrite your NULL character at some point). Instead of initializing it at declaration time, just add a line after this:
shell[sizeof(shell) - 1] = '\0' 

It is way faster than setting NULLs on every byte (which you dont need) and fool proof, if somebody overwrites the last NULL

Shrini-h: I can see that `get_cmd_shell()` is clearing out the entire shell, but I agree that we should have a explicit guard that @subhasisb suggested, for the sake of assurance while reading the code.

PR #1828, opened 2020-06-11
Fixed reconfirming multinode reservations

bhroam:
#### Describe Bug or Feature
The scheduler would fail to reconfirm a multi-node reservation

#### Describe Your Change
In order to reconfirm a running reservation, the scheduler has to remove the currently allocated resources the reservation is using.  This is so when we go to find nodes for the reservation, there are resources free on those nodes.  The scheduler was not removing the resources on the nodes, so the reservation basically conflicted with itself and failed to reconfirm.

The function that removes the resources from a reservation only worked for reservations in state RESV_BEING_ALTERED.  I modified it to also with for reservations in state RESV_DEGRADED.

[resv.log](https://github.com/openpbs/openpbs/files/4767732/resv.log)

### DIFF
--- a/src/scheduler/resv_info.c
+++ b/src/scheduler/resv_info.c
@@ -1945,7 +1945,7 @@ release_running_resv_nodes(resource_resv *resv, node_info **all_nodes)
 	if (resv == NULL || all_nodes == NULL )
 		return;
 	degraded = resv->resv->resv_state == RESV_RUNNING && resv->resv->resv_substate == RESV_DEGRADED;
-	being_altered = resv->resv->resv_state == RESV_RUNNING && resv->resv->resv_substate == RESV_DEGRADED;
+	being_altered = resv->resv->resv_state == RESV_BEING_ALTERED && resv->resv->resv_substate == RESV_RUNNING;
 	if (degraded || being_altered) {
 		resv_nodes = resv->ninfo_arr;
 		for (i = 0; resv_nodes[i] != NULL; i++) {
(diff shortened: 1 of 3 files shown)


src/scheduler/resv_info.c
```diff
@@ -1924,33 +1924,33 @@ end_resv_on_nodes(resource_resv *resv, node_info **all_nodes)
...
 		return;
-	if (resv->resv->resv_state == RESV_BEING_ALTERED) {
-		if (resv->resv->resv_substate == RESV_RUNNING) {
-			resv_nodes = resv->ninfo_arr;
-			for (i = 0; resv_nodes[i] != NULL; i++) {
-				ninfo = find_node_by_indrank(all_nodes, resv_nodes[i]->node_ind, resv_nodes[i]->rank);
-				update_node_on_end(ninfo, resv, NULL);
-			}
+	degraded = resv->resv->resv_state == RESV_RUNNING && resv->resv->resv_substate == RESV_DEGRADED;
+	being_altered = resv->resv->resv_state == RESV_RUNNING && resv->resv->resv_substate == RESV_DEGRADED;
```
arungrover, line 1948: This isn't being altered, it is the same as degraded.

bhroam:
[alter.log](https://github.com/openpbs/openpbs/files/4773191/alter.log)
I decided to run the pbs_ralter test plan.  I originally only ran the reservation test plan.  The ralter one passes now.

PR #1830, opened 2020-06-12
Fix migration upgrade and server indefinite hang

hirenvadalia:
#### Describe Bug or Feature

* Due to changes in #1505, if server receives request from incompatible client then it will hang indefinitely to receive data from incompatible client, while client is not sending any more data
* And due to above hang migration upgrade fails during moving jobs from old server to new server using qmove

#### Describe Your Change

* Add check for fd ready before calling blocking read() in tcp_recv(), if incompatible client is not sending any more data then tcp_recv() will max wait for tcp timeout before returning error
* To allow ONLY qmove from old installation (aka old client), added checks for pkt magic and dis header's first char (aka `+`) and using this two check we detect that data is coming from old client in non-pkt format, so transport layer will fallback to send/recv data for that connection in old non-pkt format.

#### Link to Design Doc

* None

* migration upgrade from 19 to after fix: [logs.txt](https://github.com/openpbs/openpbs/files/4769075/logs.txt)

* qmove 50k + sched off

|Before libauth|After fix|
|-|-|
|real	36m18.909s<br>user	1m21.727s<br>sys	2m49.302s|real	25m1.785s<br>user	1m20.452s<br>sys	2m50.000s|

* qsub 100k + sched off

|19|After fix|
|-|-|
|real	7m36.121s<br>user	1m1.326s<br>sys	4m10.856s|*Try 1:*<br>real	7m34.954s<br>user	1m0.912s<br>sys	4m11.010s<br><br>*Try 2:*<br>real	7m35.154s<br>user	1m1.473s<br>sys	4m11.201s|

* qdel 100k + sched off

|19|After fix|
|-|-|
|real	16m49.136s<br>user	1m48.411s<br>sys	11m6.905s<br>|*Try 1:*<br>real	16m4.863s<br>user	1m56.191s<br>sys	11m5.135s<br><br>*Try 2:*<br>real	16m36.950s<br>user	1m50.640s<br>sys	11m40.516s|

* Regression results coming soon...

### DIFF
--- a/src/include/dis.h
+++ b/src/include/dis.h
@@ -225,6 +225,7 @@ typedef struct pbs_tcp_auth_data {
 typedef struct pbs_tcp_chan {
 	pbs_dis_buf_t readbuf;
 	pbs_dis_buf_t writebuf;
+	int is_old_client; /* This is just for backward compatibility */
 	pbs_tcp_auth_data_t auths[2];
 } pbs_tcp_chan_t;
 
--- a/src/lib/Libdis/dis_helpers.c
+++ b/src/lib/Libdis/dis_helpers.c
@@ -49,12 +49,69 @@
 #include "pbs_internal.h"
 #include "auth.h"
 
+#define PKT_MAGIC "PKTV1"
+#define PKT_MAGIC_SZ sizeof(PKT_MAGIC)
+#define DIS_HDR_FSTCHR '+'
+
 static pbs_dis_buf_t * dis_get_readbuf(int);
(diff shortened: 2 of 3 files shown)


hirenvadalia:
> For migration upgrade support, you need to add the old authentication code also, right?

No @subhasisb, when qmove sends req to old server, old server uses privileged port as root (aka client_to_svr_extended()) to connect to new server, means no auth req is being send, old server directly sends queuejob request to new server as soon as it connects.

PR #1831, opened 2020-06-12
qsub hangs for 60 secs when job is submitted from python script on windows

RKORavi:
#### Describe Bug or Feature
When a job is submitted from python script on windows it hangs for 60 secs.
Reason: 

- qsub command is executed using subprocess.Popen, followed by proc. communicate() to get error, out of command execution.

- Since Background qsub process is attached to foreground qsub process, proc.communicate() waits for background qsub process to complete which is 60 secs.

#### Describe Your Change
1. To create background qsub process in CreateProcess function change creation flag attribute from CREATE_NO_WINDOW to CREATE_NEW_PROCESS_GROUP.

CREATE_NO_WINDOW : The process is a console application that is being run without a console window. Therefore, the console handle for the application is not set
CREATE_NEW_PROCESS_GROUP: The new process is the root process of a new process group. The process group includes all processes that are descendants of this root process. The process identifier of the new process group is the same as the process identifier, which is returned in the lpProcessInformation parameter.

Microsoft doc link: https://docs.microsoft.com/en-us/windows/win32/procthread/process-creation-flags

2. Close stdin, stdout, stderr of background qsub.

[test logs.txt](https://github.com/openpbs/openpbs/files/4768965/test.logs.txt)
[qsub_tracejob_logs.txt](https://github.com/openpbs/openpbs/files/4768967/qsub_tracejob_logs.txt)

### DIFF
--- a/src/cmds/qsub.c
+++ b/src/cmds/qsub.c
@@ -5094,7 +5094,7 @@ again:
 					qsub_exe, fl,
 					h_event, server_out);
 			if (!CreateProcess(NULL, cmd_line, &sa, &sa,
-						TRUE, CREATE_NO_WINDOW, NULL,
+						TRUE, CREATE_NEW_PROCESS_GROUP, NULL,
 						NULL, &si, &pi)) {
 				CloseHandle(h_event);
 				return rc;
@@ -5681,6 +5681,9 @@ main(int argc, char **argv, char **envp) /* qsub */
 	if ((argc == 4 || argc == 5) && (strcasecmp(argv[1], "--daemon") == 0)) {
 		/* set when background qsub is running */
 		is_background = 1;
+		(void)fclose(stdin);
+		(void)fclose(stdout);
+		(void)fclose(stderr);
 		if (argc == 4)
 			do_daemon_stuff(argv[2], argv[3], NULL);
 		else
--- a/src/cmds/qsub.c
+++ b/src/cmds/qsub.c
@@ -5043,6 +5043,7 @@ daemon_submit(char *qsub_exe, int *do_regular_submit)
 	char cmd_line[2 * MAXPATHLEN + 1];
 	int created = 0;
 	HANDLE h_event;
+	int flags = CREATE_DEFAULT_ERROR_MODE | CREATE_NEW_PROCESS_GROUP;
 
 	/* determine pipe name */
 	get_comm_filename(fl);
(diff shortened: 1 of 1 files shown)


src/cmds/qsub.c
```diff
@@ -5094,7 +5094,7 @@ daemon_submit(char *qsub_exe, int *do_regular_submit)
 					qsub_exe, fl,
 					h_event, server_out);
 			if (!CreateProcess(NULL, cmd_line, &sa, &sa,
-						TRUE, CREATE_NO_WINDOW, NULL,
+						TRUE, CREATE_NEW_PROCESS_GROUP, NULL,
```
hirenvadalia, line 5097: I think you still need CREATE_NO_WINDOW also, else it will pop up window

RKORavi: Hi Hiren,

I tried CREATE_NEW_PROCESS_GROUP | CREATE_NO_WINDOW as creation flag but it results in python script hanging for 60 secs. 
I too agree that i should add CREATE_NO_WINDOW flag, but combining it with CREATE_NEW_PROCESS_GROUP does not seem to work.

I even tried CREATE_BREAKAWAY_FROM_JOB |  CREATE_NO_WINDOW but even this does not work.

hirenvadalia: try `CREATE_DEFAULT_ERROR_MODE|CREATE_NEW_PROCESS_GROUP` and set `si.lpDesktop = NULL` and make sure you are not getting any pop up windows

RKORavi: Hi Hiren, 

Using these flag combination worked for me. But could you please explain little more on si.lpDesktop = NULL. 
I added this before createprocess(..) after referring to other parts of code. Is that the right place?

hirenvadalia: yes, `si.lpDesktop = NULL` is needed only before createprocess and that `si` is not used anywhere else, so don't worry.

RKORavi: Thankyou for review & comment. I will rerun all the tests and will update my results.

RKORavi:
PS C:\Users\pbsadmin> ps | findstr qsub
Time taken to run python script.
PS C:\Users\pbsadmin> time { python.exe .\test_qsub.py }
Command return {'out': ['8090.serverpbs'], 'err': [], 'rc': 0}


Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 0
Milliseconds      : 325
Ticks             : 3252384
TotalDays         : 3.76433333333333E-06
TotalHours        : 9.0344E-05
TotalMinutes      : 0.00542064
TotalSeconds      : 0.3252384
TotalMilliseconds : 325.2384


Python script:
PS C:\Users\pbsadmin> type .\test_qsub.py
import subprocess
import os
cmd = ['C:\\Users\\pbsadmin\\PtlPbshvihpa1w.bat']
#cmd = ['qsub', '-q@serverpbs', '--', 'pbs-sleep.exe', '10']

ret={'out':[],'err':[],'rc':0}
proc = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
(o,e)=proc.communicate()
ret['rc'] = proc.returncode
if o is not None:
        ret['out'] = [i.decode("utf-8") for i in o.splitlines()]
else:
        ret['out'] = []
if e is not None:
        ret['err'] = [i.decode("utf-8") for i in e.splitlines()]
else:
        ret['err'] = []
print("Command return %s" %ret)


PS C:\Users\pbsadmin> type .\PtlPbshvihpa1w.bat
@echo off
"C:\Program Files (x86)\PBS\exec"\bin\qsub -q@serverpbs -- pbs-sleep 10

PR #1098, opened 2019-04-19
PP-774: Ability to create file and directory as a user in PTL needs enhancement

sandisamp:
#### Describe Bug or Feature

 Currently, mkdtemp is not working as how it should work when user is trying to create a directory as another user. It creates the directory and do a chown to the specified user. However, it may fail as only root can run chown. We also need to make this interface consistent with create_temp_file().

#### Describe Your Change

the new implementation of the given function where it can be fed the argument “asuser” to create the directory as the given user rather than trying to modify it at a later stage.

PR #1833, opened 2020-06-13
[WIP] create_temp_dir does not honor permissions and group

arungrover:
#### Describe Bug or Feature
create_temp_dir creates a temporary directory but does not honor the permission bits given as input.

#### Describe Your Change
Change in #1098 changed the way a temporary directory used to be created. But this change missed checking for mode.

#### Link to Design Doc
NA

[test.txt](https://github.com/openpbs/openpbs/files/4773746/test.txt)

I'm going to run regression. PR is marked as draft.

### DIFF
--- a/test/fw/ptl/utils/pbs_dshutils.py
+++ b/test/fw/ptl/utils/pbs_dshutils.py
@@ -1540,6 +1540,8 @@ class DshUtils(object):
                            runas=runas, level=level)
         if ret['rc'] == 0:
             if gid is not None:
+                if runas is None:
+                    runas = _u
                 rv = self.chgrp(hostname, path, gid=gid, sudo=sudo,
                                 level=level, recursive=recursive, runas=runas,
                                 logerr=logerr)
@@ -2072,8 +2074,7 @@ class DshUtils(object):
             self.run_copy(hostname, src=tmpdir, dest=dirname, runas=asuser,
                           recursive=True, gid=gid, uid=uid,
                           level=level)
-            if mode is not None:
-                self.chmod(hostname, path=dirname, mode=mode, runas=asuser)
+            self.chmod(hostname, path=dirname, mode=mode, runas=asuser)
 
             tmpdir = dirname + tmpdir[4:]
 
@@ -2085,8 +2086,7 @@ class DshUtils(object):
                 self.run_copy(hostname, src=tmpdir, dest=tmpdir, runas=asuser,
                               recursive=True, uid=uid, gid=gid,
                               level=level)
-                if mode is not None:
-                    self.chmod(hostname, path=tmpdir, mode=mode, runas=asuser)
+                self.chmod(hostname, path=tmpdir, mode=mode, runas=asuser)
             else:
                 # copy temp dir created on localhost to remote as current user
                 self.run_copy(hostname, src=tmpdir, dest=tmpdir,
@@ -2102,8 +2102,7 @@ class DshUtils(object):
             # copy the orginal temp as new temp dir
             self.run_copy(hostname, src=tmpdir, dest=tmpdir2, runas=asuser,
                           recursive=True, uid=uid, gid=gid, level=level)
-            if mode is not None:
-                self.chmod(hostname, path=tmpdir2, mode=mode, runas=asuser)
+            self.chmod(hostname, path=tmpdir2, mode=mode, runas=asuser)
             # remove original temp dir
             os.rmdir(tmpdir)
             self.tmpdirlist.append(tmpdir2)
--- a/test/fw/ptl/utils/pbs_dshutils.py
+++ b/test/fw/ptl/utils/pbs_dshutils.py
@@ -2032,8 +2032,8 @@ class DshUtils(object):
         return tmpfile
 
     def create_temp_dir(self, hostname=None, suffix='', prefix='PtlPbs',
-                        dirname=None, asuser=None, asgroup=None, mode=None,
-                        level=logging.INFOCLI2, sudo=False):
+                        dirname=None, asuser=None, asgroup=None, mode=0o755,
+                        level=logging.INFOCLI2):
         """
         Create a temp dir by calling ``tempfile.mkdtemp``
         :param hostname: the hostname on which to query tempdir from
@@ -2063,26 +2063,23 @@ class DshUtils(object):
             gid = grp.getgrnam(str(asgroup))[2]
         # create a temp dir as current user
         tmpdir = tempfile.mkdtemp(suffix, prefix)
+        # By default mkdtemp creates dir according to usmask.
+        # To create dir as different user first change the dir
+        # permission to 0755 so that other user has read permission
+        self.chmod(path=tmpdir, mode=0o755)
         if dirname is not None:
             dirname = str(dirname)
-            self.chmod(path=tmpdir, mode=0o755)
             self.run_copy(hostname, src=tmpdir, dest=dirname, runas=asuser,
                           recursive=True, gid=gid, uid=uid,
                           level=level)
             if mode is not None:
                 self.chmod(hostname, path=dirname, mode=mode, runas=asuser)
-            else:
-                self.chmod(hostname, path=dirname, mode=0o755, runas=asuser)
 
             tmpdir = dirname + tmpdir[4:]
 
         # if temp dir to be created on remote host
         if not self.is_localhost(hostname):
             if asuser is not None:
-                # by default mkstemp creates dir with 0o700 permission
-                # to create dir as different user first change the dir
-                # permission to 0o755 so that other user has read permission
-                self.chmod(path=tmpdir, mode=0o755)
                 # copy temp dir created on local host to remote host
                 # as different user
                 self.run_copy(hostname, src=tmpdir, dest=tmpdir, runas=asuser,
@@ -2090,10 +2087,7 @@ class DshUtils(object):
                               level=level)
                 if mode is not None:
                     self.chmod(hostname, path=tmpdir, mode=mode, runas=asuser)
-                else:
-                    self.chmod(hostname, path=tmpdir, mode=0o755, runas=asuser)
             else:
-                self.chmod(path=tmpdir, mode=0o755)
                 # copy temp dir created on localhost to remote as current user
                 self.run_copy(hostname, src=tmpdir, dest=tmpdir,
                               level=level, preserve_permission=True,
@@ -2101,10 +2095,6 @@ class DshUtils(object):
             # remove local temp dir
             os.rmdir(tmpdir)
         if asuser is not None:
-            # by default mkdtemp creates dir with 0o700 permission
-            # to create dir as different user first change the dir
-            # permission to 0o755 so that other user has read permission
-            self.chmod(path=tmpdir, mode=0o755)
             # since we need to create as differnt user than current user
             # create a temp dir just to get temp dir name with absolute path
             tmpdir2 = tempfile.mkdtemp(suffix, prefix, dirname)
@@ -2114,8 +2104,6 @@ class DshUtils(object):
                           recursive=True, uid=uid, gid=gid, level=level)
             if mode is not None:
                 self.chmod(hostname, path=tmpdir2, mode=mode, runas=asuser)
-            else:
-                self.chmod(hostname, path=tmpdir2, mode=0o755, runas=asuser)
             # remove original temp dir
             os.rmdir(tmpdir)
             self.tmpdirlist.append(tmpdir2)
--- a/test/fw/ptl/utils/pbs_dshutils.py
+++ b/test/fw/ptl/utils/pbs_dshutils.py
@@ -2042,10 +2042,10 @@ class DshUtils(object):
         :type suffix: str
         :param prefix: the directory name will begin with this prefix
         :type prefix: str
-        :param dir: the directory will be created in this directory
-        :type dir: str or None
-        :param uid: Optional username or uid of temp directory owner
-        :param gid: Optional group name or gid of temp directory
+        :param dirname: the directory will be created in this directory
+        :type dirname: str or None
+        :param asuser: Optional username of temp directory owner
+        :param asgroup: Optional group name of temp directory
                     group owner
         :param mode: Optional mode bits to assign to the temporary
                      directory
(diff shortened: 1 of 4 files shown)


arungrover:
Thanks for reviewing Bhroam.
This change caused a specific test in direct_write test suite to fail. I have modified the test. Please have a look.

[test.txt](https://github.com/openpbs/openpbs/files/4789743/test.txt)

test/tests/functional/pbs_qsub_direct_write.py
```diff
@@ -88,11 +88,12 @@ def test_direct_write_when_job_succeeds_controlled(self):
...
         3) not accessible via other permissions
         """
-        j = Job(TEST_USER4, attrs={ATTR_k: 'doe'})
+        j = Job(TEST_USER2, attrs={ATTR_k: 'doe'})
         j.set_sleep_time(10)
         sub_dir = self.du.create_temp_dir(asuser=TEST_USER5)
         mapping_dir = self.du.create_temp_dir(
-            asuser=TEST_USER4, asgroup=TSTGRP5, mode=0o770)
+            asuser=TEST_USER2, asgroup=TSTGRP5)
+        self.du.chmod(path=mapping_dir, mode=0o770, runas=TEST_USER2)
```
nithinj, line 96: I'm trying to understand the change.

why the user has changed from TEST_USER4 to TEST_USER2?
also, why to use chmod when create_temp_dir accepts mode?

arungrover: I made the change to username because in the test at line 104/105 the test tries to access the temporary directory to list the files. Now unless the user who is running the test and the user who is running the job is part of a common group, it will fail.
pbsroot is part of tstgrp2 and so is user2 and that is why this change.

About chmod, I think you are right. I'm trying another change to ensure that create_temp_dir handles that change.

arungrover:
@bhroam and @nithinj I have made some more changes in the last two commits.
While addressing Nithin's comment about doing chmod in the dshutil itself I noticed that the function also does not honor the group name passed to it.
After making the change to the function to support groups, the direct write test started failing because pbsroot (user who runs the test) is not part of tsgrp02. I had to change that to tstgrp0 because that's the only common group between pbsuser2 and pbsroot.
I've also added a selftest for this call.
[test.txt](https://github.com/openpbs/openpbs/files/4801567/test.txt)


Please have a look again.

test/fw/ptl/utils/pbs_dshutils.py
```diff
@@ -2070,61 +2070,84 @@ def create_temp_dir(self, hostname=None, suffix='', prefix='PtlPbs',
...
         :param prefix: the directory name will begin with this prefix
         :type prefix: str
-        :param dir: the directory will be created in this directory
-        :type dir: str or None
-        :param uid: Optional username or uid of temp directory owner
-        :param gid: Optional group name or gid of temp directory
+        :param dirname: the directory will be created in this directory
+        :type dirname: str or None
+        :param asuser: Optional username of temp directory owner
+        :param asgroup: Optional group name of temp directory
```
bhroam, line 2076: Type these two parameters

test/fw/ptl/utils/pbs_dshutils.py
```diff
@@ -2070,61 +2070,84 @@ def create_temp_dir(self, hostname=None, suffix='', prefix='PtlPbs',
...
 
         # if temp dir to be created on remote host
         if not self.is_localhost(hostname):
             if asuser is not None:
-                # by default mkstemp creates dir with 0600 permission
+                # by default mkstemp creates dir with 0700 permission
                 # to create dir as different user first change the dir
                 # permission to 0755 so that other user has read permission
-                self.chmod(path=tmpdir, mode=mode)
+                self.chmod(path=tmpdir, mode=0o755)
```
bhroam, line 2110: The comment says 0700, but the code says 0o755

test/fw/ptl/utils/pbs_dshutils.py
```diff
@@ -2070,61 +2070,84 @@ def create_temp_dir(self, hostname=None, suffix='', prefix='PtlPbs',
...
                 self.run_copy(hostname, src=tmpdir, dest=tmpdir, runas=asuser,
-                              recursive=True,
+                              recursive=True, uid=uid, gid=gid,
                               preserve_permission=preserve, level=level)
+                if mode is not None:
+                    self.chmod(hostname, path=tmpdir, mode=mode, runas=asuser)
+                else:
+                    self.chmod(hostname, path=tmpdir, mode=0o755, runas=asuser)
             else:
+                self.chmod(path=tmpdir, mode=0o755)
```
bhroam, line 2121: Do you need hostname in here?

test/fw/ptl/utils/pbs_dshutils.py
```diff
@@ -2070,61 +2070,84 @@ def create_temp_dir(self, hostname=None, suffix='', prefix='PtlPbs',
...
+                              uid=uid, gid=gid)
             # remove local temp dir
             os.rmdir(tmpdir)
         if asuser is not None:
-            # by default mkdtemp creates dir with 0600 permission
+            # by default mkdtemp creates dir with 0700 permission
             # to create dir as different user first change the dir
             # permission to 0755 so that other user has read permission
-            self.chmod(path=tmpdir, mode=mode)
+            self.chmod(path=tmpdir, mode=0o755)
```
bhroam, line 2132: Same here with 0700 and 0o755

test/tests/selftest/pbs_dshutils_tests.py
```diff
@@ -86,3 +89,38 @@ def test_pbs_env(self):
...
+        self.check_access(tmpdir, user=getpass.getuser())
+
+        # create a directory as a specific user
+        tmpdir = self.du.create_temp_dir(asuser=TEST_USER2)
+        self.check_access(tmpdir, user=TEST_USER2, mode=0o755)
+
+        # create a directory as a specific user and permissions
+        tmpdir = self.du.create_temp_dir(asuser=TEST_USER2, mode=0o765)
+        self.check_access(tmpdir, mode=0o765, user=TEST_USER2)
+
```
bhroam, line 122: You're creating a directory, I wouldn't choose a mask without the execute bit.  Maybe 750?

test/fw/ptl/utils/pbs_dshutils.py
arungrover: Right, what happens is that mkdtemp creates the directory (line 2090), and that directory is created as 700 but we need to change that to ensure the directory can be copied as a different user. This is the reason it is 755.
I will update the comment to show octal numbers.

test/fw/ptl/utils/pbs_dshutils.py
arungrover: Good catch!

arungrover: on second thought, no we don't have to use hostname. Reason is that we are changing permissions of the local directory before it gets copied. So it is doing the right thing.

arungrover:
[test.txt](https://github.com/openpbs/openpbs/files/4801895/test.txt)

test/fw/ptl/utils/pbs_dshutils.py
```diff
@@ -2070,55 +2070,80 @@ def create_temp_dir(self, hostname=None, suffix='', prefix='PtlPbs',
...
             os.rmdir(tmpdir2)
             # copy the orginal temp as new temp dir
             self.run_copy(hostname, src=tmpdir, dest=tmpdir2, runas=asuser,
-                          recursive=True,
-                          preserve_permission=False, level=level)
+                          recursive=True, uid=uid, gid=gid, level=level)
+            if mode is not None:
+                self.chmod(hostname, path=tmpdir2, mode=mode, runas=asuser)
+            else:
+                self.chmod(hostname, path=tmpdir2, mode=0o755, runas=asuser)
```
bhroam, line 2146: Couldn't you avoid this if statement if you defaulted mode to 0o755?

arungrover: You are right @bhroam.
I'm thinking to just adjust the comment and just not check for 0o700 at all. The reason being, the default permissions would depend on umask value on the system. So assuming its going to be 0o700 may not be right.
I'll also remove these else statements because if the user has not given the mode, let them get what the system provides.
What do you think?

bhroam: I'm not sure.  While using the umask of the system is a very unixy thing it do, it could cause a test to fail on one system and not another due to the different umask.  If we hammer it to the default of 0o755, it will be the same on all systems.  I think I'd agree with you if this wasn't a testing environment.

arungrover: I agree, its probably better to have defined behavior for tests.

test/fw/ptl/utils/pbs_dshutils.py
```diff
@@ -2070,55 +2070,80 @@ def create_temp_dir(self, hostname=None, suffix='', prefix='PtlPbs',
...
                 self.run_copy(hostname, src=tmpdir, dest=tmpdir, runas=asuser,
-                              recursive=True,
-                              preserve_permission=False, level=level)
+                              recursive=True, uid=uid, gid=gid,
+                              level=level)
+                if mode is not None:
+                    self.chmod(hostname, path=tmpdir, mode=mode, runas=asuser)
+                else:
+                    self.chmod(hostname, path=tmpdir, mode=0o755, runas=asuser)
             else:
```
nithinj, line 2123: looks like we are repeating the code which we have inside if. Why don't we place the check just before the last chmod?
if chmod is a NOOP when asuser is None, this check may not be even required.

All three conditions in the function do a somewhat similar sequence of operations. Can it be combined to form a generic sequence without explicit conditions?

arungrover: Right, While that can be done but it will affect readability a lot. I'll have to set src and destination early up in the function to then call common lines of the code and then again based on the three conditions cleanup unnecessary directories.
I am anyways making a change to reduce the number of chmod calls, that should help make it smaller.

test/fw/ptl/utils/pbs_dshutils.py
```diff
@@ -2142,8 +2132,6 @@ def create_temp_dir(self, hostname=None, suffix='', prefix='PtlPbs',
                           recursive=True, uid=uid, gid=gid, level=level)
             if mode is not None:
                 self.chmod(hostname, path=tmpdir2, mode=mode, runas=asuser)
-            else:
```
bhroam: mode can't really be None now, unless of course someone passes in None.  Also, why do all the chmoding to 0o755 instead of mode?  It defaults to 0o755, so you should only really have to do the chmod once.

arungrover: Right, someone can pass mode as None that is why I left it as is. 
The reason I'm explicitly using chmod 755 instead of using mode is that if someone gives a very restrictive mode then the function might not even be able to do a run_copy. This is the reason I create a directory explicitly as 755 and then do a copy and then chmod it based on what was given.

bhroam: I wouldn't bother checking for None.  The type of this is an octal integer,  If someone passes None, it shouldn't work and crapping out on chmod() is fine.

test/fw/ptl/utils/pbs_dshutils.py
```diff
@@ -2071,55 +2073,65 @@ def create_temp_dir(self, hostname=None, suffix='', prefix='PtlPbs',
...
+        :type asuser: str
+        :param asgroup: Optional group name of temp directory
                     group owner
+        :type asgroup: str
         :param mode: Optional mode bits to assign to the temporary
                      directory
+        :type mode: octal integer
         :param level: logging level, defaults to INFOCLI2
         """
+        current_user_info = self.get_id_info(self.get_current_user())
```
hirenvadalia, line 2088: Optional (But I would highly recommend doing): You can make use of PbsUser.get_user

test/fw/ptl/utils/pbs_dshutils.py
```diff
@@ -2071,55 +2073,65 @@ def create_temp_dir(self, hostname=None, suffix='', prefix='PtlPbs',
...
+        uid = current_user_info['uid']
+        gid = current_user_info['gid']
+        if asuser is not None:
+            uid = self.get_id_info(str(asuser))['uid']
```
hirenvadalia, line 2092: Same as above

test/fw/ptl/utils/pbs_dshutils.py
```diff
@@ -2071,55 +2073,65 @@ def create_temp_dir(self, hostname=None, suffix='', prefix='PtlPbs',
...
+        if asgroup is not None:
+            gid = grp.getgrnam(str(asgroup))[2]
```
hirenvadalia, line 2094: Same as above, currently PbsGroup doesn't have get_group method, but you can introduce it, it will be useful in future also to someone

test/fw/ptl/utils/pbs_dshutils.py
```diff
@@ -2071,55 +2073,65 @@ def create_temp_dir(self, hostname=None, suffix='', prefix='PtlPbs',
...
         # create a temp dir as current user
         tmpdir = tempfile.mkdtemp(suffix, prefix)
+        # By default mkdtemp creates dir according to usmask.
```
hirenvadalia, line 2097: usmask == umask

arungrover: :) I guess people not wearing masks is getting into my head

test/fw/ptl/utils/pbs_dshutils.py
arungrover: I can do so but I think it will have a limitation that the user who is running PTL will then have to be one of the users PTL creates. Currently, I can run tests as any system user and it works.

hirenvadalia: I don't think PTL fw/utils or any tests go beyond what we have in `PBS_ALL_USERS`, and if we really need to go beyond `PBS_ALL_USERS` then we should enhance `get_user` to allow it.

arungrover: Well, like in this case, If I run a test as "arung" it will fail because current_user (arung) is not in the list of PBS_ALL_USERS. Is there anything that says a non-ptl user cannot run a PTL test? 
Making this change will force to run the test as one of the PBS_ALL_USERS.

arungrover:
Marking it WIP, making more changes to dshutils for remote hosts and adding another selftest

arungrover:
I could not get clean build because of recent changes. Rebased it.

test/fw/ptl/utils/pbs_dshutils.py
hirenvadalia: Sorry missed you comment somehow: No we don't have any such guidelines which say you can't run PTL as use which is not part of PBS_ALL_USERS. But usually people write tests using users which is listed on PBS_ALL_USERS so there are high chances of failures if you run PTL as a user which not listed in PBS_ALL_USERS.

arungrover:
Rebasing it again to get clean build

arungrover:
I've not been able to test it on our internal test machines. closing the PR for now.

PR #1837, opened 2020-06-15
qmove gives wrong err msg when remote host is unknown

arungrover:
#### Describe Bug or Feature
Change made in #1814 resulted into PBS server rejecting a move job request when it is not able to resolve the remote host address. But we did not set the right error number while returning from svr_movejob which means the error qmove command gets is ambiguous. 

#### Describe Your Change
This change sets the pbs_errno to PBSE_BADHOST.

#### Link to Design Doc
NA

[test_fix.txt](https://github.com/openpbs/openpbs/files/4782513/test_fix.txt)

### DIFF
--- a/test/tests/functional/pbs_moved_job.py
+++ b/test/tests/functional/pbs_moved_job.py
@@ -128,5 +128,6 @@ class TestMovedJob(TestFunctional):
         jid = self.server.submit(j)
 
         err = "Access from host not allowed, or unknown host " + jid
-        with self.assertRaises(PbsMoveError, msg=err):
+        with self.assertRaises(PbsMoveError) as e:
             self.server.movejob(jid, 'workq@unknownserver')
+        self.assertIn(err, e.exception.msg[0])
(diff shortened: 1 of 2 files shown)


test/tests/functional/pbs_moved_job.py
```diff
@@ -114,3 +114,19 @@ def test_moved_job_history(self):
...
+        """
+
+        attr = {'scheduling': 'false'}
+        self.server.manager(MGR_CMD_SET, SERVER, attr)
+
+        j = Job(TEST_USER)
+        jid = self.server.submit(j)
+
+        err = "Access from host not allowed, or unknown host " + jid
+        with self.assertRaises(PbsMoveError, msg=err):
```
bhroam, line 131: msg is the message printed in the PTL log when this exception is not raised.  I think you want to print something like 'qmove not rejected'.  You want to say 'with self.assertRaises() as e:' and then after after your with statement, you can say self.assertEqueals(e.exception.msg, err)

arungrover: ahh... my bad :)
I'll correct it :)

arungrover:
[test_fix.txt](https://github.com/openpbs/openpbs/files/4782744/test_fix.txt)

arungrover:
@bhroam I couldn't find any test in this repo that needs changing

PR #1838, opened 2020-06-16
test_epilogue_single_node fails on altix because the cgroups hook runs after the test's hook

minghui-liu:
#### Describe Bug or Feature

Test_singleNode_Job_ResourceUsed.test_epilogue_single_node fails on cpuset systems because those systems have the cgroups hook enabled by default.
test_spilogue_single_node fails because it creates an exechost_epilogue hook that sets a bunch of resources_used values and expects to find those values in the qstat of the job.
However because the cgroups hook (order=100) runs after the test's hook, the cgroups hook can overwrite values that the test expects to find.

#### Describe Your Change

The epilogue hook in this test was given a higher order number than the cgroups hook so that the test hook run last. 
#### Link to Design Doc

N/A

[pbi-24-s12-altix.txt](https://github.com/openpbs/openpbs/files/4784691/pbi-24-s12-altix.txt)
[pbi-23-r7p1-altix.txt](https://github.com/openpbs/openpbs/files/4784693/pbi-23-r7p1-altix.txt)

### DIFF
--- a/test/tests/functional/pbs_resc_used_single_node.py
+++ b/test/tests/functional/pbs_resc_used_single_node.py
@@ -147,7 +147,7 @@ e.job.resources_used["stra"] = '"glad,elated","happy"'
 """
 
         hook_name = "epi"
-        a = {'event': "execjob_epilogue", 'enabled': 'True'}
+        a = {'event': "execjob_epilogue", 'enabled': 'True', 'order': '200'}
         rv = self.server.create_import_hook(
             hook_name,
             a,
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_resc_used_single_node.py
```diff
@@ -147,7 +147,7 @@ def test_epilogue_single_node(self):
 """
 
         hook_name = "epi"
-        a = {'event': "execjob_epilogue", 'enabled': 'True'}
+        a = {'event': "execjob_epilogue", 'enabled': 'True', 'order': '200'}
```
bhroam, line 150: Add a comment explaining why you need order=200.  If the cgroups order is ever changed, and this test starts failing again, it will be helpful to know what to do.

PR #1842, opened 2020-06-17
Enhance pbs_ralter for -lselect for non-running reservations

bhroam:
#### Describe Bug or Feature
Enhance pbs_ralter to reduce the size of a select of a reservation for non-running reservations.

#### Describe Your Change
Most of this change is in the server to allow the modify reservation code path to accept -lselect and make sure the select is a subset of original select.  I also did a little refactoring where the ralter variables are stored.

#### Link to Design Doc
**[Design](https://pbspro.atlassian.net/wiki/spaces/PD/pages/1664090170/Enhancing+pbs+ralter+for+-lselect)**

[ralter.log](https://github.com/openpbs/openpbs/files/4795268/ralter.log)


src/server/req_modify.c
```diff
@@ -848,90 +911,124 @@ req_modifyReservation(struct batch_request *preq)
...
+			case RESV_ATR_resource:
+				if (strcmp(psatl->al_resc, "select") == 0) {
+					if (presv->ri_wattr[RESV_ATR_state].at_val.at_long != RESV_CONFIRMED && presv->ri_wattr[RESV_ATR_state].at_val.at_long != RESV_DEGRADED) {
+						resv_revert_alter(presv);
+						req_reject(PBSE_BADSTATE, 0, preq);
+						return;
+					}
+					send_to_scheduler = 1;
+					presv->ri_alter.ra_flags |= RESV_SELECT_MODIFIED;
+					prdef = find_resc_def(svr_resc_def, "select", svr_resc_size);
```
arungrover, line 939: I think you can use global pseldef here.

src/server/req_modify.c
```diff
@@ -848,90 +911,124 @@ req_modifyReservation(struct batch_request *preq)
...
+					presc = find_resc_entry(&presv->ri_wattr[RESV_ATR_resource], prdef);
+					if (presc == NULL) {
+						req_reject(PBSE_INTERNAL, 0, preq);
+						resv_revert_alter(presv);
+						return;
+					}
+					if (is_standing && presv->ri_alter.ra_revert.rr_select == NULL)
```
arungrover, line 946: don't we want to safe keep the select for advance reservations too?

src/server/req_modify.c
```diff
@@ -848,90 +911,124 @@ req_modifyReservation(struct batch_request *preq)
...
-			resv_revert_alter_times(presv);
-			return;
+		if (presv->ri_alter.ra_flags & (RESV_START_TIME_MODIFIED | RESV_END_TIME_MODIFIED | RESV_DURATION_MODIFIED)) {
+			/* "start", "end", "duration", and "wall"; derive and check */
+			if (start_end_dur_wall(presv)) {
+				req_reject(PBSE_BADTSPEC, 0, preq);
+				resv_revert_alter(presv);
+				return;
+			}
+			/* walltime can changed */
```
arungrover, line 1028: Please correct the tense

src/server/svr_jobfunc.c
```diff
@@ -3029,11 +3029,31 @@ Time4occurrenceFinish(resc_resv *presv)
 	/* If the start time of the reservation was altered, copy from RESV_ATR_start
  	 * will make the next instance to have it's start time altered so take the start
  	 * time from the ri_alter_stime. */
-	if (presv->ri_alter_stime) {
-		dtstart = presv->ri_alter_stime;
-		presv->ri_alter_stime = 0;
+	if (presv->ri_alter.ra_revert.rr_stime) {
```
arungrover, line 3032: I think it is better to check for flags than the value. What do you think?

src/server/svr_jobfunc.c
```diff
@@ -3029,11 +3029,31 @@ Time4occurrenceFinish(resc_resv *presv)
...
+		dtstart = presv->ri_alter.ra_revert.rr_stime;
+		presv->ri_alter.ra_revert.rr_stime = 0;
```
arungrover, line 3034: I don't understand this part. I assume that when this function is called, the reservation is in BEING_ALTERED state and that is why we want to register the next reservation task with the original stime and select rather than using unconfirmed parameters.
I'm confused why are we resetting the original stime and select to NULL? won't we need them in case the confirmation fails? Also, if resetting the value is necessary, don't we also have to reset the flags?

src/server/svr_jobfunc.c
bhroam: There are two parts of the ra_alter structure.  There are elements to keep track of the current values, and a structure to keep track of the original values.  The current values are used to revert a reservation back when the alter fails.  The ra_revert is used for reverting subsequent occurrences of a standing reservation.  When we alter a standing reservation, we only alter the current occurrence.  We need to keep track of the original values because we might alter the same occurrence multiple times.

src/server/svr_jobfunc.c
bhroam: This is for a standing reservation occurrence that has been modified and is currently ending.  The state will be running.  We need to revert the start time, duration, and select back to the original values.  Those values are then used to move to the next occurrence.  We set the values to NULL because once reset back to the original values, we don't need them any more.  When we move onto the 3rd occurrence, they will already be the correct values.

src/server/req_modify.c
```diff
@@ -848,90 +911,124 @@ req_modifyReservation(struct batch_request *preq)
...
 
 				break;
 			case RESV_ATR_duration:
-				send_to_scheduler = RESV_DURATION_MODIFIED;
-				presv->ri_alter_flags |= RESV_DURATION_MODIFIED;
+				send_to_scheduler = 1;
+				presv->ri_alter.ra_flags |= RESV_DURATION_MODIFIED;
+				break;
+			case RESV_ATR_resource:
+				if (strcmp(psatl->al_resc, "select") == 0) {
```
vstumpf, line 931: if this strcmp fails, it means that someone submitted something like `pbs_rsub -lplace=scatter`, right?  Shouldn't this be an error?
Also, if you can flip the if statement, that'd be great.

src/server/req_modify.c
```diff
@@ -848,90 +911,124 @@ req_modifyReservation(struct batch_request *preq)
...
+					if (presv->ri_wattr[RESV_ATR_state].at_val.at_long != RESV_CONFIRMED && presv->ri_wattr[RESV_ATR_state].at_val.at_long != RESV_DEGRADED) {
```
vstumpf, line 932: just for readability's sake, can you make this `if (!(a or b))`, personally i think it's cleaner
it would read if not confirmed or degraded, then do this

src/server/req_modify.c
bhroam: I'll use pseldef in req_modifyReservation().  It is static to req_modify.c, so I'll still have to search for it Time4occurrenceFinish()

src/server/req_modify.c
```diff
@@ -848,90 +911,124 @@ req_modifyReservation(struct batch_request *preq)
...
 
+	if (presv->ri_alter.ra_flags & RESV_SELECT_MODIFIED) {
+		presc = find_resc_entry(&presv->ri_wattr[RESV_ATR_resource], prdef);
+		make_schedselect(&presv->ri_wattr[(int) RESV_ATR_resource], presc, NULL, &presv->ri_wattr[(int) RESV_ATR_SchedSelect]);
+		if (is_select_smaller(presv->ri_wattr[RESV_ATR_SchedSelect_orig].at_val.at_str, presv->ri_wattr[RESV_ATR_SchedSelect].at_val.at_str) == 0) {
+			req_reject(PBSE_SELECT_NOT_SUBSET, 0, preq);
+			resv_revert_alter(presv);
+			return;
+		}
+	}
```
vstumpf, line 1017: You should do this check before we set all the attributes of the ralter, to be consistent with the other options.

src/include/reservation.h
```diff
@@ -151,6 +153,23 @@ typedef struct pbsnode_list_ {
...
+};
+
+struct resv_alter {
+	time_t ra_stime;
+	time_t ra_etime;
+	long ra_duration;
+	char *ra_select;
+	long ra_state;
+	unsigned long ra_flags;
+	struct resv_alter_revert ra_revert;
```
vstumpf, line 170: you added a lot of values, more than what was originally there, can you add some comments to explain the difference between `resv_alter.ra_stime` and `resv_alter_revert.rr_stime`

bhroam: The only new value is ra_revert.  There was a bug in existing code that if you modify a standing reservation multiple times, it only reverts back to the most recent change (i.e., A->B->C, it'd revert to B, not A).  My options were to fix the bug or exacerbate it by making it broken with select.  I chose to fix it.  I'll add the comment.

src/server/req_modify.c
bhroam: The call to is_select_smaller() is made on schedselect to schedselect.  I need to make the schedselect first before I can compare it.

src/lib/Libattr/master_resv_attr_def.xml
```diff
@@ -801,6 +801,29 @@
...
+      <member_at_comp>comp_str</member_at_comp>
+      <member_at_free>free_str</member_at_free>
+      <member_at_action>NULL_FUNC</member_at_action>
+      <member_at_flags>
+         <both>ATR_DFLAG_MGRD</both>
+      </member_at_flags>
+      <member_at_type>
+         <both>ATR_TYPE_STR</both>
+      </member_at_type>
+      <member_at_parent>PARENT_TYPE_JOB</member_at_parent>
```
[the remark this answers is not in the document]
bhroam: I even noticed that the sched select was also PARENT_TYPE_JOB.  One wonders if this field is ever used.

src/server/req_modify.c
bhroam: We only care about standing reservations since ra_revert is used to revert standing reservations in Time4occurrenceFinish()

src/server/req_modify.c
bhroam: Good catch, pbs_ralter catches it, but if someone used the API, a -lplace=scatter would just have been ignored.

src/server/req_modify.c
bhroam: I don't think it is more readable.  I do agree this statement being so long makes it hard to read.  I set the state in a local variable and used it in the if statement.  I think it is much easier to read now.

bhroam:
@arungrover @vstumpf I've addressed your changes.  I'll post the new test logs when they are finished running.

src/server/req_modify.c
vstumpf: Can't you make the sched select in the switch-case statement?
It seems this is a test we can do before setting all the attributes, instead of comparing the new with original schedselects, you can compare the resv schedselect and a schedselect created from the attribute from the ralter request.

bhroam:
New test logs: 
[ralter.log](https://github.com/openpbs/openpbs/files/4807131/ralter.log)

src/server/req_modify.c
bhroam: I can't.  The schedselect is made from the select, so the select attribute has to be set.  The attributes are only set after the select.

PR #1845, opened 2020-06-22
Update travis to have centos:8

sandisamp:
#### Describe Bug or Feature

Replaced centos:7 with centos:8 in travis

#### Describe Your Change

Added compatibility for centos:8 in travis script

See travis results

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ addons:
 services:
   - docker
 env:
-  - OS_TYPE=centos:7
+  - OS_TYPE=centos:8
   - OS_TYPE=opensuse/leap:15
   - OS_TYPE=ubuntu:18.04
   - OS_TYPE=debian:9
--- a/ci/do.sh
+++ b/ci/do.sh
@@ -49,7 +49,7 @@ if [ -f /src/ci ]; then
   logdir=/logs
   PBS_DIR=/pbssrc
 else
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -46,7 +46,7 @@ addons:
 services:
   - docker
 env:
-  - OS_TYPE=centos:7
+  - OS_TYPE=centos:8
```
hirenvadalia, line 49: don't we need to replace centos:7 in sanitize and Kerberos?

ci/do.sh
```diff
@@ -62,20 +62,47 @@ if [ ! -r ${SPEC_FILE} -o ! -r ${REQ_FILE} ]; then
 fi
 
 if [ "x${IS_CI_BUILD}" != "x1" ] || [ "x${FIRST_TIME_BUILD}" == "x1" -a "x${IS_CI_BUILD}" == "x1" ]; then
-  if [ "x${ID}" == "xcentos" ]; then
+  if [ "x${ID}" == "xcentos" -a "x${VERSION_ID}" == "x7" ]; then
```
hirenvadalia, line 65: How about removing `"x${ID}" == "xcentos"` and put a check which works for RHEL and CentOS? as of now, this will work only for centos.

Same for suse check also?

kjakkali:
> 
> 
> Shouldn't we also update the sanitize and kerberos images to use centos8?

@agrawalravi90  :  Yes. we will updating sanitize and kerberos  builds ASAP.

PR #1854, opened 2020-06-23
Fixing a mem leak in sched

agrawalravi90:
#### Describe Bug or Feature

#1689  introduced the following memory leak in sched:

==56924== 12 bytes in 1 blocks are definitely lost in loss record 30 of 3,513
==56924==    at 0x4C30EDB: malloc (vg_replace_malloc.c:309)
==56924==    by 0x434F0F: string_dup (in /opt/pbs/sbin/pbs_sched)
==56924==    by 0x422406: ??? (in /opt/pbs/sbin/pbs_sched)
==56924==    by 0x42398B: set_validate_sched_attrs (in /opt/pbs/sbin/pbs_sched)
==56924==    by 0x4236D6: update_svr_schedobj (in /opt/pbs/sbin/pbs_sched)
==56924==    by 0x41D8C6: main (in /opt/pbs/sbin/pbs_sched)
==56924== 
==56924== 36 bytes in 3 blocks are definitely lost in loss record 307 of 3,513
==56924==    at 0x4C30EDB: malloc (vg_replace_malloc.c:309)
==56924==    by 0x434F0F: string_dup (in /opt/pbs/sbin/pbs_sched)
==56924==    by 0x422406: ??? (in /opt/pbs/sbin/pbs_sched)
==56924==    by 0x42398B: set_validate_sched_attrs (in /opt/pbs/sbin/pbs_sched)
==56924==    by 0x41EA38: schedule (in /opt/pbs/sbin/pbs_sched)
==56924==    by 0x41D936: main (in /opt/pbs/sbin/pbs_sched)

This just fixes that.

### DIFF
--- a/src/scheduler/fifo.c
+++ b/src/scheduler/fifo.c
@@ -2450,6 +2450,16 @@ next_job(status *policy, server_info *sinfo, int flag)
 static void
 init_sc_attrs(void)
 {
+	static int first_time = 1;
+
+	if (!first_time) {
+		free(sc_attrs.comment);
+		free(sc_attrs.job_sort_formula);
+		free(sc_attrs.partition);
+		free(sc_attrs.sched_log);
+		free(sc_attrs.sched_port);
+		free(sc_attrs.sched_priv);
+	}
 	sc_attrs.attr_update_period = 0;
 	sc_attrs.comment = NULL;
 	sc_attrs.do_not_span_psets = 0;
@@ -2469,6 +2479,8 @@ init_sc_attrs(void)
 	sc_attrs.server_dyn_res_alarm = 0;
 	sc_attrs.throughput_mode = 1;
 	sc_attrs.opt_backfill_fuzzy = BF_DEFAULT;
+
+	first_time = 0;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


src/scheduler/fifo.c
```diff
@@ -2450,6 +2450,16 @@ next_job(status *policy, server_info *sinfo, int flag)
...
+	static int first_time = 1;
+
+	if (!first_time) {
+		free(sc_attrs.comment);
+		free(sc_attrs.job_sort_formula);
+		free(sc_attrs.partition);
+		free(sc_attrs.sched_log);
+		free(sc_attrs.sched_port);
+		free(sc_attrs.sched_priv);
+	}
```
bhroam, line 2462: If the above elements default to NULL, you don't need first_time.  Doing a free(NULL) is as noop.

agrawalravi90: No, this is the only place where I initialize them, I figured that I had to do it every-time so I put it in a function instead of a file.

bhroam: It looks like sc_attrs is a global structure (not mallo()'d).  This means it is zeroed when it is declared.  You should be safe just freeing these.

agrawalravi90: ok, i was a bit unsure if that was reliable or not. I'll remove it then, thanks.

PR #1855, opened 2020-06-23
test_hook_interrupt fails due to not finding the log message

lisa-altair:
#### Describe Bug or Feature

test_hook_interrupt searches for "TestHook Started" in the servers log messages, but only searches 5 times before it gives up.  On slow machines or when other hook event may be happening, this may not be enough time.

#### Describe Your Change

Removed the max_attempts.  This way the default max_attempts will be used.

This is a sporadic timing issue, and I could not reproduce the problem.
[hook_interrupt_after.txt](https://github.com/openpbs/openpbs/files/4822353/hook_interrupt_after.txt)

### DIFF
--- a/test/tests/functional/pbs_hook_set_interrupt.py
+++ b/test/tests/functional/pbs_hook_set_interrupt.py
@@ -68,14 +68,14 @@ pbs.logmsg(pbs.LOG_DEBUG, "TestHook Ended")
         st = time.time()
         jid = self.server.submit(j)
 
-        self.server.log_match("TestHook Started", starttime=st, max_attempts=5)
+        self.server.log_match("TestHook Started", starttime=st)
 
         _msg = "Not Running: PBS Error: request rejected"
         _msg += " as filter hook '%s' got an alarm call." % hook_name
         _msg += " Please inform Admin"
         self.server.expect(JOB,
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_hook_set_interrupt.py
```diff
@@ -68,14 +68,14 @@ def test_hook_interrupt(self):
...
-        self.server.log_match("TestHook Started", starttime=st, max_attempts=5)
+        self.server.log_match("TestHook Started", starttime=st)
 
         _msg = "Not Running: PBS Error: request rejected"
         _msg += " as filter hook '%s' got an alarm call." % hook_name
         _msg += " Please inform Admin"
         self.server.expect(JOB,
                            {'job_state': 'Q', 'comment': _msg},
-                           id=jid, offset=5, max_attempts=5)
+                           id=jid, offset=5)
```
bhroam, line 78: Can you combine the above 2 lines without going over 79 characters?

lisa-altair: No, line 77 is already at 63.  It would go over 79 char by a few characters.

bhroam: Thanks for checking

PR #1857, opened 2020-06-24
Modify TestPbsNodeRampDown tests to work on slow machines

shilpakodli:
#### Describe Bug or Feature
Few of the TestPbsNodeRampDown tests were failing on slow machines with error "no data for job".
In these tests, Job is creating a 1Gb file to lengthen stageout process but on slow machines this would create only few 100Mbs file and when Job is deleted in test, stageout process would be done in 1 second due to which test fails in further Log matches. 

#### Describe Your Change
Increased File size which Job is creating from 1gb to 2gb also changed Fibnoacci series to create 400 series instead of 50 in Job script.
Added a sleep of 3 seconds before we delete a job which triggers stageout process in tests to create a large sized file.

TH3_results

### DIFF
--- a/test/tests/functional/pbs_node_rampdown.py
+++ b/test/tests/functional/pbs_node_rampdown.py
@@ -37,7 +37,6 @@
 # "OpenPBS®", "PBS Professional®", and "PBS Pro™" and Altair's logos is
 # subject to Altair's trademark licensing policies.
 
-
 from tests.functional import *
 
 
@@ -389,10 +388,10 @@ return i\\n return fib(i-1) + fib(i-2)\\n\\nprint(fib(400))\\\")"'
             "#PBS -l place=" + self.job1_place + "\n" + \
             "#PBS -W stageout=test.img@%s:test.img\n" % (self.n4,) + \
             "#PBS -W release_nodes_on_stageout=true\n" + \
-            "dd if=/dev/zero of=test.img count=1024 bs=1048576\n" + \
+            "dd if=/dev/zero of=test.img count=1024 bs=2097152\n" + \
             "pbsdsh -n 1 -- %s\n" % (FIB40,) + \
             "pbsdsh -n 2 -- %s\n" % (FIB40,) + \
-            "%s\n" % (FIB50,)
+            "%s\n" % (FIB400,)
 
         self.script['job1_1'] = \
             "#PBS -S /bin/bash\n" \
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_node_rampdown.py
```diff
@@ -580,6 +579,10 @@ def tearDown(self):
         self.momB.signal("-CONT")
         self.momC.signal("-CONT")
         TestFunctional.tearDown(self)
+        for host in [self.hostA, self.hostB, self.hostC]:
+            test_img = os.path.join("/home", "pbsuser", "test.img")
+            self.du.rm(hostname=host, path=test_img, force=True, runas=TEST_USER)
+        TestFunctional.tearDown(self)
```
kjakkali, line 585: remove TestFunctional.tearDown from line 581

test/tests/functional/pbs_node_rampdown.py
```diff
@@ -630,23 +634,9 @@ def test_release_nodes_on_stageout_true(self):
         # Deleting the job will trigger the stageout process
         # at which time sister nodes are automatically released
         # due to release_nodes_stageout=true set
+        time.sleep(3)
```
kjakkali, line 637: instead of 3 sec sleep can we check test.img file size? once it 2 gb we can do qdel.

shilpakodli: Hi @kjakkali We don't know the exact time for creating a 2 gb file. It might take more time to create on slow machines which leads increase in test execution time..

bayucan: @shilpakodli : I think what @kjakkali is suggesting is that instead of doing the sleep 3, just test for existence of  2gb size of test.img file on the stagein host. Perhaps retry it some set number of times, and if unable to see that size, then fail the test...

test/tests/functional/pbs_node_rampdown.py
```diff
@@ -156,10 +156,14 @@ def check_test_img_size(self):
         in 10 seconds
         """
         test_file_path = os.path.join("/home", str(TEST_USER), "test.img")
-        for i in range(0,11):
-            test_img_size = os.stat(test_file_path).st_size
+        cmd = ['stat', '-c', '%s', test_file_path]
```
hirenvadalia, line 159: you can change this method like:
```python
fpath = os.path.join(TEST_USER.home, "test.img")
cmd = ['stat', '-c', '%s', fpath]
count = 10
fsize = 0
while count > 0:
    fsize = 0
    rc = self.du.run_cmd(hosts=self.hostA, cmd=cmd, runas=TEST_USER)
    if rc['rc'] == 0 and len(rc['out']) == 1:
       try:
          fsize = int(rc['out'][0])
      except:
          pass
   # 1073741824 == 1Gb
   if fsize > 1073741824:
      break
   else:
      count -= 1
      time.sleep(1)
if fsize <= 1073741824:
    self.fail("Failed to create 1gb file at %s" % fpath)
```

test/tests/functional/pbs_node_rampdown.py
```diff
@@ -151,6 +150,32 @@ def pbs_nodefile_match_exec_host(self, jid, exec_host,
             return False
         return True
 
+    def check_test_img_size(self):
```
prakashcv13, line 153: rename the function to the effect that it refers stageout file instead of test image

test/tests/functional/pbs_node_rampdown.py
```diff
@@ -151,6 +150,32 @@ def pbs_nodefile_match_exec_host(self, jid, exec_host,
...
+        """
+        This Function will make sure that atleast 1gb of test.img
```
prakashcv13, line 155: the function does not make sure, but only checks

test/tests/functional/pbs_node_rampdown.py
```diff
@@ -151,6 +150,32 @@ def pbs_nodefile_match_exec_host(self, jid, exec_host,
...
+        file is created in 10 seconds
+        """
+        fpath = os.path.join(TEST_USER.home, "test.img")
+        cmd = ['stat', '-c', '%s', fpath]
+        count = 10
+        fsize = 0
+        while count > 0:
```
prakashcv13, line 162: to me using 'for' would be make it better, we would not need count.

PR #1858, opened 2020-06-24
Simplified ci with systemd and multi node support 

sandisamp:
#### Describe Bug or Feature

* Reduced the number of options in ci tool, so it's easier to manage ci. 
(See Readme to get info on the new CLI)
* Added systemd to running containers
* Added support for multi node tests
* As per feedback removed auto building of pbs on newer commits

#### Describe Your Change

* Removed the options 'make, install, test and start' instead ci now has a single option 'run'. This option will start the container if not already running. It will configure pbs if required or configure options have changed. Run make and install. And if any test options are provided it will run pbs_benchpress on the same.
* Added option  'set'. This will be used to set configuration option in ci. and 'run' will always make use of the configured settings.
* Removed the optional --build option as the tool now always checks for dependency installation on starting a container. If someone adds new dependecy to pbs it will install it as well.
* As per feedback removed the option pbs-dirname and pbs-stop. 
* Also removed the auto building of pbs on new commits as per feedback.
* Added systemd to container. (see container-init) 
* Added support for running multi node test cases. Set can be used to specify nodes as well if someone needs to configure ci to run multinode (see Readme on how). Multi node runs according to the tests configuration provided. It reads the requirement decorator for the given tests and reconfigures ci for multinode to accomodate running that test. The OS platform chosen for all nodes is the last running one. One can provide different OS on nodes by configuring it through set option, However the configured nodes must match the criteria for the test case provided.
* Since multi node is achieved by using exec command this update to ci will support running multi node serially. It will first install necessary files on server followed by mom and comm nodes. Will configure them accordingly and in the end run tests on server node. 

[logfile.txt](https://github.com/openpbs/openpbs/files/4824447/logfile.txt)
[build-server_centos_7_1.txt](https://github.com/openpbs/openpbs/files/4824429/build-server_centos_7_1.txt)


ci/ci
```diff
@@ -213,6 +335,100 @@ def check_for_existing_image():
...
+                            "pbs_server"
+                        ]
+                    }
+                },
+                "hostname": "pbs.ci",
+                "user": "root",
+                "privileged": True
+            }
+        },
+        "networks": {
```
sandisamp, line 400: Make node names as FQDN

ci/ci
```diff
@@ -213,6 +335,100 @@ def check_for_existing_image():
...
+                    "./:/src",
+                    "./logs:/logs",
+                    "./etc:/workspace/etc"
+                ],
+                "entrypoint": "/workspace/etc/container-init \"/src/docker-entrypoint\"",
+                "environment": [
+                    "CLUSTER_INFO='SERVER=centos:7'",
+                    "NODE_TYPE='server'",
+                    "LC_ALL=en_US.utf-8",
+                    "LANG=en_US.utf-8"
```
sandisamp, line 386: Remove LC_ALL instead use LANG and LANGUAGE(if required) only

ci/ci
```diff
@@ -213,6 +335,100 @@ def check_for_existing_image():
...
+                "image": "centos:7",
+                "volumes": [
+                    "../:/pbssrc",
+                    "./:/src",
+                    "./logs:/logs",
+                    "./etc:/workspace/etc"
+                ],
+                "entrypoint": "/workspace/etc/container-init \"/src/docker-entrypoint\"",
+                "environment": [
+                    "CLUSTER_INFO='SERVER=centos:7'",
```
sandisamp, line 383: Remove cluster info

ci/ci
```diff
@@ -213,6 +335,100 @@ def check_for_existing_image():
...
+        "services": {
+            "pbs-build": {
+                "image": "centos:7",
+                "volumes": [
+                    "../:/pbssrc",
+                    "./:/src",
+                    "./logs:/logs",
+                    "./etc:/workspace/etc"
+                ],
+                "entrypoint": "/workspace/etc/container-init \"/src/docker-entrypoint\"",
```
sandisamp, line 381: Remove fail-safe argument, not required

ci/ci
```diff
@@ -153,25 +210,76 @@ def get_pkgs(platform):
...
+            node_config = node_config + 'comm=' + node_image + ';'
+    elif no_comm_on_mom_flag == True and no_comm_on_server_flag == False:
+        if max_comms_needed > max_servers_needed:
+            for _ in range(max_comms_needed-max_servers_needed):
+                node_config = node_config + 'comm=' + node_image + ';'
+    else:
+        if max_comms_needed > only_moms:
+            for _ in range(max_comms_needed - only_moms):
+                node_config = node_config + 'comm=' + node_image + ';'
+
```
sandisamp, line 279: Use list instead of string in node_config

ci/ci
```diff
@@ -153,25 +210,76 @@ def get_pkgs(platform):
...
+    max_moms_needed = 1
+    max_comms_needed = 1
+    no_mom_on_server_flag = False
+    no_comm_on_mom_flag = True
+    no_comm_on_server_flag = False
+    try:
+        with open(os.path.join(pbs_dirname, 'ptl_ts_tree.json')) as f:
+            json_data = json.load(f)
+    except:
+        log_error('Could not find ptl tree json file')
```
sandisamp, line 239: fail execution from here if tree was not able to generate

ci/ci
```diff
@@ -234,198 +450,223 @@ def check_prerequisites():
...
+    current_file_services_list.sort()
+    if current_file_services_list != potential_list:
+        build_compose_file(node_config)
+        return True
+    else:
+        return False
+
+
+def initialise_command_loop():
+    command_path = os.path.join(pbs_dirname, '.commandloops')
```
sandisamp, line 510: change to config dir (also def name)

ci/ci
```diff
@@ -234,198 +450,223 @@ def check_prerequisites():
...
-    try:
-        os.chdir(pbs_dirname)
-    except Exception as e:
-        log_error(str(e))
-        log_error("Was correct pbs path provided?")
-        sys.exit(1)
-    build_log_path = os.path.join(pbs_dirname, 'logs', 'build')
-    if args.delete == None and args.start == None and args.local == None:
+    pbs_dirname = find_ci()
+    pbs_dirname = os.path.join(pbs_dirname, 'ci')
```
sandisamp, line 566: change dir name to ci_dirname

ci/ci
```diff
@@ -234,198 +450,223 @@ def check_prerequisites():
...
+                if is_restart_required():
+                    run_cmd(
+                        "docker-compose -f docker-compose.json down --remove-orphans")
+                    check_prerequisites()
+            target_path = os.path.join(command_path, '.configure_opt')
+            if conf_opts['configure'] != read_from_file(target_path):
+                cmd = ' export ONLY_CONFIGURE=1 && /src/do.sh 2>&1 | tee -a /logs/build-$(hostname) '
+                run_docker_cmd(cmd)
+                write_to_file(target_path, conf_opts['configure'])
+            target_path = os.path.join(command_path, '.make')
```
sandisamp, line 631: let make handle pbs compilation , remove .make file

ci/ci
```diff
@@ -234,198 +450,223 @@ def check_prerequisites():
...
+        if args.delete != None:
+            confirm = input(
+                '\033[91mAre you sure you want to delete containers (Y/N)?: \033[0m')
+            if confirm.lower() == 'no' or confirm.lower() == 'n':
+                os._exit(0)
+            elif confirm.lower() == 'yes' or confirm.lower() == 'y':
+                services = get_services_list()
+                if len(services) != 0:
+                    build_compose_file(nodes=get_current_setup())
+                    cmd = '/src/killit.sh backup'
```
sandisamp, line 654: Also take backup of configured information.

ci/do.sh
```diff
@@ -150,21 +150,25 @@ fi
...
-
-_targetdirname=target-${ID}
+hostname=$(hostname)
+_targetdirname=target-${ID}-${hostname}
 if [ "x${ONLY_INSTALL}" != "x1" -a "x${ONLY_REBUILD}" != "x1" -a "x${ONLY_TEST}" != "x1" ]; then
   rm -rf ${_targetdirname}
 fi
 mkdir -p ${_targetdirname}
-if [ "x${ONLY_REBUILD}" != "x1" -a "x${ONLY_INSTALL}" != "x1" -a "x${ONLY_TEST}" != "x1" ]; then
+if [ ! -f ${_targetdirname}/Makefile ]; then
```
sandisamp, line 167: check for configure file instead

ci/do.sh
```diff
@@ -240,13 +276,25 @@ if [ "x${RUN_TESTS}" == "x1" ]; then
...
+  if [ ! -z "$pbsnodes" ]; then
+    set -- $pbsnodes
+    while [ $# -gt 0 ]; do
+      moms="$moms$1:"
+      shift
+    done
+    moms=$(echo "${moms%?}")
+    echo "moms=$moms" >${commandloops}/.params
+    params="--param-file ${commandloops}/.params"
+  fi
```
sandisamp, line 293: Handle params in ci file instead, take care of comms also

ci/etc/configure_node.sh
```diff
@@ -0,0 +1,56 @@
...
+    for tclist in ts['tclist'].values():
+        no_mom_on_server_flag = tclist['requirements']['no_mom_on_server'] or no_mom_on_server_flag
+        no_comm_on_server_flag = tclist['requirements']['no_comm_on_server'] or no_comm_on_server_flag
+        no_comm_on_mom_flag = tclist['requirements']['no_comm_on_mom'] or no_comm_on_mom_flag
+print(str(no_mom_on_server_flag) +';' + str(no_comm_on_server_flag) + ';' + str(no_comm_on_mom_flag))
+")
+	no_mom_on_server=$(echo $node_flags | awk -F';' '{print $1}')
+	no_comm_on_server=$(echo $node_flags | awk -F';' '{print $2}')
+	no_comm_on_mom=$(echo $node_flags | awk -F';' '{print $3}')
+}
```
sandisamp, line 30: Use environment or file instead

ci/etc/container-init
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash -x
+
+if [ $(echo ${CONTAINER_PLATFORM} | grep 'ci_pbs' | grep -v grep | wc -l) -ne 1 ]; then
```
sandisamp, line 3: Remove this check

ci/etc/container-init
```diff
@@ -0,0 +1,29 @@
...
+	/workspace/etc/install-system-packages 2>&1 | tee -a /logs/build-$(hostname)
+fi
+
+# set environment var file
+touch /env-file
+set >/env-file
```
sandisamp, line 9: make it env file

ci/ci
```diff
@@ -80,23 +140,34 @@ def run_cmd(cmd, return_output=False):
...
+        try:
+            log_info(cmd)
+            docker_cmd = shlex.split(cmd)
+            a = subprocess.Popen(docker_cmd)
+            a.communicate()
+            if a.returncode != 0:
+                log_error(
+                    "docker cmd returned with non zero exit code, redirecting you to container terminal")
+                docker_cmd = shlex.split(
+                    "docker-compose -f docker-compose.json exec " + service + " bash -c \'cd /pbssrc && /bin/bash\'")
```
hirenvadalia, line 165: please run pep8 on this file

ci/ci
```diff
@@ -141,37 +233,92 @@ def commit_docker_image():
...
-            subprocess.run([pbs_dirname + '/ci', '--delete'])
-            subprocess.run([pbs_dirname + '/ci', '--start=' + platform])
-        except:
-            log_error('Failed to restart ci.')
-    log_info('The package build logs can be found in logs/pkglogs')
-    run_docker_cmd(
-        'docker-compose exec pbs-build bash -c \' ' + build_cmd + ' | tee /logs/pkglogs \' ')
+def create_ts_tree_json():
+    benchpress_opt = os.path.join(
+        ci_dirname, '.config_dir', '.benchpress_opt')
```
hirenvadalia, line 244: please use macros for `.config_dir`, '.benchpress_opt' and such hard coded strings in whole file

ci/ci
```diff
@@ -182,29 +329,43 @@ def find_ci():
     return os.path.dirname(cur_dir)
 
 
-def tail_file(build_log_path):
+def tail_file():
```
hirenvadalia, line 332: I don't understand this func, why we need to implement tail when we have `docker-compose logs -f <svc name>`?

sandisamp: The docker-compose logs will stop at the line where we start systemd. Since the do.sh script is then started as a service. It will not show up in the logs. However we have a build-<svc> log file which we can tail to show the progress.

hirenvadalia: ok got it. please rename it show_build_log or tail_build_log

ci/ci
```diff
@@ -106,11 +114,12 @@ def log_warning(str_to_print):
...
     docker_container_process = subprocess.run(
-        ["docker-compose", "-f", "docker-compose.json", "ps"], stdout=subprocess.PIPE)
-    docker_container = str(docker_container_process.stdout)
-    for val in docker_container.split('\\n'):
-        if val.find('Up') != -1:
-            services.append(val.split(' ')[0])
+        ["docker-compose", "-f", "docker-compose.json",
+         "ps", "--filter", "status=running", "--services"], stdout=subprocess.PIPE)
+    docker_container = str((docker_container_process.stdout).decode('utf-8'))
+    for val in docker_container.split('\n'):
```
hirenvadalia, line 120: instead of this for loop you can do like `return [x for x in docker_container.splitlines() if len(x) > 0]` also get rid of services variable

ci/ci
```diff
@@ -53,17 +53,25 @@ from argparse import RawTextHelpFormatter
...
 ci_dirname = ''
-
-requirements_template = Template('''num_servers=$num_servers
-num_moms=$num_moms
-num_comms=$num_comms
-no_mom_on_server=$no_mom_on_server
-no_comm_on_server=$no_comm_on_server
-no_comm_on_mom=$no_comm_on_mom
+# MACROS
+config_dir = '.config_dir'
```
hirenvadalia, line 57: I suggest using caps in name for macros

ci/etc/configure_node.sh
```diff
@@ -7,7 +7,8 @@ fi
 if [ "x${NODE_TYPE}" == "xmom" ]; then
 	sed -i "s@PBS_SERVER=$(hostname)@PBS_SERVER=pbs_server@" /etc/pbs.conf
 	sed -i "s@PBS_START_SERVER=1@PBS_START_SERVER=0@" /etc/pbs.conf
-	HOST_NAME=$(hostname -f)
+	HOST_NAME=$(hostname -s)
+	HOST_NAME="${HOST_NAME}.ci_ci"
 	ssh -t root@pbs_server " /opt/pbs/bin/qmgr -c 'c n ${HOST_NAME}'"
```
hirenvadalia, line 12: I think we should create a node with the short name instead fqdn, also just a suggestion make fqdn like `name.awesome.ci` (aka more that one dot and without `_`)

sandisamp: Using the default DNS for the container names, which docker-compose assigns as {folder_name}_{network name}, hence it will be ci_ci. While i was trying the custom domainname, although the nslookup was working for (hostname -s) but not for FQDN name. So i used the default DNS which avoids complications,and ssh works well.

hirenvadalia: we can override it, I am using custom names in my machine like `node1.dev.local`, `node2.dev.local` etc.. and everything works fine.

let's discuss offline, and will find solution for it.

ci/etc/configure_node.sh
```diff
@@ -5,26 +5,38 @@ if [ -f /src/.config_dir/.requirements_decorator ]; then
...
 	HOST_NAME=$(hostname -s)
-	ssh -t root@pbs_server " /opt/pbs/bin/qmgr -c 'c n ${HOST_NAME}'"
+	ssh -t root@pbs-server " /opt/pbs/bin/qmgr -c 'c n ${HOST_NAME}'"
 	if [ "x$no_comm_on_mom" == "xTrue" ]; then
-		sed -i "s@PBS_START_COMM=1@PBS_START_COMM=0@" /etc/pbs.conf
+		sed -i "s@PBS_START_COMM=.*@PBS_START_COMM=0@" /etc/pbs.conf
+	else
+		sed -i "s@PBS_START_COMM=.*@PBS_START_COMM=1@" /etc/pbs.conf
 	fi
+	sed -i "s@PBS_START_SCHED=.*@PBS_START_SCHED=1@" /etc/pbs.conf
```
hirenvadalia, line 17: sched should be off on mom right?

ci/etc/configure_node.sh
```diff
@@ -5,26 +5,38 @@ if [ -f /src/.config_dir/.requirements_decorator ]; then
...
+	sed -i "s@PBS_START_SERVER=.*@PBS_START_SERVER=1@" /etc/pbs.conf
+	sed -i "s@PBS_START_SCHED=.*@PBS_START_SCHED=1@" /etc/pbs.conf
 fi
 
 if [ "x${NODE_TYPE}" == "xcomm" ]; then
-	sed -i "s@PBS_SERVER=$(hostname)@PBS_SERVER=pbs_server@" /etc/pbs.conf
-	sed -i "s@PBS_START_MOM=1@PBS_START_MOM=0@" /etc/pbs.conf
-	sed -i "s@PBS_START_SERVER=1@PBS_START_SERVER=0@" /etc/pbs.conf
+	sed -i "s@PBS_START_COMM=.*@PBS_START_COMM=1@" /etc/pbs.conf
+	sed -i "s@PBS_SERVER=.*@PBS_SERVER=pbs_server@" /etc/pbs.conf
```
hirenvadalia, line 38: pbs-server

ci/README.md
```diff
@@ -1,109 +1,72 @@
...
-./ci --test="-f pbs_smoketest.py"
-./ci --test="--tags=smoke"
- 
-# The above method are examples how to call a custom test case, 
-# keep in mind that benchpress is called from the installed 'tests' dir of PTL
-```
-* **./ci --run:** This command will manually trigger the build for automatic run of 'build install and test', with last specified options for configure and test (default if no options were specified previously).
+./ci --params 'configure=CFLAGS=" -O2 -Wall -Werror" --prefix=/tmp/pbs --enable-ptl'
+
+# The above command is an example of how to define a custom configure option for pbs.
```
prakashcv13, line 34: Usually - example follows the explanation, let us follow that convention.

ci/README.md
```diff
@@ -1,109 +1,72 @@
...
 ```bash
 ./ci --build-pkgs
 # Above command builds package for the platform ci was started/currently running on.
-
-./ci --build-pkgs='ubuntu:16.04'
-# This will build package on ubuntu:16.04 platform, it will restart ci services and build for the specified platform 
 ```
 
-* **./ci --delete:** which will stop any ongoing build in the container for the moment and remove it build.
+* **./ci --delete:** will delete any running containers and take a backup of logs. Does not take any arguments.
```
prakashcv13, line 66: we should also provide the backup location.

ci/ci
```diff
@@ -38,27 +38,95 @@
...
 
 def log_info(str_to_print):
     t = time.localtime()
     current_time = time.strftime("%H:%M:%S", t)
-    print('\033[1m' + current_time + " ---> " + str_to_print + '\033[0m')
+    print(current_time + " ---> " + str(str_to_print))
+
+
+def log_warning(str_to_print):
+    print("WARNING ::: " + str(str_to_print))
```
prakashcv13, line 114: why use single line functions?

ci/ci
```diff
@@ -80,23 +148,34 @@ def run_cmd(cmd, return_output=False):
...
-    except Exception as e:
-        log_error("Failed\n:")
-        log_error(e)
+    services = get_services_list()
+    services.sort(reverse=True)  # we want server cmds to run first
+    for service in services:
+        cmd = "docker-compose -f docker-compose.json exec " + \
+            service + " bash -c \'" + run_cmd + "\'"
+        if only_run_on != 'all' and cmd.find(only_run_on) == -1:
+            log_info('Skipping on service platform ' + service +
```
prakashcv13, line 161: do you mean to say "skipping service on platform"?

ci/ci
```diff
@@ -80,23 +148,34 @@ def run_cmd(cmd, return_output=False):
...
+                     ' as command to only be run on ' + only_run_on)
```
prakashcv13, line 162: do you mean to say "as command only to be run on"

ci/ci
```diff
@@ -114,25 +193,46 @@ def read_from_file(file_path):
...
-        image_name += '_ci_pbs'
-        image_name = image_name.replace(':', '_')
+
+    images_to_commit = {}
+    timeout = 0
+    services = get_services_list()
+    service_count = len(services)
+    timeout_total = 1 * 60 * 60
+    while service_count > 0:
+        time.sleep(15)
```
prakashcv13, line 205: add a comment as to why sleep is needed.

ci/ci
```diff
@@ -114,25 +193,46 @@ def read_from_file(file_path):
...
-    image_name = image_name.split('=')[1]
-    if image_name[-7:] != '_ci_pbs':
-        image_name += '_ci_pbs'
-        image_name = image_name.replace(':', '_')
+
+    images_to_commit = {}
+    timeout = 0
+    services = get_services_list()
+    service_count = len(services)
+    timeout_total = 1 * 60 * 60
```
prakashcv13, line 203: just say timeout

ci/ci
```diff
@@ -114,25 +193,46 @@ def read_from_file(file_path):
...
+    while service_count > 0:
+        time.sleep(15)
+        timeout = timeout + 15
```
prakashcv13, line 206: instead of timeout use time_spent

ci/ci
```diff
@@ -114,25 +193,46 @@ def read_from_file(file_path):
...
+        if timeout > timeout_total:
+            log_error("build is taking too long, timed out")
+            os._exit(1)
+        status = read_from_file(os.path.join(
+            ci_dirname, CONFIG_DIR, STATUS_FILE))
+        for service in services:
+            if str(status).find(service) != -1:
+                services.remove(service)
+                service_count = service_count-1
```
prakashcv13, line 215: spaces around '-' OR just use service_count -= 1

ci/ci
```diff
@@ -226,217 +495,255 @@ def check_prerequisites():
         log_error("docker not found in PATH")
         return 1
 
+    try:
+        version = subprocess.run(
+            ["docker", "--version"], stdout=subprocess.PIPE)
+        version = str(version.stdout.decode('utf-8')
+                      ).split(' ')[2].split('-')[0].replace('.', '')
+        if int(version) < 17120:
```
prakashcv13, line 503: let us not use magic numbers and define a macro.

ci/ci
```diff
@@ -226,217 +495,255 @@ def check_prerequisites():
...
-        if docker_container.find('Up') == -1:
+        service_count = len(get_services_list())
+        if service_count == 0:
             log_info("No running service found")
             try:
                 log_info('Attempting to start container')
-                os.chdir(pbs_dirname)
-                subprocess.run(["docker-compose", "down"],
+                os.chdir(ci_dirname)
+                subprocess.run(["docker-compose", "-f", "docker-compose.json", "down", "--remove-orphans"],
```
prakashcv13, line 523: this is definitely more than 80 chars, please run pep8

ci/ci
```diff
@@ -226,217 +495,255 @@ def check_prerequisites():
...
                 log_error(e)
         else:
             log_info("running container found")
             return 0
     except:
-        log_error("failed to start ci services")
+        log_error(e)
+
+
+def is_restart_required():
```
prakashcv13, line 547: add a docstring

ci/ci
```diff
@@ -226,217 +495,255 @@ def check_prerequisites():
...
+    potential_list.sort()
+    current_file_services_list.sort()
+    if current_file_services_list != potential_list:
+        build_compose_file(node_config)
+        return True
+    else:
+        return False
+
+
+def initialise_command_loop():
```
prakashcv13, line 572: add a docstring

ci/ci
```diff
@@ -226,217 +495,255 @@ def check_prerequisites():
...
+    if not os.path.exists(target_path):
+        value = "--tags=smoke"
+        write_to_file(target_path, value)
+    target_path = os.path.join(ci_dirname, 'docker-compose.json')
+    if not os.path.exists(target_path):
+        build_compose_file('server=centos:7')
+        run_cmd('docker-compose -f docker-compose.json down --remove-orphans')
+
+
+def run_ci(build_pkgs=False):
```
prakashcv13, line 594: add a docstring

ci/ci
```diff
@@ -226,217 +495,255 @@ def check_prerequisites():
...
-        pbs_dirname = os.path.join(pbs_dirname, 'ci')
-    try:
-        os.chdir(pbs_dirname)
-    except Exception as e:
-        log_error(str(e))
-        log_error("Was correct pbs path provided?")
-        sys.exit(1)
-    build_log_path = os.path.join(pbs_dirname, 'logs', 'build')
-    if args.delete == None and args.start == None and args.local == None:
+    mangle_args = ('run', 'param', 'delete', 'build-pkgs', 'local')
```
prakashcv13, line 657: mangle means destroy. let us use a better name

ci/ci
```diff
@@ -226,217 +495,255 @@ def check_prerequisites():
...
-            cmd = "docker-compose exec pbs-build bash -c \' /src/killit.sh  \'"
-            run_docker_cmd(cmd)
-            log_info(" done stop ")
+        if args.params != None:
+            if args.params[0] != 'called':
+                container_running = False
+                conf_opts = load_conf()
+                for set_opts in args.params:
+                    key = (set_opts).split('=', 1)[0]
+                    value = (set_opts).split('=', 1)[1]
```
prakashcv13, line 679: would 
key, value = (set_opts).split('=', 1)
work?

ci/ci
```diff
@@ -226,217 +495,255 @@ def check_prerequisites():
...
+                    service_count = len(get_services_list())
+                    if service_count > 0:
+                        container_running = True
+                    if key == 'nodes':
```
prakashcv13, line 683: how about using lower(key) == 'nodes'?
similarly for os/OS

ci/do.sh
```diff
@@ -150,21 +141,41 @@ fi
...
   echo "### First time build is complete ###"
+  host=$(hostname -s)
+  echo "READY:${host}" >>${config_dir}/.status
   exit 0
 fi
 
+if [ "x${ID}" == "xcentos" -a "x${VERSION_ID}" == "x8" ]; then
+  if [ ! -f /tmp/swig/swig/configure ]; then
+    # source install swig
+    dnf -y install gcc-c++ byacc pcre-devel
```
prakashcv13, line 152: shouldn't we check if dnf was successful or not?

ci/do.sh
```diff
@@ -150,21 +141,41 @@ fi
...
+    mkdir -p /tmp/swig/
+    cd /tmp/swig
+    git clone https://github.com/swig/swig --branch rel-4.0.0 --single-branch
+    cd swig
+    ./autogen.sh
+    ./configure
```
prakashcv13, line 158: I think using CFLAGS will help.

ci/ci
sandisamp: log_warning,log_error and log_info functions provide a standard way to communicate to the user on the command line, this follows with the design standard. Also if we add future changes (such as color encoding) then it will be easier to implement.

ci/do.sh
sandisamp: This is configure for 'swig' following the same steps as prescribed for a successful installation.

ci/do.sh
sandisamp: The do.sh script has a shebang line of "!/bin/bash -ex" Thus it exits with appropriate exit codes if any command fails.

ci/ci
```diff
@@ -213,230 +400,475 @@ def check_for_existing_image():
...
+            "done delete container and services")
     except Exception as e:
-        log_error(str(e))
-        log_error("Was correct pbs path provided?")
+        log_error("Failed to destroy container and services: " + e)
+
+
+def parse_params(params_list):
+    '''
+    Update given params and run ci
```
prakashcv13, line 696: does it run ci?

ci/ci
```diff
@@ -213,230 +400,475 @@ def check_for_existing_image():
...
+                build_compose_file(value)
+            elif key.lower() == 'os':
+                if container_running:
+                    log_warning(
+                        "Deleting existing containers first, \
+                        find backup in logs folder")
+                    delete_ci()
+                node_string = value.replace('"', '')
+                node_string = 'server=' + node_string
+                build_compose_file(nodes=node_string)
```
prakashcv13, line 721: why nodes=?

ci/ci
```diff
@@ -213,230 +400,475 @@ def check_for_existing_image():
...
+
+
+def parse_params(params_list):
+    '''
+    Update given params and run ci
+    '''
+    if params_list[0] != 'called':
+        container_running = False
+        conf_opts = load_conf()
+        for set_opts in params_list:
```
prakashcv13, line 701: if the user provides both os and nodes - we will call docker_compose twice

ci/ci
```diff
@@ -213,230 +400,475 @@ def check_for_existing_image():
...
-                    write = False
+        if args.params is not None:
+            parse_params(args.params)
+        if args.build_pkgs is not None:
+            build_pkgs = True
+        if args.delete is True:
+            confirm = input(
+                'Are you sure you want to delete containers (Y/N)?: ')
+            if confirm[0].lower() == 'n':
+                os._exit(0)
```
prakashcv13, line 862: in some places you use os._exit and in other sys.exit

ci/ci
```diff
@@ -114,64 +199,152 @@ def read_from_file(file_path):
...
-            log_error('Failed to restart ci.')
-    log_info('The package build logs can be found in logs/pkglogs')
-    run_docker_cmd(
-        'docker-compose exec pbs-build bash -c \' ' + build_cmd + ' | tee /logs/pkglogs \' ')
+def create_ts_tree_json():
+    benchpress_opt = os.path.join(
+        ci_dirname, MACROS['CONFIG_DIR'], MACROS['BENCHPRESS_OPT_FILE'])
+    benchpress_value = read_from_file(benchpress_opt)
+    try:
+        cmd = '/src/etc/gen_ptl_json.sh "' + benchpress_value + '"'
```
prakashcv13, line 265: there are some hard coded values, should they become part of MACROS?

ci/ci
```diff
@@ -38,27 +38,98 @@
...
 
-def log_error(str_to_print):
-    t = time.localtime()
-    current_time = time.strftime("%H:%M:%S", t)
-    print('\033[91m' + current_time + " ---> " + str_to_print + '\033[0m')
 
+def get_services_list():
+    docker_container_process = subprocess.run(
+        ["docker-compose", "-f", "docker-compose.json",
+         "ps", "--filter", "status=running", "--services"],
```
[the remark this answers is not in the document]
hirenvadalia: @prakashcv13 No it won't since docker-compose does list anything except what's defined in compose.json

ci/ci
```diff
@@ -38,27 +38,98 @@
...
+def log_warning(msg):
+    print("WARNING ::: " + str(msg))
 
-def log_error(str_to_print):
-    t = time.localtime()
-    current_time = time.strftime("%H:%M:%S", t)
-    print('\033[91m' + current_time + " ---> " + str_to_print + '\033[0m')
 
+def get_services_list():
+    docker_container_process = subprocess.run(
```
hirenvadalia, line 120: just say `_ps` as variable name and below `_p` for `docker_container`

ci/ci
```diff
@@ -80,23 +151,37 @@ def run_cmd(cmd, return_output=False):
...
-            docker_cmd = shlex.split(
-                "docker-compose exec pbs-build bash -c \'cd /pbssrc && /bin/bash\'")
-            subprocess.run(docker_cmd)
-    except Exception as e:
-        log_error("Failed\n:")
-        log_error(e)
+    services = get_services_list()
+    services.sort(reverse=True)  # we want server cmds to run first
+    for service in services:
+        cmd = "docker-compose -f docker-compose.json exec " + \
```
hirenvadalia, line 162: change this two lines to:
```python
cmd = "docker-compose -f docker-compose.json exec "
cmd += service + "/bin/bash -c \'" + run_cmd + "\'"
```

ci/ci
```diff
@@ -80,23 +151,37 @@ def run_cmd(cmd, return_output=False):
         log_error(e)
 
 
-def run_docker_cmd(cmd):
+def run_docker_cmd(run_cmd, only_run_on='all'):
```
hirenvadalia, line 154: I think we don't need `only_` word in an argument, `run_on` will be good

ci/ci
```diff
@@ -80,23 +151,37 @@ def run_cmd(cmd, return_output=False):
...
+                     ' as command only to be run on ' + only_run_on)
+            continue
+        try:
+            log_info(cmd)
+            docker_cmd = shlex.split(cmd)
+            a = subprocess.Popen(docker_cmd)
+            a.communicate()
+            if a.returncode != 0:
+                log_error(
+                    "docker cmd returned with non zero exit code, \
```
hirenvadalia, line 175: use like below instead using `\` in the string:
_msg = "....."
_msg += "...."
log_error(_msg)

Same for commands

ci/ci
```diff
@@ -80,23 +151,37 @@ def run_cmd(cmd, return_output=False):
...
+                    redirecting you to container terminal")
+                docker_cmd = shlex.split(
+                    "docker-compose -f docker-compose.json exec " +
+                    service + " bash -c \'cd /pbssrc && /bin/bash\'")
+                subprocess.run(docker_cmd)
+                os._exit(1)
```
hirenvadalia, line 181: why exit here? is this child? no right?

ci/ci
```diff
@@ -114,64 +199,152 @@ def read_from_file(file_path):
...
+    service_count = len(services)
+    timeout = 1 * 60 * 60
+    while service_count > 0:
+        time.sleep(15)
+        # Check every 15 seconds if container build is ready
+        # Do not want to check constantly as it increases cpu load
+        time_spent = time_spent + 15
+        if time_spent > timeout:
+            log_error("build is taking too long, timed out")
+            os._exit(1)
```
hirenvadalia, line 217: not child, use sys.exit

ci/ci
sandisamp: We do not want to continue with execution if any command has failed. If we don't use os._exit(1) here ; when user exits from container the process continues.... for example configuration failed , user was redirected to container terminal but when the user exits the container terminal it will continue with make command.

ci/ci
sandisamp: There is no scenario i can see where os and nodes both are required; in any case it will overwrite the params with the latter called command, not breaking anything.

ci/ci
sandisamp: My opinion is that they are not used multiple times; and are unlikely to change in the future. It also provides for better readability of code. MACROS have all those strings which are likely to change or have multiple reference in the file.

PR #1861, opened 2020-06-25
Fixed memory leak in scheduler HUP

bhroam:
#### Describe Bug or Feature
Memory leak when the scheduler is HUP'd

#### Describe Your Change
The scheduler dups some strings when it parsing server_dyn_res.  It does not free these strings on HUP.

BEFORE:
[bmann@mars scheduler]$ grep parse_config valgrind-before.log 
==814486==    by 0x44A932: parse_config (parse.c:611)
==814486==    by 0x44A932: parse_config (parse.c:611)
==814486==    by 0x44A9AA: parse_config (parse.c:620)
==814486==    by 0x44A9BA: parse_config (parse.c:621)
==814486==    by 0x44A9AA: parse_config (parse.c:620)
==814486==    by 0x44A9BA: parse_config (parse.c:621)

AFTER:
[bmann@mars scheduler]$ sudo valgrind --leak-check=full --track-origins=yes --log-file=valgrind.log ./pbs_sched -N
[bmann@mars scheduler]$ grep parse_config valgrind.log

### DIFF
--- a/src/scheduler/parse.c
+++ b/src/scheduler/parse.c
@@ -321,10 +321,6 @@ parse_config(char *fname)
 					if (!type.is_time)
 						error = 1;
 				}
-				else if (!strcmp(config_name, PARSE_SYNC_TIME)) {
-					obsolete[0] = PARSE_SYNC_TIME;
-					obsolete[1] = "nothing - syncs happen automatically";
-				}
 				else if (!strcmp(config_name, PARSE_UNKNOWN_SHARES))
 					conf.unknown_shares = num;
 				else if (!strcmp(config_name, PARSE_FAIRSHARE_DECAY_FACTOR)) {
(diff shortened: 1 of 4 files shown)


src/scheduler/parse.c
```diff
@@ -934,7 +943,6 @@ init_config()
 	conf.update_comments = 1;
 
 	conf.decay_time = 86400;	/* default decay time period is 24 hours */
-	conf.sync_time = 86400;
```
agrawalravi90, line 937: was this intentional?

bhroam: Oh, yeah, it was.  I can back it out if you want.  That's literally the only place sync time is set/used.

agrawalravi90: oh, so should we just remove conf.sync_time altogether?

bhroam: Sure, let me remove it.  I should probably file a doc bug to get it removed from the man page because you can't set it.

PR #1864, opened 2020-06-25
Fixed test_stale_degraded failure due to nodes not being free when reservation is submitted

Klovely:
#### Describe Bug or Feature
test_stale_degraded fails due to nodes not being free when reservation is submitted.
Test is failing  because at the time of reservation submission , node was in initializing state so reservation got denied and deleted immediately .
Race condition in test script 

#### Describe Your Change
In test, before reservation submission confirm node state in free state.

[Test_stale_degraded_before_fix.txt](https://github.com/openpbs/openpbs/files/4833845/Test_stale_degraded_before_fix.txt)

[Test_stale_degraded_after_fix.txt](https://github.com/openpbs/openpbs/files/4833892/Test_stale_degraded_after_fix.txt)

### DIFF
--- a/test/tests/functional/pbs_conf_resv_stale_vnode.py
+++ b/test/tests/functional/pbs_conf_resv_stale_vnode.py
@@ -105,6 +105,8 @@ class TestResvStaleVnode(TestFunctional):
         Test that a reservation goes into the degraded state
         when one of its vnodes go stale
         """
+        self.server.expect(NODE, {'state': 'free'}, id='vn[0]')
+        self.server.expect(NODE, {'state': 'free'}, id='vnode[0]')
         now = int(time.time())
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_conf_resv_stale_vnode.py
```diff
@@ -105,6 +105,8 @@ def test_stale_degraded(self):
         Test that a reservation goes into the degraded state
         when one of its vnodes go stale
         """
+        self.server.expect(NODE, {'state': 'free'}, id='vn[0]')
+        self.server.expect(NODE, {'state': 'free'}, id='vnode[0]')
```
bhroam, line 109: How about doing this all in one expect() by saying {'state=free': 2} or however many there are.

PR #1866, opened 2020-06-26
qrun subjob fails

arungrover:
#### Describe Bug or Feature
Qrun of a subjob fails

#### Describe Your Change
Changes made in #1188 caused scheduler to lower down the preemption priority of the subjob given with qrun command.
This causes scheduler to reject preemption because we don't preempt when the job being run lowers down its preemption priority.
Change is to make sure that in case of qrun, we check the preemption priority of its array parent instead of the subjob.

#### Link to Design Doc
NA

[test.txt](https://github.com/openpbs/openpbs/files/4834604/test.txt)

### DIFF
--- a/src/scheduler/job_info.c
+++ b/src/scheduler/job_info.c
@@ -3897,8 +3897,11 @@ set_preempt_prio(resource_resv *job, queue_info *qinfo, server_info *sinfo)
 	jinfo->preempt = 0;
 	jinfo->preempt_status = 0;
 
-	if (job == sinfo->qrun_job)
-		jinfo->preempt_status |= PREEMPT_TO_BIT(PREEMPT_QRUN);
+	if (sinfo->qrun_job != NULL) {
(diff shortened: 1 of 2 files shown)


src/scheduler/job_info.c
```diff
@@ -3897,8 +3897,11 @@ set_preempt_prio(resource_resv *job, queue_info *qinfo, server_info *sinfo)
 	jinfo->preempt = 0;
 	jinfo->preempt_status = 0;
 
-	if (job == sinfo->qrun_job)
-		jinfo->preempt_status |= PREEMPT_TO_BIT(PREEMPT_QRUN);
+	if (sinfo->qrun_job != NULL) {
+		if (job == sinfo->qrun_job ||
+		    (job->job->is_subjob && (strcmp(job->job->array_id, sinfo->qrun_job->name) == 0)))
```
bhroam, line 3902: Use jinfo instead of job->job

PR #1867, opened 2020-06-29
race condition in execution of endjob hook and job purge leads to mom crash

prakashcv13:
#### Describe Bug or Feature

When we have a mix of cpuset and non-cpuset nodes in a cluster, TestPbsExecutePrologue.test_prologue_exception_sisters crashes the mom process on cpuset node.
a) This test creates a hook that raises exception in a execjob_prologue hook in sister moms.
b) As at least one of the sister mom is cpuset, cgroups hook is enabled with vnode_per_numa=true.
This is a race condition - events are as follows -
1) job is sent to moms for execution.
2) pbs_cgroups hook sets up the cgroups for the job in execjob_begin event. 
3) job fails on sister moms because of exception in execjob_prologue hook.
4) on cpuset mom, pbs_cgroups execjob_end hook is run to clean up the cgroups created in execjob_begin.
5) execjob_end runs in a separate process. pbs_mom creates a task (post_run_hook) to come back to it once the
execjob_end hook completes.
6) On the other hand, the job is continuously rerun, and on receiving the rerun request, the mom purges the job.
7) Now, when execjob_end hook finishes execution, mom tries to access pjob structure that was purged.
 
#### Describe Your Change

To prevent mom from accessing the free-ed data, we delete the post_run_hook task before purging the job in step 6.

[after-fix.txt](https://github.com/openpbs/openpbs/files/4847425/after-fix.txt)

### DIFF
--- a/src/include/job.h
+++ b/src/include/job.h
@@ -459,6 +459,7 @@ struct job {
 	int		ji_parent2child_moms_status_pipe;	/* write pipe for parent mom to send sister moms status to child starter process */
 	int		ji_updated;	/* set to 1 if job's node assignment was updated */
 	time_t		ji_walltime_stamp;	/* time stamp for accumulating walltime */
+	struct work_task *ji_bg_hook_task;
 #ifdef WIN32
 	HANDLE		ji_momsubt;	/* process HANDLE to mom subtask */
 #else	/* not WIN32 */
--- a/src/include/job.h
+++ b/src/include/job.h
@@ -753,7 +753,7 @@ typedef struct	infoent {
 #define IM_RESTART		16
 #define IM_DELETE_JOB		17
 #define IM_REQUEUE		18
-#define IM_DELETE_JOB_REPLY		19
+#define IM_DELETE_JOB_REPLY	19
 #define IM_SETUP_JOB		20
 #define IM_DELETE_JOB2		21	/* sent by sister mom to delete job early */
 #define IM_SEND_RESC		22
--- a/src/resmom/catch_child.c
+++ b/src/resmom/catch_child.c
@@ -2472,7 +2472,7 @@ mom_deljob(job *pjob)
 /**
  * @brief
  * 	mom_deljob_wait - deletes most of the job stuff, job entry not deleted
- *	untill the sisters have rplied or are down
+ *	until the sisters have replied or are down
  *	This version DOES wait for the Sisters to reply, see processing of
  *	IM_DELETE_JOB_REPLY in mom_comm.c
  *	IT should only be called for a job for which this is Mother Superior.
--- a/src/resmom/catch_child.c
+++ b/src/resmom/catch_child.c
@@ -2698,36 +2698,6 @@ mom_set_use_all(void)
 void
 job_purge_mom(job *pjob)
 {
-	extern pbs_list_head task_list_event;
-	extern int post_run_hook(struct work_task *wtask);
-
-	struct work_task *wtask = NULL;
-	struct work_task *wtask_next = NULL;
-
-	/* Check for other task lists */
-	for (wtask = (struct work_task *) GET_NEXT(task_list_event); wtask; wtask = wtask_next) {
-		wtask_next = (struct work_task *) GET_NEXT(wtask->wt_linkall);
-
-		if ((wtask->wt_type == WORK_Deferred_Child) &&
-			((int (*)())wtask->wt_func == post_run_hook) &&
-			(wtask->wt_parm2 != NULL)) {
-			mom_process_hooks_params_t *php = NULL;
-			mom_hook_input_t *hook_input = NULL;
-			job   *task_pjob = NULL;
-
-			php = (mom_process_hooks_params_t *)wtask->wt_parm2;
-			hook_input = (mom_hook_input_t *) php->hook_input;
-			if (hook_input) {
-				task_pjob = (job *)hook_input->pjob;
-				if (strcmp(task_pjob->ji_qs.ji_jobid, pjob->ji_qs.ji_jobid) == 0) {
-					sprintf(log_buffer, "deleting stale hook task for job %s", pjob->ji_qs.ji_jobid);
-					log_err(0, __func__, log_buffer);
-					delete_task(wtask);
-				}
-			}
-		}
-	}
-
 	if (mock_run)
 		mock_run_job_purge(pjob);
 	else
--- a/src/resmom/catch_child.c
+++ b/src/resmom/catch_child.c
@@ -2698,6 +2698,36 @@ mom_set_use_all(void)
 void
 job_purge_mom(job *pjob)
 {
+	extern pbs_list_head task_list_event;
+	extern int post_run_hook(struct work_task *wtask);
+
+	struct work_task *wtask = NULL;
+	struct work_task *wtask_next = NULL;
+
+	/* Check for other task lists */
+	for (wtask = (struct work_task *) GET_NEXT(task_list_event); wtask; wtask = wtask_next) {
+		wtask_next = (struct work_task *) GET_NEXT(wtask->wt_linkall);
+
+		if ((wtask->wt_type == WORK_Deferred_Child) &&
+			((int (*)())wtask->wt_func == post_run_hook) &&
+			(wtask->wt_parm2 != NULL)) {
+			mom_process_hooks_params_t *php = NULL;
+			mom_hook_input_t *hook_input = NULL;
+			job   *task_pjob = NULL;
+
+			php = (mom_process_hooks_params_t *)wtask->wt_parm2;
+			hook_input = (mom_hook_input_t *) php->hook_input;
+			if (hook_input) {
+				task_pjob = (job *)hook_input->pjob;
+				if (strcmp(task_pjob->ji_qs.ji_jobid, pjob->ji_qs.ji_jobid) == 0) {
+					sprintf(log_buffer, "deleting stale hook task for job %s", pjob->ji_qs.ji_jobid);
+					log_err(0, __func__, log_buffer);
+					delete_task(wtask);
+				}
+			}
+		}
+	}
+
 	if (mock_run)
 		mock_run_job_purge(pjob);
 	else
--- a/src/resmom/mom_hook_func.c
+++ b/src/resmom/mom_hook_func.c
@@ -954,7 +954,7 @@ run_hook(hook *phook, unsigned int event_type, mom_hook_input_t *hook_input,
 
 	if (runas_jobuser) {
 
-		/* use world writtable spool directory */
+		/* use world writable spool directory */
 
 		snprintf(hook_inputfile, MAXPATHLEN,
 			FMT_HOOK_INFILE, path_spool,
@@ -1177,7 +1177,7 @@ run_hook(hook *phook, unsigned int event_type, mom_hook_input_t *hook_input,
 
 		log_file[0] = '\0';
 
-		if ((fp=fopen(hook_inputfile, "w")) == NULL) {
+		if ((fp = fopen(hook_inputfile, "w")) == NULL) {
 			sprintf(log_buffer,
 				"open of input file %s failed!",
 				hook_inputfile);
--- a/src/resmom/mom_hook_func.c
+++ b/src/resmom/mom_hook_func.c
@@ -3230,7 +3230,7 @@ record_job_last_hook_executed(unsigned int hook_event,
  * @return 0 a hook rejected
  * @return -1 an internal error occurred
  */
-static int
+int
 post_run_hook(struct work_task *ptask)
 {
 
--- a/src/resmom/mom_hook_func.c
+++ b/src/resmom/mom_hook_func.c
@@ -886,12 +886,13 @@ run_hook(hook *phook, unsigned int event_type, mom_hook_input_t *hook_input,
 				log_err(errno, __func__, msg_err_malloc);
 				return (-1);
 			}
-		 	if (php)
+		 	if (php) {
 				ptask->wt_parm2 = (void *)php;
+				php->hook_input->pjob->ji_bg_hook_task = ptask;
+			}
 			return (0);	/* no hook output file at this time */
 		} else if (php)
 			php->child = child;
-
 
 		set_alarm(phook->alarm, run_hook_alarm);
 		while (waitpid(child, &waitst, 0) < 0) {	/* error on wait */
@@ -3477,10 +3478,9 @@ post_run_hook(struct work_task *ptask)
 	}
 
 	if (!php->parent_wait) {
-		/* Backround hook */
+		/* Background hook */
 		/* Create task to check and run next hook script */
-		new_task = set_task(WORK_Immed, 0,
-						(void *)mom_process_background_hooks, phook);
+		new_task = set_task(WORK_Immed, 0, (void *)mom_process_background_hooks, phook);
 		if (!new_task)
 			log_err(errno, __func__,
 				"Unable to set task for mom_process_background_hooks");
@@ -3502,7 +3502,8 @@ post_run_hook(struct work_task *ptask)
  * @return void
  */
 
-void reply_hook_bg(job *pjob)
+void
+reply_hook_bg(job *pjob)
 {
 	int	n = 0;
 	int	ret = 0;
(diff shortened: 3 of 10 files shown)


src/resmom/catch_child.c
```diff
@@ -2694,6 +2694,36 @@ mom_set_use_all(void)
...
 job_purge_mom(job *pjob)
 {
+	extern pbs_list_head task_list_event;
+	extern int post_run_hook(struct work_task *wtask);
+
+	struct work_task *wtask = NULL;
+	struct work_task *wtask_next = NULL;
+
+	/* Check for other task lists */
+	for (wtask = (struct work_task *) GET_NEXT(task_list_event); wtask; wtask = wtask_next) {
```
hirenvadalia, line 2704: I don't think this is right fix and place, this will kill run performance as for every job purge we will go through whole task event list. Also this function is just wrapper of actual job purge vs mock run job purge. And in mock run mode mom doesn't run any hook (aka no fork at all).

I think what we should do is hang work task pointer in pjob struct, inside run_hook and in job_purge() just make that work task's wt_param2 null, and in post_run_hook check for null of param2, if null don't do anything just return.
that way we don't have to delete task also, or if we want we can delete task also with ensuring child process is also killed and reaped.

bhroam:
Shouldn't the job stick around until the end hook is complete?  What if the end hook does something to the job that needs to be reported back to the server?

prakashcv13:
@bhroam - the job has actually reported back an exception through the prologue hook, that is the reason it is rerun.

bhroam:
@prakashcv13 I admit I don't know a whole lot about hooks, but it seems to me that we'd want the job to stick around until the end hook finished.  Even though the job ever started, we don't know what the end hooks are doing.  Can an end hook do anything that might require the job?  In this instance, we don't because we are tearing down cgroups.  This is not the only purpose of end hooks though.

prakashcv13:
@bhroam - same here - not an expert on hooks. But, few months back we decided that we would run execjob_end hooks in a separate process so as to free the mom to proceed with other requests.

bhroam:
I think @vstumpf can answer why the end hook now runs in the background.  I think it has to do with not having the mom marked down or allowing it to service multiple jobs while the hook is running.  I don't think it was done so the job could continue on and get purged.  I might be wrong.

prakashcv13:
right @bhroam - job being purged is the side effect of running the end job hook in the background.

prakashcv13:
test and valgrind logs
a) PTL logs
[ptl.txt](https://github.com/openpbs/openpbs/files/4866664/ptl.txt)

b) PTL with mom under valgrind
[ptl_val.txt](https://github.com/openpbs/openpbs/files/4866663/ptl_val.txt)

c) valgrind logs - MS
[val_ms.txt](https://github.com/openpbs/openpbs/files/4866661/val_ms.txt)

d) valgrind logs - sister - showing mem leak related to send_hook_fail_action()
[val_sister_before.txt](https://github.com/openpbs/openpbs/files/4866658/val_sister_before.txt)

e) valgrind logs - sister - after fixing mem leak related to send_hook_fail_action()
[val_sister_after.txt](https://github.com/openpbs/openpbs/files/4866659/val_sister_after.txt)

bayucan:
@bhroam : the only thing that could get lost if the post_run_hook() is not run after execjob_end hook complete, are any vnode updates (like setting state=offline) to be sent to the server. This would be the side effect of this fix. Perhaps it should just be documented. That is, if a job is requeued/rerun after some mom hooks have failed, any execjob_end action to update vnodes info could get lost. There'll be no job updates (like resouces_used)  by execjob_end hook that would take effect, and this is documented, since job is already being deleted.

src/server/job_func.c
```diff
@@ -566,6 +566,18 @@ job_free(job *pj)
...
+	if (pj->ji_bg_hook_task) {
+		mom_process_hooks_params_t *php;
+		php = pj->ji_bg_hook_task->wt_parm2;
+		if (php->hook_output) {
+			free(php->hook_output->reject_errcode);
+			free(php->hook_output);
+		}
+		free(php->hook_input);
+		free(php);
+		pj->ji_bg_hook_task->wt_parm2 = NULL;
```
bayucan, line 578: I think you should just delete the task altogether here, like adding delete_task(pj-JI_bg_hook_task). Otherwise, main mom would continue to see the task, process it, and although upon seeing pj->ji_bg_hook_task->wt_parm2 (under post_run_hook()) being NULL, it would return immediately. But this would result in a mom_logs message "missing hook params argument to event". So I think it's better to delete the task now.

src/server/job_func.c
```diff
@@ -566,6 +566,18 @@ job_free(job *pj)
 	reliable_job_node_free(&pj->ji_failed_node_list);
 	reliable_job_node_free(&pj->ji_node_list);
 
+	if (pj->ji_bg_hook_task) {
+		mom_process_hooks_params_t *php;
+		php = pj->ji_bg_hook_task->wt_parm2;
```
bayucan, line 571: php could be NULL here, so best to check for NULL-ness before use.

src/include/job.h
```diff
@@ -570,6 +570,7 @@ struct job {
 	int		ji_parent2child_moms_status_pipe;	/* write pipe for parent mom to send sister moms status to child starter process */
 	int		ji_updated;	/* set to 1 if job's node assignment was updated */
 	time_t		ji_walltime_stamp;	/* time stamp for accumulating walltime */
+	struct work_task *ji_bg_hook_task;
```
bayucan, line 573: Since this a new pjob field, add to job_alloc() in job_func.c:
#if defined(PBS_MOM) && defined(WIN32)
        pj->ji_momsubt = NULL;
+     pj->ji_bg_hook_task = NULL;           <-- initialize here
#endif

prakashcv13:
Hey @bayucan, I have addressed your comments.
test logs -
[ptl_cpuset.txt](https://github.com/openpbs/openpbs/files/4884867/ptl_cpuset.txt)
[ptl_non_cpuset.txt](https://github.com/openpbs/openpbs/files/4884874/ptl_non_cpuset.txt)

bhroam:
@bayucan thank you for your inputs.  Maybe I am worrying about nothing if we already document things that might not be set in the end hook already.  My only worry is that it makes the hook situational.  The hook writer can not be certain the updates they are trying to do will be done based on other situations they can not detect.  Like I said before, hooks are not really my thing, so if you think this is fine, so do I.

bayucan:
@bhroam : thanks. We'll have this side effect documented.

prakashcv13:
@bayucan - thank you for your review and approval. I will create a doc bug as you suggested.

prakashcv13:
@bhroam - could you please review/merge if you are happy with the changes.

src/resmom/mom_hook_func.c
```diff
@@ -886,16 +885,18 @@ run_hook(hook *phook, unsigned int event_type, mom_hook_input_t *hook_input,
 				log_err(errno, __func__, msg_err_malloc);
 				return (-1);
 			}
-		 	if (php)
+		 	if (php) {
 				ptask->wt_parm2 = (void *)php;
+				if (php->hook_input && php->hook_input->pjob)
+					php->hook_input->pjob->ji_bg_hook_task = ptask;
```
bhroam, line 891: This assumes there is only one end hook.  There can be multiple hooks with the same hook event.  How do they get run in the background?  One after the other?  If we continue to run and purge the job after the first one, how do we run the subsequent ones?  Also, this is assuming there is just one task.  Will this single task pointer work with multiple end hooks?

src/resmom/mom_hook_func.c
```diff
@@ -3081,17 +3079,17 @@ struct work_task *pwt;
...
+	int vret = -1;
 
 	if ((phook == NULL) || (phook->hook_name == NULL)) {
 		log_event(PBSEVENT_ERROR, PBS_EVENTCLASS_HOOK,
 			  LOG_ERR, __func__, "Hook received is NULL");
 		return;
 	}
 
-	snprintf(hook_buf, sizeof(hook_buf)-1, "1,%s", phook->hook_name);
+	snprintf(hook_buf, sizeof(hook_buf) - 1, "1,%s", phook->hook_name);
```
bhroam, line 3092: No need for the sizeof() -1.  snprintf() will handle the null byte. (here and other places)

src/resmom/mom_hook_func.c
```diff
@@ -3944,8 +3944,7 @@ mom_process_hooks(unsigned int hook_event, char *req_user, char *req_host,
 		/* Set it only once, and only if there's a hook to execute  */
 		/* since we're affecting the job directly.		    */
 		if( ((hook_event == HOOK_EVENT_EXECJOB_END) || \
-		     (hook_event == HOOK_EVENT_EXECJOB_EPILOGUE)) &&
-			!set_job_exit) {
+		     (hook_event == HOOK_EVENT_EXECJOB_EPILOGUE)) && !set_job_exit) {
```
bhroam, line 3947: No need for  the \\ on the above line

src/resmom/mom_hook_func.c
prakashcv13: this was existing code - I just added spaces. should this be taken care of in this PR?

src/resmom/mom_hook_func.c
prakashcv13: existing code - I will remove the '\'

src/resmom/mom_hook_func.c
prakashcv13: mom_process_background_hooks() handles running subsequent end job hooks.
After the first background hook runs to completion - post_run_hook() is run.
post_hook_run() will create a task (with func = mom_process_background_hooks) to check and run next hook script.
So, there is actually only one task at a time, subsequent end job hooks are run by creating tasks.

With this fix, we do not run the subsequent endjob hooks.

src/resmom/mom_hook_func.c
bhroam: Up to you.  I guess if it is existing, then there is more cases than just the few your modified.

prakashcv13: I changed the ones I could find :).

src/resmom/mom_hook_func.c
bhroam: I think it is a bad idea to not run all of the end hooks.  It would be possible for the site to add several end hooks with order < 100, and then the cgroups hook might not run and the cgroups won't be torn down.  Even if we document this, we're telling the sites that if they have multiple end hooks, they might not all run.  To me this breaks our hook infrastructure.  How hard would it be to keep the job around and then purge it in the last call of post_run_hook()?
Hmm here's another wrinkle.  What happens if we run one end hook, and  by the time it finishes, the job is requeued and run again on this node.  Will post_run_hook() find the new version of the job and run more end hooks?

bayucan: @bhroam : you've made a good point! True, if there are multiple end hooks, the pbs_cgroups end hook might not execute, and cgroups won't get cleaned up. That would be bad. If there was only one end hook, which can be checked by doing:

GET_NEXT(pjob->ji_bg_hook_task->wt_parm1->hi_execjob_end_hooks) == NULL
then the current fix scheme would work. But if there are other execjob_end hooks, we might just let the remaining execjob_end hooks execute in blocking mode. This means mom would block until all the execjob_end hooks execute on behalf of the job, before proceeding to free the job totally.
Now trying to keep the job around and then purge in the last call of post_run_hook() seemed tricky. We would need to go through the mom code looking for the job_purge_mom() calls. @prakashcv13 : try to see the actual sequence that happens when a job is interrupted (like the prologue hook rejecting request), and then requeued/rerun. I just don't know where in mom code that particular job_purge_mom() call happens, and if it's safe to not let it happen...

bayucan: @prakashcv13 : basically, we'd like to achieve what happens when a normal,  non-interrupted job is executing end hooks, where it would linger around in Exiting state, until the last end hook executes. So if a job is interrupted, the job would still stay around until the final end hook executes. This means requeueing and reruning would get delayed as well.

prakashcv13: the sequence of events goes like this:
1) MS joins job with sisters
2) sisters reject the prologue event
3) MS aborts the job - runs the execjob_abort hooks including the cgroup hook which deletes the cgroups related to the job
4) MS sends abort request to sisters - sisters also run execjob_abort hooks
5) sisters purge the job
6) MS sends job obit to server
7) server reruns the job - sends delete job request from on_job_rerun
8) MS relays the request to sisters
9) MS runs endjob hooks
10) sisters do not find the job as it was purged in step   above - return error PBSE_JOBEXIST
11) MS rejects the delete job request from the server (chk_del_job) BUT adds the job the list of dead jobs
12) MS on itself purges the job in it's main loop
13) server on getting the reject reply (15059) sends a discard job request
14) MS does not find the job as it is already purged AND sends a success reply
15) server requeues the job
16) scheduler runs the job on the same nodes
17) post_run_hook() runs and MS crashes

prakashcv13: @bayucan, @bhroam - if the concern is whether the cgroups related to the job are getting deleted - they are getting deleted in the abort job event and we can go ahead with the fix, this is a very corner case according to me.
if the concern is that *all* endjob hooks should run - it is a race. Please let me know what do you guys think.

bhroam: Interesting, so the abort hook and the end hook runs?  I thought the purpose of the abort hook was to run it and not the end hook.  There were cases where the end hook wasn't running.  We didn't want to run the end hook there because existing end hooks might be run when the customer didn't expect them.  This is really weird because the abort hooks run everywhere, but only the end hook runs on MS.  I really dislike the fact that all the end hooks won't run.  The customer is telling us that they have N hooks and expect them to all run at the end of a job.  In this case, only some of them will run, and the number won't be predictable.  If the hooks are fast, you might run several, but if the hooks take a while, you'll run less.  Maybe the answer is to make the abort hook or the end hook run, but not both.  That will fix this problem since the abort hook will run.

bayucan: @prakashcv13 : Thank you for enumerating the steps that mom and server go through. This is really helpful. 
it is true that pbs_cgroups hook guards possible failures by doing clean up in the execjob_abort, in case execjob_end hook is not called. So the concern for the pbs_cgroups hook side is alleviated. 
However, we should explore if there's another way so that all execjob_end hooks are given a chance to execute before a job is rerun. Looking at the 4 steps below:
11. MS rejects the delete job request from the server (chk_del_job) BUT adds the job the list of dead jobs
12. MS on itself purges the job in it's main loop
13. server on getting the reject reply (15059) sends a discard job request
14. MS does not find the job as it is already purged AND sends a success reply

I believe the 2 places in mom that looks into mom_deadjobs list is as follows:
```
             i = 0;
                while ((pjob = (job *)GET_NEXT(mom_deadjobs)) != NULL) {
                        /* sometimes this purge is happening earlier than
                         * IS_DISCARD_JOB, which then does not get the pjob
                         * pointer to call kill_job().
                         *
                         * Fixed by adding a kill_job here, which should be
                         * no harm anyway.
                         */
                        (void)kill_job(pjob, SIGKILL);
                        job_purge_mom(pjob);
                        ++i;
                }
```
...
   ```
   while ((pjob = (job *)GET_NEXT(mom_deadjobs)) != NULL) {
                job_purge_mom(pjob);
        }
```

`
What if we prevent the call to job_purge_mom() from taking place in the loop body, until all end hooks have been processed. Anyway, the job would still be in the mom_deadjobs() list which mom's main loop would continue to check.

Now in step 14 above, perhaps MS should not only check if it does not find the job and reply success, but check also if job is NOT in mom_deadjobs, before replying success. Maybe that would prevent server from requesting the job to be requeued and prevent it from rerunning as job would still be in a R state (or actually E state). Eventually, job would disappear in mom_deadjobs list, and mom and server would be in sync and job goes back to Q state...

prakashcv13:
hi @bayucan, @bhroam - PR is ready for review.
[PTL - ptl_cpuset_1307.txt](https://github.com/openpbs/openpbs/files/4920949/ptl_cpuset_1307.txt)
[PTL with valgrind enabled ptl_cpuset_val_1307.txt](https://github.com/openpbs/openpbs/files/4920950/ptl_cpuset_val_1307.txt)
[valgrind logs val_1307.txt](https://github.com/openpbs/openpbs/files/4920951/val_1307.txt)

bhroam:
I'm confused.  I thought the idea was to leave the job around until all the hooks finished.  Now you removed the check if ji_hook_running_bg_on around the purge.  Won't the purge always happen now?

prakashcv13:
hey @bhroam - I have removed the check that I had put in mom_main.c
Instead of not purging a dead job - what I am doing now is that if ji_hook_running_bg_on is not BG_NONE, the job does not become a deadjob itself, so that check was not required.

prakashcv13:
Hey Al, thank you for the review again :).

> Please attach the latest test logs with these changes.

logs are already attached in my [comment](https://github.com/openpbs/openpbs/pull/1867#issuecomment-658376446) yesterday.

Thanks!

PR #1868, opened 2020-06-30
Reservation jobs might run on non reservation nodes

arungrover:
#### Describe Bug or Feature
Sometimes when a degraded reservation starts running and when it has no nodes associated to it, the jobs in these reservations might run on any available nodes.
This issue can happen only when all the reservation's nodes are taken away from reservation before it could start. Maintenance reservations can cause this issue when it completely takes over the nodes from and existing confirmed reservation, leaving the confirmed reservation with no nodes at all.

#### Describe Your Change

The scheduler already detects and ignores any running reservation that does not have any reserved nodes in them. But it does not mark the jobs in the reservation queues as can-not-run. This makes the scheduler consider those queued jobs as any other normal queued job and makes the job run on nodes that are not part of the reservation.
After this change, when the scheduler ignores the reservations, it will also mark all the jobs in the associated queue as can-never-run.

#### Link to Design Doc
NA

[test.txt](https://github.com/openpbs/openpbs/files/4848949/test.txt)

### DIFF
--- a/src/scheduler/resv_info.c
+++ b/src/scheduler/resv_info.c
@@ -119,6 +119,7 @@ stat_resvs(int pbs_sd)
 
 /**
  *
+ * @brief
  *	query_reservations - query the reservations from the server.
  *
  *  Each reservation, is created to reflect its current state in the server.
@@ -132,14 +133,15 @@ stat_resvs(int pbs_sd)
  *  reservation retains its currently allocated resources, such that no other
  *  requests make use of the same resources.
  *
- *	  sinfo - the server to query from
- *	  resvs - batch status of the stat'ed reservations
+ * @param[in] sinfo  - the server to query from
+ * @param[in] resvs  - batch status of the stat'ed reservations
+ * @param[in] pbs_sd - connection to the pbs server
  *
- *	returns an array of reservations
+ * @return    An array of reservations
  *
  */
 resource_resv **
-query_reservations(server_info *sinfo, struct batch_status *resvs)
+query_reservations(server_info *sinfo, struct batch_status *resvs, int pbs_sd)
 {
 	/* the current reservation in the list */
 	struct batch_status *cur_resv;
@@ -239,6 +241,15 @@ query_reservations(server_info *sinfo, struct batch_status *resvs)
 
 		if (ignore_resv == 1) {
 			sinfo->num_resvs--;
+			/* mark all the jobs of the associated queue as can never run */
+			if (resresv->resv->queuename != NULL) {
+				queue_info *qinfo = find_queue_info(sinfo->queues, resresv->resv->queuename);
+				if (qinfo != NULL) {
+				    clear_schd_error(err);
+				    set_schd_error_codes(err, NEVER_RUN, NO_TOTAL_NODES);
+				    update_jobs_cant_run(pbs_sd, qinfo->jobs, NULL, err, START_WITH_JOB);
+				}
+			}
 			free_resource_resv(resresv);
 			continue;
 		}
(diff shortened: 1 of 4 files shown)


test/tests/functional/pbs_maintenance_reservations.py
```diff
@@ -639,3 +639,58 @@ def test_maintenance_progressive_degrade_reservation(self):
         time.sleep(30)
 
         self.server.expect(JOB, {'job_state': 'Q'}, id=jid1)
+
+    def test_maintenance_degrade_reservation_jobs_dont_run(self):
```
vstumpf, line 643: use the @requirements decorator?

test/tests/functional/pbs_maintenance_reservations.py
```diff
@@ -639,3 +639,58 @@ def test_maintenance_progressive_degrade_reservation(self):
...
+        resv_state = {'reserve_state': (MATCH_RE, 'RESV_RUNNING|5')}
+        self.logger.info('Sleeping until reservation starts')
+        offset = start - int(time.time())
+        self.server.expect(RESV, resv_state, id=rid,
+                           offset=offset)
+        self.server.expect(RESV, resv_state, id=rid2,
+                           offset=offset)
+
+        self.logger.info('wait 5 seconds to make sure scheduler looks at job')
+        time.sleep(5)
```
vstumpf, line 693: maybe use offset instead of a raw sleep?

src/scheduler/resv_info.c
```diff
@@ -132,14 +133,15 @@ stat_resvs(int pbs_sd)
...
+ * @param[in] resvs  - batch status of the stat'ed reservations
+ * @param[in] pbs_sd - connection to the pbs server
  *
- *	returns an array of reservations
+ * @return    An array of reservations
  *
  */
 resource_resv **
-query_reservations(server_info *sinfo, struct batch_status *resvs)
+query_reservations(server_info *sinfo, struct batch_status *resvs, int pbs_sd)
```
bhroam, line 144: Please swap the order of arguments.  If we pass in a pbs_sd it comes first (unless policy is there, then it comes second).

src/scheduler/resv_info.c
```diff
@@ -239,6 +241,16 @@ query_reservations(server_info *sinfo, struct batch_status *resvs)
 
 		if (ignore_resv == 1) {
 			sinfo->num_resvs--;
+			/* mark all the jobs of the associated queue as can never run */
+			if (resresv->resv->queuename != NULL) {
+				queue_info *qinfo = find_queue_info(sinfo->queues, resresv->resv->queuename);
+				if (qinfo != NULL) {
+				    clear_schd_error(err);
+				    set_schd_error_arg(err, SPECMSG, "Reservation is in invalid state");
```
bhroam, line 249: Either make this 'Reservation is in an invalid state' or 'Reservation is invalid'

src/scheduler/resv_info.h
```diff
@@ -55,7 +55,7 @@ struct batch_status *stat_resvs(int pbs_sd);
 /*
  *	query_reservations - query the reservations from the server
  */
-resource_resv **query_reservations(server_info *sinfo, struct batch_status *resvs);
+resource_resv **query_reservations(server_info *sinfo, struct batch_status *resvsi, int pbs_sd);
```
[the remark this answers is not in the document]
arungrover: oops

test/tests/functional/pbs_maintenance_reservations.py
arungrover: Based on Bhroam's comments, I'll just trigger cycle and remove sleep.

arungrover:
[test.txt](https://github.com/openpbs/openpbs/files/4886690/test.txt)

arungrover:
Thanks @bhroam and @vstumpf for reviewing the PR.
It's good to go.

PR #1876, opened 2020-07-02
Check if enough numa vnodes before running TestPbsCpuset

vccardenas:
#### Describe Bug or Feature
TestPbsCpuset.test_reliable_job_startup_on_cpuset fails on systems that have less than the needed 2 numa vnodes.

#### Describe Your Change
Added a check if there are enough needed numa vnodes on the system. Otherwise the test will skip.

[TestPbsCpuset_before.txt](https://github.com/openpbs/openpbs/files/4866100/TestPbsCpuset_before.txt)
[TestPbsCpuset_after.txt](https://github.com/openpbs/openpbs/files/4866103/TestPbsCpuset_after.txt)

### DIFF
--- a/test/tests/functional/pbs_cpuset.py
+++ b/test/tests/functional/pbs_cpuset.py
@@ -87,6 +87,11 @@ time.sleep(20)
         The released numa node can be used in another job.
         """
 
+        # Check that there are at least a natural vnode + 2 numa vnodes
+        nodeinfo = self.server.status(NODE)
+        if len(nodeinfo) < 3:
+            self.skipTest("Not enough vnodes to run the test.")
+
         # instantiate execjob_launch hook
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_cpuset.py
```diff
@@ -87,6 +87,11 @@ def test_reliable_job_startup_on_cpuset(self):
         The released numa node can be used in another job.
         """
 
+        # Check that there are at least a natural vnode + 2 numa vnodes
+        nodeinfo = self.server.status(NODE)
```
vstumpf, line 91: Shouldn't this be VNODE? Or is there no difference

vccardenas: Either NODE or VNODE can be used here.

PR #1880, opened 2020-07-03
implicit commit with queuejob for simple jobs

subhasisb:
#### Describe Bug or Feature

In case of job submissions with a executable or credentials, we still require qsub to exchange the commit message with the server. In a heavily loaded server, cutting this single message can improve submission (and overall server) performance. The same set of message exchange happens between the server and the mom

#### Describe Your Change

Check that the job does not have a script or credential buffer to pass to server. If yes, then add a special marker to the extend parameter of the QUEUEJOB batch request. The server code assumed an implicit commit when this option is noticed. This also ensures that older clients still function their usual way, and the server does not fuse the commit for older clients. The reduction in job queue messages also takes effect between the server and the mom.

Profile: regression
|--|--|--|--|--|--|--|
|2421|1794|1|0|4|0|1789|

The 4 timedout and 1 failed are part of unstable tests that are in the process of being taken out

[svr_logs.txt](https://github.com/openpbs/openpbs/files/4883092/svr_logs.txt)
[mom_logs.txt](https://github.com/openpbs/openpbs/files/4883095/mom_logs.txt)
[valgrind_svr.txt](https://github.com/openpbs/openpbs/files/4883097/valgrind_svr.txt)
[valgrind_svr.txt](https://github.com/openpbs/openpbs/files/4883098/valgrind_svr.txt)


src/server/svr_movejob.c
```diff
@@ -603,6 +592,7 @@ send_job_exec(job *jobp, pbs_net_t hostaddr, int port, struct batch_request *req
 	if (PBSD_commit(stream, job_id, PROT_TPP, &dup_msgid) != 0)
 		goto send_err;
 
+done:
```
hirenvadalia, line 595: I think we still need to transfer the previous run's out/err/chkpt file in case of rerun + express commit, right?
(aka above line no. 584 to 590)

src/lib/Libifl/pbsD_submit.c
```diff
@@ -201,10 +200,24 @@ __pbs_submit(int c, struct attropl  *attrib, char *script, char *destination, ch
 	for (pal = attrib; pal; pal = pal->next)
 		pal->op = SET;		/* force operator to SET */
 
+	cred_info = (struct cred_info *) ptr->th_cred_info;
+
+	if ((!script || (*script == '\0')) && (!cred_info || (cred_info->cred_len <= 0))) {
+		/* no cred and no script, let's request implicit commit to cut one message exchange */
+		if (!(extend = pbs_strcat(&extend, NULL, EXTEND_OPT_IMPLICIT_COMMIT))) {
+			if (set_conn_errtxt(c, "Failed to allocated memory") != 0)
```
arungrover, line 208: Please reword it to "Failed to allocate memory"

src/lib/Libifl/pbsD_submit.c
```diff
@@ -201,10 +200,24 @@ __pbs_submit(int c, struct attropl  *attrib, char *script, char *destination, ch
 	for (pal = attrib; pal; pal = pal->next)
 		pal->op = SET;		/* force operator to SET */
 
+	cred_info = (struct cred_info *) ptr->th_cred_info;
+
+	if ((!script || (*script == '\0')) && (!cred_info || (cred_info->cred_len <= 0))) {
+		/* no cred and no script, let's request implicit commit to cut one message exchange */
+		if (!(extend = pbs_strcat(&extend, NULL, EXTEND_OPT_IMPLICIT_COMMIT))) {
```
arungrover, line 207: I think you missed freeing extend? or is it intentional because this is a client call?

src/server/req_quejob.c
```diff
@@ -1707,36 +1710,29 @@ req_mvjobfile(struct batch_request *preq)
...
  *
- *  @param[in]	preq	-	The batch request structure
+ * @param[in]	preq	-	The batch request structure
+ * @param[in]	pj		-   Pointer to the job structure
  *
  */
-
 void
-req_commit(struct batch_request *preq)
+req_commit_now(struct batch_request *preq,  job *pj)
```
arungrover, line 1718: For jobs without scripts, does it make sense in req_commit_now to even try to save job script to db?

src/lib/Libifl/pbsD_submit.c
subhasisb: i will free it

src/server/req_quejob.c
subhasisb: Yes makes sense - i always thought i had a check before writing the job script - turns out i don't - adding now

src/server/svr_movejob.c
subhasisb: please check the updated code

subhasisb:
Profile: regression
|--|--|--|--|--|--|--|
|2421|1794|1|0|4|0|1789|


The 4 timedout and 1 failed are part of unstable tests that are in the process of being taken out

subhasisb:
@arungrover  pushed changes, please check further

src/server/svr_movejob.c
hirenvadalia: @subhasisb I understand send_job correctly then we still need to call move_job_file() in case of a rerun, for example: if the job without a script and cred is re-running and has stderr and stdout files in spool dir on the server node, but both files size is <2MB, then send job will come to this func, and in this func we only check for `((jobp->ji_qs.ji_svrflags & JOB_SVFLG_SCRIPT) == 0) && (credlen <= 0)` to decide whether to do implicit commit or not. And since we are not checking `jobp->ji_qs.ji_svrflags & JOB_SVFLG_HASRUN` it will go for implicit commit, which will miss transferring previous run's stderr and stdout files

subhasisb: Pushed a change with a check for hAS_RUN, please check

src/lib/Libifl/pbsD_submit.c
```diff
@@ -196,19 +195,23 @@ __pbs_submit(int c, struct attropl  *attrib, char *script, char *destination, ch
...
-			if (set_conn_errtxt(c, "Failed to allocated memory") != 0)
-				pbs_errno = PBSE_SYSTEM;
-			goto error;
-		}
+		if (extend) {
+			lextend = EXTEND_OPT_IMPLICIT_COMMIT;
+			if (!pbs_strcat(&lextend, NULL, extend)) {
+				if (set_conn_errtxt(c, "Failed to allocate memory") != 0)
+					pbs_errno = PBSE_SYSTEM;
+				goto error;
```
arungrover, line 210: I think it will crash in this case because lextend is a string literal.

subhasisb: Hi @arungrover i think it will not crash and i did test it under valgrind and without valgrind as well. I think what happens is that lextend points to a literal to start with. However, inside the pbs_strcat() we call realloc(lextend) right? Realloc merely allocated a new memory location and does a memmove internally to copy over the stuff to the new location. This does not change the address of the literal at all.

PR #1885, opened 2020-07-07
test fails because cput isn't accruing on cpuset system with cgroups enabled

lisa-altair:
#### Describe Bug or Feature

The test is testing whether a job is killed because it's cput goes above the limit set.  The job should be killed once it is noticed that the job's cput goes above 5sec.  However, on a machine where cgroups is enabled, then we rely on the cgroups periodic hook to provide the updated cput.

#### Describe Your Change

I cause the cgroups exechost_periodic hook to run more frequently so the cput can be updated by the cgroups hook.

[cput_before.txt](https://github.com/openpbs/openpbs/files/4886724/cput_before.txt)
[cput_after.txt](https://github.com/openpbs/openpbs/files/4886725/cput_after.txt)

[cput_after2.txt](https://github.com/openpbs/openpbs/files/4887077/cput_after2.txt)
[cput_after3.txt](https://github.com/openpbs/openpbs/files/4887397/cput_after3.txt)

### DIFF
--- a/test/tests/functional/pbs_soft_walltime.py
+++ b/test/tests/functional/pbs_soft_walltime.py
@@ -950,7 +950,8 @@ done
         self.server.alterjob(jid, {'Resource_List.soft_walltime': 300})
         self.server.expect(JOB, {'job_state': 'R'}, id=jid)
 
-        # verify that job is deleted when cput limit is reached
+        self.logger.info("Sleep 10 secs waiting for cput to cause the"
+            " job to be deleted")
         time.sleep(10)
         self.server.expect(JOB, 'queue', op=UNSET, id=jid)
 
--- a/test/tests/functional/pbs_soft_walltime.py
+++ b/test/tests/functional/pbs_soft_walltime.py
@@ -928,6 +928,16 @@ do
     dd if=/dev/zero of=/dev/null;
 done
 """
+	# If it is a cpuset mom, the cgroups hook relies on the periodic hook
+        # to update cput, so make the periodic hook run more often for the
+        # purpose of this test.
+        if self.mom.is_cpuset_mom():
+            attrs = {'freq':1}            
+            rv = self.server.manager(MGR_CMD_SET, HOOK, attrs, "pbs_cgroups")
+            self.assertEqual(rv,0)
+            # cause the change to take effect now
+            self.mom.restart()
+
         j1 = Job(TEST_USER, {'Resource_List.cput': 5})
         j1.create_script(body=script)
         jid = self.server.submit(j1)
@@ -935,8 +945,15 @@ done
         self.server.expect(JOB, {'job_state': 'R'}, id=jid)
 
         # verify that job is deleted when cput limit is reached
+        self.logger.info("Sleep for 10 waiting for cput to cause job to be deleted")
         time.sleep(10)
         self.server.expect(JOB, 'queue', op=UNSET, id=jid)
+
+        if self.mom.is_cpuset_mom():
+            # reset the freq value
+            attrs = {'freq':120}            
+            rv = self.server.manager(MGR_CMD_SET, HOOK, attrs, "pbs_cgroups")
+            self.assertEqual(rv,0)
 
     def test_soft_walltime_resv(self):
         """
--- a/test/tests/functional/pbs_soft_walltime.py
+++ b/test/tests/functional/pbs_soft_walltime.py
@@ -951,7 +951,7 @@ done
         self.server.expect(JOB, {'job_state': 'R'}, id=jid)
 
         self.logger.info("Sleep 10 secs waiting for cput to cause the"
-            " job to be deleted")
+                         " job to be deleted")
         time.sleep(10)
         self.server.expect(JOB, 'queue', op=UNSET, id=jid)
 
--- a/test/tests/functional/pbs_soft_walltime.py
+++ b/test/tests/functional/pbs_soft_walltime.py
@@ -79,7 +79,7 @@ class TestSoftWalltime(TestFunctional):
         if self.mom.is_cpuset_mom():
             # reset the freq value
             attrs = {'freq': 120}
-            rv = self.server.manager(MGR_CMD_SET, HOOK, attrs, "pbs_cgroups")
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_soft_walltime.py
```diff
@@ -928,16 +928,33 @@ def test_soft_less_cput(self):
     dd if=/dev/zero of=/dev/null;
 done
 """
+	# If it is a cpuset mom, the cgroups hook relies on the periodic hook
+        # to update cput, so make the periodic hook run more often for the
+        # purpose of this test.
+        if self.mom.is_cpuset_mom():
+            attrs = {'freq':1}            
+            rv = self.server.manager(MGR_CMD_SET, HOOK, attrs, "pbs_cgroups")
+            self.assertEqual(rv,0)
```
bhroam, line 937: No need  to set rv or check rv.  Manager will raise an  exception if it fails

test/tests/functional/pbs_soft_walltime.py
```diff
@@ -928,16 +928,33 @@ def test_soft_less_cput(self):
...
+            # cause the change to take effect now
+            self.mom.restart()
```
bhroam, line 939: Why do you need to restart the mom to change the freq?  You set it on the server and the server sends  it to the mom.  Restarting  the mom shouldn't affect this.

test/tests/functional/pbs_soft_walltime.py
```diff
@@ -928,16 +928,33 @@ def test_soft_less_cput(self):
...
         # verify that job is deleted when cput limit is reached
+        self.logger.info("Sleep for 10 waiting for cput to cause job to be deleted")
         time.sleep(10)
         self.server.expect(JOB, 'queue', op=UNSET, id=jid)
 
+        if self.mom.is_cpuset_mom():
+            # reset the freq value
+            attrs = {'freq':120}            
+            rv = self.server.manager(MGR_CMD_SET, HOOK, attrs, "pbs_cgroups")
+            self.assertEqual(rv,0)
```
bhroam, line 956: This should be set in tearDown() or setUp() in revert_to_defaults().  If you put  it here, it won't be reverted  if the test fails.

bhroam:
@lisa-altair travis failed due to pycodestyle.

test/tests/functional/pbs_soft_walltime.py
bayucan: @bhroam : Unfortunately, the way the exechost_periodic hook has been implemented is that there'll be a delay when there's already one periodic hook queued up to execute at time X where X = <time previous hook executed and completed> + old 'freq' value. Only after that existing periodic hook executes, will a new periodic hook instance be queued to execute at a new 'freq' value. With pbs_cgroups hook having freq=120 default value, setting freq=1 won't actually take effect until that previous hook has awaken and completed execution. So setting 'freq=1' could actually get delayed up to the previous 'freq' value of 120 seconds + some leway time when mom is busy doing other tasks.
So the workaround is to either sleep 120+ seconds or just restart pbs_mom and the new 'freq' value immediately would take effect.  We decided it's best to just restart. 
Perhaps we should also file a ticket to change this behavior of exechost_periodic hook so that setting 'freq' value should be more instantaneous, on the mom side.

bhroam: Ahh, thanks for the explanation.  This makes sense.   Do we need to wait for the new freq value to be sent from the server before we restart?  Or will that be taken care of in the mom<->server handshake on startup?

lisa-altair: The new freq value is sent right away (as can be seen in the mom log), so no waiting is required.

test/tests/functional/pbs_soft_walltime.py
lisa-altair: Okay, I added a tearDown to the test suite.

lisa-altair:
I have made the requested changes and fixed my python style.  Please have another look

test/tests/functional/pbs_soft_walltime.py
```diff
@@ -945,16 +951,9 @@ def test_soft_less_cput(self):
         self.server.expect(JOB, {'job_state': 'R'}, id=jid)
 
         # verify that job is deleted when cput limit is reached
-        self.logger.info("Sleep for 10 waiting for cput to cause job to be deleted")
```
bhroam: Why the removal of the log message?  It's kind to your test runner to tell them why they are sleeping.

test/tests/functional/pbs_soft_walltime.py
```diff
@@ -928,6 +935,15 @@ def test_soft_less_cput(self):
     dd if=/dev/zero of=/dev/null;
 done
 """
+        # If it is a cpuset mom, the cgroups hook relies on the periodic hook
+        # to update cput, so make the periodic hook run more often for the
+        # purpose of this test.
+        if self.mom.is_cpuset_mom():
+            attrs = {'freq': 1}
+            rv = self.server.manager(MGR_CMD_SET, HOOK, attrs, "pbs_cgroups")
```
bayucan, line 943: You can rid of the unused 'rv'.

test/tests/functional/pbs_soft_walltime.py
lisa-altair: It was my debug statement.

bhroam: Could you put it back?  I don't view these as debug statements.  I view them as telling your test runner what is going on.  You  could even merge it into the comment right above it since they both really serve the same purpose.

lisa-altair: Sure!  I can add the log message back. Can you clarify what you mean by "merge it into the comment right above it since they both really serve the same purpose"?  I don't understand.

bhroam: You have a code comment to tell the person reading the code what you are sleeping for, and then you have a log message which is telling the test runner what you are sleeping for.  If you express them both in the log message, you can remove the code comment.

PR #1886, opened 2020-07-08
TestQrun.test_qrun_hangs does not clean up child process

saimanikant:
#### Describe Bug or Feature

In test_qrun_hangs:
We fork and and don't exit in child process. So both child and parent process invoke Teardown.
This results in 2 test results being generated. One of them errors out in test_qrun_subjob .

#### Describe Your Change

Exit in child process

[TestQrun_fail.txt](https://github.com/openpbs/openpbs/files/4888727/TestQrun_fail.txt)
[TestQrun_pass.txt](https://github.com/openpbs/openpbs/files/4888728/TestQrun_pass.txt)

### DIFF
--- a/test/tests/functional/pbs_qrun.py
+++ b/test/tests/functional/pbs_qrun.py
@@ -129,7 +129,7 @@ class TestQrun(TestFunctional):
             try:
                 self.server.runjob(jobid=jid)
                 self.logger.info("Successfully runjob. Child process exit.")
-                os._exit(5)
+                os._exit(0)
             except PbsRunError as e:
                 self.logger.info("Runjob throws error: " + e.msg[0])
         else:
--- a/test/tests/functional/pbs_qrun.py
+++ b/test/tests/functional/pbs_qrun.py
@@ -129,6 +129,7 @@ class TestQrun(TestFunctional):
             try:
                 self.server.runjob(jobid=jid)
                 self.logger.info("Successfully runjob. Child process exit.")
+                os._exit(5)
             except PbsRunError as e:
                 self.logger.info("Runjob throws error: " + e.msg[0])
         else:


saimanikant:
latest logs:
[TestQrun_pass.txt](https://github.com/openpbs/openpbs/files/4888983/TestQrun_pass.txt)

test/tests/functional/pbs_qrun.py
```diff
@@ -129,6 +129,7 @@ def test_qrun_hangs(self):
             try:
                 self.server.runjob(jobid=jid)
                 self.logger.info("Successfully runjob. Child process exit.")
+                os._exit(0)
```
agrawalravi90, line 132: Just say 'return' ? \_exit sounds like an internal function since it starts with an underscore

saimanikant: I had tried return but that does not terminate the current process right. With return, I still see 2 test complete results being generated.

agrawalravi90: ah yes, return will cause it to go through the same post run functions that the normal process goes through. I still don't like the idea of using an internal function because that can get renamed one fine day. How about just doing "sys.exit(0)" ?

hirenvadalia: @agrawalravi90 `os._exit(0)` is not an internal method. it is exposed a valid method that anyone can use it. And python doc itself say child should use `os._exit(0)`. See https://docs.python.org/3.8/library/os.html#os._exit

prakashcv13: I do not think this is something internal. It is a documented API and it is not easy to change it's name.
Also, from the [documentation](https://docs.python.org/3/library/os.html#os._exit)
_exit() should normally only be used in the child process after a fork().

agrawalravi90: Ok, weird that they underscore it then, thanks for the pointer

PR #1888, opened 2020-07-09
Fix accounting log match and race conditions in TestPbsReliableJobStartup

shilpakodli:
#### Describe Bug or Feature
Few of the tests of TestPbsReliableJobStartup are failing due to below reasons:
1. Accounting Log match
2. Race condition while searching log in mom logs
3.  Incorrcet Job state

Test was searching for accounting logs with specific number of lines which internally runs tail on specified accounting logs file resulting in command execution failure on our machines.
In few tests, we are checking for "Job;<jobid>;task.+started, hostname" log in mom logs. This message is appearing 4 minutes after hook execution. This is because job script has 3 "Fib 37" series to run after which "pbsdsh -n 0 -- hostname -s" is called.
**Job Script:**
echo 'pbsdsh -n 1 fib 37'
pbsdsh -n 1 – fib 37
echo 'pbsdsh -n 2 fib 37'
pbsdsh -n 2 – fib 37
echo 'fib 37'
fib 37
echo 'HOSTNAME TESTS'
echo 'pbsdsh -n 0 hostname'
pbsdsh -n 0 --  hostname -s   

#### Describe Your Change
- Replaced number of lines in accounting log match with "ALL" where it will search in entire file.
- Increased max attempts and interval in mom log match
- Changed Substate of Job from 41-> 42

TH3_Reliablejob_test_results

### DIFF
--- a/test/tests/functional/pbs_reliable_job_startup.py
+++ b/test/tests/functional/pbs_reliable_job_startup.py
@@ -151,41 +151,42 @@ class TestPbsReliableJobStartup(TestFunctional):
         """
 
         if atype == 'e':
-            self.mom.log_match("Job;%s;Obit sent" % (jid,), n=100,
-                               max_attempts=5, interval=5)
+            self.mom.log_match("Job;%s;Obit sent" % (jid,), n="ALL",
+                               max_attempts=5, interval=5,
+                               starttime=self.stime)
 
         self.server.accounting_match(
             msg=".*%s;%s.*exec_host=%s" % (atype, jid, exec_host),
-            regexp=True, n=20, max_attempts=3)
+            regexp=True, n="ALL", max_attempts=3, starttime=self.stime)
 
         self.server.accounting_match(
             msg=".*%s;%s.*exec_vnode=%s" % (atype, jid, exec_vnode),
-            regexp=True, n=20, max_attempts=3)
+            regexp=True, n="ALL", max_attempts=3, starttime=self.stime)
 
         self.server.accounting_match(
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_reliable_job_startup.py
```diff
@@ -1797,12 +1799,11 @@ def test_t2(self):
...
                                   self.job1_place,
                                   self.job1_sel_esc)
-
         self.momA.log_match("Job;%s;task.+started, hostname" % (jid,),
-                            n=10, max_attempts=60, interval=2, regexp=True)
+                            n=10, max_attempts=60, interval=5, regexp=True)
 
         self.momA.log_match("Job;%s;copy file request received" % (jid,),
-                            n=10, max_attempts=10, interval=2)
+                            n=10, max_attempts=60, interval=5)
```
dinesh1306, line 1806: remove max_attempts if it is 60.

dinesh1306:
please fix pep8 issue.

test/tests/functional/pbs_reliable_job_startup.py
bayucan: Maybe just remove max_attempts altogether, and just let the default happen, which I think is now 60. This default might actually be increased to 100 to handle slower systems or systems running pbs_cgroups hook (cpuset system)...

bhroam:
Travis failed due  to pycodestyle.  It's the missing space I pointed out and a couple of blank lines with whitespace.

PR #1889, opened 2020-07-09
Fixed accounting log issue in Test_singleNode_Job_ResourceUsed

Klovely:
#### Describe Bug or Feature
Fixed accounting log issue in Test_singleNode_Job_ResourceUsed

#### Describe Your Change
Test was searching for accounting logs with specific number of lines which internally runs tail on specified accounting logs file resulting in command execution failure on testbed machines.

* Replaced number of lines in accounting log match with "ALL" where it will search in entire file.

### DIFF
--- a/test/tests/functional/pbs_resc_used_single_node.py
+++ b/test/tests/functional/pbs_resc_used_single_node.py
@@ -130,6 +130,7 @@ e.vnode_list[localnode].resources_available['foo_str2'] = "seven"
         Test accumulation of resources of a single node job from an
         exechost_epilogue hook.
         """
+        self.stime = time.time()
         self.logger.info("test_epilogue_single_node")
         hook_body = """
 import pbs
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_resc_used_single_node.py
```diff
@@ -190,44 +191,53 @@ def test_epilogue_single_node(self):
 
         acctlog_match = 'resources_used.foo_f=0.09'
         s = self.server.accounting_match(
-            "E;%s;.*%s.*" % (jid, acctlog_match), regexp=True, n=100)
+            "E;%s;.*%s.*" % (jid, acctlog_match), regexp=True, n="ALL",
```
hirenvadalia, line 194: I think you don't need `n="ALL"` when you pass `starttime`, with `starttime` it should work (no matter what internally it does tail/head/cat). With only `starttime` if it is not working then its fw bug

Also, you don't need to pass `starttime` as it's by default to test start time, in other words, to fix this test, you simply need to remove `n=100`, and after that if still failing then its fw bug

hirenvadalia:
@Klovely Are you still working on this? Can you please mark it as WIP or close if you are not working

hirenvadalia:
closing inactive PR.

PR #1890, opened 2020-07-10
TestPbsAccumulateRescUsed doesn't show accumulation of resources if r…

lisa-altair:
…emote mom is cpusetmom

#### Describe Bug or Feature

The problem with the tests is that the tests look for a specific value of cput.  On cpuset machines where the cgroups hook is enabled cput is something the cgroups hook polls for and sets.  Thus anything set in a hook that runs before the cgroups hook  will be overwritten by the cgroups hook.

#### Describe Your Change

To fix this I ordered the tests epilogue hook to run after the cgroups hook.
And for the other tests, I changed the test to not compare cput on cpuset systems.  
This shouldn't be a problem since the test wasn't testing the accumulation of cput, but rather testing that it and other values are set.

[accum_before.txt](https://github.com/openpbs/openpbs/files/4900595/accum_before.txt)
[accum_resc_after.txt](https://github.com/openpbs/openpbs/files/4900596/accum_resc_after.txt)

### DIFF
--- a/test/tests/functional/pbs_accumulate_resc_used.py
+++ b/test/tests/functional/pbs_accumulate_resc_used.py
@@ -335,10 +335,10 @@ else:
         will be enabled and will overwrite the cput value set in the prologue
         hook
         """
-        has_cpuset = 0
+        has_cpuset = False
         for mom in self.moms.values():
             if mom.is_cpuset_mom():
-                has_cpuset = 1
+                has_cpuset = True
 
         self.logger.info("test_prologue")
         hook_body = """
@@ -414,7 +414,7 @@ else:
             'resources_used.vmem': '35gb',
             'resources_used.ncpus': '3'}
 
-        if has_cpuset is 0:
+        if not has_cpuset:
             a['resources_used.cput'] = '00:00:35'
 
         self.server.expect(JOB, a, extend='x', offset=10,
@@ -471,7 +471,7 @@ else:
         self.server.accounting_match(
             "E;%s;.*%s.*" % (jid, acctlog_match), regexp=True, n=100)
 
-        if has_cpuset is 0:
+        if not has_cpuset:
             acctlog_match = 'resources_used.cput=00:00:35'
             self.server.accounting_match(
                 "E;%s;.*%s.*" % (jid, acctlog_match), regexp=True, n=100)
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_accumulate_resc_used.py
```diff
@@ -397,16 +405,20 @@ def test_prologue(self):
...
             'resources_used.foo_i': '35',
             'resources_used.foo_str4': "eight",
             'resources_used.stra': "\"glad,elated\",\"happy\"",
             'resources_used.vmem': '35gb',
-            'resources_used.cput': '00:00:35',
-            'resources_used.ncpus': '3'},
-            extend='x', offset=10, attrop=PTL_AND, id=jid)
+            'resources_used.ncpus': '3'}
+
+        if has_cpuset is 0:
```
bhroam, line 417: Say if not  has_cpuset: (here and other places)

test/tests/functional/pbs_accumulate_resc_used.py
```diff
@@ -331,7 +331,15 @@ def test_prologue(self):
...
         Test accumulatinon of resources of a multinode job from an
         exechost_prologue hook.
+        On cpuset systems don't check for cput because the pbs_cgroups hook
+        will be enabled and will overwrite the cput value set in the prologue
+        hook
         """
+        has_cpuset = 0
+        for mom in self.moms.values():
+            if mom.is_cpuset_mom():
+                has_cpuset = 1
```
bhroam, line 341: I'd use True and False instead of 1 and 0.  It's more pythony

test/tests/functional/pbs_accumulate_resc_used.py
```diff
@@ -1180,13 +1208,15 @@ def test_epi_pro(self):
...
-            'resources_used.cput': '30',
-            'job_state': 'F'}, attrop=PTL_AND,
-            extend='x', id=jid, offset=5,
-            max_attempts=60, interval=1)
+            'job_state': 'F'}
+
+        if has_cpuset is 0:
+            b['resources_used.cput'] = '30'
+        self.server.expect(JOB, b, attrop=PTL_AND, extend='x', id=jid,
+                           offset=5, max_attempts=60, interval=1)
```
bhroam, line 1219: You can drop the PTL_AND.  It is the default now.  Also max_attempts=60 is default as well.

lisa-altair: Okay, this was there before, I just moved the lines up.  But I can remove these now that they are the default.

test/tests/functional/pbs_accumulate_resc_used.py
lisa-altair: hahaha.  I originally had that, then the grammar of it bugged me, so I changed it.  I'll change it back.

PR #1892, opened 2020-07-12
make find_attr, find_resc_def use index search

subhasisb:
#### Describe Bug or Feature

find_attr() and find_resc_def() are used all through the pbs code base. However they both employ a serial search. This negatively impacts pbs performance. For example, the job object currently has almost 100 attributes to search from, and find_attr() is frequently called for jobs. 

#### Describe Your Change

Use an avltree based search index to make the search much faster. Since avltrees have O(log N) time complexity, a search for 100 attributes should take about 2 string comparisons.

Along with this have streamlines the attributes save performance as well

Profile: regression
|--|--|--|--|--|--|--|
|2529|1794|2|0|0|0|1792|

valgrind logs: 
[valgrind_current.log](https://github.com/openpbs/openpbs/files/4918718/valgrind_current.log)

The 2 failures are unstable test cases


src/resmom/mom_main.c
```diff
@@ -9148,6 +9156,9 @@ main(int argc, char *argv[])
 	CLEAR_HEAD(task_list_timed);
 	CLEAR_HEAD(task_list_event);
 
+	cr_attrdef_idx(job_attr_def, JOB_ATR_LAST);
```
hirenvadalia, line 9159: error check?

src/server/attr_recov_db.c
```diff
@@ -113,7 +106,7 @@ compare_obj_hash(void *qs, int len, void *oldhash)
  *
  */
 int 
-encode_single_attr_db(struct attribute_def *padef, struct attribute *pattr, pbs_db_attr_list_t *db_attr_list)
+encode_single_attr_db(struct attribute_def *padef, struct attribute *pattr, int index, pbs_db_attr_list_t *db_attr_list)
```
hirenvadalia, line 109: Looks like we don't need `index` arg anymore?

src/server/setup_resc.c
```diff
@@ -367,11 +367,17 @@ expand_resc_array(char *rname, int rtype, int rflag)
...
 	pnew->rs_free   = p_resc_type_map->rtm_free;
 	pnew->rs_action = NULL_FUNC_RESC;
+	pnew->rs_custom = 1; /*  built-in resources are loaded from XML defn and initialized to 0 */
 	pnew->rs_flags = rflag;
 	pnew->rs_type  = rtype;
 	pnew->rs_entlimflg = 0;
 	pnew->rs_next  = NULL;
 
+	if (pbs_idx_insert(resc_attrdef_idx, pnew->rs_name, pnew) != PBS_IDX_RET_OK) {
+		free(pnew);
```
hirenvadalia, line 377: need to free pnew->rs_name

src/lib/Libattr/attr_fn_resc.c
```diff
@@ -453,72 +454,61 @@ free_resc(attribute *pattr)
...
- * @retval 	-1 	on error
+ * @return	pointer to structure
+ * @retval	pointer to resource_def structure - Success
+ * @retval	NULL - Error
  *
  */
-int
-is_builtin(resource_def *rscdef)
+resource_def *
+find_resc_def(resource_def *resc_def, char *name, int limit)
```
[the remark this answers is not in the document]
subhasisb: yes we don't need, but there is also no benefit of removing it. I left it in case we ever need this in future. In the case of find_attr() i removed it since i had to change the definition anyway (to include the idx parameter), however find_resc_def() defn does not require any change. I can remove it, of course, if i find some time

src/lib/Libtpp/tpp_router.c
```diff
@@ -155,7 +155,7 @@ alloc_router(char *name, tpp_addr_t *address)
 	}
 
 	/* initialize the routers leaf tree */
-	r->my_leaves_idx = pbs_idx_create(PBS_IDX_DUPS_NOT_OK, sizeof(tpp_addr_t));
+	r->my_leaves_idx = pbs_idx_create(0, sizeof(tpp_addr_t));
```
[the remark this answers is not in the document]
subhasisb: No - now PBS_IDX_DUPS_OK is a bit wise flag, and 0 is default value for such flags meaning no bits are set - it is actually more explicit than adding another macro for that

src/server/attr_recov_db.c
subhasisb: yeah i had it to save the index in the database as well, however in our updated multiserver approach since we will not be loading from the database that often - that optimization does not make sense, so i will remove this

src/lib/Libattr/attr_fn_resc.c
```diff
@@ -453,72 +454,61 @@ free_resc(attribute *pattr)
...
-		rscdf = rscdf->rs_next;
+	/* create the attribute index */
+	if ((resc_attrdef_idx = pbs_idx_create(PBS_IDX_CASE_CMP, 0)) == NULL)
+		return -1;
+
+	/* add all attributes to the tree with key as the attr name */
+	for (i = 0; i < limit; i++) {
+		if (strcmp(resc_def->rs_name, RESC_NOOP_DEF) != 0) {
+			if (pbs_idx_insert(resc_attrdef_idx, resc_def->rs_name, resc_def) != PBS_IDX_RET_OK)
+				return -1;
```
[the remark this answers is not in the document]
subhasisb: i don't see these in valgrind definitely lost reports, so why add to valgrind.supp?

src/lib/Libattr/attr_fn_resc.c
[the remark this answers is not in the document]
subhasisb: Update: found time and removed it :)

src/resmom/mom_main.c
subhasisb: oops looks like i had it coded twice, removed the older one :)

subhasisb:
pushed changes, please check further

src/include/resource.h
```diff
@@ -104,6 +106,7 @@ typedef struct resource_def {
 	unsigned int rs_type : ATRDTYPE;  /* type of resource,see attribute.h */
 	unsigned int rs_entlimflg;	  /* tracking entity limits for this  */
 	struct resource_def *rs_next;
+	unsigned int rs_custom; /* bit flag to indicate custom resource or builtin */
```
prakashcv13, line 109: shall we align the comment.

src/lib/Libattr/attr_func.c
```diff
@@ -113,34 +90,68 @@ clear_attr(attribute *pattr, struct attribute_def *pdef)
...
+		return NULL;
+
+	/* create the attribute index */
+	if ((attrdef_idx = pbs_idx_create(PBS_IDX_CASE_CMP, 0)) == NULL)
+		return NULL;
+
+	/* add all attributes to the tree with key as the attr name */
+	for (i = 0; i < limit; i++) {
+		if (pbs_idx_insert(attrdef_idx, adef->at_name, adef) != PBS_IDX_RET_OK) {
+			/* destroy this index */
```
prakashcv13, line 119: we are not destroying the index.

src/lib/Libdb/db_postgres_attr.c
```diff
@@ -237,37 +239,28 @@ dbarray_to_attrlist(char *raw_array, pbs_db_attr_list_t *attr_list)
...
-		if (keys_only)
-			attr_val_len = 0;
-		else
-			attr_val_len = (pal->al_atopl.value == NULL ? 0 : strlen(pal->al_atopl.value));
-		len += sizeof(int32_t) + 3 + attr_val_len + 1; /* include space for dot */
-		count++;
+	char *p;
+	int spc_avl, spc_req;
+	/* len attr_name . attr_resc len attr_flags . attr_value */ 
+	static int fixed_part_req = sizeof(int32_t)*2 + PBS_MAXATTRNAME + PBS_MAXATTRRESC + 3 /* max 3 digits flags */ + 2 /* 2 dots */ + 1 /* null terminator */;
```
prakashcv13, line 256: space around '*' is needed.
Also, having comments within the statement is not much readable.

src/lib/Libdb/db_postgres_attr.c
```diff
@@ -281,21 +274,38 @@ attrlist_to_dbarray_ex(char **raw_array, pbs_db_attr_list_t *attr_list, int keys
 	val = (struct str_data *)((char *) array + sizeof(struct pg_array));
 
 	for (pal = (svrattrl *)GET_NEXT(attr_list->attrs); pal != NULL; pal = (svrattrl *)GET_NEXT(pal->al_link)) {
-		sprintf(val->str, "%s.%s", pal->al_atopl.name, ((pal->al_atopl.resource == NULL) ? "" : pal->al_atopl.resource));
-		val->len = htonl(strlen(val->str));
-
-		val = (struct str_data *)(val->str + ntohl(val->len)); /* point to end */
+		spc_avl =  len - ((void *) val - (void *) array);
```
prakashcv13, line 277: shall we typecast to (char *) instead of (void *).
arithmetic on void * is allowed only as an extension in gcc.
https://gcc.gnu.org/onlinedocs/gcc-4.5.0/gcc/Pointer-Arith.html#Pointer-Arith

src/include/libutil.h
```diff
@@ -196,6 +196,14 @@ enum vnode_sharing str_to_vnode_sharing(char *vn_str);
  */
 char *pbs_strcat(char **strbuf, int *ssize, char *str);
 
+/*
+ * like strcpy, but returns pointer to end of copied data
+ * useful for chain copies instead of sprintf which is very
+ * slow
+ * 
+ */
+char *pbs_strcpy(char *dest, const char *src);
```
prakashcv13, line 205: would calling it pbs_strcat() make more sense.

src/lib/Libutil/avltree.c
```diff
@@ -219,8 +219,17 @@ freenode(node *n)
 static int
 compkey(rectype *r1, rectype *r2)
 {
-	int n = ix_keylength ? memcmp(r1->key, r2->key, ix_keylength) : strcmp(r1->key, r2->key);
-	if (n || ix_dupkeys == AVL_NO_DUP_KEYS)
+	int n;
+	if (ix_keylength)
+		n = memcmp(r1->key, r2->key, ix_keylength);
+	else {
```
prakashcv13, line 225: we can remove the braces. upto you.

src/lib/Libutil/misc_utils.c
```diff
@@ -306,6 +306,35 @@ pbs_strcat(char **strbuf, int *ssize, char *str)
...
+ *        added to the returned pointer.  
+ *
+ * @param[in] dest - pointer to the destination buffer
+ * @param[in] src  - pointer to the source buffer
+ *
+ * @return char *
+ * @retval pointer to the end of the resulting string
+ *	
+ * @note: Caller needs to ensure space and non-NULL pointers
+ *        This function is created for performance so does
```
prakashcv13, line 324: does **not**

src/server/pbsd_init.c
```diff
@@ -360,6 +343,42 @@ pbsd_init(int type)
...
+	}
+	if (cr_rescdef_idx(svr_resc_def, svr_resc_size) != 0) {
+		log_err(errno, __func__, "Failed creating resc definition search index");
+		return (-1);
+	}
+
+	/* initialize the pointers in the resource_def array */
+
+	for (i = 0; i < (svr_resc_size - 1); ++i)
+		svr_resc_def[i].rs_next = &svr_resc_def[i+1];
```
prakashcv13, line 379: spaces around '+'

src/server/pbsd_main.c
```diff
@@ -299,6 +269,14 @@ int svr_unsent_qrun_req = 0;	/* Set to 1 for scheduling unsent qrun requests */
 void *jobs_idx = NULL;
 void *queues_idx = NULL;
 void *resvs_idx = NULL;
+
+void *job_attr_idx = NULL;
+void *resv_attr_idx = NULL;
+void *node_attr_idx = NULL;
+void *que_attr_idx = NULL;
+void *svr_attr_idx = NULL;
+void *sched_attr_idx = NULL;
```
prakashcv13, line 278: no need to initialise - they are global variables.

subhasisb:
Profile: regression
|--|--|--|--|--|--|--|
|2548|1792|0|0|0|0|1792|

subhasisb:
latest tests

src/server/pbsd_main.c
hirenvadalia: curious to know: does it matter? anyway, it will same asm code right? no?

src/include/pbs_idx.h
```diff
@@ -43,8 +43,8 @@
 extern "C" {
 #endif
 
-#define PBS_IDX_DUPS_NOT_OK 0 /* duplicate key not allowed in index */
-#define PBS_IDX_DUPS_OK     1 /* duplicate key allowed in index */
+#define PBS_IDX_DUPS_OK     0x01 /* duplicate key allowed in index */
+#define PBS_IDX_CASE_CMP    0x02 /* set case-insensitive compare */
```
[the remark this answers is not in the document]
subhasisb: actually the linux case insensitive strcmp is called strcasecmp (no i in the name) :) 
But i do agree, ICASE is probably more intuitive, changing

src/include/resource.h
subhasisb: these are all "aligned" by the "clang-formatter"  - the guideline is that are no longer required (actually were never required) to align variables into a column

src/server/pbsd_main.c
prakashcv13: I believe in not doing things that will be automatically done for you...
it is an optional comment

src/lib/Libattr/attr_func.c
subhasisb: yeah i will remove this comment - we would quit PBS at this point, so no point destroying the index

src/lib/Libdb/db_postgres_attr.c
subhasisb: will put space, but not sure how to put comment to match each piece, let me try

src/include/libutil.h
subhasisb: that would be confusing since strcat actually walks the dest string to its end, and this "expects" the end of the dest char to be provided. The only difference with actually strcpy is the return value. Besides, we already have a pbs_strcat()

src/lib/Libutil/avltree.c
subhasisb: i would keep it - makes more readable i feel

src/server/pbsd_main.c
prakashcv13: another advantage is - uninitialized global data gets stored in the .bss section.
The nice part about this is, the .bss section data doesn't have to be included in the ELF file on disk (ie. there isn't a whole region of zeros in the file for the .bss section). Instead, the loader knows from the section headers how much to allocate for the .bss section, and simply zero it out before handing control over to your program.

subhasisb:
Description: Tests from given sources on platforms CENTOS7, UBUNTU1804, SUSE15, UBUNTU1604
Platforms: CENTOS7,UBUNTU1804,SUSE15,UBUNTU1604
Profile: regression
|--|--|--|--|--|--|--|
|2550|3508|0|0|0|0|3508|

PR #1896, opened 2020-07-14
Fix ralter -D not changing end time and doubling resources

vstumpf:
### Describe Bug or Feature
If pbs_ralter -D would change the end time to a later time, the reservation would still end at the original time.
Also, if pbs_ralter -D would change the end time, the server would double the resources used on the nodes.

#### Describe Your Change
When we calculate the new end time, add end time to the modified flags.
When generating a Time4resv task, instead of checking if we are not altering only the End time, we check if we are altering the start time.

[after-fix3.txt](https://github.com/openpbs/openpbs/files/4921057/after-fix3.txt)
[beforefix_small.txt](https://github.com/openpbs/openpbs/files/4921058/beforefix_small.txt)

### DIFF
--- a/src/server/svr_jobfunc.c
+++ b/src/server/svr_jobfunc.c
@@ -4221,6 +4221,10 @@ start_end_dur_wall(resc_resv *presv)
 				(petime->at_val.at_long <= pstime->at_val.at_long))
 				rc = -1;
 			else {
+				if (pstate == RESV_BEING_ALTERED) {
+					presv->ri_alter.ra_flags |= RESV_DURATION_MODIFIED;
+					presv->ri_alter.ra_duration = pduration->at_val.at_long;
+				}
 				atemp.at_val.at_long = (petime->at_val.at_long -
 					pstime->at_val.at_long);
 				pddef->at_set(pduration, &atemp, SET);
@@ -4234,6 +4238,10 @@ start_end_dur_wall(resc_resv *presv)
 				(pduration->at_val.at_long <= 0))
 				rc = -1;
 			else {
+				if (pstate == RESV_BEING_ALTERED) {
+					presv->ri_alter.ra_flags |= RESV_END_TIME_MODIFIED;
+					presv->ri_alter.ra_etime = petime->at_val.at_long;
+				}
 				petime->at_flags |= ATR_SET_MOD_MCACHE;
 				petime->at_val.at_long = pstime->at_val.at_long +
 					pduration->at_val.at_long;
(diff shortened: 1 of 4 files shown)


test/tests/functional/pbs_ralter.py
```diff
@@ -2234,3 +2235,32 @@ def test_select_fail_revert(self):
...
+        self.server.status(RESV, 'resv_nodes', id=rid)
+        resv_node = self.server.reservations[rid].get_vnodes()[0]
+
+        sleepdur = start - time.time()
+        self.logger.info('Sleeping until reservation starts')
+        self.server.expect(RESV,
+                           {'reserve_state': (MATCH_RE, 'RESV_RUNNING|5')},
+                           offset=sleepdur)
+
+        self.server.alterresv(rid, {ATTR_resv_duration: '10:00'})
```
bhroam, line 2261: I'd use alter_a_reservation().  It checks to make sure the alter succeeded.

src/server/svr_jobfunc.c
```diff
@@ -4306,6 +4327,10 @@ start_end_dur_wall(resc_resv *presv)
 					prsc->rs_value.at_val.at_long))
 				rc = -1;
 			else {
+				if (pstate == RESV_BEING_ALTERED) {
+					presv->ri_alter.ra_flags |= RESV_DURATION_MODIFIED;
+					presv->ri_alter.ra_duration = pduration->at_val.at_long;
+				}
```
bhroam, line 4333: Any time you change duration, you need to change walltime as well.

vstumpf: we don't store walltime in the resv_alter struct, it's only for when we need to revert

test/tests/functional/pbs_ralter.py
vstumpf: That function checks to see if the reservation is either confirmed or running. since the reservation is deleted, it fails.
I can add a log_match if the reservation alter was confirmed

arungrover, line 2261: can we reduce the duration instead of increasing it? we can then wait to check that the reservation does not exist when the duration is reduced.

vstumpf: Oops, I was looking at the wrong test. I will take another look.

vstumpf: im sorry arun, i don't really understand
the length of the duration doesn't really matter, it just needs to change. we don't ahve to wait for the duration.

arungrover: Sorry about being unclear. What I mean to say was that the bug description states - "_If pbs_ralter -D would change the end time to a later time, the reservation would still end at the original time._" So would it not be useful to have a test that captures the modified reservation actually ends at the modified time.

The reason I suggested reducing the duration was that I thought it does not matter if duration is extended or reduced, in both the cases (because of the bug) the reservation would end at the original end time.

vstumpf: There is a test that decreases the duration, then checks to see if the reservation is ended, so I think that's proof enough for the reservation's duration being reduced.
There is a test that increases the duration, I make sure that the test is still running after the original end time has passed. That should be enough proof for the reservation's duration being increased.

This last test is for the resources, to make sure they're not being double counted.

arungrover: ah ok, if it already exists then fine.
Thanks!

PR #1897, opened 2020-07-14
except() should first trigger a scheduling cycle and then verify the value

tpai26:
#### Describe Bug or Feature
Updating expect() such that it first triggers a scheduling cycle and then looks for the value. Currently in PTL we look for an attribute value and then trigger a scheduling cycle if the expected value is not found. This can cause an error if the attribute itself is not present. For example if a job is in Q state, then expect(exec_host=x) will fail since exec_host is not present yet.  Therefore we must trigger a scheduling cycle first in expect() and then look for the value

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -8330,6 +8330,10 @@ class Server(PBSService):
                             str(offset))
             time.sleep(offset)
 
+        if trigger_sched_cycle and attempt == 0:
+            self.manager(MGR_CMD_SET,
+                         SERVER, {'scheduling': 'True'}, sudo=True)
+
         if attrib is None:
             attrib = {}
 
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -8330,10 +8330,6 @@ class Server(PBSService):
                             str(offset))
             time.sleep(offset)
 
-        if trigger_sched_cycle and attempt == 0:
-            self.manager(MGR_CMD_SET,
-                         SERVER, {'scheduling': 'True'}, sudo=True)
-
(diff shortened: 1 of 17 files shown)


test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -8330,6 +8330,10 @@ def expect(self, obj_type, attrib=None, id=None, op=EQ, attrop=PTL_AND,
                             str(offset))
             time.sleep(offset)
 
+        if trigger_sched_cycle and attempt == 0:
+            self.manager(MGR_CMD_SET,
```
hirenvadalia, line 8334: I think this is wrong, 4 reasons:
1. If the test has turned off scheduling, this will enable it as default `trigger_sched_cycle` is True
2. As per the current design of expect() sched cycle should be triggered only when someone is doing expect(JOB, ...) not for any other objects (like NODE, VNODE, etc...)
3. As per the current design of expect(), sched cycle is triggered via action func (aka ExpectAction()), but here we are adding it directly to expect() which conflicts with ExpectAction()'s functionality
4. With this change, now every expect has to take care of passing the correct value of `trigger_sched_cycle` else it can give you unexpected results (Actually the original problem is the introduction of `trigger_sched_cycle` only, expect() was never designed to handle such thing, all those kinda stuff should be handled via ExpectAction())

hirenvadalia: Just had chat with @visheshh for the use case of introduction of `trigget_sched_cycle`, it was added so that test can disable triggering sched cycle, but as per ExpectAction() correct way of disabling expect action and stopping trigger of sched cycle is using `self.server.actions.disable_action(None, 'kicksched')` and in tearDown re-enable it using `self.server.actions.enable_action(None, 'kicksched')`

PR #1900, opened 2020-07-14
Running job in a resv causes future resvs to go unconfirmed

arungrover:
#### Describe Bug or Feature
Scheduler failed to confirm future reservations when it sees the node having a running job inside a running reservation.

#### Describe Your Change
The issue was that when the ghost job issue was addressed, scheduler started considering all running jobs to assign them to nodes. In reality, it only needs to see non-reservation running jobs to assign them to nodes because all reservation jobs resources are already accounted for by the reservation.

#### Link to Design Doc
NA

Existing tests - test_resv_excl_with_jobs & test_future_resv_with_non_excl_jobs were failing.
[test.txt](https://github.com/openpbs/openpbs/files/4921341/test.txt)

### DIFF
--- a/src/scheduler/server_info.c
+++ b/src/scheduler/server_info.c
@@ -2079,8 +2079,8 @@ check_job_running(resource_resv *job, void *arg)
  * @param[in]	arg	-	argument (not used here)
  *
  * @return	int
- * @retval	1	: if job is running and in reservation passed in arg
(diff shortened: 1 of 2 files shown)


arungrover:
Thanks for the sign-off. 
Since we have other similar filter function declaration in server_info.h, I'd rather retain it.

vstumpf:
I'm going to merge it, since the only thing that changed was a function header.

PR #1901, opened 2020-07-15
Increased default max_attempts to 180

visheshh:
#### Describe Bug or Feature
On cpuset machines 60 attempts is not sufficient in some cases

#### Describe Your Change
It has been observed that sometimes it takes longer for a test job to reach a particular state when it is run on a cgroups enabled system (cpuset system) due to the many hook events involved with the cgroups hook.  To help the tests actually pass increasing max_attempts to 180.

The change is made to default value of MAX_ATTEMPTS in ptl configuration options.

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -249,7 +249,7 @@ class PtlConfig(object):
             'PTL_SUDO_CMD': 'sudo -H',
             'PTL_RSH_CMD': 'ssh',
             'PTL_CP_CMD': 'scp -p',
-            'PTL_MAX_ATTEMPTS': 100,
+            'PTL_MAX_ATTEMPTS': 180,
             'PTL_ATTEMPT_INTERVAL': 0.5,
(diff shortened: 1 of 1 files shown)


test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -249,7 +249,7 @@ def __init__(self, conf=None):
             'PTL_SUDO_CMD': 'sudo -H',
             'PTL_RSH_CMD': 'ssh',
             'PTL_CP_CMD': 'scp -p',
-            'PTL_MAX_ATTEMPTS': 60,
+            'PTL_MAX_ATTEMPTS': 100,
```
hirenvadalia, line 252: I suggest this to increase to 180, that's what we use in our internal tool

prakashcv13:
@visheshh - a better PR summary would also help. Right now it is not clear where are we changing the max_attempts.

visheshh:
@hirenvadalia Changes done.

hirenvadalia:
@visheshh looks like Travis is stuck in reporting status back to github. Can you please do force push to re-trigger it?

PR #1903, opened 2020-07-15
scheduler might leak node spec in corner cases

arungrover:
#### Describe Bug or Feature
Scheduler might leak node spec while trying to run a job

#### Describe Your Change
This leak can only happen in error cases.
I've also addressed set_tpp_config freeing partial data in case of error. It is anyways expected that upon failure of set_tpp_config function the daemon needs to exit. So it does not help to free partial data. Added these possible leaks to the suppression file as well.

#### Link to Design Doc
NA

Since all of these leaks are only possible in corner cases, I couldn't test them out.
#### Regression results
Profile: regression
|--|--|--|--|--|--|--|
|2606|1794|0|0|0|0|1794|


arungrover:
Thanks @vstumpf!
One more reviewer needed please...

PR #1906, opened 2020-07-16
save job/resv/node state before server quits gracefully

subhasisb:
#### Describe Bug or Feature

Job's resource_assigned.mem is not as expected after a server restart with qterm -t quick

#### Describe Your Change

The pbs_server needs to save any job updates before gracefully quitting. This was mistakenly removed in PR #1847 
Added the code back

Profile: noderampdown
|--|--|--|--|--|--|--|
|2599|2|0|0|0|0|2|

### DIFF
--- a/src/server/pbsd_main.c
+++ b/src/server/pbsd_main.c
@@ -701,36 +701,36 @@ int
 main(int argc, char **argv)
 {
 	char *nodename = NULL;
-	int			are_primary;
-	int			c, rc;
-	int			i;
-	int			tppfd;		/* fd to receive is HELLO's */
-	struct			tpp_config tpp_conf;
-	char			lockfile[MAXPATHLEN+1];
-	char			**origevp;
-	char			*pc;
-	pbs_queue		*pque;
-	char			*servicename;
-	time_t			svrlivetime;
-	int			sock;
-	struct stat 		sb_sa;
-	struct batch_request	*periodic_req;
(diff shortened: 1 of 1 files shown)


src/server/pbsd_main.c
```diff
@@ -1784,6 +1779,16 @@ main(int argc, char **argv)
...
 
+	/* save any jobs that need saving */
+	for (pjob = (job *)GET_NEXT(svr_alljobs); pjob; pjob = (job *)GET_NEXT(pjob->ji_alljobs))
+		job_save_db(pjob);
+
+	/* save any reservations that need saving */
+	for (presv = (resc_resv *)GET_NEXT(svr_allresvs); presv; presv = (resc_resv *)GET_NEXT(presv->ri_allresvs))
+		resv_save_db(presv);
+
+	save_nodes_db(0, NULL);
```
bhroam, line 1790: Do we need to save any other object like a queue or server?

subhasisb: This is basically a old code block before my attribute refactoring change. I had missed to pull in this block of code. The old code only saved these objects, so i just brought back these. Since it has survived so long without saving anything else, i guess we have set expectations here.

bhroam:
@subhasisb you have a conflict now.  Fix that and I'll merge the PR.

subhasisb:
@bhroam i rebased and pushed, had to do a force

PR #1907, opened 2020-07-17
removed system update in container

kjakkali:
#### Describe Bug or Feature
Updating build script to build and test PBS with vanilla OS. 

#### Describe Your Change
Updated build script

No need to extra testing results. Travis results will suffice


hirenvadalia:
@kjakkali Looks like you have conflict now, please resolve it.

PR #1908, opened 2020-07-17
PTL supports running tests multiple times

birbalsain:
#### Describe Bug or Feature
Test Framework should allow testers to run a test for multiple times. Tester can use delay between two repetition.

#### Describe Your Change
Introduces new pbs_benchpress options --repeat_tests=<int> for how many times test will run and
--repeat_tests_delay for delay=<seconds> delay between repeated test.

Each repetition includes all submitted tests. next repetition starts after finishing of all tests of first repetition.
If submits three tests then in first repetition all three tests run and after finishing all three tests, next repetition starts and again all three tests run.

At the end we get message " All tests are repeated times"

#### Link to Design Doc

https://pbspro.atlassian.net/wiki/spaces/PD/pages/1386938446/Test+Framework+should+allow+testers+to+run+a+test+for+multiple+times

[repetition_mixed_testsuite.txt](https://github.com/openpbs/openpbs/files/4947277/repetition_mixed_testsuite.txt)
[repetition_multiple_tests.txt](https://github.com/openpbs/openpbs/files/4947278/repetition_multiple_tests.txt)
[no_repetition.txt](https://github.com/openpbs/openpbs/files/4947306/no_repetition.txt)
[json_report.txt](https://github.com/openpbs/openpbs/files/4947305/json_report.txt)

### DIFF
--- a/test/fw/bin/pbs_benchpress
+++ b/test/fw/bin/pbs_benchpress
@@ -145,8 +145,8 @@ def usage():
             'fails\n']
     msg += ['--timeout=<seconds>: duration after which no test suites are '
             'run\n']
-    msg += ['--repeat-tests=<int>: number of times test is going to repeat\n']
-    msg += ['--repeat-tests-delay=<seconds>: delay between two repetition\n']
+    msg += ['--repeat-count=<count>: repeat given all tests <count> times\n']
+    msg += ['--repeat-delay=<seconds>: delay between two repetition\n']
     msg += ['--follow-child: if set, walk the test hierarchy and run ' +
             'each test\n']
     msg += ['--tags=<tag>: Select only tests that have <tag> tag.']
@@ -216,8 +216,8 @@ if __name__ == '__main__':
     lcov_baseurl = None
     genhtml_bin = None
     timeout = None
-    repeat_tests = 1
-    repeat_tests_delay = 0
+    repeat_count = 1
+    repeat_delay = 0
     nosedebug = False
     only_info = False
     tags = []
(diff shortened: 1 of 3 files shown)


test/fw/bin/pbs_benchpress
```diff
@@ -145,6 +145,8 @@ def usage():
             'fails\n']
     msg += ['--timeout=<seconds>: duration after which no test suites are '
             'run\n']
+    msg += ['--repeat-tests=<int>: number of times test is going to repeat\n']
```
hirenvadalia, line 148: How about `--repeat-count=<count>: repeat given all tests <count> times\n`?

test/fw/bin/pbs_benchpress
```diff
@@ -145,6 +145,8 @@ def usage():
...
+    msg += ['--repeat-tests-delay=<seconds>: delay between two repetition\n']
```
hirenvadalia, line 149: I think `--repeat-delay` will be a better name

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -490,7 +493,17 @@ def run(self, test):
         self.result = result = self._makeResult()
         self.result.start = datetime.datetime.now()
         try:
-            test(result)
+            for i in range(self.repetition):
+                self.logger.info("====================================")
+                self.logger.info("tests are running (" + str(i + 1) + ") time")
```
hirenvadalia, line 498: `self.logger.info("Running tests %d time" % i + 1)`

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -490,7 +493,17 @@ def run(self, test):
...
+                self.logger.info("====================================")
+                if i is not 0:
```
hirenvadalia, line 500: I suggest not to use `is not` (py3.8 gives a warning for it), also you can do `if i != 0:` here

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -490,7 +493,17 @@ def run(self, test):
...
+                    time.sleep(self.repetition_delay)
+                test(result)
+            self.logger.info("=============================================")
+            self.logger.info(" All tests are repeated " +
```
hirenvadalia, line 504: `self.logger.info("All Tests are repeated %d times" % self.repeat_count)`

Also shouldn't we print this message only if repeat count is > 0?

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -466,9 +467,11 @@ class PtlTextTestRunner(TextTestRunner):
     """
 
     def __init__(self, stream=sys.stdout, descriptions=True, verbosity=3,
-                 config=None):
+                 config=None, repetition=1, repetition_delay=0):
         self.logger = logging.getLogger(__name__)
         self.result = None
+        self.repetition = repetition
```
hirenvadalia, line 473: how about `repeat_count` as a variable name? and for below `repeat_delay`

hirenvadalia:
Also, PR title and commit-msg can better like `Add support for running tests multiple time in PTL`

test/fw/ptl/utils/plugins/ptl_test_runner.py
bhroam: Actually repeat_count = 1 by default, so it should be > 1.

test/fw/ptl/utils/plugins/ptl_test_runner.py
bhroam: I don't think you need both this message and the below one.  We only need to be told once that the tests were repeated N times.

hirenvadalia: Agree with @bhroam.

birbalsain:
Thanks @hirenvadalia @bhroam , I have done the changes.

PR #1909, opened 2020-07-17
reconfirming running reservations can move available nodes

bhroam:
#### Describe Bug or Feature
When a running degraded reservation is reconfirmed, the scheduler  can replace all nodes, not just the unavailable ones.   This means the node a running job is on can be replaced.  This leaves the reservation with a job on a node that it doesn't own.

#### Describe Your Change
The main reason this happened is because when we were only reconfirming non-running reservations, we would mark the reservation as unconfirmed.  To the scheduler, a degraded reservation and an unconfirmed reservation were the same.  When we started reconfirming  running  reservations, this changed.  The problem is when we were choosing which select spec to use to confirm the reservation, we didn't detect the reservation was running.   We used the original select spec and treated the reservation like it was when it was not running.  This meant we could confirm it anywhere.  

I changed the code to leave  the reservation in the running state and have is_ok_to_run() detect it as confirmable in the degraded substate.  This meant check_nodes() sees it as running and we use the execselect.

[resv.log](https://github.com/openpbs/openpbs/files/4940516/resv.log)

### DIFF
--- a/src/scheduler/resource.c
+++ b/src/scheduler/resource.c
@@ -872,12 +872,12 @@ collect_resources_from_requests(resource_resv **resresv_arr)
 		 * execselect: select created from the exec_vnode.  This is likely to
 		 * be a subset of resources from schedselect + 'vnode'.  It is possible
 		 * that a job was run with qrun -H (res=val) where res is not part of
-		 * the schedselect.  The exec_vnode is taken directly from the -H  argument
+		 * the schedselect.  The exec_vnode is taken directly from the -H argument
 		 * The qrun -H case is why we need to do this check.
 		 */
 		if (r->execselect != NULL && r->execselect->defs != NULL) {
-			if ((r->is_job && r->job != NULL && in_runnable_state(r)) ||
-				(r->is_resv && r->resv != NULL && r->resv->resv_state == RESV_BEING_ALTERED)) {
+			if ((r->job != NULL && in_runnable_state(r)) ||
+				(r->resv != NULL && (r->resv->resv_state == RESV_BEING_ALTERED || r->resv->resv_substate == RESV_DEGRADED))) {
 				for (j = 0; r->execselect->defs[j] != NULL;j++) {
 					if (!resdef_exists_in_array(defarr, r->execselect->defs[j]))
 						add_resdef_to_array(&defarr, r->execselect->defs[j]);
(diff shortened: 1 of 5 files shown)


src/scheduler/resource.c
```diff
@@ -872,12 +872,12 @@ collect_resources_from_requests(resource_resv **resresv_arr)
...
 		 * that a job was run with qrun -H (res=val) where res is not part of
-		 * the schedselect.  The exec_vnode is taken directly from the -H  argument
+		 * the schedselect.  The exec_vnode is taken directly from the -H argument
 		 * The qrun -H case is why we need to do this check.
 		 */
 		if (r->execselect != NULL && r->execselect->defs != NULL) {
-			if ((r->is_job && r->job != NULL && in_runnable_state(r)) ||
-				(r->is_resv && r->resv != NULL && r->resv->resv_state == RESV_BEING_ALTERED)) {
+			if ((r->job != NULL && in_runnable_state(r)) ||
+				(r->resv != NULL && (r->resv->resv_state == RESV_BEING_ALTERED || r->resv->resv_substate == RESV_DEGRADED))) {
```
vstumpf, line 880: do you not need is_job or is_resv?

test/tests/functional/pbs_reservations.py
```diff
@@ -1598,6 +1654,9 @@ def test_ASAP_resv_request_same_time(self):
         time doesn't crashes PBS daemon.
         """
         self.common_steps()
+        # turn opt_backfill_fuzzy so both reservations are at the same time
+        self.server.manager(MGR_CMD_SET, SCHED,
+                            {'opt_backfill_fuzzy': 'off'}, id='default')
```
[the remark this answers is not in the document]
bhroam: opt_backfill_fuzzy is unset  by default.  When it is unset, it is low.  I need to turn it completely off.

src/scheduler/resource.c
bhroam: Yeah, I don't necessary need it, but unless we refactor is_job and is_resv out out of the code, I guess they should be used.  I'll remove them if you want.

bhroam:
@vstumpf I addressed the typo.

test/tests/functional/pbs_reservations.py
```diff
@@ -376,6 +376,62 @@ def test_degraded_advanced_reservation_superchunk(self):
         self.assertNotEqual(nds1, nds2)
         self.assertEquals(sc, nds1.split('+')[0])
 
+    def test_degraded_running_only_replace(self):
```
vstumpf, line 379: I think this needs to skip on cpuset machines

PR #1912, opened 2020-07-20
Server crash on make install

ashwathraop:
#### Describe Bug or Feature

Libdb makefile was using cp to make a copy of the db shared library. When pbs server is running if 'make install' is done, it will corrupt the address space where libdb was loaded since with cp only directory entry (the name) is removed.

Also fixed a configure warning:
configure.ac:75: warning: LT_INIT was called before AM_PROG_AR
/usr/share/aclocal-1.16/ar-lib.m4:13: AM_PROG_AR is expanded from...
configure.ac:75: the top level

#### Describe Your Change

One way to solve this to do a rm of the file before doing cp. This will mark the libdb.so as deleted in /proc/server_pid/maps. So when we remove and copy, we basically create a "different" file.  Libtool uses install command to copy files. I replaced cp with install now and this makes sure the file is removed before copy.

For configure warning I had to move AM_PROG_LIBTOOL after AM_PROG_AR.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -68,7 +68,6 @@ AC_PROG_CC
 AC_PROG_LEX
 AC_PROG_INSTALL
 AC_PROG_LN_S
-AM_PROG_LIBTOOL
 
 # Automake macros
 #AM_PROG_AR macro is defined with automake version >= 1.12
@@ -76,6 +75,7 @@ m4_ifdef([AM_PROG_AR], [AM_PROG_AR])
 AM_PROG_CC_C_O
 
 # Initialize libtool
+AM_PROG_LIBTOOL
 LT_INIT([shared static])
 
 # Checks for libraries.
--- a/src/lib/Libdb/Makefile.am
+++ b/src/lib/Libdb/Makefile.am
@@ -43,7 +43,7 @@ SUBDIRS = \
 	pgsql
 
 all-local:
-	install -D $(top_builddir)/src/lib/Libdb/pgsql/.libs/libdb.so.0.0.0 $(top_builddir)/src/lib/Libdb/libdb.so
+	cp -p $(top_builddir)/src/lib/Libdb/pgsql/.libs/libdb.so.0.0.0 $(top_builddir)/src/lib/Libdb/libdb.so
 
 clean-local:
 	rm -f $(top_builddir)/src/lib/Libdb/libdb.so
--- a/src/lib/Libdb/Makefile.am
+++ b/src/lib/Libdb/Makefile.am
@@ -43,7 +43,7 @@ SUBDIRS = \
 	pgsql
 
 all-local:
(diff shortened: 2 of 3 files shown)


arungrover:
well according to this page - https://superuser.com/questions/229883/how-is-install-c-different-from-cp It seems that install already removes the destination file before replacing it. Is cp the right way to go?

agrawalravi90:
"One significant difference is that cp truncates the destination file and starts copying data from the source into the destination file.

install, on the other hand, removes the destination file first.

This is significant because if the destination file is already in use, bad things could happen to whomever is using that file in case you cp a new file on top of it. e.g. overwriting an executable that is running might fail. Truncating a data file that an existing process is busy reading/writing to could cause pretty weird behavior. If you just remove the destination file first, as install does, things continue much like normal - the removed file isn't actually removed until all processes close that file."

This does make it sound like install is the way to go .. Ashwath had shared another article with us which said that apparently processes don't load a .so fully, so they have to access the copy on the disk from time to time. So, if a cp happens while the copy on the disk is being accessed, it could cause the crash that Ashwath saw.

arungrover:
That make it sound like even if we do cp instead of install it might still fail. Then maybe we should stop PBS in make install before copying over files.

ashwathraop:
@arungrover Yes, ideally we should always stop the server and other components before doing make install. This crash most likely will be seen only by developers who may not want to stop every time they are making quick changes and end up having core files inside server_priv directory. This might be annoying since every restart you'll be notified of core files.

If you look at make install logs, you'll see libtool mostly uses install to copy files to the destination. 

This is the page Ravi was talking about: https://stackoverflow.com/questions/7767325/replacing-shared-object-so-file-while-main-program-is-running

ashwathraop:
Reverted to cp in one of the makefile since it was copying inside build directory itself and rpm tools complain usage of build paths.

Error was:
+ /usr/lib/rpm/check-buildroot
Binary file /root/rpmbuild/BUILDROOT/openpbs-20.0.0-0.x86_64/opt/pbs/sbin/pbs_ds_password.bin matches
Binary file /root/rpmbuild/BUILDROOT/openpbs-20.0.0-0.x86_64/opt/pbs/sbin/pbs_dataservice.bin matches
Binary file /root/rpmbuild/BUILDROOT/openpbs-20.0.0-0.x86_64/opt/pbs/sbin/pbs_server.bin matches
Binary file /root/rpmbuild/BUILDROOT/openpbs-20.0.0-0.x86_64/opt/pbs/sbin/pbs_ds_monitor matches
Binary file /root/rpmbuild/BUILDROOT/openpbs-20.0.0-0.x86_64/opt/pbs/bin/printjob_svr.bin matches
Found '/root/rpmbuild/BUILDROOT/openpbs-20.0.0-0.x86_64' in installed files; aborting
error: Bad exit status from /var/tmp/rpm-tmp.OGD3Xp (%install)
    Bad exit status from /var/tmp/rpm-tmp.OGD3Xp (%install)

PR #1914, opened 2020-07-21
optimizing free_nodes() function in server.

bremanandjk:
#### Describe Bug or Feature

On every invocation of free_nodes(), especially each job exit, it is consuming more time for releasing the nodes associated with the job object.

#### Describe Your Change

Optimized in a way that instead of looping through to "each subnode" in "each node" in "each mom", identified the exact node directly from chunks and releasing nodes using deallocate_job_from_node() function.

Executed the below steps to identify the time consumption in each invocation. 

a) Install pbs
b) kill running mom kill -9 $(pidof pbs_mom)
c) run pbs_mom -m as root to start mock mom
d) pbs_config --vnodify -N 10 -a resources_available.ncpus=500 (to create 10 vnodes with 500 ncpus each)
e) turn off scheduling qmgr -c "s s scheduling=0"
f) submit 50k jobs (for i in $(seq 1 1 5000); do qsub -koe -- /bin/date; done; as pbsroot user)
g) start only one sched cycle (qmgr -c "s s scheduling=1" && qmgr -c "s s scheduling=0")
h) wait for jobs to finish

Before fix:
Before_fix_2_master

After fix:
After_fix_2_master

Profile: regression
|--|--|--|--|--|--|--|
|2708|1794|2|0|0|0|1792|

### DIFF
--- a/src/server/node_manager.c
+++ b/src/server/node_manager.c
@@ -6839,52 +6839,21 @@ set_nodes(void *pobj, int objtype, char *execvnod_in, char **execvnod_out, char 
 	return 0;
 }
 
-/**
- * @brief
- * 		free nodes allocated to a job
- *
- * @param[in,out]	pjob	- job structure
- *
- * @return	void
- */
 void
-free_nodes(job *pjob)
+disassociate_job_from_moms(job *pjob, struct pbsnode *pnode)
 {
-
-	struct pbsnode *pnode;
-	mom_svrinfo_t *psvrmom;
-	char *execvnod_in = NULL;
 	int i;
 	int j;
-	char *execvncopy;
-	char *chunk;
-	char *last;
-	int hasprn;
-	char *vname;
-	int nelem;
-	struct key_value_pair *pkvp;
-	char *execvnod = NULL;
 	int special_case = 0;
+	mom_svrinfo_t *psvrmom;
 	
-	/* decrement number of jobs on the Mom who is the first Mom */	
-	/* for the job, Mother Superior; incremented in set_nodes() */	
-	/* and saved in ji_destin in assign_hosts()		    */	
-	if (((pjob->ji_qs.ji_svrflags & JOB_SVFLG_HasNodes) != 0) &&	
-		(pjob->ji_qs.ji_destin[0] != '\0')) {	
-		pnode = find_nodebyname(pjob->ji_qs.ji_destin);	
-		if (pnode) {	
-			psvrmom = pnode->nd_moms[0]->mi_data;	
-			if (--psvrmom->msr_numjobs < 0)	
-				psvrmom->msr_numjobs = 0;	
-		}	
-	}
-	
-	/* Now loop through the Moms and remove the jobindx entry */	
-	/* and remove this jobs's jobinfo entry from each vnode   */	
-	for (i = 0; i < mominfo_array_size; i++) {	
-		if (mominfo_array[i] == NULL)	
+	if (pnode == NULL)
+		return;
+			
+	for (i = 0; i < pnode->nd_nummoms; i++) {	
+		if (pnode->nd_moms[i] == NULL)	
 			continue;	
-		psvrmom = (mom_svrinfo_t *)(mominfo_array[i]->mi_data);	
+		psvrmom = (mom_svrinfo_t *)(pnode->nd_moms[i]->mi_data);	
 
 		for (j=0; j<psvrmom->msr_jbinxsz; j++) {	
 			if (psvrmom->msr_jobindx[j] == pjob) {	
@@ -6911,7 +6880,51 @@ free_nodes(job *pjob)
 			log_event(PBSEVENT_DEBUG, PBS_EVENTCLASS_JOB, LOG_INFO, pjob->ji_qs.ji_jobid, log_buffer);	
 		}
 	}
+	
+}
 
+
+/**
+ * @brief
+ * 		free nodes allocated to a job
+ *
+ * @param[in,out]	pjob	- job structure
+ *
+ * @return	void
+ */
+void
+free_nodes(job *pjob)
+{
+
+	struct pbsnode *pnode;
+	mom_svrinfo_t *psvrmom;
+	char *execvnod_in = NULL;
+	char *execvncopy;
+	char *chunk;
+	char *last;
+	int hasprn;
+	char *vname;
+	int nelem;
+	struct key_value_pair *pkvp;
+	char *execvnod = NULL;
+
+	
+	/* decrement number of jobs on the Mom who is the first Mom */	
+	/* for the job, Mother Superior; incremented in set_nodes() */	
+	/* and saved in ji_destin in assign_hosts()		    */	
+	if (((pjob->ji_qs.ji_svrflags & JOB_SVFLG_HasNodes) != 0) &&	
+		(pjob->ji_qs.ji_destin[0] != '\0')) {	
+		pnode = find_nodebyname(pjob->ji_qs.ji_destin);	
+		if (pnode) {	
+			psvrmom = pnode->nd_moms[0]->mi_data;	
+			if (--psvrmom->msr_numjobs < 0)	
+				psvrmom->msr_numjobs = 0;	
+		}	
+	}
+	
+	/* Now loop through the Moms and remove the jobindx entry */
+	/*  remove this jobs's jobinfo entry from each vnode   */
+	
 	if (pjob->ji_wattr[(int)JOB_ATR_exec_vnode].at_flags & ATR_VFLAG_SET) {
 		execvnod_in = pjob->ji_wattr[(int)JOB_ATR_exec_vnode].at_val.at_str;
 		if (!execvnod_in)
@@ -6938,6 +6951,7 @@ free_nodes(job *pjob)
 	while (chunk) {
 		if (parse_node_resc(chunk, &vname, &nelem, &pkvp) == 0) {
 			pnode = find_nodebyname(vname);
+			disassociate_job_from_moms(pjob, pnode);
 			deallocate_job_from_node(pjob, pnode);
 			chunk = parse_plus_spec_r(last, &last, &hasprn);
 		}
--- a/src/server/node_manager.c
+++ b/src/server/node_manager.c
@@ -6839,8 +6839,8 @@ set_nodes(void *pobj, int objtype, char *execvnod_in, char **execvnod_out, char 
 	return 0;
 }
 
-void
-disassociate_job_from_moms(job *pjob, struct pbsnode *pnode)
+static void
+remove_job_index_from_mom(job *pjob, struct pbsnode *pnode)
 {
 	int i;
 	int j;
@@ -6925,11 +6925,7 @@ free_nodes(job *pjob)
 	/* Now loop through the Moms and remove the jobindx entry */
 	/*  remove this jobs's jobinfo entry from each vnode   */
 	
-	if (pjob->ji_wattr[(int)JOB_ATR_exec_vnode].at_flags & ATR_VFLAG_SET) {
-		execvnod_in = pjob->ji_wattr[(int)JOB_ATR_exec_vnode].at_val.at_str;
-		if (!execvnod_in)
-			return;
-	} else {
+	if ((pjob->ji_wattr[JOB_ATR_exec_vnode].at_flags & ATR_VFLAG_SET) == 0 || ((execvnod_in = pjob->ji_wattr[JOB_ATR_exec_vnode].at_val.at_str) == NULL)) {
 		log_event(PBSEVENT_DEBUG2, PBS_EVENTCLASS_JOB, LOG_DEBUG, pjob->ji_qs.ji_jobid, "in free_nodes and no exec_vnode");
 		return;
 	}
@@ -6950,7 +6946,7 @@ free_nodes(job *pjob)
 	while (chunk) {
 		if (parse_node_resc(chunk, &vname, &nelem, &pkvp) == 0) {
 			pnode = find_nodebyname(vname);
-			disassociate_job_from_moms(pjob, pnode);
+			remove_job_index_from_mom(pjob, pnode);
 			deallocate_job_from_node(pjob, pnode);
 			chunk = parse_plus_spec_r(last, &last, &hasprn);
 		}
--- a/src/server/node_manager.c
+++ b/src/server/node_manager.c
@@ -6850,124 +6850,100 @@ set_nodes(void *pobj, int objtype, char *execvnod_in, char **execvnod_out, char 
 void
 free_nodes(job *pjob)
 {
-	struct	pbssubn	*np;
-	mom_svrinfo_t	*psvrmom;
-	struct  pbsnode *pnode;
-	struct	jobinfo	*jp, *prev, *next;
-	int     i;
-	int     j;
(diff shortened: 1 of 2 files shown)


bremanandjk:
Thanks for your reviews, I will remove the loop which iterates over mominfo structures to disassociate jobs in it. Let me replace those functionalities by a performant code to get a high percentage results.

src/server/node_manager.c
```diff
@@ -6850,124 +6850,100 @@ set_nodes(void *pobj, int objtype, char *execvnod_in, char **execvnod_out, char
...
+	char *vname;
+	int nelem;
+	struct key_value_pair *pkvp;
+	char *execvnod = NULL;
+	int special_case = 0;
+	
+	/* decrement number of jobs on the Mom who is the first Mom */	
+	/* for the job, Mother Superior; incremented in set_nodes() */	
+	/* and saved in ji_destin in assign_hosts()		    */	
+	if (((pjob->ji_qs.ji_svrflags & JOB_SVFLG_HasNodes) != 0) &&
```
nithinj, line 6872: this can be part of deallocate_job_from_node instead of handling separately.

src/server/node_manager.c
```diff
@@ -6850,124 +6850,100 @@ set_nodes(void *pobj, int objtype, char *execvnod_in, char **execvnod_out, char
...
+	/* Now loop through the Moms and remove the jobindx entry */	
+	/* and remove this jobs's jobinfo entry from each vnode   */	
+	for (i = 0; i < mominfo_array_size; i++) {	
+		if (mominfo_array[i] == NULL)	
+			continue;	
+		psvrmom = (mom_svrinfo_t *)(mominfo_array[i]->mi_data);	
+
+		for (j=0; j<psvrmom->msr_jbinxsz; j++) {	
+			if (psvrmom->msr_jobindx[j] == pjob) {	
+				psvrmom->msr_jobindx[j] = NULL;
```
nithinj, line 6891: Find the mom from node struct rather than looping through.

src/server/node_manager.c
bremanandjk: Yes, that is what I am trying now.

bremanandjk:
> I do not see any point of this change due to two things:
> 
> 1. We are still looping over every mom in the system (think of an exascale system with 100k nodes)
> 2. For a multivnoded system, the gain is very miniscule, from 2% to .5% - not seeing much justification. Maybe for a very large multivnoded system there is some ROI, but then not sure how common those are (and what is that typical size)?
> 
> In other words, the enhancement is not buying us enough. I would suggest rework

I have modified the code such that not to parse all the moms in the system. Now we are considering only the moms associated with the node object. Thereby, this change would give a dramatic performance gain. 

Also, I have modified the test to show the substantial results, test steps follows. 

Test setup:

1) Install pbs
2) kill running mom kill -9 $(pidof pbs_mom)
3) run pbs_mom -m as root to start mock mom
4) pbs_config --vnodify -N 100 -a resources_available.ncpus=500 (to create 100 vnodes with 500 ncpus each)
5) Run 50 moms with unique port.
6) turn off scheduling qmgr -c "s s scheduling=0"
7) submit 50k jobs (for i in $(seq 1 1 50000); do qsub -koe -- /bin/date; done; as pbsroot user)
8) start only one sched cycle (qmgr -c "s s scheduling=1" && qmgr -c "s s scheduling=0")
9) wait for jobs to finish
10) Run perf tools to gauge the performance. 

Before fix: 
trial_2_before_fix

After fix: 
trial_2_after_fix

TH3 results:
Screen Shot 2020-07-22 at 12 05 56 AM

P.S: I have maintained the setup I used in the test machine as it is. If there are any issues in my test to set up, please let me know. I would redo the tests. Thanks

src/server/node_manager.c
```diff
@@ -6839,6 +6839,51 @@ set_nodes(void *pobj, int objtype, char *execvnod_in, char **execvnod_out, char
 	return 0;
 }
 
+void
+disassociate_job_from_moms(job *pjob, struct pbsnode *pnode)
```
hirenvadalia, line 6843: I don't think we need this, we can put this code in `deallocate_job_from_node()`

src/server/node_manager.c
```diff
@@ -6839,6 +6839,51 @@ set_nodes(void *pobj, int objtype, char *execvnod_in, char **execvnod_out, char
...
+
+		for (j=0; j<psvrmom->msr_jbinxsz; j++) {	
+			if (psvrmom->msr_jobindx[j] == pjob) {	
+				psvrmom->msr_jobindx[j] = NULL;	
+				if (is_called_by_job_purge)
+					special_case = 1;
+			}	
+		}
+		
+		if (special_case) {
```
hirenvadalia, line 6866: Do we really need this now? I think we can get rid of this

src/server/node_manager.c
```diff
@@ -6839,6 +6839,51 @@ set_nodes(void *pobj, int objtype, char *execvnod_in, char **execvnod_out, char
...
+	
+	if (pnode == NULL)
+		return;
+			
+	for (i = 0; i < pnode->nd_nummoms; i++) {	
+		if (pnode->nd_moms[i] == NULL)	
+			continue;	
+		psvrmom = (mom_svrinfo_t *)(pnode->nd_moms[i]->mi_data);	
+
+		for (j=0; j<psvrmom->msr_jbinxsz; j++) {
```
hirenvadalia, line 6858: I think we should work on this and try to get rid of this loop also, as it is going to all jobs assigned to mom

src/server/node_manager.c
```diff
@@ -6850,124 +6895,69 @@ set_nodes(void *pobj, int objtype, char *execvnod_in, char **execvnod_out, char
...
-		}
+	/*  remove this jobs's jobinfo entry from each vnode   */
+	
+	if (pjob->ji_wattr[(int)JOB_ATR_exec_vnode].at_flags & ATR_VFLAG_SET) {
+		execvnod_in = pjob->ji_wattr[(int)JOB_ATR_exec_vnode].at_val.at_str;
+		if (!execvnod_in)
+			return;
+	} else {
+		log_event(PBSEVENT_DEBUG2, PBS_EVENTCLASS_JOB, LOG_DEBUG,
+		pjob->ji_qs.ji_jobid, "in free_nodes and no exec_vnode");
```
hirenvadalia, line 6934: please merge this with the above line

src/server/node_manager.c
```diff
@@ -6850,124 +6895,69 @@ set_nodes(void *pobj, int objtype, char *execvnod_in, char **execvnod_out, char
...
-			for (j=0; j<psvrmom->msr_jbinxsz; j++)
-				if (psvrmom->msr_jobindx[j] != NULL)
-					log_event(PBSEVENT_DEBUG, PBS_EVENTCLASS_JOB, LOG_INFO, pjob->ji_qs.ji_jobid, psvrmom->msr_jobindx[j]->ji_qs.ji_jobid);
-			sprintf(log_buffer, "===================================================================="
-				"==================================================");
-			log_event(PBSEVENT_DEBUG, PBS_EVENTCLASS_JOB, LOG_INFO, pjob->ji_qs.ji_jobid, log_buffer);
-		}
+	/*  remove this jobs's jobinfo entry from each vnode   */
+	
+	if (pjob->ji_wattr[(int)JOB_ATR_exec_vnode].at_flags & ATR_VFLAG_SET) {
```
hirenvadalia, line 6928: Do we really need this if/else? old code was not checking for this, and I don't think free_nodes() will be called if job is allocated any nodes

src/server/node_manager.c
```diff
@@ -6850,124 +6895,69 @@ set_nodes(void *pobj, int objtype, char *execvnod_in, char **execvnod_out, char
...
-				set_vnode_state(pnode,
-					~(INUSE_JOB|INUSE_JOBEXCL),
-					Nd_State_And);
-
-				/* call function to check and free the node from the prov list
-				 and reset wait_prov flag, if set */
-				if (pjob->ji_qs.ji_substate == JOB_SUBSTATE_PROVISION)
-					free_prov_vnode(pnode);
-			}
+	chunk = parse_plus_spec_r(execvncopy, &last, &hasprn);
```
agrawalravi90, line 6949: I'm a bit concerned that parse_plus_spec can be slow for long exec_vnodes, can you please also do a performance comparison with multi-node jobs which occupy 100+ nodes and post results before and after your changes?

src/server/node_manager.c
bremanandjk: In general, its good to remove the loop. But here we need to consider this case in detail. 
Whenever a job queued in MoM, the job object would be registered in mominfo structure and the same would be removed after completion. For short jobs, it is so quick.  Even if we try with AVL tree DS to store job id versus mominfo index, the operation would be more complex in adding/removing/ frequently. The gain we get on from that would be trivial. (We are already getting 93% to 0.5 %).  I feel those operations are effective in terms of a large number of elements like tens of hundreds of thousands. 

This in case, at any moment, the number of runtime jobs is close to a few hundred, I would prefer the current code is much better in comparison to having another DS. If you think otherwise, plz explain with a technical explanation, I will consider in another PR.

src/server/node_manager.c
bremanandjk: This portion of code was already present for some other bug debugging, If the developer who added this code felt that it solved their purpose, let them remove it. That's why I preserved this change as it is. It is out of the scope of job disassociation from node and mom PR.

src/server/node_manager.c
bremanandjk: It is a good practice to check the attribute flag and retrieve it. I am seeing this practice everywhere. please refer same file, you could see many references.

src/server/node_manager.c
bremanandjk: Agree, I have addressed it.

src/server/node_manager.c
bremanandjk: This parse_plus_spec_r() is used extensively in all many places in Server, Sched & MoM. While optimizing this function in general, this code also would get optimized. 

Let me try a multi-node job test also and post the results. Thanks for your review.

bremanandjk: @agrawalravi90 I tried sampling server with 5000 number of multi-node jobs with each job requesting 100 chunks of 10 cpus. 

The main issue is observed only in single node job submission. It seems fine in multi-node jobs. 

Before fix: 
Before_fix_multinode

After fix: 

After_fix_multinode

agrawalravi90: A few questions:
- How many nodes did you have? how many ncpus on each node?
- I'm a bit confused why req_modifyjob() is taking so much time, especially in the first graph, did you not have enough cpus for ALL jobs to run?
- Can you please highlight free_nodes() in the graphs? You can search for a symbol in perf report just like you do in vim

bremanandjk: * How many nodes did you have? how many ncpus on each node?
 In my setup, I have 500 vnodes and each node is having 100 cpus each. In total 50k cpus. 
* I'm a bit confused why req_modifyjob() is taking so much time, especially in the first graph, did you not have enough cpus for ALL jobs to run?
 The requested amount cpus are available in the test system
* Can you please highlight free_nodes() in the graphs? You can search for a symbol in perf report just like you do in vim
 I am not able to locate the free_nodes() while running in multi-node tests. I have tried various test setups, Its happening mostly in single node tests. Even searched in vim too.

bremanandjk: Please ignore the previous results:

I have tried the setup suggested by @agrawalravi90  (Thanks) , In this setup, I could get the free_nodes symbol perfectly. 

Setup details: 

1) 50000 vnodes, each with 10 cpus. 
2) Submitted 5k jobs, each requesting 10 chunks of 10 ncpus,

Scanned the server in perf till the job completion with one scheduling cycle. 

Before fix:
master_multinode_50knodes

After fix: 
master_after_fix_50knodes

src/server/node_manager.c
hirenvadalia: Thanks, @bremanandjk for the explanation. I got it that we can't put this in `deallocate_job_from_node()`. But I would suggest to make it static and name it like `remove_job_index_from_mom()` (as we have `add_job_index_in_mom()` in the same file, which is exactly reverse of this). Also please add function header.

src/server/node_manager.c
hirenvadalia: This code was added ~3 years back and I think it served its purpose. But still, let's hear from the author of this code, @nithinj Do we need this special case of debug message anymore? If not then I would like to get rid of it.

src/server/node_manager.c
hirenvadalia: Thanks, @bremanandjk for the explanation, but my comment was in a generic or architectural way (sorry if it was not clear enough), as currently our mom structures are using array and looping through those arrays, which is slow, we can refactor those and make them faster.

src/server/node_manager.c
hirenvadalia: I agree that its good practice checks things before accessing, but we can improve this if/else as below:
```c
if ((pjob->ji_wattr[JOB_ATR_exec_vnode].at_flags & ATR_VFLAG_SET) == 0 || ((execvnod_in = pjob->ji_wattr[JOB_ATR_exec_vnode].at_val.at_str) == NULL)) {
	log_event(PBSEVENT_DEBUG2, PBS_EVENTCLASS_JOB, LOG_DEBUG, pjob->ji_qs.ji_jobid, "in free_nodes and no exec_vnode");
	return;
}
```

src/server/node_manager.c
hirenvadalia: @bremanandjk Looks like you missed push this change?

bremanandjk: On debugging further, I have realized we should not add this code in deallocate_job_from_node(). So left this code as it is.

src/server/node_manager.c
bremanandjk: I have renamed the function name.

src/server/node_manager.c
bremanandjk: As I had mentioned earlier, whether we decide to remove this bug specific debugging code or not, that should be handled separately by the original developer in separate PR. I don't want to delay this PR merge further with the "out of scope" code changes.

src/server/node_manager.c
bremanandjk: I have changed this code.

bremanandjk:
@hirenvadalia I have renamed the function as you suggest. 

@subhasisb & @agrawalravi90 Thanks for your approvals, This PR is ready to merge.

src/server/node_manager.c
nithinj: This is added at least two releases before. If it did not have helped so far, then I don't see any point in keeping it.

hirenvadalia:
@bremanandjk I think you need to do force push again to resolve 2 Travis status issue. (This happens if you do back to back force push in PR)

bremanandjk:
Not sure about 2 Travis builds, But the Travis build is successful. One build is showing a green tick and another one not getting results back.

bremanandjk:
@hirenvadalia  I have already done force push, cant try it again, 

git push -f mine freenodes_master_pr
Everything up-to-date

bremanandjk:
Thanks, @hirenvadalia  for approval and merging.

PR #1915, opened 2020-07-21
Have PTL raise an error if it failed to set sched config and modified the error message in _log_match() on non-existence 

birbalsain:
#### Describe Bug or Feature
PTL doesn't through any error if failed to set sched config. It deletes applied new sched configuration and revert back to  old configuration

In _log_match() If existence=False and log message is present then  _log_match() through  error message "log match: searching  for 'msg' with non-existence - No Match" . It is not a proper message.

If existence=True and log message is not present then it through error message "log match: searching  for 'msg' with existence - No Match" . It is meaningful.
So error message should be different.

#### Describe Your Change
PTL raise error if failed  to set sched configuration

For non-existence _log_match() through error message "log match: searching for msg - with non-existence - Match"

[sched_config_not_matched.txt](https://github.com/openpbs/openpbs/files/4953145/sched_config_not_matched.txt)
[sched_config_matched.txt](https://github.com/openpbs/openpbs/files/4953146/sched_config_matched.txt)

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -4061,7 +4061,7 @@ class PBSService(PBSObject):
                                          endtime=endtime)
             if not existence:
                 if rv:
-                    _msg = infomsg + ' - Match'
+                    _msg = infomsg + ' - but exists'
                     raise PtlLogMatchError(rc=1, rv=False, msg=_msg)
                 else:
                     self.logger.log(level, infomsg + attemptmsg + '... OK')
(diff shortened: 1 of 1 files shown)


test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4059,9 +4059,13 @@ def _log_match(self, logtype, msg, id=None, n=50, tail=True,
             rv = self.logutils.match_msg(lines, msg, allmatch=allmatch,
                                          regexp=regexp, starttime=starttime,
                                          endtime=endtime)
-            if rv is None and not existence:
-                self.logger.log(level, infomsg + attemptmsg + '... OK')
-                break
+            if not existence:
+                if rv:
+                    _msg = infomsg + ' - Match'
```
hirenvadalia, line 4064: Just minor nit: Instead `- Match` how about `- but exists` or `- exists`, so it will be more meaningful like `log match: searching for "msg" - with non-existence - but exists`

birbalsain: @hirenvadalia done the changes.

PR #1916, opened 2020-07-21
Improve conversion subjob index <-> tracking table index without looping whole table

hirenvadalia:
#### Describe Bug or Feature

* Submission and run path (aka req_runjob) take huge time when an array job is submitted with a high number of sub-jobs.
* This is due to almost everywhere we use `numindex_to_offset()` and `subjob_index_to_offset()` to get index in tracking table of sub-job (aka `struct ajtrkhd->tkm_tbl[]`). And `numindex_to_offset()` and `subjob_index_to_offset()` loops through whole table till it finds matching trk_index from entry in tkm_tbl[] (aka `struct ajtrk->trk_index`).
* But this looping is unnecessary and can be avoided by using simple calculatation as below:
  * for `numindex_to_offset()` -- `((int)(numindex - start) / step)`
  * for reverse -- `(sj index * step) + start`

#### Describe Your Change

* Use simple calculation to find offset in the tracking table, and remove trk_index from `struct ajtrk` as it's not used anymore

#### Link to Design Doc

* None as this is just code improvement, no external change

||normal job|with -J1-100|with -J1-10000|with -J1-200000|with -J1-20000000|
|-|-|-|-|-|-|
|Before changes|real	0m0.021s<br>user	0m0.002s<br>sys	0m0.007s|real	0m0.012s<br>user	0m0.000s<br>sys	0m0.002s|real	0m0.089s<br>user	0m0.000s<br>sys	0m0.002s|real	0m30.696s<br>user	0m0.002s<br>sys	0m0.000s|At this point server<br>goes to 100% cpu usage<br>and this qsub takes<br>long time to complete and<br>server stops responding<br>to any command as it is<br>looping through 20M<br>times 20M indexes<br>in tracking table|
|After changes|real	0m0.020s<br>user	0m0.008s<br>sys	0m0.000s|real	0m0.015s<br>user	0m0.002s<br>sys	0m0.000s|real	0m0.013s<br>user	0m0.002s<br>sys	0m0.000s|real	0m0.017s<br>user	0m0.002s<br>sys	0m0.000s|real	0m0.204s<br>user	0m0.002s<br>sys	0m0.000s|

Profile: regression
|--|--|--|--|--|--|--|
|2715|1794|2|0|0|0|1792|

(NOTE: 2 failures are known issues in master also)


src/include/svrfunc.h
```diff
@@ -54,6 +54,9 @@ extern "C" {
 #include "pbs_sched.h"
 #include "pbs_entlim.h"
 
+#define SJ_IDX_2_TBLIDX(pj, sjidx) ((int)((sjidx - pj->ji_ajtrk->tkm_start) / pj->ji_ajtrk->tkm_step))
```
ashwathraop, line 57: I would suggest you add a line or two of the comments explaining what this macro is.

src/server/array_func.c
```diff
@@ -625,19 +590,20 @@ subst_array_index(job *pjob, char *path)
...
-		return path;	/* unchanged */
+		return path; /* unchanged */
 
-	sprintf(cvt, "%d", pjob->ji_parentaj->ji_ajtrk->tkm_tbl[pjob->ji_subjindx].trk_index);
+	sprintf(cvt, "%d", SJ_TBLIDX_2_IDX(ppjob, pjob->ji_subjindx));
 	*pindorg = '\0';
-	(void)strcpy(trail, pindorg+strlen(index_tag));
-	(void)strcat(path, cvt);
-	(void)strcat(path, trail);
+	(void) strcpy(trail, pindorg + strlen(index_tag));
```
ashwathraop, line 604: Remove void.

src/server/array_func.c
```diff
@@ -791,29 +757,28 @@ setup_arrayjob_attrs(attribute *pattr, void *pobj, int mode)
 int
 fixup_arrayindicies(attribute *pattr, void *pobj, int mode)
 {
-	int   i;
-	int   x, y, z, ct;
+	int i;
+	int x, y, z, ct;
```
ashwathraop, line 761: Break the declaration to multi-line. And wouldn't it be better to use something meaningfull than x,y,z?

src/server/req_delete.c
```diff
@@ -363,8 +363,6 @@ req_deletejob(struct batch_request *preq)
 	int x, y, z;
```
ashwathraop, line 363: Here too.

src/server/array_func.c
```diff
@@ -791,29 +757,28 @@ setup_arrayjob_attrs(attribute *pattr, void *pobj, int mode)
...
 	char *ep;
-	job  *pjob = pobj;
+	job *pjob = pobj;
```
ashwathraop, line 763: Why do you need this? Why not use pobj directly?

src/server/array_func.c
```diff
@@ -1077,13 +1042,13 @@ mk_subjob_id(job *parent, int offset)
...
+ * @param[in]	pjob - job pointer
  * @param[in]	state -  job state.
  * @return	Pointer to static buffer
  * @par	MT-safe: No - uses a global buffer, "buf" and "buflen".
  */
 char *
-cvt_range(struct ajtrkhd *t, int state)
+cvt_range(job *pjob, int state)
 {
 	unsigned int f;	/* first of a pair or range   */
```
ashwathraop, line 1053: Again, I dont like the choice of variable names here. Can we have something meaningful?

src/server/req_stat.c
```diff
@@ -199,13 +199,13 @@ do_stat_of_a_job(struct batch_request *preq, job *pjob, int dohistjobs, int dosu
 static int
 stat_a_jobidname(struct batch_request *preq, char *name, int dohistjobs, int dosubjobs)
 {
-	int   i, indx, x, y, z;
+	int i, indx, x, y, z;
```
ashwathraop, line 202: Break declaration and have better variable names?

src/server/array_func.c
hirenvadalia: yes, pobj is a void pointer, to an operation like `pjob->ji_qs.ji_svrflags` we need job pointer. Also, all funcs called by this func expect job * so there also we need to typecast, so instead of doing typecase everywhere we just do in one place.

hirenvadalia:
@bhroam / @arungrover I still don't know many of scheduler code so asking you guys: We have `parse_subjob_index()` in range.c (used in the scheduler) and another in array_func.c (used in the server). So my question is can we move that to libutils and use in both places?

Both func is almost same, the only diff I can see is how it looks for char `]` in the given string, so I don't know which one is correct, and which one I should move to libutils and use it both places.

bhroam:
@hirenvadalia I reviewed the functions and they are practically identical.  They were probably written before libutil existed, so the only way was to copy the function.  From the looks of it, the server's function is a little more robust.  We should keep that one.

src/include/svrfunc.h
```diff
@@ -54,6 +54,25 @@ extern "C" {
...
+/*
+ * Convert given subjob index to tracking table array index
+ *
+ * For example: -J10-20:7 creates 2 subjob
+ *   1. [10] on 0th index in table
+ *   2. [17] on 1st index in table
+ * So SJ_IDX_2_TBLIDX(pj, 17) == ((17 - 10) / 7) == 1
+ */
+#define SJ_IDX_2_TBLIDX(pj, sjidx) ((int)((sjidx - pj->ji_ajtrk->tkm_start) / pj->ji_ajtrk->tkm_step))
+
```
bhroam, line 66: A downside to this over searching is you can get a false positive.  If I search for subjob index 15, I won't find it.  If I do the above math, I'll find subjob 17.

src/server/array_func.c
```diff
@@ -625,19 +590,20 @@ subst_array_index(job *pjob, char *path)
...
 
-	if (pjob->ji_parentaj == NULL)
+	if (ppjob == NULL)
 		return path;
 	if ((pindorg = strstr(path, index_tag)) == NULL)
-		return path;	/* unchanged */
+		return path; /* unchanged */
 
-	sprintf(cvt, "%d", pjob->ji_parentaj->ji_ajtrk->tkm_tbl[pjob->ji_subjindx].trk_index);
+	sprintf(cvt, "%d", SJ_TBLIDX_2_IDX(ppjob, pjob->ji_subjindx));
```
bhroam, line 602: Could you make cvt 21 or 22 characters (or maybe just 100).  20 is the length of a max signed long, 21 is the size of a max unsigned long.

src/server/req_delete.c
```diff
@@ -525,56 +521,48 @@ req_deletejob(struct batch_request *preq)
...
-				continue;
-			}
+		for (i = start; i <= end; i += step) {
+			int idx = SJ_IDX_2_TBLIDX(parent, i);
 
-			sjst = get_subjob_state(parent, i);
-			if ((sjst == JOB_STATE_EXITING) && !forcedel) {
-				x += z; /* ignore it */
+			sjst = get_subjob_state(parent, idx);
+			if ((sjst == JOB_STATE_EXITING) && !forcedel)
```
bhroam, line 547: You are doing a continue in a while loop without changing anything.  This could lead to an infinite loop.

src/server/req_runjob.c
```diff
@@ -432,21 +432,18 @@ req_runjob(struct batch_request *preq)
...
+		char fixjid[PBS_MAXSVRJOBID + 1];
 
 		/* if runjob request is from the Scheduler, */
 		/* it must have a destination specified     */
 		if (preq->rq_conn == psched->scheduler_sock) {
-			sprintf(log_buffer,
-				"runjob request from scheduler with null destination");
-			log_event(PBSEVENT_SYSTEM, PBS_EVENTCLASS_JOB, LOG_INFO,
-				jid, log_buffer);
+			log_event(PBSEVENT_SYSTEM, PBS_EVENTCLASS_JOB, LOG_INFO, jid, "runjob request from scheduler with null destination");
```
bhroam, line 442: Since you've been making other changes to bring this code in line with our standard, consider fixing the multiple single line comments to being one multiline comment.  Up to you.

src/server/req_runjob.c
```diff
@@ -457,15 +454,15 @@ req_runjob(struct batch_request *preq)
...
 			*pc = '\0';
-		pc = strchr(parent->ji_qs.ji_jobid, (int)'.');
+		pc = strchr(parent->ji_qs.ji_jobid, (int) '.');
 		if (pc)
-			(void)strcat(fixjid, pc);
+			(void) strcat(fixjid, pc);
 
-		(void)strncpy(pdefr->dr_id, fixjid, PBS_MAXSVRJOBID);
-		pdefr->dr_id[PBS_MAXSVRJOBID] ='\0';
+		(void) strncpy(pdefr->dr_id, fixjid, PBS_MAXSVRJOBID);
```
bhroam, line 464: Consider removing the (void)'s.  They don't do anything (not even remove the new gcc warning about not using return values).

src/include/svrfunc.h
hirenvadalia: Good catch @bhroam, I should have thought of negative case... will fix it.

src/server/req_delete.c
hirenvadalia: @bhroam I changed while loop to for loop, so I don't think it will be an infinite loop.

hirenvadalia:
@bhroam and @ashwathraop I have addressed all comments, please have a look at the latest commits.

PR #1917, opened 2020-07-21
socket lib init and socket close made platform independent.

Shrini-h:
#### Describe Bug or Feature
socket library initialization and socket close calls have been made platform independent.

#### Describe Your Change

Introduced initsocketlib() call to initialize any socket library. and use closesocket() to close socket fd or handle.

Platforms: CENTOS7,UBUNTU1804,SUSE15,UBUNTU1604
Profile: regression
|--|--|--|--|--|--|--|
|3041|3507|3|4|0|2|3498|

### DIFF
--- a/src/cmds/pbs_attach.c
+++ b/src/cmds/pbs_attach.c
@@ -261,9 +261,9 @@ main(int argc, char *argv[])
 
 	PRINT_VERSION_AND_EXIT(argc, argv);
 
-	if (initsocketlib()) {
+	if (initsocketlib())
 		return 1;
-	}
+
 #ifdef WIN32
 	/* Windows has an additional option -c, in order to run built-in DOS commands using a new command shell */
 	while ((c = getopt(argc, argv, "+j:p:h:m:csP")) != EOF) {
--- a/src/cmds/pbs_attach.c
+++ b/src/cmds/pbs_attach.c
@@ -261,10 +261,10 @@ main(int argc, char *argv[])
 
 	PRINT_VERSION_AND_EXIT(argc, argv);
 
-#ifdef WIN32
-	if (winsock_init()) {
+	if (initsocketlib()) {
 		return 1;
 	}
+#ifdef WIN32
 	/* Windows has an additional option -c, in order to run built-in DOS commands using a new command shell */
 	while ((c = getopt(argc, argv, "+j:p:h:m:csP")) != EOF) {
 #else
--- a/src/cmds/qsub.c
+++ b/src/cmds/qsub.c
@@ -2127,9 +2127,6 @@ set_sig_handlers(void)
 	signal(SIGBREAK, win_blockint);
 	signal(SIGTERM, win_blockint);
 
-	if (winsock_init()) {
-		return 1;
-	}
 	InitializeCriticalSection(&continuethread_cs);
 #else
 	/* Catch SIGPIPE on write() failures. */
@@ -5635,6 +5632,10 @@ main(int argc, char **argv, char **envp) /* qsub */
 	int daemon_up = 0;
 	char **argv_cpy; /* copy argv for getopt */
 	int i;
+
+	if (initsocketlib()) {
+		return 1;
+	}
 
 	/* Set signal handlers */
 	set_sig_handlers();
--- a/src/cmds/qsub.c
+++ b/src/cmds/qsub.c
@@ -1403,7 +1403,7 @@ no_suspend(int sig)
 
 /**
  * @brief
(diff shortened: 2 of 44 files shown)


src/include/pbs_internal.h
```diff
@@ -391,11 +391,12 @@ enum accrue_types {
 #endif
 #define LOCALHOST_SHORTNAME "localhost"
 #ifdef WIN32
-#define CLOSESOCKET(X) (void)closesocket(X)
 #define ERRORNO        WSAGetLastError()
+#define initsocketlib() winsock_init()
```
SudeshnaMoulik, line 395: How about moving this to pbs_config.h ?

src/cmds/qsub.c
```diff
@@ -5640,6 +5633,10 @@ main(int argc, char **argv, char **envp) /* qsub */
 	char **argv_cpy; /* copy argv for getopt */
 	int i;
 
+	if (initsocketlib()) {
```
SudeshnaMoulik, line 5636: Why are we moving this from set_sig_handlers() ?

Shrini-h: 1. Initializing socket library operation is not related to signal handling.
2. It is not possible to make it platform independent there

src/include/pbs_internal.h
Shrini-h: I agree this is not quite a right place. but since pbs_config.h is dynamically generated in some platforms it will not be possible to move it there.  May be we can find a clean place in next PRs, for now we can keep it here. Is it agreeable?

Shrini-h: I think I found a better place and moved it there, see https://github.com/openpbs/openpbs/pull/1917/commits/7efec9bf92b4900035dd88946aecda941176c6be

src/cmds/qsub.c
SudeshnaMoulik: Yes agree, I was thinking why it was like that. But as you noticed in other part of the code also we are initializing in main, so to make it consistent, it will be better to move in main.

src/lib/Libifl/tm.c
```diff
@@ -502,11 +498,10 @@ localmom()
 				case ECONNREFUSED:
 #ifdef WIN32
 				case WSAEINTR:
```
hirenvadalia, line 500: I think we should put `#define` for this one to EINTR in portability.h and remove this case block

src/cmds/pbs_attach.c
```diff
@@ -261,10 +261,10 @@ main(int argc, char *argv[])
 
 	PRINT_VERSION_AND_EXIT(argc, argv);
 
-#ifdef WIN32
-	if (winsock_init()) {
+	if (initsocketlib()) {
```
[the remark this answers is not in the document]
Shrini-h: Was expecting this comment. Fixed it everywhere in https://github.com/openpbs/openpbs/pull/1917/commits/813100c9daa674ebe91d61a749a7a80d03b8bd0c

src/cmds/qsub.c
```diff
@@ -1412,14 +1412,10 @@ no_suspend(int sig)
  *
  */
 static void
-close_sock(int sock)
+shut_close_sock(int sock)
```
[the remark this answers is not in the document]
Shrini-h: Though not my change, removed the line in https://github.com/openpbs/openpbs/pull/1917/commits/813100c9daa674ebe91d61a749a7a80d03b8bd0c

src/lib/Libifl/tm.c
Shrini-h: Its not a straight forward change as we need to carefully study the repercussions of doing such a change. So not touching it in this PR. may be we should handle in a future PR.

hirenvadalia: Currently there is only 9 place we are using WSAEINTR in whole code base and everywhere its doing exactly same stuff as EINTR. So I think its pretty straight forward change which will remove few more ifdef.

Shrini-h: Its not the number of places that I'm concerned off, but how best we can make the changes around it as platform independent, Its easy to textually change 9 lines but we don't know what impact it has on compilation and run time on different platforms. So lets go without it now and I promise to make this change in a proper way in an upcoming PR. If you are really insisting this change in the current PR for it to get your approval, please let me know why you think so that it has to be in the current PR.

Thank you in advance for your understanding

hirenvadalia: If you making changes in next PR already, I am fine with going without it in this PR.

PR #1918, opened 2020-07-21
added mom restart in delete_vnode_def()

vccardenas:
#### Describe Bug or Feature
Extra vnodes that remain from a previous test did not get cleaned up before running a test. This issue can be seen easily on a cpuset system.

#### Describe Your Change
In revert_mom(), moved the creation of node to after vnode def file is deleted and mom restarted.

[before_cpuset.txt](https://github.com/openpbs/openpbs/files/4956641/before_cpuset.txt)
[after_cpuset.txt](https://github.com/openpbs/openpbs/files/4962800/after_cpuset.txt)

[before_linux.txt](https://github.com/openpbs/openpbs/files/4956642/before_linux.txt)
[after_linux.txt](https://github.com/openpbs/openpbs/files/4962803/after_linux.txt)

### DIFF
--- a/test/fw/ptl/utils/pbs_testsuite.py
+++ b/test/fw/ptl/utils/pbs_testsuite.py
@@ -1462,7 +1462,6 @@ class PBSTestSuite(unittest.TestCase):
             self.assertTrue(mom.isUp(), msg)
         mom.pbs_version()
         restart = False
-        enabled_cpuset = False
         if ((self.revert_to_defaults and self.mom_revert_to_defaults and
              mom.revert_to_default) or force):
             # no need to delete vnodes as it is already deleted in
@@ -1476,25 +1475,6 @@ class PBSTestSuite(unittest.TestCase):
             if 'clienthost' in self.conf:
                 conf.update({'$clienthost': self.conf['clienthost']})
             mom.apply_config(conf=conf, hup=False, restart=False)
-            if mom.is_cpuset_mom():
-                self.server.manager(MGR_CMD_CREATE, NODE, None, mom.shortname)
-                # In order to avoid intermingling CF/HK/PY file copies from the
-                # create node and those caused by the following call, wait
-                # until the dialogue between MoM and the server is complete
-                time.sleep(4)
-                just_before_enable_cgroup_cset = time.time()
-                mom.enable_cgroup_cset()
-                # a high max_attempts is needed to tolerate delay receiving
-                # hook-related files, due to temporary network interruptions
-                mom.log_match('pbs_cgroups.CF;copy hook-related '
-                              'file request received', max_attempts=120,
-                              starttime=just_before_enable_cgroup_cset-1,
-                              interval=1)
-                # Make sure that the MoM will generate per-NUMA node vnodes
-                # when the natural node is created below
-                # HUP may not be enough if exechost_startup is delayed
-                restart = True
-                enabled_cpuset = True
         if restart:
             mom.restart()
         else:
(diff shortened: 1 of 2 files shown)


vccardenas:
I agree there are too many restarts and HUPs - seems more scope is involved so I'll file a separate ticket for that. Unfortunately we need to restart the mom after deleting the vnode def file to fix this issue. This is documened in the PBS Professional 19.4 Admin Guide, section 3.1.6.2.

bayucan:
@bhroam : Restarting pbs_mom after doing 'pbs_mom -s remove <vnodedef config file>' (which really does a rm <vnode def config file>),  ensures those vnodes are not remembered. It would be a clean segway to creating new vnodes, which uses insert_vnode_def() and that is also requiring a separate restart. Otherwise, we run into cases where previous vnodes are sometimes left behind...

hirenvadalia:
> We restart mom several times during setUp(). Is it wise to add another? Maybe add a way to mark the mom as needing a restart and then restart it once at the end of setUp()?

I agree with this, and as now PBSTestSuite->setUp()->revert_moms()->revert_mom() already restarts mom when it deletes vnodes... see below:
https://github.com/openpbs/openpbs/blob/b936d8ffb7bbb785a321a10cac1b435af3fb35f2/test/fw/ptl/utils/pbs_testsuite.py#L1464-L1499

lisa-altair:
@vccardenas I think you meant to remove the changes to test/fw/ptl/lib/pbs_testlib.py?

lisa-altair:
@hirenvadalia  is right, I looked at all of the calls to delete_vnode_defs and after that call, a restart is done.  I'm not sure why we couldn't remove those restarts and put it within delete_vnode_defs itself...

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -1476,42 +1475,34 @@ def revert_mom(self, mom, force=False):
...
         if not mom.isUp():
             self.logger.error('mom ' + mom.shortname + ' is down after revert')
         a = {'state': 'free'}
-        if enabled_cpuset:
-            # Checking whether the CF file was copied really belongs in code
-            # that changes the config file -- i.e. after enable_cgroup_cset
-            # called above. We're not sure it is always called here,
-            # since that call is in an if
+        self.server.manager(MGR_CMD_CREATE, NODE, None, mom.shortname)
+        if mom.is_cpuset_mom():
```
bayucan, line 1486: I don't think it's correct to move the enable pbs_cgroups hook code snippet here. It's meant to be done only when revert_to_default option is set in revert_mom().

hirenvadalia:
> I'm not sure why we couldn't remove those restarts and put it within delete_vnode_defs itself.

Hey @lisa-altair, This is because we want to minimize restarts of mom during setUp() else it can slow down tests execution, as setUp() will run for every test. HTH!

@vccardenas and/or @bayucan I am still not able to understand why we need this change? As I can see current setUp() code, it is deleting all nodes (including vnodes) from the server (when it reverts server) and also deletes vnode def file when it reverts mom and restarts mom

test/fw/ptl/utils/pbs_testsuite.py
bayucan: One suggestion is to bring back 'enable_cpuset' flag, set it True under if *revert_to_default, then later, instead of if mom.is_cpuset(), check for if 'enable_cpuset'.

vccardenas:
Marked as WIP.

@hirenvadalia and @bhroam , the problem is that after a delete_vnode_def() we need a restart.  I'll address minimizing the restarts in the next commit as well as the ordering of enable_cpuset_group() when revert_to_defaults is True.

vccardenas:
Removed WIP from the title.  Reviewers kindly review once more.

PR #1921, opened 2020-07-22
ralter on standing resv may cause oversubscription

arungrover:
#### Describe Bug or Feature
Doing ralter on a standing reservation changes its first upcoming occurrence. The server creates the tasks for future occurrences at the correct time. But, once a standing reservation's start time is modified, there is no way for scheduler to know what was the original start time of the reservation before ralter happened. This makes scheduler create all future reservation occurrences according to the modified start time.
This anomaly can then cause the scheduler to confirm/deny other reservations or top jobs at wrong times and can potentially cause oversubscription.

#### Describe Your Change
When start time of a reservation is changed, the original value is already stored inside a structure. This change makes that original value into an attribute on the reservation and makes the scheduler use it to create future reservation events.

#### Link to Design Doc
NA

[test_before.txt](https://github.com/openpbs/openpbs/files/4957064/test_before.txt)
[test_after.txt](https://github.com/openpbs/openpbs/files/4957065/test_after.txt)

more test logs coming soon


src/lib/Libattr/master_resv_attr_def.xml
```diff
@@ -852,6 +852,23 @@
...
+   <attributes>
+      <member_index>RESV_ATR_start_revert</member_index>
+      <member_name>ATTR_resv_start_revert</member_name>
+      <member_at_decode>decode_l</member_at_decode>
+      <member_at_encode>encode_l</member_at_encode>
+      <member_at_set>set_l</member_at_set>
+      <member_at_comp>comp_l</member_at_comp>
+      <member_at_free>free_null</member_at_free>
+      <member_at_action>NULL_FUNC</member_at_action>
+      <member_at_flags>READ_WRITE </member_at_flags>
```
bhroam, line 864: make this READ_ONLY, and possibly only by managers

src/scheduler/data_types.h
```diff
@@ -756,6 +756,7 @@ struct resv_info
 	int resv_idx;			/* the index of standing resv occurrence */
 	int count;			/* the total number of occurrences */
 	time_t req_start;		/* user requested start time of resv */
+	time_t req_start_future;	/* For standing reservations being modified, this will be the start time of future occurrences */
```
bhroam, line 759: I'd call this req_start_orig.  The comment is wrong too.  This isn't the start of the future occurrences, this is used to create the start time of the future occurrences.

src/scheduler/resv_info.c
```diff
@@ -485,6 +485,13 @@ query_reservations(int pbs_sd, server_info *sinfo, struct batch_status *resvs)
 				* execvnode on which the occurrence is confirmed to run.
 				*/
 			for (j = 0; occr_idx <= count; occr_idx++, j++, degraded_idx++) {
+				/* If it is not the first occurrence then update the start time as
+				 * req_start_future (if set). This is to ensure that if the first
+				 * occurrence has been changed, othe future occurrences are not
+				 * affected.
```
bhroam, line 491: othe=other

test/tests/functional/pbs_ralter.py
```diff
@@ -2263,3 +2263,36 @@ def test_resv_resc_assigned(self):
...
+        offset = 3600
+        shift = -3000
+
+        rid, start, end = self.submit_and_confirm_reservation(
+            offset, duration, select="2:ncpus=4", standing=True,
+            rrule="FREQ=HOURLY;COUNT=3")
+
+        self.server.expect(RESV,
+                           {'reserve_state': (MATCH_RE, 'RESV_CONFIRMED|2')},
+                           rid)
```
bhroam, line 2284: No need for this, submit_and_confirm_reservation() does this

test/tests/functional/pbs_ralter.py
```diff
@@ -2263,3 +2263,36 @@ def test_resv_resc_assigned(self):
...
+        # move the reservation 10 mins in future
+        self.alter_a_reservation(rid, start, end, confirm=True, alter_s=True,
+                                 alter_e=True, shift=shift)
```
bhroam, line 2287: I think the comment is incorrect, shift is -3600.  Won't this try and move the start time to now?  Actually -3600 will likely be in the past because some time has passed.  Don't you want something like -1800?

test/tests/functional/pbs_ralter.py
```diff
@@ -2263,3 +2263,36 @@ def test_resv_resc_assigned(self):
...
+        self.server.expect(RESV,
+                           {'reserve_state': (MATCH_RE, 'RESV_CONFIRMED|2')},
+                           rid)
```
bhroam, line 2290: No need for this, alter_a_reservation() does this

test/tests/functional/pbs_ralter.py
```diff
@@ -2263,3 +2263,36 @@ def test_resv_resc_assigned(self):
...
+        # Now make another reservation which runs in about 50 mins from now.
+        # Ideally this reservation should confirm because second occurrence
+        # of the first reservation happens in almost 2 hrs from now.
+        rid2, start, end = self.submit_and_confirm_reservation(
+            3000, 1800, select="2:ncpus=4")
+        self.server.expect(RESV,
+                           {'reserve_state': (MATCH_RE, 'RESV_CONFIRMED|2')},
+                           id=rid2)
```
bhroam, line 2298: Same as above, this isn't required.

src/lib/Libattr/master_resv_attr_def.xml
arungrover: How about if I make this PRIV_READ?

test/tests/functional/pbs_ralter.py
arungrover: No, shift is 3000. It's the offset that's set to 3600.

src/lib/Libattr/master_resv_attr_def.xml
bhroam: ATR_DFLAG_MGRD is probably better because this is an internal value that only the scheduler/pbs_sim/pbs_est  need.  You need to make managers see it because pbs_est runs as a root which is a manager.

arungrover:
[test_ralter.txt](https://github.com/openpbs/openpbs/files/4963212/test_ralter.txt)

arungrover:
[test_ralter.txt](https://github.com/openpbs/openpbs/files/4967158/test_ralter.txt)

src/include/pbs_ifl.h
```diff
@@ -85,31 +85,32 @@ extern "C" {
...
+#define ATTR_auth_h			"Authorized_Hosts"
+#define ATTR_cred			"cred"
+#define ATTR_nodemux			"no_stdio_sockets"
+#define ATTR_umask			"umask"
+#define ATTR_block			"block"
+#define ATTR_convert			"qmove"
+#define ATTR_DefaultChunk		"default_chunk"
+#define ATTR_X11_cookie			"forward_x11_cookie"
+#define ATTR_X11_port			"forward_x11_port"
+#define ATTR_GUI			"gui"
```
bhroam, line 113: If you are going  to format these to line them up, some of them are further indented than others.

src/include/reservation.h
```diff
@@ -124,7 +118,7 @@ struct resv_alter {
 	char *ra_select;
 	long ra_state;
 	unsigned long ra_flags;
-	struct resv_alter_revert ra_revert;
+	char *rr_select; /* used to revert select*/
```
bhroam, line 121: Change this to ra_select_revert.  Add 'for future occurrences' to the end of the comment.

src/lib/Libattr/master_resv_attr_def.xml
```diff
@@ -869,6 +869,23 @@
...
+      <member_at_encode>encode_l</member_at_encode>
+      <member_at_set>set_l</member_at_set>
+      <member_at_comp>comp_l</member_at_comp>
+      <member_at_free>free_null</member_at_free>
+      <member_at_action>NULL_FUNC</member_at_action>
+      <member_at_flags>ATR_DFLAG_MGRD</member_at_flags>
+      <member_at_type>ATR_TYPE_LONG</member_at_type>
+      <member_at_parent>PARENT_TYPE_RESV</member_at_parent>
+      <member_verify_function>
+         <ECL>verify_datatype_time</ECL>
```
bhroam, line 885: No need for this, this value is set by the server.

test/tests/functional/pbs_ralter.py
```diff
@@ -2287,6 +2287,35 @@ def test_alter_standing_resv_future_occrs(self):
         rid2, start, end = self.submit_and_confirm_reservation(
             3000, 1800, select="2:ncpus=4")
 
+    def test_alter_start_standing_resv_future_occrs(self):
+        """
+        Test that when duration of a confirmed standing reservation is
+        altered, only the upcoming occurence changes and not all occurences
+        are modified.
+        """
```
bhroam, line 2295: Don't both the tests have the same function name?

test/tests/functional/pbs_ralter.py
```diff
@@ -2287,6 +2287,35 @@ def test_alter_standing_resv_future_occrs(self):
...
+
+        duration = 180
+        offset = 30
```
bhroam, line 2298: Since you aren't waiting for these reservations to start or end, can you make the offset and duration larger?

test/tests/functional/pbs_ralter.py
```diff
@@ -2279,20 +2279,38 @@ def test_alter_standing_resv_future_occrs(self):
...
         # Ideally this reservation should confirm because second occurrence
         # of the first reservation happens in almost 2 hrs from now.
         rid2, start, end = self.submit_and_confirm_reservation(
             3000, 1800, select="2:ncpus=4")
+
+    def test_alter_standing_resv_check_start_time(self):
+        """
+        Test that when start time of a confirmed standing reservation is
+        altered, all its occurrences start at the right time.
+        """
```
bhroam, line 2323: I don't  think this test is required.  The server always did the right thing.  I don't even think you need the other test you wrote either.  If you want to write a test, then write one where the scheduler overlaps a reservation with a future occurrence because it thinks the occurrences are sooner than they are.

src/scheduler/data_types.h
```diff
@@ -759,6 +759,7 @@ struct resv_info
 	time_t req_start_orig;		/* For standing reservations being modified, this will used to get start time of future occurrences */
 	time_t req_end;			/* user requested end tiem of resv */
 	time_t req_duration;		/* user requested duration of resv */
+	time_t req_duration_orig;	/* For standing reservations being modified, this will used to get duration of future occurrences */
```
bhroam, line 762: I see you query this, but don't actually use it.

src/include/pbs_ifl.h
arungrover: Oh no,
It shows up perfectly in my editor
Screen Shot 2020-07-23 at 9 59 18 AM

src/include/reservation.h
arungrover: not sure of 'rr' goes with '_revert'. How about select_revert?

src/lib/Libattr/master_resv_attr_def.xml
arungrover: I'll make both server only attributes

test/tests/functional/pbs_ralter.py
arungrover: oops

src/include/reservation.h
bhroam: The prefix of all members of a structure should be the same.  It's an indication to the user what structure it is at its use.  That's why it needs to start with 'ra_'

test/tests/functional/pbs_ralter.py
arungrover: In this test we are checking that server is doing the right thing by starting the reservation at the right time. @tpai26 suggested me to write this test because of the recent inflow of problems that we have seen with the ralter changes.

I think the other test does check for the reservation to be denied. Since we change the duration of standing reservation from 180 to 20 seconds, buggy behavior will confirm the second reservation.

src/scheduler/data_types.h
arungrover: It is used in resv_info.c (line 558)

test/fw/ptl/lib/pbs_ifl_mock.py
```diff
@@ -94,6 +94,7 @@
 ATTR_stageout = 'stageout'
 ATTR_resvTag = 'reserve_Tag'
 ATTR_resv_start = 'reserve_start'
+ATTR_resv_start_revert = 'reserve_start_revert'
```
vstumpf, line 97: Do you need one of these for reserve_duration_revert?

arungrover: That's a good catch. Now that the attribute is only MGR read-only I'm not sure I need an entry here.  I think I should remove start_revert as well.
@bhroam what do you think?

bhroam: You should still have the ATTR_* macros, but they should be in pbs_internal.h.  I guess then they should be here too.

arungrover: @vstumpf I've addressed your comment.

src/include/pbs_ifl.h
```diff
@@ -85,30 +85,32 @@ extern "C" {
...
+#define ATTR_inter			"interactive"
+#define ATTR_sandbox			"sandbox"
+#define ATTR_stagein			"stagein"
+#define ATTR_stageout			"stageout"
+#define ATTR_resvTag			"reserve_Tag"
+#define ATTR_resv_start			"reserve_start"
+#define ATTR_resv_start_revert		"reserve_start_revert"
+#define ATTR_resv_end			"reserve_end"
+#define ATTR_resv_duration		"reserve_duration"
+#define ATTR_resv_duration_revert	"reserve_duration_revert"
```
bhroam, line 98: This are internal.  Move them to pbs_internal.h

PR #1923, opened 2020-07-22
Fix microseconds conversion in TestCrayAlpsReleaseTunables

shilpakodli:
#### Describe Bug or Feature
Two tests from TestCrayAlpsReleaseTunables were failing with error "ValueError: unconverted data remains: .621222".
In PTL, we are adding high resolution logging in all daemons logs. In test case, we are getting only till seconds of string in strptime as "%m/%d/%Y %H:%M:%S" and hence the error.

After fixing above issue, test started failing with error : AssertionError: 3.2769370079040527 not less than or equal to 3 : alps_release_jitter not working

#### Describe Your Change
Added microseconds formatting while converting string to date time.
Add math.floor to get the time difference to an integer value.

[TestCrayAlpsReleaseTunables_after_fix.txt](https://github.com/openpbs/openpbs/files/4959829/TestCrayAlpsReleaseTunables_after_fix.txt)
[TestCrayAlpsReleaseTunables_before_fix.txt](https://github.com/openpbs/openpbs/files/4959830/TestCrayAlpsReleaseTunables_before_fix.txt)

### DIFF
--- a/test/tests/functional/pbs_alps_release_tunables.py
+++ b/test/tests/functional/pbs_alps_release_tunables.py
@@ -3,49 +3,41 @@
 # Copyright (C) 1994-2020 Altair Engineering, Inc.
 # For more information, contact Altair at www.altair.com.
 #
-# This file is part of both the OpenPBS software ("OpenPBS")
-# and the PBS Professional ("PBS Pro") software.
+# This file is part of the PBS Professional ("PBS Pro") software.
 #
 # Open Source License Information:
 #
-# OpenPBS is free software. You can redistribute it and/or modify it under
-# the terms of the GNU Affero General Public License as published by the
-# Free Software Foundation, either version 3 of the License, or (at your
-# option) any later version.
+# PBS Pro is free software. You can redistribute it and/or modify it under the
+# terms of the GNU Affero General Public License as published by the Free
+# Software Foundation, either version 3 of the License, or (at your option) any
+# later version.
 #
-# OpenPBS is distributed in the hope that it will be useful, but WITHOUT
-# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
-# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Affero General Public
-# License for more details.
+# PBS Pro is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.
+# See the GNU Affero General Public License for more details.
 #
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 # Commercial License Information:
 #
-# PBS Pro is commercially licensed software that shares a common core with
-# the OpenPBS software.  For a copy of the commercial license terms and
-# conditions, go to: (http://www.pbspro.com/agreement.html) or contact the
-# Altair Legal Department.
+# For a copy of the commercial license terms and conditions,
+# go to: (http://www.pbspro.com/UserArea/agreement.html)
+# or contact the Altair Legal Department.
 #
-# Altair's dual-license business model allows companies, individuals, and
-# organizations to create proprietary derivative works of OpenPBS and
+# Altair’s dual-license business model allows companies, individuals, and
+# organizations to create proprietary derivative works of PBS Pro and
 # distribute them - whether embedded or bundled with other software -
 # under a commercial license agreement.
 #
-# Use of Altair's trademarks, including but not limited to "PBS™",
-# "OpenPBS®", "PBS Professional®", and "PBS Pro™" and Altair's logos is
-# subject to Altair's trademark licensing policies.
-
+# Use of Altair’s trademarks, including but not limited to "PBS™",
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_alps_release_tunables.py
```diff
@@ -44,7 +44,7 @@
 
 def get_epoch(msg):
     # Since its a log message split on ';' to get timestamp
-    a = time.strptime(msg.split(';')[0], "%m/%d/%Y %H:%M:%S")
+    a = time.strptime(msg.split(';')[0], "%m/%d/%Y %H:%M:%S.%f")
```
bhroam, line 47: Use LogUtil's convert_date_time().  This handles microseconds properly.  time's %f is broken.  You need to use datetime which convert_date_time() does.

shilpakodli:
@bhroam I have addressed your review comments. Please find latest execution logs.
[Latest_execution_logs.txt](https://github.com/openpbs/openpbs/files/5014099/Latest_execution_logs.txt)

I have changed math.ceil to math.floor because, test was failing with assertion error.

lisa-altair:
Hi @shilpakodli, one thing I do not understand is why things are being made to be integers.  
Previously there was no high resolution logging available, and so the float values of "alps_release_wait_time" and "alps_release_jitter" had to be compared against integer log values.  But now that there is high resolution logging, then it should be even easier to use the float values, right?

test/tests/functional/pbs_alps_release_tunables.py
```diff
@@ -151,10 +149,11 @@ def test_alps_release_jitter(self):
                 retry -= 1
                 # variable 'out' is a list of tuples and every second element
                 # in a tuple is the matched log message
-                time_prev = get_epoch(out[0][1])
+                time_prev = self.get_epoch(out[0][1])
                 for data in out[1:]:
-                    time_current = get_epoch(data[1])
-                    self.assertLessEqual(time_current - time_prev, max_delay,
+                    time_current = self.get_epoch(data[1])
+                    self.assertLessEqual(math.floor(time_current - time_prev),
```
lisa-altair, line 155: You shouldn't need math.floor here.  Just compare the high res times against the real max_delay time.  The logs should not be spread out more than max_delay.  However, it may be possible that if the system is a bit slow and the random number generated for jitter makes us equal to max_delay, then the logs can be logged a bit slower and thus be farther apart than max_delay.  In which case I think you can increase the alps_release_wait_time to workaround the slower system.

test/tests/functional/pbs_alps_release_tunables.py
```diff
@@ -96,12 +94,12 @@ def test_alps_release_wait_time(self):
...
             # in a tuple is the matched log message
-            time_prev = get_epoch(out[0][1])
+            time_prev = self.get_epoch(out[0][1])
             for data in out[1:]:
-                time_current = get_epoch(data[1])
+                time_current = self.get_epoch(data[1])
                 fail_msg = "alps_release_wait_time not working"
                 self.assertGreaterEqual(time_current - time_prev,
-                                        math.ceil(arwt),
+                                        math.floor(arwt),
```
lisa-altair, line 102: Now that we have high res timestamps, do we still need to make arwt into an integer?

shilpakodli:
@bhroam and @lisa-altair I have addressed your review comments. Please review latest code changes.
[Latest_execution_logs.txt](https://github.com/openpbs/openpbs/files/5069235/Latest_execution_logs.txt)

test/tests/functional/pbs_alps_release_tunables.py
```diff
@@ -3,49 +3,41 @@
...
-# OpenPBS is free software. You can redistribute it and/or modify it under
-# the terms of the GNU Affero General Public License as published by the
-# Free Software Foundation, either version 3 of the License, or (at your
-# option) any later version.
+# PBS Pro is free software. You can redistribute it and/or modify it under the
+# terms of the GNU Affero General Public License as published by the Free
+# Software Foundation, either version 3 of the License, or (at your option) any
+# later version.
 #
-# OpenPBS is distributed in the hope that it will be useful, but WITHOUT
```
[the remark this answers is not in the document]
lisa-altair: This copyright/license header still looks wrong.

test/tests/functional/pbs_alps_release_tunables.py
```diff
@@ -151,10 +149,11 @@ def test_alps_release_jitter(self):
                 retry -= 1
                 # variable 'out' is a list of tuples and every second element
                 # in a tuple is the matched log message
-                time_prev = get_epoch(out[0][1])
+                time_prev = self.get_epoch(out[0][1])
                 for data in out[1:]:
-                    time_current = get_epoch(data[1])
-                    self.assertLessEqual(time_current - time_prev, max_delay,
+                    time_current = self.get_epoch(data[1])
+                    self.assertLessEqual(int(time_current - time_prev),
```
lisa-altair, line 155: I'm still not understanding why you are doing an int() here.  I must be missing something.  Can you please explain?  Thanks.

lisa-altair:
@shilpakodli Please attach latest test logs.

shilpakodli:
Attaching latest execution logs:
[Latest_execution_logs.txt](https://github.com/openpbs/openpbs/files/5082779/Latest_execution_logs.txt)

lisa-altair:
Thanks @shilpakodli looks good to me.

PR #1924, opened 2020-07-22
Updated man pages for OpenPBS 20

agurban:
#### Describe Bug or Feature
Updated man pages for OpenPBS 20


hirenvadalia:
@agurban I think we need to update Makefile.am for deleted man pages, Let me know if you need help in that. Also not sure but I don't think we need `win_postinstall.py.8B` here.

agurban:
There is also at least one new man page (pbs_login).  For win_postinstall, we tell the admin to run
the command, so I think we need a man page for it.  Thank you for looking at this.

PR #1925, opened 2020-07-22
Allow provisioned jobs to run even with pending hook copy action from nodes not part of job

bayucan:
#### Describe Bug or Feature

* Provisioning jobs stay in substate 71 after provisioning succeeds, when there's a pending copy hook operation.
* Server would continue to have pending copy hook operation when some of the mom nodes in the complex are down.
* Provisioning jobs would get blocked even if nodes that are not part of the job having a pending copy hook operation.
* Sample reproduction:
  

 ```
# cat prov.py
bayucan@corretja:~/bugs/pbs_23641> cat prov.py
import pbs
import time
pbs.logmsg(pbs.LOG_DEBUG, "prov hook: sleeping for 20 seconds")
time.sleep(20)
pbs.logmsg(pbs.LOG_DEBUG, "prov hook: done sleeping..")
pbs.event().accept(0)
# qmgr -c "create hook prov event=provision"
# qmgr -c "import hook prov application/x-python default prov.py"
Have 2 nodes that are not on the server host.
# qmgr -c "list node @default"
federer
     Mom = federer.pbspro.com
     state = free
     provision_enable = true
     resources_available.aoe = osimage1
    ...
lendl
     Mom = lendl.pbspro.com
     state = free

Bring down the node without provisioning enabled:
# ssh lendl systemctl stop pbs
Force a mom hook copy operation:
# qmgr -c "create hook begin event=execjob_begin"

At this point, the copy hook operation to the downed host 'lendl' would be pending.

Submit job:
% qsub -l select=aoe=osimage1 <job.scr>
1.corretja


% qstat
1.corretja     STDIN            user1                    0 R workq

At this point, the provision hook has succeeded, and a reboot of the host needs to be done to complete provisioning:
# ssh federer systemctl stop pbs && sleep 5 && systemctl start pbs
 
Job would show up as stuck in substate=71 due to copy hook pending operation on the downed lendl, which is not even part of the job:
# qstat -f 1.corretja | egrep exec
exec_vnode = (federer:aoe=osimage1:ncpus=1)

# qstat -f 1.corretja | grep state
    job_state = R
    substate = 71
```
* One thing to note, if any of the non-provision nodes that are part of the job have a pending hook copy operation, it would not prevent the provision job from starting. This also happens with regular/non-provision jobs. This fix will only prevent starting the provisioned job, if any of the provision nodes have pending hook copy operation. 
#### Describe Your Change

* Updated code so that only if the nodes that have pending copy hook operation are the provisioned nodes assigned to the job.

* [ptl.test_os_provisioning_hook_copy.FAIL.txt](https://github.com/openpbs/openpbs/files/4963085/ptl.test_os_provisioning_hook_copy.FAIL.txt)
* [ptl.test_os_provisioning_hook_copy.PASS.txt](https://github.com/openpbs/openpbs/files/4963088/ptl.test_os_provisioning_hook_copy.PASS.txt)
* valgrind log after running pbs_provisioning.py testsuite:
[valgrind.running_provision.txt](https://github.com/openpbs/openpbs/files/4974715/valgrind.running_provision.txt)

* [Results_after_running_provisioning_tests.pdf](https://github.com/openpbs/openpbs/files/4963263/Results_after_running_provisioning_tests.pdf)

### DIFF
--- a/src/server/svr_func.c
+++ b/src/server/svr_func.c
@@ -5524,12 +5524,12 @@ prov_startjob(struct work_task *ptask)
 
 	assert(ptask->wt_parm1 != NULL);
 	pjob = (job *) ptask->wt_parm1;
-        if (pjob == NULL) {
+	if (pjob == NULL) {
 		DBPRT(("%s: pjob is NULL\n", __func__))
 		return;
 	}
 	/* task being serviced here */
-        pjob->ji_prov_startjob_task = NULL;
+	pjob->ji_prov_startjob_task = NULL;
 	if ((do_sync_mom_hookfiles || sync_mom_hookfiles_proc_running) &&
 	    (prov_vnode_pending_hook_copy(pjob))) {
 
@@ -5543,14 +5543,13 @@ prov_startjob(struct work_task *ptask)
 			"hookfiles is not completed\n"
 			, __func__))
 
-                /* set a work task to run after 5 sec from now */
+		/* set a work task to run after 5 sec from now */
 		pjob->ji_prov_startjob_task = set_task(WORK_Timed, time_now + 5,
                         				prov_startjob, pjob);
-                if (pjob->ji_prov_startjob_task == NULL) {
-                        log_err(errno, __func__,
-				"Unable to set task for prov_startjob; requeuing the job");
+		if (pjob->ji_prov_startjob_task == NULL) {
+			log_err(errno, __func__, "Unable to set task for prov_startjob; requeuing the job");
 			(void)force_reque(pjob);
-                }
+		}
 		return;
         }
 
--- a/src/server/svr_func.c
+++ b/src/server/svr_func.c
@@ -5490,9 +5490,7 @@ prov_vnode_pending_hook_copy(job *pjob)
 		for (j = 0; j < np->nd_nummoms; j++) {
 
 			if ((np->nd_moms[j] != NULL) && (sync_mom_hookfiles_count(np->nd_moms[j]) > 0)) {
-				snprintf(log_buffer, sizeof(log_buffer),
-						"prov vnode %s's parent mom %s:%d has a pending copy hook or delete hook request", np->nd_name,  np->nd_moms[j]->mi_host, np->nd_moms[j]->mi_port);
-				log_event(PBSEVENT_DEBUG3, PBS_EVENTCLASS_NODE, LOG_WARNING, pjob->ji_qs.ji_jobid, log_buffer);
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/server/svr_func.c [hunk 1]
```diff
@@ -5431,6 +5431,81 @@ is_vnode_prov_done(char * vnode)
...
+				__func__, prov_vnode_list[i]))
+			goto prov_vnode_label;
+		}
+		/* hook has not been sent */
+		for (j = 0; j < np->nd_nummoms; j++) {
+
+			if ((np->nd_moms[j] != NULL) && (sync_mom_hookfiles_count(np->nd_moms[j]) > 0)) {
+				snprintf(log_buffer, sizeof(log_buffer),
+						"prov vnode %s's parent mom %s:%d has a pending copy hook or delete hook request", np->nd_name,  np->nd_moms[j]->mi_host, np->nd_moms[j]->mi_port);
+				log_event(PBSEVENT_DEBUG3, PBS_EVENTCLASS_NODE, LOG_WARNING, pjob->ji_qs.ji_jobid, log_buffer);
```
hirenvadalia, line 5495: Please use log_eventf

test/tests/functional/pbs_provisioning.py
```diff
@@ -177,3 +177,45 @@ def test_app_provisioning(self):
...
+        self.server.expect(NODE, {'state': 'free'}, id=self.hostA)
+
+        job = Job(TEST_USER1, attrs={ATTR_l: 'aoe=osimage1'})
+        job.set_sleep_time(30)
+        jid = self.server.submit(job)
+        self.server.expect(NODE, {'state': 'provisioning'}, id=self.hostA)
+        self.server.expect(JOB, {'job_state': 'R', 'substate': 71}, id=jid)
+        self.server.log_match(
+            "fake calling os provisioning script",
+            max_attempts=20,
```
hirenvadalia, line 209: I don't think we need max_attempts here as its < default value (180)

src/server/svr_func.c
bayucan: Will do.

test/tests/functional/pbs_provisioning.py
bayucan: Ok, will do.

bayucan:
Besides addressing review comment, I had to also handle a race condition where a job that has a pending hook copy operation on a provisioned node, is deleted, causing a server crash. Culprit was a leftover task in the server (for checking if job has moved past the pending hook operation) operating on a freed pjob structure.

src/server/svr_func.c
```diff
@@ -5526,6 +5524,12 @@ prov_startjob(struct work_task *ptask)
 
 	assert(ptask->wt_parm1 != NULL);
 	pjob = (job *) ptask->wt_parm1;
+        if (pjob == NULL) {
```
hirenvadalia, line 5527: needs indentation fix here and below also

bayucan:
Thanks @hirenvadalia ! I've fixed the indentation issue.
@minghui-liu : I need your review.

PR #1932, opened 2020-07-25
Changed all strcpy to pbs_strncpy in scheduler

arungrover:
#### Describe Your Change
Using strcpy at times can be risky and cause problems like buffer overrun.

In this PR, I've added a generic pbs_strncpy function that always puts a '\0' at the end of the buffer.

Changed all strcpy in the scheduler module to pbs_strncpy.

In future, we need to also make changes to other modules.

#### Link to Design Doc
N/A

[smoketest.txt](https://github.com/openpbs/openpbs/files/4974728/smoketest.txt)

### DIFF
--- a/src/lib/Libutil/misc_utils.c
+++ b/src/lib/Libutil/misc_utils.c
@@ -334,6 +334,26 @@ pbs_strcpy(char *dest, const char *src)
 
 	return dest;
 }
+/**
+ *
+ * @brief general purpose strncpy function that will make sure to
+ *        copy '\0' at the end of the buffer.
+ *
+ * @param[in] dest - pointer to the destination buffer
+ * @param[in] src  - pointer to the source string
+ * @param[in] n    - size of destination buffer
+ *
+ * @return char *
+ * @retval pointer to the destination string
+ *
+ * @note: Caller needs ensure non-NULL pointers
+ */
+char *pbs_strncpy(char *dest, const char *src, size_t n)
+{
+    strncpy(dest, src, n-1);
+    dest[n-1] = '\0';
+    return dest;
+}
 
 /**
  * @brief
@@ -348,7 +368,6 @@ pbs_strcpy(char *dest, const char *src)
  * @retval pointer to *pbuf(the string pbuf points at) on successful read
  * @retval NULL on EOF or error
  */
-#define PBS_FGETS_LINE_LEN 8192
 char *
 pbs_fgets(char **pbuf, int *pbuf_size, FILE *fp)
 {
--- a/src/lib/Libutil/misc_utils.c
+++ b/src/lib/Libutil/misc_utils.c
@@ -350,8 +350,8 @@ pbs_strcpy(char *dest, const char *src)
  */
 char *pbs_strncpy(char *dest, const char *src, size_t n)
 {
-    strncpy(dest, src, n-1);
-    dest[n-1] = '\0';
+    strncpy(dest, src, n - 1);
+    dest[n - 1] = '\0';
     return dest;
 }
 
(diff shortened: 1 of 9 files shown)


agrawalravi90:
@arungrover  just curious, are we adding this because snprintf() can be slow?

src/lib/Libutil/misc_utils.c
```diff
@@ -334,6 +334,26 @@ pbs_strcpy(char *dest, const char *src)
...
+ *
+ * @return char *
+ * @retval pointer to the destination string
+ *
+ * @note: Caller needs ensure non-NULL pointers
+ */
+char *pbs_strncpy(char *dest, const char *src, size_t n)
+{
+    strncpy(dest, src, n-1);
+    dest[n-1] = '\0';
```
bhroam, line 354: Add spaces around the '-'

src/scheduler/fairshare.c
```diff
@@ -627,7 +627,7 @@ write_usage(char *filename, fairshare_head *fhead)
 	 * ...
 	 */
 
-	strcpy(head.tag, USAGE_MAGIC);
+	pbs_strncpy(head.tag, USAGE_MAGIC, 9);
```
bhroam, line 630: You use a combination of static numbers and macros for the size of pbs_strncpy().  Just use sizeof(dest) everywhere.

src/scheduler/parse.c
```diff
@@ -103,7 +103,7 @@ parse_config(char *fname)
 	FILE *fp;			/* file pointer to config file */
 	char *buf = NULL;
 	int buf_size = 0;
-	char buf2[8192];		/* general purpose buffer */
+	char buf2[PBS_FGETS_LINE_LEN];	/* general purpose buffer */
```
bhroam, line 106: Why PBS_FGETS_LINE_LEN?  This has nothing to do with fgets.

arungrover: You are right, this buffer isn't used with fgets. Reverting it.

src/scheduler/fairshare.c
arungrover: True, It didn't occur to me :)
I'll change it.

src/scheduler/pbs_sched.c
```diff
@@ -732,7 +732,7 @@ lock_out(int fds, int op)
 	flock.l_start  = 0;
 	flock.l_len    = 0;	/* whole file */
 	if (fcntl(fds, F_SETLK, &flock) < 0) {
-		pbs_strncpy(log_buffer, "pbs_sched: another scheduler running\n", LOG_BUF_SIZE);
+		pbs_strncpy(log_buffer, "pbs_sched: another scheduler running\n", sizeof(log_buffer));
 		log_err(errno, msg_daemonname, log_buffer);
```
bhroam, line 736: This is a static string, mind skipping the strcpy() and moving it directly into the log_err()?

arungrover: I've checked in all places, there isn't an instance where I'm doing sizeof on a pointer pointing to dynamically allocated memory.

arungrover:
sorry, I missed that @agrawalravi90 
Yes, I ran some tests and found that snprintf can be many times slower than strncpy. In one of the test, strncpy took 1 second to copy a small string 100 million times and snprintf took 10 seconds.

vstumpf:
We have a bunch of strncpys, shoudl we change those too?

arungrover:
@vstumpf while I agree with you that all calls should use one function, this particular PR was to address issues with strcpy only. Maybe we can change strncpys as and when we encounter them in future work.

arungrover:
@agrawalravi90 I've changed the description.

PR #1934, opened 2020-07-28
Fix habitat script missing pbs_db_utility

vstumpf:
#### Describe Bug or Feature
Execution rpm installations failed after #1641.
The habitat script expects a file to be present that is only present in the server rpm.

#### Describe Your Change
Only check for pbs_db_utility if PBS_START_SERVER=1 is set.

```
[vstumpf@shecil x86_64]$ sudo /etc/init.d/pbs start
Starting PBS
PBS Home directory /var/spool/pbs needs updating.
Running /opt/pbs/libexec/pbs_habitat to update it.
***
*** End of /opt/pbs/libexec/pbs_habitat
Home directory /var/spool/pbs updated.
PBS mom
[vstumpf@shecil x86_64]$ sudo /etc/init.d/pbs stop
```
I dont really think I need to write a test, so here's a manual version.

### DIFF
--- a/src/cmds/scripts/pbs_habitat.in
+++ b/src/cmds/scripts/pbs_habitat.in
@@ -244,13 +244,13 @@ if [ -f "$PBS_HOME/mom_priv/config" ]; then
 	done
 fi
 
-# Check for the db install script
-if [ ! -x "${PBS_EXEC}/libexec/pbs_db_utility" ]; then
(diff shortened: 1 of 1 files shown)


bayucan:
This looks the same fix as the one in https://github.com/openpbs/openpbs/pull/1935!

vstumpf:
Haha, oops, we'll use ashwath's

PR #1937, opened 2020-07-30
[WIP]DshUtil's function get_pbs_conf_file() incorrectly calls out to python

Klovely:
#### Describe Bug or Feature
DshUtil's function get_pbs_conf_file() incorrectly calls out to python

#### Describe Your Change
The following code is in DshUtils.get_pbs_conf_file():
` pc = ('"import os;print [False, os.environ[\'PBS_CONF_FILE\']]'

                  '[\'PBS_CONF_FILE\' in os.environ]"')

            cmd = [_'python'_, _'-c'_, pc]

            ret = self.run_cmd(hostname, cmd, logerr=False)

            if ((ret[_'rc'_] != 0) and (len(ret[_'out'_]) > 0) and

                    (ret[_'out'_][0] != 'False')):

                dflt_conf = ret[_'out'_][0]`
          

The whole list indexing idea is not needed.  os.environ is a dictionary.  Do a os.environ.get('PBS_CONF_FILE', False).  The second argument to get() is returned if the key doesn't exist.

### DIFF
--- a/test/fw/ptl/utils/pbs_dshutils.py
+++ b/test/fw/ptl/utils/pbs_dshutils.py
@@ -404,7 +404,7 @@ class DshUtils(object):
                 dflt_conf = ret['out'][0]
             """
             print("******Executing else from get_conf_file****")
-            val = os.environ.get('PBS_CONF_FILE', False)
+            value = os.environ.get('PBS_CONF_FILE', False)
             if value is not False:
                 dflt_conf = value
 
--- a/test/fw/ptl/utils/pbs_dshutils.py
+++ b/test/fw/ptl/utils/pbs_dshutils.py
@@ -404,7 +404,7 @@ class DshUtils(object):
                 dflt_conf = ret['out'][0]
             """
             print("******Executing else from get_conf_file****")
-            val = os.environ.geti('PBS_CONF_FILE', False)
+            val = os.environ.get('PBS_CONF_FILE', False)
             if value is not False:
                 dflt_conf = value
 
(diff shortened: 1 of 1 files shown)


test/fw/ptl/utils/pbs_dshutils.py
```diff
@@ -401,6 +402,11 @@ def get_pbs_conf_file(self, hostname=None):
             if ((ret['rc'] != 0) and (len(ret['out']) > 0) and
                     (ret['out'][0] != 'False')):
                 dflt_conf = ret['out'][0]
+            """
+            print("******Executing else from get_conf_file****")
```
hirenvadalia, line 406: @Klovely I know this is PR WIP, but I don't think we need to fix anything in this method as I don't see any problem with current code (unless I am missing something), but problem with fix. We need to get PBS_CONF_FILE value on given host (aka hostname) and it can be remote host also. But you current fix will always get PBS_CONF_FILE value from current host (where you are invoking pbs_benchpress).

Try your fix on this setup:
2 hosts, 1st host - server/sched/comm install, 2nd host mom install on non-default location and change rename /etc/pbs.conf to /etc/pbs_xyz.conf, and make sure if you do `ssh host2 echo \$PBS_CONF_FILE` it returns /etc/pbs_xyz.conf (you can use /etc/profile.d or ~/.bashrc to setup this)

Now invoke pbs_benchpress on host1 as `pbs_benchpress -t SmokeTest -p "servers=host1;nomom=host1;moms=host2"`

hirenvadalia:
@Klovely Are you still working on this PR? If not can you please close it?

hirenvadalia:
Closing inactive PR.

PR #1941, opened 2020-07-31
Submit test job with a run_count=18 so test_execjob_abort_ms_prologue…

lisa-altair:
… can be in the H state quickly

#### Describe Bug or Feature

TestPbsExecjobAbort.test_execjob_abort_ms_prologue sometimes fails when the machines are too overloaded and thus running more slowly.  The job is expected to be Held, but doesn't get to run enough times to be held before the expect checking hits the max attempts.

#### Describe Your Change

Increasing max_attempts to be higher than the current default of 180 was a possibility, but it seemed likely this timing issue would show up again.  It seemed like a better answer was to submit a run_count when submitting the job.  That way the test doesn't have to wait for many job runs for the run_count to equal 21 and be in the Held state.
Changing the job's starting run_count also allowed a 7 second offset to be removed.
Now the test will pass even faster on fast machines, and still pass on slow machines.

[2021.after.txt](https://github.com/openpbs/openpbs/files/5004601/2021.after.txt)

### DIFF
--- a/test/tests/functional/pbs_hook_execjob_abort.py
+++ b/test/tests/functional/pbs_hook_execjob_abort.py
@@ -133,15 +133,18 @@ time.sleep(2)
             hook_name, a, self.prolo_hook_body % (''))
         # Submit a job that eventually goes in H state
         start_time = time.time()
-        jid = self.server.submit(self.j)
+        j1attr = {ATTR_l + '.select': '3:ncpus=1', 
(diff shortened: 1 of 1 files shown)


bhroam:
@lisa-altair travis failed due to pycodestyle.

bayucan:
Merged #1941 into master.

PR #1945, opened 2020-08-01
Fixed issues when the scheduler got confused when a resv was running or not

bhroam:
#### Describe Bug or Feature
The scheduler was getting confused on whether a reservation was running or not.  If a reservation is running, it must place nodes on the nodes which are associated with the reservation.  If it isn't running, it can move anywhere.  The scheduler was getting confused that anytime a reservation was being altered, it was also "running"

#### Describe Your Change
Since we can't always detect if a reservation is running with its state/substate, we now detect it if the time is between the start/end of a reservation.  I use this instead of the state/substate to determine if a resv is running.

[ralter.log](https://github.com/openpbs/openpbs/files/5009447/ralter.log)
[resv.log](https://github.com/openpbs/openpbs/files/5009448/resv.log)

I reran the failed ralter test several times and it passed.  The 3 failed resv tests are expected and they are slated to be removed.

### DIFF
--- a/src/scheduler/check.c
+++ b/src/scheduler/check.c
@@ -1704,7 +1704,7 @@ should_check_resvs(server_info *sinfo, node_info *ninfo, resource_resv *job)
 
 	/* check if a job is in a reservation */
 	if (job->is_job && job->job != NULL && job->job->resv != NULL) {
-		if (job->job->resv->resv->resv_state == RESV_RUNNING) {
+		if (job->job->resv->resv->is_running) {
 			/* if we are not checking a specific node and the job is in a
 			 * running reservation, there can't be any conflicts
 			 */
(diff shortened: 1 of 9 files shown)


src/scheduler/data_types.h
```diff
@@ -748,6 +748,7 @@ struct node_info
 struct resv_info
 {
 	unsigned is_standing:1;		/* set to 1 for a standing reservation */
+	unsigned is_running;		/* the reservation is running (not necessarily in the running state) */
```
arungrover, line 751: Please make this a bit field.

src/scheduler/check.c
```diff
@@ -1989,7 +1989,12 @@ void get_resresv_spec(resource_resv *resresv, selspec **spec, place **pl)
 			*spec = resresv->select;
 		}
 	} else if (resresv->is_resv && resresv->resv != NULL) {
-		if (resresv->resv->resv_state == RESV_BEING_ALTERED || resresv->resv->resv_state == RESV_RUNNING)
+		/* The execselect should be used when the resv is running.  We can't
+		 * trust the state/substate to be RESV_RUNNING when a reservation is both
+		 * RESV_DEGRADED and RESV_BEING_ALTERED and is running.
+		 */
+		if (resresv->resv->is_running)
```
vstumpf, line 1996: should this also check RESV_BEING_ALTERED?

bhroam: No, we need to use the execselect only when the reservation is running.  It can be in BEING_ALTERED when it is confirmed.  In that case, the reservation can be confirmed anywhere.  There are 3 cases where we need to use the execselect (all with the reservation running):
1) state = RESV_RUNNING substate = RESV_DEGRADED
2) state = BEING_ALTERED substate = RESV_RUNNING
3) state = BEING_ALTERED substate = RESV_DEGRDED while the reservation is running.  The state/substate will be the same if the reservation is running or not.  I needed to come up with another way to determine if the reservation was running than its state/substate.

vstumpf: Oh, this is the actual reason for the PR. I see now.

PR #1948, opened 2020-08-03
Remove '_' from special charcter list in convert_to_cli method

shilpakodli:
#### Describe Bug or Feature
Tests on Cray were failing due to qsub error while submitting a job requesting vntype as "cray_compute"/"cray_login" nodes.
This is due to we are adding escape character "\" to all special characters in PTL fw, which are given in qsub and this is how qsub is getting submitted in PTL:
qsub -l vntype=cray_login -N cray\_login job.script
This special character list has underscore as well "_" which is not special character. Hence, qsub is failing with PbsSubmitError.

#### Describe Your Change
Removed underscore "_" from special character list in conver_to_cli method.

[After_fix.txt](https://github.com/openpbs/openpbs/files/5014651/After_fix.txt)
[Before_fix.txt](https://github.com/openpbs/openpbs/files/5014652/Before_fix.txt)

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -2391,7 +2391,7 @@ class BatchUtils(object):
                                       self.platform == 'craysim' or
                                       self.platform == 'shasta'):
                 v = v.translate({ord(c): "\\" +
-                                 c for c in r"~`!@#$%^&*()[]{};:,./<>?\|-=+"})
+                                 c for c in r"~`!@#$%^&*()[]{};:,/<>?\|="})
             if exclude_attrs is not None and a in exclude_attrs:
                 continue
 
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -2391,7 +2391,7 @@ class BatchUtils(object):
                                       self.platform == 'craysim' or
                                       self.platform == 'shasta'):
                 v = v.translate({ord(c): "\\" +
-                                 c for c in r"~`!@#$%^&*()[]{};:,./<>?\|-=_+"})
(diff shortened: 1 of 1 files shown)


test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -2391,7 +2391,7 @@ def convert_to_cli(self, attrs, op=None, hostname=None, dflt_conf=True,
                                       self.platform == 'craysim' or
                                       self.platform == 'shasta'):
                 v = v.translate({ord(c): "\\" +
-                                 c for c in r"~`!@#$%^&*()[]{};:,./<>?\|-=_+"})
+                                 c for c in r"~`!@#$%^&*()[]{};:,./<>?\|-=+"})
```
hirenvadalia, line 2394: Not sure but as per qsub code, we don't escape -N value while sending to server, so I am guessing this escaping is needed for python->shell when PTL runs qsub command, if that's true then we have `@`, `%`, `:`, `,`, `.`,`/`,`-`,`=`,`+` also in list which I don't think need escape in shell

shilpakodli: Hi @hirenvadalia, above code is for PTL where it runs on remote hosts. Where qsub command runs from a mom or a client.  In this case we need escape character to be appended for special characters.

shilpakodli: @hirenvadalia As discussed offline, I have verified characters you have mentioned by giving in qsub -N option. Allowed characters are "_ , . , + , - ". I have removed them from special character list.

shilpakodli:
Attaching latest execution logs.
[latest_execution_logs.txt](https://github.com/openpbs/openpbs/files/5033653/latest_execution_logs.txt)

PR #1952, opened 2020-08-04
Server should not contact Scheduler when scheduling is off

suresh-thelkar:
#### Describe Bug or Feature

Server should contact to Scheduler only when scheduling is true.

#### Describe Your Change

- At times when scheduling is off, Server keeps contacting Scheduler. This might result in eating up disk space due to which Server can totally become unresponsive.

- Following is a scenario where Server continuously tries to connect to Scheduler when scheduling is off.
_Scenario - If there is a high priority command like SCH_CONFIGURE pending by the time Scheduler goes down_
- However Server does not contact Scheduler if there is no high priority command in case of Scheduling is off.
This PR consists of code changes to make sure that Server doest not contact Scheduler when scheduling is off even in case of the above mentioned scenario.

Due to the lack of test case for the above said scenario this bug was un caught earlier.  Hence added a new PTL test case. Attached are results of this test case before and after fix.

Also removed couple of test cases which does not make any sense after this change.

[test_scheduling_behaviour_afix.log](https://github.com/openpbs/openpbs/files/5021259/test_scheduling_behaviour_afix.log)
[test_scheduling_behaviour_bfix.log](https://github.com/openpbs/openpbs/files/5021261/test_scheduling_behaviour_bfix.log)
Attached are the new PTL test case logs before and after fix.

Also ran TH3 after this change. All test cases passed.
image
image


bhroam:
I don't think this is the right direction to go in.  This is an edge case in the overall problem.  The overall problem is that if the scheduler is down, the server will try and constantly connect to it.  This will fill up the server log.  You also don't want to not send the SCH_CONFIGURE when something changes and the scheduler is up.  This will cause the scheduler to miss updates to the scheduler object.  The two daemons should always be in sync.  I think the real solution will be to start increasing the time between checks to the scheduler up to a max.  I think the current time between attempts is 2s.  Maybe do a doubling until 32s?

agrawalravi90:
weren't we going to change things so that scheduler reaches out to the server and not the other way around? Wouldn't that fix this issue as well?

bhroam:
If the scheduler contacts the server, we need a way to authenticate it as the scheduler rather than another client running as root/the daemon user.  Having the scheduler contact the server does solve a couple of problems on the scheduler side.  There are certain things it can't do until after the first cycle.

suresh-thelkar:
> I don't think this is the right direction to go in. This is an edge case in the overall problem. The overall problem is that if the scheduler is down, the server will try and constantly connect to it. This will fill up the server log. You also don't want to not send the SCH_CONFIGURE when something changes and the scheduler is up. This will cause the scheduler to miss updates to the scheduler object. The two daemons should always be in sync. I think the real solution will be to start increasing the time between checks to the scheduler up to a max. I think the current time between attempts is 2s. Maybe do a doubling until 32s?

Thanks @bhroam for giving us the insight behind Server contacting Scheduler even when scheduling is off in the above scenario. It looks like we cannot directly modify the function default_next_task for the following reason outlined in the comments.
        /*
	 * tilwhen is the basic "idle" time if there is nothing pending sooner
	 * for the Server (timed-events, call scheduler, IO)
	 * It used to be 10, but that caused a delay of outgoing TPP packets
	 * in some cases, and we don't burn too many extra cycles doing nothing
	 * if the delay is shorted to 2.
	 */
 
In order to achieve the solution that you suggested following is one of the ways. Please go through it and let me know if it looks good.
```
			for (psched = (pbs_sched*) GET_NEXT(svr_allscheds); psched; psched = (pbs_sched*) GET_NEXT(psched->sc_link)) {
			/* if time or event says to run scheduler, do it */
				
				curr_time = time(NULL);
				/* if we have a high prio sched command, send it 1st */
				if (psched->svr_do_sched_high != SCH_SCHEDULE_NULL) {
					if (curr_time - time_sched_conn > 0) {
						if (schedule_high(psched) == -1) {
							if (sched_conn_int <= 30)
								sched_conn_int += 2;
							time_sched_conn = curr_time + sched_conn_int;
						} else {
							sched_conn_int = 0;
							time_sched_conn = 0;	
						}
					}
				}
				if (psched->svr_do_schedule == SCH_SCHEDULE_RESTART_CYCLE) {
                                        ....				
					}
					psched->svr_do_schedule = SCH_SCHEDULE_NULL;
				} else if (svr_unsent_qrun_req || ((psched->svr_do_schedule != SCH_SCHEDULE_NULL) && can_schedule())) {
					/*
						* If svr_unsent_qrun_req is set to one there are pending qrun
						* request, then do schedule_jobs irrespective of the server scheduling
						* state.
						* If svr_unsent_qrun_req is not set then do the existing checking and do
						* scheduling only if server scheduling is turned on.
						*/
					int ret;

					psched->sch_next_schedule = time_now +
							psched->sch_attr[(int)	SCHED_ATR_schediteration].at_val.at_long;

					if (curr_time - time_sched_conn > 0) {
						if ((ret = schedule_jobs(psched)) == -1) {
							if (sched_conn_int <= 30)
								sched_conn_int += 2;
							time_sched_conn = curr_time + sched_conn_int;
						} else {
							sched_conn_int = 0;
							time_sched_conn = 0;	
						}
						if ((ret == 0) && (svr_unsent_qrun_req))
							svr_unsent_qrun_req = 0;
					}
				}
			}
```

agrawalravi90:
@suresh-thelkar what do you think about my suggestion to reverse the connection between server and scheduler so that server doesn't have to reach out to scheduler at all?

bhroam:
Maybe a hybrid approach.  When the server notices a scheduler is down, it stops trying to connect to it in the main loop,  It sets a timed task for like 5m in the future to check it so to not lose track of it.  When the scheduler comes up, it connects to the server to query all of its attributes.  While that connection is open, it can set the state back to idle.  The next time through the main loop, the state will be idle, and the server should pick back up.

suresh-thelkar:
> @suresh-thelkar what do you think about my suggestion to reverse the connection between server and scheduler so that server doesn't have to reach out to scheduler at all?

@agrawalravi90 , It looks like there are still some concerns if we change the direction of sched connection. This is what I understood from the Bhroam's comment https://github.com/openpbs/openpbs/pull/1952#issuecomment-668796649

And also if we follow the new approach suggested by Bhroam we can resolve the issue in hand.

suresh-thelkar:
> Maybe a hybrid approach. When the server notices a scheduler is down, it stops trying to connect to it in the main loop, It sets a timed task for like 5m in the future to check it so to not lose track of it. When the scheduler comes up, it connects to the server to query all of its attributes. While that connection is open, it can set the state back to idle. The next time through the main loop, the state will be idle, and the server should pick back up.

suresh-thelkar:
> Maybe a hybrid approach. When the server notices a scheduler is down, it stops trying to connect to it in the main loop, It sets a timed task for like 5m in the future to check it so to not lose track of it. When the scheduler comes up, it connects to the server to query all of its attributes. While that connection is open, it can set the state back to idle. The next time through the main loop, the state will be idle, and the server should pick back up.

Thanks @bhroam. I like this hybrid approach and proceed accordingly. When

subhasisb:
> If the scheduler contacts the server, we need a way to authenticate it as the scheduler rather than another client running as root/the daemon user. Having the scheduler contact the server does solve a couple of problems on the scheduler side. There are certain things it can't do until after the first cycle.

I like this approach. I think this is the best final solution for many problems. Given that there is work going on to start the scheduler as a specific configured user, it becomes easy to identify the scheduler connection. (i.e., if the connection is from user that is same as the configured scheduler user, then that connection is deemed as the scheduler) - of course, the admin has to ensure that the scheduler account itself is protected and no arbitrary user can gain access to that account.

Instead of taking stop-gap measure of this kind, we should simply flip the scheduler -  just like the moms now connect to the server than the other way around.

subhasisb:
I suggest we shelve this change for now and wait for the non-root scheduler work to go forward - things can become easier to implement after that

bhroam:
@subhasisb I have a doubt with your solution.  The doubt is that root can be the scheduler user (and is by default).  This would mean we'd authenticate all root connections as the scheduler.  This isn't right because any admin qstat would look like it is coming from the scheduler.  To do this properly, we need to have some way to authenticate as the scheduler.  Maybe we extend pbs_connect() to understand that we're the scheduler (or some other IFL call).  The server only authenticates as the scheduler if the connection is coming from root or the scheduler user.

subhasisb:
> @subhasisb I have a doubt with your solution. The doubt is that root can be the scheduler user (and is by default). This would mean we'd authenticate all root connections as the scheduler. This isn't right because any admin qstat would look like it is coming from the scheduler. To do this properly, we need to have some way to authenticate as the scheduler. Maybe we extend pbs_connect() to understand that we're the scheduler (or some other IFL call). The server only authenticates as the scheduler if the connection is coming from root or the scheduler user.

You are right. However that is about "identification" as opposed to classic "authentication" though they have an intersection. What i mean is that if root "wants" to act as a scheduler, it is okay - as in, if the admin wants to play nasty, things would go down anyway. So, we need to add a "flag" of some kind to the scheduler connection to identify that "i am the scheduler" - that is the identification part. Whether the pbs_server believes that is true or not, is based on the authentication of the user - so if the connection is from "a designated scheduler user" or "root" then we will believe that flag, else we will assume somebody trying to fake it.

suresh-thelkar:
> You are right. However that is about "identification" as opposed to classic "authentication" though they have an intersection. Exactly, what you mentioned! Connect should add a flag saying "i am scheduler" (identification) which can be be trusted only if the user is "authenticated" as either the "designated scheduler user" or "root"

@subhasisb, Thanks. I am assuming the flag should also contain "scheduler_name+ port+ primary/secondary" which can either be sent via an existing IFL or a new IFL.  This information is useful for the Server to identify which scheduler it is and whether it is primary or secondary connection.

If the user authenticated is not a "designated scheduler user" i.e. root by default, can we also add SHA256(some key) to the above information("scheduler_name+ port+ primary/secondary") for better security or is it an overkill. Just curious to know.

subhasisb:
> > You are right. However that is about "identification" as opposed to classic "authentication" though they have an intersection. Exactly, what you mentioned! Connect should add a flag saying "i am scheduler" (identification) which can be be trusted only if the user is "authenticated" as either the "designated scheduler user" or "root"
> 
> @subhasisb, Thanks. I am assuming the flag should also contain "scheduler_name+ port+ primary/secondary" which can either be sent via an existing IFL or a new IFL. This information is useful for the Server to identify which scheduler it is and whether it is primary or secondary connection.
> 
> If the user authenticated is not a "designated scheduler user" i.e. root by default, can we also add SHA256(some key) to the above information("scheduler_name+ port+ primary/secondary") for better security or is it an overkill. Just curious to know.

We can add any information we want when we extend the protocol of connection. I dont think the sha256 checksum solves anything. If somebody can change the header, they can replace the checksum too, for example a MITM attack case. The only sure-shot way to thwart that attempt is by using encryption (if MITM is actually of concern for the site)

suresh-thelkar:
> We can add any information we want when we extend the protocol of connection. I dont think the sha256 checksum solves anything. If somebody can change the header, they can replace the checksum too, for example a MITM attack case. The only sure-shot way to thwart that attempt is by using encryption (if MITM is actually of concern for the site)

@subhasisb, Thanks for the explaining MITM attack case and the solution for it.

bhroam:
@subhasisb I see your point.  I agree with you that we should only care about identification, not authentication.  If the site wants a root/sched user connection to act as the scheduler, they are really only confusing themselves.  I thought you wanted all connections by the sched user to be considered as the scheduler.  That could lead to confusion.  One way to do this without creating a new IFL call is to call pbs_connect_extend() directly and pass in something with the extend field.  I don't know what that extend field is currently being used for, so this might not work

subhasisb:
> @subhasisb I see your point. I agree with you that we should only care about identification, not authentication. If the site wants a root/sched user connection to act as the scheduler, they are really only confusing themselves. I thought you wanted all connections by the sched user to be considered as the scheduler. That could lead to confusion. One way to do this without creating a new IFL call is to call pbs_connect_extend() directly and pass in something with the extend field. I don't know what that extend field is currently being used for, so this might not work

agreed, sounds reasonable to me to try to first reuse the extend parameter

agrawalravi90:
@suresh-thelkar is this PR still relevant, or has this been taken over by https://github.com/openpbs/openpbs/pull/2000 ?

suresh-thelkar:
> #2000

@agrawalravi90, we have kept it open to get more discussions on to it. Now that we got enough information on this and this can be taken over by #2000, I am going to close this.

suresh-thelkar:
This is addressed as part of #2000.

PR #1953, opened 2020-08-05
Reduce use of malloc/memcpy in DIS routines and support for part reply send in statjob or selstat

hirenvadalia:
#### Describe Bug or Feature

* Currently DIS routines does multiple copy of data with malloc and memcpy, which is redundant as we already have buffer allocated (tdis_thebuf)
* When sending large data (like stat of large number jobs), Server tries to send data in one shot, which is time consuming + mem hungry as DIS will allocate buffer for whole reply which is will millions of bytes of data + while server is generating data client is sitting ideal doing nothing, instead this we can send status data in chunks and let client decode it while server is preparing next chunk, this will help in small DIS buffer + less data in one send + both server/client are doing some work at a time
* Servers fails to write when sending large data due to 10sec of timeout in tcp_send(), and due to this client hangs forever on waiting for data
* Currently on client side, we read data first in brp_cmdstat struct then in PBSD_status_get we convert brp_cmdstat to batch_status, which is redundant as we can directly read data in batch_status

#### Describe Your Change

* Improved DIS routines not to make copy of given data or make only absolutely needed copy (like when encryption is enabled)
* Implemented part send status data for statjob and selstat
* Improved PBSD_status_get and remove brp_cmdstat by using batch_status instead while reading data from DIS

#### Link to Design Doc

* None

* 1M jobs queued

||time qstat >/dev/null|time qstat -f >/dev/null|
|-|-|-|
|19.2|real    0m6.314s<br>user    0m3.163s<br>sys    0m0.406s|real    1m8.033s<br>user    0m54.410s<br>sys    0m6.895s|
|Before fix|qstat hangs|qstat hangs|
|Before fix (+ removing 10s timeout in tcp_send)|real    0m9.237s<br>user    0m3.836s<br>sys    0m0.390s|real    1m24.364s<br>user    1m1.697s<br>sys    0m6.226s|
|After fix|real    0m4.071s<br>user    0m3.605s<br>sys    0m0.443s|real    1m2.617s<br>user    0m56.562s<br>sys    0m5.891s|

Profile: regression
|--|--|--|--|--|--|--|
|2999|1788|2|0|0|0|1786|

(NOTE: 2 failures are known issue)

* valgrind result coming soon


hirenvadalia:
Latest tests results:

Description: Rerun All Tests From #2999
Profile: regression
|--|--|--|--|--|--|--|
|3012|1788|2|0|0|0|1786|

hirenvadalia:
Valgrind results: [valgrind.txt](https://github.com/openpbs/openpbs/files/5032843/valgrind.txt)

PR #1955, opened 2020-08-05
Add possibility of resuming multinode jobs

zv0n:
#### Describe Bug or Feature
I have added possibility of resuming jobs that are running on multiple nodes. It is already possible to resume a job running on single node after mom restarts, so allowing the same for multinode jobs seems like a good idea. This is done when `-p` flag is passed on command line.

Possible problems with the implementation:
  - If you run `pbsdsh` while one of the nodes is down it enters an infinite loop as it doesn't check for node availability
  - If you restart mother superior while `pbsdsh` is running, it will exit with error as it loses connection with mother superior
  - Starting `pbsdsh` right after a node restart may result in error on task spawn because the task can be sent to the node before the node rejoins the job (and receives ports through which it should communicate)

#### Describe Your Change
  - Mother superior now saves and recovers `ji_ports`, so it can communicate with demux program after recovering the job.
  - When a non-superior mom recovers multinode job, it sends `IM_RECONNECT_TO_MS` to mother superior, MS then sends `IM_JOIN_RECOV_JOB` to all sisters along with port numbers through which they should communicate
  - When MS recovers multinode job, it sends `IM_JOIN_RECOV_JOB` along with port numbers to all sisters
  - Obit information (if there is any) is saved/recover along with tasks, so moms know wheter to send an obit after task completion
  - Multinode jobs now check whether they have a connection to MS inside `scan_for_exiting`, so when they send an obit they can be sure that it is delivered

[test_multinode_restart.txt](https://github.com/openpbs/openpbs/files/5061012/test_multinode_restart.txt)

### DIFF
--- a/src/include/job.h
+++ b/src/include/job.h
@@ -492,6 +492,8 @@ struct job {
 	int ji_stderr;				    /* socket for stderr */
 	int ji_ports[2];			    /* ports for stdout/err */
 	enum bg_hook_request ji_hook_running_bg_on; /* set when hook starts in the background*/
+	int		ji_msconnected; /* 0 - not connected, 1 - connected */
+	pbs_list_head	ji_multinodejobs;	/* links to recovered multinode jobs */
 #else						    /* END Mom ONLY -  start Server ONLY */
 	struct batch_request *ji_pmt_preq; /* outstanding preempt job request for deleting jobs */
 	int ji_discarding;		   /* discarding job */
@@ -592,6 +594,8 @@ struct job {
 #ifdef PBS_MOM
 			tm_host_id ji_nodeidx; /* my node id */
 			tm_task_id ji_taskidx; /* generate task id's for job */
+			int ji_stdout;
+			int ji_stderr;
 #if MOM_ALPS
 			long ji_reservation;
 			/* ALPS reservation identifier */
@@ -745,6 +749,8 @@ typedef struct	infoent {
 #define IM_EXEC_PROLOGUE	24
 #define IM_CRED 		25
 #define IM_PMIX			26
+#define IM_RECONNECT_TO_MS			27
+#define IM_JOIN_RECOV_JOB		28
 
 #define IM_ERROR		99
 #define IM_ERROR2		100
--- a/src/resmom/mom_comm.c
+++ b/src/resmom/mom_comm.c
@@ -45,6 +45,7 @@
 #include	<assert.h>
 #include	<stdio.h>
 #include	<stdlib.h>
+#include	<sys/stat.h>
 
 #include	<unistd.h>
 #include	<dirent.h>
@@ -117,6 +118,7 @@ extern  char   *msg_err_malloc;
 extern int
 write_pipe_data(int upfds, void *data, int data_size);
 char	task_fmt[] = "/%8.8X";
+extern void resume_multinode(job *pjob);
 
 
 /* Function pointers
@@ -2315,7 +2317,7 @@ term_job(job *pjob)
 	int	num;
 
 	for (num=0, np = pjob->ji_hosts;
-		num<pjob->ji_numnodes;
+		num < pjob->ji_numnodes;
 		num++, np++) {
 		if (np->hn_stream >= 0) {
 			np->hn_stream = -1;
@@ -2380,6 +2382,7 @@ im_eof(int stream, int ret)
 			np->hn_stream = -1;
 			if (np->hn_eof_ts == 0)
 				np->hn_eof_ts = time(0);
+			pjob->ji_msconnected = 0;
 
 			/*
 			 ** In case connection to pbs_comm is down/recently established, do not kill a job that is actually running.
@@ -2493,6 +2496,7 @@ check_ms(int stream, job *pjob)
 		np->hn_stream = stream;
 	}
 	np->hn_eof_ts = 0;
+	pjob->ji_msconnected = 1;
 	return FALSE;
 }
 
@@ -3020,6 +3024,27 @@ im_request(int stream, int version)
 	BAIL("fromtask")
 	switch (command) {
 
+		case IM_JOIN_RECOV_JOB:
+			reply = 1;
+
+			hnodenum = disrsi(stream, &ret);
+			BAIL("JOINJOB nodenum")
+
+			np = NULL;
+			/* job should already exist */
+			pjob = find_job(jobid);
+			if( pjob == NULL ) {
+				SEND_ERR(PBSE_SYSTEM)
+				goto done;
+			}
+			pjob->ji_stdout = disrsi(stream, &ret);
+			BAIL("JOINJOB stdout")
+			pjob->ji_stderr = disrsi(stream, &ret);
+			BAIL("JOINJOB stderr")
+			pjob->ji_qs.ji_un.ji_momt.ji_exuid = pjob->ji_grpcache->gc_uid;
+			pjob->ji_qs.ji_un.ji_momt.ji_exgid = pjob->ji_grpcache->gc_gid;
+			pjob->ji_msconnected = 1;
+			goto done;
 		case IM_JOIN_JOB:
 			/*
 			 ** Sender is mom superior sending a job structure to me.
@@ -3064,6 +3089,7 @@ im_request(int stream, int version)
 				info = disrcs(stream, &len, &ret);
 				BAIL("JOINJOB credential")
 			}
+			pjob->ji_msconnected = 1;
 
 			pjob->ji_numnodes = hnodenum;
 			CLEAR_HEAD(lhead);
@@ -4030,6 +4056,7 @@ join_err:
 				op->oe_u.oe_tm.oe_node = pvnodeid;
 				op->oe_u.oe_tm.oe_event = event;
 				op->oe_u.oe_tm.oe_taskid = fromtask;
+				task_save(ptask);
 				reply = 0;
 			}
 			break;
@@ -5402,6 +5429,10 @@ join_err:
 				event, fromtask, IM_OLD_PROTOCOL_VER);
 			if (ret != DIS_SUCCESS)
 				goto err;
+			break;
+		case IM_RECONNECT_TO_MS:
+			if (pjob->ji_qs.ji_svrflags & JOB_SVFLG_HERE)
+				resume_multinode(pjob);
 			break;
 
 		default:
--- a/src/resmom/start_exec.c
+++ b/src/resmom/start_exec.c
@@ -5961,7 +5961,7 @@ void resume_multinode(job *pjob)
 
 		int stream = np->hn_stream;
 		im_compose(stream, pjob->ji_qs.ji_jobid,
-			pjob->ji_wattr[(int)JOB_ATR_Cookie].at_val.at_str,
+			get_jattr_str(pjob, JOB_ATR_Cookie),
 			com, ep->ee_event, TM_NULL_TASK,  IM_OLD_PROTOCOL_VER);
 		(void)diswsi(stream, pjob->ji_numnodes);
 		(void)diswsi(stream, pjob->ji_ports[0]);
--- a/src/resmom/start_exec.c
+++ b/src/resmom/start_exec.c
@@ -5929,6 +5929,52 @@ job_nodes(struct job *pjob)
 
 /**
  * @brief
+ * 	Resume multinode job after one or more sisters has been restarted
+ *
+ * @param[in] pjob - job pointer
+ *
+ * @return	Void
+ *
+ */
+
+void resume_multinode(job *pjob)
+{
+	if (pjob->ji_hosts == NULL)
+		return;
+
+	int com = IM_JOIN_RECOV_JOB;
+	hnodent *np = NULL;
+	eventent *ep = NULL;
+	int i;
+	for(i = 1; i < pjob->ji_numnodes; i++) {
+		np = &pjob->ji_hosts[i];
+
+		if( i == 1 )
+			ep = event_alloc(pjob, com, -1, np, TM_NULL_EVENT, TM_NULL_TASK);
+		else
+			ep = event_dup(ep, pjob, np);
+
+		if (ep == NULL) {
+			exec_bail(pjob, JOB_EXEC_FAIL1, NULL);
+			return;
+		}
+
+		int stream = np->hn_stream;
+		im_compose(stream, pjob->ji_qs.ji_jobid,
+			pjob->ji_wattr[(int)JOB_ATR_Cookie].at_val.at_str,
+			com, ep->ee_event, TM_NULL_TASK,  IM_OLD_PROTOCOL_VER);
+		(void)diswsi(stream, pjob->ji_numnodes);
+		(void)diswsi(stream, pjob->ji_ports[0]);
+		(void)diswsi(stream, pjob->ji_ports[1]);
+		dis_flush(stream);
+#if defined(PBS_SECURITY) && (PBS_SECURITY == KRB5)
+		send_cred_sisters(pjob);
+#endif
+	}
+}
+
+/**
+ * @brief
  * 	start_exec() - start execution of a job
  *
  * @param[in] pjob - job pointer
@@ -6189,6 +6235,8 @@ start_exec(job *pjob)
 			}
 			pjob->ji_stdout = socks[0];
 			pjob->ji_stderr = socks[1];
+			pjob->ji_extended.ji_ext.ji_stdout = pjob->ji_ports[0];
+			pjob->ji_extended.ji_ext.ji_stderr = pjob->ji_ports[1];
 		}
 
 		for (i = 1; i < nodenum; i++) {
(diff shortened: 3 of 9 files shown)


zv0n:
Hi,
Can someone please remove the random redundant Travis job (that doesn't seem to be starting at all)?
Thanks!

arungrover:
Hi @zv0n, I think I have seen this problem with Travis in the past. What might work is that either you can close the PR and reopen it OR force push your branch. In both these cases, Travis will run again but I think it will get rid of that redundant job.

zv0n:
Thanks @arungrover, seems like force pushing fixed it!

zv0n:
Can someone take a look at this change, please?

src/include/job.h
```diff
@@ -501,6 +501,8 @@ struct job {
 	int		ji_stderr;	/* socket for stderr */
 	int		ji_ports[2];	/* ports for stdout/err */
 	enum	bg_hook_request	ji_hook_running_bg_on; /* set when hook starts in the background*/
+	int		ji_msconnected; /* 0 - not connected, 1 - connected */
+	pbs_list_head	ji_multinodejobs;	/* links to recovered multinode jobs */
```
bayucan, line 505: Please initialize ji_msconnected in server/job_func.c:job_alloc() under #ifdef PBS_MOM section.
Initialize also ji_multinodejobs in server/job_func.c:job_alloc() by calling CLEAR_HEAD(ji_multinodejobs).

test/tests/functional/pbs_node_jobs_restart_multinode.py
```diff
@@ -0,0 +1,118 @@
...
+
+        if len(self.moms) != 2:
+            self.skipTest("test requires atleast two MoMs as input, " +
+                          "use -p moms=<mom1:mom2>")
+        momA = self.moms.values()[0]
+        momB = self.moms.values()[1]
+
+        # Make sure moms are running with -p flag
+        momA.restart(args=['-p'])
+        momB.restart(args=['-p'])
```
bayucan, line 61: What happens if -p is not specified by at least one of the moms? Please add a test case for this.

test/tests/functional/pbs_node_jobs_restart_multinode.py
```diff
@@ -0,0 +1,118 @@
...
+        script = "%s sleep 120" % pbsdsh_path
+        j = Job(TEST_USER)
+        j.set_attributes({'Resource_List.select': '2',
+                          'Resource_List.place': 'scatter'})
+        j.create_script(script)
+        jid = self.server.submit(j)
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid)
+
+        # only restart 1 node as restarting momA would result in a pbsdsh crash
+        momB.restart(args=['-p'])
```
bayucan, line 74: Please add a test case where momB is prematurely killed, as in mom.signal("-KILL"), and then issue a mom.restart("-p"). Also a case where primary mom is also killed prematurely, and then restart with -p. Is the behavior as you expected?

src/resmom/mom_comm.c
```diff
@@ -217,6 +219,41 @@ pbs_jobnodevoid_t	job_free_node = NULL;
...
+ * @retval  -1 Failure
+ *
+ */
+int
+ensure_write(int fd, char *write_ptr, int size)
+{
+	int i = 0;
+	while ((i = write(fd, write_ptr, size)) != size) {
+		if ((i < 0) && (errno == EINTR)) {	/* retry the write */
+			if (lseek(fd, (off_t)0, SEEK_SET) < 0) {
```
hirenvadalia, line 241: Doing lseek 0 here can lead to corrupted data in .TK file...
For example, if ensure_write is called for `obit->oe_u.oe_tm.oe_node` (4th write in file), and if we do lseek 0 then will loose data for `ptask->ti_qs`, `pobit->oe_type`, `pobit->oe_u.oe_tm.oe_fd`

src/resmom/mom_comm.c
```diff
@@ -278,8 +315,26 @@ task_save(pbs_task *ptask)
 			return (-1);
 		}
 	}
+	obitent *pobit = (obitent *)GET_NEXT(ptask->ti_obits);
```
hirenvadalia, line 318: I am not sure but ptask->ti_obits is list, and we are only storing first? don't we need to store others in list?

zv0n:
> @zv0n : Do you also have an accompanying design for this? It would be better to add a design under https://openpbs.atlassian.net/wiki/spaces/PD/pages outlining your approach of when pbs_mom is started with -p option, multinode jobs can also be recovered. Then announce it to the community form: https://community.openpbs.org/ for comments. People in the community can chime in and suggest improvements to the design...

@bayucan I don't have a design, sorry. I did this because we needed this functionality at my workplace, so I started coding before I looked at how merging to the project works. I'll go ahead and create one now.

src/include/job.h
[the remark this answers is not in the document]
zv0n: I believe I'm already initializing these variables in job_func.c:355,356, should I do something more than that?

Since ji_multinodejobs only contains pointers to jobs that are already managed, will CLEAR_HEAD(ji_multinodejobs) suffice in job_free()?

src/include/job.h
```diff
@@ -612,6 +614,8 @@ struct job {
 #ifdef PBS_MOM
 			tm_host_id ji_nodeidx;	/* my node id */
 			tm_task_id ji_taskidx;	/* generate task id's for job */
+			int ji_stdout;
+			int ji_stderr;
```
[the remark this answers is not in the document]
zv0n: No other vars in ji_extended are allocated in job_alloc(), should I really do that?

src/resmom/mom_comm.c
zv0n: You're probably right, I thought that obits are only created when MS first sends a task, but now I see they can be created in other ways as well. I'll rethink how the obits are saved, thanks for the catch!

src/resmom/mom_comm.c
```diff
@@ -3186,6 +3289,9 @@ im_request(int stream, int version)
 
 			/* np is set from job_nodes_inner */
 
+			if (check_ms(stream, pjob))
```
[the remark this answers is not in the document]
zv0n: Sorry, I forgot to revert this change (I did it during development when connection to MS was not working and I was trying various fixes). I'll revert it, but I think the early check isn't really needed, as it was running check_ms(stream, NULL) and check_ms always returns FALSE when pjob == NULL, so it seems like useless line to me

src/resmom/mom_comm.c
```diff
@@ -278,8 +317,43 @@ task_save(pbs_task *ptask)
...
+	unsigned int obitcount = 0;
+	if (ensure_write(fds, (char *)&obitcount, sizeof(unsigned int)) != 0)
+		goto failedwrite;
+
+	obitent *pobit = (obitent *)GET_NEXT(ptask->ti_obits);
+	while (pobit) {
+		obitcount++;
+		if (ensure_write(fds, (char *)&pobit->oe_type, sizeof(pobit->oe_type)) != 0)
+			goto failedwrite;
+		if (ensure_write(fds, (char *)&pobit->oe_u.oe_tm.oe_fd, sizeof(pobit->oe_u.oe_tm.oe_fd)) != 0)
```
hirenvadalia, line 336: @bayucan Please correct me if I am wrong...
but @zv0n I don't think saving obits help us regain connection with clients waiting for obits as if mom process goes then oe_fd will be released and clients (aka pbsdsh or pbs_tmrsh) connected to oe_fd will get disconnect error right? like below:
```
pbsroot@node1:~$ qsub -I
qsub: waiting for job 2.node1 to start
qsub: job 2.node1 ready

pbsroot@node1:~$ pbsdsh sleep 100 
pbsdsh: Event poll failed, error not connected (17002)
```
I just killed pbs_mom when sleep 100 process started

zv0n: Yes, that's one of the issues I've written about in the initial pull request comment, pbsdsh fails when mother superior is restarted. But if you restart one of the sisters, pbsdsh continues going and expects obit from the sisters, so if we don't save the obit information, it would result in an infinite loop within pbsdsh.

In this commit I wanted to create server-side (mom) support for the multi-node resume and thought client side might be done in a different commit.

This problem is within pbsdsh only and I think it should be fixable, for example mpirun (at least during my testing) didn't error out during MS restart and managed to reconnect and after receiving obits terminate successfuly.

hirenvadalia: So if I understand correctly @zv0n, then if we fix clients not to error out and wait for MS or sister nodes to come up and reconnect, then also we don't need save obit info, because once clients reconnects it will do tm_obit (to register obit) and tm_poll (wait for obit to finish) again right?

zv0n: @hirenvadalia Yes, that sounds right. If the clients can recognize when a node reconnects, so they can delete the old event and register a new obit, that would remove the need to save obits.

hirenvadalia: exactly @zv0n!, only thing is clients has to try (may be forever or for some timeout) reconnecting to node... and once connection re-establish, no need to delete old event, just register obit for given task and wait for its finish

hirenvadalia: Also doing above will eliminate all possible problems which you mentioned in PR description...

zv0n: Wouldn't the old event need to be deleted? `tm_obit` calls `add_event` inside the function body, which allocates a new event, wouldn't that be a problem? If we don't want to delete the old event I think we might need to add a new function that only re-sends the obit number without creating a new event.
I've only skimmed through the function, so I might be wrong about how it works though.

hirenvadalia: Since obits are not saved in .TK file so on mom restart mom wouldn't have that information, if we save obits in .TK and recover that on mom restart, then only we need to delete that, but as we don't need to save obits so no need of deleting it... HTH!

zv0n: I see! Great, thanks!

zv0n: So, @hirenvadalia, do you think I should remove the obit-saving parts and leave this PR as is? Because so far I have no idea how to make pbsdsh operate without an open socket to MS 😅

src/include/job.h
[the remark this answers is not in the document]
bayucan: I missed that you were already initializing them in job_func.c under job_alloc() so that's good. For the job_free(), yes, just do the CLEAR_HEAD.

bayucan:
Thanks @zv0n for creating the design.

subhasisb:
hi @zv0n I was looking at your proposal. I looked at your design page, but could not go to the associated design discussion page - it shows "oops topic not found". Hence adding my thoughts here;

I was trying to understand your primary usecase to make this change:
Are you trying to handle the case where only pbs_mom is restarted/upgraded? Or for a host level restart. If it is a host level restart/OS-update etc that needs a reboot, pbsdsh, or MPI job will typically die anyway, so allowing mom to restart with -p to preserve jobs is not useful there. 
PBS supports ways to make node maintenance easy, by allowing to turn nodes offline to drain off jobs from them, or to enable admin-suspend: maintenance_jobs in the guide: https://www.altair.com/pdfs/pbsworks/PBS19.2.3_BigBook.pdf

Also, there is a thread about new maintenance windows enhancement: 
https://openpbs.atlassian.net/wiki/spaces/PD/pages/879493121/Node+maintenance+window+enhancement

Please look through these and see if they satisfy your use cases already or not?

I do realize that your effort is to allow restarting of mom while preserving multinode jobs. The part i am trying to understand is that "is it really necessary, when there are ways to take a node down for maintenance in an orderly fashion anyway?"

vchlum:
Hi @subhasisb , my colleague @zv0n works on this because we would appreciate this feature/fix for the upgrading process, and I think this could help others too.

Of course, this can not cover the host-level restart need. In such a case, the maintenance windows are suitable.

Our current solution: When we need to update the PBS service on nodes (because applying some patch or bug fix), we often do the "live update". Right now, we have the following patch in use: After upgrading the PBS package on the system, a signal is sent to mom. The mom will wait until no more multi-node jobs are running on the node. Once no multi-node job is present the self restart is engaged (with -p). This has obviously a huge drawback - the restart may never happen because a new multi-node job could start before the last one is finished. Not ideal.

I think it would be good and useful for all to be able to update (minor? versions of) PBS without setting any maintenance. We could just install a new version of the PBS package and the post-install could initiate the restart and that's it. This "live update" would be much simpler for admins. It would also increase the utilization of nodes because you do not need to drain nodes before updating. So, no resources are wasted during the PBS package update process.

subhasisb:
> Our current solution: When we need to update the PBS service on nodes (because applying some patch or bug fix), we often do the "live update". Right now, we have the following patch in use: After upgrading the PBS package on the system, a signal is sent to mom. The mom will wait until no more multi-node jobs are running on the node. Once no multi-node job is present the self restart is engaged (with -p). This has obviously a huge drawback - the restart may never happen because a new multi-node job could start before the last one is finished. Not ideal.

Actually, if you set the node to "offline" no new jobs will be sent to the mom. So eventually, the node will be free. I do understand  that even that could take a good amount of time depending on how long the currently running job takes to complete.

> I think it would be good and useful for all to be able to update (minor? versions of) PBS without setting any maintenance. We could just install a new version of the PBS package and the post-install could initiate the restart and that's it. This "live update" would be much simpler for admins. It would also increase the utilization of nodes because you do not need to drain nodes before updating. So, no resources are wasted during the PBS package update process.

yes, this, of course, makes sense. I was just thinking whether lots of sites really need this. But, i understand if you do.

bhroam:
@vchlum another one of our customers handles rolling updates with provisioning.  They even handle OS updates this way.  They set the default aoe of the new jobs to the new node image.  Eventually a new job will run with the new aoe and that will trigger the node to be provisioned.  This will work better for them than for you because they allocate their nodes exclusively.  In your case, the scheduler will only run jobs with shared aoe values on the same nodes.  Eventually all of the older aoe jobs will be finished and a new aoe will run.  The node will eventually provision.

vchlum:
Thx @bhroam for sharing the solution. A solution such as this is not suitable for our site. We have heterogeneous nodes and heterogeneous jobs. We allow up to one-month walltime.

Since the -p parameter already works for singlenode job, I also think this is a logical improvement.

src/resmom/mom_comm.c
hirenvadalia: @zv0n, sorry for delay, I think we can proceed with this PR after removing obit-saving parts but without fixing pbsdsh/pbs_tmrsh, will not get full resume of multinode.
Based on your one of above comment, where you say mpirun was able to reconnect on MS restart and continue, so if we don't fix pbsdsh and pbs_tmrsh then only multinode jobs which uses mpirun part will able resume i think.

zv0n:
My forum post got approved, so we can move future discussion there if need be - https://community.openpbs.org/t/option-to-resume-multinode-jobs/2244/2

src/resmom/mom_comm.c
zv0n: @hirenvadalia I've removed the obit saving bits and did a test run, as you expected mpirun works without a problem, pbsdsh gets stuck on waiting for obit.

hirenvadalia: Thanks @zv0n... will continue code and design review...

zv0n:
Forgot about that, sorry, it's removed now

hirenvadalia (approved):
@zv0n I don't have any further comments, but you still have few open question/comments from @bayucan. Also you need to update design page as suggestion by @bayucan.

zv0n:
Hi, sorry for the long delay, I was on vacation.
@bayucan I think I've fixed all problems with code you've found and just added your requested tests. I've removed the test which restarted during pbsdsh since we've decided to remove obit saving and that test would either fail or enter an infinite loop

bayucan:
@zv0n : It looks good now. Travis check seemed to be stuck. Please rebase your branch to the latest master.

zv0n:
@bayucan I've rebased to master, hopefully Travis check will be happy and cooperative this time around :D

src/resmom/start_exec.c
```diff
@@ -5927,6 +5927,52 @@ job_nodes(struct job *pjob)
...
+			ep = event_dup(ep, pjob, np);
+
+		if (ep == NULL) {
+			exec_bail(pjob, JOB_EXEC_FAIL1, NULL);
+			return;
+		}
+
+		int stream = np->hn_stream;
+		im_compose(stream, pjob->ji_qs.ji_jobid,
+			pjob->ji_wattr[(int)JOB_ATR_Cookie].at_val.at_str,
```
hirenvadalia, line 5964: @zv0n sorry but one very small change: (as now things have changed in mainline about how we access job attributes using getters/setters):
Can you please repalce this line with `get_jattr_str(pjob, JOB_ATR_Cookie)`?

hirenvadalia (approved):
Thanks @zv0n. LGTM.

bayucan (approved):
Code looks good.

PR #1960, opened 2020-08-08
Multi sched performance test

visheshh:
#### Describe Bug or Feature
Added multisched performance test

#### Describe Your Change
Added multisched performance test to measure time taken by single scheduler to schedule 5k jobs and time taken by 5 schedulers when workload is divided.


test/tests/performance/pbs_multi_sched_perf.py
```diff
@@ -0,0 +1,169 @@
...
+        return ret_jids
+
+    @timeout(3600)
+    def test_multi_sched_perf(self):
+        """
+        Test time taken to schedule and run 5k jobs with
+        single scheduler and workload divided among 5 schedulers.
+        """
+        a = {'resources_available.ncpus': 1000}
+        self.server.create_vnodes(self.mom.shortname, a, 5, self.mom)
```
arungrover, line 111: Maybe creating vnodes is better suited for setup_queue_nodes()

test/tests/performance/pbs_multi_sched_perf.py
```diff
@@ -0,0 +1,169 @@
...
+        self.submit_jobs(1000, a)
+        a = {ATTR_q: 'wq4'}
+        self.submit_jobs(1000, a)
+        a = {ATTR_q: 'wq5'}
+        self.submit_jobs(1000, a)
+        start = time.time()
+        for sc in self.scheds:
+            a = {'scheduling': 'True'}
+            self.server.manager(MGR_CMD_SET, SCHED, a, id=sc)
+        for sc in self.scheds:
```
arungrover, line 154: I don't think you need to turn off scheduling. All jobs should ideally run in the same cycle.

test/tests/performance/pbs_multi_sched_perf.py
```diff
@@ -0,0 +1,169 @@
...
+            a = {'scheduling': 'False'}
+            self.server.manager(MGR_CMD_SET, SCHED, a, id=sc)
+        for sc in self.scheds:
+            self.scheds[sc].log_match("Leaving Scheduling Cycle",
+                                      starttime=int(start),
+                                      max_attempts=30, interval=1)
+        for sc in self.scheds:
+            if sc != 'default':
+                sclg = PBSLogAnalyzer()
```
arungrover, line 163: There is already a LogAnalyzer object you created above, can you use that?

test/tests/performance/pbs_multi_sched_perf.py
```diff
@@ -0,0 +1,169 @@
...
+        self.server.manager(MGR_CMD_SET, SERVER, a)
+        self.scheduler.log_match("Leaving Scheduling Cycle",
+                                 starttime=int(start) + 1,
+                                 max_attempts=30, interval=1)
+        sclg = PBSLogAnalyzer()
+        md = sclg.analyze_scheduler_log(
+            filename=self.scheduler.logfile, start=int(start))
+        dur = md['summary']['duration']
+        dur = float(dur.split(':')[2])
+        self.perf_test_result(dur, "single_sched_cycle_duration", "secs")
```
arungrover, line 133: maybe call it default_cycle_duration. That will be in sync with <sched name>_cycle_duration.

test/tests/performance/pbs_multi_sched_perf.py
```diff
@@ -0,0 +1,169 @@
...
+#
+# Use of Altair’s trademarks, including but not limited to "PBS™",
+# "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
+# trademark licensing policies.
+
+
+from tests.performance import *
+
+
+class TestMultipleSchedulersPerf(TestPerformance):
```
bhroam, line 42: Just add this to the current scheduler perf test.  We don't need a separate one for multisched.

test/tests/performance/pbs_multi_sched_perf.py
```diff
@@ -0,0 +1,169 @@
...
+        p4 = {'partition': 'P4'}
+        self.server.manager(MGR_CMD_SET, QUEUE, p4, id='wq4')
+        p5 = {'partition': 'P5'}
+        self.server.manager(MGR_CMD_SET, QUEUE, p5, id='wq5')
+        node = str(self.mom.shortname)
+        self.server.manager(MGR_CMD_SET, NODE, p1, id=node + '[0]')
+        self.server.manager(MGR_CMD_SET, NODE, p2, id=node + '[1]')
+        self.server.manager(MGR_CMD_SET, NODE, p3, id=node + '[2]')
+        self.server.manager(MGR_CMD_SET, NODE, p4, id=node + '[3]')
+        self.server.manager(MGR_CMD_SET, NODE, p5, id=node + '[4]')
```
bhroam, line 87: Maybe do this in a loop like you did in setup_scheds()?

test/tests/performance/pbs_multi_sched_perf.py
```diff
@@ -0,0 +1,169 @@
...
+        self.server.expect(SCHED, {'state': 'scheduling'},
+                           id='default', max_attempts=10)
+        self.scheduler.log_match(
+            "Starting Scheduling Cycle", starttime=int(start),
+            max_attempts=30)
+        a = {'scheduling': 'False'}
+        self.server.manager(MGR_CMD_SET, SERVER, a)
+        self.scheduler.log_match("Leaving Scheduling Cycle",
+                                 starttime=int(start) + 1,
+                                 max_attempts=30, interval=1)
```
bhroam, line 127: When you set scheduling to false, it will wait for the cycle to finish.  This log_match() should immediately get a match.

test/tests/performance/pbs_multi_sched_perf.py
```diff
@@ -0,0 +1,169 @@
...
+        self.submit_jobs(5000)
+        start = time.time()
+        a = {'scheduling': 'True'}
+        self.server.manager(MGR_CMD_SET, SERVER, a)
+        self.server.expect(SCHED, {'state': 'scheduling'},
+                           id='default', max_attempts=10)
+        self.scheduler.log_match(
+            "Starting Scheduling Cycle", starttime=int(start),
+            max_attempts=30)
+        a = {'scheduling': 'False'}
```
bhroam, line 123: Here you can use self.scheduler.run_scheduling_cycle().  This will handle turning on and off scheduling to run a cycle and wait for the cycle to finish.  You can't use it below because you'd end up running the multischeds in a serial fashion.

test/tests/performance/pbs_multi_sched_perf.py
```diff
@@ -0,0 +1,169 @@
...
+        self.server.manager(MGR_CMD_SET, SERVER, a)
+        self.scheduler.log_match("Leaving Scheduling Cycle",
+                                 starttime=int(start) + 1,
+                                 max_attempts=30, interval=1)
+        sclg = PBSLogAnalyzer()
+        md = sclg.analyze_scheduler_log(
+            filename=self.scheduler.logfile, start=int(start))
+        dur = md['summary']['duration']
+        dur = float(dur.split(':')[2])
```
bhroam, line 132: If you want the duration of a cycle, consider using self.scheduler.cycles(lastN=1).  A cycles object is returned and it has a start and end of the cycle.  You can subtract them to get the duration.

test/tests/performance/pbs_multi_sched_perf.py
```diff
@@ -0,0 +1,169 @@
...
+        start = time.time()
+        for sc in self.scheds:
+            a = {'scheduling': 'True'}
+            self.server.manager(MGR_CMD_SET, SCHED, a, id=sc)
+        for sc in self.scheds:
+            a = {'scheduling': 'False'}
+            self.server.manager(MGR_CMD_SET, SCHED, a, id=sc)
+        for sc in self.scheds:
+            self.scheds[sc].log_match("Leaving Scheduling Cycle",
+                                      starttime=int(start),
```
bhroam, line 159: No need to int() start.

test/tests/performance/pbs_multi_sched_perf.py
```diff
@@ -0,0 +1,169 @@
...
+                                      max_attempts=30, interval=1)
+        for sc in self.scheds:
+            if sc != 'default':
+                sclg = PBSLogAnalyzer()
+                filename = self.scheds[sc].attributes['sched_log']
+                md = sclg.analyze_scheduler_log(
+                    filename=filename, start=int(start))
+                dur = md['summary']['duration']
+                dur = float(dur.split(':')[2])
+                self.perf_test_result(dur, sc + "_cycle_duration", "secs")
```
bhroam, line 169: Along with logging it to the json file, do you want to log the output as well?

test/tests/performance/pbs_multi_sched_perf.py
visheshh: Need this for single scheduler also, before setting up queues and partitions, so added it here.

test/tests/performance/pbs_multi_sched_perf.py
visheshh: As i am using log_analyzer i want to read only 1 cycle information. So its better to keep it off. Some empty cycles also show up as 1 second sometimes.

arungrover: Ideally, everything should run in one and only one cycle. If you doubt that then even setting scheduling to True and then False does not mean scheduler will run only one cycle. But I understand this is the best one can do to run only one cycle.

visheshh:
Addressed @bhroam and @arungrover 's comments and moved the test to pbs_sched_perf.py.

test/tests/performance/pbs_sched_perf.py
```diff
@@ -359,3 +361,105 @@ def test_attr_update_period_perf(self):
...
+        for sc in self.scheds:
+            a = {'scheduling': 'True'}
+            self.server.manager(MGR_CMD_SET, SCHED, a, id=sc)
+        for sc in self.scheds:
+            a = {'scheduling': 'False'}
+            self.server.manager(MGR_CMD_SET, SCHED, a, id=sc)
+        for sc in self.scheds:
+            self.scheds[sc].log_match("Leaving Scheduling Cycle",
+                                      starttime=int(start),
+                                      max_attempts=30, interval=1)
```
bhroam, line 452: log_match() will return the matched line.  You can save a lot of log analyzing if you just call PBSLogUtils.convert_date_time(log.split(';')[0]) and take the largest value.  That's the end time of the longest cycle.

visheshh: @bhroam , I have to parse the starting scheduling cycle in that case to find the difference. Is it fine ?

bhroam: I don't see why you have to parse the scheduling cycle.  log_match() will return the 'Leaving Scheduling Cycle' line with the timestamp.  You pass the timestamp to convert_date_time() and it returns you a epoch time.  Pick the max of all the epoch times and you have your cycle end time.

visheshh: @bhroam Yes i can do that to findout cycle end time, but i want to calculate the cycle duration time , for that i need both start time and end time.
So i need both starting and leaving scheduling cycle time. so i want to log_match both the messages.

bhroam: You can find  the start by doing the log_match() or you'll likely be close enough if you just do a start_time = time.time() before you start kicking cycles.

visheshh:
@bhroam , Addressed comments, Please review.

test/tests/performance/pbs_sched_perf.py
```diff
@@ -446,20 +446,15 @@ def test_multi_sched_perf(self):
...
-                end = time.time()
-                sl.analyze(filename, start, end, self.server.hostname)
-                cycles = sl.cycles
-                if cycles is None or len(cycles) == 0:
-                    return []
-                dur = cycles[-1:][0].end - cycles[-1:][0].start
+                self.logger.info("searching log for scheduler " + str(sc))
+                log_msg = self.scheds[sc].log_match("Leaving Scheduling Cycle",
+                                                    starttime=int(start),
+                                                    max_attempts=30)
```
bhroam, line 454: Leave max_attempts as the default just incase the scheduling cycle takes longer that 15s

test/tests/performance/pbs_sched_perf.py
```diff
@@ -446,20 +446,15 @@ def test_multi_sched_perf(self):
...
+                endtime = PBSLogUtils.convert_date_time(
+                    log_msg[1].split(';')[0])
+                dur = endtime - start
                 self.perf_test_result(dur, sc + "_cycle_duration", "secs")
                 self.logger.info('Time taken by ' + str(sc) +
                                  ' scheduler to run 1k jobs is ' + str(dur))
```
bhroam, line 460: I think this message is too detailed for what we want.  The purpose of this test is to see the difference between running 1 cycle with 5k jobs and 5 cycles with 1k jobs and see the latter is faster.  I don't think we care the individual cycle times for the 5 cycles.  We want to see the difference between the first cycle, and the longest of the 5 cycles.

visheshh: done. @bhroam Please review

test/tests/performance/pbs_sched_perf.py
```diff
@@ -455,6 +457,17 @@ def test_multi_sched_perf(self):
...
+                else:
+                    if dur > max:
+                        max = dur
+            i = 1
+        self.perf_test_result(max, "max_multisched_cycle_duration", "secs")
+        msg = 'Max time taken by one of the multi sched to run 1k jobs is '
+        self.logger.info(msg + str(max))
+        self.perf_test_result(
+            cyc_dur - max, "multisched_defaultsched_cycle_diff", "secs")
+        msg1 = 'Multi scheduler is faster than single scheduler by '
```
bhroam, line 471: Just in case, add a self.assertLess(max, cyc_dur)

test/tests/performance/pbs_sched_perf.py
```diff
@@ -455,6 +457,17 @@ def test_multi_sched_perf(self):
...
                     log_msg[1].split(';')[0])
                 dur = endtime - start
-                self.perf_test_result(dur, sc + "_cycle_duration", "secs")
-                self.logger.info('Time taken by ' + str(sc) +
-                                 ' scheduler to run 1k jobs is ' + str(dur))
+                if i == 0:
+                    max = dur
+                else:
+                    if dur > max:
+                        max = dur
```
bhroam, line 464: For one, you don't need i.  dur can't be negative, so this if statement will work.  The other option is to create a list of durations, and then call max(list) at the end.

visheshh: i is just used as a flag to know that it is first attempt. I am just storing the max value of the durations.

PR #1964, opened 2020-08-10
PTL expect doesn't handle default values for unset operation

dinesh1306:
#### Describe Bug or Feature
PTL expect doesn't handle default values for unset operation.
current code of expect function assumes that 'unset' operation means that attribute will disappear from stat output. there are many default attributes in PBS, when we unset those attributes get 'reset' by PBS to their default values.
for example, if we unset scheduling attribute of sched object gets set to True.
we need to update code to unset those value and check that value reset to there default value.

#### Describe Your Change
we have created object wise default attribute dictionary and after unset those values we checked that value get reset to the expected vaue.

3088 | regression | Tests from given sources on platforms UBUNTU1804, SUSE15, CENTOS8 | dinesh1306/openpbs@expect_default_attr | 10th Aug 2020 09:20:58.250 AM +05:30 | 00:37:19.452 | 2703 | 0 | 0 | 4 | 2 | 0 | 0 | 2697
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --


test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -13152,6 +13196,8 @@ class MoM(PBSService):
                        'PBS_MANAGER_SERVICE_PORT': '-R',
                        'PBS_HOME': '-d'}
 
+    revert_default_attrib = {'resources_available.ncpus': 1}
+
```
shilpakodli, line 13200: Instead of checking this to 1, we can get ncpus values when we create nodes for the first time in setUp class. Not all systems will have ncpus value as 1 by default.

shilpakodli:
> Instead of keeping a list of default attributes which will go stale, maybe we can take a more proactive approach. We do a revert_to_defaults as the first thing. How about we do a status() and capture those as 'defaults'. We know the only things which aren't default (I think managers/operators is it).

If we this then PTL is not exactly setting all attributes to defaults. We have few skip list where we are not reverting few server attribs. For Ex: pbs_license_min, managers and operators. We can still take status of server in revert_to_defaults and include/exclude/update from this list to default attributes.

bhroam:
@shilpakodli I see your point.  How about we keep a list of 'default values' for just the attributes we don't set back to defaults?  It'll be much smaller than the full list.  That way we can still do the status for everything else.  By doing the status(), we catch any new attributes which revert to their defaults without having to update a list in PTL.

shilpakodli:
> @shilpakodli I see your point. How about we keep a list of 'default values' for just the attributes we don't set back to defaults? It'll be much smaller than the full list. That way we can still do the status for everything else. By doing the status(), we catch any new attributes which revert to their defaults without having to update a list in PTL.

Sounds good.

dinesh1306:
@shilpakodli , I have incorporate review comment. Please have a look
image

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -4683,18 +4683,15 @@ class Server(PBSService):
...
-                             'pbs_license_min': 0,
-                             'pbs_license_max': 2147483647,
-                             'pbs_license_linger_time': 31536000}
+    server_ignore_attrs = [ATTR_SvrHost, ATTR_total, ATTR_count, ATTR_managers,
+                           ATTR_operators, ATTR_dfltque, ATTR_FlatUID,
+                           ATTR_FLicenses, ATTR_nodefailrq,
+                           ATTR_pbs_license_info, ATTR_license_count,
+                           ATTR_version, ATTR_power_provisioning,
+                           ATTR_DefaultChunk + '.ncpus',
+                           ATTR_max_job_sequence_id, ATTR_status, 'id']
```
shilpakodli, line 4692: Why are you ignoring these attributes ? Not all in this list are unsettable. For example, we can unset flatuid, node_fail_requeue, pbs_license_info, default_chunk.ncpus, max_job_sequence_id. All these values will be removed when unset. Remove these from the list and add an else condition for if condition at line number 8474

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -8474,16 +8471,24 @@ def expect(self, obj_type, attrib=None, id=None, op=EQ, attrop=PTL_AND,
...
+                                                   level=logging.DEBUG,
+                                                   extend=extend,
+                                                   runas=runas,
+                                                   logerr=False)
+                            if 'pcpus' not in statlist[0].keys():
+                                val = 1
+                            else:
+                                val = MoM.mom_dflt_attr[id][key]
+                        attrib = {key: val}
                         self.manager(MGR_CMD_UNSET, obj_type, key, id=id)
```
shilpakodli, line 8492: Move this out of if condition. below three lines

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -8474,16 +8471,24 @@ def expect(self, obj_type, attrib=None, id=None, op=EQ, attrop=PTL_AND,
         else:
             if op == UNSET:
                 for key in attrib.keys():
-                    if (key in Scheduler.revert_default_attrib.keys() or
-                            key in Server.revert_default_attrib.keys() or
-                            key in MoM.revert_default_attrib.keys()):
+                    if (key not in Scheduler.sched_ignore_attrs or
```
shilpakodli, line 8474: Add an else condition for unset attributes

dinesh1306:
@shilpakodli , I have incorporate review comments, Please have a look.
[res_TestExpect.txt](https://github.com/openpbs/openpbs/files/5143343/res_TestExpect.txt)

shilpakodli (commented):
Thanks Dinesh for making changes. Could you please attach regression test report with latest changes.

bhroam:
@hirenvadalia I was going to review this but got confused.  Could you review it?  If you are too busy, I'll do it.

hirenvadalia:
@bhroam Sure, I will review...

dinesh1306:
@shilpakodli , please find regression result.

|--|--|--|--|--|--|--|
|3448|7821|6|1|2|6|7806|

shilpakodli (approved):
Thanks Dinesh.

hirenvadalia:
@bhroam and others, just FYI: I had offline discussion with @dinesh1306, and we found new approach similar to this but with little change, so @dinesh1306 will post details about new approach

dinesh1306:
New approach is as follows,
We are not using class level dictionary and list for special attributes (Attributes which are getting reset when we unset it) for every object, instead we are using dictionary in Class Server with following approch.

__special_attr_keys = {
                       SERVER: [k1, k2],
					   SCHED: [k1, k2, k3]
					  }
above dictionary having key-value of object_type and list of there special attributes.

__special_attr = {SERVER: {id: {k1: None, k2: None}}
                  NODE: {id: {k1: None, k2: None},
				         id1: {k1: None, k2: None}
						}
			     }

we are using __special_attr dictionary having object_type and nested dictionary for different id's of obsject and there attributes.
e.g. for NODE object we can have multiple object with different values, to manage we are create nested dictionary with id.

we are adding new function 'update_special_attr' in Class Server to update dictionary (__special_attr). the function takes 3 parameter as input i.e. object_type, object_stat, id.

in expect function we are not using manager to unset any attribute, only we are checking that attribute is unset and getting there respective value.

dinesh1306:
[res_test_revert_attributes.txt](https://github.com/openpbs/openpbs/files/5259293/res_test_revert_attributes.txt)

hirenvadalia (approved):

shilpakodli (approved):

hirenvadalia:
@dinesh1306 Please rebase your branch and run full regression and post results.

dinesh1306:
@hirenvadalia , Please find regression result.
Description: Tests from given sources on platforms UBUNTU1804, SUSE15, CENTOS8, CENTOS7
Platforms: UBUNTU1804,SUSE15,CENTOS8,CENTOS7
Profile: regression
|--|--|--|--|--|--|--|
|4379|1067|1|1|0|0|1065|

PR #1965, opened 2020-08-10
Set JOB_EXEC_RERUN as the Exit_status when doing a qrerun

lisa-altair:
#### Describe Bug or Feature

JOB_EXEC_RERUN (-11) was not always set depending on whether the qrerun was done with -Wforce and the state of the job.  Thus the accounting log "R" record (for rerun jobs) did not always have an Exit_status =-11.

Also, I noticed 2 "R" records in the accounting log for the same job for the same rerun.  I removed one of the duplicate logs.

#### Describe Your Change

Set JOB_EXEC_RERUN in one path where it was not previously being set.
Reorder some code and remove a call to account_jobend() since account_jobend will be called in force_reque anyway.

### DIFF
--- a/src/server/node_manager.c
+++ b/src/server/node_manager.c
@@ -900,12 +900,11 @@ post_discard_job(job *pjob, mominfo_t *pmom, int newstate)
 		sprintf(log_buffer, ndreque, downmom);
 		log_event(PBSEVENT_JOB, PBS_EVENTCLASS_JOB, LOG_INFO,
 			pjob->ji_qs.ji_jobid, log_buffer);
-		account_jobend(pjob, pjob->ji_acctrec, PBS_ACCT_RERUN);
+		force_reque(pjob);
 		if (pjob->ji_acctrec) {
 			free(pjob->ji_acctrec);	/* logged, so clear it */
 			pjob->ji_acctrec = NULL;
 		}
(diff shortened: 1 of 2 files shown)


src/server/node_manager.c
```diff
@@ -900,12 +900,11 @@ post_discard_job(job *pjob, mominfo_t *pmom, int newstate)
 		sprintf(log_buffer, ndreque, downmom);
 		log_event(PBSEVENT_JOB, PBS_EVENTCLASS_JOB, LOG_INFO,
 			pjob->ji_qs.ji_jobid, log_buffer);
-		account_jobend(pjob, pjob->ji_acctrec, PBS_ACCT_RERUN);
+		force_reque(pjob);
 		if (pjob->ji_acctrec) {
 			free(pjob->ji_acctrec);	/* logged, so clear it */
 			pjob->ji_acctrec = NULL;
 		}
```
bhroam, line 907: If force_reque() is logging, maybe this should go into force_reque()?

lisa-altair: force_reque is calling account_jobend which does the logging.  I just reordered what was here to prevent logging twice.  Not all places that call account_jobend or force_reque free ji_acctrec, so I don't want to add it to within force_reque nor account_jobend.

PR #1968, opened 2020-08-11
Fixed race condition in test by adding expect

Klovely:
#### Describe Bug or Feature
On few machine test_psets_calendaring from TestNodeBuckets failed with IndexError: list index out of range
Some time test is failed at below step 
`ev = self.server.status(JOB, 'estimated.exec_vnode', id=jid3)`
 `used_nodes3 = j3.get_vnodes(ev[0]['estimated.exec_vnode'])`
with error : -
 used_nodes3 = j3.get_vnodes(ev[0]['estimated.exec_vnode'])

#### Describe Your Change
Test throwing error because of race condition as we are taking 'estimated.exec_vnode'  from JOB status immediately after 
scheduler.log_match ';Job is a top job' and here is some delay between scheduler update estimated.start_time and estimated.exec_vnode to server and if  in between self.server.status will run its not getting estimated.exec_vnode value and in that case test throw error

To fix that adding expect on to confirm 'estimated.start_time' is set on Q jobs, this check will cover delay as well.

[test_psets_calendering_after_fix.txt](https://github.com/openpbs/openpbs/files/5058738/test_psets_calendering_after_fix.txt)

### DIFF
--- a/test/tests/functional/pbs_node_buckets.py
+++ b/test/tests/functional/pbs_node_buckets.py
@@ -639,6 +639,8 @@ class TestNodeBuckets(TestFunctional):
 
         n = self.server.status(NODE, 'resources_available.shape')
 
+        self.server.expect(JOB, {'job_state': 'Q'}, id=jid2)
+        self.server.expect(JOB, 'estimated.start_time', id=jid2, op=SET)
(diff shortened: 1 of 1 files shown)


prakashcv13:
merging based on reviews from @bhroam and @dinesh1306

PR #1969, opened 2020-08-12
 Tests "test_direct_write_job_array" and "test_direct_write_job_array_custom_dir" skip if don't get required ncpus

birbalsain:
#### Describe Bug or Feature
Tests "test_direct_write_job_array" and "test_direct_write_job_array_custom_dir" need 4 ncpus but on cpuset machines cpus set by cgroup are 1 so tests failed on cpuset machines.

#### Describe Your Change

Added skipOnCpuSet decorator to the tests

[direct_write_skip.txt](https://github.com/openpbs/openpbs/files/5067651/direct_write_skip.txt)

### DIFF
--- a/test/tests/functional/pbs_qsub_direct_write.py
+++ b/test/tests/functional/pbs_qsub_direct_write.py
@@ -54,6 +54,14 @@ class TestQsub_direct_write(TestFunctional):
         self.msg = "Job is sleeping for 10 secs as job should  be running"
         self.msg += " at the time we check for directly written files"
 
+    def checks_available_ncpus(self, ncpus=1):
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_qsub_direct_write.py
```diff
@@ -330,6 +330,12 @@ def test_direct_write_job_array(self):
         accessible from mom and direct_files option is used
         but submission directory is not mapped in mom config file.
         """
+        nodes = self.server.counter(NODE, 'resources_available.ncpus',
```
bayucan, line 333: Rather than repeating this block of code, just put it in  a function.

hirenvadalia:
Merging based on reviews...

PR #1970, opened 2020-08-13
Increase reliability with altering reservations

vstumpf:
#### Describe Bug or Feature
On startup, if the server recovered a reservation in the RESV_BEING_ALTERED state, it would not correctly un-alter it. This could happen in multiple ways:
 - A reservation was being altered and the scheduler crashed
 - A reservation was being altered and the server crashed
 - A reservation was being altered and the scheduler was off or unresponsive
 - A reservation was being altered and scheduling was turned off.

In addition, if a standing reservation was altered, and the server went down, the standing reservation would not be recovered correctly.

#### Describe Your Change
Create two new attributes, one for the temporary pre-alter state of a reservation, and a longer one for standing reservations.
Since they are attributes, they will be written to the database.
If a reservation is in RESV_BEING_ALTERED when recovered from the database, revert the reservation.

[resv4.txt](https://github.com/openpbs/openpbs/files/5093152/resv4.txt)
[resv4a.txt](https://github.com/openpbs/openpbs/files/5093154/resv4a.txt)


src/server/req_modify.c
```diff
@@ -743,6 +739,154 @@ is_select_smaller(char *select_orig, char *select_smaller)
...
+	attribute *alter_revert;
+	resource_def *presdef = NULL;
+	resource *resc = NULL, *pselect = NULL;
+	attribute atemp = {0};
+
+	alter_revert = &presv->ri_wattr[RESV_ATR_alter_revert];
+	if (alter_revert->at_flags & ATR_VFLAG_SET)
+		return;
+
+	alter_revert->at_flags |= ATR_SET_MOD_MCACHE;
```
arungrover, line 759: We are trying to move away from setting attributes without calling the setter functions. So in this case, can we use set_resc() instead? You can probably make a list of new resources to add and then call set_resc()?

src/server/req_modify.c
```diff
@@ -743,6 +739,154 @@ is_select_smaller(char *select_orig, char *select_smaller)
...
+	int sub = 0;
+
+	alter_revert = &presv->ri_wattr[RESV_ATR_alter_revert];
+	if (!alter_revert->at_flags & ATR_VFLAG_SET)
+		return;
+
+	/* revert start time */
+	presdef = &svr_resc_def[RESC_START_TIME];
+	resc = find_resc_entry(alter_revert, presdef);
+	presv->ri_wattr[RESV_ATR_start].at_val.at_long = resc->rs_value.at_val.at_long;
```
arungrover, line 806: can we call setter functions here as well?

src/server/req_modify.c
vstumpf: is add_resource_entry + resource_def::rs_set not a valid way of doing this?
add_resource_entry does set alter_revert's at_flags to ATR_SET_MOD_MCACHE, so i can remove this line

bhroam:
@agrawalravi90 is converting attributes to use setters.  We might want to talk with him to do this the right way.

src/server/svr_jobfunc.c
```diff
@@ -3460,9 +3454,28 @@ eval_resvState(resc_resv *presv, enum resvState_discrim s, int relVal, int *psta
...
+		if (presv->ri_alter.ra_state) {
+			*pstate = presv->ri_alter.ra_state;
+		} else if (*psub == RESV_IN_CONFLICT) {
+			if (is_running) {
+				*pstate = RESV_RUNNING;
+			} else {
+				*pstate = RESV_DEGRADED;
+			}
+		} else if (*psub == RESV_DEGRADED) {
+			*pstate = RESV_DEGRADED;
```
bhroam, line 3467: If substate is degraded and is_running, pstate needs to be running.  Collapse this condition into the above condition.

test/tests/functional/pbs_ralter.py
```diff
@@ -2595,3 +2598,45 @@ def test_alter_duration_standing_resv_future_occrs(self):
...
+            offset, duration)
+
+        attrs = {'reserve_start':
+                 self.bu.convert_seconds_to_datetime(start, self.fmt),
+                 'reserve_end':
+                 self.bu.convert_seconds_to_datetime(end, self.fmt),
+                 'reserve_state': (MATCH_RE, 'RESV_CONFIRMED|2')}
+        self.server.expect(RESV, attrs, id=rid)
+
+        self.scheduler.stop()
```
bhroam, line 2624: Turn scheduling off, don't stop the scheduler.

PR #1976, opened 2020-08-18
[WIP] Optimize updating of array_indices_remaining in job array during run/end/del job path

bremanandjk:
#### Describe Bug or Feature

Identified a possibility to optimize the job array run path while running a large number of sub-jobs.  Modified the code in a way that would save Server's execution time.

#### Describe Your Change

For every sub job run request, there is a function [**update_array_indices_remaining_attr()**] getting called to update Job array's "array_indices_remaining" attributes in the run path. This function is a fully time-consuming operation that involves loops & massive string operations. It is not necessary to update in every job run path since the same attr would be updated eventually in the subsequent **stat_job** calls. Nonetheless, there wouldn’t be any difference to the external user since the user would see the same value as earlier.

#### Link to Design Doc

NA

Before optimization: 

Before optimization

After optimization:

After OPptimization

**Point to Note:** Here in perf results, I have shown the optimization of time from update_array_indices_remaining_attr, It's not just shown percent, we should consider the entire function time is reduced. For more volume of jobs, we could see more reductions.

### DIFF
--- a/src/server/array_func.c
+++ b/src/server/array_func.c
@@ -380,6 +380,11 @@ update_array_indices_remaining_attr(job *parent)
 {
 	struct ajtrkhd	*ptbl = parent->ji_ajtrk;
 
+	if (ptbl->tkm_flags & TKMFLG_RUN_JOB_REQ) {
+		ptbl->tkm_flags &= ~TKMFLG_RUN_JOB_REQ;
+		return;
+	}
+
 	if (ptbl->tkm_flags & TKMFLG_REVAL_IND_REMAINING) {
 		attribute *premain = &parent->ji_wattr[(int)JOB_ATR_array_indices_remaining];
 		char *pnewstr = cvt_range(parent, JOB_STATE_QUEUED);
--- a/src/server/req_runjob.c
+++ b/src/server/req_runjob.c
@@ -307,6 +307,7 @@ req_runjob(struct batch_request *preq)
 	job *pjob = NULL;
 	job *pjobsub = NULL;
 	job *parent = NULL;
+	struct ajtrkhd  *ptbl = NULL;
 	char *range;
 	int start;
 	int end;
@@ -545,6 +546,10 @@ req_runjob(struct batch_request *preq)
 		if (pjob) {
 			/* free prov_vnode before use */
 			job_attr_def[(int) JOB_ATR_prov_vnode].at_free(&pjob->ji_wattr[(int) JOB_ATR_prov_vnode]);
+			ptbl = parent->ji_ajtrk;
+			if (ptbl == NULL)
+				return;
+			ptbl->tkm_flags |= TKMFLG_RUN_JOB_REQ;
 			req_runjob2(preq, pjob);
 		}
 		return;
--- a/src/include/job.h
+++ b/src/include/job.h
@@ -372,6 +372,7 @@ struct jbdscrd {
 #define TKMFLG_NO_DELETE           0x01
 #define TKMFLG_REVAL_IND_REMAINING 0x02 /* Flag to re-evaluate "array_indices_remaining" */
 #define TKMFLG_CHK_ARRAY           0x04 /* chk_array_doneness() already in call stack*/
+#define TKMFLG_RUN_JOB_REQ         0x08 /* For denoting run_job request */
 
 /* Structure for block job reply processing */
 struct block_job_reply {


bremanandjk:
**Master branch:**
Description: Tests from given sources on platforms UBUNTU1804, SUSE15, UBUNTU1604, CENTOS8, RHEL8
Platforms: UBUNTU1804,SUSE15,UBUNTU1604,CENTOS8,RHEL8
Profile: regression
|--|--|--|--|--|--|--|
|3210|3506|14|0|0|2|3490|

**Regression results:** 
Description: Tests from given sources on platforms UBUNTU1804, SUSE15, UBUNTU1604, CENTOS8, RHEL8
Platforms: UBUNTU1804,SUSE15,UBUNTU1604,CENTOS8,RHEL8
Profile: regression
|--|--|--|--|--|--|--|
|3213|3506|7|0|0|2|3497|

Description: **Rerun Only Failed Tests** From #3213
Platforms: UBUNTU1804,SUSE15,UBUNTU1604,CENTOS8,RHEL8
Profile: regression
|--|--|--|--|--|--|--|
|3217|7|4|0|0|0|3|

* Those 4 test failures are also observed in the mainline.

agrawalravi90:
@bremanandjk can you please also measure timing numbers? If you submit a job array with a million jobs, 1 node with 50k or 100k ncpus, mock mom, how soon do the jobs run and exit the system?

agrawalravi90:
I'm also a bit confused about your regression results .. master branch is 14 failures? But at the end you say "Those 4 test failures are also observed in the mainline" can you please explain?

src/server/array_func.c
```diff
@@ -380,6 +380,11 @@ update_array_indices_remaining_attr(job *parent)
 {
 	struct ajtrkhd	*ptbl = parent->ji_ajtrk;
 
+	if (ptbl->tkm_flags & TKMFLG_RUN_JOB_REQ) {
```
agrawalravi90, line 383: please add a comment here to explain what you are doing

bremanandjk:
> I'm also a bit confused about your regression results .. master branch is 14 failures? But at the end you say "Those 4 test failures are also observed in the mainline" can you please explain?

Those 4 failures in my branch:
image

**None related to job arrays.** 

Please refer the built id and cross verify the test results: Build Ids 3213 & 3217

A rerun of those 14 failures in mainline:

Description: Rerun Only Failed Tests From #3210
Platforms: UBUNTU1804,SUSE15,UBUNTU1604,CENTOS8,RHEL8
Profile: regression
|--|--|--|--|--|--|--|
|3215|14|3|2|0|0|9|

Now, let me know your confusion, I will clarify.

src/server/array_func.c
```diff
@@ -380,6 +380,11 @@ update_array_indices_remaining_attr(job *parent)
...
+		ptbl->tkm_flags &= ~TKMFLG_RUN_JOB_REQ;
```
agrawalravi90, line 384: it seems like req_runjob has the capability to run a range of subjobs. By removing the flag here, it seems like this will only work for the first subjob in a range. So, I think we should not remove the flag here, what do you think?

src/server/req_runjob.c
```diff
@@ -545,6 +546,10 @@ req_runjob(struct batch_request *preq)
 		if (pjob) {
 			/* free prov_vnode before use */
 			job_attr_def[(int) JOB_ATR_prov_vnode].at_free(&pjob->ji_wattr[(int) JOB_ATR_prov_vnode]);
+			ptbl = parent->ji_ajtrk;
+			if (ptbl == NULL)
+				return;
```
agrawalravi90, line 551: I don't think you should return/bail here, this is just a performance optimization, so just add the flag if parent->ji_ajtrk exists so that functionally it still works like before.

src/server/req_runjob.c
```diff
@@ -545,6 +546,10 @@ req_runjob(struct batch_request *preq)
...
+			ptbl->tkm_flags |= TKMFLG_RUN_JOB_REQ;
```
agrawalravi90, line 552: I think you need to also do this for the range of subjobs code below

bremanandjk: Please refer to the line what you are talking about?

agrawalravi90:
> A rerun of those 14 failures in mainline

Are you saying that, on master, 14 tests failed when you ran all tests, but 9 of them passed when you re-ran just the 14 ? @kjakkali Is this possible on TH3?

src/server/array_func.c
bremanandjk: I ran a 40k job array test, I didn't see a single occurrence of a range of sub-jobs to run job request. Looks like this is not a supported feature. In the future, if we add support for it, the functionality should be added with full optimization. It's not required today.

src/server/req_runjob.c
bremanandjk: Ignoring the range case, since it is not a supported feature. I didn't see any occurrences in the 40k Job array run.

src/server/array_func.c
hirenvadalia: @bremanandjk 'req_runjob' is also used for qrun, which supports range of subjob, run same test with sched off and do qrun of range of subjob

bremanandjk: Yeah, for supporting those user-triggered corner cases we would need to add code logic, that may burden the scheduler triggered run path. What do you think? Is there a way to separate out both the paths. I feel It's not a common path to consider. It's my opinion.

subhasisb:
Firstly, and thinking aloud, who cares for this attributes JOB_ATR_array_indices_remaining?

1. Stats need to show a string of count, etc, but then stat calls update_array_indices_remaining_attr() every time.
2. Sched needs to know pending subjobs, instead of searching, so yes, it also need, but then stat cover it also.
3. I think we DO NOT need to save this parameter, so we should mark this as NO_SAVM. I think server recomputes this at startup.

If all the above is true, then:
a) Eliminating calls to "cvt_range" for every subjob run sounds like a fine approach, since the scheduler is likely to run thousands of subjobs in one cycle, for which it would do one stat and thus one update_array_indices_remaining_attr(), instead of invoking it after each subjob run/end/del.
b) Still mark the attribute as NO_SAVM.
c) In fact we should remove the 1 lines at the bottom of function chk_array_doneness() so that for each subjob end/del, we DO NOTsave the arrayjob.

This leaves behind the question of how will cvt_range perform for each time stat() is called? Currently, it will go over 1 million entries to find out how many are still queued. This is still bad, quite bad, as we will often loop close to a million times (and what if we have 20 million jobs?). so, the problem is that it recomputes everything from scratch for a million jobs. If it persisted the JOB_ATR_array_indices_remaining and used it, then it would be faster. Now, even if you keep 4-1M as a range, then checking (at stat time) from 4 to 1M is expensive, so the only other way is to update as we go - which means updating this attribute (or the seed to it) at every runjob and endjob. The obvious and simple way to make this work is to simply use the same attribute value, parse and update on the fly. This is same as @hirenvadalia idea, but somewhat shorter and simply using string manipulation.

src/server/array_func.c
bhroam: There are two cases here.  The admin can do a qrun 123[1-10] saying they want to run those subjobs.  That is then passed to the scheduler.  The scheduler has code to handle this request.  The second is a qrun -H.  This is how the admin overrides the scheduler and does what they want to do.  We shouldn't break either of these.

src/server/req_runjob.c
bhroam: I believe running a range of subjobs is a supported feature.  The scheduler doesn't run a range because it is hard for it.  It needs to provide a destination, so it would only be able to run a range if they were all going to the same place.  Running a range is mainly used for qrun.

bhroam:
@subhasis I'm a little confused about your comment.  You first say who cares about array indices remaining, and then you go over ways to make it faster.  The answer to your question is the scheduler cares about the attribute.  The scheduler uses it to determine what subjobs are remaining to be run.  You are right that it will only stat it once per cycle, so there could be many runs per stat.  The scheduler doesn't care about the internals of the attribute, as long it is correct when it is stated at the start of the cycle.

@hirenvadalia I haven't looked at the server's structures for job array's, but the scheduler has such a range list.  It keeps track of the ranges remaining to be run in the scheduler.  Take a look at src/scheduler/range.c.

src/server/array_func.c
bremanandjk: Thanks for your comment Bhroam, just wanted to point that the proposed change won't break the range of sub-jobs functionality, this optimization would not be applied to a range of sub-jobs path. 

Anyhow, further discussion converged not to remove the attribute computation in the run path. There won't be any difference between the single sub job and range of sub-jobs then.

agrawalravi90:
@bremanandjk sounds like you are going to change approach, so please mark this as a WIP.

hirenvadalia:
> @subhasis I'm a little confused about your comment. You first say who cares about array indices remaining, and then you go over ways to make it faster. The answer to your question is the scheduler cares about the attribute. The scheduler uses it to determine what subjobs are remaining to be run. You are right that it will only stat it once per cycle, so there could be many runs per stat. The scheduler doesn't care about the internals of the attribute, as long it is correct when it is stated at the start of the cycle.
> 
> @hirenvadalia I haven't looked at the server's structures for job array's, but the scheduler has such a range list. It keeps track of the ranges remaining to be run in the scheduler. Take a look at src/scheduler/range.c.

Thank @bhroam for pointing to range.c, that's exactly what I was taking about, like when we submit array job then use range_parse(submitted_indices) to create xyz list, then use range_remove_value() (but modified version of it as right range_remove_value doesn't support removal of value in middle of range) then call range_as_str to get remain_indices string...

But Subhasis went one step ahead of me, where he is suggesting to make use of remain_indices attr's encoded data directly with string manipulation so that we don't have to maintain xyz list.

subhasisb:
> @subhasis I'm a little confused about your comment. You first say who cares about array indices remaining, and then you go over ways to make it faster. The answer to your question is the scheduler cares about the attribute. The scheduler uses it to determine what subjobs are remaining to be run. You are right that it will only stat it once per cycle, so there could be many runs per stat. The scheduler doesn't care about the internals of the attribute, as long it is correct when it is stated at the start of the cycle.

@bhroam sorry for my confusing comment. I did know that the sched uses it. What meant was we do not need to store it in the database, and thus we can eliminate one job_save() since the server can reconstruct it on startup

bhroam:
@subhasisb Thanks for the clarification.  As long as it doesn't take a very long time to recreate it, there is no need to save it.
@hirenvadalia It is true that you can't remove a value from the middle of a range.  I thought about implementing it when I originally wrote the range code, but since it was kind of complex and the scheduler didn't need it (scheduler always runs the lowest index), I chose to leave the limitation in.  The server would need it.  You can do things like qdel an index from the middle of the array.  It sounds like using the range code won't be necessary though.

bremanandjk:
Closing this PR, decided to do another approach.

PR #1978, opened 2020-08-18
reverthooks will disable cgroups hook by default

vccardenas:
#### Describe Bug or Feature
In the server revert to defaults reverthooks is True by default and is changed to False if platform is not 'cray', craysim' or 'shasta'.

One failure case is when there are more than one mom in a cluster and one of them is a cpuset machine we
would expect that the cgroups hook has been enabled.  If in the next test run, where setUp is called which calls
server revert to defaults, and the one and only cpuset mom is replaced in the cluster by a machine that is not a
cpuset machine and not cray-like (see above), the still enabled cgroups hook can get in the way and cause tests to fail.

#### Describe Your Change
With this change the cgroups hook is reverted to default during server revert to default on all platforms.
The PBS hooks are not reverted to default. As usual the cgroups hook is enabled as needed in mom revert to defaults which happens after server revert to defaults. 

The test is skipped when one of the moms is a cpuset mom:
[x56_x70_before.txt](https://github.com/openpbs/openpbs/files/5092947/x56_x70_before.txt)
The enabled cgroups hook gets in the way when using moms that are not cpuset machines:
[x56_x114_before.txt](https://github.com/openpbs/openpbs/files/5092948/x56_x114_before.txt)

The test is skipped when one of the moms is a cpuset mom:
[x56_x70_after.txt](https://github.com/openpbs/openpbs/files/5092949/x56_x70_after.txt)
The cgroups hook is now disabled when using moms that are not cpuset machines and test completes:
[x56_x114_after.txt](https://github.com/openpbs/openpbs/files/5092950/x56_x114_after.txt)

Test results after moving changes to server revert to defaults:
[x56_x70_after.1.txt](https://github.com/openpbs/openpbs/files/5098192/x56_x70_after.1.txt)
[x56_x70_x114_after.1.txt](https://github.com/openpbs/openpbs/files/5098193/x56_x70_x114_after.1.txt)
[x56_x114_after.1.txt](https://github.com/openpbs/openpbs/files/5098194/x56_x114_after.1.txt)

Test results with revert_hooks changes:
[x56_x114_after.2.txt](https://github.com/openpbs/openpbs/files/5110670/x56_x114_after.2.txt)

Before and after tests on x81:
[SmokeTest_x81_after.txt](https://github.com/openpbs/openpbs/files/5125664/SmokeTest_x81_after.txt)
[SmokeTest_x81_before.txt](https://github.com/openpbs/openpbs/files/5125663/SmokeTest_x81_before.txt)

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -5185,6 +5185,8 @@ class Server(PBSService):
 
         if delnodes:
             self.delete_nodes()
+        self.manager(MGR_CMD_SET, MGR_OBJ_HOOK,
+                     {'enabled': 'False'}, 'pbs_cgroups')
         if reverthooks:
             if self.platform == 'shasta':
                 dohup = False
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -5166,11 +5166,6 @@ class Server(PBSService):
         if self.platform == 'cray' or self.platform == 'craysim':
             setdict[ATTR_restrict_res_to_release_on_suspend] = 'ncpus'
         if delhooks:
-            if (self.platform == 'cray' or self.platform == 'craysim' or
-                    self.platform == 'shasta'):
-                reverthooks = True
-            else:
-                reverthooks = False
             self.delete_site_hooks()
         if delqueues:
             revertqueues = False
@@ -5185,9 +5180,6 @@ class Server(PBSService):
 
         if delnodes:
             self.delete_nodes()
-        # disable cgroups hook - mom revert will enable only if needed
-        self.manager(MGR_CMD_SET, MGR_OBJ_HOOK,
-                     {'enabled': 'False'}, 'pbs_cgroups')
         if reverthooks:
             if self.platform == 'shasta':
                 dohup = False
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -5185,6 +5185,7 @@ class Server(PBSService):
 
         if delnodes:
             self.delete_nodes()
+        # disable cgroups hook - mom revert will enable only if needed 
         self.manager(MGR_CMD_SET, MGR_OBJ_HOOK,
                      {'enabled': 'False'}, 'pbs_cgroups')
         if reverthooks:
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -5185,7 +5185,7 @@ class Server(PBSService):
 
         if delnodes:
             self.delete_nodes()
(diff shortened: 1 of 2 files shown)


test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -5185,6 +5185,9 @@ def revert_to_defaults(self, reverthooks=True, revertqueues=True,
 
         if delnodes:
             self.delete_nodes()
+        # disable cgroups hook - mom revert will enable only if needed
```
agrawalravi90, line 5188: I'm a bit confused, doesn't the "if reverthooks:" section below this take care of the cgroups hook? If not, then I think we should fix that instead of especially disabling the cgroups hook here. The "if reverthooks:" section is supposed to disable all hooks

agrawalravi90:
Thanks @vccardenas can you please run a full regression just to make sure that these changes don't cause any unexpected side effects?

vccardenas:
Results from full regression (the one failure is from a perl test):

Description: Tests from given build on platforms UBUNTU1804, SUSE15, UBUNTU1604, CENTOS8, RHEL8
Platforms: UBUNTU1804,SUSE15,UBUNTU1604,CENTOS8,RHEL8
Profile: regression
|--|--|--|--|--|--|--|
|3364|3443|1|0|0|2|3440|

bayucan:
@agrawalravi90 : is this PR looks good to you?

lisa-altair:
Hi @vccardenas  thanks for changing the title.  But, it's not just the cgroups hook that will be disabled by default.  It's all hooks, right?

agrawalravi90:
@bayucan  and @vccardenas I'm happy with the changes, but i do think @lisa-altair is correct and we should modify the title of the PR before we merge this in. Thanks.

vccardenas:
In server.revert_to_defaults if reverthooks is True then the hooks selection:
    hooks = self.status(HOOK, level=logging.DEBUG)
will not include the PBS hooks which may selected using
    self.status(PBS_HOOK, level=logging.DEBUG)

Before and after the changes, the selection for PBS hooks does not appear.

bayucan:
@vccardenas : It's fine that we don't include PBS hooks in revert_hook() since they can't be deleted anyway:
Qmgr: d h PBS_alps_inventory_check
qmgr obj=PBS_alps_inventory_check svr=default: cannot delete a PBS hook
q

bayucan:
@vccardenas : and sure, revert_hooks() could disable the hook instead of deleting, but I don't think PBS hooks should be disabled automatically by revert_hooks(). They are shipped with PBS, and even though all current PBS hooks are shipped enabled=false, a future PBS hook might be shipped with enabled=True, to patch a bug. If a test really manipulates the 'enabled' value of a PBS hook, that test should take care or reversing it.

lisa-altair:
@vccardenas Thanks for the reminder that this isn't affecting the pbshooks.  I am fine with the PR title then.

PR #1979, opened 2020-08-19
Increased reservation start time for test_advance_reservation to fix race condition

birbalsain:
#### Describe Bug or Feature
Test "test_advance_reservation" doesn't catch reservation confirmation message. Reservation submission and confirmation takes too long on slow machines, reservation comes in running state.

#### Describe Your Change

Increased reservation start time

[advance_reservation_fail.txt](https://github.com/openpbs/openpbs/files/5102531/advance_reservation_fail.txt)

[advance_reservation.txt](https://github.com/openpbs/openpbs/files/5109220/advance_reservation.txt)

### DIFF
--- a/test/tests/pbs_smoketest.py
+++ b/test/tests/pbs_smoketest.py
@@ -103,7 +103,8 @@ class SmokeTest(PBSTestSuite):
         jid2 = self.server.submit(j2)
 
         a = {'reserve_state': (MATCH_RE, "RESV_RUNNING|5")}
-        self.server.expect(RESV, a, id=rid, offset=20, interval=1)
+        self.server.expect(RESV, a, id=rid, interval=1,
+                           offset=(now + 30 - int(time.time())))
         self.server.expect(JOB, {'job_state': 'R'}, jid1)
         self.server.expect(JOB, {'job_state': 'B'}, jid2)
 
--- a/test/tests/pbs_smoketest.py
+++ b/test/tests/pbs_smoketest.py
@@ -83,7 +83,7 @@ class SmokeTest(PBSTestSuite):
         r = Reservation(TEST_USER)
         now = int(time.time())
         a = {'Resource_List.select': '1:ncpus=4',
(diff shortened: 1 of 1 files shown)


shilpakodli:
@birbalsain Could you please post failure logs in description.

test/tests/pbs_smoketest.py
```diff
@@ -83,7 +83,7 @@ def test_advance_reservation(self):
         r = Reservation(TEST_USER)
         now = int(time.time())
         a = {'Resource_List.select': '1:ncpus=4',
-             'reserve_start': now + 10,
+             'reserve_start': now + 20,
```
bhroam, line 86: I suggest using 30 instead of 20.  Some of the testbed machines are really slow.

shilpakodli: I agree, That's why I wanted to see failure logs first.

shilpakodli: Also, add an offset of 20 seconds while checking reservation in RUNNING state at line number 106.

shilpakodli (changes requested):

shilpakodli (approved):
Thanks for making changes. LGTM.

test/tests/pbs_smoketest.py
```diff
@@ -103,7 +103,7 @@ def test_advance_reservation(self):
         jid2 = self.server.submit(j2)
 
         a = {'reserve_state': (MATCH_RE, "RESV_RUNNING|5")}
-        self.server.expect(RESV, a, id=rid, interval=1)
+        self.server.expect(RESV, a, id=rid, offset=20, interval=1)
```
bhroam, line 106: It's best to set a variable to the start time (now + 30) above.  You then set the offset here to start - time.time().  it will take into account the fact that we have slow machines in our testbed.

shilpakodli (approved):
Looks good to me.

PR #1981, opened 2020-08-19
Fix unprotected access of pjob->ji_hosts

riyazhakki:
#### Describe Bug or Feature

*Unprotected access of pjob->ji_hosts would lead to mom crash*

#### Describe Your Change

*Added a null check before accessing pjob->ji_hosts*

[SmokeTest.txt](https://github.com/openpbs/openpbs/files/5096182/SmokeTest.txt)


bayucan:
Merged #1981 to master.

PR #1983, opened 2020-08-19
Fix node bucket marked as free when job is running

vstumpf:
### Describe Bug or Feature
In simualtion, when a job finished, the scheduler would mark the node's bucket as free, even if there was another job running on the node. 

#### Describe Your Change
Only change the state of the bucket if there are no more jobs on the node.

[afterfix.txt](https://github.com/openpbs/openpbs/files/5099360/afterfix.txt)

### DIFF
--- a/test/tests/functional/pbs_calendaring.py
+++ b/test/tests/functional/pbs_calendaring.py
@@ -40,6 +40,7 @@
 
 import time
 from tests.functional import *
+from ptl.utils.pbs_logutils import PBSLogUtils
 
 
 class TestCalendaring(TestFunctional):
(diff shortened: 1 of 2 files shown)


test/tests/functional/pbs_calendaring.py
```diff
@@ -191,3 +191,45 @@ def test_topjob_fail(self):
...
+        jid3 = self.server.submit(j3)
+
+        self.server.expect(JOB, {'job_state': 'R'}, jid1)
+        self.server.expect(JOB, {'job_state': 'R'}, jid2)
+        self.server.expect(JOB, {'job_state': 'Q'}, jid3)
+        job1 = self.server.status(JOB, id=jid1)
+        job2 = self.server.status(JOB, id=jid2)
+        job3 = self.server.status(JOB, id=jid3)
+
+        stime = time.strptime(job2[0]['stime'], "%a %b %d %H:%M:%S %Y")
```
bhroam, line 230: If you don't want to jump through all the hoops of using strptime() and mktime(), you can use the log utils convert_datetime() function.  It takes an optional format.  Also, consider using %c instead of the longer format string.

vstumpf: I'd rather use strptime and mktime, in case convert_date_time changes I really don't want to have to deal with changing this later.
I will switch to using %c though

PR #1984, opened 2020-08-20
fix acl hosts on queues (using gss)

vchlum:
#### Describe Bug or Feature
Using GSS, the acl_hosts on the queue do not work. The reason is the preq->rq_host contains the realm and not the hostname itself. It means the acl_hosts is checked against the realm, not the hostname.

#### Describe Your Change
With Kerberos and GSS support, the job attribute "[Submit_Host](https://openpbs.atlassian.net/wiki/spaces/PD/pages/29655043/PP-468+Kerberos+support)" was introduced. This attribute holds the hostname of job submission. This value is independent of the used auth method. Therefore the attribute can safely substitute the preq->rq_host with non-gss auth and it will fix the issue while using GSS.

[ptl_acl_queue.log](https://github.com/openpbs/openpbs/files/5101331/ptl_acl_queue.log)

### DIFF
--- a/src/server/req_quejob.c
+++ b/src/server/req_quejob.c
@@ -1003,7 +1003,7 @@ req_quejob(struct batch_request *preq)
 	 * job structure and attributes already set up.
 	 */
 
-	rc = svr_chkque(pj, pque, preq->rq_host, MOVE_TYPE_Move);
+	rc = svr_chkque(pj, pque, pj->ji_wattr[(int)JOB_ATR_submit_host].at_val.at_str, MOVE_TYPE_Move);
 	if (rc) {
 		if (pj->ji_clterrmsg)
 			reply_text(preq, rc, pj->ji_clterrmsg);
--- a/test/tests/functional/pbs_acl_host_queue.py
+++ b/test/tests/functional/pbs_acl_host_queue.py
@@ -1,0 +1,94 @@
+# coding: utf-8
+
+# Copyright (C) 1994-2020 Altair Engineering, Inc.
+# For more information, contact Altair at www.altair.com.
+#
+# This file is part of both the OpenPBS software ("OpenPBS")
+# and the PBS Professional ("PBS Pro") software.
+#
+# Open Source License Information:
+#
+# OpenPBS is free software. You can redistribute it and/or modify it under
+# the terms of the GNU Affero General Public License as published by the
+# Free Software Foundation, either version 3 of the License, or (at your
+# option) any later version.
+#
+# OpenPBS is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Affero General Public
+# License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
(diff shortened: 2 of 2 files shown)


test/tests/functional/pbs_acl_host_queue.py
```diff
@@ -0,0 +1,94 @@
...
+from tests.functional import *
+
+
+class Test_acl_host_queue(TestFunctional):
+    """
+    This test suite is for testing the queue attributes acl_host_enable
+    and acl_hosts.
+    """
+
+    def setUp(self):
```
hirenvadalia, line 50: you can remove of this.

test/tests/functional/pbs_acl_host_queue.py
```diff
@@ -0,0 +1,94 @@
...
+        else:
+            self.assertFalse(True, "Queue is violating acl_hosts")
+
+    def test_acl_host_enable_allow(self):
+        """
+        Set acl_host_enable = True along with acl_hosts and check
+        whether or not a job can be submitted.
+        """
+        a = {"acl_host_enable": True,
+             "acl_hosts": self.mom.hostname}
```
hirenvadalia, line 84: I think you should use self.server.hostname here

hirenvadalia (changes requested):

hirenvadalia (approved):

src/server/req_quejob.c
```diff
@@ -1003,7 +1003,7 @@ req_quejob(struct batch_request *preq)
 	 * job structure and attributes already set up.
 	 */
 
-	rc = svr_chkque(pj, pque, preq->rq_host, MOVE_TYPE_Move);
+	rc = svr_chkque(pj, pque, pj->ji_wattr[(int)JOB_ATR_submit_host].at_val.at_str, MOVE_TYPE_Move);
```
bhroam, line 1006: Is it possible for this attribute to be unset?  Do you need to check if it is set first and then use preq if it isn't?

test/tests/functional/pbs_acl_host_queue.py
```diff
@@ -0,0 +1,87 @@
...
+                            self.server.default_queue)
+
+        j = Job(TEST_USER)
+        try:
+            self.server.submit(j)
+        except PbsSubmitError as e:
+            error_msg = "qsub: Access from host not allowed, or unknown host"
+            if e.msg[0] != error_msg:
+                raise self.failureException("qsub unexpected err message: " +
+                                            e.msg[0])
```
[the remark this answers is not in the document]
vstumpf: Why not just use self.assertEquals(e.msg[0], error_msg)

src/server/req_quejob.c
vchlum: It is set on job creation and it never changes. It is similar to JOB_ATR_ctime. The source for this attribute is conn->cn_physhost, which comes from request->rq_host (before changing to the realm). IMHO, it is safe to consider the attribute is always set.

test/tests/functional/pbs_acl_host_queue.py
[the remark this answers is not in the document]
vchlum: assertEquals() makes sense here. Thx.

PR #1989, opened 2020-08-22
Enforce pulling queue's limits when peering

vstumpf:
#### Describe Bug or Feature
When a cluster has local peering set up, when the scheduler pulls a job from the furnishing queue to the pulling queue, the job can bypass the pulling queue's limits.

Scheduler is considered as a manager, and for local move jobs, the manager can bypass queue limit checks. 

#### Describe Your Change
Make local move requests from a scheduler unprivileged. 

[afterfix.txt](https://github.com/openpbs/openpbs/files/5111793/afterfix.txt)

### DIFF
--- a/test/tests/functional/pbs_peer.py
+++ b/test/tests/functional/pbs_peer.py
@@ -142,6 +142,6 @@ class TestPeering(TestFunctional):
         s1.expect(JOB, {'job_state': 'Q'}, id=jid)
         msg = (jid + r';Failed to run: .* \(15039\)')
         self.scheduler.log_match(msg, regexp=True)
-        msg = jid + ';send of job to workq@shecil:15001 failed error = 15036'
-        s2.log_match(msg)
+        msg = jid + ';send of job to workq@.* failed error = 15036'
+        s2.log_match(msg, regexp=True)
         s1.expect(JOB, {'job_state': 'R'}, id=jid2)
--- a/test/tests/functional/pbs_peer.py
+++ b/test/tests/functional/pbs_peer.py
@@ -1,0 +1,147 @@
+# coding: utf-8
+
+# Copyright (C) 1994-2020 Altair Engineering, Inc.
+# For more information, contact Altair at www.altair.com.
+#
+# This file is part of both the OpenPBS software ("OpenPBS")
+# and the PBS Professional ("PBS Pro") software.
+#
+# Open Source License Information:
+#
+# OpenPBS is free software. You can redistribute it and/or modify it under
+# the terms of the GNU Affero General Public License as published by the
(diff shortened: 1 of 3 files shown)


test/tests/functional/pbs_peer.py
```diff
@@ -0,0 +1,147 @@
...
+        s1.manager(MGR_CMD_SET, SERVER, a)
+        s2.manager(MGR_CMD_SET, SERVER, a)
+
+        a = {'peer_queue': '"workq workq@' + s2.hostname + '"'}
+        self.scheduler.set_sched_config(a)
+
+        a = {'Resource_List.' + self.cres: 100,
+             ATTR_queue: 'workq@' + s2.hostname}
+        j = Job(TEST_USER, attrs=a)
+        jid = s1.submit(j)
```
arungrover, line 132: If s1 is the pulling complex then shouldn't this be s2.submit()?

test/tests/functional/pbs_peer.py
```diff
@@ -0,0 +1,147 @@
...
+        a['Resource_List.' + self.cres] = 1
+        j = Job(TEST_USER, attrs=a)
+        jid2 = s1.submit(j)
```
arungrover, line 135: Same here

test/tests/functional/pbs_peer.py
vstumpf: PTL is a little screwy and won't let me submit from s2. However, because I set the queue to workq@s2, it will submit the job from s1's host to s2.

vstumpf: To be more clear, the first job submits just fine, the second fails with "Failed to receive data from background qsub" or something like that.

test/tests/functional/pbs_peer.py
```diff
@@ -0,0 +1,147 @@
...
+    def test_local_resc_limits(self):
+        """
+        Test that a local peering queue enforces new limits
+        """
+        self.create_resource()
+        self.create_queue()
+        a = {'resources_max.' + self.cres: 4}
+        self.server.manager(MGR_CMD_SET, QUEUE, a, id=self.cqueue)
+        a = {'started': False}
+        self.server.manager(MGR_CMD_SET, QUEUE, a, id='workq')
```
bhroam, line 85: Instead of stopping the queues, just turn scheduling off.  You need to set it to true to kick the cycle anyway, so you might as well kill two birds with one  stone. (here and below)

vstumpf: Don't I need to stop the pulled queue? Otherwise the scheduler will schedule and run jobs from that queue.

PR #1990, opened 2020-08-22
sched fails to reconfirm running degraded resv

arungrover:
#### Describe Bug or Feature
Sometimes when a multi-chunk running reservation goes into degraded state scheduler fails to reconfirm the reservation back.

#### Describe Your Change
What happens is that when one of the vnode that's associated to one of the chunk of a running reservation goes down, the scheduler tries to reconfirm this reservation buy creating a new execselect where it tries to retain the vnodes that are up and associated with the reservation and tries to replace the vnodes that are down.
In this process scheduler first start finding the solution for a chunk whose vnode is down. While doing so it finds out a vnode that is associated with one of the other chunks of the reservation. It allocates that node to the current chunk and moves on to the next chunk only to find out that the node next chunk had been running on, now has been allocated to the another chunk earlier. This results into a failure to reconfirm a running reservation.

In this change, we try to sort all the chunks associated with up and running nodes in front of the reservation's execselect. This makes scheduler assign nodes to chunks they were already associated with and helps to find new nodes for chunks associated with node that are in down state. Once we have the solution, we sort the nodes back to the way they were supposed to be (in sync with the schedselect).

#### Link to Design Doc
NA

[test_ralter.txt](https://github.com/openpbs/openpbs/files/5111830/test_ralter.txt)
[test_reservation.txt](https://github.com/openpbs/openpbs/files/5111831/test_reservation.txt)

### DIFF
--- a/src/scheduler/resv_info.c
+++ b/src/scheduler/resv_info.c
@@ -1524,7 +1524,7 @@ confirm_reservation(status *policy, int pbs_sd, resource_resv *unconf_resv, serv
 					 */
 					sel = create_select_from_nspec(nresv->orig_nspec_arr);
 					nresv->execselect = parse_selspec(sel);
-					for (ind=0; nresv->orig_nspec_arr[ind] != NULL; ind++) {
+					for (ind = 0; nresv->orig_nspec_arr[ind] != NULL; ind++) {
 					    nresv->execselect->chunks[ind]->seq_num = nresv->orig_nspec_arr[ind]->seq_num;
 					}
 
@@ -1594,9 +1594,7 @@ confirm_reservation(status *policy, int pbs_sd, resource_resv *unconf_resv, serv
 		if (!(simrc & TIMED_ERROR) && resv_start_time >= 0) {
 			clear_schd_error(err);
 			if ((ns = is_ok_to_run(nsinfo->policy, nsinfo, NULL, nresv, NO_ALLPART, err)) != NULL) {
-				if (vnodes_down == 1)
-					/* if we are confirming a running reservation, sort nspec array by seq number */
-					qsort(ns, count_array(ns), sizeof(nspec *), cmp_nspec);
+				qsort(ns, count_array(ns), sizeof(nspec *), cmp_nspec);
 				tmp = create_execvnode(ns);
 				free_nspecs(ns);
 				if (tmp == NULL) {
@@ -2043,8 +2041,12 @@ check_vnodes_unavailable(resource_resv *resv)
 			remove_ptr_from_array(resv->orig_nspec_arr, chunks_to_remove[i]);
 		}
 	}
-	/* move all nspec with down nodes at the back of orig_ncpec_arr */
-	qsort(resv->orig_nspec_arr, count_array(resv->orig_nspec_arr), sizeof(nspec *), cmp_nspec_by_sub_seq);
+	if (has_down_node == 1)
+		/* Move all the specific chunks ahead of the generic chunks.
+		 * This will ensure that we get all the nodes back we need to,
+		 * before we start looking for replacements
+		 */
+		qsort(resv->orig_nspec_arr, count_array(resv->orig_nspec_arr), sizeof(nspec *), cmp_nspec_by_sub_seq);
 
 	free(chunks_to_remove);
 
--- a/src/scheduler/resv_info.c
+++ b/src/scheduler/resv_info.c
@@ -1516,6 +1516,7 @@ confirm_reservation(status *policy, int pbs_sd, resource_resv *unconf_resv, serv
 				nresv->resv->resv_state == RESV_BEING_ALTERED) {
 				if (nresv->resv->is_running) {
 					char *sel;
+					int ind;
 					free_selspec(nresv->execselect);
 					/* Use orig_nspec_arr over nspec_arr because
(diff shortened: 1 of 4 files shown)


src/scheduler/sort.h
```diff
@@ -58,6 +58,11 @@ int cmpres(sch_resource_t r1, sch_resource_t r2);
  */
 int cmp_nspec(const void *v1, const void *v2);
 
+/*
+ *cmp_nspec_by_sub_seq_dsc - sort nspec by sub sequence number
+ */
+int
+cmp_nspec_by_sub_seq(const void *v1, const void *v2);
```
bhroam, line 65: Declarations are all on one line.

test/tests/functional/pbs_reservations.py
```diff
@@ -2270,3 +2270,45 @@ def test_resv_job_hard_walltime(self):
         J = Job(TEST_USER, attrs=a)
         jid = self.server.submit(J)
         self.server.expect(JOB, {'job_state': 'R'}, id=jid)
+
+    def test_resv_reconfirm_holding_partial_nodes(self):
+        """
+        Test that scheduler is able to reconfirm a reservation when
+        only of the nodes reservation was running on goes down. Also
```
bhroam, line 2277: I think there is a word missing between only and of

test/tests/functional/pbs_reservations.py
```diff
@@ -2270,3 +2270,45 @@ def test_resv_job_hard_walltime(self):
...
+        self.server.manager(MGR_CMD_SET, SERVER, a)
+        a = {'resources_available.ncpus': 2}
+        self.server.create_vnodes('vn', a, num=3, mom=self.mom)
+        vn_list = ['vn[0]', 'vn[1]', 'vn[2]']
+
+        now = int(time.time())
+        sel = '1:ncpus=2+1:ncpus=1'
+        rid = self.submit_reservation(user=TEST_USER, select=sel,
+                                      start=now + 5, end=now + 300)
+
```
bhroam, line 2290: If you want to truly follow the pep8 standard, it says to put spaces around your operators except in function calls.  It doesn't really matter though.

test/tests/functional/pbs_reservations.py
```diff
@@ -2270,3 +2270,45 @@ def test_resv_job_hard_walltime(self):
...
+        a = {'reserve_state': (MATCH_RE, 'RESV_RUNNING|5')}
+        self.server.expect(RESV, a, id=rid)
+
+        self.server.status(RESV, 'resv_nodes', id=rid)
+        resv_node_list = self.server.reservations[rid].get_vnodes()
+        resv_node = resv_node_list[0]
+        resv_node2 = resv_node_list[1]
+        print(resv_node_list)
+        print(vn_list)
```
bhroam, line 2299: Remove these debug statements.

src/scheduler/resv_info.c
```diff
@@ -2035,8 +2043,8 @@ check_vnodes_unavailable(resource_resv *resv)
 			remove_ptr_from_array(resv->orig_nspec_arr, chunks_to_remove[i]);
 		}
 	}
-
-	qsort(resv->orig_nspec_arr, count_array(resv->orig_nspec_arr), sizeof(nspec *), cmp_nspec);
+	/* move all nspec with down nodes at the back of orig_ncpec_arr */
```
bhroam, line 2046: While this is true, it doesn't explain why you are doing it.  Maybe something more like: "Move all the specific chunks ahead of the generic chunks.  This will ensure that we get all the nodes back we need to, before we start looking for replacements"

src/scheduler/resv_info.c
```diff
@@ -1589,6 +1594,9 @@ confirm_reservation(status *policy, int pbs_sd, resource_resv *unconf_resv, serv
 		if (!(simrc & TIMED_ERROR) && resv_start_time >= 0) {
 			clear_schd_error(err);
 			if ((ns = is_ok_to_run(nsinfo->policy, nsinfo, NULL, nresv, NO_ALLPART, err)) != NULL) {
+				if (vnodes_down == 1)
```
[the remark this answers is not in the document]
arungrover: But we don't sort in the cases when check_vnodes_unavailable returns -1 or -2. 
I think maybe I should add a check in check_vnodes_unavailable() to sort only when we see a vnode unavailable and avoid sorting when all vnodes are up.
Since we don't sort in all cases, I thought we don't have to re-sort to bring nspec back in the original order.

test/tests/functional/pbs_reservations.py
arungrover: oops

arungrover:
[test_reservation.txt](https://github.com/openpbs/openpbs/files/5125976/test_reservation.txt)

src/scheduler/sort.h
```diff
@@ -58,6 +58,10 @@ int cmpres(sch_resource_t r1, sch_resource_t r2);
  */
 int cmp_nspec(const void *v1, const void *v2);
 
+/*
+ *cmp_nspec_by_sub_seq_dsc - sort nspec by sub sequence number
```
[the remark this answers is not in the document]
arungrover: ah, good catch. Initially, I thought sub-seq-num is set to 0 for nspec with down nodes and then I realized it was being set to 0.
I have changed the name.

PR #1993, opened 2020-08-24
Fix server crash on job recovery

vstumpf:
#### Describe Bug or Feature
If `restrict_res_to_release_on_suspend` is set and a job is suspended, it creates and sets `resources_released` on the job. If `restrict_res_to_release_on_suspend` is unset and the server restarts, the server crashes in recovering the job.
It tries to access the values of `restrict_res_to_release_on_suspend`.

#### Describe Your Change
Check if `restrict_res_to_release_on_supsend` is set before accessing it.

[after-fix.txt](https://github.com/openpbs/openpbs/files/5120698/after-fix.txt)
[before-fix.txt](https://github.com/openpbs/openpbs/files/5120699/before-fix.txt)

### DIFF
--- a/src/server/node_manager.c
+++ b/src/server/node_manager.c
@@ -7854,30 +7854,31 @@ int update_resources_rel(job *pjob, attribute *attrib, enum batch_op op)
 			return 1;
 	}
 	/* Now iterate through all of the job resources that are present on at
-	 * queue/server level and add them to resource_release_list.
+	 * queue/server level and add them to resource_release_list. Only do this if
+	 * restrict_res_to_release_on_suspend is set
 	 */
-	presc_sq = (resource *) GET_NEXT(pjob->ji_wattr[(int) JOB_ATR_resource].at_val.at_list);
-	for (;presc_sq != NULL; presc_sq = (resource *)GET_NEXT(presc_sq->rs_link)) {
-		prdef = presc_sq->rs_defin;
-		/* make sure it is a server/queue level consumable resource and not
-		 * set in resource_released_list already
-		 */
-		if ((prdef->rs_flags & ATR_DFLAG_RASSN) &&
-			server.sv_attr[(int)SVR_ATR_restrict_res_to_release_on_suspend].at_flags & ATR_VFLAG_SET &&
-			(find_resc_entry(&pjob->ji_wattr[(int) JOB_ATR_resc_released_list], prdef) == NULL)) {
-			for (j = 0; j < server.sv_attr[(int)SVR_ATR_restrict_res_to_release_on_suspend].at_val.at_arst->as_usedptr; j++) {
-				if (strcmp(server.sv_attr[(int)SVR_ATR_restrict_res_to_release_on_suspend].at_val.at_arst->as_string[j],
-				    prdef->rs_name) == 0) {
-					presc = add_resource_entry(&pjob->ji_wattr[(int) JOB_ATR_resc_released_list], prdef);
-					if (presc == NULL)
-						return 1;
-					prdef->rs_set(&presc->rs_value, &presc_sq->rs_value, op);
-					break;
+	if (server.sv_attr[(int)SVR_ATR_restrict_res_to_release_on_suspend].at_flags & ATR_VFLAG_SET) {
+		presc_sq = (resource *) GET_NEXT(pjob->ji_wattr[(int) JOB_ATR_resource].at_val.at_list);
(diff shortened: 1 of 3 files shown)


test/tests/functional/pbs_release_limited_res_suspend.py
```diff
@@ -1059,3 +1059,56 @@ def test_server_restart_with_suspened_job(self):
         self.server.sigjob(jobid=jid1, signal="resume")
         self.server.expect(JOB, {ATTR_state: 'R'}, id=jid1)
         self.server.expect(NODE, 'jobs', op=SET, id=self.mom.shortname)
+
+    def test_server_restart_with_suspened_job_unset(self):
+        """
+        Test that when the attribute is set and unset,
+        the server does not crash on restart with a suspended job.
+        """
+        # Set ncpus in restrict_res_to_release_on_suspend server attribute
```
bhroam, line 1068: All the comments in this test explain what the code does.  The code does that.  Either explain why you are doing something or remove them.

src/server/node_manager.c
```diff
@@ -7854,29 +7854,31 @@ int update_resources_rel(job *pjob, attribute *attrib, enum batch_op op)
...
-			(find_resc_entry(&pjob->ji_wattr[(int) JOB_ATR_resc_released_list], prdef) == NULL)) {
-			for (j = 0; j < server.sv_attr[(int)SVR_ATR_restrict_res_to_release_on_suspend].at_val.at_arst->as_usedptr; j++) {
-				if (strcmp(server.sv_attr[(int)SVR_ATR_restrict_res_to_release_on_suspend].at_val.at_arst->as_string[j],
-				    prdef->rs_name) == 0) {
-					presc = add_resource_entry(&pjob->ji_wattr[(int) JOB_ATR_resc_released_list], prdef);
-					if (presc == NULL)
-						return 1;
-					prdef->rs_set(&presc->rs_value, &presc_sq->rs_value, op);
-					break;
+	if (server.sv_attr[(int)SVR_ATR_restrict_res_to_release_on_suspend].at_flags & ATR_VFLAG_SET) {
```
arungrover, line 7860: doing this will make it troublesome to recover suspended jobs with limited released resources.

src/server/node_manager.c
```diff
@@ -7854,29 +7854,31 @@ int update_resources_rel(job *pjob, attribute *attrib, enum batch_op op)
...
+			prdef = presc_sq->rs_defin;
+			/* make sure it is a server/queue level consumable resource and not
+			* set in resource_released_list already
+			*/
+			if ((prdef->rs_flags & ATR_DFLAG_RASSN) &&
+				(find_resc_entry(&pjob->ji_wattr[(int) JOB_ATR_resc_released_list], prdef) == NULL)) {
+				for (j = 0; j < server.sv_attr[(int)SVR_ATR_restrict_res_to_release_on_suspend].at_val.at_arst->as_usedptr; j++) {
+					if (strcmp(server.sv_attr[(int)SVR_ATR_restrict_res_to_release_on_suspend].at_val.at_arst->as_string[j],
+						prdef->rs_name) == 0) {
+						presc = add_resource_entry(&pjob->ji_wattr[(int) JOB_ATR_resc_released_list], prdef);
```
arungrover, line 7872: add_sresource_entry only adds modcache bit, I think that's the reason this attribute is not saved into db.  So either we should add a modify flag in this function.
Also, if possible update post_signal_job to release resources before setting job state because then we can save this attribute along with job state into db.

vstumpf: No, it sets ATR_SET_MOD_MCACHE, which is set + mod + modcache.
See: https://github.com/openpbs/openpbs/blob/69d4e0258bd62541409305e3a0bd8b083dc52149/src/lib/Libattr/attr_fn_resc.c#L591

src/server/node_manager.c
vstumpf: I don't understand this, the test shows the job was recovered with limited released resources

arungrover: This check is covering all RASSN resources not node level resources. So I think there will be a problem with server/queue level resources.
I think the test might be passing only because  
- mem is a RASSN resource
- you are checking node resources and not server/queue resources.

vstumpf: I've updated the test to check a server-level resource

arungrover:
Thanks for making the change @vstumpf

PR #1996, opened 2020-08-25
Install pbs and pbs_ifl as loadable Python modules

toonen:
#### Describe Bug or Feature

These changes build and install the pbs and pbs_ifl Python modules as loadable shared library modules.  This allows a standard Python 3.x interpreter to use either pbs or pbs_ifl, which is desirable for external Python programs that need to perform actions such as monitoring queued/running jobs or manipulate reservations.  This also allows PTL tests to import either module to obtain constants defined by those modules and them in the validation portion of the test.

#### Describe Your Change

The loadable pbs_ifl module is generated from the same SWIG generated code that is statically incorporated in the PBS server and pbs_python code.  The loadable pbs module is created in part by a new C module, src/modules/python/pbs_v1_module_init.c, that uses techniques similar to those used by pbs_python.  Both modules are installed in %(libdir)/python/altair.

Two additional smoke tests were added.  Please see `test_import_pbs_module` and `test_import_pbs_ifl_module` in test/tests/pbs_smoketest.py.

[install-pbs-py-smoketest.log](https://github.com/openpbs/openpbs/files/5125854/install-pbs-py-smoketest.log)


src/modules/python/pbs_v1_module_init.c
```diff
@@ -0,0 +1,815 @@
...
+		PyDict_DelItemString(py_sys_modules,
+			PBS_PYTHON_V1_MODULE_EXTENSION_NAME);
+		return PyErr_Format(PyExc_Exception,
+			"pbs_python_load_python_types() failed to load Python types");
+	}
+
+	return module;
+}
+
+/*
```
bayucan, line 188: The following "dummy" variable declarations and function definitions are the same as the ones in the standalone src/tools/pbs_python.c. Please merge them somehow, so that there'll be only one place to update or add to, when new functions are introduced.

src/lib/Libifl/Makefile.am
```diff
@@ -0,0 +1,46 @@
...
+# Use of Altair’s trademarks, including but not limited to "PBS™",
+# "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's
+# trademark licensing policies.
+#
+
+BUILT_SOURCES = pbs_ifl_wrap.c
+EXTRA_DIST = pbs_ifl.i
+
+pbs_ifl_wrap.c: pbs_ifl.i
+	@swig_dir@/bin/swig -python -outcurrentdir -I$(top_srcdir)/src/include $^
```
hirenvadalia, line 42: I think we need to include `@swig_py_inc@` in this command

test/fw/ptl.sh
```diff
@@ -71,6 +71,8 @@ else
 			python_dir=$( /bin/ls -1 ${PTL_PREFIX}/lib )/site-packages
 			[ -d "${PTL_PREFIX}/bin" ] && export PATH="${PATH}:${PTL_PREFIX}/bin"
 			[ -d "${PTL_PREFIX}/lib/${python_dir}" ] && export PYTHONPATH="${PYTHONPATH}:${PTL_PREFIX}/lib/${python_dir}"
+			[ -d "${__PBS_EXEC}/lib/python/altair" ] && export PYTHONPATH="${PYTHONPATH}:${__PBS_EXEC}/lib/python/altair"
```
hirenvadalia, line 74: with only this change it will force PTL to go in PTL_API mode. we need to remove [this](https://github.com/openpbs/openpbs/blob/master/test/fw/ptl/lib/pbs_testlib.py#L4762) if condition so by default PTL go in PTL_CLI mode (which is current default)

hirenvadalia (changes requested):

src/modules/python/Makefile.am
```diff
@@ -39,14 +39,79 @@
...
+	-module \
+	-shared \
+	-export-dynamic \
+	-avoid-version \
+	-no-undefined \
+	@PYTHON_LDFLAGS@
+_pbs_ifl_la_LIBADD = \
+	$(top_builddir)/src/lib/Libpbs/libpbs.la \
+	@PYTHON_LIBS@
+
```
bayucan, line 107: I tried compiling your code, but was getting an error under src/modules/python:
```
src/modules/python> make
/bin/sh ../../../libtool  --tag=CC   --mode=link gcc  -g -Wall -Werror -Wno-unused-result -fPIC -D_LARGEFILE_SOURCE -O2     -module -shared -export-dynamic -avoid-version -L/home/bayucan/2021.1.0/2020.1.1.20200901012250/linux312_x86_64-work/python/lib/python3.6/config-3.6m-x86_64-linux-gnu -L/home/bayucan/2021.1.0/2020.1.1.20200901012250/linux312_x86_64-work/python/lib -lpython3.6m -lpthread -ldl  -lutil -lm  -Xlinker -export-dynamic -L/home/bayucan/2021.1.0/2020.1.1.20200901012250/linux312_x86_64-work/tcltk/lib -L/home/bayucan/2021.1.0/2020.1.1.20200901012250/linux312_x86_64-work/alsdk/lib/linux_x64 -o _pbs_v1.la -rpath /opt/pbs/lib/python/altair _pbs_v1_la-attr_fn_acl.lo _pbs_v1_la-attr_fn_entlim.lo _pbs_v1_la-attr_fn_resc.lo _pbs_v1_la-attr_fn_time.lo _pbs_v1_la-attr_node_func.lo _pbs_v1_la-job_attr_def.lo _pbs_v1_la-node_attr_def.lo _pbs_v1_la-queue_attr_def.lo _pbs_v1_la-resc_def_all.lo _pbs_v1_la-resv_attr_def.lo _pbs_v1_la-sched_attr_def.lo _pbs_v1_la-svr_attr_def.lo _pbs_v1_la-shared_python_utils.lo _pbs_v1_la-common_python_utils.lo _pbs_v1_la-pbs_python_external.lo _pbs_v1_la-pbs_python_svr_external.lo _pbs_v1_la-module_pbs_v1.lo _pbs_v1_la-pbs_python_svr_internal.lo _pbs_v1_la-pbs_python_svr_size_type.lo _pbs_v1_la-pbs_python_import_types.lo _pbs_v1_la-entlim.lo _pbs_v1_la-hook.lo _pbs_v1_la-work_task.lo _pbs_v1_la-svr_attr.lo _pbs_v1_la-resc_attr.lo _pbs_v1_la-setup_resc.lo _pbs_v1_la-vnparse.lo _pbs_v1_la-pbs_v1_module_init.lo ../../../src/lib/Libpbs/libpbs.la -lpython3.6m -lpthread -ldl  -lutil -lm  -ldl -lcrypt -lc -lc 
libtool: link: gcc -shared  -fPIC -DPIC  .libs/_pbs_v1_la-attr_fn_acl.o .libs/_pbs_v1_la-attr_fn_entlim.o .libs/_pbs_v1_la-attr_fn_resc.o .libs/_pbs_v1_la-attr_fn_time.o .libs/_pbs_v1_la-attr_node_func.o .libs/_pbs_v1_la-job_attr_def.o .libs/_pbs_v1_la-node_attr_def.o .libs/_pbs_v1_la-queue_attr_def.o .libs/_pbs_v1_la-resc_def_all.o .libs/_pbs_v1_la-resv_attr_def.o .libs/_pbs_v1_la-sched_attr_def.o .libs/_pbs_v1_la-svr_attr_def.o .libs/_pbs_v1_la-shared_python_utils.o .libs/_pbs_v1_la-common_python_utils.o .libs/_pbs_v1_la-pbs_python_external.o .libs/_pbs_v1_la-pbs_python_svr_external.o .libs/_pbs_v1_la-module_pbs_v1.o .libs/_pbs_v1_la-pbs_python_svr_internal.o .libs/_pbs_v1_la-pbs_python_svr_size_type.o .libs/_pbs_v1_la-pbs_python_import_types.o .libs/_pbs_v1_la-entlim.o .libs/_pbs_v1_la-hook.o .libs/_pbs_v1_la-work_task.o .libs/_pbs_v1_la-svr_attr.o .libs/_pbs_v1_la-resc_attr.o .libs/_pbs_v1_la-setup_resc.o .libs/_pbs_v1_la-vnparse.o .libs/_pbs_v1_la-pbs_v1_module_init.o   -Wl,-rpath -Wl,/home/bayucan/2021.1.0/2020.1.1.20200901012250/linux312_x86_64-work.orig/src/lib/Libpbs/.libs -Wl,-rpath -Wl,/opt/pbs/lib -L/home/bayucan/2021.1.0/2020.1.1.20200901012250/linux312_x86_64-work/python/lib/python3.6/config-3.6m-x86_64-linux-gnu -L/home/bayucan/2021.1.0/2020.1.1.20200901012250/linux312_x86_64-work/python/lib -L/home/bayucan/2021.1.0/2020.1.1.20200901012250/linux312_x86_64-work/tcltk/lib -L/home/bayucan/2021.1.0/2020.1.1.20200901012250/linux312_x86_64-work/alsdk/lib/linux_x64 ../../../src/lib/Libpbs/.libs/libpbs.so -lz -lcrypto -lpython3.6m -lpthread -lutil -lm -ldl -lcrypt -lc  -O2 -Wl,-export-dynamic   -Wl,-soname -Wl,_pbs_v1.so -o .libs/_pbs_v1.so
/usr/lib64/gcc/x86_64-suse-linux/4.8/../../../../x86_64-suse-linux/bin/ld: /home/bayucan/2021.1.0/2020.1.1.20200901012250/linux312_x86_64-work/python/lib/python3.6/config-3.6m-x86_64-linux-gnu/libpython3.6m.a(abstract.o): relocation R_X86_64_32S against `_Py_NotImplementedStruct' can not be used when making a shared object; recompile with -fPIC
/home/bayucan/2021.1.0/2020.1.1.20200901012250/linux312_x86_64-work/python/lib/python3.6/config-3.6m-x86_64-linux-gnu/libpython3.6m.a: error adding symbols: Bad value
collect2: error: ld returned 1 exit status
Makefile:665: recipe for target '_pbs_v1.la' failed
make: *** [_pbs_v1.la] Error 1
```
I was able to get around this with this Makefile.am fix:
```
diff -c Makefile.am.orig Makefile.am
*** Makefile.am.orig    2020-09-14 13:30:35.169228291 -0700
--- Makefile.am 2020-09-14 14:42:46.649020870 -0700
***************
*** 84,94 ****
        -module \
        -shared \
        -export-dynamic \
!       -avoid-version \
!       @PYTHON_LDFLAGS@
  _pbs_v1_la_LIBADD = \
!       $(top_builddir)/src/lib/Libpbs/libpbs.la \
!       @PYTHON_LIBS@
  
  nodist__pbs_ifl_la_SOURCES = $(top_builddir)/src/lib/Libifl/pbs_ifl_wrap.c
  _pbs_ifl_la_CPPFLAGS = \
--- 84,92 ----
        -module \
        -shared \
        -export-dynamic \
!       -avoid-version
  _pbs_v1_la_LIBADD = \
!       $(top_builddir)/src/lib/Libpbs/libpbs.la
  
  nodist__pbs_ifl_la_SOURCES = $(top_builddir)/src/lib/Libifl/pbs_ifl_wrap.c
  _pbs_ifl_la_CPPFLAGS = \
***************
*** 99,109 ****
        -shared \
        -export-dynamic \
        -avoid-version \
!       -no-undefined \
!       @PYTHON_LDFLAGS@
  _pbs_ifl_la_LIBADD = \
!       $(top_builddir)/src/lib/Libpbs/libpbs.la \
!       @PYTHON_LIBS@
  
  pbsmoduledir = $(libdir)/python/altair/pbs
  
--- 97,105 ----
        -shared \
        -export-dynamic \
        -avoid-version \
!       -no-undefined
  _pbs_ifl_la_LIBADD = \
!       $(top_builddir)/src/lib/Libpbs/libpbs.la
  
  pbsmoduledir = $(libdir)/python/altair/pbs
  
```

test/tests/pbs_smoketest.py
```diff
@@ -1462,3 +1462,30 @@ def test_jobscript_max_size(self):
...
+        is able to be loaded and a connection to the server can be established.
+        This test requires the PYTHONPATH environment variable contain the path
+        PBS_EXEC/{lib,lib64}/python/altair.
+        """
+        import pbs_ifl
+        server_conn = pbs_ifl.pbs_connect(None)
+        server_stat = pbs_ifl.pbs_statserver(server_conn, None, None)
+        pbs_ifl.pbs_disconnect(server_conn)
+        msg = "server name is %s" % (server_stat.name,)
+        self.logger.info(msg)
```
bayucan, line 1491: @checkmodule() test may not be reliable,  so I suggest doing:
```
 diff -c pbs_smoketest.py.orig pbs_smoketest.py
*** pbs_smoketest.py.orig       2020-09-15 01:29:21.387629113 -0400
--- pbs_smoketest.py    2020-09-15 01:29:34.083629310 -0400
***************
*** 1463,1469 ****
          self.server.log_match("Req;req_reject;Reject reply code=15175",
                                max_attempts=5)
  
-     @checkModule("pbs")
      def test_import_pbs_module(self):
          """
          Test that the pbs module located in the PBS installation directory is
--- 1463,1468 ----
***************
*** 1471,1481 ****
          the PYTHONPATH environment variable contain the path
          PBS_EXEC/{lib,lib64}/python/altair.
          """
          import pbs
          msg = "pbs.JOB_STATE_RUNNING=%s" % (pbs.JOB_STATE_RUNNING,)
          self.logger.info(msg)
  
-     @checkModule("pbs_ifl")
      def test_import_pbs_ifl_module(self):
          """
          Test that the pbs_ifl module located in the PBS installation directory
--- 1470,1482 ----
          the PYTHONPATH environment variable contain the path
          PBS_EXEC/{lib,lib64}/python/altair.
          """
+         sys.path.append(os.path.join(self.server.pbs_conf['PBS_EXEC'], 'lib',
+                              'python', 'altair'))
          import pbs
          msg = "pbs.JOB_STATE_RUNNING=%s" % (pbs.JOB_STATE_RUNNING,)
          self.logger.info(msg)
  
      def test_import_pbs_ifl_module(self):
          """
          Test that the pbs_ifl module located in the PBS installation directory
***************
*** 1483,1488 ****
--- 1484,1492 ----
          This test requires the PYTHONPATH environment variable contain the path
          PBS_EXEC/{lib,lib64}/python/altair.
          """
+         sys.path.append(os.path.join(self.server.pbs_conf['PBS_EXEC'], 'lib',
+                              'python', 'altair'))
          import pbs_ifl
          server_conn = pbs_ifl.pbs_connect(None)
          server_stat = pbs_ifl.pbs_statserver(server_conn, None, None)
```

bayucan (changes requested):

src/modules/python/pbs_v1_module_init.c
toonen: I agree.  And my apologies for the delay in responding.  I was on vacation for a couple of weeks and then playing catch-up.  I will strip out the common code as much as seems reasonable and push an update once that is done.

test/fw/ptl.sh
toonen: As I understand Python modules, this should not be an issue.  The PTL modules explicitly import `ptl.lib.pbs_ifl`.  The Python import mechanism shouldn't ever match the module in the PBS_EXEC directory space given that the ptl/lib directory does not exist in that space.  Did I overlook something?  Granted I am making the assumption that ptl/lib doesn't exist in path I added; and while presently true, I'm happy to address the issue further if you have suggestions on further ways to mitigate a possible inappropriate match and import.

test/tests/pbs_smoketest.py
toonen: I'm more than happy to make the changes you suggest; however, I must ask, if @checkModule() is unreliable. then shouldn't it be either fixed or eliminated.  I ask because I used other test code that made use of this decorator as an example, so any issues with the decorator may extend beyond the tests I added.

src/modules/python/Makefile.am
toonen: Any information you can provide to replicate the link error would be greatly appreciated. 
 Please let me know which version of autoconf, automake and libtool you were using.  Also, were those versions of the autotools the default installation for a specific OS and version?  If so, which OS and version?

bayucan: I compiled it under SLES12, GNU Autoconf 2.69, libtool.m4 (macro_version=2.4.2
macro_revision=1.3337), automake (GNU automake) 1.13.4

test/tests/pbs_smoketest.py
bayucan: @toonen : I would have to check what's going on with @checkModule. Maybe it's assuming that the environment where pbs_benchpress is invoked already contains path $PBS_EXEC/lib/python/altair.  Maybe I didn't have that in my path when I invoked the testsuite, and it kept failing on @checkModule. When took out that check, and explictly added the path in the test case itself, it worked. This sounds more like a fail-safe way of running the test...

test/fw/ptl.sh
hirenvadalia: @toonen You are right, sorry I overlooked, it won't cause any problem as we don't have ptl/lib in PBS_EXEC/lib/python/altair. No need of any changes for this.

src/modules/python/Makefile.am
bayucan: and linked against Python 3.6.

src/modules/python/pbs_v1_module_init.c
toonen: I attempted to address the issue you raised about code duplication with the following commit: f54836830e25d5d8efa3a9845f74336c747b0a37.  Please let me know if you these changes are acceptable.

src/modules/python/Makefile.am
toonen: @bayucan Much thanks for the info and the patch.  After some additional testing, it appears that you are correct in that those flags and libraries are not required.  I'm not entirely sure why given that I obviously added them for a reason, but I'm guessing I introduced them during the dev process to eliminate some error, and the changes weren't ultimately necessary and were resolved by another change.  I've applied and committed the changes you proposed, test them locally, and they appear to have passed CI testing.  Again, thanks for both the feedback and the patch.

test/tests/pbs_smoketest.py
toonen: @bayucan: I applied the suggested patch in the following commit: 3a61204951b88280b82cc47f5d9ab9912ab31c6c.  Thanks for the suggestion.

src/lib/Libifl/Makefile.am
bayucan: @toonen : I agree with Hiren here too. WIthout the @swig_py_inc, pbs_ifl_wrap.c is not getting created. It errors out with "Error: Unable to find 'python.swg". I actually had to add it in order to make it work.

test/tests/pbs_smoketest.py
toonen: This had the side effect of identifying an issue with -fsanitize=address, which is currently resulting in the import tests failing when that flag is included.  I'm working on an autotools appropriate way to address the issue, which only reminds me that I'm a decade or so behind on frequent autotools use. :-)

test/tests/pbs_smoketest.py
```diff
@@ -1487,6 +1487,8 @@ def test_import_pbs_ifl_module(self):
         PBS_EXEC/{lib,lib64}/python/altair.
         """
         import pbs_ifl
+        sys.path.append(os.path.join(
+            self.server.pbs_conf['PBS_EXEC'], 'lib','python', 'altair'))
```
bayucan: The 'import pbs_ifl' needs to go after the ..sys.append() otherwise we'll get an import error.

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -1657,6 +1657,22 @@ def skipTest(self, reason=None):
...
+    def add_pbs_python_path_to_sys_path(self):
+        """
+        Add the path to the installed PBS Python modules located in the PBS
+        installation directory to the module search path if the path is not
+        already present.
+        """
+        for lib_dir in ['lib64', 'lib']:
+            pbs_python_path = os.path.join(self.server.pbs_conf['PBS_EXEC'],
+                lib_dir, 'python', 'altair')
+        if os.path.isdir(pbs_python_path):
```
bayucan, line 1669: Did you mean to indent this if statement to be under the for loop body?

toonen: Yes I did.  Thanks for catching that!  It's fixed in commit 5591dd0695e62133343861880db91042ebc47ffa.

toonen:
I believe I have addressed all of the issues that were pointed out.  If I missed anything, please let me know.

bayucan (approved):
Looks good now! This is a really nice, useful, addition to PBS. Thanks.

src/include/pbs_v1_module_common.i
```diff
@@ -0,0 +1,720 @@
+/*
```
hirenvadalia, line 1: I think file should `.h` not `.i` as its header file. Also I would suggest filename to be like `pbs_module_common.h`

m4/pbs_patch_libtool.m4
```diff
@@ -0,0 +1,55 @@
...
+# subject to Altair's trademark licensing policies.
+
+#
+
+AC_DEFUN([PBS_AC_PATCH_LIBTOOL], [
+	AC_CONFIG_COMMANDS([patch-libtool], [
+		AS_IF([! grep '[[-]]fsanitize=\*' libtool 2>&1 >/dev/null], [
+			AC_MSG_NOTICE([patching libtool to support -fsanitize])
+			mv libtool libtool.orig
+			sed -e 's/\(-fuse-linker-plugin\)\([[|)]]\)/\1|-fsanitize=\*\2/' \
```
subhasisb, line 47: Hi @toonen  this sed line does not seem to be adding -fsanitize on my copy of libtool. Can you check further?

m4/pbs_patch_libtool.m4
```diff
@@ -0,0 +1,55 @@
...
+				libtool.orig >libtool
+			rm -f libtool.orig
```
subhasisb, line 49: If the sed did something bad, we would be deleting the original libtool without checking the error status - perhaps we redirect sed output to libtool.upd (or similar name), check that the change we wanted is actually good and then mv libtool.upd to libtool when we are sure.

src/include/pbs_v1_module_common.i
subhasisb: @hirenvadalia this is not a header file really. In fact this contains defintions of functions..It is really the interface file (with stub implementations)

subhasisb (changes requested):

src/include/pbs_v1_module_common.i
toonen: @hirenvadalia: As @subhasisb stated, the file consists of global variable definitions and non-inline function definitions that really don't belong in a `.h` file so I used `.i` instead.  I included "v1" in the filename since it seemed specific to that version and to be consistent with `src/lib/Libpython/module_pbs_v1.c`.  I also included the version in `src/modules/python/pbs_v1_module_init.c`.  If you would still like the "v1" removed in either or both of those filenames, please let me know.

m4/pbs_patch_libtool.m4
toonen: Good point.  I made the modifications you suggested and also added code to generate an error file when something goes wrong.  In addition, I changed the option used by sed to one that I hope is more common.  The changes are in commit fa6afda76d973f41547eaa8ee6cef7e163173ea0.

m4/pbs_patch_libtool.m4
toonen: I'm hoping the changes in fa6afda76d973f41547eaa8ee6cef7e163173ea0 will resolve this; however, if it is still failing, please send me the contents of `libtool.patched.err` and the specific OS and libtool versions you are using.

src/include/pbs_v1_module_common.i
hirenvadalia: @toonen thanks for explaining... I am good with keeping v1 in filename, if needed we can rename anytime in future.

m4/pbs_patch_libtool.m4
subhasisb: Does not seem to work still, here is my container info:
[root@pbs pbspro]# libtool --version
libtool (GNU libtool) 2.4.2
Written by Gordon Matzigkeit <gord@gnu.ai.mit.edu>, 1996

Copyright (C) 2011 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
[root@pbs pbspro]# uname -a
Linux pbs 4.19.76-linuxkit #1 SMP Tue May 26 11:42:35 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
[root@pbs pbspro]# cat /etc/os-release 
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:7"
HOME_URL="https://www.centos.org/"
BUG_REPORT_URL="https://bugs.centos.org/"

CENTOS_MANTISBT_PROJECT="CentOS-7"
CENTOS_MANTISBT_PROJECT_VERSION="7"
REDHAT_SUPPORT_PRODUCT="centos"
REDHAT_SUPPORT_PRODUCT_VERSION="7"

I ran this command from your last commit:
sed 's/\(-pg\)\([[|)]]\)/\1|-fsanitize=\*\2/' libtool >libtool.patched 2>libtool.patched.err

and found a empty (0 bytes) libtool.patched.err file. The libtool.patched file is exactly same as libtool. This means sed passed through this file without making any changes.

subhasisb (approved):

hirenvadalia (approved):

PR #1997, opened 2020-08-25
Introduce $jobdir_root 'shared' directive

bayucan:
#### Describe Bug or Feature

* If a job has been submitted with sandbox attribute set to PRIVATE, (i.e. qsub -W sandbox=PRIVATE), a special job’s staging and execution directory is created under the directory specified in MoM $jobdir_root configuration option.
* Putting jobdir_root to a shared location for multiple MoMs breaks when using the [Node Ramp down Feature](https://openpbs.atlassian.net/wiki/spaces/PD/pages/49897561/PP-339+and+PP-647+node+ramp+down+feature+release+vnodes+early+from+running+jobs). 
* When a node is released early from a job by calling 'pbs_release_nodes', any job-specific specific files including job temporary directories sitting on the $jobdir_root directory, are removed by the sister mom managing that node.
* Unfortunately, the job-specific files are sitting on the same shared location seen by primary mom and the other sister moms, causing user's files to disappear. Primary mom would lose the ability to  stage out files and return stdout/stderr files, and executing job could actually abort.

#### Describe Your Change

* Introduce a new, optional directive 'shared'  to the $jobdir_root mom configuration option, that tells mom that the specified path is a shared location among primary mom and sister moms.
$jobdir_root <stage directory root>[shared]
* Having this option set, the sister mom would not cleanup the job's directories, when a request to delete the job on the node managed by that sister mom is received. 
* At the end of the job, the primary mom would take care of cleaning up the files.
#### Link to Design Doc
 [Option for sister moms to not delete job's files sitting on shared location](https://openpbs.atlassian.net/wiki/spaces/PD/pages/1979482121/Option+for+sister+moms+to+not+delete+job+s+files+sitting+on+shared+location)

* [ptl.sandbox_private_jobdir_shared.FAIL.txt](https://github.com/openpbs/openpbs/files/5126354/ptl.sandbox_private_jobdir_shared.FAIL.txt)
* [ptl.sandbox_private_jobdir_shared.PASS.txt](https://github.com/openpbs/openpbs/files/5126357/ptl.sandbox_private_jobdir_shared.PASS.txt)
* [valgrind.qsub_sandbox_private_jobdir_shared.txt](https://github.com/openpbs/openpbs/files/5126358/valgrind.qsub_sandbox_private_jobdir_shared.txt)
* [valgrind.qsub_sandbox_private_jobdir_shared.sismom.txt](https://github.com/openpbs/openpbs/files/5126359/valgrind.qsub_sandbox_private_jobdir_shared.sismom.txt)

* [ptl.sandbox_private_jobdir_shared.PASS3.txt](https://github.com/openpbs/openpbs/files/5159372/ptl.sandbox_private_jobdir_shared.PASS3.txt)
* [valgrind.msmom.jobdir_root_shared.txt](https://github.com/openpbs/openpbs/files/5159376/valgrind.msmom.jobdir_root_shared.txt)
* [valgrind.sismom.jobdir_root_shared.txt](https://github.com/openpbs/openpbs/files/5159377/valgrind.sismom.jobdir_root_shared.txt)

### DIFF
--- a/src/resmom/mom_main.c
+++ b/src/resmom/mom_main.c
@@ -234,6 +234,7 @@ char            pbs_jobdir_root[MAX_PATH]= "";
 char            pbs_tmpdir[_POSIX_PATH_MAX] = TMP_DIR;
 char            pbs_jobdir_root[_POSIX_PATH_MAX]= "";
 #endif
+int		pbs_jobdir_root_shared = FALSE;
 vnl_t		*vnlp = NULL;			/* vnode list */
 unsigned long	hooks_rescdef_checksum = 0;
 
@@ -3614,6 +3615,15 @@ static handler_ret_t
 set_jobdir_root(char *value)
 {
 	char	*cleaned_value;
+	char    *endptr;
+
+	endptr = value;
+	while ((!isspace((int)*endptr)) && *endptr)
+		endptr++;
+	if (*endptr != '\0') {
+		*endptr = '\0';
+		endptr++;
+	}
 
 	log_event(PBSEVENT_SYSTEM, PBS_EVENTCLASS_SERVER,
 		LOG_INFO, __func__, value);
@@ -3644,6 +3654,11 @@ set_jobdir_root(char *value)
 
 	strcpy(pbs_jobdir_root, cleaned_value);
 	free(cleaned_value);
+
+	if (*endptr != '\0') {
+		if (strstr(endptr, "shared"))
+			pbs_jobdir_root_shared = TRUE;
+	}
 	return HANDLER_SUCCESS;
 }
 
--- a/src/resmom/mom_main.c
+++ b/src/resmom/mom_main.c
@@ -3656,7 +3656,7 @@ set_jobdir_root(char *value)
 	free(cleaned_value);
 
 	if (*endptr != '\0') {
-		if (strstr(endptr, "shared"))
+		if (strcmp(endptr, "shared") == 0)
 			pbs_jobdir_root_shared = TRUE;
 	}
 	return HANDLER_SUCCESS;
--- a/src/resmom/mom_main.c
+++ b/src/resmom/mom_main.c
@@ -3615,16 +3615,13 @@ static handler_ret_t
 set_jobdir_root(char *value)
 {
 	char	*cleaned_value;
-	char    *endptr;
+	char    *savep = NULL;
+	char	*directive;
+	char	*p;
 
-	endptr = value;
-	while ((!isspace((int)*endptr)) && *endptr)
-		endptr++;
-	if (*endptr != '\0') {
-		*endptr = '\0';
-		endptr++;
-	}
-
+	p = value;
+	value = strtok_r(p, " ", &savep);
+	directive = strtok_r(NULL, " ", &savep);
 	log_event(PBSEVENT_SYSTEM, PBS_EVENTCLASS_SERVER,
 		LOG_INFO, __func__, value);
 	cleaned_value = remove_quotes(value); /* remove quotes if any present */
@@ -3655,8 +3652,8 @@ set_jobdir_root(char *value)
 	strcpy(pbs_jobdir_root, cleaned_value);
 	free(cleaned_value);
 
-	if (*endptr != '\0') {
-		if (strcmp(endptr, "shared") == 0)
+	if (directive != NULL) {
+		if (strcmp(directive, "shared") == 0)
 			pbs_jobdir_root_shared = TRUE;
 	}
 	return HANDLER_SUCCESS;
--- a/src/resmom/stage_func.c
+++ b/src/resmom/stage_func.c
@@ -1184,6 +1184,9 @@ error:
  * @param[in] gid    - the group id of the user under which the job will run.
  *		       Currenty this parameter only used in *nix so for
  *		       windows this parameter should be NULL
+ * @param[in] check_shared - if set to '1', test if the staging and execution
+ *			directory is sitting on a shared location, and if so,
+ *			do not remove any files in it.
  * @return void
  *
  * @note	This may take awhile so the task is forked and execed to another
@@ -1192,7 +1195,7 @@ error:
  *
  */
 void
-rmjobdir(char *jobid, char *jobdir, uid_t uid, gid_t gid)
+rmjobdir(char *jobid, char *jobdir, uid_t uid, gid_t gid, int check_shared)
 {
 	static	char	rmdir_buf[MAXPATHLEN+1] = {'\0'};
 	struct	stat	sb = {0};
@@ -1211,6 +1214,15 @@ rmjobdir(char *jobid, char *jobdir, uid_t uid, gid_t gid)
 
 	if (jobdir == NULL)
 		return;
+
+	if (check_shared &&
+	    (pbs_jobdir_root[0] != '\0') &&
+	    pbs_jobdir_root_shared &&
+	    (strncmp(pbs_jobdir_root, jobdir, strlen(pbs_jobdir_root)) == 0)) {
+		log_eventf(PBSEVENT_DEBUG3, PBS_EVENTCLASS_JOB, LOG_DEBUG, jobid?jobid:__func__, "shared jobdir %s to be removed by primary mom", jobdir);
+
+		return;
+	}
 
 #ifndef WIN32
 	if (pbs_jobdir_root[0] == '\0') {
--- a/src/resmom/stage_func.c
+++ b/src/resmom/stage_func.c
@@ -1219,7 +1219,7 @@ rmjobdir(char *jobid, char *jobdir, uid_t uid, gid_t gid, int check_shared)
 	    (pbs_jobdir_root[0] != '\0') &&
 	    pbs_jobdir_root_shared &&
 	    (strncmp(pbs_jobdir_root, jobdir, strlen(pbs_jobdir_root)) == 0)) {
-		log_eventf(PBSEVENT_DEBUG3, PBS_EVENTCLASS_JOB, LOG_DEBUG, jobid?jobid:__func__, "shared jobdir %s to be removed by primary mom", jobdir);
+		log_eventf(PBSEVENT_DEBUG3, PBS_EVENTCLASS_JOB, LOG_DEBUG, jobid?jobid:"", "shared jobdir %s to be removed by primary mom", jobdir);
 
 		return;
 	}
--- a/test/tests/functional/pbs_qsub_opts_args.py
+++ b/test/tests/functional/pbs_qsub_opts_args.py
@@ -61,17 +61,16 @@ class TestQsubOptionsArguments(TestFunctional):
         self.remove_jobdir = False
 
     def tearDown(self):
+        TestFunctional.tearDown(self)
         if self.remove_jobdir:
             for mom in self.moms.values():
                 for d in self.jobdir:
                     if d.startswith(self.jobdir_root):
                         self.logger.info('%s:remove jobdir %s' % (
                                          mom.hostname, d))
-                        cmd = ['/bin/rm', '-rf', d]
-                        self.du.run_cmd(mom.hostname, cmd=cmd,
-                                        sudo=True)
+                        self.du.rm(hostname=mom.shortname, sudo=True,
+                                   path=d, recursive=True, force=True)
             self.remove_jobdir = False
-        TestFunctional.tearDown(self)
 
     def validate_error(self, err):
         ret_msg = 'qsub: Failed to save job/resv, '\
@@ -204,12 +203,11 @@ bhtiusabsdlg' % (os.environ['HOME'])
         momA = self.moms.values()[0]
         momB = self.moms.values()[1]
 
+        loglevel = {'$logevent': 4095}
+        momB.add_config(loglevel)
         c = {'$jobdir_root': '%s shared' % self.jobdir_root}
-        l = {'$logevent': 4095}
         for mom in [momA, momB]:
             mom.add_config(c)
-            if mom == momB:
-                mom.add_config(l)
             mom.restart()
 
         a = {'Resource_List.select': '2:ncpus=1',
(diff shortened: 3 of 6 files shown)


test/tests/functional/pbs_qsub_opts_args.py
```diff
@@ -56,6 +56,17 @@ def setUp(self):
         self.fn = self.du.create_temp_file(body=script)
         self.qsub_cmd = os.path.join(
             self.server.pbs_conf['PBS_EXEC'], 'bin', 'qsub')
+        self.jobdir_root = "/tmp"
+        self.jobdir = []
+
+    def tearDown(self):
+        for mom in self.moms.values():
```
lisa-altair, line 63: Only your new test case needs this.  But putting it here means all 9 other tests will have to run through the setup and teardown.  Is it better to put this here or within the new test?

src/resmom/mom_main.c
```diff
@@ -3697,6 +3707,11 @@ set_jobdir_root(char *value)
 
 	strcpy(pbs_jobdir_root, cleaned_value);
 	free(cleaned_value);
+
+	if (*endptr != '\0') {
+		if (strstr(endptr, "shared"))
```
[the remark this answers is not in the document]
bayucan: Good point. I'll update the fix.

test/tests/functional/pbs_qsub_opts_args.py
bayucan: I have to put it there as the test case could actually throw an exception on a self.assert() marking the test case failure, and this would be the only place where things can be cleaned up.

lisa-altair: Thanks for reminding me about the exception case.  Makes sense to me.

bhroam: What I usually do for things like this is to create a class variable which is False by default.  You set it to True at the start of the test.  If the variable is True, then you do something special in tearDown() and set it back to False.

bayucan: Good idea @bhroam : Let me do that.

bayucan:
@bhroam : can you review the entire PR? It should be straightforward with design. Thanks.

src/resmom/mom_main.c
```diff
@@ -3667,6 +3668,15 @@ static handler_ret_t
...
 	char	*cleaned_value;
+	char    *endptr;
+
+	endptr = value;
+	while ((!isspace((int)*endptr)) && *endptr)
+		endptr++;
+	if (*endptr != '\0') {
+		*endptr = '\0';
+		endptr++;
+	}
```
bhroam, line 3679: What about using strchr(endptr, ' ') or strpbrk()?

src/resmom/stage_func.c
```diff
@@ -1210,6 +1213,15 @@ rmjobdir(char *jobid, char *jobdir, uid_t uid, gid_t gid)
 	if (jobdir == NULL)
 		return;
 
+	if (check_shared &&
+	    (pbs_jobdir_root[0] != '\0') &&
+	    pbs_jobdir_root_shared &&
+	    (strncmp(pbs_jobdir_root, jobdir, strlen(pbs_jobdir_root)) == 0)) {
+		log_eventf(PBSEVENT_DEBUG3, PBS_EVENTCLASS_JOB, LOG_DEBUG, jobid?jobid:__func__, "shared jobdir %s to be removed by primary mom", jobdir);
+
```
bhroam, line 1221: I'd say just jobid, not jobid with the function.  That way you'll see this message with tracejob.

test/tests/functional/pbs_qsub_opts_args.py
```diff
@@ -56,6 +56,22 @@ def setUp(self):
...
+    def tearDown(self):
+        if self.remove_jobdir:
+            for mom in self.moms.values():
+                for d in self.jobdir:
+                    if d.startswith(self.jobdir_root):
+                        self.logger.info('%s:remove jobdir %s' % (
+                                         mom.hostname, d))
+                        cmd = ['/bin/rm', '-rf', d]
+                        self.du.run_cmd(mom.hostname, cmd=cmd,
+                                        sudo=True)
```
bhroam, line 72: Use self.du.rm()

test/tests/functional/pbs_qsub_opts_args.py
```diff
@@ -56,6 +56,22 @@ def setUp(self):
...
+            self.remove_jobdir = False
+        TestFunctional.tearDown(self)
```
bhroam, line 74: Will removing the directories cause any problems when the jobs are removed?  If so, remove them after TestFunctional.tearDown()

test/tests/functional/pbs_qsub_opts_args.py
```diff
@@ -177,3 +193,55 @@ def test_qsub_with_option_a(self):
...
+        self.remove_jobdir = True
+        momA = self.moms.values()[0]
+        momB = self.moms.values()[1]
+
+        c = {'$jobdir_root': '%s shared' % self.jobdir_root}
+        l = {'$logevent': 4095}
+        for mom in [momA, momB]:
+            mom.add_config(c)
+            if mom == momB:
+                mom.add_config(l)
```
bhroam, line 212: Move this outside the for and just say momB.add_config()

test/tests/functional/pbs_qsub_opts_args.py
```diff
@@ -177,3 +193,55 @@ def test_qsub_with_option_a(self):
...
+             }
+        j = Job(TEST_USER, attrs=a)
+        j.set_sleep_time(10)
+        jid = self.server.submit(j)
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid)
+        attribs = self.server.status(JOB, id=jid)
+        jobdir = attribs[0]['jobdir']
+        self.jobdir.append(jobdir)
+        relpath = os.path.join(
+            self.server.pbs_conf['PBS_EXEC'], 'bin', 'pbs_release_nodes')
```
bhroam, line 227: Isn't there a PTL wrapper for release nodes?

test/tests/functional/pbs_qsub_opts_args.py
```diff
@@ -177,3 +193,55 @@ def test_qsub_with_option_a(self):
...
+
+        rel_cmd = [relpath, '-j', jid, momB.shortname]
+        ret = self.server.du.run_cmd(self.server.hostname, cmd=rel_cmd,
+                                     runas=TEST_USER)
+        self.assertEqual(ret['rc'], 0)
+        file_exists_cmd = '"import os; print(os.path.exists(\'%s\'))"'
+        pbs_python = os.path.join(self.server.pbs_conf['PBS_EXEC'],
+                                  "bin", "pbs_python")
+        file_exists = [pbs_python, "-c", file_exists_cmd % jobdir]
```
bhroam, line 236: use self.du.isdir()

src/resmom/mom_main.c
bayucan: Not in this case, since I need to terminate the current string on the first non-space character found. For example, given '$jobdir /tmp share', endptr at this time will be pointing to the space (' ') after '/tmp', and I need that terminated with \0. '/tmp' is then passed on as the jobdir_root value.
I do see that if there are multiple spaces after '/tmp' and before 'shared', the current code does not move past the space to get to 'shared', so I'll fix that.

src/resmom/stage_func.c
bayucan: The function where this is called tolerates a NULL jobid. I don't think that will really happen but just to guard against crash, I'm handling such a case. Instead of '__func__', I'll replace with just the string "jobid".

src/resmom/mom_main.c
bhroam: I still don't see why the functions won't work.  They search for the space, and you want to replace the space with a null byte.  To handle the multiple space case, maybe strtok_r() is the right answer.  That handles the case when the token separator appears multiple times between tokens.

src/resmom/stage_func.c
bhroam: use "", not "jobid"

bayucan: Ok, will use "".

src/resmom/mom_main.c
bayucan: Ok, I'll re-implement like this:
```
--- mom_main.c  2020-09-01 13:06:16.324785944 -0700
***************
*** 3668,3681 ****
  set_jobdir_root(char *value)
  {
        char    *cleaned_value;
        char    *endptr;
  
!       endptr = value;
!       while ((!isspace((int)*endptr)) && *endptr)
!               endptr++;
!       if (*endptr != '\0') {
                *endptr = '\0';
                endptr++;
        }
  
        log_event(PBSEVENT_SYSTEM, PBS_EVENTCLASS_SERVER,
--- 3668,3682 ----
  set_jobdir_root(char *value)
  {
        char    *cleaned_value;
+       char    *p;
        char    *endptr;
  
!       p = value;
!       if ((endptr = strchr(p, ' ')) != NULL) {
                *endptr = '\0';
                endptr++;
+       } else {
+               endptr = p;
        }
  
        log_event(PBSEVENT_SYSTEM, PBS_EVENTCLASS_SERVER,
***************
*** 3708,3713 ****
--- 3709,3716 ----
        strcpy(pbs_jobdir_root, cleaned_value);
        free(cleaned_value);
  
+       while (isspace((int)*endptr))
+               endptr++;
        if (*endptr != '\0') {
                if (strcmp(endptr, "shared") == 0)
                        pbs_jobdir_root_shared = TRUE;
```
-----
After the function processed the jobdir_root value, I'll move past any spaces to look for the optional 'shared' directive.

bhroam: That will work, but you could replace it all with the following two lines:
A = strtok_r(p, " ", &savep);
B = strtok_r(NULL, " ", &savep);
A will always return p regardless if there are any spaces.  B will be the first character after any number of spaces that separate the end of A and the start of B.  If there are no spaces, B will be NULL.

test/tests/functional/pbs_qsub_opts_args.py
bayucan: Will use.

test/tests/functional/pbs_qsub_opts_args.py
bayucan: Yes, TestFunctional.tearDown() can be called first.

test/tests/functional/pbs_qsub_opts_args.py
bayucan: If I do that, then it would look like this:
```
c = {'$jobdir_root': '%s shared' % self.jobdir_root}
for mom in [momA, momB]:
    mom.add_config(c)
    mom.restart()
l = {'$logevent': 4095}
momB.add_config(l)
momB.restart()
```
Or would you prefer this:
```
 c = {'$jobdir_root': '%s shared' % self.jobdir_root}
momA.add_config(c)
momA.restart()
l = {'$logevent': 4095}
momB.add_config(c)
momB.add_config(l)
momB.restart()
```

test/tests/functional/pbs_qsub_opts_args.py
bayucan: None right now.

test/tests/functional/pbs_qsub_opts_args.py
bayucan: Will use. It will simplify things.

src/resmom/mom_main.c
bayucan: Ok, will do this.

test/tests/functional/pbs_qsub_opts_args.py
bhroam: The first restarts momB twice.  This is unnecessary.  If you add 'l' first, you can then do your for loop and just restart the mom there.  It is going to be the shortest.  The second is also fine.  Whichever you want.  Just don't restart momB twice.  As a note, pycodestyle will complain at you for the use of a variable called 'l'.  It considers it confusing.  You might consider renaming it.

bayucan: I'll rename the variable.

test/tests/functional/pbs_qsub_opts_args.py
```diff
@@ -230,18 +228,16 @@ def test_qsub_sandbox_private_jobdir_shared(self):
...
-        pbs_python = os.path.join(self.server.pbs_conf['PBS_EXEC'],
-                                  "bin", "pbs_python")
-        file_exists = [pbs_python, "-c", file_exists_cmd % jobdir]
-        ret = self.du.run_cmd(momB.hostname, cmd=file_exists, sudo=True)
         # sister mom has preserved the file
         errmsg = "sister mom deleted jobdir %s" % jobdir
-        self.assertEqual(''.join(ret['out']), 'True', errmsg)
+        rc = self.du.isdir(hostname=momB.shortname, path=jobdir,
+                           sudo=True)
+        self.assertEqual(rc, True, errmsg)
```
bhroam, line 235: Since you are comparing this to True, consider using assertTrue()

bayucan: Yes, I think it's a good idea to use assertTrue() (and assertFalse()). Ok, I pushed a new commit.

test/tests/functional/pbs_qsub_opts_args.py
```diff
@@ -177,3 +192,52 @@ def test_qsub_with_option_a(self):
...
+        # sister mom has preserved the file
+        errmsg = "sister mom deleted jobdir %s" % jobdir
+        rc = self.du.isdir(hostname=momB.shortname, path=jobdir,
+                           sudo=True)
+        self.assertTrue(rc, errmsg)
+        msg = "shared jobdir %s to be removed by primary mom" % jobdir
+        momB.log_match(msg)
+        self.server.expect(JOB, 'job_state', op=UNSET, id=jid)
+        # primary mom has deleted the file
+        errmsg = "MS mom preserved jobdir %s" % jobdir
```
lisa-altair, line 240: nit: I'm sorry to be mentioning this only now...this doesn't seem like an error to me.
Shouldn't it say what went wrong?  Something like, "MS mom did not delete jobber %s"?

lisa-altair: Okay.  I see what you are saying.  I am okay with the current code.

PR #1998, opened 2020-08-26
Job array optimization in run path

bremanandjk:
#### Describe Bug or Feature

Identified a possibility to optimize the job array run path while running a large number of sub-jobs. Modified the code in a way that would save Server's execution time.

#### Describe Your Change

Optimized the computation part for identifying the remaining sub-job indices. For every job stat, the server would walk through all the submitted sub-jobs to find the queued jobs remaining in the array job. In this PR, the new range structures are introduced to maintain the remaining indices for quick reference. This change would increase overall performance in the job array. 

Submitted 1 million jobs in the server and measured the job run path utilization. 

Before change:
Before_job_array_run_path

After change:
After_job_array_runpath

Valgrind logs: 
[valgrind.log](https://github.com/openpbs/openpbs/files/5127428/valgrind.log)

### DIFF
--- a/src/include/job.h
+++ b/src/include/job.h
@@ -325,6 +325,18 @@ typedef struct	noderes {
  * specific structures for Job Array attributes
  */
 
+typedef struct {
+	int start;
+	int end;
+} range_t;
+
+typedef struct {
+	int count;
+	int step;
+	range_t *ranges;
+} range_arr_t;
+
+
 /* individual entries in array job index table */
 struct ajtrk {
 	int trk_status;		 /* status */
@@ -346,6 +358,7 @@ struct ajtrkhd {
 	int tkm_flags;			  /* special flags for array job */
 	int tkm_subjsct[PBS_NUMJOBSTATE]; /* count of subjobs in various states */
 	int tkm_dsubjsct;		  /* count of deleted subjobs */
+	range_arr_t tkm_rarr;		 	/* array of ranges */
 	struct ajtrk tkm_tbl[1];	  /* ptr to array of individual entries */
 	/*
 	 * when table is malloced, room for the additional required number
--- a/src/server/array_func.c
+++ b/src/server/array_func.c
@@ -137,6 +137,257 @@ static enum job_atr attrs_to_copy[] = {
 	JOB_ATR_LAST /* This MUST be LAST	*/
 };
 
+
+/**
+ * @brief
+ * 			init_range - Initialize the sub-job indices
+ *
+ * @param[in]	rarr - pointer to the range_arr_t 
+ * @param[in]	start - start of the subjob indices
+ * @param[in]	end - end of the subjob indices
+ * @param[in]	step - step factor in subjob indices
+ *
+ * @return      Int
+ * @retval	 0  - Intialization success
+ * @retval	 1  - failure 
+ */
+
+int init_range(range_arr_t *rarr, int start, int end, int step)
+{
+	rarr->count = 1;
+	rarr->step= step;
+	rarr->ranges = malloc(sizeof(range_t));
+
+	rarr->ranges[0].start = start;
+	rarr->ranges[0].end = end;
+	return 0;
+}
+
+/**
+ * @brief
+ * 			add_value - Adding a value in the range structures
+ * 						handles adding value in mid of the ranges.
+ *
+ * @param[in]	rarr - pointer to the range_arr_t 
+ * @param[in]	val - Value to add
+ *
+ * @return      Int
+ * @retval	 0  - Intialization success
+ * @retval	 1  - failure 
+ */
+
+int
+add_value(range_arr_t *rarr, int val)
+{
+	int i = 0,j = 0;
+	range_t *new;
+	
+	if (rarr->ranges == NULL) {
+		init_range(rarr, val, val, rarr->step);
+		return 0;
+	}
+
+	/*check the extreme left corner case */
+	if ( val < rarr->ranges[i].start) {
+		rarr->count++;
+		new = realloc(rarr->ranges, sizeof(range_t) * rarr->count);
+		if (new)
+			rarr->ranges = new;
+		else
+			return 1;
+		for (j = rarr->count - 1; j > i; j--) { /* move the pointers right */
+			rarr->ranges[j] = rarr->ranges[j - 1];
+		}
+		/* now we have space for the index in between */
+		rarr->ranges[i].start = val;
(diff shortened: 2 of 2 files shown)


src/include/job.h
```diff
@@ -325,6 +325,18 @@ typedef struct	noderes {
...
  */
 
+typedef struct {
+	int start;
+	int end;
+} range_t;
+
+typedef struct {
+	int count;
+	int step;
```
hirenvadalia, line 335: why add step here? we can use `tkm_step` instead

src/server/array_func.c
```diff
@@ -137,6 +137,257 @@ static enum job_atr attrs_to_copy[] = {
...
+
+	/*check the extreme left corner case */
+	if ( val < rarr->ranges[i].start) {
+		rarr->count++;
+		new = realloc(rarr->ranges, sizeof(range_t) * rarr->count);
+		if (new)
+			rarr->ranges = new;
+		else
+			return 1;
+		for (j = rarr->count - 1; j > i; j--) { /* move the pointers right */
```
hirenvadalia, line 198: however this loop and above realloc will be less in number but using sched's range struct we can get rid of this loop and above realloc

src/server/array_func.c
```diff
@@ -137,6 +137,257 @@ static enum job_atr attrs_to_copy[] = {
...
+			/* value falls in this range, split range */
+				if ((val - rarr->ranges[i - 1].end) % rarr->step != 0)
+					return 1;
+				rarr->count++;
+				new = realloc(rarr->ranges, sizeof(range_t) * rarr->count);
+				if (new)
+					rarr->ranges = new;
+				else 
+					return 1;
+				for (j = rarr->count - 1; j > i; j--) { /* move the pointers right (assiging j-1th = jth)  */
```
hirenvadalia, line 243: same as above, we can get rid of this

src/server/array_func.c
```diff
@@ -137,6 +137,257 @@ static enum job_atr attrs_to_copy[] = {
...
+ * @param[in]	rarr - pointer to the range_arr_t 
+ * @param[in]	val - Value to delete
+ *
+ * @return      Int
+ * @retval	 0  - Intialization success
+ * @retval	 1  - failure 
+ */
+
+int
+del_value(range_arr_t *rarr, int val)
```
hirenvadalia, line 293: we can merge this with range_del_value() and use it in both server and sched. for sched, sched never calls range_del_value to delete any value in middle of range, as sched runs lowest index always (@bhroam please correct me if I am wrong)

src/server/array_func.c
```diff
@@ -137,6 +137,257 @@ static enum job_atr attrs_to_copy[] = {
...
+ * @param[in]	rarr - pointer to the range_arr_t 
+ * @param[in]	val - Value to delete
+ *
+ * @return      Int
+ * @retval	 0  - Intialization success
+ * @retval	 1  - failure 
+ */
+
+char *
+encode_range(range_arr_t *rarr)
```
hirenvadalia, line 348: use `range_as_str` instead introducing new func for same functionality

src/server/array_func.c
```diff
@@ -716,6 +980,19 @@ setup_arrayjob_attrs(attribute *pattr, void *pobj, int mode)
 		if ((pjob->ji_ajtrk = mk_subjob_index_tbl(pjob->ji_wattr[(int)JOB_ATR_array_indices_submitted].at_val.at_str,
 			                                      JOB_STATE_QUEUED, &pbs_error, mode)) == NULL)
 			return pbs_error;
+		
+		prange = pjob->ji_wattr[(int)JOB_ATR_array_indices_submitted].at_val.at_str;
+		if (!prange)
```
hirenvadalia, line 985: there is no need to check for not NULL, it is NULL will never reach this point as it will crash server in mk_subjob_index_tbl()->parse_subjob_index()->`while (isspace((int) *pc)` + setup_arrayjob_attrs is action func of ATTR_array_indices_submitted, means it will be called only when value is set from req_queuejob()

src/server/array_func.c
```diff
@@ -716,6 +980,19 @@ setup_arrayjob_attrs(attribute *pattr, void *pobj, int mode)
...
+			return PBSE_BADATVAL; 
+			
+		i = parse_subjob_index(prange, &eptr, &start, &end, &step, &count);
+		
+		if (i != 0) 
+			return PBSE_BADATVAL; /* parse error */
+
+		if(init_range(&pjob->ji_ajtrk->tkm_rarr, start, end, step))
```
hirenvadalia, line 993: use `range_parse` here

src/server/array_func.c
```diff
@@ -741,9 +1018,9 @@ setup_arrayjob_attrs(attribute *pattr, void *pobj, int mode)
 	}
 
 	/* set "array_indices_remaining" if not already set */
-	if ((pjob->ji_wattr[(int)JOB_ATR_array_indices_remaining].at_flags & ATR_VFLAG_SET) == 0)
+	if ((pjob->ji_wattr[(int)JOB_ATR_array_indices_remaining].at_flags & ATR_VFLAG_SET) == 0) {
```
hirenvadalia, line 1021: plz remove `{}` for single line if statement

src/server/array_func.c
```diff
@@ -382,7 +639,7 @@ update_array_indices_remaining_attr(job *parent)
 
 	if (ptbl->tkm_flags & TKMFLG_REVAL_IND_REMAINING) {
 		attribute *premain = &parent->ji_wattr[(int)JOB_ATR_array_indices_remaining];
-		char *pnewstr = cvt_range(parent, JOB_STATE_QUEUED);
+		char *pnewstr = encode_range(&ptbl->tkm_rarr);
```
hirenvadalia, line 642: [adding comment on this line as somehow github is not allowing to add comment on exact line in chk_array_doneness()]

@bremanandjk I don't see any changes related to no to save JOB_ATR_array_indices_remaining in db and removal of job_save_db in chk_array_doneness()...

hirenvadalia (changes requested):

subhasisb:
Hi @bremanandjk i agree with @hirenvadalia  that we should reuse the range.c from scheduler. In fact, we should pull out the range.c from scheduler into libutils and add the required functionality to it. That way we do not have 2 sets of functions. Also the memory required by pointers is not much of a concern since i think in most cases, the list will only have a few nodes in it.

bremanandjk:
Raised another PR to implement the requested approach, closing this one.

PR #2000, opened 2020-08-26
Make scheduler to connect server and keep that connection persistent 

hirenvadalia:
#### Describe Bug or Feature

This change proposes the follows:
* Change the direction of connection initiation: Currently the server connects to the scheduler. However, this means the scheduler has to listen on a port and accept connections. On the other hand the server already does that very well, so we can simply reverse the direction of connection such that scheduler acts as a client (just like moms or IFL clients) and connects to the server after proper authentication.
* Make the scheduler connections persistent. There is no need to disconnect the connection between the server and scheduler just to indicate the end of a cycle, for which a message can be sent. That is less expensive than setting up and tearing down TCP connections. Besides, keeping a connection allows the scheduler to remain connected to all the servers irrespective of which server wanted to start the scheduling.

#### Describe Your Change

* 3 major changes as described in design:
  * Now scheduler will initiate connection with server
  * Now connection between server and sched will be persistent, and sched will not close it to notify server for end cycle, instead it will send, end cycle notification to server on secondary connection
  * Segregation of connection between server-sched: primary connection will be used only for IFL calls, and secondary will be used only for sending commads to sched (by server) and sending end cycle notification to server (by sched) 

#### Link to Design Doc

Forum discussion: [here](https://community.openpbs.org/t/make-scheduler-to-connect-server-and-keep-that-connection-persistent/2251)
Design page: [here](https://openpbs.atlassian.net/wiki/spaces/PD/pages/2047344647/WIP+Make+scheduler+to+connect+server+and+keep+that+connection+persistent)

* see in last few comments


hirenvadalia:
Hey reviewers, This PR is ready for review, please review and provide your feedback.

src/scheduler/pbs_sched.c
```diff
@@ -791,46 +502,285 @@ are_we_primary()
...
+
+	while (!hascmd) {
+		sigemptyset(&emptyset);
+		nfds = tpp_em_pwait(poll_context, &events, timeout, &emptyset);
+		err = errno;
+
+		if (nfds < 0) {
+			if (!(err == EINTR || err == EAGAIN || err == 0)) {
+				log_errf(err, __func__, " tpp_em_wait() error, errno=%d", err);
+				sleep(1); /* wait for 1s for not to burn too much CPU */
```
suresh-thelkar, line 707: In case if server is down even 1 second sleep also makes the disk full. One thought is we can make this configurable.

suresh-thelkar (changes requested):
I started reviewing the changes. I will continue to review. I will add comments as and when I find something.

src/scheduler/pbs_sched.c
```diff
@@ -791,46 +502,285 @@ are_we_primary()
...
+		if (sfd < 0) {
+			sleep(2);
+			continue;
+		}
+	}
+	tpp_em_add_fd(poll_context, sfd, EM_IN | EM_HUP | EM_ERR);
+
+	svr = malloc(sizeof(svr_t));
+	if (!svr) {
+		log_errf(-1, __func__, "Out of memory while allocating svr struct");
```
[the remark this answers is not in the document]
hirenvadalia: I understand that current code of sched is using MEM_ERR_MSG but its very generic message which is not quite helpful, saying explicit for what mem alloc failed can help more in debugging.
In fact I would suggest other way that change all MEM_ERR_MSG error in sched and put explicit message rather than very generic message.

src/scheduler/pbs_sched.c
```diff
@@ -791,46 +502,285 @@ are_we_primary()
...
+{
+	int pfd = -1;
+	int sfd = -1;
+	svr_t *svr;
+
+	while (pfd < 0 || sfd < 0) {
+		if (pfd < 0) {
+			pfd = connect_server_helper(svrhost, 1);
+			if (pfd < 0) {
+				sleep(2);
```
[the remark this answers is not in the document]
hirenvadalia: this sleep is not for timeout, its for not to burn too much cpu... I will add comment explaining it.

src/scheduler/pbs_sched.c
```diff
@@ -791,46 +502,285 @@ are_we_primary()
...
+		log_errf(-1, __func__, "Out of memory while allocating svr struct");
+		exit(1);
+	}
+
+	svr->svrhost = svrhost;
+	svr->pfd = pfd;
+	svr->sfd = sfd;
+	CLEAR_LINK(svr->all_svrs_link);
+	append_link(&servers, &svr->all_svrs_link, svr);
+	set_validate_sched_attrs(svr->pfd);
```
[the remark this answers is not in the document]
hirenvadalia: without this, there few tests fails related to fairshare in multisched tests, its not able to find correct log message, but if I put this call here then test gets its message properly

src/scheduler/pbs_sched.c
hirenvadalia: this is not related to server is down, if server is down then sched will be in connect_server_helper and it will be in pbs_connect() + sleep(2) loop, where we don't have any message.

src/scheduler/data_types.h
```diff
@@ -176,6 +177,19 @@ typedef sch_resource_t usage_t;
 typedef void event_ptr_t;
 typedef int (*event_func_t)(event_ptr_t*, void *);
 
+typedef struct sched_cmd {
+	int cmd;   /* sched command */
+	char *jid; /* jobid assisiated with cmd if any else NULL */
+} sched_cmd_t;
+
+typedef struct svr {
+	char *svrhost;
```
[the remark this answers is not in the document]
hirenvadalia: We don't need port here as pbs_connect/pbs_connect_extend, needs `host:port` or `host` (for dflt port) or `NULL` (for current host with dflt port) syntax.

src/scheduler/data_types.h
```diff
@@ -176,6 +177,19 @@ typedef sch_resource_t usage_t;
...
+	int pfd;		     /* primary connection to server - used for IFL calls */
+	int sfd;		     /* secondary connection to server - used to receive sched cmds from server */
+	sched_cmd_t cmd;	     /* command received from server */
+	pbs_list_link all_svrs_link; /* link to all servers */
```
suresh-thelkar, line 190: We can just name it as next_svr_link.

src/scheduler/fifo.c
```diff
@@ -798,6 +745,25 @@ scheduling_cycle(int sd, char *jobid)
 	return 0;
 }
 
+static int
```
suresh-thelkar, line 748: doxygen header please.

src/scheduler/fifo.c
```diff
@@ -798,6 +745,25 @@ scheduling_cycle(int sd, char *jobid)
...
+{
+	sched_cmd_t pcmd;
+
+	if (sconn->sfd == -1)
+		return 0;
+
+	if (!get_sched_cmd_noblk(sconn->sfd, &pcmd))
+		return 0;
+
+	if (pcmd.cmd == SCH_SCHEDULE_RESTART_CYCLE) {
```
suresh-thelkar, line 759: What if the command is other than SCH_SCHEDULE_RESTART_CYCLE. If this is the case we actually lose that scheduling command.

src/scheduler/fifo.c
```diff
@@ -1094,18 +1058,10 @@ main_sched_loop(status *policy, int sd, server_info *sinfo, schd_error **rerr)
...
-				if ((get_sched_cmd_noblk(second_connection, &cmd, &jid) == 1) &&
-					(cmd == SCH_SCHEDULE_RESTART_CYCLE)) {
-					log_event(PBSEVENT_SCHED, PBS_EVENTCLASS_JOB, LOG_WARNING,
-						njob->name, "Leaving scheduling cycle as requested by server.");
-					end_cycle = 1;
-				}
-				if (jid != NULL)
-					free(jid);
+			/* TODO: how to handle disconnet from server here */
+			if (has_prio_cmd(sconn)) {
```
suresh-thelkar, line 1062: We can actually rename this function definition as has_prty_cmd.

src/scheduler/fifo.c
```diff
@@ -1094,18 +1058,10 @@ main_sched_loop(status *policy, int sd, server_info *sinfo, schd_error **rerr)
...
-				/* get_sched_cmd_noblk() located in file get_4byte.c */
-				if ((get_sched_cmd_noblk(second_connection, &cmd, &jid) == 1) &&
-					(cmd == SCH_SCHEDULE_RESTART_CYCLE)) {
-					log_event(PBSEVENT_SCHED, PBS_EVENTCLASS_JOB, LOG_WARNING,
-						njob->name, "Leaving scheduling cycle as requested by server.");
-					end_cycle = 1;
-				}
-				if (jid != NULL)
-					free(jid);
+			/* TODO: how to handle disconnet from server here */
```
suresh-thelkar, line 1061: Are you going to remove TODO and put the corresponding code here ?

src/scheduler/get_4byte.c
```diff
@@ -66,41 +60,33 @@
...
+ * @retval	-1	: for error
  *
- * @note
- *		Callers need to free the value pointed to by 'jid' after returning
- *		from this call.
  */
-
 int
-get_sched_cmd(int sock, int *val, char **jid)
+get_sched_cmd(int fd, sched_cmd_t *cmd)
```
suresh-thelkar, line 73: I think we better call it as sock only instead of fd as it is not a file descriptor.

src/scheduler/get_4byte.c
```diff
@@ -113,35 +99,29 @@ get_sched_cmd(int sock, int *val, char **jid)
...
  * @retval	-1	for error.
- *
- * @note
- *		Callers need to free the value pointed to by 'jid' after returning
- *		from this call.
  */
 
 int
-get_sched_cmd_noblk(int sock, int *val, char **jid)
+get_sched_cmd_noblk(int fd, sched_cmd_t *cmd)
```
suresh-thelkar, line 114: Here also we can call this as sock because it is not a file descriptor but it is actually socket descriptor.

suresh-thelkar (changes requested):
Here are few more comments.

src/scheduler/fifo.c
hirenvadalia: This is same as earlier, earlier also it was ignoring any other command.

src/scheduler/fifo.c
hirenvadalia: why?  I fix has_prio_cmd more readable.

src/scheduler/get_4byte.c
hirenvadalia: why? there is no difference between socket and file descriptor...

src/scheduler/get_4byte.c
hirenvadalia: same as above, no diff between socket and file descriptor.

src/scheduler/data_types.h
hirenvadalia: why? its link to all svrs list...

src/include/pbs_sched.h
```diff
@@ -83,22 +83,21 @@ extern void *sched_attr_idx;
...
-	pbs_net_t pbs_scheduler_addr;
-	unsigned int pbs_scheduler_port;
-	time_t sch_next_schedule;		/* when to next run scheduler cycle */
-	char sc_name[PBS_MAXSCHEDNAME + 1];
-	struct preempt_ordering preempt_order[PREEMPT_ORDER_MAX + 1];
-	/* sched object's attributes  */
-	attribute sch_attr[SCHED_ATR_LAST];
-	short newobj;
+	pbs_list_link sc_link;					      /* link to all scheds known to server */
+	int sc_tmp_primary_conn;				      /* this is just to hold primary connection fd till secondary connection comes */
```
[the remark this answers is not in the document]
hirenvadalia: No, we can't without performance penalty, as we have to go though each connection in server, try to find respective sched (aka this struct) by matching primary connection (aka loop through each sched in server), then if get sched struct then match sched name with name came in secondary connection.

src/include/batch_request.h
```diff
@@ -277,6 +282,7 @@ struct batch_request {
 	char *tppcmd_msgid;			/* msg id for tpp commands */
 	struct batch_reply rq_reply;		/* the reply area for this request */
 	union indep_request {
+		struct rq_register_sched rq_rsched;
```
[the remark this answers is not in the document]
hirenvadalia: why? I find rq_rsched more readable then rq_sched...

src/include/batch_request.h
```diff
@@ -253,6 +253,11 @@ struct rqfpair {
 	char *fp_rmt;   /* used in Copy only     */
 };
 
+struct rq_register_sched {
+	char *rq_name;
+	char *rq_conn_type;
```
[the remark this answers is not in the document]
hirenvadalia: why? what we will gain by using enum?

src/scheduler/pbs_sched.c
```diff
@@ -791,46 +502,285 @@ are_we_primary()
...
+	CLEAR_HEAD(servers);
+}
+
+static int
+connect_server_helper(char *host, int is_primary)
+{
+	int rc;
+	int fd = -1;
+	struct batch_reply *reply = NULL;
+	char *conn_type = is_primary ? "primary" : "secondary";
```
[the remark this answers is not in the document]
hirenvadalia: why? this is as per design, and moreover what will gain by using enum?

src/scheduler/pbs_sched.c
```diff
@@ -791,46 +502,285 @@ are_we_primary()
...
+	while (cur) {
+		next = GET_NEXT(cur->all_svrs_link);
+		close_server(cur);
+		cur = next;
+	}
+	tpp_em_destroy(poll_context);
+	CLEAR_HEAD(servers);
+}
+
+static int
```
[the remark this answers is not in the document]
hirenvadalia: already done, in last commit.

src/scheduler/pbs_sched.c
```diff
@@ -791,46 +502,285 @@ are_we_primary()
...
+static void
+reconnect_server(svr_t *sconn)
+{
+	char *svrhost = sconn->svrhost;
+
+	close_server(sconn);
+	connect_server(svrhost);
+}
+
+static int
```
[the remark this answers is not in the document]
hirenvadalia: already done, please check last commit

src/scheduler/pbs_sched.c
```diff
@@ -791,46 +502,285 @@ are_we_primary()
...
+			log_errf(-1, __func__, "Failed to init cmd connections context");
+			exit(1);
+		}
+
+		CLEAR_HEAD(servers);
+	}
+
+	connect_server(NULL);
+}
+
```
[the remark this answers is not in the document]
hirenvadalia: already done, please last commit

src/scheduler/fifo.c
hirenvadalia: changed to `has_priority_cmd`

src/scheduler/get_4byte.c
hirenvadalia: As get_sched_cmd changed to take `svr_t *`, this comment becomes invalid now.

src/server/process_request.c
```diff
@@ -261,6 +261,98 @@ req_authenticate(conn_t *conn, struct batch_request *request)
...
+	 * scheduler (re-)connected, ask it to configure itself
+	 *
+	 * this must come after above reply_ack
+	 */
+	send_sched_cmd(sched, SCH_CONFIGURE, NULL);
+	return;
+
+rerr:
+	req_reject(rc, 0, preq);
+	close_client(preq->rq_conn);
```
bhroam, line 352: You shouldn't need to close the connection after a req_reject().

src/server/process_request.c
```diff
@@ -1410,7 +1514,12 @@ free_br(struct batch_request *preq)
 			break;
 
 #ifndef PBS_MOM		/* Server Only */
-
+		case PBS_BATCH_RegisterSched:
+			if (preq->rq_ind.rq_rsched.rq_name)
+				free(preq->rq_ind.rq_rsched.rq_name);
+			if (preq->rq_ind.rq_rsched.rq_conn_type)
+				free(preq->rq_ind.rq_rsched.rq_conn_type);
```
bhroam, line 1521: free(NULL) is a noop, there is no need to check if there is anything in them first.

src/server/run_sched.c
```diff
@@ -253,94 +202,69 @@ find_sched_from_sock(int sock)
...
 			return psched;
 	}
 	return NULL;
 }
 
 /**
  * @brief
- * 		contact_sched - open connection to the scheduler and send it a command
- *		Jobid is passed if and only if the cmd is SCH_SCHEDULE_AJOB
+ * 		Connects to the Scheduler requested.
```
bhroam, line 213: This isn't correct, you aren't connecting to a scheduler.

src/server/run_sched.c
```diff
@@ -253,94 +202,69 @@ find_sched_from_sock(int sock)
...
  *
- * @param[in]	cmd	- the command to send
- * @param[in]	jobid	- the jobid to send if 'cmd' is SCH_SCHEDULE_AJOB
+ * @return	void
  */
+void
+set_sched_state(pbs_sched *psched, char *state)
+{
+	set_attr_generic(&(psched->sch_attr[SCHED_ATR_sched_state]), &sched_attr_def[SCHED_ATR_sched_state], state, NULL, SET);
+	server.sv_attr[SVR_ATR_State].at_flags |= ATR_VFLAG_MODCACHE;
```
bhroam, line 224: How does the server's state change when the scheduler's state change?  There is no action function for the scheduler's state.

src/server/run_sched.c
```diff
@@ -253,94 +202,69 @@ find_sched_from_sock(int sock)
...
-	net_add_close_func(sock, scheduler_close);
+	pbs_sched *psched;
 
-	if (set_nodelay(sock) == -1) {
-		snprintf(log_buffer, sizeof(log_buffer), "cannot set nodelay on connection %d (errno=%d)\n", sock, errno);
-		log_err(-1, __func__, log_buffer);
-		return (-1);
-	}
+	for (psched = (pbs_sched*) GET_NEXT(svr_allscheds); psched; psched = (pbs_sched*) GET_NEXT(psched->sc_link)) {
+		if (psched->sc_secondary_conn == sock) {
```
bhroam, line 243: Just call find_sched_by_sock()

src/server/run_sched.c
```diff
@@ -253,94 +202,69 @@ find_sched_from_sock(int sock)
...
+			int rc;
+
+			DIS_tcp_funcs();
+			psched->sc_cycle_started = disrsi(sock, &rc);
```
bhroam, line 247: What is the purpose of this?  Why do we need to get this information from the scheduler?

src/server/sched_func.c
```diff
@@ -981,10 +957,21 @@ set_sched_default(pbs_sched *psched, int from_scheduler)
 		psched->sch_attr[SCHED_ATR_throughput_mode].at_flags |= ATR_SET_MOD_MCACHE | ATR_VFLAG_DEFLT;
 	}
 
+	if (!(psched->sch_attr[SCHED_ATR_version].at_flags & ATR_VFLAG_SET)) {
+		set_attr_generic(&(psched->sch_attr[SCHED_ATR_version]), &sched_attr_def[SCHED_ATR_version], PBS_VERSION, NULL, SET);
```
bhroam, line 961: If we are just going to set this to the server's version, remove the attribute.  You can get it from the server itself.

src/server/sched_func.c
```diff
@@ -981,10 +957,21 @@ set_sched_default(pbs_sched *psched, int from_scheduler)
...
+		psched->sch_attr[SCHED_ATR_version].at_flags |= ATR_SET_MOD_MCACHE | ATR_VFLAG_DEFLT;
```
bhroam, line 962: If you do keep this, no need to set ATR_SET_MOD_MCACHE.  set_attr_generic() sets that.

src/scheduler/pbs_sched.c
```diff
@@ -1279,84 +1185,49 @@ main(int argc, char *argv[])
...
+			strftime(log_buffer, sizeof(log_buffer), "%Y-%m-%d %H:%M:%S", localtime(&now));
+			DBPRT(("%s Scheduler received command %d\n", log_buffer, sconn->cmd));
+		}
+#endif
 
-			if (sigprocmask(SIG_SETMASK, &oldsigs, NULL) == -1)
-				log_err(errno, __func__, "sigprocmask(SIG_SETMASK)");
+		if (schedule(sconn)) /* magic happens here */ {
+			go = 0;
+		} else {
```
bhroam, line 1216: Why an else?  Don't you want to do a schedule() and then a send_end_cycle() right after it?

bhroam (changes requested):

src/server/run_sched.c
```diff
@@ -490,57 +398,35 @@ schedule_jobs(pbs_sched *psched)
...
-
-	set_sched_sock(-1, psched);
+	set_sched_state(psched, SC_DOWN);
+	log_eventf(PBSEVENT_SYSTEM, PBS_EVENTCLASS_SCHED, LOG_CRIT, psched->sc_name, "scheduler disconnected");
+	if (sock == psched->sc_primary_conn)
+		other_conn = psched->sc_secondary_conn;
+	else
+		other_conn = psched->sc_primary_conn;
+	psched->sc_secondary_conn = -1;
+	psched->sc_primary_conn = -1;
```
suresh-thelkar, line 419: other_conn can be either primary or secondary, but why are you assigning -1 to both of them in one shot. scheduler_close can actually get called twice once for primary and once per secondary right ?

src/server/sched_func.c
suresh-thelkar: I agree with @bhroam. We can just remove this attribute for Scheduler.

suresh-thelkar (changes requested):

src/include/batch_request.h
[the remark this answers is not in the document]
hirenvadalia: @bhroam string comparison you are talking about will be 2 times only per sched, so I don't think will get any real benefits in terms of performance. Anyways, I don't mind changing to enum but its design change, so will change both.

src/include/pbs_sched.h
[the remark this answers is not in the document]
hirenvadalia: @bhroam true number of sched will be less but number of conn in server can be high if many clients are connected to servers, and will have to loop through each connection -> loop though each sched per connection, so it will be huge performance penalty. And we can avoid those by just using one temporary variable so why not?
Also if assign sc_primary_conn then find_sched_from_sock which finds sched by comparing primary_conn or secondary conn will get sched which is not yet connected fully. so it will cause problem in wherever it is used.

src/include/sched_cmds.h
```diff
@@ -58,6 +58,3 @@
 #define SCH_SCHEDULE_ETE_ON 15		/* eligible_time_enable is turned ON */
 #define SCH_SCHEDULE_RESV_RECONFIRM 16		/* Reconfirm a reservation */
 #define SCH_SCHEDULE_RESTART_CYCLE 17		/* Restart a scheduling cycle */
-
-
-extern int schedule(int cmd, int sd, char *runjid);
```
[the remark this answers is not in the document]
hirenvadalia: its already declared in fifo.h

src/scheduler/Makefile.am
```diff
@@ -65,6 +65,7 @@ libpbs_sched_a_CPPFLAGS = ${common_cflags}
 
 libpbs_sched_a_SOURCES = \
 	$(top_builddir)/src/lib/Libpython/shared_python_utils.c \
+	$(top_builddir)/src/lib/Libutil/misc_utils.c \
```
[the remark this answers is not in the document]
hirenvadalia: I am not sure, but without this it was giving me linker error. will check it again.

src/scheduler/data_types.h
```diff
@@ -59,6 +59,7 @@ extern "C" {
 #include <time.h>
 #include <pbs_ifl.h>
 #include <libutil.h>
+#include "list_link.h"
```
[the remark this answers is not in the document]
hirenvadalia: Why can't we use link list in sched? I understand till now we have used it because we didn't needed it.

src/scheduler/data_types.h
```diff
@@ -176,6 +177,19 @@ typedef sch_resource_t usage_t;
 typedef void event_ptr_t;
 typedef int (*event_func_t)(event_ptr_t*, void *);
 
+typedef struct sched_cmd {
+	int cmd;   /* sched command */
+	char *jid; /* jobid assisiated with cmd if any else NULL */
+} sched_cmd_t;
```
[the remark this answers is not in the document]
hirenvadalia: As said, I am just following PBS standards... and we don't have different standards for sched...

src/scheduler/data_types.h
```diff
@@ -176,6 +177,19 @@ typedef sch_resource_t usage_t;
...
+
+typedef struct svr {
+	char *svrhost;
+	int primary_fd;		     /* primary connection to server - used for IFL calls */
+	int secondary_fd;	     /* secondary connection to server - used to receive sched cmds from server */
+	sched_cmd_t cmd;	     /* command received from server */
```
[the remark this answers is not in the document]
hirenvadalia: why separate them out when we need to pass all objects in svr_t (aka primary fd, secondary fd, cmd and jid) to schedule() and calls inside it so why not put it in one struct only and pass only one struct?

src/scheduler/data_types.h
```diff
@@ -176,6 +177,19 @@ typedef sch_resource_t usage_t;
...
+	pbs_list_link all_svrs_link; /* link to all servers */
+} svr_t;
```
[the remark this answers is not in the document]
hirenvadalia: why? its PBS standards to put _t suffix, and sched is part of PBS not an separate entity.

src/scheduler/fifo.c
```diff
@@ -798,6 +745,42 @@ scheduling_cycle(int sd, char *jobid)
...
+	sconn_copy.cmd.jid = NULL;
+	CLEAR_LINK(sconn_copy.all_svrs_link);
+
+	if (!get_sched_cmd_noblk(&sconn_copy))
+		return 0;
+
+	if (sconn_copy.cmd.cmd == SCH_SCHEDULE_RESTART_CYCLE) {
+		if (sconn_copy.cmd.jid)
+			free(sconn_copy.cmd.jid);
+		return 1;
```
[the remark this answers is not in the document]
hirenvadalia: Well my idea is to change only this func in case we ever add another high priority command so we don't have to sched main_sched_loop()

src/scheduler/fifo.h
```diff
@@ -217,7 +222,7 @@ void update_cycle_status(struct status *policy, time_t current_time);
  *			     deal with normal job can't run stuff
 
  */
-int main_sched_loop(status *policy, int sd, server_info *sinfo, schd_error **rerr);
+int main_sched_loop(status *, svr_t *, server_info *, schd_error **);
```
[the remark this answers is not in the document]
hirenvadalia: Again I am following PBS standards, where we don't keep names in declaration. And we have been removing from many places, so why not here?

src/scheduler/pbs_sched.c
```diff
@@ -793,44 +504,366 @@ are_we_primary()
...
+
+	if (!svrhost)
+		svrhost = pbs_default();
+
+	while (primary_fd < 0 || primary_fd < 0) {
+		if (primary_fd < 0) {
+			primary_fd = connect_server_helper(svrhost, 1);
+			if (primary_fd < 0) {
+				/* wait for 2s for not to burn too much CPU, and then retry connection */
+				sleep(2);
```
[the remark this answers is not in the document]
hirenvadalia: yes, one while making primary connection and another while making secondary connection.

src/scheduler/pbs_sched.c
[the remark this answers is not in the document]
hirenvadalia: @bhroam does sched has separate style? if yes then why? And how does it help in search exact line of code where mem error occurs if you are using generic message?

src/scheduler/pbs_sched.c
```diff
@@ -793,44 +504,366 @@ are_we_primary()
...
+	int max_connection = 1;
+
+	if (poll_context == NULL) {
+		poll_context = tpp_em_init(max_connection);
+		if (poll_context == NULL) {
+			log_errf(-1, __func__, "Failed to init cmd connections context");
+			die(-1);
+		}
+
+		CLEAR_HEAD(servers);
```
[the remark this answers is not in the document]
hirenvadalia: Why? what's problem in link list? in PBS we use link list when you want to group and link multiple same structure.

src/scheduler/pbs_sched.c
```diff
@@ -793,44 +504,366 @@ are_we_primary()
...
+{
+	int nfds;
+	int i;
+	em_event_t *events;
+	int err;
+	int hascmd = 0;
+	sigset_t emptyset;
+
+
+again:
```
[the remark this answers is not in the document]
hirenvadalia: Well I didn't find any better way to do it, if you know any better way to handle this without goto please let me know, I will change code.

src/scheduler/pbs_sched.c
hirenvadalia: if schedule() returns !0 then will be go out of this loop, and very next thing we do is close server connections, which can lead server to read on closed fd, aka server will get EOF not what sched sent, so no in cause if we are going out of loop we don't have to send end cycle notification as server will get notified by closed connection.

src/server/pbsd_main.c
```diff
@@ -1436,13 +1393,6 @@ main(int argc, char **argv)
 	}
 	process_hooks(periodic_req, hook_msg, sizeof(hook_msg), pbs_python_set_interrupt);
 
-	/*
-	 * Make the scheduler (re)-read the configuration
-	 * and fairshare usage.
-	 */
-	(void)contact_sched(SCH_CONFIGURE, NULL, pbs_scheduler_addr, pbs_scheduler_port);
-	(void)contact_sched(SCH_SCHEDULE_NULL, NULL, pbs_scheduler_addr, pbs_scheduler_port);
-
```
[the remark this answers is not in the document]
hirenvadalia: we are sending SCH_CONFIGURE as soon as sched connects to server, please see req_register_sched() in process_request.c, and despite that also fairshare test fails.

src/server/process_request.c
hirenvadalia: @bhroam req_reject() doesn't close connection, we have to explicitly call close connection after it.

src/server/run_sched.c
hirenvadalia: find_sched_by_sock, checks primary and secondary connection with given sock, but here we have to check only secondary connection, not primary connection.

src/server/run_sched.c
hirenvadalia: this is cycle end notification, as we are not closing connection to notify server like earlier. Please see send_cycle_end() in pbs_sched.c

src/server/run_sched.c
hirenvadalia: Yes, if any of the connections is closed, server will close both connection, and if we depend on close_func() of other conn then it will be try to close each other in loop, also see blow we are removing close_func and closing other connection here only.

src/server/sched_func.c
hirenvadalia: @bhroam You only told in one of our mail discussion that there is someone who is using this attribute, if that's not true then sure I will remove this attribute.

src/scheduler/data_types.h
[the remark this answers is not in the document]
suresh-thelkar: I am actually talking about the Server's port which is useful in future. Since this svr_t is meant for server's info better to have the port on which Server is using.

src/scheduler/pbs_sched.c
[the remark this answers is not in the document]
suresh-thelkar: Ideally it should not happen when server sends SCH_CONFIGURE to scheduler as soon as it comes. Can you please elaborate more on which log message you are not seeing and what is hapenning.

src/scheduler/data_types.h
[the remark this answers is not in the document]
hirenvadalia: @suresh-thelkar I understand you point, but we use pbs_connect() where we need `host[:port]` syntax, and I am not able to collect any use case for server port for now, if we have any use case where we will use it in future I am happy to add it.

src/scheduler/pbs_sched.c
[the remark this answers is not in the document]
suresh-thelkar: The gain is readability and also we can avoid string comparison.

src/include/batch_request.h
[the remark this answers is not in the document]
suresh-thelkar: I generally read rq_sched as "request from sched". rq-> stands for request. Just wondering what does 'r' in rq_rsched stands for ? If it also stands for request then we should not have it, it is redundant else we are good.

src/scheduler/pbs_sched.c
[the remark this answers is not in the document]
hirenvadalia: As I have explained as part of other comment also, a) this is design change b) I don't mind to change it to enum but will not gain anything out of it as using explicit string is more readable then enum and c) they won't be much string comparison, its just 2 times for one sched

src/scheduler/data_types.h
suresh-thelkar: This is because it only has the link for next server. That link will again have a link for further servers. Look at other places for example structure pbs_sched also we only call the link as sc_link we don't call this as all_sc_link.

src/include/batch_request.h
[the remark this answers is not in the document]
hirenvadalia: 'r' stand for 'register' not request, and if you see below we use variable name same as struct name, but I find that's too long here, so I made it short, but removing 'r' will loose readability, I happy to make it long like other if 'r' is not more readable.

src/scheduler/data_types.h
hirenvadalia: pbs_list_head and pbs_list_link is doubly link list, and this link can be used to find next as well as prior also... and in my view sc_link in pbs_sched is not correct it should be all_sc_link only, see in job->ji_alljobs in job struct.

src/scheduler/fifo.c
suresh-thelkar: No. This is not same as earlier. It is because prior to this change only SCH_SCHEDULE_RESTART_CYCLE used to be sent on secondary connection. Apart from this nothing else was sent. But as per the current design all scheduling commands are sent on the secondary connection and hence there is a possibility that we might end up losing a command other than SCH_SCHEDULE_RESTART_CYCLE. This definitely causes problems going forward especially if we have multiple servers.

src/scheduler/get_4byte.c
suresh-thelkar: why no difference. The underlying mechanism of socket vs file descriptor are different right. Eager to what is the problem if we call it as sd instead of fd?  It is more readable that way.

src/scheduler/fifo.c
hirenvadalia: @suresh-thelkar server doesn't send any other command except restart_cycle till its get end cycle notification. and we read prio command only from server who triggered sched cycle so in case multiple server will not it won't any problem.

src/scheduler/get_4byte.c
hirenvadalia: @suresh-thelkar signature of this func is changed as per your comment, so no there is not more fd or sock, means this comment is invalid now.

src/include/batch_request.h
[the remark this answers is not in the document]
suresh-thelkar: May be make it longer which avoids the confusion.

src/server/run_sched.c
hirenvadalia: Remove this func as we now have easy & good attr setters

src/server/run_sched.c
hirenvadalia: I am also not sure where did that come, may copy paste error. anyways this func is removed wrt to set_attr_generic

src/server/pbsd_main.c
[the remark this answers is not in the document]
suresh-thelkar: If PTL test case is only the problem and it is looking for the message "Scheduler is reconfiguring" and it is not getting it because of the general default log_event_mask can we move the following statement after the function call set_validate_sched_attrs in case SCH_CONFIGURE: of schedule() function.
log_event(PBSEVENT_SCHED, PBS_EVENTCLASS_SCHED, LOG_INFO,
		"reconfigure", "Scheduler is reconfiguring");
If this is the last line of the above case then I think it should work. We can give a try.

If we do this as part of the function call set_validate_sched_attrs we fetch log_events from server.

src/server/run_sched.c
suresh-thelkar: If we are calling close() for both the sockets then it should fine to put -1 for both of them. If by any chance we do not call close() but we are assign -1 to one of the sockets might cause a problem. Please cross check.

src/server/pbsd_main.c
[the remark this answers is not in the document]
hirenvadalia: Yes @suresh-thelkar that's the case, and changing position log_event will work, but I have alternate solution, by default almost all log events (except high debug levels - aka DEBUG3 and DEBUG4) is added to log_event_lvl_priv in pbs_event.c but PBSEVENT_SCHED is not added to it and I am not sure why?
@bhroam Do you know why we have not added PBSEVENT_SCHED in log_event_lvl_priv (aka default log event mask)? Can we add it or not?

src/scheduler/pbs_sched.c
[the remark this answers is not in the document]
hirenvadalia: removed sleep from connect_server_helper.

src/server/pbsd_main.c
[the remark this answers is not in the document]
bhroam: @suresh-thelkar We don't want to send a SCH_CONFIGURE when the server starts.  There is a specific scheduler command which is sent (SCH_SCHEDULE_FIRST).  If there is something to be done when the server starts, it should be done there.  @hirenvadalia I can only guess at the answer to your question because it was done before my time.  My guess is that originally only the server and mom were written, and a shell of a scheduler.  Since the scheduler wasn't written, PBSEVENT_SCHED was either left out, or it was a mistake.  For 20+ years, the scheduler didn't use log_event(), so it didn't matter.  Now that the scheduler uses log_event(), we need to add PBSEVENT_SCHED into the default log mask.  I wouldn't add DEBUG2-4 though.

src/include/pbs_sched.h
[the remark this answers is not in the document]
bhroam: Ahh, you are looping through the connections.  Yes, that will slow things down.  As for the temp variable, I am not sure I find a problem with storing the primary connection in the primary connection variable.  Finding a scheduler by the primary sock shouldn't be a problem unless you are trying to run a cycle.  You can easily add a check when you are about to run a cycle that both the primary and secondary descriptors are there before you start it.  Otherwise, I think it would be OK to find it.  It IS the scheduler's descriptor.  If you don't like this, just modify find_sched_by_sock() to make sure there is a valid primary and secondary connection before you search.

src/include/sched_cmds.h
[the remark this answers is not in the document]
bhroam: Remove the copy in fifo.h and declare it here.  It is part of the external interface for pbs_sched.c

src/scheduler/data_types.h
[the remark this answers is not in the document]
bhroam: It's how the scheduler is designed.  I see no reason to start using it as a one-off.

src/scheduler/data_types.h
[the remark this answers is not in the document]
bhroam: I find consistency more important.  Please follow the style of the scheduler.

src/scheduler/data_types.h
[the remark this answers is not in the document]
bhroam: Please follow scheduler style.  It is different than the server/mom.  The scheduler has a distinct style that I set down when I originally wrote the scheduler.  I try and stick to it for consistency.  In this case, we have data which defines a server, and transient data which defines a cycle.  The server is defined by the host and the connection descriptors.  The cycle is defined by the command.  They are separate.

src/scheduler/data_types.h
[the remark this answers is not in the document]
bhroam: Like I said before, the consistency within the scheduler is more important than strict adherence to the standard.

src/scheduler/fifo.c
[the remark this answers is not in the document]
bhroam: How is knowing there is a high priority command useful without knowing the command?  Right now there is only one, so returning 1/0 works.  Once there are two, you won't know what the high priority command is, so you would have to change main_sched_loop(), because you wouldn't want to restart the cycle if the second command is sent.

src/scheduler/fifo.h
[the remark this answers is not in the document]
bhroam: I just reviewed the standard and there is no guidance for declarations/prototype variable names.  It talks about where to put them (private vs public), but not whether or not to put variable names in them.

src/scheduler/pbs_sched.c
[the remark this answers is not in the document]
bhroam: The scheduler does not use linked lists for its primary data structures.  The server/mom do.  They are different daemons and have different purposes.  Please follow scheduler style and keep the scheduler consistent.

src/server/sched_func.c
bhroam: I'm sorry if I gave that impression.  I thought I said I didn't know if anyone was using the attribute.  In any case, right now it is not the scheduler's version, it is the server's version.  If for any reason someone did start up a scheduler with a different version, we'd print the wrong one.  So either remove the attribute, or have the scheduler set it.  I say remove it.  You might want to check with the PMs first though.

src/scheduler/pbs_sched.c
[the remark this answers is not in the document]
bhroam: The scheduler has its own way of doing things that is different than the server/mom.  The daemons were written at different times for different purposes.  For example, the scheduler uses pointer arrays for ease of sorting and because the scheduler needs to have multiple arrays of the same data for different purposes (e.g. running jobs, exiting jobs, jobs in reservations that are running, etc).  Doing this with linked lists would be much much more difficult.  This is just one of the things I decided that was how the scheduler would do things early on and we always follow it.  For the scheduler, determining which malloc() fails doesn't really help.  When a malloc() fails we try and gracefully leave the cycle and hope more memory is available for the next one.  We don't care which malloc() failed, just that one did.  Being able to search for a specific string helps with that.

src/scheduler/pbs_sched.c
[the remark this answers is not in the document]
bhroam: Replace the goto again: by a *sconn = NULL and break and add the hascmd = 1 in an else case.

src/server/process_request.c
bhroam: I guess what confused me is that this is the first time I've seen a req_reject() followed by a close_client().  Although I think the req_reject() frees the preq.  If that is the case, you'll have to save the rq_conn to close.

src/server/run_sched.c
bhroam: Is this the right thing to do, or should we make find_sched_by_sock() more generic and be able to search by primary connection, secondary connection, or both?

src/server/run_sched.c
bhroam: I guess my confusion is why we're reading something into this variable.  Is it possible that we might read true or will we be always be setting it to false?

src/scheduler/pbs_sched.c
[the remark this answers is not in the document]
bhroam: I'd vote for the enum because I generally shy away from strings in C.  If we were in python, I'd be fine with it.

src/scheduler/fifo.c
bhroam: The command to restart a cycle comes from a mom hook.  If we only read high priority commands from the server which started our cycle, what happens if a node from a different server sends a restart cycle?  Don't we need to read high priority commands from all servers?

src/server/run_sched.c
bhroam: Well set_attr_generic() will set SET_MOD_MCACHE for the scheduler attribute since it is modifying the scheduler attribute.  I'm just unfamiliar with what the server's state changes to when a scheduler goes into scheduling.  Does the server's state go into scheduling if any scheduler moves to scheduling, or just the default scheduler?  Maybe @suresh-thelkar might know since he was involved with the multi-sched feature.

suresh-thelkar: Yes @bhroam, server's state go to scheduling if only default scheduler moves to scheduling state.

src/scheduler/pbs_sched.c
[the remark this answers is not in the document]
hirenvadalia: conn type is removed from register batch request, as now first connection will be primary and second connection will secondary connection, so now need of string or enum and this comment is invalid

src/include/batch_request.h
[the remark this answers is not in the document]
hirenvadalia: rq_conn_type is removed from rq_register_sched, so this comment is invalid now.

src/include/pbs_sched.h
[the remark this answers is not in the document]
hirenvadalia: @bhroam initially when I started developing this changes, I used primary connection variable only (earlier it was named as 'scheduler_sock'), and when I run regression I show failures and and while debugging I found that root cause was early assignment of primary connection, so I used another tmp variable, but if you are saying that it should cause problem then I will give a try again without tmp variable and see whether it works or not.

src/scheduler/fifo.c
hirenvadalia: @bhroam True we will have to read commands from all servers in case of multi server, but that change will come (soon) as part of multi server changes (as there are few dependent code which we need first before we can make change here for multi-server).

src/server/process_request.c
hirenvadalia: Yes, I have handled saving preq->rq_conn in last commit.

src/server/run_sched.c
hirenvadalia: It would be good if we can make find_sched_by_sock more generic, I will give a try (may be I will need help as I still don't know many part of sched in server)

src/server/run_sched.c
hirenvadalia: I will change this to make more readable, actually since current cycle_end_marker is 0 and here we have to make 'sc_cycle_started' 0 as we got end cycle notification, i just combined those to in one like, ideally it should have been:
```c
disrsi(sock, &rc); /* read end cycle marker and ignore as we don't need its value */
psched->sc_cycle_started = 0;
// check for rc, and make sched state based on rc - aka SC_DOWN or SC_IDLE
```

src/server/pbsd_main.c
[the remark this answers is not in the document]
hirenvadalia: @bhroam Thank for details, I will add PBSEVENT_SCHED in default log mask, however DEBUG2 is already part of default log mask, not sure why, but will keep that as it is.

On:
> @suresh-thelkar We don't want to send a SCH_CONFIGURE when the server starts. There is a specific scheduler command which is sent (SCH_SCHEDULE_FIRST). If there is something to be done when the server starts, it should be done there. 

@bhroam Are you sure that SCH_SCHEDULE_FIRST should go before SCH_CONFIGURE as soon as sched connects to server? (or earlier when server was connection to sched)
As what I see in mainline code (and which I retained in this PR also), is as soon as server starts the very first command it sends to sched is SCH_CONFIGURE -> SCH_SCHEDULE_NULL-> SCH_SCHEDULE_FIRST, this is the sequence goes to sched.
And in this PR changes, I removed SCH_SCHEDULE_NULL as it doesn't do anything in sched, sched just ignores it.

src/scheduler/fifo.c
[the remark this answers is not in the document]
hirenvadalia: removed this func and added this code in main_sched_loop()... this comment is invalid now.

src/server/pbsd_main.c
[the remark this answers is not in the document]
bhroam: Interesting, I didn't realize a SCH_CONFIGURE was sent.  I thought it was only a SCH_SCHEDULE_FIRST.  I'm not necessarily sure we need both, but I don't see any reason to change it right now if that's what is happening.  Knowing the SCH_CONFIGURE is sent, I probably can remove some code from SCH_SCHEDULE_FIRST.

src/scheduler/pbs_sched.c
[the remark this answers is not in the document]
bhroam: I actually liked the fact that the batch request registered a connection as primary or secondary.  We've run into trouble before when the scheduler and server got out of sync.  Think of a case when the scheduler makes the primary connection, crashes, restarts, and then makes the primary connection again.  The server will mark the second connection as secondary, and then when the scheduler connects for the third time, it will be rejected.  If the scheduler registers two primary connections, you can discard the first and then everything is still fine.

bhroam (commented):
Sorry for the same comment twice.  I made it and didn't send it, and then I found the correct conversation and made it again.  Now I can't find the one I need to delete.  Sometimes github can be annoying and hiding things when I'd rather it didn't.

src/include/pbs_sched.h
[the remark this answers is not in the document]
bhroam: I'm really surprised there was a problem.  The only time in the code we should have had a socket to search for was if a cycle was active.  When we're registering connections, we can't have a cycle active, so none of that code should be executing.  If it is, then we're assuming that if we can find a scheduler by a socket, then a cycle is active. If that is the case, that code is now wrong.  In any case, I don't like the idea of the temp variable.  I just don't think it is necessary.  If you truly find that it is, then so be it, but I'd rather find out why things weren't working right before.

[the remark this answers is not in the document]
hirenvadalia: > The only time in the code we should have had a socket to search for was if a cycle was active. When we're registering connections, we can't have a cycle active, so none of that code should be executing. If it is, then we're assuming that if we can find a scheduler by a socket, then a cycle is active. If that is the case, that code is now wrong.

Yes, this is the case. Since earlier server was connecting to sched so we had sockets only when cycle was active, but now since we will always have socket no matter whether cycle is active or not. But I agree with you that code shouldn't assume then if it finds sched by a socket then it shouldn't assume that cycle is active, I will look into this and try to fix it.

src/server/process_request.c
```diff
@@ -287,19 +288,15 @@ req_register_sched(conn_t *conn, struct batch_request *preq)
 		rc = PBSE_BADHOST;
 		goto rerr;
 	}
-	if (sched->sc_primary_conn != -1 || sched->sc_secondary_conn != -1) {
+	if (sched->sc_primary_conn != -1 && sched->sc_secondary_conn != -1) {
 		rc = PBSE_SCHEDCONNECTED;
 		goto rerr;
 	}
-	if (strcmp(preq->rq_ind.rq_rsched.rq_conn_type, "primary") == 0) {
+	if (sched->sc_tmp_primary_conn == -1 && sched->sc_primary_conn == -1) {
```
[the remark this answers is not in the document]
hirenvadalia: No, we are not longer sending 'conn_type' anymore, as it will allow us (in future) to fulfill use cases like your 2nd pbs_est use case, where will have only connection (not 2 connections like in sched) and still mark connection as super privileged connection. (just FYI, for now we are not saying anything for other use cases in design or not making any code/func names generic, those will be taken care by next project (aka - non-root daemons project - which is dependent on this changes), as will have better picture there)

About, out of sync, it shouldn't happen as server will get notified for close connection if sched crashes or restarts just after making first connection (aka primary connection), as its tcp connection and if process dies then that connection will be closed and other side of connection (aka server in this case) will get notified. And once server gets notified it should remove socket from sched obj.

hirenvadalia:
@suresh-thelkar and @bhroam I have addressed all address given, please review the latest commit and provide your feedback.

src/scheduler/pbs_sched.c
```diff
@@ -616,48 +617,54 @@ connect_server_helper(char *host, int is_primary)
...
-	svr->svrhost = svrhost;
-	svr->primary_fd = primary_fd;
-	svr->secondary_fd = secondary_fd;
-	CLEAR_LINK(svr->all_svrs_link);
-	append_link(&servers, &svr->all_svrs_link, svr);
-	set_validate_sched_attrs(svr->primary_fd);
+	svr->svrhost = strdup(svrhost);
+	if (!svr->svrhost) {
+		free(svr);
+		log_errf(-1, __func__, MEM_ERR_MSG);
```
bhroam, line 660: shouldn't strdup() set errno?

src/scheduler/pbs_sched.c
```diff
@@ -616,48 +617,54 @@ connect_server_helper(char *host, int is_primary)
...
+		die(-1);
+	}
+	svr->primary_sock = primary_sock;
+	svr->secondary_sock = secondary_sock;
+	i = 0;
+	while (servers[i] != NULL) i++;
+	servers[i] = svr;
```
bhroam, line 667: Consider using add_ptr_to_array()

src/scheduler/pbs_sched.c
```diff
@@ -693,79 +710,138 @@ connect_servers(void)
...
+			if (!ds_enqueue(sched_cmds, cmd_prio)) {
+				if (cmd_prio->jid)
+					free(cmd_prio->jid);
+				free(cmd_prio);
+				/*
+				 * no error return here,
+				 * as we have successfully enqueued first command
+				 */
+			}
+			/* no need of return here, fall to last return */
```
bhroam, line 817: I think the code easily shows this.  Remove this comment.

src/server/run_sched.c
```diff
@@ -386,24 +388,29 @@ void
...
-	int other_conn;
+	int other_conn = -1;
 
-	psched = find_sched_from_sock(sock);
-
-	if (psched == NULL)
-		return;
+	psched = find_sched_from_sock(sock, CONN_SCHED_PRIMARY);
+	if (psched == NULL) {
+		psched = find_sched_from_sock(sock, CONN_SCHED_SECONDARY);
```
bhroam, line 395: Consider adding a 3rd option of CONN_SCHED_ANY.  It'll save you from calling this twice.  Since the number of schedulers will be small, it isn't a big deal either way.

bhroam (changes requested):

src/scheduler/pbs_sched.c
hirenvadalia: Thanks, I was looking for func like that in sched code, but couldn't find it, as add_* is in server_info.c/.h and remove_* is in misc.c/.h, will change to use add_ptr_to_array()

bhroam: Oh, that sounds like a mistake.  The add and remove functions should both be in misc.c.  I'll have to move it at some point.  Unless you want to do it.  Don't feel obligated to, though.

src/scheduler/pbs_sched.c
```diff
@@ -792,45 +506,424 @@ are_we_primary()
...
+		 */
+		if (cmd_prio->cmd == SCH_SCHEDULE_RESTART_CYCLE) {
+			if (cmd_prio->jid)
+				free(cmd_prio->jid);
+			free(cmd_prio);
+			/*
+			 * no return here,
+			 * as we have successfully enqueued first command
+			 */
+		} else if (cmd_prio->cmd != -1) {
```
suresh-thelkar, line 806: Just a question. Is this else if block required. We can at most get two commands from a single server in the worst case i.e. single normal priority command followed by a super high priority command like SCH_RESTART_SCHED_CYCLE but we never get normal priority command followed by an other normal priority command at a time right ?

src/scheduler/pbs_sched.c
```diff
@@ -792,45 +506,424 @@ are_we_primary()
...
+		 * return any failure if fails to read, as we have
+		 * successfully enqueued first command
+		 */
+		cmd_prio = malloc(sizeof(sched_cmd));
+		if (cmd_prio == NULL)
+			return 1; /* return 1 as we have enqueued first command */
+		cmd_prio->cmd = -1;
+		cmd_prio->jid = NULL;
+
+		get_sched_cmd_noblk(sock, cmd);
```
suresh-thelkar, line 791: I think it is better to test the return value of get_sched_cmd_noblk. It is because it might return 0 which indicates EOF i.e. server is closed.

src/scheduler/fifo.c
```diff
@@ -1094,18 +1039,40 @@ main_sched_loop(status *policy, int sd, server_info *sinfo, schd_error **rerr)
...
+					sched_cmd *cmd;
+					if (servers[i]->secondary_sock <= 0)
+						continue;
+					cmd = malloc(sizeof(sched_cmd));
+					if (cmd == NULL) {
+						log_err(errno, __func__, MEM_ERR_MSG);
+						continue;
+					}
+					cmd->jid = NULL;
+					if (get_sched_cmd_noblk(servers[i]->secondary_sock, cmd) != 1) {
```
suresh-thelkar, line 1054: Here also I think we need to handle the case of server going down i.e. if get_sched_cmd_noblok returns EOF.

src/scheduler/pbs_sched.c
hirenvadalia: You are right, but I wrote it so that in future we if ever change that (aka sending non-prio command after prio command or add a new prio command) then we just have to add only one else if for that new command. If you still think we don't need it then I will remove it.

src/scheduler/pbs_sched.c
hirenvadalia: good point, missed that. will change

src/scheduler/fifo.c
hirenvadalia: I am not sure what to do here if we detect EOF as its main_sched_loop() and I still understanding internals of sched. May be you or @bhroam can guide me what should we do here if we detect EOF from server side here?

src/scheduler/pbs_sched.c
hirenvadalia: Sure, will move it, (assuming) its just copy paste from one file to another.

src/server/dis_read.c
```diff
@@ -566,6 +566,10 @@ dis_request_read(int sfds, struct batch_request *request)
 			break;
 
 #ifndef PBS_MOM
+		case PBS_BATCH_RegisterSched:
+			request->rq_ind.rq_register_sched.rq_name = disrst(sfds, &rc);
```
suresh-thelkar, line 570: Hope we are initialising rq_name. If not if disrst() returns EOF then Scheduler should have been killed and causes problems.

hirenvadalia: Sorry didn't get you question completely...
If you are asking for initializing rq_name, its handled by alloc_br().
If you are asking for checking rc for EOF (or any other error) after disrst() then it handled down in code, after this switch statement.

suresh-thelkar: I was just asking about initialisation only which is handled by alloc_br() as we are doing memset inside. So we are good there.

src/server/process_request.c
```diff
@@ -261,6 +261,96 @@ req_authenticate(conn_t *conn, struct batch_request *request)
 	reply_ack(request);
 }
 
+#ifndef PBS_MOM
+static void
+req_register_sched(conn_t *conn, struct batch_request *preq)
```
suresh-thelkar, line 266: Can you please add doxygen header.

src/scheduler/pbs_sched.c
hirenvadalia: @suresh-thelkar I tried adding check for 0 here and it always return after this call as for cmd_noblk, 0 means no cmd to read also not just EOF (for blocking version 0 means EOF). So I think we can't add check for 0 here.
And also I think we don't need any check for this call, as we are not returning any error once we reach this point due we already enqueued one command in cmds queue and from this point we are just checking whether prio command is there or not before we call to schedule()

suresh-thelkar: > @suresh-thelkar I tried adding check for 0 here and it always return after this call as for cmd_noblk, 0 means no cmd to read also not just EOF (for blocking version 0 means EOF). So I think we can't add check for 0 here.

The problem is zero being used for both 1) No command  and 2)EOF case. We can actually modify this function to return a separate return code for both these cases. Don't know whether we need to go to that extent.
I was thinking if we catch server going down here it is an early catch. Anyways because of the above reasons it is fine not do this for the time being I think.

Seems the following line of function header is wrong to me. I prefer to change it to as follows so that it is more clear. It is up to you.

* @retval	0	for EOF
We can change this to 
* @retval	0	if there is no super high priority command or EOF in case of server going down

hirenvadalia: sure I will change func header to mention it.

src/scheduler/pbs_sched.c
suresh-thelkar: Please remove this as it adds an extra confusion to have this code now itself. May be when we have a new super high priority command in future we can add it accordingly.

src/scheduler/fifo.c
suresh-thelkar: Just my thought. I think it is better to end the current cycle when we get EOF here which indicates server is down. Better to catch server going down early. Even if it is ignored here it can fail further for example like while sending attribute updates or when it calls any IFL call after this because it can get SIGPIPE. And also our final goal is we should not continue scheduling even when one server is down. 

Earlier case is little different because super high priority commands were only sent on secondary connection, but now we send both normal/medium/super high priority commands on the same secondary connection.

Having said that,  Let us take Bhroam's suggestion on this and based on his inputs we can proceed accordingly. @bhroam can you please let us know your opinion.

bhroam: Usually what happens when the server goes down is we try and write over the primary connection and get a SIGPIPE.  We take note of this and exit the cycle.  Now that we are in charge of opening the connections, just exiting the cycle won't be enough.  Won't we need to reopen those connections?  In any case, you will need to return a value from this function that we can catch in main_sched_loop and set end_cycle = 1.

src/scheduler/fifo.c
```diff
@@ -745,6 +745,47 @@ scheduling_cycle(sched_svrconn *sconn, sched_cmd *cmd)
...
+{
+	if (servers) {
+		int i = 0;
+		for (; servers[i] != NULL; i++) {
+			sched_cmd *cmd;
+
+			if (servers[i]->secondary_sock <= 0)
+				continue;
+			cmd = new_sched_cmd();
+			if (!cmd)
```
bhroam, line 769: Say if (cmd == NULL)

src/scheduler/pbs_sched.c
```diff
@@ -781,41 +781,16 @@ read_sched_cmd(int sock)
...
-				free(cmd_prio);
-				/*
-				 * no error return here,
-				 * as we have successfully enqueued first command
-				 */
-			}
-			/* no need of return here, fall to last return */
-		}
+		get_sched_cmd_noblk(sock, cmd_prio);
+		free_sched_cmd(cmd_prio);
```
bhroam, line 793: How can we be sure this is a high priority command?

src/scheduler/fifo.c
hirenvadalia: Also note that there is issue in get_sched_cmd_noblk() then it doesn't differenciate between 'no super high prio command from server' vs EOF on closed connection. In both case it returns 0 only.
@bhroam (Assuming we are able to fix get_sched_cmd_noblk()), How about we do something like in main_sched_loop():
```c
if (!end_cycle) {
	int rc = -1;
	sched_cmd *cmd = get_high_prio_cmd(&rc);
	if (rc == 0) {
		log_event(PBSEVENT_SCHED, PBS_EVENTCLASS_JOB, LOG_WARNING,
				njob->name, "Looks like we lost connection with one of the server, leaving scheduling cycle");
		end_cycle = 1;
	} else if (cmd && cmd->cmd == SCH_SCHEDULE_RESTART_CYCLE) {
		log_event(PBSEVENT_SCHED, PBS_EVENTCLASS_JOB, LOG_WARNING,
				njob->name, "Leaving scheduling cycle as requested by server.");
		end_cycle = 1;
	}
	free_sched_cmd(cmd);
}
```

src/scheduler/pbs_sched.c
```diff
@@ -642,29 +646,33 @@ connect_server(char *svrhost)
...
 	}
 	svr->primary_sock = primary_sock;
 	svr->secondary_sock = secondary_sock;
-	i = 0;
-	while (servers[i] != NULL) i++;
-	servers[i] = svr;
+	svrs_newarr = add_ptr_to_array(servers, svr);
+	if (svrs_newarr == NULL) {
+		close_server(svr);
+		/* no need of logging as add_ptr_to_array already logged error */
```
bhroam, line 672: It is standard practice in the scheduler to log at the point of the malloc().  There is no need for this comment.

bhroam (changes requested):
Thanks for making the changes @hirenvadalia.  Just a couple of minor comments left.

src/scheduler/fifo.c
bhroam: This looks fine to me.  I'd rephrase the first message slightly to be more definite.  Drop the 'looks like' part.

src/scheduler/pbs_sched.c
hirenvadalia: @bhroam We already read one command from given secondary connection, so if we get 2nd command from same connection (aka same server) then it must be priority command only. As once server send non-priority command, it won't any other command (except priority command) till it receives end cycle notification.

However my earlier code was not dependent on above fact, where it was enqueuing command in sched_cmds if not priority command, but @suresh-thelkar finds that confusing code and gave comment to remove it. ([this](https://github.com/openpbs/openpbs/pull/2000/commits/a612648971d373d7281e90e29ae076b2c2ae4c93#diff-5379b5f9c171cb04e099383cb416bf83L798) is the if/else block I am talking about)

suresh-thelkar: It looks like this got resolved now. I saw thumbs up sign from @bhroam. 

But just adding my view why I felt the earlier code i.e. having that if - else if block causes confusing.

As per the current code, server at the most can send two commands i.e. regular sched command followed by a super high priority command before it receives an end of cycle. This is possible, what is not possible is server sending regular command followed by another regular command before it receives an end of cycle and the earlier code with that else if block was doing that and hence requested Hiren to remove that part as this will never happen.

src/scheduler/fifo.c
```diff
@@ -798,6 +745,57 @@ scheduling_cycle(int sd, char *jobid)
...
+static sched_cmd *
+get_high_prio_cmd(int *is_conn_lost)
+{
+	if (servers) {
+		int i = 0;
+		for (; servers[i] != NULL; i++) {
+			sched_cmd *cmd;
+			int rc;
+
+			if (servers[i]->secondary_sock <= 0)
```
suresh-thelkar, line 770: Isn't it servers[i]->secondary_sock < 0. Because 0 is valid descriptor.

src/scheduler/pbs_sched.c
```diff
@@ -792,45 +506,400 @@ are_we_primary()
...
+/**
+ * @brief close connections to all servers
+ *
+ * @return void
+ */
+static void
+close_servers(void)
+{
+	while (count_array(servers))
+		close_server(servers[0]);
```
suresh-thelkar, line 543: Do we need to have a while loop to free servers here. Why cannot we make use of free_ptr_array() function.

src/scheduler/pbs_sched.c
```diff
@@ -792,45 +506,400 @@ are_we_primary()
...
+ * @param[in] sconn - pointer server struct to reconnect
+ *
+ * @return void
+ */
+static void
+reconnect_server(sched_svrconn *sconn)
+{
+	char svrhost[PBS_MAXSERVERNAME + PBS_MAXPORTNUM + 2];
+
+	strcpy(svrhost, sconn->svrhost);
```
suresh-thelkar, line 719: please use strncpy.

src/scheduler/pbs_sched.c
hirenvadalia: free_ptr_array() just frees array but doesn't free anything in ptr (aka entries of array).

src/scheduler/pbs_sched.c
hirenvadalia: I don't think we need it. Because, sconn->svrhost comes from pbs.conf and max value of it what I have used as a sizeof svrhost[] above. So strcpy will be suffice.

src/scheduler/pbs_sched.c
suresh-thelkar: It also frees the individual elements of the array provided the members of the individual elements do not contain pointers. In our case we have to free svrhost of sch_svrconn structure in which case we cannot directly use it. But regardless we can follow the same style.

Better not to call count_array function every time in a loop rather can we simply do like the below which avoids calling count_array() function every iteration of the loop.
```
for (i = 0; servers[i] != NULL; i++)
	close_server(servers[i]);
free(servers);
```

hirenvadalia: I had initially used same only, but it give me leak and other problems like invalid read while accessing server[i], as we have to remove give sconn ptr from servers inside close_server(), so here servers[i] will change at every call to close_server().

Also since no. servers won't be too much so count_array won't be much performance issue.

I can get rid of all this by simply using link list, but as @bhroam said sched doesn't use it.

suresh-thelkar: Okay. Just another thought. How about if we do the following. If we follow this the above style of code works fine. What do you think ?
```
static void
close_server(sched_svrconn *sconn, int remove_entry)
{
	if (sconn) {
		if (remove_entry)
			remove_ptr_from_array(servers, sconn);
		if (count_array(servers) == 0) {
			free(servers);
			servers = NULL;
		}
		tpp_em_del_fd(poll_context, sconn->secondary_sock);
		pbs_disconnect(sconn->primary_sock);
		pbs_disconnect(sconn->secondary_sock);
		log_eventf(PBSEVENT_ADMIN | PBSEVENT_FORCE, PBS_EVENTCLASS_SCHED, LOG_INFO,
			   msg_daemonname, "Disconnected from server %s", sconn->svrhost);
		free(sconn->svrhost);
		free(sconn);
	}

}```

Wherever we want to close all servers i.e. close_servers() function call it as
	close_servers(servers[i], 0)
In all other places where only one server needs to be closed call it as
	close_servers(svr_entry, 1)

Comment (hirenvadalia) [reply src/scheduler/pbs_sched.c:543]: @suresh-thelkar But why to add new argument when we can do work without it? Also, no matter from where you call close_server, we have to do remove_ptr_from_array in it so I will keep current code as it.

Comment (suresh-thelkar) [reply src/scheduler/pbs_sched.c:543]: In general it is not good to alter the array which is referenced a while loop from a function which is called inside the same loop i.e. we are basically altering(removing elements from the array) the array within the same loop. Although here it does not harm because we are closing for all the servers. Secondly we can avoid calling count_array() in loop. 

This is just another way of doing it. Having said that I am fine if you want to leave the code as is.

Comment (hirenvadalia) [reply src/scheduler/pbs_sched.c:543]: Thanks @suresh-thelkar for information. I will keep code as it.

**Review by suresh-thelkar — APPROVED** [review]
Thanks for making all the code changes. I sign off. 

Please remove the WIP tag unless you have something more to test and add.

**Comment by hirenvadalia** [comment]
Rebased with master to get clean test reports...

File: src/scheduler/pbs_sched.c (hunk 56 above)
Comment (bhroam) [reply src/scheduler/pbs_sched.c:543]: Since we know close_server() will decrement the size of servers, why don't we set count_array() into a variable and decrement it.  The way you currently have it, count_array() is an O(N^2) loop.  It only needs to be O(N)

File: src/scheduler/pbs_sched.c [hunk 58]
```diff
@@ -701,7 +701,7 @@ connect_servers(void)
 		}
 	}
 
-	connect_server(pbs_default());
+	connect_server(svrhost ? svrhost : "");
```
bhroam, line 704: Since svrhost is a string, say svthost != NULL

src/scheduler/pbs_sched.c
bhroam: I'm not quite sure I understand.  Are you saying that when we read out of pbs.conf, we do the truncation of a very long host there?

src/scheduler/pbs_sched.c
```diff
@@ -786,11 +780,18 @@ read_sched_cmd(int sock)
...
-			return 1; /* return 1 as we have enqueued first command */
-
-		get_sched_cmd_noblk(sock, cmd_prio);
-		free_sched_cmd(cmd_prio);
+		if (cmd_prio != NULL) {
+			int rc_prio = get_sched_cmd_noblk(sock, cmd_prio);
+			free_sched_cmd(cmd_prio);
+			if (rc_prio == -2) {
+				free_sched_cmd(cmd);
+				return 0;
```
bhroam, line 788: This means the secondary connection has dropped.  It doesn't necessarily mean the primary has.  Could we gt into a problem where we only have one connection?

src/scheduler/pbs_sched.c
hirenvadalia: Even better I can do like:
```c
while (servers && servers[0])
   close_server(severs[0]);
```
As while closing last server, close_server will set servers = NULL. I will update code.

src/scheduler/pbs_sched.c
hirenvadalia: No, if any of connection is dropped, both (server and sched) will close other open connection, and sched will reopen both connection fresh.

src/scheduler/pbs_sched.c
hirenvadalia: Yes, see [this](https://github.com/openpbs/openpbs/blob/master/src/lib/Libifl/pbsD_connect.c#L107) for single server and for multi-server see [this](https://github.com/openpbs/openpbs/blob/master/src/lib/Libifl/pbs_loadconf.c#L295) and [this](https://github.com/openpbs/openpbs/blob/master/src/lib/Libutil/misc_utils.c#L2329)

bhroam (approved):
LGTM - When this PR is ready to merge, un-WIP It, and let me know.

hirenvadalia:
Valgrind logs: [valgrind.txt](https://github.com/openpbs/openpbs/files/5260949/valgrind.txt)

hirenvadalia:
Latest regression results:

Description: Rerun All Tests From #3839
Profile: regression
|--|--|--|--|--|--|--|
|3857|3543|23|1|0|3|3516|

Description: Rerun Only Failed Tests From #3857
|--|--|--|--|--|--|--|
|3860|24|11|0|0|0|13|

(NOTE: All failures are known test and/or product issues)

hirenvadalia:
@bhroam This PR is ready for merge.

PR #2001, opened 2020-08-26
Confirm that all execjob_end hooks execute

prakashcv13:
#### Describe Bug or Feature
PR #1867 resolved a mom crash. Changes were related to making sure that all execjob_end hooks run before the job gets deleted at the mom. However, the test case to confirm this was not automated.

#### Describe Your Change
Modified the test TestPbsExecutePrologue.test_prologue_exception_sisters (that caused the crash earlier) to add two execjob_end hooks and confirm that they are executed.

[ptl.txt](https://github.com/openpbs/openpbs/files/5132867/ptl.txt)


riyazhakki (approved):
Looks good to me

hirenvadalia (approved):

PR #2011, opened 2020-08-31
Fix multisched fairshare test

vstumpf:
#### Describe Bug or Feature
Multisched fairshare test was failing

#### Describe Your Change
Fix the multisched fairshare test.
Also, now when a scheduler's `scheduling` attribute is changed, do not send SCH_CONFIGURE.

[after-fix.txt](https://github.com/openpbs/openpbs/files/5152861/after-fix.txt)

### DIFF
--- a/src/server/req_manager.c
+++ b/src/server/req_manager.c
@@ -51,8 +51,6 @@
 #include <sys/socket.h>
 
 #include <arpa/inet.h>
-
-#include "libpbs.h"
 #include <ctype.h>
 #include <memory.h>
 #include <string.h>
@@ -60,6 +58,8 @@
 #include <unistd.h>
 #include <netdb.h>
 #include <errno.h>
+
+#include "libpbs.h"
 #include "server_limits.h"
 #include "list_link.h"
 #include "work_task.h"
@@ -89,8 +89,6 @@
 #include "sched_cmds.h"
 #include "pbs_sched.h"
 #include "pbs_share.h"
-#include "pbs_license.h"
-#include <arpa/inet.h>
 
 
 #define PERM_MANAGER (ATR_DFLAG_MGWR | ATR_DFLAG_MGRD)
@@ -1619,6 +1617,7 @@ mgr_sched_set(struct batch_request *preq)
 	pbs_list_head unsetlist;
 	int	  rc;
 	pbs_sched *psched;
+	int only_scheduling = 0;
 
 	psched = find_sched(preq->rq_ind.rq_manager.rq_objname);
 	if (!psched) {
@@ -1629,6 +1628,11 @@ mgr_sched_set(struct batch_request *preq)
 	CLEAR_HEAD(unsetlist);
 	plist = (svrattrl *)GET_NEXT(preq->rq_ind.rq_manager.rq_attr);
 	while (plist) {
+		if (strcmp(plist->al_atopl.name, ATTR_scheduling) == 0) {
+			only_scheduling++;
+		} else {
+			only_scheduling--;
+		}
 		if (plist->al_atopl.value == NULL || plist->al_atopl.value[0] == '\0') {
 			tmp = (struct svrattrl *)GET_NEXT(plist->al_link);
 			delete_link(&plist->al_link);
@@ -1662,8 +1666,8 @@ mgr_sched_set(struct batch_request *preq)
 		reply_badattr(rc, bad_attr, plist, preq);
 		return;
 	}
-
-	set_scheduler_flag(SCH_CONFIGURE, psched);
+	if (only_scheduling != 1)
+		set_scheduler_flag(SCH_CONFIGURE, psched);
 
 	sched_save_db(psched);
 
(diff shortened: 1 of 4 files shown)


src/server/req_manager.c
```diff
@@ -89,8 +89,6 @@
 #include "sched_cmds.h"
 #include "pbs_sched.h"
 #include "pbs_share.h"
-#include "pbs_license.h"
-#include <arpa/inet.h>
```
bhroam, line 93: Did you mean to remove these?

src/server/req_manager.c
```diff
@@ -1629,6 +1628,11 @@ mgr_sched_set(struct batch_request *preq)
 	CLEAR_HEAD(unsetlist);
 	plist = (svrattrl *)GET_NEXT(preq->rq_ind.rq_manager.rq_attr);
 	while (plist) {
+		if (strcmp(plist->al_atopl.name, ATTR_scheduling) == 0) {
+			only_scheduling++;
+		} else {
+			only_scheduling--;
+		}
```
bhroam, line 1635: How about calling this not_scheduling, and check strcmp() != 0, and then set the variable?  You only really care you got something that wasn't scheduling.  It's probably a bit clearer.

test/tests/functional/pbs_multi_sched.py
```diff
@@ -796,24 +796,24 @@ def test_fairshare_usage(self):
         self.server.expect(JOB, {'job_state': 'R'}, id=sc1_jid1)
         self.server.expect(JOB, {'job_state': 'Q'}, id=sc1_jid3)
         self.server.expect(JOB, {'job_state': 'Q'}, id=sc1_jid2)
-        self.server.manager(MGR_CMD_SET, SCHED, {'scheduling': 'True'},
-                            id='sc1')
         # need to delete the running job because PBS has only 1 ncpu and
         # our work is also done with the job.
         # this step will decrease the execution time as well
+        self.server.manager(MGR_CMD_SET, SCHED, {'scheduling': 'True'},
+                            id='sc1')
```
bhroam, line 803: What's the purpose of the manager() call?  Each of the submits above caused a cycle.  The delete below will cause one as well.

test/tests/functional/pbs_multi_sched.py
```diff
@@ -796,24 +796,24 @@ def test_fairshare_usage(self):
...
         self.server.delete(sc1_jid1, wait=True)
         # pbsuser3 job will run after pbsuser1
         self.server.expect(JOB, {'job_state': 'R'}, id=sc1_jid3)
         self.server.expect(JOB, {'job_state': 'Q'}, id=sc1_jid2)
+        # deleting the currently running job
```
bhroam, line 808: Why move these comments?

test/tests/functional/pbs_multi_sched.py
```diff
@@ -825,8 +825,13 @@ def test_fairshare_usage(self):
         sc1_fs_user4 = self.scheds['sc1'].query_fairshare(name=str(TEST_USER4))
         self.assertEqual(sc1_fs_user4.usage, 1)
         # Restart the scheduler
+        t = time.time()
         self.scheds['sc1'].restart()
         # Check the multisched 'sc1' usage file whether it's updating or not
+        self.assertTrue(self.scheds['sc1'].isUp())
+        self.server.manager(MGR_CMD_SET, SCHED, {'scheduling': 'True'},
+                            id='sc1')
+        self.scheds['sc1'].log_match("Scheduler is reconfiguring", starttime=t)
```
bhroam, line 834: With your above change, this won't happen any more.  We no longer send a configure if we only set scheduling.

src/server/req_manager.c
vstumpf: these are duplicates

test/tests/functional/pbs_multi_sched.py
vstumpf: We need to run a scheduling cycle to make sure that the scheduler records fairshare usage data.

bhroam: Got it, maybe add a comment to say that.  Plus, depending what your fairshare_usage_res is, the mom might not have reported any usage and you'll still have 0.

test/tests/functional/pbs_multi_sched.py
vstumpf: I was running into an issue where after the restart, the scheduler would schedule jobs, then the on next scheduling cycle it would reconfigure. I assume this is due to the scheduler coming back up from being down.

test/tests/functional/pbs_multi_sched.py
vstumpf: hahaha, i didn't mean to, it was a combination of removing the scheduling=True and putting it back in the wrong spot.
I'll change it

test/tests/functional/pbs_multi_sched.py
bhroam: I don't know why it would reconfigure.  When it comes up it just queries the sched attributes directly.  It doesn't do a configure.  The server log might help you there, or PTL is HUPing the scheduler for some reason.

test/tests/functional/pbs_multi_sched.py
vstumpf: the test uses ncpus*100, so we don't use mom values.

arungrover (approved):

bhroam (approved):
LGTM

PR #2013, opened 2020-09-01
scheduler does not update node partitions properly

arungrover:
#### Describe Bug or Feature
When a job runs, the scheduler does not update the buckets in node partitions.
TestNodeBuckets.test_psets_calendaring is failing because of this issue.

#### Describe Your Change
When a job runs, the scheduler will update buckets of node partitions of each node the job runs on.

#### Link to Design Doc
NA

Description: Rerun All Tests From #3452
Platforms: UBUNTU1804,SUSE15,CENTOS8
Profile: regression
|--|--|--|--|--|--|--|
|3471|1726|3|2|0|0|1721|

### DIFF
--- a/src/scheduler/fifo.c
+++ b/src/scheduler/fifo.c
@@ -1733,7 +1733,7 @@ run_update_resresv(status *policy, int pbs_sd, server_info *sinfo,
 						if (!ns[i]->ninfo->is_free)
 							npar[j]->free_nodes--;
 						sort_nodepart = 1;
-						update_buckets_for_node_array(npar[j]->bkts, npar[j]->ninfo_arr);
+						update_buckets_for_node(npar[j]->bkts, ns[i]->ninfo);
 					}
 				}
 				/* if the node is being provisioned, it's brought down in
--- a/src/scheduler/fifo.c
+++ b/src/scheduler/fifo.c
@@ -1727,11 +1727,13 @@ run_update_resresv(status *policy, int pbs_sd, server_info *sinfo,
 				int j;
 				update_node_on_run(ns[i], rr, &old_state);
 				if (ns[i]->ninfo->np_arr != NULL) {
-					for (j = 0; ns[i]->ninfo->np_arr[j] != NULL; j++) {
-						modify_resource_list(ns[i]->ninfo->np_arr[j]->res, ns[i]->resreq, SCHD_INCR);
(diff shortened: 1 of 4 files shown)


src/scheduler/fifo.c
```diff
@@ -1727,11 +1727,13 @@ run_update_resresv(status *policy, int pbs_sd, server_info *sinfo,
...
-					for (j = 0; ns[i]->ninfo->np_arr[j] != NULL; j++) {
-						modify_resource_list(ns[i]->ninfo->np_arr[j]->res, ns[i]->resreq, SCHD_INCR);
+					node_partition **npar = ns[i]->ninfo->np_arr;
+					for (j = 0; npar[j] != NULL; j++) {
+						modify_resource_list(npar[j]->res, ns[i]->resreq, SCHD_INCR);
 						if (!ns[i]->ninfo->is_free)
-							ns[i]->ninfo->np_arr[j]->free_nodes--;
+							npar[j]->free_nodes--;
 						sort_nodepart = 1;
+						update_buckets_for_node_array(npar[j]->bkts, npar[j]->ninfo_arr);
```
bhroam, line 1736: This loops over each node, and you are updating the node partition's buckets for an array.  This means if we will likely update the same node partition multiple times since we run a job on nodes from a single node partition.  You should rip the guts out of update_buckeys_for_node_array() and make it work for a single node.  You can call the new function in both the old function and here.

src/scheduler/node_partition.c
```diff
@@ -602,6 +602,49 @@ create_node_partitions(status *policy, node_info **nodes, char **resnames, unsig
...
+	for (i = 0; bkts[i] != NULL; i++) {
+		int node_ind = ninfo->node_ind;
+
+		/* Is this node in the bucket? */
+		if (pbs_bitmap_get_bit(bkts[i]->bkt_nodes, node_ind)) {
+			/* First turn off the current bit */
+			if (pbs_bitmap_get_bit(bkts[i]->free_pool->truth, node_ind)) {
+				pbs_bitmap_bit_off(bkts[i]->free_pool->truth, node_ind);
+				bkts[i]->free_pool->truth_ct--;
+				} else if (pbs_bitmap_get_bit(bkts[i]->busy_later_pool->truth, node_ind)) {
```
bhroam, line 623: Is the indentation off?

arungrover:
There was a null check needed in the new function, I've added that too.

PR #2014, opened 2020-09-02
Deprecate skipOnCpuset decorator

visheshh:
#### Describe Bug or Feature
Deprecate skipOnCpuset decorator from tests.

#### Describe Your Change
Deprecate skipOnCpuset decorator from tests and skip tests directly from library if tried to create more vnodes or set ncpus more than available or if tried to set mem.

Once all tests are fixed. cpuset decorator will be removed.

skipcpuset branch:

Profile: regression
|--|--|--|--|--|--|--|
|4100|2677|26|2|3|3|2643|

mainline : 
Profile: regression
|--|--|--|--|--|--|--|
|4131|2677|25|2|9|3|2638|

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -87,6 +87,11 @@ except:
         raise ImportError
     API_OK = False
 
+try:
+    from nose.plugins.skip import SkipTest
+except ImportError:
+    class SkipTest(Exception):
+        pass
 
 # suppress logging exceptions
 logging.raiseExceptions = False
@@ -6693,6 +6698,19 @@ class Server(PBSService):
         :type logerr: bool
         :raises: PbsManagerError
         """
+        if cmd == MGR_CMD_SET and id is not None and obj_type == NODE:
+            for cmom, momobj in self.moms.items():
+                cpuset_nodes = []
+                if momobj.is_cpuset_mom():
+                    momobj.check_mem_request(attrib)
+                    if len(attrib) == 0:
+                        return True
+                    vnodes = self.status(HOST, id=cmom)
+                    del vnodes[0]  # don't set anything on a naturalnode
+                    for vn in vnodes:
+                        momobj.check_ncpus_request(attrib, vn)
+                    if len(attrib) == 0:
+                        return True
 
         if isinstance(id, str):
             oid = id.split(',')
@@ -13483,6 +13501,49 @@ class MoM(PBSService):
             self._is_cpuset_mom = False
         return self._is_cpuset_mom
 
+    def skipTest(self, reason=None):
+        """
+        Skip Test
+
+        :param reason: message to indicate why test is skipped
+        :type reason: str or None
+        """
+        if reason:
+            self.logger.warning('test skipped: ' + reason)
+        else:
+            reason = 'unknown'
+        raise SkipTest(reason)
+
+    def check_mem_request(self, attrib):
+        if 'resources_available.mem' in attrib:
+            del attrib['resources_available.mem']
+            self.skipTest(
+                'mem requested cannot be set on cpuset mom')
+
+    def check_ncpus_request(self, attrib, vn):
+        skipmsg = 'ncpus requested are not same as available'
+        skipmsg += ' on a cpuset node'
+        at = 'resources_available.ncpus'
+        if at in attrib:
+            if int(attrib[at]) != int(vn[at]):
+                self.skipTest(skipmsg)
+            else:
+                del attrib['resources_available.ncpus']
+
+    def set_node_attrib(self, vnode, attrib):
+        """
+        set attribute on a node
+        """
+        for res, value in attrib.items():
+            ecmd = 'set node '
+            ecmd += vnode['id'] + ' ' + res + '=' + str(value)
+            pcmd = [os.path.join(self.pbs_conf['PBS_EXEC'],
+                                 'bin', 'qmgr'), '-c', ecmd]
+            ret = self.du.run_cmd(self.hostname, pcmd,
+                                  sudo=True,
+                                  level=logging.INFOCLI,
+                                  logerr=True)
+
     def create_vnodes(self, attrib=None, num=1,
                       additive=False, sharednode=True, restart=True,
                       delall=True, natvnode=None, usenatvnode=False,
@@ -13533,6 +13594,38 @@ class MoM(PBSService):
         if attrib is None:
             self.logger.error("attributes are required")
             return False
+
+        if self.is_cpuset_mom():
+            if vname:
+                msg = "cpuset nodes cannot have vnode names"
+                self.skipTest(msg)
+            self.check_mem_request(attrib)
+            if len(attrib) == 0:
+                return True
+            nodes = self.server.status(HOST, id=self.shortname)
+            del nodes[0]  # don't set any attribute on natural node
+            if len(nodes) < num:
+                msg = 'cpuset mom does not have required number of nodes'
+                self.skipTest(msg)
+            elif len(nodes) == num:
+                if attrib:
+                    for vnode in nodes:
+                        self.check_ncpus_request(attrib, vnode)
+                        if len(attrib) != 0:
+                            self.set_node_attrib(vnode, attrib)
+                return True
+            else:
+                i = 0
+                for vnode in nodes:
+                    if i < num:
+                        i += 1
+                        self.check_ncpus_request(attrib, vnode)
+                        if len(attrib) != 0:
+                            self.set_node_attrib(vnode, attrib)
+                    else:
+                        at = {'state': 'offline'}
+                        self.set_node_attrib(vnode, at)
+                return True
 
         if natvnode is None:
             natvnode = self.shortname
@@ -13875,6 +13968,10 @@ class MoM(PBSService):
         :param restart: If True, restart the MoM. Default is True
         :type restart: bool
         """
+        if self.is_cpuset_mom():
+            msg = 'Creating multiple vnodes is not supported on cpuset mom'
+            self.skipTest(msg)
+
         try:
             fn = self.du.create_temp_file(self.hostname, body=vdef)
         except:
--- a/test/tests/functional/pbs_Rrecord_resources_used.py
+++ b/test/tests/functional/pbs_Rrecord_resources_used.py
@@ -186,7 +186,6 @@ class Test_Rrecord_with_resources_used(TestFunctional):
 
         return jid1, jid2, jid3s1
 
-    @skipOnCpuSet
     def test_Rrecord_with_nodefailrequeue(self):
         """
         Scenario: The node on which the job was running goes down and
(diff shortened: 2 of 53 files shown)


test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -6655,6 +6660,21 @@ def selstat(self, select_list, rattrib, runas=None, extend=None):
...
+        :param reason: message to indicate why test is skipped
+        :type reason: str or None
+        """
+        if reason:
+            self.logger.warning('test skipped: ' + reason)
+        else:
+            reason = 'unknown'
+        raise SkipTest(reason)
+
+    skip_test = skipTest
```
hirenvadalia, line 6676: skip_test was just for backward compatibility, we can remove it now, I don't think any tests uses it, + if some test uses it then they will use from PBSTestSuite.skip_test not this one.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -6693,6 +6713,27 @@ def manager(self, cmd, obj_type, attrib=None, id=None, extend=None,
         :type logerr: bool
         :raises: PbsManagerError
         """
+        skipmsg = 'ncpus requested are not same as available'
```
hirenvadalia, line 6716: can we add `on cpuset mom` at end of this msg?

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -6693,6 +6713,27 @@ def manager(self, cmd, obj_type, attrib=None, id=None, extend=None,
...
+        if cmd == MGR_CMD_SET and id is not None:
+            if obj_type == NODE:
+                for cmom, momobj in self.moms.items():
+                    cpuset_nodes = []
+                    if momobj.is_cpuset_mom():
+                        vnodes = self.status(NODE)
```
hirenvadalia, line 6722: I think this should be `self.status(VNODE, id=cmom)` (aka only get vnodes of given natural node name)

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -6693,6 +6713,27 @@ def manager(self, cmd, obj_type, attrib=None, id=None, extend=None,
...
+                        del vnodes[0]  # remove natural node
```
hirenvadalia, line 6723: can you add more details why we need to remove natural node from list?

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -6693,6 +6713,27 @@ def manager(self, cmd, obj_type, attrib=None, id=None, extend=None,
...
+                        for vn in vnodes:
+                            at = 'resources_available.ncpus'
+                            if 'resources_available.ncpus' in attrib:
+                                if int(attrib[at]) != int(vn[at]):
+                                    self.skipTest(skipmsg)
+                                else:
+                                    del attrib['resources_available.ncpus']
+                        if 'resources_available.mem' in attrib:
```
hirenvadalia, line 6731: I would suggest put this check before we check ncpus, it will save one call to pbsnodes if mem is requested

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -9576,6 +9617,77 @@ def create_vnodes(self, name=None, attrib=None, num=1, mom=None,
             self.logger.error("name, attributes, and mom object are required")
             return False
 
+        if mom.is_cpuset_mom():
+            nodes = self.status(NODE)
```
hirenvadalia, line 9621: same as above, I think you need `self.status(VNODE, id=mom.shortname)` here

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -9576,6 +9617,77 @@ def create_vnodes(self, name=None, attrib=None, num=1, mom=None,
...
+            del nodes[0]  # remove natural node]
```
hirenvadalia, line 9622: remove `]` from end, and add more details about why we need to do this

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -9576,6 +9617,77 @@ def create_vnodes(self, name=None, attrib=None, num=1, mom=None,
...
+            if len(nodes) < num:
+                self.skipTest('cpuset mom')
```
hirenvadalia, line 9624: Can we add more descriptive reason? like 'Given mom is cpuset mom, and alteration of vnodes count is not allowed on it`

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -9576,6 +9617,77 @@ def create_vnodes(self, name=None, attrib=None, num=1, mom=None,
...
+            elif len(nodes) == num:
+                if attrib:
+                    skipmsg = 'ncpus requested are not same as available'
```
hirenvadalia, line 9627: Looks like this is repetitive code, can we put into common funcs?

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -9576,6 +9617,77 @@ def create_vnodes(self, name=None, attrib=None, num=1, mom=None,
...
+                            else:
+                                del attrib['resources_available.ncpus']
+                        if 'resources_available.mem' in attrib:
+                            del attrib['resources_available.mem']
+                            self.skipTest(
+                                'mem requested cannot be set on cpuset mom')
+                        if len(attrib) == 0:
+                            return True
+                        for res, value in attrib.items():
+                            ecmd = 'set node ' + \
```
hirenvadalia, line 9642: please use `+=` instead ` + \ `

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -9576,6 +9617,77 @@ def create_vnodes(self, name=None, attrib=None, num=1, mom=None,
...
+                                vnode['id'] + ' ' + res + '=' + str(value)
+                            print(ecmd)
```
hirenvadalia, line 9644: remove this debugging print

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -13868,6 +13980,9 @@ def insert_vnode_def(self, vdef, fname=None, additive=False, restart=True):
         :param restart: If True, restart the MoM. Default is True
         :type restart: bool
         """
+        if self.is_cpuset_mom():
+            self.server.skipTest('do not modify vnodes on cpuset mom')
```
hirenvadalia, line 13984: I think this msg should like: `Creating multiple mom is not supported on cpuset mom`

test/tests/functional/pbs_Rrecord_resources_used.py
```diff
@@ -186,7 +186,6 @@ def common(self, is_nonrerunnable, restart_mom):
 
         return jid1, jid2, jid3s1
 
-    @skipOnCpuSet
```
hirenvadalia, line 189: I think you need to remove `skipOnCpuSet` method from pbs_testsuite.py also

hirenvadalia (changes requested):

test/tests/functional/pbs_Rrecord_resources_used.py
visheshh: Kind of deprecating it right now. Some tests needs fix which check for specific node name or use specific node name. Listed out those tests and will work on them separately. Till then we will use skip decorator on those tests. Will remove the decorator function. Once completely removed.

hirenvadalia: ok, then please update your PR title and description with same

test/fw/ptl/lib/pbs_testlib.py
visheshh: Hi @hirenvadalia , id=cmom runs /opt/pbs/bin/pbsnodes -s <server> -v <vnodename> this can be only used to fetch one vnode on a host not all the vnodes of a cpuset host. To get that we need to run pbsnodes -H <momhost>. Should we fix that in PTL ?

hirenvadalia: Yes, please fix it in PTL (IIRC, that is long pending issue in PTL status())

visheshh: @hirenvadalia , If someone only wants to get information of one vnode on a cpusetmom, I don't know if somebody is using it. but it would break current functionality.

hirenvadalia: @visheshh I am talking about pbsnodes -s server -H momhost, aka (only) all vnodes on given mom host... if that is working fine then we don't need to fix anything in PTL.
But `vnodes = self.status(NODE)` will give you all vnodes on all moms

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -9576,6 +9623,58 @@ def create_vnodes(self, name=None, attrib=None, num=1, mom=None,
...
+                return True
+            else:
+                i = 0
+                for vnode in nodes:
+                    if i < num:
+                        i += 1
+                        self.check_ncpus_request(attrib, vnode)
+                        if len(attrib) == 0:
+                            return True
+                        for res, value in attrib.items():
```
hirenvadalia, line 9660: we can create common func for this also

hirenvadalia:
@visheshh Changes looks fine, can you please attach test log of test getting skipped and test run on cpuset and non-cpuset machine.

kjakkali:
@visheshh  : Also add link to design document.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -9576,6 +9637,37 @@ def create_vnodes(self, name=None, attrib=None, num=1, mom=None,
...
+                        self.set_node_attrib(vnode, attrib)
+                return True
+            else:
+                i = 0
+                for vnode in nodes:
+                    if i < num:
+                        i += 1
+                        self.check_ncpus_request(attrib, vnode)
+                        if len(attrib) == 0:
+                            return True
```
bayucan, line 9664: I'm trying to understand this block of code. I assume 'attrib' is something that needs to be set on all the vnodes that satisfy the # of vnodes requested. So for the extra vnodes, we offline. But here, I see a 'return True' if ncpus request is already satisfied by a vnode < the # of vnodes requested. Am I getting this confused? It doesn't look like the extra vnodes would get offlined in this case...

visheshh: Thanks for catching it. I added this code late, didnt check the implications. Will modify the condition to set the attrib if len(attrib) is not 0.

hirenvadalia:
@visheshh I believe this PR is depended on https://github.com/openpbs/openpbs/pull/2033 right? Can you please mark this PR as WIP?

visheshh:
The dependent PR is merged. I had to rebase to get the new code. I have force pushed a single commit. 
@hirenvadalia @bayucan @kjakkali , Please review

test/fw/ptl/lib/pbs_testlib.py
bayucan: @visheshh : has this been addressed?

visheshh: @bayucan : Yes Al, This has been addressed in create_vnodes. In Manager function however we should return true as we wont be having anything else to do if there are no attributes to set.

bayucan (approved):
Looks good.

kjakkali (approved):

hirenvadalia:
@visheshh changes looks good, please add regression results

hirenvadalia (approved):

PR #2015, opened 2020-09-02
node rampdown test on cpuset sister mom

vccardenas:
#### Describe Bug or Feature
A test for the Node Rampdown Feature on a cgroup cpuset mom is needed.

#### Describe Your Change
I have added two new PTL tests for the Node Rampdown Feature on a cgroup cpuset mom in pbs_cpuset.py.
The one test that was in pbs_cpuset.py was also modified to run in the new setup.

The tests ran on cluster with cgroup cpuset system as the second mom:
- first mom is a cpuset system
[TestPbsCpuset_x69_uv02.txt](https://github.com/openpbs/openpbs/files/5264618/TestPbsCpuset_x69_uv02.txt)

- first mom is not a cpuset system
[TestPbsCpuset_x83_uv02.txt](https://github.com/openpbs/openpbs/files/5264623/TestPbsCpuset_x83_uv02.txt)

The tests are skipped
- if second mom is not a cgroup cpuset system.
[TestPbsCpuset_uv02_x83.txt](https://github.com/openpbs/openpbs/files/5264626/TestPbsCpuset_uv02_x83.txt)

- if second mom is a cgroup cpuset system with less than two NUMA nodes
[TestPbsCpuset_uv02_x69.txt](https://github.com/openpbs/openpbs/files/5264628/TestPbsCpuset_uv02_x69.txt)


test/tests/functional/pbs_node_rampdown_cset.py
```diff
@@ -0,0 +1,156 @@
...
+        if len(nodeinfo) < 4:
+            self.skipTest("Not enough vnodes to run the test.")
+
+        # skip if there are no cpuset systems in the test cluster
+        no_csetmom = True
+        for mom in self.moms.values():
+            if mom.is_cpuset_mom():
+                no_csetmom = False
+        if no_csetmom:
+            self.skipTest("Skip on cluster without cgroup cpuset system.")
```
bayucan, line 68: Can move lines 62-68 before line 58 for an early check.

test/tests/functional/pbs_node_rampdown_cset.py
```diff
@@ -0,0 +1,156 @@
...
+        self.n3 = '%s[1]' % (self.n1,)
+
+        # skip if second mom has less than two NUMA vnodes
+        try:
+            self.server.expect(NODE, {'state': 'free'},
+                               id=self.n3, max_attempts=10)
+        except PtlExpectError:
+            self.skipTest("Second mom has less than 2 vnodes")
+
+        a = {'state': 'free', 'resources_available.ncpus': (GE, 1)}
```
bayucan, line 83: I think this is unused.

test/tests/functional/pbs_node_rampdown_cset.py
```diff
@@ -0,0 +1,156 @@
...
+        self.server.expect(VNODE, {'state=free': 4}, op=EQ, count=True,
+                           max_attempts=10, interval=2)
```
bayucan, line 85: Is line 58 even necessary given that there's this check.

test/tests/functional/pbs_node_rampdown_cset.py
```diff
@@ -0,0 +1,156 @@
...
+
+        self.pbs_release_nodes_cmd = os.path.join(
+            self.server.pbs_conf['PBS_EXEC'], 'bin', 'pbs_release_nodes')
+
+        # number of resource ncpus to request initially
+        ncpus = self.server.status(NODE, 'resources_available.ncpus',
+                                   id=self.n3)[0]
+        ncpus = int(ncpus['resources_available.ncpus'])
+        self.ncpus2 = ncpus / 2 - 1
```
bayucan, line 94: So if 'ncpus' is 4, initial self.ncpus2 = 4/2 -1 = 1, and if 'ncpus' is 8, initial self.ncpus2 = 8/2 - 1 = 3. So what's the significance of the '-1'?

test/tests/functional/pbs_node_rampdown_cset.py
```diff
@@ -0,0 +1,156 @@
...
+            "1:ncpus=1:mem=2gb+1:ncpus=%d:mem=2gb+" % self.ncpus2 + \
+            "1:ncpus=%d:mem=2gb" % self.ncpus2
+        self.job1_exec_host = "%s/0+%s/0*%d+%s/1*%d" % (
+            self.n0, self.n1, self.ncpus2, self.n1, self.ncpus2)
+        self.job1_exec_vnode = \
+            "(%s:ncpus=1:mem=2097152kb)+" % (self.n0,) + \
+            "(%s:ncpus=%d:mem=2097152kb)+" % (self.n2, self.ncpus2) + \
+            "(%s:ncpus=%d:mem=2097152kb)" % (self.n3, self.ncpus2)
+
+        # expected values after release of vnode
```
bayucan, line 107: expected values after release of vnode self.n3...

bayucan (changes requested):

test/tests/functional/pbs_node_rampdown_cset.py
vccardenas: line 58 checks that there are enough vnodes in the cluster. This part checks there are 4 vnodes that are in state = free.

test/tests/functional/pbs_node_rampdown_cset.py
```diff
@@ -0,0 +1,156 @@
...
+        Submit job that will use cpus on two NUMA vnodes on second mom,
+        goes in R state. Use pbs_release_node to release one of the NUMA
+        vnodes. The NUMA vnode and its resources get released. Submit a
+        second job requesting the released NUMA vnode. Job should run and
+        a new cpuset gets created that uses the released cpus.
+        """
+        # Submit a job that uses second mom's two NUMA nodes, in R state
+        a = {'Resource_List.select':
+             'ncpus=1:mem=2gb+ncpus=%d:mem=2gb+ncpus=%d:mem=2gb' %
+             (self.ncpus2, self.ncpus2), 'Resource_List.place': 'vscatter'}
```
lisa-altair, line 127: Can't this potentially end up being scheduled on the non-NUMA nodes too?  The test doesn't specifically request that vnode...maybe it should?

test/tests/functional/pbs_node_rampdown_cset.py
```diff
@@ -0,0 +1,156 @@
...
+                                 'schedselect': self.job1_schedsel1,
+                                 'exec_host': self.job1_exec_host1,
+                                 'exec_vnode': self.job1_exec_vnode1}, id=jid1)
+
+        # Submit a second job requesting the released NUMA node, in R state
+        a = {'Resource_List.select': 'ncpus=%d:mem=2gb:vnode=%s' %
+             (self.ncpus2 + 2, self.n3)}
+        j2 = Job(TEST_USER, attrs=a)
+        jid2 = self.server.submit(j2)
+        self.server.expect(JOB, {'job_state': 'R'}, id=jid2, max_attempts=10)
```
lisa-altair, line 156: Shouldn't you be checking to see that the specific vnodes are being used?
Also, don't put a max_attempts here.

lisa-altair (changes requested):

test/tests/functional/pbs_node_rampdown_cset.py
vccardenas: I have removed the -1.

test/tests/functional/pbs_node_rampdown_cset.py
```diff
@@ -0,0 +1,156 @@
...
+class TestPbsNodeRampDownCset(TestFunctional):
+
+    """
+    This tests the Node Rampdown Feature on a cluster where the
+    sister mom is a cgroup cpuset system.
+    """
+
+    def setUp(self):
+
+        TestFunctional.setUp(self)
```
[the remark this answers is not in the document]
vccardenas: TestFunctional.setUp will enable cgroups and get the cluster to a point where vnodes needed are ready for testing. Only then can we determine if the test should be skipped or not.

test/tests/functional/pbs_node_rampdown_cset.py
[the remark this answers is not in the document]
vccardenas: I have removed this part and instead loop through each vnode to check if state = free.

test/tests/functional/pbs_node_rampdown_cset.py
vccardenas: It should request a specific NUMA vnode and has been changed accordingly.

test/tests/functional/pbs_node_rampdown_cset.py
vccardenas: I have removed the submission of the second job.  I have changed the test to check that the cpuset cpu resource got freed after releasing the specific vnode requested in the initial job.  It is not guaranteed that the system will reuse the recently freed cgroup cpuset resource in the second job.

test/tests/functional/pbs_node_rampdown_cset.py
```diff
@@ -0,0 +1,175 @@
...
+            "%s/0+%s/0*%d" % (self.n0, self.n1, self.ncpus2)
+        self.job1_exec_vnode1 = \
+            "(%s:ncpus=1:mem=2097152kb)+" % (self.n0,) + \
+            "(%s:ncpus=%d:mem=2097152kb)" % (self.n2, self.ncpus2)
+
+        # cgroup cpuset path on second node
+        cmd = ['grep cgroup', '/proc/mounts', '|', 'grep cpuset', '|',
+               'grep -v', '/dev/cpuset']
+        ret = self.server.du.run_cmd(self.n1, cmd, runas=TEST_USER)
+        self.cset_path = ret['out'][0].split()[1]
```
lisa-altair, line 118: It seems to me that most of lines 86-118 are specifically for test_release_nodes_on_cpuset_sis(), and although I realize there is only one test in this suite right now...won't having these in the setUp make it harder to add test later?  What I'm trying to say is I think those lines would be more appropriate in the test.

test/tests/functional/pbs_node_rampdown_cset.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def test_release_nodes_on_cpuset_sis(self):
+        """
+        Submit job that will use cpus on two NUMA vnodes on second mom,
+        goes in R state. Use pbs_release_node to successfully release one
+        of the NUMA vnodes and its resources used in the job.
```
lisa-altair, line 124: Update the comment to be something like this:
```
Submit job that will use cpus on two NUMA vnodes on second mom.  
The wait for the job to run and then use PBS_release_node to successfully release one
of the NUMA vnodes and it's resources used in the job.
Compare the cgroup cpuset info to ensure the second NUMA node is released from the cgroup.
```

test/tests/functional/pbs_node_rampdown_cset.py
```diff
@@ -0,0 +1,175 @@
...
+        """
+        # Submit a job that uses second mom's two NUMA nodes, in R state
+        a = {'Resource_List.select': 'ncpus=1:mem=2gb+' +
+             'ncpus=%d:mem=2gb:vnode=%s+' % (self.ncpus2, self.n2) +
+             'ncpus=%d:mem=2gb:vnode=%s' % (self.ncpus2, self.n3),
+             'Resource_List.place': 'vscatter'}
```
lisa-altair, line 130: Now that you request the NUMA node vnodes specifically, is the "vscatter" still necessary because you want the first chunk to not be put on the 2 NUMA nodes?

test/tests/functional/pbs_node_rampdown_cset.py
```diff
@@ -0,0 +1,175 @@
...
+                                 'Resource_List.ncpus': 1 + self.ncpus2,
+                                 'Resource_List.nodect': 2,
+                                 'schedselect': self.job1_schedsel1,
+                                 'exec_host': self.job1_exec_host1,
+                                 'exec_vnode': self.job1_exec_vnode1}, id=jid1)
+
+        # Check if sister mom updated its internal nodes table after release
+        self.moms.values()[1].log_match('Job;%s;updated nodes info' % jid1,
+                                        starttime=before_release-1,
+                                        max_attempts=10)
```
lisa-altair, line 166: Don't set a max_attempts.

test/tests/functional/pbs_node_rampdown_cset.py
```diff
@@ -0,0 +1,175 @@
...
+
+    """
+    This tests the Node Rampdown Feature on a cluster where the
+    sister mom is a cgroup cpuset system with two NUMA nodes.
+    """
+
+    def setUp(self):
+
+        TestFunctional.setUp(self)
+        Job.dflt_attributes[ATTR_k] = 'oe'
```
lisa-altair, line 54: Why are you setting -koe?  Shouldn't pbs_release_nodes work with staging out?

lisa-altair (changes requested):
Should there be another test where the node_rampdown happens as the sister moms are not needed during stageout (i.e. using release_nodes_on_stageout)?

test/tests/functional/pbs_node_rampdown_cset.py
```diff
@@ -0,0 +1,175 @@
...
+        self.server.expect(JOB, {'job_state': 'R',
+                                 'Resource_List.mem': '6gb',
+                                 'Resource_List.ncpus': 1 + self.ncpus2 * 2,
+                                 'Resource_List.nodect': 3,
+                                 'schedselect': self.job1_schedselect,
+                                 'exec_host': self.job1_exec_host,
+                                 'exec_vnode': self.job1_exec_vnode}, id=jid1)
+
+        # Check the cpuset before releasing self.n3 from jid1
+        cset_file = self.cset_path + '/pbs_jobs.service/jobid/' + jid1 + \
```
bayucan, line 142: Use os.path.join() here when constructing the path name.

vccardenas:
Converted to a draft while I address comments.

test/tests/functional/pbs_node_rampdown_cset.py
vccardenas: Without the vscatter there is a chance that the first chunk uses one of the NUMA vnodes having sharing=default_shared. I wanted the job's exec_vnode to use the first mom for the first chunk.

test/tests/functional/pbs_node_rampdown_cset.py
vccardenas: I have removed this.

test/tests/functional/pbs_cpuset.py
```diff
@@ -38,24 +38,86 @@
...
-            self.skipTest("Tests meant to run only on cgroup cpuset system.")
         TestFunctional.setUp(self)
 
+        # skip if there are no cpuset systems in the test cluster
+        no_csetmom = True
+        for mom in self.moms.values():
+            if mom.is_cpuset_mom():
+                no_csetmom = False
+        if no_csetmom:
+            self.skipTest("Skip on cluster without cgroup cpuset system.")
```
bayucan, line 86: Some clarification here, you allow to have at least one cpuset mom to be in the cluster. So if primary mom is cpuset while sister mom is a non-cpuset, then all the tests in the cluster can run. Or should it be that you want all the mom hosts in the cluster be cpusetted?

test/tests/functional/pbs_cpuset.py
```diff
@@ -142,7 +246,7 @@ def test_reliable_job_startup_on_cpuset(self):
 
         # Submit job2 requesting the released vnode, job runs
         j2 = Job(TEST_USER, {
-            ATTR_l + '.select': '1:ncpus=1:mem=2gb:vnode=%s' % vnodeB})
+            ATTR_l + '.select': '1:ncpus=4:mem=2gb:vnode=%s' % vnodeB})
```
bayucan, line 249: The hardwired 'ncpus=4', shouldn't this be according the actual number of 'resources_available.ncpus' in vnodeB?

bayucan (changes requested):

test/tests/functional/pbs_cpuset.py
```diff
@@ -38,24 +38,86 @@
...
-    A job can be started with extra nodes with node failures tolerated
-    during job start. Released cpuset resources can be reused by another job.
+    This testsuite covers various features using cgroup cpuset systems
+        - Reliable Job Startup Feature
+        - Node Rampdown Feature
     """
 
+    def check_stageout_file_size(self):
+        """
+        This Function will check that atleast 1gb of test.img
```
bayucan, line 55: Function -> function
atleast -> at least

bayucan (changes requested):

test/tests/functional/pbs_cpuset.py
vccardenas: At this point the test is checking if there is at least one cpuset machine in the cluster. Later there will be a check if the second mom has two vnodes since the test cluster should have a cpuset system for a second mom.

bayucan: @vccardenas : if all the tests in this testsuite requires 2 cpusetted moms, best to just check the condition early.

bayucan (approved):
Looks good.

test/tests/functional/pbs_cpuset.py
```diff
@@ -140,17 +244,202 @@ def test_reliable_job_startup_on_cpuset(self):
...
+
+        # Compare the before and after cpusets info
+        msg = "%s: cpuset cpus remain after release of %s" % (jid1, self.n3)
+        self.assertNotEqual(cset_j1_before, cset_j1_after, msg)
+
+    def test_release_nodes_on_stageout_cset(self):
+        """
+        Submit a job, with -W release_nodes_on_stageout=true as a PBS directive
+        in the job script, that will use cpus and mem on two NUMA vnodes on the
+        second mom. The job goes in R state. The job creates a huge staegout
```
lisa-altair, line 325: typo "staegout" -> "stageout"

test/tests/functional/pbs_cpuset.py
```diff
@@ -140,17 +244,202 @@ def test_reliable_job_startup_on_cpuset(self):
...
+        # Check to make sure 'c' (next) record got generated
+        msg0 = ".*%s;%s.*exec_host=%s" % ('c', jid, self.job1_new_exec_host)
+        msg1 = ".*exec_vnode=%s" % self.job1_new_exec_vnode_esc
+        msg2 = ".*Resource_List\.mem=%s" % '2097152kb'
+        msg3 = ".*Resource_List\.ncpus=%d" % 1
+        msg4 = ".*Resource_List\.place=%s" % self.job1_place
+        msg5 = ".*Resource_List\.select=%s.*" % self.job1_newsel
+        msg = msg0 + msg1 + msg2 + msg3 + msg4 + msg5
+        self.server.accounting_match(msg=msg, regexp=True, n="ALL",
+                                     starttime=stime)
```
lisa-altair, line 445: I thought you were also going to show that a job can use the released vnodes of the cpuset system?  Even though the job is still being staged out?

lisa-altair (changes requested):
It's looking almost ready.  I have just a few more questions/comments

test/tests/functional/pbs_cpuset.py
```diff
@@ -80,18 +142,60 @@ def setUp(self):
...
         A job is started with two numa nodes and goes in R state.
         An execjob_launch hook will force job to have only one numa node.
         The released numa node can be used in another job.
         """
-
-        # Check that there are at least a natural vnode + 2 numa vnodes
-        nodeinfo = self.server.status(NODE)
-        if len(nodeinfo) < 3:
-            self.skipTest("Not enough vnodes to run the test.")
-
```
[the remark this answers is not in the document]
vccardenas: The test cases were written with the cluster set up so the checks are made here.

test/tests/functional/pbs_cpuset.py
```diff
@@ -101,13 +205,13 @@ def test_reliable_job_startup_on_cpuset(self):
 
         # Check mom logs that the launch hook got propagated
         msg = "Hook;launch.PY;copy hook-related file request received"
-        self.mom.log_match(msg, starttime=stime, interval=2, max_attempts=60)
+        self.moms.values()[1].log_match(msg, starttime=stime, interval=2)
```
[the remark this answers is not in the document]
vccardenas: I will remove the interval.

test/tests/functional/pbs_cpuset.py
vccardenas: The test will check after job is in stageout if self.n3 has status = free, resources_assigned.ncpus and resources_assigned.mem are 0 and that should be sufficient.

test/tests/functional/pbs_cpuset.py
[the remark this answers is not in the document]
lisa-altair: Are you saying that all three tests need 2 different systems, one system of which needs to be a cpuset system that has 2 NUMA nodes?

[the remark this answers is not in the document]
vccardenas: Yes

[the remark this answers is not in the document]
lisa-altair: Thanks for explaining.  I understand now.

test/tests/functional/pbs_cpuset.py
```diff
@@ -38,24 +38,96 @@
...
+        for node in nodeinfo:
+            if node['state'] != 'free':
+                self.skipTest("Not all the vnodes are in free state")
+
+        self.pbs_release_nodes_cmd = os.path.join(
+            self.server.pbs_conf['PBS_EXEC'], 'bin', 'pbs_release_nodes')
+        # number of resource ncpus to request initially
+        ncpus = self.server.status(NODE, 'resources_available.ncpus',
+                                   id=self.n3)[0]
+        # request a partial amount of ncpus in self.n3
```
lisa-altair, line 123: I'm afraid this comment is what I had in mind.  It is clear from the code that you are requesting only half the amount of ncpus, what I don't understand, and would like to see a comment about, is why only a partial number of ncpus instead of all of them?  Is it important to the test somehow?

lisa-altair (changes requested):
Thanks for testing on various different configurations and making the test more robust.
Just one more comment that is looking more for the "Why" explanation than the "what".  Thanks.

test/tests/functional/pbs_cpuset.py
vccardenas: There was specific reason for using half of the available ncpus on the node.

lisa-altair: Ah, okay.  Then since everything else looks good, I'll let this be.  Thanks!

lisa-altair (approved):
LGTM

bayucan (approved):
Looks good.

PR #2018, opened 2020-09-03
Race condition in receiving resources_used from sister MoM and sending to server

lisa-altair:
#### Describe Bug or Feature

The resources_used reported in the accounting logs could (infrequently) be less than actual amount of resources used.

There is a race condition that causes the main mom not to wait for the resources_used info from the sister moms before sending the job information to the server.  The race happens when the job ends on the main mom just before the reservation the job was in is ending.  
In req_signaljob, because there are no tasks to kill (because they already ended) kill_job would return a tasks killed count of 0.  The job substate would be set to JOB_SUBSTATE_EXITING which would cause main mom to ultimately proceed to sending the job's resources_used info and obit, even if not all of the sisters have reported back yet.
#### Describe Your Change

In terminate_job which is called by req_signaljob, removed setting the job substate to JOB_SUBSTATE_EXITING.
This leaves the job substate as JOB_SUBSTATE_KILLSIS until all the sisters have reported in.  Then JOB_SUBSTATE_EXITING will be set in im_request (IM_KILL_JOB).  And all of the moms' job resources_used info will be sent to the server.

Extra debug statements show the same race condition is being hit, but with the fix the outcome is the desired outcome.
[before.txt](https://github.com/openpbs/openpbs/files/5171889/before.txt)
[fixed.txt](https://github.com/openpbs/openpbs/files/5171890/fixed.txt)

### DIFF
--- a/src/resmom/requests.c
+++ b/src/resmom/requests.c
@@ -1862,7 +1862,7 @@ terminate_job(job *pjob, int internal)
 			next_sample_time = min_check_poll;
 		}
 		if (kill_job(pjob, s) == 0) {
-			/* no processes around, force into exiting */
+			/* no processes around, time to exit */
 			exiting_tasks = 1;
 		}
 		i = -2;
--- a/src/resmom/requests.c
+++ b/src/resmom/requests.c
@@ -1863,7 +1863,6 @@ terminate_job(job *pjob, int internal)
 		}
 		if (kill_job(pjob, s) == 0) {
 			/* no processes around, force into exiting */
-			pjob->ji_qs.ji_substate = JOB_SUBSTATE_EXITING;
 			exiting_tasks = 1;
 		}
 		i = -2;


src/resmom/requests.c
```diff
@@ -1863,7 +1863,6 @@ terminate_job(job *pjob, int internal)
 		}
 		if (kill_job(pjob, s) == 0) {
 			/* no processes around, force into exiting */
```
vstumpf, line 1865: Should you remove this comment? I don't think it makes much sense now

vstumpf (approved):
Just a nitpick

src/resmom/requests.c
lisa-altair: I think the comment still fits since we are setting exiting_tasks=1 which will cause the rest of the job exit to take place.

vstumpf: `exiting_tasks` just means we will eventually call `scan_for_exiting()`, right? I dunno what we're `force into exiting`

lisa-altair: updated the comment

bayucan:
@lisa-altair : have you also tested your fix under a single node job? Like is it still okay if the job has been signaled to die, and terminate_job is called, and without the substate being set to EXITING, does it still proceed normal job termination?

lisa-altair:
Yes @bayucan I ran the basic regression tests, along with specific tests for signaling a job.

bayucan:
@lisa-altair : good, thanks.

bayucan (approved):

bayucan:
@openpbs/openpbs-maintainers : I plan to take care of this PR in regards to merging. But for now, Lisa has requested to hold off merging as she might have found some issue...

vstumpf (approved):
LGTM

bhroam:
@bayucan if @lisa-altair found an issue and wants to hold off of merging, she should WIP her PR.

vstumpf:
> 
> 
> @openpbs/openpbs-maintainers : I plan to take care of this PR in regards to merging. But for now, Lisa has requested to hold off merging as she might have found some issue...

@bayucan , the easiest way to prevent merges as a maintainer is to review it again and "Request Changes"; we are trained to look at that.

bayucan:
@bhroam : ok, I'll do that next time.  In the meantime, Lisa said it turned out there's no problem with her PR, so it's all good.

lisa-altair:
False alarm, I just scared myself because the job wasn't going away when I just tried something...but it's because it couldn't do the stageout.  All good to merge once Travis is done.

PR #2020, opened 2020-09-04
Fixing test test_res_string

tpai26:
#### Describe Bug or Feature
The job sleep time is set to 5 seconds. This sometimes fails on slower machines because the job ends before the expect can check if the job is in R state. Removing the custom added sleep time for the job and keeping it to default sleep time.

[test_res_string.txt](https://github.com/openpbs/openpbs/files/5177188/test_res_string.txt)
[test_res_string_array.txt](https://github.com/openpbs/openpbs/files/5177189/test_res_string_array.txt)
[test_res_size.txt](https://github.com/openpbs/openpbs/files/5177190/test_res_size.txt)
[test_multiple_res.txt](https://github.com/openpbs/openpbs/files/5177191/test_multiple_res.txt)

### DIFF
--- a/test/tests/functional/pbs_svr_dyn_res.py
+++ b/test/tests/functional/pbs_svr_dyn_res.py
@@ -293,6 +293,8 @@ class TestServerDynRes(TestFunctional):
         a = {'job_state': 'R', 'Resource_List.foobar': 'abc'}
         self.server.expect(JOB, a, id=jid)
 
+        self.server.delete(jid, wait=True)
+
         # Submit job
         a = {'Resource_List.foobar': 'xyz'}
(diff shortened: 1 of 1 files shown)


bhroam (approved):
LGTM

vstumpf (changes requested):
I think you need to delete the job; on a machine with one cpu, the limiting resource will probably be ncpus

vstumpf:
It looks like the 5 second jobs were added in #1572, maybe you should check the other tests to see if the same need to be done?

bhroam (approved):

vstumpf (approved):
LGTM

bhroam (approved):

PR #1936, opened 2020-07-29
pbs_ralter -lselect for running reservations

bhroam:
#### Describe Bug or Feature
Enhance PBS to be able to shrink running reservations with pbs_alter -lselect

#### Describe Your Change
To be able to correctly keep the right nodes, the scheduler needs to map the unaltered resv_nodes to the unaltered schedselect.  It then maps the new select to this.  When releasing nodes, it will first release any unavailable nodes, and then go on to release other nodes.  No nodes that have running jobs on them can be released.  Once the nodes have been released, an internal select spec is generated based on the remaining nodes.  This is used to generate the new resv_nodes.

#### Link to Design Doc
 **[project documentation area](https://openpbs.atlassian.net/wiki/spaces/PD/pages/1664090170/Enhancing+pbs+ralter+for+-lselect)**

#### Attach Test and Valgrind Logs/Output
[ralter.log](https://github.com/openpbs/openpbs/files/4997141/ralter.log)
[resv.log](https://github.com/openpbs/openpbs/files/4997143/resv.log)
The 3 failures in the reservation test plan are known failures for tests that are slated to be removed.

PR #2021, opened 2020-09-04
scheduler fails to calendar a job

arungrover:
#### Describe Bug or Feature
Due to a bug in #1936 scheduler, while performing a run event scheduler tries to calendar a job, the scheduler tries to find the node solution of the job present in the run event even when that job already has a node solution associated with it.
This makes the scheduler look into events it is not supposed to and fails perform_event call, eventually resulting in failure to add a job to the calendar.

#### Describe Your Change
in run_update_resresv scheduler checks for orig_nspec_arr on the job and if not found it tries to find a node solution for the job by calling check_nodes. Orig_nspec_arr is only set for reservations and not for jobs. 
This change moves orig_nspec_arr inside the resv_info structure and also checks for orig_nspec_arr only when the scheduler is trying to run reservations.

[test_reservation.txt](https://github.com/openpbs/openpbs/files/5177103/test_reservation.txt)
[pbs_ralter.txt](https://github.com/openpbs/openpbs/files/5177104/pbs_ralter.txt)
Description: Rerun All Tests From #3614
Profile: regression
|--|--|--|--|--|--|--|
|3621|3478|4|0|0|2|3472|

### DIFF
--- a/src/scheduler/fifo.c
+++ b/src/scheduler/fifo.c
@@ -1680,8 +1680,13 @@ run_update_resresv(status *policy, int pbs_sd, server_info *sinfo,
 		 */
 		rr->can_not_run = 1;
 
-		if (rr->nspec_arr != NULL && rr->nspec_arr != ns && rr->nspec_arr != ns_arr)
+		if (rr->nspec_arr != NULL && rr->nspec_arr != ns && rr->nspec_arr != ns_arr && rr->nspec_arr != orig_ns)
 			free_nspecs(rr->nspec_arr);
+
+		if (orig_ns != ns_arr) {
+			free_nspecs(ns_arr);
+			ns_arr= NULL;
+		}
 		/* The nspec array coming out of the node selection code could
 		 * have a node appear multiple times.  This is how we need to
 		 * send the execvnode to the server.  We now need to combine
@@ -1692,6 +1697,9 @@ run_update_resresv(status *policy, int pbs_sd, server_info *sinfo,
 
 		if (rr->resv != NULL)
 			rr->resv->orig_nspec_arr = orig_ns;
+		else
+			free_nspecs(orig_ns);
+
 		rr->nspec_arr = ns;
 
 		if (rr->is_job && !(flags & RURR_NOPRINT)) {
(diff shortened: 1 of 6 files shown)


src/scheduler/fifo.c
```diff
@@ -1676,16 +1680,26 @@ run_update_resresv(status *policy, int pbs_sd, server_info *sinfo,
 		 */
 		rr->can_not_run = 1;
 
-		if (rr->nspec_arr != NULL && rr->nspec_arr != ns && rr->nspec_arr != ns_arr)
+		if (rr->nspec_arr != NULL && rr->nspec_arr != ns && rr->nspec_arr != ns_arr && rr->nspec_arr != orig_ns)
 			free_nspecs(rr->nspec_arr);
+
+		if (orig_ns != ns_arr) {
+			free_nspecs(ns_arr);
+			ns_arr= NULL;
```
bhroam, line 1688: Add a space before the =

src/scheduler/job_info.c
```diff
@@ -1314,9 +1313,15 @@ query_job(struct batch_status *job, server_info *sinfo, schd_error *err)
...
-			execvnode = attrp->value;
-		else if (!strcmp(attrp->name, ATTR_l)) { /* resources requested*/
+		else if (!strcmp(attrp->name, ATTR_execvnode)) {
+			nspec **tmp_nspec_arr;
+			tmp_nspec_arr = parse_execvnode(attrp->value, sinfo, resresv->select);
+			resresv->nspec_arr = combine_nspec_array(tmp_nspec_arr);
+			free_nspecs(tmp_nspec_arr);
+
+			if (resresv->nspec_arr != NULL)
+				resresv->ninfo_arr = create_node_array_from_nspec(resresv->nspec_arr);
```
bhroam, line 1323: The reason this was at the bottom was the fact that resresv->select might not be queried at this point.  In any case, you probably don't need to worry about that because you don't need to map nspecs to chunks.  The chunk mappings are discarded in combine_nspec_array().  Just pass NULL instead of resresv->select.

bhroam (changes requested):

bhroam (approved):
LGTM

vstumpf:
do you need to copy the new `orig_nspec_arr` in `dup_resv_info`? You added it to new and free

arungrover:
@vstumpf it is duped in dup_resource_resv because I can't access the select spec in dup_resv_info.

vstumpf (approved):
LGTM

PR #2025, opened 2020-09-11
Adding configure option to disable security checks

agrawalravi90:
#### Describe Bug or Feature

With the recent change to how security is verified in scheduler, I was having trouble using the sched library directly for third party tools where such security checks don't apply. Since there was no configurable option to disable these checks, this PR just adds that option. By default, they will be enabled. Passing --disable-security-check to configure disables them.

#### Describe Your Change

Added a macro file for disabling security checks

No external change here, so if existing tests pass, it'll test the changes sufficiently.


subhasisb:
While i have no issues with this switch, i think the sched library should not be doing security checks in the first place. The security checks should be in the daemons using the library, right?

agrawalravi90:
> The security checks should be in the daemons using the library, right?

True, and that was my first thought as well, but sched today does validation checks when someone sets the sched_priv/sched_log attribute, and I'm not sure if there's a clean way to pull it out of the main sched library. Now that I think about it, I'm not sure why sched needs to do this check when it can be easily done at the server side. @bhroam what do you think?

bhroam:
That's an interesting idea.  Originally we had to be careful with security because we listened on a socket.  Anyone could connect to us as long as they did so over a privileged port.  WIth @hirenvadalia's PR reversing the direction of the connection, we know who we are connecting to.  We probably can trust the other side more now.  It is also less of a problem with the scheduler not running as root.

Question, don't we also call chk_file_sec() with server_dyn_res?  Won't that also cause a problem?

agrawalravi90:
> Question, don't we also call chk_file_sec() with server_dyn_res? Won't that also cause a problem?

I don't think so, I don't see any other call to it in the scheduler's core.

arungrover:
Scheduler calls tmp_file_sec_user at - https://github.com/openpbs/openpbs/blob/master/src/scheduler/server_info.c#L667

agrawalravi90:
Thanks for catching that Arun. I think that should also be surrounded by #ifndef NO_SECURITY_CHECK.

agrawalravi90:
@bhroam  and @arungrover  will we still need to do this security check after Hiren's changes?

bhroam:
Yes we do.  We aren't querying this data from the server.  It comes from sched_config.  The script needs to have a security check.

agrawalravi90:
@bhroam and @arungrover  thanks for clarifying. I guess then we cannot pull the security checks from the scheduler core right now, and the only thing that we can do is surround the call to tmp_file_sec_user also with #ifndef NO_SECURITY_CHECK and use this PR's switch when the security checks are not desired. Please let me know if you guys disagree.

agrawalravi90:
@subhasisb and @bhroam gentle reminder, thanks.

src/scheduler/server_info.c
```diff
@@ -665,14 +664,19 @@ query_server_dyn_res(server_info *sinfo)
...
-			if (err != 0) {
-				log_eventf(PBSEVENT_SECURITY, PBS_EVENTCLASS_SERVER, LOG_ERR, "server_dyn_res",
-					"error: %s file has a non-secure file access, setting resource %s to 0, errno: %d",
-					filename, res->name, err);
-				set_resource(res, res_zero, RF_AVAIL);
-				continue;
-			}
+			#if !defined(DEBUG) && !defined(NO_SECURITY_CHECK)
+				int err;
+				char *filename = conf.dynamic_res[i].script_name;
```
bhroam, line 669: You're defining variables not at the top of a block.  This will fail on earlier versions of C.  We generally stick with c89, which doesn't support this.  Put these back at the top and put a #if around them.

bhroam (changes requested):
What about the code around sched_priv and sched_logs?  I thought they called the security checks as well?

src/scheduler/server_info.c
agrawalravi90: really? we use things like \_\_func\__ and snprintf() which are C99 right?

agrawalravi90:
"What about the code around sched_priv and sched_logs? I thought they called the security checks as well?"
are there any checks outside of parse_sched_obj as well? The ones in there are already surrounded with #if !defined(NO_SECURITY_CHECK), please let me know if there are any that i missed.

src/scheduler/server_info.c
bhroam: I believe we have had compilers in the past which were partially c99 compliant which allowed us to use that.  Until we're given the go-ahead to use c99, I'd shy away from it.  Unless @subhasisb disagrees.

agrawalravi90: ok, that's fine, i was just hesitant to add an ugly #ifdef just to declare variables at the top of the block, I'll make the change.

hirenvadalia: @agrawalravi90 you can do something like below (which is also ugly but will save u from another #ifdef):
```c
#if !defined(DEBUG) && !defined(NO_SECURITY_CHECK)
{
    int err;
    char *filename = conf.dynamic_res[i].script_name;
    ...
}
#endif
```

agrawalravi90:
@bhroam  and @subhasisb gentle reminder, thanks

src/scheduler/parse.c
```diff
@@ -594,13 +594,15 @@ parse_config(char *fname)
...
 									error = 1;
 								else {
-									int err;
-									err = tmp_file_sec_user(filename, 0, 1, S_IWGRP|S_IWOTH, 1, getuid());
-									if (err != 0) {
-										snprintf(errbuf, sizeof(errbuf),
-											"error: %s file has a non-secure file access, errno: %d", filename, err);
-										error = 1;
-									}
+									#if !defined(DEBUG) && !defined(NO_SECURITY_CHECK)
```
hirenvadalia, line 597: nit: I think #if should start from starting of line, also block inside is not aligned wrt outside of #if block (same for other file too)

subhasisb (approved):

bhroam (approved):
LGTM

PR #2027, opened 2020-09-12
Multi-sched test fails on slow machines

arungrover:
#### Describe Bug or Feature
Multi-sched test submits a reservation and then waits for it to be confirmed. This test then offlines the vnode where the reservation was confirmed and then checks that the reservation gets reconfirmed on another vnode.
This test fails on slow machines because the reservation is submitted to start 20 seconds in the future and by the time the test creates all schedulers, offlines the node, and checks the reservation's node and status, the reservation moves into running state.

#### Describe Your Change
This change now submits a reservation that will start 40 seconds in the future. This gives the test enough time on slower machines to offline a vnode, and create schedulers. This change should not delay the test on faster machines because the test does not wait for the reservation to start.

#### Link to Design Doc
NA

[test_multisched.txt](https://github.com/openpbs/openpbs/files/5212091/test_multisched.txt)

### DIFF
--- a/test/tests/functional/pbs_multi_sched.py
+++ b/test/tests/functional/pbs_multi_sched.py
@@ -2152,7 +2152,7 @@ e.accept()
         """
         self.common_setup()
         now = int(time.time())
-        self.degraded_resv_reconfirm(start=now + 20, end=now + 200)
+        self.degraded_resv_reconfirm(start=now + 40, end=now + 200)
 
     def test_advance_running_resv_reconfirm(self):
         """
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_multi_sched.py
```diff
@@ -2170,7 +2170,7 @@ def test_standing_confimred_resv_reconfirm(self):
         """
         self.common_setup()
         now = int(time.time())
-        self.degraded_resv_reconfirm(start=now + 20, end=now + 200,
```
bhroam, line 2173: Since the test does not wait for the reservation to start, why not use now + 600 and now + 800 just to be safe?

bhroam (changes requested):

bhroam (approved):
LGTM

vstumpf (approved):
LGTM

PR #2029, opened 2020-09-14
configurable mailer

vchlum:
#### Describe Bug or Feature
This change adds the possibility to change the sendmail to another mailer (like sendmail wrapper) by a server attribute.

#### Describe Your Change
A new server attribute 'mailer' has been added. The value of the attribute is by default the original path to sendmail. If the attribute changes the new value is used as mailer instead of sendmail.

Also, an example of sendmail wrapper called 'pbs-mailer' is added into the 'unsupported' directory.

#### Link to Design Doc
* [Design document here](https://openpbs.atlassian.net/wiki/spaces/PD/pages/2032402433/Configurable+mailer).
* [Discussion here](https://community.openpbs.org/t/configurable-mailer/2242).

Test that emailing is not broken:
[ptl_pbs_array_job_mail.log](https://github.com/openpbs/openpbs/files/5217253/ptl_pbs_array_job_mail.log)


subhasisb:
hi @vchlum this looks very nice. Would it be too much to request if you could add the mailer script for non-debian packaging as well?

vchlum:
Sure @subhasisb, which packaging(s) would you like to see? There is not only deb-packaging but also RPM already.

vchlum:
Oh, sorry. I didn't notice the *.spec files are in .gitignore. I have force-added the spec file.

subhasisb:
Thanks @vchlum sorry i missed that there was already a RPM also, so i guess we are all set.

src/server/svr_mail.c
```diff
@@ -180,6 +181,9 @@ svr_mailowner_id(char *jid, job *pjob, int mailpoint, int force, char *text)
 	/* Unprotect child from being killed by kernel */
 	daemon_protect(0, PBS_DAEMON_PROTECT_OFF);
 
+	if ((mailer = server.sv_attr[(int)SVR_ATR_mailer].at_val.at_str) == 0)
```
subhasisb, line 184: Perhaps you want to use the new function is_attr_set() along with checking the value is NULL (and using the NULL word explicitly instead of 0 is nicer to read)

src/server/svr_mail.c
```diff
@@ -255,7 +259,7 @@ svr_mailowner_id(char *jid, job *pjob, int mailpoint, int force, char *text)
 
 	/* setup sendmail command line with -f from_whom */
 
-	margs[0] = SENDMAIL_CMD;
+	margs[0] = mailer;
```
subhasisb, line 262: This block of code to setup the sendmail execution seems exactly same between here and in the reservation case. This is not your code change, something older. Care to make this a common function for better modularity?

subhasisb (changes requested):

src/server/svr_mail.c
vchlum: I used is_attr_set() and tested for true. I did the same for _SVR_ATR_mailfrom_.

src/server/svr_mail.c
vchlum: OK, please see the suggested changes. Not sure how large block of code do you mean... I added svr_exec_mailer() function not only for setup but for also for execv. Redundancy removed.

subhasisb (approved):
Thanks @vchlum this is exactly what i meant

src/unsupported/pbs-mailer/README.md
```diff
@@ -0,0 +1,30 @@
+### pbs-mailer
+
+OpenPBS can easily send a huge amount of emails to one user/email. pbs-mailer is a service for an aggregation of emails send by OpenPBS to the same user/email. The first email is send immediately and subsequent emails are sent after configurable time. The emails are squashed into one email.
```
prakashcv13, line 3: emails "sent" by OpenPBS
first email is "sent" immediately
after a configurable time "period".

prakashcv13 (changes requested):
only cosmetic comments

vchlum:
> only cosmetic comments

@prakashcv13 Thank you for your comments. Please, see the latest changes.

prakashcv13 (approved):

prakashcv13:
@vchlum - Thank you once again for your contribution!

PR #2030, opened 2020-09-15
Separating out job attr get/set functions

agrawalravi90:
#### Describe Bug or Feature

This is a follow-up to https://github.com/openpbs/openpbs/pull/2016 where I refactored some job object related code. I decided to separate out the job getters and setters into a dedicated file and fixed some minor issues.


SudeshnaMoulik (commented):
Thanks @agrawalravi90 for the clear separation of the code. Can you please attach the regression result too.

subhasisb (approved):

bhroam (approved):
LGTM

bhroam:
@agrawalravi90 once you add the regression test results that @SudeshnaMoulik asked for, I'll merge this.

agrawalravi90:
Regression result:
Screen Shot 2020-09-16 at 5 07 31 PM

I investigated the failed tests and only one of them seems like a new failure, it's related to run_count. I've confirmed that this test also fails without this change. It might have been caused by my attr get/set changes, so I'm still investigating it, but I'd like to fix it in a separate PR. So, hopefully we can go ahead with this, please let me know. Thanks!

agrawalravi90:
Actually i tried running that test on a version before the attr get/set changes and it still failed, so it seems independent of those changes as well.

SudeshnaMoulik (approved):
Thanks @agrawalravi90 for the confirmation.

PR #2033, opened 2020-09-17
Move PTL create_vnodes method from class Server to Mom

visheshh:
#### Describe Bug or Feature
Move PTL create_vnodes method from class Server to Mom

#### Describe Your Change
1.Moved PTL create_vnodes method from class Server to Mom
2. Removed the mom object and the vnode_prefix name sent to create_vnodes function and added back an optional vnode_prefix which can be given under necessary conditions.
3.Make mom hostname as the default vnode_prefix and removed the vnode_prefix from tests if not absolutely necessary.
4. Modified TestOfflineVnode test suite to verify vnodes from both the moms.

#### Link to Design Doc

https://openpbs.atlassian.net/wiki/spaces/PD/pages/2083028993/Move+create+vnodes+from+Class+Server+to+Mom+in+PTL

Mainline :
Description: Tests from given sources on platforms UBUNTU1804, SUSE15, CENTOS8, CENTOS7
Platforms: UBUNTU1804,SUSE15,CENTOS8,CENTOS7
Profile: regression
|--|--|--|--|--|--|--|
|3759|968|3|2|1|1|961|

create_vnodes branch:
Description: Tests from given sources on platforms UBUNTU1804, SUSE15, CENTOS8, CENTOS7
Platforms: UBUNTU1804,SUSE15,CENTOS8,CENTOS7
Profile: regression
|--|--|--|--|--|--|--|
|3758|968|3|1|1|1|962|

### DIFF
--- a/test/tests/functional/pbs_test_tpp.py
+++ b/test/tests/functional/pbs_test_tpp.py
@@ -358,8 +358,7 @@ class TestTPP(TestFunctional):
         if not self.mom.is_cpuset_mom():
             a = {'resources_available.ncpus': 1}
             self.mom.create_vnodes(a, 2)
-            vnode_val = "vnode=" + vn + \
-                "[0]:ncpus=1+vnode=" + vn + "[1]:ncpus=1"
+            vnode_val = "vnode=%s[0]:ncpus=1+vnode=%s[1]:ncpus=1" % (vn, vn)
         else:
             vnode_val = "vnode=%s:ncpus=1" % self.server.status(NODE)[1]['id']
             vnode_val += "+vnode=%s:ncpus=1" % self.server.status(NODE)[
@@ -373,8 +372,8 @@ class TestTPP(TestFunctional):
                           resv_set_attr=resv_set_attr)
         self.comm.stop('-KILL')
         if self.mom.is_cpuset_mom():
-            vnode_list = [self.server.status(NODE)[1]['id'],
-                          self.server.status(NODE)[2]['id']]
+            ret = self.server.status(NODE)
+            vnode_list = [ret[i]['id'] for i in range(1, 3)]
         else:
             vnode_list = [vn + "[0]", vn + "[1]"]
         a = {'state': (MATCH_RE, "down")}
--- a/test/tests/functional/pbs_test_tpp.py
+++ b/test/tests/functional/pbs_test_tpp.py
@@ -354,10 +354,12 @@ class TestTPP(TestFunctional):
         """
         Test for verifying vnode insertion when using TPP
         """
+        vn = self.mom.shortname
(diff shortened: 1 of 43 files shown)


test/tests/functional/pbs_test_tpp.py
```diff
@@ -354,10 +354,12 @@ def test_comm_with_vnode_insertion(self):
...
         Test for verifying vnode insertion when using TPP
         """
+        vn = self.mom.shortname
         if not self.mom.is_cpuset_mom():
             a = {'resources_available.ncpus': 1}
-            self.server.create_vnodes('vn', a, 2, self.mom)
-            vnode_val = "vnode=vn[0]:ncpus=1+vnode=vn[1]:ncpus=1"
+            self.mom.create_vnodes(a, 2)
+            vnode_val = "vnode=" + vn + \
+                "[0]:ncpus=1+vnode=" + vn + "[1]:ncpus=1"
```
bhroam, line 362: Consider using string formatting.  If you do that, you might be able to put the string on the first line and the (vn, vn) on the second.

test/tests/functional/pbs_test_tpp.py
```diff
@@ -374,7 +376,7 @@ def test_comm_with_vnode_insertion(self):
             vnode_list = [self.server.status(NODE)[1]['id'],
                           self.server.status(NODE)[2]['id']]
```
bhroam, line 377: Only call status() once, and and call ret[0]['id'] and ret[1]['id]], or use list comprehension and say [ret[i]['id'] for i in range(2)] (here and above)

bhroam (changes requested):

agrawalravi90 (dismissed):
"Move PTL create_vnodes method from class Server to Mom"

Can you please explain the motivation to do this?

visheshh:
> 
> 
> "Move PTL create_vnodes method from class Server to Mom"
> 
> Can you please explain the motivation to do this?

Hi @agrawalravi90 ,
Motivation to do this was to make vnode prefixes same across all platforms such as cpuset moms.
We are removing skipcpuset decorator and trying to return available vnodes in the system which have mom name as prefix and tests use vn or vnode as prefix, so we removed them and in the process moved create_vnodes function which has nothing do with server object and instead of calling from server object and passing mom object as argument we moved it under class mom.

visheshh:
Thanks @bhroam for taking your time and doing a quick review for all of these tests.

visheshh:
Hi @bhroam, I have addressed all comments. Please review.

bhroam (changes requested):

visheshh:
@bhroam Please review

bhroam (approved):
LGTM

arungrover (approved):
LGTM

PR #2034, opened 2020-09-17
PP-1333: Subjobs are sometimes aborted on server restart

lisa-altair:
#### Describe Bug or Feature

Subjobs are sometimes aborted on a server restart.
This is because the ji_qrank attribute is defined as a INTEGER in the postgres database.  But JOB_ATR_qrank job attribute is a long type.
Storing this attribute to the database results in misleading qrank. Once the server is restarted, the order of recovered jobs is sorted by the qrank. The inconsistency in data type may lead to reading subjob before parentjob, which leads to aborting the subjob because no parent job is read yet.

#### Describe Your Change

Changed the ji_qrank in the database from an INTEGER to a BIGINT.

Testing was done manually by changing the system time and starting subjobs at different times so they would get different qrank values.
[subjobs_before.txt](https://github.com/openpbs/openpbs/files/5242304/subjobs_before.txt)
[subjobs_after.txt](https://github.com/openpbs/openpbs/files/5242305/subjobs_after.txt)

### DIFF
--- a/src/include/pbs_db.h
+++ b/src/include/pbs_db.h
@@ -180,7 +180,7 @@ struct pbs_db_job_info {
 	BIGINT   ji_fromaddr;	/* host job coming from   */
 	char     ji_jid[8];	/* extended job save data */
 	INTEGER  ji_credtype;	/* credential type */
-	INTEGER  ji_qrank;	/* sort key for db query */
+	BIGINT   ji_qrank;	/* sort key for db query */
 	pbs_db_attr_list_t db_attr_list; /* list of attributes for database */
 };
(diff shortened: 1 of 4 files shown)


lisa-altair:
Description: Tests from given sources on platforms SUSE15, CENTOS8, CENTOS7, UBUNTU1804
Platforms: SUSE15,CENTOS8,CENTOS7,UBUNTU1804
Profile: regression
|--|--|--|--|--|--|--|
|3705|968|2|1|0|1|964|

bhroam (approved):
LGTM

vstumpf (approved):
LGTM, I like the tests :D

PR #2032, opened 2020-09-17
Introducing multi-server switch in pbs.conf

bremanandjk:
#### Describe Bug or Feature

Introducing multi-server switch in pbs.conf

#### Describe Your Change

Added the necessary code to parse this new switch.

#### Link to Design Doc

https://openpbs.atlassian.net/wiki/spaces/PD/pages/2099249173/Switches+related+to+multi-server

#### Attach Test and Valgrind Logs/Output

Description: Tests from given sources on platforms UBUNTU1804, SUSE15, UBUNTU1604, CENTOS8, RHEL8, CENTOS7, RHEL7, SLES12, SLES15, WIN2K16
Platforms: UBUNTU1804,SUSE15,UBUNTU1604,CENTOS8,RHEL8,CENTOS7,RHEL7,SLES12,SLES15,WIN2K16
Profile: regression
|ID|TOTAL|FAIL|ERROR|TIMEDOUT|SKIP|PASS|
|--|--|--|--|--|--|--|
|3744|3550|7|1|0|3|3539|

Description: Rerun Only Failed Tests From #3744
Platforms: UBUNTU1804,SUSE15,UBUNTU1604,CENTOS8,RHEL8,CENTOS7,RHEL7,SLES12,SLES15,WIN2K16
Profile: regression
|ID|TOTAL|FAIL|ERROR|TIMEDOUT|SKIP|PASS|
|--|--|--|--|--|--|--|
|3747|8|5|1|0|0|2|

Those failed tests are already failing in the mainline.

PR #2036, opened 2020-09-18
 pbs_connect changes for multi-server support

agrawalravi90:
#### Describe Bug or Feature

This changes pbs_connect (and disconnect) to add support for the multi-server feature. It will allow client commands to connect to one or more servers, when configured via the PBS_SERVER_INSTANCES config option (#2032 )

#### Link to Design Doc

This doesn't really add any external design change. For info on multi-server, refer to this: https://openpbs.atlassian.net/wiki/spaces/PD/pages/1507164161/PBS+Pro+multi-server+project+architecture+and+design

For single server, existing tests passing should be good enough.

I did manually test that when there are multiple servers running, qstat was able to connect to all of them. We still need to add the necessary IFL changes before qstat can actually collate the output from multiple servers and show it to the user. But here's a gdb log which shows how it was able to connect to 2 servers:
```
[ravi@sim ~]$ ps -ef | grep pbs_server
root     40377     1  0 22:02 ?        00:00:00 /opt/pbs/sbin/pbs_server.bin
root     41578     1  0 22:03 ?        00:00:00 /opt/pbs/sbin/pbs_server.bin
ravi     41585 97816  0 22:03 pts/1    00:00:00 grep --color=auto pbs_server
[ravi@sim ~]$ 
[ravi@sim ~]$ 
[ravi@sim ~]$ 
[ravi@sim ~]$ 
[ravi@sim ~]$ 
[ravi@sim ~]$ gdb qstat
GNU gdb (GDB) Red Hat Enterprise Linux 7.6.1-119.el7
Copyright (C) 2013 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-redhat-linux-gnu".
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>...
Reading symbols from /opt/pbs/bin/qstat...Reading symbols from /usr/lib/debug/opt/pbs/bin/qstat.debug...done.
done.
(gdb) b __pbs_connect_extend
Breakpoint 1 at 0x41cb4c: file ../../../../src/lib/Libifl/pbsD_connect.c, line 515.
(gdb) run
Starting program: /opt/pbs/bin/qstat 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib64/libthread_db.so.1".

Breakpoint 1, __pbs_connect_extend (server=0x7fffffffdfc0 "", extend_data=0x0) at ../../../../src/lib/Libifl/pbsD_connect.c:515
515		if (pbs_client_thread_init_thread_context() != 0)
Missing separate debuginfos, use: debuginfo-install glibc-2.17-307.el7.1.x86_64 nss-softokn-freebl-3.44.0-8.el7_7.x86_64
(gdb) n
518		if (pbs_loadconf(0) == 0)
(gdb) 
521		server = PBS_get_server(server, server_name, &server_port);
(gdb) 
522		if (server == NULL) {
(gdb) 
527		if ((sock = connect_to_servers(server_name, server_port, extend_data)) == -1) {
(gdb) s
connect_to_servers (server_name=0x7fffffffd790 "sim", port=15001, extend_data=0x0) at ../../../../src/lib/Libifl/pbsD_connect.c:450
450		int fd = -1;
(gdb) n
451		int ret_fd = -1;
(gdb) 
452		int multi_flag = 0;
(gdb) 
453		int num_conf_servers = get_num_servers();
(gdb) 
454		svr_conn_t *svr_connections = get_conn_servers();
(gdb) 
456		if (svr_connections == NULL)
(gdb) 
459		multi_flag = getenv(CONN_MULTI) != NULL;
(gdb) 
460		if (!multi_flag) {	/* Client wants to connect to only 1 server */
(gdb) 
482		for (i = 0; i < num_conf_servers; i++) {
(gdb) 
483			fd = connect_to_server(i, svr_connections, extend_data);
(gdb) 
[Detaching after fork from child process 41594]
485			if (ret_fd == -1 && fd != -1) {
(gdb) p fd
$1 = 7
(gdb) n
486				ret_fd = fd;
(gdb) 
487				if (!multi_flag)
(gdb) 
482		for (i = 0; i < num_conf_servers; i++) {
(gdb) 
483			fd = connect_to_server(i, svr_connections, extend_data);
(gdb) 
[Detaching after fork from child process 41595]
485			if (ret_fd == -1 && fd != -1) {
(gdb) p fd
$2 = 8
(gdb) n
482		for (i = 0; i < num_conf_servers; i++) {
(gdb) 
492		return ret_fd;
(gdb) 
493	}
(gdb) c
Continuing.
[Inferior 1 (process 41589) exited normally]
(gdb) q
```

### DIFF
--- a/src/cmds/pbsnodes.c
+++ b/src/cmds/pbsnodes.c
@@ -895,28 +895,28 @@ marknode(int con, char *name,
 int
 main(int argc, char *argv[])
 {
-	time_t		     timenow;
-	struct attrl	    *pattr = NULL;
-	int		     con;
-	char		    *def_server;
-	int		     errflg = 0;
-	char		    *errmsg;
-	int		     i;
-	int		     rc = 0;
-	extern	char	    *optarg;
-	extern	int	     optind;
-	char		   **pa;
-	char		    *comment = NULL;
+	time_t timenow;
+	struct attrl *pattr = NULL;
+	int con;
+	char *def_server = NULL;
+	int errflg = 0;
+	char *errmsg;
+	int i;
+	int rc = 0;
+	extern char *optarg;
+	extern int optind;
+	char **pa;
+	char *comment = NULL;
 	struct batch_status *bstat = NULL;
 	struct batch_status *bstat_head = NULL;
 	struct batch_status *next_bstat = NULL;
-	int		     do_vnodes = 0;
-	mgr_operation_t      oper = LISTSP;
-	int		     ret = 0;
-	int		     job_summary  = 0;
-	int		     long_summary = 0;
-	int		     format = 0;
-	int		     prt_summary = 0;
+	int do_vnodes = 0;
+	mgr_operation_t oper = LISTSP;
+	int ret = 0;
+	int job_summary = 0;
+	int long_summary = 0;
+	int format = 0;
+	int prt_summary = 0;
 
 	/*test for real deal or just version and exit*/
 
@@ -924,12 +924,6 @@ main(int argc, char *argv[])
 
 	if (initsocketlib())
 		return 1;
-
-	/* get default server, may be changed by -s option */
-
-	def_server = pbs_default();
-	if (def_server == NULL)
-		def_server = "";
 
 	if (argc == 1)
 		errflg = 1;
@@ -1075,6 +1069,12 @@ main(int argc, char *argv[])
 				"\t%s --version\n\n",
 				argv[0], argv[0], argv[0], argv[0], argv[0], argv[0]);
 		exit(1);
+	}
+
+	if (def_server == NULL) {
+		def_server = pbs_default();
+		if (def_server == NULL)
+			def_server = "";
 	}
 
 	if (CS_client_init() != CS_SUCCESS) {
--- a/src/cmds/qmgr.c
+++ b/src/cmds/qmgr.c
@@ -110,6 +110,7 @@
 #define QMGR_TIMEOUT 900 /* qmgr connection timeout set to 15 min */
 time_t start_time = 0;
 time_t check_time = 0;
+int num_connected_svrs = 0;
 
 char prompt[]="Qmgr: "; /* Prompt if input is from terminal */
 char contin[]="Qmgr< "; /* Prompt if input is continued across lines */
--- a/src/cmds/qmgr.c
+++ b/src/cmds/qmgr.c
@@ -110,7 +110,6 @@
 #define QMGR_TIMEOUT 900 /* qmgr connection timeout set to 15 min */
 time_t start_time = 0;
 time_t check_time = 0;
-int num_connected_svrs = 0;
 
 char prompt[]="Qmgr: "; /* Prompt if input is from terminal */
 char contin[]="Qmgr< "; /* Prompt if input is continued across lines */
--- a/src/include/libpbs.h
+++ b/src/include/libpbs.h
@@ -110,6 +110,7 @@ extern int *__pbs_tcperrno_location(void);
 #endif
 
 extern char pbs_current_group[];
+extern pthread_key_t psi_key;
 
 #define NCONNECTS 50 /* max connections per client */
 #define PBS_MAX_CONNECTIONS 5000 /* Max connections in the connections array */
@@ -131,8 +132,23 @@ pbs_tcp_chan_t * get_conn_chan(int);
 int set_conn_chan(int, pbs_tcp_chan_t *);
 pthread_mutex_t * get_conn_mutex(int);
 
+#define SVR_CONN_STATE_DOWN 0
+#define SVR_CONN_STATE_UP 1
+
+int get_num_servers(void);
+
 /* max number of preempt orderings */
 #define PREEMPT_ORDER_MAX 20
+
+typedef struct svr_conn {
+	int sd;                      /* File descriptor for the open socket */
+	int secondary_sd;            /* Secondary File descriptor for the open socket */
+	int state;                   /* Connection state */
+	time_t last_used_time;       /* Last used time for the connection */
+	char name[PBS_MAXSERVERNAME];  /* server name */
+	int port;                    /* server port */
+	int from_sched;              /* flag to indicate whether this conn is from sched or not */
+} svr_conn_t;
 
 /* PBS Batch Reply Structure */
 
@@ -299,69 +315,71 @@ struct batch_reply
 
 #define EXTEND_OPT_IMPLICIT_COMMIT ":C:" /* option added to pbs_submit() extend parameter to request implicit commit */
 
-extern int is_compose(int, int);
-extern int is_compose_cmd(int, int, char **);
-extern void PBS_free_aopl(struct attropl *);
-extern void advise(char *, ...);
-extern int PBSD_rdytocmt(int, char *, int, char **);
-extern int PBSD_commit(int, char *, int, char **);
-extern int PBSD_jcred(int, int, char *, int, int, char **);
-extern int PBSD_jscript(int, char *, int, char **);
-extern int PBSD_jscript_direct(int, char *, int, char **);
-extern int PBSD_copyhookfile(int, char *, int, char **);
-extern int PBSD_delhookfile(int, char *, int, char **);
-extern int PBSD_mgr_put(int, int, int, int, char *, struct attropl *, char *, int, char **);
-extern int PBSD_manager(int, int, int, int, char *, struct attropl *, char *);
-extern int PBSD_msg_put(int, char *, int, char *, char *, int, char **);
-extern int PBSD_relnodes_put(int, char *, char *, char *, int, char **);
-extern int PBSD_py_spawn_put(int, char *, char **, char **, int, char **);
-extern int PBSD_sig_put(int, char *, char *, char *, int, char **);
-extern int PBSD_term_put(int, int, char *);
-extern int PBSD_jobfile(int, int, char *, char *, enum job_file, int, char **);
-extern int PBSD_status_put(int, int, char *, struct attrl *, char *, int, char **);
-extern struct batch_reply *PBSD_rdrpy(int);
-extern struct batch_reply *PBSD_rdrpy_sock(int, int *);
-extern void PBSD_FreeReply(struct batch_reply *);
-extern struct batch_status *PBSD_status(int, int, char *, struct attrl *, char *);
-extern preempt_job_info *PBSD_preempt_jobs(int, char **);
-extern struct batch_status *PBSD_status_get(int);
-extern char *PBSD_queuejob(int, char *, char *, struct attropl *, char *, int, char **, int *);
-extern int decode_DIS_svrattrl(int, pbs_list_head *);
-extern int decode_DIS_attrl(int, struct attrl **);
-extern int decode_DIS_JobId(int, char *);
-extern int decode_DIS_replyCmd(int, struct batch_reply *);
-extern int encode_DIS_JobCred(int, int, char *, int);
-extern int encode_DIS_UserCred(int, char *, int, char *, int);
-extern int encode_DIS_JobFile(int, int, char *, int, char *, int);
-extern int encode_DIS_JobId(int, char *);
-extern int encode_DIS_Manage(int, int, int, char *, struct attropl *);
-extern int encode_DIS_MessageJob(int, char *, int, char *);
-extern int encode_DIS_MoveJob(int, char *, char *);
-extern int encode_DIS_ModifyResv(int, char *, struct attropl *);
-extern int encode_DIS_RelnodesJob(int, char *, char *);
-extern int encode_DIS_PySpawn(int, char *, char **, char **);
-extern int encode_DIS_QueueJob(int, char *, char *, struct attropl *);
-extern int encode_DIS_SubmitResv(int, char *, struct attropl *);
-extern int encode_DIS_JobCredential(int, int, char *, int);
-extern int encode_DIS_ReqExtend(int, char *);
-extern int encode_DIS_ReqHdr(int, int, char *);
-extern int encode_DIS_Rescq(int, char **, int);
-extern int encode_DIS_Run(int, char *, char *, unsigned long);
-extern int encode_DIS_ShutDown(int, int);
-extern int encode_DIS_SignalJob(int, char *, char *);
-extern int encode_DIS_Status(int, char *, struct attrl *);
-extern int encode_DIS_attrl(int, struct attrl *);
-extern int encode_DIS_attropl(int, struct attropl *);
-extern int encode_DIS_CopyHookFile(int, int, char *, int, char *);
-extern int encode_DIS_DelHookFile(int, char *);
-extern int encode_DIS_PreemptJobs(int, char **);
-extern char *PBSD_submit_resv(int, char *, struct attropl *, char *);
-extern int DIS_reply_read(int, struct batch_reply *, int);
-extern int tcp_pre_process(conn_t *);
-extern char *PBSD_modify_resv(int, char *, struct attropl *, char *);
-extern int PBSD_cred(int, char *, char *, int, char *, long, int, char **);
-
-extern int tcp_send_auth_req(int, unsigned int, char *, char *, char *);
+int is_compose(int, int);
+int is_compose_cmd(int, int, char **);
+void PBS_free_aopl(struct attropl *);
+void advise(char *, ...);
+int PBSD_rdytocmt(int, char *, int, char **);
+int PBSD_commit(int, char *, int, char **);
+int PBSD_jcred(int, int, char *, int, int, char **);
+int PBSD_jscript(int, char *, int, char **);
+int PBSD_jscript_direct(int, char *, int, char **);
+int PBSD_copyhookfile(int, char *, int, char **);
+int PBSD_delhookfile(int, char *, int, char **);
+int PBSD_mgr_put(int, int, int, int, char *, struct attropl *, char *, int, char **);
+int PBSD_manager(int, int, int, int, char *, struct attropl *, char *);
+int PBSD_msg_put(int, char *, int, char *, char *, int, char **);
+int PBSD_relnodes_put(int, char *, char *, char *, int, char **);
+int PBSD_py_spawn_put(int, char *, char **, char **, int, char **);
+int PBSD_sig_put(int, char *, char *, char *, int, char **);
+int PBSD_term_put(int, int, char *);
+int PBSD_jobfile(int, int, char *, char *, enum job_file, int, char **);
+int PBSD_status_put(int, int, char *, struct attrl *, char *, int, char **);
+struct batch_reply *PBSD_rdrpy(int);
+struct batch_reply *PBSD_rdrpy_sock(int, int *);
+void PBSD_FreeReply(struct batch_reply *);
+struct batch_status *PBSD_status(int, int, char *, struct attrl *, char *);
+struct batch_status *PBSD_status_random(int c, int function, char *id, struct attrl *attrib, char *extend, int parent_object);
+struct batch_status *PBSD_status_aggregate(int c, int cmd, char *id, struct attrl *attrib, char *extend, int parent_object);
+preempt_job_info *PBSD_preempt_jobs(int, char **);
+struct batch_status *PBSD_status_get(int);
+char *PBSD_queuejob(int, char *, char *, struct attropl *, char *, int, char **, int *);
+int decode_DIS_svrattrl(int, pbs_list_head *);
+int decode_DIS_attrl(int, struct attrl **);
+int decode_DIS_JobId(int, char *);
+int decode_DIS_replyCmd(int, struct batch_reply *);
+int encode_DIS_JobCred(int, int, char *, int);
+int encode_DIS_UserCred(int, char *, int, char *, int);
+int encode_DIS_JobFile(int, int, char *, int, char *, int);
+int encode_DIS_JobId(int, char *);
+int encode_DIS_Manage(int, int, int, char *, struct attropl *);
+int encode_DIS_MessageJob(int, char *, int, char *);
+int encode_DIS_MoveJob(int, char *, char *);
+int encode_DIS_ModifyResv(int, char *, struct attropl *);
+int encode_DIS_RelnodesJob(int, char *, char *);
+int encode_DIS_PySpawn(int, char *, char **, char **);
+int encode_DIS_QueueJob(int, char *, char *, struct attropl *);
+int encode_DIS_SubmitResv(int, char *, struct attropl *);
+int encode_DIS_JobCredential(int, int, char *, int);
+int encode_DIS_ReqExtend(int, char *);
+int encode_DIS_ReqHdr(int, int, char *);
+int encode_DIS_Rescq(int, char **, int);
+int encode_DIS_Run(int, char *, char *, unsigned long);
+int encode_DIS_ShutDown(int, int);
+int encode_DIS_SignalJob(int, char *, char *);
+int encode_DIS_Status(int, char *, struct attrl *);
+int encode_DIS_attrl(int, struct attrl *);
+int encode_DIS_attropl(int, struct attropl *);
+int encode_DIS_CopyHookFile(int, int, char *, int, char *);
+int encode_DIS_DelHookFile(int, char *);
+int encode_DIS_PreemptJobs(int, char **);
+char *PBSD_submit_resv(int, char *, struct attropl *, char *);
+int DIS_reply_read(int, struct batch_reply *, int);
+int tcp_pre_process(conn_t *);
+char *PBSD_modify_resv(int, char *, struct attropl *, char *);
+int PBSD_cred(int, char *, char *, int, char *, long, int, char **);
+int tcp_send_auth_req(int, unsigned int, char *, char *, char *);
+void *get_conn_servers(void);
 
 #ifdef __cplusplus
 }
--- a/src/include/libpbs.h
+++ b/src/include/libpbs.h
@@ -145,7 +145,7 @@ typedef struct svr_conn {
 	int secondary_sd;            /* Secondary File descriptor for the open socket */
 	int state;                   /* Connection state */
 	time_t last_used_time;       /* Last used time for the connection */
-	char name[PBS_MAXSERVERNAME];  /* server name */
+	char name[PBS_MAXSERVERNAME + 1];  /* server name */
 	int port;                    /* server port */
 	int from_sched;              /* flag to indicate whether this conn is from sched or not */
 } svr_conn_t;
--- a/src/include/libpbs.h
+++ b/src/include/libpbs.h
@@ -142,13 +142,18 @@ int get_num_servers(void);
 
 typedef struct svr_conn {
 	int sd;                      /* File descriptor for the open socket */
-	int secondary_sd;            /* Secondary File descriptor for the open socket */
 	int state;                   /* Connection state */
 	time_t last_used_time;       /* Last used time for the connection */
 	char name[PBS_MAXSERVERNAME + 1];  /* server name */
 	int port;                    /* server port */
 	int from_sched;              /* flag to indicate whether this conn is from sched or not */
 } svr_conn_t;
+
+typedef struct svr_conns_list {
+	svr_conn_t *conn_arr;
+	struct svr_conns_list *next; /* Pointer to next set of server connections,
+								in case client calls pbs_connect multiple times */
+} svr_conns_list_t;
 
 /* PBS Batch Reply Structure */
 
@@ -379,7 +384,8 @@ int tcp_pre_process(conn_t *);
 char *PBSD_modify_resv(int, char *, struct attropl *, char *);
 int PBSD_cred(int, char *, char *, int, char *, long, int, char **);
 int tcp_send_auth_req(int, unsigned int, char *, char *, char *);
-void *get_conn_servers(void);
+void *get_conn_servers(int);
+void dealloc_conn_list_single(int parentfd);
 
 #ifdef __cplusplus
 }
--- a/src/include/pbs_ifl.h
+++ b/src/include/pbs_ifl.h
@@ -468,6 +468,7 @@ enum mgr_obj {
 #define PBS_MAXQUEUENAME	15		/* max queue name length */
 #define PBS_MAXJOBNAME  	230		/* max job name length */
 #define PBS_MAXSERVERNAME	PBS_MAXHOSTNAME	/* max server name length */
+#define MAX_SVR_ID (PBS_MAXSERVERNAME + 6)	/* max server name + port */
 #define PBS_MAXSEQNUM		12		/* max sequence number length */
 #define PBS_DFLT_MAX_JOB_SEQUENCE_ID 9999999	/* default value of max_job_sequence_id server attribute */
 #define PBS_MAXPORTNUM	5		/* udp/tcp port numbers max=16 bits */
--- a/src/include/pbs_ifl.h
+++ b/src/include/pbs_ifl.h
@@ -468,7 +468,6 @@ enum mgr_obj {
 #define PBS_MAXQUEUENAME	15		/* max queue name length */
 #define PBS_MAXJOBNAME  	230		/* max job name length */
 #define PBS_MAXSERVERNAME	PBS_MAXHOSTNAME	/* max server name length */
-#define MAX_SVR_ID (PBS_MAXSERVERNAME + 6)	/* max server name + port */
 #define PBS_MAXSEQNUM		12		/* max sequence number length */
 #define PBS_DFLT_MAX_JOB_SEQUENCE_ID 9999999	/* default value of max_job_sequence_id server attribute */
 #define PBS_MAXPORTNUM	5		/* udp/tcp port numbers max=16 bits */
--- a/src/include/pbs_internal.h
+++ b/src/include/pbs_internal.h
@@ -58,6 +58,14 @@ extern "C" {
  *
  */
 
+#ifdef WIN32
+#define CLOSESOCKET(X) (void)closesocket(X)
+#define ERRORNO        WSAGetLastError()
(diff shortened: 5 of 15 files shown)


src/lib/Libifl/pbsD_connect.c
```diff
@@ -470,12 +349,186 @@ __pbs_connect_extend(char *server, char *extend_data)
...
+ * @retval	fd of connection
+ */
+static int
+connect_to_server(int idx, svr_conn_t *conn_arr, char *extend_data)
+{
+	if (conn_arr[idx].state != SVR_CONN_STATE_UP) {
+		if ((conn_arr[idx].sd = tcp_connect(conn_arr[idx].name, conn_arr[idx].port, extend_data)) != -1) {
+			conn_arr[idx].state = SVR_CONN_STATE_UP;
+			pbs_client_thread_lock_conntable();
+			add_connection(conn_arr[idx].sd);
```
hirenvadalia, line 425: This seems wrong to me, this func was made only for conn_table.c (the one file who manipulates entries in connection table).
Also you don't need to do this, as tcp_connect calls to DIS routines (aka sends PBS_BATCH_Connect request and does authentication) which means given sd is already part of connection table. Here we are adding it 2nd time (although it won't add but it will just reset ch_errtxt and ch_errno of entry).

I would suggest, please revert add_connection() to static and remove this (along with lock/unlock of conntable).

hirenvadalia (changes requested):

src/cmds/pbsnodes.c
```diff
@@ -1077,6 +1071,15 @@ main(int argc, char *argv[])
 		exit(1);
 	}
 
+	if (def_server == NULL) {
+		def_server = pbs_default();
+		if (def_server == NULL)
+			def_server = "";
+
+		/* User didn't specify a server, talk to all */
+		setenv(CONN_MULTI, "ENABLED", 1);
```
hirenvadalia, line 1080: I am not entirely sure that we want to go this route only (IIRC, (I may have completely misunderstood it but) this was quick fix we did in POC branch), @subhasisb please correct me if I am misunderstood the discussion in which we decided to use this approach in POC branch.

If we decide to go this route then I think we need design page which mention this as its something to be set before anyone calls IFL APIs, so if someone using directly IFL APIs, they need to take care of this.

src/lib/Libifl/pbsD_connect.c
```diff
@@ -254,101 +216,30 @@ get_hostsockaddr(char *host, struct sockaddr_in *sap)
...
-
-	/* get server host and port	*/
-
-	server = PBS_get_server(server, server_name, &server_port);
-	if (server == NULL) {
-		pbs_errno = PBSE_NOSERVER;
-		return -1;
-	}
-
-	if (pbs_conf.pbs_primary && pbs_conf.pbs_secondary) {
```
suresh-thelkar, line 304: This code actually handles failover code. Failover may not work without this. Are we going to handle it in another place by another method or so ?

src/lib/Libifl/pbsD_connect.c
```diff
@@ -470,12 +349,186 @@ __pbs_connect_extend(char *server, char *extend_data)
...
+					/* Client wants to connect to a specific server, so we return here regardless of result */
+					return connect_to_server(i, svr_connections, extend_data);
+				}
+			}
+		}
+
+		/* Client doesn't care which server to connect to, try random first */
+		i = rand_num() % num_conf_servers;
+		fd = connect_to_server(i, svr_connections, extend_data);
+		if (fd != -1)
```
suresh-thelkar, line 473: If multi_flag is enabled means we have to connect to only one random server if server name is not provided. Suppose if fd == -1 here then we can actually try to find random servers from the rest of the active servers instead of connecting to all servers. This way we accidentally don't connect to all servers unless the client has real need to talk to all servers.

agrawalravi90: there is a break in loop below which will make sure that it doesn't connect to all when multi flag is absent.

src/lib/Libifl/pbsD_connect.c
agrawalravi90: I remember that there were issues without this call, i can try removing it to see what happens.

src/cmds/pbsnodes.c
agrawalravi90: hmm, good point, I'll think about the design again, I'm not sure that there's a better design for this. I guess the easiest option for clients would be if IFL always connected to all servers and then depending on the type of batch request, it will either send it to one more more servers. But each client request connecting to all servers might cause performance issues.

hirenvadalia: >  I guess the easiest option for clients would be if IFL always connected to all servers and then depending on the type of batch request, it would either send it to one more more servers. But each client request connecting to all servers might cause performance issues.

This will be definitely performance issue, as its not just connect but its connect + authenticate (+ establish encrypted channel if encryption is enabled).

src/lib/Libifl/pbsD_connect.c
agrawalravi90: we have plans to deprecate the existing failover infrastructure, there will be an announcement/design discussion for the same soon.

src/cmds/pbsnodes.c
agrawalravi90: Please provide feedback: https://openpbs.atlassian.net/wiki/spaces/PD/pages/2099249173/Switches+related+to+multi-server

src/lib/Libifl/pbsD_connect.c
agrawalravi90: > Also you don't need to do this, as tcp_connect calls to DIS routines (aka sends PBS_BATCH_Connect request and does authentication) which means given sd is already part of connection table. Here we are adding it 2nd time (although it won't add but it will just reset ch_errtxt and ch_errno of entry).

Can you please tell me where it's added the first time ?

src/lib/Libifl/pbsD_connect.c
```diff
@@ -470,12 +349,186 @@ __pbs_connect_extend(char *server, char *extend_data)
...
+			return fd;
+
+		/* didn't work, we'll try all servers below */
```
suresh-thelkar, line 476: This comment is misleading. In the absence of multi_flag we only connect to one of the random servers which is active. Can you please remove the comment or rephrase it.

src/cmds/pbsnodes.c
agrawalravi90: Update: outcome of the discussion above was that we don't need an option for client commands/programs to connect to just one server, so all client commands will connect to all by default. So, I'll change the code to remove the multi flag altogether

src/lib/Libifl/pbsD_connect.c
```diff
@@ -470,12 +349,186 @@ __pbs_connect_extend(char *server, char *extend_data)
...
+{
+	svr_conn_t *conn_arr = NULL;
+
+	conn_arr = pthread_getspecific(psi_key);
+	if (conn_arr == NULL && pbs_conf.psi != NULL) {
+		int num_svrs;
+		int i;
+
+		num_svrs = get_num_servers();
+		conn_arr = calloc(num_svrs, sizeof(svr_conn_t));
```
suresh-thelkar, line 386: It would be better to store pointers to the structure of type svr_conn_t i.e. basically making conn_arr as follows
svr_conn_t **conn_array = NULL;

If we do this we can make use of the following utility functions.

- void * add_ptr_to_array(void *ptr_arr, void *ptr) 
- int remove_ptr_from_array(char **arr, char *str)
- void ** dup_array(void *ptr)
- int count_array(void *arr)

We need most of these functions for the following scenarios.

- When a Server is decommissioned i.e. When one of the servers is removed. We need to shrink the array for which we can use  remove_ptr_from_array
- When a new server is added. We to need to realloc the array and add details of new servers for which we can use add_ptr_to_array

Scheduler code already makes use of these functions wherever needed. These are actually part of Scheduler/misc.c
In fact we can make this part of Libutil.

src/lib/Libifl/pbsD_connect.c
hirenvadalia: It will be added at `__pbs_connect_extend()`->`encode_DIS_ReqHdr()`->`diswui()` (aka `diswul()`)->`dis_puts()`->`dis_get_writebuf()`->`transport_get_chan()` (aka `tcp_get_chan()` for tcp connection)->`get_conn_chan()`->`get_connection()`->`add_connection()`

src/lib/Libifl/pbsD_connect.c
agrawalravi90: > When a Server is decommissioned i.e. When one of the servers is removed. We need to shrink the array for which we can use remove_ptr_from_array
When a new server is added. We to need to realloc the array and add details of new servers for which we can use add_ptr_to_array

When an admin changes the PBS_SERVER_INSTANCES list to decommission/add a new server, we would want to recreate the server connections array instead of modifying the existing array, right? Users might change the order of servers in the PBS_SERVER_INSTANCES list, so better to reallocate. This will be rare and the list will be small, so there'll be minimum performance hit. So, I think we are good there. Once we move to C++, we can use a vector for this list if needed. Please let me know what you think.

agrawalravi90:
@hirenvadalia and @suresh-thelkar I've removed the redundant call to add_connection, and added code to connect_to_server to handle secondary_sd, please review the latest commit and let me knw what you think. Thanks.

src/lib/Libifl/pbsD_connect.c
```diff
@@ -357,63 +248,46 @@ __pbs_connect_extend(char *server, char *extend_data)
...
 		}
 #endif
-		sock = socket(AF_INET, SOCK_STREAM, 0);
+	sd = socket(AF_INET, SOCK_STREAM, 0);
+	if (sd == -1) {
+		pbs_errno = PBSE_SYSTEM;
+		return -1;
+	}
 
+	strncpy(pbs_server, server, sizeof(pbs_server)-1); /* set for error messages from commands */
```
hirenvadalia, line 257: Use pbs_strncpy

src/lib/Libifl/pbsD_connect.c
```diff
@@ -357,63 +248,46 @@ __pbs_connect_extend(char *server, char *extend_data)
...
-			if (f != -1)
-				(void)close(f);
-		}
+	if (get_hostsockaddr(server, &server_addr) != 0)
+		return -1;
+
+	server_addr.sin_port = htons(server_port);
+	if (connect(sd, (struct sockaddr *)&server_addr, sizeof(struct sockaddr)) != 0) {
+		/* connect attempt failed */
+		CLOSESOCKET(sd);
```
hirenvadalia, line 278: please change this and other places according to #1917

src/lib/Libifl/pbsD_connect.c
```diff
@@ -440,27 +314,32 @@ __pbs_connect_extend(char *server, char *extend_data)
...
-	if (dis_flush(sock)) {
-		closesocket(sock);
+	if (dis_flush(sd)) {
+		CLOSESOCKET(sd);
 		pbs_errno = PBSE_SYSTEM;
 		return -1;
 	}
-	reply = PBSD_rdrpy(sock);
+
+	reply = PBSD_rdrpy_sock(sd, &rc);
```
hirenvadalia, line 329: Any reason to change from PBSD_rdrpy to PBSD_rdrpy_sock() ?

src/lib/Libifl/pbsD_connect.c
```diff
@@ -470,12 +349,162 @@ __pbs_connect_extend(char *server, char *extend_data)
...
+ */
+static int
+connect_to_server(int idx, svr_conn_t *conn_arr, char *extend_data)
+{
+	if (conn_arr[idx].state != SVR_CONN_STATE_UP) {
+		int sd;
+
+		if ((sd = tcp_connect(conn_arr[idx].name, conn_arr[idx].port, extend_data)) != -1) {
+			conn_arr[idx].state = SVR_CONN_STATE_UP;
+			if (conn_arr[idx].sd > 0 && conn_arr[idx].secondary_sd < 0)
```
hirenvadalia, line 426: How we ensure here that this is secondary_sd of same server for which sd belong ?

src/lib/Libifl/pbsD_connect.c
```diff
@@ -470,12 +349,162 @@ __pbs_connect_extend(char *server, char *extend_data)
...
+	int ret_fd = -1;
+	int num_conf_servers = get_num_servers();
+	svr_conn_t *svr_connections = get_conn_servers();
+
+	if (svr_connections == NULL)
+		return -1;
+
+	/* try to connect to as many as possible */
+	for (i = 0; i < num_conf_servers; i++) {
+		fd = connect_to_server(i, svr_connections, extend_data);
```
hirenvadalia, line 463: So if I understand this correctly then caller of pbs_connect() will get only ret_fd, but internally we opened connections to all servers, now when caller is done with connection so it will call pbs_disconnect() on ret_fd (returned by pbs_connect()), how we will close all connections to other servers which we opened in pbs_connect()?

src/lib/Libifl/pbsD_connect.c
```diff
@@ -594,6 +623,22 @@ __pbs_disconnect(int connect)
...
 
+	/* Update the server connection cache */
+	svr_conns = get_conn_servers();
+	if (svr_conns != NULL) {
+		int i;
+
+		for (i = 0; i < get_num_servers(); i++) {
+			if (svr_conns[i].sd == connect) {
+				svr_conns[i].sd = -1;
+				svr_conns[i].secondary_sd = -1;
```
hirenvadalia, line 634: Shouldn't we need to close this and free its resources?

src/include/pbs_ifl.h
```diff
@@ -558,6 +559,18 @@ typedef struct preempt_job_info {
         char	order[PREEMPT_METHOD_HIGH + 1];
 } preempt_job_info;
 
+typedef struct svr_conn {
+	int sd;                      /* File descriptor for the open socket */
+	int secondary_sd;            /* Secondary File descriptor for the open socket */
+	int state;                   /* Connection state */
+	time_t state_change_time;    /* Connnetion state change time */
+	time_t last_used_time;       /* Last used time for the connection */
+	char svr_id[MAX_SVR_ID];     /* svr_id of the form server_name:port */
```
hirenvadalia, line 568: remove this and MAX_SVR_ID, as its not going to used anywhere, this was needed in POC branch but not in here.
For more details see https://github.com/openpbs/openpbs/pull/2032#discussion_r490686158

src/include/pbs_ifl.h
```diff
@@ -558,6 +559,18 @@ typedef struct preempt_job_info {
...
+	char name[PBS_MAXHOSTNAME];  /* server name */
```
hirenvadalia, line 569: I think this should be PBS_MAXSERVERNAME

src/include/pbs_ifl.h
```diff
@@ -558,6 +559,18 @@ typedef struct preempt_job_info {
         char	order[PREEMPT_METHOD_HIGH + 1];
 } preempt_job_info;
 
+typedef struct svr_conn {
+	int sd;                      /* File descriptor for the open socket */
+	int secondary_sd;            /* Secondary File descriptor for the open socket */
+	int state;                   /* Connection state */
+	time_t state_change_time;    /* Connnetion state change time */
```
hirenvadalia, line 566: I think we are not using this and `last_used_time` anywhere, and I don't think we will need it also, as this was for early version of POC where we had sharding algo

src/include/pbs_ifl.h
```diff
@@ -558,6 +559,18 @@ typedef struct preempt_job_info {
         char	order[PREEMPT_METHOD_HIGH + 1];
 } preempt_job_info;
 
+typedef struct svr_conn {
```
hirenvadalia, line 562: Shouldn't this part of pbs_internal.h? Do we really want to expose this internal structure of IFL as public interface?

hirenvadalia (changes requested):

src/lib/Libifl/pbsD_connect.c
agrawalravi90: via 'idx'

src/lib/Libifl/pbsD_connect.c
agrawalravi90: Check the code block under /* Update the server connection cache */ inside \__pbs_disconnect

src/lib/Libifl/pbsD_connect.c
agrawalravi90: good catch, ill make the change, thanks

src/include/pbs_ifl.h
agrawalravi90: @nithinj and @bremanandjk please confirm

src/lib/Libifl/pbsD_connect.c
agrawalravi90: we should just rename pbs_strncpy -> strncpy

src/include/pbs_ifl.h
nithinj: It is not used in the POC. I can see it is useful for clients which are also daemons to decide when to reconnect, something which could be introduced later.

src/lib/Libifl/pbsD_connect.c
agrawalravi90: good question, @bremanandjk this was actually your code, can you please clarify?

agrawalravi90:
Things are breaking after the sched connection reversal change, so marking WIP until I've weeded out the issues

src/include/pbs_internal.h
```diff
@@ -58,6 +58,14 @@ extern "C" {
  *
  */
 
+#ifdef WIN32
```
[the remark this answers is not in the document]
agrawalravi90: I'll take care of this when i resolve conflicts

agrawalravi90:
@hirenvadalia I've addressed your comments, please provide further feedback. I have found that my changes cause issues in some commands like qsig, qdisable etc., especially when used with \@servername, so I'll keep the WIP tag while i debug them, but do continue the review.

src/lib/Libifl/pbsD_connect.c
suresh-thelkar: Now that we are thinking to re-use the same slot for a server which gone down and came back and also going to use same connection array for scheduler I think we are good here. Also in the first phase if somebody wants to decommission a server, we are asking Admin to restart scheduler this is because he can inadvertently change the order of servers in PBS_SERVER_INSTANCES we are good here. May be in future based on the need we can revisit this and use a vector like array if needed.

src/cmds/qmgr.c
```diff
@@ -110,6 +110,7 @@
 #define QMGR_TIMEOUT 900 /* qmgr connection timeout set to 15 min */
 time_t start_time = 0;
 time_t check_time = 0;
+int num_connected_svrs = 0;
```
subhasisb, line 113: Why do you need this variable at the qmgr.c level. I mean, if the connections are managed by IFL transparently, do we need to track anything at the level of the clients (or commands)?

Besides, it appears that num_connected_svrs is not being used in qmgr.c?

src/include/pbs_ifl.h
```diff
@@ -468,6 +468,7 @@ enum mgr_obj {
 #define PBS_MAXQUEUENAME	15		/* max queue name length */
 #define PBS_MAXJOBNAME  	230		/* max job name length */
 #define PBS_MAXSERVERNAME	PBS_MAXHOSTNAME	/* max server name length */
+#define MAX_SVR_ID (PBS_MAXSERVERNAME + 6)	/* max server name + port */
```
subhasisb, line 471: Want to add the word "INSTANCE" to this variable?

src/lib/Libifl/pbsD_connect.c
```diff
@@ -254,101 +216,30 @@ get_hostsockaddr(char *host, struct sockaddr_in *sap)
 }
 
 /**
- * @brief
- *	Makes a PBS_BATCH_Connect request to 'server'.
+ * @brief	This function establishes the network connection to the choose server.
```
subhasisb, line 219: typo - choose --> chosen

src/lib/Libifl/pbsD_connect.c
```diff
@@ -357,63 +248,46 @@ __pbs_connect_extend(char *server, char *extend_data)
...
-		if (pbs_conf.pbs_public_host_name) {
-			/* my address will be in my_sockaddr,  bind the socket to it */
-			my_sockaddr.sin_port = 0;
-			if (bind(sock, (struct sockaddr *)&my_sockaddr, sizeof(my_sockaddr)) != 0) {
-				return -1;
-			}
-		}
-
-		if (get_hostsockaddr(server, &server_addr) != 0)
+	/* If a specific host name is defined which the client should use */
```
subhasisb, line 261: What does this comment mean?

src/lib/Libifl/pbsD_connect.c
```diff
@@ -357,63 +248,46 @@ __pbs_connect_extend(char *server, char *extend_data)
...
+	if (pbs_conf.pbs_public_host_name) {
+		if (get_hostsockaddr(pbs_conf.pbs_public_host_name, &my_sockaddr) != 0)
```
subhasisb, line 263: I feel we can keep tcp_connect() devoid of any pollution from pbs_conf. The goal for tcp_connect() can be very simple, given the hostname and port, connect to it, do not try to interpret it any further than what is provided. The callers of this function should decide what actual hostname should be passed to it (and how pbs_conf.pbs_public_host_name) affects what is passed to this routine.

src/lib/Libifl/pbsD_connect.c
```diff
@@ -470,12 +349,167 @@ __pbs_connect_extend(char *server, char *extend_data)
...
+	int ret_fd = -1;
+	int num_conf_servers = get_num_servers();
+	svr_conn_t *svr_connections = get_conn_servers();
+
+	if (svr_connections == NULL)
+		return -1;
+
+	if (server_name != NULL && pbs_conf.pbs_server_name != NULL) {
+		/* Make sure that the name matches PBS_SERVER_NAME */
+		if (!is_same_host(server_name, pbs_conf.pbs_server_name)) {
```
subhasisb, line 462: Are we saying that users cannot do qstat @clusterB? There could be valid cases of user wanting to connect to a cluster that is not the one defined in its local conf file in PBS_SERVER, no?

src/lib/Libifl/pbsD_connect.c
```diff
@@ -510,9 +544,8 @@ pbs_connection_set_nodelay(int connect)
 }
 
 /**
- * @brief
- *	A wrapper progarm to pbs_connect_extend() but this one not
- *	passing any 'extend' data to the connection.
+ * @brief	A wrapper progarm to pbs_connect_extend() but this one not
```
subhasisb, line 547: program? it is just an API, no?

src/lib/Libifl/pbsD_connect.c
```diff
@@ -592,7 +623,45 @@ __pbs_disconnect(int connect)
...
+ *
+ * @param[in] connect - socket descriptor
+ *
+ * @return	int
+ * @retval	0	success
+ * @retval	-1	error
+ *
+ */
+int
+__pbs_disconnect(int connect)
```
subhasisb, line 643: so pbs_connect() will automatically connect to N underlying server instances, but pbs_disconnect() will disconnect only from the one exact fd (connect) passed to it. That sounds like a disparity in the design, no?

In other words, which call, or what party would know to call pbs_disconnect() on each of the underlying sever instance connections?

src/lib/Libifl/pbsD_connect.c
```diff
@@ -525,18 +558,16 @@ __pbs_connect(char *server)
...
  *
  * @return	int
  * @retval	0	success
  * @retval	-1	error
- *
  */
-int
-__pbs_disconnect(int connect)
+static int
+__pbs_disconnect_2(int connect)
```
subhasisb, line 570: perhaps a more meaningful name is called for - how about disconnect_from_server() (in synergy with the corresponding connect function which is called connect_to_server())

subhasisb (changes requested):

src/include/libpbs.h
```diff
@@ -131,9 +132,24 @@ pbs_tcp_chan_t * get_conn_chan(int);
...
+
 /* max number of preempt orderings */
 #define PREEMPT_ORDER_MAX 20
 
+typedef struct svr_conn {
+	int sd;                      /* File descriptor for the open socket */
+	int secondary_sd;            /* Secondary File descriptor for the open socket */
+	int state;                   /* Connection state */
+	time_t last_used_time;       /* Last used time for the connection */
+	char name[PBS_MAXSERVERNAME];  /* server name */
```
hirenvadalia, line 148: I think this should be `PBS_MAXSERVERNAME + 1`

src/include/pbs_internal.h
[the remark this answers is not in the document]
hirenvadalia: Can you please resolve conflicts if any and change this?

src/include/pbs_ifl.h
hirenvadalia, line 471: I suggest to use PBS_MAXPORTNUM instead hard coding 6 (aka `PBS_MAXSERVERNAME + PBS_MAXPORTNUM + 1`

src/lib/Libifl/pbsD_connect.c
```diff
@@ -357,63 +248,46 @@ __pbs_connect_extend(char *server, char *extend_data)
...
 		}
 #endif
-		sock = socket(AF_INET, SOCK_STREAM, 0);
+	sd = socket(AF_INET, SOCK_STREAM, 0);
+	if (sd == -1) {
+		pbs_errno = PBSE_SYSTEM;
+		return -1;
+	}
 
+	pbs_strncpy(pbs_server, server, sizeof(pbs_server)-1); /* set for error messages from commands */
```
hirenvadalia, line 257: in pbs_strncpy already take care of -1, so no need to add -1 here in sizeof(pbs_server)

src/lib/Libifl/pbsD_connect.c
hirenvadalia: @bremanandjk Can you please clarify this change?

src/lib/Libifl/pbsD_connect.c
```diff
@@ -470,12 +349,167 @@ __pbs_connect_extend(char *server, char *extend_data)
...
+	if (pbs_loadconf(0) == 0)
+		return -1;
+
+	server = PBS_get_server(server, server_name, &server_port);
+	if (server == NULL) {
+		pbs_errno = PBSE_NOSERVER;
+		return -1;
+	}
+
+	if ((sock = connect_to_servers(server_name, server_port, extend_data)) == -1)
```
hirenvadalia, line 510: How about simply `return connect_to_servers(...`

hirenvadalia (changes requested):

src/lib/Libifl/pbsD_connect.c
bremanandjk: In POC implementation, there was a need to convert the connection handle to the respective socket since the connection handle was used as a virtual fd. Here, now I think we could use PBSD_rdrpy(sock) directly and also get the error code as sd represents the physical socket.

src/include/pbs_ifl.h
agrawalravi90: It's not used anymore so ill remove it

src/lib/Libifl/pbsD_connect.c
agrawalravi90: @bremanandjk  can you please clarify?

bremanandjk: It seems the code associated with this comment is refactored or removed. Probably we can remove this line.

src/include/pbs_internal.h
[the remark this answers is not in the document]
agrawalravi90: im surprised that it's still there, maybe i messed it up while resolving conflicts, ill take care it, thanks

src/lib/Libifl/pbsD_connect.c
agrawalravi90: Good catch, i think the only check we can do here then is to see if the host is reachable or not, please let me know what you think.

src/lib/Libifl/pbsD_connect.c
agrawalravi90: good point, we should disconnect from all if we are connecting to all inside pbs_connect

src/lib/Libifl/pbsD_connect.c
```diff
@@ -592,7 +667,46 @@ __pbs_disconnect(int connect)
...
+ * @retval	0	success
+ * @retval	-1	error
+ *
+ */
+int
+__pbs_disconnect(int connect)
+{
+	svr_conn_t *svr_conns = NULL;
+
+	if (connect <= 0)
```
hirenvadalia, line 691: isn't 0 valid fd?

src/lib/Libifl/pbsD_connect.c
```diff
@@ -470,13 +335,226 @@ __pbs_connect_extend(char *server, char *extend_data)
...
+	if (pbs_loadconf(0) == 0)
+		return -1;
+
+	server = PBS_get_server(server, server_name, &server_port);
+	if (server == NULL) {
+		pbs_errno = PBSE_NOSERVER;
+		return -1;
+	}
+
+	return connect_to_servers(server_name, server_port, extend_data);
```
hirenvadalia, line 557: Not sure but shouldn't we need to do this for pbs_connect_noblk() also?

hirenvadalia (changes requested):

src/lib/Libifl/pbsD_connect.c
agrawalravi90: 0, 1 and 2 are reserved (so ideally it should be <=2), but here i can just check for -1 i think.

src/lib/Libifl/pbs_loadconf.c
```diff
@@ -1109,6 +1128,12 @@ __pbs_loadconf(int reload)
 
 	pbs_conf.loaded = 1;
 
+	if (parse_psi(psi_value ? psi_value : pbs_default()) == -1) {
```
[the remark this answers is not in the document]
agrawalravi90: Nope, since we call pbs_default(), which can again call pbs_loadconf, we need to call it after pbs_conf.loaded is set.

src/lib/Libifl/pbsD_connect.c
agrawalravi90: Thanks for mentioning that, I hadn't thought about it. AFAIK, right now we don't have plans to support a multi-multi-server configuration, so when the scheduler calls pbs_connect_noblk(), it will connect to the one remote server that it wants to connect to, and then can call pbs_disconnect to disconnect from that single server. I have made a small change to pbs_disconnect to support this. Other than that, we don't need to make any changes to pbs_connect_noblk() to support multi-server since it's an internal interface. Let me know what you think.

src/lib/Libifl/pbsD_connect.c
```diff
@@ -505,12 +505,39 @@ connect_to_servers(char *svrhost, uint port, char *extend_data)
...
+				ipaddr1 = &(dflt_sockaddr.sin_addr);
+				ipaddr2 = &(tmp_sockaddr.sin_addr);
+				inet_ntop(AF_INET, ipaddr1, ipstr1, sizeof(ipstr1));
+				inet_ntop(AF_INET, ipaddr2, ipstr2, sizeof(ipstr2));
+				if (strcmp(ipstr1, ipstr2) != 0) {
+					/* The client is trying to reach a different cluster than what's known
+					 * So, just reach out to the one host provided and reply back instead
+					 * of connecting to the PBS_SERVER_INSTANCES of the default cluster
+					 */
+					return connect_to_server(0, svr_connections, extend_data);
```
subhasisb, line 537: dont we need to pass the svrhost to connect_to_server() in this case?

subhasisb (changes requested):

src/lib/Libifl/pbsD_connect.c
```diff
@@ -505,12 +505,39 @@ connect_to_servers(char *svrhost, uint port, char *extend_data)
...
+			if (get_hostsockaddr(dflt_server, &dflt_sockaddr) == 0) {
+				char ipstr1[INET_ADDRSTRLEN];
+				char ipstr2[INET_ADDRSTRLEN];
+				void *ipaddr1;
+				void *ipaddr2;
+
+
+				ipaddr1 = &(dflt_sockaddr.sin_addr);
+				ipaddr2 = &(tmp_sockaddr.sin_addr);
+				inet_ntop(AF_INET, ipaddr1, ipstr1, sizeof(ipstr1));
```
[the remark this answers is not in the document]
subhasisb: but this is also fine, i will leave it upto u to change or not

subhasisb (approved):

hirenvadalia (approved):
@agrawalravi90 Thanks for addressing comments, I don't have any further comments. Please resolve CLOSESOCKET one and post latest regression results.

src/lib/Libifl/pbsD_connect.c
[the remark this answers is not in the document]
agrawalravi90: I wanted to cover the cases where user provides fqdn vs shortname or provides an ip address itself.

agrawalravi90:
Regression results:

Full run:
Screen Shot 2020-10-07 at 8 47 46 PM

Re-running failed tests:
Screen Shot 2020-10-07 at 8 48 17 PM

I believe the rest of the failures are common with master

subhasisb (approved):

PR #2037, opened 2020-09-18
Comment out code in test_reservation due to a PBS bug

lisa-altair:
#### Describe Bug or Feature

There was a dangling server restart at the end of a test.  The check that would have happened after the restart was commented out for a "known issue".  But we didn't know what the issue was.  And we should have also commented out the server restart.

#### Describe Your Change

I added details about what the issue is that is causing the code to be commented out.
I also commented out the server restart.

Showing that the test still passes.
[after.txt](https://github.com/openpbs/openpbs/files/5248668/after.txt)

### DIFF
--- a/test/tests/functional/pbs_accumulate_resc_used.py
+++ b/test/tests/functional/pbs_accumulate_resc_used.py
@@ -753,6 +753,9 @@ else:
     def test_reservation(self):
         """
         Test that job inside reservations works same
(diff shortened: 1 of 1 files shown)


hirenvadalia:
@lisa-altair Can you please attach log showing test is failing (aka before fix test log)?

bayucan (approved):
Looks good.

bhroam (approved):
LGTM

PR #2038, opened 2020-09-21
Adding max_run_subjobs limit to array jobs

arungrover:
#### Describe Bug or Feature
This change adds a new attribute to array jobs that allow users to limit the number of subjobs they want to run concurrently.

#### Describe Your Change
The new option goes as an input to PBS scheduler. Scheduler makes sure that any array job does not exceed its max_run_subjobs limit.
If user does not specify this new limit, the behavior in the scheduler remains as it is.

#### Link to Design Doc
[Forum Discussion](https://community.openpbs.org/t/limit-on-maximum-number-of-subjobs-that-can-run-at-a-time/2274/17)
[Design Doc](https://openpbs.atlassian.net/wiki/spaces/PD/pages/2093973505/Limit+on+maximum+number+of+subjobs+that+can+run+at+a+time)

[test_array.txt](https://github.com/openpbs/openpbs/files/5258371/test_array.txt)
running more regression tests


test/tests/functional/pbs_job_array.py
```diff
@@ -39,12 +39,50 @@
...
     Test suite for job array feature
     """
+    lu = PBSLogUtils()
+    qjh = """
+import pbs
+
+e = pbs.event()
+j = e.job
+if j.max_run_subjobs is None:
+    j.max_run_subjobs = %d
```
bhroam, line 56: This just made think of something.  This will set max_run_subjobs on all jobs, not just subjobs.  Maybe there should be an action function which rejects it for non-job arrays?

test/tests/functional/pbs_job_array.py
```diff
@@ -39,12 +39,50 @@
...
+e = pbs.event()
+j = e.job
+if j.max_run_subjobs != 0:
+    if j.max_run_subjobs > 10:
+        j.max_run_subjobs = %d
+pbs.logmsg(pbs.LOG_DEBUG, "max_run_subjobs set to %%d" %% j.max_run_subjobs)
+e.accept()
+"""
+
+    def create_hook(self, max_run, event, name):
```
bhroam, line 72: Maybe make this function name less generic?  This is a generic test plan.  Something like create_max_run_subjobs_hook() ?

test/tests/functional/pbs_job_array.py
```diff
@@ -744,3 +782,203 @@ def test_recover_big_array_job(self):
...
+        a = {'resources_available.ncpus': 8}
+        self.server.manager(MGR_CMD_SET, NODE, a, self.mom.shortname)
+        j = Job(attrs={ATTR_J: '1-20%2'})
+        j_id = self.server.submit(j)
+        self.server.expect(JOB, {ATTR_state: 'B'}, id=j_id)
+        self.server.expect(JOB, {'job_state=R': 2}, extend='t')
+
+        self.server.alterjob(j_id, {ATTR_W: 'max_run_subjobs=5'})
+        a = {'scheduling': 'True'}
+        self.server.manager(MGR_CMD_SET, SERVER, a)
```
bhroam, line 802: This manager() call can fit within 79 characters.  No need to split it into 2 lines.

test/tests/functional/pbs_job_array.py
```diff
@@ -744,3 +782,203 @@ def test_recover_big_array_job(self):
...
+        a = {ATTR_J: '1-20%3', 'Resource_List.select': 'ncpus=2',
+             ATTR_q: 'wq2'}
+        j_arr = Job(attrs=a)
+        jid_arr = self.server.submit(j_arr)
+        self.server.expect(JOB, {ATTR_state: 'B'}, id=jid_arr)
+        self.server.expect(JOB, {'job_state=R': 4}, extend='t')
+        self.server.expect(JOB, {ATTR_state: 'R'}, id=jid)
+        msg = "Job cannot run due to max_run_subjobs limit,"
+        " preemption not attempted"
+        self.scheduler.log_match(jid_arr + ';' + msg)
```
bhroam, line 903: I don't think this test is necessary

test/tests/functional/pbs_job_array.py
```diff
@@ -744,3 +782,203 @@ def test_recover_big_array_job(self):
...
+
+    @skipOnCpuSet
+    def test_max_run_subjobs_qrun(self):
+        """
+        Submit array job with max_run_subjobs limit and see if such a job
+        is run using qrun, max_run_subjobs limits is ignored.
```
bhroam, line 909: I think you don't need the s on limits

test/tests/functional/pbs_job_array.py
```diff
@@ -744,3 +782,203 @@ def test_recover_big_array_job(self):
...
+        """
+        a = {'queue_type': 'execution',
+             'started': 'True',
+             'enabled': 'True',
+             'Priority': 200}
+        self.server.manager(MGR_CMD_CREATE, QUEUE, a, "wq2")
```
bhroam, line 915: Why do you need wq2 for this test?

bhroam (changes requested):

src/lib/Libattr/master_job_attr_def.xml
```diff
@@ -2054,6 +2054,23 @@
...
+      <member_at_set>set_l</member_at_set>
+      <member_at_comp>comp_l</member_at_comp>
+      <member_at_free>free_null</member_at_free>
+      <member_at_action>NULL_FUNC</member_at_action>
+      <member_at_flags>READ_WRITE | ATR_DFLAG_ALTRUN</member_at_flags>
+      <member_at_type>ATR_TYPE_LONG</member_at_type>
+      <member_at_parent>PARENT_TYPE_JOB</member_at_parent>
+      <member_verify_function>
+         <ECL>verify_datatype_long</ECL>
+         <ECL>verify_value_zero_or_positive</ECL>
```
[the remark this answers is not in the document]
arungrover: It's zero or positive because if someone wants to hold off running subjobs without penalizing the accrue type or its position in the calendar then they can do so.
I'll add this to the design doc as well.

src/scheduler/job_info.c
```diff
@@ -3037,6 +3057,13 @@ find_and_preempt_jobs(status *policy, int pbs_sd, resource_resv *hjob, server_in
 	if (hjob->aoename != NULL)
 		return 0;
 
+	/* If a job couldn't run because of max_run_subjobs limit, do not try preemption */
+	if (hjob->job->is_array && err->error_code == MAX_RUN_SUBJOBS) {
+		log_event(PBSEVENT_DEBUG, PBS_EVENTCLASS_JOB, LOG_DEBUG, hjob->name,
+			  "Job cannot run due to max_run_subjobs limit, preemption not attempted");
+		return 0;
+	}
```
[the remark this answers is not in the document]
arungrover: I'll add it there because that seems to be the right place for it. But the reason I added it up top was that there isn't really a need to go forward and call is_ok_to_run again. 
In the case of qrun request, the error structure main_sched_loop passes to this function will not have MAX_RUN_SUBJOBS error set in it because is_ok_to_run would not set it and it runs the job. Same will happen with the second call to is_ok_to_run in this function.

src/scheduler/resource_resv.c
```diff
@@ -1726,6 +1730,14 @@ update_resresv_on_end(resource_resv *resresv, char *job_state)
 			resresv->job->is_susp_sched = 1;
 			for (i = 0; ns[i] != NULL; i++)
 				ns[i]->ninfo->num_susp_jobs++;
+		} else {
+			if (resresv->job->is_subjob && resresv->job->parent_job &&
+			    resresv->job->parent_job->job->max_run_subjobs != UNSPECIFIED) {
+				resresv->job->parent_job->job->running_subjobs--;
+				if (resresv->job->parent_job->job->running_subjobs <
+				    resresv->job->parent_job->job->max_run_subjobs)
+					resresv->job->parent_job->can_not_run = 0;
```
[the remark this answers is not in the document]
arungrover: Wouldn't we need this in case of calendaring? When we simulate END event of a subjob, we should ideally be marking the parent as runnable.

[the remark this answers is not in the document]
bhroam: Calendaring doesn't care about can_not_run, it only cares about  run and end events.

src/scheduler/server_info.c
```diff
@@ -388,9 +389,28 @@ query_server(status *pol, int pbs_sd)
...
+	if (job_arrays_associated == FALSE) {
+		for (i = 0; sinfo->running_jobs[i] != NULL; i++) {
+			if ((sinfo->running_jobs[i]->job->is_subjob) &&
+			    (1 == associate_array_parent(sinfo->running_jobs[i], sinfo))) {
+				sinfo->fairshare = NULL;
+				free_server(sinfo);
+				return NULL;
+			}
+		}
+	}
```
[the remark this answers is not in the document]
arungrover: The reason being that we already run a loop on running jobs when a run limit is set. I wanted to avoid looping on running jobs again. But then for cases when run limits are not set, we need to loop on running jobs anyways and that's why the second loop. It looks vague but saves us looping on running jobs in some cases.

test/tests/functional/pbs_job_array.py
arungrover: I'll add action functions

test/tests/functional/pbs_job_array.py
arungrover: This test was to make sure that we don't try preemption when the job hits max_run_subjobs limit.

test/tests/functional/pbs_job_array.py
arungrover: You are right, that is not necessary

src/scheduler/resource_resv.c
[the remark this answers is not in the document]
arungrover: ah, ok then :)

test/tests/functional/pbs_job_array.py
```diff
@@ -744,3 +782,237 @@ def test_recover_big_array_job(self):
...
+        self.server.log_match("max_run_subjobs set to 3")
+        self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'True'})
+        self.server.expect(JOB, {'job_state=R': 3}, extend='t')
+
+        # Modify a normal job and see if queuejob hook cannot set the
+        # attribute.
+        self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': 'False'})
+        nj = self.server.submit(Job())
+        with self.assertRaises(PbsAlterError) as e:
+            self.server.alterjob(nj, {ATTR_W: 'max_run_subjobs=20'})
```
vstumpf, line 886: Can you try this with altering ncpus or something and see if the hook rejects?

vstumpf (changes requested):
some questions

test/tests/functional/pbs_job_array.py
```diff
@@ -744,3 +782,237 @@ def test_recover_big_array_job(self):
...
+        a = {ATTR_J: '1-20%3', 'Resource_List.select': 'ncpus=2',
+             ATTR_q: 'wq2'}
+        j_arr = Job(attrs=a)
+        jid_arr = self.server.submit(j_arr)
+        self.server.expect(JOB, {ATTR_state: 'B'}, id=jid_arr)
+        self.server.expect(JOB, {'job_state=R': 4}, extend='t')
+        self.server.expect(JOB, {ATTR_state: 'R'}, id=jid)
+
+    @skipOnCpuSet
+    def test_max_run_subjobs_qrun(self):
```
vstumpf, line 918: this test currently checks the admin's ability to qrun
can this also test what happens if the user tries to qrun

src/scheduler/job_info.c
```diff
@@ -2835,6 +2854,7 @@ dup_job_info(job_info *ojinfo, queue_info *nqinfo, server_info *nsinfo)
 	njinfo->array_index = ojinfo->array_index;
 	njinfo->array_id = string_dup(ojinfo->array_id);
 	njinfo->queued_subjobs = dup_range_list(ojinfo->queued_subjobs);
+	njinfo->max_run_subjobs = ojinfo->max_run_subjobs;
```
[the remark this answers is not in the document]
arungrover: Well in case of subjobs it will be 0 and for parent jobs it will be set and get copied.
running_subjobs is calculated in dup_server_info because we need to associate the parent job pointer as well. So associate_parent_job() function increments running_subjobs as well.

src/scheduler/server_info.c
```diff
@@ -361,7 +362,7 @@ query_server(status *pol, int pbs_sd)
 		allcts = find_alloc_counts(sinfo->alljobcounts, PBS_ALL_ENTITY);
 		if (sinfo->alljobcounts == NULL)
 			sinfo->alljobcounts = allcts;
-
+		job_arrays_associated = TRUE;
```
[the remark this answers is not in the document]
arungrover: In the loop right below this function I call associate_parent_job function. However, control goes to this block only when there are limits set in the system. To avoid looping on running jobs twice when limit is set, I set this flag here and check for it outside. If no limits were set then I loop on running jobs to associate parent jobs.

test/tests/functional/pbs_job_array.py
arungrover: I can do that, however it will require me to write a new modify job hook because the one I have does not set this attribute if it is already set to 0 (which when modifying ncpus will be the case). I'll add that to the test.

test/tests/functional/pbs_job_array.py
arungrover: I think qrun can only be executed by a manager or operator privilege user.

vstumpf: oh, you're right

arungrover:
[test_array.txt](https://github.com/openpbs/openpbs/files/5270259/test_array.txt)

I have addressed more comments

test/tests/functional/pbs_job_array.py
bhroam: I understand that.  Not trying preemption when the limit is hit is an optimization.  I wouldn't test for it.  Also, this message might go away when you move the code into the switch statement.

arungrover: Since this is a different kind of limit for which we have special code written to not try preemption, I thought we should add a test for it. I did remove the check for the log message while addressing your review comment.
I'd like to know what @tpai26 and @vstumpf think about it. If there is a consensus that the test is unnecessary, I'll remove it.

vstumpf (approved):
LGTM

src/include/pbs_error.h
```diff
@@ -286,6 +286,7 @@ extern "C" {
 #define PBSE_NODE_BUSY	15227		 /* Node is busy */
 #define PBSE_DEFAULT_PARTITION 15228	/* Default partition name is not allowed */
 #define PBSE_HISTDEPEND  15229		/* Finished job did not satisfy dependency */
+#define PBSE_NOTARRAY  15230		/* Not an array job */
```
bhroam, line 289: Do we want this or just use PBSE_BADATVAL?

src/lib/Liblog/pbs_messages.c
```diff
@@ -419,6 +419,7 @@ char *msg_node_busy = "Node is busy";
 char *msg_default_partition = "Default partition name is not allowed";
 char *msg_depend_runone = "Job deleted, a dependent job ran";
 char *msg_histdepend = "Finished job did not satisfy dependency";
+char *msg_notarray = "Not an array job";
```
bhroam, line 422: How about "Attribute has to be set on an array job"?

src/server/svr_jobfunc.c
```diff
@@ -5671,3 +5671,28 @@ recreate_exec_vnode(job *pjob, char *vnodelist, char *keep_select, char *err_msg
...
+ *			   It verifies that the attribute is being set only on array jobs.
+ *
+ * @param[in]	pattr	-	attribute structure
+ * @param[in]	pobject	-	job object
+ * @param[in]	actmode	-	action mode
+ */
+int
+action_max_run_subjobs(attribute *pattr, void *pobject, int actmode)
+{
+	job * pjob = (job*)pobject;
```
bhroam, line 5687: Attach the * to pjob

src/server/svr_jobfunc.c
```diff
@@ -5671,3 +5671,28 @@ recreate_exec_vnode(job *pjob, char *vnodelist, char *keep_select, char *err_msg
...
+	int jtype;
+
+	if (pjob == NULL)
+		return PBSE_INTERNAL;
+
+	jtype = is_job_array(pjob->ji_qs.ji_jobid);
```
bhroam, line 5693: Don't you need to check for NEW/ALTER?

test/tests/functional/pbs_job_array.py
```diff
@@ -982,3 +987,32 @@ def test_max_run_subjobs_eligible_time(self):
...
+        Test that setting max_run_subjobs on non-array jobs is rejected.
+        """
+
+        qsub_cmd = os.path.join(self.server.pbs_conf['PBS_EXEC'],
+                                'bin', 'qsub')
+
+        cmd = [qsub_cmd, '-Wmax_run_subjobs=4', '--', '/bin/sleep 100']
+        rv = self.du.run_cmd(self.server.hostname, cmd=cmd)
+        self.assertNotEqual(rv['rc'], 0, 'qsub must fail')
+        self.assertEqual(rv['err'][0], "qsub: Not an array job")
```
bhroam, line 1009: You should be able to do this self.server.submit()

test/tests/functional/pbs_job_array.py
```diff
@@ -982,3 +987,32 @@ def test_max_run_subjobs_eligible_time(self):
...
+
+        qsub_cmd = os.path.join(self.server.pbs_conf['PBS_EXEC'],
+                                'bin', 'qsub')
+
+        cmd = [qsub_cmd, '-J1-4%2', '-Wmax_run_subjobs=4', '--',
+               '/bin/sleep 100']
+        rv = self.du.run_cmd(self.server.hostname, cmd=cmd)
+        self.assertNotEqual(rv['rc'], 0, 'qsub must fail')
+        msg = "qsub: multiple max_run_subjobs values found"
+        self.assertEqual(rv['err'][0], msg)
```
bhroam, line 1025: You can do this with self.server.submit().  You do array_indices_submitted = '1-4%2'

bhroam (changes requested):

src/include/pbs_error.h
arungrover: To me, PBSE_BADATVAL seems unrelated to this particular error. It's not that attribute as been given a bad value but its that the attribute isn't related to the object being created.

src/server/svr_jobfunc.c
arungrover: It does not matter, action function must always reject if the attribute is being set on a normal job.

test/tests/functional/pbs_job_array.py
arungrover: I think for this, it has to come through qsub because the error check is in qsub command itself.

test/tests/functional/pbs_job_array.py
arungrover: Sure, I'll change it

src/server/svr_jobfunc.c
bhroam: What about job recovery?  What happens if we reject there?  Does the job get deleted?

arungrover:
Pushed more changes and resolved the conflict. Please review the last two commits, those are new.
[test_array.txt](https://github.com/openpbs/openpbs/files/5272796/test_array.txt)

vstumpf (approved):
Still looks good to me

bhroam (commented):
LGTM

bhroam (approved):
LGTM

PR #2046, opened 2020-09-25
Fix server_state not changing when scheduling

vstumpf:
#### Describe Bug or Feature
Server state would not change to Scheduling when the server would be scheduling

#### Describe Your Change
Set the modified bit and save to the database.

[after-fix.txt](https://github.com/openpbs/openpbs/files/5285621/after-fix.txt)
[before-fix.txt](https://github.com/openpbs/openpbs/files/5285622/before-fix.txt)

### DIFF
--- a/src/server/run_sched.c
+++ b/src/server/run_sched.c
@@ -228,7 +228,7 @@ set_sched_state(pbs_sched *psched, char *state) {
 	if (psched == NULL)
 		return;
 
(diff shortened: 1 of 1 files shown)


bhroam (approved):
LGTM

bhroam (approved):

arungrover (approved):
LGTM

PR #2048, opened 2020-09-28
fixing a minor bug in pbs_logutils

agrawalravi90:
#### Describe Bug or Feature

With https://github.com/openpbs/openpbs/pull/2016 server now prints letter job states instead of numerics. But I changed pbs_logutils to add the wrong job state letter for a particular log message. This just corrects that.

#### Describe Your Change

Changed state letter from R to Q in a server log message matching inside pbs_logutils.py

Existing tests passing should be good enough for this.


bhroam (approved):
LGTM

arungrover (approved):

agrawalravi90:
@arungrover and @bhroam thanks!!

PR #2049, opened 2020-09-28
Fix JSON errors

bayucan:
#### Describe Bug or Feature

The -Fjson feature in qstat and pbsnodes are reporting JSON errors when some conflicting substrings are read from PBS.
* If the following environment is passed to PBS, the resulting json output is an error:
```
export MYVAR=\'\"asads\"\'
qsub -V -- /bin/sleep 30
qstat -f shows:
MYVAR=\'\"asads\"\'
qstat -f -Fjson outputs:
"MYVAR":"\'\"asads\"\'"
but the \' is an error in JSON when fed to json.loads(). Fix is to drop the unnecessary \ (backslash):
"MYVAR":"'\"asads\"'"
```
* Similarly, if that \ is escaped, as in:
```
export MYVAR2=/home/pbstest01/Mo\\\'
echo $MYVAR2
/home/pbstest01/Mo\'
qsub -V -- /bin/sleep 30
qstat -f shows:
MYVAR2=/home/pbstest01/Mo\\\'
but qstat -f -Fjson would output:
"MYVAR2":"/home/pbstest01/Mo\\\'",
When fed to json.loads(), JSON error is returned for that \'. Fix is to drop that un-escaped \ (backslash).
"MYVAR2":"/home/pbsetst01/Mo\\'"
```
* If there's a comma in the environment variable value, PBS stores it with a backslash since comma has a special meaning in PBS. Comma is the delimiter  when PBS stores environment variables.
```
export MYVAR2="ab,c"
qsub -V -- /bin/sleep 30
qstat -f would show:
MYVAR="ab\,c"
qstat -f -Fjson would output:
"MYVAR"="ab\,c"
which would produce an error by json.loads() because of \, . Fix is to drop the unescaped \ (backslash).
"MYVAR":"ab,c"
```
* If a job is submitted with an environment variable value with leading 0s that is not all 000 or is part of a decimal number (i.e. 0.1020), the resulting int from json  would produce an error. This is because Python3 would not recognize such a value as int.
```
qsub -v foo=005 --/bin/sleep 30
qstat -f shows:
foo=005
qstat -f -Fjson outputs:
"foo":005
```
Feeding the above data to json.loads() produces JSON error as 005 is not recognized as an int in Python3. Fix is to detect the situation and output as a string value:
```
"foo":"005"
```
* The following sequence of steps would produce an attribute value of pset="" and would produce a bad json output:
```
Create a string array resource: qmgr -c 'c r iru type=string_array, flag=h'
Set a node group key: qmgr -c 's s node_group_key = iru'
Enable node grouping: qmgr -c 's s node_group_enable=True'
Submit a job which runs on a node where iru is not set.
Look at the qstat -f output:
qstat -f| grep pset:
pset = iru=""
Look at the qstat -f -F json output:
"pset":"iru="""
When fed to json.loads(), this results in an error due to the unescaped double quote ("). Fix is to escape
the embedded double quotes:
"pset":"\"\""
```
* If a node is given a comment with only whitespace characters, this produces a bad json output:
```
qmgr -c "set node nodeA comment='   '"
pbsndnodes -av
comment =   
pbsnodes -av -Fjson produces:
"comment": 
which is an error when fed to json.loads(). Fix is to show the whitespace values:
"comment":"   "
```

[ptl.qstat_formats_special_env.FAIL.txt](https://github.com/openpbs/openpbs/files/5295314/ptl.qstat_formats_special_env.FAIL.txt)

[ptl.qstat_formats_special_env.PASS.txt](https://github.com/openpbs/openpbs/files/5295315/ptl.qstat_formats_special_env.PASS.txt)
[ptl.json_nodes.FAIL.txt](https://github.com/openpbs/openpbs/files/5295318/ptl.json_nodes.FAIL.txt)
[ptl.json_nodes.PASS.txt](https://github.com/openpbs/openpbs/files/5295320/ptl.json_nodes.PASS.txt)
[ptl.qstat_formats.full2.txt](https://github.com/openpbs/openpbs/files/5313957/ptl.qstat_formats.full2.txt)
 [ptl.nodes_json.full2.txt](https://github.com/openpbs/openpbs/files/5313958/ptl.nodes_json.full2.txt)
 [ptl.pbs_nodes_json2.PASS.txt](https://github.com/openpbs/openpbs/files/5313959/ptl.pbs_nodes_json2.PASS.txt)
 [ptl.qstat_formats2.PASS.txt](https://github.com/openpbs/openpbs/files/5313960/ptl.qstat_formats2.PASS.txt)

The following testsuite have some json-related test cases:
[ptl.json_nodes_full.txt](https://github.com/openpbs/openpbs/files/5295324/ptl.json_nodes_full.txt)
[ptl.qstat_formats_full.txt](https://github.com/openpbs/openpbs/files/5295325/ptl.qstat_formats_full.txt)
[ptl.nonprint_chars_full.txt](https://github.com/openpbs/openpbs/files/5295327/ptl.nonprint_chars_full.txt)
(NOTE: The two test failures in nonprint_chars* are known test issues and is not related to this bug fix.)
[ptl.snapshot_json_full.txt](https://github.com/openpbs/openpbs/files/5295331/ptl.snapshot_json_full.txt)
[ptl.json_report.txt](https://github.com/openpbs/openpbs/files/5295332/ptl.json_report.txt)

Valgrind output of qstat, pbsnodes:
[valgrind.qstat_formats.txt](https://github.com/openpbs/openpbs/files/5295336/valgrind.qstat_formats.txt)
[valgrind.json_nodes.txt](https://github.com/openpbs/openpbs/files/5295337/valgrind.json_nodes.txt)

### DIFF
--- a/src/lib/Libcmds/pbs_json.c
+++ b/src/lib/Libcmds/pbs_json.c
@@ -108,6 +108,20 @@ link_node(JsonNode *node) {
  * @retval JSON conforming string   success
  * @retval NULL                     error
  *
+ * @note
+ * - If a string value contains one of the following:
+ * 	\'
+ *	\,
+ * and the \ is not escaped with another backslash, then drop that \.
+ * Otherwise, both forms produce JSON error.
+ * For instance:
+ *	\'     -> '
+ *	\\'    -> \\'
+ *	\\\'   -> \\'
+ *	\\\\'  -> \\\\'
+ *	\\\\\' -> \\\\'
+ * - If a string value contains " (embedded double quote) and it's not escaped
+ *   (not preceded by a backslash), then escape it, resulting in \".
  */
 char*
 strdup_escape(JsonEscapeType esc_type, const char *str)
@@ -116,7 +130,10 @@ strdup_escape(JsonEscapeType esc_type, const char *str)
 	int       len = 0;
 	char      *temp = NULL;
 	char      *buf = NULL;
+	const char *bufstr = NULL;
+	const char *orig_str = NULL;
 	if (str != NULL) {
+		orig_str = str;
 		i = 0;
 		len = MAXBUFLEN;
 		buf = (char *) malloc(len);
@@ -151,6 +168,11 @@ strdup_escape(JsonEscapeType esc_type, const char *str)
 				break;
 			case '"':
 				if (esc_type == JSON_ESCAPE) {
+					bufstr = str;
+					bufstr--;
+					if ((bufstr >= orig_str) && (*bufstr != '\\')) {
+						buf[i++] = '\\';
+					}
 					buf[i++] = *str++;
 					break;
 				} /* else JSON_FULLESCAPE */
@@ -160,7 +182,24 @@ strdup_escape(JsonEscapeType esc_type, const char *str)
 				break;
 			case '\\':
 				if (esc_type == JSON_ESCAPE) {
-					buf[i++] = *str++;
+					bufstr = str + 1;
+					if (*bufstr && ((*bufstr == '\'') || (*bufstr == ','))) {
+						int slash_ct = 0;
+						bufstr--;
+						while ((bufstr >= orig_str) && (*bufstr == '\\')) {
+							slash_ct++;
+							bufstr--;
+						}
+						/* detected balanced escaping (even # slashes) */
+						if ((slash_ct % 2) == 0) {
+							buf[i++] = *str++;
+						} else {
+							str++;
+						}
+
+					} else {
+						buf[i++] = *str++;
+					}
 					break;
 				} /* else JSON_FULLESCAPE */
 				buf[i++] = '\\';
@@ -226,6 +265,30 @@ free_json_node_list()
 
 /**
  * @brief
+ *	Determines if 'str' contains only white-space characters.
+ *
+ * @param[in] str - string to check
+ *
+ * @return	int
+ * @retval	1	if 'str' contains only white-space characters.
+ * @retval	0	error or otherwise.
+ *
+ */
+static int
+whitespace_only(char *str) {
+
+	if (str == NULL)
+		return (0);
+
+	while ((*str != '\0') && isspace(*str))
+		str++;
+	if (*str == '\0')
+		return (1);
+	return (0);
+}
+
+/**
+ * @brief
  *	add node to json list
  *
  * @param[in] ntype - node type
@@ -244,6 +307,7 @@ add_json_node(JsonNodeType ntype, JsonValueType vtype, JsonEscapeType esc_type, 
 	char	  *ptr = NULL;
 	char 	  *pc  = NULL;
 	JsonNode  *node = NULL;
+	int	  value_is_whitespace = 0;
 
 	node = create_json_node();
 	if (node == NULL) {
@@ -260,15 +324,22 @@ add_json_node(JsonNodeType ntype, JsonValueType vtype, JsonEscapeType esc_type, 
 		}
 		node->key = ptr;
 	}
-	if (vtype == JSON_NULL && value != NULL) {
-		(void)strtod(value, &pc);
+	value_is_whitespace = whitespace_only(value);
+	if (vtype == JSON_NULL && value != NULL && !value_is_whitespace) {
+		double val;
+		val = strtod(value, &pc);
 		while (pc) {
 			if (isspace(*pc))
 				pc++;
 			else
 				break;
 		}
-		if (strcmp(pc, "") == 0) {
+		/* In Python3, value with leading zeroes is
+		 * represented as a string, unless all zeroes (00000...)
+		 * or is part of a decimal number < 1 (0.0001 ... 0.99999).
+		 */
+		if ((strcmp(pc, "") == 0) &&
+			((*(char *)value != '0') || (val < 1))) {
 			node->value_type = JSON_NUMERIC;
 			ptr = strdup(value);
 			if (ptr == NULL) {
@@ -276,8 +347,11 @@ add_json_node(JsonNodeType ntype, JsonValueType vtype, JsonEscapeType esc_type, 
 				return NULL;
 			}
 			node->value.string = ptr;
-		} else
+		} else {
 			node->value_type = JSON_STRING;
+		}
+	} else if (value_is_whitespace) {
+		node->value_type = JSON_STRING;
 	} else {
 		node->value_type = vtype;
 		if (node->value_type == JSON_INT)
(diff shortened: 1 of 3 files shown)


src/lib/Libcmds/pbs_json.c
```diff
@@ -151,6 +168,11 @@ strdup_escape(JsonEscapeType esc_type, const char *str)
 				break;
 			case '"':
 				if (esc_type == JSON_ESCAPE) {
+					bufstr = str;
+					bufstr--;
```
minghui-liu, line 172: Is there any risk of `bufstr` running out of boundary?

bayucan: @minghui-liu : I'm checking for that condition immediately after this, before looking at content of bufstr:
```
173               if ((bufstr >= orig_str) && (*bufstr != '\\')) {
174		          buf[i++] = '\\';
175		     }
```

src/lib/Libcmds/pbs_json.c
```diff
@@ -224,6 +263,30 @@ free_json_node_list()
...
+ *
+ * @param[in] str - string to check
+ *
+ * @return	int
+ * @retval	1	if 'str' contains only white-space characters.
+ * @retval	0	error or otherwise.
+ *
+ */
+static int
+whitespace_only(char *str) {
```
vstumpf, line 278: can you make this `const char *str`

bayucan: Sure

minghui-liu (approved):

test/tests/functional/pbs_qstat_formats.py
```diff
@@ -580,6 +580,20 @@ def test_qstat_json_valid_job_special_env(self):
         """
         os.environ["DOUBLEQUOTES"] = 'hi"ha'
         os.environ["REVERSESOLIDUS"] = r'hi\ha'
+        os.environ["MYVAR"] = """\'\"asads\"\'"""
+        os.environ["MYVAR2"] = """/home/pbstest01/Mo\\'"""
+        os.environ["FOO"] = """005"""
+        os.environ["MYVAR1"] = """\'"""
+        os.environ["MYVAR2"] = """\\'"""
+        os.environ["MYVAR3"] = """\\\\'"""
```
vstumpf, line 588: Can you test `"""\\\'"""`?
That's three backslashes followed by a single quote.

test/tests/functional/pbs_qstat_formats.py
```diff
@@ -593,6 +607,7 @@ def test_qstat_json_valid_job_special_env(self):
             ' -f -F json ' + str(jid)
         ret = self.du.run_cmd(self.server.hostname, cmd=qstat_cmd_json)
         qstat_out = "\n".join(ret['out'])
+        self.logger.info(qstat_out)
```
vstumpf, line 610: Do you still need this debug output?

test/tests/functional/pbs_nodes_json.py
```diff
@@ -126,3 +126,43 @@ def test_check_reverse_solidus_escape_json(self):
...
+                           'bin', 'pbsnodes') + ' -av -Fjson'
+        ret = self.du.run_cmd(self.server.hostname, cmd=cmd)
+        n_out = "\n".join(ret['out'])
+        self.logger.info(n_out)
+        try:
+            json.loads(n_out)
+        except ValueError:
+            self.assertFalse(True, "Json failed to load")
+
+    def test_empty_job_pset_json(self):
```
vstumpf, line 147: shouldn't this be under qstat_formats? It's a qstat test.

src/lib/Libcmds/pbs_json.c
bayucan: @vstumpf : I've made the change. I need a re-review and feel free to merge if this PR is acceptable to you. Thanks.

src/lib/Libcmds/pbs_json.c
```diff
@@ -160,7 +182,24 @@ strdup_escape(JsonEscapeType esc_type, const char *str)
...
+						/* detected balanced escaping (even # slashes) */
+						if ((slash_ct % 2) == 0) {
+							buf[i++] = *str++;
+						} else {
+							str++;
+						}
+
+					} else {
+						buf[i++] = *str++;
+					}
```
vstumpf, line 202: can we handle the backslashes in pairs?
if you encounter a backslash, check the next character: if it's a comma/quote, only add the comma/quote.
Otherwise, add the backslash and next character.
This seems it would be simpler

vstumpf (changes requested):

test/tests/functional/pbs_qstat_formats.py
bayucan: Sure.

test/tests/functional/pbs_qstat_formats.py
bayucan: It's very helpful to see the json output especially if the test fails.

test/tests/functional/pbs_nodes_json.py
bayucan: I was thinking since it's setting a node group key, I placed the test under pbs_nodes_json.py. But I think you're right, this should be in pbs_qstat_formats.py testsuite. I'll move it.

src/lib/Libcmds/pbs_json.c
bayucan: I can't do what you suggested since the function parses the input string character by character, and it will operate only the current character, and not the next character like if it's a comma/quote. So if it encounters a backslash, function is deciding to include the backslash, not to include (if followed by comma or quote where the preceding backslash is not escaped), or add an additional backslash. It's not operating on the next character...

test/tests/functional/pbs_qstat_formats.py
vstumpf: Can you put it in the except block then?

src/lib/Libcmds/pbs_json.c
vstumpf: The function takes a string and returns it, right?
I see the while loop checks `(*str)`, we increment str once in each case. I'm saying we increment it twice in this one.

bayucan: @vstumpf Got it. I'll adjust the fix.

test/tests/functional/pbs_qstat_formats.py
bayucan: Done.

minghui-liu (approved):

bayucan:
Thanks @minghui-liu . @vstumpf : this is ready for merge. Thanks.

PR #2050, opened 2020-09-29
Hiding set_attr_* calls behind job object calls

agrawalravi90:
#### Describe Bug or Feature

This is just refactoring. The attribute level functions should be hidden behind job level functions if working on a job object's attribute. So, here I'm just converting calls to set_attr_* with set_jattr_* calls. 

Since there's no functional change, existing tests passing should be good enough

### DIFF
--- a/src/include/job.h
+++ b/src/include/job.h
@@ -1048,6 +1048,7 @@ void set_job_substate(job *pjob, long val);
 int set_jattr_str_slim(job *pjob, int attr_idx, char *val, char *rscn);
 int set_jattr_l_slim(job *pjob, int attr_idx, long val, enum batch_op op);
 int set_jattr_b_slim(job *pjob, int attr_idx, long val, enum batch_op op);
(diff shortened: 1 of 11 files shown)


hirenvadalia:
@agrawalravi90 Shouldn't we need to do these changes in mom code also?

subhasisb:
Hey @raviv1995  i was checking all access to ji_wattr array (the job attribute variable). There are a lot of cases of direct access (reading) of this array, for example:
peh = &pjob->ji_wattr[(int)JOB_ATR_exec_vnode];

When i did a grep "ji_wattr[", i find 334 entries. 
Similarly, a grep of "nd_attr[" shows 149 direct accesses. 
I am sure attribute array of other objects have similar direct access.

Were you planning to cover these?

bhroam (approved):
LGTM

subhasisb (approved):

PR #2052, opened 2020-09-30
cleanup *e and *o files in user homedir

visheshh:
#### Describe Bug or Feature
Large number of *e and *o files in user home directory fills up the system after tests which may submit thousands of jobs.

#### Describe Your Change
Added cleanup_homedir function under class MoM and calling cleanup_homedir for all the PBS_USERS from tearDown.

### DIFF
--- a/test/fw/ptl/lib/pbs_testlib.py
+++ b/test/fw/ptl/lib/pbs_testlib.py
@@ -14107,6 +14107,31 @@ class MoM(PBSService):
                               self.hostname)
             raise AssertionError('cgroup subsystems not mounted')
 
+    def cleanup_homedir(self, user):
+        self.logger.info('Cleaning %s\'s home directory' % (str(user)))
+        runas = PbsUser.get_user(user)
+        hostname = self.hostname
+        ret = self.du.run_cmd(hostname, cmd=['echo', '$HOME'], sudo=True,
+                              runas=runas, logerr=False, as_script=True)
+        if ret['rc'] == 0:
+            path = ret['out'][0].strip()
+        else:
+            return None
+
+        ftd = []
+        files = self.du.listdir(hostname, path=path, runas=user)
+        bn = os.path.basename
+        ftd.extend([f for f in files if bn(f).startswith('PtlPbs')])
+        ftd.extend([f for f in files if bn(f).startswith('STDIN')])
+        ftd.extend([f for f in files if re.match('.*\.o.*', f)])
+        ftd.extend([f for f in files if re.match('.*\.e.*', f)])
+
+        if len(ftd) > 1000:
+            for i in range(0, len(ftd), 1000):
+                j = i + 1000
+                self.du.rm(hostname, path=ftd[i:j], runas=user,
+                           force=True, level=logging.DEBUG2)
+
 
 class Hook(PBSObject):
 
--- a/test/fw/ptl/utils/pbs_testsuite.py
+++ b/test/fw/ptl/utils/pbs_testsuite.py
@@ -1721,6 +1721,8 @@ class PBSTestSuite(unittest.TestCase):
 
         for mom in self.moms.values():
             mom.cleanup_files()
+            for u in PBS_USERS:
+                mom.cleanup_homedir(u.name)
 
         for sched in self.scheds:
             self.scheds[sched].cleanup_files()


test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -14107,6 +14107,31 @@ def enable_cgroup_cset(self):
...
+        else:
+            return None
+
+        ftd = []
+        files = self.du.listdir(hostname, path=path, runas=user)
+        bn = os.path.basename
+        ftd.extend([f for f in files if bn(f).startswith('PtlPbs')])
+        ftd.extend([f for f in files if bn(f).startswith('STDIN')])
+        ftd.extend([f for f in files if re.match('.*\.o.*', f)])
+        ftd.extend([f for f in files if re.match('.*\.e.*', f)])
```
bhroam, line 14127: The files starting with STDIN should be the .o and .e files.  Either turn the above 4 lines into one list comprehension with or's, or set() the output to remove the duplicate entries.  You also need to add '.oe' and '.eo' files for tests using -j.

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -1721,6 +1721,8 @@ def tearDown(self):
 
         for mom in self.moms.values():
             mom.cleanup_files()
+            for u in PBS_USERS:
+                mom.cleanup_homedir(u.name)
```
bhroam, line 1725: Should this be done in tearDown() or tearDownClass()?  Also, you str() the user in cleanup_homedir(), so you can pass 'u'

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -14107,6 +14107,31 @@ def enable_cgroup_cset(self):
                               self.hostname)
             raise AssertionError('cgroup subsystems not mounted')
 
+    def cleanup_homedir(self, user):
```
bhroam, line 14110: Add a spynx header for this function.

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -14107,6 +14107,31 @@ def enable_cgroup_cset(self):
...
+        hostname = self.hostname
+        ret = self.du.run_cmd(hostname, cmd=['echo', '$HOME'], sudo=True,
+                              runas=runas, logerr=False, as_script=True)
+        if ret['rc'] == 0:
+            path = ret['out'][0].strip()
+        else:
+            return None
+
+        ftd = []
+        files = self.du.listdir(hostname, path=path, runas=user)
```
bhroam, line 14122: Question, is listdir() smart enough to take a '~' ?

test/fw/ptl/lib/pbs_testlib.py
```diff
@@ -14107,6 +14107,31 @@ def enable_cgroup_cset(self):
...
+        bn = os.path.basename
+        ftd.extend([f for f in files if bn(f).startswith('PtlPbs')])
+        ftd.extend([f for f in files if bn(f).startswith('STDIN')])
+        ftd.extend([f for f in files if re.match('.*\.o.*', f)])
+        ftd.extend([f for f in files if re.match('.*\.e.*', f)])
+
+        if len(ftd) > 1000:
```
bhroam, line 14129: This removal of files is done in test_runner() in PbsTestSuite.  You should probably remove the duplicate code there if you are removing it here.

bhroam (changes requested):

test/fw/ptl/utils/pbs_testsuite.py
visheshh: I want this to be done in tearDown() to cleanup files after each test. Most performance tests submit more than thousands of jobs in each test.

sakshamgarg:
The output and error files are staged out to the host from where job submission happens. Since in PTL, we submit the jobs from the host where pbs-server is running, I think we should cleanup the user's homedir in server host instead of mom host. So we can move your changes from MoM class to Server class. Correct me if I am understanding it wrong.

visheshh:
> 
> 
> The output and error files are staged out to the host from where job submission happens. Since in PTL, we submit the jobs from the host where pbs-server is running, I think we should cleanup the user's homedir in server host instead of mom host. So we can move your changes from MoM class to Server class. Correct me if I am understanding it wrong. What do you and others think?

Thanks Saksham, You are correct. The files will be staged back. So i will move the function to Server and have an optional hostname if in case we need to pass client hostname.

test/fw/ptl/lib/pbs_testlib.py
hirenvadalia: I think right place for this clean code is in runner plugin where we already have other cleanup code. https://github.com/openpbs/openpbs/blob/master/test/fw/ptl/utils/plugins/ptl_test_runner.py#L987

hirenvadalia (changes requested):

PR #2053, opened 2020-10-01
Increase job run time so slower machines catch the running state

lisa-altair:
#### Describe Bug or Feature

On some very slow machines, or if the machine has a pause while running the expect(), a test might fail due not finding the job in the expected state.  
For example:
ptl.lib.pbs_testlib.PtlExpectError: rc=1, rv=False, msg=expected on server xarm-05-c8p1: no data for job_state = R && substate = 42 job 12.xarm-05-c8p1 attempt: 180

The longest pause I saw was for 6 seconds.

#### Describe Your Change

I made the job run for 10 seconds in hopes that it is a long enough time to see the job in the expected state, but without adding too much time to running the test.

[after.txt](https://github.com/openpbs/openpbs/files/5308872/after.txt)


bayucan (approved):
Looks good.

vccardenas (approved):
Looks good to me.

PR #2054, opened 2020-10-01
removing the segmentation fault

Jitendra-Rajak:
#### Describe Bug or Feature
Server segmentation fault while trying to access a void pointer value.

#### Describe Your Change
Added a check before accessing the void pointer value, if the void pointer hold any value than only try to access. 

Logs attached to the ticket.

### DIFF
--- a/src/server/licensing_func.c
+++ b/src/server/licensing_func.c
@@ -544,8 +544,8 @@ license_one_node(pbsnode *pnode)
 				update_license_highuse();
 			} else {
 				add_to_unlicensed_node_list(pnode);
-				if (pnode->nd_attr[ND_ATR_LicenseInfo].at_val.at_long) {
-					licensing_control.licenses_total_needed += pnode->nd_attr[ND_ATR_LicenseInfo].at_val.at_long;
+				if (is_attr_set(&(pnode->nd_attr[ND_ATR_LicenseInfo]))) {
+					licensing_control.licenses_total_needed += get_attr_l(&(pnode->nd_attr[ND_ATR_LicenseInfo]));
 				}
 				if (get_more_licenses_task)
 					delete_task(get_more_licenses_task);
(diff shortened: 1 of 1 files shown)


src/server/licensing_func.c
```diff
@@ -544,6 +544,9 @@ license_one_node(pbsnode *pnode)
 				update_license_highuse();
 			} else {
 				add_to_unlicensed_node_list(pnode);
+				if (pnode->nd_attr[ND_ATR_LicenseInfo].at_val.at_long) {
+					licensing_control.licenses_total_needed += pnode->nd_attr[ND_ATR_LicenseInfo].at_val.at_long;
```
arungrover, line 548: Please use get_attr_l instead of accessing it directly.

src/server/licensing_func.c
```diff
@@ -694,8 +697,10 @@ init_licensing(struct work_task *ptask)
 		if (pbsndlist[i]->nd_attr[ND_ATR_LicenseInfo].at_val.at_long) {
 			licensing_control.licenses_total_needed += pbsndlist[i]->nd_attr[ND_ATR_LicenseInfo].at_val.at_long;
 		} else {
-			set_node_lic_info_attr(pbsndlist[i]);
-			licensing_control.licenses_total_needed += pbsndlist[i]->nd_attr[ND_ATR_LicenseInfo].at_val.at_long;
+			if (pbsndlist[i]->nd_lic_info != NULL) {
+				set_node_lic_info_attr(pbsndlist[i]);
+				licensing_control.licenses_total_needed += pbsndlist[i]->nd_attr[ND_ATR_LicenseInfo].at_val.at_long;
```
arungrover, line 702: Please use get_attr_l instead of accessing it directly.

arungrover (changes requested):

src/server/licensing_func.c
Jitendra-Rajak: made the changes please have a look @arungrover

src/server/licensing_func.c
Jitendra-Rajak: done

arungrover (approved):

subhasisb (approved):

PR #2056, opened 2020-10-02
fix pbs_rsub segfault while parsing hosts

vchlum:
#### Describe Bug or Feature
Submitting maintenance, pbs_rsub can segfault. The reason is the array **hostp initializes only the last item (to NULL) and it is supposed the array will be full in the end. The problem is the array is checked for duplicate items while an item is not NULL before. An inaccessible part of memory can occur within the array like this: 

```
311					for (; *hostp; hostp++) {
(gdb) 
312						if (strcmp(*hostp, argv[optind]) == 0) {
(gdb) print *hostp
$16 = 0x10271c25 <error: Cannot access memory at address 0x10271c25>
(gdb) n

Program received signal SIGSEGV, Segmentation fault.
__strcmp_avx2 () at ../sysdeps/x86_64/multiarch/strcmp-avx2.S:101
101	../sysdeps/x86_64/multiarch/strcmp-avx2.S: No such file or directory.
```

#### Describe Your Change
Initialize the first item in the array **hostp to NULL, and once a valid item is added to the array the subsequent item is always initialized to NULL. This way the array always ends with NULL.

This bug can not be reliable reproduced because the inaccessible part of memory is not always present in the array. The bigger the array is, the bigger chance to reproduce. So, numerous random hostnames are conveyed to pbs_rsub in the PTL test.
[ptl_maintenance.log](https://github.com/openpbs/openpbs/files/5316307/ptl_maintenance.log)

### DIFF
--- a/test/tests/functional/pbs_maintenance_reservations.py
+++ b/test/tests/functional/pbs_maintenance_reservations.py
@@ -694,3 +694,25 @@ class TestMaintenanceReservations(TestFunctional):
         a = {'comment': 'Can Never Run: Reservation is in an invalid state',
              'job_state': 'Q'}
         self.server.expect(JOB, a, id=jid)
+
+    def test_maintenance_parse_numerous_hosts(self):
+        """
+        Test if the parsing of numerous hosts succeed.
+        """
+        now = int(time.time())
+
+        a = {'reserve_start': now + 3600,
+             'reserve_end': now + 7200}
+        chars = 'abcdefghijklmnopqrstuvwxyz'
+        h = [''.join(random.choices(chars, k=8)) for i in range(200)]
+        r = Reservation(TEST_USER, attrs=a, hosts=h)
+
+        msg = ""
+
+        try:
+            self.server.submit(r)
+        except PbsSubmitError as err:
+            msg = err.msg[0].strip()
+
+        regex = "^pbs_rsub: Host with resources not found: .*"
+        self.assertTrue(re.search(regex, msg))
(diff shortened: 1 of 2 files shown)


test/tests/functional/pbs_maintenance_reservations.py
```diff
@@ -694,3 +694,25 @@ def test_maintenance_degrade_reservation_jobs_dont_run(self):
...
+
+        a = {'reserve_start': now + 3600,
+             'reserve_end': now + 7200}
+        chars = 'abcdefghijklmnopqrstuvwxyz'
+        h = [''.join(random.choices(chars, k=8)) for i in range(200)]
+        r = Reservation(TEST_USER, attrs=a, hosts=h)
+
+        msg = ""
+
+        try:
```
arungrover, line 712: If you are expecting an exception, please consider using "with assertRaises(PbsSubmitError) as" syntax.

arungrover (commented):
Small suggestion on the test, looks good otherwise.

test/tests/functional/pbs_maintenance_reservations.py
vchlum: Thank you @arungrover I applied your suggestion for the whole test suite.

vchlum:
New PTL log: 
[ptl_maintenance_output.txt](https://github.com/openpbs/openpbs/files/5325670/ptl_maintenance_output.txt)

arungrover (approved):
Thanks for making this change.
LGTM

bhroam (approved):
LGTM

PR #2057, opened 2020-10-03
Altering standing reservations do not affect future occurrences

vstumpf:
#### Describe Bug or Feature
If a standing reservation has its select altered, and then goes degraded, the scheduler will attempt to reconfirm the current and future occurrences with the existing altered select.  This will cause the future occurrences to be short nodes.

#### Describe Your Change
For future occurrences of standing reservations, use the saved standing reservation's select.

This also fixes a potential crash in the scheduler

[after-fix-all.txt](https://github.com/openpbs/openpbs/files/5329448/after-fix-all.txt)

[after-fix.txt](https://github.com/openpbs/openpbs/files/5320817/after-fix.txt)
[before_fix.txt](https://github.com/openpbs/openpbs/files/5320818/before_fix.txt)
[crash.txt](https://github.com/openpbs/openpbs/files/5320819/crash.txt)

### DIFF
--- a/src/scheduler/resv_info.c
+++ b/src/scheduler/resv_info.c
@@ -535,7 +535,7 @@ query_reservations(int pbs_sd, server_info *sinfo, struct batch_status *resvs)
 					 */
 					release_nodes(resresv_ocr);
 
-					if (resresv_ocr->resv->select_standing) {
+					if (resresv_ocr->resv->select_standing != NULL) {
 						free_selspec(resresv_ocr->select);
 						resresv_ocr->select = dup_selspec(resresv_ocr->resv->select_standing);
 					}
@@ -1166,7 +1166,7 @@ check_new_reservations(status *policy, int pbs_sd, resource_resv **resvs, server
 							nresv_copy = dup_resource_resv(nresv_copy, sinfo, NULL, err);
 							if (nresv_copy == NULL)
 								break;
-							if (nresv_copy->resv->select_standing) {
+							if (nresv_copy->resv->select_standing != NULL) {
 								free_selspec(nresv_copy->select);
 								nresv_copy->select = dup_selspec(nresv_copy->resv->select_standing);
 							}
--- a/src/scheduler/resv_info.c
+++ b/src/scheduler/resv_info.c
@@ -1164,6 +1164,9 @@ check_new_reservations(status *policy, int pbs_sd, resource_resv **resvs, server
 							nresv_copy = dup_resource_resv(nresv_copy, sinfo, NULL, err);
 							if (nresv_copy == NULL)
 								break;
+
+							free_selspec(nresv_copy->select);
+							nresv_copy->select = dup_selspec(nresv_copy->resv->select_standing);
 						}
 						/* Duplication deep-copies node info array. This array gets
 						 * overwritten and needs to be freed. This is an alternative
@@ -1172,8 +1175,6 @@ check_new_reservations(status *policy, int pbs_sd, resource_resv **resvs, server
 						 */
 						release_nodes(nresv_copy);
 
-						free_selspec(nresv_copy->select);
-						nresv_copy->select = dup_selspec(nresv_copy->resv->select_standing);
 						nresv_copy->resv->orig_nspec_arr = parse_execvnode(occr_execvnodes_arr[j], sinfo, nresv_copy->select);
 						nresv_copy->ninfo_arr = create_node_array_from_nspec(nresv_copy->nspec_arr);
 						nresv_copy->nspec_arr = combine_nspec_array(nresv_copy->resv->orig_nspec_arr);
--- a/src/scheduler/resv_info.c
+++ b/src/scheduler/resv_info.c
@@ -535,6 +535,9 @@ query_reservations(int pbs_sd, server_info *sinfo, struct batch_status *resvs)
 					 */
 					release_nodes(resresv_ocr);
 
+					free_selspec(resresv_ocr->select);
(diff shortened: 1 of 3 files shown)


test/tests/functional/pbs_ralter.py
```diff
@@ -2842,3 +2842,42 @@ def test_restart_revert(self):
...
+    def test_alter_degrade_reconfirm_standing(self):
+        """
+        Test that if a standing reservation is altered, degraded,
+        then reconfirmed, the reservation will use the original
+        select
+        """
+        duration = 60
+        offset = 60
+
+        self.server.manager(MGR_CMD_SET, SERVER, {'reserve_retry_time': 2})
```
bhroam, line 2855: Use 5s.  2s means we'll try and reconfirm the resv every 2s.  We want to make sure there is enough time to have the cycle.  Also you don't want the reservation to be reconfirmed too quickly or your expects will fail.

test/tests/functional/pbs_ralter.py
```diff
@@ -2842,3 +2842,42 @@ def test_restart_revert(self):
...
+
+        rid, start, end = self.submit_and_confirm_reservation(
+            offset, duration, standing=True, select="2:ncpus=2")
+
+        confirmed = {'reserve_state': (MATCH_RE, 'RESV_CONFIRMED|2')}
+        self.server.expect(RESV, confirmed, id=rid)
+
+        self.server.alterresv(rid, {ATTR_l: "select=1:ncpus=2"})
+        self.server.expect(RESV, confirmed, id=rid)
```
bhroam, line 2864: Why call out to alterresv() directly and not use alter_a_reservation()?  It handles making sure it's confirmed properly.

test/tests/functional/pbs_ralter.py
```diff
@@ -2842,3 +2842,42 @@ def test_restart_revert(self):
...
+
+        self.server.status(RESV, id=rid)
+        resv_node = self.server.reservations[rid].get_vnodes()[0]
+
+        offline = {'state': 'offline'}
+        self.server.manager(MGR_CMD_SET, NODE, offline, id=resv_node)
+        degraded = {'reserve_state': (MATCH_RE, 'RESV_DEGRADED|10')}
+        self.server.expect(RESV, degraded, id=rid)
+        self.server.manager(MGR_CMD_SET, SERVER, {'scheduling': True})
```
bhroam, line 2873: Why run a cycle?  The server will run one when the retry_time comes up.

test/tests/functional/pbs_ralter.py
```diff
@@ -2842,3 +2842,42 @@ def test_restart_revert(self):
...
+        self.server.expect(RESV, confirmed, id=rid)
+
+        stat = self.server.status(RESV, id=rid)[0]
+        resvnodes = stat['resv_nodes']
+        self.assertEquals(1, len(resvnodes.split('+')))
```
bhroam, line 2878: Instead of this assert, do a get_vnodes() and check that the node is different than the first time you queried it.

bhroam (changes requested):

test/tests/functional/pbs_ralter.py
vstumpf: because i didn't see alter_a_reservation needs to pass start and end time, i thought it would just be easier to alter the reservation without the helper function.

test/tests/functional/pbs_ralter.py
vstumpf: i forgot to turn off scheduling before offlining the node. This will prevent reserve_retry_time being too short.

test/tests/functional/pbs_ralter.py
vstumpf: I can do both, I like having this one to say that there's only 1 chunk.

test/tests/functional/pbs_ralter.py
bhroam: You have the start and end from submit_and_confirm_reservation().  I'd use the helper function.  It does a lot of stuff that is good to make sure the alter worked.

test/tests/functional/pbs_ralter.py
```diff
@@ -2842,3 +2842,42 @@ def test_restart_revert(self):
...
+        """
+        duration = 60
+        offset = 60
+
+        self.server.manager(MGR_CMD_SET, SERVER, {'reserve_retry_time': 2})
+
+        rid, start, end = self.submit_and_confirm_reservation(
+            offset, duration, standing=True, select="2:ncpus=2")
+
+        confirmed = {'reserve_state': (MATCH_RE, 'RESV_CONFIRMED|2')}
```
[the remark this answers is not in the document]
vstumpf: haha, of course

vstumpf:
running tests right now, pls no merge

src/scheduler/resv_info.c
```diff
@@ -535,8 +535,10 @@ query_reservations(int pbs_sd, server_info *sinfo, struct batch_status *resvs)
 					 */
 					release_nodes(resresv_ocr);
 
-					free_selspec(resresv_ocr->select);
-					resresv_ocr->select = dup_selspec(resresv_ocr->resv->select_standing);
+					if (resresv_ocr->resv->select_standing) {
```
bhroam, line 538: Add != NULL (here and below)

vstumpf:
I've attached logs from running hte entire ralter test suite
It's ready for review

bhroam (approved):
LGTM

arungrover (approved):

PR #2058, opened 2020-10-05
Fixing a printjob bug caused by recent refactoring for job state

agrawalravi90:
#### Describe Bug or Feature

After the refactoring change for job attributes, the server now logs job states in state characters instead of numerics, and we removed state and substate from job object's quick save area. This caused printjob to print the wrong output as it reads the job file as it is, tries to print the job's quick save area, which earlier had the state and substate, before reading the attribute list and then prints the attribute list. Since we removed state and substate, the summary prints bogus (null) values for state and substate. This causes cgroups hook to fail as well as it reads the output of printjob.

Fix:
This calls for a refactoring of printjob. I had to read the attribute list completely before printing the job summary because i needed the state and substate from the attr list.

**Printjob header, Before fix:**
```
---------------------------------------------------
jobid:      26.stblr3
---------------------------------------------------
state:                     0x
substate:             0x0 (0)
```

**After fix:**
```
---------------------------------------------------
jobid:	1003.pbs2
---------------------------------------------------
state:		R
substate:	0x42 (42)
```

### DIFF
--- a/src/hooks/cgroups/pbs_cgroups.PY
+++ b/src/hooks/cgroups/pbs_cgroups.PY
@@ -670,7 +670,7 @@ def job_is_running(jobid):
     Returns True if job shows a running state and substate
     """
     jobinfo = printjob_info(jobid)
-    if 'state' in jobinfo and jobinfo['state'] != 4:
+    if 'state' in jobinfo and jobinfo['state'] != 'R':
         return False
     if 'substate' in jobinfo:
         return jobinfo['substate'] == 42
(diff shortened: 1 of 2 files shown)


bhroam (approved):
LGTM

arungrover:
@agrawalravi90 does this issue also affect upgrades?

agrawalravi90:
@arungrover interesting point, i had not thought about that. Let me try that out, install a newer version after running jobs from the old one.

agrawalravi90:
@arungrover  so you are correct, I tried installing the version with my job state and substate changes and pbs_mom didn't identify the job which was created by the previous version. This is going to be problematic for upgrades. I'll file a ticket for it and try to think of ways to fix it. In the meantime, I believe this particular PR can still go in, please let me know what you think.

arungrover (approved):

PR #2063, opened 2020-10-08
 Qdel optimization for more number of jobs

bremanandjk:
#### Describe Bug or Feature
Objective: Since qdel is taking a huge amount of time to delete the tens of thousands of jobs in PBS_SERVER, there is a need for optimization to improve its performance. 

On analysis, observed that the qdel client is iterating over all the jobs and sending the pbs_deljob IFL in a serial fashion, there is a conspicuous design problem. However this approach works well with a minimum set of jobs, It is not a beneficial solution with a large number of jobs, say 100k jobs.

#### Describe Your Change

* Introduced new IFL pbs_deljoblist().
* Sending all jobs in a single batch request using new IFL API and processing them together in server.
* Since the server is processing a lot of jobs in a single request, the server would collect all the replies and send in delete status records. 
* This multiple jobid transmission would reduce the network latency between client and server by eliminating pbs_connect/pbs_disconnect. 

#### Link to Design Doc

https://openpbs.atlassian.net/wiki/spaces/PD/pages/2184544264/Qdel+optimization+for+a+huge+number+of+jobs

Description: Rerun All Tests From #4497
Profile: regression
|--|--|--|--|--|--|--|
|4513|3848|1|0|0|0|3847|

In mainline, qdel took 13m11.304s for deleting 100k jobs and after enhancement, the qdel took 25 seconds.

[root@blrentperf04 pbspro]# qstat |wc
 100002  600014 8100140
[root@blrentperf04 pbspro]# time qselect | xargs qdel

real	0m25.663s
user	0m0.404s
sys	0m0.166s
[root@blrentperf04 pbspro]#


src/lib/Libifl/int_delete2.c
```diff
@@ -0,0 +1,128 @@
...
+	int i;
+	struct batch_reply *reply;
+	struct batch_status *rbsp = NULL;
+
+	/* initialize the thread context data, if not initialized */
+	if (pbs_client_thread_init_thread_context() != 0)
+		return NULL;
+
+	/* verify the object name if creating a new one */
+	if (command == MGR_CMD_CREATE)
```
subhasisb, line 84: if you have a separate API for delete, then why do you need parameter command? And how can MGR_CMD_CREATE ever land up here?

src/lib/Libifl/int_delete2.c
```diff
@@ -0,0 +1,128 @@
...
+		command, aoplp)) != 0)
+		return NULL;
+
+	/* lock pthread mutex here for this connection */
+	/* blocking call, waits for mutex release */
+	if (pbs_client_thread_lock_connection(c) != 0)
+		return NULL;
+
+	/* send the manage request */
+	i = PBSD_mgr_put(c, function, command, objtype, objname, aoplp, extend, PROT_TCP, NULL);
```
subhasisb, line 99: if u are using PBSD_mgr_put itself, you could use the same old IFL API calls, and handle the case of two different return types there

subhasisb (dismissed):
I am not sure you need a separate IFL call, rather it seems a variant reply type could suffice. In case multiple jobids are sent, u could return a different reply choice - for which u can use a different reply structure, of course

bremanandjk:
> I am not sure you need a separate IFL call, rather it seems a variant reply type could suffice. In case multiple jobids are sent, u could return a different reply choice - for which u can use a different reply structure, of course

Some of my thoughts on why I have created new IFL for multiple job deletion. 

1) Supporting backward compatibility, 
2) Eliminating a lot of changes in Sched specific code while preempting the running job. (Sending the del request and process the reply)
3) Even reservation deletion causes to create multiple delete job requests, need modifications there.
4) By having new IFL, we could maintain the existing IFL calls (pbs_deljob) between Server->Mom for job deletion. else it would take time to implement to support these functionalities.
5) Changing the existing IFL would impact the IFL wraps and, also welcomes more changes on hooks in server and mom. (pbs_ifl_wrap.c & pbs_tclWrap.c)
6) PTL and Perl framework also needs to be addressed to accommodate IFL signature change. 
 
It would be a large refactoring. I was doing those and it was taking the time. So, I thought to have a new IFL, which is used only in qdel command, and it won't have any impacts on other areas and would be less time to achieve the goal for the fastest qdel with multiple jobs.

subhasisb:
> > I am not sure you need a separate IFL call, rather it seems a variant reply type could suffice. In case multiple jobids are sent, u could return a different reply choice - for which u can use a different reply structure, of course
> 
> Some of my thoughts on why I have created new IFL for multiple job deletion.
> 
>     1. Supporting backward compatibility,
> 
>     2. Eliminating a lot of changes in Sched specific code while preempting the running job. (Sending the del request and process the reply)
> 
>     3. Even reservation deletion causes to create multiple delete job requests, need modifications there.
> 
>     4. By having new IFL, we could maintain the existing IFL calls (pbs_deljob) between Server->Mom for job deletion. else it would take time to implement to support these functionalities.
> 
>     5. Changing the existing IFL would impact the IFL wraps and, also welcomes more changes on hooks in server and mom. (pbs_ifl_wrap.c & pbs_tclWrap.c)
> 
>     6. PTL and Perl framework also needs to be addressed to accommodate IFL signature change.
> 
> 
> It would be a large refactoring. I was doing those and it was taking the time. So, I thought to have a new IFL, which is used only in qdel command, and it won't have any impacts on other areas and would be less time to achieve the goal for the fastest qdel with multiple jobs.

1. Backward compatibility should not get hurt. Older clients will not know to pass multiple job-ids or reservation ids, so they will just get the older type return choice, so nothing will fall apart.

2. Need to understand/see what this is

3. For reservations, pass only one id at a time, so nothing dealing with reservations need to change even if the api has internally changed.

4. modified pbs_deljob() will have zero impact between server and mom, since it will pass one jobid at a time to the mom. So, server will act as if it is an older client (to mom) which does not know that the api actually support multiple jobids. So, as far as server-mom communication is concerned, nothing needs to be changed.

5. Hooks can behave as older clients (again not know that the api was augmented to take on multiple-ids, if passed). So hooks wont need change. I think u are confused about the wrappers part. Since the API itself is not changing the wrappers do not need to change. For new clients, only need is to process a updated return value structure/array.

I feel the change is smaller than using a new API

subhasisb:
> > 1. Eliminating a lot of changes in Sched specific code while preempting the running job. (Sending the del request and process the reply)
> 
> @subhasisb the scheduler uses pbs_preemption() to preempt jobs. It does not call pbs_deljob() itself. The server receives the preemption request, and will create an internal batch deljob request for each job that is being deleted.

Thanks @bhroam - so the scheduler would not have needed any change anyway

src/cmds/qdel.c
```diff
@@ -167,19 +171,41 @@ char **envp;
...
 		pbs_strncpy(job_id, argv[optind], sizeof(job_id));
 		if (get_server(job_id, job_id_out, server_out)) {
 			fprintf(stderr, "qdel: illegally formed job identifier: %s\n", job_id);
 			any_failed = 1;
 			continue;
 		}
-cnt:
+		
+		if (count == 0)
+			memset(job_list, 0, job_list_size);
```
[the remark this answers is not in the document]
bremanandjk: I have changed this approach, not using this heavy memset call.

src/lib/Libifl/int_delete2.c
bremanandjk: Got it, removed this unnecessary check.

src/lib/Libifl/int_delete2.c
bremanandjk: Since there is a difference in IFL signature, I have introduced a new IFL function namely PBSD_delete_put.

src/cmds/qdel.c
```diff
@@ -167,20 +240,33 @@ char **envp;
...
+		fprintf(stderr, "qdel: unable to allocate memory.\n");
+		exit(2);
+	}
+		
 	for (; optind < argc; optind++) {
-		int connect;
-		int stat=0;
-		int located = FALSE;
-
 		pbs_strncpy(job_id, argv[optind], sizeof(job_id));
```
hirenvadalia, line 253: Can't we argv[optind] directly in get_server()?

src/cmds/qdel.c
```diff
@@ -167,20 +240,33 @@ char **envp;
 		fprintf(stderr, "qdel: unable to initialize security library.\n");
 		exit(1);
 	}
-
+	
+	/* allocate enough memory to store array of server instances */
+	no_of_pbsservers = 10;
+	pbs_servers = (pbs_svr_t *)calloc(no_of_pbsservers, sizeof(pbs_svr_t));
```
hirenvadalia, line 246: You can use add_ptr_to_array() and remove this code.

src/cmds/qdel.c
```diff
@@ -241,25 +327,23 @@ char **envp;
 			pbs_strncpy(warg, warg1, sizeof(warg));
 		}
 
-		stat = pbs_deljob(connect, job_id_out, warg);
+		p_delstatus = pbs_deljoblist(connect, pbs_svr->jobids, warg);
+		
+		while (p_delstatus != NULL) {
+			if ((pbse_to_txt(p_delstatus->code) != NULL) && (p_delstatus->code != PBSE_HISTJOBDELETED)) {
```
hirenvadalia, line 333: store output of `pbse_to_txt()` in some variable and use in below like instead calling it again

src/cmds/qdel.c
```diff
@@ -241,25 +327,23 @@ char **envp;
...
-			if (locate_job(job_id_out, server_out, rmt_server)) {
-				pbs_disconnect(connect);
-				pbs_strncpy(server_out, rmt_server, sizeof(server_out));
-				goto cnt;
+			if (p_delstatus->code != PBSE_HISTJOBDELETED)
+				num_deleted++;
+			
+			if (p_delstatus->code == PBSE_UNKJOBID) {
+				if (locate_job(p_delstatus->name, server_out, rmt_server))
+					add_jobid_to_pbs_servers(rmt_server, rmt_server, &pbs_servers);
```
hirenvadalia, line 343: What will happen if rmt_server is server on which we already sent deljoblist request and finished its processing?

src/include/batch_request.h
```diff
@@ -289,7 +301,7 @@ struct batch_request {
 		struct rq_jobfile rq_jobfile;
 		char rq_rdytocommit[PBS_MAXSVRJOBID + 1];
 		char rq_commit[PBS_MAXSVRJOBID + 1];
-		struct rq_manage rq_delete;
+		struct rq_deljob rq_delete;
```
hirenvadalia, line 304: This is bit confusing as you are using same structure for both deljob and deljoblist. I would suggest to keep rq_manage as it for deljob and use new struct (rename it to rq_deljoblist) for deljoblist

src/include/batch_request.h
```diff
@@ -103,6 +103,18 @@ struct rq_manage {
 	pbs_list_head rq_attr; /* svrattrlist */
 };
 
+struct rq_deljob {
+	int rq_cmd;
```
hirenvadalia, line 107: Use rq_manage for rq_manage for deljob and remove this duplicate members from this struct, we need only rq_count and rq_jobids in this struct

src/include/libpbs.h
```diff
@@ -213,6 +222,8 @@ struct batch_reply
 		struct brp_select *brp_select; /* select replies */
 		pbs_list_head brp_status; /* status (svr) replies */
 		struct batch_status *brp_statc; /* status (cmd) replies) */
+		pbs_list_head brp_delstat; /* status (deleted jobs) replies) */
```
hirenvadalia, line 225: We don't need this and `brp_deletejobstat` struct, use `batch_deljob_status` on server side also while creating replies.

src/include/libpbs.h
```diff
@@ -332,7 +344,9 @@ int PBSD_jscript_direct(int, char *, int, char **);
 int PBSD_copyhookfile(int, char *, int, char **);
 int PBSD_delhookfile(int, char *, int, char **);
 int PBSD_mgr_put(int, int, int, int, char *, struct attropl *, char *, int, char **);
+int PBSD_delete_put(int, int, int, int, char **, struct attropl *, char *, int, char **);
```
hirenvadalia, line 347: I suggest to rename this to say its for deljoblist like PBSD_deljoblist_put

src/include/libpbs.h
```diff
@@ -332,7 +344,9 @@ int PBSD_jscript_direct(int, char *, int, char **);
...
 int PBSD_manager(int, int, int, int, char *, struct attropl *, char *);
+struct batch_deljob_status *PBSD_delete(int, int, int, int, char **, struct attropl *, char *);
```
hirenvadalia, line 349: same as above

src/lib/Libifl/pbsD_deljob2.c
```diff
@@ -0,0 +1,82 @@
...
+ * @param[in] extend - string to encode req
+ *
+ * @return	struct batch_status *
+ * @retval	0	success
+ * @retval	!0	error
+ *
+ */
+
+struct batch_deljob_status *
+__pbs_deljoblist(int c, char **jobid, char *extend)
```
hirenvadalia, line 69: rename 2nd arg to `jobids`

src/lib/Libifl/pbs_delstatfree.c
```diff
@@ -0,0 +1,73 @@
...
+ * @brief
+ *	-The function that deallocates a "batch_deljob_status" structure
+ *
+ * @param[in] bsp - pointer to batch request.
+ *
+ * @return	Void
+ *
+ */
+void
+__pbs_delstatfree(struct batch_deljob_status *bsp)
```
hirenvadalia, line 63: Do we really need separate file for this small func? Can't we put this func in int_rdrpy.c itself?

src/server/req_delete.c
```diff
@@ -337,6 +341,30 @@ decr_single_subjob_usage(job *parent)
 	parent->ji_qs.ji_svrflags |= JOB_SVFLG_ArrayJob; /* setting arrayjob flag back */
 }
 
+
+int update_deletejob_stat(char *jid, struct batch_request *preq, int errcode, pbs_list_head *pdelstathd)
```
hirenvadalia, line 345: You don't need last arg as you as access to preq in 2nd arg.

src/server/req_delete.c
```diff
@@ -362,230 +390,288 @@ req_deletejob(struct batch_request *preq)
...
-		/* just a regular job, pass it on down the line and be done
-		 * If the request is to purge the history of the job then set ji_deletehistory to 1
+	
+	preq->rq_ind.rq_delete.tot_jobs = count;
+	
+	for (j = 0; list[j] != NULL; j++) {
+		
+		snprintf(jid, sizeof(jid), "%s", list[j]);
+		
+		if (preq->rq_extend && strstr(preq->rq_extend, DELETEHISTORY))
```
hirenvadalia, line 415: move this (and below all where check extend and decided what is requested) code outside of for loop

src/server/req_delete.c
```diff
@@ -362,230 +390,288 @@ req_deletejob(struct batch_request *preq)
...
-	if (jt == IS_ARRAY_NO) {
-
-		/* just a regular job, pass it on down the line and be done
-		 * If the request is to purge the history of the job then set ji_deletehistory to 1
+	
+	preq->rq_ind.rq_delete.tot_jobs = count;
+	
+	for (j = 0; list[j] != NULL; j++) {
+		
+		snprintf(jid, sizeof(jid), "%s", list[j]);
```
hirenvadalia, line 413: why we need to create another copy of jobid? can't we use directly list[j]?

src/server/req_delete.c
```diff
@@ -362,230 +390,288 @@ req_deletejob(struct batch_request *preq)
...
-		if (pjob != NULL && pjob->ji_pmt_preq != NULL)
-			reply_preempt_jobs_request(err, PREEMPT_METHOD_DELETE, pjob);
-		return; /* note, req_reject already called */
-	}
-
-	if (delhist) {
-		rc = check_deletehistoryjob(preq);
-		if (rc == TRUE)
-			return;
+	char **list;
```
hirenvadalia, line 393: rename this to jobids

src/server/req_delete.c
```diff
@@ -362,230 +390,288 @@ req_deletejob(struct batch_request *preq)
...
+		count = 1;
 	}
-
-
-	if (jt == IS_ARRAY_NO) {
-
-		/* just a regular job, pass it on down the line and be done
-		 * If the request is to purge the history of the job then set ji_deletehistory to 1
+	
+	preq->rq_ind.rq_delete.tot_jobs = count;
```
hirenvadalia, line 409: why we need this variable, we have rq_count in rq_delete

src/server/req_delete.c
```diff
@@ -362,230 +390,288 @@ req_deletejob(struct batch_request *preq)
...
+					return;
+				} else
+					continue;
+			}
 
-		++preq->rq_refct;
+			subjobstate = get_subjob_state(parent, offset);
+			if (subjobstate == -1) {
+				update_deletejob_stat(jid, preq, PBSE_IVALREQ, &preply->brp_un.brp_delstat);
+				if (preq->rq_ind.rq_delete.tot_rpys == preq->rq_ind.rq_delete.tot_jobs) {
```
hirenvadalia, line 477: Looks like we are doing this if condition after every call to update_deletejob_stat(), can't move this in update_deletejob_stat() itself?

src/server/req_delete.c
```diff
@@ -362,230 +390,288 @@ req_deletejob(struct batch_request *preq)
...
-			if ((sjst == JOB_STATE_LTR_EXITING) && !forcedel)
-				continue;
-			if ((pjob = parent->ji_ajtrk->tkm_tbl[i].trk_psubjob)) {
+			if ((subjobstate == JOB_STATE_LTR_EXITING) && (forcedel == 0)) {
+				if (parent->ji_pmt_preq != NULL) {
+					pjob = find_job(jid);
+					reply_preempt_jobs_request(PBSE_BADSTATE, PREEMPT_METHOD_DELETE, pjob);
+				}
+				update_deletejob_stat(jid, preq, PBSE_BADSTATE, &preply->brp_un.brp_delstat);
+				if (preq->rq_ind.rq_delete.tot_rpys == preq->rq_ind.rq_delete.tot_jobs) {
```
hirenvadalia, line 490: Looks like this variables are used to decide when to reply, if so plz move this to batch_reply struct instead rq_delete (which is struct used for incoming request)

src/server/req_delete.c
```diff
@@ -362,230 +390,288 @@ req_deletejob(struct batch_request *preq)
...
+			 * the parent array.The function chk_array_doneness() will take care of eventually
+			 *  purging the history .
+			 */
+			if (delhist)
+				parent->ji_deletehistory = 1;
+			/* The Array Job itself ... */
+			/* for each subjob that is running, delete it via req_deletejob2 */
+
+			++preq->rq_refct;
+			preq->rq_ind.rq_delete.tot_arr_jobs++;
```
hirenvadalia, line 536: I don't see any usage of this variable other than this, can you please point me where its getting used?

src/server/req_jobobit.c
```diff
@@ -1010,8 +1009,7 @@ on_job_rerun(struct work_task *ptask)
 				/* need to have MOM delete her copy of the job */
 				preq = alloc_br(PBS_BATCH_DeleteJob);
 				if (preq) {
-					(void)strcpy(preq->rq_ind.rq_delete.rq_objname,
-						pjob->ji_qs.ji_jobid);
+					preq->rq_ind.rq_delete.rq_objname = strdup(pjob->ji_qs.ji_jobid);
```
hirenvadalia, line 1012: U need to check for strdup() (all changed strdup)

hirenvadalia (changes requested):

src/cmds/qdel.c
```diff
@@ -58,6 +58,75 @@
...
+	
+tryagain:
+	for (i = 0; i < no_of_pbsservers; i++) {
+		if (pbs_servers[i]->server_name == NULL) {
+			pbs_servers[i]->server_name = strdup(server_out);
+			if (pbs_servers[i]->server_name == NULL) {
+				fprintf(stderr, "qdel: unable to allocate memory.\n");
+				exit(2);
+			}
+			append_jobid(pbs_servers[i], job_id);
```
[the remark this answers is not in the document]
bremanandjk: We cant use this function since these functions would allocate new heap memory and return the passed siring. Here the objective is we need to reuse the reallocated memory for *arv[].

src/cmds/qdel.c
```diff
@@ -58,6 +58,75 @@
...
+		} else if (strcmp(server_out, pbs_servers[i]->server_name) == 0) {
+			append_jobid(pbs_servers[i], job_id);
+			found = 1;
+			break;
+		}
+	}
+	
+	if (!found) {
+		no_of_pbsservers *= 2;
+		pbs_servers = realloc(&pbs_servers, no_of_pbsservers);
```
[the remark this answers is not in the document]
bremanandjk: Here, we are allocating pbs_servers struct, it's not possible to use string addition functions.

src/cmds/qdel.c
bremanandjk: It's not a string addition operation. we can't use that function.

src/cmds/qdel.c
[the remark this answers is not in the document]
hirenvadalia: add_ptr_to_array does exactly that. it doesn't do strdup() like add_str_to_array

src/cmds/qdel.c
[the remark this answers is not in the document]
hirenvadalia: add_ptr_to_array is not str operation, its pointer operation.

src/server/req_delete.c
bremanandjk: Please refer to the code following this condition, in case of success, we need to reply to continue to the next jobid or we need to return the call for req_delete, still, we need to keep {if, else} block, that we can't avoid.

src/cmds/qdel.c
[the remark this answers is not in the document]
bremanandjk: please have a look into "add_ptr_to_array", it is counting the number of elements in each call and reallocating the size of the string array and string the new string. If we want to use this we need to refactor a lot of code in the scheduler wherever it is getting used. That refactor should be taken separately. I don't want to mix fixing that functions and extend this qdel refactoring work. 

If the function is a good performant to use, it makes sense to use it directly.

src/cmds/qdel.c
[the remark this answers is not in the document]
bremanandjk: explained above.

src/cmds/qdel.c
bremanandjk: okay, I agree, I will change it.

src/server/req_jobobit.c
bremanandjk: okay, I agree

src/include/libpbs.h
bremanandjk: Please explain, how can we use the same struct. It's not feasible in req_delete() function. Its possible in req_stat() & req_select()

There is one difference between req_stat and req_delete. In req_stat(), all the select records are added within the same function, so they are maintaining the next points in pselx and passing the same to each call. 
 
But in case of req_delete, the addition of delete status records needs to be added from 3 different functions, like req_delete(), req_delete2() and post_delete_mom1().

src/include/libpbs.h
bremanandjk: okay, I agree

src/cmds/qdel.c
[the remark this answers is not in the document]
hirenvadalia: We have using it in scheduler for long time @bremanandjk so I don't think it will cause any slowness. Also this code is client side so it won't matter much.

src/server/req_delete.c
bremanandjk: You were suggesting in an earlier comment to remove the pbs list head from batch_reply, And here you are suggesting to remove this variable also. 

then how shall we track the next pointer to add the new delete job record. ? plz explain.

src/cmds/qdel.c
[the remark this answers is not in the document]
bremanandjk: I think we can't say that since we are using in the scheduler, we can use it here also. Even though it's a client program, if we have a huge number say(1 million), jobs we are going to iterate over all jobs in each job addition. like 1million factorial. Do you still think, it's not a performance issue.

src/cmds/qdel.c
bremanandjk: I will change this.

src/include/libpbs.h
hirenvadalia: I doesn't matter how many func its adding record as there is only one func `update_deletejob_stat()` - who is adding record in this list. Anyways its fairly simple to use batch_deljob_status and blp_delstatc instead this, u can do like this:
preply = &preq->rq_reply; // preq is passed as arugment
newstat = malloc(batch_deljob_status)
fill newstat with its name and code
newstat->next = preply->brp_delstatc;
preply->brp_delstatc = newstat;

bremanandjk: You should also think about encoding and free this list part. If the "preply->brp_delstatc" also holds the lastly added struct, how do you traverse to firstly added struct(remember it is a single linked list)? You are having only the pointer to the last struct. We need head pointer at two places, One at encoding operating where we are actually sending the records and another place at freeing the records after seeing(or at failure).

src/cmds/qdel.c
bremanandjk: Good thought, what you suggest for that corner case. ?

src/lib/Libifl/pbs_delstatfree.c
bremanandjk: That's what the convention we are following in PBS, refer __pbs_statfree(), I thought we should follow the same.

agrawalravi90 (changes requested):
Please post performance numbers for this change

src/include/libpbs.h
hirenvadalia: @bremanandjk You already have func called `pbs_delstatfree()` to free batch_deljob_status structure, and traversing will be same as how you are traversing in `pbs_delstatfree()`

> What is the benefit we are achieving by we get by doing all these code changes, do you think, this much refactoring is worth to get the final outcome. I think this will simply delay this qdel optimization work without much benefit out of it.

Yes, I strongly believe that doing this change will be beneficial in future and now also (as you will have less new structures and pointers in code to maintain + code will be much more easily understandable)

src/cmds/qdel.c
hirenvadalia: I would look into moving locate_job() in req_deletejob on server side, as ultimately locate_job does pbs_connect() and then pbs_locjob() which goes to same server where we send deljob request, so same server can reply PBSE_UNKJOBID with located location as extra thing while replying to deljoblist request - this will save pbs_connect() (which is 1 req, 1 reply + all auth related part) and another pair of req/reply for pbs_locjob() - aka total 2 req/reply pair + auth stuff

src/lib/Libifl/pbs_delstatfree.c
hirenvadalia: I understand that in past we have added in separate files which I don't like at all for small funcs, but for this I think we should put this small func in int_rdrpy.c and in future if needed then we put all pbs_*free in one file rather than in individual file

src/server/req_delete.c
hirenvadalia: explained in https://github.com/openpbs/openpbs/pull/2063#discussion_r514255359

src/cmds/qdel.c
bremanandjk: How this optimization would solve the question you have raised earlier? However, we find the moved destination server name, and that server was being processed already, how to send this deljoblist request to it. 

Whatever you are suggesting is applicable to many commands like, qsig, qstat, qhold, qmove, qmsg. Everywhere we are sending the request to the server to get the moved server name only if the return code is PBSE_UNKJOBID. Why now we need all the new optimization in this qdel PR. Do you feel is it worth to spend extra time to optimize this PR. It will simply cause us to delay the merge, I think the first priority should be to do the changes required to get the fastest qdel without breaking main functionalities, then later we can do all other innovative optimizations in pbs_server.

hirenvadalia: @bremanandjk I think you misunderstand my point - Optimization is 2nd part, first I gave you suggestion for fixing issue which will come with this PR changes if we don't do anything.

> I would look into moving locate_job() in req_deletejob on server side, as ultimately locate_job does pbs_connect() and then pbs_locjob() which goes to same server where we send deljob request, so same server can reply PBSE_UNKJOBID with located location as extra thing while replying to deljoblist request

This is suggestion to fix the issue

>  this will save pbs_connect() (which is 1 req, 1 reply + all auth related part) and another pair of req/reply for pbs_locjob() - aka total 2 req/reply pair + auth stuff

This is optimization which will get free of cost if we do above suggestion.

hirenvadalia: Another solution you can do is create another list of servers with list of located jobids (like you have pbs_servers now) then once all given jobids are processed you can process new servers list with located jobids. That way you won't have to move locate_job() code in req_deletejob().

bremanandjk: Right, this would be an apt solution. I am also implementing the same, Thanks

src/cmds/qdel.c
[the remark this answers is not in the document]
bremanandjk: In our internal discussion, @nithinj we, decided to proceed with append_jobid().

src/include/libpbs.h
bremanandjk: I have pushed the changes as per the suggestion.

bremanandjk:
> Please post performance numbers for this change

I have added the results in the description.

src/cmds/qdel.c
```diff
@@ -241,25 +328,23 @@ char **envp;
...
-			if (locate_job(job_id_out, server_out, rmt_server)) {
-				pbs_disconnect(connect);
-				pbs_strncpy(server_out, rmt_server, sizeof(server_out));
-				goto cnt;
+			if (p_delstatus->code != PBSE_HISTJOBDELETED)
+				num_deleted++;
+			
+			if (p_delstatus->code == PBSE_UNKJOBID) {
+				if (locate_job(p_delstatus->name, server_out, rmt_server))
+					add_jobid_to_pbs_servers(rmt_server, rmt_server, &pbs_servers);
```
nithinj, line 344: Since you are adding to the bucket again on failure, you might have to loop through the list again to delete located jobs. Also take care of the already existing jobs on that bucket

src/lib/Libifl/pbsD_deljob2.c
```diff
@@ -0,0 +1,82 @@
...
+{
+	struct attropl *aoplp = NULL;
+
+	if ((jobid == NULL) || (**jobid == '\0'))
+		return NULL;
+
+	return PBSD_delete(c, PBS_BATCH_DeleteJobList,
```
nithinj, line 76: Do you need separate files for int_delete2.c and int_deleteput.c? Can we keep those functions within pbsD_deljob2?

nithinj (changes requested):

nithinj (changes requested):

src/cmds/qdel.c
```diff
@@ -241,25 +328,23 @@ char **envp;
...
-			prt_job_err("qdel", connect, job_id_out);
-			any_failed = pbs_errno;
-		} else if (stat && (pbs_errno == PBSE_UNKJOBID) && !located) {
-			located = TRUE;
-			if (locate_job(job_id_out, server_out, rmt_server)) {
-				pbs_disconnect(connect);
-				pbs_strncpy(server_out, rmt_server, sizeof(server_out));
-				goto cnt;
+			if (p_delstatus->code != PBSE_HISTJOBDELETED)
+				num_deleted++;
```
agrawalravi90, line 340: So, if the code is 'PBSE_UNKJOBID', you will still increment num_deleted here, which seems incorrect. Maybe check for code to be PBSE_NONE to increment num_deleted?

src/include/batch_request.h
agrawalravi90: @bremanandjk  did you miss this comment?

src/include/libpbs.h
agrawalravi90: @bremanandjk  this hasn't been addressed yet

src/lib/Libifl/dec_DelJobList.c
```diff
@@ -0,0 +1,133 @@
...
+ * "OpenPBS®", "PBS Professional®", and "PBS Pro™" and Altair's logos is
+ * subject to Altair's trademark licensing policies.
+ */
+
+
+
+/**
+ * @file	dec_Manage.c
+ * @brief
+ * decode_DIS_Manage() - decode a Manager Batch Request
```
agrawalravi90, line 45: this comment is related to Manage and Manager, i think you should remove it completely

src/lib/Libifl/dec_DelJobList.c
```diff
@@ -0,0 +1,133 @@
...
+#include "attribute.h"
+#include "credential.h"
+#include "batch_request.h"
+#include "dis.h"
+/**
+ * @brief
+ *	-decode a Delete Job Batch Request
+ *
+ * @par	Functionality:
+ *	This request is used for delete operations  deleted.
```
agrawalravi90, line 76: delete operations deleted?

src/lib/Libifl/dec_DelJobList.c
```diff
@@ -0,0 +1,133 @@
...
+	preq->rq_ind.rq_delete.rq_cmd = disrui(sock, &rc);
+	if (rc) return rc;
+	preq->rq_ind.rq_delete.rq_objtype = disrui(sock, &rc);
+	if (rc) return rc;
+	preq->rq_ind.rq_delete.rq_count = disrui(sock, &rc);
+	if (rc) return rc;
+	
+	count = preq->rq_ind.rq_delete.rq_count;
+	
+	tmp_jobslist = (char **)calloc(count + 1, sizeof(char *));
```
agrawalravi90, line 116: why calloc?

src/lib/Libifl/dec_DelJobList.c
```diff
@@ -0,0 +1,133 @@
...
+	if (tmp_jobslist == NULL) return DIS_NOMALLOC;
+	
+	for (i = 0; i < count; i++) {
+		tmp_jobslist[i] = disrst(sock, &rc);
+		if (rc) {
+			free(tmp_jobslist);
+			return rc;
+		}
+	}
+
```
agrawalravi90, line 126: tmp_joblist[i] = NULL ?

src/lib/Libifl/enc_Delete.c
```diff
@@ -0,0 +1,88 @@
...
+ * @param[in] aoplp - pointer to attropl structure(list)
+ *
+ * @return      int
+ * @retval      DIS_SUCCESS(0)  success
+ * @retval      error code      error
+ *
+ */
+
+int
+encode_DIS_Delete(int sock, int command, int objtype, char **objname, struct attropl *aoplp)
```
agrawalravi90, line 78: how about 'jobids' instead of 'objname' ?

src/lib/Libifl/ifl_impl.c
```diff
@@ -215,6 +215,26 @@ pbs_deljob(int c, char *jobid, char *extend) {
...
+ * @param[in] jobid - job identifier
+ * @param[in] extend - string to encode req
+ *
+ * @return	struct batch_status *
+ * @retval	0	success
+ * @retval	!0	error
+ *
+ */
+struct batch_deljob_status *
+pbs_deljoblist(int c, char **jobid, char *extend) {
```
agrawalravi90, line 233: how about 'jobids' instead of 'jobid' ?

src/lib/Libifl/int_delete2.c
```diff
@@ -0,0 +1,124 @@
...
+{
+	int i;
+	struct batch_reply *reply;
+	struct batch_deljob_status *rbsp = NULL;
+
+	/* initialize the thread context data, if not initialized */
+	if (pbs_client_thread_init_thread_context() != 0)
+		return NULL;
+
+
```
agrawalravi90, line 83: 2 blank lines

src/lib/Libifl/int_delete2.c
```diff
@@ -0,0 +1,124 @@
...
+	if ((pbs_verify_attributes(c, function, objtype,
+		command, aoplp)) != 0)
+		return NULL;
+
+	/* lock pthread mutex here for this connection */
+	/* blocking call, waits for mutex release */
+	if (pbs_client_thread_lock_connection(c) != 0)
+		return NULL;
+
+	/* send the manage request */
```
agrawalravi90, line 94: 'delete' request, not manage

src/lib/Libifl/int_delete2.c
```diff
@@ -0,0 +1,124 @@
...
+	i = PBSD_delete_put(c, function, command, objtype, objnames, aoplp, extend, PROT_TCP, NULL);
+	if (i) {
+		(void)pbs_client_thread_unlock_connection(c);
```
agrawalravi90, line 97: remove  (void)

src/lib/Libifl/int_delete2.c
```diff
@@ -0,0 +1,124 @@
...
+		return NULL;
+	}
+
+	/* read reply from stream into presentation element */
+	reply = PBSD_rdrpy(c);
+
+	if (reply == NULL) {
+		pbs_errno = PBSE_PROTOCOL;
```
agrawalravi90, line 105: PBSD_rdrpy already sets pbs_errno, so I think you should set it here only if reply == NULL && pbs_errno == PBSE_NONE

src/lib/Libifl/pbsD_deljob2.c
agrawalravi90: @bremanandjk  you missed this comment

src/lib/Libifl/pbsD_deljob2.c
```diff
@@ -0,0 +1,82 @@
...
+		MGR_CMD_DELETE,
+		MGR_OBJ_JOB,
```
agrawalravi90, line 78: why does PBS_BATCH_DeleteJobList need MGR_CMD_DELETE and MGE_OBJ_JOB ? it'll be handled by req_deletejob() at the server, not req_manager(), right?

agrawalravi90 (changes requested):
Haven't reviewed the server side changes yet

bremanandjk:
@agrawalravi90 @nithinj 
I have addressed all functionality related review comments, let me know those changes are fine.

src/lib/Libifl/dec_DelJobList.c
agrawalravi90: please address this

src/lib/Libifl/pbsD_deljob2.c
agrawalravi90: please address this

src/lib/Libifl/pbsD_deljob2.c
```diff
@@ -73,10 +77,138 @@ __pbs_deljoblist(int c, char **jobid, char *extend)
...
+{
+	int i;
+	struct batch_reply *reply;
+	struct batch_deljob_status *rbsp = NULL;
+
+	/* initialize the thread context data, if not initialized */
+	if (pbs_client_thread_init_thread_context() != 0)
+		return NULL;
+
+
```
agrawalravi90, line 116: 2 blank lines

src/lib/Libifl/pbsD_deljob2.c
```diff
@@ -73,10 +77,138 @@ __pbs_deljoblist(int c, char **jobid, char *extend)
...
+	if ((pbs_verify_attributes(c, function, objtype,
+		command, aoplp)) != 0)
+		return NULL;
+
+	/* lock pthread mutex here for this connection */
+	/* blocking call, waits for mutex release */
+	if (pbs_client_thread_lock_connection(c) != 0)
+		return NULL;
+
+	/* send the manage request */
```
agrawalravi90, line 127: manage request?

src/server/process_request.c
```diff
@@ -843,6 +843,11 @@ dispatch_request(int sfds, struct batch_request *request)
 			break;
 
 		case PBS_BATCH_DeleteJobList:
+			log_event(PBSEVENT_DEBUG, PBS_EVENTCLASS_JOB, LOG_INFO,
+				request->rq_ind.rq_deletejoblist.rq_jobslist[0],
+				"delete job request received");
```
agrawalravi90, line 848: add the word 'list' here to differentiate it from a regular deljob

src/server/reply_send.c
```diff
@@ -447,6 +453,15 @@ reply_free(struct batch_reply *prep)
 			(void)free(pstat);
 			pstat = pstatx;
 		}
+		
+	} else if (prep->brp_choice == BATCH_REPLY_CHOICE_Delete) {
+		pdelstat = prep->brp_un.brp_delstatc;
+		while (pdelstat) {
+			pdelstatx = pdelstat->next;
+			(void)free(pdelstat);
```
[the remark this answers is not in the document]
agrawalravi90: @bremanandjk  you resolved this without removing the cast, please remove it

src/server/req_delete.c
```diff
@@ -314,7 +312,8 @@ issue_delete(job *pjob)
 	if (preq == NULL)
 		return;
 
-	preq->rq_ind.rq_delete.rq_objname = strdup(pjob->ji_qs.ji_jobid);
+	strncpy(preq->rq_ind.rq_delete.rq_objname, pjob->ji_qs.ji_jobid, sizeof(preq->rq_ind.rq_delete.rq_objname) - 1);
```
agrawalravi90: use pbs_strncpy

src/lib/Libifl/enc_Delete.c
agrawalravi90: please address this

src/lib/Libifl/ifl_impl.c
agrawalravi90: please address this

src/lib/Libifl/pbsD_deljob2.c
agrawalravi90: please address this

src/server/svr_chk_owner.c
```diff
@@ -384,20 +384,24 @@ chk_job_request(char *jobid, struct batch_request *preq, int *rc, int *err)
 			jobid, msg_unkjobid);
 		if (err != NULL)
 			*err = PBSE_UNKJOBID;
-		req_reject(PBSE_UNKJOBID, 0, preq);
+
+		if (preq->rq_type != PBS_BATCH_DeleteJobList)
```
agrawalravi90, line 388: you are doing this 5 times in this function, how about adding a goto at the bottom of the function which will do this check, reject if not PBS_BATCH_DeleteJobList and return NULL?

src/server/svr_jobfunc.c
```diff
@@ -2719,7 +2719,7 @@ Time4resvFinish(struct work_task *ptask)
 				newreq->rq_perm |= ATR_DFLAG_MGWR;
 				strcpy(newreq->rq_user, pbs_current_user);
 				strcpy(newreq->rq_host, server_host);
-				strcpy(newreq->rq_ind.rq_delete.rq_objname, presv->ri_qs.ri_resvID);
+				strcpy(newreq->rq_ind.rq_manager.rq_objname, presv->ri_qs.ri_resvID);
```
agrawalravi90, line 2722: why did you change rq_delete to rq_manager?

agrawalravi90 (changes requested):
There are many review comments from previous rounds which have not been addressed yet, please address them and don't resolve conversations, it's hard to know what you addressed and what not.

Also, you need to do a qdel test with multiple clusters to make sure that deleting jobs on remote clusters still works after your changes

src/server/svr_chk_owner.c
bremanandjk: The error code for each req_reject() is different, hence we can't have a common code in the goto function.

src/server/svr_jobfunc.c
bremanandjk: The data structure used to store the information for ResvOccurEnd request is rq_manager, it would be ideal to use rq_manager for better code readability.

bremanandjk:
@agrawalravi90 I have pushed a commit that addressed all the remaining and new comments. In an earlier round of reviews, some comments are invalid due to changes in logic/approach and some comments are hidden conversation. Now, I have gone through each comment and addressed it. Please let me know your further comments/sign-off. Thanks

src/server/process_request.c
bremanandjk: Some test cases depend on this log msg. I think, for both the delete and deletejoblist request this log msg would be appropriate. If you feel otherwise, Let me know, I will change this log msg and its corresponding test cases.

src/cmds/qdel.c
```diff
@@ -241,25 +341,23 @@ char **envp;
...
-				pbs_disconnect(connect);
-				pbs_strncpy(server_out, rmt_server, sizeof(server_out));
-				goto cnt;
+		p_delstatus = pbs_deljoblist(connect, svr_itr->jobids, warg);
+		
+		while (p_delstatus != NULL) {
+			if (p_delstatus->code == PBSE_UNKJOBID) {
+				if (locate_job(p_delstatus->name, server_out, rmt_server))
+					add_jid_to_list(rmt_server, rmt_server, &svr_itr);
+				continue;
```
nithinj, line 350: calling continue without updating the variable will trigger an infinite loop.

src/cmds/qdel.c
```diff
@@ -58,6 +58,99 @@
...
+
+	for (iter_list = *svr_jobid_list_hd; iter_list != NULL; prev = iter_list, iter_list = iter_list->next) {
+		if (strcmp(server, iter_list->server_name) == 0) {
+			append_jobid(iter_list, job_id);
+			break;
+		}
+	}
+
+	if (iter_list == NULL) {
+		svr_jobid_list_t *new_node = NULL;
```
nithinj, line 141: declare the variables at the beginning of the function or leave a line after.

src/cmds/qdel.c
```diff
@@ -58,6 +58,99 @@
 #include <pbs_config.h>   /* the master config generated by configure */
 #include <pbs_version.h>
 
+#define NO_OF_JOBIDS 1000 /* Using this macro to allocate array of pointers */
```
nithinj, line 61: In your earlier approach, you were sending jobs by batches to the server if it is a huge array.  That will give room for server to do other useful work. Are we achieving the same now,  if not why?

nithinj (changes requested):

agrawalravi90:
@bremanandjk In my previous round of comments, i had also suggested that you do a remote server test: set up 2  clusters, submit jobs to both, then delete jobs from both and make sure it works correctly. Did you get a chance to do it? Please post logs here, thanks.

agrawalravi90 (changes requested):

bremanandjk:
@agrawalravi90 
I forgot to share the test results, sorry about it, here, it is
[Remote_host_qdel_test_logs.txt](https://github.com/openpbs/openpbs/files/5533921/Remote_host_qdel_test_logs.txt)

agrawalravi90:
Thanks for sharing!

src/cmds/qdel.c
bremanandjk: I have brought back the initial approach I followed that is, sending jobs in batches. Now, in each request, the initial number of jobs based on the email suppression number and the following iteration will be sending 10k jobs.

src/cmds/qdel.c
bremanandjk: Changed to for loop, it's okay to call continue now.

src/server/req_delete.c
```diff
@@ -1424,7 +1429,7 @@ post_delete_mom1(struct work_task *pwt)
 		if (rc == PBSE_UNKSIG) {
 			if (tries++) {
 				update_deletejob_stat(pjob->ji_qs.ji_jobid, preq_clt, rc, &preply->brp_un.brp_delstat);
-				if (preq_clt->rq_ind.rq_delete.tot_rpys == preq_clt->rq_ind.rq_delete.tot_jobs)
+				if (preply->brp_un.brp_deletejoblist.tot_rpys == preply->brp_un.brp_deletejoblist.tot_jobs)
```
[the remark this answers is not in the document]
bremanandjk: I prefer brp_deletejoblist since it matches the request name. More readability.

bremanandjk:
@agrawalravi90 @nithinj 
I have pushed a commit with further review comments, plz have a look for further reviews/sign-off. 

With these changes, the test results are 

[root@testmachine pbspro]# time qselect | xargs qdel -Wforce 
real	**0m25.411s**
user	0m0.350s
sys	0m0.182s

src/cmds/qdel.c
```diff
@@ -334,30 +346,38 @@ char **envp;
...
+		numofjobs = (mails <= svr_itr->job_ct) ? mails : svr_itr->job_ct;
 		
-		while (p_delstatus != NULL) {
-			if (p_delstatus->code == PBSE_UNKJOBID) {
-				if (locate_job(p_delstatus->name, server_out, rmt_server))
-					add_jid_to_list(rmt_server, rmt_server, &svr_itr);
-				continue;
+		for (i = 0; i < svr_itr->job_ct; i = i + numofjobs) {
+			if (mails_suppressed)
+				numofjobs = ((i + NUM_JOBS_IN_A_BATCH) <= svr_itr->job_ct) ? NUM_JOBS_IN_A_BATCH : (svr_itr->job_ct - i);
```
agrawalravi90, line 353: why i + NUM_JOBS_IN_A_BATCH ? this will mean that the first pbs_deljoblist will delete 10k, the next one 20k, and so on, right?

agrawalravi90 (changes requested):
just one comment

src/cmds/qdel.c
bremanandjk: This condition "(i + NUM_JOBS_IN_A_BATCH) <= svr_itr->job_ct)" equals to "current job index + 10k" is less then total count. If this condition succeeds means, qdel has enough jobs further to delete, then assigning 10k to "numofjobs". So in each iteration, it will be 10k always.

agrawalravi90: ok yes, true .. but why are you checking i + NUM_JOBS_IN_A_BATCH ? after the first iteration, i will be 10k, so you'll check whether 20k is less than svr_itr->job_ct ? why not just check if NUM_JOBS_IN_A_BATCH <= svr_itr->job_ct ?

agrawalravi90: thanks for explaining offline that 'job_ct' is the non-changing total number of jobs. Then this is fine, but please rename job_ct to something like 'total_jobs' to avoid confusion.

nithinj: Both the conditional operators does the same thing. You only have to decide the batch size in the if condition.
```
			if (!mails_suppressed)
				batch_size = mail_count;
			numofjobs = ((i + batch_size) <= svr_itr->job_ct) ? batch_size : (svr_itr->job_ct - I);
```

nithinj (approved):
Rest looks good to me!

bremanandjk:
@nithinj Thanks for your review, I have removed the debug line.

agrawalravi90 (approved):
LGTM

PR #2064, opened 2020-10-08
IFL changes for multi-server

nithinj:
For multi-server, PBS complex will be consisting of multiple server instances. Clients will connect to all server instances and aggregate the values and make it look like coming from a single cluster.

* stat_job(), stat_node()
IFL will simply append values coming from several instances as jobs and nodes are sharded.
* stat_queue(), stat_server()
IFL will aggregate the resources assigned and jobs counts per state before delivering the output.
* Manager() API
This is manly used by qmgr. These requests will be delivered to each instances. If it cannot reach one of the server, the whole operation is aborted.
It is important that all common objects should be in sync across the instances. Right now admin has to make sure that the objects are in sync if qmgr operation fails after it is rejected by one of the server instances. Further enhancements will be made in the future. 
* Operations on a specific job ID. qstat, qalter, qdel:
IFL will parse the server name and reach out to that server for the operation. Upon failure, it will contact other servers in the complex.

Detailed logs are attached:
[logs.txt](https://github.com/openpbs/openpbs/files/5350873/logs.txt)


src/cmds/qmgr.c
```diff
@@ -1204,6 +1204,17 @@ make_connection(char *name)
 	else
 		PSTDERR1("qmgr: cannot connect to server %s\n", name)
 
+	if (msvr_mode()) {
+		int i;
+		svr_conn_t *svr_connections = get_conn_servers(connection);
```
subhasisb, line 1209: feels like a null terminated array is better than having to asking for the number of connections.

src/cmds/qstat.c
```diff
@@ -804,7 +806,31 @@ altdsp_statjob(struct batch_status *pstat, struct batch_status *prtheader, int a
 		pstat = bs_isort(pstat, cmp_est_time);
 
 	if (prtheader) {
-		printf("\n%s: ", prtheader->name);
+		svr_conn_t *svr_connections = get_conn_servers(conn);
+		int num_cfg_svrs = get_num_servers();
```
subhasisb, line 810: same comment as above. Could we do away with get_num_servers() here and just return a null terminated array? Of course, this is no problem per se either

src/cmds/qstat.c
```diff
@@ -2889,26 +2915,28 @@ qstat -B [-f] [-F format] [-D delim] [ server_name... ]\n";
...
-					connect = cnt2server(prev_server);
+					conn = cnt2server(prev_server);
 				else
-					connect = cnt2server(server_out);
+					conn = cnt2server(server_out);
 
-				if (connect <= 0) {
-					fprintf(stderr, "qstat: cannot connect to server %s (errno=%d)\n",
+				if (conn <= 0) {
+					fprintf(stderr, "qstat: cannot conn to server %s (errno=%d)\n",
```
subhasisb, line 2923: looks like the word connect got inadvertently changed to conn due to search and replace

src/cmds/qstat.c
```diff
@@ -3076,21 +3104,21 @@ qstat -B [-f] [-F format] [-D delim] [ server_name... ]\n";
 						server_out[0] = '\0';
 				}
 que_no_args:
-				connect = cnt2server(server_out);
-				if (connect <= 0) {
-					fprintf(stderr, "qstat: cannot connect to server %s (errno=%d)\n", pbs_server, pbs_errno);
+				conn = cnt2server(server_out);
+				if (conn <= 0) {
+					fprintf(stderr, "qstat: cannot conn to server %s (errno=%d)\n", pbs_server, pbs_errno);
```
subhasisb, line 3109: same comment here -word in error message got changed inadvertently

src/cmds/qstat.c
```diff
@@ -3117,28 +3145,28 @@ qstat -B [-f] [-F format] [-D delim] [ server_name... ]\n";
...
 
 			case SERVERS:           /* get status of batch servers */
 				pbs_strncpy(server_out, operand, sizeof(server_out));
 svr_no_args:
-				connect = cnt2server(server_out);
-				if (connect <= 0) {
-					fprintf(stderr, "qstat: cannot connect to server %s (errno=%d)\n",
+				conn = cnt2server(server_out);
+				if (conn <= 0) {
+					fprintf(stderr, "qstat: cannot conn to server %s (errno=%d)\n",
```
subhasisb, line 3156: word got changed

src/cmds/qstat.c
```diff
@@ -3407,3 +3435,21 @@ cnvt_est_start_time(char *est_time, int wide)
...
 	return timebuf;
 }
+
+/**
+ * @brief used to display server instance failures in case of  MULTI_SERVER
+ * 
+ * @return void
+ */
+static void
+show_svr_inst_fail(int fd)
```
subhasisb, line 3445: Don't we need this function in several commands? If so, we could move this to libcmds and reused from all commands

src/cmds/qsub.c
```diff
@@ -4390,8 +4390,18 @@ do_connect(char *server_out, char *retmsg)
...
 	refresh_dfltqsubargs();
-	if (pbs_errno != PBSE_NONE)
-		return (pbs_errno);
+
+	if (pbs_errno != PBSE_NONE) {
+		if (msvr_mode() && (pbs_errno == PBSE_NOSERVER)) {
+			int i;
+			svr_conn_t *svr_connections = get_conn_servers(sd_svr);
+			for (i = 0; i < get_num_servers(); i++) {
+				if (svr_connections[i].state != SVR_CONN_STATE_UP)
```
subhasisb, line 4399: could we not simply call function show_svr_inst_fail()?

src/include/pbs_ifl.h
```diff
@@ -526,6 +526,7 @@ struct batch_status {
 	char		    *name;
 	struct attrl	    *attribs;
 	char		    *text;
+	struct batch_status *last;
```
subhasisb, line 529: since the forward point is called next, do you want to call the backward pointer as prev? or did u really mean last?

src/lib/Libifl/pbsD_submit.c
```diff
@@ -164,6 +166,8 @@ __pbs_submit(int c, struct attropl  *attrib, char *script, char *destination, ch
 	struct cred_info *cred_info = NULL;
 	int commit_done = 0;
 	char *lextend = NULL;
+	svr_conn_t *svr_connections = get_conn_servers(c);
```
subhasisb, line 169: would it be lesser confusion if we rename the function get_conn_servers() to get_conn_svr_instances()? That might re-enforce the fact that these are merely server instances?

src/lib/Libifl/pbsD_selectj.c
```diff
@@ -80,29 +178,71 @@ char **
...
+	if (pbs_client_thread_unlock_connection(c) != 0)
+		goto done;
+
+	if (rc)
+		pbs_errno = rc;
+
+	free(failed_conn);
+
+done:
+	ret = reply_to_jobarray(rlist);
```
subhasisb, line 240: can we rename this function from reply_to_jobarray() to reply_to_joblist() or something similar, so that it is not confued with array jobs?

src/lib/Libifl/pbsD_selectj.c
```diff
@@ -58,9 +58,107 @@
...
+		if (*rlist) {
+			(*rlist)->last->next = new;
+			(*rlist)->last = new;
+		} else
+			*rlist = new;
+	}
+}
+
+char **
+reply_to_jobarray(struct reply_list *rlist) {
```
subhasisb, line 112: please add a proper function header. Also, if this function seems more like an utility function move this to libcmds or libutil

subhasisb (changes requested):

src/cmds/qmgr.c
nithinj: Yes, we can do that as well. There are places where we need to make use of num_of_servers even if we have a null-terminated array. We make use of num_of_servers in case of finding a random server index. Also when finding the server to connect to based on the object id (job.servername). Using no_of_servers does cater all our use cases.
Also no of servers is a global var accessed only using a getter. Let me know if you still thinks it needs to be changed.

src/include/pbs_ifl.h
nithinj: I meant last. This points to the last object in the current server. So that we do not have to iterate in IFL layer while aggregating. 
In fact, this is quite a waste of memory as only the head needs to have the last pointer. In our case, each batch status structure contains the last pointer. I've moved it into batch_reply structure to avoid this.

src/lib/Libifl/pbsD_selectj.c
nithinj: It reads brp_select and decodes them into array of job ids. This is very specific to pbsD_select.
I've added the header.

nithinj:
@subhasisb ,

src/lib/Libifl/int_status.c
```diff
@@ -81,7 +221,472 @@ PBSD_status(int c, int function, char *objid, struct attrl *attrib, char *extend
...
+			break;
+	}
+
+	if (orig_st_ct)
+		encode_states(orig_st_ct, cur_st_ct, nxt_st_ct);
+	if (tot_jobs_attr)
+		sprintf(tot_jobs_attr, "%ld", tot_jobs);
+}
+
+#define DOUBLE 0
```
arungrover, line 284: Use ATR_TYPE_ defines

src/lib/Libifl/pbsD_selectj.c
```diff
@@ -58,9 +58,118 @@
...
+
+	/* read reply from stream */
+
+	reply = PBSD_rdrpy(c);
+	if (reply == NULL) {
+		pbs_errno = PBSE_PROTOCOL;
+	} else if (reply->brp_choice != BATCH_REPLY_CHOICE_NULL &&
+		   reply->brp_choice != BATCH_REPLY_CHOICE_Text &&
+		   reply->brp_choice != BATCH_REPLY_CHOICE_Select) {
+		pbs_errno = PBSE_PROTOCOL;
```
arungrover, line 92: don't we need to free reply here?

src/lib/Libifl/pbsD_selectj.c
```diff
@@ -58,9 +58,118 @@
...
+	} else if (get_conn_errno(c) == 0) {
+		if (reply->brp_un.brp_select == NULL) {
+			PBSD_FreeReply(reply);
+			return;
+		}
+
+		struct reply_list *new = malloc(sizeof(struct reply_list));
```
arungrover, line 99: check for null

src/lib/Libifl/pbsD_selectj.c
```diff
@@ -58,9 +58,118 @@
...
+	}
+}
+
+/**
+ * @brief
+ *	-convert the reply list to array of jobids
+ *
+ * @param[in] rlist - reply list
+ *
+ * @return	string list
```
arungrover, line 117: This list you are returning has data that points to location within the array. Please write a comment about format of this list.

arungrover (changes requested):

src/cmds/qmgr.c
subhasisb: its fine for now

src/include/pbs_ifl.h
subhasisb: Understood and this is fine. The list_link structure is actually very nice in this regard. You can do a GET_PRIOR(head) to get the last element, and use the same macro to go backwards if you wanted to as well. But it does not have to be done now. I will take it up later on when merging the diffstat stuff.

src/cmds/qmgr.c
```diff
@@ -1238,9 +1241,9 @@ connect_servers(struct objname *server_names, int numservers)
 		cur_obj = server_names;
 
 		/* if numservers == -1 (all servers) the var i will never equal zero */
-		for (i = numservers; i != 0 && cur_obj != NULL; i--, cur_obj=cur_obj->next) {
+		for (i = numservers; i && cur_obj; i--, cur_obj = cur_obj->next) {
 			nservers++;
-			if ((cur_svr = make_connection(cur_obj->svr_name)) ==NULL) {
+			if ((cur_svr = make_connection(cur_obj->svr_name)) == NULL || pbs_errno) {
 				nservers--;
 				error = TRUE;
```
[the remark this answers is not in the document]
nithinj: qmgr can be used to set server attributes of a list of servers. qmgr then attempts to connect to the server name mentioned and operates on that cluster. This functionality should behave the same.
```
[root@centos8 pbspro]# qmgr -c 's s host1,host2 scheduling=0'
Unknown Host.
qmgr: cannot connect to server host1
Unknown Host.
qmgr: cannot connect to server host2
```
If one of the server name is a multi-server cluster and one of the instance fails among them, qmgr should not operate on that cluster as a whole. For that, we do not create that server object.

I've moved my error checking inside make_connection so server structure is not created.

src/cmds/qstat.c
```diff
@@ -804,7 +805,31 @@ altdsp_statjob(struct batch_status *pstat, struct batch_status *prtheader, int a
 		pstat = bs_isort(pstat, cmp_est_time);
 
 	if (prtheader) {
-		printf("\n%s: ", prtheader->name);
+		svr_conn_t *svr_connections = get_conn_svr_instances(conn);
```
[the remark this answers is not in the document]
nithinj: get_conn_svr_instances will return a single instance in case of a single server.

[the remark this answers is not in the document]
nithinj: get_conn_svr_instances will return a single instance in the case of a single server.

src/lib/Libcmds/err_handling.c
```diff
@@ -0,0 +1,66 @@
...
+ * 
+ * @return void
+ */
+void
+show_svr_inst_fail(int fd, char *client)
+{
+	if (msvr_mode()) {
+		int i;
+		svr_conn_t *svr_connections = get_conn_svr_instances(fd);
+		for (i = 0; i < get_num_servers(); i++) {
```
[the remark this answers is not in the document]
nithinj: It's a single liner. Compiler should be making it inline.

src/lib/Libifl/int_manager.c
```diff
@@ -86,31 +103,58 @@ PBSD_manager(int c, int function, int command, int objtype, char *objname, struc
...
+			return rc;
+		}
+
+		/* read reply from stream into presentation element */
+		reply = PBSD_rdrpy(c);
+		PBSD_FreeReply(reply);
+
+		rc = get_conn_errno(c);
+		if (rc != 0)
+			agg_rc = rc;
```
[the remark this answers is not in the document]
nithinj: We will bail out if we cannot reach one of the servers. Which is handled in make_connection function within qmgr itself. This is a slightly bigger issue. The command can fail due to internal server error such as cannot delete an object which is busy etc. This will be handled in the future with two-phase commit or so.

Right now the admin will get an error message on the failure and had to make sure that objects are in sync.

src/lib/Libifl/int_status.c
```diff
@@ -46,7 +46,148 @@
...
+ * @return int
+ * @retval -1: error
+ * @retval != -1 fd corresponding to the connection
+ */
+int
+get_available_conn(svr_conn_t *svr_connections)
+{
+	int i;
+
+	for (i = 0; i < get_num_servers(); i++)
```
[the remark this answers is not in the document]
nithinj: inline function.

src/lib/Libifl/int_status.c
```diff
@@ -81,7 +221,472 @@ PBSD_status(int c, int function, char *objid, struct attrl *attrib, char *extend
...
+ * aggr_job_ct:
+ *	aggregate the job counts attribute of servers
+ *	total_jobs and state_counts are the attributes aggregated here.
+ * @param[in] cur - batch status to first server
+ * @param[in] next - batch status to second server
+ * 
+ * @return void
+ */
+static void
+aggr_job_ct(struct batch_status *cur, struct batch_status *nxt)
```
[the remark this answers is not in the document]
nithinj: IFL is not waiting for all server response to start processing the result. It just starts with two and processes the rest. It makes sense as servers might be busy processing other things and that time is getting utilized. We can also go one step further and process the ones which are readily available instead of following the order.

If we just want to cut multiple decodes and encodes of the first server, I could save them in a static variable. This will make the function not multi-threadable. But we can revisit this later if we attempt multi-threading for this functions.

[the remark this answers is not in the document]
nithinj: In fact, aggr_job_ct is common for both server and queue objects. This optimization does not work with queues as there can be multiple objects with the same id. Can we take it up later if we see it as a performance bottleneck?

src/lib/Libifl/int_status.c
```diff
@@ -81,7 +221,472 @@ PBSD_status(int c, int function, char *objid, struct attrl *attrib, char *extend
...
+ * @brief
+ * aggregate_svr:
+ *	aggregate two servers reported by two server instances
+ * @param[in,out] sv1 - server1 list. This will be also the final aggregated list.
+ * @param[in] sv2 - Server2 list.
+ * 
+ * @return void
+ */
+static void
+aggregate_svr(struct batch_status *sv1, struct batch_status *sv2)
```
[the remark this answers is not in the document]
nithinj: Answered above!

src/lib/Libifl/int_status.c
```diff
@@ -81,7 +221,472 @@ PBSD_status(int c, int function, char *objid, struct attrl *attrib, char *extend
...
+ * @param[in,out] a - attr list where b needs to be appended
+ * @param[in] b - batch status which needs to be appended to a
+ * @param[in,out] prev_b - previous list element of b for which references needs to be updated
+ * @param[in] head_a - head element of list contains a
+ * @param[in,out] head_b - reference to head element of list contains b. Reference is updated if prev_b is null
+ * 
+ * @return void
+ */
+static void
+append_bs(struct batch_status *a, struct batch_status *b, struct batch_status *prev_b, struct batch_status *head_a, struct batch_status **head_b)
```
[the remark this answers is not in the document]
nithinj: This is a static function and only getting used within this file. I can change it to append_bs_spl if you still feel this to be changed.

nithinj:
Hi @arungrover ,

src/lib/Libifl/int_status.c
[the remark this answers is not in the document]
arungrover: Your reasoning sounds good to me.
I am not worried about performance in this case because there are only a handful of servers or queues, it just looked odd that you had to encode every time only to decode it back.

src/lib/Libifl/int_status.c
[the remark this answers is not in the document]
arungrover: Yes, please change it, since it moves batch status from one and appends into another, maybe name it as move_and_append_bs or something similar.

src/cmds/qstat.c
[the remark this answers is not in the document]
arungrover: That's true, but you were not using anything it returns unless it is in multiserver mode. So why even call it if you are not in multiserver mode.
I'm fine with how you have it now.

src/lib/Libifl/int_status.c
```diff
@@ -128,38 +129,21 @@ static char *statename[] = { "Transit", "Queued", "Held", "Waiting",
 static void
 decode_states(char *string, long *count)
 {
-	char *c, *s;
-	long *d;
+	char *c;
 	int i;
+	char format[10 * MAX_STATE];
```
arungrover, line 134: It's rather better to make it a static string than to create it everytime. Also, if someone changes the order of statenames it will mess your string up.

src/lib/Libifl/int_status.c
```diff
@@ -128,38 +129,21 @@ static char *statename[] = { "Transit", "Queued", "Held", "Waiting",
...
-				c++;
-		} else {
-			while (*c != ' ' && *c != '\0')
-				c++;
-		}
-	}
+	for (i = 0; i < MAX_STATE; i++)
+		len += sprintf(format + len, "%s:%%ld ", statename[i]);
+
+	sscanf(c, format, &count[TRANSIT_STATE], &count[QUEUE_STATE],
```
arungrover, line 144: check for its return value to be 7 to ensure the order is matching your format.

arungrover (changes requested):
only two nits

src/lib/Libecl/ecl_verify_object_name.c
```diff
@@ -56,6 +56,7 @@
 #include "pbs_ecl.h"
 #include "pbs_error.h"
 #include "pbs_nodes.h"
+#include "libpbs.h"
```
suresh-thelkar, line 59: This is the only addition in this file. Do we really need this header ?

src/lib/Libifl/int_manager.c
```diff
@@ -86,31 +103,58 @@ PBSD_manager(int c, int function, int command, int objtype, char *objname, struc
...
 
 	/* now verify the attributes, if verification is enabled */
-	if ((pbs_verify_attributes(c, function, objtype,
-		command, aoplp)) != 0)
-		return pbs_errno;
-
-	/* lock pthread mutex here for this connection */
-	/* blocking call, waits for mutex release */
-	if (pbs_client_thread_lock_connection(c) != 0)
+	if ((pbs_verify_attributes(random_srv_conn(svr_connections), function, objtype, command, aoplp)))
```
suresh-thelkar, line 106: It would be better if we use pbs_verify_attributes_wrapper() here. It is because we can catch early if one of the servers is down i.e. this wrapper function returns PBSE_NOSERVER.

src/lib/Libifl/int_status.c
```diff
@@ -46,7 +46,132 @@
...
+#include "pbs_internal.h"
+
+
+extern char * PBS_get_server(char *, char *, uint *);
+
+
+/**
+ * @brief
+ * 	get_available_conn:
+ *	get one of the available connection from multisvr sd
```
suresh-thelkar, line 64: It looks like this function returns a socket descriptor given a list of server connections. The input is a server instances list not a sd. Please update the description accordingly.

src/lib/Libifl/int_status.c
```diff
@@ -46,7 +46,132 @@
...
+random_srv_conn(svr_conn_t *svr_connections)
+{
+	int ind = 0;
+
+	ind =  rand_num() % get_num_servers();
+
+	if (svr_connections[ind].state == SVR_CONN_STATE_UP)
+		return svr_connections[ind].sd;
+		
+	return get_available_conn(svr_connections);
```
suresh-thelkar, line 104: When we reach here which means the chosen random server is down. So when we call the function get_available_conn() we should only look a random server from the remaining servers right ?. Here it seems again we are looping through all the available servers inside get_available_conn.

src/lib/Libifl/int_status.c
```diff
@@ -46,7 +46,132 @@
...
+ * 
+ * @return void
+ *
+ */
+static void
+decode_states(char *string, long *count)
+{
+	char *c;
+	int i;
+	char format[10 * MAX_STATE];
```
suresh-thelkar, line 134: can you please add a comment what the number 10 stands for ?

src/lib/Libifl/int_status.c
```diff
@@ -46,7 +46,132 @@
...
+	int len = 0;
+	char buf[256];
+
+	buf[0] = '\0';
+	for (index = 0; index < MAX_STATE; index++) {
+		len += sprintf(buf + len, "%s:%ld ", statename[index],
+			       cur[index] + nxt[index]);
+	}
+	free(*val);
+	*val = strdup(buf);
```
suresh-thelkar, line 173: is NULL check not required for val here ?

src/lib/Libifl/int_status.c
```diff
@@ -81,7 +205,471 @@ PBSD_status(int c, int function, char *objid, struct attrl *attrib, char *extend
...
+
+	if (!svr_connections)
+		return NULL;
+
+	failed_conn = calloc(num_cfg_svrs, sizeof(int));
+
+	if (pbs_client_thread_init_thread_context() != 0)
+		return NULL;
+
+	if (pbs_verify_attributes(random_srv_conn(svr_connections), cmd, parent_object, MGR_CMD_NONE, (struct attropl *) attrib) != 0)
```
suresh-thelkar, line 544: Here also it is better to use pbs_verify_attributes_wrapper() so that we can catch PBSE_NOSERVER early if it happens.

src/lib/Libifl/int_status.c
nithinj: It does not worth saving the failed connection as the cost we are saving will be spent in the form of an if condition.

src/lib/Libifl/int_status.c
nithinj: It does not matter as status API's has to give partial output.

nithinj:
@arungrover  and @suresh-thelkar 
I have addressed your recent comments. Please have a look.

arungrover (approved):
Thanks for making the change, LGTM.

subhasisb (approved):

src/lib/Libecl/ecl_verify_values.c
```diff
@@ -1999,3 +1999,52 @@ verify_value_preempt_sort(int batch_request, int parent_object, int cmd,
...
+ * @retval 	-1 - System error verifying attributes (pbs_errno is set)
+ *
+ * @par		Side effects:
+ *		Modifies the TLS data for this thread\n
+ *		pbs_errno is set on encourtering error
+ *
+ * @par MT-safe: Yes
+ */
+int
+pbs_verify_attributes_wrapper(int fd, int batch_request,int parent_object, int cmd, struct attropl *attribute_list)
```
subhasisb, line 2028: pbs_verify_attributes() actually does not connect to any server at all. It takes the fd only to store the error code in it. So do we really need to loop over all the server instances to verify the same attribute list again and again?

subhasisb (changes requested):

subhasisb:
@nithinj please also attach valgrind test results. Also check for compilation/test on other platforms

src/lib/Libifl/int_status.c
suresh-thelkar: Why can't we simply mark the connection as SVR_CONN_STATE_DOWN so that this particular connection is not exercised again in get_available_conn. If we do this then we no need to change the code of get_available_conn.

src/lib/Libifl/int_status.c
suresh-thelkar: Yes but what if the server that pbs_verify_attributs() chooses is down, we still need to talk to another server right ?

suresh-thelkar:
@nithinj, Code changes look good to me. I have couple of minor comments. Can you please address them.

src/lib/Libifl/int_status.c
nithinj: Can you clarify a bit more? This function will find a random fd. If that fd is not connected then it will loop and get an available connection. That fd would have already marked as down. Otherwise, the function will return the same fd from here.

src/lib/Libifl/int_status.c
nithinj: That is handled today within random_srv_conn

src/lib/Libecl/ecl_verify_values.c
nithinj: You are right. I had placed this to make sure that IFL does not proceed without making sure all connections are up as qmgr stays connected. But qmgr gets disconnected if one of the connection breaks anyway.

nithinj:
Hi @subhasisb , @arungrover , @suresh-thelkar 

Thanks for the reviews and sign-off. I've checked valgrind and built on other platforms. Please take it further.

arungrover:
LGTM

subhasisb (approved):

src/lib/Libifl/pbsD_connect.c
```diff
@@ -428,7 +429,8 @@ dealloc_conn_list_single(int parentfd)
 
 	conn_list = pthread_getspecific(psi_key);
 	for (iter_conns = conn_list; iter_conns != NULL; prev = iter_conns, iter_conns = iter_conns->next) {
-		if (iter_conns->conn_arr[0].sd == parentfd) {
+		if ((iter_conns->conn_arr[0].sd == parentfd) ||
+		    (parentfd == -1 && !iter_conns->next)) {
```
agrawalravi90, line 433: im not sure i understand this change. it seems like we want to delete iter_conns[0] when parentfd is -1, but only when there's a single set of connections. Can you please explain why this is needed?

src/lib/Libifl/pbsD_connect.c
```diff
@@ -932,6 +911,7 @@ pbs_connect_noblk(char *server, int tout)
 			default:
 err:
 				closesocket(sock);
+				dealloc_conn_list_single(-1);
```
agrawalravi90, line 914: Instead of doing this, why not connect() first, and only create a new svr_connections[] if connect was successful? You won't need any of these dealloc calls then

agrawalravi90 (changes requested):
Reviewed the pbs_connect part of it only

PR #2068, opened 2020-10-09
Making Scheduler ready for Multi-Server

suresh-thelkar:
#### Describe Bug or Feature

This is an attempt to make Scheduler ready for Multi-Server as much as possible without impacting the single server functionality and also test Multi-Server as much as possible. For more details on Multi-Server feature please refer to the following.

https://openpbs.atlassian.net/wiki/spaces/PD/pages/1507164161/PBS+Pro+multi-server+project+architecture+and+design

#### Describe Your Change

Following changes are done as part of this PR.

- Scheduler as of today maintains its own connection table to connect to all servers in different PBS complexes where as IFL layer already has its mechanisum/data structures to talk to multiple servers which is already getting used by other clients like qsub, qstat etc. Please refer to the PR #2036 for more details. Having the connection info in both Scheduler and IFL layer is redundant and very hard to maintain the code. So Scheduler code is refactored to use the same connection table that is present inside IFL layer to accomplish its needs.

- Since pbs_connect itself connects to all the servers configured via PBS_SERVER_INSTANCES switch, Scheduler does not need to connect to each server independently. So it calls pbs_connect once which returns a socket descriptor which represents the whole set of servers i.e. this descriptor can be treated as a key to get all the individual servers connection information. (named as clust_primary_sock)

- Since Scheduler also needs a secondary fd to talk to each server to get scheduling commands and super high priority commands it again calls pbs_connect which returns a socket descriptor(named as clust_secondary_sock) which again represents the whole set of servers configured. 

- To get the individual servers connection details we can make use of the following APIs

```
svr_conn_t *svr_conns_primary;
svr_conn_t *svr_conns_secondary;

svr_conns_primary =  get_conn_servers(clust_primary_sock);
svr_conns_secondary =  get_conn_servers(clust_secondary_sock);
```
- Scheduler receives scheduling commands from a server and puts them in a queue. There are no duplicate commands if there is a single server, but in case of multiple servers there is a possibility that duplicate commands get accumulated in the queue which we should eliminate. Code changes are added to filter all the received commands so as to
        1) eliminate duplicate scheduling commands in the queue
        2) eliminate commands that represents a subset when there is already a super set 
             command. For example suppose if we have SCH_SCHEDULE_AJOB and 
             SCH_SCHDULE_NEW we can discard SCH_SCHEDULE_AJOB.

- The criteria to filter scheduling commands and to find out which is superset or subset command is all hidden inside the function compare_modify_sched_cmd. This way it is more flexible even if a new scheduling command is added in future.


suresh-thelkar:
As I have to run regression and valgrind marking this as WIP. Meanwhile the PR is open for reviews.

hirenvadalia:
@suresh-thelkar One thought on overall architecture of this PR:
How about we introduce below very special IFL calls:
* `pbs_register_sched()`/`pbs_connect_and_register_sched()` - This will internally connect to all servers and send register sched request to all server and return cluster fd (aka fd of 0th index in svr conn array)
* `pbs_get_sched_cmds()`/`pbs_pending_sched_cmds()` - This will read all sched commands from all server and return array of sched cmds (if needed we can put sorting and removing duplicates/filtering etc also in this api). Also this api will take one argument which will indicate that cmds should be read in blocking mode or non-blocking mode (to read high prio cmd while we are in-between sched cycle)

By doing above we can completely hide accessing svr conn struct (which is internal to IFL) in sched code.

suresh-thelkar:
> @suresh-thelkar One thought on overall architecture of this PR:
> How about we introduce below very special IFL calls:
> 
> * `pbs_register_sched()`/`pbs_connect_and_register_sched()` - This will internally connect to all servers and send register sched request to all server and return cluster fd (aka fd of 0th index in svr conn array)
> * `pbs_get_sched_cmds()`/`pbs_pending_sched_cmds()` - This will read all sched commands from all server and return array of sched cmds (if needed we can put sorting and removing duplicates/filtering etc also in this api). Also this api will take one argument which will indicate that cmds should be read in blocking mode or non-blocking mode (to read high prio cmd while we are in-between sched cycle)
> 
> By doing above we can completely hide accessing svr conn struct (which is internal to IFL) in sched code.

Thanks Hiren. As discussed I have coded them which appears like IFL API for example pbs_register_sched() which is used to register Scheduler with all the configured servers. I purposefully did not make them part of Libifl. It is because they are not useful for all other clients like qstat,qsub etc.  except for Scheduler. 

Similarly filtering scheduling commands so as to eliminate duplicates and having only the deserved commands in the scheduling commands queue is actually application logic which we can always build on top of generic IFL or IFL like functions. Again this is something that is not used by other clients except the Scheduler.

hirenvadalia:
@suresh-thelkar I agree with you that those IFL calls will be used only by sched (and possibly by other sched like tools/programs), but having that will benefit us in hiding internals of IFL (aka svr conn struct) completely in IFL layer, which is our one of the goal also right?

@subhasisb What you think? Having IFL calls will be good or not?

bhroam:
@suresh-thelkar your example says you will replace SCHEDULE_AJOB with SCHEDULE_NEW.  This will break qrun.  The server keeps track of several scheduler commands, including *ALL* qruns.  It will send them to the scheduler one after another.  If you replace SCHEDULE_AJOB with SCHEDULE_NEW, the scheduler will not run the qrun job, and that qrun will sit there forever waiting.  You need to queue up all SCHEDULE_AJOB commands.

suresh-thelkar:
> @suresh-thelkar your example says you will replace SCHEDULE_AJOB with SCHEDULE_NEW. This will break qrun. The server keeps track of several scheduler commands, including _ALL_ qruns. It will send them to the scheduler one after another. If you replace SCHEDULE_AJOB with SCHEDULE_NEW, the scheduler will not run the qrun job, and that qrun will sit there forever waiting. You need to queue up all SCHEDULE_AJOB commands.

Thanks @bhroam. I am replacing SCHEDULE_AJOB with SCHEDULE_NEW only when Scheduler sees this two commands together in its scheduling queue(This can happen if one server sending SCHEDULE_AJOB and another server sending SCHEDULE_NEW)  . I am not doing this replacement at the Server side.  Also when both the servers send SCHEDULE_AJOB independently that time I am not replacing. For more details please look into the function compare_modify_sched_cmd and where it gets called. I have tried the following scenario just to make sure whether this is ok. Please let me know if I have to try any other scenario where it does not work.

1. Turn off scheduling.
`[root@stblr3 pbspro]# qmgr -c "s s scheduling=0"`

2. Submit couple of jobs.
```
[root@stblr3 pbspro]# qsub -- date
9.stblr3
[root@stblr3 pbspro]# qsub -- date
10.stblr3
[root@stblr3 pbspro]# qstat -s

stblr3: 
                                                            Req'd  Req'd   Elap
Job ID          Username Queue    Jobname    SessID NDS TSK Memory Time  S Time
--------------- -------- -------- ---------- ------ --- --- ------ ----- - -----
9.stblr3        root     workq    STDIN         --    1   1    --    --  Q   -- 
    -- 
10.stblr3       root     workq    STDIN         --    1   1    --    --  Q   -- 
```
   
3. Put Scheduler under debug mode(having breakpoint at schedule function).

4. qrun the above jobs
` [root@stblr3 pbspro]# qrun 9 10`

5. Before schedule function gets called through gdb change SCH_SCHEDULE_AJOB to SCH_SCHEDULE_NEW as follows.
```
(gdb) p	*cmd
$19 = {cmd = 12, jid = 0x2195470 "9.stblr3", from_sock = 7}
(gdb) s	cmd->cmd=1
schedule (sd=6,	cmd=0x21975e0) at fifo.c:491

(gdb) p	*cmd
$20 = {cmd = 1,	jid = 0x2195470 "9.stblr3", from_sock = 7}

(gdb) p	*cmd
$22 = {cmd = 12, jid = 0x216c910 "10.stblr3", from_sock = 7}
(gdb) s cmd->cmd=1
schedule (sd=6,	cmd=0x2197600) at fifo.c:491
(gdb) p	*cmd
$23 = {cmd = 1,	jid = 0x216c910 "10.stblr3", from_sock = 7}

```
Had to do the above twice because scheduler gets two commands.

5. Run qstat to make sure that jobs ran fine.
```
[root@stblr3 pbspro]# qstat -s
[root@stblr3 pbspro]# 
```
qrun succeeded and both jobs ran fine.

suresh-thelkar:
> @suresh-thelkar I agree with you that those IFL calls will be used only by sched (and possibly by other sched like tools/programs), but having that will benefit us in hiding internals of IFL (aka svr conn struct) completely in IFL layer, which is our one of the goal also right?

Following is my opinion on this. We can wait and see what others thoughts on this.

Scheduler reads commands from a server through well defined APIs like get_sched_cmd and get_sched_cmd_noblk which means they are already encapsulated only thing is they are not part of libifl. The reasons for not keeping them in LibIFL is
- No other clients need them
- these APIs do not send any batch request to server and gets a response. These APIs directly read the data from server's raw sockets and don't send any response back to server that the commands are read successful or any sort of return code saying that scheduler received them successfully i.e. in other words they are not directly adhering to batch request/response protocol. 
So I think the way they are designed is good enough for now. We can always revisit them and make them 100% encapsulated or put them as part of another layer like IFL if there is any need in future.

subhasisb:
> @suresh-thelkar I agree with you that those IFL calls will be used only by sched (and possibly by other sched like tools/programs), but having that will benefit us in hiding internals of IFL (aka svr conn struct) completely in IFL layer, which is our one of the goal also right?
> 
> @subhasisb What you think? Having IFL calls will be good or not?

Perhaps an "internal only" IFL call is good enough? I was thinking more like pbs_internal.h

src/scheduler/fifo.c
```diff
@@ -479,14 +479,14 @@ init_scheduling_cycle(status *policy, int pbs_sd, server_info *sinfo)
...
- * @param[in]	sconn	-	connection and command info from server
+ * @param[in]	sd	-	primary socket descriptor to the server pool
  *
  * @return	int
  * @retval	0	: continue calling scheduling cycles
  * @retval	1	: exit scheduler
  */
 int
-schedule(sched_svrconn *sconn, sched_cmd *cmd)
+schedule(int sd, sched_cmd *cmd)
```
hirenvadalia, line 489: Do we really need sd as argument? now even with multi-server we have only one sd only 'cluster fd', and which is global variable anyway. So can we get rid of sd argument (including all dependent funcs) and use global cluster fd in IFL calls during sched cycle?

suresh-thelkar:
> Perhaps an "internal only" IFL call is good enough? I was thinking more like pbs_internal.h

Myself and @subhasisb had a offline discussion regarding the above comments. Following is the summary of our discussion.

- Since get_sched_cmd()/get_sched_cmd_no_blk() are very much scheduler specific we are not going to make them part of Libifl. Also they do not send any batch request and get a response like the other IFL APIs.
- However pbs_register_sched() internally calls send_register_sched() which registers Scheduler with a specific server by sending a batch request PBS_BATCH_RegisterSched. So the function send_register_sched() can be made of as part of Libifl.

hirenvadalia:
> > Perhaps an "internal only" IFL call is good enough? I was thinking more like pbs_internal.h
> 
> Myself and @subhasisb had a offline discussion regarding the above comments. Following is the summary of our discussion.
> 
> * Since get_sched_cmd()/get_sched_cmd_no_blk() are very much scheduler specific we are not going to make them part of Libifl. Also they do not send any batch request and get a response like the other IFL APIs.
> * However pbs_register_sched() internally calls send_register_sched() which registers Scheduler with a specific server by sending a batch request PBS_BATCH_RegisterSched. So the function send_register_sched() can be made of as part of Libifl.

But just moving only send register part in ifl doesn't help anything... we still be accessing internal structure of IFL... If we move all part where we are accessing internal structure of IFL to libifl then only will get full benefits, else there is no benefit at all.

Also if needed we can convert send/recv ched cmd part to have proper batch request/reply as they are internal and never exposed so chnaging then wouldn't break anything.

suresh-thelkar:
> But just moving only send register part in ifl doesn't help anything... we still be accessing internal structure of IFL... If we move all part where we are accessing internal structure of IFL to libifl then only will get full benefits, else there is no benefit at all.
> 
> Also if needed we can convert send/recv ched cmd part to have proper batch request/reply as they are internal and never exposed so chnaging then wouldn't break anything.

According to the discussion that I had with Subhasis as I summarised already the idea is not to put anything scheduler specific in Libifl and hence we decided only to put send_register_sched in IFL as it perfectly qualifies to be an IFL API.
I believe there is no value add if we also come up with IFLs for reading a sched command also. This unnecessary brings lot of confusion to the table because these IFLs now act on secondary set of connections and they don't adhere to batch request/response. If we really want to make this, then we need to send and receive batch requests just to read commands from Server that is kind of overkill I think.

hirenvadalia:
> > But just moving only send register part in ifl doesn't help anything... we still be accessing internal structure of IFL... If we move all part where we are accessing internal structure of IFL to libifl then only will get full benefits, else there is no benefit at all.
> > Also if needed we can convert send/recv ched cmd part to have proper batch request/reply as they are internal and never exposed so chnaging then wouldn't break anything.
> 
> According to the discussion that I had with Subhasis as I summarised already the idea is not to put anything scheduler specific in Libifl and hence we decided only to put send_register_sched in IFL as it perfectly qualifies to be an IFL API.
> I believe there is no value add if we also come up with IFLs for reading a sched command also. This unnecessary brings lot of confusion to the table because these IFLs now act on secondary set of connections and they don't adhere to batch request/response. If we really want to make this, then we need to send and receive batch requests just to read commands from Server that is kind of overkill I think.

True, but my original point was not to access internal structures of IFL outside of IFL/libifl, but @subhasisb and others think its ok to access it then I am fine with it, but we can't truly say that server connection structure in IFL is internal to IFL, means that structure is neither properly exposed structure nor internals of IFL :smile:

subhasisb:
> > > But just moving only send register part in ifl doesn't help anything... we still be accessing internal structure of IFL... If we move all part where we are accessing internal structure of IFL to libifl then only will get full benefits, else there is no benefit at all.
> > > Also if needed we can convert send/recv ched cmd part to have proper batch request/reply as they are internal and never exposed so chnaging then wouldn't break anything.
> > 
> > 
> > According to the discussion that I had with Subhasis as I summarised already the idea is not to put anything scheduler specific in Libifl and hence we decided only to put send_register_sched in IFL as it perfectly qualifies to be an IFL API.
> > I believe there is no value add if we also come up with IFLs for reading a sched command also. This unnecessary brings lot of confusion to the table because these IFLs now act on secondary set of connections and they don't adhere to batch request/response. If we really want to make this, then we need to send and receive batch requests just to read commands from Server that is kind of overkill I think.
> 
> True, but my original point was not to access internal structures of IFL outside of IFL/libifl, but @subhasisb and others think its ok to access it then I am fine with it, but we can't truly say that server connection structure in IFL is internal to IFL, means that structure is neither properly exposed structure nor internals of IFL smile

The two functions in question in get_4byte.c does not necessarily access internal IFL structures. The caller passes an fd and these functions simply work on that fd. If you recall, we anyway work with the fd directly in many other cases, so that should be okay here. What we do not want to do, is access the connections structure inside these functions directly.

subhasisb:
> > > > But just moving only send register part in ifl doesn't help anything... we still be accessing internal structure of IFL... If we move all part where we are accessing internal structure of IFL to libifl then only will get full benefits, else there is no benefit at all.
> > > > Also if needed we can convert send/recv ched cmd part to have proper batch request/reply as they are internal and never exposed so chnaging then wouldn't break anything.
> > > 
> > > 
> > > According to the discussion that I had with Subhasis as I summarised already the idea is not to put anything scheduler specific in Libifl and hence we decided only to put send_register_sched in IFL as it perfectly qualifies to be an IFL API.
> > > I believe there is no value add if we also come up with IFLs for reading a sched command also. This unnecessary brings lot of confusion to the table because these IFLs now act on secondary set of connections and they don't adhere to batch request/response. If we really want to make this, then we need to send and receive batch requests just to read commands from Server that is kind of overkill I think.
> > 
> > 
> > True, but my original point was not to access internal structures of IFL outside of IFL/libifl, but @subhasisb and others think its ok to access it then I am fine with it, but we can't truly say that server connection structure in IFL is internal to IFL, means that structure is neither properly exposed structure nor internals of IFL smile
> 
> The two functions in question in get_4byte.c does not necessarily access internal IFL structures. The caller passes an fd and these functions simply work on that fd. If you recall, we anyway work with the fd directly in many other cases, so that should be okay here. What we do not want to do, is access the connections structure inside these functions directly.
> 
> I checked once more and see that get_sched_cmd() also does a disrst(), it would be better to move this function to IFL, since it is actually doing a internal read

on second thoughts and further discussion, i agree that it is best to not change anything right now. The way we read the commands is a bit dependent on the scheduler also, so if we hide the usage or disrst inside an IFL call, then that function will also need to interpret the "scheduler command" and do actions specific to it, like reading further data - jobid. Of course, we can modify the server to always send command/data pairs (even if data is empty string) and then IFL would not need to know about the command - but that is more change than required right now

suresh-thelkar:
> on second thoughts and further discussion, i agree that it is best to not change anything right now. The way we read the commands is a bit dependent on the scheduler also, so if we hide the usage or disrst inside an IFL call, then that function will also need to interpret the "scheduler command" and do actions specific to it, like reading further data - jobid. Of course, we can modify the server to always send command/data pairs (even if data is empty string) and then IFL would not need to know about the command - but that is more change than required right now

Thank you @subhasisb.

src/scheduler/pbs_sched.c
```diff
@@ -1273,3 +1188,140 @@ main(int argc, char *argv[])
...
+	
+	if (new_cmd->cmd == SCH_SCHEDULE_AJOB &&
+		(cmd_cache->cmd != SCH_SCHEDULE_AJOB && cmd_cache->cmd != SCH_CONFIGURE &&
+		 cmd_cache->cmd != SCH_RULESET && cmd_cache->cmd != SCH_ERROR)) {
+		return 1;
+	} else if (cmd_cache->cmd == SCH_SCHEDULE_AJOB &&
+		(new_cmd->cmd != SCH_SCHEDULE_AJOB && new_cmd->cmd != SCH_CONFIGURE &&
+		 new_cmd->cmd != SCH_RULESET && new_cmd->cmd != SCH_ERROR)) {
+		cmd_cache->cmd = new_cmd->cmd;
+		cmd_cache->jid = NULL;
```
arungrover, line 1225: You loose the job id here. How will we ever honor this qrun request?

src/scheduler/pbs_sched.c
```diff
@@ -1273,3 +1188,140 @@ main(int argc, char *argv[])
...
+	if (cmd_enq == NULL)
+		goto error;
+
+	cmd_enq->cmd = tmp_cmd->cmd;
+	cmd_enq->from_sock = tmp_cmd->from_sock;
+	if (tmp_cmd->jid != NULL) {
+		cmd_enq->jid = malloc(strlen(tmp_cmd->jid) + 1);
+		if (cmd_enq->jid == NULL)
+			goto error;
+		strcpy(cmd_enq->jid, tmp_cmd->jid);
```
arungrover, line 1281: why are we not checking this strcpy?

src/scheduler/pbs_sched.c
```diff
@@ -1273,3 +1188,140 @@ main(int argc, char *argv[])
...
+ *
+* @param[in] sched_cmds	- scheduler command queue
+ *
+ * @return	ds_queue *
+ * @retval	success	- returns the filtered queue
+ * @reval	 fail -	NULL
+ *
+ */
+ds_queue *
+filter_sched_cmds(ds_queue *sched_cmds)
```
arungrover, line 1253: Can you please add some comments to this function? It isn't easy to follow.

src/scheduler/pbs_sched.c
```diff
@@ -1273,3 +1188,140 @@ main(int argc, char *argv[])
...
+		} else {
+			if (cmd_cache->cmd == new_cmd->cmd && !strcmp(cmd_cache->jid, new_cmd->jid))
+				return 1;
+		}
+	}
+	return 0;
+}
+
+/**
+ * @brief Filters scheduler commands queue according to the criteria given in the function
```
arungrover, line 1242: I think that since in our case commands have always been internal to PBS, we can renumber them in the priority order and use their id as an index in commands array.
Like I'll probably make RESTART_CYCLE as number 0, SCH_SCHEDULE_AJOB as 1.... and SCH_SCHEDULE_NEW as number 17.
So the highest priority jobs will be the first in the array and others will be behind it.
When you set an index based on what you receive, you can also turn off other similar commands like SCH_SCHEDULE_STARTQ will unset SCH_SCHEDULE_NEW etc.
Scheduler can then go through the array of commands and start to honor them one after another.
The only exception I think is in the case of SCH_SCHEDULE_AJOB. In that case, you will have to append the job id into another list and not turn off the bit in commands array until all are done.

src/scheduler/pbs_sched.c
```diff
@@ -1273,3 +1188,140 @@ main(int argc, char *argv[])
...
+				goto error;
+			tmp_cmd = sched_cmds->content_arr[cmds_front];
+
+			cmd_enq->cmd = tmp_cmd->cmd;
+			cmd_enq->from_sock = tmp_cmd->from_sock;
+			if (tmp_cmd->jid != NULL) {
+				cmd_enq->jid = malloc(strlen(tmp_cmd->jid) + 1);
+				if (cmd_enq->jid == NULL)
+					goto error;
+				strcpy(cmd_enq->jid, tmp_cmd->jid);
```
arungrover, line 1313: here as well, not sure why we are not checking the return value.

src/scheduler/pbs_sched.c
```diff
@@ -1273,3 +1188,140 @@ main(int argc, char *argv[])
...
+		if (cmd_enq->jid == NULL)
+			goto error;
+		strcpy(cmd_enq->jid, tmp_cmd->jid);
+	}
+
+	if (!ds_enqueue(sched_cmds_filtered, cmd_enq)) 
+		goto error;
+	cmds_front++;
+
+	for (; cmds_front < cmds_rear; cmds_front++) {
```
arungrover, line 1288: Can you please explain how it will work when multiple connected servers are sending multiple schedule_ajob requests with unique job-ids.

arungrover (changes requested):

src/scheduler/fifo.c
hirenvadalia: @arungrover We can set global cluster fd variable to simulation fd.

src/lib/Libifl/pbsD_connect.c
```diff
@@ -1011,3 +1013,95 @@ pbs_connect_noblk(char *server, int tout)
...
+	svr_conn_t *svr_conns_primary = NULL;
+	svr_conn_t *svr_conns_secondary = NULL;
+	svr_conns_list_t *iter_conns = NULL;
+
+	if (client_id == NULL)
+		return 0;
+
+	conn_list = pthread_getspecific(psi_key);
+
+	for (iter_conns = conn_list; iter_conns != NULL; iter_conns = iter_conns->next) {
```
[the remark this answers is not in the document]
suresh-thelkar: Yes Arun. Thanks for catching this. We can directly make use of iter_conns->conn_arr here. Please look into my latest commit for code changes.

src/lib/Libifl/pbsD_connect.c
```diff
@@ -1011,3 +1013,95 @@ pbs_connect_noblk(char *server, int tout)
...
+		/* conn_list has secondary set first followed by primary set of connections */
+		if (svr_conns_secondary == NULL)
+			svr_conns_secondary = get_conn_servers(iter_conns->conn_arr[0].sd);
```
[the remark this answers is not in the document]
suresh-thelkar: Yes this assumption is always true. This is how IFL layer stores a set of connections. It is because when a new set of connections come it is always inserted at the beginning.

src/lib/Libifl/pbsD_connect.c
```diff
@@ -1011,3 +1013,95 @@ pbs_connect_noblk(char *server, int tout)
...
+		else if (svr_conns_primary == NULL)
+			svr_conns_primary = get_conn_servers(iter_conns->conn_arr[0].sd);
+	}
+
+	for (i = 0; i < get_num_servers(); i++) {
+		if (!svr_conns_primary[i].registered) {
+			if (send_register_sched(svr_conns_primary[i].sd, client_id) == 0)
+				return 0;
```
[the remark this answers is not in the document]
suresh-thelkar: The connection entries are cleaned up in close_servers() function which in turn calls pbs_disconnect() on primary and secondary cluster sockets. Please look into these functions for more details. 
We intentionally do not a clean up a particular connection entry i.e. if there are three servers but one of the server is going down temporarily connection table at IFL layer still holds an entry for the server which is down. The idea is when the server again comes back it uses its own slot. This helps for upcoming features of Multi-Server like "Diff stat" feature for example.

src/scheduler/pbs_sched.c
suresh-thelkar: Here I am deep copying all the contents of tmp_cmd to cmd_enq. If malloc fails we goto the label "error" from which we return. Generally we don't check return value for strcpy. When we actually make use of cmd->jid then we anyways check for not null.

src/scheduler/pbs_sched.c
suresh-thelkar: I have added more comments in the code along with the algorithm. Please check my latest commit.

src/scheduler/pbs_sched.c
suresh-thelkar: Please look into my reply given above for the similar comment.

src/scheduler/pbs_sched.c
hirenvadalia: how about using strdup instead malloc+strcpy?

src/scheduler/pbs_sched.c
suresh-thelkar: I have added more details to the function header of compare_modify_sched_cmd along with couple of examples which covers this. Please check.

src/scheduler/pbs_sched.c
suresh-thelkar: As we know as of today Server marks scheduling commands as super high priority, medium priority and normal priority and sends them accordingly. Current PR does not change this behaviour. 

The job of this function filter_sched_cmds is 
- Filter duplicate scheduling commands. Duplicate commands are possible since multiple servers can kick off scheduling cycle. For example if we have two servers and both of them issuing a SCH_SCHEDULE_NEW. Scheduler needs to discard one command.
- Similarly if a scheduler has SCH_SCHEDULE_NEW and SCH_SCHEDULE_AJOB at a time it discards  SCH_SCHEDULE_AJOB
- If two servers sends SCH_SCHEDULE_AJOB each and we don't have any commands in scheduler's queue both of these commands are retained because jobids are different.
- All of this criteria is hidden inside the function compare_modify_sched_cmd.

I have added detailed descriptions in the code itself in these functions. Please take a look.

src/scheduler/pbs_sched.c
suresh-thelkar: Done. Please check my latest commit.

src/scheduler/fifo.c
hirenvadalia: Ah, got it.

src/scheduler/pbs_sched.c
bhroam: You can't discard SCHEDULE_AJOB.  When the scheduler receives a SCHEDULE_AJOB cycle, it does a special cycle for a specific job (and only that job).  This job gets super high priority during this cycle.  This includes doing things like ignoring limits and doing preemption with the highest possible preemption priority.  *EVERY* qrun job must have its own cycle.  We can't bunch these together or they lose their special priority.

In your previous testing, you just made sure that the jobs ran.  You didn't check to make sure they got the priority they needed and we overrode the policies they required.  Also, a qrun job has a special IFL call to return an error message if that job fails to run.  If you bunch these up, you will lose these (look at pbs_defschreply() at the end of scheduling_cycle())

arungrover: I feel the current PR is already changing the behavior in some way. The scheduler is now taking a decision that out of a handful of requests that has received from different servers which is the one it needs to honor and which is the one it can avoid. Isn't this prioritizing the commands?

In relation to the second point you mentioned, I think that is wrong. We have to preserve the job id because these jobs are treated with elevated preemption priority.
The way you have it now, if a qrun request gets overwritten by a normal sched cycle request, the job that is being qrun will not be treated in special way and may not preempt.

bhroam: So to keep things "the same" between a single server, and multi-server, we can do some command filtering.  Right now, the server keeps two command variables, and a list of all qrun requests.  The two command variables are a high priority one and a low priority one.  The high priority one is for things like SCH_CONFIGURE.  The normal priority one is for running scheduling cycles.  If multiple normal priority commands come in, the current one overwrites the previous one.  For the most part, all the normal scheduling commands will run a cycle.  If we wanted to extend this to multi-server, we could filter all the normal priority scheduling commands down to just one.  Until the server will send all non-duplicate commands, the scheduler can't trust the command anyway.  If it receives a SCH_SCHEDULE_NEW, it can't trust that there wasn't a SCH_SCHEDULE_TERM that came in right behind it.  If we want to start sending all non-duplicate commands from all servers, then the scheduler can make inferences based on the command.  Until we decide to do that, they all just get dumped into a 'run a scheduling cycle' bucket.

suresh-thelkar: > You can't discard SCHEDULE_AJOB. When the scheduler receives a SCHEDULE_AJOB cycle, it does a special cycle for a specific job (and only that job). This job gets super high priority during this cycle. This includes doing things like ignoring limits and doing preemption with the highest possible preemption priority. _EVERY_ qrun job must have its own cycle. We can't bunch these together or they lose their special priority.
> 
> In your previous testing, you just made sure that the jobs ran. You didn't check to make sure they got the priority they needed and we overrode the policies they required. Also, a qrun job has a special IFL call to return an error message if that job fails to run. If you bunch these up, you will lose these (look at pbs_defschreply() at the end of scheduling_cycle())

Thanks @bhroam for the detailed explanation. Modified the code changes so as not to discard or replace SCHEDULE_AJOB. Please look into my latest commit.

suresh-thelkar: > So to keep things "the same" between a single server, and multi-server, we can do some command filtering. Right now, the server keeps two command variables, and a list of all qrun requests. The two command variables are a high priority one and a low priority one. The high priority one is for things like SCH_CONFIGURE. The normal priority one is for running scheduling cycles. If multiple normal priority commands come in, the current one overwrites the previous one.
> 
> For the most part, all the normal scheduling commands will run a cycle. If we wanted to extend this to multi-server, we could filter all the normal priority scheduling commands down to just one. Until the server will send all non-duplicate commands, the scheduler can't trust the command anyway. If it receives a SCH_SCHEDULE_NEW, it can't trust that there wasn't a SCH_SCHEDULE_TERM that came in right behind it. If we want to start sending all non-duplicate commands from all servers, then the scheduler can make inferences based on the command. Until we decide to do that, they all just get dumped into a 'run a scheduling cycle' bucket.

Thank you @bhroam. It looks like you are suggesting to divide around 18 scheduling commands that we have right now into 3 buckets/groups 

Bucket-1: SCH_CONFIGURE/SCH_QUIT
Bucket-2: List of qrun requests
Bucket-3: Rest all commands to say "Run a sched cycle"

So Server will only send one command for each group. At the most Scheduler gets three types of commands and it filters the duplicates when there are multiple servers.  

I think these changes are really good to do which can actually go independently after the current PR. Hope this fine. Following are the reasons that made me to think in this direction.

- In the current PR I am not changing the number of scheduling commands and the way server sends commands to scheduler    i.e. I am not having any changes in Server whereas the proposed changes have lot of code changes in both server and scheduler side.
- The code changes done in this PR like making Scheduler ready for Multi-Server,  filtering sched commands mostly won't change even after the proposed changes happen after the current PR. Also completing the current PR as a first step helps other Multi-Server PRs which are dependent on this like IFL aggregation, implementing a scheduling policy that gives first preference to the local nodes, "Run and Movejob to other servers changes" in case if the server does not have the nodes associated with a job.

src/scheduler/pbs_sched.c
```diff
@@ -1273,3 +1193,127 @@ main(int argc, char *argv[])
...
+ *
+ * @return	ds_queue *
+ * @retval	success	- returns the filtered queue
+ * @reval	 fail -	NULL
+ *
+ */
+ds_queue *
+filter_sched_cmds(ds_queue *sched_cmds)
+{
+	/* sched_cmds_filtered is a  resultant queue which consists of sched commands after discarding
```
bhroam, line 1239: Looks like there is an extra space between 'a' and 'resultant'

src/scheduler/pbs_sched.c
```diff
@@ -1273,3 +1193,127 @@ main(int argc, char *argv[])
...
+	 * the duplicate commands which is finally returned from this function.
+	 */
```
bhroam, line 1241: Could you move the contents of this comment into the @brief above?

src/scheduler/pbs_sched.c
```diff
@@ -1273,3 +1193,127 @@ main(int argc, char *argv[])
...
+	int cmds_front;
+	int cmds_rear;
+	sched_cmd *cmd_enq;
+
+	if (sched_cmds_filtered == NULL) 
+		goto error;
+
+	/* 1. Traverse sched_cmds queue
+	 * 2. While we traverse we keep pushing the unique elements into the sched_cmds_filtered queue
+	 * 3. To find out if a command is duplicate or not here we are making use of the function compare_sched_cmd
```
bhroam, line 1253: Add () to the function

src/scheduler/pbs_sched.c
```diff
@@ -1273,3 +1193,127 @@ main(int argc, char *argv[])
...
+	 * 4. Finally return the sched_cmds_filtered_queue which now only consists of unique sched commands.
+	 */
```
bhroam, line 1255: Move this to an @note or @par in the doxygen header.  It is really the whole function in a nutshell.  It really should be there.

src/scheduler/pbs_sched.c
```diff
@@ -1273,3 +1193,127 @@ main(int argc, char *argv[])
...
+			cmd_enq->cmd = tmp_cmd->cmd;
+			cmd_enq->from_sock = tmp_cmd->from_sock;
+			if (tmp_cmd->jid != NULL) {
+				cmd_enq->jid = string_dup(tmp_cmd->jid);
+				if (cmd_enq->jid == NULL)
+					goto error;
+			}
+
+			if (!ds_enqueue(sched_cmds_filtered, cmd_enq)) 
+				goto error;
```
bhroam, line 1309: I don't think a queue is the right data structure for what you want to do now.  Before it was fine.  Once you start making your queue a hybrid, you should rethink your data structure.  What I think you should do is have 2 data structures.  One is just an array which is the size of the number of commands (17?).  When a new command comes in, you set array[cmd] = 1.  If 20 of the same command come in, you'll just set the same element 20 times.  When you need to run commands, run down the array, and run a cycle for every index which is set to 1.  The second data structure is a list of qruns.  I'd run the qruns first, and then the array of normal cycles.

bhroam (changes requested):

src/lib/Libifl/pbsD_connect.c
```diff
@@ -1011,3 +1013,98 @@ pbs_connect_noblk(char *server, int tout)
...
+
+	conn_list = pthread_getspecific(psi_key);
+
+	for (iter_conns = conn_list; iter_conns != NULL; iter_conns = iter_conns->next) {
+		/* conn_list has secondary set first followed by primary set of connections */
+		if (svr_conns_secondary == NULL)
+			svr_conns_secondary = iter_conns->conn_arr;
+		else if (svr_conns_primary == NULL)
+			svr_conns_primary = iter_conns->conn_arr;
+	}
```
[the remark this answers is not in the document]
suresh-thelkar: Thanks Bhroam I have added this check.

src/scheduler/fifo.c
```diff
@@ -761,35 +761,35 @@ scheduling_cycle(sched_svrconn *sconn, sched_cmd *cmd)
...
+
+	for (i = 0; i < get_num_servers(); i++) {
+		sched_cmd *cmd;
+		int rc;
+
+		if (svr_conns[i].sd < 0)
+			continue;
+		cmd = new_sched_cmd();
+		if (cmd == NULL)
+			continue;
```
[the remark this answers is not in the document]
suresh-thelkar: I have now refactored these code changes. Please take a look at my latest commits.

src/scheduler/fifo.c
```diff
@@ -761,35 +761,35 @@ scheduling_cycle(sched_svrconn *sconn, sched_cmd *cmd)
...
+	int i;
+	svr_conn_t *svr_conns = get_conn_servers(clust_secondary_sock);
+
+	for (i = 0; i < get_num_servers(); i++) {
+		sched_cmd *cmd;
+		int rc;
+
+		if (svr_conns[i].sd < 0)
+			continue;
+		cmd = new_sched_cmd();
```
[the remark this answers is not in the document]
suresh-thelkar: I have refactored the way we create sched commands. Please look into my latest commit.

src/scheduler/pbs_sched.c
suresh-thelkar: This function is not needed now.

src/scheduler/pbs_sched.c
suresh-thelkar: I have removed this function in the latest commit.

src/scheduler/pbs_sched.c
suresh-thelkar: This function is removed in the latest commit.

src/scheduler/pbs_sched.c
suresh-thelkar: I have removed this function in the latest commit.

src/scheduler/pbs_sched.c
suresh-thelkar: Thanks for suggesting these changes @bhroam. Apart from this I had to do some more changes. Summary of the changes is given below.

- Instead of storing the commands in a queue now I am storing them in a hash table implemented using an array like the way you suggested.
- Storing the list of SCHEDULE_AJOBS commands in a separate list 
- Processing all the qrun_list first followed by the rest of the commands
- End of cycle indication is now sent to all the servers to accommodate the above changes.
Please look into my latest commit for the code changes.

suresh-thelkar:
@bhroam / @arungrover, I have implemented all of your review comments. Please take a look and provide your further feedback.

I am also planning to rebase my branch with the master which helps me in running regressions once again and also do few more changes if required as Scheduler is now in C++.

bhroam (changes requested):

src/scheduler/fifo.c
```diff
@@ -763,26 +763,25 @@ get_high_prio_cmd(int *is_conn_lost)
 {
 	int i;
 	svr_conn_t *svr_conns = get_conn_servers(clust_secondary_sock);
+	static sched_cmd *cmd = NULL;
```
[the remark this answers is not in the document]
suresh-thelkar: In the later commit I have changed removed static. Please check.

src/scheduler/pbs_sched.c
```diff
@@ -689,20 +664,24 @@ read_sched_cmd(int sock)
 		 * and if we get priority command then just ignore it
 		 * since we are not yet in middle of schedule cycle
 		 */
-		cmd_prio = new_sched_cmd();
 		if (cmd_prio != NULL) {
-			int rc_prio = get_sched_cmd_noblk(sock, cmd_prio);
+			int rc_prio = get_sched_cmd_noblk(sock, &cmd_prio);
```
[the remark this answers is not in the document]
suresh-thelkar: NULL check for cmd_prio is removed in the later commit 99593d93461d8a48e3c26de890c4698d6d822bec. 

A single server can at the most send one normal/medium priority command followed by a super high priority command. So maximum Scheduler can get two commands from each server at a time. Here since we already read a command at the start of the function (which is normal/medium priority command) the next command that can be if it all if it is received is super high priority command i.e. SCH_SCHEDULE_RESTART_CYCLE. In fact these changes were there earlier I just changed the prototype of the function get_sched_cmd_noblk.

src/scheduler/pbs_sched.c
```diff
@@ -1192,61 +1169,43 @@ main(int argc, char *argv[])
...
- *		from another server then SCH_SCHEDULE_AJOB is replaced with SCH_SCHEDULE_NEW.
- *		Example #2: Suppose if SCH_SCHEDULE_AJOB is present in the queue but we also received SCH_SCHEDULE_AJOB
- *		received from another server we still treat these two commands as differnt because jobids are different.
- *
+ *		Compares two sched commands and tells whether they are duplicate or not.
  *
- * @param[in]	cmd_in_que	-	command in Sched's cache
- * @param[in]	new_cmd		-	command from socket buffer
+ * @param[in]	cmd1	-	command1
+ * @param[in]	cmd2	-      command2
```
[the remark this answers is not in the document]
suresh-thelkar: This function is no longer present in my latest commit.

src/scheduler/pbs_sched.c
```diff
@@ -689,20 +664,24 @@ read_sched_cmd(int sock)
...
-	if (!ds_enqueue(sched_cmds, cmd)) {
-		free_sched_cmd(cmd);
-		return -1;
+
+	if (cmd->cmd != SCH_SCHEDULE_RESTART_CYCLE)  {
+		if (!ds_enqueue(sched_cmds, cmd)) {
+			free_sched_cmd(cmd);
+			return -1;
+		} else
+			*hascmd = 1;
```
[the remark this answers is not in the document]
suresh-thelkar: Refactored this function so as not to use the flag hascmd.

src/scheduler/pbs_sched.c
```diff
@@ -1167,129 +1150,49 @@ main(int argc, char *argv[])
...
-			}
 
-			if (!ds_enqueue(sched_cmds_filtered, cmd_enq)) 
-				goto error;		
-		}
-	}
+	if (schedule(clust_primary_sock, cmd)) /* magic happens here */
+		return 1;
+	else
+		send_cycle_end();
```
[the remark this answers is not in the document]
suresh-thelkar: Yes if we are returning 1 here means we exit so we don't need to send end of cycle indication.

src/scheduler/pbs_sched.c
```diff
@@ -712,6 +709,9 @@ wait_for_cmds()
 				sleep(1); /* wait for 1s for not to burn too much CPU */
 			}
 		} else {
+			for (i = 0; i < TOTAL_SCHED_CMDS;i++) {
+				sched_cmds[i] = 0;
+			}
```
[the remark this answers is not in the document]
suresh-thelkar: Yes. I am now clearing the command as just before we call schedule(). So I have removed this now.

src/scheduler/pbs_sched.c
```diff
@@ -516,19 +515,20 @@ are_we_primary()
...
-		sched_cmd *cmd = ds_dequeue(sched_cmds);
-		free_sched_cmd(cmd);
-	}
-	free_ds_queue(sched_cmds);
-	sched_cmds = NULL;
+
+	/* free qrun_list */
+	for (i = 0; i < get_num_servers(); i++)
+		free(qrun_list[i].jid);
+	free(qrun_list);
```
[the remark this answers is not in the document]
suresh-thelkar: Even though user submits a qrun with list of jobs, server sends SCH_SCHEDULE_AJOB to Scheduler one at a time. Scheduler also reads one command at a time. So if we have 3 servers configured for example, Scheduler can at the most get 3 SCH_SCHEDULE_AJOB commands at the same time. So the number of these commands received at the Scheduler is directly proportional to the number of servers configured. Thinking this I have allocated the size of qrun_list equal to the number of servers configured in the function sched_svr_init().  However while freeing we can actually make use of qrun_list_size and free it which is what I am planning to do.

src/scheduler/globals.h
```diff
@@ -50,7 +50,15 @@ extern "C" {
 #include "sched_cmds.h"
 
 extern void *poll_context;
-extern ds_queue *sched_cmds;
+/* total number of sched commands as of today */
+#define TOTAL_SCHED_CMDS 18
```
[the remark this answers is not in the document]
suresh-thelkar: Thanks Bhroam. I have created an enum for the sched commands as you suggested. I did not make it an enum earlier because there was a gap between SCH_SCHEDULE_CMD(value 5) and SCH_CONFIGURE(value 7). Number 6 was missing. Don't know the reason for this. But since these are internal numbers used to identify commands I don't think it really matters. Anyways now it is enum. Please look into my latest commits for changes.

suresh-thelkar:
@bhroam, I have addressed your review comments. Please look into the following commits.

20330ed Code changes to make it compatible to C++ ...
2cd95eb Implementing review comments round-4 ...

@bhroam / @arungrover, Please look into the above commits and provide your further feedback if any.

src/scheduler/pbs_sched.cpp
```diff
@@ -530,204 +506,117 @@ close_server(sched_svrconn *sconn)
...
-	svrs_newarr = add_ptr_to_array(servers, svr);
-	if (svrs_newarr == NULL) {
-		close_server(svr);
-		die(-1);
+	svr_conns =  static_cast<svr_conn_t *>(get_conn_servers(clust_secondary_sock));
+
+	for (i = 0; i < get_num_servers(); i++) {
+		if (tpp_em_add_fd(poll_context, svr_conns[i].sd, EM_IN | EM_HUP | EM_ERR) < 0) {
+			log_errf(errno, __func__, "Couldn't add secondary connection to poll list for server %s", svr_conns[i].name);
+			pbs_disconnect(clust_primary_sock);
```
arungrover, line 578: Don't think you have to do this. Die() already calls close_servers()

src/scheduler/pbs_sched.cpp
```diff
@@ -530,204 +506,117 @@ close_server(sched_svrconn *sconn)
...
-		log_err(errno, __func__, MEM_ERR_MSG);
-		die(-1);
-	}
-	svr->primary_sock = primary_sock;
-	svr->secondary_sock = secondary_sock;
-	svrs_newarr = add_ptr_to_array(servers, svr);
-	if (svrs_newarr == NULL) {
-		close_server(svr);
-		die(-1);
+	svr_conns =  static_cast<svr_conn_t *>(get_conn_servers(clust_secondary_sock));
```
arungrover, line 573: Same comment, shouldn't we be passing -1?

src/scheduler/pbs_sched.cpp
```diff
@@ -530,204 +506,117 @@ close_server(sched_svrconn *sconn)
...
-	while (primary_sock < 0 || secondary_sock < 0) {
-		if (primary_sock < 0) {
-			primary_sock = connect_server_helper(svrhost, 1);
-			if (primary_sock < 0) {
+	int i;
+	svr_conn_t *svr_conns;
+	int registered = 0;
+
+
+	while (clust_primary_sock < 0 || clust_secondary_sock < 0 || registered == 0) {
```
arungrover, line 547: Earlier the function was connecting to only one server, now the function is meant to connect to a server pool. But this while loop does not really loop on the server pool. 
Since pbs_connect() connects to all servers but returns only one descriptor, there is no way you can tell that it connected to all server instances or not.

src/scheduler/pbs_sched.cpp
```diff
@@ -835,29 +717,34 @@ wait_for_cmds()
...
  * @return void
  */
 static void
-send_cycle_end(sched_svrconn *sconn)
+send_cycle_end()
 {
+	svr_conn_t *svr_conns;
+	int i;
 	static int cycle_end_marker = 0;
+	svr_conns =  static_cast<svr_conn_t *>(get_conn_servers(clust_secondary_sock));
```
arungrover, line 732: shouldn't this be get_conn_servers(-1)?

src/scheduler/pbs_sched.cpp
```diff
@@ -770,20 +654,19 @@ read_sched_cmd(int sock)
...
+		if (rc_prio == -2)
+			return 0;
 	}
-	if (!ds_enqueue(sched_cmds, cmd)) {
-		free_sched_cmd(cmd);
-		return -1;
+
+	if (cmd.cmd != SCH_SCHEDULE_RESTART_CYCLE)  {
+		if (cmd.cmd == SCH_SCHEDULE_AJOB)
+			qrun_list[qrun_list_size++] = cmd;
```
arungrover, line 664: SInce one server can get more than one qrun request, you might have to realloc this list array otherwise you will have invalid memory access.

src/scheduler/pbs_sched.cpp
```diff
@@ -1199,53 +1084,45 @@ main(int argc, char *argv[])
...
+			cmd.cmd = i;
 
-			free_sched_cmd(cmd);
+			/* jid is always NULL since this list does not contain SCHEDULE_AJOB commands */
+			cmd.jid = NULL;
 
-			if (sigprocmask(SIG_SETMASK, &oldsigs, NULL) == -1)
-				log_err(errno, __func__, "sigprocmask(SIG_SETMASK)");
+			/* clear the entry of sched_cmds[i] as we are going to process this command now */
+			sched_cmds[i] = 0;
```
arungrover, line 1118: Since '0' is a valid command, I'd set it to -1 and check for it to be not -1 before calling the wrapper.

src/scheduler/pbs_sched.cpp
```diff
@@ -1199,53 +1084,45 @@ main(int argc, char *argv[])
...
-			now = time(NULL);
-			if (!opt_no_restart)
-				segv_last_time = now;
-#ifdef DEBUG
-			{
-				strftime(log_buffer, sizeof(log_buffer), "%Y-%m-%d %H:%M:%S", localtime(&now));
-				DBPRT(("%s Scheduler received command %d\n", log_buffer, sconn->cmd));
-			}
-#endif
+			if (sched_cmds[i] == 0)
```
arungrover, line 1108: Since '0' is a valid command, I'd set cmd to -1 when it is processed and check for it to be not -1 before calling the wrapper.

arungrover (changes requested):

arungrover:
@suresh-thelkar How do you plan to test this change? Are you going to add some PTL tests with multiple servers?

src/include/sched_cmds.h
```diff
@@ -43,53 +43,42 @@
...
-#define SCH_SCHEDULE_JOBRESV 11		/* Arrival of an existing reservation time */
-#define SCH_SCHEDULE_AJOB  12		/* run one, named job */
-#define SCH_SCHEDULE_STARTQ 13		/* Stopped queue started */
-#define SCH_SCHEDULE_MVLOCAL 14		/* Job moved to local queue */
-#define SCH_SCHEDULE_ETE_ON 15		/* eligible_time_enable is turned ON */
-#define SCH_SCHEDULE_RESV_RECONFIRM 16		/* Reconfirm a reservation */
-#define SCH_SCHEDULE_RESTART_CYCLE 17		/* Restart a scheduling cycle */
+enum svr_sched_cmd
+{
+	SCH_ERROR = -1,
```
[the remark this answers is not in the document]
suresh-thelkar: As this command is not being used scheduler does not receive it. So we don't store it in sched_cmds table so there is no problem. Since there is no usage of this command I have now removed it in server/sched code.

src/scheduler/fifo.cpp
```diff
@@ -751,49 +751,50 @@ scheduling_cycle(sched_svrconn *sconn, sched_cmd *cmd)
...
-			if (cmd->cmd == SCH_SCHEDULE_RESTART_CYCLE) {
-				return cmd;
-			} else {
-				if (!ds_enqueue(sched_cmds, cmd)) {
-					log_eventf(PBSEVENT_SCHED, PBS_EVENTCLASS_SCHED, LOG_WARNING,
-							__func__, "Failed to enqueue sched cmd in cmds queue: cmd=%d jid=%s", cmd->cmd, cmd->jid ? cmd->jid : "(none)");
-					free_sched_cmd(cmd);
-				}
+	int i;
+	svr_conn_t *svr_conns = static_cast<svr_conn_t *>(get_conn_servers(clust_secondary_sock));
```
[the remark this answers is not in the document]
suresh-thelkar: As discussed this is right. It is because clust_secondary_sock acts as a key to fetch all the connections in secondary set.

src/scheduler/pbs_sched.cpp
suresh-thelkar: Same as my earlier comment. This is the correct usage.

src/scheduler/pbs_sched.cpp
suresh-thelkar: As discussed pbs_connect() connects to all servers and returns the first socket which later can be used to fetch all the servers in that set.
Also just added couple of checks to make sure whether pbs_connect indeed connected to all the configured servers or not. If not connected then we retry pbs_connect.

src/scheduler/pbs_sched.cpp
suresh-thelkar: As discussed Server only sends once SCH_SCHEDULE_AJOB command at a time even though somebody submitted a qrun with list of jobs. So no need to realloc. At any time scheduler can get at the max the number of qrun requests which is proportional to the number of configured servers. For some more info please look at the following reply.
https://github.com/openpbs/openpbs/pull/2068#discussion_r508242755

src/scheduler/pbs_sched.cpp
suresh-thelkar: The command is actually stored as an index, the value at this index just indicates whether a particular command is received or not. So even though if zero is valid command we still honor it so it might look like sched_cmds[SCH_SCHEDULE_NULL] = 1 (SCH_SCHEDULE_NULL = 0) if we receive SCH_SCHEDULE_NULL or
sched_cmds[SCH_SCHEDULE_NULL] = 0 if we don't receive it

I feel 0 or 1 is more readable here and it is not causing any problem in my testing. So I want to leave this as it. Hope this is fine.

We can actually initialize all the elements of the array to -1 in the beginning once also, But this needs an extra loop. As of now sched_cmds[] is global so by default it is initialized to zero which is what is needed for us.

src/scheduler/pbs_sched.cpp
suresh-thelkar: Same as my earlier reply.

src/scheduler/pbs_sched.cpp
suresh-thelkar: Please look into the above reply.

suresh-thelkar:
@arungrover, I have addressed all your comments. Please take a look at the following commits.

0a02bf9 Making couple of changes as get_conn_servers() is no longer available in master ...
3aac7cf Implementing review comments round-5 ...

I had to do 0a02bf9 to make Travis run smoothly.

@arungrover / @bhroam, can you please review the latest commits and provide your further feedback.

src/scheduler/pbs_sched.cpp
```diff
@@ -540,26 +540,54 @@ static void
...
 			clust_primary_sock = pbs_connect(NULL);
+			/* clust_primary_sock later is used to fetch primary connections for all the servers */
 			if (clust_primary_sock < 0) {
 				/* wait for 2s for not to burn too much CPU, and then retry connection */
 				sleep(2);
 				continue;
 			}
 		}
 		clust_secondary_sock = pbs_connect(NULL);
+		/* clust_secondary_sock later is used to fetch secondary connections for all the servers */
```
bhroam, line 560: Merge these two comments into 1 and put it at the top of the if() block.  Say 'pbs_connect() will return a connection handle for all servers'

src/scheduler/pbs_sched.cpp
```diff
@@ -540,26 +540,54 @@ static void
...
+
+		for (i = 0; i < num_conf_svrs; i++) {
+			if (svr_conns_primary[i].state == SVR_CONN_STATE_DOWN) {
+				clust_primary_sock = -1;
+				break;
+			}
+			if (svr_conns_primary[i].state == SVR_CONN_STATE_DOWN) {
+				clust_secondary_sock = -1;
+				break;
+			}
```
bhroam, line 578: The problem with putting the break where you have it is that you won't catch the case where both are down.  This is likely if a server crashed.  I'd change this to check if (prim == DOWN || sec == DOWN) then check individually for prim and sec setting the sockets to -1.  At the end of of the outer if statement, break.  Although in the situation where we do have one server which crashed, the other servers pick up for it.  Won't this cause us to wait for it to come back up before we continue?

src/include/sched_cmds.h
```diff
@@ -55,24 +55,28 @@ struct sched_cmd {
...
+	SCH_SCHEDULE_FIRST,	/* First schedule after server starts */
+	SCH_SCHEDULE_JOBRESV,	/* Arrival of an existing reservation time */
+	SCH_SCHEDULE_AJOB,	/* run one, named job */
+	SCH_SCHEDULE_STARTQ,	/* Stopped queue started */
+	SCH_SCHEDULE_MVLOCAL,	/* Job moved to local queue */
+	SCH_SCHEDULE_ETE_ON,	/* eligible_time_enable is turned ON */
+	SCH_SCHEDULE_RESV_RECONFIRM,	/* Reconfirm a reservation */
+	SCH_SCHEDULE_RESTART_CYCLE, 	/* Restart a scheduling cycle */
+	SCH_CMD_HIGH	/* This has to be the last command always. Any new command can be inserted above if required */
+};
```
bhroam, line 78: Thanks for turning this into an enum.  I just had a thought about the new direct hashing mechanism used with this enum.  There is nothing saying a bug won't send a number higher than SCH_CMD_HIGH.  Maybe do validity checking before you hash into the array that the input is valid.

bhroam (changes requested):

src/include/sched_cmds.h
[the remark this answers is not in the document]
bhroam: Sure, remove it.  We could probably remove a few more if we wanted.  Originally (like >20 yrs ago), the idea was that scheduling was site dependent, so each site would write their own scheduler.  We provided pbs_sched.c which would link into the sites scheduler.  They would provide 'schedinit()' and schedule()' which we would call.  The sched commands which were defined were a guess at what would be useful.  When I wrote the scheduler, I chose the commands I thought were useful, and turned the rest  into a noop.

src/scheduler/fifo.cpp
```diff
@@ -762,8 +761,9 @@ static int
 get_high_prio_cmd(int *is_conn_lost, sched_cmd *high_prior_cmd)
 {
 	int i;
-	svr_conn_t *svr_conns = static_cast<svr_conn_t *>(get_conn_servers(clust_secondary_sock));
 	sched_cmd cmd;
+	int num_conf_servers = get_num_servers();
+	svr_conn_t *svr_conns = static_cast<svr_conn_t *>(get_conn_servers(clust_secondary_sock));
 
 	for (i = 0; i < get_num_servers(); i++) {
```
arungrover, line 768: please use num_conf_servers here as well.

arungrover (dismissed):

src/scheduler/pbs_sched.cpp
suresh-thelkar: Thanks @bhroam. I have implemented as you suggested. Our goal is to wait till all servers come up.

src/include/sched_cmds.h
suresh-thelkar: Have added the corresponding validation checks while storing commands into the hash table.

src/scheduler/fifo.cpp
suresh-thelkar: done

suresh-thelkar:
@bhroam / @arungrover, I have implemented your comments. Please check the latest commit and provide your feedback.

suresh-thelkar:
All the regression test cases are fine. I am planning to run full regressions once again after all the reviews. Also I am almost done with multi-server testing. I will attach the logs shortly.

suresh-thelkar:
All the regression test cases are fine. I am planning to run full regressions once again after all the reviews. Also I am almost done with multi-server testing. I will attach the logs shortly.

src/scheduler/pbs_sched.cpp
```diff
@@ -530,204 +506,139 @@ close_server(sched_svrconn *sconn)
...
+
+
+	for (i = 0; i < get_num_servers(); i++) {
+		if (tpp_em_add_fd(poll_context, svr_conns_secondary[i].sd, EM_IN | EM_HUP | EM_ERR) < 0) {
+			log_errf(errno, __func__, "Couldn't add secondary connection to poll list for server %s", svr_conns_secondary[i].name);
+			die(-1);
+		}
 	}
-	servers = (sched_svrconn **)svrs_newarr;
-	log_eventf(PBSEVENT_ADMIN | PBSEVENT_FORCE, PBS_EVENTCLASS_SCHED, LOG_INFO, msg_daemonname, "Connected to server %s", svrhost);
```
agrawalravi90, line 666: I think this log message can be useful, why did you decide to remove it?

src/scheduler/pbs_sched.cpp
```diff
@@ -530,204 +506,139 @@ close_server(sched_svrconn *sconn)
...
-		die(-1);
-	}
 
-	svr = static_cast<sched_svrconn *>(malloc(sizeof(sched_svrconn)));
-	if (!svr) {
-		log_err(errno, __func__, MEM_ERR_MSG);
-		die(-1);
-	}
+		svr_conns_primary =  static_cast<svr_conn_t *>(get_conn_svr_instances(clust_primary_sock));
+		svr_conns_secondary =  static_cast<svr_conn_t *>(get_conn_svr_instances(clust_secondary_sock));
```
agrawalravi90, line 567: Will declaring svr_conns_primary and svr_conns_secondary as 'auto' variables help remove the static_casts here?

src/scheduler/pbs_sched.cpp
```diff
@@ -530,204 +506,139 @@ close_server(sched_svrconn *sconn)
...
 			die(-1);
 		}
+	}
 
-		sched_cmds = new_ds_queue();
-		if (sched_cmds == NULL) {
-			log_errf(errno, __func__, MEM_ERR_MSG);
-			die(-1);
-		}
+	qrun_list = static_cast<sched_cmd *>(malloc((get_num_servers() + 1) * sizeof(sched_cmd)));
```
agrawalravi90, line 623: malloc returns a void pointer, does it still need to be casted?

src/scheduler/pbs_sched.cpp
```diff
@@ -835,29 +741,34 @@ wait_for_cmds()
...
  * @return void
  */
 static void
-send_cycle_end(sched_svrconn *sconn)
+send_cycle_end()
 {
+	svr_conn_t *svr_conns;
+	int i;
 	static int cycle_end_marker = 0;
+	svr_conns =  static_cast<svr_conn_t *>(get_conn_svr_instances(clust_secondary_sock));
```
agrawalravi90, line 756: Can we find a way to avoid the explicit casts? they are so ugly

src/scheduler/pbs_sched.cpp
```diff
@@ -835,29 +741,34 @@ wait_for_cmds()
...
 
-	if (diswsi(sconn->secondary_sock, cycle_end_marker) != DIS_SUCCESS) {
-		log_eventf(PBSEVENT_SYSTEM | PBSEVENT_FORCE, PBS_EVENTCLASS_SCHED, LOG_ERR, __func__,
-			   "Not able to send end of cycle, errno = %d", errno);
-		goto err;
-	}
+	for (i = 0; i < get_num_servers(); i++) {
+		if (diswsi(svr_conns[i].sd, cycle_end_marker) != DIS_SUCCESS) {
```
agrawalravi90, line 759: just checking, there might be servers which never started a cycle, is it ok to send them the end cycle message?

agrawalravi90 (changes requested):

src/lib/Libifl/pbsD_connect.c
```diff
@@ -997,3 +999,99 @@ pbs_connect_noblk(char *server, int tout)
...
+	for (i = 0; i < get_num_servers(); i++) {
+		if (!svr_conns_primary[i].registered) {
+			if (send_register_sched(svr_conns_primary[i].sd, client_id) == 0)
+				return 0;	
+			svr_conns_primary[i].registered = 1;
+		}
+		if (!svr_conns_secondary[i].registered) {
+			if (send_register_sched(svr_conns_secondary[i].sd, client_id) == 0)
+				return 0;
+			svr_conns_secondary[i].registered = 1;
```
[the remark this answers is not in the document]
agrawalravi90: also this seems like a sched specific information, so we should avoid putting it in svr_conn_t which is used by all clients

src/scheduler/pbs_sched.cpp
```diff
@@ -530,204 +506,139 @@ close_server(sched_svrconn *sconn)
...
+		svr_conns_secondary =  static_cast<svr_conn_t *>(get_conn_svr_instances(clust_secondary_sock));
 
-	svr->svrhost = strdup(svrhost);
-	if (!svr->svrhost) {
-		free(svr);
-		log_err(errno, __func__, MEM_ERR_MSG);
-		die(-1);
+		for (i = 0; i < num_conf_svrs; i++) {
+			if (svr_conns_primary[i].state == SVR_CONN_STATE_DOWN || svr_conns_primary[i].state == SVR_CONN_STATE_DOWN) {	
+				if (svr_conns_primary[i].state == SVR_CONN_STATE_DOWN)
```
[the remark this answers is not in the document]
bhroam: That is what he had.  I suggested this because both primary and secondary might be down.  He needs to set both clust_primary_sock and clust_seconday_sock to -1 before he breaks.

bhroam:
I have no more comments other than what @agrawalravi90 has mentioned.

src/lib/Libifl/pbsD_connect.c
```diff
@@ -997,3 +999,99 @@ pbs_connect_noblk(char *server, int tout)
...
+	int i;
+	svr_conns_list_t *conn_list;
+	svr_conn_t *svr_conns_primary = NULL;
+	svr_conn_t *svr_conns_secondary = NULL;
+	svr_conns_list_t *iter_conns = NULL;
+
+	if (client_id == NULL)
+		return 0;
+
+	conn_list = pthread_getspecific(psi_key);
```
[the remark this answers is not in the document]
suresh-thelkar: clust_primary_sock and clust_secondary_sock are not available in this function and hence I had to use that. Now that  we have refactored pbs_register_sched to take both primary and secondary client ids as parameters, we are now making use of get_conn_svr_instances here as well.

src/lib/Libifl/pbsD_connect.c
```diff
@@ -997,3 +999,99 @@ pbs_connect_noblk(char *server, int tout)
...
+/**
+ * @brief Registers the Scheduler with all the Servers configured
+ *
+ * param[in]	client_id -  client identifier which is known to server
+ * @return int
+ * @retval 0  - failure
+ * @return 1  - success
+ */
+int
+pbs_register_sched(const char *client_id)
```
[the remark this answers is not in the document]
suresh-thelkar: As discussed this is like a true IFL function which sends a batch request and gets a batch response. As per the discussion with @subhasisb I have moved it to Libifl. We can consider this as an internal IFL function.

src/lib/Libifl/pbsD_connect.c
```diff
@@ -997,3 +999,99 @@ pbs_connect_noblk(char *server, int tout)
...
+	svr_conn_t *svr_conns_primary = NULL;
+	svr_conn_t *svr_conns_secondary = NULL;
+	svr_conns_list_t *iter_conns = NULL;
+
+	if (client_id == NULL)
+		return 0;
+
+	conn_list = pthread_getspecific(psi_key);
+
+	for (iter_conns = conn_list; (iter_conns != NULL) && (svr_conns_primary == NULL || svr_conns_secondary == NULL);
```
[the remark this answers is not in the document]
suresh-thelkar: Now that pbs_register_sched() is refactored as I mentioned above this is possible now and made these changes.

src/lib/Libifl/pbsD_connect.c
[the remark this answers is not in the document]
suresh-thelkar: Removed registered flag in svr_conn_t. This is ok since we anyways have to reattempt pbs_connect's again even if one of the servers is down. This is how pbs_connect works.

src/scheduler/pbs_sched.cpp
suresh-thelkar: Since pbs_connect now connects to all the configured servers I removed it earlier. Added a similar log message which says all the configured servers are connected.

src/scheduler/pbs_sched.cpp
suresh-thelkar: All the scheduler code is using static_cast in many places. I followed the same approach. Don't know the exact reason. May be we prefer explicit and well defined casting. It is better if we follow one style so that is consistent across all scheduler code and hence leaving this as is.

src/scheduler/pbs_sched.cpp
suresh-thelkar: Same as the reason https://github.com/openpbs/openpbs/pull/2068#discussion_r510746731

src/scheduler/pbs_sched.cpp
suresh-thelkar: Same as the reason https://github.com/openpbs/openpbs/pull/2068#discussion_r510746731

src/scheduler/pbs_sched.cpp
suresh-thelkar: Yes it is ok even if one of the servers does not start a cycle but getting cycle end notification. That is the best we can do. This has been discussed earlier with Bhroam.

suresh-thelkar:
> All the regression test cases are fine. I am planning to run full regressions once again after all the reviews. Also I am almost done with multi-server testing. I will attach the logs shortly.

Attached are test logs.
[PR2068_logs.txt](https://github.com/openpbs/openpbs/files/5428915/PR2068_logs.txt)

suresh-thelkar:
@agrawalravi90, I have implemented your review comments. Please check the latest commit.
@bhroam, Thank you very much for your sign off. I have done few more changes after that. Can you please look into the latest commit.
@arungrover, Can you please go through the latest commit.

src/scheduler/pbs_sched.cpp
```diff
@@ -582,20 +586,23 @@ connect_svrpool()
...
 			continue;
 		}
 
-		if (pbs_register_sched(sc_name) == 0) {
+		if (pbs_register_sched(sc_name, clust_primary_sock, clust_secondary_sock) == 0) {
 			log_errf(pbs_errno, __func__, "Couldn't register the scheduler %s with the configured servers", sc_name);
 			/* wait for 2s for not to burn too much CPU, and then retry connection */
-			sleep(2);			
+			sleep(2);
+			close_servers();
```
agrawalravi90, line 597: you will need to set clust_primary_sock and clust_secondary_sock to -1 every time you call close_servers() so that you can connect with them again. I think you should just have a while(1) loop instead of while (clust_primary_sock < 0 || clust_secondary_sock < 0 || registered == 0). Then, you don't need clust_primary_sock, clust_secondary_sock or registered, and you can just break out of the loop at the end when you have connected and registered both primary and secondary.

agrawalravi90 (changes requested):

arungrover (commented):
I have no more comments on the change @suresh-thelkar

src/scheduler/pbs_sched.cpp
suresh-thelkar: > you will need to set clust_primary_sock and clust_secondary_sock to -1 every time you call close_servers() so that you can connect with them again

This is not required since I am setting set clust_primary_sock and clust_secondary_sock in close_servers() itself.

suresh-thelkar: > I think you should just have a while(1) loop instead of while (clust_primary_sock < 0 || clust_secondary_sock < 0 || registered == 0). Then, you don't need clust_primary_sock, clust_secondary_sock or registered, and you can just break out of the loop at the end when you have connected and registered both primary and secondary.

I have kept all the conditions in while loop just for better readability. Now I have changed it to while(1) loop and breaking it if everything is succeeded. Please check my latest commit.

suresh-thelkar:
> I have no more comments on the change @suresh-thelkar

Thank you @arungrover for the reviews and sign off.

suresh-thelkar:
Thanks @agrawalravi90, I have addressed your latest comments as well. Please take a look.

@arungrover, Thanks for all the reviews and sign off.

@bhroam, Thank you very much for all the reviews and sign off. I have made a simple change after that. Please take a look at the latest commit and kindly take the PR forward.

agrawalravi90 (commented):
LGTM, can you please post latest regression results?

suresh-thelkar:
Following are the regression results

Profile: regression
|--|--|--|--|--|--|--|
|4535|3848|3|1|0|0|3844|

The test cases that are failed are known issues.

suresh-thelkar:
> LGTM, can you please post latest regression results?

Thank you for the sign off. I have added the results. Please check.

agrawalravi90 (approved):

agrawalravi90:
@suresh-thelkar  thanks, @bhroam still needs to sign-off on this

suresh-thelkar:
> @suresh-thelkar thanks, @bhroam still needs to sign-off on this


> @suresh-thelkar thanks, @bhroam still needs to sign-off on this

suresh-thelkar:
> @suresh-thelkar thanks, @bhroam still needs to sign-off on this

@bhroam, Thanks for your sign off. Can you please look at the latest commit and approve the PR if it still looks good.

bhroam (approved):
Thanks for making all the changes, @suresh-thelkar

PR #2071, opened 2020-10-09
fixed standing resv first occurrence issue

arungrover:
#### Describe Bug or Feature
If a user tries to submit a standing reservation where the start time is in past and the end time is future, PBS server rejects such a job. 
This problem only happens when the start and end time is given in hhmm format.

#### Describe Your Change
The root of the problem is that when pbs_rsub command submits this request, it detects that the start time is in the past and adds a day to it. But since the command finds end time in the future, it does not move the end time.
PBS server then rejects this submission because for the server, the start time is set in the future and the end time is set before the start time.
Change is to detect in the command itself that start time is in the past. If we detect that then we add 24 hrs to both the start and end time of the reservation.

#### Link to Design Doc
NA

[test_standing_before.txt](https://github.com/openpbs/openpbs/files/5356173/test_standing_before.txt)
[test_standing_after.txt](https://github.com/openpbs/openpbs/files/5356174/test_standing_after.txt)

### DIFF
--- a/test/tests/functional/pbs_reservations.py
+++ b/test/tests/functional/pbs_reservations.py
@@ -2332,7 +2332,7 @@ class TestReservations(TestFunctional):
         runcmd = [os.path.join(self.server.pbs_conf['PBS_EXEC'], 'bin',
                                'pbs_rsub')]
         tz = ['PBS_TZID=' + self.get_tz()]
-        rule = ["-r \"freq=WEEKLY;BYDAY=SU;COUNT=3\""]
+        rule = ["-r 'freq=WEEKLY;BYDAY=SU;COUNT=3'"]
 
         runcmd = tz + runcmd + start + end + rule
         ret = self.du.run_cmd(self.server.hostname, runcmd, as_script=True)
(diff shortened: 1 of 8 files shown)


test/tests/functional/pbs_reservations.py
```diff
@@ -2311,3 +2314,43 @@ def test_resv_reconfirm_holding_partial_nodes(self):
...
+
+        # we cannot use self.server.submit to submit the reservation
+        # because we don't want to specify date in start and end options
+        start = [" -R " + time.strftime('%H%M', time.localtime(now - 3600))]
+        end = [" -E " + time.strftime('%H%M', time.localtime(now + 3600))]
+
+        runcmd = [os.path.join(self.server.pbs_conf['PBS_EXEC'], 'bin',
+                               'pbs_rsub')]
+        tz = ['PBS_TZID=' + self.get_tz()]
+        rule = ["-r \"freq=WEEKLY;BYDAY=SU;COUNT=3\""]
```
vstumpf, line 2335: you can use single quotes for the python string so you don't have to escape the double quotes

test/tests/functional/pbs_reservations.py
```diff
@@ -2311,3 +2314,43 @@ def test_resv_reconfirm_holding_partial_nodes(self):
...
+
+        a = {'reserve_state': (MATCH_RE, 'RESV_CONFIRMED|2')}
+        self.server.expect(RESV, a, id=rid)
+        d = datetime.datetime.today()
+
+        # weekday() returns the weekday's index. Monday to Sunday (0 to 6)
+        # We calculate how far away is today than Sunday and move
+        # the day ahead.
+        n = d.weekday()
+        d += datetime.timedelta(6-n)
```
vstumpf, line 2350: can you specify `day` here? to make it more clear

bhroam (approved):
LGTM

bhroam (approved):

vstumpf (approved):
LGTM

PR #2077, opened 2020-10-18
License refactoring modification

Jitendra-Rajak:
#### Describe Bug or Feature
Changes and modification in the recent license refactoring .

#### Describe Your Change
Changes made to the recent license refactoring for a better management of nodes and vnodes.

Logs attached to the ticket.

### DIFF
--- a/src/include/pbs_license.h
+++ b/src/include/pbs_license.h
@@ -86,10 +86,6 @@ enum node_topology_type {
 };
 typedef enum node_topology_type ntt_t;
 
-typedef struct {
-	struct pbsnode *pnode;
-	pbs_list_link link;
-} unlicensed_node;
 
 pbs_list_head unlicensed_nodes_list;
 
--- a/src/include/pbs_license.h
+++ b/src/include/pbs_license.h
@@ -86,6 +86,13 @@ enum node_topology_type {
 };
 typedef enum node_topology_type ntt_t;
 
+typedef struct {
+	struct pbsnode *pnode;
+	pbs_list_link link;
+} unlicensed_node;
+
+pbs_list_head unlicensed_nodes_list;
+
 #define PBS_MIN_LICENSING_LICENSES	0
 #define PBS_MAX_LICENSING_LICENSES	INT_MAX
 #define PBS_LIC_LINGER_TIME		31536000 /* keep extra licenses 1 year by default */
@@ -99,6 +106,7 @@ extern int release_node_lic(void *);
 extern void license_nodes();
 extern void init_licensing(struct work_task *ptask);
 extern void reset_license_counters(pbs_license_counts *);
+extern void remove_from_unlicensed_node_list(struct pbsnode *pnode);
 
 /* Licensing-related variables */
 extern char *pbs_licensing_location;
--- a/src/include/pbs_license.h
+++ b/src/include/pbs_license.h
@@ -86,13 +86,6 @@ enum node_topology_type {
 };
 typedef enum node_topology_type ntt_t;
 
-typedef struct {
-	struct pbsnode *pnode;
-	pbs_list_link link;
-} unlicensed_node;
-
-pbs_list_head unlicensed_nodes_list;
-
 #define PBS_MIN_LICENSING_LICENSES	0
 #define PBS_MAX_LICENSING_LICENSES	INT_MAX
 #define PBS_LIC_LINGER_TIME		31536000 /* keep extra licenses 1 year by default */
@@ -106,7 +99,6 @@ extern int release_node_lic(void *);
 extern void license_nodes();
 extern void init_licensing(struct work_task *ptask);
 extern void reset_license_counters(pbs_license_counts *);
-extern void remove_from_unlicensed_node_list(struct pbsnode *pnode);
 
 /* Licensing-related variables */
 extern char *pbs_licensing_location;
--- a/src/server/licensing_func.c
+++ b/src/server/licensing_func.c
@@ -102,64 +102,6 @@ return_licenses(long count)
 
 /**
  * @brief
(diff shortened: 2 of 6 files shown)


src/include/pbs_license.h
```diff
@@ -106,7 +99,6 @@ extern int release_node_lic(void *);
 extern void license_nodes();
 extern void init_licensing(struct work_task *ptask);
 extern void reset_license_counters(pbs_license_counts *);
-extern void remove_from_unlicensed_node_list(struct pbsnode *pnode);
```
prakashcv13, line 109: why are you removing this code?

Jitendra-Rajak: reverted the change as discussed

Jitendra-Rajak:
@prakashcv13 and @nithinj , I have made the change as per offline discussion please have a look.

src/server/licensing_func.c
```diff
@@ -286,8 +286,13 @@ propagate_licenses_to_vnodes(mominfo_t *pmom)
 	if (node_index_start)
 		distribute_licenseinfo(pmom, lic_count);
 
-	if (pfrom_Lic == NULL)
+	if (pfrom_Lic == NULL) {
```
nithinj, line 289: why need a different loop for this? Couldn't you accommodate this with the previous loop without a break.

nithinj (changes requested):

src/server/licensing_func.c
Jitendra-Rajak: done!

src/server/licensing_func.c
```diff
@@ -624,6 +624,9 @@ license_nodes()
 				for (i = 0; i < np->nd_nummoms; i++)
 					propagate_licenses_to_vnodes(np->nd_moms[i]);
 			}
+		} else {
```
nithinj, line 627: Could you make use of remove_from_unlicensed_node_list() for this? Why do you need a unlicensed_node struct? Can't you add the link to original node strcture and get rid of unlicensed_node struct. So that you dont have to loop inside remove_from_unlicensed_node_list to find the node.

src/server/node_manager.c
```diff
@@ -4947,6 +4947,7 @@ is_request(int stream, int version)
 			}
 			if (made_new_vnodes || cr_node) {
 				save_nodes_db(1, pmom); /* update the node database */
+				propagate_licenses_to_vnodes(pmom);
```
nithinj, line 4950: This must be a duplicate call. The same is invoked in line 4525

nithinj (changes requested):

src/server/node_manager.c
Jitendra-Rajak: That call will not be invoked incase of update from hook so I have invoked the call here

nithinj: Okay. Is it possible to remove the first invocation then?

Jitendra-Rajak: No ! ,  that will be used in case of updating vnodes which are not created via hook

nithinj: I agree. But we are not returning from function anywhere in between. So it should hit your newly introduced call as well.

Jitendra-Rajak: Yes but we are inside different cases and condition for both the calls

src/server/licensing_func.c
Jitendra-Rajak: As its a list of nodes to find that node in the list we have to traverse we cant be sure where and at what position desired node would be, so instead of using remove_from_unlicensed_node_list() I am directly removing the node from that list as we have already traversed above.

src/server/node_manager.c
nithinj: got it. Can you move the first call alongside with save_nodes_db within the same made_new_vnodes and cr_node check then?

src/server/licensing_func.c
nithinj: I was proposing a change to remove_from_unlicensed_node_list so it does not have to traverse to remove the node from list
```
void
remove_from_unlicensed_node_list(struct pbsnode *pnode)
{
	if(!pnode->nd_added_to_unlicensed_list)
		return;

	pnode->nd_added_to_unlicensed_list = 0;
	licensing_control.licenses_total_needed -= pnode->nd_attr[ND_ATR_LicenseInfo].at_val.at_long;
	delete_link(&pnode->un_lic_link);
}
```

src/server/node_manager.c
Jitendra-Rajak: yes made the change!

src/server/licensing_func.c
Jitendra-Rajak: I am not able to understand this change I think what we have now is good to go !

Jitendra-Rajak:
I have attached the latest logs to the ticket

src/server/licensing_func.c
prakashcv13: @nithinj - are you proposing to remove the unlicensed_node_list itself.

nithinj: No, I was suggesting to use a link in pnode struct itself rather than in a separate struct like unlicensed_node

Jitendra-Rajak: I have made the changes @nithinj , please have a look on the latest commit

nithinj (approved):
LGTM!

Jitendra-Rajak:
@nithinj  I have rebased my branch as some ptl framework changes has been merged recently so to run regression I need those changes.

Jitendra-Rajak:
Profile: regression
|--|--|--|--|--|--|--|
|4553|3848|5|1|0|0|3842|

Regression report attched

prakashcv13 (approved):

PR #2082, opened 2020-10-21
Stop and remove pbs_jobs.service when uninstalling pbs

minghui-liu:
#### Describe Bug or Feature

`pbs_cgroups` hook creates the `pbs_job.service` to handle cpuset jobs under PBS. The service remains active on the system even after PBS is uninstalled, and might prevent other programs from being able to use cpuset. One known case is when downgrading to a previous version of PBS that has cpuset mom: cpuset mom fails to start because it cannot use any cpu. 

#### Describe Your Change

In the RPM pre-uninstall script, stop pbs_jobs.service and remove its cgroup directory. 

#### Link to Design Doc

N/A

Manual test results
[before fix.txt](https://github.com/openpbs/openpbs/files/5441974/before.fix.txt)
[after fix.txt](https://github.com/openpbs/openpbs/files/5441973/after.fix.txt)
[after fix with cgroup prefix renamed.txt](https://github.com/openpbs/openpbs/files/5438330/after.fix.with.cgroup.prefix.renamed.txt)


src/cmds/scripts/pbs_preuninstall
```diff
@@ -61,6 +61,12 @@ server|execution)
         echo "have systemd"
         systemctl disable pbs
         rm -f /usr/lib/systemd/system-preset/95-pbs.preset
+        systemctl stop pbs_jobs.service
```
bayucan, line 64: If pbs_jobs.service does not exist, would this call error out?

vstumpf:
What if someone changed the cgroup-prefix in the config file?
Should there be some documentation for this?

minghui-liu:
> What if someone changed the cgroup-prefix in the config file?
Good point. I'll need to parse the cgroup config file for the cgroup prefix. I'll update this PR with that feature. 
> Should there be some documentation for this?
Do you mean a design document?

src/cmds/scripts/pbs_preuninstall
minghui-liu: It will throw an error message but the preuninstall script will continue to run to the end. I think your concern is the same as what Vincent raise below: the service name can change with cgroup-prefix. I am working on this.

vstumpf:
Ah, I meant admin guide. 
In theory, the admin could have removed the cgroup hook, then decided to uninstall, causing this lookup to fail. So something to the effect of "You may have to manually remove the pbs cgroups service when uninstalling pbs"

src/cmds/scripts/pbs_preuninstall
```diff
@@ -61,9 +62,10 @@ server|execution)
         echo "have systemd"
         systemctl disable pbs
         rm -f /usr/lib/systemd/system-preset/95-pbs.preset
-        systemctl stop pbs_jobs.service
-        [ -d /sys/fs/cgroup/cpuset/pbs_jobs.service/jobid ] && rmdir /sys/fs/cgroup/cpuset/pbs_jobs.service/jobid
-        [ -d /sys/fs/cgroup/cpuset/pbs_jobs.service ] && rmdir /sys/fs/cgroup/cpuset/pbs_jobs.service
+        service_name=`cat ${pbs_home}/server_priv/hooks/pbs_cgroups.CF | python -c "import sys, json; print(json.load(sys.stdin)['cgroup_prefix'])"`
```
bayucan, line 65: Need to guard against error doing cat or python. Maybe check first for existence of ${pbs_home}/server_priv/hoks/pbs_cgroups.CF, and also if 'python' is in path.

bayucan (changes requested):

bayucan (approved):
Looks good. By the way, I think the "before_fix.txt" and "after_fix.txt" names are switched. I see no errors in the former, and errors in the latter.

minghui-liu:
> Looks good. By the way, I think the "before_fix.txt" and "after_fix.txt" names are switched. I see no errors in the former, and errors in the latter.

Thanks. I switched them back.

src/cmds/scripts/pbs_preuninstall
```diff
@@ -61,6 +62,15 @@ server|execution)
...
+            [ -d /sys/fs/cgroup/cpuset/${service_name}.service/jobid ] && rmdir /sys/fs/cgroup/cpuset/${service_name}.service/jobid
+            [ -d /sys/fs/cgroup/cpuset/${service_name}.service ] && rmdir /sys/fs/cgroup/cpuset/${service_name}.service
+        fi
+        [ -d /sys/fs/cgroup/cpuset/pbspro.service/jobid ] && rmdir /sys/fs/cgroup/cpuset/pbspro.service/jobid
+        [ -d /sys/fs/cgroup/cpuset/pbspro.service ] && rmdir /sys/fs/cgroup/cpuset/pbspro.service
+        exit 0
     fi
     ;;
-esac
+esac
```
vstumpf, line 76: Can you readd the newline here?

vstumpf (commented):
Just one nitpick

src/cmds/scripts/pbs_preuninstall
minghui-liu: Done

bayucan (commented):
still good.

src/cmds/scripts/pbs_preuninstall
```diff
@@ -61,6 +62,15 @@ server|execution)
         echo "have systemd"
         systemctl disable pbs
         rm -f /usr/lib/systemd/system-preset/95-pbs.preset
+        if command -v python &> /dev/null && [ -f ${pbs_home}/server_priv/hooks/pbs_cgroups.CF ]; then
```
vstumpf, line 65: Shouldn't this be mom_priv?

src/cmds/scripts/pbs_preuninstall
```diff
@@ -61,6 +62,15 @@ server|execution)
...
+            service_name=`cat ${pbs_home}/server_priv/hooks/pbs_cgroups.CF | python -c "import sys, json; print(json.load(sys.stdin)['cgroup_prefix'])"`
```
vstumpf, line 66: Same here

vstumpf (changes requested):
I'm sorry, i totally missed this.

src/cmds/scripts/pbs_preuninstall
bayucan: Good point, Vincent. Since we're concerned with removing the service handling cpuset jobs, then it should be on the mom side.

vstumpf (approved):
LGTM

bayucan (approved):
Looks good.

PR #2084, opened 2020-10-22
test_resv_reconfirm_holding_partial_nodes test fix

visheshh:
#### Describe Bug or Feature
test_resv_reconfirm_holding_partial_nodes fails after vnode name was removed from create_vnodes 

#### Describe Your Change
Correct the vnode name list.

Profile: regression
|--|--|--|--|--|--|--|
|4419|49|0|0|0|1|48|

### DIFF
--- a/test/tests/functional/pbs_reservations.py
+++ b/test/tests/functional/pbs_reservations.py
@@ -2268,7 +2268,7 @@ class TestReservations(TestFunctional):
         self.server.manager(MGR_CMD_SET, SERVER, a)
         a = {'resources_available.ncpus': 2}
         self.mom.create_vnodes(a, num=3)
-        vn_list = ['vn[0]', 'vn[1]', 'vn[2]']
+        vn_list = ["%s[%d]" % (self.mom.shortname, i) for i in range(3)]
 
         now = int(time.time())
         sel = '1:ncpus=2+1:ncpus=1'


kjakkali (approved):

test/tests/functional/pbs_reservations.py
```diff
@@ -2268,7 +2268,7 @@ def test_resv_reconfirm_holding_partial_nodes(self):
         self.server.manager(MGR_CMD_SET, SERVER, a)
         a = {'resources_available.ncpus': 2}
         self.mom.create_vnodes(a, num=3)
-        vn_list = ['vn[0]', 'vn[1]', 'vn[2]']
+        vn_list = ["%s[%d]" % (self.mom.shortname, i) for i in range(3)]
```
hirenvadalia, line 2271: May for future we should change signature of create_vnodes to return either vnode names or list of newly create vnode objects (aka MoM objects or introduce new Vnode object in PTL)

hirenvadalia (approved):

test/tests/functional/pbs_reservations.py
bhroam: I'll +1 returning a list of vnode names.  I think they'll be more useful than creating a vnode object, but that works too.

PR #2090, opened 2020-10-27
Remove max_attempts except from the existence=False cases

lisa-altair:
#### Describe Bug or Feature

PTL tests fail because the test machine may be slow and max_attempts is set to a small value within the test.

#### Describe Your Change

I removed the specific max_attempts settings within the tests so the test would continue to look for the expected values/settings.
I left max_attemps for the existence=False cases.  For example, that's where we check to make sure we aren't getting some error statement.

[rampdown.txt](https://github.com/openpbs/openpbs/files/5442168/rampdown.txt)


vccardenas (approved):
Looks good to me.

bayucan (approved):
Looks good.

bayucan:
Merged #2090 to master.

PR #2092, opened 2020-10-28
PTL fixes

ericpershey:
These are fixes and enhancements to PTL.  

- pbs_benchpress
    - fixes the nose debug output
    - added logger name when -F is specified

- fixes duplicate options for pbs_swigify

- removed erroneous copy of a file in ptl_resourceresv.py
- enhancement for logging in pbs_api_as
- added time to the message for _log_match

- pbs_dshutils.py
    - added get_method_name
    - added context in log message for run_cmd 
    - added context and a log message for run_copy
    - enhancement for chmod to not call out to os commands when the file is local
    - added logging to create_temp_file
    
- fixed problem with a KeyError when the server is down and pbs_testsuite.py searches for the default scheduler
- fixed problem with uninitialized self.jdata when no tests are run
- changed the logger of PTLTestLoader to use the logger in the class


vishwaks (approved):
Changes looks good to me.
@ericpershey thanks for making changes :+1:

test/fw/bin/pbs_benchpress
```diff
@@ -62,6 +64,23 @@ from ptl.utils.plugins.ptl_test_tags import PTLTestTags
 from ptl.utils.plugins.ptl_test_data import PTLTestData
 
 
+class PTLNoseConfig(nose.config.Config):
+    def __init__(self, log_format=None, outfile_handler=None, **kw):
+        super().__init__(**kw)
+        self.outfile_handler = outfile_handler
```
agrawalravi90, line 70: Does it make sense to accept a list of custom handlers here instead of the specific 'outfile_handler' and adding all of them inside configureLogging() ?

test/fw/ptl/lib/ptl_resourceresv.py
```diff
@@ -503,12 +503,6 @@ def receive_alarm(signum, stack):
         script_path = self.du.create_temp_file(hostname=hostname,
                                                body=script_body,
                                                suffix='.py')
-        if not self.du.is_localhost(hostname):
```
agrawalravi90, line 506: Why did you remove this?

test/fw/bin/pbs_benchpress
```diff
@@ -507,13 +530,14 @@ if __name__ == '__main__':
...
-    else:
-        os.environ['NOSE_VERBOSE'] = '2'
-    if stoponfail:
-        os.environ['NOSE_STOP'] = '1'
-    nose.main(defaultTest=tests, argv=[sys.argv[0]], plugins=plugins)
+    nose_config = PTLNoseConfig(
+        log_format=fmt,
+        outfile_handler=outfile_hdlr,
+        plugins=PluginManager(plugins=plugins),
+        testMatchPat=test_regex,
```
agrawalravi90, line 537: I didn't really find any documentation for 'testMatchPat', or much documentation for invoking nose.config.Config's constructor. I don't know a lot about nose, is it actually recommended to invoke nose.config.Config's constructor directly? If these are internal parameters then they might change and we might not want to use them.

agrawalravi90 (changes requested):
sorry for the late review on this, slipped through the cracks.

test/fw/ptl/lib/ptl_resourceresv.py
ericpershey: It's checking to see if it's not localhost, and if it's not it copies the file out there.  Then a few lines lower ("if not self.du.chmod.."), it does a chmod on that hostname/script_path again and that chmod does a copy again by running run_cmd.  This should reduce the copy calls and increase the speed of some of the tests.

test/fw/bin/pbs_benchpress
ericpershey: I don't know if it's worth it.  The code was added to allow nose debugging lines to be emitted, with timestamps and not duplicated.  If you think otherwise, I can update with master and update this, it's quite easy.

agrawalravi90: I just realized that this is not part of the library, it's just inside pbs_benchpress, so ya, don't bother.

As a note, you don't need to update with master when making further changes on your PRs since we do a rebase when merging the PRs.

test/fw/ptl/lib/ptl_resourceresv.py
agrawalravi90: From the code, it looks like create_temp_file() does a remote copy as well if hostname is not localhost. So, ya, this seems unnecessary, thanks for removing it.

test/fw/bin/pbs_benchpress
agrawalravi90: @ericpershey this is my only outstanding comment

agrawalravi90:
@ericpershey just want to make sure that this is not waiting on me, I think there's still one review comment that was unaddressed.

test/fw/bin/pbs_benchpress
toonen: Commit 7f0124ba44964be4bafcf17637e913380a9875aa removes the use of `testMatchPat` and instead sets the `NOSE_TESTMATCH` environment variable before creating the config object.  Please let us know if this is acceptable.

agrawalravi90: Thanks, but I'd like @hirenvadalia to vet the nose and test loader related changes once

test/fw/bin/pbs_benchpress
```diff
@@ -63,6 +64,23 @@ from ptl.utils.plugins.ptl_test_tags import PTLTestTags
 from ptl.utils.plugins.ptl_test_data import PTLTestData
 
 
+class PTLNoseConfig(nose.config.Config):
+    def __init__(self, log_format=None, outfile_handler=None, **kw):
+        super().__init__(**kw)
```
hirenvadalia, line 69: @ericpershey We need to add `env` in kw here so that it can pick `NOSE_TESTMATCH` environment while calling `nose.config.Config.__init__` else it will use default value of testmatch regex, which will break PTL.

See this: https://github.com/nose-devs/nose/blob/master/nose/config.py#L181 for how it uses `env` in Config class
See this: https://github.com/nose-devs/nose/blob/master/nose/core.py#L108 for how nose.main() passes env if config argument is None

hirenvadalia (changes requested):

agrawalravi90:
@ericpershey  and @toonen gentle reminder, thanks.

test/fw/bin/pbs_benchpress
toonen: Good catch.  Commit ebb19ce should fix this issue.

hirenvadalia (approved):

agrawalravi90 (approved):

PR #2095, opened 2020-10-30
Automate test from reservations and dedicated time test suite

shilpakodli:
#### Describe Bug or Feature
Automated test from reservations and dedicated time test suite.
Intention of test is to set sched attribute resv_confirm_ignore to dedicated_time and check reservation submission which conflicts with dedicated time

[test_reservation_logs.txt](https://github.com/openpbs/openpbs/files/5464043/test_reservation_logs.txt)

### DIFF
--- a/test/tests/functional/pbs_reservations.py
+++ b/test/tests/functional/pbs_reservations.py
@@ -2464,9 +2464,11 @@ class TestReservations(TestFunctional):
         next_day_start = PBSLogUtils.convert_date_time(
             next_day_start.strftime("%m/%d/%Y %H:%M:%S"))
         next_day_end = next_day_start + 86400
-        self.scheduler.set_sched_config({'resv_confirm_ignore': 'dedicated_time'})
+        self.scheduler.set_sched_config({'resv_confirm_ignore':
+                                         'dedicated_time'})
 
-        self.scheduler.add_dedicated_time(start=next_day_start, end=next_day_end)
+        self.scheduler.add_dedicated_time(start=next_day_start,
+                                          end=next_day_end)
         now = time.time()
         rid = self.submit_reservation(user=TEST_USER, select='1:ncpus=1',
                                       start=now + 5, end=now + 3600)
--- a/test/tests/functional/pbs_reservations.py
+++ b/test/tests/functional/pbs_reservations.py
@@ -2452,3 +2452,72 @@ class TestReservations(TestFunctional):
         self.assertEqual(
             jid2[1], '(' + vn + '[1]:ncpus=4+' + vn + '[2]:ncpus=4)')
         self.assertNotIn(job1_node, job2_node, errmsg)
+
+    def test_resv_confirm_ignore_set_dedicated_time(self):
+        """
+        Test reservation timing with resv_confirm_ignore set to dedicated_time
+        in sched_config
+        """
+        next_day = datetime.date.today() + datetime.timedelta(days=1)
+        next_day_start = datetime.datetime.combine(
+            next_day, datetime.time(0, 0, 0, 0))
+        next_day_start = PBSLogUtils.convert_date_time(
+            next_day_start.strftime("%m/%d/%Y %H:%M:%S"))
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_reservations.py
```diff
@@ -2452,3 +2452,74 @@ def test_shared_exclusive_job_not_in_same_rsv_vnode(self):
...
+        J1 = Job(attrs=a)
+        jid1 = self.server.submit(J1)
+        self.server.expect(JOB, {ATTR_state: 'R'}, id=jid1)
+        self.server.delete(jid1, wait=True)
+        a = {'Resource_List.walltime': '23:00:00', ATTR_queue: resv_queue}
+        J2 = Job(attrs=a)
+        J2.set_sleep_time(3600)
+        jid2 = self.server.submit(J2)
+        msg = "Job would cross dedicated time boundary"
+        attrs = {ATTR_comment: "Not Running: " + msg, ATTR_state: 'Q'}
```
agrawalravi90, line 2488: i don't see this getting used (missing an expect?). From what I understand, since you are submitting a reservation that's only 1 hour long but try to submit a job to it which is 23 hours long, the job will not run because it will cross the reservation's boundary itself, right?

test/tests/functional/pbs_reservations.py
```diff
@@ -2452,3 +2452,74 @@ def test_shared_exclusive_job_not_in_same_rsv_vnode(self):
...
+        self.server.expect(JOB, attrs, id=jid2)
+        self.server.delete(jid2, wait=True)
+        self.server.delete(rid, wait=True)
+
+        now = time.time()
+        self.scheduler.add_dedicated_time(start=now+10, end=now+300)
```
agrawalravi90, line 2514: i think you should delete the previous dedicated time window before creating this otherwise if this test is run around midnight, the two dedicated time windows will conflict with each other

test/tests/functional/pbs_reservations.py
```diff
@@ -2452,3 +2452,74 @@ def test_shared_exclusive_job_not_in_same_rsv_vnode(self):
...
+        rid = self.submit_reservation(user=TEST_USER, select='1:ncpus=1',
+                                      start=now + 5, end=now + 60)
+        a = {'reserve_state': (MATCH_RE, 'RESV_RUNNING|5')}
+        self.server.expect(RESV, a, id=rid, offset=5)
+        resv_queue = rid.split('.')[0]
+        a = {'Resource_List.walltime': '00:02:00', ATTR_queue: resv_queue}
+        J1 = Job(attrs=a)
+        jid1 = self.server.submit(J1)
+        msg = "Dedicated time conflict"
```
agrawalravi90, line 2523: how is this scenario different from the previous one? Both create a reservation which overlaps with a dedicated time window. Both submit a job which will cross into the dedicated time window. So, why are the rejection reasons different? Am I misunderstanding something about this scenario?

agrawalravi90 (changes requested):

test/tests/functional/pbs_reservations.py
bhroam: No, we don't backfill till the end of the reservation.  We run a job because there is a chance its walltime is wrong and it'll finish.  The jobs are going to be deleted at the end of the reservation anyway, so there is no reason to not run it.  The scheduler is smart enough to set the end of the job running in a reservation to the end of the reservation if the end time of the job is past the end of the reservation.  This is actually wrong, the scheduler shouldn't complain about dedicated time since the reservation ends before dedicated time starts.

bhroam:
@shilpakodli this is an undocumented feature, and one that really doesn't work all that well.  You can submit the reservation that crosses over dedicated time, but you still can't run jobs during dedicated time.  What's the real point?  I wouldn't test it.

shilpakodli:
HI @bhroam and @agrawalravi90 for the review. As this is an undocumented feature, as suggested by Bhroam, I am closing this PR.

PR #2099, opened 2020-10-30
Add gcc-c++ to INSTALL

### LINKED ISSUE #2098: INSTALL doesn't have gcc-c++ requirement
The install document is wrong; since the scheduler is now built with g++, we need to add that requirement.

Thanks to : https://community.openpbs.org/t/error-on-make-rhel-7-7/2337

vstumpf:
#### Describe Bug or Feature
gcc-c++ wasn't listed as a requirement in the INSTALL document
Fixes #2098 

#### Describe Your Change
Add it to the list

#### Link to Design Doc
N/A

N/A


bhroam:
Don't we need to update the spec file too?

vstumpf:
The spec file was already updated with the C++ change. It's just this that needs fixing.

prakashcv13 (approved):

kjakkali (approved):

PR #2100, opened 2020-11-02
job_is_running compatible with any printjob output

alexis-cousein:
#### Describe Bug or Feature

In "master" the external interface to printjob has changed -- states are now displayed as a string value (e.g. R for running) instead of hex values prepended with 0x (e.g. 0x4 for running state).

Regardless of the wisdom to change the external interface of a program while fixing a bug ( in PR #2058 ), this breaks backward compatibility of the master cgroup hook for no particularly good reason, and some sites have been relying on the master hook as backward compatible (because pre-2020.x the shipped hook is broken). 

Perhaps there will come a day in which maintaining backward compatibility will have to be abandoned (which will lead into having to fork the cgroup hook if extra features or bug fixes are introduced after that), but today is not that day (in my opinion).

Unfortunately, customers might be unaware the hook is broken on first inspection (and one such customer ran into this) -- there is no crash or syntax error, but job resource updates are simply completely disabled. 

#### Describe Your Change

It is easy not to "pick sides" and to make a hook that will tolerate _both_ flavours of "printjob" output. 

We don't HAVE to break compatibility with the older releases of PBSPro just to make master work:

```
    if ('state' in jobinfo
            and jobinfo['state'] != 'R'
            and jobinfo['state'] != 4):
        return False
```

I have also at DEBUG4 added a log message that will output what the hook thinks printjob returned -- right now that is silent, so any problems in running printjob are harder to debug than necessary.

#### Link to Design Doc

No design doc -- this is a change to retain backward compatible. In fact, where a design doc would have been needed is where we changed the interface of printjob silently while fixing a bug ;-).

Without changes, the last output of a job returns this (with full logging) when using the master hook on an 18.x (or 19.x or 2020.x) MoM:
```
11/02/2020 17:09:54;0800;pbs_python;Hook;pbs_python;update_job_usage: Method called
11/02/2020 17:09:54;0800;pbs_python;Hook;pbs_python;update_job_usage: resc_used = cpupercent=0,cput=00:00:00,ncpus=1,vmem=302228kb,mem=0b,walltime=00:00:01
11/02/2020 17:09:54;0800;pbs_python;Hook;pbs_python;Running: ['/opt/pbs/bin/printjob', '-a', '/var/spool/PBS18/mom_priv/jobs/342.tc72.JB']
11/02/2020 17:09:54;0800;pbs_python;Hook;pbs_python;update_job_usage: Job 342.tc72 is not running
```
With the addition of a debug message this will reveal why the hook thinks this:

```
11/02/2020 17:09:54;0800;pbs_python;Hook;pbs_python;JB file info returned: {'svrflgs': 1, 'substate': 42, 'ordering': 0, 'inter prior': 0, 'jobid': '342.tc72', 'state': 4, 'stime': 1604333391, 'file base': '342.tc72'}
```

With the changes, the job updates will be performed:
```
11/02/2020 19:02:11;0800;pbs_python;Hook;pbs_python;update_job_usage: Method called
11/02/2020 19:02:11;0800;pbs_python;Hook;pbs_python;update_job_usage: resc_used = cpupercent=0,cput=00:00:00,ncpus=1,vmem=302228kb,mem=0b,walltime=00:00:01
11/02/2020 19:02:11;0800;pbs_python;Hook;pbs_python;Running: ['/opt/pbs/bin/printjob', '-a', '/var/spool/PBS18/mom_priv/jobs/343.tc72.JB']
11/02/2020 19:02:11;0800;pbs_python;Hook;pbs_python;JB file info returned: {'svrflgs': 1, 'substate': 42, 'ordering': 0, 'inter prior': 0, 'jobid': '343.tc72
', 'state': 4, 'stime': 1604340130, 'file base': '343.tc72'}
11/02/2020 19:02:11;0008;pbs_python;Job;343.tc72;update_job_usage: CPU percent: 0
11/02/2020 19:02:11;0800;pbs_python;Hook;pbs_python;_get_cpu_usage: Method called
11/02/2020 19:02:11;0800;pbs_python;Hook;pbs_python;_cgroup_path: Method called
11/02/2020 19:02:11;0008;pbs_python;Job;343.tc72;update_job_usage: CPU usage: 0.097 secs
11/02/2020 19:02:11;0800;pbs_python;Hook;pbs_python;_get_max_mem_usage: Method called
11/02/2020 19:02:11;0800;pbs_python;Hook;pbs_python;_cgroup_path: Method called
11/02/2020 19:02:11;0008;pbs_python;Job;343.tc72;update_job_usage: Memory usage: mem=5356kb
11/02/2020 19:02:11;0800;pbs_python;Hook;pbs_python;_get_mem_failcnt: Method called
```

No PTL test can be written for this and run on a regular test harness, since the error will only reveal itself while running the hook in master on an older version of PBSPro/OpenPBS. Existing PTL tests that check resources_used will fail, though (when run with the current master hook on an older verson of PBSPro/OpenPBS).


alexis-cousein:
There is an alternative and even leaner and more robust way of handling this: since the routine is already checking for substate 42 (which has not changed), and jobs in substate 42 are always in state 'R', I think that we could just as well nuke the first "if" in the routine.

I'll defer to the opinion of the reviewers as to which solution is preferable.

bayucan:
@agrawalravi90 : I need you to review this as well...

agrawalravi90 (changes requested):
@alexis-cousein sorry that you ran into this. I did think about creating a design document for the printjob change, but I couldn't find printjob's output format documented in our docs, so I didn't think that it was necessary. Seems like I was wrong so I've created one now: https://openpbs.atlassian.net/wiki/spaces/PD/pages/2217574428/Printjob+command+will+print+job+state+letter+instead+of+numeric

I do like your idea of using only the substate, can you please implement that?

alexis-cousein:
Change implemented.

Thanks for creating a design document, but normally these should be discussed on the forum as well, and I haven't seen that.

THe gist is that printjob is old -- old enough not to have a _documented_ interface. That means that within reason we're free to change its interface, but that doesn't mean we should.

Customers rely on printjob of JB files to get information about jobs without spamming the server, and for that I do not think there is an alternative (which is why the cgroup hook uses it). In that use case, the printjob man page's suggestion to use qstat instead is, of course, not applicable, since calling qstat is exactly what the site using printjob is trying to avoid.

Yes, they are indeed relying on an undocumented interface, but that's because we didn't care to produce a documented one.

Changing its (admittedly undocumented) interface will break customer scripts in the same way it broke the cgroup hook, so we should be careful to do this only with a strong reason for doing so, especially if we haven't given any warning of an impending change (through e.g. a deprecation).

I don't think that reason exists yet -- it's not because the attribute value reads 'R' that we cannot look up PBS_STATE_RUNNING,  convert it to hex and then still print 0x4 as the 'state' in the beginning of the output, regardless of how things are now encoded in the server. At the very least, if we do change the interface we should probably also implement a flag so that the "old" output format can still be produced.

But that's something I should be writing in the forum discussion, not here. This is supposed to be just about the cgroup hook, and I agree that checking only the substate seems the most parsimonious solution.

agrawalravi90:
It sounds like sites are using it for a use case that it was never intended to be used for. If they don't want to use qstat because pinging the server makes things slow, then we should fix that performance problem (and we are, in fact, in the process of improving it with the diff-stat project). TBH, it doesn't really make sense to deprecate an unsupported interface, I'm also wondering how my design document explaining this change in printjob's output will be documented in our guides since we don't document what the previous output looked like. So, I have to disagree, I don't think this needs a design discussion.

agrawalravi90 (approved):

agrawalravi90 (changes requested):
Actually i just noticed that you've changed the permissions of the file as well, was that intentional?

vstumpf (approved):
LGTM, with or without the permission change.

alexis-cousein:
Changed the permissions back to what they were.

alexis-cousein:
> It sounds like sites are using it for a use case that it was never intended to be used for.

Uhm -- even Altair does that (wittness the cgroup hook). 

I am also not a firm believer in the theory you can just change interfaces as long as it only makes a difference in the "anticipated usage", especially when there is no documentation limiting  the anticipated usage. BTW, I've seen people use it in this way _before_ OpenPBS became PBSPro.

> then we should fix that performance problem

I disagree. A program on MoM should not need to contact the server to know what the state of the job is *on MoM*; there's no need to create a problem and then after that "fix" it; if you have large systems it's essentially unfixable anyway as a scalability issue, because you are turning an O(1) problem (where every MoM never contacts the server) into an O(N) prolbem. 

There IS a reason why printjob was written to accept EITHER a job ID OR a JB file, so I'm rather skeptical about your hypothesis that the second usage was "not anticipated". If the author had desired the server to be consulted at all times, then they would not have coded the option to consult the JB file at all.

If you want to get rid of that "unanticipated usage", then we should probably extend the hook interface so that execjob_* can query MoMs, but given the architecture of MoM hooks that is a wide river to cross (unless of course that interface actually looks things up in the JB file as well). 

But even that won't change the "unantitcipated usage" that customers may make. As an L2 support person, "it's broken because you aren't using it the way we thought you would" is usually not an answer appreciated by customers. It's saying that the external interface to programs is whatever we decide to make it, regardless of backward compatibility.  While there _are_ situations where interface changes are warranted, they are usually announced, and there is a very good reason for them. I don't think not wanting to look up what internal value 'R' has and figuring out it's 0x4 warrants that.

vstumpf:
Hey alexis, I don't see an update to the permissions. Did you forget to push?

agrawalravi90:
Thanks for providing more details Alexis. Let me try to clarify where I'm coming from. I see this being similar to daemon logs (server, mom, comm and sched logs). It's also data that's printed by PBS, and a site can write their own script which parses this data and does something useful from it. But daemon logs are not a documented/supported interface (unlike accounting logs), we are free to add/remove daemon log messages. So, if a site uses them then it's at their own risk right? We never announce changes to unsupported interfaces, why start now? Why should printjob get special treatment?

I don't know the history of printjob, but I fail to understand why it was not documented in the same way that qstat was, whenever we decided to document stuff? In fact, we clearly document that it's only intended for troubleshooting purposes, and that sites should use qstat for all generic purposes. If qstat cannot fulfill a certain customer requirement (ability to query data from .JB files on mom), then I think we should enhance it instead of asking customers to rely on an unsupported interface.

If you still disagree, then maybe we should get opinions from others. I request @openpbs/openpbs-maintainers  to chime in as well, thanks.

alexis-cousein:
"In fact, we clearly document that it's only intended for troubleshooting purposes"

Verbatim quote, emphasis mine:
" This command is _**mainly**_ useful for troubleshooting"

to me, "only" and "mainly" are clearly different. the more salient point which you're still not acknowledging is that it is our _only_ documented interface to the opaque JB file structures, and the only way _I_ know of fiunding out what's in MoM's head about a job.

"I don't know the history of printjob, but I fail to understand why it was not documented in the same way that qstat was, whenever we decided to document stuff?"
I think you weren't around in the early days of PBSPro ;-).

But look, here we are having a discussion, but this is not the right forum.

alexis-cousein:
> Hey alexis, I don't see an update to the permissions. Did you forget to push?

Guilty as charged, your honour. Pushed.

vstumpf (approved):
> Guilty as charged, your honour. Pushed.

Judge number 2 approves this motion.

agrawalravi90 (approved):

agrawalravi90:
@bayucan over to you, thanks

agrawalravi90:
@alexis-cousein https://community.openpbs.org/t/do-changes-to-printjobs-output-need-documentation/2345

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -670,8 +671,6 @@ def job_is_running(jobid):
     Returns True if job shows a running state and substate
     """
     jobinfo = printjob_info(jobid)
-    if 'state' in jobinfo and jobinfo['state'] != 'R':
-        return False
```
saimanikant, line 674: Should the values in job_is_suspended() method also be updated? I see that it is also using printjob_info()

alexis-cousein: No, that was already using only the substate, which is always numerical. In fact, this pull request makes job_is_running consistent with the current job_is_suspended routine.

mike0042 (commented):
LGTM

bayucan (approved):
Looks reasonable.

bayucan:
Merged #2100 into master.

alexis-cousein:
Thanks all!

PR #2102, opened 2020-11-03
ubuntu20.04 compilation issues

birbalsain:
#### Describe Bug or Feature
Getting warnings while compiling on ubuntu20 that is using higher gcc version 9.3. Read and write size doesn't match in snprintf(), structure sockaddr doesn't give correct size of ipv6. reading data from null string.
[ubuntu_issue.txt](https://github.com/openpbs/openpbs/files/5481183/ubuntu_issue.txt)

#### Describe Your Change
Uses structure sockaddr_storage to store ipv6. Provides required size to the writing destination. Uses if condition to check null string,

[ubuntu_build.txt](https://github.com/openpbs/openpbs/files/5483651/ubuntu_build.txt)

Description: Tests from given build on platforms UBUNTU1804, SUSE15, UBUNTU1604, CENTOS8, RHEL8, SLES12, SLES15
Platforms: UBUNTU1804,SUSE15,UBUNTU1604,CENTOS8,RHEL8,SLES12,SLES15
Profile: regression
|--|--|--|--|--|--|--|
|4663|3845|14|0|0|0|3831|

### DIFF
--- a/src/cmds/qsub.c
+++ b/src/cmds/qsub.c
@@ -1833,7 +1833,7 @@ read_job_script(char *script)
 	int errflg; /* error code from get_script() */
 	struct stat statbuf;
 	char *bnp;
-	char basename[PBS_MAXJOBNAME + 1]; /* base name of script for job name*/
+	char basename[MAXPATHLEN + 1]; /* base name of script for job name*/
 	FILE *f; /* FILE pointer to the script */
 
 	/* if script is empty, get standard input */
--- a/src/cmds/qsub.c
+++ b/src/cmds/qsub.c
@@ -1833,7 +1833,7 @@ read_job_script(char *script)
 	int errflg; /* error code from get_script() */
 	struct stat statbuf;
 	char *bnp;
-	char basename[MAXPATHLEN + 1]; /* base name of script for job name*/
+	char basename[PBS_MAXJOBNAME + 1]; /* base name of script for job name*/
 	FILE *f; /* FILE pointer to the script */
 
 	/* if script is empty, get standard input */
--- a/src/include/pbs_ifl.h
+++ b/src/include/pbs_ifl.h
@@ -466,7 +466,7 @@ enum mgr_obj {
 #define PBS_MAXPWLEN		256		/* max password length */
 #define PBS_MAXGRPN		256		/* max group name length */
 #define PBS_MAXQUEUENAME	15		/* max queue name length */
-#define PBS_MAXJOBNAME  	MAXPATHLEN + 1	/* max job name length */
+#define PBS_MAXJOBNAME  	230		/* max job name length */
 #define PBS_MAXSERVERNAME	PBS_MAXHOSTNAME	/* max server name length */
 #define PBS_MAXSEQNUM		12		/* max sequence number length */
 #define PBS_DFLT_MAX_JOB_SEQUENCE_ID 9999999	/* default value of max_job_sequence_id server attribute */
--- a/src/include/pbs_ifl.h
+++ b/src/include/pbs_ifl.h
@@ -466,7 +466,7 @@ enum mgr_obj {
 #define PBS_MAXPWLEN		256		/* max password length */
 #define PBS_MAXGRPN		256		/* max group name length */
 #define PBS_MAXQUEUENAME	15		/* max queue name length */
-#define PBS_MAXJOBNAME  	230		/* max job name length */
+#define PBS_MAXJOBNAME  	MAXPATHLEN + 1	/* max job name length */
 #define PBS_MAXSERVERNAME	PBS_MAXHOSTNAME	/* max server name length */
 #define PBS_MAXSEQNUM		12		/* max sequence number length */
 #define PBS_DFLT_MAX_JOB_SEQUENCE_ID 9999999	/* default value of max_job_sequence_id server attribute */
--- a/src/lib/Libtpp/tpp_util.c
+++ b/src/lib/Libtpp/tpp_util.c
@@ -2081,18 +2081,19 @@ tpp_get_addresses(char *names, int *count)
 tpp_addr_t *
 tpp_get_local_host(int sock)
 {
-	struct sockaddr addr;
+	struct sockaddr_storage addrs;
+	struct sockaddr *addr = (struct sockaddr *)&addrs;
 	struct sockaddr_in *inp = NULL;
 	struct sockaddr_in6 *inp6 = NULL;
 	tpp_addr_t *taddr = NULL;
-	socklen_t len = sizeof(addr);
+	socklen_t len = sizeof(struct sockaddr);
 
-	if (getsockname(sock, &addr, &len) == -1) {
+	if (getsockname(sock, addr, &len) == -1) {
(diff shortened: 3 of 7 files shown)


src/tools/pbs_idled.c
```diff
@@ -117,7 +117,7 @@ main(int argc, char *argv[], char *envp[])
 	char *username;
 	struct xy cur_xy, prev_xy;
 	struct stat st;
-	char errbuf[256];
+	char errbuf[1037];
```
hirenvadalia, line 120: Can't we use some macro here instead this magic number?

src/tools/pbs_python.c
```diff
@@ -1648,7 +1648,7 @@ main(int argc, char *argv[], char *envp[])
 		int	print_env = 0;
 		char	*tmp_str = NULL;
 		char	perf_label[MAXBUF];
-		char	perf_action[MAXBUFLEN];
+		char	perf_action[MAXBUFLEN + 13];
```
hirenvadalia, line 1651: Please add comment saying we need `+ 13`

src/resmom/mom_hook_func.c
```diff
@@ -750,7 +750,7 @@ run_hook(hook *phook, unsigned int event_type, mom_hook_input_t *hook_input,
 	pid_t myseq; /* just some unique sequence number */
 	char logmask[BUFSIZ];
 	char path_hooks_rescdef[MAXPATHLEN + 1];
-	char cmdline[2 * BUFSIZ + 1];
+	char cmdline[3 * BUFSIZ + 1];
```
hirenvadalia, line 753: better to add comment saying why `3 * size + 1`

src/tools/pbs_python.c
birbalsain: perf_action is used to write hook perf output and description string, where hook perf output size is MAXBUFLEN so need additional size for description string.

hirenvadalia: @birbalsain Can you please add whatever you said in code comment?

src/cmds/qsub.c
```diff
@@ -1833,7 +1833,7 @@ read_job_script(char *script)
 	int errflg; /* error code from get_script() */
 	struct stat statbuf;
 	char *bnp;
-	char basename[PBS_MAXJOBNAME + 1]; /* base name of script for job name*/
+	char basename[MAXPATHLEN + 1]; /* base name of script for job name*/
```
agrawalravi90, line 1836: why did you change this? PBS_MAXJOBNAME is supposed to be used for this exact purpose.

src/lib/Libtpp/tpp_util.c
```diff
@@ -2081,18 +2081,19 @@ tpp_get_addresses(char *names, int *count)
 tpp_addr_t *
 tpp_get_local_host(int sock)
 {
-	struct sockaddr addr;
+	struct sockaddr_storage addrs;
+	struct sockaddr *addr = (struct sockaddr *)&addrs;
```
agrawalravi90, line 2085: is this really ok? it's usually very dangerous to cast pointer of one type to another, unless you are casting a void pointer, It can lead to unexpected runtime issues.

agrawalravi90 (changes requested):

src/cmds/qsub.c
birbalsain: basename holds script file that is static char script[MAXPATHLEN + 1] = "";

agrawalravi90: then we should change PBS_MAXJOBNAME to be MAXPATHLEN + \<N>.

src/lib/Libtpp/tpp_util.c
vstumpf: It should be fine, according to [spec](https://pubs.opengroup.org/onlinepubs/009695399/basedefs/sys/socket.h.html)

> When a sockaddr_storage structure is cast as a sockaddr structure, the ss_family field of the sockaddr_storage structure shall map onto the sa_family field of the sockaddr structure. When a sockaddr_storage structure is cast as a protocol-specific address structure, the ss_family field shall map onto a field of that structure that is of type sa_family_t and that identifies the protocol's address family.

agrawalravi90: Thanks Vincent. From what I understand, this states that it's ok to cast sockaddr_storage to sockaddr. My concern is regarding the pointer cast from sockaddr_storage * to sockaddr *. I've been bitten in the past by casting int * to long *, the compiler did weird things with it at higher optimization levels. So, just to be on the safe side, can we do "struct sockaddr addr = (struct sockaddr ) addrs;" instead ?

src/include/pbs_ifl.h
```diff
@@ -466,7 +466,7 @@ enum mgr_obj {
 #define PBS_MAXPWLEN		256		/* max password length */
 #define PBS_MAXGRPN		256		/* max group name length */
 #define PBS_MAXQUEUENAME	15		/* max queue name length */
-#define PBS_MAXJOBNAME  	230		/* max job name length */
+#define PBS_MAXJOBNAME  	MAXPATHLEN + 1	/* max job name length */
```
hirenvadalia, line 469: This is not right... PBS_MAXJOBNAME is what we can pass in `-N` in qsub/pbs_rsub, not complete path to script. see `decode_jobname()` action func... Please revert this.

hirenvadalia (changes requested):

src/cmds/qsub.c
hirenvadalia: @agrawalravi90 PBS_MAXJOBNAME is not full path of job script, its max job name (`-N` in qsub/pbs_rsub) which is derived from job script's path if `-N` not given. So I don't agree with you that we should change PBS_MAXJOBNAME.

src/lib/Libtpp/tpp_util.c
hirenvadalia: @agrawalravi90 Its ok to cast pointer also in this case, if you read spec link given by @vstumpf then it says that:
> The <sys/socket.h> header shall define the sockaddr_storage structure. This structure shall be:
>
>    Large enough to accommodate all supported protocol-specific address structures
>
>    Aligned at an appropriate boundary so that pointers to it can be cast as pointers to protocol-specific address structures and used to access the fields of those structures without alignment problems

src/cmds/qsub.c
agrawalravi90: If job name is derived from job script's path then does it not make sense that the job name buffer be at least MAXPATHLEN in size? Anyways, the variable name and comment here seemed to suggest that it should use buffer size for job name, but I'm ok with anything as long as compiler clears it.

src/lib/Libtpp/tpp_util.c
agrawalravi90: Thanks for looking into it, I guess if the spec guarantees the safety then we can use it.

agrawalravi90 (approved):
Approving this since it seems urgent and I have no problems with the changes.

hirenvadalia (approved):

PR #2103, opened 2020-11-03
Remove duplicate current_user if already exists in manager user list

visheshh:
#### Describe Bug or Feature
If current user is a manager, a duplicate entry to the list of manager users

#### Describe Your Change
Remove duplicate current_user if already exists in manager user list

[nightly.txt](https://github.com/openpbs/openpbs/files/5482869/nightly.txt)


hirenvadalia (approved):
@visheshh You need to fix pep8 issues

kjakkali (approved):

PR #2107, opened 2020-11-04
Remove indexed access of attributes from objects

hirenvadalia:
#### Describe Bug or Feature

* Remove indexed access of attributes of objects by using getter/setter of object
* Add getter/setter for queue, node, server, reservation, sched object (job object already has getter/setter)
* Use getter/setter to get/set value of attribute

#### Describe Your Change

* Added getter/setter for renaming objects
* Removed indexed access of attributes of object by using getter/setter of object

#### Link to Design Doc

None

Coming soon


hirenvadalia:
@ashwathraop has taken over this, he will raise PR once ready. Closing this one.

PR #2108, opened 2020-11-04
Scheduler server_dyn_res leaves trail of zombies depleting host PID space

alexis-cousein:
#### Describe Bug or Feature

We've just discovered a PID leak in 2020 in the scheduler: the method to spawn a child was changed from using popen() to fork() to allow for setting up an alarm and killing a child that would hang the scheduler, but code is missing to use waitpid() to clean up the zombies after reading the result of the server_dyn_res child.

The result is a PID leak (a zombie process left in the process table for every server_dyn_res resource every cycle):

[root@tc72linux2 sched_priv]# grep wobbly sched_config
    server_dyn_res: "wobbly !/bin/echo 10"
[root@tc72linux2 sched_priv]# qmgr
Max open servers: 49
Qmgr: s s scheduling=true
Qmgr: s s scheduling=true
Qmgr: s s scheduling=true
Qmgr: s s scheduling=true
Qmgr: s s scheduling=true
Qmgr: s s scheduling=true
Qmgr: s s scheduling=true
Qmgr: s s scheduling=true
Qmgr: s s scheduling=true
Qmgr: s s scheduling=true

[root@tc72linux2 sched_priv]# ps -ef | grep pbs_sched
root     11206     1  0 17:24 ?        00:00:00 /opt/pbs/sbin/pbs_sched
root     11649  2013  0 17:29 pts/0    00:00:00 grep --color=auto pbs_sched
[root@tc72linux2 sched_priv]# pstree -c 11206
pbs_sched─┬─echo
          ├─echo
          ├─echo
          ├─echo
          ├─echo
          ├─echo
          ├─echo
          ├─echo
          ├─echo
          ├─echo
          └─{pbs_sched}

To work around this, sites will have to periodically restart the scheduler manually, or they'll risk depleting the PID space on their hosts.

#### Describe Your Change

Insert the same loop that uses waitpid() to clean up children that is already there when there is an error in the case that everything actually works. Initialize pid to zero in the loop and add a guard to avoid risking calling waitpid() on an uninitialized PID.
#### Link to Design Doc

N/A -- bug fix

See above for test showing the bug. When fixed, pstree shows a pbs_sched that has no "echo" zombie children.

### DIFF
--- a/src/scheduler/server_info.cpp
+++ b/src/scheduler/server_info.cpp
@@ -749,6 +749,8 @@ query_server_dyn_res(server_info *sinfo)
 				} else if (ret == 0) {
 					log_eventf(PBSEVENT_DEBUG, PBS_EVENTCLASS_SERVER, LOG_DEBUG, "server_dyn_res",
 					"Program %s timed out", conf.dynamic_res[i].command_line);
+				}
+				if (ret == -1 || ret == 0) {
 					/* we know pid is initialized here, otherwise pipe_err would have been set */
 					kill(-pid, SIGTERM);
 					if (waitpid(pid, NULL, WNOHANG) == 0) {
--- a/src/scheduler/server_info.cpp
+++ b/src/scheduler/server_info.cpp
@@ -751,7 +751,7 @@ query_server_dyn_res(server_info *sinfo)
 					log_eventf(PBSEVENT_DEBUG, PBS_EVENTCLASS_SERVER, LOG_DEBUG, "server_dyn_res",
 					"Program %s timed out", conf.dynamic_res[i].command_line);
 				}
-				if ( pid > 0 && ret > 0 ) {
+				if (pid > 0 && ret > 0) {
 					/* Parent; only open if child created and select showed sth to read,
 					 * but assume fdopen can't fail
 					 */
--- a/src/scheduler/server_info.cpp
+++ b/src/scheduler/server_info.cpp
@@ -682,6 +682,8 @@ query_server_dyn_res(server_info *sinfo)
 				sinfo->res = res;
 
 			pipe_err = errno = 0;
+			pid = 0;
+
 			/* Make sure file does not have open permissions */
 			#if !defined(DEBUG) && !defined(NO_SECURITY_CHECK)
 				filename = conf.dynamic_res[i].script_name;
@@ -748,11 +750,13 @@ query_server_dyn_res(server_info *sinfo)
 					log_eventf(PBSEVENT_DEBUG, PBS_EVENTCLASS_SERVER, LOG_DEBUG, "server_dyn_res",
 					"Program %s timed out", conf.dynamic_res[i].command_line);
 					kill(-pid, SIGTERM);
-					if (waitpid(pid, NULL, WNOHANG) == 0) {
-						usleep(250000);
-						if (waitpid(pid, NULL, WNOHANG) == 0) {
-							kill(-pid, SIGKILL);
-							waitpid(pid, NULL, 0);
+					if (pid != 0) {
+					    if (waitpid(pid, NULL, WNOHANG) == 0) {
+							usleep(250000);
+							if (waitpid(pid, NULL, WNOHANG) == 0) {
+								kill(-pid, SIGKILL);
+								waitpid(pid, NULL, 0);
+							}
 						}
 					}
 				}
@@ -797,7 +801,15 @@ query_server_dyn_res(server_info *sinfo)
 			else
 				log_eventf(PBSEVENT_DEBUG2, PBS_EVENTCLASS_SERVER, LOG_DEBUG, "server_dyn_res",
 					"%s = %s (\"%s\")", conf.dynamic_res[i].command_line, res_to_str(res, RF_AVAIL), buf);
-
+			if (pid != 0) {
+				if (waitpid(pid, NULL, WNOHANG) == 0) {
+					usleep(250000);
+					if (waitpid(pid, NULL, WNOHANG) == 0) {
+						kill(-pid, SIGKILL);
+						waitpid(pid, NULL, 0);
+					}
+				}
+			}
 		}
 	}
 
--- a/src/scheduler/server_info.cpp
+++ b/src/scheduler/server_info.cpp
@@ -749,14 +749,13 @@ query_server_dyn_res(server_info *sinfo)
 				} else if (ret == 0) {
 					log_eventf(PBSEVENT_DEBUG, PBS_EVENTCLASS_SERVER, LOG_DEBUG, "server_dyn_res",
 					"Program %s timed out", conf.dynamic_res[i].command_line);
+					/* we know pid is initialized here, otherwise pipe_err would have been set */
 					kill(-pid, SIGTERM);
-					if (pid != 0) {
-					    if (waitpid(pid, NULL, WNOHANG) == 0) {
-							usleep(250000);
-							if (waitpid(pid, NULL, WNOHANG) == 0) {
-								kill(-pid, SIGKILL);
-								waitpid(pid, NULL, 0);
-							}
+					if (waitpid(pid, NULL, WNOHANG) == 0) {
+						usleep(250000);
+						if (waitpid(pid, NULL, WNOHANG) == 0) {
+							kill(-pid, SIGKILL);
+							waitpid(pid, NULL, 0);
 						}
 					}
 				}
@@ -801,7 +800,9 @@ query_server_dyn_res(server_info *sinfo)
 			else
 				log_eventf(PBSEVENT_DEBUG2, PBS_EVENTCLASS_SERVER, LOG_DEBUG, "server_dyn_res",
 					"%s = %s (\"%s\")", conf.dynamic_res[i].command_line, res_to_str(res, RF_AVAIL), buf);
+
 			if (pid != 0) {
+				kill(-pid, SIGTERM);
 				if (waitpid(pid, NULL, WNOHANG) == 0) {
 					usleep(250000);
 					if (waitpid(pid, NULL, WNOHANG) == 0) {
--- a/src/scheduler/server_info.cpp
+++ b/src/scheduler/server_info.cpp
@@ -760,19 +760,20 @@ query_server_dyn_res(server_info *sinfo)
 							waitpid(pid, NULL, 0);
 						}
 					}
+					pid = 0; /* don't kill it again later */
 				}
-
-				/* Parent; assume fdopen can't fail. */
-				fp = fdopen(pdes[0], "r");
-				close(pdes[1]);
-
-				if (fgets(buf, sizeof(buf), fp) == NULL) {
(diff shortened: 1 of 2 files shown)


src/scheduler/server_info.cpp
```diff
@@ -748,11 +750,13 @@ query_server_dyn_res(server_info *sinfo)
 					log_eventf(PBSEVENT_DEBUG, PBS_EVENTCLASS_SERVER, LOG_DEBUG, "server_dyn_res",
 					"Program %s timed out", conf.dynamic_res[i].command_line);
 					kill(-pid, SIGTERM);
-					if (waitpid(pid, NULL, WNOHANG) == 0) {
-						usleep(250000);
-						if (waitpid(pid, NULL, WNOHANG) == 0) {
-							kill(-pid, SIGKILL);
-							waitpid(pid, NULL, 0);
+					if (pid != 0) {
```
vstumpf, line 753: I'm not sure this check is currently necessary, since pid will always be set if we're running this code.

src/scheduler/server_info.cpp
```diff
@@ -797,7 +801,15 @@ query_server_dyn_res(server_info *sinfo)
 			else
 				log_eventf(PBSEVENT_DEBUG2, PBS_EVENTCLASS_SERVER, LOG_DEBUG, "server_dyn_res",
 					"%s = %s (\"%s\")", conf.dynamic_res[i].command_line, res_to_str(res, RF_AVAIL), buf);
-
+			if (pid != 0) {
+				if (waitpid(pid, NULL, WNOHANG) == 0) {
```
vstumpf, line 805: Do we need to kill the process here?
If we made it here, then we've already set the resource, but the script could technically still be running (if the script prints, then sleeps). 
```suggestion
				kill(-pid, SIGTERM);
				if (waitpid(pid, NULL, WNOHANG) == 0) {
```

vstumpf (changes requested):
Thanks for this, I don't know how I screwed this up so much. I have some comments though.

src/scheduler/server_info.cpp
alexis-cousein: For this one, yes. But I found it more consistent to have this check in both places.

src/scheduler/server_info.cpp
alexis-cousein: Yes, that's an omission.

src/scheduler/server_info.cpp
alexis-cousein: I am always enclined to _directly_ check whether arguments to a system call are sane rather than rely on the code before (in this case that pipe_err was set, so that we don't end up here). 

Frequently, what works without being defensive becomes a bug when you refactor the other code.

But I'll remove it and replace it with a comment (just as a warning that there _is_ an assumption there you'd better not break).

alexis-cousein:
I've pushed a new version. Rerequesting review.

vstumpf (approved):
LGTM

alexis-cousein:
@vstumpf I see that we've forgotten to deal with one corner case -- if the select fails, then we _also_ do not kill the child. I will be submitting a new version.

src/scheduler/server_info.cpp
```diff
@@ -747,6 +749,9 @@ query_server_dyn_res(server_info *sinfo)
 				} else if (ret == 0) {
 					log_eventf(PBSEVENT_DEBUG, PBS_EVENTCLASS_SERVER, LOG_DEBUG, "server_dyn_res",
 					"Program %s timed out", conf.dynamic_res[i].command_line);
+				}
+				if (ret == -1 || ret == 0) {
```
vstumpf, line 753: I'm not sure this is necessary, since it will get cleaned up in the lower check.

src/scheduler/server_info.cpp
```diff
@@ -798,6 +803,16 @@ query_server_dyn_res(server_info *sinfo)
 				log_eventf(PBSEVENT_DEBUG2, PBS_EVENTCLASS_SERVER, LOG_DEBUG, "server_dyn_res",
 					"%s = %s (\"%s\")", conf.dynamic_res[i].command_line, res_to_str(res, RF_AVAIL), buf);
 
+			if (pid != 0) {
```
vstumpf, line 806: Since we are checking pid again, shouldn't we set it to 0 if we wait for it above? That way, we don't go through this code again.

vstumpf (changes requested):
Hey alexis, some more changes.

src/scheduler/server_info.cpp
bhroam, line 806: I think you want pid > 0.  If pid is negative, the fork() failed.

bhroam (changes requested):

src/scheduler/server_info.cpp
alexis-cousein: Then why don't we just do it once at the end? Lean and mean...

src/scheduler/server_info.cpp
alexis-cousein: See above -- I think I will just change it so that we always kill the child at the end, regardless of what happens.

Bhroam is right that we should only do that if pid is positive.

alexis-cousein: Correct. If the fork fails, though, then we set pipe_err, but in this one we no longer have that guard.

src/scheduler/server_info.cpp
alexis-cousein: We can do several things. We can leave things as is, but that would entail reading a file descriptor tying us to a killed program in case of errors. That will work, but it looks weird to even attempt to read there. It _is_ the minimal change, though.

The second alternative is to just set pid to 0 when we've killed the child, set k to 0,  and surround the parent fdopen/fgets section with if (pid>0). In that case we'll just skip even trying to read something if we had an error on the select earlier i.e. if there's nothing to be read we shouldn't even try to read.

The third alternative is to just use "continue" to go on to the next server_dyn_res resource if we had an error on the select and we killed the child.

I don't have a preference. I'll defer to @vstumpf and @bhroam as to which os the variants they like best. I'll pick the first one because it is the minimally invasive one, but I may change that when you two review it and come to a consensus.

vstumpf: I'm not sure we can use `continue` without also logging "Setting resource to 0" and calling `set_resource()`.
I like the second option.

bhroam: We could change the code flow a little and fix this.  We really have two code flows.  One error case where we kill the process and the normal case where we read/set the resource.  How about instead of checking for error first, we check for not error and do the read/set.  Either side of that if will drop us to the kill at the bottom.

vstumpf:
I wrote a test for this, if you want to add it:

```Python
from ptl.utils.pbs_procutils import ProcUtils
...
    def test_res_cleanup(self):
        """
        Test that the scheduler cleans up its children
        """
        pu = ProcUtils()
        resname = ["normal", "invalid", "timeout"]
        restype = ["long", "long", "long"]

        # Prep for server_dyn_resource scripts.
        script_body = ["echo 8", "echo hello", "sleep 40; echo 20"]

        filenames = self.setup_dyn_res(resname, restype, script_body)

        a = {'Resource_List.normal': '2',
             'Resource_List.invalid': '8',
             'Resource_List.timeout': 10}
        j = Job(TEST_USER, attrs=a)
        jid = self.server.submit(j)
        self.logger.info('Sleeping 30 seconds to wait for script to timeout')
        time.sleep(30)
        self.scheduler.log_match("%s timed out" % filenames[2])
        children = pu.get_proc_children(hostname=self.scheduler.hostname,
                                        ppid=self.scheduler.get_pid())
        self.assertFalse(children)
```

If the children aren't cleaned up, `children` will not be False:
```
2020-11-06 11:22:29,250 INFO     ===================================
2020-11-06 11:22:29,250 INFO     Completed TestServerDynRes tearDown
2020-11-06 11:22:29,251 INFO     ===================================
2020-11-06 11:22:29,256 INFO     FAILED

2020-11-06 11:22:29,259 INFO     
2020-11-06 11:22:29,259 INFO     ======================================================================
2020-11-06 11:22:29,259 INFO     FAILED: test_res_cleanup (tests.functional.pbs_svr_dyn_res.TestServerDynRes)

2020-11-06 11:22:29,259 INFO     ___m_oo_m___
2020-11-06 11:22:29,259 INFO     Traceback (most recent call last):
  File "/home/vstumpf/dev/pbspro/test/tests/functional/pbs_svr_dyn_res.py", line 615, in test_res_cleanup
    self.assertFalse(children)
AssertionError: ['78976', '78977'] is not false
```

alexis-cousein:
@vstumpf Any guidance on what you want me to do wrt the code changes (continuing vs. skipping the rest of the loop body when the select has failed)?

In what file do you want me to add the PTL test?

vstumpf:
You can add it to `functional/pbs_svr_dyn_res.py`

alexis-cousein:
Done. BTW, how does pbs_benchpress know what class methods to call to do the tests? Does it go through all the methods whose name starts with "test"?

alexis-cousein:
BTW, how come in your example you set up three server dyn resources, but the assert only finds two children? Ah yes, one got killed by the original code since it timed out.

alexis-cousein:
Awaiting requested review now...

vstumpf:
> BTW, how does pbs_benchpress know what class methods to call to do the tests? Does it go through all the methods whose name starts with "test"?

Yeah, it runs all the functions that start with `test_`, you can (try to, I don't fully understand it) find it in `test/fw/ptl/utils/plugins/ptl_test_loader.py`

> BTW, how come in your example you set up three server dyn resources, but the assert only finds two children? Ah yes, one got killed by the original code since it timed out.

Yeah, I did it on purpose to make sure to check _all_ the possible processes.

vstumpf (approved):
LGTM

src/scheduler/server_info.cpp
```diff
@@ -750,21 +751,10 @@ query_server_dyn_res(server_info *sinfo)
...
-							kill(-pid, SIGKILL);
-							waitpid(pid, NULL, 0);
-						}
-					}
-					pid = 0; /* don't kill it again later */
-				}
-				/* Parent; assume fdopen can't fail, but only open if child was not killed. */
-				k = 0;
-				if (pid > 0) {
+				if ( pid > 0 && ret > 0 ) {
```
bhroam, line 754: Remove the space after ( and before )

bhroam (approved):
Just a nit.  I'll sign off.

src/scheduler/server_info.cpp
alexis-cousein: Done (although personally I find that convention jarring when you have operators that _are_ separated from their arguments with spaces, which gives a counterintuitive visual grouping, it is what it is).

alexis-cousein: In another life, someone insisted on the if( blah >= 0 ) style...so I'm not religious about these things ;-).

PR #2112, opened 2020-11-06
Making Scheduler's job related APIs work on the right server as jobs …

suresh-thelkar:
…are sharded in Multi-Server

#### Describe Bug or Feature

In a Multi-Server world as we are sharding jobs, Scheduler should run the jobs in the right server from which they are originated. In order to do this Scheduler should act on the right sd/socket of the server for which the jobs are submitted. In other words Scheduler should call its APIs like run_job, send_run_job, update_job_attr etc on the right sever's sd i.e. it should not use cluster sd for this purpose. 

**Design Document**: https://openpbs.atlassian.net/wiki/spaces/PD/pages/2233991174/Introducing+an+attribute+server+instance+id+for+sharded+objects
**Forum Discussion**:  https://community.openpbs.org/t/adding-a-new-member-to-batch-status-structure/2353/41

For more details on Multi-Server feature please refer to the following.

https://openpbs.atlassian.net/wiki/spaces/PD/pages/1507164161/PBS+Pro+multi-server+project+architecture+and+design

#### Describe Your Change

- Populate Scheduler's internal data structures like job/node objects with the right server sd. So when scheduler accesses the job object it can directly get its server sd easily. 
  The population of the server sds to the job/node objects is done inside IFL layer i.e. while aggregating jobs/nodes itself an extra memeber by the name "server_sock" is added to the batch status. This way by the time Scheduler queries jobs/nodes this info is handy. Scheduler upon querying jobs/nodes it stores "server_sock" in its internal data structures i.e. job_info/node_info objects contain server_sock now.
  
    For example we can make use of this info while calling run_job API inside scheudler
    pbsrc = run_job(resresv->job->server_sock, rr, execvnode, sinfo->has_runjob_hook, err);

-  Scheduler's APIs like run_job, send_attr_updates, pbs_sigjob etc need to act on the right server sd. This PR concentrates on most of the job related APIs like these to act on right server sd so that we can actually see multiple servers functioning in action. There are a few more APIs like pbs_defschreply and handling qrun requests which might also need to act upon right server sd which will be taken up in the next PR. The idea here is other multi-server features like "Scheduler preferring local nodes first, move jobs from a source server to destination server in case if the source server do not have nodes" can benefit out of this PR and they can be worked on immediately.

- Please also note that pbs_asyalterjob is not made multi-server compatible as part of this PR. This is because we have to have special treatment for this like pbs_preempt_jobs() IFL when calling from Scheduler which we are taking care of as part of upcoming Multi-Server PRs. This does not cause any problem as of now for the following reasons.

   - No impact on server functionality
   - As we are making all IFLs work in case of Multi-Server even when they are called from a non scheduler like client/s. So when an IFL is called on virtual fd, the API contacts server instances in some random fashion and get to know the right server for which it sends the request. These changes are already happening as part of the PR #2086

Attached are the detailed test logs for Multi-Server.
[SchedJobRelatedAPIS.txt.log](https://github.com/openpbs/openpbs/files/5500242/SchedJobRelatedAPIS.txt.log)

Following are the regression results.
Profile: regression
|--|--|--|--|--|--|--|
|4692|3845|2|0|0|0|3843|

The test cases failed are known issues and happening in master also.


suresh-thelkar:
I am in the process of running valgrind on these code changes. Meanwhile requesting the reviews for this PR.

src/lib/Libcmds/batch_status.c
```diff
@@ -126,3 +126,20 @@ bs_find(struct batch_status *bs, const char *name)
...
+/**
+ * @brief
+ *	init_bstat - Initialize batch status
+ *
+ * @param[in] bstat - batch_status struct
+ *
+ * @return 	void
+ */
+void
+init_bstat(struct batch_status *bstat)
```
agrawalravi90, line 139: i think we should follow the name scheme in this file and name this 'bs_init' instead

src/lib/Libcmds/batch_status.c
```diff
@@ -126,3 +126,20 @@ bs_find(struct batch_status *bs, const char *name)
...
+{
+	bstat->next = NULL;
```
agrawalravi90, line 141: check for NULL first?

src/scheduler/data_types.h
```diff
@@ -573,6 +573,7 @@ struct job_info
 	unsigned topjob_ineligible:1;	/* Job is ineligible to be a top job */
 
 	char *job_name;			/* job name attribute (qsub -N) */
+	int server_sock;		/* server socket associated with this job */
```
agrawalravi90, line 576: can we add it to resource_resv instead? one less indirection to go through when accessing it, and we might need it for reservations as well.

src/scheduler/pbs_sched.cpp
```diff
@@ -570,6 +570,9 @@ connect_svrpool()
 		svr_conns_secondary =  static_cast<svr_conn_t *>(get_conn_svr_instances(clust_secondary_sock));
 
 		for (i = 0; i < num_conf_svrs; i++) {
+			if (svr_conns_primary == NULL || svr_conns_secondary == NULL)
```
agrawalravi90, line 573: Move this check before the for loop

src/scheduler/pbs_sched.cpp
```diff
@@ -570,6 +570,9 @@ connect_svrpool()
...
+				break;
+				
 			if (svr_conns_primary[i].state == SVR_CONN_STATE_DOWN || svr_conns_primary[i].state == SVR_CONN_STATE_DOWN) {	
 				if (svr_conns_primary[i].state == SVR_CONN_STATE_DOWN)
```
agrawalravi90, line 577: i think there's a bug here, you are checking svr_conns_primary.state twice (line 577 and line 579), one of them should be checking svr_conns_secondary.state right?

agrawalravi90 (changes requested):

bhroam:
I see two things here which are major changes to our IFL interface with no design.  First, you are changing the batch_status.  This is the main structure we pass PBS objects in.  Second, you are changing how the different IFL interfaces use the pbs sockets.  Before you do a pbs_connect() and then use *ANY* IFL interface on that socket.  Now that returns a multi-server socket that can only be used for certain calls.  Now you need to use that multi-server interface to do a stat, and then find the socket you need to use for other calls like alterjob or runjob.  Please write up a design for these changes so they can be discussed on the forums.

agrawalravi90:
> Before you do a pbs_connect() and then use ANY IFL interface on that socket. Now that returns a multi-server socket that can only be used for certain calls. Now you need to use that multi-server interface to do a stat, and then find the socket you need to use for other calls like alterjob or runjob. Please write up a design for these changes so they can be discussed on the forums.

Suresh can explain more, but just to add some clarification, clients can still use the fd returned by pbs_connect directly  without having to care about the fd inside each batch_status .. it's just that if they use the fd that comes within the batch_status, then IFL will not bother trying to figure out which server(s) to send the alterjob or runjob request to, so it'll be more performant.

src/lib/Libcmds/batch_status.c
suresh-thelkar: Renamed this as bs_init.

src/scheduler/data_types.h
suresh-thelkar: I thought of this originally and did not add it for reservations also because they are shared across all the servers and we learn more when we implement them. Due to this I thought it is better add this to job_info object itself.

src/scheduler/pbs_sched.cpp
suresh-thelkar: Yes. Thanks for catching this. Please look into my latest commit for changes.

suresh-thelkar:
> > Before you do a pbs_connect() and then use ANY IFL interface on that socket. Now that returns a multi-server socket that can only be used for certain calls. Now you need to use that multi-server interface to do a stat, and then find the socket you need to use for other calls like alterjob or runjob. Please write up a design for these changes so they can be discussed on the forums.
> 
> Suresh can explain more, but just to add some clarification, clients can still use the fd returned by pbs_connect directly without having to care about the fd inside each batch_status .. it's just that if they use the fd that comes within the batch_status, then IFL will not bother trying to figure out which server(s) to send the alterjob or runjob request to, so it'll be more performant.

Thanks Ravi. There is no change in using the IFLs as far as single server is concerned. It is because the virtual socket  returned by pbs_connect is actually real socket when it comes to single server configuration. 

When it comes to Multi-Server we use virtual socket to fetch node/job related stats. IFLs like pbs_sigjob, pbs_runjob from clients like scheduler are called on the real socket of the corresponding server for performance reasons.  I have created a design document and forum discussion for the same and tried to explain it in detail. Please look into that and provide your feedback.

Design Doc:
https://openpbs.atlassian.net/wiki/spaces/PD/pages/2233991174/Adding+a+new+member+to+batch+status+structure

Forum Discusssion:
https://community.openpbs.org/t/adding-a-new-member-to-batch-status-structure/2353

suresh-thelkar:
@agrawalravi90 , I have implemented your review comments. Please check the latest commit.

@bhroam / @raviv1995 ,  Following are the links to design document and forum discussions. Please go through them and provide your feedback on the following also.
https://openpbs.atlassian.net/wiki/spaces/PD/pages/2233991174/Adding+a+new+member+to+batch+status+structure
https://community.openpbs.org/t/adding-a-new-member-to-batch-status-structure/2353

src/scheduler/data_types.h
agrawalravi90: Well we haven't designed reservations yet, so I think we shouldn't assume that they will be shared, but at least while confirming a reservation, the sched does need to know which server sent it the reservation submit request so that it can respond it to that server, right?

openpbs.spec
```diff
@@ -405,21 +405,21 @@ ldconfig %{_libdir}
 if [ "$1" != "1" ]; then
 	# This is an uninstall, not an upgrade.
 	${RPM_INSTALL_PREFIX:=%{pbs_prefix}}/libexec/pbs_preuninstall server \
-		%{version} ${RPM_INSTALL_PREFIX:=%{pbs_prefix}} %{defined have_systemd}
+		%{version} ${RPM_INSTALL_PREFIX:=%{pbs_prefix}} %{pbs_home} %{defined have_systemd}
```
agrawalravi90, line 408: why were these changes needed?

agrawalravi90 (changes requested):

openpbs.spec
suresh-thelkar: After dev install this file showed up in check-out list along with other files and checked it in without noticing it. We don't need this. I have removed it in the latest commit. Please check.

agrawalravi90 (changes requested):

openpbs.spec
```diff
@@ -1,626 +0,0 @@
-
```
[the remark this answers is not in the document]
hirenvadalia: Its used by Travis (to install build nd runtime dependency), its not safe to delete.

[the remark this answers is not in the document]
suresh-thelkar: I was thinking pbspro.spec.in is enough because psbpro.spec is generated from it automatically. It seems pbspro.spec is also used by Travis. So I keep both of them in my next commit.

bhroam:
I've WIP'd this PR until the design discussion on the forum is complete (which it almost is)

src/scheduler/data_types.h
```diff
@@ -738,6 +739,7 @@ struct node_info
 	int bucket_ind;			/* index in server's bucket array */
 	int node_ind;			/* node's index into sinfo->unordered_nodes */
 	node_partition **np_arr;	/* array of node partitions node is in */
+	int server_sock;		/* server socket associated with this node */
```
bhroam, line 742: We might not need the node socket (and we don't need it right now).  We always talk to the job's server.  It might be true the job might move to another server, but the scheduler talks to the server which owns the job now.  We do need some way to match server to node, but that might be done via name and not socket.

suresh-thelkar: Yes @bhroam. As per the updated design doc we don't need server_sock. We are actually getting server's fd by using server_instance_id which actually consists of "server_instance_name:port". I request you to go through the updated design document.

suresh-thelkar:
> I've WIP'd this PR until the design discussion on the forum is complete (which it almost is)

I have now modified the design document to reflect the approach where we store server instance name as part of server_instance_id due to the reasons that I have mentioned in the forum. So yes I hope the design document is complete now unless there are any further questions. 

Marking this PR as WIP till I finish and test the code changes as per the updated design document.

src/scheduler/data_types.h
bhroam: Ahh, the design has indeed updated since I last reviewed it.  I'd rename this from server_sock, to svr_inst_id since it is not a socket.

suresh-thelkar:
@bhroam / @agrawalravi90,  As the design is signed off, and we are going with the new approach I have made the code changes accordingly. Had to squash and rebase to resolve conflicts and to make Travis work and run smoothly. It helps if you could review all the changes since beginning and provide your comments.

Also attached are the latest detailed test logs for Multi-Server.
[SchedJobRelatedAPIS.txt.log](https://github.com/openpbs/openpbs/files/5595994/SchedJobRelatedAPIS.txt.log)

suresh-thelkar:
It seems Travis builds are not working at the moment due to some issue. I have run all the tests we usually run in Travis. All of them ran successfully.

src/include/svrfunc.h
```diff
@@ -78,6 +78,7 @@ extern char *get_hostPart(char *);
 extern int is_compose(int, int);
 extern int is_compose_cmd(int, int, char **);
 extern char *get_servername(unsigned int *);
+extern char *get_svr_inst_id();
```
agrawalravi90, line 81: void the argument

src/lib/Libattr/master_job_attr_def.xml
```diff
@@ -2077,6 +2077,23 @@
...
+   <attributes>
+      <member_index>JOB_ATR_server_inst_id</member_index>
+      <member_name>ATTR_server_inst_id</member_name>
+      <member_at_decode>decode_str</member_at_decode>
+      <member_at_encode>encode_str</member_at_encode>
+      <member_at_set>set_str</member_at_set>
+      <member_at_comp>comp_str</member_at_comp>
+      <member_at_free>free_str</member_at_free>
+      <member_at_action>NULL_FUNC</member_at_action>
+      <member_at_flags>PRIV_READ | ATR_DFLAG_SSET | ATR_DFLAG_NOSAVM</member_at_flags>
```
agrawalravi90, line 2089: Why 'ATR_DFLAG_NOSAVM' ? if you retrieve a job from db, won't this information be useful ?

src/scheduler/job_info.cpp
```diff
@@ -1234,6 +1236,8 @@ query_job(struct batch_status *job, server_info *sinfo, schd_error *err)
 			else
 				resresv->qrank = -1;
 		}
+		else if (!strcmp(attrp->name, ATTR_server_inst_id))
+			resresv->job->svr_inst_id = string_dup(attrp->value);
```
agrawalravi90, line 1240: do a NULL check here as well?

src/server/process_request.c
```diff
@@ -1654,3 +1654,37 @@ get_servername(unsigned int *port)
...
+	char svr_inst_name[PBS_MAXHOSTNAME + 1];
+	unsigned int svr_inst_port;
+	char *svr_inst_id = NULL;
+
+	/* server_instance_id is of the form "server_inst_name:svr_inst_port". 2 for  ":"" and NULL char */
+	svr_inst_id = malloc(PBS_MAXHOSTNAME + PBS_MAXPORTNUM + 2);
+	if (svr_inst_id == NULL)
+		return NULL;
+
+	if (gethostname(svr_inst_name, PBS_MAXHOSTNAME) == 0)
```
agrawalravi90, line 1678: the svr_inst_name should match its PSI value. I think when you use gethostname(), you will get the same svr_inst_name for all servers running on that host, even though in PSI the the server name values might be different. So, I think we should use PBS_SERVER_NAME value instead of the hostname here, which is what PSI uses.

agrawalravi90 (changes requested):

src/scheduler/job_info.cpp
bhroam: No need to do a NULL check to just print a message.  That happens inside string_dup().  Although error recovery here is kind of important.

bhroam (changes requested):

src/lib/Libattr/master_job_attr_def.xml
suresh-thelkar: Yes, there is a possibility that the server instance name can be changed before it starts. This is the reason I have marked this attribute as ATR_DFLAG_NOSAVM. This is my understanding when I looked at the following code(pbsd_init_job function) with respect to the single server.

We are actually freeing the attribute JOB_ATR_at_server and every time populating its value. So Ideally this could have also been marked as ATR_DFLAG_NOSAVM.


```
	/* update at_server attribute in case name changed */
	free_jattr(pjob, JOB_ATR_at_server);
	set_jattr_generic(pjob, JOB_ATR_at_server, server_name, NULL, SET)
```
@subhasisb, please let us know your opinion on this.

src/server/process_request.c
suresh-thelkar: When it comes to the context of Multi-Server, PBS_SERVER refers to cluster name. This may or may not be a hostname of one of the server instances. It could totally be different from a hostname. 

Yes the idea is svr_inst_name should match with one of the PSI values. We also compare port numbers when all the server instances run on a same machine. So even if gethostname()/get_fullhostname() are the same in the above case, we also compare its port numbers, so it is not a problem. Also, Admin when configuring PBS_SERVER_INSTANCES he should make sure that he always either uses short hostname or FQDN to refer to a server instance.  

In case if we are using PBS_SERVER while populating pbs_conf.psi while only port number is provided by Admin for a server instance, we should change that to actually use the respective server's instance name. We are not supposed to use PBS_SERVER name here.

@subhasisb, please correct me If I am wrong.

hirenvadalia:

src/server/process_request.c
agrawalravi90: ok, can you please document this? That PSI should have valid hostnames, not PBS_SERVER (if it isn't documented already). Also, we should probably add a check inside pbs_loadconf() which verifies that each PSI value is a valid hostname.

agrawalravi90: This also raises questions about how we'll handle talking to multiple servers in some of the IFL calls. Right now, we are in the process of adding multi-server support to IFLs like pbs_deljob, pbs_sigjob etc., which just take a jobid, which won't help identify the server if PSI will not be created using PBS_SERVER. So, if we do what you are proposing, then we'll need a different way to map jobids to servers.

agrawalravi90: Actually, with this restriction, all servers will generate the same jobids, there will be job id conflicts, regardless of whether the servers run on the same host or on different hosts. I think your change cannot go in without fixing the jobid namespace conflict. Please let me know if I'm misunderstanding the implications of this.

suresh-thelkar: > ok, can you please document this? That PSI should have valid hostnames, not PBS_SERVER (if it isn't documented already). Also, we should probably add a check inside pbs_loadconf() which verifies that each PSI value is a valid hostname.

Yes we have not documented the above when PBS_SERVER_INSTANCES are used. Yes  I am planning to do exactly same thing i.e. add code changes to make sure PSI is a valid hostname. Also planning to document it. Don't know why we missed it earlier.

suresh-thelkar: > This also raises questions about how we'll handle talking to multiple servers in some of the IFL calls. Right now, we are in the process of adding multi-server support to IFLs like pbs_deljob, pbs_sigjob etc., which just take a jobid, which won't help identify the server if PSI will not be created using PBS_SERVER. So, if we do what you are proposing, then we'll need a different way to map jobids to servers.

Depending on jobid's hostname to figure out which is its server may or may not be the right idea. It is because going forward PBS_SERVER may or may not be hostname of one of the server instances configured. Instead, we can do the following.
An IFL API if it wants to find out which server is responsible for a given jobid, it tries to contact server instances in some random fashion and finds the right server. If the request reaches to a server which does not own the job it rejects with "unknown job id code". One of the server instances respond and process the request successfully. This way that IFL identifies which server it has to talk to. This is what @nithinj  is doing as part of his PR https://github.com/openpbs/openpbs/pull/2086

suresh-thelkar: > Actually, with this restriction, all servers will generate the same jobids, there will be job id conflicts, regardless of whether the servers run on the same host or on different hosts. I think your change cannot go in without fixing the jobid namespace conflict. Please let me know if I'm misunderstanding the implications of this.

The restriction to have same value for PBS_SERVER is not enforced in the code as of now. We need to keep this assumption in our minds and code it accordingly. So we are not using its(PBS_SERVER) value while configuring server instances and the code in this PR is exactly doing that. So given below is a mechanism how we can test as of now. 

If we make servers run on different complexes then we will get jobids differently because PBS_SERVER value is different. For example, we can get jobids like **1.spark,1.spark2, 2.spark, 2.spark2, 3.spark, 3.spark2**
In one of my test runs(as part of my PR testing) I have got the following
3.spark         root     workq    STDIN       50965   1   1    --    --  R 00:11
    --
3.spark2        root     workq    STDIN       73131   1   1    --    --  R 00:00
   Job run at Wed Nov 25 at 08:51 on (spark2:ncpus=1)
 
 
Also, when all server instances run on the same host, I think we can make aliases for PBS_SERVER in /etc/hosts and use it I believe,  in which case we get the jobids similar to the above.
 
Once we solve "jobid namespace problem" then we can strictly use only one value for PBS_SERVER across all complexes or across all server instances. In fact, we can enforce this in the code. When we do this we don't need to change any of the current code that is added as part of this PR.  This way I believe we don't need to wait till the completion of "jobid namespace feature".

suresh-thelkar:
@bhroam / @agrawalravi90, I have addressed all your comments. Please take a look.

Also, attached are Multi-Server testing logs after implementing the review comments. All test cases passed.
[SchedJobRelatedAPIS_Latest.txt](https://github.com/openpbs/openpbs/files/5607549/SchedJobRelatedAPIS_Latest.txt)

suresh-thelkar:
Attached are TH3 regression logs after implementing the review comments.

Description: Rerun All Tests From #5022
Profile: regression
|--|--|--|--|--|--|--|
|5059|3843|1|0|0|0|3842|

It seems one test case that failed is happening in master also.

suresh-thelkar:
@bhroam / @agrawalravi90, I have also added the new attribute the pbs_ifl_mock.py which is needed for few PTL tests. Please take a look at this commit as well.

suresh-thelkar:
Latest full regression logs as of today.
Description: Rerun All Tests From #5073
Profile: regression
|--|--|--|--|--|--|--|
|5076|3843|3|0|0|0|3840|

Test cases failed are also observed in nightly.

src/scheduler/job_info.cpp
```diff
@@ -125,6 +125,7 @@
 #include "server_info.h"
 #include "attribute.h"
 #include "multi_threading.h"
+#include "libpbs.h"
```
agrawalravi90, line 128: why did you include this file?

src/server/process_request.c
agrawalravi90: > An IFL API if it wants to find out which server is responsible for a given jobid, it tries to contact server instances in some random fashion and finds the right server. If the request reaches to a server which does not own the job it rejects with "unknown job id error code". One of the server instances(which is the owing server instance for the job) respond and process the request successfully. This way the IFL identifies which server it has to talk to. This is what @nithinj is doing as part of his PR #2086

Well that's not ideal for performance (both client and server side). Think about qdel list which was added recently .. we can't just broadcast the entire list of jobs to all servers, that will slow both client and server sides.

agrawalravi90 (changes requested):

src/scheduler/job_info.cpp
suresh-thelkar: As per the latest changes this header file libpbs.h inclusion is not needed. I have removed it in the latest commit. Please check.

src/server/process_request.c
suresh-thelkar: The approach that I mentioned is being followed in the PR #2086 for most of the IFL APIs in general when they are called from a non scheduler like client/s. You can go through that PR for more details. 

When it comes to IFLs like pbs_deljob where performance matters we cannot follow the same approach. Instead, we can go through the following method to achieve the desired goal. There might be other better ways. Following are just my thoughts.
- Do a stat of the required jobs inside the IFL __pbs_deljoblist. Group the jobs by server_instance_id.
- Iterate over all the resultant server instances and pass one job-list at a time to the function PBSD_deljoblist().
- As the number of server instances configured are minimum this greatly enhances the performance.

Also, we are already following a slightly different approach for some IFLs like pbs_runjob for example when qrun with '-H' option is called. We actually do a stat for the given node through which we can find the right server instance id and send the request to that server instance directly instead of contacting servers in some random fashion and finding out which is the right server instance.

suresh-thelkar:
Given below are very lates regression results. 

Profile: regression
|--|--|--|--|--|--|--|
|5082|3843|1|0|0|0|3842|

One test case that failed is also observed in nightly.

suresh-thelkar:
@agrawalravi90, I have addressed your latest review comments also. Please take a look.

src/scheduler/fifo.cpp
```diff
@@ -1646,7 +1646,7 @@ run_update_resresv(status *policy, int pbs_sd, server_info *sinfo,
 					fflush(stdout);
 #endif /* localmod 031 */
 
-					pbsrc = run_job(get_svr_inst_fd(pbs_sd, resresv->job->svr_inst_id), rr, execvnode, sinfo->has_runjob_hook, err);
+					pbsrc = run_job(get_svr_inst_fd(pbs_sd, rr->job->svr_inst_id), rr, execvnode, sinfo->has_runjob_hook, err);
```
bhroam, line 1649: Please have a shortcut in get_svr_inst_fd() if there is only 1 server.  No need to search by string if it is just one.

src/scheduler/job_info.cpp
```diff
@@ -1778,7 +1786,7 @@ update_job_attr(int pbs_sd, resource_resv *resresv, const char *attr_name,
 
 	if (pattr != NULL && (flags & UPDATE_NOW)) {
 		int rc;
-		rc = send_attr_updates(get_svr_inst_fd(pbs_sd, resresv->job->svr_inst_id), resresv->name, pattr);
+		rc = send_attr_updates(pbs_sd, resresv->name, pattr);
```
bhroam: Don't you want to send the alter to the specific server?

bhroam (changes requested):

src/server/process_request.c
bhroam: Wait, I thought a runjob always went to the server currently owning the job.  The owning server is responsible to move and run the job.

agrawalravi90: ya, we always send it to the job owner, I think Suresh is mistaken, but @nithinj can confirm once.

src/scheduler/fifo.cpp
suresh-thelkar: Optimized get_svr_inst_fd() to return early in case of single server mode. Please check the latest commit. 

Please note that in case of single server we cannot return vfd itself. It is because it may not be the first fd of the server instances after the PR #2086.

src/scheduler/job_info.cpp
suresh-thelkar: Yes, But I am not modifying pbs_asyalterjob as part of this PR as we have to handle this differently. I added a note about the same in the conversations section(last bullet point). Please check. We can take this after this PR.

suresh-thelkar:
@bhroam, I have addressed your comments. Please take a look.

src/scheduler/fifo.cpp
nithinj: just to clarify, vfd and real fd will be the same in the case of a single server cluster. Returning the first fd also works.

agrawalravi90: that's how it is right now, but aren't we planning to change it so that vfd is not the first real fd? To be able to distinguish between real and virtual fd in IFL ?

src/scheduler/job_info.cpp
agrawalravi90: you don't have to modify pbs_asyalterjob Suresh if you just send it the right fd, right?

agrawalravi90: without this, i don't see how this code will work for multi-server, it'll always send alterjob to the first server if we sent it the virtual fd

bhroam (approved):
Looks good.  Please fix the conflict.

src/scheduler/fifo.cpp
nithinj: I was talking about a single server cluster. For multi-server cluster vfd will be different than that of any instance fd.

src/scheduler/job_info.cpp
nithinj: We can take it up as part of #2086 once this gets merged.

src/scheduler/fifo.cpp
agrawalravi90: ah ok, makes sense

agrawalravi90:
> Looks good. Please fix the conflict.

Since Suresh is out on vacation and this is kind of urgent, and the conflict has to do with my master branch changes, I'll take it up

bhroam:
@agrawalravi90 do you sign off on the PR?  You are the other reviewer.  If you are happy with it, you should sign off before you work on it.

agrawalravi90 (approved):

src/server/process_request.c
nithinj: sorry, I missed this thread earlier.
qrun sends the request to the job owning server. IFL will mention where the mother superior resides by doing a node_stat. It is a special case as we are dealing with two shared objects instead of one.

PR #2122, opened 2020-11-13
Adding mock run to scheduler

agrawalravi90:
#### Describe Bug or Feature

Often, we have a need to isolate performance bottlenecks in PBS, and it’s not always straight-forward. Is it the server? scheduler? mom? is it IO? To rule out scheduler as the bottleneck and focus on testing performance of the server(s), it is desirable to add an option to scheduler which will make it very lean so that it does the bare minimum needed to schedule jobs.

#### Describe Your Change

Adding new option to pbs_sched called -m to make it run in "mock run" mode. See design doc for more info.

#### Link to Design Doc

https://openpbs.atlassian.net/wiki/spaces/PD/pages/2257944577/Mock+Run+option+for+Scheduler

**Setup**
50k vnodes, 10 ncpus each, 50k jobs, each requesting 1 ncpu, pbs_mom running in mock run mode.

**Regular Scheduling**
Time taken to schedule 50k jobs = **88 seconds**

**Mock run scheduling**
Time taken to schedule 50k jobs = **3.5 seconds (out of which ~3 seconds was time taken to query the universe)**

### DIFF
--- a/src/scheduler/Makefile.am
+++ b/src/scheduler/Makefile.am
@@ -87,6 +87,8 @@ libpbs_sched_a_SOURCES = \
 	limits.cpp \
 	misc.cpp \
 	misc.h \
+	mock_run.cpp \
+	mock_run.h \
 	multi_threading.cpp \
 	multi_threading.h \
 	node_info.cpp \
--- a/src/scheduler/buckets.h
+++ b/src/scheduler/buckets.h
@@ -44,6 +44,8 @@ extern "C" {
 #ifndef _BUCKETS_H
 #define _BUCKETS_H
 
+#include "data_types.h"
+
 /* bucket_bitpool constructor, copy constructor, destructor */
 bucket_bitpool *new_bucket_bitpool();
 void free_bucket_bitpool(bucket_bitpool *bp);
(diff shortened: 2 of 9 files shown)


bhroam:
This seems like a POC of something to fix one specific problem.  You are assuming there are just 1 cpu jobs.  This is just one instance of what a "mock" mode should do.  It should actually run any type of job.  In any case, the whole purpose of pbs_sched.c was a driver to write a scheduler.  Why not create a separate binary which is a mock mode scheduler?  You can run that instead of the normal pbs_sched.

Another option is to just do this outside the scheduler using qrun.  You could write a python scheduler pretty easily that just runs single node jobs using qrun -H.  That way you aren't writing very specific special case code into the scheduler.

agrawalravi90:
Thanks for your inputs Bhroam. I'm not sure I agree with you that this is solving a very specific problem. The requirement was to create a mode which will remove the scheduler from the critical path and allow us to focus on performance problems with the server. So, it should still schedule all the jobs that server asks it to. From server's POV, nothing changes, it still gets as many IFL requests as before. I see this being very similar to the mock run option in pbs_mom, its goal was also to remove mom from the critical path and test performance of server/sched.

"It should actually run any type of job"
It will run any type of job, it will just ignore the resource list that the job provided it and assume that it's only requesting 1 cpu. Is there any value in doing the comprehensive resource matching that the regular scheduler does?

PR #2129, opened 2020-11-18
Adding pbs_sched_bare, a minimal sched for testing

agrawalravi90:
#### Describe Bug or Feature

For performance testing scenarios where the focus is on server/mom/comm performance, it's useful to have a way to remove scheduler from the critical path. This change adds a very minimal sched called 'pbs_sched_bare' which does the following:
- query the universe from server(s)
- go through the array of jobs, assuming that each job requires only 1 cpu and no other resource
- For each job, find the next free node, allocate 1 cpu on it, create exec_vnode and send runjob request to the server

it will be installed under unsupported/ and cannot be run alongside pbs_sched, only one can be run at a time. When it does run, it will use the same PBS_HOME space as pbs_sched.

#### Forum discussion
https://community.openpbs.org/t/wip-mock-run-option-for-scheduler/2356/7

**Setup**
50k vnodes, 10 ncpus each, 50k jobs, each requesting 1 ncpu, pbs_mom running in mock run mode.

**Regular Scheduling**
Time taken to schedule 50k jobs = 88 seconds

**pbs_sched_bare scheduling**
Time taken to schedule 50k jobs = 3.5 seconds (out of which ~3 seconds was time taken to query the universe)


src/scheduler/pbs_sched_bare.cpp
```diff
@@ -0,0 +1,294 @@
...
+ *		cycle.
+ *
+ * @param[in]	sd	-	primary socket descriptor to the server pool
+ *
+ * returns 0
+ *
+ */
+static int
+intermediate_schedule_bare(int sd, const sched_cmd *cmd)
+{
```
bhroam, line 184: You probably don't need intermediate_schedule_bare().  It's to handle rare cases where we want to restart the scheduling cycle.

src/scheduler/pbs_sched_utils.cpp
```diff
@@ -0,0 +1,1204 @@
+/*
```
bhroam, line 1: In my opinion libsched.a is for scheduling policies.  The stuff that was in pbs_sched.c was separate.  Let's put this in its own library.

bhroam (changes requested):

agrawalravi90:
@bhroam thanks for the review, I've addressed your comments, please provide further feedback

bhroam (approved):
I didn't realize the only file in libschedharness would be pbs_sched_utils.cpp.  Do we even want to bother with libschedharness?  Don't include it in libsched like you had it before, just link it with both pbs_sched and pbs_sched_bare directly.  Up to you.  I'll sign off.

agrawalravi90:
@bhroam  i found an issue after i had removed intermediate schedule, please review the last commit as well.

About libschedharness, I think it's ok to keep it, we might break out the single file into multiple when we move to c++, so having it shouldn't hurt

bhroam (approved):

hirenvadalia:

ashwathraop (approved):
LGTM

agrawalravi90:
@ashwathraop thanks for approving. @bhroam i think this can be merged, thanks!

PR #2134, opened 2020-11-24
Changed the way resvs are submitted in the hour dst changes

arungrover:
#### Describe Bug or Feature
This test suite encounters failures when DST changes between the time reservation is submitted and the time reservation is supposed to end.

#### Describe Your Change
Problem is that when PTL submits this reservation it converts the end time epoch and that conversion sometimes results in end time being in the past.
Change is to detect that DST is going to change soon and submit reservations with duration instead of submitting with end time.
#### Link to Design Doc
NA

[test_resv.txt](https://github.com/openpbs/openpbs/files/5586691/test_resv.txt)

### DIFF
--- a/test/tests/functional/pbs_reservations.py
+++ b/test/tests/functional/pbs_reservations.py
@@ -1957,7 +1957,7 @@ class TestReservations(TestFunctional):
         idle_timer = 15
         extra = {'delete_idle_time': idle_timer}
         rid = self.submit_reservation(
-            user=TEST_USER, select='1:ncpus=1', rrule='freq=HOURLY;COUNT=3',
+            user=TEST_USER, select='1:ncpus=1', rrule='freq=DAILY;COUNT=3',
             start=start, end=start + 1800, extra_attrs=extra)
 
         exp_attr = {'reserve_state': (MATCH_RE, "RESV_CONFIRMED|2")}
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_reservations.py
```diff
@@ -1986,7 +1977,10 @@ def test_delete_idle_resv_job_standing(self):
         self.server.expect(JOB, {'job_state': 'R'}, id=jid)
 
         strf_str = '%a %b %d %T %Y'
-        start_str = time.strftime(strf_str, time.localtime(start + 3600))
+        delay = 3600
+        if self.dst_changes(start, start + delay) is True:
+            delay = 7200
+        start_str = time.strftime(strf_str, time.localtime(start + delay))
```
bhroam, line 1983: Is there a good reason we don't just use 7200 always?

test/tests/functional/pbs_reservations.py
```diff
@@ -1931,10 +1920,12 @@ def test_delete_idle_resv_job_delete(self):
...
-        a = {'reserve_start': start, 'reserve_end': now + 3600,
-             'delete_idle_time': 15}
-        r = Reservation(attrs=a)
-        rid = self.server.submit(r)
+        extra = {'delete_idle_time': idle_timer}
+        rid = self.submit_reservation(user=TEST_USER,
+                                      select='1:ncpus=1',
+                                      start=start,
+                                      end=now + 3600,
+                                      extra_attrs=extra)
```
bhroam, line 1928: Why submit as TEST_USER, it is the default.  Unless you are checking for it to be TEST_USER, there is no need.  Here and other places.

bhroam (changes requested):

test/tests/functional/pbs_reservations.py
arungrover: This is most likely because of copy/paste. Since there is no harm in doing this, I'd like to leave it as is.

arungrover:
@bhroam thanks for reviewing the PR. I have addressed your comment by changing the freq of the recurrence rule to daily instead of hourly. Increasing the duration of the reservation was causing issues because the freq was hourly.

bhroam (approved):
LGTM

hirenvadalia:

ashwathraop (approved):
LGTM

PR #2136, opened 2020-11-26
Fix test of Test_passing_environment_variable_via_qsub to run on remote mom

shilpakodli:
#### Describe Bug or Feature
Test of Test_passing_environment_variable_via_qsub was failing with below error when mom is on remote host:
  File "/opt/ptl/tests/functional/pbs_passing_environment_variable.py", line 139, in test_passing_shell_function
    msg="Environment variable foo content does "
AssertionError: 'foo=() {  if [ /bin/true ]; then\n echo hello;\n fi\n}\nhello' != 'BASH_FUNC_foo()=() {  if [ /bin/true ]; then\n echo hello;\n fi\n}\nhello'
- foo=() {  if [ /bin/true ]; then
+ BASH_FUNC_foo()=() {  if [ /bin/true ]; then
? ++++++++++   ++
   echo hello;
   fi
  }
  hello : Environment variable foo content does not match original
Reason: Test was executing a script on local host which is supposed to run on mom node.

#### Describe Your Change
Creating a script on mom node so that when Job is submitted with a script, it will make sure that that script runs on mom node.

Logs with no mom on server:
Profile: regression
|--|--|--|--|--|--|--|
|4998|3|0|0|0|0|3|

Logs with mom on server:
Description: Rerun All Tests From #4998
Profile: regression
|--|--|--|--|--|--|--|
|5019|3|0|0|0|0|3|

### DIFF
--- a/test/tests/functional/pbs_passing_environment_variable.py
+++ b/test/tests/functional/pbs_passing_environment_variable.py
@@ -104,10 +104,10 @@ env | grep foo
 unset -f foo
 exit 0
 """
-        fn = self.du.create_temp_file(body=foo_scr)
-        self.du.chmod(path=fn, mode=0o755)
+        fn = self.du.create_temp_file(hostname=self.mom.hostname, body=foo_scr)
+        self.du.chmod(hostname=self.mom.hostname, path=fn, mode=0o755)
         foo_msg = 'Failed to run foo_scr'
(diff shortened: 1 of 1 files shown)


Mahalty (approved):

shilpakodli:
@openpbs/openpbs-maintainers  Please take this PR forward.

hirenvadalia (approved):

PR #2140, opened 2020-12-01
Fix sleep time of jobs in test_app_provisioning 

sakshamgarg:
#### Describe Bug or Feature

test_app_provisioning fails with "no data for job" error while searching the "R" state of the job in PTL.

#### Describe Your Change

Since the current sleep time in the test is less (1 second), the job's state & substate doesn't update at the server and the job immediately ends. Hence not setting a sleep time and keeping it default.

[test_app_prov_before_fix.txt](https://github.com/openpbs/openpbs/files/5622198/test_app_prov_before_fix.txt)
[test_app_prov_after_fix_updated.txt](https://github.com/openpbs/openpbs/files/5622307/test_app_prov_after_fix_updated.txt)

### DIFF
--- a/test/tests/functional/pbs_provisioning.py
+++ b/test/tests/functional/pbs_provisioning.py
@@ -168,7 +168,6 @@ class TestProvisioningJob(TestFunctional):
         self.server.manager(MGR_CMD_SET, NODE, a, id=self.hostA)
 
         job = Job(TEST_USER1, attrs={ATTR_l: 'aoe=App1'})
-        job.set_sleep_time(3)
         jid = self.server.submit(job)
         self.server.expect(JOB, {'job_state': 'R'}, id=jid)
         self.server.log_match(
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_provisioning.py
```diff
@@ -168,7 +168,7 @@ def test_app_provisioning(self):
         self.server.manager(MGR_CMD_SET, NODE, a, id=self.hostA)
 
         job = Job(TEST_USER1, attrs={ATTR_l: 'aoe=App1'})
-        job.set_sleep_time(1)
+        job.set_sleep_time(3)
```
Shrini-h, line 171: May be its better to remove this whole line, and let the sleep duration get set to the default 60 seconds.

Shrini-h (changes requested):

Shrini-h (approved):
LGTM, Thanks for making the changes

npadole20 (approved):

hirenvadalia (approved):

PR #2142, opened 2020-12-03
Initial support for MIG-enabled GPUs

vstumpf:
#### Describe Bug or Feature
Support MIG on machines like the NVIDIA DGX-A100.

#### Describe Your Change
Add support for MIG.
Report GIs as ngpus.
Assign GIs and CIs to jobs that request them.

#### Link to Design Doc
https://openpbs.atlassian.net/wiki/spaces/PD/pages/2313453569/Nvidia+MIG+Support

[gpu_cgroups.txt](https://github.com/openpbs/openpbs/files/5669241/gpu_cgroups.txt)

### DIFF
--- a/src/hooks/cgroups/pbs_cgroups.PY
+++ b/src/hooks/cgroups/pbs_cgroups.PY
@@ -1885,9 +1885,9 @@ class NodeUtils(object):
                     if mig_mode is not None:
                         current_mig = mig_mode.find('current_mig')
                         if current_mig is not None:
-                            mig_enabled = current_mig.text == 'Enabled'
-                            if mig_enabled:
+                            if current_mig.text == 'Enabled':
                                 mig_found = True
+                                mig_enabled = True
                     gpus[name] = {
                         'pci_bus_id': (domain + ':' + instance).lower(),
                         'uuid': child.find('uuid').text,
@@ -1975,12 +1975,18 @@ class NodeUtils(object):
                        '%s: nvidia-smi call took %f seconds' %
                        (caller_name(), elapsed_time))
 
-        # format is
+        # format is either
         # | Compute instances:                                    |
         # | GPU     GPU       Name             Profile   Instance |
         # |       Instance                       ID        ID     |
         # |         ID                                            |
         # |   0      7       MIG 1g.5gb           0         0     |
+        # or
+        # | Compute instances:                                             |
+        # | GPU     GPU       Name         Profile   Instance   Placement  |
+        # |       Instance                   ID        ID       Start:Size |
+        # |         ID                                                     |
+        # |   0      7       MIG 1g.5gb       0         0          0:1     |
         r = re.compile(
             r'^\|\s+(\d+)\s+(\d+)\s+MIG\s+\S+\s+\d+\s+(\d+)\s+(\S+\s+)?\|$')
         for l in out:
@@ -2000,7 +2006,6 @@ class NodeUtils(object):
             if 'cis' not in gpus[gpuid]['gis'][giid]:
                 gpus[gpuid]['gis'][giid]['cis'] = {}
             gpus[gpuid]['gis'][giid]['cis'][ciid] = ci
-        return
 
     def _discover_mig_minor(self, gpu, gi, ci=None):
         pbs.logmsg(pbs.EVENT_DEBUG4, '%s: Method called' % caller_name())
@@ -3811,8 +3816,7 @@ class CgroupUtils(object):
                 for instance in devices[devclass]:
                     dev = devices[devclass][instance]
                     if 'extra_devs' in dev:
-                        for extra in dev['extra_devs']:
-                            accelerators.append('%s' % extra)
+                        accelerators.extend(dev['extra_devs'])
                     accelerators.append('%d:%d' % (dev['major'], dev['minor']))
         # For CentOS 7 we need to remove a *:* rwm from devices.list
         # before we can add anything to devices.allow. Otherwise our
--- a/src/hooks/cgroups/pbs_cgroups.PY
+++ b/src/hooks/cgroups/pbs_cgroups.PY
@@ -1079,7 +1079,10 @@ class HookUtils(object):
                     if key.startswith('mic'):
                         mics.append(key[3:])
                     elif key.startswith('nvidia'):
-                        gpus.append(node.devices['gpu'][key]['uuid'])
+                        if 'uuid' in node.devices['gpu'][key]:
+                            gpus.append(node.devices['gpu'][key]['uuid'])
+                        if 'uuids' in node.devices['gpu'][key]:
+                            gpus.extend(node.devices['gpu'][key]['uuids'])
                 if mics:
                     env_list.append('OFFLOAD_DEVICES=%s' %
                                     ",".join(mics))
@@ -1324,7 +1327,10 @@ class HookUtils(object):
                     if key.startswith('mic'):
                         mics.append(key[3:])
                     elif key.startswith('nvidia'):
-                        gpus.append(node.devices['gpu'][key]['uuid'])
+                        if 'uuid' in node.devices['gpu'][key]:
+                            gpus.append(node.devices['gpu'][key]['uuid'])
+                        if 'uuids' in node.devices['gpu'][key]:
+                            gpus.extend(node.devices['gpu'][key]['uuids'])
                 if mics:
                     env_list.append('OFFLOAD_DEVICES=%s' %
                                     ",".join(mics))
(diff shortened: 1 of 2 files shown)


vstumpf:
This feature works in the following way:
0) In `NodeUtils._discover_devices()`, we call `NodeUtils._discover_gpus()`
1) In `_discover_gpus()`, when parsing `nvidia-smi` output to get the dict of physical gpus, we also note if it has mig enabled.
2) If any gpus have mig enabled, we call `NodeUtils._discover_migs()`
3) `_discover_migs()` runs and parses the output of `nvidia-smi mig` to get the GIs of the gpus and the CIs of each GI. It then adds the GIs to the dictionary of physical gpus.
4) We're back in `_discover_devices()`, where we replace the physical gpu in the dict of gpus with the GIs.

The other changes (uuids, extra_devs) are for the fact that the GIs require extra device numbers to add to the cgroup device allowlist, as well as the UUID for CUDA_VISIBLE_DEVICES

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -1820,15 +1880,145 @@ class NodeUtils(object):
...
                         raise ValueError('GPU ID not recognized: ' + bus_id)
                     name = 'nvidia%s' % child.find('minor_number').text
+                    mig_enabled = False
+                    mig_mode = child.find('mig_mode')
+                    if mig_mode is not None:
+                        current_mig = mig_mode.find('current_mig')
+                        if current_mig is not None:
+                            mig_enabled = current_mig.text == 'Enabled'
+                            if mig_enabled:
+                                mig_found = True
```
bhroam, line 1890: Maybe do the following:
```
if current_mig.text == 'Enabled':
    mig_found = True
    mig_enabled = True
```

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -1820,15 +1880,145 @@ class NodeUtils(object):
...
+                       (caller_name(), elapsed_time))
+
+        # format is
+        # | Compute instances:                                    |
+        # | GPU     GPU       Name             Profile   Instance |
+        # |       Instance                       ID        ID     |
+        # |         ID                                            |
+        # |   0      7       MIG 1g.5gb           0         0     |
+        r = re.compile(
+            r'^\|\s+(\d+)\s+(\d+)\s+MIG\s+\S+\s+\d+\s+(\d+)\s+(\S+\s+)?\|$')
```
bhroam, line 1985: So the regex includes (\S\s+)? which the example output does not contain.  It will match because the ? means 0 or 1 of them, so 0 of them matches.  What is it for?

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -1820,15 +1880,145 @@ class NodeUtils(object):
...
+            minor = self._discover_mig_minor(gpu_num, giid, ciid)
+            if minor == -1:
+                continue
+            uuid = 'MIG-%s/%s/%s' % (gpus[gpuid]['uuid'], giid, ciid)
+            ci = {'minor': minor, 'gi': giid, 'ci': ciid, 'uuid': uuid}
+            pbs.logmsg(pbs.EVENT_DEBUG4, 'CI found %s' % str(ci))
+            if 'cis' not in gpus[gpuid]['gis'][giid]:
+                gpus[gpuid]['gis'][giid]['cis'] = {}
+            gpus[gpuid]['gis'][giid]['cis'][ciid] = ci
+        return
```
bhroam, line 2003: This isn't necessary, you just fall out of the function at this point.

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -1820,15 +1880,145 @@ class NodeUtils(object):
...
+        pbs.logmsg(pbs.EVENT_DEBUG4, '%s: Method called' % caller_name())
+        path = os.path.join(os.sep, 'proc', 'driver',
+                            'nvidia-caps', 'mig-minors')
+        if ci is None:
+            capname = 'gpu%d/gi%d/access' % (gpu, gi)
+        else:
+            capname = 'gpu%d/gi%d/ci%d/access' % (gpu, gi, ci)
+        with open(path, 'r') as minors:
+            for line in minors:
+                cap, minor = line.split()
```
bhroam, line 2015: If the format of the proc file changes and you get back more than 2 columns, this will barf.  How about line.split(' ', 1)?

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -3617,6 +3810,9 @@ class CgroupUtils(object):
             if devclass == 'mic' or devclass == 'gpu':
                 for instance in devices[devclass]:
                     dev = devices[devclass][instance]
+                    if 'extra_devs' in dev:
+                        for extra in dev['extra_devs']:
+                            accelerators.append('%s' % extra)
```
bhroam, line 3815: Any reason you didn't just append extra?  or str(extra)?

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -500,6 +500,16 @@ def setUp(self):
     devices_job="$devices_base/sbp.service/jobid/$PBS_JOBID"
 else
     devices_job="$devices_base/sbp.slice/sbp-${jobnum}.*.slice"
+fi
+
+device_list=`cat $devices_job/devices.list`
+grep "195" $devices_job/devices.list
```
bhroam, line 506: Couple of questions.  First do we want to hard code this?  Then you don't assign this grep into anything.  It'll just be printed to stdout.  Is that what you want?

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -500,6 +500,16 @@ def setUp(self):
...
+
+ngpus=$(nvidia-smi -L | grep "MIG-GPU" | wc -l)
+if [ "$ngpus" -eq "0" ]; then
+    ngpus=$(nvidia-smi -L | grep "GPU" | wc -l)
```
bhroam, line 510: If the call to nvidia-smi -L takes a long time, you could assign the value into a temp file or just a variable so you only need to call it once.

bhroam (changes requested):

src/hooks/cgroups/pbs_cgroups.PY
vstumpf: A newer version of nvidia-smi adds a whole new column, meaning I need to support both.
Check the design here: https://openpbs.atlassian.net/wiki/spaces/PD/pages/2313453569/Nvidia+MIG+Support#External-Dependencies

bhroam: Your comment about format is confusing then.  I'd add a couple more lines to the comment explaining the two formats.

src/hooks/cgroups/pbs_cgroups.PY
vstumpf: This is a documented, supported interface by nvidia, it shouldn't change without notice. Because of this, I don't think we need to change it.

test/tests/functional/pbs_cgroups_hook.py
vstumpf: 195 is the major number for nvidia devices, so we can hardcode this. We just want to print it out to stdout, the test will search for the output.

test/tests/functional/pbs_cgroups_hook.py
vstumpf: It shouldn't take longer than 5 seconds. I don't want to output it to a file, then i'd have to clean it up afterwards.

src/hooks/cgroups/pbs_cgroups.PY
bhroam: Well just incase it does change, I'd rather us do split(' ', 1) because then we'll fail to query the MIG and not raise an exception.  We'll have to change the hook in either case, but at least this way it won't stop working.

src/hooks/cgroups/pbs_cgroups.PY
vstumpf: I can actually just use extend.

bhroam (approved):
LGTM

bayucan (commented):
@vstumpf : This looks fine. Please run the entire pbs_cgroups_hook testsuite to make sure nothing is affected by this implementation...

vstumpf:
[full_cgroups.txt](https://github.com/openpbs/openpbs/files/5669871/full_cgroups.txt)
Here's a full run of the suite, but only ran on 1 node. Hopefully this is enough evidence.

bayucan (approved):
full cgroups hook testing on 1 node looks good!

bayucan:
Merged #2142 into master.

PR #2143, opened 2020-12-03
Preemption fails when host or vnode is requested at chunk level

arungrover:
#### Describe Bug or Feature
PBS scheduler while trying to preempt a job tries to limit the number of preemption candidates by checking the reason why a high priority job could not run and then cull the preemptible jobs to only those jobs who can help in running the high priority job.
While doing so, it takes special care about non-consumable resource vnode and host. The issue while checking these resources was that it was assumed these two are job wide resources and thus they are requested for the whole job. But this assumption is incorrect, one can request a multinode job to have one chunk running on a specific vnode or host.
Scheduler was crashing because of this wrong assumption.

#### Describe Your Change
In this change, we try to retain the value of vnode/host because of which the job could not run in the error structure. We store it in arg2 of error structure. Then in cull_preemptible_jobs, instead of searching though all the chunks and finding out which chunk requested the resource that caused the job to not run, we now use the arg2 member of error structure that stores the name of the vnode/host because of which the job could not run.

#### Link to Design Doc
NA

[test_preemption_before.txt](https://github.com/openpbs/openpbs/files/5633185/test_preemption_before.txt)
[test_preemption.txt](https://github.com/openpbs/openpbs/files/5633188/test_preemption.txt)

### DIFF
--- a/src/scheduler/check.cpp
+++ b/src/scheduler/check.cpp
@@ -1200,7 +1200,11 @@ check_avail_resources(schd_resource *reslist, resource_req *reqlist,
 							requested,
 							res_to_str_r(res, RF_AVAIL, resbuf2, sizeof(resbuf2)));
 						set_schd_error_arg(err, ARG1, buf);
-						set_schd_error_arg(err, ARG2, requested);
+						/* Set arg2 for vnode/host resource. In case of preemption, arg2 is used to cull
+						 * the list of running jobs
+						 */
+						if ((res->def == getallres(RES_HOST)) || (res->def == getallres(RES_VNODE)))
+							set_schd_error_arg(err, ARG2, requested);
 					}
 				}
 			}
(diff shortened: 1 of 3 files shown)


src/scheduler/check.cpp
```diff
@@ -1192,13 +1192,19 @@ check_avail_resources(schd_resource *reslist, resource_req *reqlist,
 				if (!compare_non_consumable(res, resreq)) {
 					fail = 1;
 					if (err != NULL) {
+						char *requested;
```
vstumpf, line 1195: Can you make this a `const char *`? Eventually I'd like `res_to_str_r` return a `const char *`,

bhroam (approved):
LGTM

vstumpf (changes requested):
LGTM

vstumpf (commented):
LGTM

vstumpf (approved):
LGTM (third time's the charm)

PR #2146, opened 2020-12-03
Restore cgroup hook backward compatibility with 13.X-2020.x

alexis-cousein:
#### Describe Bug or Feature

Recent changes for cgroup hooks break backwards compatibility with 13.x-2020.x for no good reason at all. Part of :

https://github.com/openpbs/openpbs/pull/2105

Since many sites use different versions but want to use the same cgroup hook, developers should, in my opinion, refrain to do so unless it's really needed, to avoid to fork the hook, which would require new developments to be merged into umpteen versions of the hook.

Moreover, there is NO useful functionality in the original change anyway: it doesn't achieve anything except documenting the existence of new hook events to ignore. But the cgroup hook already ignores events that are not explicitly defined without any issue, so not adding code would have been just as good as far as funcionality is concerned (although I do agree that documenting in code that indeed these events are recognized to not require any action is not bad in se).

#### Describe Your Change

If using pbs.MANAGEMENT or pbs.MODIFYVNODE, guard the section of code with an if hasattr. In fact, there _was_ already code that did that, which would have been hard to miss if the events had been added at the _end_ of the list.

Also rolled in code in add_pids that adds another guard against crashing the hook with an uncaught exception if someone uses pbs_attach on a node using a PID that does not exist (or no longer exists). os.getsid() will return an OSError, and there is no try: construct around that call in the existing hook. 

Triggering the exception was recently demonstrated by a site using PBSPro 2020.

#### Link to Design Doc

N/A

N/A on master, since this is mainly about compatibility of the source with other releases, to avoid having to needlessly fork the hook code.

### DIFF
--- a/src/hooks/cgroups/pbs_cgroups.PY
+++ b/src/hooks/cgroups/pbs_cgroups.PY
@@ -880,6 +880,9 @@ class HookUtils(object):
             self.hook_events = hook_events
         else:
             # Defined in the order they appear in module_pbs_v1.c
+            # if adding new events that may not exist in all PBS versions,
+            # use hasattr() to prevent exceptions when the hook is used
+            # on older PBS versions - see e.g. EXECJOB_ABORT
             self.hook_events = {}
             self.hook_events[pbs.QUEUEJOB] = {
                 'name': 'queuejob',
--- a/src/hooks/cgroups/pbs_cgroups.PY
+++ b/src/hooks/cgroups/pbs_cgroups.PY
@@ -3482,7 +3482,7 @@ class CgroupUtils(object):
                            ' PID %s to jobid %s'
                            % (caller_name(), pidarg, jobid))
             if (sid > 1):
-                pids = self._get_pids_in_sid(os.getsid(pidarg))
+                pids = self._get_pids_in_sid(sid)
         elif isinstance(pidarg, list):
             for pid in pidarg:
                 if not isinstance(pid, int):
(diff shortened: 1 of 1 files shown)


src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -3468,8 +3470,19 @@ class CgroupUtils(object):
...
+            try:
+                sid = os.getsid(pidarg)
+            except OSError as exc:
+                sid = -1
+                pbs.logmsg(pbs.EVENT_DEBUG2,
+                           '%s: Request to attach session of non-existing'
+                           ' PID %s to jobid %s'
+                           % (caller_name(), pidarg, jobid))
+            if (sid > 1):
+                pids = self._get_pids_in_sid(os.getsid(pidarg))
```
bayucan, line 3485: Shouldn't just use 'sid' here since it was already obtained earlier?

alexis-cousein: I almost beat you to it -- I was just baking the second commit ;-).

bayucan (approved):
This looks reasonable.

monkeystate:
Thank you for your efforts to fix this. While implementing the modifyvnode hook feature we sometimes used the modifyjob event implementation as a guide for constructing boilerplate code. If the hasattr() construct is the recommended way to add new events going forward, you might consider adding a comment indicating that. As it is now the only guidance in the block is from 4 years ago ("# Defined in the order they appear in module_pbs_v1.c")

bhroam:
@alexis-cousein could you update the PR title to be a little more specific to what you are fixing?  From your description, you are fixing the cgroups hook to be backwards compatible.

alexis-cousein:
Changed PR title, added comments on where to look for examples when adding events that do not exist on all PBS versions.

vstumpf (approved):
LGTM

PR #2148, opened 2020-12-04
Fix: Interactive job does not stay logged in

raviv1995:
#### Describe Bug or Feature

Interactive jobs don't stay logged in. The session ends abruptly without user intervention. This happens because the background qsub daemon exits one minute after it is created and the session exits with it.

#### Describe Your Change

The do_daemon_stuff function should not be invoked when the job is interactive.

[changelog.log](https://github.com/openpbs/openpbs/files/5641683/changelog.log)
TH_Out

### DIFF
--- a/src/cmds/qsub.c
+++ b/src/cmds/qsub.c
@@ -3193,7 +3193,8 @@ regular_submit(int daemon_up)
 		else
 			rc = -1;
 	}
-	do_daemon_stuff();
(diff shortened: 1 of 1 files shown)


npadole20 (approved):

bayucan (approved):
This looks like in line with the original qsub code of firing fork_and_stay().

bayucan:
Merge #2148 into master.

PR #2163, opened 2020-12-11
Enhancing pbs_preempt_job to support multi-server

agrawalravi90:
#### Describe Bug or Feature

This enhances pbs_preempt_jobs to perform preemption under a multi-server setup

#### Describe Your Change

- pbs_preempt_jobs will broadcast the list of job ids to all servers, in parallel
- each server responds back a list of jobids and preempt order. For jobs not found in each server, the preempt order says 'D'
- pbs_preempt_jobs then collates the replies, finds the jobs which were missing on a server in other servers' lists and returns the aggregated result back to the scheduler.

#### Preempting multiple jobs across multiple servers:
Since multi-node job support is missing right now for multi-server, here's a gdb output showing how pbs_preempt_jobs does the right thing. Setup: 2 jobs, each on a different server, are being preempted:
```

Thread 1 "pbs_sched" hit Breakpoint 1, pbs_preempt_jobs (c=42, preempt_jobs_list=0x1297ab0) at ../../../../src/lib/Libpbs/../Libifl/ifl_impl.c:479
479		return (*pfn_pbs_preempt_jobs)(c, preempt_jobs_list);
(gdb) s
__pbs_preempt_jobs (c=42, preempt_jobs_list=0x1297ab0) at ../../../../src/lib/Libpbs/../Libifl/pbsD_Preempt_Jobs.c:154
154		preempt_job_info **p_replies = NULL;
(gdb) n
155		svr_conn_t **svr_connections = get_conn_svr_instances(c);
(gdb) 
156		int num_cfg_svrs = get_num_servers();
(gdb) 
159		int count = 0;
(gdb) 
160		int retidx = -1;
(gdb) 
161		preempt_job_info *ret = NULL;
(gdb) 
162		void *missing_jobs = NULL;
(gdb) 
164		if (!svr_connections)
(gdb) 
167		skip = calloc(num_cfg_svrs, sizeof(int));
(gdb) 
168		if (skip == NULL) {
(gdb) 
172		p_replies = malloc(num_cfg_svrs * sizeof(preempt_job_info *));
(gdb) 
173		if (p_replies == NULL) {
(gdb) 
179		missing_jobs = pbs_idx_create(0, 0);
(gdb) 
180		if (missing_jobs == NULL) {
(gdb) 
186		if (pbs_client_thread_init_thread_context() != 0)
(gdb) 
191		if (pbs_client_thread_lock_connection(c) != 0)
(gdb) 
197		for (i = 0; i < num_cfg_svrs; i++) {
(gdb) 
198			if (preempt_jobs_send(svr_connections[i]->sd, preempt_jobs_list) != 0)
(gdb) 
197		for (i = 0; i < num_cfg_svrs; i++) {
(gdb) 
198			if (preempt_jobs_send(svr_connections[i]->sd, preempt_jobs_list) != 0)
(gdb) 
197		for (i = 0; i < num_cfg_svrs; i++) {
(gdb) 
215		for (i = 0; i < num_cfg_svrs; i++) {
(gdb) 
216			if (!skip[i]) {
(gdb) 
217				p_replies[i] = preempt_jobs_recv(svr_connections[i]->sd, &count);
(gdb) 
218				if (retidx == -1)
(gdb) 
219					retidx = i;
(gdb) 
215		for (i = 0; i < num_cfg_svrs; i++) {
(gdb) 
216			if (!skip[i]) {
(gdb) 
217				p_replies[i] = preempt_jobs_recv(svr_connections[i]->sd, &count);
(gdb) 
218				if (retidx == -1)
(gdb) 
215		for (i = 0; i < num_cfg_svrs; i++) {
(gdb) 
223		free(skip);
(gdb) 
225		if (retidx == -1) { /* Didn't receive anything/all failed */
(gdb) 
231		for (i = 0; i < count; i++) {
(gdb) 
232			if (p_replies[retidx][i].order[0] == 'D' && num_cfg_svrs > 1) {
(gdb) p p_replies[retidx][i]
$1 = {job_id = "0.pbsc8-2", '\000' <repeats 264 times>, order = "D\000\000\000\000"}
(gdb) n
233				int *data = malloc(sizeof(int));
(gdb) 
234				if (data == NULL) {
(gdb) 
238				*data = i;
(gdb) 
239				pbs_idx_insert(missing_jobs, p_replies[retidx][i].job_id, data);
(gdb) 
231		for (i = 0; i < count; i++) {
(gdb) 
232			if (p_replies[retidx][i].order[0] == 'D' && num_cfg_svrs > 1) {
(gdb) p p_replies[retidx][i]
$2 = {job_id = "2.pbsc8", '\000' <repeats 266 times>, order = "S\000\000\000\000"}
(gdb) n
231		for (i = 0; i < count; i++) {
(gdb) 
242		ret = p_replies[retidx];
(gdb) 
244		if (missing_jobs != NULL && num_cfg_svrs > 1) {
(gdb) 
245			for (i = retidx + 1; i < num_cfg_svrs; i++) {
(gdb) 
249				for (j = 0; j < count; j++) {
(gdb) 
250					if (p_replies[i] == NULL)
(gdb) 
252					if (p_replies[i][j].order[0] != 'D') {
(gdb) 
249				for (j = 0; j < count; j++) {
(gdb) 
250					if (p_replies[i] == NULL)
(gdb) 
252					if (p_replies[i][j].order[0] != 'D') {
(gdb) p p_replies[i][j]
$3 = {job_id = "0.pbsc8-2", '\000' <repeats 266 times>, order = "S\000\000\000\000"}
(gdb) 
256						int *index = NULL;
(gdb) 
257						char *jid = strdup(p_replies[i][j].job_id);
(gdb) 
258						pbs_idx_find(missing_jobs, (void **) &jid, (void **) &index, NULL);
(gdb) p jid
$4 = 0x129a370 "0.pbsc8-2"
(gdb) n
259						free(jid);
(gdb) p *index
$5 = 0
(gdb) n
260						if (index != NULL && *index < count)
(gdb) n
261							strcpy(ret[*index].order, p_replies[i][j].order);
(gdb) n
262						free(index);
(gdb) n
263						break;
(gdb) n
245			for (i = retidx + 1; i < num_cfg_svrs; i++) {
(gdb) n
270		for (i = 0; i < num_cfg_svrs; i++) {
(gdb) n
271			if (i == retidx)
(gdb) n
272				continue;
(gdb) n
270		for (i = 0; i < num_cfg_svrs; i++) {
(gdb) 
271			if (i == retidx)
(gdb) 
273			free(p_replies[i]);
(gdb) 
270		for (i = 0; i < num_cfg_svrs; i++) {
(gdb) 
275		free(p_replies);
(gdb) 
276		pbs_idx_destroy(missing_jobs);
(gdb) 
279		if (pbs_client_thread_unlock_connection(c) != 0)
(gdb) 
282		return ret;
(gdb) p ret
$6 = (preempt_job_info *) 0x129a770
(gdb) p ret[0]
$7 = {job_id = "0.pbsc8-2", '\000' <repeats 264 times>, order = "S\000\000\000\000"}
(gdb) p ret[1]
$8 = {job_id = "2.pbsc8", '\000' <repeats 266 times>, order = "S\000\000\000\000"}

```


bhroam:
@agrawalravi90 pbs_preemptjobs() reports back to the scheduler how the job was preempted.  You can't use 'D' for something else because that means the job was deleted.  It's perfectly valid for a single server to report back 'D'.

agrawalravi90:
> You can't use 'D' for something else because that means the job was deleted. It's perfectly valid for a single server to report back 'D'.

True, but that's what the server returns today when it cannot find a job. Functionally, they are equivalent. if one server preempts the job with 'D', all other servers will also have the preempt orders set to 'D' because they couldn't find it. So, scheduler will get back 'D' and can assume that the job was deleted, regardless of the method.

agrawalravi90:
Please let me know if I've understood any part of preemption incorrectly.

src/lib/Libifl/pbsD_Preempt_Jobs.c
```diff
@@ -77,7 +191,88 @@ __pbs_preempt_jobs(int c, char **preempt_jobs_list) {
...
+	/* With multi-server, jobs are sharded across multiple servers.
+	 * So, send the request to all the active servers and collate their replies
+	 */
+	for (i = 0; i < num_cfg_svrs; i++) {
+		if (preempt_jobs_send(svr_connections[i]->sd, preempt_jobs_list) != 0)
+			skip[i] = 1;
+	}
+
+	/**
+	 * We will get N lists back, one for each server, each of the same length
```
arungrover, line 203: It's a clever implementation 👍🏼 
Can I suggest an alternative - 
 We know that usually preempt order of jobs is "SCRD" so in most cases servers will be able to suspend, checkpoint or requeue job before deleting it from the system.
So creating a tree for all jobs in 'D'  and then reconciling their order as and when you go through the response of all other server is IMO not necessary.
If you make the server return a new preempt order response (say 'U') for jobs it could not find (in a multi-server scenario). You can then just ignore these jobs (with order U) from the responses. In most cases, you will find them in some other server's response. There only remains 1 case where you may not get a response from all the servers which is when job ends between the start of the cycle and the time scheduler sends preempt_job requests. In this case, the scheduler can just assume that the job is there in the system and end it anyway.

src/lib/Libifl/pbsD_Preempt_Jobs.c
```diff
@@ -77,7 +191,88 @@ __pbs_preempt_jobs(int c, char **preempt_jobs_list) {
...
+	}
+	ret = p_replies[retidx];
+
+	if (missing_jobs != NULL && num_cfg_svrs > 1) {
+		for (i = retidx + 1; i < num_cfg_svrs; i++) {
+			int j;
+
+			/* Check if the missing jobs were found on any other server */
+			for (j = 0; j < count; j++) {
+				if (p_replies[i] == NULL)
```
arungrover, line 250: I think this check should move out of the inner for loop.

arungrover (changes requested):

bhroam (changes requested):

src/lib/Libifl/pbsD_Preempt_Jobs.c
agrawalravi90: Thanks for the suggestion Arun, I didn't realize that the scheduler already can deal with missing jobs from the server's response, that makes your suggestion less complex than what i have right now. So, I'll try it out.

agrawalravi90: i had a chat with Bhroam, he mentioned that scheduler doesn't currently handle any missing jobs from the reply, it expects all jobs to be returned. So, I think it might be better to keep the behavior as it is and have IFL handle it instead. Let me know what you think.

src/lib/Libifl/pbsD_Preempt_Jobs.c
```diff
@@ -77,7 +191,88 @@ __pbs_preempt_jobs(int c, char **preempt_jobs_list) {
...
+	for (i = 0; i < num_cfg_svrs; i++) {
+		if (preempt_jobs_send(svr_connections[i]->sd, preempt_jobs_list) != 0)
+			skip[i] = 1;
+	}
+
+	/**
+	 * We will get N lists back, one for each server, each of the same length
+	 * We will do the following:
+	 * 	- go through the first list, find the index which has order == "D",
+	 * 	this implies that the job was not found on the server.
```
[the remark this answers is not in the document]
agrawalravi90: Will do, thanks.

src/lib/Libifl/pbsD_Preempt_Jobs.c
```diff
@@ -77,7 +191,88 @@ __pbs_preempt_jobs(int c, char **preempt_jobs_list) {
...
 		return NULL;
 
-	ret = PBSD_preempt_jobs(c, preempt_jobs_list);
+	/* With multi-server, jobs are sharded across multiple servers.
+	 * So, send the request to all the active servers and collate their replies
+	 */
+	for (i = 0; i < num_cfg_svrs; i++) {
+		if (preempt_jobs_send(svr_connections[i]->sd, preempt_jobs_list) != 0)
+			skip[i] = 1;
+	}
```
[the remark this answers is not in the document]
agrawalravi90: Thanks for catching that. I'll have the function barf and return failure so that scheduler can reconnect with the servers and try again.

src/lib/Libifl/pbsD_Preempt_Jobs.c
arungrover: Oh, I didn't mean to suggest that scheduler in its current form can handle this situation, It will obviously need more changes to do that.
Inside scheduler, we have a loop iterating on preempt_jobs_reply array. While doing to you might want to count how many jobs are there in this array. When the array ends check if preempted jobs count and no_of_jobs variable are equal. In most cases, it will be equal. In the cases where you don't find them equal, you remove the intersection of arrays and then consider all the jobs in the leftover list as deleted.

src/lib/Libifl/pbsD_Preempt_Jobs.c
```diff
@@ -77,7 +191,88 @@ __pbs_preempt_jobs(int c, char **preempt_jobs_list) {
...
+	free(skip);
+
+	if (retidx == -1) { /* Didn't receive anything/all failed */
+		free(p_replies);
+		return NULL;
+	}
+
+	/* Find jobs which couldn't be found on the server and store them in AVL tree for multi-server */
+	for (i = 0; i < count; i++) {
```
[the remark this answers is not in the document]
agrawalravi90: hm .. I also use count later to walk through the rest of the arrays, so if it's inconsistent, then I'll walk off the end of the array for others. Do we really want to handle this? I don't want to make the code complicated to handle a case which is highly unlikely. Let me know what you think.

agrawalravi90:
@bhroam  and @arungrover  I've addressed your comments, please provide further feedback, thanks!

src/lib/Libifl/pbsD_Preempt_Jobs.c
agrawalravi90: thanks for clarifying Arun. That sounds very similar to what i had to do inside pbs_preempt_jobs(), so I'm not sure if it will be any easier. I do like the idea of the IFL layer dealing with multi-server specific weirdness than the clients (scheduler in this case). Please let me know what you think.

arungrover: Well, I think doing the change will make your algo faster in the average case.  It is up to you if you want to leave it as is.

agrawalravi90: Just trying to understand how it'll be faster. Right now my algo does the following:
- goes through the first array, stores the jobs with state == 'D' in an AVL tree: nlogn
- goes through the rest of them, checks for state != 'D', updates value in AVL tree: mnlogn

with sched doing this, IFL will still do mn to go through n lists of m servers. Can scheduler resolve missing jobs in less than mnlogn ? You mentioned something about finding intersection of the different arrays so it sounded like a similar complexity to me. Please let me know.

agrawalravi90: Ok, I think i understand what you mean. Since in most cases we'll find all jobs which we are preempting, we'll save the logn cost of the AVL tree. TBH, I think it won't make a big difference, I like the part where all the multi-server related handling happens in IFL itself and scheduler doesn't have to bother, but let me know if you feel strongly about this.

arungrover: when scheduler computes the intersection, it happens only in cases where none of the servers respond a job id (or number of job ids), that's worst case complexity. In this implementation average is also equally complex.

arungrover: IMO, in the average case, it's not logn we save, it's "nlogn-n" that we save, which is huge. I also don't think there is any other component in PBS that uses this ifl call, plus, if you make a change in server to return 'U' only in case of multi-server, IFL call would not adversely affect backward compatibility as well.

agrawalravi90: if the difference is huge then I'd of course like to change the approach. But, can you please explain how it's nlogn less complex than what I have now? The only additional complexity in my approach now comes from insertion, search and deletion from the AVL tree, all of which should only add logn extra complexity. Am I missing something?

arungrover: So let's consider the average case where you are preempting 100 jobs which are evenly distributed between 3 servers.
All servers will respond with 33 jobs with 'S' state 67 with 'D'.
Now you loop on 100 jobs from server 1 to create an AVL tree of 67 jobs. [nlog(n)]
You then loop through all 200 jobs (from the other 2 servers) and do a find on 67 of the other deleted jobs in the tree.
in an average case, you will find all jobs and mark them preempted. [(m-1) * nlog(n)]
then you get rid of the tree. [n]
Correct me if I'm wrong, time complexity here is -  m*nlogn + n that for 100 jobs could be 3*100(6) + 100 -> 1900. On the other hand, if you just discard jobs in 'U' state, the average case drops to 3*100. Besides each AVL search operation involves strdup and strcmp whereas walking down a list and ignoring jobs wouldn't need any strcmp.

bhroam: I'm not 100% sure how this makes his algorithm faster.  He needs to combine all N server arrays into one.  He can use his tree approach for O(NlgN) or use a more conventional search for O(N^2).  Am I missing something?

I also agree with Ravi that having IFL handle this is the right place.  While the scheduler is the only one who uses pbs_preempt_jobs() now, it is part of IFL, so if someone else wanted to write a scheduler to use it, they shouldn't have to jump through any hoops.

arungrover: All I'm saying is that he does not have to do this combination in all the cases, This can be done only when it is needed. In most cases, it won't be needed.
I also don't think the reasoning that it should all be co-located in IFL is something we have been following as a thumb rule. We currently have a PR out for review where the client qdel collates responses from server to find out jobs that the server that couldn't find (Unknown job id)  and route those delete requests to the right servers. 
Plus this is a new functionality written specifically for performance in mind, I feel we should cut time expense wherever possible.
That said, If you both strongly feel that this implementation is the right way to go forward, I will stop pestering about it.

agrawalravi90: > We currently have a PR out for review where the client qdel collates responses from server to find out jobs that the server that couldn't find (Unknown job id) and route those delete requests to the right servers.

That's the qdel program's ability to connect to and delete jobs from multiple clusters. It was very much deliberate to not put that functionality in IFL because then IFL would be creating new connections itself, which is not what it's meant for. But we are indeed enhancing pbs_deljoblist to completely cover talking to multiple server instances for the cluster handler given to it and collating data itself. So, having pbs_preempt_jobs() do the same would be consistent with it. We've done the same with qrun and pbs_runjob().

> Correct me if I'm wrong, time complexity here is - m*nlogn + n that for 100 jobs could be 3*100(6) + 100 -> 1900. On the other hand, if you just discard jobs in 'U' state, the average case drops to 3*100

That's correct, but it's still O(mnlogk) vs O(mn), where k (< n) is the number of jobs not found on the first server. I still don't think that it's a big difference, we will only start seeing it for larger number of preempted jobs. This PR makes the assumption that we usually don't preempt a large number of jobs. If, that's not true, then there is a much bigger optimization that we can do which will also get rid of the AVL tree: not broadcast all jobs to all servers, group them by each server instance instead and send each server its own jobs. We aren't doing that now because it will be unnecessarily complex code (like pbs_deljoblist). So, can we wait and see if this really is a bottleneck and fix it later if it is? Right now we don't even support multi-node jobs running on multiple server instances, so it'll be hard to test and get numbers.

arungrover: I do not agree with the maths there. Unless n is exponentially bigger than k, log(n) or log(k) isn't going to be too far apart. Like in the example, log(base2) of 100 or log (base2) of 67 both lie between 6 to 7.
In any case, you can keep the implementation the way it is.

agrawalravi90: I was just trying to be technically correct by using k, I agree that it's probably the same as n, I was trying to point out that the difference is still logn, not nlogn. Anyways, thanks for letting me keep it as is, if we notice any performance issues with this in the future, we can revisit this.

src/lib/Libifl/pbsD_Preempt_Jobs.c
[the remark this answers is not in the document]
bhroam: How about this: keep 2 counts.  The last count and the current count.  If the two are ever not the same, fail the preemption attempt.  That shouldn't add much complexity.

src/lib/Libifl/pbsD_Preempt_Jobs.c
```diff
@@ -257,26 +242,35 @@ __pbs_preempt_jobs(int c, char **preempt_jobs_list)
...
-		return NULL;
-
 	return ret;
+
+err:
+	if (p_replies != NULL) {
+		for (i = 0; i < num_cfg_svrs; i++) {
+			free(p_replies[i]);
+		}
+		free(p_replies);
```
bhroam, line 273: Don't you have to free the tree too?

agrawalravi90:
@bhroam thanks for the feedback, I've addressed your comments and also fixed an issue with server_instance_id where it was not getting set correctly after a move and run. Please provide further feedback.

bhroam (approved):
LGTM.  Thanks for making the changes.

agrawalravi90:
@bhroam thanks! @arungrover please provide further feedback whenever you get a chance.

arungrover (approved):
LGTM, Thanks for making changes

PR #2165, opened 2020-12-14
 Making legacy failover to work properly

bremanandjk:
#### Describe Bug or Feature

In a failover setup after the primary server is down, the front end clients do not automatically talk to the secondary anymore. The refactoring work on pbs_connet() causes the failure in pbs_server failover. In this PR, I am fixing the broken changes and making the failover to work properly. 

#### Describe Your Change

* The failover code which recognizes the pbs_conf switches are added back and refactored to work with multi-server. 
* Fixed the bug in scheduler registering logic. Added the code to check both the primary and secondary hosts for connection validation.

#### Link to Design Doc

N/A

[Failover_test_logs.txt](https://github.com/openpbs/openpbs/files/5691017/Failover_test_logs.txt)

Profile: regression
|--|--|--|--|--|--|--|
|5316|3842|5|0|0|0|3837|

Description: Rerun Only Failed Tests From #5316
Profile: regression
|--|--|--|--|--|--|--|
|5317|5|3|0|0|0|2|

These 3 test failures are seen in the mainline branch as well.

### DIFF
--- a/src/lib/Libifl/pbsD_connect.c
+++ b/src/lib/Libifl/pbsD_connect.c
@@ -658,16 +658,17 @@ __pbs_connect_extend(char *server, char *extend_data)
 		/* failover configuered ...   */	
 		if (is_same_host(server, pbs_conf.pbs_primary)) {	
 			have_alt = 1;	
+			
+			altservers[0] = pbs_conf.pbs_primary;	
+			altservers[1] = pbs_conf.pbs_secondary;	
+			
+#ifdef CHECK_FILE	
 			/* We want to try the one last seen as "up" first to not   */	
 			/* have connection delays.   If the primary was up, there  */	
 			/* is no .pbsrc.NAME file.  If the last command connected  */	
 			/* to the Secondary, then it created the .pbsrc.USER file. */	
 
-			/* see if already seen Primary down */	
-#ifndef CHECK_FILE	
-			altservers[0] = pbs_conf.pbs_primary;	
-			altservers[1] = pbs_conf.pbs_secondary;	
-#else	
+			/* see if already seen Primary down */		
 			snprintf(pbsrc, _POSIX_PATH_MAX, "%s/.pbsrc.%s", pbs_conf.pbs_tmpdir, pbs_current_user);
 			if (stat(pbsrc, &sb) == -1) {	
 				/* try primary first */	
--- a/src/lib/Libifl/pbsD_connect.c
+++ b/src/lib/Libifl/pbsD_connect.c
@@ -632,12 +632,10 @@ __pbs_connect_extend(char *server, char *extend_data)
 	int	sock = -1;
 	int	i;
 	int	f;
-	
-#ifdef CHECK_FILE	
+		
 	char   pbsrc[_POSIX_PATH_MAX];	
 	struct stat sb;	
 	int    using_secondary = 0;	
-#endif
 	
 	/* initialize the thread context data, if not already initialized */
 	if (pbs_client_thread_init_thread_context() != 0)
@@ -661,8 +659,7 @@ __pbs_connect_extend(char *server, char *extend_data)
 			
 			altservers[0] = pbs_conf.pbs_primary;	
 			altservers[1] = pbs_conf.pbs_secondary;	
-			
-#ifdef CHECK_FILE	
+				
 			/* We want to try the one last seen as "up" first to not   */	
 			/* have connection delays.   If the primary was up, there  */	
 			/* is no .pbsrc.NAME file.  If the last command connected  */	
@@ -675,8 +672,7 @@ __pbs_connect_extend(char *server, char *extend_data)
 				altservers[0] = pbs_conf.pbs_secondary;		
 				altservers[1] = pbs_conf.pbs_primary;
 				using_secondary = 1;	
-			} 	
-#endif	
+			} 		
 		}	
 	}
 
@@ -700,7 +696,6 @@ __pbs_connect_extend(char *server, char *extend_data)
 		return -1; 		/* cannot connect */
 	}
 	
-#ifdef CHECK_FILE
 	if (have_alt && (i == 1)) {
 		/* had to use the second listed server ... */
 		if (using_secondary == 1) {
@@ -713,7 +708,6 @@ __pbs_connect_extend(char *server, char *extend_data)
 				close(f);
 		}
 	}
-#endif
 	
 	return sock;
 }
--- a/src/lib/Libifl/pbsD_connect.c
+++ b/src/lib/Libifl/pbsD_connect.c
@@ -706,7 +706,7 @@ __pbs_connect_extend(char *server, char *extend_data)
 			altservers[0] = pbs_conf.pbs_primary;	
 			altservers[1] = pbs_conf.pbs_secondary;	
 #else	
-			(void)snprintf(pbsrc, _POSIX_PATH_MAX, "%s/.pbsrc.%s", pbs_conf.pbs_tmpdir, pbs_current_user);	
+			snprintf(pbsrc, _POSIX_PATH_MAX, "%s/.pbsrc.%s", pbs_conf.pbs_tmpdir, pbs_current_user);
 			if (stat(pbsrc, &sb) == -1) {	
 				/* try primary first */	
 				altservers[0] = pbs_conf.pbs_primary;	
@@ -728,7 +728,7 @@ __pbs_connect_extend(char *server, char *extend_data)
 	 *   if attempting to connect to Primary,  try the Secondary	
 	 *   if attempting to connect to Secondary, try the Primary	
 	 */	
-	for (i=0; i<(have_alt+1); ++i) {
+	for (i = 0; i<(have_alt+1); ++i) {
 		if (have_alt) 
 			pbs_strncpy(server_name, altservers[i], PBS_MAXSERVERNAME);
 		if ((sock = connect_to_servers(server_name, server_port, extend_data)) != -1)
--- a/src/lib/Libifl/pbsD_connect.c
+++ b/src/lib/Libifl/pbsD_connect.c
@@ -690,7 +690,7 @@ __pbs_connect_extend(char *server, char *extend_data)
 	 *   if attempting to connect to Primary,  try the Secondary	
 	 *   if attempting to connect to Secondary, try the Primary	
 	 */	
-	for (i = 0; i<(have_alt+1); ++i) {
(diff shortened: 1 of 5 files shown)


src/lib/Libifl/pbsD_connect.c
```diff
@@ -171,6 +171,45 @@ get_hostsockaddr(char *host, struct sockaddr_in *sap)
...
+ * @retval	0	failure	
+ *	
+ */	
+static int	
+hostnmcmp(char *s1, char *s2)	
+{	
+	/* Return failure if any/both the names are NULL. */	
+	if (s1 == NULL || s2 == NULL)	
+		return 1;	
+#ifdef WIN32
```
agrawalravi90, line 193: I think Windows code is not needed (here and other places)

agrawalravi90 (changes requested):

src/lib/Libifl/pbsD_connect.c
bremanandjk: Why do you think? Windows command would still use pbs_connet() IFL to Linux server.

src/lib/Libifl/pbsD_connect.c
```diff
@@ -171,6 +171,45 @@ get_hostsockaddr(char *host, struct sockaddr_in *sap)
...
+ * @param[in] s1 - hostname1	
+ * @param[in] s2 - hostname2	
+ *	
+ * @return	int	
+ * @retval	1	success	
+ * @retval	0	failure	
+ *	
+ */	
+static int	
+hostnmcmp(char *s1, char *s2)
```
nithinj, line 188: Can you make use of is_same_host() instead?

src/lib/Libifl/pbsD_connect.c
```diff
@@ -638,7 +689,75 @@ __pbs_connect_extend(char *server, char *extend_data)
...
+			/* is no .pbsrc.NAME file.  If the last command connected  */	
+			/* to the Secondary, then it created the .pbsrc.USER file. */	
+
+			/* see if already seen Primary down */	
+#ifdef WIN32	
+			/* due to windows quirks, all try both in same order */	
+			altservers[0] = pbs_conf.pbs_primary;	
+			altservers[1] = pbs_conf.pbs_secondary;	
+#else	
+			(void)snprintf(pbsrc, _POSIX_PATH_MAX, "%s/.pbsrc.%s", pbs_conf.pbs_tmpdir, pbs_current_user);
```
nithinj, line 709: Remove void typecast here and elsewhere.

src/lib/Libifl/pbsD_connect.c
```diff
@@ -638,7 +689,75 @@ __pbs_connect_extend(char *server, char *extend_data)
...
+		}	
+	}
+
+	/*
+	 * connect to server ...	
+	 * If attempt to connect fails and if Failover configured and	
+	 *   if attempting to connect to Primary,  try the Secondary	
+	 *   if attempting to connect to Secondary, try the Primary	
+	 */	
+	for (i=0; i<(have_alt+1); ++i) {
```
nithinj, line 731: leave a space between operator.

nithinj (changes requested):

src/lib/Libifl/pbsD_connect.c
agrawalravi90: oops, my bad, that's true

agrawalravi90 (dismissed):
LGTM

src/lib/Libifl/pbsD_connect.c
bremanandjk: There was a special code to support "localhost" & "localhost.domain" in hostnmcmp(). we cant use is_same_host() in this place.

bremanandjk:
@agrawalravi90 Thanks for your approval, 
@nithinj Addressed your comments, please have a look.

src/lib/Libifl/pbsD_connect.c
nithinj: Does that already guaranteed by get_fullhostname() within is_same_host?
If not why don't you extend is_same_host instead of adding a new function?

bremanandjk:
@nithinj @agrawalravi90 
I have addressed further review comments, plz have a look.

src/lib/Libifl/pbsD_connect.c
```diff
@@ -638,7 +652,74 @@ __pbs_connect_extend(char *server, char *extend_data)
...
+		}	
+	}
+
+	/*
+	 * connect to server ...	
+	 * If attempt to connect fails and if Failover configured and	
+	 *   if attempting to connect to Primary,  try the Secondary	
+	 *   if attempting to connect to Secondary, try the Primary	
+	 */	
+	for (i = 0; i<(have_alt+1); ++i) {
```
nithinj, line 693: Leave space between operator

src/lib/Libifl/pbsD_connect.c
```diff
@@ -638,7 +652,74 @@ __pbs_connect_extend(char *server, char *extend_data)
...
+	if (have_alt && (i == 1)) {
+		/* had to use the second listed server ... */
+		if (using_secondary == 1) {
+			/* remove file that causes trying the Secondary first */
+			unlink(pbsrc);
+		} else {
+			/* create file that causes trying the Primary first   */
+			f = open(pbsrc, O_WRONLY|O_CREAT, 0200);
+			if (f != -1)
+				(void)close(f);
```
nithinj, line 717: remove typecast

src/server/process_request.c
```diff
@@ -293,10 +293,23 @@ req_register_sched(conn_t *conn, struct batch_request *preq)
 		rc = PBSE_UNKSCHED;
 		goto rerr;
 	}
+
 	if (sched->sc_conn_addr != conn->cn_addr) {
-		rc = PBSE_BADHOST;
-		goto rerr;
+		if (pbs_conf.pbs_primary != NULL && pbs_conf.pbs_secondary != NULL) {
+			pbs_net_t addr = get_hostaddr(pbs_conf.pbs_primary);
```
nithinj, line 299: just curious; does the server can send a register_sched? why are you checking the connection against pbs_primary and secondary?

nithinj (changes requested):

src/server/process_request.c
bremanandjk: The current design is that sched is sending the registration request to a server. At this condition, the server is verifying the requesting sched's host. Since the connection could be either from the primary or secondary host,  the validation is checking on both.

nithinj (approved):
LGTM!

src/lib/Libifl/pbsD_connect.c
```diff
@@ -638,7 +652,74 @@ __pbs_connect_extend(char *server, char *extend_data)
...
+		/* failover configuered ...   */	
+		if (is_same_host(server, pbs_conf.pbs_primary)) {	
+			have_alt = 1;	
+			/* We want to try the one last seen as "up" first to not   */	
+			/* have connection delays.   If the primary was up, there  */	
+			/* is no .pbsrc.NAME file.  If the last command connected  */	
+			/* to the Secondary, then it created the .pbsrc.USER file. */	
+
+			/* see if already seen Primary down */	
+#ifndef CHECK_FILE
```
nithinj, line 667: I think we could avoid ifndef block. Keep it as the default. Override the bheavior if CHECK_FILE is defined.

src/lib/Libifl/pbsD_connect.c
```diff
@@ -638,7 +652,75 @@ __pbs_connect_extend(char *server, char *extend_data)
...
+#ifdef CHECK_FILE	
+			/* We want to try the one last seen as "up" first to not   */	
+			/* have connection delays.   If the primary was up, there  */	
+			/* is no .pbsrc.NAME file.  If the last command connected  */	
+			/* to the Secondary, then it created the .pbsrc.USER file. */	
+
+			/* see if already seen Primary down */		
+			snprintf(pbsrc, _POSIX_PATH_MAX, "%s/.pbsrc.%s", pbs_conf.pbs_tmpdir, pbs_current_user);
+			if (stat(pbsrc, &sb) == -1) {	
+				/* try primary first */
```
nithinj, line 674: You can remove this block as well.

bremanandjk:
@agrawalravi90 
>I'm not convinced that we need the interleaved platform specific code with CHECK_FILE, I'll let others more knowledgable chime in, so dismissing my review.

There is no any platform-specific code. If the CHECK_FILE is enabled, there will be an optimization that the client connects to the last used server in failover mode then try the alternate server, if not defined, the client connects in the default order that first primary, then secondary.

subhasisb:
> @agrawalravi90
> 
> > I'm not convinced that we need the interleaved platform specific code with CHECK_FILE, I'll let others more knowledgable chime in, so dismissing my review.
> 
> There is no platform-specific code. If the CHECK_FILE is enabled, there will be an optimization that the client connects to the last used server in failover mode then try the alternate server. If not defined, the client connects in the default order that first try primary, then secondary.

I think we should remove CHECK_FILE. Just do the file check all the time - there is no need for this macro

subhasisb:
Another thing. Did you check that the secondary scheduler is being shut down in legacy failover when primary becomes active again

src/server/process_request.c
subhasisb: This is fine. However, I found in my testing that after a failover, and when primary is back, the scheduler on the secondary server does not shut down - i think that must be shut down

bremanandjk:
> > @agrawalravi90
> > > I'm not convinced that we need the interleaved platform specific code with CHECK_FILE, I'll let others more knowledgable chime in, so dismissing my review.
> > 
> > 
> > There is no platform-specific code. If the CHECK_FILE is enabled, there will be an optimization that the client connects to the last used server in failover mode then try the alternate server. If not defined, the client connects in the default order that first try primary, then secondary.
> 
> I think we should remove CHECK_FILE. Just do the file check all the time - there is no need for this macro. The code to check the file (that you have modified) is already platform independent, so why differentiate in the first place?

okay, I have removed the CHECK_FILE macro as suggested.

bremanandjk:
> Another thing. Did you check that the secondary scheduler is being shut down in legacy failover when primary becomes active again

Yes, I have tested this config too, as expected the secondary scheduler is shutting down. 
test logs
[linux_failover_sec_sched_shutdown_test.txt](https://github.com/openpbs/openpbs/files/5718275/linux_failover_sec_sched_shutdown_test.txt)

subhasisb (approved):

PR #2166, opened 2020-12-14
pbs_ralter test might fail on slow machines

arungrover:
#### Describe Bug or Feature
On some slow machines ralter test - test_alter_standing_resv_end_time_before_run fails.

#### Describe Your Change
Problem is that the test alters a reservation that is intended to start 10 seconds in the future. While altering the reservation it expects the reservation to be in a CONFIRMED state. But on a slow machine, it might happen that the reservation might move into a running state and this means the test might fail because it expects the reservation to be in the confirmed state.
I've increased the offset of the reservation to 30 seconds instead of 10 which means the reservation will start 30 seconds in the future.

#### Link to Design Doc
NA

Description: Rerun All Tests From #5318
Platforms: SUSE15,CENTOS8
Profile: regression
|--|--|--|--|--|--|--|
|5335|1929|2|0|0|1|1926|

Tests failed are not from the same testsuite.

### DIFF
--- a/test/tests/functional/pbs_ralter.py
+++ b/test/tests/functional/pbs_ralter.py
@@ -823,7 +823,7 @@ class TestPbsResvAlter(TestFunctional):
         """
         duration = 30
         shift = 10
-        offset = 10
+        offset = 30
         rid, start, end = self.submit_and_confirm_reservation(offset, duration,
                                                               standing=True)
 


vstumpf (approved):

nithinj (approved):
LGTM!

agrawalravi90 (approved):

PR #2169, opened 2020-12-16
Fix for server crash in multi-svr

nithinj:
#### Describe Bug or Feature
* Job owning server crashes when move+run job fails due to an error while submitting the job.
* Mom choosing between primary and secondary servers randomly could result in skewed results.

#### Describe Your Change
* Move and run sends two requests to the destination server - submit and run. When the submit fails due to an error, it sends the error back and the source server sends it back to the client. When it receives the error corresponding to the run request, the request is invalid as it has been already processed.
The fix is to delete any existing task corresponding to the request after replying.
* Mom will choose between primary and secondary in an interleaved way than random

[move_run_fix.log](https://github.com/openpbs/openpbs/files/5703167/move_run_fix.log)

### DIFF
--- a/src/resmom/catch_child.c
+++ b/src/resmom/catch_child.c
@@ -831,7 +831,8 @@ end_loop:
 
 /**
  * @brief
- * 		choosing one server in random if a failover server is already set up.
+ * 		choosing the server to connect to if a failover server is already set up.
+ * 		Selects primary and secondary alternatively.
  *
  * @param[out] port - Passed through to parse_servername(), not modified here.
  *
@@ -840,13 +841,16 @@ end_loop:
  * @retval !NULL - pointer to server name
  */
 static char *
-get_servername_random(unsigned int *port)
+get_servername_failover(unsigned int *port)
 {
+	static int whom_to_connect = 0;
 
 	if (!pbs_conf.pbs_secondary)
 		return get_servername(port);
 	else {
-		if (rand() % 2 == 0)
+		whom_to_connect = (whom_to_connect + 1) % 2;
+
+		if (whom_to_connect)
 			return get_servername(port);
 		else
 			return parse_servername(pbs_conf.pbs_secondary, port);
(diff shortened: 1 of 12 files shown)


src/resmom/catch_child.c
```diff
@@ -840,13 +841,16 @@ scan_for_exiting(void)
...
-get_servername_random(unsigned int *port)
+get_servername_failover(unsigned int *port)
 {
+	static int whom_to_connect = 0;
 
 	if (!pbs_conf.pbs_secondary)
 		return get_servername(port);
 	else {
-		if (rand() % 2 == 0)
+		whom_to_connect = (whom_to_connect + 1) % 2;
```
agrawalravi90, line 851: Simpler:
```suggestion
		whom_to_connect = !whom_to_connect;
```

agrawalravi90 (changes requested):
Minor comment

src/server/multi_svr.c
```diff
@@ -279,3 +279,30 @@ init_msi()
...
+char *
+get_svr_inst_id(void)
+{
+	char svr_inst_name[PBS_MAXHOSTNAME + 1];
+	unsigned int svr_inst_port;
+	char *svr_inst_id = NULL;
+
+
+	if (gethostname(svr_inst_name, PBS_MAXHOSTNAME) == 0)
+        	get_fullhostname(svr_inst_name, svr_inst_name, PBS_MAXHOSTNAME);
```
ashwathraop, line 300: No need to handle possible failure of get_fullhostname()? It can return -1 right?

nithinj: In that case, we still retain the original value

nithinj:
@ashwathraop & @agrawalravi90 ,
Thanks for the reviews. I have addressed your comments. Please have a look.

src/server/multi_svr.c
```diff
@@ -279,3 +279,30 @@ init_msi()
...
+/**
+ * @brief
+ * 	Used to get serverer_instance_id which is of the form server_instance_name:server_instance_port
+ * 
+ * @return char *
+ * @return NULL - failure
+ * @retval !NULL - pointer to server_instance_id
+ */
+char *
+get_svr_inst_id(void)
```
agrawalravi90, line 292: since this is actually allocating memory, how about 'create_svr_inst_id'/'generate_svr_inst_id' ?

agrawalravi90 (changes requested):
sorry, one more minor comment

nithinj:
@agrawalravi90 ,
I have addressed your comments. Please have a look.

agrawalravi90 (approved):
Looks good now, thanks

ashwathraop (approved):

PR #2173, opened 2020-12-17
Fixing test_fairshare_usage from multisched suite

tpai26:
#### Describe Bug or Feature
This test fails because the qdel is trying to delete a job that has already exited. The job's walltime was set to 10s, so by the time the qdel is called the job had ended.
Here I am increasing the walltime of the job to 100s

[test_fairshare_usage.txt](https://github.com/openpbs/openpbs/files/5706199/test_fairshare_usage.txt)


bhroam (approved):
LGTM

arungrover (approved):

PR #2174, opened 2020-12-17
Added a null check while confirming degraded resv

arungrover:
#### Describe Bug or Feature
It has been found in valgrind reports that in some case while trying to reconfirm a degraded reservation scheduler may try to access some invalid memory.

#### Describe Your Change
The issue is very rarely seen and is not reproducible. This change, for now, makes sure that the variables are set before accessing them.

#### Link to Design Doc
NA

[log.txt](https://github.com/openpbs/openpbs/files/5706492/log.txt)


bhroam (approved):
LGTM

vstumpf (approved):
LGTM

PR #2175, opened 2020-12-17
fixing sanitizer compilation issues

birbalsain:
#### Describe Bug or Feature
Compilation error while compiling with sanitize flags. Getting size mismatch in strncpy(), malloc is replaced by rpl_malloc and realloc is replaced by rpl_realloc because "AC_FUNC_MALLOC" and "AC_FUNC_REALLOC" are mentioned in configure.ac.

#### Describe Your Change
Uses the strcpy() instead of strncpy() if whole string is going to copied and source size is not known. removed the "AC_FUNC_MALLOC" and "AC_FUNC_REALLOC" from configure.ac file to get system malloc and realloc definition.

[before_change.txt](https://github.com/openpbs/openpbs/files/5709740/before_change.txt)
Description: Tests from given sources on platforms SUSE15, CENTOS8
Platforms: SUSE15,CENTOS8
Profile: regression
|--|--|--|--|--|--|--|
|5354|1927|2|1|0|0|1924|

Description: Tests from given sources on platforms UBUNTU1804, SUSE15, UBUNTU1604, CENTOS8, RHEL8, CENTOS7, RHEL7, SLES12, SLES15
Profile: regression
|--|--|--|--|--|--|--|
|5352|3842|4|0|0|0|3838|


src/lib/Libnet/hnls.c
```diff
@@ -294,8 +294,7 @@ get_if_info(char *msg)
 		if (!head)
 			head = curr;
 		get_sa_family(listp->ifa_addr, curr->iffamily);
-		strncpy(curr->ifname, listp->ifa_name, IFNAME_MAX);
-		curr->ifname[IFNAME_MAX - 1] = '\0';
+		strcpy(curr->ifname, listp->ifa_name);
```
prakashcv13, line 297: why not increase the size of the destination buffer and continue to use strncpy

src/server/req_delete.c
```diff
@@ -307,7 +307,7 @@ issue_delete(job *pjob)
 	if (preq == NULL)
 		return;
 
-	pbs_strncpy(preq->rq_ind.rq_delete.rq_objname, pjob->ji_qs.ji_jobid, sizeof(preq->rq_ind.rq_delete.rq_objname) - 1);
+	pbs_strncpy(preq->rq_ind.rq_delete.rq_objname, pjob->ji_qs.ji_jobid, sizeof(preq->rq_ind.rq_delete.rq_objname));
```
prakashcv13, line 310: explain this change

src/server/req_manager.c
```diff
@@ -1269,10 +1269,9 @@ mgr_queue_delete(struct batch_request *preq)
 		}
 
 		if (rc == 0) {
-			char queue_name[PBS_MAXQUEUENAME];
+			char queue_name[PBS_MAXQUEUENAME + 1];
 			/* Save the queue name before we purge it so it will appear in the log.  */
 			strncpy(queue_name, pque->qu_qs.qu_name, sizeof(queue_name));
-			queue_name[(sizeof(queue_name) - 1)] = '\0';
```
prakashcv13, line 1275: why remove this line?

prakashcv13 (changes requested):

src/server/req_manager.c
birbalsain: strncpy() copy complete src data including null.

src/server/req_delete.c
birbalsain: "-Wstringop-truncation" error because null is not copied. copied data size is 1byte less than the destination size.

src/server/req_manager.c
prakashcv13: The strncpy() function is similar, except that at most n bytes of src are copied. Warning: If there is no null byte among the first n bytes of src, the string placed in dest will not be null-terminated.

reference:https://linux.die.net/man/3/strncpy

src/lib/Libnet/hnls.c
birbalsain: I agree, pbs_strncpy() is fine

birbalsain:
Description: Tests from given sources on platforms UBUNTU1804, SUSE15, UBUNTU1604, CENTOS8, RHEL8, CENTOS7, RHEL7, SLES12, SLES15
Profile: regression
|--|--|--|--|--|--|--|
|5379|3839|0|1|0|0|3838|

Description: Tests from given sources on platforms SUSE15, CENTOS8
Platforms: SUSE15,CENTOS8
Profile: regression
|--|--|--|--|--|--|--|
|5378|1926|0|2|0|0|1924|

src/server/req_manager.c
bhroam: @prakashcv13 is right, if strncpy() will leave the string unterminated if it truncates.  We have our own version of this to handle this.  Use pbs_strncpy().

src/server/req_delete.c
bhroam: I think the new code is actually more correct.  pbs_strncpy() will copy N-1 bytes and put in a null byte.  If we subtract one from N as the size, then we'll copy in N-2 bytes and N-1 will be a null.  The Nth byte will be left alone.

src/server/svr_mail.c
```diff
@@ -514,8 +513,7 @@ svr_mailownerResv(resc_resv *presv, int mailpoint, int force, char *text)
 
 		/* no mail user list, just send to owner */
 
-		(void)strncpy(mailto, presv->ri_wattr[(int)RESV_ATR_resv_owner].at_val.at_str, sizeof(mailto));
-		mailto[(sizeof(mailto) - 1)] = '\0';
+		(void)pbs_strncpy(mailto, presv->ri_wattr[(int)RESV_ATR_resv_owner].at_val.at_str, sizeof(mailto));
```
bhroam, line 516: Remove the (void)

bhroam (approved):
I'll sign off.  Removing the (void) is optional.

prakashcv13 (approved):

PR #2178, opened 2020-12-18
Multi-Server: Handling PSI variable for qsub daemon

nithinj:
#### Describe Bug or Feature
* Handling PSI variable for qsub

#### Describe Your Change
* qsub will use a named pipe to talk to the background daemon. When qsub wants to submit a job, it will check whether a background daemon is running by checking the pipe belongs to that configuration exists. If the file could not be found, qsub will go ahead and create another daemon and a new pipe for the new configuration.
* There can be more than one daemon based on configurations.
* The name of the pipe should be the unique signature of the background daemon server it is talking to.
* Since pbs_server_instances has been added to conf file we need to account that while generating the name of the pipe.
* Since the name of the pipe is restricted to 108 in Linux, a hash of the configuration string is used as the pipe name, in case the string length exceeds 108.

* Filename with small PSI
[root@centos8 test-scripts]# qsub -- /bin/sleep 10
get_comm_filename, fname: /var/tmp/pbs_default_0_____centos8,centos8_2

* File name with large PSI
[root@centos8 test-scripts]# PBS_SERVER_INSTANCES=centos8.localhost12345.localdomain:15001,centos8_2.localhost67890.localdomain:15001 qsub -- /bin/sleep 10
get_comm_filename, fname: /var/tmp/pbs_4d05ffff73ffffffff4058ff280e3b090c3bffff

* Repeating the test to see if it generates the same file name
[root@centos8 test-scripts]# PBS_SERVER_INSTANCES=centos8.localhost12345.localdomain:15001,centos8_2.localhost67890.localdomain:15001 qsub -- /bin/sleep 10
get_comm_filename, fname: /var/tmp/pbs_4d05ffff73ffffffff4058ff280e3b090c3bffff

### DIFF
--- a/src/cmds/qsub_sup.c
+++ b/src/cmds/qsub_sup.c
@@ -1415,7 +1415,7 @@ get_comm_filename(char *fname)
 	int count = 0;
 	char buf[LARGE_BUF_LEN];
 	int len;
-	char hash[SHA_DIGEST_LENGTH];
+	unsigned char hash[SHA_DIGEST_LENGTH];
 	int i;
 
 	count = snprintf(fname, MAXPIPENAME, "%s/pbs_%.16s_%lu_%.8s_%.32s_%.16s_%.5s_%s",
@@ -1518,6 +1518,8 @@ daemon_stuff(void)
 	char *err_op = "";
 	char log_buf[LOG_BUF_SIZE];
 	int cred_timeout = 0;
+	svr_conn_t **svr_conns = get_conn_svr_instances(sd_svr);
+	int i;
 
 	/* set umask so socket file created is only accessible by same user */
 	umask(cmask);
@@ -1544,7 +1546,8 @@ daemon_stuff(void)
 	}
 
 	FD_SET(bindfd, &readset);
-	FD_SET(sd_svr, &readset);
+	for (i = 0; svr_conns[i]; i++)
+		FD_SET(svr_conns[i]->sd, &readset);
 	maxfd = (bindfd > sd_svr) ? bindfd : sd_svr;
 	while (1) {
 
@@ -1577,9 +1580,11 @@ daemon_stuff(void)
 			cred_timeout = 1;
 		}
 
-		if (FD_ISSET(sd_svr, &workset)) {
-			if (recv(sd_svr, &rc, 1, MSG_OOB) < 1)
-				goto out;
+		for (i = 0; svr_conns[i]; i++) {
+			if (FD_ISSET(svr_conns[i]->sd, &workset)) {
+				if (recv(svr_conns[i]->sd, &rc, 1, MSG_OOB) < 1)
+					goto out;
+			}
 		}
 
 		/* accept the connection */
@@ -1793,5 +1798,6 @@ again:
 		/* going down, no need to free stuff */
 		close(sock);
 	}
+
 	return rc;
 }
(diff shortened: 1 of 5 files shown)


nithinj:
Requesting review from @hirenvadalia , @raviv1995 , @ashwathraop

src/lib/Libifl/pbs_loadconf.c
```diff
@@ -66,6 +66,8 @@
 
 char *pbs_conf_env = "PBS_CONF_FILE";
 
+#define LARGE_BUF_LEN 4096
```
ashwathraop, line 69: Is there a common header where you can move? You used it in cmd.h also. You could remove double definitions.

src/lib/Libifl/pbs_loadconf.c
```diff
@@ -1188,6 +1190,107 @@ __pbs_loadconf(int reload)
...
+ */
+char *
+psi_to_str(psi_t *psi, int nsvrs)
+{
+	char buf[LARGE_BUF_LEN];
+	int len = 0;
+	int i;
+
+	if (!psi)
+		return strdup("");
```
ashwathraop, line 1215: Why not return NULL?

src/lib/Libifl/pbs_loadconf.c
```diff
@@ -1188,6 +1190,107 @@ __pbs_loadconf(int reload)
...
+		}
+		fclose(fp);
+	}
+	free(conf_file);
+	
+	if (conf_val_out)
+		return conf_val_out;
+
+	/* Finally, resort to the default. */
+	return strdup("");
```
ashwathraop, line 1291: Why not return NULL?

nithinj:
@ashwathraop ,

ashwathraop (approved):

src/cmds/qsub_sup.c
```diff
@@ -1410,32 +1411,50 @@ get_comm_filename(char *fname)
...
-		);
+		count = snprintf(fname, MAXPIPENAME, "%s/pbs_", TMP_DIR);
+		len = snprintf(buf, MAXPIPENAME, "%.16s_%lu_%.8s_%.32s_%.16s_%.5s_%s",
+			 ((server_out == NULL || server_out[0] == 0) ? "default" : server_out),
+			 (unsigned long int) getuid(),
+			 cred_name,
+			 get_conf_path(),
+			 (env_svr == NULL) ? "" : env_svr,
+			 (env_port == NULL) ? "" : env_port,
+			 psi_var ? psi_var : "");
```
suresh-thelkar, line 1442: Just a question. As server and port information is already present in psi do we still need to add evn_svr and evn_port to the pipe name ?

src/lib/Libifl/pbs_loadconf.c
```diff
@@ -1188,6 +1189,107 @@ __pbs_loadconf(int reload)
...
+ * 
+ * @note
+ * return string has to be freed by the caller
+ *
+ * @return char *
+ * @retval !NULL pointer psi string
+ * @retval NULL failure
+ */
+char *
+psi_to_str(psi_t *psi, int nsvrs)
```
suresh-thelkar, line 1207: Do we need nsvrs as a separate argument? We can just call get_num_servers() right.

raviv1995 (changes requested):

src/cmds/qsub_sup.c
nithinj: pipe name should be unique based on the value in the psi.

nithinj:
@suresh-thelkar & @raviv1995 ,

raviv1995 (approved):

suresh-thelkar:
@nithinj, thanks for making the changes. It looks good to me.

nithinj:
Thanks for the approval, @ashwathraop , @raviv1995 , @suresh-thelkar 

@openpbs/openpbs-maintainers 
Please have a look at this PR.

agrawalravi90 (approved):

PR #2181, opened 2020-12-22
Test issues due to slow server response

arungrover:
#### Describe Bug or Feature
Few tests have been seen failing because of slowness in server

#### Describe Your Change
There were mainly two issues - 
- In the first issue, we try to see if a reservation that is going to start 25 seconds in future, when degraded is able to reconfirm before moving to running state. There are various steps that happen before the job moves to degraded state and if the machine is too slow, the test may never see the reservation getting reconfirmed. It rather sees the reservation moving from a degraded state to running directly. Since the test does not wait for the reservation to start, I just pushed the reservation to 125 seconds in future.
- In the second test a job is submitted to run for only 5 seconds and by the time test tries to check its state, it's already finished. I've increased the walltime to 15 seconds instead.

#### Link to Design Doc
NA

Running tests

### DIFF
--- a/test/tests/functional/pbs_reservations.py
+++ b/test/tests/functional/pbs_reservations.py
@@ -300,7 +300,7 @@ class TestReservations(TestFunctional):
         Verify that degraded standing reservations are reconfirmed
         on other nodes
         """
-        self.degraded_resv_reconfirm(start=25, end=625,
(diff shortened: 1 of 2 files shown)


nithinj (approved):
LGTM!

agrawalravi90 (approved):

bhroam (approved):
LGTM

PR #2183, opened 2020-12-22
Fix for non-resvport auth and encryption in recent tpp changes

hirenvadalia:
#### Describe Bug or Feature

* With recent TPP changes non-resvport authentication and encryption is broken due to:
  * Double free of auth handshake data
  * Sender encrypts each chunk of pkt separately before sending, while receiver tries to decrypt whole pkt's encrypted data as it has got only two chunk - one encryption header chunk and another is encrypted data chunk
  * Invalid data pointer use in `*_pkt_handler_inner()` while encryption is enabled

#### Describe Your Change

* Fixed double free by removing unwanted `free(data_out)` in `tpp_handle_auth_handshake()`
* Fixed `tpp_encrypt_pkt()` to encrypt all chunks of pkt in one shot instead encrypting each chunk separately
* Fixed `*_pkt_handler_inner()` to use correct data pointer (aka `dhdr` instead `buf`) - as after decryption `dhdr` is pointing to decrypted data while `buf` is still pointing to original encrypted data

#### Link to Design Doc

* None

* In comment

### DIFF
--- a/src/lib/Libtpp/tpp_util.c
+++ b/src/lib/Libtpp/tpp_util.c
@@ -356,7 +356,7 @@ set_tpp_config(struct pbs_config *pbs_conf, struct tpp_config *tpp_conf, char *n
 	}
 
 	tpp_log(LOG_INFO, NULL, "TPP authentication method = %s", tpp_conf->auth_config->auth_method);
-	if (tpp_conf->auth_config->encrypt_method[0] != '\0') 
+	if (tpp_conf->auth_config->encrypt_method[0] != '\0')
 		tpp_log(LOG_INFO, NULL, "TPP encryption method = %s", tpp_conf->auth_config->encrypt_method);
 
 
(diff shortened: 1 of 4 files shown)


prakashcv13 (approved):

hirenvadalia:
[all-in-one.txt](https://github.com/openpbs/openpbs/files/5734065/all-in-one.txt)
[remote-mom.txt](https://github.com/openpbs/openpbs/files/5734066/remote-mom.txt)
[remote-mom-comm.txt](https://github.com/openpbs/openpbs/files/5734067/remote-mom-comm.txt)
[remote-mom-multi-comm-hop.txt](https://github.com/openpbs/openpbs/files/5734068/remote-mom-multi-comm-hop.txt)

Regression results:
Profile: regression
|--|--|--|--|--|--|--|
|5414|3839|0|0|0|0|3839|

src/lib/Libtpp/tpp_util.c
```diff
@@ -2332,61 +2330,65 @@ tpp_encrypt_pkt(conn_auth_t *authdata, tpp_packet_t *pkt)
...
-		/* now replace the data with encrypted buffer */
-		free(chunk->data);
-		chunk->data = data_out;
-		chunk->len = len_out;
-		chunk->pos = chunk->data;
+	if (type == TPP_AUTH_CTX && data->for_encrypt == FOR_ENCRYPT)
+		return 0;
 
-		pkt->totlen += len_out;
+	buf = malloc(totlen);
```
subhasisb, line 2344: oh but we will again malloc stuff. i think we should avoid this

subhasisb (changes requested):
is there a better way rather than making one full packet from the chunks and then encrypting it? I think we should avoid the additional malloc if possible

src/lib/Libtpp/tpp_util.c
hirenvadalia: yes agree but I couldn't find any better way unless we somehow make chunks on receiver side instead full buffer... will chat offline...

subhasisb (approved):
After speaking to Hiren, I agree to his changes. Approving

PR #2184, opened 2020-12-23
Scheduler crash while confirming multiple reservations in same cycle

arungrover:
#### Describe Bug or Feature
scheduler crashes while trying to confirm a standing reservation with any other reservation in the same cycle.

#### Describe Your Change
The problem happens at two different places - 
First, the scheduler when confirms a standing reservation it creates the node info array before it creates node_spec array even though it needs node_spec array to create node info array. So order of execution was wrong there.
Second, When the scheduler adds an occurrence of a standing reservation to the actual universe, it does not update the server's all_resresv array. This results in failure to duplicate the calendar because while duplicating we try to search for reservations in the all_resresv array.

#### Link to Design Doc
NA
In progress, will post results soon.

### DIFF
--- a/test/tests/functional/pbs_reservations.py
+++ b/test/tests/functional/pbs_reservations.py
@@ -2428,3 +2428,34 @@ class TestReservations(TestFunctional):
         self.assertEqual(
             jid2[1], '(' + vn + '[1]:ncpus=4+' + vn + '[2]:ncpus=4)')
         self.assertNotIn(job1_node, job2_node, errmsg)
+
+    def test_clashing_reservations(self):
+        """
+        Test that when a standing reservation and advance reservation
+        are submitted to start at the same time on the same set of
+        resources, then the one that is submitted first wins and second
+        is deleted.
+        """
+
+        self.common_config()
+
+        a = {'scheduling': 'False'}
+        self.server.manager(MGR_CMD_SET, SERVER, a)
+
+        start = int(time.time()) + 300
+        end = int(time.time()) + 1200
+        srid = self.submit_reservation(user=PBSROOT_USER,
+                                       select='2:ncpus=4',
+                                       rrule='FREQ=DAILY;COUNT=2',
+                                       start=start,
+                                       end=end)
+
+        arid = self.submit_reservation(user=PBSROOT_USER,
+                                       select='2:ncpus=4',
+                                       start=start,
+                                       end=end)
+        a = {'scheduling': 'True'}
+        self.server.manager(MGR_CMD_SET, SERVER, a)
+        self.server.expect(RESV, {'reserve_state':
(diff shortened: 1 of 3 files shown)


test/tests/functional/pbs_reservations.py
```diff
@@ -2428,3 +2428,34 @@ def test_shared_exclusive_job_not_in_same_rsv_vnode(self):
...
+
+        arid = self.submit_reservation(user=PBSROOT_USER,
+                                       select='2:ncpus=4',
+                                       start=start,
+                                       end=end)
+        a = {'scheduling': 'True'}
+        self.server.manager(MGR_CMD_SET, SERVER, a)
+        self.server.expect(RESV, {'reserve_state':
+                                  (MATCH_RE, 'RESV_CONFIRMED|2')}, id=srid)
+        self.server.log_match(arid + ";Reservation denied", id=arid)
```
agrawalravi90, line 2461: Just a suggestion: You could use run_scheduling_cycle() to trigger a scheduling cycle, it returns after ensuring that the sched cycle was completed. Then you can call expect() and log_match() with max_attempts=1.

agrawalravi90 (approved):
LGTM, just a small suggestion

test/tests/functional/pbs_reservations.py
```diff
@@ -2428,3 +2428,34 @@ def test_shared_exclusive_job_not_in_same_rsv_vnode(self):
...
+        """
+
+        self.common_config()
+
+        a = {'scheduling': 'False'}
+        self.server.manager(MGR_CMD_SET, SERVER, a)
+
+        start = int(time.time()) + 300
+        end = int(time.time()) + 1200
+        srid = self.submit_reservation(user=PBSROOT_USER,
```
bhroam, line 2447: Why PBSROOT_USER instead of TEST_USER?

test/tests/functional/pbs_reservations.py
```diff
@@ -2428,3 +2428,34 @@ def test_shared_exclusive_job_not_in_same_rsv_vnode(self):
         self.assertEqual(
             jid2[1], '(' + vn + '[1]:ncpus=4+' + vn + '[2]:ncpus=4)')
         self.assertNotIn(job1_node, job2_node, errmsg)
+
+    def test_clashing_reservations(self):
+        """
+        Test that when a standing reservation and advance reservation
+        are submitted to start at the same time on the same set of
+        resources, then the one that is submitted first wins and second
+        is deleted.
```
bhroam, line 2437: While deleted is technically correct, I'd rather you say 'rejected'

bhroam (approved):
I'll sign off.  My 2 comments are optional.

bhroam (approved):

PR #2187, opened 2021-01-05
Handling global counts for multi-svr

nithinj:
#### Describe Bug or Feature
* Available resources should be divided between the instances when adding/removing an instance into the cluster.
* Resource assigned will be always the resources consumed by the jobs running on that instance.
* Resources assigned and available on a stat will aggregate the counts
* When local jobs are consuming more resources than available, it bumps up the resource assigned. When the resource assigned is more than the resources available means that the resources are borrowed from the global pool.

#### Describe Your Change
* IFL Stats will accumulate resources available as well as assigned values

[globals.log](https://github.com/openpbs/openpbs/files/5772184/globals.log)

### DIFF
--- a/src/lib/Libifl/int_status.c
+++ b/src/lib/Libifl/int_status.c
@@ -268,6 +268,55 @@ struct attrl_holder {
 
 /**
  * @brief
+ *	Appending attribute to the attrl_holder
+ * @param[in,out] head - head of the list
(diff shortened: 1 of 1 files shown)


subhasisb:
We need to ascertain the pros and cons of this approach and compare with other similar approaches. There is no point in keeping this PR open for a long time. Closing it for now. Please keep the code in your repo and raise after due design considerations.

PR #2188, opened 2021-01-06
Remove indexed access of attributes from objects.

ashwathraop:
#### Describe Bug or Feature

PBS source code accesses attributes array directly using indexes and any change related to attributes array becomes very hard and prone to errors. So replacing indexed access of attributes of objects by using getter/setter of the object should make it streamlined for future changes.

#### Describe Your Change

Added getter/setter for renaming objects.
Removed indexed access of attributes of the object by using getter/setter of an object.
Added getter/setter for the queue, node, server, reservation, sched object (job object already has getter/setter functions)
Update the code to use getter/setter to get/set the value of an attribute.

#### Link to Design Doc

N/A

TODO


bremanandjk (changes requested):
Started looking into a few files. Let me add further.

bremanandjk:
Please attach the test results once they are ready and all green, Thanks

src/server/hook_func.c
```diff
@@ -3150,17 +3148,14 @@ attribute_jobmap_init(job *pjob, struct attribute_jobmap *a_map)
 	}
 
 	for (index = 0; (a_index=(int)a_map[index].attr_i) >= 0; ++index) {
-		if (is_attr_set(&a_map[index].attr_val)) {
-			if (job_attr_def[a_index].at_free) {
-				job_attr_def[a_index].at_free(&a_map[index].attr_val);
-			}
-		}
+		if (is_attr_set(&a_map[index].attr_val))
+			free_attr_generic(job_attr_def, &a_map[index].attr_val, a_index);
```
agrawalravi90, line 3152: why not just call free_attr_generic -> free_attr ? I don't think we'll ever need free_attr_long, free_attr_char etc., so, i think we can just call it free_attr

src/server/jattr_get_set.c
```diff
@@ -108,10 +115,10 @@ char
 get_job_state(const job *pjob)
 {
 	if (pjob != NULL) {
-		return get_attr_c(&pjob->ji_wattr[JOB_ATR_state]);
+		return get_attr_c(get_jattr(pjob, JOB_ATR_state));
 	}
 
-	return -1;
+	return '0';
```
agrawalravi90, line 120: use JOB_STATE_LTR_UNKNOWN instead

src/server/jattr_get_set.c
```diff
@@ -174,11 +181,45 @@ char *
...
+ * @param[in]	pjob - pointer to the job
+ * @param[in]	attr_idx - index of the attribute to return
+ *
+ * @return	pbs_list_head
+ * @retval	value of attribute
+ */
+pbs_list_head
+get_jattr_list(const job *pjob, int attr_idx)
+{
+	return get_attr_list(get_jattr(pjob, attr_idx));
```
agrawalravi90, line 219: do NULL check on pjob

src/server/job_recov_db.c
```diff
@@ -237,9 +237,11 @@ job_save_db(job *pjob)
 	int rc = -1;
 	int old_mtime, old_flags;
 	char *conn_db_err = NULL;
+	attribute *mtime;
 
-	old_mtime = get_jattr_long(pjob, JOB_ATR_mtime);
-	old_flags = pjob->ji_wattr[JOB_ATR_mtime].at_flags;
+	mtime = get_jattr(pjob, JOB_ATR_mtime);
+	old_mtime = get_attr_l(mtime);
+	old_flags = mtime->at_flags;
```
agrawalravi90, line 244: id suggest using object level getters/setters over attribute level setters. You had to do this because we don't have a getter/setter for flags yet. I suggest the following:

old_mtime = get_jattr_long(pjob, JOB_ATR_mtime);
old_flags = (get_jattr(pjob, JOB_ATR_mtime))->at_flags;

src/server/job_recov_db.c
```diff
@@ -257,8 +259,8 @@ job_save_db(job *pjob)
 
 	if (rc != 0) {
 		/* revert mtime, flags update */
-		set_jattr_l_slim(pjob, JOB_ATR_mtime, old_mtime, SET);
-		pjob->ji_wattr[JOB_ATR_mtime].at_flags = old_flags;
+		set_attr_l(mtime, old_mtime, SET);
+		mtime->at_flags = old_flags;
```
agrawalravi90, line 263: Same comment here, use:

set_jattr_l_slim(pjob, JOB_ATR_mtime, old_mtime, SET);
(get_jattr(pjob, JOB_ATR_mtime))->at_flags = old_flags;

src/server/req_holdjob.c
```diff
@@ -302,7 +302,7 @@ req_releasejob(struct batch_request *preq)
 
 	old_hold = get_jattr_long(pjob, JOB_ATR_hold);
 	rc = job_attr_def[(int)JOB_ATR_hold].
-		at_set(&pjob->ji_wattr[(int)JOB_ATR_hold],
+		at_set(get_jattr(pjob, JOB_ATR_hold),
```
agrawalravi90, line 305: There shouldn't be a single at_set() call that's outside of setters. Convert this to set_jattr_generic()

src/server/req_holdjob.c
```diff
@@ -335,14 +335,14 @@ req_releasejob(struct batch_request *preq)
 				rc =
 #endif
 					job_attr_def[(int)JOB_ATR_hold].
-					at_set(&psubjob->ji_wattr[(int)JOB_ATR_hold],
+					at_set(get_jattr(psubjob, JOB_ATR_hold),
```
agrawalravi90, line 338: use set_jattr_generic

src/server/req_quejob.c
```diff
@@ -923,7 +924,7 @@ req_quejob(struct batch_request *preq)
 		 * into an advance reservation queue, the reservation's ID
 		 * gets attached later in the code
 		 */
-		job_attr_def[(int)JOB_ATR_reserve_ID].at_decode(&pj->ji_wattr[(int)JOB_ATR_reserve_ID],
+		job_attr_def[(int)JOB_ATR_reserve_ID].at_decode(get_jattr(pj, JOB_ATR_reserve_ID),
```
agrawalravi90, line 927: replace with a job level setter

src/server/req_quejob.c
```diff
@@ -1177,8 +1178,7 @@ req_quejob(struct batch_request *preq)
 
 			/* unset the old job's JOB_ATR_block */
 			set_jattr_l_slim(pjob, JOB_ATR_block, 0, SET);
-		mark_jattr_not_set(pjob, JOB_ATR_block);
-			pjob->ji_wattr[(int)JOB_ATR_block].at_flags |= ATR_MOD_MCACHE;
+			ATR_UNSET(get_jattr(pj, JOB_ATR_block));
```
agrawalravi90, line 1181: why this change? we could have just removed the ATR_MOD_MCACHE line below, was this functionally incorrect? If needed, we should change mark_jattr_not_set to call ATR_UNSET

src/server/req_quejob.c
```diff
@@ -1719,7 +1716,7 @@ req_commit_now(struct batch_request *preq,  job *pj)
 	pj->ji_qs.ji_un.ji_momt.ji_exitstat = 0;
 	if ((pj->ji_qs.ji_svrflags & (JOB_SVFLG_CHKPT|JOB_SVFLG_ChkptMig)) == 0) {
 		pj->ji_qs.ji_stime = time_now; 	/* start of walltime */
-		pj->ji_wattr[(int)JOB_ATR_stime].at_flags |= ATR_VFLAG_MODIFY;
+		set_jattr_l_slim(pj, JOB_ATR_stime, time_now, SET);
```
agrawalravi90, line 1719: I think we need to remove ji_stime from quick save area (I'm surprised i hadn't removed it myself). Otherwise, we'll end up only setting ji_stime directly somewhere outside of the setters and we will miss the time when stime was set on the job. We should remove it and replace all calls to it with calls to getters/setters for JOB_ATR_stime

src/server/req_runjob.c
```diff
@@ -584,37 +584,52 @@ req_runjob(struct batch_request *preq)
...
 		clear_attr(&sub_prev_res, &job_attr_def[JOB_ATR_resource]);
 
 		pjobsub = get_subjob_and_state(parent, get_index_from_jid(jid), NULL, NULL);
 		if (pjobsub != NULL) {
-			sub_runcount = pjobsub->ji_wattr[JOB_ATR_runcount];
-			sub_run_version = pjobsub->ji_wattr[JOB_ATR_run_version];
+			if (is_jattr_set(pjobsub, JOB_ATR_runcount))
+				set_attr_with_attr(&job_attr_def[JOB_ATR_runcount], &sub_runcount, get_jattr(pjobsub, JOB_ATR_runcount), SET);
+			if (is_jattr_set(pjobsub, JOB_ATR_run_version))
+				set_attr_with_attr(&job_attr_def[JOB_ATR_run_version], &sub_run_version, get_jattr(pjobsub, JOB_ATR_run_version), SET);
```
agrawalravi90, line 600: this also seems like it's changing logic. We were just setting pointers to the existing attributes before, now we are making a copy of it and freeing it in various places later. The code looks a lot messier now. Why did you have to do this?

src/server/stat_job.c
```diff
@@ -268,23 +268,16 @@ status_job(job *pjob, struct batch_request *preq, svrattrl *pal, pbs_list_head *
...
-		/* Note: ATR_VFLAG_MODCACHE must be set because of svr_cached() does */
-		/*		not correctly check ATR_VFLAG_SET */
+		attribute *attr = get_jattr(pjob, JOB_ATR_eligible_time);
+		/* eligible_time_enable is off so, clear set flag so that eligible_time and accrue type dont show */
+		old_elig_flags = attr->at_flags;
+		ATR_UNSET(attr);
+
+		attr = get_jattr(pjob, JOB_ATR_accrue_type);
+		old_atyp_flags = attr->at_flags;
+		ATR_UNSET(attr);
```
agrawalravi90, line 280: Use job level function mark_jattr_not_set() instead and get flags as get_jattr(pjob, JOB_ATR_accrue_type)->flags and get_jattr(pjob, JOB_ATR_eligible_time)->flags

src/server/stat_job.c
```diff
@@ -324,17 +317,15 @@ status_job(job *pjob, struct batch_request *preq, svrattrl *pal, pbs_list_head *
...
-		}
 	} else {
 		/* reset the set flags */
-		pjob->ji_wattr[(int)JOB_ATR_eligible_time].at_flags = old_elig_flags;
-		pjob->ji_wattr[(int)JOB_ATR_accrue_type].at_flags = old_atyp_flags;
+		attribute *attr = get_jattr(pjob, JOB_ATR_eligible_time);
+		old_elig_flags = attr->at_flags = old_elig_flags;
+
+		attr = get_jattr(pjob, JOB_ATR_accrue_type);
+		old_atyp_flags = attr->at_flags = old_atyp_flags;
```
agrawalravi90, line 328: Not sure I understand what this is doing, can you please clarify? Setting old_atyp_flags changes logic, I think we should avoid that.

src/server/stat_job.c
```diff
@@ -464,16 +455,14 @@ status_subjob(job *pjob, struct batch_request *preq, svrattrl *pal, int subj, pb
...
-		mark_jattr_not_set(pjob, JOB_ATR_accrue_type);
-		pjob->ji_wattr[(int)JOB_ATR_accrue_type].at_flags |= ATR_MOD_MCACHE;
+		oldeligflags = attr->at_flags;
+		ATR_UNSET(attr);
 
-		/* Note: ATR_VFLAG_MODCACHE must be set because of svr_cached() does */
-		/* 	 not correctly check ATR_VFLAG_SET */
+		attr = get_jattr(pjob, JOB_ATR_accrue_type);
+		oldatypflags = attr->at_flags;
+		ATR_UNSET(attr);
```
agrawalravi90, line 465: same comment here about using job level function mark_jattr_not_set() instead of ATR_UNSET directly

agrawalravi90 (changes requested):
reviewed only the first commit (ji_wattr), will review other objects after this since it might end up causing changes to them ..

src/server/req_runjob.c
hirenvadalia: Just answering question:
There was bug in earlier code as it was saving pointer to existing area (aka from old pjobsub->ji_attr) but after then it calls job_purge(psubjob) which frees those saved pointers

Now in new code we make copy of needed attrs then use it later so even after job_purge() will have access to old values of saved attrs.

FYI: This is job rerun scenario.

src/resmom/mom_comm.c
```diff
@@ -2513,20 +2509,18 @@ check_ms(int stream, job *pjob)
...
 resc_used(job *pjob, char *name, u_long	(*func)(resource *))
 {
-	attribute	*at;
 	resource_def	*rd;
 	resource	*pres;
 	u_long		val = 0L;
 
-	at = &pjob->ji_wattr[(int)JOB_ATR_resc_used];
-	if (at == NULL)
+	if (!is_jattr_set(pjob, JOB_ATR_resc_used))
```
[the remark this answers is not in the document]
ashwathraop: I dont think logic changed here. `is_jattr_set` returns 0 if either `pjob` or `pjob->ji_wattr` is NULL. And we used to return anyway. And isn't this better readable than before?

src/server/jattr_get_set.c
ashwathraop: `get_attr_list` will return dummy if `pjob` is NULL.

src/resmom/mom_comm.c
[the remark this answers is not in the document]
agrawalravi90: But earlier we were not checking whether JOB_ATR_resc_used is set or not. This changes the code to return from the function if it is not set, which is a change in logic, right?

src/server/req_runjob.c
agrawalravi90: Thanks for clarifying Hiren. It sounds like a memory error so I think that it would be better to fix it independently (with valgrind traces etc.) and keep this PR limited to just refactoring.

src/server/req_quejob.c
ashwathraop: This PR at its current state itself is quite huge. And as we discussed sometime back we both agree it would have been easier to have this done in small chunks. Maybe we can tackle it later with memory reduction work.

ashwathraop:
Rebased again to get recent memory leak fixes to master.

src/server/req_quejob.c
```diff
@@ -924,8 +922,7 @@ req_quejob(struct batch_request *preq)
 		 * into an advance reservation queue, the reservation's ID
 		 * gets attached later in the code
 		 */
-		job_attr_def[(int)JOB_ATR_reserve_ID].at_decode(get_jattr(pj, JOB_ATR_reserve_ID),
-			NULL, NULL, NULL);
+		set_attr_generic(get_jattr(pj, JOB_ATR_reserve_ID), &job_attr_def[(int)JOB_ATR_reserve_ID], NULL, NULL, INTERNAL);
```
agrawalravi90, line 925: can you use set_jattr_generic here instead?

src/server/stat_job.c
```diff
@@ -458,11 +454,13 @@ status_subjob(job *pjob, struct batch_request *preq, svrattrl *pal, int subj, pb
 		attribute *attr = get_jattr(pjob, JOB_ATR_eligible_time);
 
 		oldeligflags = attr->at_flags;
-		ATR_UNSET(attr);
+		mark_jattr_not_set(pjob, JOB_ATR_eligible_time);
+		attr->at_flags |= ATR_MOD_MCACHE;
```
agrawalravi90, line 458: isnt this handled inside mark_jattr_not_set()?

agrawalravi90 (changes requested):
Mostly minor comments

src/lib/Libattr/master_node_attr_def.xml
```diff
@@ -629,10 +629,10 @@
       <member_at_comp>comp_l</member_at_comp>
       <member_at_free>free_null</member_at_free>
       <member_at_action>
-#ifdef PBS_MOM
-      NULL_FUNC
-#else
+#ifndef PBS_MOM
       chk_vnode_pool
+#else
+      NULL_FUNC
```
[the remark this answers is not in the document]
ashwathraop: @bremanandjk suggested this change just to match how it's done with other attributes.

src/lib/Libpython/pbs_python_svr_internal.c
```diff
@@ -8520,22 +8487,20 @@ pbsv1mod_meth_validate_input(PyObject *self, PyObject *args, PyObject *kwds)
 				goto validate_input_error_exit;
 			}
 
-			if (job_attr_def[attr_idx].at_decode) {
```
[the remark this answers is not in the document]
ashwathraop: I did add a check.

src/resmom/mom_comm.c
```diff
@@ -1253,21 +1252,18 @@ receive_job_update(int stream, job *pjob)
...
-                                 pjob->ji_qs.ji_jobid, log_buffer);
+				pjob->ji_qs.ji_jobid, log_buffer);
 			continue;
 		}
 
-		/* decode attribute */
-		pdef = &job_attr_def[index];
-		errcode = pdef->at_decode(get_jattr(pjob, index), psatl->al_name,
-                                         psatl->al_resc, psatl->al_value);
+		errcode = set_jattr_generic(pjob, index, psatl->al_value, psatl->al_resc, INTERNAL);;
```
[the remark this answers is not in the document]
ashwathraop: Woah, I am surprised the compiler was okay with this.

src/server/jattr_get_set.c
```diff
@@ -449,8 +449,10 @@ is_jattr_set(const job *pjob, int attr_idx)
 void
 mark_jattr_not_set(job *pjob, int attr_idx)
 {
-	if (pjob != NULL)
-		(get_jattr(pjob, attr_idx))->at_flags &= ~ATR_VFLAG_SET;
+	if (pjob != NULL) {
+		attribute *attr = get_jattr(pjob, attr_idx);
+		attr->at_flags = (attr->at_flags & ~ATR_VFLAG_SET) | ATR_MOD_MCACHE;
```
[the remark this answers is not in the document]
ashwathraop: I added this so that I can use mark_jattr_not_set() as a replacement for ATR_UNSET.

src/server/req_quejob.c
ashwathraop: Well no. set_jattr_generic() will not go further since the value is NULL.

src/server/stat_job.c
ashwathraop: Yes.

src/lib/Libpython/pbs_python_svr_internal.c
```diff
@@ -3365,6 +3367,8 @@ _pps_helper_get_queue(pbs_queue *pque, const char *que_name, char *perf_label)
 		log_err(PBSE_INTERNAL, __func__,
 			"partially populated python queue object");
 	}
+	if (is_attr_set(qattr))
+		free_attr(que_attr_def, qattr, QA_ATR_JobsByState);
```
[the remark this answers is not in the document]
ashwathraop: Yes. No need to check. And qattr will get malloced again anyway plus Valgrind complained about this.

src/lib/Libattr/attr_func.c
```diff
@@ -1803,19 +1823,47 @@ decode_project(attribute *patr, char *name, char *rescn, char *val)
...
+ *	Generic function to return the ponter to the attribute.
+ *	Use of this function only within object getter functions.
+ *
+ * @param[in]	list		- Pointer to object's attribute list
+ * @param[in]	attr_idx	- Index of the attribute to be freed.
+ *
+ */
 attribute *
-get_attr_generic(attribute *list, int attr_idx)
+_get_attr_generic(attribute *list, int attr_idx)
```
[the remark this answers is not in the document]
ashwathraop: Yes that sounds better.

src/lib/Libattr/attr_fn_c.c
```diff
@@ -99,7 +99,7 @@ decode_c(attribute *patr, char *name, char *rescn, char *val)
 		ATR_UNSET(patr);
 		patr->at_val.at_char = '\0';
 	} else {
-		patr->at_flags |= ATR_SET_MOD_MCACHE;
+		post_attr_set(patr);
```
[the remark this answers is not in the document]
ashwathraop: Inside Libattr, function attr_atomic_set() still goes through each attribute and calls decode function. So maybe we still need it. No?

src/server/jattr_get_set.c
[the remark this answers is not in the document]
agrawalravi90: Why not just call ATR_UNSET here?

agrawalravi90 (approved):
LGTM, thanks for making all the changes

ashwathraop:
Marking WIP for the test results

bremanandjk:
Please attach the latest test results.

ashwathraop:
Profile: regression
|--|--|--|--|--|--|--|
|5946|3815|2|1|0|1|3811|

Rerun:
Description: Rerun Only Failed Tests From #5946
Profile: regression
|--|--|--|--|--|--|--|
|5947|3|0|0|0|0|3|

bremanandjk (approved):
LGTM

ashwathraop:
[valgrind.log](https://github.com/openpbs/openpbs/files/5923007/valgrind.log)
Test with Valgrind

PR #2190, opened 2021-01-07
Fix nd_index duplication and mom wrong vnode map (and nodefile) due to mismatch in execvnode and exechost2 parsing

Jitendra-Rajak:
#### Describe Bug or Feature
1. nd_index stored in the database is computed based on total number of nodes, which is dynamic. Thus, the computation is wrong. This results in nodes loaded in the wrong order on server restart and other problems occur after that.

2. When using exclhost with qsub (similar to follwing), the nodefile contains wrong entries.

3. Mom dumps core on running TestPbsHookAlarmLargeMultinodeJob.

#### Describe Your Change
1. This was working fine when upon any change to any node, we used to save ALL nodes to the database, in which case the nd_index would get recomputed and saved for every node. However, saving every node does not scale well. So we had changed to save only the node that has been updated, and this exposed the issue with nd_index (of the other nodes).
Use nd_creattm (the time of creation of the vnode) as the primary means of sorting the nodes. However, in the case of Cray Xt, parent moms of vnodes having multiple parents must be loaded before such vnodes are loaded from the database (on server restart)

2. The server does not send vnodemap to moms any more. The server instead sends the exec_host2 string (along with exec_vnode and schedselect) so that the mom can deduce the information by herself. However, the mom caches the information in the vnode table after she deduces the first time (for any vnode to host map).
Move the exechost2 parsing code to a separate function and call it also from the place where to skip a vnode in execvnode. This ensures that the corresponding exechost2 entry is also skipped

3. we use variable number of args for log_eventf()
if we pass larger buffers to log_eventf()
vsnprintf() is called twice
in the first call we get to know that the internal buffer is not sufficient
we malloc another buffer and call vsnprintf() again -
we pass the same variable args list to the second call as well -
however - this list was already scanned by the first call
So, we just need to reset the list before the second call

|--|--|--|--|--|--|--|
|5484|3850|1|0|0|1|3848|

### DIFF
--- a/src/lib/Liblog/log_event.c
+++ b/src/lib/Liblog/log_event.c
@@ -144,6 +144,7 @@ void
 log_eventf(int eventtype, int objclass, int sev, const char *objname, const char *fmt, ...)
 {
 	va_list args;
+	va_list args_copy;
 	int len;
 	char logbuf[LOG_BUF_SIZE];
 	char *buf;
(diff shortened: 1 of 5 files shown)


Jitendra-Rajak:
Logs has been attached to the ticket

prakashcv13 (approved):

saimanikant (approved):

PR #2192, opened 2021-01-07
qsub should distribute jobs uniformly amongst server instances

nithinj:
#### Describe Bug or Feature
qsub daemon was distributing jobs randomly which could result in skewed submissions

#### Describe Your Change
Instead of random, pbs_submit will pick the next available instance from the list based on the last used instance. The first job will be still going to a random server in the list. All other server instances will be attempted before we reach back at this instance again.

Before:
```
[root@centos8 pbspro]# for i in {1..6}; do qsub -- /bin/sleep 100; done
2018.centos8
2019.centos8
13.centos8_2
2020.centos8
2021.centos8
2022.centos8
```
After:
```
[root@centos8 pbspro]# for i in {1..6}; do qsub -- /bin/sleep 100; done
29.centos8_2
60.centos8
30.centos8_2
61.centos8
31.centos8_2
62.centos8
````

### DIFF
--- a/src/include/libpbs.h
+++ b/src/include/libpbs.h
@@ -407,6 +407,7 @@ void *get_conn_svr_instances(int);
 int pbs_register_sched(const char *sched_id, int primary_conn_id, int secondary_conn_id);
 int get_svr_inst_fd(int vfd, char *svr_inst_id);
 int random_srv_conn(svr_conn_t **);
+int get_available_conn(svr_conn_t **);
 int starting_index(char *);
 char *PBS_get_server(char *, char *, uint *);
 int encode_DIS_JobsList(int sock, char **jobs_list, int numofjobs);
--- a/src/lib/Libifl/ifl_util.c
+++ b/src/lib/Libifl/ifl_util.c
@@ -99,7 +99,7 @@ PBS_get_server(char *server_id_in, char *server_name_out, unsigned int *port)
  * @retval -1: error
  * @retval != -1 fd corresponding to the connection
  */
-static int
+int
 get_available_conn(svr_conn_t **svr_conns)
 {
 	int i;
--- a/src/lib/Libifl/pbsD_submit.c
+++ b/src/lib/Libifl/pbsD_submit.c
@@ -67,6 +67,37 @@ struct cred_info {
 
 /**
  * @brief
+ *	get the next connection from the connection list based on the
+ *	last used value
+ *
+ * @param[in] c - connection fd
+ *
+ * @return int
+ * @retval -1: error
+ * @retval != -1 fd corresponding to the connection
+ */
+static int
+get_nxt_conn(int c)
+{
+	svr_conn_t **svr_conns = get_conn_svr_instances(c);
+	static int ind = -1;
+
+	if (!svr_conns || svr_conns[0]->sd == c)
+		return c;
+
+	if (ind == -1)
+		ind = rand_num() % get_num_servers();
+	else
+		ind = (ind + 1) % get_num_servers();
+
+	if (svr_conns[ind] && svr_conns[ind]->state == SVR_CONN_STATE_UP)
+		return svr_conns[ind]->sd;
+
+	return get_available_conn(svr_conns);
+}
+
+/**
+ * @brief
  *	-wrapper function for pbs_submit where submission takes credentials.
  *
  * @param[in] c - communication handle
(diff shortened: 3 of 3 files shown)


nithinj:
Requesting reviews from @suresh-thelkar

agrawalravi90 (changes requested):
I'm not sure that this is any better than random, if not worse. This will only work if all jobs are being submitted from the same qsub daemon process. In practice, you'll have multiple users submitting jobs from their own terminals. In the worst case, if all users submit jobs in lock step, you'll get all jobs submitted to only one server at a time (all N users will choose server 1 first and submit N jobs to server 1, then N jobs to server 2 and so on).

arungrover:
Any route you choose, please also take care of job dependency. Dependent jobs should ideally go to the same server.

agrawalravi90:
> Any route you choose, please also take care of job dependency. Dependent jobs should ideally go to the same server.

We've not added support for job dependencies yet for multi-server

arungrover:
> > Any route you choose, please also take care of job dependency. Dependent jobs should ideally go to the same server.
> 
> 
> 
> We've not added support for job dependencies yet for multi-server 

Ok, do we know how we are going to address job dependency?

agrawalravi90:
> Ok, do we know how we are going to address job dependency?

Not yet, it has not been designed yet

nithinj:
@agrawalravi90 ,
The first job being submitted from a qsub daemon will still be to a random one in the list. Then it tries to walk through the list till it completes it. I will mention it in the description.

agrawalravi90:
@nithinj thanks for clarifying. I still think that it's not really any better than the simpler randomizer that we have right now since it still will not guarantee uniform distribution. Can you do some real world tests to see how much better this is? Have distinct shell sessions submit 10k jobs before and after this change and see how the jobs get distributed between multiple servers.

suresh-thelkar:
@nithinj, The changes look good to me. As @agrawalravi90 said we can actually do some multi-server tests to know whether the distribution of jobs is fair.

agrawalravi90:
@nithinj on second thoughts, can you please do the following test instead:
- have separate qsub daemon process submit jobs after a random sleep time. This will imitate real world behavior better.
- run the test for a fixed amount of time, say 1 minute, and then measure how many jobs each server got.

nithinj:
@agrawalravi90 ,
I have updated the description with examples of submitting jobs before and after the changes.

For the sake of explanation, let's say we have 2 servers and 10 jobs. To start with we have one terminal where qsub is getting submitted. If it is purely random, in the worst case it could get distributed 10 vs 0. The best case is fair distribution 5 vs 5, which will be always guaranteed by the proposed approach.

Now you can interpolate this into N terminals. We can think optimistically that one skewed result might get offset by another terminal and so on. But it can be the other way, making the distribution even worse. OTOH, the proposed approach guarantees a fair distribution with each terminal, dividing the overall job distribution equally amongst instances.

agrawalravi90:
@nithinj I understand how the logic will work, what I'm trying to convey is that this will not guarantee uniform distribution of jobs across the server instances, so I don't like the PR title, it seems misleading. If you think that round robin is actually better than random, even though neither will guarantee uniform distribution, then please do a realistic test to show that it will be better: have multiple qsub processes submit jobs at different frequencies for a given time period and measure what the distribution was.

If round robin performs better, then of course we should do it, I'm just asking for a proper test.

nithinj:
@agrawalravi90 ,
PR doesn't claim uniform load across instances but uniform no of jobs, both are different.
I have shown how a single qsub will result in a skewed result. We have discussed what will happen if you interpolate it. It can get worse but cannot beat the proposed approach. More testing will prove how worse it can get. But I do not see much value in doing it.

agrawalravi90:
> PR doesn't claim uniform load across instances but uniform no of jobs, both are different.

That's what i said, this PR claims that jobs will be distributed uniformly across servers, which is incorrect.

> I have shown how a single qsub will result in a skewed result. We have discussed what will happen if you interpolate it. It can get better/worse but cannot beat the proposed approach. More testing will prove how better/worse it can get. But I do not see much value in doing so.

Since this change won't guarantee uniform distribution of jobs, the only potential gain from it is a "better" distribution of jobs than the current approach, so you need to show how it is better. Real sites don't have a single qsub process, so showing that a single qsub process performs better is not a good test. It's like if I created an alternative to TPP and only showed that my algorithm performs better for a single host, it's not a good test, just like yours isn't. If you still disagree, please feel free to get opinions from other maintainers, if they agree with you then I'll back off.

subhasisb:
We need to ascertain the pros and cons of this approach and compare with other similar approaches. There is no point in keeping this PR open for a long time. Closing it for now. Please keep the code in your repo and raise after due design considerations.

PR #2196, opened 2021-01-12
Redesigning jobid namespace for multi-server

agrawalravi90:
#### Describe Bug or Feature

Post multi-server, there will be multiple servers independently generating jobids. In order for these jobids to not collide, we need to relax the existing jobid namespace requirements to allow each server to generate unique jobids (explained more in the design). This PR adds ability to each server to generate unique jobids so that they don't collide using the following:

- each server will add its unique PBS_SERVER_INSTANCES index as the last 2 digits of the jobid

This allows upto 100 servers (00 to 99) to be configured at a time.

**Since the algorithm may change in the future, users are expected to treat jobids as opaque strings and not expect them to be in the above, or any other format (see design), at least for the purposes of multi-server**. Single server jobids will continue to look the same as before.

#### Link to Design Doc

https://openpbs.atlassian.net/wiki/spaces/PD/pages/2320924673/jobid+namespace+redesign+for+multi-server

Single server:
```
[ravi@pbsc8 ~]$ ps -ef | grep pbs_server
root      3620     1  0 08:41 ?        00:00:00 /opt/pbs/sbin/pbs_server.bin
ravi      3632    34  0 08:42 pts/0    00:00:00 grep --color=auto pbs_server
[ravi@pbsc8 ~]$ qsub -- /bin/sleep 1000
2014.pbsc8
[ravi@pbsc8 ~]$ qsub -- /bin/sleep 1000
2015.pbsc8
[ravi@pbsc8 ~]$ qsub -- /bin/sleep 1000
2016.pbsc8
[ravi@pbsc8 ~]$ qsub -- /bin/sleep 1000
2017.pbsc8
[ravi@pbsc8 ~]$ qsub -- /bin/sleep 1000
2018.pbsc8
```

2 servers:
```
[ravi@pbsc8 ~]$ qsub -- /bin/sleep 100
201001.pbsc8
[ravi@pbsc8 ~]$ qsub -- /bin/sleep 100
200900.pbsc8
[ravi@pbsc8 ~]$ qsub -- /bin/sleep 100
201000.pbsc8
[ravi@pbsc8 ~]$ qsub -- /bin/sleep 100
201101.pbsc8
[ravi@pbsc8 ~]$ qsub -- /bin/sleep 100
201100.pbsc8
[ravi@pbsc8 ~]$ qsub -- /bin/sleep 100
201200.pbsc8
[ravi@pbsc8 ~]$ qsub -- /bin/sleep 100
201201.pbsc8
[ravi@pbsc8 ~]$ qsub -- /bin/sleep 100
201300.pbsc8
[ravi@pbsc8 ~]$ qsub -- /bin/sleep 100
201301.pbsc8
[ravi@pbsc8 ~]$ qstat
201001.pbsc8      STDIN            ravi              00:00:00 R workq           
200900.pbsc8      STDIN            ravi              00:00:00 R workq           
201000.pbsc8      STDIN            ravi              00:00:00 R workq           
201101.pbsc8      STDIN            ravi              00:00:00 R workq           
201100.pbsc8      STDIN            ravi              00:00:00 R workq           
201200.pbsc8      STDIN            ravi              00:00:00 R workq           
201201.pbsc8      STDIN            ravi              00:00:00 R workq           
201300.pbsc8      STDIN            ravi              00:00:00 R workq           
201301.pbsc8      STDIN            ravi                     0 Q workq  
```

**Regression**
Description: Rerun All Tests From #5671
Profile: regression
|--|--|--|--|--|--|--|
|5672|3854|3|0|0|18|3833|

Description: Rerun Only Failed Tests From #5672
Profile: regression
|--|--|--|--|--|--|--|
|5675|3|0|0|0|0|3|

### DIFF
--- a/src/include/libpbs.h
+++ b/src/include/libpbs.h
@@ -75,6 +75,8 @@ extern "C" {
 #define EOF (-1)
 #endif
 
+#define MSVR_JID_NCHARS_SVR	2	/* No. of chars reserved for svr instance in job ids for multi-server */
+
 /* enums for standard job files */
 enum job_file {
 	JScript,
@@ -405,7 +407,7 @@ svr_conn_t **get_conn_svr_instances(int);
 int pbs_register_sched(const char *sched_id, int primary_conn_id, int secondary_conn_id);
 int get_svr_inst_fd(int vfd, char *svr_inst_id);
 int random_srv_conn(svr_conn_t **);
-int starting_index(char *);
+int get_job_location_hint(char *);
 char *PBS_get_server(char *, char *, uint *);
 int encode_DIS_JobsList(int sock, char **jobs_list, int numofjobs);
 int get_server_fd_from_jid(int c, char *jobid);
--- a/src/include/libpbs.h
+++ b/src/include/libpbs.h
@@ -75,8 +75,6 @@ extern "C" {
 #define EOF (-1)
 #endif
 
-#define MSVR_JID_NCHARS_SVR	2	/* No. of chars reserved for svr instance in job ids for multi-server */
-
 /* enums for standard job files */
 enum job_file {
 	JScript,
--- a/src/lib/Libifl/pbs_loadconf.c
+++ b/src/lib/Libifl/pbs_loadconf.c
@@ -275,7 +275,7 @@ parse_psi(char *conf_value)
 	char **list;
 	int i;
 	char *svrname = NULL;
-	
+
 	free(pbs_conf.psi);
 
 	if (conf_value == NULL)
@@ -308,12 +308,8 @@ parse_psi(char *conf_value)
         			get_fullhostname(pbs_conf.psi[i].name, pbs_conf.psi[i].name, PBS_MAXHOSTNAME);
 		}
 
-		if (pbs_conf.psi[i].port == 0) {
-			if (is_same_host(pbs_conf.psi[i].name, pbs_conf.pbs_server_name))
-				pbs_conf.psi[i].port = pbs_conf.batch_service_port;
-			else
-				pbs_conf.psi[i].port = PBS_BATCH_SERVICE_PORT;
-		}
+		if (pbs_conf.psi[i].port == 0)
+			pbs_conf.psi[i].port = PBS_BATCH_SERVICE_PORT;
 	}
 	free_string_array(list);
 	pbs_conf.pbs_num_servers = i;
@@ -1368,4 +1364,3 @@ get_num_servers(void)
 {
 	return pbs_conf.pbs_num_servers;
 }
-
--- a/src/server/req_quejob.c
+++ b/src/server/req_quejob.c
@@ -124,6 +124,8 @@ extern char *path_spool;
 extern struct server server;
 extern struct attribute attr_jobscript_max_size;
 extern char  server_name[];
+extern char server_host[];
+extern unsigned int pbs_server_port_dis;
 extern char *resc_in_err;
 #endif	/* PBS_MOM */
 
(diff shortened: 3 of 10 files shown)


src/include/libpbs.h
```diff
@@ -405,7 +407,7 @@ void *get_conn_svr_instances(int);
 int pbs_register_sched(const char *sched_id, int primary_conn_id, int secondary_conn_id);
 int get_svr_inst_fd(int vfd, char *svr_inst_id);
 int random_srv_conn(svr_conn_t **);
-int starting_index(char *);
+int get_svr_idx_from_jid(char *);
```
subhasisb, line 410: I would rather call this function as get_job_location_hint() or something similar. In other words, I suggest we not give away the implementation in the function name. All we are saying is that this will take a void * and return some kind of a server location hint - which may or may not be correct - and we might have to loop over servers.

src/server/req_quejob.c
```diff
@@ -241,6 +243,74 @@ validate_perm_res_in_select(char *val, int val_exist)
...
+ * @param[out]	jidbuf - buffer to fill job id in
+ * @param[in]	clusterid - cluster name (PBS_SERVER)
+ * @param[in]	jtype - job type (0 for normal job, 1 for job array)
+ *
+ * @return	int
+ * @retval	0 for Success
+ * @retval	1 for Failure
+ */
+static int
+generate_jobid(char *jidbuf, char *clusterid, int jtype)
```
subhasisb, line 283: since this is intended to be a switchable function, is it better to move it to a separate/new code file so that it can be easily patched or moved to a shared library in future? Or, should we make this a shared library to start with?

subhasisb (approved):
More like question...and a minor suggestion

subhasisb (changes requested):
one minor comment...

subhasisb:
You also need to attach test results and valgrind logs

src/include/libpbs.h
agrawalravi90: I've renamed the function, but I'd like to avoid void pointers, I can't really foresee this function ever accepting a type other than char *. We can change the signature if/when needed. What do you think?

src/server/req_quejob.c
agrawalravi90: I hesitate doing this preemptively without knowing what the future holds. Do we want to design for different sites wanting different jobid generation schemes now? Maybe it'll get designed as a configuration (qmgr/pbs.conf) switch instead of an installation time decision (via shared libraries)? So, I feel like we should wait for further feedback. What do you think?

src/include/libpbs.h
agrawalravi90: how about "get_expected_svr_idx" or "get_original_svr_idx" or "get_probable_svr_idx"? get_..._hint() makes it sound like the function is returning a hint, where it is actually returning the potential server index

src/server/req_quejob.c
subhasisb: I agree - nevertheless just wanted to get u started to think about this :) Let's leave it as is for now

src/include/libpbs.h
subhasisb: Well i was trying to intentionally keep the function name vague/abstract, so that future implementations could use it more like an hint - 

potential server index = hint --> i think both are the same things (hint being hint about the server index) but if you like the word probable more, i am fine with it...

agrawalravi90: ok, i understand your POV now, I'll make it "get_job_location_hint"

bhroam: Just for a thought way down the road, void *'s are a little bit of a pain in C++.  You have to static_cast them at the use.
(e.g. static_cast<char *>(malloc())).  We're not converting the server to C++ any time soon, but it is something to think of.

agrawalravi90: im just going to rename, not convert the argument to void *, we can do it if/when needed to support multiple different naming schemes, I'd like to wait until we design that.

subhasisb (approved):

src/lib/Libifl/pbs_loadconf.c
```diff
@@ -308,12 +308,8 @@ parse_psi(char *conf_value)
...
 		}
 
-		if (pbs_conf.psi[i].port == 0) {
-			if (is_same_host(pbs_conf.psi[i].name, pbs_conf.pbs_server_name))
-				pbs_conf.psi[i].port = pbs_conf.batch_service_port;
-			else
-				pbs_conf.psi[i].port = PBS_BATCH_SERVICE_PORT;
-		}
+		if (pbs_conf.psi[i].port == 0)
+			pbs_conf.psi[i].port = PBS_BATCH_SERVICE_PORT;
```
ashwathraop, line 312: Can you explain this change to me. I do not follow. Why you removed the if block?

agrawalravi90: The problem with the old code is this: when a server doesn't have port number in PSI, each server sets the port number to  pbs_conf.batch_service_port. So, let's say **PBS_SERVER_INSTANCES=pbs, pbs-2:18002** then, pbs_conf.batch_service_port for server 1 will be set to 15001 and pbs_conf.batch_service_port for server 2 will be set to 18002. So, when server 2 reads PSI, it will see that 'pbs' has no port number, so it will set it to its pbs_conf.batch_service_port, which will be 18002, not 15001.

In general, when a port number is absent, we should assume the default of 15001, that's why I changed it to do just that. Does it make sense?

ashwathraop: It does. Thanks.

ashwathraop (approved):

PR #2203, opened 2021-01-16
Interleave the qdel request with larger job array to handle others requests.

bremanandjk:
#### Describe Bug or Feature

Server quasihangs when large job arrays with many running sub-jobs are deleted.

#### Describe Your Change

* Introduced a new type of work task namely task_list_interleave. 
* After the fixed duration of time (say 5 sec), the qdel request would create this work task and allow the server to handle other requests. 
* In case no other pending requests are available on the server, qdel request would resume finishing the deletion operation.

After change: 
Server logs:

01/15/2021 23:29:10.343414;0008;Server@pbs-server;Job;3[].pbs-server;Job to be deleted at request of root@pbs-server.pbspro.org
01/15/2021 23:29:10.343435;0100;Server@pbs-server;Job;3[].pbs-server;in free_nodes and no exec_vnode
01/15/2021 23:29:10.344232;0100;Server@pbs-server;Job;3[].pbs-server;dequeuing from workq, state E
01/15/2021 23:29:10.345455;0100;Server@pbs-server;Req;;Type 51 request received from Scheduler@pbs-server.pbspro.org, sock=16
01/15/2021 23:29:10.345611;0100;Server@pbs-server;Req;;Type 19 request received from pbsuser@pbs-server.pbspro.org, sock=20
01/15/2021 23:29:10.346027;0040;Server@pbs-server;Svr;pbs-server;Scheduler sent command 2
01/15/2021 23:29:10.346775;0100;Server@pbs-server;Req;;Type 21 request received from Scheduler@pbs-server.pbspro.org, soc

We can see the other user requests are served while qdel request is processing.

Profile: regression
|--|--|--|--|--|--|--|
|5625|3849|2|0|0|1|3846|

### DIFF
--- a/src/lib/Libutil/work_task.c
+++ b/src/lib/Libutil/work_task.c
@@ -58,7 +58,7 @@
 /* Global Data Items: */
 
 extern pbs_list_head task_list_immed; /* list of tasks that can execute now */
-extern pbs_list_head task_list_interleave;
+extern pbs_list_head task_list_interleave; /* list of tasks that can execute after interleaving other tasks */
 extern pbs_list_head task_list_timed; /* list of tasks that have set start times */
 extern pbs_list_head task_list_event; /* list of tasks responding to an event */
 extern int svr_delay_entry;
--- a/src/lib/Libutil/work_task.c
+++ b/src/lib/Libutil/work_task.c
@@ -58,6 +58,7 @@
 /* Global Data Items: */
 
 extern pbs_list_head task_list_immed; /* list of tasks that can execute now */
+extern pbs_list_head task_list_interleave;
 extern pbs_list_head task_list_timed; /* list of tasks that have set start times */
 extern pbs_list_head task_list_event; /* list of tasks responding to an event */
 extern int svr_delay_entry;
@@ -104,6 +105,8 @@ struct work_task *set_task(enum work_type type, long event_id, void (*func)(stru
 
 	if (type == WORK_Immed)
 		append_link(&task_list_immed, &pnew->wt_linkevent, pnew);
+	else if (type == WORK_Interleave)
+		append_link(&task_list_interleave, &pnew->wt_linkevent, pnew);
 	else if (type == WORK_Timed) {
 		pold = (struct work_task *)GET_NEXT(task_list_timed);
 		while (pold) {
@@ -360,6 +363,7 @@ default_next_task(void)
 	time_t		   delay;
 	struct work_task  *nxt;
 	struct work_task  *ptask;
+	struct work_task  *last_interleave_task;
 	/*
 	 * tilwhen is the basic "idle" time if there is nothing pending sooner
 	 * for the Server (timed-events, call scheduler, IO)
@@ -384,6 +388,19 @@ default_next_task(void)
 
 	while ((ptask=(struct work_task *)GET_NEXT(task_list_immed)) != NULL)
 		dispatch_task(ptask);
+		
+	last_interleave_task = (struct work_task *) GET_PRIOR(task_list_interleave);
+	while ((ptask=(struct work_task *)GET_NEXT(task_list_interleave)) != NULL) {
+		dispatch_task(ptask);
+		if (ptask == last_interleave_task)
+			break;
+	}
+
+	if (GET_NEXT(task_list_interleave)) {
+		/* more tasks waiting, wait least */
+		tilwhen = 0;
+	}
+
 
 	while ((ptask=(struct work_task *)GET_NEXT(task_list_timed))!=NULL) {
 		if ((delay = ptask->wt_event - time_now) > 0) {
--- a/src/server/req_delete.c
+++ b/src/server/req_delete.c
@@ -80,7 +80,7 @@
 #include "pbs_nodes.h"
 #include "svrfunc.h"
 
-#define QDEL_DURATION 5
+#define QDEL_BREAKER_SECS 5
 
 /* Global Data Items: */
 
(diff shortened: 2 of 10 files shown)


bhroam:
Deleting job history does something similar to this.  You might consider merging the two methods

bremanandjk:
Thanks @bhroam for your review. 

Yeah, I have seen the history job deletion method. While setting the server attribute "ATTR_JobHistoryEnable", the corresponding action function would set the Timed work task with a 2 min delay and in each execution of this registered function, it resets the Timed work task. 

Even though "qdel resume" method and "history job deletion" looks similar, I see the following differences 

1) History job deletion method would delete the job from the server memory by iterate over the svr_alljobs list,
2) No batch request associated with it, so no need of maintaining a reply structure. and
3) It is fine to have the timed work task for 2 min since no client is waiting over TCP for a reply. 

In the case of Qdel interleave-work-task, the server gets an opportunity to process another client request in between the execution of delete operation by preserving the reply structures in the work task entry and if none of the client requests are pending to attend, immediately the server works on resuming the deletion functionality since we don't want to make the client wait for the reply. 

I couldn't find any common code between these 2 methods, plz let me know I am wrong here. Thanks

bhroam:
Thanks for explaining @bremanandjk.  They do sound different mainly because of the urgency of the qdel.  By default the job_history_duration is 7 days.  If a job sticks around for a 7 days + a few minutes it really doesn't matter.  When you have a client waiting for a response, then it matters.

src/server/req_delete.c
```diff
@@ -416,7 +449,14 @@ req_deletejob(struct batch_request *preq)
 	else
 		qdel_mail = 1;
 
-	for (j = 0; j < count; j++) {
+	if (preq->rq_ind.rq_deletejoblist.rq_resume != 1) {
+		preply->brp_un.brp_deletejoblist.tot_jobs = count;
+		preply->brp_un.brp_deletejoblist.tot_arr_jobs = 0;
+		preply->brp_un.brp_deletejoblist.tot_rpys = 0;	
+	} else 
+		start_jobid = preq->rq_ind.rq_deletejoblist.jobid_to_resume;
```
subhasisb, line 457: What happens if the start_jobid (in case of a resume) was deleted by another qdel request (or by job getting finished) by the time you resume the deletion?

src/server/req_delete.c
```diff
@@ -515,10 +557,22 @@ req_deletejob(struct batch_request *preq)
...
+			
+			if (preq->rq_ind.rq_deletejoblist.rq_resume)
+				start = preq->rq_ind.rq_deletejoblist.subjobid_to_resume;
 
 			/* keep the array from being removed while we are looking at it */
 			parent->ji_ajinfo->tkm_flags |= TKMFLG_NO_DELETE;
-			for (i = parent->ji_ajinfo->tkm_start; i <= parent->ji_ajinfo->tkm_end; i += parent->ji_ajinfo->tkm_step) {
+			for (i = start; i <= parent->ji_ajinfo->tkm_end; i += parent->ji_ajinfo->tkm_step) {
+				end_time = time(NULL);
+				if ((end_time - begin_time) > 5) {
```
subhasisb, line 568: please do not hardcode 5 - use a macro

subhasisb (changes requested):

src/server/req_delete.c
bremanandjk: I have tried to simulate the behaviour in GDB, 
1) Put the breakpoint at the subsequent call of req_deletejob(). (resume request from work task)
2) In GDB,  removed the job object from which(start_jobid) the request is going to resume. (Similar to the situation job deleted by client or job getting finished)
3) While continuing gdb, the chk_job_request() function would return NULL since the job object is not available in the system and update the error code. 
4) Qdel client would get the response as UNKNOWN JOBID for that removed one.

src/server/req_delete.c
bremanandjk: Thanks, I have added macro for this duration.

src/server/req_delete.c
subhasisb: ok that sounds like it would work

subhasisb (approved):

src/lib/Libutil/work_task.c
```diff
@@ -58,6 +58,7 @@
 /* Global Data Items: */
 
 extern pbs_list_head task_list_immed; /* list of tasks that can execute now */
+extern pbs_list_head task_list_interleave;
```
ashwathraop, line 61: It's good to be consistent with other externs and have a comment.

src/server/req_delete.c
```diff
@@ -80,6 +80,7 @@
 #include "pbs_nodes.h"
 #include "svrfunc.h"
 
+#define QDEL_DURATION 5
```
ashwathraop, line 83: Add a comment here too or have a better name that explains the need for this variable. Something on the lines of QDEL_BREAKER_SECS would be better.

ashwathraop (commented):
Small nits and rest look good.

ashwathraop:
Just to be safe add Valgrind results too.

bremanandjk:
Valgrind logs: 
[valgrind.log](https://github.com/openpbs/openpbs/files/5849747/valgrind.log)
No new errors/leaks from the changed code.

ashwathraop (approved):

bremanandjk:
@subhasisb  @ashwathraop Thanks for your approvals, 
It's ready to merge, please take it further.

PR #2214, opened 2021-01-22
Fix for runone dependency crash

arungrover:
#### Describe Bug or Feature
Server crashes while trying to get rid of a job that has runone dependency on it.

#### Describe Your Change
The problem does not show itself unless there are 2 NIC cards on the server machine. The problem happens on a machine with 2 NICs because while trying to delete a job server tries to update all other dependent jobs. When a dependent job is deleted, to update other jobs, the server usually issues a local batch request which updates other dependent jobs and then deletes the job in hand.
Now because of NICs server thinks that the server it is trying to send the batch request is remote and it sends a batch request over the network and deletes the job at hand. Now when the same server processes the batch request it had sent to itself, it tries to lookup the deleted job and crashes because the job is not there in the system.
 - I added a NULL check to make sure the server does not crash.
 - Removed runone dependency from the dependency list as soon as it is released.
 - Made a change in issue_to_svr to check all interfaces before deciding it's not a local request.

#### Link to Design Doc
NA

[test.txt](https://github.com/openpbs/openpbs/files/5858365/test.txt)

### DIFF
--- a/src/server/issue_request.c
+++ b/src/server/issue_request.c
@@ -88,6 +88,7 @@ extern char	*msg_issuebad;
 extern char     *msg_norelytomom;
 extern char	*msg_err_malloc;
 
+extern pbs_net_t	 pbs_server_addr;
 extern int max_connection;
 
 /**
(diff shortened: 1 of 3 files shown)


bhroam (approved):
LGTM

vstumpf (approved):
LGTM

PR #2218, opened 2021-01-26
scheduler might crash if secondary connection is not found

arungrover:
#### Describe Bug or Feature
If in some corner case scheduler does not make a secondary connection, it might crash while trying to find one.

#### Describe Your Change
Added a null check

#### Link to Design Doc
NA

NA


bhroam (approved):
LGTM

vstumpf (approved):
LGTM

PR #2219, opened 2021-01-26
TestCgroupsHook tests are not using the correct cgroups directory pat…

lisa-altair:
…h for the system

#### Describe Bug or Feature

The problem is that the tests are using a global self.paths, but there is only one such dictionary being used for all test systems.  Thus the information in the dictionary will be for the last host checked.  
The tests fail when the cgroups directory is located in different locations on the 2 test machines.  On TH2 test machines the cgroups directory seems to be mounted in a different path (/dev/cpuset) on the cpuset/cgroups machines versus many of the other test machines which have cgroups located at (/sys/fs/cgroup/cpuset). 

#### Describe Your Change

I created a path per host.  Many of the instances don't care which host is used, in those cases I use the first host.
When looking for the job directory within the cgroups directory, the correct path per host will be used.

Before the fix, tests fail when cgroups directory location differs on the different machines:
[checkpoint_fail.txt](https://github.com/openpbs/openpbs/files/5876700/checkpoint_fail.txt)
[cleanup_fail.txt](https://github.com/openpbs/openpbs/files/5876701/cleanup_fail.txt)

After the fix, tests pass when the cgroups directory location is in a different place:
[cgroups_cpuset_fixed.txt](https://github.com/openpbs/openpbs/files/5876676/cgroups_cpuset_fixed.txt)

[cgroups_linux_fixed.txt](https://github.com/openpbs/openpbs/files/5876678/cgroups_linux_fixed.txt)

This shows that the 5 failing tests were failing even before my fix.
[cgroups_orig_all.txt](https://github.com/openpbs/openpbs/files/5876681/cgroups_orig_all.txt)

### DIFF
--- a/test/tests/functional/pbs_cgroups_hook.py
+++ b/test/tests/functional/pbs_cgroups_hook.py
@@ -208,14 +208,15 @@ class TestCgroupsHook(TestFunctional):
         self.moms_list = []
         self.hosts_list = []
         self.nodes_list = []
+        self.paths = {}
         for cnt in range(0, len(self.moms)):
             mom = self.moms.values()[cnt]
             if mom.is_cray():
                 self.iscray = True
             host = mom.shortname
             # Check if mom has needed cgroup mounted, otherwise skip test
-            self.paths = self.get_paths(host)
-            if not self.paths['cpuset']:
+            self.paths[host] = self.get_paths(host)
+            if not self.paths[host]['cpuset']:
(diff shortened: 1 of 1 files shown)


bayucan (approved):
Looks good.

vccardenas (approved):
Changes look good to me.

bayucan:
Merged #2219 into master.

PR #2220, opened 2021-01-27
pbsfs -e/-d does not set last decay time

arungrover:
#### Describe Bug or Feature
When an admin issues pbsfs with '-e' or '-d' option to decay or erase the fair share usage, it does not set the decay time in the usage file. This can potentially lead to scheduler decaying the usage again when the scheduling cycle is triggered.

#### Describe Your Change
Updated pbsfs to set last_decay timestamp. Also, refactored the scheduler PTL class to revert fairshare.

#### Link to Design Doc
NA

Test are running

### DIFF
--- a/src/scheduler/pbsfs.cpp
+++ b/src/scheduler/pbsfs.cpp
@@ -214,8 +214,10 @@ main(int argc, char *argv[])
 	if (parse_group(RESGROUP_FILE, conf.fairshare->root) == 0)
 		return 1;
 
-	if (flags & FS_TRIM_TREE)
+	if (flags & FS_TRIM_TREE) {
 		read_usage(USAGE_FILE, FS_TRIM, conf.fairshare);
+		conf.fairshare->last_decay = time(0);
+	}
 	else
 		read_usage(USAGE_FILE, 0, conf.fairshare);
 
@@ -228,8 +230,10 @@ main(int argc, char *argv[])
 		printf("Fairshare usage units are in: %s\n", conf.fairshare_res);
 		print_fairshare(conf.fairshare->root, -1);
 	}
-	else if (flags & FS_DECAY)
+	else if (flags & FS_DECAY) {
 		decay_fairshare_tree(conf.fairshare->root);
+		conf.fairshare->last_decay = time(0);
+	}
 	else if (flags & (FS_GET | FS_SET | FS_COMP)) {
 		ginfo = find_group_info(argv[optind], conf.fairshare->root);
 
--- a/src/scheduler/pbsfs.cpp
+++ b/src/scheduler/pbsfs.cpp
@@ -216,7 +216,7 @@ main(int argc, char *argv[])
 
 	if (flags & FS_TRIM_TREE) {
 		read_usage(USAGE_FILE, FS_TRIM, conf.fairshare);
-		conf.fairshare->last_decay = time(0);
+		conf.fairshare->last_decay = time(NULL);
 	}
 	else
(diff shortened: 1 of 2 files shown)


test/fw/ptl/lib/ptl_sched.py
```diff
@@ -802,6 +802,21 @@ def set_dedicated_time_file(self, filename):
...
+        Helper method to revert scheduler's fairshare tree.
+        """
+        cmd = [os.path.join(self.pbs_conf['PBS_EXEC'], 'sbin', 'pbsfs'), '-e']
+        if self.sc_name is not 'default':
+            cmd += ['-I', self.sc_name]
+        self.du.run_cmd(cmd=cmd, runas=self.user)
+        self.parse_sched_config()
+        if self.platform == 'cray' or self.platform == 'craysim':
+            self.add_resource('vntype')
+            self.add_resource('hbmem')
```
vstumpf, line 816: Are these really reverting fairshare?

vstumpf (changes requested):

test/fw/ptl/lib/ptl_sched.py
arungrover: Well this is the default resource group we go back to on these platforms.

arungrover: oh, you are right, this isn't resource_group, I'll revert this back to where it was.

vstumpf (approved):
Thanks, I like it a bit better now.

src/scheduler/pbsfs.cpp
```diff
@@ -228,8 +230,10 @@ main(int argc, char *argv[])
 		printf("Fairshare usage units are in: %s\n", conf.fairshare_res);
 		print_fairshare(conf.fairshare->root, -1);
 	}
-	else if (flags & FS_DECAY)
+	else if (flags & FS_DECAY) {
 		decay_fairshare_tree(conf.fairshare->root);
+		conf.fairshare->last_decay = time(0);
```
bhroam, line 235: Say time(NULL).  I believe it takes a pointer, not a long.

test/fw/ptl/lib/ptl_sched.py
```diff
@@ -802,6 +802,18 @@ def set_dedicated_time_file(self, filename):
...
 
+    def revert_fairshare(self):
+        """
+        Helper method to revert scheduler's fairshare tree.
+        """
+        cmd = [os.path.join(self.pbs_conf['PBS_EXEC'], 'sbin', 'pbsfs'), '-e']
+        if self.sc_name is not 'default':
+            cmd += ['-I', self.sc_name]
+        self.du.run_cmd(cmd=cmd, runas=self.user)
+        self.parse_sched_config()
```
bhroam, line 813: I see this was there before, but I don't know why we parse the sched config here.  If it reads in the fairshare tree and resource group file, we immediately revert them below.  Otherwise rereading the sched_config doesn't have much to do with reverting fairshare.  It does need to happen, but not necessarily here.  Maybe it was just in the block of code in revert_to_defaults().  Now that you are moving it in to a new function, I don't think it belongs here.

bhroam (changes requested):

test/fw/ptl/lib/ptl_sched.py
arungrover: I'll move the call to parse config file back to revert_to_default().

arungrover:
Well, that didn't work.

bhroam (approved):
LGTM

vstumpf (approved):
LGTM

PR #2224, opened 2021-01-28
Fixing some performance issues

agrawalravi90:
#### Describe Bug or Feature

- Fixed a perf issue with get_svr_inst_sd() calls from sched by adding a hash map in sched

- Sped up check_node_array_eligibility() by preemptively  making nodes ineligible wherever possible

- Added avl caching to is_same_host() to avoid hostname lookups

Test for node_check_array_eligiblity improvements:
**Setup**: 5k nodes, 4998 nodes fully assigned, remaining 2 nodes: 15 out of 16 ncpus occupied. Then, submit 10k jobs with increasing walltime, and which request 2 ncpus, this will get past scheduler's allpart check and walk the check_node_array_eligibility code. Since most nodes are job-busy, the optimization with NSCR_CYCLE_INELIGIBLE should help. Results:
**before fix:**
Screen Shot 2021-02-12 at 4 58 11 PM

**after fix:**
Screen Shot 2021-02-12 at 5 23 56 PM


src/scheduler/node_info.cpp
```diff
@@ -1071,6 +1071,11 @@ remove_node_state(node_info *ninfo, const char *state)
 		&& !ninfo->is_unknown  && !ninfo->is_down && !ninfo->is_maintenance)
 		ninfo->is_free = 1;
 
+	if (ninfo->is_free)
+		ninfo->nscr |= NSCR_INELIGIBLE;
```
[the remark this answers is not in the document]
agrawalravi90: oops, mistake, thanks for catching

nithinj:
@agrawalravi90 ,
If you have indexing at is_same_host() level why duplicate for get_svr_inst_sd?

src/lib/Libutil/misc_utils.c
```diff
@@ -978,17 +979,45 @@ file_exists(char *path)
...
+
+	if (hostmap == NULL)
+		hostmap = pbs_idx_create(0, 0);
+
 	if (strcasecmp(host1, host2) == 0)
 		return 1;
-	if ((get_fullhostname(host1, host1_full, (sizeof(host1_full) - 1)) != 0)
-		|| (get_fullhostname(host2, host2_full, (sizeof(host2_full) - 1)) != 0))
+
+	pbs_idx_find(hostmap, (void **) &host1, (void **) &host1_f, NULL);
```
nithinj, line 996: I wish we had an LRU cache. With the AVL tree, we will be caching even obsolete objects.

nithinj:
I see this mapping is helpful in sending the jobs to the right servers. We can also keep this cache in the IFL which would provide an abstraction and help similar clients without having to do all these.

src/scheduler/globals.cpp
```diff
@@ -43,6 +43,9 @@
 #include <pthread.h>
 #include <limits.h>
 
+#include <unordered_map>
+#include <string>
+
```
bhroam, line 48: No real need to include these, you included them in globals.h

src/scheduler/node_info.cpp
```diff
@@ -1071,6 +1071,11 @@ remove_node_state(node_info *ninfo, const char *state)
 		&& !ninfo->is_unknown  && !ninfo->is_down && !ninfo->is_maintenance)
 		ninfo->is_free = 1;
 
+	if (ninfo->is_free)
+		ninfo->nscr &= ~NSCR_INELIGIBLE;
+	else
+		ninfo->nscr |= NSCR_INELIGIBLE;
+
```
bhroam, line 1078: There are quite a few more reasons a node can be marked ineligible than just state.  If you want to do this, you need to add a new element to the enum and create a ne NSCR bit.

src/scheduler/node_info.cpp
```diff
@@ -1138,8 +1143,11 @@ add_node_state(node_info *ninfo, const char *state)
 	}
 
 	/* Remove the free state unless it was specifically the state being added */
-	if (!set_free)
+	if (!set_free) {
 		ninfo->is_free = 0;
+		ninfo->nscr |= NSCR_INELIGIBLE;
+	} else
+		ninfo->nscr &= ~NSCR_INELIGIBLE;
```
bhroam, line 1150: Isn't this backwards?  You're setting ineligible if it is free and removing it when it is not free

src/scheduler/fifo.cpp
```diff
@@ -1392,7 +1392,10 @@ run_job(int pbs_sd, resource_resv *rjob, char *execvnode, int has_runjob_hook, s
 	if (rc) {
 		char buf[MAX_LOG_SIZE];
 		set_schd_error_codes(err, NOT_RUN, RUN_FAILURE);
-		errbuf = pbs_geterrmsg(get_svr_inst_fd(pbs_sd, rjob->job->svr_inst_id));
+
+		if (svrinstfd_map.find(rjob->job->svr_inst_id) == svrinstfd_map.end())
```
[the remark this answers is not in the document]
bhroam: Doesn't get_svr_inst_fd() iterate over like 5 things?  I don't thinkt his umap is going to help much and adds complexity that I don't see as needed.  If you want it, please test it to see if it really helps.

bhroam (changes requested):

src/lib/Libutil/misc_utils.c
agrawalravi90: hostnames won't go obsolete that often right?

agrawalravi90:
> I see this mapping is helpful in sending the jobs to the right servers. You could consider keeping this cache in the IFL instead which would provide an abstraction and help similar clients without having to do all these.

the cache i created is for mapping server_instance_id to server fd, IFL needs an fd itself, so we need the cache before calling IFL .. 

> If you have indexing at is_same_host() level why duplicate for get_svr_inst_sd?

get_svr_inst_sd doesn't have any indexing .. if you meant the map inside sched, then that map will prevent unnecessary calls to parse_servername() which will parse a string and the loop around servers to call is_same_host .. using the map doesn't really add much complexity, so why not avoid the unnecessary cost?

src/scheduler/fifo.cpp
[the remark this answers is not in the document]
agrawalravi90: as we discussed offline, it'll avoid string parsing that parse_servername() does, and the loop around servers, which can theoretically be up to 100, for each job. Using the map doesn't really add much complexity, but I will simplify this to move get_svr_inst_fd inside the scheduler, or create a scheduler wrapper for it so that the map is limited to that function, as Arun suggested.

agrawalravi90:
Putting WIP until i figure out the correct way to mark vnodes ineligible for check_node_array_eligibility, but please review the other changes in the meantime

src/scheduler/globals.cpp
vstumpf: I disagree with this, you should keep them here since you're using them in this file. If it gets removed from globals.h, they'll have to just add them here too.

agrawalravi90: @bhroam do you feel strongly about not including them?

bhroam:
Since you do the same lookup/add in multiple places, I'd suggest creating a function to do it.  That way it looks pretty much just like get_svr_inst_id().

src/scheduler/globals.cpp
bhroam: I still think having them in the .h file is enough.  It is a file specific .h file, so it has to be included in globals.cpp.  The only way we can remove it from the .h file is to remove the variables.  If we do that, we'll be removing them from the .cpp file as well.

nithinj:
> the cache i created is for mapping server_instance_id to server fd, IFL needs an fd itself, so we need the cache before calling IFL ..

IFL today doesn't know where a job belongs to. It can know, but it doesn't save that information. Hence it has to do trial and error till it succeeds in sending the job to the right server.
The scheduler cannot take this performance hit, so it figures out the right fd for IFL.
Instead, this indexing should be created at the IFL layer while doing a job stat. Save the job to server mapping for non-default cases (most times it can be derived from the jobid itself). That will enable the IFL to do the right thing instead of relying on the client.

src/lib/Libutil/misc_utils.c
nithinj: Might not. This is a utility function so can be used anywhere. We don't have to rely on this assumption if the element expires after a period of time.

agrawalravi90:
"The scheduler cannot take this performance hit, so it figures out the right fd for IFL."

Ok, I see your point, but it'll be a different cache if done in IFL: it'll need to be a cache of jobid to server fd instead of server_instance_id: server fd, so it'll be much larger than just a cache of svr_instance_id: fd, and it'll take longer to create and harder to maintain (invalidate and refresh). It'll also need to map multiple fds to a jobid, one for each virtual fd. And at the end of the day, it'll only help the scheduler because it's the only client that needs the performance boost, so makes much more sense to me to do it on the scheduler side and keep IFL simple. Let me know what you think

nithinj:
>  so it'll be much larger than just a cache of svr_instance_id: fd, and it'll take longer to create and harder to maintain (invalidate and refresh)

It will be definitely larger than svr_instance_id: fd, but might only introduce a few extra comparisons when indexed. Building a new cache on every job stat just for the default server will not introduce any consistency issues.
This way it is cleaner and the client does not have to be aware of IFL internals whether it is just the scheduler or other clients who will benefit from the potential speed-up.
But this requires walking over all the jobs within IFL which otherwise is not required today. We should re-consider it possibly after diff-stat when the diffs are failry low.

agrawalravi90:
Hey guys, i had added and then removed a PTL performance test since we only see noticeable difference when submitting a large number of jobs (like 100k). Attached manual test results

agrawalravi90:
@bhroam  and @nithinj  I've attached some perf results to show improvement to check_node_arr_eligiblity and also fixed another performance issues with get_svr_inst_id. I've also removed the scheduler's unordered_map cache as it didn't seem to provide significant benefit after adding the AVL tree to is_same_host. Also made sure that we don't look at unlicensed nodes for check_node_array_eligibilty, so i renamed the new nscr flag to a more generic name. Please take a look at the new commits and provide further feedback.

agrawalravi90:
@bhroam  @arungrover  and @nithinj gentle reminder, please provide further feedback, thanks.

src/scheduler/pbs_sched_utils.cpp
```diff
@@ -44,6 +44,7 @@
 
 #include <algorithm>
 #include <vector>
+#include <sstream>
```
nithinj, line 47: why just the header?

test/tests/performance/pbs_sched_perf.py
```diff
@@ -40,6 +40,7 @@
 
 from ptl.utils.pbs_logutils import PBSLogUtils
 from tests.performance import *
+import random
```
nithinj, line 43: why the header is added?

src/scheduler/globals.cpp
bhroam: You removed your unordered_map.  No need for the header any more.

src/scheduler/globals.h
```diff
@@ -42,6 +42,9 @@
 #include <pthread.h>
 #include <limits.h>
 
+#include <unordered_map>
+#include <string>
```
bhroam, line 46: Same here, no need.

src/scheduler/node_info.cpp
```diff
@@ -2368,9 +2380,13 @@ eval_selspec(status *policy, selspec *spec, place *placespec,
...
 
-	/* clear the node scratch space for use for node searching */
+	/* Remove visited, scattered and ineligible bits from ncsr for node searching
+	 * Since statebusy is a per cycle bit, it shouldn't be changed
+	 */
 	for (i = 0; ninfo_arr[i] != NULL; i++) {
-		ninfo_arr[i]->nscr = NSCR_NONE;
+		ninfo_arr[i]->nscr &= ~NSCR_VISITED;
+		ninfo_arr[i]->nscr &= ~NSCR_SCATTERED;
+		ninfo_arr[i]->nscr &= ~NSCR_INELIGIBLE;
```
bhroam, line 2389: Combine this into one line: ninfo_arr[i]->nscr &= (NSCR_VISITED | NSCR_SCATTERED | NSCR_INELIGIBLE);

src/scheduler/fifo.cpp
```diff
@@ -1510,7 +1511,9 @@ run_update_resresv(status *policy, int pbs_sd, server_info *sinfo,
 	pbs_errno = PBSE_NONE;
 	if (resresv->is_job && resresv->job->is_suspended) {
 		if (pbs_sd != SIMULATE_SD) {
-			pbsrc = pbs_sigjob(get_svr_inst_fd(pbs_sd, resresv->job->svr_inst_id), resresv->name, const_cast<char *>("resume"), NULL);
+			pbsrc = pbs_sigjob(get_svr_inst_fd(pbs_sd, resresv->job->svr_inst_id), resresv->name,
+					   const_cast<char *>("resume"), NULL);
```
bhroam, line 1515: When you rebase, you'll notice you'll need to do const_cast<char *>(resresv->name.c_str).  resresv->name is now a std::string.

bhroam (changes requested):

bhroam (approved):
LGTM

src/scheduler/node_info.cpp
```diff
@@ -2383,11 +2383,8 @@ eval_selspec(status *policy, selspec *spec, place *placespec,
 	/* Remove visited, scattered and ineligible bits from ncsr for node searching
 	 * Since statebusy is a per cycle bit, it shouldn't be changed
 	 */
-	for (i = 0; ninfo_arr[i] != NULL; i++) {
-		ninfo_arr[i]->nscr &= ~NSCR_VISITED;
-		ninfo_arr[i]->nscr &= ~NSCR_SCATTERED;
-		ninfo_arr[i]->nscr &= ~NSCR_INELIGIBLE;
-	}
+	for (i = 0; ninfo_arr[i] != NULL; i++)
+		ninfo_arr[i]->nscr &= ~(NSCR_VISITED | NSCR_SCATTERED | NSCR_INELIGIBLE);
```
arungrover, line 2387: Can we not make a macro value of all these three flags together? That would mean the scheduler will not be doing this or operation every time for every job.

arungrover (changes requested):

src/scheduler/node_info.cpp
agrawalravi90: i doubt bitwise operations cost much, these are in an enum, so making a macro for them might be a little weird .. @bhroam  what do you think about @arungrover 's idea?

agrawalravi90:
@bhroam  @nithinj  and @arungrover sorry i rebased to resolve conflicts, did performance testing, fixed minor issues on server side and refactored the calls to pbs_sigjob in scheduler, please check the changes to scheduler_ifl_wrappers.c and the second commit.

src/lib/Libutil/misc_utils.c
nithinj: I'm seeing is_same_host being invoked from svr_chk_owner.c
This might be called against any client request authorized or not. You can confirm this.

agrawalravi90: so what? we'll cache client hosts as well, which is again good, we'll save the cost of hostnames lookups there as well, right?

bhroam (approved):

src/scheduler/node_info.cpp
arungrover: I tried checking the assembly with or without macro and there is no difference at all. So I take it back, It's fine the way it is now.

arungrover (approved):

agrawalravi90:
Thanks for the sign-offs @bhroam  and @arungrover  
@nithinj  please let me know if you still have any concerns.

PR #2225, opened 2021-01-29
Removed an unnecessary hostname lookup

arungrover:
#### Describe Bug or Feature
WHile trying to issue batch request from one server to another, PBS server ends up calling get_addrinfo twice. 

#### Describe Your Change
Removed unnecessary call.

#### Link to Design Doc
NA

Profile: regression
|--|--|--|--|--|--|--|
|5846|3815|0|6|0|1|3808|

Tests failing are not related to this change

### DIFF
--- a/src/server/issue_request.c
+++ b/src/server/issue_request.c
@@ -257,10 +257,9 @@ issue_to_svr(char *servern, struct batch_request *preq, void (*replyfunc)(struct
 				svrname = server_host;
 		}
 	}
-	if (comp_svraddr(pbs_server_addr, svrname) == 0)
+	if (comp_svraddr(pbs_server_addr, svrname, &svraddr) == 0)
 		return (issue_Drequest(PBS_LOCAL_CONNECTION, preq, replyfunc, 0, 0));
 
-	svraddr = get_hostaddr(svrname);
 	if (svraddr == (pbs_net_t)0) {
 		if (pbs_errno == PBS_NET_RC_RETRY)
 			/* Non fatal error - retry */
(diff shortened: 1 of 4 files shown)


src/server/issue_request.c
```diff
@@ -257,10 +257,9 @@ issue_to_svr(char *servern, struct batch_request *preq, void (*replyfunc)(struct
 				svrname = server_host;
 		}
 	}
-	if (comp_svraddr(pbs_server_addr, svrname) == 0)
+	if (comp_svraddr(pbs_server_addr, svrname, &svraddr) == 0)
```
bhroam, line 260: Could you turn the (pbs_net_t) 0 below to NULL?

bhroam (approved):
Just one nit and not in your code.  I'll sign off.

src/server/issue_request.c
arungrover: I don't think we can, svraddr is not an address it is of type pbs_net_t which is a long. comp_svraddr sets it to 0 when it's not able to find the IP addresses associated to svrname.

subhasisb:
We have a simple avltree cache in pbs, the tree pointed by hostaddr_idx. We could possibly use that cache to reduce such ns lookups. The problem with that cache is that it does not have a auto expiry of entries though, which means if hostnames change, it is a bad fit - right now this is used for moms, and that works since PBS does not expect mom addresses to change when they are added to PBS....

Some day, this would be a nice this to modernize, consolidate...

subhasisb (approved):

src/server/issue_request.c
bhroam: Ahh, OK.  Before we used to do things like (char *) 0.  Vincent replaced most (if not all) of these with NULL at one point.  I just thought this was one that he missed.  Since it is a long, you can't use NULL.

PR #2227, opened 2021-01-29
Ignore failures of OOM and RLIMIT set

ashwathraop:
#### Describe Bug or Feature

Server start fails if OOM or RLIMIT set fails. This can make PBS start on containers fail if the privileged option is not used.

#### Describe Your Change

Ignore the failure and continue.

#### Link to Design Doc

N/A

### DIFF
--- a/src/lib/Libdb/pgsql/db_common.c
+++ b/src/lib/Libdb/pgsql/db_common.c
@@ -458,9 +458,11 @@ pbs_dataservice_control(char *cmd, char *pbs_ds_host, int pbs_ds_port)
 		}
 		if (oom_val != NULL) {
 			if ((fd = open(oom_file, O_TRUNC | O_WRONLY, 0600)) != -1) {
-				(void)write(fd, oom_val, strlen(oom_val));
+				if (write(fd, oom_val, strlen(oom_val)) == -1)
+					ret = PBS_DB_OOM_ERR;
 				close(fd);
-			}
+			} else
+				ret = PBS_DB_OOM_ERR;
 			free(oom_val);
 		}
 		sprintf(errfile, "%s/spool/pbs_ds_monitor_errfile", pbs_conf.pbs_home_path);
@@ -1349,6 +1351,10 @@ pbs_db_get_errmsg(int err_code, char **err_msg)
 
 	case PBS_DB_CONNFAILED:
 		*err_msg = strdup("Failed to connect to PBS dataservice");
+		break;
+
+	case PBS_DB_OOM_ERR:
+		*err_msg = strdup("Failed to protect PBS from Linux OOM killer. No access to OOM score file.");
 		break;
 
 	case PBS_DB_ERR:
--- a/src/lib/Libdb/pgsql/db_common.c
+++ b/src/lib/Libdb/pgsql/db_common.c
@@ -157,7 +157,7 @@ pg_db_fn_t db_fn_arr[PBS_DB_NUM_TYPES] = {
  *	Initialize a query state variable, before being used in a cursor
  *
  * @param[in]	conn - Database connection handle
- * @param[in]	query_cb - Object handler query back function 
+ * @param[in]	query_cb - Object handler query back function
  *
  * @return	void *
  * @retval	NULL - Failure to allocate memory
@@ -445,7 +445,10 @@ pbs_dataservice_control(char *cmd, char *pbs_ds_host, int pbs_ds_port)
 	}
 
 	if (!(strcmp(cmd, PBS_DB_CONTROL_START))) {
-		/* Protect self from Linux OOM killer */
+		/*
+		 * try protect self from Linux OOM killer
+		 * but don't fail if can't update OOM score
+		 */
 		if (access(oom_score_adj, F_OK) != -1) {
 			strcpy(oom_file, oom_score_adj);
 			oom_val = strdup("-1000");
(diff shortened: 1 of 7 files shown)


src/lib/Libdb/pgsql/pbs_ds_systemd
```diff
@@ -38,10 +38,10 @@
 
 . ${PBS_CONF_FILE:-/etc/pbs.conf}
 
-is_systemd=0
-systemctl --version >/dev/null 2>&1
-if [ $? -eq 0 ] ; then
-    is_systemd=1
+is_systemd=1
+_status=$(systemctl is-system-running 2>/dev/null)
+if [ "x${_status}" == "xoffline" -o "x${_status}" == "xunknown" ] ; then
```
bremanandjk, line 43: Why are we adding an extra character "x" before the variable string? I see without adding that character also the comparison works fine. Curious to know.

ashwathraop: Its extra precaution if ${_status} is empty you will compare x == xoffline.

bremanandjk: Yeah, I thought the same, But with empty string also working fine. 
#!/bin/bash

var1="same"
var2=
if [ "$var1" == "$var2" ]
then
	echo "Equal"
else
	echo "Not Equal"
fi


Did you see any error with this ?

bremanandjk: I don't see any harm in adding that check. If you see it's having any specific reason, we will add that character.

bremanandjk (approved):
LGTM

src/lib/Libdb/pgsql/db_common.c
```diff
@@ -454,20 +457,11 @@ pbs_dataservice_control(char *cmd, char *pbs_ds_host, int pbs_ds_port)
...
-				return -1;
-			}
-			if (write(fd, oom_val, strlen(oom_val)) == -1) {
-				if (errmsg_cache)
-					free(errmsg_cache);
-				errmsg_cache = strdup("OOM protect: file write failed");
-				return -1;
+			if ((fd = open(oom_file, O_TRUNC | O_WRONLY, 0600)) != -1) {
+				(void)write(fd, oom_val, strlen(oom_val));
+				close(fd);
```
agrawalravi90, line 462: I think we should still write the error message so that if OOM killer does kill it, we can see from logs that the OOM protect had failed

src/resmom/mom_main.c
```diff
@@ -7310,10 +7310,8 @@ main(int argc, char *argv[])
 
 #ifdef RLIMIT_NPROC
 		(void)getrlimit64(RLIMIT_NPROC, &orig_nproc_limit); /* get for later */
-		if (setrlimit64(RLIMIT_NPROC, &rlimit) == -1) {    /* set unlimited */
+		if (setrlimit64(RLIMIT_NPROC, &rlimit) == -1)    /* set unlimited */
 			perror(" setrlimit NPROC");
```
agrawalravi90, line 7314: convert perror() to log_err() (here and below) so that it goes into mom logs and there's a record of the fact that setting rlimit had failed, which might explain any unexpected behavior as a consequence.

agrawalravi90 (changes requested):

src/lib/Libdb/pgsql/db_common.c
vstumpf: I agree. We should have access to the logger from here, right?

ashwathraop: This is part of libdb which doesn't write to logs directly. The only way to pass the logs is through errmsg but if the next operation is a success we may not capture it.

ashwathraop:
Profile: regression
|--|--|--|--|--|--|--|
|5853|3815|1|6|0|1|3807|

A rerun of failed tests matches with master
Description: Rerun Only Failed Tests From #5853
Profile: regression
|--|--|--|--|--|--|--|
|5855|7|0|6|0|0|1|

ashwathraop:
After rebasing failed tests are passing.

Description: Rerun Only Failed Tests From #5855
Profile: regression
|--|--|--|--|--|--|--|
|5859|6|0|0|0|0|6|

src/lib/Libdb/pgsql/db_common.c
vstumpf: Ah, that sucks. In the caller, can we check if errmsg_cache is not null and log?

ashwathraop: No, not directly. We call pbs_db_get_errmsg() to get errmsg_cache in case of non-zero return code. I will have to introduce a new error code for OOM failure and make callers just log but still consider it as a successful start DB.

vstumpf (approved):
LGTM

Should we also remove privileged from our ci script and azure pipelines?

src/lib/Libdb/pgsql/db_common.c
agrawalravi90: as we discussed offline, if it's not a lot of effort to add a new return code, can you please try to do it?

ashwathraop:
@vstumpf and @agrawalravi90  I added a new error code and made the server log a warning.

[root@pbs-server build]# /etc/init.d/pbs start
Starting PBS
/opt/pbs/sbin/pbs_comm ready (pid=55720), Proxy Name:pbs-server.pbspro.org:17001, Threads:4
PBS comm
PBS sched
**WARNING: Failed to protect PBS from Linux OOM killer. No access to OOM score file.**
Connecting to PBS dataservice...connected to PBS dataservice@pbs-server
Licenses valid for 1000000 Floating hosts
PBS server

[root@pbs-server build]# grep OOM /var/spool/pbs/server_logs/20210201
**02/01/2021 19:16:56;0006;Server@pbs-server;Svr;Server@pbs-server;WARNING: Failed to protect PBS from Linux OOM killer. No access to OOM score file.
02/01/2021 19:17:07;0006;Server@pbs-server;Svr;Server@pbs-server;WARNING: Failed to protect PBS from Linux OOM killer. No access to OOM score file.**

agrawalravi90 (approved):
LGTM, thanks for making the changes

agrawalravi90 (changes requested):
blocking it until you are done with regression

ashwathraop:
Test results:
Profile: regression
|--|--|--|--|--|--|--|
|5890|3815|1|0|0|1|3813|

The rerun of failed test:
Description: Rerun Only Failed Tests From #5890
Profile: regression
|--|--|--|--|--|--|--|
|5892|1|0|0|0|0|1|

agrawalravi90 (approved):

vstumpf (approved):
LGTM

PR #2229, opened 2021-01-30
Bypass dev disc and nvidia-smi for compute nodes in cgroup hook

alexis-cousein:
#### Describe Bug or Feature

The cgroup hook spends a lot of time in _all_ events (not just exechost_startup) discovering devices and trying to run nvidia-smi. At many sites this takes up more than 2/3rds of the overhead of the cgroup hook, which gates throughput when there are a lot of small short jobs.

All of this is unnecessary on compute nodes without GPUs, or with Nvidia cards that are not used as GPUs. At many sites the cluster may have thousands of such nodes and tens or hundreds of GPU nodes only.

Also, sharing nodes between more than one job when using GPUs only works when the "devices" subsystem (cgroup controller) is enabled in the hook configuration, but the hook's execjob_startup still sets resources_available.ngpus to detected GPUs even when the devices subsystem is _dis_abled in the hook.

Another side effect of the hook's current strategy of always discovering devices when instantiating the NodeUtils classes is that at DEBUG4 the hook is slowed down by the sheer amount of lines printed in the log, which also make the log well nigh unreadable. All of this is noise if the site does not plan use 

#### Describe Your Change

The hook introduces an extra configuration variable in the config file, discover_gpus (it can be set differently on different nodes with the existing functionality to set booleans according to vntype or hostnames, so that e.g. GPU discovery). It is set to True by default except if the devices section is not enabled (when the devices section is disabled, then it is reset to False regardless of what it is set to, with a log line in the MoM log indicating it is disabled and why). Whenever it is utlitmately set to False, nvidia-smi is not called.

Alternatively, setting the 'nvidia-smi' flag to the empty string will also neuter GPU discovery. This, however does not offer the same functionality, since this is not a boolean that can be enabled/disabled depending on host vntype/hostname. 

When the devices section is disabled, device discovery is suppressed, and the "devices" discovered becomes an empty dictionary. Guards are inserted in existing code to avoid any danger of triggering an exception because of this.

Some lines were also edited to achieve a pycodestyle clean code (a recent commit introduced "l" (ell) variables that pycodestyle does not like because they are ambiguous when displayed in certain fonts. These pycodestyle errors can prevent certain QA tests from being run correctly on some build environments if they insist in pycodestyle clean hooks.

#### Link to Design Doc

N/A -- at most sites, discover_gpus would already be set to false whenever necessary (and without an explicit interface) since typically these sites do not enable the devices subsection _unless_ MIC or GPUs are being managed.

If necessary we might want a design document just to document the extra flag that is introduced just in case sites want to enable the devices controller but still _dis_able GPU discovery. I will leave that to the reviewers to decide.

### DIFF
--- a/src/hooks/cgroups/pbs_cgroups.CF
+++ b/src/hooks/cgroups/pbs_cgroups.CF
@@ -8,6 +8,7 @@
     "online_offlined_nodes" : true,
     "use_hyperthreads"      : false,
     "ncpus_are_cores"       : false,
+    "discover_gpus"         : true,
     "cgroup" : {
         "cpuacct" : {
             "enabled"            : true,
--- a/src/hooks/cgroups/pbs_cgroups.PY
+++ b/src/hooks/cgroups/pbs_cgroups.PY
@@ -1078,7 +1078,9 @@ class HookUtils(object):
                 for key in cgroup.assigned_resources['device_names']:
                     if key.startswith('mic'):
                         mics.append(key[3:])
-                    elif key.startswith('nvidia'):
+                    elif (key.startswith('nvidia')
+                            and "gpu" in node.devices
+                            and key in node.devices['gpu']):
                         if 'uuid' in node.devices['gpu'][key]:
                             gpus.append(node.devices['gpu'][key]['uuid'])
                         if 'uuids' in node.devices['gpu'][key]:
@@ -1169,7 +1171,8 @@ class HookUtils(object):
             pbs.logmsg(pbs.EVENT_DEBUG4,
                        'assigned_resources: %s' %
                        (cgroup.assigned_resources))
-            cgroup.setup_job_devices_env(node.devices['gpu'])
+            if "gpu" in node.devices:
+                cgroup.setup_job_devices_env(node.devices['gpu'])
         return True
 
     def _exechost_periodic_handler(self, event, cgroup, jobutil):
@@ -1326,7 +1329,9 @@ class HookUtils(object):
                 for key in cgroup.assigned_resources['device_names']:
                     if key.startswith('mic'):
                         mics.append(key[3:])
-                    elif key.startswith('nvidia'):
+                    elif (key.startswith('nvidia')
+                            and "gpu" in node.devices
+                            and key in node.devices['gpu']):
                         if 'uuid' in node.devices['gpu'][key]:
                             gpus.append(node.devices['gpu'][key]['uuid'])
                         if 'uuids' in node.devices['gpu'][key]:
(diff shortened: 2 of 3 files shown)


bayucan (commented):
@alexis-cousein : change looks good. Care to add also a PTL test for this given that it's a new option?

alexis-cousein:
Not sure what to test for -- in general the desired behaviour is to do as little different as possible ;-).

-A test to confirm that if the devices section is disabled, that nvidia-smi is not being called?
-Another test to confirm that even when you _do_ enable the devices section, nvidia-smi is not called if discover_gpus is set to False?

AFAIK, we don't have GPU-enabled test vehicles, so calling nvidia-smi should be a proxy for "trying to discover GPUs", but we can't really test if resources_available.ngpus is being set correctly (I did do that on my machine which does have a GPU, which is how I discovered that you'd better set ngpus to 0 *explicitly* when disabling the devices subsystem section --- otherwise after a HUP of MoM, MoM keeps the existing values).

vstumpf:
> -A test to confirm that if the devices section is disabled, that nvidia-smi is not being called?
I think this is a good starting point, if we need a test to check the stale gpus someone can write a test later.

Otherwise, the rest of the PR looks good to me.

bayucan:
I agree with @vstumpf 's last suggestion of having "-A test to confirm that if the devices section is disabled, that nvidia-smi is not being called?"

alexis-cousein:
Added three tests. One to confirm that we're still calling device and GPU discovery if devices is enabled and nothing is changed from the default, one to confirm that disabling "devices" is enough not to call either, and one to confirm that if discover_gpus is set to false, _even_ when devices are discovered _discover_gpus does not call nvidia-smi.

I also took the liberty of hardening one of the setup routines: it adds an "ngpu" resource, but on recent versions of PBSPro that may already exist, so it first queries the server for resources and only creates the ones that do not exist. It didn't make tests on my machines fail all of the time, but if you have unclean exits of a PTL test it can mess up later ones.

alexis-cousein:
@bayucan I see all checks have passed, but I also see that the cgroup hook tests aren't being run anyway. Are there any plans to enable them? Can anyone confirm they do indeed pass on anything except my build platform?

bayucan:
@alexis-cousein : I've reviewed your latest update and it looked fine. I'm currently building and testing your TestCgroupsHook update and I'll let you know when it's done.

bayucan:
@alexis-cousein : The build and pbs_cgroups_hook.py  run was successful, included passes for your 3 new tests.
[ptl.txt](https://github.com/openpbs/openpbs/files/5954559/ptl.txt)

bayucan (approved):
Looks good.

bayucan:
@vstumpf : I'll need your okay on this before I can merge.

vstumpf (approved):
LGTM

PR #2234, opened 2021-02-02
updates to test for qstat -pt

vccardenas:
#### Describe Bug or Feature
On some systems the sleep 7 is not sufficient for the first subjob to show up as 100% done in qstat -pt output.

#### Describe Your Change
Removed the sleep 7 and replaced with server.expect that the first subjob is in X state before doing qstat -pt.  Added a minor correction so that the last subjob is also checked.

[TestQstat_before.txt](https://github.com/openpbs/openpbs/files/5914618/TestQstat_before.txt)
[TestQstat_after.txt](https://github.com/openpbs/openpbs/files/5914619/TestQstat_after.txt)

### DIFF
--- a/test/tests/functional/pbs_qstat.py
+++ b/test/tests/functional/pbs_qstat.py
@@ -63,24 +63,26 @@ class TestQstat(TestFunctional):
 
         self.server.expect(JOB, {'job_state': 'B'}, id=jid)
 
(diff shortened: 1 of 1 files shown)


bayucan (approved):
Looks good.

lisa-altair (approved):
Looks good.

bayucan:
Merged #2234 into master.

PR #2235, opened 2021-02-03
Scheduler might leak memory in multi-sched mode

arungrover:
#### Describe Bug or Feature
The scheduler might leak some memory while trying to reset the scheduler comment.

#### Describe Your Change
This only happens in the case of a multi-sched when the scheduler is trying to update its comment attribute. I added a missing free().

#### Link to Design Doc
NA

[sched.log](https://github.com/openpbs/openpbs/files/5915420/sched.log)


bhroam (approved):
LGTM

subhasisb (approved):

PR #2238, opened 2021-02-04
cleanup *e and *o files in user homedir

visheshh:
#### Describe Bug or Feature
Large number of *e and *o files in user home directory fills up the system after tests which may submit thousands of jobs.

#### Describe Your Change
Added cleanup in user's homedir by the user himself in ptl_test_runner.py for all PBS_USERS

[cleanup.zip](https://github.com/openpbs/openpbs/files/5960496/cleanup.zip)

### DIFF
--- a/test/fw/ptl/utils/plugins/ptl_test_runner.py
+++ b/test/fw/ptl/utils/plugins/ptl_test_runner.py
@@ -1045,15 +1045,15 @@ class PTLTestRunner(Plugin):
         for user in PBS_USERS:
             self.logger.info('Cleaning %s\'s home directory' % (str(user)))
             runas = PbsUser.get_user(user)
-            for mom in hosts:
-                ret = du.run_cmd(mom, cmd=['echo', '$HOME'], sudo=True,
+            for host in hosts:
+                ret = du.run_cmd(host, cmd=['echo', '$HOME'], sudo=True,
                                  runas=runas, logerr=False, as_script=True)
                 if ret['rc'] == 0:
                     path = ret['out'][0].strip()
                 else:
                     return None
                 ftd = []
-                files = du.listdir(mom, path=path, runas=user)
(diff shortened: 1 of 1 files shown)


kjakkali (approved):
Changes looks good. Please add test log.

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -1038,6 +1038,28 @@ def addSuccess(self, test):
     def _cleanup(self):
         self.logger.info('Cleaning up temporary files')
         du = DshUtils()
+        for user in PBS_USERS:
+            self.logger.info('Cleaning %s\'s home directory' % (str(user)))
+            runas = PbsUser.get_user(user)
+            for mom in self.param_dict['moms']:
+                ret = du.run_cmd(mom, cmd=['echo', '$HOME'], sudo=True,
+                                 runas=runas, logerr=False, as_script=True)
```
sakshamgarg, line 1046: Why are we cleaning the home directory only for MoM hosts? It will be effective only if MoM and Server are on the same machine. We should clean the home directory of the server host as well because the job submission is done from the host where pbs-server is running.

visheshh: Hi @sakshamgarg , You are right i will add server machine also to the list.

visheshh: @sakshamgarg addressed your comments. Please review.

sakshamgarg (changes requested):
One small suggestion. The rest of the code looks good to me.

sakshamgarg (approved):
Thanks @visheshh for making the changes. PR looks good to me

visheshh:
@openpbs/openpbs-maintainers Please take this PR forward

hirenvadalia (approved):

PR #2239, opened 2021-02-05
Minor fix to sched's send_run_job

agrawalravi90:
#### Describe Bug or Feature

With some multi-server changes to sched, we added a new job/node attribute called server_instance_id. Right now, sched necessarily wants this attribute to be set, but it shouldn't be made necessary, it's just information for it to get the right server's fd, so it should just let IFL decide how to process an empty server_instance_id, if it's empty.

#### Describe Your Change

Just remove a couple of NULL checks for server_instance_id

No major change here, existing tests passing should be enough


arungrover (approved):

bhroam (approved):
LGTM

PR #2246, opened 2021-02-09
multi-svr fixes: Sched crash + PBS_SERVER_INSTANCES bug

agrawalravi90:
#### Describe Bug or Feature

- parse_psi was assigning 15001 as the server port for single server scenario, where PBS_SERVER_INSTANCES was missing from pbs.conf.
- Sched was crashing as follows because of an invalid free in aggr_job_ct(), we were freeing the attribute object's value but setting a function local pointer, so 'value' remained free, which meant that when sched free'd it, it crashed as:
```
==1389== Invalid free() / delete / delete[] / realloc()
==1389==    at 0x4C3210C: free (vg_replace_malloc.c:538)
==1389==    by 0x477EF1: __pbs_statfree (pbs_statfree.c:77)
==1389==    by 0x44278B: query_queues(status*, int, server_info*) (queue_info.cpp:322)
==1389==    by 0x44F9A2: query_server(status*, int) (server_info.cpp:268)
==1389==    by 0x4236F8: scheduling_cycle(int, sched_cmd const*) (fifo.cpp:629)
==1389==    by 0x423A46: intermediate_schedule(int, sched_cmd const*) (fifo.cpp:562)
==1389==    by 0x45A556: schedule_wrapper(sched_cmd*, int) (pbs_sched_utils.cpp:1210)
==1389==    by 0x45B522: sched_main(int, char**, int (*)(int, sched_cmd const*)) (pbs_sched_utils.cpp:1156)
==1389==    by 0x679D7B2: (below main) (in /usr/lib64/libc-2.28.so)
==1389==  Address 0xa4fc050 is 0 bytes inside a block of size 2 free'd
==1389==    at 0x4C3210C: free (vg_replace_malloc.c:538)
==1389==    by 0x4770F8: aggr_job_ct (int_status.c:227)
==1389==    by 0x477A9A: aggregate_queue (int_status.c:444)
==1389==    by 0x477A9A: PBSD_status_aggregate (int_status.c:566)
==1389==    by 0x47B4CE: __pbs_statque (pbsD_statque.c:69)
==1389==    by 0x442531: query_queues(status*, int, server_info*) (queue_info.cpp:153)
==1389==    by 0x44F9A2: query_server(status*, int) (server_info.cpp:268)
==1389==    by 0x4236F8: scheduling_cycle(int, sched_cmd const*) (fifo.cpp:629)
==1389==    by 0x423A46: intermediate_schedule(int, sched_cmd const*) (fifo.cpp:562)
==1389==    by 0x45A556: schedule_wrapper(sched_cmd*, int) (pbs_sched_utils.cpp:1210)
==1389==    by 0x45B522: sched_main(int, char**, int (*)(int, sched_cmd const*)) (pbs_sched_utils.cpp:1156)
==1389==    by 0x679D7B2: (below main) (in /usr/lib64/libc-2.28.so)
==1389==  Block was alloc'd at
==1389==    at 0x4C30F0B: malloc (vg_replace_malloc.c:307)
==1389==    by 0x46738D: disrst (disrst.c:99)
==1389==    by 0x485EC6: decode_DIS_attrl (dec_attrl.c:118)
==1389==    by 0x4769C6: read_batch_status (dec_rpyc.c:102)
==1389==    by 0x4769C6: decode_DIS_replyCmd (dec_rpyc.c:385)
==1389==    by 0x46B668: PBSD_rdrpy_sock (int_rdrpy.c:91)
==1389==    by 0x46B712: PBSD_rdrpy (int_rdrpy.c:124)
==1389==    by 0x4775ED: PBSD_status_get (int_status.c:658)
==1389==    by 0x477969: PBSD_status_aggregate (int_status.c:554)
==1389==    by 0x47B4CE: __pbs_statque (pbsD_statque.c:69)
==1389==    by 0x442531: query_queues(status*, int, server_info*) (queue_info.cpp:153)
==1389==    by 0x44F9A2: query_server(status*, int) (server_info.cpp:268)
==1389==    by 0x4236F8: scheduling_cycle(int, sched_cmd const*) (fifo.cpp:629)
```

#### Describe Your Change

- parse_psi() is now passed NULL when PSI is absent from pbs.conf, and internally uses pbs_conf.batch_service_port along with pbs_default() when it is NULL
- Fixed the sched crash by setting the total count attribute correctly.

Regression results:
Profile: regression
|--|--|--|--|--|--|--|
|6010|3888|1|0|0|1|3886|

The 1 failure is a transient issue, happens on master too, unrelated to my changes

### DIFF
--- a/src/lib/Libifl/pbs_loadconf.c
+++ b/src/lib/Libifl/pbs_loadconf.c
@@ -317,14 +317,17 @@ parse_psi(char *conf_value)
 	}
 	free_string_array(list);
 	pbs_conf.pbs_num_servers = i;
-	pbs_conf.psi_str = strdup(local_conf);
-	free(local_conf);
+	if (conf_value)
+		pbs_conf.psi_str = strdup(local_conf);
+	else
+		pbs_conf.psi_str = local_conf;
 
 	return 0;
 
 err:
 	free_string_array(list);
-	free(local_conf);
+	if (conf_value == NULL)
+		free(local_conf);
 	return -1;
 }
 
--- a/src/lib/Libifl/pbs_loadconf.c
+++ b/src/lib/Libifl/pbs_loadconf.c
@@ -272,7 +272,7 @@ parse_config_line(FILE *fp, char **key, char **val)
 int
 parse_psi(char *conf_value)
 {
-	char **list;
+	char **list = NULL;
 	int i;
 	char *svrname = NULL;
 	char *local_conf = NULL;
@@ -289,23 +289,21 @@ parse_psi(char *conf_value)
 
 	list = break_comma_list(local_conf);
 	if (list == NULL)
-		return -1;
+		goto err;
 
 	for (i = 0; list[i] != NULL; i++)
 		;
 
 	if (!(pbs_conf.psi = calloc(i, sizeof(psi_t)))) {
 		fprintf(stderr, "Out of memory while parsing configuration %s", local_conf);
-		free_string_array(list);
-		return -1;
+		goto err;
 	}
 
 	for (i = 0; list[i] != NULL; i++) {
 		svrname = parse_servername(list[i], &(pbs_conf.psi[i].port));
 		if (svrname == NULL) {
 			fprintf(stderr, "Error parsing PBS_SERVER_INSTANCES %s \n", list[i]);
-			free_string_array(list);
-			return -1;
+			goto err;
 		}
 		strcpy(pbs_conf.psi[i].name, svrname);
 
@@ -320,11 +318,14 @@ parse_psi(char *conf_value)
 	free_string_array(list);
 	pbs_conf.pbs_num_servers = i;
 	pbs_conf.psi_str = strdup(local_conf);
-
-	if (conf_value == NULL)
-		free(local_conf);
+	free(local_conf);
 
 	return 0;
+
+err:
+	free_string_array(list);
+	free(local_conf);
+	return -1;
 }
 
 
--- a/src/lib/Libifl/pbs_loadconf.c
+++ b/src/lib/Libifl/pbs_loadconf.c
@@ -275,14 +275,19 @@ parse_psi(char *conf_value)
 	char **list;
 	int i;
 	char *svrname = NULL;
+	char *local_conf = NULL;
 
 	free(pbs_conf.psi);
 	free(pbs_conf.psi_str);
 
-	if (conf_value == NULL)
-		return -1;
+	if (conf_value == NULL) {
+		pbs_asprintf(&local_conf, "%s:%d", pbs_default(), pbs_conf.batch_service_port);
+		if (local_conf == NULL)
+			return -1;
+	} else
+		local_conf = conf_value;
 
-	list = break_comma_list(conf_value);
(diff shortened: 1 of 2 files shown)


src/lib/Libifl/pbs_loadconf.c
```diff
@@ -275,22 +275,27 @@ parse_psi(char *conf_value)
...
+		pbs_asprintf(&local_conf, "%s:%d", pbs_default(), pbs_conf.batch_service_port);
+		if (local_conf == NULL)
+			return -1;
+	} else
+		local_conf = conf_value;
 
-	list = break_comma_list(conf_value);
+	list = break_comma_list(local_conf);
 	if (list == NULL)
 		return -1;
```
hirenvadalia, line 292: I think u need to free local_conf from here and before all returns also, may use `goto` and use condition like
```c
if (conf_value == NULL)
    free(local_conf)
```

src/lib/Libifl/pbs_loadconf.c
```diff
@@ -314,7 +319,10 @@ parse_psi(char *conf_value)
 	}
 	free_string_array(list);
 	pbs_conf.pbs_num_servers = i;
-	pbs_conf.psi_str = strdup(conf_value);
+	pbs_conf.psi_str = strdup(local_conf);
```
bhroam, line 322: Suggestion: Instead of calling strdup() all the time, only call it when conf_value != NULL.  Otherwise just assign psi_str = local_conf.  It's already allocated memory.  Then you can remove the free below.

bhroam (approved):
Only one minor suggestion you can take or not.

src/lib/Libifl/pbs_loadconf.c
agrawalravi90: Thanks for the suggestion, but i feel like the code will be simpler if we just do the strdup and free local_conf (actually keep it local). So I'd prefer that we leave it as is unless you feel strongly about this.

bhroam: I guess it is a bit simpler.  I just didn't like the idea of duping allocated memory, just to free the memory.  It doesn't really matter because local_conf should be just a few bytes, and this is called once at the start of the program.

bhroam (approved):

src/lib/Libifl/pbs_loadconf.c
```diff
@@ -1097,7 +1106,7 @@ __pbs_loadconf(int reload)
 
 	pbs_conf.loaded = 1;
 
-	if (parse_psi(psi_value ? psi_value : pbs_default()) == -1) {
+	if (parse_psi(psi_value) == -1) {
 		fprintf(stderr, "Couldn't find a valid server instance to connect to\n");
 		free(psi_value);
```
hirenvadalia, line 1111: Remove this free and other `free(psi_value)` else it will be double free as parse_psi is already freeing given value (even in error case also)

hirenvadalia (changes requested):

hirenvadalia:
@agrawalravi90 Changes looks good, can you please add test results?

hirenvadalia (approved):

PR #2247, opened 2021-02-09
updated sleep time for slower machines

RKORavi:
#### Describe Bug or Feature
Fix PTL tests on slower machines

#### Describe Your Change
1. Increase sleep time of job
2. Remove R state check for jobs because in next lie we check for exit_status=0

#### Link to Design Doc
NA

Id: 6000
Profile: regression

Test Summary: total: 3052, fail: 3, error: 1, timedout: 0, skip: 1, pass: 3047

### DIFF
--- a/test/tests/functional/pbs_server_periodic_hook.py
+++ b/test/tests/functional/pbs_server_periodic_hook.py
@@ -384,9 +384,8 @@ pbs.logmsg(pbs.LOG_DEBUG, "periodic hook ended at %%d" %% time.time())
         a = {'Resource_List.select': '1:ncpus=1',
              'Resource_List.walltime': 3}
         j = Job(TEST_USER, attrs=a)
-        j.set_sleep_time(3)
+        j.set_sleep_time(10)
         jid1 = self.server.submit(j)
-        self.server.expect(JOB, {'job_state': 'R'}, id=jid1)
         self.server.expect(JOB, 'queue', id=jid1, op=UNSET, offset=3)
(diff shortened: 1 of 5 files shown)


raviv1995 (approved):

test/tests/functional/pbs_server_periodic_hook.py
```diff
@@ -384,9 +384,8 @@ def test_other_pbs_operations_work(self):
         a = {'Resource_List.select': '1:ncpus=1',
              'Resource_List.walltime': 3}
         j = Job(TEST_USER, attrs=a)
-        j.set_sleep_time(3)
+        j.set_sleep_time(10)
```
bhroam, line 387: You are still submitting the job with a walltime of 3s.  PBS will kill the job.  In any case, 3s is still probably OK because you just want the job to end.  It doesn't matter if a slow machine pauses since you are no longer trying to catch the job in the 'R' state.

bhroam (changes requested):

test/tests/functional/pbs_server_periodic_hook.py
RKORavi: Yes, I agree. Thanks for catching it @bhroam

bhroam (approved):
LGTM

bhroam:
@raviv1995 let me know if you want to take another look or if I should merge.

raviv1995 (approved):
LGTM

PR #2252, opened 2021-02-11
systemd missing daemons only start

Jitendra-Rajak:
#### Describe Bug or Feature
If PBS is started using systemd and if some how any daemon get killed, systemctl status still shows pbs service running without the daemon in process list that got killed. We were not able to start that missing daemon until we restart the service all together.

#### Describe Your Change
Added a directive in the pbs.service unit file that will be used to reload pbs service by doing the reload only the missing daemon/ daemons will be started again and will be seen in the process list when we do systemctl status pbs.

[missing daemon restart.txt](https://github.com/openpbs/openpbs/files/5964855/missing.daemon.restart.txt)


Jitendra-Rajak:
Profile: regression
|--|--|--|--|--|--|--|
|6035|3887|6|0|0|0|3881|

riyazhakki (changes requested):
Attach PTL test log.
Update test_missing_daemon to test all the daemons.
Would suggest to rename pbs_wrap to pbs_reload.
How these changes will act, if more than one daemon is killed ?

riyazhakki:
How pbs_wrap script will perform, if more than one daemon is killed ?

Jitendra-Rajak:
> How pbs_wrap script will perform, if more than one daemon is killed ?

In this case when we will be doing Systemctl reload pbs, killed daemons will be started again.
please se the attached log for the same.
[more than one daemon down.txt](https://github.com/openpbs/openpbs/files/6042676/more.than.one.daemon.down.txt)

Jitendra-Rajak:
@riyazhakki , I have addressed your comments here is the PTL logs for the latest change.
[systemdptl.txt.txt](https://github.com/openpbs/openpbs/files/6042688/systemdptl.txt.txt)

riyazhakki (approved):
Thanks of adding the test logs.
Changes looks good to me

test/tests/functional/pbs_systemd.py
```diff
@@ -110,3 +112,53 @@ def test_systemd(self):
...
+        if self.mom.isUp(max_attempts=10):
+            self.logger.info("MoM started and running")
+        else:
+            self.fail("MoM not started")
+        # Sched
+        self.scheduler.signal("-KILL")
+        if self.scheduler.isUp(max_attempts=10):
+            self.fail("Sched is still running")
+        cmd = "systemctl reload pbs"
+        self.du.run_cmd(self.hostname, cmd, True)
```
bayucan, line 140: You don't need to keep reassigning the same "systemctl reload pbs" to cmd...

bayucan (changes requested):

test/tests/functional/pbs_systemd.py
Jitendra-Rajak: made the changes

bayucan (approved):
Looks good.

bayucan:
Merged #2252 into master.

PR #2253, opened 2021-02-12
Multi-Server: MN jobs crossing local server boundaries

nithinj:
#### Describe Bug or Feature
The scheduler can assign nodes reporting to different server instances to a multi-node job.

#### Describe Your Change

Server1 will send an update to book the nodes that belong to server2.

For instance, an INCR operation will be sent before running a job. A DECR will be the counter operation to decrement the resources accounted as in the case with job termination.

The update will be broadcasted using PSI values

Inter-server resource update can happen when:

→ a job gets run

→ job gets terminated

→ various other scenarios where the resource needs to be updated.

Upon receiving the request, server2 will update the job portion in the node object based on the operation. Server1 will be broadcasting the request to all the sister servers. The request will contain the following parameters:

→ jobid

→ execvnode

→ operation (INCR/DECR)

→ Job exclusivity

Challenges:

 *   Server does not know hostname and port of all the nodes where the job is being run:

when J1's run request lands on server1, it receives the execvnode which contains the list of all vnodes and how much resources is allotted in each of them. Server1 now has to translate it into exechost2, which will be hostname and port pairs instead of vnode name. This is required so that Mom1 can reach out to the other vnodes in the execvnode using the corresponding hostname:port.

As nodes are partitioned, the servers do not know the vnode to hostname:port mapping of nodes belongs to a different server.

To handle this, the server will do a node_stat() and cache the necessary information. The stat will happen only on failure to find the node either in server or join-job failure in mom (lazy pull). This will be further optimized with diff-stat.

  *  IS_CLUSTERADDR protocol cannot be reliably used for multi-server:

Even though this mechanism is not supported in multi-server mode, an attacker cannot gain control of the moms in the cluster if it is protected by any of the other supported authentication mechanisms such as TLS, munge or Kerberos.

 *   What if the resource update fails due to network failure?

Resource updates can fail due to a number of reasons including network failures and a down instance. This feature relies on the network layer to tell the application that something went wrong. If the application closes that connection channel, it will reattempt to send the resource update upon connection restoration.

 *   What happens when a server instance restarts?

Resource usage values of nodes, queue, server etc are repopulated based on the state of the job when a server is started. With multi-server, the shared objects might have some share from the job that belongs to another server. So server needs to request other servers to send any update it needs to account for. This will be done as part of the initial hello exchange.

 *   What if the scheduler does another stat before resource update reaches the sister-servers?

 This could result in an over-subscription as those nodes which need to be filled by the resource update might appear to be empty.

This is solved by pbs_server_ready which is a synchronization mechanism used by clients. pbs_server_ready is an IFL that makes sure all the servers in the cluster are ready for stat operation.

Detailed test steps and log and various scenarios which need to be automated:
[test.log](https://github.com/openpbs/openpbs/files/6127975/test.log)

Regression Results:

Description: Rerun All Tests From #6324
Platforms: UBUNTU1804,SUSE15,UBUNTU1604,CENTOS8,CENTOS7,RHEL7,SLES12,SLES15,WIN2K16
Profile: regression
|--|--|--|--|--|--|--|
|6325|3990|2|0|0|0|3988|

Description: Rerun Only Failed Tests From #6325
Platforms: UBUNTU1804,SUSE15,UBUNTU1604,CENTOS8,CENTOS7,RHEL7,SLES12,SLES15,WIN2K16
Profile: regression
|--|--|--|--|--|--|--|
|6342|2|0|0|0|0|2|

Valgrind logs:
[valgrind.log](https://github.com/openpbs/openpbs/files/6107397/valgrind.log)


nithinj:
Requesting reviews from @bhroam & @subhasisb

subhasisb (changes requested):
just started looking first few files

src/include/batch_request.h
```diff
@@ -226,6 +227,16 @@ struct rq_defschrpy {
 	char *rq_txt;
 };
 
+/* Resource update from peer server */
+struct rq_rescupdate {
+	char *jobid;
```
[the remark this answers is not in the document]
nithinj: resource update was initially part of batch request which is not the case anymore. I have removed and added this as part of server.h

src/include/pbs_nodes.h
```diff
@@ -117,8 +120,10 @@ struct mom_svrinfo {
 	struct job	**msr_jobindx;  /* index array of jobs on this Mom */
 	long		msr_vnode_pool;/* the pool of vnodes that belong to this Mom */
 	int		msr_has_inventory; /* Tells whether mom is an inventory reporting mom */
+	int		num_pending_rply; /* number of unacknowledged replied from this server */
```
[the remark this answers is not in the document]
nithinj: This comment seems off from the code I have. Do you mean to ask something else?

nithinj:
@subhasisb ,
Thanks for the initial set of comments. I have addressed them. Please have a look.

src/include/libpbs.h
```diff
@@ -338,6 +339,7 @@ struct batch_reply
 #define EXTEND_OPT_NEXT_MSG_PARAM "next_msg_param"
 
 int is_compose(int, int);
+int ps_compose(int, int);
```
subhasisb, line 342: what is ps_compose to be used for? Peer server compose? Perhaps a comment would help

src/lib/Libifl/enc_reply.c
```diff
@@ -251,17 +251,21 @@ int
 encode_DIS_replyTPP(int sock, char *tppcmd_msgid, struct batch_reply *reply)
 {
 	int rc;
-	/* first encode "header" consisting of protocol type and version */
 
-	/* since it TPP connection, write a header */
-	if ((rc = is_compose(sock, IS_CMD_REPLY)) != DIS_SUCCESS)
+	/* first encode "header" consisting of protocol type and version */
+	if (reply->brp_choice == BATCH_REPLY_CHOICE_Status) {
+		if ((rc = ps_compose(sock, PS_STAT_RPLY)) != DIS_SUCCESS)
```
subhasisb, line 257: u must be needing something like the tppcmd_msgid also for ps_compose()? If so, it is taken care of by ps_compose() itself?

src/lib/Libifl/int_rdrpy.c
```diff
@@ -83,18 +84,22 @@ PBSD_rdrpy_sock(int sock, int *rc)
 		return NULL;
 	}
 
-	DIS_tcp_funcs();
-	old_timeout = pbs_tcp_timeout;
-	if (pbs_tcp_timeout < PBS_DIS_TCP_TIMEOUT_LONG)
-		pbs_tcp_timeout = PBS_DIS_TCP_TIMEOUT_LONG;
+	if (prot == PROT_TCP) {
+		DIS_tcp_funcs();
```
subhasisb, line 88: dont u have to set it to DIS_tpp_funcs() in case of PROT_TPP? If the function pointers were pointing to TCP in an earlier call - then we would read from the wrong buffers?

Granted, the DIS_xxx_funcs() is the lousiest possible code design! We should change this as soon as we call (perhaps with the flatbuffers RFE)

src/lib/Libifl/int_rdrpy.c
```diff
@@ -83,18 +84,22 @@ PBSD_rdrpy_sock(int sock, int *rc)
...
-	if ((*rc = decode_DIS_replyCmd(sock, reply)) != 0) {
+	if ((*rc = decode_DIS_replyCmd(sock, reply, prot)) != 0) {
 		(void)free(reply);
 		pbs_errno = PBSE_PROTOCOL;
 		return NULL;
 	}
-	dis_reset_buf(sock, DIS_READ_BUF);
-	pbs_tcp_timeout = old_timeout;
+	if (prot == PROT_TCP) {
+		dis_reset_buf(sock, DIS_READ_BUF);
```
subhasisb, line 100: also why would dis_reset_buf() not be required for TPP? It clears of the read buffer from DIS layer in case something was pending - perhaps best to to leave it on for both protocols?

src/lib/Libifl/list_link.c
```diff
@@ -191,6 +189,18 @@ delete_link(struct pbs_list_link *oldp)
 	oldp->ll_prior = oldp;
 }
 
+/**
+ * @brief delete an entry from the list and clear the struct
+ *
+ * @param[in] oldp       ptr to link to delete
+ */
+void
+delete_clear_link(struct pbs_list_link *oldp)
```
subhasisb, line 198: why do we need this function? Usually the callers call delete_link(x) and then free(x). Why do we need to set the ll_struct to NULL? Is it so this object is ready to be appended/inserted to another list?

subhasisb (changes requested):
second round comments

src/lib/Libifl/dec_rpyc.c
```diff
@@ -299,16 +300,18 @@ decode_DIS_replyCmd(int sock, struct batch_reply *reply)
...
-		return rc;
-	if (i != PBS_BATCH_PROT_TYPE)
-		return DIS_PROTO;
-	i = disrui(sock, &rc);
-	if (rc != 0)
-		return rc;
-	if (i != PBS_BATCH_PROT_VER)
-		return DIS_PROTO;
+	if (prot == PROT_TCP) {
+		i = disrui(sock, &rc);
```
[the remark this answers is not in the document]
nithinj: encode_DIS_replyTPP was not encoding them. I'll change that to encode these parameters as well.

src/include/libpbs.h
nithinj: This is mentioned in the function Doxygen.

src/lib/Libifl/enc_reply.c
nithinj: Presently we don't have a case where the sender will match the msgid with the reply for peer server messages.

src/lib/Libifl/list_link.c
nithinj: free_ru will delete all the items in the list. delete_clear_link() will help me in deleting only the current item as it won't proceed after the link. I can also update free_ru to accept a new parameter just to delete the first item, but I feel this should be a capability of the list routines.

src/server/node_manager.c
```diff
@@ -6168,6 +6163,177 @@ which_parent_mom(pbsnode *pnode, mominfo_t *pcur_mom)
...
+		set_vnode_state(pnode, INUSE_JOBEXCL, Nd_State_Or);
+	} else if (pnode->nd_nsnfree <= 0) {
+		set_vnode_state(pnode, INUSE_JOB, Nd_State_Or);
+	} else {
+		set_vnode_state(pnode, ~(INUSE_JOB | INUSE_JOBEXCL), Nd_State_And);
+	}
+}
+
+/**
+ * @brief Determines job sharing type
```
subhasisb, line 6304: Same, bit more descriptive please

src/server/node_manager.c
```diff
@@ -6319,25 +6481,13 @@ set_nodes(void *pobj, int objtype, char *execvnod_in, char **execvnod_out, char
 		if (execvnod == NULL)
 			return PBSE_BADNODESPEC;
 
-		if (!strlen(execvnod))
+		if (!strlen(execvnod)) {
+			send_nodestat();
```
subhasisb, line 6485: Please describe the flow over here. That the job is being rejected, but should run in the next few cycles by which time the stat would have been answered and populated.

src/include/net_connect.h
```diff
@@ -116,7 +120,15 @@ typedef unsigned long pbs_net_t;        /* for holding host addresses */
...
-#define IS_PEERSVR_CONNECT              23 /* hello from peer server  */
+
+/*
+ **	Types of Peer Server messages (between Server and Server).
+ */
+#define PS_CONNECT		1 /* hello from peer server  */
+#define PS_RSC_UPDATE_FULL	2 /* hello from peer server  */
+#define PS_RSC_UPDATE		3 /* Peer server resource update */
+#define PS_RSC_UPDATE_ACK	4 /* Reply for peer server resource update */
+#define PS_STAT_RPLY		5 /* Asybchronous status reply */
```
subhasisb, line 131: spelling of asynchronous is wrong``

src/server/svr_comm.c
```diff
@@ -0,0 +1,413 @@
...
+ */
+int
+send_resc_usage(int c, psvr_ru_t *psvr_ru, int ct, int incr_ct)
+{
+	int rc;
+	psvr_ru_t *ru_cur = NULL;
+	server_t *psvr;
+
+	/* account messages sent */
+        for (psvr = GET_NEXT(peersvrl);
```
subhasisb, line 133: alignment issues

src/server/svr_comm.c
```diff
@@ -0,0 +1,413 @@
...
+	size_t sz;
+	int i;
+        psvr_ru_t *ru_cur = NULL;
+
+        CLEAR_HEAD((*ru_head));
+
+	ct = disrsi(sock, &rc);
+
+	for (i = 0; i < ct; i++) {
+                ru_cur = calloc(1, sizeof(psvr_ru_t));
```
subhasisb, line 187: alignment issues

src/server/svr_comm.c
```diff
@@ -0,0 +1,413 @@
...
+		ru_cur->jobid = disrcs(sock, &sz, &rc);
+		if (rc)
+                        goto err;
+
```
subhasisb, line 191: lots of alignment issues

src/server/svr_comm.c
```diff
@@ -0,0 +1,413 @@
...
+			DBPRT(("%s: stream %d from %s:%d already open on %d\n",
+				__func__, stream, pmom->mi_host,
+				ntohs(addr->sin_port), psvr_info->msr_stream))
+			tpp_close(psvr_info->msr_stream);
+			tdelete2((u_long)psvr_info->msr_stream, 0ul, &streams);
+		}
+		/* we save this stream for future communications */
+		psvr_info->msr_stream = stream;
+		psvr_info->msr_state &= ~INUSE_NEEDS_HELLOSVR;
+                tinsert2((u_long)stream, 0ul, psvr, &streams);
```
subhasisb, line 344: This file is full or alignment issues and very skimpy function headers. Please fix all of these

src/server/multi_svr.c
```diff
@@ -306,3 +524,339 @@ gen_svr_inst_id(void)
...
+							   NULL, ru_cur->execvnode);
+		update_node_rassn(&pexech, ru_cur->op);
+
+		if (ru_cur->op == DECR) {
+			delete_clear_link(&ru_cur->ru_link);
+			free_ru(ru_cur);
+		}
+	}
+
+	if (op == INCR)
```
subhasisb, line 646: Why ack in case of INCR only? DECR could also create inconsistencies!!!

src/server/process_request.c
```diff
@@ -1015,7 +1015,7 @@ dispatch_request(int sfds, struct batch_request *request)
 			break;
 
 		case PBS_BATCH_StatusNode:
-			if (set_to_non_blocking(conn) == -1) {
+			if (prot != PROT_TPP && set_to_non_blocking(conn) == -1) {
```
subhasisb, line 1018: not a comment for this time, but i could not put a comment on code you had committed in an earlier PR, hence hanging it here. the ptask structure is used only by the PBS_BATCH_ServerReady handler req_stat_svr_ready(). So why not set it inside req_stat_svr_ready() and pass only request to it. It seems contrived that we assign request to wt_parm of ptask and then pass a pointer to it. Why?

src/server/node_manager.c
```diff
@@ -3097,6 +3112,10 @@ mcast_moms(struct work_task *ptask)
 			mtfd_replyhello_noinv = -1;
 			break;
 
+		case PS_CONNECT:
+			replyhello_psvr();
```
subhasisb, line 3116: All ur acks for send_resc_usages result in decrementing the pending acks! This is fine in general. Are there cases where this count could go berserk, and when it does, what operation will reset it to a clean state? 
If a peer server was sent a resc_update() but crashed and did not respond, when it comes back up, will the pending_acks for that server be cleared and started from 0?

subhasisb (changes requested):

subhasisb:
Round 3 feedback submitted.

src/include/pbs_nodes.h
```diff
@@ -88,9 +88,12 @@ struct mominfo {
 	void	       *mi_data;	/* daemon dependent substructure */
 	mom_hook_action_t **mi_action;	/* pending hook copy/delete on mom */
 	int		mi_num_action; /* # of hook actions in mi_action */
+	pbs_list_link	mi_link; /* forward/backward links */
+	void		*mi_rsc_idx; /* avl of resc updates based on job_id as the key and psvr_ru_t as value */
+	pbs_list_head	mi_node_list; /* list of nodes corresponding to this server. Useful for clearing the nodes in case of an update */	
 };
 typedef struct mominfo mominfo_t;
-typedef struct mominfo svrinfo_t;
+typedef struct mominfo server_t;
```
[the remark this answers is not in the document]
nithinj: How about the name host_info?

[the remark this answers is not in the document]
nithinj: Perhaps creating an independent struct for the peer server should be better as most of the members in mom_svrinfo are mom-specific. I'm looking at it.

nithinj:
> so, there is not ack back to the sender? (That can't be, because if you have that, then there would be no way to fence things at the start of sched cyle, right?)
Perhaps you mean there will be only one peer message outstanding at any point of time, so you dont need to match with tpp_msgid. But how about future?

Peer server sends ack for resc update. This is a specific message and hence the sender can identify that and change the counts which are used for fencing.
In the future, if there is a requirement this needs to be introduced. That will require a change in the peer server protocol version. But the need for such a change (addition of new dialogues/change to the existing ones) might itself bring a need for changing the version.
I'm of the opinion that we can add it when there is a requirement. But let me know if you think otherwise.

src/server/node_manager.c
```diff
@@ -348,6 +347,9 @@ reply_hellosvr(int stream, int need_inv)
 	if ((ret = diswsi(stream, need_inv)) != DIS_SUCCESS)
 		return ret;
 
+	if ((ret = diswsi(stream, get_num_servers())) != DIS_SUCCESS)
```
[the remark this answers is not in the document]
nithinj: It is to confirm whether mom belongs to a single/multi-server cluster. Mom will not do IP-based (CLUSTERADDR) checks if it is part of msvr cluster. I will add a comment in the code.

src/server/node_manager.c
```diff
@@ -2898,9 +2902,14 @@ stream_eof(int stream, int ret, char *msg)
 	/* find who the stream belongs to and mark down */
 	if ((mp = tfind2((u_long)stream, 0, &streams)) != NULL) {
 		DBPRT(("%s: %s down\n", __func__, mp->mi_host))
+		log_errf(-1, __func__, "%s down", mp->mi_host);
 		if (msg == NULL)
 			msg = "communication closed";
-		momptr_down(mp, msg);
+		/* Do not invoke for a peer server */
+		if (get_peersvr_from_host_port(mp->mi_host, mp->mi_port))
+			((mom_svrinfo_t *)(mp->mi_data))->msr_state |= 	INUSE_NEEDS_HELLOSVR;
```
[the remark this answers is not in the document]
nithinj: I'm working on having a separate struct for the peer server. I'll adopt the name if that doesn't end well.

src/server/multi_svr.c
nithinj: INCR will result in over-consumption and DECR results in under-utilization. But an under-utilization can be filled in the very next scheduling cycle.
This will be similar to an under-utilization scenario in single-server mode. The job has ended and the obit has been sent. But the server processes the sched stat request before processing obit. This should be fine if the scheduler can fill those resources in the next cycle.

suresh-thelkar (changes requested):
I have started looking into the changes. Provide more comments shortly.

src/server/process_request.c
nithinj: req_stat_svr_ready replies back only if the server is in a consistent state. Otherwise, it will leave a work task with the request in it which will be converted to immediate when all acks are received. So req_stat_svr_ready needs a wrapper function that does accept request structure and set it to the ptask. I choose to do it in the caller as there is no other thing that wrapper can do. I'll leave a comment indicating this.

src/server/node_manager.c
nithinj: Counts are reset when the server receive a PS_CONNECT from one of the peer server. Then all the resource updates will be sent.

IS_CONNECT will be received in the following cases:
1. When a new server joins the cluster .
2. When an existing server gets restarted
3. When network fails and then server rejoins the cluster after the network recovery.

Application layer relies on the network layer in detecting network failure.

src/server/multi_svr.c
subhasisb: sure, that is fine. But what I am asking is that is there is a pending ack (due to not acking for DECR), then the sending server will have a "pending_acks" count right? And that would stop scheduling during the fence operation?

src/server/process_request.c
subhasisb: Well, the actual answer is that req_stat_svr_ready() is set as a work task and hence you wanted the parameter to be work_task * and not request *. I see that now and it is fine.

src/server/multi_svr.c
```diff
@@ -166,36 +169,224 @@ get_hostname_from_addr(struct in_addr addr)
...
+ * @param[in] pjob - job pointer
+ * @param[in] op - operation performed - INCR/DECR
+ * @param exec_vnode - exec_vnode string
+ * @return psvr_ru_t* 
+ * @retval NULL - on failure
+ */
+psvr_ru_t *
+init_ru(job *pjob, int op, char *exec_vnode)
+{
+	psvr_ru_t *psvr_ru = calloc(1, sizeof(psvr_ru_t));
```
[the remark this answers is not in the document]
nithinj: I'm thinking of introducing pbs_calloc which prints the mem allocation failure to log_err along with the stack trace and exit with PBSE_SYSTEM. That would be what we do eventually as PBS cannot function with memory allocation errors. Let me know if this is not acceptable.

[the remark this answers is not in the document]
subhasisb: No, don't do that. U are assuming that PBS must exit for all calloc failures. There could be cases where we try a large amount of area and if that is not available, we try smaller parts...it would not be prudent to add an exit to all calloc failures

src/server/node_manager.c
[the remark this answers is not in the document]
subhasisb: ok

src/server/multi_svr.c
nithinj: Pending ACKs are only incremented for INCR updates.

src/server/multi_svr.c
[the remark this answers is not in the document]
nithinj: I'm only adding a utility function that can be used in cases like these. We still need calloc for the scenario you have mentioned. I'm only changing this occurrence and not any other occurrence of calloc.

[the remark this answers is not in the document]
subhasisb: well don't make it a function, any log_err inside that will not log the correct function. If necessary make it a macro. But do you really need it?

src/server/multi_svr.c
subhasisb: ok

src/server/node_manager.c
subhasisb: ok sounds fine for now

subhasisb:
> > so, there is not ack back to the sender? (That can't be, because if you have that, then there would be no way to fence things at the start of sched cyle, right?)
> > Perhaps you mean there will be only one peer message outstanding at any point of time, so you dont need to match with tpp_msgid. But how about future?
> 
> Peer server sends ack for resc update. This is a specific message and hence the sender can identify that and change the counts which are used for fencing.
> In the future, if there is a requirement this needs to be introduced. That will require a change in the peer server protocol version. But the need for such a change (addition of new dialogues/change to the existing ones) might itself bring a need for changing the version.
> I'm of the opinion that we can add it when there is a requirement. But let me know if you think otherwise.

Ignore my earlier comment, this was before i went through the rest of the code

src/server/node_manager.c
subhasisb: one more question: If just the network goes down and comes back up, (the peer server did not failed and restart), then will the peer serve send PS_CONNECT to fix the counts?

subhasisb:
Please do add somewhere (in the code) a pretty detailed dscription of the pending_acks flow (for posterity)
- The design if based on counts outstanding acks of updates that went out to peer servers
- If pending acks are not down to 0, scheduler will be blocked
- only INCR's add to pending acks count
- only INCR updates are acknowledged (and why it is ok to not count/ack DECR operations)
- What happen if network breaks and comes back up again? (PS_CONNECT will happen even if only the network breaks, right?)

src/server/multi_svr.c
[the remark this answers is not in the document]
nithinj: pbs_calloc is part of the last commit. The function aborts on failure leaving the stack trace for inspection. The thought was to make the code clutter-free. macro also sounds good to me.
But if you feel there is more to think, then I can leave it for now and take it up separately.

src/server/node_manager.c
nithinj: If the application layer came to know about the network failure, it will close the peer server connection. The server will periodically attempt to re-connect and the initial exchange will get triggered upon reconnecting.

src/server/multi_svr.c
```diff
@@ -295,14 +518,367 @@ gen_svr_inst_id(void)
...
+ * to be sent to peer server in the form of resource usage / commands.
+ */
+void
+poke_peersvr(void)
+{
+	server_t *psvr;
+
+	for (psvr = GET_NEXT(peersvrl);
+	     psvr; psvr = GET_NEXT(psvr->mi_link)) {
+		connect_to_peersvr(psvr);
```
suresh-thelkar, line 573: Don't we need to collect the return value of connect_to_peersvr() ?

suresh-thelkar (changes requested):

src/server/node_manager.c
```diff
@@ -6168,6 +6160,185 @@ which_parent_mom(pbsnode *pnode, mominfo_t *pcur_mom)
...
+
+	snp = pnode->nd_psn;
+	if (hw_ncpus == 0) {
+		/* setup jobinfo struture */
+		jp = (struct jobinfo *) malloc(sizeof(struct jobinfo));
+		if (jp) {
+			jp->next = snp->jobs;
+			jp->has_cpu = 0; /* has no cpus allocatted */
+			snp->jobs = jp;
+			jp->jobid = strdup(jobid);
```
suresh-thelkar, line 6205: Please handle return value of strdup

src/server/node_manager.c
```diff
@@ -7931,6 +8033,50 @@ update_node_rassn(attribute *pexech, enum batch_op op)
 	update_job_node_rassn(NULL, pexech, op);
 }
 
+void
+update_jobs_on_node(char *jid, char *exec_vnode, int op, int share_job)
```
suresh-thelkar, line 8037: Please write Doxygen header for this function.

src/server/node_manager.c
```diff
@@ -7931,6 +8033,50 @@ update_node_rassn(attribute *pexech, enum batch_op op)
...
+			for (j = 0; j < nelem; ++j) {
+				prdef = find_resc_def(svr_resc_def, pkvp[j].kv_keyw);
+				if (prdef == NULL)
+					return;
+				/* skip all non-consumable resources (e.g. aoe) */
+				if ((prdef->rs_flags & asgn) == 0) {
+					continue;
+				}
+				if (!strcmp(prdef->rs_name, "ncpus")) {
+					ncpus = strtol(pkvp[j].kv_val, NULL, 10);
```
suresh-thelkar, line 8063: Make sure that strtol succeeds. In case of failure take the appropriate action.

src/server/req_runjob.c
```diff
@@ -320,29 +287,27 @@ get_dest(char *extend)
...
-	if (!is_same_host(destination, pbs_server_name) ||
-	    (pc && (atoi(pc + 1) != pbs_server_port_dis))) {
-		if (pc)
-			*pc = ':';
-		return destination;
+	if (!pnode) {
+		if (pnode = find_alien_node(vname))
+			svr_id = get_nattr_str(pnode, ND_ATR_server_inst_id);
+		else
+			send_nodestat();
```
suresh-thelkar, line 306: Better to rename this to send_nodestat_req()

suresh-thelkar (changes requested):

src/include/libpbs.h
nithinj: I would like to keep it similar to is_compose(). A short name helps for a function that is used quite frequently.

src/lib/Libifl/enc_reply.c
nithinj: tpp_msgid is used in finding the right task to dispatch from the deferred task list. This is useful for commands like move job/run job to execute the associated post handler function. We may not have such a use case for stat requests at least in the near future.
See: process_DreplyTPP()

src/lib/Libifl/int_status.c
```diff
@@ -641,33 +642,47 @@ PBSD_status_random(int c, int cmd, char *id, struct attrl *attrib, char *extend,
  * @brief
  *	Returns pointer to status record
  *
- * @param[in] c - index into connection table
+ * @param[in] c - connection socket
+ * @param[out] last - last batch status read
```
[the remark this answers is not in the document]
nithinj: The caller just has to pass the holding buffer. Caller can pass NULL if she does not want to read it. I'm not sure whether it qualifies as in/out.

src/lib/Libifl/pbsD_connect.c
```diff
@@ -858,20 +859,20 @@ __pbs_disconnect(int connect)
 	if (svr_conns) {
 		for (i = 0; svr_conns[i]; i++) {
 			if (disconnect_from_server(svr_conns[i]->sd) != 0)
-				return -1;
+				rc = -1;
```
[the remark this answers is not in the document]
nithinj: In that case rc will point to the last return code. Right now rc holds an error code if any of the operations fails.

src/lib/Libtpp/tpp_util.c
```diff
@@ -436,41 +437,32 @@ set_tpp_config(struct pbs_config *pbs_conf, struct tpp_config *tpp_conf, char *n
...
+			if ((*p && *p == ',') || end) {
 				*p = 0;
-				tpp_conf->routers[i++] = strdup(q);
-
-				p++; /* go past the null char and trim any spaces */
-				while (*p && (*p == ' ' || *p == '\t'))
-					p++;
-
-				q = p;
+				while (isspace(*q))
```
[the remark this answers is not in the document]
nithinj: We could, but most of this logic will still be applicable. So I thought of doing everything in one go.

src/server/multi_svr.c
```diff
@@ -40,50 +40,69 @@
...
- * @return	svrinfo_t
+ * @return	server_t
  * @retval	NULL	- Could not find peer server corresponing to the addr
  * @retval	!NULL	- svrinfo structure
  */
 void *
 get_peersvr(struct sockaddr_in *addr)
 {
-	svrinfo_t *psvr;
+	server_t *psvr;
```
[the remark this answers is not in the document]
nithinj: Why?

src/server/multi_svr.c
```diff
@@ -40,50 +40,69 @@
...
+ * @return peer server object, NULL if not found
+ */
+void *
+get_peersvr_from_host_port(char *hostname, uint port)
+{
+	server_t *psvr;
+
+	for (psvr = GET_NEXT(peersvrl);
+	     psvr; psvr = GET_NEXT(psvr->mi_link)) {
+		if (!strncmp(psvr->mi_host, hostname,
```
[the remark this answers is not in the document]
nithinj: This function is removed with the recent commits.

src/server/multi_svr.c
```diff
@@ -166,36 +168,229 @@ get_hostname_from_addr(struct in_addr addr)
...
 	return psvr;
 }
 
+/**
+ * @brief free the resource update structure
+ * 
+ * @param[in,out] ru_head - head of the resource update object list
+ */
+void
+free_ru(psvr_ru_t *ru_head)
```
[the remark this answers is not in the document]
nithinj: That's too long!
how about free_psvr_ru / delete_ru ?

src/server/multi_svr.c
```diff
@@ -166,36 +168,229 @@ get_hostname_from_addr(struct in_addr addr)
...
+ * @brief initialize resource usage stucture based on parameters
+ * 
+ * @param[in] pjob - job pointer
+ * @param[in] op - operation performed - INCR/DECR
+ * @param exec_vnode - exec_vnode string
+ * @return psvr_ru_t* 
+ * @retval NULL - on failure
+ */
+psvr_ru_t *
+init_ru(job *pjob, int op, char *exec_vnode)
```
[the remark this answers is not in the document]
nithinj: how about init_psvr_ru ?

src/server/multi_svr.c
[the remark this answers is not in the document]
subhasisb: The main question is whether there is any way to proceed without exiting the daemon. Quitting the daemon could be the last resort, but otherwise we might be able to just reject a user request, for example, if we are low on memory.

src/server/multi_svr.c
```diff
@@ -166,36 +168,229 @@ get_hostname_from_addr(struct in_addr addr)
...
+		log_errf(-1, __func__, "Resource update from unknown stream %d", conn);
+		return;
+	}
+
+	if (*pending_rply == 0 && num_pending_peersvr_rply() == 0) {
+		ptask = find_work_task(WORK_Deferred_Reply, NULL, req_stat_svr_ready);
+		if (ptask) {
+			log_event(PBSEVENT_DEBUG, PBS_EVENTCLASS_SERVER, LOG_DEBUG, __func__,
+				  "All peer server acks received. Processing pbs_server_ready");
+			convert_work_task(ptask, WORK_Immed);
```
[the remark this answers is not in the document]
nithinj: Isn't it evident from the log we are printing?

src/server/multi_svr.c
nithinj: The error is logged at the inner level functions.

src/server/multi_svr.c
```diff
@@ -295,14 +518,367 @@ gen_svr_inst_id(void)
...
+	svrinfo_t *psvr_info;
+
+	for (psvr = GET_NEXT(peersvrl); psvr; psvr = GET_NEXT(psvr->mi_link)) {
+		psvr_info = psvr->mi_data;
+		if (psvr_info->msr_stream < 0) {
+			if (connect_to_peersvr(psvr) < 0)
+				continue;
+		}
+
+		mcast_add(psvr, &mtfd);
```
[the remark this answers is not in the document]
nithinj: I've changed the log message. I'm not handling the error at this level as we need to proceed sending to other servers.

src/server/node_manager.c
```diff
@@ -7261,6 +7360,12 @@ update_job_node_rassn(job *pjob, attribute *pexech, enum batch_op op)
 			return;
 	}
 
+	if (pjob && (pjob->ji_qs.ji_svrflags & JOB_SVFLG_Broadcast_Rqd)) {
+		psvr_ru = init_ru(pjob, op, pexech->at_val.at_str);
+		mcast_resc_usage(psvr_ru);
```
suresh-thelkar, line 7365: Instead of sending resource update to all peer servers why can't we send it to the servers which need them ?

src/server/multi_svr.c
```diff
@@ -295,14 +518,367 @@ gen_svr_inst_id(void)
...
+		if (psvr_ru->op == INCR)
+			incr_ct++;
+
+		if ((ret = ps_compose(mtfd, PS_RSC_UPDATE)) != DIS_SUCCESS)
+			close_streams(mtfd, ret);
+
+		/* broadcast resc usage */
+		if ((ret = send_resc_usage(mtfd, psvr_ru, 1, incr_ct)) != DIS_SUCCESS)
+			close_streams(mtfd, ret);
+
```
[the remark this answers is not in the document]
nithinj: handled through close_streams()

src/server/multi_svr.c
```diff
@@ -295,14 +518,367 @@ gen_svr_inst_id(void)
...
+		tpp_mcast_close(mtfd);
+		mtfd = -1;
+	}
+}
+
+/**
+ * @brief adding an alien node to the cache
+ * 
+ * @param[in,out] psvr - peer server where the node belongs to
```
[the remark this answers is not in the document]
nithinj: append_link() modifies psvrinfo->ps_node_list

src/server/multi_svr.c
```diff
@@ -295,14 +518,367 @@ gen_svr_inst_id(void)
...
+		if ((decode_attr_db(pnode, &attrs, node_attr_idx, node_attr_def, pnode->nd_attr, ND_ATR_LAST, 0)) != 0) {
+			log_errf(PBSE_INTERNAL, __func__, "Decode of node %s received from peer server has failed!",
+				 pnode->nd_name);
+			free(pnode);
+			free_attrlist(&attrs);
+			pbs_statfree(bstat);
+			return -1;
+		}
+
+		add_node_to_cache(psvr, pnode);
```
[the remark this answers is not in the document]
nithinj: Isn't it fine as the function is static?

src/server/multi_svr.c
```diff
@@ -295,14 +518,367 @@ gen_svr_inst_id(void)
...
+ * @brief
+ * 		find_alien_node() - find an alien node by its name
+ * @param[in]	nodename	- node being searched
+ *
+ * @return	strcut pbsnode *
+ * @retval	!NULL - success
+ * @retval	NULL  - failure
+ */
+pbs_node *
+find_alien_node(char *nodename)
```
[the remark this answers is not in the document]
nithinj: This function doesn't expect nodes in a special format. But I have kept the same checks which were present in find_nodebyname which was skipping over a leading parenthesis and terminate at a forward slash if they were any. Let me know if you think they will be no longer relevant.

src/server/multi_svr.c
```diff
@@ -295,14 +518,367 @@ gen_svr_inst_id(void)
...
+		return pbs_errno;
+
+	switch (obj_type) {
+
+	case MGR_OBJ_NODE:
+		update_node_cache(c, bstat);
+		break;
+
+	default:
+		break;
```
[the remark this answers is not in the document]
nithinj: This is as per the examples given in the coding standards: https://openpbs.atlassian.net/wiki/spaces/DG/pages/1481670667/OpenPBS+Coding+Standards

src/server/multi_svr.c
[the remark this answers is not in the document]
suresh-thelkar: This is just to distinguish between server instance vs server from another cluster.

src/server/multi_svr.c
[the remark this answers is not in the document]
suresh-thelkar: free_psvr_ru() looks better.

src/server/multi_svr.c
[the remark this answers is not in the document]
suresh-thelkar: works for me.

src/server/multi_svr.c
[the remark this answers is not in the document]
suresh-thelkar: Better to be specific. "cache" is very generic.

src/server/multi_svr.c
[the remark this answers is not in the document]
suresh-thelkar: I think they are relevant. The Only thing is if we add an example it would be better to follow the code. In other words If it follows the same semantics as that of find_nobdebyname we can just mention the same in Doxygen header.

src/server/multi_svr.c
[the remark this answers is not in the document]
nithinj: I've changed it to handle the error now. Will reflect in the next commit.

src/server/multi_svr.c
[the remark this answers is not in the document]
nithinj: I see your point. But we should be able to make use of the same structure if we ever represent a remote server internally. You should be able to distinguish with the variable names.

src/server/multi_svr.c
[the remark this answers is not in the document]
nithinj: updated as add_node_to_psvr_cache

src/include/job.h
```diff
@@ -355,7 +355,7 @@ typedef struct ajinfo {
  */
 
 struct jbdscrd {
-	struct	mominfo *jdcd_mom;	/* ptr to Mom */
+	struct	machine_info *jdcd_mom;	/* ptr to Mom */
```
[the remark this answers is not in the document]
nithinj: This is a pointer to mom although the type is generic.

src/server/mom_info.c
```diff
@@ -310,6 +312,35 @@ find_mom_entry(char *hostname, unsigned int port)
...
+ * inlcluding mom/peer-svr
+ * 
+ * @param[in] pul - list of IP addresses of host; will be freed on error
+ *			or saved in structure; caller must not free pul
+ * @param[in] port - port of service end point 
+ * @param[in] pmi - machine info struct
+ * @return dmn_info_t* 
+ */
+dmn_info_t*
+init_daemon_info(ulong *pul, uint port, mominfo_t *pmi)
```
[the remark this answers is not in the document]
nithinj: These are neither utility functions. I can create a daemon_info.c and keep it there or leave it in the same file as these are only two functions. Let me know...

nithinj:
@subhasisb & @suresh-thelkar ,
I've addressed all your comments. Please have a look.

src/server/req_stat.c
```diff
@@ -804,6 +797,12 @@ req_stat_svr(struct batch_request *preq)
 /**
  * @brief
  * 		service the Server Ready request
+ * Replies back only if the server is in a consistent state.
+ * Otherwise, it will leave a work task with the request in it
+ * which will be converted to immediate when all acks are received.
+ * 
+ * Scheduler can proceed only when all the servers anwers to this which
```
suresh-thelkar, line 804: Typo in anwers

src/server/multi_svr.c
```diff
@@ -756,31 +852,47 @@ init_node_from_bstat(struct batch_status *bstat, server_t *psvr)
...
 	struct batch_status *cur;
+	int rc = 0;
 
 	for (cur = bstat; cur; cur = cur->next) {
+
 		pnode = calloc(1, sizeof(pbs_node));
+		if (!pnode) {
+			log_errf(PBSE_SYSTEM, __func__, "Failed to allocate memory for pnode");
+			rc = -1;
+			continue;
```
suresh-thelkar, line 860: Don't we want to return immediately in case of memory failure ?

src/server/mom_info.c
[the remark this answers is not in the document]
suresh-thelkar: It would be better if we keep them in daemon_info.c.

src/server/multi_svr.c
nithinj: trying to cache the remaining elements if the memory is available for future operations.

suresh-thelkar (approved):
Thanks for making all the changes. Code changes look good to me. 

After resolving the conflicts and all the review, please run full regressions once again just to make sure everything is fine.  Also attach the detail test scenarios and logs that you are coming up with.

src/server/req_runjob.c
```diff
@@ -386,6 +386,8 @@ req_runjob(struct batch_request *preq)
 		return; /* note, req_reject already called */
 
 	if ((dest = need_to_run_elsewhere(preq))) {
+		log_eventf(PBSEVENT_DEBUG, PBS_EVENTCLASS_JOB, LOG_INFO,
+			   jid, "Moving job to server %s in order to run", dest);
```
suresh-thelkar, line 390: Can we also print job id in the log message?

suresh-thelkar:
@nithinj, code changes still look good to me. I just left a minor comment. Please address that. 

Also, the PR is becoming unresponsive some times. If required you can open up another PR and refer this PR in that.

src/lib/Libifl/pbs_loadconf.c
```diff
@@ -257,6 +257,39 @@ parse_config_line(FILE *fp, char **key, char **val)
...
+	char *svrname;
+
+	if (!psi || !svr_id)
+		return -1;
+
+	svrname = parse_servername(svr_id, &psi->port);
+	if (svrname == NULL)
+		return -1;
+
+	strcpy(psi->name, svrname);
```
subhasisb, line 281: Could result in a buffer overflow, since the size of psi->name is smaller than what svrname can return

src/server/multi_svr.c
```diff
@@ -40,50 +40,77 @@
...
+ * @retval NULL - failure
+ */
+void *
+get_peersvr_from_svrid(char *sv_id)
+{
+	psi_t psi;
+	server_t *psvr;
+
+	if (!sv_id)
+		return NULL;
```
subhasisb, line 99: one indent more?

src/server/multi_svr.c
```diff
@@ -40,50 +40,77 @@
...
+
+	if (frame_psi(&psi, sv_id) != 0)
+		log_errf(PBSE_INTERNAL, __func__, "Failed to parse server instance id %s", sv_id);
```
subhasisb, line 102: one indent more?

src/server/multi_svr.c
```diff
@@ -40,50 +40,77 @@
...
+
+	for (psvr = GET_NEXT(peersvrl);
```
subhasisb, line 104: why break the for into two lines?

src/server/multi_svr.c
```diff
@@ -40,50 +40,77 @@
...
+	     psvr; psvr = GET_NEXT(psvr->mi_link)) {
+		if (is_same_host(psi.name, psvr->mi_host) &&
```
subhasisb, line 106: and the if?

src/server/req_manager.c
```diff
@@ -2939,10 +2939,12 @@ create_pbs_node2(char *objname, svrattrl *plist, int perms, int *bad, struct pbs
 				 * Since we are going ahead nevertheless, we need to allocate
 				 * an "empty" pul list
 				 */
+				free(pul);
```
subhasisb, line 2942: was there a leak here?

src/server/svr_comm.c
```diff
@@ -0,0 +1,445 @@
...
+	int rc;
+	psvr_ru_t *ru_cur = NULL;
+	server_t *psvr;
+	int i;
+	int count;
+	int *strms;
+
+	/* account messages sent */
+	if (psvr_ru->broadcast) {
+		for (psvr = GET_NEXT(peersvrl);
```
subhasisb, line 138: it might be better to have the for in one line for understanding

src/server/svr_comm.c
```diff
@@ -0,0 +1,445 @@
...
+	}
+
+	if ((rc = dis_flush(mtfd)) != DIS_SUCCESS) {
+		pbs_errno = PBSE_PROTOCOL;
+		goto err;
+	}
+
+	return 0;
+err:
+	log_errf(pbs_errno, __func__, "%s from stream %d",
```
subhasisb, line 172: There is no need to break lines to 80 chars these days

src/server/svr_comm.c
```diff
@@ -0,0 +1,445 @@
...
+{
+	int rc;
+	int ct;
+	size_t sz;
+	int i;
+	psvr_ru_t *ru_cur = NULL;
+
+	CLEAR_HEAD((*ru_head));
+
+	ct = disrsi(sock, &rc);
```
subhasisb, line 198: not check if this disrsi fails?

src/server/svr_comm.c
```diff
@@ -0,0 +1,445 @@
...
+			goto err;
+
+		CLEAR_LINK(ru_cur->ru_link);
+		append_link(ru_head, &ru_cur->ru_link, ru_cur);
+	}
+
+	return 0;
+
+err:
+	free_psvr_ru(ru_cur);
```
subhasisb, line 225: No log in this error block

src/server/svr_comm.c
```diff
@@ -0,0 +1,445 @@
...
+	pat = calloc(1, sizeof(struct attrl));
+	pat->name = ATTR_NODE_Port;
+	tail->next = pat;
+	tail = tail->next;
+	pat = calloc(1, sizeof(struct attrl));
+	pat->name = ATTR_server_inst_id;
+	tail->next = pat;
+	tail = tail->next;
+
+	rc = PBSD_status_put(mtfd, PBS_BATCH_StatusNode,
```
subhasisb, line 294: no need to break the line

src/server/svr_comm.c
```diff
@@ -0,0 +1,445 @@
...
+	int command;
+	struct sockaddr_in *addr;
+	svrinfo_t *psvr_info;
+	dmn_info_t *pdmn_info;
+	pbs_list_head ru_head;
+
+	DBPRT(("%s: stream %d version %d\n", __func__, stream, version))
+	addr = tpp_getaddr(stream);
+	if (version != PS_PROTOCOL_VER) {
+		log_errf(-1, __func__, "protocol version %d unknown from %s",
```
subhasisb, line 336: no need to break the line

src/server/svr_comm.c
```diff
@@ -0,0 +1,445 @@
...
+		mcast_resc_update_all(psvr);
+		return;
+	} else if ((psvr = tfind2((u_long) stream, 0, &streams)) != NULL) {
+		psvr_info = psvr->mi_data;
+		goto found;
+	}
+
+badcon:
+	sprintf(log_buffer, "bad attempt to connect from %s", netaddr(addr));
+	log_err(-1, __func__, log_buffer);
```
subhasisb, line 386: use log_errf()?

src/server/svr_comm.c
```diff
@@ -0,0 +1,445 @@
...
+	stream_eof(stream, 0, NULL);
+	return;
+
+found:
+
+	switch (command) {
+
+	case PS_RSC_UPDATE_ACK:
```
subhasisb, line 394: the case blocks should be one tab more indented

src/server/svr_comm.c
```diff
@@ -0,0 +1,445 @@
...
+			 dis_emsg[ret], psvr->mi_host, netaddr(addr));
+		stream_eof(stream, ret, "write_err");
+	} else {
+		log_errf(rc, __func__, "Error processing command %d from peer server %s",
+			 command, psvr->mi_host);
+		stream_eof(stream, ret, "read_err");
+	}
+
+	return;
+}
```
subhasisb, line 445: add a newline

src/include/libutil.h
```diff
@@ -219,6 +219,8 @@ char *pbs_fgets_extend(char **pbuf, int *pbuf_size, FILE *fp);
 extern int pbs_asprintf(char **dest, const char *fmt, ...);
 extern char *pbs_asprintf_format(int len, const char *fmt, va_list args);
 
+extern void *pbs_calloc(size_t n, size_t amt);
```
subhasisb, line 222: remove this extern, you have removed the usage already

subhasisb (changes requested):

subhasisb:
Can you think of any statistical information that can be logged for this change. Some ideas could be like:
- number of node cache misses per day
- time spent in refreshing cache per day
- number of peer server communications done per day
- number of times when acks were not received by the next sched cycle

nithinj:
@suresh-thelkar & @subhasisb ,
I've attached various test scenarios which can be automated as test.log
Please have a look.

src/include/pbs_error.h
```diff
@@ -288,6 +288,7 @@ extern "C" {
 #define PBSE_HISTDEPEND  15229		/* Finished job did not satisfy dependency */
 #define PBSE_SCHEDCONNECTED	15230
 #define PBSE_NOTARRAY_ATTR  15231		/* Not an array job */
+#define PBSE_UNKOBJ	15232		/* Named object is not in the list or alien cache */
```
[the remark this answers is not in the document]
nithinj: This error code is used in places where I need to distinguish between the two scenarios. Whether an object is owned by the instance versus whether the object is known to the instance. PBSE_UNKOBJ indicates the latter. See: update_job_node_rassn()
Only used internally

src/include/pbs_nodes.h
```diff
@@ -75,41 +75,48 @@ enum nodeattr {
...
  */
 
-struct mominfo {
+struct machine_info {
 	char		mi_host[PBS_MAXHOSTNAME+1]; /* hostname where mom is */
 	unsigned int	mi_port;	/* port to which Mom is listening */
 	unsigned int	mi_rmport;	/* port for MOM RM */
 	time_t		mi_modtime;	/* time configuration changed */
-	void	       *mi_data;	/* daemon dependent substructure */
+	dmn_info_t	*mi_dmn_info;	/* daemon specific data which are common for all */
```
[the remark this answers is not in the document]
nithinj: In fact, all the elements within machine_info struct are service end point-specific information. I can rename the struct to service_info but that will also require renaming all prefixes from mi_* to si_*.

machine_info is being used by both server and mom. The server used it to represent moms and peer-servers and the mom uses it to represent peer-moms.
mi_data contains daemon-dependent sub-structure. So it is different to represent a mom (mom_svrinfo_t), a peer-svr (svrinfo_t) and a peer-mom (mom_vninfo_t, when used inside mom code).
mi_dmn_info represents the elements that are common for both mom and peer-svr. and only used within server code.

Let me know what do you think about renaming?

src/include/pbs_nodes.h
```diff
@@ -75,41 +75,48 @@ enum nodeattr {
...
+	void		*mi_data;	/* daemon dependent substructure */
```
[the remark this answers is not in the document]
nithinj: These are used in hook_func.c which is linked to both mom and server.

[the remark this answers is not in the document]
nithinj: Looks like it is only used in the server part of the code and hence can be moved to mom_svrinfo_t. Copying @bayucan  for his view.

subhasisb:
@nithinj  please address my latest arounds, and we should be able to get this committed

src/server/multi_svr.c
nithinj: the "if" has been indented with one tab and "return" is indented with two.

src/server/multi_svr.c
nithinj: same as above, it looks fine in my browser and editor!

src/server/multi_svr.c
nithinj: This line is lengthier than the block it is enclosed in. looks cleaner when broken to two.

src/server/req_manager.c
nithinj: added based on code, as I was touching make_host_addresses_list(). I don't think I have seen a leak during testing.

src/server/svr_comm.c
nithinj: logged in the caller: ps_request()

src/server/svr_comm.c
nithinj: This is in accordance with the coding guidelines.

src/lib/Libifl/pbs_loadconf.c
```diff
@@ -278,7 +278,7 @@ frame_psi(psi_t *psi, char *svr_id)
 	if (svrname == NULL)
 		return -1;
 
-	strcpy(psi->name, svrname);
+	strncpy(psi->name, svrname, sizeof(psi->name));
```
hirenvadalia, line 281: AFAIK, now a days we are using pbs_strncpy() instead strncpy() in new/modified code

subhasisb (approved):

subhasisb:
I approved. The PR however is very sluggish on github, so perhaps u can close this and create another the actual merge. And you can add the test scenarios to that new PR

nithinj:
Raised a new PR: https://github.com/openpbs/openpbs/pull/2304
Closing this one.

PR #2255, opened 2021-02-16
Supporting Reservations for Multi-Server

suresh-thelkar:
#### Describe Bug or Feature

This PR consists of the first set of changes towards supporting reservations for the Multi-Server. As a first step, we are making reservations to be contained with in a partition/server instance i.e. reservations should not span across multiple partitions. This is being done with the spirit of agile one step at a time.  Links to the design document and the forum discussion are given below. 

**Design Document** -> https://openpbs.atlassian.net/wiki/spaces/PD/pages/2461171713/Making+Reservations+work+for+Multi-Server
**Forum Discussion** -> 
https://community.openpbs.org/t/design-for-making-reservations-work-for-multi-server/2440
#### Describe Your Change

- Extending jobid namespace changes to reservations as they are sharded
- Refactoring get_job_resv_location_hint() to act on job/resv id even if they may or may not contain PBS_SERVER information appended to it. 
- **Submitting reservations**
  - **Advanced or Standing Reservations**
     - Route the submission request to a random server instance
  - **Job-Specific reservations**
     - There are a couple of reservations that belong to this category i.e. job-specific ASAP and job-specific Now reservations. In order to support these type of reservations with Multi-Server, we first need to find out the partition or location of the given job i.e. we need to find out the server instance which has this job. To know this we actually do a “job stat” for the given job to fetch the server instance id of the job.  Since this information is enough we only request one attribute i.e. ATTR_server_inst_id.
     - Next route the reservation submission request to that server instance with the help of ATTR_server_inst_id found in the above step.
     - As job-specific start reservation is submitted through qsub, the corresponding reservation gets created wherever the job lands in. 
     Note: Admin for the time being needs to use these type or reservations only for the jobs that do not span across partitions/server intances.
- **Submitting jobs to a reservation - Approach followed**
   - If jobs are submitted to a non-default queue and if this queue is a reservation queue, we will find out the server instance which has this queue and route the job submission request to it. 
   - Else if jobs are submitted to default queue, we select a random server as usual and send the job submission request to it. 
- **Modifying reservations**
   We first try to identify the server instance where the given reservation belongs to by the help of get_job_resv_location_hint(). Only in case if this hint is not correct we try other server-instances where one of them will accept it and serves the modify request.
- **Deleting reservations** 
   Here also we follow the similar approach that we have for modifying the reservations. 
- **Scheduler side changes:**
  Scheduler confirms the reservation only to the server instance which has the reservation. This is possible since reservation has its owning server_instance_id.
- Refactoring generate_objid() to work for reservations as well.   
- **Miscellaneous bug fixes**
Fixing a couple of bugs which are specific to Multi-Server. They are not directly related to the changes made in this PR. But the code fixes for them lies around the functions which I am modifying in this PR and hence fixed them in the same PR for better testing.
   1.  get_job_location_hint() on a given job id is not returning the right server index at times. This makes the respective IFL to loop through all servers to find the right one which is a performance hit. 
   2. PBSD_manager() operation(altering/holding/deleting on a job) on jobid cannot stop immediately(exit this function) even after success i.e. after finding the right server instance. It still continues the operation with the other server instances which is not desired and error-prone. 

#### Link to Design Doc

https://openpbs.atlassian.net/wiki/spaces/PD/pages/2461171713/Making+Reservations+work+for+Multi-Server

The following are the details of testing performed on these code changes.
- Full PBS regression. Results are given below.

|--|--|--|--|--|--|--|
|6088|3887|2|0|0|0|3885|

I have rerun the couple of test cases failed. Both of them passed. 
|--|--|--|--|--|--|--|
|6094|2|0|0|0|0|2|

- Following is the list of scenarios that are tested for Multi-Server. More details can be seen in the logs attached.
   - Submission of advanced/standing reservations. Make sure that the submission hits one of the random servers without any issues.
   - Submitting multiple reservations and verify reservations stats
      - reservations stats with single reservation id
      - reservations stats with multiple reservations ids
      - all reservation stats
      - Use -f option for all the above
      - Making sure that reservation stats work even if one of the server instances is down. It should list reservations from all other active instances instead. It should also tell which server  instances are down
- Submitting a reservation when one of the server instances is down
- Test reservation queue stats with given reservation queue id
- Reservation queue stats with multiple reservation queue ids
- Testing the above couple of scenarios when one of the server instances is down
- Altering reservation/s. Make sure that all of these work successfully
- Modifying reservation/s. Make sure that all of these work successfully.
- Alter/modify reservations when one of the server instances is down. Requests for the active server instances should be honored.
- Deleting single reservation
- Deleting multiple reservations
- Deleting reservations when one of the server instances is down. Requests for the active server instances should be honored.
- Verifying most of the above test scenarios when pbs server name is not given as part of reservation id/s. This should also work since jobid namespace changes are extended to reservations also.
- Job-specific reservations
      a) Job-specific ASAP reservation
      b) Job-specific Now reservation
      c) Job-specific Start reservation

**Detailed test logs:** 
[Reservations_MultiServer_Testing.log](https://github.com/openpbs/openpbs/files/5988210/Reservations_MultiServer_Testing.log)


suresh-thelkar:
I am also in the process of running valgrind and make sure that there are no new issues.

src/lib/Libifl/int_status.c
```diff
@@ -572,6 +573,15 @@ PBSD_status_aggregate(int c, int cmd, char *id, void *attrib, char *extend, int
 					cur = last;
 				}
 			}
+		} else {
+			if (!single_itr && (pbs_errno == PBSE_UNKQUE || pbs_errno == PBSE_UNKRESVID)) {
```
agrawalravi90, line 577: combine the 'else' and the 'if' to a 'else if'

src/lib/Libifl/int_status.c
```diff
@@ -572,6 +573,15 @@ PBSD_status_aggregate(int c, int cmd, char *id, void *attrib, char *extend, int
...
+				if (pbs_errno_clear_cnt < (nsvrs - 1)) {
+					pbs_errno = 0;
```
agrawalravi90, line 579: use PBSE_NONE instead of 0

src/lib/Libifl/pbsD_modify_resv.c
```diff
@@ -67,33 +67,50 @@ char *
...
+	rc = pbs_verify_attributes(random_srv_conn(svr_conns), PBS_BATCH_ModifyResv,
 		MGR_OBJ_RESV, MGR_CMD_NONE, attrib);
 	if (rc)
 		return NULL;
 
-	/* lock pthread mutex here for this connection
-	 * blocking call, waits for mutex release */
-	if (pbs_client_thread_lock_connection(c) != 0)
-		return NULL;
+	if (svr_conns) {
```
agrawalravi90, line 87: seems like there's nothing that can be done if svr_conns is NULL, so just check for if (!svr_conns) instead and return NULL

src/lib/Libifl/int_modify_resv.c
```diff
@@ -77,13 +89,17 @@ PBSD_modify_resv(int connect, char *resv_id, struct attropl *attrib, char *exten
 		(rc = encode_DIS_ReqExtend(connect, extend))) {
 			if (set_conn_errtxt(connect, dis_emsg[rc]) != 0) {
 				pbs_errno = PBSE_SYSTEM;
+				(void)pbs_client_thread_unlock_connection(connect);
```
agrawalravi90, line 92: Remove the leading (void) (here and below)

src/lib/Libifl/pbsD_modify_resv.c
```diff
@@ -67,33 +67,50 @@ char *
...
+			if (!svr_conns[i] || svr_conns[i]->state != SVR_CONN_STATE_UP)
+				continue;
+
+			/*
+			 * For a single server cluster, instance fd and cluster fd are the same. 
+			 * Hence breaking the loop.
+			 */
+			if (svr_conns[i]->sd == c)
+				/* initiate the modification of the reservation  */
+				return PBSD_modify_resv(svr_conns[i]->sd, resv_id, attrib, extend);
```
agrawalravi90, line 102: You could have done this much earlier in the function right? I'd suggesting moving it before the call to get_job_resv_location_hint(), just check if (svr_conns[0]->sd == c) and return PBSD_modify_resv()

src/lib/Libifl/pbsD_submit.c
```diff
@@ -219,6 +219,22 @@ __pbs_submit(int c, struct attropl  *attrib, char *script, char *destination, ch
 			extend = EXTEND_OPT_IMPLICIT_COMMIT;
 	}	
 
+	if (destination != NULL && msvr_mode()) {
+		/* 
+		 * Reached here means job is submitted to non default queue
+		 * Also check if it is a reservation queue
+		 */
+		if (destination[0] == 'R' || destination[0] == 'S' || destination[0] == 'M') {
+			int start = 0;
```
agrawalravi90, line 228: this is not a good check, any queue can start with 'R', 'S' or 'M', not just reservation queues, right? You don't really need this check, just pass destination as it is to get_job_resv_location_hint(), if it returns -1, just proceed normally, I'm guessing the job will be submitted to a random server. get_job_resv_location_hint() should not fail if the queue name is a valid reservation queue.

agrawalravi90: Also, what will happen to existing reservations when a site switches from single to multi-server? How will this code work with old reservation ids? (which might be owned by any one of the N servers)

src/lib/Libifl/pbsD_submit_resv.c
```diff
@@ -70,9 +70,64 @@ __pbs_submit_resv(int c, struct attropl *attrib, char *extend)
...
 	int rc;
 	char *ret;
+	svr_conn_t **svr_conns;
+	char job_id_resv[PBS_MAXJOBNAME + 1];
 
-	for (pal = attrib; pal; pal = pal->next)
+	job_id_resv[0] = '\0';
+	for (pal = attrib; pal; pal = pal->next) {
 		pal->op = SET;		/* force operator to SET */
+		if (msvr_mode()) {
```
agrawalravi90, line 79: instead of calling this in a loop, call it before the loop and store it in a local var

src/lib/Libifl/pbsD_submit_resv.c
```diff
@@ -70,9 +70,64 @@ __pbs_submit_resv(int c, struct attropl *attrib, char *extend)
...
+					"",
+					SET
+				}	
+			};
+
+			ss = pbs_statjob(c, job_id_resv, attribs, NULL);
+			if (ss == NULL)
+				return NULL;
+
+			while (ss != NULL) {
```
agrawalravi90, line 108: there should only be one object in 'ss', so this loop is not needed.

src/server/req_quejob.c
```diff
@@ -271,27 +271,40 @@ get_server_index(void)
...
+ * @param[in]	objtype - object type (0 for normal job, 1 for job array, 2 for reservation)
+ * @param[in]	resv_char - character representing type of reservation
  *
  * @return	int
  * @retval	0 for Success
  * @retval	1 for Failure
  */
 static int
-generate_jobid(char *jidbuf, char *clusterid, int jtype)
+generate_objid(char *idbuf, char *clusterid, int objtype, char resv_char)
```
agrawalravi90, line 284: to keep the naming consistent, please rename "get_job_resv_location_hint" to "get_obj_location_hint"

src/server/req_quejob.c
```diff
@@ -271,27 +271,40 @@ get_server_index(void)
...
-	if (jidbuf == NULL || server_name == NULL)
+	if (idbuf == NULL || server_name == NULL)
 		return 1;
 
 	if (get_num_servers() <= 1) { /* single server setup */
-		if (jtype == 0)
-			sprintf(jidbuf, "%lld.%s", next_svr_sequence_id, clusterid);
-		else
-			sprintf(jidbuf, "%lld[].%s", next_svr_sequence_id, clusterid);
+		if (clusterid != NULL) {
```
agrawalravi90, line 292: clusterid will never be NULL, we don't need to handle that case (here and below)

src/server/req_quejob.c
```diff
@@ -2199,9 +2224,10 @@ req_resvSub(struct batch_request *preq)
 	 * but the structure field would be an addition to the
 	 * "quick save" area of the server - can't do
 	 */
-
-	(void)snprintf(qbuf, sizeof(qbuf), "%c%lld", PBS_RESV_ID_CHAR,
-			next_svr_sequence_id);
+	if (generate_objid(qbuf, NULL, 2, PBS_RESV_ID_CHAR) != 0) {
```
agrawalravi90, line 2227: This generates a queue name for the reservation, so just do a strchr on 'rid' and dup the portion before the '.' instead of adding the big if (!clusterid) blocks inside generate_objid().

agrawalravi90 (changes requested):

src/lib/Libifl/pbsD_submit_resv.c
nithinj, line 79: Use multi_svr_op() instead, here and other places...

src/lib/Libifl/pbsD_modify_resv.c
```diff
@@ -67,33 +67,50 @@ char *
 pbs_modify_resv(int c, char *resv_id, struct attropl *attrib, char *extend)
 {
 	struct attropl *pal = NULL;
-	int		rc = 0;
-	char		*ret = NULL;
+	int rc = 0;
+	char *ret = NULL;
+	int i;
+	svr_conn_t **svr_conns = get_conn_svr_instances(c);
```
nithinj, line 73: you need to also handle the case where the IFL is invoked with the instance fd. In that case get_conn_svr_instances(c) will return NULL.

src/lib/Libifl/int_manager.c
```diff
@@ -172,8 +172,20 @@ PBSD_manager(int c, int rq_type, int command, int objtype, char *objname, struct
 						objname,
 						aoplp,
 						extend);
-			if (rc && objtype == MGR_OBJ_JOB && pbs_errno != PBSE_UNKJOBID)
-				break;
+
+			if (objtype == MGR_OBJ_JOB) {
+				if (rc == 0)
+					break;
```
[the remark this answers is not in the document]
agrawalravi90: please address this comment

agrawalravi90 (changes requested):
I think you haven't addressed the previous round's comments yet, please let me know once you have

suresh-thelkar:
> I think you haven't addressed the previous round's comments yet, please let me know once you have

Thanks for the review. I am in the process of addressing all your/Nithin's comments and let you know once done so that you can provide further feedback. I actually found a couple of more bugs(one of which is critical in Multi-Server) while I am testing. Even though this bug is not related to my code changes, we cannot test the complete functionality of reservations without fixing this and hence started fixing these also as part of the same PR. I will let you know after I push a commit with these changes. Meanwhile, I am marking this as WIP till I push the next commit.

bhroam:
I have no problem with choosing a small bite in the spirit of agile, but I think you have chosen too small of a MVP.  I have no problem with containing a reservation within a server, but I completely disagree that it has to be contained in the random server that is chosen for the reservation.  When a reservation is denied, the submitter has no way to know if it really could be confirmed or not.  Does the submitter try and modify the times or select and try again or just submit the same reservation multiple times and hope it gets submitted to all the servers?  If they do modify it and it is randomly submitted to a busier server, they once again don't know if their modified request was denied because it can't be confirmed, or just on that random server.  I think you need to check if the reservation can be confirmed within any one server, and if it can, move the reservation over there.  I understand we don't have the ability to move reservations, but it shouldn't be hard to write.  A qmove is basically a qsub to another server.  A reservation move will just be a reservation submit to another server.

subhasisb:
I agree with @bhroam  - as this stands it is not very useful by itself, so we need to complete the "spanning/moving" right after. I think we are reading too much into "MVP" - by that, we are only talking about being able to have a Alpha level code to some available to be tried out. Right after this PR is merged, we should take up the "spanning/moving" reservations change, so that reservations become useful to end-users.

suresh-thelkar:
> I agree with @bhroam - as this stands it is not very useful by itself, so we need to complete the "spanning/moving" right after. I think we are reading too much into "MVP" - by that, we are only talking about being able to have Alpha level code available to be tried out. Right after this PR is merged, we should take up the "spanning/moving" reservations change, so that reservations become useful to end-users.

Thanks Subhasis. I will go ahead and proceed as you suggested.

src/lib/Libifl/ifl_util.c
```diff
@@ -145,30 +146,41 @@ random_srv_conn(svr_conn_t **svr_conns)
  * @retval < 0: could not find appropriate index
  */
 int
-get_job_location_hint(char *jobid)
+get_job_resv_location_hint(char *job_resv_id)
```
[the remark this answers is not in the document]
suresh-thelkar: I have refactored the function to work on either job id/resv id.  Also added a new parameter to specify the object type. 
This function also considers a job/resv id even if it is not qualified with PBS_SERVER name. Please look at my latest commit for more details.
To answer your other question, if we pass "0.pbspro" for single server configuration, the function returns early since we are not in Multi-Server mode. 
If we pass "0.pbspro" for multi-server configuration it returns -1 since the length of object id should be minimum >=2

src/lib/Libifl/ifl_util.c
```diff
@@ -135,8 +135,9 @@ random_srv_conn(svr_conn_t **svr_conns)
 }
 
 /**
- * @brief	Get the server instance index from the jobid, which will act as a hint
- * 			as to which server the job might possibly be located
+ * @brief	Get the server instance index from the job/resv id, which will act as a hint
+ * 		as to which server the job/reservation might possibly be located
+ * 		This function also works even if object id does not contain pbs_server name
```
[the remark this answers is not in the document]
suresh-thelkar: I have added more examples in the function header. Please check my latest commit.

src/lib/Libifl/int_manager.c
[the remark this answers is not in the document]
suresh-thelkar: > this changes logic, we were not breaking before with rc == 0, are you sure this is what you intended to do?

This was actually a bug in the earlier code. In case of a job/resv object if we get rc == 0 we should quit from the function. This is because we found the right server which serves the request. We don't need to iterate over other servers. This is actually a performance bug and I have fixed it in my latest commit. Please take a look at my latest commit.

src/lib/Libifl/int_status.c
suresh-thelkar: Done

src/lib/Libifl/pbsD_modify_resv.c
suresh-thelkar: I have added more code to this function. Please check the latest changes.

src/lib/Libifl/pbsD_submit.c
suresh-thelkar: > this is not a good check, any queue can start with 'R', 'S' or 'M', not just reservation queues, right? You don't really need this check, just pass destination as it is to get_job_resv_location_hint(), if it returns -1, just proceed normally, I'm guessing the job will be submitted to a random server. get_job_resv_location_hint() should not fail if the queue name is a valid reservation queue.

Please look into my latest commit. I have refactored the places where we check if it is resv or not.  Since we removed the above check now even if somebody submits a job to non-default queue we still call get_shard_obj_location_hint(). As this is not a very costly function with respect to performance I feel it is ok and that we are doing this in the client side not on the server side.

suresh-thelkar: > Also, what will happen to existing reservations when a site switches from single to multi-server? How will this code work with old reservation ids? (which might be owned by any one of the N servers)

I think a similar case arises for jobs as well. When we come up with an approach to handle upgrades this will be taken care of. I will make a note on it.

src/server/req_quejob.c
suresh-thelkar: I now named this function as get_shard_obj_location_hint(). This is very generic and since we need this function only for sharded objects like jobs/resvs, I have named it like this.

src/lib/Libifl/pbsD_modify_resv.c
suresh-thelkar: Thanks Nithin. I have handled this as well. Please look into my latest commit.

src/lib/Libifl/ifl_util.c
```diff
@@ -146,7 +152,7 @@ random_srv_conn(svr_conn_t **svr_conns)
  * @retval < 0: could not find appropriate index
  */
 int
-get_job_resv_location_hint(char *job_resv_id)
+get_shard_obj_location_hint(char *job_resv_id)
```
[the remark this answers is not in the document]
suresh-thelkar: Answered. Please check my responses in the above comments.

src/lib/Libifl/pbsD_submit.c
```diff
@@ -219,14 +219,14 @@ __pbs_submit(int c, struct attropl  *attrib, char *script, char *destination, ch
 			extend = EXTEND_OPT_IMPLICIT_COMMIT;
 	}	
 
-	if (destination != NULL && msvr_mode()) {
+	if (destination != NULL && multi_svr_op(c)) {
 		/* 
 		 * Reached here means job is submitted to non default queue
 		 * Also check if it is a reservation queue
 		 */
 		if (destination[0] == 'R' || destination[0] == 'S' || destination[0] == 'M') {
```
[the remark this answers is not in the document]
suresh-thelkar: I have addressed your comments. Please take a look at my lates commit.

suresh-thelkar:
During my testing I have found the following bug fixes which are SPECIFIC to Multi-Server. I have fixed all of them even though most of them are not related to the changes made in this PR. 

Since they are around the functions which I am modifying as part of this PR and also for better coverage of Multi-Server testing, I have fixed it here itself. 

1. Stats for some queues are missing while querying queues which results in some jobs not scheduled.
    For example, if a server-instance1 has queues "workq,R1" and server-instsance2 has "workq,R2,R3" then at times when 
    we query queues one of the queues is missing from the stat. 
2. get_job_location_hint() on a given job id is not returning the right server index at times. This makes the respective IFL 
    to loop through all servers to find the right one which is a performance hit. 
3. PBSD_manager() operation(altering/holding/deleting on a job) on jobid cannot stop (exit this function) even after 
    success with the right server instance id but continues with others as well. This is actually a performance bug.
4. Job submission some times does not work when we use qsub with credentials.
5. We cannot perform altering/holding/deleting etc on a job object just by using job-id without PBS_SERVER name in it. 
For example when we try to suspend a job by using its id the job is not getting suspended.
```
[root@spark pbspro]# qstat -s

spark, spark2: 
                                                            Req'd  Req'd   Elap
Job ID          Username Queue    Jobname    SessID NDS TSK Memory Time  S Time
--------------- -------- -------- ---------- ------ --- --- ------ ----- - -----
000.spark       root     workq    STDIN       19855   1   1    --    --  R 00:00
   Job run at Tue Feb 23 at 08:57 on (spark:ncpus=1)
001.spark2      root     workq    STDIN       66206   1   1    --    --  R 00:00
   Job run at Tue Feb 23 at 08:57 on (spark2:ncpus=1)
[root@spark pbspro]# qsig -s suspend 001
[root@spark pbspro]# qstat -s

spark, spark2: 
                                                            Req'd  Req'd   Elap
Job ID          Username Queue    Jobname    SessID NDS TSK Memory Time  S Time
--------------- -------- -------- ---------- ------ --- --- ------ ----- - -----
000.spark       root     workq    STDIN       19855   1   1    --    --  R 00:00
   Job run at Tue Feb 23 at 08:57 on (spark:ncpus=1)
001.spark2      root     workq    STDIN       66206   1   1    --    --  R 00:00
   Job run at Tue Feb 23 at 08:57 on (spark2:ncpus=1)

```
It actually works if we give fully qualified job id like qsig -s suspend 101.spark2.

suresh-thelkar:
@agrawalravi90 / @nithinj, thanks for the review. I have addressed all your comments. 

Also fixed five more bugs which are specific to Multi-Server. More details are provided in the above GitHub section. Please take a look at these changes also and provide your further feedback.

agrawalravi90:
@suresh-thelkar please don't rebase your branch in the middle of code review, I now have no reference of what i've reviewed so far and need to review your whole PR again. Be kind to your reviewers, the CI tool not running is no matter, conflicts can be resolved at the end when you have sign-offs

suresh-thelkar:
> @suresh-thelkar please don't rebase your branch in the middle of code review, I now have no reference of what i've reviewed so far and need to review your whole PR again. Be kind to your reviewers, the CI tool not running is no matter, conflicts can be resolved at the end when you have sign-offs

I did not squash the commits and so was assuming that you can still be able to see the changes and other reason is there were conflicts, not able to run CI and complete regressions without that. If you feel we can do these activities after the sign-offs I am fine with that.

src/server/req_quejob.c
```diff
@@ -285,26 +285,16 @@ generate_objid(char *idbuf, char *clusterid, int objtype, char resv_char)
...
-				sprintf(idbuf, "%lld", next_svr_sequence_id);
-			else if (objtype == 1)
-				sprintf(idbuf, "%lld[]", next_svr_sequence_id);
-			else if (objtype == 2)
-				sprintf(idbuf, "%c%lld", resv_char, next_svr_sequence_id);
-		}
-
+		if (objtype == 0)
+			sprintf(idbuf, "%lld.%s", next_svr_sequence_id, clusterid);
+		else if (objtype == 1)
```
nithinj, line 294: Can you use macros for objtype? There are macros defined in pbs_ifl.h for object types.

nithinj (changes requested):

src/lib/Libifl/ifl_util.c
```diff
@@ -115,17 +115,23 @@ get_available_conn(svr_conn_t **svr_conns)
...
+ * @param[in] fd - virtual fd or the actual fd
  * @param[in] svr_conns - pointer to array of server connections
  *
  * @return int
  * @retval -1: error
  * @retval != -1: fd corresponding to the connection
  */
 int
-random_srv_conn(svr_conn_t **svr_conns)
+random_srv_conn(int fd, svr_conn_t **svr_conns)
```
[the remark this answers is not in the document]
suresh-thelkar: As discussed, most of the places where we call random_srv_conn() ,  svr_conns is already present and hence I used this as a parameter which avoids calculating it again inside the function.

src/lib/Libifl/int_manager.c
```diff
@@ -173,19 +173,13 @@ PBSD_manager(int c, int rq_type, int command, int objtype, char *objname, struct
...
 					break;
-				else if (pbs_errno != PBSE_UNKJOBID)
-					break;
-			}
-
-			if (objtype == MGR_OBJ_RESV) {
-				if (rc == 0)
-					break;
-				else if (pbs_errno != PBSE_UNKRESVID)
+				else if (pbs_errno != PBSE_UNKJOBID || pbs_errno != PBSE_UNKRESVID)
```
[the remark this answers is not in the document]
suresh-thelkar: It was proper before we clubbed the conditions. Missed it while making it a single check. Thanks for catching this. I will add it to my next commit.

src/server/req_quejob.c
suresh-thelkar: There were no macros used earlier. May be because it used to act on only job/array job. Now that we introduced the reservation object also into this function, it would be better to use a macro. As discussed I am now making use of the macros MGR_OBJ_JOB/MGR_OBJ_JOBARRAY_PARENT/MGR_OBJ_RESV which makes it more readable.

suresh-thelkar:
@nithinj, thanks for the review. I have implemented your comments. Please look into my latest commit and provide further feedback.

src/lib/Libifl/pbsD_submit.c
```diff
@@ -219,6 +218,15 @@ __pbs_submit(int c, struct attropl  *attrib, char *script, char *destination, ch
 			extend = EXTEND_OPT_IMPLICIT_COMMIT;
 	}	
 
+	c = random_svr_conn;	
+	if (msvr && destination) {
+		/* Reached here means job is submitted to non default queue */
+		int start;
+
+		if ((start = get_shard_obj_location_hint(destination, MGR_OBJ_RESV)) != -1)
```
agrawalravi90, line 226: Please add a comment here explaining that since this could be a reservation queue and reservation queues are not shared, we try to find out which server this queue resides on.

src/server/req_quejob.c
agrawalravi90: then you should name this "generate_shard_objid" since the same logic applies here. TBH, I think it is clear enough even if you don't use the word 'shard'. And a simple if condition inside generate_obj_loation_hint() can ensure that it handles other objects as well (which will also help you remove similar if conditions before calling generate_obj_loation_hint())

src/lib/Libifl/pbsD_submit_resv.c
```diff
@@ -70,9 +70,64 @@ __pbs_submit_resv(int c, struct attropl *attrib, char *extend)
...
+	char job_id_resv[PBS_MAXJOBNAME + 1];
+	int msvr = multi_svr_op(c);
 
-	for (pal = attrib; pal; pal = pal->next)
+	job_id_resv[0] = '\0';
+	for (pal = attrib; pal; pal = pal->next) {
 		pal->op = SET;		/* force operator to SET */
+		if (msvr) {
+			if (strcmp(pal->name, ATTR_convert) == 0 || strcmp(pal->name, ATTR_resv_job) == 0) {
+				strcpy(job_id_resv, pal->value);
```
agrawalravi90, line 81: Can you please confirm that ATTR_convert  and ATTR_resv_job are mutually exclusive and that ATTR_convert's, or qmove's, value is actually a jobid? Please add a comment here to explain what this if condition does.

src/lib/Libifl/pbsD_submit_resv.c
```diff
@@ -70,9 +70,64 @@ __pbs_submit_resv(int c, struct attropl *attrib, char *extend)
...
+							pbs_statfree(ss);
+							return NULL;
+						}
+						break;
+					}
+				}
+			}    
+			c = get_svr_inst_fd(c, svr_inst_id);
+			free(svr_inst_id);
+			pbs_statfree(ss);
```
agrawalravi90, line 122: can you please put this in a separate function? something like get_resvjob_svr_inst_id()? it'll make this function look much cleaner.

src/scheduler/resv_info.cpp
```diff
@@ -1710,7 +1711,7 @@ confirm_reservation(status *policy, int pbs_sd, resource_resv *unconf_resv, serv
 		snprintf(confirm_msg, LOG_BUF_SIZE, "%s:partition=%s", PBS_RESV_CONFIRM_SUCCESS,
 			 sc_attrs.partition ? sc_attrs.partition : DEFAULT_PARTITION);
 
-		pbsrc = pbs_confirmresv(pbs_sd, const_cast<char *>(nresv_parent->name.c_str()), short_xc,
+		pbsrc = pbs_confirmresv(get_svr_inst_fd(pbs_sd, nresv_parent->svr_inst_id), const_cast<char *>(nresv_parent->name.c_str()), short_xc,
```
agrawalravi90, line 1714: Please add a sched level wrapper function for pbs_confirmresv so that the caller doesn't have to worry about multi-server stuff. Refer to sched_ifl_wrappers.cpp

src/server/req_quejob.c
```diff
@@ -466,7 +471,7 @@ req_quejob(struct batch_request *preq)
 			return;
 		}
 		created_here = JOB_SVFLG_HERE;
-		if (generate_jobid(jidbuf, server_name, i) != 0) {
+		if (generate_objid(jidbuf, server_name, i, ' ') != 0) {
```
agrawalravi90, line 474: pass '\0' as resv_char instead of a valid space character?

agrawalravi90 (changes requested):

src/lib/Libifl/ifl_util.c
```diff
@@ -135,40 +141,64 @@ random_srv_conn(svr_conn_t **svr_conns)
...
 int
-get_job_location_hint(char *jobid)
+get_shard_obj_location_hint(char *obj_id, int obj_type)
 {
 	int nsvrs = get_num_servers();
 	char *ptr = NULL;
+	char *ptr_idx = NULL;
 	int svridx = -1;
 	char *endptr = NULL;
+	int id_len = 0;
```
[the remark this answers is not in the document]
suresh-thelkar: Now that I am implementing your next comment https://github.com/openpbs/openpbs/pull/2255#discussion_r582375355, I think it is better to leave it as "id_len" only. Because we are not decrementing id_len value.

src/lib/Libifl/ifl_util.c
```diff
@@ -135,40 +141,64 @@ random_srv_conn(svr_conn_t **svr_conns)
...
+	}
+
+	/* Minimum length of sequence will be MSVR_JID_NCHARS_SVR + 1 */
+	if (id_len <= MSVR_JID_NCHARS_SVR)
 		return -1;
+	
+	if (obj_type == MGR_OBJ_RESV) {
+		/* Restore id_len to original value */
+		id_len++;
 	}
```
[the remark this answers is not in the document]
suresh-thelkar: I have added all of them into one condition as mentioned. Please check my latest commit.

src/lib/Libifl/ifl_util.c
```diff
@@ -135,40 +141,64 @@ random_srv_conn(svr_conn_t **svr_conns)
...
+ * @param[in] obj_id - job or resv object id
+ * @param[in] obj_type - job or resv object
  *
  * @return int
  * @retval >= 0: start_ind - index where the request should be fired first
  * @retval < 0: could not find appropriate index
  */
 int
-get_job_location_hint(char *jobid)
+get_shard_obj_location_hint(char *obj_id, int obj_type)
```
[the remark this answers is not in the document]
suresh-thelkar: I renamed the function to get_obj_location_hint().

src/lib/Libifl/int_status.c
```diff
@@ -398,28 +398,33 @@ aggr_resc_ct(struct batch_status *st1, struct batch_status *st2)
...
 static void
-move_append_bs(struct batch_status *a, struct batch_status *b, struct batch_status *prev_b, struct batch_status *head_a, struct batch_status **head_b)
+move_append_bs(struct batch_status *b, struct batch_status *prev_b, struct batch_status **head_a, struct batch_status **head_b)
 {
-	if (!a) {
-		for (a = head_a; a->next; a = a->next)
-			;
+	struct batch_status *a = NULL;
+
+	for (a = *head_a; a->next; a = a->next)
```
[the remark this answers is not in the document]
suresh-thelkar: Yes, there is no need to traverse the list 'a' every time inside the function, instead we can pass "tail of a". Added the code changes for it.  We actually don't need a static pointer, the function in hand, gets called when we reach the end of the first list. In other words we can easily get the tail of list 'a'. Please look into my latest commit.

src/lib/Libifl/int_status.c
```diff
@@ -398,28 +398,33 @@ aggr_resc_ct(struct batch_status *st1, struct batch_status *st2)
...
-		b->next = NULL;
-	}
+	else
+		*head_a = b;
+
+
+	if (prev_b)
+		prev_b->next = b->next;
+	else
+		*head_b = b->next;
```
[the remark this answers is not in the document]
suresh-thelkar: This function is actually called inside a loop. The intent of calling this function is to append 'b' to the end of 'a' and is called when there is no aggregation required. We can actually hide the append operation inside a function. I believe this is the original intent of having this in a function. 

@nithinj, can you please let us know your views on this as you are aware of this code.

src/lib/Libifl/int_status.c
```diff
@@ -448,7 +452,9 @@ aggregate_queue(struct batch_status *sv1, struct batch_status **sv2)
 		}
 
 		if (!a)
-			move_append_bs(a, b, prev_b, sv1, sv2);
+			move_append_bs(b, prev_b, &sv1, sv2);
+		else
+			prev_b = b;
```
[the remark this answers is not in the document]
suresh-thelkar: Added my response in the above comment.

src/lib/Libifl/int_status.c
```diff
@@ -572,6 +579,13 @@ PBSD_status_aggregate(int c, int cmd, char *id, void *attrib, char *extend, int
 					cur = last;
 				}
 			}
+		} else if (!single_itr && (pbs_errno == PBSE_UNKQUE || pbs_errno == PBSE_UNKRESVID)) {
```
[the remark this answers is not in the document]
suresh-thelkar: Even in the said case, we are not masking PBSE_UNKQUE if it is non-reservation queue. This is because we are not clearing pbs_errno for all iterations.

src/lib/Libifl/pbsD_submit_resv.c
suresh-thelkar: I have added a detailed comment in the code. Please check.
In both the above cases the value is a jobid.

src/lib/Libifl/pbsD_submit_resv.c
suresh-thelkar: I have moved the above code into the function get_job_svr_inst_fd(). It is actually getting called only once at the moment. But since it can be used by any client, I have moved this to ifl_utils. c

src/scheduler/resv_info.cpp
suresh-thelkar: I have actually looked at one of the functions send_attr_updates() inside sched_ifl_wrappers.cpp. This is still getting called as follows.
```
rc = send_attr_updates(get_svr_inst_fd(pbs_sd, resresv->svr_inst_id), resresv->name, pattr);
rc = send_attr_updates(get_svr_inst_fd(pbs_sd, job->svr_inst_id), job->name, job->job->attr_updates);

```
And also I see there are other calls like pbs_sigjob(), pbs_geterrmsg(), are also getting called in the similar fashion i.e. using get_svr_inst_fd().

Are you suggesting wrapping all of these functions like sig_job(), geterrmsg(), send_attr_updates() etc and internally call get_svr_inst_fd() ? Please clarify.

suresh-thelkar:
@nithinj, Please let me know if you still have any further comments. 
@agrawalravi90, I have addressed all your comments. Please look at the latest commit.

src/scheduler/resv_info.cpp
agrawalravi90: Please look at the latest code, I recently refactored it to make things simpler for callers and also added a wrapper for pbs_sigjob. I missed out pbs_geterrmsg(), if you want you can take it up, or leave it since it's not really in the scope of your PR. But please do add a wrapper for pbs_confirmresv() since you are touching it.

src/lib/Libifl/int_status.c
```diff
@@ -518,12 +518,11 @@ PBSD_status_aggregate(int c, int cmd, char *id, void *attrib, char *extend, int
...
-		if ((start = get_shard_obj_location_hint(id, parent_object)) == -1)
-			start = 0;
-		else
-			single_itr = 1;
-	}
+	if (c == svr_conns[0]->sd)
+		single_itr = 1;
+
+	if ((start = get_obj_location_hint(id, parent_object)) == -1)
+		start = 0;
```
agrawalravi90, line 525: This changes logic: earlier, we were setting single_iter for the case where we found the hint to where the object is located. Now, you are setting it when it's a single server setup. Why the change?

src/lib/Libifl/pbsD_submit_resv.c
```diff
@@ -77,50 +77,16 @@ __pbs_submit_resv(int c, struct attropl *attrib, char *extend)
...
-						break;
-					}
-				}
-			}    
-			c = get_svr_inst_fd(c, svr_inst_id);
-			free(svr_inst_id);
-			pbs_statfree(ss);
-		} else { 
+		if (job_id_resv[0] != '\0') /* Job specific reservation */
+			c = get_job_svr_inst_id(c, job_id_resv);
```
agrawalravi90, line 88: I think we should rename this function to "stat_job_svr_inst_id" to make it clear that this will perform a stat IFL call to get the instance id

agrawalravi90 (approved):

src/lib/Libifl/int_status.c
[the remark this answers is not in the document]
agrawalravi90: ok, it's more involved than id previously thought, im ok with it being in a separate function.

src/lib/Libifl/ifl_util.c
```diff
@@ -164,31 +164,18 @@ get_shard_obj_location_hint(char *obj_id, int obj_type)
 	char *endptr = NULL;
 	int id_len = 0;
 
-	if (obj_id == NULL || !msvr_mode())
+	if (obj_id == NULL || !msvr_mode() || (obj_type != MGR_OBJ_JOB && obj_type == MGR_OBJ_RESV))
```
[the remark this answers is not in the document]
suresh-thelkar: I actually observed that in my testing and made changes and forgot to push this change. Thanks for catching. I will add it to my next commit.

src/lib/Libifl/int_status.c
suresh-thelkar: There are a couple of reasons to change this. 

1. get_obj_location_hint() now returns -1 even if it is not a job/resv object. So if we get -1 in the caller of this function this may or may not indicate that it is a single server which needs single iteration 
2. Since single_itr is to find out if it is a single server or not  we can derive this early i.e. without the help of the function get_obj_location_hint().

src/lib/Libifl/pbsD_submit_resv.c
suresh-thelkar: By keeping the function name as get_job_svr_inst_id(), we are trying to hide the implementation details i.e. the user does not need to know that we are doing a job-stat to get the server instance id. Tomorrow we might choose some other logic to get server instance id. I feel this is a good abstraction/encapsulation. 

Just worried that if we change the name to "stat_job_svr_inst_id" the name of the function itself says that we are doing a job stat. So just wanted to leave this as is for now and if required we can always change the name whenever needed.

suresh-thelkar:
@agrawalravi90, thanks for all the review and sign off. I am also going to add a commit with a single change that we discussed above. Please also take a look at that.

nithinj (approved):
LGTM!

suresh-thelkar:
> LGTM!

Thanks for your review and sign off.

suresh-thelkar:
@agrawalravi90 / @nithinj, I have once again tested all the Multi-Server test cases. All of them are passing. Attached are the logs for the same. 
[Reservations_Msvr_after_review.log](https://github.com/openpbs/openpbs/files/6061099/Reservations_Msvr_after_review.log)


The following are full regression results. All the test cases are passing. 
Profile: regression
|--|--|--|--|--|--|--|
|6258|3990|0|0|0|0|3990|

suresh-thelkar:
@agrawalravi90 / @nithinj, I have also added a wrapper function for pbs_confirmresv(). Please check latest commit.

suresh-thelkar:
@agrawalravi90, I have resolved the conflicts and rebased my branch. Also ran valgrind, attached are the logs for the same. No new issues are observed. 
[valgrind (1).log](https://github.com/openpbs/openpbs/files/6068431/valgrind.1.log)

suresh-thelkar:
@agrawalravi90, thanks for taking this PR forward.

PR #2256, opened 2021-02-16
Fix TestQstatFormats on slow machines

shilpakodli:
#### Describe Bug or Feature
One of the test from TestQstatFormats  was failing with "UnboundLocalError: local variable 'attr_vals' referenced before assignment". This is due to attr_vals is the output of qstat -f of Jobid. and on slow machines, this qstat takes little longer to fetch the Job's status and by the time it gives the output, job would have finished running.

#### Describe Your Change
Removed set_sleep_time from all the tests as default sleep time is 100 seconds which is decent to avoid race conditions.

[TestQstatFormats_after_fix.txt](https://github.com/openpbs/openpbs/files/5988226/TestQstatFormats_after_fix.txt)
[TestQstatFormats_before_fix.txt](https://github.com/openpbs/openpbs/files/5988228/TestQstatFormats_before_fix.txt)

### DIFF
--- a/test/tests/functional/pbs_qstat_formats.py
+++ b/test/tests/functional/pbs_qstat_formats.py
@@ -190,7 +190,6 @@ class TestQstatFormats(TestFunctional):
         test qstat outputs job info in dsv format with default delimiter pipe
         """
         j = Job(TEST_USER)
-        j.set_sleep_time(10)
         jid = self.server.submit(j)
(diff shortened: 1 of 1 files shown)


bhroam (approved):
LGTM

bayucan (approved):
Looks good.

bayucan:
Merged #2256 into master.

PR #2264, opened 2021-02-19
Fixing a leak in parse_psi

agrawalravi90:
#### Describe Bug or Feature

Observed a mem leak for pbs_conf.psi_str, which is allocated inside parse_psi():

13 bytes in 1 blocks are definitely lost in loss record 12 of 39
   at 0x4C28C23: malloc (vg_replace_malloc.c:299)
   by 0x418D12: pbs_asprintf_format (misc_utils.c:529)
   by 0x418E74: pbs_asprintf (misc_utils.c:570)
   by 0x422ED2: parse_psi (pbs_loadconf.c:284)
   by 0x42614E: __pbs_loadconf (pbs_loadconf.c:1112)
   by 0x4275BE: pbs_loadconf (ifl_impl.c:325)
   by 0x405C32: main (pbs_comm.c:606)

### DIFF
--- a/src/lib/Libifl/pbs_loadconf.c
+++ b/src/lib/Libifl/pbs_loadconf.c
@@ -1114,7 +1114,7 @@ __pbs_loadconf(int reload)
 
 	if (!reload) {
 		if (parse_psi(psi_value) == -1) {
-			fprintf(stderr, "Couldn't find a valid server instance to connect to\n");
+			fprintf(stderr, "Couldn't parse PBS_SERVER_INSTANCES\n");
 			free(psi_value);
 			goto err;
 		}
--- a/src/lib/Libifl/pbs_loadconf.c
+++ b/src/lib/Libifl/pbs_loadconf.c
@@ -1112,11 +1112,14 @@ __pbs_loadconf(int reload)
 
 	pbs_conf.loaded = 1;
 
-	if (parse_psi(psi_value) == -1) {
-		fprintf(stderr, "Couldn't find a valid server instance to connect to\n");
-		free(psi_value);
-		goto err;
-	}
+	if (!reload) {
+		if (parse_psi(psi_value) == -1) {
+			fprintf(stderr, "Couldn't find a valid server instance to connect to\n");
+			free(psi_value);
+			goto err;
+		}
+	} else
+		fprintf(stderr, "Warning: PBS_SERVER_INSTANCES will not be re-read from pbs.conf, restart to take effect\n");
 
 	if (pbs_client_thread_unlock_conf() != 0)
 		return 0;
--- a/src/lib/Libifl/pbs_loadconf.c
+++ b/src/lib/Libifl/pbs_loadconf.c
@@ -356,7 +356,9 @@ err:
  *	will be void. In that case, access to every pbs_conf.variable has to be
  *	synchronized against the reload of those variables.
  *
- * @param[in] reload		Whether to attempt a reload
+ * @param[in] reload	Whether to attempt a reload
+ * 						Note: The following parameters will not be reloaded:
+ * 						- PBS_SERVER_INSTANCES
  *
  * @return int
  * @retval 1 Success
--- a/src/lib/Libifl/pbs_loadconf.c
+++ b/src/lib/Libifl/pbs_loadconf.c
@@ -1120,8 +1120,7 @@ __pbs_loadconf(int reload)
 			free(psi_value);
 			goto err;
 		}
-	} else if (psi_value && (!pbs_conf.psi_str || strcmp(psi_value, pbs_conf.psi_str)))
-		fprintf(stderr, "Warning: PBS_SERVER_INSTANCES will not be re-read from pbs.conf, restart to take effect\n");
+	}
 
 	if (pbs_client_thread_unlock_conf() != 0)
 		return 0;
--- a/src/lib/Libifl/pbs_loadconf.c
+++ b/src/lib/Libifl/pbs_loadconf.c
@@ -1118,7 +1118,7 @@ __pbs_loadconf(int reload)
 			free(psi_value);
 			goto err;
 		}
-	} else
+	} else if (psi_value && (!pbs_conf.psi_str || strcmp(psi_value, pbs_conf.psi_str)))
 		fprintf(stderr, "Warning: PBS_SERVER_INSTANCES will not be re-read from pbs.conf, restart to take effect\n");
 
 	if (pbs_client_thread_unlock_conf() != 0)
--- a/src/lib/Libifl/pbs_loadconf.c
+++ b/src/lib/Libifl/pbs_loadconf.c
@@ -281,6 +281,9 @@ parse_psi(char *conf_value)
 	free(pbs_conf.psi_str);
 
 	if (conf_value == NULL) {
+		char *dfltsvr = pbs_default();
+		if (dfltsvr == NULL)
+			return -1;
 		pbs_asprintf(&local_conf, "%s:%d", pbs_default(), pbs_conf.batch_service_port);
 		if (local_conf == NULL)
 			return -1;
--- a/valgrind.supp
+++ b/valgrind.supp
@@ -728,12 +728,3 @@
    fun:dis_flush
    ...
 }
-{
-   From PBS - Suppress intentional unfreed memory of psi_str inside pbs_conf
-   Memcheck:Leak
-   fun:malloc
-   ...
-   fun:pbs_asprintf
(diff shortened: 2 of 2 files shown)


src/lib/Libifl/pbs_loadconf.c
```diff
@@ -281,6 +281,9 @@ parse_psi(char *conf_value)
 	free(pbs_conf.psi_str);
 
 	if (conf_value == NULL) {
+		char *dfltsvr = pbs_default();
```
subhasisb, line 284: i don't see how this fixes a leak. Or even valgrind report is correct (valgrind shows spurious leaks all the time). 
More importantly, we need to add code to prevent reloading of the PBS_SERVER_INSTANCES variable. That should only be loaded once at daemon startup

subhasisb (changes requested):

src/lib/Libifl/pbs_loadconf.c
agrawalravi90: I'm not sure either, but it's a good check to have anyways. I've added code to prevent re-read of PBS_SERVER_INSTANCES on a reload, thanks for catching that.

src/lib/Libifl/pbs_loadconf.c
```diff
@@ -1109,11 +1112,14 @@ __pbs_loadconf(int reload)
 
 	pbs_conf.loaded = 1;
 
-	if (parse_psi(psi_value) == -1) {
-		fprintf(stderr, "Couldn't find a valid server instance to connect to\n");
-		free(psi_value);
-		goto err;
-	}
+	if (!reload) {
```
nithinj, line 1115: psi_str was used to get the raw PSI value. If PSI is NULL, just make it NULL and do not have to use a framed value using pbs_default().

Now if you have that, just compare the strings and return when they are the same within parse_psi.

Instead what you have done is to choose not to re-read the values even when reload=TRUE. which is not the case with any other conf parameters and goes against the definition of the function.

nithinj (changes requested):

src/lib/Libifl/pbs_loadconf.c
```diff
@@ -1109,11 +1112,14 @@ __pbs_loadconf(int reload)
...
+		if (parse_psi(psi_value) == -1) {
+			fprintf(stderr, "Couldn't find a valid server instance to connect to\n");
```
subhasisb, line 1117: BTW, this error message does not sound all that nice!

src/lib/Libifl/pbs_loadconf.c
```diff
@@ -1109,11 +1112,14 @@ __pbs_loadconf(int reload)
...
+			free(psi_value);
+			goto err;
+		}
+	} else
+		fprintf(stderr, "Warning: PBS_SERVER_INSTANCES will not be re-read from pbs.conf, restart to take effect\n");
```
nithinj, line 1122: I feel this is something to be handled at the daemon level. Some may choose to deal with it differently. Leave this function common for all daemons and let the daemons handle it the way it wants.

src/lib/Libifl/pbs_loadconf.c
subhasisb: What is the difference?
pbs_loadconf() does error check, and we can enforce a type of check there. In this case the check is that, we will not reload the value of pbs_conf.psi. I do not see any issue with that. As a precedence, look at pbs_conf.pbs_leaf_routers which is a value that is made-up by loadconf (not exactly what was loaded).

subhasisb (changes requested):

nithinj:
@subhasisb ,
I do not see it as an error check. There is nothing that goes wrong with pbs_loadconf, and hence it is not an error in its perspective. If a daemon chooses to reload these values upon HUP and decides to act upon it, why should that be disallowed? We do not have similar checks for similar parameters like PBS_SERVER.

subhasisb:
u are ignore how many values returned from pbs_loadconf() is not what is in the pbs.conf file, but is synthetic. So, we can always return a synthetic value (which is the old value) (and log that we are intentionally not returning the updated value). So, i think this is perfectly fine!

src/lib/Libifl/pbs_loadconf.c
agrawalravi90: pbs_loadconf isn't just meant to parse pbs.conf, it's meant to do it smartly. We create pbs_conf.psi even when PBS_SERVER_INSTANCES is NULL so that the code which needs it doesn't need special case logic for when pbs_conf.psi is NULL, there can be just a single code flow.

About reload: TBH, 'reload' sounds like it was meant to 'reload' a select few pbs.conf parameters, but we didn't implement it that way. That's why some daemons don't even seem to call pbs_loadconf on a HUP (e.g - server's HUP handler just closes and reopens the logs, same with sched). So, even if this is the first parameter that's not reloaded on 'reload', I think it's not wrong to do so. We should reload only those params which make sense to be reloaded, others need a restart.

src/lib/Libifl/pbs_loadconf.c
agrawalravi90: but it is common, no daemon wants this param to be reloaded. In the future we might enhance them to be able to handle real time changes to PBS_SERVER_INSTANCES, but in the future PBS_SERVER_INSTANCES might not even exist, so let's not plan for that now.

subhasisb (approved):

src/lib/Libifl/pbs_loadconf.c
nithinj: I was referring to psi_str, not pbs_conf.psi. psi_str is used to point to the PSI data and it does not need a framed data. But that does not require to have a special case logic. Now using that, you can just exit if you have the same old data without parsing any further.

You mention how specific daemons behave with the re-read data. Some may act upon it and some may not. That is exactly what I am saying. There is a division of responsibility. pbs_loadconf just reads and makes the data available for the daemons to act upon.

If the daemons do not reload the conf file upon HUP, then isn't it evident that the daemons are responsible to decide that and what is the benefit of adding such a check here?

Now if you want to head in the other direction where only selected parameters are re-read, what will be those parameters? Does it vary from daemon to daemon? So why not leave pbs_loadconf as today which will re-read all parameters upon invocation and leave the rest for the caller.

src/lib/Libifl/pbs_loadconf.c
nithinj: But how do you decide that? pbs_loadconf is an IFL.

Also, Are we throwing this warning on every reload irrespective of whether conf has PSI set or not? And what if PSI hasn't changed?

agrawalravi90: "Also, Are we throwing this warning on every reload irrespective of whether conf has PSI set or not? And what if PSI hasn't changed?"

I can add checks for that.

agrawalravi90:
@nithinj I'm not entirely sure I understand your concern. Is it that pbs_loadconf not loading PBS_SERVER_INSTANCES on a 'reload' is going against the design of pbs_loadconf ?

nithinj:
@agrawalravi90 ,
Yes. That's my primary concern.
Apart from that, I failed to see any value addition we get by blocking parse_psi() on reload.

PR also lacks proper checks in its current form: https://github.com/openpbs/openpbs/pull/2264#discussion_r582234524

agrawalravi90:
> Yes. That's my primary concern.

@nithinj ok, can you please point me to the documented behavior of pbs_loadconf? from what i can tell, we don't really document it.

agrawalravi90:
> Apart from that, I failed to see any value addition we get by blocking parse_psi() on reload.

well, right now the daemons cannot handle PBS_SERVER_INSTANCES being altered while they are up and running, so the value is in preventing them from behaving/failing unexpectedly.

nithinj:
> > Yes. That's my primary concern.
> 
> @nithinj ok, can you please point me to the documented behavior of pbs_loadconf? from what i can tell, we don't really document it.

Please add a note in the Doxygen of the function if you're proceeding in this direction so that the API user can know what he is doing.

agrawalravi90:
@nithinj thanks, I've added info in doxygen, please let me know if it looks okay now

nithinj (approved):

subhasisb (changes requested):
I am not sure this warning message helps. So, we should silently discard the updated value. This is what happens with leaf_routers, comm_routers and such variables. pbs_loadconf() returns a synthetic value back and that is okay.

subhasisb (approved):

subhasisb (approved):

PR #2271, opened 2021-02-26
Implement swap and no-mem/vmem-requested management

alexis-cousein:
#### Describe Bug or Feature
See https://openpbs.atlassian.net/wiki/spaces/PD/pages/2576613377/Improve+memory+swap+management+in+the+cgroup+hook

#### Describe Your Change
-Added queuejob/modifyjob hook to compute "cgswap" from mem and vmem requests
-If "manage_cgswap" is enabled, then resources_available.cgswap is set to the difference between vmem and mem
-Introduce enforce_default and exclhost_ignore_default config parameters in the memory sections (memory/memsw/hugetlb). Since they are boolean, they can be set differently for different cluster hosts.
--enforce_default set to true assigns a default value for mem/swap/hpmem if the job does not explicitly request the resource; enforce_default set to false lets the job use all available resources of that type present on the host and not excluded from use by the job cgroups (cfr. reserved_amount/reserved_percent).
--exclhost_ignore_default enabled will force enforce_default==false behaviour for exclhost placement jobs.
-configure_job and _get_assigned_cgroup_resources will now get the totals available from the parent cgroup set up by exechost_startup, to avoid small changes in reported values due to kernel drivers changing the OS reported totals. exechost_startup will compute the values as before
-memory resources assigned by existing jobs and tallied by the cgroup hook will not include values that have been raised through enforce_default==false or exclhost_ignore_default==true. Since the hook always 'hides' at least one MB from the server/scheduler (and rounds down), the only way for a job to have its cgroup limits identical to the "jobid" parent directory ones is through these mechanisms.
-memsw "default" is now accurately reflecting default extra swap to be added to the memory limit, consistent with the current behaviour of reserved_amount and reserved_percentage. This avoids sites being able to specify nonsensical configurations as well.
-Added default configuration file entries. Changed the default for mem_fences to false, since the combination of mem_fences enabled and vnode_per_numa_node disabled risks erecting fences that may kill jobs. Changed the default reserve_amount for mem to something more appropriate for modern systems with tens of GB (i.e. 1GB) -- the old default is often left as-is by sites and is invariably not enough to protect OS daemons. It's still on the low side but at least sane.

#### Link to Design Doc
See https://openpbs.atlassian.net/wiki/spaces/PD/pages/2576613377/Improve+memory+swap+management+in+the+cgroup+hook

@bayucan ran the TH TestCgroupsHook test on  your cgroup_cgswap branch. (http://th.prog.altair.com/results/6401/all) It's all passing now:

Id: 6401
Description: Tests from given sources on platforms CENTOS7, SUSE15
Platforms: CENTOS7,SUSE15
Profile: Cgroups

Test Summary: total: 52, fail: 0, error: 0, timedout: 0, skip: 13, pass: 39

### DIFF
--- a/src/hooks/cgroups/pbs_cgroups.PY
+++ b/src/hooks/cgroups/pbs_cgroups.PY
@@ -2444,8 +2444,9 @@ class NodeUtils(object):
                 pbs.size(convert_size(val, 'mb'))
             if (self.cfg['cgroup']['memsw']['enabled']
                     and self.cfg['cgroup']['memsw']['manage_cgswap']):
-                host_resc_avail['cgswap'] = \
-                    host_resc_avail['vmem']-host_resc_avail['mem']
+                host_resc_avail['cgswap'] = max(0,
+                                                host_resc_avail['vmem']
+                                                - host_resc_avail['mem'])
         else:
             # set the value on the host to 0
             host_resc_avail['mem'] = pbs.size('0')
@@ -2585,8 +2586,9 @@ class NodeUtils(object):
                 vnode_resc_avail = vnode_list[vnode_key].resources_available
                 if (self.cfg['cgroup']['memsw']['enabled']
                         and self.cfg['cgroup']['memsw']['manage_cgswap']):
-                    vnode_resc_avail['cgswap'] = (vnode_resc_avail['vmem']
-                                                  - vnode_resc_avail['mem'])
+                    vnode_resc_avail['cgswap'] = max(0,
+                                                     vnode_resc_avail['vmem']
+                                                     - vnode_resc_avail['mem'])
                 pbs.logmsg(pbs.EVENT_DEBUG4, '%s: %s vnode_resc_avail: %s' %
                            (caller_name(), vnode_key, vnode_resc_avail))
         pbs.logmsg(pbs.EVENT_DEBUG4, '%s: host_resc_avail: %s' %
--- a/src/hooks/cgroups/pbs_cgroups.PY
+++ b/src/hooks/cgroups/pbs_cgroups.PY
@@ -5890,6 +5890,9 @@ def missing_str(memspecs):
             mem_size = pbs.size(memspecs['mem'])
             if mem_size > vmem_size:
                 event.reject("invalid specification: vmem>mem")
+            elif mem_size == vmem_size:
+                # Avoid creating a "0mb" pbs.size
+                return("cgswap=0b")
             else:
                 cgswap_size = vmem_size - mem_size
                 return ("cgswap="
@@ -5913,13 +5916,18 @@ def missing_str(memspecs):
     if 'vmem' in memspecs and 'cgswap' in memspecs:
         try:
             vmem_size = pbs.size(memspecs['vmem'])
-            cgswap_size = pbs.size(memspecs['cgswap'])
-            if cgswap_size >= vmem_size:
-                event.reject("invalid specification: mem<=0")
+            # Woraround for bug: no arithmetic on zero pbs.size
+            # since user may have specified e.g. "0mb" cgswap
+            if size_as_int(memspecs['cgswap']) == 0:
+                mem_size = vmem_size
+                return ("mem=" + str(mem_size))
             else:
-                mem_size = vmem_size - cgswap_size
-                return ("mem="
-                        + str(mem_size))
+                cgswap_size = pbs.size(memspecs['cgswap'])
+                if cgswap_size >= vmem_size:
+                    event.reject("invalid specification: mem<=0")
+                else:
+                    mem_size = vmem_size - cgswap_size
+                    return ("mem=" + str(mem_size))
         except Exception:
             pbs.logmsg(pbs.EVENT_DEBUG4,
                        str(traceback.format_exc().strip().splitlines()))
--- a/src/hooks/cgroups/pbs_cgroups.PY
+++ b/src/hooks/cgroups/pbs_cgroups.PY
@@ -1061,7 +1061,7 @@ class HookUtils(object):
                                'deleting cpuset cgroup')
 
         # Configure the new cgroup
-        cgroup.configure_job(event.job.id, jobutil.assigned_resources,
+        cgroup.configure_job(event.job, jobutil.assigned_resources,
                              node, cgroup, event.type)
 
         # Write out the assigned resources
@@ -2442,11 +2442,18 @@ class NodeUtils(object):
             val -= val % (1024 * 1024)  # round down to MB
             host_resc_avail['hpmem'] = \
                 pbs.size(convert_size(val, 'mb'))
+            if (self.cfg['cgroup']['memsw']['enabled']
+                    and self.cfg['cgroup']['memsw']['manage_cgswap']):
+                host_resc_avail['cgswap'] = \
+                    host_resc_avail['vmem']-host_resc_avail['mem']
         else:
             # set the value on the host to 0
             host_resc_avail['mem'] = pbs.size('0')
             host_resc_avail['vmem'] = pbs.size('0')
             host_resc_avail['hpmem'] = pbs.size('0')
+            if (self.cfg['cgroup']['memsw']['enabled']
+                    and self.cfg['cgroup']['memsw']['manage_cgswap']):
+                host_resc_avail['cgswap'] = pbs.size('0')
             # Some kernel drivers "reserve" memory on /proc/meminfo
             # and this is not reflected in the node meminfo; if necessary,
             # adjust the values published as resources_available.mem
@@ -2576,6 +2583,10 @@ class NodeUtils(object):
             for nnid in self.numa_nodes:
                 vnode_key = vnode_name + '[%d]' % nnid
                 vnode_resc_avail = vnode_list[vnode_key].resources_available
+                if (self.cfg['cgroup']['memsw']['enabled']
+                        and self.cfg['cgroup']['memsw']['manage_cgswap']):
+                    vnode_resc_avail['cgswap'] = (vnode_resc_avail['vmem']
+                                                  - vnode_resc_avail['mem'])
                 pbs.logmsg(pbs.EVENT_DEBUG4, '%s: %s vnode_resc_avail: %s' %
                            (caller_name(), vnode_key, vnode_resc_avail))
         pbs.logmsg(pbs.EVENT_DEBUG4, '%s: host_resc_avail: %s' %
@@ -2927,7 +2938,7 @@ class CgroupUtils(object):
             if jobid:
                 # Caller wants parent directory of job
                 return os.path.join(subdir, jobid, '')
-            # Caller wants parent directory of subsystem
+            # Caller wants job parent directory for subsystem
             return os.path.join(subdir, '')
         # Caller wants full path to file
         if cgfile == 'tasks':
@@ -3193,20 +3204,14 @@ class CgroupUtils(object):
         defaults['cgroup']['devices']['exclude_hosts'] = []
         defaults['cgroup']['devices']['exclude_vntypes'] = []
         defaults['cgroup']['devices']['allow'] = []
-        defaults['cgroup']['hugetlb'] = {}
-        defaults['cgroup']['hugetlb']['enabled'] = False
-        defaults['cgroup']['hugetlb']['exclude_hosts'] = []
-        defaults['cgroup']['hugetlb']['exclude_vntypes'] = []
-        defaults['cgroup']['hugetlb']['default'] = '0MB'
-        defaults['cgroup']['hugetlb']['reserve_percent'] = 0
-        defaults['cgroup']['hugetlb']['reserve_amount'] = '0MB'
-        defaults['cgroup']['hugetlb']['vnode_hidden_mb'] = 1
         defaults['cgroup']['memory'] = {}
         defaults['cgroup']['memory']['enabled'] = False
         defaults['cgroup']['memory']['exclude_hosts'] = []
         defaults['cgroup']['memory']['exclude_vntypes'] = []
         defaults['cgroup']['memory']['soft_limit'] = False
         defaults['cgroup']['memory']['default'] = '0MB'
+        defaults['cgroup']['memory']['enforce_default'] = True
+        defaults['cgroup']['memory']['exclhost_ignore_default'] = False
         defaults['cgroup']['memory']['reserve_percent'] = 0
         defaults['cgroup']['memory']['reserve_amount'] = '0MB'
         defaults['cgroup']['memory']['vnode_hidden_mb'] = 1
@@ -3216,9 +3221,22 @@ class CgroupUtils(object):
         defaults['cgroup']['memsw']['exclude_hosts'] = []
         defaults['cgroup']['memsw']['exclude_vntypes'] = []
         defaults['cgroup']['memsw']['default'] = '0MB'
+        defaults['cgroup']['memsw']['enforce_default'] = True
+        defaults['cgroup']['memsw']['exclhost_ignore_default'] = False
         defaults['cgroup']['memsw']['reserve_percent'] = 0
         defaults['cgroup']['memsw']['reserve_amount'] = '0MB'
-        defaults['cgroup']['memsw']['vnode_hidden_mb'] = 0
+        defaults['cgroup']['memsw']['vnode_hidden_mb'] = 1
+        defaults['cgroup']['memsw']['manage_cgswap'] = False
+        defaults['cgroup']['hugetlb'] = {}
+        defaults['cgroup']['hugetlb']['enabled'] = False
+        defaults['cgroup']['hugetlb']['exclude_hosts'] = []
+        defaults['cgroup']['hugetlb']['exclude_vntypes'] = []
+        defaults['cgroup']['hugetlb']['default'] = '0MB'
+        defaults['cgroup']['hugetlb']['enforce_defaults'] = True
+        defaults['cgroup']['hugetlb']['exclhost_ignore_default'] = False
+        defaults['cgroup']['hugetlb']['reserve_percent'] = 0
+        defaults['cgroup']['hugetlb']['reserve_amount'] = '0MB'
+        defaults['cgroup']['hugetlb']['vnode_hidden_mb'] = 1
         defaults['cgroup']['net_cls'] = {}
         defaults['cgroup']['net_cls']['enabled'] = False
         defaults['cgroup']['net_prio'] = {}
@@ -3467,6 +3485,29 @@ class CgroupUtils(object):
         """
         pbs.logmsg(pbs.EVENT_DEBUG4, '%s: Method called' % caller_name())
         assigned = {}
+
+        # if jobs have a limit *exactly* equal to the total available limit
+        # then they have no explicit memory req and the limit was set to the
+        # available total -- get totals to compare
+        mem_avail = None
+        filename = self._cgroup_path('memory', 'limit_in_bytes')
+        if filename and os.path.isfile(filename):
+            with open(filename) as desc:
+                mem_avail = int(desc.readline())
+        vmem_avail = None
+        filename = self._cgroup_path('memsw', 'limit_in_bytes')
+        if filename and os.path.isfile(filename):
+            with open(filename) as desc:
+                vmem_avail = int(desc.readline())
+        hpmem_avail = None
+        filename = self._cgroup_path('hugetlb', 'limit_in_bytes')
+        if filename and os.path.isfile(filename):
+            with open(filename) as desc:
+                hpmem_avail = int(desc.readline())
+        pbs.logmsg(pbs.EVENT_DEBUG4, "get_assigned_cgroup_resources:"
+                   " total host avail mem=%s vmem=%s hpmem=%s"
+                   % (mem_avail, vmem_avail, hpmem_avail))
+
         for key in self.paths:
             if key in ('blkio', 'cpu', 'cpuacct', 'freezer', 'systemd'):
                 continue
@@ -3499,18 +3540,24 @@ class CgroupUtils(object):
                                                 jobid)) as desc:
                         assigned[jobid][key]['limit_in_bytes'] = \
                             int(desc.readline())
+                    if assigned[jobid][key]['limit_in_bytes'] == mem_avail:
+                        assigned[jobid][key]['limit_in_bytes'] = 0
                     with open(self._cgroup_path(key, 'soft_limit_in_bytes',
                                                 jobid)) as desc:
                         assigned[jobid][key]['soft_limit_in_bytes'] = \
                             int(desc.readline())
+                    if (assigned[jobid][key]['soft_limit_in_bytes']
+                            == mem_avail):
+                        assigned[jobid][key]['soft_limit_in_bytes'] = 0
                 elif key == 'memsw':
                     filename = self._cgroup_path('memsw', 'limit_in_bytes',
                                                  jobid)
-                    if os.path.isfile(filename):
+                    if filename and os.path.isfile(filename):
                         with open(filename) as desc:
-                            assigned[jobid]['memsw'] = {}
                             assigned[jobid]['memsw']['limit_in_bytes'] = \
                                 int(desc.readline())
+                    if assigned[jobid][key]['limit_in_bytes'] == vmem_avail:
+                        assigned[jobid][key]['limit_in_bytes'] = 0
                     else:
                         pbs.logmsg(pbs.EVENT_DEBUG4, '%s: No such file: %s' %
                                    (caller_name(), filename))
@@ -3519,6 +3566,8 @@ class CgroupUtils(object):
                                                 jobid)) as desc:
                         assigned[jobid][key]['limit_in_bytes'] = \
                             int(desc.readline())
+                    if assigned[jobid][key]['limit_in_bytes'] == hpmem_avail:
+                        assigned[jobid][key]['limit_in_bytes'] == 0
                 elif key == 'devices':
                     path = self._cgroup_path(key, 'list', jobid)
                     pbs.logmsg(pbs.EVENT_DEBUG4, '%s: Devices path is %s' %
@@ -4726,17 +4775,40 @@ class CgroupUtils(object):
             finally:
                 os.umask(old_umask)
 
-    def configure_job(self, jobid, hostresc, node, cgroup, event_type):
+    def configure_job(self, job, hostresc, node, cgroup, event_type):
         """
         Determine the cgroup limits and configure the cgroups
         """
+        jobid = job.id
+
+        # Get available totals for mem, vmem and hpmem from parent cgroup
+        # Calling node.get_memory_on_node etc. may give slightly different
+        # results if kernel drivers have freshly reserved memory since
+        # exechost_startup was called
+        mem_avail = None
+        filename = self._cgroup_path('memory', 'limit_in_bytes')
+        if filename and os.path.isfile(filename):
+            with open(filename) as desc:
+                mem_avail = int(desc.readline())
+        vmem_avail = None
+        filename = self._cgroup_path('memsw', 'limit_in_bytes')
+        if filename and os.path.isfile(filename):
+            with open(filename) as desc:
+                vmem_avail = int(desc.readline())
+        hpmem_avail = None
+        filename = self._cgroup_path('hugetlb', 'limit_in_bytes')
+        if filename and os.path.isfile(filename):
+            with open(filename) as desc:
+                hpmem_avail = int(desc.readline())
+        pbs.logmsg(pbs.EVENT_DEBUG4, "configure_job:"
+                   " total host avail mem=%s vmem=%s hpmem=%s"
+                   % (mem_avail, vmem_avail, hpmem_avail))
+
         pbs.logmsg(pbs.EVENT_DEBUG4, '%s: Method called' % caller_name())
         mem_enabled = 'memory' in self.subsystems
         vmem_enabled = 'memsw' in self.subsystems
         if mem_enabled or vmem_enabled:
             # Initialize mem variables
-            mem_avail = node.get_memory_on_node()
-            pbs.logmsg(pbs.EVENT_DEBUG4, 'mem_avail %s' % mem_avail)
             mem_requested = None
             if 'mem' in hostresc:
                 mem_requested = convert_size(hostresc['mem'], 'kb')
@@ -4744,8 +4816,6 @@ class CgroupUtils(object):
             if mem_enabled:
                 mem_default = self.default('memory')
             # Initialize vmem variables
-            vmem_avail = node.get_vmem_on_node()
-            pbs.logmsg(pbs.EVENT_DEBUG4, 'vmem_avail %s' % vmem_avail)
             vmem_requested = None
             if 'vmem' in hostresc:
                 vmem_requested = convert_size(hostresc['vmem'], 'kb')
@@ -4777,7 +4847,14 @@ class CgroupUtils(object):
                 mem_limit = mem_requested
             else:
                 # mem was not requested
-                if mem_default is None:
+                if (mem_default is None
+                        or not self.cfg['cgroup']['memory']
+                                       ['enforce_default']
+                        or (self.cfg['cgroup']['memory']
+                                    ['exclhost_ignore_default']
+                            and "place" in job.Resource_List
+                            and 'exclhost'
+                                in repr(job.Resource_List['place']))):
                     mem_limit = mem_avail
                 else:
                     mem_limit = mem_default
@@ -4791,10 +4868,20 @@ class CgroupUtils(object):
                 vmem_limit = vmem_requested
             else:
                 # vmem was not requested
-                if vmem_default is None:
+                if (vmem_default is None
+                        or not self.cfg['cgroup']['memsw']
+                                       ['enforce_default']
+                        or (self.cfg['cgroup']['memsw']
+                                    ['exclhost_ignore_default']
+                            and "place" in job.Resource_List
+                            and 'exclhost'
+                                in repr(job.Resource_List['place']))):
+
                     vmem_limit = vmem_avail
                 else:
-                    vmem_limit = vmem_default
+                    vmem_limit = str(min(size_as_int(vmem_avail),
+                                         size_as_int(mem_limit)
+                                         + size_as_int(vmem_default)))
             # Ensure vmem is at least as large as mem
             if size_as_int(vmem_limit) < size_as_int(mem_limit):
                 vmem_limit = mem_limit
@@ -4855,10 +4942,15 @@ class CgroupUtils(object):
         # Initialize hpmem variables
         hpmem_enabled = 'hugetlb' in self.subsystems
         if hpmem_enabled:
-            hpmem_avail = node.get_hpmem_on_node(ignore_reserved=False)
             hpmem_limit = None
             hpmem_default = self.default('hugetlb')
-            if hpmem_default is None:
+            if (hpmem_default is None
+                    or not self.cfg['cgroup']['hugetlb']
+                                   ['enforce_default']
+                    or (self.cfg['cgroup']['hugetlb']
+                                ['exclhost_ignore_default']
+                        and "place" in job.Resource_List
+                        and 'exclhost' in repr(job.Resource_List['place']))):
                 hpmem_default = hpmem_avail
             if 'hpmem' in hostresc:
                 hpmem_limit = convert_size(hostresc['hpmem'], 'kb')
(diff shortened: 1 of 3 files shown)


alexis-cousein:
BTW, I changed the "default" for memsw to be 0MB in the shipped CF file -- most sites either do not want to allow jobs to swap or in many cases have no swap on compute nodes. That is probably the most sane and useful starting point, rather than an arbitrary 256MB. in "memory" of course a 0 default would not allow any job to start, but that is not the case with "memsw" (if indeed "default", reserve_amount and reserve_percent refer to swap).

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -3499,18 +3540,24 @@ class CgroupUtils(object):
                                                 jobid)) as desc:
                         assigned[jobid][key]['limit_in_bytes'] = \
                             int(desc.readline())
+                    if assigned[jobid][key]['limit_in_bytes'] == mem_avail:
+                        assigned[jobid][key]['limit_in_bytes'] = 0
```
bayucan, line 3544: Can you give the context of why this is being set to 0 if current value is equal to mem_avail?

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -2442,11 +2442,18 @@ class NodeUtils(object):
             val -= val % (1024 * 1024)  # round down to MB
             host_resc_avail['hpmem'] = \
                 pbs.size(convert_size(val, 'mb'))
+            if (self.cfg['cgroup']['memsw']['enabled']
+                    and self.cfg['cgroup']['memsw']['manage_cgswap']):
+                host_resc_avail['cgswap'] = \
+                    host_resc_avail['vmem']-host_resc_avail['mem']
```
bayucan, line 2448: Should we guard against what if host_resc_avail['vmem'] - host_resc_avail['mem'] yield a negative value?

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -2576,6 +2583,10 @@ class NodeUtils(object):
             for nnid in self.numa_nodes:
                 vnode_key = vnode_name + '[%d]' % nnid
                 vnode_resc_avail = vnode_list[vnode_key].resources_available
+                if (self.cfg['cgroup']['memsw']['enabled']
+                        and self.cfg['cgroup']['memsw']['manage_cgswap']):
+                    vnode_resc_avail['cgswap'] = (vnode_resc_avail['vmem']
+                                                  - vnode_resc_avail['mem'])
```
bayucan, line 2589: Same as before, what if the vnode_resc_avail['vmem'] - node_resc_avail['mem'] results in negative value?

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -3519,6 +3566,8 @@ class CgroupUtils(object):
                                                 jobid)) as desc:
                         assigned[jobid][key]['limit_in_bytes'] = \
                             int(desc.readline())
+                    if assigned[jobid][key]['limit_in_bytes'] == hpmem_avail:
+                        assigned[jobid][key]['limit_in_bytes'] == 0
```
bayucan, line 3570: I think you meant '=' rather than '=='

bayucan (changes requested):

src/hooks/cgroups/pbs_cgroups.PY
alexis-cousein: See comment above: if the per-job limit is that for the overarching "jobid" directory, then it was set because the job did not request memory and the limit was raised because default enforcement was disabled. Since the values reported as available to the server and scheduler are _lower_, it's impossible for an explicitly requested memory value to be this high.

These jobs did not request memory, so when determining how much memory was assigned, it is zero. Otherwise you'd have the paradoxical effect that if _one_ individual manages to get a job without an explicit memory request, then _no_ other job that asks even for one byte would be able to land on the node and get accepted by the cgroup hook. Since the _scheduler_ thinks that jobs that do not explicitly ask for memory ask for 0 bytes, that would lead to jobs being scheduled and rejected by MoM -- possibly several times, until the job was held.

src/hooks/cgroups/pbs_cgroups.PY
alexis-cousein: It is (not strictly) positive by construction. The difference is the swap not reserved for other purposes, in this snippet in get_vmem_on_node:

`        if swap <= 0:
            pbs.logmsg(pbs.EVENT_DEBUG4,
                       '%s: No swap space detected' %
                       caller_name())
            swap = 0
        pbs.logmsg(pbs.EVENT_DEBUG4, 'total swap: %d' % swap)
        # Calculate reserved swap
        reserved = 0
        if not ignore_reserved:
            reserve_pct = int(self.cfg['cgroup']['memsw']['reserve_percent'])
            reserved += int(swap * (reserve_pct / 100.0))
            reserve_amount = self.cfg['cgroup']['memsw']['reserve_amount']
            reserved += size_as_int(reserve_amount)
            pbs.logmsg(pbs.EVENT_DEBUG4, 'reserved swap: %d' % reserved)
            if reserved > swap:
                reserved = swap
        # Calculate remaining vmem
        remaining = total + swap - reserved
` 
That last guard will guarantee that swap - reserved is always positive and that vmem remaining reported is at least mem.

But of course, if you'd rather see us add a max(0,...) I'm not opposed to it.

src/hooks/cgroups/pbs_cgroups.PY
alexis-cousein: It never does, by construction, but same comment -- if you think inserting a max(0,...) makes sense, I'm not opposed to it.

src/hooks/cgroups/pbs_cgroups.PY
alexis-cousein: In fact, given this is only done in exechost_startup, I'll add the max. No point in making a reader even think about that possibility...

src/hooks/cgroups/pbs_cgroups.PY
alexis-cousein: Ach -- I see I forgot to push a commit.

alexis-cousein:
As for PTL tests, here would be my proposed test. The first one would work both if there were and were not swap available, but it would do more testing if there is swap found on the node.

Test #1

-define resource cgswap
-load new config file (NUMA off, manage_cgswap on)
-enable queuejob and modifyjob events for cgroup hook
-restart MoM
-killall -HUP pbs_mom
-do a pbsnodes of the server
-add cgswap to the scheduler resources, HUP scheduler
-see if resources_available.cgswap is set
if it is strictly positive:
  -submit one job with mem=256MB and vmem 256MB+2/3rds of cgswap
  -see if cgswap is defined in schedselect
  -wait until it runs
  -confirm cgroup limits are as per request
  -submit second job with identical specs
  -wait 20 seconds -- it must not run
  -kill first job
  -confirm second job now runs
if it is 0:
-submit job with -lselect=1:mem=256MB:vmem=1gb
-See if cgswap is defined in schedselect for job

Test #2

Ditto, but with NUMA enabled
pbsnodes -v server_name[0] instead of pbsnodes server_name

Test #3 (enforce_default for vmem)

-Load config: enforce_dfault enabled for mem, disabled for vmem.
-restart MoM
-killall -HUP MoM
-Submit job with -lselect=1:ncpus=1
-Wait until it runs
Check cgroup limits

Test #4 (do not enforce default for mem, enforce for vmem)

-Idem, different config file (see title)
-If resources_available.vmem - resources_available.mem on node is less than 
 256MB, then memsw limit should be that of parent dir and should be >= 
 resources_available.vmem
-If more than that, then vmem limit should be mem limit +256MB

bayucan:
@alexis-cousein : the updates to the code look good. And the tests that you proposed seemed fine.

alexis-cousein:
Added PTL tests.

During PTL testing I discovered it's possible to trip over a bug in the hook library -- pbs.size("0mb") quantities, both directly inserted (from e.g. the config file) or computed (e.g. by subtracting large mb-round values from each other with a 0 result) can make hooks crash. That led to some extra changes, and to avoid that in one routine the "_limit" and "_requested" quantities are strings (since you don't want these converted to bytes when the limit is what was requested) but the "_default" and "_avail" are now simply number of bytes instead of strings. "0" has to be spacial-cased in a couple of places to avoid arithmetic that on some machines may crash the hook.

The PTL tests are:

TesCgroupsHook:
	test_cgroup_cgswap
	test_cgroup_cgswap_numa
	test_cgroup_enforce_default
	test_cgroup_enforce_default_tf
	test_cgroup_enforce_default_ft
	test_cgroup_enforce_default_exclhost

The first two check whether cgswap is correctly configured on the nodes and whether jobs asking for mem and vmem get a cgswap added to the select specification. The second test always sets vnode_per_numa_node to True to exercise the other code paths involved.

The other four test what happens to a job that specifies no mem and vmem requests. All combinations of enforcement on mem and memsw are tested, but the last (no enforcement on either subsystem) is done by specifying enforcement as enabled but by specifying an exclhost job (which then disables it again by virtue of exclhost_ignore_default).

alexis-cousein:
THe PTL tests seem not to work on hosts where PTL "hostname" is the long name -- I'll make some changes.

bayucan:
@alexis-cousein : I just noticed that in your design, there was no mention of a queuejob/modifyjob component of pbs_cgroups hook. There'll need to be an instruction for manually enabling it  to compute "cgswap" from mem and vmem requests. Currently, the default pbs_cgroups.HK is left alone, without 'queuejob' and 'modifyjob' in hook events. I think it's better to leave it like this, and not add these server hook events, as there might be conflicts with sites that have queujob or modifyjob hook events. Care must be taken as to the order of these hook events. Perhaps an existing queuejob hook could be modifying vmem/mem, which can be fed to the pbs_cgroups queuejob hook. We just probably mention that 'order' must be carefully selected. I think it's reasonable to specify an 'order' attribute value that will make the pbs_cgroups queuejob or modifyjob hook execute last.

alexis-cousein:
> @alexis-cousein : I just noticed that in your design, there was no mention of a queuejob/modifyjob component of pbs_cgroups hook. There'll need to be an instruction for manually enabling it to compute "cgswap" from mem and vmem requests. Currently, the default pbs_cgroups.HK is left alone, without 'queuejob' and 'modifyjob' in hook events. I think it's better to leave it like this, and not add these server hook events, as there might be conflicts with sites that have queujob or modifyjob hook events. Care must be taken as to the order of these hook events. Perhaps an existing queuejob hook could be modifying vmem/mem, which can be fed to the pbs_cgroups queuejob hook. We just probably mention that 'order' must be carefully selected. I think it's reasonable to specify an 'order' attribute value that will make the pbs_cgroups queuejob or modifyjob hook execute last.

The document mentions:
'To make the new feature as transparent as possible, a site may enable a queuejob and modifyjob event on the cgroup hook, which will complete either "old" style -lncpus=X,mem=Y,vmem=Z or new-style -lselect=N:ncpus=X:mem=Y:vmem=Z with the computed "cgswap"'

But you're right the step is missing in the bullet list.

bayucan:
@alexis-cousein : I'm satisfied with the design update. The pbs_cgroups hook is actually has a high 'order' attribute value of 100 to hope that it executes last. The max order value is 1000.

bayucan (approved):
Looks good.

alexis-cousein:
Added some more guards to avoid pbs.size("0mb") injections, since this crashes hooks on some OpenPBS versions.

alexis-cousein:
PTL tests succeed, both NUMA and non-NUMA. Same output as before.

nishiya:
I haven't gone through all the code or comments yet, but I did read the design, and I do want to note a few things in the PR description should be noted in the design:
- exclhost_ignore_default enabled will force enforce_default==false behaviour for exclhost placement jobs
- Changed the default for mem_fences to false
- Changed the default reserve_amount for mem
- changed "default" for memsw to be 0MB in the shipped CF file

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -4334,6 +4476,274 @@ def test_cgroup_default_config(self):
...
+            self.logger.info('job cgswap detected to be correct, roughly %s'
+                             % str(cgswap))
+            j = Job(TEST_USER, attrs=a)
+            j.create_script(self.sleep100_job)
+            jid = self.server.submit(j)
+
+            # Second job should not run - not enough cgswap
+            # offset large enough to give scheduler a chance
+            bs2 = {'job_state': 'Q'}
+            self.server.expect(JOB, bs2, jid, offset=20, max_attempts=1)
```
nishiya, line 4595: Rather than wait for a scheduling cycle.  Set scheduling=t to cause a scheduling cycle to run.
And get rid off the offset.

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -4334,6 +4476,274 @@ def test_cgroup_default_config(self):
...
+            self.skipTest('Test requires memory subystem mounted')
+        if not self.swapctl:
+            self.skipTest('Test requires memsw accounting enabled')
+
+        self.load_config(self.cfg16
+                         % enforce_flags)
+        # Restart mom for changes made by cgroups hook to take effect
+        begin = time.time()
+        # sleep 2s to ensure log matching will not catch older log lines
+        time.sleep(2)
```
nishiya, line 4613: Same comment as earlier, can we get rid of the 2 second sleep?

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -4334,6 +4476,274 @@ def test_cgroup_default_config(self):
...
+        vmem_avail_in_bytes = None
+        try:
+            vmem_avail_in_bytes = int(result['out'][0])
+        except Exception:
+            # None will be seen as a failure, nothing to do
+            pass
+        self.logger.info("total available memsw: %d"
+                         % vmem_avail_in_bytes)
+        self.assertTrue(vmem_avail_in_bytes is not None,
+                        "Unable to read total memsw available")
```
nishiya, line 4668: Can you make this error assertion unique to make it easier to tell which part is failing?  Right now it matches mem_avail_in_bytes.

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -4334,6 +4476,274 @@ def test_cgroup_default_config(self):
...
+
+        # Check results correspond to enforcement flags and job placement
+        swap_avail = vmem_avail_in_bytes - mem_avail_in_bytes
+        if enforce_flags[0] == 'true' and not exclhost:
+            self.assertTrue(mem_limit_in_bytes == 100 * 1024 * 1024,
+                            "Job mem limit is %d expected %d"
+                            % (mem_limit_in_bytes, 100 * 1024 * 1024))
+        else:
+            self.assertTrue(mem_avail_in_bytes == mem_limit_in_bytes,
+                            "job mem limit is not total mem available")
```
nishiya, line 4710: Can this message be more like the one on line 4717 and print the expected and got values?

nishiya (changes requested):
Thank you for adding tests!  Could you please add test descriptions to the tests.
Especially to clarify enforce_flags and which setting is being set to what without having to find the self.cfg16 and find it, i.e. memory.enforce_default=true and memsw.enforce_default=false

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -3216,9 +3248,22 @@ class CgroupUtils(object):
         defaults['cgroup']['memsw']['exclude_hosts'] = []
         defaults['cgroup']['memsw']['exclude_vntypes'] = []
         defaults['cgroup']['memsw']['default'] = '0MB'
+        defaults['cgroup']['memsw']['enforce_default'] = True
+        defaults['cgroup']['memsw']['exclhost_ignore_default'] = False
         defaults['cgroup']['memsw']['reserve_percent'] = 0
         defaults['cgroup']['memsw']['reserve_amount'] = '0MB'
-        defaults['cgroup']['memsw']['vnode_hidden_mb'] = 0
+        defaults['cgroup']['memsw']['vnode_hidden_mb'] = 1
```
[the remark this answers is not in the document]
alexis-cousein: That is, of course, not an interface change, so I don't think it necessarily belongs in a _design_ document. But I'll mention it.

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -4334,6 +4476,274 @@ def test_cgroup_default_config(self):
...
+        # Enable the cgroups hook new events
+        conf = {'enabled': 'True', 'freq': 10, 'event': events}
+        self.server.manager(MGR_CMD_SET, HOOK, conf, self.hook_name)
+
+        self.load_config(self.cfg15
+                         % ('true' if vnode_per_numa_node else 'false'))
+        # Restart mom for changes made by cgroups hook to take effect
+        begin = time.time()
+        # sleep 2s to ensure log matching will not catch older log lines
+        time.sleep(2)
```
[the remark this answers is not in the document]
alexis-cousein: No, and yes.
 
We need a time that is guaranteed to be before the action so that we know from when to search in the logs, but we also need to avoid catching the results of earlier actions; the way we do that is to take time, then not do anything for 2 seconds. 

1 second is not enough for reasons of rounding -- time.time is a float but times in the logs are integers, and very small rounding differences might (rarely) make PTL and the daemons come to a different conslusion.

test/tests/functional/pbs_cgroups_hook.py
alexis-cousein: ?

test/tests/functional/pbs_cgroups_hook.py
alexis-cousein: Sure.

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -5767,6 +5883,168 @@ def set_global_vars():
...
+            pbs.logmsg(pbs.EVENT_DEBUG4,
+                       str(traceback.format_exc().strip().splitlines()))
+            event.reject("Invalid mem or cgswap size requested")
+
+    if 'vmem' in memspecs and 'cgswap' in memspecs:
+        try:
+            vmem_size = pbs.size(memspecs['vmem'])
+            # Woraround for bug: no arithmetic on zero pbs.size
+            # since user may have specified e.g. "0mb" cgswap
+            if size_as_int(memspecs['cgswap']) == 0:
```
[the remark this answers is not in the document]
alexis-cousein: 0 for mem is nonsensical, whether expressed in mb or b -- such a job would never run. Whereas specifying that you want the job to consume no swap is entirely legitimate.

 But you're right, someone could actually set that default in an _attempt_ to ensure jobs that do not specify memory would not run (of course, crashing the hook would also have that effect). I'll add it.

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -4334,6 +4476,274 @@ def test_cgroup_default_config(self):
...
+            a = {'Resource_List.select':
+                 '1:ncpus=1:mem=100mb:vmem=1100mb:vnode=%s'
+                 % vnode_name}
+
+            j = Job(TEST_USER, attrs=a)
+            j.create_script(self.sleep30_job)
+            jid = self.server.submit(j)
+            # give the scheduler time to schedule, but job should not run
+            a = {'job_state': 'Q'}
+            self.server.expect(JOB, a, jid, offset=5)
```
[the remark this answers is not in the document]
alexis-cousein: The scheduling cycle *is* already triggered by the job submission; there is no need to trigger it. The offset is 5 seconds in order to avoid consuming some attempts in the "expect" when we do not expect the job to run (it's been tuned to the behaviour of a well-known test harness). Some test harness virtual machines are fickle beasts that can take ages to do a scheduling cycle because of resource starvation (competition with other VMs on the host).

On my personal host, offset=1 is perfectly sane. It is futile to expect it to make tests faster on a well-known test harness with VMs.

test/tests/functional/pbs_cgroups_hook.py
alexis-cousein: There is only ONE attempt -- the offset is necessary or it won't work.

As said before, we don't need to "trigger a cycle" -- a cycle *is* triggered. The problem is that on a test harness with variable performance and variable times to actually run a cycle to completion, we do not know exactly how long to wait before the scheduler has had a chance to run and MoM has had a chance to accept the job and tell the server. The 20 seconds is safe on a well-known test harness. Less, and you're going to run the chance that you're not seeing the job rujn simply because the test harness is too slow toi actually run the job, even though it (mistakenly) does.

 This works, and if you want to propose an alternative, I'm open to suggestions, but one proven to work on a system where operations that could take less than a second on a "regular" host would take 10 seconds.

alexis-cousein: Note: I am going to increase a sleep in the setup routine from 4 seconds to 6 seconds, BTW. A MoM bug together with the extreme slowness of the tests harness VMs has shown once case where "only" 4 seconds has introduced a false failure because when waiting 4 seconds, we were sending a HUP signal just when MoM was setting up the TPP thread, which triggered a race condition and a hang.

And yes: after 4 whole seconds the MoM wasn't fully restarted, and had not even begun running the exechost_startup hooks.

test/tests/functional/pbs_cgroups_hook.py
alexis-cousein: I would suggest not to do so, but I will rewrite it as before.

alexis-cousein:
[- exclhost_ignore_default enabled will force enforce_default==false behaviour for exclhost placement jobs]
That is not the way I would word it. enforce_default is a flag, not a behaviour, and that flag operates on all jobs. Both flags modulate the behaviour of the hook, one for all jobs, the other for host exclusive jobs only. But I'll reword the design document.

[- Changed the default for mem_fences to false]
That does warrant being discussed in the design document

[- Changed the default reserve_amount for mem]

[- changed "default" for memsw to be 0MB in the shipped CF file (however, the code change I see is for memsw "reserve_amount" being set to 0MB instead of the prior 64MB, so update the design doc to match what truly got changed)]
There are two defaults -- one in the code when there is no entry in the CF file, and one in the CF file we ship. I prefer these two not to be inconsistent, but I will discuss exactly why the current default is not intuitive for sites (and thus plainly wrong).

test/tests/functional/pbs_cgroups_hook.py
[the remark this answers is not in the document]
alexis-cousein: In the interest of readability:
-sleep 1 (to ensure that any 'old' log line no longer matches the "begin" timestamp)
-begin = time.time()
-sleep 1 (to ensure that rounding errors do not make us "miss" a line caused by the next action
-action
-log inspection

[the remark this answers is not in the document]
alexis-cousein: I have adapted all tests to use this newer flow. And yes, those time.sleep(1) are pretty minimal -- we don't have microsecond logging enabled everywhere ;-). And unfortunately on a wll-know test harness I have proof by existence that small time differences in recorded log times between PTL and the daemons can wreak havoc if you try not to sleep, and the resolution of the log lines is 1s.

test/tests/functional/pbs_cgroups_hook.py
alexis-cousein: Yeah, there is a typo in the earlier message.

src/hooks/cgroups/pbs_cgroups.PY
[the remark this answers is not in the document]
alexis-cousein: Done. Also made some other changes following suggestions.

alexis-cousein:
Actually, I'll have to increase the sleep between the restrt and HUP in setUp once more, because on one (really slow) test harness we got this:

03/17/2021 15:36:06.477656;0100;pbs_mom;Req;;Type 85 request received from root@10.130.166.248:15001, sock=4
03/17/2021 15:36:06.477662;0080;pbs_mom;Hook;pbs_cgroups.PY;copy hook-related file request received
03/17/2021 15:36:06.527347;0c00;pbs_mom;TPP;pbs_mom(Main Thread);tpp_send;**** sd=4, compr_len=41, len=41, dest_sd=26
03/17/2021 15:36:06.527391;0002;pbs_mom;n/a;catch_hup;caught signal 1

And apparently, sending a HUP signal while MoM is processing hook copy requests is unsafe and leaves you with a corrupted hook:
03/17/2021 15:36:06.648844;0400;pbs_python;Svr;pbs_python;Compiling script file: </var/spool/pbs/mom_priv/hooks/pbs_cgroups.PY>
03/17/2021 15:36:06.727167;0001;pbs_python;Svr;pbs_python;PBS server internal error (15011) in Failed to compile script, <class 'SyntaxError'>
03/17/2021 15:36:06.727199;0001;pbs_python;Svr;pbs_python;PBS server internal error (15011) in Failed to compile script, ('EOL while scanning string literal', ('<embedded code object>', 4448, 31, "                if 'limit_in_b\n"))
03/17/2021 15:36:06.727217;0800;pbs_python;Hook;hook_perf_stat;label=hook_exechost_startup_pbs_cgroups_4641.in action=run_code walltime=0.078367 cputime=0.020000
03/17/2021 15:36:06.727320;0100;pbs_python;Hook;pbs_cgroups;exechost_startup hook 'pbs_cgroups' encountered an exception, request rejected

alexis-cousein:
All pbs_cgroups PTL tests run on Test Harness. 

Results:


  | TEST INSTANCES

6569 | cgroups_cousein | Tests from given sources on platforms CENTOS7, SUSE15, CENTOS8 | alexis-cousein/pbspro@cgroups_cgswap | 17th Mar 2021 05:41:43.264 PM +01:00 | 00:19:53.859 | 52 | 0 | 0 | 0 | 0 | 0 | 13 | 39

13 skipped, 39 passed, 0 failures.

I reran the PTL tests that were skipped and are relevant to this PR on my own machine - all seem to pass as well.

test/tests/functional/pbs_cgroups_hook.py
nishiya: How about:
1) submit job 1
2) expect job 1 to be in R state
3) submit job 2
4) call run_scheduling_cycle 
5) expect job 2 in Q state

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -4334,6 +4488,284 @@ def test_cgroup_default_config(self):
         self.server.delete(id=jid, wait=True)
         self.assertFalse(self.is_dir(ehjd1, ehost1), "job cpuset dir found")
 
+    def test_cgroup_cgswap(self, vnode_per_numa_node=False):
+        if not self.mem:
```
nishiya, line 4492: Please add a docstring.

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -4334,6 +4488,284 @@ def test_cgroup_default_config(self):
...
+            j = Job(TEST_USER, attrs=a)
+            j.create_script(self.sleep100_job)
+            jid = self.server.submit(j)
+
+            # Second job should not run - not enough cgswap
+            # offset large enough to give scheduler a chance
+            bs2 = {'job_state': 'Q'}
+            self.server.expect(JOB, bs2, jid, offset=20, max_attempts=1)
+
+    def test_cgroup_cgswap_numa(self):
```
nishiya, line 4613: please add a docstring

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -4334,6 +4488,284 @@ def test_cgroup_default_config(self):
...
+        self.test_cgroup_cgswap(vnode_per_numa_node=True)
+
+    def test_cgroup_enforce_default(self,
```
nishiya, line 4616: please add a doctoring

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -4334,6 +4488,284 @@ def test_cgroup_default_config(self):
...
+                                 " (%d)" % vmem_avail_in_bytes)
+            else:
+                self.assertTrue(mem_limit_in_bytes == vmem_limit_in_bytes,
+                                "no swap, mem (%d) and vmem (%d) limits "
+                                "should be identical but are not"
+                                % (mem_limit_in_bytes, vmem_limit_in_bytes))
+                self.logger.info("no swap: job memsw limit is job mem limit")
+
+    def test_cgroup_enforce_default_tf(self):
+        self.test_cgroup_enforce_default(enforce_flags=('true', 'false'))
```
nishiya, line 4759: Please add a docstring.  Especially to clarify enforce_flags and which setting is being set to what without having to find the self.cfg16 and find it, i.e. memory.enforce_default=true and memsw.enforce_default=false

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -4334,6 +4488,284 @@ def test_cgroup_default_config(self):
...
+
+    def test_cgroup_enforce_default_ft(self):
```
nishiya, line 4761: Please add a docstring

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -4334,6 +4488,284 @@ def test_cgroup_default_config(self):
...
+        self.test_cgroup_enforce_default(enforce_flags=('false', 'true'))
+
+    def test_cgroup_enforce_default_exclhost(self):
```
nishiya, line 4764: Please add a docstring

nishiya (changes requested):
Thanks for making changes.  Things are looking pretty good.  I have a few more comments.

test/tests/functional/pbs_cgroups_hook.py
alexis-cousein: Done

test/tests/functional/pbs_cgroups_hook.py
alexis-cousein: Done

test/tests/functional/pbs_cgroups_hook.py
alexis-cousein: Done

test/tests/functional/pbs_cgroups_hook.py
alexis-cousein: Done

test/tests/functional/pbs_cgroups_hook.py
alexis-cousein: Done

test/tests/functional/pbs_cgroups_hook.py
alexis-cousein: Done

alexis-cousein:
"Can you add a comment to the code? Sorry if I missed where you added it..."

You missed it: it was at the top of the function. But I reworded that top section and added reminders before the lines that puzzled you.

alexis-cousein:
At some places in pbs_cgroups_hook.py PTL tests, I see that we have (rare) failures when doing an action and scanning the logs for the results: we record a start time to look at the logs but there is some jitter between the time sen by daemons (and thuis the logs) and PTL, and time.time() records a float whereas time in the logs jumps in discrete 1s steps.

I've normalized all tests to sleep 2 seconds before and after we record the time at which we start looking in the logs (we want to avoid matching old log lines, but we don't want to miss new ones from the action), and I also convert the minimal timestamp to look for to int (because the logs themselves do not use fractional time, so if we say "start at 10:23:2.3" we may miss events at 10:23:2.9 since they will be logged as occuring "at 10:23:2", which is smaller).

test/tests/functional/pbs_cgroups_hook.py
alexis-cousein: Done

test/tests/functional/pbs_cgroups_hook.py
alexis-cousein: Done

test/tests/functional/pbs_cgroups_hook.py
alexis-cousein: Done

alexis-cousein:
New tests all passed now, and they should be more robust:
6599
cgroups_cousein
Tests from given sources on platforms CENTOS7, SUSE15, CENTOS8
alexis-cousein/pbspro@cgroups_cgswap
19th Mar 2021 02:09:03.792 PM +01:00
00:27:15.204
Total: 52
Skip: 13
Pass: 39

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -1660,7 +1660,7 @@ def load_config(self, cfg, mom_checks=True):
         # copies from setup, node creations, MoM restarts etc.
         # are all finished, so that we don't match a CF copy
         # message in the logs from someone else!
-        time.sleep(2)
+        time.sleep(5)
```
nishiya, line 1663: Did you only change this to get a build?  Can it be changed back?

nishiya (changes requested):
Thanks again for making changes and adding doc strings.  It's getting really close, just a few more comments.

test/tests/functional/pbs_cgroups_hook.py
alexis-cousein: No, I mistakenly changed that to 2 and reverted to 5. The original had 5 -- it's not only sleeping to get a point-in-time, but also to ensure the CF copies are done. I think it's my remit to lower it here.

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -4756,12 +4801,34 @@ def test_cgroup_enforce_default(self,
...
         self.test_cgroup_enforce_default(enforce_flags=('false', 'true'))
 
     def test_cgroup_enforce_default_exclhost(self):
+        """
+        Test to verify if the flags to enforce default mem are working
+        and to ensure mem and memsw limits are set as expected;
+        enforce neither mem nor memsw by enabling flags to ignore
+        enforcement for exclhost jobs and submitting an exclhost job:
+        job should be able to consume all physical memory
+        fand memsw set as limit for all jobs
```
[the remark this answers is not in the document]
alexis-cousein: Fixed.

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -4489,6 +4490,13 @@ def test_cgroup_default_config(self):
         self.assertFalse(self.is_dir(ehjd1, ehost1), "job cpuset dir found")
 
     def test_cgroup_cgswap(self, vnode_per_numa_node=False):
+        """
+        Test to verify (with vnode_per_numa_node enabled):
```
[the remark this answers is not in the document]
alexis-cousein: Yes, cut and paste error.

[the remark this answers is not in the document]
alexis-cousein: Fixed.

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -4539,18 +4547,20 @@ def test_cgroup_cgswap(self, vnode_per_numa_node=False):
             self.logger.info('First Mom has no swap, test will just '
                              'check if job cgswap is added')
             a = {'Resource_List.select':
-                 '1:ncpus=1:mem=100mb:vmem=1100mb:vnode=%s'
+                 '1:ncpus=0:mem=100mb:vmem=1100mb:vnode=%s'
```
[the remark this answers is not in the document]
alexis-cousein: It is foolish to _assume_ that your test harness has two cpus if you don't need to assume it.

If you don't do this, then you would in theory need to add a skip test so that if resources_available.ncpus is less than 2 you skip part of the test -- cumbersome.

Here we're only testing memory, so we can get away with asking for zero cpus (which will simply put job tasks in /sys/fs/cgroup/cpuset/pbs_jobs.service/jobid by default and give them a low number of shares).

alexis-cousein:
BTW, I reworked most log_match() in the file. They often used a default interval and value of n, which means 'scan the log from the end for 50 lines, and repeat as fast as you can'. Given the cgroup hook can actually be really verbose and that PTL can be slowed down on some VMs that don't run PTL and the daemons at a decent speed, I've seen spurious failures in which one attempt didn't see the lines because they weren't there yet, and the next attempt (a whole _three seconds_ later despite interval=None!) didn't see them because they had already "scrolled through" the 50 line buffer at the end of the file.

Again, that is not something you'll see on a "regular" testing machine, but if you run tests on small VMs that compate with other VMs on a host you'll sometimes trip over it.

The most sane thing to avoid this is to set a timestamp we must not cross when reading from the end, and then always read from the end back _to that point_ in the consecutive attempts to find the line. (i.e. use starttime=stime, n='ALL')

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -1539,13 +1684,16 @@ def load_config(self, cfg, mom_checks=True):
                 r2 = self.du.run_cmd(cmd=['cat', mom_conf], sudo=True)
                 if r1['out'] != r2['out']:
                     self.logger.info('server & mom pbs_cgroups.CF differ')
-                    just_before_import = time.time() - 1
+                    time.sleep(2)
+                    just_before_import = int(time.time())
+                    time.sleep(2)
```
agrawalravi90, line 1689: How about calling self.log_lines() with n=1 on the mom where you want to call log_match() later on, this will give you a log line from that host's mom logs. you can then use it to get a timestamp from that machine. Something like this (untested):
lastlog = self.moms_list[0].log_lines(n=1)
timestamp = float(self.logutils.convert_date_time(lastlog[0].split(";", 1)[0]))

alexis-cousein: That is a good idea -- feel free to submit your own PR to make it all more elegant, but I'm now just going to ensure it works -- including the tests that I did not write.

As you said: your idea has merit, but it's "untested". This isn't, it's tested _ad nauseam_ by now.

alexis-cousein: I chose not to engage in something like that because it would gain us 2 seconds in tests at the cost of readability.

alexis-cousein:
Had to fix one test on a related subject that was wrong. In "memsw", both reserve_amount and default denote additional _swap_ to reserve or add to the physical memory quantities in "memory" (otherwise one could obtain nonsensical configurations where the default for memsw would be lower than that for memory).

Yet for a memory default of 96MB and a memsw default of 96MB, the test expected the memory.memsw.limit_in_bytes to be equal to memory.limit_in_bytes, instead of double the amount (96MB of memory and 96MB of swap).

Unfortunately, the behaviour of the test is indeed consistent with current documentation, but this is nonsensical and a defect in functionality -- it prevents people from saying "by default, no swap"  by setting the memsw default to "0MB" and would make the value in the "memsw" section _depend_ on the value in the "memory" section. That was changed for reserve_amount and reserve_percent for similar reasons, and those reasons apply also to the default. We want both sections to control _different_ things, one physical memory, the other swap.

That will have to be mentioned in the design document, though.

alexis-cousein:
Clarified the meaning of "default" in the "memsw" section in the design document (and its new default of '0B' that means 'add no swap usage by default').

alexis-cousein:
There are a number of other problems in mem/memsw tests. One of them is that the tests assume that if self.swapctl is 'true', then there _is_ swap. But on some hosts (including the test harness tried), swap management is turned on in the kernel (which makes swapctl 'true') but there is no swap device, so the amount of swap is 0. So a correct test for 'memsw' should determine what the limit should be depending on what the hook saw as available total vmem.

In another test purely on mem, a test is also skipped if self.swpactl is not 'true', but it should check on self.mem instead (or it will skip on hosts that should run the test).

alexis-cousein:
Found another race in a PTL test: test_cgfroup_find_gpus loads a config file with the "devices" subsection enabled, but does not wait for the exechost_startup to run. 

Since in the current master cgroup device and gpu discovery is suppressed if "devices" is disabled, the server.status(NODE,...) that follows can pick up a node status that does not yet reflect the configuration file that was loaded.

That test was broken before this PR was submitted and _should_ have started to fail because of it when an earlier PR was accepted and merged, but was probably masked by the fact no one ran _that_ test on a platform with GPUs that was also fast enough to query the node status too early.

Incidentally, it also launches an nvidia-smi command without specifying the hosts, which will run it on the server, and not the first MoM. It "happens" to work most of the time since the server is often the first MoM as well, but is not guaranteed to work if you run it differently.

I took the liberty to also fix that test. It now runs properly on my GPU host, and also runs if you run the server and MoM on different hosts.

alexis-cousein:
Made another small change. The find_gpus test would end with "Error" rather than "Failure" if it failed to find resources_available.ngpus on the test node:


2021-03-20 21:46:58,265 INFO     ___m_oo_m___
2021-03-20 21:46:58,265 INFO     Traceback (most recent call last):
  File "/home/alexis/git/github/alexis-cousein/pbspro/test/tests/functional/pbs_cgroups_hook.py", line 3321, in test_cgroup_find_gpus
    id=self.nodes_list[0])[0]

Instead if just dying on an exception, this should be logged as a test _failure_. Now it stats all nodes (so there is always index 0 in the answer) and it asserts that resources_available.ngpus should be present on the test node where nvidia-smi was run before continuing. If it isn't there the test will fail cleanly instead of crashing.

alexis-cousein:
I also saw some spurious failures in test_cgroup_periodic_update_check_values : the test does a loop in which it repeatedly scans to see updates. When using the default n=50, you sometimes miss seeing some updates; when using n='ALL', the different attempts can all match the same update, and in that update the memory usage might not be large enough for success yet.

The fix is to move the starttime for the log matching so that in 10 attempts you are sure that you match 10 different updates (or 9 if you're really unlucky and you match the first update twice).

alexis-cousein:
Hardened test_cgroup_periodic_update_check_values about coming to conclusions too early.

alexis-cousein:
Since there are *so* many tests that need extra care to make sure they don't run off and do things before the exechost_startup was run after they load a config file, I willl move the loop to _check_ whether the config file was fully loaded and the HUP was actually _processed_ completely in load_config itself, which will allow us to simplify a lot of tests again.

alexis-cousein:
Well _that_ simplified a lot of things and allowed me to remove quite a bit of duct tape.

alexis-cousein:
I'll have to match another message in load_config -- the message "hook handler returned success" will only work if the cgroup hook itself is not completely disabled when the configuration file is loaded -- otherwise exchost_startup runs but the message isn't there.

So I'll have to commit once more to make it work...

alexis-cousein:
Fixed another race: the code did a MoM restart and then changed the config file, which involves a HUP. If the HUP arrives too early and MoM is still setting up networking and hasn't sent the restart to the server, it can get confused and hang.

Possibly a MoM bug (already known), but the tests will have to live with it.

alexis-cousein:
There seems to also be a race condition in the test_cgroup_offline_node test. The test starts a job and then tries to freeze processes (using the cgroup freezer controller), which should lead to a failure to clean up the cgroups.

But sometimes the job gets killed "regularly" before the freeze:

03/21/2021 22:32:07.294346;0008;pbs_mom;Job;0.x6631-2-0;walltime 8 exceeded limit 3
03/21/2021 22:32:07.295560;0008;pbs_mom;Job;0.x6631-2-0;kill_job

Obviously, setting a walltime limit of 3 seconds for a job when the speed at which PTL itself acts upon the job is unknown is hazardous at best...

alexis-cousein:
All tests now pass now even with some heave competition on the test harness:

6639 | cgroups_cousein | Tests from given sources on platforms CENTOS7, SUSE15, CENTOS8 | alexis-cousein/pbspro@cgroups_cgswap | 22nd Mar 2021 09:52:39.346 AM +01:00 | 00:31:35.688 | 52 | 0 | 0 | 0 | 0 | 0 | 13 | 39

alexis-cousein:
Reran all cgroup tests on test harness, on three OSes (CentOS7, CentOS8, SLES15) and for good measure repeated with daemons running under Valgrind -- no failures, 117 passes, 39 skips.

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -2039,10 +2039,11 @@ def test_cgroup_periodic_update_check_values(self):
         cput_usage = 0.0
         mem_usage = 0
         vmem_usage = 0
+        time.sleep(8)
         for count in range(10):
+            time.sleep(2)
             # Faster systems might have expected usage after 8 seconds
             # TH3 can take up to a minute
```
nishiya, line 2070: Probably should move this comment to go with the sleep you moved.

nishiya (commented):
Nice work!  I had only the one minor comment about moving the comment along with the sleep.  Other than that it looks good.  Thank you @alexis-cousein for putting in the hardworking to figure out how to make the tests run reliably!

test/tests/functional/pbs_cgroups_hook.py
alexis-cousein: Done -- I'll commit once I dealt with any other change you might suggest (and I agree with ;-) ).

nishiya (approved):
Thanks for the last comment change.
Looks good to me.

bayucan (approved):
Looks good! Thanks @alexis-cousein for improving the code and TestCgroupsHook testsuite! This will greatly benefit the feature!

bayucan:
Merged #2271 into master.

PR #2272, opened 2021-02-26
Snapshot obfuscate errors out + optimizing obfuscate

agrawalravi90:
#### Describe Bug or Feature

Errors out as follows:
  File "/opt/ptl/bin/pbs_snapshot", line 188, in obfuscate_snapshot_wrapper
    obj.obfuscate_snapshot(snap_name, map_file, sudo_val)
  File "/opt/ptl/lib/python3.6/site-packages/ptl/utils/pbs_snaputils.py", line 508, in obfuscate_snapshot
    self.obfuscate_acct_logs(snap_dir, sudo_val)
  File "/opt/ptl/lib/python3.6/site-packages/ptl/utils/pbs_snaputils.py", line 430, in obfuscate_acct_logs
    self._obfuscate_acct_file(attrs_to_obf, acct_fpath)
  File "/opt/ptl/lib/python3.6/site-packages/ptl/utils/pbs_snaputils.py", line 357, in _obfuscate_acct_file
    with open(file_path, "r") as fd, open(fout, "w") as fdout:
FileNotFoundError: [Errno 2] No such file or directory: 'new_snapshot_infra_bigmem/server_priv/accounting/new_snapshot_infra_bigmem/server_priv/accounting/20170617'

- Also, --obfuscate had a performance inefficiency: towards the end, it would open, parse, modify and write all of the files in the snapshot multiple times, once for each key-val pair in the obfuscation map, in order to obfuscate each value. I changed it to open the file, swap all the values in the map and write back, just once.

#### Describe Your Change

DshUtils.listdir returns full paths, not just file names, which is what we were assuming in the obfuscation codepath.

[testsnap.log](https://github.com/openpbs/openpbs/files/6052817/testsnap.log)

[test_capture_server.log](https://github.com/openpbs/openpbs/files/6126788/test_capture_server.log)

### DIFF
--- a/test/tests/functional/pbs_snapshot_unittest.py
+++ b/test/tests/functional/pbs_snapshot_unittest.py
@@ -258,10 +258,11 @@ class TestPBSSnapshot(TestFunctional):
                     common_path = os.path.commonprefix([file_fullpath,
                                                         svr_fullpath])
                     try:
-                        self.assertEqual(os.path.basename(common_path), "server")
+                        self.assertEqual(os.path.basename(common_path),
+                                         "server")
                     except AssertionError:
-                        # Check if this was a core file, which would explain why it
-                        # it was captured
+                        # Check if this was a core file, which would
+                        # explain why it was captured
                         if CORE_DIR in file_fullpath:
                             continue
                         raise
--- a/test/tests/functional/pbs_snapshot_unittest.py
+++ b/test/tests/functional/pbs_snapshot_unittest.py
@@ -261,9 +261,10 @@ class TestPBSSnapshot(TestFunctional):
                         self.assertEqual(os.path.basename(common_path),
                                          "server")
                     except AssertionError:
-                        # Check if this was a core file, which would
+                        # Check if this was a server core file, which would
                         # explain why it was captured
-                        if CORE_DIR in file_fullpath:
+                        svrcorepath = os.path.join(CORE_DIR, "server_priv")
+                        if svrcorepath in file_fullpath:
                             continue
                         raise
             # Check 3: qstat_Bf.out exists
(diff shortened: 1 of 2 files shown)


test/tests/functional/pbs_snapshot_unittest.py
```diff
@@ -257,7 +257,15 @@ def test_capture_server(self):
...
                     common_path = os.path.commonprefix([file_fullpath,
                                                         svr_fullpath])
-                    self.assertEqual(os.path.basename(common_path), "server")
+                    try:
+                        self.assertEqual(os.path.basename(common_path),
+                                         "server")
+                    except AssertionError:
+                        # Check if this was a core file, which would
+                        # explain why it was captured
+                        if CORE_DIR in file_fullpath:
```
npadole20, line 266: As part of capture_server(), will we also capture core files in another directory apart from the server ones?

agrawalravi90: No we don't, we only capture server's core files. The thing is, we store them in a separate path inside the snapshot: <snapshot name>/core_file_bt
The test is checking that the only data captured was inside directories that shared a common path with <snapshot name>/server, which 'core_file_bt' will not. So, that's why I made the exception.

Did you want me to check that CORE_DIR didn't capture core files from other daemons?

npadole20: Yes. I think we should do that. It should be simple as we are creating a different directory for every daemon. If we have only the server_priv directory in core_file_bt/ that should be good enough check.

agrawalravi90:
@npadole20 thanks for your review, I've addressed your comment, attached test log after making the latest changes, i had a core file in mom_priv as well as server_priv and it was able to capture only server .. please provide further feedback

npadole20 (approved):
Thanks for making the changes.

agrawalravi90:
@openpbs/openpbs-maintainers  need a maintainer review for this
@npadole20 thanks for the review!

agrawalravi90:
@openpbs/openpbs-maintainers gentle reminder, thanks

hirenvadalia (approved):

PR #2277, opened 2021-03-02
Incorrect Job array ids generated for multi-server

agrawalravi90:
#### Describe Bug or Feature

After the jobid namespace redesign change went in, a bug was introduced which caused job array ids to be generated incorrectly: the '[]' came before the numeric portion of server id, leading to jobids such as 0[]00.pbspro.

Also, changed msvr_node_group's value to be set to the position of the server in PBS_SERVER_INSTANCES so that even if PBS_SERVER is the same for all servers, msvr_node_group is set to different values for nodes of each server.

### DIFF
--- a/src/server/multi_svr.c
+++ b/src/server/multi_svr.c
@@ -62,6 +62,7 @@ typedef struct peersvr_list {
 static struct peersvr_list *peersvrl;
 extern char server_host[PBS_MAXHOSTNAME + 1];
 extern unsigned int	pbs_server_port_dis;
+static int svridx = -1;
 
 /**
  * @brief
@@ -279,6 +280,18 @@ init_msi()
 {
 	peersvrl = NULL;
 
+	/* Determine the index of the server in PBS_SERVER_INSTANCES list */
+	if (svridx == -1) {
+		int i;
+		for (i = 0; i < get_num_servers(); i++) {
+			if (pbs_conf.psi[i].port == pbs_server_port_dis &&
+			    is_same_host(pbs_conf.psi[i].name, server_host)) {
+				svridx = i;
+				break;
+			}
+		}
+	}
+
 	return 0;
 }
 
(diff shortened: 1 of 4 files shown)


src/server/node_func.c
```diff
@@ -360,19 +361,18 @@ initialize_pbsnode(struct pbsnode *pnode, char *pname, int ntype)
...
-		 * server_id is the id of the server for the node */
-		prd = &svr_resc_def[RESC_MSVR_ND_GROUP];
-		if ((svr_nd_grp = add_resource_entry(pat1, prd)) != NULL) {
-			char buf[PBS_MAXHOSTNAME];
-
-			snprintf(buf, sizeof(buf), "%s", pbs_server_name);
-			decode_arst(&svr_nd_grp->rs_value, NULL, NULL, buf);
-		}
-	}
+    if (get_num_servers() > 1) {
```
nithinj, line 364: Use msvr_mode()

src/server/multi_svr.c
```diff
@@ -306,3 +308,31 @@ gen_svr_inst_id(void)
...
+ * @return	int
+ * @retval	index of the server
+ * @retval	-1 if couldn't be determined
+ */
+int
+get_server_index(void)
+{
+	static int svridx = -1;
+
+	if (svridx == -1) {
```
nithinj, line 326: Instead of calculating here, set the Index when we find the self in init_msi.

nithinj (changes requested):

src/server/multi_svr.c
agrawalravi90: "set the Index when we find the self in init_msi."

I'm not sure I understand. Is there any advantage of calculating it in init_msi? I'll have to create a global if I calculate it in init_msi() so that get_server_index() can return it, I'd rather not go in that direction. How does it help?

agrawalravi90:
@nithinj thanks for the review, I've addressed your comments, please provide further feedback.

src/server/multi_svr.c
nithinj: init_msi will change and follows a similar logic with https://github.com/openpbs/openpbs/pull/2253
so it will be easier to merge.
We can just call one function whenever PSI changes.
The variable is already static, the suggested change requires it to be in the file's scope.

agrawalravi90:
@nithinj thanks, I've moved the calculation to init_msi(), let me know if it looks ok now.

agrawalravi90:
@openpbs/openpbs-maintainers need a maintainer review for this PR, thanks.

arungrover (approved):
LGTM

agrawalravi90:
@arungrover  thanks for taking this up!

nithinj (approved):
LGTM!

subhasisb (approved):

agrawalravi90:
@nithinj  @subhasisb and @arungrover thanks for the sign-off, I just resolved a conflict, please check it out and let me know if it still looks good.

agrawalravi90:
Regression results:

Profile: regression
|--|--|--|--|--|--|--|
|6359|3990|0|0|0|0|3990|

PR #2279, opened 2021-03-03
Fix qdel race condition in test_job_sort_formula_threshold

visheshh:
#### Describe Bug or Feature
qdel failed on some machines as the job with walltime 4 was over by the time we delete the job.

#### Describe Your Change
Let the qdel unknown job id error passby and use force delete 

[smoke_threshold.txt](https://github.com/openpbs/openpbs/files/6092612/smoke_threshold.txt)

Ran smoketest on th3.
Profile: regression
|--|--|--|--|--|--|--|
|6290|52|0|0|0|0|52|

### DIFF
--- a/test/tests/pbs_smoketest.py
+++ b/test/tests/pbs_smoketest.py
@@ -985,9 +985,12 @@ class SmokeTest(PBSTestSuite):
         msg = "Checking the job state of %s, runs after %s is deleted" % (j3id,
                                                                           j4id)
         self.logger.info(msg)
-        attribs = self.server.status(JOB, 'job_state', id=j4id)
-        if attribs[0]['job_state'] == 'R':
-            self.server.deljob(id=j4id, wait=True)
+        try:
+            self.server.deljob(id=j4id, wait=True, extend='force',
+                               runas=MGR_USER)
+        except PbsDeljobError as e:
+            self.assertIn(
+                'qdel: Unknown Job Id', e.msg[0])
         self.server.expect(JOB, {'job_state': 'R'}, id=j3id, max_attempts=30,
                            interval=2)
         self.server.expect(JOB, {'job_state': 'Q'}, id=j2id, max_attempts=30,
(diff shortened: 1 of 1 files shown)


birbalsain (approved):
LGTM

test/tests/pbs_smoketest.py
```diff
@@ -985,7 +985,9 @@ def test_job_sort_formula_threshold(self):
         msg = "Checking the job state of %s, runs after %s is deleted" % (j3id,
                                                                           j4id)
         self.logger.info(msg)
-        self.server.deljob(id=j4id, wait=True)
+        attribs = self.server.status(JOB, 'job_state', id=j4id)
+        if attribs[0]['job_state'] == 'R':
+            self.server.deljob(id=j4id, wait=True)
```
bayucan, line 990: I think the fix should be in self.server.deljob() to handle a case when the job actually exited/no longer exist, and should not hang...

bayucan (changes requested):

test/tests/pbs_smoketest.py
visheshh: @bayucan It does not hang. But it throws unknown job id error as qdel would if the job didnt exist.

bayucan: @visheshh : Then how about just doing a try...else around self.server.deljob() and just allow for the "Unkonwn jobid" condition? Even if you check if the job's state is R, what if by the time the self.server.deljob() executes, job could have disappeared quickly...

visheshh: @bayucan Thanks for the suggestion. I will do that.

agrawalravi90 (changes requested):
There's still a race condition that this doesn't fix: by the time you did a status() and called deljob(), the job might have gotten deleted, right? How about making deljob() not throw any errors? what if you pass logerr=False ?

visheshh:
> 
> 
> There's still a race condition that this doesn't fix: by the time you did a status() and called deljob(), the job might have gotten deleted, right? How about making deljob() not throw any errors? what if you pass logerr=False ?

Yes , I was thinking to go ahead with @bayucan suggestion, to use try and let the error pass by.

test/tests/pbs_smoketest.py
bhroam: I'm not sure we want to have unknown job id not throw an exception.  I mean it IS an error.  It just happens that some of our tests don't care.  I think it should be up to the test to catch the exception if they don't care.  Either that or pass an argument that you don't care about it.  In any case, using a walltime of 4s is too small for our slow testbed machines.

agrawalravi90:
> Yes , I was thinking to go ahead with @bayucan's suggestion, to use try and let the error pass by.

Sure, whatever you prefer

test/tests/pbs_smoketest.py
vstumpf: This also won't help if the job is in R state when you status it, but is deleted by the time you deljob it. 

I agree with bhroam's suggestion of passing an argument if you want to throw an exception if the jobid doesn't exit.

visheshh:
@agrawalravi90 @bayucan @bhroam @vstumpf Please review the changes.
I got unknown Job id and could not qdel as job was in E state error.
I reduce job sleep to 2 secs and was able to reproduce it consistently on testbed and with these changes,
i.e, with Try expect and qdel -W force I was able to bypass them.

agrawalravi90 (approved):
LGTM

agrawalravi90 (changes requested):
Actually, please change the title of your PR to describe the problem.

bayucan (approved):
Looks good.

bhroam (approved):

visheshh:
@agrawalravi90 Please review

agrawalravi90 (approved):

PR #2280, opened 2021-03-03
Enhancing pbs_locjob for multi-server

agrawalravi90:
#### Describe Bug or Feature

Right now, pbs_locjob cannot handle a multi-server setup. This leads to unexpected behavior of commands like qdel which use it  to locate jobs which were moved to remote clusters. So, this PR enhances it to handle a multi-server configuration.

The following shows how qdel was not reporting "unknown job" for an already deleted job, and that gdb showed how locate_job was getting a SIGPIPE
```
[ragrawal@m2 ~]$ qstat
101.m2-3          STDIN            ragrawal          00:00:00 R workq           
201.m2-3          STDIN            ragrawal          00:00:00 R workq           
[ragrawal@m2 ~]$ qdel 900.m2
[ragrawal@m2 ~]$ gdb qdel
GNU gdb (GDB) Red Hat Enterprise Linux 8.2-12.el8
Copyright (C) 2018 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-redhat-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from qdel...
warning: the debug information found in "/usr/lib/debug//opt/pbs/bin/qdel-20.0.0-0.x86_64.debug" does not match "/opt/pbs/bin/qdel" (CRC mismatch).

Missing separate debuginfo for /opt/pbs/bin/qdel
Try: yum --enablerepo='*debug*' install /usr/lib/debug/.build-id/f0/29deed9dc7f8e5b9a6ca220f749c91358bbfad.debug
Reading symbols from .gnu_debugdata for /opt/pbs/bin/qdel...(no debugging symbols found)...done.
(no debugging symbols found)...done.
(gdb) b locate_job
Breakpoint 1 at 0x403860
(gdb) run 900.m2
Starting program: /opt/pbs/bin/qdel 900.m2
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
warning: Loadable section ".note.gnu.property" outside of ELF segments
warning: Loadable section ".note.gnu.property" outside of ELF segments
[Detaching after fork from child process 38457]
[Detaching after fork from child process 38458]

Breakpoint 1, 0x0000000000403860 in locate_job ()
(gdb) finish
Run till exit from #0  0x0000000000403860 in locate_job ()
[Detaching after fork from child process 38459]
[Detaching after fork from child process 38460]

Program received signal SIGPIPE, Broken pipe.
0x00007ffff7bc5ac5 in write () from /lib64/libpthread.so.0
(gdb) 
```

After fix:
```
[ragrawal@m2 ~]$ qstat
1.m2              STDIN            ragrawal          00:00:00 R workq           
2.m2              STDIN            ragrawal          00:00:00 R workq           
3.m2              STDIN            ragrawal          00:00:00 R workq           
4.m2              STDIN            ragrawal          00:00:00 R workq           
700.m2            STDIN            ragrawal                 0 Q workq           
600.m2            STDIN            ragrawal          00:00:00 R workq           
001.m2-3          STDIN            ragrawal                 0 Q workq           
[ragrawal@m2 ~]$ qdel 700.m2
[ragrawal@m2 ~]$ qdel 700.m2
qdel: Unknown Job Id 700.m2
```

### DIFF
--- a/src/lib/Libifl/pbsD_locjob.c
+++ b/src/lib/Libifl/pbsD_locjob.c
@@ -51,31 +51,23 @@
 #include "dis.h"
 #include "pbs_ecl.h"
 
-
 /**
- * @brief
- *      This function does the LocateJob request.
+ * @brief	Helper function for __pbs_locjob, sends the locjob request using a real sd
  *
  * @param[in] c - connection handler
  * @param[in] jobid - job identifier
  * @param[in] extend - string to encode req
  *
- * @return      string
- * @retval      destination name	success
- * @retval      NULL      		error
- *
+ * @return	string
+ * @retval	destination name	success
+ * @retval	NULL	error
  */
 char *
-__pbs_locjob(int c, char *jobid, char *extend)
+pbs_locjob2(int c, char *jobid, char *extend)
 {
-	int	rc;
+	int rc;
 	struct batch_reply *reply;
-	char       *ploc = NULL;
-
-	if ((jobid == NULL) || (*jobid == '\0')) {
-		pbs_errno = PBSE_IVALREQ;
-		return (ploc);
-	}
+	char *ploc = NULL;
 
 	/* initialize the thread context data, if not already initialized */
 	if (pbs_client_thread_init_thread_context() != 0)
@@ -133,4 +125,48 @@ __pbs_locjob(int c, char *jobid, char *extend)
 		return NULL;
 
 	return ploc;
+}
+
+/**
+ * @brief
+ *      This function does the LocateJob request.
+ *
+ * @param[in] c - connection handler
+ * @param[in] jobid - job identifier
+ * @param[in] extend - string to encode req
+ *
+ * @return	string
+ * @retval	destination name	success
+ * @retval	NULL	error
+ */
+char *
+__pbs_locjob(int c, char *jobid, char *extend)
+{
+	if ((jobid == NULL) || (*jobid == '\0')) {
+		pbs_errno = PBSE_IVALREQ;
+		return NULL;
+	}
+
+	if (multi_svr_op(c)) {
+		svr_conn_t **svr_conns = get_conn_svr_instances(c);
+		int start = 0;
+		int ct;
+		int nsvrs = get_num_servers();
+		char *ret = NULL;
+		int i;
+
+		if ((start = get_obj_location_hint(jobid, MGR_OBJ_JOB)) == -1)
+		    start = 0;
+
+		for (i = start, ct = 0; ct < nsvrs; i = (i + 1) % nsvrs, ct++) {
+			if (!svr_conns[i] || svr_conns[i]->state != SVR_CONN_STATE_UP)
+				continue;
+			ret = pbs_locjob2(svr_conns[i]->sd, jobid, extend);
+			if (ret)
+				break;
+		}
+		return ret;
+	} else
+		return pbs_locjob2(c, jobid, extend);
+
 }
--- a/src/lib/Libifl/int_manager.c
+++ b/src/lib/Libifl/int_manager.c
@@ -166,7 +166,6 @@ PBSD_manager(int c, int rq_type, int command, int objtype, char *objname, struct
 				if (rc == PBSE_NONE || (pbs_errno != PBSE_UNKJOBID && pbs_errno != PBSE_UNKRESVID))
 					break;
 			}
-			
 		}
 
 		return rc;


src/lib/Libifl/pbsD_locjob.c
```diff
@@ -134,3 +126,47 @@ __pbs_locjob(int c, char *jobid, char *extend)
...
+		int i;
+
+		if ((start = get_obj_location_hint(jobid, MGR_OBJ_JOB)) == -1)
+		    start = 0;
+
+		for (i = start, ct = 0; ct < nsvrs; i = (i + 1) % nsvrs, ct++) {
+			if (!svr_conns[i] || svr_conns[i]->state != SVR_CONN_STATE_UP)
+				continue;
+			ret = pbs_locjob2(svr_conns[i]->sd, jobid, extend);
+			if (ret)
```
nithinj, line 165: If you are getting this return value correctly, then we might be recording the internal server moves into the tracking table. Should we do that? That might be a huge no of jobs to track and what's the point if they are part of the same cluster?
Instead, we should record into the tracking table only if the job moves to a remote cluster.
The problem now would be that if a qmove happens after the internal move; pbs_locjob will fail to report that.

agrawalravi90: I don't think we support qmove between servers of the same cluster (especially with PBS_SERVER values being the same, I don't see how it'll work). So, can we wait until we add support for qmove between internal servers (if at all)? If not, please let me know if you have a suggestion, I don't think we can handle it without knowing how qmove between internal servers will work.

nithinj: PBS_SERVER values can be distinct as well between instances in a multi-server right?
I was talking about the internal move that happens when the scheduler decides to run the job in a different instance.

agrawalravi90: > PBS_SERVER values can be distinct as well between instances in a multi-server right?
Even if PBS_SERVER is different, if they are on the same host, they will all resolve to the same hostname, so it still cannot work.

> I was talking about the internal move that happens when the scheduler decides to run the job in a different instance.
If a job is moved from one multi-server to another, do we keep its information around in db? I thought we delete it from the original server instance. If we don't, I think we should just delete it as the job is still in the same queue, so moving from one multi-server to another isn't technically a qmove. Refer to the definition of qmove from the ref guide:

_"2.48 qmove
Moves a PBS job from one queue to another
"_

nithinj: > Even if PBS_SERVER is different, if they are on the same host, they will all resolve to the same hostname, so it still cannot work.

What if they are on different hosts?

>  so moving from one multi-server to another isn't technically a move

Agreed! That's what I'm trying to say. 
As you are enhancing pbs_locjob for multi-svr, we should make sure that internal moves are not being tracked. If we do so, pbs_locjob could still return the wrong internal server instead of a remote server for a multi-svr case.

agrawalravi90: ok, I see your concern, and I'd have loved to test it out, but right now qmove between a cluster with multi-server and a cluster with single server is broken, so I cannot test out whether this works or not. But, the point that I'm trying to make is, if it doesn't work, then it'd be something to fix on the server side, not pbs_locjob. The server should delete a job after it's moved to another instance in the local cluster, and then the changes that this PR adds will make sure that all instances are sent the request to know which local/remote server the job was moved to. Does that make sense?

arungrover: Interesting conversation! When the qmove across complexes issue is resolved, we should test what happens when a job is moved from a single server cluster to a multi-server cluster and then moved locally by the scheduler. Does the pulling complex report back the finished status of the job? and does pbs_locjob issued on furnishing complex work in this scenario?

arungrover (approved):
LGTM

src/lib/Libifl/pbsD_locjob.c
agrawalravi90: ya, that'd be interesting to test .. i'm not completely sure how it'll work since the single server cluster won't know much about the multi-server cluster .. this is something that we'll probably not be able to support in the first version.

nithinj (approved):
LGTM!

PR #2281, opened 2021-03-04
Fix interactive test in PTL

RKORavi:
#### Describe Bug or Feature
Interactive job in PTL executes only for few seconds, irrespective of job's sleep time

#### Describe Your Change
Remove timeout argument from read_nonblocking function call. Instead a default timeout will be used 

Added sleep time of 8 secs in test_interactive_job to show changes
[after_fix.txt](https://github.com/openpbs/openpbs/files/6083424/after_fix.txt)
[before_fix.txt](https://github.com/openpbs/openpbs/files/6083425/before_fix.txt)

### DIFF
--- a/test/fw/ptl/lib/ptl_resourceresv.py
+++ b/test/fw/ptl/lib/ptl_resourceresv.py
@@ -652,7 +652,8 @@ class InteractiveJob(threading.Thread):
             _p.sendline("exit")
             while True:
                 try:
-                    _p.read_nonblocking(timeout=5)
+                    # When -1 (default), use self.timeout
+                    _p.read_nonblocking()
                 except Exception:
                     break
             if _p.isalive():
--- a/test/fw/ptl/lib/ptl_resourceresv.py
+++ b/test/fw/ptl/lib/ptl_resourceresv.py
@@ -642,8 +642,13 @@ class InteractiveJob(threading.Thread):
                 self.job.interactive_handle = None
                 return None
             self.logger.debug(_p.after.decode())
+            _to = 5
             for _l in _sc:
                 (cmd, out) = _l
+                if 'sleep ' in cmd:
+                    timev = cmd.split(' ')[1]
+                    if timev.isnumeric():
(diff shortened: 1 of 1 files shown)


test/fw/ptl/lib/ptl_resourceresv.py
```diff
@@ -652,7 +652,8 @@ def run(self):
             _p.sendline("exit")
             while True:
                 try:
-                    _p.read_nonblocking(timeout=5)
+                    # When -1 (default), use self.timeout
+                    _p.read_nonblocking()
```
Shrini-h, line 656: LGTM, @RKORavi  it would be great if you can simulate a time out during non blocking read

Shrini-h (approved):

test/fw/ptl/lib/ptl_resourceresv.py
birbalsain: while creating "_p"  value of timeout is set by "pexpect_timeout = 15". So self.timeout in this case is 15 sec. does it sufficient for interactive job?

RKORavi: Hey @birbalsain, timeout in this case will be 30 secs. Because when we do not pass timeout it will use self.timeout value in pexpect. Please refer this link: https://pexpect.readthedocs.io/en/3.x/_modules/pexpect.html
I will test it again with job's sleep greater than 30 secs.

RKORavi: @birbalsain yes, you are right. timeout will be 15 & not 30 secs. Thanks for pointing that out. I will make the changes accordingly.

RKORavi:
I have changed timeout value to be same as job's sleep time. If a job doesn't have a sleep time, then timeout is set to 5.

birbalsain (approved):

Shrini-h (approved):

hirenvadalia (approved):

PR #2282, opened 2021-03-04
Fix race condition in TestAcctlogRescUsedWithTwoMomHooks

Klovely:
#### Describe Bug or Feature
'test_Rrecord'and 'test_Erecord' of 'TestAcctlogRescUsedWithTwoMomHooks' failed on slow machine  some time with below error 

ptl.lib.ptl_error.PtlExpectError: rc=1, rv=False, msg=expected on server x50-xyz:  no data for job_state = R && substate = 42 job 2.xyz attempt: 180

Reason :- Job is of only sleep 5 sec in test_Rrecord and of sleep 1sec in test_Erecord , so there is race condition  on slow machine where some time there is delay of 5 sec to respond in command , in that case job cross the sleep time by the time job test look for job in R state with substate 42 , job is already went into E state  or finished and fail in PTL expect while 

self.server.expect(JOB, {ATTR_state: 'R'}, jid1)

#### Describe Your Change
To avoid the race condition increase sleep time in job script to 15sec

Before fix 

[TestAcctlogRescUsedWithTwoMomHooks_before_fix_1.txt](https://github.com/openpbs/openpbs/files/6083470/TestAcctlogRescUsedWithTwoMomHooks_before_fix_1.txt)

After fix 

[TestAcctlogRescUsedWithTwoMomHooks_after_fix_10.txt](https://github.com/openpbs/openpbs/files/6083471/TestAcctlogRescUsedWithTwoMomHooks_after_fix_10.txt)

### DIFF
--- a/test/tests/functional/pbs_two_mom_hooks_resources_used.py
+++ b/test/tests/functional/pbs_two_mom_hooks_resources_used.py
@@ -96,15 +96,11 @@ class TestAcctlogRescUsedWithTwoMomHooks(TestFunctional):
         resources_used value to be present in the 'R' record.
         """
 
-        test = []
-        test += ['#PBS -N NodeFailRequeueTest\n']
-        test += ['echo Starting test at `date`\n']
-        test += ['sleep 15\n']
-
+        # Submit job
         select = "vnode=" + self.hostA + "+vnode=" + self.hostB
         j1 = Job(TEST_USER, attrs={
-            'Resource_List.select': select})
-        j1.create_script(body=test)
+             ATTR_N: 'NodeFailRequeueTest',
+             'Resource_List.select': select})
         jid1 = self.server.submit(j1)
 
         # Wait for the job to start running.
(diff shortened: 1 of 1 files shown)


Klovely:
Note :  After fixing the race condition for test_Erecord observed its failing on further step due to product issue resource_used.walltime is coming as 0  in qstat -fx output after job finishes
Logged issue for same .

But race condition issue is got fixed and test is working fine  at step where test looking for Job  is in R state .

dinesh1306 (approved):

test/tests/functional/pbs_two_mom_hooks_resources_used.py
```diff
@@ -99,7 +99,7 @@ def test_Rrecord(self):
         test = []
         test += ['#PBS -N NodeFailRequeueTest\n']
         test += ['echo Starting test at `date`\n']
-        test += ['sleep 5\n']
```
bhroam, line 102: Nothing in this test requires this job script.  It also doesn't wait for the job to finish, so using the default sleep of 100 is probably best.  I'd remove the script, and just leave the sleep time alone.

test/tests/functional/pbs_two_mom_hooks_resources_used.py
```diff
@@ -129,7 +129,7 @@ def test_Erecord(self):
         test = []
         test += ['#PBS -N JobEndTest\n']
         test += ['echo Starting test at `date`\n']
-        test += ['sleep 1\n']
+        test += ['sleep 15\n']
```
bhroam, line 132: This test also doesn't care about this job script.  I'd remove the job script, but the sleep time is right.  I'd set the sleep time of the job to 15 like you are (but in j1.set_sleep_time())

bhroam (changes requested):

Klovely:
Latest log

[TestAcctlogRescUsedWithTwoMomHooks_after_fix2.txt](https://github.com/openpbs/openpbs/files/6087787/TestAcctlogRescUsedWithTwoMomHooks_after_fix2.txt)

bhroam (approved):

PR #2283, opened 2021-03-04
[WIP]Removing an additional job_save_db call for subjob's run path

agrawalravi90:
#### Describe Bug or Feature

When running a new subjob, we call job_save_db twice: once from inside req_runjob2->svr_startjob->..->svr_setjobstate, and once inside req_runjob2:
before

This is redundant, so we'll save some time if we only saved once if possible.

#### Describe Your Change

Moved the job_save_db call to after svr_setjobstate(). This causes svr_setjobstate() to skip calling job_save_db as it's a new subjob so that newobj bit is still set. So, we only save the subjob once.

Profile: regression
|--|--|--|--|--|--|--|
|6471|3990|0|0|0|0|3990|

### DIFF
--- a/src/server/req_runjob.c
+++ b/src/server/req_runjob.c
@@ -772,7 +772,6 @@ req_runjob2(struct batch_request *preq, job *pjob)
 		reply_ack(preq);
 		if (save_subjob(pjob)) {
 			free_nodes(pjob);
-			req_reject(PBSE_SAVE_ERR, 0, preq);
 			return;
 		}
 	}
--- a/src/server/req_runjob.c
+++ b/src/server/req_runjob.c
@@ -728,6 +728,16 @@ req_runjob(struct batch_request *preq)
 		reply_send(preq);
 	return;
 }
+
+static int
+save_subjob(job *pjob)
+{
+	if (pjob->ji_qs.ji_svrflags & JOB_SVFLG_SubJob)
+		return job_save_db(pjob);
+
+	return 0;
+}
+
 /**
  * @brief
  * 		req_runjob - service the Run Job and Asyc Run Job Requests
@@ -748,19 +758,8 @@ req_runjob2(struct batch_request *preq, job *pjob)
 	/* Check if prov is required, if so, reply_ack and let prov finish */
 	/* else follow normal flow */
 	prov_rc = check_and_provision_job(preq, pjob, &need_prov);
-
-	/* In case of subjob, save it to the database now because
-	 * not saved to the database so far.
-	 */
-	if (pjob->ji_qs.ji_svrflags & JOB_SVFLG_SubJob) {
-		if (job_save_db(pjob)) {
-			free_nodes(pjob);
-			req_reject(PBSE_SAVE_ERR, 0, preq);
-			return;
-		}
-	}
-
 	if (prov_rc) { /* problem with the request */
+		save_subjob(pjob);
 		free_nodes(pjob);
 		req_reject(prov_rc, 0, preq);
 		return;
@@ -771,7 +770,11 @@ req_runjob2(struct batch_request *preq, job *pjob)
 		/* provisioning was needed and was qneueued successfully */
 		/* Allways send ack for prov jobs, even if not async run */
 		reply_ack(preq);
-		return;
+		if (save_subjob(pjob)) {
+			free_nodes(pjob);
+			req_reject(PBSE_SAVE_ERR, 0, preq);
+			return;
+		}
 	}
 
 	/* if need_prov ==0 then no prov required, so continue normal flow */
@@ -784,6 +787,8 @@ req_runjob2(struct batch_request *preq, job *pjob)
 		}
 	}
 	if ((dest == NULL) || (*dest == '\0')) {
+		save_subjob(pjob);
+
 		/* Neither the run request nor the job specified an execvnode. */
 		free_nodes(pjob);
 		req_reject(PBSE_IVALREQ, 0, preq);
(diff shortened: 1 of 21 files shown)


src/server/req_runjob.c
```diff
@@ -771,7 +770,11 @@ req_runjob2(struct batch_request *preq, job *pjob)
 		/* provisioning was needed and was qneueued successfully */
 		/* Allways send ack for prov jobs, even if not async run */
 		reply_ack(preq);
-		return;
+		if (save_subjob(pjob)) {
+			free_nodes(pjob);
+			req_reject(PBSE_SAVE_ERR, 0, preq);
```
subhasisb, line 775: a req_reject() after a reply_ack() is not meaningful. I suggest we move the reply_ack() to after this block

src/server/req_runjob.c
```diff
@@ -748,19 +758,8 @@ req_runjob2(struct batch_request *preq, job *pjob)
...
-	if (pjob->ji_qs.ji_svrflags & JOB_SVFLG_SubJob) {
-		if (job_save_db(pjob)) {
-			free_nodes(pjob);
-			req_reject(PBSE_SAVE_ERR, 0, preq);
-			return;
-		}
-	}
-
 	if (prov_rc) { /* problem with the request */
+		save_subjob(pjob);
```
subhasisb, line 762: if you fail to save, you want to reject with the DB save error?

src/server/req_runjob.c
```diff
@@ -802,14 +807,20 @@ req_runjob2(struct batch_request *preq, job *pjob)
...
 		((rq_type == PBS_BATCH_AsyrunJob_ack) || preq)) {
+		save_subjob(pjob);
 		free_nodes(pjob);
-		if (preq)
-			req_reject(rc, 0, preq);
+		req_reject(rc, 0, preq);
+	}
+
+	/* if a subjob, the job wouldn't have gotten saved yet, let's save it now */
+	if (save_subjob(pjob)) {
```
subhasisb, line 821: u seem to have just shifted the location of the subjob save, how is it cutting a job save. Certainly the save inside of svr_setjobstate() is also still happening?

subhasisb (changes requested):

src/server/req_runjob.c
bhroam: Don't all of the reply/reject functions free the batch reply structure?  Won't doing 2 of them double free?

src/server/req_runjob.c
agrawalravi90: svr_setjobstate() doesn't save to db if newobj is set, which it is for the subjob. Calling job_save_db() before svr_setjobstate() clears that bit and causes svr_setjobstate() to call job_sabe_db again. That's how moving the job_save_db call after svr_setjobstate keeps the bit set and svr_setjobstate doesn't call job_save_db. I've added this explanation to the PR description now.

src/server/req_runjob.c
agrawalravi90: this was intentional, we are in an error case anyways, I think it's better to return the original error than return the db save error, that's why i didn't handle it. Let me know what you think.

src/server/req_runjob.c
subhasisb: Thanks for explaining, makes sense.

src/server/req_runjob.c
subhasisb: Sounds fine

src/server/req_runjob.c
subhasisb: Could you add a comment explaining why we are not saving the subjob before and how setjobstate() will not save it - somebody else will certainly forget this detail (i myself added the check for newjobstate in setjobstate() but did not recall that)

subhasisb (changes requested):

subhasisb (approved):

ashwathraop (approved):

subhasisb (approved):

subhasisb:
@agrawalravi90 if the server process was to crash right after this state change to 41, can you describe what would happen in that case? In other words, we need to test for that case (gdb would be handy here)

src/server/svr_jobfunc.c
```diff
@@ -506,7 +506,7 @@ svr_dequejob(job *pjob)
  */
 
 int
-svr_setjobstate(job *pjob, char newstate, int newsubstate)
+svr_setjobstate(job *pjob, char newstate, int newsubstate, bool db_write)
```
ashwathraop, line 509: Update Doxygen with the new argument.

agrawalravi90:
I have come across the following issue with this enhancement, for subjobs:
- if server process is killed before it's had a chance to save it to db (because of my change, the save to db happens after sending the job to mom), then the subjob is never saved to db
- So, when server wakes up, mom tells it about the subjob, but server doesn't have it (subjobs are created only when they are run), so it tells the mom to discard the job, and it ends up getting terminated

As far as I can tell, this is not trivial to fix. We'll need to recreate the subjob object and set all of the metadata on it that is set when a job is run. This means walking down some of the runjob path, but skipping the rest of it (since the array parent and nodes are already allocated). I'm still trying to figure out if this is worth the effort, in the meantime, if you guys have any thoughts/suggestions, do let me know. I'll investigate this for a week more at max and then decide whether to close this PR or not.

agrawalravi90:
Another thought: instead of this, one, possibly lower hanging fruit, might be to not send exec_vnode for jobs from moms to server when doing registermom. The server doesn't use it, so we could save some data transfer and not send this over. Of course this would mean that there's no way to do the fix needed for this PR as we'll no longer have an exec_vnode from the mom, but since that's a messy route anyways, maybe it's better to go this other route.

PR #2288, opened 2021-03-09
qrank of jobs can be same for high submit rates

agrawalravi90:
#### Describe Bug or Feature

queue_rank's precision is milliseconds, which means that jobs submitted within the same millisecond can have the same queue_rank, which would mean that scheduler won't be able to sort them properly, especially for multi-server where scheduler gets jobs from servers as: (all jobs from server 1) -> (all jobs from server 2) .. and so on

#### Describe Your Change

Changed the precision of queue_rank to microseconds

```
Thread 2 "pbs_sched" hit Breakpoint 1, query_job (job=<optimized out>, sinfo=0x2122d70, err=0x14a740000b60) at ../../../src/scheduler/job_info.cpp:1223
1223				count = strtol(attrp->value, &endp, 10);
(gdb) p attrp->value
$1 = 0x21235e0 "1615311926959806"
(gdb) n
1224				if (*endp == '\0')
(gdb) p count
$2 = 1615311926959806
(gdb) c
Continuing.
```


src/server/array_func.c
```diff
@@ -682,7 +682,7 @@ create_subjob(job *parent, char *newjid, int *rc)
 	svrattrl  *psatl;
 	job 	  *subj;
 	long	   eligibletime;
-	long	    time_msec;
+	long	    time_usec;
```
arungrover, line 685: I know I wrote time_msec but I think it was a mistake not making it an unsigned long. If you want, you can do that. Not that it matters because we will be all gone before it fills up :)

arungrover (approved):
LGTM!

vstumpf:
How big is a `long`? It needs to be 64 bits. 

At some point we should replace `long` with `int64_t`, but that's for another time.

bhroam:
Actually the ANSI standard defines a long as at least as large as an int.  It just happens that all logs these days are 64bits.

vstumpf:
> 
> 
> > How big is a `long`? It needs to be 64 bits.
> > At some point we should replace `long` with `int64_t`, but that's for another time.
> 
> It is 8 bytes.

I only ask because we do have both `long` and `long long` available to us in attr_val
```
union attr_val {	      /* the attribute value	*/
	long		      at_long;	/* long integer */
	Long		      at_ll;	/* largest long integer */
	...
}
```

agrawalravi90:
@vstumpf thanks, i didn't know that we have a long long type, ive changed qrank's type to long long, also added job level getters and setters for long long, please review the latest commit @bhroam and @arungrover

bhroam:
I don't think I agree with using a long long for qrank.  From what I understand about a long long, is the compiler knits two registers together to create a long long.  This adds extra work.  I can't imagine that qrank will ever exceed a long.  A long is 19 characters long and an unsigned long is 20.  Currently a time + usec is 16 characters long.  This means we have 3-4 orders of magnitude before we need to worry.

bhroam:
So the standard says a long is at least as wide as an int, and a long long is at least as wide as a long.  By this logic, a long long could be as long as an int.  In practice, an int is 32bits.  A long is 64 bits.  A long long is 128 bits.  Like Vincent said, we should probably be using the uint64 type so we can guarinty it to be 64 bits.  Of course that'd need to completely revamp the at_val union, so that should be done at a different time.

We ran into issues with a 32bit int using when we switched to miliseconds.  When we switched that to a long, it fixed the problem on all of our platforms.  I think we can safely say that all the platforms we build on have a 64bit long.

arungrover:
I didn't know long long could be 128 bits in size, if it is, it will cause problems with how we store ji_qrank as bigint in db.
I thought size of long was 4 bytes only on windows compilers, all other compilers have long as 8 byte long. Since we don't support server/scheduler on a windows config anyways, In my opinion, an unsigned long was just fine.

agrawalravi90:
@bhroam can you please provide reference to documentation that says that long long is 128 bits? Everywhere I saw, it seems to be listed as 8 bytes (or 64 bits), for both 32 and 64 bit architectures, which is why i think it's the right choice. long is 32 bits on 32 bit architectures, 64 bits on 64, so I think we should use long long to be on the safer side. Of course if it can be 128 bits, or 16 bytes, then it doesn't make sense to use it.

src/lib/Libattr/attr_fn_l.c
```diff
@@ -277,6 +277,44 @@ set_attr_l(attribute *pattr, long value, enum batch_op op)
...
+ * @param[in]	op		-	operation to do
+ *
+ * @return	void
+ *
+ * @par MT-Safe: No
+ * @par Side Effects: None
+ *
+ */
+void
+set_attr_ll(attribute *pattr, Long value, enum batch_op op)
```
nithinj, line 294: shouldn't this function be inside attr_fn_ll.c ?

subhasisb:
> Actually the ANSI standard defines a long as at least as large as an int. It just happens that all logs these days are 64bits.

Right. A 32 bit architecture, defined as ILP32 means Int, long, pointers are all 32 bits (4 bytes), whereas the 64 bit architecture is usually defined as LP64 (notice there is no i in that) - which means Long and pointer are 64 bits, whereas int remains 32 bits.

subhasisb:
> > > How big is a `long`? It needs to be 64 bits.
> > > At some point we should replace `long` with `int64_t`, but that's for another time.
> > 
> > 
> > It is 8 bytes.
> 
> I only ask because we do have both `long` and `long long` available to us in attr_val
> 
> ```
> union attr_val {	      /* the attribute value	*/
> 	long		      at_long;	/* long integer */
> 	Long		      at_ll;	/* largest long integer */
> 	...
> }
> ```

Well, that attr_val union has other members like struct pbsnode, and struct array_strings so it is probably going to much larger than even long long in size. Some day, we should refactor that code to not take that much space for a attribute value which is only an int or even char (but sadly, today it does)

arungrover:
You are right @subhasisb. I think only problem here is LLP64 systems that have long as big as an int (4bytes). Google search says that other than windows there isn't any that follow this convention.

subhasisb:
> I didn't know long long could be 128 bits in size, if it is, it will cause problems with how we store ji_qrank as bigint in db.
> I thought size of long was 4 bytes only on windows compilers, all other compilers have long as 8 byte long. Since we don't support server/scheduler on a windows config anyways, In my opinion, an unsigned long was just fine.

long long is 64 bits only (not 128 bits) - long long and "just" long differs only in the case of ILP32 architecture (32 bit machines) where "just" long is still 32 bits but long long is 64 bits. On a 64 bit machine, both long long and just long are the same.

src/lib/Libattr/attr_fn_l.c
subhasisb: yeah, this should be in the attr_fn_ll.c file

subhasisb (changes requested):
Another similar change sought:

I noticed that for some reason even though pointers and longs are now 64 bit, we've kept the at_float member of the union to store attribute/resource values as a C 'float', and when we encode (and e.g. send to the scheduler) we chop off anything after the first four digits (which creates trouble when people use these to express fractions).

Now that 64-bit machines are ubiquitous, isn't it time for us to store that in a double, and to set FLOA_NUM_DIGITS to 12 or 14 rather than 4?

... as far as I can see all that is needed is a one word header change and another for FLOAT_NUM_DIGITS. The decode is already using strtod, but we just throw away some of the bits.

bhroam:
I'm sorry for stirring the pot.  I must have been living in the past when on a 32bit architecture, the compiler had to stitch two registers together to make a 64bit number.  This does not happen today.  I'm fine with long long.  qrank has to be a 64bit number.  Back when we were using miliseconds, we overflowed every 48 days when we were accidently using an int.  Now that we're using microseconds, it'll be even worse.

arungrover:
I am ok using long long.

arungrover:
I am ok using long long.

agrawalravi90:
@bhroam  and @arungrover thanks! I've addressed Nithin's comment about moving a function, please provide further feedback.

bhroam (commented):
Why Long instead of long long?  You won't have to include a header file if you use long long.

agrawalravi90:
>Another similar change sought:

I noticed that for some reason even though pointers and longs are now 64 bit, we've kept the at_float member of the union to store attribute/resource values as a C 'float', and when we encode (and e.g. send to the scheduler) we chop off anything after the first four digits (which creates trouble when people use these to express fractions).

Now that 64-bit machines are ubiquitous, isn't it time for us to store that in a double, and to set FLOA_NUM_DIGITS to 12 or 14 rather than 4?

... as far as I can see all that is needed is a one word header change and another for FLOAT_NUM_DIGITS. The decode is already using strtod, but we just throw away some of the bits.

Ok, I've increased the precision of floats to 12, please review the latest commit

agrawalravi90:
> Why Long instead of long long? You won't have to include a header file if you use long long.

I followed the existing code. 'Long' is something we've created, I think to work on all platforms. Do you still think I should change it to long long?

bhroam:
@agrawalravi90 long long is part of the c99 standard which we now use.  I suspect Long was created before we were able to use the c99 standard.

agrawalravi90:
> So this macro has nothing to do with what we store. It has to do with what we print into our logs. In order to change this, you need to test if the FLOAT_DIGITS() macro will work properly with 12. When I originally wrote the macro, I found the longer you got, the more crap the double would have in it, and the macro would start printing inaccuracies that were not really there. This is mostly used by the scheduler which already uses a double. So don't just make this change without some testing of what it'll do. Plus do we really want 12 digits printed for all of our doubles? 4 probably is enough.

Thanks for pointing this out, I didn't realize that increasing it to 12 can add this complication. @subhasisb I think this change would be better done separately from the qrank change as it needs a dedicated discussion and vetting, and really has nothing much to do with qrank. So, I'll revert this for now.

agrawalravi90:
@bhroam thanks, I've addressed your comment, please provide further feedback.

@arungrover can you please review the latest changes as well?

vstumpf (approved):
Thanks for making the changes Ravi, I like the use of `long long`.

arungrover (approved):
LGTM!

agrawalravi90:
@arungrover  @vstumpf thanks for the sign-off guys.

@subhasisb please let me know if you have more feedback.

bhroam (approved):
LGTM

subhasisb (approved):

PR #2291, opened 2021-03-09
Refactor queue_info, config, and status into C++

bhroam:
#### Describe Bug or Feature
Due to some forthcoming multiserver changes, the queue_info structure needed to be refactored into C++.  That led to the config/status also being refactored.

#### Describe Your Change
The queue_info structure (now class), has a constructor, destructor, and copy constructor now.
The config structure no longer has any memory directly allocated for it.  All of it's pointer arrays have been converted to vectors.  The status structure was similarly modified.

Description: Rerun All Tests From #6387
Profile: regression
|--|--|--|--|--|--|--|
|6395|3990|2|0|0|0|3988|

I reran the 2 failed tests and they passed.


src/scheduler/parse.cpp
```diff
@@ -344,415 +385,278 @@ parse_config(const char *fname)
...
+					free_string_array(strarr);
+
+					if (need_host)
+						tmpconf.res_to_check.emplace_back("host");
+					if (need_vnode)
+						tmpconf.res_to_check.emplace_back("vnode");
+				} else if (!strcmp(config_name, PARSE_DEDICATED_PREFIX))
+					tmpconf.ded_prefix = config_value;
+				else if (!strcmp(config_name, PARSE_PRIMETIME_PREFIX))
+						tmpconf.pt_prefix = config_value;
```
vstumpf, line 434: extra indentation here

src/scheduler/parse.cpp
```diff
@@ -1093,43 +861,6 @@ preempt_cmp(const void *p1, const void *p2)
...
-		for (i = 0; i < MAX_SORTS; i++) {
-			if (so_info[i].res_name != NULL)
-				free(so_info[i].res_name);
-		}
-		free(so_info);
-	}
-}
+peer_queue::peer_queue(const char *lqueue, const char *rqueue, const char *rserver): local_queue(lqueue), remote_queue(rqueue), remote_server(rserver) {}
+timegap::timegap(time_t from, time_t to): from(from), to(to) {}
+dyn_res::dyn_res(const char *res, const char *cmdline, const char *fname): res(res), command_line(cmdline), script_name(fname) {}
```
vstumpf, line 866: if you're not going to define them in the header file, i would put `{` and `}` on their own lines.

vstumpf (changes requested):
quick comments

src/scheduler/job_info.cpp
```diff
@@ -2473,42 +2475,28 @@ resresv_set_which_selspec(resource_resv *resresv)
...
  * @retval NULL on error
  */
-resdef **
-create_resresv_sets_resdef(status *policy, server_info *sinfo) {
-	resdef **defs;
-	int def_ct;
-	int limres_ct = 0;
-	int i;
+std::vector<resdef *>
+create_resresv_sets_resdef(status *policy) {
```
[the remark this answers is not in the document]
bhroam: This isn't the only function we pass a status *policy into.  If I change it here, I'll have to change it everywhere.  Also, sinfo->policy is a pointer, so I'd have to pass *sinfo->policy into the function.  Might as well leave it as a pointer.

src/scheduler/data_types.h
```diff
@@ -325,35 +328,32 @@ struct status
...
+	std::vector<resdef *> resdef_to_check_rassn;		/* resdef_to_check intersects res_rassn */
+	std::vector<resdef *> resdef_to_check_rassn_select;	/* resdef_to_check intersects res_rassn and host level resource */
+	std::vector<resdef *> resdef_to_check_noncons;	/* non-consumable resources to match */
+	std::vector<resdef *> equiv_class_resdef;		/* resources to consider for job equiv classes */
 
 
 	time_t prime_status_end;		/* the end of prime or nonprime */
 
-	resdef **rel_on_susp;	    /* resources to release on suspend */
+	std::vector<resdef *> rel_on_susp;	    /* resources to release on suspend */
```
agrawalravi90, line 349: this as well

src/scheduler/data_types.h
```diff
@@ -1126,54 +1133,50 @@ struct config
...
 	bool prime_exempt_anytime_queues:1; /* backfill affects anytime queues */
-	bool assign_ssinodes:1;	/* assign the ssinodes resource */
 	bool preempt_starving:1;	/* once jobs become starving, it can preempt */
 	bool preempt_fairshare:1; /* normal jobs can preempt over usage jobs */
 	bool dont_preempt_starving:1; /* don't preempt staving jobs */
 	bool enforce_no_shares:1;	/* jobs with 0 shares don't run */
 	bool node_sort_unused:1;	/* node sorting by unused/assigned is used */
-	bool resv_conf_ignore:1;  /* if we want to ignore dedicated time when confirming reservations.  Move to enum if ever expanded */
-	bool allow_aoe_calendar:1;        /* allow jobs requesting aoe in calendar*/
+	bool resv_conf_ignore:1;	/* if we want to ignore dedicated time when confirming reservations.  Move to enum if ever expanded */
```
agrawalravi90, line 1141: do we need to the ":1" with bool?

agrawalravi90 (changes requested):
didn't intend to review the whole PR, just reviewed some data structure stuff

src/scheduler/data_types.h
bhroam: Yes, otherwise the size is undefined (likely 1 byte per bool)

arungrover (changes requested):
Minor comments

src/scheduler/check.cpp
```diff
@@ -1143,128 +1263,89 @@ check_avail_resources(schd_resource *reslist, resource_req *reqlist,
...
-					}
-				}
-			}
-			if(fail && (flags & RETURN_ALL_ERR)) {
-				fail = 0;
-				any_fail = 1;
+	return num_chunk;
+}
+
+// overloaded check_avail_resources() that checks all resources
```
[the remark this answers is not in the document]
bhroam: The arguments to both functions are the same (except the missing one).  I figured it would be redundant.  I can write one if you want.

src/scheduler/check.cpp
```diff
@@ -1076,6 +1075,146 @@ is_ok_to_run(status *policy, server_info *sinfo,
...
+ * @param[in] fail_code - fail code to use in schd_error if resources don't match
+ * @param[out] err - if resources don't match, reason not matched
+ * @return int
+ * @retval number of chunks matched if matched and consumable
+ * @retval SCHD_INFINITY if matched and non-consumable
+ * @retval 0 of resources failed to match
+ */
+
+int
+do_compare(schd_resource *res, resource_req *resreq, unsigned int flags, enum sched_error_code fail_code, schd_error *err)
```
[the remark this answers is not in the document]
bhroam: This compares a schd_resource to a resource_req.  Having one name without the other wouldn't work.  I can try and come up with a better name.

src/scheduler/buckets.cpp
```diff
@@ -784,7 +785,8 @@ chunk_to_nspec(status *policy, chunk *chk, node_info *node, char *aoename)
 		}
 	}
 	for (cur_req = chk->req; cur_req != NULL; cur_req = cur_req->next) {
-		if (resdef_exists_in_array(policy->resdef_to_check, cur_req->def) && cur_req->def->type.is_consumable) {
+		const auto& rdc = policy->resdef_to_check;
+		if (std::find(rdc.begin(), rdc.end(), cur_req->def) != rdc.end() && cur_req->def->type.is_consumable) {
```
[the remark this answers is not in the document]
bhroam: I'm not sure I understand.  When we get to this function, we've already matched the bucket to the job.  We know the bucket matches.  This function creates the nspec.  nspec's only contain consumable resources and aoe.  This line says, if the resource is in the resources line and it is consumable, add it to the nspec.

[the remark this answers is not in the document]
arungrover: My comment was to do find() only when we know the resource is consumable.

src/scheduler/check.cpp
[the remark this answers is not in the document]
agrawalravi90: I think it would be helpful, otherwise the doxygen documentation will not report overloaded functions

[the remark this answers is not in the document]
arungrover: Really? I thought doxygen will report the function but it wouldn't display anything else because there is no header.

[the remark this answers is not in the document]
agrawalravi90: hmm, ya you are probably right, so it will report the function name, but not describe the arguments .. anyways, I'm not particular about this, I'll leave it to you to decide.

src/scheduler/buckets.cpp
```diff
@@ -37,6 +37,7 @@
  * subject to Altair's trademark licensing policies.
  */
 
+#include <algorithm>
```
agrawalravi90, line 40: just curious, what exactly is this needed for?

src/scheduler/check.h
```diff
@@ -41,6 +41,8 @@
 #ifndef	_CHECK_H
 #define	_CHECK_H
 
+#include <vector>
```
agrawalravi90, line 44: is this still needed?

agrawalravi90: don't you need to include unordered_set instead?

src/scheduler/data_types.h
```diff
@@ -1126,54 +1133,50 @@ struct config
...
 	enum smp_cluster_dist non_prime_smp_dist;/* how do dist jobs during nonprime*/
 	time_t prime_spill;			/* the amount of time a job can
 						 * spill into primetime
 						 */
 	time_t nonprime_spill;			/* vice versa for prime_spill */
-	fairshare_head *fairshare;		/* fairshare tree */
 	time_t decay_time;			/*  time in seconds for the decay period*/
-	struct t prime[HIGH_DAY][2];	/* prime time start and prime time end*/
-	int holidays[MAX_HOLIDAY_SIZE];		/* holidays in Julian date */
+	struct t prime[HIGH_DAY][2];		/* prime time start and prime time end*/
```
agrawalravi90, line 1158: for future, i think we should start using std::array for static arrays, it provides similar advantages as std::string vs char array, and comes with a useful set of built-in functions.

src/scheduler/dedtime.cpp
```diff
@@ -88,16 +92,16 @@ parse_ded_file(const char *filename)
...
 		sprintf(log_buffer, "Error opening file %s", filename);
 		log_err(errno, "parse_ded_file", log_buffer);
 		return 1;
 	}
 
-	i = 0;
-	memset(conf.ded_time, 0, MAX_DEDTIME_SIZE * sizeof(struct timegap));
+	// We are rereading the dedtime file.  The current dedtime might not exist any more.
+	cstat.is_ded_time = false;
+
```
agrawalravi90, line 104: should we clear the vector ded_time here?

agrawalravi90 (changes requested):
minor comments/questions .. btw, the code looks so much better!

src/scheduler/buckets.cpp
bhroam: Oh, it's for std::find().  I should remove it from a number of files.

src/scheduler/check.cpp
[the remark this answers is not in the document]
bhroam: I'll add a @brief to the overload and point it to look at the main function for the arguments.

bhroam:
Latest TH3 results:
Description: Rerun All Tests From #6462
Profile: regression
|--|--|--|--|--|--|--|
|6467|3990|0|0|0|0|3990|

agrawalravi90 (approved):
LGTM

src/scheduler/buckets.cpp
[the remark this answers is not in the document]
bhroam: Even with a constant time find of the unordered_set, I think your comment is still valid.  Checking !bit is still going to be faster than doing a constant time hash into the set.

bhroam:
@arungrover @vstumpf I fixed your comments.  Gentle reminder to rereview.

arungrover (approved):

vstumpf (approved):
LGTM

PR #2293, opened 2021-03-10
test_invoke_pbs_tmrsh_from_sister_mom fails with empty output file

saimanikant:
#### Describe Bug or Feature

TestJobTask.test_invoke_pbs_tmrsh_from_sister_mom is failing ,generating error file and empty output file.

test is failing with below error:

  File "/home/pbsroot/TEST/tmp/tests/functional/pbs_job_task.py", line 142, in test_invoke_pbs_tmrsh_from_sister_mom
    self.assertEqual(ret['out'][0], mom3, "pbs_tmrsh invoked from sister"

error file has below content:
[pbsroot@x05-c8p1 functional]$ *sudo -H -u pbsuser /opt/tools/wrappers/cat /home/pbsuser/PtlPbsJobScriptpwnfznj5.e2
/opt/pbs/bin/pbs_tmrsh: host "x02-s15" is not a node in job <2.x05-c8p1>*

#### Describe Your Change

The test scripts need to use FQDNs.

Logs for failed run:
[logfile_ERROR.txt](https://github.com/openpbs/openpbs/files/6116025/logfile_ERROR.txt)

Logs for pass run:
[test_invoke_pbs_tmrsh_from_sister_mom_pass.txt](https://github.com/openpbs/openpbs/files/6116026/test_invoke_pbs_tmrsh_from_sister_mom_pass.txt)

### DIFF
--- a/test/tests/functional/pbs_job_task.py
+++ b/test/tests/functional/pbs_job_task.py
@@ -117,15 +117,17 @@ class TestJobTask(TestFunctional):
         mom1 = self.moms.keys()[0]
         mom2 = self.moms.keys()[1]
         mom3 = self.moms.keys()[2]
+        fqdn_mom2 = socket.getfqdn(mom2)
+        fqdn_mom3 = socket.getfqdn(mom3)
         pbstmrsh_cmd = os.path.join(self.server.pbs_conf['PBS_EXEC'],
                                     'bin', 'pbs_tmrsh')
 
         script_mom2 = """#!/bin/bash\n%s %s hostname""" % \
-                      (pbstmrsh_cmd, mom3)
+                      (pbstmrsh_cmd, fqdn_mom3)
         fn = self.du.create_temp_file(hostname=mom2, body=script_mom2)
         self.du.chmod(hostname=mom2, path=fn, mode=0o755)
         a = {ATTR_S: '/bin/bash'}
-        script = ['%s %s %s' % (pbstmrsh_cmd, mom2, fn)]
+        script = ['%s %s %s' % (pbstmrsh_cmd, fqdn_mom2, fn)]
         job = Job(TEST_USER, attrs=a)
         job.set_attributes({'Resource_List.select': '3:ncpus=1',
                             'Resource_List.place': 'scatter'})


bayucan (approved):
Looks good.

riyazhakki (approved):

test/tests/functional/pbs_job_task.py
```diff
@@ -117,15 +117,17 @@ def test_invoke_pbs_tmrsh_from_sister_mom(self):
         mom1 = self.moms.keys()[0]
         mom2 = self.moms.keys()[1]
         mom3 = self.moms.keys()[2]
+        fqdn_mom2 = socket.getfqdn(mom2)
+        fqdn_mom3 = socket.getfqdn(mom3)
```
prakashcv13, line 121: what if getfqdn fails?

saimanikant: why would getfqdn fail? if we run with wrong hostname values to -p argument? 
PTL would catch that in setup() before starting the test.

hirenvadalia: PTL setUp() doesn't use getfqdn(), see PBSService.\_\_init__() (https://github.com/openpbs/openpbs/blob/master/test/fw/ptl/lib/ptl_service.py#L509)

So, you can use self.moms.values()[0].hostname (or .fqdn) to get FQDN of that mom

saimanikant: Thanks @hirenvadalia. I would still like to understand why getfqdn() could cause a problem

hirenvadalia: @saimanikant 
I didn't say getfqdn() will cause a problem, my point is whatever you are trying to do is already there, so use it.

@prakashcv13
getfqdn will never fail, it will try to find fqdn if available, if not then it will return hostname returned by socket.gethostname() (which can be fqdn or short name, based on system configuration)

prakashcv13: @hirenvadalia and @saimanikant - I should have been clear - by 'fail' i mean what if getfqdn() returns the short name itself. what should the test case do in that case.

saimanikant: @prakashcv13, My understanding is that pbs_tmrsh rejects with usage of short name and we expect pbs_tmrsh to work with hostnames listed in $PBS_NODEFILE and $PBS_NODEFILE shows Fqdn names. 
I don't think the fqdn names in $PBS_NODEFILE will be different from the values returned by getfqdn()?

prakashcv13: I am not of the view that we can be absolutely sure about both being same, I would recommend skipping the test case or failing purposely if getfqdn returns the short name. what do you think?

saimanikant: I don't think that is necessary. I think the value returned by getfqdn() would be the same value written in $PBS_NODEFILE by PBS.

prakashcv13: alright. do we have logs supporting your statement. If not, let us err on the side of caution and add the skip/fail code.

saimanikant: The logs for the test run with changes are attached in PR description, where pbs_tmrsh runs as expected.

prakashcv13: well, I am not sure how can I be more clear about this. I would go out on a limb and say that we add the code to check if getfqdn() returns a value equal to the short name and skip/fail the test if needed.

prakashcv13 (changes requested):

saimanikant:
Hi @bayucan . I think this PR is ready for merge. Please have a look.

bayucan:
@saimanikant : I'm not sure if you've fully addressed Hiren's and Prakash's comments. @hirenvadalia @prakashcv13 : are you okay with the code as it stands?

prakashcv13:
hey @bayucan - I am fine with the changes.

bayucan:
Thanks @prakashcv13 . How about you @hirenvadalia , are you okay with the changes?

hirenvadalia (approved):

PR #2295, opened 2021-03-10
Changed PTL to use pbs_sleep instead of /bin/sleep

arungrover:
#### Describe Bug or Feature
Some of the preemption PTL tests have been failing because ptl framework by default submits a /bin/sleep job. Since sleep process cannot really be suspended, it turns out that in some tests job finishes before it can be resumed back.

#### Describe Your Change
Changed the PTL framework to use pbs_sleep binary for all jobs as much as possible. 

#### Link to Design Doc
NA

Profile: regression
|--|--|--|--|--|--|--|
|6419|3990|1|0|0|8|3981|

Test that failed is a known failure.
[test_smoke.txt](https://github.com/openpbs/openpbs/files/6119168/test_smoke.txt)

### DIFF
--- a/test/fw/ptl/lib/ptl_mom.py
+++ b/test/fw/ptl/lib/ptl_mom.py
@@ -165,7 +165,10 @@ class MoM(PBSService):
         # This is true by default, but can be set to False if
         # required by a test
         self.revert_to_default = True
-        self.sleep_cmd = '/bin/sleep'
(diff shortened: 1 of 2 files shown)


bhroam (approved):
LGTM

birbalsain (approved):
Changes LGTM

PR #2299, opened 2021-03-12
preventing child hang from inherited nslookup lock from parent process

vishwaks:
#### Describe Bug or Feature
The periodic hook which used to make http call(will do getaddrinfo sys call), is getting hung permanently even after alarm(timeout) call. 
One possible reason of this hang could be that when the hook process was forked, tpp thread was in getaddrinfo call. Since this library call takes a lock, it is possible that after fork, child inherited this lock but since child was forked by a non-tpp thread it can never release lock.

#### Describe Your Change
Only solution to this problem is to take mutex lock in tpp thread before calling getaddrinfo() and register pthread_atfork handlers to take/release these locks in pre/post handlers respectively.

Id: 6455
Description: Rerun All Tests From #6433(without valgrind)
Profile: regression

Test Summary: total: 3990, fail: 1, error: 0, timedout: 0, skip: 0, pass: 3989

### DIFF
--- a/src/lib/Libtpp/tpp_internal.h
+++ b/src/lib/Libtpp/tpp_internal.h
@@ -508,6 +508,17 @@ int tpp_add_fd(int, int, int);
 int tpp_del_fd(int, int);
 int tpp_mod_fd(int, int, int);
 
+/* a new mutex introduced to prevent inheriting lock from tpp thread
+ * from getaddrinfo(nslookup) during fork for periodic hook
+ * set handlers using pthread_atfork.
+ */
+pthread_mutex_t tpp_nslookup_mutex;
+int tpp_nslookup_mutex_lock();
+int tpp_nslookup_mutex_unlock();
+void tpp_nslookup_atfork_prepare();
+void tpp_nslookup_atfork_parent();
+void tpp_nslookup_atfork_child();
+
 int tpp_validate_hdr(int, char *);
 tpp_addr_t *tpp_get_addresses(char *, int *);
 tpp_addr_t *tpp_get_local_host(int);
--- a/src/lib/Libtpp/tpp_internal.h
+++ b/src/lib/Libtpp/tpp_internal.h
@@ -513,8 +513,6 @@ int tpp_mod_fd(int, int, int);
  * set handlers using pthread_atfork.
  */
 pthread_mutex_t tpp_nslookup_mutex;
-int tpp_nslookup_mutex_lock();
-int tpp_nslookup_mutex_unlock();
 void tpp_nslookup_atfork_prepare();
 void tpp_nslookup_atfork_parent();
 void tpp_nslookup_atfork_child();
--- a/src/lib/Libtpp/tpp_platform.c
+++ b/src/lib/Libtpp/tpp_platform.c
@@ -710,10 +710,15 @@ tpp_sock_resolve_host(char *host, int *count)
 	hints.ai_socktype = SOCK_STREAM;
 	hints.ai_protocol = IPPROTO_TCP;
 
+	tpp_nslookup_mutex_lock();
+
 	if ((rc = getaddrinfo(host, NULL, &hints, &pai)) != 0) {
+		tpp_nslookup_mutex_unlock();
 		tpp_log(LOG_CRIT, NULL, "Error %d resolving %s", rc, host);
 		return NULL;
 	}
+	/* unlock nslook up mutex */
+	tpp_nslookup_mutex_unlock();
 
 	*count = 0;
 	for (aip = pai; aip != NULL; aip = aip->ai_next) {
--- a/src/lib/Libtpp/tpp_platform.c
+++ b/src/lib/Libtpp/tpp_platform.c
@@ -710,8 +710,10 @@ tpp_sock_resolve_host(char *host, int *count)
 	hints.ai_socktype = SOCK_STREAM;
 	hints.ai_protocol = IPPROTO_TCP;
 
-	/* introducing a new mutex to prevent child process from 
-	inheriting getaddrinfo mutex using pthread_atfork handlers*/
+	/* 
+	 * introducing a new mutex to prevent child process from 
+	 * inheriting getaddrinfo mutex using pthread_atfork handlers
+	 */
 	tpp_lock(&tpp_nslookup_mutex);
 
 	if ((rc = getaddrinfo(host, NULL, &hints, &pai)) != 0) {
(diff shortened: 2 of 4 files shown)


src/lib/Libtpp/tpp_transport.c
```diff
@@ -505,6 +505,19 @@ tpp_transport_init(struct tpp_config *conf)
...
+	if (pthread_mutex_init(&tpp_nslookup_mutex, NULL) != 0) {
+		tpp_log(LOG_CRIT, __func__, "Failed to initialize nslookup mutex");
+		return -1;
+	}
+	
+#ifndef WIN32
+	/* for unix, set a pthread_atfork handler */
+	if (pthread_atfork(tpp_nslookup_atfork_prepare, tpp_nslookup_atfork_parent, tpp_nslookup_atfork_child) != 0) {
+		fprintf(stderr, "tpp nslookup mutex atfork handler failed\n");
+		exit(1);
```
subhasisb, line 518: Is there any need to exit here? We can simply return -1, like any other failures

subhasisb: also why the fprintf, you can simply use tpp_log() like the mutex_init function case

src/lib/Libtpp/tpp_util.c
```diff
@@ -2332,6 +2332,89 @@ tpp_set_logmask(long logmask)
...
+ * @retval  0 - success
+ *
+ * @par Side Effects:
+ *	None
+ *
+ * @par MT-safe: Yes
+ *
+ */
+int
+tpp_nslookup_mutex_unlock()
```
subhasisb, line 2374: you dont need these new routines, you can simply use tpp_lock() and pass the tpp_nslookup_mutex address to it.

src/lib/Libtpp/tpp_util.c
```diff
@@ -2332,6 +2332,89 @@ tpp_set_logmask(long logmask)
...
+ * @retval  0 - success
+ *
+ * @par Side Effects:
+ *	None
+ *
+ * @par MT-safe: Yes
+ *
+ */
+int
+tpp_nslookup_mutex_lock()
```
subhasisb, line 2350: you don't need these new routines, you can simply use tpp_lock() and pass the tpp_nslookup_mutex address to it.

src/lib/Libtpp/tpp_util.c
```diff
@@ -2332,6 +2332,89 @@ tpp_set_logmask(long logmask)
...
+
+/**
+ * @brief
+ *	wrapper function for tpp_nslookup_mutex_lock().
+ *
+ */
+void
+tpp_nslookup_atfork_prepare()
+{
+	tpp_nslookup_mutex_lock();
```
subhasisb, line 2393: Change this to call tpp_lock(&tpp_nslookup_mutex);

src/lib/Libtpp/tpp_util.c
```diff
@@ -2332,6 +2332,89 @@ tpp_set_logmask(long logmask)
...
+
+/**
+ * @brief
+ *	wrapper function for tpp_nslookup_mutex_unlock().
+ *
+ */
+void
+tpp_nslookup_atfork_parent()
+{
+	tpp_nslookup_mutex_unlock();
```
subhasisb, line 2404: Change this to call tpp_unlock(&tpp_nslookup_mutex);

src/lib/Libtpp/tpp_util.c
```diff
@@ -2332,6 +2332,89 @@ tpp_set_logmask(long logmask)
...
+
+/**
+ * @brief
+ *	wrapper function for tpp_nslookup_mutex_unlock().
+ *
+ */
+void
+tpp_nslookup_atfork_child()
+{
+	tpp_nslookup_mutex_unlock();
```
subhasisb, line 2415: Change this to call tpp_unlock(&tpp_nslookup_mutex);

src/lib/Libtpp/tpp_internal.h
```diff
@@ -508,6 +508,17 @@ int tpp_add_fd(int, int, int);
 int tpp_del_fd(int, int);
 int tpp_mod_fd(int, int, int);
 
+/* a new mutex introduced to prevent inheriting lock from tpp thread
+ * from getaddrinfo(nslookup) during fork for periodic hook
+ * set handlers using pthread_atfork.
+ */
+pthread_mutex_t tpp_nslookup_mutex;
+int tpp_nslookup_mutex_lock();
```
subhasisb, line 516: u can use the tpp_lock() tpp_unlock() routines, no need to add another pair that does exactly the same

src/lib/Libtpp/tpp_platform.c
```diff
@@ -710,10 +710,15 @@ tpp_sock_resolve_host(char *host, int *count)
 	hints.ai_socktype = SOCK_STREAM;
 	hints.ai_protocol = IPPROTO_TCP;
 
+	tpp_nslookup_mutex_lock();
```
subhasisb, line 713: Change this to call tpp_lock(&tpp_nslookup_mutex);

src/lib/Libtpp/tpp_platform.c
```diff
@@ -710,10 +710,15 @@ tpp_sock_resolve_host(char *host, int *count)
...
+
 	if ((rc = getaddrinfo(host, NULL, &hints, &pai)) != 0) {
+		tpp_nslookup_mutex_unlock();
 		tpp_log(LOG_CRIT, NULL, "Error %d resolving %s", rc, host);
 		return NULL;
 	}
+	/* unlock nslook up mutex */
+	tpp_nslookup_mutex_unlock();
```
subhasisb, line 721: Change this to call tpp_unlock(&tpp_nslookup_mutex);

subhasisb (changes requested):

src/lib/Libtpp/tpp_platform.c
hirenvadalia: Since as per document of `getaddrinfo()`, its thread safe. I would suggest to add comment here to explain why we are adding lock/unlock around `getaddrinfo()` so in future it won't confuse ppl/dev.

subhasisb (approved):

src/lib/Libtpp/tpp_platform.c
```diff
@@ -710,10 +710,17 @@ tpp_sock_resolve_host(char *host, int *count)
 	hints.ai_socktype = SOCK_STREAM;
 	hints.ai_protocol = IPPROTO_TCP;
 
+	/* introducing a new mutex to prevent child process from
```
hirenvadalia, line 713: Please use correct multiline comment format as below:
```
/*
 * ...
 */
```

hirenvadalia (approved):

PR #2302, opened 2021-03-12
updated std_dev, minimum and maximum of average_measurements in json report

birbalsain:
#### Describe Bug or Feature
For test run multiple times  minimum, maximum and  std_dev are average of the all run count.

#### Describe Your Change
In average measurement, for minimum and maximum taking minimum, maximum  among all the run count.
taking standard deviation of mean. 

[average_measurement.txt](https://github.com/openpbs/openpbs/files/6132812/average_measurement.txt)

### DIFF
--- a/test/fw/ptl/utils/plugins/ptl_report_json.py
+++ b/test/fw/ptl/utils/plugins/ptl_report_json.py
@@ -212,6 +212,8 @@ class PTLJsonData(object):
                 m_avg['testsuites'][tsname]['testcases'][tcname] = []
                 t_sum = []
                 sum_std = []
+                sum_min = []
+                sum_max = []
                 count = 0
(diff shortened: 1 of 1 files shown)


visheshh (changes requested):
You can use the same kind of list creation and use min() and max() for values.

birbalsain:
> You can use the same kind of list creation and use min() and max() for values.

Already one for loop is in use if we use that same loop to compare  then it is good. If we use that loop only to store value in list and pass that list to min() then it is not good. In this case min() is not good.

birbalsain:
[average_measurements.txt](https://github.com/openpbs/openpbs/files/6147510/average_measurements.txt)

visheshh (approved):
LGTM

prakashcv13 (approved):

PR #2304, opened 2021-03-15
multi-server: Multi-node jobs crossing local server boundaries

nithinj:
#### Describe Bug or Feature
The scheduler can assign nodes reporting to different server instances to a multi-node job.

#### Describe Your Change

Server1 will send an update to book the nodes that belong to server2.

For instance, an INCR operation will be sent before running a job. A DECR will be the counter operation to decrement the resources accounted as in the case with job termination.

The update will be broadcasted using PSI values

Inter-server resource update can happen when:

→ a job gets run

→ job gets terminated

→ various other scenarios where the resource needs to be updated.

Upon receiving the request, server2 will update the job portion in the node object based on the operation. Server1 will be broadcasting the request to all the sister servers. The request will contain the following parameters:

→ jobid

→ execvnode

→ operation (INCR/DECR)

→ Job exclusivity

Challenges:

 *   Server does not know hostname and port of all the nodes where the job is being run:

when J1's run request lands on server1, it receives the execvnode which contains the list of all vnodes and how much resources is allotted in each of them. Server1 now has to translate it into exechost2, which will be hostname and port pairs instead of vnode name. This is required so that Mom1 can reach out to the other vnodes in the execvnode using the corresponding hostname:port.

As nodes are partitioned, the servers do not know the vnode to hostname:port mapping of nodes belongs to a different server.

To handle this, the server will do a node_stat() and cache the necessary information. The stat will happen only on failure to find the node either in server or join-job failure in mom (lazy pull). This will be further optimized with diff-stat.

  *  IS_CLUSTERADDR protocol cannot be reliably used for multi-server:

Even though this mechanism is not supported in multi-server mode, an attacker cannot gain control of the moms in the cluster if it is protected by any of the other supported authentication mechanisms such as TLS, munge or Kerberos.

 *   What if the resource update fails due to network failure?

Resource updates can fail due to a number of reasons including network failures and a down instance. This feature relies on the network layer to tell the application that something went wrong. If the application closes that connection channel, it will reattempt to send the resource update upon connection restoration.

 *   What happens when a server instance restarts?

Resource usage values of nodes, queue, server etc are repopulated based on the state of the job when a server is started. With multi-server, the shared objects might have some share from the job that belongs to another server. So server needs to request other servers to send any update it needs to account for. This will be done as part of the initial hello exchange.

 *   What if the scheduler does another stat before resource update reaches the sister-servers?

 This could result in an over-subscription as those nodes which need to be filled by the resource update might appear to be empty.

This is solved by pbs_server_ready which is a synchronization mechanism used by clients. pbs_server_ready is an IFL that makes sure all the servers in the cluster are ready for stat operation.

Detailed test steps and log and various scenarios which need to be automated:
[test.log](https://github.com/openpbs/openpbs/files/6127975/test.log)

Regression Results:

Description: Rerun All Tests From #6324
Platforms: UBUNTU1804,SUSE15,UBUNTU1604,CENTOS8,CENTOS7,RHEL7,SLES12,SLES15,WIN2K16
Profile: regression
|--|--|--|--|--|--|--|
|6325|3990|2|0|0|0|3988|

Description: Rerun Only Failed Tests From #6325
Platforms: UBUNTU1804,SUSE15,UBUNTU1604,CENTOS8,CENTOS7,RHEL7,SLES12,SLES15,WIN2K16
Profile: regression
|--|--|--|--|--|--|--|
|6342|2|0|0|0|0|2|

Valgrind logs:
[valgrind.log](https://github.com/openpbs/openpbs/files/6107397/valgrind.log)

Continuation of PR: https://github.com/openpbs/openpbs/pull/2253


subhasisb (changes requested):

subhasisb (approved):

src/server/multi_svr.c
```diff
@@ -292,11 +292,11 @@ log_msvr_stat()
 
 	log_eventf(PBSEVENT_DEBUG, PBS_EVENTCLASS_SERVER, LOG_DEBUG, __func__,
 		   "Average msvr statistical info for last 24 hours\n"
-		   "CACHE_MISS = %ld\n"
-		   "CACHE_REFR_TM = %ld\n"
-		   "NUM_RESC_UPDATE = %ld\n"
-		   "NUM_MOVE_RUN = %ld\n"
-		   "NUM_SCHED_MISS = %ld\n",
+		   "{\n\"CACHE_MISS\" : %ld,\n"
+		   "\"CACHE_REFRESH_TM\" : %ld,\n"
```
suresh-thelkar, line 296: No need to give new lines after every entry in JSON object. If we want to do that then it is better we put a tab to every line at the start for better readability.

suresh-thelkar (approved):
Code changes still look good to me. I just have a minor comment. Please take a look. Also implement the suggestions given by Subhasis.

nithinj:
@subhasisb & @suresh-thelkar ,

PR #2305, opened 2021-03-15
Fix test_revert_attributes of TestExpect

sakshamgarg:
#### Describe Bug or Feature

test_revert_attributes test was failing due to multiple reasons:

1. After unsetting the _sched_host_ attribute of scheduler, it was expecting that attribute to be not present. But the sched_host attribute was still there.
2. It was giving an error saying "Undefined attribute" after trying to create a scheduler partition.
3. Lastly it was throwing an error that the server object didn't have a definition for the "create_vnodes" function.

#### Describe Your Change

Following are the changes for the above described failures
1. Since sched_host attribute is set to server's hostname while unsetting it, I added "sched_host" attribute to special attribute list. This would take the default value of sched_host while unsetting it in PTL
2. As "sched_port" attribute has been removed from the scheduler, we don't need to create sched partition using this.
3. "create_vnodes" function was moved to MoM class from server class. So now calling the function using correct object.

Before the change:
Description: Tests from given build on platforms UBUNTU1804, SUSE15, UBUNTU1604, CENTOS8, RHEL8, CENTOS7, RHEL7, SLES15
Platforms: UBUNTU1804,SUSE15,UBUNTU1604,CENTOS8,RHEL8,CENTOS7,RHEL7,SLES15
Profile: regression
|--|--|--|--|--|--|--|
|6504|1|1|0|0|0|0|

After the change:
Profile: regression
|--|--|--|--|--|--|--|
|6508|1|0|0|0|0|1|

### DIFF
--- a/test/fw/ptl/lib/ptl_wrappers.py
+++ b/test/fw/ptl/lib/ptl_wrappers.py
@@ -131,6 +131,7 @@ class Wrappers(PBSService):
                            SCHED: [ATTR_sched_cycle_len, ATTR_scheduling,
                                    ATTR_schedit, ATTR_logevents,
                                    ATTR_sched_server_dyn_res_alarm,
+                                   ATTR_SchedHost,
                                    'preempt_prio', 'preempt_queue_prio',
                                    'throughput_mode', 'job_run_wait',
                                    'partition', 'sched_priv', 'sched_log'],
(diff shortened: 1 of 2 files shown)


raviv1995 (approved):

birbalsain (approved):
Changes LGTM

birbalsain:
@sakshamgarg Can you change the bug description as per offline discussion.

bhroam (approved):
LGTM

PR #2310, opened 2021-03-17
Short execvnode_seq can crash the scheduler and server

bhroam:
#### Describe Bug or Feature
If the scheduler reconfirms a degraded reservation with fewer execvnodes than it needs, this reservation can cause the scheduler and server to crash (until the reservation finishes or is deleted).

#### Describe Your Change
I only found two ways this could happen.  

We were calling pbs_asprintf() but only checking for -1 for error.  It returns the return value of vsnprintf().  The man page for vsnprintf() says it returns a negative number on error.  It is possible it was returning a different negative number than -1 and failing.

The other case was in condense_execvnode_seq().  This internally creates a linked list of execvnodes and then condenses them.  When adding an execvnode to the list, if malloc() failed, it would silently not add the execvnode and not detect the error.

The fix for this four fold:
1) The use of pbs_asprintf() was horribly inefficient.  We were having it allocate new memory to concatenate two strings, and then freeing the first.  I converted this to use a std::string which is very efficient concatenating strings.
2) I fixed the internals of condense_execvnode_seq() to detect a memory error and return NULL.
3) I catch if condense_execvnode_seq() returns NULL or if the number of execvnodes does not equal to the number we should have, we exit the cycle and immediately try again
4) The reservation's reserve_count was being set at confirmation time to the number of execvnodes sent over.  I changed this to set it at submission time based on the rrule.  This allows us to check at confirmation time if the incorrect number of execvnodes were sent.

This issue showed up once in a weekend valgrind run.  I've never been able to reproduce it outside of the debugger.  Before my fix I was able to crash both the scheduler and server.  I submitted a standing reservation with 10 occurrences, but I only returned 5.  When the daemons tried to access the execvnode_seq, it would use a negative index and crash.

Post fix:
This shows up in the scheduler log:
03/16/2021 16:45:44.172642;0080;pbs_sched;Resv;S2.henke;Invalid execvnode_seq while confirming reservation

and if I bypass the check in the scheduler and send the server an invalid execvnode_seq:
03/16/2021 16:36:49.962786;0200;Server@henke;Resv;S0.henke;Number of execvnodes given does not equal the number of occurrences left

The one failed test is not for reservations (preemption).  I reran it and it passed.

Profile: regression
|--|--|--|--|--|--|--|
|6553|3996|1|0|0|0|3995|

### DIFF
--- a/src/lib/Libutil/execvnode_seq_util.c
+++ b/src/lib/Libutil/execvnode_seq_util.c
@@ -81,16 +81,16 @@ static struct map *new_map(int);
 /* Map a sequence of words to a list of indices by which
  * they are represented in the passed string str.
  */
-static void direct_map(dictionary *, char *);
+static int direct_map(dictionary *, char *);
 
 /* find a word in a dictionary */
 static struct word *find_word(dictionary *, char *);
 
 /* Append word to dictionary */
-static void append_to_dict(dictionary *, char *, int);
+static int append_to_dict(dictionary *, char *, int);
 
 /* Append the index of a word in a string to the mapping of the word */
-static void append_to_word(dictionary *, struct word *, int);
+static int append_to_word(dictionary *, struct word *, int);
 
 /* Create a string out of all words in a dictionary and their corresponding mapped indices */
 static char *dict_to_str(dictionary *);
@@ -209,9 +209,11 @@ static struct map *new_map(int val)
  *	If the string exists in the dictionary, its index is appended
  * 	to the word entry. Otherwise it is added as a new word to the dictionary
  *
- * @return Fills the dictionary structure.
+ * @return int
+ * @retval 1 error
+ * @retval 0 success
  */
-static void
+static int
 direct_map(dictionary *dict, char *str)
 {
 	struct word *w;
@@ -220,26 +222,32 @@ direct_map(dictionary *dict, char *str)
 	int i = 0;
 
 	if (dict == NULL || str == NULL)
-		return;
+		return 1;
 
 	if ((str_copy = strdup(str)) == NULL) {
 		DBPRT(("new_word: %s\n", MALLOC_ERR_MSG));
-		return;
+		return 1;
 	}
 
 	str_tok = strtok(str_copy, TOKEN_SEPARATOR);
-	while (str_tok!=NULL) {
+	while (str_tok != NULL) {
 		w = (struct word *)find_word(dict, str_tok);
-		if (w == NULL)
-			append_to_dict(dict, str_tok, i);
-		else
-			append_to_word(dict, w, i);
+		if (w == NULL) {
+			if (append_to_dict(dict, str_tok, i)) {
+				free(str_copy);
+				return 1;
+			}
+		} else
+			if (append_to_word(dict, w, i)) {
+			free(str_copy);
+			return 1;
+			}
 		i++;
 		str_tok = strtok(NULL, TOKEN_SEPARATOR);
 	}
 	dict->max_idx = i;
 	free(str_copy);
-	return;
+	return 0;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


bhroam:
Please review.  I've WIP'd it until I get the results from TH3.

bhroam:
Tests finished.  I un-WIP'd the PR.

src/lib/Libutil/execvnode_seq_util.c
```diff
@@ -554,6 +574,9 @@ get_execvnodes_count(char *str)
 	if (str == NULL)
 		return 0;
 
+	if (str[0] == '(')
```
arungrover, line 577: Can you give an example of the str this function expects? not sure why it is returning the number of occurrences as 1 when it sees open braces.

src/lib/Libutil/execvnode_seq_util.c
```diff
@@ -220,26 +222,32 @@ direct_map(dictionary *dict, char *str)
...
-			append_to_word(dict, w, i);
+		if (w == NULL) {
+			if (append_to_dict(dict, str_tok, i)) {
+				free(str_copy);
+				return 1;
+			}
+		} else
+			if (append_to_word(dict, w, i)) {
+			free(str_copy);
+			return 1;
```
agrawalravi90: indentation looks off

src/lib/Libutil/execvnode_seq_util.c
```diff
@@ -351,14 +367,14 @@ append_to_word(dictionary *dict, struct word *w, int val)
 		}
 		tmp->next = new_map(val);
 
-		if (tmp->next == NULL)
-			return;
+		if (tmp->next == NULL)
```
agrawalravi90, line 370: nit: unwanted space at the end

arungrover (changes requested):
1 minor comment

src/lib/Libutil/execvnode_seq_util.c
agrawalravi90, line 579: can you please add a comment here to explain why you are checking for parenthesis?

bhroam: An execvnode_seq looks like N#(execvnode){0-N-1}.  This looked for the N at the front and failed with a normal exec_vnode.  A single exec_vnode will start with a '('.  I'll update the doxygen comment.

src/scheduler/resv_info.cpp
```diff
@@ -1698,20 +1658,26 @@ confirm_reservation(status *policy, int pbs_sd, resource_resv *unconf_resv, serv
 		 * each execvnode and condensing the concatenated string.
 		 */
 		if (nresv_parent->resv->is_standing)
-			short_xc = condense_execvnode_seq(execvnodes);
+			short_xc = condense_execvnode_seq(execvnodes.c_str());
 		else
-			short_xc = string_dup(execvnodes);
+			short_xc = string_dup(execvnodes.c_str());
+		
+		if (short_xc == NULL || get_execvnodes_count(short_xc) != occr_count) {
```
agrawalravi90, line 1665: If get_execvnodes_count() is only needed to make sure that condense_execvnode_seq() didn't screw up, maybe just have condense_execvnode_seq() check that the exec vnode count matches occr_count and return NULL if it doesn't? otherwise we will incur the overhead of get_execvnodes_count() for advance reservations as well. I think we can even make direct_map() return the count instead of calling get_execvnodes_count() for standing reservations.

agrawalravi90 (changes requested):
Mostly minor comments

src/scheduler/resv_info.cpp
bhroam: condense_execvnode_seq() works off of a string.  You don't pass in a reservation.  We'd have to pass in the occr_count into the function.  I don't think condense_execvnode_seq() is the right place for this check.  I think it the right place is here.

src/lib/Libutil/execvnode_seq_util.c
bhroam: I updated the doxygen comment for the expected input.

src/scheduler/resv_info.cpp
agrawalravi90: condense_execvnode_seq() is condensing an exec vnode, so it seems well within its purview to ensure that it kept the number of occurrences the same while condensing it. If you still think that it doesn't belong there, then please move the get_execvnodes_count(short_xc) != occr_count check to the if (nresv_parent->resv->is_standing) block so that it doesn't get performed for advance reservations

agrawalravi90 (changes requested):

bhroam:
I have the check where I want it.  I want this check to happen right before we confirm the reservation.  This check is very efficient.  For advanced reservations, get_execvnode_count() checks if str[0] == '(' and then returns 1.  This check is also checking if short_xc == NULL, so that is valid for advance reservations (since it is doing a string_dup()).

agrawalravi90:
I was suggesting to only move the get_execvnodes_count(short_xc) != occr_count check, not short_xc == NULL. But since get_execvnodes_count() will not be expensive for advance reservations, I'm ok with it. Thanks for clarifying.

agrawalravi90 (approved):
LGTM

arungrover (approved):
LGTM!

PR #2312, opened 2021-03-17
Fixing the proper server name at connection failure for qstat  in multi-server environement

bremanandjk:
#### Describe Bug or Feature

In a multi-server environment, qstat throws a connection error with the wrong server name.

#### Describe Your Change

Need to print the default server name in case of connection failure. 

After fix:

[root@pbs-server pbspro]# cat /etc/pbs.conf 
PBS_SERVER=pbs-server
PBS_START_SERVER=1
PBS_START_SCHED=1
PBS_START_COMM=1
PBS_START_MOM=1
PBS_EXEC=/opt/pbs
PBS_HOME=/var/spool/pbs
PBS_CORE_LIMIT=unlimited
PBS_SCP=/bin/scp
PBS_SERVER_INSTANCES=pbs-server,spark
[root@pbs-server pbspro]# 

[root@pbs-server pbspro]# qstat
Connection refused
qstat: cannot connect to server pbs-server (errno=111)
[root@pbs-server pbspro]# pbsnodes -av
Connection refused
pbsnodes: cannot connect to server pbs-server, error=111
[root@pbs-server pbspro]#

### DIFF
--- a/src/cmds/qstat.c
+++ b/src/cmds/qstat.c
@@ -2921,7 +2921,7 @@ job_no_args:
 
 				if (conn <= 0) {
 					fprintf(stderr, "qstat: cannot connect to server %s (errno=%d)\n",
-						pbs_server, pbs_errno);
+						def_server, pbs_errno);
 #ifdef NAS /* localmod 071 */
 					(void)tcl_stat(error, NULL, tcl_opt);
 #else
@@ -3106,7 +3106,7 @@ job_no_args:
 que_no_args:
 				conn = cnt2server(server_out);
 				if (conn <= 0) {
-					fprintf(stderr, "qstat: cannot connect to server %s (errno=%d)\n", pbs_server, pbs_errno);
+					fprintf(stderr, "qstat: cannot connect to server %s (errno=%d)\n", def_server, pbs_errno);
 #ifdef NAS /* localmod 071 */
 					(void)tcl_stat(error, NULL, tcl_opt);
 #else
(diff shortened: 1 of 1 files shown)


agrawalravi90 (changes requested):
@bremanandjk  Can you please clarify which server was qstat not able to connect to in your example? (pbs-server or spark?). From your code changes it seems like we'll always say that qstat was not able to connect to the default server, even if it was actually not able to connect to one of the other multi-servers, is that true?

bremanandjk:
> @bremanandjk Can you please clarify which server was qstat not able to connect to in your example? (pbs-server or spark?). From your code changes it seems like we'll always say that qstat was not able to connect to the default server, even if it was actually not able to connect to one of the other multi-servers, is that true?

In this example, both the servers were down. 
The expected error msg should be with the default server name in the case of multi-server mode. This is the same behaviour in pbsnodes o/p also.

bremanandjk:
@agrawalravi90 Plz let me know if you have any further questions.

agrawalravi90 (approved):
I think it would be more useful if we said which multi-server couldn't be connected to instead of the default server, for both qstat and pbsnodes. But this is better than being inconsistent, so I'm ok with it.

ashwathraop (approved):
LGTM, do you have test results?

bremanandjk:
Thanks for your approval @agrawalravi90 & @ashwathraop 
regression results:
Profile: regression
|--|--|--|--|--|--|--|
|6684|3999|3|0|0|5|3991|

Description: Rerun Only Failed Tests From #6684
Profile: regression
|--|--|--|--|--|--|--|
|6687|3|1|0|0|0|2|

That one test case failure observed in mainline without my changes also. 

Profile: regression
|--|--|--|--|--|--|--|
|6688|3|1|0|0|0|2|

bremanandjk:
> LGTM, do you have test results?

I have shared regression results and test results in description page. Its ready to merge.

PR #2314, opened 2021-03-18
Fixed possible free of unitialized variable

bhroam:
#### Describe Bug or Feature
If the server passes execvnode="" to the scheduler, the scheduler will try and free an unitialized value.

It currently isn't possible for the server to pass "" to the scheduler, but it will be possible in the upcoming multi-server persistent scheduler.

#### Describe Your Change
Initialized variable to NULL

### DIFF
--- a/src/scheduler/node_info.cpp
+++ b/src/scheduler/node_info.cpp
@@ -4293,7 +4293,7 @@ parse_execvnode(char *execvnode, server_info *sinfo, selspec *sel)
 	char *excvndup;
 	char *node_name;
 	int num_el;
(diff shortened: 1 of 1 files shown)


arungrover (approved):
LGTM!

vstumpf (commented):
can you use `nullptr` c:

bhroam:
@vstumpf I think I'll leave this as NULL for now and then take up a refactoring project later to turn all NULL pointers into nullptr.

vstumpf (approved):

PR #2316, opened 2021-03-18
Updated test_advance_reservation_with_job_array for slow machines

bhroam:
#### Describe Bug or Feature
The test was failing on slow machines due to the static sleep times.  For example if you set a job's sleep time to 60s and sleep for 60s, there is always some difference between when the job starts and the sleep time.

#### Describe Your Change
Modified all reservations to sleep the real amount of time between when the sleep starts and the reservation starts.  Also modified a sleep for job start to better catch the end of it.


birbalsain (approved):
Changes LGTM

arungrover (approved):
LGTM

PR #2317, opened 2021-03-19
Initialize nslookup mutex before tpp_lock

vishwaks:
#### Describe Bug or Feature
server, sched processes are exiting after the start with log message "Failed to lock mutex".

#### Describe Your Change
nslookup mutex which introduced in #2299 was acquiring the lock before initializing from tpp_init -> tpp_get_addresses -> tpp_resolv_sock_host. The nslookup mutex is initialized in tpp_transport_init which will be called after tpp_get_addresses.
Solving this issue by initializing nslookup mutex using PTHREAD_MUTEX_INITIALIZER macro(nitializes the static mutex mutex, setting its attributes to default values) And also apply this mutex only to linux platform as fork() API used for linux platforms only.

Id: 6593
Profile: regression

Test Summary: total: 1956, fail: 0, error: 0, timedout: 0, skip: 0, pass: 1956

### DIFF
--- a/src/lib/Libtpp/tpp_platform.c
+++ b/src/lib/Libtpp/tpp_platform.c
@@ -710,19 +710,24 @@ tpp_sock_resolve_host(char *host, int *count)
 	hints.ai_socktype = SOCK_STREAM;
 	hints.ai_protocol = IPPROTO_TCP;
 
+#ifndef WIN32
 	/* 
 	 * introducing a new mutex to prevent child process from 
 	 * inheriting getaddrinfo mutex using pthread_atfork handlers
 	 */
 	tpp_lock(&tpp_nslookup_mutex);
-
+#endif
 	if ((rc = getaddrinfo(host, NULL, &hints, &pai)) != 0) {
+#ifndef WIN32
 		tpp_unlock(&tpp_nslookup_mutex);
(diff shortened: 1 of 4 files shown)


src/lib/Libtpp/tpp_platform.c
```diff
@@ -710,19 +710,24 @@ tpp_sock_resolve_host(char *host, int *count)
...
+#ifndef WIN32
 	/* 
 	 * introducing a new mutex to prevent child process from 
 	 * inheriting getaddrinfo mutex using pthread_atfork handlers
 	 */
 	tpp_lock(&tpp_nslookup_mutex);
-
+#endif
 	if ((rc = getaddrinfo(host, NULL, &hints, &pai)) != 0) {
+#ifndef WIN32
```
subhasisb, line 721: u could change the code such that there a single unlock only

rc = getaddrinfo(...);
#ifndef WIN32
      tpp_unlock(&tpp_nslookup_mutex);
#endif
     if (rc != 0) {
          error message
     }

subhasisb (changes requested):

subhasisb (approved):

vstumpf (approved):
LGTM

PR #2318, opened 2021-03-19
Remove variable definition from tpp_internal.h

toonen:
This PR fixes a _multiple definition of `tpp_nslookup_mutex'_ error by moving the definition into a source file and changing the instance in the header file to a declaration.


vstumpf:
I believe this is now fixed with #2317.

toonen:
It appears so.  Closing this PR.

PR #2319, opened 2021-03-19
Add pbs_errno access to the pbs_ifl Python module

toonen:
This PR adds an accessor function for the thread specific variable pbs_errno to the pbs_ifl Python module, as well as the PBSE constants and the pbs_err_to_txt() translation function.  An accessor function is also included for the other pbs_ifl thread specific variable, pbs_server.


bayucan (approved):
Looks good.

PR #2323, opened 2021-03-23
Fix race condition in TestJobArray

sakshamgarg:
#### Describe Bug or Feature

"test_running_subjob_survive_restart" & "test_running_subjob_survive_restart_with_history" fails intermittently to verify job state by giving below error

_ptl.lib.ptl_error.PtlExpectError: rc=1, rv=False, msg=expected on server x6668-5-0: job_state = R && substate = 42 job 0[2].x6668-5-0  got: job_state = X_

Since the sleep time of the array job was less, sometimes it took longer for the server to restart and connect to the MoM. By the time server would restart and connect to MoM, the subjob finishes and it is goes to 'X' state.

#### Describe Your Change

Instead of increasing the sleep time, I am checking for both 'R' and 'X' state of the subjob. This way we don't need to worry about the time it takes to restart server on any slow machine

Result before change:
Description: Tests from given build on platforms UBUNTU1804, SUSE15, UBUNTU1604, CENTOS8, RHEL8, CENTOS7, RHEL7, SLES12, SLES15
Profile: regression
|--|--|--|--|--|--|--|
|6668|18|4|0|0|0|14|

Result after change:
Description: Rerun All Tests From #6666
Profile: regression
|--|--|--|--|--|--|--|
|6667|18|0|0|0|0|18|

### DIFF
--- a/test/tests/functional/pbs_job_array.py
+++ b/test/tests/functional/pbs_job_array.py
@@ -144,47 +144,23 @@ e.accept()
         j = Job(TEST_USER, attrs={
             ATTR_J: '1-3', 'Resource_List.select': 'ncpus=1'})
 
-        j.set_sleep_time(20)
-
         j_id = self.server.submit(j)
-        subjid_2 = j.create_subjob_id(j_id, 2)
+        subjid_1 = j.create_subjob_id(j_id, 1)
 
         # 1. check job array has begun
         self.server.expect(JOB, {'job_state': 'B'}, j_id)
 
-        # 2. wait till subjob 2 starts running
-        self.server.expect(JOB, {'job_state': 'R'}, subjid_2, offset=20)
+        # 2. check subjob 1 started running
+        self.server.expect(JOB, {'job_state': 'R'}, subjid_1)
 
         # 3. Kill and restart the server
         self.kill_and_restart_svr()
 
         # 4. array job should be B
-        self.server.expect(JOB, {'job_state': 'B'}, j_id, max_attempts=1)
+        self.server.expect(JOB, {'job_state': 'B'}, j_id)
 
-        # 5. subjob 1 should be X
-        self.server.expect(JOB, {'job_state': 'X'},
-                           j.create_subjob_id(j_id, 1), max_attempts=1)
-
-        subjob2_finished = False
-        # 6. subjob 2 should be in R or X state
-        try:
-            self.server.expect(JOB, {'job_state': 'R'}, subjid_2,
-                               max_attempts=1)
-        except PtlExpectError:
-            self.server.expect(JOB, {'job_state': 'X'}, subjid_2,
-                               max_attempts=1)
-            subjob2_finished = True
-
-        subjid_3 = j.create_subjob_id(j_id, 3)
-        # 7. subjob 3 should be Q if subjob2 is in R state
-        if not subjob2_finished:
-            self.server.expect(JOB, {'job_state': 'Q'}, subjid_3,
-                               max_attempts=1)
-        else:
-            try:
-                self.server.expect(JOB, {'job_state': 'R'}, subjid_3)
-            except PtlExpectError:
-                self.server.expect(JOB, {'job_state': 'X'}, subjid_3)
+        # 5. subjob 1 should be R
+        self.server.expect(JOB, {'job_state': 'R'}, subjid_1)
 
     def test_running_subjob_survive_restart_with_history(self):
         """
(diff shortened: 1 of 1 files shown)


birbalsain (approved):
Changes LGTM

raviv1995 (approved):

test/tests/functional/pbs_job_array.py
```diff
@@ -165,12 +165,26 @@ def test_running_subjob_survive_restart(self):
...
-        # 6. subjob 2 should be R
-        self.server.expect(JOB, {'job_state': 'R'}, subjid_2, max_attempts=1)
+        subjob2_finished = False
+        # 6. subjob 2 should be in R or X state
+        try:
+            self.server.expect(JOB, {'job_state': 'R'}, subjid_2,
+                               max_attempts=1)
+        except PtlExpectError:
+            self.server.expect(JOB, {'job_state': 'X'}, subjid_2,
+                               max_attempts=1)
```
bhroam, line 175: Since you do max_attempts=1, it might be better to call status(), and check the job_state.  The purpose of expect() is to check over time.  You only check once, so it isn't any different than a status(), except the two expect() calls will call status() twice.

test/tests/functional/pbs_job_array.py
```diff
@@ -165,12 +165,26 @@ def test_running_subjob_survive_restart(self):
...
+        subjid_3 = j.create_subjob_id(j_id, 3)
+        # 7. subjob 3 should be Q if subjob2 is in R state
+        if not subjob2_finished:
+            self.server.expect(JOB, {'job_state': 'Q'}, subjid_3,
+                               max_attempts=1)
+        else:
+            try:
+                self.server.expect(JOB, {'job_state': 'R'}, subjid_3)
+            except PtlExpectError:
+                self.server.expect(JOB, {'job_state': 'X'}, subjid_3)
```
bhroam, line 187: Instead of the 2 expects, do {'job_state': 'R', 'job_state': 'X'}, attrop=PTL_OR

bhroam (changes requested):

test/tests/functional/pbs_job_array.py
```diff
@@ -165,12 +165,26 @@ def test_running_subjob_survive_restart(self):
...
 
-        # 6. subjob 2 should be R
-        self.server.expect(JOB, {'job_state': 'R'}, subjid_2, max_attempts=1)
+        subjob2_finished = False
+        # 6. subjob 2 should be in R or X state
+        try:
+            self.server.expect(JOB, {'job_state': 'R'}, subjid_2,
+                               max_attempts=1)
+        except PtlExpectError:
+            self.server.expect(JOB, {'job_state': 'X'}, subjid_2,
```
arungrover, line 174: I think by doing this you are changing the intent of the test. If server has a bug where it starts considering running subjobs as exitied upon a restart then this test will never catch that.

IMO it is better to increase the sleep time.

sakshamgarg: I had reservations about this, but I think you are right. After going through the documentation I found this

> Subjob exit status is available only as long as the subjob is in job history. When a subjob is not in job history, a failed or terminated subjob will show an exit status of Finished, instead of failed or terminated.

Would it be better to verify the exit status as zero of that subjob from the server logs? Or checking the exit_status is also not a good idea? What do you think? I just want to avoid increasing the sleep as much as possible, so that in slower machines it does not fail again.

test/tests/functional/pbs_job_array.py
sakshamgarg: Thanks for the suggestion @bhroam. I have been looking for something like this in PTL but couldn't find it.

test/tests/functional/pbs_job_array.py
arungrover: I wouldn't rely on exit status either, as that can change in future. How about we increase the sleep time of the job and not wait for the first subjob to finish.  In this test, step 2 where we wait for second subjob to start really serves no purpose. 
If we restart the server as soon as we find the first subjob running, that is all we need. Doing this will speed up the test anyway, since we are not going to wait for a subjob to finish.

sakshamgarg:
I have added sleep time in "test_multiple_server_restarts" as this test case was also failing for same the reason. The fix has been added in this commit https://github.com/openpbs/openpbs/pull/2323/commits/5e0e63ba83a580b86b2aa67899e517765de11fc6

bhroam (approved):
LGTM

arungrover (approved):
Thanks for making the change.
LGTM

PR #2328, opened 2021-03-24
C++ refactor of resdef structure

bhroam:
#### Describe Bug or Feature
The scheduler uses a resdef structure to store resource definitions it gets from the server.  It used to store this in a NULL terminated array.  It was semi-sorted so some definitions appeared in well known locations.  This was so you could directly index  into the array for often used definitions.  Other than that, finding a definition was an O(N) search.

#### Describe Your Change
I replaced the NULL terminated array with a std::unordered_map.  This allows for O(1) searches for all elements.
I did 2 performance tests (they're in TestSchedPerf).  The calendar test showed no difference.  The test which runs 10k 1 cpu jobs was was about 5s faster.  Master ran in about 45s.  The new code ran in about 40s.

The 3 failed tests were hook tests which I reran and they passed.

Profile: regression
|--|--|--|--|--|--|--|
|6696|3999|3|0|0|0|3996|

### DIFF
--- a/src/scheduler/buckets.cpp
+++ b/src/scheduler/buckets.cpp
@@ -911,9 +911,9 @@ int job_should_use_buckets(resource_resv *resresv) {
 
 	/* Job's requesting specific hosts or vnodes use the standard path */
 	const auto& defs = resresv->select->defs;
-	if (defs.find(getallres(RES_HOST)) != defs.end())
+	if (defs.find(allres["host"]) != defs.end())
 		return 0;
-	if (defs.find(getallres(RES_VNODE)) != defs.end())
+	if (defs.find(allres["vnode"]) != defs.end())
 		return 0;
 	/* If a job has an execselect, it means it's requesting vnode */
 	if (resresv->execselect != NULL)
--- a/src/scheduler/job_info.cpp
+++ b/src/scheduler/job_info.cpp
@@ -667,12 +667,12 @@ query_jobs_chunk(th_data_query_jinfo *data)
 		/* Find out if it is a shrink-to-fit job.
 		 * If yes, set the duration to max walltime.
 		 */
-		req = find_resource_req(resresv->resreq, getallres(RES_MIN_WALLTIME));
+		req = find_resource_req(resresv->resreq, allres["min_walltime"]);
 		if (req != NULL) {
 			resresv->is_shrink_to_fit = 1;
 			/* Set the min duration */
 			resresv->min_duration = (time_t) req->amount;
-			req = find_resource_req(resresv->resreq, getallres(RES_MAX_WALLTIME));
+			req = find_resource_req(resresv->resreq, allres["max_walltime"]);
 
 #ifdef NAS /* localmod 026 */
 			/* if no max_walltime is set then we want to look at what walltime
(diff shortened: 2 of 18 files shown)


src/scheduler/resource.cpp
```diff
@@ -96,129 +88,72 @@
...
  * @brief
- * 		query a pbs server for the resources it knows about
+ * 		query a pbs server for the resources it knows about and fill in the global unordered_map
  *
  * @param[in]	pbs_sd	-	communication descriptor to pbs server
  *
- * @return	resdef**
- * @retval	array of resources from server
- * @retval	NULL	: on error
+ * @return	bool
```
arungrover, line 95: Looks like you are returning an unordered map from this function.

src/scheduler/resource.cpp
```diff
@@ -696,71 +340,66 @@ create_resource_signature(schd_resource *reslist, std::unordered_set<resdef *>&
...
-				def_is_bool, NULL, NO_FLAGS);
-			if (boolres == NULL)
-				error = 1;
-		}
-
-		if (!conf.res_to_check.empty()) {
-			conf.resdef_to_check = resstr_to_resdef(conf.res_to_check);
-			if (conf.resdef_to_check.empty())
-				error = 1;
+	clear_limres();
```
arungrover, line 355: Should we move this to the end of the function? In this change, We might end up clearing it when when query_resources fail.

src/scheduler/buckets.cpp
```diff
@@ -911,9 +911,9 @@ int job_should_use_buckets(resource_resv *resresv) {
 
 	/* Job's requesting specific hosts or vnodes use the standard path */
 	const auto& defs = resresv->select->defs;
-	if (defs.find(getallres(RES_HOST)) != defs.end())
+	if (defs.find(allres["host"]) != defs.end())
```
arungrover, line 914: I actually like accessing the resource via this getter function that we previously had. If you want, you can just turns the enums to macros of the resource name and then you don't have to make this change in so many places.

arungrover (changes requested):
Minor comments

src/scheduler/resource.cpp
bhroam: ahh yeah.  When I originally modified this function, I had it directly modify allres.  I then realized I needed the old and new maps so I could update the defs in last_running.  I'll update it.

src/scheduler/job_info.cpp
```diff
@@ -4248,14 +4247,14 @@ formula_evaluate(const char *formula, resource_resv *resresv, resource_req *resr
 	}
 
 
-	for (i = 0; consres[i] != NULL; i++) {
-		req = find_resource_req(resreq, consres[i]);
+	for (const auto& cr: consres) {
```
vstumpf, line 4250: add a space before `:`

src/scheduler/resource.cpp
```diff
@@ -96,129 +88,72 @@
...
-				break;
-			}
-		}
-		if (j == RES_HIGH) {
-			defarr[i] = def;
-			i++;
-			defarr[i] = NULL;
+		def = new resdef(cur_bs->name, flags, rtype);
+		if (def == NULL) {
+			for (auto& d: tmpres)
```
vstumpf, line 134: space before `:`

src/scheduler/resource.cpp
```diff
@@ -96,129 +88,72 @@
...
-			 * the rest.  We need to fill in the holes so we can free the entire thing
-			 */
-			defarr[i] = new_resdef();
-			error = 1;
+	/**
+	 * @par Make sure all the well known resources are sent to us.
+	 *      This is to allow us to directly index into the allres umap.
+	 *      Do not directly index into the allres umap for non-well known resources.  Use find_resdef()
+	 */
+	for (const auto& r: well_known_res) {
```
vstumpf, line 146: space before `:`

src/scheduler/resource.cpp
```diff
@@ -696,71 +350,71 @@ create_resource_signature(schd_resource *reslist, std::unordered_set<resdef *>&
...
-	if (error) {
-		if (consres != NULL) {
-			free(consres);
-			consres = NULL;
-		}
-		if (boolres != NULL) {
-			free(boolres);
-			boolres = NULL;
-		}
+	for (auto& d: allres)
```
vstumpf, line 392: space before `:`

src/scheduler/resource.cpp
```diff
@@ -696,71 +350,71 @@ create_resource_signature(schd_resource *reslist, std::unordered_set<resdef *>&
...
 
-			conf.resdef_to_check.clear();
+	allres = tmpres;
 
-		if (allres != NULL) {
-			free_resdef_array(allres);
-			allres = NULL;
-		}
+	consres.clear();
+	for (const auto& def: allres) {
```
vstumpf, line 398: space before `:`

vstumpf (changes requested):
thoughts on defining the resource names as macros? something like: 
`#define RES_NAME_HOST "host"`?

bhroam:
@vstumpf I'd love it if we had macros for our resources like we do for our attributes, but those would have to reside in pbs_ifl.h.  They'd be used by the xml files, the server, and the scheduler.  Just creating macros in the scheduler doesn't really help because the resources are quoted strings in the XML file.

I think this is an excellent idea, but it is a lot of work.  I'll consider doing it for this PR, or just filing a ticket for us to do later.

vstumpf:
We can leave it for now, and hope that we don't change the resource names.

vstumpf (approved):

arungrover (approved):

PR #2332, opened 2021-03-26
Implement manage_rlimit_as; add PTL test

alexis-cousein:
#### Describe Bug or Feature

When enabling the "memsw" functionality in the cgroup hook, the semantics of "vmem" change.

Without it "vmem" is managed by MoM and denotes the sum of the address spaces used by all processes on the mother superior node, and MoM sets a per process limit for RLIMIT_AS (since clearly a single process cannot use less address space than the sum for all processes).

But when the cgroup hook "memsw" functionality is enabled, "vmem" requests instead specify the sum of physical memory plus swap usage for the job, which is often smaller than the address space limit.

Indeed, it is doubtful that in the face of kernel limits on real memory plus swap usage anyone would care about the address space limits, especially on 64-bit systems (which have an 8 Exabyte positive address address space). Address space limits were used in the past when there were no physical memory and swap limits enforced by the kernel, but cgroups make these unnecessary.

But on most current versions of PBSPro, the MoM does not know about this change in semantics for vmem, and it still sets a RLIMIT_AS limit for all processes in the job. Since many applications allocate much more address space than they actually use memory, that can make applications fail even though they are staying well within the memory+swap usage corresponding to their vmem requests.

Even MoM itself can be a victim for small vmem limits: when the job starter child forks to run the execjob_launch hook, the fork and exec of pbs_python can make it use up to 150mb-200mb of address space, so setting a small RLIMIT_AS limit can make either the job starter or the hooks fail.

Furthermore, a per-job 'pvmem' resource request unambiguously requests exactly a RLIMIT_AS, so there is no real need to use vmem to specify this limit if one is desired.

To summarise: when enabling the "memsw" functionality in the cgroup hook, RLIMIT_AS should be unlimited, unless the job also requests "pvmem" (which is an unambiguous request to set RLIMIT_AS).

Note: this really is a MoM bug; ideally it should be possible to enable or disable automatic setting of the RLIMIT_AS limit when 'vmem' is requested in the MoM config file. But until such a mechanism exists, at least the cgroup hook can mitigate the effect for values of 'vmem' large enough to allow the job starter and the hooks to run.

#### Describe Your Change

A new parameter is introduced to the cgroup configuration file, manage_rlimit_as, which by default is true.

If enabled, then the cgroup hook will reset the RLIMIT_AS process limit for task processes to either unlimited (if pvmem is not specified) or the value of pvmem requested for the job.

This functionality requires a kernel that supports the prlimit system call (i.e. Linux kernel 2.6.36 and above).

If hooks use Python 3, i.e. for PBS versions 2020 and above, that is the only requirement.

For older versions of PBS, the 'prlimit' command should be present. This command is available in util-linux versions 2.21 and above on:
-CentOS7/RedHat from version 7.0
-SLES from version 12 onward,
-Ubuntu from version Ubuntu 16.04 onward
-Debian versions from version 8.0 onward
(i.e. most OS versions with a kernel that supports the prlimit system call and have a version of util-linux >= 2.21).

If the support for changing process limits of other processes is not present, then limits set by MoM are never changed (i.e. the hook behaves as before, and as if manage_rlimit_as were disabled).

When the flag is disabled, limits set by MoM will not be changed. It is unlikely anyone would want this behaviour, except on MoMs where it is possible to disable enforcement of 'vmem' as an RLIMIT_AS limit set on processes.
#### Link to Design Doc

https://openpbs.atlassian.net/wiki/spaces/PD/pages/2668199965/Fix+incorrect+MoM+RLIMIT+AS+when+vmem+is+used+for+the+group+hook+s+memsw+functionality

A PTL test was added, test_manage_rlimit_as. 

A job requests -lselect=1:ncpus=0:mem=300mb:vmem=300mb and the test checks that ulimit -v in a job returns 'unlimited'. A job requests -lselect=1:ncpus=0:mem=300mb:vmem=300mb,pvmem=300mb, and then ulimit -v should return 307200 (since ulimit -v output is in kb units).

Relevant portion of test_manage_rlimit_as output:
```
2021-03-26 13:08:54,255 INFO     manager on dragon: set server {'job_history_enable': 'True'}
2021-03-26 13:08:54,255 INFOCLI  dragon: /opt/pbs/bin/qmgr -c set server job_history_enable=True
2021-03-26 13:08:54,283 INFOCLI  dragon: /opt/pbs/bin/qstat -Bf dragon
2021-03-26 13:08:54,294 INFO     expect on server dragon: job_history_enable = True server dragon ...  OK
2021-03-26 13:08:54,294 INFO     Current config: {
    "cgroup_prefix"         : "pbs_jobs",
    "exclude_hosts"         : [],
    "exclude_vntypes"       : ["no_cgroups"],
    "run_only_on_hosts"     : [],
    "periodic_resc_update"  : true,
    "vnode_per_numa_node"   : false,
    "online_offlined_nodes" : true,
    "use_hyperthreads"      : false,
    "ncpus_are_cores"       : false,
    "manage_rlimit_as"      : true,
    "cgroup" : {
        "cpuacct" : {
            "enabled"            : true,
            "exclude_hosts"      : [],
            "exclude_vntypes"    : []
        },
        "cpuset" : {
            "enabled"            : true,
            "exclude_cpus"       : [],
            "exclude_hosts"      : [],
            "exclude_vntypes"    : [],
            "mem_fences"         : true,
            "mem_hardwall"       : false,
            "memory_spread_page" : false
        },
        "devices" : {
            "enabled"            : true,
            "exclude_hosts"      : [],
            "exclude_vntypes"    : [],
            "allow"              : [
                "b *:* rwm",
                "c *:* rwm"
            ]
        },
        "hugetlb" : {
            "enabled"            : false,
            "exclude_hosts"      : [],
            "exclude_vntypes"    : [],
            "default"            : "0MB",
            "reserve_percent"    : 0,
            "reserve_amount"     : "0MB"
        },
        "memory" : {
            "enabled"            : true,
            "exclude_hosts"      : [],
            "exclude_vntypes"    : [],
            "soft_limit"         : false,
            "default"            : "100MB",
            "reserve_percent"    : 0,
            "swappiness"         : 0,
            "reserve_amount"     : "2GB",
            "enforce_default"    : true,
            "exclhost_ignore_default" : true
        },
        "memsw" : {
            "enabled"            : true,
            "exclude_hosts"      : [],
            "exclude_vntypes"    : [],
            "default"            : "100MB",
            "reserve_percent"    : 0,
            "reserve_amount"     : "10GB",
            "manage_cgswap"      : true,
            "enforce_default"    : true,
            "exclhost_ignore_default" : true
        }
    }
}

2021-03-26 13:09:01,301 INFO     manager on dragon: import hook pbs_cgroups {'content-type': 'application/x-config', 'content-encoding': 'default', 'input-file': '/tmp/PtlPbs9u4sidjl'}
2021-03-26 13:09:01,302 INFOCLI  dragon: sudo -H /opt/pbs/bin/qmgr -c import hook pbs_cgroups application/x-config default /tmp/PtlPbs9u4sidjl
2021-03-26 13:09:01,326 ERROR    err: ["hook 'pbs_cgroups' contents overwritten"]
2021-03-26 13:09:04,892 INFO     mom dragon log match: searching for "pbs_cgroups.CF;copy hook-related file request received" - with existence... OK
2021-03-26 13:09:04,903 INFOCLI2 dragon: sudo -H cat /var/spool/PBS/server_priv/hooks/pbs_cgroups.CF
2021-03-26 13:09:04,923 INFOCLI2 dragon: sudo -H cat /var/spool/PBS/mom_priv/hooks/pbs_cgroups.CF
2021-03-26 13:09:04,941 INFO     server & mom pbs_cgroups.CF match
2021-03-26 13:09:08,945 INFO     mom dragon: sent signal -HUP
2021-03-26 13:09:08,945 INFOCLI2 dragon: sudo -H /usr/bin/cat /var/spool/PBS/mom_priv/mom.lock
2021-03-26 13:09:08,964 INFOCLI2 dragon: sudo -H kill -HUP 554166
2021-03-26 13:09:12,543 INFO     mom dragon log match: searching for "hook_perf_stat;label=hook_exechost_startup_pbs_cgroups_.* profile_stop" - using regular expression  - with existence - No match
2021-03-26 13:09:17,143 INFO     mom dragon log match: searching for "hook_perf_stat;label=hook_exechost_startup_pbs_cgroups_.* profile_stop" - using regular expression  - with existence... OK
2021-03-26 13:09:17,153 INFO     job: executable set to /bin/sleep with arguments: 100
2021-03-26 13:09:17,154 INFOCLI2 dragon: which chmod
2021-03-26 13:09:17,163 INFOCLI2 dragon: /usr/bin/chmod 755 /tmp/PtlPbsJobScript_fdaldlg
2021-03-26 13:09:17,168 INFOCLI  dragon: sudo -H -u pbsuser /opt/pbs/bin/qsub -l select=1:ncpus=0:mem=300mb:vmem=300mb:vnode=dragon /tmp/PtlPbsJobScript_fdaldlg
2021-03-26 13:09:17,203 INFO     submit to dragon as pbsuser: job 4424.dragon OrderedDict([('Resource_List.select', '1:ncpus=0:mem=300mb:vmem=300mb:vnode=dragon')])
2021-03-26 13:09:17,203 INFOCLI  job script /tmp/PtlPbsJobScript_fdaldlg
---
#!/bin/bash
ulimit -v
---
2021-03-26 13:09:17,203 INFO     server dragon: expect offset set to 1
2021-03-26 13:09:18,204 INFOCLI  dragon: /opt/pbs/bin/qstat -x -f 4424.dragon
2021-03-26 13:09:18,218 INFO     expect on server dragon: job_state = F job 4424.dragon  got: job_state = R
2021-03-26 13:09:18,719 INFOCLI  dragon: /opt/pbs/bin/qmgr -c set server scheduling=True
2021-03-26 13:09:18,738 INFOCLI  dragon: /opt/pbs/bin/qstat -x -f 4424.dragon
2021-03-26 13:09:18,950 INFO     expect on server dragon: job_state = F job 4424.dragon attempt: 2  got: job_state = E
2021-03-26 13:09:19,451 INFOCLI  dragon: /opt/pbs/bin/qmgr -c set server scheduling=True
2021-03-26 13:09:19,470 INFOCLI  dragon: /opt/pbs/bin/qstat -x -f 4424.dragon
2021-03-26 13:09:19,482 INFO     expect on server dragon: job_state = F job 4424.dragon attempt: 3 ...  OK
2021-03-26 13:09:19,482 INFO     status on dragon: job 4424.dragon
2021-03-26 13:09:19,482 INFOCLI  dragon: /opt/pbs/bin/qstat -x -f 4424.dragon
2021-03-26 13:09:19,693 INFOCLI2 dragon: sudo -H /usr/bin/cat /home/pbsuser/PtlPbsJobScript_fdaldlg.o4424
2021-03-26 13:09:19,714 INFO     job_out=unlimited
2021-03-26 13:09:19,714 INFO     Job that requests vmem but no pvmem correctly has unlimited RLIMIT_AS
2021-03-26 13:09:19,714 INFO     job: executable set to /bin/sleep with arguments: 100
2021-03-26 13:09:19,715 INFOCLI2 dragon: /usr/bin/chmod 755 /tmp/PtlPbsJobScriptld59nol_
2021-03-26 13:09:19,723 INFOCLI  dragon: sudo -H -u pbsuser /opt/pbs/bin/qsub -l select=1:ncpus=0:mem=300mb:vmem=300mb:vnode=dragon -l pvmem=300mb /tmp/PtlPbsJobScriptld59nol_
2021-03-26 13:09:19,745 INFO     submit to dragon as pbsuser: job 4425.dragon OrderedDict([('Resource_List.select', '1:ncpus=0:mem=300mb:vmem=300mb:vnode=dragon'), ('Resource_List.pvmem', '300mb')])
2021-03-26 13:09:19,745 INFOCLI  job script /tmp/PtlPbsJobScriptld59nol_
---
#!/bin/bash
ulimit -v
---
2021-03-26 13:09:19,745 INFO     server dragon: expect offset set to 1
2021-03-26 13:09:20,746 INFOCLI  dragon: /opt/pbs/bin/qstat -x -f 4425.dragon
2021-03-26 13:09:20,761 INFO     expect on server dragon: job_state = F job 4425.dragon  got: job_state = E
2021-03-26 13:09:21,261 INFOCLI  dragon: /opt/pbs/bin/qmgr -c set server scheduling=True
2021-03-26 13:09:21,281 INFOCLI  dragon: /opt/pbs/bin/qstat -x -f 4425.dragon
2021-03-26 13:09:21,293 INFO     expect on server dragon: job_state = F job 4425.dragon attempt: 2  got: job_state = E
2021-03-26 13:09:21,794 INFOCLI  dragon: /opt/pbs/bin/qmgr -c set server scheduling=True
2021-03-26 13:09:21,811 INFOCLI  dragon: /opt/pbs/bin/qstat -x -f 4425.dragon
2021-03-26 13:09:22,023 INFO     expect on server dragon: job_state = F job 4425.dragon attempt: 3 ...  OK
2021-03-26 13:09:22,024 INFO     status on dragon: job 4425.dragon
2021-03-26 13:09:22,024 INFOCLI  dragon: /opt/pbs/bin/qstat -x -f 4425.dragon
2021-03-26 13:09:22,237 INFOCLI2 dragon: sudo -H /usr/bin/cat /home/pbsuser/PtlPbsJobScriptld59nol_.o4425
2021-03-26 13:09:22,257 INFO     job_out=307200
2021-03-26 13:09:22,257 INFO     Job that requests 300mb pvmem correctly has 300mb RLIMIT_AS
```

### DIFF
--- a/src/hooks/cgroups/pbs_cgroups.PY
+++ b/src/hooks/cgroups/pbs_cgroups.PY
@@ -2846,6 +2846,29 @@ class CgroupUtils(object):
         else:
             self.subsystems = self._target_subsystems()
 
+        # Check if memsw is enabled despite lack of kernel support
+        # if so disable memsw
+        if 'memsw' in self.subsystems:
+            memsw_limit_file = (self.paths['memsw']
+                                + 'limit_in_bytes')
+            if not os.path.isfile(memsw_limit_file):
+                pbs.logmsg(pbs.EVENT_ERROR,
+                           'CONFIG ERROR: CF enables memsw '
+                           'but %s is missing'
+                           % memsw_limit_file)
+                pbs.logmsg(pbs.EVENT_ERROR,
+                           'CONFIG ERROR: kernel swapaccount parameter is 0, '
+                           'disabling memsw subsystem')
+                pbs.logmsg(pbs.EVENT_ERROR,
+                           'CONFIG ERROR: to fix this, either add '
+                           'swapaccount=1 to kernel command line '
+                           'or disable memsw')
+                self.cfg['cgroup']['memsw']['enabled'] = False
+                self.subsystems.remove('memsw')
+                pbs.logmsg(pbs.EVENT_DEBUG4,
+                           'remaining enabled subsystems: '
+                           + repr(self.subsystems))
+
         # Return now if nothing is enabled
         if not self.subsystems:
             pbs.logmsg(pbs.EVENT_DEBUG2, '%s: No cgroups enabled' %
--- a/src/hooks/cgroups/pbs_cgroups.PY
+++ b/src/hooks/cgroups/pbs_cgroups.PY
@@ -2889,13 +2889,23 @@ class CgroupUtils(object):
         Write a message to the job stderr file
         """
         pbs.logmsg(pbs.EVENT_DEBUG4, '%s: Method called' % caller_name())
+        filename = None
         try:
-            filename = job.stderr_file()
+            if str(job.Join_Path) == 'oe':
+                # If we write in stderr we might do it after the join!
+                # so write in stdout instead
+                filename = job.stdout_file()
+            else:
+                filename = job.stderr_file()
+            pbs.logmsg(pbs.EVENT_DEBUG4,"Message to be written to %s: %s"
+                       % (str(filename), msg.strip()))
             if filename is None:
                 return
             with open(filename, 'a') as desc:
                 desc.write(msg)
         except Exception:
+            # tough luck for user, but not fatal to system
+            # Note: sometimes normal (e.g. interactive jobs)
             pass
 
     def _target_subsystems(self):
@@ -4751,7 +4761,7 @@ class CgroupUtils(object):
                                 and pbs.event().job.in_ms_mom()):
                             self.write_to_stderr(pbs.event().job,
                                                  "Cgroup mem limit "
-                                                 "exceeded: %s" % (err_msg))
+                                                 "exceeded: %s\n" % (err_msg))
             elif subsys == 'memsw':
                 max_vmem = self._get_max_memsw_usage(jobid)
                 if max_vmem is None:
@@ -4771,7 +4781,7 @@ class CgroupUtils(object):
                     if vmem_failcnt > 0:
                         err_msg = self._get_error_msg(jobid)
                         pbs.logjobmsg(jobid,
-                                      'Cgroup memsw limit exceeded: %s' %
+                                      'Cgroup memsw limit exceeded: %s\n' %
                                       (err_msg))
                         if (pbs.event().type == pbs.EXECJOB_EPILOGUE
                                 and pbs.event().job.in_ms_mom()):
@@ -5703,6 +5713,16 @@ class CgroupUtils(object):
             if job_start < 0:
                 continue
             return kill_line
+
+        # Found nothing -- more recent kernels have different messages
+        for line in out_split:
+            start = line.find('oom-kill')
+            kill_line = line[start:]
+            job_start = line.find(jobid)
+            if job_start < 0:
+                continue
+            return kill_line
+
         return ''
 
     def write_job_env_file(self, jobid, env_list):
--- a/src/hooks/cgroups/pbs_cgroups.PY
+++ b/src/hooks/cgroups/pbs_cgroups.PY
@@ -1153,6 +1153,53 @@ class HookUtils(object):
                            'Error removing file: %s' % (filename))
         return True
 
+    def manage_rlimit_as(self, job):
+        """
+        Sets rlimit_as according to pvmem (if present)
+        if pvmem is not present, sets it to unlimited
+        (since we know vmem specifies memory+swap usage limit,
+         not address space limit)
+        """
+        import resource
+        rlimit_as = None
+        parent_pid = os.getppid()
+        if 'pvmem' in job.Resource_List and job.Resource_List['pvmem']:
+            rlimit_as = size_as_int(job.Resource_List['pvmem'])
+        else:
+            rlimit_as = resource.RLIM_INFINITY
+
+        prlimit_command = None
+        if hasattr(resource, 'prlimit'):
+            # No need to call command -- Python has direct support
+            prlimit_command = ''
+            pbs.logmsg(pbs.EVENT_DEBUG3,
+                       'Calling resource.prlimit(%s, resource.RLIMIT_AS, %s)'
+                       % (parent_pid, str((rlimit_as, rlimit_as))))
+            resource.prlimit(parent_pid, resource.RLIMIT_AS,
+                             (rlimit_as, rlimit_as))
+        elif os.path.isfile('/usr/bin/prlimit'):
+            prlimit_command = '/usr/bin/prlimit'
+        elif os.path.isfile('/bin/prlimit'):
+            prlimit_command = '/bin/prlimit'
+
+        if prlimit_command is None:
+            pbs.logmsg(pbs.EVENT_ERROR, "cannot set rlimit_as for task: "
+                       "prlimit command not found")
+        elif prlimit_command != '':
+            cmd = [prlimit_command,
+                   '--as=' + str(rlimit_as) + ':' + str(rlimit_as),
+                   '--pid=' + str(parent_pid)]
+            pbs.logmsg(pbs.EVENT_DEBUG3, 'Running: ' + ' '.join(cmd))
+            try:
+                # Try running the prlimit command
+                process = subprocess.Popen(cmd, shell=False,
+                                           stdout=subprocess.PIPE,
+                                           stderr=subprocess.PIPE)
+                out = process.communicate()[0]
+            except Exception:
+                pbs.logmsg(pbs.EVENT_DEBUG4, 'Failed to execute: %s' %
+                           ' '.join(cmd))
+
     def _execjob_launch_handler(self, event, cgroup, jobutil):
         """
         Handler for execjob_launch events.
@@ -1173,6 +1220,13 @@ class HookUtils(object):
                        (cgroup.assigned_resources))
             if "gpu" in node.devices:
                 cgroup.setup_job_devices_env(node.devices['gpu'])
+
+        # If vmem was requested, set per process address space limit
+        # or clear the one MoM has set,
+        # if possible (requires prlimit support)
+        if (cgroup.cfg['manage_rlimit_as']
+                and cgroup.cfg['cgroup']['memsw']['enabled']):
+            self.manage_rlimit_as(pbs.event().job)
         return True
 
     def _exechost_periodic_handler(self, event, cgroup, jobutil):
@@ -3199,6 +3253,7 @@ class CgroupUtils(object):
         defaults['job_setup_timeout'] = 30
         defaults['placement_type'] = 'load_balanced'
         defaults['propagate_vntype_to_server'] = True
+        defaults['manage_rlimit_as'] = True
         defaults['cgroup'] = {}
         defaults['cgroup']['cpu'] = {}
         defaults['cgroup']['cpu']['enabled'] = False
@@ -3518,7 +3573,7 @@ class CgroupUtils(object):
         pbs.logmsg(pbs.EVENT_DEBUG4, '%s: Method called' % caller_name())
         assigned = {}
 
-        # Get totals available for all jobs to compare with later        
+        # Get totals available for all jobs to compare with later
         mem_avail = None
         filename = self._cgroup_path('memory', 'limit_in_bytes')
         if filename and os.path.isfile(filename):
@@ -4968,6 +5023,7 @@ class CgroupUtils(object):
                                'scheduler configuration file' %
                                caller_name())
                     hostresc['vmem'] = pbs.size(vmem_limit)
+
         # Initialize hpmem variables
         hpmem_default = None
         hpmem_limit = None
--- a/src/hooks/cgroups/pbs_cgroups.PY
+++ b/src/hooks/cgroups/pbs_cgroups.PY
@@ -1183,7 +1183,7 @@ class HookUtils(object):
             prlimit_command = '/bin/prlimit'
 
         if prlimit_command is None:
-            pbs.logmsg(pbs.EVENT_ERROR, "cannot set rlimit_as for task: "
+            pbs.logmsg(pbs.EVENT_DEBUG, "cannot set rlimit_as for task: "
                        "prlimit command not found")
         elif prlimit_command != '':
             cmd = [prlimit_command,
@@ -1197,7 +1197,8 @@ class HookUtils(object):
                                            stderr=subprocess.PIPE)
                 out = process.communicate()[0]
             except Exception:
-                pbs.logmsg(pbs.EVENT_DEBUG4, 'Failed to execute: %s' %
+                pbs.logmsg(pbs.EVENT_ERROR,
+                           'Found but to failed to execute: %s' %
                            ' '.join(cmd))
 
     def _execjob_launch_handler(self, event, cgroup, jobutil):
@@ -2897,7 +2898,7 @@ class CgroupUtils(object):
                 filename = job.stdout_file()
             else:
                 filename = job.stderr_file()
-            pbs.logmsg(pbs.EVENT_DEBUG4,"Message to be written to %s: %s"
+            pbs.logmsg(pbs.EVENT_DEBUG4, "Message to be written to %s: %s"
                        % (str(filename), msg.strip()))
             if filename is None:
                 return
--- a/src/hooks/cgroups/pbs_cgroups.PY
+++ b/src/hooks/cgroups/pbs_cgroups.PY
@@ -2203,8 +2203,15 @@ class NodeUtils(object):
                        caller_name())
         cpuinfo['physical_cpus'] = int(cpuinfo['logical_cpus']
                                        // cpuinfo['hyperthreads_per_core'])
-        pbs.logmsg(pbs.EVENT_DEBUG4, '%s returning: %s' %
-                   (caller_name(), cpuinfo))
+        wanted_keys = ['physical_cpus', 'logical_cpus','hyperthreads_per_core',
+                       'hyperthreads']
+        cpuinfo_short = dict((k, cpuinfo[k])
+                             for k in wanted_keys if k in cpuinfo)
+        pbs.logmsg(pbs.EVENT_DEBUG4, "%s returning: %s"
+                   % (caller_name(), cpuinfo_short))
+        if 0 in cpuinfo['cpu']:
+            pbs.logmsg(pbs.EVENT_DEBUG4, "%s For CPU 0: cpuinfo['cpu'][0]: %s"
+                       % (caller_name(), cpuinfo['cpu'][0]))
         return cpuinfo
 
     def gather_jobs_on_node(self, cgroup):
--- a/src/hooks/cgroups/pbs_cgroups.PY
+++ b/src/hooks/cgroups/pbs_cgroups.PY
@@ -344,7 +344,13 @@ def size_as_int(value):
     """
     Convert a size string to an integer representation of size in bytes
     """
-    return int(convert_size(value).rstrip(string.ascii_lowercase))
+    in_bytes = convert_size(value, units='b')
+    if in_bytes is not None:
+        return int(convert_size(value).rstrip(string.ascii_lowercase))
+    else:
+        pbs.logmsg(pbs.EVENT_ERROR, "size_as_int: Value %s "
+                   "does not convert to int, returning None" % value)
+        return None
 
 
 #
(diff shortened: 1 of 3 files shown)


alexis-cousein:
Test also passed on Test Harness, on CentOS7, CentOS8 and SLES15.

6752 | Custom | Tests from given sources on platforms CENTOS7, CENTOS8, SUSE15 | alexis-cousein/pbspro@cgroups_rlimit_as | 26th Mar 2021 05:35:50.619 PM +01:00 | 00:07:24.458 | 3 | 0 | 0 | 0 | 0 | 0 | 0 | 3

When rerunning all cgroup tests I see there's still a race condition in the test to see if stuck nodes are correctly offlined, because the job that should get stuck can still be deleted before the processes are frozen. I'm going to implement a more robust test to confirm the freezer has frozen the processes (instead of "if I sleep 1 second it should be OK").

Stay tuned for another commit.

alexis-cousein:
It will also be necessary to rewrite test_cgroup_enforce_memsw, since the 'success' criterion depends on the bug worked around by manage_rlimit_as! It currently tests on whether Python was stopped because it ran out of address space, rather than because the memsw limit was enforce.

alexis-cousein:
Since the Test Harness has no swap and will not run it, the relevant output for the fixed test_cgroup_enforce_memsw test on my system:

```
2021-03-27 01:45:11,433 INFO     Current config: {
    "exclude_hosts"         : [],
    "exclude_vntypes"       : [],
    "run_only_on_hosts"     : [],
    "periodic_resc_update"  : true,
    "vnode_per_numa_node"   : false,
    "online_offlined_nodes" : true,
    "use_hyperthreads"      : true,
    "cgroup":
    {
        "cpuacct":
        {
            "enabled"         : true,
            "exclude_hosts"   : [],
            "exclude_vntypes" : []
        },
        "cpuset":
        {
            "enabled"         : true,
            "exclude_hosts"   : [],
            "exclude_vntypes" : []
        },
        "devices":
        {
            "enabled"         : false
        },
        "hugetlb":
        {
            "enabled"         : false
        },
        "memory":
        {
            "enabled"         : true,
            "default"         : "96MB",
            "reserve_amount"  : "50MB",
            "exclude_hosts"   : [],
            "exclude_vntypes" : []
        },
        "memsw":
        {
            "enabled"         : true,
            "default"         : "96MB",
            "reserve_amount"  : "45MB",
            "exclude_hosts"   : [],
            "exclude_vntypes" : []
        }
    }
}

2021-03-27 01:45:18,435 INFO     manager on dragon: import hook pbs_cgroups {'content-type': 'application/x-config', 'content-encoding': 'default', 'input-file': '/tmp/PtlPbspszxq2fy'}
2021-03-27 01:45:18,435 INFOCLI  dragon: sudo -H /opt/pbs/bin/qmgr -c import hook pbs_cgroups application/x-config default /tmp/PtlPbspszxq2fy
2021-03-27 01:45:18,453 ERROR    err: ["hook 'pbs_cgroups' contents overwritten"]
2021-03-27 01:45:18,558 INFO     mom dragon log match: searching for "pbs_cgroups.CF;copy hook-related file request received" - with existence... OK
2021-03-27 01:45:18,558 INFOCLI2 dragon: sudo -H cat /var/spool/PBS/server_priv/hooks/pbs_cgroups.CF
2021-03-27 01:45:18,570 INFOCLI2 dragon: sudo -H cat /var/spool/PBS/mom_priv/hooks/pbs_cgroups.CF
2021-03-27 01:45:18,581 INFO     server & mom pbs_cgroups.CF match
2021-03-27 01:45:22,585 INFO     mom dragon: sent signal -HUP
2021-03-27 01:45:22,586 INFOCLI2 dragon: sudo -H /usr/bin/cat /var/spool/PBS/mom_priv/mom.lock
2021-03-27 01:45:22,598 INFOCLI2 dragon: sudo -H kill -HUP 713397
2021-03-27 01:45:22,715 INFO     mom dragon log match: searching for "hook_perf_stat;label=hook_exechost_startup_pbs_cgroups_.* profile_stop" - using regular expression  - with existence - No match
2021-03-27 01:45:23,825 INFO     mom dragon log match: searching for "hook_perf_stat;label=hook_exechost_startup_pbs_cgroups_.* profile_stop" - using regular expression  - with existence... OK
2021-03-27 01:45:23,825 INFOCLI2 dragon: sudo -H -u pbsuser /tmp/PtlPbsej6r65hs
Contents of /tmp/PtlPbsej6r65hs:
----------------------------------------
#!/bin/bash
/usr/bin/rm ~/CGROUP6.*
----------------------------------------
2021-03-27 01:45:23,838 INFOCLI2 dragon: /usr/bin/rm /tmp/PtlPbsej6r65hs
2021-03-27 01:45:23,840 INFO     job: executable set to /bin/sleep with arguments: 100
2021-03-27 01:45:23,840 INFOCLI2 dragon: which chmod
2021-03-27 01:45:23,841 INFOCLI2 dragon: /usr/bin/chmod 755 /tmp/PtlPbsJobScriptiwry015u
2021-03-27 01:45:23,843 INFOCLI  dragon: sudo -H -u pbsuser /opt/pbs/bin/qsub -l select=1:ncpus=1:mem=300mb:vmem=320mb:host=dragon -N CGROUP6 /tmp/PtlPbsJobScriptiwry015u
2021-03-27 01:45:23,876 INFO     submit to dragon as pbsuser: job 4465.dragon OrderedDict([('Resource_List.select', '1:ncpus=1:mem=300mb:vmem=320mb:host=dragon'), ('Job_Name', 'CGROUP6')])
2021-03-27 01:45:23,876 INFOCLI  job script /tmp/PtlPbsJobScriptiwry015u
---
#PBS -joe
#PBS -S /bin/bash
timeout -s KILL 60 sync
sleep 10
python_path=`which python 2>/dev/null`
python3_path=`which python3 2>/dev/null`
python2_path=`which python2 2>/dev/null`
if [ -z "$python_path" ]; then
    if [ -n "$python3_path" ]; then
        python_path=$python3_path
    else
        python_path=$python2_path
    fi
fi
if [ -z "$python_path" ]; then
    echo Exiting -- no python found
    exit 1
fi
$python_path - 80 10 10 <<EOF

import sys
import time
MB = 2 ** 20
iterations = 1
chunkSizeMb = 1
sleeptime = 0
if (len(sys.argv) > 1):
    iterations = int(sys.argv[1])
if (len(sys.argv) > 2):
    chunkSizeMb = int(sys.argv[2])
if (len(sys.argv) > 3):
    sleeptime = int(sys.argv[3])
if (iterations < 1):
    print('Iteration count must be greater than zero.')
    exit(1)
if (chunkSizeMb < 1):
    print('Chunk size must be greater than zero.')
    exit(1)
totalSizeMb = chunkSizeMb * iterations
print('Allocating %d chunk(s) of size %dMB. (%dMB total)' %
      (iterations, chunkSizeMb, totalSizeMb))
buf = ''
for i in range(iterations):
    print('allocating %dMB' % ((i + 1) * chunkSizeMb))
    buf += ('#' * MB * chunkSizeMb)
if sleeptime > 0:
    time.sleep(sleeptime)

EOF

---
2021-03-27 01:45:23,876 INFOCLI  dragon: /opt/pbs/bin/qstat -f 4465.dragon
2021-03-27 01:45:23,882 INFO     expect on server dragon: job_state = R && substate = 42 job 4465.dragon  got: substate = 41
2021-03-27 01:45:24,383 INFOCLI  dragon: /opt/pbs/bin/qmgr -c set server scheduling=True
2021-03-27 01:45:24,394 INFOCLI  dragon: /opt/pbs/bin/qstat -f 4465.dragon
2021-03-27 01:45:24,400 INFO     expect on server dragon: job_state = R && substate = 42 job 4465.dragon attempt: 2 ...  OK
2021-03-27 01:45:24,400 INFO     status on dragon: job 4465.dragon
2021-03-27 01:45:24,400 INFOCLI  dragon: /opt/pbs/bin/qstat -f 4465.dragon
2021-03-27 01:45:24,607 INFO     Reading file: /home/pbsuser/CGROUP6.o4465 on host: dragon
2021-03-27 01:45:37,120 INFOCLI2 dragon: sudo -H /usr/bin/cat /home/pbsuser/CGROUP6.o4465
2021-03-27 01:45:37,135 INFO     Joined stdout/stderr contained expected string: Cgroup memsw limit exceeded
```

alexis-cousein:
Note a cgroup hook fix was also necessary: for joined stderr/stdout, if the cgroup limit detection happens only in the execjob_epilogue, it is too late to append to stderr (it does not make it to the joined file), so we need to write in stdout if we detect Join_Path set to 'oe'.

That was indeed a valid 'test failure', because users who use this feature would otherwise never see the cgroup limit exceeded messages in what is supposed to be a joined stdout/stderr.

The dmesg searcher was also extended to support more recent kernels, which no longer have lines matching the old parser; for more recent kernels the lines added to stderr (or the joined stdout/stderr) are in the form:

```
Cgroup mem limit exceeded: oom-kill:constraint=CONSTRAINT_MEMCG,nodemask=(null),cpuset=4464.dragon,mems_allowed=0,oom_memcg=/pbs_jobs.service/jobid/4464.dragon,task_memcg=/pbs_jobs.service/jobid/4464.dragon,task=python,pid=711137,uid=4359
Cgroup memsw limit exceeded: oom-kill:constraint=CONSTRAINT_MEMCG,nodemask=(null),cpuset=4464.dragon,mems_allowed=0,oom_memcg=/pbs_jobs.service/jobid/4464.dragon,task_memcg=/pbs_jobs.service/jobid/4464.dragon,task=python,pid=711137,uid=4359
```
As before, they identify the limit for which failcnt was non-zero (here both limits were exceeded) , the cgroup (which validates it's about the correct job) and the victim selected by the OOM killer. More information can be found by site admins in dmesg output.

alexis-cousein:
The sibling test test_cgroup_enforce_memory is also incorrect: it uses a job that sets only a mem limit, and then uses a configuration file that enables "memory" and disables "memsw", leaving "swappiness" to the default 1. That will not make the job exceed the memory limit in all cases: if the machine has swap, once the job reaches its memory limit, it is allowed to swap out pages to swap (and since memsw is disabled there is no limit to it). On some kernels even when swap is depleted there may not even be a failcount set on the memory cgroup before the OOM killer wakes up because the swap is depleted -- so the job may get killed because of a "general" lack of swap, rather than a cgrou limit violation.

To avoid this, you can either enable memsw and set a vmem limit, or set swappiness to 0 in the configuration file. Since 
-there is already a separate test_cgroup_enforce_memsw for the former 
-_and_ since the latter also represents a typical configuration for sites that have machines with little disk swap (typically, you leave memsw disabled since there is no swap to manage, but set swappiness to 0 to ensure the OOM killer wakes up if the job cannot satisfy memory requests from physical memory)

the latter is preferable, and was implemented by setting up a slightly different configuration file.

alexis-cousein:
All tests that run without swap passed on CentOS7, CentOS8 and SLES15 on the test harness:

6763 | cgroups_cousein | Tests from given sources on platforms CENTOS7, SUSE15, CENTOS8 | alexis-cousein/pbspro@cgroups_rlimit_as | 27th Mar 2021 03:09:14.645 AM +01:00 | 00:32:36.812 | 159 | 0 | 0 | 0 | 0 | 0 | 39 | 120

159 run, 39 skipped (for lack of swap or a number of nodes that is too low), 120 successes.

All tests also pass on my machine which has swap enabled.

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -2649,8 +2742,29 @@ def cgroup_offline_node(self, name, vnpernuma=False):
...
+        for count in range(30):
+            ret = self.du.cat(hostname=self.hosts_list[0],
+                              filename=freezer_file,
+                              sudo=True)
+            if ret['rc'] != 0:
+                self.logger.info("Cannot confirm freezer state"
+                                 "sleeping 30 seconds instead")
+                confirmed_frozen = True
+                time.sleep(30)
+                break
```
bayucan, line 2756: I see a "break" here, after "Cannot confirm freezer state sleeping for 30 seconds instead". Shouldn't retry checking freezer file and not have the 'break'?

alexis-cousein: If for some reason the commands to confirm the state of the freezer don't work, the workaround is to wait 30 seconds and then bail out -- we don't know whether it's "enough" but it gives us at least some chance that the processes are frozen. The 30 seconds is the maximum we'd wait if the queries did not fail.

The problem is that we can't just skip the test when something goes wrong and be done with it. If we're flying blind and the freezer _is_ active, we can't just give up: it would poison the host for any subsequent PTL tests, until it is rebooted.

If you prefer we could time.sleep(1) and then not break, but if things are iffy I'd rather wait for as long as reasonable.

bayucan: So the way I see the code is if it can't confirm the frozen state as the 'cat freezer_file' returned a non-zero exit status, it would say confirm_frozen=true, and then sleep 30 seconds, and break, which means to not do a retry anymore of 'cat freezer_file' to see if we can get a better status. I just want to make sure that is what you intended.

test/tests/functional/pbs_cgroups_hook.py
```diff
@@ -2649,8 +2742,29 @@ def cgroup_offline_node(self, name, vnpernuma=False):
...
+                self.logger.info("job processes reported as FROZEN")
+                confirmed_frozen = True
+                break
+            else:
+                self.logger.info("freezer state reported as "
+                                 + ret['out'][0])
+                time.sleep(1)
+
+        if not confirmed_frozen:
+            self.logger.info("Freezer did not work -- skip test")
```
bayucan, line 2767: So when this is reached, somewhere along the line, the test will get skipped? There's no need for a self.skip_test(...) call?

alexis-cousein: As said before -- we MUST NOT just do a skipTest unless we're sure the freezer created is thawed.

bayucan: Got it.

test/tests/functional/pbs_cgroups_hook.py
alexis-cousein: Yes, that's what I intended. Except that we should probably set confirmed_frozen to False --  we can't decide whether the test failed or succeeded.

I'll change that once you confirm you finished your (first) review.

alexis-cousein: Update made.

bayucan (approved):
Looks good.

src/hooks/cgroups/pbs_cgroups.PY
```diff
@@ -1153,6 +1153,53 @@ class HookUtils(object):
...
+                   '--pid=' + str(parent_pid)]
+            pbs.logmsg(pbs.EVENT_DEBUG3, 'Running: ' + ' '.join(cmd))
+            try:
+                # Try running the prlimit command
+                process = subprocess.Popen(cmd, shell=False,
+                                           stdout=subprocess.PIPE,
+                                           stderr=subprocess.PIPE)
+                out = process.communicate()[0]
+            except Exception:
+                pbs.logmsg(pbs.EVENT_DEBUG4, 'Failed to execute: %s' %
```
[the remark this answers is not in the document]
alexis-cousein: Yes, because there are systems where this exception is normal (when they're using Python2 and there is no prlimit command), for which this change indeed fails to reset the rlimit_as set by MoM. 

I don't want to pollute the logs there too much unless the site admin wants to really see all messages, and many sites do show everything to DEBUG2.

Of course whether it is preferable to be at DEBUG3 or DEBUG4 is open for debate, but I usually err on the side of caution for messages that may be legitimate but look like 'failures'.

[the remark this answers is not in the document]
alexis-cousein: Changed log level of some messages, made hook pycodestyle-clean again.

alexis-cousein:
Fixing a few other problems:
-cleanup was not accurate; if 'cleaning up' freezers, it would "rm" rather than rmdir the freezer and would also not clean up processes. The current cleanup thaws all freezers (including those not tied to jobs, e.g. the PtlPbs freezer made by the test_cgroup_offline_node test) but then kills processes in freezers tied to directories that are probably jobs (i.e. a number followed by . and something else) before doing an rmdir.
-Ditto for cleaning up other cgroups. There was no attempt at killing the processes before doing an rmdir, so if processes were lingering the cleanup function would fail to mop up everything. I've also made it resistant to future tests that would make sub-cgroups in the job cgroups.
-the PtlPbs freezer was madein the freezer root. We have no delegate control, so under some OSes systemd can move processes out of it into the (unfrozen) root freezer and clean it up because it "should not be there" (since we haven't registered a service or slice with systemd), making the test fail. There was also an error in error handling in that case (which I had not seen since the freezer worked on my system) that was throwing an exception (hardly the thing you want when cleaning up a freezer!) and that was corrected.

alexis-cousein:
All tests passed or (correctly) skipped on test harness; now also on Ubunto 20.04:
PTL_results

alexis-cousein:
There were a number of issues with exception handling in the hook and in the tests.

1) 
The tests weren't correctly skipping some tests when self.swapctl was set to 'false' (because it's set to a string you can't use 'if not self.swapctl'), on nodes that have no kernel support for swap accounting (aka memsw). 

Corrected that in the tests, but also made sure that the hook would not fail with incomprehensible exceptions when memsw was enabled but kernel support for it was missing. I added guards around all the messages encountered, reinstalled some sanity checks for vmem_avail and vmem_default, and fixed one indentation issue when we were looking up memory.memsw.limit_in_bytes that would make the periodic hook fail when hosts have no swap accounting.

 Finally, the cgroup hook was modified to neuter this "impossible" configuration: if memsw is enabled in the config file but kernel support is lacking, memsw will be disabled and every hook event will print three messages in the logs at EVENT_ERROR level starting with "CONFIG ERROR:" telling the site something needs to be fixed.

2) 
one test was forcibly creating two vnodes on all hosts, named hostname[0] and hostname[1], but was also doing so when there is only one socket. Needless to say, configure_job does not take kindly to a request by the scheduler to assign memory on "hostname[1]", which it takes to be a request to assign it to the non-existing socket #1. The hook did crash with an ungainly exception (a KeyError on 1), so I modified the hook to fail with "Failed to assign resources" and a specific error message in the log.

The test, though, needs to be rewritten to let the cgroup hook create vnodes and work with what it gets. It now sets vnode_per_numa_node, lets MoM create the extra vnodes, checks for the presence of a vnode for socket [1] (and skips if not present), tallies the total memory available on all vnodes of the first host, and then creates a job that uses 1 CPU and every memory available except the two last 2mb, plus a job that uses 1024mb more (to ensure that the first job run was not successful because memory requests are ignored.

The test success criterion was also changed; it was looking for a "Can Never Run" message for insufficient mem, but if there is more than one node the message can also start with "Not Running" instead of that string, so it now just matches the "insufficient resource" part of the string.

alexis-cousein:
Also changed some tests that will be problematic on hosts that are not "real" VMs or hosts but containers. Some tests were using sync(), which if unfiltered is going to hang until the entire host kernel has flushed its buffer cache (with very uncertain timings, and creating a process that is unkillable while the sync is in progress and may mess up test cleanup).

I replaced it with a command to echo 3 to /proc/sys/vm/drop_caches, which on hosts and VMs will _ask_ to flush the caches (but in contrast with sync will not wait around and will not create an unkillable process), on privileged containers will _ask_ to flush the host caches, and on unprivileged containers will simply return an error (as /proc is not writeable on these containers).

nishiya (approved):
LGTM.  Please attach latest test results.

alexis-cousein:
The drop_caches turns out to be a bad idea -- all it does on recent kernels is mark clean pages free, but if the tunables of the host OS are sane (i.e. vm.min_free_kbytes is not too low) that's pointless -- it's the dirty pages that are hard to get rid of..

alexis-cousein:
Tests on test harness on all the OSes I could find all either passed or skipped.

Description: Rerun All Tests From #6905
Profile: cgroups_cousein
|--|--|--|--|--|--|--|
|6910|424|0|0|0|105|319|

nishiya:
Thanks for posting test results, looks good.
Need a maintainer review.

alexis-cousein:
the rlimit_as test was not robust to hosts that add extra lmines of spam in job outputs; that was changed. 

New test results on the Test Harness:

Profile: cgroups_cousein
|--|--|--|--|--|--|--|
|6957|424|0|0|0|104|320|

bayucan (approved):
Looks good.

bayucan:
@nishiya : All the tests are passing now after the latest commit from Alexis. I'll wait until you've seen the latest commit before I merge this in.

nishiya (approved):
LGTM

bayucan:
Merged #2332 into master.

PR #2338, opened 2021-04-01
Tests having "mom_on_server=True" and _no_mom_on_server=true" are not behaving as expected if mom is on remote host

birbalsain:
#### Describe Bug or Feature
Test having "no_mom_on_server=True" are skipping even mom is on remote host and "mom_on_server=True" should be skipped when mom is on remote host but tests are failing. Because the set holding the mom hostname is updated with server hostname.

#### Describe Your Change
In _cleanup(), used list to store server and mom hostname. because when we assign one set to other set and other set is updated then both set will be updated. so here mom hostname set is updated with server hostname

**test having "mom_on_Server=True"**
Before the changes
Profile: regression
|--|--|--|--|--|--|--|
|6839|12|12|0|0|0|0|

After the changes
Description: Rerun All Tests From #6839
Profile: regression
|--|--|--|--|--|--|--|
|6843|12|0|0|0|12|0|

**Tests having "no_mom_on_server=True"**
Before the changes
Description: Tests from given build on platforms UBUNTU1804, SUSE15
Platforms: UBUNTU1804,SUSE15
Profile: regression
|--|--|--|--|--|--|--|
|6835|7|0|0|0|7|0|

After the changes
Profile: regression
|--|--|--|--|--|--|--|
|6841|4|1|0|0|1|2|

### DIFF
--- a/test/fw/ptl/utils/plugins/ptl_test_runner.py
+++ b/test/fw/ptl/utils/plugins/ptl_test_runner.py
@@ -1041,7 +1041,7 @@ class PTLTestRunner(Plugin):
         hosts = set(self.param_dict['moms'])
         for server in self.param_dict['servers']:
             if server not in hosts:
-                hosts.extend(server)
+                hosts.update(server)
         for user in PBS_USERS:
             self.logger.debug('Cleaning %s\'s home directory' % (str(user)))
             runas = PbsUser.get_user(user)
--- a/test/fw/ptl/utils/plugins/ptl_test_runner.py
+++ b/test/fw/ptl/utils/plugins/ptl_test_runner.py
@@ -1038,10 +1038,10 @@ class PTLTestRunner(Plugin):
     def _cleanup(self):
         self.logger.info('Cleaning up temporary files')
         du = DshUtils()
-        hosts = self.param_dict['moms']
+        hosts = list(self.param_dict['moms'])
         for server in self.param_dict['servers']:
             if server not in self.param_dict['moms']:
-                hosts.update(self.param_dict['servers'])
+                hosts.extend(self.param_dict['servers'])
         for user in PBS_USERS:
             self.logger.debug('Cleaning %s\'s home directory' % (str(user)))
             runas = PbsUser.get_user(user)
--- a/test/fw/ptl/utils/plugins/ptl_test_runner.py
+++ b/test/fw/ptl/utils/plugins/ptl_test_runner.py
@@ -1038,10 +1038,10 @@ class PTLTestRunner(Plugin):
     def _cleanup(self):
         self.logger.info('Cleaning up temporary files')
         du = DshUtils()
-        hosts = list(self.param_dict['moms'])
+        hosts = set(self.param_dict['moms'])
         for server in self.param_dict['servers']:
-            if server not in self.param_dict['moms']:
-                hosts.extend(self.param_dict['servers'])
+            if server not in hosts:
+                hosts.extend(server)
         for user in PBS_USERS:
             self.logger.debug('Cleaning %s\'s home directory' % (str(user)))
             runas = PbsUser.get_user(user)


visheshh (approved):
LGTM

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -1038,10 +1038,10 @@ def addSuccess(self, test):
     def _cleanup(self):
         self.logger.info('Cleaning up temporary files')
         du = DshUtils()
-        hosts = self.param_dict['moms']
+        hosts = list(self.param_dict['moms'])
```
hirenvadalia, line 1041: Why are u converting `hosts` from `set` to `list`? we need it as set as we want to keep unique keys in it.

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -1038,10 +1038,10 @@ def addSuccess(self, test):
...
         for server in self.param_dict['servers']:
             if server not in self.param_dict['moms']:
-                hosts.update(self.param_dict['servers'])
+                hosts.extend(self.param_dict['servers'])
```
hirenvadalia, line 1044: Python's `set` doesn't have extend, so `.update` was right, but there is bug in old (as well as) new like is if `server` not in `moms` set then we add whole `servers` set to `hosts`, where we need to add only `server` in `hosts`

So this lines should be like:
```python
hosts = self.param_dict['moms']
for server in self.param_dict['servers']:
    if server not in hosts:
        hosts.update(server)
```

hirenvadalia (changes requested):

test/fw/ptl/utils/plugins/ptl_test_runner.py
birbalsain: Because if we use hosts = self.param_dict['moms'] and in later step we do hosts.update(server) then not only hosts updated with server but also origin self.param_dict['moms']. that is the set behavior. 

We can get unique value by using if server not in hosts:

hirenvadalia: for that you just need to create copy of self.param_dict['moms'] using `hosts = set(self.param_dict['moms'])`

birbalsain:
no_mom_on_server=True is  passed after changes [after_changes.txt](https://github.com/openpbs/openpbs/files/6283483/after_changes.txt)

test/fw/ptl/utils/plugins/ptl_test_runner.py
```diff
@@ -1038,10 +1038,10 @@ def addSuccess(self, test):
     def _cleanup(self):
         self.logger.info('Cleaning up temporary files')
         du = DshUtils()
-        hosts = self.param_dict['moms']
+        hosts = set(self.param_dict['moms'])
         for server in self.param_dict['servers']:
-            if server not in self.param_dict['moms']:
-                hosts.update(self.param_dict['servers'])
+            if server not in hosts:
+                hosts.extend(server)
```
hirenvadalia, line 1044: Python `set` doesn't have `.extend()`:
```
>>> t = set()
>>> t.extend('a')
AttributeError: 'set' object has no attribute 'extend'
>>> 
```

So your newly modified changes are not getting tested in whatever test your are running. Means you need to run test with `-pmoms=<moms on remote host>,servers=<server on host other than mom>,nomom=<server hostname>` to `pbs_benchpress`

hirenvadalia (changes requested):

hirenvadalia:
@birbalsain please add test results with latest changes

birbalsain:
[after_changes_no_mom.txt](https://github.com/openpbs/openpbs/files/6285187/after_changes_no_mom.txt)

birbalsain:
mom_on_server=True tests are skipped while mom is not on the server, previously failed
Profile: regression
|--|--|--|--|--|--|--|
|6983|12|0|0|0|12|0|

hirenvadalia (approved):

PR #2341, opened 2021-04-06
test fix: node state check before qdel post server restart

saritakh:
#### Describe Bug or Feature

Qdel command is issued in the same second when the server mom connection is resumed; hence qdel fails. 

#### Describe Your Change

Fixed by inserting node state check before issuing qdel command.

Before:
[logfile_ERROR.txt](https://github.com/openpbs/openpbs/files/6264431/logfile_ERROR.txt)

After:
[logfile_PASS.txt](https://github.com/openpbs/openpbs/files/6264435/logfile_PASS.txt)


visheshh:
Hi @saritakh , shouldn't the check be part of svr kill and restart method to make sure svr establishes connection with all daemons?

saritakh:
> Hi @saritakh , shouldn't the check be part of svr kill and restart method to make sure svr establishes connection with all daemons?

ya Vishesh, sounds correct. It would avoid race condition failures of any other tests as well. Will update it, Thanks.

visheshh (approved):
LGTM

test/tests/functional/pbs_job_array.py
```diff
@@ -133,6 +133,7 @@ def kill_and_restart_svr(self):
         rv = self.is_server_licensed(self.server)
         _msg = 'No license found on server %s' % (self.server.shortname)
         self.assertTrue(rv, _msg)
+        self.mom.isUp()
```
bhroam, line 136: isUp() returns True/False if the mom is up or not.  If the mom fails to come up, just waits some time and returns false.  Since you are not checking for that, it might not work.  I'd put this in a self.assertTrue()

saritakh: Hi @bhroam While working on this, I realized that this case does not involve mom restart; but involves only server restart. Hence, node state check would suffice here and check for mom pid is not necessary. I have updated the same in the code. Please have a look and let me know your view on this.

saritakh:
@bhroam A gentle reminder for review of the latest changes.

bhroam (approved):
LGTM

PR #2344, opened 2021-04-08
multi-server: enhancing client commands

nithinj:
#### Describe Bug or Feature

* pbsnodes -av, and possibly other client commands don't throw any error message letting users know that a server is down and the data reported might be incomplete. It should behave similarly to qstat which does the following:

[ravi@pbspro ~]$ qstat

qstat: cannot connect to server pbspro

000.pbspro        STDIN            ravi                     0 Q workq 

* pbs_release_nodes command does not work with multi-server
* pbs_movejob does not work with multi-server.
* pbs_msgjob does not work with multi-server.
* pbs_orderjob does not work with multi-server.
* pbs_rerunjob does not work with multi-server
* export from init script masks PSI value for daemons.

[test.log](https://github.com/openpbs/openpbs/files/6304478/test.log)

#### Describe Your Change

* pbsnodes and other commands will throw an error if it cannot connect to one of the instances.
* pbs_release_node should iterate through instance fds when a virtual fd is provided.
* pbs_movejob should iterate through instance fds when a virtual fd is provided.
* pbs_msgjob should iterate through instance fds when a virtual fd is provided.
* pbs_orderjob should iterate through instance fds when a virtual fd is provided. It only works when both the jobid belongs to the same server.
* pbs_rerunjob should iterate through instance fds when a virtual fd is provided.
* set PSI while executing command instead of exporting

[test.log](https://github.com/openpbs/openpbs/files/6304479/test.log)

Regression Results:
Profile: regression
|--|--|--|--|--|--|--|
|7037|2023|6|0|0|0|2017|

Description: Rerun Only Failed Tests From #7037
Profile: regression
|--|--|--|--|--|--|--|
|7039|6|1|0|0|0|5|

### DIFF
--- a/src/cmds/pbs_ralter.c
+++ b/src/cmds/pbs_ralter.c
@@ -294,7 +294,7 @@ main(int argc, char *argv[], char *envp[])		/* pbs_ralter */
 			pbs_server, pbs_errno);
 		CS_close_app();
 		exit(pbs_errno);
-	} else
+	} else if (pbs_errno)
 		show_svr_inst_fail(connect, argv[0]);
 
 	pbs_strncpy(resv_id, argv[optind], sizeof(resv_id));
--- a/src/cmds/pbs_ralter.c
+++ b/src/cmds/pbs_ralter.c
@@ -294,7 +294,8 @@ main(int argc, char *argv[], char *envp[])		/* pbs_ralter */
 			pbs_server, pbs_errno);
 		CS_close_app();
 		exit(pbs_errno);
-	}
+	} else
+		show_svr_inst_fail(connect, argv[0]);
 
 	pbs_strncpy(resv_id, argv[optind], sizeof(resv_id));
 	if (get_server(resv_id, resv_id_out, server_out)) {
--- a/src/cmds/pbs_rdel.c
+++ b/src/cmds/pbs_rdel.c
@@ -127,7 +127,7 @@ main(int argc, char **argv, char **envp)
 				pbs_server, pbs_errno);
 			any_failed = pbs_errno;
 			continue;
-		} else
+		} else if (pbs_errno)
 			show_svr_inst_fail(connect, argv[0]);
 
 		stat = pbs_delresv(connect, resv_id_out, dest_queue);
--- a/src/cmds/pbs_rdel.c
+++ b/src/cmds/pbs_rdel.c
@@ -120,14 +120,15 @@ main(int argc, char **argv, char **envp)
 			any_failed = 1;
 			continue;
 		}
-		connect = cnt2server(server_out);
 
+		connect = cnt2server(server_out);
 		if (connect <= 0) {
 			fprintf(stderr, "pbs_rdel: cannot connect to server %s (errno=%d)\n",
 				pbs_server, pbs_errno);
 			any_failed = pbs_errno;
 			continue;
-		}
+		} else
+			show_svr_inst_fail(connect, argv[0]);
 
 		stat = pbs_delresv(connect, resv_id_out, dest_queue);
 		if (stat) {
--- a/src/include/libpbs.h
+++ b/src/include/libpbs.h
@@ -138,7 +138,6 @@ pthread_mutex_t * get_conn_mutex(int);
 #define SVR_CONN_STATE_DOWN 0
 #define SVR_CONN_STATE_UP 1
 
-#define NSVR pbs_conf.pbs_num_servers
 int get_num_servers(void);
 
 /* max number of preempt orderings */
--- a/src/include/libpbs.h
+++ b/src/include/libpbs.h
@@ -138,6 +138,7 @@ pthread_mutex_t * get_conn_mutex(int);
 #define SVR_CONN_STATE_DOWN 0
 #define SVR_CONN_STATE_UP 1
 
+#define NSVR pbs_conf.pbs_num_servers
 int get_num_servers(void);
 
 /* max number of preempt orderings */
--- a/src/lib/Libifl/pbsD_connect.c
+++ b/src/lib/Libifl/pbsD_connect.c
@@ -532,7 +532,6 @@ static bool
 part_of_cluster(char *svrhost, uint port, svr_conn_t **svr_conns)
 {
 	int i;
-	int nsvrs = get_num_servers();
 
 	if (!svrhost)
 		return true;
@@ -540,7 +539,7 @@ part_of_cluster(char *svrhost, uint port, svr_conn_t **svr_conns)
 	if (is_same_host(svrhost, pbs_default()) && port == pbs_conf.batch_service_port)
 		return true;
 
-	for (i = 0; i < nsvrs; i++) {
+	for (i = 0; i < NSVR; i++) {
 		if (is_same_host(svrhost, pbs_conf.psi[i].name) &&
 		    port == pbs_conf.psi[i].port)
 			return true;
@@ -567,7 +566,6 @@ connect_to_servers(char *svrhost, uint port, char *extend_data)
 	int fd = -1;
 	svr_conns_list_t *new_conns = create_conn_svr_instances();
 	svr_conn_t **svr_conns;
-	int nsvrs = get_num_servers();
 
 	if (new_conns == NULL)
 		return -1;
@@ -585,7 +583,7 @@ connect_to_servers(char *svrhost, uint port, char *extend_data)
 	}
 
 	/* Try to connect to all servers in the cluster */
-	for (i = 0; i < nsvrs; i++) {
+	for (i = 0; i < NSVR; i++) {
 		svr_conns[i] = add_instance(pbs_conf.psi[i].name, pbs_conf.psi[i].port);
 		if (!svr_conns[i])
 			goto err;
@@ -634,7 +632,6 @@ __pbs_connect_extend(char *server, char *extend_data)
 	unsigned int server_port;
 	char	*altservers[2];
 	int	have_alt = 0;
-	int	nsvrs;
 	int	sock = -1;
 	int	i;
 	int	f;
@@ -656,9 +653,7 @@ __pbs_connect_extend(char *server, char *extend_data)
 		return -1;
 	}
 
-	nsvrs = get_num_servers();
-
-	if (nsvrs == 1 && pbs_conf.pbs_primary && pbs_conf.pbs_secondary) {	
+	if (NSVR == 1 && pbs_conf.pbs_primary && pbs_conf.pbs_secondary) {	
 		/* failover configuered ...   */	
 		if (is_same_host(server, pbs_conf.pbs_primary)) {	
 			have_alt = 1;	
@@ -695,7 +690,7 @@ __pbs_connect_extend(char *server, char *extend_data)
 			break; 
 	}
 	
-	if (nsvrs > 1)
+	if (NSVR > 1)
 		return sock;
 	
 	if (i >= (have_alt+1) && sock == -1) {
@@ -1235,7 +1230,7 @@ pbs_register_sched(const char *sched_id, int primary_conn_id, int secondary_conn
 	if (svr_conns_secondary == NULL)
 		return -1;
 
-	for (i = 0; i < get_num_servers(); i++) {
+	for (i = 0; i < NSVR; i++) {
 		if (svr_conns_primary[i]->sd < 0 ||
 		    send_register_sched(svr_conns_primary[i]->sd, sched_id) != 0)
 			return -1;
@@ -1305,7 +1300,7 @@ server_name_to_fd(int c, char *svrname)
 		pbs_errno = PBSE_NOCONNECTS;
 		return -1;
 	}
-	for (i = 0; i < get_num_servers(); i++) {
+	for (i = 0; i < NSVR; i++) {
 		if (strcmp(conns[i]->name, svrname) == 0)
 			return conns[i]->sd;
 	}
@@ -1360,9 +1355,8 @@ int
 multi_svr_op(int fd)
 {
 	svr_conn_t **conns = get_conn_svr_instances(fd);
-	int num_svrs = get_num_servers();
 
-	if (conns == NULL || num_svrs == 1 || fd == conns[0]->sd)
+	if (conns == NULL || get_num_servers() == 1 || fd == conns[0]->sd)
 		return FALSE;
 
 	return TRUE;
--- a/src/lib/Libifl/pbsD_connect.c
+++ b/src/lib/Libifl/pbsD_connect.c
@@ -566,7 +566,7 @@ connect_to_servers(char *svrhost, uint port, char *extend_data)
 	int fd = -1;
 	svr_conns_list_t *new_conns = create_conn_svr_instances();
 	svr_conn_t **svr_conns;
-	int last_err;
+	int last_err = PBSE_NONE;
 
 	if (new_conns == NULL)
 		return -1;
@@ -1360,7 +1360,7 @@ multi_svr_op(int fd)
 {
 	svr_conn_t **conns = get_conn_svr_instances(fd);
 
-	if (conns == NULL || get_num_servers() == 1 || fd == conns[0]->sd)
+	if (conns == NULL || NSVR == 1 || fd == conns[0]->sd)
 		return FALSE;
 
 	return TRUE;
--- a/src/lib/Libifl/pbsD_connect.c
+++ b/src/lib/Libifl/pbsD_connect.c
@@ -539,7 +539,7 @@ part_of_cluster(char *svrhost, uint port, svr_conn_t **svr_conns)
 	if (is_same_host(svrhost, pbs_default()) && port == pbs_conf.batch_service_port)
 		return true;
 
-	for (i = 0; i < NSVR; i++) {
+	for (i = 0; i < get_num_servers(); i++) {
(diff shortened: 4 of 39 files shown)


src/cmds/pbs_ralter.c
```diff
@@ -294,7 +294,8 @@ main(int argc, char *argv[], char *envp[])		/* pbs_ralter */
 			pbs_server, pbs_errno);
 		CS_close_app();
 		exit(pbs_errno);
-	}
+	} else
+		show_svr_inst_fail(connect, argv[0]);
```
agrawalravi90, line 298: why are we showing that a server instance failed if connect > 0 ?

agrawalravi90: Now that I look at the function, I see that it checks whether any server instance was not successfully connected to. Can you please change the function name? It's kind of misleading. Maybe something like "check_msvr_conn_fail" ?

src/cmds/pbs_rdel.c
```diff
@@ -120,14 +120,15 @@ main(int argc, char **argv, char **envp)
...
 
+		connect = cnt2server(server_out);
 		if (connect <= 0) {
 			fprintf(stderr, "pbs_rdel: cannot connect to server %s (errno=%d)\n",
 				pbs_server, pbs_errno);
 			any_failed = pbs_errno;
 			continue;
-		}
+		} else
+			show_svr_inst_fail(connect, argv[0]);
```
agrawalravi90, line 131: How about moving the call to 'show_svr_inst_fail' inside cnt2server()? That way it'll be guaranteed to be called for every client command

src/include/libpbs.h
```diff
@@ -138,6 +138,7 @@ pthread_mutex_t * get_conn_mutex(int);
 #define SVR_CONN_STATE_DOWN 0
 #define SVR_CONN_STATE_UP 1
 
+#define NSVR pbs_conf.pbs_num_servers
```
agrawalravi90, line 141: I don't understand this. You are defining a pre-processor symbol as a variable name? Why?

src/server/node_manager.c
```diff
@@ -6581,10 +6581,10 @@ set_nodes(void *pobj, int objtype, char *execvnod_in, char **execvnod_out, char
 			    (pnode = find_alien_node(vname)) == NULL) {
 				if (svr_init && (pjob->ji_qs.ji_svrflags & JOB_SVFLG_RescUpdt_Rqd))
 					continue;
+				log_eventf(PBSEVENT_DEBUG, PBS_EVENTCLASS_SERVER, LOG_INFO,
+					   __func__, "Unknown node %s received", vname);
```
agrawalravi90, line 6585: Knowing the node name is good, but since this happens while parsing the select spec for a job, I think you should keep the class as PBS_EVENTCLASS_JOB and add the jobid back because the error has occurred wrt running a job.

agrawalravi90 (changes requested):

agrawalravi90:
Also, please post regression results

src/cmds/pbs_ralter.c
nithinj: It does more than just checking. The important function is to write to stderr upon failure. How about show_svr_inst_fail_if_any() ?

src/cmds/pbs_rdel.c
nithinj: We could have unless there are special checks in some clients which do not need this message to be printed. For instance, if a client is invoked in the quiet mode it is not expecting any message in stderr.

nithinj: and there could be client-specific error handling in the future than just printing a log to the console. we are not there yet.

src/include/libpbs.h
nithinj: This saves the need of storing the variable in a temporary variable for iterating. See eefd054884c68a853edac96874b5b5c2d6229da6

agrawalravi90: why not just use pbs_conf.pbs_num_servers directly then?

src/cmds/pbs_rdel.c
agrawalravi90: i didn't know we had a quiet mode :) The other thing we can do is have cnt2server return a specific PBSE_ code when one of the instances fail to connect, and then the clients can print that error in the if (connect <= 0) {} block itself. I think there's already such a code actually. What do you think?

nithinj: but connect will be >=0 if one of the instance gets connected.
Also, pbs_errno will get overridden with each new connection.

src/include/libpbs.h
nithinj: that is a lengthy one and NSVR is used very often within for loop and such...

agrawalravi90: As such, all of us use IDEs, I don't think long variable names are an issue, we have much longer pbs_conf variable names than this. But if you do want to shorten it, then just rename pbs_conf.pbs_num_servers instead of creating a level of indirection.

nithinj: I would prefer to keep NSVR if there are no issues with it.

src/cmds/pbs_rdel.c
agrawalravi90: here's what I'm suggesting:
- cnt2server_extend() calls a function to check if any of the instances failed, even though pbs_connect_extend returned > 0
- if an instance failed, cnt2server sets pbs_errno to a PBSE_ code which will tell clients that one of the instances is down. But it will still return the cluster fd, not -1
- Clients will then check if (connect < 0 || pbs_errno == PBSE_) {print connection error, if needed, exit from program}

That way cnt2server handles the check for any instance connection failures and client decides what to do with it.

src/cmds/pbs_ralter.c
agrawalravi90: no that's too ugly. See my suggestion for cnt2server(), if you agree then this function can just be for checking while the client does the printing of error if needed.

src/include/libpbs.h
agrawalravi90: As such, all of us use IDEs, I don't think long variable names are an issue, we have much longer pbs_conf variable names than this. But if you do want to shorten it, then just rename pbs_conf.pbs_num_servers instead of creating a level of indirection, or use a local variable like before.

src/lib/Libifl/pbsD_connect.c
```diff
@@ -1360,9 +1355,8 @@ int
 multi_svr_op(int fd)
 {
 	svr_conn_t **conns = get_conn_svr_instances(fd);
-	int num_svrs = get_num_servers();
 
-	if (conns == NULL || num_svrs == 1 || fd == conns[0]->sd)
+	if (conns == NULL || get_num_servers() == 1 || fd == conns[0]->sd)
```
agrawalravi90, line 1359: why get_num_servers() here instead of NSVR ?

agrawalravi90 (changes requested):
Minor comments. For the changes to commands, sorry to ask you for this, but can you please set up multi-server and post logs showing that these commands work with multi-server? Also, please do a regression.

src/lib/Libifl/pbsD_orderjo.c
```diff
@@ -65,21 +65,12 @@
  *
  */
 
-int
-__pbs_orderjob(int c, char *job1, char *job2, char *extend)
+static int
```
ashwathraop, line 68: update doxygen

src/lib/Libifl/pbsD_rerunjo.c
```diff
@@ -66,8 +66,8 @@
  *
  */
```
ashwathraop, line 67: Update doxygen

ashwathraop (changes requested):

src/cmds/pbs_ralter.c
nithinj: Right now we are more specific on which of the instance is down. We don't print the port number now but I am going to add that. It will be helpful in printing the specifics than leaving a generic connection error.

src/cmds/pbs_rdel.c
nithinj: > if needed, exit from program

I'm not sure whether I understand you correctly. That again requires another condition to be added which is similar to what we have now. So what is the advantage of this approach?
We don't want to exit from the program if one of the instances is connected.

To address the readability concerns, I'm calling show_svr_inst_failure only if pbs_errno is set. See: 97cc87e3b918de5101b19e658c98834c689940ed

src/include/libpbs.h
nithinj: There will not be an indirection as it is a macro.
The purpose of this change was to avoid creating a temporary variable every time you need to access frequently used variable like nsvr.

src/lib/Libifl/pbsD_connect.c
```diff
@@ -656,9 +653,7 @@ __pbs_connect_extend(char *server, char *extend_data)
 		return -1;
 	}
 
-	nsvrs = get_num_servers();
-
-	if (nsvrs == 1 && pbs_conf.pbs_primary && pbs_conf.pbs_secondary) {	
+	if (NSVR == 1 && pbs_conf.pbs_primary && pbs_conf.pbs_secondary) {
```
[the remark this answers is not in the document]
nithinj: > Minor comments. For the changes to commands, sorry to ask you for this, but can you please set up multi-server and post logs showing that these commands work with multi-server? Also, please do a regression.

I have attached the test logs in the PR description. Please have a look.

src/lib/Libifl/pbsD_movejob.c
```diff
@@ -49,7 +49,6 @@
 #include "dis.h"
 #include "pbs_ecl.h"
 
-
 /**
  * @brief
  *	send move job request
```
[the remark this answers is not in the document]
nithinj: Isn't that evident with static? Do I have to make any changes to Doxygen to indicate this?

src/lib/Libifl/pbsD_msgjob.c
```diff
@@ -52,7 +52,6 @@
 #include "dis.h"
 #include "pbs_ecl.h"
 
-
 /**
  * @brief
  *	-send the MessageJob request and get the reply.
```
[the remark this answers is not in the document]
nithinj: Isn't that evident with static? Do I have to make any changes to Doxygen to indicate this?

src/lib/Libifl/pbsD_msgjob.c
```diff
@@ -199,17 +258,70 @@ pbs_py_spawn(int c, char *jobid, char **argv, char **envp)
  * @retval	!0	error
```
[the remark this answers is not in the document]
nithinj: Isn't that evident with static? Do I have to make any changes to Doxygen to indicate this?

src/include/libpbs.h
agrawalravi90: what's wrong with a temporary variable? by indirection i was not talking about performance, i was talking about readability. This is totally unnecessary.

agrawalravi90:
@nithinj thanks for attaching logs. Just to confirm something:
- for pbs_release_nodes and qorder, the logs show an error. Does that mean that they need more work to be supported for multi-server?

src/cmds/pbs_ralter.c
```diff
@@ -294,7 +294,7 @@ main(int argc, char *argv[], char *envp[])		/* pbs_ralter */
 			pbs_server, pbs_errno);
 		CS_close_app();
 		exit(pbs_errno);
-	} else
+	} else if (pbs_errno)
```
agrawalravi90, line 297: just to confirm, will any pbs_errno be set when there's only 1 instance down and rest are up?

agrawalravi90 (changes requested):

src/include/libpbs.h
nithinj: The earlier temporary variable used was nsvr
That will now will be NSVR. which conveys the same meaning. How does it affect readability?

src/cmds/pbs_ralter.c
nithinj: no, if any of them are down. If all of them are down then it will hit the if case.

nithinj:
> @nithinj thanks for attaching logs. Just to confirm something:
> 
>     * for pbs_release_nodes and qorder, the logs show an error. Does that mean that they need more work to be supported for multi-server?

No, you are seeing the error in the log due to the wrong job state. This means the request lands on the server and the client got the response back. Without the proposed. changes, the request will not reach the server.

nithinj:
@agrawalravi90 & @ashwathraop 
Thanks for the review. I have addressed all your comments. Please have a look.

src/include/libpbs.h
agrawalravi90: you are creating a global reference to another global object just because you think it's too long, it will affect readability because it's not obvious why it's needed. Both of your reviewers are saying that they'd rather see pbs_conf.pbs_num_servers instead of NSVR everywhere, doesn't that tell you that your change might be affecting readability?

agrawalravi90:
@nithinj thanks. It does seem like qorder of jobs from different servers doesn't work, so I guess that needs more work, right?

the only other outstanding comment from me is 'NSVR'.

src/include/libpbs.h
nithinj: Let me try to explain:
This is one of the instances where we have to replace NSVR with pbs_conf.pbs_num_servers
`for (i = start, ct = 0; ct < NSVR; i = (i + 1) % NSVR, ct++) {`
=>
`for (i = start, ct = 0; ct < pbs_conf.pbs_num_servers; i = (i + 1) % pbs_conf.pbs_num_servers, ct++) {`
This is the readability issue this change is trying to address. This is not an isolated occurrence. Most of the lines changed for NSVR are similar.

Accessing data using a getter function is a good practice. If such a function has to be used within a loop and such and does not want the overhead of function invocation, then it has to be treated as an inline function or declare as a macro. This what is done here.

nithinj:
> @nithinj thanks. It does seem like qorder of jobs from different servers doesn't work, so I guess that needs more work, right?

yes, I have mentioned that in the PR description. That will have to be taken up separately.

ashwathraop (commented):
I dont have further comments. I will sign off once there is an agreement on the NSVR comment.

src/include/libpbs.h
agrawalravi90: if you wanted to replace pbs_conf.pbs_num_servers with a getter then you should have used get_num_servers(). NSVR is not a getter, it won't help when we move to C++ as we'll need a proper function, not macro.

It sounds like the only "readability issue" that you are trying to solve is that "pbs_conf.pbs_num_servers" is too long. So, I again suggest to either rename it to make it shorter, or use get_num_servers(). If you still disagree, can we take this up later? The fixes that this PR adds are urgent, we are spending a lot of time discussing refactoring that's not needed for those fixes.

nithinj: get_num_servers() was the one that we were using initially. We started using a temp variable as most of the usage of this variable falls within loops and to avoid the function invocation overhead. Any names which you can form by reducing pbs_conf.pbs_num_servers will be still long for its use within a loop. I can declare a new global called nsvr and assign the value to that. But the present approach is even simpler.

It addresses the following:
+ Smaller name
+ Better readability
+ Does not have to use a temp variable.

If you think it won't help in the future, let's take it up at that point. what say?

agrawalravi90: Sorry Nithin, I don't like this change, I don't think get_num_servers() will add any perceivable overhead (you are welcome to prove me wrong), I don't think the name is too long, and if it is then I don't see any issue with using local variables.

So, I'll again say, let's do this refactoring separately and we can involve more people and take their opinion. Since the fixes are urgent, I don't want to involve more folks as it'll delay the check-in.

nithinj: > I don't like this change

I read the whole thread again, the only point against the proposed change was that it adds a level of indirection. This is the case even if I revert back.

agrawalravi90: Ok, I guess we are at a stalemate, I request opinions from other @openpbs/openpbs-maintainers on Nithin's change. Once we have more opinions and reach an agreement, we can proceed with the check-in.

bhroam: OK, since @agrawalravi90 wants our opinion, I'll give one.  I actually agree with Ravi.  NSVR isn't a getter.  If at some point in the future we want to replace it with a real getter, then we'll have to rename it to be a function (NSVR->get_nsvr()).  I don't mind the use of a temp variable because the compiler will optimize it out.  It'll see that the only use is pbs_conf.pbs_num_servers and either optimize both into a register, or just optimize it out and use the pbs_conf address itself.  I'd vote for a combination of the use of a temp variable or not.  Use the temp variable when the line gets too long, and don't use one if it isn't too long.

nithinj:
@bhroam 
Thanks for the feedback.

@agrawalravi90 & @ashwathraop ,
I've replaced NSVR for get_num_servers() and used a temporary variable when the line gets too long. Please review ee46b710cdc44895eb19e0e055453aaec07bc938

ashwathraop (approved):
LGTM

ashwathraop:
Please see failed builds.

agrawalravi90 (approved):
LGTM, please see build failures and do one final regression

nithinj:
Please find the regression results:

Profile: regression
|--|--|--|--|--|--|--|
|7114|2023|2|0|0|1|2020|

PR #2346, opened 2021-04-09
multi-server: server local scheduling is broken

agrawalravi90:
#### Describe Bug or Feature

Server local scheduling seems to be broken. This was likely introduced by https://github.com/openpbs/openpbs/pull/2304 where we changed the location of calling init_msi() to be after setup_nodes(). This meant that inside setup_nodes() when we didn't set the node resource msvr_node_group correctly (it calls get_server_index(), which relies on multi-server metadata which is set by init_msi()). So, it gets set to -1, which means scheduler cannot do server local scheduling properly and it ends up running jobs on nodes which don't belong to the server.

#### Describe Your Change

moved the call to init_msi() to before setup_nodes(). Also, inside get_server_index(), we check to see if init_msi() was called, and if not then we call it to ensure that the metadata is set up before we return the server index.

### DIFF
--- a/src/server/multi_svr.c
+++ b/src/server/multi_svr.c
@@ -680,12 +680,13 @@ gen_svr_inst_id(void)
 	if (svr_inst_id == NULL) {
 		unsigned int svr_inst_port;
 		char svr_inst_name[PBS_MAXHOSTNAME + 1];
+		char svr_inst_fname[PBS_MAXHOSTNAME + 1];
 
 		if (gethostname(svr_inst_name, PBS_MAXHOSTNAME) == 0)
-			get_fullhostname(svr_inst_name, svr_inst_name, PBS_MAXHOSTNAME);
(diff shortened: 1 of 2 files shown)


nithinj (approved):
LGTM!

nithinj:
Won't the jobs get to run as the title mentions? Servers are supposed to run the jobs even if get non-local nodes using move+run or resc update. If the server does not have node-cache, it should get run in the next cycle.

agrawalravi90:
> Won't the jobs get to run as the title mentions? Servers are supposed to run the jobs even if get non-local nodes using move+run or resc update. If the server does not have node-cache, it should get run in the next cycle.

Thanks for clarifying, I didn't know that servers work off of a node-cache now, I've changed the PR title and description to focus on server local scheduling

arungrover (approved):

PR #2347, opened 2021-04-09
In revert_mom() allow time for mom to sync with server prior to create node.

vccardenas:
#### Describe Bug or Feature
During revert_mom(), intermittently observable on cpuset systems, the mom would fail to sync with the server before create node. The mom ends up with a corrupted pbs_cgroups.PY and the create node will fail to create the NUMA vnodes on cpuset systems.

#### Describe Your Change
Added a sleep after a mom restart or HUP before qmgr create node.

Note: Results show tests are skipped due to a separate existing issue.
[Test_Rrecord_with_resources_used.txt](https://github.com/openpbs/openpbs/files/6287220/Test_Rrecord_with_resources_used.txt)
[TestCgroupsHook.txt](https://github.com/openpbs/openpbs/files/6287221/TestCgroupsHook.txt)

test results after using sleep on cpuset systems only:
[th3_results.txt](https://github.com/openpbs/openpbs/files/6340478/th3_results.txt)
[jenkins_results.txt](https://github.com/openpbs/openpbs/files/6340479/jenkins_results.txt)

### DIFF
--- a/test/fw/ptl/utils/pbs_testsuite.py
+++ b/test/fw/ptl/utils/pbs_testsuite.py
@@ -1531,6 +1531,8 @@ class PBSTestSuite(unittest.TestCase):
             mom.signal('-HUP')
         if not mom.isUp():
             self.logger.error('mom ' + mom.shortname + ' is down after revert')
+        # give mom enough time to network sync with the server
+        time.sleep(4)
         a = {'state': 'free'}
         self.server.manager(MGR_CMD_CREATE, NODE, None, mom.shortname)
         if enabled_cpuset:
--- a/test/fw/ptl/utils/pbs_testsuite.py
+++ b/test/fw/ptl/utils/pbs_testsuite.py
@@ -1531,8 +1531,9 @@ class PBSTestSuite(unittest.TestCase):
             mom.signal('-HUP')
         if not mom.isUp():
             self.logger.error('mom ' + mom.shortname + ' is down after revert')
-        # give mom enough time to network sync with the server
-        time.sleep(4)
+        # give mom enough time to network sync with the server on cpuset system
+        if enabled_cpuset:
+            time.sleep(4)
         a = {'state': 'free'}
         self.server.manager(MGR_CMD_CREATE, NODE, None, mom.shortname)
         if enabled_cpuset:


test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -1531,6 +1531,8 @@ def revert_mom(self, mom, force=False):
             mom.signal('-HUP')
         if not mom.isUp():
             self.logger.error('mom ' + mom.shortname + ' is down after revert')
+        # give mom enough time to network sync with the server
+        time.sleep(4)
```
hirenvadalia, line 1535: @vccardenas We already have fix for avoiding corruption of CF/HK/PY files below in `if enabled_cpuset:` block (just after `self.server.manager()` line, then why we need this `sleep()`?

And if we need it for cpuset moms then I would suggest to put it under `if enabled_cpuset:` condition, that way it won't affect other platforms, else it will add 4s delay for each mom in every test, which will cost us huge amount time in regression where we run ~4k tests.

hirenvadalia (changes requested):

test/fw/ptl/utils/pbs_testsuite.py
bayucan: @hirenvadalia : the sleep delay is needed before the qmgr create node call to give mom enough time to network sync up with the server. just before making the actual create node calls and hook files start copying. I agree with you that we should wrap the sleep under the if enabled_cpuset condition. to: @vccardenas

vstumpf: Is there a log message we can use to match instead of just sleeping for 4 seconds?

bayucan: @vstumpf : Unfortunately,  I have not found any mom_logs message that will guarantee that mom and server are synced up. I really tried to find it, experimented, but all attempts failed. This is the best that we can do to make revert_to_mom() more robust.

hirenvadalia: @bayucan then why don't we introduce needed messages in mom logs?

bayucan: @hirenvadalia : It's hard to pinpoint where to put such a message in mom to ensure that mom and server are network synced. If server is restarted, mom gets the "Hello from server" message. But if mom is restarted, or HUP-ed, I'm not sure where to guarantee that sync, unless mom sends a special message to server and gets an ack from the server. I think mom sends a periodic update internally to the server, but I don't think it waits for an ack.
In my opinion, I don't think we need to be adding a special mom_logs just to help out PTL with this problem. This incomplete pbs_cgroups hook problem happens intermittently under cpuset systems, and appears to be a consequence of the frequent mom hups/restarts and hook copying in PTL. I believe this is not the normal thing that happens in actual sites. So adding this stable mom_logs message may only be useful to PTL and nothing else. 
I feel the sleep fix is the best we can do for now without having to touch mom code.
As a separate issue, we will look into having mom detecth a hook being incomplete and not execute it, in particular exechost_startup hook, and perhaps would ask the server for a fresh copy.not execute an incomplete (corrupted) exechost_startup hook, and instead, would request a server a fresh copy of the hook.

hirenvadalia: Well, my point for adding a message was not only for PTL, that message might be usefull for general debugging purpose also, but if you are saying its hard to pinpoint location to add message, then I am fine with sleep().

hirenvadalia (approved):

bayucan (approved):
Looks good.

bayucan:
Merged #2347 into master.

PR #2350, opened 2021-04-12
Test fix for TestPbsAccumulateRescUsed.test_mom_down & test_mom_down2 failures at job attribute verification

saritakh:
#### Describe Bug or Feature

TestPbsAccumulateRescUsed.test_mom_down and TestPbsAccumulateRescUsed.test_mom_down2 occasionally fail at job attribute verification test steps.

#### Describe Your Change

**test_mom_down:**
In the case of failure, the job of 100 seconds gets over in and around same time when the mom is stopped. Hence, the job is not found and no data got when queried for its attribute resources_used.foo_i. Hence fixing this by by increasing the walltime of this job to 300 seconds instead of 100 seconds.

**test_mom_down2:** 
In the case of failure, the Job finish timestamp is happening the same time when the sister mom is killed, whose killing as per test case should happen before job finishes. Hence, the job's walltime needs to be increased well enough to not get over before sister mom is killed. The resource accumulation at execjob_epilogue should happen only from 2 moms i.e. 9 (at mother superior) and 10 (at sister mom), but since the hook from sister mom to be killed also executes (because job finishes before mom is killed) the accumulation is 29 instead of 19.

Before:
[logfile_FAIL.txt](https://github.com/openpbs/openpbs/files/6297787/logfile_FAIL.txt)
[logfile_FAIL_2.txt](https://github.com/openpbs/openpbs/files/6297825/logfile_FAIL_2.txt)
After:
[logfile_PASS.txt](https://github.com/openpbs/openpbs/files/6297826/logfile_PASS.txt)
[logfile_PASS_2.txt](https://github.com/openpbs/openpbs/files/6297828/logfile_PASS_2.txt)

### DIFF
--- a/test/tests/functional/pbs_accumulate_resc_used.py
+++ b/test/tests/functional/pbs_accumulate_resc_used.py
@@ -951,6 +951,7 @@ for jj in e.job_list.keys():
             overwrite=True)
 
         a = {'Resource_List.select': '3:ncpus=1',
+             'Resource_List.walltime': 300,
              'Resource_List.place': 'scatter'}
         j = Job(TEST_USER)
         j.set_attributes(a)
--- a/test/tests/functional/pbs_accumulate_resc_used.py
+++ b/test/tests/functional/pbs_accumulate_resc_used.py
@@ -961,6 +961,7 @@ for jj in e.job_list.keys():
         a = {'Resource_List.select': '5:ncpus=1',
              'Resource_List.place': 'scatter'}
         j.set_attributes(a)
+        j.set_sleep_time("300")
         jid2 = self.server.submit(j)
 
         # Wait for 10s approx for hook to get executed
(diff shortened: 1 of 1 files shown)


kjakkali (approved):

test/tests/functional/pbs_accumulate_resc_used.py
```diff
@@ -951,6 +951,7 @@ def test_mom_down(self):
             overwrite=True)
 
         a = {'Resource_List.select': '3:ncpus=1',
+             'Resource_List.walltime': 300,
              'Resource_List.place': 'scatter'}
         j = Job(TEST_USER)
         j.set_attributes(a)
```
hirenvadalia, line 957: Don't you need to increase job's sleep time as well to 300? else even if u have job with 300s walltime still it will finish in 100s as by default job runs `/bin/sleep 100`

saritakh: I agree Hiren, thanks for catching it.

saritakh:
[logfile_PASS_3.txt](https://github.com/openpbs/openpbs/files/6335678/logfile_PASS_3.txt)

Thanks for the review @hirenvadalia , I request to take this PR forward.
Thanks @kjakkali for sign off.

hirenvadalia (approved):

PR #2352, opened 2021-04-12
[WIP] fix TPP mbox full error

subhasisb:
#### Describe Bug or Feature

TPP does some rudimentary amount of throttling, so that no one client can flood the pipes. The pipes are managed by the mbox data structures, which have max_size and a mbox_size. The mbox_size parameter was not being computed properly (not correctly reset to zero in certain cases). The other issue was that when receiver buffer was full, we had the logic to check the buffer at periodic intervals, but that did not work due to a missing socket fd in the comparison.

#### Link to Design Doc

- Set mbox_size (current size of the pipe) to 0 at some places where we clear the mbox
- passing conn->sock_fd to the deferred event enquing, since the fd is checked at the time of dequeuing (the fd was being set to -1 earlier)

Manual test procedure:
- Stand up a single host pbs installation
- Set the mom to have 1000 ncpus
- Turn off scheduling and submit 600 jobs
- Turn on scheduling
- The target receiver buffer will get full due to the quick TPP actions, but due to the bug will not get cleared
- This will result in obit acks no longer reaching the mom and many jobs will remain stuck in 41 substate

Profile: regression
|--|--|--|--|--|--|--|
|7032|2023|3|1|0|0|2019|

Master results are as follows:
Profile: regression
|--|--|--|--|--|--|--|
|7033|2023|2|0|0|0|2021|

### DIFF
--- a/src/lib/Libtpp/tpp_em.c
+++ b/src/lib/Libtpp/tpp_em.c
@@ -1083,6 +1083,7 @@ tpp_mbox_read(tpp_mbox_t *mbox, unsigned int *tfd, int *cmdval, void **data)
 
 	/* if no more data, clear all notifications */
 	if (cmd == NULL) {
+		mbox->mbox_size = 0;
 #ifdef HAVE_SYS_EVENTFD_H
 		read(mbox->mbox_eventfd, &u, sizeof(uint64_t));
 #else
@@ -1091,7 +1092,7 @@ tpp_mbox_read(tpp_mbox_t *mbox, unsigned int *tfd, int *cmdval, void **data)
 	} else {
 		/* reduce from mbox size during read */
 		mbox->mbox_size -= cmd->sz;
-#ifdef DEBUG
+#ifdef TPPDEBUG
 		if ((mbox->max_size != -1) && (cmd->sz > 0)) {
 			TPP_DBPRT("Mbox %s, after reading %d size = %d", mbox->mbox_name, cmd->sz, mbox->mbox_size);
 		}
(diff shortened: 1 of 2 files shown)


nithinj (approved):
LGTM!

ashwathraop (approved):

nithinj:
@subhasisb ,
Why a physical machine is required to reproduce this issue?

subhasisb:
> @subhasisb ,
> Why a physical machine is required to reproduce this issue?

It is possible that in usual container setup, TPP will get much less chances to be fast enough to produce this issue. In this particular case (fix part 2) the application (pbs_mom) is consuming slower than the number of messages arriving to it via TPP. Anyway, it is only a speculation right now - it might be possible to reproduce it. The point, though, is whether we can make a reliable automated test that can reliably catch the problem on most platforms.

agrawalravi90:
@subhasisb  changes look fine to me, please run a full regression before this can be merged, thanks.

hirenvadalia (approved):

agrawalravi90 (changes requested):
Blocking until Subhasis completes testing

subhasisb:
Description: Rerun Only Failed Tests From #7032
Profile: regression
|--|--|--|--|--|--|--|
|7034|4|1|1|0|0|2|

subhasisb:
The failed and error tests are also present on master

subhasisb:
closing this PR - i am updating the alternate approach, which is to remove the checks for mbox sizes, since it does not make much difference and is adding extra complexity (the previous versions of PBS did not have this check as well, so we won't miss any functionality)

PR #2356, opened 2021-04-14
Only attempt to extend job script directive lines

awellington:
PBS job scripts should only extend PBS directives in the script. Line
extensions in the rest of the script can change the meaning of the job
script. For example, a comment line in a shell script that ended with
'\' would cause the next line of the script to be commented out if the
line is processed with the line extension code.

#### Describe Bug or Feature

When submitting a job script PBS should not attempt to perform line extension on the non-directive lines (non-#PBS lines). For example:

```
#!/bin/sh
#PBS -m n
        
# This is a test comment that shouldn't be extended \
cat $0
```

By incorrectly performing line extension on the comment line above, the line below it in the script becomes part of the comment and is never executed.

#### Describe Your Change

We now read the job script with `pbs_fgets`, and while processing a PBS directive line (#PBS) we check if the line needs extension. To do the check, `pbs_extendable_line` has been factored out of `pbs_fgets_extend` so we can use the same logic as `pbs_fgets_extend` does.

If it's a line that requires extension, we read any further lines using `pbs_fgets_extend` and `pbs_strcat` them together for the final extended line that we should be processing.

[functional_test_output.txt](https://github.com/openpbs/openpbs/files/6307641/functional_test_output.txt)

### DIFF
--- a/src/cmds/qsub_sup.c
+++ b/src/cmds/qsub_sup.c
@@ -284,9 +284,7 @@ get_script(FILE *file, char *script, char *prefix)
 			 */
 			if (do_dir(sopt, CMDLINE - 1, retmsg, MAXPATHLEN) != 0) {
 				fprintf(stderr, "%s", retmsg);
-				if (extend_in != NULL) {
-					free(extend_in);
-				}
+				free(extend_in);
 				free(s_in);
 				return (-1);
 			}
@@ -298,17 +296,13 @@ get_script(FILE *file, char *script, char *prefix)
 			fprintf(stderr, "qsub: error writing copy of script, %s\n",
 				tmp_name);
 			fclose(TMP_FILE);
-			if (extend_in != NULL) {
-				free(extend_in);
-			}
+			free(extend_in);
 			free(s_in);
 			return (3);
 		}
 	}
 
-	if (extend_in != NULL) {
-		free(extend_in);
-	}
+	free(extend_in);
 	free(s_in);
 	if (fclose(TMP_FILE) != 0) {
 		perror(" qsub: copy of script to tmp failed on close");
--- a/src/cmds/qsub_sup.c
+++ b/src/cmds/qsub_sup.c
@@ -235,6 +235,9 @@ get_script(FILE *file, char *script, char *prefix)
 	char *in;
 	char *s_in = NULL;
 	int s_len = 0;
+	char *extend;
+	char *extend_in = NULL;
+	int extend_len;
 	static char tmp_template[] = "pbsscrptXXXXXX";
 	int fds;
 
@@ -258,8 +261,21 @@ get_script(FILE *file, char *script, char *prefix)
 		return (4);
 	}
 
-	while ((in = pbs_fgets_extend(&s_in, &s_len, file)) != NULL) {
+	while ((in = pbs_fgets(&s_in, &s_len, file)) != NULL) {
 		if (!exec && ((sopt = pbs_ispbsdir(s_in, prefix)) != NULL)) {
+			/* Check if this is a directive line that should be extended */
+			extend_len = pbs_extendable_line(in);
+			if (extend_len >= 0) {
+				in[extend_len] = '\0'; /* remove the backslash (\) */
+				extend = pbs_fgets_extend(&extend_in, &extend_len, file);
+				if (extend != NULL) {
+					if (pbs_strcat(&s_in, &s_len, extend) == NULL)
+						return (5);
+				}
+				in = s_in;
+				sopt = pbs_ispbsdir(s_in, prefix);
+			}
+
 			/*
 			 * Setting options from the job script will not overwrite
 			 * options set on the command line. CMDLINE-1 means
--- a/src/cmds/qsub_sup.c
+++ b/src/cmds/qsub_sup.c
@@ -236,8 +236,8 @@ get_script(FILE *file, char *script, char *prefix)
 	char *s_in = NULL;
 	int s_len = 0;
 	char *extend;
-	char *extend_in;
-	int extend_len;
+	char *extend_in = NULL;
+	int extend_len = 0;
 	int extend_loc;
 	static char tmp_template[] = "pbsscrptXXXXXX";
 	int fds;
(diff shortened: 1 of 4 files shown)


src/cmds/qsub_sup.c
```diff
@@ -258,8 +261,21 @@ get_script(FILE *file, char *script, char *prefix)
...
 	}
 
-	while ((in = pbs_fgets_extend(&s_in, &s_len, file)) != NULL) {
+	while ((in = pbs_fgets(&s_in, &s_len, file)) != NULL) {
 		if (!exec && ((sopt = pbs_ispbsdir(s_in, prefix)) != NULL)) {
+			/* Check if this is a directive line that should be extended */
+			extend_len = pbs_extendable_line(in);
+			if (extend_len >= 0) {
+				in[extend_len] = '\0'; /* remove the backslash (\) */
+				extend = pbs_fgets_extend(&extend_in, &extend_len, file);
```
bhroam, line 270: extend_len is the amount of space that is currently in extend_in.  At the start of the loop, extend_in == NULL, this will cause a crash.  If you don't preallocate any memory for extend_in, extend_len == 0.

Preallocating memory for extend_in is unnecessary.  pbs_fgets_extend() will do it for you.

src/cmds/qsub_sup.c
```diff
@@ -258,8 +261,21 @@ get_script(FILE *file, char *script, char *prefix)
...
+				if (extend != NULL) {
+					if (pbs_strcat(&s_in, &s_len, extend) == NULL)
+						return (5);
+				}
+				in = s_in;
+				sopt = pbs_ispbsdir(s_in, prefix);
```
bhroam, line 276: I don't see that sopt here being used.  The only time it is used here is when it is called with pbs_ispbsdir() up at line 265.  Also, since s_in would have started with the #PBS directive back there, it wouldn't have changed down here.  The only change to s_in is if we strcat() something onto it.

test/tests/functional/pbs_qsub_script.py
```diff
@@ -81,3 +81,31 @@ def test_qsub_basic_job(self):
...
+    def test_qsub_line_extensions(self):
+        """
+        This test case ensures that only #PBS directive lines are treated
+        as candidates for line extension
+        """
+        script = """#!/bin/sh
+        #PBS -m n
+        # This is a test comment that shouldn't be extended \\
+        cat $0
+        """
```
bhroam, line 94: Please extend a #PBS directive as well as a comment.

test/tests/functional/pbs_qsub_script.py
```diff
@@ -81,3 +81,31 @@ def test_qsub_basic_job(self):
...
+
+        fn = self.du.create_temp_file(body=script, asuser=TEST_USER)
+        cmd = [self.qsub_cmd, fn]
+        rv = self.du.run_cmd(self.server.hostname,
+                             cmd=cmd,
+                             runas=TEST_USER,
+                             cwd=self.sub_dir)
```
bhroam, line 101: No need to use create_temp_file()/run_cmd().  Use j = Job(), and j.create_script(body=script)

bhroam (changes requested):

src/cmds/qsub_sup.c
awellington: New push corrects to use `extend_loc` to save location from `pbs_extendable_line`
Also some corrections to memory management, ensuring that `pbs_fgets_extend` return is freed, etc.

src/cmds/qsub_sup.c
awellington: `sopt` is used in `do_dir` immediately below this block. Moved these two lines up inside the `if` however.

test/tests/functional/pbs_qsub_script.py
awellington: Updated script includes extended directive, with a blank extended line as well to verify this works correctly

test/tests/functional/pbs_qsub_script.py
awellington: Done

src/cmds/qsub_sup.c
```diff
@@ -264,16 +265,19 @@ get_script(FILE *file, char *script, char *prefix)
...
 		if (!exec && ((sopt = pbs_ispbsdir(s_in, prefix)) != NULL)) {
 			/* Check if this is a directive line that should be extended */
-			extend_len = pbs_extendable_line(in);
-			if (extend_len >= 0) {
-				in[extend_len] = '\0'; /* remove the backslash (\) */
+			extend_loc = pbs_extendable_line(in);
+			if (extend_loc >= 0) {
+				in[extend_loc] = '\0'; /* remove the backslash (\) */
+				extend_in = NULL;
+				extend_len = 0;
```
bhroam, line 272: You want to set extend_in to NULL and extend_len to 0 at the top of the function once.  After the first call to pbs_fgets_extend() there will be memory allocated to extend_len, so it can be reused.

src/cmds/qsub_sup.c
```diff
@@ -264,16 +265,19 @@ get_script(FILE *file, char *script, char *prefix)
...
 				extend = pbs_fgets_extend(&extend_in, &extend_len, file);
 				if (extend != NULL) {
 					if (pbs_strcat(&s_in, &s_len, extend) == NULL)
 						return (5);
+					in = s_in;
+					sopt = pbs_ispbsdir(s_in, prefix);
```
bhroam, line 278: I still don't understand what the purpose of this is.  sopt is set and checked above.  You don't check sopt again before you reset it on line 266.

bhroam (changes requested):

src/cmds/qsub_sup.c
bhroam: Oh, sorry about questioning this again.  I hadn't seen your response yet.  Thank you for your answer.  I see why you need to call it twice.

bhroam (commented):
There is no need to check if extend_in != NULL. free(NULL) is a noop.

awellington:
> There is no need to check if extend_in != NULL. free(NULL) is a noop.

No worries, memories of writing for PalmOS where `free(NULL)` crashed are still there :)

bhroam:
I never wrote for PalmOS.  I had a Palm Pilot though.  They were way cool :)

bhroam (approved):
Thanks for making all the changes!  LGTM

raviv1995 (approved):

prakashcv13:
merged based on review from Bhroam and Ravi

PR #2357, opened 2021-04-14
Remove mbox size check

subhasisb:
#### Describe Bug or Feature

Remove the checks for mbox sizes, since it does not make much difference and is adding extra complexity (the previous versions of PBS did not have this check as well, so we won't miss any functionality). The functionality that was added needs proper overhaul to add some kind of throttling on a sending flooding receiver with messages, but the current code is not useful, and hence the removal.

#### Describe Your Change

Keep the code that computes the mbox sizes, just do not reject messages on a particular mbox size (just remove the checks)

Profile: regression
|--|--|--|--|--|--|--|
|7070|2023|3|7|0|0|2013|

Failures are present in master as well

### DIFF
--- a/src/lib/Libtpp/tpp_client.c
+++ b/src/lib/Libtpp/tpp_client.c
@@ -2324,7 +2324,6 @@ leaf_next_event_expiry(time_t now)
  * @return Error code
  * @retval 0 - Success
  * @retval -1 - Failure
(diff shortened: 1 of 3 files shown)


hirenvadalia (approved):
@subhasisb Please post regression results

agrawalravi90 (approved):
LGTM

PR #2360, opened 2021-04-16
pbs_snapshot --obfuscate performance and bugfixes

agrawalravi90:
#### Describe Bug or Feature

Fixes the following bugs:
- —obfuscate for filenames with sensitive data would fail to delete the file with sensitive name
- --obfuscate captures sched_logs without scrubbing all sensitive data

And --obfuscate can take a very long time to complete for sites with a large number of accounting logs and topology/ files

#### Changes
- Fixed the filename issue
- Deleted sched logs with --obfuscate
- Added multi-processing + some optimizations to make obfuscating accounting logs faster, and deleted topology/ instead of scrubbing it.

Obfuscation of a snapshot with 3.3G of accounting logs used to take a very long time, I ran it for **5.5 hours** before I killed it. Now it takes around **12.5 mins**.

### DIFF
--- a/test/fw/ptl/utils/pbs_snaputils.py
+++ b/test/fw/ptl/utils/pbs_snaputils.py
@@ -52,6 +52,7 @@ import time
 import platform
 from subprocess import STDOUT
 from pathlib import Path
+from multiprocessing import Process
 
 from ptl.lib.pbs_ifl_mock import *
 from ptl.lib.pbs_testlib import (SCHED, BatchUtils, Scheduler, Server,
@@ -348,14 +349,13 @@ class ObfuscateSnapshot(object):
         """
         Helper function to anonymize
 
-        :param attrs_obf - list of attributes to obfuscate
-        :type attrs_obf - list
+        :param attrs_obf - set of attributes to obfuscate
+        :type attrs_obf - set
         :param file_path - path of acct log file
         :type file_path - str
         """
-        fout = self.du.create_temp_file()
-
-        with open(file_path, "r") as fd, open(fout, "w") as fdout:
+        newcontent = []
+        with open(file_path, "r") as fd:
             for record in fd:
                 # accounting log format is
                 # %Y/%m/%d %H:%M:%S;<Key>;<Id>;<key1=val1> <key2=val2> ...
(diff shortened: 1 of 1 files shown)


ashwathraop (approved):

agrawalravi90:
@openpbs/openpbs-maintainers  I need a maintainer review for this

@ashwathraop thanks for the sign-off!

test/fw/ptl/utils/pbs_snaputils.py
```diff
@@ -402,9 +402,10 @@ def _obfuscate_acct_file(self, attrs_obf, file_path):
                 if not skip_record:
                     record = ";".join(record_list[:3]) + ";" + \
                         " ".join(["=".join(n) for n in kvl_list])
-                    fdout.write(record + "\n")
+                    newcontent.append(record + "\n")
 
-        shutil.move(fout, file_path)
+        with open(file_path, "w") as fd:
+            fd.write("".join(newcontent))
```
bhroam, line 408: Wouldn't it be better to use fd.writelines(newcontent)?

test/fw/ptl/utils/pbs_snaputils.py
```diff
@@ -420,13 +421,34 @@ def obfuscate_acct_logs(self, snap_dir, sudo_val):
...
-            self._obfuscate_acct_file(attrs_to_obf, acct_fpath)
+
+        # Limit the number of cores used to 10
+        ncpus = os.cpu_count()
+        ncpus = min(ncpus, 10)
+        nfiles = len(acct_fpaths)
+        i = 0
+        while True:
+            if i >= nfiles:
+                break
```
bhroam, line 438: You could move i >= nfiles as the while conditional.

bhroam (changes requested):

test/fw/ptl/utils/pbs_snaputils.py
agrawalravi90: I think writelines() is slower than doing a join and a write(), I saw posts like this and decided to just use write: https://stackoverflow.com/questions/44560655/python-writelines-and-write-huge-time-difference

bhroam (approved):
LGTM

PR #2364, opened 2021-04-20
Reservation test fails on cpuset machines

arungrover:
#### Describe Bug or Feature
A reservation test fails on a cpuset machine

#### Describe Your Change
On a cpuset machine the test submits a job to run on a particular vnode but it does not submit the reservation to also run on the same vnode. Because of this, the reservation gets confirmed elsewhere and the test fails.

#### Link to Design Doc
NA

[test_resv.txt](https://github.com/openpbs/openpbs/files/6339730/test_resv.txt)

### DIFF
--- a/test/tests/functional/pbs_reservations.py
+++ b/test/tests/functional/pbs_reservations.py
@@ -789,9 +789,14 @@ class TestReservations(TestFunctional):
         jid = self.server.submit(j)
         self.server.expect(JOB, {'job_state': 'R'}, id=jid)
 
+        if self.mom.is_cpuset_mom():
+            select_val = vnode_val
+        else:
+            select_val = '1:ncpus=1'
+
         # Submit a reservation that will start after the job starts running
         rid1 = self.submit_reservation(user=TEST_USER,
-                                       select='1:ncpus=1',
+                                       select=select_val,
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_reservations.py
```diff
@@ -789,9 +789,14 @@ def test_future_resv_conflicts_running_job(self):
         jid = self.server.submit(j)
         self.server.expect(JOB, {'job_state': 'R'}, id=jid)
 
+        if self.mom.is_cpuset_mom():
+            select_val = vnode_val
+        else:
+            select_val = '1:ncpus=1'
+
```
bhroam, line 796: Why don't you combine this if statement with the if self.mom.is_coupsaet_mom() if statement above

bhroam (changes requested):

test/tests/functional/pbs_reservations.py
arungrover: I've made the change

bhroam (approved):
LGTM

bayucan (approved):
Looks good.

bayucan:
Merged #2364 into master.

PR #2366, opened 2021-04-20
fix test_terminal_control_shell_function on RHEL7

RKORavi:
#### Describe Bug or Feature
Test case: test_nonprint_shell_function, test_terminal_control_shell_function fails on rhel7 due to bash version.
Bash version required to run these test cases must be greater than or equal to 4.2.46.

#### Describe Your Change
1. Add a decorator to check for bash version
2. Skip tests if bash version is less than 4.2.46

#### Link to Design Doc
NA

[7194-regression_report.txt](https://github.com/openpbs/openpbs/files/6342947/7194-regression_report.txt)

[7194-ALL.zip](https://github.com/openpbs/openpbs/files/6342977/7194-ALL.zip)

### DIFF
--- a/test/fw/ptl/utils/pbs_testsuite.py
+++ b/test/fw/ptl/utils/pbs_testsuite.py
@@ -611,10 +611,10 @@ class PBSTestSuite(unittest.TestCase):
         skip_test = False
         msg = 'capability supported only for bash version >= 4.2.46'
         for mom in cls.moms.values():
-            if mom.check_mom_bash_version():
+            if not mom.check_mom_bash_version():
                 skip_test = True
                 break
-        if skip_test:
+        if not skip_test:
             return
         if cls.__dict__.get('__check_mom_bash_version__', False):
             # skip all test cases in this test suite
--- a/test/fw/ptl/utils/pbs_testsuite.py
+++ b/test/fw/ptl/utils/pbs_testsuite.py
@@ -168,6 +168,12 @@ def runOnlyOnLinux(function):
     function.__run_only_on_linux__ = True
     return function
 
+def checkMomBashVersion(function):
+    """
+    Decorator to skip a test if bash version is less than 4.2.46
+    """
+    function.__check_mom_bash_version__ = True
+    return function
 
 def requirements(*args, **kwargs):
     """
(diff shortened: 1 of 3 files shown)


test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -597,6 +606,23 @@ def run_only_on_linux(cls):
                     test_item.__unittest_skip__ = True
                     test_item.__unittest_skip_why__ = msg
 
+    @classmethod
+    def check_mom_bash_version(cls):
+        if cls.mom.check_mom_bash_version():
```
vishwaks, line 611: this check we have to do for all moms, if multiple moms are configured.

vishwaks (changes requested):

RKORavi:
Id: 7306
Description: Tests from given sources on platforms UBUNTU1804, SUSE15, UBUNTU1604, CENTOS8, RHEL8, CENTOS7, RHEL7, SLES12, SLES15
Profile: regression

Test Summary: total: 243, fail: 0, error: 9, timedout: 0, skip: 2, pass: 232

test/fw/ptl/utils/pbs_testsuite.py
```diff
@@ -597,6 +606,28 @@ def run_only_on_linux(cls):
...
 
+    @classmethod
+    def check_mom_bash_version(cls):
+        skip_test = False
+        msg = 'capability supported only for bash version >= 4.2.46'
+        for mom in cls.moms.values():
+            if mom.check_mom_bash_version():
+                skip_test = True
+                break
+        if skip_test:
```
vishwaks, line 617: This variable name might confuse. As I understand skip_test = True, then test will not skip. Right ? Then can we rename skip_test to no_skip(or something similar) or just swap the values default skip_test = True in loop update it to False and then use it in if condition?

vishwaks (changes requested):

test/fw/ptl/utils/pbs_testsuite.py
RKORavi: done

vishwaks (approved):
thanks for making changes. I sign off.

SudeshnaMoulik (approved):
LGTM

prakashcv13 (approved):

PR #2368, opened 2021-04-21
server can go in infinite loop with deljoblist

agrawalravi90:
#### Describe Bug or Feature

Server could go into an infinite loop if qdel was executed while some jobs were in the "prerun" substate, and mom takes some time to send updates to the server. It's caused by the recent deljoblist enhancement as follows:
- when server encounters a job in the prerun state, it adds a work task (post_delete_route) to get back to it after a second.
- The problem is: it uses the original batch_request, so it adds a post_delete_route for ALL jobs.
- It also does this for every job that's in the prerun substate
- So, if mom takes longer than a second, the work tasks start getting executed. Since they also try to delete ALL jobs, they end up creating work tasks of their own, again each of them trying to delete ALL jobs.
- Server gets stuck in this cycle and goes into an infinite loop where it stops responding to any requests, including the updates from mom which would have saved it

#### Describe Your Change

req_deletejob2 now sets a flag which tells req_deletejob that a post_delete_route work task has been added so it needs to suspend the deletion of jobs. So, it resumes deleting the jobs from the first job that it finds in the prerun state, and keeps after every second until the mom sends updates. This is similar to how it used to behave before deljoblist enhancement.

After the change, the server no longer goes into an infinite loop. There's no good test for this since it's a race condition. Please inspect the code. I'll do regressions to make sure nothing breaks

### DIFF
--- a/src/server/array_func.c
+++ b/src/server/array_func.c
@@ -816,7 +816,7 @@ dup_br_for_subjob(struct batch_request *opreq, job *pjob, int (*func)(struct bat
 
 	npreq = alloc_br(opreq->rq_type);
 	if (npreq == NULL)
-		return 0;
+		return 1;
 
 	npreq->rq_perm    = opreq->rq_perm;
 	npreq->rq_fromsvr = opreq->rq_fromsvr;
@@ -859,7 +859,7 @@ dup_br_for_subjob(struct batch_request *opreq, job *pjob, int (*func)(struct bat
 		default:
 			delete_link(&npreq->rq_link);
 			free(npreq);
-			return 0;
+			return 1;
 	}
 
 	npreq->rq_parentbr = opreq;
--- a/src/server/array_func.c
+++ b/src/server/array_func.c
@@ -806,15 +806,17 @@ create_subjob(job *parent, char *newjid, int *rc)
  *		PBS_Batch_Rerun, and PBS_Batch_RunJob subjob requests.
  *		For any other request types, be sure to add another switch case below
  *		(matching request type).
+ * @return int
+ * @retval return value of the callback function
  */
-void
-dup_br_for_subjob(struct batch_request *opreq, job *pjob, void (*func)(struct batch_request *, job *))
+int
+dup_br_for_subjob(struct batch_request *opreq, job *pjob, int (*func)(struct batch_request *, job *))
 {
 	struct batch_request  *npreq;
 
 	npreq = alloc_br(opreq->rq_type);
 	if (npreq == NULL)
-		return;
+		return 0;
 
 	npreq->rq_perm    = opreq->rq_perm;
 	npreq->rq_fromsvr = opreq->rq_fromsvr;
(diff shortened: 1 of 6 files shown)


src/server/req_delete.c
```diff
@@ -785,6 +805,10 @@ req_deletejob2(struct batch_request *preq, job *pjob)
 
 					kill((pid_t) pwtold->wt_event, SIGTERM);
 					set_job_substate(pjob, JOB_SUBSTATE_ABORT);
+					if (preq->rq_type == PBS_BATCH_DeleteJobList) {
+						/* let the caller know that the deljoblist request needs to be suspended */
+						suspenddelete = true;
```
subhasisb, line 810: Why can't we use a simple return value here, instead of a global variable?

src/server/req_delete.c
```diff
@@ -824,6 +847,9 @@ req_deletejob2(struct batch_request *preq, job *pjob)
 			update_deletejob_stat(pjob->ji_qs.ji_jobid, preq, PBSE_SYSTEM);
 			if (preply->brp_un.brp_deletejoblist.tot_rpys == preply->brp_un.brp_deletejoblist.tot_jobs)
 				req_reject(PBSE_SYSTEM, 0, preq);
+		} else if (preq->rq_type == PBS_BATCH_DeleteJobList) {
+			/* let the caller know that the deljoblist request needs to be suspended */
+			suspenddelete = true;
```
subhasisb, line 852: same here - seems like a return value (boolean) would have sufficed?

subhasisb (changes requested):

src/server/req_delete.c
agrawalravi90: i cannot change the function prototype of req_deletejob2 since it gets passed as a callback to dup_br_for_subjob

subhasisb: but we can change the definition of dup_br_for_subjob() itself - it is barely called from 7-8 places and not all of them need a return value anyway. I just feel that a global like this is quite dangerous

agrawalravi90: dup_br_for_subjob takes multiple callback functions as input, not just req_deletejob2. If I change its function pointer argument's prototype, then I'll also have to change the prototype of req_signaljob2, req_runjob2 and req_rerunjob2. But I do see your point and I don't like this either, so let me change it.

src/server/array_func.c
```diff
@@ -806,15 +806,17 @@ create_subjob(job *parent, char *newjid, int *rc)
  *		PBS_Batch_Rerun, and PBS_Batch_RunJob subjob requests.
  *		For any other request types, be sure to add another switch case below
  *		(matching request type).
+ * @return int
+ * @retval return value of the callback function
```
subhasisb, line 810: please add return value descriptions in doxygen.
also here you return 0 for failure, but in req_deljob2 0 is success - can we make it the same?

subhasisb (changes requested):

subhasisb (approved):

ashwathraop (approved):

ashwathraop:
Please run Valgrind regression once since we have seen many leaks around these changes before.

bremanandjk:
Please try to run these changes under valgrind, especially t2.t of 8345-qsubqdel_regression_tests and 4046_support_site_allocation_mgmt.

agrawalravi90:
@bremanandjk  thanks for the suggestion, i did run valgrind for those tests, also included any test with 'qdel' in its name, the report seems to be ok.

bremanandjk (approved):
LGTM

ashwathraop:
Thanks for running the tests.

PR #2371, opened 2021-04-23
Fix: qsub does not accept paths with colons for output files

raviv1995:
#### Describe Bug or Feature

Linux allows the file path to contain colons(:). However, while specifying a file path for ATTR_o or ATTR_e (output and error paths) qsub rejects a path with a colon in it.

#### Describe Your Change

If a path with a colon is provided to qsub, it considers it to include a hostname compulsorily. In `prepare_path` when we parse the string to extract a hostname we break out of the loop if the character is not a valid character for hosts (alphanumeric, periods, and hyphens). So we check for a '/' there (signifying a path with a possible colon) and continue the function with no host.

[logs_qsub_colon.txt](https://github.com/openpbs/openpbs/files/6392327/logs_qsub_colon.txt)

[ptl.log](https://github.com/openpbs/openpbs/files/6396649/ptl.log)
th3_open

### DIFF
--- a/src/lib/Libcmds/prepare_path.c
+++ b/src/lib/Libcmds/prepare_path.c
@@ -165,9 +165,23 @@ prepare_path(char *path_in, char *path_out)
 				}
 			}
 			if (*c != ':')
-				return 1;
-			/* Advance past the colon */
-			c++;
+			{
+				if (*c == '/') {
+					/* There's a colon in the path */ 
+					host_given = NULL;
+					host_name[0] = '\0';
+					for (c = path_in; *c; c++) {
+						if (isspace(*c) == 0)
+							break;
+					}
+				}
+				else
+					return 1;
+			}
+			else {
+				/* Advance past the colon */
+				c++;
+			}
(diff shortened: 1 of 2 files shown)


npadole20:
@raviv1995  please attach test logs

src/lib/Libcmds/prepare_path.c
```diff
@@ -165,9 +165,23 @@ prepare_path(char *path_in, char *path_out)
...
+				if (*c == '/') {
+					/* There's a colon in the path */ 
+					host_given = NULL;
+					host_name[0] = '\0';
+					for (c = path_in; *c; c++) {
+						if (isspace(*c) == 0)
+							break;
+					}
+				}
+				else
```
npadole20, line 178: Move this to the line above, same for else at line no. 181

npadole20 (changes requested):

npadole20 (approved):

prakashcv13 (approved):
LGTM

PR #2374, opened 2021-04-24
fix tpp large data packet handling

subhasisb:
#### Describe Bug or Feature

the tcp recv() call's return value was being collected in a short. This is an obvious oversight. A packet size larger than can hold in a short resulted in a negative value of the return, and thus resulted in communication breakdown.

#### Describe Your Change

Fixed the return value (and for send as well) to be ssize_t.
Along with that fixed a few ints to be unsigned ints, some logging fixes.

Profile: regression
|--|--|--|--|--|--|--|
|7296|2036|1|0|0|8|2027|

The single failure is unrelated to TPP anyway

### DIFF
--- a/src/lib/Libtpp/tpp_internal.h
+++ b/src/lib/Libtpp/tpp_internal.h
@@ -139,7 +139,7 @@ typedef struct {
 typedef struct {
 	pbs_list_link chunk_link;
 	char *data;	/* pointer to the data buffer */
-	int len;	/* length of the data buffer */
+	size_t len;	/* length of the data buffer */
 	char *pos;	/* current position - till which data is consumed */
 } tpp_chunk_t;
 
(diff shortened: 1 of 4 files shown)


src/lib/Libtpp/tpp_internal.h
```diff
@@ -150,20 +150,20 @@ typedef struct {
...
 	pbs_list_head chunks;
 	tpp_chunk_t *curr_chunk;
-	int totlen;
+	size_t totlen;
 	int ref_count;	/* number of accessors */
 } tpp_packet_t;
 
 typedef struct {
-	int ntotlen;
+	unsigned int ntotlen;
```
nithinj, line 158: why not use ssize_t here and elsewhere for ntotlen?

agrawalravi90: ssize_t is a signed type, why would you want to use that?

agrawalravi90 (approved):
LGTM

src/lib/Libtpp/tpp_internal.h
subhasisb: That would be changing the data size to long and will render the packets no longer backward compatible. Not required right now.

nithinj (approved):

PR #2376, opened 2021-04-26
Sched multi-threading doesn't block relevant signals

agrawalravi90:
#### Describe Bug or Feature

Scheduler's MT code blocks SIGHUP, but no other signals. Since the main thread blocks the relevant signals after the threads are already created, the threads don't inherit the main thread's sigmask. This means that they can be interrupted while doing critical operations and lead to unexpected behavior.

#### Describe Your Change

Added SIGHUP to the sigmask that sched blocks instead of only adding SIGHUP. Now worker threads will block all relevant signals

**Before**
```
Thread 2 "pbs_sched" hit Breakpoint 1, check_node_eligibility_chunk (data=0x1655530) at ../../../src/scheduler/node_info.cpp:5567
5567		if (data == NULL)
(gdb) n

Thread 2 "pbs_sched" received signal SIGTERM, Terminated.
check_node_eligibility_chunk (data=0x1655530) at ../../../src/scheduler/node_info.cpp:5567
5567		if (data == NULL)
(gdb) n
die (sig=15) at ../../../src/scheduler/pbs_sched_utils.cpp:235
235		ret_lock = pthread_mutex_trylock(&cleanup_lock);
(gdb) n
236		if (ret_lock != 0)
(gdb) 
```

**After**
```
Thread 2 "pbs_sched" hit Breakpoint 1, check_node_eligibility_chunk (data=0xde3920) at ../../../src/scheduler/node_info.cpp:5567
5567		if (data == NULL)
(gdb) n

Thread 2 "pbs_sched" hit Breakpoint 1, check_node_eligibility_chunk (data=0xde3920) at ../../../src/scheduler/node_info.cpp:5570
5570		err = new_schd_error();
(gdb) n
5571		if (err == NULL) {
(gdb) n
5575		misc_err = new_schd_error();
(gdb) n
5576		if (misc_err == NULL) {
(gdb) finish
Run till exit from #0  check_node_eligibility_chunk (data=0xde3920) at ../../../src/scheduler/node_info.cpp:5576
[Switching to Thread 0x14c82f4e9700 (LWP 9933)]

Thread 3 "pbs_sched" hit Breakpoint 1, check_node_eligibility_chunk (data=0x90c8c0) at ../../../src/scheduler/node_info.cpp:5567
5567		if (data == NULL)
(gdb) c
Continuing.
[Switching to Thread 0x14c82f6ea700 (LWP 9932)]

Thread 2 "pbs_sched" hit Breakpoint 1, check_node_eligibility_chunk (data=0x90c8f0) at ../../../src/scheduler/node_info.cpp:5567
5567		if (data == NULL)
(gdb) c
Continuing.
[Switching to Thread 0x14c82f4e9700 (LWP 9933)]

Thread 3 "pbs_sched" hit Breakpoint 1, check_node_eligibility_chunk (data=0x90c8c0) at ../../../src/scheduler/node_info.cpp:5570
5570		err = new_schd_error();
(gdb) c
Continuing.
[Switching to Thread 0x14c82f6ea700 (LWP 9932)]

Thread 2 "pbs_sched" hit Breakpoint 1, check_node_eligibility_chunk (data=0x90c8f0) at ../../../src/scheduler/node_info.cpp:5570
5570		err = new_schd_error();
(gdb) c
Continuing.

Thread 1 "pbs_sched" received signal SIGTERM, Terminated.
[Switching to Thread 0x14c83454a740 (LWP 9924)]
0x000014c83266dac1 in sigprocmask () from /lib64/libc.so.6
```

### DIFF
--- a/src/scheduler/multi_threading.cpp
+++ b/src/scheduler/multi_threading.cpp
@@ -60,6 +60,8 @@
 #include "resource_resv.h"
 #include "multi_threading.h"
 
+extern sigset_t allsigs;
+
 /**
  * @brief	create the thread id key & set it for the main thread
  *
@@ -237,8 +239,8 @@ worker(void *tid)
 	pthread_setspecific(th_id_key, tid);
 	ntid = *(int *)tid;
 
-	/* Block HUPs, if we ever unblock this, we'll need to modify 'restart()' to handle MT */
-	sigemptyset(&set);
+	/* Add HUP to the list of signals to block, if we ever unblock this, we'll need to modify 'restart()' to handle MT */
+	set = allsigs;
 	sigaddset(&set, SIGHUP);
 
 	if (pthread_sigmask(SIG_BLOCK, &set, NULL) != 0) {


src/scheduler/multi_threading.cpp
```diff
@@ -237,8 +239,8 @@ worker(void *tid)
 	pthread_setspecific(th_id_key, tid);
 	ntid = *(int *)tid;
 
-	/* Block HUPs, if we ever unblock this, we'll need to modify 'restart()' to handle MT */
-	sigemptyset(&set);
+	/* Add HUP to the list of signals to block, if we ever unblock this, we'll need to modify 'restart()' to handle MT */
+	set = allsigs;
```
arungrover, line 243: Shouldn't the threads block allsigs all the time? This looks like that it will block momentarily and unblock later on. So if the scheduler receives a signal after it is out of threading code but still within a cycle, one of these threads can potentially honor the signal even though the scheduler blocked it for the entire cycle.

agrawalravi90: they do block them all the time right? each thread will block these signals until they die, we never unblock them right? The main thread will unblock allsigs and block them again every cycle, but that will not affect the fact that the worker threads have these blocked, right?

bhroam (approved):
LGTM

src/scheduler/multi_threading.cpp
arungrover: Oh ok, I thought worker threads unblock them later on. This looks good then.

arungrover (approved):

PR #2378, opened 2021-04-28
Refactoring: removing the global for signal mask in sched

agrawalravi90:
#### Describe Bug or Feature

Recently I fixed a signal blocking issue with scheduler multi-threading (https://github.com/openpbs/openpbs/pull/2376), but I added a global dependency on libsched in the process, which can cause issues to anything that links with libsched other than pbs_sched, and is a bad coding practice as well. So, this PR refactors the original fix.

### DIFF
--- a/src/scheduler/multi_threading.cpp
+++ b/src/scheduler/multi_threading.cpp
@@ -238,7 +238,7 @@ worker(void *tid)
 	ntid = *(int *)tid;
 
 	/* Add HUP to the list of signals to block, if we ever unblock this, we'll need to modify 'restart()' to handle MT */
-	pthread_sigmask(SIG_BLOCK, NULL, &set);
+	sigemptyset(&set);
 	sigaddset(&set, SIGHUP);
 
 	if (pthread_sigmask(SIG_BLOCK, &set, NULL) != 0) {
--- a/src/scheduler/multi_threading.cpp
+++ b/src/scheduler/multi_threading.cpp
@@ -60,8 +60,6 @@
 #include "resource_resv.h"
 #include "multi_threading.h"
 
-extern sigset_t allsigs;
-
 /**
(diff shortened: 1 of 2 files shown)


arungrover (approved):
LGTM

src/scheduler/multi_threading.cpp
```diff
@@ -240,7 +238,7 @@ worker(void *tid)
 	ntid = *(int *)tid;
 
 	/* Add HUP to the list of signals to block, if we ever unblock this, we'll need to modify 'restart()' to handle MT */
-	set = allsigs;
+	pthread_sigmask(SIG_BLOCK, NULL, &set);
```
arungrover, line 241: Just a minor comment, shouldn't this be SIG_SETMASK? something like pthread_sigmask(SIG_SETMASK, &set, NULL)?

agrawalravi90: As we discussed Arun, when 'set' is NULL, the first argument is ignored, so it doesn't really matter.

arungrover: Nevermind, I was confused by how man page has explained SIG_SETMASK.  https://www.gnu.org/software/libc/manual/html_node/Process-Signal-Mask.html explains better. Please ignore the comment.

subhasisb: I think u could do this in one pthread_sigmask() call right? SIG_BLOCK is supposed to "add" to the current mask-set
By doing this:
sigemptyset(&set);
sigaddset(&set, SIGHUP);
pthread_sigmask(SIG_BLOCK, &set, NULL);

subhasisb (changes requested):

subhasisb (approved):

arungrover (approved):

PR #2382, opened 2021-05-05
handling queued limits with multi-server

nithinj:
#### Describe Bug or Feature
Multi-server doesn't enforce queued limits globally. qsub will return an error if the random server it reaches out to has exhausted its limits.

#### Describe Your Change
If the qsub reaches a random server where the limit is reached, it should attempt to queue jobs in the alternate servers before returning an error.

[test.log](https://github.com/openpbs/openpbs/files/6430106/test.log)

Profile: regression
|--|--|--|--|--|--|--|
|7648|4176|0|10|0|0|4166|

Description: Rerun Only Failed Tests From #7648
Platforms: SUSE15,SLES15
Profile: regression
|--|--|--|--|--|--|--|
|7659|10|0|0|0|0|10|

### DIFF
--- a/src/include/libutil.h
+++ b/src/include/libutil.h
@@ -50,7 +50,7 @@ extern "C" {
 
 /* misc_utils specific */
 
-#define IS_EMPTY(str) (str && str[0] != '\0')
+#define IS_EMPTY(str) (!str || str[0] == '\0')
 
 /* replace - Replace sub-string  with new pattern in string */
 void replace(char *, char *, char *, char *);
--- a/src/include/libutil.h
+++ b/src/include/libutil.h
@@ -50,6 +50,8 @@ extern "C" {
 
 /* misc_utils specific */
 
+#define IS_EMPTY(str) (str && str[0] != '\0')
+
 /* replace - Replace sub-string  with new pattern in string */
 void replace(char *, char *, char *, char *);
 
--- a/src/lib/Libifl/pbsD_submit.c
+++ b/src/lib/Libifl/pbsD_submit.c
@@ -167,9 +167,11 @@ __pbs_submit(int c, struct attropl  *attrib, char *script, char *dest, char *ext
 	int msvr = multi_svr_op(c);
 	svr_conn_t **svr_conns = get_conn_svr_instances(c);
 	int nsvr = get_num_servers();
-	int start = rand_num() % nsvr;
+	int rand = rand_num() % nsvr;
 	int ct;
+	int start = rand;
 	int i;
+	
 
 	/* initialize the thread context data, if not already initialized */
 	if ((pbs_errno = pbs_client_thread_init_thread_context()) != 0)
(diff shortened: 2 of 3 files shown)


agrawalravi90 (changes requested):
Since we are still trying to figure out how to handle global data, I think we should postpone this change until we have more clarity on the future direction.

subhasisb (changes requested):
Blocking this for now to get to a design consensus before we commit this. Thanks!

nithinj:
@agrawalravi90 & @subhasisb ,
This PR is ready for review. Please have a look.

src/include/libutil.h
```diff
@@ -50,6 +50,8 @@ extern "C" {
 
 /* misc_utils specific */
 
+#define IS_EMPTY(str) (str && str[0] != '\0')
```
agrawalravi90, line 53: Shouldn't this be (!str || str[0] == '\0') ? right now, IS_EMPTY(NULL) and IS_EMPTY("\0") will return false

src/lib/Libifl/pbsD_submit.c
```diff
@@ -218,23 +221,37 @@ __pbs_submit(int c, struct attropl  *attrib, char *script, char *destination, ch
 			extend = EXTEND_OPT_IMPLICIT_COMMIT;
 	}	
 
-	c = random_svr_conn;	
-	if (msvr && destination) {
+	if (msvr && IS_EMPTY(dest)) {
```
agrawalravi90, line 224: Shouldn't this be !IS_EMPTY(dest) ?

src/lib/Libifl/pbsD_submit.c
```diff
@@ -218,23 +221,37 @@ __pbs_submit(int c, struct attropl  *attrib, char *script, char *destination, ch
...
+			rc = PBSE_NOSERVER;
+			continue;
+		}
+
+		c = svr_conns[i]->sd;
+		return_jobid = PBSD_queuejob(c, "", dest, attrib, extend, PROT_TCP, NULL, &commit_done);
+		if (return_jobid)
+			break;
+		else if (IS_EMPTY(dest) || (pbs_errno != PBSE_MAXQUED && pbs_errno != PBSE_ENTLIMCT && pbs_errno != PBSE_ENTLIMRESC))
+				goto error;
```
agrawalravi90, line 249: Not sure I understand this. If 'dest' was given, we'll not try other servers? Why? 'dest' is just a queue name, which can be shared for all servers.

Also, I think we should try all servers regardless of what the error was. Is there any case where it'd be bad if we tried all servers without checking for error?

src/lib/Libifl/pbsD_submit.c
```diff
@@ -218,23 +221,37 @@ __pbs_submit(int c, struct attropl  *attrib, char *script, char *destination, ch
...
 		/* Reached here means job is submitted to non default queue */
-		int start;
 
 		/* Since this could be a reservation queue and reservation queues are not shared,
 		 * we try to find out which server this queue resides on
 		 */
-		if ((start = get_obj_location_hint(destination, MGR_OBJ_RESV)) != -1)
-			c = svr_conns[start]->sd;
+		if ((start = get_obj_location_hint(dest, MGR_OBJ_RESV)) == -1)
+			start = 0;
```
agrawalravi90, line 231: Why not leave start as the random value that we got earlier?

agrawalravi90 (changes requested):
Please fix CI issues

nithinj:
@agrawalravi90 ,

agrawalravi90 (approved):
LGTM

nithinj:
Thanks @agrawalravi90 for the approval.

@subhasisb could you take a look now.

ashwathraop (approved):
LGTM

PR #2383, opened 2021-05-06
Check return output of mom bash version

RKORavi:
#### Describe Bug or Feature
Check return output of mom bash version. 
If it fails to get mom bash version then it results in : IndexError: list index out of range

#### Describe Your Change
Check if command execution to get mom version fails then return True

#### Link to Design Doc
NA

[TestNonprintingCharacters.zip](https://github.com/openpbs/openpbs/files/6435094/TestNonprintingCharacters.zip)

### DIFF
--- a/test/fw/ptl/lib/ptl_mom.py
+++ b/test/fw/ptl/lib/ptl_mom.py
@@ -699,10 +699,9 @@ class MoM(PBSService):
         if len(ret['out']) > 0:
             mom_bash_version = ret['out'][0]
         else:
-            # Test case will be skipped when this function returns False,
-            # if it fails to get bash version. Run test if it fails to
-            # get bash version. On some platform command to get bash version
-            # might fail.
+            # If we can't get the bash version, there is no harm
+            # in trying to run the test. It might fail in an error,
+            # but at least we tried.
             return True
         if mom_bash_version >= req_bash_version:
             return True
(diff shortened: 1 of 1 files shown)


prakashcv13 (approved):

npadole20 (approved):

test/fw/ptl/lib/ptl_mom.py
```diff
@@ -696,7 +696,11 @@ def check_mom_bash_version(self):
         ret = self.du.run_cmd(self.hostname, cmd=cmd, sudo=True,
                               as_script=True)
         req_bash_version = "4.2.46"
-        mom_bash_version = ret['out'][0]
+        if len(ret['out']) > 0:
+            mom_bash_version = ret['out'][0]
+        else:
+            # If it fails to get bash version return True
```
bhroam, line 702: Comments should explain why, not what the code is doing.  Why are you returning true if we can't get the bash version?

bhroam (commented):
I'm still not sure I understand.  If you need the bash version to be a certain version, and you can't get the version, shouldn't you skip the test?  Or are you saying that if you can't figure it out, try anyway?  If that's the case, the comment should say that.

RKORavi:
> I'm still not sure I understand. If you need the bash version to be a certain version, and you can't get the version, shouldn't you skip the test? Or are you saying that if you can't figure it out, try anyway? If that's the case, the comment should say that.

@bhroam please let me know if comment is clear now.

bhroam:
@RKORavi I'm not sure the comment is helpful.  Let me know if this is wrong: If we can't get the bash version, there is no harm in trying to run the test.  It might fail in an error, but at least we tried, right?

Otherwise I'm still not understanding why you want to run the test if you can't find the bash version.  You are querying the bash version to see if it will run properly.  Why are you still running the test if you don't know if it will run properly?

prakashcv13 (approved):

PR #2389, opened 2021-05-12
scheduler hangs on HUP while connecting to servers

arungrover:
#### Describe Bug or Feature
Scheduler hangs when during startup scheduler gets a SIGHUP while it is trying to connect to the server(s)

#### Describe Your Change
Issue happens because in order to connect to server scheduler forks out a pbs_iff process for authentication. Due to fork, an at-fork handler is called in the parent process which acquires a lock on the logging mutex. Now, what happens is that if scheduler receives HUP while it is in the middle of fork, it calls "restart" signal handler which logs. Since logging mutex isn't recursive, "restart" signal handler hangs forever to acquire the logging mutex.

Change is to block all signals while trying to connect/reconnect the server. This may not apply to peer servers because that happens in the middle of a cycle and signals are anyways blocked.

#### Link to Design Doc
NA

Running regression tests.
[post_fix.txt](https://github.com/openpbs/openpbs/files/6462732/post_fix.txt)
[pre_fix.txt](https://github.com/openpbs/openpbs/files/6462733/pre_fix.txt)

### DIFF
--- a/src/scheduler/pbs_sched_utils.cpp
+++ b/src/scheduler/pbs_sched_utils.cpp
@@ -593,34 +593,25 @@ connect_svrpool()
 		 */
 		if (clust_primary_sock < 0) {
 			clust_primary_sock = pbs_connect(NULL);
-			if (clust_primary_sock < 0) {
+			if (clust_primary_sock < 0)
 				/* wait for 2s for not to burn too much CPU, and then retry connection */
-				sleep(2);
-				close_servers();
-				goto CONTINUE;
-			}
+				goto unmask_continue;
 		}
 		clust_secondary_sock = pbs_connect(NULL);
-		if (clust_secondary_sock < 0) {
+		if (clust_secondary_sock < 0)
 			/* wait for 2s for not to burn too much CPU, and then retry connection */
-			sleep(2);
-			close_servers();
-			goto CONTINUE;
-		}
+			goto unmask_continue;
 
 		svr_conns_primary = get_conn_svr_instances(clust_primary_sock);
 		svr_conns_secondary = get_conn_svr_instances(clust_secondary_sock);
-		if (svr_conns_primary == NULL || svr_conns_secondary == NULL) {
+		if (svr_conns_primary == NULL || svr_conns_secondary == NULL)
 			/* wait for 2s for not to burn too much CPU, and then retry connection */
-			sleep(2);
-			close_servers();
-			goto CONTINUE;
-		}
+			goto unmask_continue;
 
 		for (i = 0; svr_conns_primary[i] != NULL && svr_conns_secondary[i] != NULL; i++) {
 			if (svr_conns_primary[i]->state == SVR_CONN_STATE_DOWN ||
 			    svr_conns_secondary[i]->state == SVR_CONN_STATE_DOWN)
-				goto BREAK;
+				goto unmask_break;
 		}
 
 		if (i != get_num_servers()) {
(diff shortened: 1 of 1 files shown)


src/scheduler/pbs_sched_utils.cpp
```diff
@@ -776,7 +776,11 @@ wait_for_cmds()
 				if (err != 1) {
 					/* if memory error ignore, else reconnect server */
 					if (err != -2) {
+						if (sigprocmask(SIG_BLOCK, &allsigs, &oldsigs) == -1)
```
agrawalravi90, line 779: Id suggest adding these to inside connect_svrpool and reconnect_servers, because these functions might get called from multiple places, so it’ll ensure that we don’t miss blocking the signals.

agrawalravi90 (changes requested):

src/scheduler/pbs_sched_utils.cpp
arungrover: So the problem is that in some cases reconnect_server is already called from a place where we block signals before calling it. See, schedule_wrapper()->send_cycle_end()->reconnect_server(). 
If I move it inside connect_svrpool or reconnect_server, I'll make scheduler block signals again for no particular reason. This is the reason why I am adding the change only where it is needed.

agrawalravi90: ok, I see your point, but is there any harm in potentially blocking them again? If there's no harm then I still think that it might be safer to block them inside these functions. The reason why I brought this up is because I'm currently working on a change around these functions and I'd have probably not realized that we need to block signals before calling them from somewhere else in the code and injected this bug again.

arungrover: I've addressed your comment Ravi, please have a look.

agrawalravi90 (approved):
LGTM

bhroam (approved):
LGTM.  Consider adding a comment for why you are blocking signals.

arungrover:
Thanks Bhroam!
I've added the comment.

bhroam (approved):

src/scheduler/pbs_sched_utils.cpp
```diff
@@ -623,19 +631,25 @@ connect_svrpool()
...
 			log_errf(pbs_errno, __func__, "Couldn't register the scheduler %s with the configured servers", sc_name);
 			/* wait for 2s for not to burn too much CPU, and then retry connection */
 			sleep(2);
 			close_servers();
-			continue;
+			goto CONTINUE;
 		}
 
 		/* Reached here means everything is success, so we will break out of the loop */
+BREAK:		if (sigprocmask(SIG_SETMASK, &prevsigs, NULL) == -1)
```
bhroam, line 646: Put the code below the label.  Also labels are usually lower case.  I'd rename them and use a lower case label.  How about unmask_break and unmask_continue

bhroam (changes requested):

src/scheduler/pbs_sched_utils.cpp
```diff
@@ -588,15 +597,15 @@ connect_svrpool()
 				/* wait for 2s for not to burn too much CPU, and then retry connection */
 				sleep(2);
 				close_servers();
-				continue;
+				goto CONTINUE;
```
agrawalravi90, line 600: How about moving the sleep(2) and close_servers() to the CONTINUE label, and unblocking the signals before going to sleep, so that scheduler doesn't block them while it's sleeping

agrawalravi90 (changes requested):

arungrover:
Thanks @bhroam  and @agrawalravi90, I've addressed your comments.

bhroam (approved):

agrawalravi90 (approved):

PR #2393, opened 2021-05-17
Fix for test_staging when mom and server are not collocated

toonen:
A fix to sys_copy() which corrects a problem with test_staging when the mom(s) and server are not collocated.   At the beginning of the test, test_staging calls self.mom.create_and_format_stagein_path() which in turn calls self.du.create_temp_file().  create_temp_file() copies the file to the mom node if the mom and server are not running on the same node.  test_staging then submits a job which results in the same file being copied again to the mom node via `scp -p` in sys_copy().  If create_temp_file() creates and then copies the temp file to the mom in the same second, then mtime of the existing file on the mom and the one later copied with `scp -p` in sys_copy() will be the same, and sys_copy() will return that the copy has failed.  Again, this is only an issue if the the server and mom are not on the same hosts.

This commit also includes a fix to the creation of a hosts set from the set of moms and the server specified to pbs_benchpress.  At present, if mom hosts are specified and none are the server host, then _cleanup() will add each character in the server name to the set of hosts rather than the server's hostname.

### DIFF
--- a/src/resmom/stage_func.c
+++ b/src/resmom/stage_func.c
@@ -1650,16 +1650,15 @@ sys_copy(int dir, int rmtflg, char *owner, char *src, struct rqfpair *pair, int 
 		}
 
 		/*
-		 ** There is a problem where scp can return zero
-		 ** indicating success even though the copy did
-		 ** not work.  In the case of a remote stagein,
-		 ** save the mtime of an existing file so we can
-		 ** check the mtime after the copy to be sure it
-		 ** worked.
+		 * There is a problem where scp can return zero indicating success even
+		 * though the copy did not work.  In the case of a remote stagein,
+		 * where the file already exists, save the ctime of an existing file so
+		 * we can check the ctime after the copy to be sure it worked.  ctime
+		 * is used instead of mtime because scp may reset mtime.
 		 */
 		if ((rmtflg != 0) && (dir == STAGE_DIR_IN)) {
 			if (stat(ag3, &sb) != -1)
-				original = sb.st_mtime;
+				original = sb.st_ctime;
 		}
 
 		DBPRT(("%s: %s %s %s %s\n", __func__, ag0, ag1, ag2, ag3))
@@ -1681,7 +1680,7 @@ sys_copy(int dir, int rmtflg, char *owner, char *src, struct rqfpair *pair, int 
 				if ((rc = WEXITSTATUS(rc)) == 0) {
 					if ((rmtflg != 0) && (dir == STAGE_DIR_IN)) {
 						if ((stat(ag3, &sb) == -1) ||
-							(original == sb.st_mtime))
+							(original == sb.st_ctime))
 							rc = 13;
 					}
 					return (rc);		/* good,  stop now */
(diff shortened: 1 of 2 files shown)


src/resmom/stage_func.c
```diff
@@ -1650,16 +1650,15 @@ sys_copy(int dir, int rmtflg, char *owner, char *src, struct rqfpair *pair, int
...
-		 ** There is a problem where scp can return zero
-		 ** indicating success even though the copy did
-		 ** not work.  In the case of a remote stagein,
-		 ** save the mtime of an existing file so we can
-		 ** check the mtime after the copy to be sure it
-		 ** worked.
+		 * There is a problem where scp can return zero indicating success even
+		 * though the copy did not work.  In the case of a remote stagein,
+		 * where the file already exists, save the ctime of an existing file so
+		 * we can check the ctime after the copy to be sure it worked.  ctime
```
arungrover, line 1656: I am a little confused here, scp is issued with "-p" options which should ideally preserve the m_time from the original file. Why are we switching to c_time?

toonen: For exactly that reason.  If the same file is copied with `scp -p`, the mtime will not change and thus the original code would not detect any change even if the file transfer were successful.  The test_staging test does exact that; stage in a file that had already exist and preserves the mtime.  As I noted in the comments, ctime changes anytime the file metadata is changed, which includes changing the mtime.

src/resmom/stage_func.c
```diff
@@ -1681,7 +1680,7 @@ sys_copy(int dir, int rmtflg, char *owner, char *src, struct rqfpair *pair, int
 				if ((rc = WEXITSTATUS(rc)) == 0) {
 					if ((rmtflg != 0) && (dir == STAGE_DIR_IN)) {
 						if ((stat(ag3, &sb) == -1) ||
-							(original == sb.st_mtime))
+							(original == sb.st_ctime))
 							rc = 13;
```
toonen, line 1684: @arungrover in case I wasn't clear earlier, this is where using ctime is important.  Because `scp -p` is used, mtime might not change in value, causing this code to report that the scp failed.  By using ctime, we can detect that the file metadata was changed, which I'm pretty sure always happens when the `-p` option is used, and assume as was done previous that the copy succeeded.

agrawalravi90:
"A fix to sys_copy() which corrects a **problem** with test_staging when the mom(s) and server are not collocated."

Can you please describe the problem in the PR description?

src/resmom/stage_func.c
arungrover: Thanks for the explanation @toonen , I am a little surprised that you are hitting this issue at all. Ideally, the PTL test you mentioned should ideally create a temp file and it is highly unlikely (but possible) that it will come up with a name of a file that has already been staged.
If we change the mtime to ctime, it is almost always going to pass, isn't it? Because with each scp, it will update ctime of the file when it is copied. So, should we even be doing this check?

toonen: I added more details about why a file with the same name already exists to the PR description.

As for the need for the check, the comment at line 1653 states that apparently scp can fail to perform the copy and still return success.  The mtime check was there as an attempt to catch this situation.  Unfortunately, `scp -p` resets mtime possibly back to the same as the original value.  Using ctime is more likely to catch whatever issue resulted in scp failing since it will increase in value when mtime is set.  BTW, if scp fails part way through the copy (having actually written some data) and returns success, then checking for a lack of change in mtime or ctime won't catch the issue.

I don't know what the original situation was when this check was added, so I can't say if it is still necessary.  I can say that in my test environment test_staging fails frequently when the server and mom are not collocated and that switching to ctime fixed the failure.

agrawalravi90 (approved):
Thanks for modifying the description, changes look good to me

arungrover (approved):
Thanks @toonen, LGTM!

PR #2395, opened 2021-05-18
Scheduler quits if any server is down in msvr setup

agrawalravi90:
#### Describe Bug or Feature

The commit https://github.com/openpbs/openpbs/commit/05e16c3419bb47340c1c5fefb6404d456085033a introduced a bug in scheduler's connect_svrpool() function which makes the code break out if even one of the servers is down (the label unmask_break caused the outer loop to break out instead of the inner for loop which was checking if all servers were connected to), and it ends up exiting the program incorrectly.

#### Describe Your Change

Removed the label which was causing the bug

### DIFF
--- a/src/scheduler/pbs_sched_utils.cpp
+++ b/src/scheduler/pbs_sched_utils.cpp
@@ -611,7 +611,7 @@ connect_svrpool()
 		for (i = 0; svr_conns_primary[i] != NULL && svr_conns_secondary[i] != NULL; i++) {
 			if (svr_conns_primary[i]->state == SVR_CONN_STATE_DOWN ||
 			    svr_conns_secondary[i]->state == SVR_CONN_STATE_DOWN)
-				goto unmask_break;
+				break;
 		}
 
 		if (i != get_num_servers()) {
@@ -629,12 +629,6 @@ connect_svrpool()
 			goto unmask_continue;
 		}
 
-		/* Reached here means everything is success, so we will break out of the loop */
-unmask_break:
-		if (sigprocmask(SIG_SETMASK, &prevsigs, NULL) == -1)
-			log_err(errno, __func__, "sigprocmask(SIG_SETMASK)");
-		break;
-
 unmask_continue:
 		if (sigprocmask(SIG_SETMASK, &prevsigs, NULL) == -1)
 			log_err(errno, __func__, "sigprocmask(SIG_SETMASK)");
@@ -642,6 +636,7 @@ unmask_continue:
 		close_servers();
 		continue;
 	}
+
 	log_eventf(PBSEVENT_ADMIN | PBSEVENT_FORCE, PBS_EVENTCLASS_SCHED,
 		   LOG_INFO, msg_daemonname, "Connected to all the configured servers");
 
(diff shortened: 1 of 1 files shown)


nithinj (approved):

bhroam (changes requested):
Your description and your code do not match.  Your description talks about if any server is down, the scheduler goes down.  All you have done is replace a unmasking break with a break.  This will cause the code to not unmask, so all the signals will be masked forever.

If I am misunderstanding what is going on, please update your description.

agrawalravi90:
@bhroam I did mention that it ends up exiting the program, which happens because of the check that follows the connect code:

	for (i = 0; svr_conns_secondary[i] != NULL; i++) {
		if (tpp_em_add_fd(poll_context, svr_conns_secondary[i]->sd, EM_IN | EM_HUP | EM_ERR) < 0) {
			log_errf(errno, __func__, "Couldn't add secondary connection to poll list for server %s",
				 svr_conns_secondary[i]->name);
			die(-1);
		}
	}

src/scheduler/pbs_sched_utils.cpp
```diff
@@ -629,19 +629,14 @@ connect_svrpool()
 			goto unmask_continue;
 		}
 
-		/* Reached here means everything is success, so we will break out of the loop */
-unmask_break:
-		if (sigprocmask(SIG_SETMASK, &prevsigs, NULL) == -1)
-			log_err(errno, __func__, "sigprocmask(SIG_SETMASK)");
-		break;
```
arungrover, line 636: Thanks for fixing this issue.
I don't think this change is needed. Before I changed it, it was breaking out of the outer for loop here. If you remove the break here then the while loop can potentially become endless.

arungrover (changes requested):

src/scheduler/pbs_sched_utils.cpp
agrawalravi90: oops, thanks for catching this, we need to add the break back.

arungrover: Also, retain the sigprocmask() call.

arungrover: If you don't add back the sigprocmask call, the signals will stay blocked.

agrawalravi90:
@bhroam  and @arungrover  I've added more explanation to the description to try to explain the problem, which is that the label was causing the outer loop to break instead of the inner loop, causing sched to come out of the outer loop without connecting to all servers, and causing it to exit in the check that followed.

arungrover (approved):
LGTM!

bhroam:
Thanks for explaining further, @agrawalravi90.  Could you explain why you removed the unmasking of the signals, but didn't add it elsewhere?  To me, this sounds like we'd mask the signals for this loop, but never unmask them.

bhroam (approved):
LGTM

bhroam:
@nithinj do you want to take another look, or should we merge this PR?

nithinj:
@bhroam ,
Looks good to me. Please proceed with merging!

PR #2398, opened 2021-05-19
Fixed resv overlap issue

bhroam:
#### Describe Bug or Feature
When multiple reservations were confirmed in the same cycle, it was possible for the scheduler to oversubscribe the same resources

#### Describe Your Change
When confirming a reservation, the scheduler creates a scratch universe to confirm the reservation.  When it has successfully confirmed the reservation, it copies the relevant data from the scratch reservation back to the real reservation.  The scheduler wasn't copying back the resv_nodes, so it would think the reservation didn't use any resources.  Since the scheduler didn't think it used any resources, those same resources could be used again to confirm another reservation.

Fix: copy back the resv_nodes (nspec array) to the real reservation.

I modified an existing test to show the bug.  I ran a full regression and it passed with no failures (including the modified test).

### DIFF
--- a/src/scheduler/resv_info.cpp
+++ b/src/scheduler/resv_info.cpp
@@ -1182,27 +1182,20 @@ check_new_reservations(status *policy, int pbs_sd, resource_resv **resvs, server
 								nresv_copy->select = new selspec(*nresv_copy->resv->select_standing);
 							}
 						}
-						/* Duplication deep-copies node info array. This array gets
-						 * overwritten and needs to be freed. This is an alternative
-						 * to creating another duplication function that only duplicates
-						 * the required fields.
-						 */
-						release_nodes(nresv_copy);
-
-						nresv_copy->resv->orig_nspec_arr = parse_execvnode(occr_execvnodes_arr[j], sinfo, nresv_copy->select);
-						nresv_copy->nspec_arr = combine_nspec_array(nresv_copy->resv->orig_nspec_arr);
-						nresv_copy->ninfo_arr = create_node_array_from_nspec(nresv_copy->nspec_arr);
-						nresv_copy->resv->resv_nodes = create_resv_nodes(nresv_copy->nspec_arr, sinfo);
 					}
+					release_nodes(nresv_copy);
+
+					nresv_copy->resv->orig_nspec_arr = parse_execvnode(occr_execvnodes_arr[j], sinfo, nresv_copy->select);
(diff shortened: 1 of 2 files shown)


arungrover (approved):
LGTM

src/scheduler/resv_info.cpp
```diff
@@ -1182,27 +1182,20 @@ check_new_reservations(status *policy, int pbs_sd, resource_resv **resvs, server
...
-						 * the required fields.
-						 */
-						release_nodes(nresv_copy);
-
-						nresv_copy->resv->orig_nspec_arr = parse_execvnode(occr_execvnodes_arr[j], sinfo, nresv_copy->select);
-						nresv_copy->nspec_arr = combine_nspec_array(nresv_copy->resv->orig_nspec_arr);
-						nresv_copy->ninfo_arr = create_node_array_from_nspec(nresv_copy->nspec_arr);
-						nresv_copy->resv->resv_nodes = create_resv_nodes(nresv_copy->nspec_arr, sinfo);
 					}
+					release_nodes(nresv_copy);
```
agrawalravi90, line 1186: I just want to confirm, in the first iteration of the loop, we set nresv_copy = sinfo->resvs[i];, by taking release_nodes() outside the if-else block, we'll call release_nodes() on  sinfo->resvs[i] for the first iteration. That won't cause any issues right?

agrawalravi90 (changes requested):
Just one clarification

src/scheduler/resv_info.cpp
bhroam: The first time this reservation is confirmed, release_nodes() will do nothing.  We then copy in where the reservation will run.  If we reconfirm the reservation, its location is moving, so we need to clear what is there and copy in the new location.  So no, this won't cause any problems.

bhroam:
I can run another regression to show you the results if you want, but I did run one.  It just took a couple of retries to get the counts to 0, so I didn't think that was too useful to show.  I thought saying the regression was successful would be enough.

agrawalravi90:
Can you just run that test once again and post the logs from it?

bhroam:
@agrawalravi90 
Here's a TH3 run of the whole reservation test suite (which includes the new test):
Description: Tests from given sources on platforms CENTOS8
Platforms: CENTOS8
Profile: regression
|--|--|--|--|--|--|--|
|7714|49|0|0|0|0|49|

agrawalravi90 (approved):
Thanks for providing the test logs

PR #2400, opened 2021-05-21
[WIP]Implement parallel tm_spawn (tm_spawn_multi)

nishiya:
#### Describe Bug or Feature

tm_spawn and pbsdsh can only start an executable on one mom at a time.  If an executable has to be sent to 10 moms, it would be sent one at a time.  We can speed things up and make things happen in parallel by sending to all of the moms at once using tm_spawn_multi.

#### Describe Your Change

Introducing tm_spawn_multi() that will multicast the execution request to all moms.

#### Link to Design Doc

https://openpbs.atlassian.net/wiki/spaces/PD/pages/2664988673/Proposal+for+tm+spawn+multi

Profile: regression
|--|--|--|--|--|--|--|
|7667|4176|0|0|0|0|4176|

### DIFF
--- a/src/lib/Libifl/tm.c
+++ b/src/lib/Libifl/tm.c
@@ -218,6 +218,7 @@ del_event(event_info *ep)
 
 		case TM_INIT:
 		case TM_SPAWN:
+		case TM_SPAWN_MULTI:
 		case TM_ATTACH:
 		case TM_SIGNAL:
 		case TM_OBIT:
@@ -802,6 +803,84 @@ tm_nodeinfo(tm_node_id **list, int *nnodes)
  * @param[in] argc - argument count
  * @param[in] argv - argument list
  * @param[in] envp - environment variable list
+ * @param[in] list_size - number of nodes in where
+ * @param[in] where - job relative node ID(s)
+ * @param[out] tids - task ID(s)
+ * @param[out] event - event info
+ *
+ * @return	int
+ * @retval	TM_SUCCESS	success
+ * @retval	TM_ER*		error
+ *
+ */
+int
+tm_spawn_multi(int argc, char **argv, char **envp,
+		int list_size, tm_node_id where[], tm_task_id *tids,
+		tm_event_t *event)
+{
+	char		*cp;
+	int		i;
+
+	if (!init_done)
+		return TM_BADINIT;
+	if (argc <= 0 || argv == NULL || argv[0] == NULL || *argv[0] == '\0')
+		return TM_ENOTFOUND;
+	
+	if (list_size < 1)
+		return TM_EBADENVIRONMENT; 
+
+	*event = new_event();
+	if (startcom(TM_SPAWN_MULTI, *event) != DIS_SUCCESS)
+		return TM_ENOTCONNECTED;
+	
+	/* send list_size */
+	if (diswui(local_conn, list_size) != DIS_SUCCESS)
+		return TM_ENOTCONNECTED;
+	
+	/* send where */
+	for (i = 0; i < list_size; i++) {
+		if (diswui(local_conn, where[i]) != DIS_SUCCESS)
+			return TM_ENOTCONNECTED;
+	}
+
+	if (diswsi(local_conn, argc) != DIS_SUCCESS)	/* send argc */
+		return TM_ENOTCONNECTED;
+
+	/* send argv strings across */
+
+	for (i=0; i < argc; i++) {
+		cp = argv[i];
+		if (diswcs(local_conn, cp, strlen(cp)) != DIS_SUCCESS)
+			return TM_ENOTCONNECTED;
+	}
+
+	/* send envp strings across */
+	if (envp != NULL) {
+		for (i=0; (cp = envp[i]) != NULL; i++) {
+#if defined(PBS_SECURITY) && (PBS_SECURITY == KRB5)
+			/* never send KRB5CCNAME; it would rewrite the value on target host */
+			if (strncmp(envp[i], "KRB5CCNAME", strlen("KRB5CCNAME")) == 0)
+				continue;
+#endif
+			if (diswcs(local_conn, cp, strlen(cp)) != DIS_SUCCESS)
+				return TM_ENOTCONNECTED;
+		}
+	}
+	if (diswcs(local_conn, "", 0) != DIS_SUCCESS)
+		return TM_ENOTCONNECTED;
+	dis_flush(local_conn);
+	/* Passing the Mother Superior */
+	add_event(*event, where[0], TM_SPAWN_MULTI, (void *)tids);
+	return TM_SUCCESS;
+}
+
+/**
+ * @brief
+ *	-Starts <argv>[0] with environment <envp> at <where>.
+ *
+ * @param[in] argc - argument count
+ * @param[in] argv - argument list
+ * @param[in] envp - environment variable list
  * @param[in] where - job relative node
  * @param[out] tid - task id
  * @param[out] event - event info
@@ -1336,6 +1415,7 @@ tm_poll(tm_event_t poll_event, tm_event_t *result_event, int wait, int *tm_errno
 	int		ret, mtype, nnodes;
 	int		prot, protver;
 	int		*obitvalp;
+	int		count = 0;
 	event_info	*ep = NULL;
 	tm_task_id	tid, *tidp;
 	tm_event_t	nevent;
@@ -1526,6 +1606,43 @@ tm_poll(tm_event_t poll_event, tm_event_t *result_event, int wait, int *tm_errno
 			}
 			tidp = (tm_task_id *)ep->e_info;
 			*tidp = new_task(tm_jobid, ep->e_node, tid);
+			break;
+
+		case TM_SPAWN_MULTI:
+			/*
+			** read (
+			** 	int		count
+			** 	tm_task_id	taskid 0
+			** 	tm_node_id	node 0
+			**	...
+			** 	tm_task_id	taskid count-1
+			** 	tm_node_id	node count-1
+			** )
+			*/
+			count = disrui(local_conn, &ret);
+			if (ret != DIS_SUCCESS) {
+				DBPRT(("%s: SPAWN_MULTI failed count\n", __func__))
+				goto err;
+			}
+			for (i = 0; i < count; i++) {
+				/*
+				** Read a task ID and node ID for each
+				** task that was spawned
+				*/
+				tid = disrui(local_conn, &ret);
+				if (ret != DIS_SUCCESS) {
+					DBPRT(("%s: SPAWN_MULTI failed tid\n", __func__))
+					goto err;
+				}
+
+				node = disrui(local_conn, &ret);
+				if (ret != DIS_SUCCESS) {
+					DBPRT(("%s: SPAWN_MULTI failed node\n", __func__))
+					goto err;
+				}
+				tidp = ((tm_task_id *)ep->e_info)+i;
+				*tidp = new_task(tm_jobid, node, tid);
+			}
 			break;
 
 		case TM_SIGNAL:
--- a/src/lib/Libifl/tm.c
+++ b/src/lib/Libifl/tm.c
@@ -815,7 +815,7 @@ tm_nodeinfo(tm_node_id **list, int *nnodes)
  */
 int
 tm_spawn_multi(int argc, char **argv, char **envp,
-		int list_size, tm_node_id where[], tm_task_id *tids,
+		int list_size, tm_node_id where[], tm_task_id tids[],
 		tm_event_t *event)
 {
 	char		*cp;
--- a/src/resmom/mom_comm.c
+++ b/src/resmom/mom_comm.c
@@ -122,6 +122,8 @@ write_pipe_data(int upfds, void *data, int data_size);
 char task_fmt[] = "/%8.8X";
 extern void resume_multinode(job *pjob);
 
+static tm_task_id taskident;
+
 /* Function pointers
  **
  ** These are functions to provide extra interaction between mother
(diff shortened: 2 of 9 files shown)


agrawalravi90:
Do you have any numbers which show the boost in performance?

src/resmom/mom_comm.c
```diff
@@ -6156,14 +6525,459 @@ tm_request(int fd, int version)
...
+				}
+			
+				/* send body of the message: 
+				** start with the identifier,
+				** then the number of items in node_list, 
+				** then send the list of where in a loop
+				** then argc, argv, env, etc.
+				*/
+				ret = diswui(mtfd, taskident);
+				if (ret != DIS_SUCCESS) {
```
subhasisb, line 6798: I think we can avoid the if (ret != DIS_SUCCESS) blocks being repeated. How about we check for ret being DIS_SUCCESS in every next step, like so:

ret = diswui(mtfd, taskident);
if (ret == DIS_SUCCESS) 
            ret = diswui(mtfd, list_size);
for (i = 0; i < list_size, ret == DIS_SUCCESS; i++)
           ret = diswui(mtfd, node_list[i]);
...

src/server/job_func.c
```diff
@@ -357,6 +357,7 @@ job_alloc(void)
 	CLEAR_HEAD(pj->ji_multinodejobs);
 	pj->ji_extended.ji_ext.ji_stdout = 0;
 	pj->ji_extended.ji_ext.ji_stderr = 0;
+	pj->ji_spawninfo = NULL;
```
subhasisb, line 360: Since we are setting every field member of job by hand, should we remove the memset on top. That just seems to be wasteful

src/resmom/mom_comm.c
```diff
@@ -6156,14 +6525,459 @@ tm_request(int fd, int version)
...
+					index++;
+				} else {
+					/*
+					 ** Sending to another MOM.
+					 */
+					/* 
+					** Have we already added a stream for this mom?
+					** We don't want to open another stream to the same mom.
+					*/	
+					for (k = 0; k < momcnt; k++) {
```
subhasisb, line 6714: I think this check in a loop is not required. tpp_open() will prevent opening a duplicate stream to the same target. If the stream already exists (opened), then tpp_open() will return the same stream fd back.

src/resmom/mom_comm.c
```diff
@@ -6156,14 +6525,459 @@ tm_request(int fd, int version)
...
+					if (phost->hn_stream < 0) {
+						sprintf(log_buffer, "tpp_open failed on %s:%d",
+							phost->hn_host, phost->hn_port);
+						log_joberr(-1, __func__, log_buffer, jobid);
+						continue;
+					}
+				
+					if (pbs_conf.pbs_use_mcast == 1 && !alreadyadded) {
+						/* add each of the tpp streams to the tpp mcast channel */
+						if (tpp_mcast_add_strm(mtfd, phost->hn_stream, FALSE) == -1) {
```
subhasisb, line 6735: U can call tpp_mcast_add_strm( mtfd, stream, unique) - the third parameter can be set to true which allows tpp itself to check if the stream was already added or not. So the fd will be added if the stream was not already added (if unique was set to true)

src/resmom/mom_comm.c
```diff
@@ -6156,14 +6525,459 @@ tm_request(int fd, int version)
...
+							log_joberr(-1, __func__, log_buffer, jobid);
+							continue;
+						}
+						/* add the mom to the list of moms who have streams added */
+						streamlist[momcnt] = tvnodeid;
+						momcnt++;
+					}
+					/* Add the IM_SPAWN_MULTI event to the list */	
+					if (nep == NULL) {
+						nep = event_alloc(pjob, IM_SPAWN_MULTI, fd, phost,
```
subhasisb, line 6748: are we still following the 80 char wrap arounds? I was  not - and in modern day displays it does not seem to make sense?

subhasisb (changes requested):
BTW, do we not need to update pbs_tmrsh() as well?

nishiya:
> BTW, do we not need to update pbs_tmrsh() as well?

I don't think we need to.  When I looked at pbs_tmrsh() as far as I could tell it is only called for one hostname.  It will not benefit from tm_spawn_multi.

subhasisb:
> > BTW, do we not need to update pbs_tmrsh() as well?
> 
> I don't think we need to. When I looked at pbs_tmrsh() as far as I could tell it is only called for one hostname. It will not benefit from tm_spawn_multi.

Yes, you are right. tmrsh itself in the current form will not be able to use this.

src/server/job_func.c
nishiya: Interesting question.  I took a quick look to see if I could find any fields from the job structure that were not specifically listed in in job_alloc.  By looking at the first 10-15 fields there were already a handful that were relying on the memset (i.e. not listed specifically in job_alloc()).  I would not recommend removing memset without a deeper dive.

src/resmom/mom_comm.c
```diff
@@ -4719,6 +4976,108 @@ im_request(int stream, int version)
...
+						tempnid[index] = tvnodeid;
+						index++;
+					}
+					/* We heard back from one more sister */
+					siscnt--;
+
+					/* Don't reply until we heard back from all sisters */
+					if (siscnt != 0) {
+						tempinfo->ji_num_sisters = siscnt;
+						tempinfo->ji_taskid_index=index;
```
vstumpf, line 5046: whitespace issue

src/resmom/mom_comm.c
```diff
@@ -4719,6 +4976,108 @@ im_request(int stream, int version)
...
+						(void)diswui(efd, temptaskid[i]);
+						(void)diswui(efd, tempnid[i]);
+					}
+					(void)dis_flush(efd);
+					tempinfo->ji_num_sisters = 0;
+					tempinfo->ji_taskid_index = 0;
+					free(tempinfo->ji_taskid_list);
+					free(tempinfo->ji_nid_list);
+					tempinfo->ji_taskid_list = NULL;
+					tempinfo->ji_nid_list = NULL;
```
vstumpf, line 5078: Should you free tempinfo here?

src/resmom/mom_comm.c
```diff
@@ -6156,14 +6525,459 @@ tm_request(int fd, int version)
...
+					(void)diswui(fd, temptaskid[k]);
+					(void)diswui(fd, tempnid[k]);
+				}
+				(void)dis_flush(fd);
+				/* Reset the task counts */
+				tempinfo->ji_num_sisters = 0;
+				tempinfo->ji_taskid_index = 0;
+				/* These aren't needed anymore */
+				free(temptaskid);
+				free(tempnid);
```
vstumpf, line 6825: should you free tempinfo here?

src/lib/Libifl/tm.c
```diff
@@ -795,6 +796,84 @@ tm_nodeinfo(tm_node_id **list, int *nnodes)
...
+ * @retval	TM_SUCCESS	success
+ * @retval	TM_ER*		error
+ *
+ */
+int
+tm_spawn_multi(int argc, char **argv, char **envp,
+		int list_size, tm_node_id where[], tm_task_id *tids,
+		tm_event_t *event)
+{
+	char		*cp;
```
vstumpf, line 821: some weird whitespace

src/resmom/mom_comm.c
```diff
@@ -3941,6 +3954,250 @@ im_request(int stream, int version)
...
+			if (ret != DIS_EOD) {
+				arrayfree(argv);
+				arrayfree(envp);
+				free(node_list);
+				sprintf(log_buffer, "SPAWN_MULTI read of envp array");
+				goto err;
+			}
+
+			/* Allocate a temporary place to store task IDs */
+                	taskidlist = (tm_task_id *)calloc(list_size, sizeof(tm_task_id));
```
vstumpf, line 4091: whitespace

vstumpf (changes requested):
some initial comments

src/resmom/mom_comm.c
nishiya: Sorry, I don't see what whitespace issue you are referring to...

src/resmom/mom_comm.c
nishiya: No, tempinfo is freed when the job structure is freed as part of job_free()

src/resmom/mom_comm.c
nishiya: No, tempinfo is freed when the job structure is freed

src/resmom/mom_comm.c
vstumpf: Ah, here it's the spaces around the `=`

src/resmom/mom_comm.c
```diff
@@ -3941,6 +3954,250 @@ im_request(int stream, int version)
...
+				ptask->ti_qs.ti_parenttask = fromtask;
+				if (task_save(ptask) == -1) {
+					sprintf(log_buffer, "unable to save "
+						"task info\n");
+					log_event(PBSEVENT_DEBUG3,
+						PBS_EVENTCLASS_JOB, LOG_DEBUG,
+						jobid, log_buffer);
+					break;
+				}
+				errcode = start_process(ptask, argv, envp, false);
```
subhasisb, line 4126: I think this is the culprit!! Can we do the work of starting the process AFTER we have ack'd (sent IM_ALL_OKAY)?  
Of course, we have a small risk that start_process() actually fails later, even though we would have ack'd okay, but that is as good as the process starting and failing immediately (or similar). In fact, we do the same for job starts in the mom anyway, we ack the send from the server server before actually starting the job/task. So, it does make sense in a parallel spawn to be asynchronous in that way.

src/resmom/mom_comm.c
```diff
@@ -3941,6 +3954,250 @@ im_request(int stream, int version)
...
+					taskidlist[created] = ptask->ti_qs.ti_task;
+					created++;
+				}
+			} /* end of for loop of node_list */
+
+			/* We finished going through the node_list, now send im_compose IM_ALL_OKAY */
+			ret = im_compose(stream, jobid, cookie, IM_ALL_OKAY,
+					event, fromtask, IM_OLD_PROTOCOL_VER);
+			if (ret != DIS_SUCCESS) {
+				arrayfree(argv);
```
subhasisb, line 4146: You can consolidate the "if ret != DIS_SUCCESS" checks to just one block...
ret = im_compose(...);
if (ret == DIS_SUCCESS) diswsi(..);
if (ret == DIS_SUCCESS) diswsi(..);
...
...
if (ret != DIS_SUCCESS) {
     arrayfree(argv);
     arrayfree(envp);
     ....
}

src/resmom/mom_comm.c
```diff
@@ -6156,14 +6524,332 @@ tm_request(int fd, int version)
...
+			taskident++;
+			
+			tempinfo->ji_ident = taskident;
+			index = 0;
+			
+			/* open the tpp mcast channel */
+			if ((mtfd = tpp_mcast_open()) == -1) {
+				sprintf(log_buffer, "mcast open failed");
+				log_joberr(-1, __func__, log_buffer, jobid);
+				free(node_list);
```
subhasisb, line 6636: you could free them in the err block

src/resmom/mom_comm.c
```diff
@@ -6156,14 +6524,332 @@ tm_request(int fd, int version)
...
+				}
+				for (i = 0; envp[i] && ret == DIS_SUCCESS; i++) {
+					ret = diswst(mtfd, envp[i]);
+				}
+				if (ret == DIS_SUCCESS) {
+					ret = (dis_flush(mtfd) == -1) ?
+						DIS_NOCOMMIT : DIS_SUCCESS;
+				}
+				if (ret != DIS_SUCCESS) {
+					arrayfree(argv);
```
subhasisb, line 6794: we could consolidate all the frees to the err / done blocks

src/server/job_func.c
subhasisb: hmm - then there are certainly fields which we are setting to 0, even if memset did the job. We should delete those lines - small, but when we want thousands of job per second, every nanosecond counts :) That said, this is a general thought, not for this PR, so you may ignore it.

subhasisb (changes requested):

src/resmom/mom_comm.c
```diff
@@ -122,6 +122,8 @@ write_pipe_data(int upfds, void *data, int data_size);
 char task_fmt[] = "/%8.8X";
 extern void resume_multinode(job *pjob);
 
+static tm_task_id taskident;
```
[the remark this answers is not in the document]
nishiya: static variables are initialized

src/lib/Libifl/tm.c
nishiya: Sorry, I'm not finding it.

src/resmom/mom_comm.c
nishiya: Now I see what you are asking, yes, I can also free tempinfo here.

src/resmom/mom_comm.c
nishiya: Since there are not many, I'd rather just take care of it here.  There is a history of each case statement taking care of freeing what it allocated.

src/resmom/mom_comm.c
nishiya: Hmm... I thought about this some more and I think you are right, we should free in the err and done blocks.  I was just following the existing code style, but I think your suggestion is better, so we don't leak memory in the future if code is added that returns without freeing.

src/resmom/mom_comm.c
nishiya: never mind, I decided it made sense to do it in err/done after all.

nishiya:
@agrawalravi90, sorry for taking so long to get back to you.  I finally got some numbers.
I ran with 50 moms all running on one large machine.  I submit a job that does a pbsdsh -c 5000 /usr/bin/true, which basically means each mom has to run 100 tasks.   I ran this 20 times and took an average of those runs, these are the rough numbers:
before - it took 8.3M microseconds to send and receive task IDs (at this point the tasks have not necessarily completed).  After my changes - it took 3.1K microseconds to send and receive task IDs.

PR #2402, opened 2021-05-22
Multi-server: Enhancing sched to work with partition tolerance, 1

agrawalravi90:
#### Describe Bug or Feature

This enhances scheduler to work even when one/more servers down in a multi-server setup. This is first of a multi-part change. Here, we are enhancing sched to not stall scheduling when one/more servers are down. This doesn't handle multi-node job spanning server boundaries, this will be handled in further PRs.

```
[root@pbspro ravi]# PBS_CONF_FILE=/etc/pbs_1.conf /etc/init.d/pbs start
Starting PBS
/opt/pbs/sbin/pbs_comm ready (pid=83015), Proxy Name:pbspro:17001, Threads:4
PBS comm
PBS mom
PBS sched
Connecting to PBS dataservice...connected to PBS dataservice@pbspro
PBS server
[ravi@pbspro ~]$ PBS_SERVER_INSTANCES=pbspro:15001 qsub -- /bin/sleep 10000
1400.pbspro
[ravi@pbspro ~]$ qstat
qstat: cannot connect to server pbspro runs on port 18000 (errno=111)
1200.pbspro       STDIN            ravi              00:00:00 R workq           
501.pbspro        STDIN            ravi              00:00:00 R workq           
1300.pbspro       STDIN            ravi              00:00:00 R workq           
1400.pbspro       STDIN            ravi                     0 R workq           
[ravi@pbspro ~]$ qdel 1200.pbspro 1300.pbspro
qdel: cannot connect to server pbspro runs on port 18000 (errno=111)
[root@pbspro ravi]# PBS_CONF_FILE=/etc/pbs_2.conf /etc/init.d/pbs start
Starting PBS
Connecting to PBS dataservice...connected to PBS dataservice@pbspro
PBS server
[root@pbspro ravi]# 
[ravi@pbspro ~]$ PBS_SERVER_INSTANCES=pbspro:18000 qsub -- /bin/sleep 10000
601.pbspro
[ravi@pbspro ~]$ sudo /opt/pbs/bin/qmgr -c 's s scheduling=t'
[ravi@pbspro ~]$ qstat
501.pbspro        STDIN            ravi              00:00:00 R workq           
1400.pbspro       STDIN            ravi              00:00:00 R workq           
601.pbspro        STDIN            ravi                     0 R workq           
[ravi@pbspro ~]$ ps -ef | grep pbs_server
root     83182     1  0 02:07 ?        00:00:00 /opt/pbs/sbin/pbs_server.bin
root     83435     1  0 02:08 ?        00:00:00 /opt/pbs/sbin/pbs_server.bin
ravi     83477    36  0 02:10 pts/0    00:00:00 grep --color=auto pbs_server
[ravi@pbspro ~]$ sudo kill 83435
[ravi@pbspro ~]$ qstat
qstat: cannot connect to server pbspro runs on port 18000 (errno=15031)
501.pbspro        STDIN            ravi              00:00:00 R workq           
1400.pbspro       STDIN            ravi              00:00:00 R workq           
601.pbspro        STDIN            ravi              00:00:00 R workq           
[ravi@pbspro ~]$ PBS_SERVER_INSTANCES=pbspro:15001 qsub -- /bin/sleep 10000
1500.pbspro
[ravi@pbspro ~]$ qstat
qstat: cannot connect to server pbspro runs on port 18000 (errno=111)
501.pbspro        STDIN            ravi              00:00:00 R workq           
1400.pbspro       STDIN            ravi              00:00:00 R workq           
601.pbspro        STDIN            ravi              00:00:00 R workq           
1500.pbspro       STDIN            ravi                     0 Q workq  
[ravi@pbspro ~]$ sudo /opt/pbs/bin/qmgr -c 's s scheduling=t'
[ravi@pbspro ~]$ qstat
qstat: cannot connect to server pbspro runs on port 18000 (errno=111)
501.pbspro        STDIN            ravi              00:00:00 R workq           
1400.pbspro       STDIN            ravi              00:00:00 R workq           
601.pbspro        STDIN            ravi              00:00:00 R workq           
1500.pbspro       STDIN            ravi                     0 R workq 
```
**Preemption with 1st server down:**
```
[ravi@pbspro ~]$ PBS_SERVER_INSTANCES=pbspro:15001 qsub -- /bin/sleep 1000
2600.pbspro
[ravi@pbspro ~]$ qstat
qstat: cannot connect to server pbspro runs on port 18000 (errno=111)
2600.pbspro       STDIN            ravi              00:00:00 R workq           
[ravi@pbspro ~]$ PBS_SERVER_INSTANCES=pbspro:15001 qsub -lncpus=3 -- /bin/sleep 1000
2700.pbspro
[ravi@pbspro ~]$ qstat
qstat: cannot connect to server pbspro runs on port 18000 (errno=111)
2600.pbspro       STDIN            ravi              00:00:00 R workq           
2700.pbspro       STDIN            ravi                     0 R workq           
[ravi@pbspro ~]$ PBS_SERVER_INSTANCES=pbspro:15001 qsub -q highp -- /bin/sleep 1000
2800.pbspro
[ravi@pbspro ~]$ qstat
qstat: cannot connect to server pbspro runs on port 18000 (errno=111)
2600.pbspro       STDIN            ravi              00:00:00 R workq           
2700.pbspro       STDIN            ravi              00:00:00 S workq           
2800.pbspro       STDIN            ravi                     0 R highp     
```

**preemption with both servers up, preempting jobs from both**:
```
[ragrawal@m1 ~]$ qstat
900.m1            STDIN            ragrawal          00:00:00 R workq           
1000.m1           STDIN            ragrawal          00:00:00 R workq           
501.m1            STDIN            ragrawal          00:00:00 R workq           
601.m1            STDIN            ragrawal          00:00:00 R workq           
[ragrawal@m1 ~]$ qsub -q highp -lselect=2:ncpus=4 -- /bin/sleep 100
1100.m1
[ragrawal@m1 ~]$ qstat
900.m1            STDIN            ragrawal          00:00:00 S workq           
1000.m1           STDIN            ragrawal          00:00:00 S workq           
1100.m1           STDIN            ragrawal                 0 R highp           
501.m1            STDIN            ragrawal          00:00:00 S workq           
601.m1            STDIN            ragrawal          00:00:00 S workq 
```

Regression Results:
Profile: regression
|--|--|--|--|--|--|--|
|7696|4178|0|0|0|15|4163|

### DIFF
--- a/src/include/libpbs.h
+++ b/src/include/libpbs.h
@@ -146,10 +146,8 @@ int get_num_servers(void);
 typedef struct svr_conn {
 	int sd;                      /* File descriptor for the open socket */
 	int state;                   /* Connection state */
-	time_t last_used_time;       /* Last used time for the connection */
 	char name[PBS_MAXSERVERNAME + 1];  /* server name */
 	int port;                    /* server port */
-	int from_sched;              /* flag to indicate whether this conn is from sched or not */
 } svr_conn_t;
 
 typedef struct svr_conns_list {
@@ -414,6 +412,9 @@ int get_server_fd_from_jid(int c, char *jobid);
 int multi_svr_op(int fd);
 int get_job_svr_inst_id(int c, char *job_id);
 
+int pbs_register_sched_single_msvr(const char *sched_id, int primary_conn_id, int secondary_conn_id);
+void pbs_connect_single_msvr(svr_conn_t *conn);
+int pbs_disconnect_single_msvr(svr_conn_t *svr_conn);
 #ifdef __cplusplus
 }
 #endif
--- a/src/include/libpbs.h
+++ b/src/include/libpbs.h
@@ -412,9 +412,9 @@ int get_server_fd_from_jid(int c, char *jobid);
 int multi_svr_op(int fd);
 int get_job_svr_inst_id(int c, char *job_id);
 
-int pbs_register_sched_single_msvr(const char *sched_id, int primary_conn_id, int secondary_conn_id);
-void pbs_connect_single_msvr(svr_conn_t *conn);
-int pbs_disconnect_single_msvr(svr_conn_t *svr_conn);
+int pbs_register_sched_msvr_instance(const char *sched_id, int primary_conn_id, int secondary_conn_id);
+void pbs_connect_msvr_instance(svr_conn_t *conn);
+int pbs_disconnect_msvr_instance(svr_conn_t *svr_conn);
 #ifdef __cplusplus
 }
 #endif
--- a/src/lib/Libifl/pbsD_Preempt_Jobs.c
+++ b/src/lib/Libifl/pbsD_Preempt_Jobs.c
@@ -87,7 +87,8 @@ preempt_jobs_send(int connect, char **preempt_jobs_list)
 	}
 
 	if (dis_flush(connect)) {
-		pbs_errno = PBSE_PROTOCOL;
+		if (pbs_errno == PBSE_NONE)
+			pbs_errno = PBSE_PROTOCOL;
 		return 1;
 	}
 	return 0;
@@ -110,9 +111,7 @@ preempt_jobs_recv(int connect, int *ret_count)
 	preempt_job_info *ppj_reply = NULL;
 
 	reply = PBSD_rdrpy(connect);
-	if (reply == NULL)
-		pbs_errno = PBSE_PROTOCOL;
-	else {
+	if (reply != NULL) {
 		int i = 0;
 		int count = 0;
 		preempt_job_info *ppj_temp = NULL;
@@ -156,9 +155,10 @@ __pbs_preempt_jobs(int c, char **preempt_jobs_list)
 	int i;
 	int count = 0;
 	int last_count = -1;
-	int retidx = 0;	/* first server's list will be used to collate results from all */
+	int retidx = -1;	/* first server's list will be used to collate results from all */
 	preempt_job_info *ret = NULL;
 	void *missing_jobs = NULL;
+	int num_servers = get_num_servers();
 
 	if (!svr_connections)
 		return NULL;
@@ -187,10 +187,17 @@ __pbs_preempt_jobs(int c, char **preempt_jobs_list)
 	/* With multi-server, jobs are sharded across multiple servers.
 	 * So, send the request to all the active servers and collate their replies
 	 */
-	for (i = 0; i < get_num_servers(); i++) {
-		if (preempt_jobs_send(svr_connections[i]->sd, preempt_jobs_list) != 0)
-			goto err;
+	for (i = 0; i < num_servers; i++) {
+		if (svr_connections[i]->sd > 0) {
+			if (preempt_jobs_send(svr_connections[i]->sd, preempt_jobs_list) != 0) {
+				goto err;
+			}
+			if (retidx == -1)
+				retidx = i;
+		}
 	}
+	if (retidx == -1)
+		goto err;
 
 	/**
 	 * Each server will return an array containing all jobs
@@ -206,8 +213,8 @@ __pbs_preempt_jobs(int c, char **preempt_jobs_list)
 	 * 	- Using the AVL tree avoids the cost of N strcmps otherwise needed to find the jobid
 	 *  in the first array
 	 */
-	for (i = 0; i < get_num_servers(); last_count = count, i++) {
-		if ((p_replies[i] = preempt_jobs_recv(svr_connections[i]->sd, &count)) == NULL)
+	for (i = 0; i < num_servers; last_count = count, i++) {
+		if (svr_connections[i]->sd > 0 && (p_replies[i] = preempt_jobs_recv(svr_connections[i]->sd, &count)) == NULL)
 			goto err;
 		if (last_count != -1 && count != last_count) {
 			/* something went wrong, this server did not return the same count, abort */
@@ -215,9 +222,9 @@ __pbs_preempt_jobs(int c, char **preempt_jobs_list)
 		}
 	}
 
-	/* Find jobs which couldn't be found on the first server and store them in AVL tree */
+	/* Find jobs which couldn't be found on the retidx server and store them in AVL tree */
 	for (i = 0; i < count; i++) {
-		if (p_replies[retidx][i].order[0] == 'D' && get_num_servers() > 1) {
+		if (p_replies[retidx][i].order[0] == 'D' && num_servers > 1) {
 			int *data = malloc(sizeof(int));
 			if (data == NULL) {
 				pbs_errno = PBSE_SYSTEM;
@@ -229,14 +236,14 @@ __pbs_preempt_jobs(int c, char **preempt_jobs_list)
 	}
 	ret = p_replies[retidx];
 
-	if (missing_jobs != NULL && get_num_servers() > 1) {
-		for (i = 1; i < get_num_servers(); i++) {	/* Starting from 1 as retidx == 0 */
+	if (missing_jobs != NULL && num_servers > 1) {
+		for (i = retidx + 1; i < num_servers; i++) {
 			int j;
 
 			if (p_replies[i] == NULL)
 				continue;
 
-			/* Check if the missing jobs were found on any other server */
+			/* Check if the missing jobs were found on this server */
 			for (j = 0; j < count; j++) {
 				if (p_replies[i][j].order[0] != 'D') {
 					/* Job was found on another server!
@@ -261,8 +268,9 @@ __pbs_preempt_jobs(int c, char **preempt_jobs_list)
 		goto err;
 
 	/* Free up all the other lists */
-	for (i = 1; i < get_num_servers(); i++) {
-		free(p_replies[i]);
+	for (i = 0; i < num_servers; i++) {
+		if (i != retidx)
+			free(p_replies[i]);
 	}
 	free(p_replies);
 	pbs_idx_destroy(missing_jobs);
@@ -271,7 +279,7 @@ __pbs_preempt_jobs(int c, char **preempt_jobs_list)
 
 err:
 	if (p_replies != NULL) {
-		for (i = 0; i < get_num_servers(); i++) {
+		for (i = 0; i < num_servers; i++) {
 			free(p_replies[i]);
 		}
 		free(p_replies);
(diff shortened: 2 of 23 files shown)


src/include/libpbs.h
```diff
@@ -414,6 +412,9 @@ int get_server_fd_from_jid(int c, char *jobid);
 int multi_svr_op(int fd);
 int get_job_svr_inst_id(int c, char *job_id);
 
+int pbs_register_sched_single_msvr(const char *sched_id, int primary_conn_id, int secondary_conn_id);
```
nithinj, line 415: why single and msvr both appear in the name?

src/lib/Libifl/pbsD_connect.c
```diff
@@ -618,6 +621,43 @@ connect_to_servers(char *svrhost, uint port, char *extend_data)
...
 
+/**
+ * @brief	(Re)Connect to a single server in a multi-server setup
+ *
+ * @param[in,out]	conn - svr_conn_t object for the server to be connected to
+ *
+ * @return void
+ */
+void
+pbs_connect_single_msvr(svr_conn_t *conn)
```
nithinj, line 632: can you rename it to pbs_connect_msvr_instance or similar?

src/lib/Libifl/pbsD_connect.c
```diff
@@ -618,6 +621,43 @@ connect_to_servers(char *svrhost, uint port, char *extend_data)
...
+ * @brief	Disconnect from a single multi-server instance
+ *
+ * @param[in,out]	svr_conn - connection structure for the instance
+ *
+ * @return	int
+ * @retval	0 for success
+ * @retval	-1 for error
+ */
+int
+pbs_disconnect_single_msvr(svr_conn_t *svr_conn)
```
nithinj, line 647: can you rename it to pbs_disconnect_msvr_instance or similar?

nithinj (changes requested):

src/scheduler/pbs_sched_utils.cpp
```diff
@@ -684,13 +744,56 @@ sched_svr_init(void)
...
 
-	close_servers();
-	connect_svrpool();
+		return;
+	}
+
+	part_tolerance = true;
+
+	if (time(NULL) - last_reconnect < min_reconnect_period) {
+		return;	/* To throttle the reconnect frequency, don't connect if under min_reconnect_period */
```
arungrover, line 767: Unlock the mutex. We should consider making a "Locker" class that takes a given mutex in the constructor and releases it in the destructor. That way we don't have to worry about forgetting to release mutexes.

src/scheduler/pbs_sched_utils.cpp
```diff
@@ -684,13 +744,56 @@ sched_svr_init(void)
...
+		close_servers();
+		connect_svrpool();
+		pthread_mutex_unlock(&cleanup_lock);
 
-	close_servers();
-	connect_svrpool();
+		return;
+	}
+
+	part_tolerance = true;
```
arungrover, line 764: Shouldn't we be setting partition_tolerance after we have checked min_reconnect_period? Can it happen that reconnect_servers is called frequently twice and it makes the scheduler go into partition tolerance mode even when no server was down?

src/scheduler/pbs_sched_utils.cpp
```diff
@@ -684,13 +744,56 @@ sched_svr_init(void)
...
+	svr_conn_t **prim_conn = get_conn_svr_instances(clust_primary_sock);
+	svr_conn_t **sec_conn = get_conn_svr_instances(clust_secondary_sock);
+	int num_down = 0;
+	static time_t last_reconnect = -1;
+	const int min_reconnect_period = 30;	/* Reconnect at least 30secs after last */
+
 	pthread_mutex_lock(&cleanup_lock);
+	if (prim_conn == NULL || sec_conn == NULL || get_num_servers() == 1) {
+		close_servers();
+		connect_svrpool();
```
arungrover, line 758: Since connect_svrpool can also set part_tolerance to True, I feel we should always set it to False here if the number of servers configured is only 1.

src/scheduler/sched_ifl_wrappers.cpp
```diff
@@ -153,7 +177,11 @@ send_attr_updates(int virtual_sd, resource_resv *resresv, struct attrl *pattr)
 preempt_job_info *
 send_preempt_jobs(int virtual_sd, char **preempt_jobs_list)
 {
-    return pbs_preempt_jobs(virtual_sd, preempt_jobs_list);
+	preempt_job_info *ret;
+
+    ret = pbs_preempt_jobs(virtual_sd, preempt_jobs_list);
```
arungrover, line 182: As of today, pbs_preempt_jobs already detects if it is not able to read data from one of the servers it clears the responses from other servers too. Are you going to address that in later PRs?

src/scheduler/sched_ifl_wrappers.cpp
```diff
@@ -153,7 +177,11 @@ send_attr_updates(int virtual_sd, resource_resv *resresv, struct attrl *pattr)
...
+
+	return static_cast<preempt_job_info *>(handle_part_tolerance(ret));
```
arungrover, line 184: This can potentially become a leak if upon a sigpipe this function returns 0 and does not free up responses in "ret" variable.
It can also happen in all other stat functions below.

arungrover (changes requested):

src/scheduler/pbs_sched_utils.cpp
```diff
@@ -534,6 +534,60 @@ are_we_primary()
...
+		close_msvr_conn(prim_conn, sec_conn);	/* Close both connections if either is down */
+		return 1;
+	}
+
+	if (pbs_register_sched_single_msvr(sc_name, prim_conn->sd, sec_conn->sd) != 0) {
+		close_msvr_conn(prim_conn, sec_conn);
+		return 1;
+	}
+
+	/* Both primary and secondary successful, add secondary to poll list */
```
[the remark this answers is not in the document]
agrawalravi90: Actually this means to say that both were successfully connected, so we can add secondary to the poll list. I'll make it more clear.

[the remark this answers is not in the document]
arungrover: ah, no worries, I misread it.

src/scheduler/pbs_sched_utils.cpp
```diff
@@ -610,28 +666,30 @@ connect_svrpool()
 
 		for (i = 0; svr_conns_primary[i] != NULL && svr_conns_secondary[i] != NULL; i++) {
 			if (svr_conns_primary[i]->state == SVR_CONN_STATE_DOWN ||
-			    svr_conns_secondary[i]->state == SVR_CONN_STATE_DOWN)
-				break;
+			    svr_conns_secondary[i]->state == SVR_CONN_STATE_DOWN) {
+				/* At least one server is down, operate in partition tolerance mode */
+				part_tolerance = true;
```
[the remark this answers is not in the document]
agrawalravi90: this is not secondary server, it's the secondary connection to the same server. I think it's safer to disconnect both primary and secondary connections when either of them couldn't be connected, so I think I'll make that change. We will try reconnecting both connections after every 30 seconds, so it should be ok to be conservative. Let me know what you think.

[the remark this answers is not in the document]
arungrover: Sorry, I again misread it. If it is the secondary connection then you are right. There is no point in keeping partially connected server, better to reconnect both connections.

src/scheduler/pbs_sched_utils.cpp
agrawalravi90: Great idea, but as we discussed, let's wait until we refactor code to use std:thread and std:mutex, then we can use STL's std::lock_guard or std::unique_lock to do this.

src/scheduler/pbs_sched_utils.cpp
agrawalravi90: reconnect_servers() should only be called when one/more servers is down, so I wanted to set part_tolerance upfront .. but I'll move it to the if condition that checks how many are down ..

src/scheduler/pbs_sched_utils.cpp
agrawalravi90: connect_svrpool() sets it to False the first thing, so we should be ok.

src/scheduler/sched_ifl_wrappers.cpp
agrawalravi90: Thanks for pointing this out, I hadn't thought about it, yes, it'll need to be updated.

agrawalravi90:
@arungrover  and @nithinj  thanks for your review, I've addressed your comments. I also found a bug in preemption with multi-server, we were not updating svr_to_psets in update_all_nodepart, which was causing preemption to fail as the node partition could show free_nodes=0 even when some nodes were free'd up by the simulation. Please check it out and let me know if it looks okay.

src/lib/Libifl/pbsD_Preempt_Jobs.c
```diff
@@ -156,9 +155,10 @@ __pbs_preempt_jobs(int c, char **preempt_jobs_list)
 	int i;
 	int count = 0;
 	int last_count = -1;
-	int retidx = 0;	/* first server's list will be used to collate results from all */
+	int retidx = -1;	/* first server's list will be used to collate results from all */
 	preempt_job_info *ret = NULL;
 	void *missing_jobs = NULL;
+	int num_servers = get_num_servers();
```
arungrover, line 161: you can use num_servers on line 166 as well

arungrover (approved):
I just have this minor comment. LGTM!

agrawalravi90:
@arungrover  thanks! I've made a fix to server local scheduling, which was using psets and seeing that a superchunk was split across nodes of different hosts because allpart had the ok_break bit set. I changed it to be set to false by default, which fixed the issue. Please look at the latest changes.

@nithinj please provide further feedback

src/lib/Libifl/pbsD_Preempt_Jobs.c
```diff
@@ -87,7 +87,8 @@ preempt_jobs_send(int connect, char **preempt_jobs_list)
 	}
 
 	if (dis_flush(connect)) {
-		pbs_errno = PBSE_PROTOCOL;
+		if (pbs_errno == PBSE_NONE)
```
nithinj, line 90: why checking for PBSE_NONE? I could not see this check with any of the errors above.

nithinj (commented):
Rest looks good to me!

src/lib/Libifl/pbsD_Preempt_Jobs.c
nithinj: Does dis layer set pbs error code? I don't think so. but if that's the case you might want to add the check to other places in the function.

nithinj (approved):
LGTM!

arungrover (approved):

PR #2404, opened 2021-05-25
multi-server: jobs which span boundaries stay queued when it is eligible to run

nithinj:
#### Describe Bug or Feature
multi-server: jobs which span boundaries stay queued when it is eligible to run as initially servers doesn't have the node cache. It will get in the next scheduling cycle.

#### Describe Your Change
Server will fetch the node cache when it comes up.
Server will inform the scheduler once it receives the cache to trigger another cycle.

[test.log](https://github.com/openpbs/openpbs/files/6541716/test.log)

Profile: regression
|--|--|--|--|--|--|--|
|7699|4178|1|5|0|15|4157|

Description: Rerun Only Failed Tests From #7699
Platforms: UBUNTU1804,SLES12
Profile: regression
|--|--|--|--|--|--|--|
|7712|6|0|0|0|0|6|

### DIFF
--- a/src/server/multi_svr.c
+++ b/src/server/multi_svr.c
@@ -613,8 +613,9 @@ connect_to_peersvr(void *psvr)
 	if (send_connect(psvr) < 0)
 		return -1;
 
-	if (resc_upd_reqd && svr_info->ps_pending_replies) {
-		mcast_resc_update_all(psvr);
+	if (resc_upd_reqd) {
+		if (svr_info->ps_pending_replies)
+			mcast_resc_update_all(psvr);
 		send_nodestat_req();
 	}
 
--- a/src/server/multi_svr.c
+++ b/src/server/multi_svr.c
@@ -613,8 +613,10 @@ connect_to_peersvr(void *psvr)
 	if (send_connect(psvr) < 0)
 		return -1;
 
-	if (resc_upd_reqd && svr_info->ps_pending_replies)
+	if (resc_upd_reqd && svr_info->ps_pending_replies) {
 		mcast_resc_update_all(psvr);
+		send_nodestat_req();
+	}
 
 	return 0;
 }
@@ -1028,6 +1030,10 @@ update_node_cache(int stream, struct batch_status *bstat)
 	clear_node_cache(psvr);
 
 	init_node_from_bstat(bstat, psvr);
+
+	/* Trigger default scheduler to schedule the job which has failed due
+	to missing vnode cache information */
+	set_scheduler_flag(SCH_SCHEDULE_NEW, dflt_scheduler);
 
 	return 0;
 }
--- a/src/server/multi_svr.c
+++ b/src/server/multi_svr.c
@@ -319,6 +319,9 @@ log_msvr_stat()
 void
 update_msvr_stat(ulong val, msvr_stat_type_t type)
 {
+	if (type < 0 || type >= END_OF_STAT)
+		return;
+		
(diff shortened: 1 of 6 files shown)


src/server/multi_svr.c
```diff
@@ -1029,6 +1032,10 @@ update_node_cache(int stream, struct batch_status *bstat)
 
 	init_node_from_bstat(bstat, psvr);
 
+	/* Trigger default scheduler to schedule the job which has failed due
+	to missing vnode cache information */
+	set_scheduler_flag(SCH_SCHEDULE_NEW, dflt_scheduler);
```
agrawalravi90, line 1037: why only default? I think server should trigger sched cycles for all schedulers that it knows about, otherwise there can be jobs on other scheds which the default sched will ignore

src/server/multi_svr.c
```diff
@@ -613,8 +613,11 @@ connect_to_peersvr(void *psvr)
 	if (send_connect(psvr) < 0)
 		return -1;
 
-	if (resc_upd_reqd && svr_info->ps_pending_replies)
-		mcast_resc_update_all(psvr);
+	if (resc_upd_reqd) {
+		if (svr_info->ps_pending_replies)
+			mcast_resc_update_all(psvr);
+		send_nodestat_req();
```
agrawalravi90, line 619: Shouldn't we remove the other calls to send_nodestat_req()? are they still needed?

src/server/svr_comm.c
```diff
@@ -271,15 +271,15 @@ send_nodestat_req(void)
 
 	update_msvr_stat(1, CACHE_MISS);
 
+	if (mtfd == -1)
+		return;
```
agrawalravi90, line 275: shouldn't this be done before the call to update_msvr_stat() ?

agrawalravi90 (changes requested):

src/server/multi_svr.c
nithinj: We should think about whether we need to support multi-sched in its current form with multi-svr. We can revisit this if we decide to do so.

src/server/multi_svr.c
nithinj: yes, send_nodestat_req() gets invoked when node cache becomes invalid

src/server/svr_comm.c
nithinj: update_msvr_stat() is updating a potential cache miss. that has to be done anyway except when it is initially invoked. I'll add the changes for that.

src/server/svr_comm.c
```diff
@@ -257,29 +259,30 @@ replyhello_psvr(struct work_task *ptask)
...
 	struct attrl *head;
 	struct attrl *tail;
 	struct attrl *nxt;
 	int mtfd = open_ps_mtfd();
-	int rc;
 	static int time_last_sent = 0;
+	int rc;
 
-	update_msvr_stat(1, CACHE_MISS);
+	update_msvr_stat(1, type);
```
ashwathraop, line 275: All the calls of send_nodestat_req() take CACHE_MISS, so, why did you change this to access CACHE_MISS through `type` variable?

nithinj: type can also be -1. Which is handled within update_msvr_stat instead of the caller.

agrawalravi90 (approved):
LGTM

ashwathraop (approved):

PR #2407, opened 2021-05-27
Change sched to not consider prerunning job as running

arungrover:
#### Describe Bug or Feature
Scheduler does not differentiate between pre-running and running jobs and considers them all as running. This works fine for most cases but in the case of finding the fairshare usage of jobs, it causes trouble. Since prerunning jobs don't have any resources_used.<res> attribute set on them, scheduler ends up using the resources_list (requested resources) attributes.
This results in a sudden jump in the resource usage on an entity thereby depleting the entity's chances of running a job.

#### Describe Your Change
A pre-running job should ideally not be considered a running job. This change ensures that jobs in pre-running substate don't make it into running_jobs list.

#### Link to Design Doc
NA

Profile: regression
|--|--|--|--|--|--|--|
|7760|4178|1|0|0|0|4177|

The test that failed in regression is not related to fairshare.

### DIFF
--- a/src/scheduler/fifo.cpp
+++ b/src/scheduler/fifo.cpp
@@ -349,7 +349,7 @@ init_scheduling_cycle(status *policy, int pbs_sd, server_info *sinfo)
 				if (user != NULL) {
 					auto rj = find_resource_resv(sinfo->running_jobs, lj.name);
 
-					if (rj != NULL && rj->job != NULL && rj->job->is_prerunning != 1) {
+					if (rj != NULL && rj->job != NULL && !rj->job->is_prerunning) {
 						/* just in case the delta is negative just add 0 */
 						delta = formula_evaluate(conf.fairshare_res.c_str(), rj, rj->job->resused) -
 							formula_evaluate(conf.fairshare_res.c_str(), rj, lj.resused);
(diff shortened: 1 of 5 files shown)


bhroam (commented):
Only one minor nit: You need to copy is_prerunning in dup_job_info()

bhroam (approved):
LGTM

bhroam:
Azure Pipelines failed because a few smoke tests failed.  You might want to look into that.

arungrover:
Yeah, i saw those failures, That's happening because now that jobs in 41 substates don't make it into running_jobs array, limits count is off.
I think I'll have to change the implementation, to only make the change for fairshare and not affect other features. I can put a prerunning check when we are calculating delta for running jobs and let these jobs remain in running_jobs array. That way limits code will remain unaffected. What do you think?

arungrover:
@bhroam  please review this implementation, let me know what you think.

src/scheduler/fifo.cpp
```diff
@@ -349,7 +349,7 @@ init_scheduling_cycle(status *policy, int pbs_sd, server_info *sinfo)
 				if (user != NULL) {
 					auto rj = find_resource_resv(sinfo->running_jobs, lj.name);
 
-					if (rj != NULL && rj->job != NULL) {
+					if (rj != NULL && rj->job != NULL && rj->job->is_prerunning != 1) {
```
bhroam, line 352: is_prerunning is a bool.  Say either !rj->job->is_prerunning, or rj->job->is_prerunning == false

vstumpf (approved):
LGTM

bhroam (approved):
LGTM

PR #2408, opened 2021-05-27
Use pbs_sleep inside PTL tests

sakshamgarg:
#### Describe Bug or Feature

PTL framework already uses pbs_sleep in job submission. Doing the same for tests which uses sleep in job scripts or as executable inside them

#### Describe Your Change

Replacing /bin/sleep in PTL tests with pbs_sleep.
The job scripts doesn't require full path of pbs_sleep but the tests which use sleep as executable, they require full path. Hence using full path of pbs_sleep from MoM class to make it platform independent.

Description: Rerun All Tests From #7694
Profile: regression
|--|--|--|--|--|--|--|
|7719|173|0|0|0|2|171|

### DIFF
--- a/test/tests/functional/pbs_hooksmoketest.py
+++ b/test/tests/functional/pbs_hooksmoketest.py
@@ -54,7 +54,7 @@ class TestHookSmokeTest(TestFunctional):
 
         self.script = []
         self.script += ['echo Hello World\n']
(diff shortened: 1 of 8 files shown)


arungrover (approved):
LGTM

raviv1995 (approved):

PR #2410, opened 2021-06-02
Multi-Server: Discard job for job spanning boundaries

nithinj:
#### Describe Bug or Feature
qdel -Wforce / qrerun -Wforce does not delete the job on mom's belongs to other partitions when mother superior is down

#### Describe Your Change
Discard job is a special message by the server to discard a job from all moms participating in that job. In multi-server cases, some of those moms can belong to other partitions. The server will not be able to contact those moms and hence the job won't get discarded.

This change will broadcast the discard job request to peer servers if it cannot reach the target mother superior.

[test.log](https://github.com/openpbs/openpbs/files/6611496/test.log)

Profile: regression
|--|--|--|--|--|--|--|
|7768|4178|2|0|0|0|4176|

Description: Rerun Only Failed Tests From #7768
Profile: regression
|--|--|--|--|--|--|--|
|7778|3|0|0|0|0|3|

### DIFF
--- a/src/server/multi_svr.c
+++ b/src/server/multi_svr.c
@@ -878,6 +878,7 @@ open_ps_mtfd_for_execvnode(char *exec_vnode)
 		if (parse_node_resc(chunk, &noden, &nelem, &pkvp) == 0) {
 			if ((pnode = find_nodebyname(noden)) == NULL &&
 			    (pnode = find_alien_node(noden)) == NULL) {
+				/* Broadcast when we dont have the node cache */
 				close_streams(mtfd, rc);
 				mtfd = open_ps_mtfd();
 				return mtfd;
--- a/src/server/multi_svr.c
+++ b/src/server/multi_svr.c
@@ -320,7 +320,7 @@ update_msvr_stat(ulong val, msvr_stat_type_t type)
 {
 	if (type < 0 || type >= END_OF_STAT)
 		return;
-		
+
 	msvr_stat.stat[type] += val;
 
 	log_msvr_stat();
@@ -851,6 +851,42 @@ req_resc_update(int stream, pbs_list_head *ru_head, void *psvr)
 	*/
 	if (op == INCR)
 		send_command(stream, PS_RSC_UPDATE_ACK);
+}
+
+/**
+ * @brief open a multicast fd for peer servers corresponds to execvnode
+ * 
+ * @param[in] exec_vnode - exec vnode
+ * 
+ * @return int
+ * @retval -1 : for failure
+ */
+int
+open_ps_mtfd_for_execvnode(char *exec_vnode)
+{
+	int mtfd = -1;
+	char *chunk;
+	int rc = 0;
+	char *noden;
+	int nelem;
+	struct key_value_pair *pkvp;
+	struct pbsnode *pnode;
+
+	for (chunk = parse_plus_spec(exec_vnode, &rc);
+	     chunk && !rc; chunk = parse_plus_spec(NULL, &rc)) {
+
+		if (parse_node_resc(chunk, &noden, &nelem, &pkvp) == 0) {
+			if ((pnode = find_nodebyname(noden)) == NULL &&
+			    (pnode = find_alien_node(noden)) == NULL) {
+				close_streams(mtfd, rc);
+				mtfd = open_ps_mtfd();
+				return mtfd;
+			} else
+				mcast_add(get_peersvr_from_svrid(get_nattr_str(pnode, ND_ATR_server_inst_id)), &mtfd, TRUE);
+		}
+	}
+
+	return mtfd;
 }
 
 /**
--- a/src/server/multi_svr.c
+++ b/src/server/multi_svr.c
@@ -876,8 +876,11 @@ open_ps_mtfd_for_execvnode(char *exec_vnode)
 	     chunk && !rc; chunk = parse_plus_spec(NULL, &rc)) {
 
 		if (parse_node_resc(chunk, &noden, &nelem, &pkvp) == 0) {
-			if ((pnode = find_nodebyname(noden)) == NULL &&
-			    (pnode = find_alien_node(noden)) == NULL) {
+
+			if ((pnode = find_nodebyname(noden)))
+				continue;
+
+			if ((pnode = find_alien_node(noden)) == NULL) {
 				/* Broadcast when we dont have the node cache */
 				close_streams(mtfd, rc);
 				mtfd = open_ps_mtfd();
--- a/src/server/node_manager.c
+++ b/src/server/node_manager.c
@@ -2394,6 +2394,7 @@ discard_job(job *pjob, char *txt, int noack)
 	struct pbsnode *pnode;
 	int	 rc;
 	int	 rver;
+	int	broadcast = 0;
 
 	/* We're about to discard the job, reply to a preemption.
 	 * This serves as a catch all just incase the code doesn't reply on its own.
@@ -2462,7 +2463,8 @@ discard_job(job *pjob, char *txt, int noack)
 				}
 				nmom++;
 			}
-		}
+		} else
+			broadcast = 1;
 		pn = parse_plus_spec(NULL, &rc);
 	}
 
@@ -2470,7 +2472,7 @@ discard_job(job *pjob, char *txt, int noack)
 	rver = get_jattr_long(pjob, JOB_ATR_run_version);
 
 	/* ask each peer server to handle discard job for its share of moms running the job */
-	if (!(pjob->ji_qs.ji_svrflags & JOB_SVFLG_AlienJob))
+	if (broadcast && !(pjob->ji_qs.ji_svrflags & JOB_SVFLG_AlienJob))
 		ps_send_discard(pjob->ji_qs.ji_jobid,
 				get_jattr_str(pjob, JOB_ATR_exec_vnode),
 				get_jattr_str(pjob, JOB_ATR_exec_host), rver);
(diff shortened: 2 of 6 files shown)


nithinj:
Requesting reviews from @agrawalravi90 , @subhasisb

src/server/node_manager.c
```diff
@@ -2466,15 +2470,21 @@ discard_job(job *pjob, char *txt, int noack)
 		pn = parse_plus_spec(NULL, &rc);
 	}
 
+	/* Get run vervion of this job */
+	rver = get_jattr_long(pjob, JOB_ATR_run_version);
+
+	if ((mom_superior->mi_dmn_info->dmn_state & INUSE_DOWN) &&
```
agrawalravi90, line 2476: Pardon my lack of familiarity with discard job, but I'm a bit confused by this check. It seems like discard_job()'s purpose is to have the server(s) send IS_DISCARD_JOB to each mom in case even one of the moms for a job is down. So, even if mother superior is up, it doesn't seem like MS will keep trying to send discard/delete job to the down sister(s) until they come back up, it seems that's the responsibility of the server, not MS. If my understanding is correct, then this check would be incorrect, you'd want the server(s) to ensure that all moms receive IS_DISCARD_JOB, regardless of the MS. Can you please clarify?

src/server/svr_comm.c
```diff
@@ -310,6 +310,112 @@ send_nodestat_req(enum msvr_stats_type type)
...
+		goto err;
+
+	if ((rc = diswcs(mtfd, jobid, strlen(jobid))) != 0)
+		goto err;
+
+	if ((rc = diswcs(mtfd, execvnode, strlen(execvnode))) != 0)
+		goto err;
+
+	if ((rc = diswcs(mtfd, exechost, strlen(exechost))) != 0)
+		goto err;
```
agrawalravi90, line 349: We already sent exec_vnode to the peer server with PS_RSC_UPDATE, do we really need to send it again?

src/server/svr_comm.c
```diff
@@ -310,6 +310,112 @@ send_nodestat_req(enum msvr_stats_type type)
...
+int
+ps_send_discard(char *jobid, char *execvnode, char *exechost, int rver)
+{
+	int mtfd = -1;
+	int rc;
+
+	if (!exechost)
+		return -1;
+
+	mtfd = open_ps_mtfd();
```
agrawalravi90, line 334: so which peer server will this request be sent to?

src/server/svr_comm.c
```diff
@@ -310,6 +310,112 @@ send_nodestat_req(enum msvr_stats_type type)
...
+		goto end;
+
+	pjob = calloc(1, sizeof(job));
+	pbs_strncpy(pjob->ji_qs.ji_jobid, jid, sizeof(pjob->ji_qs.ji_jobid));
+	pjob->ji_qs.ji_svrflags |= JOB_SVFLG_AlienJob;
+	set_jattr_str_slim(pjob, JOB_ATR_exec_vnode, execvnode, NULL);
+	set_jattr_str_slim(pjob, JOB_ATR_exec_host, exechost, NULL);
+	set_jattr_l_slim(pjob, JOB_ATR_run_version, rver, SET);
+
+	discard_job(pjob, "", 1);
```
agrawalravi90, line 408: why not use the 'txt' argument to provide a log message? Maybe something like "Discard request for non local job"?

agrawalravi90 (changes requested):

src/server/node_manager.c
nithinj: I could not find a reason why the server is sending the discard to all moms when the MS is up.
If the MS receives discard, it will send delete job to its sisters which will do the purpose. If some of the sisters are down they will get the request from MS when they come up.
So we should remove the case where the server sends the request to all moms when the MS is up. But that can be taken up independently.

src/server/svr_comm.c
nithinj: The server will broadcast the request to all peer servers. We should be able to optimize it by reusing an already opened stream by saving it in the job struct.

src/server/svr_comm.c
nithinj: PS_RSC_UPDATE and PS_DISCARD_JOB are independent messages. It is not necessary that a resource update will be sent when a discard job is sent and vice versa.

src/server/node_manager.c
subhasisb: > I could not find a reason why the server is sending the discard to all moms when the MS is up.
> If the MS receives discard, it will send delete job to its sisters which will do the purpose. If some of the sisters are down they will get the request from MS when they come up.
> So we should remove the case where the server sends the request to all moms when the MS is up. But that can be taken up independently.

There is a reason. It could be that the mother superior is damaged and cannot be up or will late a huge amount of time to get up. If so, the job will remain stuck on the sisters and cannot be deleted, so a force delete makes the server talk to the sisters directly to reclaim the resources blocked by the job.

agrawalravi90: >If some of the sisters are down they will get the request from MS when they come up.

Can you please point me to the code in mom which keeps a track of which sisters haven't responded to delete job request and sends it again when they come up? I couldn't find anything on the mom side to ensure this.

Regardless, I think I agree with Subhasis, it's safer for the server to ensure that discard_job reaches every mom, this is a rare but critical case, so it's important to be correct rather than performant.

subhasisb: It's a question of "how long" the resources are blocked. MS could track it, but she herself is down, then the sister resources have no other way of getting cleaned up other than a direct intervention from server. And that intervention is "IS_DISCARD"

src/server/svr_comm.c
agrawalravi90: The job might span over multiple servers so we'd need to store multiple streams .. let's wait until we see a performance problem and then fix this, for now broadcast should be ok.

src/server/node_manager.c
agrawalravi90: ya makes sense ..

src/server/svr_comm.c
nithinj: It will be a multicast fd. no need to store multiple streams. But we could take it up separately.

agrawalravi90 (changes requested):
From your test logs, it seems like you didn't test that the down mom, when it comes back up, gets the discard job as well .. can you please show logs for that as well?

Also, please test the scenario where you kill a sister mom, not MS, and show that discard job reaches the sister mom when it comes back up.

nithinj:
@agrawalravi90 ,
the scenarios mentioned do not involve peer-server communication and are covered in the regression. I'll update the log as I test more.

agrawalravi90:
> the scenarios mentioned do not involve peer-server communication and are covered in the regression

The regression won't have any scenarios where there are multiple servers, but if you want to make it more interesting then can you try out the following:
- 2 servers, one mom each
- multi-node job spanning across both nodes
- kill sister mom
- discard job is issued by server (qdel -Wforce), only MS receives it, not sister mom as it's down
- a second later, kill MS itself
- bring sister mom back up, should now receive discard job from the server, not MS
- bring MS back up
- submit another job which spans both nodes and verify that it runs as the previous was successfully discarded

nithinj:
@agrawalravi90 ,
I've updated the test logs.
The server will never send discard twice. If all nodes/some of the nodes are down, then it is up to the MS to handle it.

agrawalravi90 (changes requested):
@nithinj I just tested this out, it seems like when a mom is down and comes back up, it discards the job, regardless of whether MS is up or not. So, the scenario i was thinking of already works ok. Your change will ensure that when MS goes down, all sister moms get a discard from the server(s), so this looks good.

It seems like you added changes to send the discard job only to select servers. Looking at the change, I'm not sure if it'll be better or worse than before. Now you are parsing exec_vnode to fetch out each vnode, checking if it's alien or not, parsing server_inst_id to get peer-server and then adding it to mcast. For jobs which are spanning servers, the exec_vnode might be huge .. so this might actually be worse than just broadcasting discard job to all servers. If we want to do this then we should instead create a cache of "jobid: [server1, server2,.. etc.]" at run time so that we can just fetch the servers where a job is running and send discard_job to them. But I think for now we can leave it as broadcast to all. What do you think?

nithinj:
@agrawalravi90 ,
The change about broadcasting is keeping in mind that IO is expensive than computation. Yes, the originating server walks through the execvnode to decide where to send, but it is saving from the destination server (one or more) from doing the same. Each destination server has to walk through this list when to decide where to send the update even if it doesn't have any mom belongs to.

src/server/multi_svr.c
```diff
@@ -853,6 +853,42 @@ req_resc_update(int stream, pbs_list_head *ru_head, void *psvr)
...
+
+	for (chunk = parse_plus_spec(exec_vnode, &rc);
+	     chunk && !rc; chunk = parse_plus_spec(NULL, &rc)) {
+
+		if (parse_node_resc(chunk, &noden, &nelem, &pkvp) == 0) {
+			if ((pnode = find_nodebyname(noden)) == NULL &&
+			    (pnode = find_alien_node(noden)) == NULL) {
+				close_streams(mtfd, rc);
+				mtfd = open_ps_mtfd();
+				return mtfd;
```
agrawalravi90, line 883: Does this make it so that we broadcast to all if the node cache doesn't have info about a foreign node? Please add a comment above this block to explain the behavior.

agrawalravi90 (approved):
> Yes, the originating server walks through the execvnode to decide where to send, but it is saving from the destination server (one or more) from doing the same. Each destination server has to walk through this list when to decide where to send the update even if it doesn't have any mom belongs to.

Ah, got it, thanks for clarifying. Ideally, I think each server should keep track of the "foreign" jobs which are currently consuming its resources, I think it might come handy for fast lookups, but for now this looks good.

Just one minor comment, but I'll approve.

ashwathraop (approved):
LGTM

PR #2412, opened 2021-06-03
Python tool to convert system monitoring data from test output json fie to csv

visheshh:
#### Describe Bug or Feature
Convert system monitoring data from PTL in json file to easily readable csv format

#### Describe Your Change
Python script which takes JSON file as input and converts system monitoring data to csv file. 

Ex : csv file 

[sys_report_jts.csv.txt](https://github.com/openpbs/openpbs/files/6593627/sys_report_jts.csv.txt)
[sys_report_new.csv.txt](https://github.com/openpbs/openpbs/files/6621827/sys_report_new.csv.txt)

### DIFF
--- a/test/fw/bin/pbs_sys_report
+++ b/test/fw/bin/pbs_sys_report
@@ -43,16 +43,7 @@ import json
 import sys
 import os
 import getopt
-
-
-def usage():
-    msg = []
-    msg += ['Usage: ' + os.path.basename(sys.argv[0])]
-    msg += [' [PTL_JSON_FILE] \n\n']
-    msg += [' System monitoring results to csv conversion tool']
-    msg += ['--output-file : path to generate file\n']
-    msg += ['--help or -h : To display usage information\n']
-    print(''.join(msg))
+import argparse
 
 
 def write_to_csv(head, csvrows, i, filepath):
@@ -64,28 +55,13 @@ def write_to_csv(head, csvrows, i, filepath):
 
 if __name__ == '__main__':
 
-    if len(sys.argv) < 2:
-        usage()
-        sys.exit(1)
-
-    try:
-        opts, args = getopt.getopt(sys.argv[2:], "h",
-                                   ["help", "output-file="])
-    except getopt.GetoptError as err:
-        print(err)
-        usage()
-        sys.exit(1)
-
-    with open(sys.argv[1]) as fp:
-        td = json.load(fp)
-
-    filepath = None
-    for o, val in opts:
-        if o in ("-h", "--help"):
-            usage()
-            sys.exit(0)
-        elif o == "--output-file":
-            filepath = val
+    msg = 'System monitoring results to csv conversion tool'
+    parser = argparse.ArgumentParser(description=msg)
+    parser.add_argument('PTL_JSON_FILE', help='path to ptl_test_results.json')
+    parser.add_argument('--outputfile', nargs='?',
+                        help='path to generate file')
+    args = parser.parse_args()
+    filepath = args.outputfile
 
     if not filepath:
         filepath = "sys_report.csv"
@@ -93,7 +69,9 @@ if __name__ == '__main__':
     if os.path.exists(filepath):
         os.remove(filepath)
 
-    i = 0
+    with open(args.PTL_JSON_FILE) as fp:
+        td = json.load(fp)
+
     for k, v in sorted(td['testsuites'].items()):
         _otcs = v['testcases']
         for _k, _v in sorted(v['testcases'].items()):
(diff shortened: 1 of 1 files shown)


shilpakodli (commented):
Have argument parser to show help if user passes any other arguments. also type of the file user should input.

test/fw/bin/pbs_sys_report
```diff
@@ -0,0 +1,96 @@
...
+        csv.writer(fp).writerow(csvrows)
+
+
+if __name__ == '__main__':
+
+    with open(sys.argv[1]) as fp:
+        td = json.load(fp)
+
+    if os.path.exists("sys_report.csv"):
+        os.remove("sys_report.csv")
```
bhroam, line 60: Consider using argparse.  You can then create options for the input and output files.  I would rather not see the output file be hard coded.

test/fw/bin/pbs_sys_report
```diff
@@ -0,0 +1,96 @@
...
+                for item in val['measurements']:
+                    for ky, vl in item.items():
+                        if ky == 'procs':
+                            timelist = []
+                            name = ''
+                            time = ''
+                            for mes in vl:
+                                for nm, value in mes.items():
+                                    if nm == 'time':
+                                        time = value
```
bhroam, line 78: While you don't import it, time is a module name.  Consider renaming this variable.

test/fw/bin/pbs_sys_report
```diff
@@ -0,0 +1,96 @@
...
+                                timelist = list(set(timelist))
+                            timelist.sort()
+                            for tm in timelist:
+                                head = ['Time']
+                                csvrows = []
+                                csvrows.append(tm)
+                                for mes in vl:
+                                    for nm, value in mes.items():
+                                        if nm == 'name':
+                                            name = value
```
bhroam, line 89: isn't this just mes['name']?

test/fw/bin/pbs_sys_report
```diff
@@ -0,0 +1,96 @@
...
+                            timelist = []
+                            name = ''
+                            time = ''
+                            for mes in vl:
+                                for nm, value in mes.items():
+                                    if nm == 'time':
+                                        time = value
+                                        timelist.append(value)
+                                timelist = list(set(timelist))
+                            timelist.sort()
```
bhroam, line 81: Since you only really care about item['procs'], why not just use that?  Why iterate over the whole thing?  Also, if procs is not in item.items(), you won't initialize timelist, and this will barf.

bhroam (changes requested):

test/fw/bin/pbs_sys_report
visheshh: measurements is a list of dictionaries, so iterate through the whole thing to find procs key in the list of dictionaries.

test/fw/bin/pbs_sys_report
```diff
@@ -42,22 +42,56 @@ import csv
...
 
 if __name__ == '__main__':
 
+    if len(sys.argv) < 2:
+        usage()
+        sys.exit(1)
+
+    try:
+        opts, args = getopt.getopt(sys.argv[2:], "h",
+                                   ["help", "output-file="])
```
bhroam, line 73: I still suggest looking into argparse over getopt().  It allows it wraps up the usage and all the arguments when you define the parser.  There's a thing called a positional argument you can use for the json file.  You won't have to check if the output file is specified, because you can set a default when you define the parser.

This is just a suggestion though.  I'll sign off.

bhroam (approved):

test/fw/bin/pbs_sys_report
visheshh: Updated file with argparse

visheshh:
@shilpakodli @bhroam Please review the changes

test/fw/bin/pbs_sys_report
```diff
@@ -64,36 +55,23 @@ def write_to_csv(head, csvrows, i, filepath):
...
-        if o in ("-h", "--help"):
-            usage()
-            sys.exit(0)
-        elif o == "--output-file":
-            filepath = val
+    msg = 'System monitoring results to csv conversion tool'
+    parser = argparse.ArgumentParser(description=msg)
+    parser.add_argument('PTL_JSON_FILE', help='path to ptl_test_results.json')
+    parser.add_argument('--outputfile', nargs='?',
+                        help='path to generate file')
```
bhroam: This should be nargs=1, default='sys_report.csv'

bhroam (changes requested):

test/fw/bin/pbs_sys_report
visheshh: @bhroam added default, adding nargs would make it a list of 1 variable, without nargs its just a string.

shilpakodli (approved):
Changes LGTM. Please attach the csv file with latest changes.

visheshh:
> 
> 
> Changes LGTM. Please attach the csv file with latest changes.

attached

bhroam (approved):
LGTM

PR #2416, opened 2021-06-08
TestReservations.test_ASAP_resv_with_multivnode_job is failing intermittently

SudeshnaMoulik:
#### Describe Bug or Feature

TestReservations.test_ASAP_resv_with_multivnode_job is failing intermittently on few platforms.

#### Describe Your Change

For ASAP reservation, there could be scenario, when job finishes earlier than the reservation and the corresponding mom sends the server obit before the server could send TermJob signal to mom.
So in that scenario, PTL test fails as it keep checking for "job_state = F && substate = 91".

This PR is for checking for "job_state = F" (irrespective of substate is 91 or 92).

Profile: regression
|--|--|--|--|--|--|--|
|7807|2050|0|1|0|0|2049|

### DIFF
--- a/test/tests/functional/pbs_reservations.py
+++ b/test/tests/functional/pbs_reservations.py
@@ -2090,7 +2090,7 @@ class TestReservations(TestFunctional):
         jids = [jid1, jid2]
         for job in jids:
             self.server.expect(JOB, 'queue', op=UNSET, id=job)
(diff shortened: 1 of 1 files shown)


shilpakodli (approved):
LGTM

bhroam (approved):
LGTM

PR #2421, opened 2021-06-11
Fixing the server crash while moving job across queue

bremanandjk:
#### Describe Bug or Feature
Server crashes when jobs are moved across queue with differing priority.

#### Describe Your Change

The dangling pointer is causing the server to walk into unintended code and causing it core dump. Setting the NULL to an uninitialized pointer would stave off the crash.

Steps to reproduce:

1) Run the server in Valgrind, submit a job that spans across 2 nodes.
2) Submit 2 more jobs that depend on the last submitted job.
3) Delete the first dependent job.
4) Make the mom down and that causes the parent's job to requeue.
5) Once requeued, move the parent job across the queue.
6) Move the second dependent job across the queue, which causes the job state to change from "H" to "Q" state.
7) Move back the second dependent job to the first submitted queue, which causes the below crash at the accounting function.

[Test_logs.txt](https://github.com/openpbs/openpbs/files/6639273/Test_logs.txt)

### DIFF
--- a/src/server/accounting.c
+++ b/src/server/accounting.c
@@ -587,7 +587,7 @@ acct_job(const job *pjob, int type, char *buf, int len)
 		len -= i;
 	} else if (is_jattr_set(pjob, JOB_ATR_depend)) {
 		pbs_list_head phead;
-		svrattrl *svrattrl_list;
+		svrattrl *svrattrl_list = NULL;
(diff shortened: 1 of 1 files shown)


bhroam (approved):
LGTM

agrawalravi90 (approved):

PR #2422, opened 2021-06-11
Fixing scheduler hang

nithinj:
#### Describe Bug or Feature
When job spans and consumes resources belonging to another server. The server informs the peer server about the resource which needed to be accounted for. The source server sends this asynchronously and moves on by keeping track of unacknowledged requests. Servers cannot deliver a consistent state unless all the servers are in a ready state without having any pending acks.

Scheduler asks whether servers are ready for querying before stating the universe. This will be a blocking call (for now) and the server will keep the request in a deferred task until it becomes ready. In some cases, the server was not triggering the deferred task call even when the server is ready.

#### Describe Your Change
Trigger the deferred task to reply to the scheduler when we reset the counts in the server

[Test.log](https://github.com/openpbs/openpbs/files/6641127/Test.log)

Profile: regression
|--|--|--|--|--|--|--|
|7851|4177|0|1|0|0|4176|


nithinj:
Requesting reviews from @agrawalravi90

PR #2424, opened 2021-06-16
Addressed static code analysis issues

bhroam:
#### Describe Bug or Feature
I ran cppcheck and clang's static code analyzer on the scheduler and addressed the findings.
[report.txt](https://github.com/openpbs/openpbs/files/6666160/report.txt)

Here is cppcheck's report.  I'm not attaching clang's because it was filled with false positives.  I only addressed the real issues.  Some of cppcheck's issues were false positives, but not many.

#### Describe Your Change
Addressed issues reported by static code analyzers

A lot of minor fixes.  I moved a lot of variables closer to their use.  There were 3 actual bugs:
1) If we compiled with DEBUG mode, we wouldn't chdir() into our priv dir
2) There was a path through the STF code where if a job needed to shrink close to its min_walltime, it could cause a FPE.
3) for the daemon service user, we called setuid() and setgid() without checking their return value.  It turned out that setgid() was failing.  I believe this is because we set the uid first, and then we no longer have the privilege to set the gid.  I now set the gid first. 

Profile: regression
|--|--|--|--|--|--|--|
|7916|4177|2|5|0|0|4170|

I reran the failed/error tests and they passed.

### DIFF
--- a/src/scheduler/resource_resv.cpp
+++ b/src/scheduler/resource_resv.cpp
@@ -303,7 +303,7 @@ free_resource_resv_array(resource_resv **resresv_arr)
 }
 
 /**
- * @brie& 
+ * @brief
  *		resource_resv destructor
  *
  */
@@ -2037,7 +2037,7 @@ new_chunk()
  * @return	duplicate chunk array.
  */
 chunk **
-dup_chunk_array(chunk **old_chunk_arr)
+dup_chunk_array(const chunk * const *old_chunk_arr)
 {
 	int i;
 	int ct;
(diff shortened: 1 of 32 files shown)


src/scheduler/resource_resv.cpp
```diff
@@ -305,7 +303,7 @@ free_resource_resv_array(resource_resv **resresv_arr)
 }
 
 /**
- * @brief
+ * @brie&
```
vstumpf, line 306: Is this intended?

src/scheduler/resource_resv.cpp
```diff
@@ -2258,6 +2196,15 @@ selspec::selspec(selspec& oldspec)
 	defs = oldspec.defs;
 }
 
+selspec& selspec::operator=(selspec& oldspec)
```
vstumpf, line 2199: this should be a const ref

vstumpf (changes requested):
Some comments

arungrover (approved):
LGTM!

vstumpf (approved):
LGTM

PR #2426, opened 2021-06-22
Fix for the test failure to validate a long string set to customer resource

bremanandjk:
#### Describe Bug or Feature

The test is setting a string custom resource for 4096 long value and validating this long resource value in accounting logs, server logs, and mail. Seems the sendmail has a limitation that it can't have a single line with more than 2048 characters, this truncating it. 
(For reference: https://www.cs.ait.ac.th/~on/O/oreilly/tcpip/sendmail/ch18_08.htm)
<html>
<body>

MAXLINE | 2048 | conf.h | Length of an input line
-- | -- | -- | --

</body>
</html>

#### Describe Your Change

It's not an expected behavior to see 4096 bytes long string in the mail, thus removing this invalid validation. 

[testlog.txt](https://github.com/openpbs/openpbs/files/6693650/testlog.txt)

### DIFF
--- a/test/tests/functional/pbs_acct_log.py
+++ b/test/tests/functional/pbs_acct_log.py
@@ -57,20 +57,6 @@ class TestAcctLog(TestFunctional):
         in the job's resources_used attr or the accounting log at job end
         """
 
-        # Make sure emails are not truncated
-        try:
-            mailfile = os.environ['MAIL']
-        except KeyError:
-            self.skip_test(
-                "mail is not setup. " +
-                "Hence this step would be skipped. " +
-                "Please setup the mail.")
-        if not os.path.isfile(mailfile):
-            self.skip_test(
-                "Mail file does not exist. " +
-                "Hence this step would be skipped. " +
-                "Please check manually.")
-
         self.server.manager(MGR_CMD_SET, SERVER,
                             {'job_history_enable': 'True'})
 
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_acct_log.py
```diff
@@ -107,18 +93,6 @@ def test_long_resource_end(self):
         log_match = 'resources_used.foo_str=' + hstr
         self.server.log_match("%s;.*%s.*" % (jid, log_match), regexp=True)
 
-        mailpass = 0
```
bhroam, line 110: If you are not going to look in the mail any more, you don't need to set ATTR_m to 'e' above.
Also, if you want to, you could change the for loop which creates hstr to hstr = '1' * 4096.  It'd probably be faster.

bhroam (changes requested):

test/tests/functional/pbs_acct_log.py
bremanandjk: Thanks, @bhroam for your review comments,
I have addressed your comments, please review further.

bhroam (approved):
I think PEP8 will want spaces around the '*', but if it doesn't complain, neither will I.

LGTM

bayucan (approved):
Looks fine.

PR #2428, opened 2021-06-22
DBPRT to log file

toonen:
At times, it is useful to have the debugging printf (DBPRT) messages sent to the log files rather than stdout so that they are interspersed with the other log messages.

This pull request defines an alternative implementation of the DBPRT macro when DBPRT_LOG is defined.  This implementation uses the logging module to log the message at the PBSEVENT_DEBUG3 level.  A few fixes to code using DBPRT are also included.

### DIFF
--- a/src/include/log.h
+++ b/src/include/log.h
@@ -92,7 +92,7 @@ extern "C" {
 #define STRIP_PARENS(...) __VA_ARGS__
 #undef DBPRT
 #define	DBPRT(x) \
-	if (will_log_event(PBSEVENT_DEBUG3)) { \
+	if (will_log_event(PBSEVENT_DEBUGPRT)) { \
 		char * msg_; \
 		int msg_len_; \
 		int save_errno_ = errno; \
@@ -101,7 +101,7 @@ extern "C" {
 			if (msg_len_ > 0 && msg_[msg_len_ - 1] == '\n') { \
 				msg_[msg_len_ - 1] = '\0'; \
 			} \
-			log_record(PBSEVENT_DEBUG3, 0, LOG_DEBUG, __func__, msg_); \
+			log_record(PBSEVENT_DEBUGPRT, 0, LOG_DEBUG, __func__, msg_); \
 			free(msg_); \
 		} \
 		errno = save_errno_; \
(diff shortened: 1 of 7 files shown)


src/include/log.h
```diff
@@ -84,8 +84,31 @@ extern "C" {
...
+#include <errno.h>
+#include "libutil.h"
+#define STRIP_PARENS(...) __VA_ARGS__
+#undef DBPRT
+#define	DBPRT(x) \
+	if (will_log_event(PBSEVENT_DEBUG3)) { \
+		char * msg_; \
+		int msg_len_; \
+		int save_errno_ = errno; \
+		msg_len_ = pbs_asprintf(&msg_, STRIP_PARENS x); \
```
bhroam, line 99: Can you use log_eventf() here?  It's like log_event() and printf() combined into one.

src/include/log.h
```diff
@@ -84,8 +84,31 @@ extern "C" {
...
-#define	DBPRT(x)
+#endif
+#ifdef DBPRT_LOG
+#include <stdlib.h>
+#include <errno.h>
+#include "libutil.h"
+#define STRIP_PARENS(...) __VA_ARGS__
+#undef DBPRT
+#define	DBPRT(x) \
+	if (will_log_event(PBSEVENT_DEBUG3)) { \
```
bhroam, line 95: Maybe instead of using DEBUG3, we create a new DEBUG level called PBSEVENT_DEBUGPRT or something like that.  This way it won't be interspersed with the normal DEBUG3 messages.

bhroam (changes requested):

src/include/log.h
toonen: I used asprintf() to construct the message so that I could strip off the trailing newline, which was resulting in a considerable number of blank lines in the log files.  I might be able to add another macro layer to strip out the format string, but I would still need to copy it into another buffer in order to remove the newline.   While the asprintf() results in extra memory allocation, DBPRT only produces the extra code and memory overhead when DBPRT_LOG is defined, and I was only expecting it to be defined when `-O0 -g` had already been added and gdb wasn't helping.  In other words, when debugging frustration levels had reached critical.  :-)

bhroam (approved):
LGTM

src/include/log.h
toonen: Done in 43931b90596f6c47652a84f1981a4de00c305ff7.  This change does allow the level value to be overridden by compiler flags or a temporary addition to source and/or header files in order to make developer's desire of "I want the output now!" while debugging as simple as possible and yet minimize the impact on others.

toonen:
@bhroam Thanks for the feedback!  As always, it's greatly appreciated.

bayucan (approved):
Looks good.

PR #2430, opened 2021-06-22
Fixed possible memory leak when querying resources

bhroam:
#### Describe Bug or Feature
Possible memory leak if some built-in resources aren't sent to the scheduler.

#### Describe Your Change
Moved pbs_statfree() to the point when we are done with it.

Also, new can't return NULL.  No reason to check for it.

Profile: regression
|--|--|--|--|--|--|--|
|7966|4177|1|3|0|0|4173|

I reran the failed/error tests.  All but 1 failed which is a known test bug in master.

### DIFF
--- a/src/scheduler/resource.cpp
+++ b/src/scheduler/resource.cpp
@@ -115,11 +115,10 @@ query_resources(int pbs_sd)
 	for (cur_bs = bs; cur_bs != NULL; cur_bs = cur_bs->next) {
 		int flags = NO_FLAGS;
 		resource_type rtype;
-		char *endp;
-
-		attrp = cur_bs->attribs;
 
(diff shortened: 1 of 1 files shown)


src/scheduler/resource.cpp
```diff
@@ -116,27 +116,20 @@ query_resources(int pbs_sd)
 		int flags = NO_FLAGS;
 		resource_type rtype;
 		char *endp;
-		resdef *def;
 
 		attrp = cur_bs->attribs;
 
-		while (attrp != NULL) {
+		for (attrp = cur_bs->attribs; attrp != NULL; attrp = attrp->next) {
```
agrawalravi90, line 122: you can remove line 120

agrawalravi90 (changes requested):
Minor comment

agrawalravi90 (approved):

PR #2432, opened 2021-06-25
Refactored query code

bhroam:
#### Describe Bug or Feature
When querying the universe, querying an object type is broken up into two functions.  One to query the container of objects (usually an array), and the other to query a single object itself.  An example of this is query_jobs() queries an array of jobs in which is loops over what pbs_selstat() gave us and calls query_job() on each one.

Right now query_job() really only converts the batch_status into a job.  The rest of querying the job happens in query_jobs() itself.  This is similar for reservations, queues, etc.

#### Describe Your Change
This PR is about refactoring the code so everything pertaining to the querying of an object is in the function that queries the object.  This means the part of querying a job that was in query_jobs() has been moved into query_job().  Really all query_jobs() does now is stat the jobs and loop over and call query_job() on each.

Profile: regression
|--|--|--|--|--|--|--|
|7981|4177|0|3|0|0|4174|

I reran the failed tests and they passed.

### DIFF
--- a/src/scheduler/server_info.cpp
+++ b/src/scheduler/server_info.cpp
@@ -160,7 +160,7 @@ extern char **environ;
  *		and all the queues and jobs that reside in that server
  *
  * @par Order of Query
- *		query_universe()
+ *		query_server()
  *      -> query_sched()
  *	 	-> query_nodes()
  *	 	-> query_queues()
@@ -176,7 +176,7 @@ extern char **environ;
  *
  */
 server_info *
-query_universe(status *pol, int pbs_sd)
+query_server(status *pol, int pbs_sd)
 {
 	struct batch_status *server;	/* info about the server */
 	struct batch_status *bs_resvs;	/* batch status of the reservations */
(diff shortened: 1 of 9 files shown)


src/scheduler/server_info.cpp
```diff
@@ -177,12 +176,11 @@ extern char **environ;
  *
  */
 server_info *
-query_server(status *pol, int pbs_sd)
+query_universe(status *pol, int pbs_sd)
```
arungrover, line 179: I was thinking, should we really rename it to query_universe? If we move this method to the server_info class then it should ideally be called query_server. Another reason was that we already have functions like update_universe_on_run that call update_server_on_run so query_universe may seem confusing to people.

src/scheduler/server_info.cpp
```diff
@@ -431,6 +423,7 @@ query_server(status *pol, int pbs_sd)
 		return NULL;
 	}
 
+	/* Clean up some node tasks which we couldn't do in query_nodes() because we didn't have everything we needed */
```
arungrover, line 426: I am a little confused here, what are we trying to clean here?

arungrover (changes requested):
Minor comments

src/scheduler/server_info.cpp
bhroam: I'll update the comment.  I was just trying to be terse.  I was a little too terse.  Ideally we'd do everything to query the nodes in query_nodes().  Unfortunately we don't have all the data we need to do that.  Some of it is queried/created after the query_nodes() call.  It had to be delayed until near the end of query_server().

arungrover (approved):
LGTM!

sakshamgarg (approved):

PR #2434, opened 2021-06-29
Convert counts, np_cache, node_group_key to C++ containers

arungrover:
#### Describe Bug or Feature
As part of converting C code to C++ containers, this PR addresses some of server_info members and converts them into unordered_map and vectors.

Each of these changes is an independent commit for ease of review.

#### Link to Design Doc
NA

Description: Rerun All Tests From #8151
Profile: regression
|--|--|--|--|--|--|--|
|8170|4177|2|0|0|1|4174|
Some of the tests are failing, I'm looking into those to see if they are failing because of this change.

### DIFF
--- a/src/include/limits_if.h
+++ b/src/include/limits_if.h
@@ -240,7 +240,7 @@ void update_soft_limits(server_info *, queue_info *, resource_resv *);
  *		the counts structure
  * @return	int
  */
-int find_preempt_bits(counts *, const char *, resource_resv *);
+int find_preempt_bits(counts *, std::string &, resource_resv *);
 #ifdef	__cplusplus
 }
 #endif
--- a/src/scheduler/buckets.cpp
+++ b/src/scheduler/buckets.cpp
@@ -393,7 +393,7 @@ char *create_node_bucket_name(status *policy, node_bucket *nb) {
  * @retval NULL on error
  */
 node_bucket **
-create_node_buckets(status *policy, node_info **nodes, queue_info **queues, unsigned int flags) {
+create_node_buckets(status *policy, node_info **nodes, qinfo_vector &queues, unsigned int flags) {
 	int i;
 	int j = 0;
 	node_bucket **buckets = NULL;
@@ -421,7 +421,7 @@ create_node_buckets(status *policy, node_info **nodes, queue_info **queues, unsi
 		if (nodes[i]->is_down || nodes[i]->is_offline || node_ind == -1 || nodes[i]->lic_lock == 0)
 			continue;
 
-		if (queues != NULL && !nodes[i]->queue_name.empty())
+		if (!queues.empty() && !nodes[i]->queue_name.empty())
 			qinfo = find_queue_info(queues, nodes[i]->queue_name);
 
 		bkt_ind = find_node_bucket_ind(buckets, nodes[i]->res, qinfo, nodes[i]->priority);
--- a/src/scheduler/buckets.cpp
+++ b/src/scheduler/buckets.cpp
@@ -1077,7 +1077,7 @@ check_node_buckets(status *policy, server_info *sinfo, queue_info *qinfo, resour
 	 * If it doesn't exist, we'll create it and add it to the cache
 	 */
 	if (resresv->place_spec->group != NULL) {
-		char *grouparr[2] = {0};
+		string_vector groupvec{resresv->place_spec->group};
 		np_cache *npc = NULL;
 		node_info **ninfo_arr;
 
@@ -1086,9 +1086,7 @@ check_node_buckets(status *policy, server_info *sinfo, queue_info *qinfo, resour
 		else
 			ninfo_arr = sinfo->unassoc_nodes;
 
-		grouparr[0] = resresv->place_spec->group;
-		grouparr[1] = NULL;
-		npc = find_alloc_np_cache(policy, &(sinfo->npc_arr), grouparr, ninfo_arr, cmp_placement_sets);
+		npc = find_alloc_np_cache(policy, &(sinfo->npc_arr), groupvec, ninfo_arr, cmp_placement_sets);
 		if (npc != NULL)
 			nodepart = npc->nodepart;
 	}
--- a/src/scheduler/buckets.cpp
+++ b/src/scheduler/buckets.cpp
@@ -1086,7 +1086,7 @@ check_node_buckets(status *policy, server_info *sinfo, queue_info *qinfo, resour
 		else
 			ninfo_arr = sinfo->unassoc_nodes;
 
-		npc = find_alloc_np_cache(policy, &(sinfo->npc_arr), groupvec, ninfo_arr, cmp_placement_sets);
+		npc = find_alloc_np_cache(policy, sinfo->npc_arr, groupvec, ninfo_arr, cmp_placement_sets);
 		if (npc != NULL)
 			nodepart = npc->nodepart;
 	}
--- a/src/scheduler/buckets.cpp
+++ b/src/scheduler/buckets.cpp
@@ -400,7 +400,7 @@ create_node_buckets(status *policy, node_info **nodes, std::vector<queue_info *>
 	node_bucket **tmp;
 	int node_ct;
 
-	if (policy == NULL || nodes == NULL)
+	if (policy == NULL || nodes == NULL || queues.empty())
 		return NULL;
 
 	node_ct = count_array(nodes);
@@ -421,7 +421,7 @@ create_node_buckets(status *policy, node_info **nodes, std::vector<queue_info *>
 		if (nodes[i]->is_down || nodes[i]->is_offline || node_ind == -1 || nodes[i]->lic_lock == 0)
 			continue;
 
-		if (!queues.empty() && !nodes[i]->queue_name.empty())
+		if (!nodes[i]->queue_name.empty())
 			qinfo = find_queue_info(queues, nodes[i]->queue_name);
 
 		bkt_ind = find_node_bucket_ind(buckets, nodes[i]->res, qinfo, nodes[i]->priority);
--- a/src/scheduler/buckets.cpp
+++ b/src/scheduler/buckets.cpp
@@ -393,7 +393,7 @@ char *create_node_bucket_name(status *policy, node_bucket *nb) {
  * @retval NULL on error
  */
 node_bucket **
-create_node_buckets(status *policy, node_info **nodes, qinfo_vector &queues, unsigned int flags) {
+create_node_buckets(status *policy, node_info **nodes, std::vector<queue_info *> &queues, unsigned int flags) {
 	int i;
 	int j = 0;
 	node_bucket **buckets = NULL;
@@ -1077,7 +1077,7 @@ check_node_buckets(status *policy, server_info *sinfo, queue_info *qinfo, resour
 	 * If it doesn't exist, we'll create it and add it to the cache
 	 */
 	if (resresv->place_spec->group != NULL) {
-		string_vector groupvec{resresv->place_spec->group};
+		std::vector<std::string> groupvec{resresv->place_spec->group};
 		np_cache *npc = NULL;
 		node_info **ninfo_arr;
 
--- a/src/scheduler/constant.h
+++ b/src/scheduler/constant.h
@@ -41,6 +41,7 @@
 #define	_CONSTANT_H
 
 #include <math.h>
+#include <string>
 
 /* macro to turn a value from enum preempt into it's bit for the bitfield */
 #define PREEMPT_TO_BIT(X) (1 << (X) )
--- a/src/scheduler/constant.h
+++ b/src/scheduler/constant.h
@@ -41,7 +41,6 @@
 #define	_CONSTANT_H
 
 #include <math.h>
-#include <string>
 
 /* macro to turn a value from enum preempt into it's bit for the bitfield */
 #define PREEMPT_TO_BIT(X) (1 << (X) )
--- a/src/scheduler/data_types.h
+++ b/src/scheduler/data_types.h
@@ -431,19 +431,19 @@ struct server_info
 	 */
 	int preempt_count[NUM_PPRIO + 1];
 
-	counts *group_counts;		/* group resource and running counts */
-	counts *project_counts;		/* project resource and running counts */
-	counts *user_counts;		/* user resource and running counts */
-	counts *alljobcounts;		/* overall resource and running counts */
+	counts_map group_counts;		/* group resource and running counts */
+	counts_map project_counts;		/* project resource and running counts */
+	counts_map user_counts;		/* user resource and running counts */
+	counts_map alljobcounts;		/* overall resource and running counts */
 
 	/*
 	 * Resource/Run counts list to store counts for all jobs which
 	 * are running/queued/suspended.
 	 */
-	counts *total_group_counts;
-	counts *total_project_counts;
-	counts *total_user_counts;
-	counts *total_alljobcounts;
+	counts_map total_group_counts;
+	counts_map total_project_counts;
+	counts_map total_user_counts;
+	counts_map total_alljobcounts;
 
 	node_partition **nodepart;	/* array pointers to node partitions */
 	int num_parts;			/* number of node partitions(node_group_key) */
@@ -516,18 +516,18 @@ class queue_info
 	resource_resv **jobs;		/* array of jobs that reside in queue */
 	resource_resv **running_jobs;	/* array of jobs in the running state */
 	node_info **nodes;		/* array of nodes associated with the queue */
-	counts *group_counts;		/* group resource and running counts */
-	counts *project_counts;		/* project resource and running counts */
-	counts *user_counts;		/* user resource and running counts */
-	counts *alljobcounts;		/* overall resource and running counts */
+	counts_map group_counts;		/* group resource and running counts */
+	counts_map project_counts;		/* project resource and running counts */
+	counts_map user_counts;		/* user resource and running counts */
+	counts_map alljobcounts;		/* overall resource and running counts */
 	/*
 	 * Resource/Run counts list to store counts for all jobs which
 	 * are running/queued/suspended.
 	 */
-	counts *total_group_counts;
-	counts *total_project_counts;
-	counts *total_user_counts;
-	counts *total_alljobcounts;
+	counts_map total_group_counts;
+	counts_map total_project_counts;
+	counts_map total_user_counts;
+	counts_map total_alljobcounts;
 
 	char **node_group_key;		/* node grouping resources */
 	struct node_partition **nodepart; /* array pointers to node partitions */
@@ -690,8 +690,8 @@ struct node_info
 
 	int priority;			/* node priority */
 
-	counts *group_counts;		/* group resource and running counts */
-	counts *user_counts;		/* user resource and running counts */
+	counts_map group_counts;	/* group resource and running counts */
+	counts_map user_counts;		/* user resource and running counts */
 
 	int max_running;		/* max number of jobs on the node */
 	int max_user_run;		/* max number of jobs running by a user */
@@ -893,15 +893,13 @@ class prev_job_info
 class counts
 {
 	public:
-	char *name;			/* name of entitiy */
+	std::string name;		/* name of entitiy */
 	int running;			/* count of running jobs in object */
 	int soft_limit_preempt_bit;	/* Place to store preempt bit if entity is over limits */
 	resource_count *rescts;		/* resources used */
-	counts *next;
 	counts(const std::string &);
-	counts(const char *);
 	counts(const counts &);
-	counts& operator=(const counts&);
+	counts& operator=(const counts &);
 	~counts();
 };
 
--- a/src/scheduler/data_types.h
+++ b/src/scheduler/data_types.h
@@ -139,10 +139,7 @@ typedef struct th_data_dup_resresv th_data_dup_resresv;
 typedef struct th_data_query_jinfo th_data_query_jinfo;
 typedef struct th_data_free_resresv th_data_free_resresv;
 
-typedef std::unordered_map<std::string, counts *> counts_map;
-typedef std::vector<std::string> string_vector;
-typedef std::vector<np_cache *> np_cache_vector;
-typedef std::vector<queue_info *> qinfo_vector;
+using counts_umap = std::unordered_map<std::string, counts *>;
 #ifdef NAS
 /* localmod 034 */
 /*
@@ -413,9 +410,9 @@ class server_info
 	int num_nodes;			/* number of nodes associated with the server */
 	int num_resvs;			/* number of reservations on the server */
 	int num_preempted;		/* number of jobs currently preempted */
-	string_vector node_group_key;		/* the node grouping resources */
+	std::vector<std::string> node_group_key;		/* the node grouping resources */
 	state_count sc;			/* number of jobs in each state */
-	qinfo_vector queues;		/* array of queues */
+	std::vector<queue_info *> queues;		/* array of queues */
 	queue_info ***queue_list;	/* 3 dimensional array, used to order jobs in round_robin */
 	node_info **nodes;		/* array of nodes associated with the server */
 	node_info **unassoc_nodes;	/* array of nodes not associated with queues */
@@ -435,19 +432,19 @@ class server_info
 	 */
 	int preempt_count[NUM_PPRIO + 1];
 
-	counts_map group_counts;		/* group resource and running counts */
-	counts_map project_counts;		/* project resource and running counts */
-	counts_map user_counts;		/* user resource and running counts */
-	counts_map alljobcounts;		/* overall resource and running counts */
+	counts_umap group_counts;		/* group resource and running counts */
+	counts_umap project_counts;		/* project resource and running counts */
+	counts_umap user_counts;		/* user resource and running counts */
+	counts_umap alljobcounts;		/* overall resource and running counts */
 
 	/*
 	 * Resource/Run counts list to store counts for all jobs which
 	 * are running/queued/suspended.
 	 */
-	counts_map total_group_counts;
-	counts_map total_project_counts;
-	counts_map total_user_counts;
-	counts_map total_alljobcounts;
+	counts_umap total_group_counts;
+	counts_umap total_project_counts;
+	counts_umap total_user_counts;
+	counts_umap total_alljobcounts;
 
 	node_partition **nodepart;	/* array pointers to node partitions */
 	int num_parts;			/* number of node partitions(node_group_key) */
@@ -462,7 +459,7 @@ class server_info
 	 * be duplicated.  It would be difficult to duplicate correctly, and it is
 	 * just a cache.  It will be regenerated when needed
 	 */
-	np_cache_vector npc_arr;
+	std::vector<np_cache *> npc_arr;
 
 	resource_resv *qrun_job;	/* used if running a job via qrun request */
 	/* policy structure for the server.  This is an easy storage location for
@@ -484,7 +481,7 @@ class server_info
 	server_info();
 	server_info(int);
 	server_info(const server_info &);
-	~server_info();
+	virtual ~server_info();
 	server_info & operator=(const server_info &);
 	void dup_server_psets(const std::unordered_map<std::string, node_partition*>& spsets);
 };
@@ -529,20 +526,20 @@ class queue_info
 	resource_resv **jobs;		/* array of jobs that reside in queue */
 	resource_resv **running_jobs;	/* array of jobs in the running state */
 	node_info **nodes;		/* array of nodes associated with the queue */
-	counts_map group_counts;		/* group resource and running counts */
-	counts_map project_counts;		/* project resource and running counts */
-	counts_map user_counts;		/* user resource and running counts */
-	counts_map alljobcounts;		/* overall resource and running counts */
+	counts_umap group_counts;		/* group resource and running counts */
+	counts_umap project_counts;		/* project resource and running counts */
+	counts_umap user_counts;		/* user resource and running counts */
+	counts_umap alljobcounts;		/* overall resource and running counts */
 	/*
 	 * Resource/Run counts list to store counts for all jobs which
 	 * are running/queued/suspended.
 	 */
-	counts_map total_group_counts;
-	counts_map total_project_counts;
-	counts_map total_user_counts;
-	counts_map total_alljobcounts;
+	counts_umap total_group_counts;
+	counts_umap total_project_counts;
+	counts_umap total_user_counts;
+	counts_umap total_alljobcounts;
 
-	string_vector node_group_key;		/* node grouping resources */
+	std::vector<std::string> node_group_key;		/* node grouping resources */
 	struct node_partition **nodepart; /* array pointers to node partitions */
 	struct node_partition *allpart;   /* partition w/ all nodes assoc with queue*/
 	int num_parts;			/* number of node partitions(node_group_key) */
@@ -703,8 +700,8 @@ struct node_info
 
 	int priority;			/* node priority */
 
-	counts_map group_counts;	/* group resource and running counts */
-	counts_map user_counts;		/* user resource and running counts */
+	counts_umap group_counts;	/* group resource and running counts */
+	counts_umap user_counts;		/* user resource and running counts */
 
 	int max_running;		/* max number of jobs on the node */
 	int max_user_run;		/* max number of jobs running by a user */
@@ -787,9 +784,9 @@ class resource_resv
 	bool will_use_multinode:1;	/* res resv will use multiple nodes */
 
 	const std::string name;		/* name of res resv */
-	std::string user;			/* username of the owner of the res resv */
-	std::string group;			/* exec group of owner of res resv */
-	std::string project;			/* exec project of owner of res resv */
+	std::string user;		/* username of the owner of the res resv */
+	std::string group;		/* exec group of owner of res resv */
+	std::string project;		/* exec project of owner of res resv */
 	char *nodepart_name;		/* name of node partition to run res resv in */
 
 	long sch_priority;		/* scheduler priority of res resv */
@@ -797,7 +794,7 @@ class resource_resv
 	int ec_index;			/* Index into server's job_set array*/
 
 	time_t qtime;			/* time res resv was submitted */
-	long long qrank;			/* time on which we might need to stabilize the sort */
+	long long qrank;		/* time on which we might need to stabilize the sort */
 	time_t start;			/* start time (UNDEFINED means no start time */
 	time_t end;			/* end time (UNDEFINED means no end time */
 	time_t duration;		/* duration of resource resv request */
@@ -1005,7 +1002,7 @@ struct node_partition
 class np_cache
 {
 	public:
-	string_vector resnames;		/* resource names used to create partitions */
+	std::vector<std::string> resnames;		/* resource names used to create partitions */
 	node_info **ninfo_arr;		/* ptr to array of nodes used to create pools */
 	int num_parts;			/* number of partitions in nodepart */
 	node_partition **nodepart;	/* node partitions */
--- a/src/scheduler/data_types.h
+++ b/src/scheduler/data_types.h
@@ -139,6 +139,7 @@ typedef struct th_data_query_jinfo th_data_query_jinfo;
 typedef struct th_data_free_resresv th_data_free_resresv;
 
 typedef std::unordered_map<std::string, counts *> counts_map;
+typedef std::vector<std::string> string_vector;
 #ifdef NAS
 /* localmod 034 */
 /*
@@ -409,7 +410,7 @@ struct server_info
 	int num_nodes;			/* number of nodes associated with the server */
 	int num_resvs;			/* number of reservations on the server */
 	int num_preempted;		/* number of jobs currently preempted */
-	char **node_group_key;		/* the node grouping resources */
+	string_vector node_group_key;		/* the node grouping resources */
 	state_count sc;			/* number of jobs in each state */
 	queue_info **queues;		/* array of queues */
 	queue_info ***queue_list;	/* 3 dimensional array, used to order jobs in round_robin */
@@ -529,7 +530,7 @@ class queue_info
 	counts_map total_user_counts;
 	counts_map total_alljobcounts;
 
-	char **node_group_key;		/* node grouping resources */
+	string_vector node_group_key;		/* node grouping resources */
 	struct node_partition **nodepart; /* array pointers to node partitions */
 	struct node_partition *allpart;   /* partition w/ all nodes assoc with queue*/
 	int num_parts;			/* number of node partitions(node_group_key) */
@@ -977,13 +978,13 @@ struct node_partition
 {
 	bool ok_break:1;	/* OK to break up chunks on this node part */
 	bool excl:1;		/* partition should be allocated exclusively */
-	char *name;			/* res_name=res_val */
+	char *name;		/* res_name=res_val */
 	/* name of resource and value which define the node partition */
 	resdef *def;
 	char *res_val;
 	int tot_nodes;		/* the total number of nodes  */
 	int free_nodes;		/* the number of nodes in state Free  */
-	schd_resource *res;		/* total amount of resources in node part */
+	schd_resource *res;	/* total amount of resources in node part */
 	node_info **ninfo_arr;	/* array of pointers to node structures  */
 	node_bucket **bkts;	/* node buckets for node part */
 	int rank;		/* unique numeric identifier for node partition */
@@ -991,9 +992,9 @@ struct node_partition
 
 struct np_cache
 {
-	char **resnames;		/* resource names used to create partitions */
-	node_info **ninfo_arr;	/* ptr to array of nodes used to create pools */
-	int num_parts;		/* number of partitions in nodepart */
+	string_vector resnames;		/* resource names used to create partitions */
+	node_info **ninfo_arr;		/* ptr to array of nodes used to create pools */
+	int num_parts;			/* number of partitions in nodepart */
 	node_partition **nodepart;	/* node partitions */
 };
 
--- a/src/scheduler/data_types.h
+++ b/src/scheduler/data_types.h
@@ -383,10 +383,6 @@ struct schedattrs
 
 class server_info
 {
-	void init_server_info();
-	void free_server_info();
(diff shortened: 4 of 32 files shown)


bhroam (changes requested):

bhroam:
I mean to say I gave comments on the first commit.  I'll wait for the other 3 until you are ready.

src/scheduler/limits.cpp
```diff
@@ -774,34 +782,12 @@ free_limcounts(limcounts *lc)
  * @retval	NULL	: on error
  */
 static limcounts *
-make_limcounts(counts *user, counts *group, counts *project, counts *all)
+make_limcounts(counts_map &user, counts_map &group, counts_map &project, counts_map &all)
 {
-	limcounts *lc;
-	lc = new_limcounts();
+	limcounts *lc = NULL;
+	lc = new limcounts(user, group, project, all);
```
bhroam, line 788: New won't return NULL.  It'll throw an exception.

src/scheduler/limits.cpp
```diff
@@ -912,43 +898,34 @@ check_limits(server_info *si, queue_info *qi, resource_resv *rr, schd_error *err
...
 								break;
 							}
 
-							update_counts_on_run(svr_counts->all, te_rr->resreq);
-							svr_counts_max->all =
-								counts_max(svr_counts_max->all, svr_counts->all);
-							if (svr_counts_max->all == NULL) {
-								error = true;
+							cts = find_alloc_counts(svr_counts->all, std::string(PBS_ALL_ENTITY));
+							update_counts_on_run(cts, te_rr->resreq);
```
bhroam, line 925: You can pass the char * PBS_ALL_ENTITY in.  C++ will do the conversion.  The same thing will happen though.  Here you are just being more explicit. (here and other places)

src/scheduler/limits.cpp
```diff
@@ -1158,13 +1115,13 @@ void update_soft_limits(server_info *si, queue_info *qi, resource_resv *rr)
...
  */
-int find_preempt_bits(counts *entity_counts, const char *entity_name, resource_resv *rr)
+int find_preempt_bits(counts_map &entity_counts, const char *entity_name, resource_resv *rr)
 {
 	counts *cnt = NULL;
 	resource_count *res_c;
 	int rc = 0;
 
-	if (entity_counts == NULL || entity_name == NULL)
+	if (entity_counts.size() == 0 || entity_name == empty_str)
```
bhroam, line 1124: entity_name is a char *.  I don't think you meant this.  You'd check it against NULL.  In any case, you don't need this overload.

src/scheduler/limits.cpp
```diff
@@ -1260,15 +1217,14 @@ check_server_max_user_run(server_info *si, queue_info *qi, resource_resv *rr,
...
-	counts		*cts = NULL;
 
 	if ((si == NULL) || (user == empty_str) || (sc == NULL))
 		return (SCHD_ERROR);
 
 	if (!si->has_user_limit)
 	    return (0);
 
-	cts = sc->user;
+	counts_map &cts = sc->user;
```
bhroam, line 1227: use auto (here and below)

src/scheduler/limits.cpp
```diff
@@ -1768,17 +1717,16 @@ check_queue_max_res(server_info *si, queue_info *qi, resource_resv *rr,
 	schd_resource	*res;
 	resource_count	*used_res;
 	counts		*c;
-	counts		*cts = NULL;
 
 	if ((qi == NULL) || (rr == NULL))
 		return (SCHD_ERROR);
 
-	if ((qc == NULL) || (qc->all == NULL))
+	if ((qc == NULL) || (qc->all.size() == 0))
```
bhroam, line 1724: no need to check for size().  It'll just fail to find something and return 0 further down.  The original check against NULL was to guard against crashing.

src/scheduler/server_info.cpp
```diff
@@ -2670,19 +2617,10 @@ free_counts(counts *cts)
...
-		return;
-
-	cur = prev = ctslist;
-
-	while (cur != NULL) {
-		prev = cur->next;
-		free_counts(cur);
-		cur = prev;
+	for (auto it: ctslist) {
+		free_counts(it.second);
```
bhroam, line 2619: No need for free_counts().  Just call delete it.second

src/scheduler/server_info.cpp
```diff
@@ -2901,52 +2783,48 @@ update_counts_on_end(counts *cts, resource_req *resreq)
...
 
-	cmax_head = cmax;
+	if (cmax.size() == 0) {
+		cmax = ncounts;
+		return;
+	}
 
-	for (cur = ncounts; cur != NULL; cur = cur->next) {
-		cur_fmax = find_counts(cmax, cur->name);
+	for (auto cur: ncounts) {
```
bhroam, line 2803: Make this const auto&

src/scheduler/server_info.cpp
```diff
@@ -2958,7 +2836,48 @@ counts_max(counts *cmax, counts *ncounts)
 			}
 		}
 	}
-	return cmax_head;
+}
+
+// overloaded 
+void
+counts_max(counts_map &cmax, counts *ncounts)
```
bhroam, line 2839: There is a lot of code between these two functions, any chance some of it can be moved into a helper function to be called by both overloads?

src/scheduler/server_info.cpp
```diff
@@ -3559,31 +3455,54 @@ void
...
+		for (; i < sinfo->num_queues; i++) {
+			free_counts_list(sinfo->queues[i]->total_group_counts);
+			free_counts_list(sinfo->queues[i]->total_user_counts);
+			free_counts_list(sinfo->queues[i]->total_project_counts);
+			free_counts_list(sinfo->queues[i]->total_alljobcounts);
+			create_total_counts(NULL, sinfo->queues[i], NULL, QUEUE);
+		}
+	}
+	return;
+}
```
bhroam, line 3500: I got rid of this function because it wasn't used.  Are you using it now?

bhroam (changes requested):

src/scheduler/buckets.cpp
```diff
@@ -1077,7 +1077,7 @@ check_node_buckets(status *policy, server_info *sinfo, queue_info *qinfo, resour
 	 * If it doesn't exist, we'll create it and add it to the cache
 	 */
 	if (resresv->place_spec->group != NULL) {
-		char *grouparr[2] = {0};
+		string_vector groupvec{resresv->place_spec->group};
```
bhroam, line 1080: While the counts_map typedef is converting a long type into a shorter type, std::vector<std::string> isn't very long.  Use the explicit type.

src/scheduler/misc.cpp
```diff
@@ -489,7 +491,31 @@ enum match_string_array_ret match_string_array(const char * const *strarr1, cons
...
+		return SA_FULL_MATCH;
+
+	if (match == strarr1.size() || match == strarr2.size())
+		return SA_SUB_MATCH;
+
+	if (match)
+		return SA_PARTIAL_MATCH;
 
+	return SA_NO_MATCH;
+}
```
bhroam, line 518: If we only ever search for full matches, and you are writing this function over again, consider just shrinking both functions to search for a full match.

src/scheduler/node_partition.cpp
```diff
@@ -328,6 +328,12 @@ find_node_partition(node_partition **np_arr, char *name)
 
 	return np_arr[i];
 }
+// overloaded
+node_partition *
+find_node_partition(node_partition **np_arr, const std::string &name)
+{
+	return (find_node_partition(np_arr, name.c_str()));
+}
```
bhroam, line 336: No need for the overload.  Just let C++ do the string conversion.

src/scheduler/node_partition.cpp
```diff
@@ -780,7 +762,7 @@ new_np_cache(void)
 		return NULL;
 	}
 
-	npc->resnames = NULL;
+	npc->resnames = {};
```
bhroam, line 765: This needs to be a constructor if you have C++ types in it.

src/scheduler/node_partition.cpp
```diff
@@ -919,10 +898,10 @@ find_alloc_np_cache(status *policy, np_cache ***pnpc_arr,
 			npc = new_np_cache();
 			if (npc != NULL) {
 				npc->ninfo_arr = ninfo_arr;
-				npc->resnames = dup_string_arr(const_cast<char **>(resnames));
+				npc->resnames = resnames;
 				npc->num_parts = num_parts;
 				npc->nodepart = nodepart;
-				if (npc->resnames == NULL || add_np_cache(pnpc_arr, npc) ==0) {
+				if (npc->resnames.empty() || add_np_cache(pnpc_arr, npc) ==0) {
```
bhroam: I'd remove the npc->resnames.empty().  If the memory failed to be allocated, an exception would have been thrown.  Also, add a space after the '=='

src/scheduler/server_info.cpp
```diff
@@ -441,8 +441,8 @@ query_server(status *pol, int pbs_sd)
 	 * we don't want to account for resources consumed by ghost jobs
 	 */
 	create_placement_sets(policy, sinfo);
-	if (!sinfo->node_group_enable && sinfo->node_group_key != NULL &&
-	    strcmp(sinfo->node_group_key[0], "msvr_node_group") == 0) {
+	if (!sinfo->node_group_enable && !sinfo->node_group_key.empty() &&
+	    (sinfo->node_group_key[0] == std::string("msvr_node_group"))) {
```
bhroam, line 445: No need to convert msvr_node_group into a string.  operator= of string is overloaded to take a char *.

bhroam (changes requested):

src/scheduler/data_types.h
```diff
@@ -140,6 +140,7 @@ typedef struct th_data_free_resresv th_data_free_resresv;
 
 typedef std::unordered_map<std::string, counts *> counts_map;
 typedef std::vector<std::string> string_vector;
+typedef std::vector<np_cache *> np_cache_vector;
```
bhroam, line 143: This isn't much shorter.  Don't do the typedef.

src/scheduler/server_info.cpp
```diff
@@ -1805,9 +1804,9 @@ update_server_on_end(status *policy, server_info *sinfo, queue_info *qinfo,
 	/* We're ending a job or reservation, which will affect the cached data.
 	 * We'll flush the cache and rebuild it if needed
 	 */
-	if (sinfo->npc_arr != NULL) {
+	if (!sinfo->npc_arr.empty()) {
 		free_np_cache_array(sinfo->npc_arr);
-		sinfo->npc_arr = NULL;
+		sinfo->npc_arr = {};
```
bhroam, line 1809: Do sinfo->npc_arr.clear()

bhroam (changes requested):

src/scheduler/data_types.h
```diff
@@ -139,7 +138,7 @@ typedef struct th_data_dup_resresv th_data_dup_resresv;
 typedef struct th_data_query_jinfo th_data_query_jinfo;
 typedef struct th_data_free_resresv th_data_free_resresv;
 
-
+typedef std::unordered_map<std::string, counts *> counts_map;
```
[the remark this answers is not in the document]
arungrover: As we discussed, I'll start using "using" keyword instead.

src/scheduler/constant.h
```diff
@@ -41,6 +41,7 @@
 #define	_CONSTANT_H
 
 #include <math.h>
+#include <string>
```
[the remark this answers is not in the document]
arungrover: yeah,  that's there by mistake.

src/scheduler/data_types.h
```diff
@@ -891,13 +890,19 @@ class prev_job_info
...
+class counts
 {
+	public:
 	char *name;			/* name of entitiy */
 	int running;			/* count of running jobs in object */
 	int soft_limit_preempt_bit;	/* Place to store preempt bit if entity is over limits */
 	resource_count *rescts;		/* resources used */
 	counts *next;
+	counts(const std::string &);
+	counts(const char *);
```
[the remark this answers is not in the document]
arungrover: True, I removed it in later commits

src/scheduler/limits.cpp
```diff
@@ -1501,20 +1509,20 @@ check_queue_max_user_run(server_info *si, queue_info *qi, resource_resv *rr,
...
+	if ((qi == NULL) || (user == empty_str) || (qc == NULL))
 		return (SCHD_ERROR);
 
 	if (!qi->has_user_limit)
 	    return (0);
 
 	cts = qc->user;
 
-	if ((key = entlim_mk_runkey(LIM_USER, user)) == NULL)
+	if ((key = entlim_mk_runkey(LIM_USER, user.c_str())) == NULL)
```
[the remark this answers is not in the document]
arungrover: That's not possible, it is in libutils that's common across daemons.

src/scheduler/server_info.cpp
```diff
@@ -2600,33 +2599,46 @@ is_unassoc_node(node_info *ninfo, void *arg)
...
- * @retval	NULL	: malloc failed
- *
- * @par MT-Safe:	yes
- */
-counts *
-new_counts(void)
-{
+// counts constructor
+counts::counts(const std::string &name_){
+	name = strdup(name_.c_str());
```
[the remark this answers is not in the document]
arungrover: I changed it in later commits

src/scheduler/server_info.cpp
```diff
@@ -2699,17 +2702,7 @@ dup_counts(counts *octs)
...
-	if (ncts != NULL) {
-		if (octs->name != NULL)
-			ncts->name = string_dup(octs->name);
-
-		ncts->running = octs->running;
-		ncts->soft_limit_preempt_bit = octs->soft_limit_preempt_bit;
-
-		ncts->rescts = dup_resource_count_list(octs->rescts);
-	}
+	ncts = new counts(*octs);
```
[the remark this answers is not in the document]
arungrover: I removed the dup function in later commits.

src/scheduler/server_info.cpp
```diff
@@ -2779,6 +2772,12 @@ find_counts(counts *ctslist, const char *name)
 
 	return cur;
 }
+// Overloaded function
+counts *
+find_counts(counts *ctslist, std::string &name)
+{
+	return find_counts(ctslist, name.c_str());
+}
```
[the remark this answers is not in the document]
arungrover: Changed this in later commits

src/scheduler/limits.cpp
arungrover: oops :)

src/scheduler/sched_exception.cpp
```diff
@@ -0,0 +1,90 @@
...
+// Assignment Operator
+sched_exception &sched_exception::operator =(const sched_exception &e)
+{
+	message = e.get_message();
+	error_code = e.get_error_code();
+	return (*this);
+}
+
+// Parametrized Constructor
+sched_exception::sched_exception(const std::string &str, const int &err)
```
bhroam, line 59: No need to pass err in by reference.  It's a scalar.

src/scheduler/sched_exception.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+/*
+ * @brief Overridden function to return char * of message
+ *
+ * @return const char *
+ */
+const char * sched_exception::what()
+{
+    return message.c_str();
+}
```
bhroam, line 90: Attach the * to the function name.

src/scheduler/server_info.cpp
```diff
@@ -56,14 +56,13 @@
...
  * 	free_resource()
- * 	new_server_info()
  * 	new_resource()
  * 	create_resource()
  * 	add_resource_list()
  * 	add_resource_value()
  * 	add_resource_str_arr()
  * 	add_resource_bool()
- * 	free_server()
+ * 	delete )
```
bhroam, line 65: Remove this

src/scheduler/server_info.cpp
```diff
@@ -115,8 +113,10 @@
 #include <ctype.h>
 #include <signal.h>
 #include <sys/wait.h>
-
+// STL headers
```
bhroam, line 114: No need for this.

src/scheduler/server_info.cpp
```diff
@@ -974,104 +973,104 @@ free_server_psets(std::unordered_map<std::string, node_partition *>& spsets)
...
-		newpset[spset.first] = dup_node_partition(spset.second, sinfo);
-		if (newpset[spset.first] == NULL) {
-			free_server_psets(newpset);
-			return {};
-		}
-	}
 
-	return newpset;
+	for (const auto& spset : spsets)
+		svr_to_psets[spset.first] = dup_node_partition(spset.second, this);
```
bhroam, line 985: Can dup_node_partition() no longer return NULL?

src/scheduler/server_info.cpp
```diff
@@ -1132,89 +1131,86 @@ free_resource(schd_resource *resp)
...
+// Parametrized Constructor
+server_info::server_info(int limallocflag)
+{
+	init_server_info();
 	if ((limallocflag != 0))
-		sinfo->liminfo = lim_alloc_liminfo();
-	init_state_count(&(sinfo->sc));
-	memset(sinfo->preempt_count, 0, (NUM_PPRIO + 1) * sizeof(int));
+		liminfo = lim_alloc_liminfo();
+}
```
bhroam, line 1199: This shouldn't be needed.  Now that we have a constructor and copy constructor, one passes 1, and the other passes 0.  Look how I did queue_info.

src/scheduler/server_info.cpp
```diff
@@ -1132,89 +1131,86 @@ free_resource(schd_resource *resp)
...
- * @see	lim_alloc_liminfo
- *
- * @par MT-Safe:	no
- */
-server_info *
-new_server_info(int limallocflag)
-{
-	server_info *sinfo;			/* the new server */
+// Init function
+void server_info::init_server_info ()
```
bhroam, line 1133: This function is only called by the constructors, make it private.

src/scheduler/server_info.cpp
```diff
@@ -2107,156 +2072,131 @@ check_resv_running_on_node(resource_resv *resv, const void *arg)
...
-	nsinfo->node_group_key = osinfo->node_group_key;
-	nsinfo->nodesigs = dup_string_arr(osinfo->nodesigs);
-
-	nsinfo->policy = dup_status(osinfo->policy);
-
-	nsinfo->num_nodes = osinfo->num_nodes;
+// Copy constructor
+server_info::server_info(const server_info &osinfo)
+{
+	init_server_info();
```
bhroam, line 2075: I wouldn't init everything just to copy over everything again.  Just keep in mind that the constructor hasn't been called, so you do actually need to set every member.

src/scheduler/server_info.cpp
```diff
@@ -2107,156 +2072,131 @@ check_resv_running_on_node(resource_resv *resv, const void *arg)
...
 	/* dup the nodes, if there are any nodes */
-	nsinfo->nodes = dup_nodes(osinfo->nodes, nsinfo, NO_FLAGS);
+	nodes = dup_nodes(osinfo.nodes, this, NO_FLAGS);
 
-	if (nsinfo->has_nodes_assoc_queue) {
-		nsinfo->unassoc_nodes =
-			node_filter(nsinfo->nodes, nsinfo->num_nodes, is_unassoc_node, NULL, 0);
+	if (has_nodes_assoc_queue) {
+		unassoc_nodes =
+			node_filter(nodes, num_nodes, is_unassoc_node, NULL, 0);
```
bhroam, line 2120: Combine these to 1 line and ditch the {}

src/scheduler/server_info.cpp
```diff
@@ -2107,156 +2072,131 @@ check_resv_running_on_node(resource_resv *resv, const void *arg)
...
 
 	/* duplicate the queues */
-	nsinfo->queues = dup_queues(osinfo->queues, nsinfo);
-	if (nsinfo->queues.empty()) {
-		free_server(nsinfo);
-		return NULL;
+	queues = dup_queues(osinfo.queues, this);
+	if (queues.empty()) {
+		free_server_info();
+		throw sched_exception("Unable to duplicate queues", SCHD_ERROR);
```
bhroam, line 2138: If the int in sched_exception is a sched error code, then it shouldn't be int.  It should be the enum.

src/scheduler/server_info.cpp
```diff
@@ -2265,72 +2205,70 @@ dup_server_info(server_info *osinfo)
...
-			return NULL;
+	for (int i = 0; osinfo.nodes[i] != NULL; i++)
+		nodes[i]->job_arr =
+			copy_resresv_array(osinfo.nodes[i]->job_arr, jobs);
+
+	num_parts = osinfo.num_parts;
+	if (osinfo.nodepart != NULL) {
+		nodepart = dup_node_partition_array(osinfo.nodepart, this);
+		if (nodepart == NULL) {
+			free_server_info();
```
bhroam, line 2213: I don't think this will work.  Even though we threw an exception, we'll still leak.  When the caller does a foo = new server_info(*osinfo), they'll still get a new, but empty sinfo back.  They'll have to call delete on it.  If they're calling delete on it, then you don't have to free everything here.

bhroam (changes requested):

src/scheduler/server_info.cpp
arungrover: I've changed it to initialize it only in the parameterized constructor in later commits. Do you think we don't need parameterized constructor and should always allocate liminfo?

src/scheduler/limits.cpp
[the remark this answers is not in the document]
bhroam: Depending how often you have to call out user.c_str(), you can make a local overload.  This is what I did for log_eventf(), so it could take a string.

src/scheduler/server_info.cpp
bhroam: So new_server_info() took a parameter to allocate this or not.  When new_server_info() was called in query, it was 1.  When it was called in dup_server_info(), it was 0.  The reason we did that is we didn't want to allocate one in dup just to copy the old one over.  Since these are two different constructors now, you don't need the parameter.

src/scheduler/data_types.h
```diff
@@ -477,13 +478,13 @@ class server_info
...
 	// Class methods
 	void free_server_info();
-	void init_server_info();
 	server_info();
 	server_info(int);
 	server_info(const server_info &);
 	virtual ~server_info();
 	server_info & operator=(const server_info &);
 	void dup_server_psets(const std::unordered_map<std::string, node_partition*>& spsets);
+	void free_server_psets();
```
bhroam, line 487: Can this be private?

src/scheduler/server_info.cpp
```diff
@@ -980,9 +976,15 @@ free_server_psets(std::unordered_map<std::string, node_partition *>& spsets)
 void
 server_info::dup_server_psets(const std::unordered_map<std::string, node_partition*>& spsets)
 {
-
-	for (const auto& spset : spsets)
-		svr_to_psets[spset.first] = dup_node_partition(spset.second, this);
+	for (const auto& spset : spsets) {
```
bhroam, line 979: Can this be private?

src/scheduler/server_info.cpp
```diff
@@ -980,9 +976,15 @@ free_server_psets(std::unordered_map<std::string, node_partition *>& spsets)
...
+		node_partition *temp = dup_node_partition(spset.second, this);
+		if (temp != NULL)
+			svr_to_psets[spset.first] = temp;
```
bhroam, line 982: You can shrink this code a little by getting rid of temp and assigning directly into svr_to_psets.  You can then check svr_to_psets[spset.first] == NULL

bhroam (changes requested):

src/include/limits_if.h
```diff
@@ -240,7 +240,7 @@ void update_soft_limits(server_info *, queue_info *, resource_resv *);
  *		the counts structure
  * @return	int
  */
-int find_preempt_bits(counts *, const char *, resource_resv *);
+int find_preempt_bits(counts *, std::string &, resource_resv *);
```
vstumpf, line 243: doing this makes this a c++ header, and therefore you can remove the `#ifdef __cplusplus` and `extern "C" {` stuff, since it can no longer be imported by C.

bhroam: You might as well move this into the scheduler directory while you are at it.  The scheduler is the only one who uses it.

arungrover: done!

arungrover:
Removed WIP status. One of the two test failures reported in regression passed when I ran it again. Another failure is in the qdel test which isn't related.

src/scheduler/server_info.cpp
```diff
@@ -286,15 +282,14 @@ query_server(status *pol, int pbs_sd)
...
 	 */
 	if ((policy->round_robin == 1) || (policy->by_queue == 1))
-		qsort(sinfo->queues, sinfo->num_queues, sizeof(queue_info *),
-			cmp_queue_prio_dsc);
+		std::sort(sinfo->queues.begin(), sinfo->queues.end(), cmp_queue_prio_dsc);
 	if (policy->round_robin == 1) {
 		/* queues are already sorted in descending order of their priority */
-		for (i = 0; i < sinfo->num_queues; i++) {
-			auto ret_val = add_queue_to_list(&sinfo->queue_list, sinfo->queues[i]);
+		for (auto queue: sinfo->queues) {
```
vstumpf, line 288: can you put a space before the `:`

src/scheduler/server_info.cpp
```diff
@@ -2599,60 +2467,34 @@ is_unassoc_node(node_info *ninfo, void *arg)
...
-		return;
-
-	if (cts->name != NULL)
-		free(cts->name);
-
-	if (cts->rescts != NULL)
-		free_resource_count_list(cts->rescts);
+// counts copy constructor
+counts::counts(const counts &rcount) {
+        name = rcount.name;
```
vstumpf, line 2480: I see some weird tabs here

src/scheduler/server_info.cpp
```diff
@@ -2666,89 +2508,26 @@ free_counts(counts *cts)
...
-				prev->next = ncts;
-
-			prev = ncts;
-		}
-		cur = cur->next;
-	}
-
-	return nhead;
+	counts_umap nmap;
+	for (auto iter: omap)
```
vstumpf, line 2528: This creates a copy, do you want to get it by reference?

src/scheduler/sort.cpp
```diff
@@ -280,15 +280,10 @@ cmp_nspec_by_sub_seq(const void *v1, const void *v2)
  * @retval	0 	: if q1 == q2
  * @retval	-1  : if q1 > q2
  */
-int
-cmp_queue_prio_dsc(const void *q1, const void *q2)
+bool
+cmp_queue_prio_dsc (const queue_info *q1, const queue_info *q2)
```
vstumpf, line 284: What if either are null? Do you want to pass in references instead?

vstumpf (changes requested):

src/scheduler/limits.cpp
```diff
@@ -717,91 +725,49 @@ has_softlimits(void *p)
...
  * @brief
- *		create a new limit count structure and initialize it.
- *
- * @return	limit count structure
- * @retval	NULL	: Failed to create a limit count structure.
+ *		limitcount class constructor.
  */
-static limcounts *
-new_limcounts(void)
+limcounts::limcounts()
```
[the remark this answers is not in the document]
arungrover: No reason, I should probably delete this constructor.

src/scheduler/limits.cpp
```diff
@@ -717,91 +725,49 @@ has_softlimits(void *p)
...
-	}
-	lc->all = dup_counts_list(all);
-	if ((lc->all == NULL) && (all != NULL)) {
-		free_limcounts(lc);
-		return NULL;
-	}
+// Assignment operator
+limcounts & limcounts::operator=(const limcounts &rlimit)
+{
+	user = rlimit.user;
```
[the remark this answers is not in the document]
arungrover: Yeah, that caused me trouble. I've changed the code to do deep copy in copy constructor but forgot to add it here. Good catch!

src/scheduler/sort.cpp
vstumpf: Oh i see, it's a vector of pointers. You still might wnat to check for `nullptr` though.

src/scheduler/limits.cpp
```diff
@@ -1249,20 +1157,19 @@ check_server_max_user_run(server_info *si, queue_info *qi, resource_resv *rr,
 	limcounts *sc, limcounts *qc, schd_error *err)
 {
 	char		*key;
-	char		*user = rr->user;
+	std::string	user = rr->user;
```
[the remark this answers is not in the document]
arungrover: As discussed, we will keep it as it is as the strings it is going to create are not going to be very small strings.

src/scheduler/node_partition.cpp
```diff
@@ -426,17 +423,17 @@ create_node_partitions(status *policy, node_info **nodes, const char * const *re
...
 		for (node_i = 0; nodes[node_i] != NULL; node_i++) {
 			if (nodes[node_i]->is_stale)
 				continue;
 
 			res = find_resource(nodes[node_i]->res, def);
 
 			if (res == NULL && (flags & NP_CREATE_REST)) {
-				unset_res->name = resnames[res_i];
+				unset_resname = res_i;
+				unset_res->name = unset_resname.c_str();
```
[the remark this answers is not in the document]
arungrover: You are right, unset_resname isn't needed.

src/scheduler/queue_info.cpp
```diff
@@ -108,7 +108,7 @@ query_queues(status *policy, int pbs_sd, server_info *sinfo)
 	struct batch_status *cur_queue;
 
 	/* array of pointers to internal scheduling structure for queues */
-	queue_info **qinfo_arr;
+	std::vector<queue_info *> qinfo_arr = {};
```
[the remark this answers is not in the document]
arungrover: It is hard to say how many queues would a complex have.

src/scheduler/sort.cpp
arungrover: These can't be nullptr right? it is called from std::sort which shouldn't be passing a null pointer to sort function.

src/scheduler/node_partition.cpp
```diff
@@ -770,23 +746,40 @@ node_partition_update(status *policy, node_partition *np)
...
  */
+np_cache::np_cache() {
+	ninfo_arr = NULL;
+	nodepart = NULL;
+	num_parts = UNSPECIFIED;
+}
+// copy constructor
+np_cache::np_cache(const np_cache & rnp_cache) {
+	ninfo_arr = rnp_cache.ninfo_arr;
+	nodepart = rnp_cache.nodepart;
```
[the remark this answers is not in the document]
arungrover: You are right, on second thought I don't think we need the copy constructor or assignment operator. We don't even copy over npc_arr when we duplicate server_info class. I guess I'll delete both copy constructor and assignment operator.

src/scheduler/sort.cpp
vstumpf: std::sort will pass a reference to the item being sorted, in this case, we're sorting `queue_info *`. Someone can do `vec.push_back(nullptr);`, sort the vector, and this will crash.

arungrover:
@vstumpf Thanks for the review, I have addressed your comments.

src/scheduler/sort.cpp
arungrover: True, but I think we should be adding such a check when we push_back into vectors. In our case, it is not possible for now, because we are adding queue_info pointers after allocating them and if new fails it is going to raise an exception anyways.

arungrover:
@vstumpf @bhroam  Can you please review the latest commits? I have also rebased it.

src/scheduler/misc.cpp
```diff
@@ -489,6 +491,31 @@ enum match_string_array_ret match_string_array(const char * const *strarr1, cons
...
+
+	if (strarr1.empty() || strarr2.empty())
+		return SA_NO_MATCH;
+
+	for (auto &str1: strarr1) {
+		if (std::find(strarr2.begin(), strarr2.end(), str1) != strarr2.end())
+			match++;
+	}
+
+	/* i is the length of strarr1 since we just looped through the whole array */
```
vstumpf, line 507: This comment doesn't make sense anymore

src/scheduler/node_partition.cpp
```diff
@@ -916,16 +869,13 @@ find_alloc_np_cache(status *policy, np_cache ***pnpc_arr,
 			if (sort_func != NULL)
 				qsort(nodepart, num_parts, sizeof(node_partition *), sort_func);
 
-			npc = new_np_cache();
+			npc = new np_cache();
```
vstumpf, line 872: npc will never be null, as new will raise an exception if it fails.

do you want to put a try/catch block here?

src/scheduler/queue_info.cpp
```diff
@@ -568,19 +532,16 @@ queue_info::queue_info(const char *qname): name(qname)
  */
 
 void
-free_queues(queue_info **qarr)
+free_queues(std::vector<queue_info *> &qarr)
```
vstumpf, line 535: This function will delete all the queue_info *s, but will not actually clear the vector, is that ok? You'll have a vector of freed pointers. Maybe add a .clear()?

src/scheduler/sort.cpp
vstumpf: That makes sense.

vstumpf (changes requested):
just some nitpicks

src/scheduler/simulate.cpp
```diff
@@ -1919,25 +1919,25 @@ policy_change_info(server_info *sinfo, resource_resv *resresv)
 		return 0;
 	}
 
-	if (sinfo->queues != NULL) {
+	if (!sinfo->queues.empty()) {
```
bhroam, line 1922: I don't think you need this block any more.  It was there to stop dereferencing a NULL pointer.  The bottom 3 ifs will just fall through if sinfo->queues is empty.

src/scheduler/sort.cpp
bhroam: remove the space between the function name and the paren.

bhroam (changes requested):

src/scheduler/limits.cpp
```diff
@@ -131,33 +128,41 @@
...
+	public:
+	counts_umap user;
+	counts_umap group;
+	counts_umap project;
+	counts_umap all;
+	limcounts() = delete;
+	limcounts(const counts_umap &user_,
+		  const counts_umap &group_,
+		  const counts_umap &project_,
+		  const counts_umap &all_);
```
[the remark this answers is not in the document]
arungrover: I changed in most places, must have missed this one. I'll change it.

src/scheduler/limits.cpp
```diff
@@ -2653,11 +2546,12 @@ check_queue_max_res_soft(server_info *si, queue_info *qi, resource_resv *rr)
...
  */
 static int
-check_max_group_res(resource_resv *rr, counts *cts_list, resdef **rdef, void *limitctx)
+check_max_group_res(resource_resv *rr, counts_umap &cts_list,
+	resdef **rdef, void *limitctx)
 {
 	char		*groupreskey;
 	char		*gengroupreskey;
-	char		*group;
+	std::string	group = rr->group;
```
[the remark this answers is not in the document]
arungrover: oops

src/scheduler/limits.cpp
```diff
@@ -3418,7 +3330,12 @@ schderr_args_server(const char *entity, schd_error *err)
...
 	set_schd_error_arg(err, ARG1, (char*)entity);
 }
-
+// overloaded
+static void
+schderr_args_server(const std::string &entity, schd_error *err)
+{
+	set_schd_error_arg(err, ARG1, entity.c_str());
+}
 /**
```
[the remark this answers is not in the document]
arungrover: As discussed, In some places It is called with NULL as entity. So we will need the overloaded function.

src/scheduler/limits.cpp
```diff
@@ -3429,11 +3346,11 @@ schderr_args_server(const char *entity, schd_error *err)
  * @param[out]	err	-	schd_error structure to return error information
  */
 static void
-schderr_args_server_res(const char *entity, const char *res, schd_error *err)
+schderr_args_server_res(std::string &entity, const char *res, schd_error *err)
 {
 	set_schd_error_arg(err, ARG1, (char*)res);
```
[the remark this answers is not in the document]
arungrover: Actually it already does take const char *, this must be old code.

src/scheduler/limits.cpp
```diff
@@ -3456,20 +3373,20 @@ schderr_args_server_res(const char *entity, const char *res, schd_error *err)
...
-	char		*project;
+	std::string	project;
 	sch_resource_t	max_project_res;
 	sch_resource_t	max_genproject_res;
 	sch_resource_t	used = 0;
 
 	if (rr == NULL)
 		return (-1);
-	if ((limres == NULL) || (rr->resreq == NULL) || (rr->project == NULL))
+	if ((limres == NULL) || (rr->resreq == NULL) || (rr->project.empty()))
```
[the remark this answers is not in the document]
arungrover: That check is right above this if condition.

src/scheduler/limits.cpp
```diff
@@ -3548,7 +3465,7 @@ check_max_project_res_soft(resource_resv *rr, counts *cts_list, void *limitctx,
...
```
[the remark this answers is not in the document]
arungrover: Same here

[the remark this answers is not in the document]
bhroam: oops, I didn't notice.

src/scheduler/node_partition.cpp
```diff
@@ -916,16 +869,13 @@ find_alloc_np_cache(status *policy, np_cache ***pnpc_arr,
...
 			if (npc != NULL) {
```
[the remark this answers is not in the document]
arungrover: You are right, based on Vincent's comment, I added a try-catch block here.

src/scheduler/queue_info.cpp
```diff
@@ -795,30 +731,22 @@ queue_info::~queue_info()
...
-		(nsinfo->num_queues + 1) * sizeof(queue_info*)))) == NULL) {
-		log_err(errno, __func__, MEM_ERR_MSG);
-		return NULL;
-	}
-
-	for (i = 0; oqueues[i] != NULL; i++) {
-		if ((new_queues[i] = new queue_info(*oqueues[i], nsinfo)) == NULL) {
+	for (auto queue: oqueues) {
+		queue_info *temp;
+		if ((temp = new queue_info(*queue, nsinfo)) == NULL) {
```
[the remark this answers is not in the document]
arungrover: I'll add this into a try-catch block.

src/scheduler/sched_exception.cpp
```diff
@@ -0,0 +1,90 @@
...
+// Getter function for message
+const std::string sched_exception::get_message() const
+{
+    return message;
+}
+
+// Destructor
+sched_exception::~sched_exception()
+{
+}
```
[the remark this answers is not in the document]
arungrover: True, I'll remove it.

src/scheduler/server_info.cpp
```diff
@@ -504,19 +489,15 @@ query_server_info(status *pol, struct batch_status *server)
 	if (pol == NULL || server == NULL)
 		return NULL;
 
-	if ((sinfo = new_server_info(1)) == NULL)
-		return NULL;		/* error */
+	sinfo = new server_info(1);
```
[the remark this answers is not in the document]
arungrover: I must have missed the comment

src/scheduler/server_info.cpp
arungrover: Sorry, I forgot to respond. dup_node_partition() can return NULL in some cases.

src/scheduler/server_info.cpp
```diff
@@ -286,15 +282,14 @@ query_server(status *pol, int pbs_sd)
...
 	 */
 	if ((policy->round_robin == 1) || (policy->by_queue == 1))
-		qsort(sinfo->queues, sinfo->num_queues, sizeof(queue_info *),
-			cmp_queue_prio_dsc);
+		std::sort(sinfo->queues.begin(), sinfo->queues.end(), cmp_queue_prio_dsc);
 	if (policy->round_robin == 1) {
 		/* queues are already sorted in descending order of their priority */
-		for (i = 0; i < sinfo->num_queues; i++) {
-			auto ret_val = add_queue_to_list(&sinfo->queue_list, sinfo->queues[i]);
+		for (auto queue : sinfo->queues) {
```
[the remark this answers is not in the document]
arungrover: I'll change the variable name, I can't combine the loops because this loop has to happen only in the case of round_robin policy and that too after the queues have been sorted.

src/scheduler/server_info.cpp
```diff
@@ -2468,23 +2449,26 @@ is_unassoc_node(node_info *ninfo, void *arg)
 }
 
 // counts constructor
-counts::counts(const std::string &name_){
+counts::counts(const std::string &name_)
```
bhroam, line 2452: Still one left with an underscore

bhroam (changes requested):
Just a few really minor comments.

src/scheduler/server_info.cpp
arungrover: oops

arungrover:
Description: Tests from given sources on platforms UBUNTU1804, UBUNTU1604, CENTOS8, RHEL8, CENTOS7, RHEL7, SLES12, SLES15, UBUNTU2004
Platforms: UBUNTU1804,UBUNTU1604,CENTOS8,RHEL8,CENTOS7,RHEL7,SLES12,SLES15,UBUNTU2004
Profile: regression
|--|--|--|--|--|--|--|
|8342|4117|3|0|0|0|4114|

bhroam (approved):
Thanks for making all the changes.  I sign off.

arungrover:
@vstumpf can you please have a look again?

vstumpf (approved):
LGTM

PR #2436, opened 2021-06-30
Fixed preemption by checkpointing to fail if it is not enabled

bhroam:
#### Describe Bug or Feature
If checkpoint is in the preempt_order, but it is not enabled (no checkpoint_abort script), the checkpoint preempt method will "succeed", but the job will remain in the running state.  This will cause the scheduler to think it has been preempted, and will run the high priority job.  This oversubscribes resources.

#### Describe Your Change
If there is no checkpoint_abort script, the mom returns PBSE_NOSUSP.  In post_hold() on the server side, it ignores PBSE_NOSUSP when checking for the error code.  This would then fall into the case for successful preemption.

The fix is to fail preemption if the pbs error code returned from mom is not PBSE_NONE, regardless if it is PBSE_NOSUSP or not.

I also refactored the if statement in post_hold() a little to make it more clear what was going on.

Profile: regression
|--|--|--|--|--|--|--|
|8018|4177|1|109|0|0|4067|

And the rerun:
Description: Rerun Only Failed Tests From #8018
Profile: regression
|--|--|--|--|--|--|--|
|8030|110|0|0|0|0|110|

So all tests passed.

### DIFF
--- a/src/server/req_holdjob.c
+++ b/src/server/req_holdjob.c
@@ -471,22 +471,22 @@ post_hold(struct work_task *pwt)
 		conn->cn_authen &= ~PBS_NET_CONN_NOTIMEOUT;
 	}
 
-	if (code != 0) {
+	if (code != PBSE_NONE) {
 		/* Checkpoint failed, remove checkpoint flags from job */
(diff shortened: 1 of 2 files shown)


arungrover (approved):
LGTM.

agrawalravi90 (approved):

PR #2437, opened 2021-06-30
fix tpp crash and simplify locks to avoid races

subhasisb:
#### Describe Bug or Feature

1. The comm router code incorrectly accesses a fd which has been set to -1 earlier and indexes on it and crashes. The stack trace is as follows:
`#0 0x0000000000409155 in handle_disconnect (conn=conn@entry=0x7efef8002500) at /home/pbsbuild/ramdisk/workspace/build/openpbs/src/lib/Libtpp/tpp_transport.c:1745
#1 0x0000000000409274 in add_pkt (conn=conn@entry=0x7efef8002500) at /home/pbsbuild/ramdisk/workspace/build/openpbs/src/lib/Libtpp/tpp_transport.c:1892
#2 0x0000000000409378 in handle_incoming_data (conn=conn@entry=0x7efef8002500)
at /home/pbsbuild/ramdisk/workspace/build/openpbs/src/lib/Libtpp/tpp_transport.c:1850
#3 0x000000000040a1e0 in work (v=0x24d5e20) at /home/pbsbuild/ramdisk/workspace/build/openpbs/src/lib/Libtpp/tpp_transport.c:1576
#4 0x00007eff06943e65 in start_thread () from /lib64/libpthread.so.0
#5 0x00007eff0601b88d in clone () from /lib64/libc.so.6`

The change to fix this it to not set the conn_fd = -1 in function router_close_handler_inner(). Also the caller, handle_disconnect() has been modified to deal with it, in case the handler reset the fd to -1. 

2. While walking the leaves tree in the comm router, the locks are released too early, and another thread could remove the leaves in between the time the index walk happened, and the usage. Moved the lock/unlocks to a more consistent around the entire functions of "broadcast_to_my_leaves" "broadcast_to_my_routers" and "send_leaves_to_router" functions (instead of inside them)

Helgrind test logs
[comm.log](https://github.com/openpbs/openpbs/files/6740473/comm.log)
BTW, the single stack trace listed by helgrind is well understood and is a false positive. 

Description: Rerun All Tests From #8037
Profile: regression
|--|--|--|--|--|--|--|
|8040|4175|1|2|0|1|4171|

### DIFF
--- a/src/lib/Libtpp/tpp_transport.c
+++ b/src/lib/Libtpp/tpp_transport.c
@@ -947,23 +947,17 @@ static int
 tpp_post_cmd(int tfd, char cmd, tpp_packet_t *pkt)
 {
 	int rc;
+	int slot_state;
 	phy_conn_t *conn = NULL;
 	thrd_data_t *td = NULL;
 
 	errno = 0;
 
-	if (tpp_read_lock(&cons_array_lock))
-		return -1;
+	conn = get_transport_atomic(tfd, &slot_state);
+	if (conn)
+		td = conn->td;
 
-	if (tfd >= 0 && tfd < conns_array_size) {
-		if (conns_array[tfd].conn && conns_array[tfd].slot_state == TPP_SLOT_BUSY) {
-			conn = conns_array[tfd].conn;
-			td = conn->td;
-		}
-	}
-
-	if (!td || !conn) {
-		tpp_unlock_rwlock(&cons_array_lock);
+	if (!conn || slot_state != TPP_SLOT_BUSY || !td) {
 		errno = EBADF;
 		return -1;
 	}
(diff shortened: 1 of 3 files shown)


src/lib/Libtpp/tpp_transport.c
```diff
@@ -980,8 +974,6 @@ tpp_post_cmd(int tfd, char cmd, tpp_packet_t *pkt)
```
hirenvadalia, line 974: we don't need `tpp_unlock_rwlock(&cons_array_lock);` anymore at line 970 above

Also at line number 866, it will be safe to put `free_router(r);` under lock/unlock

hirenvadalia (changes requested):

hirenvadalia (approved):

suresh-thelkar (approved):

PR #2446, opened 2021-07-08
Fix deleting multi sched in TestMultipleSchedulers.test_list_multi_sched

shilpakodli:
#### Describe Bug or Feature
Test test_multiple_queue_same_partition and test_pbsfs_revert_to_defaults of TestMultipleSchedulers are failing while verifying log match due to cleanup has not done properly in previous test which is test_list_multi_sched.

#### Describe Your Change
In test case test_list_multi_sched deleting scheduler sc1 and sc3 using qmgr command which leaves sched processed "pbs_sc1" and "pbs_sc3" running behind and also respective directories will be present under PBS Home directory.
In order to delete sched, we need to first kill the process, remove relative folders from PBS Home directory and then delete using qmgr.

Profile: regression
|--|--|--|--|--|--|--|
|8113|570|0|0|0|10|560|

### DIFF
--- a/test/tests/functional/pbs_multi_sched.py
+++ b/test/tests/functional/pbs_multi_sched.py
@@ -136,6 +136,17 @@ class TestMultipleSchedulers(TestFunctional):
                 self.server.manager(MGR_CMD_SET, SCHED, {'scheduling': op},
                                     id=each)
 
+    def delete_sched(self, sched_name):
+        """
+        Helper function to delete sched"
+        """
+        self.scheds[sched_name].terminate()
+        sched_log = self.scheds[sched_name].attributes['sched_log']
+        sched_priv = self.scheds[sched_name].attributes['sched_priv']
+        self.du.rm(path=sched_log, sudo=True, recursive=True, force=True)
+        self.du.rm(path=sched_priv, sudo=True, recursive=True, force=True)
+        self.server.manager(MGR_CMD_DELETE, SCHED, id=sched_name)
+
     def test_job_sort_formula_multisched(self):
         """
         Test that job_sort_formula can be set for each sched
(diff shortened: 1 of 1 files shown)


test/tests/functional/pbs_multi_sched.py
```diff
@@ -136,6 +136,17 @@ def set_scheduling(self, scheds=None, op=False):
...
+    def delete_sched(self, sched_name):
+        """
+        Helper function to delete sched"
+        """
+        self.scheds[sched_name].terminate()
+        sched_log = self.scheds[sched_name].attributes['sched_log']
+        sched_priv = self.scheds[sched_name].attributes['sched_priv']
+        self.du.rm(path=sched_log, sudo=True, recursive=True, force=True)
+        self.du.rm(path=sched_priv, sudo=True, recursive=True, force=True)
+        self.server.manager(MGR_CMD_DELETE, SCHED, id=sched_name)
```
bhroam, line 148: This helper function is fine where it is, but it could be useful in the PTL scheduler class.  We already have a create_scheduler() function which makes them.  Having a function which deletes one would be useful too.

bhroam (approved):
I'll sign off.  I'll leave it up to you if you want to move delete_sched() into the scheduler class.

dinesh1306 (approved):

test/tests/functional/pbs_multi_sched.py
shilpakodli: Hi @bhroam , we have already function which deletes multisched in ptl_server.py file "delete_sched_config" but this deletes all the schedulers in setup. I had to write to delete individual sched.

bhroam: Maybe break the function into 2 parts?  One to delete one sched, and the other to loop over all the scheds and call the other function?  It'll be more useful in the future, but isn't necessary now.  Do you want to do that, or should I merge?

shilpakodli: Hey Bhroam, I have created a separate ticket for these changes.

PR #2448, opened 2021-07-09
Fasten log analysis by searching start timestamp from bottom using tail

visheshh:
Fasten log analysis by searching start timestamp from bottom using tail

#### Describe Bug or Feature
When running tests for a long duration as PBS logs grow, log analysis takes long amount of time as each and every time we parse all the logs for analysis.

Tests take 5-10 min extra as PBS logs become greater than 1gb and tests does log analysis.

#### Describe Your Change
When start time stamp is given check timestamp from last 10000th line in iteration of 10k lines to find the timestamp.
Create a tail command if start timestamp is given for n number of lines where the timestamp was found.

Observation : pbs_loganalyzer on a 4gb server log file took 30 mins
Before fix: pbs_loganalyzer on a 4gb server log file with begin timestamp of last one hour took 20 mins.
After fix: pbs_loganalyzer on a 4gb server log file with begin timestamp of last one hour took less than 5 secs.

time pbs_loganalyzer -s /var/spool/pbs/server_logs/20210715 -b '07/15/2021 17:00:00'
Parsing /var/spool/pbs/server_logs/20210715: |0%-100%|

real    0m36.034s
user    0m29.771s
sys     0m10.042s

 time pbs_loganalyzer -s /var/spool/pbs/server_logs/20210715 -b '07/15/2021 17:00:00'
Parsing /var/spool/pbs/server_logs/20210715: |0%-10%-20%-30%-40%-50%-60%-70%-80%-90%-100%-100%|

real    11m44.785s
user    11m38.054s
sys     0m4.440s

[Test_scof_wfix.txt](https://github.com/openpbs/openpbs/files/6824429/Test_scof_wfix.txt)
[Test_scof_wofix.txt](https://github.com/openpbs/openpbs/files/6824430/Test_scof_wofix.txt)

With change test took : 1:48 mins
Without change test took: 11:42 mins

### DIFF
--- a/test/fw/ptl/utils/pbs_logutils.py
+++ b/test/fw/ptl/utils/pbs_logutils.py
@@ -291,6 +291,13 @@ class PBSLogUtils(object):
         taillogs = 10000
         tailcmd = [self.du.which(hostname, 'tail')]
         headcmd = [self.du.which(hostname, 'head')]
+        if sudo:
+            tailcmd = self.du.sudo_cmd + tailcmd
+            headcmd = self.du.sudo_cmd + headcmd
+        if hostname:
+            tailcmd = ['ssh', hostname] + tailcmd
+            headcmd = ['ssh', hostname] + headcmd
+
         if start:
             i = 0
             while(True):
--- a/test/fw/ptl/utils/pbs_logutils.py
+++ b/test/fw/ptl/utils/pbs_logutils.py
@@ -256,7 +256,7 @@ class PBSLogUtils(object):
             # Get epoch-timestamp assuming local timezone
             tm = t.timestamp()
         except ValueError:
-            cls.logger.debug("could not convert date time: " + str(dt))
+            cls.logger.info("could not convert date time: " + str(dt))
             return None
 
         if micro is True:
--- a/test/fw/ptl/utils/pbs_logutils.py
+++ b/test/fw/ptl/utils/pbs_logutils.py
@@ -289,13 +289,15 @@ class PBSLogUtils(object):
         """
         readcmd = ['cat', log]
         taillogs = 10000
+        tailcmd = [self.du.which(hostname, 'tail')]
+        headcmd = [self.du.which(hostname, 'head')]
         if start:
             i = 0
             while(True):
                 i += 1
                 taillogs = 10000 * i
-                args = ['tail', '-n', str(taillogs), log]
-                args2 = ['head', '-n', '1']
+                args = tailcmd + ['-n', str(taillogs), log]
+                args2 = headcmd + ['-n', '1']
                 process_tail = Popen(args, stdout=PIPE,
                                      shell=False)
                 process_head = Popen(args2, stdin=process_tail.stdout,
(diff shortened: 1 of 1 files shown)


test/fw/ptl/utils/pbs_logutils.py
```diff
@@ -286,20 +287,51 @@ def open_log(self, log, hostname=None, sudo=False):
         :type sudo: boolean
         :returns: A file instance
         """
+        readcmd = ['cat', log]
+        taillogs = 10000
+        if start:
+            i = 0
+            while(True):
```
bhroam, line 294: This function works on log files on machines other than this one (the hostname parameter).  This code expects the log file to be local.

test/fw/ptl/utils/pbs_logutils.py
```diff
@@ -286,20 +287,51 @@ def open_log(self, log, hostname=None, sudo=False):
...
+                    break
+                elif taillogs > num_records:
+                    readcmd = ['cat', log]
+                    break
+
         try:
             if hostname is None or self.du.is_localhost(hostname):
                 if sudo:
-                    cmd = copy.copy(self.du.sudo_cmd) + ['cat', log]
+                    cmd = copy.copy(self.du.sudo_cmd) + readcmd
```
bhroam, line 322: No need for the copy.  When you add 2 lists together, you get a new list.  Now if you did
```
cmd = self.du.sudo_cmd
cmd += readcmd
```
You'd screw yourself because now sudo_cmd has changed as well.

test/fw/ptl/utils/pbs_logutils.py
```diff
@@ -286,20 +287,51 @@ def open_log(self, log, hostname=None, sudo=False):
...
+            i = 0
+            while(True):
+                i += 1
+                taillogs = 10000 * i
+                args = ['tail', '-n', str(taillogs), log]
+                args2 = ['head', '-n', '1']
+                process_tail = Popen(args, stdout=PIPE,
+                                     shell=False)
+                process_head = Popen(args2, stdin=process_tail.stdout,
+                                     stdout=PIPE, shell=False)
```
bhroam, line 302: There is self.du.tail().  You could use that.

bhroam (changes requested):

bhroam:
@visheshh Thanks for making the changes.  There are still two comments you haven't replied to.  The first is that your change uses Popen(), and that will only work on the local machine.  The second is why aren't you using self.du.head() and self.du.tail()?  If you were to use those functions, you could pass in a hostname.

visheshh:
> 
> 
> @visheshh Thanks for making the changes. There are still two comments you haven't replied to. The first is that your change uses Popen(), and that will only work on the local machine. The second is why aren't you using self.du.head() and self.du.tail()? If you were to use those functions, you could pass in a hostname.

Hi @bhroam , I have added sudo and hostname to the command if present.
with du.tail it returns all the lines and stores all the lines . I tried getting the output['out'][0] the first line from output and it took nearly 10 times more time than Popen tail which Pipes out data to head -n 1.

2021-07-12 22:58:02,372 INFOCLI2 blrentperf03(run_cmd): /usr/bin/tail -n 100000 /var/spool/pbs/server_logs/20210712
2021-07-12 22:58:02,569 INFO     Time taken to du.tail
2021-07-12 22:58:02,570 INFO     0.19795727729797363
2021-07-12 22:58:02,588 INFO     Time taken to tail head
2021-07-12 22:58:02,589 INFO     0.018006086349487305

test/fw/ptl/utils/pbs_logutils.py
```diff
@@ -291,6 +291,13 @@ def open_log(self, log, hostname=None, sudo=False, start=None,
         taillogs = 10000
         tailcmd = [self.du.which(hostname, 'tail')]
         headcmd = [self.du.which(hostname, 'head')]
+        if sudo:
+            tailcmd = self.du.sudo_cmd + tailcmd
+            headcmd = self.du.sudo_cmd + headcmd
+        if hostname:
+            tailcmd = ['ssh', hostname] + tailcmd
+            headcmd = ['ssh', hostname] + headcmd
+
```
bhroam, line 300: hostname can be None.  Why don't you want to use self.du.head() and self.du.tail()?  Those take care of all of this for you.  You are basically reinventing the wheel here.

bhroam:
I don't think you should be trying to get around issues in run_cmd() by reimplementing run_cmd() locally.  If you think run_cmd() is too slow, fix run_cmd().

bhroam (approved):
Thanks for making the changes, LGTM

bayucan (approved):
Looks good.

bayucan:
@visheshh : before I can merge, can you attach the test logs?

visheshh:
@bayucan , Please find the test logs.

PR #2450, opened 2021-07-13
Fixed post condition of *server_restart_immed tests

bayucan:
#### Describe Bug or Feature

* Two tests in the TestPbsNodeRampDown testsuite, test_release_nodes_excl_server_restart_immed and test_release_nodes_shared_server_restart_immed need to be updated due to an  incorrect assumption on node assignment after job restart.
* The two tests incorrectly assume that the job will get assigned exactly the same set of vnodes upon job rerun, after a 'qterm -t immediate' is done and server is restarted. 
* There's no guarantee that the SAME set of vnodes would be re-assigned to the job. For instance, say job requests 3 vnodes and has been assigned vnode1+vnode2+vnode3. After a bunch of pbs_release_nodes calls and then a qterm -t immediate followed up by a server start, the job is requeued to go back to requesting 3 vnodes. Upon job rerun, it could be assigned to another set of nodes like vnode4+vnode2+vnode6.

#### Describe Your Change

* Upon job restart, instead of matching exactly the exec_vnode value of before, do a regular expression match instead and not include the exact vnode names.

[logfile_PASS_test_release_nodes_excl_server_restart_immed.txt](https://github.com/openpbs/openpbs/files/6811053/logfile_PASS_test_release_nodes_excl_server_restart_immed.txt)
[logfile_PASS_test_release_nodes_shared_server_restart_immed.txt](https://github.com/openpbs/openpbs/files/6811054/logfile_PASS_test_release_nodes_shared_server_restart_immed.txt)

### DIFF
--- a/test/tests/functional/pbs_node_rampdown.py
+++ b/test/tests/functional/pbs_node_rampdown.py
@@ -332,7 +332,7 @@ class TestPbsNodeRampDown(TestFunctional):
         self.server.manager(MGR_CMD_SET, NODE, a, id=self.hostC)
 
         a = {'state': 'free', 'resources_available.ncpus': (GE, 1)}
-        self.server.expect(VNODE, {'state=free': 11}, op=EQ, count=True,
+        self.server.expect(VNODE, {'state=free': 11}, count=True,
                            interval=2)
 
         # Various node names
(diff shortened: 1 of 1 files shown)


nishiya (changes requested):
Just one nit: op=EQ is the default, can you please remove it  (line 335)?

bhroam (approved):
LGTM

nishiya (approved):
Thanks for making that addition change @bayucan.
Looks good!

PR #2451, opened 2021-07-14
Make sure no nodes are in down state after server restarts

visheshh:
#### Describe Bug or Feature
After server restart, nodes may be down for a few seconds till they reconnect with server, which causes qdel failures where it cannot connect to Mom which are issues as soon as server restart.

#### Describe Your Change
Check nodes are in same state as before and after server restart

Description: Tests from given sources on platforms SUSE15, CENTOS8, CENTOS7, UBUNTU2004
Platforms: SUSE15,CENTOS8,CENTOS7,UBUNTU2004
Profile: regression
|--|--|--|--|--|--|--|
|8154|1243|3|1|0|0|1239|

### DIFF
--- a/test/fw/ptl/lib/ptl_server.py
+++ b/test/fw/ptl/lib/ptl_server.py
@@ -286,7 +286,7 @@ class Server(Wrappers):
         Terminate and start a PBS server.
         """
         cmd = os.path.join(self.client_conf['PBS_EXEC'], 'bin',
-                          'pbsnodes') + ' -av' + ' -Fjson'
+                           'pbsnodes') + ' -av' + ' -Fjson'
         cmd_out = self.du.run_cmd(self.hostname, cmd, sudo=True)
         pbsnodes_json = json.loads('\n'.join(cmd_out['out']))
         if self.isUp():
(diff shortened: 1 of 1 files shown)


kjakkali (approved):

bhroam (commented):
I'm not sure I agree with this.  This can take a long time if you have a lot of vnodes.  Not only that, it adds a race.  If  node is in job-busy before the restart, the job might end during the restart.  The node will come up in state free.

visheshh:
Hi @bhroam, Can we check If node state is state-unknown,down and wait till it gets to free state.

bhroam:
Well there are two issues here.  I can tell you that creating 10k vnodes takes a long time because it will check that all 10k vnodes are in the free state.  You are making restart() do the same thing.  Then there is the race where the state could change during the restart.  Maybe do an expect for {'state=down,unknown': 0}'

visheshh:
@bhroam, @kjakkali,  Added an expect {'state=down,unknown': 0}' after server restart

bhroam (approved):
LGTM

bhroam:
@kjakkali can you take another look?  The implementation changed since you last looked.

kjakkali (approved):

PR #2454, opened 2021-07-16
Check nodes are free after create_moms creates them

visheshh:
#### Describe Bug or Feature
After moms are created by create_moms, nodes take some time to connect server.

#### Describe Your Change
After moms are created wait for the nodes to be in free state.

### DIFF
--- a/test/fw/ptl/lib/ptl_server.py
+++ b/test/fw/ptl/lib/ptl_server.py
@@ -1920,6 +1920,8 @@ class Server(Wrappers):
                     if rc != 0:
                         self.logger.error("error creating node " + _n)
                         error = True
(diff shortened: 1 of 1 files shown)


bhroam:
Won't this happen in setUp() when the nodes are deleted and recreated?

visheshh:
@bhroam , The function create_moms is called from tests, where we want to create start multiple mom processes on different ports. Now used in job performance tests, the nodes are not checked for free state. So adding this check to function itself.

bhroam (changes requested):
This will fail if delall is not True.  You'd be adding a mom then, and then the number of free nodes would be greater than num_nodes.  Could you change this to (GE, num_nodes)

PR #2459, opened 2021-07-21
qdel is getting hanged while deleting array jobs

nithinj:
#### Describe Bug or Feature
* qdel is getting hanged while deleting array jobs
The hang is because of two reasons:
1. Server doesn't create a worktask to track the deljob list when a subjob deletion fails
2. Total array job reply count was not getting updated

#### Describe Your Change
* Create work tasks to handle the deljoblist even when subjob deletion fails
* Update total array job count whenever the reply count is getting updated for an array job.
* Reorder the deljoblist to avoid any deletion triggered runs from the same list
* Modified PTL framework to delete list of id's together instead of deleting one by one.

```
[root@head1 tmp]# qstat
[root@head1 tmp]# for i in {1..4}; do qsub -J1-6 -- /bin/sleep 100; done; qstat; qdel `qselect`
586[].head1
587[].head1
588[].head1
589[].head1
586[].head1       STDIN            root                     0 B workq           
587[].head1       STDIN            root                     0 Q workq           
588[].head1       STDIN            root                     0 Q workq           
589[].head1       STDIN            root                     0 Q workq
```

### DIFF
--- a/src/include/batch_request.h
+++ b/src/include/batch_request.h
@@ -107,7 +107,7 @@ struct rq_manage {
 struct rq_deletejoblist {
 	int rq_count;
 	char **rq_jobslist;
-	int rq_resume;
+	bool rq_resume;
 	int jobid_to_resume;
 	int subjobid_to_resume;
 };
--- a/src/include/job.h
+++ b/src/include/job.h
@@ -1008,7 +1008,7 @@ extern int   modify_job_attr(job *, svrattrl *, int, int *);
 extern char *prefix_std_file(job *, int);
 extern void  cat_default_std(job *, int, char *, char **);
 extern int   set_objexid(void *, int, attribute *);
-extern bool update_deljob_rply(struct batch_request *, int, char *, int);
+extern bool update_deljob_rply(struct batch_request *, char *, int);
 #if 0
 extern int   site_check_user_map(job *, char *);
 #endif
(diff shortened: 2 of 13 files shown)


bremanandjk:
One overarching comment, since you are adding a sorting block for each qdeljoblist request with time complexity of O(n), It would be better to run the qdel performance test with 100K jobs. 

The earlier performance metrics are
qdel request took 13m11.304s for deleting 100k jobs and after the enhancement, the qdel took 25 seconds.

nithinj:
@bremanandjk 
Please find the performance numbers:
```
Mainline
====================================================
[root@blrentperf04 target]# qselect |wc -l
100000
[root@blrentperf04 target]# time qselect | xargs qdel

real	0m29.521s
user	0m0.313s
sys	0m0.144s
[root@blrentperf04 target]# qselect |wc -l
0

Using counters with fixes and ordering
=====================================================
[root@blrentperf04 target]# qselect |wc  -l
100001
[root@blrentperf04 target]# time qselect | xargs qdel

real	0m29.763s
user	0m0.339s
sys	0m0.155s
[root@blrentperf04 target]# qselect |wc  -l
0

Using AVL tree instead of counters
======================================================
[root@blrentperf04 target]# qselect |wc  -l
100001
[root@blrentperf04 target]# time qselect | xargs qdel

real	0m30.590s
user	0m0.345s
sys	0m0.150s
[root@blrentperf04 target]# qselect |wc  -l
0
0
```

PR #2460, opened 2021-07-21
Add generic interface to allow external libraries to handle web requests 

vstumpf:
#### Describe Bug or Feature
Add generic interface to allow external libraries to handle web requests.
Also, some *totally* unrelated bug fixes are introduced in this as well.
Introduced pbs_asprintcatf, a performant way to concatenate a formatted string to a buffer, with automatic reallocation if the buffer must be increased.

#### Describe Your Change
Add interface to handle webapi calls.

#### Link to Design Doc
N/A - There is no external functional changes in openpbs, so therefore no design documents are required.

Description: Tests from given sources on platforms UBUNTU1804, UBUNTU1604, CENTOS8, RHEL8, CENTOS7, RHEL7, SLES12, UBUNTU2004
Platforms: UBUNTU1804,UBUNTU1604,CENTOS8,RHEL8,CENTOS7,RHEL7,SLES12,UBUNTU2004
Profile: regression
|--|--|--|--|--|--|--|
|8238|3282|1|0|0|0|3281|

The failure is unrelated, TestMultipleSchedulers.test_prime_time_backfill

### DIFF
--- a/src/lib/Libutil/misc_utils.c
+++ b/src/lib/Libutil/misc_utils.c
@@ -618,7 +618,7 @@ pbs_asprintf_exit:
  * @param[in] ... - arguments to format string
  *
  * @return int
- * @retval -1 - Error
+ * @retval -1 - Error - No new allocation or freeing done.
  * @retval >=0 - Length of new string, not including terminator
  */
 int
@@ -629,6 +629,10 @@ pbs_asprintcatf(char **dest, size_t *size, size_t *idx, const char *fmt, ...)
 	size_t new_size = *size;
 	char * newbuf = NULL;
 
+	if (!size)
+		return -1;
+	if (!idx)
+		return -1;
 	if (!dest)
 		return -1;
 	if (!fmt)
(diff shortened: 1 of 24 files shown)


src/lib/Libutil/misc_utils.c
```diff
@@ -605,6 +605,69 @@ pbs_asprintf(char **dest, const char *fmt, ...)
...
+	if (len < 0)
+		goto pbs_asprintcatf_exit;
+
+	while (*idx + len > new_size) {
+		new_size *= 2;
+	}
+	if (new_size != *size) {
+		newbuf = realloc(*dest, new_size);
+		if (newbuf == NULL) {
+			len = -1;
```
arungrover, line 655: So if we allocate dest in this function and then fails to realloc it then do we expect the caller to free dest?

arungrover (commented):
minor comments

src/lib/Libutil/misc_utils.c
vstumpf: That is correct. I'll add it to the docstring

